PR #1345, opened 2019-01-31
feat(error-handling): add missing error handlers, upgrade a11y pkgs

seanforyou23:
This PR aims to further stablize our a11y test suite by upgrading related packages to the latest version and adding some strategic timeouts before running each page audit so that we can be sure we're auditing pages when they are in an ideal state, and not still in the process of rendering. I also added some missing catch() handlers to some of the reporter internals, these were resulting in `UnhandledPromiseRejectionWarning` errors being thrown in some of the builds from earlier today. This change should make those occurrences produce friendlier error messages. 

Before I added the these waits, the reporter reported a different number of errors most of the times I ran it locally. After the adjustments, I'm getting reports of 19 errors each run.

bump version of chromedriver
bump version of axe-core
wait for gastby root to receive child nodes
allow adequate time for CSS to paint DOM nodes
add missing catch handlers
refactor audit, cleanup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ deploy:
   local_dir: public
 
 before_script:
-  - wget https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip
+  - wget https://chromedriver.storage.googleapis.com/2.45/chromedriver_linux64.zip
   - unzip chromedriver_linux64.zip -d /home/travis/bin/
   - export CHROME_BIN=chromium-browser
   - npm install -g gulp-cli
--- a/package-lock.json
+++ b/package-lock.json
@@ -4950,9 +4950,9 @@
       "dev": true
     },
     "axe-core": {
-      "version": "3.0.3",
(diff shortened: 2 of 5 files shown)


seanforyou23:
Updated the description to reflect latest. Looked at it with @michaelkro today and seems like you need to do some hand-holding with selenium to wait for each step of the rendering process to complete before performing the audit. I seem to get 19/22 errors pretty consistently now (using a 1 second buffer), but I have seen the occasional run where less errors are reported. Seems like the farther < 1 second we go, the more likely the report will miss problems. At one full second, it seems stable.

I suspect we could be experiencing some strangeness with how axe handles some of our animations, selenium not handling async very well, or I suppose it's also possible for gatsby to be adding to the variance as it's quickly being asked for over 60 page requests in a short amount of time and it's just a dev server. Not uncommon for these things to get overloaded and hiccup.

The timeouts do seem to address the problem though, and again while it's not perfect, seems to be more stable with this refactoring and extra attention to render times than without. The entire audit runs in about 2 min, which isn't bad in the scheme of things. Just wanted to point it out in case someone had a concern.

PR #1350, opened 2019-01-31
fix(dropdown): adds back the toggle text, updates docs

### LINKED ISSUE #1348: dropdown toggle text wrapper missing
We've somehow dropped the toggle text that was added via #971

matthewcarleton:
fixes #1348


matthewcarleton:
@mcoker ya that’s a good point. I’d guess we don’t but maybe we can pull in design here.

christiemolloy:
This looks good. Why is the max-width for a truncated example added to workspace.scss ?

mcoker:
@mcarrano @mceledonia - should the text in a dropdown toggle ever wrap to 2 lines? Or should it always be one line, and if the text overflows the dropdown toggle width, the text truncates with `...`? Or maybe it should do something else instead?

here's how it currently behaves:
screen shot 2019-02-01 at 10 52 06 am

or should it behave like this by default?
screen shot 2019-02-01 at 10 54 32 am

mcarrano:
I guess that I would vote for truncation @mcoker . We should never have a dropdown label that needs two lines. What does a Button do? I see this as like a button. Do you agree @mceledonia ?

mcoker:
@mcarrano agree, the toggle for a dropdown is basically a button (it's an actual `<button>` in most cases). Currently the button component overflow is hidden, but not truncated with ellipses. 

screen shot 2019-02-01 at 12 02 40 pm

matthewcarleton:
> This looks good. Why is the max-width for a truncated example added to workspace.scss ?

hey @christiemolloy this is because we want to show the truncated text, but not force a max-width. This should be set based on context. There is an example of this in the page demo where we apply a max-width to the user dropdown.

matthewcarleton:
@mcoker I guess we'd have this as required then in the docs and add it to all the examples.

mceledonia:
Just got back from PTO but I would say I agree, I think the main utility of the expansion panel is to supplement a truncated input. We may want to look into a more clear definition of when to use a single-line input versus a larger text field area. For example an email address field vs. a description field. I think there's certainly a case for both, and there's a question of if we also want the expansion panel to be available for both as well.

christiemolloy:
@matthewcarleton ah gotcha! looks good then

mcoker:
Looks good! Can you also update the login component examples, login upgrade demo, and input group example?

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -24,6 +24,7 @@
   --pf-c-dropdown__toggle--m-plain--BorderColor: transparent;
   --pf-c-dropdown__toggle--m-plain--Color: var(--pf-global--Color--200);
   --pf-c-dropdown__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
+  --pf-c-dropdown__toggle-text--Width: calc(100% - var(--pf-global--spacer--lg));
```
rhamilto: This addition broke the Dropdown - Icon Only case at http://patternfly-react.surge.sh/patternfly-4/components/dropdown.  :(

cc: @spadgett, @janwright73, @priley86

rhamilto: Before: 
screen shot 2019-03-05 at 2 41 25 pm

After:
screen shot 2019-03-05 at 2 41 05 pm

rhamilto: Digging deeper, this kooks like this is bug in @patternfly/react-core.

PR #1354, opened 2019-01-31
fix(dropdown): fix z-index issue with toggle border appearing on top of buttons in toggle

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1349


mcoker:
@christiemolloy that should be fixed in https://github.com/patternfly/patternfly-next/pull/1353

PR #1359, opened 2019-01-31
fix(about-modal): fixes image stretch with scroll

### LINKED ISSUE #1351: about modal hero image stretches
the hero image should be set as background and fixed so it doesn't stretch if there is a lot of content in the content area.
screen shot 2019-01-28 at 6 45 49 pm

matthewcarleton:
closes #1351 
I've also added in `position:sticky;` on the close button so it's stays in place when there is a lot of content and you scroll.


src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -193,6 +197,9 @@
 // Close
 .pf-c-about-modal-box__close {
   grid-area: close;
+  position: sticky;
```
mcoker: 🔥

PR #1362, opened 2019-02-01
feat(AlertDemo): Toasts

christiemolloy:
FYI, spoke with @kybaker and he said that the design for inline notifications will change - so leave it for the time being.

Also, the toast alerts are meant to be the highest in the page hierarchy (over the nav and header)


src/patternfly/components/AlertGroup/alert-group.scss
```diff
@@ -4,6 +4,7 @@
 .pf-c-alert-group {
   // Alert Group variables
   --pf-c-alert-group__item--MarginTop: var(--pf-global--spacer--sm);
+  --pf-c-alert-group__item-m-toast--Width: var(--pf-c-alert-group--m-toast--MaxWidth);
```
mcoker: Curious why this is necessary? Also the var should be `--pf-c-alert-group__item--m-toast--Width`

mcoker:
Hey Christie, looks good. A few questions:

* I wonder if we should create a page demo hbs include that gives us a default page demo without having to recreate a page demo in this demo. This is one more place we'll need to update if we change something in the Page component that needs to be reflected in the demos. 
* Even the way you have it now though, you don't need `demos/Alert/page-template*.hbs` in this directory. A component, layout, demo, or utility will find a hbs include from any other component, layout, demo, or utility dir. At a minimum, I would just remove those files.
* I think we should ask the react devs how this alertGroup will be added to the DOM. I see you've added it in the page header. My presumption is react will just append it to the DOM, before `</body>`. Depending on where they put it in the page, it might need different properties, like `z-index`. My preference would be for them to just append it to the end of document, and we add a high `z-index` to `.pf-c-alert-group.pf-m-toast`
* I'm not sure why the `width/max-width` changes were needed, or `display: flex; flex-direction: column;`. The demo seems to work as expected without them.

I went ahead and dropped in all of these changes in this branch - https://github.com/mcoker/patternfly-next/tree/alertdemo2

And that  built an preview link here - https://alertdemo2--thirsty-noether-30ec26.netlify.com/demos/Alert/examples/

What do you think?

I'm curious what you and others think about having a single hbs include for a sample page demo, so we can reference it in other demos that just need to include a page demo.

mcoker:
Looks like there is a conflict with the page component demo.

PR #1365, opened 2019-02-04
fix(form): update horizontal margin between action buttons

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1364


srambach:
LGTM 👍

PR #1372, opened 2019-02-05
fix(toolbar): update total items text in checked toolbar demos

### LINKED ISSUE #1358: Update toolbar demos
In toolbar demos that have filters applied and are not using pagination, update the "[n] Items" in `.pf-c-toolbar__total-items` to represent a subset of items, like "12 of 37 Items"

boaz0:
fixes #1358 
//cc @mcoker


mcoker:
@boaz0 thanks for doing this! And this was my bad, I should have been more specific in the original issue. I _believe_ the number of items should reflect the sum of the numbers in the filter badges. So in those examples, since the sum of the badges is 36 (18 + 12 + 6), it should read "36 of 37 Items". @gdoyle1 can you confirm?

gdoyle1:
@mcoker Yes that's correct! Sum of badges (# of items with filters added) / total items

PR #1374, opened 2019-02-05
Expandable

### LINKED ISSUE #169: Expandable section
Need an expandable container that allows content to be shown or hidden. Current design can be found here: http://www.patternfly.org/pattern-library/forms-and-controls/expand-collapse-section/#/design

Visuals to refine design approach. We should have all the CSS to create this. 

This can be seen on a form but might have other use cases.

christiemolloy:
Close #169 

Creates the expandable component.


mcoker:
@christiemolloy the way I'm envisioning this component is really simple - it would just be a container (`.pf-c-expandable`) with a button inside (`.pf-c-expandable__toggle`) and a `div` that is hidden by default, and is shown when you click the toggle button. Currently you call that `.pf-c-expandable__expanded`, but I would probably call it `.pf-c-expandable__content` to represent an element, not a state (because it's still called "expanded" when it isn't expanded, and that's confusing).

 I think the form that you have in the `.pf-c-expandable__section` is just a regular form, and shouldn't be part of this component. You would instead add your `.pf-c-expandable` component to an existing form, probably in a `.pf-c-form__group` element.

Also I think the width should just consume the full width of the parent. Our components should always do that unless there is a specific use case to not do that. And for the examples in this component, I would just include the `.pf-c-expandable` stuff, and not any form stuff around it as a demo. We could then add `.pf-c-expandable` to the [Basic Form demos](https://pf4.patternfly.org/demos/BasicForms/examples/) (or create a new demo) to show how it looks in the context of a form.

What do you think @mcarrano @mceledonia?

mcarrano:
That makes sense to me @mcoker

christiemolloy:
@mcoker yes I definitely agree I guess I was confused at first about the requirements. Made those changes. I was thinking of holding off on adding this to a demo for right now until we work out what exactly we want in them.

src/patternfly/components/Expandable/docs/code.md
```diff
@@ -0,0 +1,21 @@
+## Accessibility
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-expandable__toggle` | Indicated that the expandable content is visible. **Required** |
+| `hidden` | `.pf-c-expandable__content` | Indicates that the expandable content is hidden. **Required**. |
+| `aria-hidden="true"` | `.pf-c-expandable__toggle-icon` | Hides from screen readers. |
+| `aria-controls="[id of element controlled]"` | `.pf-c-expandable__toggle-icon` | Identifies the section controlled by the toggle button. |
```
mcoker: I would also add an `id` line that applies to `.pf-c-expandable__content` since that's needed to work with `aria-controls`

src/patternfly/components/Expandable/docs/code.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+## Usage
+
+| Class | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-expandable` | `<div>` | Initiates the expandable component. **Required** |
+| `.pf-c-expandable__section` | `<div>` | Initiates the expandable component section. **Required** |
```
mcoker: I don't think `__section` needs to be a part of this component... this component should just handle the expandable toggle/content, and that should be able to be plugged into whatever. The content above it should just be whatever, probably form data, and the footer below, if it were buttons, would be the buttons in the form, not part of this component. So you can remove `__footer`, too

src/patternfly/components/Expandable/examples/expandable-expanded-example.hbs
```diff
@@ -0,0 +1,15 @@
+{{#> expandable expandable--IsExpanded="true"}}
+  {{#> expandable-section}}
+    This section is always visible.
+  {{/expandable-section}}
+  {{#> expandable-toggle expandable-toggle--attribute='aria-controls="expanded-content-1" aria-expanded="true"'}}
```
mcoker: nitpicky but can you change the ID for this one to be "2" since it's the 2nd example, and change the ID for the non-expanded example to be "1"?

src/patternfly/components/Expandable/expandable.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-expandable {{#if expandable--IsExpanded}} pf-m-expanded{{/if}}{{#if expandable--modifier}} {{expandable--modifier}}{{/if}}"
```
mcoker:
```suggestion
<div class="pf-c-expandable{{#if expandable--IsExpanded}} pf-m-expanded{{/if}}{{#if expandable--modifier}} {{expandable--modifier}}{{/if}}"
```

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
+@import "../../patternfly-imports";
+
+.pf-c-expandable {
+  // Expandable
+  --pf-c-expandable--Width: #{pf-size-prem(450px)};
```
mcoker: We can remove the width (so it's auto/100% by default) and let the parent element control the width.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+  --pf-c-expandable__content--MarginTop: var(--pf-global--spacer--xl);
+
+  // Expandable Footer
+  --pf-c-expandable__footer--Margin-Top: var(--pf-global--spacer--2xl);
+
+  width: var(--pf-c-expandable--Width);
+  padding: var(--pf-c-expandable--PaddingTop) var(--pf-c-expandable--PaddingRight) var(--pf-c-expandable--PaddingBottom) var(--pf-c-expandable--PaddingLeft);
+  background-color: var(--pf-c-expandable--BackgroundColor);
+  background-clip: padding-box;
+  border: var(--pf-c-expandable--BorderWidth) solid transparent;
```
mcoker: I don't know that we need this. It should just display like regular content. Shouldn't need this for high contrast mode like a dropdown menu.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+  box-shadow: var(--pf-c-expandable--BoxShadow);
+
+  &.pf-m-expanded {
+    .pf-c-expandable__toggle {
+      .pf-c-expandable__toggle-icon {
```
mcoker: can you move this and nest it under `&.pf-m-expanded`? This adds specificity that isn't necessary.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+        transform: rotate(90deg);
+      }
+
+      &::after {
+        content: "Show less";
```
mcoker: Should the text be customizable? If so, it could be a variable. Though I might just make it text in the toggle. You might ask a react dev the best way to include the text in the component so it's customizable by the react component. Currently to change the text, you'll need to write CSS.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+  }
+
+  .pf-c-title {
```
mcoker: I think you can remove this

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+    margin-bottom: var(--pf-c-expandable--c-title--MarginBottom);
+  }
+
+  .pf-c-expandable__toggle {
+    margin-top: var(--pf-c-expandable__toggle--MarginTop);
+    font-weight: var(--pf-c-expandable__toggle--FontWeight);
+    color: var(--pf-c-expandable__toggle--Color);
+    border: none;
+
+    .pf-c-expandable__toggle-icon {
```
mcoker: Can you move this out so it isn't nested in `.pf-c-expandable__toggle`?

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+  }
+
+  .pf-c-title {
+    margin-bottom: var(--pf-c-expandable--c-title--MarginBottom);
+  }
+
+  .pf-c-expandable__toggle {
+    margin-top: var(--pf-c-expandable__toggle--MarginTop);
```
mcoker: You shouldn't need this. The spacing above the expandable component can be handled by whatever component it's included in.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
...
+  --pf-c-expandable--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+  --pf-c-expandable--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-expandable--BoxShadow: var(--pf-global--BoxShadow--md);
+  --pf-c-expandable--PaddingTop: var(--pf-global--spacer--2xl);
+  --pf-c-expandable--PaddingRight: var(--pf-global--spacer--3xl);
+  --pf-c-expandable--PaddingBottom: var(--pf-global--spacer--2xl);
+  --pf-c-expandable--PaddingLeft: var(--pf-global--spacer--2xl);
+
+  // Title
+  --pf-c-expandable--c-title--MarginBottom: var(--pf-global--spacer--lg);
```
mcoker: you can remove this

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,84 @@
+@import "../../patternfly-imports";
+
+.pf-c-expandable {
+  // Expandable
+  --pf-c-expandable--Width: #{pf-size-prem(450px)};
+  --pf-c-expandable--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
```
mcoker: If this should always have a white background (assuming it would always be a light theme), you can use `--pf-global--BackgroundColor--100` as the background color and add `@include pf-t-light;`

mcoker:
Looks a lot better! Left a few more comments.

src/patternfly/components/Expandable/docs/code.md
christiemolloy: Okay makes sense

jgiardino:
This looks great!

I'm curious what everyone thinks about the use of `aria-controls` in this component. The example provided by [WAI-ARIA authoring practices](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure) describes the use of `aria-controls` as optional:
> Optionally, the element with role button has a value specified for aria-controls that refers to the element that contains all the content that is shown or hidden.

Also, when working with the students at GMS, their expectation with `aria-expanded` is that when `"true"` the contents are immediately after. I've been thinking that the primary benefit of `aria-controls` is for those cases where the controlled contents are not immediately after (in JAWS, it provides a way to jump to the controlled contents).

Does anyone object to removing it? Is there ever a case where the expanded contents would not come immediately after the toggle?

If we decide to keep it, should we add this everywhere that has the `aria-expanded` attribute?

christiemolloy:
@jgiardino I agree with that. I think in this case the expanded content would always come after the toggle button so its redundant here.

christiemolloy:
@mcoker updated! thanks for reviewing

src/patternfly/components/Expandable/docs/code.md
```diff
@@ -0,0 +1,17 @@
+## Accessibility
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-expandable__toggle` | Indicated that the expandable content is visible. **Required** |
+| `hidden` | `.pf-c-expandable__content` | Indicates that the expandable content is hidden. **Required**. |
+| `aria-hidden="true"` | `.pf-c-expandable__toggle-icon` | Hides from screen readers. |
```
mcoker:
```suggestion
| `aria-hidden="true"` | `.pf-c-expandable__toggle-icon` | Hides from screen readers. **Required** |
```

mcoker:
@mcarrano currently this component is in a container with a background color, padding on all sides, and box-shadow. I was expecting this component to be used anywhere you have additional content you want to show via the "show more" link, and that it could be placed in a modal or form or whatever, and it would be spaced based on the component it's used in. Meaning it would have no box shadow, background-image, or padding - except for the padding/margin between the "show/hide" link and the expanded content below.

This is how it is now:
screen shot 2019-02-13 at 12 39 07 pm


And this is how I was expecting it to look:
screen shot 2019-02-13 at 12 39 35 pm

Also this component is forced to use the light theme (meaning it will stay light in a dark theme/component), and the toggle link color is `--pf-global--active-color--300`. I would expect it to adapt to either the light or dark theme, and the toggle color would just to be a normal link color.

Can you clarify how all of those points should look?

mcarrano:
I think this makes sense @mcoker . I'll defer to @mceledonia on any color or spacing questions.

src/patternfly/components/Expandable/docs/code.md
```diff
@@ -0,0 +1,17 @@
+## Accessibility
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-expandable__toggle` | Indicated that the expandable content is visible. **Required** |
```
mcoker:
```suggestion
| `aria-expanded="true"` | `.pf-c-expandable__toggle` | Indicates that the expandable content is visible. **Required** |
```

mcoker: also can you add a line for `aria-expanded="false"`? It's required too

src/patternfly/components/Expandable/examples/expandable-expanded-example.hbs
```diff
@@ -0,0 +1,7 @@
+{{#> expandable expandable--IsExpanded="true"}}
+  {{#> expandable-toggle expandable-toggle--attribute='aria-expanded="true"'}}
```
mcoker: aria-expanded should be false. however, could you just put a conditional in `expandable-toggle.hbs` that looks for `expandable--IsExpanded`, and generates the proper `aria-expanded` property so all you have to do is specify `expandable--IsExpanded` and everything else updates appropriately?

src/patternfly/components/Expandable/expandable-toggle.hbs
```diff
@@ -0,0 +1,12 @@
+<button class="pf-c-expandable__toggle{{#if expandable-toggle--modifier}} {{expandable-toggle--modifier}}{{/if}}"
```
mcoker: I wonder if the toggle area for this button should match our minimum target for buttons, dropdowns, etc? If so, I wonder if it would be acceptable to use a `.pf-c-button.pf-m-link` here?

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,37 @@
...
+  --pf-c-expandable__toggle--FontWeight: var(--pf-global--FontWeight--semi-bold);
+  --pf-c-expandable__toggle--Color: var(--pf-global--link--Color);
+
+  // Expandable Toggle Icon
+  --pf-c-expandable__toggle-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-expandable__toggle-icon--Color: var(--pf-global--Color--100);
+
+  // Expandable Expanded
+  --pf-c-expandable__content--MarginTop: var(--pf-global--spacer--xl);
+  --pf-c-expandable__content--Color: var(--pf-global--Color--100);
```
mcoker: can you add `.pf-c-expandable` to the list at the top of `_base.scss`? If you add that, I think you can remove this var & color property.

christiemolloy: Yes!

src/patternfly/components/Expandable/docs/code.md
```diff
@@ -0,0 +1,18 @@
+## Accessibility
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-expandable__toggle` | Indicates that the expandable content is visible. **Required** |
+| `aria-expanded="false"` | `.pf-c-expandable__toggle` | Indicates the the expandable content is hidden. **Required** |
+| `hidden` | `.pf-c-expandable__content` | Indicates that the expandable content is hidden. **Required**. |
```
mcoker: you might indicate that it's required when the menu is hidden, or maybe list this the way you did in the accordion.

> Indicates that the expandable content element is hidden. Use with `aria-expanded="false"` **Required** |

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,38 @@
...
+  --pf-c-expandable__toggle--PaddingLeft: 0;
+  --pf-c-expandable__toggle--FontWeight: var(--pf-global--FontWeight--semi-bold);
+  --pf-c-expandable__toggle--Color: var(--pf-global--link--Color);
+
+  // Expandable Toggle Icon
+  --pf-c-expandable__toggle-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-expandable__toggle-icon--Color: var(--pf-global--Color--100);
+
+  // Expandable Expanded
+  --pf-c-expandable__content--MarginTop: var(--pf-global--spacer--xl);
```
mcoker: Talked to @mceledonia about this and it's too much space. Let's go with 16px instead.

```suggestion
  --pf-c-expandable__content--MarginTop: var(--pf-global--spacer--md);
```

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -0,0 +1,38 @@
...
+    .pf-c-expandable__toggle-icon {
+      transform: rotate(90deg);
+    }
+  }
+
+  .pf-c-expandable__toggle {
+    padding: var(--pf-c-expandable__toggle--PaddingTop) var(--pf-c-expandable__toggle--PaddingRight) var(--pf-c-expandable__toggle--PaddingBottom) var(--pf-c-expandable__toggle--PaddingLeft);
+    font-weight: var(--pf-c-expandable__toggle--FontWeight);
+    color: var(--pf-c-expandable__toggle--Color);
+    border: none;
```
mcoker: should we add....

```
  --pf-c-expandable__toggle--Color: var(--pf-global--link--Color--hover);
```

christiemolloy: ah yeah theres no hover/focus styles specified. If we add this, @mceledonia should there be an active/focus state too?

PR #1384, opened 2019-02-07
Replace buttons with dropdowns, fix sizing, fix focus state

### LINKED ISSUE #1183: Table component: update classes used for Kebab
In reviewing the Table component for pf-react (https://github.com/patternfly/patternfly-react/pull/821), the kebab button in the core example appears to be incorrect. 

Currently, the pf-react Table component has the following:
```
<td data-key="5" class="pf-c-table__action">
  <div class="pf-c-dropdown pf-m-expanded">
    <button aria-label="Actions" id="pf-toggle-id-14" class="pf-c-dropdown__toggle pf-m-plain" aria-expanded="true">
      ...
```

But the example in core is missing the dropdown `<div>` and uses different classes on the `<button>`:
```
<td class="pf-c-table__action">
  <button class="pf-c-button pf-m-plain" id="simple-action1" aria-label="Actions">
    ...
```

Also, currently, neither core nor react components convey the expected alignment of the menu. This is set on the menu `<ul>` element, which is currently excluded from the core example, since the core example isn't interactive anyway. Should we add the following snippet to the code so that the `pf-m-align-right` class is included, or would this be noted somewhere else?:
```
<div>
  <ul aria-labelledby="pf-toggle-id-14" class="pf-c-dropdown__menu pf-m-align-right">
    <li>
      ...
```

### LINKED ISSUE #1286: Dropdown hover text is white in dark theme
> Looking at https://patternfly-react.surge.sh/patternfly-4/demos/pagelayout/examples/page-layout-default-nav, if you go to the toolbar dropdown that says "Kyle Baker", the hover state on the items turns the text white. Is there already a bug report opened for this?

### LINKED ISSUE #1379: Update dropdowns
Dropdowns shouldn't use .pf-c-button, they should use a standard <button> class.
Dropdowns should take on the same padding as buttons.
Buttons that will dropdown in react, should use the dropdown component.

- Currently the App Launcher dropdown is using .pf-c-button - it should use have its own class.
- Use button component padding for button.pf-m-plain.pf-c-*__toggle

### LINKED ISSUE #1381: Dropdown Focus States
Dropdown and Select variations dont have focus styles in Core - we should use the styles that we already have for active/expanded.

christiemolloy:
Fix #1183, fix #1381 , fix #1379, fix #1286

Issue: 
- The kebabs in the table were using the button component not the dropdown component.
- The dropdown toggle has the correct padding but the icon is not inheriting the line-height of its parent therefore is at 28px in height not 36px.
- Focus states for the dropdown/ option menu button were not present.


mcoker:
Looks good. Curious about these nested kebabs in the compound expansion example - should these be dropdowns, too?

screen shot 2019-02-07 at 1 33 44 pm

Also can you use unique ID's for the dropdowns? The a11y checker compares all examples for a component on a single page, so ID's between examples need to be unique.

christiemolloy:
@mcoker thanks for the review! I pushed changes -- added dropdowns for the nested table in compound expansion (for the time being I think its safe to assume those will be dropdowns), I added unique IDs, and I also replaced the kebab buttons in the data list with dropdowns.

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -11,7 +11,15 @@
   --pf-c-app-launcher__menu--ZIndex: var(--pf-global--ZIndex--xs);
```
mcoker: Looks like we need to add a normal `font-weight` to `.pf-c-app-launcher__menu-item`. There is a global style for `a` elements that uses a `semi-bold` font-weight. You can mimic the dropdown component for this change.

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -19,6 +19,8 @@
   --pf-c-dropdown__toggle--hover--BorderBottomColor: var(--pf-global--active-color--100);
```
mcoker: can you add `--pf-c-dropdown__toggle--Color: var(--pf-global--Color--100);` (then set the `color` property) so that when you toggle the dropdown in a dark theme, the toggle text updates?

And on L255, we're redefining `--pf-c-dropdown__menu-item--Color` to `var(--pf-c-dropdown__menu-item--Color)` which doesn't do anything :) Can you change that to `--pf-c-dropdown__menu-item--Color: var(--pf-c-dropdown__menu-item--hover--Color);` and then add `  --pf-c-dropdown__menu-item--hover--Color: var(--pf-global--Color--dark-100);`. That will also fix https://github.com/patternfly/patternfly-next/issues/1286

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -160,6 +171,10 @@
   &.pf-m-plain {
     color: var(--pf-c-dropdown__toggle--m-plain--Color);
 
+    > i {
```
mcoker: We can't reliably count on an `<i>` element here. For example, they're `<svg>` in react. I think you could change this to `> *`...

src/patternfly/components/OptionsMenu/options-menu.scss
```diff
@@ -8,6 +8,7 @@
   --pf-c-options-menu__toggle--PaddingBottom: var(--pf-global--spacer--form-element);
   --pf-c-options-menu__toggle--PaddingLeft: var(--pf-global--spacer--sm);
```
mcoker: can you add a `color` to the toggle, similar to my comment about the dropdown, so that the color toggles properly in dark mode?

src/patternfly/components/OptionsMenu/options-menu.scss
```diff
@@ -132,10 +145,16 @@
       color: var(--pf-c-options-menu__toggle--m-plain--Color);
     }
 
+    > i {
```
mcoker: same comment here as with dropdown.

src/patternfly/components/DataList/examples/data-list-check-action-example.hbs
```diff
@@ -18,6 +21,9 @@
     {{#> data-list-cell}}
       Secondary content
     {{/data-list-cell}}
-    {{#> data-list-action button--attribute='aria-labelledby="check-action-item2 check-action-action2" id="check-action-action2" aria-label="Actions"'}}{{/data-list-action}}
+    {{#> data-list-action button--attribute='aria-labelledby="check-action-item2 check-action-action2" id="check-action-action2" aria-label="Actions"'}}
```
mcoker: did you mean to leave `button--attribute` on here? you might also see if you can add the `{{#> dropdown` to `data-list-action.hbs` except omit the `id` attribute and add it to `{{#> data-list-action` instead.

src/patternfly/components/Dropdown/dropdown.scss
christiemolloy: Great! It works now.

src/patternfly/components/DataList/examples/data-list-check-action-example.hbs
christiemolloy: I moved `{{#> dropdown` to `data-list-action.hbs` , and the id is now on data-list-action. But my concern is that I believe the button inside the dropdown needs an ID. It takes in `dropdown-id` and concats it onto `-button` but I'm having trouble with handlebars passing the `data-list-action--ID` into the dropdown id.

If this is not possible I think I may have to add the dropdown back in otherwise it will not have a unique ID. @mcoker

mcoker: You don't have `button--attribute` on any of the other `data-list-action` partials - why do you need it here? And re: the ID, I meant do `{{#> data-list-action id="dropdown-kebab-right-aligned-2"}}` - that should pass `id` in the context of `data-list-action`, which can be consumed by `{{#> dropdown}}` inside of it.

christiemolloy: I had removed button--attribute not sure why it was showing up. Should be good now

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -9,6 +9,7 @@
   --pf-c-dropdown__toggle--MinWidth: var(--pf-global--target-size--MinWidth);
```
mcoker: can you also add `--pf-c-dropdown__toggle-button: var(--pf-global--color--100);` and then `color: var(--pf-c-dropdown__toggle-button);` to `pf-c-dropdown__toggle-button`? You can verify this fixes the split button dropdown in a dark theme. Currently the toggle button is black in a dark theme and it should turn white.

src/patternfly/components/Dropdown/docs/code.md
```diff
@@ -38,7 +38,7 @@ The dropdown menu can contain either links or buttons, depending on the expected
 | `.pf-m-align-right` | `.pf-c-dropdown__menu` | Modifies to display the menu aligned to the right edge of the toggle |
```
mcoker: in these docs, there is a line for

> aria-expanded=“true” | .pf-c-dropdown__menu | ...

I don’t think we need that? `aria-expanded` goes on the toggle, not the menu.

src/patternfly/components/OptionsMenu/docs/code.md
```diff
@@ -24,5 +24,8 @@
 | `.pf-m-top` | `.pf-c-options-menu` | Modifies to display the menu above the toggle. |
 | `.pf-m-align-right` | `.pf-c-options-menu__menu` | Modifies to display the menu aligned to the right edge of the toggle |
 | `.pf-m-expanded` | `.pf-c-options-menu` |  Modifies for the expanded state. |
+| `.pf-m-hover` | `.pf-c-options-menu__toggle`, `.pf-c-options-menu__toggle .pf-m-plain` | Modifies for the hover state. |
```
mcoker: do you need `.pf-c-options-menu__toggle .pf-m-plain` here? Seems like that's a typo, too (should be ``.pf-c-options-menu__toggle.pf-m-plain`) but it seems like listing `.pf-c-options-menu__toggle` covers it.

mcoker:
Great work! Left a few comments for some small things.

src/patternfly/components/Dropdown/dropdown.scss
christiemolloy: Yes working now!

src/patternfly/components/Dropdown/docs/code.md
christiemolloy: yeah this was also on the app launcher component so removed them both.

src/patternfly/components/OptionsMenu/docs/code.md
christiemolloy: Agreed.

src/patternfly/components/Dropdown/docs/code.md
mcoker: Can you also remove this? applied to the select dropdown which is it's own component now.

> aria-pressed="true" | button.pf-c-dropdown__menu-item | Select only The attribute aria-pressed="true" should be set programmatically to the active item. Value should be false when not selected.
-- | -- | --

PR #1386, opened 2019-02-07
fix(modalbox): remove max-height from body

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1385

### DIFF
--- a/src/patternfly/components/ModalBox/modal-box.scss
+++ b/src/patternfly/components/ModalBox/modal-box.scss
@@ -18,7 +18,6 @@
 
   // Body
   --pf-c-modal-box--body--MinHeight: calc(var(--pf-global--FontSize--md) * var(--pf-global--LineHeight--md)); // Allow for at least 1 line of content in the body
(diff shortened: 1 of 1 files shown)


mcoker:
@mcarrano the 2 widths just set a `max-width`. If you don't specify a width for the modal, the default width is `100%` (it will consume the available space on the screen), and that allows you to use a custom width if you want. You would specify your custom width in core via overwriting the `--pf-c-modal-box--Width` variable. Or in react, you would pass a `width` prop to the modal component and that will override the width to whatever you specify.

src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -89,7 +88,6 @@
 .pf-c-modal-box__body {
   flex: 1 1 auto;
   min-height: var(--pf-c-modal-box--body--MinHeight);
-  max-height: var(--pf-c-modal-box--body--MaxHeight);
   overflow-x: hidden;
```
matthewcarleton: do we want to add `-webkit-overflow-scrolling: touch;` here?

PR #1397, opened 2019-02-07
fix(base.scss): add default color to form elements

mcoker:
I considered just adding `color: var(--pf-global--Color--100;` to the `button,
  input,
  optgroup,
  select,
  textarea ` selector in `_base.scss`, since those elements don't inherit color anyways. This would just make those elements (not all in all browsers, but most) our default color instead of the browser's default. I put together an example to test at https://codepen.io/mcoker/pen/NoXwgd.

The only problem I could see with that approach is if someone loaded the pf4 CSS after their app's CSS, and if they had declared a unique button color in their CSS with the same specificity as our element selector, we would change the color of their buttons.  But we're already going to do that with a lot of things in our base scss, like setting `font-weight` on heading elements, `color` on `a`/`a:hover` elements, etc. 

What do y'all think?

fixes https://github.com/patternfly/patternfly-next/issues/1360

### DIFF
--- a/src/patternfly/_base.scss
+++ b/src/patternfly/_base.scss
@@ -147,6 +147,7 @@ html {
     font-family: inherit;
     font-size: 100%;
     line-height: var(--pf-global--LineHeight--md);
+    color: var(--pf-global--Color--100);
   }
 
   img,


matthewcarleton:
@mcoker I see in your codepen that works for everything but the select. I think that approach makes sense, so then in the modifier for each (`.pf-c-button.pf-m-primary`) we we would just set the colour property?

mcoker:
@matthewcarleton yep, though look in other browsers like Firefox. FF lets you specify color for all of those elements. And yep for any button we either 1) want the user to be able to customize via a variable, or 2) want the color to be something other than our default, we would just set the color property.

matthewcarleton:
@mcoker ok I'm cool with putting it in `_base.scss` @srambach do you have anything wisdom to lay on this?

srambach:
I think it seems good to me. The concern in the description beginning with "The only problem I could see with that approach is if someone loaded the pf4 CSS after their app's CSS..." doesn't seem like a worry because I believe their app's CSS should always be loaded after PF, no?

mcoker:
@srambach I would expect so, I'm just trying to anticipate how it could break a user's layout, and that's one way. There are users who may be using Patternfly's components on another system, like PF3 or something else, so they may load PF4 after that. That said, I wouldn't expect them to import our base scss if that's the case, but you never know :) For example, PF3 sets a button's color to `inherit`. Maybe we should use that instead?

I think it's fine to add to the base scss if y'all agree :)

srambach:
👍

mcoker:
@matthewcarleton @srambach ran this by @dgutride and he agrees it's fine to add to add this to the base reset scss.

PR #1399, opened 2019-02-08
fix(select): fixes spacing within select toggle

### LINKED ISSUE #1310: Custom select spacing
When the custom select has multiple items in the toggle and it wraps the spacing is off.

screen shot 2019-01-24 at 4 00 55 pm

srambach:
Fixes #1310

### DIFF
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -28,11 +28,8 @@
   --pf-c-select__toggle--m-plain--Color: var(--pf-global--Color--200);
   --pf-c-select__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
 
-  // Toggle wrapper doesn't need top padding unless there's a typeahead with adjustment for the input
-  --pf-c-select__toggle-wrapper--PaddingTop: 0;
-
   // When there is a typeahead (input box) adjust the top padding of the wrapper in case it wraps to the second line
-  --pf-c-select__toggle-wrapper--typeahead--PaddingTop: var(--pf-global--spacer--xs);
+  --pf-c-select__toggle-wrapper--m-typeahead--PaddingTop: var(--pf-global--spacer--xs);
 
   // Space for the children of the toggle
   --pf-c-select__toggle-wrapper--not-last-child--MarginRight: var(--pf-global--spacer--xs);
@@ -196,7 +193,8 @@
   &.pf-m-typeahead {
     --pf-c-select__toggle--PaddingTop: 0;
     --pf-c-select__toggle--PaddingBottom: 0;
-    --pf-c-select__toggle-wrapper--PaddingTop: var(--pf-c-select__toggle-wrapper--typeahead--PaddingTop);
+
+    padding-top: var(--pf-c-select__toggle-wrapper--m-typeahead--PaddingTop);
 
     input {
       overflow: hidden;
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -187,6 +196,7 @@
   &.pf-m-typeahead {
     --pf-c-select__toggle--PaddingTop: 0;
     --pf-c-select__toggle--PaddingBottom: 0;
+    --pf-c-select__toggle-wrapper--PaddingTop: var(--pf-c-select__toggle-wrapper--typeahead--PaddingTop);
```
mcoker: can you update the var to be `m-typeahead`?

src/patternfly/components/Select/select.scss
```diff
@@ -28,13 +28,22 @@
...
+
   // Space for the children of the toggle
   --pf-c-select__toggle-wrapper--not-last-child--MarginRight: var(--pf-global--spacer--xs);
 
   // Max-width is needed for ellipsing the text. This leaves enough space for the toggle caret.
   // However, for cases where the clear icon button or a badge are added, the correct width will need to be calculated in script
   --pf-c-select__toggle-wrapper--MaxWidth: calc(100% - var(--pf-global--spacer--lg));
 
+  // Chip group within a toggle-wrapper needs a margin-bottom in case it wraps
+  --pf-c-select__toggle-wrapper--c-chip-group--MarginBottom: 0;
```
mcoker: do you need this? doesn't look like it's doing anything.

src/patternfly/components/Select/select.scss
```diff
@@ -224,6 +234,7 @@
   // Max-width is needed for ellipsing the text. This leaves enough space for the toggle caret.
   // However, for cases where the clear icon button or a badge are added, the correct width will need to be calculated in script
   max-width: var(--pf-c-select__toggle-wrapper--MaxWidth);
+  padding-top: var(--pf-c-select__toggle-wrapper--PaddingTop);
```
mcoker: If I understand this part correctly, I think you can remove this and just add `padding-top` in the `m-typeahead` block. We usually don't add a property/variable that doesn't do anything, just so it can be overwritten later in a modifier.

src/patternfly/components/Select/select.scss
```diff
@@ -237,6 +248,16 @@
     padding-top: var(--pf-c-select__toggle-wrapper--c-chip--c-button--PaddingTop);
     padding-bottom: var(--pf-c-select__toggle-wrapper--c-chip--c-button--PaddingBottom);
   }
+
+  // if there is an input in the toggle-wrapper, then give it a negative top margin to match the typeahead top padding
+  > input {
```
mcoker: would this ever be an `input` that isn't a `.pf-c-form-control`?

src/patternfly/components/Select/select.scss
srambach: The chip group comes with a margin of -4px. It's hard to see but you can see the difference when the typeahead input is wrapped below the chip group and you click on the input to get the focus ring. Without this adjustment, the input is exactly at the bottom of the chip group, with it, there's a little space. @kybaker do you want to say which is correct?

mcoker: Ah, gotcha!

Issue #756, opened 2018-09-24
List/Table Demo


mcarrano:
Create a demo that illustrates usage of data lists and data tables in different contexts. When do we use a list? When do we use a table? Consider following use cases:

- Editable rows
- Expandable rows
- Drag and drop to reorder rows

IxD: 
Visual: 
CSS:

gdoyle1:
@mcarrano To confirm what examples we need for this story based on our last meeting:

1. Compound data list with new toolbar design (One with header and navigation, one without)
2. Expanded data list with new toolbar design (One with header and navigation, one without)
3. Compound data table with new toolbar design (One with header and navigation, one without)
4. Expanded data table with new toolbar design (One with header and navigation, one without)

Does that sound right to you? Am I also addressing the capabilities mentioned in this issue (i.e editable rows, expandable rows, drag and drop to reorder rows)?

mcarrano:
@gdoyle1 Not quite sure what you mean by "Compound data list/table"  There is a data list and a data table. The table can take two types of styling, the "hybrid list" type styling and the compact styling. Any of these can also support expansion. If this is still confusing, I'm glad to go over again via bluejeans.

As for the other capabilities, I had not thought about those. But if you have time to address them, we will need this eventually. I left out because they are not high priorities right now to implement.

Does that help?

gdoyle1:
@mcarrano That helps! I'll send over what I have at that moment to make sure we're on the same page.

gdoyle1:
@mcarrano Here's what I have ([link to Sketch file in GD](https://drive.google.com/open?id=1MeaDaCAzCo_g-lxj3vcQ0YGiII3JyhFO)). I haven't added it to the master file yet and won't do so until we meet on Tuesday. Let me know if you need any other examples that are not there.

Examples that are included are:
1. Simple Data List
2. Actionable (with actions and check boxes) Data List
3. Expanded Data List
4. Compact Data Table
5. Expanded Date Table
6. Simple Data Table
7. Sortable Data Table
8. Compound Expanded Data Table

gdoyle1:
[...]

PR #1401, opened 2019-02-08
Data List Demo

christiemolloy:
Part 1/2 of issue #756 

Four List Demo examples
1. Simple
2. Actionable
3. Expandable
4. Simple without Pagination 

- Update the toolbar when the filter component is added
- How should we handle padding around pagination? Right now I added it to the .pf-m-footer modifier for pagination, but I'm very open to changing that.
- There are no mobile examples for the data list but confirmed with @kybaker that it should style the exact same as a responsive table.
- I updated breakpoints in the data list component to use- `pf-global--breakpoint--mobile` rather than the variable which wasn't working.


gdoyle1:
@christiemolloy This is looking good so far! If you could also check out how to add some more white space padding underneath the bottom pagination, that'd be great! It looks almost like it's falling off the list now. And then the other points you noted above overlap with the other points I noticed, so it sounds good!

PR #1402, opened 2019-02-08
feat(sizing): add viewport units and 1/3 and 2/3 width/height

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1400


mcoker:
@matthewcarleton I just added them, was curious what you thought :) "thirds" seem pretty common. Do you think they're common enough to add to the utility or is this getting towards creating a utility for too many widths?

matthewcarleton:
@mcoker ya I think it's ok. I don't think it's bad to offer more. I guess we'll see if they are useful :)

PR #1403, opened 2019-02-09
Feat(skiptocontent): adds component

### LINKED ISSUE #924: Skip to Content component
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/skip-to-content.html

matthewcarleton:
closes #924 

@mceledonia design was never discussed with this. We could make this look however we want. @jgiardino do you have any opinion on how this should look?


jgiardino:
This looks great! I don't have any suggestions about appearance, although looking at the example for [WebAIM](https://webaim.org/techniques/skipnav/), their skip link is animated and has more contrast between the link background and page background, which help make this link more noticeable. That said, this implementation seems noticeable enough, assuming the user is expecting the hamburger to be the first element to receive focus. 

The method you used is noted in the [WebAIM article for skip links](https://webaim.org/techniques/skipnav/), but they include the following note:

> Alternatively, you can use a named anchor to identify the target for the link, though named anchors are no longer conforming in HTML5.

And `name` is also listed in the list of [obsolete attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Obsolete_attributes) on MDN. Do you know if `name` is needed for this method to be used for older browsers?

matthewcarleton:
@jgiardino ya I wasn't sure about the `name` attribute either and now that you mention it I bet we can just pull it out since we don't need to be too concerned with legacy browsers.

src/patternfly/components/SkipToContent/docs/code.md
```diff
@@ -0,0 +1,12 @@
+## Overview
+
+Skip to content allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
+
+When using `.pf-c-skip-to-content` you must also provide an anchor with the matching id that is applied to the `href` attribute of `.pf-c-skip-to-content`. You will typically place this just before the main content of the page. Ex: `<a id="main-content" name="main-content"></a>` For a demo of this see the <a href="/demos/Page/examples">Page demo</a>.
```
mcoker: should you remove `name="main-content"`?

src/patternfly/components/SkipToContent/skip-to-content.hbs
```diff
@@ -0,0 +1,7 @@
+<a class="pf-c-skip-to-content pf-c-button pf-m-primary{{#if skip-to-content--modifier}} {{skip-to-content--modifier}}{{/if}}"
+  href="#main-content"
```
mcoker: should you remove this so the user can specify their own ID?

src/patternfly/demos/Page/examples/page-component-expandable-nav-example.hbs
```diff
@@ -2,6 +2,9 @@
 {{/background-image}}
 
 {{#> page page--id="page-layout-expandable-nav"}}
+  {{#> skip-to-content skip-to-content--attribute='href="#content"'}}
```
mcoker: the ID you're using is `#main-content`.

PR #1404, opened 2019-02-09
feat(wizard): add wizard component

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/755

* need to figure out a good way to have ellipsis overflow on the mobile steps toggle button

### DIFF
--- a/src/patternfly/components/Wizard/docs/code.md
+++ b/src/patternfly/components/Wizard/docs/code.md
@@ -1,0 +1,43 @@
+## Accessibility
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-wizard__toggle` | Indicates that the steps menu is visible. **Required** |
+| `aria-expanded="false"` | `.pf-c-wizard__toggle` | Indicates that the steps menu is hidden. **Required** |
+| `aria-label="close"` | `.pf-c-wizard__toggle-icon` | Gives the close button an accessible name. **Required** |
+| `aria-hidden="true"` | `.pf-c-wizard__toggle-icon`, `.pf-c-wizard__toggle-divider` | Hides the icon from assistive technologies. **Required** |
+| `aria-label="Steps"` | `.pf-c-wizard__nav` | Gives the steps nav element an accessible name. **Required** |
+| `aria-disabled="true"` | `.pf-c-wizard__nav-link` | Indicatese that the element is disabled. **Required for disabled links with `.pf-m-disabled`** |
+| `aria-current="page"` | `.pf-c-wizard__nav-link` | Indicates the current page link. Can only occur once on page. **Required for the current link** |
+| `tabindex="-1"` | `.pf-c-wizard__nav-link` | Removes a link from keyboard focus. **Required for disabled links with `.pf-m-disabled`** |
+
+## Usage
+
+| Class | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-wizard` | `<div>` | Initiates the wizard component. **Required** |
+| `.pf-c-wizard__header` | `<header>` | Initiates the header. **Required** |
+| `.pf-c-wizard__close` | `.pf-c-button.pf-m-plain` | Initiates the close button. **Required** |
+| `.pf-c-wizard__title` | `.pf-c-title.pf-m-3xl` | Initiates the title. **Required** |
+| `.pf-c-wizard__description` | `<p>` | Initiates the description. |
+| `.pf-c-wizard__toggle` | `<button>` | Initiates the mobile steps menu toggle button. **Required** |
+| `.pf-c-wizard__toggle-list` | `<ol>` | Initiates the toggle list. **Required** |
+| `.pf-c-wizard__toggle-list-item` | `<li>` | Initiates a toggle list item. **Required** |
+| `.pf-c-wizard__toggle-num` | `<span>` | Initiates the step number. **Required** |
+| `.pf-c-wizard__toggle-divider` | `<i>` | Initiates the divider between steps. |
+| `.pf-c-wizard__toggle-icon` | `<i>` | Initiates the toggle icon. **Required** |
+| `.pf-c-wizard__outer-wrap` | `<div>` | Initiates the outer wrapper. **Required** |
+| `.pf-c-wizard__inner-wrap` | `<div>` | Initiates the inner wrapper. **Required** |
+| `.pf-c-wizard__nav` | `<nav>` | Initiates the steps nav. **Required** |
+| `.pf-c-wizard__nav-list` | `<ol>` | Initiates a list of steps. **Required** |
+| `.pf-c-wizard__nav-item` | `<li>` | Initiates a step list item. **Required** |
+| `.pf-c-wizard__nav-link` | `<a>` | Initiates a step link. **Required** |
+| `.pf-c-wizard__main` | `<main>` | Initiates the main container. **Required** |
+| `.pf-c-wizard__body` | `<div>` | Initiates the main body container. |
+| `.pf-c-wizard__footer` | `<footer>` | Initiates the footer. **Required** |
+| `.pf-m-expanded` | `.pf-c-wizard__toggle`, `.pf-c-wizard__nav` | Modifies the mobile steps toggle and steps menu for the expanded state. |
+| `.pf-m-current` | `.pf-c-wizard__nav-link` | Modifies a step link for the current state. **Required** |
+| `.pf-m-disabled` | `.pf-c-wizard__nav-link` | Modifies a step link for the disabled state. |
+| `.pf-m-hover` | `.pf-c-wizard__nav-link` | Modifies a step link for the hovered state. This state is primarily for demonstration purposes and would not normally be used in lieu of the `:hover` pseudo-class. |
+| `.pf-m-focus` | `.pf-c-wizard__nav-link` | Modifies a step link for the focus state. This state is primarily for demonstration purposes and would not normally be used in lieu of the `:focus` pseudo-class.|
+| `.pf-m-align-right` | `.pf-c-wizard__footer` | Modifies the footer to right align it's content. |
--- a/src/patternfly/components/Wizard/examples/wizard-example.hbs
+++ b/src/patternfly/components/Wizard/examples/wizard-example.hbs
@@ -1,0 +1,90 @@
+{{#> backdrop}}
+{{#> bullseye}}
+  {{#> wizard}}
+    {{#> background-image}}{{/background-image}}
+    {{#> wizard-header}}
+      {{#> button button--modifier="pf-m-plain pf-c-wizard__close" button--attribute='aria-label="Close"'}}
+        {{> button-icon button-icon--type="times"}}
+      {{/button}}
+      {{#> title title--modifier="pf-m-3xl pf-c-wizard__title"}}Wizard Title{{/title}}
+      {{#> wizard-description}}
+        here is where the description goes
+      {{/wizard-description}}
+    {{/wizard-header}}
+    {{#> wizard-toggle}}
+      {{#> wizard-toggle-list}}
+          {{#> wizard-toggle-list-item}}
+            {{#> wizard-toggle-num}}2{{/wizard-toggle-num}}
+            Configuration
+            {{> wizard-toggle-divider}}
+          {{/wizard-toggle-list-item}}
+          {{#> wizard-toggle-list-item}}
+            Substep B
+          {{/wizard-toggle-list-item}}
+        {{/wizard-toggle-list}}
+        {{> wizard-toggle-icon}}
+      {{/wizard-toggle}}
+      {{#> wizard-outer-wrap}}
+        {{#> wizard-inner-wrap}}
+          {{#> wizard-nav}}
+            {{#> wizard-nav-list}}
+              {{#> wizard-nav-item}}
+                {{#> wizard-nav-link}}
+                  Information
+                {{/wizard-nav-link}}
+              {{/wizard-nav-item}}
+              {{#> wizard-nav-item}}
+                {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-current"}}
+                  Configuration
+                {{/wizard-nav-link}}
+                {{#> wizard-nav-list}}
+                  {{#> wizard-nav-item}}
+                    {{#> wizard-nav-link}}
+                      Substep A
+                    {{/wizard-nav-link}}
+                  {{/wizard-nav-item}}
+                  {{#> wizard-nav-item}}
+                    {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-current" wizard-nav-link--IsCurrent="true"}}
+                      Substep B
+                    {{/wizard-nav-link}}
+                  {{/wizard-nav-item}}
+                  {{#> wizard-nav-item}}
+                    {{#> wizard-nav-link}}
+                      Substep C
+                    {{/wizard-nav-link}}
+                  {{/wizard-nav-item}}
+                {{/wizard-nav-list}}
+              {{/wizard-nav-item}}
+              {{#> wizard-nav-item}}
+                {{#> wizard-nav-link}}
+                  Additional
+                {{/wizard-nav-link}}
+              {{/wizard-nav-item}}
+              {{#> wizard-nav-item}}
+                {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-disabled" wizard-nav-link--attribute='aria-disabled="true" tabindex="-1"'}}
+                  Review
+                {{/wizard-nav-link}}
+              {{/wizard-nav-item}}
+            {{/wizard-nav-list}}
+          {{/wizard-nav}}
+        {{#> wizard-main}}
+          {{#> wizard-body}}
+            <p>wizard content goes here</p>
+          {{/wizard-body}}
+        {{/wizard-main}}
+      {{/wizard-inner-wrap}}
+      {{#> wizard-footer}}
+        {{#> button button--modifier="pf-m-primary" button--attribute='type="submit"'}}
+          Next
+        {{/button}}
+        {{#> button button--modifier="pf-m-secondary"}}
+          Back
+        {{/button}}
+        {{#> button button--modifier="pf-m-link"}}
+          Cancel
+        {{/button}}
+      {{/wizard-footer}}
+    {{/wizard-outer-wrap}}
+  {{/wizard}}
+{{/bullseye}}
+{{/backdrop}}
--- a/src/patternfly/components/Wizard/examples/wizard-expanded-example.hbs
+++ b/src/patternfly/components/Wizard/examples/wizard-expanded-example.hbs
@@ -1,0 +1,90 @@
+{{#> backdrop}}
+{{#> bullseye}}
+  {{#> wizard wizard--IsExpanded="true"}}
+    {{#> background-image}}{{/background-image}}
+    {{#> wizard-header}}
+      {{#> button button--modifier="pf-m-plain pf-c-wizard__close" button--attribute='aria-label="Close"'}}
+        {{> button-icon button-icon--type="times"}}
+      {{/button}}
+      {{#> title title--modifier="pf-m-3xl pf-c-wizard__title"}}Wizard Title{{/title}}
+      {{#> wizard-description}}
+        here is where the description goes
+      {{/wizard-description}}
+    {{/wizard-header}}
+    {{#> wizard-toggle}}
+      {{#> wizard-toggle-list}}
+        {{#> wizard-toggle-list-item}}
+          {{#> wizard-toggle-num}}2{{/wizard-toggle-num}}
+          Configuration
+          {{> wizard-toggle-divider}}
+        {{/wizard-toggle-list-item}}
+        {{#> wizard-toggle-list-item}}
+          Substep B
+        {{/wizard-toggle-list-item}}
+      {{/wizard-toggle-list}}
+      {{> wizard-toggle-icon}}
+    {{/wizard-toggle}}
+    {{#> wizard-outer-wrap}}
+      {{#> wizard-inner-wrap}}
+        {{#> wizard-nav}}
+          {{#> wizard-nav-list}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link}}
+                Information
+              {{/wizard-nav-link}}
+            {{/wizard-nav-item}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-current"}}
+                Configuration
+              {{/wizard-nav-link}}
+              {{#> wizard-nav-list}}
+                {{#> wizard-nav-item}}
+                  {{#> wizard-nav-link}}
+                    Substep A
+                  {{/wizard-nav-link}}
+                {{/wizard-nav-item}}
+                {{#> wizard-nav-item}}
+                  {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-current" wizard-nav-link--IsCurrent="true"}}
+                    Substep B
+                  {{/wizard-nav-link}}
+                {{/wizard-nav-item}}
+                {{#> wizard-nav-item}}
+                  {{#> wizard-nav-link}}
+                    Substep C
+                  {{/wizard-nav-link}}
+                {{/wizard-nav-item}}
+              {{/wizard-nav-list}}
+            {{/wizard-nav-item}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link}}
+                Additional
+              {{/wizard-nav-link}}
+            {{/wizard-nav-item}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-disabled" wizard-nav-link--attribute='aria-disabled="true" tabindex="-1"'}}
+                Review
+              {{/wizard-nav-link}}
+            {{/wizard-nav-item}}
+          {{/wizard-nav-list}}
+        {{/wizard-nav}}
+        {{#> wizard-main}}
+          {{#> wizard-body}}
+            <p>wizard content goes here</p>
+          {{/wizard-body}}
+        {{/wizard-main}}
+      {{/wizard-inner-wrap}}
+      {{#> wizard-footer wizard-footer--modifier="pf-m-align-right"}}
+        {{#> button button--modifier="pf-m-primary" button--attribute='type="submit"'}}
+          Next
+        {{/button}}
+        {{#> button button--modifier="pf-m-secondary"}}
+          Back
+        {{/button}}
+        {{#> button button--modifier="pf-m-link"}}
+          Cancel
+        {{/button}}
+      {{/wizard-footer}}
+    {{/wizard-outer-wrap}}
+  {{/wizard}}
+{{/bullseye}}
+{{/backdrop}}
--- a/src/patternfly/components/Wizard/wizard-toggle-icon.hbs
+++ b/src/patternfly/components/Wizard/wizard-toggle-icon.hbs
@@ -1,0 +1,1 @@
+<i class="fas fa-caret-down pf-c-wizard__toggle-icon" aria-hidden="true"></i>
--- a/src/patternfly/components/Wizard/wizard-toggle.hbs
+++ b/src/patternfly/components/Wizard/wizard-toggle.hbs
@@ -1,0 +1,7 @@
+<button class="pf-c-wizard__toggle{{#if wizard--IsExpanded}} pf-m-expanded{{/if}}{{#if wizard-toggle--modifier}} {{wizard-toggle--modifier}}{{/if}}"
+  aria-expanded="{{#if wizard--IsExpanded}}true{{else}}false{{/if}}"
+  {{#if wizard-toggle--attribute}}
+    {{{wizard-toggle--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</button>
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -1,0 +1,451 @@
+.pf-c-wizard {
+  --pf-c-wizard--Height: 100%;
+  --pf-c-wizard--Width: 100vw;
+  --pf-c-wizard--lg--Height: #{pf-size-prem(762px)}; // Shares height with about modal
+  --pf-c-wizard--lg--MaxHeight: calc(100% - var(--pf-global--spacer--xl));
+  --pf-c-wizard--lg--Width: calc(100% - (var(--pf-global--spacer--3xl) * 2));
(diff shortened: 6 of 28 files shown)


mcarrano:
This is looking good @mcoker . Just a few questions for you and @kybaker .

* When I hover over a step item, I see that the text label turns blue, but the step number is static. Should that have a hover state, too? I think it should.
* I notice that the current step is implemented as a link. But we are already on that page. In general, I would not expect the selected item in a Nav menu to have a hover.
* Should the step number in the disabled state have a circular background like the others? Looks strange without it. I'm thinking just a gray circle with no fill. What do you guys think?

kybaker:
> * When I hover over a step item, I see that the text label turns blue, but the step number is static. Should that have a hover state, too? I think it should.
> I don't think so. I think just changing text color it is a great way to differentiate hover from selected without having to deal with color contrast.
> * I notice that the current step is implemented as a link. But we are already on that page. In general, I would not expect the selected item in a Nav menu to have a hover.
> Agreed.
> * Should the step number in the disabled state have a circular background like the others? Looks strange without it. I'm thinking just a gray circle with no fill. What do you guys think?
> We should have the filled circle just use the next step lighter gray than the unselected filled circle.

mcoker:
> We should have the filled circle just use the next step lighter gray than the unselected filled circle.

@mcarrano @kybaker 

The non-current bg color is `$pf-color-black-200`, so using `$pf-color-black-100`, though it's barely visible. We have `$pf-color-black-150` (`#f5f5f5`) but we haven't introduced that var into the system as a standard global var yet.

If you'd like me to use black-150, would this be a one-off use, or should that color be added as an additional global background-color var to use system-wide?

mcarrano:
Wow. That is pretty perceptible. @kybaker thoughts?

src/patternfly/components/Wizard/examples/wizard-example.hbs
```diff
@@ -0,0 +1,82 @@
...
+              {{/wizard-nav-link}}
+            {{/wizard-nav-item}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-current"}}
+                Sub-step B
+              {{/wizard-nav-link}}
+            {{/wizard-nav-item}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link}}
+                Subs-tep C
```
srambach: On a roll today... typo here.

mcoker: i'm grate at teh spel-ling.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,398 @@
...
+  // Toggle
+  --pf-c-wizard__toggle--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-wizard__toggle--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-wizard__toggle--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-wizard__toggle--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-wizard__toggle--BoxShadow: var(--pf-global--BoxShadow--md-bottom);
+  --pf-c-wizard__toggle--ZIndex: var(--pf-global--ZIndex--sm);
+
+  // Toggle Label
+  --pf-c-wizard__toggle-label--MarginRight: var(--pf-global--spacer--sm);
```
christiemolloy: should this instead have a (max-width of 100% - width of icon), so that the text truncates rather than cutting off? Not sure what design calls for.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,398 @@
...
+
+  // Header
+  --pf-c-wizard__header--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-wizard__header--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-wizard__header--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-wizard__header--PaddingLeft: var(--pf-global--spacer--xl);
+
+  // Close
+  --pf-c-wizard__close--Top: calc(var(--pf-global--spacer--lg) - var(--pf-global--spacer--form-element));
+  --pf-c-wizard__close--Right: calc(var(--pf-global--spacer--xl) - var(--pf-global--spacer--md));
```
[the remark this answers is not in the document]
mcoker: Yep, that's a good idea. I originally wrote it this way to represent that the right value is the right padding of the container minus the right padding of the button. But unless we wrote it like `--pf-c-wizard__close--Right: calc(var(--pf-c-wizard__header--PaddingRight) - var(--pf-c-button--PaddingRight));`, I don't think it makes sense to have this `calc()`

src/patternfly/components/Wizard/docs/code.md
```diff
@@ -0,0 +1,43 @@
...
+| `aria-disabled="true"` | `.pf-c-wizard__nav-link` | Indicatese that the element is disabled. **Required for disabled links with `.pf-m-disabled`** |
+| `tabindex="-1"` | `.pf-c-wizard__nav-link` | Removes a link from keyboard focus. **Required for disabled links with `.pf-m-disabled`** |
+
+## Usage
+
+| Class | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-wizard` | `<div>` | Initiates the wizard component. |
+| `.pf-c-wizard__header` | `<header>` | Initiates the header . |
+| `.pf-c-wizard__close` | `.pf-c-button.pf-m-plain` | Initiates the close button. |
```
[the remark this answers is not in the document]
mcoker: We can, but per the [BEM guidelines](https://cssguidelin.es/#more-layers)

> If we were to add another Element—called, let’s say, .person__eye {}—to this .person {} component, we would not need to step through every layer of the DOM. That is to say, the correct notation would be .person__eye {}, and not .person__head__eye {}. Your classes do not reflect the full paper-trail of the DOM.

If there were more than 1 description, close, and title, it would make sense to use `header-[element]`, but in this case there aren't. Even then, I think we could add `.pf-c-wizard__footer-description` and leave `.pf-c-wizard__description` and it wouldn't be confusing. Some of our classes get unnecessarily long :) But I'm fine with it either way. What do you think, still want to change it?

src/patternfly/components/Wizard/docs/code.md
```diff
@@ -0,0 +1,43 @@
...
+| `.pf-c-wizard__title` | `.pf-c-title.pf-m-3xl` | Initiates the title. |
+| `.pf-c-wizard__description` | `<p>` | Initiates the description. |
+| `.pf-c-wizard__steps` | `<div>` | Initiates the steps container. |
+| `.pf-c-wizard__toggle` | `<button>` | Initiates the mobile steps menu toggle button. |
+| `.pf-c-wizard__toggle-label` | `<div>` | Initiates the label area in the toggle. |
```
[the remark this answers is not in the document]
mcoker: good call!

src/patternfly/components/Wizard/docs/code.md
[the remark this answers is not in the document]
christiemolloy: Oh I never knew this I thought it was a hard rule. I'm definitely fine with leaving this!

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,398 @@
...
+  padding: var(--pf-c-wizard__toggle--PaddingTop) var(--pf-c-wizard__toggle--PaddingRight) var(--pf-c-wizard__toggle--PaddingBottom) var(--pf-c-wizard__toggle--PaddingLeft);
+  border: 0;
+  box-shadow: var(--pf-c-wizard__toggle--BoxShadow);
+
+  &.pf-m-expanded {
+    .pf-c-wizard__toggle-icon {
+      transform: rotate(180deg);
+    }
+
+    + .pf-c-wizard__nav {
```
jgiardino: It might be worth checking with a react dev on this, but using the `+` in this selector makes it dependent on HTML structure, and it would be ideal to have styles applied that are not dependent on HTML structure. 

For example, something I'd like to test for components like this menu is if we can implement them in a way that is similar to how we handle popovers (wrt handling and trapping focus). Currently, the Popover displays in a separate section of the DOM from the toggle

src/patternfly/components/Wizard/examples/wizard-example.hbs
```diff
@@ -0,0 +1,82 @@
...
+      {{/wizard-nav-list}}
+    {{/wizard-nav}}
+  {{/wizard-steps}}
+  {{#> wizard-main}}
+    <p>wizard step content</p>
+  {{/wizard-main}}
+  {{#> wizard-footer}}
+    {{#> form}}
+      {{#> form-actions}}
+        {{#> button button--modifier="pf-m-primary"}}
```
jgiardino: I'm not sure if this is easy to do, or something we want to show here, but when I use this component in my product, I _think_ I would want this button to be `type="submit"` so that I can hit Enter in the context of the form to move on to the next page. Is that something we would want to document here, or is that something that we handle when building the React component?

src/patternfly/components/Wizard/wizard.scss
mcoker: Yep, good call! Updated the toggle/menu expanded styles to respond to a class on their parent element instead that can be toggled in react.

src/patternfly/components/Wizard/examples/wizard-example.hbs
mcoker: I agree. Do you think `type="submit"` should be optional, or can we just ship it that way? Seems to me like the primary action in the wizard (including if it changes from "next" to "review" or "finish") should always have `type="submit"`, so I would be fine just adding it. We do that with the Login component's form button currently. It's `type="submit"` in core and react, but not documented in either. Do you think this (and the login form button?) should be documented as needing `type="submit"`?

jgiardino: I can't think of a reason why you wouldn't want `type="submit"` (or if there's a reason to use `input type="submit"` instead) so I think it would be fine to just ship it that way.

I don't think it hurts to document that, but at the same time, I don't know that we need to document it—it seems like basic, standard HTML stuff and I'm not sure where that would live in our current document structure.

mcoker: Cool! Added that in my last commit.

mcoker:
Thanks @christiemolloy @jgiardino, made those updates.

src/patternfly/components/Wizard/examples/wizard-example.hbs
```diff
@@ -0,0 +1,86 @@
...
+            Configuration
+          {{/wizard-nav-link}}
+          {{#> wizard-nav-list}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link}}
+                 Substep A
+              {{/wizard-nav-link}}
+            {{/wizard-nav-item}}
+            {{#> wizard-nav-item}}
+              {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-current"}}
```
jgiardino: Similar to our Nav component, we should include `aria-current="page` for the current selected page.

src/patternfly/components/Wizard/wizard-toggle.hbs
```diff
@@ -0,0 +1,9 @@
+<button class="pf-c-wizard__toggle{{#if wizard-toggle--modifier}} {{wizard-toggle--modifier}}{{/if}}"
+  aria-expanded="{{#if wizard--IsExpanded}}true{{else}}false{{/if}}"
+  aria-controls="{{id}}-nav"
+  aria-haspopup="true"
```
jgiardino: I'm finding that this attribute in JAWS is really weird when you use it outside the context of a form. For example, if I'm navigating a form in JAWS, the components we have with `aria-haspopup` work great. But if you encounter this component outside of a form, it's behavior is a little janky. 

I'm hoping that what we have labeled as "Basic Dropdown" (without aria-haspopup, and implemented with similar keyboard interaction of a popover) will work for these use cases of Dropdowns and Selects that are not embedded in a form, but I haven't been able to test a POC of that to know for sure. For now, I'd suggest just keeping `aria-expanded` and removing `aria-haspopup`.

jgiardino: Aaaaand I've been commenting on all the PRs that I think we can remove `aria-controls` from cases like this. I think we only need it for things like Tabs or buttons in a toolbar, where the controlled element is not immediately after the toggle (or focus isn't automatically shifted like in a modal or popover). Currently this attribute only seems to benefit JAWS users, where JAWS provides a way to navigate to the controlled element.

mcoker:
thanks @jgiardino! made those updates.

christiemolloy:
Hey Coker, still seeing this run off the page

screen shot 2019-02-13 at 9 23 56 am

mcoker:
@christiemolloy updated

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,419 @@
...
+
+  // Toggle Divider
+  --pf-c-wizard__toggle-divider--MarginLeft: var(--pf-global--spacer--sm);
+
+  // Toggle Icon
+  --pf-c-wizard__toggle-icon--MarginTop: #{pf-size-prem(3px)};
+
+  // Nav
+  --pf-c-wizard__nav--Top: 100%;
+  --pf-c-wizard__nav--MaxHeight: #{pf-size-prem(200px)};
```
christiemolloy: Looks like the max height is larger in the invision file and has more padding

screen shot 2019-02-18 at 11 02 28 am

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,419 @@
...
+  --pf-c-wizard__nav--ZIndex: var(--pf-global--ZIndex--xs);
+
+  @media screen and (min-width: $pf-global--breakpoint--md) {
+    --pf-c-wizard__nav--Top: var(--pf-c-wizard__nav--md--Top);
+    --pf-c-wizard__nav--MaxHeight: var(--pf-c-wizard__nav--md--MaxHeight);
+    --pf-c-wizard__nav--BoxShadow: var(--pf-c-wizard__nav--md--BoxShadow);
+  }
+
+  // Nav List (nested)
+  --pf-c-wizard__nav--nested--MarginLeft: calc(var(--pf-c-wizard__nav-link--PaddingLeft) + var(--pf-global--spacer--md));
```
christiemolloy: Where is this variable - `--pf-c-wizard__nav-link--PaddingLeft`. Might be wrong but I can't see it

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,419 @@
+.pf-c-wizard {
+  --pf-c-wizard--GridTemplateRows: auto auto 1fr auto;
+  --pf-c-wizard--md--GridTemplateRows: auto 1fr auto;
+  --pf-c-wizard--md--GridTemplateColumns: 300px 1fr;
+  --pf-c-wizard--Width: calc(100% - var(--pf-global--spacer--xl) * 2);
```
christiemolloy: Does it need a width and height?

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,419 @@
...
+
+  // Title
+  --pf-c-wizard__title--PaddingRight: var(--pf-global--spacer--xl);
+
+  // Description
+  --pf-c-wizard__description--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-wizard__description--Color: var(--pf-global--Color--400);
+
+  // Steps
+  --pf-c-wizard__steps--md--BoxShadow: var(--pf-global--BoxShadow--md-right);
```
christiemolloy: does the `md` refer to the box shadow size?

src/patternfly/components/Wizard/examples/wizard-expanded-example.hbs
```diff
@@ -0,0 +1,86 @@
...
+            Review
+          {{/wizard-nav-link}}
+        {{/wizard-nav-item}}
+      {{/wizard-nav-list}}
+    {{/wizard-nav}}
+  {{/wizard-steps}}
+  {{#> wizard-main}}
+    <p>wizard step content</p>
+  {{/wizard-main}}
+  {{#> wizard-footer wizard-footer--modifier="pf-m-align-right"}}
```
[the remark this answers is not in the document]
mcoker: To show the modifier class in use. Do you think it shouldn't?

src/patternfly/components/Wizard/wizard.scss
mcoker: it's because it's a var in the `md` media query: `@media screen and (min-width: $pf-global--breakpoint--md)`

src/patternfly/components/Wizard/wizard.scss
mcoker: It's similar to a modal, so I think it would. That said, I'll verify with Kyle that it will always present as a modal, and assuming it fits in a `.pf-l-bullseye` I'll make sure the height/width works the same as a modal

src/patternfly/components/Wizard/wizard.scss
mcoker: nice, the build was failing because of the missing var :) that's a leftover from a previous change. addressed that and another var.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,419 @@
...
+    &::before {
+      --pf-c-wizard__nav-link--before--BackgroundColor: var(--pf-c-wizard__nav-link--m-disabled--before--BackgroundColor);
+      --pf-c-wizard__nav-link--before--Color: var(--pf-c-wizard__nav-link--m-disabled--before--Color);
+    }
+  }
+}
+
+.pf-c-wizard__main {
+  flex-grow: 1;
+  padding-top: var(--pf-c-wizard__main--PaddingTop);
```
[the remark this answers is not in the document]
mcoker: I'll setup a meeting with @kybaker and go over all of the spacing and the dropdown menu height (that's matched from the context selector)

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,419 @@
...
+  border: 0;
+  box-shadow: var(--pf-c-wizard__toggle--BoxShadow);
+
+  @media screen and (min-width: $pf-global--breakpoint--md) {
+    display: none;
+    visibility: hidden;
+  }
+}
+
+.pf-c-wizard__toggle-list {
```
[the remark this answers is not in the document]
mcoker: fixed

src/patternfly/components/Wizard/wizard.scss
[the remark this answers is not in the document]
mcoker: Updated the spacing. The header left/right padding, main padding, and nav padding should match the page component.

src/patternfly/components/Wizard/wizard-toggle-icon.hbs
```diff
@@ -0,0 +1 @@
+<i class="fas fa-chevron-down pf-c-wizard__toggle-icon" aria-hidden="true"></i>
```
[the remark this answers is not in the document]
mcoker: fixed

src/patternfly/components/Wizard/wizard.scss
mcoker: Talked to @kybaker and just allowing this to consume the available area when the nav is open.

src/patternfly/components/Wizard/wizard.scss
mcoker: @kybaker said to match about modal box sizing. updated to match.

src/patternfly/components/Wizard/wizard.scss
christiemolloy: oh nice!

src/patternfly/components/Wizard/examples/wizard-expanded-example.hbs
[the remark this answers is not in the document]
christiemolloy: I would confirm with design that it the position of these buttons can vary

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,457 @@
+.pf-c-wizard {
+  --pf-c-wizard--GridTemplateRows: auto auto 1fr auto;
+  --pf-c-wizard--lg--GridTemplateRows: auto 1fr auto;
+  --pf-c-wizard--GridTemplateColumns: 100%;
+  --pf-c-wizard--lg--GridTemplateColumns: 300px 1fr;
+  --pf-c-wizard--Height: 100%;
+  --pf-c-wizard--Width: 100vw;
+  --pf-c-wizard--lg--Height: #{pf-size-prem(762px)}; // Shares height with about modal
```
christiemolloy: The way that the wizard truncates when the viewport height reduces is different to the about modal. Not sure if the content should remain and the buttons are hidden or vice versa. But I think we should decide here and then create an issue to update the about modal.

screen shot 2019-02-19 at 10 19 16 am

screen shot 2019-02-19 at 10 20 55 am

src/patternfly/components/Wizard/examples/wizard-example.hbs
```diff
@@ -0,0 +1,92 @@
...
+            {{#> wizard-nav-link wizard-nav-link--modifier="pf-m-disabled" wizard-nav-link--attribute='aria-disabled="true" tabindex="-1"'}}
+              Review
+            {{/wizard-nav-link}}
+          {{/wizard-nav-item}}
+        {{/wizard-nav-list}}
+      {{/wizard-nav}}
+    {{/wizard-steps}}
+    {{#> wizard-main}}
+      {{#> wizard-body}}
+        wizard content goes were
```
christiemolloy: here*

src/patternfly/components/Wizard/examples/wizard-example.hbs
```diff
@@ -0,0 +1,92 @@
...
+      {{/wizard-body}}
+    {{/wizard-main}}
+    {{#> wizard-footer}}
+      {{#> form}}
```
christiemolloy: Should we remove the form from the footer and add custom spacing

src/patternfly/components/Wizard/docs/code.md
```diff
@@ -0,0 +1,42 @@
...
+| `aria-disabled="true"` | `.pf-c-wizard__nav-link` | Indicatese that the element is disabled. **Required for disabled links with `.pf-m-disabled`** |
+| `aria-current="page"` | `.pf-c-wizard__nav-link` | Indicates the current page link. Can only occur once on page. **Required for the current link** |
+| `tabindex="-1"` | `.pf-c-wizard__nav-link` | Removes a link from keyboard focus. **Required for disabled links with `.pf-m-disabled`** |
+
+## Usage
+
+| Class | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-wizard` | `<div>` | Initiates the wizard component. **Required** |
+| `.pf-c-wizard__header` | `<header>` | Initiates the header . **Required** |
```
christiemolloy: change to - header.

src/patternfly/components/Wizard/docs/code.md
```diff
@@ -0,0 +1,42 @@
...
+| `.pf-c-wizard__toggle-list` | `<ol>` | Initiates the toggle list. **Required** |
+| `.pf-c-wizard__toggle-list-item` | `<li>` | Initiates a toggle list item. **Required** |
+| `.pf-c-wizard__toggle-num` | `<span>` | Initiates the step number. **Required** |
+| `.pf-c-wizard__toggle-divider` | `<i>` | Initiates the divider between steps. |
+| `.pf-c-wizard__toggle-icon` | `<i>` | Initiates the toggle icon. **Required** |
+| `.pf-c-wizard__nav` | `<nav>` | Initiates the steps nav. **Required** |
+| `.pf-c-wizard__nav-list` | `<ol>` | Initiates a list of steps. **Required** |
+| `.pf-c-wizard__nav-item` | `<li>` | Initiates a step list item. **Required** |
+| `.pf-c-wizard__nav-link` | `<a>` | Initiates a step link. **Required** |
+| `.pf-c-wizard__main` | `<main>` | Initiates the main container. **Required** |
```
christiemolloy: Could `.pf-c-wizard__main` take on the padding of `.pf-c-wizard__body`. I don't see styles for `.pf-c-wizard__main` so I don't see the need for both of these classes, but let me know if I'm wrong :)

src/patternfly/components/Wizard/wizard.scss
mcoker: fwiw, your screenshot is of the normal modal, not the [about modal](https://pf4.patternfly.org/components/AboutModalBox/examples/).

@kybaker what do you think? This is on the edge case that a person has a super short viewport. 

Here are some screenshots:

## normal height

screen shot 2019-02-19 at 9 52 46 am
screen shot 2019-02-19 at 9 53 07 am

## super short height

screen shot 2019-02-19 at 9 52 55 am
screen shot 2019-02-19 at 9 53 12 am

src/patternfly/components/Wizard/docs/code.md
mcoker: I talked to @kybaker and we're treating this similar to a card. With the body element, you get the proper spacing that allows you just to put content in it and it works. Without that element, the main area is just a container where you can put anything you want, and we don't impose padding or anything.

christiemolloy: Ok that sounds good to me

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,470 @@
...
+  counter-reset: wizard-nav-count;
+
+  // Nested sub menus
+  & & {
+    margin-top: var(--pf-c-wizard__nav--nested--MarginTop);
+    margin-left: var(--pf-c-wizard__nav--nested--MarginLeft);
+
+    // Reset styles for nested nav links
+    .pf-c-wizard__nav-link {
+      padding-left: 0;
```
matthewcarleton: do you need to reset padding here? I don't see it being set anywhere else.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,470 @@
...
+  grid-column: 1 / -1;
+  grid-row: 4 / 5;
+
+  @media screen and (min-width: $pf-global--breakpoint--lg) {
+    grid-column: 2 / 3;
+    grid-row: 3 / 4;
+  }
+
+  > * {
+    margin-bottom: var(--pf-c-wizard__footer--child--MarginBottom);
```
matthewcarleton: Is this needed? Is it for wrapping buttons?

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,470 @@
...
+  }
+
+  > *:not(.pf-c-wizard__header) {
+    background-color: var(--pf-c-wizard--children--BackgroundColor);
+  }
+
+  .pf-c-background-image {
+    &::before {
+      position: absolute;
+      top: 0;
```
matthewcarleton: Shouldn't you only need to redefine `position` here?

src/patternfly/components/Wizard/examples/wizard-expanded-example.hbs
```diff
@@ -0,0 +1,88 @@
+{{#> backdrop}}
+{{#> bullseye}}
+  {{#> wizard wizard--IsExpanded="true"}}
+    {{#> background-image}}{{/background-image}}
+    {{#> wizard-header}}
+      {{#> button button--modifier="pf-m-plain pf-c-wizard__close" button--attribute='aria-list="Close"'}}
```
matthewcarleton: aria-label="close"?

src/patternfly/components/Wizard/examples/wizard-example.hbs
```diff
@@ -0,0 +1,88 @@
+{{#> backdrop}}
+{{#> bullseye}}
+  {{#> wizard}}
+    {{#> background-image}}{{/background-image}}
+    {{#> wizard-header}}
+      {{#> button button--modifier="pf-m-plain pf-c-wizard__close" button--attribute='aria-list="Close"'}}
```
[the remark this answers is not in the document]
mcoker: doh!

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -0,0 +1,470 @@
...
+    --pf-c-wizard__toggle--PaddingRight: var(--pf-c-wizard__toggle--lg--PaddingRight);
+    --pf-c-wizard__toggle--PaddingLeft: var(--pf-c-wizard__toggle--lg--PaddingLeft);
+  }
+
+  // Toggle List
+  --pf-c-wizard__toggle-list--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-wizard__toggle-list--MarginBottom: calc(var(--pf-c-wizard__toggle-list-item--MarginBottom) * -1);
+
+  // Toggle List Item
+  --pf-c-wizard__toggle-list-item--not-first-child--FontSize: var(--pf-global--FontSize--sm);
```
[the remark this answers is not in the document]
mcoker: oh good catch!

src/patternfly/components/Wizard/wizard.scss
mcoker: nope! that's leftover from a previous iteration.

src/patternfly/components/Wizard/wizard.scss
mcoker: Yeah, and the footer's `padding-bottom` is adjusted to account for the button's `margin-bottom

mcoker:
thanks @matthewcarleton! updated.

matthewcarleton:
nice! Did you see my comment above too? 
> I've noticed in the design the expanded steps don't take full height in mobile but they do in yours, was this changed?

mcoker:
@matthewcarleton refactored the layout to support the mobile expanded nav state. mind looking at it again?

matthewcarleton:
@mcoker ya I think that works. Although did you notice that you can scroll the content behind outside of that area? So if I drag overtop the expanded menu it still scrolls the content. This is only an issue when the expanded area isn't taking full height.

mcoker:
@matthewcarleton today I learned that [webkit-overflow-scrolling: touch; creates a new stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context#The_stacking_context), but it seems to ignore `z-index`. On top of that, I moved the scrolling behavior of the `nav` to the `nav-list`, then gave `nav` (which is adjacent to `main`) a higher `z-index` than `main`, so that `main`'s stacking context shouldn't compete with `nav-list` and it still didn't work! As best I can tell, if an element _that doesn't scroll, either at all, or hasn't overflowed any content__, is displayed on top of another element with `webkit-overflow-scrolling: touch;`, the underlying element will scroll when you drag the element on top of it. 

After spending way too much time trying workarounds I saw posted online and having no luck, I just removed `webkit-overflow-scrolling: touch;` from `.pf-c-wizard__main` and will open a follow up issue to investigate this later so that this bug isn't blocking the component.

PR #1417, opened 2019-02-13
Data Table Demo

christiemolloy:
Part 1/2 of issue #756 

Five Table Demo examples

- 1. Simple
- 2. Sortable
- 3. Expandable
- 4. Compact
- 5. Compound Expansion

As of, 02/27 waiting on these PRs to merge:
- #1472
- #1482 (breaking change)
- #1495  (breaking change)
- #1497 
- #1498 
- #1499


src/patternfly/components/Page/page.scss
```diff
@@ -82,24 +82,13 @@
...
   // Main section
-  --pf-c-page__main-section--PaddingTop: var(--pf-global--spacer--md);
-  --pf-c-page__main-section--PaddingRight: var(--pf-global--spacer--md);
-  --pf-c-page__main-section--PaddingBottom: var(--pf-global--spacer--md);
-  --pf-c-page__main-section--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-page__main-section--md--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-page__main-section--md--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-page__main-section--md--PaddingBottom: var(--pf-global--spacer--xl);
-  --pf-c-page__main-section--md--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__main-section--PaddingTop: var(--pf-global--spacer--xl);
```
mcoker: Just want to verify that this is purposeful. The spacing in other elements in the page component change at a medium breakpoint, so the content no longer aligns properly. 

This is how it was before:

screen shot 2019-02-19 at 4 11 43 pm

And here's what happens now, including the code below that removes the padding on the last section area on a mobile breakpoint

screen shot 2019-02-19 at 4 11 49 pm

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -2,6 +2,10 @@
   // Base
   --pf-c-pagination--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-pagination--child--MarginRight: var(--pf-global--spacer--lg);
+  --pf-c-pagination--m-footer--PaddingTop: var(--pf-global--spacer--md);
```
mcoker: Should the footer pagination be in a toolbar, so the spacing can just come from the toolbar in both instances and we don't have to match it here?

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
```diff
@@ -8,9 +8,8 @@
     {{/button}}
   {{/page-header-tools-group}}
   {{#> page-header-tools-group}}
-    {{#> button button--modifier="pf-m-plain pf-m-mobile" button--attribute='aria-label="Overflow actions"'}}
-      {{> button-icon button-icon--type="ellipsis-v"}}
-    {{/button}}
+    {{#> dropdown id="dropdown-kebab-right-aligned-1" dropdown--IsActionMenu="true" dropdown-menu--modifier="pf-m-align-right" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
```
mcoker: you need to add the `.pf-m-mobile` class so it only shows as you shrink the page down. Currently it's shown on desktop.

screen shot 2019-02-19 at 4 32 05 pm

src/patternfly/demos/DataTable/examples/data-table-sortable-demo.hbs
```diff
@@ -0,0 +1,29 @@
+{{#> background-image}}
```
mcoker: Good job consolidating the code for these demos into something manageable.

src/patternfly/demos/DataTable/docs/code.md
```diff
@@ -0,0 +1,3 @@
+<!-- ## Overview -->
+
+This demo's the table component.
```
mcoker: I don't know that you need this but it should be `demos`

mcoker:
Just want to verify that this is correct. The expandable table changes at 2 breakpoints. 

desktop

screen shot 2019-02-19 at 4 49 05 pm

then at a breakpoint there is a small grey border under the expanded content

screen shot 2019-02-19 at 4 49 16 pm

then at another, there is a thick grey border above the expanded content

screen shot 2019-02-19 at 4 49 24 pm

mcoker:
Also just want to verify this is how the responsive view of tables should display in general. In particular, should there be a thick border before the first row? And a thin border after the last?

screen shot 2019-02-19 at 4 52 22 pm

mcoker:
the sortable table also has 2 breakpoints

desktop
screen shot 2019-02-19 at 4 54 06 pm

first breakpoint
screen shot 2019-02-19 at 4 54 12 pm

second breakpoint
screen shot 2019-02-19 at 4 54 16 pm

jgiardino:
The value defined in `aria-controls` for the following `<button>` should be used as an `id` for `nav.pf-c-nav`. Since these demos only display in full page mode, it might be possible to use the same id for all of these demos, if that makes it simpler to update.  I'm not sure if that would fail the accessibility audit or not. 

`<button class="pf-c-button pf-m-plain" id="page-layout-data-list-actionable-nav-toggle" aria-label="Global navigation" aria-expanded="true" aria-controls="page-layout-data-list-actionable-primary-nav">`

src/patternfly/components/Page/page.scss
christiemolloy: This is a really good point. We want this styling for the table examples where the `.pf-c-page__main-section` has no padding when the page responds. But I wonder if we want it for every example for the page on mobile - and I think not. I wonder if we should create a modifier to add to `.pf-c-page__main-section` that has no padding on mobile so that it can hold a table, or other examples that dont need padding in the responsive state. What do you think? @mcoker

src/patternfly/components/Pagination/pagination.scss
christiemolloy: Yes good idea!

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
christiemolloy: Yep fixed

src/patternfly/demos/DataTable/docs/code.md
christiemolloy: Going to remove it for consistency

christiemolloy:
@gdoyle1 made updates from what we spoke about yesterday.
@jgiardino thanks for that  - made the update.
@mcoker verified that the table has two breakpoints, and a thinner border at the bottom/thicker at the top. And I fixed those borders you pointed out. Also updated the page main section to have a modifier for when the table uses it - but let me know what you think of that.

src/patternfly/components/Page/page.scss
mcoker: Right, I don't think we want to change the default page layout just to accommodate these examples or use cases. FWIW, it still looks like you've removed the variable main-section padding between small/md viewports, and are using the desktop padding on both breakpoints except the bottom padding of the main section is larger at both breakpoints now, too.

We used to use `md` padding all around on mobile, and `xl` padding on desktop. Now we're using `xl` for top, left, and right, and `2xl` for bottom padding on both mobile and desktop. I just want to verify that this fundamental change to the page layout spacing is part of this PR. If so, this also implies that we need to make global spacing changes to the masthead and navigation in a mobile viewport to match this new spacing.

This is how the masthead, navigation, and content align on the left edge currently.

screen shot 2019-02-21 at 5 06 25 pm

screen shot 2019-02-21 at 5 06 28 pm

This is how the content now aligns with this change. If we keep this main section padding as it is, I'm assuming we'll need to update the masthead left/right padding so those components align on the left edge, and since the masthead padding changed, the navigation padding will need to change, so that when it's open, it now aligns with the new padding in the masthead.

screen shot 2019-02-21 at 5 07 11 pm

src/patternfly/components/Page/page.scss
```diff
@@ -82,24 +82,17 @@
...
+  --pf-c-page__main-section--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-page__main-section--PaddingBottom: var(--pf-global--spacer--2xl);
+  --pf-c-page__main-section--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__main-section--m-table--PaddingTop: calc(var(--pf-global--spacer--sm) / 2);
```
mcoker: any reason you wouldn't do this?

```suggestion
  --pf-c-page__main-section--m-table--PaddingTop: var(--pf-global--spacer--xs);
```

src/patternfly/components/Page/page.scss
mcoker: re: the modifier for an alternate spacing on the main section, I assume it's not an option just to keep the spacing that currently exists? (16px or `--pf-global--spacer--xl` all around).  And we've talked recently about putting together demos that have custom styling or use a utility for spacing as a demonstration of _something a user can do_, instead of _something you get by default from patternfly_, specifically around building card demos. Just want to verify this is something we want to ship with? If so I could see the use for a simple modifier that just removes the padding entirely or maybe removing the default spacing and adding an element that creates the spacing (like we do in `Card` and are doing in `Wizard`) and just leave the area up to the user to handle their own spacing, but that wouldn't work for the examples in this PR because of the small `padding-top` applied. 

Assuming this change should ship with patternfly, it's fine as is, but I doubt this spacing will just be used for a table (or maybe it will?), so I think a more meaningful class would be better. I'm not sure what that name would be, so maybe we just leave it :)

src/patternfly/components/Page/page.scss
christiemolloy: my bad this is meant to be 2px

src/patternfly/components/Page/page.scss
mcoker: hey @christiemolloy just checking to see if you saw my comment above? I'm assuming this would also impact https://github.com/patternfly/patternfly-next/pull/1401

christiemolloy: @mcoker Based on the designs, this padding around the card is something that we would want to ship with Patternfly it shoudn't be custom because it is the default styling for cards that have data tables/data lists, on mobile. IMO we haven't had enough examples on cards to see all the variations for page mobile styling. As of right now, I think we should have the standard padding and then a modifier for this padding but I'm going to check with design!

mcarrano:
@christiemolloy Just one small thing I noticed when looking at this again today. The pagination control in the footer should be right-aligned. Pagination is always the right-most element whether above or below the list/table.

christiemolloy:
@mcarrano yes this is being addressed in #1482

src/patternfly/demos/DataTable/examples/data-table-sortable-demo.hbs
```diff
@@ -0,0 +1,29 @@
...
+{{/background-image}}
+{{#> page page--id="page-layout-data-table-sortable"}}
+  {{#> page-header}}
+    {{> data-table-page-header}}
+  {{/page-header}}
+  {{#> page-sidebar}}
+    {{> data-table-page-nav}}
+  {{/page-sidebar}}
+  {{#> page-main}}
+    {{#> page-main-section page-main-section--modifier="pf-m-light"}}
```
mcoker: do these sections need `.pf-m-no-padding-mobile`?

src/patternfly/demos/DataTable/examples/data-table-compact-demo.hbs
```diff
@@ -0,0 +1,29 @@
...
+{{#> page page--id="page-layout-data-table-compact"}}
+  {{#> page-header}}
+    {{> data-table-page-header}}
+  {{/page-header}}
+  {{#> page-sidebar}}
+    {{> data-table-page-nav}}
+  {{/page-sidebar}}
+  {{#> page-main}}
+    {{#> page-main-section page-main-section--modifier="pf-m-light"}}
+      {{#> grid grid--modifier="pf-m-gutter"}}
```
mcoker: Why use `grid` here? I don't think you need `grid` in these demos at all. Also you should use `.pf-c-page__main-nav` instead of `.pf-c-page__main-section` to hold the tertiary nav. It applies less top padding so the nav sits closer to the header.

mcoker: BTW the datalist demos have the same problem.

christiemolloy: thanks for letting me know about that! I'll create an issue to update data list.

mcoker:
In this demo the row borders don't extend to the right edge of the parent

https://deploy-preview-1417--pf-next.netlify.com/demos/DataTable/examples-full/?component=Data%20Table%20-%20Expandable%20Demo

Screen Shot 2019-03-26 at 1 40 24 PM


In this demo, looks like the table needs to break to the grid layout earlier https://deploy-preview-1417--pf-next.netlify.com/demos/DataTable/examples-full/?component=Data%20Table%20-%20Compact%20Demo

Screen Shot 2019-03-26 at 1 39 47 PM

mcarrano:
@christiemolloy Something to note that I may have missed before. When the toolbar is used together with a table, the bulk selector (first item in the toolbar) is not needed. This is because the table has built in bulk selection via the checkbox in the header area.

PR #1428, opened 2019-02-15
feat(table): add inline editing + buttons

suomiy:
Hello,

these are the styles for pf-react Table (https://github.com/patternfly/patternfly-react/pull/1227). 

It was inspired by pf3 inline-edit (https://github.com/patternfly/patternfly-react/blob/aa839a80a005f5d540faeb0e1a0320b7df1cc982/packages/patternfly-3/patternfly-react/less/inline-edit.less)

few issues:
- I didn't put the confirmation buttons into the examples because their location is computed (depends on scroll and resize)
- not sure if the colors should be the same as in pf3 (light/dark theme?)
- the borders seem to misbehave in firefox


suomiy:
this PR has a lower priority and the styles were temporarily moved to [pf-react](https://github.com/patternfly/patternfly-react/pull/1227) (after a recent discussion)

for future reference: there is a bug in firefox which breaks the table borders: https://bugzilla.mozilla.org/show_bug.cgi?id=688556

Issue #1435, opened 2019-02-18
Register patternfly npm package with a cdn


dgutride:
Should be consistent with patternfly cdn registration

redallen:
Quick comparison:
|                             | Monthly Cost | Hosting | Image Optimization | Cache TTL | Instant purging | Configurability | Points of Presence |
|-----------------------------|--------------|---------|--------------------|-----------|-----------------|-----------------|--------------------|
| Unpkg                       | $0           | Yes     | No                 | Forever   | No              | None            | Uses Cloudflare    |
| jsDelivr                    | $0           | Yes     | No                 | Forever   |  No             | None            | Best Worldwide     |
| cdnjs                       | $0           | Yes     | No                 | Forever   | No              | None            | Uses Cloudflare    |
| Cloudflare                  | $0-50        | No      | Yes                | 2hr-30s   | Yes             | Low             | Worldwide          |
| Fastly                      | $50-100      | No      | Yes                | <1s       | Yes             | High            | Worldwide          |
| KeyCDN                      | $0-20        | Yes     | No                 | ?         | Yes             | High            | Less Worldwide     |
| MetaCDN                     | $0-100       | Yes     | No                 | ?         | Yes             | Low             | Best Worldwide     |
| Amazon Cloudfront           | $0-20        | S3      | No                 | ?         | Yes             | High            | Worldwide, growing |
| Stackpath (formerly MaxCDN) | $10          | Yes     | No                 | ?         | Yes             | High            | Worldwide          |

redallen:
Closed in #1442.

PR #1442, opened 2019-02-19
fix(docs): update CDN url

redallen:
Use unpkg as our CDN instead of https://pf4.patternfly.org/styles.css. Reference #1435.

Example [codepen.](https://codepen.io/anon/pen/wNOJre?&editable=true) There's still the detail of figuring out assets other than CSS, like the Patternfly logo.


redallen:
Closes #1435 .

PR #1444, opened 2019-02-19
feat(dropdown): add support for groups

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1436

@jgiardino how does this look?

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
+++ b/src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
@@ -1,23 +1,21 @@
-<ul class="pf-c-dropdown__menu{{#if dropdown-menu--modifier}} {{dropdown-menu--modifier}}{{/if}}"
+<div class="pf-c-dropdown__menu{{#if dropdown-menu--modifier}} {{dropdown-menu--modifier}}{{/if}}"
   aria-labelledby="{{id}}-button"
   {{#unless dropdown--IsExpanded}}hidden{{/unless}}
   {{#if dropdown-menu--attribute}}
 	  {{{dropdown-menu--attribute}}}
   {{/if}}>
-  <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
-  <li><button class="pf-c-dropdown__menu-item">Action</button></li>
-  <li>
-    <div class="pf-c-dropdown__group-title">Group 1</div>
+  <section class="pf-c-dropdown__group">
+    <h1 class="pf-c-dropdown__group-title">Group 1</h1>
     <ul>
-      <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
-      <li><button class="pf-c-dropdown__menu-item">Action</button></li>
+        <li><a class="pf-c-dropdown__menu-item" href="#">Group 1 Link</a></li>
+        <li><button class="pf-c-dropdown__menu-item">Group 1 Action</button></li>
     </ul>
-  </li>
-  <li>
-    <div class="pf-c-dropdown__group-title">Group 2</div>
+  </section>
+  <section class="pf-c-dropdown__group">
+    <h1 class="pf-c-dropdown__group-title">Group 2</h1>
     <ul>
-      <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
-      <li><button class="pf-c-dropdown__menu-item">Action</button></li>
+        <li><a class="pf-c-dropdown__menu-item" href="#">Group 2 Link</a></li>
+        <li><button class="pf-c-dropdown__menu-item">Group 2 Action</button></li>
     </ul>
-  </li>
-</ul>
+  </section>
+</div>
--- a/src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
+++ b/src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
@@ -4,18 +4,20 @@
   {{#if dropdown-menu--attribute}}
 	  {{{dropdown-menu--attribute}}}
   {{/if}}>
+  <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
+  <li><button class="pf-c-dropdown__menu-item">Action</button></li>
   <li>
-    <div class="pf-c-dropdown__group-title">App</div>
+    <div class="pf-c-dropdown__group-title">Group 1</div>
     <ul>
-      <li><a href="#" class="pf-c-dropdown__menu-item">Configuration</a></li>
-      <li><a href="#" class="pf-c-dropdown__menu-item">Stats</a></li>
+      <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
+      <li><button class="pf-c-dropdown__menu-item">Action</button></li>
     </ul>
   </li>
   <li>
-    <div class="pf-c-dropdown__group-title">User</div>
+    <div class="pf-c-dropdown__group-title">Group 2</div>
     <ul>
-      <li><a href="#" class="pf-c-dropdown__menu-item">Profile</a></li>
-      <li><a href="#" class="pf-c-dropdown__menu-item">Privacy</a></li>
+      <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
+      <li><button class="pf-c-dropdown__menu-item">Action</button></li>
     </ul>
   </li>
 </ul>
--- a/src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
+++ b/src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
@@ -5,10 +5,9 @@
 	  {{{dropdown-menu--attribute}}}
   {{/if}}>
   <section class="pf-c-dropdown__group">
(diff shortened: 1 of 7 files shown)


jgiardino:
Yes, I think this structure makes sense. 

I'm on the fence about whether `.pf-c-dropdown__group-title` should be a heading, but I'm not sure what heading level would make sense. The benefits of using a heading for these is it helps clarify that this text is a heading that starts a new section. What do you think?

src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
```diff
@@ -0,0 +1,21 @@
+<ul class="pf-c-dropdown__menu{{#if dropdown-menu--modifier}} {{dropdown-menu--modifier}}{{/if}}"
+  aria-labelledby="{{id}}-button"
+  {{#unless dropdown--IsExpanded}}hidden{{/unless}}
+  {{#if dropdown-menu--attribute}}
+	  {{{dropdown-menu--attribute}}}
+  {{/if}}>
+  <li>
+    <div class="pf-c-dropdown__group-title">App</div>
+    <ul>
+      <li><a href="#" class="pf-c-dropdown__menu-item">Configuration</a></li>
```
christiemolloy: The other menu-items that I'm seeing in examples above, use button rather than <a></a> tag, unless it says "Link" was this purposeful?

src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
```diff
@@ -0,0 +1,21 @@
+<ul class="pf-c-dropdown__menu{{#if dropdown-menu--modifier}} {{dropdown-menu--modifier}}{{/if}}"
+  aria-labelledby="{{id}}-button"
+  {{#unless dropdown--IsExpanded}}hidden{{/unless}}
+  {{#if dropdown-menu--attribute}}
+	  {{{dropdown-menu--attribute}}}
+  {{/if}}>
+  <li>
+    <div class="pf-c-dropdown__group-title">App</div>
```
christiemolloy: I agree with Jenn that this should use heading tags.

src/patternfly/components/Dropdown/examples/dropdown-groups-example.hbs
```diff
@@ -0,0 +1,5 @@
+{{#> dropdown id="dropdown-basic-example-expanded" dropdown--IsExpanded="true" dropdown--HasToggleIcon="true" dropdown--IsGroups="true"}}
```
christiemolloy: maybe change the ID to dropdown-group-example-expanded.
and ...
dropdown--IsGroups , to dropdown--HasGroups.

src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
mcoker: Yeah, you can use links or buttons in a dropdown. I can change if you you'd prefer it be a button.

christiemolloy: I think it said in the docs that links go off the page and buttons stay on. So i'd update it to what the other menu items use. But thats just me.

mcoker:
@jgiardino I think that makes sense. We do that for nav sections, and [bootstrap does it, too](https://v4-alpha.getbootstrap.com/components/dropdowns/#menu-headers). Though I was looking at some WAI examples and while these are for navigation, it seems similar, and they don't use headings.

* https://www.w3.org/WAI/tutorials/menus/flyout/
* https://www.w3.org/WAI/tutorials/menus/application-menus-code/

Agree re: which level to use. Following the traditional approach to hierarchy of headings, seems like these should use whatever makes sense  based on where it is in the page? ie, if the dropdown is contextually in the content under an `<h2>` on the page, they would be `<h3>`'s?

One way could be to wrap the section title and nested `ul` in a `<section>` element, then use an `<h1>` since [sections can have their own heading hierarchy](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#Defining_section).

src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
mcoker: The other menus use both. I think as long as it's clear that the component can take either, and we have examples of that, I think it's fine to use either. I'll update it to have both like the other examples though. 

screen shot 2019-02-20 at 1 22 04 pm

src/patternfly/components/Dropdown/examples/dropdown-groups-example.hbs
mcoker: good catch. to stay consistent with the other menus, changed it to `dropdown--IsGroupsMenu`

mcoker:
@christiemolloy thanks! updated, except for the heading level until we sort that out. fwiw, that element can be any level heading and it displays properly.

jgiardino:
> One way could be to wrap the section title and nested `ul` in a `<section>` element, then use an `<h1>` since [sections can have their own heading hierarchy](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#Defining_section).

I like this approach. It's consistent with what we're doing for Grouped Nav, except for the heading level, which is defined as `<h2>` in the nav. I think `<h1>` makes sense for this case, based on the assumption that the primary use case for a menu is to treat it like a modal (this method still needs to be tested, though, so that could change). And if the menu behaves like a modal, then only the contents  in the menu are visible. 

If we go with this approach, I don't think you need the nested `<ul>` structure, similar to how each section in the Grouped Nav is a separate section with its own `<ul>`.

mcoker:
@jgiardino thanks. What HTML structure would you propose for a menu with groups?

Currently it's something like:

```
<ul class="pf-c-dropdown__menu">
    <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
    <li><button class="pf-c-dropdown__menu-item">Action</button></li>
</ul>
```

If a menu has groups, would you propose something like:

```
<div class="pf-c-dropdown__menu">
    <ul> // not in a group
        <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li> 
        <li><button class="pf-c-dropdown__menu-item">Action</button></li>
    </ul>
    <section>
        <h1>Group title</h1>
        <ul>
            <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
            <li><button class="pf-c-dropdown__menu-item">Action</button></li>
        </ul>
    </section>
    <section>
        <h1>Another Group title</h1>
        <ul>
            <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
            <li><button class="pf-c-dropdown__menu-item">Action</button></li>
        </ul>
    </section>
</div>

**Comment by jgiardino** [comment]
@mcoker Your suggestion looks good to me. 

Do you know if the use case for having menu items not in a group is to always have them be first? If so, I think this solution is nice. But if we were to have something like:
- section of items not in a group
- section of items in a group
- section of items in a group
- section of items not in a group

Then I'd expect to use sections as parents of each `<ul>`, and to include a separator element, at least before the last section of items not in a group. For example:
```
<section><ul>...</ul></section> // menu items not in a group
<div role="separator>...</div>     // maybe include a separator here for visual separation, but for sr users, the use of the heading is sufficient
<section><h1>...</h1><ul>...</ul></section> /// menu items in a group
<section><h1>...</h1><ul>...</ul></section> /// menu items in a group
<div role="separator>...</div>  // include a separator here for visual separation, also not really sure the separate section is sufficient in all screen readers
<section><ul>...</ul></section> // menu items not in a group
```

If we want to allow this type of flexibility, then _maybe_ all sections of menu items should be wrapped into a `<section>` element, and only what we call groups would include the `<h1>`?

src/patternfly/components/Dropdown/dropdown-menu-groups.hbs
christiemolloy: okay agree with that!

mcoker:
What do you think @kybaker @mcarrano re: jenn's question about use case in https://github.com/patternfly/patternfly-next/pull/1444#issuecomment-466008067?

Would we ever have a group of items followed by items that are not in a group? If so, how would we separate those? 

@jgiardino curious why you couldn't just have a `ul` adjacent to a `section` like in the example I showed? 

And re:

> If we want to allow this type of flexibility, then _maybe_ all sections of menu items should be wrapped into a `<section>` element, and only what we call groups would include the `<h1>`?

do you mean updating all dropdowns, including ones that have a single list of items (no groups) to be in a `<section>` or would this only apply to dropdowns that include at least 2 groups?

jgiardino:
> @jgiardino curious why you couldn't just have a `ul` adjacent to a `section` like in the example I showed?

I would treat each section consistently—named or not—so that in cases/screen readers where the sections are announced, each section is announced in a consistent way. So either they all have sections, or they don't have sections. 

If we expect the organization to always be as you show it, where there's a clear separation between named groups and the other menu items, then I think what you have is fine. But given that this is an open source design system, we can't expect that's always going to be the case.

mcoker:
@jgiardino updated

jgiardino:
Looks great!

PR #1445, opened 2019-02-20
feat(icons): add font fabulous icons

karelhala:
This will add to font definitions https://github.com/ManageIQ/font-fabulous font.


karelhala:
I don't know about font files and their definition, is there any automation how to do that or is that by hand?

mcoker:
Hey @karelhala, that file manages the custom pf3 icons we're carrying over to pf4. Adding a font library to that file likely isn't ideal as I believe we'll run into namespace issues doing that in the future, and a person should be able to opt in or out of that if they aren't required by default.

What are these icons needed for and what's the reasoning behind adding them to core?

I believe we had talked about adding Font Fabulous before if I'm not mistaken and we chose not to. I don't remember the details, @dgutride do you remember anything about that?

All that said, there is an npm script `npm run buildpficons` that creates SVG files out of the definitions in `pf-icons.json`, then `npm run build:pficonfont` builds the icon font files. Then you run `./build/iconList.sh` and check in to update the list of icons served from the `/icons` in the dev workspace.

karelhala:
@mcoker I believe that it was requested by ManageIq team as nice to have. We don't have to merge this PR if we don't want to expose these icons. I just stumbled upon these icons and heard that it would be nice to have in MiQ (they could replace all icons with react-icons).

mcoker:
@karelhala great, thanks! I'm inclined to only ship icons with PF that are used by PF. Otherwise, it should be up to the individual projects to add whatever icons are needed to their product. What do you think @dgutride @kybaker

karelhala:
Mhhh, this makes complete sense to ship only FA icons. But as far as I know the initial idea behind [@patternfly/react-icons](https://www.npmjs.com/package/@patternfly/react-icons) was to ship as many icons as possible. And since FF is not perfectly styled I thought that it would be good to put them here, I think that the best way would be to add this file (icon definition file) to pf-react icons and expose them over this package only.

mcoker:
@karelhala I'll let @dgutride or @kybaker respond about `react-icons`. One issue with adding them to this file in core is that we build an icon font out that definition file. By default, those icons would then be added to the `pf-icon` icon font we have (you can see that at https://pf4.patternfly.org/icons). So if we do decide to keep them, we would need to change the way we build the icon font file so that we don't build font fabulous by default, or we might need to create a separate JSON file for FF, and then `react-icons` could read in that file on top of `pf-icons.json`

rachael-phillips:
Approved to be closed

PR #1447, opened 2019-02-20
Change padding around expandable rows.

### LINKED ISSUE #1424: Table expanding rows have large margins
Implemented the Table component from patternfly/patternfly-react, but the expanding row layout looks odd in the Insights/Cost Management UI.

It appears that the expandable row reserves space on the left side, presumably for additional `td` tags containing toggles, check boxes, etc? As a result, it looks like there is a large margin on the left. 

Looking at the core examples, the expandable row also has extra `td` tags.

In order to implement the designs created for Insights/Cost Management, I want to make use of that additional space in the expandable row. Currently, everything is shifted to the right.

screen shot 2019-02-12 at 1 59 47 pm

I filed the issue below against patternfly/patternfly-core, but it was closed stating this was deliberate? Seems that we could make the td tags span the entire row, so we can make use of the space.

Still, the margins for the implementation in patternfly/patternfly-react don't appear to match core. The margin on the left appears to be larger than core, while right is minimal.

https://github.com/patternfly/patternfly-react/issues/1376

christiemolloy:
Fix #1424 

Expandable rows in tables should have the same padding as expandable rows in data lists. Currently in tables, we use the columns from the above row to determine spacing around content, but this is not robust enough for products to use.

Fix: expandable rows by default have no padding i.e. if you want to include something that has its own padding / the nested table in compound expansion / or when there is no required padding. 
using `.pf-c-table__expandable-row-content` as the child of `.pf-c-table__expandable-row` will apply the necessary spacing.

### DIFF
--- a/src/patternfly/components/Table/docs/table-expandable.md
+++ b/src/patternfly/components/Table/docs/table-expandable.md
@@ -6,6 +6,8 @@
 <br>
 <mark>**All checkbox/action button accessibility and usage requirements apply.**</mark>
 
+Note: To apply padding to `.pf-c-table__expandable-row` wrap the content in `.pf-c-table__expandable-row-content`. 
+
 ### Accessibility
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
--- a/src/patternfly/components/Table/docs/table-expandable.md
+++ b/src/patternfly/components/Table/docs/table-expandable.md
@@ -6,7 +6,7 @@
 <br>
 <mark>**All checkbox/action button accessibility and usage requirements apply.**</mark>
 
-Note: To apply padding to `.pf-c-table__expandable-row` wrap the content in `.pf-c-table__expandable-row-content`. 
+Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.pf-c-table__expandable-row-content`. 
 
 ### Accessibility
 | Attribute | Applied To | Outcome |
--- a/src/patternfly/components/Table/examples/table-expandable-example.hbs
+++ b/src/patternfly/components/Table/examples/table-expandable-example.hbs
@@ -43,21 +43,16 @@
       {{/table-td}}
     {{/table-tr}}
 
-    {{#> table-tr table-tr--expandable="true" table-tr--modifier="pf-m-expanded"}}
-      {{#> table-td table-td--attribute='colspan="7"'}}
-        <div class="pf-c-table__expandable-row-content" id="{{table--id}}-content1">
-          Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
-          tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
-          quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
-          consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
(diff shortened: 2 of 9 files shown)


src/patternfly/components/Table/docs/table-expandable.md
```diff
@@ -6,6 +6,8 @@
 <br>
 <mark>**All checkbox/action button accessibility and usage requirements apply.**</mark>
 
+Note: To apply padding to `.pf-c-table__expandable-row` wrap the content in `.pf-c-table__expandable-row-content`.
```
mcoker:
```suggestion
Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.pf-c-table__expandable-row-content`. 
```

mcoker:
Do you think it's worth adding a default padding or margin to the expandable row that accounts for the expanded left border-width?

screen shot 2019-02-20 at 10 34 57 am

Can you update the other expandable example that isn't expanded? I wonder if that example should be expanded and not include the expandable content wrapping element?

screen shot 2019-02-20 at 10 39 40 am

christiemolloy:
Thanks @mcoker !! Agree with your last two comments so made those updates

src/patternfly/components/Table/examples/table-expandable-example.hbs
```diff
@@ -43,21 +43,16 @@
...
-      {{#> table-td table-td--attribute='colspan="7"'}}
-        <div class="pf-c-table__expandable-row-content" id="{{table--id}}-content1">
-          Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
-          tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
-          quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
-          consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
-          cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
-          proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+    {{#> table-tr table-tr--expandable="true" table-tr--IsExpanded="true"}}
+      {{#> table-td table-td--attribute=(concat 'colspan="7" id="' table-id '-content1"')}}
```
mcoker: the `id` isn't generating. `table-id` should be `table--id`

src/patternfly/components/Table/examples/table-expandable-example.hbs
```diff
@@ -82,22 +77,9 @@
...
-          tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
-          quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
-          consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
-          cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
-          proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
-        </div>
-      {{/table-td}}
-      {{#> table-td}}
+    {{#> table-tr table-tr--expandable="true" table-tr--IsExpanded="true"}}
+      {{#> table-td table-td--attribute=(concat 'colspan="7" id="' table-id '-content1"')}}
```
mcoker: `table-id` to `table--id`

PR #1451, opened 2019-02-20
Add icon in Data List 

### LINKED ISSUE #1389: Add support on adding an Icon to the DataListItem.
Issue [#1338](https://github.com/patternfly/patternfly-react/issues/1338) was opened in the React repo.  The consumer would like to add an icon to the DataListCell.  The consumer followed the [pattern library](https://www.patternfly.org/pattern-library/content-views/list-view/) and tried to add an icon on the left.  This Pattern Library is PF3 specific and currently PF4 does not support the icon being placed to the left as in PF3.

christiemolloy:
Closes #1389 

Adds a class called `.pf-c-data-list__icon` which styles an icon similar to a checkbox in the data table. Added icon the expandable table example - but it would be good to clarify which examples should use this icon/if any.

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -58,6 +58,14 @@ $pf-data-list--breakpoint--mobile: 380px;
     --pf-c-data-list__check--MarginRight: var(--pf-c-data-list__check--mobile--MarginRight);
   }
 
+  // Icon
+  --pf-c-data-list__icon--MarginRight: var(--pf-global--spacer--xl);
+  --pf-c-data-list__icon--mobile--MarginRight: var(--pf-global--spacer--md);
+
+  @media screen and (max-width: $pf-data-list--breakpoint--mobile) {
+    --pf-c-data-list__icon--MarginRight: var(--pf-c-data-list__icon--mobile--MarginRight);
+  }
+
   // Actions
   --pf-c-data-list__action--MarginLeft: var(--pf-global--spacer--lg);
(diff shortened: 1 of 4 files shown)


mcarrano:
Can you elaborate on this @christiemolloy ? In not sure what this change is doing. What am I looking for?

christiemolloy:
@mcarrano sorry ill update the description. Take a look at the expandable table with the updated icons.

mcarrano:
Ah, OK I see it @christiemolloy . I was getting confused because the issue says Data List, but then you talk about the expandable table :)  When you did looks fine to me, although I'm still unclear what the purpose of this example is.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -162,6 +171,11 @@ $pf-data-list--breakpoint--mobile: 380px;
   margin-right: var(--pf-c-data-list__check--MarginRight);
 }
 
+.pf-c-data-list__icon {
+  display: flex;
```
mcoker: is this needed?

mcoker:
@christiemolloy looks great, just one comment.

src/patternfly/components/DataList/data-list.scss
christiemolloy: no, also took it off data-list__check.

christiemolloy:
@mcoker updated. @mcarrano the purpose of adding the icons is to demonstrate that it is possible to have icons in the data list, before this, it was not possible to add an icon and have it style correctly in the data list - refer to the issue #1389

PR #1454, opened 2019-02-21
feat(ProgressBarUpdate): aria update for progress bar component

### LINKED ISSUE #1128: Progress bar example updates
Some updates to our progress bar examples are needed to better align with our design documentation and also provide better clarity regarding how to utilize the progress bar for different use cases. 

- The text string "Title" should display as the primary text
  - For most examples, "Title" would replace "Descriptive text here"
  - For the success and failure examples, "Title" would replace the strings "Success" and "Failure due to an error"
- The example labeled "Progress with Dynamic Description" can support a couple of use cases. To more clearly represent these use cases in our examples, we should create two versions of this example showing the following use cases:
  1. progress value is numeric but not a percentage, example 1
      - In this case, the value displays on the right where the percentage value currently displays, but in the format "2 of 5 units"
      - use `aria-valuetext="2 of 5 units"` in this example
      - other attributes can be set to `aria-valuemin="0" aria-valuemax="5" aria-valuenow="2"`
  2. progress value is numeric but not a percentage, example 2
      - In this case, the value displays on the right where the percentage value currently displays, but as a text string that describes the current value. 
      - The text string for the example we show could say "Step 2: Copying files"
      - use `aria-valuetext="Step 2: Copying files"`
      - other attributes should still be included if they can accurately reflect current progress, e.g.: `aria-valuenow="2" aria-valuemin="0" aria-valuemax="4"`

- The following documentation about `aria-valuetext` should be included:
    > Provide a text string that communicates current status. Only use if the important information about status is included in the text string. Do no use if percentage is the most important value to communicate. Some screen readers will ignore the percentage value determined from `aria-valuenow` when `aria-valuetext` is used.

jameslau:
fixes #1128 

- Adding new progress bar updates to reflect different use cases
- Adjusted handlebar configuration to better reflect dynamic aria values
- Replaced documentation instructions

screen shot 2019-02-20 at 10 05 31 pm

### DIFF
--- a/src/patternfly/components/Progress/docs/code.md
+++ b/src/patternfly/components/Progress/docs/code.md
@@ -11,7 +11,7 @@ Note: If this component is describing the loading progress of a particular regio
 | `aria-valuemin="0"` | `.pf-c-progress` |  The minimum value for the progress bar. |
 | `aria-valuemax="100"` | `.pf-c-progress` |  The maximum value for the progress bar. |
 | `aria-describedby="[id of .pf-c-progress__description]"` | `.pf-c-progress` |  The description of what progress is being shown. |
-| `aria-valuetext="[loading state]"` | `.pf-c-progress` |  Text that explains the current state of the progress (used in place of aria-describedby only when the text is updated dynamically as progress proceeds). |
+| `aria-valuetext="[loading state]"` | `.pf-c-progress` |  Provide a text string that communicates current status. Only use if the important information about status is included in the text string. Do no use if percentage is the most important value to communicate. Some screen readers will ignore the percentage value determined from `aria-valuenow` when `aria-valuetext` is used. |
 
 
 ## Usage
--- a/src/patternfly/components/Progress/docs/code.md
+++ b/src/patternfly/components/Progress/docs/code.md
@@ -9,9 +9,9 @@ Note: If this component is describing the loading progress of a particular regio
 | `role="progressbar"` | `.pf-c-progress` |  This role is used for an element that displays the progress status for a task that takes a long time or consists of several steps. |
 | `aria-valuenow=""` | `.pf-c-progress` |  This value needs to be updated as progress continues. |
 | `aria-valuemin="0"` | `.pf-c-progress` |  The minimum value for the progress bar. |
-| `aria-valuemax="100"` | `.pf-c-progress` |  The maximum value for the progress bar. |
+| `aria-valuemax="100"` | `.pf-c-progress` |  The maximum value for the progress bar. If the progress is only defined using `aria-valuenow` (e.g a percentage), the value should be set to "100". If the progress is defined using `aria-valuetext`, then this value can be a number other than 100. For example, if `aria-valuetext` is "2 of 5 units", then `aria-valuemax` can be "5" and `aria-valuenow` can be "2". |
 | `aria-describedby="[id of .pf-c-progress__description]"` | `.pf-c-progress` |  The description of what progress is being shown. |
-| `aria-valuetext="[loading state]"` | `.pf-c-progress` |  Provide a text string that communicates current status. Only use if the important information about status is included in the text string. Do no use if percentage is the most important value to communicate. Some screen readers will ignore the percentage value determined from `aria-valuenow` when `aria-valuetext` is used. |
+| `aria-valuetext="[loading state]"` | `.pf-c-progress` |  Provide a text string that communicates current status. Only use if the important information about status is included in the text string. Do not use if percentage is the most important value to communicate. Some screen readers will ignore the percentage value determined from `aria-valuenow` when `aria-valuetext` is used. |
 
 
 ## Usage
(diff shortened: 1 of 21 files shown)


src/patternfly/components/Progress/docs/code.md
```diff
@@ -11,7 +11,7 @@ Note: If this component is describing the loading progress of a particular regio
 | `aria-valuemin="0"` | `.pf-c-progress` |  The minimum value for the progress bar. |
 | `aria-valuemax="100"` | `.pf-c-progress` |  The maximum value for the progress bar. |
 | `aria-describedby="[id of .pf-c-progress__description]"` | `.pf-c-progress` |  The description of what progress is being shown. |
-| `aria-valuetext="[loading state]"` | `.pf-c-progress` |  Text that explains the current state of the progress (used in place of aria-describedby only when the text is updated dynamically as progress proceeds). |
+| `aria-valuetext="[loading state]"` | `.pf-c-progress` |  Provide a text string that communicates current status. Only use if the important information about status is included in the text string. Do no use if percentage is the most important value to communicate. Some screen readers will ignore the percentage value determined from `aria-valuenow` when `aria-valuetext` is used. |
```
dgutride: 'Do not use' (not Do no use)

src/patternfly/components/Progress/progress-indicator.hbs
```diff
@@ -1,5 +1,5 @@
 <div class="pf-c-progress__indicator{{#if progress-indicator--modifier}} {{progress-indicator--modifier}}{{/if}}" 
-  style="width: {{progress__value}}%;"
+style="width: {{~#if progress__width}}{{progress__width}}{{else}}{{progress__value}}{{/if}}%;"
```
dgutride: Keep a close eye on the handlebar spacing - the two spaces at the beginning of the second line should not be removed.

src/patternfly/components/Progress/progress__measure.hbs
```diff
@@ -1,4 +1,8 @@
 {{!-- This provides the measure with a % --}}
 {{#if progress__value}}
-  <span class="pf-c-progress__measure">{{progress__value}}%</span>
+  {{~#if progress--dynamic}}
+    <span class="pf-c-progress__measure">{{progress__valuetext}}</span>
+    {{else}}
```
dgutride: Else should be moved back two spaces to align with the if start/end blocks and add visual continuity to the logic functions.

src/patternfly/components/Progress/docs/code.md
jgiardino: 😬 Doh! that was me. Good catch!

jameslau: Teamwork!

jgiardino:
Thanks for making these updates, @jameslau! They look great! 

The only fix I see (besides the typo @dgutride found) is that the example labeled "Progress with Dynamic Description" can be deleted from this page. We're replacing that one with the two new examples you added.

Also, it might be good to get a designer to take a look and make sure these updates match what they were expecting. I had discussed these changes with @rachael-phillips and @mceledonia when I created the issue for this. Could one of you take a quick look?

jgiardino:
I have one more request for this PR. 

1. Where `aria-valuemax="100"` is included in the documentation, I think it would be helpful to clarify when it should be "100" versus some other value. Here's what we have currently:
    > The maximum value for the progress bar.

    Here's an updated description:
    > The maximum value for the progress bar. If the progress is only defined using `aria-valuenow` (e.g a percentage), the value should be set to "100". If the progress is defined using `aria-valuetext`, then this value can be a number other than 100. For example, if `aria-valuetext` is "2 of 5 units", then `aria-valuemax` can be "5" and `aria-valuenow` can be "2".

2. Where the instructions display for the new examples, I think it would be helpful to revise the text to reflect what's happening in the new examples. Here's what we have currently:
    > If the description is updated dynamically, then the ARIA tag aria-valuetext should be used to provide this dynamic description.

    Here's an updated description:
    > If the status that displays with the bar is not a percentage, then the ARIA tag `aria-valuetext` should be used to provide this status to screen reader users. This is the only case when setting the `aria-valuemax` to a value other than "100" is recommended, given how different screen readers handle these attributes.

src/patternfly/components/Progress/docs/code.md
jameslau: Typo correction made.

src/patternfly/components/Progress/progress-indicator.hbs
jameslau: Line break removed.

src/patternfly/components/Progress/progress__measure.hbs
jameslau: Spacing adjusted.

jameslau:
Hi @jgiardino, I just pushed up new updates for the progress bar. Please have a look.

@mceledonia and @rachael-phillips, can we get some design feedback on the new changes?

mceledonia:
This looks good to me! Thanks again!

srambach:
👍

PR #1459, opened 2019-02-21
Fix Data List w Modifiers Example

### LINKED ISSUE #1327: Data-List with Modifiers issue
In the Data List with modifiers example, at a certain screen size, the Data List looks like this (this doesn't happen in React):
screen shot 2019-01-28 at 10 57 13 am

-  the kebab icon is filtering into .pf-c-data-list__expandable-content
- .pf-c-data-list__expandable-content is not full width

christiemolloy:
Fix #1327

Currently, the data list doesn't support three columns of content in its row, thats why the action kebab is overflowing to the next line.  
Two fixes: larger breakpoint or remove the third column from the "Flex Modifiers - Example 3".
I wonder if it should support three columns without breaking though?

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -58,14 +58,6 @@ $pf-data-list--breakpoint--mobile: 380px;
     --pf-c-data-list__check--MarginRight: var(--pf-c-data-list__check--mobile--MarginRight);
   }
 
-  // Icon
-  --pf-c-data-list__icon--MarginRight: var(--pf-global--spacer--xl);
-  --pf-c-data-list__icon--mobile--MarginRight: var(--pf-global--spacer--md);
-
-  @media screen and (max-width: $pf-data-list--breakpoint--mobile) {
-    --pf-c-data-list__icon--MarginRight: var(--pf-c-data-list__icon--mobile--MarginRight);
-  }
-
   // Actions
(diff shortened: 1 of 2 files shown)


mcoker:
Hmm... yeah, seems kinda fragile the way it's currently set up. I wonder if everything in the row except for the kebab should be an a flex row that wraps, and the kebab is outside of that row and is always aligned on the right. Then the kebab would always be on the right and we wouldn't have to change `order` on `.pf-c-data-list__cell`. But I think if we want to fix this example without introducing a breaking change, this is probably the only way to do it.

christiemolloy:
@mcoker when you say - "But I think if we want to fix this example without introducing a breaking change, this is probably the only way to do it."  Do you mean what I had done or to implement what you suggested?

mcoker:
The changes you have in this PR just address the breaking example in the workspace, which is fine as a fix for now. There is still the underlying issue of when a person adds more content in the row than will allow and the row wraps before the breakpoint that changes the `order` on the row cells. So I think we can merge this to fix the workspace, then at another point, fix the layout so that you can put as many items as you want in the row and those items wrap but the kebab stays in the top/right. You could just open another issue to address that and we can make it in the April release since that likely requires a different HTML structure and is a breaking change.

christiemolloy:
@mcoker Yes definitely agree, I'll create an issue now

PR #1460, opened 2019-02-21
fix(patternfly.scss): add extension to import for patternfly-base.scss

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1370


mcoker:
Can we merge this one?

PR #1461, opened 2019-02-22
Unique ids

### LINKED ISSUE #1439: IDs used in ARIA and labels must be unique
Below are the URLs that illustrate the problem:

-  /components/Table/examples-full/
- /components/Login/examples-full/

This is a `critical` a11y issue.

To learn more about how to build and test locally, check out: https://github.com/patternfly/patternfly-next#testing-for-accessibility

**As part of this issue, please complete the following items.** These will help reviewers know what issues were addressed in the PR, and also will help us keep our total a11y violation count down.

1. Run the a11y audit, copy the issues from the audit that were fixed in the PR, and paste them into the PR description. 
    - Instructions for running the audit are here: https://github.com/patternfly/patternfly-next#testing-for-accessibility
    - Issues that are returned in the terminal look like this. Find the issues you are addressing and copy the following information provided for them or take a screencap to include in the PR.
      ```
      1 error found in: /upgrade-examples/AlertUpgradeExamples/examples-full/

      1) Buttons must have discernible text -- https://dequeuniversity.com/rules/axe/3.1/button-name -- impact: critical
      
      -------html violation instances-------

      <button class="btn btn-link dropdown-toggle" type="button" id="dropdownKebabRight" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
        <span class="fa fa-ellipsis-v"></span>
      </button> 
      ```
2. Reduce the threshold by number of issues being addressed 
    - This update is made [here](https://github.com/patternfly/patternfly-next/blob/master/test-a11y/config.js#L2)
    - The new value should be derived by: 
      - subtracting the number of issues you are addressing in the PR (e.g. the line that looks like this in the info you copied from the audit report in item 1 above):
        `5 errors found in: /components/Toolbar/examples-full/`
      - from current violations (e.g. this line in the audit report, near the end):
        `20 violations in total`

srambach:
Fixes duplicate IDs on the table and toolbar components. This fixes part of #1439. The duplicate IDs on SVG filters will be addressed separately.
This fix allows the tolerance to be reduced by 3.

image

### DIFF
--- a/src/patternfly/components/Table/table-nested.hbs
+++ b/src/patternfly/components/Table/table-nested.hbs
@@ -13,7 +13,7 @@
       {{#> table-td table-td--data-label="Date"}}May 9, 2018{{/table-td}}
       {{#> table-td table-td--data-label="Status"}}Active{{/table-td}}
       {{#> table-td table-td--action="true"}}
-        {{#> dropdown id="dropdown-kebab-right-aligned-nested-1" dropdown--IsActionMenu="true" dropdown-menu--modifier="pf-m-align-right" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+        {{#> dropdown id=(concat table-nested--id '-dropdown-kebab-' table-tr--id) dropdown--IsActionMenu="true" dropdown-menu--modifier="pf-m-align-right" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
         {{/dropdown}}
       {{/table-td}}
     {{/table-tr}}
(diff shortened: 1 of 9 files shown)


jameslau:
This looks good @srambach! The a11y checks in and the errors dropped from 20 to 16. 👍

jgiardino:
Thanks for making this update!

I'm a little confused by the a11y audit tool. When I look at the [details](https://travis-ci.com/patternfly/patternfly-next/builds/101945647), it still returns the same issues for the Table.

But the audit shows that this element has an issue:
`<button class="pf-c-dropdown__toggle pf-m-plain" id="-dropdown-kebab-nested-tr1-button" aria-expanded="false" aria-label="Actions">`

Whereas the actual HTML in the preview for this PR has this (the id values are different):
`<button class="pf-c-dropdown__toggle pf-m-plain" id="nested-table-1-dropdown-kebab-nested-tr1-button" aria-expanded="false" aria-label="Actions">`

Also, when I try to find the duplicate values for this on the page, I'm not finding it. Maybe this is just an issue with the a11y audit (@seanforyou23)?

srambach:
How's it look after re-running, @jgiardino ? My only concern is that I had it reducing from 20 to 17 errors, @jameslau saw it as down to 16 errors, and the latest run says 18 errors.

mcoker:
It's 16 for me locally, too.

PR #1462, opened 2019-02-22
feat(popover): adds sm and lg width modifiers for the popover

### LINKED ISSUE #1101: Popover Enhancement
Currently we have set min and max width variables for the popover - as per design specs. 
React was asking how a consumer would override the max-width for the popover? 
Should we create a pf-m-lg modifier to modify the max-width of the popover so that we can handle it in Core first?

 @mcarrano @jschuler @mceledonia @jgiardino

christiemolloy:
Closes #1101 

Similar to the modal, the popover now has a default width of 100%, with sm and lg modifiers to modify its max-width. React can pass a width prop to specify a custom width.


mcoker:
@christiemolloy this looks really good. My only concern is that you can use `appendTo` in react to append the popover to an element. I'm curious if, for example, you have an icon or small text box that you append the popover, too, and if the popover has `width: 100%; max-width: [whatever];`, then the width of the popover will only be as wide as the width of the thing it's going inside of if the popover is `position: absolute` and the element it's appended to is `position: relative;`.  I think this might require testing in react to see how it behaves.

Also I'm wondering if we should remove `width: 100%`, and follow the behavior of [pf3 popovers](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/popovers.html) where the popover is only as big as the content inside of it up to a `max-width`? We could do something similar to the button component where by default the popover is only as wide as the content inside of it, but you can use a `.pf-m-block` modifier to make it `width: 100%`? What do you think? That's likely also a question for design.

christiemolloy:
@mcoker agreed I had actually made those changes to the tooltip, and I confirmed with @mceledonia that we want the same styles for both. Just need to decide should we have a small max-width as `.pf-m-sm` and a large max-width as `.pf-m-lg`,  or one standard max-width on the component.

src/patternfly/components/Popover/popover.scss
```diff
@@ -34,8 +34,7 @@
   --pf-c-popover__footer--MarginTop: var(--pf-global--spacer--lg);
 
   position: relative;
-  min-width: var(--pf-c-popover--MinWidth);
-  max-width: var(--pf-c-popover--MaxWidth);
+  display: inline-block;
```
mcoker: this shouldn't be necessary. The popover will always be in an element with `position: absolute` in react, so it will only be as big as is needed to fit the content inside up to the specified `max-width` (if that's specified)

mcoker:
> Just need to decide should we have a small max-width as `.pf-m-sm` and a large max-width as `.pf-m-lg`, or one standard max-width on the component.

The pros of having 2 classes are it's more flexible and allows for people to have a big popover. The con is the component now requires a modifier class or react prop (I'm assuming anyways, maybe `isSmall` or whatever would be the default?). Are there any other considerations. I could go either way with it. It would depend on what users want/expect. I'm inclined to offer a single width just to make it easy, and something like a tooltip/popover will likely run into issues if it's too big, since it's positioned absolutely on the page and has the potential to go off screen easily unless the tooltip is in the middle of the page. FWIW, looks like carbon and bootstrap offer a single width as far as I can tell, and that's 240px and 276px.

christiemolloy:
Made changes @mcoker . I agree with just having one max-width. @mceledonia mentioned that the popover should wrap to the size of the content in it. And for that to happen it looks like the popover needs `position: relative/display: inline-block` OR `position: absolute` . only having position: relative and a max-width doesn't allow the popover to wrap to the content inside of it. Let me know what you think about this.

mcoker:
> the popover needs `position: relative/display: inline-block` OR `position: absolute`

react sets the position of the popover to `absolute`. You can see in the [popover component on react](http://patternfly-react.surge.sh/patternfly-4/components/popover) how that works. Here's a screenshot of what it does currently with with little content.

screen shot 2019-02-26 at 9 25 00 am

Does anything need to change with this? the only difference I see is there is no longer a `min-width` and the old `max-width` was `calc(var(--pf-c-popover--c-button--sibling--PaddingRight) + #{pf-size-prem(300px)});`, which accounts for the right-padding of the element that follows the close button. I think your change to remove it is fine if visuals are fine with it, noting that the overall popover width is 300px, but the content is more narrow because of the popover's left/right padding and then the element that follows the close button is more narrow than that.

christiemolloy:
Thanks for explaining the positioning - makes sense now. I think we should just close this Pull Request since we decided not to have the small and large modifiers like the issue was intended.

PR #1466, opened 2019-02-22
refactor(tooltip): change width

### LINKED ISSUE #1463: Update width of tooltip
Design suggests changing the max-width from 200px to 300px.

christiemolloy:
Close #1463 .

The tooltip currently has a set width. The width of the tooltip should be the width of the content inside of the tooltip + padding, and wrap at a max-width of 540px.


src/patternfly/components/Tooltip/tooltip.scss
```diff
@@ -20,6 +20,7 @@
   --pf-c-tooltip__arrow--m-left--Transform: translate(50%, -50%) rotate(45deg);
 
   position: relative;
+  display: inline-block;
```
mcoker: Same deal here as popover, I don't think we need `display` as it will be set by react. Here's what the component looks like currently with a small amount of text, shrinking to fit the content.

screen shot 2019-02-26 at 12 30 48 pm

mcoker:
This seems pretty wide for a tooltip. It will exceed the viewport width of a mobile device. What's the reasoning behind changing it to 540px? For reference, bootstrap uses 200px, material uses 300px, and carbon looks like it uses ~180px.

christiemolloy:
@mcoker I think we can also close this. I had spoken to Michael briefly about this width but its definitely too wide now that I think about.

mceledonia:
Good call, let's go with 300px... If I recall that was a late friday afternoon decision :)

PR #1467, opened 2019-02-22
chore(packaging): dismantle patternfly-base into multiple files

seanforyou23:
This PR breaks up patternfly-base.scss into multiple files. Why? It was doing too much! Multiple applications have run into issues when attempting to include only a single component from patternfly4 (masthead) and global styles (included as part of patternfly-base) were upsetting fonts and various spacing throughout the application. These changes make it possible for users to import only the bits they need and in turn keeps people from having to take further steps to exclude the things they never wanted in the first place.

So, with these changes, we have an alternate path for incrementally integrating patternfly4 into patternfly3 (and other) existing UI's. For example, I don't want fonts so I just don't include them like so:

```scss
@import "~@patternfly/patternfly/sass-utilities/all";
@import "~@patternfly/patternfly/variables";
@import "~@patternfly/patternfly/patternfly-shield-noninheritable";
@import "~@patternfly/patternfly/patternfly-common";
@import "~@patternfly/patternfly/themes";
@import "~@patternfly/patternfly/icons";
```

(You can do the same thing using the generated CSS files)

This is sort of the inverse from what we recommend currently, where consumers have to import the entire `patternfly-base` and must set variables to exclude the things they don't want. This requires more maintenance on our part and also has noteworthy limitations like not being able to conditionally include an import reference. I think we should consider deprecating this approach:

```scss
$pf-global--image-path: "~@patternfly/patternfly/assets/images" !default;
$pf-global--enable-reset: false !default;
$pf-global--enable-font-overpass-cdn: false !default;
$pf-global--enable-fontawesome-cdn: true !default;
$pf-global--disable-fontawesome: false !default;
$pf-global--load-pf-3: false !default;
@import "~@patternfly/patternfly/patternfly-base";
```

This should not be a breaking change. Please test and verify this!


seanforyou23:
Awesome! I've received some offline feedback that I think will make a good addition to this PR. It's taking this same strategy a little further by ensuring there are CSS files emitted for each of those `@import "./_*";` files. This way, users who don't use SCSS will have the same advantage of being able to import only what they need. Working on this now, will update this PR once that's all in place so it's all captured in one easy to understand PR. Thanks for all the feedback!

seanforyou23:
Alrighty, I've addressed all feedback to date and think this work is ready for a final review. I've tested with latest Syndesis UI and also verified the IE-11 post process still works using the seed app. The new files that are now available as part of the core distribution are:

- _common.scss
- _fa-icons.scss
- _globals.scss
- _pf-icons.scss
- _shield-inheritable.scss
- _shield-noninheritable.scss
- patternfly-common.css
- patternfly-fa-icons.css
- patternfly-fonts.css
- patternfly-globals.css
- patternfly-icons.css
- patternfly-pf-icons.css
- patternfly-shield-inheritable.css
- patternfly-shield-noninheritable.css
- patternfly-themes.css
- patternfly-variables.css

Let me know if you find any issues, otherwise please feel free to go ahead and merge when ready. Thx @mcoker for all the help and making this possible!

PR #1472, opened 2019-02-25
fix(table): update compact table check box

### LINKED ISSUE #1450: Fix Compact Table
A list of issues to fix with the compact table.

1. At a certain breakpoint the data in the compact table is not aligning with the checkboxes. 
screen shot 2019-02-20 at 3 05 04 pm

2. The top and bottom rows have a larger padding/margin which is impacting the alignment.

3. Create a class for the icon that handles it centering on desktop and left aligned on mobile.

4. Change real names to fake names.

christiemolloy:
Fix #1450 

`.pf-c-table__check` is currently set to `vertical-align: top`. This works for "table with checkbox and actions" / "table expandable" because the checkbox lines up with the expandable arrow and the table columns are wider so it is less likely for the content to wrap more than 2 lines. However in the "compact table" the columns are so narrow that content will wrap over multiple lines - the checkbox set to top alignment looks visually off - note the image in #1450. Setting `.pf-c-table__check` to vertical-align: middle` for the compact table solves this problem.

### DIFF
--- a/src/patternfly/components/Table/examples/table-compact-example.hbs
+++ b/src/patternfly/components/Table/examples/table-compact-example.hbs
@@ -19,7 +19,7 @@
       {{#> table-th table-th--attribute='scope="col"'}}
         Numbers
       {{/table-th}}
-      {{#> table-th table-th--attribute='scope="col"' table-th--modifier="pf-m-center"}}
+      {{#> table-th table-th--attribute='scope="col"'}}
         Icons
       {{/table-th}}
       {{#> table-th}}{{/table-th}}
@@ -32,8 +32,8 @@
       {{#> table-td table-td--check="true"}}
         <input type="checkbox" name="checkrow1" aria-labelledby="name1">
       {{/table-td}}
-      {{#> table-th table-th--data-label="Contributor Name"}}
-        <span id="name1">Andres Galant</span>
+      {{#> table-th table-th--data-label="Contributor"}}
+        <span id="name1">Sam Jones</span>
       {{/table-th}}
       {{#> table-td table-td--data-label="Position"}}
         CSS Guru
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -455,6 +455,14 @@
     --pf-c-table-cell--PaddingLeft: var(--pf-c-table__action--PaddingLeft);
   }
 
+  .pf-c-table__check {
```
mcoker: All of the table cells should be top-aligned. The table vertical alignment should match the data list component. Looks like you could probably just add `vertical-align: baseline;` to `.pf-c-table tr > *`

src/patternfly/components/Table/table-grid.scss
```diff
@@ -150,6 +150,10 @@ $pf-c-table-xs-breakpoint: 380px;
     }
   }
 
+  .pf-c-table__icon > * {
```
mcoker: This is another instance where I wonder if we're better off with a `__content` element in the cells that holds the content. Right now if you put 2 icons in this cell, they stack on top of one another. Or if you put text beside the icon, the text displays in the left column.

screen shot 2019-03-01 at 1 26 16 pm


Someone would need to know to put this content in a div for things to behave as they're intended. Then we could also style `.pf-c-table__icon > .pf-c-table__cell-content { text-align: left; }`. What do you think? I think the way it is right now isn't necessarily intuitive. Also it feels kinda weird to center the icon this way. If this column is only supposed to have icons, you could use `justify-items: start;` on `.pf-c-table__icon` to align the content to the start of the grid columns.

src/patternfly/components/Table/examples/table-compact-example.hbs
```diff
@@ -19,7 +19,7 @@
       {{#> table-th table-th--attribute='scope="col"'}}
         Numbers
       {{/table-th}}
-      {{#> table-th table-th--attribute='scope="col"' table-th--modifier="pf-m-center"}}
```
mcoker: The `th` doesn't align with the content now. Is that intentional?

screen shot 2019-03-01 at 4 10 02 pm

mcoker: Per https://github.com/patternfly/patternfly-org/pull/732, looks like the content should align with the header.

src/patternfly/components/Table/table.scss
christiemolloy: Yes this looks good now!

src/patternfly/components/Table/examples/table-compact-example.hbs
christiemolloy: Fixed this now :)

PR #1473, opened 2019-02-25
fix(switch): remove label from css

### LINKED ISSUE #1452: Switch enhancements - move text strings from css to html
Currently the text strings that are included in the Switch component are defined in the css. This can be problematic for consumers that want to either translate their text strings or customize these strings to be different (e.g. enabled/disabled vs on/off).

I copied bits from comments in issue #915 to show a solution that might work for this.

With the html shown below, a couple of things are happening:

- The checkbox is always labelled by the positive label, so that what gets announced makes sense. E.g. "On, checkbox, checked" or "Enabled, checkbox, checked"
- The text inside `.pf-c-switch__label is-on` gets announced when the checkbox receives focus because of the `aria-labelledby` on the checkbox. But otherwise, the visible labels are ignored by the screen reader due to `aria-hidden`.
- The combination of the classes used on these labels plus the CSS shown below will control which label is visible depending on the state of the checkbox.

HTML
```
<label class="pf-c-switch" for="switch-on-example">
  <input 
    class="pf-c-switch__input" 
    type="checkbox" 
    id="switch-on-example" 
    aria-labelledby="switch-label" 
    checked
  >
  <span class="pf-c-switch__toggle">
  </span>
  <span class="pf-c-switch__label is-off" aria-hidden="true">
      Message when off
  </span>
  <span class="pf-c-switch__label is-on" aria-hidden="true" id="switch-label">
      Message when on
  </span>
</label>
```

CSS
```
.pf-c-switch__input:not(:checked) ~ .is-on,
.pf-c-switch__input:checked ~ .is-off {
  display: none;
}
```

christiemolloy:
Fix #1452 

Fixes what we discussed in CSS Army. The switch label is not written in the CSS anymore, it is in the html. I feel like there are many variations on how to do this, so if you have any suggestions let me know :)

### DIFF
--- a/src/patternfly/components/Switch/docs/code.md
+++ b/src/patternfly/components/Switch/docs/code.md
@@ -13,6 +13,7 @@ Use checkbox if your user has to perform additional steps for changes to become 
 | `aria-labelledby="..."` or `aria-label="..."` | `.pf-c-switch__input` | Indicates the action triggered by the switch. If an additional text label is included with the switch besides `.pf-c-switch__label.pf-m-on`, then `aria-labelledby` can reference the `id` of this text, or this text can be used as the value for `aria-label`. If the text included for `.pf-c-switch__label.pf-m-on` provides additional meaning to the primary label that's referenced, then it can also be represented as part of the `aria-labelledby` or `aria-label` attribute. **Required** |
 | `for` | `<label>` | Each `<label>` must have a `for` attribute that matches its input id. **Required** |
 | `id` | `<input type="checkbox">` | Each `<input>` must have an `id` attribute that matches its label's `for` value. **Required** |
+| `id` | `.pf-c-switch__label` | Each `.pf-c-switch__label` must have an `id` attribute that matches the `aria-labelledby` on `.pf-c-switch__input`. |
 | `checked` | `.pf-c-switch__input` |  Indicates that the input is checked |
 | `disabled` | `.pf-c-switch__input` |  Indicates that the input is disabled |
 | `aria-hidden="true"` | `.pf-c-switch__label` | Hides the text from the screen reader. |
--- a/src/patternfly/components/Switch/docs/code.md
+++ b/src/patternfly/components/Switch/docs/code.md
@@ -15,7 +15,6 @@ Use checkbox if your user has to perform additional steps for changes to become 
 | `id` | `<input type="checkbox">` | Each `<input>` must have an `id` attribute that matches its label's `for` value. **Required** |
 | `checked` | `.pf-c-switch__input` |  Indicates that the input is checked |
 | `disabled` | `.pf-c-switch__input` |  Indicates that the input is disabled |
-| `role=presentational` | `.pf-c-switch__label` | Hides the generated content from the screen reader. **Required** |
 | `aria-hidden="true"` | `.pf-c-switch__label` | Hides the text from the screen reader. |
 
 ## Usage
@@ -27,3 +26,5 @@ Use checkbox if your user has to perform additional steps for changes to become 
 | `.pf-c-switch__toggle` | `<span>` |  Initiates the toggle inside the switch. **required**  |
 | `.pf-c-switch__toggle-icon` | `<i>` | Initiates an icon inside the switch toggle. **required when the switch is used without a label** |
 | `.pf-c-switch__label` | `<span>` |  Initiates a label inside the switch. |
+| `.pf-m-on` | `.pf-c-switch__label` | Modifies the switch label to display the on message. |
+| `.pf-m-off` | `.pf-c-switch__label` | Modifies the switch label to display the off message. |
--- a/src/patternfly/components/Switch/docs/code.md
+++ b/src/patternfly/components/Switch/docs/code.md
@@ -10,7 +10,7 @@ Use checkbox if your user has to perform additional steps for changes to become 
 
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `aria-labelledby="..."` | `.pf-c-switch__input` |  Indicates the action triggered by the switch. **required**  |
+| `aria-labelledby="..."` or `aria-label="..."` | `.pf-c-switch__input` | Indicates the action triggered by the switch. If an additional text label is included with the switch besides `.pf-c-switch__label.pf-m-on`, then `aria-labelledby` can reference the `id` of this text, or this text can be used as the value for `aria-label`. If the text included for `.pf-c-switch__label.pf-m-on` provides additional meaning to the primary label that's referenced, then it can also be represented as part of the `aria-labelledby` or `aria-label` attribute. **Required** |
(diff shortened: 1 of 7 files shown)


jgiardino:
This looks great! It works as I would expect it to. There's just one thing missing...

In this element, you include `aria-labelledby`, but the `id` with the same value is missing:
`<input class="pf-c-switch__input" type="checkbox" id="on" aria-labelledby="switch-is-on" name="switchExample3" checked>`

That `id` should be included on this element:
`<span class="pf-c-switch__label pf-m-on" role="presentational" aria-hidden="true" id="switch-is-on">`

jgiardino:
As a followup to my previous comment, if the switch has no visible label, then instead of `aria-labelledby` on the `<input>`, you would just use something like `aria-label="Run security audit"` for the `<input>`.

In actual usage, I would expect an additional text label to be included with most switches, that describe what you're actually controlling. For example "Run security audit  [switch] On". In this case, "Run security audit" is the more descriptive text string to use and would be more appropriate than "on" for the aria-label. 

I don't think we should update the examples to show this extra string, since it isn't really part of the component, but it would probably be helpful to include info about the aria-label/aria-labelledby attributes in the documentation that describes this.

Here's a stab at this. Feel free to use whatever makes sense. 


-- | -- | --
`aria-labelledby="..."` or `aria-label="..."`  | .pf-c-switch__input | Indicates the action triggered by the switch. If an additional text label is included with the switch besides `.pf-c-switch__label.pf-m-on`, then `aria-labelledby` can reference the `id` of this text, or this text can be used as the value for `aria-label`. If the text included for `.pf-c-switch__label.pf-m-on` provides additional meaning to the primary label that's referenced, then it can also be represented as part of the `aria-labelledby` or `aria-label` attribute. **required**

jgiardino:
Would it be out of scope to remove `role="presentational"` from the `<span>` elements in this component? The [`presentation` role](https://www.w3.org/TR/wai-aria-1.1/#presentation) is intended to remove the implicit role associated with an element (e.g. if you're using a `<li>` but more for presentation/html and not for the "listitem" role). In this case, it's not needed because the [`<span>` element has no implicit role](https://www.w3.org/TR/html-aria/#span).

christiemolloy:
Thanks @jgiardino . Made those changes for now until it is tested on Jaws. Also removed `role=presentational` - @matthewcarleton had discovered that this hid the "content" in the CSS from screen readers. But now that we aren't going with that approach we can definitely remove.

jgiardino:
Oh, that's interesting to know about the `role="presentation"`. 

I tested different variations of labeling the switch in different screen readers.

VO works best when aria-label/aria-labelledby is on the `<input>`. JAWS and NVDA didn't change how the control is announced when I placed this attribute on `<label>` vs `<input>`.

christiemolloy:
@jgiardino updated this. aria-label and aria-labelledby are on the `<input>`

Let me know how it looks now :)

jgiardino:
It looks good. The one thing that's off is the values don't match between `aria-labelledby` on the `<input>` and `id` on the `.pf-c-switch__label.pf-m-on`.
```
<label class="pf-c-switch" for="switch-with-label-1">
  <input class="pf-c-switch__input" type="checkbox" id="switch-with-label-1" aria-labelledby="switch-with-label-1-on" name="switchExample1" checked>
  <span class="pf-c-switch__toggle">
  </span>
  <span class="pf-c-switch__label pf-m-on" id="on" aria-hidden="true">
```

christiemolloy:
@jgiardino yes was actually just fixing that! Updated.

dgutride:
@tlabaj  - can you take a look at this to see if we'd need to introducing a breaking change to react-core?  If so, we'll need to hold off on merging this for a little while.

tlabaj:
@dana it looks like we ideally we would want to change our interface.  There is actually a bug in the current code.   We are taking in a label prop and applying it and css is also applying a label.  So in the Docs we see a double label (e.g. "OnOn").  For the change in this PR, we would want to take in 2 labels.  One for On and one for Off.  Or, we could possible change the current label prop to take in an object or a string.  The object would be the labels for on and off.   If the prop is a string we would could apply it to both the on/off spans (one would be hidden).  The consumer would be responsible for passing in the correct label as they do today.

jgiardino:
I could totally be misunderstanding something, but I'm not sure this PR would require breaking changes in the react component. The [react Switch component](https://patternfly-react.surge.sh/patternfly-4/components/switch) currently has a label prop which accepts both labels for the switch as shown below. The issue we're seeing in react is that in a previous PR, the labels On/Off were included in the CSS, but this PR would revert that change.

image

tlabaj:
@jgiardino the consumer determines the label here based on the boolean isChecked..  The line of code 
`label={isChecked ? 'Yes' : 'No'}`
passes in 'Yes' or No as a the label prop.  Yes if isChecked is true, No is it is false. SO if we want to keep the interface the same, we would need to accept a string (like we do now), and apply it to both spans since the component will not know which span the label is for.  Based on the is checked value, the appropriate span will be hidden.  Or, we can accept a string or an object as a prop.  The object would have both the On and off labels and we could apply the correct label to the correct span.

jgiardino:
Ah, ok. Thanks for the clarification, @tlabaj! So if I understand correctly, this PR is a breaking change for the react component, but the updates to the react component would not be a breaking change for the consumer if we continue to use the `label` prop as-is?

tlabaj:
@jgiardino Correct!

jgiardino:
If we can get this fix in, then we can update the react component to display text strings defined by the user. I think we can avoid having this PR be a breaking change for core consumers if we add the css back in that displays the labels, but then for the newer example that this PR introduces, we can modify the css so that it doesn't pull the labels in from the css.

For example, this html from before this PR would still display the labels from the CSS:
```
<label class="pf-c-switch" for="on">
  <input class="pf-c-switch__input" type="checkbox" id="on" aria-labelledby="switch-is-on" name="switchExample2" checked="">
  <span class="pf-c-switch__toggle">
  </span>
  <span class="pf-c-switch__label" role="presentational" aria-hidden="true">
  </span>
</label>
```
But if this newer html is being used, then the labels would not be pulled in from the css, and we would use the labels as defined in the html.
```
<label class="pf-c-switch" for="switch-with-label-1">
   <input class="pf-c-switch__input" type="checkbox" id="switch-with-label-1" aria-labelledby="switch-with-label-1-on" name="switchExample1" checked="">
  <span class="pf-c-switch__toggle">
  </span>
  <span class="pf-c-switch__label pf-m-on" id="switch-with-label-1-on" aria-hidden="true">
    Message when on</span>
  <span class="pf-c-switch__label pf-m-off" id="switch-with-label-1-off" aria-hidden="true">
    Message when off</span>
</label>
```

I don't think we would include an example of the deprecated html on our page, but if we do, we should label it as deprecated. 

Later when we are ready for a breaking change release, we can remove that css that enables the older, deprecated html to still work.

jgiardino:
Thanks, @mcoker! I pulled the `breaking change` label off. This looks good to me. I tested the previous version of the HTML and it still works.

PR #1475, opened 2019-02-25
fix(emptyState): add sm and lg modifiers for max-width

### LINKED ISSUE #1098: Empty state doesn't stretch to fill container
The empty state currently has a max-width on it. Our convention is that a component fills available space, and the parent container is responsible for sizing and placement. I recommend removing the max-width. A modifier could be used to limit the width if absolutely necessary.

@mcoker @mceledonia what do you think?

christiemolloy:
Fix #1098 

Empty state is 100% full width without modifiers. With `.pf-m-sm` its max-width is 300px. With `.pf-m-lg` its max-width is 600px. I'm open to suggestions for the sizing @mceledonia @kybaker


mcoker:
@christiemolloy looks like this branch has a conflict.

kybaker:
@christiemolloy Those are good sizes to start with. Thank you for making the change.

PR #1477, opened 2019-02-25
fix(accordion, data-list, page): add padding to child of expandable parent to fix overflow bug.

### LINKED ISSUE #1405: Firefox padding issue with elements that overflow
On an element that has overflow that scrolls, any bottom padding will be ignored.

Read more about it here - https://github.com/w3c/csswg-drafts/issues/129

Some components we should look at
* expanded content in data list
* ^ in table
* page component vertical navigation (`.pf-c-page__sidebar`)

christiemolloy:
Fix #1405 

In Firefox on an element that has overflow that scrolls, any bottom padding will be ignored.

When the parent has the `max-height` and `overflow` properties, the child must have the padding. 

I updated: accordion (fixed example), data-list (expandable example) and page (side bar). Data table works in Firefox.


mcoker:
An issue with this approach is that we're adding HTML elements to support scrolling, and in the case of the accordion, for example, every direct child of the accordion expandable element gets the padding. What if there were multiple `p`'s or other children in the accordion? 

I fixed this in the wizard nav by adding an `::after` pseudo element to the element with the `overflow` that is used in place of `padding-bottom` that doesn't require extra HTML or cause any issues with children you might add, like in the accordion example. 

There may be a better way to address this, too. There is a GH issue that talks about the problem https://github.com/w3c/csswg-drafts/issues/129 and it links to a couple of good SO questions with workarounds, one of which is the pseudo element trick I used.

https://stackoverflow.com/questions/10722367/bottom-padding-not-working-on-overflow-element-in-non-chrome-browsers

https://stackoverflow.com/questions/20624938/padding-bottom-is-ignored-in-firefox-and-ie-on-overflowing-elements-with-no-cont

christiemolloy:
@mcoker I saw that fix but thought it was too "hacky". But after looking further into this issue it looks like its a bug and I guess we need to hack it? Do we usually hack > add extra HTML? 

I updated the issue to use the pseudo element anyway :)

mcoker:
@christiemolloy it is hacky :) We can add extra HTML, especially where it makes sense to have an extra element, but I think in cases like this, it's fine. This is a good topic for our CSS meeting tomorrow.

mcoker:
Oh forgot to mention that if we did go the route of adding the padding to a child of the element that has the overflow property, we would need to add a specific class to that element. So instead of

`.pf-c-accordion__expanded-content > * { padding... }`, and then relying on someone putting a single `<p>` in that element to create the padding around the content (because they're likely to put muliple `p`'s or `div`'s or whatever), add a new element (`.pf-c-accordion__expanded-content-body`), and move the padding to that, so that they can put whatever kind of content in it that they want.

christiemolloy:
@mcoker updated this one based on our conversation yesterday.

src/patternfly/components/Accordion/accordion-expanded-content-body.hbs
```diff
@@ -0,0 +1,6 @@
+<p class="pf-c-accordion__expanded-content-body{{#if accordion-expanded-content-body--modifier}} {{accordion-expanded-content-body--modifier}}{{/if}}"
```
mcoker: Should this be a `<p>`? The content model for a paragraph only allows [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content). So they couldn't use a `<div>`, `<h1>`, `<ul>`, etc. I think this would be better as a `<div>`.

src/patternfly/components/DataList/data-list-expandable-content-body.hbs
```diff
@@ -0,0 +1,6 @@
+<p class="pf-c-data-list__expandable-content-body{{#if data-list-expandable-content-body--modifier}} {{data-list-expandable-content-body--modifier}}{{/if}}"
```
mcoker: Same comment here. I don't think we want to limit the elements they can use in this element.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -213,32 +217,30 @@ $pf-data-list--breakpoint--mobile: 380px;
...
-  // reset li padding on expandable content
-  padding-right: var(--pf-c-data-list__expandable-content--PaddingRight);
-  padding-left: var(--pf-c-data-list__expandable-content--PaddingLeft);
+  max-height: var(--pf-c-data-list__expandable-content--MaxHeight);
 
   // offset for li padding for border to extend to li edges
   margin-right: var(--pf-c-data-list__expandable-content--MarginRight);
   margin-left: var(--pf-c-data-list__expandable-content--MarginLeft);
-  overflow: hidden;
+  overflow-y: auto;
```
mcoker: I was curious why you changed the `max-height` since this created an effect of the content scrolling down as the menu opened when you toggle `.pf-m-expanded` on the list item. That said, the react implementation doesn't surface any sort of animation and the `@keyframes` animation makes the overflow weird. I think as part of this PR we can remove the `opacity`, `transition`, `animation` and the `pf-c-data-list--expandable-content--in` `@keyframes` animation from this.

src/patternfly/components/Page/page.scss
```diff
@@ -272,6 +270,12 @@
     box-shadow: var(--pf-c-page__sidebar--BoxShadow);
   }
 
+  .pf-c-page__sidebar-body {
+    height: 100%;
```
mcoker: is this necessary?

christiemolloy: I was mimicking what was there before adding the child element.

**Before:**
screen shot 2019-03-05 at 3 10 21 pm

**After (without height: 100%)**
screen shot 2019-03-05 at 3 09 58 pm

After (with height: 100%)
screen shot 2019-03-05 at 3 10 07 pm

src/patternfly/components/DataList/data-list.scss
christiemolloy: That was actually unintentional. I didn't realize that was for animation :(  I agree with removing this animation though, as we have removed it in almost (I think) every other component.

src/patternfly/components/Page/page.scss
mcoker: Gotcha, the sidebar needs to be full height because it has styles (like `box-shadow`) that rely on the element being full height. But the body element shouldn't, just like how the nav component inside doesn't. Or do you see a reason that we need it?

christiemolloy: Ah I don't see a reason that we need it! So good call!

christiemolloy:
*Note: the mobile breakpoint on the expandable body isn't working...

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -115,11 +115,14 @@
   }
 
   .pf-c-accordion__expanded-content {
-    padding: var(--pf-c-accordion__expanded-content--PaddingTop) var(--pf-c-accordion__expanded-content--PaddingRight) var(--pf-c-accordion__expanded-content--PaddingBottom) var(--pf-c-accordion__expanded-content--PaddingLeft);
     font-size: var(--pf-c-accordion__expanded-content--FontSize);
     color: var(--pf-c-accordion__expanded-content--Color);
     border-left: var(--pf-c-accordion__expanded-content--m-expanded--BorderWidth) solid var(--pf-c-accordion__expanded-content--BorderLeftColor);
 
+    .pf-c-accordion__expanded-content-body {
```
mcoker: can you add this class to the docs?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -85,24 +85,26 @@ $pf-data-list--breakpoint--mobile: 380px;
...
-  --pf-c-data-list__expandable-content--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-data-list__expandable-content--mobile--PaddingRight: var(--pf-global--spacer--md);
-  --pf-c-data-list__expandable-content--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-data-list__expandable-content--mobile--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-data-list__expandable-content--MarginRight: calc(var(--pf-c-data-list__expandable-content--PaddingRight) * -1);
-  --pf-c-data-list__expandable-content--MarginLeft: calc(var(--pf-c-data-list__expandable-content--PaddingLeft) * -1);
+  --pf-c-data-list__expandable-content--MarginRight: calc(var(--pf-c-data-list__expandable-content-body--PaddingRight) * -1);
+  --pf-c-data-list__expandable-content--MarginLeft: calc(var(--pf-c-data-list__expandable-content-body--PaddingLeft) * -1);
   --pf-c-data-list__expandable-content--BoxShadow: var(--pf-global--BoxShadow--md-bottom);
+  --pf-c-data-list__expandable-content--MaxHeight: #{pf-size-prem(600px)};
```
mcoker: do we want a `max-height` on this? Should it match the table, where you only get a max-height on mobile?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -222,41 +224,26 @@ $pf-data-list--breakpoint--mobile: 380px;
...
-    max-height: var(--pf-c-data-list__item--m-expanded__expandable-content--MaxHeight);
-    padding-top: var(--pf-c-data-list__item--m-expanded__expandable-content--PaddingTop);
-    padding-bottom: var(--pf-c-data-list__item--m-expanded__expandable-content--PaddingBottom);
     border-top-width: var(--pf-c-data-list__item--m-expanded__expandable-content--BorderTopWidth);
-    opacity: 1;
+  }
 
-    // use animation to control overflow value to prevent flicker if content scrolls
-    animation: var(--pf-c-data-list__item--m-expanded__expandable-content--TransitionDuration) pf-c-data-list--expandable-content--in forwards;
+  .pf-c-data-list__expandable-content-body {
```
mcoker: Can you document both `.pf-c-data-list__expandable-content` and `.pf-c-data-list__expandable-content-body` in the expandable section code?

src/patternfly/components/Page/page.scss
```diff
@@ -67,8 +67,8 @@
   --pf-c-page__sidebar--ZIndex: var(--pf-global--ZIndex--sm);
   --pf-c-page__sidebar--Width: 80%;
   --pf-c-page__sidebar--md--Width: #{pf-size-prem(290px)};
-  --pf-c-page__sidebar--PaddingTop: var(--pf-global--spacer--md);
-  --pf-c-page__sidebar--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-page__sidebar-body--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-page__sidebar-body--PaddingBottom: var(--pf-global--spacer--md);
```
mcoker: can you move these vars to the bottom of the `--pf-c-page__sidebar` var list?

src/patternfly/components/Page/page.scss
```diff
@@ -276,6 +274,11 @@
     box-shadow: var(--pf-c-page__sidebar--BoxShadow);
   }
 
+  .pf-c-page__sidebar-body {
```
mcoker: can you move this selector outside of/underneath `.pf-c-page__sidebar`?

mcoker: And can you add this class to the documentation?

src/patternfly/components/DataList/data-list-expandable-content-body.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-data-list__expandable-content-body{{#if data-list-expandable-content-body--modifier}} {{data-list-expandable-content-body--modifier}}{{/if}}"
```
mcoker: Looks like you need to update the new datalist demos, too. Comment below in the sidebar change about doing this in one place instead of each demo/example.

src/patternfly/components/Page/page-sidebar-body.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-page__sidebar-body{{#if page-sidebar-body--modifier}} {{page-sidebar-body--modifier}}{{/if}}"
```
mcoker: Looks like you need to update the dataTable and dataList demos with this change, too. How about just adding this to `page-sidebar.hbs`, so updating `page-sidebar.hbs` to...


```
<div class="pf-c-page__sidebar{{#if page-sidebar--modifier}} {{page-sidebar--modifier}}{{/if}}" 
  {{#if page-sidebar--attribute}}
    {{{page-sidebar--attribute}}}
  {{/if}}>
  {{#> page-sidebar-body}}
    {{> @partial-block}}
  {{/page-sidebar-body}}
</div>
```

You could do that to the rest of these changes (accordion and datalist), since the `-body` element is required in each of these components.

src/patternfly/components/DataList/data-list.scss
christiemolloy: Thats a great idea

src/patternfly/components/Page/page-sidebar-body.hbs
christiemolloy: I added it to accordion, but not datalist because although its required, we need the option to add a modifier to remove padding.

src/patternfly/components/Page/page.scss
mcoker: @christiemolloy still not seeing `pf-c-page__sidebar-body` in the docs.

mcoker:
Looks like there are a11y errors in the datalist component and demo.

christiemolloy:
@mcoker I just fixed this PR from what you bought up Friday with the expandable content breaking. It was because the data list item padding was set to be md size at an md breakpoint. But the expandable content area was set to be md padding at a sm breakpoint. So I actually wonder if this is something we should address because it seems like an easy break...

PR #1479, opened 2019-02-25
fix(form): replace aria-errormessage with aria-describedby

### LINKED ISSUE #1420: Forms - replace `aria-errormessage` with `aria-describedby`
The attribute `aria-errormessage` is very similar in behavior to `aria-describedby`, but isn't widely supported yet. 

For now, we should only use `aria-describedby` to reference the error message that displays for a form input.

This issue should find all instances of `aria-errormessage` and simply replace the attribute with `aria-describedby` while keeping the same attribute value.

christiemolloy:
fix #1420

### DIFF
--- a/src/patternfly/components/Form/docs/code.md
+++ b/src/patternfly/components/Form/docs/code.md
@@ -7,7 +7,7 @@
 | `required aria-required="true"` | `<input>`, `<select>`, `<textarea>` | Required fields must include these attributes. |
 | `id="{helper_text_id}"` | `.pf-c-form__helper-text` | Form fields with related `.pf-c-form__helper-text` require this attribute. Usage `<p class="pf-c-form__helper-text" id="{helper_text_id}">`.  |
 | `aria-describedby="{helper_text_id}"` | `<input>`, `<select>`, `<textarea>` | Form fields with related `.pf-c-form__helper-text` require this attribute. Usage `<input aria-describedby="{helper_text_id}">`.  |
-| `aria-invalid="true" aria-errormessage="{helper_text_id}"` | `<input>`, `<select>`, `<textarea>` | Where form validation fails, `aria-errormessage` is used instead of `aria-describedby`. `aria-errormessage` and `aria-invalid="true"` are only present when validation fails so this needs to be handled with Javascript. For proper styling of errors `aria-invalid="true"` is required.|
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Form/docs/code.md
```diff
@@ -7,7 +7,7 @@
...
 | `id="{helper_text_id}"` | `.pf-c-form__helper-text` | Form fields with related `.pf-c-form__helper-text` require this attribute. Usage `<p class="pf-c-form__helper-text" id="{helper_text_id}">`.  |
 | `aria-describedby="{helper_text_id}"` | `<input>`, `<select>`, `<textarea>` | Form fields with related `.pf-c-form__helper-text` require this attribute. Usage `<input aria-describedby="{helper_text_id}">`.  |
-| `aria-invalid="true" aria-errormessage="{helper_text_id}"` | `<input>`, `<select>`, `<textarea>` | Where form validation fails, `aria-errormessage` is used instead of `aria-describedby`. `aria-errormessage` and `aria-invalid="true"` are only present when validation fails so this needs to be handled with Javascript. For proper styling of errors `aria-invalid="true"` is required.|
+| `aria-invalid="true" aria-describedby="{helper_text_id}"` | `<input>`, `<select>`, `<textarea>` | Where form validation fails `aria-describedby` is used. `aria-describedby` and `aria-invalid="true"` are only present when validation fails so this needs to be handled with Javascript. For proper styling of errors `aria-invalid="true"` is required.|
```
jgiardino: Parts of this statement seem a little confusing. What are your thoughts on this?

> Where form validation fails `aria-describedby` is used to communicate the error to the user. These attributes need to be handled with Javascript so that `aria-describedby` only references help text that explains the error, and so that `aria-invalid="true"` is only present when validation fails. For proper styling of errors `aria-invalid="true"` is required.

jgiardino:
Thanks for making this update! I just had a suggestion for the documentation. Feel free to revise.

christiemolloy:
Agreed @jgiardino updated it

dgutride:
I'd say it's not a breaking change in core since core only distributes css.  If react adopts this and has to modify attributes, they can just add a new one and mark the old one as deprecated so as not to introduce a breaking change (with an issue to be cleaned up later that would remove the deprecated propType if it exists).

jgiardino:
Just FYI, this update was previously made in the React component (https://github.com/patternfly/patternfly-react/pull/917). A react issue isn't needed for this one.

PR #1480, opened 2019-02-25
fix(a11y): fixes duplicate ids by moving SVG filter to scss

### LINKED ISSUE #1912: id attribute value must be unique
On this path we find an issue that comes from the background image component providing a non-unique ID.

`/components/Login/examples-full/`

https://pf4.patternfly.org/components/Login/examples-full/

Screen Shot 2019-06-10 at 11 14 14 AM

srambach:
Having the background filter SVG inline in the HTML causes duplicate IDs when the A11y checker is run. This PR moves the SVG out of the HTML and into the SCSS for the background image component. 

However, this fix works in Chrome and Firefox, but not in Safari. In Safari, the filter is correctly assigned to the variable, the variable is showing up for the filter property, however, the SVG is unaffected. In the inspector, everything appears to be properly escaped.

Turning the SVG to Base64 did not work for me at all, so I may have been doing it wrong. However, according to CSS-Tricks that should not be necessary.

Some relevant links:
https://stackoverflow.com/questions/10768451/inline-svg-in-css
https://stackoverflow.com/questions/13925736/inline-svg-filter-in-css
https://github.com/Fyrd/caniuse/issues/3803
https://benfrain.com/applying-multiple-svg-filter-effects-defined-in-css-or-html/
https://css-tricks.com/probably-dont-base64-svg/

Useful tools:
SVG minifier: https://jakearchibald.github.io/svgomg/
https://www.mobilefish.com/services/base64/base64.php

Would fix #1912  if it worked in Safari.

### DIFF
--- a/src/patternfly/components/BackgroundImage/background-image.hbs
+++ b/src/patternfly/components/BackgroundImage/background-image.hbs
@@ -3,20 +3,5 @@
     {{{background-image--attribute}}}
   {{/if}}>
   {{> @partial-block}}
-  <svg xmlns="http://www.w3.org/2000/svg" class="pf-c-background-image__filter" width="0" height="0">
-  <filter id="image_overlay" width="">
-    <feColorMatrix type="matrix"
-      values="1 0 0 0 0
-              1 0 0 0 0
-              1 0 0 0 0
-              0 0 0 1 0" />
-    
-    <feComponentTransfer color-interpolation-filters="sRGB" result="duotone">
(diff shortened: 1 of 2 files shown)


mcoker:
I tinkered with this and couldn't get it to work in Safari, either.

mcoker:
@srambach this PR is pretty stale. Any objections to closing it and re-visiting it if someone reports an issue resulting from the duplicate ID's? I'm fine just living with this error in the workspace a11y checks.

srambach:
Yeah, I guess if we can live with the a11y checks then it's fine until someone hits the problem irl.

PR #1482, opened 2019-02-26
fix(pagination): update content at breakpoints

### LINKED ISSUE #1449: Update Pagination and Toolbar (break-points)
Confirmed with @gdoyle1 about the following changes that need to be made to the pagination and toolbar component.

- Currently pagination top on mobile truncates to "10 items", however it should say the full text from the options menu - "1-10 of 20 items".
- pagination bottom should truncate to just `.pf-c-pagination__nav' and the options-menu should be hidden.

For toolbar: the breakpoint on the toolbar when used in a page needs to be tweaked. See data table demo example - https://deploy-preview-1417--pf-next.netlify.com/demos/DataTable/examples/

screen shot 2019-02-20 at 2 58 42 pm

### LINKED ISSUE #1470: Pagination component - update css used to hide elements
The Pagination component currently includes the following element, which only displays in smaller views:
`<div class="pf-c-pagination__total-items">20 items</div>`

Currently when viewing the full size version of the component, this element is hidden using `display: none;`. This property alone doesn't reliably hide elements from screen readers. Whenever hiding elements from screen readers using only css, the following properties should be defined:
```
display: none;
visibility: hidden;
```

These properties should also be applied to the elements that become hidden in the responsive view of this component.

christiemolloy:
fix #1470 , fix #1449


christiemolloy:
@gdoyle1 updated :)

rachael-phillips:
Hi @christiemolloy we're going to hold off on merging this until April 5 because it is a breaking change. For any issue that may require a breaking change we will wait until the release.

src/patternfly/components/Pagination/docs/code.md
```diff
@@ -20,6 +20,7 @@ Note: `<button>` or `<a>` elements can be used in `.pf-c-pagination__nav-page-se
 | Class | Applied To | Outcome |
 | -- | -- | -- |
 | `.pf-c-pagination` | `<div>` |  Initiates pagination. |
+| `.pf-c-pagination__items` | `<div>` | Initiates pagination items. |
```
mcoker: If you didn't change this element's name, would it still be a breaking change?

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -28,18 +28,31 @@
...
-        display: flex;
+      @media screen and (max-width: $pf-global--breakpoint--md) {
+        display: none;
+        visibility: hidden;
+      }
+    }
+  }
+
+  &.pf-m-footer {
+    margin-left: auto;
```
mcoker: is this necessary? I don't see what it's doing.

christiemolloy: When I put pagination in the toolbar (table and list demo)  it defaults to a left alignment, so this fixes that.

src/patternfly/components/Pagination/docs/code.md
christiemolloy: I meant to remove this class.

src/patternfly/components/Pagination/pagination.scss
mcoker: Shouldn't the toolbar be responsible for positioning it's elements?

christiemolloy: Yes updated

christiemolloy:
Hey @mcarrano,  @gdoyle1 and I have been discussing whether or not the top pagination on mobile should shrink down to just a text string or the options menu. If we choose the first option (like we have right now), we won't have options menu at all on mobile, as the bottom pagination now shrinks to `pagination__nav`

mcarrano:
@christiemolloy @gdoyle1 I think that the issue is one of space. I was willing to loose functionality to save space in the toolbar. Can you show what this would look like? I'm having a hard time getting this by looking at the demo. This is what I'm seeing in responsive view for an iPhone: 

screen shot 2019-02-28 at 12 01 13 pm

christiemolloy:
Sorry @mcarrano I had to break the demos into individual PR's so we could capture any breaking changes. But with this PR pulled into the demo it would look like this:

screen shot 2019-02-28 at 2 42 19 pm

christiemolloy:
@mcarrano right now I think im leaning towards removing the options menu on mobile and keeping just the string of text. This will avoid breaking changes and save space in the toolbar.

I had asked @gdoyle1 if it was okay to opt for saving space over functionality so that answers the question :)

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -28,7 +28,7 @@
     white-space: nowrap;
   }
 
-  > *:not(:last-child):not(.pf-c-pagination__total-items) {
+  > *:not(:last-child):not(.pf-c-pagination__items) {
```
mcoker: I think you want to change this back? Currently there is a margin on `.pf-c-pagination__total-items` and `.pf-c-pagination__items` doesn't exist

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -133,6 +133,7 @@
 }
 
 .pf-c-toolbar__pagination,
+.pf-c-pagination,
```
mcoker: why do we need 2 selectors for the toolbar pagination?

src/patternfly/components/Pagination/pagination.scss
christiemolloy: Oh yessss i forgot to change this back!! thanks for catching that

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -132,7 +132,6 @@
   --pf-c-toolbar--child--MarginLeft: var(--pf-c-toolbar__action-list--MarginLeft);
 }
 
-.pf-c-toolbar__pagination,
```
christiemolloy: this will make it breaking now @mcoker

mcoker: will it though? the current toolbar top pagination is `.pf-c-pagination.pf-c-toolbar__pagination`. Seems like you could remove that selector and the `.pf-c-pagination` selector will still position the pagination component.

christiemolloy: Oh yeah the functionality is there so it should be good! Didn't see those classes, removed them.

mcoker: Ah... removing the classes from the elements will be a breaking change. If someone was using them to style the pagination, they're no longer available. I was just referring to removing `.pf-c-toolbar__pagination` from the CSS that applies `margin-left: auto` and only using `.pf-c-pagination` there. That shouldn't introduce a breaking change. We could create a follow-up issue to remove `.pf-c-toolbar_pagination` from those elements later though.

christiemolloy: Ohh ok so ill go back to what I had and then create a follow up issue

PR #1497, opened 2019-02-27
fix(page): add modifier to page main section

### LINKED ISSUE #1490: Add modifier class to `.pf-c-page__main-section`
Add a modifier class to `.pf-c-page__main-section` that removes padding from the main section, to allow cards to sit on the edges of the viewport. This is necessary for the data table and data list demos.

christiemolloy:
Fix #1490

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -90,10 +90,10 @@
   --pf-c-page__main-section--md--PaddingRight: var(--pf-global--spacer--xl);
   --pf-c-page__main-section--md--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-page__main-section--md--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-page__main-section--m-nopadding--PaddingTop: 0;
-  --pf-c-page__main-section--m-nopadding--PaddingRight: 0;
-  --pf-c-page__main-section--m-nopadding--PaddingBottom: 0;
-  --pf-c-page__main-section--m-nopadding--PaddingLeft: 0;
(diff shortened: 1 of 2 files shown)


mcoker:
can you change this to `pf-m-no-padding`?

christiemolloy:
Updated @mcoker

mcoker:
Thanks! What's the use for the media query? Is there only supposed to be no padding on mobile? In that case, I don't know that `no-padding` is a good modifier name since I would expect that to remove the padding regardless the viewport width.

christiemolloy:
@mcoker yes no padding is only for mobile, I accidentally left that out of this PR. What about `.pf-m-no-padding-mobile'

mcoker:
@christiemolloy I'm ok with that. @srambach @matthewcarleton wdyt?

christiemolloy:
@mcoker I went ahead and made the name change - I think it makes sense.

src/patternfly/components/Page/page.scss
```diff
@@ -90,6 +90,10 @@
   --pf-c-page__main-section--md--PaddingRight: var(--pf-global--spacer--xl);
   --pf-c-page__main-section--md--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-page__main-section--md--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__main-section--md--m-no-padding-mobile--PaddingTop: 0;
```
mcoker: This looks great. Can you change the vars so that `--md` is the last thing before the property? The breakpoint should go last (assuming there aren't properties defined after the breakpoint in the SCSS). The list of things in the var should also generally follow the nesting in SCSS.

PR #1503, opened 2019-02-27
fix(table): change padding

### LINKED ISSUE #1499: Update table padding-left/padding-right for mobile
Spoke with @kybaker ... table rows should be 16px for padding rather than 32px. (on mobile)

christiemolloy:
fix #1499


mcoker:
This changes padding-top/bottom, too, is that intentional?

christiemolloy:
@mcoker yes ill update the issue

PR #1504, opened 2019-02-27
fix(wizard): adjust padding, var names

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1501

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -76,22 +76,26 @@
   --pf-c-wizard__nav--lg--BoxShadow: var(--pf-global--BoxShadow--lg-right);
   --pf-c-wizard__nav--Width: 100%;
   --pf-c-wizard__nav--lg--Width: #{pf-size-prem(300px)};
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -422,12 +421,13 @@
 }
 
 .pf-c-wizard__main {
+  padding: var(--pf-c-wizard__main--PaddingTop) var(--pf-c-wizard__main--PaddingRight) var(--pf-c-wizard__main--PaddingLeft) var(--pf-c-wizard__main--PaddingRight);
```
christiemolloy: Should the last two variables be "PaddingBottom" and "PaddingLeft"

mcoker:
@christiemolloy doh! thanks, updated. It is technically a breaking change, but the component was just added this week, I wonder if it would be OK. @jschuler is working on the pf-react component, and I would like to see these changes there. @dgutride what do you think?

dgutride:
I'd say it's still ok @mcoker - nobody is using it yet

srambach:
LGTM.

PR #1506, opened 2019-02-27
fix(datalist): for mobile styling

### LINKED ISSUE #1505: Changes to Data List for Mobile
This issue is for changes to the data list from the data list demo.

christiemolloy:
fix #1505

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -2,7 +2,7 @@ $pf-data-list--breakpoint--mobile: 380px;
 
 .pf-c-data-list {
   --pf-c-data-list--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
-  --pf-c-data-list--BorderTopColor: var(--pf-global--Color--light-200);
+  --pf-c-data-list--BorderTopColor: var(--pf-global--BorderColor--light-200);
   --pf-c-data-list--BorderTopWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-data-list--mobile--BorderTopWidth: var(--pf-global--spacer--sm);
   --pf-c-data-list--BorderBottomColor: var(--pf-global--Color--light-200);
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -44,7 +44,7 @@ $pf-data-list--breakpoint--mobile: 380px;
   --pf-c-data-list__cell-cell--md--PaddingTop: var(--pf-global--spacer--xl);
   --pf-c-data-list__cell-cell--md--MarginLeft: var(--pf-global--spacer--xl);
 
-  @media screen and (min-width: $pf-data-list--breakpoint--mobile) {
+  @media screen and (min-width: $pf-global--breakpoint--md) {
     --pf-c-data-list__cell-cell--PaddingTop: var(--pf-c-data-list__cell-cell--md--PaddingTop);
   }
 
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -20,7 +20,7 @@ $pf-data-list--breakpoint--mobile: 380px;
   --pf-c-data-list__item--m-expanded--before--BackgroundColor: var(--pf-global--active-color--100);
   --pf-c-data-list__item-item--BorderTopColor: var(--pf-global--Color--light-200);
   --pf-c-data-list__item-item--BorderTopWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-data-list__item-item--mobile--BorderTopWidth: var(--pf-global--spacer--sm);
+  --pf-c-data-list__item-item--mobile--BorderTopWidth: calc(var(--pf-global--BorderWidth--xl) * 2);
 
   @media screen and (max-width: $pf-data-list--breakpoint--mobile) {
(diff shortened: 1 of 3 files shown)


src/patternfly/components/DataList/data-list.scss
```diff
@@ -2,15 +2,25 @@ $pf-data-list--breakpoint--mobile: 380px;
 
 .pf-c-data-list {
   --pf-c-data-list--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+  --pf-c-data-list--BorderTopColor: var(--pf-global--Color--light-200);
```
mcoker: This is a `color` var. Can you use `--pf-global--BorderColor--light-200` instead?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -34,7 +44,7 @@ $pf-data-list--breakpoint--mobile: 380px;
   --pf-c-data-list__cell-cell--md--PaddingTop: var(--pf-global--spacer--xl);
   --pf-c-data-list__cell-cell--md--MarginLeft: var(--pf-global--spacer--xl);
 
-  @media screen and (min-width: $pf-global--breakpoint--md) {
+  @media screen and (min-width: $pf-data-list--breakpoint--mobile) {
```
mcoker: Shouldn't this match the breakpoint on the page content and pagination components where the left/right padding changes?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -2,15 +2,25 @@ $pf-data-list--breakpoint--mobile: 380px;
 
 .pf-c-data-list {
   --pf-c-data-list--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+  --pf-c-data-list--BorderTopColor: var(--pf-global--Color--light-200);
+  --pf-c-data-list--BorderTopWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-data-list--mobile--BorderTopWidth: var(--pf-global--spacer--sm);
+  --pf-c-data-list--BorderBottomColor: var(--pf-global--Color--light-200);
+  --pf-c-data-list--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
```
mcoker: The top/bottom border of expanded rows doesn't extend to the top and bottom of the component because of the border additions. I added a PR with one way to fix if you're interested - https://github.com/christiemolloy/patternfly-next/pull/10

src/patternfly/components/DataList/data-list.scss
christiemolloy: Yes it should, I had changed it to use `$pf-global--breakpoint--md` to match pagination/page breakpoints but I didn't want to make this PR a breaking change.. what are your thoughts?

src/patternfly/components/DataList/data-list.scss
mcoker: Agreed, you definitely don't want it to overlap on mobile, but on desktop, I think it looks better if the blue is flush with the top border (and the last expanded row is flush with the bottom border).

It's also worth noting that in this PR, the non-first rows blue border overlaps the border above it

Screen Shot 2019-03-08 at 5 02 25 PM

And the case you brought up re: mobile is an issue there.

Screen Shot 2019-03-08 at 5 02 50 PM

I'm in favor of the blue border overlapping the borders on desktop (like they do now), and updating what's there now so they don't overlap on mobile. Or we could have them not overlap anywhere. It might be worth asking @mceledonia, too.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -2,15 +2,25 @@ $pf-data-list--breakpoint--mobile: 380px;
...
   --pf-c-data-list__item--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-data-list__item--mobile--PaddingRight: var(--pf-global--spacer--lg); // custom name for one-off breakpoint
+  --pf-c-data-list__item--mobile--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-data-list__item--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-data-list__item--mobile--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-data-list__item--mobile--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-data-list__item--m-expanded--before--BackgroundColor: var(--pf-global--active-color--100);
   --pf-c-data-list__item-item--BorderTopColor: var(--pf-global--Color--light-200);
   --pf-c-data-list__item-item--BorderTopWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-data-list__item-item--mobile--BorderTopWidth: var(--pf-global--spacer--sm);
```
mcoker: I wonder if we should add a new `--pf-global--BorderWidth` var for this? Or is this just a one-off? If it's just a one-off, is this supposed to present itself as an 8px border, or more of an 8px gap that we would expect to change if someone decided to change `--pf-global--spacer--sm` here? I'm wondering if this should just be `#{pf-size-prem(8px)}` instead.

christiemolloy: I was thinking about this. Its a border and its used multiple times. I just added an xl border width variable. My only concern is that the variables increment by 1 each time:  sm - 1px, md - 2px, lg - 3px and now xl would be 8px which is a jump. How do you feel about making the xl - 4px and then I can calc and multiply it by 2. @mcoker

christiemolloy: after discussing this further we will be going with 8px hard coded height, because this is a one-off .. made those adjustments.

PR #1512, opened 2019-02-28
feat(dropdown): add support for disabled dropdown menus

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1511


src/patternfly/components/Dropdown/docs/code.md
```diff
@@ -40,3 +41,4 @@ The dropdown menu can contain either links or buttons, depending on the expected
 | `.pf-m-plain` | `.pf-c-dropdown__toggle` | Modifies to display the toggle with no border |
 | `.pf-m-active` | `.pf-c-dropdown__toggle` | Forces display of the active state of the element. This state is primarily for demonstration purposes and would not normally be used in lieu of the `:active` pseudo-class. |
 | `.pf-m-disabled` | `a.pf-c-dropdown__menu-item` | Modifies to display the menu item as disabled. This applies to `a.pf-c-dropdown__menu-item` and should not be used in lieu of the `disabled` attribute on `button.pf-c-dropdown__menu-item`|
+| `.pf-m-disabled` | `div.pf-c-dropdown__toggle.pf-m-split-button` | Modifies to display the dropdown toggle as disabled. This applies to `div.pf-c-dropdown__toggle.pf-m-split-button` and should not be used in lieu of the `disabled` attribute on `button.pf-c-dropdown__toggle`, `.pf-c-dropdown__toggle-button`, or `.pf-c-dropdown__toggle-check > input[type="checkbox"]`|
```
christiemolloy: To me this seems like `.pf-m-disabled` can only apply to the dropdown with `.pf-m-split-button`. But can it be applied to every dropdown no matter what modifier is applied, if any modifier at all?

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -95,6 +97,19 @@
...
 
+  &.pf-m-disabled,
+  &:disabled {
+    pointer-events: none;
+
+    &:not(.pf-m-plain) {
+      --pf-c-dropdown__toggle--BackgroundColor: var(--pf-c-dropdown__toggle--disabled--BackgroundColor);
+
+      &::before {
+        border-color: var(--pf-c-dropdown__toggle--disabled--BackgroundColor);
```
christiemolloy: could you just do border: 0; to unset the border since technically it doesn't have a border?

src/patternfly/components/Dropdown/examples/dropdown-disabled-example.hbs
```diff
@@ -0,0 +1,5 @@
+{{#> dropdown id="dropdown-example-collapsed" dropdown--IsActionMenu="true" dropdown--HasToggleIcon="true" dropdown-toggle--IsDisabled="true"}}
```
christiemolloy: Should you update the id to `id="dropdown-disabled-example-collapsed` as it currently shares the same id as the example above and they may cause a11y issues.

src/patternfly/components/Dropdown/docs/code.md
mcoker: You would only use it on `div.pf-c-dropdown__toggle` (not a button, you should use `disabled` instead), when you want the grey background color to appear - so not on a `.pf-m-plain` dropdown. Currently the only case for that in the dropdown is the split button toggle.

src/patternfly/components/Dropdown/dropdown.scss
mcoker: You can, I was following what the button component does, but now that i look at it again, the button uses `transparent` as the color. 

Seems like I should either use `border: 0;` or set `--pf-c-dropdown__toggle--BorderTopColor`, `--pf-c-dropdown__toggle--BorderRightColor`, etc to `transparent`, which allows a person to use a color if they want to. What do you think?

src/patternfly/components/Dropdown/docs/code.md
mcoker: Though this could probably just be written as `div.pf-c-dropdown__toggle`.

christiemolloy: ^ I think that would be better

src/patternfly/components/Dropdown/dropdown.scss
christiemolloy: AH thats a tricky one. I'm in favor of `border: 0` because the design doesn't have a border so why give people the opportunity to add a border when the design doesn't call for it.

mcoker:
Thanks @christiemolloy updated

src/patternfly/components/Dropdown/docs/code.md
```diff
@@ -40,3 +41,4 @@ The dropdown menu can contain either links or buttons, depending on the expected
 | `.pf-m-plain` | `.pf-c-dropdown__toggle` | Modifies to display the toggle with no border |
 | `.pf-m-active` | `.pf-c-dropdown__toggle` | Forces display of the active state of the element. This state is primarily for demonstration purposes and would not normally be used in lieu of the `:active` pseudo-class. |
 | `.pf-m-disabled` | `a.pf-c-dropdown__menu-item` | Modifies to display the menu item as disabled. This applies to `a.pf-c-dropdown__menu-item` and should not be used in lieu of the `disabled` attribute on `button.pf-c-dropdown__menu-item`|
+| `.pf-m-disabled` | `div.pf-c-dropdown__toggle` | Modifies to display the dropdown toggle as disabled. This applies to `div.pf-c-dropdown__toggle` and should not be used in lieu of the `disabled` attribute on `button.pf-c-dropdown__toggle`. When this is used, `disabled` should also be added to any form elements in `div.pf-c-dropdown__toggle`|
```
jgiardino: I like this description. 👍

mcoker: :D @jgiardino not sure if you had a chance to review the rest of the PR or not, but when you can would you mind leaving a review?

PR #1514, opened 2019-03-01
fix(tabs): hide scrollbars on firefox

### LINKED ISSUE #1103: Scrollbars in Firefox for Tabs (and maybe Wizard)
As of today, Firefox has increased support for scrollbars in terms of color and width. This means that it is possible for us to hide scrollbars across Chrome, Firefox and IE (need to check with Edge). 
This affects the Tabs component and Wizard component.
The scrollbar should be hidden in Chrome and its width set to none in Firefox aligning with what the original design had specified.

- https://drafts.csswg.org/css-scrollbars-1/  
- https://hacks.mozilla.org/2018/12/firefox-64-released/ 

Related issue #859

christiemolloy:
Close #1103 

Spoke to @kybaker and he is pushing for us to explore options to hide the scrollbar in the tab component. 

This is a solution that that Material uses - https://material-components.github.io/material-components-web-catalog/#/component/tabs

Parent has overflow-y:hidden, child has overflow-x: scroll, margin-bottom: -15px to hide the scrollbar.


andybraren:
I dug into this a bit more. 3 things:

1. This `-15px` margin fix only works on devices with a physical mouse attached. On touch devices (including laptops with trackpads) the tabs now look like the below. We might be able to use some [clever media queries](https://css-tricks.com/touch-devices-not-judged-size/) to only apply the negative margin on devices with a mouse attached, but I don't think we need to (read on).



2. The `-15px` that Material's example uses seems to be applied dynamically via JavaScript, and only on Firefox. They do this seemingly because Firefox on Windows' scrollbar is `17px` instead of `15px` so one size wouldn't have fit all.

3. As noted in #1103, since [Firefox 64 supports `scrollbar-width: none;`](https://hacks.mozilla.org/2018/12/firefox-64-released/) we now have the ability to reliably hide scrollbars without fixed negative margins or relying on JavaScript like Material does.

The bigger question here is whether or not we *should* hide scrollbars in Tabs, given the potential a11y implications.

[@matthewcarleton pointed out](https://github.com/patternfly/patternfly-next/issues/1103#issuecomment-448343569) that `scrollbar-width: none;` overrides the user's system-wide accessibility setting to "Always show scroll bars", but the same could be said of the `overflow-x: scroll; overflow-y: hidden;` stuff we'd be using to hide scrollbars in other browsers.

Unfortunately JavaScript doesn't seem to have the ability to detect the user's system-wide scrollbar preference, so we can't re-enable scrollbars for those users. It's an all-or-nothing decision.

Since PatternFly assumes the user has JavaScript enabled and we tend to assume that devs will include the Tab component's scroll buttons to indicate that scrolling is possible (right?), I lean toward being okay with hiding scrollbars for everyone. My uninformed hunch is that users of the "Always show scroll bars" system preference care far more about keeping the [page-long scrollbar](https://css-tricks.com/designing-show-scroll-bars/) along the right-hand side of their browser window visible than they do horizontal scrollbars on tabs.

Any other thoughts on this?

mcoker:
Nice work @andybraren. I don't know if I speak for the majority, but I set scrollbars to always show for things like this. I'm not really interested in the main window vertical scrollbar as it's pretty predictable and easy enough to discover if it's scrollable. Little widgets and windows that scroll aren't always obvious unless I see that content is cut off or I just happen to scroll in an area and discover that it's scrollable. If I don't see a scrollbar on something on a page, I assume the area isn't scrollable. 

That said, the scroll buttons are also an indicator that there is overflow content. I may not know that it's _scrollable_, but I would have an indication I can use the buttons to navigate.

christiemolloy:
scrollbar-width doesn't seem to be very supported unless you have the latest versions of all browsers 
https://css-tricks.com/almanac/properties/s/scrollbar-width/

andybraren:
@christiemolloy IIRC at one point we had an implementation that worked in everything but Firefox, right? I was thinking`scrollbar-width` would just give us the Firefox support that blocked this effort previously.

christiemolloy:
Yes you're right @andybraren I didn't mean to post that there. 

Thanks so much for looking into this further. It appears that Material's solution was not the best solution for an only CSS implementation. There is no way for us to target different browsers and devices for different scrollbar widths - see the list here - https://codepen.io/sambible/post/browser-scrollbar-widths. 

So I think the best solution that works for Chrome, Firefox, Edge, Safari is: (mobile devices have their own scrollbars so no need to hide them). 

Chrome:
> ::-webkit-scrollbar {
>   display: none;
> }

Firefox 64 and up: 
>   scrollbar-width: none; 

Edge: (this auto hides them, when you hover the scrollbars appear
> -ms-overflow-style: -ms-autohiding-scrollbar;

mcoker:
Looks good to me. My only concerns are:

* The scrollbars appear on edge when you hover, which is inconsistent, but I think it's fine.
* No indication that you can scroll secondary tabs isn't ideal, but I think that will be covered in https://github.com/patternfly/patternfly-next/issues/1580
* If you're using a [device that doesn't scroll sideways](https://i.stack.imgur.com/Bx37r.jpg), it might not be intuitive to scroll secondary tabs. I have to shift+scroll in Firefox and Chrome on Win10 to scroll sideways.

christiemolloy:
Closing this issue in favor of #1676 which addresses hiding scrolling.

PR #1516, opened 2019-03-01
fix(inputGroup): background is all white now

### LINKED ISSUE #1255: Should the InputGroup's text area have a background
This is small but I noticed this when I was doing the react review for Input Group, and I think its something that we should be consistent on.

The screenshot attached shows that the components making up the input group vary in their background color (transparent/white). I think we should be consistent here, where possible. And I'm not sure if the text area needs to have a white background, it looks inconsistent next to the button, unless it is for accessibility reasons. 
screen shot 2019-01-17 at 12 55 10 pm

christiemolloy:
Fix #1255

### DIFF
--- a/src/patternfly/components/InputGroup/input-group.scss
+++ b/src/patternfly/components/InputGroup/input-group.scss
@@ -1,4 +1,8 @@
 .pf-c-input-group {
+  // Input Group
+  --pf-c-input-group--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+
+  // Input Group Text
   --pf-c-input-group__text--FontSize: var(--pf-global--FontSize--md);
   --pf-c-input-group__text--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-input-group__text--PaddingLeft: var(--pf-global--spacer--sm);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/InputGroup/input-group.scss
```diff
@@ -1,4 +1,8 @@
 .pf-c-input-group {
+  // Input Group
+  --pf-c-input-group--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
```
mcoker: can you use `var(--pf-global--BackgroundColor--100);` here? `@include pf-t-light;` in this component should ensure that's white in light/dark mode.

mcoker:
@christiemolloy looks good, left a comment. Also, how do we feel about these white corners outside of the elements with a border-radius?

screen shot 2019-03-01 at 10 40 32 am

christiemolloy:
I don't think the background of the input-group should show through. I added the same size border-radius to the input group component and this solves the problem @mcoker

srambach:
Has visuals taken a look at this? @mceledonia  When put into dark mode, the white buttons appear to be a pixel smaller since the border is now not visible. (It's there but you can't really see it.) I'm not sure about whether the blues should be changing as well.

before change: 
image

after this PR:
image

mceledonia:
Yeah I think we are going to want to explore some other options for these button styles in dark mode. I will take a look at this and I'll keep you posted!

christiemolloy:
We can close this PR now, spoke to @mceledonia and I created #1556 to look at the redesign of buttons in dark mode

PR #1521, opened 2019-03-01
fix(multiple components): add visibility:hidden where there is display: none

### LINKED ISSUE #1421: Page component - Nav is visibly hidden but not hidden from AT
Currently, when the page navigation is hidden using CSS, it isn't hidden from assistive technology like screen readers.

Whenever CSS is used to hide elements (instead of using the `hidden` attribute on the HTML element), the following properties should be defined to ensure that the contents are completely hidden from screen readers:
```
visibility: hidden; 
display: none;
```

christiemolloy:
Fix #1421 but also addresses every component where this applies:
- about-modal
- form
- nav
- page
- pagination --- addressed this in #1482 
- progress
- switch
- table
- toolbar
- wizard

Note: `visibility: hidden` has to be set second in the case of breakpoints, but for modifiers use `visibility: visible` to make the element visible again.

### DIFF
--- a/src/patternfly/components/AboutModalBox/about-modal-box.scss
+++ b/src/patternfly/components/AboutModalBox/about-modal-box.scss
@@ -238,15 +238,16 @@
 
 // Hero
 .pf-c-about-modal-box__hero {
-  display: none; // Don't display the hero at the narrowest breakpoint
+  display: block;
+  background-image: var(--pf-c-about-modal-box__hero--sm--BackgroundImage);
+  background-repeat: no-repeat;
+  background-attachment: fixed;
+  background-position: var(--pf-c-about-modal-box__hero--sm--BackgroundPosition);
+  background-size: var(--pf-c-about-modal-box__hero--sm--BackgroundSize);
+  grid-area: hero;
 
-  @media only screen and (min-width: $pf-global--breakpoint--sm) {
-    display: block;
-    background-image: var(--pf-c-about-modal-box__hero--sm--BackgroundImage);
-    background-repeat: no-repeat;
-    background-attachment: fixed;
-    background-position: var(--pf-c-about-modal-box__hero--sm--BackgroundPosition);
-    background-size: var(--pf-c-about-modal-box__hero--sm--BackgroundSize);
-    grid-area: hero;
+  @media only screen and (max-width: $pf-global--breakpoint--sm) {
+    display: none; // Don't display the hero at the narrowest breakpoint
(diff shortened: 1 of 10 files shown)


src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -238,15 +238,16 @@
...
-    grid-area: hero;
+  display: block;
+  background-image: var(--pf-c-about-modal-box__hero--sm--BackgroundImage);
+  background-repeat: no-repeat;
+  background-attachment: fixed;
+  background-position: var(--pf-c-about-modal-box__hero--sm--BackgroundPosition);
+  background-size: var(--pf-c-about-modal-box__hero--sm--BackgroundSize);
+  grid-area: hero;
+
+  @media only screen and (max-width: $pf-global--breakpoint--sm) {
```
mcoker: Curious why you changed the media query? It makes this PR a breaking change and goes against mobile-first CSS.

christiemolloy: Initially, I thought that this would reduce one line of code because I wouldn't need to add the line `visibility: visible` to turn back on the visibility. But thanks for catching that this is a breaking change! 
I also updated the wizard description and toolbar action group which did the same thing @mcoker

src/patternfly/components/Page/page.scss
```diff
@@ -228,6 +228,7 @@
 
     @media screen and (min-width: $pf-global--breakpoint--lg) {
       display: none;
+      visibility: hidden;
```
jgiardino: Looking at this one, I'm a little confused about what's happening. It looks like this class is used to hide the kebab in the example below at larger screen sizes:

image


And at larger screen sizes, it disappears and I see two icons instead. But is there a reason why a similar class isn't defined for these icons? Instead, these icons have `.pf-screen-reader` applied which makes them not visible, but they're still there for screen readers. But aren't the actions associated with these icons included in the kebab (i.e. can we just make them `display:none; visibility:hidden`?
image

Not sure if this is in the scope of this PR or not (and probably not if that change would break our react implementation), but wanted to mention it.

jgiardino: I captured this here https://github.com/patternfly/patternfly-next/issues/1538

christiemolloy: Good pick up! I didn't see that. I wonder if its any other components?

PR #1527, opened 2019-03-04
refactor(select): use focusable elements

### LINKED ISSUE #1280: Select component should use focusable elements
The elements used in the Select component that are interactive should use native html elements so that they can receive keyboard focus.

- In examples where clicking anywhere on the toggle has the same behavior, the whole toggle wrapping element should be a button. For example in the first example, a `<div>` is used for for `.pf-c-select__toggle`. instead of `<button>`.
- In examples where the toggle includes separate interactive elements with distinct behaviors, check that each element uses an element type that is appropriate for the behavior associated with it. For example, if clicking on the left places focus in an input, but clicking on the caret on the right displays a menu, then the caret icon should be wrapped in a button. For example, **Select - Single with Typeahead (Expanded and Selected)** uses`<i class="fas fa-caret-down pf-c-select__toggle-arrow" aria-hidden="true"></i>` without a button element. If this should be clickable, it should be wrapped with a button.

As part of this issue, any updates to the html should be reflected in the documentation.

christiemolloy:
fix #1280 

@jgiardino will you double check that the a11y is okay after refactoring this component?

@mcoker to be consistent I know that the toggle should just use `button` over `.pf-c-button` however just using button wouldn't show the focusable state. Which I tried to debug for a while but no luck. Let me know what you think :)


mcoker:
@christiemolloy the reason the button isn't focusable is because the pseudo element that draws the border has `position: absolute`, which gives it a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). Since the pseudo element is the entire width and height of the button, it effectively covers the button and everything inside of it that doesn't have a stacking context. The reason `.pf-c-button` works is because it has `position: relative`, which gives it a stacking context, allowing it to display on top of the pseudo element. You just need to give `.pf-c-select__toggle-button` `position: relative`

src/patternfly/components/Select/select.scss
```diff
@@ -257,7 +273,10 @@
 
 .pf-c-select__toggle-clear + .pf-c-select__toggle-arrow {
```
mcoker: Does this selector apply anymore? I don't see it in the DOM.

src/patternfly/components/Select/select.scss
```diff
@@ -257,7 +273,10 @@
...
   // This is adjusted so that the clear button can maintain even padding around while not creating too much space between the clear and arrow
-  --pf-c-select__toggle-arrow--MarginLeft: var(--pf-c-select__toggle-arrow--with-clear--MarginLeft);
+
+  .pf-c-select__toggle-arrow {
+    padding-left: -16px;
```
mcoker: you can't have a negative padding :)

src/patternfly/components/Select/select.scss
```diff
@@ -57,9 +57,16 @@
...
   // Toggle Arrow
-  --pf-c-select__toggle-arrow--MarginLeft: var(--pf-global--spacer--md);
+  --pf-c-select__toggle-arrow--MarginLeft: var(--pf-global--spacer--sm);
+  --pf-c-select__toggle-arrow--MarginRight: var(--pf-global--spacer--sm);
   --pf-c-select__toggle-arrow--with-clear--MarginLeft: var(--pf-global--spacer--xs);
 
+  // Toggle Button
+  --pf-c-select__toggle-button--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-select__toggle-button--PaddingLeft: var(--pf-global--spacer--sm);
+  --pf-c-select__toggle-button--MarginLeft: #{pf-size-prem(-8px)};
```
mcoker: I don't think you always want a negative margin here? Looks like you only want it when it's next to the clear button? Otherwise I don't understand why you would want it when it's next to the `input[type="text"]`.

But I don't know that you want to do it that way anyways, it causes a weird overlap of the clear button and toggle arrow.

screen shot 2019-03-04 at 3 48 15 pm
screen shot 2019-03-04 at 3 50 10 pm

I wonder if it's worth asking design if there can be more space between those elements so they don't have to overlap like that. If not, I would just use a smaller padding on the sides the clear and arrow button meet, like how we do in the options menu. 

screen shot 2019-03-04 at 3 50 44 pm

srambach:
I see more space between the toggle arrow and the right side of the toggle. I think if the toggle arrow is assumed to always be there, which I think is true, and the toggle arrow now has its own padding, then the `padding-right` on `.pf-c-select__toggle` could safely be removed. 
image

srambach:
^^ unless of course, the new spacing is the desired one...personally I like it better but that's not my call :)

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -26,6 +26,7 @@
   --pf-c-context-selector__toggle-text--MaxWidth: calc(var(--pf-c-context-selector--Width) - var(--pf-c-context-selector__toggle-icon--MarginLeft) - 30px); // hard coded last variable
```
srambach: Apologies, this isn't new to this PR, but this comment needs to be more descriptive. Why is it hardcoded to 30px?

src/patternfly/components/Select/select.scss
christiemolloy: Ah I forgot to remove these, thanks!

christiemolloy:
@srambach this spacing update was intentional as I checked with the sketch file. Do you think I still should remove padding-right from the toggle?

srambach:
@christiemolloy No, if the spacing is correct, then don't remove the padding-right. It must have been incorrect before.

src/patternfly/components/ContextSelector/context-selector.scss
christiemolloy: Yes updated this thanks for catching that!!

mcoker:
Looks like if you make the select component 100% width (what react does by default), the toggle button doesn't grow. That should grow to fill the width of the component.

screen shot 2019-03-05 at 1 44 13 pm


Also, this is an existing problem, so let me know if you'd like to address this in another issue/PR, but the badge is currently centered in the toggle because of `justify-items: space-between;` on the parent.

screen shot 2019-03-05 at 1 44 24 pm

christiemolloy:
@mcoker made updates to this. Let me know about the focusable elements. I changed the padding on only the clear button which means now it doesn't overlap with the arrow. I'm curious do we care about the different shapes of focusable elements and whether an element is centered inside of its focusable area?

christiemolloy:
I think we should address the issues you mentioned in another PR. I went ahead and created - #1539

src/patternfly/components/Select/select.scss
```diff
@@ -255,11 +275,6 @@
   }
 }
 
-.pf-c-select__toggle-clear + .pf-c-select__toggle-arrow {
-  // This is adjusted so that the clear button can maintain even padding around while not creating too much space between the clear and arrow
-  --pf-c-select__toggle-arrow--MarginLeft: var(--pf-c-select__toggle-arrow--with-clear--MarginLeft);
```
mcoker: if you're removing the selector, you should remove `--pf-c-select__toggle-arrow--with-clear--MarginLeft`, too. Though I wonder if this should now become:

```
.pf-c-select__toggle-clear + .pf-c-select__toggle-button {
       --pf-c-select__toggle-arrow--MarginLeft: var(--pf-c-select__toggle-arrow--with-clear--MarginLeft); 
}
```

But update `--pf-c-select__toggle-arrow--with-clear--MarginLeft` to be 8px (`--pf-global--spacer--sm`)

mcoker: Also, if you leave this original selector, and just write a new one with what I have above, I believe this PR isn't a breaking change. You could create a follow up issue to remove `.pf-c-select__toggle-clear + .pf-c-select__toggle-arrow` with a breaking change release.

mcoker:
Looks like there is quite a bit of spacing between the clear and toggle button now. Has design looked at that? 

Screen Shot 2019-03-08 at 4 49 47 PM

Left a comment about updating that.

src/patternfly/components/Select/select.scss
christiemolloy: Awesome I added it back in and will create an issue to remove it!

christiemolloy:
@mcoker made the change you suggested but going to confirm with design the new spacing

mcoker:
Looking at the typeahead examples, the shape of the toggle button looks off. At a minimum, I think it should extend to the top and bottom of the parent element with a height of 36px, like the "clear" button. Just use standard button spacing to achieve that height - I would make it a `.pf-c-button.pf-m-plain` and change the left/right padding, to match the "clear" button.

The unequal padding also looks odd here, though I'm not sure what to do about it. Maybe we should remove the right padding on the parent and increase the width of the button on the right side so it's flush with the right of the parent.

Screen Shot 2019-03-12 at 3 08 19 PM

And I wonder if we should do the same thing when there is a clear button present and allow the right side of the toggle button to be wider than the left, and increase the width on the left side of the clear button. Currently there is more space between the input + toggle button than there is between input + clear button in the typeahead examples.

I tried to capture those changes in this PR which has 2 commits. First is where the work is. Second is just formatting, plus I refactored the toggle-button element to use `{{#> button}}`

https://github.com/christiemolloy/patternfly-next/pull/11

christiemolloy:
@mcoker thanks for doing that, I definitely like this approach better, and although I still think the focus area looks off, I'm happy that it aligns with that we have in the split button component. This is also something we can probably revisit when we enable the new focus states.

PR #1535, opened 2019-03-06
fix(table): refactor table row borders, workspace styles

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1520

I couldn't get borders on `tr` or `td` to behave reliably, but a pseudo element seems to work everywhere. You'll need to test in Firefox on windows to verify the borders no longer disappear.

### DIFF
--- a/src/patternfly/components/Table/examples/table-simple-example.hbs
+++ b/src/patternfly/components/Table/examples/table-simple-example.hbs
@@ -26,17 +26,17 @@
       {{/table-td}}
       {{#> table-td table-td--data-label="Branches"}}
         10
-      {{/table-td}}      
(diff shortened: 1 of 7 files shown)


suomiy:
nice! 

Btw, is this approach going to work for the left and right border of the row (in the future)? Live example https://1227-pr-patternfly-react-patternfly.surge.sh/patternfly-4/components/inlineedit (you just need to hover on the row)

mcoker:
@suomiy Looking at that preview, it should. This approach uses a pseudo element of the `td` to draw the border. The left/right border of the row in the preview you linked to uses a border on the `td` (`td:first-child, td:last-child`).

PR #1541, opened 2019-03-08
feat(table): add support for borders in compact table

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1540

### DIFF
--- a/src/patternfly/components/Table/docs/table-compact.md
+++ b/src/patternfly/components/Table/docs/table-compact.md
@@ -12,4 +12,3 @@
 | Class | Applied To | Outcome |
 | -- | -- | -- |
 | `.pf-m-compact` | `.pf-c-table` | Modifies for a compact table. |
-| `.pf-m-border-rows` | `.pf-c-table` | Modifies border rows to have a border. |
--- a/src/patternfly/components/Table/docs/table-compact.md
+++ b/src/patternfly/components/Table/docs/table-compact.md
@@ -12,3 +12,4 @@
 | Class | Applied To | Outcome |
 | -- | -- | -- |
 | `.pf-m-compact` | `.pf-c-table` | Modifies for a compact table. |
+| `.pf-m-border-rows` | `.pf-c-table` | Modifies border rows to have a border. |
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -437,6 +437,10 @@
       tr:first-child > * {
         --pf-c-table-cell--PaddingTop: calc(var(--pf-c-table--m-compact-cell--PaddingTop) * 2);
       }
+
+      tr:last-child > * {
+        --pf-c-table-cell--PaddingBottom: calc(var(--pf-c-table--m-compact-cell--PaddingBottom) * 2);
+      }
     }
   }
 
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -433,6 +433,10 @@
     tbody {
       --pf-c-table--BorderWidth: 0;
       --pf-c-table--BorderColor: transparent;
+
+      tr:first-child > * {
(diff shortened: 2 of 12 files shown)


mcoker:
@srambach @mceledonia updated to remove the extra top padding when the compact table has borders.

src/patternfly/components/Table/table.scss
```diff
@@ -432,26 +432,23 @@
 // Compact table
 // ==================================================================
 .pf-c-table.pf-m-compact {
+  &:not(.pf-m-border-rows) {
```
christiemolloy: I think it would be good to show an example for this modifier

src/patternfly/components/Table/docs/table-compact.md
```diff
@@ -12,3 +12,4 @@
 | Class | Applied To | Outcome |
 | -- | -- | -- |
 | `.pf-m-compact` | `.pf-c-table` | Modifies for a compact table. |
+| `.pf-m-border-rows` | `.pf-c-table` | Modifies border rows to have a border. |
```
christiemolloy: I think you should update the description to be more specific -  does this modifier give all rows  a border or can you modify individual rows?
andddd since all other tables have borders can it be applied to them? or should this just apply to `.pf-c-table .pf-c-compact`

src/patternfly/components/Table/table.scss
mcoker: added

src/patternfly/components/Table/docs/table-compact.md
mcoker: since it only applies to `.pf-c-table`, I think it's implied you just get borders on the whole table, not on individual rows. Updated the docs so it only applies to `.pf-c-table.pf-m-compact`

src/patternfly/components/Table/table.scss
```diff
@@ -432,26 +432,23 @@
...
+    tbody {
+      --pf-c-table--BorderWidth: 0;
+      --pf-c-table--BorderColor: transparent;
+
+      tr:first-child > * {
+        --pf-c-table-cell--PaddingTop: calc(var(--pf-c-table--m-compact-cell--PaddingTop) * 2);
```
christiemolloy: What was the reason for this change? Its now making the compact table in the compound expansion look off:

Screen Shot 2019-03-14 at 9 13 10 AM

src/patternfly/components/Table/table.scss
```diff
@@ -432,26 +432,23 @@
...
-  tbody {
-    --pf-c-table--BorderWidth: 0;
-    --pf-c-table--BorderColor: transparent;
-
-    tr:first-child > * {
-      --pf-c-table-cell--PaddingTop: calc(var(--pf-c-table--m-compact-cell--PaddingTop) * 2);
-    }
-
-    tr:last-child > * {
-      --pf-c-table-cell--PaddingBottom: calc(var(--pf-c-table--m-compact-cell--PaddingBottom) * 2);
```
christiemolloy: What was the reason for removing this? I think now when we insert a compact table the spacing will be off like in the compound expansion:

Screen Shot 2019-03-14 at 9 13 10 AM

mcoker: @mceledonia had asked to remove the extra padding on the last row, but maybe he meant only when we apply borders? I added this space back to the default compact table, and am removing it when we have borders (along with the extra top padding on the first row).

Here's what it looks like now. WDYT?

Screen Shot 2019-03-14 at 8 58 26 AM
Screen Shot 2019-03-14 at 8 58 17 AM

mceledonia: @mcoker Much better, you nailed it. Only need that bottom padding removed when we use borders because the borders create a visual height for each row, and not removing it makes the bottom row look awkwardly larger than the rest. But we do want it without the borders because sort of the opposite happens :)

mcoker: Thanks @mceledonia! @christiemolloy updated that.

PR #1545, opened 2019-03-08
fix(table): fix disappearing borders in firefox, workspace styles

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1520

You'll need to test in Firefox on windows to verify the borders no longer disappear. Also updated a bunch of workspace styles so they don't bleed into the examples.

### DIFF
--- a/src/patternfly/components/Table/examples/table-compact-example.hbs
+++ b/src/patternfly/components/Table/examples/table-compact-example.hbs
@@ -126,7 +126,7 @@
         <input type="checkbox" name="checkrow4" aria-labelledby="name4">
       {{/table-td}}
       {{#> table-th table-th--data-label="Contributor Name"}}
-        <span id="name4">Matt Carrano</span>
+        <span id="name4">Sarah Jackson</span>
       {{/table-th}}
       {{#> table-td table-td--data-label="Position"}}
         Interaction Design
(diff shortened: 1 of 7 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -119,9 +119,6 @@
   // Standard table row (non-expandable)
   // exclude expandable rows
   tr:not(.pf-c-table__expandable-row) {
-    // set position relative for absolutely positioned elements in responsive view
```
suomiy: do you know why was this "necessary"?

mcoker: My guess is it was used previously but not anymore. There are no `position: absolute;` items that depend on the `tr` having `relative` positioning.

src/patternfly/components/Table/examples/table-compact-example.hbs
```diff
@@ -120,5 +120,36 @@
         {{/dropdown}}
       {{/table-td}}
     {{/table-tr}}
+
+    {{#> table-tr}}
```
christiemolloy: I know adding these extra rows were used to check the bug but I think you should remove these before the PR is merged.

christiemolloy:
Strange, in this PR i still see those two issues that I had fixed in recent PRs:

Screen Shot 2019-03-13 at 2 27 41 PM
Screen Shot 2019-03-13 at 2 27 46 PM

src/patternfly/components/Table/examples/table-compact-example.hbs
mcoker: What's the harm in leaving them? I left them on purpose so that if this bug crops up again, we should be able to see it in the workspace.

mcoker:
@christiemolloy I rebased, let me know if you still see those issues.

src/patternfly/components/Table/examples/table-compact-example.hbs
christiemolloy: Okay thats fine then!!

christiemolloy:
@mcoker not seeing those issues anymore

PR #1548, opened 2019-03-11
chore(tabs): remove aria-label from examples

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1537

### DIFF
--- a/src/patternfly/components/Tabs/docs/tabs-primary.md
+++ b/src/patternfly/components/Tabs/docs/tabs-primary.md
@@ -3,7 +3,6 @@
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
 | `aria-controls=[ID of tab content element]` | `.pf-c-tabs__button` | Identifies the section controlled by the tab. **Required**       |
-| `aria-label="Local"` | `<nav class="pf-c-tabs">` | Gives the `<nav>` element an accessible name. Note that using `<nav>` with this `aria-label` should only be used for a tab list that is unique on the current page. Cases where the same set of tabs are duplicated in multiple regions on a page (e.g. cards on a dashboard), are less likely to benefit from using the `<nav>` element. **Required when `.pf-c-tabs` is used with `<nav>`**
 
 ### Usage
 
--- a/src/patternfly/components/Tabs/docs/tabs-primary.md
+++ b/src/patternfly/components/Tabs/docs/tabs-primary.md
@@ -3,7 +3,7 @@
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
 | `aria-controls=[ID of tab content element]` | `.pf-c-tabs__button` | Identifies the section controlled by the tab. **Required**       |
-| `aria-label="Local"` | `<nav class="pf-c-tabs">` | Gives the `<nav>` element an accessible name. Note that using `<nav>` with this aria-label should only be used for a tab list that is unique on the current page. Cases where the same set of tabs are duplicated in multiple regions on a page (e.g. cards on a dashboard), are less likely to benefit from using the `<nav>` element.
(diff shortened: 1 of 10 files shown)


src/patternfly/components/Tabs/docs/tabs-primary.md
```diff
@@ -3,7 +3,7 @@
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
 | `aria-controls=[ID of tab content element]` | `.pf-c-tabs__button` | Identifies the section controlled by the tab. **Required**       |
-| `aria-label="Local"` | `<nav class="pf-c-tabs">` | Gives the `<nav>` element an accessible name. Note that using `<nav>` with this aria-label should only be used for a tab list that is unique on the current page. Cases where the same set of tabs are duplicated in multiple regions on a page (e.g. cards on a dashboard), are less likely to benefit from using the `<nav>` element.
+| `aria-label="Local"` | `<nav class="pf-c-tabs">` | Gives the `<nav>` element an accessible name. Note that using `<nav>` with this `aria-label` should only be used for a tab list that is unique on the current page. Cases where the same set of tabs are duplicated in multiple regions on a page (e.g. cards on a dashboard), are less likely to benefit from using the `<nav>` element. **Required when `.pf-c-tabs` is used with `<nav>`**
```
christiemolloy: Would it be worth having an example where tabs use `<nav>` so that we can demonstrate this?

mcoker: I think so, what do you think @jgiardino?

jgiardino: I'm not opposed to including this in an example. I do think that in cases where a component can have more generic vs semantic elements, to have the first example use generic elements, to avoid cases where consumers don't read documentation and just copy the first example to paste without fully understanding what's what.

What are your thoughts on including it on both **Scroll buttons** and **Primary tabs with secondary tabs**? In the case of **Primary tabs with secondary tabs**, we would need to have a different label for the secondary set of tabs. My best suggestion is keep "Local" for primary and use "Local secondary" for secondary tabs (i.e. not all primary tabs would have a secondary set of tabs). @stacymcauliffe might have a better suggestion for this. And it might also be good to call out the different `aria-label` in the documentation for `.pf-c-tabs.pf-m-tabs-secondary`. 

If any of this seems out of scope for this PR, we can capture it in a separate issue.

mcoker: @jgiardino That's fine by me. Curious what he point of having the Scroll button example would be? Wouldn't a primary + secondary illustrate the use of `<nav>` with the `aria-label` that you would use if you only had primary, as well as the `aria-label` if you had secondary? Would the scroll button add any other value? Maybe just illustrating that `<nav>` can be used with and without scroll buttons?

mcoker: I went ahead and added a new example with both scroll and primary+secondary tabs. What do you think?

jgiardino: > Curious what he point of having the Scroll button example would be? Wouldn't a primary + secondary illustrate the use of <nav> with the aria-label that you would use if you only had primary, as well as the aria-label if you had secondary?

Oh, yeah, good point. I think originally I thought that the primary tabs would be labeled differently. But now that I see your new example, I like how you have it. 

I'm not sure what the best approach for the accessibility documentation should be. I like that it's mentioned at the top so that consumers know about that without having to scroll through all the examples to discover it, but I also feel like the details should be next to the example. Let's leave it like you have it, except in the A11y table by the new example, include the `aria-label="Local secondary"` for the `.pf-c-tabs.pf-m-tabs-secondary` element.

mcoker: Oh I meant to remove that section from the top example! Do you still think I should keep it?

PR #1555, opened 2019-03-12
feat(chip): add read-only variation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1553

### DIFF
--- a/src/patternfly/components/Chip/chip.scss
+++ b/src/patternfly/components/Chip/chip.scss
@@ -7,12 +7,12 @@
   --pf-c-chip--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-chip--BorderRadius: var(--pf-global--BorderRadius--sm);
 
-  // modified chip
+  // overflow chip
   --pf-c-chip--m-overflow--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-chip--m-overflow--PaddingLeft: 0;
   --pf-c-chip--m-overflow--BorderWidth: 0;
 
-  // modified chip button
+  // overflow chip button
   --pf-c-chip--m-overflow--c-button--BorderRadius: var(--pf-global--BorderRadius--sm);
   --pf-c-chip--m-overflow--c-button--BorderWidth: 0;
   --pf-c-chip--m-overflow--c-button--PaddingLeft: var(--pf-global--spacer--sm);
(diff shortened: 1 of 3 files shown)


srambach:
LGTM 👍

src/patternfly/components/Chip/chip.scss
```diff
@@ -24,6 +24,11 @@
   --pf-c-chip--m-overflow--c-button--focus--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-chip--m-overflow--c-button--focus--BorderColor: var(--pf-global--secondary-color--100);
 
+  // read-only chip
+  --pf-c-chip--m-read-only--PaddingTop: var(--pf-global--spacer--form-element);
```
christiemolloy: It seems weird to me that the modified version is the simplest chip we have with all padding present..would it be wrong to have read only be the simplified chip component and then have a modifier for the chip with the "remove" button. I guess though the chip is always meant to come with the remove button so this should be fine... im just also thinking if we have chips with icons etc.... you can really put anything into a chip.

src/patternfly/components/Chip/examples/chip-example.hbs
```diff
@@ -23,14 +23,21 @@
...
+    7
+  {{/badge}}
   {{#> button button--modifier="pf-m-plain" button--attribute='aria-labelledby="remove_chip_three chip_three" aria-label="Remove" id="remove_chip_three"'}}
     {{> button-icon button-icon--type="times-circle"}}
   {{/button}}
 {{/chip}}
 <br>
 <br>
+{{#> chip chip--type="div" chip--modifier="pf-m-read-only"}}
+  {{#> chip-text}}
```
christiemolloy: should this chip-text have an id

src/patternfly/components/Chip/examples/chip-example.hbs
```diff
@@ -23,14 +23,21 @@
...
+    Read-only Chip
+  {{/chip-text}}
+{{/chip}}
+<br>
+<br>
 {{#> chip chip--type="div" chip--modifier="pf-m-overflow"}}
   {{#> button button--modifier="pf-m-plain"}}
```
christiemolloy: should this chip-text have an id?

src/patternfly/components/Chip/chip.scss
mcoker: >  I guess though the chip is always meant to come with the remove button so this should be fine

That's what I'm thinking, too. Plus that would be a breaking change. Can you give an example of where you would want to use an icon in a chip and how that would affect the layout?

src/patternfly/components/Chip/examples/chip-example.hbs
mcoker: I don't think so. The `id` is only used with `aria-labelledby` when there is an icon button (the close button)

src/patternfly/components/Chip/examples/chip-example.hbs
mcoker: same as above

src/patternfly/components/Chip/chip.scss
christiemolloy: Ok lets stick with this then and avoid a breaking change. I have only seen icons used at the beginning and last element of the chip, but we dont have a case for Patternfly yet so all good

PR #1557, opened 2019-03-12
fix(form): update basic form demo, form action overflow

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1549

To see the issue @karelhala brought up, put a form with an action row in a parent that has no padding and `overflow: auto`. For example, you can drop it in the body of a modal:

Screen Shot 2019-03-12 at 1 02 05 PM

### DIFF
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -167,6 +167,7 @@
   display: flex;
   flex-wrap: wrap;
   margin-bottom: var(--pf-c-form__actions--MarginBottom);
+  overflow: hidden; // keeps the margin bottom on children from causing overflow issues
 
   > * {
     margin-bottom: var(--pf-c-form__actions--child--MarginBottom);
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -21,9 +21,14 @@
   --pf-c-form__group--m-action--MarginTop: var(--pf-global--spacer--xl);
 
   // actions
+  --pf-c-form__actions--MarginTop: calc(var(--pf-c-form__actions--child--MarginTop) * -1);
+  --pf-c-form__actions--MarginRight: calc(var(--pf-c-form__actions--child--MarginRight) * -1);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Form/form.scss
```diff
@@ -167,6 +167,7 @@
   display: flex;
   flex-wrap: wrap;
   margin-bottom: var(--pf-c-form__actions--MarginBottom);
+  overflow: hidden; // keeps the margin bottom on children from causing overflow issues
```
christiemolloy: overflow: hidden is causing the area around the buttons to be cut off - which i dont know is the best option especially when its a submit button it should be clear that its being submitted 

Screen Shot 2019-03-13 at 2 52 34 PM
Screen Shot 2019-03-13 at 2 52 41 PM

src/patternfly/components/Form/form.scss
```diff
@@ -167,6 +167,7 @@
   display: flex;
   flex-wrap: wrap;
   margin-bottom: var(--pf-c-form__actions--MarginBottom);
```
christiemolloy: Looking at this more, I don't know if form__actions should have a margin? It seems so specific to one example and maybe grid can handle the layout.

src/patternfly/components/Form/form.scss
mcoker: Ah good catch. Updated.

src/patternfly/components/Form/form.scss
mcoker: It just needs to offset the bottom margin of the buttons, and the margin on the buttons is for when the buttons wrap, so they have space between them. We don't want to use grid because we don't want columns. What we need is `row-gap` with `display: flex`, which [should be coming soonish](https://github.com/w3c/csswg-drafts/issues/1696). Until then, this is the poor man's form of a `row-gap`.

karelhala:
Thank you for taking care of this! Roping in @Hyperkid123 so he knows when this is merged.

src/patternfly/components/Form/form.scss
christiemolloy: Ok no worries! It would be awesome if there was a selector for overflown-content

PR #1559, opened 2019-03-12
fix(form): helper text missing top margin

mcoker:
> helper text will always appear below a form field

Adds a default top margin to helper text when it's in a form group. An example of when helper text is not in a form group is in the [login component invalid state](https://pf4.patternfly.org/components/Login/examples-full?component=Login%20Invalid)

fixes https://github.com/patternfly/patternfly-next/issues/1546


srambach:
looks fine. :+1:

PR #1563, opened 2019-03-12
fix(table): compact table expanded state styling

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1560

### DIFF
--- a/src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
+++ b/src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
@@ -21,13 +21,12 @@
 
   {{#> table-tbody table-tbody--modifier="pf-m-expanded"}}
     {{#> table-tr table-tr--expanded="true"}}
-      {{#> table-td table-td--toggle="true" table-td--button--attribute=(concat 'aria-labelledby="' table--id '-node1 expandable-toggle1" id="compact-expandable-toggle1" aria-label="Details" aria-controls="' table--id '-content1"')}}{{/table-td}}
+      {{#> table-td table-td--toggle="true" table-td--button--attribute=(concat 'aria-labelledby="' table--id '-node1 ' table--id '-expandable-toggle1" id="' table--id '-expandable-toggle1" aria-label="Details" aria-controls="' table--id '-content1"')}}{{/table-td}}
       {{#> table-td table-td--check="true"}}
         <input type="checkbox" name="checkrow1" aria-labelledby="{{concat table--id '-node1'}}" aria-label="Details">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository Name"}}
         <p id="{{table--id}}-node1">Node 1</p>
-        <a href="#">siemur/test-space</a>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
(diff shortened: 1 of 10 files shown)


src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
```diff
@@ -0,0 +1,87 @@
+{{#> table table--id="compact-expandable-table" table--grid="true" table--modifier="pf-m-compact pf-m-grid-md" table--expandable="true" table--attribute='aria-label="Compact expandable table example"'}}
```
christiemolloy: I wonder if we should design if compact expandable tables need to have the border modifier (once we merge your PR that adds them). 

This looks off, but it might just be me
Screen Shot 2019-03-13 at 3 03 21 PM

src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
```diff
@@ -0,0 +1,87 @@
...
+  {{#> table-thead}}
+    {{#> table-tr}}
+      {{#> table-td}}{{/table-td}}
+      {{#> table-td table-td--check="true"}}
+        <input type="checkbox" name="check-all" aria-label="Select all rows">
+      {{/table-td}}
+      {{#> table-th table-th--attribute='scope="col"' table-th--sortable="true" table-th--modifier="pf-m-width-30" table-th--selected="true" table-th--desc="true"}}
```
christiemolloy: I dont know if the compound expansion table needs the sortable feature in its demo

src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
mcoker: re: borders, I think so. Is what looks off the lack of a border? between the top row and expanded row?

src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
mcoker: removed

src/patternfly/components/Table/examples/table-compact-expandable-example.hbs
christiemolloy: yeah exactly

christiemolloy: but I think its a question from design, do you have the design specs?

mcoker: No, @mceledonia just said that the left border should behave like the regular expanded table example, but I agree it does look weird. Also not sure about the shadow on the right of the expanded area. That shadow fits better when there is a top border.

FWIW, here's what the compact table looks like in a better example with collapsed/expanded rows.

Screen Shot 2019-03-13 at 3 06 01 PM

mcoker:
@mceledonia @christiemolloy updated the spacing and border between the expandable row and the expanded content.

src/patternfly/components/Table/table.scss
```diff
@@ -499,6 +512,15 @@
       --pf-c-table-cell--PaddingRight: var(--pf-c-table--nested--first-last-child--PaddingRight);
     }
   }
+
+  .pf-c-table__expandable-row {
+    .pf-c-table__expandable-row-content {
+      --pf-c-table__expandable-row-content--PaddingTop: var(--pf-c-table--m-compact__expandable-row-content--PaddingTop);
```
christiemolloy: The paddings on all expandable-row-contents should be 24px. 

I had updated table-grid.scss to this, but for some reason the compact table is not assigning it to the only child example (second one).
 `tr > * {
    // Remove First child padding left
    &:first-child {
      --pf-c-table-cell--PaddingLeft: 0;
    }
    // Remove Last child padding right
    &:last-child {
      --pf-c-table-cell--PaddingRight: 0;
    }
  }
`
Screen Shot 2019-03-28 at 11 15 51 AM

mcoker: ah! Thanks. Updated, let me know if that addresses it :)

src/patternfly/components/Table/table-nested.hbs
```diff
@@ -1,4 +1,4 @@
-{{#> table newcontext table--id=table--id table--attribute=(concat table-nested--attribute ' id="' table--id '" aria-label="Nested table"') table--modifier="pf-m-compact" table--grid="true"}}
+{{#> table newcontext table--id=table--id table--attribute=(concat table-nested--attribute ' id="' table--id '" aria-label="Nested table"') table--modifier="pf-m-compact pf-m-no-border-rows" table--grid="true"}}
```
christiemolloy: What do you think about updating the border modifier to .pf-m-borderless-rows , .pf-m-no-border-rows sounds a little odd to me

PR #1566, opened 2019-03-13
fix(alertGroup): fix a11y

### LINKED ISSUE #1562: role="status" on alert-group component causing a11y violations
Should we just increase the a11y threshold to account for these, or is there another preferred approach?

Screen Shot 2019-03-12 at 4 28 14 PM
Screen Shot 2019-03-12 at 4 27 57 PM

cc @jgiardino @seanforyou23

christiemolloy:
fix #1562

### DIFF
--- a/src/patternfly/components/AlertGroup/docs/alert-group.md
+++ b/src/patternfly/components/AlertGroup/docs/alert-group.md
@@ -6,7 +6,6 @@
 
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `role="status"` | `.pf-c-alert-group` |  Communicates contents of the alert message to the user. **Required** |
 
 ### Usage
 
--- a/src/patternfly/components/AlertGroup/docs/alert-group.md
+++ b/src/patternfly/components/AlertGroup/docs/alert-group.md
@@ -2,11 +2,6 @@
 
 `.pf-c-alert-group` is optional when only one alert is needed. It becomes required when more than one alert is used in a list.
 
-## Accessibility
-
-| Attribute | Applied To | Outcome |
-| -- | -- | -- |
-
 ### Usage
 
 | Attribute | Applied To | Outcome |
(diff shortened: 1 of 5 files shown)


christiemolloy:
@srambach still getting this error:

1 error found in: /components/AlertGroup/examples-full/
1) <li> elements must be contained in a <ul> or <ol> -- https://dequeuniversity.com/rules/axe/3.1/listitem -- impact: serious
      
-------html violation instances-------
<li class="pf-c-alert-group__item">
<li class="pf-c-alert-group__item">
<li class="pf-c-alert-group__item">

src/patternfly/demos/Alert/examples/alert-toast-example.hbs
```diff
@@ -1,5 +1,5 @@
 {{> page-demo-default}}
-{{#> alert-group alert-group--attribute='role="status"' alert-group--modifier="pf-m-toast"}}
+{{#> alert-group alert-group--modifier="pf-m-toast"}}
```
jgiardino: So weird! I see that you removed role="status", but it's still showing up in this example:

```
<ul class="pf-c-alert-group pf-m-toast" role="status">
  <li class="pf-c-alert-group__item">
```

christiemolloy:
Ah looks like it was in two examples! should be good now - @srambach @jgiardino

src/patternfly/components/AlertGroup/docs/alert-group.md
```diff
@@ -6,7 +6,6 @@
 
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `role="status"` | `.pf-c-alert-group` |  Communicates contents of the alert message to the user. **Required** |
```
mcoker: Can you remove this whole a11y section?

Screen Shot 2019-03-13 at 1 54 37 PM

srambach:
Latest changes fix the Alert and AlertGroup errors I saw before...but the total increased because it tacked more onto the end. Reran it, now it's down to 18. I am fine with merging (with @mcoker's suggestion above) because the errors this intended to fix seem to be fixed.

src/patternfly/components/AlertGroup/docs/alert-group.md
christiemolloy: Yep

christiemolloy:
@srambach if you take a look now I merged the table changes, we are down to 15 violations and a threshold of 16

mcoker:
Should we have the threshold match the current number of violations?

PR #1568, opened 2019-03-13
docs(readme): add pf3/pf4 compatibility css

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1567

@dgutride per our discussion, I added a blurb to the README for this. I think it's inevitable this will become a stylesheet, and I imagine anyone using PF3 and PF4 together will run into this issue and benefit from this CSS.

I was chatting with @seanforyou23 about it and he mentioned that this is really similar to https://github.com/patternfly/patternfly-next/wiki/IE11-Support#step-2-add-ie11-compatible-styling-overrides and we could host these in a common place. Just wanted to start the discussion about whether the README is the best spot for these, or if we want to explore other options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,16 @@ If you have any suggestions about ways that we can improve how we use this tool,
 
 [How do I use CSS variables to customize
 the library?](https://pf4.patternfly.org/guidelines#variables)
+
+### PatternFly 3 Compatibility
+
+Due to differences in how PatternFly 3 and PatternFly 4 CSS is written, there are cases where PatternFly 3 CSS may take precedence over PatternFly 4 CSS in an application that uses both systems. Below are some snippets that will help ensure PatternFly 4 components are displayed consistently.
+
+```
+.pf-c-dropdown__menu-item {
+  &:hover,
+  &:focus {
+    color: var(--pf-c-dropdown__menu-item--hover--Color);
+  }
+}
+```


mcoker:
FWIW, I'm seeing this in UHC, too, since they're using PF3/PF4 together. And like I mentioned when we met about this, there are issues popping up from UHC not using our global reset styles, too. Note the white bullets beside the dropdown menu items.

Screen Shot 2019-03-19 at 9 57 51 AM

If we offer styles to correct these types of issues, I think since problems like these are likely going to keep popping up and will likely be an issue with lots of apps, having them copy/paste or grab styles from somewhere manually likely isn't the best approach. I'm more in favor of having a stylesheet that folks using PF3/PF4 together and aren't using the global reset can include in their project to automatically get these kinds of updates. To that point, we could also move towards asking products to include the global stylesheet by default and make it a requirement for using PF4, and then we would only need to provide overrides for PF3/PF4 compatibility.

cc @janwright73

seanforyou23:
+1 to all of this, I strongly prefer these types of styling patches be centralized in a single place that's included as part of the core distribution so it can be consumed in an automated fashion. Placing it in a README I think will not be good for us or our consumers, it forces a manual workflow for each update.

If we're going to do this for PF3/PF4 compatibility styles, it would make sense to go ahead and decide where the IE11 compatibility styles should live also. Seems like they might be found along side one another in a `compat` directory or something of the sort.

Anything I can do to help?

seanforyou23:
Looks like just needs a rebase? Also, you can group this along side the ie11 stylesheet by adjusting this path https://github.com/patternfly/patternfly-next/blob/master/build/npm-scripts/ie-convert-all.js#L7

mcoker:
@dgutride do you know if this has continued to be an issue? I haven't come across a product needing support for this since the PR was opened. Curious if we should close this PR for now and revisit if it comes up again?

PR #1586, opened 2019-03-15
feat(nav): add nav item separator

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1574

### DIFF
--- a/src/patternfly/components/Nav/examples/nav-example.hbs
+++ b/src/patternfly/components/Nav/examples/nav-example.hbs
@@ -10,6 +10,7 @@
         Link 2
       {{/nav-link}}
     {{/nav-item}}
+    {{#> nav-item nav-item--modifier="pf-m-separator" nav-item--attribute='role="separator"'}}{{/nav-item}}
     {{#> nav-item}}
       {{#> nav-link nav-link--href="#"}}
         Link 3
--- a/src/patternfly/components/Nav/examples/nav-example.hbs
+++ b/src/patternfly/components/Nav/examples/nav-example.hbs
@@ -10,7 +10,7 @@
         Link 2
       {{/nav-link}}
     {{/nav-item}}
-    {{#> nav-item nav-item--modifier="pf-m-separator" nav-item--attribute='role="separator"'}}{{/nav-item}}
+    {{#> nav-separator}}{{/nav-separator}}
     {{#> nav-item}}
       {{#> nav-link nav-link--href="#"}}
         Link 3
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -157,6 +157,8 @@
   // List item seperator
   --pf-c-nav__separator--Height: var(--pf-global--BorderWidth--sm);
   --pf-c-nav__separator--BackgroundColor: var(--pf-global--BorderColor--light);
+  --pf-c-nav__separator--MarginTop: var(--pf-global--spacer--sm);
+  --pf-c-nav__separator--MarginBottom: var(--pf-global--spacer--sm);
   --pf-c-nav__separator--MarginLeft: var(--pf-c-nav__list-link--PaddingLeft);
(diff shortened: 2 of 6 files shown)


src/patternfly/components/Nav/examples/nav-example.hbs
```diff
@@ -10,6 +10,7 @@
         Link 2
       {{/nav-link}}
     {{/nav-item}}
+    {{#> nav-item nav-item--modifier="pf-m-separator" nav-item--attribute='role="separator"'}}{{/nav-item}}
```
christiemolloy: In the options menu we also have a separator with the class - pf-c-options-menu__separator. I wonder if we should update that component and potentially other components to use the modifier like here. And would we add a utility class for this?

src/patternfly/components/Nav/nav.scss
```diff
@@ -154,6 +154,13 @@
   --pf-c-nav__simple-list-link--m-current--FontWeight: var(--pf-global--FontWeight--semi-bold);
   --pf-c-nav__simple-list-link--m-current--BackgroundColor: var(--pf-global--BackgroundColor--150);
 
+  // List item seperator
+  --pf-c-nav__item--m-separator--Height: var(--pf-global--BorderWidth--sm);
+  --pf-c-nav__item--m-separator--BackgroundColor: var(--pf-global--BorderColor--light);
+  --pf-c-nav__item--m-separator--MarginLeft: var(--pf-c-nav__list-link--PaddingLeft);
```
christiemolloy: Could you check with design about the top and bottom margins between the separator. Its looking extra tight between items with a separator, compared with not. 
Screen Shot 2019-03-20 at 11 28 30 AM

For example we did this in the options menu:
Screen Shot 2019-03-20 at 11 29 38 AM

If its not in the scope of this PR we should create an issue!

mceledonia: Yeah let's go ahead and add the same spacers as in the options menu separator. Consistency and also it aligns with the function of the separator - to separate groups of content. It also just looks nicer!

src/patternfly/components/Nav/examples/nav-example.hbs
mcoker: Ah good catch! I just used what was already in the nav component for this. Personally I think it's better as it's own element, and not as a modifier. If we were to add any sort of styles to a regular nav-item (or options menu item, etc) like padding, borders, background color, hover/focus treatment, etc, I can't imagine we would ever want those styles to apply to the separator. So in the CSS, we would need to use `:not(.pf-m-separator)` or whatever to target non-separator items. To me a regular item and separator item are separate elements. WDYT?

And can you give me an example of how we would benefit from adding this as a utility class? The approach so far has been to not create a utility class unless it's obviously useful, and instead wait for folks to request one.

christiemolloy: I agree with what you said above I dont think it should be a modifier. Would we make the separator a component? or just an element in the nav?

To me a separator has the same properties as a border. For example in the data table and data list examples I had to add extra borders to the component in order to meet the design specs, however maybe a separator would have been better there? I can see a separator utility being added to a lot of cases to separate content or add an extra border between two elements. But in that case, what if we made it its own component?

mcoker:
@christiemolloy @mceledonia updated the separator to be its own element and add top/bottom margin. Once https://github.com/patternfly/patternfly-next/pull/1600 is merged, I'll make sure it uses the new common border color.

src/patternfly/components/Nav/nav.scss
christiemolloy: @mceledonia , @mcoker updated the margins for the separator which looks really good, but I think this looks odd. If you're fine with it thats great, but if you have any suggestions let us know (maybe adding more padding not sure)..
Screen Shot 2019-03-27 at 5 11 38 PM

mceledonia: That does look a bit odd, but is that how I would expect it to work. Since it's only on hover and only when you're before a divider we can be totally fine with it.

PR #1595, opened 2019-03-18
fix(select): make 100% width and fix button and badge position

### LINKED ISSUE #1539: Select Component Width / Badge fixes
Looks like if you make the select component 100% width (what react does by default), the toggle button doesn't grow. That should grow to fill the width of the component.

screen shot 2019-03-05 at 1 44 13 pm

Also, this is an existing problem, so let me know if you'd like to address this in another issue/PR, but the badge is currently centered in the toggle because of justify-items: space-between; on the parent.

screen shot 2019-03-05 at 1 44 24 pm

christiemolloy:
fix #1539

### DIFF
--- a/src/patternfly/components/Select/select-toggle.hbs
+++ b/src/patternfly/components/Select/select-toggle.hbs
@@ -37,9 +37,7 @@
 				<span>Select</span>
 		{{/if}}
 
-	</div>
-
-	{{!-- Third part, badge --}}
+		{{!-- Third part, badge --}}
 		{{#if select--IsCheckboxSelect}}
 			{{#if select--ItemIsSelected}}
 				<div class="pf-c-select__toggle-badge">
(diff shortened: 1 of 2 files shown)


mcoker:
This is outside of the scope of this PR, but should the selector for this block https://github.com/patternfly/patternfly-next/blob/ecdb09577c4ed2fc462211cd4a425c454dfa2043/src/patternfly/components/Select/select.scss#L263-L266

be `  > *:not(:last-child)` or `  > * + * { margin-left: ... }` instead? Right now there is a margin-right on the badge.

src/patternfly/components/Select/select-toggle.hbs
```diff
@@ -37,9 +37,7 @@
 				<span>Select</span>
```
mcoker: Can you fix the indentation in this file? Looks like it's using the wrong tab space value

mcoker:
Looks great, left a couple of comments.

christiemolloy:
@mcoker ready for another review

PR #1601, opened 2019-03-19
refactor(context-selector): aria refinements

rebeccaalpert:
Fixes https://github.com/patternfly/patternfly-next/issues/1531.

Added ids to `.pf-c-context-selector__toggle` and `.pf-c-button pf-m-tertiary`, updated the aria-label on `.pf-c-button pf-m-tertiary`, and added aria-labelledby to `.pf-c-context-selector__toggle` and `.pf-c-form-control`.

### DIFF
--- a/src/patternfly/components/ContextSelector/examples/context-selector-example.hbs
+++ b/src/patternfly/components/ContextSelector/examples/context-selector-example.hbs
@@ -1,5 +1,5 @@
-{{#> context-selector id="context-selector-collapsed-example" context-selector--label-text="Selected Project"}}
-  {{#> context-selector-toggle context-selector-toggle--attribute='id="context-selector-expanded-example-toggle" aria-labelledby="context-selector-expanded-example-toggle context-selector-expanded-example-label"'}}
+{{#> context-selector context-selector--id="context-selector-collapsed-example" context-selector--label-text="Selected Project"}}
+  {{#> context-selector-toggle context-selector-toggle--attribute=(concat 'id="' context-selector--id '-toggle"' 'aria-labelledby="' context-selector--id '-toggle ' context-selector--id '-label"')}}
     {{#> context-selector-toggle-text}}
       My Project
     {{/context-selector-toggle-text}}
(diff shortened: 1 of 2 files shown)


src/patternfly/components/ContextSelector/examples/context-selector-example.hbs
```diff
@@ -1,5 +1,5 @@
 {{#> context-selector id="context-selector-collapsed-example" context-selector--label-text="Selected Project"}}
-  {{#> context-selector-toggle}}
+  {{#> context-selector-toggle context-selector-toggle--attribute='id="context-selector-expanded-example-toggle" aria-labelledby="context-selector-expanded-example-toggle context-selector-expanded-example-label"'}}
```
christiemolloy: We should give the context selector an id: "context-selector--id" and then reuse that in the context-selector-toggle  as the ID's need to be unique.  For example you can look at the table.
it would be:

`{{#> context-selector-toggle context-selector-toggle--attribute=(concat 'id="' context-selector--id '-toggle" aria-labelledby="context-selector-expanded-example-toggle context-selector-expanded-example-label"')}}`

src/patternfly/components/ContextSelector/examples/context-selector-example.hbs
```diff
@@ -82,9 +82,9 @@
   {{#> context-selector-menu}}
     {{#> context-selector-menu-input}}
       {{#> input-group}}
-        {{#> form-control controlType="input" input="true" form-control--attribute='type="search" placeholder="Search" id="textInput2" name="textInput2" aria-label="search through menu items"'}}
+        {{#> form-control controlType="input" input="true" form-control--attribute='type="search" placeholder="Search" id="textInput2" name="textInput2" aria-labelledby="search-button"'}}
         {{/form-control}}
-        {{#> button button--modifier="pf-m-tertiary" button--attribute='aria-label="search button for search input"'}}
+        {{#> button button--modifier="pf-m-tertiary" button--attribute='id="search-button" aria-label="Search menu items"'}}
```
christiemolloy: id needs to be unique! You should use the same comments that I made above and concat the context-selector--id onto the search button.

rebeccaalpert:
I added the unique ids.

src/patternfly/components/ContextSelector/examples/context-selector-example.hbs
```diff
@@ -1,5 +1,5 @@
-{{#> context-selector id="context-selector-collapsed-example" context-selector--label-text="Selected Project"}}
-  {{#> context-selector-toggle}}
+{{#> context-selector context-selector--id="context-selector-collapsed-example" context-selector--label-text="Selected Project"}}
+  {{#> context-selector-toggle context-selector-toggle--attribute=(concat 'id="' context-selector--id '-toggle"' 'aria-labelledby="' context-selector--id '-toggle ' context-selector--id '-label"')}}
```
jgiardino: Can you switch the order in the `aria-labelledby` value so that it's: 
`'aria-labelledby="' context-selector--id '-label ' context-selector--id '-toggle"`

This will announce the label text "Selected Project:" first before announcing the toggle label (i.e. the selected project)

jgiardino:
Thanks for making these updates! There is just one minor update needed, which I noted inline.

PR #1605, opened 2019-03-20
chore(browser): add browser support statement

dgutride:


README.md
```diff
@@ -94,5 +94,15 @@ If you have any suggestions about ways that we can improve how we use this tool,
 
 ## FAQ
 
+#### CSS Variables
 [How do I use CSS variables to customize
 the library?](https://pf4.patternfly.org/guidelines#variables)
+
+#### Browser Support
+PatternFly 4 is supported on the following browsers:
```
mcoker: Should we mention that we support the latest version?

README.md
```diff
@@ -94,5 +94,15 @@ If you have any suggestions about ways that we can improve how we use this tool,
...
 the library?](https://pf4.patternfly.org/guidelines#variables)
+
+#### Browser Support
+PatternFly 4 is supported on the latest version of the following browsers:
+- Chrome
+- Firefox
+- Safari
+- Edge
+
+We also provide documentation for IE11 support.  This isn't baked into the library directly, but we provide information on how to achieve it in our [wiki](https://github.com/patternfly/patternfly-next/wiki/IE11-Support).
```
seanforyou23: I wonder if this will raise more questions than it answers? I hope we can avoid saying support isn't baked into the lib directly, (the [tooling](https://github.com/patternfly/patternfly-next/blob/master/build/npm-scripts/ie-conversion-utils.js) to achieve it is) just that we currently only have fallback styles for a subset of the components. I think by enumerating which components have these fallback styles, it might better help people understand what the lift looks like for them. Maybe this information is better suited for the wiki and not this readme, just something to think about.

README.md
```diff
@@ -94,5 +94,15 @@ If you have any suggestions about ways that we can improve how we use this tool,
 
 ## FAQ
 
+#### CSS Variables
 [How do I use CSS variables to customize
 the library?](https://pf4.patternfly.org/guidelines#variables)
+
+#### Browser Support
+PatternFly 4 is supported on the latest version of the following browsers:
```
christiemolloy: Should we talk about how many of the previous versions we support - I think you mentioned the latest two?

README.md
```diff
@@ -94,5 +94,15 @@ If you have any suggestions about ways that we can improve how we use this tool,
 
 ## FAQ
 
+#### CSS Variables
 [How do I use CSS variables to customize
 the library?](https://pf4.patternfly.org/guidelines#variables)
+
+#### Browser Support
```
christiemolloy: Should we also have a section for device support?

README.md
dgutride: Added this

README.md
dgutride: I'd rather be pretty vague here for now than enumerating specific components.  If you'd like to update the wiki with that, it would be a better place.

README.md
dgutride: Let's leave it at latest initially.

README.md
dgutride: I think we should leave device support out, we should endeavor to support all devices that run the latest versions of the specified browsers.

PR #1610, opened 2019-03-20
fix(login): language selector should use select

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1588

### DIFF
--- a/src/patternfly/components/Login/examples/login-invalid-example.hbs
+++ b/src/patternfly/components/Login/examples/login-invalid-example.hbs
@@ -12,11 +12,9 @@
         {{#> login-main-header-desc}}
           Lorem ipsum dolor sit amet, consectetur adipiscing elit.
         {{/login-main-header-desc}}
-        {{#> dropdown id="invalid-login-demo-dropdown-select" dropdown--IsSelect="true" dropdown--ItemIsSelected="true" dropdown--HasToggleIcon="true"}}
-          {{#> dropdown-toggle-text}}
-            English
-          {{/dropdown-toggle-text}}
-        {{/dropdown}}
+        {{#> select id="login-invalid-select-single" select--IsLanguageMenu="true"}}
+          English
+        {{/select}}
       {{/login-main-header}}
       {{#> login-main-body}}
         {{#> form}}
--- a/src/patternfly/components/Login/examples/login-simple-example.hbs
+++ b/src/patternfly/components/Login/examples/login-simple-example.hbs
@@ -12,11 +12,9 @@
         {{#> login-main-header-desc}}
           Lorem ipsum dolor sit amet, consectetur adipiscing elit.
         {{/login-main-header-desc}}
-        {{#> dropdown id="login-demo-dropdown-select" dropdown--IsSelect="true" dropdown--ItemIsSelected="true" dropdown--HasToggleIcon="true"}}
-          {{#> dropdown-toggle-text}}
-            English
-          {{/dropdown-toggle-text}}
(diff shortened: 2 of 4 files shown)


src/patternfly/components/Login/examples/login-invalid-example.hbs
```diff
@@ -12,11 +12,9 @@
         {{#> login-main-header-desc}}
           Lorem ipsum dolor sit amet, consectetur adipiscing elit.
         {{/login-main-header-desc}}
-        {{#> dropdown id="invalid-login-demo-dropdown-select" dropdown--IsSelect="true" dropdown--ItemIsSelected="true" dropdown--HasToggleIcon="true"}}
-          {{#> dropdown-toggle-text}}
-            English
-          {{/dropdown-toggle-text}}
-        {{/dropdown}}
+        {{#> select id="login-invalid-select-single" select--IsLanguageMenu="true"}}
```
christiemolloy: The select is taking up 100% width on mobile due to the change we made in Core. I wonder if it should be wrapped in a container here.
Screen Shot 2019-03-22 at 1 01 59 PM

christiemolloy: This is what I'm seeing on Sketch
Screen Shot 2019-03-22 at 1 01 59 PM

christiemolloy: Screen Shot 2019-03-22 at 1 04 42 PM

mcoker: @gdoyle1 @mcarrano can you advise here? Since these designs were created, we have made the width of the select component 100%, so it fills the width of it's parent. We don't have a variation of the select component that will shrink back down to what it was before, where the toggle is as wide as what's in it, and the menu is as wide as the widest thing inside of it.

Just want to verify that we want selects to be 100% width by default? That's inconsistent with the other types of menus (dropdown, options menu, etc). Not saying it should be consistent, just some feedback.

And assuming so, should this select be full-width on mobile as seen in the first screenshot @christiemolloy provided? Or should we limit the width of the select? 

If we limit the width, should that be a variation we offer in the select component itself (for example, to offer a modifier that displays a select's toggle/menu width like a dropdown?)? Or should I use the layout in the login component to make the select some percentage or whatever of the available space?

mcarrano: @mcoker I would not expect the Select to always fill it's container at 100%. I do remember some prior conversation on this where we said that the width or the toggle should not adjust based on the length of the selected item. But I didn't realize this meant that we were always sizing to 100%. I think I expected that you could set a fixed width on a Select list.

How do we deal with this in a toolbar, for example? We don't stretch the contents of the toolbar to fill the width of the screen, right?

mcarrano:
@mcoker kmcfaul, @mceledonia and I met to discuss this issue. Seems that this is a bigger problem with the current approach to the language selector where it has the potential to break the layout if the selected value is long.

Screen Shot 2019-03-25 at 12 57 15 PM

Our recommendation is to move the language selector to the background (dark) area of the page where it will fall in the DOM before other login form elements. This will require a design change to the login page and a version of Select that works on the dark theme.

Since this is a potential breaking change, our recommendation is to remove the language selector for release and introduce this at a future date after all design contingencies can be worked out. @rachael-phillips can you comment? Is there any exposure in removing the Language Selector in the short term?

mcoker:
Closing this for now until we figure out the design.

PR #1614, opened 2019-03-21
fix(a11y): fix switch, table, and about modal violations

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/619

There were also a bunch of table violations.

### DIFF
--- a/test-a11y/config.js
+++ b/test-a11y/config.js
@@ -1,5 +1,5 @@
 module.exports = {
-  toleranceThreshold: 13,
+  toleranceThreshold: 4,
   host: 'localhost',
   port: '8000',
   protocol: 'http',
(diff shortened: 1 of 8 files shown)


test-a11y/config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  toleranceThreshold: 13,
+  toleranceThreshold: 5,
```
jgiardino: bitmoji

jgiardino:
You rock!

If you're fixing weird `id`s in the Table component, I came across this in the **Table with Compound Expansion** example
image

mcoker:
@jgiardino nice find! fixed that.

PR #1615, opened 2019-03-22
feat(nav): add overflow styling to horizontal nav examples

### LINKED ISSUE #1519: Fix Tertiary Nav Scroll
This is how the Tertiary Nav looks in Chrome vs Firefox. We should hide the scrollbar in the css like we will do in the Tab component #1514 Also fix Horizontal Nav

screen shot 2019-03-01 at 11 06 06 am

christiemolloy:
fix #1519 

- Is it possible for the box shadow to change to white in the dark theme?

Visuals:
Screen Shot 2019-03-22 at 10 00 03 AM


mcoker:
Does this component's scroll buttons need the current styles, like used in tabs?

christiemolloy:
@mcoker I don't think so as its just meant to be an addition to overflow scrolling but I can check

mcoker:
Since the UI and basic interaction is the same, seems like they should match in that regard. Also, since this pattern is effectively the same between the components, I'm curious if this should be it's own component that nav and tabs use. And maybe it has a modifier to create both the UI used for the primary and secondary tabs examples.

christiemolloy:
In terms of a new component would you be talking about the overflow button as a component or the entire horizontal list overflow. I don't know how I feel about that because one is a nav and one is a secondary tab list, and the lists have really different styles. and then we have the horizontal nav version for the masthead that seems like it should be an example of the nav component rather than using a component. I could also see the styles for secondary tabs maybe changing too @mcoker

mcoker:
Looks like it would be an overflow-scroll component that you put a list inside of, and it could apply to tabs, secondary tabs, horizontal navigation, or if you had another list somewhere you wanted to treat similarly. Just in general, it's the same structure, and if I came across this pattern in the nav and tabs, I would expect it to behave the same or at least very similarly. The visual differences could be handled in the choice of icon or modifiers applied. It just looks like there is a lot of copy and paste between the two currently. What I would want to avoid is having 2 unique implementations because there are slight design changes between them. And question the design differences anyways, and ask if, since they're very similar patterns, they should present themselves more similarly, too.

I don't think we need to worry about it for this PR, but something to think about in the future.

src/patternfly/components/Nav/nav.scss
```diff
@@ -550,3 +585,34 @@
...
+.pf-c-nav__scroll-button {
+  position: absolute;
+  top: var(--pf-c-nav__scroll-button--Top);
+  bottom: 0;
+  z-index: var(--pf-c-nav__scroll-button--ZIndex);
+  width: var(--pf-c-nav__scroll-button--Width);
+  padding-right: var(--pf-c-nav__scroll-button--PaddingRight);
+  padding-left: var(--pf-c-nav__scroll-button--PaddingLeft);
+  background-color: var(--pf-c-nav__scroll-button--BackgroundColor);
+  border: initial;
```
mcoker: Is there a reason you use `initial` over `0`?

christiemolloy: This was used in the tab example so I thought it was purposeful. I can make a note to update both.

christiemolloy:
@mcoker the reason I moved `overflow-x: auto` to the horizontal/tertiary list elements is because on the nav component itself, it was causing the buttons to scroll with the list. But placing it on the list fixes this.
Also something strange is happening in Firefox: the button isn't positioning absolute to its parent, its extending a certain amount below (hard to see), I don't think its because we're hiding the scrollbar with width: none, because this hasn't been a problem before.

Screen Shot 2019-03-26 at 10 57 51 AM

mcoker:
Just want to verify that the right-side scroll button should extend to the edge of the viewport like this?

Screen Shot 2019-03-26 at 1 57 17 PM

Also when I place the horizontal navigation in the page header nav, I can't scroll and the scroll button isn't visible.

Screen Shot 2019-03-26 at 2 02 53 PM

mcoker:
And this is how tertiary nav looks with both scroll buttons visible.

Screen Shot 2019-03-26 at 2 05 11 PM

christiemolloy:
`--pf-c-page__main-nav--PaddingRight: 0` is assigned to 0 for some reason

christiemolloy:
@mcoker updated this, visuals have approved of the box shadows.
Two things:
- The box shadow for light and dark are hard coded. I didn't make them global vars because I think its a one off (i.e. we dont have a var for a white box-shadow or one that is the same color as dark mode), but that means the light/dark theming doesn't apply here and I had to add the dark var to the component...which I dont think we want.
- Also I added a media query to the tertiary-list-item so that the first child has a padding-left of 32px on desktop, and then a padding-left of 16px on mobile. This looks really good, but I don't think we want to handle the padding in the component, and it would definitely be better for `.pf-c-page-main-nav` to handle the padding....so maybe we do explore negative positioning on the button?

I also added some styles to the page to show how the nav would work.. Ill remove those, but definitely check out the nav on the page component.

christiemolloy:
@mcoker fyi if you click on "the second option for nav" I added a different variation for this option that adds negative margin to the buttons

mcoker:
Re: box shadows, that's fine with me. Are we sure we don't want to introduce dark mode box shadows?

I like option 2. by...

> also what if the .pf-c-page__main-nav handled the pf-c-nav__scroll-button , since its a one off case.

Do you mean that element handles the negative positioning on the buttons? If so, I think that's ideal. Considering this should be a modular component, it should follow the rules of the component and the parent should handle the positioning for one-off stuff like this.

Though just tossing it out there, what if the buttons didn't go so far into the negative space, and the icon aligned with the content?

Screen Shot 2019-03-27 at 4 50 28 PM

side note - do you know why this and the tabs button use `width` and don't let the padding determine the width?

christiemolloy:
@mcoker I successfully added the horizontal nav to the masthead in the page demo & updated the box shadow color as per @mceledonia  recommendation

src/patternfly/components/Nav/docs/code.md
```diff
@@ -24,6 +24,27 @@ The navigation system relies on several different sub-components:
 
 | Class | Applied To | Outcome |
 | -- | -- | -- |
+<<<<<<< HEAD
```
mcoker: Looks like you have some merge remnants left in here...

src/patternfly/components/Nav/nav.scss
```diff
@@ -164,20 +164,46 @@
...
+  --pf-c-nav__scroll-button--m-dark--nth-of-type-1--BoxShadow: 20px 0 10px -4px rgba(21, 21, 21, .7); // This is a special one-off shadow for nav
+  --pf-c-nav__scroll-button--m-dark--nth-of-type-2--BoxShadow: -20px 0 10px -4px rgba(21, 21, 21, .7); // This is a special one-off shadow for nav
+
   // Base styling
-  overflow-x: auto;
+  position: relative;
 
-  // Hide scrollbars
-  &::-webkit-scrollbar {
+  &.pf-m-dark {
```
mcoker: I wonder if instead of `pf-m-dark`, you should just style the button `box-shadow` in the horizontal nav? And can the color and background colors not inherit from `@include pf-t-dark;` in `.pf-c-nav__horizontal-list`?

src/patternfly/components/Page/page.scss
```diff
@@ -204,6 +215,10 @@
   overflow: hidden;
   background-color: var(--pf-c-page__header-nav--BackgroundColor);
 
+  > * {
```
mcoker: is there a reason you're using `*` instead of `.pf-c-nav`?

src/patternfly/demos/Page/page-template-title.hbs
```diff
@@ -1,3 +1,38 @@
+{{#> page-main-nav page-main-section--modifier="pf-m-light"}}
```
mcoker: Are you sure you want to add tertiary nav to every demo? Also the expandable nav already had tertiary nav, so now you have 2 in that demo - https://deploy-preview-1615--pf-next.netlify.com/demos/Page/examples-full/?component=Page%20Component%20Expandable%20Nav%20Example

src/patternfly/components/Page/page.scss
christiemolloy: I was thinking for flexibility but I don't think that's necessary here so good point!

src/patternfly/components/Nav/nav.scss
```diff
@@ -164,20 +164,33 @@
...
+  --pf-c-nav__scroll-button--PaddingLeft: var(--pf-global--spacer--sm);
+  --pf-c-nav__scroll-button--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-nav__scroll-button--hover--Color: var(--pf-global--link--Color);
+  --pf-c-nav__scroll-button--nth-of-type-1--BoxShadow: 20px 0 10px -4px rgba(255, 255, 255, .7); // This is a special one-off shadow for nav
+  --pf-c-nav__scroll-button--nth-of-type-2--BoxShadow: -20px 0 10px -4px rgba(255, 255, 255, .7); // This is a special one-off shadow for nav
+  --pf-c-nav__scroll-button--m-dark--nth-of-type-1--BoxShadow: 20px 0 10px -4px rgba(21, 21, 21, .7); // This is a special one-off shadow for nav
+  --pf-c-nav__scroll-button--m-dark--nth-of-type-2--BoxShadow: -20px 0 10px -4px rgba(21, 21, 21, .7); // This is a special one-off shadow for nav
+
   // Base styling
-  overflow-x: auto;
```
mcoker: Sorry if I asked you this already, but why move this from nav to the horizontal nav components? Curious if this is necessary?

mcoker:
@christiemolloy this looks great! Just one question about moving the `overflow` property from the main nav component into the horizontal nav variations.

Also, I see you are not using pf-m/t-dark, or including `%pf-t-dark` anywhere. Is the only remaining issue with the examples? If so, you can update the example component for those 2 examples to also include the `pf-t-dark` class.


```
      <Example
        heading="Horizontal Nav (only in masthead)"
        handlebars={navHorizontalListExampleRaw}
        className="is-dark-preview pf-t-dark"
      >
        {navHorizontalListExample}
      </Example>
      <Example
        heading="Horizontal Nav Overflow (only in masthead)"
        handlebars={navHorizontalListOverflowExampleRaw}
        className="is-dark-preview pf-t-dark"
      >
```

mcoker:
Ah, I see what you're saying about needing a dark class on the nav if the buttons are outside of the list. In this case, we style another nav element based on the page element it's applied in, and I think that's fine here, too. The buttons have to be outside of the list though, otherwise you get a11y errors. Plus if we make this a component in the future, the buttons will need to be outside of the list (and probably the entire nav element). 

In this PR - https://github.com/christiemolloy/patternfly-next/pull/13 I refactored that and fixed a couple of a11y issues, as well as used `aria-label="Global"` on the main nav components in the demos, to match what's on react and what was on master before this PR. I'm assuming that was lost in a merge conflict? I'm not sure about whether to use `aria-label="Global"` or `aria-label="Local"` on the tertiary nav in the demos though. The tertiary nav in the current page demo on master uses "Global" but the tertiary nav in the Nav component examples uses "Local". Maybe @jgiardino can guide us here?

christiemolloy:
@mcoker thanks for making those changes! This is happening in the example because the page__header-nav has 'width: auto` on it rather than width: 100%, but it works in the demo, so not sure if it even matters:

Screen Shot 2019-04-04 at 10 43 19 AM

mcoker:
@christiemolloy I think that's fine. This is one of those cases where the example doesn't look right because it's outside of the context where it is designed to be used. Though looks like you could add `.pf-c-page__header` as a wrapper to those examples and that fixes it.

jgiardino:
@mcoker I'm not seeing what you're describing. 

We're using what's called **Horizontal Nav**  not **Tertiary Nav** in our Page component. And on the the Nav component page, it does have `aria-label="Global"`, which is correct.

image

I did notice that before this PR, the page component has two aria-label attributes though. One of them was changed to "Global" in this PR, so now at least they're both correct, but we only need one. 😉

src/patternfly/demos/DataList/data-list-page-nav.hbs
```diff
@@ -1,4 +1,4 @@
-{{#> nav nav--attribute=(concat 'id="' page--id '-primary-nav" aria-label="Grouped Nav Example"')}}
+{{#> nav nav--attribute=(concat 'id="' page--id '-primary-nav" aria-label="Global"')}}
```
jgiardino: `aria-label` appears to already be defined on the `nav` handlebar. It's showing up twice in the Page and Datalist demos.

mcoker:
@jgiardino we use tertiary nav on the [expandable nav page demo](https://deploy-preview-1615--pf-next.netlify.com/demos/Page/examples-full/?component=Page%20Component%20Expandable%20Nav%20Example). 

Screen Shot 2019-04-04 at 10 28 38 AM

It's also in all of the [data list demos](https://pf4.patternfly.org/demos/DataList/examples/) and [data table demos](https://pf4.patternfly.org/demos/DataTable/examples/)

jgiardino:
> we use tertiary nav on the [expandable nav page demo](https://deploy-preview-1615--pf-next.netlify.com/demos/Page/examples-full/?component=Page%20Component%20Expandable%20Nav%20Example).

Thanks for the example, @mcoker! That example uses the correct aria-labels. The primary nav is referred to as "global" navigation. Then the tertiary nav region is "local" navigation relative to the "Resource Usage" page that's selected in the global navigation.

src/patternfly/demos/DataList/data-list-page-nav.hbs
mcoker: Nice find! Looks like we should just remove any reference to `aria-label` from all of the nav components in the examples, except for the tertiary nav used in the page expandable demo, data list demo, and data table demos. And in those cases, define `aria-label` as an attribute of the hbs partial, not in the `nav--attribute` attribute value, since this is the logic that generates `aria-label` in `nav.hbs`

```
<nav class="pf-c-nav{{#if nav--modifier}} {{nav--modifier}}{{/if}}"
  {{#if aria-label}}
    aria-label="{{aria-label}}"
  {{else}}
    aria-label="Global"
  {{/if}}
  {{#if nav--attribute}}
    {{{nav--attribute}}}
  {{/if}}>
  {{> @partial-block}}
</nav>
```

mcoker: Or alternatively, we could remove the `aria-label` logic from `nav.hbs`, and include it in `nav--attribute` instead. That would probably be more consistent with our other partials.

src/patternfly/demos/DataList/data-list-page-nav.hbs
```diff
@@ -1,4 +1,4 @@
-{{#> nav nav--attribute=(concat 'id="' page--id '-primary-nav" aria-label="Grouped Nav Example"')}}
+{{#> nav nav--attribute=(concat 'id="' page--id '-primary-nav" aria-label="Local"')}}
```
mcoker:
```suggestion
{{#> nav nav--attribute=(concat 'id="' page--id '-primary-nav" aria-label="Global"')}}
```

mcoker:
Left one change suggestion.

Also. Can you just remove the nav component from this page component example, and replace the component with `pf-c-nav` text like the other examples?

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Page/examples/page-component-nav-vertical-example.hbs#L17-L19

I also wonder if we should add `aria-label` to the nav component examples? It's on some of them but not all of them. We can create a follow up issue to do that, too.

PR #1619, opened 2019-03-25
feat(clipboardcopy): add ellipsis when text overflows in the input box

### LINKED ISSUE #1618: form control should have ellipsis when text overflows
According to the image below when the text overflows, ellipsis should be shown:

my image

I believe it's easier to implement that using css's `text-overflow` than writing it in JavaScript.

boaz0:
Screenshot_2019-03-26 PatternFly 4(1)
Screenshot_2019-03-26 PatternFly 4

closes #1618

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -90,6 +90,9 @@ $pf-c-form-control__select--ViewBox: "320" !default;
 
   &:not(textarea) {
     height: var(--pf-c-form-control--Height);
+
+    // truncate overflow text
+    text-overflow: ellipsis;
   }
 
   &[readonly] {


christiemolloy:
@mcoker I think there was a discussion about adding the ellipsis to the actual form control component, would you be opposed to adding it there rather than in just this component.

boaz0:
@mcoker @christiemolloy I don't mind adding it to the form control component.

mcoker:
@christiemolloy ah, good question. I can see how that would be useful. What do you think @mcarrano @kybaker, is there value in adding a `pf-m-truncate` modifier to the form control component itself so a person can allow overflow text to truncate with an ellipsis on any form control element? Or do you think we should only apply the truncation in the copy to clipboard component?

mcarrano:
I think that any time we have a text field where the text does not fit, it should be truncated with an ellipses. If the field is editable, then I would expect the ellipses to hide when the user places focus in the field and the user can use the arrow keys to to scroll the contents in either direction.

Does this make sense?

mcoker:
@mcarrano yep. So it sounds like we should change the default text input form control component to always truncate with an ellipsis any time there is overflow text? And this will occur globally across patternfly if the contents of a text input are wider than the text input itself, not just in the copy clipboard component?

mcoker:
@boaz0 can you go ahead and do that? ^^ The form control text input should overflow with `...` any time there is overflow content, except when focused. And just verify it works in FF, Chrome, Edge, and Safari. Feel free to update the original issue and this PR or create a new one, whatever works best.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -90,6 +90,11 @@ $pf-c-form-control__select--ViewBox: "320" !default;
 
   &:not(textarea) {
     height: var(--pf-c-form-control--Height);
+
+    // truncate overflow text
+    overflow: hidden;
```
mcoker: in this case, it doesn't look like `overflow` or `white-space` are needed in the case of an `input`. I tested on Chrome/FF/Safari, and Firefox on Windows (`text-overflow` doesn't work on Edge).

boaz0: Um, from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#Browser_compatibility) Edge 12 supports this. What version of Edge are we supporting?

mcoker: Oh sorry, I don't mean the property itself, I mean just on `input`s. I found this issue - https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15550368/

And here's a screenshot

Screen Shot 2019-03-28 at 8 29 59 AM

boaz0:
Thanks @mcoker and @christiemolloy

PR #1621, opened 2019-03-25
docs(login): update required components for the login

### LINKED ISSUE #1572: Login component - which parts are required?
Most of the elements in the login component docs aren't marked as **Required**. We should go through the elements in that component and ask IXD/design which elements are required, and 1) update the docs table and 2) ensure the component layout doesn't break if optional elements are removed.

christiemolloy:
fix #1572

### DIFF
--- a/src/patternfly/components/Login/docs/code.md
+++ b/src/patternfly/components/Login/docs/code.md
@@ -6,17 +6,19 @@
 | `.pf-c-login__container` | `<div>` |  Positions the login component content. |
 | `.pf-c-login__header` | `<header>` |  Positions the login header. |
 | `.pf-c-login__header .pf-c-brand` | `<img>` |  Creates a brand image inside of login header. |
-| `.pf-c-login__main` | `<main>` |  Positions the login main area. |
-| `.pf-c-login__main-header` | `<header>` |  Creates the header of the main area. |
+| `.pf-c-login__main` | `<main>` |  Positions the login main area. **Required** |
+| `.pf-c-login__main-header` | `<header>` |  Creates the header of the main area. **Required** |
+| `.pf-c-login__main-header .pf-c-title` | `<h1>,<h2>,<h3>,<h4>,<h5>,<h6>` |  Creates a title in the main header area. **Required** |
 | `.pf-c-login__main-header-desc` | `<p>` |  Creates the description in the main area header. |
-| `.pf-c-login__main-header .pf-c-title` | `<h1>,<h2>,<h3>,<h4>,<h5>,<h6>` |  Creates a title in the main header area. |
 | `.pf-c-login__main-header .pf-c-dropdown` | `<div>` |  Creates a language selector dropdown in the main header area. |
-| `.pf-c-login__main-body` | `<div>` |  Creates the body of the main area. |
+| `.pf-c-login__main-body` | `<div>` |  Creates the body of the main area. **Required** |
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Login/docs/code.md
```diff
@@ -6,17 +6,19 @@
 | `.pf-c-login__container` | `<div>` |  Positions the login component content. |
```
mcoker: In other components, we add **Required** to all of the structural elements required to make the component, too. So `.pf-c-login`, `.pf-c-login__container`, `.pf-c-login__header`, etc

src/patternfly/components/Login/docs/code.md
```diff
@@ -6,17 +6,19 @@
...
-| `.pf-c-login__main-header` | `<header>` |  Creates the header of the main area. |
+| `.pf-c-login__main` | `<main>` |  Positions the login main area. **Required** |
+| `.pf-c-login__main-header` | `<header>` |  Creates the header of the main area. **Required** |
+| `.pf-c-login__main-header .pf-c-title` | `<h1>,<h2>,<h3>,<h4>,<h5>,<h6>` |  Creates a title in the main header area. **Required** |
 | `.pf-c-login__main-header-desc` | `<p>` |  Creates the description in the main area header. |
-| `.pf-c-login__main-header .pf-c-title` | `<h1>,<h2>,<h3>,<h4>,<h5>,<h6>` |  Creates a title in the main header area. |
 | `.pf-c-login__main-header .pf-c-dropdown` | `<div>` |  Creates a language selector dropdown in the main header area. |
-| `.pf-c-login__main-body` | `<div>` |  Creates the body of the main area. |
+| `.pf-c-login__main-body` | `<div>` |  Creates the body of the main area. **Required** |
 | `.pf-c-login__main-body .pf-c-form` | `<form>` |  Creates the login form in the main body area. |
```
mcoker: Is the form not required?

christiemolloy:
@mcoker updated

src/patternfly/components/Login/docs/code.md
```diff
@@ -2,21 +2,23 @@
 
 | Class | Applied To | Outcome |
 | -- | -- | -- |
-| `.pf-c-login` | `<div>` |  Initializes the login component. |
-| `.pf-c-login__container` | `<div>` |  Positions the login component content. |
-| `.pf-c-login__header` | `<header>` |  Positions the login header. |
+| `.pf-c-login` | `<div>` |  Initializes the login component. **Required**|
+| `.pf-c-login__container` | `<div>` |  Positions the login component content. **Required**|
+| `.pf-c-login__header` | `<header>` |  Positions the login header. **Required**|
 | `.pf-c-login__header .pf-c-brand` | `<img>` |  Creates a brand image inside of login header. |
```
mcoker: Currently the margin between the header and login box comes from `.pf-c-login__header .pf-c-brand`, but if the brand isn't required, what's acceptable to put in the header?

mcoker: If the header is required, maybe we should move the margin from the brand to `.pf-c-login__header`?

christiemolloy: IMO this gives no flexibility for the user to add different content other than an image. But I'm fine to make it required for now, since its dictating space.

mcoker: Cool. I'm not sure if we want someone to have anything other than an image in there? If we want to support text, along with moving the margin for the header from the image to the header itself, we might also want to recommend what size title component to use in that space.

PR #1629, opened 2019-03-25
fix(page): address hidden elements that are visible to screen readers

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1538


mcoker:
@jgiardino agreed! I created this issue a while back - https://github.com/patternfly/patternfly-next/issues/1430

Also, I believe we will have a new design for the header icons/menus soon, and assuming so, I'm hoping to solve that issue then.

jgiardino:
Oh, awesome! Thanks for the info!

PR #1642, opened 2019-03-26
fix(a11y): remove aria-required, remove required attr from labels

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/769


jgiardino:
These updates look great! 

Also, I didn't capture this in the issue, so not sure if it's in scope for this PR, but I'm noticing that we have instances of `required="true"`, which is incorrect according to [the spec for boolean attributes](http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes). Instead, it should be  `required`, `required=""` or `required="required"`.

Should I capture this in a separate issue, or would this be in scope for this PR?

mcoker:
@jgiardino thanks! Updated that in this PR. There was only one instance of `required="true"` on a form control element.  The rest are on `{{# form-label}}` as a handlebars attribute to create the red asterisk by the form label.

PR #1643, opened 2019-03-26
fix(table): remove unnecessary aria-label

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1637

### DIFF
--- a/src/patternfly/components/Table/docs/table-checkboxes-actions.md
+++ b/src/patternfly/components/Table/docs/table-checkboxes-actions.md
@@ -17,12 +17,12 @@ Note: When header cells are empty or they contain interactive elements, `<th>` s
 ### Tbody Accessibility
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `aria-labelledby="{title_cell_id}"` | `.pf-c-table__check input[type="checkbox"]` | Provides an accessible name for the checkbox. **Required** |
-| `id`  | title cell `<th>` > `*` | Provides an accessible description for the checkbox. **Required** |
+| `aria-labelledby="[title_cell_id]"` or `aria-label="[descriptive text]` | `.pf-c-table__check input[type="checkbox"]` | Provides an accessible name for the checkbox. **Required** |
+| `id` | title cell `<th>` > `*` | Provides an accessible description for the checkbox. **Required if using `aria-labelledby` for `.pf-c-table__check input[type="checkbox"]`** |
 
 
 ### Usage
 | Class | Applied To | Outcome |
 | -- | -- | -- |
-| `.pf-c-table__check`  | `<th>`, `<td>` | Initiates a check table cell. |
+| `.pf-c-table__check` | `<th>`, `<td>` | Initiates a check table cell. |
 | `.pf-c-table__action` | `<th>`, `<td>` | Initiates an action table cell. |
--- a/src/patternfly/components/Table/docs/table-checkboxes-actions.md
+++ b/src/patternfly/components/Table/docs/table-checkboxes-actions.md
@@ -12,18 +12,16 @@ Note: When header cells are empty or they contain interactive elements, `<th>` s
 ### Thead Accessibility
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `aria-label="[descriptive text]"` | `.pf-c-table__check` | Provides an accessible name for checkbox header cells. **Required** |
+| `aria-label="[descriptive text]"` | `.pf-c-table__check input[type="checkbox"]` | Provides an accessible name for the checkbox. **Required** |
 
 ### Tbody Accessibility
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `aria-labelledby="{title_cell_id} {element_id}"` | `.pf-c-check__input`, `.pf-c-button`  | Provides an accessible name for special cells containing interactive elements. **Required** |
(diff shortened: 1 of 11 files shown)


christiemolloy:
Looks like the sortable table in the table demos doesnt have an ID, no idea why accessibility didn't catch that, so the aria-labelledby etc aren't rendering properly. Also would be good to update the checks from `{{#> check input}} to <input type="check">` like you did in the expandable.

mcoker:
nice, I totally missed the other demo checkboxes. updated those and the sortable and simple table id's.

src/patternfly/demos/DataTable/data-table-expandable-table.hbs
```diff
@@ -77,7 +77,7 @@
     {{#> table-tr}}
       {{#> table-td table-td--toggle="true" table-td--button--attribute=(concat 'aria-labelledby="' table--id '-node2 expandable-toggle2" id="expandable-toggle2" aria-label="Details" aria-controls="' table--id '-content2"')}}{{/table-td}}
       {{#> table-td table-td--check="true"}}
-        {{#> check-input check-input--attribute=(concat 'type="checkbox" name="checkrow2" aria-labelledby="' table--id '-node2" aria-label="Details"')}}{{/check-input}}
+        <input type="checkbox" name="checkrow2" aria-labelledby="{{table--id}}-node2">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository Name"}}
         <div id="{{table--id}}-node2">Node 2<div><a href="#">siemur/test-space</a></div></div>
```
jgiardino: Not sure if this is relative to this PR, but the structure we have in the component example is different:
```
<div>
  <div id="simple-table-with-checkboxes-node2">Node 2</div>
  <a href="#">siemur/test-space</a>
</div>
```
In the component example, the html illustrates how the the id is only on the meaningful portion of the `<th>` if there are a lot of contents included there.

jgiardino:
The documentation for the checkbox currently states this:


-- | -- | --
`aria-label="[descriptive text]"` | `.pf-c-table__check` | Provides an accessible name for checkbox header cells. Required

Can we update this to include `aria-labelledby`? `aria-label` is still valid, so I would keep it, but also include what we show in our examples:
> `aria-label="[descriptive text]"` or `aria-labelledby="[id of descriptive text]"`

src/patternfly/demos/DataTable/data-table-sortable-table.hbs
```diff
@@ -141,7 +141,7 @@
 
     {{#> table-tr}}
       {{#> table-td table-td--check="true"}}
-        {{#> check-input check-input--attribute=(concat 'type="checkbox" name="checkrow5" aria-labelledby="' table--id '-node5"')}}{{/check-input}}
+        <input type="checkbox" name="checkrow4" aria-labelledby="{{table--id}}-node4">
```
jgiardino: `node4` should be `node5` here

src/patternfly/demos/DataTable/data-table-expandable-table.hbs
mcoker: thanks! updated the structure.

mcoker:
@jgiardino re: the documentation, it looks like `aria-labelledby` is in the docs, but under the `tbody` section. In the examples, we use `aria-label` in the `thead` and `aria-labelledby` in the `tbody`. I took a stab at updating the docs, and removing the reference to buttons since we don't use them in the examples (those are used in the expandable examples, which have their own docs). Let me know what you think.

src/patternfly/components/Table/docs/table-checkboxes-actions.md
```diff
@@ -12,18 +12,16 @@ Note: When header cells are empty or they contain interactive elements, `<th>` s
...
+| `aria-label="[descriptive text]"` | `.pf-c-table__check input[type="checkbox"]` | Provides an accessible name for the checkbox. **Required** |
 
 ### Tbody Accessibility
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `aria-labelledby="{title_cell_id} {element_id}"` | `.pf-c-check__input`, `.pf-c-button`  | Provides an accessible name for special cells containing interactive elements. **Required** |
-| `aria-label="[descriptive text]"`  | `.pf-c-check__input`, `.pf-c-button`  | Provides an accessible name for `checkboxes` and `buttons`. **Required** |
-| `id`  | title cell `<th>` > `<span>`, `.pf-c-check__input`, `.pf-c-button`  | Provides an accessible description for special cells. **Required** |
+| `aria-labelledby="{title_cell_id}"` | `.pf-c-table__check input[type="checkbox"]` | Provides an accessible name for the checkbox. **Required** |
+| `id`  | title cell `<th>` > `*` | Provides an accessible description for the checkbox. **Required** |
```
jgiardino: In chatting with @TheRealJon about [this implementation for react](https://github.com/patternfly/patternfly-react/issues/1460), he had mentioned that setting up the `aria-labelledby`/`id` relationship is more complex than just using `aria-label`. I would still keep our examples as they are, but since the react component will have a different (but also valid) method for providing the accessible name, should we show that variation as well?

If so, I would just add "or `aria-label="[descriptive text]"` for line 20. For line 21, I'm not sure if it's better to remove that and assume that they would know that `{title_cell_id}` means they need to add an `id` somewhere, or if we should say "**Required if** using `aria-labelledby` for `.pf-c-table__check input[type="checkbox"]`

And this is a total nit, but should we use consistent formatting for `[descriptive text]` and `{title_cell_id}`?

PR #1651, opened 2019-03-27
refactor(DataList): remove custom breakpoint

### LINKED ISSUE #1583: Update Data List Padding Breakpoint for Mobile
Currently the data list item has a set padding for mobile at a breakpoint of `max-width: $pf-data-list--breakpoint--mobile` this needs to break earlier to align with when the toolbar and page etc. break. This should be updated to `max-width: $pf-global--breakpoint--md` the associated variables should also change their naming. 

This will update #1401

christiemolloy:
fix #1583


christiemolloy:
Closing this issue in favor of #1658

PR #1653, opened 2019-03-27
chore(demos): remove pf-c-toolbar__pagination from toolbar in demos

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1652


mcoker:
@christiemolloy oh that's in https://github.com/patternfly/patternfly-next/pull/1627/

PR #1657, opened 2019-03-28
refactor(flex-layout): convert device breakpoints to grid breakpoints

### LINKED ISSUE #1654: Update `pf-l-flex` to use grid breakpoints

mattnolting:
fix #1654


mcoker:
looks good! could you also remove the breakpoints from [scss-variables.scss](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/sass-utilities/scss-variables.scss#L156-L159)?

PR #1671, opened 2019-04-01
 refactor(optionsmenu): refactor demos, per page text

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1666

@jgiardino should we review the `options-menu--label-text` and toggle `aria-label` for this component, or wait for the react implementation? I believe someone is working on the react options menu in this upcoming sprint.

### DIFF
--- a/src/patternfly/components/OptionsMenu/options-menu.scss
+++ b/src/patternfly/components/OptionsMenu/options-menu.scss
@@ -47,6 +47,7 @@
 
   // Menu Item
   --pf-c-options-menu__menu-item--Background: transparent;
+  --pf-c-options-menu__menu-item--FontSize: var(--pf-global--FontSize--md);
   --pf-c-options-menu__menu-item--PaddingTop: var(--pf-global--spacer--sm);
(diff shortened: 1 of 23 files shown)


src/patternfly/components/OptionsMenu/options-menu.scss
```diff
@@ -47,6 +47,7 @@
 
   // Menu Item
   --pf-c-options-menu__menu-item--Background: transparent;
+  --pf-c-options-menu__menu-item--FontSize: var(--pf-global--FontSize--md);
```
christiemolloy: Not sure if you need to define 16px font-size here as it is that by default.

mcoker: I talked to @mceledonia about it and we don't want the menu item to inherit a `font-size` from a parent. For example, the options menu in the pagination - the pagination component is 14px, but the options menu items should still be 16px. This is also true with the dropdown and select, which have `font-size` set on their menu-items, too.

christiemolloy: oh cool i was just wondering if it would get overwritten. lets stick with it.

PR #1672, opened 2019-04-01
feat(wizard): add support for finished state

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1670

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -6,7 +6,7 @@
   --pf-c-breadcrumb__item--MarginRight: var(--pf-global--spacer--sm);
 
   // Breadcrumb Divider
-  --pf-c-breadcrumb__item-divider--Color: var(--pf-global--BorderColor--100);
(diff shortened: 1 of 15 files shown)


mcoker:
cc @jschuler

mcoker:
@christiemolloy, @jknilans found a bug where if there is a lot of content in the main section, the footer isn't sticky and the content doesn't scroll. Also fixed that in this PR.

christiemolloy:
@kybaker you requested light grey arrows between list items in the breadcrumb component, do you want to update the color here to match:
Screen Shot 2019-04-01 at 3 32 13 PM

christiemolloy:
@mcoker do you think we need x and y scroll? I think that might be alot for the user on mobile, maybe we could just set it to y, or maybe ask design

mcoker:
@christiemolloy re: overflow, looks like we use `y: auto, x: hidden` almost everywhere else so I think that's ok here, too.

Chatted with @kybaker and we decided to go a darker shade, using `$pf-global--BorderColor--200` (`$pf-color-black-500/#8b8d8f`)

@kybaker can you confirm this looks OK? Updated Wizard and Breadcrumbs.

kybaker:
@christiemolloy @mcoker Looks good to me.

PR #1676, opened 2019-04-01
Add mixin for Scrollbar Hiding

### LINKED ISSUE #1641: Consider a mixin for scrollbar hiding
Currently, there are three vars we are adding to hide the scrollbar on Edge, Firefox and Chrome. We should add these as a mixin and apply it anywhere we are hiding them so that we cover all browsers.

 scrollbar-width: none; // hides scrollbars in Firefox 64 and up

-ms-overflow-style: -ms-autohiding-scrollbar; // auto hides scrollbars in Edge

::-webkit-scrollbar {
 display: none;
} // hides scrollbars in Chrome

christiemolloy:
fix #1641

### DIFF
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -91,13 +91,13 @@
 
 // Tab List
 .pf-c-tabs__list {
+  @include pf-overflow-hide-scroll;
+
   position: relative;
   display: flex;
   max-width: 100%;
   overflow-x: auto;
   -webkit-overflow-scrolling: touch;
-
-  @include pf-overflow-hide-scroll;
 
   // Right border on last tab
   .pf-c-tabs__item:last-of-type .pf-c-tabs__button::before {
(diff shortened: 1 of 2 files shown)


src/patternfly/sass-utilities/mixins.scss
```diff
@@ -143,3 +143,10 @@
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+@mixin pf-overflow-hide-scroll {
+  &::-webkit-scrollbar { display: none; } // hides scrollbars in Chrome
```
mcoker: Can you break that out into 3 lines?

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -97,6 +97,8 @@
   overflow-x: auto;
   -webkit-overflow-scrolling: touch;
 
+  @include pf-overflow-hide-scroll;
```
mcoker: one more thing, can you move this to the top of the selector? `@include` should come after variables but before properties.

christiemolloy: thanks @mcoker updated

PR #1681, opened 2019-04-02
refactor(Table): update code to style the table borders for mobile

### LINKED ISSUE #1665: update table to design for mobile
This is a redo of #1495 now that all other breaking changes have been merged for the table.

christiemolloy:
fix #1665


mcoker:
Just want to verify that we need all of these breakpoints, or if any of then could be combined?

Grid layout with thin border and default left padding:
Screen Shot 2019-04-02 at 5 32 57 PM

Reduced left padding:
Screen Shot 2019-04-02 at 5 33 04 PM

Thick borders:
Screen Shot 2019-04-02 at 5 33 07 PM

And there should be no border above the caption on mobile?
Screen Shot 2019-04-02 at 5 33 51 PM

In both of the expandable tables (regular and compact), there is no border on the last row until it breaks to a grid, then on mobile, if the last row is an expanded row, it has a thick bottom border. If it's not an expanded row, it has a thin bottom border. Every other table has a thin bottom border on the last row.

#### expanded row is last row

Screen Shot 2019-04-02 at 5 37 41 PM

#### non expanded row is last row
Screen Shot 2019-04-02 at 5 40 50 PM

christiemolloy:
@mcoker, re your comments: We need the grid layout with thin border default padding and left padding, because the padding breakpoint follows the page/toolbar.

I don’t think there should be a border above caption, there isn’t in the normal table.

I fixed the expandable table and now they all have a thin border on the bottom - thanks for catching this!

PR #1682, opened 2019-04-02
refactor(page): do not limit brand img width, use grid in header

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1678

I've been testing with different image sizes by using placehold.it as the src. For example `http://placehold.it/800x50` or `http://placehold.it/350x50`

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -10,7 +10,6 @@
   --pf-c-page__header-brand--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-page__header-brand--md--PaddingRight: var(--pf-global--spacer--xl);
   --pf-c-page__header-brand--md--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-page__header-brand--md--FlexBasis: #{pf-size-prem(290px)};
 
   @media (min-width: $pf-global--breakpoint--md) {
     --pf-c-page__header-brand--PaddingLeft: var(--pf-c-page__header-brand--md--PaddingLeft);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -145,26 +143,29 @@
...
 }
 
 // Brand
 .pf-c-page__header-brand {
+  grid-column: 1 / 2;
   padding-left: var(--pf-c-page__header-brand--PaddingLeft);
 
   @media (min-width: $pf-global--breakpoint--md) {
-    flex: 0 0 var(--pf-c-page__header-brand--md--FlexBasis); // Hardcode header brand width based on design spec, must match --pf-c-nav--Width
     align-self: stretch;
```
christiemolloy: do you still need this without flex?

christiemolloy:
Say you have a logo with a really large width, say 1000px, is Kyle fine, from a design perspective for it to be the only element in that header that scales down with size of the viewport. I wonder if we should put a max-width on it.

src/patternfly/components/Page/page.scss
mcoker: It's a grid property, too, though I don't think it's necessary. I think it was there since this was the element that had the semi-opaque dark background color.

mcoker:
@christiemolloy re:

> Say you have a logo with a really large width, say 1000px

Yep, I had originally applied some kind of reasonable max-width and @kybaker  said remove it entirely. Let them put a huge image in there if they want to, even though it will scale funny. They should be able to put any kind of image they want in there and figure out the right size based on how their header scales without us putting a limitation on the image width.

PR #1689, opened 2019-04-03
feat(page): add fill/no-fill modifiers for main-section elements

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1492


christiemolloy:
@mcoker what do you think about adding the pf-m-fill/pf-m-no-fill modifiers to the page component to demonstrate how they work?

mcoker:
@christiemolloy good idea, added.

PR #1702, opened 2019-04-04
 feat(dataList): modifier to handle alignment of data-cells

### LINKED ISSUE #1701: Data List alignment modifier
The Data List demos in Sketch have examples where the secondary content in a data-list__cell is positioned to the right of the container. We need to add a modifier that can handle this and then align it back left on mobile.

christiemolloy:
close #1701


mcoker:
Can you include the designs? I don't see them in the sketch cloud link.

christiemolloy:
Screen Shot 2019-04-04 at 5 18 54 PM

mcoker:
Ah, would this also apply as content that should align right by default, then left when it breaks to the mobile view?

Screen Shot 2019-04-04 at 4 35 10 PM

PR #1709, opened 2019-04-08
fix(wizard): no-padding modifier should modify main-body element

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1708

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -441,7 +441,9 @@
   word-break: break-word;
 
   &.pf-m-no-padding {
-    padding: 0;
+    .pf-c-wizard__main-body {
+      padding: 0;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -441,7 +441,9 @@
   word-break: break-word;
 
   &.pf-m-no-padding {
```
jschuler: Do you just want to move this into the pf-c-wizard__main-body class? It's okay if it's breaking

jschuler: lol nevermind was just thinking from react perspective, no more breaking for core :D

jschuler: Thanks for the quick turnaround!~

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -441,7 +441,9 @@
...
-    padding: 0;
+    .pf-c-wizard__main-body {
```
christiemolloy: Why did you put the modifier on `.pf-c-wizard__main` instead of `.pf-c-wizard__main-body`? Is it because of breaking changes? Maybe create a follow up issue?

PR #1716, opened 2019-04-10
feat(carddemo): add demo

### LINKED ISSUE #297: Card View Demo
Card view demo includes: 
- Navigation/header 
- page layout
- Toolbar 
- pagination 
- card examples as seen here: http://www.patternfly.org/pattern-library/content-views/card-view/#design

christiemolloy:
close #297 

@gdoyle1 
- should I add a logo for each card that relates to its project?
- in your example the card padding looks like 32px/some of the cards in Sketch are too @mcoker cards in pf are currently 24px. Do either of you know what it should be? IMO the content in the cards looks like it needs more padding.

@mcoker 
- not sure how to handle the spacing between the kebab and the check because the kebab has default padding around it and I don't want to add custom styles.
- not sure how to handle the content, the paragraph text should be darker and maybe the content shouldn't come with margin bottoms, not sure. there wasn't currently a modifier to remove the margin so I added one.


gdoyle1:
@christiemolloy 
- If you have time to add those logos, then yeah that'd be great! PF logo works too if you don't have time.
- I used 24px padding - @mceledonia recently did a lot of work figuring out the padding! Here's the spec of the card
Screen Shot 2019-04-11 at 9 16 10 AM

src/patternfly/demos/Card/card-content-1.hbs
```diff
@@ -0,0 +1,26 @@
+{{#> card-header}}
+  {{#> l-flex flex--modifier="pf-m-align-right"}}
```
mcoker: `pf-m-align-right` isn't doing anything. Also I would add `pf-m-align-center` to vertically align the row.

src/patternfly/demos/Card/card-content-1.hbs
```diff
@@ -0,0 +1,26 @@
...
+    {{#> l-flex flex--modifier="pf-m-grow"}}
+      {{#> brand brand--attribute='src="/assets/images/pf-logo-small.svg" alt="Logo"'}} 
+      {{/brand}}
+    {{/l-flex}}
+    {{#> dropdown id=(concat card--id "-dropdown-kebab-right-aligned") dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+    {{/dropdown}}
+    {{#> check}} 
+      {{#> check-input check-input--attribute=(concat 'id="' card--id '-check" name="' card--id '-check" aria-labelledby="' card--id '-check-label"')}}{{/check-input}}
```
mcoker: This should just be a checkbox. The check component is designed to be used with a label.

src/patternfly/demos/Card/card-content-1.hbs
```diff
@@ -0,0 +1,26 @@
...
+    {{/check}}
+  {{/l-flex}}
+{{/card-header}}
+{{#> card-body}}
+  {{#> content content--modifier="pf-m-no-margin"}}
+    <h4 id="{{concat card--id "-check-label"}}">Patternfly</h4>
+  {{/content}}
+  {{#> content}}
+    <small>Provided by Red Hat</small>
```
mcoker: Is this text appropriate for a `<small>` element or are you just using it to get the lighter text color?

> In HTML5, this element is repurposed to represent side-comments and small print, including copyright and legal text, independent of its styled presentation.

src/patternfly/demos/Card/card-content-1.hbs
```diff
@@ -0,0 +1,26 @@
...
+    {{/l-flex}}
+    {{#> dropdown id=(concat card--id "-dropdown-kebab-right-aligned") dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+    {{/dropdown}}
+    {{#> check}} 
+      {{#> check-input check-input--attribute=(concat 'id="' card--id '-check" name="' card--id '-check" aria-labelledby="' card--id '-check-label"')}}{{/check-input}}
+    {{/check}}
+  {{/l-flex}}
+{{/card-header}}
+{{#> card-body}}
+  {{#> content content--modifier="pf-m-no-margin"}}
```
mcoker: Are we adding this just for the purpose of this demo? Or is it a purposeful design decision to offer this? FWIW, you can get all of this spacing using a flex column layout instead. Just use a spacer between the "provided by" and paragraph text.

christiemolloy: I more so used content to get the sizes and coloring

src/patternfly/demos/Card/card-content-1.hbs
christiemolloy: awesome thanks!

src/patternfly/demos/Card/card-content-1.hbs
christiemolloy: Oh really, I didn't even realize the distinction until now, maybe we should clarify that in the docs

src/patternfly/demos/Card/card-content-1.hbs
mcoker: Ah ok. I think this is a great use of the element.

src/patternfly/demos/Card/card-content-2.hbs
```diff
@@ -0,0 +1,26 @@
...
+{{/card-header}}
+{{#> card-body}}
+  {{#> content content--modifier="pf-m-no-margin"}}
+    <h4 id="{{concat card--id "-check-label"}}">ActiveMQ</h4>
+  {{/content}}
+  {{#> content}}
+    <small>Provided by Red Hat</small>
+  {{/content}}
+  {{#> content content--modifier="pf-m-no-margin"}}
+    <small>
```
mcoker: I think this is an OK use of the `<small>` element, too. But we want the text to be black. Not sure how to do that.

christiemolloy:
@mcoker made updates

src/patternfly/demos/Card/card-content-5.hbs
```diff
@@ -0,0 +1,26 @@
+{{#> card-header}}
+  {{#> l-flex flex--modifier="pf-m-align-right"}}
+    {{#> l-flex flex--modifier="pf-m-grow"}}
+      {{#> brand brand--attribute='src="/assets/images/pf-logo-small.svg" alt="Logo"'}}
```
mcoker: We probably shouldn't use the brand component for this. The brand was specifically meant for the header, not just a generic image element, though I don't think we've said specifically it can't be used elsewhere. I think this would just be better as an `` or inline `<svg>`. Also I'm assuming a card like this, for amq, apache spark, etc, the patternfly logo would likely be replaced with the individual product logos.

mcoker:
@christiemolloy re: the remaining style issues, I think it depends on whether we need to add modifiers for text color (to adjust the description text color), add a layout or something that puts a kebab next to a check with minimal spacing  (maybe a generic layout, or something specific in the card header), maybe something for font-size (assuming the description text is better suited for a `<p>` than `<small>`). 

This was my original question when I saw the designs - do we actually want to build support for specific card layouts, or just use what we have to achieve the designs, and will we need to add any add'l support for layouts or utilities that we need for the demo that we don't currently have?

gdoyle1:
@christiemolloy Kebabs and checkboxes look good! 
- Can we change the text in the toolbar dropdown to "creator" and add a filter search with placeholder text "Find a project..."? 
- And also change the total items in the pagination to 20 so it makes more sense?

src/patternfly/demos/Card/card-demo-toolbar.hbs
```diff
@@ -0,0 +1,45 @@
...
+  {{#> toolbar-filter}}
+    {{#> select id=(concat toolbar--id '-select-checkbox') select--IsCheckboxSelect="true"}}
+      <i class="fas fa-filter" aria-hidden="true"></i>
+      Filter
+    {{/select}}
+  {{/toolbar-filter}}
+  {{#> toolbar-action-group}}
+    {{#> button button--modifier="pf-m-primary"}}
+      Create a project
+    {{/button}}
```
mcoker: This just disappears. Should we display the kebab so it's still available?

src/patternfly/components/Card/card.scss
```diff
@@ -13,6 +14,12 @@
   background-color: var(--pf-c-card--BackgroundColor);
   box-shadow: var(--pf-c-card--BoxShadow);
 
+  &.pf-m-hover {
+    &:hover {
+      box-shadow: var(--pf-c-card--m-hover--hover--BoxShadow);
```
mcoker: Just want to verify this was an approved add from design? The shadow seems to big to me personally, like we need something between `md` and `lg` box shadows.

src/patternfly/components/Page/page.scss
```diff
@@ -408,6 +408,10 @@
     --pf-c-page__main-section--BackgroundColor: var(--pf-c-page__main-section--m-dark-200--BackgroundColor);
   }
 
+  &.pf-m-no-padding {
```
mcoker: think it's worth noting this in the next demo?

src/patternfly/components/Card/card.scss
mcoker: And assuming so, we should add it to the documentation.

mcoker: and... potentially call it something else. we use `pf-m-hover` to force the hover state in components. Maybe `pf-m-hoverable`?

src/patternfly/components/Page/page.scss
mcoker: also add it to the docs.

src/patternfly/demos/Card/card-content-2.hbs
```diff
@@ -0,0 +1,24 @@
...
+    {{#> flex-item flex-item--modifier="pf-u-mr-0"}}
+      {{#> dropdown id=(concat card--id "-dropdown-kebab-right-aligned") dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+      {{/dropdown}}
+    {{/flex-item}}
+     <input type="checkbox" id="{{card--id}}-check" name="{{card--id}}-check" aria-labelledby="{{card--id}}-check-label">
+  {{/l-flex}}
+{{/card-header}}
+{{#> card-body}}
+  {{#> l-flex flex--modifier="pf-m-column"}}
+  <p style="font-size:18px" id="{{concat card--id "-check-label"}}">Patternfly</p>
```
mcoker: why not use a title component?

src/patternfly/demos/Card/card-content-2.hbs
```diff
@@ -0,0 +1,24 @@
...
+  <p style="font-size:14px;color:#8B8D8F;">Provided by Red Hat</p>
```
mcoker: I think this was fine as a `<small>` used inside of `pf-c-content`

src/patternfly/demos/Card/card-content-2.hbs
```diff
@@ -0,0 +1,24 @@
...
+  {{#> flex-item flex-item--modifier="pf-u-mt-md"}}
+    <p style="font-size:14px">
```
mcoker: what if you had for this section...

```
{{#> title}} Patternfly{{/title}}
{{#> content}}
    <small>provided by Red Hat</small>
    <p>The ActiveMQ component allows messages to be sent to a JMS Queue or Topic; or messages to be consumed from a JMS Queue or Topic using Apache ActiveMQ.</p>
{{/content}}
```

would that create the same spacing?

mcoker: We should also probably talk about whether we want to write styles for demos. I think a stylesheet for the demo is better than inline styles, if we're going to start writing custom CSS in demos that we don't support using default layouts/components.

Another option is to keep our demos within the bounds of what our components and layouts provide. If we don't currently have support for styles needed in a demo, we could 1) simplify the demo, or 2) expand our components/layouts so that we don't have to write custom CSS.

src/patternfly/demos/Card/card-content-3.hbs
```diff
@@ -0,0 +1,24 @@
+{{#> card-header}}
+  {{#> l-flex flex--modifier="pf-m-align-center"}}
+    {{#> l-flex flex--modifier="pf-m-grow"}}
```
mcoker: You can simplify this section quite a bit for the layouts in these demos.

```
<div class="pf-c-card__header pf-c-title pf-m-lg pf-l-flex pf-m-space-items-none pf-m-align-center">
    <img class="pf-m-align-left" src="/assets/images/camel-spark_200x150.png" width="60px" alt="Logo">
    <div class="pf-c-dropdown">...</div>
    <input type="checkbox" id="card-3-check" name="card-3-check" aria-labelledby="card-3-check-label">
</div>
```

Or you could add a bit more HTML and do it like

```
<div class="pf-c-card__header pf-c-title pf-m-lg">
    <div class="pf-l-flex pf-m-space-items-none pf-m-align-center">
        <img class="pf-m-align-left" src="/assets/images/camel-spark_200x150.png" width="60px" alt="Logo">
        <div class="pf-c-dropdown">...</div>
       <input type="checkbox" id="card-3-check" name="card-3-check" aria-labelledby="card-3-check-label">
    </div>
</div>
```

src/patternfly/demos/Card/card-demo-toolbar.hbs
christiemolloy: Yeah good idea, wasn't in design, but makes sense with the current toolbar

src/patternfly/demos/Card/card-demo-template-gallery.hbs
```diff
@@ -0,0 +1,52 @@
+{{#> gallery gallery--modifier="pf-m-gutter"}}
+  {{#> gallery-item}}
```
mcoker: You also don't have to use a `pf-c-gallery__item` here. You can just use a `pf-c-card`.

src/patternfly/components/Card/card.scss
christiemolloy: - the shadow has been approved by design, @gdoyle1 want to make sure you're fine with the lg size box shadow.  I can test out the medium one also?
- agree with the hoverable name.

src/patternfly/demos/Card/examples/card-demo-example.hbs
```diff
@@ -0,0 +1,70 @@
...
+        {{#> nav-item}}
+          {{#> nav-link nav-link--href="#"}}
+            Server
+          {{/nav-link}}
+        {{/nav-item}}
+      {{/nav-list}}
+    {{/nav}}
+  {{/page-sidebar}}
+  {{#> page-main}}
+    {{#> page-template-breadcrumb}}
```
mcoker: The design shows tertiary nav instead of breadcrumbs, but I wonder if we should have either in these demos and just keep them simple? It's less things to follow up and update if we update breadcrumbs, tertiary nav, or the containers for them in the page component.

src/patternfly/components/Page/page.scss
christiemolloy: Added to the docs, what does your first comment mean?

src/patternfly/demos/Card/card-demo-template-gallery.hbs
christiemolloy: Oh yes nice that allows them to all be the same size according to the tallest card in the row.

src/patternfly/demos/Card/examples/card-demo-example.hbs
christiemolloy: I definitely agree going to remove it

src/patternfly/demos/Card/card-content-2.hbs
christiemolloy: Agreed I think it would be a great topic for CSS Army. To me, the demos showcase how our components fit together, and shouldn't need to add custom styles. I think custom styles are obvious and infinite, and we shouldn't need to worry about them in demos. So unless the design is necessary to the component, we don't add custom styles.

src/patternfly/components/Page/page.scss
mcoker: Oh this is a new feature of the page component, but is being bundled in this PR to create a card demo. Wondering if we should mention in the next PF demo that you can add `pf-m-no-padding` to the page sections. Up to you, it's a small thing so it doesn't matter to me. Just wanted to mention it.

src/patternfly/demos/Card/card-content-2.hbs
mcoker: Agreed. I had talked with @mceledonia about creating these demos and we mentioned creating more complex demos that contain custom CSS to show what a user could do (with these different styles in the cards), but I don't know if that's a good idea or not. I certainly like the idea of just building what we can without the need for custom CSS.

src/patternfly/components/Page/page.scss
christiemolloy: oh gotcha!

src/patternfly/demos/Card/card-content-2.hbs
christiemolloy: I also think that if we add custom CSS, users will think that those styles are bundled with PF and it will create confusion as to what Patternfly really offers

mcoker: this^^. That's also why we don't include any JS in our demos. We used to include a little just for basic functionality but we don't want users to think we're suggesting or support that.

christiemolloy:
@mcoker I updated this card based on the new card updates, took out all of the extra classes except I'm not sure what to do with the grey line of text under the header. 
It could use the content class and then remove the margin-bottom using a utility.
`{{#> content}}
    <small class="pf-u-mb-0">Provided by Red Hat </small>
{{/content}}`

Or use a regular `<p>` tag and set the size and color to match the designs:

`<p style="color: #000; font-size=14px";"> Provided by Red Hat </small>`

mcoker:
Can you think of a reason why we shouldn't remove the margin-bottom from `<small>` and add `small` to the list of elements in this block https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Content/content.scss#L117-L125. That would keep `<small>` from having a bottom margin if it's the only child. I'm also inclined to do the same thing to `<h1-6>`. What do you think? I think we could safely change `<small>` now and just position the issue as a bug. Changing the headings might be more invasive, so I'd consider that a breaking change.

christiemolloy:
@mcoker I think its a great idea to make all those elements not have a bottom margin if they're the last child, also goes with the rule that our components touch all edges of their parent. I can make the issues and work on the one for small now :)

christiemolloy:
@mcoker updated

christiemolloy:
@mcoker great point, I started that way because I was going to repeat the card information... but just moved all that content into the one file now

PR #1717, opened 2019-04-10
feat(select): add selected state

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1410

### DIFF
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -80,6 +80,7 @@
   // Menu Item
   --pf-c-select__menu-item--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-select__menu-item--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-select__menu-item--m-selected--PaddingRight: var(--pf-global--spacer--2xl);
   --pf-c-select__menu-item--PaddingBottom: var(--pf-global--spacer--sm);
   --pf-c-select__menu-item--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-select__menu-item--FontSize: var(--pf-global--FontSize--md);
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -356,6 +363,20 @@
     pointer-events: none;
     background-color: var(--pf-c-select__menu-item--disabled--BackgroundColor);
   }
+
+  &.pf-m-selected {
```
christiemolloy: I feel like this is where the focus styles should go? i.e. the background color change

src/patternfly/components/Select/select.scss
```diff
@@ -356,6 +363,20 @@
...
+    --pf-c-select__menu-item--PaddingRight: var(--pf-c-select__menu-item--m-selected--PaddingRight);
+  }
+}
+
+.pf-c-select__menu-item-icon {
+  position: absolute;
+  top: 50%;
+  right: var(--pf-c-select__menu-item-icon--Right);
+  margin-left: auto;
```
christiemolloy: do we need margin-left: auto?

src/patternfly/components/Select/select.scss
```diff
@@ -356,6 +363,20 @@
...
   }
+
+  &.pf-m-selected {
+    --pf-c-select__menu-item--PaddingRight: var(--pf-c-select__menu-item--m-selected--PaddingRight);
+  }
+}
+
+.pf-c-select__menu-item-icon {
+  position: absolute;
+  top: 50%;
```
christiemolloy: I wonder if, seeing as you have a var for "right" if we should have one for top also even though its a %

mcoker: We can. We would also need one for `translateY()` since those 2 properties do the vertical alignment.

mcoker:
@christiemolloy re: "Isn't this the selected state?" I believe the selected indicator is just a check mark, like the Options Menu. @mcarrano or @mceledonia can you confirm?

mceledonia:
@mcoker @christiemolloy Yeah let's go with just the checkmark for selected everywhere. The gray background should be hover only if I'm remembering correctly.

christiemolloy:
@mcoker  re: @mceledonia 's comment, I think you can remove the focus states on the menu item from styles and docs

mceledonia:
Apologies, hover and focus! @christiemolloy

PR #1722, opened 2019-04-11
fix(tabs): show tabs as links in <nav> example

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1585

### DIFF
--- a/src/patternfly/components/Tabs/docs/tabs-nav.md
+++ b/src/patternfly/components/Tabs/docs/tabs-nav.md
@@ -8,3 +8,9 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 | -- | -- | -- |
 | `aria-label="Local"` | `nav.pf-c-tabs` | Gives the `<nav>` element an accessible name. **Required when `.pf-c-tabs` is used with `<nav>`**
 | `aria-label="Local secondary"` | `nav.pf-c-tabs.pf-m-tabs-secondary` | Gives the `<nav>` element an accessible name. **Required when `.pf-c-tabs.pf-m-tabs-secondary` is used with `<nav>`**
+
+### Usage
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-tabs__button` | `<a>` | Creates a tab link. **Required** |
--- a/src/patternfly/components/Tabs/docs/tabs-nav.md
+++ b/src/patternfly/components/Tabs/docs/tabs-nav.md
@@ -13,4 +13,4 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `.pf-c-tabs__button` | `<a>`, `<button>` | Creates a tab link/button. **Required** |
+| `.pf-c-tabs__button` | `<a>`, `<button>` | Creates a tab link/button. **Note: `<a>` should be used when there is a URL associated with the tab, and should only be used when `.pf-c-tabs` is used with `<nav>`.** **Required** |
--- a/src/patternfly/components/Tabs/docs/tabs-nav.md
+++ b/src/patternfly/components/Tabs/docs/tabs-nav.md
@@ -13,4 +13,4 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 
 | Attribute | Applied To | Outcome |
 | -- | -- | -- |
-| `.pf-c-tabs__button` | `<a>` | Creates a tab link. **Required** |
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Tabs/docs/tabs-nav.md
```diff
@@ -8,3 +8,9 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 | -- | -- | -- |
 | `aria-label="Local"` | `nav.pf-c-tabs` | Gives the `<nav>` element an accessible name. **Required when `.pf-c-tabs` is used with `<nav>`**
 | `aria-label="Local secondary"` | `nav.pf-c-tabs.pf-m-tabs-secondary` | Gives the `<nav>` element an accessible name. **Required when `.pf-c-tabs.pf-m-tabs-secondary` is used with `<nav>`**
+
+### Usage
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-tabs__button` | `<a>` | Creates a tab link. **Required** |
```
jgiardino: Since this class can be used on either `<button>` or `<a>`, it seems like we should mention `<button>` here.

I'm just now noticing we don't have any usage documentaton on Tabs. Should we create an issue to add all the other classes here, too?

jgiardino:
These updates look good! I just had some comments/questions about the documentation.

src/patternfly/components/Tabs/docs/tabs-nav.md
mcoker: Hmm... do you see the docs for the other sections? I was just commenting on the deviation here. The docs for the rest of the elements Or can/should you use `<button>` with `<nav>`, too? 

I just noticed we have `.pf-c-tabs | <div>, <nav>` in the section above. Do you think I should pull `<nav>` out of that, and move it down to the nav example? Or leave `<nav>` and add `<a>` to the elements `.pf-c-tabs__button` applies to in the first example's docs?

jgiardino: Doh, I completely somehow missed the other doc sections. 😞 

It might be possible that consumers would need to use `<button>` due to not having urls defined for each tab, but want the benefits of using `<nav>`. So I think we'd still list `<button>` here, too.

mcoker: What do you think about `.pf-c-tabs | <div>, <nav>` in the first section? Should I remove `<nav>` so that's only in the nav example's docs?

jgiardino: I think it's fine as it is. On one hand, I don't think we need to show every possible html element variation that's possible, and can have the documentation mention elements that could be used but not used in an example. 

But for the `<a>`/`<button>` row for the `<nav>` example, I'm inclined to say that the `<a>` is only relevant to the `<nav>` use case, and therefore, it makes sense to call that out with this example specifically, and not in the global documentation.

src/patternfly/components/Tabs/docs/tabs-nav.md
```diff
@@ -8,3 +8,9 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 | -- | -- | -- |
 | `aria-label="Local"` | `nav.pf-c-tabs` | Gives the `<nav>` element an accessible name. **Required when `.pf-c-tabs` is used with `<nav>`**
 | `aria-label="Local secondary"` | `nav.pf-c-tabs.pf-m-tabs-secondary` | Gives the `<nav>` element an accessible name. **Required when `.pf-c-tabs.pf-m-tabs-secondary` is used with `<nav>`**
+
+### Usage
+
+| Attribute | Applied To | Outcome |
+| -- | -- | -- |
+| `.pf-c-tabs__button` | `<a>`, `<button>` | Creates a tab link/button. **Required** |
```
christiemolloy: Can tab buttons be links only when tabs use? If so I think you should update the docs here, I don't think just including these docs in the nav example is enough

src/patternfly/components/Tabs/examples/tabs-nav.hbs
```diff
@@ -59,22 +59,22 @@
 
 <br><br>
 
-{{#> tabs tabs--id="primary-nav" tabs--type="nav" tabs--attribute='aria-label="Local"'}}
+{{#> tabs tabs--id="primary-nav" tabs--type="nav" tabs--attribute='aria-label="Local"' tabs-button--type="a"}}
```
christiemolloy: If my previous comment applies (if tabs = nav, tabs button = a) I wonder if tabs--type is enough and we can assign the tabs-button--type through there

PR #1726, opened 2019-04-12
feat(alert): add live region support for toast alerts

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1375

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -163,20 +163,23 @@
   }
 
   &.pf-m-live {
+    opacity: 0;
+    animation: pf-show-live 0s 1s forwards;
+
     .pf-c-alert__content {
-      animation: pf-load-live 1s 1;
+      visibility: hidden;
+      animation: pf-show-content 0s 1s forwards;
     }
   }
 
-  @keyframes pf-load-live {
-    0%,
-    95% {
-      display: none;
-      visibility: hidden;
-    }
-
+  @keyframes pf-show-live {
     100% {
-      display: block;
+      opacity: 1;
+    }
+  }
+
+  @keyframes pf-show-content {
+    100% {
       visibility: visible;
     }
   }
@@ -189,6 +192,10 @@
   font-size: var(--pf-c-alert__icon--FontSize);
   color: var(--pf-c-alert__icon--Color);
   background-color: var(--pf-c-alert__icon--BackgroundColor);
+}
+
+.pf-c-alert__content {
+  display: contents;
 }
 
 .pf-c-alert__title {
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -161,6 +161,25 @@
       font-size: var(--pf-c-alert__m-inline__icon--FontSize);
     }
   }
+
+  &.pf-m-live {
+    .pf-c-alert__content {
+      animation: pf-load-live 1s 1;
+    }
+  }
+
+  @keyframes pf-load-live {
+    0%,
+    95% {
+      display: none;
+      visibility: hidden;
+    }
+
+    100% {
+      display: block;
+      visibility: visible;
+    }
+  }
 }
 
 .pf-c-alert__icon {
--- a/src/patternfly/components/Alert/alert-content.hbs
+++ b/src/patternfly/components/Alert/alert-content.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-alert__content{{#if alert-content--modifier}} {{alert-content--modifier}}{{/if}}"
+  {{#if alert-content--attribute}}
+    {{{alert-content--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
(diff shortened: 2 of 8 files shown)


mcoker:
@matthewcarleton astutely pointed out that the current layout supports the description content area wrapping underneath actions, and this new layout does not. Leaving a note to address in this PR.

Screen Shot 2019-04-15 at 12 58 12 PM

Screen Shot 2019-04-15 at 1 04 10 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -88,6 +88,25 @@
     --pf-c-alert__icon--BackgroundColor: var(--pf-c-alert--m-info__icon--BackgroundColor);
     --pf-c-alert__title--Color: var(--pf-c-alert--m-info__title--Color);
   }
+
+  &.pf-m-live {
+    .pf-c-alert__content {
+      animation: pf-load-live 1s 1;
```
jgiardino: Just to confirm, I updated my [codepen POC](https://codepen.io/jenngiardino/full/PLyQGm) and tested in all screen readers to make sure a shorter time would work. 

I used `.25s` for the animation property, and `80%` instead of `95%` which should be 200ms.

I also played around with the css animation to see if there's a way to hide the wrapper visually but not from assistive technology and have that appear at the same time as the contents. But I couldn't find anything that worked.

mcoker:
`display: contents;` on `.pf-c-alert__content` looks like it would work, but it doesn't work in Edge even though they have partial support.

Another option could be to modify the HTML structure in the `.pf-m-live` variation of the toast alert, and leave support for the existing structure so not to introduce a breaking change, then clean that up during our next breaking change release.

@jgiardino what do you think?

jgiardino:
To recap our discussion from today, we talked about using `display: contents` to get the desired layout where supported, and live with the layout as shown above in browsers where not supported (i.e. Edge).

And we would modify the animation to have the Alert block and Alert text display at the same time. I updated the same [codepen](https://codepen.io/jenngiardino/pen/PLyQGm?editors=0100) I shared earlier with a POC that I tested. Opacity does work in this case, but so do the screen-reader-only styles. I'd suggest using the sr-only styles for consistency with the method we use to hide things visually but not hide things from screen readers. 

I also put a comment in https://github.com/patternfly/patternfly-next/issues/938 to capture that the Alert is an ideal candidate to start off with for handling animation.

src/patternfly/components/Alert/alert.scss
jgiardino: Can we adjust the timing so that the delay is equivalent to 200 ms (e.g. `.25s` and `80%`)

> I also played around with the css animation to see if there's a way to hide the wrapper visually but not from assistive technology and have that appear at the same time as the contents. But I couldn't find anything that worked.

I was missing the `s` in `.25s` 🤦‍♀. This is totally doable.

mcoker:
Thanks @jgiardino! In your codepen, I don't see the "visible" styles (to undo the "sr-only" styles) on `@keyframes display_alert`. 

I took a stab at rewriting the animation time to be easier to manage by starting it immediately `animation-start: 0s`, and setting `animation-delay: .2s` so it starts at 2s, and defining the initial hidden styles in the elements' CSS, then only applying the visible styles when the animation starts. I ran into a couple of problems with this though.

You can't animate `display`, so I wasn't able to animate `visibility: hidden; display: none;` on `.pf-c-alert__content` - only able to animate `visibility: hidden;`. I'm also assuming `display` was never being animated before in the previous way of writing the animation, either, since it doesn't animate?

Also if you animate `position: fixed;` to `position: static;` in safari, this happens.

Screen Shot 2019-05-23 at 4 54 26 PM

So instead, I'm simply animating `opacity` on the alert component, and `visibility` on the `pf-c-alert__content` element. Would you mind letting me know if this works, and if so, what you think of that? If not, I'll just go back to what you've already tested!

jgiardino:
> In your codepen, I don't see the "visible" styles (to undo the "sr-only" styles) on @keyframes display_alert.

I could be wrong about this, but I was expecting that the styles applied to the element outside the context of the animation are the "visible" styles. And that the animation is only applying styles to hide the element, and when their time is up, they're no longer applied, and the styles applied outside the animation are?

> I took a stab at rewriting the animation time to be easier to manage by starting it immediately animation-start: 0s, and setting animation-delay: .2s so it starts at 2s, and defining the initial hidden styles in the elements' CSS, then only applying the visible styles when the animation starts. I ran into a couple of problems with this though.

That's an interesting approach. Maybe this is a larger conversation about how we want to move forward with animation? I can see value in having the styles that are applied to the element being the styles that should be applied without any animation (e.g. how this thing looks on the page), and the animation styles that get defined are the styles used to transition that element on or off the page. But I also like the use of delay in your example. Maybe it's possible to merge these ideas?

> You can't animate display, so I wasn't able to animate visibility: hidden; display: none; on .pf-c-alert__content - only able to animate visibility: hidden;. I'm also assuming display was never being animated before in the previous way of writing the animation, either, since it doesn't animate?

It's very possible that `visibility: hidden` is enough for screen readers to not see the contents.

That said, I did test the updated preview in this PR, but to really test this solution, I will need something like a codepen that's using these styles and has a js timeout on displaying alerts. I have not had time to create that yet.

jgiardino:
@mcoker I took a codepen that you started and forked it to include a timeout so that one of the alerts displays after the page has loaded (i.e. to simulate an actual alert scenario): https://codepen.io/jenngiardino/full/zQgBxx

I only got as far as testing it in VO, and it isn't working there. I'm kinda wondering if maybe what they say where css animation doesn't affect `display` is untrue for assistive technologies. I tried changing the css that you had specified so that it incorporated `display: none`. But with the way your animation is set up, the properties that are statically applied to the element outside of the animation are to hide the element, and then the animation dynamically adds properties to show the element. If I add `display: none` to the element, then `display:block` to the animation, the contents never become visible on screen, and these contents are not announced either.

I can't figure out why the animation in my original codepen works, and this one doesn't (e.g. is it the delay, or whether styles are defined statically on the element vs dynamically within the animation?). Let me know if you find a variation that works in VO, and then I can test those in other screen readers.

PR #1727, opened 2019-04-12
feat(button): add inline variation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1182

### DIFF
--- a/src/patternfly/components/Button/examples/button-types-example.hbs
+++ b/src/patternfly/components/Button/examples/button-types-example.hbs
@@ -1,19 +1,22 @@
 {{#> button button--modifier="pf-m-primary"}}
-    Primary
+  Primary
 {{/button}}
 {{#> button button--modifier="pf-m-secondary"}}
-    Secondary
+  Secondary
 {{/button}}
 {{#> button button--modifier="pf-m-tertiary"}}
-    Tertiary
+  Tertiary
 {{/button}}
 {{#> button button--modifier="pf-m-danger"}}
-    Danger
+  Danger
 {{/button}}
 {{#> button button--modifier="pf-m-link"}}
-    {{> button-icon button-icon--type="plus-circle" button-icon--modifier="pf-c-button__icon"}}
-    Link Button
+  {{> button-icon button-icon--type="plus-circle" button-icon--modifier="pf-c-button__icon"}}
+  Link Button
 {{/button}}
 {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Remove"'}}
-    {{> button-icon button-icon--type="times"}}
+  {{> button-icon button-icon--type="times"}}
+{{/button}}
+{{#> button button--modifier="pf-m-inline pf-m-link"}}
+  Inline Button
 {{/button}}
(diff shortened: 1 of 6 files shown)


jgiardino:
Also looks good to me! Just one question about hover styles. Should a button that displays within a paragraph of text have the same hover effects as a link that displays within a paragraph of text? Currently, a link will include an underline on hover. cc'ing @mceledonia for his input, too.

jgiardino:
I think you captured the use case correctly, but just to confirm...

The use case for this is when we want to include an interactive element within a chunk of text, but want to use a `<button>` instead of a link (`<a>`) due to not having a url to define for the element. 

And below is how it would look, in the context of our Content component (i.e. identical to the link, except for hover at the moment):
    image

> do you know if the current design shown here has any accessibility concerns?

The best resource I can find on this is this technique: https://www.w3.org/TR/WCAG20-TECHS/G183.html

Essentially, if color alone is used to differentiate a link or control from surrounding text, then they recommend providing some additional visual cue on hover and focus. With focus we have the default behavior, but for hover, nothing changes.

I know we have had conversations in the past about users knowing when something is a link vs a button, and the opposing arguments are that users are indifferent and don't care if what they're about to click on is a button or link VS it should be clear if the control they are about to click on is a link or a button so they know what to expect. Until we have user research to verify which approach is accurate, I'm putting all my money on the former, that they don't care. Additionally, in the reverse case where links are styled like buttons, we don't do anything to show that it's really a link that looks like a button.

Based on the recommendation to show a change on hover, my assumption about user expectation, and the behavior of links styled as buttons, my suggestion would be to have the buttons styled as links to have the same hover effect as a link. 

I'm totally open to treating them differently, though, as long as we apply a similar guideline to the links as buttons. And maybe we wait to treat them differently until we have had a chance to do some user research to know for sure if it's a requirement to differentiate links and buttons?

mcoker:
> my suggestion would be to have the buttons styled as links to have the same hover effect as a link.

+1

mceledonia:
@jgiardino @mcoker Thanks for the explanation, I understand this a lot more clearly now. I agree that the majority won't care (until proven otherwise :) and that aligning with the link hover makes sense here, especially given how the context of being within a block of text impacts the accessibility needs.

mcoker:
@jgiardino just to confirm, are you referring to both the normal link button and inline link button variations? Or just the inline link?

jgiardino:
> are you referring to both the normal link button and inline link button variations? Or just the inline link?

I was only referring to the inline link button variation. The recommendations I mentioned above aren't  applicable to the normal link buttons, so I would defer to others on what we want there.

mcoker:
@jgiardino @mceledonia updated.

src/patternfly/components/Button/examples/button-types-example.hbs
```diff
@@ -1,19 +1,22 @@
...
-    {{> button-icon button-icon--type="plus-circle" button-icon--modifier="pf-c-button__icon"}}
-    Link Button
+  {{> button-icon button-icon--type="plus-circle" button-icon--modifier="pf-c-button__icon"}}
+  Link Button
 {{/button}}
 {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Remove"'}}
-    {{> button-icon button-icon--type="times"}}
+  {{> button-icon button-icon--type="times"}}
+{{/button}}
+{{#> button button--modifier="pf-m-inline pf-m-link"}}
```
christiemolloy: I wonder if this should be called "Inline Link Button"

mcoker: updated.

PR #1732, opened 2019-04-15
fix(card): update font sizes

### LINKED ISSUE #1729: Card body/footer text size should be 14px by default and header to 16px
Card body and footer text size should be 14px by default and header should be 16px.

Related to #1716

christiemolloy:
Close #1729

### DIFF
--- a/src/patternfly/components/Card/card-header.hbs
+++ b/src/patternfly/components/Card/card-header.hbs
@@ -1,4 +1,4 @@
-<div class="pf-c-card__header pf-c-title pf-m-lg{{#if card-header--modifier}} {{card-header--modifier}}{{/if}}"
+<div class="pf-c-card__header pf-c-title pf-m-md{{#if card-header--modifier}} {{card-header--modifier}}{{/if}}"
   {{#if card-header--attribute}}
     {{{card-header--attribute}}}
   {{/if}}>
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Card/card-header.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="pf-c-card__header pf-c-title pf-m-lg{{#if card-header--modifier}} {{card-header--modifier}}{{/if}}"
+<div class="pf-c-card__header pf-c-title pf-m-md{{#if card-header--modifier}} {{card-header--modifier}}{{/if}}"
```
mcoker: Has @mceledonia looked at this, too? This just changed recently as a result of this conversation with the Integreatly team - https://github.com/patternfly/patternfly-next/issues/626#issuecomment-466500428

src/patternfly/components/Card/card.scss
```diff
@@ -27,6 +29,11 @@
   }
 }
 
+.pf-c-card__body,
+.pf-c-card__footer {
```
mcoker: Can you separate this out to separate vars?

mcoker: Looks like you added `--pf-c-card__footer--FontSize`, just aren't using it.

christiemolloy:
@mcoker updated

PR #1733, opened 2019-04-15
fix(page): add no padding modifier to main section

### LINKED ISSUE #1730: Page Main Section should have a modifier for no padding
There is a modifier for no-padding on mobile, but we need a modifier for no padding at all times when components like the Toolbar have their own padding.

Related to #1716

christiemolloy:
Closes #1730


src/patternfly/components/Page/examples/index.js
```diff
@@ -33,6 +36,9 @@ export default () => {
       >
         {pageComponentFillExample}
       </Example>
+      <Example heading="Page Component Main Section Modifiers" handlebars={PageComponentMainSectionModifiersRaw}>
```
mcoker: Can you add "Padding" so it's "Page Component Main Section Padding Modifiers", since there are other main section modifiers available that aren't in this example?

christiemolloy: Should we consider doing an example for those other modifiers then? I think its worth it.

PR #1734, opened 2019-04-15
fix(card): add hoverable shadow

### LINKED ISSUE #1728: Card needs to have a box-shadow on hover (modifier)
Add a modifier to the card component so that on hover it has a lg size box shadow.

Part of #1716

christiemolloy:
Close #1728


src/patternfly/components/Card/examples/card-hover-example.hbs
```diff
@@ -0,0 +1,11 @@
+{{#> card card--modifier="pf-m-hoverable"}}
+  {{#> card-body}}
```
mcoker: this should be `{{#> card-header}}`.

christiemolloy:
@mcoker updated

PR #1746, opened 2019-04-17
feat(wizard): add modifier for compact nav width

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1737

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -8,6 +8,7 @@
   --pf-c-wizard--lg--MaxHeight: calc(100% - (var(--pf-global--spacer--2xl) * 2));
   --pf-c-wizard--m-full-width--lg--MaxWidth: auto;
   --pf-c-wizard--m-full-height--lg--Height: 100%;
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -8,6 +8,7 @@
   --pf-c-wizard--lg--MaxHeight: calc(100% - (var(--pf-global--spacer--2xl) * 2));
   --pf-c-wizard--m-full-width--lg--MaxWidth: auto;
   --pf-c-wizard--m-full-height--lg--Height: 100%;
+  --pf-c-wizard--m-compact-nav__nav--lg--Width: #{pf-size-prem(250px)};
```
christiemolloy: IMO the modifier `--m-compact-nav` should go on the nav instead of the wizard so you'd update the var here and on line 189 i wonder if you can move the modifier into the nav element?

mcoker: Good call, moved the variable to the nav section.

The nav width var is used elsewhere due to how the component is positioned, so it needs to exist outside of the nav selector.

Issue #1725, opened 2019-04-12
Table Drop Column


rachael-phillips:
Add option to drop column in mobile view.

mcarrano:
My understanding here is that we will provide the following:

* An option to flag any column to be dropped when the table converts to a mobile table (stacked table).
* When the width of the screen goes below a certain breakpoint and columns designated to be hidden will not be visible on the screen and other data will be moved up to take it's place.

@rachael-phillips @LHinson @mcoker can you just confirm that this is consistent with your interpretation? If so, this can be placed in the queue for CSS.

mcoker:
@mcarrano yep. The way I envision it, it will just be a simple modifier (something like `.pf-m-show-on-lg`) that only shows a particular cell at one of our global breakpoints (in that example, the cell is hidden at any point before the `lg` breakpoint), and you would apply that modifier to all cells in the column that you want to show only at/after a particular breakpoint. @tlabaj does this sound like an approach that would work with the react implementation, or would you propose something different?

LHinson:
@mcarrano agreed. Removing the "ixd" tag and adding "css"

dlabaj:
Sounds good.  We can do this in react.  @mcoker @rachael-phillips

PR #1748, opened 2019-04-17
feat(table): add support to hide/show cells at breakpoints

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1725

I had originally made this block of hidden/visible-on-* CSS a mixin that could be included in any component that takes an argument for the `display` property (for `block`, `table`, `flex`, etc). That might be a good idea in the future if we want to add this kind of functionality to other components.

### DIFF
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -1,6 +1,8 @@
 /* stylelint-disable */
 @mixin pf-mobile-layout {
-  .pf-m-grid.pf-c-table { @content; }
+  .pf-m-grid.pf-c-table {
+    @content;
+  }
 
   .pf-m-grid-md.pf-c-table {
     @media screen and (max-width: $pf-global--breakpoint--md) {
@@ -58,6 +60,9 @@
   --pf-c-table-tr--responsive--nested-table--PaddingRight: var(--pf-global--spacer--lg);
   --pf-c-table-tr--responsive--nested-table--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-table-tr--responsive--nested-table--PaddingLeft: var(--pf-global--spacer--lg);
+
+  // Cell
+  --pf-c-table-cell--m-grid--hidden-visible--Display: grid;
 
   // Td
   --pf-c-table-td--responsive--GridColumnGap: var(--pf-global--spacer--md);
@@ -194,11 +199,11 @@
   }
 
   // Standard td, in standard row (non-expandable)
-  // exclude specifically styled tds, as they are placed explicitly within grid
-  // :not(.pf-c-table__toggle):not(.pf-c-table__check):not(.pf-c-table__action):not(.pf-c-table__sort)
-  tr:not(.pf-c-table__expandable-row) > :not(.pf-c-table__toggle):not(.pf-c-table__check):not(.pf-c-table__action):not(.pf-c-table__sort) {
+  [data-label] {
+    // default pf-hidden-visible() mixin is called in table.scss. redefining variable here
+    --pf-c-table-cell--hidden-visible--Display: var(--pf-c-table-cell--m-grid--hidden-visible--Display);
+
     // responsive layout, set td to two column grid
-    display: grid;
     grid-column: 1;
     grid-column-gap: var(--pf-c-table-td--responsive--GridColumnGap);
     align-items: start;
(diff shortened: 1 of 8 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -174,10 +174,12 @@
...
     }
 
-    /* stylelint-disable-next-line */
     &.pf-m-center {
       text-align: center;
     }
+
+    // Hide/show cells at different breakpoints
+    @include pf-hidden-visible("table-cell");
   }
```
mattnolting: Love this!

src/patternfly/sass-utilities/mixins.scss
```diff
@@ -152,3 +152,70 @@
   scrollbar-width: none; // hides scrollbars in Firefox 64 and up
   -ms-overflow-style: -ms-autohiding-scrollbar; // auto hides scrollbars in Edge
 }
+
+@mixin pf-hidden-visible($val: "block") {
```
christiemolloy: Were you thinking of updating the other components that use `display: none; visibility: hidden` within this PR?

src/patternfly/sass-utilities/mixins.scss
```diff
@@ -152,3 +152,70 @@
...
+  &.pf-m-hidden {
```
christiemolloy: I feel like we should make a note that you dont want to use these modifiers on cells like inside of the expandable content. This is what happened when I added those modifiers and we run into the problem where hiding the cell doesnt hide those styles that come with the expansion.

Screen Shot 2019-04-18 at 10 08 59 AM

src/patternfly/sass-utilities/mixins.scss
mcoker: oh nah, I wasn't planning on it. this mixin was just for the hidden/visible functionality.

src/patternfly/sass-utilities/mixins.scss
mcoker: nice, I'll look at that

mcoker:
@mattnolting @christiemolloy updated.

src/patternfly/sass-utilities/mixins.scss
mcoker: I tried this out and it seems to work OK for me. It may have been an issue with how it was written previously.

priley86:
the demo [here](https://deploy-preview-1748--pf-next.netlify.com/components/Table/examples-full/?component=Table%20with%20hidden/visible%20breakpoint%20modifiers) looks good to me and I would like to prototype w/ the `pf-m-hidden-on-*` and `pf-m-visible-on-*`  classes downstream. Will also need to think about flexible row widths next (i.e. `col-xs-3`, `col-sm-3`, `col-md-3`, etc.). 

One additional thing I noted on visibility while looking at the w3 specs during the table's work so far is that we can use `aria-colcount` when columns are visible/hidden. We could probably add this during the downstream pf-react integration I'm guessing.
https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html

christiemolloy:
This is an example of hiding all of the `td's` in the first `<tr>` would you expect that bottom line to also be removed in this case?

Screen Shot 2019-04-23 at 10 30 58 AM

mcoker:
@christiemolloy this feature is for dropping columns. From https://github.com/patternfly/patternfly-next/issues/1725

> Add option to drop column in mobile view

So I would consider that a mis-use of the feature 😁

mattnolting:
> @christiemolloy this feature is for dropping columns. From #1725
> 
> > Add option to drop column in mobile view
> 
> So I would consider that a mis-use of the feature 😁

You could hide the entire row with the same method

mcoker:
@mattnolting for sure, this feature is 🔥. Let's add that if we get a request for it.

src/patternfly/components/Table/table-grid.scss
```diff
@@ -194,11 +199,11 @@
   }
 
   // Standard td, in standard row (non-expandable)
-  // exclude specifically styled tds, as they are placed explicitly within grid
-  // :not(.pf-c-table__toggle):not(.pf-c-table__check):not(.pf-c-table__action):not(.pf-c-table__sort)
-  tr:not(.pf-c-table__expandable-row) > :not(.pf-c-table__toggle):not(.pf-c-table__check):not(.pf-c-table__action):not(.pf-c-table__sort) {
+  [data-label] {
+    // default pf-hidden-visible() mixin is called in table.scss. redefining variable here
+    --pf-c-table-cell--hidden-visible--Display: var(--pf-c-table-cell--m-grid--hidden-visible--Display);
```
mattnolting: 👍

PR #1765, opened 2019-04-25
fix(vars): update padding var vals - page, nav, table, toolbar, datalist

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1757


src/patternfly/components/Page/page.scss
```diff
@@ -132,9 +132,9 @@
   --pf-c-page__main-breadcrumb--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-page__main-breadcrumb--PaddingBottom: 0;
   --pf-c-page__main-breadcrumb--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-page__main-breadcrumb--md--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-page__main-breadcrumb--md--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-page__main-breadcrumb--md--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__main-breadcrumb--md--PaddingTop: var(--pf-global--spacer--lg);
```
christiemolloy: Sidebar nav and the breadcrumb/page-nav aren't lining up now.

Screen Shot 2019-04-25 at 5 12 26 PM

src/patternfly/components/Table/table.scss
```diff
@@ -42,8 +42,8 @@
   --pf-c-table-thead-cell--PaddingBottom: var(--pf-global--spacer--md);
 
   // Td
-  --pf-c-table-tbody-cell--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-table-tbody-cell--PaddingBottom: var(--pf-global--spacer--xl);
+  --pf-c-table-tbody-cell--PaddingTop: var(--pf-global--spacer--lg);
```
christiemolloy: You'll have to update the padding when the table breaks into its grid, but before the mobile view: this is currently still 32px.
Screen Shot 2019-04-25 at 5 27 06 PM

src/patternfly/components/Page/page.scss
mcoker: Yeah that's a tricky one. The main content sections don't line up, either. Tertiary nav does though.

We could make whatever section is first (breadcrumbs, tertiary nav, main content section) have padding that aligns it with the sidebar, or we could move the nav up a little, then update the tertiary nav to match. 

What do you think?

kybaker: @mcoker We should move the sidebar nav up a little to match.

PR #1776, opened 2019-04-29
chore(links): update documentation links to support path prefixing

dgutride:


dgutride:
Helps with github.com/patternfly/patternfly-org/issues/909

srambach:
LGTM. Tested all the links and I don't see any more `<mark>` tags in the .md files.

PR #1780, opened 2019-04-30
feat(grid): add grid-gap responsive modifiers

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1739

also removed the spacing utility specific items in `$pf-global--spacer-map` and left that var as a global var that maps to our spacers so we can use it elsewhere and am using `map-merge()` to add things to that map in specific utilities/layouts.

### DIFF
--- a/src/patternfly/layouts/Grid/docs/code.md
+++ b/src/patternfly/layouts/Grid/docs/code.md
@@ -8,7 +8,7 @@ A Grid layout framework including responsive columns, row spans and gutters. Bre
 | -- | -- | -- |
 | `.pf-l-grid` | `<div>` | Initializes the grid layout. |
 | `.pf-l-grid__item` | `<div>` | Explicitly sets a child of the grid. This class isn't necessary, but it is included to keep inline with BEM convention, and to provide an entity that will later be used for applying modifiers. |
-| `.pf-m-gutter` | `.pf-l-grid` | Adds grid gutters. |
+| `.pf-m-gutter{-[size]{-on[breakpoint]}}` | `.pf-l-grid` | Adds grid gutters. |
 | `.pf-m-all-{1-12}-col{-on-[breakpoint]}` | `.pf-l-grid` | Defines grid item size on grid container. |
 | `.pf-m-{1-12}-col{-on-[breakpoint]}` | `.pf-l-grid__item` | Defines grid item size.  Although not required, they are strongly suggested. If not used, grid item will default to 12 col. |
 | `.pf-m-{2-x}-row{-on-[breakpoint]}` | `.pf-l-grid__item` | Defines grid item row span.  For row spans to function correctly, the value of of the current row plus the grid items to span must be equal to or less than 12. Example: .pf-m-8-col.pf-m-2-row + .pf-m-4-col + .pf-m-4-col. There is no limit to number of spanned rows. |
--- a/src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
+++ b/src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
@@ -1,6 +1,8 @@
+<h2 class="example-title"><code>.pf-m-gutter.pf-m-gutter-3xl-on-lg</code></h2>
+
 {{#> grid grid--modifier="pf-m-gutter pf-m-gutter-3xl-on-lg"}}
   {{#> grid-item grid-item--modifier="pf-m-6-col"}}
-      <strong>.pf-m-gutter.pf-m-gutter-3xl-on-lg</strong> - Item 1
+      Item 1
   {{/grid-item}}
   {{#> grid-item grid-item--modifier="pf-m-6-col"}}
       Item 2
@@ -18,10 +20,12 @@
       Item 6
   {{/grid-item}}
 {{/grid}}
-<br><br>
+
+<h2 class="example-title"><code>.pf-m-gutter-on-md</code></h2>
+
 {{#> grid grid--modifier="pf-m-gutter-on-md"}}
   {{#> grid-item grid-item--modifier="pf-m-6-col"}}
-      <strong>.pf-m-gutter-on-md</strong> - Item 1
+      Item 1
   {{/grid-item}}
   {{#> grid-item grid-item--modifier="pf-m-6-col"}}
       Item 2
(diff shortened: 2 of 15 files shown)


src/patternfly/layouts/Grid/docs/code.md
```diff
@@ -8,7 +8,7 @@ A Grid layout framework including responsive columns, row spans and gutters. Bre
 | -- | -- | -- |
 | `.pf-l-grid` | `<div>` | Initializes the grid layout. |
 | `.pf-l-grid__item` | `<div>` | Explicitly sets a child of the grid. This class isn't necessary, but it is included to keep inline with BEM convention, and to provide an entity that will later be used for applying modifiers. |
-| `.pf-m-gutter` | `.pf-l-grid` | Adds grid gutters. |
+| `.pf-m-gutter{-[size]{-on[breakpoint]}}` | `.pf-l-grid` | Adds grid gutters. |
```
christiemolloy: I could see a case for wanting a certain gutter size on the x-axis and a different gutter size on the y-axis, do you think this is worth exploring?

src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
```diff
@@ -0,0 +1,62 @@
...
+      Item 4
+  {{/grid-item}}
+  {{#> grid-item grid-item--modifier="pf-m-3-col"}}
+      Item 5
+  {{/grid-item}}
+  {{#> grid-item grid-item--modifier="pf-m-3-col"}}
+      Item 6
+  {{/grid-item}}
+{{/grid}}
+<br><br>
```
christiemolloy: In between these examples could you add a description about each variation, looking at this seems like they're all one example and its a little confusing

Screen Shot 2019-05-13 at 10 32 16 AM

src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
```diff
@@ -0,0 +1,62 @@
+{{#> grid grid--modifier="pf-m-gutter pf-m-gutter-3xl-on-lg"}}
```
christiemolloy: Why do you need to set `pf-m-gutter pf-m-gutter-3xl-on-lg' twice

src/patternfly/layouts/Grid/docs/code.md
mcoker: I do! I'd like to start here though. Maybe we can wait until someone requests this to begin that work. Or do you think we should go ahead and create an issue and put it on the back burner?

src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
mcoker: That will apply `.pf-m-gutter` spacing (lg) by default, then `.pf-m-gutter-3xl` spacing on the `lg` viewport. Or are you asking something else?

src/patternfly/layouts/Grid/docs/code.md
christiemolloy: Maybe we start here and then in the next breaking change period we could get it in? so creating an issue would be good!

src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
christiemolloy: I guess where I'm getting confused is that we have a default gutter size of lg. why do we need a default and a lg?

mcoker: We can't get rid of `.pf-m-gutter` since that would be a breaking change. We could leave out `.pf-m-gutter-lg`, though do you think it would be weird to offer every size except for `lg`? Also, the spacing for `$pf-global--gutter` might change at some point, so `.pf-m-gutter` would always represent the value of our global gutter value, so I can see value in leaving `.pf-m-gutter` as separate from the explicit widths. What do you think?

christiemolloy: I'm in favor of removing pf-m-gutter next breaking change period and removing it from examples now.
I don't think we need both and I can't think of anywhere else across the system where when we give options for a range of sizes, that we have had one default value, and then repeated that default value within the range of the sizes. i.e. Table has grid breakpoint modifiers but no default grid breakpoint, our utilities don't have one default.

mcoker: An argument for leaving it would be for consistency in implementation - out of the box you get our default gutter. Otherwise it would be up to apps to choose any time they want a gutter, when I presume most will just want some reasonable default like `md` or `lg`. Also what about split, stack, and gallery, which all use `$pf-global--gutter`, too?

mceledonia: I'd also be in favor of getting rid of the default variable then just assigning one of the sized variables to default, I assume it will be less confusing than having a default var which also happens to me -med for example.

mattnolting: All valid points. It makes sense to me to set `pf-m-gutter` as a default and not include breakpoints for that selector. So, non-responsive support for default `pf-m-gutter`. Then if gutters are being changed responsively, a size would be expected. Yes, `pf-m-gutter-lg` would be redundant, but only once for the non-responsive version.

**Example**
```
<div class="pf-l-grid pf-m-gutter"></div>
```
vs (what I expect to be the most common use case)
```
<div class="pf-l-grid pf-m-gutter pf-m-gutter-xl-on-xl"></div>
```
or
```
<div class="pf-l-grid pf-m-gutter-sm pf-m-gutter-xl-on-xl"></div>
```

src/patternfly/layouts/Grid/grid.scss
```diff
@@ -36,15 +36,23 @@
...
   --pf-l-grid--m-gutter--GridGap: var(--pf-global--gutter);
   --pf-l-grid__item--GridColumnStart: auto;
   --pf-l-grid__item--GridColumnEnd: span 12;
 
   display: grid;
   grid-template-columns: repeat(12, [col-start] 1fr);
+  grid-gap: var(--pf-l-grid--GridGap);
 
   > *,
   .pf-l-grid__item {
```
mattnolting: Can we drop `pf-l-grid__item`?

src/patternfly/layouts/Grid/examples/grid-responsive-gutters-example.hbs
mcoker: Personally I'm in favor of either removing the default `pf-m-gutter` or leaving it along with all of the breakpoints. @mattnolting I can see a use case of someone wanting like `.pf-m-gutter.pf-m-gutter-none-on-md.pf-m-gutter-on-lg`, and I think it might be confusing to offer specific gutter sizes with breakpoints without the default gutter supporting breakpoints. I don't see why we wouldn't want to offer that aside from it being a few more lines of code.

And I'm more in favor of leaving `pf-m-gutter` as the default gutter and including in the documentation that it is the same as `pf-m-gutter-lg`. So you can either use `gutter` (with all of the breakpoints) for a default gutter, or `gutter-[size]` for a custom gutter. But I'm also fine dropping `pf-m-gutter`, too. 

That said, I think we should add this same option to the split, stack, and gallery layouts (or anywhere we offer a generic layout with the option of a gutter - unless anyone disagrees?), and removing `.pf-m-gutter` will probably be a fairly invasive change since that would require folks define a gutter size where they're just using the default gutter currently... unless we offer this in a backwards compatible way like leaving it in the code and not documenting it.

Another argument in favor of leaving the default gutter and its breakpoints is to offer the most flexibility to the user and encourage them to use it correctly in the documentation.

This also begs the question that if we drop `pf-m-gutter`, do we need `$pf-m-global--gutter`? Was that ever needed in the first place?

mcoker:
After talking to @kybaker we're going to hold off on this. However we should update the components that use the grid-gap/gutter to use a smaller (md) gutter on mobile. https://github.com/patternfly/patternfly-next/issues/1822

PR #1790, opened 2019-05-01
fix(docs): for downstream patternfly org

redallen:
Also lowercase examples and headings.


mcoker:
@redallen can you also update the new component [blueprint](https://github.com/patternfly/patternfly-next/blob/master/build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/index.js) so it generates lower case headings?

PR #1798, opened 2019-05-03
fix(nav): fix horizontal nav spacing, background color

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1794, fixes https://github.com/patternfly/patternfly-next/issues/1797

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -34,20 +34,27 @@
 
   // Header nav
   --pf-c-page__header-nav--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-page__header-nav--md--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__header-nav--md--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-page__header-nav--lg--PaddingLeft: 0;
   --pf-c-page__header-nav--lg--MarginLeft: var(--pf-global--spacer--xl);
   --pf-c-page__header-nav--lg--MarginRight: var(--pf-global--spacer--xl);
-  --pf-c-page__header-nav--BackgroundColor: var(--pf-global--BackgroundColor--dark-transparent-100);
+  --pf-c-page__header-nav--BackgroundColor: #242424; // Custom color for this element only
   --pf-c-page__header-nav--lg--BackgroundColor: transparent;
+  --pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--Left: calc(-1 * (var(--pf-global--spacer--md) - var(--pf-global--spacer--xs))); // This sets the the button to the edge of its container with 4px clearance on the left
+  --pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--md--Left: calc(-1 * (var(--pf-global--spacer--md) - var(--pf-global--spacer--xs)));
+  --pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--lg--Left: 0;
+  --pf-c-page__header-nav--c-nav__scroll-button--lg--BackgroundColor: var(--pf-c-page__header-nav--BackgroundColor);
+  --pf-c-page__header-nav--c-nav__scroll-button--lg--Top: 0;
 
   @media screen and (min-width: $pf-global--breakpoint--md) {
     --pf-c-page__header-nav--PaddingLeft: var(--pf-c-page__header-nav--md--PaddingLeft);
+    --pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--Left: var(--pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--md--Left);
   }
 
   @media screen and (min-width: $pf-global--breakpoint--lg) {
     --pf-c-page__header-nav--PaddingLeft: var(--pf-c-page__header-nav--lg--PaddingLeft);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -34,20 +34,27 @@
...
   // Header nav
   --pf-c-page__header-nav--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-page__header-nav--md--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__header-nav--md--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-page__header-nav--lg--PaddingLeft: 0;
   --pf-c-page__header-nav--lg--MarginLeft: var(--pf-global--spacer--xl);
   --pf-c-page__header-nav--lg--MarginRight: var(--pf-global--spacer--xl);
-  --pf-c-page__header-nav--BackgroundColor: var(--pf-global--BackgroundColor--dark-transparent-100);
+  --pf-c-page__header-nav--BackgroundColor: #242424; // Custom color for this element only
   --pf-c-page__header-nav--lg--BackgroundColor: transparent;
```
christiemolloy: Do we need to set this variable, could you change the breakpoint to max-width and just set the background color there. I know its a breaking change but maybe we could create an issue to remove it?

src/patternfly/components/Page/page.scss
```diff
@@ -224,12 +232,23 @@
 .pf-c-page__header-nav {
   grid-column: 1 / -1;
   grid-row: 2 / 3;
+  min-width: 0;
```
christiemolloy: Is the min-width just for FF?

src/patternfly/components/Page/page.scss
```diff
@@ -34,20 +34,27 @@
...
-  --pf-c-page__header-nav--md--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__header-nav--md--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-page__header-nav--lg--PaddingLeft: 0;
   --pf-c-page__header-nav--lg--MarginLeft: var(--pf-global--spacer--xl);
   --pf-c-page__header-nav--lg--MarginRight: var(--pf-global--spacer--xl);
-  --pf-c-page__header-nav--BackgroundColor: var(--pf-global--BackgroundColor--dark-transparent-100);
+  --pf-c-page__header-nav--BackgroundColor: #242424; // Custom color for this element only
   --pf-c-page__header-nav--lg--BackgroundColor: transparent;
+  --pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--Left: calc(-1 * (var(--pf-global--spacer--md) - var(--pf-global--spacer--xs))); // This sets the the button to the edge of its container with 4px clearance on the left
+  --pf-c-page__header-nav--c-nav__scroll-button--nth-of-type-1--md--Left: calc(-1 * (var(--pf-global--spacer--md) - var(--pf-global--spacer--xs)));
```
christiemolloy: Should we also do this to the 2nd button to make padding balanced on both sides

Screen Shot 2019-05-13 at 9 56 26 AM

mcoker: To do that, if we move the button to the left a little, I think we'll need to add right padding to the horizontal nav so that the nav links don't show on the right side of the button (behind/underneath the button, since the nav extends to the right edge of the page). That would probably be a breaking change for folks who don't have the overflow buttons enabled, since their nav would no longer cut off/overflow to the edge of the browser and would cut off at the padding instead. But if we're OK with the way this looks for folks that don't have the overflow buttons enabled, I think that's a good change.

What do you think @kybaker, is this OK?

Screen Shot 2019-05-17 at 9 28 13 AM

FWIW, this is what it looks like currently.

Screen Shot 2019-05-17 at 9 29 42 AM

src/patternfly/components/Page/page.scss
mcoker: no, it breaks in chrome without it, too. the default is `min-width: min-content;` so by default, the header-nav element doesn't overflow and just grows to fit it's content. That causes the nav and masthead to grow too wide, resulting in this brokenness.

Screen Shot 2019-05-17 at 9 31 53 AM

src/patternfly/components/Page/page.scss
mcoker: Doesn't matter to me. It's probably the same amount of code either way - you're losing a variable but adding a media query. This way is technically more flexible, too, not that we need it to be. Feel free to create an issue if you think we should remove it.

src/patternfly/components/Page/page.scss
kybaker: @mcoker I do not think we want padding here. It looks like a bug and it is not as clear that there is overflow.

mcoker: @christiemolloy opened https://github.com/patternfly/patternfly-next/issues/1839

src/patternfly/components/Page/page.scss
christiemolloy: Thats ok then

src/patternfly/components/Page/page.scss
christiemolloy: ok great

src/patternfly/components/Page/page.scss
christiemolloy: Sounds good thanks @mcoker

PR #1801, opened 2019-05-06
feat(megamenu): adds mega menu

mcoker:
**WIP, don't merge!**

fixes https://github.com/patternfly/patternfly-next/issues/1767

cc @jgiardino


jgiardino:
This looks great! No suggestions from me on the implementation. I took a stab at the a11y documentation. Feel free to revise this.


-- | -- | --
`aria-label` | `.pf-c-mega-menu__toggle` and `.pf-c-mega-menu__menu .pf-c-nav` | Provides an accessible label for the toggle and navigation menu. The same label should be used for both. 
`aria-current="page"` | `.pf-c-nav__link.pf-m-current` | Indicates the current page link. 
`aria-label="[label of current selected item]"` | `ul.pf-c-nav__list` | The next sibling to the MegaMenu component should be the `ul.pf-c-nav__list` element in the Nav component. This attribute provides an accessible label for assistive technologies, and names the list in the global navigation after the current selected item in the mega menu.

Also, related to the `aria-current` attribute, I went back and tested this again to see if aria-current is supported in the structure we would create for the global nav that includes a mega menu, and it can exist in both menus. Could you update the a11y documentation for Nav to not say "Can only occur once on page" in the **Outcome** column?

mcoker:
@jgiardino thanks, added! re: `aria-current`, we also have that wording in the wizard docs

> `aria-current="page"` | `.pf-c-wizard__nav-link` | Indicates the current page link. Can only occur once on page. **Required for the current link** |

Do you think I should I remove it there, too?

jgiardino:
> Do you think I should I remove it there, too?

Yes, good call! Thanks for catching that!

mcoker:
Mega menu was put on hold per the section enhancements in app launcher. Closing for now.

PR #1806, opened 2019-05-08
feat(drawer): adds drawer to page component

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1786


mcoker:
FWIW, I just made up the widths, and added the small and large variations since we do that in some other places. We can change that in any way if someone has a more specific intent for the drawer width(s).

> What happens if contents of the drawer exceeds this width?

Currently it scrolls, but we can have it cut off instead. Ideally what users put in it will be fluid and not wider than the drawer width.

rachael-phillips:
@mceledonia and @kybaker would love your review here, as well!

mcoker:
As [commented in the issue](https://github.com/patternfly/patternfly-next/issues/1786#issuecomment-508156683) after talking to @mcarrano @kybaker, we're just going to add the drawer with a static width that's overridable, and the drawer should always display on top of the content, and the sidebar should always display on top of the drawer.

@jgiardino @christiemolloy ready for review when you have a chance.

src/patternfly/components/Page/page.scss
```diff
@@ -1,6 +1,20 @@
 // URL.com/guidelines#layout
 .pf-c-page {
   --pf-c-page--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);
+  --pf-c-page--GridTemplateColumns: 1fr;
```
christiemolloy: It's probably hard to tell because the content isn't inside of the drawer, but I think it might look odd that the notification drawer has a width of 450px, and then at the md breakpoint, the drawer almost doubles in width. Was this intentional? Also what happens when the sidebar is visible again, does it display on top of the drawer?

Screen Shot 2019-07-08 at 10 03 51 AM

src/patternfly/components/Page/page.scss
```diff
@@ -155,22 +171,39 @@
...
   --pf-c-page__main-section--m-dark-200--BackgroundColor: var(--pf-global--BackgroundColor--dark-transparent-200);
 
+  // Drawer
+  --pf-c-page__drawer--ZIndex: var(--pf-global--ZIndex--xs);
+  --pf-c-page__drawer--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-page__drawer--BoxShadow: var(--pf-global--BoxShadow--md-left);
+  --pf-c-page__drawer--GridArea: main;
+  --pf-c-page__drawer--md--GridArea: drawer;
+  --pf-c-page__drawer--Width: 100%;
+  --pf-c-page__drawer--md--Width: #{pf-size-prem(450px)};
```
christiemolloy: @maryshak1996 I remember us talking about the width of the drawer being the same as the width of the sidebar, how do you feel about this width?

src/patternfly/components/Page/page.scss
```diff
@@ -155,22 +171,39 @@
   --pf-c-page__main-section--m-dark-100--BackgroundColor: var(--pf-global--BackgroundColor--dark-transparent-100);
   --pf-c-page__main-section--m-dark-200--BackgroundColor: var(--pf-global--BackgroundColor--dark-transparent-200);
 
+  // Drawer
+  --pf-c-page__drawer--ZIndex: var(--pf-global--ZIndex--xs);
```
christiemolloy: Are you going to add a transition like the sidebar has?

src/patternfly/demos/Drawer/drawer.scss
```diff
@@ -0,0 +1,26 @@
+// Component variables follow this formula:
```
christiemolloy: Just a reminder to delete all of these extra files

src/patternfly/components/Page/page.scss
mcoker: Yep, I went over the design with @kybaker and @mcarrano last week, and we decided to have it full width on mobile (like the wizard) and 450px on desktop. And yep, the sidebar should always appear on top of the drawer.

src/patternfly/components/Page/page.scss
mcoker: Nope, but we can. What do you think @maryshak1996 @mceledonia?

maryshak1996: @mcoker I would think that the same transition used with the vertical nav is what would make sense here

src/patternfly/components/Page/page.scss
christiemolloy: Ok sounds good, I think it would be good to see an example of the sidebar opened on top of the drawer.

src/patternfly/components/Page/page.scss
mcoker: @christiemolloy added the transition if you want to have another look.

src/patternfly/components/Page/examples/page-component-drawer-example.hbs
```diff
@@ -0,0 +1,26 @@
...
+      header-tools
+    {{/page-header-tools}}
+  {{/page-header}}
+  {{#> page-main}}
+    {{#> page-main-section page-main-section--modifier="pf-m-light"}}
+      page-main-section
+    {{/page-main-section}}
+  {{/page-main}}
+  {{#> page-drawer}}
+    page-drawer
```
christiemolloy: Should this example should the drawer expanded? Otherwise I think its a little unclear where it is

src/patternfly/components/Page/page.scss
mcoker: I created another example so we show the collapsed and expanded states, so the react devs can see the classes needed to build for the two states. I think for how it looks with the sidebar open, we could point users to the react demo, since that would only be a visual demo and doesn't demo any code needed for a drawer state.What do you think?

We don't currently have any other page demos with the sidebar open, either. That said, I wonder if we should add 2 new demos to the page component to illustrate the expanded sidebar on mobile, and collapsed sidebar on desktop. We could always incorporate those states into the existing demos, too.

src/patternfly/components/Page/page.scss
christiemolloy: Transition looks really good!!

christiemolloy: Can you show an example of the sidebar expanded and the notification drawer expanded @mcoker

src/patternfly/components/Page/docs/code.md
```diff
@@ -42,3 +45,4 @@ This component provides the basic chrome for a page, including sidebar, header,
 | `.pf-m-no-padding-mobile` | `.pf-c-page__main-section` | Modifies a main page section to have no padding on mobile. |
 | `.pf-m-fill` | `.pf-c-page__main-section` | Modifies a main page section to grow to fill the available vertical space. |
 | `.pf-m-no-fill` | `.pf-c-page__main-section` | Modifies a main page section to not grow to fill the available vertical space. |
+| `.pf-m-expanded` | `.pf-c-page__drawer` | Modifies the drawer for the expanded state. |
```
christiemolloy: Do you think its ok that we have an expanded modifier for the drawer but not the sidebar ?

src/patternfly/components/Page/examples/page-component-drawer-example.hbs
mcoker: good call, added

src/patternfly/components/Page/page.scss
christiemolloy: I just added a comment underneath before I saw this one. But yeah I think it would be good to add new examples to the page component that show the sidebar opened and closed, I dont see an expanded state on it, so not sure what do to there.. but that sounds good then we can show what it looks like with both.

src/patternfly/components/Page/docs/code.md
mcoker: Hmm... we have `pf-m-expanded` for the sidebar, or are you referring to something else?

Screen Shot 2019-07-08 at 4 28 14 PM

christiemolloy: I missed that class, when there are two modifiers the same do we typically add them on one line in the docs like this:

`.pf-m-expanded` | `.pf-c-page__sidebar`, `.pf-c-page__drawer` | Modifies the sidebar and drawer for the expanded state.  |

christiemolloy:
When the drawer is collapsed, its box-shadow is still present on the page:

Screen Shot 2019-07-09 at 9 46 34 AM

Also were you going to remove the Drawer demo and just add it as an example in the page component?

mcoker:
closing in favor of https://github.com/patternfly/patternfly-next/pull/2069

PR #1810, opened 2019-05-10
fix(guidelines): punctuation, formatting

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1809

### DIFF
--- a/src/site/pages/guidelines.md
+++ b/src/site/pages/guidelines.md
@@ -19,7 +19,7 @@ Layouts are the containers that allow for organizing and grouping its immediate 
 
 - A layout never imposes padding or element styles on its children.
 
-The classes are prefixed with `-l` (after the patterfly prefix `pf-`), for example: `.pf-l-split` or `.pf-l-stack`.
+The classes are prefixed with `-l` (after the PatternFly prefix `pf-`), for example: `.pf-l-split` or `.pf-l-stack`.
 
 ### Components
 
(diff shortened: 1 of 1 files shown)


src/site/pages/guidelines.md
```diff
@@ -64,77 +64,76 @@ PatternFly follows a two-layer theming system where **global variables** always
 The main reason to have global variables is to maintain consistency. They adhere to the following rules:
 
 - They are prefixed with the word `global` and follow the formula `--pf-global--concept--PropertyCamelCase--modifier--state`.
```
christiemolloy: Above where it says: `The component itself never has backgrounds, widths, floats or margins.` we give components backgrounds so should we remove that

src/site/pages/guidelines.md
```diff
@@ -64,77 +64,76 @@ PatternFly follows a two-layer theming system where **global variables** always
...
 ```
 
 ### Component variables
 
 The second layer is scoped to themeable component custom properties. This setup allows for consistency across components, generates a common language between designers and developers, and gives granular control to authors. The rules are as follows:
 
-- They follow this general formula `--pf-c-block__element--modifier--state--breakpoint--pseudo-element--PropertyCamelCase`.
+- They follow this general formula `--pf-c-block[__element][--modifier][--state][--breakpoint][--pseudo-element]--PropertyCamelCase`.
   - `pf-c-block` refers to the block, usually the component or layout name (i.e., `pf-c-alert`).
   - `__element` refers to the element inside of the block (i.e., `__title`).
```
seanforyou23: `__element` is the only piece of syntax that's illustrated with the decorating dashes/underscores. To be consistent we could remove them from this one so it just reads `element` like the others, or we could add to the other examples their corresponding underscores/dashes like `--pf-c-block`, `--modifier`, etc. I think the latter is more literal and leaves little to be misunderstood, but using either approach across the board would be better regardless of which one we use.

Love the examples for each one btw, really helps.

PR #1813, opened 2019-05-10
feat(font): add option to use new red hat font

mcoker:
WIP - don't merge!!

fixes https://github.com/patternfly/patternfly-next/issues/1524

@kybaker @mceledonia

Things to note:
* I made a global change so that all `<h1-6>` elements will use the "display" font. The "normal" font weight between the 2 is different (display is bolder). So if we had an `<h3>` before with a normal font weight, it looked like normal text. Now an `<h3>` with a normal font weight is bolder than normal text.
* We had 4 font weights before (light, normal, semi-bold, and bold), and now we have 3 in Display (regular, bold, and a special, lighter, weight for rare use) and 2 in Text (regular and bold).
    * I've updated the new font system so that our existing declarations in components for "light" and "normal" font weights in Overpass use the "regular" font-weight in the new red hat font, and "semi-bold" and "bold" in Overpass use the "bold" weight in the new font.
    * I'm not using the lighter font weight with the Display font. According to Kendall, that's rarely used.

Let me know what you think.


christiemolloy:
I know this is a WIP, but noticed that the h4 here looks too bold, because it wasn't for display purposes but rather screen readers.

Screen Shot 2019-05-13 at 2 08 42 PM

mcoker:
@christiemolloy yep, all of the `<h1-6>` elements now use the "display" font, and it's "normal" font-weight is bolder than the text version of the new font. All of this may change so that we're, for example, only applying the "display" font to `.pf-c-content h1-6` and/or `.pf-c-title`, and leaving random uses of `<h1-6>` as the "text" font, but I'm not sure. Want to go through it with @kybaker and see what he expects.

christiemolloy:
@mcoker great yeah I saw your comment above, I think that option would work better.

rhamilto:
With this implementation, two sets of font files are shipping regardless of which one is being used.  This seems less than ideal.  Any reason why the SASS doesn't only include the relevant fonts based upon a variable?

cc:  @spadgett

mcoker:
Hey @rhamilto! We discussed this approach and since we have a lot of users that don't use SCSS and import CSS directly, we wanted to have a single approach for everyone. Firefox, Chrome, and Safari only download fonts that are used on the page, so those browsers shouldn't download Overpass, even if it's included in the CSS.

That said, if this approach doesn't work for you, I'm happy to offer an alternate solution for you or anyone else that might have these same concerns.

rhamilto:
The issue is not the browsers downloading the fonts, but the fonts being *distributed* with the app.  We're shipping an extra 500 kb of fonts our app doesn't use!

rhamilto:
Seems like there should be an optional SASS variable available to switch the inclusion of the correct fonts.

redallen:
Added a wiki page on how to achieve what @rhamilto wants: https://github.com/patternfly/patternfly-next/wiki/Overpass-vs-Red-Hat-Font

PR #1819, opened 2019-05-15
fix(title): add hyphen

### LINKED ISSUE #1762: How to handle word-wrapping
In cases like below where text wraps to the next line should we used the CSS hyphen option to add a hyphen between letters in words that wrap to the next line, like typical text editors do. It would be great to get some guidance from @stacymcauliffe  here.

Related to #1761 

Screen Shot 2019-04-24 at 4 54 43 PM

christiemolloy:
close #1762 

WIP
- still have to test across browsers
- test more cases
- use real words when testing

These are the components with titles that hyphen: auto would effect. I actually think that we should change the title in modal to not truncate using ellipsis for consistency. 

Screen Shot 2019-05-15 at 9 35 58 AM
Screen Shot 2019-05-15 at 9 38 03 AM
Screen Shot 2019-05-15 at 9 40 47 AM
Screen Shot 2019-05-15 at 9 41 56 AM
Screen Shot 2019-05-15 at 9 42 37 AM
Screen Shot 2019-05-15 at 9 44 05 AM

### DIFF
--- a/src/patternfly/components/Title/title.scss
+++ b/src/patternfly/components/Title/title.scss
@@ -29,6 +29,8 @@
   --pf-c-title--m-md--FontSize: var(--pf-global--FontSize--md);
   --pf-c-title--m-md--FontWeight: var(--pf-global--FontWeight--semi-bold);
 
+  hyphens: auto;
+
   &.pf-m-4xl {
     font-size: var(--pf-c-title--m-4xl--FontSize);
     font-weight: var(--pf-c-title--m-4xl--FontWeight);


mceledonia:
Big +1 from me here @christiemolloy @mcoker @kybaker

christiemolloy:
I've done a little more research into this issue.

As this issue first occurred in the Wizard I'm going to start there. At the minimum mobile screen width of 320px, using a word length of 15 characters or less, using capitalized letters, and using the default title font size, you can see in the example below that **there are no words that break to the next line ** . This is an ideal use case.

Screen Shot 2019-05-23 at 3 08 44 PM

When testing for edge cases like words 15 characters and above **hyphens should ideally be used so that the broken string of text can be identified. Look at these two examples:**

Screen Shot 2019-05-23 at 3 15 12 PM

Screen Shot 2019-05-23 at 3 15 22 PM

This is a use case where you have generic long words and edge case long words. It seems unideal that the words less than 15 characters would wrap, shouldnt the long ones? **There is currently no support for CSS Text Level 4 which would support this cause.**
Screen Shot 2019-05-23 at 3 22 32 PM

Back to the generic 15 characters words at a medium size viewport. The first is good typography, the second is poor typography. I wonder though because the web is only just catching up to these more "editorial" styling choices, if users will see the hyphens as unexpected.

Screen Shot 2019-05-23 at 3 26 23 PM

Screen Shot 2019-05-23 at 3 28 08 PM

This is a really great [article](https://medium.com/clear-left-thinking/all-you-need-to-know-about-hyphenation-in-css-2baee2d89179) that addresses hyphenation in CSS and how we may be able to set hyphenation for limiting characters so that we could only have words over 15 characters wrap, but **unfortunately this is only supported by Edge and Safari.**

Hyphenation also requires that a language be set in the HTML. It looks like Patternfly uses English US Language, which is why it hyphenates the words well in the examples above.

IMO until CSS Text Level 4 is accepted into Chrome/Firefox, we should wait to add hyphens into the Patternfly design system as we can't set the hyphen on words with a particular character count.

rachael-phillips:
@christiemolloy is it ok to close this one for now?

christiemolloy:
Closing this PR, we can discuss further in the future

PR #1829, opened 2019-05-16
fix(gutter): update mobile gutter spacing in layouts that have gutters

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1822

### DIFF
--- a/src/patternfly/_variables.scss
+++ b/src/patternfly/_variables.scss
@@ -86,6 +86,7 @@
 
   // Gutter
   --pf-global--gutter: #{$pf-global--gutter};
+  --pf-global--gutter--md: #{$pf-global--gutter--md};
 
   // Ratio
   --pf-global--golden-ratio: #{$pf-global--golden-ratio};
--- a/src/patternfly/sass-utilities/scss-variables.scss
+++ b/src/patternfly/sass-utilities/scss-variables.scss
@@ -134,6 +134,7 @@ $pf-global--spacer--form-element: pf-size-prem(6px) !default;
 
 // Gutter
 $pf-global--gutter: $pf-global--spacer--lg !default;
+$pf-global--gutter--md: $pf-global--spacer--md !default;
 
 // Ratio
 $pf-global--golden-ratio: 1.681 !default;
--- a/src/patternfly/layouts/Gallery/docs/code.md
+++ b/src/patternfly/layouts/Gallery/docs/code.md
@@ -8,4 +8,4 @@ A Gallery layout makes all of the children a uniform size, the children are disp
 | -- | -- | -- |
 | `.pf-l-gallery` |  `<div>` |  Initializes a Gallery layout |
 | `.pf-l-gallery__item` | `<div>` |  Explicitly sets the child for the gallery. This class isn't necessary, but it is included to keep inline with BEM convention, and to provide an entity that will later be used for applying modifiers. |
-| `.pf-m-gutter` | `.pf-l-gallery` | Adds space between children |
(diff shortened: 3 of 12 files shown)


src/patternfly/_variables.scss
```diff
@@ -86,6 +86,7 @@
 
   // Gutter
   --pf-global--gutter: #{$pf-global--gutter};
+  --pf-global--gutter--md: #{$pf-global--gutter--md};
```
christiemolloy: In the documentation I think it would be worth noting the spacer size that the gutter uses and then how it changes on a medium breakpoint.

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -134,6 +134,7 @@ $pf-global--spacer--form-element: pf-size-prem(6px) !default;
 
 // Gutter
 $pf-global--gutter: $pf-global--spacer--lg !default;
+$pf-global--gutter--md: $pf-global--spacer--md !default;
```
christiemolloy: Are you naming this variable by the breakpoint that it relates to? IMO when i first glance at the variable it seems like its a larger version of the default variable which could be confusing for users since it really is smaller. I wonder if it should be named with sm because its a smaller version of the gutter, in the case that the variable is used elsewhere when it doesnt relate to a breakpoint? I think most of the variables in here are named based on size over breakpoint, but let me know your reasoning :)

src/patternfly/_variables.scss
mcoker: You think this should be in the layouts' documents, or somewhere else?

src/patternfly/sass-utilities/scss-variables.scss
mcoker: Not by the breakpoint - I'm using `--md` as the modifier to represent "medium" since that's the size spacer it maps to. I wondered about this, too. The way I see it, gutters are similar to our spacing system, but they are used in a different context. We could potentially have gutters that map 1:1 with our spacers. So I think having gutter names that match the spacer names makes sense. In this case, I'm going off the assumption that `pf-global--gutter` is the same as `pf-global--gutter--lg`, so `pf-global--gutter--md` is `pf-global--spacer--md`, `pf-global--gutter--sm` would map to `pf-global--spacer--sm`, etc. If they don't match, I think that's confusing, and if we decided to have a gutter that matched `pf-global--spacer--xs`, we would need to create `pf-global--gutter--xxs`.

christiemolloy: Ok great that makes sense and I like that approach

src/patternfly/_variables.scss
christiemolloy: For example here in the Gallery Layout. "Adds space between children" is a little too broad and I think it would help if the consumer knew what size gutter was being applied.

Screen Shot 2019-05-20 at 2 19 36 PM

christiemolloy: You may need to add it to every layout that uses the gutter

mcoker: What do you have in mind for documentation that? Not to say we shouldn't, but we don't really document spacing values currently. And if we update the value of the spacing, it would be good if we don't have to update every component/layout that uses it. I wonder if we could say it uses the globally defined gutter value and generally matches the gutters used for sections in the page layout?

christiemolloy: Ok yeah I get that we don't want to explicitly state the size but I think this will help to inform the user:
"Adds space between children by using the globally defined gutter value"

mcoker: Sounds good! done.

PR #1843, opened 2019-05-20
fix(content): change margin bottom for small element

### LINKED ISSUE #1841: Small element in Content
Remove the bottom margin from the small element in the Content Component when its the last child. This goes along with our rule that components should touch all edges of their parents. It will also solve the issue in the Card Demo - https://github.com/patternfly/patternfly-next/pull/1716#issuecomment-492288569

christiemolloy:
Close #1841 

@mcoker not sure how you expected this to go, `<small>` was already added to lines 117 - 130 but the small element already has a margin bottom set, so I just added in the last-child selector there.

### DIFF
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -51,7 +51,6 @@
   --pf-c-content--h6--FontWeight: var(--pf-global--FontWeight--semi-bold);
 
   // Small text
-  --pf-c-content--small--MarginBottom: var(--pf-global--spacer--md);
   --pf-c-content--small--LineHeight: var(--pf-global--LineHeight--md);
   --pf-c-content--small--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-content--small--Color: var(--pf-global--Color--200);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Content/content.scss
```diff
@@ -208,10 +208,13 @@
 
   small {
     display: block;
-    margin-bottom: var(--pf-c-content--small--MarginBottom);
     font-size: var(--pf-c-content--small--FontSize);
     line-height: var(--pf-c-content--small--LineHeight);
     color: var(--pf-c-content--small--Color);
+
```
mcoker: Couldn't you just remove the `margin-bottom` from `small` here, and it will get a margin bottom from L122?

christiemolloy: Oh my bad didn't realize they're the same size

mcoker: Doh! I should have thought about that a little more. Technically it's a breaking change now, where the way you did it before was not, even though it's redundant. How about we do it the way you did it originally, the create a follow up issue with a breaking change label just to remove the margin from `small`?

christiemolloy: Yep that sounds good, created this issue #1844 and updated this PR

PR #1846, opened 2019-05-21
CSS Variables for Charts

### LINKED ISSUE #1723: CSS Needs for Charting
- Need a grey color pallet for bullet chart ranges. This is the only chart that uses grey.

- Need to create CSS variables for charts; margins, padding, font sizes, etc. Each chart type has its own style properties, which should all be defined by PatternFly core. An extensive list of properties can be found here https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartTheme/theme-common.js

- Need to decide what color themes we want to support for charts? There are multiple color pallets, each could be it's own theme. Currently, we have Victory based themes for both light and dark screens, but should have a discussion with design folks. For example, we have a green-dark theme, which uses the green color pallet and white text for dark backgrounds. And a green-light theme which uses black text for light backgrounds. (This also allows tooltip background colors also change for a dark screen.) We also have a multi-color theme, based on Michael's mocks.

christiemolloy:
1 /2 of #1723 

This is just the Color Variables for now.


mcoker:
From Michael's document, looks like he added a bunch of new colors. I had done some initial work here if you want to compare. 

https://github.com/mcoker/patternfly-next/blob/issue-1723/src/patternfly/sass-utilities/chart-colors.scss

I also went down the road of creating a separate variable system entirely for charts so it isn't included in the main core CSS, and the path to the CSS/SCSS would need to be manually included by someone using charts. At this point, I think it's fine just to include these vars with the main  dist package, since there aren't that many additions. But thinking forward a little, if/when we add more vars, it would cut down on our main CSS size to include them separately. Though if you go that route, I would place the file outside of `sass-utilities`, as we include all of `sass-utilities` with core's CSS and we wouldn't want to do that with the charting vars.

For now it looks like we can just focus on colors anyways, so we can look at what approach makes most sense re: incorporating the chart detail vars (borders, paddings, etc).

dlabrecq:
This PR shouldn't close #1723. I see some color variables defined, but we still need margins, padding, font sizes, height, width, etc. 

Each chart type has its own style properties, which should all be defined by PatternFly core. We also have some new components like the percentage donut and threshold donut charts, which will have their own theme properties.

An extensive list of properties can be found here https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-victory.js

christiemolloy:
@mcoker I moved just the color variables to a separate style sheet for now we want to merge those ASAP

src/patternfly/_charts.scss
```diff
@@ -0,0 +1,50 @@
...
+
+// purple
+$pf-chart-color-purple-100: #b2b0ea !default;
+$pf-chart-color-purple-200: #8481dd !default;
+$pf-chart-color-purple-300: #5752d1 !default;
+$pf-chart-color-purple-400: #3c3d99 !default;
+$pf-chart-color-purple-500: #2a265f !default;
+
+// yellow
+$pf-chart-color-yellow-100: #f9e0a2 !default;
```
mcoker: What do you think about naming these "gold" to match our core styles? I don't see why we should name them differently. Same applies for "grey" (black).

Also for these colors that match the core colors, how about we just import the `sass-utilities/colors.scss` file and reference the core variable names? That would be presuming that we want the colors to match if we update the colors in core. I think that's a safe bet. Then if we make updates in core that we don't want reflected in the charts, we can just remove the reference to the core var and put in the hex color instead.

Also can you rename this to `charts.scss`? The `_` denotes that the file is a [scss partial](https://sass-lang.com/guide#topic-4) to be included from another scss file.

christiemolloy: > What do you think about naming these "gold" to match our core styles? I don't see why we should name them differently. Same applies for "grey" (black).

what do you think about that? @gdoyle1

gdoyle1: @christiemolloy that's good with me!

PR #1857, opened 2019-05-24
fix(popover): scope title class to popover component

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1856

### DIFF
--- a/src/patternfly/components/Popover/popover.scss
+++ b/src/patternfly/components/Popover/popover.scss
@@ -78,7 +78,6 @@
 
   > .pf-c-title {
     margin-bottom: var(--pf-c-popover--c-title--MarginBottom);
-    line-height: 1;
   }
 
   // Close button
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Popover/popover.scss
```diff
@@ -76,6 +76,11 @@
   padding: var(--pf-c-popover__content--PaddingTop) var(--pf-c-popover__content--PaddingRight) var(--pf-c-popover__content--PaddingBottom) var(--pf-c-popover__content--PaddingLeft);
   background-color: var(--pf-c-popover__content--BackgroundColor);
 
+  > .pf-c-title {
+    margin-bottom: var(--pf-c-popover--c-title--MarginBottom);
+    line-height: 1;
```
christiemolloy: When I inspect this, it looks like the line-height from the title component (1.5) is overriding the line-height (1) that is set here. Do you see that too?

mcoker: Ah nice catch, I don't know why `line-height: 1` is there. If the title wraps to 2 lines, there isn't enough space/leading. I'm going to remove that :)

Screen Shot 2019-05-24 at 3 35 33 PM

PR #1859, opened 2019-05-27
fix: remove colour override from title

matthewcarleton:
A quick fix to have the title text colour align with the alert type. This was included in the original design and seems like it should be included. 
@bdellasc @srambach let me know if this was intentional and we can just close this :)

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -132,7 +132,6 @@
 
   &.pf-m-inline {
     --pf-c-alert--BoxShadow: var(--pf-c-alert--m-inline--BoxShadow);
-    --pf-c-alert__title--Color: var(--pf-c-alert--m-inline__title--Color);
     --pf-c-alert__icon--BackgroundColor: var(--pf-c-alert--BackgroundColor);
     --pf-c-alert--m-success__icon--Color: var(--pf-c-alert--m-inline--m-success__icon--Color);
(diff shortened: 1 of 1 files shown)


bdellasc:
I recall having a conversation with @kybaker that the color coordinated title text within the alerts was being changed to the standard dark gray PF text color, so I aligned my designs to that. Has that text change been reconsidered since I had that conversation with Kyle? CC: @mceledonia @maryshak1996

kybaker:
@bdellasc @matthewcarleton We do want to keep that alert color for the text. What Matt posted in the deployment site looks correct.

src/patternfly/components/Alert/alert.scss
```diff
@@ -132,7 +132,6 @@
 
   &.pf-m-inline {
     --pf-c-alert--BoxShadow: var(--pf-c-alert--m-inline--BoxShadow);
-    --pf-c-alert__title--Color: var(--pf-c-alert--m-inline__title--Color);
     --pf-c-alert__icon--BackgroundColor: var(--pf-c-alert--BackgroundColor);
```
srambach: You should probably remove the variable then also, line 63.

matthewcarleton: :)

bdellasc:
OK, thanks for the update on where that stands. I've also updated @mcarrano 's inline pattern graphics source file: "design-doc-assets.sketch" to reflect the color alert titles across all examples. It's uploaded to google drive.

PR #1862, opened 2019-05-28
feat(notificationbadge): adds notification badge

### LINKED ISSUE #1789: Icon w/ badge (for notification drawer)
Introduce a generic component for an icon with badge that will be leveraged for the notification drawer  (seen in masthead/header)

srambach:
Fixes #1789 

This adds a new component that is the blue dot on an icon showing that there are read/unread notifications.

### DIFF
--- a/src/patternfly/components/NotificationBadge/docs/code.md
+++ b/src/patternfly/components/NotificationBadge/docs/code.md
@@ -12,6 +12,6 @@ Be sure that the component associated with this indicator handles screenreader t
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-notification-badge` | `<div>` |  Initiates a notification badge. **Always use it with a modifier class.** |
-| `.pf-c-read` | `.pf-c-notification-badge` |  Applies read notification badge styling. |
+| `.pf-m-read` | `.pf-c-notification-badge` |  Applies read notification badge styling. |
 | `.pf-c-unread` | `.pf-c-notification-badge` |  Applies unread notification badge styling. |
 
--- a/src/patternfly/components/NotificationBadge/docs/code.md
+++ b/src/patternfly/components/NotificationBadge/docs/code.md
@@ -1,0 +1,17 @@
+## Overview
+
+Always add a modifier class. Never use the class `.pf-c-notification-badge` on its own.
+
+## Accessibility
+
+Be sure that the component associated with this indicator handles screenreader text indicating read or unread notifications.
+
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-notification-badge` | `<div>` |  Initiates a notification badge. **Always use it with a modifier class.** |
+| `.pf-c-read` | `.pf-c-notification-badge` |  Applies read notification badge styling. |
+| `.pf-c-unread` | `.pf-c-notification-badge` |  Applies unread notification badge styling. |
+
--- a/src/patternfly/components/NotificationBadge/docs/code.md
+++ b/src/patternfly/components/NotificationBadge/docs/code.md
@@ -13,5 +13,5 @@ Be sure that the component associated with this indicator handles screenreader t
 | -- | -- | -- |
 | `.pf-c-notification-badge` | `<div>` |  Initiates a notification badge. **Always use it with a modifier class.** |
 | `.pf-m-read` | `.pf-c-notification-badge` |  Applies read notification badge styling. |
-| `.pf-c-unread` | `.pf-c-notification-badge` |  Applies unread notification badge styling. |
+| `.pf-m-unread` | `.pf-c-notification-badge` |  Applies unread notification badge styling. |
 
--- a/src/patternfly/components/NotificationBadge/examples/index.js
+++ b/src/patternfly/components/NotificationBadge/examples/index.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import Documentation from '@siteComponents/Documentation';
+import Example from '@siteComponents/Example';
+import notificationBadgeSimpleExampleRaw from '!raw!./notification-badge-simple-example.hbs';
+import NotificationbadgeSimpleExample from './notification-badge-simple-example.hbs';
+import docs from '../docs/code.md';
+
+export const Docs = docs;
+
+export default props => {
+  const notificationBadgeSimpleExample = NotificationbadgeSimpleExample();
+  const headingText = 'Notification Badge';
+  const variablesRoot = 'pf-c-notification-badge';
+
+  return (
+    <Documentation data={props} docs={Docs} heading={headingText} variablesRoot={variablesRoot}>
+      <Example heading="Notification Badge" handlebars={notificationBadgeSimpleExampleRaw}>
+        {notificationBadgeSimpleExample}
+      </Example>
+    </Documentation>
+  );
+};
--- a/src/patternfly/components/NotificationBadge/examples/index.js
+++ b/src/patternfly/components/NotificationBadge/examples/index.js
@@ -9,12 +9,12 @@ export const Docs = docs;
 
 export default props => {
   const notificationBadgeSimpleExample = NotificationbadgeSimpleExample();
-  const headingText = 'Notification Badge';
+  const headingText = 'Notification badge';
   const variablesRoot = 'pf-c-notification-badge';
 
   return (
     <Documentation data={props} docs={Docs} heading={headingText} variablesRoot={variablesRoot}>
-      <Example heading="Notification Badge" handlebars={notificationBadgeSimpleExampleRaw}>
+      <Example heading="Notification badge" handlebars={notificationBadgeSimpleExampleRaw}>
         {notificationBadgeSimpleExample}
       </Example>
     </Documentation>
--- a/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
+++ b/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
@@ -9,27 +9,3 @@
       {{> button-icon button-icon--type="bell"}}
     {{/notification-badge}}
 {{/button}}
-
-{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
-    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
-      {{> button-icon button-icon--type="cog"}}
-    {{/notification-badge}}
-{{/button}}
-
-{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
-    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
-      {{> button-icon button-icon--type="file"}}
-    {{/notification-badge}}
-{{/button}}
-
-{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
-    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
-      {{> button-icon button-icon--type="info"}}
-    {{/notification-badge}}
-{{/button}}
-
-{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
-    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
-      {{> button-icon button-icon--type="port"}}
-    {{/notification-badge}}
-{{/button}}
--- a/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
+++ b/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
@@ -9,3 +9,27 @@
       {{> button-icon button-icon--type="bell"}}
     {{/notification-badge}}
 {{/button}}
+
+{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
+    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
+      {{> button-icon button-icon--type="cog"}}
+    {{/notification-badge}}
+{{/button}}
+
+{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
+    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
+      {{> button-icon button-icon--type="file"}}
+    {{/notification-badge}}
+{{/button}}
+
+{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
+    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
+      {{> button-icon button-icon--type="info"}}
+    {{/notification-badge}}
+{{/button}}
+
+{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
+    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
+      {{> button-icon button-icon--type="port"}}
+    {{/notification-badge}}
+{{/button}}
--- a/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
+++ b/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
@@ -1,7 +1,7 @@
 {{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Unread notifications"'}}
     {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
       {{> button-icon button-icon--type="bell"}}
-    {{/ notification-badge}}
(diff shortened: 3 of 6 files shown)


src/patternfly/components/NotificationBadge/docs/code.md
```diff
@@ -0,0 +1,17 @@
...
+
+Be sure that the component associated with this indicator handles screenreader text indicating read or unread notifications.
+
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-notification-badge` | `<div>` |  Initiates a notification badge. **Always use it with a modifier class.** |
+| `.pf-c-read` | `.pf-c-notification-badge` |  Applies read notification badge styling. |
```
mcoker:
```suggestion
| `.pf-m-read` | `.pf-c-notification-badge` |  Applies read notification badge styling. |
```

src/patternfly/components/NotificationBadge/docs/code.md
```diff
@@ -0,0 +1,17 @@
...
+| `.pf-c-unread` | `.pf-c-notification-badge` |  Applies unread notification badge styling. |
```
mcoker:
```suggestion
| `.pf-m-unread` | `.pf-c-notification-badge` |  Applies unread notification badge styling. |
```

src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
```diff
@@ -0,0 +1,11 @@
+{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Unread notifications"'}}
+    {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
+      {{> button-icon button-icon--type="bell"}}
+    {{/ notification-badge}}
```
mcoker: nit - we normally don't put a space here.

```suggestion
    {{/notification-badge}}
```

src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
```diff
@@ -0,0 +1,11 @@
...
+{{/button}}
+
+{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Notifications"'}}
+    {{#> notification-badge notification-badge--modifier="pf-m-read"}}
+      {{> button-icon button-icon--type="bell"}}
+    {{/ notification-badge}}
```
mcoker:
```suggestion
    {{/notification-badge}}
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,35 @@
...
+
+  // Modifiers
+  --pf-c-notification-badge--m-read--after--BackgroundColor: transparent;
+  --pf-c-notification-badge--m-unread--after--BackgroundColor: var(--pf-global--active-color--400);
+
+  position: relative;
+
+  &::after {
+    position: absolute;
+    top: var(--pf-c-notification-badge--pf-m-top-right--after--Top);
```
mcoker:
```suggestion
    top: var(--pf-c-notification-badge--after--Top);
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,35 @@
...
+    right: var(--pf-c-notification-badge--pf-m-top-right--after--Right);
```
mcoker:
```suggestion
    right: var(--pf-c-notification-badge--after--Right);
```

mcoker:
@mcarrano agree. Using `translate(-50%,0)` seems to align it more like I would expect.

Screen Shot 2019-05-28 at 12 54 08 PM

src/patternfly/components/NotificationBadge/docs/code.md
```diff
@@ -0,0 +1,17 @@
...
+## Accessibility
+
+Be sure that the component associated with this indicator handles screenreader text indicating read or unread notifications.
+
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-notification-badge` | `<div>` |  Initiates a notification badge. **Always use it with a modifier class.** |
```
mcoker: I wonder if we should suggest this be a `<span>` instead? Seems like it would almost always be used inline to wrap an icon.

srambach:
Visual spec I was going by:
https://redhat.invisionapp.com/share/7FQDO4RP462#/screens/345357488_Visual_Spec

src/patternfly/components/NotificationBadge/docs/code.md
srambach: changed.

maryshak1996:
@srambach @mcarrano @mcoker I think that the badge should not extend taller than the bell icon (this would just make things look misaligned when the hover affect is applied). 
It's hard to see in that spec, so here's a zoom-in. You can see that there is also a 1px wide "cut" that the blue dot makes into the bell (increases contrast better than blue on white)
Screen Shot 2019-05-28 at 2 28 33 PM

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,35 @@
+.pf-c-notification-badge {
+  --pf-c-notification-badge--after--Top: 0;
+  --pf-c-notification-badge--after--Right: 0;
+  --pf-c-notification-badge--after--Width: var(--pf-global--spacer--sm);
+  --pf-c-notification-badge--after--Height: var(--pf-global--spacer--sm);
+  --pf-c-notification-badge--after--BackgroundColor: var(--pf-global--active-color--400);
```
mcoker: should the default be `transparent` since you have to use a modifier? We also wouldn't need to update the default if we updated the "unread" color (or "read" color, if we introduce a color for that in the future) to match.

src/patternfly/components/NotificationBadge/examples/index.js
```diff
@@ -0,0 +1,22 @@
...
+export const Docs = docs;
+
+export default props => {
+  const notificationBadgeSimpleExample = NotificationbadgeSimpleExample();
+  const headingText = 'Notification Badge';
+  const variablesRoot = 'pf-c-notification-badge';
+
+  return (
+    <Documentation data={props} docs={Docs} heading={headingText} variablesRoot={variablesRoot}>
+      <Example heading="Notification Badge" handlebars={notificationBadgeSimpleExampleRaw}>
```
mcoker:
```suggestion
      <Example heading="Notification badge" handlebars={notificationBadgeSimpleExampleRaw}>
```

src/patternfly/components/NotificationBadge/examples/index.js
```diff
@@ -0,0 +1,22 @@
...
+import Example from '@siteComponents/Example';
+import notificationBadgeSimpleExampleRaw from '!raw!./notification-badge-simple-example.hbs';
+import NotificationbadgeSimpleExample from './notification-badge-simple-example.hbs';
+import docs from '../docs/code.md';
+
+export const Docs = docs;
+
+export default props => {
+  const notificationBadgeSimpleExample = NotificationbadgeSimpleExample();
+  const headingText = 'Notification Badge';
```
mcoker:
```suggestion
  const headingText = 'Notification badge';
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,35 @@
...
+  --pf-c-notification-badge--m-read--after--BackgroundColor: transparent;
+  --pf-c-notification-badge--m-unread--after--BackgroundColor: var(--pf-global--active-color--400);
+
+  position: relative;
+
+  &::after {
+    position: absolute;
+    top: var(--pf-c-notification-badge--after--Top);
+    right: var(--pf-c-notification-badge--after--Right);
+    display: inline-block;
```
mcoker: you shouldn't need this. setting `position: absolute;` will make it a block element.

srambach:
@maryshak1996  I can put the border around the badge, which gives the cut but must match the background or else you will see a phantom circle. If we make the assumption that it will always be on a BackgroundColor-100 then it will be ok on a white or black background, but not if it's placed on something in the middle. @kybaker thinks it will always be on black, but I can't guarantee it. 
image

Also, I have the circle as a small spacer size - so 8px I believe. Yours looks a little larger - can you tell me what it should be?

And third, if this badge is placed on a different icon or something else - will the positioning need to change? How should it be calculated?
image

mcoker:
@srambach something else we might want to consider re: placement of the blue circle is how react's SVGs are shaped differently than FA icons. The SVGs in react have a square aspect ratio, and the FA icons don't, so the badge may be positioned differently in react compared to core.

maryshak1996:
@srambach 8px is exactly right :) 
this is the 'algorithm' I used to align the indicator dot
Screen Shot 2019-05-29 at 10 43 38 AM

srambach:
Because icons vary in size and shape, it is impossible to guarantee that the blue circle will fall exactly where the designers would like based on any sort of algorithm if this component is provided as a notification badge to be used generically. For this reason, it seems like the best option would be to abandon the notification badge as a component and rely on creating a variant of the icon that includes the notification badge. 
This would also have the advantage of handling the cut between the icon and circle more reliably, as then the background can show through properly rather than simulating it with a border around the circle in the background color, which cannot be guaranteed to be correct.

If a SVG icon can be used, a separate layer can be created for the notification badge, which can then be turned on and off with classes.

If there are no objections, I will close this PR and update the issue to ask to create the bell icon with notification badge and others as needed.
@maryshak1996 @kybaker @mcarrano @mcoker @matthewcarleton @rachael-phillips @LHinson @gdoyle1

mcarrano:
@srambach the only potential problem I can see with this approach is that it introduces color into an icon. All of our icons are currently monochromatic. But I'd like to hear from @kybaker or @mcoker on this as I don't fully understand the constraints here. It would also be good for someone to weight in from the React side. @tlabaj @dlabaj @jschuler ?

srambach:
Proceeding with this approach. I've normalized the icon to be square in the hope that it will position correctly in React as well. The dot is currently translated to be in the correct place for the bell icon, which is the only icon to use it currently. If other icons are used in the future, there may need to be a modifier added to adjust the position.
Also note that the cut is created by a border on the circle in the background-100 color. If the icon is placed on another background color, then a modifier or some other accommodation will need to be made.
@maryshak1996 @kybaker @mcarrano @mcoker @matthewcarleton

mcarrano:
Looks good in Chrome @srambach . But I see something very different when I preview this in FireFox.

Screen Shot 2019-05-31 at 4 43 22 PM

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,52 @@
...
+  --pf-c-notification-badge--after--Width: calc(var(--pf-global--spacer--sm) + var(--pf-c-notification-badge--after--BorderWidth) + var(--pf-c-notification-badge--after--BorderWidth));
+  --pf-c-notification-badge--after--Height: calc(var(--pf-global--spacer--sm) + var(--pf-c-notification-badge--after--BorderWidth) + var(--pf-c-notification-badge--after--BorderWidth));
+  --pf-c-notification-badge--after--BackgroundColor: transparent;
+
+  // This translation works for the Bell icon. If other icons are used, a modifier may need to be created to allow for proper positioning on that icon.
+  --pf-c-notification-badge--after--TranslateX: calc(var(--pf-global--spacer--xs) + var(--pf-c-notification-badge--after--BorderWidth));
+  --pf-c-notification-badge--after--TranslateY: var(--pf-c-notification-badge--after--BorderWidth);
+
+  // The icon is normalized to be square in order to facilitate consistency in positioning
+  --pf-c-notification-badge--after__i--Width: var(--pf-global--spacer--md);
```
mcoker:
```suggestion
  --pf-c-notification-badge__i--Width: var(--pf-global--spacer--md);
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,52 @@
...
+  --pf-c-notification-badge--after__i--Height: var(--pf-global--spacer--md);
```
mcoker:
```suggestion
  --pf-c-notification-badge__i--Height: var(--pf-global--spacer--md);
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,52 @@
...
+  --pf-c-notification-badge--after--BorderRadius: var(--pf-global--BorderRadius--lg);
+  --pf-c-notification-badge--after--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-notification-badge--after--Top: 0;
+  --pf-c-notification-badge--after--Right: 0;
+  --pf-c-notification-badge--after--Width: calc(var(--pf-global--spacer--sm) + var(--pf-c-notification-badge--after--BorderWidth) + var(--pf-c-notification-badge--after--BorderWidth));
+  --pf-c-notification-badge--after--Height: calc(var(--pf-global--spacer--sm) + var(--pf-c-notification-badge--after--BorderWidth) + var(--pf-c-notification-badge--after--BorderWidth));
+  --pf-c-notification-badge--after--BackgroundColor: transparent;
+
+  // This translation works for the Bell icon. If other icons are used, a modifier may need to be created to allow for proper positioning on that icon.
+  --pf-c-notification-badge--after--TranslateX: calc(var(--pf-global--spacer--xs) + var(--pf-c-notification-badge--after--BorderWidth));
```
mcoker: If the goal is to offset the badge by half of it's width, you could use `50%` instead of `var(--pf-global--spacer--xs)`. That's how it looked to me from the designs.

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,52 @@
...
+    height: var(--pf-c-notification-badge--after--Height);
+    content: "";
+    background-color: var(--pf-c-notification-badge--after--BackgroundColor);
+    border: var(--pf-c-notification-badge--after--BorderWidth) solid var(--pf-c-notification-badge--after--BorderColor);
+    border-radius: var(--pf-c-notification-badge--after--BorderRadius);
+    transform: translate(var(--pf-c-notification-badge--after--TranslateX), var(--pf-c-notification-badge--after--TranslateY));
+  }
+
+  & > i {
+    width: var(--pf-c-notification-badge--after__i--Width);
```
mcoker:
```suggestion
    width: var(--pf-c-notification-badge__i--Width);
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,52 @@
...
+    height: var(--pf-c-notification-badge--after__i--Height);
```
mcoker:
```suggestion
    height: var(--pf-c-notification-badge__i--Height);
```

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -0,0 +1,52 @@
...
+  // The icon is normalized to be square in order to facilitate consistency in positioning
+  --pf-c-notification-badge--after__i--Width: var(--pf-global--spacer--md);
+  --pf-c-notification-badge--after__i--Height: var(--pf-global--spacer--md);
+
+  // Modifiers
+  --pf-c-notification-badge--m-read--after--BackgroundColor: transparent;
+  --pf-c-notification-badge--m-read--after--BorderColor: transparent;
+  --pf-c-notification-badge--m-unread--after--BackgroundColor: var(--pf-global--active-color--400);
+
+  position: relative;
```
mcoker: looks like we should also add `display: inline-block;` so it wraps the icon properly in FF and Edge.

srambach: This fixes it for FF but causes Chrome to position the icon slightly differently, so the translation for the badge is off by 2 px. I adjusted it by 1px to create a happy medium between FF and Chrome. If you know some other magic to position it consistently between the browsers, let me know.

src/patternfly/components/NotificationBadge/notification-badge.scss
srambach: I used the spacer plus the border width because @maryshak1996 said above that it was pushed out by 4px. 50% works in this case so I went ahead and changed it in the hope that it might work for other icons without adjustment.

src/patternfly/components/NotificationBadge/notification-badge.scss
mcoker: Does using that in combination with `line-height: 1` improve/normalize the positioning? That should cause the element to wrap the icon better.

PR #1863, opened 2019-05-28
feat(charts): add remaining variables

### LINKED ISSUE #1723: CSS Needs for Charting

christiemolloy:
closes #1723

This Pull Request is for the remaining variables for Charts. I spoke to @dlabrecq and these vars should be specific to the chart variations which I followed [here](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-victory.js#L42) . @gdoyle1 could you review to check that I properly covered the variables [here](https://docs.google.com/document/d/1cw10pJFXWruB1SA8TQwituxn5Ss6KpxYPCOYGrH8qAY/edit) ?


gdoyle1:
Looks good! @christiemolloy

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
 $pf-chart-color-black-300: $pf-color-black-500;
 $pf-chart-color-black-400: $pf-color-black-600;
 $pf-chart-color-black-500: $pf-color-black-700;
+
+
+// typography
+$pf-chart-global--FontSize: $pf-global--FontSize--sm;
```
christiemolloy: @mcoker do I need to pull in the Red Hat Font to these style sheet?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Layout Props
+$pf-chart-global-layout--Padding: 50px;
+$pf-chart-global-layout--Height: 300px;
+$pf-chart-global-layout--Width: 450px;
+
+// Individual Charts
+
+// Area Chart
+$pf-chart-area--Opacity: .4;
+$pf-chart-area-stroke--Width: 2px;
```
christiemolloy: @mcoker  im not sure whether to break stroke width into two words or not

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-bullet--zero-line--BorderColor: $pf-color-black-700;
+$pf-chart-bullet--zero-line--BorderWidth: 1px;
+$pf-chart-bullet--negative-primary-measure--BorderColor: $pf-color-red-100;
+$pf-chart-bullet--negative-primary-measure--BorderWidth: 2px;
+
+// Candlestick
+$pf-chart-candelstick-data-stroke--Width: 1px;
+
+// Chart
+$pf-chart-chart--BorderWidth: 1px;
```
christiemolloy: Should the "chart" variation of charts be defined globally? or is it actually a variation @dlabrecq

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Donut Chart
+$pf-chart-donut-color-threshold-large: $pf-color-black-200;
+$pf-chart-donut-color-threshold-medium: $pf-color-black-300;
```
christiemolloy: @gdoyle1 does small/med/large work here for naming?

gdoyle1: first/second/third thresholds

src/patternfly/patternfly-charts.scss
gdoyle1: font should be 12px

gdoyle1: for donut charts with the a single label in the middle representing data value, should be 24px

gdoyle1: legend should be using 14px text

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-area--Opacity: .4;
+$pf-chart-area-stroke--Width: 2px;
+
+// Axis Chart
+$pf-chart-axis-axis-stroke--Width: 1px;
+$pf-chart-axis-axis-label--Padding: 100px;
+$pf-chart-axis--ticks--Width: 1px;  // this should be a var for size so not sure if width is appropriate
+
+// Bar Chart
+$pf-chart-bar--Width: 10px;
```
dlabrecq: Do we want a var for stroke? I have this set to none, but we could easily change that via the theme. 

This is the color around the bar. Below I set this to yellow.

Screen Shot 2019-05-29 at 11 03 57 AM

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
 $pf-chart-color-black-300: $pf-color-black-500;
 $pf-chart-color-black-400: $pf-color-black-600;
 $pf-chart-color-black-500: $pf-color-black-700;
+
+
+// typography
+$pf-chart-global--FontSize--xs: $pf-global--FontSize--xs; // Axis Labels
+$pf-chart-global--FontSize--sm: $pf-global--FontSize--sm; // For Legend
+$pf-chart-global--FontSize--2xl: $pf-global--FontSize--2xl; // For Large Numbers
```
dlabrecq: Do we want to have a var for the chart font family? This is currently hard coded as global_FontFamily_sans_serif.value

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-boxplot-median--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-median-stroke--Width: 1px;
+$pf-chart-boxplot-min--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-min-stroke--Width: 1px;
+$pf-chart-boxplot-q1--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-q3--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-box--Width: 20px;
+
+// Bullet Chart
+$pf-chart-bullet--comparative-measure-line-warning--BorderColor: $pf-color-orange-300;
```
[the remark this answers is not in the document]
dlabrecq: Yes, please. 

We will have a separate theme files for bullet chart, donut, and the new donut utilization charts below. However, we still need the vars to use with the components.

The props for the donut chart can be found here:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-donut.js

The props for the donut utilization chart are here:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-donut-utilization.js

The props for the new donut utilization chart with static thresholds are here:
https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-charts/src/components/ChartTheme/themes/theme-donut-threshold.js

src/patternfly/patternfly-charts.scss
dlabrecq: Chart is a separate component. Basically it's a container which can be used with any chart to provide default functionality like a border, x and y axes, etc.

dlabrecq: FYI, the border is currently border: `1px solid`. Do we have `solid` defined anywhere?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Chart
+$pf-chart-chart--BorderWidth: 1px;
+
+// Donut Chart
+$pf-chart-donut-color-threshold-first: $pf-color-black-200;
+$pf-chart-donut-color-threshold-second: $pf-color-black-300;
+$pf-chart-donut-color-threshold-third: $pf-color-black-400;
+$pf-chart-donut-color-threshold-success: $pf-chart-color-blue-300;
+$pf-chart-donut-color-threshold-warning: $pf-chart-color-gold-400;
+$pf-chart-donut-color-threshold-danger: $pf-color-red-100;
```
dlabrecq: Will need other vars here as described above.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Error Bar
+$pf-chart-errorbar--BorderWidth: $pf-global--spacer--sm;
+$pf-chart-errorbar-data-stroke--Width: 2px;
```
dlabrecq: Should we have a var for the fill? Currently it's hard coded as 'transparent', but could be changed via the theme

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Legend
+$pf-chart-legend-gutter--Width: 20px;
+$pf-chart-legend-title--FontSize: $pf-chart-global--FontSize;
+$pf-chart-legend-title--Padding: 2px;
+
```
dlabrecq: In general, each component has their own layout and label props. For example, legend looks like this:

Screen Shot 2019-05-29 at 11 21 16 AM

The layout and label props that are typically applied are:

Screen Shot 2019-05-29 at 11 22 09 AM

I'm wondering if these props should be applied globally? Or, should var be defined for each chart component? Thus, allowing a single component to be overriding individually, without effecting other charts.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Line Chart
+$pf-chart-line-stroke--Width: 2px;
```
dlabrecq: Should we have a var for the fill? Currently it's hard coded as 'transparent', but could be changed via the theme.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
```
dlabrecq: I noticed that there are some properties that are not defined in victory-theme.js. Apparently, we're getting some defaults by relying on the Victory grayscale theme. However, we probably want to add our own vars. For example, line chart can have these props:

Screen Shot 2019-05-29 at 11 32 42 AM

Victory has a grayscale.js theme, which shows the other available props. For example, line can have fill, stroke, strokeWidth, etc.
https://github.com/FormidableLabs/victory/blob/master/packages/victory-core/src/victory-theme/grayscale.js

Victory also has a material.js theme which uses some props that are not listed in grayscale.js. For example, this theme defines an opacity for line chart.
https://github.com/FormidableLabs/victory/blob/master/packages/victory-core/src/victory-theme/material.js
https://github.com/FormidableLabs/victory/blob/master/packages/victory-core/src/victory-theme/material.js

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Pie Chart
+$pf-chart-pie--Padding: $pf-global--spacer--sm;
+$pf-chart-pie-data-Padding: $pf-global--spacer--sm;
```
dlabrecq: I don't believe we can use `rem` values here. This must be pixels. For example, the donut relies on this being 8px.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-pie-data-stroke--Width: 1px;
+$pf-chart-pie-labels--Padding: $pf-global--spacer--sm;
+$pf-chart-pie--Height: 230px;
+$pf-chart-pie--Width: 230px;
+
+// Scatter Chart
+$pf-chart-stack-data-stroke--Width: 1px;
```
dlabrecq: The scatter chart currently has strokeWidth set to `0` and stroke as `transparent`

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Tooltip
+$pf-chart-tooltip-pointer--Height: 10px; // this is equivalent to length
+$pf-chart-tooltip-pointer--Width: 20px;
+$pf-chart-tooltip--Padding: $pf-global--spacer--md;
```
dlabrecq: The tooltip has a bunch of important props that effect the styling. For example, cornerRadius, strokeWidth, padding, etc.

Screen Shot 2019-05-29 at 11 47 37 AM

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Voronoi Chart
+$pf-chart-voronoi-label--Padding: $pf-global--spacer--sm;
+$pf-chart-voronoi-flyout-stroke--Width: 1px;
```
dlabrecq: Missing a few important props here. For example, padding, strokeWdith, etc., which are different than the flyout tooltip styles

Screen Shot 2019-05-29 at 11 50 49 AM

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// typography
+$pf-chart-global--FontSize: $pf-global--FontSize--sm;
+
+// label
+$pf-chart-global-label--FontSize: $pf-chart-global--FontSize;
+$pf-chart-global-label--Padding: 10px;
+
+// Layout Props
+$pf-chart-global-layout--Padding: 50px;
```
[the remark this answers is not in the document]
mcoker: If victory doesn't work with rems, then I think we're fine using pixels.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Individual Charts
+
+// Area Chart
+$pf-chart-area--Opacity: .4;
+$pf-chart-area-stroke--Width: 2px;
+
+// Axis Chart
+$pf-chart-axis-axis-stroke--Width: 1px;
+$pf-chart-axis-axis-label--Padding: 100px;
+$pf-chart-axis--ticks--Width: 1px;  // this should be a var for size so not sure if width is appropriate
```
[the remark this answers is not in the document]
mcoker: I think width works since the goal is for this stylesheet to be used outside of the context of victory as well.

src/patternfly/patternfly-charts.scss
mcoker: I think this is fine, though I think we should follow the same general formula we do for other [global vars](https://pf4.patternfly.org/guidelines) and use `$pf-chart-area--stroke--Width`. Though what is the "stroke" width? Is that the same as a border?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-global-layout--Width: 450px;
+
+// Individual Charts
+
+// Area Chart
+$pf-chart-area--Opacity: .4;
+$pf-chart-area-stroke--Width: 2px;
+
+// Axis Chart
+$pf-chart-axis-axis-stroke--Width: 1px;
```
mcoker: why is `axis` in the var name twice?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-axis-axis-label--Padding: 100px;
+$pf-chart-axis--ticks--Width: 1px;  // this should be a var for size so not sure if width is appropriate
+
+// Bar Chart
+$pf-chart-bar--Width: 10px;
+
+// Box Plot Chart
+$pf-chart-boxplot-max--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-max-stroke--Width: 1px;
```
mcoker: Similar to my other comment https://github.com/patternfly/patternfly-next/pull/1863/files#r288707836

I think the prefix should be `$pf-chart-global` or `$pf-chart-boxplot` followed by `--concept` (or thing you're styling) followed by `--modifier` (if applicable), followed by `--PropertyCamelCase`

So this would become

`$pf-chart-boxplot--max--stroke--Width`
`$pf-chart-boxplot--median--stroke--Width`
`$pf-chart-boxplot--min--stroke--Width`
etc...

src/patternfly/patternfly-charts.scss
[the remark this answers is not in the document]
christiemolloy: Updated based on these files. Some questions. 
- In pie chart, what is padAngle?
- Those font sizes should be 14 and 24 instead of 16 and 26, so they relate to the global font sizes 
- Do you need a variable for: orientation: vertical?

- inside of the donut key value pairs, is "pie". Do those variables relate to pie or donut? I've seen the same height/width of 230 be used under pie but in the pie variation and in the pie inside of donut.

src/patternfly/patternfly-charts.scss
christiemolloy: @mcoker not sure if you missed this comment :)

src/patternfly/patternfly-charts.scss
christiemolloy: Thats the thing I was debating. Stroke width is different from border width, so I thought it was more of a concept rather than a property. What do you think @mcoker

src/patternfly/patternfly-charts.scss
christiemolloy: @dlabrecq when we usually define vars we wouldn't do one for solid, but this makes sense since its the only css defined for use in the victory charts

src/patternfly/patternfly-charts.scss
christiemolloy: i.e. here there is a stroke width and a border width?

Screen Shot 2019-05-30 at 11 27 41 AM

src/patternfly/patternfly-charts.scss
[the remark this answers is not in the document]
dlabrecq: padAngle is the space between the pie (i.e., dount) chart segments. This is an example with padAngle=1, which looks like 3 pixels in the design.

Screen Shot 2019-05-30 at 11 48 36 AM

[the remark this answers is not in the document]
dlabrecq: pie and donut are separate components, but Victory only supports pie charts. When creating a donut theme, we merge the donut properties on top of the pie properties. That gives us the look of the donut, but it's actually a pie chart.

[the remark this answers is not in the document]
dlabrecq: The legend component has a default orientation, so probably not necessary?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,97 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Axis Chart
+$pf-chart-axis-axis-stroke--Width: 1px;
+$pf-chart-axis-axis-label--Padding: 100px;
+$pf-chart-axis--ticks--Width: 1px;  // this should be a var for size so not sure if width is appropriate
+
+// Bar Chart
+$pf-chart-bar--Width: 10px;
+
+// Box Plot Chart
+$pf-chart-boxplot-max--Padding: $pf-global--spacer--sm;
```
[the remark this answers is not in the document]
dlabrecq: I'll defer to @mcoker, but if we have separate vars for each component, then devs can override properties for just that one chart component. That is, without changing properties for all charts.

src/patternfly/patternfly-charts.scss
dlabrecq: Yes, they are different.

src/patternfly/patternfly-charts.scss
dlabrecq: They could also be separate vars and we concatinate them in the Victory theme?

src/patternfly/patternfly-charts.scss
christiemolloy: Added a var

src/patternfly/patternfly-charts.scss
christiemolloy: I tagged @mcoker about this as I'm not sure the best way to import the new font family

src/patternfly/patternfly-charts.scss
christiemolloy: I spoke with @gdoyle1 about this and she said it would be best to keep those global properties global, i.e. font size / font family to ensure consistency across charts.

src/patternfly/patternfly-charts.scss
christiemolloy: Added in the remaining properties from both of those files

src/patternfly/patternfly-charts.scss
christiemolloy: It refers to the axis within the axis variation of the chart, it was for consistency:
Screen Shot 2019-05-30 at 3 17 40 PM

src/patternfly/patternfly-charts.scss
christiemolloy: Tried to follow this throughout the doc, but it would be great if you could double check the changes

src/patternfly/patternfly-charts.scss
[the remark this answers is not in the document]
christiemolloy: That makes sense about pie and donut, and since only colors are defined for donut it keeps the properties of the pie separate.

Also I see in the material.js theme that the legend has an orientation of 'vertical' which is different to the victory file where the legend has an orientation of 'horizontal', so i think ill keep it in there for now?

src/patternfly/patternfly-charts.scss
christiemolloy: @mcoker what do you think? borders by default are solid right. Across the system I see alot of border widths without the 'solid'

src/patternfly/patternfly-charts.scss
mcoker: I think so. Seems like we would want the chart theme to only use `$pf-chart` vars? Then it allows us the flexibility in the future to deviate from core by simply updating the variable value, instead of also needing to update the react chart theme code.

src/patternfly/patternfly-charts.scss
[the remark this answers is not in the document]
mcoker: @dlabrecq agreed.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-boxplot-max-stroke--Width: 1px;
+$pf-chart-boxplot-median--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-median-stroke--Width: 1px;
+$pf-chart-boxplot-min--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-min-stroke--Width: 1px;
+$pf-chart-boxplot-q1--Padding: $pf-global--spacer--sm;
+$pf-chart-boxplot-q3--Padding: $pf-global--spacer--sm;
```
mcoker: what do q1 and q3 represent?

christiemolloy: @gdoyle1 ?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-boxplot-box--Width: 20px;
+
+// Bullet Chart
+$pf-chart-bullet--comparative-measure-line-warning--BorderColor: $pf-color-orange-300;
+$pf-chart-bullet--comparative-measure-line-warning--BorderWidth: 1px;
```
mcoker: in this line and the ones below "warning" and "error" are states, so there should be 2 hyphens before that text.

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-bullet--comparative-measure-line-error--BorderColor: $pf-color-red-100;
+$pf-chart-bullet--comparative-measure-line-error--BorderWidth: 1px;
+$pf-chart-bullet--zero-line--BorderColor: $pf-color-black-700;
+$pf-chart-bullet--zero-line--BorderWidth: 1px;
+$pf-chart-bullet--negative-primary-measure--BorderColor: $pf-color-red-100;
+$pf-chart-bullet--negative-primary-measure--BorderWidth: 2px;
+
+// Candlestick
+$pf-chart-candelstick-data-stroke--Width: 1px;
```
mcoker: should this be `$pf-chart-candelstick--data--stroke--Width`?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,99 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Chart
+$pf-chart-chart--BorderWidth: 1px;
+
+// Donut Chart
+$pf-chart-donut-color-threshold-first: $pf-color-black-200;
```
mcoker: Should this be `$pf-chart-donut--threshold--first--Color`?

src/patternfly/patternfly-charts.scss
dlabrecq: The only border I see in the Victory themes is for the Chart component -- not including borderWidth props. Keep in mind that these are SVGs, not CSS props.

src/patternfly/patternfly-charts.scss
dlabrecq: These are Victory BoxPlot component properties. You can read about them here: https://formidable.com/open-source/victory/docs/victory-box-plot#q1

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Scatter Chart
+$pf-chart-stack--data--stroke--Width: 1;
+$pf-chart-stack--colorScale: "colors";
+
+// Tooltip
+$pf-chart-tooltip--cornerRadius: 0; // this is 5 in grayscale theme
+$pf-chart-tooltip--pointerLength: 10; // grayscale theme
+$pf-chart-tooltip--flyoutStyle--cornerRadius: 0;
+$pf-chart-tooltip--flyoutStyle--stroke--Width: 0; // this is 1 in grayscale theme
+$pf-chart-tooltip--flyoutStyle--pointerEvents: "none";
```
mcoker: `pointer-events` is a property name, so you could call this `PointerEvents`

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-scatter--data--BackgroundColor: "charcoal"; // grayscale
+
+// Scatter Chart
+$pf-chart-stack--data--stroke--Width: 1;
+$pf-chart-stack--colorScale: "colors";
+
+// Tooltip
+$pf-chart-tooltip--cornerRadius: 0; // this is 5 in grayscale theme
+$pf-chart-tooltip--pointerLength: 10; // grayscale theme
+$pf-chart-tooltip--flyoutStyle--cornerRadius: 0;
```
mcoker: can we just hyphenate names that aren't CSS properties? so this would be `corner-radius`. Also could we use `BorderRadius` here?

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-tooltip--flyoutStyle--stroke--Width: 0; // this is 1 in grayscale theme
+$pf-chart-tooltip--flyoutStyle--pointerEvents: "none";
+$pf-chart-tooltip--flyoutStyle--stroke--Color: "charcoal"; // grayscale theme
+$pf-chart-tooltip--flyoutStyle--BackgroundColor: #f0f0f0; // grayscale theme
+$pf-chart-tooltip--pointer--Height: 10; // this is equivalent to length
+$pf-chart-tooltip--pointer--Width: 20;
+$pf-chart-tooltip--Padding: 16;
+$pf-chart-tooltip--pointerEvents: "none";
```
mcoker:
```suggestion
$pf-chart-tooltip--PointerEvents: "none";
```

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-tooltip--labels--Padding: 5; // grayscale theme
+$pf-chart-tooltip--labels--pointerEvents: "none"; // grayscale theme
```
mcoker:
```suggestion
$pf-chart-tooltip--labels--PointerEvents: "none"; // grayscale theme
```

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Voronoi Chart
+$pf-chart-voronoi--data--BackgroundColor: transparent;
+$pf-chart-voronoi--data--stroke--Color: transparent;
+$pf-chart-voronoi--data--stroke--Width: 0;
+$pf-chart-voronoi--labels--Padding: 8; // this is 5 in grayscale theme
+$pf-chart-voronoi--labels--pointerEvents: "none";
```
mcoker:
```suggestion
$pf-chart-voronoi--labels--PointerEvents: "none";
```

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-voronoi--flyout--stroke--Width: 1;
+$pf-chart-voronoi--flyout--pointerEvents: "none";
```
mcoker:
```suggestion
$pf-chart-voronoi--flyout--PointerEvents: "none";
```

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-voronoi--flyout--stroke--Color: "charcoal";
+$pf-chart-voronoi--flyout--stroke--BackgroundColor: #f0f0f0;
+$pf-chart-voronoi--flyout--pointerEvents: "none";
```
mcoker:
```suggestion
$pf-chart-voronoi--flyout--PointerEvents: "none";
```

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Axis Chart
+$pf-chart-axis--axis--stroke--Width: 1;
+$pf-chart-axis--axis--stroke--Color: "charcoal"; // Grayscale theme
+$pf-chart-axis--axis--BackgroundColor: transparent; // Grayscale theme
+$pf-chart-axis--axis-label--Padding: 100; // I see 25px padding for this in the grayscale theme
+$pf-chart-axis--axis-label--stroke--Color: "transparent"; // material theme
+$pf-chart-axis--grid--BackgroundColor: transparent; // is transparent ok for none?
+$pf-chart-axis--grid--stroke--Color: "blueGrey50"; // material theme
+$pf-chart-axis--grid--BorderColor: transparent;
+$pf-chart-axis--grid--pointerEvents: "painted";
```
mcoker:
```suggestion
$pf-chart-axis--grid--PointerEvents: "painted";
```

christiemolloy:
@gdoyle1 there are two themes , one called grayscale and one called material. Right now we are relying on them so we want to use the same props that they include. Could you take a look at: https://github.com/FormidableLabs/victory/blob/master/packages/victory-core/src/victory-theme/grayscale.js and when you see the assignments at the top of the file could you take a look at these colors and let me know what they "should" represent in our color system.
- there are 7 colors in the 'color' array --> what should these represent, the grey chart family?
- what should 'charcoal' represent?
- what should 'grey' represent?

I believe we will also need a conversion for the Material.js theme :) https://github.com/FormidableLabs/victory/blob/master/packages/victory-core/src/victory-theme/material.js

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,172 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-boxplot--max--Padding: 8;
+$pf-chart-boxplot--max--stroke--Color: "charcoal"; // grayscale theme
+$pf-chart-boxplot--max--stroke--Width: 1;
+$pf-chart-boxplot--median--Padding: 8;
+$pf-chart-boxplot--median--stroke--Color: "charcoal"; // grayscale theme
+$pf-chart-boxplot--median--stroke--Width: 1;
+$pf-chart-boxplot--min--Padding: 8;
+$pf-chart-boxplot--min--stroke--Width: 1;
+$pf-chart-boxplot--min--stroke--Color: "charcoal"; // grayscale theme
+$pf-chart-boxplot--q1--Padding: 8;
```
mcoker: Lost the previous comment. @dlabrecq you pointed me [here](https://formidable.com/open-source/victory/docs/victory-box-plot/#q1) but I'm still not sure what q1 is :) I'm just curious if we can give a more accessible name/description for that var instead of something that's victory specific.

gdoyle1:
@christiemolloy Regarding the material color replacements, see here! 
colors

src/patternfly/patternfly-charts.scss
christiemolloy: q1 means Lower Quartile and q3 means Upper Quartile, basically the beginning and end of the median.

src/patternfly/patternfly-charts.scss
christiemolloy: Updated all names that aren't CSS properties to have hyphens.

@dlabrecq do you know if corner radius and border radius are the same things?

christiemolloy:
@gdoyle1 from the colors you posted there, color #F5C12E isnt a variable in our global colors or our chart colors, was this intended?

christiemolloy:
@dlabrecq I chatted with @mcoker briefly last week about the grayscale/material theme variables. When I was adding these variables I noticed that alot of the props cross over with the props in the main victory style sheet, except material/grayscale have different values. I'm curious as to how these stylesheets will be used? Will they be used separately from the main victory or will they be used on top of the main victory, and when they're used on top of, will they replace just the value from the main victory?  
I'm wondering if we should:
1. split the variables for material/grayscale into their own stylesheets
2. add "material" or "grayscale" into the variable name and keep them all in the one stylesheet.

dlabrecq:
The material theme isn't used at all. However, the default is currently grayscale. If we don't provide a property in our theme, the grayscale property will be used by default.

Basically, our theme properties are merged on top of the grayscale theme.

src/patternfly/patternfly-charts.scss
dlabrecq: corner radius is different than border. For example, we want the cornerRadius=0 so our  tooltip boxes look be square.

gdoyle1:
> @gdoyle1 from the colors you posted there, color #F5C12E isnt a variable in our global colors or our chart colors, was this intended?

I meant #F0AB00! Sorry

src/patternfly/patternfly-charts.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-line--data--stroke--Width: 2;
+$pf-chart-line--data--stroke--Color: $pf-color-black-900; // grayscale theme
+
+// Pie Chart
+$pf-chart-pie--Padding: 8; // 50 in grayscale theme
+$pf-chart-pie--data--Padding: 8; // this is 10px in grayscale theme
+$pf-chart-pie--data--stroke--Width: 1;
+$pf-chart-pie--data--stroke--Color: transparent; // grayscale theme
+$pf-chart-pie--labels--Padding: 8; // this is 20 in grayscale theme
+$pf-chart-pie--static--Height: 230; // 400 in grayscale theme
```
christiemolloy: @dlabrecq do you know what the difference is between static height, dynamic height, and just height, do you want vars for all three?

dlabrecq: Yes, we need all 3.

The height represents the overall SVG height. That can be used to change the scale of the SVG.

The static height is for the static portion of the chart. That is, the outer circle that represents the threshold -- that visualization doesn't change. 

The dynamic height is for the dynamic part of the chart. That is, the inner circle that changes dynamically based on the current value.

They are actually two separate pie charts that are used for the "donut utilization chart with thresholds".

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Layout Props
+$pf-chart-global--layout--Padding: 50;
+$pf-chart-global--layout--Height: 300;
+$pf-chart-global--layout--Width: 450;
+
+// Stroke
+$pf-chart-global--stroke-line-cap: "round";
+$pf-chart-global--stroke-line-join: "round";
+$pf-chart-global--stroke-dash-array: "10, 5";
```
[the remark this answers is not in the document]
christiemolloy: Added, and yes I assumed the label variables are global for all charts

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Individual Charts
+
+// Area Chart
+$pf-chart-area--Opacity: .4;
+$pf-chart-area--stroke--Width: 2;
+$pf-chart-area--data--BackgroundColor: $pf-color-black-900;
```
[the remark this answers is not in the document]
christiemolloy: Ok updated all the background colors to fills, that makes more sense

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Bullet Chart
+$pf-chart-bullet--comparative-measure-line--warning--BorderColor: $pf-color-orange-300;
+$pf-chart-bullet--comparative-measure-line--warning--BorderWidth: 1;
+$pf-chart-bullet--comparative-measure-line--error--BorderColor: $pf-color-red-100;
+$pf-chart-bullet--comparative-measure-line--error--BorderWidth: 1;
+$pf-chart-bullet--zero-line--BorderColor: $pf-color-black-700;
+$pf-chart-bullet--zero-line--BorderWidth: 1;
+$pf-chart-bullet--negative-primary-measure--BorderColor: $pf-color-red-100;
+$pf-chart-bullet--negative-primary-measure--BorderWidth: 2;
```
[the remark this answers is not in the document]
christiemolloy: Ah yes I was looking at the Bullet Chart variables inside of this document: https://docs.google.com/document/d/1cw10pJFXWruB1SA8TQwituxn5Ss6KpxYPCOYGrH8qAY/edit which I assumed was the latest. Maybe zero-line and negative-primary-measure need to be called something different to match up with your names? Or if you think its redundant I can remove them.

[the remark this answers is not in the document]
christiemolloy: Also updated it with those props you mentioned above ^

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Candlestick
+$pf-chart-candelstick--data--stroke--Width: 1;
+$pf-chart-candelstick--data--stroke--Color: $pf-color-black-900;
+$pf-chart-candelstick--candle--positive--Color: #fff;
```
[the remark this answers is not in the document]
christiemolloy: Yep added it in, thanks for finding that

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-candelstick--candle--negative--Color: $pf-color-black-900;
+
+// Chart
+$pf-chart-chart--BorderWidth: 1 solid;
+
+// Donut Chart
+$pf-chart-donut--threshold--first--Color: $pf-color-black-200;
+$pf-chart-donut--threshold--second--Color: $pf-color-black-300;
+$pf-chart-donut--threshold--third--Color: $pf-color-black-400;
```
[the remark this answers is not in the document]
christiemolloy: Broke up those variables into two color scales, one for those grey background colors and then one for the colored indicators that sit on top.

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-donut--threshold--success--Color: $pf-chart-color-blue-300;
+$pf-chart-donut--threshold--warning--Color: $pf-chart-color-gold-400;
+$pf-chart-donut--threshold--danger--Color: $pf-color-red-100;
+$pf-chart-donut--label--subtitle--Color: $pf-color-black-500;
```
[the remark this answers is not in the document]
christiemolloy: @gdoyle1 should these font sizes be 14 and 24 for the donut, instead of 16 and 26?

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Error Bar
+$pf-chart-errorbar--BorderWidth: 8;
+$pf-chart-errorbar--data--BackgroundColor: transparent;
+$pf-chart-errorbar--data--Opacity: 1;
+$pf-chart-errorbar--data-stroke--Width: 2;
+$pf-chart-errorbar--data-stroke--Color: $pf-color-black-900;
+
+// Group
+$pf-chart-group--grey-scale-color-scale: ($pf-color-black-900, $pf-chart-color-black-500, $pf-chart-color-black-400, $pf-chart-color-black-300, $pf-chart-color-black-200, $pf-chart-color-black-100); // grayscale theme
```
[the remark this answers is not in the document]
christiemolloy: Ok so should I remove all color scale variables from this sheet? Makes sense that you just use the colors already merged but want to confirm? @dlabrecq

src/patternfly/_chart-globals.scss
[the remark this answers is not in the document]
gdoyle1: @christiemolloy Yes, you're correct!

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-pie--Padding: 8;
+$pf-chart-pie--data--Padding: 8;
+$pf-chart-pie--data--stroke--Width: 1;
+$pf-chart-pie--data--stroke--Color: transparent;
+$pf-chart-pie--labels--Padding: 8;
+$pf-chart-pie--static--Height: 230;
+$pf-chart-pie--static--Width: 230;
+$pf-chart-pie--dynamic--Height: 202;
+$pf-chart-pie--dynamic--Width: 202;
+$pf-chart-pie--angle--Padding: 1;
```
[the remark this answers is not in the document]
christiemolloy: separated these donuts out. was a little confused, the title for the threshold donuts is "utilization" and the title for the utilization donuts is "threshold"

gina also confirmed that those font sizes should be 14/24 to match our globals.

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,175 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Axis Chart
+$pf-chart-axis--axis--stroke--Width: 1;
+$pf-chart-axis--axis--stroke--Color: $pf-color-black-900;
+$pf-chart-axis--axis--BackgroundColor: transparent;
+$pf-chart-axis--axis-label--Padding: 100;
+$pf-chart-axis--axis-label--stroke--Color: transparent;
+$pf-chart-axis--grid--BackgroundColor: transparent;
+$pf-chart-axis--grid--stroke--Color: $pf-color-black-200;
+$pf-chart-axis--grid--BorderColor: transparent;
```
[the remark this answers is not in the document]
christiemolloy: Ah thanks replaced with stroke--Color, I think it was from the start where I thought stroke == border

christiemolloy:
Updated @dlabrecq

src/patternfly/_chart-globals.scss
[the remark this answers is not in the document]
dlabrecq: Would you mind leaving the warning and error colors as is? Those colors are applied individually, so they're not really a good candidate for a color scale. Just the grey threshold colors are a candidate for color scales.

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,179 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-candelstick--data--stroke--Color: $pf-color-black-900;
+$pf-chart-candelstick--candle--positive--Color: $pf-color-white;
+$pf-chart-candelstick--candle--negative--Color: $pf-color-black-900;
+
+// Chart
+$pf-chart-chart--BorderWidth: 1 solid;
+
+// Donut Threshold Chart
+$pf-chart-donut--threshold--background--color-scale: ($pf-color-black-200, $pf-color-black-300, $pf-color-black-400);
+$pf-chart-donut--threshold--indicator--color-scale: ($pf-chart-color-blue-300, $pf-chart-color-gold-400, $pf-color-red-100);
```
dlabrecq: Thinking we should just remove all color scale vars. Just define the colors.

In react, we need to create an array, so it will likely be easier to work with individual colors. For a blue theme; for example, we merge blue colors on top of this color scale.

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,179 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-donut--threshold--dynamic--pie--Height: 202;
+$pf-chart-donut--threshold--dynamic--pie--Width: 202;
+$pf-chart-donut--threshold--static--pie--Height: 230;
+$pf-chart-donut--threshold--static--pie--angle--Padding: 1;
+$pf-chart-donut--threshold--static--pie--Width: 230;
```
dlabrecq: Can we add the font sizes here for title (26) and subtitle (16)?

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,179 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+
+// Donut Utilization Chart
+$pf-chart-donut--utilization--label--subtitle--fill: $pf-chart-color-black-200;
+$pf-chart-donut--utilization--label--orientation: "vertical";
+$pf-chart-donut--utilization--dynamic--pie--Height: 230;
+$pf-chart-donut--utilization--dynamic--pie--angle--Padding: 1;
+$pf-chart-donut--utilization--dynamic--pie--Width: 230;
```
dlabrecq: Can we add the font sizes here for title (26) and subtitle (16)?

src/patternfly/_chart-globals.scss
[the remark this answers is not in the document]
dlabrecq: I'm sorry, the way you had it is probably best.

In react, we need to create an array, so working with individual colors will work better.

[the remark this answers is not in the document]
dlabrecq: We should just remove all color scales

src/patternfly/_chart-globals.scss
dlabrecq: Can we also add properties for the donut chart? We have pie, donut, donut threshold, and donut utilization charts.

Screen Shot 2019-06-11 at 10 18 12 AM

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,189 @@ $pf-chart-color-black-200: $pf-color-black-400;
 $pf-chart-color-black-300: $pf-color-black-500;
 $pf-chart-color-black-400: $pf-color-black-600;
 $pf-chart-color-black-500: $pf-color-black-700;
+
+// typography
+$pf-chart-global--FontSize--xs: 12; // Axis Labels
```
mcoker: Are we going to stick with unitless numbers or add units so it's clear what this is supposed to convey? The plan was that this stylesheet should also serve as the base styles for someone building a charting library that doesn't use victory.

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,189 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-donut--pie--angle--Padding: 1;
+$pf-chart-donut--pie--Width: 230;
+
+// Donut Threshold Chart
+$pf-chart-donut--threshold--first--Color: $pf-color-black-200;
+$pf-chart-donut--threshold--second--Color: $pf-color-black-300;
+$pf-chart-donut--threshold--third--Color: $pf-color-black-400;
+$pf-chart-donut--threshold--success--Color: $pf-chart-color-blue-300;
+$pf-chart-donut--threshold--warning--Color: $pf-chart-color-gold-400;
+$pf-chart-donut--threshold--danger--Color: $pf-color-red-100;
```
mcoker: for vars like this where we reference the PF4 color var directly, I'm curious if we should...

1) add this as another `$pf-chart-color-red-[...]` color?
2) add another layer of meaningful color var names, like we have in core, for `$pf-chart-global--danger-color--100` and use it where we use "danger" colors?
3) if we don't want to do that, would it be better to reference `$pf-global--danger-color--100` instead of `$pf-color-red-100` so that this danger chart color always aligns with our primary danger color in core?

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,189 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+// Donut Utilization Chart
+$pf-chart-donut--utilization--label--subtitle--fill: $pf-chart-color-black-200;
+$pf-chart-donut--utilization--label--orientation: "vertical";
+$pf-chart-donut--utilization--dynamic--pie--Height: 230;
+$pf-chart-donut--utilization--dynamic--pie--angle--Padding: 1;
+$pf-chart-donut--utilization--dynamic--pie--Width: 230;
+
+// Error Bar
+$pf-chart-errorbar--BorderWidth: 8;
+$pf-chart-errorbar--data--fill: transparent;
```
mcoker: `fill` is a CSS property so you can capitalize that, too.

src/patternfly/_chart-globals.scss
christiemolloy: spoke with @gdoyle1 and she said it would be good to replace those red colors with `$pf-global--danger-color--100` and `pf-color-black-900` with the global text color. Updated

src/patternfly/_chart-globals.scss
dlabrecq: I would prefer that we stick with unitless numbers. We don't currently have a solution for react-tokens to prune units out

src/patternfly/_chart-globals.scss
dlabrecq: I believe these particular vars will be specific to the donut utilization chart. 

We could reference global colors in react, but I suspect this will make it easier to override just the donut utilization chart?

dlabrecq: I believe these particular vars will be specific to the donut utilization chart. For example, the bullet chart has different color thresholds for warning and error.

We could reference global colors in react, but I suspect this will help make it easier to override?

src/patternfly/_chart-globals.scss
mcoker: Since it's all built out now, I talked to @dgutride and we think it's going to be difficult to position this as a generic stylesheet for other libraries to use, so seems fine to make it specific to victory.

src/patternfly/_chart-globals.scss
```diff
@@ -50,3 +51,189 @@ $pf-chart-color-black-200: $pf-color-black-400;
...
+$pf-chart-global--stroke-line-cap: "round";
+$pf-chart-global--stroke-line-join: "round";
+$pf-chart-global--stroke-dash-array: "10, 5";
+
+// Individual Charts
+
+// Area Chart
+$pf-chart-area--Opacity: .4;
+$pf-chart-area--stroke--Width: 2;
+$pf-chart-area--data--Fill: $pf-global--Color--100;
```
mcoker: There are only a handful of pf core variables that we're using here as values for chart specific vars. Seems odd that we have a mix of pf4 and chart specific vars as the values. Why wouldn't we create the same kind of system we have in core, like 

```
$pf-chart-color-black-600: $pf-color-black-900; // add this to the global colors

$pf-chart-global--Fill--Color--100: $pf-chart-color-black-600;
$pf-chart-global--Fill--Color--200: $pf-chart-color-black-200;
$pf-chart-global--Fill--Color--300: $pf-chart-color-black-300;
...
$pf-chart-global--stroke--Color--100: $pf-chart-color-black-600;
...
$pf-chart-global--warning--Color--100: $pf-chart-color-orange-300;
$pf-chart-global--warning--Color--200: $pf-chart-color-gold-300;
...
$pf-chart-global--danger--Color--100: $pf-global--danger-color--100;
...
```

Then all of the values for the chart specific variables could use chart global variables. Seems like there is a common enough pattern in the use of colors for chart properties to do something like that.

dlabrecq:
Are the font sizes in rem? I’m thinking we will need vars in pixels. The SVGs cannot use rem.

src/patternfly/_chart-globals.scss
christiemolloy: Okay I updated the file so that there are global fill colors. For colors like black-200 and black-900 I didn't add those to the color family of 5 as these should be kept to the family of 5 to match the rest of the color sets. For the global fill variables that I just added, they reference the global colors rather than the chart color families, in case those families change.

src/patternfly/_chart-globals.scss
```diff
@@ -76,16 +77,29 @@ $pf-chart-global--stroke-line-cap: "round";
 $pf-chart-global--stroke-line-join: "round";
 $pf-chart-global--stroke-dash-array: "10, 5";
 
+// Fills and Strokes
+$pf-chart-global--danger--Color--100: $pf-global--danger-color--100;
+$pf-chart-global--warning--Color--100: $pf-color-orange-300;
```
christiemolloy: Are there meant to be two variations of warning colors? @gdoyle1 looks like one is orange and one is gold.

dlabrecq:
One last request to add a two margin vars for the donut label and legend. This is due to a design change so we can position the subtitle to the right or bottom of a donut chart.

Screen Shot 2019-06-13 at 4 51 47 PM

dlabrecq:
Looks like this is ready to go. Can it be merged?

mcoker:
@dlabrecq I'd like to review it again. I'll do that after lunch today.

src/patternfly/components/Nav/examples/index.js
```diff
@@ -30,7 +30,7 @@ import docs from '../docs/code.md';
 
 export const Docs = docs;
 
-export default (props) => {
+export default props => {
```
mcoker: did you mean to commit this?

@redallen do you know if this is necessary? Looks like every component/layout has them except for

```
marvin:patternfly-next cmichael$ grep -ri 'export default' src/ | grep props | grep -v '(props)'
src//patternfly/components/Alert/examples/index.js:export default props => {
src//patternfly/components/AppLauncher/examples/index.js:export default props => {
src//patternfly/components/NotificationBadge/examples/index.js:export default props => {
src//patternfly/components/AlertGroup/examples/index.js:export default props => {
```

mcoker:
* we use danger and error in the var names, and values for each in the other's namespace - should we normalize and use one or the other or do they represent different things? 

* should we have vars for stroke/border width? (1 & 2 mostly it seems)

* should we have vars for padding/gutter (or just spacer like core)? (1, 8, 10, 16, 20, maybe 50?)
    * I don't want to get us into a situation where if we introduce a new var, it messes up the naming. For example, if `$spacer-1: 1; $spacer-2: 8;` and we add a new one as "6", what var do we call it? Do we have the concept of a "spacer system" here?

* we use "230" for pie height/width in multiple spots - should that be a var?

* does `$pf-chart-global--layout--Height` and `$pf-chart-global--layout--Width` need to be a var since we don't define height/width anywhere else (except maybe pie above)? Or should we create a var for all of the chart width/heights if they apply?

* in this block below, we don't normally name the concept var by the color var it represents. That kinda defeats the purpose - if we change one of the "fill" vars to a different color, we wouldn't want to have to rename the fill var. Do we have a fill (looks like this is also used for stroke) color system, where we would have `Fill--Color--100, Fill--Color--200, etc`, similar to how we have other vars defined like that (ie, `$pf-chart-global--warning--Color--100`, `$pf-chart-global--warning--Color--200`, etc)

```
$pf-chart-global--Fill--Color--900: $pf-color-black-900;
$pf-chart-global--Fill--Color--700: $pf-color-black-700;
$pf-chart-global--Fill--Color--500: $pf-color-black-500;
$pf-chart-global--Fill--Color--400: $pf-color-black-400;
$pf-chart-global--Fill--Color--300: $pf-color-black-300;
$pf-chart-global--Fill--Color--200: $pf-color-black-200;
$pf-chart-global--Fill--Color--white: $pf-color-white;
```

christiemolloy:
@dlabrecq could you provide some insight here. The layout vars are used in alot of charts right so they should stay global? or could it be moved to under the globals like the legend/axis? I think what @mcoker is saying is that we declare it at the top of the file but its not used in each chart type 

> does $pf-chart-global--layout--Height and $pf-chart-global--layout--Width need to be a var since we don't define height/width anywhere else (except maybe pie above)? Or should we create a var for all of the chart width/heights if they apply?

christiemolloy:
@mcoker I hadn't done that yet because I wasn't sure how extensive the system will get. @gdoyle1 @dlabrecq do you know if we expect to add more colors to these fill colors or is this set ? 

> in this block below, we don't normally name the concept var by the color var it represents. That kinda defeats the purpose - if we change one of the "fill" vars to a different color, we wouldn't want to have to rename the fill var. Do we have a fill (looks like this is also used for stroke) color system, where we would have Fill--Color--100, Fill--Color--200, etc, similar to how we have other vars defined like that (ie, $pf-chart-global--warning--Color--100, $pf-chart-global--warning--Color--200, etc)
> 
> $pf-chart-global--Fill--Color--900: $pf-color-black-900;
> $pf-chart-global--Fill--Color--700: $pf-color-black-700;
> $pf-chart-global--Fill--Color--500: $pf-color-black-500;
> $pf-chart-global--Fill--Color--400: $pf-color-black-400;
> $pf-chart-global--Fill--Color--300: $pf-color-black-300;
> $pf-chart-global--Fill--Color--200: $pf-color-black-200;
> $pf-chart-global--Fill--Color--white: $pf-color-white;

christiemolloy:
@mcoker I don't think 230 should be a variable because pies/donuts overlap anyway so that seems clear .

christiemolloy:
@dlabrecq @gdoyle1 what do you think about using a spacer system here for padding? The values in the charts seem very much component specific, I can't see the use of a global spacer system here really. IMO I see a spacer system using a value set that has a pattern to how it increments.

> should we have vars for padding/gutter (or just spacer like core)? (1, 8, 10, 16, 20, maybe 50?)
> 
> I don't want to get us into a situation where if we introduce a new var, it messes up the naming. For example, if $spacer-1: 1; $spacer-2: 8; and we add a new one as "6", what var do we call it? Do we have the concept of a "spacer system" here?
>

christiemolloy:
@gdoyle1 @dlabrecq  do you know if error and danger are the same thing?

mcoker:
> The layout vars are used in alot of charts right so they should stay global?

Ah ok that makes sense. 

>  I don't think 230 should be a variable because pies/donuts overlap anyway so that seems clear .

I just mean that the width/height 230 pair is defined in 4 spots, so seems easier to maintain if that's a common var instead of hardcoded in each place, as long as we want those to display consistently. 

For example, this...
```
$pf-chart-donut--pie--Height: 230;
$pf-chart-donut--pie--Width: 230;
$pf-chart-donut--threshold--static--pie--Height: 230;
$pf-chart-donut--threshold--static--pie--Width: 230;
$pf-chart-donut--utilization--dynamic--pie--Height: 230;
$pf-chart-donut--utilization--dynamic--pie--Width: 230;
$pf-chart-pie--Height: 230;
$pf-chart-pie--Width: 230;
```

could become....
```
$pf-chart-donut--pie--Height: 230;
$pf-chart-donut--pie--Width: 230;
$pf-chart-donut--threshold--static--pie--Height: $pf-chart-donut--pie--Height;
$pf-chart-donut--threshold--static--pie--Width: $pf-chart-donut--pie--Width;
$pf-chart-donut--utilization--dynamic--pie--Height: $pf-chart-donut--pie--Height
$pf-chart-donut--utilization--dynamic--pie--Width: $pf-chart-donut--pie--Width;
$pf-chart-pie--Height: $pf-chart-donut--pie--Height;
$pf-chart-pie--Width: $pf-chart-donut--pie--Width;
```

gdoyle1:
@christiemolloy Danger and error are the same, yes - I think more products are using "error" now instead of danger. I think a spacer system would be a good thing to look into making for charts. @mceledonia did you picture using the same number system as we have for our other spacers or a different incremental order? I'm not 100% sure about the colors thing although I don't see more colors being added to the chart colors. @mceledonia would you agree?

mceledonia:
@gdoyle1 yeah I think we can use the same spacer system. I used it when designing the charts, but I have a feeling it will really only come into play for padding of things like axis labels and such. Whether or not it makes sense on the dev side to do that, I'm not confident enough to give an answer there..

dlabrecq:
The width/height 230 is common to donut and donut utilization and donut threshold. However, they are separate components. I don't know that they should share global var here. Ideally, all components would have their own vars.

dlabrecq:
It's my understanding that the layout vars are global because they are expected to remain consistent for all. Although, they're not used by all components; for example, legend, pie, tooltip. It's your call as to whether they are global -- I don't feel strongly about the layout props.

I don't see a global spacer system being used here. Maybe we can speak more about what that means?

We have a default padding of 50px for the chart axis. Legend has a gutter of 20px, which probably should be 10px, IMO. Other than that, we only have a couple places where we use a padding of 8px (label margin) and 16px (legend margin)

dlabrecq:
$pf-chart-global--layout--Height and $pf-chart-global--layout--Width would apply to all charts as a defult height and width. Donuts don't need to be that large.

I would like to have separate vars for padding/gutter and  stroke/border width. These properties have very different effects on charts -- they are not the same.

christiemolloy:
- If error and danger are the same do I need to make a new global variable for error?
- I added vars for stroke and border width but kept them separate
- I agree with @dlabrecq when he said that we shouldn't change the spacers until he tests it in React

@mcoker what do you think

mcoker:
> If error and danger are the same do I need to make a new global variable for error?

I think what you have now is fine. We use danger in the global color var names, which matches core. And I'm assuming the use of danger and error in the chart specific vars matches the theme properties.

> I added vars for stroke and border width but kept them separate

Did you push those changes up? I don't see them.

> I agree with @dlabrecq when he said that we shouldn't change the spacers until he tests it in React

👍

christiemolloy:
@mcoker here is the change: https://github.com/patternfly/patternfly-next/pull/1863/commits/9ee4958aded27cdbb360a9b9ccba35bf535b9290 I must not have pushed it up

mcoker:
@christiemolloy lgtm!

Looks like you missed one:

> $pf-chart-area--stroke--Width: 2;

And I'm not sure what this is - would it apply as a stroke/border, too? Or is it something else?

> $pf-chart-bullet--threshold-error--Width: 2;

christiemolloy:
Updated! $pf-chart-bullet--threshold-error--Width: 2; is just a width not a stroke or border.

christiemolloy:
@dlabrecq and @mcoker can you re-approve this PR, I just added the files from `_chart-globals.scss` to `patternfly-charts.scss`

src/patternfly/patternfly-charts.scss
```diff
@@ -53,4 +53,215 @@
...
+  --pf-chart-legend--title--orientation: #{$pf-chart-legend--title--orientation};
+  --pf-chart-legend--data--type: #{$pf-chart-legend--data--type};
+  --pf-chart-legend--title--Padding: #{$pf-chart-legend--title--Padding};
+  --pf-chart-legend--Margin: #{$pf-chart-legend--Margin};
+
+  // Line Chart
+  --pf-chart-line--data--Fill: #{$pf-chart-line--data--Fill};
+  --pf-chart-line--data--Opacity: #{$pf-chart-line--data--Opacity};
+  --pf-chart-line--data--stroke--Width: #{$pf-chart-line--data--stroke--Width};
+  --pf-chart-line--data--stroke--Color: #{$pf-chart-global--Fill--Color--900};
```
mcoker:
```suggestion
  --pf-chart-line--data--stroke--Color: #{$pf-chart-line--data--stroke--Color};
```

PR #1871, opened 2019-05-30
fix(form/login/wizard): fixed undefined issues with IE11 script

dlabaj:
Fixed order of variables which was causing the IE conversion script to generate undefined.  Also removed a login variable that was not needed.  This was also causing the IE scripts to generate undefined. 

This resolves issue: https://github.com/patternfly/patternfly-next/issues/1869

### DIFF
--- a/src/patternfly/components/Login/login.scss
+++ b/src/patternfly/components/Login/login.scss
@@ -76,7 +76,6 @@
   --pf-c-login__main-footer-links--PaddingRight: var(--pf-global--spacer--3xl);
   --pf-c-login__main-footer-links--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-login__main-footer-links--PaddingLeft: var(--pf-global--spacer--3xl);
-  --pf-c-login__main-footer-links--MarginBottom: calc(var(--pf-c-login__main-footer-links--MarginBottom) * -1);
   --pf-c-login__main-footer-links-item--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-login__main-footer-links-item--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-login__main-footer-links-item--MarginBottom: var(--pf-global--spacer--sm);
@@ -238,7 +237,6 @@
   flex-wrap: wrap;
   justify-content: center;
   padding: var(--pf-c-login__main-footer-links--PaddingTop) var(--pf-c-login__main-footer-links--PaddingRight) var(--pf-c-login__main-footer-links--PaddingBottom) var(--pf-c-login__main-footer-links--PaddingLeft);
-  margin-bottom: var(--pf-c-login__main-footer-links--MarginBottom);
 }
 
 .pf-c-login__main-footer-links-item {
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -45,6 +45,38 @@
   --pf-c-wizard__description--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-wizard__description--Color: var(--pf-global--Color--light-200);
 
+  // Nav Link Before
+  --pf-c-wizard__nav-link--before--Width: 1.5rem;
+  --pf-c-wizard__nav-link--before--Height: 1.5rem;
+  --pf-c-wizard__nav-link--before--Top: 0;
+
+  // Toggle
+  --pf-c-wizard__toggle--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-wizard__toggle--ZIndex: var(--pf-global--ZIndex--sm);
+  --pf-c-wizard__toggle--BoxShadow: var(--pf-global--BoxShadow--md-bottom);
+  --pf-c-wizard__toggle--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-wizard__toggle--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-wizard__toggle--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-wizard__toggle--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-wizard__nav-link--before--Width) + var(--pf-global--spacer--sm));
+
+  // Toggle number
+  --pf-c-wizard__toggle-num--before--Top: #{pf-size-prem(2px)};
+
+  // Toggle list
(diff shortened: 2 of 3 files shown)


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -59,7 +64,6 @@
 
   // Toggle list
   --pf-c-wizard__toggle-list--MarginRight: var(--pf-global--spacer--sm);
-  --pf-c-wizard__toggle-list--MarginBottom: calc(var(--pf-c-wizard__toggle-list-item--MarginBottom) * -1);
```
mcoker: Did you mean to remove this? Seems like you want to just move it below where `--pf-c-wizard__toggle-list-item--MarginBottom` is defined.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -270,7 +271,6 @@
   flex-wrap: wrap;
   align-items: baseline;
   margin-right: var(--pf-c-wizard__toggle-list--MarginRight);
-  margin-bottom: var(--pf-c-wizard__toggle-list--MarginBottom);
```
mcoker: Then add this back if you add `--pf-c-wizard__toggle-list--MarginBottom` back.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -115,9 +119,6 @@
   --pf-c-wizard__nav-link--m-disabled--Color: var(--pf-global--Color--dark-200);
 
   // Nav link number
-  --pf-c-wizard__nav-link--before--Width: 1.5rem;
-  --pf-c-wizard__nav-link--before--Height: 1.5rem;
-  --pf-c-wizard__nav-link--before--Top: 0;
```
mcoker: instead of moving just these 3 lines, can you move this whole block up so that the `// Nav link number` CSS is all in one spot?

src/patternfly/components/Login/login.scss
```diff
@@ -76,7 +76,6 @@
   --pf-c-login__main-footer-links--PaddingRight: var(--pf-global--spacer--3xl);
   --pf-c-login__main-footer-links--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-login__main-footer-links--PaddingLeft: var(--pf-global--spacer--3xl);
-  --pf-c-login__main-footer-links--MarginBottom: calc(var(--pf-c-login__main-footer-links--MarginBottom) * -1);
```
srambach: Same question as @mcoker had on removing `--pf-c-wizard__toggle-list--MarginBottom` from wizard.scss - did you mean to remove it completely or just move this line below `--pf-c-login__main-footer-links--MarginBottom`?

srambach: Wait, nevermind - I see this is actually self-referential?

mcoker: Yep. If you look now the margin there just computes to 0 since `--pf-c-login__main-footer-links--MarginBottom` isn't defined anywhere else.

src/patternfly/components/Wizard/wizard.scss
dlabaj: I did mean to remove it because this one also came out to be zero.  I'll double check to make sure.

mcoker: it's `-4px` for me.

dlabaj: yeah my mistake thanks will add it back.

src/patternfly/components/Wizard/wizard.scss
dlabaj: Done.

src/patternfly/components/Wizard/wizard.scss
dlabaj: Done.

dlabaj:
@srambach and @mcoker Thanks for the feedback.  I think it's good now.  Let me know if you see anything else.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -106,18 +117,7 @@
   // Nav item
   --pf-c-wizard__nav-item--MarginTop: var(--pf-global--spacer--md);
 
-  // Nav link
```
mcoker: Oh you moved this block instead of the `// Nav link number block` 

It should be 

```
  // Nav link
  --pf-c-wizard__nav-link--Color: var(--pf-global--Color--100);
  --pf-c-wizard__nav-link--TextDecoration: var(--pf-global--link--TextDecoration);
  --pf-c-wizard__nav-link--hover--Color: var(--pf-global--link--Color);
  --pf-c-wizard__nav-link--focus--Color: var(--pf-global--link--Color);
  --pf-c-wizard__nav-link--m-current--Color: var(--pf-global--link--Color);
  --pf-c-wizard__nav-link--m-disabled--Color: var(--pf-global--Color--dark-200);

// Nav link number
  --pf-c-wizard__nav-link--before--Width: 1.5rem;
  --pf-c-wizard__nav-link--before--Height: 1.5rem;
  --pf-c-wizard__nav-link--before--Top: 0;
  --pf-c-wizard__nav-link--before--BackgroundColor: var(--pf-global--BackgroundColor--300);
  --pf-c-wizard__nav-link--before--BorderRadius: var(--pf-global--BorderRadius--lg);
  --pf-c-wizard__nav-link--before--Color: var(--pf-global--Color--100);
  --pf-c-wizard__nav-link--before--FontSize: var(--pf-global--FontSize--sm);
  --pf-c-wizard__nav-link--before--Transform: translateX(calc(-100% - var(--pf-global--spacer--sm)));
  --pf-c-wizard__nav-link--m-current--before--BackgroundColor: var(--pf-global--active-color--100);
  --pf-c-wizard__nav-link--m-current--before--Color: var(--pf-global--Color--light-100);
  --pf-c-wizard__nav-link--m-disabled--before--BackgroundColor: transparent;
  --pf-c-wizard__nav-link--m-disabled--before--Color: var(--pf-global--Color--dark-200);
```

src/patternfly/components/Login/login.scss
srambach: As long as this wasn't a mistake where it really should have been assigned to another variable. I'm not sure why it would have been put in there that way. But, I don't know what the original specs were.

mcoker: Yeah I looked at the commit where this was introduced (https://github.com/patternfly/patternfly-next/pull/1281) and looks like a mistake. The layout matches the design in sketch, too.

redallen:
You all are crazy fast with this, thanks for helping my #1876 :)

PR #1874, opened 2019-05-31
fix(switch): add focus indicator

### LINKED ISSUE #1824: Focus state for switch
Since the Switch is not a native html element (like a button or text input), but instead is a checkbox styled to look like a switch, we cannot use native browser styles for indicating focus.

For this issue, we should add styles that display on focus to indicate to the user when the switch has focus.

srambach:
This adds a focus indicator to the switch component. 
After discussion with @mceledonia, we'll go with a 2px #06C border, 8 px from the edge, but no fill. This allows us to use the outline property rather than creating a pseudo-element, which would have potentially caused z-index problems.

Fixes #1824

### DIFF
--- a/src/patternfly/components/Switch/examples/switch-no-label-example.hbs
+++ b/src/patternfly/components/Switch/examples/switch-no-label-example.hbs
@@ -5,7 +5,8 @@
     {{#> switch-toggle-icon}}{{/switch-toggle-icon}}
   {{/switch-toggle}}
 {{/switch}}
-
+<br/>
+<br/>
 {{#> switch switch--attribute='for="switch-with-icon-2"'}}
   {{#> switch-input id="switch-with-icon-2" switch-input--attribute='name="switchExample4"'}}{{/switch-input}}
   {{#> switch-toggle}}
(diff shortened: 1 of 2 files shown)


srambach:
To quote @mceledonia, 4px rather than the 8px used felt "like an impending doom slowly closing in on the innocent switch." 😂Also, he felt that the switches shouldn't be placed that close together.

mceledonia:
Yeah I felt like the 4px was just a bit too tight, but if everyone disagrees I can concede :) 

I certainly want to avoid that above - hopefully we can do that with spacing as each toggle should have some kind of label which keeps them further apart. I can't think of many other areas where we have two components less than 8px away from each other... thinking of things like the toolbar, button layouts, etc..

christiemolloy:
Ok thanks for explaining @mceledonia ! So this is with 8px margin between them and I think that will work

Screen Shot 2019-06-04 at 9 16 18 AM

mceledonia:
Awesome, yeah let's go with that for now... if we have problems in the future we can always bring it down.

mattnolting:
Beautiful 👍

PR #1876, opened 2019-05-31
feat(ie11): add ie11 to build

redallen:
Use existing conversion utilities to write `patternfly-ie.css` which contains all components + base styles.

Closes https://github.com/patternfly/patternfly-react/issues/2092

### DIFF
--- a/build/npm-scripts/ie-convert-all.js
+++ b/build/npm-scripts/ie-convert-all.js
@@ -14,7 +14,12 @@ getStylesheetPaths(pfStylesheetsGlob, [], [])
     }, '')
   )
   .then(concatCss => transform(concatCss, patternflyBasePath))
-  .then(transformedCss => transformedCss.replace(/\.\.\/\.\.\/assets/gm, './assets'))
+  .then(transformedCss => {
+    if (transformedCss.indexOf('undefined') >= 0) {
+      throw new Error('Stylesheet contains undefined');
+    }
+    return transformedCss.replace(/\.\.\/\.\.\/assets/gm, './assets');
+  })
   .then(ie11ReadyStylesheet => fs.writeFileSync(outPath, ie11ReadyStylesheet))
   .catch(error => {
     throw new Error(error);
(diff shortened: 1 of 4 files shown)


seanforyou23:
Actually just realized I hadn't built patternfly yet, which I suppose others may also do when developing locally. Could we add something like the following to `getStylesheetPaths`? You could add it [right before the filter](https://github.com/patternfly/patternfly-next/blob/master/build/npm-scripts/ie-conversion-utils.js#L116).

```js
if (files.length === 0) {
  throw new Error(`getStylesheetPaths didn't receive any stylesheets to parse`);
}
```

seanforyou23:
It would also probably also be a good time to fix the [copy/pasta](https://github.com/patternfly/patternfly-next/blob/master/build/npm-scripts/ie-conversion-utils.js#L105) in the doc, and use something like `/Path/to/patternfly-next/dist/{components,layouts,utilities}/**/*.css` instead.

seanforyou23:
So the main problem I see now is that when an instance of `undefined` is encountered, the script doesn't inform the user of where. The resulting stylesheet just contains `Error: Stylesheet contains undefined`.

What I'd think to do here is first produce the problematic file and after that throw the error. This way users can search the resulting file to get a clue as to what went wrong.

Also, we probably want to handle the error better?

Screen Shot 2019-06-04 at 2 23 17 PM

build/npm-scripts/ie-convert-all.js
```diff
@@ -0,0 +1,37 @@
...
+  )
+  .then(concatCss => transform(concatCss, patternflyBasePath))
+  .then(ie11ReadyStylesheet => {
+    fs.writeFileSync(outPath, ie11ReadyStylesheet);
+    return ie11ReadyStylesheet;
+  })
+  .then(transformedCss => {
+    transformedCss.split('\n').forEach((line, index) => {
+      if (line.indexOf('undefined') >= 0) {
+        throw new Error(`Stylesheet contains undefined at line ${index + 1}`);
```
seanforyou23: Looking good! One last request, to help people find the problematic file faster, could we include something to point users directly to the file? As is, they only know the line number, and need to figure out what file it's in. You could add something like the following just before throwing the error.

```js
console.log(
  `\x1b[31m%s\x1b[0m`,
  `Problem in ${outPath}\n`
);
```

We use these log colors also in the a11y audit, maybe it makes sense to make these shared colors available in a better place than the a11y test config?

Then, users are guided directly to the problem.
Screen Shot 2019-06-05 at 10 39 40 AM

redallen: The [Chalk](https://www.npmjs.com/package/chalk) package exists for the colors, but I just used the hardcoded value for the `console.error`.

PR #1880, opened 2019-06-03
fix(table): add an extra breakpoint

### LINKED ISSUE #1640: add an extra breakpoint to the table
In the data table sortable demo the table needs a larger breakpoint, so that the rows don't overflow outside of their container #1417 

Our breakpoint global vars stop at xl size, I think we need to investigate a var that is bigger than this, to allow more flexibility.

christiemolloy:
close #1640 

This is related to a fix for the table in the website: https://github.com/patternfly/patternfly-org/issues/1021


mattnolting:
> Re: the flex layout, @mattnolting does your work impact this? Seems like all of the breakpoints in the flex layout could be refactored to use the global lists/maps we have already defined for breakpoints?

@mcoker Yep, you're exactly right. The flex refactoring I'm currently working on relies on the spacer and breakpoint maps, rather than explicit values. Change away, this won't affect my work:D

PR #1881, opened 2019-06-03
fix(table): media query matches border

### LINKED ISSUE #1823: Update border for tables to match Media query break point
Not sure if this is a bug or not but the table component adds a 8px bottom border on smaller screens. It might make more sense to have this align with the same break point that stacks the rows so the larger bottom border is shown as soon as the rows stack.
Screen Shot 2019-05-15 at 5 05 55 PM
Screen Shot 2019-05-15 at 5 08 38 PM

christiemolloy:
fix #1823 

Fixes the table so that when the table breaks to grid form, the border width changes to 8px, and the border color matches the background color.

Screen Shot 2019-06-03 at 3 48 32 PM

### DIFF
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -34,6 +34,9 @@
   // Responsive vars
   // ==================================================================
 
+  // Table
+  --pf-c-table--responsive--BorderColor: var(--pf-global--BorderColor--300);
+
   // Body
   --pf-c-table-tbody--responsive--MarginTop: var(--pf-global--spacer--xs);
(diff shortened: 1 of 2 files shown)


mcoker:
did you mean to commit the chart changes in https://github.com/patternfly/patternfly-next/pull/1881/commits/84fad8fbc1e742be209c6387a350399dec1ff271?

mcoker:
Looks like we lost the bottom border on expandable table rows if it's the last thing in the table (expandable, compact expandable, compound expansion)

Screen Shot 2019-06-04 at 2 44 50 PM

christiemolloy:
Thanks so much for catching that @mcoker 

Screen Shot 2019-06-05 at 10 07 02 AM

PR #1886, opened 2019-06-04
fix(datalist): adds breakpoints for actions

### LINKED ISSUE #1704: fix(Data List) Investigate Data List Actions Element
In the new Data List updates from #1695, there is a `pf-c-data-list_item-action` that handles all actions. We need to investigate a way to determine when the buttons in this element should wrap to the next line. Currently it breaks at a lg breakpoint, but if we have 4 buttons in there, this will break.

srambach:
This collapses actions into a kebab menu by using modifiers specifying what breakpoint to show/hide the menu toggle and action buttons.

Closes #1704

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -73,6 +73,7 @@
   }
 
   // Actions
+  --pf-c-data-list__item-action--Display: flex;
   --pf-c-data-list__item-action--PaddingTop: var(--pf-global--spacer--lg);
   --pf-c-data-list__item-action--PaddingBottom: var(--pf-global--spacer--lg);
   --pf-c-data-list__item-action--MarginLeft: var(--pf-global--spacer--xl);
@@ -80,7 +81,6 @@
   --pf-c-data-list__item-action--not-last-child--MarginRight: var(--pf-global--spacer--md);
   --pf-c-data-list__item-action--not-last-child--lg--MarginBottom: var(--pf-global--spacer--md);
   --pf-c-data-list__action--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
-  --pf-c-data-list__item-action--hidden-visible--Display: flex;
 
   @media screen and (max-width: $pf-global--breakpoint--md) {
     --pf-c-data-list__item-action--MarginLeft: var(--pf-c-data-list__item-action--md--MarginLeft);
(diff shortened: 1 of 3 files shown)


src/patternfly/components/DataList/data-list.scss
```diff
@@ -80,6 +80,7 @@
   --pf-c-data-list__item-action--not-last-child--MarginRight: var(--pf-global--spacer--md);
   --pf-c-data-list__item-action--not-last-child--lg--MarginBottom: var(--pf-global--spacer--md);
   --pf-c-data-list__action--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-data-list__item-action--hidden-visible--Display: flex;
```
mcoker: how about we change this to `--pf-c-data-list__item-action--Display: flex;`, then use `var(--pf-c-data-list__item-action--Display)` for the `@include` on L182 and for the `display` property on L184?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -178,6 +179,8 @@
 }
 
 .pf-c-data-list__item-action {
+  @include pf-hidden-visible(var(--pf-c-data-list__item-action--hidden-visible--Display));
```
mcoker: nice!

srambach:
I read the issue as not wanting a wrap in lieu of the breakpoints but @mcarrano can you specify whether you want it to wrap or not?

mcarrano:
When you say wrap here, are you asking if we should start to stack the buttons below a certain breakpoint? That was not my intent. I think that we want to collapse the buttons into the kabob when they no longer fit. Keep in mind that the example here is not a likely one or a design I would encourage. I would not recommend including more than one or two inline action buttons in a row. Maybe we should remove that example?

srambach:
I can make the example whatever you want. I was showing that if you did have a lot, it could collapse earlier than if you had fewer. I assume you would never collapse 1 button, so do you just want one example with 2 buttons that collapse into a kebab? @mcarrano

mcarrano:
I don't feel strongly about whether the 4 button example should be removed or not. I'm struggling with whether you would collapse one button or not. What if you scale down to mobile? A single button could occupy half the screen. But then it's odd to have a kabob with one action. @srambach @christiemolloy what would you expect?

srambach:
I would not want a single button to collapse to a kebab. But it would just not be coded with the modifier saying to collapse it. Or maybe that single button would switch to an icon button. That would be possible with the changes in this PR. This PR gives you the mechanism to show/hide actions depending on a breakpoint. @mcarrano

PR #1890, opened 2019-06-04
fix(button): add class around the icon to fix bug

### LINKED ISSUE #1883: Link Button: look at adjusting spacing
I was using our "Link Button" for the website and I had some feedback that there is no room between the icon and the text, which I also agree. @kybaker or @mceledonia what do you think? Should we look at adjusting it a little?

**Update** React Icons dont take classes, so lets create another div around the icon and add the margin to that div.

Screen Shot 2019-06-04 at 9 22 29 AM

christiemolloy:
fix #1883 

Create follow up issue in React.


src/patternfly/components/Button/button.scss
```diff
@@ -106,6 +106,12 @@
   border: 0;
   border-radius: var(--pf-c-button--BorderRadius);
 
+  .pf-c-button__icon {
+    display: inline;
```
mcoker: You could use a `<span>` as the wrapping element and remove this.

src/patternfly/components/Button/button-icon.hbs
```diff
@@ -1,5 +1,6 @@
-<i class="fas{{#if button-icon--type}} fa-{{button-icon--type}}{{/if}}{{#if button-icon--modifier}} {{button-icon--modifier}}{{/if}}"
-  aria-hidden="true"
+<div class="pf-c-button__icon{{#if button-icon--modifier}} {{button-icon--modifier}}{{/if}}"
```
mcoker: the build is failing because this partial is used just to generate an icon, ie `{{> button-icon button-icon--type="bars"}}` or an icon with the class `pf-c-button__icon`, ie `{{> button-icon button-icon--type="plus-circle" button-icon--modifier="pf-c-button__icon"}}`, so `@partial-block` in this new partial is erroring since nothing is passed to the partial in places that use it currently.

You could test for `{{# if button-icon--modifier == "pf-c-button__icon"}}`, and spit out the wrapping `<div>` and `</div>` (or `<span>` per my comment below), but that feels a little hacky.

I might create a new partial for defining a button icon (just the `<i class="fas ..." aria-hidden="true"></i>` element), or just update the examples that use this partial currently just to be an `<i>` element instead of using a partial to generate the icon. Probably the former. What do you think?

christiemolloy: Ah thats funny i replaced the examples that were using this partial to just use the <i> element but that was only in the Button component. That partial is in hundreds of places, so instead of updating them all I just changed the new partial to `{{#> button-link-icon }}` .. thoughts?

mcoker: This works, but the only issue is the file isn't named by the BEM class used inside of it like the rest of our files. I think it would work better if we called the current `button-icon.hbs` something else like `button-icon-include` or something, though I'm not sure what to call it. That way `button-icon.hbs` can be used with `pf-c-button__icon`.

christiemolloy:
@mcoker  updated

mcoker:
looks like you need to update https://github.com/christiemolloy/patternfly-next/blob/master/src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs since the merge

christiemolloy:
updated @mcoker

src/patternfly/components/NotificationBadge/examples/notification-badge-simple-example.hbs
```diff
@@ -1,11 +1,11 @@
-{{#> button button--modifier="pf-m-icon" button--attribute='aria-label="Unread notifications"'}}
```
mcoker: oh nice, this should have been `pf-m-plain`. Can you add that class to these buttons?

PR #1896, opened 2019-06-05
fix(clipboardcopy): fix focus ring and add content editable to expanded

### LINKED ISSUE #1764: Copy to clipboard component - minor issues
The react implementation highlighted some changes needed:

- [x] Right-hand border on focus for the expand button is cut off by the input
- [x] When using expandable content and after speaking with @mcarrano, we should disable the top input and use a contenteditable div for the section below.  Not sure what the treatment of the div should be (readonly/leave white?)

christiemolloy:
close #1764 

@srambach I know you were working on these focus rings. Do you think it looks odd for only one button to have a negative outline-offset. In tabs they aren't consistent so maybe this is fine for now...

Screen Shot 2019-06-05 at 1 27 40 PM

@mceledonia the input field when the clipboardcopy component is expanded, is now "Read-Only" which means that to be consistent it takes on the gray background of our "Read-Only" form control. I can override the background-color to be white here like it was before. I dont know if that will have a usability problem as users may think it is editable. Was wondering what you thought?

### DIFF
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
@@ -1,4 +1,4 @@
-<div class="pf-c-clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div contenteditable="true" class="pf-c-clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
   {{#unless clipboard-copy--IsExpanded}}hidden{{/unless}}
   {{#if clipboard-copy-expandable-content--attribute}}
     {{{clipboard-copy-expandable-content--attribute}}}
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
@@ -1,4 +1,4 @@
-<div contenteditable="true" class="pf-c-clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div class="pf-c-clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
   {{#unless clipboard-copy--IsExpanded}}hidden{{/unless}}
   {{#if clipboard-copy-expandable-content--attribute}}
     {{{clipboard-copy-expandable-content--attribute}}}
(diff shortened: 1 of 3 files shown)


srambach:
Visuals are not my call, but fwiw, I barely notice the difference when tabbing from one to the other and you'd never have two focused at the same time, which makes it more obvious when shown that way. The outline-offset can be changed on the other controls if it's deemed important but I'd be careful of what else it affects.

mcarrano:
I agree with @srambach here. I took me a bit to see the difference. But not sure I understand why these are different. Is that intentional?

srambach:
@mcarrano The focus outline on the left button was moved in so it's within the bounds of the button because otherwise it was cut off by the input box. It's a one-off because you don't really want to do that for all buttons. The right button didn't need it because it's not cut off.

mceledonia:
@christiemolloy Yeah let's keep it consistent with read-only style for now. With this in mind, in the future we might want to refine the read-only styling to work a little better in an input group like the above. Something more subtle.. I wonder if we even need a visual change at all, or if the interaction of being not editable is enough.

mcoker:
FWIW, if the copy to clipboard is expandable, in the expanded state, the "readonly" and regular copy to clipboard components look exactly the same. That seems confusing.

Also I think we have a bug in the react component where if you make the expanded copy to clipboard example `isReadOnly`, you can still edit it by editing the text in the expanded panel. @mcarrano is that a bug or is that intentional?

IMO this is how I would expect a normal (not readonly) copy to clipboard component to look:

Screen Shot 2019-06-06 at 3 16 50 PM

And this is how I would expect it to look like if it's readonly (and you can't edit the text in the input or the expanded box below)

Screen Shot 2019-06-06 at 3 17 05 PM

We also have a bit of a focus ring issue with the expanded content as you can see in the screenshot above.

mcarrano:
> Also I think we have a bug in the react component where if you make the expanded copy to clipboard example isReadOnly, you can still edit it by editing the text in the expanded panel. @mcarrano is that a bug or is that intentional?

Sounds like a bug

christiemolloy:
I actually dont think thats a bug. From my understanding, the content in the expanded section should match the content in the input above at all times as its just an expanded version of the text. So when its expanded, the input should be read only, and only the expanded section should be editable.

And @mcoker to your point about how you expect it to work, I think that what this issue is trying to address, is that the input should always be read-only when the content is expanded. 

But @mcarrano could you help out here

christiemolloy:
Updated:

Screen Shot 2019-06-07 at 10 51 40 AM

mcarrano:
I'm sorry. I think I misunderstood @mcoker 's question. @christiemolloy your description is correct. When the field is expanded, the user can edit the text in the expansion area and those changes are reflected back in the read-only field.

mcoker:
> So when its expanded, the input should be read only, and only the expanded section should be editable.

Why couldn't both be editable?

And if the copy to clipboard is `isReadOnly`, when it's expanded, shouldn't the content in the expandable area not be editable? I would expect that if it's an editable copy to clipboard, whether it's expandable or not, you can edit either the input or the expanded area. And if it's `isReadOnly` you can't edit either of them.

mcarrano:
I can't remember the exact reasoning, but I looked at this with @dgutride prior to the April release and we decided on the current approach. It was something to do with the difficulty in keeping the two fields in sync.

christiemolloy:
I updated this with a varation of the read only @mcoker @mcarrano 

@mceledonia could you also check this out to see if you think that this styling for the read-only expansion is ok?

mcoker:
@mcarrano the part you replied to is a bug, I believe. If you have an expandable, read only copyclipboard component, in the react implementation, you can still edit what's in the form control input by editing the expanded section. Seems like if the component is read only, neither the input or expanded section should be editable.

mcarrano:
Oh,OK I see what you are saying @mcoker . Sorry if I'm finding this confusing. Yes, if it's read-only then I should not be able to edit the text, period. If it's editable and expandable, then I should only be able to edit from the expandable text area when the expansion is open.

@christiemolloy is that not how it is now?

christiemolloy:
@mcarrano yes thats how it is in Core, we should file the issue in React, I can

mattnolting:
Lgtm!

PR #1898, opened 2019-06-05
docs(card): update action order in img/action examples, rename demo

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1897


christiemolloy:
What was the reasoning behind the example you removed? I remember seeing a product's card design have a an image in the body and actions parallel to the right of it. Cant seem to find it now. But if this is a design treatment going forward then that seems ok. I guess if they want to achieve the body next to actions they can.

mcoker:
> BTW, I'm assuming that the actions area can be customized, right. I might want actions without selection checkboxes in some cases.

@mcarrano yep, you can place anything you want in the action area. The only styling we do to anything there is put an 8px spacer between whatever actions are present.

PR #1911, opened 2019-06-07
chore(a11y): use tablet viewport size when auditing

seanforyou23:
This PR updates the window size used for the browser that our a11y audit is run against. I've noticed for a while that we get more errors reported on CI that we do locally, was about to file a bug when I realized using a smaller than "maximized" size causes the audit to produce consistent results across both local and CI. We've been wanting to start testing smaller viewports anyway, as many components change in appearance and behavior at smaller dimensions. So, here we use the size of an iPad in portrait orientation.

Should help close; https://github.com/patternfly/patternfly-next/issues/1602 and https://github.com/patternfly/patternfly-next/issues/1906

### DIFF
--- a/test-a11y/index.js
+++ b/test-a11y/index.js
@@ -17,7 +17,7 @@ let chromeOptions = {};
 if (process.env.CI) {
   chromeOptions = { args: ['--headless'] };
 } else {
-  chromeOptions = { args: ['--start-maximized', '--incognito'] };
+  chromeOptions = { args: ['--incognito', '--window-size=768,1024'] };
 }
 
 const chromeCapabilities = selenium.Capabilities.chrome();
--- a/src/patternfly/components/Toolbar/examples/toolbar-expanded-example.hbs
+++ b/src/patternfly/components/Toolbar/examples/toolbar-expanded-example.hbs
@@ -3,7 +3,7 @@
     {{> dropdown-toggle-check dropdown-toggle-check--id=(concat toolbar--id '-split-button-dropdown-check') aria-label="Select all"}}
     {{> dropdown-toggle-button dropdown--IsToggleButton="true" dropdown-toggle-button--id=(concat toolbar--id '-split-button-dropdown-button') aria-label="Select"}}
   {{/dropdown}}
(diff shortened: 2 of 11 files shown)


test-a11y/index.js
```diff
@@ -17,7 +17,7 @@ let chromeOptions = {};
 if (process.env.CI) {
   chromeOptions = { args: ['--headless'] };
 } else {
-  chromeOptions = { args: ['--start-maximized', '--incognito'] };
+  chromeOptions = { args: ['--incognito', '--window-size=768,1024'] };
```
christiemolloy: Could you explain, does this mean that it looks at the window sizes 768 and 1024?

I'm wondering if its possible to check the screen at all breakpoint sizes that we have, because changes can happen at all of them:

// Grid breakpoints
$pf-global--breakpoint--xs: 0 !default;
$pf-global--breakpoint--sm: 576px !default;
$pf-global--breakpoint--md: 768px !default;
$pf-global--breakpoint--lg: 992px !default;
$pf-global--breakpoint--xl: 1200px !default;
$pf-global--breakpoint--2xl: 1450px !default;

seanforyou23: @christiemolloy that will set the viewport to 768px (w) by 1024px (h) . To extend the audit to check for each of the supported breakpoints is a bit beyond the scope of this PR as we'd then want to make sure we have a way of collecting and publishing the results for each of the breakpoints. Currently, we're just using a single reporting mechanism and I found that cranking down the viewport size in general does wonders for bringing alignment between local dev and CI.

We have these two issues in the backlog which would be a good point in time to address checking all supported breakpoints: https://github.com/patternfly/patternfly-next/issues/635 and https://github.com/patternfly/patternfly-next/issues/649

srambach:
Probably a future idea, but I was wondering if it would be worth making a parameter to the test to say which breakpoint to use.

test-a11y/index.js
christiemolloy: Ok thanks for clarifying that! Sounds good to me :)

Issue #1429, opened 2019-02-15
Create common toggle and context menu components


mcoker:
I would like to propose creating common components for a toggle button and context menu that are used in components like dropdown, app launcher, options menu, select, context selector, etc.

The toggle button and context menus in our current dropdown menus are 90% the same, with relatively small variations in how they present themselves per component. These differences could be built into the shared components themselves and the variations exposed with the use of modifier classes, and instead of dropdown, options-menu, select, etc, we could have a common "dropdown" component that is modified based on whether you're building a select menu, app launcher, etc, or maybe we continue to support each individual type of dropdown component, and those components simply reference the shared toggle and context menu components with defaults that apply to that specific component.

Modular dropdown design recomendations here: https://docs.google.com/document/d/1GyVgog_SbErqbYyF08_N2rfLDUgwAVbdKP5udBjdxYI/edit?usp=sharing

jgiardino:
One idea is that to have 3 types of components:

- Dropdown - basically a wrapper, similar to how card is a wrapper, and just controls positioning of the toggle relative to the menu
- Menu Item
- Toggle

Then maybe the demos are where we show different variations of these components based on the design documentation we have. 

It might also be nice to decouple the a11y stuff from the component level, at least the aspects that are heavily tied to context (like form or not a form), use case (like menu of options vs actions), and/or specific javascript implementations (like the keyboard behavior that's required when using aria-haspopup). These could potentially be documented at the Demo level.

karelhala:
@jgiardino this is really great approach how to do this! This will heavilly impact the implementation of dropdown in react (and eventually in any framework) component. The best way to implement this (interaction in JS) would be to build on top of current implementation and just improve the API with new capabilities.

With this approach we will not be limited at all in using dropdown
* horizontal menu [1]
* Custom datepicker (this would be really good!)
* Notification drawer (accordion with some styling)
* Literally anything conumer might think of

Currently I can think of 4 elements as dropdown item
* `button` - to interact with dropdown (selectable item - both with and without checkboxes and horizontal menu?)
* `a` - to navigate
* `input` - to filter group, not every dropdown implementation supports this (they usually allow to filter trough all items), but it would be really cool to allow consumers to filter trough part of the items and rest to be static
* `div` - container to hold any item consumer wishes to show in dropdown menu

[1] (just a quick mock-up)
[...]

PR #1916, opened 2019-06-11
feat(applauncher): add support for sections/icons/divider

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1914

@kybaker I just put images and SVGs of various shapes I found online in the example to show how they will look. There is a 24x24 area to the left of the app launcher link text where the icon will go and the log edge of the icon will stretch to fill the 24px length, and the shorter edge will scale down. And if the logo is less than 24x24, it will just be smaller. All the images will be vertically and horizontally centered.

@jgiardino I changed the title `<h2>`s to `<h1>`s, inline with what we're doing in the dropdown component, since you can establish a new heading hierarchy in `<section>` elements. And added the scree-reader text.


jgiardino:
👍 This looks great! 

I really like how you combined the icon with the screen reader text, since they are synonymous. Looking at this html example below, this seems like a pattern that should be reusable in other places. A similar case came up with my product team today where they want to include a link to the product documentation in a dropdown menu in the masthead. Would it be possible to make this chunk of the menu item a reusable component? Or is that just something we should capture as part of https://github.com/patternfly/patternfly-next/issues/1429? 

```
<span class="pf-c-app-launcher__menu-item-external-icon">
  <i class="fas fa-external-link-alt" aria-hidden="true"></i>
  <span class="pf-screen-reader">(opens new window)</span>
</span>
```

One change that I recommend is to not set opacity on `.pf-c-app-launcher__menu-item-external-icon` and instead set it on just the icon, to avoid any potential issues with opacity affecting screen reader perception of that text.

mcoker:
@jgiardino thanks, updated so that we don't have a wrapper for the external link icon, it seems unnecessary. Also updated the documentation. Should `aria-hidden="true"` be applied to the icon on the left of the list item, too?

mcoker:
@jgiardino 

> Would it be possible to make this chunk of the menu item a reusable component? 

What would be the benefit of it being a component? I'm also curious how we would position it as a component - what might it be named? I also changed the structure with an update (removed the wrapper and now just have the icon + screen-reader text) and, assuming that works, curious if that impacts this potentially becoming a component?

src/patternfly/components/AppLauncher/examples/app-launcher-sections-icons-example.hbs
```diff
@@ -0,0 +1,266 @@
+{{#> app-launcher id="app-launcher-example-sections-icons" app-launcher--IsExpanded="true" aria-label="Actions" app-launcher--IsNav="true" app-launcher--attribute='aria-label="global"'}}
+  {{#> app-launcher-group}}
+    <ul>
+      <li>
+        {{#> app-launcher-menu-item}}
+          {{#> app-launcher-menu-item-icon}}
+            <img src="https://kubernetes.io/images/favicon.png" alt="">
```
jgiardino: > Should aria-hidden="true" be applied to the icon on the left of the list item, too?

Since you are using `` elements for these icons, AND since these images are redundant (i.e. the text that follows is considered synonymous with these icons), then having `alt=""` like you have it here is the correct way to tell assistive technology that this image provides no extra meaning and to ignore it.

jgiardino:
> What would be the benefit of it being a component?

I think this just gets back to the need to have menu items and toggles as separate components so that we can build these things just once (e.g. this would be a variation of the MenuItem component). Should I just add this as something to #1429 for now?

This is definitely something that could be useful in the Dropdown menu, only because this exact use case came up in a product planning meeting yesterday. :-)

> I also changed the structure with an update (removed the wrapper and now just have the icon + screen-reader text) and, assuming that works, curious if that impacts this potentially becoming a component?

I think this works. I think if it's just a variation of a future menu item component, this would be fine.

mcoker:
> Should I just add this as something to #1429 for now?

@jgiardino that sounds good to me, thanks!

jgiardino:
I have a question for design (@mcarrano, @kybaker)...

Is the use case for this component that it will always have links? 

A question came up off thread about whether the app launcher should always be wrapped in a `<nav>` element. My thinking is that if the consumer needs a similar interaction pattern that provides a menu of action buttons instead of navigational links, then that's what the Dropdown should be used for. Does that match your expectation?

Also, in other components where we use the `<nav>` element, we try to be prescriptive in how we define the labels for those regions. This is really important when screen reader users are navigating by region and there are multiple navigation regions on the page. Some of these labels we decided on are "Global" for primary nav, "Local" for tertiary nav, and "Breadcrumb" for breadcrumbs (e.g. VO announces this region as "Breadcrumb navigation").

If the app launcher is wrapped in a navigation region, what would that label be? Would "Application" match our use case for this component?

@mcoker If we do decide that all app launchers should use `<nav>` and that all aria-labels should be more specific to our use case for this component, should that be captured in a separate issue (since that would require updating all examples and not just this new one)?

mcoker:
> @mcoker If we do decide that all app launchers should use <nav> and that all aria-labels should be more specific to our use case for this component, should that be captured in a separate issue (since that would require updating all examples and not just this new one)?

@jgiardino I'm fine with making that update in this PR if you would be OK with verifying that changeset.

mcarrano:
> My thinking is that if the consumer needs a similar interaction pattern that provides a menu of action buttons instead of navigational links, then that's what the Dropdown should be used for. Does that match your expectation?

Yes, this matches my expectations @jgiardino . Regarding the naming, I think "Application" could be ambiguous. Maybe "Application lancher" or just "launcher?"

jgiardino:
> Regarding the naming, I think "Application" could be ambiguous. Maybe "Application lancher" or just "launcher?"

I was also wondering the same thing, and thinking that "Application launcher" might be better. 

"Launcher" alone seems ambiguous to me, too.

And something to keep in mind is that this label gets combined with "navigation", so for the user it would get announced as "Application launcher navigation".

mcoker:
@jgiardino updated the examples and documentation.

mcoker:
@mattnolting thanks, copied and pasted that code from the dropdown component, so committed your suggestions and updated the dropdown component if you want to take another look.

PR #1918, opened 2019-06-11
fix(nav): change breakpoint to fix bug

### LINKED ISSUE #1917: fix(nav): button has transparent background at lg breakpoint
When working on the website I noticed that at 992px (lg breakpoint) the scroll button background turns transparent. Its happening because the scroll button background relies on the background of the nav, and at a lg breakpoint the nav background color turns from dark grey to transparent and so just and so based on the order that the variables are assigned, the background color changes, before the new nav background color is assigned.

Screen Shot 2019-06-11 at 11 05 00 AM

christiemolloy:
fix #1917 

When I remove 1px from the breakpoint it fixes the issue of `top: var(--pf-c-page__header-nav--c-nav__scroll-button--lg--Top);` and `background-color: var(--pf-c-page__header-nav--c-nav__scroll-button--lg--BackgroundColor);` being assigned too early and causing a bug.

This also fixes the nav background-color changing to transparent before it is used for the background color of the scroll buttons.

I think its happening because the grid in the header nav changes at a minimum breakpoint of lg, and these assignments are happening at a maximum breakpoint of lg, although I can't fully pinpoint the issue. @mcoker wondering if you have any insight into this?

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -240,7 +240,7 @@
     min-width: 0;
 
     .pf-c-nav__scroll-button {
-      @media screen and (max-width: 991px) {
+      @media screen and (max-width: #{$pf-global--breakpoint--lg - 1}) {
         top: var(--pf-c-page__header-nav--c-nav__scroll-button--lg--Top);
         background-color: var(--pf-c-page__header-nav--c-nav__scroll-button--lg--BackgroundColor);
       }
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -240,7 +240,7 @@
     min-width: 0;
 
     .pf-c-nav__scroll-button {
-      @media screen and (max-width: $pf-global--breakpoint--lg) {
+      @media screen and (max-width: 991px) {
```
mattnolting: Can you use a subtraction operator here `#{$pf-global--breakpoint--lg - 1}`?

Issue #1833, opened 2019-05-16
Introduce masthead utility menus to Page Layout demo


mcarrano:
Define standard masthead utility menus - organization and content and incorporate into the standard page layout demo. Need to consider what menus are required and how content is organized. Consider requirements for user management, help, support, language selection, and other settings.

Results should be incorporated into the existing page layout demo.

Related issues: #1832 , https://github.com/patternfly/patternfly-org/issues/714

mcarrano:
@dongniwang per our discussion, I have created this issue and assigned in to you for discovery work. I am working with @maryshak1996 on a separate issue to better define the behavior of utility items in general as a first priority. Once we have standard menus defined, the goal would be to implement as a demo and publish design guidelines to the website. Let me know if you have questions.

jeperry:
FYI, if you log in with your Red Hat Account you can see the early attempt at cloud.redhat.com. Screen shots here for convenience.  

![image](https://user-images.githubusercontent.com/2206023/57889588-315e2800-7803-11e9-996a-a3c00218aeca.png)
 
![image](https://user-images.githubusercontent.com/2206023/57889600-3de28080-7803-11e9-827a-1c74990f162c.png)

dongniwang:
@mcarrano thank you! 

@jeperry thank you for the screenshots! Two questions: 
1. The "username" and "account number" in the first screenshot are greyed out, is there a reason for it? For the user avatar, is it clickable? How would users update their avatars? 
2. The second screenshot (help dropdown menu) - is there a reason behind opening the links in the existing browser tab instead of a new tab?

jeperry:
1. username and account number are included as information only. These are necessary for account differentiation, but they are static info and cannot be clicked, so that is why the text treatment is different than actionable menu items below the divider. The clickable area to invoke this menu is the user name and the arrow,  the avatar is not included in this. Avatar image is updated via Customer Portal profile (on another domain) which is reached via the link "My Profile." 
2. The first three links are on different domains and should probably open in another tab, but have not been implemented this way. Is there a design pattern to indicate links in these menus that open new tabs vs those that do not? 'About' simply opens the about modal.

jeperry:
fyi @KendraMar

KendraMar:
@dongniwang Hi, the Cloud Services team is asking about the avatar as well. Here's what one of the team reported in Jira:

"currently the user avatar icon sits without a way to interact with it directly beside the username/action drop-down

[...]

PR #1925, opened 2019-06-13
fix(page): update utility item responsiveness in masthead

### LINKED ISSUE #1832: Define responsive behavior of masthead utility items
Current implementations of the masthead utility items require rework to display desired responsive behaviors. These changes are required to the Page Layout demos in both [core](https://www.patternfly.org/v4/documentation/core/demos/page-full?component=Page%20component%20default%20nav%20header%20example) and [React](https://www.patternfly.org/v4/documentation/react/demos/pagelayout/pagelayoutsimplenav/). 

New implementation must support the following requirements:

(1) Must support the following Utility bar items
* User Avatar (optional per application support), 
* User menu - dropdown with contents TBD)
* Notification icon - alerts the user of unread notifications and toggles notification drawer. The notification icon must be visible at all viewport sizes.
* utility menus - one or more items represented by an icon. Menus can be either Dropdowns or Options Menus. Typically includes Help and Settings but other menus may be included.

(2) At first breakpoint (tablet), collapse  utility items to kabob.
(3) At second breakpoint, (mobile) collapse user menu into kabob and hide avatar. (Only notification icon and kabob will be shown.)

This issue should address how the combined utility menu is displayed when collapsed, i.e. if User, Settings, and Help live in the same menu, what does that look like - how do we maintain hierarchy.

christiemolloy:
Closes ~#1832~ https://github.com/patternfly/patternfly-next/issues/2482

This is a breaking change we won't be merging it until after the next breaking change release.


christiemolloy:
Understood @mcarrano if you can confirm that, i can update.

I think originally we didn't add dropdowns to these demos because it can be alot of maintenance, @mcoker what do you think about updating these buttons to dropdowns and then showing them in expanded state?

mcoker:
> This looks good, but I believe there is an intermediate state where the user menu collapses into the kabob but the avatar is still visible. Then at mobile the avatar hides. @maryshak1996 can you confirm this?

I see this in the designs here

Screen Shot 2019-06-13 at 2 16 53 PM


> what do you think about updating these buttons to dropdowns and then showing them in expanded state?

It is additional maintenance but I wonder if we can abstract the demo in a way where we share the overall page demo code between collapsed and expanded state, then just modify the kebab menu code in the expanded demo to show the expanded state.

Also are we adding a `:focus` state where the dropdowns/buttons in the masthead will have a circular background color?

And from the designs, it looks like on desktop we show all of the masthead items, then on tablet, we collapse everything but the avatar into the kebab, then on mobile, we hide the avatar. Is that right? In this demo, I see we're 1) hiding the gear icon at one breakpoint and moving it into a kebab menu, which I don't think even needs to happen. Maybe if we had a gear icon and something else, we could collapse into a kebab, but I don't see the point in moving one menu into a kebab like that. Then 2) at another breakpoint  we're moving the user menu into the kebab. Seems like from the designs, in this demo when we're collapsing the gear icon into the kebab, we would also collapse the user menu into the kebab, too. That state would persist until mobile where we then hide the avatar.

mcarrano:
Yes, @mcoker your description of how this should work is accurate. Interesting question about how to reflect the menus. There is a separate issue open here to better define these menus and what goes in them: https://github.com/patternfly/patternfly-next/issues/1833 Maybe we should wait for that to be better defined before adding the actual menus? But the thing we were trying to illustrate in this issue is how the utility items collapse into an accordion menu when multiple items are combined into the kabob. I don't think we have anything to date that behaves like that. Maybe this is a pattern that we should codify as a new component? Interested in your thoughts @christiemolloy @mcoker

mcoker:
> Maybe this is a pattern that we should codify as a new component? 

I'm not sure. I was able to mostly plug and play the existing accordion into the dropdown component and get something like this

Screen Shot 2019-06-13 at 4 30 40 PM

It would be great if we could build this in a modular way where this would just be a different variation of the dropdown showing an accordion menu in the dropdown menu instead of a traditional list of links/buttons, depending on how we choose to modify the styles needed to place an accordion in a dropdown. It may be better off as a new component if we want to have more control over how the accordion is displayed in a dropdown.

mcarrano:
I could go either way @mcoker Visually is this what you expected to see @maryshak1996 ?

christiemolloy:
- Updated so that the gear icon and user menu collapse into the kabob at the same time. 
- I agree with @mcarrano in that we can address the dropdown/menu in this issue: #1833 
- For the background behind the notification icon when its in focus/active, I was thinking that could be a modification to the button component when an icon is in it. @mcoker what do you think

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
```diff
@@ -1,15 +1,17 @@
 {{#> page-header-tools}}
   {{#> page-header-tools-group page-header-tools-group--modifier="pf-m-icons"}}
-    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Alerts"'}}
+    {{#> button button--modifier="pf-m-plain pf-c-page__header-notifications" button--attribute='aria-label="Unread notifications"'}}
```
christiemolloy: @mattnolting thanks for that! I merged in your suggestions, do you think its ok that we add these classes: `pf-c-page__header-notifications` and `pf-c-page__header-settings` to the button modifier and `pf-c-page__header-overflow-menu` to the dropdown modifier. I was tossing up between doing this also because i'm not sure if it follows what we typically do. @mcoker what do you think?

mattnolting: Yeah, probably should wrap those. Looks a lot cleaner in the code too.

mcoker: by wrap, do you mean create elements in the page header where we place the buttons? If so, that's what I would do, too.

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
```diff
@@ -1,15 +1,17 @@
...
+      {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
         <i class="fas fa-bell" aria-hidden="true"></i>
+      {{/notification-badge}}
     {{/button}}
-    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Settings"'}}
+    {{#> button button--modifier="pf-m-plain pf-c-page__header-settings" button--attribute='aria-label="Settings"'}}
```
mcoker: Working with different apps, I think the settings button here is just to serve as an example of an icon that someone might put in their header. We're being very prescriptive with the things in the masthead currently, and I think we should be more generic, and only specific with things we know we want to control. So far I think we have 3 things we want to control:

* Notification badge - always shown
* User menu - shown on desktop
* Overflow menu - shown on mobile/tablet

Otherwise, I think we just need a class that allows for a button, dropdown, or whatever in it, and hides it on mobile/tablet. That may be a settings button/menu, help button, both, something else, etc...

I wonder if we should have like `pf-c-page__header-item` with modifiers for `pf-m-desktop` (shows on desktop, hides on mobile) and `pf-m-mobile` (shows on mobile, hides on desktop), so we would use no modifier for notification icon, the `pf-m-mobile` on the overflow menu, and `pf-m-desktop` on the settings/whatever icon and user menu?

We could also have specific classes like you have now for notification icon and user menu, then create a generic class like `pf-c-page__header-item` that you place things in that will show on desktop and hide on mobile, then if you want your menu to persist in both desktop/mobile, you just don't wrap the button in that `__header-item` element. 

I think the former might be easier to document and more intuitive. 

We could also extend the hidden/visible SCSS here and use that, though it's less prescriptive and would allow people to change the breakpoints at which things are hidden/shown more easily (which is good and bad).

mcoker:
> For the background behind the notification icon when its in focus/active, I was thinking that could be a modification to the button component when an icon is in it.

@christiemolloy do you mean updating the button component itself to use that background with the `.pf-m-plain` modifier?

christiemolloy:
@mcoker @mattnolting updated. Should we keep the modifier for avatar or should we have a modifier for pf-m-tablet instead, since it displays on tablet size and up. We could just target the margin on the avatar by wrapping it in a page-header-tools-group and then saying the last-child has a different margin. Since the avatar will always be there, I think its fine to use the avatar class ?

srambach:
What do you think about using `.pf-m-hidden{-on-[breakpoint]}` or `.pf-m-visible{-on-[breakpoint]}` modifiers instead of `.pf-m-desktop` and `.pf-m-mobile`? (like on table/data table) Desktop and mobile are somewhat subjective terms, and giving breakpoint options would also make it more flexible in case there is something else taking space in the header besides the standard components.

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +279,51 @@
...
     @media screen and (min-width: $pf-global--breakpoint--lg) {
       display: none;
       visibility: hidden;
     }
   }
 
-  .pf-m-user {
+  .pf-c-avatar {
     display: none;
+    margin-left: 0;
```
mcoker: is this necessary?

mcoker:
@srambach yeah, I mentioned that at the bottom of this comment https://github.com/patternfly/patternfly-next/pull/1925#discussion_r294008165. I think it's a good idea under the principle of being flexible and encouraging proper use of the component. 

What elements would you suggest the hidden/visible modifiers apply to? Maybe `.pf-c-page__header-tools-group > *`? Or that and `.pf-c-page__header-tools-group`?

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +279,51 @@
...
   }
 }
 
 .pf-c-page__header-tools-group {
   display: flex;
-
-  & + & {
-    margin-left: var(--pf-c-page__header-tools-group--MarginLeft);
-  }
+  margin-left: var(--pf-c-page__header-tools-group--MarginLeft);
```
mcoker: With this change, this is no longer true

> there is only a margin-left on the .pf-c-page__header-tools-group when its next to .pf-c-page__header-tools-group .pf-m-icons

Is that expected? Personally I think it's good. I might even increase the margin-left on mobile, since the logo can span as wide as the available space in the header now, meaning you can end up with only 8px space between the utility icons and the header logo if the logo is wide enough

Screen Shot 2019-07-02 at 1 38 37 PM

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +279,51 @@
...
 
-  .pf-m-user {
+  .pf-c-avatar {
     display: none;
+    margin-left: 0;
     visibility: hidden;
 
-    @media screen and (min-width: $pf-global--breakpoint--md) {
+    @media screen and (min-width: $pf-global--breakpoint--lg) {
+      margin-left: var(--pf-c-page__header-tools--c-avatar--MarginLeft);
```
mcoker: couldn't you just leave this on `.pf-c-avatar` since it won't apply when it's `display: none;`?

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
```diff
@@ -1,22 +1,22 @@
...
     {{/button}}
+    {{#> dropdown id=(concat page--id '-dropdown-kebab-right-aligned-1') dropdown--IsActionMenu="true" dropdown--modifier="pf-m-mobile" dropdown-menu--modifier="pf-m-align-right" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+    {{/dropdown}}
   {{/page-header-tools-group}}
   {{#> page-header-tools-group}}
-    {{#> dropdown id=(concat page--id '-dropdown-kebab-right-aligned-1') dropdown--IsActionMenu="true" dropdown-menu--modifier="pf-m-align-right" dropdown-toggle--modifier="pf-m-mobile pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
-    {{/dropdown}}
-    <div class="pf-m-user">
-      {{#> dropdown id=(concat page--id '-dropdown') dropdown-toggle--modifier="pf-m-plain" dropdown--HasToggleIcon="true"}}
+      {{#> dropdown id=(concat page--id '-dropdown') dropdown-toggle--modifier="pf-m-plain pf-m-desktop" dropdown--HasToggleIcon="true"}}
```
mcoker: you would want to apply the desktop class to `dropdown--modifier`, not `dropdown-toggle--modifier`

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +279,51 @@
   margin-bottom: var(--pf-c-page__header-tools--MarginBottom);
   margin-left: auto; // to push it to the right
 
-  .pf-m-icons {
```
mcoker: Removing this will make it a breaking change.

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +279,51 @@
...
+  // Hides on mobile visible on desktop
+  .pf-m-desktop {
     display: none;
     visibility: hidden;
 
     @media screen and (min-width: $pf-global--breakpoint--lg) {
-      display: block;
+      display: inline-block;
```
mcoker: I think this should have been `display: flex;` since it applies to `.pf-c-page__header-tools-group`, which is `display: flex;`. But if you add `.pf-m-icons` back, just make sure whatever you use here works with that class and what you're using to show the desktop items.

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +279,51 @@
...
+    display: inline-block;
+    visibility: visible;
 
     @media screen and (min-width: $pf-global--breakpoint--lg) {
       display: none;
       visibility: hidden;
     }
   }
 
-  .pf-m-user {
```
mcoker: removing this is also a breaking change.

christiemolloy: is removing pf-m-user a breaking change now that the display is handled by its parent?

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
```diff
@@ -1,22 +1,22 @@
...
+        {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
+          <i class="fas fa-bell" aria-hidden="true"></i>
+        {{/notification-badge}}
     {{/button}}
-    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Settings"'}}
+    {{#> button button--modifier="pf-m-plain pf-m-hidden pf-m-visible-on-lg" button--attribute='aria-label="Settings"'}}
         <i class="fas fa-cog" aria-hidden="true"></i>
     {{/button}}
+    {{#> dropdown dropdown--modifier="pf-m-hidden-on-lg" id=(concat page--id '-dropdown-kebab-right-aligned-1') dropdown--IsActionMenu="true" dropdown-menu--modifier="pf-m-align-right" dropdown-toggle--modifier="pf-m-mobile pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+    {{/dropdown}}
```
christiemolloy: @mcoker @mattnolting what are your thoughts on moving the dropdown kebab to the icons group?

src/patternfly/components/Page/page.scss
```diff
@@ -273,35 +284,18 @@
...
-    }
-  }
-
-  .pf-m-mobile {
-    margin-right: 0;
-
-    @media screen and (min-width: $pf-global--breakpoint--lg) {
-      display: none;
-      visibility: hidden;
+    .pf-c-button {
```
christiemolloy: @srambach @mcoker added the `@includes pf-hidden-visible`  here, what do you think?

src/patternfly/components/Page/page.scss
```diff
@@ -63,12 +63,24 @@
   --pf-c-page__header-tools--MarginBottom: var(--pf-global--spacer--sm);
   --pf-c-page__header-tools--md--MarginRight: var(--pf-global--spacer--lg);
   --pf-c-page__header-tools--c-avatar--MarginLeft: var(--pf-global--spacer--md);
-  --pf-c-page__header-tools-group--MarginLeft: var(--pf-global--spacer--xl);
+  --pf-c-page__header-tools-group--Display: flex;
+  --pf-c-page__header-tools-group--m-icons--c-button--Display: block;
+  --pf-c-page__header-tools-group--m-icons--c-dropdown--Display: block;
+  --pf-c-page__header-tools-group--c-avatar--Display: inline-block;
```
mcoker: What do you think about adding a `.pf-c-page__header-tools-item` that can be a child of `.pf-c-page__header-tools-group`, then we apply the hidden-visible extend to the `tools-item` and `tools-group` classes? Then a user can put whatever they want in that and we won't need to add hidden-visible support for each thing we/users add to the header tools (buttons and dropdowns currently).

So each button/dropdown in a group would be wrapped in an item, and the avatar would be wrapped in an item. That's also a more modular approach.

src/patternfly/components/Page/page.scss
mcoker: Yep, think about people who aren't using react and pull in these changes without updating their markup. A good way to test this is to copy the stylesheet from this PR into a local build of master and test all of the page component pages in the dev workspace. Everything should work as it did before this update.

src/patternfly/components/Page/page.scss
christiemolloy: Ohh I like that idea. I updated the PR

src/patternfly/components/Page/page-header-tools-group-item.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-page__header-tools-group-item{{#if page-header-tools-group-item--modifier}} {{page-header-tools-group-item--modifier}}{{/if}}"
```
mattnolting: @mcoker were you saying `pf-c-page__header-tools-item` rather than `pf-c-page__header-tools-group-item`?
```suggestion
<div class="pf-c-page__header-tools-item{{#if page-header-tools-item--modifier}} {{page-header-tools-item--modifier}}{{/if}}"
```

src/patternfly/components/Page/page.scss
```diff
@@ -272,50 +280,36 @@
   margin-bottom: var(--pf-c-page__header-tools--MarginBottom);
```
mattnolting: Removing `margin-top` and `margin-bottom` here doesn't seem to affect layout. Is it needed? I think originally, it was necessary for the avatar, but I can't remember.

src/patternfly/components/Page/page-header-tools-group-item.hbs
mcoker: I think so, just because the class is getting really long, `group` in the name doesn't seem necessary, but I'm fine with it either way.

mattnolting: I'm just thinking back to the old discussion, is a `group` wrapper `necessary` when there is nothing to group/items exists on its own..

mcoker: So adjacent groups and items would have the same spacing as groups currently, and items in groups would have no spacing (like how buttons, dropdowns, etc in groups now have no spacing)?

mcoker: Looks like that's how @christiemolloy set it up, so that groups with just 1 item are just an item, which is how I would expect it. Though I think we should apply the same `margin-left` that is on groups to items that are adjacent to a group, then disable that margin on the item with `.pf-m-avatar`, since we'll need to leave the `margin-left` on `.pf-c-avatar` so as not to make this a breaking change. Then in a breaking change release we can remove the margin on `.pf-c-avatar` and apply that to the item with `.pf-m-avatar`. What do you think?

src/patternfly/demos/Page/page-template-header-tools-elements.hbs
mcoker: I think that works since we want the kebab grouped menu to display adjacent to the notification bell.

src/patternfly/components/Page/page-header-tools-group-item.hbs
christiemolloy: Ah yes I meant to do `pf-c-page__header-tools-item` , which works better because the item doesn't necessarily need to be part of the group.

src/patternfly/components/Page/page.scss
christiemolloy: I think you're right, created this issue to remove it during the breaking change period: #2040

src/patternfly/components/Page/page-header-tools-group-item.hbs
mattnolting: @mcoker I think that icon button groups, because there is no border on the button and padding makes them appear to be nicely spaced, are an exception to what might appear in the page header tools group. So, target first child and apply large spacer value, then inside groups apply spacer value (defined by design) to children. The exception would be group of things that should not have margins, like `-m-icons`. 

Screen Shot 2019-07-11 at 11 50 35 AM
Screen Shot 2019-07-11 at 12 06 14 PM

What do you think?

mcoker: I suppose it depends on what all goes in the masthead. If we're adding buttons, I wonder if that would be another case where the action list layout would be used.

christiemolloy:
@mcarrano and @maryclarke , @mcoker mentioned some good points that I'm wondering if you can help answer:
1. Can they add another group/item outside of the existing `.pf-m-icons` group? Or if they add things, should it be in that list of icons?
2. If so, can they add things other than a plain button?
3. Should there be “group” spacing between all groups/non-grouped-items in the masthead?
4. Should there be space between items in a group? Or do we just remove space in a class like `.pf-m-icons` (or give it a less specific name that conveys no space between items)

mcarrano:
@christiemolloy not sure that I can answer all of your questions, but it sounds like we may be over-complicating this a bit. I can't really see a need for making the masthead into a general purpose toolbar. I see the need for an icon group with a flexible number of icons along with the user menu and avatar. I doubt that we would want other types of controls like text buttons populating the masthead.

Does that help?

PR #1928, opened 2019-06-14
fix(tables): fixed word wrap (iss1791)

### LINKED ISSUE #1791: Word Wrap in Table <td>
We should add functionality to wrap words in td's in our Table, this issue arose from using the table in the website:

https://github.com/patternfly/patternfly-org/issues/881

lareplog:
Added word-wrap to td's in table, now wraps when on mobile size.  

Close #1791

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -144,7 +144,7 @@
   background-color: var(--pf-c-table--BackgroundColor);
 
   td {
-    word-break: break-all;
+    word-break: break-word;
   }
 
   // Standard table row (non-expandable)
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -143,6 +143,10 @@
   width: 100%;
   background-color: var(--pf-c-table--BackgroundColor);
 
+  td {
+    word-break: break-all;
```
christiemolloy: Actually Lauren will you change this to "word-break: break-word"

I tested this in the table, and some words were breaking when they shouldn't. Using "break-all" breaks every word, but "break-word" breaks just the really long words.

This is with **break-all** : 
Screen Shot 2019-06-14 at 10 44 50 AM


This is with "break-word":
Screen Shot 2019-06-14 at 10 44 43 AM

lareplog: Changed it!

PR #1932, opened 2019-06-14
fix(toggle): Expanded/collapsed arrow rotation in components

### LINKED ISSUE #1920: Expanded/collapsed arrow rotation in components
The datalist is the only component that uses a var for its arrow's `transform` property for the expanded state. Let's be consistent and add a var to these other components, too.

* accordion
* clipboard copy
* expandable
* expandable table
* wizard mobile nav toggle

christiemolloy:
close #1920

Test: added in `:focus` to test the arrow rotating will remove prior to merging
- accordion
- clipboard copy
- data list
- expandable
- table
- wizard toggle


src/patternfly/components/Accordion/accordion.scss
```diff
@@ -31,6 +31,8 @@
 
   // accordion toggle icon
   --pf-c-accordion__toggle-icon--LineHeight: var(--pf-global--LineHeight--md);
+  --pf-c-accordion__toggle-icon--expanded--Transition: .2s ease-in 0s;
```
christiemolloy: Does anyone recall why these use `expanded` over `m--expanded` `--pf-c-accordion__toggle-text--expanded--Color` `--pf-c-accordion__toggle-text--expanded--FontWeight`

mcoker: Looks like an oversight to me.

christiemolloy: will create a follow up issue thanks @mcoker

christiemolloy: will create a follow up issue thanks @mcoker

christiemolloy: will create a follow up issue thanks @mcoker

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -80,7 +83,8 @@
       }
 
       .pf-c-accordion__toggle-icon {
-        transform: rotate(90deg);
+        transition: var(--pf-c-accordion__toggle-icon--expanded--Transition);
```
mcoker:
```suggestion
        transition: var(--pf-c-accordion__toggle-icon--m-expanded--Transition);
```

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -80,7 +83,8 @@
...
+        transform: var(--pf-c-accordion__toggle-icon--expanded--Transform);
```
mcoker:
```suggestion
        transform: var(--pf-c-accordion__toggle-icon--m-expanded--Transform);
```

src/patternfly/components/Accordion/accordion.scss
mcoker: You also want to move the `transition` to `.pf-c-accordion__toggle-icon` outside of the context of `.pf-m-expanded`, otherwise the transition will only happen when the element already has `.pf-m-expanded` (it will transition "collapsed", but won't transition "expanded")

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
```diff
@@ -16,6 +16,10 @@
   --pf-c-clipboard-copy__group-toggle--m-expanded--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-clipboard-copy__group-toggle--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
 
+  // Icon
+  --pf-c-clipboard-copy__group-toggle-icon--m-expanded--Transition: .2s ease-in 0s;
```
mcoker: `.2s ease-in 0s;` looks like it should be a global var.

src/patternfly/components/Accordion/accordion.scss
mcoker: also just left a comment in https://github.com/patternfly/patternfly-next/issues/1938, but let's change this var and the one below to `--pf-c-accordion__toggle--m-expanded__toggle-icon--[Transition/Transform]`

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
```diff
@@ -41,15 +45,18 @@
...
+  &.pf-m-expanded {
+    .pf-c-clipboard-copy__group-toggle {
+      &::before {
+        --pf-c-clipboard-copy__group-toggle--BorderBottomColor: var(--pf-c-clipboard-copy__group-toggle--m-expanded--BorderBottomColor);
+
+        border-bottom-width: var(--pf-c-clipboard-copy__group-toggle--m-expanded--BorderBottomWidth);
+      }
+
+      &:focus .pf-c-clipboard-copy__group-toggle-icon {
+        transition: var(--pf-c-clipboard-copy__group-toggle-icon--m-expanded--Transition);
```
mcoker: you'll want to move the `transition` here, too, so it applies to `.pf-c-clipboard-copy__group-toggle-icon` on it's own, not just in the context of `pf-m-expanded`. Also can we update the var name so it matches the selector here, too? And create a follow up issue with a breaking change label to address the existing vars we have where the var name doesn't match the selectors?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -213,13 +213,12 @@
...
   // rotate icon
-  .pf-c-button > * {
+  .pf-c-button:focus > * {
     pointer-events: none;
-    transition: var(--pf-c-data-list__toggle--c-button--Transition);
-    transform: var(--pf-c-data-list__toggle--c-button--Transform);
 
     .pf-c-data-list__item.pf-m-expanded & {
-      --pf-c-data-list__toggle--c-button--Transform: var(--pf-c-data-list__item--m-expanded__toggle--c-button--Transform);
+      transition: var(--pf-c-data-list__toggle-icon--m-expanded--Transition);
```
mcoker: Same here, you'll want to move the `transition` to `.pf-c-button > *`, then just use `transform` for the expanded state. And update the var name so it matches the selectors. Looks like it was in the correct format before.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -18,20 +20,22 @@
...
     font-weight: var(--pf-c-expandable__toggle--FontWeight);
     color: var(--pf-c-expandable__toggle--Color);
     border: none;
 
+    &:focus,
+    &.pf-m-expanded {
+      --pf-c-expandable__toggle--Color: var(--pf-c-expandable__toggle--m-expanded--Color);
+
+      .pf-c-expandable__toggle-icon {
+        transition: var(--pf-c-expandable__toggle-icon--m-expanded--Transition);
```
mcoker: same comment as above applies here

src/patternfly/components/Table/table.scss
```diff
@@ -282,9 +284,10 @@
       transform: rotate(270deg);
     }
 
-    &.pf-m-expanded {
+    &.pf-m-expanded:focus {
       > * {
-        transform: rotate(0deg);
+        transition: var(--pf-c-table__toggle--c-button--m-expanded--Transition);
```
mcoker: same comments as above

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -298,6 +303,7 @@
 
 .pf-c-wizard__toggle-icon {
   margin-top: var(--pf-c-wizard__toggle-icon--MarginTop);
+  transform: var(--pf-c-wizard__toggle-icon--Transform);
```
mcoker: This shouldn't be necessary, it should point down when collapsed, and up when expanded (`rotate(180deg)`)

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -256,9 +259,11 @@
     visibility: hidden;
   }
 
-  &.pf-m-expanded {
+  &.pf-m-expanded,
+  &:focus {
     .pf-c-wizard__toggle-icon {
-      transform: rotate(180deg);
+      transition: var(--pf-c-wizard__toggle-icon--m-expanded--Transition);
```
mcoker: Same comments here as above.

src/patternfly/components/Expandable/expandable.scss
```diff
@@ -18,20 +20,22 @@
...
-  }
-
   .pf-c-expandable__toggle {
     padding: var(--pf-c-expandable__toggle--PaddingTop) var(--pf-c-expandable__toggle--PaddingRight) var(--pf-c-expandable__toggle--PaddingBottom) var(--pf-c-expandable__toggle--PaddingLeft);
     font-weight: var(--pf-c-expandable__toggle--FontWeight);
     color: var(--pf-c-expandable__toggle--Color);
     border: none;
 
+    &:focus,
+    &.pf-m-expanded {
```
mcoker: the expanded class is applied to the component, not the `.pf-c-expandable__toggle` element

mcoker:
I forgot to include the nav component's expandable variation in this list. Can you add these changes there, too?

src/patternfly/components/Accordion/accordion.scss
christiemolloy: @mcoker why did you add in the extra toggle into the name?

mcoker: It's just following the selector names, and `toggle-icon` is the BEM element name for the toggle icon. Since the selector order is `.pf-c-accordion__toggle.pf-m-expanded` -> `.pf-c-accordion__toggle-icon`, the var would be `--pf-c-accordion__toggle--m-expanded__toggle-icon`

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,6 +1,7 @@
 .pf-c-nav {
   --pf-c-nav--Width: #{pf-size-prem(290px)};
   --pf-c-nav--Transition: var(--pf-global--Transition);
+  --pf-c-nav__item--m-expanded__toggle--child--Transform: rotate(90deg);
```
christiemolloy: @mcoker Since there is no icon class here should I use `child` in the selector name, or should we create an icon class to be consistent.

mcoker: An icon class would be great. The data-list would need that update, too. Let's be sure and leave the `*` as the selector so the icon still rotates for people who don't have the new markup, then create a follow up issue to target the icon class in the selector in a breaking change release.

src/patternfly/components/DataList/docs/data-list-expandable.md
```diff
@@ -20,7 +20,8 @@
 | -- | -- | -- |
 | `.pf-c-data-list__item-control` | `<div>` | Initiates a container for data list controls. For example, add `.pf-c-data-list__toggle` here. **Required** |
 | `.pf-c-data-list__toggle` | `<div>` | Initiates a toggle button. |
-| `.pf-c-data-list__icon`  | `<i>` | Initiates a toggle icon. |
+| `pf-c-data-list__toggle-icon` | `<i>` | Initiates a toggle icon. |
```
mcoker:
```suggestion
| `.pf-c-data-list__toggle-icon` | `<i>` | Initiates a toggle icon. |
```

src/patternfly/components/Table/table.scss
```diff
@@ -61,6 +61,9 @@
 
   // Table cell - toggle
   --pf-c-table__toggle--c-button--MarginTop: calc(#{pf-size-prem(6px)} * -1);
+  --pf-c-table__toggle--c-button-icon--Transform: rotate(270deg);
```
mcoker: should this be...

```suggestion
  --pf-c-table__toggle--c-button__toggle-icon--Transform: rotate(270deg);
```

src/patternfly/components/Table/table.scss
```diff
@@ -61,6 +61,9 @@
...
+  --pf-c-table__toggle--c-button-icon--Transition: var(--pf-global--toggle--Transition);
```
mcoker:
```suggestion
  --pf-c-table__toggle--c-button__toggle-icon--Transition: var(--pf-global--toggle--Transition);
```

src/patternfly/components/Table/table.scss
```diff
@@ -61,6 +61,9 @@
...
+  --pf-c-table__toggle--c-button--m-expanded__icon--Transform: rotate(360deg);
```
mcoker:
```suggestion
  --pf-c-table__toggle--c-button--m-expanded__toggle-icon--Transform: rotate(360deg);
```

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -218,6 +218,7 @@ $pf-global--ListStyle: disc outside !default;
 
 // Transitions
 $pf-global--Transition: all 250ms cubic-bezier(.42, 0, .58, 1);
+$pf-global--toggle--Transition: .2s ease-in 0s;
```
mcoker: `transition-delay`'s initial value is `0s` so you shouldn't need that.

I'm also not sure about the var name. Changing a global var name is a fairly significant breaking change. Do we want to create transitions for behaviors, like spinning a toggle icon that are different from something like opening/closing a drawer or hiding the sidebar? FWIW, `.2s` is pretty similar to `250ms`, which is what we're already using for `$pf-global--TransitionDuration`. I also wonder if we should separate this out like the other var where we have maybe....

```
$pf-global--TransitionTimingFunction--100: ease-in;
$pf-global--TransitionTimingFunction--200: cubic-bezier(.645, .045, .355, 1);

$pf-global--TransitionDuration--100: .2s;
$pf-global--TransitionDuration--200: .25s;

$pf-global--Transition--toggle: $pf-global--TransitionDuration--100 $pf-global--TransitionTimingFunction--100;
$pf-global--Transition--menu: $pf-global--TransitionDuration--200 $pf-global--TransitionTimingFunction--200;
```

Although I think the vars we would make here depend on the different types of animations we want to offer, and I don't think we've explored that yet.

Maybe at a minimum I'd rename the var to `$pf-global--Transition--toggle;` as I imagine we'll have different variations of a transition, so the variation is a modifier of the thing it applies to. Or maybe we should wait and define the vars when we have a better transition/animation strategy. WDTY?

christiemolloy: @mcoker I really like the idea you have for separating out the vars etc. but I do think that we shouldn't make a global variable now when we don't know the full strategy behind the animation. I don't think theres any harm in keeping the values hard-coded and then creating a variable in the future?

mcoker: Agreed, sounds good!

mcoker: Agreed, sounds good!

src/patternfly/components/Select/select.scss
```diff
@@ -60,6 +60,7 @@
   --pf-c-select__toggle-arrow--MarginLeft: var(--pf-global--spacer--md);
   --pf-c-select__toggle-arrow--MarginRight: var(--pf-global--spacer--sm);
   --pf-c-select__toggle-arrow--with-clear--MarginLeft: var(--pf-global--spacer--sm);
+  --pf-c-select__toggle-arrow--m-top--m-expanded__toggle-arrow--Transform: rotate(-180deg);
```
christiemolloy: @mattnolting should we make this a positive number for consistency, or does it matter?

mattnolting: That sounds right, good catch!

srambach:
Select and options menu are not using a transition, while the wizard toggle is also a 180 transform but does use the transition. Just checking whether those should be consistent or not.

christiemolloy:
@srambach I spoke to @mceledonia and he said that components that use the caret icon: i.e. wizard, select, dropdown, options menu, context selector.... don't need a transition. Updated the PR accordingly.

srambach:
OK, to clarify, so up/down, no transition, but if it's going 90 degrees, use the transition?

christiemolloy:
@mceledonia  correct me if I'm wrong but it depends on the type of icon. For the smaller caret icon, no transition, but for the arrow, transition: 

Screen Shot 2019-06-27 at 4 15 02 PM

Screen Shot 2019-06-27 at 4 15 13 PM

srambach:
Ah.... I totally did not notice that difference! bitmoji

christiemolloy:
@srambach  bitmoji

PR #1936, opened 2019-06-17
Reduce size of BG image

psav:
As this background should be covered up by content most of the time, a small reduction in quality yields a big reduction in filesize.


mcoker:
@psav this looks great, thanks! Would you mind also copying this file to `src/patternfly/assets/images` and overwrite the existing `pfbg_2000.jpg` file there, too?

psav:
Sure thing

psav:
I can also do another PR to do the same type of transformation to the smaller size images

mcoker:
@psav that would be great, thanks!

PR #1982, opened 2019-06-25
fix(expandable): add type to the button

### LINKED ISSUE #1980: PF4 Expandable: button is missing type attribute
The Expandable component button is missing the type attribute.  React has a related issue [#2339](https://github.com/patternfly/patternfly-react/issues/2339)

christiemolloy:
close #1980 

I think the problem was that there was no "type" set on the toggle button in the expandable component, so when it was added to a form the type defaulted to submit. 

I think its safe to add the type as button to this component because I couldnt see it being a reset/submit type. 

Maybe we should also evaluate the variations of the button component.

### DIFF
--- a/src/patternfly/components/Button/docs/code.md
+++ b/src/patternfly/components/Button/docs/code.md
@@ -5,6 +5,9 @@ Always add a modifier class to add color to the button.
 ## Button vs link
 Semantic buttons and links are important for usability as well as accessibility. Using an `a` instead of a `button` element to perform user initiated actions should be avoided, unless absolutely necessary.
 
+## Button Types
+The default type for a button is button, which must be set to avoid having a button in a form incorrectly default to type "submit". If a button is used in a submit form, the type should be set to "submit". If a button is used to reset a form, the type should be set to "reset".
+
 ## Accessibility
 
 | Attribute | Applied to | Outcome |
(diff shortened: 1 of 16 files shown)


srambach:
I agree with @mattnolting - let's add this to the button component. And should we add `type="button"` on a button by default then?

christiemolloy:
Updated @srambach and @mattnolting

src/patternfly/components/Button/docs/code.md
```diff
@@ -5,6 +5,9 @@ Always add a modifier class to add color to the button.
 ## Button vs link
 Semantic buttons and links are important for usability as well as accessibility. Using an `a` instead of a `button` element to perform user initiated actions should be avoided, unless absolutely necessary.
 
+## Button Types
+The default type for a button is button, which must be set to avoid having a button in a form incorrectly default to type "submit". If a button is used in a submit form, the type should be set to "submit". If a button is used to reset a form, the type should be set to "reset".
```
jgiardino: Do we want to include this statement here?  The `type` attribute is documented on MDN as standard html, and not specific to patternfly. And the statement about what the default is is specific to our handlebar, but is not something our consumers would care about.

christiemolloy: Agreed I honestly wasn't too sure about putting it in but wanted to see what others thought.

PR #1989, opened 2019-06-26
fix(wizard): adjust spacing for cancel action

### LINKED ISSUE #1894: Update spacer for wizard cancel action
Currently in the wizard there is a 16px spacer for the cancel button
Screen Shot 2019-06-05 at 1 00 56 PM
The white space around this action is misleading as well because it is done with padding so the click area is much larger than it appears visually.

For an action that is so destructive I propose we use a 48px spacer to avoid accidentally hitting this cancel action instead of the back button. 

Screen Shot 2019-06-05 at 1 08 19 PM

matthewcarleton:
closes #1894 

This tackles just the wizard and the form actions, I think the modal actions are out of scope for this pr so I will create a follow up issue.

FYI: @mceledonia @mcarrano

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -39,6 +39,13 @@
     --pf-c-wizard__close--Right: var(--pf-c-wizard__close--lg--Right);
   }
 
+  // Cancel
+  --pf-c-wizard__cancel--MarginLeft: var(--pf-global--spacer--lg);
+
+  @media screen and (min-width: $pf-global--breakpoint--lg) {
+    --pf-c-wizard__cancel--MarginLeft: var(--pf-global--spacer--xl);
+  }
+
   // Title
   --pf-c-wizard__title--PaddingRight: var(--pf-global--spacer--2xl);
 
@@ -488,4 +495,8 @@
       margin-right: var(--pf-c-wizard__footer--child--MarginRight);
     }
   }
+}
+
+.pf-c-wizard__cancel {
+  margin-left: var(--pf-c-wizard__cancel--MarginLeft);
 }
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -40,10 +40,10 @@
   }
 
   // Cancel
-  --pf-c-wizard__cancel--MarginLeft: var(--pf-global--spacer--lg);
+  --pf-c-wizard__footer__link--MarginLeft: var(--pf-global--spacer--lg);
 
   @media screen and (min-width: $pf-global--breakpoint--lg) {
-    --pf-c-wizard__cancel--MarginLeft: var(--pf-global--spacer--xl);
+    --pf-c-wizard__footer__link--MarginLeft: var(--pf-global--spacer--xl);
   }
 
   // Title
@@ -497,6 +497,6 @@
   }
 }
 
-.pf-c-wizard__cancel {
-  margin-left: var(--pf-c-wizard__cancel--MarginLeft);
+.pf-c-wizard__footer__link {
+  margin-left: var(--pf-c-wizard__footer__link--MarginLeft);
 }
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -39,13 +39,6 @@
     --pf-c-wizard__close--Right: var(--pf-c-wizard__close--lg--Right);
   }
 
-  // Cancel
-  --pf-c-wizard__footer__link--MarginLeft: var(--pf-global--spacer--lg);
-
-  @media screen and (min-width: $pf-global--breakpoint--lg) {
-    --pf-c-wizard__footer__link--MarginLeft: var(--pf-global--spacer--xl);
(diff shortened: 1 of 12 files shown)


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -39,6 +39,13 @@
     --pf-c-wizard__close--Right: var(--pf-c-wizard__close--lg--Right);
   }
 
+  // Cancel
+  --pf-c-wizard__footer__link--MarginLeft: var(--pf-global--spacer--lg);
```
srambach: I don't think we use two __ separators, do we? Instead I think I'd use `--pf-c-wizard__footer-link` (and so forth throughout)

src/patternfly/demos/BasicForms/examples/horizontal-form-demo-example.hbs
```diff
@@ -57,7 +57,7 @@
         {{#> button button--modifier="pf-m-primary"}}
           Submit form
         {{/button}}
-        {{#> button button--modifier="pf-m-secondary"}}
+        {{#> button button--modifier="pf-m-link pf-c-form__actions__link"}}
```
srambach: Also here I'd call it `pf-c-form__action-link`

christiemolloy:
I think we should update the margin-left to be 16px rather than 24px, it seems like too much space to me. Curious what designers think @mceledonia 

Screen Shot 2019-06-27 at 4 44 01 PM


Screen Shot 2019-06-27 at 4 44 10 PM

christiemolloy:
@mcoker this reminds me of a potential place to implement an offset for the button component.

mcarrano:
This looks good to me @matthewcarleton . I will defer to @mceledonia to verify the proper spacing.

matthewcarleton:
@christiemolloy @srambach thanks for the quick review :) 
I've updated based on your feedback. I think that spacer suggestion works nicely @christiemolloy, @mceledonia feel free to chime in.
@christiemolloy are you saying we might have a modifier for buttons in general that will allow for offset? If that's the case we might not need this at all. How would it work? Would it be a fixed value? I'm interested! :)

christiemolloy:
@matthewcarleton I can't find the PR where @mcoker mentioned it and I'm not entirely sure of the plan, so should wait until then to merge this issue in case it can be fixed with this new method.

mattnolting:
Seems like, if this is a repeating design pattern, we could consider a `.pf-c-button-group` component like `.pf-c-chip-group`. If this pattern only exists in the wizard, it could be added to `.pf-c-wizard`.

Maybe something like:
```
--pf-c-button-group--child--MarginRight: var(--pf-global--spacer--md);
--pf-c-button-group--m-primary-actions--child--MarginRight = var(--pf-global--spacer--xl);

<div class="pf-c-button-group">
  <div class="pf-c-button-group pf-m-primary-actions">
    <button>button 1</button>
    <button>button 2</button>
  </div>
  <button>button 3</button>
</div>
```
or 
```
--pf-c-wizard__button-group--child--MarginRight: var(--pf-global--spacer--md);
--pf-c-wizard__button-group--m-primary-actions--child--MarginRight = var(--pf-global--spacer--xl);

<div class="pf-c-wizard__button-group">
  <div class="pf-c-wizard__button-group pf-m-primary-actions">
    <button>button 1</button>
    <button>button 2</button>
  </div>
  <button>button 3</button>
</div>
```

matthewcarleton:
@christiemolloy @mattnolting that works for me. This solution came out of the conversation I had with @mcoker here #1894 
I could make sense to have a more global approach, this pattern shows up in Wizard, Modal, and form. 
How do we proceed?

mcoker:
> @matthewcarleton I can't find the PR where @mcoker mentioned it and I'm not entirely sure of the plan, so should wait until then to merge this issue in case it can be fixed with this new method.

The idea is that we could have a modifier for a button that offsets the button's position to the top, right, bottom, or left the amount of padding on the side of the offset, so that the icon/text in the button is better aligned with whatever's around it. I'm not sure how I feel about it since it will cause some odd positioning for visual purposes, which will impact elements around it, the target/click/focusable area, stuff like that. Just an idea after @christiemolloy had mentioned a couple of times that it would be useful in certain contexts.

Screen Shot 2019-07-01 at 4 29 51 PM
Screen Shot 2019-07-01 at 4 30 00 PM

mcoker:
@matthewcarleton @mattnolting that's the approach I suggested with the action row (or whatever we call it), too. A layout that spaces actions, and allows for different spacing for cancel or whatever we might want to put in a row of actions. Though I do think it should be a layout rather than a component, if we're going by our rule that a layout doesn't style anything and simply positions its children. The input group applies quite a few styles outside of positioning its children.

matthewcarleton:
@mcoker @mattnolting @christiemolloy it sounds like I jumped the gun a bit with this PR so we can close it if we want to take the layout approach. It sounds reasonable to me.

mattnolting:
@mcoker @christiemolloy @matthewcarleton I agree w/all, layout makes sense here. However, I think button offsets would lead to inconsistencies across our apps.

matthewcarleton:
Closing this in favour of #2021

PR #1990, opened 2019-06-26
feat(accordion): add two variations of html markup

### LINKED ISSUE #1913: Accordion - update examples to show two html variations
Here are a few examples of accordions out there:
https://www.w3.org/TR/wai-aria-practices-1.1/#accordion
https://inclusive-components.design/collapsible-sections/
https://www.scottohara.me/blog/2017/10/25/accordion-release.html
https://a11yproject.com/patterns/

The first 3 use headings, without a definition list. The last link includes an example, "Simple Accordion", that uses a definition list and not headings.

Our pattern uses both, which isn't considered valid html. Instead, we should support one or the other. Our core css currently supports not including a heading, but it doesn't support not using the `<dl>` elements.

**Updates needed for this issue:**

- Update the examples shown to use either headings or definition list
- Include an extra example showing the other variation
- Make sure the css gets applied to both variations as expected
- Create an issue in react to make similar updates when this issue closes

christiemolloy:
closes #1913

### DIFF
--- a/src/patternfly/components/Accordion/docs/code.md
+++ b/src/patternfly/components/Accordion/docs/code.md
@@ -1,10 +1,11 @@
 ## Overview
 
-The accordion component can be built in two different ways. 
-The default way uses `<div>` and `<h3>` tags to build the component. 
-`.pf-c-accordion` uses `<div>`, `.pf-c-accordion__toggle` uses `<h3><button>`, and `.pf-c-accordion__expanded-content` uses `<div>`.
-This is the definition list variation:
-`.pf-c-accordion` uses `<dl>`, `.pf-c-accordion__toggle` uses `<dt><button>`, and `.pf-c-accordion__expanded-content` uses `<dd>`.
+There are two variations to build the accordion component:
+One way uses `<div>` and `<h1 - h6>` tags to build the component. 
+In these examples `.pf-c-accordion` uses `<div>`, `.pf-c-accordion__toggle` uses `<h3><button>`, and `.pf-c-accordion__expanded-content` uses `<div>`. The heading level that you use should fit within the rest of the headings outlined on your page.
+
+Another variation is using the definition list:
+In these examples `.pf-c-accordion` uses `<dl>`, `.pf-c-accordion__toggle` uses `<dt><button>`, and `.pf-c-accordion__expanded-content` uses `<dd>`.
 
 
 ## Accessibility
@@ -22,7 +23,7 @@ This is the definition list variation:
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-accordion` | `<div>`, `<dl>` | Initiates an accordion component. **Required**|
-| `.pf-c-accordion__toggle` | `<h3><button>`, `<dt><button>` | Initiates a toggle in the accordion. **Required** |
+| `.pf-c-accordion__toggle` | `<h1-h6><button>`, `<dt><button>` | Initiates a toggle in the accordion. **Required** |
 | `.pf-c-accordion__toggle-text` | `<span>` | Initiates the text inside the toggle. **Required** |
 | `.pf-c-accordion__toggle-icon` | `<i>` | Initiates the toggle icon. **Required** |
 | `.pf-c-accordion__expanded-content` | `<div>`, `<dd>` | Initiates expanded content. **Must be paired with a button** |
--- a/src/patternfly/components/Accordion/docs/code.md
+++ b/src/patternfly/components/Accordion/docs/code.md
@@ -21,11 +21,11 @@ This is the definition list variation:
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-accordion` | `<div>` `<dl>` | Initiates an accordion component. **Required**|
-| `.pf-c-accordion__toggle` | `<h3><button>` `<dt><button>` | Initiates a toggle in the accordion. **Required** |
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Accordion/docs/code.md
```diff
@@ -15,11 +21,11 @@
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-accordion` | `<dl>` | Initiates an accordion component. **Required**|
-| `.pf-c-accordion__toggle` | `<dt><h3><button>` | Initiates a toggle in the accordion. **Required** |
+| `.pf-c-accordion` | `<div>` `<dl>` | Initiates an accordion component. **Required**|
```
mattnolting:
```suggestion
| `.pf-c-accordion` | `<div>`, `<dl>` | Initiates an accordion component. **Required**|
```

src/patternfly/components/Accordion/docs/code.md
```diff
@@ -15,11 +21,11 @@
...
+| `.pf-c-accordion__toggle` | `<h3><button>` `<dt><button>` | Initiates a toggle in the accordion. **Required** |
```
mattnolting:
```suggestion
| `.pf-c-accordion__toggle` | `<h3><button>`, `<dt><button>` | Initiates a toggle in the accordion. **Required** |
```

src/patternfly/components/Accordion/docs/code.md
```diff
@@ -15,11 +21,11 @@
...
 | `.pf-c-accordion__toggle-text` | `<span>` | Initiates the text inside the toggle. **Required** |
 | `.pf-c-accordion__toggle-icon` | `<i>` | Initiates the toggle icon. **Required** |
-| `.pf-c-accordion__expanded-content` | `<dd>` | Initiates expanded content. **Must be paired with a button** |
+| `.pf-c-accordion__expanded-content` | `<div>` `<dd>` | Initiates expanded content. **Must be paired with a button** |
```
mattnolting:
```suggestion
| `.pf-c-accordion__expanded-content` | `<div>`, `<dd>` | Initiates expanded content. **Must be paired with a button** |
```

mattnolting:
Lgtm 👍

jgiardino:
Thanks for updating this!!

Off-thread, @christiemolloy had asked about what the default variation should be, between the one with definition list and one with headings. My understanding with the react component is that if `headingLevel` is not included, then the definition list variation would be used. If `headingLevel` is specified, then we use the variation with headings and use the heading level specified in that prop. In that case, I think the variation with definition list would be considered the default, at least in the context of react. In the context of core, do we have a concept of default? Do our defaults in react mean anything in the context of core? I'm a little inclined to say that what's more important is we match examples and example order between the two. And the first react example doesn't _have_ to be the variant that uses defaults. 

I'm curious what others think, but in this example, I think we can leave the order as it is, and have the react examples updated to match the examples shown in core.

Other than that question, there's one minor issue I noticed when I expanded the code for the preview of the `<dl>` variation, where an extra `</dt>` is included:

```
  <dt>
    <button class="pf-c-accordion__toggle" aria-expanded="false">
      <span class="pf-c-accordion__toggle-text">
   Item one</span>
      <i class="fas fa-angle-right pf-c-accordion__toggle-icon" aria-hidden="true"></i>
    </button>
    </h3>
  </dt>
  </dt> <<<<<< EXTRA </dt> HERE
  <dd class="pf-c-accordion__expanded-content" hidden>
    <div class="pf-c-accordion__expanded-content-body">
      This text is hidden
    </div>
  </dd>
```

src/patternfly/components/Accordion/docs/code.md
```diff
@@ -1,5 +1,11 @@
 ## Overview
 
+The accordion component can be built in two different ways. 
+The default way uses `<div>` and `<h3>` tags to build the component.
```
jgiardino: Related to my previous comment, I'm not sure that we want to label anything in core as a default, since the concept of defaults doesn't really exist here the way they do in the react repo. In react, there are defaults based on empty props, and for this component, I think the default variation would be the definition list variation.

Here, I would just say that one variation uses definition list and another uses headings. And also that we are using `<h3>` but really the heading levels that you use should fit within the rest of the heading outline on your page.

src/patternfly/components/Accordion/docs/code.md
```diff
@@ -15,11 +21,11 @@
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-accordion` | `<dl>` | Initiates an accordion component. **Required**|
-| `.pf-c-accordion__toggle` | `<dt><h3><button>` | Initiates a toggle in the accordion. **Required** |
+| `.pf-c-accordion` | `<div>`, `<dl>` | Initiates an accordion component. **Required**|
+| `.pf-c-accordion__toggle` | `<h3><button>`, `<dt><button>` | Initiates a toggle in the accordion. **Required** |
```
jgiardino: Since this is the only `<button>` in the accordion, maybe we can change this:
`<h3><button>`, `<dt><button>`
to just be this:
`<button>`

One benefit of this is not having to explain that the `<h3>` should really be a heading level that fits the rest of the heading outline on your page.

christiemolloy: per our offline conversation, updating this to be <h1 - h6>

christiemolloy:
Updated @jgiardino

PR #2002, opened 2019-06-28
fix(padding): split out padding vars (iss1773)

### LINKED ISSUE #1773: Split out padding vars
```
src//patternfly/components/EmptyState/empty-state.scss:  padding: var(--pf-c-empty-state--Padding);
src//patternfly/layouts/Bullseye/bullseye.scss:  padding: var(--pf-l-bullseye--Padding);
src//patternfly/components/Alert/alert.scss:  padding: var(--pf-c-alert__icon--Padding);
```

1. We should use vars for padding-top, padding-right, padding-bottom, and padding-left, instead of a single value for all 4. 
1. Also make sure any modifiers or changes to these paddings in other selectors override the original vars where applicable. 
1. Bullseye shouldn't need `padding: 0;` and `margin: 0;` as it gets that by default.

lareplog:
Added vars for padding-top, padding-right, padding-bottom, padding-left. Replaced any use of old padding value with the new 4 vars where applicable. closes #1773

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -6,7 +6,10 @@
   --pf-c-alert--grid-template-rows: 1fr auto;
 
   // Icon
-  --pf-c-alert__icon--Padding: var(--pf-global--spacer--md);
+  --pf-c-alert__icon--PaddingTop: var(--pf-global--spacer--md);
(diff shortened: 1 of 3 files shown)


mcoker:
Thanks for the PR @lareplog! Since this introduces breaking changes, I created a separate branch where we will track our breaking change commits and added this work there https://github.com/mcoker/patternfly-next/tree/breaking-changes-10.28.19, so we can close this PR.

PR #2005, opened 2019-07-01
feat(button): add input modifier

### LINKED ISSUE #1926: Update input group button styling to match CopyClipboard
The button styling for copy to clipboard looks better than the input group button styling, and in my opinion these should be consistent. Despite being separate components, to the end user they are the same, a button attached to an input field.

Copy to clipboard styling (currently hovering over copy button, invisible pointer)
https://deploy-preview-1813--pf-next.netlify.com/components/ClipboardCopy/examples/
image
As a side note here, I think changing to 2px/blue on hover looks better than the current black to blue change, which isn't very noticeable. Then in the case of the dropdown arrow, leaving the 2px blue style active. If we made this change it would apply to all buttons/inputs/dropdowns that use this style, something to consider.

Current input group button styling
https://deploy-preview-1813--pf-next.netlify.com/components/InputGroup/examples/
image

How I think it should be:
image

image

CC @kybaker @mcarrano @rachael-phillips @LHinson @mcoker

christiemolloy:
closes #1926 

As per @mcoker suggestion, added a new modifier to the button component that styles it for use in with an input.

Also updated the clipboard copy component, context selector component and input component to now use this button.


src/patternfly/components/ClipboardCopy/clipboard-copy-group-toggle.hbs
```diff
@@ -1,4 +1,4 @@
-<button class="pf-c-clipboard-copy__group-toggle{{#if clipboard-copy-group-toggle--modifier}} {{clipboard-copy-group-toggle--modifier}}{{/if}}"
+<div class="pf-c-clipboard-copy__group-toggle{{#if clipboard-copy-group-toggle--modifier}} {{clipboard-copy-group-toggle--modifier}}{{/if}}"
```
christiemolloy: @mcoker @mattnolting I removed the hbs file for `pf-c-clipboard-copy__group-copy` because it was only just a wrapper. I didn't remove the hbs for `pf-c-clipboard-copy__group-toggle` because its used to determine whether its expanded or not, what do you both think about that?

mcoker:
Would we also want to use this variation of the button as the dropdown (and components like it) toggle with left/right padding changes?

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
```diff
@@ -1,38 +1,40 @@
...
+  // --pf-c-clipboard-copy__group-toggle--BorderTopColor: var(--pf-global--BorderColor--300);
+  // --pf-c-clipboard-copy__group-toggle--BorderRightColor: var(--pf-global--BorderColor--300);
+  // --pf-c-clipboard-copy__group-toggle--BorderBottomColor: var(--pf-global--BorderColor--200);
+  // --pf-c-clipboard-copy__group-toggle--BorderLeftColor: var(--pf-global--BorderColor--300);
+  // --pf-c-clipboard-copy__group-toggle--hover--BorderBottomColor: var(--pf-global--active-color--100);
+  // --pf-c-clipboard-copy__group-toggle--active--BorderBottomWidth: var(--pf-global--BorderWidth--md);
+  // --pf-c-clipboard-copy__group-toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
+  // --pf-c-clipboard-copy__group-toggle--focus--BorderBottomWidth: var(--pf-global--BorderWidth--md);
+  // --pf-c-clipboard-copy__group-toggle--focus--BorderBottomColor: var(--pf-global--active-color--100);
+  --pf-c-clipboard-copy__group-toggle--c-button--m-expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
```
mcoker: I wonder if we should add a modifier to the button for this, like `pf-m-expanded/selected/active` or something like that, since I imagine we will want to apply these styles in other contexts, too?

src/patternfly/components/ClipboardCopy/examples/clipboard-copy-expansion-example.hbs
```diff
@@ -3,11 +3,15 @@ Editable
 {{#> clipboard-copy clipboard-copy--id="3"}}
 	{{#> clipboard-copy-group}}
     {{#> clipboard-copy-group-toggle clipboard-copy-group-toggle--attribute=(concat 'id="toggle-' clipboard-copy--id '" aria-labelledby="toggle-' clipboard-copy--id ' text-input-' clipboard-copy--id '" aria-controls="content-' clipboard-copy--id '"')}}
```
mcoker: Is there a reason we don't just add the group-toggle class and all of the attributes to the button, like we did with the group-copy button?

christiemolloy: I added a comment here about it: https://github.com/patternfly/patternfly-next/pull/2005#discussion_r299140165  
but i think its possible to give it a try

christiemolloy:
> > As a side note here, I think changing to 2px/blue on hover looks better than the current black to blue change, which isn't very noticeable. Then in the case of the dropdown arrow, leaving the 2px blue style active. If we made this change it would apply to all buttons/inputs/dropdowns that use this style, something to consider.
> 
> Should the `:hover` weight be 2px ^^?

yep I added a variable for the hover weight:  `--pf-c-button--m-input--hover--BorderBottomWidth` can you see it

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
christiemolloy: My take was that the button should just hold the generic hover, active and focus styles, and then components that use the button can add their own modifiers since they're specific to that component, and we can't forsee all of the modifier names. But what is your main reasoning for adding these modifiers to the button I'm open to adding them.

christiemolloy:
> Would we also want to use this variation of the button as the dropdown (and components like it) toggle with left/right padding changes?

@mceledonia do you have an opinion here? I don't see why we couldn't, but it would likely be a breaking change and take a lot of consideration it might get tricky, because when we do it for dropdown, should we also do it for select? and then some selects are built with a div.

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
mcoker: The name is the tricky part, but it's a common form control display for active/focus - the thicker, blue bottom border. We would also use it for expanded as the dropdown toggle, too, if we use it there.

christiemolloy: So are you suggesting adding `pf-m-expanded` to the button since we know we're going to be using that in the clipboard copy at least.

```
  --pf-c-button--m-input--m-expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
  --pf-c-button--m-input--m-expanded--BorderBottomColor: var(--pf-global--active-color--100);
```

**pf-m-selected/active:** not sure about these classes because I think the active/focus styles cover these, or am I missing something?

mcoker: Yep, that's it. We can always not since it's only used here for now, and add it later if we end up using it elsewhere (like the dropdown or other components). It just looks like a state this type of button would need. 

Also what's the reasoning behind `pf-m-input` as the modifier name? Is it because the intended use is always going to be "for use in with an input." (as in connected to an input?) I wonder if that's going to end up being limiting.

christiemolloy: Added the expanded modifier to the button.. @mceledonia do you have any ideas for a name for this button?

mceledonia: @christiemolloy Is this for the overall modifier to change styling to input group? If so, maybe something like input-group-button or something specific like that. Maybe something that rolls off the tongue a bit better :)

christiemolloy: Its for these buttons, thanks for the suggestion!

Screen Shot 2019-07-10 at 10 39 12 AM

mceledonia:
@christiemolloy Yeah we would want this styling for all interactive elements in an input group, so both buttons and dropdowns!

mcoker:
> Yeah we would want this styling for all interactive elements in an input group

@mceledonia what about for use outside of an input group, could this variation of the button be used as the toggle for standalone dropdown, select, options menu, etc components? The style is 

@matthewcarleton has also mentioned there is interest from products in a primary button style toggle for dropdown toggles, too. If we support that, I think it makes sense that the toggle for those components is a variation of our button, whether it's the style we're adding in this PR, the primary button, etc.

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
christiemolloy: @mceledonia on second thought, if we use this button in the individual dropdown/select etc. then i dont know if `input-group-button` would be correct, since it wouldn't be part of an input. Maybe a name that refers to the style more so the usage?

matthewcarleton:
just to add to what @mcoker is saying, this issue is tracking what's been seen as a useful addition on the product side. https://github.com/patternfly/patternfly-next/issues/1872

christiemolloy:
@mcoker @mattnolting should I make this not a breaking change by putting back in the code for copy to clipboard etc. or do you want to leave this until the breaking change period?

christiemolloy:
@mceledonia should there be a disabled variation of the input modifier? By default it looks like this when disabled is added: 

Screen Shot 2019-09-16 at 2 38 25 PM


All buttons have a disabled styling so was wondering what you thought we should do?

mceledonia:
> @mceledonia should there be a disabled variation of the input modifier? By default it looks like this when disabled is added:
> 
> Screen Shot 2019-09-16 at 2 38 25 PM
> 
> All buttons have a disabled styling so was wondering what you thought we should do?

Yeah we should have a disabled style here, I wonder if we go with the styling for disabled like what we saw in @maryshak1996 's touchspin design? Essentially - text/icon goes to secondary text color and we remove the underline. I think that would work well, what do you think @kybaker @mcarrano @christiemolloy

mcarrano:
That works for me.

mattnolting:
> @mcoker @mattnolting should I make this not a breaking change by putting back in the code for copy to clipboard etc. or do you want to leave this until the breaking change period?

Leave the code in for this PR and create an issue to remove it during breaking change period?

christiemolloy:
> > @mcoker @mattnolting should I make this not a breaking change by putting back in the code for copy to clipboard etc. or do you want to leave this until the breaking change period?
> 
> Leave the code in for this PR and create an issue to remove it during breaking change period?

sorry I wrote that back in July @mattnolting

christiemolloy:
@mceledonia how does this look:

**input button disabled:**

Screen Shot 2019-09-16 at 4 52 43 PM

**input button disabled in context of a form:**

Screen Shot 2019-09-16 at 4 52 25 PM

christiemolloy:
@mcarrano @mceledonia @kybaker we have had some trouble naming this variation because we don't technically know the scope of it yet... it will likely be used as a button in a dropdown, which means we have to change it from `pf-m-input`, right now we are going with `pf-m-control` since the button is always used to control something, whether its a dropdown menu, an input, a search field, a clipboard copy... but wondering what you guys thought because it will be a breaking change to change the name down the road.

maryshak1996:
This looks great to me @christiemolloy @mceledonia !

mcarrano:
@christiemolloy I don't see the problem with calling this 'input'. An input group is always used to supply an input, right? That may be from a text field as in the search or a select list.

Regarding the disabled question, keep in mind that if this will be disabled then the entire input group will show disabled, i.e. it does not make any sense to allow the user to type into the search field and then have the action/input disabled. Don't know if this changes any thinking on this, but wanted to make sure that we are thinking of the entire input group as one thing.

christiemolloy:
@mcarrano  I'm not sure if we could say that `pf-m-input` would hold when its used in a dropdown component?

mattnolting:
> @christiemolloy I don't see the problem with calling this 'input'. An input group is always used to supply an input, right? That may be from a text field as in the search or a select list.

@mcarrano Also can be used with textareas, html selects, dropdown, custom select, other things down the road. That's why we thought `pf-m-control` worked, because it implies that it controls something, but not necessarily always an input :)

mcarrano:
@christiemolloy @mattnolting I must be missing a use case here. In what circumstances would this be used as part of a dropdown, a text area, or another control? I though this was specific to an input group of things built from an input group like a Touchspin.

christiemolloy:
Sorry about not communicating fully. The potential plan down the road would be to replace the button in the dropdown, select, context selector etc components with this button, as they all share exactly the same styles. This will allow us to reuse a lot of code and make it easier to update all components that use this button @mcarrano

mcarrano:
@christiemolloy Thanks for the clarification. I understand better now. I'd say go with pf-m-control unless someone has a better idea.

src/patternfly/components/Button/button.scss
```diff
@@ -316,6 +332,66 @@
     }
   }
 
+  &.pf-m-control {
+    position: relative;
```
mcoker: this comes with `.pf-c-button`

src/patternfly/components/Button/button.scss
```diff
@@ -316,6 +332,66 @@
...
+    border: none;
```
mcoker: same ^

src/patternfly/components/Button/button.scss
```diff
@@ -316,6 +332,66 @@
...
+
+    &::before {
```
mcoker: any reason you're using `::before` instead of `::after`? Seems like you would get most of the definitions for free, then you can just redefine `--pf-c-button--BorderWidth`. Then I might separate `border: var(--pf-c-button--BorderWidth) solid var(--pf-c-button--BorderColor);` into `border: var(--pf-c-button--BorderWidth) solid;` and `border-color: var(--pf-c-button--BorderColor);`, then redefine `--pf-c-button--BorderColor` as

src/patternfly/components/Button/button.scss
```diff
@@ -316,6 +332,66 @@
...
+        border-bottom-width: var(--pf-c-button--m-control--m-expanded--before--BorderBottomWidth);
+      }
+    }
+
+    &:disabled,
+    &.pf-m-disabled {
+      background-color: var(--pf-c-button--m-control--disabled--BackgroundColor);
+
+      &::before {
+        --pf-c-button--m-control--before--BorderBottomColor: var(--pf-c-button--m-control--before--BorderTopColor);
```
mcoker: This works for me, though we could also be safe and create `--pf-c-button--m-control--disabled--before--BorderBottomColor`, which allows flexbility for this to be its own var if we want to change it in the future without a breaking change. Or if you go with the approach above and create `--pf-c-button--m-control--before--BorderColor`, then you could create `--pf-c-button--m-control--disabled--before--BorderColor` and just redefine `--pf-c-button--m-control--before--BorderColor: var(--pf-c-button--m-control--disabled--before--BorderColor)` here.

src/patternfly/components/Button/docs/code.md
```diff
@@ -30,6 +30,8 @@ Semantic buttons and links are important for usability as well as accessibility.
 | `.pf-m-plain` | `.pf-c-button` | Modifies for icon styles. This button has no background or border, uses a standard text color, and is used for `.pf-m-plain` icon buttons such as close, expand, kebab, etc. |
 | `.pf-m-inline` | `.pf-c-button.pf-m-link` | Modifies for inline styles. This button is presented similar to a normal link, has no padding, and is displayed inline with other inline content. |
 | `.pf-m-block` | `.pf-c-button` | Creates a block level button. |
+| `.pf-m-control` | `.pf-c-button` | Modifies for control styles. **Note: ** This modifier should only be used when using buttons in the Input or ClipboardCopy components. |
```
mcoker:
```suggestion
| `.pf-m-control` | `.pf-c-button` | Modifies for control styles. **Note:** This modifier should only be used when using buttons in the Input Group or Clipboard Copy components. |
```

src/patternfly/components/Button/docs/code.md
```diff
@@ -30,6 +30,8 @@ Semantic buttons and links are important for usability as well as accessibility.
...
+| `.pf-m-expanded` | `.pf-c-button.pf-m-input` | Modifies an input button to be expanded. |
```
mcoker: Maybe instead of "to be expanded" change to "for the expanded state" or something. Currently it reads like the button is expandable, instead of representing the expanded state of something else.

src/patternfly/components/Button/examples/button-states-example.hbs
```diff
@@ -100,3 +100,19 @@
 {{#> button button--modifier="pf-m-link pf-m-inline" button--attribute="disabled"}}
   Inline link button
 {{/button}}
+<br><br>
+{{#> button button--modifier="pf-m-control"}}
+  Control button
```
mcoker: Do you think this variations and "link" and "inline link" above need the word "button" in the examples?

src/patternfly/components/ContextSelector/examples/context-selector-example.hbs
```diff
@@ -11,7 +11,7 @@
       {{#> input-group}}
         {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="search"' 'placeholder="Search"' 'id="textInput1"' 'name="textInput1"' 'aria-labelledby="' context-selector--id '-search-button"')}}
         {{/form-control}}
-        {{#> button button--modifier="pf-m-tertiary" button--attribute=(concat 'id="' context-selector--id '-search-button"' 'aria-label="Search menu items"')}}
+        {{#> button button--modifier="pf-m-input" button--attribute=(concat 'id="' context-selector--id '-search-button"' 'aria-label="Search menu items"')}}
```
mcoker: the context selector and input group examples should use `pf-m-control` instead of `pf-m-input`

src/patternfly/components/Button/examples/button-states-example.hbs
christiemolloy: No I think its redundant, updated all the examples to remove button

src/patternfly/components/Button/button.scss
christiemolloy: Ah the reason I used before instead of after was because I based it off the clipboard-copy-group-copy, but after definitely makes sense in this case. Updated.

src/patternfly/components/Button/button.scss
```diff
@@ -316,6 +334,59 @@
     }
   }
 
+  &.pf-m-control {
+    &::after {
+      --pf-c-button--BorderWidth: var(--pf-c-button--m-control--after--BorderWidth);
+      --pf-c-button--BorderColor: var(--pf-c-button--m-control--after--BorderTopColor) var(--pf-c-button--m-control--after--BorderRightColor) var(--pf-c-button--m-control--after--BorderBottomColor) var(--pf-c-button--m-control--after--BorderLeftColor);
+
+      border-radius: unset;
```
mcoker: Any reason we wouldn't just use `0` here? We use `unset` very sparingly, and in this context, while it will still evaluate to `initial` (which is `0` for this property) we could just set that directly since we don't want it to have a `border-radius`.

src/patternfly/components/Button/examples/button-variations-example.hbs
```diff
@@ -10,15 +10,29 @@
...
+  Inline link
 {{/button}}
+
+<br>
+<br>
+
+{{#> button button--modifier="pf-m-control"}}
+  Control
+{{/button}}
+{{#> button button--modifier="pf-m-control" button--attribute='aria-label="copy input"'}}
```
mcoker:
```suggestion
{{#> button button--modifier="pf-m-control" button--attribute='aria-label="Copy input"'}}
```

src/patternfly/components/InputGroup/examples/input-group-example.hbs
```diff
@@ -96,7 +96,7 @@
 {{#> input-group}}
   {{#> form-control controlType="input" input="true" form-control--attribute='type="text" id="textInput10" name="textInput10" aria-label="input example with popover"'}}
```
mcoker: super nit - would you mind sentence casing these `aria-label` values?

```suggestion
  {{#> form-control controlType="input" input="true" form-control--attribute='type="text" id="textInput10" name="textInput10" aria-label="Input example with popover"'}}
```

christiemolloy: Updated the file, most of the aria-labels weren't capitalized

mcoker:
@christiemolloy @mcarrano is the goal to not support the previous ways of attaching buttons to the  input group component  where we square the borders except for the left side of the first button and right side of the last one? Or are we going to continue to allow that indefinitely?

If we aren't going to allow it, we should create a follow up issue to remove the CSS that does that in a breaking change release.

PR #2006, opened 2019-07-01
feat(applauncher): add button example to menu item

### LINKED ISSUE #1944: AppLauncher: pf-c-app-launcher__menu-item should be applicable to non-anchor element as well
in addition to `a` element for .pf-c-app-launcher__menu-item, make it work also with `button` and maybe `div` as well. In react, if consumer wants to use a routing library, it will often clash if it internally renders also to an anchor and thus it would be a doubly nested `a`.

christiemolloy:
closes #1944 

@mcoker @mattnolting what do you guys think about showing a `<div>` as an example in the menu. Do you know if thats a common case? I don't think we do it in the dropdown.

### DIFF
--- a/src/patternfly/components/AppLauncher/examples/app-launcher-collapsed-example.hbs
+++ b/src/patternfly/components/AppLauncher/examples/app-launcher-collapsed-example.hbs
@@ -1,2 +1,8 @@
-{{#> app-launcher id="app-launcher-example"}}{{/app-launcher}}
+{{#> app-launcher id="app-launcher-example"}}
+  {{#> app-launcher-menu}}
+    <li>{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link{{/app-launcher-menu-item}}</li>
+    <li>{{#> app-launcher-menu-item app-launcher-menu-item--type="div" app-launcher-menu-item--attribute='href="#"'}}Link two{{/app-launcher-menu-item}}</li>
+    <li>{{#> app-launcher-menu-item app-launcher-menu-item--modifier="pf-m-disabled" app-launcher-menu-item--attribute='href="#" aria-disabled="true" tabindex="-1"'}}Disabled link{{/app-launcher-menu-item}}</li>
+  {{/app-launcher-menu}}
+{{/app-launcher}}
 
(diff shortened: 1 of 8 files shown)


src/patternfly/components/AppLauncher/examples/app-launcher-expanded-example.hbs
```diff
@@ -1 +1,8 @@
-{{#> app-launcher id="app-launcher-example-expanded" app-launcher--IsExpanded="true"}}{{/app-launcher}}
\ No newline at end of file
+{{#> app-launcher id="app-launcher-example-expanded" app-launcher--IsExpanded="true"}}
+  {{#> app-launcher-menu}}
+    <li>{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link one{{/app-launcher-menu-item}}</li>
```
mcoker: Instead of "Link one, Action two, Div three", I think you could just list "Link, Action, Div". Or if you want to include the numbers, it would be like "Link 1, Link 2, Action 1, Action 2" etc.

src/patternfly/components/AppLauncher/examples/app-launcher-collapsed-example.hbs
```diff
@@ -1,2 +1,8 @@
-{{#> app-launcher id="app-launcher-example"}}{{/app-launcher}}
+{{#> app-launcher id="app-launcher-example"}}
+  {{#> app-launcher-menu}}
+    <li>{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link{{/app-launcher-menu-item}}</li>
```
mcoker: Can we make the menu items in the collapsed and expanded examples match, so they're basically the same example, just in the collapsed and expanded states? I think that's why it was a single include before.

mattnolting:
@jschuler using a `<div>` instead of `<button>`, `<a>` or `<input type="button">` seems inherently wrong to do. Also, I can't find any accessible examples of converting a `<div>` to an accessible, clickable object. Are you saying use `<div>` as a wrapper with a nested `<a>`? In this case, I think it may be ok to support wrapping in this way, but I'm hesitant to show this as an example in core..
@christiemolloy @mcoker @matthewcarleton thoughts?

mcoker:
@mattnolting I left a similar comment in the issue - https://github.com/patternfly/patternfly-next/issues/1944#issuecomment-507443506. I think it depends on how it renders and what classes are applied to what elements as to whether we can even support it currently and if it's something we should show in an example.

matthewcarleton:
+1 on what @mattnolting is saying, it seems most reasonable to stick with just adding a button and dropping the `div` unless there's something I'm missing as a valid use case.

christiemolloy:
@mcoker  @mattnolting @matthewcarleton thanks for the input, I also agree with that. Updated

PR #2010, opened 2019-07-02
 feat(page): add utility icon focus styles in the masthead

### LINKED ISSUE #1998: Focus state for icons in masthead (notification badge)
from @maryshak1996: 

@mcarrano I wanted to ping this issue again, as I talked to @dlabaj about this a bit more today. I wanted to more specifically indicate the states of the masthead icons:
- DEFAULT: global-background-300
- HOVER: white
- OPENED: 36x36px black-800 circle appears behind icon (this is for ALL masthead icons, including notification bell)
I've updated the notification drawer docs as well as the docs for this issue, but here is a quick snapshot of what I'm referring to: 
Default:
Screen Shot 2019-06-27 at 3 04 47 PM
Hover:
Screen Shot 2019-06-27 at 3 04 57 PM
Opened:
Screen Shot 2019-06-27 at 3 05 10 PM

_Originally posted by @maryshak1996 in https://github.com/patternfly/patternfly-next/issues/1832#issuecomment-506472615_

christiemolloy:
closes #1998 

Hold off merging until #1925 is merged.


src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -21,6 +21,7 @@
   --pf-c-notification-badge--m-read--after--BackgroundColor: transparent;
   --pf-c-notification-badge--m-read--after--BorderColor: transparent;
   --pf-c-notification-badge--m-unread--after--BackgroundColor: var(--pf-global--active-color--400);
+  --pf-c-notification-badge--m-unread--c-button--m-masthead--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
```
christiemolloy: Any suggestions on naming this variable. Since the button wraps around the notification badge component, I'm wondering if it should be `--pf-c-button--m-masthead__notification-badge--m-unread--after--BorderColor` because then it matches the selector order. But it seems odd to have a variable in a component start with another variable's name. I think this is the first time this has occurred. Maybe theres another solution here. @mcoker @mattnolting @srambach

mcoker: You could move these styles to the button component since the styles you're trying to match come from the button.

But if I'm not mistaken, like we talked about in the last css meeting, I think you're creating a variation of the button here instead of just overwriting the button's styles in the page component? But then you're overwriting the notification badge styles based on being inside of another component (or modifying the notification badge styles in the button component, if you moved the styles). You could create a modifier for the notification badge for this use case. Though I'm wondering if it might be better just to overwrite the button and notification badge styles in the masthead instead. Maybe it wasn't a great idea to create a modifier for use when we need to style one component in another, unless it's a repeatable variation we'll use in other components.

maryshak1996:
@christiemolloy this is looking good! The only thing that I would point out from the design side is that the outline around the badge should match the background color of the grey circle when it that background is applied (below is the screenshot that I saw, and you can see the color is off here)
unnamed

christiemolloy:
where do you see that @maryshak1996 it should be updated

maryshak1996:
@christiemolloy woops, my bad I was looking at something old that was in my inbox! Just looked at the new stuff and it looks great to me!

src/patternfly/components/NotificationBadge/notification-badge.scss
christiemolloy: This background style is for all icons in the masthead so I can't create a modifier for just the notification badge, I don't think, that's why I targeted the plain modification of the button. 

@mcarrano  @maryshak1996 can you for see this background style for icons outside of the masthead?  Or is it only for the icons in the masthead?

mcoker: Couldn't you use `.pf-m-masthead` with the notification drawer as you've done for the button? I'm also not sure about masthead as the class. What if we have 2 modifiers for buttons in the masthead? That class also doesn't age well if we end up doing this somewhere else.

christiemolloy: Okay so are you suggesting to just overwrite the styles of buttons in the masthead of the page component, and then if another use case comes up for them, we can look at making it its own example in the button component ?

mcoker: I think so, it seems simpler and more appropriate just to make the change in the context of where it needs to change than trying to abstract it out this way. What do you think?

christiemolloy: Yeah I agree!

christiemolloy:
@mcoker updated so that the styles from the button are now in the page component, let me know what you think.

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -53,3 +54,18 @@
     --pf-c-notification-badge--after--BorderColor: var(--pf-c-notification-badge--m-read--after--BorderColor);
   }
 }
+
+// When the notification badge in the masthead is focused, the pseudo element on the notification badge will have the same border color as the background color on the button.
+/* stylelint-disable */
+.pf-c-page__header-tools {
```
mcoker: should this also be in the page component's CSS?

christiemolloy: I wasn't sure what to call the variable in that case, because its overriding a variable in the notification badge component, or is it ok to use these variable names here @mcoker

mcoker: Oh if you added this in the page component, you would just overwrite `border-color` since you don't overwrite variables in other components. And I think it would be something like

```
--pf-c-page__header-tools__c-button--focus__c-notification-badge--m-unread--after--BorderColor
```

christiemolloy: thanks for clarifying got a little confused there, updated. I want to come back to this PR once we merge #1925

src/patternfly/components/Page/page.scss
```diff
@@ -280,6 +290,36 @@
       display: block;
       visibility: visible;
     }
+
+    > .pf-c-button {
```
mcoker: These can also be dropdowns. I wonder if someone would put something else in there? Maybe we should style the header tools item in https://github.com/patternfly/patternfly-next/pull/1925 instead?

Screen Shot 2019-07-10 at 1 31 59 PM

christiemolloy:
closing this PR as it is now addressed in: #2010

PR #2026, opened 2019-07-08
docs(select): add a top expanded select example

### LINKED ISSUE #1986: PF4 Select:  Add the ability to set the direction of the component
The consumer would like to set the direction of the select menu (top or down) similar to the Dropdown component.

@srambach said there is already a modifier, can we also add an example to core.

boaz0:
closes #1986 

//cc @tlabaj @mcoker

### DIFF
--- a/src/patternfly/components/Select/examples/index.js
+++ b/src/patternfly/components/Select/examples/index.js
@@ -5,6 +5,7 @@ import Example from '@siteComponents/Example';
 import SelectSingleExampleRaw from '!raw!./select-single-example.hbs';
 import SelectDisabledExampleRaw from '!raw!./select-disabled-example.hbs';
 import SelectSingleExpandedExampleRaw from '!raw!./select-single-expanded-example.hbs';
+import SelectSingleTopExpandedExampleRaw from '!raw!./select-single-top-expanded-example.hbs';
 import SelectSingleExpandedSelectedExampleRaw from '!raw!./select-single-expanded-selected-example.hbs';
 import SelectSingleTypeaheadExampleRaw from '!raw!./select-single-typeahead-example.hbs';
 import SelectSingleTypeaheadExpandedExampleRaw from '!raw!./select-single-typeahead-expanded-example.hbs';
@@ -22,6 +23,7 @@ import SelectIconExampleRaw from '!raw!./select-icon-example.hbs';
 import SelectSingleExample from './select-single-example.hbs';
 import SelectDisabledExample from './select-disabled-example.hbs';
 import SelectSingleExpandedExample from './select-single-expanded-example.hbs';
+import SelectSingleTopExpandedExample from './select-single-top-expanded-example.hbs';
 import SelectSingleExpandedSelectedExample from './select-single-expanded-selected-example.hbs';
 import SelectSingleTypeaheadExample from './select-single-typeahead-example.hbs';
 import SelectSingleTypeaheadExpandedExample from './select-single-typeahead-expanded-example.hbs';
@@ -42,6 +44,7 @@ import SelectMultiTypeaheadExampleDoc from '../docs/select-multi-typeahead.md';
 import SelectCheckboxExampleDoc from '../docs/select-checkbox.md';
 import SelectPlainExampleDoc from '../docs/select-plain.md';
 import SelectIconExampleDoc from '../docs/select-icon.md';
+import SelectSingleTopExampleDoc from '../docs/select-single-top.md';
 import docs from '../docs/code.md';
 
 export const selectSingleExampleDoc = SelectSingleExampleDoc;
@@ -50,12 +53,14 @@ export const selectMultiTypeaheadExampleDoc = SelectMultiTypeaheadExampleDoc;
 export const selectCheckboxExampleDoc = SelectCheckboxExampleDoc;
 export const selectPlainExampleDoc = SelectPlainExampleDoc;
 export const selectIconExampleDoc = SelectIconExampleDoc;
+export const selectSingleTopExampleDoc = SelectSingleTopExampleDoc;
 export const Docs = docs;
 
 export default props => {
   const selectSingleExample = SelectSingleExample();
   const selectDisabledExample = SelectDisabledExample();
   const selectSingleExpandedExample = SelectSingleExpandedExample();
+  const selectSingleTopExpandedExample = SelectSingleTopExpandedExample();
   const selectSingleExpandedSelectedExample = SelectSingleExpandedSelectedExample();
   const selectSingleTypeaheadExample = SelectSingleTypeaheadExample();
   const selectSingleTypeaheadExpandedExample = SelectSingleTypeaheadExpandedExample();
@@ -82,6 +87,14 @@ export default props => {
       </Example>
       <Example heading="Select - single (expanded)" handlebars={SelectSingleExpandedExampleRaw} minHeight="20em">
         {selectSingleExpandedExample}
+      </Example>
+      <Example
+        className="is-expanded-top"
+        heading="Select - single (top expanded)"
+        docs={selectSingleTopExampleDoc}
+        handlebars={SelectSingleTopExpandedExampleRaw}
+      >
+        {selectSingleTopExpandedExample}
       </Example>
       <Example
         heading="Select - single (expanded and selected)"
--- a/src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
+++ b/src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
@@ -1,0 +1,3 @@
+{{#> select id="select-single-top-expanded" select--modifier="pf-m-top" select--IsExpanded="true"}}
+  Filter by birth month
+{{/select}}
--- a/src/patternfly/components/Select/select-single.hbs
+++ b/src/patternfly/components/Select/select-single.hbs
@@ -1,4 +1,4 @@
-<ul class="pf-c-select__menu{{#if select--modifier}} {{select--modifier}}{{/if}}"
+<ul class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
   aria-labelledby="{{id}}-label"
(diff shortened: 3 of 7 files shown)


src/patternfly/components/Select/examples/index.js
```diff
@@ -61,6 +64,13 @@ export default (props) => {
       <Example heading="Select - single (expanded)" handlebars={SelectSingleExpandedExampleRaw} minHeight="20em">
         {selectSingleExpandedExample}
       </Example>
+      <Example
+        heading="Select - single direction top (expanded)"
```
mcoker: What do you think about renaming this to `Select - single (top expanded)`? "single direction top" reads like "direction" is part of "single," as in "single direction," instead of "direction top".

src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
```diff
@@ -0,0 +1,4 @@
+<div style="padding:130px"></div>
```
mcoker: Users should be able to copy and paste the example code from the workspace into their application, so we try to avoid unnecessary elements with inline styles like this.

Have a look at how the top aligned dropdown example is setup.

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Dropdown/examples/index.js#L70-L72

https://github.com/patternfly/patternfly-next/blob/master/src/site/workspace.scss#L424-L428

mcoker:
thanks @boaz0! left a couple of comments.

src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
boaz0: Thanks!

src/patternfly/components/Select/examples/index.js
```diff
@@ -61,6 +64,14 @@ export default (props) => {
       <Example heading="Select - single (expanded)" handlebars={SelectSingleExpandedExampleRaw} minHeight="20em">
         {selectSingleExpandedExample}
       </Example>
+      <Example
+        className="is-expanded-top"
+        heading="Select - single (top expanded)"
+        handlebars={SelectSingleTopExpandedExampleRaw}
+        minHeight="20em"
```
mcoker: You don't need this now that you're using `className="is-expanded-top"`

src/patternfly/components/Select/select.scss
```diff
@@ -234,6 +234,12 @@
   .pf-c-select__toggle-clear + .pf-c-select__toggle-button {
     padding-left: var(--pf-c-select__toggle-button--PaddingLeft);
   }
+
+  .pf-c-select__toggle-arrow {
+    .pf-m-top.pf-m-expanded > & {
+      transform: var(--pf-c-select__toggle-arrow--m-top--m-expanded__toggle-arrow--Transform);
+    }
+  }
```
mcoker: nice! Can you remove this block - https://github.com/patternfly/patternfly-next/blob/e7fbc2dca3e3ad139ab85f66b65445cae66db587/src/patternfly/components/Select/select.scss#L252-L254?

src/patternfly/components/Select/examples/index.js
boaz0: Thanks a lot. I did miss that! :smile:

src/patternfly/components/Select/select.scss
```diff
@@ -234,6 +234,12 @@
   .pf-c-select__toggle-clear + .pf-c-select__toggle-button {
     padding-left: var(--pf-c-select__toggle-button--PaddingLeft);
   }
+
+  .pf-c-select__toggle-arrow {
+    .pf-m-top.pf-m-expanded > & {
```
mcoker: Can you scope the `.pf-m-top.pf-m-expanded` selector to `.pf-c-select` so that, in the case that the select is nested inside of something else with `.pf-m-top.pf-m-expanded`, this CSS doesn't accidentally apply? See the dropdown for a similar example

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Dropdown/dropdown.scss#L248-L257

```suggestion
    .pf-c-select.pf-m-top.pf-m-expanded > & {
```

src/patternfly/components/Select/select.scss
```diff
@@ -236,6 +236,13 @@
   .pf-c-select__toggle-clear + .pf-c-select__toggle-button {
     padding-left: var(--pf-c-select__toggle-button--PaddingLeft);
   }
+
+  .pf-c-select__toggle-arrow {
```
mcoker: Can you use the existing selector on L249? That will also keep the specificity down.

https://github.com/patternfly/patternfly-next/blob/26b59efaefdf9f58f1b7e5611ac503e844d4d33f/src/patternfly/components/Select/select.scss#L249

mcoker:
Can you also document this example so that we have the `.pf-m-top` class documented?

src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
```diff
@@ -0,0 +1,3 @@
+{{#> select id="select-single-top-expanded" select--modifier="pf-m-top" select--IsExpanded="true"}}
```
mcoker: using `select--modifier="pf-m-top"` is adding `.pf-m-top` to both `.pf-c-select` (not needed) and to `.pf-c-select__menu` (the element that needs it). Can you update `select-multi.hbs` and `select-single.hb` to use `{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}` instead of `{{#if select--modifier}} {{select--modifier}}{{/if}}`, and then change `select--modifier="pf-m-top"` to `select-menu--modifier="pf-m-top"` in this file?

FWIW I completely forgot about this PR (doh!!!) and opened another with the changes if you want to reference - https://github.com/patternfly/patternfly-next/pull/2164/files

There is also a typo in `select-multi.hbs` referenced in #2164 if you want to fix that, too.

boaz0:
Thanks @mcoker 
I updated this PR according to your PR. :smile:

src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
```diff
@@ -0,0 +1,3 @@
+{{#> select id="select-single-top-expanded" select-menu-modifier="pf-m-top" select--IsExpanded="true"}}
```
mcoker:
```suggestion
{{#> select id="select-single-top-expanded" select-menu--modifier="pf-m-top" select--IsExpanded="true"}}
```

src/patternfly/components/Select/select-single.hbs
```diff
@@ -1,4 +1,4 @@
-<ul class="pf-c-select__menu{{#if select--modifier}} {{select--modifier}}{{/if}}"
+<ul class="pf-c-select__menu{{#if select-menu-modifier}} {{select-menu-modifier}}{{/if}}"
```
mcoker:
```suggestion
<ul class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
```

src/patternfly/components/Select/select-typeahead.hbs
```diff
@@ -1,4 +1,4 @@
-<ul class="pf-c-select__menu{{#if select--modifier}} {{select--modifier}}{{/if}}"
+<ul class="pf-c-select__menu{{#if select-menu-modifier}} {{select-menu-modifier}}{{/if}}"
```
mcoker:
```suggestion
<ul class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
```

src/patternfly/components/Select/select.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="pf-c-select{{#if select--IsExpanded}} pf-m-expanded{{/if}}{{#if select--modifier}} {{select--modifier}}{{/if}}"
+<div class="pf-c-select{{#if select--IsExpanded}} pf-m-expanded{{/if}}{{#if select-menu-modifier}} {{select-menu-modifier}}{{/if}}"
```
mcoker:
```suggestion
<div class="pf-c-select{{#if select--IsExpanded}} pf-m-expanded{{/if}}{{#if select--modifier}} {{select--modifier}}{{/if}}"
```

mcoker: Since this is the `select` component/hbs file, we want to be able to pass `select--modifier` to it to pass css classes.

src/patternfly/components/Select/select-typeahead.hbs
mcoker: the format for these names are `hbs-file-name--modifier` (for css classes) and `hbs-file-name--attribute` (for html attributes)

src/patternfly/components/Select/examples/select-single-top-expanded-example.hbs
boaz0: shouldn't it be `select--modifier`?

mcoker: @boaz0 oh nope, since we're adding `.pf-m-top` to `.pf-m-select__menu`, not `.pf-c-select`.

src/patternfly/components/Select/examples/index.js
```diff
@@ -83,6 +86,13 @@ export default props => {
       <Example heading="Select - single (expanded)" handlebars={SelectSingleExpandedExampleRaw} minHeight="20em">
         {selectSingleExpandedExample}
       </Example>
+      <Example
+        className="is-expanded-top"
+        heading="Select - single (top expanded)"
+        handlebars={SelectSingleTopExpandedExampleRaw}
```
mcoker: need to add `import SelectSingleTopExampleDoc from '../docs/select-single-top.md';` toward the top of this file, then `docs={SelectSingleTopExampleDoc}` here in this `<Example>` to hook up the documentation.

boaz0: Thanks!

boaz0:
Thank you @mcoker !!!

PR #2028, opened 2019-07-08
feat(select): add a disable modifier

### LINKED ISSUE #1985: PF4 Select: Add the ability to  disable the component
The consumer should be able to disable the Select component similar to the Dropdown.

boaz0:
closes #1985 

//cc @tlabaj

### DIFF
--- a/src/patternfly/components/Select/docs/select-single.md
+++ b/src/patternfly/components/Select/docs/select-single.md
@@ -7,6 +7,7 @@ The single select should be used when the user is selecting an option from a lis
 | Attribute | Applied to | Outcome |
 | -- | -- | -- |
 | `role` or `aria` | `pf-c-select` |  accessibility notes. |
+| `disabled` | `.pf-c-select__toggle` | Disables the dropdown toggle and removes it from keyboard focus. |
 *Note:* The attribute `aria-pressed="true"` should be set programmatically to the active item.
 
 
@@ -23,3 +24,4 @@ The single select should be used when the user is selecting an option from a lis
 | `.pf-c-select__menu-item-icon` | `<i>` |  Initiates the selected item icon. |
 | `.pf-m-expanded` | `.pf-c-select` |  Indicates the select is expanded. |
 | `.pf-m-selected` | `.pf-c-select__menu-item` |  Indicates the menu item is selected. |
+| `.pf-m-disabled` | `div.pf-c-select__toggle` | Modifies to display the select toggle as disabled. This applies to `div.pf-c-select__toggle` and should not be used in lieu of the `disabled` attribute on `button.pf-c-select__toggle`. When this is used, `disabled` should also be added to any form elements in `div.pf-c-select__toggle`|
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -22,6 +22,7 @@
   --pf-c-select__toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-select__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-select__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
+  --pf-c-select__toggle--disabled--BackgroundColor: var(--pf-global--disabled-color--300);
   --pf-c-select__toggle--m-plain--BorderColor: transparent;
   --pf-c-select__toggle--m-plain--Color: var(--pf-global--Color--100);
   --pf-c-select__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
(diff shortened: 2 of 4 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -22,9 +22,11 @@
   --pf-c-select__toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-select__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-select__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
+  --pf-c-select__toggle--disabled--BackgroundColor: var(--pf-global--disabled-color--300);
   --pf-c-select__toggle--m-plain--BorderColor: transparent;
   --pf-c-select__toggle--m-plain--Color: var(--pf-global--Color--200);
   --pf-c-select__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
+  --pf-c-select__toggle--m-plain--disabled--Color: var(--pf-global--disabled-color--200);
```
mcoker: This var is unused. I'll comment in the issue and see if we even want to support a plain variation of the select component.

mcoker:
@mcarrano we have support for a "plain" variation of the select, which would be similar to a kebab dropdown, or a sort icon options menu. Looks like it was added in the [original custom select implementation](https://github.com/patternfly/patternfly-next/pull/1096/), but isn't documented or in use anywhere as far as I know. Do we want to support that as a variation in the select, or should we just remove that variation?

mcarrano:
Not sure that I entirely understand the question @mcoker . Are you asking if the Select should support a variation? Not sure what this would be used for because by definition a select list to picking a value to show from a list of values.

mcoker:
@mcarrano Yeah, in our "dropdown" menus (dropdown, options menu, etc), we support a "plain" variation that is like a kebab menu, and in the options menu we support a "plain text" variation where there is simply no border around the select toggle, like we use in pagination.

Screen Shot 2019-07-15 at 10 21 38 AM

Currently we have support for the plain variation of a select (in the code, but not in an example) but not the plain text. I was referring to the plain variation in this PR. I'm assuming we don't want to support a plain select component, but we do want to add support for the plain text variation?

mcarrano:
I see. So this would be something that works like a dropdown but without the border. Do I understand that right @mcoker ? I'm struggling to think of a use case for that. If we want a text label, wouldn't we just use the standard dropdown/select? Can you think of a unique use case that requires this? I don't have a problem keeping this as an undocumented feature if it's there. If I'm still misunderstanding, let's discuss.

src/patternfly/components/Select/select.scss
mcoker: Hey @boaz0 we do want to support this variation, but you can remove this var. The disabled plain select should display the same as the regular disabled select.

src/patternfly/components/Select/select.scss
```diff
@@ -144,6 +146,19 @@
   background-color: var(--pf-c-select__toggle--BackgroundColor);
   border: none;
 
+  &.pf-m-disabled,
+  &:disabled {
+    pointer-events: none;
+
+    &:not(.pf-m-plain) {
```
mcoker: you can remove this selector, too. we want the plain variation to get the background color and `border: 0`

src/patternfly/components/Select/docs/select-single.md
```diff
@@ -7,6 +7,7 @@ The single select should be used when the user is selecting an option from a lis
 | Attribute | Applied to | Outcome |
 | -- | -- | -- |
 | `role` or `aria` | `pf-c-select` |  accessibility notes. |
+| `disabled` | `.pf-c-select__toggle` | Disables the dropdown toggle and removes it from keyboard focus. |
```
christiemolloy: I think you also want to add in the modifier class for disabled.

src/patternfly/components/Select/select.scss
```diff
@@ -144,6 +145,17 @@
   background-color: var(--pf-c-select__toggle--BackgroundColor);
   border: none;
 
+  &.pf-m-disabled,
+  &:disabled {
+    pointer-events: none;
+
+    --pf-c-select__toggle--BackgroundColor: var(--pf-c-select__toggle--disabled--BackgroundColor);
```
mcoker: Can you move the variable to the top? We typically list variables before declarations.

mcoker:
@boaz0 can you make sure you have the latest from master? Still doesn't look like this pulled in the select plain example.

boaz0:
@mcoker I fetched the latest updates. Can you make sure I didn't miss anything?
Thanks.

boaz0:
@mcoker done!

mcoker:
@boaz0 can you pull the latest from master and rebase? That should fix the failing build.

boaz0:
Yes sure!

boaz0:
Thanks a lot @mcoker for your help, thanks @christiemolloy for your feedback and thanks @redallen for merging :smile_cat:

PR #2030, opened 2019-07-09
fix(components): reverse max-width to be min-width

### LINKED ISSUE #1923: Investigate components using max-width
Related to https://github.com/patternfly/patternfly-next/pull/1918#pullrequestreview-248905813

The PR above raised a good point about max-widths and min-widths, and when both are used at the same breakpoint, breakpoint collisions will arise.

Can we investigate the components that use max-width and see if we can reverse them to be min-width, without causing a breaking change. If they are breaking changes then we can can do it closer to the next breaking change release period.

@srambach to investigate if its possible to make the table component mobile-first so that it displays as a grid and then at a table at a min-width size.

christiemolloy:
closes #1923 

- I didn't update the table yet, because wasn't sure if @srambach had started yet or not
- I also didn't update the last max-width in the page component, the problem here is that if we default to mobile first, then we need to make the default background color in the nav component the color of the header on mobile. I think im fine with making the change but I wanted to see what others thought.

```
    .pf-c-nav__scroll-button {
      @media screen and (max-width: #{$pf-global--breakpoint--lg - 1}) {
        top: var(--pf-c-page__header-nav--c-nav__scroll-button--lg--Top);
        background-color: var(--pf-c-page__header-nav--c-nav__scroll-button--lg--BackgroundColor);
      }
```


src/patternfly/components/DataList/data-list.scss
```diff
@@ -90,18 +92,19 @@
   --pf-c-data-list__expandable-content--MarginRight: calc(var(--pf-c-data-list__expandable-content-body--PaddingRight) * -1);
   --pf-c-data-list__expandable-content--MarginLeft: calc(var(--pf-c-data-list__expandable-content-body--PaddingLeft) * -1);
   --pf-c-data-list__expandable-content--BoxShadow: var(--pf-global--BoxShadow--md-bottom);
-  --pf-c-data-list__expandable-content--md--MaxHeight: #{pf-size-prem(600px)};
+  --pf-c-data-list__expandable-content--MaxHeight: #{pf-size-prem(600px)};
+  --pf-c-data-list__expandable-content--md--MaxHeight: unset;
```
mcoker: Can you use `initial` or `none` here?

src/patternfly/components/Login/login.scss
```diff
@@ -17,21 +17,33 @@
 
   // header
   --pf-c-login__header--MarginBottom: var(--pf-global--spacer--md);
-  --pf-c-login__header--sm--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-login__header--sm--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-login__header--PaddingLeft: var(--pf-global--spacer--md);
```
mcoker: Just noting, this is an example of 6 lines of code turning into 12. There are a couple of those in this component.

mcoker:
@christiemolloy can we close this in favor of https://github.com/patternfly/patternfly/pull/2816?

PR #2032, opened 2019-07-10
fix(login): fix a11y

### LINKED ISSUE #1907: Login: Elements must have sufficient color contrast
This is one of only **two** remaining color contrast issues! It's for login, details from the a11y audit below. The other error in the screenshot is already captured in a separate issue.

Screen Shot 2019-06-06 at 9 20 35 PM

It's basically those three links Terms of Use, Help, and Privacy Policy. The axe browser extension is flagging issues with other text on the page also, but I'm not as worried about that as I am getting the automated tooling to a point where we are free of errors. 

Screen Shot 2019-06-06 at 9 24 10 PM

christiemolloy:
fixes #1907 

unsets the link color text so that its white. Looks like even when its blue at a max-width of "xl" it still doesnt pass the color-contrast test, so I made it white at all sizes @mceledonia is this ok?

### DIFF
--- a/src/patternfly/components/Login/login.scss
+++ b/src/patternfly/components/Login/login.scss
@@ -282,6 +282,12 @@
   }
 
   .pf-c-list {
+    a {
+      @media screen and (min-width: $pf-global--breakpoint--xl) {
+        color: unset;
+      }
+    }
+
     &:not(:only-child) { // paragraph text might not always be included.
       padding-top: var(--pf-c-login__footer--c-list--PaddingTop);
     }
--- a/src/patternfly/components/Login/login.scss
+++ b/src/patternfly/components/Login/login.scss
@@ -283,9 +283,11 @@
 
   .pf-c-list {
     a {
-      @media screen and (min-width: $pf-global--breakpoint--xl) {
-        color: unset;
-      }
+      color: unset;
+
+      // @media screen and (min-width: $pf-global--breakpoint--xl) {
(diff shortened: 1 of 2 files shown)


mcoker:
FWIW, these links used to be white, then we made them the normal link color in https://github.com/patternfly/patternfly-next/pull/1281/files#diff-85d2441dd6ceb0f491290b7c882e0547L72

christiemolloy:
@mcoker Are you suggesting that we add that variable back in rather than "unsetting" the color?

mcoker:
If we want the links to be white, I think we should. But I think the reason we changed them from white to blue initially is because it isn't clear that they're links when they're white.

christiemolloy:
From what I can see, to pass a11y violation, the background either needs to be a solid color, or the links need to be white, unless @mceledonia you want to try test another color -- its just hard because the background behind the links changes at every size.

christiemolloy:
@mceledonia do you have any input here. Basically to remove a11y violation we need to change the background gradient, or change the color of the links to white.

mceledonia:
This shouldn't be an issue with our new login screens with the red hat logo in the background... @kybaker should we move forward with the assumption we will have that (and avoid the a11y issue) or find another fix for this in the short term? I'm ok with the white for now as a temporary measure until we get the new login.

kybaker:
@mceledonia I agree - I think the white is a great temporary solution.

PR #2033, opened 2019-07-10
fix(tooltip): break words

### LINKED ISSUE #2013: Long texts in tooltip are truncated
When there is long text in tooltip it is runcated and not visible. Can we add `word-break: break-all;` in there?

christiemolloy:
closes #2013 

cc: @karelhala 

Screen Shot 2019-07-10 at 11 05 46 AM


mcoker:
`break-all` will break anywhere in a word, even when it isn't necessary.

Screen Shot 2019-07-10 at 10 12 33 AM

`break-word` would work better.

Screen Shot 2019-07-10 at 10 13 03 AM

PR #2036, opened 2019-07-10
fix(select): add documentation for pf-m-plain

### LINKED ISSUE #1987: PF4 Select: add example and documentation for pf-m-plain
There is a request to have a pf-m-plain select, similar to the options menu and dropdown. Lets add an example and also add the documentation for the pf-m-plain modifier for the button is not documented in usage

jenny-s51:
Closes #1987


christiemolloy:
@mcoker I agree with adding an example of the plain variation, @jenny-s51 do you want to add that as part of this issue?

jenny-s51:
@christiemolloy I think we should make another PR for that -- I'm booked with react issues for this sprint! Dana told me to make this PR so I could start on this issue https://github.com/patternfly/patternfly-react/issues/2371 which is what I'm working on currently 🙂

jenny-s51:
Can this PR be merged?

christiemolloy:
@jenny-s51 , I talked to @mcoker and we shouldn't add the documentation when we don't have an example to support it. I'm going to update the original issue and we can close this PR

PR #2049, opened 2019-07-12
fix(switch): remove dependency on font-size for switch's height

### LINKED ISSUE #2025: The pf-c-switch renders awkwardly if --pf-global--FontSize--md variable is changed from 1rem
We noticed in the OpenShift console that when changing the `--pf-global--FontSize--md` value caused  the `.pf-c-switch` to not render correctly, since it uses that value to calculat  `--pf-c-switch--Height`. 

https://github.com/patternfly/patternfly-next/blob/4a2dd6b099c3b1625dd1a37992cb6f7fbb83a8e3/src/patternfly/components/Switch/switch.scss#L3

christiemolloy:
closes #2025


src/patternfly/components/Switch/switch.scss
```diff
@@ -1,7 +1,7 @@
 .pf-c-switch {
   // Switch
-  --pf-c-switch--Height: calc(#{var(--pf-global--FontSize--md)} * #{var(--pf-global--LineHeight--md)});
-  --pf-c-switch--Width: calc(#{var(--pf-c-switch--Height)} * #{var(--pf-global--golden-ratio)});
+  --pf-c-switch--Height: calc(#{var(--pf-c-switch__label--FontSize)} * #{var(--pf-c-switch--LineHeight)});
+  --pf-c-switch--Width: calc(#{var(--pf-c-switch--Height)} + #{var(--pf-c-switch__label--FontSize)});
```
mcoker: Do you think we should use `--pf-c-switch__toggle--before--Width` instead of `--pf-c-switch__label--FontSize` here? They're the same thing at the moment, but that's really just to create space for the other toggle circle, which would be represented by `--pf-c-switch__toggle--before--Width`

src/patternfly/components/Switch/switch.scss
```diff
@@ -31,13 +31,14 @@
   --pf-c-switch__toggle-icon--FontSize: var(--pf-global--icon--FontSize--sm);
   --pf-c-switch__toggle-icon--Color: var(--pf-global--Color--light-100);
   --pf-c-switch__toggle-icon--PaddingLeft: var(--pf-global--spacer--sm);
-  --pf-c-switch__toggle-icon--Top: #{pf-size-prem(7px)};
+  --pf-c-c-switch__toggle-icon--LineHeight: var(--pf-c-switch--Height);
```
mcoker: Looks like this can be removed.

src/patternfly/components/Switch/switch.scss
```diff
@@ -139,9 +140,9 @@
 
   .pf-c-switch__toggle-icon {
     position: absolute;
-    top: var(--pf-c-switch__toggle-icon--Top);
```
mcoker: If you're going to remove `top`, there's no need to leave the variable around since it would not do anything.

Instead of `line-height` you could use `top: 50%; left: 0; transform: translateY(50%);` to center it. I think that would be more reliable.

Also looks like we need some more space between this icon and the circle. 

I checked out your branch to work on this a bit and created a PR - https://github.com/christiemolloy/patternfly-next/pull/17

Made some of these changes, kept this `top` and removed `line-height` below. Also reduced the size of the toggle icon - it's 14px on master currently, so we can keep it that width. Only thing I'm not sure about is I hardcoded a `2px` spacer into the switch `width` and toggle circle `translateX` to match what's on master now and create space between the toggle icon (check) and toggle circle. Feel free to update that if you can think of a better way.

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -136,9 +136,6 @@ $pf-global--spacer--form-element: pf-size-prem(6px) !default;
 $pf-global--gutter: $pf-global--spacer--lg !default;
 $pf-global--gutter--md: $pf-global--spacer--md !default;
 
-// Ratio
-$pf-global--golden-ratio: 1.681 !default;
```
mcoker: Removing this is a breaking change, too. Let's add it back and leave a comment in the code to remove it later or create an issue or however you'd prefer deal with it.

src/patternfly/components/Switch/switch.scss
christiemolloy: yes good idea!

src/patternfly/sass-utilities/scss-variables.scss
christiemolloy: Okay added them back in captured it in an issue here: #2054

src/patternfly/components/Switch/switch.scss
christiemolloy: Thanks for adding that branch, I also updated some things.
- for the toggle::before width I subtracted 2px because the current toggle was 2px too big from what was on master:

Screen Shot 2019-07-16 at 9 46 56 AM
Screen Shot 2019-07-16 at 9 47 01 AM

- removed the var you added for transform because it looks like it isn't needed
- i think we can remove the font-weight and line-height at the next breaking change
- since we use #{pf-size-prem(2px)} 3 times now, I wonder if we should store it in a var like "offset" or something

src/patternfly/components/Switch/switch.scss
```diff
@@ -1,12 +1,12 @@
 .pf-c-switch {
   // Switch
-  --pf-c-switch--Height: calc(#{var(--pf-global--FontSize--md)} * #{var(--pf-global--LineHeight--md)});
-  --pf-c-switch--Width: calc(#{var(--pf-c-switch--Height)} * #{var(--pf-global--golden-ratio)});
+  --pf-c-switch--Height: calc(var(--pf-c-switch__label--FontSize) * var(--pf-c-switch--LineHeight));
+  --pf-c-switch--Width: calc(var(--pf-c-switch--Height) + #{pf-size-prem(2px)} + var(--pf-c-switch__toggle--before--Width));
   --pf-c-switch--LineHeight: var(--pf-global--LineHeight--md);
```
mcoker: Curious why we have `line-height` on `.pf-c-switch` and `font-size` on `.pf-c-switch__label`. Can we move them both to the same element, or is there a reason to keep them separate?

src/patternfly/components/Switch/switch.scss
```diff
@@ -19,25 +19,28 @@
   // Switch toggle
   --pf-c-switch__toggle--BackgroundColor: var(--pf-global--disabled-color--200);
   --pf-c-switch__toggle--BorderRadius: var(--pf-c-switch--Height);
-  --pf-c-switch__toggle--before--Top: calc(#{var(--pf-c-switch--Height)} / 2 * (1 - 1 / #{var(--pf-global--golden-ratio)}));
+  --pf-c-switch__toggle--before--Top: calc((var(--pf-c-switch--Height) - var(--pf-c-switch__toggle--before--Height)) / 2);
   --pf-c-switch__toggle--before--Left: var(--pf-c-switch__toggle--before--Top);
-  --pf-c-switch__toggle--before--Width: calc(#{var(--pf-c-switch--Height)} / #{var(--pf-global--golden-ratio)});
+  --pf-c-switch__toggle--before--Width: calc(var(--pf-c-switch__label--FontSize) - #{pf-size-prem(2px)});
```
mcoker: It's worth noting that `--pf-global--FontSize--sm` would work here, too. That's what I had in the original PR, and it's the same value used for `--pf-c-switch__toggle-icon--FontSize`. Though I think this is better since it resizes the toggle relative to the font size used in the switch instead of needing to be adjusted manually if you change the switch font size. That said, I think we should update the `--pf-c-switch__toggle-icon--FontSize` var, too, since it's the only thing that doesn't scale when you change the switch font size.

src/patternfly/components/Switch/switch.scss
mcoker: > since we use #{pf-size-prem(2px)} 3 times now, I wonder if we should store it in a var like "offset" or something

Works for me!

src/patternfly/components/Switch/switch.scss
christiemolloy: Great point, moved the font-size from the label to the switch

christiemolloy: I did have to remove the font-size variable from the label, because now if the font-size changes the say 24px, the var set on the label overrides it to still be 16px. Is this a breaking change?

src/patternfly/components/Switch/switch.scss
christiemolloy: Updated the font-size of the icon to be relative to the switch's font-size. However, we might also need to make the padding-left on the icon, relative to the icon size too..

Screen Shot 2019-07-17 at 4 10 14 PM

christiemolloy: I tested the value 0.4 and it seems to be consistent across all font-sizes in the switch:

Screen Shot 2019-07-17 at 4 16 08 PM
Screen Shot 2019-07-17 at 4 16 26 PM
Screen Shot 2019-07-17 at 4 17 58 PM
Screen Shot 2019-07-17 at 4 18 16 PM
Screen Shot 2019-07-17 at 4 18 37 PM
Screen Shot 2019-07-17 at 4 18 58 PM

src/patternfly/components/Switch/switch.scss
```diff
@@ -1,12 +1,13 @@
 .pf-c-switch {
   // Switch
-  --pf-c-switch--Height: calc(var(--pf-c-switch__label--FontSize) * var(--pf-c-switch--LineHeight));
-  --pf-c-switch--Width: calc(var(--pf-c-switch--Height) + #{pf-size-prem(2px)} + var(--pf-c-switch__toggle--before--Width));
+  --pf-c-switch--Height: calc(var(--pf-c-switch--FontSize) * var(--pf-c-switch--LineHeight));
+  --pf-c-switch--Width: calc(var(--pf-c-switch--Height) + var(--pf-c-switch__toggle-icon--Offset) + var(--pf-c-switch__toggle--before--Width));
   --pf-c-switch--LineHeight: var(--pf-global--LineHeight--md);
+  --pf-c-switch--FontSize: var(--pf-global--FontSize--sm);
```
mcoker: Should this be `--pf-global--FontSize--sm`? Everything is too small by 2px now :)

src/patternfly/components/Switch/switch.scss
```diff
@@ -21,31 +22,33 @@
...
+  --pf-c-switch__toggle--before--Width: calc(var(--pf-c-switch--FontSize) - var(--pf-c-switch__toggle-icon--Offset));
   --pf-c-switch__toggle--before--Height: var(--pf-c-switch__toggle--before--Width);
   --pf-c-switch__toggle--before--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-switch__toggle--before--BorderRadius: var(--pf-global--BorderRadius--lg);
   --pf-c-switch__toggle--before--BoxShadow: var(--pf-global--BoxShadow--md);
   --pf-c-switch__toggle--before--Transition: transform .25s ease 0s, width .25s ease 0s, left .25s ease 0s;
 
   // Switch icon
-  --pf-c-switch__toggle-icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-switch__toggle-icon--FontSize: calc(var(--pf-c-switch--FontSize) * .625);
```
mcoker: Since we don't normally do this, can you add a comment here explaining why we don't use an `--icon--FontSize` var b/c this allows the icon to scale?

src/patternfly/components/Switch/switch.scss
```diff
@@ -21,31 +22,33 @@
...
   --pf-c-switch__toggle-icon--Color: var(--pf-global--Color--light-100);
-  --pf-c-switch__toggle-icon--PaddingLeft: var(--pf-global--spacer--sm); // during breaking change period, remove this and make it the value for `left`
+  --pf-c-switch__toggle-icon--PaddingLeft: calc(var(--pf-c-switch--FontSize) * .4); // during breaking change period, remove this and make it the value for `left`
```
mcoker: Same here, can you add a comment saying something like this allows it to scale relative to the switch and icon font sizes?

src/patternfly/components/Switch/switch.scss
christiemolloy: yep forgot to change it back to md

src/patternfly/components/Switch/switch.scss
mcoker: I think it's fine, but could you add the var and declaration back and just make the value `--pf-c-switch__label--FontSize: var(--pf-c-switch--FontSize)`?

src/patternfly/components/Switch/switch.scss
```diff
@@ -19,30 +20,35 @@
...
-  --pf-c-switch__toggle--before--Top: calc(#{var(--pf-c-switch--Height)} / 2 * (1 - 1 / #{var(--pf-global--golden-ratio)}));
+  --pf-c-switch__toggle--before--Top: calc((var(--pf-c-switch--Height) - var(--pf-c-switch__toggle--before--Height)) / 2);
   --pf-c-switch__toggle--before--Left: var(--pf-c-switch__toggle--before--Top);
-  --pf-c-switch__toggle--before--Width: calc(#{var(--pf-c-switch--Height)} / #{var(--pf-global--golden-ratio)});
+  --pf-c-switch__toggle--before--Width: calc(var(--pf-c-switch--FontSize) - var(--pf-c-switch__toggle-icon--Offset));
   --pf-c-switch__toggle--before--Height: var(--pf-c-switch__toggle--before--Width);
   --pf-c-switch__toggle--before--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-switch__toggle--before--BorderRadius: var(--pf-global--BorderRadius--lg);
   --pf-c-switch__toggle--before--BoxShadow: var(--pf-global--BoxShadow--md);
+  --pf-c-switch__toggle--before--Transition: transform .25s ease 0s, width .25s ease 0s, left .25s ease 0s;
```
mcoker: why transition `width` and `left?`

christiemolloy: not sure, removed, its not needed

PR #2053, opened 2019-07-15
feat(select): add plain variation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1987

### DIFF
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -1,6 +1,4 @@
 .pf-c-select {
-  @include pf-t-light; // force the container follow the light theme
-
   // Toggle
   --pf-c-select__toggle--PaddingTop: var(--pf-global--spacer--form-element);
   --pf-c-select__toggle--PaddingRight: var(--pf-global--spacer--sm);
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -23,8 +21,6 @@
   --pf-c-select__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-select__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-select__toggle--m-plain--BorderColor: transparent;
-  --pf-c-select__toggle--m-plain--Color: var(--pf-global--Color--200);
-  --pf-c-select__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
```
christiemolloy: Are these two lines breaking changes

src/patternfly/components/Select/select.scss
```diff
@@ -127,6 +123,11 @@
   width: 100%;
 }
 
+.pf-c-select__toggle:not(.pf-m-plain),
+.pf-c-select__menu {
+  @include pf-t-light; // other than a plain select toggle, the select component should be a light theme
```
christiemolloy: Was there a way to disable the dark-mode icon in pf so that if its only meant to be light, the consumer can't see it like this:

Screen Shot 2019-07-15 at 4 37 45 PM

src/patternfly/components/Select/select.scss
mcoker: Hmm... technically yes, but it was never a version of the component that we supported. It wouldn't make sense to have the "plain" code we have here for the select. I'm fine with deleting it but we could be safe and remove it later. What do you think?

src/patternfly/components/Select/select.scss
mcoker: Ah good catch, I forgot to test that. I changed the theme since I assumed the plain variation would have a dark background and white text in a dark theme, but assuming the select will be used in a form with other form controls, I suppose this variation should always follow the light theme, too. Updated the PR.

christiemolloy: yep good idea until we decide that it should be used in dark mode

src/patternfly/components/Select/select.scss
christiemolloy: I think be safer and remove it later. Just in case someone is using one of these variables, even though we aren't using them

src/patternfly/components/Select/select.scss
```diff
@@ -1,5 +1,5 @@
 .pf-c-select {
-  @include pf-t-light; // force the container follow the light theme
+  @include pf-t-light; // other than a plain select toggle, the select component should be a light theme
```
christiemolloy: I think you can update the comment here now that you added this back

PR #2058, opened 2019-07-16
chore(a11y): ensure SkipToContent sends focus to page content

seanforyou23:
This PR swaps the target element for SkipToContent from being an anchor to being the primary page container. It requires that we supply the `<main>` element with `tabindex="-1" so this region can receive focus programmatically. It's related to https://github.com/patternfly/patternfly-react/pull/2519

I did notice that unless a heading is an immediate child of the main container, we still don't get the headings read aloud for free when using a screen reader, however, I believe this is still a step in the right direction. The technique is described in one of the training courses over at dequeuniversity.com.

Should we do something about the new focus ring that appears as a result of focusing on the main content region?

add doc for new attribute requirements in Page, SkipToContent
update corresponding demos/exmaples
skip to content sends focus to content container

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -369,6 +369,10 @@
   overflow-x: hidden;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
+
+  &:target {
+    outline: 0;
+  }
 }
 
 .pf-c-page__main-nav {
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -370,7 +370,7 @@
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
 
-  &:target {
+  &:focus {
     outline: 0;
   }
 }
--- a/src/patternfly/components/SkipToContent/docs/code.md
+++ b/src/patternfly/components/SkipToContent/docs/code.md
@@ -2,7 +2,7 @@
 
 Skip to content allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
 
-When using `.pf-c-skip-to-content` you must provide an `href` attribute who's value corresponds to the `id` attribute of the primary content container for your application. In most cases this is the `<main>` element. For a demo of this see the [page demo](../../../demos/Page/examples), and note the use of `tabindex="-1" which allows the element to receive focus programmatically.
+When using `.pf-c-skip-to-content` you must provide an `href` attribute whose value corresponds to the `id` attribute of the primary content container for your application. In most cases this is the `<main>` element. For a demo of this see the [page demo](../../../demos/Page/examples), and note the use of `tabindex="-1"` which allows the element to receive focus programmatically.
 
 ## Accessibility
 
--- a/src/patternfly/components/SkipToContent/docs/code.md
+++ b/src/patternfly/components/SkipToContent/docs/code.md
@@ -2,7 +2,13 @@
 
 Skip to content allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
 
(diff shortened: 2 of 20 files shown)


matthewcarleton:
thanks for doing this @seanforyou23 
I think dropping the focus ring could make sense. It might be a little strange for sighted user. Although it's pretty subtle so not sure it's a big deal.

src/patternfly/components/SkipToContent/docs/code.md
```diff
@@ -2,7 +2,13 @@
 
 Skip to content allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
 
-When using `.pf-c-skip-to-content` you must also provide an anchor with the matching id that is applied to the `href` attribute of `.pf-c-skip-to-content`. You will typically place this just before the main content of the page. Ex: `<a id="main-content"></a>` For a demo of this see the [page demo](../../../demos/Page/examples).
+When using `.pf-c-skip-to-content` you must provide an `href` attribute who's value corresponds to the `id` attribute of the primary content container for your application. In most cases this is the `<main>` element. For a demo of this see the [page demo](../../../demos/Page/examples), and note the use of `tabindex="-1" which allows the element to receive focus programmatically.
```
mcoker:
```suggestion
When using `.pf-c-skip-to-content` you must provide an `href` attribute whose value corresponds to the `id` attribute of the primary content container for your application. In most cases this is the `<main>` element. For a demo of this see the [page demo](../../../demos/Page/examples), and note the use of `tabindex="-1"` which allows the element to receive focus programmatically.
```

mcoker:
There are some other places we'll need to update the anchor link:

```
src//patternfly/demos/DataList/examples/data-list-simple-without-pagination-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/DataList/examples/data-list-simple-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/DataList/examples/data-list-expandable-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/DataList/examples/data-list-actionable-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/Table/examples/table-sortable-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/Table/examples/table-compact-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/Table/examples/table-simple-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/Table/examples/table-expandable-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/Table/examples/table-compound-expansion-demo.hbs:    <a id="main-content-{{page--id}}"></a>
src//patternfly/demos/CardView/examples/card-view-demo-example.hbs:      <a id="main-content-{{page--id}}"></a>
```
```

**Comment by mcoker** [comment]
> I think dropping the focus ring could make sense. It might be a little strange for sighted user.

+1. You can just add `&:target { outline: 0; }` in [this block](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Page/page.scss#L364-L372).

File: src/patternfly/components/Page/page.scss [hunk 2]
```diff
@@ -369,6 +369,10 @@
   overflow-x: hidden;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
+
+  &:target {
```
jgiardino: @seanforyou23 and I are reviewing this, and the question came up about whether we need this or not. After some discussion, we're leaning toward not including `&:target` and just always having `outline: 0;` set for the main region.

seanforyou23: +1 - realized that the style rule doesn't apply when sending focus to the main content area programmatically with JavaScript, which we need to do after route changes to ensure assistive technology is made aware of the new content. If it's ok, I'd like to remove it. WDYT @mcoker ?

mcoker: Sounds good to me!

seanforyou23: Ended up going with `:focus`, this most closely represents the state of the element we're looking to style and signals to developers the purpose of the style. Let me know if you see any issues with this.

PR #2065, opened 2019-07-16
fix(table): add example of modifier

### LINKED ISSUE #2048: Support column headers that wrap
Update: this issue should show an example of the table headers wrapping.

cc: @mcoker  @mcarrano

christiemolloy:
closes #2048 

@KendraMar I added an example of how to make column headers wrap. We should create a new issue if you would like to be able to set max-widths on the columns.

Screen Shot 2019-07-16 at 2 44 21 PM

### DIFF
--- a/src/patternfly/components/Table/docs/table-headers-wrap.md
+++ b/src/patternfly/components/Table/docs/table-headers-wrap.md
@@ -1,0 +1,9 @@
+### Table expandable notes
+
+**All simple table accessibility and usage requirements apply.**
+
+### Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-wrap` | `<th>`, `<td>` | Modifies content to wrap. |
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Table/docs/table-headers-wrap.md
```diff
@@ -0,0 +1,9 @@
+### Table expandable notes
+
+**All simple table accessibility and usage requirements apply.**
+
+### Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-wrap` | `<th>`, `<td>` | Modifies content to wrap. |
```
christiemolloy: @mcoker not sure if its necessary to be on the `<td>` here because the td wraps by default. Only thead has `white-space: nowrap;` set on it. Maybe I'm missing something.

mcoker: it's because `<td>` is valid in a `<thead>`

PR #2069, opened 2019-07-16
feat(drawer): adds drawer component

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1786

refactored this so that it's a component, and it can be used outside of the context of the page component.

### DIFF
--- a/src/patternfly/components/Drawer/docs/code.md
+++ b/src/patternfly/components/Drawer/docs/code.md
@@ -7,5 +7,5 @@
 | `.pf-c-drawer__panel` | `<aside>` | Creates the drawer panel container. |
 | `.pf-c-drawer__panel-body` | `<div>` | Creates the drawer panel body container. |
 | `.pf-m-expanded` | `.pf-c-drawer` | Modifies the drawer panel for the expanded state. |
-| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the content element and panel element are displayed side by side. **This variation is not supported at this time for use in the page component when the drawer is used with the main content area.** |
+| `.pf-m-inline` | `.pf-c-drawer` | Modifies the drawer so the content element and panel element are displayed side by side. **This variation is not supported at this time for use in the page component when the drawer is used with the main content area.** |
 | `.pf-m-no-padding` | `.pf-c-drawer__panel-body` | Modifies the panel body to remove the padding. |
--- a/src/patternfly/components/Drawer/docs/code.md
+++ b/src/patternfly/components/Drawer/docs/code.md
@@ -3,9 +3,9 @@
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-drawer` | `<div>` | Creates the drawer container. |
-| `.pf-c-drawer__main` | `<div>` | Creates the drawer main container. |
+| `.pf-c-drawer__content` | `<div>` | Creates the drawer content container. |
 | `.pf-c-drawer__panel` | `<aside>` | Creates the drawer panel container. |
 | `.pf-c-drawer__panel-body` | `<div>` | Creates the drawer panel body container. |
 | `.pf-m-open` | `.pf-c-drawer` | Modifies the drawer for the open state. |
-| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the main element and panel element are displayed side by side. |
+| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the content element and panel element are displayed side by side. **This variation is not supported at this time for use in the page component when the drawer is used with the content content area.** |
 | `.pf-m-no-padding` | `.pf-c-drawer__panel-body` | Modifies the panel body to remove the padding. |
--- a/src/patternfly/components/Drawer/docs/code.md
+++ b/src/patternfly/components/Drawer/docs/code.md
@@ -6,6 +6,6 @@
 | `.pf-c-drawer__content` | `<div>` | Creates the drawer content container. |
 | `.pf-c-drawer__panel` | `<aside>` | Creates the drawer panel container. |
 | `.pf-c-drawer__panel-body` | `<div>` | Creates the drawer panel body container. |
-| `.pf-m-open` | `.pf-c-drawer` | Modifies the drawer for the open state. |
-| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the content element and panel element are displayed side by side. **This variation is not supported at this time for use in the page component when the drawer is used with the content content area.** |
+| `.pf-m-expanded` | `.pf-c-drawer` | Modifies the drawer panel for the expanded state. |
+| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the content element and panel element are displayed side by side. **This variation is not supported at this time for use in the page component when the drawer is used with the main content area.** |
 | `.pf-m-no-padding` | `.pf-c-drawer__panel-body` | Modifies the panel body to remove the padding. |
--- a/src/patternfly/components/Drawer/docs/code.md
+++ b/src/patternfly/components/Drawer/docs/code.md
@@ -1,0 +1,11 @@
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-drawer` | `<div>` | Creates the drawer container. |
+| `.pf-c-drawer__main` | `<div>` | Creates the drawer main container. |
+| `.pf-c-drawer__panel` | `<aside>` | Creates the drawer panel container. |
+| `.pf-c-drawer__panel-body` | `<div>` | Creates the drawer panel body container. |
+| `.pf-m-open` | `.pf-c-drawer` | Modifies the drawer for the open state. |
+| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the main element and panel element are displayed side by side. |
+| `.pf-m-no-padding` | `.pf-c-drawer__panel-body` | Modifies the panel body to remove the padding. |
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -1,15 +1,15 @@
 .pf-c-drawer {
   --pf-c-drawer__main--MarginRight: calc(var(--pf-c-drawer__panel--Width) * -1);
-  --pf-c-drawer--m-open__main--MarginRight: 0;
-  --pf-c-drawer__main--Transition: margin .25s ease;
+  --pf-c-drawer--m-expanded__main--MarginRight: 0;
+  --pf-c-drawer__main--Transition: margin-right .25s ease;
   --pf-c-drawer__panel--ZIndex: var(--pf-global--ZIndex--sm);
   --pf-c-drawer__panel--Width: 100%;
   --pf-c-drawer__panel--md--Width: #{pf-size-prem(450px)};
   --pf-c-drawer__panel--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-drawer__panel--Transition: transform .25s ease;
   --pf-c-drawer__panel--Transform: translateX(100%);
-  --pf-c-drawer--m-open__panel--Transform: translateX(0);
-  --pf-c-drawer--m-open__panel--BoxShadow: var(--pf-global--BoxShadow--md-left);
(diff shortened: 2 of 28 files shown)


src/patternfly/components/Drawer/docs/code.md
```diff
@@ -0,0 +1,11 @@
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-drawer` | `<div>` | Creates the drawer container. |
+| `.pf-c-drawer__main` | `<div>` | Creates the drawer main container. |
+| `.pf-c-drawer__panel` | `<aside>` | Creates the drawer panel container. |
+| `.pf-c-drawer__panel-body` | `<div>` | Creates the drawer panel body container. |
+| `.pf-m-open` | `.pf-c-drawer` | Modifies the drawer for the open state. |
```
christiemolloy: Why did you switch `pf-m-expanded` to `pf-m-open`? I liked expanded because its consistent with the naming for many of our other modifiers.

src/patternfly/components/Page/docs/code.md
```diff
@@ -30,6 +30,7 @@ This component provides the basic chrome for a page, including sidebar, header,
 | `.pf-c-page__main-nav` | `<section>` |   Creates a container to nest the navigation component in the main page area. |
 | `.pf-c-page__main-breadcrumb` | `<section>` |   Creates a container to nest the breadcrumb component in the main page area. |
 | `.pf-c-page__main-section` | `<section>` |  Creates a section container in the main page area. **Note: The last/only `.pf-c-page__main-section` element will grow to fill the availble vertical space. You can change this behavior using `.pf-m-fill` and `.pf-m-no-fill`, which are documented below.**  |
+| `.pf-c-page__drawer` | `<div>` |  Creates a container for the drawer component when placing the main page element in the drawer body. |
```
christiemolloy: Is it required that the main page element is included inside of here? I don't see it in the second example of the drawer demo.

src/patternfly/components/Drawer/docs/code.md
```diff
@@ -0,0 +1,11 @@
...
+| `.pf-m-inline` | `.pf-c-drawer` | Modifies drawer so the main element and panel element are displayed side by side. |
```
christiemolloy: I don't know if `inline` works here for the modifier name. Or maybe you need to update the description, because its more obvious in the demo then the component. I also wonder if these "inline" styles should be in the page component since they modify the drawer in the page component specifically.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -0,0 +1,70 @@
...
+  --pf-c-drawer__main--MarginRight: calc(var(--pf-c-drawer__panel--Width) * -1);
+  --pf-c-drawer--m-open__main--MarginRight: 0;
+  --pf-c-drawer__main--Transition: margin .25s ease;
+  --pf-c-drawer__panel--ZIndex: var(--pf-global--ZIndex--sm);
+  --pf-c-drawer__panel--Width: 100%;
+  --pf-c-drawer__panel--md--Width: #{pf-size-prem(450px)};
+  --pf-c-drawer__panel--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-drawer__panel--Transition: transform .25s ease;
+  --pf-c-drawer__panel--Transform: translateX(100%);
+  --pf-c-drawer--m-open__panel--Transform: translateX(0);
```
christiemolloy: should all the `--pf-c-drawer--m-open..` go before anything else in this list

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -0,0 +1,70 @@
...
+  --pf-c-drawer--m-open__panel--BoxShadow: var(--pf-global--BoxShadow--md-left);
+  --pf-c-drawer__panel-body--PaddingTop: var(--pf-global--spacer--md);
```
christiemolloy: At this size, this is happening

Screen Shot 2019-07-17 at 2 49 52 PM

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -0,0 +1,70 @@
+.pf-c-drawer {
+  --pf-c-drawer__main--MarginRight: calc(var(--pf-c-drawer__panel--Width) * -1);
+  --pf-c-drawer--m-open__main--MarginRight: 0;
+  --pf-c-drawer__main--Transition: margin .25s ease;
+  --pf-c-drawer__panel--ZIndex: var(--pf-global--ZIndex--sm);
+  --pf-c-drawer__panel--Width: 100%;
+  --pf-c-drawer__panel--md--Width: #{pf-size-prem(450px)};
+  --pf-c-drawer__panel--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-drawer__panel--Transition: transform .25s ease;
```
christiemolloy: Why do `--pf-c-drawer__main` and `--pf-c-drawer__panel` both have a transition?

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  display: flex;
+  overflow-x: hidden;
+
+  &.pf-m-open {
+    --pf-c-drawer__panel--Transform: var(--pf-c-drawer--m-open__panel--Transform);
+
+    &.pf-m-inline {
+      --pf-c-drawer__main--MarginRight: var(--pf-c-drawer--m-open__main--MarginRight);
```
christiemolloy: How come you override `--pf-c-drawer__main--MarginRight` in `pf-c-drawer` rather than `pf-c-drawer__main`, also I don't see these styles for margin-right working could you explain what they do?

src/patternfly/demos/Drawer/drawer-demo-default.hbs
```diff
@@ -0,0 +1,71 @@
...
+          {{#> nav-link nav-link--href="#"}}
+            Server
+          {{/nav-link}}
+        {{/nav-item}}
+      {{/nav-list}}
+    {{/nav}}
+  {{/page-sidebar}}
+  {{#> page-drawer}}
+    {{#> drawer}}
+      {{#> drawer-main}}
```
christiemolloy: Having the `page-main section` inside of the page-drawer doesn't seem intuitive to me.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -0,0 +1,70 @@
+.pf-c-drawer {
```
christiemolloy: when the panel is closed, should it still be visible?

Screen Shot 2019-07-17 at 3 15 20 PM

src/patternfly/components/Drawer/docs/code.md
mcoker: `inline` just displays the drawer inline with the content instead of on top of it. I'm open to another name if you want to suggest one. But they aren't just used in the page component. In fact, we don't support them in the page component at the moment. They would be supported in other contexts though, like [topology](https://patternfly-react.surge.sh/patternfly-4/components/topology/).

src/patternfly/components/Drawer/drawer.scss
mcoker: It can, I don't feel strongly about it. I grouped it by the value the var modifies, not the selector, so that the modifier var is after the thing it modifies. Same with

```
  --pf-c-drawer__main--MarginRight: calc(var(--pf-c-drawer__panel--Width) * -1);
  --pf-c-drawer--m-expanded__main--MarginRight: 0;
```

Want me to re-arrange them?

src/patternfly/components/Drawer/drawer.scss
mcoker: `__main` transitions when `pf-c-drawer` has `.pf-m-inline`. I can limit the `transition` to when the drawer has that class if you think this will cause problems. Only problem I see is if someone dynamically changes the `margin-right` (I changed the transition from `margin` to `margin-right`) outside of the context of the drawer opening/closing, that margin change will be transitioned. Moving the transition to `.pf-m-inline` will still cause that, but only when the drawer also has `.pf-m-inline`

src/patternfly/components/Drawer/drawer.scss
mcoker: one of the benefits of using vars like this is that we can change a subset of children vars on a parent like this. I'm changing the var on `.pf-c-drawer.pf-m-inline` instead of `.pf-c-drawer.pf-m-inline .pf-c-drawer__main`. Progress is a good example of doing this, too - https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Progress/progress.scss#L50-L52

src/patternfly/components/Drawer/drawer.scss
mcoker: That's just a workspace thing, the examples have padding so it's showing in the padding. If you remove the padding, it's outside of `.pf-c-drawer`, and `overflow-x` is `hidden`

Screen Shot 2019-07-17 at 4 16 57 PM

src/patternfly/components/Drawer/drawer.scss
christiemolloy: As long as it follows the rules for overriding a variable, then I'm ok with it. To me its harder to follow but maybe thats just me.

christiemolloy: Also I just realized should the modifier var name change from `--pf-c-drawer--m-open__main--MarginRight` to `--pf-c-drawer--m-open--m-inline__main--MarginRight` ?

src/patternfly/components/Drawer/drawer.scss
christiemolloy: All good

src/patternfly/components/Drawer/drawer.scss
mcoker: I just took a glance at other components, looks like we're doing it. We're probably doing both :)

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Dropdown/dropdown.scss#L44-L47

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Dropdown/dropdown.scss#L55-L58

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Table/table.scss#L61-L67

src/patternfly/components/Drawer/docs/code.md
```diff
@@ -0,0 +1,11 @@
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-drawer` | `<div>` | Creates the drawer container. |
+| `.pf-c-drawer__main` | `<div>` | Creates the drawer main container. |
```
christiemolloy: Is the reasoning behind calling this `pf-c-drawer__main` because content from the page's main section goes inside of it? I feel like main implies that its the main thing which I'm not sure it is, what about `pf-c-drawer__content`?

src/patternfly/components/Drawer/docs/code.md
christiemolloy: Ok im good with that, I can't think of another name

src/patternfly/components/Drawer/drawer.scss
christiemolloy: I'm still a little confused here could you explain more as to why they both have a transition

LHinson:
@mcoker is this drawer component also have the capability to push content over? if so, it was mentioned that the drop shadow doesn't make as much sense in that use case since it's not hovering and is pushing. what do you recommend?

mcoker:
@LHinson it does. Initially we discussed leaving that feature out as we didn't want to support it with the notification drawer yet since pushing the content over would make the main content area very narrow in certain viewports and we would need to reconsider our responsive breakpoints to support that. However after looking at the [TopologyView component](https://patternfly-react.surge.sh/patternfly-4/components/topology/), I refactored this drawer to support that feature, since that is how they use the drawer in that component. I just made note in the Drawer documentation that the inline variation (pushes the content over) isn't supported for use in the page component where the drawer is displayed relative to the main content area.

That's a little background for that feature, and it looks like there is a drop-shadow in the TopologyView component.

Do we want to remove the shadow all together for the inline variation and encourage that change on TopologyView? Or maybe disable the drop-shadow by default for the inline variation, and offer a modifier to add the drop-shadow? Or leave it as it is, and add a modifier to disable the box-shadow (either just for the inline variation, or for the default variation, too)?

PR #2080, opened 2019-07-18
feat(divider): add divider component

### LINKED ISSUE #1612: Create divider component
Divider should support both `<hr>` and a `<div>` (or maybe just any block element?). The `<hr>` should be used when you want the implicit `role="separator"` (between siblings, like `li`s in a `ul`) and the `div` should be used between blocks (like `divs`s or `sections`s)

cc @jgiardino @rachael-phillips

christiemolloy:
closes #1612 

- I thought about offering a vertical variation of the divider component - but I can't see any use for it right now, so I was thinking we could add it as we go.
- I thought about using this component in the breadcrumb and the wizard, but those seem to be their own component that is a divider, but is styled very differently (you need to undo all divider styles and add new styles)..

### DIFF
--- a/src/patternfly/components/Divider/divider.hbs
+++ b/src/patternfly/components/Divider/divider.hbs
@@ -1,0 +1,12 @@
+<{{# if divider--type}}{{divider--type}}{{else}}hr{{/if}} class="pf-c-divider{{#if divider--modifier}} {{divider--modifier}}{{/if}}"
+{{#if divider--attribute}}
+  {{divider--attribute}}
+{{/if}}
+{{#if divider--type}}
+  role="separator"
+{{/if}}
+{{#if divider--orientation}}
+  aria-orientation="vertical"
+{{/if}}>
+  {{> @partial-block}}
+</{{#if divider--type}}{{divider--type}}{{else}}hr{{/if}}>
--- a/src/patternfly/components/Divider/divider.hbs
+++ b/src/patternfly/components/Divider/divider.hbs
@@ -4,9 +4,6 @@
 {{/if}}
 {{#if divider--type}}
   role="separator"
-{{/if}}
-{{#if divider--orientation}}
-  aria-orientation="vertical"
 {{/if}}>
   {{> @partial-block}}
 </{{#if divider--type}}{{divider--type}}{{else}}hr{{/if}}>
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -1,0 +1,8 @@
+.pf-c-divider {
+  --pf-c-divider--Height: var(--pf-global--BorderWidth--sm);
+  --pf-c-divider--BackgroundColor: var(--pf-global--BorderColor--100);
+
+  height: var(--pf-c-divider--Height);
+  background-color: var(--pf-c-divider--BackgroundColor);
+  border: 0; // removes the default border styling on an hr
+}
--- a/src/patternfly/components/Divider/docs/code.md
+++ b/src/patternfly/components/Divider/docs/code.md
@@ -4,7 +4,7 @@ The divider renders as an `<hr>` by default. It is possible to make the divider 
 
 | Attribute | Applied to | Outcome |
 | -- | -- | -- |
-| `role="separator"` | `.pf-c-divider.li`, `.pf-c-divider.div` | Indicates that the separator is a separator. |
+| `role="separator"` | `li.pf-c-divider`, `div.pf-c-divider` | Indicates that the separator is a separator. |
 
 ## Usage
 
--- a/src/patternfly/components/Divider/docs/code.md
+++ b/src/patternfly/components/Divider/docs/code.md
@@ -2,7 +2,7 @@
 
 The divider renders as an `<hr>` by default. It is possible to make the divider render as an `li` or a `div` to match the HTML5 specification and context of the divider.
 
-| Title | Margin top | Margin bottom |
+| Attribute | Applied to | Outcome |
 | -- | -- | -- |
 | `role="separator"` | `.pf-c-divider.li`, `.pf-c-divider.div` | Indicates that the separator is a separator. |
 
--- a/src/patternfly/components/Divider/docs/code.md
+++ b/src/patternfly/components/Divider/docs/code.md
@@ -1,0 +1,15 @@
+## Overview
+
+The divider renders as an `<hr>` by default. It is possible to make the divider render as an `li` or a `div` to match the HTML5 specification and context of the divider.
+
+| Title | Margin top | Margin bottom |
+| -- | -- | -- |
+| `role="separator"` | `.pf-c-divider.li`, `.pf-c-divider.div` | Indicates that the separator is a separator. |
+| `aria-orientation="vertical"` | `.pf-c-divider` | Indicates that the separator is oriented vertically. |
+
+
+## Usage
+
+| Class | Applied | Outcome |
+| -- | -- | -- |
+| `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
--- a/src/patternfly/components/Divider/docs/code.md
+++ b/src/patternfly/components/Divider/docs/code.md
@@ -5,8 +5,6 @@ The divider renders as an `<hr>` by default. It is possible to make the divider 
 | Title | Margin top | Margin bottom |
 | -- | -- | -- |
 | `role="separator"` | `.pf-c-divider.li`, `.pf-c-divider.div` | Indicates that the separator is a separator. |
-| `aria-orientation="vertical"` | `.pf-c-divider` | Indicates that the separator is oriented vertically. |
-
 
 ## Usage
 
--- a/src/patternfly/components/Divider/docs/code.md
+++ b/src/patternfly/components/Divider/docs/code.md
@@ -8,6 +8,6 @@ The divider renders as an `<hr>` by default. It is possible to make the divider 
 
 ## Usage
 
-| Class | Applied | Outcome |
+| Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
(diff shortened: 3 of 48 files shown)


src/patternfly/components/Divider/divider.hbs
```diff
@@ -0,0 +1,9 @@
+<{{# if divider--type}}{{divider--type}}{{else}}hr{{/if}} class="pf-c-divider{{#if divider--modifier}} {{divider--modifier}}{{/if}}"
+{{#if divider--attribute}}
+  {{divider--attribute}}
+{{/if}}
+{{#if divider--type}}
+  role="separator"
```
mcoker: nice!

src/patternfly/components/Divider/docs/code.md
```diff
@@ -0,0 +1,13 @@
+## Overview
+
+The divider renders as an `<hr>` by default. It is possible to make the divider render as an `li` or a `div` to match the HTML5 specification and context of the divider.
+
+| Title | Margin top | Margin bottom |
```
mcoker:
```suggestion
| Attribute | Applied to | Outcome |
```

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -303,13 +303,19 @@
...
 
+// remove this block at breaking change period
 .pf-c-dropdown__separator {
   height: var(--pf-c-dropdown__separator--Height);
   margin-top: var(--pf-c-dropdown__separator--MarginTop);
   margin-bottom: var(--pf-c-dropdown__separator--MarginBottom);
   background-color: var(--pf-c-dropdown__separator--BackgroundColor);
 }
 
+.pf-c-divider {
```
mcoker: With all of these changes for the divider in components, you've added `.pf-c-divider` to the global scope. You'll need to move the CSS blocks so they're scoped in the component they apply to.

src/patternfly/components/Divider/docs/code.md
christiemolloy: bitmoji

src/patternfly/components/Divider/docs/code.md
```diff
@@ -0,0 +1,13 @@
+## Overview
+
+The divider renders as an `<hr>` by default. It is possible to make the divider render as an `li` or a `div` to match the HTML5 specification and context of the divider.
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `role="separator"` | `.pf-c-divider.li`, `.pf-c-divider.div` | Indicates that the separator is a separator. |
```
mcoker: Moved the elements before the class. We do that everywhere else. Buttons and tabs (the nav examples) are a couple of examples. 

```suggestion
| `role="separator"` | `li.pf-c-divider`, `div.pf-c-divider` | Indicates that the divider is a separator. |
```

src/patternfly/components/Nav/nav.scss
```diff
@@ -218,6 +218,21 @@
...
     visibility: visible;
   }
+
+  // Separator
+  .pf-c-divider {
+    margin-top: var(--pf-c-nav__separator--MarginTop); // change var name at breaking change
+    margin-bottom: var(--pf-c-nav__separator--MarginBottom); // change var name at breaking change
+    margin-left: var(--pf-c-nav__separator--MarginLeft); // change var name at breaking change
+
+    .pf-c-nav__simple-list & {
```
mcoker: This selector compiles to `.pf-c-nav__simple-list .pf-c-nav .pf-c-divider`.

Maybe you should go ahead and create vars for `--pf-c-nav__c-divider--MarginTop/Bottom/Left`, then modify the vars in the CSS blocks for `.pf-c-nav__simple-list` and `.pf-c-nav__item`?

src/patternfly/components/Nav/nav.scss
```diff
@@ -218,6 +218,21 @@
...
+      --pf-c-nav__separator--MarginLeft: var(--pf-c-nav__simple-list__separator--MarginLeft); // change var name at breaking change
+    }
+
+    .pf-c-nav__item & {
```
mcoker: Same here

src/patternfly/components/Nav/nav.scss
christiemolloy: Yep I updated all the components to be like this now, and then we can just remove vars at breaking change rather than changing var names etc.

src/patternfly/components/Divider/docs/code.md
```diff
@@ -0,0 +1,13 @@
...
+
+The divider renders as an `<hr>` by default. It is possible to make the divider render as an `li` or a `div` to match the HTML5 specification and context of the divider.
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `role="separator"` | `li.pf-c-divider`, `div.pf-c-divider` | Indicates that the separator is a separator. |
+
+## Usage
+
+| Class | Applied | Outcome |
```
mcoker:
```suggestion
| Class | Applied to | Outcome |
```

src/patternfly/components/Nav/nav.scss
mcoker: Still seeing the problem with the nav selectors. https://www.sassmeister.com/gist/966ff367643b853eea5b2dc4a35e5d09

It works with `.pf-c-nav__separator` because that element is outside of the scope of `.pf-c-nav`

src/patternfly/components/Divider/divider.scss
```diff
@@ -0,0 +1,8 @@
+.pf-c-divider {
+  --pf-c-divider--Height: var(--pf-global--BorderWidth--sm);
+  --pf-c-divider--BackgroundColor: var(--pf-global--BorderColor--100);
+
+  height: var(--pf-c-divider--Height);
```
mattnolting:
```suggestion
  height: var(--pf-c-divider--Height);
  margin-top: var(--pf-c-divider--MarginTop);  
  margin-right: var(--pf-c-divider--MarginRight);  
  margin-bottom: var(--pf-c-divider--MarginBottom);  
  margin-left: var(--pf-c-divider--MarginLeft);  
```

src/patternfly/components/Divider/divider.scss
```diff
@@ -0,0 +1,8 @@
+.pf-c-divider {
+  --pf-c-divider--Height: var(--pf-global--BorderWidth--sm);
```
mattnolting:
```suggestion
  --pf-c-divider--Height: var(--pf-global--BorderWidth--sm);
  --pf-c-divider--MarginTop: 0;
  --pf-c-divider--MarginRight: 0;
  --pf-c-divider--MarginBottom: 0;
  --pf-c-divider--MarginLeft: 0;
```

src/patternfly/components/Nav/nav.scss
christiemolloy: Oh sorry I didn't understand that the first time. Could you check that they're in the right blocks I wasn't 100% sure @mcoker

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -91,6 +95,11 @@
   position: relative;
   display: inline-block;
   max-width: 100%;
+
+  .pf-c-divider {
+    margin-top: var(--pf-c-dropdown__c-divider--MarginTop);
+    margin-bottom: var(--pf-c-dropdown__c-divider--MarginBottom);
+  }
```
mattnolting:
```suggestion
  --pf-c-divider--MarginTop: var(--pf-c-dropdown__c-divider--MarginTop);
  --pf-c-divider--MarginBottom: var(--pf-c-dropdown__c-divider--MarginBottom);
  }
```

src/patternfly/components/Nav/nav.scss
```diff
@@ -218,6 +225,13 @@
     display: block;
     visibility: visible;
   }
+
+  // Divider
+  .pf-c-divider {
+    margin-top: var(--pf-c-nav__c-divider--MarginTop);
+    margin-bottom: var(--pf-c-nav__c-divider--MarginBottom);
+    margin-left: var(--pf-c-nav__c-divider--MarginLeft);
+  }
```
mattnolting:
```suggestion
      --pf-c-divider--MarginTop: var(--pf-c-nav__c-divider--MarginTop);
      --pf-c-divider--MarginRight: var(--pf-c-nav__c-divider--MarginRight);
      --pf-c-divider--MarginLeft: var(--pf-c-nav__c-divider--MarginLeft);
    }
```

src/patternfly/components/OptionsMenu/options-menu.scss
```diff
@@ -64,15 +64,24 @@
...
+  --pf-c-options-menu__separator--MarginBottom: var(--pf-global--spacer--sm); // remove at breaking change
 
   position: relative;
   display: inline-block;
   max-width: 100%;
+
+  .pf-c-divider {
+    margin-top: var(--pf-c-options-menu__c-divider--MarginTop);
+    margin-bottom: var(--pf-c-options-menu__c-divider--MarginBottom);
+  }
```
mattnolting:
```suggestion
    --pf-c-divider--MarginTop: var(--pf-c-options-menu__c-divider--MarginTop);
    --pf-c-divider--MarginBottom: var(--pf-c-options-menu__c-divider--MarginBottom);
  }
```

christiemolloy:
@mattnolting I didn't do that because we don't have padding/margins on an individual component, although I agree it would be helpful in this case

mattnolting:
@christiemolloy Great point, great point. I could be persuaded either way. This is definitely a special type of component, as it would never really exist outside of another component. Just food for thought. @mcoker wyt?

mcoker:
Yeah... we don't define margins on components, but here we aren't actually defining a margin, sooo :) We also decided a while back that when we overwrite variables, we only do so if the original variable was defined and used. We don't create vars only so that we can overwrite them, which is what we would be doing in this case. (the [button component's](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Button/button.scss) `color` and `background-color` are an example). So I don't think we would do that according to how we've written components thus far.

Also, don't modify other components' variables in a component currently. We always overwrite the property instead. That came after we decided it's easier to maintain if we overwrite the properties, since if we change a variable's name, that breaks any other components that reverence that var name. 

I'm happy to revisit all of that. I think it's smart use of the variable system and we could end up with less code that way. But I think we would need to discuss it (and adding support for a margin) as a group if we want to do that. It might also be hard to define (for consistency's sake) when we choose to create a var for something that isn't used, just so that it can be overwritten.

Issue #1972, opened 2019-06-21
AppLauncher - add support for modifiers top and align right and disabled (at both the item level and whole toggle button level)


jschuler:
AppLauncher should support similar modifiers to dropdown so that the menu can be right-aligned or the direction flipped up

Also consumer should be able to add `pf-m-disabled` at the button toggle as well as the menu item level

https://github.com/patternfly/patternfly-react/issues/2322

jschuler:
And also for disabled item `pf-m-disabled`

PR #2081, opened 2019-07-22
fix(applauncher): add support for right and top alignment

mattnolting:
fixes: #1972 
relates to: https://github.com/patternfly/patternfly-react/issues/2322


mattnolting:
@christiemolloy yep, added disabled styling. Disabled already exists at menu-item level.

src/patternfly/components/AppLauncher/examples/app-launcher-disabled-example.hbs
```diff
@@ -0,0 +1,7 @@
+{{#> app-launcher id="app-launcher-example" app-launcher--IsDisabled="true"}}
```
mcoker:
```suggestion
{{#> app-launcher id="app-launcher-example-disabled" app-launcher--IsDisabled="true"}}
```

PR #2083, opened 2019-07-22
fix(datalistdemo): add in header to avoid a11y issue

### LINKED ISSUE #2082: Give Data List Demo a header
Related to: https://github.com/patternfly/patternfly-org/issues/1348

christiemolloy:
closes #2082


christiemolloy:
I have another issue for that one, I split them up because the basic forms has another issue also.

PR #2086, opened 2019-07-23
fix(a11y): basic forms demo

### LINKED ISSUE #2085: Fix a11y issues in Core Form Demo
Related to: https://github.com/patternfly/patternfly-org/issues/1346

These a11y issues are coming from the Core demo for Basic forms

- Checkbox inputs with the same name attribute value must be part of a group
- Headings must not be empty

christiemolloy:
closes #2085 

- Added headingtext to the forms demo 
- made the `name=...` assignments from the `check-input` unique as this is what we have in the check component. This fixes the error for "Checkbox inputs with the same name attribute value must be part of a group"... I am wondering though @jgiardino do these checks need a form fieldset around them like in the select component?


src/patternfly/demos/BasicForms/examples/index.js
```diff
@@ -11,12 +11,13 @@ import docs from '../docs/code.md';
 
 export const Docs = docs;
 
-export default (props) => {
+export default props => {
   const simpleFormDemoExample = SimpleFormDemoExample();
   const horizontalFormDemoExample = HorizontalFormDemoExample();
+  const headingText = 'Basic Forms Demo';
```
mcoker: should be sentence case

```suggestion
  const headingText = 'Basic forms demo';
```

redallen:
Merging master to get latest CircleCI changes.

src/patternfly/demos/BasicForms/examples/index.js
redallen: Done.

PR #2088, opened 2019-07-23
fix(radio): give unique names to fix a11y

### LINKED ISSUE #2087: fix(a11y): Radio
Related to: https://github.com/patternfly/patternfly-org/issues/1337

There is an issue coming from Core: "Radio inputs with the same name attribute value must be part of a group"

christiemolloy:
fix #2087 

@jgiardino I followed what we have in the check component where we have unique names for the radio inputs, this is also what I did in #2086


redallen:
Merged to get latest CircleCI changes.

PR #2101, opened 2019-07-24
fix(logo): PF logo missing width attribute

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2100


kybaker:
@mcoker Looks good!

mcoker:
That's just the width for our image. We don't want to impose a width (or height) on comsumers' logos since they can have logos that vary in size.

PR #2107, opened 2019-07-25
feat(alert): add default alert

### LINKED ISSUE #1749: Additional default state for notifications
There are a few needs on the product side to have a default notification.

This would require a generic colour and icon that would show in inline/toast/notification drawer implementations of the notification.

See this comment for a better understanding: 
https://docs.google.com/a/redhat.com/document/d/1ujaWYYvuKE_lK4JmD1DLHwr1zYJ-h5652QMba6TcWo8/edit?disco=AAAACsEDtzU

OpenShift also has a need for this as @cshinn has pointed out. Where a user could create an alert that doesn't fit into the four options we provide. He has made the case for a more generic notification that could be tied to this custom alert.

christiemolloy:
closes #1749 

spoke with @mceledonia about replacing the cyan vars in `colors.scss` with the new colors we added to the global chart vars. my only question is @mceledonia do you have any thoughts on what you would like to name them, now that we're assigning the cyan colors as variables? I went with default for the moment as that is what I inferred from the issue.


src/patternfly/sass-utilities/colors.scss
```diff
@@ -18,13 +18,11 @@ $pf-color-blue-400:              #06c !default;
 $pf-color-blue-500:              #004080 !default;
 $pf-color-blue-600:              #004368 !default;
 $pf-color-blue-700:              #002235 !default;
-$pf-color-cyan-100:              #8bb4b9 !default;
-$pf-color-cyan-200:              #5c979d !default;
-$pf-color-cyan-300:              #2e7981 !default;
-$pf-color-cyan-400:              #005c66 !default;
-$pf-color-cyan-500:              #00434b !default;
-$pf-color-cyan-600:              #003d44 !default;
-$pf-color-cyan-700:              #001f22 !default;
```
mattnolting: Should we keep `-600` and `-700` to avoid breaking change?

src/patternfly/components/Alert/alert.scss
```diff
@@ -133,15 +148,16 @@
...
     --pf-c-alert--m-inline--before--BackgroundColor: var(--pf-c-alert--m-inline--m-info--before--BackgroundColor);
+
+    &.pf-m-inline {
+      --pf-c-alert__icon--Color: var(--pf-c-alert--m-inline--m-info__icon--Color);
+    }
   }
 
   &.pf-m-inline {
     --pf-c-alert--BoxShadow: var(--pf-c-alert--m-inline--BoxShadow);
     --pf-c-alert__icon--BackgroundColor: var(--pf-c-alert--BackgroundColor);
```
mcoker: Looks like we have a var for this value that is currently unused `--pf-c-alert--m-inline__icon--BackgroundColor`, and I think it should just be `transparent` instead of `var(--pf-c-alert--BackgroundColor)`.

mcoker:
@mattnolting would you mind looking at this again? We'd like to get it in while @christiemolloy is out. Created 2 follow up issues to [remove cyan-600 and cyan-700](https://github.com/patternfly/patternfly-next/issues/2115) and [rename the grid vars](https://github.com/patternfly/patternfly-next/issues/2114) as noted in your PR.

mcoker:
@mceledonia or @lboehling can you review when you get a chance?

PR #2119, opened 2019-08-01
feat(data toolbar): add data toolbar

mattnolting:
WIP

- Finishing overflow menu documentation; toolbar is ready for review

more info to come...


src/site/workspace.scss
```diff
@@ -536,14 +536,39 @@ html {
...
+.Preview__body {
+  .example-border {
+    background: none;
+    border: 2px dashed #393f44;
+    box-shadow: none;
+  }
+}
 
+// Toolbar documentation
+.toolbar-documentation {
```
dgutride: These will all be lost when we show the documentation on patternfly.org - please chat with me so we can discuss a different way to accomplish this.  If something in the preview is deficient for showing toolbar examples - we'll need to make a global change instead.

jgiardino:
Great job on this so far. This is impressive work and it's great to see this pattern starting to get implemented.

I reviewed this in person with @mattnolting, and here are some of the things we chatted about...

My original expectation when I saw the design pattern was that for the scenario where the filters are available from a toggle button, that the filters would display within something like a popover/modal type of container. This is consistent with my very black and white perspective on the world of UI, where things that display on top of the page use a modal interaction pattern, and things that display inline use the disclosure interaction pattern. But at the time of reviewing the design, I didn't realize that there are really 3 variations of this toolbar:

1. All controls are inline, no toggle - at the larger breakpoint
2. A toggle button that shows the filters group, and the group is inline
3. A toggle button that shows the filters group, and the group is on top of contents - at the smaller breakpoint

This raised some questions about how we really want to handle the different breakpoints, and whether html should handle all of this, or if we should instead handle the breakpoints in the react implementation given that each variation has slightly different behavior and aria semantics to implement.

We did talk about why the trapping of keyboard focus and screen reader focus would matter at smaller breakpoints. Here are a couple of scenarios where that could occur:
- a screen reader user is using a phone with a bluetooth keyboard
- a low vision user has zoomed in on the browser, triggering the responsive layout AND the low vision user is also a keyboard only user. 
  - NOTE: I don't think breakpoints are set up to use rems currently, but we should so that low vision users can take advantage of the responsive layouts for cases like this

We also talked about updating the a11y documentation to clearly identify which aria attributes are important for which layout/breakpoint size, so that it's more obvious which patterns are needed for which layout.

redallen:
Merged master to get latest CircleCI changes.

src/patternfly/components/DataToolbar/data-toolbar-expandable-content.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="pf-c-data-toolbar__expandable-content{{#if data-toolbar-expandable-content--IsExpanded}} pf-m-expanded{{/if}}{{#if data-toolbar-expandable-content--modifier}} {{data-toolbar-expandable-content--modifier}}{{/if}}" id="{{data-toolbar--id}}"
```
mcoker: can you put `id="{{data-toolbar--id}}"` on a new line so it's before `{{#if data-toolbar-expandable-content--attribute}}`?

src/patternfly/components/DataToolbar/data-toolbar-group-toggle-group.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-data-toolbar__group pf-m-toggle-group{{#if data-toolbar-group--modifier}} {{data-toolbar-group--modifier}}{{/if}}{{#if data-toolbar-group-toggle-group--Reveal}} pf-m-reveal{{data-toolbar-group-toggle-group--Reveal}}{{/if}}{{#if data-toolbar-toggle-group--modifier}} {{data-toolbar-toggle-group--modifier}}{{/if}}"
+  {{#if data-toolbar-toggle-group--attribute}}
```
mcoker: should you use `data-toolbar-group-toggle-group--[modifier/attribute]` as the prefix for this partial's attribute vars? You're using `data-toolbar-group-toggle-group--Reveal`. Generally the attribute matches the partial name.

src/patternfly/components/DataToolbar/data-toolbar-item-overflow-menu.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-data-toolbar__item pf-m-overflow-menu{{#if data-toolbar-overflow-menu--modifier}} {{data-toolbar-overflow-menu--modifier}}{{/if}}"
+  {{#if data-toolbar-overflow-menu--attribute}}
```
mcoker: same here, should you use `data-toolbar-item-overflow-menu--[attribute/modifier]`?

src/patternfly/components/DataToolbar/data-toolbar-item-search-filter.hbs
```diff
@@ -0,0 +1,11 @@
+{{#> data-toolbar-item data-toolbar-item--modifier=(concat 'pf-m-search-filter ' data-toolbar-item-search-filter--modifier)}}
```
mcoker: 🔥

src/patternfly/components/DataToolbar/data-toolbar-group-toggle-group.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-data-toolbar__group pf-m-toggle-group{{#if data-toolbar-group--modifier}} {{data-toolbar-group--modifier}}{{/if}}{{#if data-toolbar-group-toggle-group--Reveal}} pf-m-reveal{{data-toolbar-group-toggle-group--Reveal}}{{/if}}{{#if data-toolbar-toggle-group--modifier}} {{data-toolbar-toggle-group--modifier}}{{/if}}"
```
mcoker: Should you use something like `{{#> data-toolbar-group data-toolbar-group--modifier=(concat 'pf-m-toggle-group ' data-toolbar-toggle-group--modifier)}}`

And do you need `{{#if data-toolbar-group--modifier}} {{data-toolbar-group--modifier}}{{/if}}` here? Or could modifiers to this partial be covered by `data-toolbar-toggle-group--modifier`?

src/patternfly/components/DataToolbar/data-toolbar-item-sort.hbs
```diff
@@ -0,0 +1,14 @@
+{{#> data-toolbar-item}}
+  {{#> button button--modifier="pf-m-plain"}}
+    <i class="fas fa-sort-amount-down" aria-hidden="true"></i>
+  {{/button}}
+{{/data-toolbar-item}}
+
+{{!-- {{#> data-toolbar-item}}
```
mcoker: any reason to keep this block around?

src/patternfly/components/DataToolbar/data-toolbar-item-total-items.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="pf-c-data-toolbar__item pf-m-pagination">
```
mcoker: Can you use `{{#> data-toolbar__item}}` here?

src/patternfly/components/DataToolbar/data-toolbar-item-sort.hbs
```diff
@@ -0,0 +1,14 @@
+{{#> data-toolbar-item}}
```
mcoker: Can you use {{#> data-toolbar__item}} here?

src/patternfly/components/DataToolbar/data-toolbar-item-overflow-menu.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-data-toolbar__item pf-m-overflow-menu{{#if data-toolbar-overflow-menu--modifier}} {{data-toolbar-overflow-menu--modifier}}{{/if}}"
```
mcoker: Can you use {{#> data-toolbar-item}} here and `concat()` `pf-m-overflow-menu` and `data-toolbar-overflow-menu--modifier` in `data-toolbar-item--modifier`?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+  --pf-c-data-toolbar__group--spacer: var(--pf-c-data-toolbar--spacer--base);
+
+  // Item
+  --pf-c-data-toolbar--item--Display: flex;
+  --pf-c-data-toolbar--item--Visibility: visible;
+
+  // Toggle group
+  --pf-c-data-toolbar__group--m-toggle-group--spacer: var(--pf-global--spacer--sm);
+  --pf-c-data-toolbar__group--m-toggle-group--m-reveal--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__toggle--c-badge: var(--pf-global--spacer--sm);
```
mcoker: looks like this should be `--pf-c-data-toolbar__toggle--c-badge--MarginLeft`

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+
+  // Label
+  --pf-c-data-toolbar--m-label--spacer: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar--m-label--FontWeight: var(--pf-global--FontWeight--bold);
+
+  // Search filter
+  --pf-c-data-toolbar--m-search-filter--spacer: var(--pf-global--spacer--sm);
+  --pf-c-data-toolbar--m-search-filter--space-items: 0;
+
+  // Separator
```
mcoker: Can we use the divider component here?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+
+  @media screen and (min-width: $pf-global--breakpoint--lg) {
+    position: static;
+    box-shadow: none;
+  }
+
+  &.pf-m-expanded {
+    display: grid;
+    grid-template-columns: 1fr;
+    grid-row-gap: var(--pf-c-data-toolbar__expandable-content--m-expanded--GridGap);
```
mcoker: Can you use `GridRowGap` in the var?

src/patternfly/components/DataToolbar/examples/index.js
```diff
@@ -0,0 +1,101 @@
...
+export default props => {
+  const dataToolbarItemsExample = DataToolbarItemsExample();
+  const dataToolbarGroupTypesExample = DataToolbarGroupTypesExample();
+  const dataToolbarToggleGroupExample = DataToolbarToggleGroupExample();
+  const dataToolbarSelectedFiltersExample = DataToolbarSelectedFiltersExample();
+  const dataToolbarSpacersExample = DataToolbarSpacersExample();
+  const dataToolbarGroupSpacersExample = DataToolbarGroupSpacersExample();
+  const dataToolbarStackedExample = DataToolbarStackedExample();
+
+  const headingText = 'Data Toolbar';
```
mcoker:
```suggestion
  const headingText = 'Data toolbar';
```

src/patternfly/components/OptionsMenu/options-menu.scss
```diff
@@ -155,6 +155,8 @@
   }
 
   &.pf-m-plain {
+    justify-content: center;
```
mcoker: This declaration is in the selector below that excludes the `.pf-m-text` variation. If it applies to the `.pf-m-text` variation and you set a `width` on the options menu, you get this:

Screen Shot 2019-08-13 at 4 46 50 PM

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+
+  // Expanded
+  --pf-c-data-toolbar__expandable-content--m-expanded--GridGap: var(--pf-global--gutter);
+
+  // Base spacer - shared value
+  --pf-c-data-toolbar--spacer--base: var(--pf-global--spacer--md);
+
+  // Spacer values
+  --pf-c-data-toolbar__item--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--spacer: var(--pf-c-data-toolbar--spacer--base);
```
mcoker: is it no longer the case that standalone items and/or groups have a 24px spacer and that items in a group have 16px?

src/patternfly/components/DataToolbar/docs/code.md
```diff
@@ -0,0 +1,56 @@
+## Overview
+
+As the data toolbar component is a hybrid layout and component, some of its elements are presentational, while some require accessibility support.
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-data-toolbar` | `<div>` | Initiates the toolbar component. **Required** |
+| `.pf-c-data-toolbar__item` | `<div>` | Initiates the toolbar component item. **Required** |
```
mcoker: I would use "Initiates _a_ toolbar component item" since you can have more than 1 of them. And continue to use "the" when you can only have 1.

src/patternfly/components/DataToolbar/docs/code.md
```diff
@@ -0,0 +1,56 @@
...
+| `.pf-c-data-toolbar__group` | `<div>` | Initiates the toolbar component group. |
+| `.pf-c-data-toolbar__content` | `<div>` | Initiates the toolbar component content. |
+| `.pf-c-data-toolbar__expandable-content` | `<div>` | Initiates the toolbar component expandable content. |
+| `.pf-m-expanded` | `.pf-c-data-toolbar__expandable-content` | Modifies the component for the expanded state. |
+| `.pf-c-data-toolbar__divider` | `<div>` | Initiates the toolbar component divider. |
+| `.pf-m-bulk-select` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. |
```
mcoker: Usually we'll put a description of what it modifies, like "modifies bulk select toolbar item spacing"

src/patternfly/components/DataToolbar/docs/code.md
```diff
@@ -0,0 +1,56 @@
...
+| `.pf-m-overflow-menu` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. |
+| `.pf-m-pagination` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. |
+| `.pf-m-separator` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. |
+| `.pf-m-button-group` | `.pf-c-data-toolbar__group` | Modifies toolbar group spacing. |
+| `.pf-m-icon-button-group` | `.pf-c-data-toolbar__group` | Modifies toolbar group spacing. |
+| `.pf-m-filter-group` | `.pf-c-data-toolbar__group` | Modifies toolbar group spacing. |
+| `.pf-m-contextual-search` | `.pf-c-data-toolbar__group` | Modifies toolbar group spacing. |
+| `.pf-m-hidden{-on-[breakpoint]}` | `.pf-c-data-toolbar > *` | Modifies toolbar element to hidden at breakpoint. |
```
mcoker: Instead of "to hidden" made "to hide" or "to be hidden"?

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: That is correct, @mceledonia and I reviewed the spec and arrived at this conclusion.

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: I don't think so, we don't have a vertical divider option yet...

mcoker: Oh sorry, I meant adding that as an enhancement to that component and using it here. We didn't create the vertical divider because we didn't need that variation yet, but I think we do now. I'll create an issue for that and we can use this as it is and update it once we add the variation to the divider.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+.pf-c-data-toolbar__item {
+  // reset spacer
+  --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar__item--spacer);
+}
+
+.pf-c-data-toolbar__group,
+.pf-c-data-toolbar__item {
+  margin-right: var(--pf-c-data-toolbar--spacer);
+
+  &:last-child:not([class*="pf-m-spacer"]) {
```
mcoker: nice! probably best to allow them to define this spacing even if it's the last child.

src/patternfly/components/DataToolbar/docs/data-toolbar-spacers-intro.md
```diff
@@ -0,0 +1,3 @@
+<h3 class="Example_heading">Data toolbar spacers</h3>
+
+In some instances, it may be necessary to adjust spacing explicitly where items are being hidden/revealed. For example, if a `.pf-m-toogle-group` is adjacent to an elements being hidden/show or hidden/revealed, the spacing may appear to be inconsistent. If possible, rely on modifier values. Available spacer modifiers are `.pf-m-spacer-{none, sm, md, lg}{-on-md, -on-lg, -on-xl}` and `.pf-m-space-items-{none, sm, md, lg}{-on-md, -on-lg, -on-xl}`. These modifiers will overwrite existing modifiers provided by `.pf-c-data-toolbar`.
```
mcoker:
```suggestion
In some instances, it may be necessary to adjust spacing explicitly where items are being hidden/revealed. For example, if a `.pf-m-toogle-group` is adjacent to an element being hidden/show or hidden/revealed, the spacing may appear to be inconsistent. If possible, rely on modifier values. Available spacer modifiers are `.pf-m-spacer-{none, sm, md, lg}{-on-md, -on-lg, -on-xl}` and `.pf-m-space-items-{none, sm, md, lg}{-on-md, -on-lg, -on-xl}`. These modifiers will overwrite existing modifiers provided by `.pf-c-data-toolbar`.
```

src/patternfly/components/DataToolbar/docs/data-toolbar-toggle-group-intro.md
```diff
@@ -0,0 +1,3 @@
+### Toggle group modifier
+The `.pf-m-toggle-group` controls when, and at which breakpoint, filters will be hidden and revealed. By default, all filters are hidden until the specified breakpoint is reached.
+- `.pf-m-reveal-on-{md, lg, xl}`: controls when filters are revealed and when toggle button is hidden.
```
mcoker: bullet isn't showing up, this example doesn't have `className="is-layout-page"`. Some others are missing it, too, not sure if it should be added to those.

src/patternfly/components/DataToolbar/docs/data-toolbar-toggle-group.md
```diff
@@ -0,0 +1,27 @@
+
+
+## Accessibility
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `hidden` | `.pf-c-data-toolbar__item`, `.pf-c-data-toolbar__group`, `.pf-c-data-toolbar__toggle`, `.pf-c-data-toolbar__expandable-content` |  Indicates that the toggle group element is hidden. **Required** |
+| `aria-expanded="true"` | `.pf-c-data-toolbar__toggle > .pf-c-button` |  Indicates that the expandable content is visible. **Required** |
+| `aria-expanded="false"` | `.pf-c-data-toolbar__toggle > .pf-c-button` |  Indicates the the expandable content is hidden. **Required** |
+| `aria-controls="[id of expandable content]"` | `.pf-c-data-toolbar__toggle > .pf-c-button` |  Identifies the expanded content controlled by the toggle button when . **Required** |
```
mcoker: seems like part of the sentence is missing.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+
+  // Content
+  --pf-c-data-toolbar__content--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__content--PaddingRight: var(--pf-global--spacer--lg);
+  --pf-c-data-toolbar__content--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__content--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-data-toolbar__content--md--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__content--md--PaddingLeft: var(--pf-global--spacer--md);
+
+  @media screen and (max-width: $pf-global--breakpoint--md) {
```
mcoker: Can we use `min-width` here?

src/patternfly/components/DataToolbar/data-toolbar-toggle.hbs
```diff
@@ -0,0 +1,13 @@
...
+  {{/if}}>
+  {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'aria-label="Reveal filters" aria-expanded="' data-toolbar-group--IsExpanded '" aria-controls="' data-toolbar--id '-expandable-content"')}}
+    <i class="fas fa-filter" aria-hidden="true"></i>
+    {{#if data-toolbar-toggle--SelectedFilters}}
+      {{#> badge badge--modifier="pf-m-read"}}
+        {{{data-toolbar-toggle--SelectedFilters}}}
+      {{/badge}}
+    {{/if}}
+  {{/button}}
+</div>
```
mcoker: can you put an empty blank line at the end of this file?

src/patternfly/components/DataToolbar/data-toolbar-item-separator.hbs
```diff
@@ -0,0 +1 @@
+{{#> data-toolbar-item data-toolbar-item--attribute='aria-hidden="true"' data-toolbar-item--modifier="pf-m-separator"}}{{/data-toolbar-item}}
```
mcoker: and an empty blank line here?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
+$pf-c-data-toolbar--breakpoint-map: build-breakpoint-map("base", "md", "lg", "xl", "2xl");
+$pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
+
+// Data toolbar
+.pf-c-data-toolbar {
+  --pf-c-data-toolbar--BackgroundColor: var(--pf-global--BackgroundColor--100);
+
+  // Content
+  --pf-c-data-toolbar__content--PaddingTop: var(--pf-global--spacer--md);
```
mcoker: looks like all of the examples use the `toolbar__content` element. From the original issue:

> There should be no spacers on the left and right edges of the toolbar if there are already padding spacers being used with whatever component the toolbar is placed within (e.g. cards).

Did that requirement change?

src/patternfly/components/DataToolbar/docs/data-toolbar-intro.md
```diff
@@ -0,0 +1,45 @@
...
+
+<br>
+### Data toolbar item types
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-bulk-select` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. Spacer value is set to `var(--pf-c-data-toolbar__item--m-bulk-select--spacer)`. |
+| `.pf-m-overflow-menu` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. Spacer value is set to `var(--pf-c-data-toolbar__item--m-overflow-menu--spacer)`. |
+| `.pf-m-pagination` | `.pf-c-data-toolbar__item` | Modifies toolbar item spacing. Spacer value is set to `var(--pf-c-data-toolbar__item--m-pagination--spacer)`. |
+<br>
+<h3 class="Example_heading">Spacer system</h3>
```
mcoker: The indentation on this gets a little wonky in markdown. I replaced this through the "available breakpoints" line with this code and it looks better

```
<h3 class="Example_heading">Spacer system</h3>

**Spacer** - `.pf-m-spacer-{none, sm, md, lg}`.
- This modifier can be applied to a group or item.
- This modifier changes the spacer value for only the element to which it is applied.
- Responsive spacers can be used by appending `{-on-[breakpoint]}` to `.pf-m-spacer-{size}`. **Example:** `.pf-m-spacer-lg-on-xl`.

<br>
**Space items** - `.pf-m-space-items-{none, sm, md, lg}`.
- This modifier affects only groups.
- It changes the spacing of direct child items and can be overwritten with any other modifier.
- Responsive spacers can be used by appending `{-on-[breakpoint]}` to `.pf-m-space-items-{size}`. **Example:** `.pf-m-space-items-lg-on-xl`.

<br>
### Available breakpoints are: `-on-md, -on-lg, -on-xl, -on-2xl`.
```

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+  --pf-c-data-toolbar__group--m-button-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-button-group--space-items: var(--pf-global--spacer--sm);
+
+  // Icon button group spacer
+  --pf-c-data-toolbar__group--m-icon-button-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-icon-button-group--space-items: 0;
+
+  // Icon group spacer
+  --pf-c-data-toolbar__group--m-filter-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-filter-group--space-items: 0;
```
mcoker: Have you considered applying spacing similar to the input group here so that you don't end up with double borders where elements meet adjacent elements?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+  // Button group
+  --pf-c-data-toolbar__group--m-button-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-button-group--space-items: var(--pf-global--spacer--sm);
+
+  // Icon button group spacer
+  --pf-c-data-toolbar__group--m-icon-button-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-icon-button-group--space-items: 0;
+
+  // Icon group spacer
+  --pf-c-data-toolbar__group--m-filter-group--spacer: var(--pf-c-data-toolbar--spacer--base);
```
mcoker: Do you think the `-group` on the group modifiers is redundant? I think it's fine, but think it would be fine to drop it, too. Just wanted to mention it if you hadn't considered it.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+  box-shadow: var(--pf-c-data-toolbar__expandable-content--BoxShadow);
+
+  @media screen and (min-width: $pf-global--breakpoint--lg) {
+    position: static;
+    box-shadow: none;
+  }
+
+  &.pf-m-expanded {
+    display: grid;
+    grid-template-columns: 1fr;
```
mcoker: do you need to define this?

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: Nope! Good catch.

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: Yeah, I thought about omitting it. `-group` was more readable to me than:
- `.pf-c-data-toolbar__group.pf-m-filters`
- `.pf-c-data-toolbar__group.pf-m-buttons`
- `.pf-c-data-toolbar__group.pf-m-icon-buttons`

I could go either way, though

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: nice catch. I had it there to place `clear all filters` button to [auto]column: 2, but I don't think we need it.

src/patternfly/components/DataToolbar/docs/data-toolbar-intro.md
```diff
@@ -0,0 +1,46 @@
+Data toolbar relies on groups (`.pf-c-data-toolbar__group`) and items (`.pf-c-data-toolbar__item`), with default spacer values. Groups and items can be siblings and/or items can be nested within groups. Modifier selectors adjust spacing based on the type of group or item. Each modifier applies a unique CSS variable, therefore, the base spacer value for all elements can be customized and item/groups spacers can be themed individually. The default spacer value for items and groups is set to `--pf-global--spacer--md` or 16px.
```
mcoker: do you think it's worth saying the default spacer value is set to `--pf-c-data-toolbar--spacer--base`, in the context of customizing/theming items/groups? And then you could mention that the value for that is `--pf-global--spacer--md` or 16px. Like

> The default spacer value for items and groups is set to `--pf-c-data-toolbar--spacer--base`, whose value is `--pf-global--spacer--md` or 16px.

src/patternfly/components/DataToolbar/docs/data-toolbar-intro.md
```diff
@@ -0,0 +1,46 @@
...
+- This modifier can be applied to a group or item.
+- This modifier changes the spacer value for only the element to which it is applied.
+- Responsive spacers can be used by appending `{-on-[breakpoint]}` to `.pf-m-spacer-{size}`. **Example:** `.pf-m-spacer-lg-on-xl`.
+
+<br>
+**Space items** - `.pf-m-space-items-{none, sm, md, lg}`.
+- This modifier affects only groups.
+- It changes the spacing of direct child items and can be overwritten with any other modifier.
+- Responsive spacers can be used by appending `{-on-[breakpoint]}` to `.pf-m-space-items-{size}`. **Example:** `.pf-m-space-items-lg-on-xl`.
+
```
mcoker: Could this also not be a standard usage table? Like

| -- | -- | -- |
| `.pf-m-spacer-{none, sm, md, lg}{-on-[breakpoint]}` | `.pf-c-data-toolbar__item`, `.pf-c-data-toolbar__group` | Modifies the spacer value for only the element to which it is applied. **Example:** `.pf-m-spacer-lg-on-xl`.|
| `.pf-m-space-items-{none, sm, md, lg}{-on-[breakpoint]}` | `.pf-c-data-toolbar__group` | Modifies the spacing of direct child items and can be overwritten with any other modifier. **Example:** `.pf-m-space-items-lg-on-xl`. |

mcoker: er actually, this table basically exists under "adjusting item spacers" and "adjusting group spacers." Do you think it's best to document it in the intro and under those sections?

src/patternfly/components/DataToolbar/docs/data-toolbar-toggle-group-intro.md
```diff
@@ -0,0 +1,2 @@
+### Toggle group modifier
+The `.pf-m-toggle-group` controls when, and at which breakpoint, filters will be hidden and revealed. By default, all filters are hidden until the specified breakpoint is reached. `.pf-m-reveal-on-{md, lg, xl}`: controls when filters are revealed and when toggle button is hidden.
```
mcoker:
```suggestion
The `.pf-m-toggle-group` controls when, and at which breakpoint, filters will be hidden and revealed. By default, all filters are hidden until the specified breakpoint is reached. `.pf-m-reveal-on-{md, lg, xl}` controls when filters are revealed and when the toggle button is hidden.
```

src/patternfly/components/DataToolbar/docs/code.md
```diff
@@ -0,0 +1,55 @@
...
+As the data toolbar component is a hybrid layout and component, some of its elements are presentational, while some require accessibility support.
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-data-toolbar` | `<div>` | Initiates the toolbar component. **Required** |
+| `.pf-c-data-toolbar__item` | `<div>` | Initiates a the toolbar item. **Required** |
+| `.pf-c-data-toolbar__group` | `<div>` | Initiates a toolbar group. |
+| `.pf-c-data-toolbar__content` | `<div>` | Initiates a toolbar content section. |
```
mcoker: is this required?

src/patternfly/components/DataToolbar/docs/code.md
```diff
@@ -0,0 +1,55 @@
...
+| `.pf-c-data-toolbar__expandable-content` | `<div>` | Initiates a toolbar expandable content section. |
+| `.pf-m-expanded` | `.pf-c-data-toolbar__expandable-content` | Modifies expandable content for the expanded state. |
+| `.pf-c-data-toolbar__divider` | `<div>` | Initiates the toolbar divider. |
```
mcoker: can you remove this?

src/patternfly/components/DataToolbar/docs/data-toolbar-intro.md
mattnolting: Removed from intro section

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,351 @@
...
+  .pf-m-clear {
+    justify-self: end;
+  }
+
+  // If pagination is sibling of align-right
+  .pf-m-align-right ~ * {
+    margin-left: 0;
+  }
+
+  .pf-c-chip-group {
```
mcoker: Should we have a modifier for the chip group item?

src/patternfly/components/DataToolbar/data-toolbar-item-search-filter.hbs
mcoker: Looks like you're allowing for `data-toolbar-item-search-filter--modifier` in this item modifier partial, but not the others (bulk select, chip group, pagination, separator). You also support `data-toolbar-group-toggle-group--modifier` in the toggle group modifier partial. Do you think we should add it to the others for consistency?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,351 @@
...
+  // Filter group
+  .pf-m-filter-group {
+    --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar__group--m-filter-group--spacer);
+
+    > * {
+      --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar__group--m-filter-group--space-items);
+    }
+
+    > * + * {
+      margin-left: -1px;
```
mcoker: sweeet. The only issue with this (and in the input group) is if you focus on an element that's to the left of another element in the group, the focus ring is hidden behind the right element, since the right element overlaps the left and has a higher stacking order since it comes after in the DOM.

Maybe we could use `:focus-within`? Here's a codepen as an example - https://codepen.io/mcoker/pen/BaBKORg. I tried it in the toolbar and needed to add a `z-index`, too, but didn't look at it much further as to why. Not sure that we want to add a `z-index`, which could cause random issues with that element popping out from underneath other elements.

mattnolting: `:focus-within` didn't work well for `.pf-c-input-group` text input. Setting `:focus` on `__item` children worked. My main concern w/setting `z-index` was related to dropdowns within the same toolbar. Setting the `z-index` value to `var(--pf-global--ZIndex--xs)` ensures that a `:focus`ed toolbar element won't appear above a dropdown. Other `z-index`ed elements have higher `z-index`es, but it could still see it causing issues..

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,353 @@
...
+  // reset spacer
+  --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar__item--spacer);
+
+  &:last-child {
+    --pf-c-data-toolbar--spacer: 0;
+  }
+
+  :focus {
+    position: relative;
+    z-index: var(--pf-global--ZIndex--xs);
```
mcoker: need a component var for this one.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,353 @@
...
+  // Expandable content
+  --pf-c-data-toolbar__expandable-content--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__expandable-content--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__expandable-content--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__expandable-content--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__expandable-content--lg--PaddingTop: 0;
+  --pf-c-data-toolbar__expandable-content--lg--PaddingRight: 0;
+  --pf-c-data-toolbar__expandable-content--lg--PaddingBottom: 0;
+  --pf-c-data-toolbar__expandable-content--lg--PaddingLeft: 0;
+  --pf-c-data-toolbar__expandable-content--ZIndex: var(--pf-global--ZIndex--xs);
```
mcoker: Looking at @christiemolloy's [zindex refactor](https://github.com/patternfly/patternfly-next/pull/1901), this component should have a z-index of `--pf-global--ZIndex--xs`. That said, we're going to end up with situations like this - I just added an expanded dropdown after the expanded toolbar in the DOM. I think this is a likely scenario in a table or datalist with an open kebab menu in the first row on mobile.

Screen Shot 2019-08-16 at 1 33 37 PM

I'm not sure how this behaves in react, but I think we should consider closing open menus when you open another menu. I know this happens with dropdowns or selects on the page, but not sure what happens if you have a mix of things like dropdowns, selects, toolbar, options menu, etc. WDYT?

mattnolting: @mcoker Valid point, for sure. The React dropdown closes itself with tab exit, but not with VO exit, even when another menu is opened. I think 100% agree, open menus should close when another menu is open.

src/patternfly/components/DataToolbar/docs/data-toolbar-toggle-group.md
```diff
@@ -0,0 +1,27 @@
...
+| -- | -- | -- |
+| `aria-haspopup="true"` | `.pf-c-data-toolbar__toggle > .pf-c-button` | When expandable content appears above content, `aria-haspopup` should be applied to indicate that focus should be trapped. **Required** |
+
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-reveal{-on-[breakpoint]}` | `.pf-c-data-toolbar__group.pf-m-toggle-group`, `.pf-c-data-toolbar__expandable-content` | Modifies toolbar element to hidden at breakpoint. This selector must be applied consistently to toggle group and expandable content. |
+| `.pf-m-shared-item` | `.pf-c-data-toolbar__item`, `.pf-c-data-toolbar__group` | Identifies a toolbar item or group that appear within toggle group and expandable content. |
```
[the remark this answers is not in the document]
mattnolting: I agree. This is something @mcoker and I talked about as well. Really, this is more of a fallback for Core users as React will likely move DOM elements around to accomplish this. I'll expand on description.

src/patternfly/components/DataToolbar/docs/data-toolbar-spacers-intro.md
```diff
@@ -0,0 +1,3 @@
+<h3 class="Example_heading">Data toolbar spacers</h3>
+
+In some instances, it may be necessary to adjust spacing explicitly where items are being hidden/revealed. For example, if a `.pf-m-toogle-group` is adjacent to an element being hidden/show or hidden/revealed, the spacing may appear to be inconsistent. If possible, rely on modifier values. Available spacer modifiers are `.pf-m-spacer-{none, sm, md, lg}{-on-md, -on-lg, -on-xl}` and `.pf-m-space-items-{none, sm, md, lg}{-on-md, -on-lg, -on-xl}`. These modifiers will overwrite existing modifiers provided by `.pf-c-data-toolbar`.
```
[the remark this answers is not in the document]
mattnolting: `.pf-m-toggle` is part of `.pf-m-toggle-group` and gets hidden, so I think it's best to leave this `.pf-m-toggle-group`.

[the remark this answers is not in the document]
mattnolting: doh

src/patternfly/components/DataToolbar/examples/data-toolbar-selected-filters-example.hbs
```diff
@@ -0,0 +1,75 @@
...
+    {{#> data-toolbar-item data-toolbar-item--modifier="pf-m-shared-item"}}
+      {{#> select id=(concat data-toolbar--id '-select-checkbox-risk-expanded') select--IsCheckboxSelect="true"}}
+        Risk
+      {{/select}}
+    {{/data-toolbar-item}}
+
+    {{> data-toolbar-item-chip-group chip-group--label="Risk" chip-group--id=(concat data-toolbar--id '-group2-')}}
+
+    {{!-- Clear all filters button --}}
+    {{#> data-toolbar-item data-toolbar-item--modifier="pf-m-clear"}}
```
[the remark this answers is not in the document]
mattnolting: Consulted @mceledonia, he says align it right. The button looks misaligned when aligned left. @kybaker @mcarrano wyt?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+
+  // Spacer values
+  --pf-c-data-toolbar__item--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--spacer: var(--pf-c-data-toolbar--spacer--base);
+
+  // Focus state
+  --pf-c-data-toolbar__item--child--focus--ZIndex: var(--pf-global--ZIndex--xs);
+
+  // Item
+  --pf-c-data-toolbar--item--Display: flex;
```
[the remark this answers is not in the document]
mattnolting: nope

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+  --pf-c-data-toolbar--item--Visibility: visible;
```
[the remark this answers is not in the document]
mattnolting: nope

src/patternfly/components/DataToolbar/examples/data-toolbar-selected-filters-example.hbs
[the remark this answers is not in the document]
matthewcarleton: ya I saw that to b/c of the spacing around the button. I wonder if this is showing a larger issue of the type of button. I'd guess it will come up again, it feels like a problem with the button itself if we are trying to hide it by aligning it left, but I understand that for the target area we might be stuck with it.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,356 @@
...
+  // Button group
+  --pf-c-data-toolbar__group--m-button-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-button-group--space-items: var(--pf-global--spacer--sm);
+
+  // Icon group spacer
+  --pf-c-data-toolbar__group--m-filter-group--spacer: var(--pf-c-data-toolbar--spacer--base);
+  --pf-c-data-toolbar__group--m-filter-group--space-items: 0;
+
+  // Overflow menu item
+  --pf-c-data-toolbar--m-overflow-menu--spacer: var(--pf-c-data-toolbar--spacer--base);
```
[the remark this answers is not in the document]
mattnolting: Yeah, the spacing is a little different when `revealed`

mattnolting:
@matthewcarleton updated

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -0,0 +1,357 @@
...
+  --pf-c-data-toolbar__group--spacer: var(--pf-c-data-toolbar--spacer--base);
+
+  // Focus state
+  --pf-c-data-toolbar__item--child--focus--ZIndex: var(--pf-global--ZIndex--xs);
+
+  // Separator
+  --pf-c-data-toolbar--m-separator--spacer: var(--pf-c-data-toolbar__item--spacer);
+  --pf-c-data-toolbar--m-separator--BackgroundColor: var(--pf-global--BorderColor--100);
+  --pf-c-data-toolbar--m-separator--Width: var(--pf-global--BorderWidth--md);
+  --pf-c-data-toolbar--m-separator--Height: var(--pf-global--FontSize--lg);
```
mcoker: This doesn't really follow the var naming convention, except that it follows the selector order in the SCSS. Though ideally in the SCSS, because of how we write BEM modifiers, the modifier should be combined with the selector it applies to, otherwise (since we share modifier class names, and the class name doesn't include the element it applies to) we run into potential conflicts with `.pf-c-data-toolbar .pf-m-separator {}` applying to multiple elements  that share the modifier class name (items, groups, and any other child of the toolbar - dropdowns, etc - that may use that class). And writing the var this way opens us up to conflicting names where, as an example, if we were to add `.pf-m-separator` to `.pf-c-data-toolbar`, the variable name (`--pf-c-data-toolbar--m-separator`) is already taken.

In a perfect world, the `.pf-m-separator` selector would be combined with the element it modifies, so `.pf-c-toolbar__item.pf-m-separator`, then the var name would be `--pf-c-data-toolbar__item--m-separator`. What do you think?

mattnolting: @mcoker updated, how's that look?

mattnolting:
@mcoker This should be more inline with naming convention and compound selector config.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -124,63 +120,81 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
...
 
-.pf-c-data-toolbar__content,
-.pf-c-data-toolbar__group {
-  display: flex;
-  align-items: center;
-}
-
-// Group
+// __group and __item spacer values have to be defined after base styling and modifiers
+/* stylelint-disable no-duplicate-selectors */
```
mcoker: Why is this the case? I'm probably missing it, but I don't see why you couldn't move these up into their respective selectors above.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -189,11 +203,12 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   }
 }
 
-// Item
 .pf-c-data-toolbar__item {
   // reset spacer
   --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar__item--spacer);
 
+  margin-right: var(--pf-c-data-toolbar--spacer);
```
mcoker: This is duplicated in the selector that follows it.

mcoker:
On https://github.com/patternfly/patternfly-next/blob/1f46b717a5c365659997bf79e112a47799adf6a8/src/patternfly/components/DataToolbar/data-toolbar.scss#L216 I think that `:focus` selector is going to cause problems. Here's an example where setting `position: relative` on any focused child will mess up positioning https://codepen.io/mcoker/pen/PoYWpPj

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -38,44 +38,44 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
...
+  --pf-c-data-toolbar__group--m-icon-button-group--spacer: var(--pf-c-data-toolbar__group--spacer);
+  --pf-c-data-toolbar__group--m-icon-button-group--space-items: 0;
 
   // Button group
-  --pf-c-data-toolbar--m-button-group--spacer: var(--pf-c-data-toolbar__group--spacer);
-  --pf-c-data-toolbar--m-button-group--space-items: var(--pf-global--spacer--sm);
+  --pf-c-data-toolbar__group--m-button-group--spacer: var(--pf-c-data-toolbar__group--spacer);
+  --pf-c-data-toolbar__group--m-button-group--space-items: var(--pf-global--spacer--sm);
 
   // Icon group spacer
```
mcoker: Should be "// Filter group spacer"

src/patternfly/components/DataToolbar/data-toolbar.scss
mcoker: Perfect, thanks!

mattnolting:
> I think that :focus selector is going to cause problems. Here's an example where setting position: relative on any focused child will mess up positioning 

@mcoker Let's revisit this and fix the problem globally. Created a new issue here: https://github.com/patternfly/patternfly-next/issues/2184

PR #2137, opened 2019-08-06
fix(page): take main elements grid-area value out of custom property

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2136


mcoker:
You can verify the broken behavior by opening a page demo's codepen link and seeing the layout is broken.

Then you can verify the fix by removing the link to `https://unpkg.com/@patternfly/patternfly/patternfly.min.css` in the codepen's CSS, check out this branch, run `npm run build-patternfly`, and copy the contents of `dist/patternfly.min.css` into the CSS panel in the codepen.

Issue #279, opened 2018-06-01
Spinner 


LHinson:
The spinner component needs introduced. Check out the accessibility concerns https://github.com/twbs/bootstrap/pull/22960

Design Folder: https://drive.google.com/open?id=1ObtKD1QavVL3p1hQ0aIXpoaJCzLHDYDx

Component lifecycle: 
~- [ ] Interaction design requirements~ 
- [ ] Visuals 
- [ ] First checkpoint complete and approved 
- [ ] CSS 
- [ ] React
- [ ] Second checkpoint complete and approved 

Contacts: 
- IxD 
- Visual 
- CSS

bdellasc:
Here's a really basic proposal for the spinner. If there is more needed from a visuals standpoint, please let me know. Maybe small and large variants in addition to this default medium one?

https://redhat.invisionapp.com/share/2FQI0EGQ7NR

bdellasc:
Here is the Sketch source file
https://drive.google.com/open?id=1D0j5GL4NVajwElfS9tuyR22jXv6TsS2q

mcarrano:
@kybaker @mceledonia can you guys take a look?

matthewcarleton:
just had a conversation with @mceledonia about this. He walked me through the design that he came up with.
for reference we also looked at 
https://material.io/design/components/progress-indicators.html#circular-progress-indicators
https://codepen.io/mrrocks/pen/EiplA

rachael-phillips:
@matthewcarleton just to clarify, do you want to take on this issue?

rachael-phillips:
Leaving an explanation here of @mceledonia, @kybaker's and my recent discussion on Spinner. @bdellasc I took the design back to Michael, we had a discussion, and we decided to move forward with a more robust spinner design which should work well across products. Let me know if you have any questions!

bdellasc:
@rachael-phillips thanks for the update. Do we have a sense of when the chosen spinner will be implemented? We have need for one in the short term for our upcoming release. @dlabaj built the simple one I proposed already for our immediate needs/current page. I assume we'd just consume and replace it with the chosen one once it's been implemented.

rachael-phillips:
Hi @bdellasc that's a great question, and yes, that sounds like a good plan. Spinner is on the roadmap to be available on the React side by March 22, and I will keep you updated.

ryelo:
Is there any update on the spinner and if it'll hit the March 22 date?

[...]

PR #2142, opened 2019-08-09
feat(spinner): implements a loading indicator

srambach:
Fixes: #279


bdellasc:
This looks great, @srambach !

SpyTec:
Looks great! I noticed a small difference in the spinner compared to the proposed spinner design. 

The proposed design turns the line smaller as it spins and when it slows down to the starting position it gets bigger again. The PR doesn't have this feature, was this dropped?

andybraren:
I notice my computer's fan spin up quite a bit when viewing the preview page. I'm not super familiar with CSS GPU/hardware acceleration, but when I applied `transform: translateZ(0);` to `.pf-c-spinner` (suggested [here](https://stackoverflow.com/questions/13176746/css-keyframe-animation-cpu-usage-is-high-should-it-be-this-way)) my fans immediately started calming down. Chrome's Task Manager doesn't show a significant change in CPU utilization with this applied, but my computer definitely sounds like it's working less. 🤷‍♂️

`will-change` might be the more modern way of activating hardware acceleration and is [well-supported](https://caniuse.com/#feat=will-change), but [MDN's docs](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) consider its usage to be a last resort.

Does anyone else see similar behavior?

andybraren:
Testing with iOS and desktop Safari, the `.pf-m-sm`'s line seems to disappear while spinning. Maybe something related to the line squashing behavior?

image

Weirdly, the ball color in Safari also seems to be different than the line when the dark theme is applied. Not quite sure why.

image

src/patternfly/components/Spinner/spinner.scss
```diff
@@ -0,0 +1,252 @@
+// Spinner Component
+
+.pf-c-spinner {
+  --pf-c-spinner--AnimationDuration: 1.5s;
+  --pf-c-spinner--AnimationTimingFunction: cubic-bezier(.77, .005, .315, 1);
+
+  // set the overall width and height of the spinner to be consistent with the icon size
+  --pf-c-spinner--Width: var(--pf-global--icon--FontSize--xl);
+  --pf-c-spinner--Height: var(--pf-global--icon--FontSize--xl);
```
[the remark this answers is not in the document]
srambach: I was in a constant internal struggle over all the shared variables. You'd do this then?
```  --pf-c-spinner--Width: var(--pf-global--icon--FontSize--xl);
  --pf-c-spinner--Height: var(--pf-c-spinner--Width);```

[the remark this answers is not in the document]
Comment (SpyTec) [reply src/patternfly/components/Spinner/spinner.scss]: I believe having Width and Height be declared separately by the same variable is better than basing Height off of Width.

Given height is based off of width, if the user wants to change width they'd also change the height as well without wanting to. But the same wouldn't happen the other way around if they changed height instead

If we want to keep the aspect ratio and have it linked, there's always the possibility of using Radius instead of Width and Height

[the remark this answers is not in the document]
Comment (srambach) [reply src/patternfly/components/Spinner/spinner.scss]: Yeah, good point. I personally like using radius or diameter but was trying to play nicely with our variable rules :) Anyone have a problem with defining radius or diameter and using that?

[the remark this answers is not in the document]
Comment (srambach) [reply src/patternfly/components/Spinner/spinner.scss]: For now, I have left it as Width and Height and added a comment noting that they need to be the same.

**Comment by srambach** [comment]
@mceledonia @andybraren On the small version, I took out the narrowing of the line as it zooms around. This seems to fix the problem of the stroke disappearing on Safari. Can you both verify that this is an acceptable solution? Thanks!

**Comment by andybraren** [comment]
Small version looks good to me now in Safari. 👍

**Comment by mcoker** [comment]
This is awesome!! Can you be sure to add this to the [experimental-features.js](https://github.com/patternfly/patternfly-next/blob/master/experimental-features.js) file so that it shows up as an experimental component?

**Comment by srambach** [comment]
Note about performance: I've put this into an application that is doing something very cpu intensive. In this case, the spinner comes apart and skips. I've done some variations with simulated CPU loading, which I will attach. (thank you @jschuler!)
I have found that the problem lies in the narrowing of the stroke width. This is done by animating the box-shadow property, which is in the paint layer. I also tried animating the border, but that is in the layout layer and causes even worse behavior.
I also tried removing the balls that round out the ends of the stroke. When these come loose, the animation looks very bad. However, without the narrowing of the line, they seem to behave well. The previous commit changing their animation to a scale puts them into the composite layer, which is the "good" one.

One possibility is to remove the narrowing of the line. I hate to do this to @mceledonia and also I think it really differentiates the look of this spinner.

Another possibility is to create a variant or modifier that makes a "high performance" spinner that doesn't include the narrowing of the line, but leave the full effect for when the waiting is expected to be based on network activity or for a reason that isn't CPU-bound. IDK if having these two variants is a good idea or not.

[SpinnerVariations.zip](https://github.com/patternfly/patternfly-next/files/3560444/SpinnerVariations.zip)

@mceledonia 😢 @kybaker @bdellasc

**Comment by mceledonia** [comment]
Bummer! Per our convo @srambach I'm thinking the rounded terminals will have a larger impact in terms of making it feel less like the initial design when compared to the stroke width change, so I'd be in favor of dropping the width change if that lets us keep the rounded terminals. Let me know if you disagree though.

File: src/patternfly/components/Spinner/docs/code.md [hunk 2]
```diff
@@ -0,0 +1,25 @@
+## Accessibility
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `role` or `aria` | `pf-c-spinner` |  accessibility notes. |
+| `role="progressbar` | `pf-c-spinner` |  Applying a role of progressbar lets the reader know that this is an indeterminate progress indicator |
+| `aria-valuetext="Loading..."` | `pf-c-spinner` |  This text can be used to say what is loading |
+
+Note: If the spinner is showing the loading in progress of a particular region of a page, the author should use `aria-describedby` to point to the status, and set the `aria-busy` attribute to `true` on the region until it is finished loading.
```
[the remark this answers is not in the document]
mcoker: Should "If the spinner is showing the loading in progress of a particular region of a page" be "If the spinner is showing the loading ~in~ progress of a particular region of a page"?

src/patternfly/components/Spinner/docs/code.md
```diff
@@ -0,0 +1,24 @@
+## Accessibility
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `role="progressbar"` | `pf-c-spinner` |  Indicates to assitive technologies that this is an indeterminate progress indicator. |
```
mcoker:
```suggestion
| `role="progressbar"` | `.pf-c-spinner` |  Indicates to assitive technologies that this is an indeterminate progress indicator. |
```

src/patternfly/components/Spinner/docs/code.md
```diff
@@ -0,0 +1,24 @@
...
+| `aria-valuetext="Loading..."` | `pf-c-spinner` |  Describes content that is being loaded, while it is loading. |
```
mcoker:
```suggestion
| `aria-valuetext="Loading..."` | `.pf-c-spinner` |  Describes content that is being loaded, while it is loading. |
```

src/patternfly/components/Spinner/spinner.scss
```diff
@@ -0,0 +1,210 @@
...
+  --pf-c-spinner__clipper--Width: var(--pf-c-spinner--diameter);
+  --pf-c-spinner__clipper--Height: var(--pf-c-spinner--diameter);
+
+  // Clipper:after (circle) values
+  --pf-c-spinner__clipper--after--BoxShadowColor: var(--pf-c-spinner--Color);
+  --pf-c-spinner__clipper--after--Width: var(--pf-c-spinner--diameter);
+  --pf-c-spinner__clipper--after--Height: var(--pf-c-spinner--diameter);
+
+  // This creates the stroke and needs to match the spinner__ball:after Height and Width
+  --pf-c-spinner__clipper--after--BoxShadowSize: var(--pf-c-spinner--stroke-width);
```
mcoker: should we use `BoxShadowSpreadRadius` here in case we introduce other `box-shadow` values?

src/patternfly/components/Spinner/spinner.scss
```diff
@@ -0,0 +1,210 @@
+// Spinner Component
+
+.pf-c-spinner {
+  --pf-c-spinner--AnimationDuration: 1.5s;
+  --pf-c-spinner--AnimationTimingFunction: cubic-bezier(.77, .005, .315, 1);
+
+  // This conceptual variable is used to set other values all based on the diameter of the spinner
+  --pf-c-spinner--diameter: var(--pf-global--icon--FontSize--xl);
```
mcoker: Should the spinner have a default size in core? Or should we always require use with a modifier, and we can offer a default size in react? If we keep the default size in core, I think we should mention in the documentation the default is `pf-m-xl`.

matthewcarleton: +1 it seems confusing to have a modifier that does nothing, otherwise we should remove the default and require a modifier. I don't feel strongly about either, I guess It could depend on how we've handled this in other places.

src/patternfly/components/Spinner/spinner.scss
```diff
@@ -0,0 +1,210 @@
...
+
+  // Tail ball variables
+  --pf-c-spinner__tail-ball--after--BackgroundColor: var(--pf-c-spinner--Color);
+
+  position: relative;
+  display: inline-block;
+  width: var(--pf-c-spinner--Width);
+  height: var(--pf-c-spinner--Height);
+  text-align: left;
+  animation: pf-animation-spinner-parent calc(var(--pf-c-spinner--AnimationDuration) * 2) var(--pf-c-spinner--AnimationTimingFunction) infinite;
```
matthewcarleton: Nitpic: Should we move `@keyframe` animations to the top, rr maybe have a standard where we declare it before it's used.

src/patternfly/components/Spinner/spinner.scss
```diff
@@ -0,0 +1,210 @@
...
+
+// Elements
+
+// Clipper is a rotating rectangle clipped to one quarter the size of the spinner, starting in the upper right
+.pf-c-spinner__clipper {
+  position: absolute;
+  width: var(--pf-c-spinner__clipper--Width);
+  height: var(--pf-c-spinner__clipper--Height);
+  clip-path: inset(0 0 50% 50%);
+  animation:
```
matthewcarleton: should this all be on one line like the rest of the animations?

src/patternfly/components/Spinner/docs/code.md
```diff
@@ -0,0 +1,24 @@
+## Accessibility
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `role="progressbar"` | `.pf-c-spinner` |  Indicates to assitive technologies that this is an indeterminate progress indicator. |
```
matthewcarleton:
```suggestion
| `role="progressbar"` | `.pf-c-spinner` |  Indicates to assistive technologies that this is an indeterminate progress indicator. |
```

src/patternfly/components/Spinner/docs/code.md
```diff
@@ -0,0 +1,24 @@
...
+
+Note: If the spinner is showing that loading of a particular region of a page is in process, the author should use `aria-describedby` to point to the status, and set the `aria-busy` attribute to `true` on the region until it is finished loading. 
+
+Note: A live region must be present before changing its status in order for the change to be read. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
+
+## Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-spinner` | `<span>` |  Creates a spinner component. The default is an extra large spinner. |
```
matthewcarleton: Should we state that each of these classes are required?

mcoker:
@matthewcarleton I'm inclined to remove the default for now and require a modifier like we do for buttons and badge, particularly since the default here is the XL variation. Otherwise I think we could do like the progress component where the default size is medium, and we have small and large variations.

That will allow us to get it in and maybe we can update the default either in core or just react after users use it and we get feedback there should be a default. For example, with the `<Badge>` in react, the read/unread is a boolean, `isRead` that is set to false by default (so `.pf-m-unread` is applied by default).

Since it's experimental, we should have no problem updating that later.

srambach:
If you don't have a default size, you get no spinner at all if the user doesn't think to add a modifier. I don't think this is ok.
If we don't define all size modifiers, then someone who has customized the default will have no way of creating one in that size without further customization.
It is worth 1 extra line of code to make this resilient.

mcoker:
@srambach that's fine, we can come back to it if we need to. Can you document that XL is the default size?

PR #2143, opened 2019-08-09
Feature/bug fix 1577 fix table checkbox alignment

tqle:
@mcoker - Created the PR. Please let me know if there is anything i missed or do in terms of comments, formatting, etc...  Do I need to create something you can preview this fix outside of this is PR.


src/patternfly/components/Table/table-grid.scss
```diff
@@ -92,7 +92,6 @@
 
   // Table check
   --pf-c-table__check--responsive--MarginLeft: var(--pf-global--spacer--sm);
-  --pf-c-table__check--responsive--MarginTop: #{pf-size-prem(6px)};
```
mcoker: We have to leave variables in place and not rename them, otherwise the change is considered breaking, since a user could have overridden this variable in their local app, and if we remove it, their override no longer applies. Instead of removing it, you can apply a value of "0". You'll also need to add back the `margin-top` declaration that references it.

src/patternfly/components/Table/table.scss
```diff
@@ -303,14 +303,14 @@
...
 // ==================================================================
-.pf-c-table__check {
-  vertical-align: top;
-
-  > * {
-    margin-top: var(--pf-c-table__check--input--MarginTop);
-    vertical-align: top;
-  }
-}
+// FIX #1577: Remove this to fix the table checkbox alignment
```
mcoker: We don't comment out blocks (if it's not needed, you can just remove it), and ideally we would be able to test the change in the PR build preview that is posted as a comment to this PR (https://deploy-preview-2143--pf-next.netlify.com/), and not have to build it locally and uncomment the code. Can you go ahead and uncomment it?

src/patternfly/components/Table/table-grid.scss
tqle: Ok. I've added it back in.

src/patternfly/components/Table/table.scss
tqle: Thanks for the review. I've removed the commented code out.

mattnolting:
Hey @tqle @mcoker, I'm not successfully reproducing the bug. Can you tell me how you were able to do that?

tqle:
Hi @mattnolting.
You can reproduce this off the main page from the Table -> Table with checkboxes and actions example. Just open it to a new window and resize the browser. Or you can open the debug console and switch to an iphone or tablet view and you can see the checkbox and kebab in the responsive view.  This should match the original bug screenshots. https://pf4.patternfly.org/components/Table/examples-full/?component=Table%20with%20checkboxes%20and%20actions.

Also how do I resolve this ci/circleci build to make it pass>  I'm new to Patternfly, so I am not familar with all the build processes. Please let me know what I need to do or correct.

Regards,
Tien

mattnolting:
Hey @tqle, you should be able to rebase, then npm install or npm ci, then force push to resolve circleci issues.

src/patternfly/components/Table/table-grid.scss
```diff
@@ -393,11 +393,10 @@
 
   // Check table cell
   .pf-c-table__check {
-    margin-top: var(--pf-c-table__check--responsive--MarginTop);
```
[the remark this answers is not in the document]
tqle: @mattnolting @mcoke
 If add --pf-c-table__check--responsive--MarginTop) back in do I need to set the variable to 0, will that fix this issue?
The issue with the line heights is that the native checkbox and the kebab button have different heights. Even if you set them in the same line height they will be off.  The align-self is self correcting. If the kebab button height or the checkbox gets replaced by a custom checkbox, you may have the height and line height issue again. Also the button has default padding which, the checkbox does not.  It is also causing it to be off center.  Adding the margin top and vertical align top will change the height again also and may cause things to be off center. If you add outlines to the kebab button and checkbox in your browser you can see why they are in the current pattern fly site.

Just let me know what you need me to fix or change. I'm here to help :)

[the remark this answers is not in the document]
mattnolting: > If add --pf-c-table__check--responsive--MarginTop) back in do I need to set the variable to 0, will that fix this issue?

Yep, that's right 👍

tqle:
@mattnolting @mcoker - Hi. Just wondering if this PR is still valid. Just haven't heard anything back in awhile.  Please let me know if there is any thing else to be reviewed. Thanks.

mcoker:
Hey @tqle my sincerest apologies for the delay. Since reporting the original issue, the spacing has improved to where it's acceptable. As @mattnolting mentioned, the issue with using `align-self: center;` on the check/action in responsive view is that, if the first grid row is very tall, the check/action will be vertically centered within the row, when it should maintain its positioning at the top of the row. Here is a screenshot of that:

Screen Shot 2019-12-12 at 4 31 29 PM

However, I think we could improve the spacing a bit more as it seems there is a discrepancy between the spacing between regular and compact (`pf-m-compact`) tables, as the checkboxes in a compact table are smaller (14px) than in a regular table (16px), but they use the same `margin-top` to offset the spacing.

My suggestion would be to either:

1. Update the checkboxes in the compact tables so that they use the normal font-size and aren't smaller than checkboxes in regular tables. This would also be better for accessibility as the target area of a checkbox is already really small.
  * Side note, I would also like to do this to checkboxes in the `<thead>`, since those checkboxes are smaller than checkboxes in the `<tbody>` of a non-compact table. It seems odd that those checkboxes would be different sizes.
1. Use a different `margin-top` for the checkboxes in compact tables to account for the size difference between compact and regular tables. For example, if the margin-top for checkboxes is currently 6px in a responsive view, increase that to 8px in a compact table to account for the 2px font size difference between regular and compact tables.

Thanks for your work on this, and again apologies for the delay in getting to it. I'll follow up with design about the preferred approach and create a separate issue if we choose to address it.

PR #2146, opened 2019-08-12
Updated release notes to fix bullet placement

rachael-phillips:
This PR updates the release notes to fix the bullet placement.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -16,15 +16,14 @@ Packages released:
 
 ## Chore
 - **A11y:** Ensure SkipToContent sends focus to page content ([#2058](https://github.com/patternfly/patternfly-next/pull/2125))
-- **Build:**
-  - Moved to circle ci. This PR:
+- **Build:** Moved to circle ci
   - Replaces .travis.yml with .circleci/config.yml and split up test steps to allow testing in parallel and nicer Github status checks
   - Fixes our our a11y Selenium script to not have Travis integrations
   - Uploads PR previews to https://surge.sh on every pushed commit (no need for Netlify to also build a preview)
   - Still updates the Github Pages with a new npm run publish:docs target on merges to master ([#2121](https://github.com/patternfly/patternfly-next/pull/212))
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -17,12 +17,12 @@ Packages released:
...
 - **A11y:** Ensure SkipToContent sends focus to page content ([#2058](https://github.com/patternfly/patternfly-next/pull/2125))
 - **Build:**
-- Moved to circle ci. This PR:
- - Replaces .travis.yml with .circleci/config.yml and split up test steps to allow testing in parallel and nicer Github status checks
- - Fixes our our a11y Selenium script to not have Travis integrations
- - Uploads PR previews to https://surge.sh on every pushed commit (no need for Netlify to also build a preview)
- - Still updates the Github Pages with a new npm run publish:docs target on merges to master ([#2121](https://github.com/patternfly/patternfly-next/pull/212))
-- Temporarily disable problematic ci checks ([#2125](https://github.com/patternfly/patternfly-next/pull/2101))
+  - Moved to circle ci. This PR:
+  - Replaces .travis.yml with .circleci/config.yml and split up test steps to allow testing in parallel and nicer Github status checks
```
mcoker: Should these be sub-items of the line above? Or is it OK for them to be in the same hierarchy?

RELEASE-NOTES.md
```diff
@@ -17,12 +17,12 @@ Packages released:
 ## Chore
 - **A11y:** Ensure SkipToContent sends focus to page content ([#2058](https://github.com/patternfly/patternfly-next/pull/2125))
 - **Build:**
```
mcoker: This could change to `**Build: moved to circle ci`, then we could remove the "Moved to circleci. This PR:" line. I don't feel strongly about this, it's fine as is, too.

RELEASE-NOTES.md
rachael-phillips: What do you think makes the most sense? Right now it looks like it lives directly under Build.

rachael-phillips: After I made the first edit, I think it resolves this comment. Let me know if it doesn't!

PR #2164, opened 2019-08-19
fix(select): add top aligned example, docs and fix icon rotation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2163


mcoker:
duplicate of https://github.com/patternfly/patternfly-next/pull/2026/

PR #2166, opened 2019-08-20
fix(build): remove github pages deploy

redallen:
Second attempt to fix the master build since #2160 only partially worked :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,9 +109,3 @@ jobs:
     - run:
         name: Release to NPM
         command: npm run semantic-release
-    - run:
-        name: Avoid Unknown Host for github.com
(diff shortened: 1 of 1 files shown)


dgutride:
@redallen  - do we still need the github pages deploy?

redallen:
I believe the gh-pages is what's deployed to pf4.patternfly.org .

dgutride:
The DNS entry points at pf-next.netlify.com - this only pushes to patternfly.org/patternfly-next and I'm pretty sure we don't link to that anywhere.  Can we turn this off and remove any gh deploy settings from this repo?

redallen:
Yep, you're correct! Need reapproval @dgutride .

PR #2182, opened 2019-08-22
fix(modal): add left aligned footer buttons variation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1599


mcoker:
@mcarrano 

> submit (Save) button should come first, i.e. Save, Cancel.

D'oh! Updated.

> Also something seems to be messed up with the examples as I see the same thing for each example. I assume you are keeping around the right aligned examples as not to break anything, but can you reorganize the workspace to place the left-aligned example first as the default placement? The way it is now, it feels like this is a secondary variant as opposed to the preferred approach.

Currently the left-aligned buttons is just a variation, and not the default, so I listed it as we would any other variation from the default. But I see your point. Do we even want to show an example with right aligned buttons, or should they all simply be updated to be left aligned? Or do we want them all to be left aligned, then we show a new example that illustrates the old/legacy alignment where they're right aligned?

mcarrano:
@mcoker I would vote for only showing left aligned examples, otherwise it feels like we are saying that you have the option to do either. @LHinson do you agree with that?

mcoker:
@mcarrano @mattnolting updated the examples so they all use the left aligned modifier, and added right aligned demo

mcarrano:
Seem to be having a problem with the preview @mcoker . All examples are layering to top of each other.

Screen Shot 2019-08-29 at 10 24 17 AM

mcoker:
@rachael-phillips see @mcarrano's comment ^. That's the workspace problem we discussed yesterday.

@mcarrano there's a bug in the preview where all of the examples for a component/demo all show up on one page, so since these demos all occupy the entire viewport width/height, they're overlapped on top of one another. Here are screenshots of the demos:

# Modal demo

Screen Shot 2019-08-29 at 4 16 18 PM


# Modal demo - content long enough to scroll

Screen Shot 2019-08-29 at 4 16 20 PM

# Modal demo - large

Screen Shot 2019-08-29 at 4 16 23 PM

# Modal demo - legacy footer button alignment (right aligned)

Screen Shot 2019-08-29 at 4 16 25 PM

PR #2186, opened 2019-08-23
feat(wizard): add in-page variation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2073

### DIFF
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -5,4 +5,5 @@
   height: var(--pf-c-divider--Height);
   background-color: var(--pf-c-divider--BackgroundColor);
   border: 0; // removes the default border styling on an hr
+  flex-shrink: 0;
 }
--- a/src/patternfly/components/Page/page-main-wizard.hbs
+++ b/src/patternfly/components/Page/page-main-wizard.hbs
@@ -1,0 +1,6 @@
+<section class="pf-c-page__main-wizard{{#if page-main-wizard--modifier}} {{page-main-wizard--modifier}}{{/if}}"
+  {{#if page-main-wizard--attribute}}
+    {{{page-main-wizard--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</section>
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -84,10 +84,9 @@
   --pf-c-wizard--m-in-page__toggle--md--PaddingLeft: calc(var(--pf-global--spacer--xl) + var(--pf-c-wizard__nav-link--before--Width) + var(--pf-global--spacer--sm));
 
   // Toggle number
-  --pf-c-wizard__toggle-num--before--Top: #{pf-size-prem(2px)};
+  --pf-c-wizard__toggle-num--before--Top: 0;
 
   // Toggle list item
-  --pf-c-wizard__toggle-list-item--first-child--FontSize: var(--pf-global--FontSize--lg);
   --pf-c-wizard__toggle-list-item--not-last-child--MarginRight: var(--pf-global--spacer--sm);
   --pf-c-wizard__toggle-list-item--MarginBottom: var(--pf-global--spacer--xs);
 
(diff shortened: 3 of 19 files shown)


src/patternfly/demos/Wizard/examples/wizard-in-page-example.hbs
```diff
@@ -0,0 +1,136 @@
...
+      {{/nav-list}}
+    {{/nav}}
+  {{/page-sidebar}}
+  {{#> page-main page-main--attribute=(concat 'id="main-content-' page--id '"')}}
+    {{#> page-template-breadcrumb}}
+    {{/page-template-breadcrumb}}
+    {{#> page-template-title}}
+    {{/page-template-title}}
+    {{#> page-main-wizard}}
+      {{#> divider}}{{/divider}}
```
matthewcarleton: just curious why we use this rather than adding a top border to `.pf-c-page__main-wizard`?

mattnolting:
@mcoker I think it looks pretty good. My questions have been asked by @matthewcarleton. I would rather see the changes apply to `.pf-c-wizard` rather than `.pf-m-in-page`, but as long as we revisit during breaking change period, it looks good to me.

src/patternfly/components/Page/page-main-wizard.hbs
```diff
@@ -0,0 +1,6 @@
+<section class="pf-c-page__main-wizard{{#if page-main-wizard--modifier}} {{page-main-wizard--modifier}}{{/if}}"
```
christiemolloy: Did you decide to go with using the page main section here or this new element?

src/patternfly/components/Divider/divider.scss
```diff
@@ -5,4 +5,5 @@
   height: var(--pf-c-divider--Height);
   background-color: var(--pf-c-divider--BackgroundColor);
   border: 0; // removes the default border styling on an hr
+  flex-shrink: 0;
```
christiemolloy: If you're not using the divider component now, do you need this?

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -8,6 +8,11 @@
   --pf-c-wizard--lg--MaxHeight: calc(100% - (var(--pf-global--spacer--2xl) * 2));
   --pf-c-wizard--m-full-width--lg--MaxWidth: auto;
   --pf-c-wizard--m-full-height--lg--Height: 100%;
+  --pf-c-wizard--m-in-page--BoxShadow: none;
```
christiemolloy: I can't completely remember, but didn't we decide that if you want to style a component in another component for one specific use case, then you should style that component in the component that has the new variation. Similar to the case of the notification icon in the page header... instead of making a variation of the button that has a background we style the button/icon in the page component. Maybe I'm wrong, but that seems like a similar case here where you should make the variation of the wizard inside the page. Seems like thats the only place where this variation will occur?

src/patternfly/components/Divider/divider.scss
mcoker: I think we should keep it. It keeps the element from shrinking when used as a flex child between sections since we use `height` (and presumably `width` in a vertical divider) instead of `border` or something like that. [Material's divider](https://material-ui.com/components/dividers/) also sets it.

src/patternfly/components/Wizard/wizard.scss
mcoker: Oh this variation basically removes the wizard from being constrained to a width/height for use in it's own version of a modal. This variation can be used in the page, in a `pf-c-modal-box` or whatever, and will eventually just be the default wizard (in a breaking change release).

mcoker: I left a comment related to this here, specifically around the modifier name. https://github.com/patternfly/patternfly-next/pull/2186#discussion_r318142645

src/patternfly/components/Page/page-main-wizard.hbs
mcoker: New element. To make it work with the main page section would require setting a `flex-basis` or `height`, which could be a breaking change.

mcoker:
@christiemolloy @matthewcarleton  decreased the font size of the first step name in the toggle.

PR #2200, opened 2019-08-28
fix(form): update label alignment

### LINKED ISSUE #2167: Horizontal form label alignment
The labels in a horizontal form group layout have top-padding to align the label text with a form control element.

Screen Shot 2019-08-20 at 9 36 29 AM

This alignment doesn't work when there is something other than a form control element in the horizontal form group.

Screen Shot 2019-08-20 at 9 35 17 AM

Also if the label text is more than a single line, the alignment looks a little off, though I'd like to confirm with design if this is intended or not.

Screen Shot 2019-08-20 at 9 38 57 AM

christiemolloy:
closes #2167 

Spoke with @kybaker about how we should handle form labels. We came to the conclusion that: - 
- for form labels adjacent to a form control with one line, the label should be vertically aligned in the center.
- for form labels adjacent to a form control with two lines, the label should be aligned to the top.

This should allow for these labels to more flexible, especially with font-sizes changing. The downside is that the user will have to add a modifier to change the label to align to the top.

Here is the example in the horizontal form demo: 
Screen Shot 2019-08-28 at 10 05 46 AM

Bootstrap does something similar:
Screen Shot 2019-08-28 at 9 56 51 AM

### DIFF
--- a/src/patternfly/components/Form/docs/code.md
+++ b/src/patternfly/components/Form/docs/code.md
@@ -28,4 +28,5 @@
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
+| `.pf-m-align-top` | `.pf-c-form__label` | Modifies form label to align to the top of its parent. |
 | `.pf-m-inline` | `.pf-c-form__group` | Modifies form group children to be inline (this is primarily for radio buttons and checkboxes). |
--- a/src/patternfly/components/Form/docs/code.md
+++ b/src/patternfly/components/Form/docs/code.md
@@ -28,5 +28,5 @@
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
-| `.pf-m-no-padding` | `.pf-c-form__label` | Removes top padding from the label element for labels adjacent to an element that isn't a form control. |
+| `.pf-m-no-padding-top` | `.pf-c-form__label` | Removes top padding from the label element for labels adjacent to an element that isn't a form control. |
 | `.pf-m-inline` | `.pf-c-form__group` | Modifies form group children to be inline (this is primarily for radio buttons and checkboxes). |
--- a/src/patternfly/components/Form/docs/code.md
+++ b/src/patternfly/components/Form/docs/code.md
@@ -28,5 +28,5 @@
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
-| `.pf-m-align-top` | `.pf-c-form__label` | Modifies form label to align to the top of its parent. |
+| `.pf-m-no-padding` | `.pf-c-form__label` | Removes top padding from the label element for labels adjacent to an element that isn't a form control. |
 | `.pf-m-inline` | `.pf-c-form__group` | Modifies form group children to be inline (this is primarily for radio buttons and checkboxes). |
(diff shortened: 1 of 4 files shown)


mcoker:
If a user has a horizontal group with multiple form control elements, or just a tall element, the label is going to be vertically aligned to the center. Is that OK or would that be breaking?

This is what that will look like if a user just pulls in this PR without adding the `pf-m-align-top` class.

Screen Shot 2019-08-28 at 6 03 45 PM


And personally I think labels adjacent to groups with multiple lines where the first line is a form control element, I would still expect the label text to align with the text in the form control element.

So this (top padding)

Screen Shot 2019-08-28 at 6 05 22 PM

instead of this (no padding)

Screen Shot 2019-08-28 at 6 05 29 PM

christiemolloy:
@mcoker Yeah I think that will be a breaking change, would you recommend that I make this an opt-in solution or release it during the breaking change period?

In the screenshot below is this how you would expect to see it happen? By default the label is vertically-aligned in the center (helper text is outside of `pf-c-form__horizontal-group`) so that it doesn't make the form control taller than it needs to be. For labels adjacent to a text area they can use the modifier `pf-m-align-top` so that is aligned to the top and has a top padding so that the text is aligned horizontally?

Screen Shot 2019-08-30 at 10 57 11 AM

mcoker:
> In the screenshot below is this how you would expect to see it happen?

It is, but that's just my opinion :)

If we went that way, couldn't we just add a `pf-m-no-padding-top` modifier to the existing form labels that removes the top padding when the label is adjacent to something that isn't a form control element?

I was also going to say that we could also just apply the `padding-top` that is applied to labels to both the label and the horizontal form group to the right, then say `.pf-c-form__horizontal-group > .pf-c-form-control:first-child, .pf-c-form__horizontal-group > .pf-c-select:first-child, [etc] { margin-top: -[the padding top of the horizontal group]; }` to offset the padding when one of those components that uses `--pf-global--spacer--form-element` as the top padding is the first child of the horizontal form group. I don't actually think we should do that, as I think it could get tricky and we might end up with a really long list of components in the selector, but that would be one way to make the same adjustment and a user wouldn't need to use a modifier for the label and content in the horizontal form group to align.

christiemolloy:
@mcoker my thoughts on the padding for alignment is that it isn't the most flexible option especially when font-sizes change

@kybaker what are your thoughts on having the label that is aligned to the top, have a padding-top? Or should it just be aligned to the top?

christiemolloy:
@mcoker @mattnolting made updates so that this isn't breaking... adding a modifier for when the label is adjacent to a non-formcontrol element

Screen Shot 2019-09-05 at 1 40 53 PM

Also confirmed with @mcarrano that wrapping the label is fine, but from a design perspective long labels should be avoided.

src/patternfly/components/Form/form.scss
```diff
@@ -69,10 +69,16 @@
         grid-template-columns: var(--pf-c-form--m-horizontal--md__group--GridTemplateColumns);
       }
 
+      /* stylelint-disable */
       .pf-c-form__group .pf-c-form__label {
         padding-top: var(--pf-c-form__label--PaddingTop);
         padding-bottom: 0;
+
+        &.pf-m-no-padding {
+          padding-top: 0;
```
mcoker:
```suggestion
          --pf-c-form__label--PaddingTop: 0;
```

src/patternfly/components/Form/docs/code.md
```diff
@@ -28,4 +28,5 @@
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
+| `.pf-m-no-padding` | `.pf-c-form__label` | Removes top padding from the label element for labels adjacent to an element that isn't a form control. |
```
mcoker: should this be `.pf-m-no-padding-top`?

christiemolloy: updated

src/patternfly/components/Form/form.scss
christiemolloy: updated

mcoker:
There is some extra space under the textarea for some reason...

Screen Shot 2019-09-05 at 1 38 01 PM

Also since we have a helper text example, I wonder if we should remove that and just show helper text in the vertical and horizontal examples.

christiemolloy:
Thanks for reviewing. I think I captured all of your comments and removed the text underneath

PR #2210, opened 2019-08-30
fix(dropdown): add primary toggle variation to dropdown

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1872


src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -27,6 +27,13 @@
   --pf-c-dropdown__toggle--m-plain--Color: var(--pf-global--Color--200);
   --pf-c-dropdown__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
   --pf-c-dropdown__toggle--m-plain--disabled--Color: var(--pf-global--disabled-color--200);
+  --pf-c-dropdown__toggle--m-primary--Color: var(--pf-global--Color--light-100);
+  --pf-c-dropdown__toggle--m-primary--BorderColor: transparent;
```
christiemolloy: Do you think we need a variable for the primary border color? I think we could just unset it here, because we don't want to encourage users to add a border color here, looks like it should never have a border color

PR #2211, opened 2019-08-30
Update issue templates

rachael-phillips:
Requesting this is added as our issue template.


rachael-phillips:
Relevant issue #2211

rachael-phillips:
> I think this should be the same as what we settle on for React.

Hi @redallen yes, it will be the same as the React template! Thanks!

SpyTec:
> What is your product and what release version are you targeting?

Release version of my product seems unnecessary. Would be better to ask about a release window since we can't know that from a version.

There's also nothing indicating severity of the issue, if it's blocking or just low priority

rachael-phillips:
@tlabaj do you want me to update this with @SpyTec 's suggestion?

PR #2229, opened 2019-09-06
fix(data toolbar): format groups correctly inside expandable content

nicolethoen:
Fixes needed for the Data Toolbar React component PR.

### DIFF
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -245,6 +245,11 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     margin: 0;
   }
 
+  .pf-c-data-toolbar__group {
+    display: grid;
+    grid-row-gap: var(--pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap);
+  }
+
   // If item is used within a group, add margin back
   .pf-c-data-toolbar__group .pf-c-data-toolbar__item {
     margin-right: var(--pf-c-data-toolbar--spacer);
@@ -331,6 +336,13 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
         .pf-m-space-items-#{$spacer}#{$breakpoint-name} {
           > * {
             --pf-c-data-toolbar--spacer: #{$spacer-value};
+          }
+
+          &.pf-c-data-toolbar__expandable-content {
+            .pf-c-data-toolbar__group,
+            .pf-c-data-toolbar__item {
+              margin-right: var(--pf-c-data-toolbar--spacer);
+            }
           }
 
           > :last-child {
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -254,11 +254,6 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     display: none;
     visibility: hidden;
   }
-
-  // If item is used within a group, add margin back
-  .pf-c-data-toolbar__group .pf-c-data-toolbar__item {
-    margin-right: var(--pf-c-data-toolbar--spacer);
-  }
 }
 
 .pf-c-data-toolbar__toggle .pf-c-badge {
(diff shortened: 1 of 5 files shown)


src/patternfly/components/DataToolbar/examples/data-toolbar-stacked-example.hbs
```diff
@@ -75,20 +75,35 @@
...
-      {{#> select id=(concat data-toolbar--id '-select-checkbox-status') select--IsCheckboxSelect="true"}}
-        Running
-      {{/select}}
-    {{/data-toolbar-item}}
-    {{#> data-toolbar-item}}
-      {{#> select id=(concat data-toolbar--id '-select-checkbox-type') select--IsCheckboxSelect="true"}}
-        Any
-      {{/select}}
-    {{/data-toolbar-item}}
+      {{#> data-toolbar-group newcontext}}
```
mattnolting: Line [8](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-8af5b5bc5ea0f51d0c1286e2bd8baf7bR8), [18](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-8af5b5bc5ea0f51d0c1286e2bd8baf7bR18), [28](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-8af5b5bc5ea0f51d0c1286e2bd8baf7bR28), here at [78](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-8af5b5bc5ea0f51d0c1286e2bd8baf7bR78), [88](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-8af5b5bc5ea0f51d0c1286e2bd8baf7bR88), and [98](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-8af5b5bc5ea0f51d0c1286e2bd8baf7bR98) would all be
```
{{#> data-toolbar-group newcontext data-toolbar-group--modifier="pf-m-shared-item"}}
```

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -245,9 +245,14 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     margin: 0;
   }
 
-  // If item is used within a group, add margin back
-  .pf-c-data-toolbar__group .pf-c-data-toolbar__item {
-    margin-right: var(--pf-c-data-toolbar--spacer);
+  .pf-c-data-toolbar__group {
+    display: grid;
+    grid-row-gap: var(--pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap);
```
christiemolloy: Trying to catch up on the issue so I may be incorrect, but should we be using this variable here for the `grid-row-gap` property? Looks like the class that its being applied to is `pf-c-data-toolbar__group` rather than `expandable content`

mattnolting: @christiemolloy `.pf-c-data-toolbar__group`'s grid row gap, in this case, should always match that of `--pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap`. We're passing the parent's layout to any descendant `__group` so that `.pf-c-data-toolbar__content` and `.pf-c-data-toolbar__expandable-content`'s structure can match. Expandable content is a single column grid when this applies.

christiemolloy:
Just want to confirm the grid gap between `__content` and `__expandable-content` should be smaller than the gaps between the `__expandable-content`'s?

Screen Shot 2019-09-09 at 1 54 50 PM

christiemolloy:
I think that we should create a new example that shows the expandable-content class, aside from all of the other classes. It took me a while to find out which examples they were in. Maybe 1 example that has two components with `expandable-content` classes and one with `pf-m-expanded` applied and one without?

We can do it outside of the scope of this PR... @mattnolting @nicolethoen @mcoker what do you think? If so I can create a follow up issue

nicolethoen:
@christiemolloy I agree that we should do something to demonstrate the pf-m-expanded state in action. They could be added to the appropriate existing examples (toggle group, select filter, and stacked)

mattnolting:
> Just want to confirm the grid gap between `__content` and `__expandable-content` should be smaller than the gaps between the `__expandable-content`'s?

NICE CATCH! 

at [line 7](https://github.com/patternfly/patternfly-next/pull/2229/files#diff-aa7455b75cbb315661ed0ae6dfb39b10R7) can we update?
```
--pf-c-data-toolbar--GridGap: var(--pf-global--gutter);
```

mattnolting:
> I think that we should create a new example that shows the expandable-content class, aside from all of the other classes. It took me a while to find out which examples they were in. Maybe 1 example that has two components with `expandable-content` classes and one with `pf-m-expanded` applied and one without?

I agree, we should update the examples.

> We can do it outside of the scope of this PR... @mattnolting @nicolethoen @mcoker what do you think? If so I can create a follow up issue

I'm open to either updating the existing examples in this PR, or opening a new issue. The primary purpose of this PR is to unblock @nicolethoen, so I'll leave that up to her..

mcoker:
@christiemolloy would you mind opening a new issue for the expanded demos so we can get this PR in ASAP?

PR #2231, opened 2019-09-07
docs(radio): update the radio component example to single components

### LINKED ISSUE #2221: Update radio component examples
Update the examples so we have a single radio component in each example, and an example to illustrate each state - checked, unchecked, reversed, and disabled. We can show both checked and unchecked in the disabled example since it isn't interactive.

This is just to un-group the radios so it doesn't appear they're in the same group and should have the same `name` so you can toggle between them and any other elements/attributes needed to properly display radio elements in a group in the context of a form. The component examples should just show and document the radio component itself (ie, how to create a radio + label pair).

boaz0:
closes #2221

### DIFF
--- a/src/patternfly/components/Radio/examples/index.js
+++ b/src/patternfly/components/Radio/examples/index.js
@@ -2,13 +2,25 @@ import React from 'react';
 import Documentation from '@siteComponents/Documentation';
 import Example from '@siteComponents/Example';
 import RadioExampleRaw from '!raw!./radio-example.hbs';
+import RadioRevExampleRaw from '!raw!./radio-reversed-example.hbs';
+import RadioWrapExampleRaw from '!raw!./radio-wrap-example.hbs';
+import RadioDisabledExampleRaw from '!raw!./radio-disabled-example.hbs';
+import RadioCheckedExampleRaw from '!raw!./radio-checked-example.hbs';
 import RadioExample from './radio-example.hbs';
+import RadioRevExample from './radio-reversed-example.hbs';
+import RadioWrapExample from './radio-wrap-example.hbs';
+import RadioDisabledExample from './radio-disabled-example.hbs';
+import RadioCheckedExample from './radio-checked-example.hbs';
 import docs from '../docs/code.md';
 
 export const Docs = docs;
 
-export default (props) => {
+export default props => {
   const radioExample = RadioExample();
+  const radioRevExample = RadioRevExample();
+  const radioWrapExample = RadioWrapExample();
+  const radioDisabledExample = RadioDisabledExample();
+  const radioCheckedExample = RadioCheckedExample();
   const headingText = 'Radio';
   const variablesRoot = 'pf-c-radio';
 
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Radio/examples/index.js
```diff
@@ -17,6 +29,18 @@ export default (props) => {
       <Example heading="Radio example" handlebars={RadioExampleRaw}>
         {radioExample}
       </Example>
+      <Example heading="Radio reversed example" handlebars={RadioRevExampleRaw}>
+        {radioRevExample}
+      </Example>
+      <Example heading="Radio label wraps input example" handlebars={RadioWrapExampleRaw}>
+        {radioWrapExample}
+      </Example>
+      <Example heading="Radio checked example" handlebars={RadioCheckedExampleRaw}>
```
mcoker: Can we move this as the 2nd example?

src/patternfly/components/Radio/examples/radio-disabled-example.hbs
```diff
@@ -0,0 +1,4 @@
+{{#> radio}}
+  {{#> radio-input radio-input--attribute='id="radio4" name="exampleRadio5" disabled'}}{{/radio-input}}
+  {{#> radio-label radio-label--modifier="pf-m-disabled" radio-label--attribute='for="radio4"'}}Radio 4 disabled{{/radio-label}}
+{{/radio}}
```
mcoker: Can you add a checked variation in here? I think it's OK to have both in one example since they aren't interactive.

src/patternfly/components/Radio/examples/radio-checked-example.hbs
```diff
@@ -0,0 +1,4 @@
+{{#> radio}}
+  {{#> radio-input radio-input--attribute='id="radio3" name="exampleRadio4" checked'}}{{/radio-input}}
```
mcoker: curious why the `exampleRadio[n]` names don't match the ID?

src/patternfly/components/Radio/examples/radio-checked-example.hbs
```diff
@@ -0,0 +1,4 @@
...
+  {{#> radio-label radio-label--attribute='for="radio3"'}}Radio 3 checked{{/radio-label}}
```
mcoker: I don't think the number in the label text is necessary anymore. We can just use "Radio," "Radio reversed," "Radio label wraps input," etc

boaz0:
Thanks @mcoker - updated the PR. Feel free to review.

mcoker:
@boaz0 oh sorry, I must have miscommunicated. The way you had it before was fine where we have 1 radio per example, except for the disabled example, where I think it's OK to have checked and unchecked since they aren't interactive.

We should have a

* Radio (unchecked - no need to mention it's unchecked)
* Radio checked
* Radio label wraps input (unchecked - no need to mention it's unchecked)
* Radio reversed (unchecked - no need to mention it's unchecked)
* Radio disabled (checked and unchecked in a single example as you have it now)

boaz0:
No problem :smile_cat:

boaz0:
Done

christiemolloy:
For this example I think it should help to add more text like this to demonstrate how the text wraps..
Screen Shot 2019-09-23 at 5 16 17 PM

boaz0:
Thanks @mcoker & @christiemolloy

PR #2248, opened 2019-09-10
fix(data-toolbar): update clear button mods

### LINKED ISSUE #2179: Button alignment in data toolbar
See my comment here: https://github.com/patternfly/patternfly-next/pull/2119#discussion_r315355517

I think this could be worth investigating as it seems like we could see other scenarios where the `pf-c-button.pf-m-link` button doesn't get left aligned b/c the spacing makes it look misaligned. 
Screen Shot 2019-08-22 at 10 55 08 AM

mattnolting:
closes #2179


mcoker:
Looks good! Do we need the `.pf-m-clear` class and docs anymore?

.pf-m-clear | .pf-c-data-toolbar__expandable-content .pf-c-data-toolbar__item | Positions clear all filters button.
-- | -- | --

PR #2255, opened 2019-09-12
fix(typeahead-select): add form wrapper, udpate css

### LINKED ISSUE #2130: Select typeahead does not scale input with component width
With Select's input field it does not scale depending on the component size. Meaning that on smaller sizes it's easier to click the input on smaller sizes than bigger ones.

Images with input field highlighted:
small size

big size

mattnolting:
closes #2130 

- `.pf-c-form-control` `height` property throws off vertical alignment of typeahead text. Setting to `auto` allows for `input` to scale correctly.
- added `form.pf-c-select__toggle-form` to match [PF-React implementation](https://patternfly-react.surge.sh/patternfly-4/components/select/)
Screen Shot 2019-09-12 at 8 42 01 AM
- remove `pf-c-select__toggle.pf-m-typeahead` `padding-top` as this adjusts height of select 
- add `margin-top` to `.pf-c-chip-group` to retain proper spacing

### DIFF
--- a/src/patternfly/components/Select/select-toggle-typeahead.hbs
+++ b/src/patternfly/components/Select/select-toggle-typeahead.hbs
@@ -1,13 +1,15 @@
-{{#if select--ItemIsSelected}}
-	{{#> form-control controlType="input"
-		input="true"
-		form-control--modifier="pf-c-select__toggle-typeahead"
-		form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" value="' select-typeahead--Placeholder '"')}}
-	{{/form-control}}
-{{else}}
-	{{#> form-control controlType="input"
-		input="true"
-		form-control--modifier="pf-c-select__toggle-typeahead"
-		form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
-	{{/form-control}}
-{{/if}}
+{{#> select-toggle-typeahead-form}}
+  {{#if select--ItemIsSelected}}
+    {{#> form-control controlType="input"
+      input="true"
+      form-control--modifier="pf-c-select__toggle-typeahead"
+      form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" value="' select-typeahead--Placeholder '"')}}
+    {{/form-control}}
+  {{else}}
+    {{#> form-control controlType="input"
+      input="true"
+      form-control--modifier="pf-c-select__toggle-typeahead"
+      form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
+    {{/form-control}}
+  {{/if}}
+{{/select-toggle-typeahead-form}}
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -27,8 +27,9 @@
   --pf-c-select__toggle--m-plain--Color: var(--pf-global--Color--100);
   --pf-c-select__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
 
+  // Remove during breaking change
   // When there is a typeahead (input box) adjust the top padding of the wrapper in case it wraps to the second line
-  --pf-c-select__toggle-wrapper--m-typeahead--PaddingTop: var(--pf-global--spacer--xs);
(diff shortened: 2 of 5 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -50,6 +48,10 @@
   --pf-c-select__toggle-typeahead--BorderTop: none;
   --pf-c-select__toggle-typeahead--BorderRight: none;
   --pf-c-select__toggle-typeahead--BorderLeft: none;
+  --pf-c-select__toggle-typeahead--Padding: 0;
```
mcoker: this is unused

src/patternfly/components/Select/select-toggle-typeahead.hbs
```diff
@@ -1,13 +1,15 @@
...
-		form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" value="' select-typeahead--Placeholder '"')}}
-	{{/form-control}}
-{{else}}
-	{{#> form-control controlType="input"
-		input="true"
-		form-control--modifier="pf-c-select__toggle-typeahead"
-		form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
-	{{/form-control}}
-{{/if}}
+<form class="pf-c-select__toggle-form">
```
mcoker: Can you check the indentation in this file? Looks like tabs are creating too many spaces.

src/patternfly/components/Select/select.scss
```diff
@@ -27,9 +27,6 @@
   --pf-c-select__toggle--m-plain--Color: var(--pf-global--Color--100);
   --pf-c-select__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
 
-  // When there is a typeahead (input box) adjust the top padding of the wrapper in case it wraps to the second line
-  --pf-c-select__toggle-wrapper--m-typeahead--PaddingTop: var(--pf-global--spacer--xs);
```
mcoker: Instead of removing this, can we just set it to `0` and create a follow-up issue for a breaking change release to remove the var?

src/patternfly/components/Select/select.scss
```diff
@@ -273,7 +274,10 @@
 
 .pf-c-select__toggle-wrapper {
   display: flex;
+  flex-basis: 0;
```
mcoker: can we use shorthand `flex` for grow/shrink/basis?

src/patternfly/components/Select/select.scss
```diff
@@ -334,6 +334,13 @@
   }
 }
 
+.pf-c-select__toggle-wrapper-form {
+  flex-basis: 0;
+  flex-grow: 1;
+  flex-shrink: 1;
```
mcoker: use `flex` here, too?

src/patternfly/components/Select/select-toggle-typeahead.hbs
mcoker: Can you create a `.hbs` file for this element?

mcoker: @tlabaj in this PR @mattnolting has added a BEM element for the `<form>` inside of the select typeahead toggle to match the react examples. Do you know if that form will always exist? Or would some users exclude it? I remember seeing an issue about nested `<form>` elements in the select component (I think it was with the checkbox select), and how that might be some sort of problem, but I don't remember what came of that.

mcoker: Oh and the class should be `.pf-c-select__toggle-wrapper-form`

src/patternfly/components/Select/select.scss
```diff
@@ -224,16 +226,15 @@
...
     .pf-c-form-control {
       @include pf-text-overflow;
 
       position: relative;
+      height: auto;
     }
 
     // When the input is the first thing then use a negative left padding so it matches the toggle container
-    input:first-child {
+    form:first-child {
```
mcoker:
```suggestion
    .pf-c-select__toggle-wrapper-form:first-child {
```

src/patternfly/components/Select/select.scss
mattnolting: Updated

mattnolting:
@mcoker @christiemolloy updated

mcoker:
Nice work! I noticed it breaks a bit when the chip group wraps, but I think that's more of an edge case and not sure we could fix that without a different approach to how the children are included in the toggle.

Screen Shot 2019-09-17 at 5 58 40 PM

PR #2262, opened 2019-09-12
chore(tabs): remove unused var

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2253


mcoker:
@christiemolloy good question. I was going to say I don't think so, because the variable isn't used anywhere, so it doesn't do anything. The only case I can think where it would be breaking is if the user had written their own declaration and used the var as the value, but I don't think we can worry about that level of customization.

That said, if we look at it like this is like a global variable we have defined but don't use - I might expect a user to use it, so removing it would be breaking IMO.

This is so low priority and not a big deal I think it's safe to close it and just add it to the breaking change queue just to be safe.

PR #2266, opened 2019-09-13
fix(table): unset word-break, add modifiers

### LINKED ISSUE #2205: Bug in table cell word-break

```
.pf-c-table td {
    word-break: break-word;
}
```

### LINKED ISSUE #2325: CSS tables agressively wrap, breaking words apart

The code in question is @

The resulting code is:

```css
.pf-c-table td {
  word-break: break-word;
}
```

...and the CSS should be:

```css
.pf-c-table td {
  overflow-wrap: break-word;
}
```

mattnolting:
fixes #2205 
fixes #2325 

- remove `word-break: break-word;` from `<td>`
- ~~add `.pf-c-table-container` - set `overflow-x: auto:` for when table run out of available space~~
- add `.pf-m-break-word` - sets `word-break: break-word;`
- add `.pf-m-text-overflow` - sets text overflow to ellipsis
- add `.pf-m-nowrap` - sets `white-space: pre`
- removed `--pf-c-table--BorderWidth: 0;` as this causes `overflow-y` to scroll

### DIFF
--- a/src/patternfly/components/Table/docs/code.md
+++ b/src/patternfly/components/Table/docs/code.md
@@ -11,7 +11,7 @@ Applying `role="grid"` to tables enhances accessible interaction while in table 
 Table columns may shift when expanding/collapsing. To address this, set `.pf-m-fit-content`, or assign a width `.pf-m-width-[width]` to the corresponding `<th>` defining the column or `<td>` within the column. Width values are `[10, 15, 20, 25, 30, 35, 40, 45, 50, 60, 70, 80, 90]` or `max`.
 
 ## Table header cells
-By default, all table header cells are set to `white-space: nowrap`. If a `<th>`'s content needs to wrap, apply `.pf-m-wrap`.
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Table/docs/table-simple.md
```diff
@@ -9,6 +9,7 @@
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
+| `.pf-c-table-container` | `<div>` | Initiates a table container element. The table container element controls overflow when `.pf-c-table` runs out of space within its container. |
```
mcoker: this doesn't follow any of our other components, since it's a new component. I'm also not sure we need a component for this. If a user's content overflows the table, they could try and ensure that doesn't happen, or if it does, they may need to write CSS, or we could offer overflow options on the elements that wrap the table, like page sections and/or cards. Just some other ideas...

mattnolting:
@mcoker updated

- @mceledonia signed off on headers wrapping
- added modifiers to control wrapping and text overflow

mattnolting:
holding this until breaking change release

PR #2273, opened 2019-09-16
fix(datatoolbar): add in examples for expandable

### LINKED ISSUE #2237: Data Toolbar: Update examples
We need to create a new example that demonstrates the `pf-m-expanded` state in action. They could be added to the appropriate existing examples (toggle group, select filter, and stacked) or in a new example.. Right now it is hard to find which examples `__expandable-content` is in

Related to: https://github.com/patternfly/patternfly-next/pull/2229

christiemolloy:
closes #2237 

Added two examples to distinguish the difference between the expandable content wrapper, and just added pf-m-expanded to components. @mattnolting @mcoker let me know what you think about just having one over the other, or if you think that we should have both.


mcoker:
I like the expanded elements example, though I think we should check items in the menus to match the badge counts.

I think the expanded wrapper example is kind of confusing, since you would never see that and you have duplicate filters. 

I wonder if we should add a 2nd example after "Data toolbar toggle group (responsive)" and label it "expanded" or "expanded on mobile" or something that shows the content expanded. And/or add "expanded on mobile" to the "Data toolbar selected filters" example, too.

src/patternfly/components/DataToolbar/examples/data-toolbar-expanded-elements-example.hbs
```diff
@@ -0,0 +1,90 @@
+Not Expanded Toolbar
+<br>
+{{#> data-toolbar data-toolbar--id="data-toolbar-not-expanded-example"}}
```
mcoker: duplicate ID in this and the other example

src/patternfly/components/DataToolbar/examples/data-toolbar-expanded-elements-example.hbs
```diff
@@ -0,0 +1,119 @@
...
+        {{#> select id=(concat data-toolbar--id '-select-checkbox-risk-expanded') select--IsCheckboxSelect="true"}}
+          Risk
+        {{/select}}
+       {{/data-toolbar-item}}
+    {{/data-toolbar-group}}
+  {{/data-toolbar-expandable-content}}
+{{/data-toolbar}}
+<br>
+<br>
+Expanded Toolbar
```
mcoker:
```suggestion
Expanded toolbar
```

mcoker:
```suggestion
Toolbar expanded
```

src/patternfly/components/DataToolbar/examples/data-toolbar-expanded-elements-example.hbs
```diff
@@ -0,0 +1,119 @@
+Not Expanded Toolbar
```
mcoker:
```suggestion
Toolbar not expanded
```

mcoker:
In the expanded elements example, in the responsive/expanded state, the status menu isn't expanded and nothing is selected. Also should the "Name" menu be expected in both the content and expanded content?

The other 2 demos lgtm!

christiemolloy:
@mcoker re your comment above, I included the "name" in the expanded area because that was what was implemented in the expanded version of the "selected" example. @mattnolting can you confirm that this is the correct implementation?

Screen Shot 2019-09-25 at 9 24 59 AM
Screen Shot 2019-09-25 at 9 25 03 AM

christiemolloy:
Also @mcoker I only expanded the bottom select because it looks like an incorrect implementation if they're both expanded, what are your thoughts?

Screen Shot 2019-09-25 at 9 29 38 AM

mcoker:
@christiemolloy gotcha, that makes sense. Only thing left is that nothing under "Risk" is selected in the responsive view. Seems like the same 3 items should be selected there as in the non-responsive view.

Screen Shot 2019-09-25 at 8 47 47 AM
Screen Shot 2019-09-25 at 8 47 38 AM

christiemolloy:
Updated @mcoker 

Screen Shot 2019-09-25 at 10 24 55 AM

src/patternfly/components/DataToolbar/examples/data-toolbar-expanded-elements-example.hbs
```diff
@@ -0,0 +1,119 @@
...
+  {{#> data-toolbar-expandable-content data-toolbar-expandable-content--modifier="pf-m-reveal-on-xl" data-toolbar-expandable-content--IsExpanded="true"}}
+    {{> data-toolbar-item-search-filter button--id="expanded-content" data-toolbar-item-search-filter--modifier="pf-m-shared-item"}}
+    {{#> data-toolbar-group data-toolbar-group--modifier="pf-m-filter-group"}}
+      {{#> data-toolbar-item data-toolbar-item--modifier="pf-m-shared-item"}}
+        {{#> select id=(concat data-toolbar--id '-select-checkbox-status-expanded')}}
+          Status
+        {{/select}}
+      {{/data-toolbar-item}}
+      {{#> data-toolbar-item data-toolbar-item--modifier="pf-m-shared-item"}}
+        {{#> select id=(concat data-toolbar--id '-select-checkbox-risk') select--IsCheckboxSelect="true" select--IsChecked="true" select--ItemIsSelected="true" select--IsExpanded="true"}}
```
mattnolting:
```suggestion
        {{#> select id=(concat data-toolbar--id '-select-checkbox-risk-expanded') select--IsCheckboxSelect="true" select--IsChecked="true" select--ItemIsSelected="true" select--IsExpanded="true"}}
```

mattnolting: id throws axe error

christiemolloy: thank you! updated!

PR #2275, opened 2019-09-16
feat(pagination): offer compact variation

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2110

### DIFF
--- a/src/patternfly/components/Pagination/pagination.scss
+++ b/src/patternfly/components/Pagination/pagination.scss
@@ -1,6 +1,7 @@
 .pf-c-pagination {
   // Base
   --pf-c-pagination--child--MarginRight: var(--pf-global--spacer--lg);
+  --pf-c-pagination--m-compact--child--MarginRight: var(--pf-global--spacer--sm);
 
   // Dropdown
   --pf-c-pagination--c-options-menu__toggle--FontSize: var(--pf-global--FontSize--sm);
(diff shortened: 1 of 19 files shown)


src/patternfly/components/Pagination/pagination.scss
```diff
@@ -12,6 +13,7 @@
   --pf-c-pagination__nav--c-button--PaddingLeft: var(--pf-global--spacer--sm);
   --pf-c-pagination__nav--c-button--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-pagination__nav--c-button--FontSize: var(--pf-global--FontSize--md);
+  --pf-c-pagination--m-compact__nav--c-button--MarginLeft: var(--pf-global--FontSize--md);
```
christiemolloy: Why does this use a global FontSize var for the Margin?

christiemolloy:
@mcoker was the plan to update this in any spots or leave that to another issue? It would be good to see an example of this in the table/data list demos to test how it looks

src/patternfly/components/Pagination/pagination.scss
mcoker: Doh! Because I copied and pasted the wrong thing :) Updated.

gdoyle1:
@mcoker Hmm for some reason the compact pagination example looks exactly like the normal pagination. Not sure what's going on! This is what I'm seeing
Screen Shot 2019-09-18 at 10 41 25 PM

mcoker:
@gdoyle1 oh sorry about that, the PR build site has a bug where if you click on an example that opens in a full page, it takes you to a page with all of the examples for that component. To see the compact navigation, just scroll to the very bottom since it's the last example for the component.

mcoker:
@mcarrano there are no other differences between the compact and full pagination other than the spacing in the top level pagination, illustrated in the compact example. The expanded menu, disabled buttons, etc would look the same as they do in the full pagination, which there are already examples of. But if you think it's helpful to include those for reference, I'm happy to create a follow up issue and add them.

gdoyle1:
> @mcarrano there are no other differences between the compact and full pagination other than the spacing in the top level pagination, illustrated in the compact example. The expanded menu, disabled buttons, etc would look the same as they do in the full pagination, which there are already examples of. But if you think it's helpful to include those for reference, I'm happy to create a follow up issue and add them.

I agree with you @mcoker on this one!

PR #2278, opened 2019-09-17
fix(chip-group): add overflow to group

### LINKED ISSUE #2140: Chip group alignment
The "more" alignment is off in Firefox. The alignment in Chrome is OK.

Screen Shot 2019-08-08 at 9 09 12 AM

mattnolting:
closes #2140

### DIFF
--- a/src/patternfly/components/ChipGroup/chip-group.scss
+++ b/src/patternfly/components/ChipGroup/chip-group.scss
@@ -31,14 +31,22 @@
 
 
   &.pf-m-toolbar {
-    & > li {
+    > li {
       display: inline-flex;
       flex-wrap: wrap;
       align-items: center;
       padding: var(--pf-c-chip-group--m-toolbar-PaddingTop) var(--pf-c-chip-group--m-toolbar-PaddingRight) var(--pf-c-chip-group--m-toolbar-PaddingBottom) var(--pf-c-chip-group--m-toolbar-PaddingLeft);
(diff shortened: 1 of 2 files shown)


mcoker:
@matthewcarleton and @christiemolloy do you remember if there was a reason we introduced that overflow chip changed in this PR outside of the chip group as a `<div>`, instead of inside the chip group as another `<li>`? Looks like it was introduced in this PR - https://github.com/patternfly/patternfly-next/pull/1287 and I remember conversation about it, but not specifically why we chose that structure. It looks like we use the structure recommended in this PR for multi-select, any reason we wouldn't use that here that you can think of?

matthewcarleton:
@mcoker I don't recall exactly but reading through that PR it seems like it's related to this comment by @jgiardino 
>Also, I think it makes sense for .pf-m-overflow to be outside the list. I think this is consistent with our plans for Alert Group (i.e. the controls for viewing more alerts are external to the list of alerts).
@jgiardino do you recall anything about this?
We should have one approach for this so this PR LGTM

jgiardino:
These updates look good. I don't have a strong opinion on whether the overflow button is a list item or not, as long as we use a consistent approach for similar patterns in other components.

Is there an issue for making similar updates in react? If so, there's another issue regarding this structure that would be good to work at the same time: https://github.com/patternfly/patternfly-react/issues/1454

PR #2283, opened 2019-09-18
fix(nav): styling updates to dark nav

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2279


src/patternfly/components/Nav/nav.scss
```diff
@@ -4,10 +4,13 @@
   --pf-c-nav__item--m-expanded__toggle-icon--Transform: rotate(90deg);
```
mattnolting: Not part of PR, but this var isn't used

kybaker:
Looks great!

PR #2314, opened 2019-09-27
Add strokeDasharray variable for ChartThreshold

dlabrecq:
Added a variable for the new ChartThreshold React component. Victory's, `strokeDasharray` prop defines the length and spacing between dashes in the lines below.

Screen Shot 2019-09-27 at 9 52 43 AM

Fixes https://github.com/patternfly/patternfly-next/issues/2313

### DIFF
--- a/src/patternfly/patternfly-charts.scss
+++ b/src/patternfly/patternfly-charts.scss
@@ -256,6 +256,9 @@
   // Scatter Chart
   --pf-chart-stack--data--stroke--Width: #{$pf-chart-stack--data--stroke--Width};
 
+  // Threshold
+  --pf-chart-threshold--stroke-dash-array: #{$pf-chart-threshold--stroke-dash-array};
+
   // Tooltip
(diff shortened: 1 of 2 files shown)


dlabrecq:
Any chance we can review / merge? I need this down stream in react-charts

src/patternfly/patternfly-charts.scss
```diff
@@ -256,6 +256,9 @@
   // Scatter Chart
   --pf-chart-stack--data--stroke--Width: #{$pf-chart-stack--data--stroke--Width};
 
+  // Threshold strokeDasharray: '6,6'
```
mcoker: nit - can you remove `'6,6'`? Also if we expect we'll add other "Threshold" vars at some point, I think this comment can just say `// Threshold`

dlabrecq: sure, no problem

PR #2331, opened 2019-10-03
feat(textarea): add horizontal and vertical resize variants

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2251

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -151,4 +151,12 @@ $pf-c-form-control__select--ViewBox: "320" !default;
       --pf-c-form-control--invalid--Background: var(--pf-c-form-control__select--invalid--Background);
     }
   }
+
+  &.pf-m-resize-vertical {
+    resize: vertical;
+  }
+
+  &.pf-m-resize-horizontal {
+    resize: horizontal;
+  }
 }
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -6,16 +6,16 @@ cssPrefix: pf-c-form-control
 
 ## Examples
 ```hbs title=Input
-{{#> form-control controlType="input" input="true" form-control--attribute='disabled type="text" value="disabled" id="textInput1" aria-label="disabled input example"'}}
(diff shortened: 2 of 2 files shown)


src/patternfly/components/FormControl/form-control.scss
```diff
@@ -149,4 +149,12 @@ $pf-c-form-control__select--ViewBox: "320" !default;
...
     }
   }
+
+  &.pf-m-resize-vertical {
+    resize: vertical;
+  }
+
+  &.pf-m-resize-horizontal {
+    resize: horizontal;
+  }
```
mattnolting: Can you add `max-width: 100%` and a `min-width` to this?

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -149,4 +149,12 @@ $pf-c-form-control__select--ViewBox: "320" !default;
       --pf-c-form-control--invalid--Background: var(--pf-c-form-control__select--invalid--Background);
     }
   }
+
+  &.pf-m-resize-vertical {
+    resize: vertical;
+  }
```
mattnolting: Maybe a `min-height` here?

src/patternfly/components/FormControl/form-control.scss
mcoker: @mceledonia should we add a `min-width` to the textarea when you can resize it horizontally, to keep a user from sizing it down too small?

src/patternfly/components/FormControl/form-control.scss
mcoker: @mceledonia what about a `min-height` when you can resize it vertically to prevent sizing it down too short?

mcoker:
@mcarrano would you mind verifying the behavior from our discussion this morning?

mcoker:
@mattnolting re: https://github.com/patternfly/patternfly-next/issues/2357 we discussed that this morning and decided not to impose a min/max-height/width beyond the changes in this PR. @mceledonia @mcarrano do you want to weigh in?

mcarrano:
@mcoker This looks great.

@mattnolting we decided to keep this simple and not impose a max/min width or height. There does not seem to be any harm in letting the user control this other than constraining the resizing as implemented by this PR.

PR #2332, opened 2019-10-04
fix(Table): fix issue with agressive wrap of words

### LINKED ISSUE #2325: CSS tables agressively wrap, breaking words apart

KKoukiou:
This commit replaces the use of word-break: break-word for table columns
to overflow-wrap: break-word instead.

According to MDN, the difference between the two is:

`Note: In contrast to word-break, overflow-wrap will only create a break if an
entire word cannot be placed on its own line without overflowing.`

We obviously do not want to break the words by default — they should
only break when absolutely necessary.

Fixes #2325

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -146,7 +146,7 @@
   background-color: var(--pf-c-table--BackgroundColor);
 
   td {
-    word-break: break-word;
+    overflow-wrap: break-word;
   }
 
   // Standard table row (non-expandable)


mattnolting:
Hey @KKoukiou, thank for working on this. We flagged this issue as a breaking change and will be releasing the fix during breaking change period. The PR is here: https://github.com/patternfly/patternfly-next/pull/2266/files, so I think you can close this.

PR #2334, opened 2019-10-04
feat(chip-group): add closable chip-group

### LINKED ISSUE #2322: Add new variant for closable ChipGroup
Want to create a new variant of a ChipGroup that includes a Close button. For visuals and spacing see: https://github.com/patternfly/patternfly-next/issues/2225 

cc: @rachael-phillips @mcoker

mattnolting:
closes #2322


src/patternfly/components/ChipGroup/examples/ChipGroup.md
```diff
@@ -130,19 +130,51 @@ cssPrefix: pf-c-chip-group
...
+      {{#> chip}}
+        {{#> chip-text chip-text--attribute=(concat 'id="' chip-group--id 'chip_two_toolbar"')}}
+          Chip two
+        {{/chip-text}}
+        {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'aria-labelledby="' chip-group--id 'remove_chip_two_toolbar ' chip-group--id 'chip_two_toolbar" aria-label="Remove" id="' chip-group--id 'remove_chip_two_toolbar"')}}
+          <i class="fas fa-times-circle" aria-hidden="true"></i>
+        {{/button}}
+      {{/chip}}
+    {{/chip-group}}
+    {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'aria-labelledby="' chip-group--id '-label remove_all_chips" aria-label="Remove all chips" id="remove_all_chips"')}}
```
mcoker: What do you think about creating a `chip-group__close-all` element or something like that that we style instead of styling the button component and it needing to be adjacent to the chip group? That would allow for flexibility if someone adds functionality to the button that creates a wrapper for some reason, future proofs for if we add another button that does something else, and we aren't styling a component in a component - which just gets tricky, though styling the margin currently doesn't pose much risk.

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -25,13 +25,15 @@
   --pf-c-chip-group--c-chip--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-chip-group--c-chip--MarginBottom: var(--pf-global--spacer--xs);
 
+  // Close
+  --pf-c-chip-group--button--MarginLeft: var(--pf-global--spacer--xs);
```
christiemolloy: change var: `--pf-c-chip-group--button--MarginLeft` to `--pf-c-chip-group--c-button--MarginLeft`

src/patternfly/components/ChipGroup/chip-group-action.hbs
```diff
@@ -0,0 +1,6 @@
+<button class="pf-c-chip-group__action{{#if chip-group-action--modifier}} {{chip-group-action--modifier}}{{/if}}"
```
christiemolloy: @mcoker what was your reasoning behind wanting to create a new element in the chip group vs using pf-c-button?

mcoker: The previous selector used `.pf-c-chip-group + .pf-c-button` (or something like that) to target the close all button. Ideally we wouldn't use the sibling selector since it can be fragile, and would not style the button (a component in a component), and instead just use a BEM class selector that's part of the chip-group component to style it. Also that doesn't scale if we end up adding another `.pf-c-button` later in that same spot that does something else.

mcoker: Could this be a container for the close all button instead, then we just use a plain button in that, so that we don't have to replicate the button component styles for this element?

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -3,10 +3,10 @@
 .pf-c-chip-group {
   @include pf-t-light; // force the container follow the light theme
 
-  --pf-c-chip-group--MarginRight: calc(var(--pf-c-chip-group--c-chip--MarginRight) * -1);
+  --pf-c-chip-group--MarginRight: var(--pf-global--spacer--xs);
```
mcoker: Do we need this margin? It's creating this extra space on the right here

Screen Shot 2019-10-23 at 12 48 18 PM

If it's to space the group from the action, could we apply a left margin to the action?

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -60,5 +78,37 @@
...
     margin-right: var(--pf-c-chip-group--c-chip--MarginRight);
     margin-bottom: var(--pf-c-chip-group--c-chip--MarginBottom);
+
+    &:last-child {
+      --pf-c-chip-group--c-chip--MarginRight: 0;
+    }
+  }
+}
+
+.pf-c-chip-group__action {
```
mcoker: should we add an "xs" `margin-right` to this to retain the extra space on the right side of the chip-group?

To illustrate, the extra space on the right here

Screen Shot 2019-10-23 at 12 53 57 PM

isn't here

Screen Shot 2019-10-23 at 12 53 53 PM

src/patternfly/components/ChipGroup/chip-group-action.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="pf-c-chip-group__action{{#if chip-group-action--modifier}} {{chip-group-action--modifier}}{{/if}}">
+  {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'aria-labelledby="' chip-group--id '-button ' chip-group--id '-label" aria-label="Close chip group" id="' chip-group--id '-button"')}}
```
mcoker: > 'aria-labelledby="' chip-group--id '-button ' chip-group--id '-label" aria-label="Close chip group" id="' chip-group--id '-button"'

🔥

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -19,7 +19,7 @@
   --pf-c-chip-group__label--PaddingBottom: var(--pf-global--spacer--xs);
   --pf-c-chip-group__label--PaddingLeft: 0;
   --pf-c-chip-group__label--FontSize: var(--pf-global--FontSize--sm);
-  --pf-c-chip-group__label--Maxwidth: #{pf-size-prem(120px)};
+  --pf-c-chip-group__label--MaxWidth: #{pf-size-prem(120px)};
```
mcoker: changing the case [breaks the var](https://codepen.io/mcoker/pen/ExxWMra) :( Can you leave a comment or create an issue to fix this in a breaking change release?

src/patternfly/components/ChipGroup/chip-group-close.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="pf-c-chip-group__close{{#if chip-group-close--modifier}} {{chip-group-close--modifier}}{{/if}}">
```
mcoker: can you add the `chip-group-close--attribute` bit in this file?

mattnolting:
@mcoker @christiemolloy RTG

https://github.com/patternfly/patternfly-next/issues/2390

PR #2337, opened 2019-10-04
chore(select): add example of empty menu with a div

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2333

### DIFF
--- a/src/patternfly/components/Select/select-single.hbs
+++ b/src/patternfly/components/Select/select-single.hbs
@@ -1,24 +1,28 @@
-<ul class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
+<{{#if select-menu--type}}{{select-menu--type}}{{else}}ul{{/if}} class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
   aria-labelledby="{{id}}-label"
   {{#unless select--IsExpanded}}hidden{{/unless}}
   {{#if select--attribute}}
     {{{select--attribute}}}
   {{/if}}>
-  <li><button class="pf-c-select__menu-item" aria-pressed="false">Running</button></li>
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Select/select-single.hbs
```diff
@@ -1,9 +1,12 @@
-<ul class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
+<{{#if select-menu--type}}{{select-menu--type}}{{else}}ul{{/if}} class="pf-c-select__menu{{#if select-menu--modifier}} {{select-menu--modifier}}{{/if}}"
   aria-labelledby="{{id}}-label"
   {{#unless select--IsExpanded}}hidden{{/unless}}
   {{#if select--attribute}}
     {{{select--attribute}}}
   {{/if}}>
+  {{#if select--IsEmptyMenu}}
```
christiemolloy: should this be `select--HasEmptyMenu` ?

mcoker: I dunno... I think `Is[whatever]` seems to show an example variation. Like `IsExpanded`, `IsTypeahead`, or `IsMultiselect` identifies the example. `IsEmptyMenu` shows the empty menu example. I see `HasGroups` or `HasCustomLabel` more as a toggle to include something that is optional, though the line seems blurry and I could totally go either way with it :) Let me know if you'd still like me to update it.

mcoker:
@christiemolloy rebased and updated since this was using the old hbs format

src/patternfly/components/Select/select-single.hbs
christiemolloy: Thanks for the explanation that definitely makes sense from the example side. "Is Empty Menu" just reads strange, but no worries!

christiemolloy:
Approved but there is a conflict

PR #2338, opened 2019-10-07
feat(form): add validated form modifier

### LINKED ISSUE #2320: Add styling for validated text input
Would like to have common styling for text inputs that need to indicate the value has been validated.

Similar to:

image

jeff-phillips-18:
Fixes #2320 

Updates colors used for --pf-global--success-color--100 and --pf-global--success-color--200 for better a11y. See #2320

New Validated input:
image

Updates due to success color changes:

image

image

image

cc @mceledonia @kybaker @serenamarie125

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -91,10 +91,10 @@ cssPrefix: pf-c-form
     {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-comment"')}}
       Comment
     {{/form-label}}
-    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-validated" form-control--attribute=(concat 'value="This is a valid comment"' 'type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
+    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute=(concat 'value="This is a valid comment"' 'type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
     {{/form-control}}
-    {{#> form-helper-text form-helper-text--modifier="pf-m-validated" form-helper-text--attribute=(concat 'id="' form--id '-simple-form-comment-help" aria-live="polite"')}}
-      This is helper text for validated input
+    {{#> form-helper-text form-helper-text--modifier="pf-m-success" form-helper-text--attribute=(concat 'id="' form--id '-simple-form-comment-help" aria-live="polite"')}}
+      This is helper text for success input
     {{/form-helper-text}}
   {{/form-group}}
 {{/form}}
@@ -140,7 +140,7 @@ cssPrefix: pf-c-form
 | `.pf-c-form__actions` | `<div>` | Iniates a row of actions. |
 | `.pf-m-action` | `.pf-c-form__group` | Modifies form group margin-top. |
 | `.pf-m-error` | `.pf-c-form__helper-text`| Modifies text color of helper text. |
-| `.pf-m-validated` | `.pf-c-form__helper-text` | Modifies text color of helper text. |
+| `.pf-m-success` | `.pf-c-form__helper-text` | Modifies text color of helper text. |
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -87,6 +87,16 @@ cssPrefix: pf-c-form
       This is helper text for an invalid input
     {{/form-helper-text}}
   {{/form-group}}
+  {{#> form-group}}
+    {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-comment"')}}
+      Comment
+    {{/form-label}}
+    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-validated" form-control--attribute=(concat 'value="This is a valid comment"' 'type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
+    {{/form-control}}
+    {{#> form-helper-text form-helper-text--modifier="pf-m-validated" form-helper-text--attribute=(concat 'id="' form--id '-simple-form-comment-help" aria-live="polite"')}}
+      This is helper text for validated input
+    {{/form-helper-text}}
+  {{/form-group}}
 {{/form}}
 ```
 
@@ -130,6 +140,7 @@ cssPrefix: pf-c-form
 | `.pf-c-form__actions` | `<div>` | Iniates a row of actions. |
 | `.pf-m-action` | `.pf-c-form__group` | Modifies form group margin-top. |
 | `.pf-m-error` | `.pf-c-form__helper-text`| Modifies text color of helper text. |
+| `.pf-m-validated` | `.pf-c-form__helper-text` | Modifies text color of helper text. |
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -139,8 +139,8 @@ cssPrefix: pf-c-form
 | `.pf-c-form__horizontal-group` | `<div>`| Wraps `.pf-c-form-control` when using `.pf-m-horizontal` on `.pf-c-form` to provide proper spacing for longer labels. |
 | `.pf-c-form__actions` | `<div>` | Iniates a row of actions. |
 | `.pf-m-action` | `.pf-c-form__group` | Modifies form group margin-top. |
-| `.pf-m-error` | `.pf-c-form__helper-text`| Modifies text color of helper text. |
-| `.pf-m-success` | `.pf-c-form__helper-text` | Modifies text color of helper text. |
+| `.pf-m-error` | `.pf-c-form__helper-text`| Modifies text color of helper text for error state. |
+| `.pf-m-success` | `.pf-c-form__helper-text` | Modifies text color of helper text for success state. |
 | `.pf-m-inactive` | `.pf-c-form__helper-text`| Modifies display of helper text to none. |
 | `.pf-m-border` | `.pf-c-form__section` | Modifies form element border-bottom. |
 | `.pf-m-disabled` | `.pf-c-form__label` | Modifies form label to show disabled state. |
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -46,6 +46,7 @@
 
   // States
   --pf-c-form--m-error--Color: var(--pf-global--danger-color--100);
+  --pf-c-form--m-validated--Color: var(--pf-global--success-color--200);
 
 
   display: grid;
@@ -158,13 +159,14 @@
 
 .pf-c-form__helper-text {
   font-size: var(--pf-c-form__helper-text--FontSize);
-
-  &:not(.pf-m-error) {
-    color: var(--pf-c-form__helper-text--Color);
-  }
+  color: var(--pf-c-form__helper-text--Color);
 
   &.pf-m-error {
-    color: var(--pf-c-form--m-error--Color);
+    --pf-c-form__helper-text--Color: var(--pf-c-form--m-error--Color);
+  }
+
+  &.pf-m-validated {
+    --pf-c-form__helper-text--Color: var(--pf-c-form--m-validated--Color);
   }
 
   &.pf-m-inactive {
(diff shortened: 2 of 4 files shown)


jeff-phillips-18:
Updated per comments in #2320 

image

image

image

image

mceledonia:
Looks great. Thanks Jeff!

mcoker:
We will need to hold off on introducing this change until we have a breaking change release as the color changes will likely be visually breaking for anyone currently using `pf-color-light-green-[400/600]` (or the hex equivalents) in their app to represent a success state, and that color will no longer match the success state color in PF components.

jeff-phillips-18:
Removed the color changes introduced by this PR. Those can be addressed separately. Though the validated input text is not _quite_ what we wanted for colors, those can be updated in the separate PR.

image

@serenamarie125

src/patternfly/components/Form/examples/Form.md
```diff
@@ -87,6 +87,16 @@ cssPrefix: pf-c-form
...
+  {{#> form-group}}
+    {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-comment"')}}
+      Comment
+    {{/form-label}}
+    {{#> form-control controlType="input" input="true" form-control--value="This is a valid comment" form-control--modifier="pf-m-validated" form-control--attribute=(concat 'type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
+    {{/form-control}}
+    {{#> form-helper-text form-helper-text--modifier="pf-m-validated" form-helper-text--attribute=(concat 'id="' form--id '-simple-form-comment-help" aria-live="polite"')}}
+      This is helper text for validated input
+    {{/form-helper-text}}
+  {{/form-group}}
```
mcoker: Can you also add an example(s) to the form control component examples? Is this intended to work with selects and textareas, too?

src/patternfly/components/Form/form.scss
```diff
@@ -167,6 +168,10 @@
     color: var(--pf-c-form--m-error--Color);
   }
 
+  &.pf-m-validated {
```
mcoker: I wonder if we should remove the `:not(.pf-m-error)` selector, and just apply `color: var(--pf-c-form__helper-text--Color);` to `.pf-c-form__helper-text`, then overwrite `--pf-c-form__helper-text--Color` in `.pf-m-error` and `.pf-m-validated`. Can you try that and see if it works?

src/patternfly/components/FormControl/form-control.hbs
```diff
@@ -2,4 +2,4 @@
   {{#if form-control--attribute}}
     {{{form-control--attribute}}}
 	{{/if}}
-{{#if input}}>{{else}}>{{> @partial-block}}</{{controlType}}>{{/if}}
+{{#if input}}{{#if form-control--value}} value="{{form-control--value}}"{{/if}}>{{else}}>{{> @partial-block}}</{{controlType}}>{{/if}}
```
mcoker: You shouldn't need this, you can pass `value="foo"` to `form-control--attribute`

Here are some examples

```
src/patternfly//components/Pagination/examples/Pagination.md:      {{> form-control input="true" controlType="input" form-control--attribute='aria-label="Current page" type="number" min="1" max="4" value="1"'}}
src/patternfly//components/Pagination/examples/Pagination.md:      {{> form-control input="true" controlType="input" form-control--attribute='aria-label="Current page" type="number" min="1" max="4" value="1"'}}
src/patternfly//components/Pagination/examples/Pagination.md:      {{> form-control input="true" controlType="input" form-control--attribute='aria-label="Current page" type="number" min="1" max="4" value="1"'}}
src/patternfly//components/ClipboardCopy/examples/ClipboardCopy.md:    {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" value="This is editable" id="text-input-' clipboard-copy--id '" aria-label="Copyable input"')}}
src/patternfly//components/ClipboardCopy/examples/ClipboardCopy.md:    {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'readonly type="text" value="This is read-only" id="text-input-' clipboard-copy--id '" aria-label="Copyable input"')}}
src/patternfly//components/ClipboardCopy/examples/ClipboardCopy.md:    {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" value="This is an editable version of the Copy to Clipboard Component that has an expandable section. Got a lot of text here, need to see all of it? Click that arrow on the left side and check out the resulting expansion." id="text-input-' clipboard-copy--id '" aria-label="Copyable input"')}}
```

src/patternfly/components/Form/form.scss
jeff-phillips-18: seems to work fine as it should. Not sure why the `:not(.pf-m-error)` and then `&.pf-m-error` was there to begin with.

src/patternfly/components/Form/examples/Form.md
jeff-phillips-18: Yes, and done.

src/patternfly/components/FormControl/form-control.hbs
jeff-phillips-18: great, thanks.

src/patternfly/components/Form/form.scss
```diff
@@ -158,15 +159,16 @@
...
 .pf-c-form__helper-text {
   font-size: var(--pf-c-form__helper-text--FontSize);
-
-  &:not(.pf-m-error) {
-    color: var(--pf-c-form__helper-text--Color);
-  }
+  color: var(--pf-c-form__helper-text--Color);
 
   &.pf-m-error {
     color: var(--pf-c-form--m-error--Color);
```
mcoker:
```suggestion
    --pf-c-form__helper-text--Color: var(--pf-c-form--m-error--Color);
```

src/patternfly/components/Form/form.scss
```diff
@@ -158,15 +159,16 @@
...
   }
 
+  &.pf-m-validated {
+    color: var(--pf-c-form--m-validated--Color);
```
mcoker:
```suggestion
    --pf-c-form__helper-text--Color: var(--pf-c-form--m-validated--Color);
```

src/patternfly/components/FormControl/examples/FormControl.md
```diff
@@ -52,6 +55,18 @@ cssPrefix: pf-c-form-control
     <option value="Option 4">The fourth option</option>
   </optgroup>
 {{/form-control}}
+<br /><br />
+{{#> form-control controlType="select" form-control--modifier="pf-m-validated" form-control--attribute='required id="selectExample3" name="selectExample3" aria-label="error state select group example"'}}
```
mcoker:
```suggestion
{{#> form-control controlType="select" form-control--modifier="pf-m-validated" form-control--attribute='required id="selectExample3" name="selectExample3" aria-label="validated state select group example"'}}
```

mcoker:
```suggestion
{{#> form-control controlType="select" form-control--modifier="pf-m-validated" form-control--attribute='id="selectExample3" name="selectExample3" aria-label="validated state select group example"'}}
```

src/patternfly/components/FormControl/examples/FormControl.md
```diff
@@ -67,6 +82,10 @@ Readonly
 Error
 {{/form-control}}
 <br /><br />
+{{#> form-control controlType="textarea"  form-control--modifier="pf-m-validated" form-control--attribute='required name="textarea" id="textareavalidated" aria-label="Validated state textarea example"'}}
```
mcoker:
```suggestion
{{#> form-control controlType="textarea" form-control--modifier="pf-m-validated" form-control--attribute='name="textarea" id="textareavalidated" aria-label="Validated state textarea example"'}}
```

src/patternfly/components/FormControl/examples/FormControl.md
```diff
@@ -15,6 +15,9 @@ cssPrefix: pf-c-form-control
 {{#> form-control controlType="input" input="true" form-control--attribute='required type="text" value="Error" id="textInput4" aria-invalid="true" aria-label="Error state input example"'}}
 {{/form-control}}
 <br /><br />
+{{#> form-control controlType="input" input="true" form-control--modifier="pf-m-validated" form-control--attribute='required type="text" value="Validated" id="textInput4" aria-label="Validated state input example"'}}
```
mcoker:
```suggestion
{{#> form-control controlType="input" input="true" form-control--modifier="pf-m-validated" form-control--attribute='type="text" value="Validated" id="textInput4" aria-label="Validated state input example"'}}
```

mcoker:
@jeff-phillips-18 looks great! One small thing - can you add a line to the documentation for `.pf-m-validated`?

Something like

```
| `.pf-m-validated` | `.pf-c-form-control` | Modifies a form control for the validated state. |
```

Also would you mind fixing these typos? `textarea.pf-m-form-control` should be `textarea.pf-c-form-control`

```
| `.pf-m-resize-vertical` | `textarea.pf-m-form-control` | Modifies a `<textarea>` element to resize vertically. |
| `.pf-m-resize-horizontal` | `textarea.pf-m-form-control` | Modifies a `<textarea>` element to resize horizontally. |
```

jeff-phillips-18:
@mcoker Done. PTAL. Thanks.

src/patternfly/components/Form/examples/Form.md
```diff
@@ -130,6 +140,7 @@ cssPrefix: pf-c-form
 | `.pf-c-form__actions` | `<div>` | Iniates a row of actions. |
 | `.pf-m-action` | `.pf-c-form__group` | Modifies form group margin-top. |
 | `.pf-m-error` | `.pf-c-form__helper-text`| Modifies text color of helper text. |
+| `.pf-m-validated` | `.pf-c-form__helper-text` | Modifies text color of helper text. |
```
christiemolloy: what is the reason for the name being `validated` over something like `success` just because that would match the color name that it is using. I also think that the term "validated" is dependent on each product that the form is in.. i.e. one validation may have passed, so its successful, but it might not have passed a second validation on submit or something... maybe im overthinking it, but with naming i think keeping it as open as possible is better.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -60,6 +64,17 @@ $pf-c-form-control__select--ViewBox: "320" !default;
   --pf-c-form-control--invalid--exclamation--Background: var(--pf-c-form-control--invalid--BackgroundUrl) var(--pf-c-form-control--invalid--BackgroundPosition) / var(--pf-c-form-control--invalid--BackgroundSize) no-repeat;
   --pf-c-form-control--invalid--Background: var(--pf-c-form-control--BackgroundColor) var(--pf-c-form-control--invalid--exclamation--Background);
 
+  // Input m-validated
+  --pf-c-form-control--validated--BorderBottomWidth: var(--pf-global--BorderWidth--md);
+  --pf-c-form-control--validated--PaddingBottom: calc(var(--pf-global--spacer--form-element) - var(--pf-c-form-control--validated--BorderBottomWidth));
+  --pf-c-form-control--validated--BorderBottomColor: var(--pf-global--success-color--100);
+  --pf-c-form-control--validated--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-form-control--validated--BackgroundPosition: calc(100% - var(--pf-c-form-control--PaddingLeft)) var(--pf-c-form-control--PaddingLeft);
```
christiemolloy: I wonder if we could group the BackgroundPosition and BackgroundURL that is set on the validated/invalid modifier to `pf-c-form-control` since its used in both places.

src/patternfly/components/Form/examples/Form.md
mcoker: This works for me. I think "success" covers something being validated, and can be used outside of the context of validated to indicate a success state. And it matches the alert and progress components. @mceledonia @mcarrano what do you think about using a more generic "success" modifier name instead of "validated"?

src/patternfly/components/FormControl/form-control.scss
mcoker: I could go either way with dropping `BackgroundPosition` - both the default (for input and textarea), as well as the ones for the select variation. It is duplicated, but that also allows for flexibility in case we were to use different icons  and wanted different spacing. 

I think we want to keep BackgroundURL unique between the default select, invalid and validated since the function that generates the value takes different arguments between the 3.

src/patternfly/components/Form/examples/Form.md
mcoker: I could also see using "valid" since the error state is styled via `aria-invalid`

jeff-phillips-18: My original thought was this was really only to be used in the case where the input value has been `validated`, not just `valid`. Most input values that are not `invalid` would thus be `valid`. However, `validated` has a different meaning and should only be used in cases where there was some effort involved in validation, not just `yeah, that is a number`.

src/patternfly/components/FormControl/form-control.scss
jeff-phillips-18: Just let me know what we want here.

jeff-phillips-18:
Any decisions?

src/patternfly/components/Form/examples/Form.md
mcarrano: I like calling this 'success' to be consistent with alert states. Agree with @jeff-phillips-18 that 'valid' does not feel right.

mcoker: @christiemolloy updated to be `pf-m-success`

src/patternfly/components/FormControl/form-control.scss
mcoker: It's just an extra variable and allows for more flexibility. I think this is fine to leave as-is unless you disagree @christiemolloy.

mcoker:
@christiemolloy should be good for review now

src/patternfly/components/FormControl/form-control.scss
christiemolloy: yep thats ok with me

PR #2340, opened 2019-10-07
Add size variables for scatter chart

dlabrecq:
Adds size variables to support interactive data points with the `ChartScatter` component in react-charts.

Fixes https://github.com/patternfly/patternfly-next/issues/2339


dlabrecq:
Thank you!

PR #2342, opened 2019-10-07
refactor(data-toolbar): refactor examples to match React implementation

mattnolting:
- removed `.pf-m-shared`
- changed `.pf-m-reveal-{bp}` to `.pf-m-show-{bp}`
- chip groups placement alteration
- `.pf-c-data-toolbar__expandable-content` moved into the `.pf-c-data-toolbar__content` to which it relates
- added padding to toolbar and expandable-content
- added selected filters example
- adds `.pf-m-toggle-group-container`. Handles item/group spacing when `__expandable-content` is present in `.pf-c-data-toolbar__expandable-content` as the `last-child`.
- adds hidden expandable content to toolbar as it appears in React


nicolethoen:
@mcarrano I hope so. I never actually ended up having a conversation with him about that, but I did brainstorm a bit with @mturley and I think it's possible. It'll just require some rewiring. It'll definitely be the most complicated aspect of this update.

mturley:
@nicolethoen let's spend some time pair programming tomorrow if we can, I have some ideas.

mattnolting:
@nicolethoen Would `.pf-c-data-toolbar__group.pf-m-toggle-group-summary` be better, worse, or make no difference if it were a sibling, `.pf-c-data-toolbar__content` instead?

Screen Shot 2019-10-08 at 2 14 57 PM

Screen Shot 2019-10-08 at 2 15 01 PM

nicolethoen:
@mturley That would be great if you have time! Attempt #1 was not altogether successful.

@mattnolting I dont think it would make a huge difference if the .pf-m-toggle-group-summary was a modifier to a .pf-c-datatoolbar__group or to a .pf-c-datatoolbar__content from the react perspective.

christiemolloy:
A few of the examples are missing their expanded examples:

Screen Shot 2019-10-10 at 9 17 18 AM

Screen Shot 2019-10-10 at 9 13 40 AM

christiemolloy:
I haven't seen any designs for this state below, but I am assuming that there should be vertical space between the chip groups. I also wonder where the "clear filters" should be? Should it wrap vertically under the chips at this stage? IMO it looks odd on the right there, maybe @mcarrano you can add some insight here

Screen Shot 2019-10-10 at 9 25 27 AM

src/patternfly/components/DataToolbar/data-toolbar-group.hbs
```diff
@@ -1,6 +1,9 @@
-<div class="pf-c-data-toolbar__group{{#if data-toolbar-group--modifier}} {{data-toolbar-group--modifier}}{{/if}}"
+<div class="pf-c-data-toolbar__group{{#if data-toolbar-group--modifier}} {{data-toolbar-group--modifier}}{{/if}}{{#if data-toolbar-group--IsHidden}} pf-m-hidden{{/if}}"
   {{#if data-toolbar-group--attribute}}
     {{{data-toolbar-group--attribute}}}
+  {{/if}}
+  {{#if data-toolbar-group--IsHidden}}
```
christiemolloy: This looks like its repeated twice

mattnolting:
@mcarrano met with @mceledonia about chip group wrapping. Update has been applied. 

@mcoker @christiemolloy ready for review

mcarrano:
Yes, much better @mattnolting !

christiemolloy:
Still not seeing some of those extra examples that were added to show the expanded state i.e. https://pf4.patternfly.org/documentation/core/experimental/datatoolbar#expanded-elements-responsive, that should fix this problem here:

Screen Shot 2019-10-11 at 1 56 59 PM

Stacked example also isn't responsive:

Screen Shot 2019-10-11 at 1 58 32 PM

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -6,22 +6,28 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
 // Data toolbar
 .pf-c-data-toolbar {
   --pf-c-data-toolbar--BackgroundColor: var(--pf-global--BackgroundColor--100);
-  --pf-c-data-toolbar--GridGap: var(--pf-global--gutter);
+  --pf-c-data-toolbar--Gap: var(--pf-global--gutter);
```
christiemolloy: should this var: `--pf-c-data-toolbar--Gap` be named `--pf-c-data-toolbar--Gutter`

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -30,16 +36,20 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   // Expanded
   --pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap: var(--pf-global--gutter);
 
+  // Chip container
+  --pf-c-data-toolbar--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
```
christiemolloy: It looks like this modifier is on: `pf-c-data-toolbar__content` so the var name should be updated accordingly, same for the name below

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -253,6 +292,10 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   }
 }
 
+.pf-c-data-toolbar__toggle.pf-m-expanded .pf-c-button {
+  color: var(--pf-c-data-toolbar__toggle--expanded--button--Color);
```
christiemolloy: could we update the color in the button (.pf-m-plain) by adding a `pf-m-expanded` modifier like we did to the control modifier. seems like its something we would repeat

nicolethoen:
> Stacked example also isn't responsive:

@christiemolloy the stacked example (and really all the examples) should collapse down to be more responsive if the viewport size changes. You have to drag the whole window to be smaller, not just expand the inspection tools on the right.

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: Using this as a reference, seems like we're moving toward `gap` instead of `gutter`: https://github.com/patternfly/patternfly-next/issues/2336. I could use `row-gap` instead.

mcoker: I think `Gap` works since the primary function is to define the `gap` property, and isn't used as a concept like a gutter.

mattnolting:
> @mattnolting it does appear that the collapsed state of the stacked example is missing. Are you no longer using `.pf-m-reveal-on-{breakpoint}` mods on the `.pf-m-toggle-group` anymore?

@nicolethoen Yes, that's correct. That was one of the changes we discussed. JS to apply `pf-m-hidden` and `hidden` when breakpoint prop is reached. Am I misremembering?

src/patternfly/components/DataToolbar/data-toolbar.scss
christiemolloy: thats cool with me

nicolethoen:
> @nicolethoen Yes, that's correct. That was one of the changes we discussed. JS to apply `pf-m-hidden` and `hidden` when breakpoint prop is reached. Am I misremembering?

nah, sounds good! :)

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -253,6 +292,10 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   }
 }
 
+.pf-c-data-toolbar__toggle.pf-m-expanded .pf-m-plain {
+  color: var(--pf-c-data-toolbar__toggle--m-expanded--m-plain--Color);
```
mcoker:
```suggestion
  color: var(--pf-c-data-toolbar__toggle--m-expanded__m-plain--Color);
```

nicolethoen:
I like it!

mattnolting:
@mcoker @nicolethoen one last update. 

- Added `.pf-c-data-toolbar__expandable-content` (regardless of whether it contains elements or not ) to toolbar examples that use `.pf-m-toggle-group`. This maintains parity w/React implementation and fixes a small bug that relates to `:last-child` spacer value for `__items`/`__group`s. 

- Added `.pf-m-toggle-group-container` that offsets last-child target via `:nth-last-of-type(2)` to reset spacer value.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -237,6 +237,12 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   &.pf-m-chip-container .pf-c-data-toolbar__item {
     margin-top: var(--pf-c-data-toolbar__content--m-chip-container__item--MarginTop);
   }
+
+  &.pf-m-toggle-group-container {
+    > :nth-last-of-type(2) {
```
mcoker: Can you comment the intent of this selector, that it’s removing the spacer from the last thing that isn’t the expandable content?

nicolethoen: +1 ^ 
these changes are fine if they are necessary. I get to break another rule of react to implement but that's sort of the theme with this component, so no big deal at this point.

mattnolting: @nicolethoen I don't yet know that implementing this is necessary, but the selector is there if we need it. Might be a good idea see how the updates affect the React implementation before baking it in. If it's unnecessary, we can remove it later. @mcoker wdyt?

PR #2364, opened 2019-10-11
fix(select): add plain modifier to button

### LINKED ISSUE #2127: Select component's toggle-button element should use plain modifier
`.pf-c-select__toggle-button` is a button component and it's missing a modifier. That should have `.pf-m-plain`.

christiemolloy:
closes #2127


mattnolting:
@christiemolloy @mcoker Should we use `.pf-c-button.pf-m-control` for the button since it shared the same basic styling?

christiemolloy:
Thats a good point, the only thing is that the control button has borders so we would have to override this:

Screen Shot 2019-10-11 at 11 25 10 AM

And the pf-c-button.pf-m-control has a different hover style to the rest of the select

Screen Shot 2019-10-11 at 11 28 10 AM

and if you focus on the select itself, then the entire select is focused but the button isn't so it appears like this? I think if we did `.pf-c-button.pf-m-control` then we would potentially need to move the button outside of the select component

Screen Shot 2019-10-11 at 11 30 06 AM

mcoker:
Oh missed something. Can you update the docs so that

```
.pf-c-select__toggle-button | <button>
```

is

```
.pf-c-select__toggle-button | button.pf-m-plain
```
 similar to the `.pf-c-select__toggle-clear` docs?

We could also update both of those to `button.pf-c-button.pf-m-plain` to be more clear. WDYT?

christiemolloy:
@mcoker updated docs

PR #2382, opened 2019-10-21
feat(font): add monospace stack for opt in redhat font

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2268

To test this, you can pull down the branch and update some component to use `font-family: var(--pf-global--FontFamily--monospace);`, then make sure the computed `font-family` is "liberation mono", then remove `.pf-m-redhat-font` from the `<html>` element and make sure the computed `font-family` is "overpass-mono".


mcoker:
@mceledonia approves

christiemolloy:
Confirmed that it works! All I had to do was add `font-family: var(--pf-global--FontFamily--monospace);` to the pf-c-badge component

Screen Shot 2019-10-24 at 1 18 59 PM

PR #2384, opened 2019-10-22
fix(menus): fix menu group space - dropdown, options menu, app launcher

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2102

Updated our examples to show a divider in a list (as an `<li>`), as the last `<li>` in a list as a way to separate groups (how the react examples use it), and also as an `<hr>` between groups per the new spacing @mceledonia and I discussed.

### DIFF
--- a/src/patternfly/components/AppLauncher/app-launcher-separator.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-separator.hbs
@@ -1,7 +1,10 @@
 <{{#if app-launcher-separator--IsHr}}hr{{else}}li{{/if}} class="pf-c-app-launcher__separator{{#if app-launcher-separator--modifier}} {{app-launcher-separator--modifier}}{{/if}}"
   {{#if app-launcher-separator--attribute}}
 	  {{{app-launcher-separator--attribute}}}
-  {{/if}}>
+  {{/if}}
+  {{#unless app-launcher-separator--IsHr}}
+    role="separator"
+  {{/unless}}>
 {{#unless app-launcher-separator--IsHr}}
 </li>
 {{/unless}}
--- a/src/patternfly/components/AppLauncher/app-launcher.scss
+++ b/src/patternfly/components/AppLauncher/app-launcher.scss
@@ -191,7 +191,6 @@
   transform: var(--pf-c-app-launcher__menu-item-external-icon--Transform); // move icon a bit to appear more visually centered
 }
 
-// remove this block at breaking change period
 .pf-c-app-launcher__separator {
   height: var(--pf-c-app-launcher__separator--Height);
   margin-top: var(--pf-c-app-launcher__separator--MarginTop);
--- a/src/patternfly/components/AppLauncher/app-launcher.scss
+++ b/src/patternfly/components/AppLauncher/app-launcher.scss
@@ -54,8 +54,10 @@
   --pf-c-app-launcher__separator--MarginBottom: var(--pf-global--spacer--sm);
 
   // Groups
-  --pf-c-app-launcher__group--PaddingTop: var(--pf-global--spacer--md);
-  --pf-c-app-launcher__group--first-child--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__group--PaddingTop: 0; // remove at breaking change
+  --pf-c-app-launcher__group--group--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__group--first-child--PaddingTop: 0; // remove at breaking change
+  --pf-c-app-launcher__group-title--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-app-launcher__group-title--PaddingRight: var(--pf-c-app-launcher__menu-item--PaddingRight);
   --pf-c-app-launcher__group-title--PaddingBottom: var(--pf-c-app-launcher__menu-item--PaddingBottom);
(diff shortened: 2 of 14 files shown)


src/patternfly/components/AppLauncher/app-launcher-separator.hbs
```diff
@@ -1,7 +1,10 @@
 <{{#if app-launcher-separator--IsHr}}hr{{else}}li{{/if}} class="pf-c-app-launcher__separator{{#if app-launcher-separator--modifier}} {{app-launcher-separator--modifier}}{{/if}}"
   {{#if app-launcher-separator--attribute}}
 	  {{{app-launcher-separator--attribute}}}
-  {{/if}}>
+  {{/if}}
+  {{#unless app-launcher-separator--IsHr}}
```
christiemolloy: Were you thinking of permanently keeping an app-launcher-separator element? or replacing it with the divider when the divider isn't experimental

christiemolloy:
For this separator element which adds a visual break between the exact same groups but in a different way, is going to be confusing for a user to wrap their head around, and seems a little hacky.. is this the only solution to the problem? If so i think we need clear documentation that if they have more than 1 group then those subsequent groups should nest their divider inside of the group.

Screen Shot 2019-10-23 at 10 38 09 AM

src/patternfly/components/OptionsMenu/options-menu-multiple.hbs
```diff
@@ -15,7 +15,7 @@
     {{/options-menu-menu-item}}
   </ul>
 </li>
-{{#> divider divider--type="li"}}{{/divider}}
+{{#> options-menu-separator}}{{/options-menu-separator}}
```
christiemolloy: Background color should be the same as the separator in the dropdown etc. 

Screen Shot 2019-10-23 at 10 44 59 AM

src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -149,7 +157,7 @@ The dropdown menu can contain either links or buttons, depending on the expected
 | `.pf-c-dropdown__menu` | `<ul>`, `<div>` | Defines the parent wrapper of the menu items. |
 | `.pf-c-dropdown__menu-item` | `<a>` | Defines a menu item that navigates to another page. |
 | `.pf-c-dropdown__menu-item` | `<button>` | Defines a menu item that performs an action on the current page. |
-| `.pf-c-dropdown__separator` | `<div>` | Defines a separator within the menu. |
+| `.pf-c-dropdown__separator` | `<li>`, `<h4>` | Defines a separator within the menu. |
```
christiemolloy: maybe I missed this but where is the separator an `h4` ?

mcoker:
@christiemolloy just to make sure, you're referring to adding a separator at the end of a group as a means to divide 2 groups - as opposed to placing the separator between the groups? If so, I agree, but that's how the react examples look currently, so to not support that use case would be a breaking change since it's one way we show users how to separate groups. I think we should update our examples and remove support for placing the separator at the end of a list in an upcoming breaking change release.

I only added that example here so that we can verify the new spacing works if a user does that. That's not part of the solution, either - the solution focuses on addressing the spacing between groups with and without a divider, I just built in support for the use case mentioned above.

src/patternfly/components/AppLauncher/app-launcher-separator.hbs
mcoker: We'll replace it. Dana mentioned that we shouldn't use or ship experimental features in non-experimental components. I have another issue to update the nav dividers to separators. We can update these components to use the divider once it's moved out of experimental.

src/patternfly/components/OptionsMenu/options-menu-multiple.hbs
mcoker: updated

src/patternfly/components/Dropdown/examples/Dropdown.md
mcoker: typo :) should be `<hr>` - updated

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -356,23 +358,32 @@
...
-// remove this block at breaking change period
 .pf-c-dropdown__separator {
   height: var(--pf-c-dropdown__separator--Height);
   margin-top: var(--pf-c-dropdown__separator--MarginTop);
   margin-bottom: var(--pf-c-dropdown__separator--MarginBottom);
   background-color: var(--pf-c-dropdown__separator--BackgroundColor);
+  border: 0; // for use when separator is an <hr>
+
+  &:last-child {
+    margin-bottom: 0;
```
mattnolting: should you `--pf-c-dropdown__separator--MarginBottom: 0`

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -189,24 +191,32 @@
...
-// remove this block at breaking change period
 .pf-c-app-launcher__separator {
   height: var(--pf-c-app-launcher__separator--Height);
   margin-top: var(--pf-c-app-launcher__separator--MarginTop);
   margin-bottom: var(--pf-c-app-launcher__separator--MarginBottom);
   background-color: var(--pf-c-app-launcher__separator--BackgroundColor);
   border: 0; // for use when separator is an <hr>
+
+  &:last-child {
+    margin-bottom: 0;
```
mattnolting: should you `--pf-c-app-launcher__separator--MarginBottom: 0` as it keeps specificity of `margin-bottom` at 10?

PR #2385, opened 2019-10-22
fix(accordion): add no-box-shadow variation, refactor expanded border

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2376

fixes https://github.com/patternfly/patternfly-next/issues/2375

### DIFF
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -4,6 +4,7 @@
 .pf-c-accordion {
   // accordion
   --pf-c-accordion--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-accordion--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-accordion--BoxShadow: var(--pf-global--BoxShadow--md);
   --pf-c-accordion--PaddingTop: var(--pf-global--spacer--xl);
   --pf-c-accordion--PaddingBottom: var(--pf-global--spacer--xl);
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -81,17 +81,6 @@
     background-color: var(--pf-c-accordion__toggle--BorderLeftColor);
   }
 
-  .pf-c-accordion__toggle-text {
-    @include pf-text-overflow;
-
-    max-width: var(--pf-c-accordion__toggle-text--MaxWidth);
-  }
-
-  .pf-c-accordion__toggle-icon {
-    line-height: var(--pf-c-accordion__toggle-icon--LineHeight);
-    transition: var(--pf-c-accordion__toggle-icon--Transition);
-  }
-
   &.pf-m-expanded {
     --pf-c-accordion__toggle--BorderLeftColor: var(--pf-c-accordion__toggle--m-expanded--BorderLeftColor);
 
@@ -133,6 +122,17 @@
       color: var(--pf-c-accordion__toggle-text--focus--Color);
     }
   }
+}
+
+.pf-c-accordion__toggle-text {
+  @include pf-text-overflow;
+
+  max-width: var(--pf-c-accordion__toggle-text--MaxWidth);
+}
+
+.pf-c-accordion__toggle-icon {
+  line-height: var(--pf-c-accordion__toggle-icon--LineHeight);
+  transition: var(--pf-c-accordion__toggle-icon--Transition);
 }
 
 .pf-c-accordion__expanded-content {
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -136,9 +136,22 @@
 }
 
 .pf-c-accordion__expanded-content {
-  position: relative;
   font-size: var(--pf-c-accordion__expanded-content--FontSize);
   color: var(--pf-c-accordion__expanded-content--Color);
+
+  &.pf-m-fixed {
+    max-height: var(--pf-c-accordion__expanded-content--m-fixed--MaxHeight);
+    overflow-y: auto;
+  }
+
+  &.pf-m-expanded {
+    --pf-c-accordion__expanded-content--BorderLeftColor: var(--pf-c-accordion__expanded-content--m-expanded--BorderLeftColor);
+  }
+}
+
+.pf-c-accordion__expanded-content-body {
+  position: relative;
+  padding: var(--pf-c-accordion__expanded-content-body--PaddingTop) var(--pf-c-accordion__expanded-content-body--PaddingRight) var(--pf-c-accordion__expanded-content-body--PaddingBottom) var(--pf-c-accordion__expanded-content-body--PaddingLeft);
 
   &::before {
     position: absolute;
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Accordion/accordion.scss
```diff
@@ -4,7 +4,6 @@
 .pf-c-accordion {
   // accordion
   --pf-c-accordion--BackgroundColor: var(--pf-global--BackgroundColor--100);
-  --pf-c-accordion--BorderWidth: var(--pf-global--BorderWidth--sm);
```
christiemolloy: are you making this PR a breaking change? or is this considered a bug?

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -57,85 +56,109 @@
...
 
-  .pf-c-accordion__toggle {
-    display: flex;
-    align-items: center;
-    justify-content: space-between;
-    width: 100%;
-    padding: var(--pf-c-accordion__toggle--PaddingTop) var(--pf-c-accordion__toggle--PaddingRight) var(--pf-c-accordion__toggle--PaddingBottom) var(--pf-c-accordion__toggle--PaddingLeft);
-    border: 0;
-    border-left: var(--pf-c-accordion__toggle--m-expanded--BorderWidth) solid var(--pf-c-accordion__toggle--BorderLeftColor);
+  &.pf-m-no-box-shadow {
```
christiemolloy: Would you call this `pf-m-plain` since this modifier basically removes any border/shadow/line around the accordion?

src/patternfly/components/Accordion/accordion.scss
mcoker: This var isn't used anywhere in the component, so it's safe to remove.

src/patternfly/components/Accordion/accordion.scss
mcoker: We could call it something else, but it's just removing the box-shadow at the moment. I don't know if there is anything else we plan on offering in a variation of the accordion without a box shadow.

@mcarrano what do you think, should we give this variation (no box shadow) a more meaningful name? You had also mentioned that you weren't sure why the accordion had a box shadow to begin with. One thing we could do (in a breaking change release) is remove the box shadow from the accordion entirely, and if you want a box shadow, you place it in a card. Though it's worth noting the default box shadow on the accordion is `--pf-global--BoxShadow--md`, and the box shadow on the card is `--pf-global--BoxShadow--sm`.

src/patternfly/components/Accordion/accordion.scss
mcoker: I can add it back and create a follow up issue to remove it in a breaking change release if you can think of a use case where removing it would break someone's design.

src/patternfly/components/Accordion/accordion.scss
mcarrano: I don't really have a good answer @mcoker .  @mceledonia do you know why we put a box shadow around the accordion? I agree that without the shadow, the accordion could be placed in a card to get this effect.

src/patternfly/components/Accordion/accordion.scss
christiemolloy: I'm not sure if I can think of a use case of it breaking someones design, but to be safe I think we've been following that method right? @mcoker

mcoker: I think we did leave it one place before in a similar scenario. I think the use case would be if someone had written their own CSS that used `--pf-c-accordion--BorderWidth`, removing the var would break what they're doing. I'll leave it and leave a comment to remove it at a breaking change release just to be sure.

src/patternfly/components/Accordion/accordion.scss
mceledonia: I believe the accordion had the shadow as a stylistic effect for a particular use case where it existed above other content similar to a popover. 

As for using a card to achieve the same effect, the default card shadow is quite different than the one I see on the accordion (the card's shadow is smaller/tighter), so I would say as long as the card is flexible enough to utilize whatever type of shadow is needed, that should work. 

It makes sense to offer a version without the shadow for usage within a broader context, Kyle worked on this and is out for a while so I will investigate further as I believe this was in collaboration with the digital design team.

mceledonia: > Though it's worth noting the default box shadow on the accordion is `--pf-global--BoxShadow--md`, and the box shadow on the card is `--pf-global--BoxShadow--sm`.

Woops, missed that! Yeah we would want to be sure we cover that as I have a feeling the box shadow was added to cover a use case where this existed above other content, but I can't say for sure. Going to investigate that like I said - but I think offering a modifier to remove it makes sense.

As for naming, I expect this non-shadow version to be used in a larger context like a form or some kind of content layout. I expect the shadow version to be used more or less on it's own, possibly as a result of an action to trigger it to appear. Maybe that can give some ideas on what to name it.

mceledonia: Sorry for the rapid response, but I am thinking if you need the larger shadow for this, you might be able to utilize a popover to get the same effect. Card would potentially work as well like you previously mentioned. In that case, we could remove that shadow altogether.

christiemolloy:
@mcoker I'm seeing this, are you as well?

Screen Shot 2019-10-23 at 9 21 43 AM

mcoker:
@christiemolloy yep! Updated.

mcoker:
Also created https://github.com/patternfly/patternfly-next/issues/2388 to clean up these var names.

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -57,85 +57,109 @@
...
+    bottom: 0;
+    left: 0;
+    width: var(--pf-c-accordion__toggle--m-expanded--BorderWidth);
+    content: "";
+    background-color: var(--pf-c-accordion__toggle--BorderLeftColor);
+  }
 
-      max-width: var(--pf-c-accordion__toggle-text--MaxWidth);
-    }
+  .pf-c-accordion__toggle-text {
```
mattnolting: Should this move out of `__toggle`?

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -57,85 +57,109 @@
...
-    .pf-c-accordion__toggle-icon {
-      line-height: var(--pf-c-accordion__toggle-icon--LineHeight);
-      transition: var(--pf-c-accordion__toggle-icon--Transition);
-    }
+    max-width: var(--pf-c-accordion__toggle-text--MaxWidth);
+  }
 
-    &.pf-m-expanded {
-      --pf-c-accordion__toggle--BorderLeftColor: var(--pf-c-accordion__toggle--m-expanded--BorderLeftColor);
+  .pf-c-accordion__toggle-icon {
```
mattnolting: Same here, should it move out of `__toggle`?

mcoker:
@mattnolting I'm seeing these in use still

```
marvin:patternfly-next cmichael$ grep -ri 'pf-c-accordion__toggle-text--active--FontWeight\|pf-c-accordion__toggle-text--focus--Color' src/patternfly/
src/patternfly//components/Accordion/accordion.scss:  --pf-c-accordion__toggle-text--active--FontWeight: var(--pf-global--FontWeight--semi-bold);
src/patternfly//components/Accordion/accordion.scss:  --pf-c-accordion__toggle-text--focus--Color: var(--pf-global--link--Color);
src/patternfly//components/Accordion/accordion.scss:      font-weight: var(--pf-c-accordion__toggle-text--active--FontWeight);
src/patternfly//components/Accordion/accordion.scss:      color: var(--pf-c-accordion__toggle-text--focus--Color);
```

Let me know if you still think they need to be removed. Added the other vars to https://github.com/patternfly/patternfly-next/issues/2388 so we can remove them in a breaking change release.

PR #2387, opened 2019-10-22
fix(nav): update examples so simple-list is only used in expandable nav

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2272


mcarrano:
Looks good to me. I approve.

PR #2396, opened 2019-10-24
fix(releasenotes): update release notes for 2019.09 release

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2395

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,20 +4,64 @@ section: root
 hideTOC: true
 ---
 
+# 2019.09 release notes (2019-10-29)
+Packages released:
+- @patternfly/patternfly@v2.40.2
+
+## Components
+- **Charts:**
+  - Added strokeDasharray variable for ChartThreshold ([#2314](https://github.com/patternfly/patternfly-next/pull/2314))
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -4,20 +4,64 @@ section: root
 hideTOC: true
 ---
 
+# 2019.09 release notes (2019-10-29)
+Packages released:
+- @patternfly/patternfly@v2.40.1
```
dgutride: I believe this should be 2.40.2 to be in sync with the build that this will be released in.

mcoker: Updated, and also updated commit message so it will trigger a release.

mcoker:
@rachael-phillips updated the messages to be past tense

RELEASE-NOTES.md
```diff
@@ -4,20 +4,64 @@ section: root
 hideTOC: true
 ---
 
+# 2019.09 release notes (2019-10-29)
+Packages released:
+- @patternfly/patternfly@v2.40.2
+
+## Components
+- **Charts:**
```
redallen: Usually we alphabetize these, but I'll worry about it later.

PR #2397, opened 2019-10-24
refactor(data-toolbar): move and wrap chips

### LINKED ISSUE #2394: DataToolbar: reorganize chip groups in expandable content
This should only affect how chip groups are displayed in the expandable content. They should remain in their own content row when the viewport is wider than the consumer provided breakpoint.

- Move all chip groups to the bottom of the expandable content, rather than displaying each of them directly under their associated filter input.

- Display chip groups in line and have them wrap as needed, rather than displaying each one as a block element

mattnolting:
closes #2394

### DIFF
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -36,8 +36,8 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   --pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap: var(--pf-global--gutter);
 
   // Chip container
-  --pf-c-data-toolbar--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
-  --pf-c-data-toolbar--m-chip-container__item--MarginTop: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__group--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-data-toolbar__group--m-chip-container--item--MarginTop: var(--pf-global--spacer--md);
 
   // Base spacer - shared value
   --pf-c-data-toolbar--spacer--base: var(--pf-global--spacer--md);
@@ -294,7 +294,7 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   .pf-m-chip-container {
     display: flex;
     flex-wrap: wrap;
-    margin-top: var(--pf-c-data-toolbar--m-chip-container--MarginTop);
+    margin-top: var(--pf-c-data-toolbar__group--m-chip-container--MarginTop);
     margin-left: 0;
 
     .pf-c-data-toolbar__item,
@@ -303,7 +303,7 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     }
 
     .pf-c-data-toolbar__item {
-      margin-top: var(--pf-c-data-toolbar--m-chip-container__item--MarginTop);
+      margin-top: var(--pf-c-data-toolbar__group--m-chip-container--item--MarginTop);
     }
   }
 }
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -298,6 +298,13 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     display: none;
     visibility: hidden;
   }
+
+  .pf-c-data-toolbar__content.pf-m-chip-container {
+    --pf-c-data-toolbar__content--PaddingRight: 0;
+    --pf-c-data-toolbar__content--PaddingLeft: 0;
+
+    margin-left: 0;
+  }
 }
 
 .pf-c-data-toolbar__toggle.pf-m-expanded .pf-c-button.pf-m-plain {
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -36,8 +36,8 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   --pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap: var(--pf-global--gutter);
 
   // Chip container
-  --pf-c-data-toolbar__group--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
-  --pf-c-data-toolbar__group--m-chip-container--item--MarginTop: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__expandable-content--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-data-toolbar__expandable-content--m-chip-container__item--MarginTop: var(--pf-global--spacer--md);
 
   // Base spacer - shared value
   --pf-c-data-toolbar--spacer--base: var(--pf-global--spacer--md);
@@ -294,7 +294,7 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   .pf-m-chip-container {
     display: flex;
     flex-wrap: wrap;
-    margin-top: var(--pf-c-data-toolbar__group--m-chip-container--MarginTop);
+    margin-top: var(--pf-c-data-toolbar__expandable-content--m-chip-container--MarginTop);
     margin-left: 0;
 
     .pf-c-data-toolbar__item,
@@ -303,7 +303,7 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     }
 
     .pf-c-data-toolbar__item {
-      margin-top: var(--pf-c-data-toolbar__group--m-chip-container--item--MarginTop);
+      margin-top: var(--pf-c-data-toolbar__expandable-content--m-chip-container__item--MarginTop);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


nicolethoen:
Are we sure that the chip groups should be in a `.pf-c-data-toolbar__content`? Right now we have a rule that every content row has its own expandable content (except for the , so putting a content row in an expandable content feels weird and means I'll have to code this content row very differently than the others. Would it make more sense for it to be a `.pf-c-data-toolbar__group`?

I'm wondering if the same is true for the content row under the toolbar used for displaying the chip groups. should it be a content row containing a group using the `.pf-m-chip-container` modifier rather than putting the modifier on the content row? 

Would that semantically make a little more sense than we have it now? I'm just processing how I'm going to implement these changes.

mattnolting:
> Are we sure that the chip groups should be in a .pf-c-data-toolbar__content INSIDE the expandable content?

@nicolethoen my thought was hardcoding the chip-group wrapper with `__content.pf-m-chip-container`. I can make the wrapper whatever you want it to be. It will share the layout CSS currently assigned to `.pf-m-chip-container`.

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -36,8 +36,8 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   --pf-c-data-toolbar__expandable-content--m-expanded--GridRowGap: var(--pf-global--gutter);
 
   // Chip container
-  --pf-c-data-toolbar__content--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
-  --pf-c-data-toolbar__content--m-chip-container__item--MarginTop: var(--pf-global--spacer--md);
+  --pf-c-data-toolbar__group--m-chip-container--MarginTop: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-data-toolbar__group--m-chip-container--item--MarginTop: var(--pf-global--spacer--md);
```
mcoker: why wouldn't this var and the one above it be

`--pf-c-data-toolbar__expandable-content--m-chip-container--MarginTop` and `--pf-c-data-toolbar__expandable-content--m-chip-container__item--MarginTop`?

mattnolting: @mcoker updated

src/patternfly/components/DataToolbar/examples/DataToolbar.md
```diff
@@ -291,7 +291,7 @@ The `.pf-m-toggle-group` controls when, and at which breakpoint, filters will be
 | -- | -- | -- |
 | `.pf-m-show{-on-[breakpoint]}` | `.pf-c-data-toolbar__group.pf-m-toggle-group`, `.pf-c-data-toolbar__expandable-content` | Modifies toolbar element visibility at breakpoint. This selector must be applied consistently to toggle group and expandable content. |
 | `.pf-m-toggle-group-container` | `.pf-c-data-toolbar__content` | Modifies the data toolbar content for toggle group, expandable content layout. |
-| `.pf-m-chip-container` | `.pf-c-data-toolbar__content` | Modifies the data toolbar content for selected filters applied filters layout. |
+| `.pf-m-chip-container` | `.pf-c-data-toolbar__group` | Modifies the data toolbar group for selected filters applied filters layout. |
```
mcoker:
```suggestion
| `.pf-m-chip-container` | `.pf-c-data-toolbar__expandable-content` | Modifies the data toolbar expandable content for selected filters applied filters layout. |
```

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -298,6 +290,22 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     display: none;
     visibility: hidden;
   }
+
+  .pf-m-chip-container {
```
mcoker: I should have caught this before, but I think here and in a few other places in this file, we should make sure the modifier selector applies to the element it modifies, since we reuse modifier class names. `.pf-m-chip-container` could be a modifier from another component that would inherit styles we don't want here. I think we updated `.pf-m-hidden` in this way here https://github.com/patternfly/patternfly-next/blob/95ed2defabe187ccb6898cdecbb19a0cc07230ff/src/patternfly/components/DataToolbar/data-toolbar.scss#L241-L251

Can we update `.pf-m-modifier`, `.pf-m-label` ([link](https://github.com/patternfly/patternfly-next/blob/95ed2defabe187ccb6898cdecbb19a0cc07230ff/src/patternfly/components/DataToolbar/data-toolbar.scss#L289)), and the modifiers in the SCSS loops?

src/patternfly/components/DataToolbar/examples/DataToolbar.md
```diff
@@ -309,13 +309,15 @@ The `.pf-m-toggle-group` controls when, and at which breakpoint, filters will be
...
         {{/data-toolbar-item}}
-        {{> data-toolbar-item-chip-group chip-group--label="Status" chip-group--id=(concat data-toolbar--id '-group1-')}}
         {{#> data-toolbar-item}}
           {{#> select id=(concat data-toolbar--id '-select-checkbox-risk-expanded') select--IsCheckboxSelect="true" select--ItemIsSelected="true"}}
             Risk
           {{/select}}
         {{/data-toolbar-item}}
-        {{> data-toolbar-item-chip-group chip-group--label="Risk" chip-group--id=(concat data-toolbar--id '-group2-')}}
+        {{#> data-toolbar-group data-toolbar-group--modifier="pf-m-chip-container"}}
+          {{> data-toolbar-item-chip-group chip-group--label="Status" chip-group--id=(concat data-toolbar--id '-group1-')}}
```
mcoker: do you need the trailing `-`?

```suggestion
          {{> data-toolbar-item-chip-group chip-group--label="Status" chip-group--id=(concat data-toolbar--id '-group1')}}
```

src/patternfly/components/DataToolbar/examples/DataToolbar.md
```diff
@@ -309,13 +309,15 @@ The `.pf-m-toggle-group` controls when, and at which breakpoint, filters will be
...
+          {{> data-toolbar-item-chip-group chip-group--label="Risk" chip-group--id=(concat data-toolbar--id '-group2-')}}
```
mcoker: same here

```suggestion
          {{> data-toolbar-item-chip-group chip-group--label="Risk" chip-group--id=(concat data-toolbar--id '-group2')}}
```

mcoker: looks like this is in a few other spots, too, I won't tag 'em all unless you want me to

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -298,6 +290,22 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
...
+    display: flex;
+    flex-wrap: wrap;
+    margin-top: var(--pf-c-data-toolbar__expandable-content--m-chip-container--MarginTop);
```
mcoker: Sorry about this, I was reading the selector wrong before and had you update it incorrectly. I think the way you had it before was fine, though technically it should be this if we don't want to run into CSS var naming conflicts

`--pf-c-data-toolbar__expandable-content__group--m-chip-container--MarginTop`

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -298,6 +290,22 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
...
+    margin-left: 0;
+
+    .pf-c-data-toolbar__item,
+    .pf-c-data-toolbar__group {
+      margin-right: var(--pf-c-data-toolbar--spacer);
+    }
+
+    .pf-c-data-toolbar__item {
+      margin-top: var(--pf-c-data-toolbar__expandable-content--m-chip-container__item--MarginTop);
```
mcoker: and this to `--pf-c-data-toolbar__expandable-content__group--m-chip-container__item--MarginTop`

mcoker:
is this chip container group in the right spot? From the CSS, I thought it would only exist in expandable content:

https://github.com/patternfly/patternfly-next/blame/7b9d47546225cd710311d0bab2a8a4007a41169f/src/patternfly/components/DataToolbar/examples/DataToolbar.md#L317-L320

src/patternfly/components/DataToolbar/examples/DataToolbar.md
mattnolting: the trailing dash constructs a complex `id` correctly

src/patternfly/components/DataToolbar/examples/DataToolbar.md
mattnolting: same here, the trailing dash constructs a complex `id` correctly

mattnolting:
> is this chip container group in the right spot? From the CSS, I thought it would only exist in expandable content:
> 
> https://github.com/patternfly/patternfly-next/blame/7b9d47546225cd710311d0bab2a8a4007a41169f/src/patternfly/components/DataToolbar/examples/DataToolbar.md#L317-L320

Yes, it moves locations

> is this chip container group in the right spot? From the CSS, I thought it would only exist in expandable content:
> 
> https://github.com/patternfly/patternfly-next/blame/7b9d47546225cd710311d0bab2a8a4007a41169f/src/patternfly/components/DataToolbar/examples/DataToolbar.md#L317-L320

@mcoker Nope, not in the right spot. Updated.

nicolethoen:
So just to be clear, the last couple pushes removed the pf-m-chip-container modifier from the content row below the toolbar that displays chips and the X filters applied message depending on viewport width? so the pf-m-chip-container modifier is only used in the expandable content. Did anything else change from an html hierarchy perspective?

mattnolting:
> So just to be clear, the last couple pushes removed the pf-m-chip-container modifier from the content row below the toolbar that displays chips and the X filters applied message depending on viewport width? so the pf-m-chip-container modifier is only used in the expandable content. Did anything else change from an html hierarchy perspective?

No, I made a change to that. I've extended the `chip-container` layout to `__content` as well as `__group` so that, if users want to use custom filtering, the can do so outside of the expandable content.

PR #2405, opened 2019-10-28
fix issues when trying to have patternfly.scss as a dependency

cntlsn:
When importing `patternfly.scss` as a dependency in a project, a series of errors are thrown for not finding `/components/_all.scss` and `/layouts/_all.scss` files. 
In fact, when installing PF4 via `npm install @patternfly/patternfly --save` the two files are not currently copied over.

The PR also fixes a second error thrown by gulp after manually patching the `_all.scss` files: the `@import "../../patternfly-imports"` in `components/SkipToContent/skip-to-content.scss` doesn't specify an extension and gulp is not sure whether it would need to import `patternfly-imports.css` or `patternfly-imports.scss`.


redallen:
Didn't mean to close, reopening in #2407

PR #2419, opened 2019-10-31
fix(patternfly-imports): don't compile patternfly-imports

redallen:
Don't compile this file for consumers. Let them default to the SCSS.


mcoker:
Should we also update the release notes with the version bump?

redallen:
@mcoker Yeah. I'll do that along with the site work on the gatsby-theme-patternfly-org branch today/tomorrow.

PR #2421, opened 2019-11-01
fix(build): move build back to separate component build for correct dist

dgutride:
We need the component structure and sass-variables/patternfly.scss to match what it used to be.  This puts it back and works with the new gatsby config.


dgutride:
I also copied this over to integr8ly to see if it works there and it's fine on that end, too @redallen .

redallen:
I fixed the watch task to work. Compiling our SASS separately during the watch task has been problematic for hot-reloading, so we've moved back to Gatsby compiling our SASS for now.

dgutride:
Please do not commit static/assets/*, they do not belong in that location in the repo.

PR #2428, opened 2019-11-06
feat(app-launcher): added support for search input, favorites

mcoker:
fix https://github.com/patternfly/patternfly-next/issues/2365

### DIFF
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-input.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-input.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-app-launcher__menu-input{{#if app-launcher-menu-input--modifier}} {{app-launcher-menu-input--modifier}}{{/if}}"
+  {{#if app-launcher-menu-input--attribute}}
+    {{{app-launcher-menu-input--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-input.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-input.hbs
@@ -1,6 +1,0 @@
-<div class="pf-c-app-launcher__menu-input{{#if app-launcher-menu-input--modifier}} {{app-launcher-menu-input--modifier}}{{/if}}"
-  {{#if app-launcher-menu-input--attribute}}
-    {{{app-launcher-menu-input--attribute}}}
-  {{/if}}>
-  {{> @partial-block}}
-</div>
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-search.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-search.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-app-launcher__menu-search{{#if app-launcher-menu-search--modifier}} {{app-launcher-menu-search--modifier}}{{/if}}"
+  {{#if app-launcher-menu-search--attribute}}
+    {{{app-launcher-menu-search--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
--- a/src/patternfly/components/AppLauncher/app-launcher.scss
+++ b/src/patternfly/components/AppLauncher/app-launcher.scss
@@ -22,6 +22,14 @@
   --pf-c-app-launcher__toggle--disabled--Color: var(--pf-global--disabled-color--200);
   --pf-c-app-launcher__toggle--m-expanded--Color: var(--pf-global--Color--100);
 
+  // Menu input
+  --pf-c-app-launcher__menu-input--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-input--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-input--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-input--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-input--BottomBorderColor: var(--pf-global--BorderColor--100);
+  --pf-c-app-launcher__menu-input--BottomBorderWidth: var(--pf-global--BorderWidth--sm);
+
   // Menu item
   --pf-c-app-launcher__menu-item--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-app-launcher__menu-item--PaddingRight: var(--pf-global--spacer--md);
@@ -29,8 +37,29 @@
   --pf-c-app-launcher__menu-item--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-app-launcher__menu-item--Color: var(--pf-global--Color--dark-100);
   --pf-c-app-launcher__menu-item--FontWeight: var(--pf-global--FontWeight--normal);
+  --pf-c-app-launcher__menu-item--Width: 100%;
   --pf-c-app-launcher__menu-item--disabled--Color: var(--pf-global--Color--dark-200);
   --pf-c-app-launcher__menu-item--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
+  --pf-c-app-launcher__menu-item--m-wrapper--PaddingTop: 0;
+  --pf-c-app-launcher__menu-item--m-wrapper--PaddingRight: 0;
+  --pf-c-app-launcher__menu-item--m-wrapper--PaddingBottom: 0;
+  --pf-c-app-launcher__menu-item--m-wrapper--PaddingLeft: 0;
+  --pf-c-app-launcher__menu-item--m-link--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-item--m-link--PaddingRight: 0;
+  --pf-c-app-launcher__menu-item--m-link--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-item--m-link--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-item--m-link--hover--BackgroundColor: transparent;
+  --pf-c-app-launcher__menu-item--m-action--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-item--m-action--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-item--m-action--PaddingBottom: calc(var(--pf-global--spacer--sm) + 2px); // extra padding to vertically center icon with text
+  --pf-c-app-launcher__menu-item--m-action--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-item--m-action--hover--BackgroundColor: transparent;
+  --pf-c-app-launcher__menu-item--m-action--Color: var(--pf-global--disabled-color--200);
+  --pf-c-app-launcher__menu-item--m-action--Width: auto;
+  --pf-c-app-launcher__menu-item--m-action--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-app-launcher__menu-item--hover__menu-item--m-action--Color: var(--pf-global--Color--200);
+  --pf-c-app-launcher__menu-item--m-action--hover--Color: var(--pf-global--Color--100);
+  --pf-c-app-launcher__menu-item--m-favorite__menu-item--m-action--Color: #{$pf-color-gold-400};
 
   // Menu item icon
   --pf-c-app-launcher__menu-item-icon--MarginRight: var(--pf-global--spacer--sm);
@@ -41,7 +70,7 @@
   --pf-c-app-launcher__menu-item-external-icon--Color: var(--pf-global--link--Color);
   --pf-c-app-launcher__menu-item-external-icon--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-app-launcher__menu-item-external-icon--Transform: translateY(#{pf-size-prem(-1px)});
-  --pf-c-app-launcher__menu-item-external-icon--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-app-launcher__menu-item-external-icon--FontSize: var(--pf-global--icon--FontSize--sm);
 
   // Divider
   --pf-c-app-launcher__c-divider--MarginTop: var(--pf-global--spacer--sm);
@@ -132,10 +161,15 @@
   }
 }
 
+.pf-c-app-launcher__menu-input {
+  padding: var(--pf-c-app-launcher__menu-input--PaddingTop) var(--pf-c-app-launcher__menu-input--PaddingRight) var(--pf-c-app-launcher__menu-input--PaddingBottom) var(--pf-c-app-launcher__menu-input--PaddingLeft);
+  border-bottom: var(--pf-c-app-launcher__menu-input--BottomBorderWidth) solid var(--pf-c-app-launcher__menu-input--BottomBorderColor);
+}
+
 .pf-c-app-launcher__menu-item {
   display: flex;
   align-items: center;
-  width: 100%;
+  width: var(--pf-c-app-launcher__menu-item--Width);
   padding: var(--pf-c-app-launcher__menu-item--PaddingTop) var(--pf-c-app-launcher__menu-item--PaddingRight) var(--pf-c-app-launcher__menu-item--PaddingBottom) var(--pf-c-app-launcher__menu-item--PaddingLeft);
   font-weight: var(--pf-c-app-launcher__menu-item--FontWeight);
   color: var(--pf-c-app-launcher__menu-item--Color);
@@ -146,6 +180,8 @@
   &.pf-m-hover,
   &:focus,
   &.pf-m-focus {
+    --pf-c-app-launcher__menu-item--m-action--Color: var(--pf-c-app-launcher__menu-item--hover__menu-item--m-action--Color);
+
     text-decoration: none;
     background-color: var(--pf-c-app-launcher__menu-item--hover--BackgroundColor);
   }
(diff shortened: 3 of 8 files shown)


maryshak1996:
this looks great to me :)

src/patternfly/components/AppLauncher/app-launcher-menu-input.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-app-launcher__menu-input{{#if app-launcher-menu-input--modifier}} {{app-launcher-menu-input--modifier}}{{/if}}"
```
matthewcarleton: do you think this should be `pf-c-app-launcher__menu-search-input`? to be more specific? Maybe more bulletproof in the future?

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -22,15 +25,44 @@
...
+  --pf-c-app-launcher__menu-item--m-link--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-item--m-link--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-item--m-link--hover--BackgroundColor: transparent;
+  --pf-c-app-launcher__menu-item--m-action--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-item--m-action--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-item--m-action--PaddingBottom: calc(var(--pf-global--spacer--sm) + 2px); // extra padding to vertically center icon with text
+  --pf-c-app-launcher__menu-item--m-action--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-item--m-action--hover--BackgroundColor: transparent;
+  --pf-c-app-launcher__menu-item--m-action--Color: var(--pf-global--disabled-color--200);
+  --pf-c-app-launcher__menu-item--m-action--Width: auto;
```
matthewcarleton: Looks like this isn't being used.

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -166,6 +205,44 @@
...
+
+  &.pf-m-link {
+    --pf-c-app-launcher__menu-item--PaddingTop: var(--pf-c-app-launcher__menu-item--m-link--PaddingTop);
+    --pf-c-app-launcher__menu-item--PaddingRight: var(--pf-c-app-launcher__menu-item--m-link--PaddingRight);
+    --pf-c-app-launcher__menu-item--PaddingBottom: var(--pf-c-app-launcher__menu-item--m-link--PaddingBottom);
+    --pf-c-app-launcher__menu-item--PaddingLeft: var(--pf-c-app-launcher__menu-item--m-link--PaddingLeft);
+    --pf-c-app-launcher__menu-item--hover--BackgroundColor: var(--pf-c-app-launcher__menu-item--m-link--hover--BackgroundColor);
+  }
+
+  &.pf-m-action {
```
matthewcarleton: I wonder if this should be a more specific modifier name? Maybe the default is `pf-m-favorite` and the active modifier is `pf-m-favorite-active` or `pf-m-favorited`? They seem like they should be more connected.

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -166,6 +205,44 @@
...
+    --pf-c-app-launcher__menu-item--PaddingLeft: var(--pf-c-app-launcher__menu-item--m-action--PaddingLeft);
+    --pf-c-app-launcher__menu-item--hover--BackgroundColor: var(--pf-c-app-launcher__menu-item--m-action--hover--BackgroundColor);
+    --pf-c-app-launcher__menu-item--Color: var(--pf-c-app-launcher__menu-item--m-action--Color);
+    --pf-c-app-launcher__menu-item--Width: var(--pf-c-app-launcher__menu-item--Width);
+
+    font-size: var(--pf-c-app-launcher__menu-item--m-action--FontSize);
+
+    &:hover,
+    &.pf-m-hover {
+      --pf-c-app-launcher__menu-item--m-action--Color: var(--pf-c-app-launcher__menu-item--m-action--hover--Color);
```
matthewcarleton: Should this be applied to the focus as well?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -197,6 +197,112 @@ import './AppLauncher.css'
...
+    {{/app-launcher-menu-input}}
+    {{#> app-launcher-group}}
+      {{#> app-launcher-group-title}}
+        Favorites
+      {{/app-launcher-group-title}}
+      <ul>
+        {{#> app-launcher-menu-item app-launcher-menu-item--type="li" app-launcher-menu-item--modifier="pf-m-wrapper pf-m-external pf-m-favorite" app-launcher-menu-item--attribute='href="#"'}}
+          {{#> app-launcher-menu-item app-launcher-menu-item--modifier="pf-m-link" app-launcher-menu-item-favorite-link--attribute='href="#" target="_blank"'}}
+            {{#> app-launcher-menu-item-icon}}
+              <img src="/assets/images/pf-logo-small.svg" alt="">
```
matthewcarleton: are we supposed to have html here or should this be a .hbs partial?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -197,6 +197,112 @@ import './AppLauncher.css'
...
+  {{#> app-launcher-menu}}
+    {{#> app-launcher-menu-input}}
+      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="search" placeholder="Filter by name..." id="' id '-text-input" name="textInput1"')}}
+        {{/form-control}}
+    {{/app-launcher-menu-input}}
+    {{#> app-launcher-group}}
+      {{#> app-launcher-group-title}}
+        Favorites
+      {{/app-launcher-group-title}}
+      <ul>
```
matthewcarleton: Same for this? Should this be a partial?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -197,6 +197,112 @@ import './AppLauncher.css'
...
+        {{#> app-launcher-menu-item app-launcher-menu-item--type="li" app-launcher-menu-item--modifier="pf-m-wrapper pf-m-external pf-m-favorite" app-launcher-menu-item--attribute='href="#"'}}
+          {{#> app-launcher-menu-item app-launcher-menu-item--modifier="pf-m-link" app-launcher-menu-item-favorite-link--attribute='href="#" target="_blank"'}}
+            {{#> app-launcher-menu-item-icon}}
+              <img src="/assets/images/pf-logo-small.svg" alt="">
+            {{/app-launcher-menu-item-icon}}
+            Link 3
+          {{> app-launcher-menu-item-external-icon}}
+          {{/app-launcher-menu-item}}
+          {{#> app-launcher-menu-item app-launcher-menu-item--type="button" app-launcher-menu-item--modifier="pf-m-action"}}
+            <i class="fas fa-star" aria-label="Favorite"></i>
```
matthewcarleton: Should this be a partial?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -197,6 +197,112 @@ import './AppLauncher.css'
...
+    {{#> app-launcher-menu-input}}
+      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="search" placeholder="Filter by name..." id="' id '-text-input" name="textInput1"')}}
+        {{/form-control}}
+    {{/app-launcher-menu-input}}
+    {{#> app-launcher-group}}
+      {{#> app-launcher-group-title}}
+        Favorites
+      {{/app-launcher-group-title}}
+      <ul>
+        {{#> app-launcher-menu-item app-launcher-menu-item--type="li" app-launcher-menu-item--modifier="pf-m-wrapper pf-m-external pf-m-favorite" app-launcher-menu-item--attribute='href="#"'}}
```
matthewcarleton: you've included `app-launcher-menu-item--attribute='href="#"'` on all the `li`s

src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -197,6 +197,112 @@ import './AppLauncher.css'
...
+          {{#> app-launcher-menu-item app-launcher-menu-item--modifier="pf-m-link" app-launcher-menu-item-favorite-link--attribute='href="#" target="_blank"'}}
```
matthewcarleton: Should both the `li` and the child `a` have the same class with just a different modifier? Shouldn't they just have different classes via BEM?

matthewcarleton: `app-launcher-menu-item-favorite-link--attribute` wasn't added so `app-launcher-menu-item-favorite-link--attribute='href="#"` isn't adding the href attribute.

src/patternfly/components/AppLauncher/app-launcher-menu-input.hbs
mcoker: Yep, I named this after the context switcher, but agreed and I think this element could just be named `__menu-search`

src/patternfly/components/AppLauncher/app-launcher.scss
mcoker: doh! it should be.

src/patternfly/components/AppLauncher/app-launcher.scss
mcoker: The way I did it is an attempt to make a generic action button that is separate from the link in a menu item. The intent is that we could add more actions here and it would scale, then add unique modifiers (like `.pf-m-favorite`) for unique things that apply to that individual action.

That said, I can take a more specific approach and name these elements (the link, too) something specific to the favorite feature if you think that's better. 

Do you think these existing classes will scale in a generic way?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
mcoker: I leave it as HTML because it allows you to pass a unique image as the `@partial-block` to the partial.

src/patternfly/components/AppLauncher/examples/AppLauncher.md
mcoker: Since it's not a BEM element, I'm not really concerned with it being in a partial. Same way for the dropdown and options menu group lists.

src/patternfly/components/AppLauncher/examples/AppLauncher.md
mcoker: > Should both the li and the child a have the same class with just a different modifier? Shouldn't they just have different classes via BEM?

I wondered the same thing. I started out building them as their own BEM elements but realized they're basically just items and have the same basic style as the original item, with a few adjustments in padding and background color mostly. They all share some styles of the original item CSS, with some changes. I ended up with this, and it felt better than creating unique elements.

That said I pushed a unique commit in https://github.com/patternfly/patternfly-next/pull/2428/commits/c3a0f0a3f09c9369fd861645bd89d2415e09ca1d where I did the converstion of changing the outer `.pf-c-app-launcher__menu-item.pf-m-wrapper` to `.pf-c-app-launcher__menu-item-wrapper` and updated the rest of the styles to work. I think it makes things a bit more complicated but have a look and let me know what you think.

src/patternfly/components/AppLauncher/app-launcher.scss
matthewcarleton: ah ok, I see what you mean. That's smart, I like it! 👍

src/patternfly/components/AppLauncher/examples/AppLauncher.md
matthewcarleton: 🔥

src/patternfly/components/AppLauncher/examples/AppLauncher.md
matthewcarleton: 👍

src/patternfly/components/AppLauncher/examples/AppLauncher.md
matthewcarleton: ya it's definitely not as straight forward. My main concern is how this could cause confusion because they are not siblings. I suppose it's just down to how strict are we when following BEM and how concerned are we that this could be implemented incorrectly? Will this have any impact on the React implementation?

christiemolloy:
Is it possible that these stars will be repeated in other components? Did you consider making them a checkbox? I'm not sure if it has to be a checkbox for a11y reasons or if a button is fine.

src/patternfly/components/AppLauncher/app-launcher-menu-search.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-app-launcher__menu-search{{#if app-launcher-menu-search--modifier}} {{app-launcher-menu-search--modifier}}{{/if}}"
```
christiemolloy: In the context selector, this wrapper is named as `pf-c-context-selector__menu-input`. I think we should be consistent here. I don't have an opinion on the name I think search is definitely more specific if know its only going to be search there. We could update the context selector in a breaking change too.

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -22,15 +25,44 @@
   --pf-c-app-launcher__toggle--disabled--Color: var(--pf-global--disabled-color--200);
   --pf-c-app-launcher__toggle--m-expanded--Color: var(--pf-global--Color--100);
 
+  // Menu input
+  --pf-c-app-launcher__menu-search--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-app-launcher__menu-search--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-search--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-search--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-app-launcher__menu-search--BottomBorderColor: var(--pf-global--BorderColor--100);
```
christiemolloy: Would you not use the divider component here?

src/patternfly/components/AppLauncher/app-launcher-menu-search.hbs
mcoker: Agreed, I'd like to update the context selector in a breaking change release to rename it. I went ahead and renamed it here to avoid a breaking change in the future, or do you think we should name this `menu-input` now and update them both in a breaking change release?

src/patternfly/components/AppLauncher/app-launcher.scss
mcoker: It's still experimental, we shouldn't be using it in components until it's promoted to a regular component. I have another issue open to update our examples in the components where we're using it now.

christiemolloy: Oh yep thats right!

src/patternfly/components/AppLauncher/app-launcher-menu-search.hbs
christiemolloy: I think search is good here and then create a follow up issue to change the context selector

mcoker: https://github.com/patternfly/patternfly-next/issues/2438

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -41,7 +73,7 @@
   --pf-c-app-launcher__menu-item-external-icon--Color: var(--pf-global--link--Color);
   --pf-c-app-launcher__menu-item-external-icon--PaddingLeft: var(--pf-global--spacer--lg);
```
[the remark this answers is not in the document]
mcoker: Nice, I think that looks a lot better.

src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -197,6 +197,112 @@ import './AppLauncher.css'
 {{/app-launcher}}
 ```
 
+```hbs title=Favorites
+{{#> app-launcher id="app-launcher-sections-dividers-icons-and-external-links" app-launcher--IsExpanded="true" app-launcher--IsGrouped="true"}}
+  {{#> app-launcher-menu}}
+    {{#> app-launcher-menu-search}}
+      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="search" placeholder="Filter by name..." id="' id '-text-input" name="textInput1"')}}
```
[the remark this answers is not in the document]
mcoker: @mcarrano @maryshak1996 what do you think about adding a search button here? We have a search button in the context selector, which is very similar. That would provide a label for the input.

mcoker:
@maryshak1996 can you clarify the text truncation behavior you'd expect in this menu? Currently the menu width is determined by the width of the items in it, so it will grow to be as wide as the longest item in the menu. Same behavior for the dropdown and options menu.

src/patternfly/components/AppLauncher/examples/AppLauncher.md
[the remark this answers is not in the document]
mcarrano: I'd like this to filter as you type (i.e. a type ahead filter). In retrospect I wish we did not have the search button on the Context Selector as this is a better interaction pattern IMO. It was also the request from the OS team.

As for the label, how do we handle this for a type-ahead select?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
mcoker: Are you referring to how the structure is `__menu-item-wrapper` -> `__menu-item`, when we normally name classes so that `__menu-item[-whatever]` would be a child of `__menu-item`?

src/patternfly/components/AppLauncher/examples/AppLauncher.md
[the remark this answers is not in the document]
mcoker: Ah thanks @mcarrano, added an `aria-label`

maryshak1996:
@mcoker my thought, with the width, would be that it would still keep that behavior for the default (un-hovered) state of the menu items (takes on the longest item width).I figured that truncation would only happen when, the added external icon appears on hover (since this would need extra room). or does that auto-sizing including the width of the external icon? if so, then nevermind about the truncation!

mcoker:
@maryshak1996 gotcha thanks! We do create space for the external icon, so when it appears on hover, it doesn't impact the text, item width, or overall menu width. Does the behavior of the [example in the PR](https://patternfly-next-pr-2428.surge.sh/documentation/core/components/applauncher#favorites) look like you would expect?

maryshak1996:
@mcoker yes that looks great! I just tested your example with a long menu item example and it looks great :)

maryshak1996:
Approved on my side

src/patternfly/components/AppLauncher/examples/AppLauncher.md
matthewcarleton: ya exactly.

mcoker: Ah gotcha, I renamed it to `__menu-wrapper`. Do you think this approach of it being it's own BEM element is preferred to it being `.pf-c-app-launcher__menu-item.pf-m-wrapper` (or use some other word rather than "wrapper" in the modifier or BEM name)? I think the modifier works as well. I'm curious what @mattnolting and @christiemolloy think about that, too.

mcoker:
@mattnolting @christiemolloy this should be good for another review.

PR #2430, opened 2019-11-07
fix(select): removed use of form element in toggle and menu

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2379

### DIFF
--- a/src/patternfly/components/Select/select-checkbox-groups-checked.hbs
+++ b/src/patternfly/components/Select/select-checkbox-groups-checked.hbs
@@ -3,50 +3,46 @@
   {{#if select-multi--attribute}}
     {{{select-multi-attribute}}}
   {{/if}}>
-  {{#> form}}
-    {{#> form-group form-group--attribute=(concat 'aria-labelledby="' id '-label"')}}
-      <div class="pf-c-select__menu-group">
-        <div aria-hidden="true" id="{{id}}-group-status" class="pf-c-select__menu-group-title">Status</div>
-         {{#> form-fieldset form-fieldset--attribute=(concat 'aria-labelledby="' id '-group-status"')}}
-          {{#> check check--type="label" check--attribute=(concat 'for="' id '-running"') check--modifier="pf-c-select__menu-item"}}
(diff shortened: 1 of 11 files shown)


mcoker:
@mattnolting ah good catch! Also, doesn't look like `.pf-c-select__toggle-typeahead-form` was ever added to the toggle `<form>` in react, so the typeahead input doesn't span full width in that component. It should now.

src/patternfly/components/Select/select-checkbox-groups.hbs
```diff
@@ -3,50 +3,46 @@
...
-            {{/check}}
-            {{#> check check--type="label" check--attribute=(concat 'for="' id '-hp"') check--modifier="pf-c-select__menu-item"}}
-              {{#> check-input check-input--attribute=(concat 'type="checkbox" id="' id '-hp" name="hp"')}}{{/check-input}}
-              {{#> check-label check-label--type="span"}}Hewlett-Packard{{/check-label}}
-            {{/check}}
-        {{/form-fieldset}}
-      </div>
-    {{/form-group}}
-  {{/form}}
+  <div class="pf-c-select__menu-group">
```
mattnolting: Can we make this a .hbs partial?

src/patternfly/components/Select/select-checkbox-groups-checked.hbs
```diff
@@ -3,50 +3,46 @@
...
-            {{/check}}
-            {{#> check check--type="label" check--attribute=(concat 'for="' id '-hp"') check--modifier="pf-c-select__menu-item"}}
-              {{#> check-input check-input--attribute=(concat 'type="checkbox" id="' id '-hp" name="hp"')}}{{/check-input}}
-              {{#> check-label check-label--type="span"}}Hewlett-Packard{{/check-label}}
-            {{/check}}
-          {{/form-fieldset}}
-      </div>
-    {{/form-group}}
-  {{/form}}
+  <div class="pf-c-select__menu-group">
```
mattnolting: Can we make this a .hbs partial?

Issue #2321, opened 2019-10-01
Inline edit for forms


mcarrano:
Want to support editable fields in a definition list (re-only form) allowing user to place either a single field or an entire form into edit mode.

Design doc: https://docs.google.com/document/d/15aOWgpkd0Zp0hvbKM4jfG6Jf00FHjY_f92bkosbo1Ds/edit?usp=sharing

Design issue: https://github.com/patternfly/patternfly-next/issues/295

cc: @rachael-phillips @mcoker Not sure if this is a new component or just a demo

mcarrano:
@mattnolting @mcoker Is this functionality supported today or is additional work needed? If there is, I will get this on the roadmap. It's functionality we should support.

mattnolting:
@mcarrano It's additional work. The core component is done, but the only react implementation is embedded in `pf-c-table`. https://www.patternfly.org/v4/documentation/react/components/table#editable-rows. I'd defer to React @tlabaj as to a standalone implementation.

mcarrano:
Thanks @mattnolting . In that case I am going to transfer this issue over to React and will get it on the roadmap.

Issue #2319, opened 2019-10-01
Add bulk inline editing to Table


mcarrano:
Adds bulk inline editing capability to a table. 

Design doc: https://docs.google.com/document/d/15aOWgpkd0Zp0hvbKM4jfG6Jf00FHjY_f92bkosbo1Ds/edit?usp=sharing

Design issue: https://github.com/patternfly/patternfly-next/issues/404

mcarrano:
@rachael-phillips @mcoker FYI

LHinson:
Putting this one on hold while we wait on  #2315 to complete

mcarrano:
Will add this to the roadmap.

mcoker:
@mcarrano should we leave this open?

mcarrano:
Yes. We do have an approved design for this and therefore I'd like to keep it in the backlog. But we have not had any recent request.

nicolethoen:
@mcarrano is the design for this feature still relevant/current? Should we add this to an upcoming milestone?

mcarrano:
@nicolethoen I see this as a "nice-to-have" feature. No one is asking for it, specifically. What is the effort/risk to implement this? If you feel it can be easily implemented on top of existing row edit capability, I'd be interested in seeing a demo. Otherwise I can close.

mcarrano:
That makes sense @nicolethoen . I have added the demo tag. Do you think we need a core demo of this or can this go directly to React? Not sure what the core demo would be.

mcoker:
@mcarrano we can do it as a core demo to make sure it works and show the structure.

PR #2446, opened 2019-11-18
feat(inline-edit): added inline-edit component

### LINKED ISSUE #2315: inline edit for single row (with no expand)
Adds single row inline editing to a Table.

Design doc: https://docs.google.com/document/d/15aOWgpkd0Zp0hvbKM4jfG6Jf00FHjY_f92bkosbo1Ds/edit?usp=sharing

Design issue: https://github.com/patternfly/patternfly-next/issues/404

mattnolting:
closes #2315 #2321 #2319 

This approach provides a flexible, reusable pattern for inline edit rather than baking `inline edit` into each context. 

Note: `<dl>` is styled by workspace `.css`. This should be supported by `.pf-c-list`. Issue is open here: https://github.com/patternfly/patternfly-next/issues/1919


src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,202 @@
...
+      input {
+        grid-column: 1;
+      }
+    }
+  }
+  /* stylelint-enable */
+}
+
+// Value group
+.pf-c-inline-edit__value-group {
```
christiemolloy: What's the purpose to splitting up `.pf-c-inline-edit__value-group` and `.pf-c-inline-edit__input-group`, I'm wondering if you can combine these containers to `pf-c-inline-edit__group` and then use modifiers on the input vs value groups. This way when content becomes editable, a class/prop could be added to the container rather than a whole container switch (same with the container for the title) But let me know what you think

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,202 @@
...
+    display: none;
+    visibility: hidden;
+  }
+
+  // Align right
+  .pf-m-align-right {
+    margin-left: auto;
+  }
+
+  // Overrides
```
christiemolloy: Remove

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: Yeah, I had thought about using a standard `__group` with modifiers. @mcoker wyt?

maryshak1996:
@mattnolting @mcarrano the only other thing that I'm seeing is that it's strange to me that the edit link button for  doesn't VISUALLY look left aligned with the content. Is there a modifier of sorts that can be added to remove that 16px of padding attached to the far left of the button in the definition list use case? 
Screen Shot 2019-11-19 at 11 44 30 AM
(make the pencil+edit flush left to the text)

mattnolting:
@maryshak1996 @mcarrano `.pf-m-link` doesn't have padding, but does have an underline on hover. `pf-m-inline` modifier removes the padding, but adds the underline on hover. I suggest we avoid creating specific modifiers for such a generic component. Rather that should be handled by a button modifier or utility class (not recommended) as this is a common issue noted in other PRs. Since the padding is invisible, the `.pf-m-link` button appears to be misaligned.

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,335 @@
+---
+title: Inline Edit
```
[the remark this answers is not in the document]
mcoker: submitted an issue for this - https://github.com/patternfly/patternfly-next/issues/2454

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,335 @@
...
+```
+
+```hbs title=Single-inline-edit-with-title-(default)
+{{#> inline-edit inline-edit--type="form" inline-edit--id="single-inline-edit-with-title-example"}}
+  {{#> inline-edit-title-group}}
+    {{#> title title--modifier="pf-m-xl" titleType="h2" title--attribute=(concat 'id="' inline-edit--id '-label"')}}
+      Single inline edit group
+    {{/title}}
+    {{#> inline-edit-toggle}}
+      {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```
mcoker:
```suggestion
      {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-label="edit" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,335 @@
...
+```
+
+```hbs title=Single-inline-edit-with-title-(active)
+{{#> inline-edit inline-edit--type="form" inline-edit--id="single-inline-edit-with-title-active-example" inline-edit--modifier="pf-m-inline-edit-active"}}
+  {{#> inline-edit-title-group}}
+    {{#> title title--modifier="pf-m-xl" titleType="h2" title--attribute=(concat 'id="' inline-edit--id '-label"')}}
+      Single inline edit group
+    {{/title}}
+    {{#> inline-edit-toggle}}
+      {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```
mcoker:
```suggestion
      {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-label="edit" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,335 @@
...
+            {{> form-control controlType="input" input="true" form-control--attribute=(concat 'type="number" value="2" id="' inline-edit--id inline-edit--row 'number-input"  aria-label="Number input"')}}
+          {{/inline-edit-input}}
+        {{/table-td}}
+        {{#> table-td table-td--action="true"}}
+          {{#> inline-edit-action-group inline-edit-action-group--IsInline="true"}}
+            {{> inline-edit-action-save inline-edit-action-save--IsIcon="true" inline-edit-action-save--modifier="pf-m-valid"}}
+            {{> inline-edit-action-cancel inline-edit-action-cancel--IsIcon="true"}}
+            {{#> inline-edit-action}}
+              {{#> inline-edit-toggle}}
+                {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```
mcoker:
```suggestion
                {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-label="edit" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,335 @@
...
+            {{> form-control controlType="input" input="true" form-control--attribute=(concat 'type="number" value="2" id="' inline-edit--id inline-edit--row 'number-input"  aria-label="Number input"')}}
+          {{/inline-edit-input}}
+        {{/table-td}}
+        {{#> table-td table-td--action="true"}}
+          {{#> inline-edit-action-group inline-edit-action-group--IsInline="true"}}
+            {{> inline-edit-action-save inline-edit-action-save--IsIcon="true"}}
+            {{> inline-edit-action-cancel inline-edit-action-cancel--IsIcon="true"}}
+            {{#> inline-edit-action}}
+              {{#> inline-edit-toggle}}
+                {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```
mcoker:
```suggestion
                {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-label="edit" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```

mcoker:
In the inline edit dl/bulk example, the edit toggle is still visible and it isn't in the designs.

Screen Shot 2019-11-19 at 5 20 20 PM

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,335 @@
...
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `id` | `[labelling element]` | Provides a reference for toggle button and/or editable content. |
+| `aria-label="descriptive content"` | `.pf-c-inline-edit__toggle > button` | Provides an accessible name for toggle button. **Required** |
+| `aria-labelledby="[labelling element] [toggle button id]"` | `.pf-c-inline-edit__toggle > button` | Provides an accessible description for toggle button. **Required** |
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-inline-edit` | `<form>`, `<div>`, `*` | Initiates the inline edit component. **Required** |
```
mcoker: why list anything along with `*` if that means it can apply to anything?

src/patternfly/components/InlineEdit/inline-edit-action-cancel.hbs
```diff
@@ -0,0 +1,11 @@
+{{#> inline-edit-action inline-edit-action--modifier="pf-m-cancel"}}
+  {{#if inline-edit-action-cancel--IsIcon}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Cancel edits"'}}
+      <i class="fas fa-times" aria-hidden="true" />
+    {{/button}}
+  {{else}}
+    {{#> button button--modifier="pf-m-secondary" button--attribute='aria-label="Cancel edits"'}}
```
mcoker: You shouldn't need an `aria-label` if the button has descriptive text.

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,202 @@
+// Don't remove this magic comment. See gulpfile.js.
+// @import "../../sass-utilities/all";
```
mcoker: You can remove this. Also added an issue to remove it from the blueprint file https://github.com/patternfly/patternfly-next/issues/2454

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  > :last-child {
+    --pf-c-inline-edit__value-group--item--MarginRight: 0;
+  }
+}
+
+// Input group
+.pf-c-inline-edit__input-group {
+  > * {
```
mcoker: why not define this as `> *:not(:last-child)` and remove the `> :last-child` selector below?

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,202 @@
...
+  /* stylelint-disable selector-max-class */
+  .pf-c-inline-edit__toggle .pf-c-button.pf-m-plain {
+    color: var(--pf-c-inline-edit--m-active__toggle--Color);
+  }
+  /* stylelint-enable */
+}
+
+/* stylelint-disable selector-max-class */
+.pf-c-inline-edit__action.pf-m-save.pf-m-valid .pf-c-button.pf-m-plain {
+  --pf-c-button--m-plain--Color: var(--pf-c-inline-edit__action--m-save--m-valid--c-button--m-plain--Color);
```
mcoker: We aren't currently editing other component variables in a component. Couldn't you just apply `color` to `.pf-c-inline-edit__action.pf-m-save.pf-m-valid` and the button will inherit it?

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: Two reasons:
1) It leaves the `:last-child` without a declaration at all. So, if a user _did_ want to update the `:last-child`, they would have to apply both property and value `margin: var(--var);` to the selector rather than updating the variable for the `:last-child`. In a theming file, that would be inconsistent with our intent of redefining variables rather than properties and values.
2) Base specificity 10 vs 20. The property is as easily overwritable as it possibly can be. Only the `:last-child` **css variable** itself has a specificity of 20. The goal is to avoid the need for complex selectors to override basic styles/dissuade developers from using `!important` tags.

src/patternfly/components/InlineEdit/inline-edit-toggle.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-inline-edit__toggle{{#if inline-edit-toggle--modifier}} {{inline-edit-toggle--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
mattnolting: Using `.pf-c-inline-edit__action` simplifies the code even more. Nice @christiemolloy!

mcarrano:
@mattnolting Thanks for updating the example "Bulk edit description list" to move the Save and Cancel buttons inside the toggle group. Can you just add some vertical spacing between the last form field and the buttons so they don't but up against each other?

src/patternfly/components/InlineEdit/inline-edit.scss
mcoker: sounds good!

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,193 @@
...
+.pf-c-inline-edit__group {
+  display: flex;
+
+  > * {
+    margin-right: var(--pf-c-inline-edit__group--item--MarginRight);
+  }
+
+  // Value group
+  &.pf-m-value-group {
+    --pf-c-inline-edit__group--item--MarginRight: var(--pf-c-inline-edit__group--m-value-group--item--MarginRight);
```
christiemolloy: Do you need to re-assign this variable here if they both have the same value of `sm` ? I think you could get rid of the var altogether

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,193 @@
...
+  .pf-c-inline-edit__action.pf-m-save,
+  .pf-c-inline-edit__action.pf-m-cancel {
+    display: inline-block;
+  }
+
+  // Action group
+  .pf-c-inline-edit__group.pf-m-action-group {
+    display: inline-flex;
+
+    // margin-top: var(--pf-c-inline-edit__action-group--MarginTop);
```
christiemolloy: remove this line

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,193 @@
...
+  --pf-c-inline-edit__group--item--MarginRight: var(--pf-global--spacer--sm);
+
+  // Value group
+  --pf-c-inline-edit__group--m-value-group--item--MarginRight: var(--pf-global--spacer--sm);
+
+  // Input group
+  --pf-c-inline-edit__group--m-input-group--item--MarginRight: var(--pf-global--spacer--sm);
+
+  // Action group
+  --pf-c-inline-edit__action-group--MarginTop: var(--pf-global--spacer--xl);
```
christiemolloy: why is this `--pf-c-inline-edit__action-group--MarginTop` instead of using `m-action-group`

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,352 @@
...
+| `.pf-c-inline-edit__action-group` | `*` | Initiates an inline edit action group that wraps action buttons within active, inline edit regions. |
+| `.pf-m-editable` | `.pf-c-inline-edit`, `.pf-c-inline-edit [block level element]` | Modifies an inline edit region for editable state. |
+| `.pf-m-edit` | `.pf-c-inline-edit__action` | Modifies for the edit action state. |
+| `.pf-m-save` | `.pf-c-inline-edit__action` | Modifies for the save action state. |
+| `.pf-m-cancel` | `.pf-c-inline-edit__action` | Modifies for the cancel action state. |
+| `.pf-m-valid` | `.pf-c-inline-edit__action.pf-m-save` | Modifies the save button state. |
+| `.pf-m-icon-group` | `.pf-c-inline-edit__action-group` | Modifies an inline edit action group item spacing. |
+| `.pf-m-inline` | `.pf-c-inline-edit__action-group` | Modifies an inline edit action group item spacing. |
+| `.pf-m-enable` | `.pf-c-inline-edit__action-group` | Modifies an inline edit action group visibility. |
+| `.pf-m-confirm` | `.pf-c-inline-edit__action-group` | Modifies an inline edit action group visibility. |
```
christiemolloy: This has the same description as `pf-m-enable`

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,352 @@
...
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-inline-edit` | `<form>`, `<div>` | Initiates the inline edit component. **Required** |
+| `.pf-c-inline-edit__value` | `*` | Initiates an inline edit value. **Required** |
+| `.pf-c-inline-edit__input` | `*` | Initiates an inline edit input. **Required** |
+| `.pf-c-inline-edit__action` | `*` | Initiates an inline edit action (visible when inline edit region is active). **Required** |
+| `.pf-c-inline-edit__value-group` | `*` | Initiates an inline edit value group. |
+| `.pf-c-inline-edit__input-group` | `*` | Initiates an inline edit input group. This group can be nested (check/radio group). |
+| `.pf-c-inline-edit__action-group` | `*` | Initiates an inline edit action group that wraps action buttons within active, inline edit regions. |
```
christiemolloy: if you have `pf-c-inline-edit__action-group` do you need `pf-m-action-group` as a modifier? Could you explain what they're doing.
Also I think `pf-m-action-group` can be changed to just `pf-m-action` 
Also would need to add `pf-m-action-group` to documentation

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,193 @@
...
+  .pf-c-inline-edit__action.pf-m-edit,
+  .pf-c-inline-edit__group.pf-m-value-group,
+  .pf-c-inline-edit__group.pf-m-action-group > * {
+    display: none;
+    visibility: hidden;
+  }
+
+  .pf-c-inline-edit__input,
+  .pf-c-inline-edit__group.pf-m-input-group,
+  .pf-c-inline-edit__action.pf-m-save,
```
christiemolloy: if pf-m-save and pf-m-cancel are just being toggled between `hidden/visible` do we even need those modifiers ? could we just toggle the group that they sit in?

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: It's not necessary, but allows for the value to be customized/themed.

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: That's correct

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: Don't need action `pf-c-inline-edit__action-group`

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: They may not always be in the group.

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: The classes override the default state for `.pf-m-action-group`, specifically for the table example. There's a group of buttons that enable edit mode, and another set that confirms changes whose visibility has to be toggled.

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: good catch, it's incorrect

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: The buttons may not always be inside the group

christiemolloy:
:gina-evil:

mattnolting:
@mcoker @christiemolloy ready for another review

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,362 @@
...
+    {{/inline-edit-group}}
+  {{/inline-edit-group}}
+{{/inline-edit}}
+```
+
+```hbs title=Single-inline-edit-with-title-(default)
+{{#> inline-edit inline-edit--type="form" inline-edit--id="single-inline-edit-with-title-example"}}
+  {{#> inline-edit-group}}
+    {{#> title title--modifier="pf-m-xl" titleType="h2" title--attribute=(concat 'id="' inline-edit--id '-label"')}}
+      Single inline edit group
```
christiemolloy: From what I see in the design document it looks like this should have more space underneath and maybe be smaller, but its hard to tell. IMO it looks too big but let me know if there is a design spec to work from
Screen Shot 2020-01-07 at 3 59 03 PM

src/patternfly/components/InlineEdit/inline-edit-dl.hbs
```diff
@@ -0,0 +1,93 @@
+{{#> title title--modifier="pf-m-xl" titleType="h3" title--attribute=(concat 'id="' inline-edit--id '-label" aria-hidden="true"')}}
+  Bulk edit description list
```
christiemolloy: Should there be space underneath the edit button? Could we have a layout for that

Screen Shot 2020-01-07 at 4 02 57 PM

christiemolloy: same thing for under the title here
Screen Shot 2020-01-07 at 4 03 56 PM

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,362 @@
...
+
+```hbs title=Inline-edit-dl-example-(active)
+{{#> inline-edit inline-edit--type="form" inline-edit--id="bulk-edit-dl-example-active" inline-edit--modifier="pf-m-editable"}}
+  {{#> inline-edit-dl}}{{/inline-edit-dl}}
+{{/inline-edit}}
+```
+
+```hbs title=Inline-edit-table-row
+{{#> inline-edit inline-edit--type="form" inline-edit--id="bulk-edit-table-example"}}
+  {{#> table table--grid="true" table--modifier="pf-m-grid-lg" table--attribute='aria-label="This is a simple table example"'}}
```
christiemolloy: Should the aria label change here to reflect that it is inline-edit

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,362 @@
...
+        {{#> table-td table-td--data-label="Number"}}
+          {{#> inline-edit-value}}
+            2
+          {{/inline-edit-value}}
+          {{#> inline-edit-input}}
+            {{> form-control controlType="input" input="true" form-control--attribute=(concat 'type="number" value="2" id="' inline-edit--id inline-edit--row 'number-input"  aria-label="Number input"')}}
+          {{/inline-edit-input}}
+        {{/table-td}}
+        {{#> table-td table-td--action="true"}}
+          {{#> inline-edit-group inline-edit-group--modifier="pf-m-action-group"}}
```
christiemolloy: Will have to add in styles to vertically center these actions

Screen Shot 2020-01-07 at 4 06 49 PM

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+    display: inline-block;
+  }
+
+  // Action group
+  .pf-c-inline-edit__group.pf-m-action-group {
+    display: inline-flex;
+  }
+
+  // Input group
+  .pf-c-inline-edit__group.pf-m-input-group {
```
christiemolloy: is this modifier necessary?

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: Inline edit doesn't style or add margin/padding/spacing to anything other than items inside its groups, as their visibility is controlled by the component. It vertical spacing is controlled by a layout or parent component, such as `.pf-c-form` or `.pf-c-list`. In this case, I added support for `<dl>` styling, based on the design files, in the components workspace .css. That may make it confusing to view...

src/patternfly/components/InlineEdit/inline-edit-dl.hbs
mattnolting: Controlled by layout or parent

src/patternfly/components/_all.scss
```diff
@@ -28,6 +28,8 @@
 @import "./Expandable/expandable.scss";
 @import "./FormControl/form-control.scss";
 @import "./Form/form.scss";
+
+// @import "./InlineEdit/inline-edit.scss";
```
seanforyou23: I think this needs to be uncommented?

mattnolting: When the component is promoted from experimental, we add/uncomment the import. While still experimental, the compiled `.css` is called here https://github.com/patternfly/patternfly-next/blob/master/gatsby-browser.js

seanforyou23: I see, I guess I was confused because it's not listed in that "// Experimental components" section like the others. For some reason, gatsby was giving me a hard time linking the projects locally to test this out without uncommenting that line, so I'll just do this locally while we work through the React implementation.

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,362 @@
...
+```
+
+```hbs title=Single-inline-edit-with-title-(default)
+{{#> inline-edit inline-edit--type="form" inline-edit--id="single-inline-edit-with-title-example"}}
+  {{#> inline-edit-group}}
+    {{#> title title--modifier="pf-m-xl" titleType="h2" title--attribute=(concat 'id="' inline-edit--id '-label"')}}
+      Single inline edit group
+    {{/title}}
+    {{#> inline-edit-toggle}}
+      {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-label="edit" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```
mcoker: small nit - values for `aria-label` should be sentence case

```suggestion
      {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' inline-edit--id '-edit-button" aria-label="Edit" aria-labelledby="' inline-edit--id '-label ' inline-edit--id '-edit-button"')}}
```

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  .pf-m-display-inline-flex {
+    display: inline-flex;
+  }
+
+  // Align right
+  .pf-m-align-right {
+    margin-left: auto;
+  }
```
mcoker: while these modifiers _should_ do the same thing between components, we can't always rely on that being the case. since `.pf-m-inline-edit` can wrap a bunch of things, can you prefix the modifiers with the selectors they apply to so they don't interfere with whatever else might be a descendent of `.pf-m-inline-edit`?

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+
+  /* stylelint-disable selector-max-compound-selectors */
+  [class*="pf-m-grid"] [data-label] {
```
mcoker: is this just for the purposes of the demo? Seems like it could be problematic if we leave it in.

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+  &.pf-m-value-group {
+    --pf-c-inline-edit__group--item--MarginRight: var(--pf-c-inline-edit__group--m-value-group--item--MarginRight);
+  }
+
+
+  // Input group
+  &.pf-m-input-group {
+    --pf-c-inline-edit__group--item--MarginRight: var(--pf-c-inline-edit__group--m-input-group--item--MarginRight);
+
+    .pf-m-action-group {
```
mcoker: can we scope this class to the bem element it applies to, too?

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+
+// Action
+.pf-c-inline-edit__action {
+  display: none;
+
+  &.pf-m-enable {
+    display: inline-block;
+  }
+
+  &.pf-m-valid .pf-c-button.pf-m-plain {
```
mcoker: modifier shouldn't be necessary since you're styling the plain var.

```suggestion
  &.pf-m-valid .pf-c-button {
```

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+    --pf-c-button--m-plain--Color: var(--pf-c-inline-edit__action--m-valid--c-button--m-plain--Color);
+
+    &:hover {
```
mcoker: this rule doesn't appear to be necessary since the rule above overriding `--pf-c-button--m-plain--Color` has higher specificity than the button components declaration to change `--pf-c-button--m-plain--Color: var(--pf-c-button--m-plain--hover--Color);`. Also if you wanted to style the hover color anyways, I believe you could just define `--pf-c-button--m-plain--hover--Color: var(--pf-c-inline-edit__action--m-valid--c-button--m-plain--hover--Color);` in the parent selector - no need to put it in a `&:hover` block

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+      --pf-c-button--m-plain--Color: var(--pf-c-inline-edit__action--m-valid--c-button--m-plain--hover--Color);
+    }
+  }
+}
+
+// Groups
+.pf-c-inline-edit [class*="group"] {
```
mcoker: can you use the inline-edit component group selectors here so the attr selector doesn't apply to things it isn't designed to?

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+  align-items: baseline;
+}
+
+// Editable
+.pf-c-inline-edit.pf-m-editable,
+.pf-c-inline-edit .pf-m-editable {
```
mcoker: the `.pf-m-editable` selector should be combined with the element(s) it applies to so it doesn't do something unintended if we use that class in something else inside and inline edit block.

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,206 @@
...
+
+  // Footer
+  &.pf-m-footer {
+    margin-top: var(--pf-c-inline-edit__group--m-footer--MarginTop);
+  }
+
+  // Column
+  &.pf-m-column {
+    --pf-c-inline-edit__group--item--MarginRight: 0;
+
```
mcoker: should you add a vertical spacer for the item(s) there? I think the same horizontal spacer size would work well.

mcoker:
@mattnolting overall this looks great! Left some comments, mainly around CSS scope, but otherwise looking really good 👍

src/patternfly/components/InlineEdit/inline-edit.scss
seanforyou23: One thing I noticed that's sorta on this topic is that the modifier `pf-m-error` doesn't receive it's full style properties when inside of a `pf-c-inline-edit` as it's scoped to `pf-c-form` with `--pf-c-form--m-error--Color`. Is there something we can do to make this work? I can add pf-c-form to the same element as pf-c-inline-edit but probably better handled on the CSS side.. wdyt?

seanforyou23: Maybe it's a matter of making `pf-m-error` more generic? Otherwise, the error text doesn't receive the red error color. Here's the markup I'm using as an immediate sibling to the TextInput react component

`<div class="pf-c-form__helper-text pf-m-error" aria-live="polite">This field is required</div>`

Is there more appropriate html to use? 

Screen Shot 2020-01-16 at 11 54 41 AM

I can raise an issue if need be, just wanna make sure I understand what exactly to ask for.

seanforyou23: Raised this issue to capture the problem better - feel free to comment/reword as necessary! https://github.com/patternfly/patternfly-next/issues/2588

mattnolting: > while these modifiers should do the same thing between components, we can't always rely on that being the case. since .pf-m-inline-edit can wrap a bunch of things, can you prefix the modifiers with the selectors they apply to so they don't interfere with whatever else might be a descendent of .pf-m-inline-edit?

It's pretty essential that we allow `*` descendants to be editable, regardless of type for a couple of reasons.

Inline edit can be self-contained or apply to a larger component. To scope `.pf-m-editable` to specific elements that could apply would be to scope it to all elements within all components. So scoping to anything within `pf-c-inline-edit` is intentional. Examples would making a `table row` editable. We can't add a wrapping element around a `table row`, so we have to apply `-editable` to `<tr>` add `-inline-edit`s components to `<td>`s. Same thing with `<dl>`s.  To consider each use case like that could inefficient bearing in mind that this is an invisible, utility component.

gatsby-browser.js
```diff
@@ -18,4 +18,9 @@ import './dist/patternfly-addons.css';
 
 // Experimental components
 import './dist/components/Drawer/drawer.css';
+<<<<<<< HEAD
+=======
```
mcoker: you've got some merge conflict code in here

src/patternfly/components/InlineEdit/inline-edit.scss
mcoker: It looks like you commented out a couple of the problematic classes except for `.pf-m-hidden` and the `.pf-m-grid [data-attr]` selector. If we don't bind those modifiers to an element, we just run the risk of them applying inadvertently to nested things we don't intend on. Would it be possible to wrap whatever you apply `.pf-m-hidden` to in an inline-edit element and apply `display: contents;` to it, then we could scope the modifier class to that?

We could always refactor patternfly to use proper BEM modifier classes so we don't have this problem :)

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: this is used to stack checkboxes and radios. it'll likely change in the future.

mattnolting: adding top margin throws off layout

src/patternfly/components/InlineEdit/examples/InlineEdit.md
christiemolloy: So should one be used over another? @mattnolting

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,562 @@
...
+| `.pf-c-inline-edit__action` | `*` | Initiates an inline edit action (visible when inline edit region is active). **Required** |
+| `.pf-m-icon-button` | `.pf-c-inline-edit__action` | Modifies the action button spacing. |
+| `.pf-m-valid` | `.pf-c-inline-edit__action` | Modifies the action button state. |
+| `.pf-m-value-group` | `.pf-c-inline-edit__group` | Modifies group for value group. |
+| `.pf-m-input-group` | `.pf-c-inline-edit__group` | Modifies group for input group. |
+| `.pf-m-action-group` | `.pf-c-inline-edit__group` | Modifies group for action group. |
+| `.pf-m-icon-group` | `.pf-c-inline-edit__group` | Modifies an action group item spacing. |
+| `.pf-m-column` | `.pf-c-inline-edit__group` | Modifies an action group flex direction. |
+| `.pf-m-footer` | `.pf-c-inline-edit__group` | Modifies an inline edit group margin-top. |
+| `.pf-m-enable` | `.pf-c-inline-edit__action` | Exposes an inline edit action by default. |
```
christiemolloy: I think `enable` may be too arbitrary, my first thought is what is being enabled? What about expose or visible

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,562 @@
...
+| `aria-label="descriptive text"` | `.pf-c-inline-edit__toggle > button` | Provides an accessible description for toggle button. **Required** |
+| `aria-labelledby="[labelling element] [toggle button id]"` | `.pf-c-inline-edit__toggle > button` | Provides an accessible description for toggle button. **Required** |
+| `aria-label="descriptive text"` | `[radio button group]` | Provides an accessible description for radio groups. **Required** |
+| `role="radiogroup"` | `[radio button group]` | Provides an accessible role for radio buttons groups. **Required** |
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-inline-edit` | `<form>`, `<div>` | Initiates the inline edit component. **Required** |
+| `.pf-m-inline-editable` | `.pf-c-inline-edit`, `.pf-c-inline-edit [block level element]` | Modifies an inline edit region for editable state. |
```
christiemolloy: should this be moved down with the modifiers?

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,562 @@
...
+| `.pf-c-inline-edit__value` | `*` | Initiates an inline edit value. **Required** |
+| `.pf-c-inline-edit__input` | `*` | Initiates an inline edit input. **Required** |
+| `.pf-c-inline-edit__label` | `*` | Initiates an inline edit label. |
+| `.pf-c-inline-edit__action` | `*` | Initiates an inline edit action (visible when inline edit region is active). **Required** |
+| `.pf-m-icon-button` | `.pf-c-inline-edit__action` | Modifies the action button spacing. |
```
christiemolloy: I wonder if instead of having the modifier on the individual buttons (as this could be a lot) the modifier goes on the parent element instead, and updates the margin between its children accordingly?

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,562 @@
...
+| `.pf-m-valid` | `.pf-c-inline-edit__action` | Modifies the action button state. |
```
christiemolloy: `pf-m-valid` is actually on `pf-c-button` in your example. i could see it going on `.pf-c-inline-edit__action` instead though.

src/patternfly/components/Table/examples/Table.md
```diff
@@ -387,6 +387,7 @@ When header cells are empty or they contain interactive elements, `<th>` should
 | -- | -- | -- |
 | `.pf-c-table__check` | `<th>`, `<td>` | Initiates a check table cell. |
 | `.pf-c-table__action` | `<th>`, `<td>` | Initiates an action table cell. |
+| `.pf-c-table__inline-edit-action` | `<th>`, `<td>` | Initiates an inline edit action table cell. |
```
christiemolloy: should you add in that `pf-m-inline-editable`applies to `<tr>`

christiemolloy: actually does that do anything to the row?

christiemolloy:
I would expect that the "edit" icon would sit a lot closer to the value that it is editing, is it possible for the icon to be aligned to the left? or does this reflect what design had?

Screen Shot 2020-01-22 at 5 06 23 PM

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,562 @@
...
+| `.pf-m-value-group` | `.pf-c-inline-edit__group` | Modifies group for value group. |
```
mcoker: is this used for anything?

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,562 @@
...
+| `.pf-m-input-group` | `.pf-c-inline-edit__group` | Modifies group for input group. |
+| `.pf-m-action-group` | `.pf-c-inline-edit__group` | Modifies group for action group. |
+| `.pf-m-icon-group` | `.pf-c-inline-edit__group` | Modifies an action group item spacing. |
```
mcoker: is this used?

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mcoker: agree with @christiemolloy, then you can just set the vars at the root of that selector and you shouldn't need the button selector

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,154 @@
...
+  --pf-c-inline-edit__action--m-icon-button--MarginRight: 0;
+
+  // Toggle
+  --pf-c-inline-edit--m-active__toggle--Color: var(--pf-global--Color--100);
+
+  // Footer group
+  --pf-c-inline-edit__group--m-footer--MarginTop: var(--pf-global--spacer--xl);
+
+  // Action group spacing
+  --pf-c-table__action--pf-c-inline-edit__group--m-action-group--MarginTop: var(--pf-global--spacer--md);
```
mcoker: doesn't look like this is used

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,154 @@
...
+
+  // Valid action
+  --pf-c-inline-edit__action--c-button--m-valid--m-plain--Color: var(--pf-global--link--Color);
+  --pf-c-inline-edit__action--c-button--m-valid--m-plain--hover--Color: var(--pf-global--link--Color--hover);
+
+  // Icon button
+  --pf-c-inline-edit__action--m-icon-button--MarginRight: 0;
+
+  // Toggle
+  --pf-c-inline-edit--m-active__toggle--Color: var(--pf-global--Color--100);
```
mcoker: doesn't appear to be used

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  > :last-child {
+    --pf-c-inline-edit__group--item--MarginRight: 0;
+  }
+}
+
+// Table action cells and action groups inside input groups
+.pf-c-inline-edit .pf-c-table__action,
```
mcoker: where does this selector contain `.pf-c-inline-edit__group.pf-m-action-group`? I'm not sure where this is used.

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,154 @@
...
+  &.pf-m-enable {
+    display: inline-block;
+    visibility: visible;
+  }
+}
+
+// offset button by top padding
+.pf-c-inline-edit__label + .pf-c-inline-edit__action.pf-m-enable > .pf-c-button {
+  margin-top: calc(var(--pf-c-button--PaddingTop) * -1);
+  margin-bottom: calc(var(--pf-c-button--PaddingBottom) * -1);
```
mcoker: This seems a bit weird since these buttons could overlap other things, but I'm not sure how else to handle it if we don't want the presence of those buttons to impact the layout.

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -0,0 +1,154 @@
...
+}
+
+.pf-c-inline-edit__label.pf-m-bold {
```
mcoker: if this is still used/needed, can you add it to the documentation? I'm curious when it would be used.

src/patternfly/components/Table/table.scss
```diff
@@ -319,8 +326,10 @@
 
 // Table action cell
 // ==================================================================
-.pf-c-table__action {
+.pf-c-table__action,
+.pf-c-table__inline-edit-action {
```
mcoker: should this be a modifier of `.pf-c-table__action`?

src/patternfly/components/Table/examples/Table.md
mattnolting: The row contains `.pf-c-inline-edit` componentry, so setting `.pf-m-inline-editable` on the row enables the editability of all its inline edit elements. The same is true for an entire form, wrapped in `.pf-c-inline-edit` or a single item (that could be in any context as its self contained). It can apply to any wrapping element or any `.pf-c-inline-edit` block. So it's not isolated to a `<tr>`, it can apply to almost anything.

mattnolting: But that's not a function of table, it's a function of `.pf-c-inline-edit`, so it's documented in that component instead.

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: Waffled over this one, I agree, group mod is best.

mattnolting:
> I would expect that the "edit" icon would sit a lot closer to the value that it is editing, is it possible for the icon to be aligned to the left? or does this reflect what design had?
> 
> Screen Shot 2020-01-22 at 5 06 23 PM

This is something we need to work through, but it's not a blocker. I discussed briefly w/@maryshak1996 and we arrived at the layout. That will likely change when we see it in action.

src/patternfly/components/Table/table.scss
mattnolting: Yes, I think it's cleanest to separate the two entirely. @seanforyou23 wdyt?

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: Noted, we can test and refine.

src/patternfly/components/InlineEdit/inline-edit.scss
mattnolting: Replaced by `pf-m-icon-group`

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: Yes, it is now in favor of `.pf-m-icon-button`

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: Nope

src/patternfly/components/Table/examples/Table.md
christiemolloy: got it, ok!

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,563 @@
...
+
+import './InlineEdit.css'
+
+## Introduction
+
+**Inline edit** is a utilitarian component that has two core elements, `.pf-c-inline-edit__value` and `.pf-c-inline-edit__input` and is based on a simple concept. When **value** is visible, **input** is hidden, and vice versa.
+
+### Generic groups
+
+`.pf-c-inline-edit__group`s provide basic layout support and have several available modifiers. The base behavior of `.pf-c-inline-edit__value` and `.pf-c-inline-edit__input` extends to `.pf-c-inline-edit__group.pf-m-value-group` and `.pf-c-inline-edit__group.pf-m-input-group`. Elements added to these groups will be inversely visible and activated with the `.pf-m-inline-editable` modifier.
```
christiemolloy: can remove `pf-m-value-group` now that it is removed

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,563 @@
...
+
+### Actions
+
+**Actions** (`.pf-c-inline-edit__action`) and **action groups** (`.pf-c-inline-edit__group.pf-m-action-group`) are hidden by default and exposed when a region of `.pf-c-inline-edit` becomes active. The default visibility of an **action** or **action group** can be inversed by adding `.pf-m-enable`. `.pf-m-enable` flags an element or group of elements as the controlling agents for enabling editable content and therefore is visible by default and hidden when inline editing is enabled.
```
christiemolloy: `pf-m-enable` needs to be updated with the new name now

src/patternfly/components/InlineEdit/inline-edit-action-save.hbs
```diff
@@ -0,0 +1,27 @@
+{{#if inline-edit-action-save--IsIcon}}
+  {{#if inline-edit-action-save--IsInvalid}}
+    {{#> inline-edit-action}}
+      {{#> button button--modifier="pf-m-plain" button--attribute='disabled aria-label="Save edits"'}}
+        <i class="fas fa-check" aria-hidden="true" />
+      {{/button}}
+    {{/inline-edit-action}}
+  {{else}}
+    {{#> inline-edit-action inline-edit-action--modifier="pf-m-valid"}}
+      {{#> button button--modifier="pf-m-plain pf-m-valid" button--attribute='aria-label="Save edits"'}}
```
christiemolloy: There is still `pf-m-valid` on the button, is this correct

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,563 @@
...
+      {{> inline-edit-action-save inline-edit-action-save--IsIcon="true"}}
+      {{> inline-edit-action-cancel inline-edit-action-cancel--IsIcon="true"}}
+    {{/inline-edit-group}}
+  {{/inline-edit-group}}
+{{/inline-edit}}
+```
+
+```hbs title=Single-inline-edit-with-label-(default)
+{{#> inline-edit inline-edit--type="form" inline-edit--id="single-inline-edit-with-label-example"}}
+  {{#> inline-edit-group}}
```
christiemolloy: why is the label and the action button inside a group but then the value is outside of the group? I thought that the value always sits inside of a group with the action button because that is what is directly being edited. Could you explain a little more :)

src/patternfly/components/InlineEdit/examples/InlineEdit.md
```diff
@@ -0,0 +1,563 @@
...
+        {{#> table-th}}
+          Checkboxes
+        {{/table-th}}
+        {{#> table-th}}
+          Radios
+        {{/table-th}}
+        {{#> table-th}}
+          Number
+        {{/table-th}}
+        {{#> table-td}}{{/table-td}}
```
christiemolloy: this should be updated to reflect the changes we made to the table: table-td cell is empty

src/patternfly/components/InlineEdit/examples/InlineEdit.md
mattnolting: Sure. The `-group` is now arbitrary and performs a basic function that `display: flex`s elements by default. The idea being that instead of having to call a layout, you can (optionally) use the `.pf-c-inline-edit__group` to provide basic alignment, but it's not necessary. `-groups` no longer hide/show, rather the elements within now hide/show.

PR #2451, opened 2019-11-18
feat(dropdown): add support for menu item icons

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2450

I think changing `pf-c-dropdown__menu-item` to `display: flex` would be a breaking change, so I've created a modifier (`.pf-m-icon`) to add that, then added a new element to support the icon configured the same as the app launcher.


src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -63,6 +63,14 @@ import './Dropdown.css'
 {{/dropdown}}
 ```
 
+```hbs title=Menu-item-icons
+{{#> dropdown id="dropdown-align-top-expanded" dropdown--IsActionMenu="true" dropdown--IsExpanded="true" dropdown--HasItemIcons="true" dropdown--HasToggleIcon="true"}}
```
mattnolting: needs a different `id` to pass axe testing

src/patternfly/components/Dropdown/dropdown-menu.hbs
```diff
@@ -4,10 +4,48 @@
...
   {{/if}}>
-  <li><a class="pf-c-dropdown__menu-item" href="#">Link</a></li>
-  <li><button class="pf-c-dropdown__menu-item">Action</button></li>
-  <li><a class="pf-c-dropdown__menu-item pf-m-disabled" aria-disabled="true" tabindex="-1" href="#">Disabled link</a></li>
-  <li><button class="pf-c-dropdown__menu-item" disabled>Disabled action</button></li>
+  <li>
+    {{#if dropdown--HasItemIcons}}
+      {{#> dropdown-menu-item dropdown-menu-item--type="a" dropdown-menu-item--modifier="pf-m-icon" dropdown-menu-item--attribute='href="#"'}}
+        {{#> dropdown-menu-item-icon}}
+          <img src="/assets/images/pf-logo-small.svg" alt="">
```
mattnolting: needs `aria-hidden` val

mcoker: from https://github.com/patternfly/patternfly-next/pull/1916#discussion_r293372913

> Since you are using `` elements for these icons, AND since these images are redundant (i.e. the text that follows is considered synonymous with these icons), then having `alt=""` like you have it here is the correct way to tell assistive technology that this image provides no extra meaning and to ignore it.

src/patternfly/components/Dropdown/examples/Dropdown.md
mcoker: updated

src/patternfly/components/Dropdown/dropdown-menu.hbs
mattnolting: Still, doesn't pass axe testing. If @jgiardino ok'd it in the past, I'm good with it.

PR #2453, opened 2019-11-19
chore(build): use patternfly-a11y

redallen:
Use new repo https://github.com/patternfly/patternfly-a11y which can be given a list of URLs to run aXe against.

Report formats are console, JUnit XML, and JSON.

Can test for errors aXe considers incomplete (like occluded element color contrast), or exclude them with `--noIncomplete`. Can aggregate JUnit results by component and React/Core using `--agregate`.


mcoker:
Tested this locally and pretty good! I'm not sure the scope of this PR but I'd like to understand better at some point

* the role of https://github.com/patternfly/patternfly-a11y, because it didn't seem necessary to run this
* how the list of pages we're testing is generated
* how to see/modify the rules
* how to blacklist or exclude certain things, and if it's possible to exclude those things on an individual component basis.
* if this can run alongside the normal dev server locally - I had to run `npm run build:site && npm run serve` to get `npm run test:a11y` to work

Though I tried to break `.pf-l-flex`'s a11y and got some odd results. I would have expected all of these to fail, but only 1 did... unless I did something wrong :)
  * `.pf-l-flex { color: #fff; }` - passed (should fail because the page background is white)
  * `.pf-l-flex { color: #fff; background: #fff; }` - passed
  * `.pf-l-flex { color: #fff; } .pf-l-flex > * { background: #fff; }` - passed
  * `.pf-l-flex { color: #fff; } .pf-l-flex > * { background: #f0f0f0; }` - failed?

redallen:
- The role of https://github.com/patternfly/patternfly-a11y for now is a CLI tool to run aXe tests on URLs and generate some reports for them. It is necessary to run this. It's been abstracted to its own package (rather than continuing to live here) so it can be used for [patternfly.org/v4](https://github.com/patternfly/gatsby-theme-patternfly-org/) and in [patternfly-react](https://github.com/patternfly/patternfly-react/pull/3339).
- The list of pages is generated [in gatsby-theme-patternfly-org](https://github.com/patternfly/gatsby-theme-patternfly-org/blob/master/gatsby-theme-patternfly-org/gatsby-node.js#L254). It's just all of the fullscreen example pages.
- Go upstream to [patternfly-a11y](https://github.com/patternfly/patternfly-a11y/blob/master/lib.js#L30) to see/modify rules. We can make everything configurable that we want.
- No blacklist/exclusion exists yet. We need to figure out how we'd like to do that.
- It can be run against the local dev server by using `patternfly-a11y --file .cache/fullscreenPages.json --prefix http://localhost:8000`. However, it's slower and buggier that running against a full build of the site.
- Testing should not have generated that file. Nothing exists under that directory in this PR. Look instead inside the `coverage` directory.

Breaking the flex layout's colors did not yield any results because:
1. We are passing `--noIncomplete` to the a11y tool per @jessiehuff 's request. These color contrast problems are usually marked `--noIncomplete` because aXe can't figure out what the background color of an element actually is.
2. We always pass the a11y tests in CI for now by using `npm run test:a11y:ci || true`

mcoker:
Sounds good. Re: the contrast/background color and `--noIncomplete`, it would be good if we can enable that at some point. We had contrast errors with the a11y tests previously from this - some components failed contrast because they're always intended for use against a different background color than the fullpage example showed them in, but those were just a few exceptions and that check caught more valid violations than false positives. 

We might also consider implementing some sort of timeout before the check happens as a lot of components have transitions that fire when the component loads, so the UI isn't in it's final form until that transition finishes.

re: https://github.com/patternfly/patternfly-a11y, ah gotcha. I wasn't sure if I needed to clone it or do something with it to run this.

Also being able to run it easily locally would be great since we often have to do that a lot troubleshooting some of the violations.

redallen:
The contrast check can be enabled whenever. I'm happy to do it in this PR if you wish.

> We might also consider implementing some sort of timeout before the check happens as a lot of components have transitions that fire when the component loads, so the UI isn't in it's final form until that transition finishes.

This is accomplished by using statically rendered HTML (vs using a dynamically rendered page like in dev mode). I can still add a timeout for possible transitions, what exactly are we targeting?

> Also being able to run it easily locally would be great since we often have to do that a lot troubleshooting some of the violations.
`npm run build:site && npm run test:a11y:ci`

PR #2483, opened 2019-12-02
fix(modal): add new variant that disables overflow scroll

### LINKED ISSUE #2194: Modal body overflow
When using a dropdown in a modal there is an overflow issue b/c of the positioning of the dropdown menu. 
This is caused by 
``` 
overflow-x: hidden; 
overflow-y: auto;
```
on
`.pf-c-modal-box__body`
This is what you get when an expanded menu is seen in the modal. I believe this is still the case in React unless there is some magic determining height.
Screen Shot 2019-08-26 at 6 23 07 PM
From what I recall this overflow was applied to avoid long scrolling viewports on smaller screens.

christiemolloy:
closes #2194

### DIFF
--- a/src/patternfly/components/ModalBox/examples/ModalBox.md
+++ b/src/patternfly/components/ModalBox/examples/ModalBox.md
@@ -76,6 +76,26 @@ cssPrefix: pf-c-modal-box
 {{/modal-box}}
 ```
 
+```hbs title=No-overflow-scroll-on-body
+{{#> modal-box modal-box--modifier="pf-m-disable-overflow-scroll" modal-box--attribute='aria-label="Example of a modal with overflow scroll disabled on the body" aria-describedby="modal-no-overflow-scroll-on-body"'}}
+  {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
+    <i class="fas fa-times" aria-hidden="true"></i>
+  {{/button}}
+  {{#> title titleType="h1" title--modifier="pf-m-2xl" title--attribute='id="modal-lg-title"'}}
+    Modal header
+  {{/title}}
+  {{#> modal-box-body modal-box-body--attribute='id="modal-description-without-scroll"'}}
+    Static text describing modal purpose. Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
+    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+    quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+    consequat.
+  {{/modal-box-body}}
+  {{#> modal-box-footer modal-box-footer--modifier="pf-m-align-left"}}
+    Modal footer
+  {{/modal-box-footer}}
+{{/modal-box}}
+```
+
 ## Documentation
 ### Overiew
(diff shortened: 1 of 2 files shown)


mcoker:
Hmm... the intent of this variation is to allow content from the body to overflow outside of the body instead of inside of it. The use case for introducing this is placing an expanded dropdown at the end of the body - the dropdown menu should overflow outside of the body. Here is what it looks like with this PR:

Screen Shot 2019-12-02 at 2 42 55 PM

I think all that's left to do is to disable the `overflow-*` properties on the `__body` element.

Also I might use `pf-m-no-` instead of `pf-m-disable-` as we're using that in other places to disable a feature. And you might consider making it `pf-m-no-body-overflow-scroll` in case we want to disable some "overflow scroll" feature of something else in the modal down the road.

src/patternfly/components/ModalBox/examples/ModalBox.md
```diff
@@ -76,6 +76,26 @@ cssPrefix: pf-c-modal-box
...
 
+```hbs title=Without-overflow-scroll-on-body
+{{#> modal-box modal-box--modifier="pf-m-disable-overflow-scroll" modal-box--attribute='aria-label="Example of a modal with overflow scroll disabled on the body" aria-describedby="modal-overflow-scroll-disabled-on-body"'}}
+  {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
+    <i class="fas fa-times" aria-hidden="true"></i>
+  {{/button}}
+  {{#> title titleType="h1" title--modifier="pf-m-2xl" title--attribute='id="modal-lg-title"'}}
+    Modal header
+  {{/title}}
+  {{#> modal-box-body modal-box-body--attribute='id="modal-description-without-scroll"'}}
```
mcoker:
```suggestion
  {{#> modal-box-body modal-box-body--attribute='id="modal-no-overflow-scroll-on-body-description"'}}
```

src/patternfly/components/ModalBox/examples/ModalBox.md
```diff
@@ -76,6 +76,26 @@ cssPrefix: pf-c-modal-box
 {{/modal-box}}
 ```
 
+```hbs title=Without-overflow-scroll-on-body
+{{#> modal-box modal-box--modifier="pf-m-disable-overflow-scroll" modal-box--attribute='aria-label="Example of a modal with overflow scroll disabled on the body" aria-describedby="modal-overflow-scroll-disabled-on-body"'}}
```
mcoker:
```suggestion
{{#> modal-box modal-box--modifier="pf-m-disable-overflow-scroll" modal-box--attribute='aria-label="Example of a modal with overflow scroll disabled on the body" aria-describedby="modal-no-overflow-scroll-on-body-description"'}}
```

src/patternfly/components/ModalBox/examples/ModalBox.md
```diff
@@ -76,6 +76,26 @@ cssPrefix: pf-c-modal-box
...
+  {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
+    <i class="fas fa-times" aria-hidden="true"></i>
+  {{/button}}
+  {{#> title titleType="h1" title--modifier="pf-m-2xl" title--attribute='id="modal-lg-title"'}}
```
mcoker:
```suggestion
  {{#> title titleType="h1" title--modifier="pf-m-2xl" title--attribute='id="modal-no-overflow-scroll-on-body-title"'}}
```

christianvogt:
We still have a need for modals to be scrollable. We often have modals with forms in them that contain dropdowns. How is this going to solve those use cases?

PR #2486, opened 2019-12-03
fix(charts): update threshold properties

dlabrecq:
Updated threshold chart properties to match design guidelines.

Fixes https://github.com/patternfly/patternfly-next/issues/2485


src/patternfly/_chart-globals.scss
```diff
@@ -256,6 +256,7 @@ $pf-chart-stack--data--stroke--Width: $pf-chart-global--stroke--Width--xs;
 
 // Threshold
 $pf-chart-threshold--stroke-dash-array: "6,6";
```
mcoker: Were you going to change this to `4,2` in this PR?

dlabrecq: Updated, thank you

PR #2487, opened 2019-12-03
fix(nav): removed bold for tertiary link current, active, focus

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2436


christiemolloy:
Should this change also apply to horizontal nav? It has the same problem when you click on it and font weight makes the text move to the right

Screen Shot 2019-12-03 at 1 53 33 PM

mcoker:
@christiemolloy good call, I thought we had a separate issue for that but that's for another issue with horizontal nav. Double checked with @mceledonia we want that change and we do. Updated the PR.

PR #2491, opened 2019-12-04
feat(datalist): added selectable/hoverable row variations

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2370

Notes:
* We removed the ability to select multiple rows in a data-list. Selecting multiple rows should be via checkbox select, not this kind of select.
* I refactored the way we draw the left border on expanded/selected rows to be able to remove `position: relative` from list items, so that I can use `position` to create a stacking context for selected rows, then only have to assign a single `z-index` to hovered rows to get the shadow stacking order correct.
* Moved the background from the data list to the items as that's needed so the shadows on a row don't show up behind adjacent rows.
* The [first commit](https://github.com/patternfly/patternfly-next/pull/2491/commits/f6f08c3714706a23b08e52a4b1ffb4eb5bb71096) has the majority of the feature work in it if you want to review that separately. The subsequent commits were to redo the hbs for data lists to make the a11y easier, plus a few minor feature tweaks.

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -16,9 +16,9 @@
   --pf-c-data-list__item--m-expanded--before--BackgroundColor: var(--pf-global--active-color--100);
   --pf-c-data-list__item--m-selected--before--BackgroundColor: var(--pf-global--active-color--100);
   --pf-c-data-list__item--m-selected--before--Top: calc(var(--pf-c-data-list__item-item--BorderTopWidth) * -1);
-  --pf-c-data-list__item--m-selected--BoxShadow: var(--pf-global--BoxShadow--sm);
+  --pf-c-data-list__item--m-selected--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
   --pf-c-data-list__item--m-selectable--hover--ZIndex: var(--pf-global--ZIndex--xs);
-  --pf-c-data-list__item--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--sm);
+  --pf-c-data-list__item--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
   --pf-c-data-list__item--m-expanded--m-selectable--before--BackgroundColor: var(--pf-global--active-color--300);
   --pf-c-data-list__item-item--BorderTopColor: var(--pf-global--BorderColor--100); // at breaking change switch var to BorderBottom
   --pf-c-data-list__item-item--BorderTopWidth: var(--pf-global--BorderWidth--sm); // at breaking change switch var to BorderBottom
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -14,6 +14,12 @@
 
   // Item
   --pf-c-data-list__item--m-expanded--before--BackgroundColor: var(--pf-global--active-color--100);
+  --pf-c-data-list__item--m-selected--before--BackgroundColor: var(--pf-global--active-color--100);
+  --pf-c-data-list__item--m-selected--before--Top: calc(var(--pf-c-data-list__item-item--BorderTopWidth) * -1);
+  --pf-c-data-list__item--m-selected--BoxShadow: var(--pf-global--BoxShadow--lg);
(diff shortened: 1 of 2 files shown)


christiemolloy:
When the data list item is selected, should it have a focus-ring to show that its selected/focusable?

Screen Shot 2019-12-04 at 3 08 28 PM

src/patternfly/components/DataList/data-list.scss
```diff
@@ -114,30 +121,15 @@
   @include pf-t-light;
 
   list-style-type: disc;
-  background-color: var(--pf-c-data-list--BackgroundColor);
```
christiemolloy: would this be breaking?

christiemolloy:
When I hover over "secondary content", the left blue border is extending too far

Screen Shot 2019-12-04 at 3 15 43 PM

christiemolloy:
No space between border and text

Screen Shot 2019-12-04 at 3 17 27 PM

src/patternfly/components/DataList/data-list.scss
mcoker: I don't think so since if someone had defined this as a custom color, the component should still present itself the same visually. Same with moving the pseudo element from the list `__item` element to 2 pseudo elements of the `__item-row` and `__expandable-content` elements.

mcoker:
@christiemolloy 

> No space between border and text

That's an existing bug, if we want to call it that. The problem with offsetting the left edge of the content by the width of the expanded border is that if someone were to use `pf-m-no-padding` and added a left padding so that the content in the expanded body matched the left alignment of the other elements in data list rows, that left alignment would then be off by 2px. I'm inclined to leave it as it is until this is filed as a bug by design or a product.

> When the data list item is selected, should it have a focus-ring to show that its selected/focusable?

I updated the shadows on hover/select so that the top shadow is more obvious. What do you think about that? Think we still need an outline on focus?

christiemolloy:
I thought for a11y reasons that the items would need a focus ring? Maybe not. @jessiehuff do you have any insight here

mcoker:
@christiemolloy good call. I added `tabindex="0"` to the selectable rows. @jessiehuff does that look ok to you?

Also chatted with @lboehling re: 

> When I hover over "secondary content", the left blue border is extending too far

She said it's not a big deal, though I did take a stab at addressing it, though it's pretty tricky because of the stacking order and how we define the borders. Let me know if you see any issues. Be sure to add expanded and selected/non-selected rows to the examples and hover over everything and try to break it since I'm redefining borders on hover so that the blue border overlaps the row borders consistently.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -114,53 +128,58 @@
...
-    top: var(--pf-c-data-list__item--before--Top);
-    bottom: 0;
-    left: 0;
-    width: var(--pf-c-data-list__item--before--Width);
-    height: var(--pf-c-data-list__item--before--Height); // remove after breaking change period
-    content: "";
-    background-color: var(--pf-c-data-list__item--before--BackgroundColor);
-    transition: var(--pf-c-data-list__item--before--Transition);
+  @media (min-width: $pf-global--breakpoint--sm) {
+    --pf-c-data-list__item--before--Top: var(--pf-c-data-list__item-item--before--Top);
```
mattnolting: Should this be `var(--pf-c-data-list__item--item--before--Top)`? 

The second `--item--` preceded with two dashes?

christiemolloy:
Seeing this on mobile / is it possible to fix the curve

Screen Shot 2019-12-06 at 10 24 15 AM
Screen Shot 2019-12-06 at 10 33 55 AM

mcoker:
@mattnolting

>   I wonder if we should extend the `.pf-m-selectable` option to `.pf-c-data-list`

I had actually done this originally, but I thought this would provide more flexibility after seeing how some users use `onRowClick` on tables, and how it is valid to only apply that handler on certain rows.

Though I did chat with @mcarrano and he said that if rows are selectable, the most likely use case is that all rows are selectable.

I guess this approach is more flexible, but there is more code and it's a bit more difficult to implement (requiring a class on each row and not just one class on the table). I could go either way...

src/patternfly/components/DataList/data-list.scss
mcoker: yep! but it was that way before, so changing it would be a breaking change. we need to update a lot of these vars in a breaking change release. I've made a note of that here - https://github.com/patternfly/patternfly-next/issues/2305

PR #2497, opened 2019-12-09
feat(selectable-card): added selectable card

### LINKED ISSUE #2470: Add hover and select states for Cards
To support the master-detail pattern we want to all for the possibility of a card grid to be used as part of the master panel. This will require the possibility for cards to have a select and hover state as was implemented for the DataList.

See designs here: https://docs.google.com/document/d/1kzGP37tb-lf7sVz43xBnnoYEWPsQN4GvNvffebxiuTE/edit?usp=sharing

master/detail sketch file: 
[Master:Detail.zip](https://github.com/patternfly/patternfly-next/files/3893514/Master.Detail.zip)

mattnolting:
Closes #2470

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -3,9 +3,12 @@
   --pf-c-card--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-card--BoxShadow: var(--pf-global--BoxShadow--sm);
   --pf-c-card--m-hoverable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
-  --pf-c-card--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
+  --pf-c-card--m-selectable--m-selected--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-card--m-selectable--m-selected--before--Height: var(--pf-global--BorderWidth--lg);
   --pf-c-card--m-selectable--m-selected--before--BackgroundColor: var(--pf-global--active-color--100);
+  --pf-c-card--m-selectable--hover--BoxShadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
+  --pf-c-card--m-selectable--focus--BoxShadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
+  --pf-c-card--m-selectable--active--BoxShadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
   --pf-c-card--m-compact__body--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-card--m-compact__footer--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-card--m-compact--first-child--PaddingTop: var(--pf-global--spacer--md);
@@ -38,13 +41,21 @@
     position: relative;
     cursor: pointer;
 
-    &.pf-m-selected,
-    &:hover,
-    &:focus {
+    &:hover {
       box-shadow: var(--pf-c-card--m-selectable--hover--BoxShadow);
     }
 
+    &:focus {
+      box-shadow: var(--pf-c-card--m-selectable--focus--BoxShadow);
+    }
+
+    &:active {
+      box-shadow: var(--pf-c-card--m-selectable--active--BoxShadow);
+    }
+
     &.pf-m-selected {
+      box-shadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
+
       &::before {
         position: absolute;
         top: 0;
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -3,12 +3,12 @@
   --pf-c-card--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-card--BoxShadow: var(--pf-global--BoxShadow--sm);
   --pf-c-card--m-hoverable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
+  --pf-c-card--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Card/card.scss
```diff
@@ -3,6 +3,9 @@
   --pf-c-card--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-card--BoxShadow: var(--pf-global--BoxShadow--sm);
   --pf-c-card--m-hoverable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
+  --pf-c-card--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
+  --pf-c-card--m-selectable--m-selected--before--Height: var(--pf-global--BorderWidth--lg);
+  --pf-c-card--m-selectable--m-selected--before--BackgroundColor: var(--pf-global--link--Color);
```
mcoker: I think we should reserve `link--Color` for links. We use `--pf-global--active-color--100` in the other components to indicate some active thing like tabs current state or the data-list/table expanded/selected state. I think an exception to that is nav link current border and I wonder if we should update that, too, or if it should always match the link color of the nav item. I'll make a note to ask @mceledonia about that when he gets back.

```suggestion
  --pf-c-card--m-selectable--m-selected--before--BackgroundColor: var(--pf-global--active-color--100);
```

mattnolting:
@mcoker updated

src/patternfly/components/Card/examples/Card.md
```diff
@@ -191,10 +191,29 @@ cssPrefix: pf-c-card
 {{/card}}
 ```
 
+```hbs title=Selectable
```
mcoker: What do you think about updating this to be `.pf-m-selectable` (remove `.pf-m-selected`) and adding another example for "selected" with `.pf-m-selectable.pf-m-selected`?

mattnolting: wfm

mattnolting:
@mcoker updated

src/patternfly/components/Card/card.scss
```diff
@@ -31,6 +34,29 @@
...
   }
 
+  &.pf-m-selectable {
+    position: relative;
+    cursor: pointer;
+
+    &.pf-m-selected,
+    &:hover,
+    &:focus {
+      box-shadow: var(--pf-c-card--m-selectable--hover--BoxShadow);
```
christiemolloy: Wonder if we should break this up into 3 variables in case it changes down the road .. i.e. they want different size shadows for the different states. Also no active state here?

src/patternfly/components/Card/examples/Card.md
```diff
@@ -207,3 +240,5 @@ A card is a generic rectangular container that can be used to build other compon
 | `.pf-m-compact` | `.pf-c-card` | Creates a compact variation of the card component that involves smaller font sizes and spacing. |
 | `.pf-m-no-fill` | `.pf-c-card__body` | Sets a `.pf-c-card__body` to not fill the available space in `.pf-c-card`. `.pf-m-no-fill` can be added to multiple card bodies. |
 | `.pf-m-hoverable` | `.pf-c-card` | Modifies the card to include hover styles on `:hover`. |
+| `.pf-m-selectable` | `.pf-c-card` | Modifies a selectable card so that it is selectable. |
+| `.pf-m-selected` | `.pf-c-card` | Modifies a selectable card for the selected state. |
```
christiemolloy:
```suggestion
| `.pf-m-selected` | `.pf-c-card.pf-m-selectable` | Modifies a selectable card for the selected state. |
```

mattnolting: 👍

src/patternfly/components/Card/card.scss
mattnolting: updated

mattnolting: I mapped the states to `--pf-c-card--m-selectable--m-selected--BoxShadow` as it is the primary modifier for selectable cards. I would expect that changing this variable would affect the default `:hover, :focus:, :active` states5

src/patternfly/components/Card/card.scss
```diff
@@ -3,9 +3,12 @@
   --pf-c-card--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-card--BoxShadow: var(--pf-global--BoxShadow--sm);
   --pf-c-card--m-hoverable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
-  --pf-c-card--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--lg);
+  --pf-c-card--m-selectable--m-selected--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-card--m-selectable--m-selected--before--Height: var(--pf-global--BorderWidth--lg);
   --pf-c-card--m-selectable--m-selected--before--BackgroundColor: var(--pf-global--active-color--100);
+  --pf-c-card--m-selectable--hover--BoxShadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
+  --pf-c-card--m-selectable--focus--BoxShadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
+  --pf-c-card--m-selectable--active--BoxShadow: var(--pf-c-card--m-selectable--m-selected--BoxShadow);
```
mcoker: The downside to sharing the vars like this is that if we need to update `--m-selected--BoxShadow` to be unique from the others, that will be a breaking change. The upside to this is if someone wants to change the box shadow for all of these values, they only have to change 1 thing. Typically we don't share the vars like this though, except in pretty specific instances (like a spacer to accommodate the value of a component's border width or something like that). The button is a good example. I would be in favor of making the value for these `var(--pf-global--BoxShadow--lg);`

mattnolting: updated

mcoker:
Nice! LGTM.

PR #2499, opened 2019-12-10
fix(empty-state): moved basic example

### LINKED ISSUE #2469: Change order of empty state examples in documentation
Goal: Shift the order of the empty state examples so that the "basic empty state" comes AFTER small, large, and [the future] extra large empty state examples, but BEFORE "with primary element" variation.

The reason for this is because this "basic" empty state variation is not one that should be used in mockups/actual implementation. Therefore, as an alternative to deleting it all together (in case a use case does come up in the future), we have decided to make it less visible in the page.

@mcarrano

mattnolting:
fixes #2469

### DIFF
--- a/src/patternfly/components/EmptyState/examples/EmptyState.md
+++ b/src/patternfly/components/EmptyState/examples/EmptyState.md
@@ -5,40 +5,6 @@ cssPrefix: pf-c-empty-state
 ---
 
 ## Examples
-```hbs title=Basic
-{{#> empty-state}}
-  {{#> empty-state-icon}}{{/empty-state-icon}}
-  {{#> title titleType="h1" title--modifier="pf-m-lg"}}
-    Empty state
-  {{/title}}
-  {{#> empty-state-body}}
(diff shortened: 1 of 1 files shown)


src/patternfly/components/EmptyState/examples/EmptyState.md
```diff
@@ -5,11 +5,12 @@ cssPrefix: pf-c-empty-state
...
 ## Examples
-```hbs title=Basic
-{{#> empty-state}}
+
+```hbs title=Small
+{{#> empty-state empty-state--modifier="pf-m-sm"}}
   {{#> empty-state-icon}}{{/empty-state-icon}}
   {{#> title titleType="h1" title--modifier="pf-m-lg"}}
-    Empty state
+    Empty State
```
mcoker: some scope creep, can we fix this so it's sentence case?

```suggestion
    Empty state
```

mattnolting:
@mcoker updated

PR #2500, opened 2019-12-10
fix(options-menu-separator): removed margin-bottom:last-child

### LINKED ISSUE #2488: Options menu - extra space at top of group when dividers are present
When dividers are used at the end of a group to separate groups, we need to remove the extra space at the top of the next group. 

This example:
https://pf4.patternfly.org/documentation/core/components/optionsmenu#with-groups-and-separators-between-items

Screen Shot 2019-12-03 at 12 55 04 PM

The spacing should match https://pf4.patternfly.org/documentation/core/components/optionsmenu#with-groups-and-separators-between-groups

For reference, the dropdown component had the correct spacing.

mattnolting:
fixes #2488

### DIFF
--- a/src/patternfly/components/OptionsMenu/options-menu.scss
+++ b/src/patternfly/components/OptionsMenu/options-menu.scss
@@ -294,6 +294,10 @@
   & + & {
     padding-top: var(--pf-c-options-menu__group--group--PaddingTop);
   }
+
+  .pf-c-options-menu__separator:last-child {
+    --pf-c-options-menu__separator--MarginBottom: 0;
+  }
 }
 
 .pf-c-options-menu__group-title {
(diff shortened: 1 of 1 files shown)


src/patternfly/components/OptionsMenu/options-menu.scss
```diff
@@ -294,6 +294,10 @@
   & + & {
     padding-top: var(--pf-c-options-menu__group--group--PaddingTop);
   }
+
+  .pf-c-options-menu__separator:last-child {
```
mcoker: can you nest this within the existing `.pf-c-options-menu__separator` selector using `:last-child` or does that break things? That that it's a requirement, but that's how the dropdown and app launcher are written, which should have the same layout.

mattnolting: Doh, yes that's how it should be.

PR #2511, opened 2019-12-11
feat(notification-drawer): added notification drawer component

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/1787

Given the number of overrides that would be needed to fit the accordion to this component, I opted just to create a unique element in this component. That may change in the future.

* The notification drawer should fill the height of its container.
* Unread notifications should be clickable to mark them as read.
* Expanded groups should push all of the other groups underneath it down until 1) all of the items in the group are visible, or 2) the groups underneath hit the bottom of the container at which point the list of notifications should scroll within the group. A couple of visuals showing this:

Screen Shot 2019-12-12 at 3 21 41 PM
Screen Shot 2019-12-12 at 3 21 47 PM

* Expanded groups have a min-height of 300px, so if the viewport shrinks in height and there isn't enough space to display the expanded group and all of the additional groups, an outer scrollbar will appear, creating 2 scrollbars and a scroll treadmill trap. This isn't ideal, but it should be a rare occurrence as there will likely only be a few groups in a notification drawer. We will likely iterate more on this interaction. Here's a screengrab of that.

Screen Shot 2019-12-12 at 3 22 02 PM

* I deviated from the `__el-nested-nested` BEM naming in a couple of places because the classnames were getting really long and it didn't seem necessary to follow that structure re: avoiding class name conflicts in the future. I'm open to changing this.

### DIFF
--- a/src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
+++ b/src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
@@ -104,6 +104,7 @@ cssPrefix: pf-c-notification-drawer
 | `aria-expanded="false"` | `.pf-c-notification-drawer__group-toggle` | Indicates that the group notification list is hidden. |
 | `aria-expanded="true"` | `.pf-c-notification-drawer__group-toggle` | Indicates that the group notification list is visible. |
 | `hidden` | `.pf-c-notification-drawer__list` | Indicates that the group notification list is hidden so that it isn't visible in the UI and isn't accessed by assistive technologies. |
+| `tabindex="0"` | `.pf-c-notification-drawer__list-item.pf-m-clickable` | Inserts the clickable list item into the tab order of the page so that it is focusable. |
 | `aria-hidden="true"` | `.pf-c-notification-drawer__group-toggle-icon > *`, `.pf-c-notification-drawer__list-item-header-icon > *` | Hides icon for assistive technologies. |
 
 ### Usage
@@ -128,3 +129,10 @@ cssPrefix: pf-c-notification-drawer
 | `.pf-c-notification-drawer__group-title` | `<div>` | Initiates a notification group toggle title. |
 | `.pf-c-notification-drawer__group-count` | `<div>` | Initiates a notification group toggle count. |
 | `.pf-c-notification-drawer__group-icon` | `<span>` | Initiates a notification group toggle icon. |
+| `.pf-m-info` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the info state. |
+| `.pf-m-warning` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the warning state. |
+| `.pf-m-danger` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the danger state. |
+| `.pf-m-success` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the success state. |
+| `.pf-m-read` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the read state. |
+| `.pf-m-clickable` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item to indicate that it is clickable. |
+| `.pf-m-expanded` | `.pf-c-notification-drawer__group` | Modifies a notification group for the expanded state. |
--- a/src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
+++ b/src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
@@ -38,62 +38,64 @@ cssPrefix: pf-c-notification-drawer
     {{/notification-drawer-header-action}}
   {{/notification-drawer-header}}
   {{#> notification-drawer-body}}
-    {{#> notification-drawer-group notification-drawer--id=(concat notification-drawer--id '-group1')}}
-      <h1>
-        {{#> notification-drawer-group-toggle}}
-          {{#> notification-drawer-group-toggle-title}}
-            First notification group
-          {{/notification-drawer-group-toggle-title}}
-          {{#> notification-drawer-group-toggle-count}}
-            {{#> badge badge--modifier="pf-m-unread"}}2{{/badge}}
-          {{/notification-drawer-group-toggle-count}}
-          {{> notification-drawer-group-toggle-icon}}
-        {{/notification-drawer-group-toggle}}
-      </h1>
-      {{> notification-drawer-basic-list}}
-    {{/notification-drawer-group}}
-    {{#> notification-drawer-group notification-drawer--id=(concat notification-drawer--id '-group2') notification-drawer-group--IsExpanded="true"}}
-      <h1>
-        {{#> notification-drawer-group-toggle}}
-          {{#> notification-drawer-group-toggle-title}}
-            Second notification group
-          {{/notification-drawer-group-toggle-title}}
-          {{#> notification-drawer-group-toggle-count}}
-            {{#> badge badge--modifier="pf-m-unread"}}2{{/badge}}
-          {{/notification-drawer-group-toggle-count}}
-          {{> notification-drawer-group-toggle-icon}}
-        {{/notification-drawer-group-toggle}}
-      </h1>
-      {{> notification-drawer-basic-list}}
-    {{/notification-drawer-group}}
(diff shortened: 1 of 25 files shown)


mcarrano:
@mcoker This is looking much better following recent changes. I was comparing this to the design doc here: https://docs.google.com/document/d/1kccmdwXgnkLZdH5OY_sr6C91Gsm0gCtF6zp9OBwC-dE/edit?usp=sharing  One thing that is missing is that the unread item should be clickable as we want the user to be able to mark as read with a single click and not need to open the kabob.

@maryshak1996 can you also take a look at this? One thing I'm wondering is whether we should have hover states for elements that are clickable, like an unread notification item. Seems like the group headers should also have a hover as they do in the Accordion component. In fact, should we just be reusing the Accordion here @mcoker ?

mcoker:
@mcarrano updated the PR so that the unread items can receive keyboard focus and have a cursor pointer to indicate they're interactive.

> In fact, should we just be reusing the Accordion here

I suppose it depends on how far the group list and accordion component end up diverging. At the moment there would be quite a bit of overrides to the component to use it in the notification drawer, then we create a dependency between those 2 components. I would say that with what we have currently, the styling differences are enough that I think we can create this as a standalone panel in the notification drawer.

christiemolloy:
@mcarrano @maryshak1996 if the title wraps like this should the paragraph of text underneath it be tabbed over more?

Screen Shot 2019-12-13 at 10 20 41 AM

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
```diff
@@ -0,0 +1,138 @@
...
+| `.pf-c-notification-drawer__list-item-header-title` | `<h2>` | Initiates a notification list item header title. |
+| `.pf-c-notification-drawer__list-item-action` | `<div>` | Initiates a notification list item action. |
+| `.pf-c-notification-drawer__list-item-description` | `<div>` | Initiates a notification list item description. |
+| `.pf-c-notification-drawer__list-item-timestamp` | `<div>` | Initiates a notification list item timestamp. |
+| `.pf-c-notification-drawer__group` | `<section>` | Initiates a notification group. |
+| `.pf-c-notification-drawer__group-toggle` | `<button>` | Initiates a notification group toggle. |
+| `.pf-c-notification-drawer__group-title` | `<div>` | Initiates a notification group toggle title. |
+| `.pf-c-notification-drawer__group-count` | `<div>` | Initiates a notification group toggle count. |
+| `.pf-c-notification-drawer__group-icon` | `<span>` | Initiates a notification group toggle icon. |
+| `.pf-m-info` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the info state. |
```
christiemolloy: Is it necessary to have a modifier class on the list-item?

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
```diff
@@ -0,0 +1,138 @@
...
+| `.pf-m-warning` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the warning state. |
+| `.pf-m-danger` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the danger state. |
+| `.pf-m-success` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the success state. |
+| `.pf-m-read` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item for the read state. |
+| `.pf-m-clickable` | `.pf-c-notification-drawer__list-item` | Modifies a notification list item to indicate that it is clickable. |
```
christiemolloy: should this be `pf-m-selectable` instead?

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -0,0 +1,265 @@
...
+  // Group toggle count
+  --pf-c-notification-drawer__group-toggle-count--MarginRight: var(--pf-global--spacer--md);
+
+  // Group toggle icon
+  --pf-c-notification-drawer__group-toggle-icon--MarginRight: var(--pf-global--spacer--md);
+  --pf-c-notification-drawer__group-toggle-icon--Color: var(--pf-global--Color--200);
+  --pf-c-notification-drawer__group-toggle-icon--Transition: .2s ease-in 0s;
+  --pf-c-notification-drawer__group--m-expanded__group-toggle-icon--Transform: rotate(90deg);
+
+  display: flex;
```
christiemolloy: Is this necessary?

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
```diff
@@ -0,0 +1,130 @@
...
+| `.pf-c-notification-drawer__header-action` | `<div>` | Initiates the notification drawer header action. |
+| `.pf-c-notification-drawer__body` | `<div>` | Initiates the notification drawer body. |
+| `.pf-c-notification-drawer__list` | `<ul>` | Initiates a notification list. |
+| `.pf-c-notification-drawer__list-item` | `<li>` | Initiates a notification list item. |
+| `.pf-c-notification-drawer__list-item-header` | `<div>` | Initiates a notification list item header. |
+| `.pf-c-notification-drawer__list-item-header-icon` | `<span>` | Initiates a notification list item header icon. |
+| `.pf-c-notification-drawer__list-item-header-title` | `<h2>` | Initiates a notification list item header title. |
+| `.pf-c-notification-drawer__list-item-action` | `<div>` | Initiates a notification list item action. |
+| `.pf-c-notification-drawer__list-item-description` | `<div>` | Initiates a notification list item description. |
+| `.pf-c-notification-drawer__list-item-timestamp` | `<div>` | Initiates a notification list item timestamp. |
```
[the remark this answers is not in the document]
mcoker: Agree it's specific. This could be a general `__list-item-footer` element, wdyt?

src/patternfly/components/NotificationDrawer/notification-drawer-header-title.hbs
```diff
@@ -0,0 +1,6 @@
+<h1 class="pf-c-notification-drawer__header-title{{#if notification-drawer-header-title--modifier}} {{notification-drawer-header-title--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
mcoker: Where do you think it could go instead?

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
mcoker: Yep to indicate what kind of notification it is.

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
mcoker: I wondered that, and if we might have a "selectable" state in the future that might be separate from what we're offering here, which is to indicate that the item is clickable. The way I envision it currently in the react component is the item can attach click events to it like, on click, add `.pf-m-read`, or execute a custom callback function. That's what this class adds are styles to say "you can click this and something will happen."

And I wonder if something could be both selectable and clickable, with different styles. Maybe `pf-m-hoverable`? We use that in other places to indicate hover styles to say something is interactive.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
mcoker: Yeah, so that the `__header` will stay fixed and `__body` will be the scrollable item if the viewport height is shorter than the overall list of things in the notification drawer. Or do you think we can achieve that without it?

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -0,0 +1,285 @@
...
+
+    position: relative;
+  }
+
+  &.pf-m-hoverable {
+    cursor: pointer;
+
+    &:hover {
+      z-index: 100;
+      box-shadow: var(--pf-global--BoxShadow--md-top), var(--pf-global--BoxShadow--md-bottom);
```
mattnolting: Should these be `--pf-c-notification-drawer__list-item--m-hoverable--BoxShadow` and `--pf-c-notification-drawer__list-item--m-hoverable--ZIndex`

PR #2541, opened 2019-12-17
feat(page): add hover focus state to icons in header

### LINKED ISSUE #1998: Focus state for icons in masthead (notification badge)

christiemolloy:
closes #1998 

Screen Shot 2019-12-17 at 3 06 18 PM

Screen Shot 2019-12-17 at 3 06 23 PM


src/patternfly/components/Page/page.scss
```diff
@@ -285,6 +295,40 @@
...
+
+    > .pf-c-button {
+      width: var(--pf-c-page__header-tools-group--m-icons--c-button--Width);
+      height: var(--pf-c-page__header-tools-group--m-icons--c-button--Height);
+      padding: unset;
+      background-color: var(--pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor);
+      border-radius: var(--pf-c-page__header-tools-group--m-icons--c-button--BorderRadius);
+
+      > i {
+        vertical-align: -.75px;
```
christiemolloy: will make this a variable if we think its needed

maryshak1996:
@christiemolloy this looks great!

src/patternfly/components/Page/page.scss
```diff
@@ -285,6 +295,40 @@
...
+        --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: var(--pf-c-page__header-tools-group--m-icons--c-button--active--BackgroundColor);
+      }
+
+      &:focus,
+      &.pf-m-focus {
+        --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: var(--pf-c-page__header-tools-group--m-icons--c-button--focus--BackgroundColor);
+
+        outline-offset: var(--pf-c-page__header-tools-group--m-icons--c-button--focus--OutlineOffset);
+
+        /* stylelint-disable */
```
mcoker: this block doesn't appear to be necessary since the border color is `var(--pf-global--BackgroundColor--100)` which changes to black in a dark theme.

src/patternfly/components/Page/page.scss
```diff
@@ -285,6 +295,40 @@
...
     }
+
+    > .pf-c-button {
+      width: var(--pf-c-page__header-tools-group--m-icons--c-button--Width);
+      height: var(--pf-c-page__header-tools-group--m-icons--c-button--Height);
+      padding: unset;
+      background-color: var(--pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor);
+      border-radius: var(--pf-c-page__header-tools-group--m-icons--c-button--BorderRadius);
+
+      > i {
```
mcoker: is this necessary? if the icon isn't centered, could you use `display: flex; align-items: center; justify-content: center;` on the button?

src/patternfly/components/Page/page.scss
```diff
@@ -285,6 +295,40 @@
...
+      &:active,
+      &.pf-m-active {
+        --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: var(--pf-c-page__header-tools-group--m-icons--c-button--active--BackgroundColor);
+      }
+
+      &:focus,
+      &.pf-m-focus {
+        --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: var(--pf-c-page__header-tools-group--m-icons--c-button--focus--BackgroundColor);
+
+        outline-offset: var(--pf-c-page__header-tools-group--m-icons--c-button--focus--OutlineOffset);
```
mcoker: is this necessary? the focus ring seems OK without it.

src/patternfly/components/Page/page.scss
```diff
@@ -285,6 +295,40 @@
       display: block;
       visibility: visible;
     }
+
+    > .pf-c-button {
```
mcoker: This works, but where it will break down is if buttons are used in a `.pf-c-page__header-tools-group` element without `.pf-m-icons` - the modifier is only used to hide the group on mobile. So if someone defined a `.pf-c-page__header-tools-group` with buttons inside of it that they want to persist on mobile (by omitting `.pf-m-icons`), they wouldn't get the focus styles. Seems like it would be safe to to move this so it applies to any `> .pf-c-button` in a `.pf-c-page__header-tools-group`.

jeperry:
Does this PR handle focus, or focus and active? I'd like to ensure all of these use cases are possible even if there is no visual distinction in any of these. 

- user is hovering on a header icon 
- user is clicking on a header icon
- user has opened a menu from an header icon, and the focus state is on while that menu is open
- user clicks a header icon which is used as a button to go into a mode (ex: setting view) and the focus state remains on as long as they are in the settings view.

src/patternfly/components/Page/page.scss
christiemolloy: I think its necessary for the focus style though, because the black border doesn't work on the grey background, see image here:
Screen Shot 2020-01-07 at 2 36 51 PM

src/patternfly/components/Page/page.scss
christiemolloy: yep that helps a lot!

src/patternfly/components/Page/page.scss
christiemolloy: agreed, removed it

src/patternfly/components/Page/page.scss
christiemolloy: agreed, moved it outside of `.pf-m-icons`

christiemolloy:
@jeperry the button in the page header has active and focus styles which control the background color change. There is no background color change on hover for the icon as stated in #1998. For the last two cases that you mention, that does raise a good question about whether we should have a modifier like `.pf-m-open` or `.pf-m-expanded` so that it can stay selected while the menu is opened / in a mode. @mcoker @mattnolting thoughts?

mattnolting:
@mcoker @christiemolloy should this be abstracted?? Seems like this would be a desirable feature in other areas of the framework..

christiemolloy:
@mattnolting potentially but I know for now we only have a case for it in the page header tools. @mcarrano can you see a need for this anywhere else?

mcarrano:
This was pretty unique to the masthead (page header) area. Can't predict whether we will want to use this effect else where in the future.

src/patternfly/components/Page/page.scss
```diff
@@ -287,6 +296,37 @@
...
+    &.pf-m-active {
+      --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: var(--pf-c-page__header-tools-group--m-icons--c-button--active--BackgroundColor);
+    }
+
+    &:focus,
+    &.pf-m-focus {
+      --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: var(--pf-c-page__header-tools-group--m-icons--c-button--focus--BackgroundColor);
+
+      /* stylelint-disable */
+      .pf-c-notification-badge {
```
mcoker: looks like this style should also apply to the `:active` state, since this is what it looks like now

Screen Shot 2020-01-10 at 12 09 12 PM

Also we should move `:active` after `:focus` since they modify the same property and an active element is also focused, otherwise `:active` styles will will always be overwritten by the `:focus` styles.

src/patternfly/components/Page/page.scss
```diff
@@ -287,6 +296,37 @@
     }
   }
 
+  .pf-c-button {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    width: var(--pf-c-page__header-tools-group--m-icons--c-button--Width);
+    height: var(--pf-c-page__header-tools-group--m-icons--c-button--Height);
```
mcoker: I saw some designs (that may or may not make their way into PF) that would imply we need to support this button focus state potentially being pill shaped if there is a lot of content in the button. Here are a couple of screenshots.

I wonder if we could support that without much effort in the original implementation? If not it's always an enhancement we could make later if needed.

Screen Shot 2020-01-10 at 12 16 39 PM

src/patternfly/components/Page/page.scss
christiemolloy: sounds good, updated

src/patternfly/components/Page/page.scss
christiemolloy: Assuming the badge and notification would be part of the same button, we could possibly have a selector for multiple elements in the button and then change the width/height/border-radius accordingly. Maybe there would be a variation of the notification badge that supports the badge component too and handles the spacing there. Either way, I think we can just update properties based on whats inside of the notification badge, with the current implementation.

mcoker: That's one way to do it, though I wonder if it could just adapt automatically. I changed the button height/width to min-height/min-width and gave the button sm (8px) padding here. Not sure if something like that would work?

Screen Shot 2020-01-10 at 4 27 09 PM
Screen Shot 2020-01-10 at 4 27 03 PM
Screen Shot 2020-01-10 at 4 26 56 PM

mcarrano:
I have a question about this PR. Looks like it is designed to apply the visual affect on focus and well as when the icon is selected/active. Is that right? This does not seem right to me. The intent is to provide a state that indicated that the feature is active (i.e. the drawer is open in this case). I don't think that we typically change the state of other icons on focus, do we? Isn't focus usually just indicated by the focus rectangle that surrounds the element? To say this another way, I would not expect tabbing through the interface to open the drawer. That should only happen if the user clicks on the element.

christiemolloy:
@mcarrano i thought the original intention of the issue was to update the active and focus state of all icons in the masthead, as Mary had stated: "this is for ALL masthead icons, including notification bell". But I think what you're saying is that the icons should only have this background circle when they have been opened? If so, I can create a modifier that can be applied to an icon that is used to open things such as the notification drawer, and it can take on the styling. Therefore there will be no active/focus style changes, it will only change when the icon has opened something?

src/patternfly/components/Page/page.scss
christiemolloy: oh thats a good point, seems easier.

maryshak1996:
@christiemolloy yes, what @mcarrano said is correct -- this treatment is for when a masthead utility item has been opened (be that in the form a dropdown, side panel, or any other potential views)

christiemolloy:
got it, thanks @maryshak1996 ! So just confirming that a masthead utility item can be multiple types of icons?

maryshak1996:
@christiemolloy at least based on how I'm using the term, yes! I think that what we want here, is to just have a way of indicating any 'opened' icon in the masthead, since it is possible that multiple could be opened at once

christiemolloy:
i updated the examples. so now in order to have the styling when the icon is opened you would add the modifier `pf-m-opened` to `pf-c-button` in the page component. I removed the example from the page component as this may be confusing because it wasn't demonstrating opening anything, so now it is under `demos/drawer` and you'll see the styling in the expanded version. let me know if this aligns with what you guys were thinking @mcarrano @maryshak1996

mcarrano:
Yes, I think this makes sense now @christiemolloy Do you agree @maryshak1996 ?

christiemolloy:
@mcarrano spoke to Mary in person and she agrees that it works as it should now

src/patternfly/components/Page/page.scss
```diff
@@ -287,6 +297,29 @@
     }
   }
 
+  .pf-c-button.pf-m-opened {
```
mattnolting: There's a noticeable jump when `.pf-m-opened` is applied to `.pf-c-button` because the widths are applied to the modifier rather than the base button. `.pf-m-opened` needs to apply to the default `.pf-c-button` to avoid shift.

src/patternfly/components/Page/page.scss
```diff
@@ -65,6 +65,16 @@
...
+  // Header tools group button
+  --pf-c-page__header-tools-group--m-icons--c-button--Width: #{pf-size-prem(36px)};
+  --pf-c-page__header-tools-group--m-icons--c-button--Height: #{pf-size-prem(36px)};
+  --pf-c-page__header-tools-group--m-icons--c-button--BorderRadius: var(--pf-global--BorderRadius--lg);
+  --pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor: transparent;
+  --pf-c-page__header-tools-group--m-icons--c-button--active--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-page__header-tools-group--m-icons--c-button--focus--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-page__header-tools-group--m-icons--c-button--focus__c-notification-badge--m-unread--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-page__header-tools-group--m-icons--c-button--active__c-notification-badge--m-unread--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
+
```
mattnolting: Looks like we need `--pf-c-page__header-tools-group--m-icons--c-button--PaddingLeft` and `--pf-c-page__header-tools-group--m-icons--c-button--PaddingRight`.

src/patternfly/components/Page/page.scss
mattnolting: `.pf-m-opened` is distinct to something being `open` in the UI, I think we can abstract to use the active state `.pf-m-active` here instead.

src/patternfly/components/Page/page.scss
```diff
@@ -287,6 +297,29 @@
...
   }
 
+  .pf-c-button.pf-m-opened {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    width: var(--pf-c-page__header-tools-group--m-icons--c-button--Width);
+    height: var(--pf-c-page__header-tools-group--m-icons--c-button--Height);
+    padding: unset;
+    background-color: var(--pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor);
```
mattnolting: Since `--pf-c-page__header-tools-group--m-icons--c-button--BackgroundColor` is transparent, this can be set on the unmodified button.

src/patternfly/components/Page/page.scss
christiemolloy: I do agree with your comment about open being too specific as a name. But I'm wondering if `pf-m-active` would be confused with the `:active` selector? 

@mcarrano @maryshak1996 any thoughts here?

christiemolloy: oh yes thanks for pointing that out!

mcarrano: I don't really have an opinion @christiemolloy . Maybe a better question for devs?

src/patternfly/components/Page/page.scss
christiemolloy: Ah I had unset padding and given the button a width and height instead... just changed it so that it uses `:after` to draw the circle background

christiemolloy:
Updated @mattnolting

mcoker:
@mattnolting agree that `.pf-m-opened` is too specific, but we preserve `.pf-m-active` for the `:active` pseudo class like @christiemolloy mentioned. I would ultimately be in favor of removing the classes we use to represent pseudo classes, as most/all of them are just for our workspace and IMO that shouldn't ship with PF. What about `pf-m-selected`, that seems appropriate? And that could apply to one or more header elements at the same time with different purposes.

mattnolting:
> Yeah pf-m-selected works for me. Thoughts @mattnolting

wfm!

christiemolloy:
updated @mattnolting @mcoker

src/patternfly/components/Page/page.scss
mattnolting: The problem with setting a strict `height/width` is that if `font-size` changes the icon size it breaks the background treatment.

src/patternfly/components/Page/page.scss
```diff
@@ -287,6 +295,39 @@
...
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        content: "";
+        background-color: var(--pf-c-page__header-tools--c-button--m-selected--before--BackgroundColor);
+        border-radius: var(--pf-c-page__header-tools--c-button--m-selected--before--BorderRadius);
+      }
+
+      /* stylelint-disable */
```
mcoker: Can you check in this file and convert any `/* comment */` line to `// comment`? That'll make it a sass comment which won't get compiled into patternfly.css

PR #2545, opened 2019-12-18
feat(empty-state): xl variation

### LINKED ISSUE #2472: Introduce Extra-Large empty state variant
We want to add a new empty state variant to be used for certain full screen empty state patterns like welcome and success screens.

Final designs here: https://marvelapp.com/504hec7

cc: @mcoker @rachael-phillips @LHinson

christiemolloy:
closes #2472

### DIFF
--- a/src/patternfly/components/EmptyState/empty-state.scss
+++ b/src/patternfly/components/EmptyState/empty-state.scss
@@ -88,7 +88,7 @@
 // RedHat Font overrides
 @include pf-m-redhat-font {
   .pf-c-empty-state {
-    > .pf-c-title:not(.pf-m-4xl) {
+    > .pf-c-title.pf-m-lg {
       font-size: var(--pf-global--FontSize--xl);
     }
   }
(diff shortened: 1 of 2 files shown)


src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -34,7 +40,28 @@
   &.pf-m-lg {
     max-width: var(--pf-c-empty-state--m-lg--MaxWidth);
   }
+
+  &.pf-m-xl {
+    > .pf-c-button.pf-m-primary,
```
christiemolloy: @mcoker @mattnolting is there a better way to write this selector

mcoker: Yeah I think so, I submitted a PR if you want to take a look and compare https://github.com/christiemolloy/patternfly-next/pull/19

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -64,7 +80,7 @@
 // RedHat Font overrides
 @include pf-m-redhat-font {
   .pf-c-empty-state {
-    > .pf-c-title {
+    > .pf-c-title:not(.pf-m-4xl) {
```
mcoker: The original selector here actually might be problematic. Since we don't set the `font-size` for the empty state title in the component's css, and instead use the title component's modifier to set that, users can modify the `font-size` if they wish by using a different title size modifier. If they've opted into the new font, we're overriding that behavior and forcing the font-size to `--pf-global--FontSize--xl` regardless the title modifier.

The intention here was just to change the default empty state title size from `lg` to `xl` when opting in to the new font without needing a markup change (modifying the title component's modifier). I think we could change this from `> .pf-c-title` to `> .pf-c-title.pf-m-lg` and consider that change fixing a bug, then we wouldn't have a need to do anything special for the xl variation title. wdyt?

PR #2548, opened 2019-12-19
feat(button): add ability to have icon on left or ride side of text

### LINKED ISSUE #2507: Add ability to add an icon to the link button to the end of the button.
Issue was tracked in React first: https://github.com/patternfly/patternfly-react/issues/3283 but we need to add styling to the icon so that it will have a margin-left if its the first-child.

christiemolloy:
closes #2507 

@mcoker @mattnolting Instead of making a new class for the text inside of the link button, I just left it as a span. In a future breaking change release I was thinking that we could remove the class around the button, and have a class around the text instead, which can control the margin-left and margin-right. This way we only need one class, rather than two.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -110,7 +110,7 @@
 
   // Styles for an icon in button
   --pf-c-button__icon--MarginRight: var(--pf-global--spacer--xs);
-  --pf-c-button__icon--MarginLeft: var(--pf-global--spacer--xs);
+  --pf-c-button__text--icon--MarginLeft: var(--pf-global--spacer--xs);
 
   position: relative;
   display: inline-block;
@@ -124,12 +124,13 @@
   border: 0;
   border-radius: var(--pf-c-button--BorderRadius);
 
-  .pf-c-button__text + .pf-c-button__icon {
-    margin-left: var(--pf-c-button__icon--MarginLeft);
+  .pf-c-button__icon {
+    margin-right: var(--pf-c-button__icon--MarginRight);
   }
 
-  .pf-c-button__icon:first-child:not(:only-child) {
-    margin-right: var(--pf-c-button__icon--MarginRight);
+  .pf-c-button__text + .pf-c-button__icon {
+    margin-right: 0;
+    margin-left: var(--pf-c-button__icon--MarginLeft);
   }
 
   &::after {
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -110,6 +110,7 @@
 
   // Styles for an icon in button
   --pf-c-button__icon--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-button__icon--MarginLeft: var(--pf-global--spacer--xs);
 
   position: relative;
   display: inline-block;
@@ -124,7 +125,13 @@
   border-radius: var(--pf-c-button--BorderRadius);
 
   .pf-c-button__icon {
-    margin-right: var(--pf-c-button__icon--MarginRight);
+    &:first-child {
+      margin-right: var(--pf-c-button__icon--MarginRight);
+    }
+
+    &:last-child {
+      margin-left: var(--pf-c-button__icon--MarginLeft);
+    }
   }
 
 
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -124,16 +124,13 @@
   border: 0;
   border-radius: var(--pf-c-button--BorderRadius);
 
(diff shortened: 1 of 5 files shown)


mcoker:
This would be a breaking change as-is because without the span, we would be creating space on either side of the icon - https://codepen.io/mcoker/pen/zYxZwpO

We could add `.pf-c-button__text` to the `<span>`, then do `.pf-c-button__text + .pf-c-button__icon { margin-left: whatever; }` as a way to opt allow users to opt in to the change with new markup.

src/patternfly/components/Button/button.scss
```diff
@@ -123,10 +124,13 @@
   border: 0;
   border-radius: var(--pf-c-button--BorderRadius);
 
-  .pf-c-button__icon {
-    margin-right: var(--pf-c-button__icon--MarginRight);
+  .pf-c-button__text + .pf-c-button__icon {
+    margin-left: var(--pf-c-button__icon--MarginLeft);
   }
 
+  .pf-c-button__icon:first-child:not(:only-child) {
```
mcoker: This introduces a breaking change because it disables the existing right margin when the icon is placed in front of text. I think you'll need to leave the original selector as it is, then set `margin-right: 0;` when it's `__text + __icon`

src/patternfly/components/Button/button.scss
```diff
@@ -110,6 +110,7 @@
 
   // Styles for an icon in button
   --pf-c-button__icon--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-button__icon--MarginLeft: var(--pf-global--spacer--xs);
```
mcoker: Can we rename this to `--pf-c-button__text--icon--MarginLeft`? The existing var would be a conflict if we ever wanted to add a regular `margin-left` to `--pf-c-button__icon`.

src/patternfly/components/Button/button.scss
christiemolloy: Ok updated, but what if there is just one button icon with no text, then it still has a margin-right.

src/patternfly/components/Button/button.scss
christiemolloy: yep good point!

src/patternfly/components/Button/button.scss
```diff
@@ -127,6 +128,10 @@
     margin-right: var(--pf-c-button__icon--MarginRight);
   }
 
+  .pf-c-button__text + .pf-c-button__icon {
+    margin-right: 0;
+    margin-left: var(--pf-c-button__icon--MarginLeft);
+  }
```
mattnolting: Instead of `.pf-c-button__text + . pf-c-button__icon`, what about:

```
> * {
  margin-right: var(--pf-c-button--item--MarginRight);
}

> :last-child {
  --pf-c-button--item--MarginRight: 0;
}
```

**Why?** 
- Maintains base specificity.
- Don't have to worry about swapping `margin-right/left`.
- Simpler

src/patternfly/components/Button/button.scss
```diff
@@ -110,6 +110,7 @@
 
   // Styles for an icon in button
   --pf-c-button__icon--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-button__text--icon--MarginLeft: var(--pf-global--spacer--xs);
```
mattnolting: Instead of `.pf-c-button__text + . pf-c-button__icon`, what about:

```
// change during breaking change period
--pf-c-button--item--MarginRight: var(--pf-c-button__icon--MarginRight);
```

src/patternfly/components/Button/button.scss
mcoker: That would change the natural spacing if people have their own elements in the button, like `<button><span>word</span> <span>word</span></button>`

mattnolting: Right, I would expect that.

What about something like this when it comes along?
Screen Shot 2020-01-09 at 11 30 43 AM

mattnolting: @mcoker This solution works, but I think it's too specific/needs to be more flexible and we'll be revisiting in the future.

mcoker: > Right, I would expect that.

I wouldn't expect a space difference between `<button>word word</button>` and `<button><span>word</span> <span>word</span></button>` though, and we never know what a person will put in the button. I agree we'll want to support space between other things in the future for sure, but as long as users can put anything they want in the button, I don't think a margin on every direct child is the best approach.

mattnolting: Ship it

christiemolloy:
added classname to docs, and created follow-up issue #2574

PR #2549, opened 2019-12-19
chore(fix): should not include "Main" property in package.json

ddonahue007:
if we are not using 'Main' property asn entry point it should be removed, it is not a required property

In the case of the codesandbox it is causing the following error

"Error: Could not fetch dependencies, please try again in a couple seconds: Something went wrong while packaging the
dependency @patternfly/patternfly@2.43.1: ENOENT: no such file or directory, scandir '/tmp/960130286/node_modules/@patternfly/patternfly/n'"

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,9 @@
 {
-  "name": "@patternfly/patternfly",
+  "name": "dd-patternfly",
   "description": "Assets, source, tooling, and content for PatternFly 4",
   "version": "0.0.0-development",
   "keywords": [],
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,10 +1,9 @@
 {
-  "name": "@patternfly/patternfly",
+  "name": "dd-patternfly",
```
dgutride: This needs to be set back to @patternfly/patternfly

PR #2551, opened 2019-12-20
fix(global-vars): added CSS vars for color palette SCSS vars

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2546


src/patternfly/_variables.scss
```diff
@@ -1,5 +1,82 @@
 :root {
   // Colors
+  // Global palette
+  --pf-global--Color--black-100: #{$pf-color-black-100};
```
christiemolloy: Should we be using these variables like `--pf-global--Color--black-100` now in places where we used values like `#{$pf-color-black-100}` ? i.e. charts?

mcoker: Yeah, great idea! I'll update them in this PR.

mcoker:
@mceledonia our global vars follow the format of `--pf-global--[thing]--[variation]`, like `--pf-global--Color--100` for our global text color and `--pf-global--Color--light-100` for the global "light" text color (white).

If we introduce these new color palette vars as `--pf-global--color--black-100`, `--pf-global--color--black-200`, etc, that will likely conflict with the existing `--pf-global--Color--*` variables that we use for text color.

I was thinking naming these vars something like `--pf-global--palette-color--black-100`, `--pf-global--palette-color--black-200` - what do you think? Do you have a name you think would work better?

mceledonia:
Michael and I talked about this over bluejeans and landed on the naming convention `--pf-global--palette--black-100`

src/patternfly/_variables.scss
mcoker: Only one I didn't update is `--pf-c-about-modal-box__close--c-button--hover--BackgroundColor: #{rgba($pf-color-black-1000, .4)};` since passing the css var to sass's `rgba()` function doesn't work.

PR #2553, opened 2019-12-23
fix(table): fix border

### LINKED ISSUE #2512: Bug - Mobile expand pane is missing styling
The expandable table has a blue line when expanded on the row and child expand pane. When responsive on mobile, this line is different.

Seen in product and also on the PF website
https://www.patternfly.org/v4/documentation/react/components/table
Xnip2019-12-10_08-13-51

Product example: https://github.com/RedHatInsights/insights-advisor-frontend/pull/500#issuecomment-564025255

christiemolloy:
closes #2512

### DIFF
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -331,6 +331,7 @@
     --pf-c-table-cell--PaddingRight: 0;
     --pf-c-table-cell--PaddingLeft: 0;
 
+    display: initial;
     max-height: var(--pf-c-table__expandable-row--MaxHeight);  // Overflow scroll should only happen on responsive
     overflow-y: auto;
     border-bottom: none;
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -298,7 +298,7 @@
   tbody {
     position: relative;
 
-    &::before {
+    &::after {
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Table/table-grid.scss
```diff
@@ -332,7 +331,7 @@
     --pf-c-table-cell--PaddingRight: 0;
     --pf-c-table-cell--PaddingLeft: 0;
 
-    display: block;
+    display: initial;
```
mcoker: This will set it to `display: inline;`, which works but I wonder if there's a better way to fix it since that element will no longer behave as a block level element (and I think it should).

It looks like the issue is due to the stacking context of `.pf-c-table__expandable-row` being placed above the blue border because the expandable row has `position: relative`. It doesn't look like `.pf-c-table__expandable-row` needs `position: relative;` in a grid (or desktop?) layout, so one option to could be to remove it.

Or another option could be - since the blue border is drawn with `.pf-m-grid-lg.pf-c-table tbody:before`, we could just switch that to `.pf-m-grid-lg.pf-c-table tbody:after` to fix the stacking context.

christiemolloy: Those are great ideas thanks, I went with changing the blue border to use `::after` instead of `::before`

mattnolting:
👍

PR #2558, opened 2020-01-06
fix(table): fixed bug with mobile inheritable grid-column property

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2557


mattnolting:
Side note: looks like we lost `table--id` prop on tables that need it. 

Screen Shot 2020-01-07 at 9 25 58 AM

Should we update that here or open a new issue?

mcoker:
@mattnolting doh! looks like it was lost in this MDX conversion https://github.com/patternfly/patternfly-next/issues/2242

updated

PR #2563, opened 2020-01-07
feat(input): add new example

### LINKED ISSUE #2299: Add input group examples with plain button modifier
We need to add examples to the input group component that show how to use a plain button in an input group

Screen Shot 2019-09-24 at 9 40 50 AM

@mcarrano can you tell us which examples we should add? Ie, should we just add the example above, or would you like to see something else, more examples, etc?

christiemolloy:
closes #2299 

@mcarrano @mcoker @mattnolting do you think I should add any documentation for when to use the plain modifier over the control modifier for the button next to the input?

### DIFF
--- a/src/patternfly/components/InputGroup/examples/InputGroup.md
+++ b/src/patternfly/components/InputGroup/examples/InputGroup.md
@@ -93,17 +93,17 @@ cssPrefix: pf-c-input-group
 {{/input-group}}
 <br>
 {{#> input-group}}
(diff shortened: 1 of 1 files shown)


christiemolloy:
Yes that makes a lot of sense to me @mcarrano

mattnolting:
> @mcarrano @mcoker @mattnolting do you think I should add any documentation for when to use the plain modifier over the control modifier for the button next to the input?

Good question. I think the search example illustrates when to use `-control`, but the next example, with 'help' as the `-control`, doesn't make sense. I think of it as `-control` having a direct effect on the adjacent input. I think your addition would replace the current help example.

PR #2573, opened 2020-01-10
feat(simplelist): introduce simple list component

### LINKED ISSUE #2468: Introduce Simple List component for master-detail pattern
One variation of the master-detail pattern calls for a simple list component in the master panel that can be implemented as a compact variant of the simple light nav. An example of what that looks like can be seen here: https://marvelapp.com/9iefh5b/screen/63567619 This is useful in cases where use of a DataList as the master is not appropriate. The component should support both grouped and ungrouped items.

cc: @mcoker @rachael-phillips @LHinson

christiemolloy:
closes #2468


src/patternfly/components/SimpleList/examples/SimpleList.md
```diff
@@ -0,0 +1,102 @@
...
+| -- | -- | -- |
+| `.pf-c-simple-list` | `<div>` | Initiates a simple list. |
+| `.pf-c-simple-list__section` | `<section>` | Initiates a simple list section. |
+| `.pf-c-simple-list__title` | `<h2>` | Initiates a simple list title. |
+| `.pf-c-simple-list__list` | `<ul>` | Initiates a simple list unordered list. |
+| `.pf-c-simple-list__item` | `<li>` | Initiates a simple list item. |
+| `.pf-c-simple-list__item-button` | `<button>` | Initiates a simple list item button. |
+| `.pf-m-hover` | `.pf-c-simple-list__item-button` | Modifies the simple list item button for the hovered state. |
+| `.pf-m-active` | `.pf-c-simple-list__item-button` | Modifies the simple list item button for the active state. |
+| `.pf-m-focus` | `.pf-c-simple-list__item-button` | Modifies the simple list item button for the focussed state. |
```
mattnolting: I think `hovered` and `focussed` => `hover` and `focus`.

lboehling: Font weight shouldn't change on hover, the color should just change to blue (w/ gray hover band). Active and focus font-weight will change to medium.

lboehling:
@mcarrano Font weight shouldn't change on hover, the color should just change to blue (w/ gray hover band). Active and focus font-weight will change to medium. If it still changes the width too much when active items change to medium, we could just use the gray band on hover with the text remaining reg/black, and then change the active state text to blue but w/o changing any weights.

src/patternfly/components/SimpleList/examples/SimpleList.md
christiemolloy: updated!

mattnolting:
Perfect! 🎉

mcarrano:
I guess that I am OK that the font weight changes (and makes the item get wider) on focus. Looks like that's what we've done before as you point out @christiemolloy 

Are you good with this @lboehling ? If so, I will approve.

lboehling:
i'm good with it @mcarrano

src/patternfly/components/SimpleList/examples/SimpleList.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-simple-list` | `<div>` | Initiates a simple list. |
+| `.pf-c-simple-list__section` | `<section>` | Initiates a simple list section. |
+| `.pf-c-simple-list__title` | `<h2>` | Initiates a simple list title. |
+| `.pf-c-simple-list__list` | `<ul>` | Initiates a simple list unordered list. |
+| `.pf-c-simple-list__item` | `<li>` | Initiates a simple list item. |
+| `.pf-c-simple-list__item-button` | `<button>` | Initiates a simple list item button. |
```
mcoker: I wonder if `__item-button` is too restrictive. I think it would be valid for this to be an `<a>` when, for example, the link updates the URL and that URL can load the page with one of these links/buttons selected.

In our current dropdown, select, etc menus, the `<li>` doesn't have a BEM class, and the link/button inside is `__item`, which is a nice and generic name, but I like giving the `<li>` a class in case that we need to style it, and `__item` or `__list-item` seems like the most appropriate name.

I wonder if we should call it `__item-link`? That seems a little more generic than `__item-button`, but that may just be me :) wdyt? 

FWIW calling the link/button `__item-link` would match the nav component, though `__item-link` is obviously appropriate in the nav component as it will almost always contain links.

mcoker:
@mcarrano @lboehling looks like we need to add a "current" variation of an item in the list to denote the currently selected item (similar to nav & tabs) - can you verify?

mcarrano:
I thought we did have that state @mcoker . Agree that we need it. @lboehling can you confirm whether it is part of the design? 

I wasn't able to view the preview anymore. The preview link seems to no longer work. Anyone else having this problem?

mcoker:
@mcarrano we have `.pf-m-active`, but that's something else (shouldn't be used by a dev, it's only used to mimic a CSS pseudo class). We'll need to add `.pf-m-current` assuming only one item can be active at a time. The preview link still works for me.

lboehling:
Yeah, there should be a current state--only one item can be active at a time. This would be the same as the "active" state I described in the mocks (medium weight, blue text, gray bkg). @mcarrano @mcoker

mcoker:
@lboehling @mcarrano thanks! @christiemolloy ^^ 😁

christiemolloy:
@mcoker added `pf-m-current` and updated the classname `__button` to `__link` as per our discussion on slack yesterday.. also added this issue to update tabs: #2594

mcoker:
@christiemolloy thanks, the changes look good, but the "grouped list" example isn't showing properly in the preview build. Can you verify if it's an actual issue or just an issue with the build? 

Screen Shot 2020-01-22 at 12 07 03 PM

christiemolloy:
Looks like I still had an instance of `__button` in there thats why it wasn't loading @mcoker

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -0,0 +1,83 @@
...
+  --pf-c-simple-list__title--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-simple-list__title--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-simple-list__title--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-simple-list__title--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-simple-list__title--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-simple-list__title--Color: var(--pf-global--Color--dark-200);
+  --pf-c-simple-list__title--FontWeight: var(--pf-global--FontWeight--semi-bold);
+
+  // Simple list section
+  --pf-c-simple-list__section-section--MarginTop: var(--pf-global--spacer--sm);
```
mcoker:
```suggestion
  --pf-c-simple-list__section--section--MarginTop: var(--pf-global--spacer--sm);
```

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -0,0 +1,83 @@
...
+
+.pf-c-simple-list__title {
+  padding: var(--pf-c-simple-list__title--PaddingTop) var(--pf-c-simple-list__title--PaddingRight) var(--pf-c-simple-list__title--PaddingBottom) var(--pf-c-simple-list__title--PaddingLeft);
+  font-size: var(--pf-c-simple-list__title--FontSize);
+  font-weight: var(--pf-c-simple-list__title--FontWeight);
+  color: var(--pf-c-simple-list__title--Color);
+}
+
+.pf-c-simple-list__section + .pf-c-simple-list__section {
+  margin-top: var(--pf-c-simple-list__section-section--MarginTop);
```
mcoker:
```suggestion
  margin-top: var(--pf-c-simple-list__section--section--MarginTop);
```

PR #2577, opened 2020-01-14
fix(table): update font-size for check

### LINKED ISSUE #2518: Table checkbox sizes
Currently the checkboxes in tables adapt with the font size of the cells. So if the font size is bigger/smaller, the checkboxes change to accommodate. 

Would it be possible to update the table so that the checkboxes have a consistent size regardless the font-size used in the table?

Currently this would change:

1. Checkboxes in the `<thead>` (column headers) in regular tables, so that the checkbox there would match the size of the checkboxes below it. 

Here is what it looks like currently:
Screen Shot 2019-12-12 at 5 02 58 PM

And here is what it looks like after the change:
Screen Shot 2019-12-12 at 5 02 29 PM

2. Checkboxes in the compact table, so that the checkboxes would be the same size between compact and regular tables. This would also improve accessibility as the target area for a checkbox is already really small.

Here is what it looks like currently:
Screen Shot 2019-12-12 at 5 02 56 PM

And here is what it looks like after the change:
Screen Shot 2019-12-12 at 5 02 14 PM

cc @mcarrano

christiemolloy:
closes #2518

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -68,6 +68,7 @@
 
   // Table cell - check
   --pf-c-table__check--input--MarginTop: #{pf-size-prem(3px)};
+  --pf-c-table__check--input--FontSize: var(--pf-global--FontSize--md);
 
   // Table cell - action
   --pf-c-table__action--PaddingTop: 0;
@@ -313,6 +314,7 @@
 
   > * {
     margin-top: var(--pf-c-table__check--input--MarginTop);
+    font-size: var(--pf-c-table__check--input--FontSize);
     vertical-align: top;
   }
 }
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -313,6 +314,7 @@
 
   > * {
     margin-top: var(--pf-c-table__check--input--MarginTop);
+    font-size: var(--pf-c-table__check--input--FontSize);
```
mcoker: wdyt about redefining `--pf-c-table-cell--FontSize` in `.pf-c-table__check` instead? So move this up to the `.pf-c-table__check` selector and change to `--pf-c-table-cell--FontSize: var(--pf-c-table__check--input--FontSize);`

PR #2578, opened 2020-01-14
fix(card): add wrapper to img

### LINKED ISSUE #2562: Card: Need to allow proper resizing
When resizing the viewport of the images and actions card, I noticed that at 771 x 686 the action items extend past the card. The image + actions area is a flex layout and things aren't always allowed to shrink naturally in a flex layout like you might expect them to. 
We should allow whatever is beside the actions to shrink in size. 

What I saw on org and react workspace: 
Screen Shot 2020-01-06 at 4 09 22 PM

Screenshots from core:
Screen Shot 2020-01-06 at 3 32 44 PM

What should happen: 
Screen Shot 2020-01-06 at 3 32 21 PM

christiemolloy:
closes #2562 

There may be a few solutions to this problem, I went with the solution of wrapping the image in its own container, because we don't know how the user will give it height/width. So even if width is added to the image in the html or the svg, it will still size proportionally down. But interested to hear what you guys think @mcoker @mattnolting

### DIFF
--- a/src/patternfly/components/Card/card-image.hbs
+++ b/src/patternfly/components/Card/card-image.hbs
@@ -1,6 +1,0 @@
-<div class="pf-c-card__main{{#if card-main--modifier}} {{card-main--modifier}}{{/if}}"
-  {{#if card-main--attribute}}
-    {{{card-main--attribute}}}
-  {{/if}}>
-  {{> @partial-block}}
-</div>
--- a/src/patternfly/components/Card/card-image.hbs
+++ b/src/patternfly/components/Card/card-image.hbs
@@ -1,6 +1,6 @@
-<div class="pf-c-card__image{{#if card-image--modifier}} {{card-image--modifier}}{{/if}}"
-  {{#if card-image--attribute}}
-    {{{card-image--attribute}}}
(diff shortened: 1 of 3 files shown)


mcoker:
@christiemolloy I dig it, I think this makes the most sense. Though I wonder if we should be more generic in the element name in case someone wants to put some text or text+icon instead. We could call it `__main` or something like that?

christiemolloy:
I agree with using a less specific name. Also added class to the docs @mcoker

src/patternfly/components/Card/card-image.hbs
```diff
@@ -1,6 +1,6 @@
-<div class="pf-c-card__image{{#if card-image--modifier}} {{card-image--modifier}}{{/if}}"
-  {{#if card-image--attribute}}
-    {{{card-image--attribute}}}
+<div class="pf-c-card__main{{#if card-main--modifier}} {{card-main--modifier}}{{/if}}"
```
mcoker: can we call it `__head-main` instead, since it should only exist in `__head`? I think we should reserve `__main` in case we want to add that to the root of the card.

christiemolloy: yeah +1

PR #2579, opened 2020-01-14
feat(radio): add optional description vertical

### LINKED ISSUE #2526: Add new variant of Radio and Check component with optional stacked description
Want to introduce optional descriptions to the Radio and Check components per the design specification here: https://marvelapp.com/50eh328/screen/64407412 The component will support vertically stacked descriptions, only per the design.

cc @mcoker @rachael-phillips

christiemolloy:
closes #2526

### DIFF
--- a/src/patternfly/components/Check/examples/Check.md
+++ b/src/patternfly/components/Check/examples/Check.md
@@ -44,26 +44,15 @@ cssPrefix: pf-c-check
 {{/check}}
 ```
 
-```hbs title=With-vertical-description
+```hbs title=With-description
 {{#> check}}
   {{#> check-input check-input--attribute='id="check7" name="exampleCheck7"'}}{{/check-input}}
-  {{#> check-label check-label--attribute='for="check7"'}}Check{{/check-label}}
+  {{#> check-label check-label--attribute='for="check7"'}}Check with description{{/check-label}}
   {{#> check-description}}
     Single-tenant cloud service hosted and managed by Red Hat that offers high-availability enterprise-grade clusters in a virtual private cloud on AWS od GCP.
   {{/check-description}}
 {{/check}}
 ```
-
-```hbs title=With-horizontal-description
-{{#> check check--modifier="pf-m-horizontal"}}
-  {{#> check-input check-input--attribute='id="check8" name="exampleCheck8"'}}{{/check-input}}
-  {{#> check-label check-label--attribute='for="check8"'}}Check{{/check-label}}
-  {{#> check-description}}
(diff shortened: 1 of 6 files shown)


christiemolloy:
@mcarrano so no vertical layout for both radio and check? is this a definite?

mcarrano:
Actually, it's no horizontal layout @christiemolloy . We *do* want the vertically stacked text for both Check and Radio

christiemolloy:
Oh sorry I got those mixed up, ok yes!

src/patternfly/components/Check/examples/Check.md
```diff
@@ -45,4 +71,5 @@ If you extend this component or modify the styles of this component, then make s
 | `.pf-c-check` | `<div>`, `<label>` |  Initiates the check component. **Required**  |
 | `.pf-c-check__input` | `<input type="checkbox">` |  Initiates a check input. **Required**  |
 | `.pf-c-check__label` | `<label>`, `<span>` |  Initiates a label. **Required**  |
-| `.pf-m-disabled` | `.pf-c-check__label` |  Initiates a disabled style for labels. **Required when input is disabled** |
\ No newline at end of file
+| `.pf-c-check__description` | `<p>` |  Initiates a check description. It is displayed vertically by default. |
```
mcoker: It's probably better to leave this as a `<div>` in case users want to use an inline list or something other than phrasing content.

mcoker: Also since we only have one layout for the description, I think we can remove the "It is displayed vertically by default" text.

src/patternfly/components/Check/examples/Check.md
```diff
@@ -8,23 +8,49 @@ cssPrefix: pf-c-check
...
 {{#> check}}
   {{#> check-input check-input--attribute='id="check1" name="exampleCheck1"'}}{{/check-input}}
-  {{#> check-label check-label--attribute='for="check1"'}}Check 1{{/check-label}}
+  {{#> check-label check-label--attribute='for="check1"'}}Check{{/check-label}}
 {{/check}}
+```
+
+```hbs title=Checked
+{{#> check}}
+  {{#> check-input check-input--attribute='id="check2" name="exampleCheck2" checked'}}{{/check-input}}
```
mcoker: A more sustainable way to manage the `id`, `name` and `for` attrs would be to use `check--id="check-checked"`, `check--id="check-label-wrapping-input"`, etc, then use that var for the values of `id`, `name`, and `for`. Just a nice to have, doesn't have to be part of this PR.

christiemolloy:
Updated @mcoker

PR #2582, opened 2020-01-15
fix(table): fix accessibility for mobile

### LINKED ISSUE #2540: Tables are inaccessible on small viewports by voiceover
Through testing, I noticed our doc tables [are not readable by voiceover at 768px viewport size](https://codesandbox.io/s/2mm0c) and appear empty with 0 columns and 0 rows:

Screen Shot 2019-12-17 at 1 56 52 PM

It seems that when table was created there was discussion about using `role=grid` and it's impact on accessibility which may be causing this issue. (https://github.com/patternfly/patternfly-next/pull/749) 

It might be worth stepping back and looking at whether combining a non-table layout with the semantics of a `<table>` element is the right approach.

christiemolloy:
closes #2540

### DIFF
--- a/src/patternfly/components/Table/table-td.hbs
+++ b/src/patternfly/components/Table/table-td.hbs
@@ -6,6 +6,7 @@
     {{else if table-td--action}} class="pf-c-table__action"
     {{else if table-td--modifier}} class="{{table-td--modifier}}"
     {{/if}}
+    role="cell"
   {{#if table-td--attribute}}
     {{{table-td--attribute}}}
   {{/if}}
--- a/src/patternfly/components/Table/table-td.hbs
+++ b/src/patternfly/components/Table/table-td.hbs
@@ -6,7 +6,9 @@
     {{else if table-td--action}} class="pf-c-table__action"
     {{else if table-td--modifier}} class="{{table-td--modifier}}"
     {{/if}}
-    role="cell"
+    {{#unless table-td--IsEmpty}}
+      role="cell"
+    {{/unless}}
   {{#if table-td--attribute}}
     {{{table-td--attribute}}}
   {{/if}}
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -12,19 +12,19 @@ cssPrefix: pf-c-table
   {{/table-caption}}
   {{#> table-thead}}
     {{#> table-tr}}
-      {{#> table-th}}
+      {{#> table-th table-th--attribute='scope="col"'}}
         Repositories
       {{/table-th}}
-      {{#> table-th}}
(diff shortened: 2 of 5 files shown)


src/patternfly/components/Table/table-td.hbs
```diff
@@ -6,6 +6,7 @@
     {{else if table-td--action}} class="pf-c-table__action"
     {{else if table-td--modifier}} class="{{table-td--modifier}}"
     {{/if}}
+    role="cell"
```
mattnolting: Within `.pf-c-table__expandable-row.pf-m-expanded`, if empty cells are present, VO announces them where it should ignore them. Omitting `role="cell"` attr on them hides. 

```
{{#unless table-td--IsEmpty}}
  role="cell"
{{/unless}}
```

christiemolloy: Oh yes good point!

christiemolloy: updated

mattnolting: Screen Shot 2020-01-21 at 1 38 21 PM

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Table/table-td.hbs#L9
```
{{#unless table-td--IsEmpty}}
  role="cell"
{{/unless}}
```

then in [table.md](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Table/examples/Table.md) add `table-cell-IsEmpty=“true”`

mattnolting: Then in [table.md](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Table/examples/Table.md), anywhere you see `{{#> table-td}}{{/table-td}}` (lines 257, 408, 409, 441, 442, 602, 603, 834, 835, 1150, 1151, 1180, 1181, 1187) add `table-td--IsEmpty="true"`. 
lines: 996 & 997 (should be `table-td table-td--IsEmpty="true"` instead of `table-th`)

mcoker:
Looks great, though I see these discrepancies in some of the examples. Is that on purpose?

* compact & compact w/o borders - wrong role on empty cells
* hidden/vislble missing scope and role
* headers that wrap missing scope

src/patternfly/components/Table/table-td.hbs
christiemolloy: This was really helpful feedback thanks @mattnolting

christiemolloy:
@mcoker thanks for finding those inconsistencies. I attempted to address them by doing this:
- Matt Noltings comment addressed this problem
- Looks like this table was missing table--grid="true", added that in and added `scope="col"`
- There was no `scope="col"` set on hidden/visible, added that in

Could you confirm that these have been fixed?

PR #2586, opened 2020-01-17
build(lint): lint dist css

### LINKED ISSUE #2441: CSS Output Linting/Verification
Recently invalid CSS functions (that were actually undefined SASS functions!) slipped into our compiled component CSS.

We should add a check to lint this CSS and catch possible linting errors.

redallen:
Closes #2441 

- Remove unused handlebars files
- Lint CSS functions
- Disallow block comments in dist (cleans up `/* stylelint-enable */` block comments in our dist)


build/helpers/contains.js
```diff
@@ -1,6 +0,0 @@
-const util = require('handlebars-utils');
```
dgutride: This file is necessary for handlebars - where did the functionality get moved to?

redallen: https://github.com/patternfly/patternfly-org/blob/master/packages/patternfly-4/gatsby-theme-patternfly-org/helpers/createHandlebars.js

gulpfile.js
```diff
@@ -211,6 +201,106 @@ function buildIE() {
...
+
+  return src('./dist/**/*.css')
+    .pipe(
+      through2.obj((chunk, _, cb2) => {
+        const css = chunk.contents.toString();
+
+        let result;
+        // eslint-disable-next-line no-cond-assign
+        while ((result = regex.exec(css))) {
+          if (!validCSSFunctions.includes(result[1])) {
```
jschuler: are you ignoring case? My regex is not good enough to tell :) I think CSS functions are not case sensitive

redallen: Nice catch! I hope `node-sass`'s compiled output would never have to put this to the test, though...

PR #2630, opened 2020-01-28
fix(DropdownItem): Added pointer cursor

### LINKED ISSUE #2629: Make all DropdownItem elements pointer cursor
As seen in SCSS file, there is no cursor styling for dropdown items. Should one use different elements than button or anchor, such as a div for more advanced dropdown items, we would not get a pointer cursor.

https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Dropdown/dropdown.scss#L357-L392

SpyTec:
Closes #2629

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -364,6 +364,7 @@
   color: var(--pf-c-dropdown__menu-item--Color);
   text-align: left;
   white-space: nowrap;
+  cursor: pointer;
   background-color: var(--pf-c-dropdown__menu-item--BackgroundColor);
   border: none;
 


mcoker:
Hey @SpyTec, while this works, would you mind showing me your use case for this? Curious what the item does when you click on it, what the markup looks like, and why you can't use an `<a>` or `<button>` or element that already has the pointer styles.

Another option could be, if appropriate, to use a `role` of either `link` or `button`, and we could add global styles to https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/_globals.scss#L125-L128 to change the cursor for elements with those roles.

SpyTec:
Hey @mcoker, there's two issues `<a>` and `<button>`. With `<a>` we can't use links within dropdowns, as nesting anchor links is not allowed. An example of this usage is [here](https://www.patternfly.org/v4/documentation/react/components/table#simple-actions), although it was recently changed to use `<div>` instead of `<a>` or `<button>`.

`<button>` isn't meant to contain richer elements either, so customizing the dropdown in this way isn't ideal. Taken from W3:

> Buttons created with the BUTTON element function just like buttons created with the INPUT element, but they offer richer rendering possibilities: the BUTTON element may have content. For example, a BUTTON element that contains an image functions like and may resemble an INPUT element whose type is set to “image”, but the BUTTON element type allows content.

Which is why I'd want the ability to use elements that are not as limited in what it can contain, such as a `<div>`.

> Another option could be, if appropriate, to use a `role` of either `link` or `button`, and we could add global styles to [/src/patternfly/_globals.scss@`master`#L125-L128](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/_globals.scss#L125-L128) [](/patternfly/patternfly-next/blob/HEAD@{2020-01-30T22:48:29Z}/src/patternfly/_globals.scss#L125-L128) to change the cursor for elements with those roles.

This seems like a better idea, makes things a bit more uniform

mcoker:
@SpyTec thanks for the explanation.

> nesting anchor links is not allowed.

Why would you need to nest a link in a link? Are you placing multiple links in the item that serve different purposes? In that case, I'm not sure why the inner link is necessary - couldn't it just be the outer link with the class `-pf-c-dropdown__menu-item`?

> <button> isn't meant to contain richer elements either

Can you can give me an example of what you're placing in a dropdown item that isn't non-interactive phrasing content?

Thanks for the link to the [simple actions table demo](https://www.patternfly.org/v4/documentation/react/components/table#simple-actions). As that example is written, I see a couple of problems I would like to follow up with one of the react developers about.

> This seems like a better idea, makes things a bit more uniform

Great! For your use case, would adding a role that generates the appropriate styling work?

SpyTec:
> 
> > nesting anchor links is not allowed.
> 
> Why would you need to nest a link in a link? Are you placing multiple links in the item that serve different purposes? In that case, I'm not sure why the inner link is necessary - couldn't it just be the outer link with the class `-pf-c-dropdown__menu-item`?

For instance, I have a `<Table>` with ellipsis dropdown for each row. In that dropdown I want an Edit button that will link the user to an Edit page of that config. [Since the recommended way of doing router links in dropdowns](https://www.patternfly.org/v4/documentation/react/components/dropdown#react-router-link-usage) is to provide a `<Link>` that's what I've gone with. I somehow want to display a different view based on the click

> > isn't meant to contain richer elements either
> 
> Can you can give me an example of what you're placing in a dropdown item that isn't non-interactive phrasing content?

[Application launcher](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/ApplicationLauncher/ApplicationLauncherItem.tsx) ([source](https://www.patternfly.org/v4/documentation/react/components/applicationlauncher#with-sections-and-icons)) is a good example of what can be done with richer elements. In their implementation they use anchor links to contain the content, but then we can't nest anchor links to go somewhere else

> Thanks for the link to the [simple actions table demo](https://www.patternfly.org/v4/documentation/react/components/table#simple-actions). As that example is written, I see a couple of problems I would like to follow up with one of the react developers about.
> 
> > This seems like a better idea, makes things a bit more uniform
> 
> Great! For your use case, would adding a role that generates the appropriate styling work?

As long as React's Table implementation with `role="menuitem"` would work then my use case would be satisfied. That's my blocker at the moment

mcoker:
>  Since the recommended way of doing router links in dropdowns is to provide a <Link> that's what I've gone with. 

I reviewed this with @tlabaj and @kmcfaul today and looks like we need to make some updates to that. As you mentioned, that example results in nested `<a>` elements, which is invalid.

I'm assuming you're doing something like this instead?

```
<DropdownItem key="link" component="div">
  <Link to={'/'}>Link</Link>
</DropdownItem>
```

Nesting the clickable element in a dropdown item that is a `<div>` (or other non-clickable element) comes with a number of problems, one of which is the cursor. I've outline some more problems that surfaces, along with a potential enhancement to the dropdown item to better support using react router (or just passing your own interactive element as a child of `<DropdownItem>`) here - https://github.com/patternfly/patternfly-react/issues/3641

I'm curious if the enhancement in that issue would work for your use case?

SpyTec:
That issue looks good to me, even brings up issues I didn't realize at the time

PR #2640, opened 2020-01-29
chore(ie11): nicer ie11 error message

### LINKED ISSUE #2595: node-sass error messaging
When a css variable is called by a selector, but undefined in component root, node-sass needs to throw an error/warning message and/or problem needs to be clearly defined. 

Screen Shot 2020-01-21 at 9 48 31 AM

redallen:
Closes #2595

Also added a quick wiki page: https://github.com/patternfly/patternfly-next/wiki/IE11-Build

### DIFF
--- a/build/npm-scripts/ie-convert-all.js
+++ b/build/npm-scripts/ie-convert-all.js
@@ -26,15 +26,17 @@ module.exports = () =>
       return ie11ReadyStylesheet;
     })
     .then(transformedCss => {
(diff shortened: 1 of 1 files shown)


mcoker:
Nice, this is a lot better. We could add a little more to the message to say how to fix it. Something like

> Stylesheet [stylesheet] contains "undefined". This is usually due to a CSS variable being referenced before it is defined. Try rearranging the order of your CSS variables.

PR #2641, opened 2020-01-29
fix(nav): updated group title spacing so it changes responsively

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2572


christiemolloy:
LPTM

PR #2660, opened 2020-01-31
fix(popover): matched min/max-width so popover has consistent width

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2632

@mceledonia can you confirm that you don't consider this a visual breaking change?

### DIFF
--- a/src/patternfly/components/Popover/popover.scss
+++ b/src/patternfly/components/Popover/popover.scss
@@ -1,6 +1,6 @@
 .pf-c-popover {
   // Component
-  --pf-c-popover--MinWidth: calc(var(--pf-c-popover--c-button--sibling--PaddingRight) + #{pf-size-prem(100px)});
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Popover/popover.scss
```diff
@@ -1,6 +1,6 @@
 .pf-c-popover {
   // Component
-  --pf-c-popover--MinWidth: calc(var(--pf-c-popover--c-button--sibling--PaddingRight) + #{pf-size-prem(100px)});
+  --pf-c-popover--MinWidth: calc(var(--pf-c-popover--c-button--sibling--PaddingRight) + #{pf-size-prem(300px)});
```
matthewcarleton: are we keeping both min and max width just for flexibility? Could we just replace it with width if they are going to be the same value?

mcoker: Yeah, @mceledonia and I discussed just offering width. The react component currently allows you to pass `maxWidth` and they have a PR open to allow passing `minWidth`, too, so we would have to be sure not to break that behavior - https://github.com/patternfly/patternfly-react/pull/3601

PR #2662, opened 2020-02-01
fix(title): enabled long strings to wrap

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2661

### DIFF
--- a/src/patternfly/components/Title/title.scss
+++ b/src/patternfly/components/Title/title.scss
@@ -32,6 +32,7 @@
   --pf-c-title--m-md--FontWeight: var(--pf-global--FontWeight--semi-bold);
 
   font-family: var(--pf-c-title--FontFamily);
+  word-break: break-word;
 
   &.pf-m-4xl {
     font-size: var(--pf-c-title--m-4xl--FontSize);


mcoker:
@boaz0 can you give me an example? I tried this in both FF and chrome, doesn't overflow.

Screen Shot 2020-02-04 at 3 17 08 PM

boaz0:
Uhh... sorry my mistake. I forgot to mention that if it is inside a `flex layout` it doesn't work.

```
  <div style={{display: 'flex'}}>
  <Title headingLevel="h1" style={{overflowWrap: "break-word"}} size="4xl">
    SuperDuperlooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongTitle
  </Title>
```

mcoker:
@boaz0 ahh good find, that's due to a flex child's default `min-width` being the width of its contents. We can use `word-break: break-word;` there instead, which is similar and will allow the string to break in a flex child. Updated.

PR #2664, opened 2020-02-02
feat(Pagination): allow updating input width with CSS vars

### LINKED ISSUE #2663: Allow setting pagination input width programmatically
related to https://github.com/patternfly/patternfly-react/issues/3530

As suggested by @mcoker [here](https://github.com/patternfly/patternfly-react/issues/3530#issuecomment-576937340) the `ch` unit is going to be used to support Firefox too and make it possible to update input width through JavaScript (see example [here](https://codepen.io/mcoker/pen/MWYZrJZ))

boaz0:
closes #2663 
related to https://github.com/patternfly/patternfly-react/issues/3530#issuecomment-576937340

//cc @mcoker

### DIFF
--- a/src/patternfly/components/Pagination/pagination.scss
+++ b/src/patternfly/components/Pagination/pagination.scss
@@ -20,7 +20,9 @@
   --pf-c-pagination__nav-page-select--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--child--MarginRight: var(--pf-global--spacer--xs);
-  --pf-c-pagination__nav-page-select--c-form-control--Width: #{pf-size-prem(48px)}; // custom value to show 2 numbers
+  --pf-c-pagination__nav-page-select--c-form-control--width-base: 3.5ch;
+  --pf-c-pagination__nav-page-select--c-form-control--width-chars: 2;
+  --pf-c-pagination__nav-page-select--c-form-control--Width: calc(var(--pf-c-pagination__nav-page-select--c-form-control--width-base) + (var(--pf-c-pagination__nav-page-select--c-form-control--width-chars) * 1ch));
 
   display: flex;
   flex-wrap: wrap;


src/patternfly/components/Pagination/pagination.scss
```diff
@@ -20,7 +20,8 @@
   --pf-c-pagination__nav-page-select--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--child--MarginRight: var(--pf-global--spacer--xs);
-  --pf-c-pagination__nav-page-select--c-form-control--Width: #{pf-size-prem(48px)}; // custom value to show 2 numbers
+  --pf-c-pagination__nav-page-select--c-form-control--Base: 3ch;
+  --pf-c-pagination__nav-page-select--c-form-control--Chars: 2; // custom value to show 2 numbers
```
mcoker: to avoid a breaking change, and for the end of the var to match the property name the var applies to, can you use this instead?

```
  --pf-c-pagination__nav-page-select--c-form-control--width-base: 3ch;
  --pf-c-pagination__nav-page-select--c-form-control--width-chars: 2;
  --pf-c-pagination__nav-page-select--c-form-control--Width: calc(var(--pf-c-pagination__nav-page-select--c-form-control--width-base) + (var(--pf-c-pagination__nav-page-select--c-form-control--width-chars) * 1ch));
```

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -102,7 +103,7 @@
   }
 
   .pf-c-form-control {
-    width: var(--pf-c-pagination__nav-page-select--c-form-control--Width);
+    width: calc(var(--pf-c-pagination__nav-page-select--c-form-control--Base) + (var(--pf-c-pagination__nav-page-select--c-form-control--Chars) * 1ch));
```
mcoker: And if you go with the code I suggested above, we can just revert this to what it was before. That way if someone has defined their own `var(--pf-c-pagination__nav-page-select--c-form-control--Width)`, it will still override PF's styles.

boaz0:
@mcoker - thanks for the review. fixed that problem.

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -20,7 +20,9 @@
   --pf-c-pagination__nav-page-select--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--child--MarginRight: var(--pf-global--spacer--xs);
-  --pf-c-pagination__nav-page-select--c-form-control--Width: #{pf-size-prem(48px)}; // custom value to show 2 numbers
+  --pf-c-pagination__nav-page-select--c-form-control--width-base: 3ch;
```
mcoker: Testing this in Safari, looks like it needs 3.5ch with `00` as the value. It makes the box a small amount larger than needed in chrome and firefox, but I think that's OK - should be no harm in making the box bigger than needed, but if it's smaller then characters get cut off. 

3ch - cut off in safari
Screen Shot 2020-02-05 at 10 24 20 AM


3.5ch
Screen Shot 2020-02-05 at 10 24 07 AM

boaz0: No problem :+1:

PR #2665, opened 2020-02-03
build(workspace): create lightweight workspace

### LINKED ISSUE #2481: Speed up CSS compilation
Engineer CSS compilation to recompile changed files rather than all files when any file is changed.

### LINKED ISSUE #2612: Development server issues
I randomly get this error in the terminal from the development server

```
[2] /Users/cmichael/repos/patternfly-next/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
[2]         throw ex;
[2]         ^
[2] 
[2] TypeError: Cannot read property 'write' of null
[2]     at Zlib.callback (zlib.js:499:33)
[2] gatsby develop exited with code 7
```

Zack recommended the fix in https://github.com/gatsbyjs/gatsby/issues/14553, which increases the system's file watcher limit. I've verified that it's writing this out to `/etc/sysctl.conf` and persists after rebooting

```
marvin:patternfly-next cmichael$ grep max_user_watches /etc/sysctl.conf
fs.inotify.max_user_watches=524288
```

Also updates to `.hbs` and the component example `.md` files aren't always reflected in the browser. However simply stopping the server and restarting (and sometimes after removing the `.cache` dir) will reflect the changes.

Occasionally something really weird happens - I'll make a change and don't see it in the browser, so I make some other change, then I see the original change but not the new change. Then I make another change, and see the previous change, but not the new one. I've verified I'm giving the server plenty of time to rebuild and reflect the change in the browser. Restarting the server will fix that, too.

cc @redallen @evwilkin

redallen:
Closes #2481 Closes #2612

Adds new `npm run dev:lite` command which has an index like this:
image

With files like this:
image

And `index.html` files like this:
image

It's not pretty, and it's not meant to be. It starts up in 16 seconds and live-reloads changes to MD, HBS, or SASS in 2 seconds.

**Why:**

Gatsby is cumbersome as a workspace for our current maintainers. It's slow to start, slow to reload MD/HBS changes, and has trouble reloading `.hbs` files because it doesn't understand how dependencies work between them and MD files. We already have a task to watch our source SASS files and update them under our `dist` folder.

**What:**
This adds a new task to watch our HBS and MD files that compiles to per-example HTML files and a single `index.html` file under a new `workspace` folder.

[browser-sync](https://www.browsersync.io/) is used to live reload after CSS or HTML files are changed.


package.json
```diff
@@ -145,16 +128,8 @@
...
     ]
   },
-  "resolutions": {
-    "node-sass": "4.7.2"
-  },
   "publishConfig": {
     "registry": "https://registry.npmjs.org/",
     "tag": "latest"
-  },
-  "husky": {
```
dgutride: Why is husky removed?

redallen: By accident from the snapshot PR, adding it back now.

mcoker:
This is really slick @redallen. Compared to the current workspace, it's lightning fast to change everything I've tried:

* CSS change - 1s vs 5s
* change text in a component example - 2s vs 20s <-- that's a big one
* changes to HBS - 2s vs 25s -- not sure if this is the same as ^
* start server - 25s vs 45s

I don't really have much feedback other than we need to include `src/patternfly/demos` in the workspace dir, and add `.pf-m-redhat-font` to `<body>`. Everything else is just around styling/presentation.

On top of the speed differences, the existing server is inconsistent in reflecting changes in the browser. Often the current contents of a file isn't shown - the previous save's changes are what I'll see in the browser. And sometimes changes aren't reflected at all after saving multiple times and waiting a minute or so to see them - restarting the server will make the changes visible.

This is a small thing, but the new server seems to preserve my selected element in dev tools when the page hot refreshes with markup changes - something the existing server doesn't do.

One thing I would like to test more is how it handles adding/removing files and updating the workspace after switching branches, something the existing server often struggles with.

As it stands, compared to the existing workspace, this will save me quite a bit of time and frustration particularly when working on anything that doesn't just involve CSS changes only. And since it runs on a different port, I can switch between the two workspaces, depending on what I want to look at.

christiemolloy:
Layout styles aren't being included:

Screen Shot 2020-02-04 at 11 06 51 AM
Screen Shot 2020-02-04 at 11 07 33 AM


Some overlap is occurring in components
Screen Shot 2020-02-04 at 11 09 16 AM

redallen:
That's correct @christiemolloy ! I'm fundamentally against workspace example styles because it causes confusion confusion for consumers when they don't get them and there's no documentation for them. These docs show the same thing that a CodeSandbox would show, for example.

Even still, I _could_ add some logic to include them in the `<head>`. Is this important @mcoker and @mattnolting ?

dgutride:
This workspace isn't for consumers, it must only be for developers if we are going to allow it in.  We need to prioritize the developer experience no matter what and the developers had originally requested the outlines for layouts.

christiemolloy:
yeah to @dgutride point, @redallen the layout styles enhance my developer experience and without them I find it hard to follow a lot of the examples, but that might just be me

redallen:
That's enough convincing for me, I'll add in some frontmatter parsing and glob for example CSS files to include in `<head>`. CSS file compatibility between `dev:lite` and Gatsby shouldn't be a problem so long as both create the example wrapper class names the same.

redallen:
Layout styles and demos are in.

mattnolting:
This is awesome!!! Two requests:
- remove the `<hr>`s between the examples
- add an empty `dev-lite.css` file that doesn't need to be compiled, watched, etc...

I'd like to add some very basic styling in `dev-lite.css` that will support this awesome development workspace!!

PR #2668, opened 2020-02-04
feat(table): allow sortable table headers to wrap

### LINKED ISSUE #2584: tables: no way to let sortable column headers wrap
This issue was first created in React: https://github.com/patternfly/patternfly-react/issues/3520 however it is an issue we need to fix in Core first.

Right now, `pf-m-wrap` applied to a sortable table's `<th>` doesn't work because the modifier only applies to a direct descender of `<th>` and in the sortable table there is a button in the `<th>`.

We need to consult design because of the arrow icon to the right of the `<th>` . When `white-space: normal` is applied, the arrow icon drops to the next line:

Screen Shot 2020-01-16 at 11 17 18 AM

So we may need a different solution here like truncating for example. @mcarrano 

@rachael-phillips making you aware too.

christiemolloy:
fixes #2584 

Got confirmation with @gdoyle1 that instead of truncating table headers with sortable, we can wrap the text.

Screen Shot 2020-02-04 at 9 58 53 AM


src/patternfly/components/Table/table-th.hbs
```diff
@@ -25,8 +25,14 @@
     {{{table-th--attribute}}}
   {{/if}}>
   {{#if table-th--sortable}}
-    {{#> button button--modifier="pf-m-plain"}}
-      {{> @partial-block}}
+    <button class="pf-c-button pf-m-pain{{#if table-th--sortable-wrap}} pf-m-sort-wrap{{/if}}">
+      {{#if table-th--sortable-wrap}}
```
christiemolloy: @mcoker @mattnolting what do you think about wrapping the text in a span only if `pf-m-sort-wrap` is applied? This was mainly so that we could target the span to apply `white-space: normal` so that only the text wraps. Or we could have a span wrap all text that is inside sortable `<th>`

src/patternfly/components/Table/table-th.hbs
```diff
@@ -25,8 +25,14 @@
     {{{table-th--attribute}}}
   {{/if}}>
   {{#if table-th--sortable}}
-    {{#> button button--modifier="pf-m-plain"}}
-      {{> @partial-block}}
+    <button class="pf-c-button pf-m-pain{{#if table-th--sortable-wrap}} pf-m-sort-wrap{{/if}}">
```
mcoker:
```suggestion
    <button class="pf-c-button pf-m-plain{{#if table-th--sortable-wrap}} pf-m-sort-wrap{{/if}}">
```

christiemolloy:
@mcarrano thanks for finding that issue. i have provided a solution to fix it let me know if you're still seeing it.

src/patternfly/components/Table/table.scss
```diff
@@ -390,10 +391,20 @@
...
     }
+
+    &.pf-m-sort-wrap {
+      display: flex;
+
+      .pf-c-table__sort--c-button-text {
+        white-space: normal;
+      }
+    }
   }
```
mattnolting: I think you can achieve the same result without having to create a new selector. Does this work for you?

```
// Table sort
.pf-c-table__sort {
  // override pf-m-plain color
  .pf-c-button {
    // stylelint-disable
    &.pf-m-plain {
      color: var(--pf-c-table__sort--c-button--Color);
    }
  }

  &.pf-m-wrap .pf-c-button {
    white-space: normal;
    display: flex;
  }
  // stylelint-enable

  &.pf-m-selected {
    // set button and indicator color
    .pf-c-button,
    .pf-c-table__sort-indicator {
      color: var(--pf-c-table__sort--sorted--Color);
    }
  }

  &:not(.pf-m-selected) {
    &:hover {
      .pf-c-table__sort-indicator {
        color: var(--pf-c-table__sort-indicator--hover--Color);
      }
    }
  }
}
```

src/patternfly/components/Table/examples/Table.md
```diff
@@ -218,6 +218,51 @@ cssPrefix: pf-c-table
   {{/table-tbody}}
 {{/table}}
 ```
+
+```hbs title=Sortable-with-headers-that-wrap
+{{#> table table--id="table-sortable-wrap" table--grid="true" table--modifier="pf-m-grid-lg" table--attribute='aria-label="This is a sortable table example with headers that wrap"'}}
+  {{#> table-thead}}
+    {{#> table-tr}}
+      {{#> table-th table-th--attribute='scope="col"' table-th--sortable="true" table-th--sortable-wrap="true" table-th--selected="true" table-th--asc="true"}}
```
mattnolting: Then you could just add `table-th--modifier="pf-m-wrap"` here instead of needing updated structure.

package.json
```diff
@@ -71,7 +71,7 @@
     "eslint-plugin-react": "^7.7.0",
     "eslint-plugin-standard": "^3.0.1",
     "fs-extra": "^5.0.0",
-    "gatsby": "2.18.6",
+    "gatsby": "^2.19.12",
```
redallen: Change to `    "gatsby": "2.19.12",` and rerun `npm i` so we don't end up on a broken Gatsby version like has happened in the past :)

src/patternfly/components/Table/table-th.hbs
```diff
@@ -2,7 +2,7 @@
     {{else if table-th--check}} class="pf-c-table__check {{table-th--modifier}}"
     {{else if table-th--action}} class="pf-c-table__action {{table-th--modifier}}"
     {{else if table-th--icon}} class="pf-c-table__icon {{table-th--modifier}}"
-    {{else if table-th--sortable}} class="pf-c-table__sort {{#if table-th--selected}} pf-m-selected{{/if}} {{table-th--modifier}}"
+    {{else if table-th--sortable}} class="pf-c-table__sort {{#if table-th--sortable-wrap}} pf-m-wrap{{/if}}{{#if table-th--selected}} pf-m-selected{{/if}} {{table-th--modifier}}"
```
mattnolting: 👍

src/patternfly/components/Table/table.scss
```diff
@@ -390,10 +391,22 @@
...
+    // stylelint-disable
     &.pf-m-plain {
       color: var(--pf-c-table__sort--c-button--Color);
     }
   }
+  // stylelint-enable
+
+  &.pf-m-wrap {
+    .pf-c-button {
+      display: flex;
```
mattnolting: I'd put `whitespace` here as it doesn't require `.pf-c-table__sort--c-button-text`.

christiemolloy: @mcoker mentioned in slack: "then we could also be more specific about what we style with white-space - not that it hurts to style all of pf-c-button, but we may not want that at some point"

src/patternfly/components/Table/table-th.hbs
```diff
@@ -25,8 +25,14 @@
     {{{table-th--attribute}}}
   {{/if}}>
   {{#if table-th--sortable}}
-    {{#> button button--modifier="pf-m-plain"}}
-      {{> @partial-block}}
+    <button class="pf-c-button pf-m-plain">
```
mcoker: can we go back to `{{#> button}}` for this?

package.json
```diff
@@ -65,13 +65,14 @@
     "eslint-config-standard-jsx": "^5.0.0",
     "eslint-config-standard-react": "^6.0.0",
     "eslint-import-resolver-alias": "^1.0.3",
+    "eslint-plugin-jsx-a11y": "^6.2.3",
```
christiemolloy: @redallen it asked me to update this, is this ok?

src/patternfly/components/Table/examples/Table.md
mcoker: +1, we use `table-th--modifier="pf-m-wrap"` on the other includes, probably good to be consistent instead of using `table-th--sortable`

src/patternfly/components/Table/table-th.hbs
```diff
@@ -25,8 +25,14 @@
...
+      {{#if table-th--sortable-wrap}}
+        <span class="pf-c-table__sort--c-button-text">
```
mcoker:
```suggestion
        <span class="pf-c-table__sort-text">
```

src/patternfly/components/Table/table.scss
```diff
@@ -390,10 +391,22 @@
...
+    }
+
+    .pf-c-table__sort--c-button-text {
```
mcoker:
```suggestion
    .pf-c-table__sort-text {
```

src/patternfly/components/Table/table.scss
```diff
@@ -390,10 +391,22 @@
...
+      white-space: normal;
```
mcoker: did you want to move this out and apply it to `.pf-c-table__sort.pf-m-wrap .pf-c-button` above? That will fix the bug for users now, rather than relying on them updating their structure to get this.

src/patternfly/components/Table/table.scss
christiemolloy: removed this anyway

src/patternfly/components/Table/table.scss
christiemolloy: oh I have it moved out this might not have the latest changes

PR #2675, opened 2020-02-04
feat(dev-lite): added basic WS styling

mattnolting:


mcoker:
What do you think about adding...

```
:not(.ws-lite-example) table code {
    background: var(--pf-global--BackgroundColor--300);
}

:not(.ws-lite-example) code {
    background: var(--pf-global--BackgroundColor--100);
    padding: 0 .5em;
}

:not(.ws-lite-example) h1, :not(.ws-lite-example) h2, :not(.ws-lite-example) h3, :not(.ws-lite-example) h4, :not(.ws-lite-example) h5, :not(.ws-lite-example) h6 {
    font-family: var(--pf-global--FontFamily--heading--sans-serif);
}
```

PR #2680, opened 2020-02-05
chore(nomenclature): phrase non production components as beta

seanforyou23:
This PR renames "experimental" things to "beta" so that the intent of the in-progress status is more clearly signaled to users.

Related to https://github.com/patternfly/patternfly-react/issues/3624


README.md
```diff
@@ -65,18 +65,9 @@ If you want to set host to a specific IP address for example `172.17.12.1` run `
...
-
-To make a component experimental - modify the `experimental-features.js` file in the root and add an entry that looks like the following:
-
-```javascript
-{
-  name: 'Alert',
-  path: 'components/Alert/'
-},
-```
+When creating a brand new component, it should be released as beta in order to get feedback.  This results in the css not being bundled into the full distribution (`patternfly.css`), but the component level css is still included in the npm package for usage.
```
redallen: Now it _is_ being bundled.

PR #2681, opened 2020-02-06
feat(file upload): add file upload component

### LINKED ISSUE #2624: Introduce File Upload component
Introduce File Upload component according to the UI design here: https://marvelapp.com/de0ggah

This component will be introduced for uploading text files only. The user can upload a file by browsing to it, pasting its contents, or dragging and dropping it.

christiemolloy:
closes #2624


src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,147 @@
...
+title: File upload
+section: experimental
+cssPrefix: pf-c-file-upload
+---
+
+## Examples
+```hbs title=Basic-file-upload
+{{#> file-upload file-upload--type="form"}}
+  {{#> input-group}}
+    {{#> form-control controlType="input" input="true" form-control--attribute='id="basic-file-upload" name="basic-file-upload" placeholder="Drag a file here or browse to upload" aria-label="upload a file"'}}
```
mcoker:
```suggestion
    {{#> form-control controlType="input" input="true" form-control--attribute='id="basic-file-upload" name="basic-file-upload" placeholder="Drag a file here or browse to upload" aria-label="Upload a file"'}}
```

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+  --pf-c-file-upload--c-button--m-control--disabled--after--BorderRightColor: var(--pf-global--BorderColor--300);
+  --pf-c-file-upload--c-button--m-control--disabled--after--BorderBottomColor: var(--pf-global--BorderColor--200);
+  --pf-c-file-upload--c-button--m-control--disabled--after--BorderLeftColor: var(--pf-global--BorderColor--300);
+  --pf-c-file-upload--c-button--m-control--disabled--after--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-file-upload--c-button--m-control--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
+
+  display: flex;
+  flex-direction: column;
+
+  &.pf-m-draganddrop {
```
mcoker: You might check with react about what states the react component will support, and we write classes that support that. It seems like this might be `pf-m-drag-hover`, and there may be another class for `pf-m-drag-drop` (once a file has dropped) that we might want to style for, things like that.

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+    &::after {
+      content: "";
+      border: var(--pf-c-file-upload--c-button--m-control--disabled--after--BorderWidth) solid;
+      border-color: var(--pf-c-file-upload--c-button--m-control--disabled--after--BorderTopColor) var(--pf-c-file-upload--c-button--m-control--disabled--after--BorderRightColor) var(--pf-c-file-upload--c-button--m-control--disabled--after--BorderBottomColor) var(--pf-c-file-upload--c-button--m-control--disabled--after--BorderLeftColor);
+      border-radius: initial;
+    }
+  }
+}
+
+.pf-c-spinner {
```
mcoker: The way the styles are written for these components (input group, button, spinner) will apply globally to any use of them in patternfly.

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+
+  &.pf-m-loading {
+    background-color: var(--pf-global--disabled-color--300);
+    border: 0;
+  }
+}
+
+.pf-c-button.pf-m-control {
+  --pf-c-button--PaddingRight: var(--pf-c-file-upload--c-button--m-control--PaddingRight);
+  --pf-c-button--PaddingLeft: var(--pf-c-file-upload--c-button--m-control--PaddingLeft);
```
mcoker: are these styles necessary? Would be good to use the defaults if we can.

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+.pf-c-input-group > .pf-c-form-control {
+  &::placeholder {
+    color: var(--pf-c-file-upload--c-input-group--c-form-control--placeholder--Color);
+  }
+}
+
+.pf-c-file-upload__text-area {
```
mcoker: I would make this a `div` that you place a `textarea.pf-c-form-control` inside of

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+      border-bottom-width: var(--pf-c-file-upload__text-area--focus--BorderBottomWidth);
+    }
+  }
+
+  &:read-only,
+  &.pf-m-readonly {
+    --pf-c-file-upload__text-area--BackgroundColor: var(--pf-c-file-upload__text-area--readonly--BackgroundColor);
+  }
+
+  &.pf-m-loading {
```
mcoker: instead of changing this element from a textarea to a div and sharing the same class, if you just make it a div and nest a textarea.pf-c-form-control in it, you can set the textarea's opacity to 0 when it's in the loading state, the absolutely position the spinner in this div as you have in your example.

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,147 @@
...
+  {{#> file-upload-text-area file-upload-text-area--attribute='aria-label="text-area-empty"'}}
+  {{/file-upload-text-area}}
+{{/file-upload}}
+```
+
+```hbs title=Browse-to-upload-file
+{{#> file-upload file-upload--type="form"}}
+  {{#> input-group}}
+    {{#> form-control controlType="input" input="true" form-control--attribute='id="basic-file-upload-browse" name="basic-file-upload-browse" placeholder="Drag a file here or browse to upload" aria-label="select browse button to upload file"'}}
+    {{/form-control}}
```
mcoker: What about a wrapping element for the input group, one for the text area, and one for the message? Like `pf-c-form-control__header` or something like that, `__main/content` for the textarea container, and `__footer` for the message.

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+      border-radius: initial;
+    }
+  }
+}
+
+.pf-c-spinner {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
mcoker: this isn't working because `.pf-c-spinner` also has a default `transform` in it's `animation` that makes it spin. In general I'd try and avoid positioning components like this directly in another component - let's use a wrapper instead to do the positioning, then just drop the component in that. So like `pf-c-form-control__content-spinner/loading` where "content" is whatever you name the main area container. That does the positioning, then you just drop the spinner in it.

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+  --pf-c-button--PaddingRight: var(--pf-c-file-upload--c-button--m-control--PaddingRight);
+  --pf-c-button--PaddingLeft: var(--pf-c-file-upload--c-button--m-control--PaddingLeft);
+
+  outline-offset: var(--pf-c-file-upload--c-button--m-control--OutlineOffset);
+
+  &:disabled,
+  &.pf-m-disabled {
+    color: initial;
+    background-color: var(--pf-c-file-upload--c-button--m-control--disabled--BackgroundColor);
+
```
mcoker: These disabled styles, too - can we just use what's in the button component?

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,132 @@
...
+      content: "";
+      background-color: var(--pf-c-file-upload--m-draganddrop--after--BackgroundColor);
+      opacity: var(--pf-c-file-upload--m-draganddrop--after--Opacity);
+    }
+  }
+}
+
+.pf-c-input-group > .pf-c-form-control {
+  &::placeholder {
+    color: var(--pf-c-file-upload--c-input-group--c-form-control--placeholder--Color);
```
mcoker: is this something we could add back to the form control component?

src/patternfly/components/FileUpload/file-upload.scss
mcoker: A trick you can use to create a crosshair element on something is absolutely position the `::before/after` elements as lines that create the crosshair - https://codepen.io/mcoker/pen/yLNygrE

Screen Shot 2020-02-07 at 7 07 32 PM

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: Good point, will do!

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: @lboehling in the designs the left and right padding around the button was smaller, are you okay with keeping the default padding style for the button?

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: the default disabled style for `pf-c-button :disabled` is below. Its very different from what @lboehling has for the disabled state. @mcoker do you mean use the variables from the top-level button disabled button component and apply them here?

Screen Shot 2020-02-10 at 9 31 18 AM

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: @lboehling @mceledonia what do you think about updating the form control component to have grey text color for its placeholder text? Currently its black and in the file upload component Lucia designed it be grey.

src/patternfly/components/FileUpload/examples/FileUpload.md
christiemolloy: I think wrapping them is a good idea, not too sure about the naming though. I think that the file upload component is always going to have an input group at the top and a text area underneath, could you think of a situation where this wouldn't be the case?

@mattnolting what are your thoughts here?

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: ahh thank you!!

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: Talked with Michael Spaxman and these are some of the events that were bought up during the draggable table columns POC:
      draggable,
      onDragStart,
      onDragEnd,
      onDrop,
      onDragOver,
      onDragEnter,
      onDragLeave

Talked with Lucia and her designs currently cover the events above

mturley: I think as far as CSS goes for this, since the element being dragged does not originate from the component, you only have to worry about the classes added/removed on the events `onDragEnter`, `onDragLeave`, and `onDrop`. I'm not sure how `onDragOver` is different from `onDragEnter` though.

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,119 @@
...
+  --pf-c-file-upload--c-button--m-control--disabled--after--BorderLeftColor: var(--pf-global--BorderColor--300);
+  --pf-c-file-upload--c-button--m-control--disabled--after--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-file-upload--c-button--m-control--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
+
+  // Message
+  --pf-c-file-upload__message--MarginTop: var(--pf-global--spacer--xs);
+  --pf-c-file-upload__message--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-file-upload__message--Color: var(--pf-global--Color--100);
+  --pf-c-file-upload__message--m-error--Color: var(--pf-global--danger-color--100);
+
```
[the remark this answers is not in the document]
christiemolloy: That would be a nice update, do you know if there was any reason why `pf-c-form__helper-text` was still inside of `.pf-c-form`?

src/patternfly/components/FileUpload/file-upload.scss
christiemolloy: Yeah i would agree with that. There is no styling difference between `onDragOver` and `onDragEnter` do you have any preference with naming? @mattnolting had suggested we use `pf-m-drag-hover` to cover both of those events. Wdyt @mturley

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -0,0 +1,119 @@
...
+  display: flex;
+  flex-direction: column;
+
+  .pf-c-file-upload__input-group + .pf-c-file-upload__text-area > .pf-c-form-control {
+    border-top: 0;
+  }
+
+  &.pf-m-drag-over {
```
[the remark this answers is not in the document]
christiemolloy: Yeah i think that aligns more with how we typically name modifiers, nice!

src/patternfly/components/FileUpload/file-upload.scss
mturley: I think that works well, and matches the naming of similar CSS `:hover` styles. 👍

src/patternfly/components/FileUpload/file-upload.scss
lboehling: I'm fine with the default padding!  @christiemolloy

src/patternfly/components/FileUpload/file-upload.scss
lboehling: Good question. I was using the sketch kit and it seems like the disabled states are treated with a gray background in split buttons/dropdowns. Is the gray bkg just pulling from the disabled form/dropdown that the action button is on? If this is just an inconsistency with the library and/or if I'm just interpreting the sketch library incorrectly, I'm fine with keeping the default disabled styling!  

what do you think? @mcarrano 
Screen Shot 2020-02-11 at 3 57 45 PM

mcarrano:
@lboehling I just spoke with @christiemolloy and we agreed that the styling in your design is preferable to the disabled button styling for the control button so she will make that update after confirming that this does not present a risk elsewhere.

src/patternfly/components/FileUpload/file-upload.scss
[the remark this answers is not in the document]
mattnolting: I think we simply didn't plan for it. `.pf-c-form` was one of the first components.

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+        }}
+      {{#> button button--modifier="pf-m-control"}}
+        Browse...
+      {{/button}}
+      {{#> button button--modifier="pf-m-control" button--attribute="disabled"}}
+        Clear
+      {{/button}}
+    {{/input-group}}
+  {{/file-upload-input-group}}
+  {{#> file-upload-textarea file-upload-textarea--aria-label="Text area empty"}}
```
srambach: "Text area empty" is really awkward to me; would "Text area is empty" or "Empty text area" be better?

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+        }}
+      {{#> button button--modifier="pf-m-control"}}
+        Browse...
+      {{/button}}
+      {{#> button button--modifier="pf-m-control"}}
+        Clear
+      {{/button}}
+    {{/input-group}}
+  {{/file-upload-input-group}}
+  {{#> file-upload-textarea file-upload-textarea--aria-label="Text area empty" file-upload-textarea--attribute='readonly'}}Ssh-Rsa AAh3zJFkzjjakCJialksjfB3zJFkzzAAhhMskjjakCJialksjfB3z89z3zJFkz3 +kzMAjsauoox88aaZXphBx4fczJFkzMAjsauoox88aaZXphBx4fczJFkzMAjsauoox88aaZXphBx4fc
```
srambach: Now the text area is not empty - should the label change?

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+cssPrefix: pf-c-file-upload
+---
+
+## Examples
+
+```hbs title=Basic-file-upload
+{{#> file-upload file-upload--id="basic-file-upload"}}
+  {{#> file-upload-input-group}}
+    {{#> input-group}}
+      {{> file-upload-text-input
```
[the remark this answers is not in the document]
christiemolloy: Yeah I think so! Good point!

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+        file-upload-text-input--aria-label="Upload a file"
```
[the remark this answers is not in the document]
christiemolloy: good suggestion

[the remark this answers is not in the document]
christiemolloy: will implement :)

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
christiemolloy: This was discussed with design, the user has the ability to change the name of the file. But im sure that with the React implementation we will be able to analyze if this is something that we want.

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+
+```hbs title=File-upload-loading
+{{#> file-upload file-upload--id="file-upload-loading" file-upload--modifier="pf-m-loading"}}
+  {{#> file-upload-input-group}}
+    {{#> input-group}}
+      {{> file-upload-text-input
+        file-upload-text-input--aria-label="Read only"
+        file-upload-text-input--attribute='readonly name="file-upload-loading" value="Sample.png"'
+      }}
+      {{#> button button--modifier="pf-m-control"}}
```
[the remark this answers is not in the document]
christiemolloy: Yep treating it so that when the file is loading, you can still browse for a file which will override the current file, or you can cancel.

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+### Overview
+
+### Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-file-upload` | `<form>`, `<div>` | Initiates the file upload component. **Required**. |
+| `.pf-c-file-upload__input-group` | `<div>` | Initiates the file upload input group. **Required** |
+| `.pf-c-file-upload__text-area` | `<div>` | Initiates the text area element. **Required** |
+| `.pf-c-file-upload__message` | `<div>` | Initiates the file upload message. |
```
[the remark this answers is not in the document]
christiemolloy: Right now there is only designs for the error state. However I could see someone having helper text underneath, thoughts?

src/patternfly/components/FileUpload/file-upload-text-area-spinner.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-file-upload__text-area-spinner{{#if file-upload-text-area-spinner--modifier}} {{file-upload-text-area-spinner--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
christiemolloy: good catch thanks!

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
mcarrano: Good question @matthewcarleton . I may be misremembering, but didn't we say that the user should not be able to enter a file name directly into this field as we wanted the browser to handle any access to the user's file system @christiemolloy . Can you confirm @mturley ? If that's the case, maybe this should be made read-only at all times.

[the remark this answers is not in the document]
matthewcarleton: ya I agree with you @mcarrano. It seems unlikely that they'd need to edit the name if they put the wrong file in there. I would guess (maybe there's research out there to back this up) most users don't expect to be able to edit a filename for this type of interaction.

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
mcarrano: I think this would be handled by the React implementation but I would not expect the user to be able to click Browse while an upload is in progress. I just seems like it over complicates the React code. I'd say to disable Browse in this state. Do you agree @mturley ?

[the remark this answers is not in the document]
matthewcarleton: ok, maybe for the react implementation we should throw a warning that if they browse they will cancel their current upload. It could be confused that you can upload more than one file rather than replace the current upload.

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
mcarrano: I don't feel strongly about this. My vote would not be to add another example.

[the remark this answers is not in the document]
matthewcarleton: Ya I think it would be good. Is there any reason for it otherwise? I assumed it was for the success state but maybe I'm wrong.

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
mturley: Yes @mcarrano you're correct, that's what we discussed. In the implementation, it won't make sense to allow the user to edit a filename here, because it's not clear what effect that would have.

If the user either presses the Browse button and selects a file, or drags and drops a file into the component, then the read-only filename is populated from the selected file. If it were possible to change, it would have no effect, the file was already loaded.

I can only think of two things typing a filename in that box would do in theory: (a) the user manually specifies a path to a file on disk instead of browsing for it, or (b) the user is entering the name of a file to be created somewhere in the consuming app as a result of this form submission. (a) is not possible due to security restrictions (we need the browser to provide the file path and contents based on user action, JavaScript can't just ask the browser for the contents of an arbitrary file path), and (b) did not seem to be a need here.

In the future if we want to support a use case like (b) we could add a prop to make that field editable as a non-breaking change.

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
mturley: I don't have a strong opinion on this, but it makes sense to me to block the Browse button during an upload since we have the Clear button which could function as a cancel action.

While the React component should support the asynchronous upload case for those who need it, I expect the majority use case (directly loading a small text file from the user's local machine into a form) to effectively happen instantly, and the in-progress view would never appear. In that case, the actual transfer of the file contents to a server would happen when the form is submitted, not when the file is selected.

In fact, I think that might be reason enough to call this component "File Input" rather than "File Upload".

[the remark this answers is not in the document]
christiemolloy: Based on everyone's thoughts I think it would be best to disable the browse button while the upload is happening.

@mcarrano what are your thoughts on @mturley name suggestion?

src/patternfly/components/FileUpload/examples/FileUpload.md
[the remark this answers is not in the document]
christiemolloy: Yeah we dont have a success state as of yet but I could see that coming in the future. For now I wont add a new example.

src/patternfly/components/FileUpload/examples/FileUpload.md
```diff
@@ -0,0 +1,150 @@
...
+  {{#> file-upload-textarea file-upload-textarea--aria-label="Text area"}}Ssh-Rsa AAh3zJFkzjjakCJialksjfB3zJFkzzAAhhMskjjakCJialksjfB3z89z3zJFkz3 +kzMAjsauoox88aaZXphBx4fczJFkzMAjsauoox88aaZXphBx4fczJFkzMAjsauoox88aaZXphBx4fc
+  {{/file-upload-textarea}}
+{{/file-upload}}
+```
+
+```hbs title=Drag-file-hover-component
+{{#> file-upload file-upload--id="drag-file-hover-component" file-upload--modifier="pf-m-drag-hover"}}
+  {{#> file-upload-input-group}}
+    {{#> input-group}}
+      {{> file-upload-text-input file-upload-text-input--aria-label="Drag a file here or browse to upload" file-upload-text-input--attribute='readonly placeholder="Drag a file here or browse to upload" aria-describedby="inputButton4"'}}
```
mattnolting:
```suggestion
      {{> file-upload-text-input 
      file-upload-text-input--aria-label="Drag a file here or browse to upload" 
      file-upload-text-input--attribute=(concat 'readonly placeholder="Drag a file here or browse to upload" aria-describedby="' file-upload--id '-browse"')}}
```

mturley:
Awesome work @christiemolloy ! This came out great.

mturley:
@christiemolloy @mcoker @mcarrano @mattnolting I realized something about this design while implementing the React version, and I have a question:

There is [an example](https://www.patternfly.org/v4/documentation/core/components/fileupload#file-upload-error) showing an error icon in the filename field (`aria-invalid="true"`), which looks like it is for errors in the file upload process itself. Two things:

1. The filename input field in this example is not `readonly`, and making it `readonly` hides that error icon even if the field has `aria-invalid="true"`.
2. If a consumer wants to display a form field validation message about the file's body itself (e.g. what they pasted in was invalid in some way, or too long, or is required but was left empty, etc) should we put the invalid styles on the `textarea`? I can imagine a scenario in which the field is required, and there is an error in the upload, so the "required" validation is failing and we have error icons in both the filename input field and the textarea.

I think we either need to change the CSS so the error icon can be displayed in a readonly text field, or we should change the examples and place the error icon in the textarea instead of the input for any invalid states of this component. What do you guys think?

If we did the latter, instead of this:
Screenshot 2020-02-26 15 06 22

we'd have this:
Screenshot 2020-02-26 15 06 04

christiemolloy:
Thanks for bringing this up @mturley we definitely didn't catch this in the Core implementation. I would be in favor of the second example where the error icon is on the text area because that is the "editable" section, so it makes sense to tell the user where the error is and where it can be fixed. I think putting the error icon within a disabled input would be misleading as to where the error is and how to fix it. Thoughts @mcarrano @mcoker @mattnolting

christiemolloy:
Thanks for bringing this up @mturley we definitely didn't catch this in the Core implementation. I would be in favor of the second example where the error icon is on the text area because that is the "editable" section, so it makes sense to tell the user where the error is and where it can be fixed. I think putting the error icon within a disabled input would be misleading as to where the error is and how to fix it. Thoughts @mcarrano @mcoker @mattnolting

christiemolloy:
Thanks for bringing this up @mturley we definitely didn't catch this in the Core implementation. I would be in favor of the second example where the error icon is on the text area because that is the "editable" section, so it makes sense to tell the user where the error is and where it can be fixed. I think putting the error icon within a disabled input would be misleading as to where the error is and how to fix it. Thoughts @mcarrano @mcoker @mattnolting

mcoker:
> The filename input field in this example is not readonly, and making it readonly hides that error icon even if the field has aria-invalid="true".

From what I can tell, if a field is `readonly` and is either in the invalid or success state, the invalid/success state overrides the `readonly` styles. So you keep the error/success icon, but lose the readonly state background. 

Screen Shot 2020-02-27 at 10 23 45 AM

Screen Shot 2020-02-27 at 10 21 18 AM

I'm pretty sure we can consider that a bug and fix it. And in testing the fix, it looks like allowing a form control component to support error/success and `readonly` states at the same time will result in a breaking change because of the way we assign `background` properties. https://github.com/patternfly/patternfly-next/pull/2753/files

Assuming we want to implement this fix, we can also patch the file upload component now, which wouldn't be a breaking change, and update the form control component later. Also curious what @mcarrano and @lboehling think about that or if maybe there is another design we should consider for indicating error(s) in this component?

> I can imagine a scenario in which the field is required, and there is an error in the upload, so the "required" validation is failing and we have error icons in both the filename input field and the textarea.

Agree, and am curious what @mcarrano and @lboehling think.

mcarrano:
@christiemolloy @mcoker I think that associating the error with the text area makes the most sense as shown in the second example above.

mturley:
In that case, in the[ "Upload complete non editable" state](https://www.patternfly.org/v4/documentation/core/components/fileupload#upload-complete-non-editable), do we still want to support the error/success icon being added to that read-only textarea? Seems we would still need a patch/fix there.

PR #2684, opened 2020-02-06
chore(components): updated _all.scss with missing components

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2683

```
cur=`pwd`
base="src/patternfly"

for dir in layouts components; do
path="$cur/$base/$dir"
file="_all.scss"

cd $path
find . -name "*.scss" | grep -v $file | sed -e 's/^/@import "/g' | sed -e 's/$/";/g' | sort > $file

done
```


redallen:
I like your bash script!

PR #2693, opened 2020-02-10
feat(select-checkbox): added filter and search input

### LINKED ISSUE #2565: Add filtering to checkbox select menu
Want the ability to filter items in a checkbox select component. This is useful when the list of potential options is very large.

Here is the target design:

Checkbox (without badges)

This takes the existing checkbox select and adds a type-ahead filter and an optional Clear button. The clear button, when present, will clear any existing selections. If nothing selected, the clear button should be hidden.

mattnolting:
closes #2565

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -8,6 +8,11 @@ $pf-c-form-control__select--ViewBox: "320" !default;
 $pf-c-form-control--success--Color: $pf-global--success-color--100 !default;
 $pf-c-form-control--success--Coordinates: "M504 256c0 136.967-111.033 248-248 248S8 392.967 8 256 119.033 8 256 8s248 111.033 248 248zM227.314 387.314l184-184c6.248-6.248 6.248-16.379 0-22.627l-22.627-22.627c-6.248-6.249-16.379-6.249-22.628 0L216 308.118l-70.059-70.059c-6.248-6.248-16.379-6.248-22.628 0l-22.627 22.627c-6.248 6.248-6.248 16.379 0 22.627l104 104c6.249 6.249 16.379 6.249 22.628.001z" !default;
 
+// Input - search
+$pf-c-form-control--m-search--Color: $pf-global--icon--Color--light !default;
+$pf-c-form-control--m-search--Coordinates: "M505 442.7L405.3 343c-4.5-4.5-10.6-7-17-7H372c27.6-35.3 44-79.7 44-128C416 93.1 322.9 0 208 0S0 93.1 0 208s93.1 208 208 208c48.3 0 92.7-16.4 128-44v16.3c0 6.4 2.5 12.5 7 17l99.7 99.7c9.4 9.4 24.6 9.4 33.9 0l28.3-28.3c9.4-9.4 9.4-24.6.1-34zM208 336c-70.7 0-128-57.2-128-128 0-70.7 57.2-128 128-128 70.7 0 128 57.2 128 128 0 70.7-57.2 128-128 128z" !default;
+
+
 .pf-c-form-control {
   // Component
   --pf-c-form-control--FontSize: var(--pf-global--FontSize--md);
@@ -71,6 +76,13 @@ $pf-c-form-control--success--Coordinates: "M504 256c0 136.967-111.033 248-248 24
   --pf-c-form-control--success--BackgroundUrl: #{pf-bg-svg($pf-c-form-control--success--Coordinates, $svg-color: $pf-c-form-control--success--Color)};
   --pf-c-form-control--success--check--Background: var(--pf-c-form-control--success--BackgroundUrl) var(--pf-c-form-control--success--BackgroundPosition) / var(--pf-c-form-control--success--BackgroundSize) no-repeat;
   --pf-c-form-control--success--Background: var(--pf-c-form-control--BackgroundColor) var(--pf-c-form-control--success--check--Background);
+
+  // Input m-search
+  --pf-c-form-control--m-search--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-form-control--m-search--BackgroundPosition: var(--pf-c-form-control--PaddingLeft);
+  --pf-c-form-control--m-search--BackgroundSize: var(--pf-c-form-control--FontSize) var(--pf-c-form-control--FontSize);
+  --pf-c-form-control--m-search--BackgroundUrl: #{pf-bg-svg($pf-c-form-control--m-search--Coordinates, $svg-color: $pf-c-form-control--m-search--Color)};
+  --pf-c-form-control--m-search--Background: var(--pf-c-form-control--m-search--BackgroundUrl) var(--pf-c-form-control--m-search--BackgroundPosition) / var(--pf-c-form-control--m-search--BackgroundSize) no-repeat;
 
   // Select
   --pf-c-form-control__select--PaddingRight: var(--pf-global--spacer--lg);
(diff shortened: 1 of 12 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -443,6 +451,23 @@
   padding-top: var(--pf-c-select__menu-group--not-first--PaddingTop);
 }
 
+.pf-c-select__menu-group {
+  &.pf-m-padding {
```
matthewcarleton: Why use a modifier here for padding? Wouldn't it always need the padding? If it is a modifier should it be called padding? I didn't think we named them like this.

src/patternfly/components/Select/select.scss
```diff
@@ -111,6 +111,14 @@
   // Match within a menu item when using typeahead
   --pf-c-select__menu-item--match--FontWeight: var(--pf-global--FontWeight--bold);
 
+  // Menu group padding
+  --pf-c-select__menu-group--m-padding--PaddingTop: var(--pf-c-select__menu--PaddingBottom);
+  --pf-c-select__menu-group--m-padding--PaddingRight: var(--pf-c-select__menu-item--PaddingRight);
+  --pf-c-select__menu-group--m-padding--PaddingBottom: var(--pf-c-select__menu--PaddingBottom);
+  --pf-c-select__menu-group--m-padding--PaddingLeft: var(--pf-c-select__menu-item--PaddingLeft);
+  --pf-c-select__menu-group--m-padding--not--first-of-type--PaddingTop: var(--pf-global--spacer--md);
```
matthewcarleton:
```suggestion
  --pf-c-select__menu-group--m-padding--not-first-of-type--PaddingTop: var(--pf-global--spacer--md);
```
do you need the `--` here?

src/patternfly/components/Select/select.scss
```diff
@@ -111,6 +111,14 @@
...
+  --pf-c-select__menu-group--m-padding--not--last-of-type--PaddingBottom: var(--pf-global--spacer--md);
```
matthewcarleton:
```suggestion
  --pf-c-select__menu-group--m-padding-not--last-of-type--PaddingBottom: var(--pf-global--spacer--md);
```
same question as above :)

src/patternfly/components/Select/select.scss
mattnolting: Yeah, we're using `.pf-m-padding` in some components (`.pf-c-drawer` most recently) where padding is a boolean and pre-configured. This is what I thought to be the most flexible approach as we build out functionality in the future opting for existing, generic elements over specific elements like `.pf-c-select__filter`.  Wdyt?

src/patternfly/components/Select/select.scss
mattnolting: Yeah, because the selector is `:not(:last-of-type)`, the second `-` denotes a pseudo selector.

@mcoker @christiemolloy wdyt?

src/patternfly/components/Select/select.scss
mattnolting: Same as above

src/patternfly/components/Select/select.scss
matthewcarleton: ah gotcha so more reusable this way. Ok, I'm behind on the newest rules for PF :)

mattnolting: and that `var` is wrong, lol

src/patternfly/components/Select/select.scss
mcoker: Agree, it's 2 pseudo selectors.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -72,6 +77,13 @@ $pf-c-form-control--success--Coordinates: "M504 256c0 136.967-111.033 248-248 24
   --pf-c-form-control--success--check--Background: var(--pf-c-form-control--success--BackgroundUrl) var(--pf-c-form-control--success--BackgroundPosition) / var(--pf-c-form-control--success--BackgroundSize) no-repeat;
   --pf-c-form-control--success--Background: var(--pf-c-form-control--BackgroundColor) var(--pf-c-form-control--success--check--Background);
 
+  // Input m-search
+  --pf-c-form-control--m-search--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-form-control--m-search--BackgroundPosition: var(--pf-c-form-control--PaddingLeft);
```
christiemolloy: Should the background position always align with the left padding of the form?

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -72,6 +77,13 @@ $pf-c-form-control--success--Coordinates: "M504 256c0 136.967-111.033 248-248 24
...
+  --pf-c-form-control--m-search--BackgroundSize: var(--pf-c-form-control--FontSize) var(--pf-c-form-control--FontSize);
```
christiemolloy: Should the background-size use font-size or line-height?

src/patternfly/components/Select/examples/Select.md
```diff
@@ -196,6 +203,7 @@ The checkbox select can select multiple items using checkboxes. The number of it
 | `.pf-c-select__menu-fieldset` | `<fieldset>` |  Initiates a fieldset for the items in a checkbox select. |
 | `.pf-c-select__menu-group` | `<div>` |  Initiates a group within a select menu. |
 | `.pf-c-select__menu-group-title` | `<div>` |  Initiates a title for a group with a select menu. |
+| `.pf-m-padding` | `.pf-c-select__menu-group` |  Adds padding to group. |
```
christiemolloy: Change comment to something similar to this: "Modifies the select menu group to add extra padding."

src/patternfly/components/Select/select.scss
christiemolloy: Yep, agreed

src/patternfly/components/Select/select-menu-filter.hbs
```diff
@@ -0,0 +1,4 @@
+{{#> select-menu-group select-menu-group--modifier="pf-m-padding"}}
```
christiemolloy: Should we follow what is in the context selector component? It has its own element wrapper for the input. The advantage of this is that it is specific for adding an input in it. While the modifier that you're applying to the `pf-c-select__menu-group` keeps it flexible, it may mean that users use it for the wrong reasons other than what its meant for.

Screen Shot 2020-02-11 at 1 49 39 PM

src/patternfly/components/FormControl/form-control.scss
mattnolting: Yeah, because the clear `(x)` appears on the right by default in the `input="search"`

src/patternfly/components/FormControl/form-control.scss
mattnolting: I copied these settings from the previous examples

src/patternfly/components/Select/select-menu-filter.hbs
mattnolting: Probably should do that. If in the future we want to add `.pf-c-select__menu-group.pf-m-padding` in the future, we can.

src/patternfly/components/Select/select-menu-input.hbs
```diff
@@ -0,0 +1,4 @@
+<div class="pf-c-select__menu-input{{#if select-menu-input--modifier}} {{select-menu-input--modifier}}{{/if}}">
+  {{#> form-control controlType="input" form-control--modifier="pf-m-search" form-control--attribute=(concat 'type="search" id="' id '-search-input" name="' id '-search-input" aria-label="Search"')}}
```
christiemolloy: What if in the future we want to put a different type of input into this wrapper? I guess we could leave this and then add a condition inside of here to either show search or a different type of input, wdyt?

src/patternfly/components/Select/select.scss
```diff
@@ -440,7 +446,14 @@
 }
 
 .pf-c-select__menu-group:not(:first-of-type) {
-  padding-top: var(--pf-c-select__menu-group--not-first--PaddingTop);
+  padding-top: var(--pf-c-select__menu-group--not--first-of-type--PaddingTop);
+}
+
+.pf-c-select__menu-input {
+  padding-top: var(--pf-c-select__menu-input--PaddingTop);
```
christiemolloy: could you just do `padding: ____ ____ ____ ____` here

src/patternfly/components/Select/select-menu-input.hbs
mattnolting: Love this comment/suggestion. **Nice**!!!

PR #2699, opened 2020-02-11
chore(fix): added files/dirs to cleanup

ddonahue007:


gulpfile.js
```diff
@@ -20,7 +20,22 @@ const hbsFiles = ['./src/patternfly/**/*.hbs'];
 const mdFiles = ['./src/patternfly/**/*.md'];
 
 function clean(cb) {
-  ['./dist', './src/icons/PfIcons'].forEach(dir => removeSync(dir));
+  const cleanDirs = [
+    './dist',
+    './src/icons/PfIcons',
+    'cache',
```
redallen: Shouldn't this be `.cache`? And doesn't `gatsby clean` already remove `.cache` (so we can just remove this line)?

ddonahue007: Done

package.json
```diff
@@ -14,7 +14,7 @@
     "build:pficonfont": "gulp pfIconFont",
     "build:ie11": "gulp buildIE11",
     "build:site": "npm run copy-assets && gatsby build",
-    "clean": "gatsby clean && shx -rf tmp",
+    "clean": "gatsby clean && gulp clean",
     "cli:setup": "cd build/patternfly-cli && npm install && cd ../../ && npm link ./build/patternfly-cli",
     "copy-assets": "gulp copyAssets",
     "dev": "concurrently \"npm run copy-assets\" \"gulp watchSASS\" \"gatsby develop\"",
```
redallen: Can you remove the "shx" dependency further down in this file and run `npm install`?

ddonahue007: I am not seeing the dependency in this branch

redallen: You're right! I must've removed it but forgotten to remove it from the "clean" target.

PR #2700, opened 2020-02-11
fix(inline-edit): hide groups by default

seanforyou23:
This PR makes a small change to the inline edit styles, to hide groups by default and then assign display once the activator class (pf-m-inline-editable) is present. This fixes a problem in Firefox where the group container reserves space even when all of its children are hidden, forcing it out of alignment with surrounding elements in table.

### DIFF
--- a/src/patternfly/components/InlineEdit/inline-edit.scss
+++ b/src/patternfly/components/InlineEdit/inline-edit.scss
@@ -74,6 +74,16 @@
 }
 
 // Editable
+.pf-c-inline-edit__group.pf-m-action-group {
+  display: none;
+  visibility: hidden;
+}
+
+.pf-m-inline-editable .pf-c-inline-edit__group {
+  display: flex;
+  visibility: visible;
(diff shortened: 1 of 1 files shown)


src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -18,7 +18,7 @@
 
 // Group
 .pf-c-inline-edit__group {
-  display: flex;
+  display: none;
```
christiemolloy: should you also add `visibility: hidden` here?

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -74,6 +74,10 @@
 }
 
 // Editable
+.pf-m-inline-editable .pf-c-inline-edit__group {
+  display: flex;
```
christiemolloy: and `visibility: visible` here too

Issue #2719, opened 2020-02-13
Add support for plain inline buttons


marshmalien:
#### SUMMARY
Button Variations: https://www.patternfly.org/v4/documentation/react/components/button#variations

The button prop `isInline` currently doesn't play well with `variant="plain"`.  In the example below, the DataListAction contains a plain button, but it adds about 10px of whitespace to the row height. 

Changing the button variant to `link` and adding `isInline` removes the extra whitespace, but we don't want the "link" styles. Could you add support for a "plain" "inline" button?

##### Plain 
<img width="501" alt="Screen Shot 2020-02-13 at 4 23 12 PM" src="https://user-images.githubusercontent.com/15881645/74479547-56102a80-4e7d-11ea-9fa5-52e5f4bd0fec.png">

##### Link and Inline
<img width="447" alt="Screen Shot 2020-02-13 at 4 22 47 PM" src="https://user-images.githubusercontent.com/15881645/74479534-527ca380-4e7d-11ea-918b-b12c3037d222.png">

marshmalien:
cc @trahman73

mcoker:
@marshmalien after looking at this some more, it looks like this is due to a bug in the react component. I opened a react issue for this yesterday https://github.com/patternfly/patternfly-react/issues/3860

The basic idea is that in core, we have a `<div class="pf-c-data-list__action"></div>` element that we use to wrap the plain button actions, and that class offsets the spacing of the button so that it better aligns with the row and doesn't push content down. That class wasn't implemented in the react component, resulting in this visual bug. You can see that class in use in core by going to the [data list examples page](https://www.patternfly.org/v4/documentation/core/components/datalist) and inspecting a row that has a kebab icon used for the dropdown action of a row.

Could you add this element to your implementation of the data list temporarily until we're able to resolve the bug in react?

marshmalien:
Great investigative work @mcoker. If there is a plan to add `.pf-c-data-list__action` to the data list action component then we can wait until the fix comes downstream. I would rather not locate and wrap every instance of our action buttons with a wrapper, only to remove it once this change lands.

mcoker:
@marshmalien agreed! Is it OK if we close this issue and you can track this bug in https://github.com/patternfly/patternfly-react/issues/3860?

marshmalien:
@mcoker Yes, sounds good to me 👍 Thank you for looking into this!

PR #2720, opened 2020-02-17
fix(button): added inline variation to plain button

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2719

### DIFF
--- a/src/patternfly/components/Button/examples/Button.md
+++ b/src/patternfly/components/Button/examples/Button.md
@@ -43,12 +43,17 @@ import './Button.css'
   {{/button-icon}}
 {{/button}}
 
+{{#> button button--modifier="pf-m-inline pf-m-link"}}
+  Inline link
+{{/button}}
+<br>
+<br>
 {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Remove"'}}
   <i class="fas fa-times" aria-hidden="true"></i>
 {{/button}}
 
-{{#> button button--modifier="pf-m-inline pf-m-link"}}
-  Inline link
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Button/examples/Button.md
```diff
@@ -50,6 +50,10 @@ import './Button.css'
 {{#> button button--modifier="pf-m-inline pf-m-link"}}
   Inline link
 {{/button}}
+
+{{#> button button--modifier="pf-m-inline pf-m-plain"}}
+  Inline link
```
mattnolting: Should this be an icon?

mcoker: Yes! Updated that and a couple of other things.

christiemolloy:
Is there no concern about target size for a button now? If not, it could be helpful to update some of the demos where the original button spacing wouldn't push the icon too far left or too far right and not align it with other text and elements.

mcoker:
@christiemolloy we still care about button size, and this would just be for edge cases - but re-reading the original issue, I originally thought this was a custom layout, but it looks like they're running into a bug I filed in react yesterday where plain button actions in the data list aren't wrapped in an element we use to offset the position of the button so it plays nicely with the row height - https://github.com/patternfly/patternfly-react/issues/3860

I could go either way merging this. On the one hand, I can see it being useful in edge cases, though it isn't accessible from the standpoint of the target size and that, unlike the `pf-c-button.pf-m-link.pf-m-inline` variation, there is no additional treatment to indicate the inline variation is clickable (ie, blue color, underline/color change on hover). WDYT?

mcoker:
Closing since the issue in #2719 is a bug with the data list component.

PR #2734, opened 2020-02-20
feat(vertical-divider): added vertical divider

### LINKED ISSUE #2148: Add vertical variation to divider component
A use case is in the new data toolbar component. Relevant - https://github.com/patternfly/patternfly-next/pull/2119#discussion_r313624056

mattnolting:
closes #2148 

**Divider component preview**
https://patternfly-next-pr-2734.surge.sh/documentation/core/components/divider

### DIFF
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -64,11 +64,14 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   --pf-c-data-toolbar__group--m-filter-group--spacer: var(--pf-c-data-toolbar__group--spacer);
   --pf-c-data-toolbar__group--m-filter-group--space-items: 0;
 
+  // remove during breaking change
   // Separator
   --pf-c-data-toolbar__item--m-separator--spacer: var(--pf-c-data-toolbar__item--spacer);
   --pf-c-data-toolbar__item--m-separator--BackgroundColor: var(--pf-global--BorderColor--100);
   --pf-c-data-toolbar__item--m-separator--Width: var(--pf-global--BorderWidth--md);
   --pf-c-data-toolbar__item--m-separator--Height: var(--pf-global--FontSize--lg);
+
+  // end - remove during breaking change
 
   // Overflow menu item
   --pf-c-data-toolbar__item--m-overflow-menu--spacer: var(--pf-c-data-toolbar__item--spacer);
@@ -89,12 +92,34 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   // Toggle
   --pf-c-data-toolbar__toggle--m-expanded__c-button--m-plain--Color: var(--pf-global--Color--100);
 
+  // Divider
+  --pf-c-data-toolbar--c-divider--m-vertical--spacer: var(--pf-c-data-toolbar--spacer--base);
+
   position: relative;
   row-gap: var(--pf-c-data-toolbar--RowGap);
   display: grid;
   padding-top: var(--pf-c-data-toolbar--PaddingTop);
   padding-bottom: var(--pf-c-data-toolbar--PaddingBottom);
   background-color: var(--pf-c-data-toolbar--BackgroundColor);
+}
+
+// Divider
+.pf-c-data-toolbar__content-section,
+.pf-c-data-toolbar__group {
+  // set this var here so specificity is 20
+  > .pf-c-divider {
+    --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar--c-divider--m-vertical--spacer);
+  }
+
+  > .pf-c-divider.pf-m-vertical {
+    align-self: stretch;
+    height: auto;
+    margin-right: var(--pf-c-data-toolbar--spacer);
+
+    &:last-child {
+      --pf-c-data-toolbar--spacer: 0;
+    }
+  }
 }
 
 // Group
@@ -171,6 +196,7 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
 
   margin-right: var(--pf-c-data-toolbar--spacer);
 
+  // remove during breaking change
   // Separator
   &.pf-m-separator {
     --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar__item--m-separator--spacer);
@@ -180,6 +206,8 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
     height: var(--pf-c-data-toolbar__item--m-separator--Height);
     background-color: var(--pf-c-data-toolbar__item--m-separator--BackgroundColor);
   }
+
+  // end - remove during breaking change
 
   // Overflow menu
   &.pf-m-overflow-menu {
--- a/src/patternfly/components/DataToolbar/data-toolbar.scss
+++ b/src/patternfly/components/DataToolbar/data-toolbar.scss
@@ -112,8 +112,6 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   }
 
   > .pf-c-divider.pf-m-vertical {
-    align-self: stretch;
-    height: auto;
     margin-right: var(--pf-c-data-toolbar--spacer);
 
     &:last-child {
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -18,6 +18,7 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
 
   display: flex;
   align-items: center;
+  align-self: stretch;
   justify-content: center;
   width: 100%;
   border: 0; // removes the default border styling on an hr
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -10,11 +10,13 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
 
   // end - update during breaking change
 
-  --pf-c-divider--m-vertical--after--Width: var(--pf-global--BorderWidth--sm);
-
   // After
   --pf-c-divider--after--FlexBasis: 100%;
   --pf-c-divider--after--Inset: 0%;
+
+  // Vertical
+  --pf-c-divider--m-vertical--after--FlexBasis: 100%;
+  --pf-c-divider--m-vertical--after--Width: var(--pf-global--BorderWidth--sm);
 
   display: flex;
   align-items: center;
@@ -41,7 +43,7 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
     max-height: 100%;
 
     &::after {
-      flex-basis: calc(var(--pf-c-divider--after--FlexBasis) - var(--pf-c-divider--after--Inset));
+      flex-basis: calc(var(--pf-c-divider--m-vertical--after--FlexBasis) - var(--pf-c-divider--after--Inset));
       width: var(--pf-c-divider--m-vertical--after--Width);
     }
   }
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -15,7 +15,6 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   // After
   --pf-c-divider--after--FlexBasis--base: 100%;
   --pf-c-divider--after--Inset: 0;
-  --pf-c-divider--after--FlexBasis: calc(var(--pf-c-divider--after--FlexBasis--base) - var(--pf-c-divider--after--Inset) - var(--pf-c-divider--after--Inset));
 
   display: flex;
   align-items: center;
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -53,6 +53,10 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
 
     @include pf-apply-breakpoint($breakpoint, $pf-c-divider--breakpoint-map) {
       @each $spacer, $spacer-value in $pf-c-divider--spacer-map {
+        @if $spacer == none {
+          $spacer-value: 0%;
+        }
+
         &.pf-m-inset-#{$spacer}#{$breakpoint-name} {
           --pf-c-divider--after--Inset: #{$spacer-value};
         }
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -20,6 +20,7 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   align-items: center;
   align-self: stretch;
   justify-content: center;
+  width: 100%;
   border: 0; // removes the default border styling on an hr
 
   &::after {
@@ -34,6 +35,7 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   &.pf-m-vertical {
     display: inline-flex;
     flex-direction: column;
+    width: auto;
     height: inherit;
     min-height: 100%;
     max-height: 100%;
--- a/src/patternfly/components/Divider/divider.scss
+++ b/src/patternfly/components/Divider/divider.scss
@@ -1,9 +1,21 @@
 .pf-c-divider {
   --pf-c-divider--Height: var(--pf-global--BorderWidth--sm);
   --pf-c-divider--BackgroundColor: var(--pf-global--BorderColor--100);
+  --pf-c-divider--m-vertical--Width: var(--pf-global--BorderWidth--md);
 
   height: var(--pf-c-divider--Height);
   background-color: var(--pf-c-divider--BackgroundColor);
   border: 0; // removes the default border styling on an hr
   flex-shrink: 0;
+
+  &.pf-m-vertical {
+    align-self: center;
+    width: var(--pf-c-divider--m-vertical--Width);
+    height: 100%;
+
+    &.pf-m-stretch {
+      align-self: stretch;
+      height: auto;
+    }
+  }
 }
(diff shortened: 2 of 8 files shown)


redallen:
Failing lint check can be ignored. It's happening because the size of divider increased 53%.

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -24,6 +25,21 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
   &:last-child {
     --pf-l-flex--spacer: 0;
   }
+
+  // set this var here so specificity is 20
```
mcoker: I really think we should leave these styles out of the flex layout. It goes against the principle that we have about the role of layouts vs components and how [layouts are used for positioning, and don't impose styles on their children](https://pf4.patternfly.org/guidelines#layouts), and I don't see a compelling reason for this exception in this PR. I would be in favor of keeping this in our back pocket and if there is a specific product requirement that makes this type of exception necessary, we can reconsider at that point.

src/patternfly/components/OverflowMenu/examples/OverflowMenu.md
```diff
@@ -276,6 +277,5 @@ The action group consists of a primary and secondary action. Any additional acti
 | `.pf-c-overflow-menu__control` | `<div>` | Initiates the overflow menu control. **Required** |
 | `.pf-c-overflow-menu__group` | `<div>` | Initiates an overflow menu group. |
 | `.pf-c-overflow-menu__item` | `<div>` | Initiates an overflow menu item. **Required** |
-| `.pf-c-overflow-menu__separator` | `<li>` | Defines a separator within the menu. |
```
mcoker: Since this is a promoted component, we want to keep the documentation for the old separator around - not only here but in examples, so that we don't inadvertently change something that breaks the existing `__separator` until it's removed and no longer supported. We have a breaking change issue to update components' implementation of their own separators with the divider component that this work would be great to included in - https://github.com/patternfly/patternfly-next/issues/2621

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,9 +1,23 @@
...
   height: var(--pf-c-divider--Height);
   background-color: var(--pf-c-divider--BackgroundColor);
   border: 0; // removes the default border styling on an hr
   flex-shrink: 0;
+
+  &.pf-m-vertical {
+    align-self: center;
+    width: var(--pf-c-divider--m-vertical--Width);
+    height: 100%;
+    min-height: var(--pf-c-divider--m-vertical--MinHeight);
```
mcoker: The CSS for this looks solid, but I'm not sure we want to impose a `min-height` on the divider as we don't know the context in which it will be used and it's something a user would need to override.

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -33,3 +37,5 @@ The divider renders as an `<hr>` by default. It is possible to make the divider
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
+| `.pf-m-vertical` | `.pf-c-divider` | Modifies the divider component from horizontal to vertical. |
+| `.pf-m-stretch` | `.pf-m-vertical` | Modifies the vertical divider to stretch to full height. This modifier requires that the parent has an explicit or implicit height value. |
```
mcoker: I think "This modifier requires that the parent has an explicit or implicit height value" belongs in the previous line (`.pf-c-divider.pf-m-vertical`) - the base variation is the standard inline/inline-block use that requires a height value. `.pf-m-stretch` removes `height`, so we should document it's use is required to be used as the child of a flex parent.

src/patternfly/components/DataToolbar/examples/DataToolbar.md
```diff
@@ -90,7 +90,6 @@ Several components in the following examples do not include functional and/or ac
 | -- | -- | -- |
 | `.pf-c-data-toolbar__item` | `<div>` | Initiates the toolbar component item. **Required** |
 | `.pf-c-data-toolbar__group` | `<div>` | Initiates the toolbar component group. |
-| `.pf-m-separator` | `.pf-c-data-toolbar__item` | Modifies item to a separator. |
```
mcoker: This is fine since it's experimental, but we could also save this change out and include it in https://github.com/patternfly/patternfly-next/issues/2621 since we're going to making this change globally, and it would keep the scope of this issue to just introducing the feature.

src/patternfly/components/Divider/divider.scss
mattnolting: I get that, but I believe that not including a `min-height` will lead to user confusion as they may think it's broken or doesn't work at all because they won't see anything unless it's put into the right context. This way, it may look incorrect, but at least it will be visible. And if inserted into an element with the needed height (but not an implicit layout height), it will still consume 100% height. It think it's a sound approach.

src/patternfly/layouts/Flex/flex.scss
mattnolting: I feel strongly that we should make this exception now and consider `pf-c-divider` a special type of component, maybe the only of its type. In the preview, try putting the divider into a `pf-l-flex__item`. It does not behave as expected. We'd be better off to just use the existing separator. So, in order to use a component like `pf-c-divider` in place of all component specific divider/separators (in this case `pf-l-flex__separator`), all or most of them will require some change to the component/layout css for margin or placement updates, including the default, horizontal `divider`. 

In the case of `.pf-l-flex`, it would require a wrapper modifier, `.pf-l-flex__item.pf-m-divider` then placing `.pf-c-divider` inside that modified `flex item`. That to me seems redundant/inefficient, whereas we can just say, `.pf-c-divider` is a wildcard component and is the exception to the [the rule above](https://pf4.patternfly.org/guidelines#layouts). Since all layouts and components can use it, it gets special consideration. 

I also think our developers will appreciate this exception, as it's one less component specific element to use. If something needs to be divided, use `pf-c-divider` and it behaves as PF intended without worrying about how to configure the correct wrapping elements. Each component considers it and controls it specifically, if needed.

This approach is only an enhancement to the system. The only drawback is that it breaks the rules we wrote before we knew we needed to make exceptions to those rules.

mattnolting: One more note: by using this method, `divider` responds to placement modifiers `.pf-m-space-items-{spacer}` applied to flex parents.  Which makes it responsive to the the built in spacing modifiers without having to explicitly adjust.

src/patternfly/components/Divider/divider.scss
mcoker: > I believe that not including a min-height will lead to user confusion as developers may think it's broken or doesn't work at all because they won't see anything unless it's put into the right context. 

That's where the documentation should come in. I don't think this would actually solve a use case for the divider, it would only serve as a hint that they need to manage the height in some regard, but it comes with the cost of potentially imposing an unwanted height. If the documentation is clear, I don't think this is necessary.

src/patternfly/layouts/Flex/flex.scss
mcoker: > So, in order to use a component like pf-c-divider in place of all component specific divider/separators (in this case pf-l-flex__separator)

The flex layout doesn't have a separator, and at the point that we add one, it becomes a component at least by our current definition of a layout vs a component. 

> In the case of .pf-l-flex, it would require a wrapper modifier, .pf-l-flex__item.pf-m-divider then placing .pf-c-divider inside that modified flex item.

A known usage of the vertical variation is that the height of the parent needs to be managed in some way. You can do that now by using what's already built into the flex layout with `pf-m-align-self-stretch` if you wanted the divider to be a child of a flex item, or use the dividers `pf-m-stretch` if the divider is itself a flex child. https://codepen.io/mcoker/pen/abOBWgX

I get that it's an enhancement and could be beneficial, but it's inconsistent with the way layouts are built and we already have in place what we need to allow a divider to be used in a flex layout based off of the documentation in the divider component. I think using the modifier that already exists on the flex layout is consistent with how you would use the vertical divider in any other context. That's why I don't see there being a compelling reason to make an exception to the guidelines for this use case. 

At a minimum though I think this is beyond the scope of this PR of just adding the variation. It sounds like this might be a feature request to add dividers to the layouts, or the flex based ones at least. Maybe if this is a constant pain point for users we can consider making an exception and revisiting?

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -89,12 +92,30 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
...
   padding-top: var(--pf-c-data-toolbar--PaddingTop);
   padding-bottom: var(--pf-c-data-toolbar--PaddingBottom);
   background-color: var(--pf-c-data-toolbar--BackgroundColor);
+
+  // set this var here so specificity is 20
+  .pf-c-divider {
+    --pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar--c-divider--m-vertical--spacer);
+  }
+
+  .pf-c-divider.pf-m-vertical {
```
mcoker: Since these styles aren't scoped, they'll apply to any use of the vertical divider in the component. I think it's fair to assume we or someone might use a divider as part of another component, say in a dropdown item or toggle or something like that. Can we scope the styles somehow? It would be fair to keep the separator container to manage the divider positioning, then use the divider in that, similar to how we do that in other components to handle the positioning of a component in a component.

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -33,3 +37,5 @@ The divider renders as an `<hr>` by default. It is possible to make the divider
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
+| `.pf-m-vertical` | `.pf-c-divider` | Modifies the divider component from horizontal to vertical. This modifier requires that the parent has an explicit or implicit height value. |
+| `.pf-m-stretch` | `.pf-m-vertical` | Modifies the vertical divider to stretch to full height. |
```
mcoker: Can we make note that this requires the parent have a flex layout, or for the divider to be a flex item?

src/patternfly/components/OverflowMenu/overflow-menu.scss
```diff
@@ -17,6 +18,21 @@
 
   display: inline-flex;
   align-items: center;
+
+  // set this var here so specificity is 20
+  .pf-c-divider {
```
mcoker: Same note here re: scope - it's fair to assume a user or we might introduce a button or split button or something with a divider and these styles would conflict.

src/patternfly/components/DataToolbar/data-toolbar.scss
mattnolting: Scoped 👍

mattnolting:
@mcoker updated

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,9 +1,63 @@
+$pf-c-divider--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl", "2xl");
+$pf-c-divider--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl", "3xl");
+
 .pf-c-divider {
+  // update during breaking change
   --pf-c-divider--Height: var(--pf-global--BorderWidth--sm);
   --pf-c-divider--BackgroundColor: var(--pf-global--BorderColor--100);
+  --pf-c-divider--after--Height: var(--pf-c-divider--Height);
```
mcoker: 👍

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -33,3 +60,5 @@ The divider renders as an `<hr>` by default. It is possible to make the divider
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
+| `.pf-m-vertical` | `.pf-c-divider` | Modifies the divider component from horizontal to vertical. This modifier requires that the parent has an explicit or implicit height, or has a flex or grid based layout parent. |
+| `.pf-m-inset-{none, sm, md, lg, xl, 2xl, 3xl}{-on-[md, lg, xl, 2xl]}` | `.pf-c-divider` | Modifies a divider inset to match spacer system. |
```
mcoker:
```suggestion
| `.pf-m-inset-{none, sm, md, lg, xl, 2xl, 3xl}{-on-[sm, md, lg, xl, 2xl]}` | `.pf-c-divider` | Modifies a divider inset to match spacer system. |
```

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -4,24 +4,51 @@ section: components
...
 </ul>
 ```
 
 ```hbs title=<div>
-{{#> divider divider--type="div"}}{{/divider}}
+{{> divider divider--type="div"}}
+```
+
+```hbs title=Insets
+<div>No inset</div>
```
mcoker: I was expecting the inset in the example to match the title - not match the previous examples plus what's in the title. Maybe we could update the descriptions to describe all of the insets it uses? I also think it's fine just to have a couple of examples. Maybe a simple one "small inset on sm" then a complex one like "md inset, no inset on md, 3xl inset on lg, lg inset on xl"

mattnolting:
@mcoker updated

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,9 +1,63 @@
...
+  display: flex;
+  align-items: center;
+  align-self: stretch;
+  justify-content: center;
+  width: 100%;
   border: 0; // removes the default border styling on an hr
-  flex-shrink: 0;
+
+  &::after {
+    flex-basis: calc(var(--pf-c-divider--after--FlexBasis) - calc(var(--pf-c-divider--after--Inset) * 2));
```
mcoker: Looks like any time `--pf-c-divider--after--Inset` is `0`, the `calc()` is going to fail due to a type mismatch.

mattnolting: Good catch!

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -33,3 +56,5 @@ The divider renders as an `<hr>` by default. It is possible to make the divider
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
+| `.pf-m-vertical` | `.pf-c-divider` | Modifies the divider component from horizontal to vertical. This modifier requires that the parent has an explicit or implicit height, or has a flex or grid based layout parent. |
+| `.pf-m-inset-{none, sm, md, lg, xl, 2xl, 3xl}{-on-[sm, md, lg, xl, 2xl]}` | `.pf-c-divider` | Modifies a divider inset to match spacer system. |
```
christiemolloy: Can you explain further what inset is here, I think it might help to include more info for designers/devs here

christiemolloy: And I'm a little confused too

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -4,24 +4,47 @@ section: components
...
+
+```hbs title=Vertical
+{{> divider divider--type="div" divider--modifier="pf-m-vertical pf-m-inset-md"}}
+```
+
+```hbs title=Vertical,-inset-medium
+{{> divider divider--type="div" divider--modifier="pf-m-vertical pf-m-inset-md"}}
+```
+
+```hbs title=Vertical,-md-inset,-no-inset-on-md,-lg-inset-on-lg,-sm-inset-on-xl
```
christiemolloy: Why does this have medium inset and no inset on medium?

mattnolting: Its purpose is to illustrate removing inset.

src/patternfly/components/Divider/examples/Divider.md
mattnolting: @christiemolloy updated

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,9 +1,67 @@
...
+    height: var(--pf-c-divider--after--Height);
+    content: "";
+    background-color: var(--pf-c-divider--after--BackgroundColor);
+    justify-self: center;
+  }
+
+  &.pf-m-vertical {
+    display: inline-flex;
+    flex-direction: column;
+    width: var(--pf-c-divider--m-vertical--Width);
```
christiemolloy: why does this have a width set here and on the pseudo class

mattnolting: Good catch, it was originally to unset `width: 100%`, but I don't think we need that either

src/patternfly/components/DataToolbar/data-toolbar.scss
```diff
@@ -97,6 +103,23 @@ $pf-c-data-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
   background-color: var(--pf-c-data-toolbar--BackgroundColor);
 }
 
+// Divider
+.pf-c-data-toolbar__content-section,
+.pf-c-data-toolbar__group {
+  // set this var here so specificity is 20
+  > .pf-c-divider {
```
christiemolloy: why is `pf-c-divider` set to `--pf-c-data-toolbar--c-divider--m-vertical--spacer` but below `.pf-c-divider.pf-m-vertical` is set to `--pf-c-data-toolbar--spacer`

mattnolting: `--pf-c-data-toolbar--spacer` is a utility variable. It isn't designed to define strict values, rather it is meant to be manipulated, but as a consequence it's very sensitive to specificity. So I set `--pf-c-data-toolbar--spacer: var(--pf-c-data-toolbar--c-divider--m-vertical--spacer);` on `__group` and `__content-section` to retain a specificity of  20, so that in can later be overwritten by a modifier or by a custom theme. Doing this allows for consumers to override the base spacer value `--pf-c-data-toolbar--c-divider--m-vertical--spacer` for the vertical spacer AND use `data-toolbar` s modifiers as if it belonged to the component. 

Because it has no assignment in horizontal/default `pf-c-divider` the variable is ignored until it's called in `pf-m-vertical`

src/patternfly/components/Divider/divider.scss
mattnolting: Update: we need `width: 100%`. Vertical width is now `auto`

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,9 +1,67 @@
...
+    height: inherit;
+    min-height: 100%;
+    max-height: 100%;
+
+    &::after {
+      flex-basis: calc(var(--pf-c-divider--after--FlexBasis) - var(--pf-c-divider--after--Inset));
```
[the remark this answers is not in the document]
mattnolting: Nice.

PR #2745, opened 2020-02-25
feat(radio/check): custom styles - WIP

mcoker:
WIP - don't merge. Just a quick demo to show what custom radio and checkboxes could look like in patternfly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build": "npm run build-patternfly && npm run build:site",
     "build:pficons": "gulp pfIcons",
     "build:pficonfont": "gulp pfIconFont",
-    "build:ie11": "gulp buildIE11",
+    "build:ie11": "true",
     "build:site": "npm run copy-assets && gatsby build",
     "clean": "gatsby clean && gulp clean",
     "cli:setup": "cd build/patternfly-cli && npm install && cd ../../ && npm link ./build/patternfly-cli",
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -48,13 +48,17 @@
   -webkit-appearance: none;
   -moz-appearance: none;
 
+  --d-t: .3s;
+  --d-t-e: ease;
+  --d-o: .2s;
+
   &::after {
     position: absolute;
     top: 0;
     left: 0;
     display: block;
     content: "";
-    transition: transform var(--d-t, .3s) var(--d-t-e, ease), opacity var(--d-o, .2s);
+    transition: transform var(--d-t) var(--d-t-e), opacity var(--d-o);
   }
 
   &:checked {
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -98,16 +98,20 @@
   }
 
   &:not(.switch) {
+    --o: 0;
+
     width: 21px;
 
     &::after {
-      opacity: var(--o, 0);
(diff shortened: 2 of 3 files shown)


mcoker:
@mcarrano they're 21x21, but that's easily adjusted.

> Is the idea that we would offer these side by side with the native controls or replace the native controls @mcoker ?

I'm inclined to suggest they're only supported in the check and radio components - not just any check or radio someone might add. So a user could still use native controls, they would just need to not use our component. This would be similar to our button and form-control text input, textarea, and select menu styles. We only style those controls when using our component.

>  My only concern would be how these affect spacing in places like table rows, for example.

Yep, we use checkboxes in a number of places we would need to update. Currently in table, data-list, etc - anywhere we just have a single checkbox with no label, we are not using the check component. Our check component is currently only used when you have a checkbox + label pair (like in the context of a form). So I think the first update here would be to enable using a standalone check/radio component without a label. Then the next step would be to swap out the current standalone checkboxes in tables, data-lists, etc with the check component, and adjust the styling in those components to work with the new check styles. We could leave in the styles for the checkboxes that are currently used, too, so that you could use the natively styled control or the newly styled control if that helps us avoid a breaking change.

mcarrano:
Thanks for the update on that @mcoker . Would be interested to hear from @jessiehuff regarding optimal sizing. 21 x 21 still feels small, but maybe we would offer some options.

While I am happy to see this demo, we need to figure out where this belongs on the roadmap. Would this change be breaking @mcoker ?

cc @LHinson

jessiehuff:
After looking into this, it does seem like this route makes sense for accessibility, and there is some [research](https://designnotes.blog.gov.uk/2016/11/30/weve-updated-the-radios-and-checkboxes-on-gov-uk/) that supports larger sizes. However, I couldn't quite find any specifications or recommendations of exactly what size that would be. There are clear WCAG standards in terms of [focus/target size](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=211%2C251%2C255#target-size) of elements, but I could not find anything on size of the elements themselves. I'm happy to leave the specific size itself up to design.

mcoker:
@mcarrano 

> Would this change be breaking @mcoker ?

Likely not breaking in the code, but breaking visually. It could be offered as a "styled" or some other named opt-in variation. You can see how it impacts components by looking at other components in the PR preview
* https://patternfly-next-pr-2745.surge.sh/documentation/core/components/login/basic
* https://patternfly-next-pr-2745.surge.sh/documentation/core/demos/basicforms
* https://patternfly-next-pr-2745.surge.sh/documentation/core/components/select
* https://patternfly-next-pr-2745.surge.sh/documentation/core/components/inlineedit

We would also need to update some components to use the check/radio component instead of just using standalone check/radio inputs. For example [data-list](https://patternfly-next-pr-2745.surge.sh/documentation/core/components/datalist) and [table](https://patternfly-next-pr-2745.surge.sh/documentation/core/components/table), but I suspect the impact there would be minimal.

mcarrano:
@mcoker Generally I like this a lot better and feel like we should just go for it. The only place this concerned me was in the inline edit example where you had checkboxes and radio buttons vertically stacked, Can you create an example of vertically stacked radios and checkboxes on a form? I'd like to see what that looks like. @mceledonia @jessiehuff you may also want to take a look.

mcoker:
closing for now, we can re-open when we are ready to try and implement.

PR #2748, opened 2020-02-26
chore(workspace-blueprints): added child and concat to bp

mattnolting:


build/patternfly-cli/pf.js
```diff
@@ -66,13 +66,21 @@ program
...
     {{{${bemEntity}--attribute}}}
   {{/if}}`,
-      moduleHbOpen: `{{#> ${dasherizedName}}}`,
+      attributeItemBlock: `id="{{${dasherizedName}-id}}-{{${dasherizedName}-item--id}}"
+  {{#if ${bemEntity}-item--attribute}}
+    {{{${bemEntity}-item--attribute}}}
+  {{/if}}`,
+      moduleHbOpen: `{{#> ${dasherizedName} ${dasherizedName}--id="${dasherizedName}--basic-example"}}`,
       moduleHbClose: `{{/${dasherizedName}}}`,
+      moduleHbIdConcatenationOpen: `{{#> ${dasherizedName}-item ${dasherizedName}--id="item"}}`,
```
mcoker: Should we use `{{#> foo-bar-item foo-bar--id=(concat foo-bar--id "-item")}}` here?

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
```diff
@@ -2,14 +2,43 @@
...
+### Overview
+
+`{{nameDasherized}}` is a component. Some of its elements are presentational, while others require accessibility support.
+
+### Accessibility
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `hidden` | `.{{nameDasherized}}__item` |  Indicates that the `.{{nameDasherized}}__item` is hidden. |
+| `aria-hidden` | `.{{nameDasherized}}__item` |  Indicates that the `.{{nameDasherized}}__item` is hidden from and unfocusable to screen readers, but remains visible for sighted users. |
```
mcoker: If the description is that the thing is hidden, we'll need to specify the value as true since the default value is false.

```suggestion
| `aria-hidden="true"` | `.{{nameDasherized}}__item` |  Indicates that the `.{{nameDasherized}}__item` is hidden from and unfocusable to screen readers, but remains visible for sighted users. |
```

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
```diff
@@ -2,14 +2,43 @@
...
+| `aria-label` | `.{{nameDasherized}}__item` | Provides an accessible name for a test element. **Required** |
```
mcoker: most of our examples use `aria-label="[descriptive text]"` in the docs

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
```diff
@@ -2,14 +2,43 @@
...
+| `aria-expanded="true"` | `.{{nameDasherized}}__item` | Indicates that the expandable content is visible. **Required** |
+| `aria-expanded="true"` | `.{{nameDasherized}}__item` | Indicates that the expandable content is hidden. **Required** |
```
mcoker: do we need `aria-expanded`? It isn't as commonly used as the others.

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
```diff
@@ -2,14 +2,43 @@
...
+
+### Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.{{nameDasherized}}` | `<div>` | Initiates the {{nameDasherized}} component. **Required** |
```
mcoker: can we be consistent with the a11y docs and make a code block with the component class in these descriptions?

```suggestion
| `.{{nameDasherized}}` | `<div>` | Initiates the `.{{nameDasherized}}` component. **Required** |
```

mcoker: is there a lower-case and spaced var for spitting out the name here? so it reads "Initiates the about modal box component" instead of "Initiates the about-modal-box component"? that's probably our most common usage.

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
mcoker: if you keep it, need to update the hidden one

```suggestion
| `aria-expanded="false"` | `.{{nameDasherized}}__item` | Indicates that the expandable content is hidden. **Required** |
```

build/patternfly-cli/pf.js
```diff
@@ -66,13 +66,21 @@ program
...
+      bemItemModifierBlock: `{{#if ${bemEntity}-item--modifier}} {{${bemEntity}-item--modifier}}{{/if}}`,
       moduleName,
       partialBlock: '{{> @partial-block}}',
-      attributeBlock: `{{#if ${bemEntity}--attribute}}
+      attributeBlock: `
+  {{#if ${bemEntity}--attribute}}
     {{{${bemEntity}--attribute}}}
   {{/if}}`,
-      moduleHbOpen: `{{#> ${dasherizedName}}}`,
+      attributeItemBlock: `id="{{${dasherizedName}-id}}-{{${dasherizedName}-item--id}}"
```
mcoker:
```suggestion
      attributeItemBlock: `id="{{${dasherizedName}--id}}-{{${dasherizedName}-item--id}}"
```

build/patternfly-cli/pf.js
mcoker:
```suggestion
      moduleHbIdConcatenationOpen: `{{#> ${dasherizedName}-item ${dasherizedName}-item--id="item"}}`,
```

build/patternfly-cli/pf.js
```diff
@@ -66,13 +66,21 @@ program
       nameCapitalized: inflection.capitalize(name),
       bemName: `${bemName}`,
       bemModifierBlock: `{{#if ${bemEntity}--modifier}} {{${bemEntity}--modifier}}{{/if}}`,
+      bemItemModifierBlock: `{{#if ${bemEntity}-item--modifier}} {{${bemEntity}-item--modifier}}{{/if}}`,
       moduleName,
       partialBlock: '{{> @partial-block}}',
-      attributeBlock: `{{#if ${bemEntity}--attribute}}
+      attributeBlock: `
+  {{#if ${bemEntity}--attribute}}
```
mcoker: This is creating an empty line above the attribute block in the main component hbs partial

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
mattnolting: My thought here is include most of the the aria attributes w/descriptions that we use to drive consistency.

mattnolting: Don't have to, just a thought that might make our lives easier w/copy / pasta

build/patternfly-cli/pf.js
mcoker: A couple of thoughts about this way of defining an `id` attribute on the generated markup of the `__item` instead of using `concat()`.

If we're going to do it this way, I think we should wrap it in a conditional so it doesn't generate an `id` by default, otherwise I think it would be confusing to consumers and react devs that we have `id`'s on elements that aren't used for anything.

Also it's a bit inconsistent if the `__item` generates an ID this way but the main component hbs partial doesn't. We've run into issues before where we've used `(concat 'id="' component--id '-item"')` on a partial when it supported `component-item--id`, generating 2 `id`'s - since it's generally the case that you specify an `id` using `concat()`. I think it would be good to have a consistent approach instead of having to check hbs components for which way it supports adding an `id` to the element in the partial.

If we want to be consistent with this new approach and update the main component hbs to also support generating an `id` in the hbs from the `component--id` attribute, that will create an unnecessary `id` on the main component element since we'll likely always have `component--id` defined, which I think we should avoid.

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
mcoker: Yeah I'm just thinking we'll end up removing it in most cases. I'm fine either way though. If you want to keep it just need to update the 2nd one to `aria-expanded="false"`

build/patternfly-cli/pf.js
mattnolting: @mcoker try that out

build/blueprints/template/files/src/patternfly/__typeDirectory__/__namePascalized__/examples/__namePascalized__.md
mattnolting: We can expand on it ;D

PR #2754, opened 2020-02-27
fix(login): centered login on larger viewports

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2729

### DIFF
--- a/src/patternfly/components/Login/login.scss
+++ b/src/patternfly/components/Login/login.scss
@@ -1,7 +1,7 @@
 .pf-c-login {
   --pf-c-login--PaddingTop: var(--pf-global--spacer--lg);
   --pf-c-login--PaddingBottom: var(--pf-global--spacer--lg);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Login/login.scss
```diff
@@ -117,7 +116,6 @@
 }
 
 .pf-c-login__container {
-  width: 100%;
```
rhamilto, line 120: This change resulted in a regression where the container can be smaller than the viewport at mobile.

oauth-openshift apps rhamilto devcluster openshift com_oauth_authorize_client_id=console redirect_uri=https%3A%2F%2Fconsole-openshift-console apps rhamilto devcluster openshift com%2Fauth%2Fcallback response_type=code scope=user%3Afull  (1)

www patternfly org_v4_documentation_react_components_loginpage_basic(iPhone 6_7_8)


I'm pretty sure we're going to want this back.

mcoker: Thanks for catching that Robb, I'll try and look at https://github.com/patternfly/patternfly/issues/2913 later today.

mcoker: I've put up a PR and added a comment to the code so hopefully that shouldn't be removed again. https://github.com/patternfly/patternfly/pull/2917

PR #2764, opened 2020-03-03
fix(about-modal): add a comment to test v4 release

redallen:
Testing if PR previews and releasing to `prerelease-v4` on NPM will work.


redallen:
Ahhhh, bad semantic release!

Debugging by just pushing to the `v4` branch now.

PR #2766, opened 2020-03-03
fix(form): refactored label/control sections, added label help

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2703

@mcarrano this adds the icon next to a form label for the help popover. I'm using 14px font size for the icon label and 4px spacer (same as spacer between label text and the required *). How does that look?

## Breaking changes
This PR adds 2 new elements used for the overall form layout. These elements are the main sections that make up a `.pf-c-form__group`.
* `.pf-c-form__group-label` - defines the label section of the form. Anything that is above the form control element(s) in a stacked form, and to the left of form control element(s) in a horizontal form. Primarily this will hold the `.pf-c-form__label`. Also as a change introduced in this PR, it will hold the field level help element `.pf-c-form__label-help`.
  * All `.pf-c-label` and any other elements that should be grouped with the label should now be wrapped in `.pf-c-form__group-label`
* `.pf-c-form__group-control` - defines the control section of the form. This is where form control elements are placed. It will hold the form elements and form element helper text.
  * All form control elements (`.pf-c-form-control`, `.pf-c-check`, `.pf-c-radio`, etc ) should be wrapped in a single `.pf-c-form__group-control` element per form group.

`.pf-m-border` and `.pf-c-form__section` have been removed from the documentation as we currently do not have styles for those. Any reference to those should be removed as patternfly does not have any styles for them.

The following modifiers have moved from one element to another. Any reference to them should be moved from the old element to the new element
* `.pf-m-no-padding-top` has moved from `.pf-c-form__label` to `.pf-c-form__group-label`
* `.pf-m-inline` has been moved from `.pf-c-form__group` to `.pf-c-form__group-control`

* `--pf-c-form__label--Color`
* `--pf-c-form__label--FontWeight`
* `--pf-c-form__group--MarginLeft`
* `--pf-c-form--m-horizontal--md__group--GridTemplateColumns`
  * This was used to define the label and form control column widths in horizontal forms. You can now modify the label column width via `--pf-c-form--m-horizontal__group-label--md--GridColumnWidth` and the form control column width via `--pf-c-form--m-horizontal__group-label--md--GridColumnWidth`

The following variables have been renamed. Any reference to the old variable should be updated to use the new variable:
* `--pf-c-form--m-inline--MarginRight` has changed to `--pf-c-form__group-control--m-inline--child--MarginRight`
* `--pf-c-form--m-error--Color` has changed to `--pf-c-form__helper-text--m-error--Color`
* `--pf-c-form--m-success--Color` has changed to `--pf-c-form__helper-text--m-success--Color`

Some of the general CSS structure has also changed, so please reference the [file changeset](https://github.com/patternfly/patternfly/pull/2766/files) or the [form component stylesheet](https://github.com/patternfly/patternfly/blob/v4/src/patternfly/components/Form/form.scss) to see the new structure.


mcarrano:
Looks good to me @mcoker .  But shouldn't the help icon have a hover state as an active element?

@mceledonia @bdellasc can you also take a look? Is this the icon we decided that we want to use for field level help on forms?

mcoker:
@mcarrano good question about the hover state. This is just a placeholder for an icon, which a user could wrap with a button, which would provide hover styles, or they could just attach a tooltip to the icon itself similar to this react example, which doesn't have any sort of hover state https://www.patternfly.org/v4/documentation/react/components/tooltip#on-icon

Is it the case that when using a tooltip or popover on an element, it should receive some sort of interactive styling, like a pointer cursor?

bdellasc:
The idea that @mceledonia and I were kicking around was: in a layout with many info icons in close proximity (form, table, list), we'd use the outline version to reduce visual noise. For one-offs and layouts with a few here and there we could use the solid version of the icon.

mcarrano:
@mcoker In this case we want the help icon to open a popover and not a tooltip so I'd expect this to use a button rather than a static icon. Per @bdellasc 's point, I think this is OK as long as it's clear that any icon can be used here. Does that make sense?

mcoker:
@mcarrano yes, in the react implementation, a user would specify the icon they want to use. If we want to use a button here, using our standard plain button would create too much space around the button and offset the label's positioning. Would this be a use case for an "inline" variation of the plain button (a plain button with no extra padding)? It's also worth noting that the target size of this button will be pretty small since I'm using 14px for the icon size.

mcarrano:
Those are really good questions about the icon @mcoker . Not sure what the right thing to do is, but does not seem right as it is now. @mceledonia any thoughts?

mcoker:
@mcarrano I originally built this just as a container for an icon, so you could put just an icon or an icon with popover or whatever you want in it. If this is going to be used specifically for a help popover, I can refactor it to handle that specific use case, and we can build the button to trigger the popover and the popover itself into the component, so a user would just need to specify the icon they want to use and popover's content. We could abstract the button used to an inline variation of the plain button if we think that variation would be useful, or I could just build that button as a unique button for this use case as the help popover, and not use the button component.

mcarrano:
I like that approach @mcoker . We might want to allow the user to substitute a different icon, but basically the intent is to open help in a popover. I don't see the need to use this for other reasons.

mmenestr:
@mcarrano I can confirm that the popover icon should be an un-filled question mark circle ([fa-question-circle, regular style](https://fontawesome.com/icons/question-circle?style=regular)) !

Also not sure if this PR affects buttons in a form at all but, if it does, I also want to mention that the action group should be a primary button and a **link button**, instead of a primary button and a **secondary **button** as currently seen in the demo!

mcoker:
@mmenestr @mcarrano 

>  I can confirm that the popover icon should be an un-filled question mark circle (fa-question-circle, regular style) !

We don't ship font awesome regular with PF4 core, however we have a custom pf "help" icon I can use. Do you see any issues with using that?

Screen Shot 2020-04-06 at 10 30 42 AM

> I'm noticing that I get the pointer cursor when hovering anywhere over the label. Why is that?

It's likely to indicate that it is interactive - clicking on the label when it is tied to a particular form control element will cause the focus to shift into that form control element.

> Also not sure if this PR affects buttons in a form at all but, if it does, I also want to mention that the action group should be a primary button and a link button, instead of a primary button and a **secondary button as currently seen in the demo!

It does not, but if the [form grid layout](https://patternfly-pr-2766.surge.sh/documentation/core/demos/basicforms#grid) looks good, I will submit a new PR with that grid layout (and remove it from this one) and update our use of buttons in the form examples/demos.

mmenestr:
@mcoker Yes, that icon works too - no issues with using it! and it would be amazing if you could make those changes to the buttons - form grid layout looks good otherwise!

mcoker:
@mmenestr OK I've created a new issue to add the grid demo and update the action buttons - https://github.com/patternfly/patternfly/issues/2929

Also updated the help icon in this PR.

christiemolloy:
When the text wraps should it wrap like this where its off-center to the input?
Screen Shot 2020-04-08 at 2 33 15 PM

Should this happen when the label is too long? 
Screen Shot 2020-04-08 at 2 34 06 PM

Do we want the help button to be vertically centered with the text ? Looks like here its sitting with the baseline
Screen Shot 2020-04-08 at 2 35 49 PM

mcarrano:
@christiemolloy I think ideally everything would be vertically centered with the field. Do you agree @mceledonia ?

mcoker:
> When the text wraps should it wrap like this where its off-center to the input?

This is what bootstrap looks like. I think you should only use the horizontal form layout if you have short labels. 

Screen Shot 2020-04-08 at 6 09 42 PM

It could get tricky if we tried to center text/form elements. By default, it might look something like this:

Screen Shot 2020-04-08 at 6 06 53 PM

Though that's something we could offer as a modifier to a horizontal `pf-c-form__group` element if we wanted - it would center the label/control elements in the layout, and remove the top padding from the label text.

I reworked the help button to have padding and use negative margins. It looks like this, and doesn't change the height of the label in any way.

Screen Shot 2020-04-08 at 6 04 36 PM
Screen Shot 2020-04-08 at 6 04 31 PM

I tried using the plain button variation but the padding is too large. It would require the icon be too far from the text to account for its horizontal padding, and it would need to cause the space between broken lines of text increase so that the padding doesn't overlap text above the button.

Screen Shot 2020-04-08 at 5 51 20 PM
Screen Shot 2020-04-08 at 5 51 24 PM

mcarrano:
I don't feel strongly about the alignment of the labels when wrapping @mcoker . I'm fine to follow your lead on this. Agree that it's not good design practice to have long field labels if using a horizontal form. Also agree with your proposal for alignment and spacing of the help icon.

PR #2769, opened 2020-03-04
feat(base): made shield styles optional by default

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2591

This PR removes

```
  [class*="pf-c-"],
  [class*="pf-l-"] {
    &,
    &::before,
    &::after {
      box-sizing: border-box;
      padding: 0;
      margin: 0;
      background-color: transparent;
    }
  }
```

and this block that applies to all components and layouts (the top level component/layout selector, ie `.pf-c-data-toolbar`, `.pf-l-grid`, etc)

```
    font-family: var(--pf-global--FontFamily--sans-serif);
    font-size: var(--pf-global--FontSize--md);
    font-weight: var(--pf-global--FontWeight--normal);
    line-height: var(--pf-global--LineHeight--md);
    color: var(--pf-global--Color--100); // this color rule is set here to be able to use themes
    text-rendering: optimizeLegibility;
```

One thing to look out for is removing `background-color: transparent;` from `*`. Buttons have a background color by default, so I updated the button component so it works with this. I glanced at the rest of the components and they looked ok, but that's something to look out for.

Also the inheritable styles will no longer apply to a top-level component, and that may or may not be a good thing. Buttons, for example, should probably always reset their `color` so that in an element whose color is `red`, the button color doesn't inherit `red` as the color. All of our button variants now set the `color` but there may be other instances to update here.


src/patternfly/components/Button/button.scss
```diff
@@ -74,38 +74,58 @@
...
   --pf-c-button--m-link--focus--Color: var(--pf-global--link--Color--hover);
+  --pf-c-button--m-link--active--BackgroundColor: transparent;
   --pf-c-button--m-link--active--Color: var(--pf-global--link--Color--hover);
   --pf-c-button--m-link--disabled--BackgroundColor: transparent;
   --pf-c-button--m-link--m-inline--hover--TextDecoration: var(--pf-global--link--TextDecoration--hover);
   --pf-c-button--m-link--m-inline--hover--Color: var(--pf-global--link--Color--hover);
 
-
   // Plain btn
+  --pf-c-button--m-plain--BackgroundColor: transparent;
```
christiemolloy: do we want these as variables ? or should they just be set as `transparent`. Just seems like its adding a lot of variables when we don't want them to be overwritten

mcoker: I don't feel strongly about it, but I lean toward making the components easier to theme than less, unless it introduces a lot of overhead, which this doesn't. For example, someone could use something like a very light, semi-transparent background here in their theme.

Plus, properties like colors (text, background, border, etc) are almost always variables that can be overridden, so there is likely a developer expectation that this background could be overridden via a variable.

christiemolloy: Ok that sounds good to me!

PR #2773, opened 2020-03-04
don't merge feat(dropdown-avatar-and-text): added avatar and text to dropdown

### LINKED ISSUE #2482: Masthead - refine responsive behavior and avatar behavior
Need to refactor the masthead utility menu area to address the following issues:

* Want to better define how the utility menu should respond with the viewport width shrinks.
* Want to integrate the user menu and avatar to make the avatar interactive (when used).

New implementation must support the following utility items:
* Integrated user menu and avatar
* Notification icon - alerts the user of unread notifications and toggles notification drawer. The notification icon must be visible at all viewport sizes.
* utility menus - one or more items represented by an icon. Menus can be either Dropdowns or Options Menus. Typically includes Help and Settings but other menus may be included.

Designs can be found here: https://docs.google.com/document/d/1iN-FV-JsS1kial7wza7GG4Vu4jA0KfaXaxz5a0lrYcU/edit?usp=sharing

This replaces issue #1832 which will be closed

mattnolting:
relates to #2482

This PR does not close #2482, rather it  
- adds avatar to dropdown
- hides the avatar at a breakpoint (in page)
- adds supporting text in the dropdown

**Dropdown component preview**
https://patternfly-next-pr-2773.surge.sh/documentation/core/components/dropdown#dropdown-with-avatar-and-info

**Page demo preview**
https://patternfly-next-pr-2773.surge.sh/documentation/core/demos/page#default-nav

## Breaking changes
1. Any **md content** here
  - More content


src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -173,11 +182,36 @@ The dropdown panel is provided for flexibility in allowing various content withi
...
+    Ned Username
+  {{/dropdown-toggle-text}}
+  {{> dropdown-toggle-icon}}
+{{/dropdown}}
+```
+
+```hbs title=Dropdown-with-avatar-and-info-dark-theme
+<div class="pf-t-dark">
+{{#> dropdown id="dropdown-with-avatar-and-info-dark-theme-example" dropdown--IsAvatarInfo="true" dropdown--IsExpanded="true"}}
+  {{> avatar avatar--attribute='src="/assets/images/img_avatar.svg" alt="Avatar image"'}}
```
christiemolloy: I do think the image looks a little unbalanced here next to the text, even though if there was a photo it would be balanced. i think for a lot of users they would probably have this generic avatar, i wonder if design could add a border or a light background behind the person

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -389,6 +415,18 @@
     display: flex;
     align-items: center;
   }
+
+  &.pf-m-text {
```
christiemolloy: i wonder if you need pf-m-text or if you could select the classes that you added for the text inside

mattnolting: Good question. I put it here because the classes within (`pf-c-page__header-text`, `pf-c-page__header-label`) actually belong to `.pf-c-page` component and may not be present at all if a single string without a wrapper is used. `.pf-m-text` belongs to `pf-c-drowpdown` and `page__label` and `page__text` belong to page as they are unique.

It can be used like this as well
```
<div class="pf-c-dropdown__menu-item pf-m-text">
  string
</div>
```

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -438,3 +476,11 @@
   font-weight: var(--pf-c-dropdown__group-title--FontWeight);
   color: var(--pf-c-dropdown__group-title--Color);
 }
+
+// Dark
+.pf-t-dark {
+  .pf-c-dropdown__toggle {
+    --pf-c-dropdown__toggle--Color: var(--pf-t-dark--c-dropdown__toggle--Color);
```
mcoker: Where are these styles needed?

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -117,6 +130,19 @@
     margin-top: var(--pf-c-dropdown__c-divider--MarginTop);
     margin-bottom: var(--pf-c-dropdown__c-divider--MarginBottom);
   }
+
+  .pf-c-avatar {
```
mcoker: If this is just for positioning, can we create a `pf-c-dropdown__image` (or whatever) element that's agnostic to what is placed in it? Then other images/icons could be used here and these rules have less specificity and there's no dependency created between the dropdown and avatar.

src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -173,11 +182,36 @@ The dropdown panel is provided for flexibility in allowing various content withi
...
+  {{#> dropdown-toggle-text}}
+    Ned Username
+  {{/dropdown-toggle-text}}
+  {{> dropdown-toggle-icon}}
+{{/dropdown}}
+```
+
+```hbs title=Dropdown-with-avatar-and-info-dark-theme
+<div class="pf-t-dark">
```
mcoker: Oh I see why the `.pf-t-dark` styles exist now. I don't think we should add the classes to patternfly.css just to illustrate the workspace example. Can we just see the dark styles in the page demo? I'll also modify the parent element of workspace examples and add `.pf-t-dark` since we lost our magic light/dark theme toggle.

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -389,6 +409,18 @@
     display: flex;
     align-items: center;
   }
+
+  &.pf-m-text {
+    color: var(--pf-c-dropdown__menu-item--m-text--Color);
+    user-select: none;
```
mcoker: why is this set? I would expect to be able to select the text.

src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -173,11 +182,24 @@ The dropdown panel is provided for flexibility in allowing various content withi
 {{/dropdown}}
 ```
 
+```hbs title=Dropdown-with-avatar-and-info
```
mcoker: instead of info could we say "text" - to illustrate use of the text modifier

mcoker:
@mcarrano what do you think about adding the "text" modifier to a dropdown item that turns the item text grey and disables the hover/focus styles? It seems like this might be used in other contexts to include text in a dropdown menu. Or do you think this would this only ever be used in the context of the masthead user menu?

Also we assumed the "username: email", and "account number: #" styles (reducing the font size of username and account number) were specific to the masthead user menu, so those styles don't exist in the dropdown component and exist in the page component instead.

src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -31,6 +32,14 @@ import './Dropdown.css'
 {{/dropdown}}
 ```
 
+```hbs title=Disabled
```
mcoker: looks like this was added by accident.

src/patternfly/components/Dropdown/examples/Dropdown.md
```diff
@@ -173,11 +182,24 @@ The dropdown panel is provided for flexibility in allowing various content withi
 {{/dropdown}}
 ```
 
+```hbs title=Dropdown-with-avatar-and-info
+{{#> dropdown id="dropdown-with-avatar-and-info-example" dropdown--IsAvatarInfo="true" dropdown--IsExpanded="true"}}
```
mcoker: I wonder if this example should just show a basic usage of the text modifier on a dropdown item since it might be confusing and not obvious that the styles in this menu depend on the page component's css.

src/patternfly/components/Dropdown/dropdown-menu-avatar-info.hbs
```diff
@@ -0,0 +1,40 @@
+<div class="pf-c-dropdown__menu{{#if dropdown-menu--modifier}} {{dropdown-menu--modifier}}{{/if}}"
+  aria-labelledby="{{id}}-button"
+  {{#unless dropdown--IsExpanded}}hidden{{/unless}}
+  {{#if dropdown-menu--attribute}}
+    {{{dropdown-menu--attribute}}}
+  {{/if}}>
+
+  {{#> dropdown-group}}
+    <dl>
+      {{#> dropdown-menu-item dropdown-menu-item--attribute=(concat 'aria-hidden="true" id="' id '-username"') dropdown-menu-item--type="div" dropdown-menu-item--modifier="pf-m-text"}}
```
mcoker: why is this `aria-hidden`?

src/patternfly/components/Dropdown/examples/Dropdown.md
mcarrano: I do see what you mean @christiemolloy . But the reality is that this will only be used in the masthead where it will be against the dark background so I don't think that will be an issue.

mcarrano:
Good question about the text modifier @mcoker . Yes, I do think that might be generally useful. We do have this issue in the backlog: https://github.com/patternfly/patternfly-design/issues/770 Would doing that help us satisfy this other request, too?

mcoker:
> We do have this issue in the backlog: patternfly/patternfly-design#770 Would doing that help us satisfy this other request, too?

@mcarrano hmm, while they're visually similar, I'm inclined to say no because I think that is something different. The text variation of the dropdown item in this PR is the item itself, and the text in patternfly/patternfly-design#770 is a description for a dropdown item. I think we would want to keep them separate.

mcarrano:
Makes sense @mcoker

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,9 +1,69 @@
...
+  // end - update during breaking change
+
+  // After
+  --pf-c-divider--after--FlexBasis: 100%;
+  --pf-c-divider--after--Inset: 0%;
 
-  height: var(--pf-c-divider--Height);
-  background-color: var(--pf-c-divider--BackgroundColor);
+  // Vertical
+  --pf-c-divider--m-vertical--after--FlexBasis: 100%;
```
christiemolloy: should the divider changes be in here?

src/patternfly/components/Dropdown/dropdown.scss
christiemolloy: okay. could this modifier be called something like `pf-m-account-info` , text might be a bit arbitrary

mattnolting:
@christiemolloy 

> okay. could this modifier be called something like pf-m-account-info , text might be a bit arbitrary

I think generic is what we want, so it can be used in the future for other things. One use case will likely be in the multi-level dropdown / overflow menu where multiple dropdowns are combined. https://github.com/patternfly/patternfly-next/issues/2792

mattnolting:
closes #2802 

Isolating this issue to dropdown additions, therefore it is not a breaking change.

src/patternfly/components/Dropdown/dropdown-menu-toggle-image-info.hbs
```diff
@@ -0,0 +1,40 @@
...
+  {{#unless dropdown--IsExpanded}}hidden{{/unless}}
+  {{#if dropdown-menu--attribute}}
+    {{{dropdown-menu--attribute}}}
+  {{/if}}>
+
+  {{#> dropdown-group}}
+    <dl>
+      {{#> dropdown-menu-item dropdown-menu-item--type="div" dropdown-menu-item--modifier="pf-m-text"}}
+        <dt class="{{dropdown-menu--dt-modifier}}">Username:</dt>
+        <dd class="{{dropdown-menu--dd-modifier}}">mshaksho@redhat.com</dd>
```
christiemolloy: maybe replace to a generic email

src/patternfly/components/Page/examples/Page.md
```diff
@@ -156,6 +156,7 @@ This component provides the basic chrome for a page, including sidebar, header,
 | `.pf-c-page__header-nav` | `<div>` |   Creates a container to nest the navigation component in the header. |
 | `.pf-c-page__header-tools` | `<div>` |   Creates a container to nest the icons and menus in header. |
 | `.pf-c-page__header-tools-group` | `<div>` |  Creates a container for grouping sets of icons and menus in header. |
+| `.pf-c-page__header-label` | `<div>` |  Creates a header label. |
```
christiemolloy: if you keep these in, would you also need `pf-c-page__header-text`

PR #2776, opened 2020-03-04
fix(fileupload): change border for drag hover outside of psuedo selector

christiemolloy:
fixes `.pf-c-file-upload.p-m-drag-hover` so that the border is set on that class rather than the `::after` pseudo selector as that also has opacity of 0.1 set on it, which was affecting the opacity of the border.

### DIFF
--- a/src/patternfly/components/FileUpload/file-upload.scss
+++ b/src/patternfly/components/FileUpload/file-upload.scss
@@ -6,8 +6,8 @@
   --pf-c-file-upload--m-loading__file-details--before--Bottom: var(--pf-global--BorderWidth--sm);
 
   // pf-m-drag-hover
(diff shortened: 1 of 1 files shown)


src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -34,6 +34,7 @@
 
   &.pf-m-drag-hover {
     position: relative;
+    border: var(--pf-c-file-upload--m-drag-hover--BorderWidth) solid var(--pf-c-file-upload--m-drag-hover--BorderColor);
```
mcoker: adding a border is going to make the component shrink/grow when the class is applied. You could use the `::before` pseudo class to draw the border

mturley:
@christiemolloy while we're in here, in the "File upload error" example can we move the `aria-invalid="true"` from the `<input>` to the `<textarea>`?

PR #2777, opened 2020-03-04
fix(datalist): update compact font-size

mcoker:
part of https://github.com/patternfly/patternfly-next/issues/2653.

This just modifies the existing compact spacing and leaves the modifier and default behavior in place.

We're also not going to update the button alignment to center if the space below exceeds 8px. 

## Breaking changes
* Changes compact data-list font-size to `--pf-global--FontSize--sm`

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -129,6 +129,8 @@
   }
 
   // Compact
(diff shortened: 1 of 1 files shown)


christiemolloy:
Also agree that that example looks cramped. the 14pt text next to the 16pt text in the button is a little visually jarring to me, but if its approved by design then im fine with it

PR #2782, opened 2020-03-05
fix(notificationdrawer): updated dropdowns so they dont overflow pare…

mattnolting:
…nt (#2767)


redallen:
Hmm, something seems to be the matter with version numbers. This PR shouldn't trigger a release since it's a PR, but it's trying to since it sees that the branch it's running on is `master`. I'm unsure of what versions CI is actually seeing (they're all git tags) and trying to release.

Let me take a look and think about how to best do this.

redallen:
Ah, part of the reason is that master is currently broken: https://circleci.com/gh/patternfly/patternfly-next/4836

I'll get a PR up to fix that first, then we can try this again.

PR #2799, opened 2020-03-09
fix(radio, check): fixed radio left margin overflow, error in selector

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2677


src/patternfly/components/Radio/radio.scss
```diff
@@ -6,6 +6,7 @@
   --pf-c-radio__label--FontSize: var(--pf-global--FontSize--md);
   --pf-c-radio__label--LineHeight: var(--pf-global--LineHeight--sm);
   --pf-c-radio__input--MarginTop: #{pf-size-prem(-3px)};
+  --pf-c-radio__input--MarginLeft: #{pf-size-prem(1px)};
```
christiemolloy: since we don't usually add a margin to elements, should you leave a note here that this fixes the OSX bug ? so devs know what this margin left is for. If they don't see the issue I think they will be confused

christiemolloy:

PR #2807, opened 2020-03-10
fix(file-upload): removed message container, added form to error example

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2797

### DIFF
--- a/src/patternfly/components/FileUpload/examples/FileUpload.md
+++ b/src/patternfly/components/FileUpload/examples/FileUpload.md
@@ -91,7 +91,7 @@ beta: true
 ```hbs title=File-upload-in-form-with-error
 {{#> form}}
   {{#> form-group}}
-    {{#> file-upload file-upload--id="file-upload-error" file-upload--type="div"}}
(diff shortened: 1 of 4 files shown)


mcoker:
@mcarrano it just removes the file message/helper text at the bottom (instructions/error text) from the file upload component and relies on the file upload component being in a form for that helper text, so the helper text comes from the form component instead.

That also means:
* Any updates we make to the form component's helper text (descriptive, error, success, etc) will be available when used with file upload since the file upload is no longer creating its own helper text.
* The text doesn't get a blue overlay when you drag/hover the file upload component, since that text is no longer part of the file upload component.

old:
Screen Shot 2020-03-11 at 2 51 59 PM

new:
Screen Shot 2020-03-11 at 2 50 00 PM

mcarrano:
Sounds good to me. Thanks for the explanation @mcoker !

mturley:
Ah, I'm glad you guys made that change, I forgot to mention I needed to make `.pf-c-file-upload` a div instead of a form in the React component so I could get my FormGroup example to work (React will refuse to let you nest a <form> in another <form>).

PR #2814, opened 2020-03-12
feat(banner): adds banner component, sticky page main sections

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2727

@mcarrano I added a "banner" demo section with a simple demo of how to add a banner to be sticky to the top of the page content area.

The banner provides:
    * outer padding
    * background and text color
        * default is black background, white text. Other variations are available as modifiers.
    * font-size (`--pf-global--FontSize--sm`)
    * modifier to make it sticky to the top of the container it's used within
        * this adds a box-shadow (`--pf-global--BoxShadow--md-bottom`)

The demo just uses our flex layout to position the text inside of the banner. None of that is built in to the banner and the application would need to handle any positioning/hiding/showing of the banner text.

Is adding the banner component outside of the existing page elements (above the header and below the nav/content) something you think patternfly should provide, or would it be up to the application to add that to their page chrome, then place patternfly's page content between the top and bottom banners?


mcarrano:
@mcoker this is looking great. @maryshak1996 can you also take a look? The green looks a bit light when compared to the design doc. Is that really `success--color--100`?

If someone wanted to set a custom background color, how would they do that? What variable would they need to override?

Regarding your question about placing above/below the page, would it be possible to provide a second demo to show that placement? I don't this wee need more than that. Thanks.

maryshak1996:
Hey @mcarrano @mcoker this is looking good! to your point about the green color:
There's an inconsistency between the `success--color--100` as it's implemented and as it is in the sketch kit. I chatted with about this with @mceledonia and he said that there is an effort to change the success color to be the darker version that you see in my mocks.

mcoker:
@mcarrano fwiw, we do have this update scheduled for this milestone as a breaking change to update the success colors - maybe this design takes this update into account? https://github.com/patternfly/patternfly-next/issues/2479

> If someone wanted to set a custom background color, how would they do that? What variable would they need to override?

They can override any of

```
  --pf-c-banner--m-info--BackgroundColor
  --pf-c-banner--m-info--BackgroundColor
  --pf-c-banner--m-danger--BackgroundColor
  --pf-c-banner--m-success--BackgroundColor
  --pf-c-banner--m-warning--BackgroundColor
```

> would it be possible to provide a second demo to show that placement?

Currently the patternfly demos only showcase things that you can build with patternfly tools (components, layouts, utilitles). We don't have anything in patternfly that would enable placing something above the header and below the page, yet still visible in the browser viewport like that. It's easily achieved using custom CSS though. I created a demo with some custom CSS that shows how it could be done, but it doesn't follow the way our other demos are built since it is a custom layout and isn't built with just patternfly tools. https://patternfly-next-pr-2814.surge.sh/documentation/core/demos/banner/topbottom

Unless we want to either enhance the page component to accept something being shown above the header and fixed at the bottom of the viewport like this, or reconsider the way we build demos and allow for custom styles like this<sup>1</sup>, I wonder if this would be better highlighted in the design documentation, mentioning that a user would need to build this custom layout themselves.

This is also something we can continue to consider while going ahead and getting the component into patternfly without this demo.

[1] - I see value in allowing for custom demos, applying custom CSS, overriding variables, etc. We don't really have a place to showcase how you can do that sort of thing now, and I think our demos could be a good place to do that. It's just not something that we do currently.

src/patternfly/demos/Banner/__banner-page-main-section-content.hbs
```diff
@@ -0,0 +1,4 @@
+{{#> content}}
```
mattnolting: Same as above, I think you can drop `-content` in the partial name or use `-demo`...

src/patternfly/demos/Banner/__banner-content.hbs
```diff
@@ -0,0 +1,14 @@
+{{#> l-flex l-flex--modifier="pf-m-justify-content-center pf-m-justify-content-space-between-on-lg pf-m-nowrap"}}
```
mattnolting: I think you can drop `-content` in the partial name or use `-demo` instead to distinguish from potential component content.

mcarrano:
Good question about the demo @mcoker . Since this configuration is specifically for Gov't systems, it may be good enough to just share with them. But I do agree it would be nice to have a place for custom demos like this to share code between developers.

I am going to share a link to the PR with the Gov't Systems folks to get their feedback and will coordinate more discussion if needed.

redallen:
I'll try to rebase this on `master` momentarily.

redallen:
Rebased! Let me know if you have further issues.

src/patternfly/components/Banner/banner.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-banner{{#if banner--modifier}} {{banner--modifier}}{{/if}}"
```
christiemolloy, line 1: Not sure if this needs a role?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_role

mcarrano:
I agree with @christiemolloy that the text should probably truncate rather than overflow if the message is long. Also that the blue color looks off. @maryshak1996 can you verify?

And yes, I'm seeing the same problem with the extra Nav element stuck in the page body. Looks like a bug.

mcoker:
Thanks @mcarrano @christiemolloy, I need to go through the examples and update them based on changes introduced in the v4 release.

maryshak1996:
> Also that the blue color looks off. @maryshak1996 can you verify?
@mcarrano @christiemolloy the color to use here is: $pf-global--info-color--100 :)

mcoker:
@mcarrano @christiemolloy updated the nav in the demos.

Re: text truncation, it's tricky because the banner is really just a container for a user to put a custom layout inside of. We tell is to have `white-space: nowrap;`, which is an inheritable property - meaning it will apply to any child of the banner component, regardless what the user puts in it. However, the properties needed to make text truncate are not inheritable. Here are a few options:

* Don't worry about text truncation. Similar to the banner layout inside the component, leave it to the user to ensure there isn't too much content in the banner, it truncates, there is a popover, etc... This puts the responsibility on the user to enable text truncation if there is overflow content.
* Enable text truncation only if text is placed directly in the banner, meaning it's `<div class="pf-c-banner">this is my text</div>`. Then we can apply the text truncation in that case - however, if the user uses a custom layout in the banner, the text no longer truncates unless they enable that somehow. This makes it easy for text to truncate when there is just text in the banner (not sure how likely that use case is?), but still puts the responsibility on the user to truncate text themselves in more complicated structures in the banner.
* Require all text in the banner be wrapped in a `.pf-c-banner__text` element, and we enable text truncation on that element. So text would be wrapped in this element if it's the direct child of the banner component, as well as if the user has specified a 3 column layout - it's up to them that in all 3 columns, they wrap the text in each column in our `.pf-c-banner__text` element. This requires more work upfront on the user to ensure all of their banner text is wrapped in our text element and adds extra markup to the component, but would reliably and flexibly provide text truncation.


@maryshak1996 we are currently using `$pf-global--info-color--100` as `--pf-c-banner--m-info--BackgroundColor: var(--pf-global--info-color--100)`. FWIW, the inline info alert uses a background of `$pf-color-blue-50`, want to try that or is it too light?

You can see our global var color mapping here - https://github.com/patternfly/patternfly/blob/master/src/patternfly/sass-utilities/scss-variables.scss, with the actual color vars & hex values here - https://github.com/patternfly/patternfly/blob/master/src/patternfly/sass-utilities/colors.scss

mcarrano:
@mcoker regarding the text truncation, do you favor one of these approaches? I lean towards option #2 since that would satisfy most cases I think and for those who want to use a custom layout they will need to add the extra code. What do you think @christiemolloy ?

mcoker:
If the most common use case is just putting text in the banner, I'm in favor of #2, too. Then the expectation is that if the user does something custom with the layout in the banner, it's up the user to manage the layout and apply the appropriate text truncation if necessary. Does that sounds good @mcarrano?

mcarrano:
Sounds like a plan @mcoker !

mcoker:
@mcarrano updated. Also updated the info background color to use blue-200, which matches the mockups.

src/patternfly/components/Banner/banner.hbs
christiemolloy: Any thoughts? @mcoker

src/patternfly/demos/Banner/__banner-table.hbs
```diff
@@ -0,0 +1,183 @@
+{{#> table table--id=(concat page--id '-table') table--grid="true" table--modifier="pf-m-grid-xl" table--attribute='aria-label="This is a table with checkboxes"'}}
```
christiemolloy, line 1: This is what the demo looks like right now,  i dont know if the banner is working

Screen Shot 2020-06-17 at 9 34 44 AM

mcarrano:
@mcoker Looks like there is a problem with this demo https://patternfly-pr-2814.surge.sh/documentation/core/demos/banner/basic  The banner is not sticking correctly.

src/patternfly/components/Banner/banner.hbs
mcoker: @jessiehuff wdyt?

mcoker:
@mcarrano @christiemolloy good catch, updated.

mcarrano:
@mcoker the demo looks good now. @maryshak1996 can you just take a look and confirm the colors once again?

src/patternfly/components/Banner/banner.hbs
jessiehuff: A banner landmark role should be reserved for the site header content that is common across the site generally at the top of every page. So if these banners are expected to be the top banner then they should have the role, but otherwise they shouldn't. Looking through the [design doc](https://docs.google.com/document/d/1QZY9EcblZaIGFB6KHqeDtqWPHqxsJn5-VGJnowmrVT8/edit), it almost looks like this is used similar to an alert so I wonder if it should be treated similar with an aria-label. wdyt @jgiardino?

src/patternfly/demos/Banner/examples/Banner.md
```diff
@@ -0,0 +1,72 @@
...
+
+```hbs title=Top/bottom isFullscreen
+{{#> l-flex l-flex--modifier="pf-m-column pf-m-nowrap pf-m-space-items-none" l-flex--attribute='style="height: 100%;"'}}
+  {{#> l-flex-item}}
+    {{#> banner banner--modifier="pf-m-sticky"}}
+      {{> __banner-demo}}
+    {{/banner}}
+  {{/l-flex-item}}
+  {{#> l-flex-item l-flex-item--modifier="pf-m-grow" l-flex-item--attribute='style="min-height: 0;"'}}
+    {{#> page page--id="page-layout-table-simple"}}
```
mattnolting, line 44:
```suggestion
    {{#> page page--id="page-layout-table-top-bottom"}}
```

src/patternfly/demos/Banner/__banner-demo.hbs
```diff
@@ -0,0 +1,14 @@
+{{#> l-flex l-flex--modifier="pf-m-justify-content-center pf-m-justify-content-space-between-on-lg pf-m-nowrap"}}
+  {{#> display display--type="none pf-u-display-block-on-lg"}}
+    Localhost
+  {{/display}}
+  {{#> display display--type="none pf-u-display-block-on-lg"}}
+    This message is sticky to the top of the page.
```
mattnolting, line 6:
```suggestion
    This message is sticky to the top or bottom of the page.
```

src/patternfly/components/Banner/banner.hbs
jgiardino: The [documentation for `role="banner"`](https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_banner) is a bit confusing. After reading the following article, I'm inclined to say that we would not use `role="banner"` for this component:
https://www.scottohara.me/blog/2018/03/03/landmarks.html

For example, the documentation includes these bullets
> - Each page may have one banner landmark.
> - When a page contains nested document and/or application roles (e.g. typically through the use of iframe and frame elements), each document or application role may have one banner landmark.
> - If a page includes more than one banner landmark, each should have a unique label

Based on the explanation in the article, I interpret these points to mean that a page can only have more than one banner if additional banners are nested inside a `document` or `application`. 

> Looking through the design doc, it almost looks like this is used similar to an alert so I wonder if it should be treated similar with an aria-label.

We might want to consider not using aria-label on `<div>` elements that don't have an explicit role defined, like we currently do with the Alert component. The following article provides a really good explanation on that. But also, if visible text exists within an element, as in this case, the use of `aria-label` could override that text.
https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/

PR #2816, opened 2020-03-13
fix(components): change components to be mobile first

### LINKED ISSUE #1923: Investigate components using max-width

christiemolloy:
closes #1923 

## Breaking changes

This PR changes components that had `max-width` breakpoint declarations, and reverses these to be `min-width` so that they are written "mobile-first". Our aim is to make PatternFly "mobile-first", however there is one component that cannot satisfy this: the table. At mobile, we change the native/default properties of the table when we set it to `display: grid` in order to effectively style the content at that viewpoint. If we were to rewrite this entire component to be mobile first, given the stylistic properties, we would be writing **a lot** of code to overwrite the grid properties with table properties as well as making sure we add all roles to the table at desktop. For that reason, we have decided to keep the table "desktop-first".

The components that are now mobile-first are: background image, data list, data toolbar, form, login, page, toolbar, and wizard.

* `--pf-c-data-list__item-content--PaddingBottom` changes to `--pf-c-data-list__item-content--md--PaddingBottom`
* `--pf-c-data-list__item-action--not-last-child--lg--MarginBottom` changes to `--pf-c-data-list__item-action--not-last-child--MarginBottom`
* `--pf-c-data-list__expandable-content--md--MaxHeight` changes to `--pf-c-data-list__expandable-content--MaxHeight`
* `--pf-c-data-list__item-content--PaddingBottom` changes to `--pf-c-data-list__item-content--md--PaddingBottom`
* `--pf-c-login__header--sm--PaddingLeft` changes to `--pf-c-login__header--PaddingLeft`
* `--pf-c-login__header--sm--PaddingRight` changes to `--pf-c-login__header--PaddingRight`
* `--pf-c-login__main--xl--MarginBottom` changes to `--pf-c-login__main--MarginBottom`
* `--pf-c-login__footer--sm--PaddingLeft` changes to `--pf-c-login__footer--PaddingLeft`
* `--pf-c-login__footer--sm--PaddingRight` changes to `--pf-c-login__footer--PaddingRight`
* Removes `--pf-c-page__main-section--m-no-padding-mobile--md--PaddingTop`
* Removes `--pf-c-page__main-section--m-no-padding-mobile--md--PaddingRight`
* Removes `--pf-c-page__main-section--m-no-padding-mobile--md--PaddingBottom`
* Removes `--pf-c-page__main-section--m-no-padding-mobile--md--PaddingLeft`

* Removes `.pf-m-no-padding-mobile` and introduces `.pf-m-no-padding{-on-[breakpoint]}` which removes padding from the main page section at an optional breakpoint.
* Introduces `.pf-m-padding{-on-[breakpoint]}` which modifies the main page section to add padding back in at a specified breakpoint. Should be used with pf-m-no-padding.

* `$pf-global--gutter:` now has the value  `$pf-global--spacer--md` and `$pf-global--gutter--md` now has the value `$pf-global--spacer--lg`


src/patternfly/components/Page/page.scss
```diff
@@ -482,12 +480,13 @@
...
-      --pf-c-page__main-section--PaddingTop: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingTop);
-      --pf-c-page__main-section--PaddingRight: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingRight);
-      --pf-c-page__main-section--PaddingBottom: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingBottom);
-      --pf-c-page__main-section--PaddingLeft: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingLeft);
+  @each $breakpoint, $breakpoint-value in $pf-c-page--breakpoint-map {
+    $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
+
+    @include pf-apply-breakpoint($breakpoint, $pf-c-page--breakpoint-map) {
+      &.pf-m-no-padding-#{$breakpoint-name} {
+        padding: 0;
```
christiemolloy: @mattnolting could you help me debug why this isn't working

src/patternfly/components/Page/page.scss
```diff
@@ -482,12 +480,13 @@
...
-    @media screen and (max-width: $pf-global--breakpoint--md) {
-      --pf-c-page__main-section--PaddingTop: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingTop);
-      --pf-c-page__main-section--PaddingRight: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingRight);
-      --pf-c-page__main-section--PaddingBottom: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingBottom);
-      --pf-c-page__main-section--PaddingLeft: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingLeft);
+  @each $breakpoint, $breakpoint-value in $pf-c-page--breakpoint-map {
+    $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
+
+    @include pf-apply-breakpoint($breakpoint, $pf-c-page--breakpoint-map) {
+      &.pf-m-no-padding-#{$breakpoint-name} {
```
mattnolting:
```suggestion
      &.pf-m-no-padding#{$breakpoint-name} {
```

src/patternfly/components/Page/examples/Page.md
```diff
@@ -174,6 +174,6 @@ This component provides the basic chrome for a page, including sidebar, header,
 | `.pf-m-dark-200` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a dark transparent background. |
 | `.pf-m-dark-100` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a darker transparent background. |
 | `.pf-m-no-padding` | `.pf-c-page__main-section` | Modifies a main page section to have no padding. |
-| `.pf-m-no-padding-mobile` | `.pf-c-page__main-section` | Modifies a main page section to have no padding on mobile. |
+| `.pf-m-no-padding-{-on-[breakpoint]}` | `.pf-c-page__main-section` | Modifies a main page section to have no padding at a given breakpoint. |
```
mcoker: I think you could consolidate `.pf-m-no-padding` and this entry to `.pf-m-no-padding{-on[breakpoint]}` to say it removes padding at an optional breakpoint.

mcoker:
@christiemolloy what do you think about reversing these? https://github.com/patternfly/patternfly-next/blob/438890b5f67cb795aef3e01926f5e5d9208bd5b8/src/patternfly/sass-utilities/scss-variables.scss#L141-L143

src/patternfly/components/DataList/data-list.scss
```diff
@@ -384,8 +384,14 @@
 
 // Expandable content
 .pf-c-data-list__expandable-content {
+  max-height: var(--pf-c-data-list__expandable-content--MaxHeight);
+  overflow-y: auto;
   box-shadow: var(--pf-c-data-list__expandable-content--BoxShadow);
 
+  @media screen and (min-width: $pf-global--breakpoint--md) {
+    max-height: initial;
```
mcoker: Can you set `overflow: visible;` here? Otherwise it will hide overflow by default, resulting in something like this (an expanded dropdown in expandable content) 

Screen Shot 2020-03-17 at 6 19 34 PM

src/patternfly/components/Page/page.scss
```diff
@@ -482,12 +480,21 @@
     padding: 0;
```
mcoker: You shouldn't need this

src/patternfly/components/Page/page.scss
```diff
@@ -482,12 +480,21 @@
...
-      --pf-c-page__main-section--PaddingTop: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingTop);
-      --pf-c-page__main-section--PaddingRight: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingRight);
-      --pf-c-page__main-section--PaddingBottom: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingBottom);
-      --pf-c-page__main-section--PaddingLeft: var(--pf-c-page__main-section--m-no-padding-mobile--md--PaddingLeft);
+  &.pf-m-no-padding-on-sm,
+  &.pf-m-no-padding-on-md,
+  &.pf-m-no-padding-on-lg,
+  &.pf-m-no-padding-on-xl,
+  &.pf-m-no-padding-on-2xl {
+    padding: 0;
```
mcoker: you shouldn't need this block

src/patternfly/components/Page/page.scss
```diff
@@ -482,12 +480,21 @@
...
+  }
+
+  @each $breakpoint, $breakpoint-value in $pf-c-page--breakpoint-map {
+    $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
+
+    @include pf-apply-breakpoint($breakpoint, $pf-c-page--breakpoint-map) {
+      &.pf-m-no-padding#{$breakpoint-name} {
+        padding: var(--pf-c-page__main-section--PaddingTop) var(--pf-c-page__main-section--PaddingRight) var(--pf-c-page__main-section--PaddingBottom) var(--pf-c-page__main-section--PaddingLeft);
```
mcoker:
```suggestion
                --pf-c-page__main-section--PaddingTop: 0;
                --pf-c-page__main-section--PaddingRight: 0;
                --pf-c-page__main-section--PaddingBottom: 0;
                --pf-c-page__main-section--PaddingLeft: 0;
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -88,19 +88,26 @@
...
+  }
+
+  @media screen and (min-width: $pf-global--breakpoint--lg) {
+    display: block;
+    visibility: visible;
 
     &.pf-m-expanded {
-      display: flex;
-      visibility: visible;
+      display: none;
```
mcoker: I think you can just get rid of this block entirely. The filter should be shown on desktop, and you wouldn't want to hide it when you apply `.pf-m-expanded`

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -393,10 +393,15 @@
...
+    display: block;
+    visibility: visible;
+  }
+
+  @media screen and (min-width: $pf-global--breakpoint--lg) {
     &.pf-m-expanded {
-      display: block;
-      visibility: visible;
+      display: none;
+      visibility: hidden;
```
mcoker: I don't think you need this block.

src/patternfly/components/Page/examples/Page.md
```diff
@@ -123,8 +123,8 @@ cssPrefix: pf-c-page
     {{#> page-main-section page-main-section--modifier="pf-m-no-padding pf-m-light"}}
       This `.pf-c-page__main-section` uses `.pf-m-no-padding` to remove all padding.
     {{/page-main-section}}
-    {{#> page-main-section page-main-section--modifier="pf-m-no-padding-mobile"}}
-      This `.pf-c-page__main-section` uses `.pf-m-no-padding-mobile` to remove padding on mobile only.
+    {{#> page-main-section page-main-section--modifier="pf-m-no-padding-on-md"}}
```
mcoker: I think this is going to need to be:

```suggestion
    {{#> page-main-section page-main-section--modifier="pf-m-no-padding pf-m-padding-on-md"}}
```

I wonder if we should keep `no-padding-mobile` for places that already have it like this, and support removing/adding padding at the breakpoints? And `no-padding-mobile` can always represent removing the padding at whatever breakpoint we change the page to a mobile view.

src/patternfly/components/DataList/data-list.scss
christiemolloy: good catch, thanks for finding that

src/patternfly/components/Page/page.scss
christiemolloy: were you able to test this? because`pf-apply-breakpoint` is set to min-width, it removes the padding at `md` and above, when we want the reverse. That's why I had removed the padding from the selectors and then added it back in at the breakpoint that is specified. but maybe im missing something! let me know what you think

src/patternfly/components/Toolbar/toolbar.scss
christiemolloy: yep good point, removed

src/patternfly/components/Page/page.scss
mcoker: Oh I see... personally I think that's confusing, since all of our other breakpoint modifiers apply the modifier name at the specified breakpoint and above, with exceptions like the `pf-m-grid` modifiers on table. wdyt @mattnolting

src/patternfly/components/Login/login.scss
```diff
@@ -136,10 +136,12 @@
   @include pf-t-dark;
 
   grid-area: header;
+  padding-right: var(--pf-c-login__header--PaddingRight);
+  padding-left: var(--pf-c-login__header--PaddingLeft);
 
-  @media (max-width: $pf-global--breakpoint--sm) {
-    padding-right: var(--pf-c-login__header--sm--PaddingRight);
-    padding-left: var(--pf-c-login__header--sm--PaddingLeft);
+  @media (min-width: $pf-global--breakpoint--sm) {
```
mcoker: Could you move this block up to under the block where these vars are defined?

src/patternfly/components/Login/login.scss
```diff
@@ -165,31 +168,30 @@
     padding-bottom: var(--pf-c-login__main-footer--PaddingBottom);
   }
 
-  // This is not mobile first but avoids removing the margin for larger devices
-  @media (max-width: $pf-global--breakpoint--xl) {
-    margin-bottom: var(--pf-c-login__main--xl--MarginBottom);
+  @media (min-width: $pf-global--breakpoint--xl) {
```
mcoker: Same for this - can you move it up? Doesn't need to be nested here.

src/patternfly/components/Login/login.scss
```diff
@@ -274,10 +276,12 @@
   @include pf-t-dark;
 
   grid-area: footer;
+  padding-right: var(--pf-c-login__footer--PaddingRight);
+  padding-left: var(--pf-c-login__footer--PaddingLeft);
 
-  @media (max-width: $pf-global--breakpoint--sm) {
-    padding-right: var(--pf-c-login__footer--sm--PaddingRight);
-    padding-left: var(--pf-c-login__footer--sm--PaddingLeft);
+  @media (min-width: $pf-global--breakpoint--sm) {
```
mcoker: same for this - can we move it up to the vars section?

src/patternfly/components/Page/examples/Page.md
christiemolloy: per offline conversation on slack, we decided to remove `no-padding-mobile` and use `pf-n-no-padding pf-m-padding-on-md` and then this shouldnt have any change to how its used in React.

src/patternfly/components/Page/examples/Page.md
```diff
@@ -173,7 +173,7 @@ This component provides the basic chrome for a page, including sidebar, header,
 | `.pf-m-light` | `.pf-c-page__main-section` | Modifies a main page section to have a light theme. |
 | `.pf-m-dark-200` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a dark transparent background. |
 | `.pf-m-dark-100` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a darker transparent background. |
-| `.pf-m-no-padding` | `.pf-c-page__main-section` | Modifies a main page section to have no padding. |
-| `.pf-m-no-padding-mobile` | `.pf-c-page__main-section` | Modifies a main page section to have no padding on mobile. |
+| `.pf-m-no-padding`, `.pf-m-no-padding-{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional breakpoint |
```
mcoker:
```suggestion
| `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional breakpoint. |
```

mcoker:
Hmm... I'm having a hard time tracking down why I'm seeing style in the PR preview:

```
.pf-c-page__main-section.pf-m-no-padding,.pf-c-page__main-section.pf-m-no-padding-on-2xl,.pf-c-page__main-section.pf-m-no-padding-on-lg,.pf-c-page__main-section.pf-m-no-padding-on-md,.pf-c-page__main-section.pf-m-no-padding-on-sm,.pf-c-page__main-section.pf-m-no-padding-on-xl {
    padding: 0
}
```

I'm looking in your PR and don't see code that would generate that line. Also when I compile the CSS locally either using `npm run build-patternfly` or just `sass patternfly.scss`, I don't see those styles. @christiemolloy any idea what that might be?

The reason I mention that is, in all of the demos where we had `.pf-m-no-padding-mobile`, this PR changes it to `.pf-m-no-padding-on-md`, but that will still leave the padding on mobile since `.pf-c-page__main-section` has padding by default. We'll either need to add `.pf-m-no-padding` to the class list on those main sections we want to remove the mobile padding from, or leave `.pf-m-no-padding-mobile` and use that instead.

src/patternfly/components/Page/page.scss
mcoker: You still shouldn't need this. It gets built using the breakpoint loop below with "base" as the value in the map. The compiled CSS currently looks like:

```
.pf-c-page__main-section.pf-m-no-padding {
  padding: 0;
}

.pf-c-page__main-section.pf-m-padding {
  padding: var(--pf-c-page__main-section--PaddingTop) var(--pf-c-page__main-section--PaddingRight) var(--pf-c-page__main-section--PaddingBottom) var(--pf-c-page__main-section--PaddingLeft);
}

.pf-c-page__main-section.pf-m-no-padding {
  padding: 0;
}

@media (min-width: 576px) {
  .pf-c-page__main-section.pf-m-padding-on-sm {
    padding: var(--pf-c-page__main-section--PaddingTop) var(--pf-c-page__main-section--PaddingRight) var(--pf-c-page__main-section--PaddingBottom) var(--pf-c-page__main-section--PaddingLeft);
  }

  .pf-c-page__main-section.pf-m-no-padding-on-sm {
    padding: 0;
  }
}
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -46,18 +46,18 @@
...
+  --pf-c-data-list__item-row--md--PaddingLeft: var(--pf-global--spacer--lg);
 
-  @media screen and (max-width: $pf-global--breakpoint--md) {
+  @media screen and (min-width: $pf-global--breakpoint--md) {
     --pf-c-data-list__item-row--PaddingRight: var(--pf-c-data-list__item-row--md--PaddingRight);
     --pf-c-data-list__item-row--PaddingLeft: var(--pf-c-data-list__item-row--md--PaddingLeft);
   }
 
   // Data list item content
-  --pf-c-data-list__item-content--PaddingBottom: var(--pf-global--spacer--lg);
```
mcoker: You still have a reference to this var here that needs to change to `--pf-c-data-list__item-content--md--PaddingBottom`

https://github.com/patternfly/patternfly/blob/2c4b42c08dfec4be8a623678bf54297c9b39b19b/src/patternfly/components/DataList/data-list.scss#L168

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -393,11 +393,9 @@
...
   box-shadow: var(--pf-c-wizard__nav--BoxShadow);
 
-  @media screen and (max-width: $pf-global--breakpoint--lg) {
-    &.pf-m-expanded {
-      display: block;
-      visibility: visible;
-    }
+  &.pf-m-expanded {
+    display: block;
+    visibility: visible;
```
mcoker: This looks good, but I'm seeing remnants of previous commits in the preview CSS. For example, the nav isn't visible in this demo https://patternfly-next-pr-2816.surge.sh/documentation/core/components/wizard/nav-expanded-mobile

It's due to this block of CSS being in the compiled CSS on the preview site, but not in our SCSS from as far as I can tell. I think it _was_ in our SCSS, but has since been removed?

```
@media screen and (min-width: 992px) {
    .pf-c-wizard__nav.pf-m-expanded {
        display:none;
        visibility: hidden
    }
}
```

@redallen can you check this out? I tried re-running the build and still seeing this. I left a similar comment here - https://github.com/patternfly/patternfly/pull/2816#issuecomment-603531287

src/patternfly/demos/Table/examples/Table.md
```diff
@@ -283,7 +283,7 @@ section: demos
     {{#> page-main-section page-main-section--modifier="pf-m-light"}}
       {{> table-main-section-content}}
     {{/page-main-section}}
-    {{#> page-main-section page-main-section--modifier="pf-m-no-padding-mobile"}}
+    {{#> page-main-section page-main-section--modifier="pf-m-no-padding-on-md"}}
```
mcoker: Anywhere we were using `pf-m-no-padding-mobile` now needs to be `pf-m-no-padding pf-m-padding-on-md`. Otherwise you're adding it on mobile (since the page section has padding by default), and removing it on desktop with `pf-m-no-padding-on-md`. Here's an example from the build in this PR:

Screen Shot 2020-03-24 at 8 11 00 PM
Screen Shot 2020-03-24 at 8 10 56 PM

src/patternfly/components/DataList/data-list.scss
christiemolloy: good catch thank you!

src/patternfly/demos/Table/examples/Table.md
christiemolloy: Thanks forgot to update those :(

src/patternfly/components/Page/examples/Page.md
```diff
@@ -173,7 +173,7 @@ This component provides the basic chrome for a page, including sidebar, header,
...
+| `.pf-m-padding-{-on-[breakpoint]}` | `.pf-c-page__main-section` | Modifies the main page section to have padding at an optional breakpoint. |
```
mcoker: The SCSS creates `.pf-m-padding`, though it won't currently be used since it has padding by default. Probably best if we indicate it's there in the selector with this change. Also since the docs say "at an optional breakpoint," meaning the breakpoint isn't a required part of the selector.

```suggestion
| `.pf-m-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Modifies the main page section to have padding at an optional breakpoint. |
```

PR #2818, opened 2020-03-13
fix(color): updated black global and background global/component vars

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2424

## Breaking changes
* `$pf-color-black-200` changes to `#f0f0f0`
* `$pf-color-black-600` changes to `#6a6e73`
* `$pf-global--BackgroundColor--200` changes to `$pf-color-black-200 `
* Removes `$pf-global--BackgroundColor--150` and `$pf-global--BackgroundColor--300`
  * Updated all components that use `$pf-global--BackgroundColor--150` or `$pf-global--BackgroundColor--300` to use  `$pf-global--BackgroundColor--200` instead

### DIFF
--- a/build/gulp/ws-lite.css
+++ b/build/gulp/ws-lite.css
@@ -1,6 +1,6 @@
 .ws-lite-index-example {
   padding: 3rem 3rem 5rem 3rem;
-  background-color: var(--pf-global--BackgroundColor--300);
+  background-color: var(--pf-global--BackgroundColor--200);
 }
 
 .ws-lite-index-example > * {
@@ -64,7 +64,7 @@
 }
 
 :not(.ws-lite-example) table code {
-  background: var(--pf-global--BackgroundColor--300);
+  background: var(--pf-global--BackgroundColor--200);
 }
 
 :not(.ws-lite-example) code {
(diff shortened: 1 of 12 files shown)


mattnolting:
We should update:
- https://github.com/patternfly/patternfly-next/blob/master/build/gulp/ws-lite.css#L3
- https://github.com/patternfly/patternfly-next/blob/master/build/gulp/ws-lite.css#L67

to `--pf-global--BackgroundColor--200` as well

mattnolting:
Andddd, updated tabs.scss here: https://github.com/patternfly/patternfly-next/pull/2757/files#diff-c8d219328bcd3d467c0732874743859fR32

mcoker:
@mattnolting nice catch! updated.

mceledonia:
Nice! This looks great @mcoker

christiemolloy:
For the review notes, when you say "most likely" maybe it could be more encouraging to say update to use... "can be updated to use BackgroundColor--200, but consult your visual designer first" ... just an idea

> Any references to the variables that were removed should be reviewed, and most likely updated to use --pf-global--BackgroundColor--200/$pf-global--BackgroundColor--200 instead, to align with patternfly's usage of BackgroundColor variables.

mcoker:
@christiemolloy updated

PR #2821, opened 2020-03-16
fix(page-header-tools-margin): removed margin-top/bottom

### LINKED ISSUE #2040: Remove margins from page header tools
cc: @mattnolting I don't think these vars are necessary in the page header tools anymore, because when removing them they don't affect layout:

```
  margin-top: var(--pf-c-page__header-tools--MarginTop);
  margin-bottom: var(--pf-c-page__header-tools--MarginBottom);
```

mattnolting:
closes #2040 

## Breaking changes
1. Removed `--pf-c-page__header-tools--MarginTop: var(--pf-global--spacer--sm);`
2. Removed `--pf-c-page__header-tools--MarginBottom: var(--pf-global--spacer--sm);`
3. Removed `margin-top`
4. Removed `margin-bottom`

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -58,9 +58,7 @@
   }
 
   // Header tools
-  --pf-c-page__header-tools--MarginTop: var(--pf-global--spacer--sm);
   --pf-c-page__header-tools--MarginRight: var(--pf-global--spacer--md);
-  --pf-c-page__header-tools--MarginBottom: var(--pf-global--spacer--sm);
   --pf-c-page__header-tools--md--MarginRight: var(--pf-global--spacer--lg);
   --pf-c-page__header-tools--c-avatar--MarginLeft: var(--pf-global--spacer--md);
   --pf-c-page__header-tools-group--MarginLeft: var(--pf-global--spacer--xl);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -279,9 +279,7 @@
 .pf-c-page__header-tools {
   grid-column: 2 / 3;
   align-items: center;
-  margin-top: var(--pf-c-page__header-tools--MarginTop);
```
mcoker: This is still defined at https://github.com/patternfly/patternfly-next/blob/829b5f6b40d992988b78482ab11aa580a326e684/src/patternfly/components/Page/page.scss#L61

src/patternfly/components/Page/page.scss
```diff
@@ -279,9 +279,7 @@
...
   margin-right: var(--pf-c-page__header-tools--MarginRight);
-  margin-bottom: var(--pf-c-page__header-tools--MarginBottom);
```
mcoker: This is defined at https://github.com/patternfly/patternfly-next/blob/829b5f6b40d992988b78482ab11aa580a326e684/src/patternfly/components/Page/page.scss#L63

mattnolting:
@mcoker updated

mcoker:
@mattnolting nice! can you also update the issue description to say removed `--pf-c-page__header-tools--MarginTop` and `--pf-c-page__header-tools--MarginBottom`, since those are what users will need to update?

mattnolting:
@mcoker updated

mcoker:
Can you say something to the effect of "The following variables were removed. Any reference to them should be removed as they are no longer used in patternfly"?

PR #2823, opened 2020-03-16
fix(chipgroup): modify max width for chip label

### LINKED ISSUE #2765: Datatoolbar: wider chip labels
Migrating Cost Management to use the new data toolbar and noticed an issue with the chip label width. 

For the "Tag" category below, the chips represent tags (tag, key, value pairs), which can be lengthy. (Note that "Project" is a different category.) The current chip implementation truncates the chip label so that it makes it difficult to read.

Screen Shot 2020-01-23 at 2 06 25 PM

When the "Tag Key" label is omitted, it's not clear what the chip represents below. I feel that the label is important here, so ideally we could use a little more width.

Screen Shot 2020-01-23 at 2 23 59 PM

christiemolloy:
fixes #2765 

## Breaking changes
1. Changed max-width for `--pf-c-chip-group__label--Maxwidth` to 160px


mattnolting:
👍 Nice!

Issue #1223, opened 2019-01-14
Gatsby - workspace should extract required css into separate css file


dgutride:
We'd like to remove the inline styling that is being used right now to work around default gatsby behavior.

- Remove the custom sass plugin added for gatsby v2 upgrade
- Use the normal mechanism in gatsby-node/gatsby-config to overwrite the bundling configuration for minicssextract and extract-text behavior

michaelkro:
Could possibly be a related issue https://github.com/gatsbyjs/gatsby/issues/9733

dgutride:
Closing via https://github.com/patternfly/patternfly-next/pull/1232

PR #2824, opened 2020-03-16
fix(colors): remove cyan vars

### LINKED ISSUE #2115: Remove cyan color vars
These are no longer used

```
$pf-color-cyan-600:              #003d44 !default;
$pf-color-cyan-700:              #001f22 !default;
```

christiemolloy:
closes #2115 

## Breaking changes
1. Remove cyan color vars: pf-color-cyan-600 and pf-color-cyan-700

### DIFF
--- a/src/patternfly/_variables.scss
+++ b/src/patternfly/_variables.scss
@@ -26,8 +26,6 @@
   --pf-global--palette--cyan-300: #{$pf-color-cyan-300};
   --pf-global--palette--cyan-400: #{$pf-color-cyan-400};
   --pf-global--palette--cyan-500: #{$pf-color-cyan-500};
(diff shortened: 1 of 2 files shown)


mceledonia:
@christiemolloy @mcoker Removing cyan-600/700 would leave it as the only color without 600/700 variables which feels weird. Coker and I realized we initially had this story because we replaced the cyan-100-500 with the chart variables. Can we pivot this to just change 600/700 to colors which match the new palette more instead of removing them?

mceledonia:
We also need to do this for the blues as well! 

`green-600` = `#002323`
`green-700` = `#000F0F`

`blue-600` = `#002952`
`blue-700` = `#001223`

Here's a quick before/after, other colors don't seem to have this issue just blue/cyan.

image

PR #2828, opened 2020-03-16
fix(button): add modifiers to position icon in button

### LINKED ISSUE #2574: Follow-up issue for button: ability to add icon to left or right side
Related to #2507 make the class `pf-c-button__text` mandatory when using text inside of the button component, so we can style the direct descendants of the button with a margin since the content a user passes to the button would be in the __text element.

christiemolloy:
closes #2574 

### Breaking changes
* Adds the modifiers `pf-m-start` and `pf-m-end` to position `pf-c-button__icon` in the button component depending on whether it comes before or after text. We made this change so that we could safely position the icon before or after text, without needing to introduce a new class that wraps the text.
* Removes `  --pf-c-button__icon--MarginRight` and `--pf-c-button__text--icon--MarginLeft`. These are no longer in use, and to modify the right margin on an icon at the beginning of the button, use `--pf-c-button__icon--m-start--MarginRight` now and `--pf-c-button__icon--m-end--MarginLeft` to modify the left margin of an icon at the end of a button.
* Removes the `.pf-c-button__text` element as it is no longer necessary. Any instance of this element or styling using this selector should be removed as it is no longer used in the button component.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -124,16 +124,6 @@
   border: 0;
   border-radius: var(--pf-c-button--BorderRadius);
 
-  .pf-c-button__icon {
-    &.pf-m-start {
-      margin-right: var(--pf-c-button__icon--m-start--MarginRight);
-    }
-
-    &.pf-m-end {
-      margin-left: var(--pf-c-button__icon--m-end--MarginLeft);
-    }
-  }
-
   &::after {
     position: absolute;
     top: 0;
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -125,12 +125,13 @@
   border-radius: var(--pf-c-button--BorderRadius);
 
   .pf-c-button__icon {
```
mcoker: can you move this selector block outside of `.pf-c-button`? doesn't seem like it needs to be nested

PR #2829, opened 2020-03-16
feat(example): add select-disabled-typeahead-example.hbs

ValerieKenyon:


src/patternfly/components/Select/examples/Select.md
```diff
@@ -100,6 +100,12 @@ The single select should be used when the user is selecting an option from a lis
 {{#> select select-toggle--type="div" id="select-single-typeahead-expanded-selected" select--ItemIsSelected="true" select--IsExpanded="true" select--IsTypeahead="true" select-toggle--type="div" select-typeahead--Placeholder="New Mexico"}}
 {{/select}}
 ```
+
+```hbs title=Disabled-with-typeahead
+{{#> select select-toggle--type="div" id="select-single-typeahead-disabled" select-toggle--attribute="disabled" select--IsTypeahead="true" select-typeahead--Placeholder="Choose a state"}}
```
mcoker: When `.pf-c-select__toggle` is a `<div>` (instead of a `<button>`, which it normally is), instead of using the `disabled` attribute, you'll need to use the class `.pf-m-disabled`, since `disabled` isn't valid on a `<div>`. You can reference that in the select component's documentation by searching for "pf-m-disabled" either in your dev environment or on https://www.patternfly.org/v4/documentation/core/components/select. 

In that documentation, you'll also see

> When [.pf-m-disabled] is used, `disabled` should also be added to any form elements in `div.pf-c-select__toggle`

That means you'll need to add `disabled` to the text `<input>` and `<button>` in the toggle. 

Here's the code for the `<input>` - https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Select/select-toggle-typeahead.hbs

And for the `<button>` https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Select/select-toggle-button.hbs

One approach could be to add `select--IsDisabled="true"` in this disabled code you've added here, then in both [select-toggle-typeahead.hbs](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Select/select-toggle-typeahead.hbs) and [select-toggle-button.hbs](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Select/select-toggle-button.hbs), add conditionals to test if `select--IsDisabled` exists, and if so add `disabled` to those elements.

 Feel free to try that or any other method you think will work well! It's also worth noting you don't need to worry about adding `disabled` to [select-toggle-button.hbs](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Select/select-toggle-button.hbs) if the select is expanded, since a disabled select is not expandable.

src/patternfly/components/Select/examples/Select.md
```diff
@@ -100,6 +100,17 @@ The single select should be used when the user is selecting an option from a lis
 {{#> select select-toggle--type="div" id="select-single-typeahead-expanded-selected" select--ItemIsSelected="true" select--IsExpanded="true" select--IsTypeahead="true" select-toggle--type="div" select-typeahead--Placeholder="New Mexico"}}
 {{/select}}
 ```
+
+```hbs title=Disabled-with-typeahead
+{{#> select select-toggle--type="div"
+id="select-single-typeahead-disabled"
+select--IsTypeahead="true"
+select--IsDisabled="true"
+select-toggle--modifier="pf-m-disabled"
```
mcoker: 👍

src/patternfly/components/Select/select-toggle-button.hbs
```diff
@@ -3,7 +3,14 @@
     <i class="fas fa-caret-down pf-c-select__toggle-arrow" aria-hidden="true"></i>
   {{/button}}
 {{else}}
-  {{#> button button--modifier="pf-m-plain pf-c-select__toggle-button" button--attribute=(concat 'id="' id '-toggle" aria-haspopup="true" aria-expanded="false" aria-labelledby="' id '-label ' id '-toggle" aria-label="Select"')}}
-    <i class="fas fa-caret-down pf-c-select__toggle-arrow" aria-hidden="true"></i>
-  {{/button}}
+  {{#if select--IsDisabled}}
+    {{#> button button--modifier="pf-m-plain pf-c-select__toggle-button pf-m-disabled"
```
mcoker: Since this renders as a `<button>` element, we want to apply the `disabled` attribute to the element, not the `.pf-m-disabled` class. We use `disabled` on form elements that support it. We use `.pf-m-disabled` on non-form elements, like `<a class="pf-c-button pf-m-disabled">a link that looks like a button</a>`, since an `<a>` doesn't support the `disabled` attribute.

src/patternfly/components/Select/select-toggle-typeahead.hbs
```diff
@@ -5,9 +5,17 @@
...
 {{else}}
-  {{#> form-control controlType="input"
-    input="true"
-    form-control--modifier="pf-c-select__toggle-typeahead"
-    form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
-  {{/form-control}}
+  {{#if select--IsDisabled}}
+    {{#> form-control controlType="input"
+      input="true"
+      form-control--modifier="pf-c-select__toggle-typeahead pf-m-disabled"
```
mcoker: Same here - since this renders as an `<input>` element, we want to use the `disabled` attribute instead of the `pf-m-disabled class`.

ValerieKenyon: I had used both the `pf-m-disabled` class and the `disabled` attribute for both of these, and it rendered correctly.  I got right of the `disabled` attributes on both the `<input>` and `<button>` because this seemed to be causing the accessibility test failures.

mcoker: Ah gotcha. Looking at [this commit](https://github.com/patternfly/patternfly/pull/2829/commits/32a62317ccb6d2a6fba32c199e976a19d96d41f2) it looks like you replaced all of `form-control--attribute` with `disabled` via `form-control--attribute="disabled"`. Instead what you should do is take the existing value of `form-control--attribute` and add ` disabled ` to it somewhere between other attributes. For example, between `type="text" id="` as `type="text" disabled id="`

src/patternfly/components/Select/select-toggle-button.hbs
```diff
@@ -3,7 +3,14 @@
     <i class="fas fa-caret-down pf-c-select__toggle-arrow" aria-hidden="true"></i>
   {{/button}}
 {{else}}
-  {{#> button button--modifier="pf-m-plain pf-c-select__toggle-button" button--attribute=(concat 'id="' id '-toggle" aria-haspopup="true" aria-expanded="false" aria-labelledby="' id '-label ' id '-toggle" aria-label="Select"')}}
-    <i class="fas fa-caret-down pf-c-select__toggle-arrow" aria-hidden="true"></i>
-  {{/button}}
+  {{#if select--IsDisabled}}
+    {{#> button button--modifier="pf-m-plain pf-c-select__toggle-button" 
+      button--attribute=(concat 'id="' id '-toggle" aria-haspopup="true" aria-expanded="false" aria-labelledby="' id '-label ' id '-toggle" aria-label="Select" disabled')}}
```
mcoker: very nice!

src/patternfly/components/Select/select-toggle-typeahead.hbs
```diff
@@ -5,9 +5,17 @@
...
-  {{#> form-control controlType="input"
-    input="true"
-    form-control--modifier="pf-c-select__toggle-typeahead"
-    form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
-  {{/form-control}}
+  {{#if select--IsDisabled}}
+    {{#> form-control controlType="input"
+      input="true"
+      form-control--modifier="pf-c-select__toggle-typeahead"
+      form-control--attribute=(concat 'type="text" id="' id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '" disabled')}}
```
mcoker: perfect

ValerieKenyon:
Thank you for your guidance @mcoker !

mcoker:
Closing this PR via https://github.com/patternfly/patternfly/pull/2882 so we can merge to `v4` instead of master.

PR #2842, opened 2020-03-19
Don't merge/review: fix(table-word-break): fixes aggressive word wrapping and <tr> height

### LINKED ISSUE #1948: Update table font-weight var
change `--pf-c-table--FontWeight` to `--pf-c-table-cell--FontWeight` throughout the component.

### LINKED ISSUE #2205: Bug in table cell word-break

### LINKED ISSUE #2261: Address table component tr/td height
Currently we use `height: 1px` for `<tr>` elements and `height: inherit` for `<td>` elements. In chrome, this allows a child of the `<td>` to have `height: 100%` and it will grow to fill the height of the `<td>`. In firefox, however, this causes the child of the `<td>` to be `height: 1px`.

At a minimum, I think we need to remove this `height` configuration for the default table. I don't believe anything in our components depends on it, and it causes issues in Firefox. We could possibly create a modifier that applies similar height styles so that a child of a `<td>` can grow to fill the height of the `<td>`, but ideally we need a cross browser solution.

### LINKED ISSUE #2325: CSS tables agressively wrap, breaking words apart

mattnolting:
Preview: https://patternfly-pr-2842.surge.sh/documentation/core/components/table

closes #1948 
closes #2261 
closes #2205 
closes #2325 

## Breaking changes
1. Updates css variable names from `selector-element` to `selector--element`
2. Sets `<th>` and `<td>` elements to wrap by default
3. Sets `overflow-wrap` and `break-word` for a more graceful, less aggressive string break
4. Adds `.pf-m-no-wrap` to override default wrapping
5. Adds `.pf-c-table__button-content`
6. Refines some special cell placement
7. Updates css var `--responsive--` to `--grid--`
8. Adds css vars:
9. Decrements `max-width` `breakpoint` by 1


src/patternfly/components/Table/table-grid.scss
```diff
@@ -126,7 +113,7 @@
     display: block;
 
     &:first-of-type {
-      border-top: var(--pf-c-table-tbody--responsive--BorderWidth) solid var(--pf-c-table--responsive--BorderColor);
+      border-top: var(--pf-c-table--m-grid--tbody--BorderWidth) solid var(--pf-c-table--m-grid--BorderColor);
```
christiemolloy: border-top seems to be missing:

Screen Shot 2020-03-20 at 3 03 49 PM

src/patternfly/components/Button/button-icon.hbs
```diff
@@ -2,5 +2,9 @@
   {{#if button-icon--attribute}}
     {{{button-icon--attribute}}}
   {{/if}}>
-  {{> @partial-block}}
+  {{#if button-icon--name}}
```
christiemolloy: i think this should be in its own issue/pr since its just affecting the button and we can keep track of it

christiemolloy:
I wonder if we should support mixed table headers with wrapping and no wrapping, or maybe we could have a better example where we support it when theres like 1-2 lines of text not a whole sentence

Screen Shot 2020-03-20 at 3 20 40 PM

christiemolloy:
Looks like the table headers truncate by default now? Before they used to extend out? Do you think they should wrap by default like the other `th`'s ?

Screen Shot 2020-03-23 at 9 47 22 AM

christiemolloy:
`min-width: 16ch;` is affecting this here

Screen Shot 2020-03-23 at 9 49 42 AM

christiemolloy:
The min-width is also affecting here
Screen Shot 2020-03-23 at 9 53 04 AM

christiemolloy:
Screen Shot 2020-03-23 at 9 54 41 AM

Issue #2845, opened 2020-03-19
Updates to box-shadow global variables


mcoker:
https://marvelapp.com/6029867/screen/67313592

mcoker:
closed via https://github.com/patternfly/patternfly/pull/2854

PR #2854, opened 2020-03-19
fix(box-shadow): update global shadows, add xl, update shadow utility

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2845

@lboehling looks like these shadows are built from absolute black. We don't use black anywhere else in patternfly. The previous shadows were built from `$pf-color-black-1000` (`#151515`). Will these shadows work using `$pf-color-black-1000` as the color used in the `rgba()` function?

PR Preview: https://patternfly-pr-2854.surge.sh/
* Any component that uses the existing box-shadow global variables will be updated in this preview.
* Link to the box-shadow utility class examples: https://patternfly-pr-2854.surge.sh/documentation/core/utilities/boxshadow

### DIFF
--- a/src/patternfly/_variables.scss
+++ b/src/patternfly/_variables.scss
@@ -131,20 +131,25 @@
 
   // Shadows
   --pf-global--BoxShadow--sm: #{$pf-global--BoxShadow--sm};
-  --pf-global--BoxShadow--sm-right: #{$pf-global--BoxShadow--sm-right};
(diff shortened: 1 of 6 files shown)


lboehling:
I just chatted with @mcoker. We'll be updated the small directional shadows to these values so they're easier to see on selectable rows with borders. Otherwise, looks perfect! 
Screen Shot 2020-03-23 at 6 57 31 PM


@mcarrano I've updated these new values in the marvel and sketch files.

christiemolloy:

Issue #2849, opened 2020-03-19
Wizard shadow updates


mcoker:
https://marvelapp.com/6029867/screen/67313592

* change shadow from lg to xl
* mobile nav toggle
  * collapsed toggle keeps shadow
  * when toggle is expanded shadow is removed and replaced with 1px #d2d2d2 border
  * expanded nav keeps shadow
* side nav
  * replace shadow with 1px #d2d2d2 border

mcoker:
closed via https://github.com/patternfly/patternfly/pull/2860

PR #2860, opened 2020-03-20
fix(wizard): updates to shadows and borders

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2849

## Breaking changes
* changes wizard box shadow from lg to xl. This is a visual breaking change only. No further action required to consume this change.
* removes box shadow from mobile nav toggle when expanded, replaces with border. This is a visual breaking change only. No further action required to consume this change.
* replaces desktop nav box shadow with border. This is a visual breaking change only. No further action required to consume this change.
* Removes the following variables. Any references to them will need to be removed as they are no longer used in patternfly.
  * `--pf-c-wizard__nav--lg--BoxShadow`
  * `--pf-c-wizard--m-in-page__nav--lg--BoxShadow`
  * `--pf-c-wizard--m-in-page__nav--lg--BorderRightWidth`
  * `--pf-c-wizard--m-in-page__nav--lg--BorderRightColor`

### DIFF
--- a/src/patternfly/components/Wizard/examples/Wizard.css
+++ b/src/patternfly/components/Wizard/examples/Wizard.css
@@ -1,3 +1,4 @@
-#ws-core-c-wizard-in-page .ws-preview-html {
-  height: 400px;
+#ws-core-c-wizard-in-page .ws-preview-html,
+#ws-core-c-wizard-full-widthheight .ws-preview-html {
+  height: 600px;
 }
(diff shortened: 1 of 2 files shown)


mcoker:
@christiemolloy IMO it's not valuable for most people. I think we should just list what breaks, and if we clearly replace one variable with another (ie, `--component--Margin` with `--component--Padding`), we could list that, but otherwise just list the breaking change. If someone sees they're using something we listed as breaking, they can reference the PR or component docs for what changed exactly.

I think [bootstrap's 4.0 migration docs](https://getbootstrap.com/docs/4.0/migration/#by-component) serve as a good reference. WDYT?

mcoker:
@lboehling you're right, it's blocked by https://github.com/patternfly/patternfly/pull/2854 since that is the PR that introduces the XL box shadow. Once we get #2854 in, I can rebase this PR and pull in those changes.

christiemolloy:

Issue #2857, opened 2020-03-20
Empty State Secondary Actions


christiemolloy:
Tthe empty state secondary actions container has a margin-right, which is causing the whole block to be larger than the rest of the content in the empty state:

<img width="409" alt="Screen Shot 2020-03-20 at 10 58 51 AM" src="https://user-images.githubusercontent.com/20118816/77176080-f15a7800-6a99-11ea-835e-f899ee6f657c.png">

This issue can be more clearly seen in the insights dashboard, where the buttons are not centered with the rest of the text:
<img width="310" alt="Screen Shot 2020-03-20 at 10 59 06 AM" src="https://user-images.githubusercontent.com/20118816/77176284-3bdbf480-6a9a-11ea-9e3c-93256332cd49.png">

cc: @mcoker

PR #2866, opened 2020-03-23
fix(empty-state): center content, fix vars, adjust secondary margin

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2568
fixes https://github.com/patternfly/patternfly/issues/2857

## Breaking changes
* Introduces a new element `.pf-c-empty-state__content` that wraps the contents of the empty state content. The `max-width` used for the size variations applies to `.pf-c-empty-state__content` instead of `.pf-c-empty-state`, and `.pf-c-empty-state__content` is centered horizontally and vertically in its parent, reducing the need for using a `.pf-l-bullseye` layout to center empty state content. If you're using the bullseye layout to center the empty state horizontally, you can remove that, though there is likely no harm in continuing to use it.
  * Adds `--pf-c-empty-state__content--MaxWidth` that is overridden with `.pf-m-sm` and `.pf-m-lg` via `--pf-c-empty-state--m-sm__content--MaxWidth` and `--pf-c-empty-state--m-lg__content--MaxWidth`
  * Adds a `.pf-m-full-height` variation to `.pf-c-empty-state` for use when the empty state component should be centered vertically in a container, and the empty state does not already consume the height of its container.
* Adjusts the left/right spacing on buttons in the secondary actions container so that instead of using a `margin-right` to space the buttons, we use a `margin-left` and `margin-right`, which allows us to get rid of the negative right margin on the secondary actions container. We're also styling all direct children of the secondary actions container instead of just `.pf-c-button` elements.
  * removes `--pf-c-empty-state__secondary--MarginRight`. This was just to offset the right margin on buttons in the `__secondary` container and is no longer necessary. No updates needed unless you were referencing that var.
  * renames `--pf-c-empty-state__secondary__c-button--MarginRight` to `--pf-c-empty-state__secondary--child--MarginRight` and `--pf-c-empty-state__secondary__c-button--MarginBottom` to `--pf-c-empty-state__secondary--child--MarginBottom`. Any references you may have to this var need to be updated to the new names.
  * adds `--pf-c-empty-state__secondary--child--MarginLeft`. The margin between secondary actions is now calculated from `--pf-c-empty-state__secondary--child--MarginLeft` and `--pf-c-empty-state__secondary--child--MarginRight` between adjacent buttons.
* Fixes vars. Any references to the old vars will need to be updated to use the new vars.
  * renames `--pf-c-empty-state--c-button--MarginTop` to `--pf-c-empty-state__primary--MarginTop`
  * renames `--pf-c-empty-state--c-button__secondary--MarginTop` to `--pf-c-empty-state__primary--secondary--MarginTop`

### DIFF
--- a/src/patternfly/components/EmptyState/empty-state.scss
+++ b/src/patternfly/components/EmptyState/empty-state.scss
@@ -45,6 +45,10 @@
       font-size: var(--pf-c-empty-state--m-xl__body--FontSize);
     }
   }
+
+  &.pf-m-full-height {
+    height: 100%;
+  }
 }
 
 .pf-c-empty-state__content {
--- a/src/patternfly/components/EmptyState/empty-state.scss
+++ b/src/patternfly/components/EmptyState/empty-state.scss
@@ -10,9 +10,9 @@
   --pf-c-empty-state__primary--secondary--MarginTop: var(--pf-global--spacer--sm);
   --pf-c-empty-state__secondary--MarginTop: var(--pf-global--spacer--xl);
   --pf-c-empty-state__secondary--MarginBottom: calc(var(--pf-global--spacer--xs) * -1);
-  --pf-c-empty-state__secondary__c-button--MarginRight: calc(var(--pf-global--spacer--xs) / 2);
-  --pf-c-empty-state__secondary__c-button--MarginBottom: var(--pf-global--spacer--xs);
-  --pf-c-empty-state__secondary__c-button--MarginLeft: calc(var(--pf-global--spacer--xs) / 2);
+  --pf-c-empty-state__secondary--child--MarginRight: calc(var(--pf-global--spacer--xs) / 2);
+  --pf-c-empty-state__secondary--child--MarginBottom: var(--pf-global--spacer--xs);
+  --pf-c-empty-state__secondary--child--MarginLeft: calc(var(--pf-global--spacer--xs) / 2);
   --pf-c-empty-state--m-sm__content--MaxWidth: #{pf-size-prem(400px)};
   --pf-c-empty-state--m-lg__content--MaxWidth: #{pf-size-prem(600px)};
   --pf-c-empty-state--m-xl__body--FontSize: var(--pf-global--FontSize--xl);
@@ -77,10 +77,10 @@
   margin-top: var(--pf-c-empty-state__secondary--MarginTop);
   margin-bottom: var(--pf-c-empty-state__secondary--MarginBottom);
 
-  > .pf-c-button {
-    margin-right: var(--pf-c-empty-state__secondary__c-button--MarginRight);
-    margin-bottom: var(--pf-c-empty-state__secondary__c-button--MarginBottom);
-    margin-left: var(--pf-c-empty-state__secondary__c-button--MarginLeft);
+  > * {
+    margin-right: var(--pf-c-empty-state__secondary--child--MarginRight);
+    margin-bottom: var(--pf-c-empty-state__secondary--child--MarginBottom);
+    margin-left: var(--pf-c-empty-state__secondary--child--MarginLeft);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -71,17 +61,26 @@
...
   flex-wrap: wrap;
   justify-content: center;
   margin-top: var(--pf-c-empty-state__secondary--MarginTop);
-  margin-right: var(--pf-c-empty-state__secondary--MarginRight);
   margin-bottom: var(--pf-c-empty-state__secondary--MarginBottom);
 
   > .pf-c-button {
-    margin-right: var(--pf-c-empty-state__secondary--c-button--MarginRight);
-    margin-bottom: var(--pf-c-empty-state__secondary--c-button--MarginBottom);
+    margin-right: var(--pf-c-empty-state__secondary__c-button--MarginRight);
```
christiemolloy: I wonder if we should also support a plain `<a>` link that doesn't use `pf-c-button` ?

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -1,64 +1,54 @@
...
+  --pf-c-empty-state__content--MaxWidth: none;
   --pf-c-empty-state__icon--MarginBottom: var(--pf-global--spacer--lg);
   --pf-c-empty-state__icon--FontSize: var(--pf-global--icon--FontSize--xl);
   --pf-c-empty-state__icon--Color: var(--pf-global--icon--Color--light);
   --pf-c-empty-state__body--MarginTop: var(--pf-global--spacer--md);
   --pf-c-empty-state__body--Color: var(--pf-global--Color--200);
-  --pf-c-empty-state--c-button--MarginTop: var(--pf-global--spacer--xl);
-  --pf-c-empty-state--c-button__secondary--MarginTop: var(--pf-global--spacer--sm);
+  --pf-c-empty-state__primary--MarginTop: var(--pf-global--spacer--xl);
+  --pf-c-empty-state__primary--secondary--MarginTop: var(--pf-global--spacer--sm);
```
mattnolting:
```suggestion
  --pf-c-empty-state__secondary--MarginTop: var(--pf-global--spacer--sm);
```

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -1,64 +1,54 @@
...
-      + .pf-c-empty-state__secondary {
-        --pf-c-empty-state--c-button__secondary--MarginTop: var(--pf-c-empty-state--m-xl--c-button__secondary--MarginTop);
-      }
-    }
   }
 }
-// stylelint-enable
+
+.pf-c-empty-state__content {
+  max-width: var(--pf-c-empty-state__content--MaxWidth);
```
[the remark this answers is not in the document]
mcoker: updated and added `.pf-m-full-height`

src/patternfly/components/EmptyState/empty-state.scss
mcoker: Not sure what you mean - this is for when the secondary element follows the primary element/button. And `--pf-c-empty-state__secondary--MarginTop` is already defined in the line below

mattnolting: Oh, I see what you're doing!

christiemolloy:
Could you simplify the breaking change description into clear categories like you did in other PRs, so visual changes, variables removed, variable names changed, etc... and start each of the sections with a description about what is happening, i.e. "Renames the following variables..." I think that really helps make the description easier to digest

mcoker:
@christiemolloy took a stab at it, let me know what you think. Feel free to edit it if you'd like.

christiemolloy:
Nice updates! Release notes approved

Issue #2864, opened 2020-03-23
Remove bs-variables.scss


mcoker:
https://github.com/patternfly/patternfly/blob/master/src/patternfly/sass-utilities/bs-variables.scss

For the most part, this file hasn't been maintained since it was created, and isn't a part of our normal development workflow or testing process. Yet it is included in every build of the [base scss](https://github.com/patternfly/patternfly/blob/master/src/patternfly/sass-utilities/_all.scss).

Let's investigate the best way to remove it, yet potentially still make it available as a reference for anyone who might be using it.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/2871

PR #2871, opened 2020-03-24
fix(sass-utilities): made bs-variables optional include

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2864


christiemolloy:
Release notes reviewed and approved.

Issue #2276, opened 2019-09-17
Use of _shield-inheritable.scss classes inside heading elements results in incorrect heading font


rhamilto:
when using `pf-m-redhat-font`.  See https://github.com/openshift/console/pull/2668/files/8984c9a50e42071dcb3b9bbb068de804e27718e6#r325190181

mattnolting:
punting

PR #2872, opened 2020-03-24
feat(base): made shield styles optional by default (#2769)

mcoker:
fixes https://github.com/patternfly/patternfly-next/issues/2591

This PR removes:

```
  [class*="pf-c-"],
  [class*="pf-l-"] {
    &,
    &::before,
    &::after {
      box-sizing: border-box;
      padding: 0;
      margin: 0;
      background-color: transparent;
    }
  }
```

and this block that applies to all components and layouts (the top level component/layout selector, ie `.pf-c-data-toolbar`, `.pf-l-grid`, etc)

```
    font-family: var(--pf-global--FontFamily--sans-serif);
    font-size: var(--pf-global--FontSize--md);
    font-weight: var(--pf-global--FontWeight--normal);
    line-height: var(--pf-global--LineHeight--md);
    color: var(--pf-global--Color--100); // this color rule is set here to be able to use themes
    text-rendering: optimizeLegibility;
```

One thing to look out for is removing `background-color: transparent;` from `*`. Buttons have a background color by default, so I updated the button component so it works with this. I glanced at the rest of the components and they looked ok, but that's something to look out for.

Also the inheritable styles will no longer apply to a top-level component, and that may or may not be a good thing. Buttons, for example, should probably always reset their `color` so that in an element whose color is `red`, the button color doesn't inherit `red` as the color. All of our button variants now set the `color` but there may be other instances to update here.

## Breaking changes

The "shield" styles were intended to help resolve styling issues when using PF3 and PF4 together, but shouldn't be necessary for most PF users and is causing issues (https://github.com/patternfly/patternfly/issues/2276 as an example) as PatternFly evolves. 

This PR removes the global styles provided by https://github.com/mcoker/patternfly-next/blob/master/src/patternfly/_shield-inheritable.scss and https://github.com/mcoker/patternfly-next/blob/master/src/patternfly/_shield-noninheritable.scss, and makes them an optional include applications can include if necessary.

If an application is importing their own SCSS, they can set `$pf-global--enable-shield: true;` to enable the shield SCSS to be built in `_base.scss` or `patternfly.scss` (which imports `_base.scss`) - https://github.com/mcoker/patternfly-next/blob/f19b660e55350d47d059bfd32cf8a402ed4a2b2e/src/patternfly/_base.scss#L10-L13

Users can also import either the SCSS or compiled CSS via `https://github.com/mcoker/patternfly-next/blob/f19b660e55350d47d059bfd32cf8a402ed4a2b2e/src/patternfly/patternfly-shield-inheritable.scss` (which is compiled to patternfly-shield-inheritable.css) and https://github.com/mcoker/patternfly-next/blob/f19b660e55350d47d059bfd32cf8a402ed4a2b2e/src/patternfly/patternfly-shield-noninheritable.scss (which is compiled to patternfly-shield-noninheritable.css)

The contents of [_shield-inheritable.scss](https://github.com/mcoker/patternfly-next/blob/f19b660e55350d47d059bfd32cf8a402ed4a2b2e/src/patternfly/_shield-inheritable.scss) and [_shield-noninheritable.scss](https://github.com/mcoker/patternfly-next/blob/f19b660e55350d47d059bfd32cf8a402ed4a2b2e/src/patternfly/_shield-noninheritable.scss) are now SCSS mixins called `shield-inheritable` and `shield-noninheritable`, which allow us to include the blocks of code in SCSS conditionals.

This PR sets new styles on buttons necessary with the removal of `background-color: transparent` from all elements that the shield styles provided previously. 
* Link and Plain buttons now define `background-color: transparent` by default and on hover, focus, and active states.

The control button now sets a white background and default text color for its normal, hover, focus, and active states.

... tbc


mattnolting:
Perf 💯

PR #2880, opened 2020-03-25
chore(repo): replace 288 occurences of patternfly-next with patternfly

redallen:
Cleanup from repo renaming.


gatsby-config.js
```diff
@@ -5,7 +5,6 @@ module.exports = {
     title: 'PatternFly 4 - HTML',
     siteUrl: 'https://pf4.patternfly.org/'
   },
-  pathPrefix: '/patternfly-next',
```
mcoker: just checking - this is no longer needed?

redallen: Correct! No longer needed.

PR #2884, opened 2020-03-25
style(vertical-nav): refactored css structure

### LINKED ISSUE #1849: Horizontal Nav items focus border issues
image

image

related: https://github.com/patternfly/patternfly-react/issues/2054

### LINKED ISSUE #2196: Nav component doesn't have a disabled state for links
In the docs there is a `.pf-m-disabled` for `.pf-c-nav__link` but no styles for the state were written.

### LINKED ISSUE #2650: Update Vertical Nav Visuals
Want to update the selection state of vertical nav items to be more visible. These styling changes should be made across all vertical nav variants per the design.

Design here: https://marvelapp.com/21a8a2g6

cc @mcoker @LHinson

mattnolting:
closes #1849 
closes #2196
closes #2650 

**Nav component preview**
https://patternfly-pr-2884.surge.sh/documentation/core/demos/page

## Breaking changes
The nav component's css was refactored, see the file list for details on what changed 
https://github.com/patternfly/patternfly/pull/2884/files

This PR:
1. Applies stylistic updates to css variables rather than redefining component elements.
2. Updates `--pf-c-nav__c-nested-component--Property` to `--pf-c-nav--c-nested-component--Property`. Any instances of the old variable should be updated to the new variable.
3. Updates visual style of navigation only, no HTML updates are necessary.

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -257,6 +257,7 @@
   font-size: var(--pf-c-nav__link--FontSize);
   font-weight: var(--pf-c-nav__link--FontWeight);
   color: var(--pf-c-nav__link--Color);
+  user-select: text;
   background-color: var(--pf-c-nav__link--BackgroundColor);
 
   &::before,
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -9,17 +9,17 @@
   --pf-c-nav--m-dark__link--focus--Color: var(--pf-global--Color--light-100);
   --pf-c-nav--m-dark__link--active--Color: var(--pf-global--Color--light-100);
   --pf-c-nav--m-dark__link--m-current--Color: var(--pf-global--Color--light-100);
-  --pf-c-nav--m-dark__link--before--BorderColor: var(--pf-c-nav--m-dark__link--hover--BackgroundColor);
   --pf-c-nav--m-dark__link--after--BorderColor: var(--pf-global--active-color--400);
   --pf-c-nav--m-dark__link--BackgroundColor: transparent;
   --pf-c-nav--m-dark__link--hover--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
   --pf-c-nav--m-dark__link--m-current--BackgroundColor: var(--pf-global--BackgroundColor--dark-400);
   --pf-c-nav--m-dark__section-title--Color: var(--pf-global--Color--light-100);
-  --pf-c-nav--m-dark__section-title--BorderBottomColor: var(--pf-c-nav__link--before--BorderColor);
-  --pf-c-nav--m-dark__c-divider--BackgroundColor: var(--pf-c-nav__link--before--BorderColor);
+  --pf-c-nav--m-dark__section-title--BorderBottomColor: var(--pf-global--BackgroundColor--dark-200);
   --pf-c-nav--m-dark__subnav__link--not--m-current--hover--BorderColor: var(--pf-global--BorderColor--200);
   --pf-c-nav--m-dark__link--disabled--Color: var(--pf-global--disabled-color--100);
   --pf-c-nav--m-dark__link--disabled--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-nav--m-dark__item--before--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-nav--m-dark__c-divider--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
 
   // Link
   --pf-c-nav__link--FontSize: var(--pf-global--FontSize--md);
@@ -44,20 +44,11 @@
   --pf-c-nav__link--m-current--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-nav__link--m-current--after--BorderWidth: var(--pf-global--BorderWidth--xl);
 
-  // Link ::before borders
-  --pf-c-nav__link--before--BorderColor: var(--pf-c-nav__link--hover--BackgroundColor);
-  --pf-c-nav__link--before--BorderWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-nav__link--before--BorderTopWidth: 0;
-  --pf-c-nav__link--before--BorderBottomWidth: 0;
-
   // Link ::after borders
   --pf-c-nav__link--after--BorderColor: var(--pf-global--active-color--400);
   --pf-c-nav__link--after--BorderWidth: var(--pf-global--BorderWidth--xl);
   --pf-c-nav__link--after--BorderBottomWidth: 0;
   --pf-c-nav__link--after--BorderLeftWidth: 0;
-
-  // Nav list
-  --pf-c-nav__list__link--before--BorderWidth: var(--pf-global--BorderWidth--sm);
 
   // Simple list
   --pf-c-nav__simple-list__link--MarginTop: var(--pf-global--spacer--sm);
@@ -73,8 +64,6 @@
   --pf-c-nav__subnav__link--not--m-current--hover--BorderColor: var(--pf-global--BorderColor--dark-100);
 
   // Horizontal list
-  // --pf-c-nav__horizontal-list__item--MarginRight: var(--pf-global--spacer--xl);
-
   --pf-c-nav__horizontal-list__link--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-nav__horizontal-list__link--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-nav__horizontal-list__link--PaddingBottom: var(--pf-global--spacer--sm);
@@ -108,7 +97,7 @@
   --pf-c-nav__tertiary-list__link--after--BorderWidth: var(--pf-global--BorderWidth--lg);
   --pf-c-nav__tertiary-list__link--after--BorderColor: var(--pf-global--active-color--100);
 
-  // Horizonta/tertiary list offsets
+  // Horizontal/tertiary list offsets
   --pf-c-nav__link--offset: calc(var(--pf-global--spacer--xs) * -1);
 
   // Sub nav
@@ -116,8 +105,6 @@
   --pf-c-nav__subnav--PaddingBottom: var(--pf-global--spacer--md);
   --pf-c-nav__subnav--md--PaddingLeft: var(--pf-c-nav__link--PaddingLeft);
   --pf-c-nav__subnav__link--FontSize: var(--pf-global--FontSize--sm);
-  --pf-c-nav__subnav__link--before--BorderTopWidth: 0;
-  --pf-c-nav__subnav__link--before--BorderBottomWidth: 0;
   --pf-c-nav--subnav__simple-list__link--MarginTop: 0;
   --pf-c-nav--subnav__simple-list__link--MarginBottom: 0;
 
@@ -136,6 +123,10 @@
   --pf-c-nav__section-title--Color: var(--pf-global--Color--dark-200);
   --pf-c-nav__section-title--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-nav__section-title--BorderBottomColor: var(--pf-global--BorderColor--300);
+
+  // Nav item
+  --pf-c-nav__item--before--Height: var(--pf-global--BorderWidth--sm);
+  --pf-c-nav__item--before--BackgroundColor: var(--pf-global--BorderColor--300);
 
   // Scroll button
   --pf-c-nav__scroll-button--Display: none;
@@ -165,6 +156,7 @@
   --pf-c-nav__c-divider--MarginBottom: var(--pf-global--spacer--sm);
   --pf-c-nav--c-divider--PaddingRight: 0;
   --pf-c-nav--c-divider--PaddingLeft: 0;
+  --pf-c-nav__c-divider--BackgroundColor: var(--pf-global--BorderColor--300);
 
   @media screen and (min-width: $pf-global--breakpoint--md) {
     --pf-c-nav__link--PaddingRight: var(--pf-c-nav__link--md--PaddingRight);
@@ -200,6 +192,7 @@
     --pf-c-nav__subnav__link--not--m-current--hover--BorderColor: var(--pf-c-nav--m-dark__subnav__link--not--m-current--hover--BorderColor);
     --pf-c-nav__link--disabled--Color: var(--pf-c-nav--m-dark__link--disabled--Color);
     --pf-c-nav__link--disabled--BackgroundColor: var(--pf-c-nav--m-dark__link--disabled--BackgroundColor);
+    --pf-c-nav__item--before--BackgroundColor: var(--pf-c-nav--m-dark__item--before--BackgroundColor);
 
     .pf-c-divider {
       --pf-c-divider--after--BackgroundColor: var(--pf-c-nav--m-dark__c-divider--BackgroundColor);
@@ -214,7 +207,7 @@
 
   // Divider
   .pf-c-divider {
-    --pf-c-divider--after--BackgroundColor: var(--pf-c-nav__link--before--BorderColor);
+    --pf-c-divider--after--BackgroundColor: var(--pf-c-nav__c-divider--BackgroundColor);
 
     padding-right: var(--pf-c-nav--c-divider--PaddingRight);
     padding-left: var(--pf-c-nav--c-divider--PaddingLeft);
@@ -248,6 +241,20 @@
   transform: var(--pf-c-nav__item--m-expanded__toggle--i--Transform);
 }
 
+.pf-c-nav__item {
+  position: relative;
+
+  &::before {
+    position: absolute;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    height: var(--pf-c-nav__item--before--Height);
+    content: "";
+    background-color: var(--pf-c-nav__item--before--BackgroundColor);
+  }
+}
+
 // Nav link
 .pf-c-nav__link {
   position: relative;
@@ -260,32 +267,15 @@
   user-select: text;
   background-color: var(--pf-c-nav__link--BackgroundColor);
 
-  &::before,
   &::after {
     position: absolute;
     top: 0;
     bottom: 0;
     left: 0;
-
-    // set pointer events so text is selectable
-    pointer-events: none;
     content: "";
-    border-width: 0;
-  }
-
-  &::before {
-    right: 0;
-    border-color: var(--pf-c-nav__link--before--BorderColor);
-    border-style: solid;
-    border-top-width: var(--pf-c-nav__link--before--BorderTopWidth);
-    border-bottom-width: var(--pf-c-nav__link--before--BorderBottomWidth);
-  }
-
-  &::after {
     border-color: var(--pf-c-nav__link--after--BorderColor);
     border-style: solid;
-    border-bottom-width: var(--pf-c-nav__link--after--BorderBottomWidth);
-    border-left-width: var(--pf-c-nav__link--after--BorderLeftWidth);
+    border-width: 0 0 var(--pf-c-nav__link--after--BorderBottomWidth) var(--pf-c-nav__link--after--BorderLeftWidth);
   }
 
   // States
@@ -324,17 +314,6 @@
   }
 }
 
-// List default
-.pf-c-nav__list {
-  --pf-c-nav__link--before--BorderTopWidth: var(--pf-c-nav__link--before--BorderWidth);
-
-  border-bottom: var(--pf-c-nav__link--before--BorderWidth) solid var(--pf-c-nav__link--before--BorderColor);
-
-  .pf-c-nav__item:first-child {
-    --pf-c-nav__link--before--BorderTopWidth: 0;
-  }
-}
-
 // Active border treatment
 .pf-c-nav__list,
 .pf-c-nav__simple-list {
@@ -354,6 +333,10 @@
   --pf-c-nav--c-divider--PaddingRight: var(--pf-c-nav__simple-list__link--PaddingRight);
   --pf-c-nav--c-divider--PaddingLeft: var(--pf-c-nav__simple-list__link--PaddingLeft);
 
+  .pf-c-nav__item {
+    --pf-c-nav__item--before--BackgroundColor: transparent;
+  }
+
   .pf-c-nav__link {
     margin-top: var(--pf-c-nav__simple-list__link--MarginTop);
     margin-bottom: var(--pf-c-nav__simple-list__link--MarginBottom);
@@ -365,6 +348,10 @@
 .pf-c-nav__tertiary-list {
   --pf-c-nav__link--Right: var(--pf-c-nav__link--PaddingRight);
   --pf-c-nav__link--Left: var(--pf-c-nav__link--PaddingLeft);
+
+  .pf-c-nav__item {
+    --pf-c-nav__item--before--BackgroundColor: transparent;
+  }
 
   @include pf-overflow-hide-scroll;
 
@@ -441,8 +428,6 @@
 
 // Subnav
 .pf-c-nav__subnav {
-  --pf-c-nav__link--before--BorderTopWidth: var(--pf-c-nav__subnav__link--before--BorderTopWidth);
-  --pf-c-nav__link--before--BorderBottomWidth: var(--pf-c-nav__subnav__link--before--BorderBottomWidth);
   --pf-c-nav__link--FontSize: var(--pf-c-nav__subnav__link--FontSize);
   --pf-c-nav__simple-list__link--MarginTop: var(--pf-c-nav--subnav__simple-list__link--MarginTop);
   --pf-c-nav__simple-list__link--MarginBottom: var(--pf-c-nav--subnav__simple-list__link--MarginBottom);
@@ -452,10 +437,6 @@
   padding-bottom: var(--pf-c-nav__subnav--PaddingBottom);
   padding-left: var(--pf-c-nav__subnav--PaddingLeft);
   transition: var(--pf-c-nav--Transition);
-
-  .pf-c-nav__item:last-child {
-    --pf-c-nav__link--before--BorderBottomWidth: 0;
-  }
 
   .pf-c-nav__item.pf-m-expanded & {
     max-height: var(--pf-c-nav__subnav--MaxHeight);
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -285,7 +285,6 @@
 
   &::before {
     right: 0;
-    z-index: -1;
     border-color: var(--pf-c-nav__link--before--BorderColor);
     border-style: solid;
     border-top-width: var(--pf-c-nav__link--before--BorderTopWidth);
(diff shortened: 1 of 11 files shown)


christiemolloy:
Did design specify how the divider should look now that the items have borders ? @mcarrano 

Screen Shot 2020-03-26 at 11 50 39 AM

mcarrano:
Yeah, this does not look right to me @christiemolloy . Should we be having separators at all in the primary navigation? @mceledonia please also take a look at this. Looks good to me (other than the above issue), but would like you to review, also.

christiemolloy:
For the default nav, im looking at the mock ups, and it looks like the border/lines are the same color as the fill, which means that you wouldn't see the darker line:

Screen Shot 2020-03-26 at 4 48 27 PM

Screen Shot 2020-03-26 at 4 49 25 PM

christiemolloy:
Looks like the subnav link should have a white left border on it

Screen Shot 2020-03-26 at 4 53 53 PM

Screen Shot 2020-03-26 at 4 53 50 PM

christiemolloy:
From the mock-ups im seeing this version without borders, are we supporting that ?

Screen Shot 2020-03-26 at 4 56 55 PM

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,169 +1,118 @@
...
+  --pf-c-nav__link--after--BorderColor: var(--pf-global--active-color--400);
+  --pf-c-nav__link--after--BorderWidth: var(--pf-global--BorderWidth--xl);
+  --pf-c-nav__link--after--BorderBottomWidth: 0;
+  --pf-c-nav__link--after--BorderLeftWidth: 0;
+
+  // Nav list
+  --pf-c-nav__list__link--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-nav__list__link--PaddingRight: var(--pf-global--spacer--lg);
+  --pf-c-nav__list__link--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-nav__list__link--PaddingLeft: var(--pf-global--spacer--lg);
```
mcoker: The horizontal padding for links should be `var(--pf-global--spacer--md)` < `$pf-global--breakpoint--md` to match the overall spacing changes that happen with the page layout between desktop and mobile. The left edge of the nav should align with the left edge of the hamburger on mobile.

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,169 +1,118 @@
...
+  --pf-c-nav__list__link--md--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-nav__list__link--before--BorderWidth: var(--pf-global--BorderWidth--sm);
+
+  // Simple list
+  --pf-c-nav__simple-list__link--MarginTop: var(--pf-global--spacer--sm);
+  --pf-c-nav__simple-list__link--MarginBottom: var(--pf-global--spacer--sm);
+  --pf-c-nav__simple-list__link--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-nav__simple-list__link--PaddingRight: var(--pf-global--spacer--lg);
+  --pf-c-nav__simple-list__link--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-nav__simple-list__link--PaddingLeft: var(--pf-global--spacer--lg);
```
mcoker: I imagine the simple-list__link padding will be the same - 16px horizontal on mobile, 24px on desktop

mcoker:
The `:active` and `:focus` styles on the horizontal navigation looks like this 

Screen Shot 2020-03-26 at 6 37 19 PM

src/patternfly/components/Nav/nav.scss
```diff
@@ -260,76 +222,12 @@
 
   // Divider
   .pf-c-divider {
-    // Update during breaking change
-    padding-left: var(--pf-c-nav__c-divider--MarginLeft);
-
-    // end - Update during breaking change
+    padding-right: var(--pf-c-nav--c-divider--PaddingRight);
```
christiemolloy: why is padding being set to 0 here ? is it to override ? should it just be the value 0

christiemolloy:
This link doesn't appear to be disabled:

Screen Shot 2020-03-26 at 7 52 58 PM

mcoker:
An odd overflow issue with the horizontal nav

Screen Shot 2020-03-26 at 7 02 07 PM

mcoker:
Should tertiary nav have changed? The current component is blue on current/hover, and the text/underline are a darker shade of blue.

This PR
Screen Shot 2020-03-26 at 7 10 16 PM

Master/v4
Screen Shot 2020-03-26 at 7 09 20 PM

src/patternfly/components/Nav/nav.scss
mattnolting: It's set once here so that it can inherit padding values in different contexts, or be themed.

mcoker:
Looks like there is too much top padding on the horizontal nav between md and lg breakpoints

this pr
Screen Shot 2020-03-26 at 7 15 52 PM

master/v4
Screen Shot 2020-03-26 at 7 15 58 PM

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,180 +252,105 @@
...
-
 
-// Separator --- remove this block at breaking change
-.pf-c-nav__separator {
-  height: var(--pf-c-nav__separator--Height);
-  margin-top: var(--pf-c-nav__separator--MarginTop);
-  margin-bottom: var(--pf-c-nav__separator--MarginBottom);
-  margin-left: var(--pf-c-nav__separator--MarginLeft);
-  background-color: var(--pf-c-nav__separator--BackgroundColor);
+  &:active {
```
mcoker: Not part of the design, but can you move anywhere we're defining `:active` along with `:focus` so that the `:active` styles come after? Otherwise they'll never work, since `:focus` will always override.

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,180 +252,105 @@
...
+  color: var(--pf-c-nav__link--Color);
+  background-color: var(--pf-c-nav__link--BackgroundColor);
+
+  &::before,
+  &::after {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
```
mcoker: This makes the text unselectable because it overlays the the entire link. Is there any reason you wouldn't use `width/height/background-color`?  That's typically how we create these treatments using pseudo elements.

src/patternfly/components/Nav/nav.scss
```diff
@@ -230,12 +146,52 @@
...
+    --pf-c-nav__section-title--Color: var(--pf-c-nav--m-dark__section-title--Color);
+    --pf-c-nav__section-title--BorderBottomColor: var(--pf-c-nav--m-dark__section-title--BorderBottomColor);
+    --pf-c-nav__subnav__link--not--m-current--hover--BorderColor: var(--pf-c-nav--m-dark__subnav__link--not--m-current--hover--BorderColor);
+
+    .pf-c-divider {
+      --pf-c-divider--after--BackgroundColor: var(--pf-c-nav--m-dark__c-divider--BackgroundColor);
+    }
   }
 
   // Disable underline
```
mcoker: Not part of the redesign, but I think we could move this to `.pf-c-nav__link` and be a little more specific with the selector than `[class*="list"] [class*="link"]`

src/patternfly/components/Nav/nav.scss
```diff
@@ -548,157 +372,91 @@
...
+
+  &::-webkit-scrollbar {
+    display: none;
+    visibility: hidden;
+  }
 }
 
+// Section
 .pf-c-nav__section {
   margin-top: var(--pf-c-nav__section--MarginTop);
```
mcoker: It's not in this changeset, but small nit - the line below this can be written as `+ &`

src/patternfly/components/Nav/nav.scss
```diff
@@ -717,8 +475,26 @@
   font-size: var(--pf-c-nav__section-title--FontSize);
   font-weight: var(--pf-c-nav__section-title--FontWeight);
```
mcoker: not part of the design, but looks like we can remove this prop & var.

src/patternfly/components/Nav/nav.scss
```diff
@@ -717,8 +475,26 @@
...
+// Toggle caret
+.pf-c-nav__toggle {
+  flex: none;
+  padding-right: var(--pf-c-nav__toggle--PaddingRight);
+  padding-left: var(--pf-c-nav__toggle--PaddingLeft);
+  margin-left: auto;
+  font-size: var(--pf-c-nav__toggle--FontSize);
+  line-height: 1;
+  pointer-events: none;
+  border: 0;
```
mcoker: looks like these are legacy styles, but are `border` and `pointer-events` needed? Looks like it might have been a button at some point but isn't now as far as I can tell.

src/patternfly/components/Nav/nav.scss
mattnolting: It's not because of the pseudo selectors. Disable them and try to select the text. It seems to be the caused by changing the `display` value and/or adding padding to an `<a>` element. Our current nav exhibits the same behavior.

mattnolting: I find this method super efficient in that we don't have to specify multiple values to say, "Enable the lefthand border", like `top: 0; bottom: 0; left: 0; width: 4px; background-color: blue;`. Then for border right say `left: auto; right: 0; ` You can just say, "all border are 0 width and color blue, then give `--BorderLeftWidth` a width value to enable. It allows us to namespace active borders.

mcoker: Yeah, I still think that as a general principle, we should not introduce an invisible overlay, just for styling purposes, over the element that keeps you from interacting with the things in the element under it. Unless it's necessary of course - though it doesn't seem necessary here. We define most borders using pseudo elements with `width/height/background-color` in other components, so it's also a more consistent approach as well.

mattnolting: We'll need to update [tabs](https://patternfly-next-pr-2757.surge.sh/documentation/core/components/tabs) as well

mattnolting: And looks like [buttons](https://pf4.patternfly.org/documentation/core/components/button)

https://github.com/patternfly/patternfly/issues/2893

christiemolloy:
can you update the breaking change notes

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
-  .pf-m-current.pf-c-nav__link,
-  .pf-m-current > .pf-c-nav__link {
-    --pf-c-nav__list-link--FontWeight: var(--pf-c-nav__list-link--m-current--FontWeight);
-    --pf-c-nav__list-link--Color: var(--pf-c-nav__list-link--m-current--Color);
-
-    &::after {
-      background-color: var(--pf-c-nav__list-link--m-current--after--BackgroundColor);
-    }
+  &:hover {
+    color: var(--pf-c-nav__link--hover--Color);
```
mcoker, line 280: should color and background-color redefine `--pf-c-nav__link--hover` and `--pf-c-nav__link--BackgroundColor` for these states?

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
 }
 
-// Rotate the icon
-.pf-c-nav__item.pf-m-expanded .pf-c-nav__toggle > * {
-  transform: var(--pf-c-nav__item--m-expanded__toggle-icon--Transform);
+// Active border treatment
+.pf-c-nav__list,
+.pf-c-nav__simple-list {
+  .pf-c-nav__link.pf-m-current {
+    --pf-c-nav__link--before--BorderLeftWidth: var(--pf-c-nav__link--before--BorderWidth);
```
mcoker, line 318: I wonder if we should separate these out so they aren't tied to one another? If we style them differently and `--pf-c-nav__link--before--BorderWidth` doesn't control both, that's a breaking change.

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
+  --pf-c-nav__link--focus--BackgroundColor: var(--pf-c-nav__horizontal-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--m-current--BackgroundColor: var(--pf-c-nav__horizontal-list__link--m-current--BackgroundColor);
+  --pf-c-nav__link--m-current--Color: var(--pf-c-nav__horizontal-list__link--m-current--Color);
+  --pf-c-nav__link--before--BorderColor: var(--pf-c-nav__horizontal-list__link--before--BorderColor);
+
+  .pf-c-nav__link:hover,
+  .pf-c-nav__link:focus,
+  .pf-c-nav__link:active,
+  .pf-c-nav__link.pf-m-current {
+    --pf-c-nav__link--before--BorderBottomWidth: var(--pf-c-nav__horizontal-list__link--before--BorderWidth);
```
mcoker, line 395: Same here - what do you think about separating them out? It's less code to write this way and easier to theme, but less customizable and introduces a breaking change if we choose to style one differently. I think it's fair to expect we might style one differently as the simple nav does that.

mcoker: Same with tertiary

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
     max-height: var(--pf-c-nav__subnav--MaxHeight);
-    margin-top: var(--pf-c-nav__subnav--MarginTop);
     overflow-y: auto;
     opacity: 1;
   }
+
+  .pf-c-nav__link:not(.pf-m-current):hover {
+    --pf-c-nav__link--before--BorderLeftWidth: var(--pf-c-nav__subnav__link--not--m-current--hover--BorderLeftWidth);
+    --pf-c-nav__link--before--BorderColor: var(--pf-c-nav__subnav__link--not--m-current--hover--BorderColor);
+    --pf-c-nav__link--before--BorderTopWidth: 0;
```
mcoker, line 447: can remove this

src/patternfly/components/Nav/nav.scss
```diff
@@ -172,55 +114,27 @@
...
-  --pf-c-nav__tertiary-list-link--m-current--Color: var(--pf-global--link--Color);
-  --pf-c-nav__tertiary-list-link--m-current--after--BackgroundColor: var(--pf-global--link--Color);
+  // Nav item
+  --pf-c-nav__item--before--Height: var(--pf-global--BorderWidth--sm);
+  --pf-c-nav__item--before--BackgroundColor: var(--pf-global--BorderColor--300);
 
   // Scroll button
+  --pf-c-nav__scroll-button--Display: none;
+  --pf-c-nav__scroll-button--Visibility: hidden;
   --pf-c-nav__scroll-button--Top: var(--pf-global--spacer--sm);
```
mcoker, line 133: looks like you can remove this

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,169 +1,111 @@
...
+  --pf-c-nav__link--md--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-nav__link--Color: var(--pf-global--Color--dark-100);
+  --pf-c-nav__link--hover--Color: var(--pf-global--Color--dark-100);
+  --pf-c-nav__link--active--Color: var(--pf-global--Color--dark-100);
+  --pf-c-nav__link--focus--Color: var(--pf-global--Color--dark-100);
+  --pf-c-nav__link--disabled--Color: var(--pf-global--disabled-color--100);
+  --pf-c-nav__link--BackgroundColor: transparent;
+  --pf-c-nav__link--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
+  --pf-c-nav__link--active--BackgroundColor: var(--pf-c-nav__link--hover--BackgroundColor);
+  --pf-c-nav__link--focus--BackgroundColor: var(--pf-c-nav__link--hover--BackgroundColor);
```
mcoker, line 41: Same here - if we're going to have a var for each, we should probably not create the dependency between them if one changes.

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
-      content: "";
-    }
+  --pf-c-nav__link--PaddingTop: var(--pf-c-nav__tertiary-list__link--PaddingTop);
+  --pf-c-nav__link--PaddingRight: var(--pf-c-nav__tertiary-list__link--PaddingRight);
+  --pf-c-nav__link--PaddingBottom: var(--pf-c-nav__tertiary-list__link--PaddingBottom);
+  --pf-c-nav__link--PaddingLeft: var(--pf-c-nav__tertiary-list__link--PaddingLeft);
+  --pf-c-nav__link--FontWeight: var(--pf-c-nav__tertiary-list__link--FontWeight);
+  --pf-c-nav__link--Color: var(--pf-c-nav__tertiary-list__link--Color);
+  --pf-c-nav__link--hover--Color: var(--pf-c-nav__tertiary-list__link--hover--Color);
+  --pf-c-nav__link--active--Color: var(--pf-c-nav__tertiary-list__link--hover--Color);
```
mcoker, line 408:
```suggestion
  --pf-c-nav__link--active--Color: var(--pf-c-nav__tertiary-list__link--active--Color);
```

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
+  --pf-c-nav__link--focus--Color: var(--pf-c-nav__tertiary-list__link--hover--Color);
```
mcoker, line 409:
```suggestion
  --pf-c-nav__link--focus--Color: var(--pf-c-nav__tertiary-list__link--focus--Color);
```

src/patternfly/demos/Page/page-demo-expandable-nav.hbs
```diff
@@ -84,11 +84,6 @@
             {{/nav-list}}
           {{/nav-subnav}}
         {{/nav-item}}
-        {{#if page--IsLegacySeparators}}
-          {{#> nav-separator}}{{/nav-separator}}
-        {{else}}
-          {{#> divider divider--type="li"}}{{/divider}}
-        {{/if}}
```
mcoker, line 91: Did you want to pull the nav-separator out of this component in this PR? Works for me if you do, I saw this and removed that element from the documentation, but realized we still need to yank out

https://github.com/mattnolting/patternfly/blob/3fa5a9097a7487fc87d84f776ea60218b1ef07d3/src/patternfly/demos/Page/page-demo-expandable-nav.hbs#L45-L49

https://github.com/mattnolting/patternfly/blob/style-vertical-nav/src/patternfly/components/Nav/nav-separator.hbs

and the 2 examples in https://github.com/mattnolting/patternfly/blob/style-vertical-nav/src/patternfly/demos/Page/examples/Page.md that reference `IsLegacySeparators`

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
+  --pf-c-nav__link--BackgroundColor: var(--pf-c-nav__tertiary-list__link--BackgroundColor);
+  --pf-c-nav__link--hover--BackgroundColor: var(--pf-c-nav__tertiary-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--active--BackgroundColor: var(--pf-c-nav__tertiary-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--focus--BackgroundColor: var(--pf-c-nav__tertiary-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--m-current--BackgroundColor: var(--pf-c-nav__tertiary-list__link--m-current--BackgroundColor);
```
mcoker, line 414: `--pf-c-nav__tertiary-list__link--m-current--BackgroundColor` is undefined.

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +235,275 @@
...
+  --pf-c-nav__link--PaddingLeft: var(--pf-c-nav__horizontal-list__link--PaddingLeft);
+  --pf-c-nav__link--Color: var(--pf-c-nav__horizontal-list__link--Color);
+  --pf-c-nav__link--hover--Color: var(--pf-c-nav__horizontal-list__link--hover--Color);
+  --pf-c-nav__link--active--Color: var(--pf-c-nav__horizontal-list__link--active--Color);
+  --pf-c-nav__link--focus--Color: var(--pf-c-nav__horizontal-list__link--focus--Color);
+  --pf-c-nav__link--BackgroundColor: var(--pf-c-nav__horizontal-list__link--BackgroundColor);
+  --pf-c-nav__link--hover--BackgroundColor: var(--pf-c-nav__horizontal-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--active--BackgroundColor: var(--pf-c-nav__horizontal-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--focus--BackgroundColor: var(--pf-c-nav__horizontal-list__link--hover--BackgroundColor);
+  --pf-c-nav__link--m-current--BackgroundColor: var(--pf-c-nav__horizontal-list__link--m-current--BackgroundColor);
```
mcoker, line 387: `--pf-c-nav__horizontal-list__link--m-current--BackgroundColor` is undefined. Another approach could be to set the following to transparent by default, then only declare them where they're needed, instead of having to unset them all in horizontal and tertiary nav

```
  --pf-c-nav__link--BackgroundColor
  --pf-c-nav__link--hover--BackgroundColor
  --pf-c-nav__link--active--BackgroundColor
  --pf-c-nav__link--focus--BackgroundColor
  --pf-c-nav__link--m-current--BackgroundColor
```

mattnolting:
@mcoker updated!

mcoker:
Doesn't look like the `pf-m-start/end` classes are working on the horizontal nav:

Screen Shot 2020-04-03 at 2 18 20 PM
Screen Shot 2020-04-03 at 2 19 37 PM

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +253,290 @@
...
 
-  // Hide scrollbars
-  &::-webkit-scrollbar {
-    display: none;
-    visibility: hidden;
-  }
-
-  // Expanded list item
   .pf-c-nav__item.pf-m-expanded & {
     max-height: var(--pf-c-nav__subnav--MaxHeight);
```
mcoker, line 471: This var name is probably better as `--pf-c-nav__item--m-expanded__subnav--MaxHeight` since `__subnav` actually has a `max-height`

And since you declare `max-height` for `__subnav`, you could use `--pf-c-nav__subnav--MaxHeight` there, then here declare `--pf-c-nav__subnav--MaxHeight: var(--pf-c-nav__item--m-expanded__subnav--MaxHeight);`

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +253,290 @@
...
+  }
 }
 
+// Section
 .pf-c-nav__section {
   margin-top: var(--pf-c-nav__section--MarginTop);
 
-  + .pf-c-nav__section {
+  + & {
     --pf-c-nav__section--MarginTop: var(--pf-c-nav__section__section--MarginTop);
```
mcoker, line 492: can we change this to `--pf-c-nav__section--section--MarginTop`?

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +251,290 @@
...
   .pf-c-nav__item.pf-m-expanded & {
     max-height: var(--pf-c-nav__subnav--MaxHeight);
-    margin-top: var(--pf-c-nav__subnav--MarginTop);
     overflow-y: auto;
     opacity: 1;
   }
+
+  .pf-c-nav__link:not(.pf-m-current):hover {
+    --pf-c-nav__link--before--BorderLeftWidth: var(--pf-c-nav__subnav__link--not--m-current--hover--BorderLeftWidth);
+    --pf-c-nav__link--before--BorderColor: var(--pf-c-nav__subnav__link--not--m-current--hover--BorderColor);
```
christiemolloy, line 478: could this variable just be `--pf-c-nav__subnav__link--hover--BorderColor`, since that is the main case

src/patternfly/components/Nav/nav.scss
```diff
@@ -462,20 +462,21 @@
   --pf-c-nav__simple-list__link--MarginBottom: var(--pf-c-nav--subnav__simple-list__link--MarginBottom);
   --pf-c-nav__simple-list__link--FontSize: var(--pf-c-nav__subnav__link--FontSize);
 
-  max-height: 0;
+  max-height: var(--pf-c-nav__subnav--MaxHeight);
```
mcoker, line 465: looks like we need to define `--pf-c-nav__subnav--MaxHeight: 0;`

src/patternfly/components/Nav/nav.scss
```diff
@@ -474,9 +485,24 @@
     opacity: 1;
   }
 
-  .pf-c-nav__link:not(.pf-m-current):hover {
-    --pf-c-nav__link--before--BorderLeftWidth: var(--pf-c-nav__subnav__link--hover--BorderLeftWidth);
-    --pf-c-nav__link--before--BorderColor: var(--pf-c-nav__subnav__link--hover--BorderColor);
+  .pf-c-nav__link:hover {
```
mcoker, line 488: very nice!

mattnolting:
@mcarrano @mceledonia updated per our meeting

- Removed references in examples
- Removed `disabled` background color in light theme

@mcoker noted that this fails color contrast
https://files.slack.com/files-pri/T0FA4HM8D-F011AUCHD8S/screen_shot_2020-04-07_at_12.52.12_pm.png

I also left the cursor as is right now. We can open another issue to track.

mcoker:
Looks like the hover state for the scroll buttons changed colors - is that intended?

Screen Shot 2020-04-07 at 1 19 09 PM

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +268,330 @@
...
-
-    &.pf-m-focus,
-    &:focus {
-      --pf-c-nav__horizontal-list-link--Color: var(--pf-c-nav__horizontal-list-link--focus--Color);
-
-      &::after {
-        background-color: var(--pf-c-nav__horizontal-list-link--focus--after--BackgroundColor);
-      }
-    }
+  .pf-c-nav__link::after {
```
mcoker, line 421: can you nest this and the `::before` in the `.pf-c-nav__link` selector above?

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +268,330 @@
...
-    &::after {
-      position: absolute;
-      bottom: 0;
-      left: 0;
-      display: block;
-      width: 100%;
-      height: var(--pf-c-nav__horizontal-list-link--after--Height);
-      content: "";
-    }
+  .pf-c-nav__link.pf-m-current {
```
mcoker, line 457: can you scss nest hover/focus/active/current?

src/patternfly/components/Nav/nav.scss
```diff
@@ -354,389 +268,330 @@
...
-      --pf-c-nav__tertiary-list-link--Color: var(--pf-c-nav__tertiary-list-link--focus--Color);
+  .pf-c-nav__link:active {
+    --pf-c-nav__link--before--BorderBottomWidth: var(--pf-c-nav__tertiary-list__link--active--before--BorderWidth);
+  }
 
-      &::after {
-        background-color: var(--pf-c-nav__tertiary-list-link--focus--after--BackgroundColor);
-      }
-    }
+  .pf-c-nav__link.pf-m-current {
```
mcoker, line 494: can you nest hover/focus/active/current?

mcoker:
@mattnolting for markup changes in the breaking changes notes, can you verify that the nav separator was removed in this PR and the divider should be used instead? Also looks like this is where there was a change to the `.pf-c-nav__toggle-icon` element applied to a wrapper instead of the icon.

Issue #2851, opened 2020-03-19
Data list shadow updates


mcoker:
https://marvelapp.com/6029867/screen/67313592

remove shadow from expanded rows, use grey border instead

PR #2885, opened 2020-03-25
fix(data-list): changed expanded content shadow to a border

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2851

## Breaking changes
* Replaces the box shadow on the expandable content element with a bottom border
* Removes `--pf-c-data-list__expandable-content--BoxShadow`,  `--pf-c-data-list__item--before--Height`, and `--pf-c-data-list__item-item--before--Height`. Any instances of these variables in your application should be removed as they are no longer used.
* Replaces `--pf-c-data-list__item--m-expanded__expandable-content--BorderTopWidth` with `--pf-c-data-list__expandable-content--BorderTopWidth`. Any instances of the old variable should be updated to the new variable.

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -41,9 +41,11 @@
   --pf-c-data-list__item--before--ZIndex: var(--pf-global--ZIndex--xl);
   --pf-c-data-list__item--before--Top: 0;
   --pf-c-data-list__item--before--Bottom: 0;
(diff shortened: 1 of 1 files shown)


christiemolloy:
Should also add to the description, that you removed these vars:

  --pf-c-data-list__item--before--Height
 
  --pf-c-data-list__item-item--before--Height: initial

mattnolting:
Release notes approved

Issue #2853, opened 2020-03-19
Drawer shadow updates


mcoker:
https://marvelapp.com/6029867/screen/67313592

* add variation that replaces shadow with grey border for use on white backgrounds when drawer expands inline with page content.
* small left shadow to be used on drawer that expand inline with content that is placed on a grey background
* medium left shadow will remain and only be used when drawers expand over content.

PR #2887, opened 2020-03-26
fix(drawer): removed m-border, enabled m-no-border on inline/static

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2853

@lboehling from the issue

> add variation that replaces shadow with grey border for use on white backgrounds when drawer expands inline with page content.

The non-inline drawers have a shadow by default and can have a border instead with use of a `pf-m-border` variation. You can see that here https://www.patternfly.org/v4/documentation/core/components/drawer

Is that already done, or did you want a different update?

> small left shadow to be used on drawer that expand inline with content that is placed on a grey background

When the panel displays inline with the content, the drawer changed the shadow to a border. In this PR, I disabled that behavior, so that the panel has a shadow by default, and you can set a border instead by using the `.pf-m-border` variation mentioned above. I also updated that shadow size when used on an inline drawer to be `sm`

> medium left shadow will remain and only be used when drawers expand over content.

I'm showing a `lg` shadow. Do you want to update it to be `md`?

----

Also currently you can apply `.pf-m-border` to a panel that expands on top of content - is that OK or do we want to not support that - or maybe not show examples of that? And any other changes to the examples? For example, none of our examples show content with a grey background because the grey background would come from the page component. I created a demo showing the inline drawer's shadow in https://patternfly-pr-2887.surge.sh/documentation/core/demos/drawer

## Breaking changes
* This PR removes the `.pf-m-border` variation as when the panel overlays content, a shadow should always be used instead of a border. And when the panel is inline with content (`pf-m-inline`, `pf-m-static`), a border should always be used instead of a shadow.
* Enables `.pf-m-no-border` on inline and static variations, to remove the border if necessary.

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -287,15 +287,9 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   // stylelint-enable
 
   // Border
-  .pf-c-drawer .pf-c-drawer__panel.pf-m-border,
-  .pf-c-drawer.pf-m-panel-left .pf-c-drawer__panel.pf-m-border,
   .pf-c-drawer .pf-c-drawer__panel.pf-m-no-border,
   .pf-c-drawer.pf-m-panel-left .pf-c-drawer__panel.pf-m-no-border {
     --pf-c-drawer__panel--BoxShadow: transparent;
-  }
-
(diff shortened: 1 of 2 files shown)


mceledonia:
@mattnolting @mcoker I definitely see what you mean but I'm leaning towards being consistent with our logic here - shadows for overlap, border for separation. I think it looks good, and in context of a larger page should usually be around some more depth!

christiemolloy:
Release notes reviewed and approved.

PR #2894, opened 2020-03-27
chore(snippets): gulpify snippets, add html validation

redallen:
This PR integrates the new `parseHBS.js` for VSCode snippets into our existing gulp scripts. Notably, it:
- Uses existing `function compileHBS` in `html.js` to create the Handlebars instance
- Uses existing `workspace/**/*.html` generated by `function compileMD` to create the HTML from MD
- Outputs JSON files to `workspace` folder

WIP because it needs more work to make `lint:hbs` work and the package.json deps need to be cleaned.

Was prettier really checking for valid HTML syntax @jschuler? I'd like to find an alternative that takes <60s and make the check a part of `compileMD`.


build/gulp/snippets.js
```diff
@@ -0,0 +1,119 @@
...
+const through2 = require('through2');
+const { render } = require('html-formatter');
+const cheerio = require('cheerio');
+const { hbsInstance, hbsFileMap } = require('./html');
+
+function prettyHTML(html) {
+  return render(` ${html} `).replace(/\t/g, '  ');
+}
+
+function generateSnippets(htmlIndexFiles) {
```
jschuler: works great in the pf snippet vscode extension

Issue #2523, opened 2019-12-16
App launcher: update to external link icon element


mcoker:
To make the use of the external link icon more flexible and match the react implementation, let's update `.pf-c-app-launcher__menu-item-external-icon` so that it is a `<span>` that wraps the icon used inside instead of applying it directly to the icon.

PR #2904, opened 2020-03-30
fix(app launcher): applied external link icon class to icon wrapper

ValerieKenyon:
Fixes https://github.com/patternfly/patternfly/issues/2523
Updated `.pf-c-app-launcher__menu-item-external-icon` so that it is a <span> that wraps the icon used inside instead of applying it directly to the icon.

## Breaking changes
Moves the class `.pf-c-app-launcher__menu-item-external-icon` from the external link icon itself to a `<span>` that wraps the external link icon. Any instances of the external link icon should be wrapped in a `<span>` and the `.pf-c-app-launcher__menu-item-external-icon` class should be moved from the icon to the `<span>`.

### DIFF
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
@@ -1,8 +1,13 @@
-<span class="pf-c-app-launcher__menu-item-external-icon">
-  <i class="fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-    aria-hidden="true"
-    {{#if app-launcher-menu-item-external-icon--attribute}}
+<span class="pf-c-app-launcher__menu-item-external-icon
+  {{#if app-launcher-menu-item-external-icon--modifier}} 
+    {{app-launcher-menu-item-external-icon--modifier}}
+  {{/if}}
+  {{#if app-launcher-menu-item-external-icon--attribute}}
 	    {{{app-launcher-menu-item-external-icon--attribute}}}
-    {{/if}}></i>
+  {{/if}}"
+>
+  <i class="fas fa-external-link-alt"
+    aria-hidden="true"
+    ></i>
 </span>
 <span class="pf-screen-reader">(opens new window)</span>
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
@@ -1,9 +1,7 @@
-<span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}}{{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
+<span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
   {{#if app-launcher-menu-item-external-icon--attribute}}
-	    {{{app-launcher-menu-item-external-icon--attribute}}}
+	  {{{app-launcher-menu-item-external-icon--attribute}}}
   {{/if}}>
-  <i class="fas fa-external-link-alt"
-    aria-hidden="true"
-    ></i>
+  <i class="fas fa-external-link-alt" aria-hidden="true"></i>
 </span>
 <span class="pf-screen-reader">(opens new window)</span>
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
@@ -1,6 +1,8 @@
-<i class="pf-c-app-launcher__menu-item-external-icon fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-  aria-hidden="true"
-  {{#if app-launcher-menu-item-external-icon--attribute}}
-	  {{{app-launcher-menu-item-external-icon--attribute}}}
-  {{/if}}></i>
+<span class="pf-c-app-launcher__menu-item-external-icon">
+  <i class="fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
+    aria-hidden="true"
+    {{#if app-launcher-menu-item-external-icon--attribute}}
+	    {{{app-launcher-menu-item-external-icon--attribute}}}
+    {{/if}}></i>
+</span>
 <span class="pf-screen-reader">(opens new window)</span>
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
@@ -1,6 +1,6 @@
 <span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
   {{#if app-launcher-menu-item-external-icon--attribute}}
-	  {{{app-launcher-menu-item-external-icon--attribute}}}
+    {{{app-launcher-menu-item-external-icon--attribute}}}
   {{/if}}>
   <i class="fas fa-external-link-alt" aria-hidden="true"></i>
 </span>
--- a/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
+++ b/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
@@ -1,11 +1,7 @@
-<span class="pf-c-app-launcher__menu-item-external-icon
-  {{#if app-launcher-menu-item-external-icon--modifier}} 
-    {{app-launcher-menu-item-external-icon--modifier}}
-  {{/if}}
+<span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}}{{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
   {{#if app-launcher-menu-item-external-icon--attribute}}
 	    {{{app-launcher-menu-item-external-icon--attribute}}}
-  {{/if}}"
->
+  {{/if}}>
   <i class="fas fa-external-link-alt"
     aria-hidden="true"
     ></i>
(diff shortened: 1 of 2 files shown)


src/patternfly/components/AppLauncher/examples/AppLauncher.md
```diff
@@ -328,7 +328,7 @@ import './AppLauncher.css'
 | `.pf-c-app-launcher__group-title` | `<h1>` | Defines a title for a group of items. |
 | `.pf-c-app-launcher__menu-item` | `<a>`, `<button>` | Defines a menu item. |
 | `.pf-c-app-launcher__menu-item-icon` | `<span>` | Defines the wrapper for the menu item icon. |
-| `.pf-c-app-launcher__menu-item-external-icon` | `<i>` | Defines the external link icon that appears on hover/focus. Use with `.pf-m-external`. |
+| `.pf-c-app-launcher__menu-item-external-icon` | `<span>` | Defines the external link icon that appears on hover/focus. Use with `.pf-m-external`. |
```
mcoker: Can you match with the working in the previous line? Something like "Defines the wrapper for the external link icon that appears on hover/focus. Use with `.pf-m-external`."

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,8 @@
-<i class="pf-c-app-launcher__menu-item-external-icon fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-  aria-hidden="true"
-  {{#if app-launcher-menu-item-external-icon--attribute}}
-	  {{{app-launcher-menu-item-external-icon--attribute}}}
-  {{/if}}></i>
+<span class="pf-c-app-launcher__menu-item-external-icon">
+  <i class="fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
```
mcoker: Now that the `<span>` is the BEM element, can you move the `{{#if}}` logic for both `app-launcher-menu-item-external-icon--modifier` and `app-launcher-menu-item-external-icon--attribute` to the span?

mcoker:
Hey @ValerieKenyon, this is looking really good! Can you also update https://github.com/patternfly/patternfly/blame/0501cc040b2846779f38d97ba84cb3108076f159/src/patternfly/components/AppLauncher/examples/AppLauncher.md#L319? Looks like you can probably just change 

```
| `aria-hidden="true"` | `.pf-c-app-launcher__menu-item-external-icon` | Hides the icon from assistive technologies. |
```

to

```
| `aria-hidden="true"` | `.pf-c-app-launcher__menu-item-external-icon > *` | Hides the icon from assistive technologies. |
```

mcoker:
Also can you update the PR title so it matches the commit message?

`feat(app launcher): update to external link icon element issue`

Though would you mind using `fix()` instead of `feat()` since the original goal for the change (made in react) was to fix an issue due to the class being on the icon itself, and that BEM element already exists, so we aren't adding any sort of functionality. And even better if that were the actual commit message, though it's OK to leave that as I can just modify it when we merge the PR.

src/patternfly/components/AppLauncher/examples/AppLauncher.md
ValerieKenyon: OK

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
ValerieKenyon: I was going to break it up into separate lines for readability like this
Screen Shot 2020-03-30 at 16 26 00

ValerieKenyon: And fixed the next item (for line 319)

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,13 @@
-<i class="pf-c-app-launcher__menu-item-external-icon fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-  aria-hidden="true"
+<span class="pf-c-app-launcher__menu-item-external-icon
+  {{#if app-launcher-menu-item-external-icon--modifier}}
```
mcoker: Currently this generates the markup `<span class="pf-c-app-launcher__menu-item-external-icon ">...` if there is nothing passted to `app-launcher-menu-item-external-icon--modifier`. Moving `{{#if ...` up a line so that it is `<span class="pf-c-app-launcher__menu-item-external-icon{{#if ...` is how we usually write it, so as to be consistent with the rest of our examples and to remove any extra space characters in the `class` attribute.

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,13 @@
-<i class="pf-c-app-launcher__menu-item-external-icon fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-  aria-hidden="true"
+<span class="pf-c-app-launcher__menu-item-external-icon
+  {{#if app-launcher-menu-item-external-icon--modifier}} 
+    {{app-launcher-menu-item-external-icon--modifier}}
+  {{/if}}
   {{#if app-launcher-menu-item-external-icon--attribute}}
```
mcoker: Since `app-launcher-menu-item-external-icon--attribute` is to generate html attributes, not extra strings to be passed to the `class` attribute, you'll need to add a close `"` between the conditional for `app-launcher-menu-item-external-icon--modifier` and the one for `app-launcher-menu-item-external-icon--attribute`, then remove the extra `"` below after the conditional for `app-launcher-menu-item-external-icon--attribute`.

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,13 @@
...
-	  {{{app-launcher-menu-item-external-icon--attribute}}}
-  {{/if}}></i>
+	    {{{app-launcher-menu-item-external-icon--attribute}}}
+  {{/if}}"
+>
```
mcoker: And can you move this up a line so it is after the closing `{{/if}}` block?

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
mcoker: For reference, you can look at most of the other handlebars partials in this directory. The [app-launcher-menu-search.hbs](https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/AppLauncher/app-launcher-menu-search.hbs) is a good one.

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
ValerieKenyon: I think I've got the last 3 things fixed

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,9 @@
-<i class="pf-c-app-launcher__menu-item-external-icon fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-  aria-hidden="true"
+<span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}}{{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
```
mcoker: need a space here in the conditional, so that if `app-launcher-menu-item-external-icon--modifier` is set, there will be a space between `pf-c-app-launcher__menu-item-external-icon` and the value of `app-launcher-menu-item-external-icon--modifier`

```suggestion
<span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
```

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,9 @@
...
   {{#if app-launcher-menu-item-external-icon--attribute}}
-	  {{{app-launcher-menu-item-external-icon--attribute}}}
-  {{/if}}></i>
+	    {{{app-launcher-menu-item-external-icon--attribute}}}
+  {{/if}}>
+  <i class="fas fa-external-link-alt"
```
mcoker: This is a short enough element that it can all be on one line.

src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs
```diff
@@ -1,6 +1,9 @@
-<i class="pf-c-app-launcher__menu-item-external-icon fas fa-external-link-alt{{#if app-launcher-menu-item-external-icon--modifier}} {{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
-  aria-hidden="true"
+<span class="pf-c-app-launcher__menu-item-external-icon{{#if app-launcher-menu-item-external-icon--modifier}}{{app-launcher-menu-item-external-icon--modifier}}{{/if}}"
   {{#if app-launcher-menu-item-external-icon--attribute}}
-	  {{{app-launcher-menu-item-external-icon--attribute}}}
-  {{/if}}></i>
+	    {{{app-launcher-menu-item-external-icon--attribute}}}
```
mcoker: Nit pick - can you update the indentation here so that there are 4 leading spaces, since we use 2 spaces for indentation and this line is indented 2 levels deep?

mcoker: Hey @ValerieKenyon looks like you pushed an update, but I'm still seeing too much indentation here https://github.com/patternfly/patternfly/blob/f52e914ad15e315bd98162a43dba5144cdd48e1b/src/patternfly/components/AppLauncher/app-launcher-menu-item-external-icon.hbs#L3

ValerieKenyon: I don't understand why that happened.  I did update it, and it showed indented 4 spaces in my VS Code.  I'm sorry about that.  I'll try to fix that now.

ValerieKenyon: I think I see where my mistake was.  The first 2 spaces in my file were from a tab, not the space bar.  I've fixed that and pushed the change, I think it looks okay now.  Again, sorry about that.

mcoker: Looks great now, thanks!

Issue #2915, opened 2020-03-24
Breadcrumbs don't wrap,


dsesami:
https://www.patternfly.org/v4/documentation/react/components/breadcrumb

Similar to https://github.com/patternfly/patternfly-react/issues/3938, but for breadcrumbs.
![Screenshot from 2020-03-24 11-22-12](https://user-images.githubusercontent.com/19942822/77446746-2bdb5200-6dc5-11ea-9801-57ec16b28530.png)

PR #2916, opened 2020-04-01
fix(breadcrumb): add word-break, display link and icon inline, move divider

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2915

@mcarrano curious what you think of this. Here is the previous behavior:

A breadcrumb item that breaks to multiple lines has it's arrow vertically centered to the right of the item link.

Screen Shot 2020-04-01 at 3 15 14 PM

An item that is a very long string won't break and just overflows whatever it is inside of.

Screen Shot 2020-04-01 at 3 15 38 PM

----

And here is the updated behavior. Is this preferred?

Screen Shot 2020-04-01 at 3 16 19 PM

Screen Shot 2020-04-01 at 3 15 58 PM

A side effect of that is that when the container is wide enough to fit the item, but not the arrow, the arrow will break to the next line separately like this

Screen Shot 2020-04-01 at 3 26 44 PM

## Breaking changes
This PR changes the visual appearance of breadcrumbs when they break to multiple lines.
* The text will now break if there is a long string in an item link that is wider than the viewport
* The divider is displayed before the link text, instead of after it
* The divider will also always appear inline with the item text instead of wrapping to a new line in some cases.

Because the divider comes before the link text now, the variable `--pf-c-breadcrumb__item-divider--MarginLeft` has been replaced with `--pf-c-breadcrumb__item-divider--MarginRight`. Any instance of the old variable should be replaced with the new one.

Any reference to the divider element (`.pf-c-breadcrumb__item-divider`) will need to be moved in the markup to come before the breadcrumb link element (`.pf-c-breadcrumb__link`) instead of after it.

### DIFF
--- a/src/patternfly/components/Breadcrumb/examples/Breadcrumb.md
+++ b/src/patternfly/components/Breadcrumb/examples/Breadcrumb.md
@@ -12,21 +12,21 @@ cssPrefix: pf-c-breadcrumb
       {{#> breadcrumb-link}}
         Section home
       {{/breadcrumb-link}}
-      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
     {{/breadcrumb-item}}
     {{#> breadcrumb-item}}
+      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
       {{#> breadcrumb-link}}
         Section title
       {{/breadcrumb-link}}
-      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
     {{/breadcrumb-item}}
     {{#> breadcrumb-item}}
+      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
       {{#> breadcrumb-link}}
         Section title
       {{/breadcrumb-link}}
-      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
     {{/breadcrumb-item}}
     {{#> breadcrumb-item}}
+      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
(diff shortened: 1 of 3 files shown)


christiemolloy:
This is definitely a better solution than what was there before. I wonder if the breadcrumb links + arrow could be width: 100% and then the breadcrumb link is always (e.g. 90%) and then the arrow is (e.g. 10%) and then the text inside the breadcrumb link would truncate. Just an idea

mcoker:
After chatting with @mceledonia and @mcarrano we decided to include the separator before the link instead, which gives us this layout:

Screen Shot 2020-04-07 at 12 27 37 PM

christiemolloy:
This looks good, Would you say its required to have the divider before the links now? Maybe update the md description / breaking change description

mcoker:
@christiemolloy thanks, updated the release notes

src/patternfly/components/Breadcrumb/examples/Breadcrumb.md
```diff
@@ -12,21 +12,21 @@ cssPrefix: pf-c-breadcrumb
       {{#> breadcrumb-link}}
         Section home
       {{/breadcrumb-link}}
-      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
     {{/breadcrumb-item}}
     {{#> breadcrumb-item}}
+      {{#> breadcrumb-item-divider}}{{/breadcrumb-item-divider}}
```
mattnolting, line 17: You can use `{{#> breadcrumb-item-divider}}`

mcoker: updated

PR #2919, opened 2020-04-02
fix(tabs): Update class name for tabs button to be tabs link

### LINKED ISSUE #2594: Update class name in Tab component
Had a conversation on slack yesterday with @mcoker and @mattnolting about the simple list component  #2573. We discussed whether a classname that can either represent a `<button>` or an `<a>` tag, should be called `__button`, `__link` or `__action`. Semantically, `<button>` is used when linking to content within a page, and `<a>` is used when linking to content outside of the page.

For the tab and simple list components, they can either use `<button>` or `<a>` depending on the context. Typically though `<button>` would be used, and a component like the `<nav>` would use `<a>`. However, we came to the conclusion that when implemented in React, the "button" is still **linking**  to new content or a new element, therefore it is more of a "link" than a "button" which performs an action. But for markup purposes a link or button can be used and shouldn't affect the naming.

We should update the tab component accordingly.

christiemolloy:
closes #2594 

## Breaking Changes

* Renames the class `pf-c-tabs__button` to be `pf-c-tabs__link`. Any instances of the old class should be updated to the new class.

* Updates any tabs CSS variable with a reference to the BEM element `__button` to `__link`. Any references to tabs variables with `__button` should be replaced with `__link`.
E.g. `--pf-c-tabs--m-box__button--BackgroundColor` becomes `--pf-c-tabs--m-box__link--BackgroundColor`

### DIFF
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -30,7 +30,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs--m-vertical__link--PaddingBottom: var(--pf-global--spacer--md);
   --pf-c-tabs--m-box__link--BackgroundColor: var(--pf-global--BackgroundColor--200);
 
-  // Button before
+  // Link before
   --pf-c-tabs__link--before--BorderStyle: solid;
   --pf-c-tabs__link--before--BorderColor: var(--pf-c-tabs--before--BorderColor);
   --pf-c-tabs__link--before--BorderWidth: var(--pf-global--BorderWidth--sm);
@@ -43,7 +43,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs__link--before--Left: calc(var(--pf-c-tabs__link--before--BorderWidth) * -1);
   --pf-c-tabs__item--m-current__link--after--BorderColor: var(--pf-global--active-color--100);
 
-  // Button after
+  // Link after
   --pf-c-tabs__link--after--BorderStyle: solid;
   --pf-c-tabs__link--after--BorderColor: var(--pf-global--BorderColor--light-100);
   --pf-c-tabs__link--after--BorderWidth: 0;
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Tabs/tabs.scss
```diff
@@ -15,46 +15,46 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs--m-vertical--m-box--Inset: var(--pf-global--spacer--xl);
 
   // Tab buttons
```
mcoker, line 17:
```suggestion
  // Tab links
```

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -15,46 +15,46 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
+  --pf-c-tabs__link--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-tabs__link--md--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-tabs__link--md--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-tabs__item--m-current__link--Color: var(--pf-global--Color--100);
+  --pf-c-tabs__item--m-current__link--Background: var(--pf-global--BackgroundColor--100);
+  --pf-c-tabs--m-vertical__link--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-tabs--m-vertical__link--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-tabs--m-box__link--BackgroundColor: var(--pf-global--BackgroundColor--200);
 
   // Button before
```
mcoker, line 33:
```suggestion
  // Link before
```

mcoker:
> Updates all instances of --pf-c-tabs__**button**--[property] to --pf-c-tabs__**link**--[property]. Any instances of the old variable should be updated to the new variable.

I think that might be better described as 

> Updates all instances of `--pf-c-tabs**__button**--[property]` to `--pf-c-tabs**__link**--[property]`. Any instances of the old variable should be updated to the new variable.

That would also cover the syntax for something like `--pf-c-tabs--m-vertical__button--PaddingTop`. Though maybe a simpler way of stating it is this?

> Updates any tabs CSS variable with a reference to the BEM element `__button` to `__link`. Any references to tabs variables with `__button` should be replaced with `__link`.

PR #2921, opened 2020-04-03
fix(alerts): update styling for better accessibility

### LINKED ISSUE #2659: Update styling of alerts for better accessibility
Improve the styling of the Alert component to work better visually and be more accessible.

Design here: https://marvelapp.com/11417fd4

cc @mcoker @LHinson

christiemolloy:
closes #2659

## Breaking Changes

* Adds `pf-c-alert__action-group` class. This class should be used to wrap the action buttons at the bottom of the alert.

* Updates the global info color - `$pf-global--info-color--100` from `$pf-color-blue-200` to  `$pf-color-blue-300`. This is a visual breaking change only. No further changes are needed to consume this change. If you were using the global info color in your application this change will visually effect your UI.

* The variable name `--pf-c-alert--grid-template-columns` changed to `--pf-c-alert--GridTemplateColumns`. Any instance of `--pf-c-alert--grid-template-columns` should be replaced with `--pf-c-alert--GridTemplateColumns`

*Removed these variables:
  *  `--pf-c-alert--grid-template-rows`
  * `--pf-c-alert__icon--Padding`
  * `--pf-c-alert__icon--BackgroundColor`
  * `--pf-c-alert__title--PaddingTop`
  * `--pf-c-alert__title--PaddingRight`
  * `--pf-c-alert__title--PaddingBottom`
  * `--pf-c-alert__title--PaddingLeft`
  * `--pf-c-alert__description--PaddingRight`
  * `--pf-c-alert__description--PaddingBottom`
  * `--pf-c-alert__description--PaddingLeft`
  * `--pf-c-alert__description--MarginTop`
  * `--pf-c-alert__action--PaddingTop`
  * `--pf-c-alert__action--PaddingRight`
  *`--pf-c-alert--m-success__icon--BackgroundColor`
  * `--pf-c-alert--m-danger__icon--BackgroundColor`
  * `--pf-c-alert--m-warning__icon--BackgroundColor`
  * `--pf-c-alert--m-warning__icon--FontSize`
  * `--pf-c-alert--m-info__icon--BackgroundColor`
  * `--pf-c-alert--m-inline--BorderColor`
  * `--pf-c-alert--m-inline--BorderStyle`
  * `--pf-c-alert--m-inline--BorderTopWidth`
  * `--pf-c-alert--m-inline--BorderRightWidth`
  * `--pf-c-alert--m-inline--BorderBottomWidth`
  * `--pf-c-alert--m-inline--BorderLeftWidth`
  * `--pf-c-alert--m-inline--before--Width`
  * `--pf-c-alert--m-inline--before--Top`
  * `--pf-c-alert--m-inline--before--Bottom`
  * `--pf-c-alert--m-inline--before--BackgroundColor`
  * `--pf-c-alert--m-inline__icon--FontSize`
  * `--pf-c-alert--m-inline__icon--Color`
  * `--pf-c-alert--m-inline__icon--BackgroundColor`
  * `--pf-c-alert--m-inline__icon--PaddingTop`
  * `--pf-c-alert--m-inline__icon--PaddingRight`
  * `--pf-c-alert--m-inline__icon--PaddingBottom`
  * `--pf-c-alert--m-inline__icon--PaddingLeft`
  * `--pf-c-alert--m-inline--m-warning__icon--FontSize`
  * `--pf-c-alert--m-inline--m-success__icon--Color`
  * `--pf-c-alert--m-inline--m-success--before--BackgroundColor`
  * `--pf-c-alert--m-inline--m-danger__icon--Color`
  * `--pf-c-alert--m-inline--m-danger--before--BackgroundColor`
  * `--pf-c-alert--m-inline--m-warning__icon--Color`
  * `--pf-c-alert--m-inline--m-warning--before--BackgroundColor`
  * `--pf-c-alert--m-inline--m-info__icon--Color`
  * `--pf-c-alert--m-inline--m-info--before--BackgroundColor`


src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
   --pf-c-alert--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-alert--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-alert--grid-template-columns: max-content 1fr max-content;
-  --pf-c-alert--grid-template-rows: 1fr auto;
+  --pf-c-alert--grid-template-rows: var(--pf-global--spacer--lg) auto auto;
```
mcoker, line 6: I don't think you need to specify rows. Giving it a fixed height creates issues if you change the size of anything in that row. Looks like maybe it's there to keep the close button from making that row so tall? The way the alert currently does it works pretty well.

Screen Shot 2020-04-03 at 6 28 56 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
-  --pf-c-alert__description--PaddingBottom: var(--pf-global--spacer--md);
-  --pf-c-alert__description--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-alert__description--MarginTop: calc(-1 * var(--pf-global--spacer--sm));
+  // Alert description
+  --pf-c-alert__description--PaddingTop: var(--pf-global--spacer--xs);
 
   // Action
-  --pf-c-alert__action--PaddingTop: #{pf-size-prem(11px)}; // one off spacer to align properly.
-  --pf-c-alert__action--PaddingRight: var(--pf-global--spacer--xs);
+  --pf-c-alert__action--MarginTop: #{pf-size-prem(-6px)};
```
mcoker, line 28: You can use `var(--pf-global--spacer--form-element)` or `var(--pf-c-button--PaddingTop)` here.

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
+  grid-column: 1 / span 1;
+  margin-top: var(--pf-c-alert__icon--MarginTop);
+  margin-right: var(--pf-c-alert__icon--MarginRight);
   font-size: var(--pf-c-alert__icon--FontSize);
   color: var(--pf-c-alert__icon--Color);
-  background-color: var(--pf-c-alert__icon--BackgroundColor);
 }
 
 .pf-c-alert__title {
+  display: flex;
```
mcoker, line 142: why make this flex? If you put `<span>word</span> <span>word</span>` it will read `wordword`
Screen Shot 2020-04-03 at 6 49 30 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
 
 .pf-c-alert__description {
   grid-area: content;
-  padding-right: var(--pf-c-alert__description--PaddingRight);
-  padding-bottom: var(--pf-c-alert__description--PaddingBottom);
-  padding-left: var(--pf-c-alert__description--PaddingLeft);
+  padding-top: var(--pf-c-alert__description--PaddingTop);
+
+  + .pf-c-alert__action-group {
+    padding-top: var(--pf-c-alert__description--action-group--PaddingTop);
```
mcoker, line 153:
```suggestion
    --pf-c-alert__action-group--PaddingTop: var(--pf-c-alert__description--action-group--PaddingTop);
```

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
   }
 }
 
 .pf-c-alert__icon {
   display: flex;
-  grid-area: icon;
-  padding: var(--pf-c-alert__icon--Padding);
+  grid-row: 1 / span 1;
+  grid-column: 1 / span 1;
+  margin-top: var(--pf-c-alert__icon--MarginTop);
```
mcoker, line 135: Since this is a flex container, what if you used `align-items: center;` instead of a margin? Then everything is still centered if you change the height of the icon/text

Screen Shot 2020-04-03 at 6 57 31 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
+    --pf-c-alert--BackgroundColor: var(--pf-c-alert--m-inline--BackgroundColor);
   }
 }
 
 .pf-c-alert__icon {
   display: flex;
-  grid-area: icon;
-  padding: var(--pf-c-alert__icon--Padding);
+  grid-row: 1 / span 1;
+  grid-column: 1 / span 1;
```
mcoker, line 134: Is there a reason you don't want it to use the `icon` grid area?` Since it's defined and the rest of the sections use `grid-area`, you could modify the template areas and use `.` for the 2nd and 3rd column `icon` areas currently.

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
+  margin-top: var(--pf-c-alert__action--MarginTop);
+  margin-right: var(--pf-c-alert__action--MarginRight);
+}
+
+.pf-c-alert__action-group {
+  grid-area: actiongroup;
+  padding-top: var(--pf-c-alert__action-group--PaddingTop);
+
+  > .pf-c-button {
+    font-weight: var(--pf-c-alert__action-group--c-button--FontWeight);
```
mcoker, line 168:
```suggestion
    font-weight: var(--pf-c-alert__action-group__c-button--FontWeight);
```

mcoker: same for all of the vars that reference `__c-button` as a child

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,174 @@
...
+    "icon actiongroup actiongroup";
+
+  &::before {
+    position: absolute;
+    top: 0;
+    right: 0;
+    left: 0;
+    content: "";
+    border: var(--pf-c-alert--before--BorderTopWidth) solid;
+    border-color: var(--pf-c-alert--before--BorderTopColor);
```
mcoker, line 94: Since you have a single border color declaration, this can all be declared in the `border` shorthand.

src/patternfly/components/Alert/alert.scss
christiemolloy: yeah not sure why that is there , removed

src/patternfly/components/Alert/alert.scss
christiemolloy: oh good point didn't realize you can do `.` nice!

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,166 @@
   --pf-c-alert--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-alert--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-alert--grid-template-columns: max-content 1fr max-content;
-  --pf-c-alert--grid-template-rows: 1fr auto;
+  --pf-c-alert--grid-template-rows: 1fr auto auto;
```
mcoker, line 6: What's the reasoning for `1fr` on the first row?

mcoker:
The presence of the close button makes the title area larger than it needs to be, and also changes the space between the title and whatever is below it when you add/remove the close button.

Screen Shot 2020-04-06 at 4 55 32 PM

Screen Shot 2020-04-06 at 4 57 15 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,166 @@
...
+  box-shadow: var(--pf-c-alert--BoxShadow);
   grid-template-columns: var(--pf-c-alert--grid-template-columns);
   grid-template-rows: var(--pf-c-alert--grid-template-rows);
   grid-template-areas:
     "icon title action"
-    "icon content content";
-  background-color: var(--pf-c-alert--BackgroundColor);
-  box-shadow: var(--pf-c-alert--BoxShadow);
-
+    ". content content"
```
mattnolting, line 81: Icon should span the rows. If icon font-size is increased, spanning all rows will prevent vertical misalignment.

```
"icon title action"
"icon content content"
"icon actiongroup actiongroup"
```
**Before**
Screen Shot 2020-04-07 at 10 05 41 AM

**After**
Screen Shot 2020-04-07 at 10 05 24 AM

src/patternfly/components/Alert/alert.scss
mattnolting: Agree, I don't think it needs to be defined.

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,166 @@
...
+  --pf-c-alert--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-alert--PaddingLeft: var(--pf-global--spacer--md);
 
   // Icon
-  --pf-c-alert__icon--Padding: var(--pf-global--spacer--md);
-  --pf-c-alert__icon--Color: var(--pf-global--default-color--300);
-  --pf-c-alert__icon--BackgroundColor: var(--pf-global--default-color--100);
-  --pf-c-alert__icon--FontSize: var(--pf-global--icon--FontSize--lg);
+  --pf-c-alert__icon--Color: var(--pf-global--default-color--200);
+  --pf-c-alert__icon--MarginRight: var(--pf-global--spacer--sm);
```
mattnolting, line 16: Suggest using `grid-column-gap` instead, with `minmax` column definition.

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,166 @@
   --pf-c-alert--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-alert--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-alert--grid-template-columns: max-content 1fr max-content;
```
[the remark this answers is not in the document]
christiemolloy: nice!

christiemolloy:
> The presence of the close button makes the title area larger than it needs to be, and also changes the space between the title and whatever is below it when you add/remove the close button.
> 
> Screen Shot 2020-04-06 at 4 55 32 PM
> 
> Screen Shot 2020-04-06 at 4 57 15 PM

That's why I added a height to the first row in the grid. I updated it so that it is based on the `font-size * line-height` of the alert title so that it would adjust as the font-size changes. Let me know if you think thats a viable option?

Screen Shot 2020-04-07 at 11 36 53 AM

src/patternfly/components/Alert/alert.scss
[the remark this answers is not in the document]
mcoker: That will work until a point. It's also only specific to core. I did a little digging and looks like FA offers a utility class for [fixed width icons](https://fontawesome.com/how-to-use/on-the-web/styling/fixed-width-icons) with this use case.

mcoker:
I don't think we want to limit the height of the title row, or we'll end up with this, unless we set the text to truncate.

Screen Shot 2020-04-07 at 7 46 50 PM

Also if you use height to define the row, I don't think you'll want to base off of line-height as someone may just leave that and modify the font-size var we have (`var(--pf-c-alert__title--FontSize)`) and then you end up with:

Screen Shot 2020-04-07 at 7 49 27 PM

mcoker:
Also `float` on the icon will let the title text wrap underneath the icon, which we don't want. It's a pattern in other components that if text is positioned adjacent to a close icon, the text doesn't wrap underneath (popover, tooltip, modal, etc).

mcoker:
The easiest way around the spacing problem caused by the button is to give it a negative margin so that the button's padding no longer affects the space at the bottom of the alert title row, the use `translateY()` to move it back down where you want it. I made that update along with some others here - https://github.com/christiemolloy/patternfly-next/pull/23/files

src/patternfly/components/Alert/alert.scss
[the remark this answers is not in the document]
mattnolting: Nice, you could do same thing as `.fa-fw` without creating a dependency by setting a width on `i`, which would make it themeable. Then you could set the first column back to `max-content`.
```
--pf-c-alert--grid-template-columns: max-content 1fr max-content;
--pf-c-alert__icon--icon--Width: 1.25em

.pf-c-alert__icon {
  ...

  i {
    width: var(--pf-c-alert__icon--icon--Width);
    text-align: center;
  }
}
```

christiemolloy:
@mceledonia just wanted to point this out, are the alerts too big relative to the page size? This is what they look like now:

Screen Shot 2020-04-08 at 9 58 13 AM

compared with Marvel: https://marvelapp.com/11417fd4/screen/67688439

src/patternfly/components/Alert/alert.scss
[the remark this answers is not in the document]
mcoker: The downside to that is we're shipping code just to style a core element. And the utility comes with patternfly, so users should have it by default. If they don't have it, it degrades gracefully as the alignment difference is really slight.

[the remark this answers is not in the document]
mattnolting: What if FA doesn't support requirements and the application needs to use custom iconography?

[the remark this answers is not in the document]
mcoker: I would say at that point it's on them to make sure their icons work for this kind of layout. We can't assume they'll be using `<i>` elements, or that this fixed width would work the same way with whatever icons they're using if they are using `<i>`. That said, the `.fa-fw` styles seem generic enough that they likely would work, but we don't know for sure.

And to my earlier point, I think we should avoid core specific styles in the components if at all possible, and it doesn't seem necessary to include this. That said, react and core use different icon systems, so in some cases we will need to account for that, and this is a small utility so it's not a big deal to include it.

mcarrano:
@christiemolloy for the demo showing toast alerts, I'm thinking they should not have actions. Typically we would not use actions like this in a toast. That would reduce the vertical space.

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,171 @@
...
+  --pf-c-alert--m-inline--m-info--BackgroundColor: var(--pf-global--palette--blue-50);
 
   @include pf-t-light; // This component always needs to be light
 
   position: relative;
   display: grid;
+  padding: var(--pf-c-alert--PaddingTop) var(--pf-c-alert--PaddingRight) var(--pf-c-alert--PaddingBottom) var(--pf-c-alert--PaddingLeft);
+  background-color: var(--pf-c-alert--BackgroundColor);
+  box-shadow: var(--pf-c-alert--BoxShadow);
   grid-template-columns: var(--pf-c-alert--grid-template-columns);
```
mcoker, line 80: Just noticed these var names are wrong. We could also specify vars for the column widths the way I did with the form component PR, so users don't have to override all of `grid-template-columns` just to change a column width. And that allows us to update `grid-template-columns` with less likelihood of it being breaking.

```suggestion
  grid-template-columns: var(--pf-c-alert--GridTemplateColumns);
```

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,171 @@
...
-  grid-template-rows: var(--pf-c-alert--grid-template-rows);
   grid-template-areas:
     "icon title action"
-    "icon content content";
-  background-color: var(--pf-c-alert--BackgroundColor);
-  box-shadow: var(--pf-c-alert--BoxShadow);
-
+    "icon description description"
+    "icon actiongroup actiongroup";
+  grid-column-gap: var(--pf-c-alert--grid-column-gap);
```
mcoker, line 85:
```suggestion
  grid-column-gap: var(--pf-c-alert--GridColumnGap);
```

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,171 @@
   --pf-c-alert--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-alert--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-alert--grid-template-columns: max-content 1fr max-content;
-  --pf-c-alert--grid-template-rows: 1fr auto;
+  --pf-c-alert--grid-column-gap: var(--pf-global--spacer--xs);
```
mcoker, line 6: renamed the var and bumped the value - the current gap is a little tight.

```suggestion
  --pf-c-alert--GridColumnGap: var(--pf-global--spacer--sm);
```

src/patternfly/components/Alert/alert.scss
```diff
@@ -3,215 +3,171 @@
   --pf-c-alert--BoxShadow: var(--pf-global--BoxShadow--lg);
   --pf-c-alert--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-alert--grid-template-columns: max-content 1fr max-content;
```
mcoker, line 5:
```suggestion
  --pf-c-alert--GridTemplateColumns: max-content 1fr max-content;
```

Or `AlertColumnWidth`, `TitleColumnWidth`, `ActionColumnWidth`

src/patternfly/components/Alert/alert.scss
```diff
@@ -2,216 +2,172 @@
...
-  --pf-c-alert__icon--Color: var(--pf-global--default-color--300);
-  --pf-c-alert__icon--BackgroundColor: var(--pf-global--default-color--100);
-  --pf-c-alert__icon--FontSize: var(--pf-global--icon--FontSize--lg);
+  --pf-c-alert__icon--Color: var(--pf-global--default-color--200);
+  --pf-c-alert__icon--MarginTop: calc(var(--pf-global--spacer--xs) / 2);
+  --pf-c-alert__icon--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-alert__icon--FontSize: var(--pf-global--icon--FontSize--md);
 
   // Title
   --pf-c-alert__title--FontSize: var(--pf-global--FontSize--md);
```
mcoker, line 20: Can you think of a reason to set this if it's the default font size?

src/patternfly/components/Alert/alert.scss
```diff
@@ -2,216 +2,172 @@
...
+  --pf-c-alert__title--LineHeight: var(--pf-global--LineHeight--md);
```
mcoker, line 21: looks like this isn't in use.

src/patternfly/components/Alert/alert.scss
christiemolloy: I was using it to calculate the height, i dont need anymore good pick up

christiemolloy:
Yep, looks good! @mcoker

mattnolting:
Release notes verified

PR #2923, opened 2020-04-03
fix(datalist): remove flex wrap for action button

### LINKED ISSUE #2029: Buttons in the Data List component
I was reviewing the data list component....
@mcarrano is it true that the recent addition of the `pf-hidden-visible-{breakpoints}` means that buttons should collapse to a kebab before they stack. That's how it looks, at 992px (lg breakpoint) width that looks like this, but it seems unintentional i think because of the min-max issue we've seen before:

Screen Shot 2019-07-09 at 2 19 25 PM

If so, I think we can remove the following classes from the data list component when the breaking change period happens, or maybe just when we make a layout for buttons: cc: @mcoker 

```
  --pf-c-data-list__item-action--not-last-child--MarginRight: var(--pf-global--spacer--md);
  --pf-c-data-list__item-action--not-last-child--MarginBottom: var(--pf-global--spacer--md);
  --pf-c-data-list__item-action--not-last-child--lg--MarginBottom: 0;
```

christiemolloy:
closes #2029

## Breaking Changes
* Removes `--pf-c-data-list__item-action--not-last-child--MarginBottom` all instances of it should be removed from your application.

* Removes `flex-wrap: wrap` from the data list actions. When the list of buttons get too long you should use the overflow menu to hide buttons, instead of flex-wrap, as we don't support this anymore.

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -99,7 +99,6 @@
   --pf-c-data-list__item-action--MarginLeft: var(--pf-global--spacer--md);
   --pf-c-data-list__item-action--md--MarginLeft: var(--pf-global--spacer--xl);
   --pf-c-data-list__item-action--not-last-child--MarginRight: var(--pf-global--spacer--md);
-  --pf-c-data-list__item-action--not-last-child--MarginBottom: var(--pf-global--spacer--md);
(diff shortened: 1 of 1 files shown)


mcoker:
@mattnolting that's a separate task. I'm fine with it being a demo in the react workspace. Jotted down some ideas here - https://github.com/patternfly/patternfly/issues/2926

Issue #1959, opened 2019-06-19
Investigate use of .pf-c-form__helper-text-icon in login component


mcoker:
https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/components/Login/login.scss#L214-L217

`.pf-c-form__helper-text-icon` doesn't exist in the form component. Should we add it? Maybe update the class to be an element of the login component instead?

mcoker:
@mcarrano we have this icon in the form helper text on the login example. This currently only works in the login component. Do you see any reason not to move support this icon and its vertical alignement/spacing into the form component, so that it can be used in the context of any form (not just on the login screen)?

mcarrano:
Is this what you are referring to @mcoker ?

![Screen Shot 2020-04-06 at 9 44 03 AM](https://user-images.githubusercontent.com/6619860/78564973-36e1a980-77eb-11ea-96e7-50e883dceeab.png)

If so, I don't see any harm in that, although we don't have any specific pattern that requires it. I think this was unique to login because we did not want to use an inline alert to conserve space.

mcoker:
@mcarrano that's it, and thanks!

PR #2928, opened 2020-04-06
fix(form): moved helper text icon from login to form component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/1959

## Breaking changes
This PR moves the form helper text icon feature from the login component back into the form component.

* The `.pf-c-form__helper-text-icon` styles have moved from the login component stylesheet to the form component stylesheet
* Removes the following variables from the login component and replaces the variables in the form component. If you're overriding any of these using the `.pf-c-login` component selector, you will need to update the selector to `.pf-c-form`.
  * `--pf-c-login__main-body--c-form__helper-text-icon--FontSize` has been renamed to `--pf-c-form__helper-text-icon--FontSize`
  * `--pf-c-login__main-body--c-form__helper-text-icon--MarginRight` has been renamed to `--pf-c-form__helper-text-icon--MarginRight`

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -91,10 +91,23 @@ cssPrefix: pf-c-form
     {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-comment"')}}
       Comment
     {{/form-label}}
-    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute=(concat 'value="This is a valid comment"' 'type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
+    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute=(concat 'value="This is a valid comment" type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
     {{/form-control}}
     {{#> form-helper-text form-helper-text--modifier="pf-m-success" form-helper-text--attribute=(concat 'id="' form--id '-simple-form-comment-help" aria-live="polite"')}}
       This is helper text for success input
+    {{/form-helper-text}}
+  {{/form-group}}
+  {{#> form-group}}
+    {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-info"')}}
+      Information
+    {{/form-label}}
+    {{#> form-control controlType="textarea" form-control--attribute=(concat 'id="' form--id '-simple-form-info" name="' form--id '-simple-form-info" aria-invalid="true" aria-describedby="' form--id '-simple-form-info-helper"')}}
+    {{/form-control}}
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Form/examples/Form.md
```diff
@@ -91,12 +91,25 @@ cssPrefix: pf-c-form
...
       This is helper text for success input
     {{/form-helper-text}}
   {{/form-group}}
+  {{#> form-group}}
+    {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-info"')}}
+      Information
+    {{/form-label}}
+    {{#> form-control controlType="textarea" form-control--attribute=(concat 'id="' form--id '-simple-form-info" name="' form--id '-simple-form-info" aria-invalid="true" aria-describedby="' form--id '-simple-form-info-helper"')}}
+    {{/form-control}}
+    {{#> form-helper-text form-helper-text--modifier="pf-m-error" form-helper-text--attribute=(concat 'id="' form--id '-simple-form-info-helper" aria-live="polite"')}}
```
christiemolloy, line 106: Could you update the handlebars file for `form-helper-text` so that if `HasIcon` is true, it adds the icon to the helper text.

mcoker: I don't think it always needs to be one kind of icon though, it can be anything. So you'd have to pass `form-helper-text-icon--type` or something, too, to define the icon type, but that gets tricky as you might want to pass something custom that isn't a `fas` icon. ^ is how we do this for most other components where you can pass a dynamic icon.

christiemolloy: Okay that makes sense

src/patternfly/components/Form/examples/Form.md
```diff
@@ -91,12 +91,25 @@ cssPrefix: pf-c-form
     {{#> form-label form-label--attribute=(concat 'for="' form--id '-simple-form-comment"')}}
       Comment
     {{/form-label}}
-    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute=(concat 'value="This is a valid comment"' 'type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
+    {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute=(concat 'value="This is a valid comment" type="text" id="' form--id '-simple-form-comment" name="' form--id '-simple-form-comment" aria-describedby="' form--id '-simple-form-comment-help"')}}
```
mattnolting, line 94: Niiice!!

mcoker:
> This is the same pattern as the pf-c-button updates. Is the icon only going to be appear first? If not, we should add .pf-m-start and .pf-m-end to apply margin.

This is basically a small version of an inline alert, so I'm assuming we would just follow the alert layout with icon + text.

@mcarrano - would we ever support having both icon + text and text + icon as part of the helper text layout like we do with buttons? Or would this always be icon + text? And if a user wanted to have text + icon, they could but they would have to use their own kind of layout to do that.

mcarrano:
@mcoker @christiemolloy let's keep this simple. It should always be an icon + text. I don't see any reason to support the reverse. As you say, someone could create that if they need to with a custom layout.

christiemolloy:
Release notes look good

Issue #2096, opened 2019-07-24
Update content component elements' top/bottom margin for first/last children


mcoker:
For example, this shouldn't have a bottom margin.

<img width="1189" alt="Screen Shot 2019-07-24 at 8 11 52 AM" src="https://user-images.githubusercontent.com/35148959/61796301-b6a91e80-adea-11e9-85a3-55e879440077.png">

And this shouldn't have a top margin

<img width="1145" alt="Screen Shot 2019-07-24 at 8 13 42 AM" src="https://user-images.githubusercontent.com/35148959/61796417-f96af680-adea-11e9-9c5a-03936f6417c3.png">

mcoker:
closed via https://github.com/patternfly/patternfly/pull/2930

Issue #1842, opened 2019-05-20
h1-h6 content component remove bottom margin


christiemolloy:
Remove the bottom margin from the heading elements in the Content Component when they're the last child. This goes along with our rule that components should touch all edges of their parent.

Also check to see if other elements in the component need the bottom margin removed if they're the last child.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/2930

PR #2930, opened 2020-04-06
fix(content): margin should not be on first-child or last-child

### LINKED ISSUE #1844: Remove margin bottom from small element in content
In the next breaking change period we should create a PR to remove the margin-bottom variable from the small element, as it is being handled in lines 117-130.

Relates to #1843

christiemolloy:
closes #1844 #2096 #1842

## Breaking Changes
* If you are using `pf-c-content` in your application you may need to check to see if this change impacts the UI. The margin-bottom may now be removed if using h1-h6, and the margin-bottom may now be removed if using `<ol>` and `<ul>`.

* Removes the margin-bottom from `h1-h6` if its the last child. 
* Removes the margin-top from `<ol>` and `<ul>` if its the first-child.

### DIFF
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -58,7 +58,6 @@
   --pf-c-content--small--LineHeight: var(--pf-global--LineHeight--md);
   --pf-c-content--small--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-content--small--Color: var(--pf-global--Color--200);
-  --pf-c-content--small--FontWeight: var(--pf-global--FontWeight--semi-bold);
 
   // Links
   --pf-c-content--a--Color: var(--pf-global--link--Color);
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -138,6 +138,14 @@
   h6 {
     margin: 0;
     font-family: var(--pf-c-content--heading--FontFamily);
+
+    &:first-child {
+      margin-top: 0;
+    }
+
+    &:last-child {
+      margin-bottom: 0;
+    }
   }
 
   ol,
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Content/content.scss
```diff
@@ -239,9 +257,12 @@
 
   ol {
     padding-left: var(--pf-c-content--ol--PaddingLeft);
-    margin-top: var(--pf-c-content--ol--MarginTop);
     margin-left: var(--pf-c-content--ol--MarginLeft);
 
+    &:not(:first-child) {
+      margin-top: var(--pf-c-content--ol--MarginTop);
```
mcoker, line 263: Can you think of a reason why we would want this?  One thing this does is cause a heading + `<ol>` (or `<ul>`) have a `16px` spacer where a heading + `<p>` will have an `8px` spacer. Do we want that?

It would also allow users to use a different spacer above lists by styling the list, versus having to style whatever it is that precedes the list. But we don't allow that with the other elements - `<p>`'s, for example.

src/patternfly/components/Content/content.scss
```diff
@@ -255,10 +276,13 @@
 
   ul {
     padding-left: var(--pf-c-content--ul--PaddingLeft);
-    margin-top: var(--pf-c-content--ul--MarginTop);
     margin-left: var(--pf-c-content--ul--MarginLeft);
     list-style: var(--pf-c-content--ul--ListStyle);
 
+    &:not(:first-child) {
```
mcoker, line 282: Since we've wrapped these `margin` declarations in `:not(:first-child)`, the styles below where we set a custom margin for nested lists isn't working since a nested list likely will be the first/only child.

Screen Shot 2020-04-06 at 6 43 53 PM

mcoker:
@christiemolloy left some comments for review. If you haven't already, we should go over these styles with design and ensure we match their expectations.

src/patternfly/components/Content/content.scss
christiemolloy: Agreed I'm not sure if we even need the margin-top because any of the elements in `pf-c-content` will have a margin-bottom if they're not the last element.

@mcarrano @kybaker @mceledonia do any of you remember why there was a margin-top of 16px set on the `<ol>` and `<ul>` when there is already a margin-bottom of 8px set on any element on top of it. Do you know if there was a reason behind having larger spacing on top of the lists?

src/patternfly/components/Content/content.scss
christiemolloy: nice catch!

src/patternfly/components/Content/content.scss
christiemolloy: @mcoker @mattnolting after speaking offline with Matt and Michael, we decided that `<ol>` and `<ul>` would only be placed after a `heading` or a `<p>` element therefore the margin-bottom on the later elements would be sufficient spacing. (So we can remove the margin-top  on `<ol>` and `<ul>`). They also said that they would prefer to see 8px space between header and `<ol>` and `<ul>` elements so we don't need to introduce a new variable to replace the margin-top vars which had a size of 16px.

src/patternfly/components/Content/content.scss
```diff
@@ -148,69 +146,87 @@
...
-    margin-bottom: var(--pf-c-content--h1--MarginBottom);
     font-size: var(--pf-c-content--h1--FontSize);
     font-weight: var(--pf-c-content--h1--FontWeight);
     line-height: var(--pf-c-content--h1--LineHeight);
 
     &:not(:first-child) {
       margin-top: var(--pf-c-content--h1--MarginTop);
     }
+
+    &:not(:last-child) {
```
mattnolting, line 157: What if you set
```
  h1 {
    margin-top: var(--pf-c-content--h1--MarginTop);
    margin-bottom: var(--pf-c-content--h1--MarginBottom);
  }

  h2 {
    margin-top: var(--pf-c-content--h2--MarginTop);
    margin-bottom: var(--pf-c-content--h2--MarginBottom);
  }

  h1,
  h2,
  h3,
  h4 {
    &:first-child {
      margin-top: 0;
    }

    &:last-child {
      margin-bottom: 0;
    }
  }
```

christiemolloy: That would work, nice idea

christiemolloy: Is there any downside to setting it and then removing it, no right

mcoker:
doesnt look like this line is used https://github.com/patternfly/patternfly/blob/3390e4fae80388f7dd0cf4da60162a3855d14f22/src/patternfly/components/Content/content.scss#L61

src/patternfly/components/Content/content.scss
mattnolting: Not that I can see:D

christiemolloy:
removed that unused var @mcoker

Issue #1828, opened 2019-05-15
Rename expandable component


ncameronbritt:
This is a very petty issue, but almost all (if not all) components have names that are nouns, whereas ["expandable"](https://www.patternfly.org/v4/documentation/react/components/expandable/) is an adjective (expandable what?). IMO this makes the naming of this pattern inconsistent with the rest of PF. 

Alternatives: "Expandable section." In PF3 this pattern was called "Expand/Collapse Section." "Accordion" or "accordion section" also work. 

I take no position on whether or not the component should be renamed in the code, but I think we should use a noun to refer to it in documentation.

rachael-phillips:
@mcarrano making you aware. What are your thoughts on renaming this component?

mcarrano:
I'd be fine with that as I agree with @ncameronbritt that the current naming is inconsistent with our standard. I believe this change needs to be made in the code so not sure of the effect.

PR #2937, opened 2020-04-08
fix(expandable): renamed expandable to expandable section

mcoker:
fixes https://github.com/patternfly/patternfly/issues/1828

## Breaking changes
This PR renames the expandable component to expandable section. As result, all of the names for elements, variables, modifiers, component classes, etc have changed to reference "expandable-section" instead of "expandable."

Any instance of the old name in an application should be updated to reference the new name. See the full list of changes for details - https://github.com/patternfly/patternfly/pull/2937/files


christiemolloy:
@mcarrano do you mind re-reviewing this one?

PR #2941, opened 2020-04-09
fix(chip): refactor styles

### LINKED ISSUE #2657: Revise Chip styling
Want to revisit styling on the Chip component.

Design updated for Chip can be found here: https://marvelapp.com/ac34bid/screen/67803286 This basically entails updating the border styling and close button inside the chip. Otherwise all of the same variants are supported. This document provides additional information on Chips used in context. https://docs.google.com/document/d/1bmsgR28EdJ1FsroTBn0crxBB4xdiGPkVGpQ__yaWeG4/edit?usp=sharing

cc @mcoker @LHinson

christiemolloy:
closes #2657 

## Breaking Changes
* The read-only modifier: `.pf-m-read-only` was removed as you can achieve the same result with the plain chip. With this the following variables were removed, all instances of them should be removed from your application:
  * `--pf-c-chip--m-read-only--PaddingTop`
  * `--pf-c-chip--m-read-only--PaddingRight`
  *  `--pf-c-chip--m-read-only--PaddingBottom`

* The following variables were removed from the overflow chip component, all instances of them should be removed from your application:
  * `--pf-c-chip--m-overflow--BackgroundColor`
  * `--pf-c-chip--m-overflow--PaddingLeft`
  * ` --pf-c-chip--m-overflow--BorderWidth`
  * `--pf-c-chip--m-overflow--c-button--BorderRadius`
  * `--pf-c-chip--m-overflow--c-button--BorderWidth`
  * ` --pf-c-chip--m-overflow--c-button--PaddingLeft`
  * `--pf-c-chip--m-overflow--c-button--PaddingRight`
  *  `--pf-c-chip--m-overflow--c-button--hover--BorderWidth`
  * `--pf-c-chip--m-overflow--c-button--active--BorderWidth`
  * `--pf-c-chip--m-overflow--c-button--focus--BorderWidth`

* The following variables were renamed from the overflow chip component:
  * `--pf-c-chip--m-overflow--c-button--hover--BorderColor` to `--pf-c-chip--m-overflow--hover--before--BorderColor`
  * `--pf-c-chip--m-overflow--c-button--active--BorderColor` to `--pf-c-chip--m-overflow--active--before--BorderColor`
  * `--pf-c-chip--m-overflow--c-button--focus--BorderColor` to `--pf-c-chip--m-overflow--focus--before--BorderColor`

* `--pf-c-chip--BorderColor` replaced `var(--pf-global--secondary-color--100)` with `var(--pf-global--BackgroundColor--200)`. This is just a visual change, no updates are needed.

### DIFF
--- a/src/patternfly/components/Chip/chip.scss
+++ b/src/patternfly/components/Chip/chip.scss
@@ -9,9 +9,9 @@
   --pf-c-chip--BorderRadius: var(--pf-global--BorderRadius--sm);
 
   // overflow chip button
-  --pf-c-chip--m-overflow--c-button--hover--BorderColor: var(--pf-global--secondary-color--100);
-  --pf-c-chip--m-overflow--c-button--active--BorderColor: var(--pf-global--secondary-color--100);
-  --pf-c-chip--m-overflow--c-button--focus--BorderColor: var(--pf-global--secondary-color--100);
+  --pf-c-chip--m-overflow--hover--before--BorderColor: var(--pf-global--secondary-color--100);
+  --pf-c-chip--m-overflow--active--before--BorderColor: var(--pf-global--secondary-color--100);
+  --pf-c-chip--m-overflow--focus--before--BorderColor: var(--pf-global--secondary-color--100);
 
   // chip text
   --pf-c-chip__text--FontSize: var(--pf-global--FontSize--xs);
@@ -57,21 +57,21 @@
     &.pf-m-hover,
     &:hover {
       &::before {
-        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--c-button--hover--BorderColor);
+        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--hover--before--BorderColor);
       }
     }
 
     &.pf-m-focus,
     &:focus {
       &::before {
-        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--c-button--focus--BorderColor);
+        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--focus--before--BorderColor);
       }
     }
 
     &.pf-m-active,
     &:active {
       &::after {
-        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--c-button--active--BorderColor);
+        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--active--before--BorderColor);
       }
     }
   }
--- a/src/patternfly/components/Chip/chip.scss
+++ b/src/patternfly/components/Chip/chip.scss
@@ -1,31 +1,17 @@
 .pf-c-chip {
+  --pf-c-chip--PaddingTop: var(--pf-global--spacer--form-element);
+  --pf-c-chip--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-chip--PaddingBottom: var(--pf-global--spacer--form-element);
   --pf-c-chip--PaddingLeft: var(--pf-global--spacer--sm);
   --pf-c-chip--BackgroundColor: var(--pf-global--Color--light-100);
-  --pf-c-chip--BorderColor: var(--pf-global--secondary-color--100);
+  --pf-c-chip--BorderColor: var(--pf-global--BackgroundColor--200);
   --pf-c-chip--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-chip--BorderRadius: var(--pf-global--BorderRadius--sm);
 
-  // overflow chip
-  --pf-c-chip--m-overflow--BackgroundColor: var(--pf-global--BackgroundColor--200);
-  --pf-c-chip--m-overflow--PaddingLeft: 0;
-  --pf-c-chip--m-overflow--BorderWidth: 0;
-
   // overflow chip button
-  --pf-c-chip--m-overflow--c-button--BorderRadius: var(--pf-global--BorderRadius--sm);
-  --pf-c-chip--m-overflow--c-button--BorderWidth: 0;
-  --pf-c-chip--m-overflow--c-button--PaddingLeft: var(--pf-global--spacer--sm);
-  --pf-c-chip--m-overflow--c-button--PaddingRight: var(--pf-global--spacer--sm);
-  --pf-c-chip--m-overflow--c-button--hover--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-chip--m-overflow--c-button--hover--BorderColor: var(--pf-global--secondary-color--100);
-  --pf-c-chip--m-overflow--c-button--active--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-chip--m-overflow--c-button--active--BorderColor: var(--pf-global--secondary-color--100);
-  --pf-c-chip--m-overflow--c-button--focus--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-chip--m-overflow--c-button--focus--BorderColor: var(--pf-global--secondary-color--100);
-
-  // read-only chip
-  --pf-c-chip--m-read-only--PaddingTop: var(--pf-global--spacer--form-element);
-  --pf-c-chip--m-read-only--PaddingRight: var(--pf-global--spacer--sm);
-  --pf-c-chip--m-read-only--PaddingBottom: var(--pf-global--spacer--form-element);
 
   // chip text
   --pf-c-chip__text--FontSize: var(--pf-global--FontSize--xs);
@@ -35,6 +21,9 @@
   // chip button
   --pf-c-chip--c-button--PaddingLeft: var(--pf-global--spacer--sm);
   --pf-c-chip--c-button--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-chip--c-button--MarginTop: calc(var(--pf-c-chip--PaddingTop) * -1);
+  --pf-c-chip--c-button--MarginRight: calc(var(--pf-c-chip--PaddingRight) / 2 * -1);
+  --pf-c-chip--c-button--MarginBottom: calc(var(--pf-c-chip--PaddingBottom) * -1);
   --pf-c-chip--c-button--FontSize: var(--pf-global--FontSize--xs);
 
   // chip badge
@@ -45,7 +34,7 @@
   position: relative;
   display: inline-flex;
   align-items: center;
-  padding-left: var(--pf-c-chip--PaddingLeft);
+  padding: var(--pf-c-chip--PaddingTop) var(--pf-c-chip--PaddingRight) var(--pf-c-chip--PaddingBottom) var(--pf-c-chip--PaddingLeft);
   list-style: none;
   background-color: var(--pf-c-chip--BackgroundColor);
   border-radius: var(--pf-c-chip--BorderRadius);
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Chip/chip.scss
```diff
@@ -1,31 +1,17 @@
 .pf-c-chip {
+  --pf-c-chip--PaddingTop: var(--pf-global--spacer--form-element);
+  --pf-c-chip--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-chip--PaddingBottom: var(--pf-global--spacer--form-element);
   --pf-c-chip--PaddingLeft: var(--pf-global--spacer--sm);
   --pf-c-chip--BackgroundColor: var(--pf-global--Color--light-100);
-  --pf-c-chip--BorderColor: var(--pf-global--secondary-color--100);
+  --pf-c-chip--BorderColor: var(--pf-global--BackgroundColor--200);
```
mcoker, line 7: 👍

mcoker, line 7: should this be `--pf-global--BorderColor--300`?

src/patternfly/components/Chip/chip.scss
```diff
@@ -63,63 +52,39 @@
...
-  }
-
   // Button
   .pf-c-button {
-    padding-right: var(--pf-c-chip--c-button--PaddingRight);
-    padding-left: var(--pf-c-chip--c-button--PaddingLeft);
+    margin-top: var(--pf-c-chip--c-button--MarginTop);
+    margin-right: var(--pf-c-chip--c-button--MarginRight);
+    margin-bottom: var(--pf-c-chip--c-button--MarginBottom);
     font-size: var(--pf-c-chip--c-button--FontSize);
```
mcoker, line 84:
```suggestion
    --pf-c-button--FontSize: var(--pf-c-chip--c-button--FontSize);
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -63,63 +52,39 @@
...
+
+    --pf-c-button--PaddingRight: var(--pf-c-chip--c-button--PaddingRight);
+    --pf-c-button--PaddingLeft: var(--pf-c-chip--c-button--PaddingLeft);
   }
```
mcoker, line 88: I wonder if we should add `position: relative` to the `.pf-c-chip__text` element so that you can interact with it (like select the text). We could also achieve the border effect with an actual `border` instead of a pseudo element (which would allow you to interact with the text), though I think the pseudo element is a little more flexible, and makes it easier to make future changes like changing the border width on hover. WDYT? If we want to change to a `border`, now would be good to do it. If not, we could add `position` or sort that out later as the current chip is has this issue so it's not pressing.

mcoker:
can you update the docs to add `<button>` as an element you can apply `.pf-c-chip` to?

mcoker:
Should we update the docs so that `.pf-m-overflow` applies to `button.pf-c-chip`?

mcoker:
need to remove `.pf-m-read-only` from the docs

src/patternfly/components/Chip/chip.scss
mcoker: Also can we make sure the border vars are named from what they apply to? Currently they apply to `::before` but don't reference that in the name.

src/patternfly/components/Chip/chip.scss
```diff
@@ -63,63 +52,39 @@
...
   // Button
   .pf-c-button {
-    padding-right: var(--pf-c-chip--c-button--PaddingRight);
-    padding-left: var(--pf-c-chip--c-button--PaddingLeft);
+    margin-top: var(--pf-c-chip--c-button--MarginTop);
+    margin-right: var(--pf-c-chip--c-button--MarginRight);
+    margin-bottom: var(--pf-c-chip--c-button--MarginBottom);
     font-size: var(--pf-c-chip--c-button--FontSize);
+
+    --pf-c-button--PaddingRight: var(--pf-c-chip--c-button--PaddingRight);
```
mcoker, line 86: Can you move these vars to the top of the block? Var declarations should be placed before property declarations.

src/patternfly/components/Chip/chip.scss
```diff
@@ -35,6 +21,9 @@
   // chip button
   --pf-c-chip--c-button--PaddingLeft: var(--pf-global--spacer--sm);
   --pf-c-chip--c-button--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-chip--c-button--MarginTop: calc(var(--pf-c-chip--PaddingTop) * -1);
```
mcoker, line 24: Can you rename these to `--pf-c-chip__c-button--`?

src/patternfly/components/Chip/chip.scss
christiemolloy: oh yep nice catch

src/patternfly/components/Chip/chip.scss
christiemolloy: I think we should keep the pseudo element for the border because i think its reasonable that design might ask for the border width to increase on H/A/F. So I just added `position: relative` to the text but you can only highlight text in the `read-only` chip now.. also updated the name

src/patternfly/components/Chip/chip.scss
```diff
@@ -57,81 +46,58 @@
...
-      &:hover {
-        // stylelint-disable-next-line
-        &::after {
-          --pf-c-chip--m-overflow--c-button--BorderWidth: var(--pf-c-chip--m-overflow--c-button--hover--BorderWidth);
-          --pf-c-chip--m-overflow--c-button--BorderColor: var(--pf-c-chip--m-overflow--c-button--hover--BorderColor);
-        }
+    &.pf-m-hover,
+    &:hover {
+      &::before {
+        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--hover--before--BorderColor);
```
mcoker, line 60:
```suggestion
        --pf-c-chip--before--BorderColor: var(--pf-c-chip--m-overflow--hover--before--BorderColor);
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -57,81 +46,58 @@
...
-      &:active {
-        // stylelint-disable-next-line
-        &::after {
-          --pf-c-chip--m-overflow--c-button--BorderWidth: var(--pf-c-chip--m-overflow--c-button--active--BorderWidth);
-          --pf-c-chip--m-overflow--c-button--BorderColor: var(--pf-c-chip--m-overflow--c-button--active--BorderColor);
-        }
+    &.pf-m-focus,
+    &:focus {
+      &::before {
+        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--focus--before--BorderColor);
```
mcoker, line 67:
```suggestion
        --pf-c-chip--before--BorderColor: var(--pf-c-chip--m-overflow--focus--before--BorderColor);
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -57,81 +46,58 @@
...
-      &:focus {
-        // stylelint-disable-next-line
-        &::after {
-          --pf-c-chip--m-overflow--c-button--BorderWidth: var(--pf-c-chip--m-overflow--c-button--focus--BorderWidth);
-          --pf-c-chip--m-overflow--c-button--BorderColor: var(--pf-c-chip--m-overflow--c-button--focus--BorderColor);
-        }
+    &.pf-m-active,
+    &:active {
+      &::after {
+        --pf-c-chip--BorderColor: var(--pf-c-chip--m-overflow--active--before--BorderColor);
```
mcoker, line 74:
```suggestion
        --pf-c-chip--before--BorderColor: var(--pf-c-chip--m-overflow--active--before--BorderColor);
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -1,51 +1,40 @@
...
-  --pf-c-chip--m-overflow--c-button--active--BorderWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-chip--m-overflow--c-button--active--BorderColor: var(--pf-global--secondary-color--100);
-  --pf-c-chip--m-overflow--c-button--focus--BorderWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-chip--m-overflow--c-button--focus--BorderColor: var(--pf-global--secondary-color--100);
-
-  // read-only chip
-  --pf-c-chip--m-read-only--PaddingTop: var(--pf-global--spacer--form-element);
-  --pf-c-chip--m-read-only--PaddingRight: var(--pf-global--spacer--sm);
-  --pf-c-chip--m-read-only--PaddingBottom: var(--pf-global--spacer--form-element);
+  --pf-c-chip--m-overflow--hover--before--BorderColor: var(--pf-global--secondary-color--100);
```
mcoker, line 12: I'm not sure about the use of `secondary-color--100`. That's black-600. Should we use `pf-global--BorderColor--200` instead (black-500) so it's more inline with our use of border colors?

christiemolloy: good point.

@mceledonia does this look ok to you?

Screen Shot 2020-04-13 at 1 43 35 PM

src/patternfly/components/Chip/chip.scss
```diff
@@ -1,54 +1,42 @@
 .pf-c-chip {
+  --pf-c-chip--PaddingTop: var(--pf-global--spacer--form-element);
```
mattnolting, line 2: Why `--pf-global--spacer--form-element`?

src/patternfly/components/Chip/chip.scss
```diff
@@ -57,81 +45,58 @@
     bottom: 0;
     left: 0;
     content: "";
-    border: var(--pf-c-chip--BorderWidth) solid var(--pf-c-chip--BorderColor);
-    border-radius: var(--pf-c-chip--BorderRadius);
+    border: var(--pf-c-chip--before--BorderWidth) solid var(--pf-c-chip--before--BorderColor);
+    border-radius: var(--pf-c-chip--before--BorderRadius);
```
mattnolting, line 49: `pointer-events: none` will ensure this doesn't interfere with selectability.

src/patternfly/components/Chip/chip.scss
```diff
@@ -1,54 +1,42 @@
...
-  --pf-c-chip--m-read-only--PaddingRight: var(--pf-global--spacer--sm);
-  --pf-c-chip--m-read-only--PaddingBottom: var(--pf-global--spacer--form-element);
+  --pf-c-chip--m-overflow--hover--before--BorderColor: var(--pf-global--BorderColor--200);
+  --pf-c-chip--m-overflow--active--before--BorderColor: var(--pf-global--BorderColor--200);
+  --pf-c-chip--m-overflow--focus--before--BorderColor: var(--pf-global--BorderColor--200);
 
   // chip text
   --pf-c-chip__text--FontSize: var(--pf-global--FontSize--xs);
   --pf-c-chip__text--Color: var(--pf-global--Color--100);
   --pf-c-chip__text--MaxWidth: #{pf-size-prem(120px)};
```
mattnolting, line 19: Maybe outside the scope of this issue, but what do you think about changing `--pf-c-chip__text--MaxWidth` to a character width, say `20ch` and/or making that a global variable? 

Reason being that if the font-size of `pf-c-chip__text` is increased, the string length doesn't adjust to that change and results in this. 
Screen Shot 2020-04-13 at 4 11 24 PM

Where a `ch` value would result in this
Screen Shot 2020-04-13 at 4 11 37 PM

src/patternfly/components/Chip/chip.scss
```diff
@@ -57,81 +45,58 @@
...
+    border: 0;
 
-      &.pf-m-hover,
-      &:hover {
-        // stylelint-disable-next-line
-        &::after {
-          --pf-c-chip--m-overflow--c-button--BorderWidth: var(--pf-c-chip--m-overflow--c-button--hover--BorderWidth);
-          --pf-c-chip--m-overflow--c-button--BorderColor: var(--pf-c-chip--m-overflow--c-button--hover--BorderColor);
-        }
+    &.pf-m-hover,
```
mattnolting, line 56: Can we remove the workspace modifiers now??

src/patternfly/components/Chip/chip.scss
christiemolloy: That's a good point, i think its outside of the scope of this issue and it might be a blocker if we need to get design involved. it also might be worth taking a look at how this is handled in all the components. Thoughts? Want me to open an issue or do you want to? @mattnolting

mattnolting: Wfm! Captured the convo here: https://github.com/patternfly/patternfly/issues/2954

Issue #2656, opened 2020-01-31
Revise Label visuals


mcarrano:
Want to update the Label component. Goals include: color coding, adding icons, and introducing a removable label.

Final designs here: https://marvelapp.com/87942e3

cc @mcoker @LHinson

mcoker:
@mcarrano what's the use case for a clickable label? Is it a link or button that performs an action? Or is it "selectable", similar to data list rows?

mcoker:
Also do we have named states for the colors - similar to alerts?

mcarrano:
@mcoker The main use case I've seen for clickable labels is to use as a link. As for the color states, no these do not map like alerts. Consumers can use the colors how ever they want.

PR #2943, opened 2020-04-10
feat(label): refactored label component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2656

## Breaking changes
The label component has been redesigned and refactored. This PR introduces breaking visual and code changes. At a minimum to update to the new label component, the `.pf-c-label` component now requires `.pf-c-label__text` element to wrap the label text. And the label will be grey by default now, so to achieve the previous blue styling, the closest match visually would be using the new `.pf-m-blue` modifier. However, reference the new component documentation and the PR changeset for more info regarding the changes https://github.com/patternfly/patternfly/pull/2943/files

### DIFF
--- a/src/patternfly/components/Label/examples/Label.md
+++ b/src/patternfly/components/Label/examples/Label.md
@@ -38,30 +38,99 @@ import './Label.css'
   {{/label-text}}
 {{/label}}
 
-{{#> label label--id="default-grey-compact" label--modifier="pf-m-compact" label--isRemovable="true"}}
-  {{#> label-icon}}
-    <i class="fas fa-info-circle" aria-hidden="true"></i>
-  {{/label-icon}}
+<br><br>
+
+{{#> label label--id="default-grey-compact" label--modifier="pf-m-compact"}}
   {{#> label-text}}
     Grey compact
   {{/label-text}}
 {{/label}}
 
-{{#> label label--id="default-grey-link" label--type="a"}}
+{{#> label label--id="default-grey-compact-icon" label--modifier="pf-m-compact"}}
+  {{#> label-icon}}
+    <i class="fas fa-info-circle" aria-hidden="true"></i>
+  {{/label-icon}}
+  {{#> label-text}}
+    Grey compact icon
+  {{/label-text}}
+{{/label}}
+
+{{#> label label--id="default-grey-compact-removable" label--modifier="pf-m-compact" label--isRemovable="true"}}
+  {{#> label-text}}
+    Grey compact removable
+  {{/label-text}}
+{{/label}}
+
+{{#> label label--id="default-grey-compact-icon-removable" label--modifier="pf-m-compact" label--isRemovable="true"}}
+  {{#> label-icon}}
+    <i class="fas fa-info-circle" aria-hidden="true"></i>
+  {{/label-icon}}
+  {{#> label-text}}
+    Grey compact icon removable
+  {{/label-text}}
+{{/label}}
+
+<br><br>
+
+{{#> label label--id="default-grey-link" label-content--type="a" label-content--attribute='href="#"'}}
   {{#> label-text}}
     Grey link
   {{/label-text}}
 {{/label}}
 
-{{#> label label--id="default-grey-button" label--type="button"}}
+{{#> label label--id="default-grey-link-icon" label-content--type="a" label-content--attribute='href="#"'}}
+  {{#> label-icon}}
+    <i class="fas fa-info-circle" aria-hidden="true"></i>
+  {{/label-icon}}
+  {{#> label-text}}
+    Grey link icon
+  {{/label-text}}
+{{/label}}
+
+{{#> label label--id="default-grey-link-close" label-content--type="a" label-content--attribute='href="#"'  label--isRemovable="true"}}
+  {{#> label-text}}
+    Grey link removable
+  {{/label-text}}
+{{/label}}
+
+{{#> label label--id="default-grey-link-icon-close" label-content--type="a" label-content--attribute='href="#"' label--isRemovable="true"}}
+  {{#> label-icon}}
+    <i class="fas fa-info-circle" aria-hidden="true"></i>
+  {{/label-icon}}
+  {{#> label-text}}
+    Grey link icon and removable
+  {{/label-text}}
+{{/label}}
+
+<br><br>
+
+{{#> label label--id="default-grey-button" label-content--type="button"}}
   {{#> label-text}}
     Grey button
   {{/label-text}}
 {{/label}}
 
-{{#> label label--id="default-grey-clickable-removable" label--modifier="pf-m-clickable" label--attribute='tabindex="0"' label--isRemovable="true"}}
+{{#> label label--id="default-grey-button-icon" label-content--type="button"}}
+  {{#> label-icon}}
+    <i class="fas fa-info-circle" aria-hidden="true"></i>
+  {{/label-icon}}
   {{#> label-text}}
-    Grey clickable removable
+    Grey button icon
+  {{/label-text}}
+{{/label}}
+
+{{#> label label--id="default-grey-button-removable" label-content--type="button" label--isRemovable="true"}}
+  {{#> label-text}}
+    Grey button removable
+  {{/label-text}}
+{{/label}}
+
+{{#> label label--id="default-grey-button-icon-removable" label-content--type="button" label--isRemovable="true"}}
+  {{#> label-icon}}
+    <i class="fas fa-info-circle" aria-hidden="true"></i>
+  {{/label-icon}}
+  {{#> label-text}}
+    Grey button icon and removable
   {{/label-text}}
 {{/label}}
 
@@ -97,12 +166,6 @@ import './Label.css'
   {{/label-text}}
 {{/label}}
 
-{{#> label label--id="default-blue" label--modifier="pf-m-blue pf-m-clickable" label--attribute='tabindex="0"'}}
-  {{#> label-text}}
-    Blue clickable label
-  {{/label-text}}
-{{/label}}
-
 <br><br>
 
 {{#> label label--id="default-green" label--modifier="pf-m-green"}}
@@ -132,12 +195,6 @@ import './Label.css'
   {{/label-icon}}
   {{#> label-text}}
     Green label
-  {{/label-text}}
-{{/label}}
-
-{{#> label label--id="default-green" label--modifier="pf-m-green pf-m-clickable" label--attribute='tabindex="0"'}}
-  {{#> label-text}}
-    Green clickable label
   {{/label-text}}
 {{/label}}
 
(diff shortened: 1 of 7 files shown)


maryshak1996:
@mcoker these look great so far! 
@mcarrano to your quesiton about removable and clickable -- I think that they can live together just fine.
When the user hovers anywhere on the label that ISN'T the 'X', it would get the hover affect that I spelled out in my designs, and could click on the non-'X' areas for the clickable label functionality.
To dismiss / remove the label, they would hover over the 'X' itself, and the 'X' would go to the default icon hover color (black), and click on that 'X' to remove it

christiemolloy:
What do you think about the button being wider and less tall so that the focus ring fits within the constrain of the label, and you can use the extra width of the label to capture the area size

Screen Shot 2020-04-13 at 2 08 23 PM

christiemolloy:
Is this hoverable or selectable ? when I think hoverable i think that i wouldn't be able to select it like I can here.

Screen Shot 2020-04-13 at 2 10 28 PM

src/patternfly/components/Label/examples/Label.md
```diff
@@ -7,22 +7,553 @@ cssPrefix: pf-c-label
...
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-label` | `<span>` | Iniates a label. |
-| `.pf-m-compact` | `.pf-c-label` | Modifies label for a compact table. |
+| `.pf-c-label` | `<span>` | Iniates a label. Without a color modifier, the label's default style is grey. Use a color modifier to change the label color.  |
+| `.pf-m-compact` | `.pf-c-label` | Modifies label for compact styles. |
+| `.pf-m-outline` | `.pf-c-label` | Modifies label for outline styles. |
+| `.pf-m-hoverable` | `.pf-c-label` | Modifies label for a hoverable styles. |
+| `.pf-m-blue` | `.pf-c-label` | Modifies label for a blue styles. |
```
christiemolloy, line 554:
```suggestion
| `.pf-m-blue` | `.pf-c-label` | Modifies the label to have blue colored styling. |
```

src/patternfly/components/Label/label.hbs
```diff
@@ -3,4 +3,10 @@
     {{{label--attribute}}}
   {{/if}}>
   {{> @partial-block}}
+  {{#if label--IsClosable}}
```
christiemolloy, line 6: Should this be `IsRemovable` ?

src/patternfly/components/Label/label.scss
```diff
@@ -1,31 +1,258 @@
...
+  // Border base styles
+  --pf-c-label--before--BorderWidth: 0;
+  --pf-c-label--before--BorderColor: transparent;
+
+  // Compact
   --pf-c-label--m-compact--FontSize: var(--pf-global--FontSize--xs);
 
-  display: inline-block;
+  // Hoverable
+  --pf-c-label--m-hoverable--hover--before--BorderWidth: var(--pf-global--BorderWidth--sm);
```
christiemolloy, line 20: I left a comment but should this be called `selectable`

src/patternfly/components/Label/examples/Label.md
```diff
@@ -7,22 +7,553 @@ cssPrefix: pf-c-label
...
+```hbs title=Filled
+{{#> label label--id="default-grey"}}
+  {{#> label-text}}
+    grey label
+  {{/label-text}}
 {{/label}}
 
-{{#> label label--modifier="pf-m-compact"}}
-  Compact label
+{{#> label label--id="default-grey-icon"}}
```
christiemolloy, line 17: did you include a compact variation in here?

maryshak1996:
> What do you think about the button being wider and less tall so that the focus ring fits within the constrain of the label, and you can use the extra width of the label to capture the area size
> 
> Screen Shot 2020-04-13 at 2 08 23 PM

@christiemolloy is that something that can be somewhat easily done? I wouldn't want to create extra work for you guys, if this alters how the icon area usually works

maryshak1996:
sorry didn't mean to accidentally close this!

maryshak1996:
> Is this hoverable or selectable ? when I think hoverable i think that i wouldn't be able to select it like I can here.
> 
> Screen Shot 2020-04-13 at 2 10 28 PM

@christiemolloy on hover, the label should get an outline if it doesn't already have one. If there already is an outline, that outline should become 2px in thickness. I think that we should have the blue here on focus, but that shouldn't appear on hover

src/patternfly/components/Label/label.scss
mcoker: After talking to @mcarrano the use case for these is as links. So I'm thinking of styling `a` and `button` this way and adding a `.pf-m-clickable` variation  to be used with a non-interactive element like `<div class="pf-c-label">some text<button class="pf-c-button pf-m-plain">x</button</div>` to create a closable/clickable label since you can't put a `<button>` in a `<button>` or `<a>` element. wdyt?

mcoker:
@maryshak1996 @christiemolloy re: the button, we can do whatever! I don't think we would want it to be any wider, unless we positioned the button farther to the right, otherwise the increased width would overlap the button text. We can make it shorter though, good call. I think the button is 2px longer on the top/bottom.

mcoker:
> When the user hovers anywhere on the label that ISN'T the 'X', it would get the hover affect that I spelled out in my designs, and could click on the non-'X' areas for the clickable label functionality.

@maryshak1996 I reworked the component so that a close button can be placed inside the label alongside the element that contains the text/image, and that element can be a link or button if the user chooses. I've updated the demo with more examples, including compact label examples. 

One change you'll see is that the text/icon now receives focus instead of the label itself, since the link/button has to be separate from the close button (a close button can't go inside of a link or another button). Let me know what you think of the focus behavior.

Since the label border changes serve as an indication that the label has focus, we could disable the default browser focus ring, too.

@christiemolloy addressed your comments

mcoker:
@maryshak1996 @mcarrano I've updated the grey/default label examples to be more robust to show all of the different states. Do you have a recommendation on how to present these examples, showing the options available while not having an overwhelming number of examples?

mcarrano:
Thanks for making all of these updates @mcoker . Here is some feedback.

* For the gray labels with icons, I think that the icon should be dark gray, not blue. Can you confirm @maryshak1996 ?
* Looks like you re-introduced the compact label? During design we had decided that there was not much benefit of having this. Do you still think that's the case @maryshak1996 ? I don't feel strongly about it, but if it is only minimal benefit then perhaps we should remove.
* I question whether we should have a label both as a link and a button. I understand the distinction, but not sure I see a use case for making a label be a button to invoke an action. Feel like having this might be misleading from a design perspective.

Good question about how best to present these examples. Maybe the first section in labeled "Variants" that shows all of the different combination with the gray filled labels. Then a second section that is "Label Styles"  and show all of the different color and border options, but I don't see a need to repeat all of the variants for both filled and outlined. Hopefully people will get that these are just different visual styles that can be applied to any label variant.

@mcoker @maryshak1996 what do you think?

maryshak1996:
@mcarrano @mcoker 
I second what Matt said about the presentation of these examples (since there is a lot of variants, this seems like the most clear way to see it all). 
In regard to the color of the icon in the grey label, my initial thinking was the to have blue (to provide more contrast from the close icon), but @mcarrano I think that now, seeing these in context, having the icon be the black icon color is the way to go -- it does look out of place with it being blue. I'll make this update in the sketch library too. 
To the question about compact sizing -- ya the consensus was definitely to have just this one size (I mocked it up in more compact spaces and it still works just fine). I would be in favor of removing the 'compact size' for sure.

mcoker:
Great thanks @maryshak1996 @mcarrano! Removed the compact variation, and support for buttons. Also trimmed down the examples just to test all of the variations behave and look as expected. I created a new issue for the next sprint to clean up the examples - https://github.com/patternfly/patternfly/issues/2959

mcarrano:
I think this looks much better now @mcoker . You might not even need to refactor the examples as I think this looks OK now that you removed some of the variants. There is just one bug that I see. The "green removable link" is colored gray

Screen Shot 2020-04-14 at 1 25 24 PM

mcoker:
@mcarrano thanks, squashed that bug.

src/patternfly/components/Label/label.scss
christiemolloy: ok that works!

mcoker: Sorry so much has changed in the last day! Since @maryshak1996 prefers the hover outline not be shown on a link label when you hover the close button, I separated the label into 2 children - a `__content` element (that can be a `<span>` or `<a>`) and the close button. So you just use an `<a>` on `__content` if you want a link, and the close button is an adjacent element (instead of a child), so no need for the `clickable` class manipulation.

maryshak1996:
@mcoker this looks perfect to me!

src/patternfly/components/Label/label.scss
```diff
@@ -1,31 +1,271 @@
...
+
+.pf-c-label__content {
+  &::before {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    content: "";
+    border: var(--pf-c-label__content--before--BorderWidth) solid var(--pf-c-label__content--before--BorderColor);
```
christiemolloy, line 234: why does this pseudo element need to be on `pf-c-label__content` ? can it just go on label?

src/patternfly/components/Label/label-content.hbs
```diff
@@ -0,0 +1,9 @@
+<{{# if label-content--IsLink}}a{{else}}span{{/if}} class="pf-c-label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
mcoker: Not necessarily. The `__content` element is the container for the icon and text, so we could just have the text inline with the icon. This gives us a container for the text so we have more control over it, though it's not necessary to have right now.

src/patternfly/components/Label/label.scss
mcoker: The hover border treatment can't apply to the label since we don't want the label border to change when it's a link and the user hovers over the close button.

src/patternfly/components/Label/label-content.hbs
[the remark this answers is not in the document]
christiemolloy: hmm.. i think the structure should be something similar to the button component. since we decided not to add the text wrapper, i don't know if its necessary to have two wrappers around the text in the label component ( `pf-c-label` pf-c-label__content` and `pf-c-label__text` ) @mattnolting thoughts?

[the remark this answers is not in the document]
mcoker: works for me!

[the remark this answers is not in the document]
mcoker: @christiemolloy actually just looked at it again and one reason we use the text element is, similar to chips, it serves as a value passed to the close button's `aria-labelledby` attribute.

PR #2955, opened 2020-04-13
feat(global): made redhat font default, add opt-in for overpass

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2600

## Breaking changes
This PR removes the `.pf-m-redhat-font` class from PatternFly that changed the base font families to a new Red Hat font as well as changed base styles for a number of components to be used with the Red Hat font. The Red Hat font and its associated component styles are now the default font in PatternFly. If an application was using `.pf-m-redhat-font` before, it can safely be removed as that class and the default styles should now match what they were before.

If an application wants to continue using the Overpass font, this PR introduces a class `.pf-m-overpass-font` that works in the exact same way as `.pf-m-redhat-font` did. It changes the base font families to Overpass fonts, and applies specific component styling to be used with Overpass. To enable Overpass in your application, simply wrap your application in an element with `.pf-m-overpass-font`. This is most commonly applied to the `<body>` element, though it can apply to any element that wraps your application. It can also be used to wrap individual components if that is more desirable.

The following are breaking changes that accompany these updates:

* If an application is not using `.pf-m-overpass-font`, the card title is now bold. This is a visual change only. No further code updates are needed to consume this change.

* The following variables have been removed as they are no longer necessary. Any reference to them should be removed from an application as they are no longer used in patternfly.
  * `--pf-global--link--FontWeight` 
  * `--pf-global--FontFamily--redhatfont--sans-serif`
  * `--pf-global--FontFamily--redhatfont--heading--sans-serif`
  * `--pf-global--FontFamily--redhatfont--monospace`
  * `--pf-global--FontWeight--redhatfont--bold`
  * `--pf-c-breadcrumb__item--FontWeight`
  * `--pf-c-breadcrumb__link--FontWeight`
  * `--pf-c-content--blockquote--FontWeight`
  * `--pf-c-expandable-section__toggle--FontWeight`

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -82,6 +82,5 @@
   .pf-c-breadcrumb__link,
   .pf-c-breadcrumb__item {
     font-weight: var(--pf-global--FontWeight--semi-bold);
-    font-weight: var(--pf-global--FontWeight--semi-bold);
   }
 }
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -1,7 +1,6 @@
 .pf-c-breadcrumb {
   // Breadcrumb item
   --pf-c-breadcrumb__item--FontSize: var(--pf-global--FontSize--sm);
-  --pf-c-breadcrumb__item--FontWeight: var(--pf-global--FontWeight--semi-bold);
   --pf-c-breadcrumb__item--LineHeight: var(--pf-global--LineHeight--sm);
   --pf-c-breadcrumb__item--MarginRight: var(--pf-global--spacer--sm);
 
@@ -11,7 +10,6 @@
   --pf-c-breadcrumb__item-divider--FontSize: var(--pf-global--FontSize--sm);
 
   // Breadcrumb link
-  --pf-c-breadcrumb__link--FontWeight: var(--pf-global--FontWeight--semi-bold);
   --pf-c-breadcrumb__link--m-current--Color: var(--pf-global--Color--100);
 
   // Heading
@@ -80,9 +78,10 @@
 }
 
 // RedHat Font overrides
-@include pf-m-redhat-font {
-  .pf-c-breadcrumb {
-    --pf-c-breadcrumb__link--FontWeight: var(--pf-global--FontWeight--normal);
-    --pf-c-breadcrumb__item--FontWeight: var(--pf-global--FontWeight--normal);
+@include pf-m-overpass-font {
+  .pf-c-breadcrumb__link,
+  .pf-c-breadcrumb__item {
+    font-weight: var(--pf-global--FontWeight--semi-bold);
+    font-weight: var(--pf-global--FontWeight--semi-bold);
   }
 }
--- a/src/patternfly/components/Card/card-title.hbs
+++ b/src/patternfly/components/Card/card-title.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-card__title{{#if card-title--modifier}} {{card-title--modifier}}{{/if}}"
+  {{#if card-title--attribute}}
+    {{{card-title--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
--- a/src/patternfly/components/Card/examples/Card.md
+++ b/src/patternfly/components/Card/examples/Card.md
@@ -7,9 +7,9 @@ cssPrefix: pf-c-card
 ## Examples
 ```hbs title=Basic
 {{#> card}}
-  {{#> card-header}}
+  {{#> card-title}}
     Header
-  {{/card-header}}
+  {{/card-title}}
   {{#> card-body}}
     Body
   {{/card-body}}
@@ -31,9 +31,9 @@ cssPrefix: pf-c-card
       <input type="checkbox" id="{{card--id}}-check" name="{{card--id}}-check" aria-labelledby="{{card--id}}-check-label">
     {{/card-actions}}
   {{/card-head}}
-  {{#> card-header card-header--attribute=(concat 'id="' card--id '-check-label"')}}
+  {{#> card-title card-title--attribute=(concat 'id="' card--id '-check-label"')}}
     Header
-  {{/card-header}}
+  {{/card-title}}
   {{#> card-body}}
     Body
   {{/card-body}}
@@ -51,9 +51,9 @@ cssPrefix: pf-c-card
       {{/dropdown}}
       <input type="checkbox" id="{{card--id}}-check" name="{{card--id}}-check" aria-labelledby="{{card--id}}-check-label">
     {{/card-actions}}
-    {{#> card-header card-header--attribute=(concat 'id="' card--id '-check-label"')}}
+    {{#> card-title card-title--attribute=(concat 'id="' card--id '-check-label"')}}
       This is a really really really really really really really really really really long header
-    {{/card-header}}
+    {{/card-title}}
   {{/card-head}}
   {{#> card-body}}
     Body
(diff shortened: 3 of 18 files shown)


mcoker:
@mcarrano @mceledonia 

To test, you can open dev tools and find the `<html>` element at the very top of the source code, double click on "pf-m-redhat-font" to edit that text, and replace it with "pf-m-overpass-font" to show the overpass font in the UI. To see the redhat font, you just can't have "pf-m-overpass-font" in the html element's class. So you can remove one character (ie, "pf-m-overpass-fon"), and that will show the redhat text, which is (maybe?) an easier way to toggle back and forth between the two fonts.

* ensure link text is semi-bold in overpass, normal in redhat text - there is an example on the [content component](https://patternfly-pr-2955.surge.sh/documentation/core/components/content)
* breadcrumb items are semi-bold overpass, normal redhat text
* button is semi-bold overpass, normal redhat text
* alert title
* card title
* empty state title
* h1-h6 on content component (all of the heading elements) are the right font-weight
* blockquote on content component is light font in overpass, regular in redhat
* expandable toggle is semi-bold in overpass, regular in redhat
* medium and large title are semi-bold in overpass, regular in redhat

Also just overall that the typography looks correct.

src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -81,9 +79,10 @@
...
 
 // RedHat Font overrides
-@include pf-m-redhat-font {
-  .pf-c-breadcrumb {
-    --pf-c-breadcrumb__link--FontWeight: var(--pf-global--FontWeight--normal);
-    --pf-c-breadcrumb__item--FontWeight: var(--pf-global--FontWeight--normal);
+@include pf-m-overpass-font {
+  .pf-c-breadcrumb__link,
+  .pf-c-breadcrumb__item {
+    font-weight: var(--pf-global--FontWeight--semi-bold);
```
christiemolloy, line 85: do you need two font-weight's here?

src/patternfly/components/Card/examples/Card.md
```diff
@@ -7,9 +7,9 @@ cssPrefix: pf-c-card
 ## Examples
 ```hbs title=Basic
 {{#> card}}
-  {{#> card-header}}
+  {{#> card-title}}
```
christiemolloy, line 10: should you also update the card demo?

src/patternfly/components/Card/examples/Card.md
```diff
@@ -250,7 +250,7 @@ A card is a generic rectangular container that can be used to build other compon
 | Class | Applied | Outcome |
 | ---- | ---- | ---- |
 | `.pf-c-card` | `<div>` | Creates a card component.  **Required** |
-| `.pf-c-card__header` | `<div>` | Creates the header of a card. |
+| `.pf-c-card__title` | `<div>` | Creates the header of a card. |
```
christiemolloy, line 253:
```suggestion
| `.pf-c-card__title` | `<div>` | Creates the title of a card. |
```

src/patternfly/components/EmptyState/empty-state-title.hbs
```diff
@@ -0,0 +1,6 @@
+<h1 class="pf-c-empty-state__title{{#if empty-state-title--modifier}} {{empty-state-title--modifier}}{{/if}}"
```
christiemolloy, line 1: do we want to make this `<h1>`?  there could be many occassions where the empty state is small and there are many of them on a page

mcarrano:
@mcoker I've tested as per your instructions above and verified that the font defaults to the Red Hat font as expected. I've asked @gdoyle1 to also verify against the Sketch library to ensure consistency with design across components.

gdoyle1:
Differences btwn code and sketch library:
**I assume the following needs to be updated in code**
1. Nav group section titles should be var(--pf-global--FontWeight--semi-bold) which equals 700 weight in Sketch library 
--> This matches to RHT medium and is the same weight as used in Application launcher, light nav and form labels
2. Dark nav selected items use var(--pf-global--FontWeight--normal) but should be var(--pf-global--FontWeight--bold) so it is RHT medium (unless this has changed @mceledonia ?)
3. Empty state titles are var(--pf-global--FontWeight--normal) but should be RHD medium... it looks like they are RHD regular in this PR
--> Sketch library has them as RHD medium
4. Login preview is showing up as blue text (but it is correct in the full page view)
Screen Shot 2020-04-15 at 11 17 41 AM
5. Pagination options menu items are using both 16px and 14px text - the full option should be 16px text and all black text. Also the pagination options menu toggle in general has some weird spacing going on
--> Sketch library has menu items as 16px and all black text
6. Previews for wizard look weird
Screen Shot 2020-04-15 at 11 32 01 AM

**Update to Sketch library (made these while I was going through)**
1. Card titles are 16px RHT medium instead of 16px RHT regular
2. Wizard selected steps and substeps get RHT medium weight and blue

cc: @mcarrano @mcoker

src/patternfly/components/Card/card-title.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-card__title{{#if card-title--modifier}} {{card-title--modifier}}{{/if}}"
+  {{#if card-title--attribute}}
+    {{{card-title--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
```
mattnolting, line 6: I wonder about changing this to `__title` for a couple of reasons. It's somewhat confusing, without supporting examples, as to when I should use `h2.pf-c-card__title` vs `.pf-c-card__title > h2.pf-c-title.pf-m-xl` (if I want to change the card header/title text styling). And, because of React's implementation, they'll need to make API changes. Should you revert this and open a follow up issue?

mcoker: re: examples, agree but that's pre-existing. All of our docs/examples use a div, and the react `<CardHeader>` just renders a `<div>`. And in react, the `<CardHeader>` isn't a `<Title>` component, it just hardcodes the `.pf-c-title.pf-m-md` classes to achieve a font-size/font-weight combo. The goal of this change is just to remove the dependency on `.pf-c-title` being hardcoded as a class to `.pf-c-card__header`, and apply the font size/weight combo to the card element instead. 

If we just remove `.pf-c-title` from `.pf-c-card__header`, and style that element directly, I'm pretty sure that doesn't break the react API. React just needs to remove the hard coded class from the [CardHeader component](https://github.com/patternfly/patternfly-react/blob/v4/packages/react-core/src/components/Card/CardHeader.tsx). But agreed I'll revert this - I'd like to verify to be sure, and it's past the point where we can add work to react's plate. Also, I'd like to rename it to `card__title` to be consistent (and not confuse with `card__head`).

I made a similar change to the empty state component in this PR I'll revert, too. We have an issue filed in react to update instances like the card title and empty state title that I think we could get this work in with since it has been planned - https://github.com/patternfly/patternfly-react/issues/3676

src/patternfly/components/Card/examples/Card.md
mcoker: reverted this change

src/patternfly/components/Card/examples/Card.md
mcoker: reverted this

src/patternfly/components/EmptyState/empty-state-title.hbs
mcoker: Reverted this. I made it an `<h1>` tho since that's what all of our existing examples are, and added this to the docs:

```
| `.pf-c-empty-state__title` | `<div>`, `<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`, `<h6>` |  Creates the empty state title. **Required** |
```

mcoker:
Thanks @gdoyle1! I've opened a separate issue to address these as they look like pre-existing issues https://github.com/patternfly/patternfly/issues/3010.

mcoker:
@mattnolting @christiemolloy ready for review

PR #2957, opened 2020-04-14
Readme fix 2947

ddonahue007:
One of the commits was to have both dev configurations in the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ To setup the PatternFly 4 development environment:
 - clone the project
 - run `npm install` from the project root
 - run `npm run build-patternfly`
+- run `npm run cli:setup` (only needed if doing development)
 - run `npm run dev`
 - open your browser to `http://localhost:8000`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,15 @@ To setup the PatternFly 4 development environment:
 
 - clone the project
 - run `npm install` from the project root
-- run `npm run dev:lite`
+- run `npm run build-patternfly`
+- run `npm run dev`
+- open your browser to `http://localhost:8000`
+
+As an alternative development environment:
```
dgutride, line 30: Instead of listing this- we should be clear about why people would use this instead of dev above - seems odd to recommend two alternate ways of running the project.

ddonahue007: decided to remove the second dev configuration from the readme in favor of communicating that to users one on one.

README.md
```diff
@@ -23,7 +23,9 @@ To setup the PatternFly 4 development environment:
 
 - clone the project
 - run `npm install` from the project root
-- run `npm run dev:lite`
+- run `npm run build-patternfly`
```
mcoker, line 26: We still need `cli:setup` since that adds `pf.js` to your path so you can use `pf generate` commands

> - run `npm run cli:setup` (only needed if doing development)

ddonahue007: sorry missed it on commit

PR #2960, opened 2020-04-14
Moving some top level scss files into directory structure"

### LINKED ISSUE #1622: Clean up SCSS/includes
This is a follow-up to https://github.com/patternfly/patternfly-next/pull/1467. We talked about renaming and potentially reorganizing some of the files so it's more intuitive to find and include files in a project.

First steps should be to come up with a plan that we can review as a group before beginning. For this release, we should not move `patternfly.scss` or `patternfly-base.scss`.

cc @dgutride

ddonahue007:
closes #1622 

Scope of this PR is to move most of the top level scss files into a directory structure.

## BREAKING CHANGES 

If you are importing any of the following files directly you will need to update import statements.

** changed **
- patternfly-common.css to base/patternfly-common.css
- patternfly-fa-icons.css to base/patternfly-fa-icons.css
- patternfly-fonts.css to base/patternfly-fonts.css
- patternfly-globals.css to base/patternfly-globals.css
- patternfly-icons.css to base/patternfly-icons.css
- patternfly-pf-icons.css to base/patternfly-pf-icons.css
- patternfly-shield-inheritable.css to base/patternfly-shield-inheritable.css
- patternfly-shield-noninheritable.css to base/patternfly-shield-noninheritable.css
- patternfly-themes.css to base/patternfly-themes.css
- patternfly-variables.css to base/patternfly-variables.css

** removed **
- patternfly-imports.scss


gatsby-browser.js
```diff
@@ -14,4 +14,4 @@ import 'gatsby-theme-patternfly-org/global.css';
 import './dist/patternfly.css';
 
 // Utilities
-import './dist/patternfly-addons.css';
```
redallen, line 17: This is probably going to break more than just our docs site. I believe many products use our addons. Can we keep it at the root of `dist` when we build?

ddonahue007: I realize we want to minimize the level of breakage we cause. One thing to note is by making this change will cause the scss and css files do not live along side of each other. We could move both charts and addons back to the top level and remove the addons directory completely for now.

redallen:
Merge conflict in `src/patternfly/_shield-inheritable.scss`

src/patternfly/base/_shield-inheritable.scss
```diff
@@ -25,7 +25,7 @@
   .pf-c-drawer,
   .pf-c-dropdown,
   .pf-c-empty-state,
-  .pf-c-expandable,
+  .pf-c-expandable-section,
```
redallen, line 28: Not sure if a merge went awry, but why is this showing up in the diff? :(

ddonahue007: mmmm, good question. let me double heck

gulpfile.js
```diff
@@ -10,9 +10,10 @@ const { generateSnippets } = require('./build/gulp/snippets');
 
 const sassFiles = [
   './src/patternfly/patternfly*.scss',
+  './src/patternfly/addons/patternfly*.scss',
```
redallen, line 13: There is no addons directory?

ddonahue007: Done

.circleci/config.yml
```diff
@@ -110,6 +110,10 @@ jobs:
           name: Lint CSS
           command: npm run lint:css
           when: always
+      - run:
```
redallen, line 113: Thanks for adding this!

PR #2961, opened 2020-04-14
fix(chipgroup): clarify chip group behavior

### LINKED ISSUE #1837: Clarify Chip Group Behavior
The Chip Group component was introduced to manage the behavior of a group of chips used in either a toolbar (i.e. filter chips), a type-ahead select, or for tagging. The current implementation and application raises some question about desired handling of overflow conditions, i.e. what happens when there are too many chips to fit into the parent container. 

Currently this is set as a hard-coded number of chips but it somehow needs to be made more dynamic and detect when to go into overflow based on the size of the parent. Additionally it would be good to arrive at a consistent experience that would support all use cases where the chips are used.

Some background and questions raised by the current design and implementation are documented here: https://docs.google.com/document/d/1bmsgR28EdJ1FsroTBn0crxBB4xdiGPkVGpQ__yaWeG4/edit?usp=sharing

christiemolloy:
closes #1837 

spoke with @mceledonia and @mcarrano  about the new chip group behavior, where the overflow button sits within the chip group.


src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -1,58 +1,50 @@
...
-  --pf-c-chip-group--m-toolbar-PaddingRight: var(--pf-global--spacer--xs);
-  --pf-c-chip-group--m-toolbar-PaddingBottom: var(--pf-global--spacer--xs);
-  --pf-c-chip-group--m-toolbar-PaddingLeft: var(--pf-global--spacer--sm);
-  --pf-c-chip-group--m-toolbar-BorderRadius: var(--pf-global--BorderRadius--sm);
-  --pf-c-chip-group--m-toolbar--BackgroundColor: var(--pf-global--BackgroundColor--200);
-  --pf-c-chip-group__li--m-toolbar--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-chip-group--m-category--PaddingTop: var(--pf-global--spacer--xs);
+  --pf-c-chip-group--m-category--PaddingRight: var(--pf-global--spacer--xs);
+  --pf-c-chip-group--m-category--PaddingLeft: var(--pf-global--spacer--sm);
+  --pf-c-chip-group--m-category-BorderRadius: var(--pf-global--BorderRadius--sm);
```
mcoker, line 7:
```suggestion
  --pf-c-chip-group--m-category--BorderRadius: var(--pf-global--BorderRadius--sm);
```

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -1,58 +1,50 @@
...
 
-    > .pf-c-chip.pf-m-overflow {
-      margin: 0;
-    }
+  &.pf-m-category {
+    padding-top: var(--pf-c-chip-group--m-category--PaddingTop);
+    padding-right: var(--pf-c-chip-group--m-category--PaddingRight);
+    padding-left: var(--pf-c-chip-group--m-category--PaddingLeft);
+    background-color: var(--pf-c-chip-group--m-category--BackgroundColor);
+    border-radius: var(--pf-c-chip-group--m-category-BorderRadius);
```
mcoker, line 32:
```suggestion
    border-radius: var(--pf-c-chip-group--m-category--BorderRadius);
```

src/patternfly/components/ChipGroup/examples/ChipGroup.md
```diff
@@ -5,269 +5,264 @@ cssPrefix: pf-c-chip-group
...
-| Class | Applied to | Outcome |
-| -- | -- | -- |
-| `.pf-c-chip-group__close` | `<div>` | Initiates the container used to close chip group. **Required.** |
-
-The chip group requires the [chip component](/documentation/core/components/chip).
-
-```hbs title=Multi-select-collapsed
-{{#> chip-group chip-group--id="select-collapsed"}}
+```hbs title=Chip-group-with-categories-overflow
+{{#> chip-group chip-group--modifier="pf-m-category" chip-group--id="toolbar-collapsed"}}
```
mcoker, line 125: `toolbar-collapsed` is used in another example.

src/patternfly/components/ChipGroup/examples/ChipGroup.md
```diff
@@ -5,269 +5,264 @@ cssPrefix: pf-c-chip-group
...
+  {{/chip}}
+  {{> chip-group-close}}
 {{/chip-group}}
 ```
 
 ### Overview
-A chip-group used in a multi-select do not require a modifier as its parent will handle its styles. When it overflows outside of the group, the chips will wrap to the next line. This overflow is indicated by a chip with the modifier `.pf-m-overflow` that acts as a button to show/hide the overflown chips by expanding the height of the container they belong to.
+A chip group is constrained to the width of its container and will wrap when it exceeds that width. An overflow value can be set and when the number of chips exceeds that value, additional chips will be hidden by default. The default overflow value will be set to 3 chips but this can be adjusted per application needs. The toggle button after the last chip allows the group to be expanded (or collapsed).
+
+If you want to create sub-groupings of chips to represent multiple values applied against the same category, chips can be grouped by category. This can be useful in filtering use cases, for example, where you want all items the match more than one value of the same attribute, e.g., ‘status = down OR needs maintenance’.
```
mcoker, line 250: "where you want all items ~the~ _that_ match"?

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -1,58 +1,50 @@
 .pf-c-chip-group {
   @include pf-t-light; // force the container follow the light theme
 
-  --pf-c-chip-group--MarginRight: var(--pf-global--spacer--xs);
-  --pf-c-chip-group--MarginBottom: calc(var(--pf-c-chip-group--c-chip--MarginBottom) * -1);
```
mcoker, line 5: Removing this negative margin will create a margin underneath chips in a chip group, since the chips have margin.

Screen Shot 2020-04-15 at 5 25 16 PM

christiemolloy: ah good catch

christiemolloy:
there is no hover/focus state for the overflow - confirmed with design @mcoker

christiemolloy:
@mcarrano what is a good modifier name for grouped chips with the label. It was originally `toolbar` but it isn't specific to the toolbar anymore. I changed it to `category` but that might be too specific, @mcoker suggested `labelled` which might be more appropriate ?

src/patternfly/components/ChipGroup/examples/ChipGroup.md
christiemolloy: updated the ID of all of them to reflect what is happening

src/patternfly/components/ChipGroup/chip-group.scss
christiemolloy: Adding the margin bottom of -4px but does this mean that we could run into the error where the component is taller than what is on screen.

mcoker:
Looks like we need to update these data-toolbar examples/demos

```
components/DataToolbar/data-toolbar-item-chip-group.hbs:  {{#> chip-group chip-group--modifier="pf-m-toolbar"}}
demos/DataToolbar/examples/DataToolbar.md:      {{#> chip-group chip-group--modifier="pf-m-toolbar" chip-group--id=(concat data-toolbar--id '-group1-')}}
demos/DataToolbar/examples/DataToolbar.md:        {{#> chip-group chip-group--id="data-toolbar-chip-group-one" chip-group--modifier="pf-m-toolbar"}}
```

mcoker:
And review the use of chip groups in the select component

Screen Shot 2020-04-17 at 7 46 31 PM
Screen Shot 2020-04-17 at 7 46 24 PM

Issue #2907, opened 2020-03-16
Improve responsive behavior of pages with vertical nav


mcarrano:
Want to improve the way our page responds as the viewport width is reduced. This is especially true for pages with tables that have lots of rows and content. Content becomes compressed before navigation sidebar hides leaving an unusable page. Need to investigate this problem together with Core to explore alternative approaches.

Here is an example of what this currently looks like: https://slack-files.com/T0FA4HM8D-F010YDKLLVB-6ecfe221e7

cc @kybaker @LHinson

mcoker:
@mcarrano @kybaker and I met today and agreed to change the breakpoint at which the page sidebar collapses from `$pf-global--breakpoint--md` to `$pf-global--breakpoint--xl`.

This will also involve updating the left/right component and page chrome padding changes that happen at `$pf-global--breakpoint--md`, so that they happen at `$pf-global--breakpoint--xl` instead.

PR #2962, opened 2020-04-14
fix(global): hid vertical nav at 1200px, updated component spacing

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2907

Testing notes:
The components that changed: data-list, nav, page, table, tabs (removed the top/bottom padding change that happened at 768px), wizard. We just need to test the spacing in these components in general, but specifically to make sure the horizontal spacing aligns with the overall page chrome so that when they're used in a page main section, their horizontal padding aligns with the page chrome. For example...

desktop table
Screen Shot 2020-04-14 at 5 25 42 PM

mobile table
Screen Shot 2020-04-14 at 5 25 27 PM

## Breaking changes
This PR changes the CSS breakpoint at which the page component's sidebar shows and hides. This used to be defined at `$pf-global--breakpoint--md` (768px) and is now defined at `$pf-global--breakpoint--xl` (1200px). This change results in a wider main content area of the page, so this may impact an application's UI. No code changes are needed to allow the breakpoint change, since the breakpoint is defined in CSS.

The overall page chrome and some individual components' horizontal spacing becomes more compact at this breakpoint (true of the old breakpoint and the new), so applications may find that custom elements that aligned with the old page chrome may not align now since the chrome spacing changes at 1200px instead of 768px. An application may need to make updates to match the chrome's spacing at the new breakpoint.

There is a long list of variable names that have changed across individual components that you can find in this PR's [file changeset](https://github.com/patternfly/patternfly/pull/2962/files). 

This PR also changes the gutter in patternfly layouts (gallery, grid, level, split, stack) to have a single gutter instead of responsive, so the gutter is always 16px, instead of being 16px on mobile and 24px on desktop. No futher changes are needed to consume this update, though applications may find that custom elements that may have aligned with the spacing at the old breakpoint no longer align. An application may need to make updates to match spacing at the new breakpoint.

The page sidebar and vertical navigation are now always the same size between mobile and desktop. No further updates are needed to consume this change.

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -50,12 +50,12 @@
   // Data list item row
   --pf-c-data-list__item-row--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-data-list__item-row--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-data-list__item-row--md--PaddingRight: var(--pf-global--spacer--lg);
-  --pf-c-data-list__item-row--md--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-data-list__item-row--xl--PaddingRight: var(--pf-global--spacer--lg);
+  --pf-c-data-list__item-row--xl--PaddingLeft: var(--pf-global--spacer--lg);
 
-  @media screen and (min-width: $pf-global--breakpoint--md) {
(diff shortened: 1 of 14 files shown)


christiemolloy:
In https://patternfly-pr-2962.surge.sh/documentation/core/demos/datalist/simple the overflow arrow is pointing the wrong way

mcoker:
Thanks for looking at it @mcarrano!!

mcoker:
@christiemolloy with the eagle eyes! Fixed.

Issue #2046, opened 2019-07-11
The Wizard is another implementation of Modal


tlabaj:
The following issue was submitted to the react repo [#2114](https://github.com/patternfly/patternfly-react/issues/2114)  It needs to be investigated by core first. 

> PatternFly should reuse its own components and it should use the Modal component.
> 
> The wizard's render function starts with
> 
> isOpen && ReactDOM.createPortal(
>         <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
>           <Backdrop>
>             <Bullseye>
> Instead of that it should be wrapped only the Modal component. (And Modal component needs to be updated to be able to use here.) Or a consumer should be able to put modal there in his own implementation, if he wants to do. There is no reason to control this in the Wizard component, when you have another component, which does it. ( => Less code to test => less code to maintain => less code to break => more satisfied consumers and happy faces)
> 
> <Modal isOpen>
>   ...
> </Modal>
> or in user's implementation
> 
> <Modal isOpen>
>    <Wizzard {...props} />
> </Modal>
>

matthewcarleton:
+1 we also have the need for the wizard to be in-page in OpenShift so it would great if the parts of the wizard could live outside of the modal that wraps it. Being in-page may also require some additional design updates as well, not sure if that would be part of this effort or not. 
FYI @rachael-phillips @LHinson

mcoker:
I tried this locally and looks like we will need the following changes

#### Wizard
* remove in-page variation, make default
* remove full-height/full-width variations

#### Modal
* add full-height/full-width variations to modal, or some way to support the wizard's default size
* add option to remove modal padding
* add option to remove modal close button

Issue #2021, opened 2019-07-03
Create action list component


mcoker:
**Updated design:** https://docs.google.com/presentation/d/1Kh8wGM27X5shhSyVpZbgjV3uidpOmlz-CLYWHhxEMEI/edit?usp=sharing

---

Let's create a layout that spaces and positions actions properly for use in the wizard, modal, and form, and anywhere you need a list of actions that would be spaced similarly.

cc @matthewcarleton @mcarrano 

Can we talk about the specifics needed for this layout? Maybe some examples of types of buttons that would go in this layout? And how they should behave responsively and maintain the proper spacing?

Could this also be used in places like the data list?

<img width="1212" alt="Screen Shot 2019-07-03 at 4 06 38 PM" src="https://user-images.githubusercontent.com/35148959/60625163-9cac8b00-9dac-11e9-862c-050f9ae4dedd.png">

matthewcarleton:
Thanks for creating this @mcoker 
This would be a good fit for the wizard action spacing issue #1894  and the modal issue #1999 
I'd assume any type of button can be used in this layout.
vertical spacing should be easy enough for wrapping actions. 
The more complicated issue is the spacer modifier for link type actions. 
Once wrapped we'll see the possibility of this:
<img width="357" alt="Screen Shot 2019-07-04 at 1 02 47 PM" src="https://user-images.githubusercontent.com/287880/60678874-3b43f500-9e5c-11e9-806c-6024f589ecea.png">

This layout will also have to handle left aligned and right aligned actions.

mcarrano:
Sounds good. I will add this to the design backlog. @LHinson @rachael-phillips FYI

mattnolting:
> Could this also be used in places like the data list?

I think we'd use the overflow menu in data list.

christiemolloy:
@mcarrano do you know if there are design stories done yet for this?

mcarrano:
No, not specifically @christiemolloy .  @mcoker do you think that the Overflow Menu component handles this? Do we need a new layout? If so, is input from design needed for this?

mattnolting:
@mcarrano I think this is a valid layout, but in components like `table` and `data-list`, where the layout switches to grid and buttons are added to a dropdown, (where we're already emulating overflow behavior) `.pf-c-overflow-menu` would handle that.

mcoker:
[...]

PR #2964, opened 2020-04-15
fix(wizard): reworked wizard so it can be used in modal

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2046

@tlabaj @mcarrano this is pretty straight forward except for one thing. The wizard when used in a modal needs a defined height, otherwise the height will be defined by the wizard contents and will change as a user progresses through the wizard. There are a couple of ways we can do this.

1) Assign a height to the wizard for use when it's in a modal, then modify that height when the wizard is used in context of the page (the "in page" variation). This seems to work just fine and allows us to define a specific height for the wizard (that the user can override), but it means that we need to maintain separate variations of the wizard for use in the page and in a modal. The user impact is that the user has to specify the modal width when using a modal, and they have to specify the "in page" variation when using the modal in a page.

2) Add an enhancement to the modal to define a modal height (that is overridable by the user, similar to the height variations we currently offer) and allow that to set the wizard modal height. The wizard can then simply occupy the height of the container it's in both the modal and in page variations, meaning we could do away with the in page wizard variation. The user impact of this is that the user must specify both the width and height when using a modal, but doesn't have to specify a variation of the wizard to use in either context.

WDYT?

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -1,6 +1,6 @@
 .pf-c-wizard {
   --pf-c-wizard--Height: 100%;
-  --pf-c-modal-box--c-wizard--Height: #{pf-size-prem(762px)}; // this is a specific height from design for the wizard in a modal
+  --pf-c-modal-box--c-wizard--FlexBasis: #{pf-size-prem(762px)}; // this is a specific height from design for the wizard in a modal
 
   // Header
   --pf-c-wizard__header--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);
@@ -192,9 +192,7 @@
 
   // When the wizard is used in a modal
   .pf-c-modal-box & {
-    --pf-c-wizard--Height: var(--pf-c-modal-box--c-wizard--Height);
-
-    max-height: var(--pf-c-modal-box--c-wizard--MaxHeight, inherit);
+    flex: 1 1 var(--pf-c-modal-box--c-wizard--FlexBasis);
   }
 
   > *:not(.pf-c-wizard__outer-wrap) {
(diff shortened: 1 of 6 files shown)


mcarrano:
@mcoker I'm having a bit of a hard time understanding the examples and what I should be looking for. As to your question, IMO the second option seems like a cleaner approach. 

Just another small thing, a while ago we decided that the Cancel button should be spaced further away from the Next and Back buttons. I don't see that here. Here is the most recent visual spec:

Wizard spec

mcoker:
@mcarrano the cancel button spacing was put on hold in favor of the action list component - https://github.com/patternfly/patternfly/issues/2021.

mcarrano:
Yes, I do recall that @mcoker . Do you know if this is on the roadmap? It was defined a while ago and I am surprised it was never implemented.

mcoker:
@mcarrano it's slated for the 2020.08 milestone - June 23.

mcoker:
@mcarrano this PR is ready for review. Our modal now behaves like a card, where if you simply omit the title/body/footer, it’s a box (with no padding) to put things in. Users only have to specify the size variation of the modal to use (large). The wizard detects that it is in a modal and retains its existing height (762px), and shrinks with the window. I tested on large and small screens and it looks good. There is only 16px padding around the window when the modal shrinks to fit the window size, so not much real estate lost from how the existing wizard goes to the edges of the window on small screens.

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -1,23 +1,6 @@
...
-  --pf-c-wizard--m-in-page--Height: 100%;
-  --pf-c-wizard--m-in-page--Width: auto;
-  --pf-c-wizard--m-in-page--lg--MaxWidth: none;
-  --pf-c-wizard--m-in-page--lg--MaxHeight: none;
-
-  @media screen and (min-width: $pf-global--breakpoint--lg) {
-    --pf-c-wizard--Width: var(--pf-c-wizard--lg--Width);
-    --pf-c-wizard--Height: var(--pf-c-wizard--lg--Height);
-  }
+  --pf-c-modal-box--c-wizard--Height: #{pf-size-prem(762px)};
```
christiemolloy, line 3: this is such a specific height, maybe add a comment explaining why it is this height

christiemolloy:
Release notes look great!

PR #2965, opened 2020-04-15
fix(table-row-height): fixed borders, row height, focus

### LINKED ISSUE #1948: Update table font-weight var

### LINKED ISSUE #2205: Bug in table cell word-break

### LINKED ISSUE #2261: Address table component tr/td height

### LINKED ISSUE #2265: Table - horizontal scroll issue
relates to #2205 

Because `.pf-c-table` relies on a breakpoint to change layout to `grid`, in some instances the table layout breaks down before it hits the prescribed breakpoint because it contains a lot of information. 

Screen Shot 2019-09-13 at 10 37 41 AM

Suggestion:
```
<div class="pf-c-table-container">
  <table class="">
    content
  </table>
</div>

.pf-c-table-container {
  overflow-x: auto;
}
```

### LINKED ISSUE #2869: Fix - table borders
Table borders are beveled. 

Screen Shot 2020-03-24 at 10 03 21 AM

**Task**
- Align border implementation with [tabs](https://patternfly-next-pr-2757.surge.sh/documentation/core/components/tabs/) for `::before` and `::after`
- Example
  - `--pf-c-table__compound-expansion-toggle--after--BorderRightWidth: 0`
  - `--pf-c-table__compound-expansion-toggle--after--BorderBottomWidth: 0`
  - `--pf-c-table__compound-expansion-toggle--after--BorderLeftWidth: 0`

mattnolting:
**Preview:** https://patternfly-pr-2965.surge.sh/documentation/core/components/table

fixes #2869 
fixes #2265 
fixes #2261
fixes #2205 
fixes #1948 

- removed `.pf-m-expanded` from compound expansion table example.

## Breaking changes
This PR fixes multiple table issues. 
* Adds dedicated table button `.pf-c-table__button`: this button relies on `::after` pseudo element to define clickable area, as previous height hacks cause cross browser bugs. `focus-ring` is disabled on this button and applied to the parent table cell.
* Hover, focus, active and selected states `.pf-c-table__sort` and `.pf-c-table__compound-expansion-toggle`: updates border discrepancies using `::before` and `::after` pseudos and negative offsets.
* Adds text wrapping control `.pf-m-truncate`, `.pf-m-wrap`, `.pf-m-no-wrap`, `.pf-m-break-word` and `.pf-m-fit-content`.
* Adds `.pf-c-table__text` wrapper for grid layout truncation/wrapping control.

// Will update when approved

The following classes have been removed. Any reference to them should be removed as they are no longer used in patternfly:
* `.`

* `--`

The following variables have been renamed. Any reference to the old variable should be updated to use the new variable:
* `--pf-c-` has changed to `--pf-c-`

Some of the general CSS structure has also changed, so please reference the [file changeset](https://github.com/patternfly/patternfly/pull/2965/files).

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -299,10 +299,10 @@ import './Table.css'
         <input type="checkbox" name="{{table--id}}-checkrow2" aria-labelledby="{{concat table--id '-node2'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node2">Node 2</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -327,10 +327,10 @@ import './Table.css'
         <input type="checkbox" name="{{table--id}}-checkrow3" aria-labelledby="{{concat table--id '-node3'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node3">Node 3</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -355,10 +355,10 @@ import './Table.css'
         <input type="checkbox" name="{{table--id}}-checkrow4" aria-labelledby="{{concat table--id '-node4'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node4">Node 4</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -435,10 +435,10 @@ When header cells are empty or they contain interactive elements, `<th>` should 
         <input type="checkbox" name="{{table--id}}-checkrow1" aria-labelledby="{{concat table--id '-node1'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node1">Node 1</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -474,10 +474,10 @@ When header cells are empty or they contain interactive elements, `<th>` should 
         <input type="checkbox" name="{{table--id}}-checkrow2" aria-labelledby="{{concat table--id '-node2'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node2">Node 2</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -511,10 +511,10 @@ When header cells are empty or they contain interactive elements, `<th>` should 
         <input type="checkbox" name="{{table--id}}-checkrow3" aria-labelledby="{{concat table--id '-node3'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node3">Node 3</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -547,10 +547,10 @@ When header cells are empty or they contain interactive elements, `<th>` should 
         <input type="checkbox" name="{{table--id}}-checkrow4" aria-labelledby="{{concat table--id '-node4'}}">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node4">Node 4</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -1372,10 +1372,10 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
         <input type="checkbox" name="{{table--id}}-checkrow2" aria-labelledby="{{table--id}}-node2">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node2">Node 2</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -1396,10 +1396,10 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
         <input type="checkbox" name="{{table--id}}-checkrow3" aria-labelledby="{{table--id}}-node3">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node3">Node 3</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -1420,10 +1420,10 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
         <input type="checkbox" name="{{table--id}}-checkrow4" aria-labelledby="{{table--id}}-node4">
       {{/table-td}}
       {{#> table-th table-th--data-label="Repository name"}}
-        {{#> table-cell-content}}
+        <div>
           <div id="{{table--id}}-node4">Node 4</div>
           <a href="#">siemur/test-space</a>
-        {{/table-cell-content}}
+        </div>
       {{/table-th}}
       {{#> table-td table-td--data-label="Branches"}}
         10
@@ -1603,40 +1603,6 @@ To better control table cell behavior, PatternFly provides a series of modifiers
       {{/table-td}}
       {{#> table-td table-td--modifier="pf-m-nowrap" table-td--data-label="No wrap"}}
         <a href="#">No wrap</a>
-      {{/table-td}}
-    {{/table-tr}}
-  {{/table-tbody}}
-{{/table}}
-```
-
-## Custom text control `.pf-c-table__cell-content`
-
-For granular control over text elements, `.pf-c-table__cell-content` is available. The `.pf-c-table__cell-content` element breaks content text out of the table flow and allows content to respond to `.pf-m-width-` modifiers set on parent elements. It also allows for content customization and individual `.pf-c-table__text` modifications. Modifiers can be applied to the parent `cell` or individual `.pf-c-table__text` elements.
-
-```hbs title=Controlling-text-elements-individually
-{{#> table table--id="controlling-text-elements-individually" table--grid="true" table--modifier="pf-m-grid-md" table--attribute='aria-label="This is a table with custom text wrappers example."'}}
-  {{#> table-thead}}
-    {{#> table-tr}}
-      {{#> table-th table-th--attribute='scope="col"'}}
-        Custom text cells
-      {{/table-th}}
-    {{/table-tr}}
-  {{/table-thead}}
-
-  {{#> table-tbody}}
-    {{#> table-tr}}
-      {{#> table-td table-td--data-label="No wrapping"}}
-        {{#> table-cell-content}}
-          {{#> table-text table-text--type="div" table-text--modifier="pf-m-truncate"}}
-            This cell's contents will truncate. Here's some more text to demonstrate how.
-          {{/table-text}}
-          {{#> table-text table-text--type="div" table-text--modifier="pf-m-nowrap"}}
-            <a href="#">No wrap link example</a>
-          {{/table-text}}
-          {{#> table-text table-text--type="div"}}
-            This cell's contents will wrap. Here's some more text to demonstrate how.
-          {{/table-text}}
-        {{/table-cell-content}}
       {{/table-td}}
     {{/table-tr}}
   {{/table-tbody}}
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -625,11 +625,8 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
     {{/table-tr}}
   {{/table-thead}}
 
-  {{#> table-tbody table-tbody--modifier="pf-m-expanded"}}
+  {{#> table-tbody}}
     {{#> table-tr table-tr--expanded="true"}}
-      {{#> table-th table-th--data-label="Repository name"}}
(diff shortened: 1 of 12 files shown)


src/patternfly/components/Table/table-grid.scss
```diff
@@ -37,16 +37,16 @@
...
-  // ==================================================================
-
   // Table
   --pf-c-table--responsive--BorderColor: var(--pf-global--BorderColor--300);
 
   // Body
   --pf-c-table-tbody--responsive--MarginTop: var(--pf-global--spacer--xs);
   --pf-c-table-tbody--m-expanded--before--Top: var(--pf-global--spacer--sm);
   --pf-c-table-tbody--responsive--BorderWidth: var(--pf-global--spacer--sm);
+  --pf-c-table--tbody--after--BorderWidth: var(--pf-global--BorderWidth--lg);
```
christiemolloy, line 47: should this be `pf-c-table--tbody` or `pf-c-table-tbody`, if we're going with `pf-c-table--tbody` looks like we should update it everywhere

src/patternfly/components/Table/table-grid.scss
```diff
@@ -105,16 +111,40 @@
...
+  }
+
+  // apply modifications to text
+  .pf-c-table__text {
+    position: relative;
+    display: block;
+    width: auto;
+    min-width: 0;
+    max-width: none;
+    overflow: var(--pf-c-table--cell--Overflow);
```
christiemolloy, line 143: could you use the overflow mixin here ?

src/patternfly/components/Table/table-grid.scss
```diff
@@ -316,33 +314,36 @@
...
   // Expandable row
   .pf-c-table__expandable-row {
     --pf-c-table-cell--responsive--PaddingTop: 0;
     --pf-c-table-cell--responsive--PaddingRight: 0;
     --pf-c-table-cell--responsive--PaddingBottom: 0;
     --pf-c-table-cell--responsive--PaddingLeft: 0;
-    --pf-c-table-cell--PaddingRight: 0;
-    --pf-c-table-cell--PaddingLeft: 0;
+    --pf-c-table--cell--PaddingRight: 0;
+    --pf-c-table--cell--PaddingLeft: 0;
```
christiemolloy, line 346: as per comment above, if we're going with `pf-c-table--cell` we should update all of the var names

src/patternfly/components/Table/table.scss
```diff
@@ -30,53 +29,100 @@
...
   // Table cell - toggle
   --pf-c-table__toggle--c-button--MarginTop: calc(#{pf-size-prem(6px)} * -1);
   --pf-c-table__toggle--c-button__toggle-icon--Transform: rotate(270deg);
   --pf-c-table__toggle--c-button__toggle-icon--Transition: .2s ease-in 0s;
   --pf-c-table__toggle--c-button--m-expanded__toggle-icon--Transform: rotate(360deg);
+
+  // Button
+  --pf-c-table__button--BackgroundColor: transparent;
+  --pf-c-table__button--Color: var(--pf-global--Color--100);
+  --pf-c-table__button--hover--Color: var(--pf-global--Color--100);
```
christiemolloy, line 106: do we need all of these if the color is the same for default and hover/active/focus

src/patternfly/components/Table/table-grid.scss
mattnolting: Great point! To limit the amount of potential breaks, I'll open another issue to update all variables to our current spec! https://github.com/patternfly/patternfly/issues/2983

src/patternfly/components/Table/table-grid.scss
mattnolting: I think using the overflow mixin wouldn't apply to this use case as I'm using the variables specifically to pass the values from the parent `th/td` to the selector. Looks like the overflow mixin would just hide the overflow, rather than respond to the new modifiers.

src/patternfly/components/Table/table.scss
mattnolting: Yes, I think so. @mcoker wdyt?

mcoker: If we anticipate that this could change, I think so. If we set all of the states with 1 variable, then decide to introduce a new var for one of the states, since the initial var that was used no longer controls that state, it's technically a breaking change.

src/patternfly/components/Table/table-grid.scss
```diff
@@ -316,33 +314,36 @@
       top: 0;
       bottom: 0;
       left: 0;
-      width: var(--pf-c-table__expandable-row--before--Width);
       content: "";
+      border: 0;
+      border-left: var(--pf-c-table--tbody--after--BorderColor) solid var(--pf-c-table--tbody--after--BorderLeftWidth);
```
mcoker, line 319:
```suggestion
      border-left: var(--pf-c-table--tbody--after--BorderLeftWidth) solid var(--pf-c-table--tbody--after--BorderColor);
```

src/patternfly/components/Table/table.scss
```diff
@@ -212,20 +275,39 @@
...
+
+    // Set default th behavior to truncate
+    --pf-c-table--cell--MinWidth: var(--pf-c-table--m-truncate--cell--MinWidth);
+    --pf-c-table--cell--MaxWidth: var(--pf-c-table--m-truncate--cell--MaxWidth);
+    --pf-c-table--cell--Overflow: var(--pf-c-table--m-truncate--cell--Overflow);
+    --pf-c-table--cell--TextOverflow: var(--pf-c-table--m-truncate--cell--TextOverflow);
+    --pf-c-table--cell--WhiteSpace: var(--pf-c-table--m-truncate--cell--WhiteSpace);
+
+    // Align content to bottom
+    --pf-c-table__button-content--AlignItems: var(--pf-c-table--thead__button-content--AlignItems);
```
mcoker, line 290: does it help for this to be a var? Looks like it's only used once in a property declaration, and the variable here as the value is only declared once.

src/patternfly/components/Table/table.scss
```diff
@@ -262,32 +344,176 @@
       // explicitly reset padding rather than css variable, as the button within uses the variable
       padding: 0;
     }
+  }
+
+  // set property here to increase specificity
+  .pf-c-table__sort {
+    min-width: var(--pf-c-table__sort--MinWidth);
+  }
+
```
mcoker, line 353: these modifiers tho! 🔥

src/patternfly/components/Table/examples/Table.md
```diff
@@ -1705,12 +1655,293 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
   {{/table-tbody}}
 {{/table}}
 ```
+
+## Controlling text
```
christiemolloy, line 1659: I think this is a great summary!

src/patternfly/components/Table/examples/Table.md
```diff
@@ -1705,12 +1655,293 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
...
+
+To better control table cell behavior, PatternFly provides a series of modifiers to help contextually control layout. By default, `thead` cells are set to truncate, whereas `tbody` cells are set to wrap. Both `th` and `td` cells use a set of shared css properties mapped to customizable css variable values. Because only the shared css variables are changed by the modifier selector and not the properties, the modifier can be applied to any parent element up until `.pf-c-table` itself [`thead`, `tbody`, `tr`, `th`, `td`, `.pf-c-table__text`].
+
+<ul>
+  <li>
+    <b>pf-m-wrap:</b> Sets text to wrap. This is the default behavior for <code>tbody</code> cells.
```
christiemolloy, line 1665: I think it would be helpful here to add a table of what these modifiers apply to. because im looking here and then im like ok now i have to search for which modifier to use for table header or for table cell, so maybe you could just add the documentation for all these modifiers here

christiemolloy:
Could you make this a bit more intuitive, maybe make it more of a table of context for what you can find in the table component

Screen Shot 2020-04-21 at 2 47 24 PM

christiemolloy:
This arrow in expandable is pointing down but it should be up, if you look at compound expansion its different

Screen Shot 2020-04-21 at 2 47 59 PM

christiemolloy:
could you re-format this and also make it so that the user knows that this is still part of controlling text, maybe you could split them up into part A and part B

Screen Shot 2020-04-21 at 2 53 55 PM

christiemolloy:
Maybe after the th truncation example you could bring down the headers that wrap example (seeing as that isn't the default).
and maybe similar to how you wrote "Controlling text" you could have a section on Headers and then have the modifications for the headers there so that they all follow on from each other

christiemolloy:
This is really cool!!!! Can you add it back up to the controlling text section

Screen Shot 2020-04-21 at 2 57 56 PM

christiemolloy:
Should this be scrollable on mobile?

Screen Shot 2020-04-21 at 3 16 26 PM

src/patternfly/components/Table/table-grid.scss
```diff
@@ -112,16 +118,40 @@
...
   display: grid;
   border: none;
 
+  // reset cell modifications
+  &.pf-c-table tr > * {
+    width: auto;
+    min-width: 0;
+    max-width: none;
+    overflow: visible;
+    text-overflow: clip;
```
christiemolloy, line 139: why does the default text-overflow for mobile need to be set to `clip`

src/patternfly/components/Table/table-grid.scss
```diff
@@ -316,33 +314,36 @@
...
       content: "";
+      border: 0;
+      border-left: var(--pf-c-table--tbody--after--BorderColor) solid var(--pf-c-table--tbody--after--BorderLeftWidth);
     }
 
     &.pf-m-expanded {
-      &::after {
-        background-color: var(--pf-c-table__expandable-row--before--BackgroundColor);
-      }
+      --pf-c-table--tbody--after--BorderLeftWidth: var(--pf-c-table--tbody--after--BorderWidth);
```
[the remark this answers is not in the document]
christiemolloy: not sure if this was fixed ^

src/patternfly/components/Table/table-grid.scss
```diff
@@ -410,6 +411,13 @@
     padding-right: 0;
   }
 
+  .pf-c-table__button {
+    --pf-c-table--cell--PaddingTop: var(--pf-c-table--m-grid--cell--PaddingTop);
```
christiemolloy, line 415: I think you'll need to change this var name

src/patternfly/components/Table/table.scss
```diff
@@ -12,49 +12,57 @@
...
+  --pf-c-table--cell--Color: var(--pf-global--Color--100);
+  --pf-c-table--cell--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-table--cell--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-table--cell--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-table--cell--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-table--cell--first-last-child--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-table--cell--first-last-child--PaddingRight: var(--pf-global--spacer--lg);
+
+  // Default cell variables
+  --pf-c-table--cell--MinWidth: 0;
```
christiemolloy, line 51: did you set all these same vars in table-grid.scss ? could you use these vars in that file to keep consistent so if something changes here you dont have to update the grid file?

src/patternfly/components/Table/table.scss
```diff
@@ -12,49 +12,57 @@
...
+  --pf-c-table--cell--MaxWidth: none;
+  --pf-c-table--cell--Width: auto;
+  --pf-c-table--cell--Overflow: visible;
+  --pf-c-table--cell--TextOverflow: clip;
+  --pf-c-table--cell--WhiteSpace: normal;
+  --pf-c-table--cell--WordBreak: normal;
+
+  // Truncate
+  --pf-c-table--m-truncate--cell--MinWidth: 10ch;
```
christiemolloy, line 60: should the truncate modifier be set directly on the table or should it be set on the `<th>` or the `<td>`

src/patternfly/components/Table/table-grid.scss
mattnolting: Just resetting to its default value

src/patternfly/components/Table/table.scss
```diff
@@ -12,49 +12,57 @@
...
-  --pf-c-table-cell--first-last-child--xl--PaddingLeft: var(--pf-global--spacer--md);
-  --pf-c-table-cell--first-last-child--xl--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-table--cell--FontSize: var(--pf-global--FontSize--md);
+  --pf-c-table--cell--FontWeight: var(--pf-global--FontWeight--normal);
+  --pf-c-table--cell--Color: var(--pf-global--Color--100);
+  --pf-c-table--cell--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-table--cell--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-table--cell--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-table--cell--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-table--cell--first-last-child--PaddingLeft: var(--pf-global--spacer--lg);
```
mcoker, line 47: looks like we lost the mobile breakpoint changes to the first/last-child. Needs to be `--md`, then `--lg` @ min-width breakpoint--xl

mcoker: grid layout spacing adapts correctly

src/patternfly/components/Table/table-grid.scss
mattnolting: The plan is to update the old vars to this pattern: https://github.com/patternfly/patternfly/issues/2983

mcoker:
Looks like we have an unwanted left blue border on the row with the compound expansion buttons 
Screen Shot 2020-04-21 at 4 03 53 PM

should look like

Screen Shot 2020-04-21 at 4 04 58 PM

src/patternfly/components/Table/table.scss
mattnolting: Yes, I'm using these vars in `table-grid.scss`, but I'm unsetting them on the `th/td` and passing them to the child `.pf-c-table__text`. Otherwise, setting these values on the `th/td`s in grid would destroy the `pf-m-grid` layout. If there's no child `.pf-c-table__text`, then the table layout styles have no affect on content.

src/patternfly/components/Table/table.scss
mattnolting: I left this variable unattached to specific table element so that it could be used on an entire thead, tbody or row so that the modifier would affect all child `th/td`s.

src/patternfly/components/Table/table-grid.scss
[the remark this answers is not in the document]
mattnolting: I'm using the variable from `table.scss` here the same way we're using it in `table.scss` by defining individual borders with `thing--after--BorderWidth` when they're active/expanded. Do you think we need a grid specific var here?

mcoker:
The expanded row is missing the left border

Screen Shot 2020-04-21 at 5 37 51 PM

mcoker:
The link in the compact table examples is breaking, doesn't seem to happen in chrome/safari.

Screen Shot 2020-04-21 at 5 40 33 PM

mcoker:
`.pf-m-fit-content` works to disable truncation on a normal header cell, but when used on a `.pf-c-table__sort` header cell, the text is still truncated.

Screen Shot 2020-04-21 at 5 56 01 PM

mcoker:
`pf-m-truncate` doesn't seem to work on compound expansion buttons - should it? 
Screen Shot 2020-04-21 at 6 00 56 PM

mcoker:
@mattnolting for the bulleted items in the release notes, can you also include actionable notes about why each change is breaking and what exactly needs to change? Like how and where there `pf-c-table__button` element is used, and what a user would need to do to update their table to the new structure. The new features aren't breaking, so unless they're used to accompany or address a breaking change, you don't have to mention them.

Issue #2967, opened 2020-04-16
Move modal padding from container to elements


mcoker:
Create the padding in a modal similar to a card, so that the elements can be removed and the modal can act as a container without padding. This will also enable layouts where one or more body elements do not have padding, which was seen in the table column management demo mockups.

PR #2968, opened 2020-04-16
Issue 2967

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2967

@tlabaj I think it would be more consistent for the modal component title to be it's own element of the modal component (ie, `pf-c-modal-box__title`), instead of using the title component. It looks like the react modal title is built in the component from a `title` prop on the component. How do you feel about removing the dependency on the title component in the modal, and wrapping the title in a new element (`pf-c-modal-box__title`) instead? Would there be much (or any) impact to the consumer? If that's OK I will update this PR.

## Breaking changes
This PR moves the outer padding on the modal from the `.pf-c-modal-box` container to the modal box elements that are children of the modal. To better achieve this spacing, we now wrap the modal's title in a new element `.pf-c-modal-box__header`, so anywhere a modal title is defined, it will need to be wrapped in this new element.

The following variables have been removed. Any references to them should be removed as they are no longer used in patternfly.
* `--pf-c-modal-box--PaddingTop`
  * This can now be achieved using `--pf-c-modal-box__[modal element]--PaddingTop`, however it's important to note that this variable changes depending on the markup present in the modal, so this variable serves multiple purpopses. And it will need to be modified for each child of the modal component that touches the top edge of the modal.
* `--pf-c-modal-box--PaddingRight`
  * This can now be achieved using `--pf-c-modal-box__[modal element]--PaddingRight`. However it will need to be modified for each child of the modal component that touches the right edge of the modal.
* `--pf-c-modal-box--PaddingBottom`
  * This can now be achieved using `--pf-c-modal-box__[modal element]--last-child--PaddingBottom`. However it will need to be modified for each child of the modal component that touches the bottom edge of the modal.
* `--pf-c-modal-box--PaddingLeft`
  * This can now be achieved using `--pf-c-modal-box__[modal element]--PaddingLeft`. However it will need to be modified for each child of the modal component that touches the left edge of the modal.

The following variables have been renamed. Any reference to the old name should be updated to reference the new name.
* `--pf-c-modal-box--body--MinHeight` renamed to `--pf-c-modal-box__body--MinHeight`
* `--pf-c-modal-box__c-title--description--MarginTop` ranamed to `--pf-c-modal-box__header--description--PaddingTop`
* `--pf-c-modal-box__description--body--MarginTop` renamed to `--pf-c-modal-box__description--body--PaddingTop`
* `--pf-c-modal-box--c-title--body--MarginTop` renamed to `--pf-c-modal-box__header--body--PaddingTop`
* `--pf-c-modal-box__footer--MarginTop` renamed to `--pf-c-modal-box__footer--PaddingTop`

### DIFF
--- a/src/patternfly/components/ModalBox/examples/ModalBox.md
+++ b/src/patternfly/components/ModalBox/examples/ModalBox.md
@@ -10,9 +10,11 @@ cssPrefix: pf-c-modal-box
   {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
    <i class="fas fa-times" aria-hidden="true"></i>
   {{/button}}
-  {{#> title titleType="h1" title--modifier="pf-m-2xl" title--attribute='id="modal-title"'}}
-    Modal header
-  {{/title}}
+  {{#> modal-box-header}}
+    {{#> title titleType="h1" title--modifier="pf-m-2xl" title--attribute='id="modal-title-basic"'}}
+      Modal header
+    {{/title}}
+  {{/modal-box-header}}
   {{#> modal-box-body modal-box-body--attribute='id="modal-description"'}}
     To support screen reader user awareness of the dialog text, the dialog text is wrapped in a div that is referenced by aria-describedby.
   {{/modal-box-body}}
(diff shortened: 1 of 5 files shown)


mcoker:
closing in favor of https://github.com/patternfly/patternfly/pull/2969

PR #2999, opened 2020-04-22
fix(nav-toggle-icon): removed nav-toggle-icon

mattnolting:
Removes `pf-c-nav__toggle-icon`

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -314,6 +314,7 @@
 .pf-c-nav__link {
   display: flex;
   align-items: baseline;
+  justify-content: space-between;
   padding: var(--pf-c-nav__link--PaddingTop) var(--pf-c-nav__link--PaddingRight) var(--pf-c-nav__link--PaddingBottom) var(--pf-c-nav__link--PaddingLeft);
   font-size: var(--pf-c-nav__link--FontSize);
   font-weight: var(--pf-c-nav__link--FontWeight);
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Nav/nav.scss
```diff
@@ -183,7 +183,8 @@
   --pf-c-nav__toggle--FontSize: var(--pf-global--icon--FontSize--md);
 
   // Toggle icon
-  --pf-c-nav__toggle-icon--Transition: var(--pf-global--TransitionDuration);
+  --pf-c-nav__toggle--icon--Transition: var(--pf-global--TransitionDuration);
```
christiemolloy, line 186: since we don't have a wrapper that is pf-c-nav__toggle-icon anymore should this just be `pf-c-nav__toggle` ?

src/patternfly/components/Nav/nav.scss
```diff
@@ -558,15 +563,11 @@
...
-}
 
-// Toggle icon
-.pf-c-nav__toggle-icon {
-  display: inline-block;
-  transition: var(--pf-c-nav__toggle-icon--Transition);
-
-  .pf-c-nav__item.pf-m-expanded & {
-    transform: var(--pf-c-nav__item--m-expanded__toggle-icon--Transform);
+  > * {
```
christiemolloy, line 567: is it possible for the toggle wrapper to rotate and not the icon inside since we want to avoid styling the child element

mattnolting:
Keeping `pf-c-nav__toggle-icon`

Issue #3022, opened 2020-04-24
Remove use of RHD font


mcoker:
@mceledonia @kybaker 

We've decided to drop the use of RHD in PF core and use RHT as the single font in PF components.

PR #3023, opened 2020-04-24
fix(global): replaced use of RHD with RHT, remove RHD

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3022

@kybaker @mceledonia @mcarrano the 3 places we used RHD were the title component, content component's h1-h6, and the modal component title. Along with swapping the font from RHD to RHT, I made the font bold, which maps to `RedHatText-Medium`. Let me know what you think and if you would like to make any other updates.

----

@mceledonia had a question about whether our `RedHatText-Medium` was actually `RedHatText-Bold`, because it appears more bold than expected, so I compared our fonts against the ones on https://github.com/RedHatOfficial/RedHatFont/, and verified the `700` font-weight in PF maps to `RedHatText-Medium`, and everything looks correct.

I temporarily introduced an extra font weight that maps to `RedHatText-Bold` (which we don't use), to demonstrate what our `RedHatText-Medium` looks like in comparison. This is regular -> medium -> bold.

Screen Shot 2020-04-24 at 6 49 00 PM


mceledonia:
The medium weight in the shot above looks correct, thanks for testing that @mcoker. Also thanks for listing the areas we used RHD, I checked those and they're all looking good.

PR #3026, opened 2020-04-27
fix(switch): clean up vars

### LINKED ISSUE #2608: Clean up switch vars and properties
```
  --pf-c-switch__toggle-icon--PaddingLeft: calc(var(--pf-c-switch--FontSize) * .4); // during breaking change period, remove this and make it the value for `left`
  --pf-c-switch__label--FontSize: var(--pf-c-switch--FontSize); // remove at breaking change period
  --pf-c-switch__label--FontWeight: var(--pf-global--FontWeight--normal); // this can be removed at breaking change period
  --pf-c-switch__label--LineHeight: var(--pf-c-switch--LineHeight); // this can be removed at breaking change period
  font-weight: var(--pf-c-switch__label--FontWeight); // remove at breaking change
  line-height: var(--pf-c-switch__label--LineHeight); // remove at breaking change
```

Also just check the var order overall to make sure it follows other components. There was some reorganization of the vars that broke our standard convention so we could get the vars to work with our ie11 script in this PR - https://github.com/patternfly/patternfly-next/pull/2610/files

Since we're removing the ie11 script, those updates are no longer necessary.

christiemolloy:
closes #2608 

## Breaking Changes

TBD


src/patternfly/components/Switch/switch.scss
```diff
@@ -28,6 +27,7 @@
   --pf-c-switch__input--focus__toggle--OutlineColor: var(--pf-global--primary-color--100);
 
   // Switch toggle
+  --pf-c-switch__toggle--Height: var(--pf-c-switch--Height);
```
mcoker, line 30: Can this be updated to match the value of the `--pf-c-switch--Height` var? Since `--pf-c-switch--Height` can be used to set the value of `--pf-c-switch__toggle--Height` now, if we choose to give `--pf-c-switch__toggle--Height` a custom value at some point, and `--pf-c-switch--Height` no longer sets the toggle height, that would technically be a breaking change.

src/patternfly/components/Switch/switch.scss
```diff
@@ -149,7 +144,6 @@
   position: absolute;
   top: var(--pf-c-switch__toggle-icon--Top);
```
mcoker, line 145: Looking at the react switch, the icon is misaligned - https://www.patternfly.org/v4/documentation/react/components/switch/

One way to fix it would be to replace this block with below and remove `--pf-c-switch__toggle-icon--Top` and `--pf-c-switch__toggle-icon--Transform`

```
.pf-c-switch__toggle-icon {
  position: absolute;
  top: 0;
  bottom: 0;
  left: var(--pf-c-switch__toggle-icon--Left);
  display: flex;
  align-items: center;
  padding-left: var(--pf-c-switch__toggle-icon--PaddingLeft);
  font-size: var(--pf-c-switch__toggle-icon--FontSize);
  color: var(--pf-c-switch__toggle-icon--Color);
}
```

src/patternfly/components/Switch/switch.scss
christiemolloy: I updated it, a lot of the toggle vars rely on `pf-c-switch--Height` so I'm wondering if I should update all of them to represent the value? But at the same time some of them should be based on the height of the switch. I could see the toggle being a var that is potentially changed though so im in favor of updating that.

but something like: `--pf-c-switch__toggle--BorderRadius: var(--pf-c-switch--Height);` doesn't really need to rely on the switch's height

christiemolloy:
thanks for the feedback, updated @mcoker

src/patternfly/components/Switch/switch.scss
mcoker: Yeah good call, not sure why that is set. That could just be `var(--pf-global--BorderRadius--lg)`.

src/patternfly/components/Switch/switch.scss
```diff
@@ -147,19 +140,19 @@
 
 .pf-c-switch__toggle-icon {
   position: absolute;
-  top: var(--pf-c-switch__toggle-icon--Top);
+  top: 0;
+  bottom: 0;
   left: var(--pf-c-switch__toggle-icon--Left);
+  display: flex;
+  align-items: center;
   padding-left: var(--pf-c-switch__toggle-icon--PaddingLeft);
```
mcoker, line 148: you can remove this

Issue #2989, opened 2020-04-20
Add an 80px spacer


mcarrano:
We've had a design request to add an 80px spacer for websites that desire larger padding between components. See https://github.com/patternfly/patternfly-design/issues/827 for the design discussion.

cc @mcoker @LHinson

coreyvickery:
@mcarrano Any movement on this yet?

mcarrano:
I believe that this is part of our current milestone. Can you confirm @mcoker ?

PR #3027, opened 2020-04-27
feat(spacer): added 80px spacer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2989


coreyvickery:
Maybe not the right place for this question, but was a color ever decided upon for the new 80px spacer?

mceledonia:
@coreyvickery Yes I believe @gdoyle1 came up with a nice pink/purple which worked quite well!

mcoker:
@coreyvickery yep https://github.com/patternfly/patternfly-design-kit/issues/205#issuecomment-620837809

> We're using #B300B3 as the fill and #8F008F as the 1px border

PR #3028, opened 2020-04-27
fix(button): clean up vars

### LINKED ISSUE #2281: Update button component border vars to include pseudo element
Right now the vars are `--pf-c-button--Border[property]` and should technically be `--pf-c-button--before--Border`

christiemolloy:
closes #2281 

## Breaking Changes

TBD


src/patternfly/components/Button/button.scss
```diff
@@ -151,29 +152,29 @@
...
-    border-radius: var(--pf-c-button--BorderRadius);
+    border: var(--pf-c-button--after--BorderWidth) solid;
+    border-color: var(--pf-c-button--after--BorderColor);
+    border-radius: var(--pf-c-button--after--BorderRadius);
   }
 
   &:hover {
     text-decoration: none;
 
     &::after {
```
mcoker, line 163: you shouldn't need to specify this here or any of the other selectors. You can override the variable with the state selector.

mcoker: I was trying something and made that change in a PR here if it helps https://github.com/christiemolloy/patternfly-next/pull/25

PR #3029, opened 2020-04-27
fix(docs): bump theme to fix top nav

redallen:
Change top nav from always showing scroll buttons to conditionally showing them.

Before:
image
After:
image


mcoker:
This doesn't fix it in PR preview as far as I can tell. Is it supposed to?

Screen Shot 2020-04-27 at 8 00 47 PM

redallen:
Ah, good point! I need to bump React as well in this PR.

redallen:
Should be fixed now @mcoker :
image

PR #3031, opened 2020-04-27
fix(table-modifier-bugs): fixed truncate and wrap mod bugs

### LINKED ISSUE #3004: Table bugs
Top padding on `thead` cells in compact table should be `12px`. Looks like everything's in place, but the `thead` specific var declarations are being overridden due to specificity.

Checkbox in table header is misaligned in FF
Screen Shot 2020-04-22 at 3 08 30 PM

Compound expansion grid view is missing blue left expanded border. Also there are dividers between rows in the expansion that weren't there before.

Before:
Screen Shot 2020-04-22 at 3 11 00 PM

After:
Screen Shot 2020-04-22 at 3 11 03 PM

The bottom border under the compound expansion buttons is missing if the first button is expanded in safari:

Screen Shot 2020-04-22 at 3 16 04 PM

^ Should look like this for comparison (chrome)

Screen Shot 2020-04-22 at 3 28 41 PM

Sortable and non-sortable headers are mis-aligned in firefox

Screen Shot 2020-04-22 at 3 44 49 PM

Screen Shot 2020-04-22 at 3 55 16 PM

mattnolting:
closes #3004 

Updates include 
- Fixed truncation is compound expansion cells
- Added examples for nested `.pf-c-table__text`

### DIFF
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -151,6 +151,10 @@
     white-space: var(--pf-c-table--cell--WhiteSpace);
   }
 
+  // [class*="pf-l"] {
+  //   width: 100%;
+  // }
+
   [class*="pf-l"],
   [class*="pf-l"] > * {
     max-width: 100%;
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -131,7 +131,7 @@
   border: none;
 
   // reset cell modifications
-  &.pf-c-table tr > * {
+  tr > * {
     width: auto;
     min-width: 0;
     max-width: none;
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Table/table-grid.scss
```diff
@@ -152,6 +151,11 @@
     white-space: var(--pf-c-table--cell--WhiteSpace);
   }
 
+  [class*="pf-l"],
+  [class*="pf-l"] > * {
```
mcoker, line 155: did you mean to leave this in?

mattnolting: I did, but using `minmax()` for grid layout columns _seems_ to fix this. Nice catch!

christiemolloy:
I think this looks a bit odd, what was the reason for having the first-child have a blue-border when its expanded?
Screen Shot 2020-05-06 at 10 37 23 AM

mattnolting:
@christiemolloy the current implementation is presented this way, but the examples don't show the first cell as a `__toggle`. 

Screen Shot 2020-05-06 at 10 56 18 AM

PR #3037, opened 2020-04-28
fix(components): split out transform variables

### LINKED ISSUE #2991: Split out --Transform vars to individual transform functions
Firefox already supports it, but soon browsers will allow us to define the individual transform functions as properties instead of values in a `transform` declaration. To prepare for that, and just to make it easier for patternfly and consumers to make adjustments to individual properties instead of having to redeclare all transform functions in a single variable, let's look anywhere we've defined a `--Transform` var and turn it into var(s) for the function(s) it contains.

christiemolloy:
closes #2991 

## Breaking changes

TBD


mcoker:
Great job @christiemolloy! This looks really good. The only thing I'm uncertain about is we have a mix of `translateX`, `translateY`, `trasnlate3d` and `translate`.

FWIW, the proposed property in the [draft for individual properties](https://drafts.csswg.org/css-transforms-2/#individual-transforms) is just `translate`, and it will take 3 values - X, Y, and Z.

I'm wondering if we should have a variable for X and Y values where we set `translate` in popover and tooltip, and a var for X where we set `translate3d` in the page component. In popover/tooltip, we would set `translate: var(--translateX) var(--translateY);` and in page we would set `translate: var(--translateX) 0 0` (specifying the z transform for performance reasons).

What do you think?

mcoker:
Also looks like drawer, tabs, file upload, and spinner have transforms, too.

mcoker:
@christiemolloy I submitted a PR here with what I was thinking https://github.com/christiemolloy/patternfly-next/pull/26

The idea is that

`--component--Rotate: 90deg;`

Allows us to define it like this now:

`transform: rotate(var(--component--Rotate));`

And when `rotate` is a property, it will become:

`rotate: var(--component--Rotate);`

Issue #2701, opened 2020-02-11
Update the disabled state for the button control component


christiemolloy:
Currently this is the disabled state for `.pf-c-button.pf-m-control`

<img width="694" alt="Screen Shot 2020-02-11 at 4 52 37 PM" src="https://user-images.githubusercontent.com/20118816/74282921-fab03200-4cee-11ea-8bb3-079ed94a198b.png">

@lboehling had designed the file upload component with a control button with a disabled state that looks like the following:

<img width="107" alt="Screen Shot 2020-02-11 at 4 55 07 PM" src="https://user-images.githubusercontent.com/20118816/74283072-47940880-4cef-11ea-8dcd-b61b700d1355.png">

I spoke with @mcarrano and he agreed with Lucia and I that we should update the disabled state for `.pf-c-button.pf-m-control` so that it aligns with Lucia's designs.

Wanted to get another opinion. @mceledonia what are your thoughts here?

christiemolloy:
After this is completed will need to update #2624 to remove the variable overrides.

Issue #2997, opened 2020-04-21
Border removed once Button is disabled inside another component


nixocio:
When the Button prop `isDisabled` is set to True inside another component, the `border` of Button is removed.

![border_button](https://user-images.githubusercontent.com/9053044/79913343-6ef30a00-83f1-11ea-8b46-418f8c909e6d.gif)

PF Version

```text
    "@patternfly/patternfly": "^2.71.3",
    "@patternfly/react-core": "^3.153.3",
    "@patternfly/react-icons": "^3.15.15",
    "@patternfly/react-tokens": "^2.8.12",
    "ansi-to-html": "^0.6.11",
```

mcoker:
cc @mcarrano, this is a result of the control button's disabled state styling

mcarrano:
It does seem odd @mcoker Can you recall why we did that? cc @mceledonia

mcoker:
@mcarrano nevermind, we have an issue where we're going to update that styling that we will get in this milestone - https://github.com/patternfly/patternfly/issues/2701

mcoker:
Hi @nixocio! Just wanted to follow up on this and say that we have updates coming to the button's disabled state that should resolve this issue. The PR is https://github.com/patternfly/patternfly/pull/3049, and I have put together a demo that shows how that update would in the use case you reported in this issue. https://codepen.io/mcoker/pen/xxwpeEa

Do you mind looking that over and letting me know if that works as you would expect it?

nixocio:
hi @mcoker, thanks for working on this. I will take a look.

nixocio:
It seems to be fixed! 

![search_icon_fixed](https://user-images.githubusercontent.com/9053044/81195081-60792680-8f8b-11ea-960e-40fd8d85255e.gif)

mcoker:
@nixocio great, thanks for verifying! This new design will be available in our upcoming v4 release. If you need the design now, you can use the CSS I provided in the codepen to set the `border-radius` and disabled `background-color`.

mcoker:
resolved via https://github.com/patternfly/patternfly/pull/3049

Issue #3048, opened 2020-04-29
Improve disabled styles in File Upload component


mturley:
Discussing with @christiemolloy and @mcarrano, we agreed that the File Upload component looks a little weird when it is disabled:

<img width="558" alt="Screenshot 2020-04-29 16 31 56" src="https://user-images.githubusercontent.com/811963/80645656-6cb82d80-8a39-11ea-9e95-fbc8b4bd8050.png">

Since there is no border on the bottom of the disabled filename input field, it blends with the textarea below into one big box, making the buttons look out of place. @christiemolloy proposed altering the `:disabled` style for this specific class of input fields (only within a file upload), so that it has the bottom border that it currently does when it is `read-only`.

mcoker:
I made a similar note in https://github.com/patternfly/patternfly/pull/3049#issuecomment-621469098.

mceledonia:
Something like this might be more ideal but I can't think of a way to do it that fits well within our initial approach. Is this something that would be a good use case for an override/exception? i.e. if disabled and if an adjacent element is disabled, use this styling?

![image](https://user-images.githubusercontent.com/30935938/81000959-f8a0cf80-8e14-11ea-9a86-8ae40ee21973.png)

mcoker:
After meeting with @mceledonia and @mcarrano, we decided to go forward manually defining a bottom border on disabled elements in the file upload input group.

mcarrano:
@mcoker is this still an issue?

mcoker:
I don't believe so. This is what it looks like when the component is disabled. I want to say this was resolved when we removed the bottom border from disabled control variation buttons. Unless we still want a border between the input group and textarea?
 
<img width="804" alt="Screen Shot 2020-08-06 at 5 19 20 PM" src="https://user-images.githubusercontent.com/35148959/89588520-6cce8300-d809-11ea-995d-079a0871d2a9.png">

PR #3049, opened 2020-04-29
fix(button): updated control variation disabled state

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2701

#### Button

This PR updates the visuals of the control button disabled state.

* The following variables were removed. All references to them should be removed as they are no longer used in patternfly.
  * `--pf-c-button--m-control--disabled--after--BorderTopColor`
  * `--pf-c-button--m-control--disabled--after--BorderRightColor`
  * `--pf-c-button--m-control--disabled--after--BorderBottomColor`
  * `--pf-c-button--m-control--disabled--after--BorderLeftColor`

#### Form Control

* `--pf-c-form-control--readonly--focus--BackgroundColor` was renamed to `--pf-c-form-control--readonly--BackgroundColor`. All references to the old variable name should be updated to reference the new variable name instead.
* `--pf-c-form-control--readonly--focus--PaddingBottom` was removed. Any reference to it should be removed as it is no longer used in patternfly.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -108,10 +108,6 @@
   --pf-c-button--m-control--after--BorderBottomColor: var(--pf-global--BorderColor--200);
   --pf-c-button--m-control--after--BorderLeftColor: var(--pf-global--BorderColor--300);
   --pf-c-button--m-control--disabled--BackgroundColor: var(--pf-global--disabled-color--300);
-  --pf-c-button--m-control--disabled--after--BorderTopColor: transparent;
-  --pf-c-button--m-control--disabled--after--BorderRightColor: transparent;
-  --pf-c-button--m-control--disabled--after--BorderBottomColor: var(--pf-global--BorderColor--200);
-  --pf-c-button--m-control--disabled--after--BorderLeftColor: transparent;
   --pf-c-button--m-control--hover--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-button--m-control--hover--Color: var(--pf-global--Color--100);
   --pf-c-button--m-control--hover--after--BorderBottomWidth: var(--pf-global--BorderWidth--md);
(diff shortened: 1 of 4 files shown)


mcoker:
@mcarrano @lboehling it's worth noting that this looks as intended when the disabled button is beside a non-disabled form control element, like in the file upload example

Screen Shot 2020-04-29 at 4 14 27 PM

Or this input group example

Screen Shot 2020-04-29 at 4 13 20 PM

But since the disabled control button has a bottom border, and disabled form control elements do not, if the form control element is also disabled, it looks like this

Screen Shot 2020-04-29 at 4 13 09 PM

mcoker:
@christiemolloy file upload is beta, so I think we can address that one separately as long as we're OK making that as an update specifically to the file upload component. Here's the issue for reference - https://github.com/patternfly/patternfly/issues/3048

That said, per my comment https://github.com/patternfly/patternfly/pull/3049#issuecomment-621469098, I don't think it's a file upload specific problem. It looks like an issue with either the input group, or just that disabled form control elements have different styling than disabled control buttons.

@mcarrano @lboehling @mceledonia can you confirm ^^?

mcoker:
@christiemolloy @mattnolting after talking to @mcarrano and @mceledonia we decided to remove the bottom border from the disabled control button so it matches disabled form control elements. Made that update, and also noticed the bottom border of `readonly` form control elements was wrong, so updated that to match the regular form control border, which matches the control button.

christiemolloy:
This works, the File Upload does look strange now where the disabled buttons used to have a bottom border. @mcarrano do you think we should add to the issue that @mturley created to add a bottom border to the disabled elements in the File Component?

Screen Shot 2020-05-05 at 3 25 47 PM
Screen Shot 2020-05-05 at 3 25 57 PM

mcoker:
> This works, the File Upload does look strange now where the disabled buttons used to have a bottom border. @mcarrano do you think we should add to the issue that @mturley created to add a bottom border to the disabled elements in the File Component?

@christiemolloy yep that's the plan, updated that issue - https://github.com/patternfly/patternfly/issues/3048#issuecomment-624266589

mcoker:
@mcarrano to circle back on the related issue another user reported, https://github.com/patternfly/patternfly/issues/2997, I put together a demo of what their use case will look like with this update - https://codepen.io/mcoker/pen/xxwpeEa

Just want to verify that this fix will resolve #2997 as well?

mcarrano:
I think it does @mcoker . You can add a comment to #2997 proposing this as our fix before closing if you want. But IMO this makes it much more clear what is going on with the Search button.

Issue #3021, opened 2020-04-24
Updates to bottom pagination on mobile


mcoker:
* Make it sticky, provide an opt-out for that
* Update UI to something like this

![Apr-24-2020 17-56-54](https://user-images.githubusercontent.com/35148959/80263070-1459e800-8655-11ea-9dea-366ddbc8fe81.gif)

PR #3050, opened 2020-04-29
fix(pagination): updated bottom pagination design

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3021

This is the custom, sticky bottom pagination design on screens < the md breakpoint. You can disable the sticky behavior (but you still get the new design) by adding `.pf-m-no-sticky`.

I added the bottom pagination to the table and data-list demos, including a non-sticky bottom pagination at the end of those demo pages.

@mcarrano feel free to add more reviewers if you'd like anyone else to look at it

## Breaking changes

This PR makes a number of changes to the pagination component.

* The bottom pagination mobile design has changed. On mobile viewports, the bottom pagination now spans the width of the parent container and is "sticky" to the bottom of the viewport and will remain at the bottom of the viewport as content above it scrolls. This behavior can be disabled by applying `.pf-m-static` to the `.pf-c-pagination` component.
  * Since the pagination now spans the width of it's parent container on mobile, it should not be used in a toolbar, and the bottom pagination should be placed directly after the element the pagination is for (data-list, table, etc).
  * On desktop screens, the pagination retains its original design, which matches the top pagination.
* Renames the bottom pagination from `.pf-m-footer` to `.pf-m-bottom` to align with React. All use of `.pf-c-pagination.pf-m-footer` should change to `.pf-c-pagination.pf-m-bottom`.
  * All pagination component CSS variables with `m-footer` in the string should change to `m-bottom`. 
* The buttons used in `.pf-c-pagination__nav` are now individually wrapped with `div.pf-c-pagination__nav-control`, and this new element has modifiers `.pf-m-first`, `.pf-m-prev`, `.pf-m-next`, and `.pf-m-last` to indicate the first, previous, next, and last buttons. The "first" button should now be wrapped in `<div class="pf-c-pagination__nav-control pf-m-first">// first button</div>`, the "next" wrapped in `<div class="pf-c-pagination__nav-control pf-m-next">// next button</div>`, and so on.
* The following variables were renamed and the old variable name should be updated to reflect the new name
  * `--pf-c-pagination__nav--c-button--PaddingLeft` changed to `--pf-c-pagination__nav-control--c-button--PaddingLeft`
  * `--pf-c-pagination__nav--c-button--PaddingRight` changed to `--pf-c-pagination__nav-control--c-button--PaddingRight`
  * `--pf-c-pagination__nav--c-button--FontSize` changed to `--pf-c-pagination__nav-control--c-button--FontSize`
  * `--pf-c-pagination--m-compact__nav--c-button--MarginLeft` changed to `--pf-c-pagination--m-compact__nav-control--nav-control--MarginLeft`


christiemolloy:
I'm a little confused by the examples, e.g. in the table a lot of the examples don't have the sticky footer, but then theres an example to remove the sticky footer. Did you talk with design about what should be the default? should sticky be the default? It might make sense to have not sticky the default and then they add the sticky pagination if they want it. I also don't know about the name for the modifier as `pf-m-no-sticky` , maybe we could do `pf-m-relative`

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -10,11 +10,28 @@
   --pf-c-pagination__menu-text--Color: var(--pf-global--Color--200);
 
   // Nav
-  --pf-c-pagination__nav--c-button--PaddingLeft: var(--pf-global--spacer--sm);
+  --pf-c-pagination__nav--c-button--PaddingTop: 0;
   --pf-c-pagination__nav--c-button--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-pagination__nav--c-button--PaddingBottom: 0;
+  --pf-c-pagination__nav--c-button--PaddingLeft: var(--pf-global--spacer--sm);
   --pf-c-pagination__nav--c-button--FontSize: var(--pf-global--FontSize--md);
+  --pf-c-pagination__nav--m-footer--c-button--PaddingTop: var(--pf-global--spacer--md);
```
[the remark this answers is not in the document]
mcoker: Nope! Updated the var names.

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -48,14 +82,71 @@
...
+    border-top: var(--pf-c-pagination--m-footer--BorderTopWidth) solid var(--pf-c-pagination--m-footer--BorderTopColor);
+
+    &.pf-m-no-sticky {
+      --pf-c-pagination--m-footer--MarginTop: 0;
+      --pf-c-pagination--m-footer--BorderTopWidth: 0;
+
+      position: relative;
+    }
+
+    .pf-c-button {
```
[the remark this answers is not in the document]
mcoker: It wouldn't apply styles since it's just var declarations, they're just scoped to `.pf-c-button`, but they really don't need to be. I scoped them to `.pf-m-footer` instead.

mcoker:
> I'm a little confused by the examples, e.g. in the table a lot of the examples don't have the sticky footer, but then theres an example to remove the sticky footer. 

All of the demos should have a bottom pagination except for the empty state and loading demos. They are only sticky < 768px. 

> Did you talk with design about what should be the default? should sticky be the default? It might make sense to have not sticky the default and then they add the sticky pagination if they want it.

Yep we talked about that in the meeting with @mcarrano and I asked @mceledonia to confirm and he agrees.

> I also don't know about the name for the modifier as pf-m-no-sticky , maybe we could do pf-m-relative

Good call. How about `.pf-m-static`?

mcoker:
@christiemolloy @mattnolting updated!

@mcarrano did you want to increase the padding on the buttons so the touch area is larger? 

@mceledonia @lboehling updated the bottom pagination to have a top shadow instead of border when it's sticky.

mcarrano:
This is looking good @mcoker . Yes, if you are able to make the touch area larger on the left/right that would be helpful, I think. But I don't think this is a significant problem if it causes any other issues.

mattnolting:
> Good call. How about .pf-m-static?

I like that ^^

mcoker:
@mcarrano added the extra spacing. It's 16px all around and 24px on the side with larger padding. Let me know if you'd like me to change that.

@christiemolloy @mattnolting ready for another review, and updated the release notes.

christiemolloy:
Should the options menu have the padding-left ? 
Screen Shot 2020-05-04 at 9 52 23 AM

christiemolloy:
Should you add an outline offset since the focus ring will be cut off in every UI
Screen Shot 2020-05-04 at 10 17 04 AM

mcoker:
@christiemolloy good call, added `outline-offset` for the focus ring.

The toggle padding comes with that variation of the options menu so I think that's fine.

mcoker:
@christiemolloy there was an unnecessary margin-right on the options-menu though, just removed that.

christiemolloy:
is there a reason why the padding left is larger than padding right?
Screen Shot 2020-05-05 at 9 20 50 AM

mcoker:
@christiemolloy re: the padding - https://github.com/patternfly/patternfly/pull/3050#issuecomment-622418270

christiemolloy:
@mcarrano Now that @mcoker added the outline offset, i think the icon in the button looks weird off-center. Could we add padding evenly to the left and right side ?

Screen Shot 2020-05-05 at 9 20 44 AM

mceledonia:
+1 to that @christiemolloy !

mcoker:
@mceledonia @christiemolloy if we do that, it will no longer align with the edge of the content

Screen Shot 2020-05-05 at 9 58 29 AM

mcoker:
@mceledonia @christiemolloy updated the buttons so they have 16px on both sides on mobile.

src/patternfly/demos/DataList/data-list-pagination-footer-static.hbs
```diff
@@ -0,0 +1,4 @@
+{{#> pagination pagination--modifier="pf-m-bottom pf-m-static"}}
+  {{> pagination-options-menu id="{{page--id}}pagination-options-menu-bottom-example-static" options-menu--IsText="true"}}
+  {{> pagination-nav-content}}
+{{/pagination}}
```
mattnolting, line 4: 👍

src/patternfly/demos/MasterDetail/master-detail-table.hbs
```diff
@@ -273,30 +273,23 @@
...
+      <span aria-hidden="true">of 4</span>
+    {{/pagination-nav-page-select}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Go to next page"'}}
+      <i class="fas fa-angle-right" aria-hidden="true"></i>
+    {{/button}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Go to last page"'}}
+      <i class="fas fa-angle-double-right" aria-hidden="true"></i>
+    {{/button}}
+  {{/pagination-nav}}
+{{/pagination}}
```
mattnolting, line 295: awesome!

src/patternfly/demos/Table/examples/Table.md
```diff
@@ -193,6 +198,34 @@ section: demos
 {{/page}}
 ```
 
+```hbs title=Static-bottom-pagination isFullscreen
+{{#> page page--id="page-layout-table-simple"}}
```
mattnolting, line 202:
```suggestion
{{#> page page--id="static-bottom-pagination"}}
```

PR #3051, opened 2020-04-29
fix(tabs-box-border-top): fixed box vertical border top

mattnolting:
Fixes `.pf-c-tabs.pf-m-box.pf-m-vertical .pf-c-tabs__item:first-child.pf-m-current` border top
Screen Shot 2020-04-29 at 7 34 17 PM
and `.pf-c-tabs.pf-m-vertical` right border position
Screen Shot 2020-04-29 at 7 34 21 PM

### DIFF
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -179,16 +179,21 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
     --pf-c-tabs__link--PaddingTop: var(--pf-c-tabs--m-vertical__link--PaddingTop);
     --pf-c-tabs__link--PaddingBottom: var(--pf-c-tabs--m-vertical__link--PaddingBottom);
 
-    max-width: var(--pf-c-tabs--m-vertical--MaxWidth);
+    display: inline-flex;
+    flex-direction: column;
+
+    // If not a flex child, set height
+    height: 100%;
 
     // Because vertical variant has no scroll buttons, reset padding
     padding: 0;
 
     .pf-c-tabs__list {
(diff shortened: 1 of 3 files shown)


mcoker:
> .pf-c-tabs.pf-m-vertical right border position

What would cause that bug?

mattnolting:
> > .pf-c-tabs.pf-m-vertical right border position
> 
> What would cause that bug?

@mcoker `max-width` on `.pf-c-tabs__list` rather than `.pf-c-tabs`

mattnolting:
@mcoker https://patternfly-react-pr-4146.surge.sh/documentation/react/components/tabs#vertical enable `isBox`

mcoker:
Gotcha, thanks. Do we need these workspace styles? They don't seem necessary and were keeping that from showing up in the core examples. https://github.com/patternfly/patternfly/blob/a8f457be7c44775376854c1e12a22d4148bb0a6b/src/patternfly/components/Tabs/examples/Tabs.css#L1-L5

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -231,6 +232,11 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
       --pf-c-tabs__link--before--BorderBottomWidth: var(--pf-c-tabs__link--before--BorderWidth);
     }
```
mcoker, line 234: it's fine the way it is, but looks like you could use this here

```
.pf-c-tabs__item {
  &:first-child{}
  &:last-child {}
  &.pf-m-current{}
}
```

mattnolting: 👍 https://github.com/patternfly/patternfly/issues/3052

Issue #3063, opened 2020-05-05
Data-list - fix selected row shadow


mcoker:
Selected rows should have a top and bottom shadow
<img width="302" alt="Screen Shot 2020-05-04 at 8 57 05 PM" src="https://user-images.githubusercontent.com/35148959/81029271-e42afa00-8e49-11ea-832a-dd69b1150c53.png">

This is what the staging branch looks like
<img width="371" alt="Screen Shot 2020-05-04 at 8 56 57 PM" src="https://user-images.githubusercontent.com/35148959/81029272-e4c39080-8e49-11ea-917f-e432f85053e9.png">

PR #3068, opened 2020-05-05
fix(datalist): fix shadows on selected rows

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3063

The way shadows should work is the selected row has a top/bottom shadow by default, and if you hover/focus over a selectable row, that row should also have a top/bottom shadow, and if you hover/focus over a selectable row that is adjacent to a selected row, the hover shadow should take precedence over the selected row's shadow.

## Breaking changes

TBD

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -11,37 +11,37 @@
 
   // Item
   --pf-c-data-list__item--BackgroundColor: var(--pf-global--BackgroundColor--100);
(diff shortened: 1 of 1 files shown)


christiemolloy:
On mobile, when the last item in the data list is hovered over, the grey border at the bottom also "raised up" which gives it a different effect then what hovering over any of the other data list items does. 

Screen Shot 2020-05-06 at 10 29 05 AM 

Screen Shot 2020-05-06 at 10 28 48 AM

christiemolloy:
Release notes look great!!

Issue #2985, opened 2020-04-20
Select toggle height when chip group is present


mcoker:
With the new chip design, the select toggle height grows when chips are present. Let's review with design and see if it's possible to adjust this so the toggle height remains static.

PR #3070, opened 2020-05-06
fix(select): adjusted chip group margin in toggle

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2985


christiemolloy:
I think that due to the border in the select toggle, the spacing underneath the chips is less than the spacing on the top

Screen Shot 2020-05-06 at 10 15 29 AM

mcoker:
@christiemolloy yeah, that's a byproduct of the underline changing width when it's expanded. It's even when the select is closed.

Screen Shot 2020-05-06 at 9 28 20 AM

PR #3074, opened 2020-05-07
fix(table-vars): updated min-width vars, removed unused

mattnolting:
Found an inconsistency in truncation string length because `--pf-c-table__text` and `--pf-c-table__sort--MinWidth` don't factor in changing padding left/right values. 

Also, @redallen found an unused variable: `--pf-c-table--cell--Width: var(--pf-c-table--thead--cell--Width);`

## Breaking changes
* `--pf-c-table--thead--cell--Width`

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -60,7 +60,7 @@
 
   // Truncate
   --pf-c-table--m-truncate--cell--MaxWidth: 1px;
-  --pf-c-table--m-truncate--cell--MinWidth: 9ch;
+  --pf-c-table--m-truncate--cell--MinWidth: calc(5ch + var(--pf-c-table--cell--PaddingRight) + var(--pf-c-table--cell--PaddingLeft));
 
   // Hidden visible
   --pf-c-table--cell--hidden-visible--Display: table-cell;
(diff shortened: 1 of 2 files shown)


christiemolloy:
Looks like the popover is in the wrong spot.

Screen Shot 2020-05-07 at 9 24 53 AM

christiemolloy:
Also, im looking at the example. So is `pf-m-truncate` set by default on the table header?
I wonder if it should be set to `pf-m-nowrap` by default and then the user can add truncate if they would like, because it took me a while to figure out how to unset it. but if truncate is the default then thats ok was just wondering. although it looks like table headers are truncated in a lot of places where there is a short amount of text which would make for a lot of tooltips, and i dont know if thats the most usable experience

Screen Shot 2020-05-07 at 9 29 58 AM

src/patternfly/components/Table/table.scss
```diff
@@ -60,7 +60,7 @@
 
   // Truncate
   --pf-c-table--m-truncate--cell--MaxWidth: 1px;
-  --pf-c-table--m-truncate--cell--MinWidth: 9ch;
+  --pf-c-table--m-truncate--cell--MinWidth: calc(5ch + var(--pf-c-table--cell--PaddingRight) + var(--pf-c-table--cell--PaddingLeft));
```
mcoker, line 63: I wonder if we should have a var for `5ch` that serves as the theme var for that width, so that's what a user changes to set this. Then we use it in this `MinWidth` var and add the padding to it.

mcoker: I did something similar in pagination to create a `--whatever--chars` var that is the number of characters you want to show in the pagination text input, and the react component takes a prop for how many chars to show that modifies that var's value.

https://github.com/patternfly/patternfly/blob/ceb30f2f12240f73063013a9edca7c9b002642d5/src/patternfly/components/Pagination/pagination.scss#L32

PR #3096, opened 2020-05-12
fix(table-sort-th-columns): fixing col def bug for nowrap

mattnolting:
fixes `pf-m-nowrap` behavior on sortable `th`

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -480,11 +480,14 @@
   // stylelint-disable
   .pf-c-table tr.pf-m-nowrap &,
   .pf-c-table thead.pf-m-nowrap &,
-  .pf-c-table th.pf-m-nowrap &,
+  .pf-c-table th.pf-m-nowrap & {
+    grid-template-columns: min-content max-content;
+  }
+
   .pf-c-table tr.pf-m-fit-content &,
   .pf-c-table thead.pf-m-fit-content &,
   .pf-c-table th.pf-m-fit-content & {
-    grid-template-columns: min-content max-content;
+    grid-template-columns: fit-content max-content;
   }
 
   .pf-c-table tr.pf-m-wrap &,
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -473,6 +473,14 @@
   justify-content: start;
   grid-template-columns: auto;
   grid-auto-columns: max-content;
+
+  .pf-m-nowrap & {
```
mcoker, line 477: can you scope these modifiers to the elements they apply to? If there were something outside of the table with `.pf-m-nowrap/truncate`, that would enable this CSS.

src/patternfly/components/Table/table.scss
```diff
@@ -473,6 +473,14 @@
...
+    grid-template-columns: min-content auto;
+  }
+
+  .pf-m-truncate & {
+    grid-template-columns: auto;
```
mcoker, line 482: is this necessary since it's the default?

src/patternfly/components/Table/table.scss
```diff
@@ -471,8 +471,27 @@
...
+  .pf-c-table tr.pf-m-nowrap &,
+  .pf-c-table thead.pf-m-nowrap &,
+  .pf-c-table th.pf-m-nowrap &,
+  .pf-c-table tr.pf-m-wrap &,
+  .pf-c-table thead.pf-m-wrap &,
+  .pf-c-table th.pf-m-wrap &,
+  .pf-c-table tr.pf-m-fit-content &,
+  .pf-c-table thead.pf-m-fit-content &,
+  .pf-c-table th.pf-m-fit-content & {
+    grid-template-columns: min-content min-content;
```
mcoker, line 486: min-content will try to take up as little space as possible, so you end up with headers like this if you use `pf-m-wrap`

Screen Shot 2020-05-13 at 4 38 05 PM

Issue #3097, opened 2020-05-12
Add specific shield styles to global CSS


mcoker:
We set the following styles in the shield CSS that patternfly and users depended on that haven't caused any issues that I'm aware of.

```
  [class*="pf-c-"],
  [class*="pf-l-"] {
    &,
    &::before,
    &::after {
      padding: 0;
      margin: 0;
      background-color: transparent;
    }
  }
```

Removing them by default could cause some inadvertent issues with elements suddenly having an unexpected background/margin/padding that didn't before. These seem OK to add to the global styles.

PR #3098, opened 2020-05-12
fix(global): added global transparent, margin, and padding styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3097


mcoker:
@christiemolloy @mattnolting updated from our conversation earlier

Issue #3102, opened 2020-05-12
Update chip docs


mcoker:
Chips are no longer used as an `li`

PR #3103, opened 2020-05-12
fix(chip): updated docs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3102

### DIFF
--- a/src/patternfly/components/Chip/examples/Chip.md
+++ b/src/patternfly/components/Chip/examples/Chip.md
@@ -56,7 +56,7 @@ cssPrefix: pf-c-chip
 
 ## Documentation
 ### Overview
-A Chip is used to display items that have been filtered or selected from a larger group. They comprise of a text element and a button component that is used to remove the chip from selection. When the text overflows it is truncated using ellipses. A chip can be grouped by using the "chip-group" layout. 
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Chip/examples/Chip.md
```diff
@@ -56,7 +56,7 @@ cssPrefix: pf-c-chip
 
 ## Documentation
 ### Overview
-A Chip is used to display items that have been filtered or selected from a larger group. They comprise of a text element and a button component that is used to remove the chip from selection. When the text overflows it is truncated using ellipses. A chip can be grouped by using the "chip-group" layout. 
+A Chip is used to display items that have been filtered or selected from a larger group. They comprise of a text element and a button component that is used to remove the chip from selection. When the text overflows it is truncated using ellipses. A chip can be grouped by using the "chip-group" layout.
```
mattnolting, line 59:
```suggestion
A Chip is used to display items that have been filtered or selected from a larger group. They comprise of a text element and a button component that is used to remove the chip from selection. When the text overflows it is truncated using ellipses. A chip can be grouped by using the "chip-group" component.
```

mcoker:
@mattnolting updated

PR #3104, opened 2020-05-13
fix(form-control-background): split declarations

mattnolting:
fixes search input background


src/patternfly/components/FormControl/form-control.scss
```diff
@@ -81,7 +81,6 @@ $pf-c-form-control--m-search--Coordinates: "M505 442.7L405.3 343c-4.5-4.5-10.6-7
   --pf-c-form-control--m-search--BackgroundPosition: var(--pf-c-form-control--PaddingLeft);
```
mcoker, line 81: What you have here works. I wonder if this background position should be setup like the others where the non-icon side horizontal padding is used to position the icon. This would also allow you to redefine `--pf-c-form-control--PaddingLeft` later in `&.pf-m-search` instead of having to declare `padding-left`

```suggestion
  --pf-c-form-control--m-search--BackgroundPosition: var(--pf-c-form-control--PaddingRight);
```

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -179,7 +178,10 @@ $pf-c-form-control--m-search--Coordinates: "M505 442.7L405.3 343c-4.5-4.5-10.6-7
 
   &.pf-m-search {
     padding-left: var(--pf-c-form-control--m-search--PaddingLeft);
```
mcoker, line 180: Making the change above to `--pf-c-form-control--m-search--BackgroundPosition` would allow you to use `--pf-c-form-control--PaddingLeft` here

```suggestion
    --pf-c-form-control--PaddingLeft: var(--pf-c-form-control--m-search--PaddingLeft);
```

mattnolting:
@mcoker updated

Issue #3114, opened 2020-05-14
Update page header demos


mcoker:
* update the help icon to use `pf-icon-help`
* start by ensuring the icons are arranged similar to what is on master, and design can review once the PR is up if we need to re-arrange things.

PR #3115, opened 2020-05-14
fix(page): updated header tools icons in demos

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3114

### DIFF
--- a/src/patternfly/demos/Page/page-template-header-tools-elements.hbs
+++ b/src/patternfly/demos/Page/page-template-header-tools-elements.hbs
@@ -1,36 +1,19 @@
 {{#> page-header-tools}}
-  {{#> page-header-tools-group page-header-tools-group--modifier="pf-m-hidden pf-m-visible-on-lg"}}
-    {{#> page-header-tools-item}}
+  {{#> page-header-tools-group}}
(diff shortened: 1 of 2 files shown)


src/patternfly/demos/Page/page-template-header-tools-elements.hbs
```diff
@@ -1,36 +1,19 @@
 {{#> page-header-tools}}
-  {{#> page-header-tools-group page-header-tools-group--modifier="pf-m-hidden pf-m-visible-on-lg"}}
-    {{#> page-header-tools-item}}
+  {{#> page-header-tools-group}}
+    {{> page-template-header-tools-notification-badge page-template-header-tools-notification-badge--item--modifier="pf-m-hidden pf-m-visible-on-lg"}}
```
christiemolloy, line 3: I don't think I see the notification bell?  https://patternfly-pr-3115.surge.sh/documentation/core/demos/page/default-nav

mcoker: Oh I removed the notification bell from the default demos, they're only on the drawer demos now.

Issue #3126, opened 2020-05-19
Select - add a focus class for menu items


mcoker:
The react select typeahead allows users to navigate the select menu items via keyboard while maintaining browser focus on the  text input in the toggle. This allows users to type in the text input while navigating the menu items.

When navigating via keyboard, since `:focus` remains on the toggle input, we need a class react can use to apply to the currently selected menu item via keyboard navigation.

PR #3127, opened 2020-05-19
fix(select): added item modifier to apply focus styles for keyboard nav

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3126


mcoker:
@christiemolloy done!

PR #3130, opened 2020-05-22
feat(dropdown, select): add description element

### LINKED ISSUE #3057: Dropdown menu with description
Want to add the ability to include descriptive text with any dropdown menu item. This variant should be available for any component that supports dropdown menu behavior including Dropdown, Select, and Options menu.

Design

image

cc @mcoker @LHinson

christiemolloy:
closes #3057


src/patternfly/components/AppLauncher/app-launcher-menu-item-description.hbs
```diff
@@ -0,0 +1,6 @@
+<p class="pf-c-app-launcher__menu-item-description{{#if app-launcher-menu-item-description--modifier}} {{app-launcher-menu-item-description--modifier}}{{/if}}"
+  {{#if app-launcher-menu-item-description--attribute}}
+	  {{{app-launcher-menu-item-description--attribute}}}
```
mcoker, line 3: The indentation is off

src/patternfly/components/AppLauncher/app-launcher-menu-item-description.hbs
```diff
@@ -0,0 +1,6 @@
+<p class="pf-c-app-launcher__menu-item-description{{#if app-launcher-menu-item-description--modifier}} {{app-launcher-menu-item-description--modifier}}{{/if}}"
```
mcoker, line 1: Do you know this should be a `<p>` for sure? Users wouldn't be able to use non-phrasing content here with a `<p>`, like a `<ul/ol/dl>`, `<div>`, etc.

mcoker, line 1: Since a `<button>` only accepts [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content), looks like this will need to be a `<span>`.

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -230,6 +234,10 @@
       --pf-c-app-launcher__menu-item--m-action--Color: var(--pf-c-app-launcher__menu-item--m-action--hover--Color);
     }
   }
+
+  &.pf-m-description {
+    display: block;
```
mcoker, line 239: Hmm... this is going to cause the styling between non-description and description items to be different. A good example is an item with favorites.

I wonder if this might be a use case for absolutely positioning a pseudo element of `__menu-item` relative to the parent `li` so it covers the entire `li`. Then placing the `__menu-item-description` element after `__menu-item` element. So technically they're different elements, but the pseudo element of the `__menu-item` covers the `__menu-item-description` element, so clicking on anywhere in the `li` triggers a click to the `__menu-item`.

christiemolloy: Tried this and it works really well from what i can see!

src/patternfly/components/AppLauncher/app-launcher.scss
```diff
@@ -246,6 +254,11 @@
   }
 }
 
+.pf-c-app-launcher__menu-item-description {
```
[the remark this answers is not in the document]
christiemolloy: fixed now

src/patternfly/components/Dropdown/dropdown-menu-item-description.hbs
```diff
@@ -0,0 +1,6 @@
+<span class="pf-c-dropdown__menu-item-description{{#if dropdown-menu-item-description--modifier}} {{dropdown-menu-item-description--modifier}}{{/if}}"
```
mcoker, line 1: use a `<div>`

src/patternfly/components/Dropdown/dropdown-menu-item-title.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-dropdown__menu-item-title{{#if dropdown-menu-item-title--modifier}} {{dropdown-menu-item-title--modifier}}{{/if}}"
```
mcoker, line 1: I might just call this `menu-item-text` instead of `menu-item-title`

mcoker: Oh sorry I read the code before looking at the examples and assumed it just wrapped the text. Maybe `-main`?

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -416,6 +420,16 @@
   }
 }
 
+.pf-c-dropdown__menu-item-description {
+  font-size: var(--pf-c-dropdown__menu-item-description--FontSize);
+  color: var(--pf-c-dropdown__menu-item-description--Color);
+}
+
+.pf-c-dropdown__menu-item-title {
+  display: flex;
```
mcoker, line 429: Hmm... this is going to make text in the item in a regular dropdown (no description) and an item with a description display differently. I think this flex styling needs to be scoped to when `.pf-m-icon` is applied to the top level item. Though I see `.pf-m-icon` is missing from the item in these description examples. Is that intentional?

One way is we could move `.pf-m-icon` from `.pf-c-dropdown__menu-item` to `.pf-c-dropdown__menu-item-title` when a description is used, and update `.pf-c-dropdown__menu-item.pf-m-icon { display: flex; align-items: center;}` to include `.pf-c-dropdown__menu-item-title.pf-m-icon`.

You could also do something like this which might be easier on the react side:

```
li // has icon
  a.pf-c-dropdown__menu-item.pf-m-description.pf-m-icon // leaves .pf-m-icon on the top level menu-item with or without a description
    div.pf-c-dropdown__menu-item-title // or whatever you end up calling it
      span.pf-c-dropdown__menu-item-icon
        // some icon
      the dropdown item text
    div.pf-c-dropdown__menu-item-description
      the dropdown item description

li // no icon
  a.pf-c-dropdown__menu-item.pf-m-description
    div.pf-c-dropdown__menu-item-title // or whatever you end up calling it
      the dropdown item text
    div.pf-c-dropdown__menu-item-description
      the dropdown item description
```

And then the CSS

```
.pf-c-dropdown__menu-item {
  &.pf-m-description {
    flex-direction: column; // fixes the direction when `.pf-m-icon` is applied
       
    &.pf-m-icon {
      .pf-c-dropdown__menu-item-main {
        display: flex;
        align-items: center;
      }
    }
}
```

src/patternfly/components/Select/select.scss
```diff
@@ -394,9 +410,25 @@
...
   }
 }
 
+.pf-c-select__menu-item-title {
+  position: relative;
+  padding-right: var(--pf-c-select__menu-item-title--PaddingRight);
+}
+
+.pf-c-select__menu-item-description {
+  display: block;
```
mcoker, line 426: is this necessary?

src/patternfly/components/Select/select.scss
```diff
@@ -346,6 +354,14 @@
     right: 0;
   }
 
+  &.pf-m-wrap {
```
mcoker, line 357: Not sure if this is the intention but is `.pf-m-wrap` added to the parent `pf-c-select__menu` element to allow its children items to wrap? Which allows the description to wrap under the menu-item-title element? If so wouldn't that mean that if `.pf-m-wrap` is used and some of the items have descriptions and others don't, all of the items will now wrap?

Also, that would require the react component to go back up to the parent `__menu` to add a class when one of its children meets some condition, which isn't ideal.

What about just adding `.pf-m-description` to all items with descriptions (not just ones that are selected, as it is currently), and we add `white-space: normal;` to them, then add `.white-space: nowrap;` to the `__menu-item-title` element?

Otherwise this lgtm!

src/patternfly/components/Dropdown/dropdown-menu-item-title.hbs
christiemolloy: yeah that sounds good to me

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -416,6 +433,16 @@
...
 
+.pf-c-dropdown__menu-item-description {
+  font-size: var(--pf-c-dropdown__menu-item-description--FontSize);
+  color: var(--pf-c-dropdown__menu-item-description--Color);
+}
+
+.pf-c-dropdown__menu-item-main {
+  display: flex;
+  align-items: center;
+}
```
mcoker, line 444: looks like this is still here and doesn't need to be since you added it when it's a child of `.pf-c-dropdown__menu-item.pf-m-icon`

src/patternfly/components/Select/select.scss
```diff
@@ -123,6 +123,14 @@
   --pf-c-select__menu-group-title--FontWeight: var(--pf-global--FontWeight--semi-bold);
   --pf-c-select__menu-group-title--Color: var(--pf-global--Color--dark-200);
 
+  // Menu item description
+  --pf-c-select__menu-item-description--FontSize: var(--pf-global--FontSize--xs);
+  --pf-c-select__menu-item-description--Color: var(--pf-global--Color--200);
+  --pf-c-select__menu-item-description--PaddingRight: var(--pf-c-select__menu-item--PaddingRight);
+
+  // Menu item title
+  --pf-c-select__menu-item-main--PaddingRight: var(--pf-global--spacer--2xl);
```
mcoker, line 132: Should this be...

```suggestion
  --pf-c-select__menu-item-main--PaddingRight: var(--pf-c-select__menu-item--m-selected--PaddingRight);
```

src/patternfly/components/Select/select.scss
```diff
@@ -395,6 +403,24 @@
   &.pf-m-selected {
     --pf-c-select__menu-item--PaddingRight: var(--pf-c-select__menu-item--m-selected--PaddingRight);
   }
+
+  &.pf-m-description {
+    --pf-c-select__menu-item--m-selected--PaddingRight: 0;
```
mcoker, line 408: and if you made the change above, this would change to 

```suggestion
    --pf-c-select__menu-item--PaddingRight: 0;
```

PR #3132, opened 2020-05-26
feat(button): add small variation

### LINKED ISSUE #3116: Updating styling in alerts + Add modifier to the Button
@mceledonia bought up that the font-size in the alerts should be 14px instead of 16px. Also the inline button should be 14px. We decided that we should introduce a modifier to the button component for 14px font-size.

christiemolloy:
closes #3116 

- adds variation to the button for 14px font-size
- updates the alert to use the 14px button, and updates font-size to be 14px. At 14px, the icon wasn't aligning with the text, so I removed the margin-top, used flex, and update the grid-template-areas, i think this is a more flexible solution (hopefully it isn't a breaking change or we could see it as a bug fix)

Screen Shot 2020-05-26 at 4 53 17 PM
Screen Shot 2020-05-26 at 4 53 22 PM

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -20,6 +20,7 @@
   // Title
   --pf-c-alert__title--FontWeight: var(--pf-global--FontWeight--bold);
   --pf-c-alert__title--Color: var(--pf-global--default-color--300);
+  --pf-c-alert__title--max-lines: 1;
 
   // Action
   --pf-c-alert__action--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
@@ -83,8 +84,8 @@
   grid-template-columns: var(--pf-c-alert--GridTemplateColumns);
   grid-template-areas:
     "icon title action"
-    "icon description description"
-    "icon actiongroup actiongroup";
+    ". description description"
+    ". actiongroup actiongroup";
 
   &.pf-m-success {
     --pf-c-alert--BorderTopColor: var(--pf-c-alert--m-success--BorderTopColor);
@@ -135,6 +136,11 @@
   grid-area: title;
   font-weight: var(--pf-c-alert__title--FontWeight);
   color: var(--pf-c-alert__title--Color);
+  word-break: break-word;
+
+  &.pf-m-truncate {
+    @include pf-line-clamp("var(--pf-c-alert__title--max-lines)");
+  }
 }
 
 .pf-c-alert__description {
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -28,7 +28,7 @@
   --pf-c-alert__action--MarginRight: calc(var(--pf-global--spacer--sm) * -1);
 
   // Action Description
-  --pf-c-alert__description--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-alert__description--PaddingTop: var(--pf-global--spacer--xs);
 
   // Alert action group
   --pf-c-alert__action-group--PaddingTop: var(--pf-global--spacer--xs);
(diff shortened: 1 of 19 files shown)


src/patternfly/components/Alert/alert.scss
```diff
@@ -120,15 +122,23 @@
 .pf-c-alert__icon {
   grid-area: icon;
   display: flex;
+  align-items: center;
```
mcoker, line 125: This will center the icon vertically when the title text wraps to more than 1 line.

I tried out removing the `line-height` added to `__icon > *` and `__title`, and adding back `--pf-c-alert__icon--MarginTop: 1px` and that seems to work, except the close button adds extra height to the top row (both in this PR and with the updates I mentioned). Adding `line-height: 1;` to the close button looks to address that.

Also now I wonder if we should add `line-height: 1` to the button component's base styles, and use the `spacer--sm` for the top/bottom padding instead of the custom `6px` form element var.

mcoker: Here are those changes in a PR - https://github.com/christiemolloy/patternfly-next/pull/27

christiemolloy: Good point I didnt think about the case where the text wraps to two lines. Merged your PR in that looks great

christiemolloy: Good point I didnt think about the case where the text wraps to two lines. Merged your PR in that looks great going to check with design about the new font-size

christiemolloy:
@mceledonia @mcarrano This is the alert with the new font-size of 14px, do you think there should be the same spacer between the title text and the description text? i think the space looks a little big now but let me know what you think?

Screen Shot 2020-06-01 at 1 58 13 PM

src/patternfly/components/Button/examples/Button.md
```diff
@@ -94,6 +130,23 @@ import './Button.css'
 {{#> button button--modifier="pf-m-control" button--attribute="disabled"}}
   Control disabled
 {{/button}}
+<br>
```
mcoker, line 133: did you mean to add this section to the disabled buttons?

mceledonia:
@christiemolloy The spacing does look a bit big, here's a spec to double check it with - 
image
image

Is there any additional spacing getting added somehow?

mceledonia:
@christiemolloy I think the spacing is correct but the text seems to be too small, check this comparison I just made with your screenshot. Is the text 14px? 
image

mceledonia:
Here's what it looks like when I line up the spacing. I can't tell what's causing the inconsistencies here, maybe worth jumping on a call at some point. 

I lined up the spacing/icon as a reference point, the transparent overlay is the designed version and the opaque layer is the screenshot. 

image

src/patternfly/components/Button/examples/Button.md
christiemolloy: yes i was assuming the small text would apply to all of the buttons with text. @mceledonia what do you think?

christiemolloy:
ahhh looks like there was 8px between the title and description, will update to 4px.

> Edit - Actually here's what it looks like if I line up the text. So maybe the text size is correct but the spacing is off?
> 
> image
> 
> Adding spec here for reference -
> Orange = 4px
> Green = 16px
> Cyan = 24px
> image

yep looks

src/patternfly/components/Button/examples/Button.md
mcoker: They're in the disabled section, but aren't disabled. You already have examples of them in the "variations" example to show what they look like. I don't know that they need to be listed under disabled, even if we update them to be disabled.

src/patternfly/components/Alert/alert.scss
```diff
@@ -128,6 +131,8 @@
 }
 
 .pf-c-alert__title {
+  display: flex;
+  align-items: center;
```
mcoker, line 135: What do we need this for? Doesn't seem necessary? If someone has `<span>`s or other elements in the title, the whitespace between them will no longer be honored.

src/patternfly/components/Button/examples/Button.md
christiemolloy: Oh yep my bad don't know how they got in there

Issue #3144, opened 2020-05-29
Notification Drawer: Allow long strings in the title to wrap (long string no spaces)


jcaianirh:
The example for notification drawer shows a long string that wraps nicely, however, in openshift we have long strings that are multiple words without spacing we would also like to wrap. Otherwise, it causes horizontal scrolling to occur in the notification drawer. See attachments.

![Screen Shot 2020-05-29 at 5 16 14 PM](https://user-images.githubusercontent.com/35978579/83306849-55539800-a1d1-11ea-814b-816970842103.png)

![Screen Shot 2020-05-29 at 5 19 59 PM](https://user-images.githubusercontent.com/35978579/83306817-48cf3f80-a1d1-11ea-83fd-42bd7d978f45.png)

PR #3145, opened 2020-05-29
fix(notificationdrawer/alert): improved title wrapping

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3144

This PR enables long strings when used as title of a notification in the notification drawer to wrap:
Screen Shot 2020-05-29 at 4 50 23 PM

----

It also enables the same behavior in the alert component for consistency:
Screen Shot 2020-05-29 at 4 54 29 PM

----

And the way notification drawer group/section title/toggle wraps currently isn't ideal, so I updated it to truncate when the title text exceeds the width of the toggle (to be consistent with the accordion) - @mcarrano wdyt of this?
Screen Shot 2020-05-29 at 4 49 26 PM

----

Alternatively, we could allow that text to wrap - it would look like this.

Screen Shot 2020-05-29 at 4 48 23 PM

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -19,6 +19,7 @@
   // Title
   --pf-c-alert__title--FontWeight: var(--pf-global--FontWeight--bold);
   --pf-c-alert__title--Color: var(--pf-global--default-color--300);
+  --pf-c-alert__title--max-lines: 1;
 
   // Action
   --pf-c-alert__action--MarginTop: calc(var(--pf-global--spacer--form-element) * 2 * -1);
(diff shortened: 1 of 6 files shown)


mattnolting:
Looks great!

@mcoker wdyt about adding a `.pf-m-truncate` modifier for `.pf-c-alert__title` and `.pf-c-notification-drawer__list-item-header-title`? I could see a use case for devs wanting to override the default word-break. 

`.pf-c-notification-drawer__group-toggle-title`

mcoker:
I think that's a good idea. We could also do it via `line-clamp` and maybe have it as a react prop to specify the number of lines to show.

wdyt @mcarrano should we add support for a user to pass a prop/modifier to these title to truncate the text instead of having it wrap? Or maybe wait until products request that? I believe the only place we give the user the option to truncate text is the table.

mcarrano:
> wdyt @mcarrano should we add support for a user to pass a prop/modifier to these title to truncate the text instead of having it wrap? Or maybe wait until products request that? I believe the only place we give the user the option to truncate text is the table.

I'm good with that. Would provide more flexibility.

mcoker:
@mcarrano @mattnolting updated. I added the option to set a line clamp value and a couple of examples. I'm not sure that we need the examples, or if we could just call it out in the documentation? Let me know what you think!

jcaianirh:
@mcoker is this breaking on tests? If not, can we get this little piece of heaven merged?

Issue #3141, opened 2020-05-29
Add $pf-global--breakpoint--sm to tabs inset breakpoint modifiers


mcoker:
Looks like we're missing the small breakpoint here:

https://github.com/patternfly/patternfly/blob/545b48279a5c3df90599fae98bcdcc19ad74b518/src/patternfly/components/Tabs/tabs.scss#L1

Issue #3140, opened 2020-05-29
Add $pf-global--breakpoint--md to drawer panel breakpoint modifiers


mcoker:
Looks like there is some conflicting logic here that keeps the medium breakpoint from being available in the modifiers?

https://github.com/patternfly/patternfly/blob/545b48279a5c3df90599fae98bcdcc19ad74b518/src/patternfly/components/Drawer/drawer.scss#L298-L300

https://github.com/patternfly/patternfly/blob/545b48279a5c3df90599fae98bcdcc19ad74b518/src/patternfly/components/Drawer/drawer.scss#L315-L319

PR #3146, opened 2020-06-01
fix(drawer/tabs): add missing breakpoints to breakpoint mods

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3141
fixes https://github.com/patternfly/patternfly/issues/3140


mattnolting:
There's a bug in the responsive presentation. https://github.com/patternfly/patternfly/pull/3146/files#diff-56492175f881007db175cb81ee985735L300 was in place so that all drawers would have the same behavior until the `md` breakpoint, but `pf-m-inline` could still be applied without requiring a breakpoint value. Because `pf-m-inline` and `pf-m-static` disable the expandability of the drawer by resetting `margin` and `transform`, those properties would have to considered for mobile viewport control..

mattnolting:
Screen Shot 2020-06-01 at 12 40 46 PM

redallen:
Aw, shucks. Is there really much value to the `md` breakpoint on drawers anyway instead of just changing the default?

redallen:
Hmmm, FWIW this PR has the a11y tests crashing on /documentation/core/demos/masterdetail/master-detail-card-data-list-expanded-on-mobile. I'll put up a PR to revert the `md` change for drawer and we can always add it later.

PR #3157, opened 2020-06-05
fix(docs): add 2020.07 release notes

redallen:
Almost forgot release notes.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,6 +4,12 @@ section: overview
 releaseNoteTOC: true
 ---
 
+## 2020.07 release notes (2020-06-05)
+Packages released:
+- @patternfly/patternfly@v4.10.31
+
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.md
```diff
@@ -4,6 +4,12 @@ section: overview
 releaseNoteTOC: true
 ---
 
+## 2020.07 release notes (2020-06-05)
+Packages released:
+- @patternfly/patternfly@v4.10.31
+
+This is our major release. Check out our [upgrade guide](/documentation/core/overview/upgrade-guide) for a list of breaking changes!
```
mcoker, line 11: does this link need a `//patternfly.org/v4` prefix?

redallen: Nope! We have other links that work the same way: 
image

The prefix gets handled automagically in org.

PR #3159, opened 2020-06-05
feat(datalist): add text modifiers

### LINKED ISSUE #3087: DataListCell doesn't wrap words
The DataListCell component doesn't wrap words and should.  

PF issue filed here https://github.com/patternfly/patternfly-react/issues/3938

DataListCell with `word-break: break-word` styling
Screen Shot 2020-03-19 at 12 47 46 PM

DataListCell with no wrapping styling
Screen Shot 2020-03-19 at 1 17 42 PM

christiemolloy:
closes #3087

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -67,6 +67,11 @@
   --pf-c-data-list__cell--cell--PaddingTop: 0;
   --pf-c-data-list__cell--cell--md--PaddingTop: var(--pf-global--spacer--lg);
   --pf-c-data-list__cell--m-icon--cell--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-data-list__cell--MaxWidth: none;
+  --pf-c-data-list__cell--Overflow: visible;
+  --pf-c-data-list__cell--TextOverflow: clip;
+  --pf-c-data-list__cell--WhiteSpace: normal;
+  --pf-c-data-list__cell--WordBreak: normal;
   --pf-c-data-list__cell--m-truncate--MaxWidth: 100%;
 
   @media screen and (min-width: $pf-global--breakpoint--md) {
@@ -181,24 +186,24 @@
   &.pf-m-truncate {
     .pf-c-data-list__cell,
     .pf-c-data-list__text {
-      max-width: var(--pf-c-data-list__cell--m-truncate--MaxWidth);
-      overflow: hidden;
-      text-overflow: ellipsis;
-      white-space: nowrap;
+      --pf-c-data-list__cell--MaxWidth: var(--pf-c-data-list__cell--m-truncate--MaxWidth);
+      --pf-c-data-list__cell--Overflow: hidden;
+      --pf-c-data-list__cell--TextOverflow: ellipsis;
+      --pf-c-data-list__cell--WhiteSpace: nowrap;
     }
   }
 
   &.pf-m-break-word {
     .pf-c-data-list__cell,
     .pf-c-data-list__text {
-      word-break: break-word;
+      --pf-c-data-list__cell--WordBreak: break-word;
     }
   }
 
   &.pf-m-nowrap {
     .pf-c-data-list__cell,
     .pf-c-data-list__text {
-      white-space: nowrap;
+      --pf-c-data-list__cell--WhiteSpace: nowrap;
     }
   }
 }
@@ -281,24 +286,24 @@
   &.pf-m-truncate {
     .pf-c-data-list__cell,
     .pf-c-data-list__text {
-      max-width: var(--pf-c-data-list__cell--m-truncate--MaxWidth);
-      overflow: hidden;
-      text-overflow: ellipsis;
-      white-space: nowrap;
+      --pf-c-data-list__cell--MaxWidth: var(--pf-c-data-list__cell--m-truncate--MaxWidth);
+      --pf-c-data-list__cell--Overflow: hidden;
+      --pf-c-data-list__cell--TextOverflow: ellipsis;
+      --pf-c-data-list__cell--WhiteSpace: nowrap;
     }
   }
 
   &.pf-m-break-word {
     .pf-c-data-list__cell,
     .pf-c-data-list__text {
-      word-break: break-word;
+      --pf-c-data-list__cell--WordBreak: break-word;
     }
   }
 
   &.pf-m-nowrap {
     .pf-c-data-list__cell,
     .pf-c-data-list__text {
-      white-space: nowrap;
+      --pf-c-data-list__cell--WhiteSpace: nowrap;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


mcoker:
Can we extend these modifiers to apply to the `.pf-c-data-list` and `.pf-c-data-list__item-row` similar to the table?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -176,6 +182,30 @@
       font-size: var(--pf-c-data-list--m-compact__check--FontSize);
     }
   }
+
+  &.pf-m-truncate {
+    .pf-c-data-list__cell,
+    .pf-c-data-list__text {
```
mcoker, line 188: in a block like this, I don't think you need the `__cell` and `__text` selectors do you?

I also think you could write the sass like

```
.pf-c-data-list,
.pf-c-data-list__item-row,
.pf-c-data-list__cell,
.pf-c-data-list__text {
  &.pf-m-truncate {
    // truncate vars
  }
  .pf-m-break-word {
    // break-word vars
  }
  ...
}
```

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -658,6 +658,124 @@ When a list item includes more than one block of content, it can be difficult fo
...
+      {{/data-list-item-content}}
+    {{/data-list-item-row}}
+  {{/data-list-item}}
+{{/data-list}}
+```
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-data-list__text` | `<span>` | Inserts the data list text element. Use this class to modify specific text directly. |
```
mcoker, line 774: We should document that this applies to `*`, which will also let react know it's a flexible element that can be modified with a prop (`component`, I think is what's typically used) to use a different element. It's documented as `*` in a table, too.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -176,6 +182,30 @@
...
+      --pf-c-data-list__cell--MaxWidth: var(--pf-c-data-list__cell--m-truncate--MaxWidth);
+      --pf-c-data-list__cell--Overflow: hidden;
+      --pf-c-data-list__cell--TextOverflow: ellipsis;
+      --pf-c-data-list__cell--WhiteSpace: nowrap;
```
mcoker, line 192: What do you think about adding a reasonable min-width when something is truncated so the cell doesn't shrink until no text is visible? The table uses `--pf-c-table--m-truncate--cell--MinWidth: calc(5ch + var(--pf-c-table--cell--PaddingRight) + var(--pf-c-table--cell--PaddingLeft));`

src/patternfly/components/DataList/data-list.scss
```diff
@@ -176,6 +182,30 @@
...
+    }
+  }
+
+  &.pf-m-break-word {
+    .pf-c-data-list__cell,
+    .pf-c-data-list__text {
+      --pf-c-data-list__cell--WordBreak: break-word;
```
mcoker, line 199: We should probably model this off of the table and add `--pf-c-data-list__cell--WhiteSpace: normal;` so that `pf-m-break-word` can be added to a cell in a table that has `.pf-m-nowrap`, and it will enable that cell to wrap. We could also add a `.pf-m-wrap` modifier to do the same thing. Looks like that's missing in the data-list

src/patternfly/components/DataList/data-list.scss
christiemolloy: I added 20ch, because the data-list cells didn't have any paddingRight or paddingLeft to base the calculation off of, so i thought keeping it to characters would make the most sense right now but let me know if you have different thoughts

src/patternfly/components/DataList/data-list.scss
christiemolloy: Ah yes good point. I didn't add pf-m-wrap because the data list cells wrap by default. Should I still add it?

christiemolloy:
When the user adds the text element to their data list cell, do you expect the data list cell to control the text effect or do you expect the text element to control it. E.g. if i add `pf-m-flex-2` to one of the data list cells, the width of the cell wont be affected because the text element is what is controlling the width. Which i dont think we want @mcoker @mattnolting 

Screen Shot 2020-06-24 at 10 23 19 AM

also I can't work out how to use min-width here without causing the text effects to not work. it seemed like `initial` would work but it keeps defaulting to auto.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -67,6 +67,14 @@
   --pf-c-data-list__cell--cell--PaddingTop: 0;
   --pf-c-data-list__cell--cell--md--PaddingTop: var(--pf-global--spacer--lg);
   --pf-c-data-list__cell--m-icon--cell--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-data-list__cell--MaxWidth: none;
+  --pf-c-data-list__cell--MinWidth: 0;
```
mattnolting, line 71:
```suggestion
  --pf-c-data-list__cell--MinWidth: initial;
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -368,6 +404,15 @@
...
 }
 
+// Data list text
+.pf-c-data-list__text {
+  max-width: var(--pf-c-data-list__cell--MaxWidth);
+  overflow: var(--pf-c-data-list__cell--Overflow);
+  text-overflow: var(--pf-c-data-list__cell--TextOverflow);
+  word-break: var(--pf-c-data-list__cell--WordBreak);
+  white-space: var(--pf-c-data-list__cell--WhiteSpace);
+}
```
mattnolting, line 414: I'd probably pair `.pf-c-data-list__cell` and `.pf-c-data-list__text` and apply the shared properties.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -368,6 +404,15 @@
   &.pf-m-flex-5 { flex-grow: 5; }
 }
 
+// Data list text
+.pf-c-data-list__text {
```
mattnolting, line 408: Careful with applying this class to `span` s. The text mods won't affect inline items, see: https://patternfly-pr-3159.surge.sh/documentation/core/components/datalist#text-modifiers-data-list-text. You'll need to set a `display` property on `__text` or only apply to block level elements.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -67,6 +67,14 @@
...
   --pf-c-data-list__cell--cell--md--PaddingTop: var(--pf-global--spacer--lg);
   --pf-c-data-list__cell--m-icon--cell--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-data-list__cell--MaxWidth: none;
+  --pf-c-data-list__cell--MinWidth: 0;
+  --pf-c-data-list__cell--Overflow: visible;
+  --pf-c-data-list__cell--TextOverflow: clip;
+  --pf-c-data-list__cell--WhiteSpace: normal;
+  --pf-c-data-list__cell--WordBreak: normal;
+  --pf-c-data-list__cell--m-truncate--MaxWidth: 100%;
+  --pf-c-data-list__cell--m-truncate--MinWidth: calc(20ch);
```
mcoker, line 77: no need for calc(), and this should be something much smaller. it's 5ch in the table

https://github.com/patternfly/patternfly/blob/1a1a1a84c6e1efaa65b1b872c0aa306755b868fb/src/patternfly/components/Table/table.scss#L63

src/patternfly/components/DataList/data-list.scss
mcoker: It's `display: block;` in the table, but you might be able to make it `inline-block` without any adverse effects?

src/patternfly/components/DataList/data-list.scss
mcoker: I would leave this as `auto` since it's the default, just to keep the default behavior. Then update `--m-truncate--MinWidth` to `5ch`

Issue #3044, opened 2020-04-28
Support for table column header info popover


mcoker:
This is a request to add support for a help/info popover to table column headers.

* We can use a plain button with modified padding for the help/info button
* Sortable column header buttons should not consume the width of the `<th>` cell. This is a global change, and applies whether there is a popover in the header cell or not.
* Spacing updates from the screenshot below:
  * the right padding of the sortable text/icon button was modified to be `8px`
  * the help/info button has left/right padding of 8px
* The help/info button should be bottom-aligned with the header cell text if the text wraps, similar to the sort direction icon
* The icon is `pf-icon pf-icon-help`

<img width="1140" alt="Screen Shot 2020-04-28 at 2 55 12 PM" src="https://user-images.githubusercontent.com/35148959/80532285-825a2400-8961-11ea-9025-9b0752a4fda6.png">

cc @beanh66 @mcarrano @mceledonia @rebeccaalpert

beanh66:
Thanks @mcoker! We have a complex table use case where we need to elaborate on the difference between two columns specifically so this will really help us :)

PR #3189, opened 2020-06-17
feat(table): added support for column header help

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3044

### DIFF
--- a/src/patternfly/components/Table/table-column-help-action.hbs
+++ b/src/patternfly/components/Table/table-column-help-action.hbs
@@ -1,0 +1,8 @@
+<span class="pf-c-table__column-help-action{{#if table-column-help-action--modifier}} {{table-column-help-action--modifier}}{{/if}}"
+  {{#if table-column-help-action--attribute}}
+    {{{table-column-help-action--attribute}}}
+  {{/if}}>
+  {{#> button button--modifier="pf-m-plain" button--attribute="More info"}}
+    <i class="pficon pf-icon-help" aria-hidden="true"></i>
+  {{/button}}
+</span>
--- a/src/patternfly/components/Table/table-th.hbs
+++ b/src/patternfly/components/Table/table-th.hbs
@@ -29,6 +29,9 @@
   {{#if table-th--attribute}}
     {{{table-th--attribute}}}
   {{/if}}>
+  {{#if table-th--IsColumnHelp}}
+    <div class="pf-c-table__column-help">
+  {{/if}}
   {{#if table-th--sortable}}
     {{#> table-button}}
       {{#> table-button-content}}
@@ -58,6 +61,7 @@
     {{> @partial-block}}
   {{/if}}
   {{#if table-th--IsColumnHelp}}
-    {{> table-column-help}}
+    {{> table-column-help-action}}
+    </div>
   {{/if}}
 </th>
--- a/src/patternfly/components/Table/table-th.hbs
+++ b/src/patternfly/components/Table/table-th.hbs
@@ -57,4 +57,7 @@
   {{else}}
     {{> @partial-block}}
   {{/if}}
+  {{#if table-th--IsColumnHelp}}
+    {{> table-column-help}}
+  {{/if}}
 </th>
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -77,10 +77,6 @@
   --pf-c-table__toggle--c-button--m-expanded__toggle-icon--Rotate: 360deg;
 
   // Button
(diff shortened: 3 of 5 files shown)


src/patternfly/components/Table/table-column-help-action.hbs
```diff
@@ -0,0 +1,8 @@
+<span class="pf-c-table__column-help-action{{#if table-column-help-action--modifier}} {{table-column-help-action--modifier}}{{/if}}"
+  {{#if table-column-help-action--attribute}}
+    {{{table-column-help-action--attribute}}}
+  {{/if}}>
+  {{#> button button--modifier="pf-m-plain" button--attribute="More info"}}
```
christiemolloy, line 5: is this meant to be an aria-label?

src/patternfly/components/Table/table-th.hbs
```diff
@@ -29,6 +29,9 @@
   {{#if table-th--attribute}}
     {{{table-th--attribute}}}
   {{/if}}>
+  {{#if table-th--IsColumnHelp}}
+    <div class="pf-c-table__column-help">
```
christiemolloy, line 33: should the class names for these be `pf-c-table__sort-column-help` because from what i can see it should only be included inside table sort? that way someone wont add this to anywhere inside the table

src/patternfly/components/Table/table.scss
```diff
@@ -638,26 +643,65 @@
...
+  --pf-c-table__button--PaddingTop: var(--pf-c-table__sort__button--PaddingTop);
+  --pf-c-table__button--PaddingRight: var(--pf-c-table__sort__button--PaddingRight);
+  --pf-c-table__button--PaddingBottom: var(--pf-c-table__sort__button--PaddingBottom);
+  --pf-c-table__button--PaddingLeft: var(--pf-c-table__sort__button--PaddingLeft);
 
-  .pf-c-table__button:focus {
-    --pf-c-table__sort-indicator--Color: var(--pf-c-table__sort__button--focus__sort-indicator--Color);
-  }
+  .pf-c-table__button {
+    display: inline-flex;
```
christiemolloy, line 689: do you need this

src/patternfly/components/Table/table.scss
```diff
@@ -638,26 +643,65 @@
...
+      outline-style: auto;
+      outline-color: -webkit-focus-ring-color;
+    }
+    // stylelint-enable
+  }
+}
+
+.pf-c-table__column-help {
+  display: flex;
+  align-items: flex-end;
```
christiemolloy, line 662: is this needed

mcoker:
@christiemolloy displaying the popovers on mobile would be tricky because of the way the column header is displayed in the grid view (via pseudo elements). @mcarrano wdyt? If we want to look at supporting that, we can open a separate issue.

src/patternfly/components/Table/table-column-help-action.hbs
mcoker: yes!

src/patternfly/components/Table/table-th.hbs
mcoker: It's a feature of column headers, not just sortable ones. It's like a "more info" blurb about what the column is for.

src/patternfly/components/Table/table.scss
mcoker: Yeah, it keeps everything bottom aligned in the cell.

Here's with

Screen Shot 2020-06-18 at 10 54 54 AM

Here's without

Screen Shot 2020-06-18 at 10 55 03 AM

src/patternfly/components/Table/table.scss
mcoker: It was `display: flex`, so I made it `inline-flex` so that it wouldn't take up the width of the cell... but I refactored that bit so it can stay `display: flex;`. I'm not sure if that's needed for one of the modifiers, but I don't see any harm in leaving it even if it isn't necessary. It will remove any gap between its children (the button text and indicator icon), and allow for better alignment (using flex properties) if that's ever needed.

mcoker:
@christiemolloy @mattnolting updated

src/patternfly/components/Table/table-th.hbs
christiemolloy: should it be `pf-c-table__column-header-help` then?

mcoker: It could but that's longer, and it's really just info about the column. It could also be `pf-c-table__th-help` or something like that. Do you see a problem with it named `__column-help`?

PR #3198, opened 2020-06-19
fix(release-notes): added 2020.08 release notes

mcoker:


RELEASE-NOTES.md
```diff
@@ -4,6 +4,42 @@ section: overview
 releaseNoteTOC: true
 ---
 
+## 2020.08 release notes (2020-06-24)
+Packages released:
+- @patternfly/patternfly@v4.16.5
```
redallen, line 9: Can we change ALL of these to link to NPM like https://www.npmjs.com/package/@patternfly/patternfly/v/4.10.31?

They look a little funny currently: https://www.patternfly.org/v4/documentation/core/overview/release-notes

redallen: This would be nice to have for 2020.08, but if you can't get a PR up tomorrow, that's fine.

PR #3202, opened 2020-06-22
fix(breadcrumb): always use divider and hide first

redallen:
Towards patternfly/patternfly-react#4422

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -84,3 +84,10 @@
     font-weight: var(--pf-global--FontWeight--semi-bold);
   }
 }
+
+// Hide divider of first child - required for react since dividers are included in each item
+// Consider revisiting this in a breaking change release
+.pf-c-breadcrumb__list > :first-child .pf-c-breadcrumb__item-divider {
+  display: none;
+  visibility: hidden;
+}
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -84,3 +84,9 @@
     font-weight: var(--pf-global--FontWeight--semi-bold);
   }
 }
+
+// Hide divider of first child
+.pf-c-breadcrumb__item:first-child .pf-c-breadcrumb__item-divider {
```
mcoker, line 89: Is it guaranteed that all of the `.pf-c-breadcrumb__item`'s will always be siblings of one another? As in a person won't wrap that element in some other thing? If I'm reading https://github.com/patternfly/patternfly-react/issues/4422 properly, seems like that is part of the problem - `.pf-c-breadcrumb__item` isn't a direct descendent of `.pf-c-breadcrumb`.

redallen: It is a direct descendant in the HTML. It is not in the React node hierarchy so the `React.cloneElement` hack I mentioned last week doesn't work :(

If there's a better way to write the selector I'm all for it but ill equipped... are there any good resources to play around with selectors?

src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -84,3 +84,9 @@
     font-weight: var(--pf-global--FontWeight--semi-bold);
   }
 }
+
+// Hide divider of first child
```
mcoker, line 88: Adding a note here for why it's necessary since it isn't intuitive. Also a "breaking change" note so we can revisit this later and try and remove the unnecessary HTML.

```suggestion
// Hide divider of first child - required for react since dividers are included in each item
// Consider revisiting this in a breaking change release
```

PR #3214, opened 2020-06-24
feat(button): add cta variation to button

### LINKED ISSUE #3099: Introduce Call to Action button variant
Want to introduce a new variant of the Button component as a Call to Action button. This intended as a larger, more prominent button or link and it optimized for use in websites vs applications.

Variations to include: Primary, Secondary, and Tertiary.

Visual spec:
CTA buttons specs:
Red Hat Display, Bold, 16pt
16px, 32px, 16px, 32px spacers

CTA link specs:
Red Hat Display, Bold, 18pt, with arrow icon

image

cc @LHinson @mcoker @ajacobs21e

christiemolloy:
closes #3099 

@mcarrano do you know why the Dark theme designs were included in the original issue?

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -135,7 +135,6 @@
   --pf-c-button--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
   --pf-c-button--m-link--m-cta--FontSize: var(--pf-global--FontSize--lg);
   --pf-c-button--m-link--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
-  --pf-c-button--m-link--m-cta--LineHeight: #{pf-size-prem(24px)};
 
   // Styles for an icon in button
   --pf-c-button__icon--m-start--MarginRight: var(--pf-global--spacer--xs);
@@ -339,7 +338,6 @@
     &.pf-m-cta {
       --pf-c-button--FontSize: var(--pf-c-button--m-link--m-cta--FontSize);
       --pf-c-button--FontWeight: var(--pf-c-button--m-link--m-cta--FontWeight);
-      --pf-c-button--LineHeight: var(--pf-c-button--m-link--m-cta--LineHeight);
     }
   }
 
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -134,7 +134,6 @@
   --pf-c-button--m-cta--PaddingLeft: var(--pf-global--spacer--xl);
   --pf-c-button--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
   --pf-c-button--m-link--m-cta--FontSize: var(--pf-global--FontSize--lg);
-  --pf-c-button--m-link--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
 
   // Styles for an icon in button
   --pf-c-button__icon--m-start--MarginRight: var(--pf-global--spacer--xs);
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -128,12 +128,12 @@
   --pf-c-button--m-small--FontSize: var(--pf-global--FontSize--sm);
 
   // CTA Button
-  --pf-c-button--m-cta--PaddingTop: var(--pf-global--spacer--md);
-  --pf-c-button--m-cta--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-button--m-cta--PaddingBottom: var(--pf-global--spacer--md);
-  --pf-c-button--m-cta--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-button--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
-  --pf-c-button--m-link--m-cta--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-button--m-display-lg--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-button--m-display-lg--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-button--m-display-lg--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-button--m-display-lg--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-button--m-display-lg--FontWeight: var(--pf-global--FontWeight--bold);
+  --pf-c-button--m-link--m-display-lg--FontSize: var(--pf-global--FontSize--lg);
 
   // Styles for an icon in button
   --pf-c-button__icon--m-start--MarginRight: var(--pf-global--spacer--xs);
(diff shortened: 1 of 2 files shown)


mcarrano:
@christiemolloy The dark theme versions were more directed to the patternfly-elements folks as I understand since they support both light and dark themes. What you implemented looks good. @ajacobs21e can you also take a look to see if this is what you expected?

ajacobs21e:
looks great @christiemolloy !
The only difference I see is that the font is Red Hat Text instead of Red Hat Display. Is there a reason we have to use RHT?

mcoker:
Just want to confirm we should only support primary, secondary, tertiary, and link as CTAs? Do we want to consider enabling support for the other buttons in the code, but not list them on the website? 

Also for the link CTA, is it intended that we should support both the normal link (with padding), and the inline link with no padding? FWIW, the current link CTA (non-inline) has padding, but it uses normal button padding, not the larger CTA padding. Not sure if we should leave that, update it to the larger padding, or only support an inline-link CTA button with no padding.

Here's what it looks like currently:

Screen Shot 2020-06-24 at 2 16 25 PM

ajacobs21e:
@mcoker I think we should wait on including the dark theme and RCUE theme styles for now since those still haven't been fully defined. 

For the CTA link I think it's almost always going to be used without padding on the left side. It might be worth including both options if there's a use case I'm not aware of. 

Currently on redhat.com the CTA link is only using the text line height without any padding. @mceledonia what do you think? It probably should have some padding and the button padding looks good to me.

mcoker:
> I think we should wait on including the dark theme and RCUE theme styles for now since those still haven't been fully defined.

@ajacobs21e gotcha, I was referring to the danger, plain, and control variations. Currently we only apply CTA styling if the button is used with primary, secondary, tertiary, or link (both regular and inline link). Wondering if we should support those in the code in case someone uses it, or if applying the CTA variation to those buttons should do nothing as it does currently.

> It might be worth including both options if there's a use case I'm not aware of.

+1. It would be needed for something like this as an example, since the padding would give these buttons consistent alignment and spacing. Just like placing a normal primary + link button pair in a row.

Screen Shot 2020-06-25 at 10 58 13 AM

What I'm proposing is updating this, since the padding isn't consistent with CTA buttons - it's CTA typography with non-CTA button padding.

Screen Shot 2020-06-25 at 10 59 52 AM

To this

Screen Shot 2020-06-25 at 11 00 34 AM

And then we would still have the "inline" link, with no padding, like:

Screen Shot 2020-06-25 at 11 00 52 AM

ajacobs21e:
@mcoker gotcha, agree on all of the spacing. Thanks for the examples to help me see what you mean. 

Danger and control don't make sense to me as a CTA. What does a plain button look like / how is it used?

christiemolloy:
hi @ajacobs21e ! A plain button with the modifier `pf-m-cta` would look like this with padding. If you think we should apply CTA on the plain button, would you suggest updating the text size?

Screen Shot 2020-06-25 at 3 02 09 PM

Also @mcoker and @ajacobs21e on the topic of updating the padding for the link button (not inline) so that it matches the primary/secondary/tertiary buttons if its placed next to them.. the design spec calls for the link button to be 18px font size, however the regular button is still 16px font-size. This looks odd when placed together and we will need to adjust the sizing for the inline button so they line up. Thoughts?

Screen Shot 2020-06-25 at 3 05 31 PM

mcarrano:
It is hard for me to foresee a case where you would want a plain button (i.e. and icon button) used as a call to action. But let me know if you think otherwise @ajacobs21e .

ajacobs21e:
Agree, I don't think that's a use case

ajacobs21e:
@christiemolloy I think it's ok that the fonts don't match up. Most of the time only one CTA will be used at a time. @mceledonia any thoughts on this one?

christiemolloy:
Made updates @ajacobs21e @mcarrano 

@mcoker i update the line-height of the 18pt button so that it lines up with the 16px button but there is still a weird discrepancy between the height of the 16pt button and the 18pt button, i fixed it by moving top down 1px but i dont think we would want to add that permanently to the button.

Screen Shot 2020-06-29 at 10 52 55 AM
 
Screen Shot 2020-06-29 at 10 46 43 AM

mcoker:
@christiemolloy since they don't need to line up, and it would look weird anyways since they're different font-sizes, I think the link CTA could just inherit the normal line height, and it will be a taller button.

christiemolloy:
Okay that sounds like a good solution thanks @mcoker

src/patternfly/components/Button/button.scss
```diff
@@ -179,6 +188,19 @@
     --pf-c-button--FontSize: var(--pf-c-button--m-small--FontSize);
   }
 
+  &.pf-m-primary,
+  &.pf-m-secondary,
+  &.pf-m-tertiary,
+  &.pf-m-link:not(.pf-m-inline) {
```
mcoker, line 194: I don't think you need the `:not()` here. `.pf-m-inline` applies `padding: 0` so the vars won't apply anyways. That will keep the specificity of `.pf-m-link.pf-m-cta` the same as the others, and then you could remove `--pf-c-button--m-link--m-cta--FontWeight`, too.

src/patternfly/components/Button/button.scss
```diff
@@ -127,6 +127,15 @@
...
   --pf-c-button--m-small--FontSize: var(--pf-global--FontSize--sm);
 
+  // CTA Button
+  --pf-c-button--m-cta--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-button--m-cta--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-button--m-cta--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-button--m-cta--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-button--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
+  --pf-c-button--m-link--m-cta--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-button--m-link--m-cta--FontWeight: var(--pf-global--FontWeight--bold);
```
mcoker, line 137: looks like you can remove this

src/patternfly/components/Button/examples/Button.md
```diff
@@ -243,3 +270,4 @@ Semantic buttons and links are important for usability as well as accessibility.
 | `.pf-m-active` | `.pf-c-button` | Forces display of the active state of the button. This modifier should be used when `aria-pressed` is set to true so that the button displays in an active state. |
 | `.pf-m-small` | `.pf-c-button` | Modifies the button so that it has small font size. |
 | `.pf-m-aria-disabled` | `.pf-c-button` | Modifies a button to be visually disabled, yet is still focusable. |
+| `.pf-m-cta` | `.pf-c-button` `pf-c-button.pf-m-link` | Modifies the button and link button for "call to action" styles. |
```
mcoker, line 273:
```suggestion
| `.pf-m-cta` | `.pf-c-button`, `pf-c-button.pf-m-link` | Modifies the button and link button for "call to action" styles. |
```

christiemolloy:
@mcarrano @mattnolting updated the modifier name of the button from `pf-m-cta` to `pf-m-display-lg` as per our conversation from last week. @mcoker updated from your review

Issue #3215, opened 2020-06-24
Update wizard nav to support buttons 


mcoker:
To support https://github.com/patternfly/patternfly-react/pull/4425, wizard nav items should support using a `<button>` element.

PR #3216, opened 2020-06-24
fix(wizard): added support for buttons as nav items

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3215


seanforyou23:
Feel free to make any adjustments as necessary to fix that chrome discrepancy you found in https://github.com/patternfly/patternfly-react/pull/4425#discussion_r446218930 - I tested the demos you showed and either way works well from my POV.

Issue #2979, opened 2020-04-18
PF4 Modal: there is not class that can be used to create medium sized modals


KKoukiou:
Similar to pf-m-sm and pf-m-lg we would need a pf-m-md.

Our use case is that we have a lot of modals where the elements inside are configured to expand to the full width with `width: 100%`. It would be ideal to have a class to limit the max width of the modals - but pf-m-sm is too small whilist of-m-lg is too big for our examples.
We can of course can adjust the scss variables, but I was hoping for some cleaner solution by getting a pf-m-md class from you.
Is that possible?

Thanks in advance.

mcoker:
I like it. wdyt @mceledonia @mcarrano

mcarrano:
Makes  sense to me.

PR #3217, opened 2020-06-24
fix(modal): added medium, adjusted maxwidths for better responsiveness

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2979

Noticed the modal demos were broken, so went ahead and added this, too.

@mattnolting @christiemolloy wdyt about this change? Since the size modifiers were applied at breakpoints before, the point at which the size modifier applies needs to be 32px below whatever the breakpoint is in order for the modal to resize smoothly between the size modifier and adjusting with the width of the browser. So the existing small and large modifiers jerk a little right at the point where the modifier is applied. That also meant this medium modifier would need to apply at 736px to not jerk when the modifier is applied, and design wanted that width to be 840px.

The changes I made don't break the variable name API, and the react component's `width` prop will still work as expected. In a breaking change release, I'd propose we change `--pf-c-modal-box--m-sm--sm--MaxWidth` to `--pf-c-modal-box--m-sm--Width` and `--pf-c-modal-box--m-lg--lg--MaxWidth` to `--pf-c-modal-box--m-lg--MaxWidth`.

### DIFF
--- a/src/patternfly/components/ModalBox/examples/ModalBox.md
+++ b/src/patternfly/components/ModalBox/examples/ModalBox.md
@@ -46,6 +46,28 @@ cssPrefix: pf-c-modal-box
 {{/modal-box}}
 ```
 
+```hbs title=Medium
+{{#> modal-box modal-box--modifier="pf-m-md" modal-box--attribute='aria-labelledby="modal-md-title" aria-describedby="modal-md-description"'}}
+  {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close dialog"'}}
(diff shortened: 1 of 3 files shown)


mcoker:
@christiemolloy it's just to keep the modal from touching the sides of the viewport. Even with a full width modal, we add some padding so that it doesn't touch the left/right side of the window.

christiemolloy:
So is it 16px on each side @mcoker ?

mcoker:
@christiemolloy yep. it should be that way with each size variation and without a size variation (full width).

PR #3218, opened 2020-06-25
Feat(Hint): Add hint component

### LINKED ISSUE #2718: Hint component
Want to introduce a Hint component. Design here: https://docs.google.com/document/d/1qAYYv-VPuOst1o-BFx0R7Wgh3jt6X9ATmV4WQU5hXcw/edit?usp=sharing

christiemolloy:
closes #2718

### DIFF
--- a/src/patternfly/components/Hint/examples/Hint.md
+++ b/src/patternfly/components/Hint/examples/Hint.md
@@ -6,10 +6,10 @@ cssPrefix: pf-c-hint
 ---
 
 ## Examples
-```hbs title=Default-with-title
+```hbs title=Hint-with-title
 {{#> hint}}
   {{#> hint-actions}}
-    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
+    {{#> dropdown id="hint-with-title-dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
   {{/hint-actions}}
   {{#> hint-title}}
     Do more with Find it Fix it capabilities
@@ -23,7 +23,7 @@ cssPrefix: pf-c-hint
 
 {{#> hint}}
   {{#> hint-actions}}
-    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
+    {{#> dropdown id="hint-with-title-with-footer-dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
   {{/hint-actions}}
   {{#> hint-title}}
     Do more with Find it Fix it capabilities
@@ -53,7 +53,7 @@ cssPrefix: pf-c-hint
 
 {{#> hint}}
   {{#> hint-actions}}
-    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
+    {{#> dropdown id="hint-with-no-title-dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
   {{/hint-actions}}
   {{#> hint-body}}
     Upgrade to Red Hat Smart Management to remediate all your systems across regions and geographies.
@@ -72,7 +72,7 @@ cssPrefix: pf-c-hint
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-hint` | `<div>` | Initiates the hint component. |
+| `.pf-c-hint` | `<div>` | Initiates the hint component. **Required** |
 | `.pf-c-hint__title` | `<div>` | Initiates the hint title element. |
 | `.pf-c-hint__body` | `<div>` | Initiates the hint body element. |
 | `.pf-c-hint__footer` | `<div>` | Initiates the hint footer element. |
--- a/src/patternfly/components/Hint/examples/Hint.md
+++ b/src/patternfly/components/Hint/examples/Hint.md
@@ -1,0 +1,114 @@
+---
+title: 'Hint'
+section: components
+beta: true
+cssPrefix: pf-c-hint
+---
+
+## Examples
+```hbs title=Default-with-header
+{{#> hint}}
+  {{#> hint-header}}
+    Do more with Find it Fix it capabilities
+  {{/hint-header}}
+  {{#> hint-body}}
+    Upgrade to Red Hat Smart Management to remediate all your systems accross regions and geographies.
+  {{/hint-body}}
+{{/hint}}
+
+{{#> hint}}
+  {{#> hint-header}}
+    Do more with Find it Fix it capabilities
+  {{/hint-header}}
+  {{#> hint-body}}
+    Upgrade to Red Hat Smart Management to remediate all your systems accross regions and geographies.
+  {{/hint-body}}
+  {{#> hint-actions}}
+    {{#> button button--modifier="pf-m-link pf-m-inline"}}
+      Try it for 90 days
+    {{/button}}
+  {{/hint-actions}}
+{{/hint}}
+```
+
+```hbs title=Default-with-no-header
+{{#> hint}}
+  {{#> hint-body}}
+    Upgrade to Red Hat Smart Management to remediate all your systems accross regions and geographies.
+  {{/hint-body}}
+{{/hint}}
+
+{{#> hint}}
+  {{#> hint-body}}
+    Upgrade to Red Hat Smart Management to remediate all your systems accross regions and geographies.
+  {{/hint-body}}
+  {{#> hint-actions}}
+    {{#> button button--modifier="pf-m-link pf-m-inline"}}
+      Try it for 90 days
+    {{/button}}
+  {{/hint-actions}}
+{{/hint}}
+```
+
+```hbs title=Dominant-with-header
+{{#> hint}}
+  {{#> hint-header}}
+    Create service binding
+  {{/hint-header}}
+  {{#> hint-body}}
+    Service bindings create a secret containing the necessary information for another application to use mongodb-persistent. Once the binding is ready, add the secret to your application’s environment variables.
+  {{/hint-body}}
+{{/hint}}
+
+{{#> hint}}
+  {{#> hint-header}}
+    Create service binding
+  {{/hint-header}}
+  {{#> hint-body}}
+    Service bindings create a secret containing the necessary information for another application to use mongodb-persistent. Once the binding is ready, add the secret to your application’s environment variables.
+  {{/hint-body}}
+  {{#> hint-actions}}
+    {{#> button button--modifier="pf-m-secondary"}}
+      Create service binding
+    {{/button}}
+    {{#> button button--modifier="pf-m-link"}}
+      Learn more
+      {{#> button-icon button-icon--modifier="pf-m-end"}}
+        <i class="fas fa-external-link-alt" aria-hidden="true"></i>
+      {{/button-icon}}
+    {{/button}}
+  {{/hint-actions}}
+{{/hint}}
+```
+
+```hbs title=Dominant-with-no-header
+{{#> hint hint--modifier="pf-m-dominant"}}
+  {{#> hint-body}}
+    Service bindings create a secret containing the necessary information for another application to use mongodb-persistent. Once the binding is ready, add the secret to your application’s environment variables.
+  {{/hint-body}}
+{{/hint}}
+
+{{#> hint hint--modifier="pf-m-dominant"}}
+  {{#> hint-body}}
+    Service bindings create a secret containing the necessary information for another application to use mongodb-persistent. Once the binding is ready, add the secret to your application’s environment variables.
+  {{/hint-body}}
+  {{#> hint-actions}}
+    {{#> button button--modifier="pf-m-secondary"}}
+      Create service binding
+    {{/button}}
+    {{#> button button--modifier="pf-m-link"}}
+      Learn more
+      {{#> button-icon button-icon--modifier="pf-m-end"}}
+        <i class="fas fa-external-link-alt" aria-hidden="true"></i>
+      {{/button-icon}}
+    {{/button}}
+  {{/hint-actions}}
+{{/hint}}
+```
+
+## Documentation
+
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
--- a/src/patternfly/components/Hint/examples/Hint.md
+++ b/src/patternfly/components/Hint/examples/Hint.md
@@ -12,12 +12,10 @@ cssPrefix: pf-c-hint
     Do more with Find it Fix it capabilities
   {{/hint-header}}
   {{#> hint-body}}
-    Upgrade to Red Hat Smart Management to remediate all your systems accross regions and geographies.
+    Upgrade to Red Hat Smart Management to remediate all your systems across regions and geographies.
   {{/hint-body}}
   {{#> hint-actions}}
-    {{#> button button--modifier="pf-m-link pf-m-inline"}}
-      Dismiss
-    {{/button}}
+    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
   {{/hint-actions}}
 {{/hint}}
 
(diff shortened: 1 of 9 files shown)


src/patternfly/components/Hint/examples/Hint.md
```diff
@@ -0,0 +1,153 @@
...
+## Examples
+```hbs title=Default-with-header
+{{#> hint}}
+  {{#> hint-header}}
+    Do more with Find it Fix it capabilities
+  {{/hint-header}}
+  {{#> hint-body}}
+    Upgrade to Red Hat Smart Management to remediate all your systems across regions and geographies.
+  {{/hint-body}}
+  {{#> hint-actions}}
```
mcoker, line 17: why do the actions come last?

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,114 @@
+.pf-c-hint {
+  --pf-c-hint__header--GridColumnWidth: auto;
+  --pf-c-hint__actions--GridColumnWidth: calc(var(--pf-global--spacer--2xl) + #{pf-size-prem(44px)});
```
mcoker, line 3: where does this value come from? Why not just let it be the size of the action + the purple margin/gap?

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,114 @@
...
+  --pf-c-hint--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-hint--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-hint--BackgroundColor: var(--pf-global--palette--blue-50);
+  --pf-c-hint--BorderColor: var(--pf-global--palette--blue-100);
+  --pf-c-hint--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-hint--BoxShadow: var(--pf-global--BoxShadow--sm);
+  --pf-c-hint--Color: var(--pf-global--Color--100);
+
+  // Hint Header
+  --pf-c-hint__header--FontSize: var(--pf-global--FontSize--lg);
```
mcoker, line 15: In keeping with the other components, `__header` designates a region (ie, `<header>`), and `__title` is the title (ie, `<h1>`). If we're setting `font-size`, I would expect this to be a title element, like the alert (which this layout seems to be modeled after).

mcoker:
@maryshak1996 looking at the design, it looks like the content to the left of the kebab (title, body, footer/actions) never wraps underneath the kebab. We have this layout in a lot of components, but the pattern is that whatever is directly beside the kebab (either the title or the body) never wraps underneath, but any elements under the kebab do. A few examples of this are the alert, card, and popover (below). Is the difference in the hint component intended?

Screen Shot 2020-06-25 at 5 02 06 PM
Screen Shot 2020-06-25 at 5 03 18 PM
Screen Shot 2020-06-25 at 5 04 06 PM

src/patternfly/components/Hint/hint.scss
mcoker: From this design, looks like this column width should fit whatever's placed in it

Screen Shot 2020-06-25 at 5 09 54 PM

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,114 @@
...
+
+  // Hint Body
+  --pf-c-hint__body--FontSize: var(--pf-global--FontSize--md);
+  --pf-c-hint__header--body--MarginTop: var(--pf-global--spacer--md);
+
+  // Hint Footer
+  --pf-c-hint__footer--child--MarginRight: var(--pf-global--spacer--md);
+  --pf-c-hint__header--footer--MarginTop: var(--pf-global--spacer--md);
+  --pf-c-hint__body--footer--MarginTop: var(--pf-global--spacer--md);
+
```
mcoker, line 25: if the gap is always 16px, and we stick with grid, couldn't you just use `grid-row-gap` on `.pf-c-hint`?

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,114 @@
...
+  display: grid;
+  grid-template-columns: var(--pf-c-hint__header--GridColumnWidth) var(--pf-c-hint__actions--GridColumnWidth);
+  padding: var(--pf-c-hint--PaddingTop) var(--pf-c-hint--PaddingRight) var(--pf-c-hint--PaddingBottom) var(--pf-c-hint--PaddingLeft);
+  color: var(--pf-c-hint--Color);
+  background-color: var(--pf-c-hint--BackgroundColor);
+  border: var(--pf-c-hint--BorderWidth) solid var(--pf-c-hint--BorderColor);
+  box-shadow: var(--pf-c-hint--BoxShadow);
+
+  &.pf-m-dominant {
+    border: unset;
```
mcoker, line 44: If you're just disabling the border, `border: 0;` is probably better. `unset` is like a 2 step evaluation depending on whether the property is inheritable. It will evaluate to `initial` for `border`.

You could also do `--pf-c-hint--BorderWidth: 0;`

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,114 @@
...
+
+    --pf-c-hint--Color: var(--pf-c-hint--m-dominant--Color);
+    --pf-c-hint--BackgroundColor: var(--pf-c-hint--m-dominant--BackgroundColor);
+    --pf-c-hint--BoxShadow: var(--pf-c-hint--m-dominant--BoxShadow);
+
+    .pf-c-button,
+    .pf-c-dropdown__toggle {
```
mcoker, line 51: The `pf-m-dominant` variation is a tricky one. It's effectively a theme. All of its contents need to change so that text, borders, shadows, etc, work against this background. For example, here's what it looks like with an expandable section:

Screen Shot 2020-06-25 at 5 27 31 PM

The button text is the same as the background. Here it is focused

Screen Shot 2020-06-25 at 5 27 41 PM

To do this right, I think we need the concept of themed components. The closest we have is our existing implementation of dark theme - it doesn't work with the blue background color, but here's black for example.

Screen Shot 2020-06-25 at 5 30 54 PM

Unless we want to restrict the type of content that can be placed in a hint (limit it to text, buttons, and dropdowns, for example, and style those elements with the correct colors), we may want to 

* Update the dominant variation so that it uses a normal background and everything that's placed inside of it doesn't need to be re-themed/styled so that it's visible and accessible. This is probably what I would recommend.
* Consider using the existing dark theme feature we have (would require use a different background than the dark blue), and testing to make sure our components work. This could open a can of worms since we don't use the dark theme extensively and might discover things don't work as expected.

mcoker: cc @maryshak1996 @mcarrano

mcarrano: These are great points @mcoker . But before we go back and redesign this, is it possible to restrict what goes inside of here? I don't think we were intending that the hint was a general container that you can drop anything into. Thoughts @maryshak1996 ?

christiemolloy:
> @maryshak1996 looking at the design, it looks like the content to the left of the kebab (title, body, footer/actions) never wraps underneath the kebab. We have this layout in a lot of components, but the pattern is that whatever is directly beside the kebab (either the title or the body) never wraps underneath, but any elements under the kebab do. A few examples of this are the alert, card, and popover (below). Is the difference in the hint component intended?
> 
> Screen Shot 2020-06-25 at 5 02 06 PM
> 
> Screen Shot 2020-06-25 at 5 03 18 PM
> 
> Screen Shot 2020-06-25 at 5 04 06 PM

@mcoker on mobile the body text wraps underneath the kebab (provided theres a header)

Screen Shot 2020-06-26 at 9 31 54 AM

src/patternfly/components/Hint/hint.scss
mcoker: @mcarrano yeah that's an option, it just currently seems like we'll run into problems when someone puts an element in the container we aren't anticipating, whether we want to support that use case or not. Like here's what it looks like if you use a link currently - 

Screen Shot 2020-06-26 at 8 26 54 AM

Screen Shot 2020-06-26 at 8 27 04 AM

mcarrano: Yeah, that's a problem if we can't put link text into the message. Let me pull together a meeting to talk this through and decide how best to address.

maryshak1996:
> @maryshak1996 looking at the design, it looks like the content to the left of the kebab (title, body, footer/actions) never wraps underneath the kebab. We have this layout in a lot of components, but the pattern is that whatever is directly beside the kebab (either the title or the body) never wraps underneath, but any elements under the kebab do. A few examples of this are the alert, card, and popover (below). Is the difference in the hint component intended?
> 
> Screen Shot 2020-06-25 at 5 02 06 PM
> 
> Screen Shot 2020-06-25 at 5 03 18 PM
> 
> Screen Shot 2020-06-25 at 5 04 06 PM

@mcoker 
I think that these are all really good points, and I think that the treatment for the content around the kebab that you suggested would be totally fine -- there really was no intentional reason for not doing that in my design to begin with!

src/patternfly/components/Hint/hint.scss
maryshak1996: @mcarrano I agree, I don't think that we want too much flexibility with what goes into the hint.
@mcoker I'm a little hesitant to support a black dominant hint, since in many cases the hint will be housed close to the masthead, which is also black -- I think that it might not have the same effect when there's a black hint underneath a black masthead

christiemolloy:
@mcoker @maryshak1996 @mcarrano removed the dominant variation

src/patternfly/components/Hint/examples/Hint.md
christiemolloy: Moved them to after the header (similar to what is in the card)

src/patternfly/components/Hint/hint.scss
christiemolloy: great point, updated header to be title

src/patternfly/components/Hint/hint.scss
christiemolloy: updated

christiemolloy:
Updated @mcoker

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,84 @@
...
+  grid-column: -1 / 1;
+  font-size: var(--pf-c-hint__body--FontSize);
+
+  + .pf-c-hint__footer {
+    margin-top: var(--pf-c-hint__body--footer--MarginTop);
+  }
+}
+
+.pf-c-hint__footer {
+  grid-column: -1 / 1;
```
mcoker, line 79:
```suggestion
  grid-column: 1 / -1;
```

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,84 @@
...
+    text-align: left;
+    white-space: pre-wrap;
+  }
+}
+
+.pf-c-hint__actions {
+  margin-left: var(--pf-c-hint__actions--MarginLeft);
+  text-align: right;
+  grid-column: 2;
+  grid-row: 1 / span 3;
```
mcoker, line 46: do you need this? It should just span 1 column, so `grid-row` shouldn't be necessary.

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,84 @@
...
+
+  .pf-c-dropdown {
```
mcoker, line 48: This should only apply to plain dropdowns, in case someone puts a regular dropdown there.

src/patternfly/components/Hint/examples/Hint.md
mcoker: This seems fine to me, the modal and popover are the same way (action comes first), though I'm not sure if that's specific to a close button in an element like that or if it can apply to general actions. @jessiehuff wdyt?

src/patternfly/components/Hint/hint.scss
christiemolloy: yep dont need this anymore

src/patternfly/components/Hint/examples/Hint.md
```diff
@@ -0,0 +1,79 @@
...
+section: components
+beta: true
+cssPrefix: pf-c-hint
+---
+
+## Examples
+```hbs title=Default-with-title
+{{#> hint}}
+  {{#> hint-actions}}
+    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
```
mattnolting, line 12:
```suggestion
    {{#> dropdown id="default-with-title-dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
```

src/patternfly/components/Hint/examples/Hint.md
```diff
@@ -0,0 +1,79 @@
...
+      Learn more about the improved features.
+    {{/button}}
+  {{/hint-body}}
+{{/hint}}
+
+<br>
+
+{{#> hint}}
+  {{#> hint-actions}}
+    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
```
mattnolting, line 56: update `id`

src/patternfly/components/Hint/examples/Hint.md
```diff
@@ -0,0 +1,79 @@
...
+  {{#> hint-body}}
+    Upgrade to Red Hat Smart Management to remediate all your systems across regions and geographies.
+  {{/hint-body}}
+{{/hint}}
+
+<br>
+
+{{#> hint}}
+  {{#> hint-actions}}
+    {{#> dropdown id="dropdown-kebab" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
```
mattnolting, line 26: update `id`

src/patternfly/components/Hint/examples/Hint.md
```diff
@@ -0,0 +1,79 @@
...
+{{/hint}}
+```
+
+## Documentation
+
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-hint` | `<div>` | Initiates the hint component. |
```
mattnolting, line 75:
```suggestion
| `.pf-c-hint` | `<div>` | Initiates the hint component. **Required** | 
```

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,85 @@
+.pf-c-hint {
+  --pf-c-hint--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-hint--PaddingRight: var(--pf-global--spacer--lg);
+  --pf-c-hint--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-hint--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-hint--BackgroundColor: var(--pf-global--palette--blue-50);
+  --pf-c-hint--BorderColor: var(--pf-global--palette--blue-100);
+  --pf-c-hint--BorderWidth: var(--pf-global--BorderWidth--sm);
```
mattnolting, line 8: Should we use a `::before` pseudo for borders?

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,85 @@
...
+  border: var(--pf-c-hint--BorderWidth) solid var(--pf-c-hint--BorderColor);
+  box-shadow: var(--pf-c-hint--BoxShadow);
+
+  .pf-c-button.pf-m-link.pf-m-inline {
+    text-align: left;
+    white-space: pre-wrap;
+  }
+}
+
+.pf-c-hint__actions {
```
mattnolting, line 42: If there's ever multiple actions in `__action` adding this would handle it.
```
  display: inline-grid;
  grid-auto-flow: column;
```

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,85 @@
...
+.pf-c-hint__title {
+  font-size: var(--pf-c-hint__title--FontSize);
+
+  + .pf-c-hint__body {
+    margin-top: var(--pf-c-hint__title--body--MarginTop);
+  }
+
+  + .pf-c-hint__footer {
+    margin-top: var(--pf-c-hint__title--footer--MarginTop);
+  }
```
mattnolting, line 67: replace w/`row-gap`, as @mcoker mentioned

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.pf-c-hint__body {
+  grid-column: -1 / 1;
+  font-size: var(--pf-c-hint__body--FontSize);
+
+  + .pf-c-hint__footer {
+    margin-top: var(--pf-c-hint__body--footer--MarginTop);
+  }
```
mattnolting, line 76: replace w/row-gap

src/patternfly/components/Hint/hint.scss
christiemolloy: I didn't add a pseudo because there's no hover/focus state on the hint

christiemolloy: but I could, in case it has a state in the future

src/patternfly/components/Hint/hint.scss
```diff
@@ -0,0 +1,75 @@
...
+  grid-column: 2;
+  grid-row: 1;
+
+  // Only applies to the plain dropdown
+  .pf-c-dropdown .pf-c-dropdown__toggle.pf-m-plain {
+    margin-top: var(--pf-c-hint__actions--c-dropdown--MarginTop);
+  }
+
+  + .pf-c-hint__body {
+    grid-column: 1;
```
mattnolting, line 56: Nice!

PR #3220, opened 2020-06-26
Feat(NotificationDrawer): Add notification demo

### LINKED ISSUE #3186: Add Notification Drawer Demo
As a follow up to #3180 we need to investigate why when the notification drawer is placed inside of the drawer, the notifications groups are not behaving as expected. You can see how they should behave by looking at the full size demo of the notification drawer.

We should also see if we can replicate the bug that Openshift was having when they were seeing double scrollbars, in order to know if its a PF or Openshift specific problem.

We should also investigate if we can adjust the css in the page/drawer components so that we can support products that do not use the `pf-c-main` component, similar to how the DOM structure is in Openshift.
cc: @mcoker and @mattnolting let me know if theres anything i missed

christiemolloy:
Closes #3186 

Adds the notification drawer demo and includes a fix for the two bugs that were happening when you add the notification drawer inside of the drawer:

- Adds `height: 100%` to `pf-c-drawer__body`
- Adds option to add `overflow: hidden` to `pf-c-drawer__panel`


christiemolloy:
@jcaianirh from what I know I think having the entire notification drawer scroll when there are groups is an override to what we offer in PatternFly, but ill defer to @mcoker and @mcarrano because that might be a variation that we can offer.

@mcarrano @mcoker I wanted to experiment with having two groups open at once to make sure double scrollbars dont appear. Is this expected in the notification drawer, or is it the case that as soon as you expand a group, the other groups close?

mcarrano:
@christiemolloy the way this is implemented now matches the design intent. @jcaianirh is there a reason you would not want it to work this way?  I would not be in favor of having multiple examples. We should just agree on the way that this works and have a consistent approach. As to the correct accordion behavior, looks like the way it is implemented in React, each group opens independently (https://www.patternfly.org/v4/documentation/react/components/notificationdrawer#groups). I'm fine wit that.

Should the demo also include an example of the basic drawer without groups?

christiemolloy:
@mcarrano that sounds good to me. I'll add that as a second example for the demo

mcarrano:
@mcoker @mceledonia interested to hear your thoughts on the scrolling issue.

mcarrano:
@christiemolloy after chatting with @jcaianirh @mcoker and @mceledonia , we decided that the default behavior should be as @jcaianirh describes - that each group opens independently and a scroll bar is exposed for the entire contents of the drawer rather than constraining the height of the group as you currently have it. So this demo should then include two versions of the drawer. One with groups and the basic drawer without groups. Thanks!

christiemolloy:
@mcarrano @jcaianirh @mcoker updated the examples so now there is both variations, and the notification drawer's default behavior has changed

jcaianirh:
@christiemolloy So, the contents of the notification drawer body are supposed to scroll, but not the header. The header remains in place while the body contents scroll vertically. Thanks for updating the behavior.

christiemolloy:
Ah thanks for catching that! will update shortly

christiemolloy:
@jcaianirh do you mind taking a look at it now ?

jcaianirh:
> * We currently have a Drawer demo and a Notification Drawer demo. Any reason we need both? THey seem like the same thing.
@mcarrano good point looks the same to me

> * @mattnolting is currently working on updating the Notification Badge in the masthead. Will that change just get picked up here? I assume so, but wanted to make sure you are aware. If needed you guys should coordinate.
@mcarrano i would think if this is gets picked up here it would still need modification since the new badge has more info to offer than just a boolean on/off...there are counts and severities needed to populate it with corresponding data. So @christiemolloy this will probably need to be updated when the new badge is delivered.

christiemolloy:
@mcarrano @jcaianirh 

- Updated the examples to say 2 unread instead of 3 unread
- We have a Drawer demo and Notification demo because the drawer is a separate component where any kind of list could go into the Drawer vs the Notification demo is its own demo, but let me know if you think otherwise.
- Ill coordinate with @mattnolting on updating this demo once the badge is complete!

mcarrano:
> We have a Drawer demo and Notification demo because the drawer is a separate component where any kind of list could go into the Drawer vs the Notification demo is its own demo, but let me know if you think otherwise.

I understand what you are saying, but it just seems odd to have two versions of the same thing. Maybe we just go ahead and merge this then and we can create something different for the Drawer demo at a later date. Or do we even need a Drawer demo on it's own? Thoughts @mcoker ?

I will go ahead and give this one more review and approve.

mcoker:
@mcarrano the drawer demo was to create it in context of the page component for the use with notification drawer, but we didn't have the notification drawer built yet. Now that we have the notification drawer demo, I think having the two doesn't seem necessary as they are, since the "drawer" demo is just an empty drawer.

> we can create something different for the Drawer demo at a later date

I'm in favor of that. Maybe showing how to use a simple list in a drawer in a page section, or other uses of the drawer. If it makes sense to you, we could go ahead and move this demo the the "drawer" demo page, and just have this as a drawer demo called "notification drawer."

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -193,6 +197,7 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 
 // Modified content children
 .pf-c-drawer__body {
+  height: 100%;
```
mattnolting, line 200: Adding this here disallows multiple `__body` sections where the last section stretches the remaining height https://github.com/patternfly/patternfly/pull/3220/files#diff-56492175f881007db175cb81ee985735L225. I think it should be
```
.pf-c-drawer__body:only-child {
  height: 100%;
}

File: src/patternfly/components/NotificationDrawer/notification-drawer.scss [hunk 2]
```diff
@@ -125,8 +125,7 @@
 }
 
 .pf-c-notification-drawer__body {
-  flex: 1;
-  overflow: auto;
+  overflow-y: scroll;
```
mattnolting, line 128:
```suggestion
  overflow-y: auto;
```

src/patternfly/demos/NotificationDrawer/drawer-demo-notification-drawer-basic.hbs
```diff
@@ -0,0 +1,16 @@
...
+      2 unread
+    {{/notification-drawer-header-status}}
+    {{#> notification-drawer-header-action}}
+      {{#> dropdown id=(concat notification-drawer--id "-header-action") dropdown-menu--modifier="pf-m-align-right" dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}{{/dropdown}}
+    {{/notification-drawer-header-action}}
+  {{/notification-drawer-header}}
+  {{#> notification-drawer-body}}
+    {{> notification-drawer-basic-list}}
+  {{/notification-drawer-body}}
+{{/notification-drawer}}
```
mattnolting, line 16:
```suggestion
{{/notification-drawer}}

```

src/patternfly/demos/NotificationDrawer/drawer-demo-notification-drawer.hbs
```diff
@@ -0,0 +1,73 @@
...
+              {{#> badge badge--modifier="pf-m-unread"}}2{{/badge}}
+            {{/notification-drawer-group-toggle-count}}
+            {{> notification-drawer-group-toggle-icon}}
+          {{/notification-drawer-group-toggle}}
+        </h1>
+        {{> notification-drawer-basic-list}}
+      {{/notification-drawer-group}}
+    {{/notification-drawer-group-list}}
+  {{/notification-drawer-body}}
+{{/notification-drawer}}
```
mattnolting, line 73:
```suggestion
{{/notification-drawer}}

```

mcarrano:
> If it makes sense to you, we could go ahead and move this demo the the "drawer" demo page, and just have this as a drawer demo called "notification drawer."

@mcoker I'd rather not do this because the notification drawer is such a simple thing and I think people will look for it under that name in the navigation. Let's leave the drawer demo as is for now (just an empty drawer). I've opened a new issue to design a more robust drawer demo in the future that can include additional features like the horizontal drawer and splitter that are planned. https://github.com/patternfly/patternfly-design/issues/891

mcoker:
I wanted to get rid of the `.pf-m-overflow-hidden` class and was able to do so, but this exposed a bug @jcaianirh and I discovered earlier in the year where there would be a lot of unnecessary white space at the bottom of the drawer, causing a double scrollbar. After troubleshooting that (for way too long), I tracked it down to `position: fixed` on the `.pf-screen-reader` text in the notification alert item title text inside of the drawer panel that has `translateX()`. This is a chrome specific issue (FF works just fine), and I created a simple demo outlining the problem here - https://codepen.io/mcoker/pen/eYJxbzb?editors=1100

I've [opened a PR](https://github.com/christiemolloy/patternfly-next/pull/29) on @christiemolloy's branch with some updates that enable dropping `.pf-m-overflow-hidden` and clean up some styles left over from when the notification drawer was more flexy in its overflow behavior.

Let's continue to discuss a solution to the screen reader text issue.

mcoker:
@christiemolloy noticed a bug in safari where `height: 100%` on the notification drawer isn't working due to `.pf-c-drawer__body:last-child {flex: 1 1 auto; }` - the `auto` flex basis isn't giving it an implicit height, so `height: 100%` won't work.

Just changing that to `flex: 1 1;` will fix it - Safari will assign `flex-basis: 0%;` in the absence of it declared int he shorthand value.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -219,11 +220,10 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   &:not(.pf-m-no-padding) + * {
     --pf-c-drawer--child--PaddingTop: 0;
   }
-}
 
-// Body
-.pf-c-drawer__body:last-of-type {
-  flex: 1 0 auto;
+  &:last-child {
+    flex: 1 1 auto;
```
mcoker, line 225:
```suggestion
    flex: 1 1;
```

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -96,6 +96,7 @@
   --pf-c-notification-drawer__group--m-expanded__group-toggle-icon--Rotate: 90deg;
 
   display: flex;
+  flex: 1 1;
```
mcoker, line 99:
```suggestion
```

Issue #3222, opened 2020-06-29
Select - add example of checkbox select with description


mcoker:
Need to show how to add a description to a checkbox select. We should reuse the check component's description feature.

PR #3224, opened 2020-06-29
fix(select): added support for checkbox select with description

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3222

this is needed for https://github.com/patternfly/patternfly-react/pull/4460


mcoker:
@christiemolloy updated!

mcoker:
@christiemolloy actually just fixed this small issue, too, where the checkbox label was wrapping when there was a description present - https://github.com/patternfly/patternfly-react/issues/4477

PR #3231, opened 2020-07-01
Feat notification badge enhancements

### LINKED ISSUE #2994: Notification badge enhancements
Update the Notification Badge component to include an optional count field and attention state. Designs here: https://marvelapp.com/4f0341a

cc @mcoker @lhinson

mattnolting:
closes #2994

### DIFF
--- a/src/patternfly/components/NotificationBadge/notification-badge.scss
+++ b/src/patternfly/components/NotificationBadge/notification-badge.scss
@@ -20,8 +20,8 @@
   --pf-c-notification-badge--after--TranslateY: 0; // remove at breaking change
 
   // The icon is normalized to be square in order to facilitate consistency in positioning
-  --pf-c-notification-badge__i--Width: auto;
-  --pf-c-notification-badge__i--Height: auto;
+  --pf-c-notification-badge__i--Width: auto; // remove at breaking change
+  --pf-c-notification-badge__i--Height: auto; // remove at breaking change
 
   // Modifiers
   --pf-c-notification-badge--m-read--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
--- a/src/patternfly/components/NotificationBadge/notification-badge.scss
+++ b/src/patternfly/components/NotificationBadge/notification-badge.scss
@@ -1,32 +1,27 @@
 .pf-c-notification-badge {
-  --pf-c-notification-badge--Color: var(--pf-global--ZIndex--sm);
-  --pf-c-notification-badge--child--ZIndex: var(--pf-global--ZIndex--sm);
-
   // Notification badge
   --pf-c-button--c-notification-badge--MarginRight: calc(var(--pf-global--spacer--xs) * -1);
   --pf-c-button--c-notification-badge--MarginLeft: calc(var(--pf-global--spacer--xs) * -1);
 
+  // update all --after variables to --before at breaking change
   // ::after border treatment
   --pf-c-notification-badge--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
   --pf-c-notification-badge--after--BorderRadius: var(--pf-global--BorderRadius--md);
   --pf-c-notification-badge--after--BorderWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-notification-badge--after--ZIndex: var(--pf-global--ZIndex--xs);
-  --pf-c-notification-badge--after--Top: 0;
-  --pf-c-notification-badge--after--Right: 0;
-  --pf-c-notification-badge--after--Bottom: 0;
-  --pf-c-notification-badge--after--Left: 0;
-  --pf-c-notification-badge--after--Width: auto;
-  --pf-c-notification-badge--after--Height: auto;
+  --pf-c-notification-badge--after--Top: 0; // remove at breaking change
+  --pf-c-notification-badge--after--Right: 0; // remove at breaking change
+  --pf-c-notification-badge--after--Width: auto; // remove at breaking change
+  --pf-c-notification-badge--after--Height: auto; // remove at breaking change
   --pf-c-notification-badge--after--BackgroundColor: transparent;
 
   // This translation works for the Bell icon, although adjusted for a happy medium between FF and Chrome.
   // If other icons are used, a modifier may need to be created to allow for proper positioning on that icon.
-  --pf-c-notification-badge--after--TranslateX: 0;
-  --pf-c-notification-badge--after--TranslateY: 0;
+  --pf-c-notification-badge--after--TranslateX: 0; // remove at breaking change
+  --pf-c-notification-badge--after--TranslateY: 0; // remove at breaking change
 
   // The icon is normalized to be square in order to facilitate consistency in positioning
-  --pf-c-notification-badge__i--Width: var(--pf-global--spacer--md);
-  --pf-c-notification-badge__i--Height: var(--pf-global--spacer--md);
+  --pf-c-notification-badge__i--Width: auto;
+  --pf-c-notification-badge__i--Height: auto;
 
   // Modifiers
   --pf-c-notification-badge--m-read--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
@@ -46,18 +41,15 @@
   --pf-c-notification-badge__count--MarginLeft: var(--pf-global--spacer--xs);
 
   position: relative;
-  display: inline-block;
-  margin-right: var(--pf-c-button--c-notification-badge--MarginRight);
-  margin-left: var(--pf-c-button--c-notification-badge--MarginLeft);
+  display: inline-flex;
   line-height: 1;
 
-  &::after {
+  &::before { // update all --after vars to --before at breaking change
     position: absolute;
     top: var(--pf-c-notification-badge--after--Top);
     right: var(--pf-c-notification-badge--after--Right);
-    bottom: var(--pf-c-notification-badge--after--Bottom);
-    left: var(--pf-c-notification-badge--after--Left);
-    z-index: var(--pf-c-notification-badge--after--ZIndex);
+    bottom: 0;
+    left: 0;
     width: var(--pf-c-notification-badge--after--Width);
     height: var(--pf-c-notification-badge--after--Height);
     content: "";
@@ -69,6 +61,8 @@
 
   .pf-c-button & {
     position: static;
+    margin-right: var(--pf-c-button--c-notification-badge--MarginRight);
+    margin-left: var(--pf-c-button--c-notification-badge--MarginLeft);
   }
 
   & > i {
@@ -78,7 +72,6 @@
 
   & > * {
     position: relative;
-    z-index: var(--pf-c-notification-badge--child--ZIndex);
   }
 
   &.pf-m-unread {
--- a/src/patternfly/components/NotificationBadge/notification-badge.scss
+++ b/src/patternfly/components/NotificationBadge/notification-badge.scss
@@ -1,5 +1,4 @@
 .pf-c-notification-badge {
-
   // update all --after variables to --before at breaking change
   // ::after border treatment
   --pf-c-notification-badge--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
@@ -58,8 +57,6 @@
 
   .pf-c-button & {
     position: static;
-    margin-right: var(--pf-c-button--c-notification-badge--MarginRight);
-    margin-left: var(--pf-c-button--c-notification-badge--MarginLeft);
   }
 
   & > i {
--- a/src/patternfly/components/NotificationBadge/notification-badge.scss
+++ b/src/patternfly/components/NotificationBadge/notification-badge.scss
@@ -35,7 +35,7 @@
   // Badge count
   --pf-c-notification-badge__count--MarginLeft: var(--pf-global--spacer--xs);
 
-  position: relative;
+  position: static;
   display: inline-flex;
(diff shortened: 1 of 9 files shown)


src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -176,6 +176,7 @@ $pf-global--BorderColor--300:    $pf-color-black-200 !default;
 $pf-global--BorderColor--dark-100:   $pf-color-black-300 !default;
 $pf-global--BorderColor--light-100:  $pf-color-black-400 !default;
 $pf-global--BorderRadius--sm:        3px !default;
+$pf-global--BorderRadius--md:        5px !default;
```
christiemolloy, line 179: did you confirm with design that they definitely wanted 5px instead of 3 ?

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -1,35 +1,63 @@
...
+  --pf-c-notification-badge--after--BorderRadius: var(--pf-global--BorderRadius--md);
   --pf-c-notification-badge--after--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-notification-badge--after--ZIndex: var(--pf-global--ZIndex--xs);
   --pf-c-notification-badge--after--Top: 0;
   --pf-c-notification-badge--after--Right: 0;
-  --pf-c-notification-badge--after--Width: calc(var(--pf-global--spacer--sm) + var(--pf-c-notification-badge--after--BorderWidth) + var(--pf-c-notification-badge--after--BorderWidth));
-  --pf-c-notification-badge--after--Height: calc(var(--pf-global--spacer--sm) + var(--pf-c-notification-badge--after--BorderWidth) + var(--pf-c-notification-badge--after--BorderWidth));
+  --pf-c-notification-badge--after--Bottom: 0;
+  --pf-c-notification-badge--after--Left: 0;
+  --pf-c-notification-badge--after--Width: auto;
```
christiemolloy, line 18: is there any point to this ? if you dont want to remove it because of breaking changes could you leave a comment next to it. same as below

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -39,17 +67,39 @@
     transform: translate(var(--pf-c-notification-badge--after--TranslateX), var(--pf-c-notification-badge--after--TranslateY));
   }
 
+  .pf-c-button & {
+    position: static;
```
mcoker, line 71: is there a use of the notification badge outside of a button? It seems like the negative left/right margin assumes this will always be placed inside of a button?

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -1,35 +1,63 @@
...
+  // Badge count
+  --pf-c-notification-badge--m-attention--Color: var(--pf-global--Color--light-100);
+  --pf-c-notification-badge--m-attention--after--BorderColor: transparent;
+  --pf-c-notification-badge--m-attention--after--BackgroundColor: var(--pf-global--danger-color--100);
+
+  // Badge count
+  --pf-c-notification-badge__count--MarginLeft: var(--pf-global--spacer--xs);
 
   position: relative;
   display: inline-block;
```
mcoker, line 49: wdyt about changing this to `inline-flex` with `align-items`? That would also normalize the spacing here in case either our or an apps react removes the space in the HTML between the icon and the count.

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -1,35 +1,63 @@
...
   // This translation works for the Bell icon, although adjusted for a happy medium between FF and Chrome.
   // If other icons are used, a modifier may need to be created to allow for proper positioning on that icon.
-  --pf-c-notification-badge--after--TranslateX: 50%;
-  --pf-c-notification-badge--after--TranslateY: calc(var(--pf-c-notification-badge--after--BorderWidth) * -1);
+  --pf-c-notification-badge--after--TranslateX: 0;
+  --pf-c-notification-badge--after--TranslateY: 0;
 
   // The icon is normalized to be square in order to facilitate consistency in positioning
   --pf-c-notification-badge__i--Width: var(--pf-global--spacer--md);
   --pf-c-notification-badge__i--Height: var(--pf-global--spacer--md);
```
mcoker, line 29: I think these could be set to `auto` now since we aren't positioning the unread dot in this component. Also with a comment to remove them in a breaking change.

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -1,35 +1,63 @@
...
+
+  // Notification badge
+  --pf-c-button--c-notification-badge--MarginRight: calc(var(--pf-global--spacer--xs) * -1);
+  --pf-c-button--c-notification-badge--MarginLeft: calc(var(--pf-global--spacer--xs) * -1);
+
+  // ::after border treatment
+  --pf-c-notification-badge--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-notification-badge--after--BorderRadius: var(--pf-global--BorderRadius--md);
   --pf-c-notification-badge--after--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-notification-badge--after--ZIndex: var(--pf-global--ZIndex--xs);
```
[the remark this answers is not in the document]
mattnolting: The notification badge uses an `::after` pseudo element for decoration. In order to retain possible color modifications to notification state while using `::after` to draw the border, it must be placed behind the text. This should be changed to a `::before` pseudo at breaking change.

mattnolting:
> Should you add this example with the exclamation mark on the bell?
> Screen Shot 2020-07-02 at 10 51 54 AM

That's a `Pro` icon, we don't have access to it.

mattnolting:
@mceledonia @mcarrano After chatting w/@mcoker and @christiemolloy, we concluded that the notification badge, inside a button, inside the page header should share the same padding as the button component as well as the other page header buttons. I initially added a larger border radius, as the Sketch file aligns stroke to `outside`, which presents as more rounded than 3px, but then reverted as this too would deviate from the button component. Please review the implementation vs design. Thanks!
- 
https://patternfly-pr-3231.surge.sh/documentation/core/demos/drawer
https://patternfly-pr-3231.surge.sh/documentation/core/components/notificationbadge

src/patternfly/components/NotificationBadge/notification-badge.scss
mattnolting: I don't think so. I removed the negative margin as this presents the badge inconsistently with the other buttons.

mcarrano:
I will defer to @mceledonia on the design question. 

It looks like what we are doing here then is to replace the old notification badge (with the blue dot) with this new design, is that right? I think that's the outcome we want But what is the ripple effect of that? I don't see a Beta tag on this, although as far as I know no one is using the old/original design. But I guess technically we are breaking this. Thoughts @mcoker ?

Also, this might be outside of the scope of this issue, but I'm looking at the Drawer demo and see this:

Screen Shot 2020-07-08 at 2 39 36 PM
 Should the Help and Settings icon have the same hover appearance? They don't currently. Also, is it just me or are the Help and Settings icons misaligned?

mcoker:
@mcarrano so far from what I've seen @mattnolting has made this change quite gracefully, so that it's not a breaking change in the code - only from a visual perspective changing the circle to a border. The only problems that might cause from what I can tell are 1) it looks different and 2) if a user had customized the size or positioning of the circle, those changes wouldn't apply to the border, and they may end up with a weird looking unread indicator. However, other modifications to the circle, such as the color, would still apply correctly to the new design.

christiemolloy:
minor update: but could you add spacing in between the notification badges

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -1,35 +1,52 @@
...
-  --pf-c-notification-badge--m-unread--after--BackgroundColor: var(--pf-global--active-color--400);
+
+  // Badge count
+  --pf-c-notification-badge--m-unread--after--BorderColor: var(--pf-global--active-color--100);
+  --pf-c-notification-badge--m-unread--after--BorderWidth: var(--pf-global--BorderWidth--md);
+  --pf-c-notification-badge--m-unread--after--BackgroundColor: transparent;
+
+  // Badge count
+  --pf-c-notification-badge--m-attention--Color: var(--pf-global--Color--light-100);
+  --pf-c-notification-badge--m-attention--after--BorderColor: transparent;
```
christiemolloy, line 33: does this need a variable or will it always be transparent ?

src/patternfly/components/Page/page.scss
```diff
@@ -51,10 +51,10 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
...
 
   // Header tools group button
-  --pf-c-page__header-tools--c-button--m-selected--before--Width: #{pf-size-prem(36px)};
-  --pf-c-page__header-tools--c-button--m-selected--before--Height: #{pf-size-prem(36px)};
+  --pf-c-page__header-tools--c-button--m-selected--before--Width: auto; // remove at breaking change
+  --pf-c-page__header-tools--c-button--m-selected--before--Height: auto; // remove at breaking change
   --pf-c-page__header-tools--c-button--m-selected--before--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
-  --pf-c-page__header-tools--c-button--m-selected--before--BorderRadius: var(--pf-global--BorderRadius--lg);
+  --pf-c-page__header-tools--c-button--m-selected--before--BorderRadius: var(--pf-global--BorderRadius--sm);
   --pf-c-page__header-tools--c-button--m-selected--c-notification-badge--m-unread--after--BorderColor: var(--pf-global--BackgroundColor--dark-200);
```
christiemolloy, line 58: Do you need this variable anymore ?

src/patternfly/demos/Page/page-template-header-tools-notification-badge.hbs
```diff
@@ -2,18 +2,18 @@
   {{#if page-header-tools--IsNotificationSelected}}
```
christiemolloy, line 2: Is this file being used any more ? you could remove it, and add it to the notification drawer specifically

mceledonia:
@mattnolting @mcarrano @mcoker The border alignment change looks fine to me, that was a design oversight more than anything. And to Matt's point, yes the new unread state (previously blue dot) is now the 2px blue border. Then the attention state is a third state triggered by critical alerts out of the box, but customizable to be triggered by any alert deemed worthy by the product leads.

src/patternfly/components/NotificationBadge/notification-badge.scss
mcoker: could `position: static` just apply to `.pf-c-notification-badge`?

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -39,17 +56,38 @@
...
   }
 
-  &.pf-m-read::after {
+  &.pf-m-read {
     --pf-c-notification-badge--after--BackgroundColor: var(--pf-c-notification-badge--m-read--after--BackgroundColor);
     --pf-c-notification-badge--after--BorderColor: var(--pf-c-notification-badge--m-read--after--BorderColor);
   }
+
+  &.pf-m-attention {
+    --pf-c-notification-badge--after--BackgroundColor: var(--pf-c-notification-badge--m-attention--after--BackgroundColor);
```
mcoker, line 84: should we add this and any other background color modifications to `::before`?

src/patternfly/components/NotificationBadge/notification-badge.scss
mattnolting: We could if the notification badge will only ever be used inside a button

src/patternfly/components/NotificationBadge/notification-badge.scss
mattnolting: Yeah, GH is collapsing the code https://github.com/patternfly/patternfly/blob/5c62a351df736a71755d51e9204135ce0d264781/src/patternfly/components/NotificationBadge/notification-badge.scss#L52

src/patternfly/components/NotificationBadge/notification-badge.scss
mcoker: Yeah that's what I meant with my original question. Is there a use case for it outside of a button? Seems like for this layout to work, it needs to be `static`, and placed in a non-`static` parent for the pseudo elements to work correctly.

src/patternfly/demos/Page/page-template-header-tools-notification-badge.hbs
mattnolting: It's included in `page-template-header-tools-elements.hbs` and conditionally rendered, I'd like to keep in case we have other demos that might use it

src/patternfly/components/Page/page.scss
```diff
@@ -258,35 +257,20 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
 .pf-c-page__header-tools-item {
   @include pf-hidden-visible(var(--pf-c-page__header-tools-item--Display));
 
-  .pf-c-button {
```
mcoker, line 261: This causes selected, non-notification badge buttons to appear behind the selected grey background.

Issue #3209, opened 2020-06-23
Select menu with filtering and favorites


mcarrano:
Need an example of  a Select menu that supports type-ahead filtering and favorities functionality equivalent to what we support for Application Launcher here: https://www.patternfly.org/v4/documentation/react/components/applicationlauncher#with-favorites-and-search

Design mockups below:

![image](https://user-images.githubusercontent.com/6619860/85412902-1a2a5780-b538-11ea-8d42-d5ae7aed406c.png)

We should also consider whether it makes sense to break Favorities out into a modular component that can be reused in multiple places (e.g. tables, etc).

cc @mcoker

mcoker:
@mcarrano 

1. What is the "Create project" text in this menu? Is that part of the feature, or just part of the design?
1. Should items retain the selected state check mark? Assuming so, what would that look like? 
1. Does this need to work with checkbox select?

mcarrano:
Good questions @mcoker .  @gdoyle1 can you help answer the first two?

As for the checkbox select, I have a hard time seeing the use case for favoriting checkbox items, but you never know. I would not include that for now unless it makes things easier in some way.

gdoyle1:
1. "Create project" is an action in this example - trying to say that these can be added into a dropdown so that some items can be favorited and some items can still act as actions
2. I thought we had discussed that we wouldn't allow for two icons in one item @mcarrano ? If that's not true, then I'd expect the checkmark to be to the right of the favorite star here

mcoker:
> "Create project" is an action in this example - trying to say that these can be added into a dropdown so that some items can be favorited and some items can still act as actions

Ah, this issue is to add the feature to the select menu. Just want to confirm we should add this to the select and not the dropdown for this task? @mcarrano 

> I thought we had discussed that we wouldn't allow for two icons in one item @mcarrano ? If that's not true, then I'd expect the checkmark to be to the right of the favorite star here

It seems like you would still want some indication of what item is selected? Here are a couple of screenshots:

<img width="653" alt="Screen Shot 2020-06-30 at 10 13 00 AM" src="https://user-images.githubusercontent.com/35148959/86143702-60446580-baba-11ea-99bc-3c0d728e53ca.png">
<img width="659" alt="Screen Shot 2020-06-30 at 10 11 50 AM" src="https://user-images.githubusercontent.com/35148959/86143704-61759280-baba-11ea-9472-d0b1de3c89fd.png">

mcarrano:
[...]

PR #3238, opened 2020-07-02
feat(select): added favorites

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3209

### DIFF
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -82,6 +82,7 @@
   --pf-c-select__menu-item--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-select__menu-item--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-select__menu-item--m-selected--PaddingRight: var(--pf-global--spacer--2xl);
+  --pf-c-select__menu-item--m-selected--m-link--PaddingRight: var(--pf-global--spacer--lg);
   --pf-c-select__menu-item--PaddingBottom: var(--pf-global--spacer--sm);
   --pf-c-select__menu-item--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-select__menu-item--FontSize: var(--pf-global--FontSize--md);
@@ -94,7 +95,6 @@
   --pf-c-select__menu-item--focus--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-select__menu-item--disabled--BackgroundColor: transparent;
   --pf-c-select__menu-item--m-link--Width: auto;
-  --pf-c-select__menu-item--m-link--PaddingRight: 0;
   --pf-c-select__menu-item--m-link--hover--BackgroundColor: transparent;
   --pf-c-select__menu-item--m-link--focus--BackgroundColor: transparent;
   --pf-c-select__menu-item--m-action--hover--BackgroundColor: transparent;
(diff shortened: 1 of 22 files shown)


christiemolloy:
Is this the intended spacing between the check and the star?
Screen Shot 2020-07-02 at 4 30 09 PM

mcoker:
@gdoyle1 @mcarrano wdyt re: [christie's question](https://github.com/patternfly/patternfly/pull/3238#issuecomment-653207432)?

For clarity, here's how that is setup. The green space is padding in the menu item to keep the text from overlapping the check icon. It uses the same spacing as a selected item without a favorite button.

Screen Shot 2020-07-02 at 3 33 22 PM

An alternative could be to move the checkmark all the way to the end of the item area, like this:

Screen Shot 2020-07-02 at 3 35 57 PM

Which looks like this without the outlines

Screen Shot 2020-07-02 at 3 36 13 PM

src/patternfly/components/Select/select.scss
```diff
@@ -88,10 +88,23 @@
   --pf-c-select__menu-item--FontWeight: var(--pf-global--FontWeight--normal);
   --pf-c-select__menu-item--LineHeight: var(--pf-global--LineHeight--md);
   --pf-c-select__menu-item--Color: var(--pf-global--Color--dark-100);
+  --pf-c-select__menu-item--Width: 100%;
   --pf-c-select__menu-item--disabled--Color: var(--pf-global--Color--dark-200);
   --pf-c-select__menu-item--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-select__menu-item--focus--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-select__menu-item--disabled--BackgroundColor: transparent;
+  --pf-c-select__menu-item--m-link--Width: auto;
+  --pf-c-select__menu-item--m-link--PaddingRight: 0;
```
christiemolloy, line 97: i dont see this used anywhere?

mcoker: removed

mcoker:
@christiemolloy updated the selected icon spacing when used with favorites.

Also realized that in the app launcher (and here), we set `.pf-m-favorite` on the parent wrapper, then change the `color` on `.pf-c-[select/app-launcher]__menu-item.pf-m-action` to gold. That works as long as the only `.pf-m-action` item is the favorite star. It breaks if we have multiple `.pf-m-action` items. To fix that, we need to identify an action as the favorite action.

I addressed it by adding `.pf-m-favorite-action` to the action item that has the star. In a perfect world, we could just use `.pf-m-favorite`, but I don't like that we would have the same class `.pf-m-favorite` on `.pf-c-select__menu-wrapper` (to indicate the whole item is favorited), and `.pf-m-favorite` on the `.pf-c-select__menu-item.pf-m-action` (to indicate which action is the favorite action), because of styling issues scoping styles for different logical modifiers because of a shared class. This is one of the reason I wonder if we could move toward traditional BEM modifiers so the class would be `.pf-c-select__menu-wrapper--favorite` and `.pf-c-select__menu-item--favorite`.

Also made a `select-menu-item.hbs` with a lot of logic in it since there is a lot of structural stuff that changes now whether an item has a description, or if items are used in a favorites menu. I didn't want to mess up examples and make them hard to work with/verify with copy/paste issues, so putting all the logic in the handlebars really helped with that. I'm only using it in the new favorites and description examples.

src/patternfly/components/Select/select.scss
```diff
@@ -82,24 +82,41 @@
...
   // Menu item icon
   --pf-c-select__menu-item-icon--Color: var(--pf-global--active-color--100);
   --pf-c-select__menu-item-icon--FontSize: var(--pf-global--icon--FontSize--sm);
   --pf-c-select__menu-item-icon--Right: var(--pf-global--spacer--md);
+  --pf-c-select__menu-item--m-link__menu-item-icon--Right: #{pf-size-prem(2px)};
   --pf-c-select__menu-item-icon--Top: 50%;
   --pf-c-select__menu-item-icon--TranslateY: -50%;
 
+  // Menu item action icon
+  --pf-c-select__menu-item-action-icon--MinHeight: calc(var(--pf-c-select__menu-item--FontSize) * var(--pf-c-select__menu-item--LineHeight));
```
christiemolloy, line 118: should the height be based off of `pf-c-select__menu-item-main` instead since that is now what the icon is adjacent to?

gdoyle1:
@mcoker @christiemolloy I think we should stick with the larger padding! Having the checkmark touch the focus area of the favorite icon seems a bit weird if you're interacting with that favorite icon

src/patternfly/components/Select/select.scss
```diff
@@ -380,44 +407,86 @@
...
     position: relative;
     background-color: var(--pf-c-select__menu-item--focus--BackgroundColor);
   }
 
-  &:disabled,
-  &.pf-m-disabled {
-    color: var(--pf-c-select__menu-item--disabled--Color);
-    pointer-events: none;
-    background-color: var(--pf-c-select__menu-item--disabled--BackgroundColor);
+  &.pf-m-link {
```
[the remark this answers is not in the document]
mcoker: That's for the main link part of the item. With the favorites feature, instead of having

```
li
  a/button.pf-c-select__menu-item
    // item contents
```

we have

```
li.pf-c-select__menu-wrapper
  a/button.pf-c-select__menu-item.pf-m-linik // menu item contents
  button.pf-c-select__menu-item.pf-m-action // favorites
```

christiemolloy:
still need to update the padding as per gina's comment (so that the check doesn't touch the side)

Screen Shot 2020-07-09 at 2 00 14 PM

PR #3241, opened 2020-07-02
refactor: udpated text max-length

### LINKED ISSUE #2954: Truncation length
Maybe outside the scope of this issue, but what do you think about changing `--pf-c-chip__text--MaxWidth` to a character width, say `20ch` and/or making that a global variable? 

Reason being that if the font-size of `pf-c-chip__text` is increased, the string length doesn't adjust to that change and results in this. 
Screen Shot 2020-04-13 at 4 11 24 PM

Where a `ch` value would result in this
Screen Shot 2020-04-13 at 4 11 37 PM

_Originally posted by @mattnolting in https://github.com/patternfly/patternfly/pull/2941_

mattnolting:
closes #2954 

Before update:
Screen Shot 2020-07-02 at 10 53 46 AM

After update:
Screen Shot 2020-07-02 at 11 01 37 AM

### DIFF
--- a/src/patternfly/components/Chip/chip.scss
+++ b/src/patternfly/components/Chip/chip.scss
@@ -15,7 +15,7 @@
   // chip text
   --pf-c-chip__text--FontSize: var(--pf-global--FontSize--xs);
   --pf-c-chip__text--Color: var(--pf-global--Color--100);
-  --pf-c-chip__text--MaxWidth: #{pf-size-prem(120px)};
+  --pf-c-chip__text--MaxWidth: 16ch;
 
   // chip button
   --pf-c-chip__c-button--PaddingTop: var(--pf-global--spacer--xs);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Chip/chip.scss
```diff
@@ -15,7 +15,7 @@
   // chip text
   --pf-c-chip__text--FontSize: var(--pf-global--FontSize--xs);
   --pf-c-chip__text--Color: var(--pf-global--Color--100);
-  --pf-c-chip__text--MaxWidth: #{pf-size-prem(120px)};
+  --pf-c-chip__text--MaxWidth: 16ch;
```
mcoker, line 18: nice! wdyt about making this update for the chip group label text, too?

https://github.com/patternfly/patternfly/blob/376703bdfbaf9ddc70f7d08e37f695079dba1c38/src/patternfly/components/ChipGroup/chip-group.scss#L19

mattnolting:
@mcoker updated

PR #3249, opened 2020-07-06
fix(chipgroup): update removable action icon

### LINKED ISSUE #3247: Update Chip component to match whats in the design kit
The close button at the end of a chip group should be changed in order to reflect this design in the sketch kit:

From @mceledonia "I think what’s in the design kit is the most ideal, we wanted the circle-x to get changed in the chip because the actual “x” was too small. But with the chip group, the circle-x can be a little bigger, plus it makes sense that it would be different than the chip X as to denote a different action (clear all)"

christiemolloy:
closes #3247 

@maryshak1996 @mceledonia I noticed the removable action doesn't have the same spacing as in the design kit. Do you know the actual px for the padding ?

Screen Shot 2020-07-06 at 11 40 03 AM


maryshak1996:
@christiemolloy I think that the spacing that you have works well (I thought that what was in the design kit was quite tight anyway) -- @mceledonia what do you think?

maryshak1996:
@christiemolloy for some reason, I am still unable to get the surge preview working -- I am just getting this failure walrus... 
Screen Shot 2020-07-08 at 12 15 08 PM

PR #3250, opened 2020-07-06
feat(nav): adds button variation to nav-item

### LINKED ISSUE #3212: Expandable nav item should be a button and not a link
Related: https://github.com/patternfly/patternfly-react/issues/4377

When an expandable nav item doesn't serve as a link and is only a subnav toggle, the element should be a `<button>` instead of an `<a>`.

jenny-s51:
Closes #3212. Once merged, I'll make the corresponding change in https://github.com/patternfly/patternfly-react/issues/4377.


jenny-s51:
The focus state previously looked like this @mcarrano  Screen Shot 2020-07-08 at 1 09 46 PM ... happy to make whatever changes are necessary.

jenny-s51:
Upon testing both `master` and this branch in both Chrome and Firefox, the focus state appears to look the same in each browser, and it looks like this PR hasn't changed the appearance of the focus state as you said @mcarrano . However there seems to be a slight difference in the appearance of the focus state between the browsers. The focused state in Chrome appears in Christie's screenshot above, while in Firefox it seems to look like this:

Screen Shot 2020-07-08 at 1 09 46 PM

If this discrepancy needs to be addressed at some point, I would be happy to open another issue for it - just let me know. In the meantime, could we merge this in?

PR #3253, opened 2020-07-07
feat(flex-and-grid): added ordering to layouts

### LINKED ISSUE #3245: Feat: add ordering to flex and grid layouts
Bootstrap users have the ability to order `flex` and `grid` items. Adding ordering to our layouts will provide the expected support and encourage users to transition to PF.

Bootstrap does it like this: https://getbootstrap.com/docs/4.5/layout/grid/#order-classes

mattnolting:
closes #3245

### DIFF
--- a/src/patternfly/layouts/Flex/examples/Flex.md
+++ b/src/patternfly/layouts/Flex/examples/Flex.md
@@ -659,7 +659,107 @@ The CSS approach, by keeping specificity low on base class properties and resett
   {{/l-flex-item}}
 {{/l-flex}}
 ```
+
+### Ordering
+
+Ordering - `.pf-m-order-[1,12]{-on-[breakpoint]}`. Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.pf-l-flex__item</code>s. Spacing may need to be managed based on how items are ordered.
+
+```hbs title=First-last-ordering
+{{#> l-flex}}
+  {{#> l-flex-item l-flex-item--modifier="pf-m-order-last pf-m-spacer-none"}}
+    Last item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item l-flex-item--modifier="pf-m-order-first pf-m-spacer-md"}}
+    First item
+  {{/l-flex-item}}
+{{/l-flex}}
+```
+
+```hbs title=Responsive-first-last-ordering
+{{#> l-flex}}
+  {{#> l-flex-item l-flex-item--modifier="pf-m-order-last-on-lg pf-m-spacer-none-on-lg"}}
+    Last item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item l-flex-item--modifier="pf-m-order-first-on-lg pf-m-spacer-md-on-lg"}}
+    First item
+  {{/l-flex-item}}
+{{/l-flex}}
+```
+
+```hbs title=Ordering
+{{#> l-flex}}
+  {{#> l-flex l-flex--modifier="pf-m-order-2 pf-m-spacer-none"}}
+    {{#> l-flex-item l-flex-item--modifier="pf-m-order-4 pf-m-spacer-none"}}
+      Set 1, Order 4
+    {{/l-flex-item}}
+    {{#> l-flex-item l-flex-item--modifier="pf-m-order-2"}}
+      Set 1, Order 2
+    {{/l-flex-item}}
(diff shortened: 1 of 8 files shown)


src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -707,8 +809,12 @@ The CSS approach, by keeping specificity low on base class properties and resett
 | `.pf-m-align-content-space-around{-on-[breakpoint]}` | `.pf-l-flex` | Modifies the flex layout align-content property to space-around. |
 | `.pf-m-align-left{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Resets the flex layout element margin-left property to 0. |
 | `.pf-m-align-right{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the flex layout element margin-left property to auto. |
+| `.pf-m-order[0-12]{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the order of the flex layout element. |
```
mcoker, line 812:
```suggestion
| `.pf-m-order-[0-12]{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the order of the flex layout element. |
```

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -244,6 +244,9 @@ $pf-global--arrow--width-lg: pf-font-prem(25px);
 $pf-global--target-size--MinWidth: 44px !default;
 $pf-global--target-size--MinHeight: 44px !default;
 
+// Order limit
+$pf-global--order--limit: 12;
```
mcoker, line 248: Why do we need this as a global variable?

src/patternfly/layouts/Grid/grid-order.scss
```diff
@@ -0,0 +1,25 @@
+$pf-l-grid--order--limit: $pf-global--order--limit;
```
mcoker, line 1: It seems like the grid could be built from like `$pf-l-grid--columns/tracks: 12;` and both the grid and ordering could reference that, so that the order limit always matches the number of tracks - versus a var for one and a hard-coded value for the other?

src/patternfly/layouts/Grid/examples/Grid.md
```diff
@@ -293,3 +396,6 @@ The grid layout is based on CSS Grid’s two-dimensional system of columns and r
 | `.pf-m-all-{1-12}-col{-on-[breakpoint]}` | `.pf-l-grid` | Defines grid item size on grid container. |
 | `.pf-m-{1-12}-col{-on-[breakpoint]}` | `.pf-l-grid__item` | Defines grid item size.  Although not required, they are strongly suggested. If not used, grid item will default to 12 col. |
 | `.pf-m-{2-x}-row{-on-[breakpoint]}` | `.pf-l-grid__item` | Defines grid item row span.  For row spans to function correctly, the value of of the current row plus the grid items to span must be equal to or less than 12. Example: .pf-m-8-col.pf-m-2-row + .pf-m-4-col + .pf-m-4-col. There is no limit to number of spanned rows. |
+| `.pf-m-order[0-12]{-on-[breakpoint]}` | `.pf-l-grid > .pf-l-grid`, `.pf-l-grid__item` | Modifies the order of the grid layout element. |
```
mcoker, line 399:
```suggestion
| `.pf-m-order-[0-12]{-on-[breakpoint]}` | `.pf-l-grid > .pf-l-grid`, `.pf-l-grid__item` | Modifies the order of the grid layout element. |
```

src/patternfly/layouts/Flex/flex-order.scss
```diff
@@ -0,0 +1,25 @@
+$pf-l-flex--order--limit: $pf-global--order--limit;
```
mcoker, line 1: It seems like `pf-m-order-[0-12]` here is a bit arbitrary, since it's not based off of a 12 column/row layout. I can see adding it for consistency to match the grid layout. Just thinking forward, how might someone set `order: 42`? Would that have any impact on how we build this today? I don't think we would build out classnames for that.

One approach could be to have a fallback variable system at different breakpoints where we read `--order{-[breakpoint]}: n`, and we pass that directly to an `order` property at that breakpoint. That would work in place of the `pf-m-order` classes, and works for any `order` value. And the react component could take like `breakpoints={'md': 1, 'lg': -1}` and create `setProperty('--order-md', '1').setProperty('--order-lg', '-1')` (or however they do that).

Here's a demo of how that might work https://codepen.io/mcoker/pen/GRoeLJo

PR #3229, opened 2020-06-30
feat(tile): add tile component

christiemolloy:
closes #3100

PR #3274, opened 2020-07-10
fix(tile): remove support for imgs

christiemolloy:
As a follow up to #3229 this PR removes support for images in the tile component and only supports SVG's which take on the font-size property set on the tile-icon.


src/patternfly/components/Tile/examples/Tile.md
```diff
@@ -164,7 +164,7 @@ import './Tile.css'
 {{#> tile}}
   {{#> tile-header tile-header--modifier="pf-m-stacked"}}
     {{#> tile-icon}}
-      <img src="/assets/images/pf-logo-small.svg" alt="PatternFly logo">
+      {{#> tile-img-pf}}{{/tile-img-pf}}
```
mattnolting, line 167: These can all be

```suggestion
      {{> tile-img-pf}}
```

src/patternfly/components/Tile/tile.scss
```diff
@@ -24,20 +27,18 @@
   --pf-c-tile__icon--Color: var(--pf-global--Color--200);
   --pf-c-tile--hover__icon--Color: var(--pf-global--primary-color--100);
   --pf-c-tile--m-selected__icon--Color: var(--pf-global--primary-color--100);
-  --pf-c-tile--m-disabled__icon--Color: var(--pf-global--disabled-color--100);
```
mcoker, line 27: We should leave this in. If someone sets `--pf-c-tile__icon--Color: red;`, then it will still be red if it's disabled since we only set the icon color for stacked icons.

mcoker:
Just an idea, but this might be a time to try something like

```
.pf-c-tile__icon {
  color: var(--pf-c-tile__icon--Color--state, var(--pf-c-tile__icon--Color--type, var(--pf-c-tile__icon--Color)));
}
```

where `--pf-c-tile__icon--Color` is the base/default color.

Then we have a theme var for the tile "type" (eg, stacked), and set `.pf-c-tile__header.pf-m-stacked { --pf-c-tile__icon--Color--type: var(--pf-c-tile__header--m-stacked__icon--Color); }`

Then for hover/focus/selected/disabled, we set `.pf-c-tile__header:hover { --pf-c-tile__icon--Color--state: var(--pf-c-tile--hover__icon--Color); }`, `.pf-c-tile__header.pf-m-selected { --pf-c-tile__icon--Color--state: var(--pf-c-tile--m-selected__icon--Color); }`, etc.

That way we're only setting 1 var on hover, focus, etc, for the icon color versus needed to set 2 separate vars. That's assuming we don't think we would separate out the icon hover/focus/selected/disabled color between normal and stacked variations, which it doesn't seem like we would. Easier for us and the user to customize.

We haven't done that before, but I'm curious what you think @christiemolloy @mattnolting

src/patternfly/components/Tile/tile.scss
mcoker: Also need to add back so the stacked icon color is disabled

`--pf-c-tile__header--m-stacked__icon--Color: var(--pf-global--disabled-color--100);`

src/patternfly/components/Tile/tile.scss
```diff
@@ -25,19 +28,18 @@
...
   --pf-c-tile--m-disabled__icon--Color: var(--pf-global--disabled-color--100);
+  --pf-c-tile--focus__icon--Color: var(--pf-global--primary-color--100);
 
   // Stacked variatiion
   --pf-c-tile__header--m-stacked__icon--MarginBottom: var(--pf-global--spacer--xs);
   --pf-c-tile__header--m-stacked__icon--FontSize: var(--pf-global--icon--FontSize--lg);
   --pf-c-tile__header--m-stacked__icon--Color: var(--pf-global--Color--100);
   --pf-c-tile--hover__header--m-stacked__icon--Color: var(--pf-global--primary-color--100);
   --pf-c-tile--m-selected__header--m-stacked__icon--Color: var(--pf-global--primary-color--100);
-  --pf-c-tile--m-disabled__header--m-stacked__icon--Color: var(--pf-global--disabled-color--100);
```
mcoker, line 35: Looks like you need to add this back. It's used but no longer defined:

https://github.com/patternfly/patternfly/blob/baa6bcbdef670d447e3831012d1c5215025c8827/src/patternfly/components/Tile/tile.scss#L101

PR #3286, opened 2020-07-13
feat(tab-background-color): added tab background modifiers

### LINKED ISSUE #3150: Box tabs enhancement for screens with gray content backgrounds
Can we have an option to reverse the box tab colors for those using a gray background in their content area? The new box tabs give a gray fill to the unselected tabs, which makes them look like they are selected when the content area is below is also gray. 

Screen Shot 2020-06-01 at 10 14 05 PM

I think having an option to reverse the colors would connect the active tab to the content area much better when gray bkgs are in use (See Apicurio example below). 

Screen Shot 2020-05-29 at 11 00 33 AM

mattnolting:
closes #3150

### DIFF
--- a/src/patternfly/components/Tabs/examples/Tabs.md
+++ b/src/patternfly/components/Tabs/examples/Tabs.md
@@ -4,6 +4,8 @@ section: components
 cssPrefix: pf-c-tabs
 ---
 
+import './Tabs.css'
+
 ## Examples
 
 ### Default
@@ -60,6 +62,14 @@ cssPrefix: pf-c-tabs
 {{#> tabs tabs--id="box-vertical-example" tabs--modifier="pf-m-box pf-m-vertical"}}
   {{> __tabs-list __tabs-list--NoScrollButtons="true"}}
 {{/tabs}}
+```
+
+### Box tabs alt color scheme
+```hbs
+{{#> tabs tabs--modifier="pf-m-box pf-m-alt-color-scheme" tabs--id="Box-tabs-alt-color-scheme"}}
+  {{> __tabs-list}}
+{{/tabs}}
+<div className="tabs-example-block tabs-example-block-m-alt-color-scheme"></div>
 ```
 
 ### Inset
@@ -194,3 +204,4 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 | `.pf-m-fill` | `.pf-c-tabs` | Modifies the tabs to fill the available space. |
 | `.pf-m-current` | `.pf-c-tabs__item` | Indicates that a tab item is currently selected. |
 | `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[md, lg, xl, 2xl]}` | `.pf-c-tabs` | Modifies the tabs component padding/inset to visually match padding of other adjacent components. |
+| `. pf-m-alt-color-scheme` | `.pf-c-tabs` | Modifies the tabs component tab background colors. |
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -177,3 +186,4 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 | `.pf-m-fill` | `.pf-c-tabs` | Modifies the tabs to fill the available space. |
 | `.pf-m-current` | `.pf-c-tabs__item` | Indicates that a tab item is currently selected. |
 | `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[md, lg, xl, 2xl]}` | `.pf-c-tabs` | Modifies the tabs component padding/inset to visually match padding of other adjacent components. |
+| `. pf-m-alt-color-scheme` | `.pf-c-tabs` | Modifies the tabs component tab background colors. |
```
mcoker, line 189: @lboehling @mcarrano we're calling this variation `pf-m-alt-color-scheme`. Is that a good name or is there a more meaningful name we could use? Also thinking forward, would the name scale well if we introduced another color variation for use in a different context?

mcoker, line 189:
```suggestion
| `.pf-m-alt-color-scheme` | `.pf-c-tabs` | Modifies the tabs component tab background colors. |
```

mcarrano: Not sure @mcoker . The purpose of this is to just swap the white for gray, correct? If so, should be just call this something more like 'inverted'?

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -91,6 +96,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   padding-right: var(--pf-c-tabs--inset);
   padding-left: var(--pf-c-tabs--inset);
   overflow: hidden;
+  background-color: var(--pf-c-tabs--BackgroundColor);
```
mcoker, line 99: is this used for anything? it's transparent by default.

mcoker:
@mattnolting looks like there is a conflict.

src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -194,3 +204,4 @@ Whenever a list of tabs is unique on the current page, it can be used in a `<nav
 | `.pf-m-fill` | `.pf-c-tabs` | Modifies the tabs to fill the available space. |
 | `.pf-m-current` | `.pf-c-tabs__item` | Indicates that a tab item is currently selected. |
 | `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[md, lg, xl, 2xl]}` | `.pf-c-tabs` | Modifies the tabs component padding/inset to visually match padding of other adjacent components. |
+| `.pf-m-current-tab-background--light-300` | `.pf-c-tabs` | Modifies the tabs component tab background colors. |
```
christiemolloy, line 207: what if this was called pf-m-inverse instead ? since it just swaps the colors around ? I would also update the description

mcoker: I think @mceledonia had recommended this naming. The issue is that it doesn't scale if we offer a different variation, but it could work if we don't think this color change is part of a theme where we might have other color/theme variations.

PR #3295, opened 2020-07-15
feat(togglebuttongroup): add component

### LINKED ISSUE #2625: Introduce Toogle Button Group
Introduce a new component called Toggle Button Group. This will be a groups of buttons that allow for selecting one items among a list of options (similar to a group of radio buttons). See design mockups below.

christiemolloy:
fix #2625


src/patternfly/components/ToggleButtonGroup/examples/ToggleButtonGroup.md
```diff
@@ -0,0 +1,190 @@
+---
+title: Toggle button group
```
mcoker, line 2: need to add `beta: true`

src/patternfly/components/ToggleGroup/examples/ToggleGroup.md
```diff
@@ -0,0 +1,158 @@
...
+## Documentation
+### Overview
+
+
+### Accessibility
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `aria-label="[button label text]"` | `.pf-c-toggle-group__button` | Provides an accessible name for the button when an icon is used instead of text. **Required when icon is used with no supporting text** |
+| `disabled` | `.pf-c-toggle-group__button` | When a button element is used, indicates that it is unavailable and removes it from keyboard focus. **Required when button is disabled** |
+| `aria-disabled="true"` | `button.pf-c-toggle-group__button` | When a button element is used, indicates that it is unavailable but does not prevent keyboard or hover interactions. Used when a disabled button provides interactive elements like a tooltip. |
```
mcoker, line 149: why do you need `aria-disabled`? shouldn't be necessary with `disabled`

src/patternfly/components/ToggleGroup/toggle-group.scss
```diff
@@ -0,0 +1,70 @@
...
+  --pf-c-toggle-group__button--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-toggle-group__button--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-toggle-group__button--first-child--BorderTopLeftRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-toggle-group__button--first-child--BorderBottomLeftRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-toggle-group__button--last-child--BorderTopRightRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-toggle-group__button--last-child--BorderBottomRightRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-toggle-group__button--BackgroundColor: var(--pf-global--BackgroundColor--200);
+  --pf-c-toggle-group__button--Color: var(--pf-global--Color--100);
+  --pf-c-toggle-group__button--m-light--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-toggle-group__button--hover--BackgroundColor: var(--pf-global--disabled-color--200);
```
mcoker, line 14: Hmm... I don't think we should use `--disabled-color--200` here. Can we confirm we can't use one of the existing background colors? Maybe we need a new global background color var? cc @mceledonia, this ties into the theming. If someone were to customize the `--disabled-color--200` var, I don't think we would want to change this button state color.

https://github.com/patternfly/patternfly/blob/2167795ab3c6d711837f96c1f76e870fd041e3ec/src/patternfly/sass-utilities/scss-variables.scss#L25-L41

src/patternfly/components/ToggleGroup/toggle-group.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  &:disabled,
+  &.pf-m-disabled {
+    pointer-events: none;
+  }
+
+  &:disabled,
+  &.pf-m-disabled,
+  &.pf-m-aria-disabled {
```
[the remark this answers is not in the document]
mcoker: I think `aria-disabled` is to support the disabled, yet focusable button - https://www.patternfly.org/v4/documentation/core/components/button#aria-disabled

mcoker:
@christiemolloy overall I think this looks great. A few things - also for @mceledonia @mcarrano :

1. Should these buttons not match our existing button dimensions?
1. If we're going to support `pf-m-aria-disabled` is there value in trying to use the button component here? I'm curious if we should have a toggle button that supports the selected states, then use it in a toggle button group component?
1. Is it valid to only have a single button in a group, to toggle something on/off?
1. Is it valid to have multiple selections at once? <sup>1</sup>
  * Might we also want to support multiple groups? <sup>2</sup>
1. I wonder if we should have a container for the buttons: `toggle-group > toggle-group__item > toggle-group__button`

<sup>1</sup> This is an example from material - Screen Shot 2020-07-27 at 3 12 21 PM
 
<sup>2</sup> bootstrap 
Screen Shot 2020-07-27 at 3 17 35 PM

mcarrano:
@mcoker Good questions about the intended interactions for this.

> Should these buttons not match our existing button dimensions?
I'll defer to @mceledonia 
> If we're going to support pf-m-aria-disabled is there value in trying to use the button component here? I'm curious if we should have a toggle button that supports the selected states, then use it in a toggle button group component? 
I can see the case for that. Maybe this depends on the answer to the above. But the styling is different than a standard button, isn't it?

> Is it valid to only have a single button in a group, to toggle something on/off?
Hmmm... I don't think that's how we intended it. But if we wanted to allow for multiple selections, like as for a checkbox group, then maybe. It becomes like a graphical checkbox. 

> Is it valid to have multiple selections at once? 
Again, we did not think about it this way, but I suppose that we could think about this as a more generic tool button. Thoughts @mceledonia ?

mcoker:
> I can see the case for that. Maybe this depends on the answer to the above. But the styling is different than a standard button, isn't it?

@mcarrano it's a little bit different. Similar to how the form control button is different - different background and hover styles. The spacing is currently a bit different, too, but I don't think that would be a problem if we wanted to keep that and have this be a variation of the button component.

mceledonia:
@mcoker 

Re: Dimensions - These buttons have 14px font size, with 8px top/bottom and 16px right/left padding.
Re: Multiple selections - I feel it would make sense to offer this as an option, I can imagine certain use cases where that would be required.
Re: Disabled - We can make that styling match our standard button disabled styling, but the spacial/typographic dimensions are different as stated above.

src/patternfly/components/ToggleGroup/examples/ToggleGroup.md
christiemolloy: I was thinking that the buttons in the toggle group would be the same as the buttons in the button component - but i guess i dont need it because they have more of a specific use case and wouldnt need an interactive element like a tooltip

christiemolloy:
@mceledonia in regards to this question - https://github.com/patternfly/patternfly/pull/3295#discussion_r461134941 i can make a new backgroundColor var for `pf-color-black-300`?

@mcarrano @mceledonia what do you think about making these buttons a variation of the button component. Similar to how we have `pf-m-control` as a variation which we use in clipboardcopy and inputgroup. the question would be whether this button style can be used outside of the toggle group component, in another component, or whether this button style is specific to the toggle group component? 
This is also based off of @mcoker mentioning whether we would need to support `pf-m-aria-disabled` in this case, which is for when buttons are interactive and need to be focused even when they're disabled.

mcarrano:
@christiemolloy My only concern with this suggestion of making this just a button variant is that it might not make sense for the React component. These buttons should behave like a radio group and not just be an alternate styling of a button.

src/patternfly/components/ToggleGroup/examples/ToggleGroup.md
```diff
@@ -0,0 +1,157 @@
...
+| `aria-label="[button label text]"` | `.pf-c-toggle-group__button` | Provides an accessible name for the button when an icon is used instead of text. **Required when icon is used with no supporting text** |
+| `disabled` | `.pf-c-toggle-group__button` | When a button element is used, indicates that it is unavailable and removes it from keyboard focus. **Required when button is disabled** |
+
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-toggle-group` | `<div>` | Initiates the toggle button group. **Required** |
+| `.pf-c-toggle-group__button` | `<button>` | Initiates the toggle button group button. **Required** |
+| `.pf-m-selected` | `.pf-c-toggle-group__button` | Modifies the toggle button group button for the selected state. |
+| `.pf-m-disabled` | `.pf-c-toggle-group__button` | Modifies the toggle button group button for the disabled state. |
```
mcoker, line 156: if these are always buttons, you shouldn't need this. If we want to leave this modifier (so that a link can be used), we should add `aria-disabled` back

mcoker:
> Re: Dimensions - These buttons have 14px font size, with 8px top/bottom and 16px right/left padding.

@christiemolloy should the buttons be 14px?

> - I feel it would make sense to offer this as an option, I can imagine certain use cases where that would be required.

Should we create an example of that so it's clear that's an option?

src/patternfly/components/ToggleGroup/examples/ToggleGroup.md
```diff
@@ -0,0 +1,157 @@
...
+  {{#> toggle-group-button toggle-group-button--modifier="pf-m-light"}}
+    Option 1
+  {{/toggle-group-button}}
+  {{> divider divider--type="div" divider--modifier="pf-m-vertical"}}
+  {{#> toggle-group-button toggle-group-button--modifier="pf-m-light"}}
+    Option 2
+  {{/toggle-group-button}}
+  {{> divider divider--type="div" divider--modifier="pf-m-vertical"}}
+  {{#> toggle-group-button toggle-group-button--modifier="pf-m-light pf-m-selected"}}
+    Option 3
```
mcoker, line 118: I'm assuming you could also put an icon in a button? Since the text and icons have no BEM element, that will require updates to handle properly if we want to manage that spacing.

src/patternfly/components/ToggleGroup/examples/ToggleGroup.md
christiemolloy: @mcarrano mentioned above that we're not supporting links in this component, so i think its safe to remove.

Issue #3297, opened 2020-07-16
Dark theme POC


mcoker:
Using the following data, create a POC to evaluate the theming work further.

UI Element | Which is this | Changes to this
-- | -- | --
  |   |  
**Backgrounds (fills)** |   |  
All white backgrounds |   | 1B1E24 (700)
All light gray backgrounds |   | 12141A (800)
Left nav background (for now) |   | 1B1E24 (700)
Masthead background |   | 05050C (900)
  |   |  
**Borders (outlines)** |   |  
Divider lines (all) |   | 292D32 (600)
  |   |  
**Main colors (primary, status)** |   |  
Primary blue (all instances) |   | 2B9AF3 (pf-blue-300)
Success green |   | 5BA352 (pf-green-400)
Warning yellow |   | F4C145 (pf-gold-300)
Very bad red |   | F64747 (NEW)
  |   |  
**Form fields** |   |  
Fill |   | 292D32 (600)
Border |   | Remove (or ^)
  |   |  
**Text (including icons)** |   |  
All black text (primary) |   | FFFFFF (pf-white)
All gray text (secondary) |   | A9ADAF (200)
All blue text |   | 2B9AF3 (pf-blue-300)
  |   |  
**New Gray Pallete** |   |  
Gray 50 | E7EBEB |  
Gray 100 | D3D6D7 |  
Gray 200 | A9ADAF |  
Gray 300 | 808487 |  
Gray 400 | 565B5F |  
Gray 500 | 42474B |  
Gray 600 | 292D32 |  
Gray 700 | 1B1E24 |  
Gray 800 | 12141A |  
Gray 900 | 05050C |

PR #3298, opened 2020-07-16
chore(global): dark theme poc

mcoker:
POC - do not merge.

fixes https://github.com/patternfly/patternfly/issues/3297

Everything was able to be updated via global vars except for

* sidebar background color
  * no global var change for `--pf-c-page__sidebar--BackgroundColor: var(--pf-global--BackgroundColor--dark-300);`
* masthead background color
  * no global var change for `--pf-c-page__header--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);`
* form control border
  * we're removing the borders in dark theme, which doesn't go with the change of `--pf-global--BorderColor--100` to `pf-global--palette--black-600`. Technically we don't have to change/remove the border colors though since the borders are now `black-600` which matches the custom background color for form control elements, though if these elements should not have a border in dark theme, we should remove them.
* form control background-color
  * doesn't go with the change of `--pf-global--BackgroundColor--100` to `--pf-global--palette--black-700`

### DIFF
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -1,7 +1,5 @@
 // Set common reset styles for patternfly components
 [class*="pf-c-"] {
-  color: var(--pf-global--Color--100);
-
   &,
   &::before,
   &::after {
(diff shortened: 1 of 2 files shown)


mcoker:
@mcarrano made an adjustment for the text so that should work better now.

PR #3301, opened 2020-07-17
feat(docs): add compiled example html to dist

redallen:
Org is currently compiling Core's handlebars files itself. This is problematic since Core and Org need to keep their build systems in sync and it takes ~15 seconds of startup time to compile the HTML for Org devs. It also limits Core to using only Handlebars.

This PR creates a `dist/docs` folder that contains MD files with compiled example HTML in favor of the current `dist/site` folder which includes the MD files with HBS templates. Once Gatsby is removed, the `dist/site` folder can be removed for a smaller NPM package.


redallen:
The CSS size lint failing can be ignored and is fixed in #3302

PR #3303, opened 2020-07-17
feat(treeview): Add tree view component

### LINKED ISSUE #3206: Introduce Tree menu
Introduce a Tree menu component per the designs outlined here: https://marvelapp.com/8cdg211/screen/68891042

christiemolloy:
closes #3206

### DIFF
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -7,7 +7,7 @@ cssPrefix: pf-c-tree-view
 ## Examples
 ```hbs title=Default
 {{#> tree-view}}
-    {{#> tree-view-list}}
+    {{#> tree-view-list tree-view-list--attribute='role="tree"'}}
         {{#> tree-view-list-item}}
             {{#> tree-view-link tree-view-link--modifier="pf-m-expandable pf-m-expanded"}}
                 {{> tree-view-link-toggle}}
@@ -668,26 +668,30 @@ cssPrefix: pf-c-tree-view
 {{#> tree-view}}
     {{#> tree-view-list}}
         {{#> tree-view-list-item}}
-            {{#> tree-view-link tree-view-link--modifier="pf-m-expandable pf-m-expanded"}}
-                {{> tree-view-link-toggle}}
-                {{#> tree-view-link-text}}
-                    Application launcher
-                {{/tree-view-link-text}}
-                {{#> tree-view-link-action}}
+            {{#> tree-view-link-wrapper}}
+                {{#> tree-view-link tree-view-link--modifier="pf-m-expandable pf-m-expanded"}}
+                    {{> tree-view-link-toggle}}
+                    {{#> tree-view-link-text}}
+                        Application launcher
+                    {{/tree-view-link-text}}
+                {{/tree-view-link}}
+                {{#> tree-view-action}}
                     <i class="fas fa-ellipsis-v" aria-hidden="true"></i>
-                {{/tree-view-link-action}}
-            {{/tree-view-link}}
+                {{/tree-view-action}}
+            {{/tree-view-link-wrapper}}
             {{#> tree-view-list tree-view-list--modifier="pf-m-nested-level-1"}}
                 {{#> tree-view-list-item}}
-                    {{#> tree-view-link tree-view-link--modifier="pf-m-first pf-m-expandable pf-m-expanded"}}
-                        {{> tree-view-link-toggle}}
-                        {{#> tree-view-link-text}}
-                            Application 1
-                        {{/tree-view-link-text}}
-                        {{#> tree-view-link-action}}
+                    {{#> tree-view-link-wrapper}}
+                        {{#> tree-view-link tree-view-link--modifier="pf-m-first pf-m-expandable pf-m-expanded"}}
+                            {{> tree-view-link-toggle}}
+                            {{#> tree-view-link-text}}
+                                Application 1
+                            {{/tree-view-link-text}}
+                        {{/tree-view-link}}
+                        {{#> tree-view-action}}
                             <i class="fas fa-clipboard" aria-hidden="true"></i>
-                        {{/tree-view-link-action}}
-                    {{/tree-view-link}}
+                        {{/tree-view-action}}
+                    {{/tree-view-link-wrapper}}
                     {{#> tree-view-list tree-view-list--modifier="pf-m-nested-level-2"}}
                         {{#> tree-view-list-item}}
                             {{#> tree-view-link}}
@@ -706,15 +710,17 @@ cssPrefix: pf-c-tree-view
                     {{/tree-view-list}}
                 {{/tree-view-list-item}}
                 {{#> tree-view-list-item}}
-                    {{#> tree-view-link tree-view-link--modifier="pf-m-expandable pf-m-expanded"}}
-                        {{> tree-view-link-toggle}}
-                        {{#> tree-view-link-text}}
-                            Application 2
-                        {{/tree-view-link-text}}
-                        {{#> tree-view-link-action}}
+                    {{#> tree-view-link-wrapper}}
+                        {{#> tree-view-link tree-view-link--modifier="pf-m-expandable pf-m-expanded"}}
+                            {{> tree-view-link-toggle}}
+                            {{#> tree-view-link-text}}
+                                Application 2
+                            {{/tree-view-link-text}}
+                        {{/tree-view-link}}
+                        {{#> tree-view-action}}
                             <i class="fas fa-bars" aria-hidden="true"></i>
-                        {{/tree-view-link-action}}
-                    {{/tree-view-link}}
+                        {{/tree-view-action}}
+                    {{/tree-view-link-wrapper}}
                     {{#> tree-view-list tree-view-list--modifier="pf-m-nested-level-2"}}
                         {{#> tree-view-list-item}}
                             {{#> tree-view-link tree-view-link--modifier="pf-m-expandable"}}
@@ -733,15 +739,17 @@ cssPrefix: pf-c-tree-view
                             {{/tree-view-link}}
                             {{#> tree-view-list tree-view-list--modifier="pf-m-nested-level-3"}}
                                 {{#> tree-view-list-item}}
-                                    {{#> tree-view-link tree-view-link--modifier="pf-m-expandable"}}
-                                        {{> tree-view-link-toggle}}
-                                        {{#> tree-view-link-text}}
-                                            Loading app 1
-                                        {{/tree-view-link-text}}
-                                        {{#> tree-view-link-action}}
+                                    {{#> tree-view-link-wrapper}}
+                                        {{#> tree-view-link  tree-view-link--modifier="pf-m-expandable"}}
+                                            {{> tree-view-link-toggle}}
+                                            {{#> tree-view-link-text}}
+                                                Loading app 1
+                                            {{/tree-view-link-text}}
+                                        {{/tree-view-link}}
+                                        {{#> tree-view-action tree-view-link-action--modifier="pf-m-favorite"}}
                                             <i class="fas fa-star" aria-hidden="true"></i>
-                                        {{/tree-view-link-action}}
-                                    {{/tree-view-link}}
+                                        {{/tree-view-action}}
+                                    {{/tree-view-link-wrapper}}
                                 {{/tree-view-list-item}}
                                 {{#> tree-view-list-item}}
                                     {{#> tree-view-link}}
@@ -780,15 +788,17 @@ cssPrefix: pf-c-tree-view
             {{/tree-view-link}}
         {{/tree-view-list-item}}
         {{#> tree-view-list-item}}
-            {{#> tree-view-link tree-view-link--modifier="pf-m-expandable"}}
-                {{> tree-view-link-toggle}}
-                {{#> tree-view-link-text}}
-                    This is a really really really long folder name that overflows from the width of the container.
-                {{/tree-view-link-text}}
-                {{#> tree-view-link-action}}
+            {{#> tree-view-link-wrapper}}
+                {{#> tree-view-link tree-view-link--modifier="pf-m-expandable"}}
+                    {{> tree-view-link-toggle}}
+                    {{#> tree-view-link-text}}
+                        This is a really really really long folder name that overflows from the width of the container.
+                    {{/tree-view-link-text}}
+                {{/tree-view-link}}
+                {{#> tree-view-action}}
                     <i class="fas fa-bell" aria-hidden="true"></i>
-                {{/tree-view-link-action}}
-            {{/tree-view-link}}
+                {{/tree-view-action}}
+            {{/tree-view-link-wrapper}}
         {{/tree-view-list-item}}
     {{/tree-view-list}}
 {{/tree-view}}
@@ -799,8 +809,19 @@ cssPrefix: pf-c-tree-view
 
 ### Accessibility
 | Attribute | Applied to | Outcome |
+| `role="grid"` | `.pf-c-tree-view__list` | Identifies the `<ul>` as a tree widget. **Place on the first <ul> only** |
+| `role="group"` | `.pf-c-tree-view__list` | Identifies the `<ul>` element as a container of treeitem elements that form a branch of the tree. **Place on all ul's except the first ul** |
 
 ### Usage
 | Class | Applied | Outcome |
 | -- | -- | -- |
-
+| `.pf-c-tree-view` | `<div>` | Initiates the tree view container. **Required** |
+| `.pf-c-tree-view__list` | `<ul>` | Initiates a tree view list. **Required** |
+| `.pf-c-tree-view__list-item` | `<li>` | Initiates a tree view list item. **Required** |
+| `.pf-c-tree-view__link` | `<button>`, `<a>` | Initiates a tree view link. **Required** |
+| `.pf-c-tree-view__toggle` | `<span>` | Initiates a tree view toggle. |
+| `.pf-c-tree-view__text | `<span>` | Initiates tree view text. |
+| `.pf-c-tree-view__icon | `<span>` | Initiates a tree view icon. |
+| `.pf-c-tree-view__link-wrapper` | `<div>` | Initiates a tree view link wrapper. **Required when using .pf-c-tree-view__action** |
+| `.pf-c-tree-view__action` | `<button>` | Initiates a tree view action. |
+| `.pf-c-tree-view__search` | `<div>` | Initiates a tree view search wrapper. |
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -10,82 +10,102 @@ beta: true
 {{#> tree-view}}
 	{{#> tree-view-list tree-view-list--attribute='role="tree"'}}
 		{{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-			{{#> tree-view-link tree-view-link--IsFirst="true"}}
-				{{> tree-view-link-toggle}}
-				{{#> tree-view-link-text}}
-					Application launcher
-				{{/tree-view-link-text}}
-      {{/tree-view-link}}
+      {{#> tree-view-node}}
+        {{#> tree-view-link tree-view-link--IsFirst="true"}}
+          {{> tree-view-link-toggle}}
+          {{#> tree-view-link-text}}
+            Application launcher
+          {{/tree-view-link-text}}
+        {{/tree-view-link}}
+      {{/tree-view-node}}
 			{{#> tree-view-list newcontext tree-view-list--IsChild="true" tree-view-list--modifier="pf-m-nested-level-1"}}
 				{{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-					{{#> tree-view-link tree-view-link--modifier="pf-m-first"}}
-						{{> tree-view-link-toggle}}
-						{{#> tree-view-link-text}}
-							Application 1
-						{{/tree-view-link-text}}
-					{{/tree-view-link}}
+          {{#> tree-view-node}}
+            {{#> tree-view-link tree-view-link--modifier="pf-m-first"}}
+              {{> tree-view-link-toggle}}
+              {{#> tree-view-link-text}}
+                Application 1
+              {{/tree-view-link-text}}
+            {{/tree-view-link}}
+          {{/tree-view-node}}
 					{{#> tree-view-list newcontext tree-view-list--modifier="pf-m-nested-level-2"}}
 						{{#> tree-view-list-item}}
-							{{#> tree-view-link}}
-								{{#> tree-view-link-text}}
-									Settings
-								{{/tree-view-link-text}}
-							{{/tree-view-link}}
-						{{/tree-view-list-item}}
-						{{#> tree-view-list-item}}
-							{{#> tree-view-link tree-view-link--modifier="pf-m-current"}}
-								{{#> tree-view-link-text}}
-									Current
-								{{/tree-view-link-text}}
-							{{/tree-view-link}}
+              {{#> tree-view-node}}
+                {{#> tree-view-link}}
+                  {{#> tree-view-link-text}}
+                    Settings
+                  {{/tree-view-link-text}}
+                {{/tree-view-link}}
+                {{/tree-view-node}}
+              {{/tree-view-list-item}}
+              {{#> tree-view-list-item}}
+                {{#> tree-view-node}}
+                  {{#> tree-view-link tree-view-link--modifier="pf-m-current"}}
+                    {{#> tree-view-link-text}}
+                      Current
+                    {{/tree-view-link-text}}
+                  {{/tree-view-link}}
+                {{/tree-view-node}}
 						{{/tree-view-list-item}}
 					{{/tree-view-list}}
 				{{/tree-view-list-item}}
 				{{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-					{{#> tree-view-link}}
-						{{> tree-view-link-toggle}}
-						{{#> tree-view-link-text}}
-							Application 2
-						{{/tree-view-link-text}}
-					{{/tree-view-link}}
+          {{#> tree-view-node}}
+            {{#> tree-view-link}}
+              {{> tree-view-link-toggle}}
+              {{#> tree-view-link-text}}
+                Application 2
+              {{/tree-view-link-text}}
+            {{/tree-view-link}}
+          {{/tree-view-node}}
 					{{#> tree-view-list newcontext tree-view-list--modifier="pf-m-nested-level-2"}}
 						{{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-							{{#> tree-view-link}}
-								{{> tree-view-link-toggle}}
-								{{#> tree-view-link-text}}
-									Settings
-								{{/tree-view-link-text}}
-							{{/tree-view-link}}
+              {{#> tree-view-node}}
+                {{#> tree-view-link}}
+                  {{> tree-view-link-toggle}}
+                  {{#> tree-view-link-text}}
+                    Settings
+                  {{/tree-view-link-text}}
+                {{/tree-view-link}}
+              {{/tree-view-node}}
 						{{/tree-view-list-item}}
 						{{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-							{{#> tree-view-link}}
-								{{> tree-view-link-toggle}}
-								{{#> tree-view-link-text}}
-									Loader
-								{{/tree-view-link-text}}
-							{{/tree-view-link}}
+              {{#> tree-view-node}}
+                {{#> tree-view-link}}
+                  {{> tree-view-link-toggle}}
+                  {{#> tree-view-link-text}}
+                    Loader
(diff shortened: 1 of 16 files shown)


src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -0,0 +1,840 @@
...
+{{/tree-view}}
+```
+
+## Documentation
+### Overview
+
+### Accessibility
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `role="tree"` | `.pf-c-tree-view__list` | Identifies the `ul` as a tree widget. **Place on the first ul only** |
```
mcoker, line 813:
```suggestion
| `role="tree"` | `.pf-c-tree-view__list` | Identifies the `ul` as a tree widget. **Place on the first `ul` only** |
```

mcoker: also it's fine as is, but maybe "outermost" instead of "first"?

mcoker: Oh after reading the rest of the docs, you're only putting the first instance in backticks? That works, too.

src/patternfly/components/TreeView/tree-view-link-icon.hbs
```diff
@@ -0,0 +1,14 @@
+<span class="pf-c-tree-view__link-icon{{#if tree-view-link-icon--modifier}} {{tree-view-link-icon--modifier}}{{/if}}"
+  {{#if tree-view-link-icon--attribute}}
+    {{{tree-view-link-icon--attribute}}}
+  {{/if}}>
+  {{#if tree-view-link-icon--IsFolder}}
+    {{#if tree-view-link-icon--IsExpanded}}
```
mcoker, line 6: I think you could write this in a single line via

`<i class="fas fa-folder{{#if tree-view-link-icon--IsExpanded}}-open{{/if}}" aria-hidden="true"></i>

src/patternfly/components/TreeView/tree-view-list-item.hbs
```diff
@@ -0,0 +1,8 @@
+<li class="pf-c-tree-view__list-item{{#if tree-view-list-item--expandable}} pf-m-expandable{{/if}}{{#if tree-view-list-item--expanded}} pf-m-expanded{{/if}}{{#if tree-view-list-item--modifier}} {{tree-view-list-item--modifier}}{{/if}}"
```
mcoker, line 1: I know we have a mix of use of attrs/vars in other components, but do you think for these boolean ones, they should be `tree-view-list-item--IsExpandable` and `tree-view-list-item--IsExpanded`?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+  --pf-c-tree-view__list--m-nested-level-3__list-item__link--PaddingLeft: calc(var(--pf-c-tree-view__list--m-nested-level-2__list-item__link--PaddingLeft) + var(--pf-global--spacer--md) + var(--pf-global--spacer--sm));
+  --pf-c-tree-view__list--m-nested-level-1__list-item--m-expandable__link--PaddingLeft: calc(var(--pf-c-tree-view__link--PaddingLeft) + var(--pf-global--spacer--md));
+  --pf-c-tree-view__list--m-nested-level-2__list-item--m-expandable__link--PaddingLeft: calc(var(--pf-c-tree-view__list--m-nested-level-1__list-item--m-expandable__link--PaddingLeft) + var(--pf-global--spacer--md));
+  --pf-c-tree-view__list--m-nested-level-3__list-item--m-expandable__link--PaddingLeft: calc(var(--pf-c-tree-view__list--m-nested-level-2__list-item--m-expandable__link--PaddingLeft) + var(--pf-global--spacer--md));
+
+  // link hover
+  --pf-c-tree-view__link--hover--BackgroundColor: var(--pf-global--BackgroundColor--200);
+
+  // toggle
+  --pf-c-tree-view__link--m-expanded__toggle--Rotate: 90deg;
```
mcoker, line 25: should this be called `__toggle-icon`?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+    --pf-c-check--GridGap: 0;
+
+    padding-right: var(--pf-c-tree-view__link--c-check--PaddingRight);
+  }
+
+  .pf-c-badge {
+    margin-left: var(--pf-c-tree-view__link--c-badge--MarginLeft);
+
+    &.pf-m-read {
+      --pf-c-badge--m-read--BackgroundColor: var(--pf-c-tree-view__link--c-badge--BackgroundColor);
```
mcoker, line 140: should probably be `--pf-c-tree-view__link--c-badge--m-read--BackgroundColor`

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+  align-items: center;
+  justify-content: center;
+  min-width: var(--pf-c-tree-view__link-toggle--MinWidth);
+  min-height: var(--pf-c-tree-view__link-toggle--MinHeight);
+  padding-right: var(--pf-c-tree-view__link-toggle--PaddingRight);
+  padding-left: var(--pf-c-tree-view__link-toggle--PaddingLeft);
+}
+
+.pf-c-tree-view__link-text {
+  overflow: hidden;
```
mcoker, line 156: wanna use `@include pf-text-overflow;` here?

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -0,0 +1,840 @@
+---
+title: Tree view
+section: components
```
mcoker, line 3: needs `beta: true`

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -0,0 +1,840 @@
...
+cssPrefix: pf-c-tree-view
+---
+
+## Examples
+```hbs title=Default
+{{#> tree-view}}
```
mcoker, line 9: this file uses 4 spaces instead of 2 for tabs

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -0,0 +1,840 @@
...
+| `role="treeitem"`| `.pf-c-tree-view__link` | Identifies the link as a tree item. **Note:** The role is set on the element that is interactive and focusable. |
+| `role="none"` | `.pf-c-tree-view__list-item` | Hides the implicit listitem role of the li element from assistive technologies. |
+| `aria-expanded="false"` | `.pf-c-tree-view__list-item.pf-m-expandable` | Indicates the parent node is closed, i.e., the descendant elements are not visible. |
+| `aria-expanded="true"` | `.pf-c-tree-view__list-item.pf-m-expandable.pf-m-expanded` | Indicates the parent node is open, i.e., the descendant elements are visible. |
+| `tabindex="-1"` | `.pf-c-tree-view__link` | Makes the element with the treeitem role focusable without including it in the tab sequence of the page. |
+| `tabindex="0"` | `.pf-c-tree-view__link` | Includes the element with the treeitem role in the tab sequence. Only one treeitem in the tree has tabindex="0". When the user moves focus in the tree, the element included in the tab sequence changes to the element with focus. |
```
mcoker, line 820: how do you navigate the tree by keyboard? All of our other menus like this (like nav) allow you to tab/focus on all of the items.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+
+  &.pf-m-first {
+    font-weight: var(--pf-c-tree-view__link--m-first--FontWeight);
+  }
+
+  &:hover {
+    background-color: var(--pf-c-tree-view__link--hover--BackgroundColor);
+  }
+
+  .pf-c-check {
```
mcoker, line 129: If we're just placing a checkbox there, I think we should just use a normal `<input type="checkbox">`. That or update the check component so it can work without a label - I just created an issue for that here https://github.com/patternfly/patternfly/issues/3311

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+    --pf-c-check__input--MarginTop: 0;
+    --pf-c-check--GridGap: 0;
+
+    padding-right: var(--pf-c-tree-view__link--c-check--PaddingRight);
```
mcoker, line 133: Also if we keep this, we should use a `margin`, as we (shouldn't anyways) ever have a `margin` on this component, but may have `padding` that this style could conflict with.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  .pf-c-badge {
+    margin-left: var(--pf-c-tree-view__link--c-badge--MarginLeft);
```
mcoker, line 137: wdyt about creating `__check/select` and `__count` elements in this component that we place the check and badge components in, then we style the elements of this component for spacing/positioning.

src/patternfly/components/TreeView/tree-view.scss
mcoker: Actually you could use the line-clamp mixin that would allow a user to specify they want to truncate after the value of an optional `--pf-c-tree-view__link-text--max-lines` var, and there could be a prop in react for that.

Screen Shot 2020-07-20 at 6 44 35 PM


https://github.com/patternfly/patternfly/blob/3ab70e9587c143ec2ff06e1b99b0cb2cefd07305/src/patternfly/sass-utilities/mixins.scss#L153-L160

mcoker: also you'll want to add `word-break: break-word;` regardless

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+}
+
+.pf-c-tree-view__link-toggle {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  min-width: var(--pf-c-tree-view__link-toggle--MinWidth);
+  min-height: var(--pf-c-tree-view__link-toggle--MinHeight);
```
mcoker, line 150: why are these necessary? Seems like you don't need them, though you could add `fa-fw` to the toggle icons so they're more of a square, more representative of how they'll look in react as svg's.

mcoker:
From the design it looks like if you have both expanandable groups, and regular items in the same list, the text should be aligned.

The design

Screen Shot 2020-07-20 at 7 07 56 PM

This PR

Screen Shot 2020-07-20 at 7 03 07 PM

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+  --pf-c-tree-view__action--Color: var(--pf-global--icon--Color--light);
+  --pf-c-tree-view__action--m-favorite--Color: var(--pf-global--palette--gold-400);
+
+  padding-top: var(--pf-c-tree-view--PaddingTop);
+  padding-bottom: var(--pf-c-tree-view--PaddingBottom);
+}
+
+// stylelint-disable
+.pf-c-tree-view__list {
+  &.pf-m-nested-level-1 {
```
mcoker, line 60: Hmmmm.... this section is tricky. Off the top of my head, I think we could do something like this, but I bet there is a better way.

```
--nested-padding: 16px;
.pf-c-tree-view__list {
  $parent = &;
  & & {
    __link--PaddingLeft: --nested-padding;
    & #{$parent} {
      __link--PaddingLeft: calc(--nested-padding * 2);
      & #{$parent} {
        __link--PaddingLeft: calc(--nested-padding * 3);
        // and so on...
      }
    }
  }
}

File: src/patternfly/components/TreeView/tree-view.scss [hunk 15]
```diff
@@ -0,0 +1,196 @@
...
+.pf-c-tree-view__search {
+  padding: var(--pf-c-tree-view__search--PaddingTop) var(--pf-c-tree-view__search--PaddingRight) var(--pf-c-tree-view__search--PaddingBottom) var(--pf-c-tree-view__search--PaddingLeft);
+}
+
+.pf-c-tree-view__link-icon {
+  padding-right: var(--pf-c-tree-view__link-icon--PaddingRight);
+  color: var(--pf-c-tree-view__link-icon--Color);
+}
+
+.pf-c-tree-view__link-wrapper {
```
mcoker, line 170: the way we do this in the app launcher really isn't ideal, it was built that way to add that functionality after the fact without breaking the component. IMO these items should always have this "wrapper", but let's call it something else. Then we can apply the hover/focus colors consistently (on this item), instead of on both this item and its children.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+  display: grid;
+  grid-template-columns: minmax(0, 100%) auto;
+
+  &:hover {
```
mcoker, line 174: I'm confused about the lack of ability to tab/focus to things. I would expect the background change to happen on focus, and with this item, `:focus-within`. Also `.pf-c-tree-view__action` is only visible on `:hover` - what about keyboard users?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+.pf-c-tree-view__action {
+  display: none;
+  padding-right: var(--pf-c-tree-view__action--PaddingRight);
+  padding-left: var(--pf-c-tree-view__action--PaddingLeft);
+  margin-left: auto;
+  color: var(--pf-c-tree-view__action--Color);
+  visibility: hidden;
+  border: 0;
+
+  &.pf-m-favorite {
```
mcoker, line 193: if this should be the same as the favorite in other components, the star should be smaller and have hover/focus color changes. Also where are favorited items stored in the layout? Or is that all left up to the user?

I wonder if favoriting should be saved for another sprint, too, if it's a lot of work and not required for the initial component.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,196 @@
...
+}
+
+.pf-c-tree-view__link-icon {
+  padding-right: var(--pf-c-tree-view__link-icon--PaddingRight);
+  color: var(--pf-c-tree-view__link-icon--Color);
+}
+
+.pf-c-tree-view__link-wrapper {
+  display: grid;
+  grid-template-columns: minmax(0, 100%) auto;
```
mcoker, line 172: This should be the same as `1fr auto;` though, I don't think we need to use grid. This is a row, not a grid - I think flex would be a lot better. Also if we end up supporting multiple actions in an item, we can just add them in a flex layout, versus having to modify the `grid-template-columns` to create new columns. Just set the link element to grow.

src/patternfly/components/TreeView/tree-view.scss
doruskova: @mcoker Agree that the action should have a hover state. However, the favorite items are not stored in a layout, because I think that it's a different scenario/pattern. The general idea for the tree view menu was to have any action item on the right (depends on a use-case). If we need a pattern for "Favorite items", then it needs more exploration, but IMHO should be possible to build a place where these items are stored from existing components. For example, add the "Favorite items" tab above the tree or drill down menu where you can see the list of items you selected.

@mcarrano What do you think? Do we have use-cases for "Selected/Favorite" items? I have a use-case for selected items in a multi-level drill-down menu ([link here](https://marvelapp.com/prototype/66b0059/screen/71366718)). But, I still think that it can be implemented from existing components.

mcoker: My concern is that we already have a specific favoriting pattern in lists like this, so it seems like it would be inconsistent if this pattern didn't match the other implemenations -- at least in the design, but also potentially in the interaction pattern

https://www.patternfly.org/v4/documentation/core/components/applauncher#favorites
https://www.patternfly.org/v4/documentation/core/components/select#favorites

doruskova: @mcoker ooohh, that makes sense. Then I think that we need to discuss the other implementation to match everything else. This feature wasn't required for the initial component. Thanks for the links :)

src/patternfly/components/TreeView/tree-view.scss
christiemolloy: Tried doing line-clamp vs text-overflow and the spacing to the left is larger for line-clamp, do you know why this might be 

Screen Shot 2020-07-24 at 3 05 34 PM

Screen Shot 2020-07-24 at 3 06 06 PM

src/patternfly/components/TreeView/examples/TreeView.md
christiemolloy: This is what I was following for the tree view accessibility https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html 

- "All treeitem elements are focusable, but only one is included in the tab sequence."
- "Only one treeitem in the tree has tabindex="0"."

src/patternfly/components/TreeView/tree-view.scss
christiemolloy: Sounds good, I'll use a normal input for now and we could follow up with adding the check component here once the issue is finished.

src/patternfly/components/TreeView/tree-view.scss
christiemolloy: Yep that sounds like a solid approach to me!

src/patternfly/components/TreeView/examples/TreeView.md
mcoker: Ah gotcha, the keyboard navigation is managed via JS.

src/patternfly/components/TreeView/tree-view.scss
mcoker: are you referring to the space after the `...`? If so it's because `line-clamp` will actually lay something out in as many lines as it would be naturally, and clip the end of the text on whatever line you specify with `line-clamp`.  If you changed `line-clamp` to "2", you'll see that the text doesn't shift. Here's an exaggerated example to show that https://codepen.io/mcoker/pen/gOPEqge

christiemolloy: @mcoker got it. So if there are multiple lines that overflow, the differences could look like this. If the different spaces on the right seems natural to you, im ok with it

Screen Shot 2020-07-27 at 9 48 11 AM

christiemolloy:
> From the design it looks like if you have both expanandable groups, and regular items in the same list, the text should be aligned.
> 
> The design
> 
> Screen Shot 2020-07-20 at 7 07 56 PM
> 
> This PR
> 
> Screen Shot 2020-07-20 at 7 03 07 PM

Where do you see this ? Or did you add the loader groups into one of the examples, if so which nested level was it ? Right now I see this
Screen Shot 2020-07-27 at 9 52 28 AM

src/patternfly/components/TreeView/tree-view.scss
mcarrano: @mcoker @doruskova Trying to catch up on this. I agree that processing favorites in the tree menu in the same way that we do for other menus might not work. I don't know that we have a requirement for this, so maybe we should just remove favoriting  from the examples for now and we can revisit this if or when it comes up?

doruskova: @mcarrano I agree with removing favoriting for now and we can add it later because it needs more exploration. It wasn't required for the initial component.

christiemolloy: Sounds good, will remove the favoriting examples and we can follow-up when its needed

src/patternfly/components/TreeView/tree-view.scss
christiemolloy: are you suggesting that I add flex: grow on the link element right now ? it already has width: 100% on it.

src/patternfly/components/TreeView/tree-view.scss
christiemolloy: Is this still a concern? or does what i said about managing focus via JS cover it?

christiemolloy: Are you suggesting I add focus-within to the wrapper (now called node) element ?

src/patternfly/components/TreeView/tree-view.scss
christiemolloy: I renamed this `.pf-c-tree-view__node` and added it around each link. I moved the hover to this element, but should the focus still remain on the link inside since its the `<button>` element?

src/patternfly/components/TreeView/tree-view.scss
mcoker: Good point, using `@include pf-text-overflow;` is fine. If users want to specify the number of lines, we can consider the other way with this caveat.

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -0,0 +1,973 @@
...
+      {{#> tree-view-node}}
+        {{#> tree-view-link tree-view-link--IsFirst="true"}}
+          {{> tree-view-link-toggle}}
+          {{#> tree-view-link-text}}
+            Application launcher
+          {{/tree-view-link-text}}
+        {{/tree-view-link}}
+      {{/tree-view-node}}
+			{{#> tree-view-list newcontext tree-view-list--IsChild="true" tree-view-list--modifier="pf-m-nested-level-1"}}
+				{{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
```
mcoker, line 22: Indentation is off in a lot of spots in this file

src/patternfly/components/TreeView/tree-view-link-icon.hbs
```diff
@@ -0,0 +1,10 @@
+<span class="pf-c-tree-view__link-icon{{#if tree-view-link-icon--modifier}} {{tree-view-link-icon--modifier}}{{/if}}"
+  {{#if tree-view-link-icon--attribute}}
+    {{{tree-view-link-icon--attribute}}}
+  {{/if}}>
+  {{#if tree-view-link-icon--IsFolder}}
+    <i class="fas fa-folder{{#if tree-view-link-icon--IsExpanded}}-open{{/if}}" aria-hidden="true"></i>
```
mcoker, line 6: 👍

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
+.pf-c-tree-view {
+  --pf-c-tree-view--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-tree-view--PaddingBottom: var(--pf-global--spacer--sm);
```
mcoker, line 3: Should this have top/bottom padding? If it's placed in a dropdown menu, for example, that padding will come from `.pf-c-dropdown__menu`

mcoker:
I'm having a hard time figuring out the pattern for the indentation... is there a pattern? Or is it just a super specific padding depending on the depth level and whether or not the item is expandable? 

Screen Shot 2020-07-28 at 6 58 57 PM

mcoker:
> Where do you see this ? Or did you add the loader groups into one of the examples, if so which nested level was it ? Right now I see this

All of the examples look that way for me.

Screen Shot 2020-07-28 at 7 04 54 PM

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -0,0 +1,973 @@
...
+```
+
+```hbs title=With-checkboxes
+{{#> tree-view}}
+    {{#> tree-view-list tree-view-list--attribute='role="tree"'}}
+        {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
+          {{#> tree-view-node}}
+            {{#> tree-view-link tree-view-link--IsFirst="true"}}
+                {{> tree-view-link-toggle}}
+								<input type="checkbox"/>
```
mcoker, line 303: wdyt about putting this in a `__link-check`element?

src/patternfly/components/TreeView/tree-view.scss
mcoker: Can we at least apply the appropriate level of padding just using nesting selectors and changing the `__link--PaddingLeft` value, so that we don't need the `pf-m-nested-level-n` modifiers, or is there value in keeping those?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
...
+      --pf-c-badge--m-read--BackgroundColor: var(--pf-c-tree-view__link-count--c-badge--m-read--BackgroundColor);
+    }
+  }
+}
+
+.pf-c-tree-view__link-toggle-icon {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  min-width: var(--pf-c-tree-view__link-toggle-icon--MinWidth);
```
mcoker, line 150: What is the width for?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
...
+  min-height: var(--pf-c-tree-view__link-toggle-icon--MinHeight);
```
mcoker, line 151: Same - how is this used?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
...
+  padding-right: var(--pf-c-tree-view__link-toggle-icon--PaddingRight);
+  padding-left: var(--pf-c-tree-view__link-toggle-icon--PaddingLeft);
```
mcoker, line 153: I don't see the left spacer in the design, is that necessary?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
...
+.pf-c-tree-view__search {
+  padding: var(--pf-c-tree-view__search--PaddingTop) var(--pf-c-tree-view__search--PaddingRight) var(--pf-c-tree-view__search--PaddingBottom) var(--pf-c-tree-view__search--PaddingLeft);
+}
+
+.pf-c-tree-view__link-icon {
+  padding-right: var(--pf-c-tree-view__link-icon--PaddingRight);
+  color: var(--pf-c-tree-view__link-icon--Color);
+}
+
+.pf-c-tree-view__node {
```
mcoker, line 173: I'm not sure about the `__node` name :) In the app launcher, we named the `li` `__wrapper`. It doesn't seem like we need this `__node` element? Or is it needed for something? 

One thing it will help with is if we add a `__description` or something, and want to separate the item's top row of things from other things below it. Maybe this could be called `__main` or just `__row`?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
...
+        --pf-c-tree-view__link--PaddingLeft: var(--pf-c-tree-view__list--m-nested-level-3__list-item__link--PaddingLeft);
+      }
+    }
+  }
+}
+// stylelint-enable
+
+.pf-c-tree-view__list-item {
+  &.pf-m-expanded {
+    > .pf-c-tree-view__link > .pf-c-tree-view__link-toggle-icon {
```
mcoker, line 104: This isn't working - `__node` is missing.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -0,0 +1,198 @@
...
+  --pf-c-tree-view__list--m-nested-level-3__list-item__link--PaddingLeft: calc(var(--pf-c-tree-view__list--m-nested-level-2__list-item__link--PaddingLeft) + var(--pf-global--spacer--md) + var(--pf-global--spacer--sm));
+  --pf-c-tree-view__list--m-nested-level-1__list-item--m-expandable__link--PaddingLeft: calc(var(--pf-c-tree-view__link--PaddingLeft) + var(--pf-global--spacer--md));
+  --pf-c-tree-view__list--m-nested-level-2__list-item--m-expandable__link--PaddingLeft: calc(var(--pf-c-tree-view__list--m-nested-level-1__list-item--m-expandable__link--PaddingLeft) + var(--pf-global--spacer--md));
+  --pf-c-tree-view__list--m-nested-level-3__list-item--m-expandable__link--PaddingLeft: calc(var(--pf-c-tree-view__list--m-nested-level-2__list-item--m-expandable__link--PaddingLeft) + var(--pf-global--spacer--md));
+
+  // link hover/focus
+  --pf-c-tree-view__node--hover--BackgroundColor: var(--pf-global--BackgroundColor--200);
+  --pf-c-tree-view__link--focus--BackgroundColor: var(--pf-global--palette--black-150);
+
```
mcoker, line 24: Are the node:hover, and link:focus colors different on purpose? The difference is nearly indiscernible (`#f5` vs `#f0`), I'm wondering if they should be the same. And if they should be triggered on the same element, using `:focus-within` if the element changing background color isn't focusable.

mcoker:
I'd like to review the designs so I can make sure I understand the variations in background colors, the meaning of the "first" item, and what it means for something to be "current".

doruskova:
> I'm having a hard time figuring out the pattern for the indentation... is there a pattern? Or is it just a super specific padding depending on the depth level and whether or not the item is expandable?
> 
> Screen Shot 2020-07-28 at 6 58 57 PM

I tried to play with it a little bit in a web console and I created this guide. https://marvelapp.com/prototype/8cdg211/screen/71593599 I used this left-padding: 8px > 24px > 48px > 72px > 96px. The difference is always 24px. The only problem I had was that the toggle icon has 4px left padding, which moves the item more to the right side and then it looks more indented. So, I set ` --pf-c-tree-view__link-toggle-icon--PaddingLeft: 0px;` 
I think that it looks good even if it's not perfectly aligned, but it should look consistent. WDYT @mcoker ?

PR #3317, opened 2020-07-23
fix(toolbar-padding): updated toolbar padding

### LINKED ISSUE #3240: Toolbar padding enhancements
- ~~Update toolbar padding to increase from `--md` to `--lg` at medium breakpoint, like `page__main-section`, `card`, etc~~
- Update toolbar padding values for `.pf-c-page__main-section`, `.pf-c-card`, `.pf-c-table` etc to contextually reflect padding left/right 
- Add `pf-m-no-padding` variant.

mattnolting:
fixes #3240

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -2,6 +2,7 @@
 // @import "../../sass-utilities/all";
 $pf-c-toolbar--breakpoint-map: build-breakpoint-map("base", "md", "lg", "xl", "2xl");
 $pf-c-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
+$pf-c-toolbar--inset-map: build-spacer-map();
 
 .pf-c-toolbar {
   --pf-c-toolbar--BackgroundColor: var(--pf-global--BackgroundColor--100);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -2,16 +2,22 @@
...
   --pf-c-toolbar--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-toolbar--RowGap: var(--pf-global--spacer--lg);
   --pf-c-toolbar--PaddingTop: var(--pf-global--spacer--md);
   --pf-c-toolbar--PaddingBottom: var(--pf-global--spacer--md);
 
+  // Gutters
+  --pf-c-toolbar--gutter: var(--pf-global--gutter);
+  --pf-c-toolbar--xl--gutter: var(--pf-global--gutter--md);
+  --pf-c-toolbar--gutter--stack: var(--pf-c-toolbar--inset, var(--pf-c-toolbar--gutter));
+
```
mcoker, line 17: This PR looks good, but I'm not sure about the way these vars are defined. In the other components that do this, we're using `--spacer` global vars, not `--gutter` global vars. Our use of the `--gutter` vars is for creating gaps. For example,

https://github.com/patternfly/patternfly/blob/4a1a56f1eb307fd25b87993687de9594069b5f20/src/patternfly/layouts/Grid/grid.scss#L71-L73

Also the `gutter--md` var was traditionally to reflect the `--md` media query, but it's being used for the `--xl` media query here. 

And the `--stack` var isn't something we're doing anywhere else. To be consistent, we would just modify `--pf-c-toolbar--gutter` in the `.pf-m-inset` sass loop. I like that, but I think we should consider what it would look like to do this across the library, and be consistent in that approach.

Just to be consistent, I would be in favor of leaving `--pf-c-toolbar__content--Padding[Right/Left]`, and setting those to `--spacer` vars, then in a MQ, update them to the wide viewport `--spacer` vars. And in the `pf-m-inset` loop, do the same (which is what you're doing now). wdyt?

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -2,16 +2,22 @@
 // @import "../../sass-utilities/all";
 $pf-c-toolbar--breakpoint-map: build-breakpoint-map("base", "md", "lg", "xl", "2xl");
 $pf-c-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
+$pf-c-toolbar--inset-map: build-spacer-map();
```
mcoker, line 5: without values passed to the function, looks like it will build these values https://github.com/patternfly/patternfly/blob/4a1a56f1eb307fd25b87993687de9594069b5f20/src/patternfly/sass-utilities/scss-variables.scss#L302-L309

If so, the docs should be updated. They currently list `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`

PR #3320, opened 2020-07-23
feat(skeleton): add skeleton loading component

### LINKED ISSUE #3105: Skeleton loading
Introduce support for Skeleton loading. Design document here: https://marvelapp.com/61e541a/screen/69171533

cc: @LHinson @mcoker

christiemolloy:
closes #3105

### DIFF
--- a/src/patternfly/components/Skeleton/examples/Skeleton.md
+++ b/src/patternfly/components/Skeleton/examples/Skeleton.md
@@ -310,7 +310,8 @@ Large rectangle
 | `.pf-c-skeleton__element` | `<div>` | Initiates the skeleton element. **Required** |
 | `.pf-c-skeleton__container` | `<div>` | Initiates the skeleton container. |
 | `.pf-m-padding` | `.pf-c-skeleton` | Modifies the skeleton to have padding. |
-| `.pf-m-spacer-[xs,sm,md,lg,xl,2xl]` | `.pf-c-skeleton__container` | Modifies the spacing between skeleton elements in the container. |
+| `.pf-m-spacer-[xs,sm,md,lg,xl,2xl]` | `.pf-c-skeleton__container` | Modifies the spacing between skeleton elements in the container. | `.pf-c-skeleton__container` | Modifies the skeleton container to center the elements inside of it. 
+| `pf-m-justify-items-center` | `
 | `.pf-m-padding-top-[xs,sm,md,lg,xl,2xl]` | `.pf-c-skeleton__container` | Modifies the padding top on the skeleton container. |
 | `.pf-m-padding-bottom-[xs,sm,md,lg,xl,2xl]` | `.pf-c-skeleton__container` | Modifies the padding bottom on the skeleton container. |
 | `.pf-m-height-[xs,sm,md,lg,xl,2xl]` | `.pf-c-skeleton__element` | Modifies the height of the skeleton element. |
--- a/src/patternfly/components/Skeleton/examples/Skeleton.md
+++ b/src/patternfly/components/Skeleton/examples/Skeleton.md
@@ -1,0 +1,127 @@
+---
+title: Skeleton
+section: components
+cssPrefix: pf-c-skeleton
+beta: true
+---
+
+## Examples
+```hbs title=Default
+{{#> skeleton}}
+  {{#> skeleton-element}}
+  {{/skeleton-element}}
+{{/skeleton}}
+```
+
+```hbs title=Width-modifiers
+{{#> skeleton}}
+  {{#> skeleton-element skeleton-element--modifier="pf-m-width-20"}}
+  {{/skeleton-element}}
+{{/skeleton}}
+<br/>
+{{#> skeleton}}
+  {{#> skeleton-element skeleton-element--modifier="pf-m-width-40"}}
+  {{/skeleton-element}}
+{{/skeleton}}
+<br/>
+{{#> skeleton}}
+  {{#> skeleton-element skeleton-element--modifier="pf-m-width-60"}}
+  {{/skeleton-element}}
+{{/skeleton}}
+<br/>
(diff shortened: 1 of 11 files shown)


src/patternfly/components/Skeleton/examples/Skeleton.md
```diff
@@ -0,0 +1,321 @@
...
+  {{#> skeleton-element skeleton-element--modifier="pf-m-rectangle-lg"}}
+  {{/skeleton-element}}
+{{/skeleton}}
+```
+
+```hbs title=Skeleton-container
+
+<div>4px spacing</div>
+{{#> skeleton}}
+  {{#> skeleton-container skeleton-container--modifier="pf-m-spacer-sm"}}
```
mcoker, line 137: did you mean to use xs here?

```suggestion
  {{#> skeleton-container skeleton-container--modifier="pf-m-spacer-xs"}}
```

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,200 @@
...
+  // Container
+  --pf-c-skeleton__container--GridGap: 0;
+  --pf-c-skeleton__container--PaddingTop: 0;
+  --pf-c-skeleton__container--PaddingBottom: 0;
+
+  // Element
+  --pf-c-skeleton__element--BackgroundColor: var(--pf-global--palette--black-150);
+  --pf-c-skeleton__element--Height: var(--pf-global--spacer--lg);
+  --pf-c-skeleton__element--Width: 100%;
+  --pf-c-skeleton__element--BorderRadius: #{pf-size-prem(4px)}; // add variable ?
```
mcoker, line 16: Seems like we should be able to use one of our existing border-radius vars?

christiemolloy: @doruskova would you be ok with using a 3px border radius to align with the system?

doruskova: @christiemolloy sure, no problem. I tried it in my codepen and the difference is minimal. Keep consistency with the system :)

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,217 @@
...
+  --pf-c-skeleton__element--m-width-35--Width: 35%;
+  --pf-c-skeleton__element--m-width-40--Width: 40%;
+  --pf-c-skeleton__element--m-width-45--Width: 45%;
+  --pf-c-skeleton__element--m-width-50--Width: 50%;
+  --pf-c-skeleton__element--m-width-60--Width: 60%;
+  --pf-c-skeleton__element--m-width-70--Width: 70%;
+  --pf-c-skeleton__element--m-width-80--Width: 80%;
+  --pf-c-skeleton__element--m-width-90--Width: 90%;
+
+  &.pf-m-padding {
```
mcoker, line 58: do we need the parent element, and this modifier?

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  &.pf-m-width-45 {
+    --pf-c-skeleton__element--Width: var(--pf-c-skeleton__element--m-width-45--Width);
+  }
+
+  &.pf-m-width-50 {
+    --pf-c-skeleton__element--Width: var(--pf-c-skeleton__element--m-width-50--Width);
+  }
+
```
mcoker, line 159: It's kind of odd we don't do +5% increments after 50%. We could also use the same widths that are in the sizing utility - https://github.com/patternfly/patternfly/blob/61898c01f7dcbb7937d9ba624ab374166f9d96cb/src/patternfly/utilities/Sizing/sizing.scss#L5-L14

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,217 @@
+.pf-c-skeleton {
+  // Padding modifier
+  --pf-c-skeleton--m-padding--Padding: var(--pf-global--spacer--lg);
+
+  // Element
+  --pf-c-skeleton__element--BackgroundColor: var(--pf-global--palette--black-150);
+  --pf-c-skeleton__element--Height: var(--pf-global--spacer--lg);
+  --pf-c-skeleton__element--Width: 100%;
```
mcoker, line 8: do we need this?

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,217 @@
+.pf-c-skeleton {
+  // Padding modifier
+  --pf-c-skeleton--m-padding--Padding: var(--pf-global--spacer--lg);
+
+  // Element
+  --pf-c-skeleton__element--BackgroundColor: var(--pf-global--palette--black-150);
+  --pf-c-skeleton__element--Height: var(--pf-global--spacer--lg);
```
mcoker, line 7: Should this just be the current font size * line height maybe? I think that would make a good default.

src/patternfly/components/Skeleton/skeleton.scss
christiemolloy: True, I was thinking for elements that didnt have padding. But maybe we remove this and if there is a use case where the component needs padding we could add it back in

src/patternfly/components/Skeleton/skeleton.scss
christiemolloy: Yeah I was following what was in the table. But Ill follow the sizing utility instead and maybe one day we can update the table

src/patternfly/components/Skeleton/skeleton.scss
christiemolloy: Yep and that aligns with the text modifiers - sounds good

Issue #3313, opened 2020-07-09
Bug: Inline Link Button Not working as expected in components 


christiemolloy:
We noticed in the Hint component, that the Inline link button is not working as expected. 

1) You should be able to put the inline link button inside of paragraph text without it wrapping to its own new line, see example below:

<img width="655" alt="Screen Shot 2020-07-09 at 2 28 59 PM" src="https://user-images.githubusercontent.com/20118816/87077089-9ab3ae00-c1f0-11ea-88eb-add59f0edf6e.png">

2) The inline link button should wrap inside of its container just like text does, this is how it appears in the Hint component:

<img width="935" alt="Screen Shot 2020-07-09 at 2 31 52 PM" src="https://user-images.githubusercontent.com/20118816/87077363-039b2600-c1f1-11ea-8537-07ceffe7af41.png">

cc: @mcoker 

Related to patternfly/patternfly#3218

mcoker:
I tried everything I can think of and as far as I can tell, this isn't possible because of how browsers treat buttons and do not allow a `display: inline;` style. More info - https://www.scottohara.me/blog/2018/10/03/unbutton-buttons.html

From there, I also tried `display: contents;`, and while it works to simply allow the text to wrap inline, the limitations around a11y and `display: contents;` are still an issue - it removes the element from the a11y tree. You also lose all of the element's agent styling.

It seems like the best solution might be to use a `<span>` and manipulate the `role`, as suggested in the link above. WDYT  @jessiehuff?

jessiehuff:
@mcoker After taking a look at this, trying a few things, and reading this article, I agree with you. That seems like the best solution. Scott O'Hara is also a fairly well known thought leader in the accessibility space so I feel like it's a solid source to base our approach on.

PR #3325, opened 2020-07-24
fix(button): allowed inline link text to wrap and use with span element

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3313


mcoker:
@christiemolloy re: @jenny-s51's comment

> can we update the hint example with the new modifier?

Does the button in the hint example need to be a button, or could that just be a link?

christiemolloy:
> @christiemolloy re: @jenny-s51's comment
> 
> > can we update the hint example with the new modifier?
> 
> Does the button in the hint example need to be a button, or could that just be a link?

I think the dropdown button and the buttons in the footer are meant to buttons, but link would go in the body paragraph in the hint . hopefully that answers the question :)

mcoker:
> link would go in the body paragraph in the hint

Great, agreed, thanks! Updated that example @christiemolloy @jenny-s51

mattnolting (approved):
Nice work @mcoker! I can see where this is needed. ~~@christiemolloy The hint component in the SS should be a standard link though, right?~~ I see you already discussed this..

christiemolloy (approved):
LGTM! conflict

mcoker:
closed via https://github.com/patternfly/patternfly/pull/3470

Issue #3260, opened 2020-07-08
Ensure nested drawers behave properly


mcoker:
re: https://github.com/patternfly/patternfly-react/issues/4497

Let's put together a demo of nested drawers and ensure any changes to the drawer are scoped to that instance, and not inherited by nested instances.

jcaianirh:
OpenShift Bug this will fix: https://bugzilla.redhat.com/show_bug.cgi?id=1859258

PR #3328, opened 2020-07-27
fix(drawer): scoped drawer styles to improve behavior of nested drawers

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3260

I took the approach of traditional style overrides using selectors and properties instead of overriding variables. There are cases where overriding a variable would work, but to be consistent, I kept the styling changes consistent. It may be a better approach to define something like - a nested drawer can only exist in the drawer__content element, so applying var overrides to elements in the drawer__panel element is fine, but this way, a nested drawer can exist anywhere inside of the drawer component and it should work fine*.

Another consideration was using the `*` selector in place of `__main` (and maybe `__body`, or maybe `.pf-c-drawer > * (main) > * (panel) > * (body) > * (head) > .pf-c-drawer__close` as a means to style the close element, since we expect it to be in a specific spot), but it seems a bit dangerous to do that. Doing that in place of `__main` seems predictable enough though. I'm curious others thoughts.

* a nested drawer can't be in the `__content > __body > __head` element.

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -74,8 +74,8 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   }
 
   // Actions
-  --pf-c-drawer__actions--MarginTop: calc(#{pf-size-prem(6px)} * -1);
-  --pf-c-drawer__actions--MarginRight: calc(#{pf-size-prem(6px)} * -1);
+  --pf-c-drawer__actions--MarginTop: calc(var(pf-global--spacer--form-element) * -1);
+  --pf-c-drawer__actions--MarginRight: calc(var(pf-global--spacer--form-element) * -1);
 
   // Box shadow
   --pf-c-drawer__panel--BoxShadow: none;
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Drawer/drawer.scss
```diff
@@ -134,18 +134,17 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 
 // Content
 .pf-c-drawer__content {
-  --pf-c-drawer--child--PaddingTop: var(--pf-c-drawer__content--child--PaddingTop);
-  --pf-c-drawer--child--PaddingRight: var(--pf-c-drawer__content--child--PaddingRight);
-  --pf-c-drawer--child--PaddingBottom: var(--pf-c-drawer__content--child--PaddingBottom);
-  --pf-c-drawer--child--PaddingLeft: var(--pf-c-drawer__content--child--PaddingLeft);
+  > .pf-c-drawer__body {
+    padding-top: var(--pf-c-drawer__content--child--PaddingTop) var(--pf-c-drawer__content--child--PaddingRight) var(--pf-c-drawer__content--child--PaddingBottom) var(--pf-c-drawer__content--child--PaddingLeft);
```
mattnolting, line 138:
```suggestion
    padding: var(--pf-c-drawer__content--child--PaddingTop) var(--pf-c-drawer__content--child--PaddingRight) var(--pf-c-drawer__content--child--PaddingBottom) var(--pf-c-drawer__content--child--PaddingLeft);
```

PR #3339, opened 2020-07-28
feat(label): add support for overflow

### LINKED ISSUE #3259: Label - support overflow ellipsis and tooltip

_Unexpected_: Labels that have long character strings grow indefinitely.
Screen Shot 2020-07-01 at 2 25 34 PM

_Expected_: Labels follow the same behavior as Chips - truncate long text, show an ellipsis and tooltip.
Screen Shot 2020-07-01 at 2 24 30 PM

Create a label with a large number of characters

Bug

Tower 
cc @trahman73

christiemolloy:
closes #3259

### DIFF
--- a/src/patternfly/components/Label/examples/Label.md
+++ b/src/patternfly/components/Label/examples/Label.md
@@ -39,11 +39,13 @@ import './Label.css'
   Grey link removable
 {{/label}}
 
-{{#> label label--id="default-grey-icon-overflow" label--isRemovable="true"}}
+{{#> label label--id="default-grey-icon-close-overflow" label--isRemovable="true"}}
   {{#> label-icon}}
     <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
   {{/label-icon}}
-  Grey icon overflow
+  {{#> label-text}}
+    Grey label with icon that overflows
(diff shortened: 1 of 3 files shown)


mcoker:
@mcarrano @mceledonia would you consider this change visually breaking? Wondering if this should be opt-in, or just the new default for labels.

mcarrano:
I don't consider it visually breaking. Why wouldn't you want this? What do you think @mceledonia ? Also, isn't the Label component still beta?

mcoker:
@mcarrano a potential implication is if there are labels currently that exceed the width, the content that is currently visible will not be without having to hover the label to see a tooltip, and that could be unwanted. I think it's fine, since it sounds like it was an oversight not to have it in the first place, so it's more of a bug fix. I just wanted to clarify since the way we make the update in react could depend on wither a prop enables it or not.

mcarrano:
@christiemolloy @mcoker I'm kind of torn on this one. I guess there could be cases where you would not want the label to truncate. Let me get some other design opinions. @mceledonia @maryshak1996 @kybaker @gdoyle1 what do you think? In your usage of labels, are there cases where you might not want a long label name to truncate?

Issue #2910, opened 2020-03-31
Introduce Max-Width Page


mcarrano:
Want to support placing a max-width on a page to make designs more usable on wide screen monitors. This should eventually become our default behavior but may initially be introduced as an opt-in. Design guidance here: https://marvelapp.com/7191idb/screen/67734666

cc @mcoker @LHinson

PR #3340, opened 2020-07-28
feat(page): added ability to limit width of content in page section

mcoker:
fixes https://github.com/patternfly/patternfly/issues/2910

@mcarrano @mceledonia you can see the enhancement in:

* The breadcrumb, tertiary nav, title/content and gallery sections on the page demos - https://patternfly-pr-3340.surge.sh/documentation/core/demos/page
* The basic table demo - https://patternfly-pr-3340.surge.sh/documentation/core/demos/table/basic
  * In this demo, I changed the table page section to `.pf-m-light` (gives it a white background instead of grey), and removed the card that the table was in.
* The in-page wizard demos
  * https://patternfly-pr-3340.surge.sh/documentation/core/demos/wizard/in-page-nav-expanded-mobile
  * https://patternfly-pr-3340.surge.sh/documentation/core/demos/wizard/in-page
  * I updated the default background color of the wizard container to be white. It was previously transparent, but never visible because the wizard would cover it.

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -94,7 +94,7 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
   --pf-c-page__main-breadcrumb--main-section--PaddingTop: var(--pf-global--spacer--md);
 
   // Main section body
-  --pf-c-page__main-section--m-limit-width__main-section-body--MaxWidth: calc(#{pf-size-prem(1710px)} - var(--pf-c-page__sidebar--Width));
+  --pf-c-page__main-section--m-limit-width__main-section-body--MaxWidth: calc(#{pf-size-prem(2000px)} - var(--pf-c-page__sidebar--Width));
 
   @media screen and (min-width: $pf-global--breakpoint--xl) {
     --pf-c-page__main-section--PaddingTop: var(--pf-c-page__main-section--xl--PaddingTop);
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -431,13 +431,11 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
 .pf-c-page__main-section,
 .pf-c-page__main-wizard {
   &.pf-m-limit-width {
-    display: flex;
-    flex-direction: column;
     padding: 0;
 
     .pf-c-page__main-body {
-      flex: 1;
       max-width: var(--pf-c-page--section--m-limit-width--MaxWidth);
+      height: 100%;
     }
   }
 }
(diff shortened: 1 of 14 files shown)


kybaker:
It looks good. One issue I see is the limit should be on all of the content below the masthead. It looks like the tabs for the third level nav in the example are not constrained. 

@mceledonia Please verify this is the content limit we want. I think we should be limiting the content to 1750px by default.

mcoker:
@kybaker @mceledonia from the design, it said 2000px overall width from the left edge of the window, so the calculation I used was `calc(#{pf-size-prem(2000px)} - var(--pf-c-page__sidebar--Width))`, which is 1710px. I can adjust that as needed.

Screen Shot 2020-07-29 at 1 00 22 PM

mcoker:
@kybaker @mcarrano updated this feature to also apply to the tertiary nav and breadcrumbs in the page demos, tertiary nav in the basic table demo, and those sections and the wizard in the in-page wizard demos. Updated the PR description and included links.

src/patternfly/components/Page/page.scss
```diff
@@ -407,17 +411,53 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
...
   }
 }
 
 .pf-c-page__main-wizard {
   flex: 1 0 0;
+  background-color: var(--pf-c-page__main-wizard--BackgroundColor);
   border-top: var(--pf-c-page__main-wizard--BorderTopWidth) solid var(--pf-c-page__main-wizard--BorderTopColor);
 }
 
+.pf-c-page__main-nav,
```
christiemolloy, line 429: what if there was a container called pf-c-page__limit-width that had no styles, but you could put any content inside of it and it would limit the width? e.g. what does someone do if they want to use a container that isnt `--nav`, `--breadcrumb`, `--section` or `--wizard` ? but they want to limit the width? do they have to use one of these containers and then remove the styles ? i guess pro to this would be flexibility in having to maintain this... con: people could limit the width of anything

src/patternfly/components/Page/page.scss
```diff
@@ -407,17 +411,53 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
...
+.pf-c-page__main-breadcrumb,
+.pf-c-page__main-section,
+.pf-c-page__main-wizard {
+  &.pf-m-limit-width {
+    display: flex;
+    flex-direction: column;
```
christiemolloy, line 435: would this have an impact on the elements that are inside of these containers ? what if they had two images in a flex row?

mcoker: Nope, all of the children of the `pf-c-page__main-[nav/section/breadcrumb/etc]` would be put in the `main-body` element, so no other children would be part of that flex layout. Though I just updated the PR to remove this flex styling and added `height: 100%` to the `main-body` element, which seems a bit safer. The main goal is just that if someone puts something with `height: 100%` in one of these sections, and applies `.pf-m-limit-width`, their thing will be 100% height - which means the `main-body` element also needs to be 100% height.

PR #3349, opened 2020-07-30
feat(button): add warning modifier to buttons

### LINKED ISSUE #3251: Warning modifier added to buttons
I've seen some examples of where an action has a warning notification tied to it 

**Having a primary action for something like this feels too positive**
Screen Shot 2020-07-06 at 4 22 54 PM
**Having a danger action feels a bit too intense**
Screen Shot 2020-07-06 at 4 23 35 PM
**Maybe a warning modifier would be most appropriate**
Screen Shot 2020-07-06 at 4 25 50 PM

I spoke with @mceledonia about this and he supports this :)

jenny-s51:
Closes #3251

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -67,12 +67,12 @@
   // Warning btn
   --pf-c-button--m-warning--BackgroundColor: var(--pf-global--warning-color--100);
   --pf-c-button--m-warning--Color: var(--pf-global--Color--dark-100);
-  --pf-c-button--m-warning--hover--BackgroundColor: var(--pf-global--warning-color--200);
-  --pf-c-button--m-warning--hover--Color: var(--pf-global--Color--light-100);
-  --pf-c-button--m-warning--focus--BackgroundColor: var(--pf-global--warning-color--200);
+  --pf-c-button--m-warning--hover--BackgroundColor: var(--pf-global--palette--gold-500);
+  --pf-c-button--m-warning--hover--Color: var(--pf-global--Color--dark-100);
+  --pf-c-button--m-warning--focus--BackgroundColor: var(--pf-global--palette--gold-500);
   --pf-c-button--m-warning--focus--Color: var(--pf-global--Color--light-100);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -64,6 +64,16 @@
   --pf-c-button--m-tertiary--active--after--BorderColor: var(--pf-global--Color--100);
   --pf-c-button--m-tertiary--active--Color: var(--pf-global--Color--100);
 
+  // Warning btn
+  --pf-c-button--m-warning--BackgroundColor: var(--pf-global--warning-color--100);
+  --pf-c-button--m-warning--Color: var(--pf-global--Color--dark-100);
+  --pf-c-button--m-warning--hover--BackgroundColor: var(--pf-global--warning-color--200);
```
mcoker, line 70: I checked with @mceledonia and he suggested using `--pf-global--palette--gold-500` for hover/focus/active background color instead, and leaving the text color as it is.

jenny-s51: Thank you Michael! Sounds good -- I'll make this change.

christiemolloy:
Looks like there are multiple styles for the states for the warning button, which makes for a jarring experience. Most buttons have an initial state and their hover/focus are the same, should we also do that for the warning button?

The initial state is:
Screen Shot 2020-08-17 at 4 24 15 PM

The hover state is:
Screen Shot 2020-08-17 at 4 23 39 PM

The focus state is:
Screen Shot 2020-08-17 at 4 23 44 PM

mcarrano:
Good catch @christiemolloy . Why would the label color invert on focus? Feels like a bug.

mcoker:
@christiemolloy good catch! I think the focus color is just an oversight. Confirmed with @mceledonia it should be the normal text color. @jenny-s51 when you get a chance, can you make that update?

jenny-s51:
> Thoughts? @mceledonia @mcoker @jenny-s51

nice catch @christiemolloy! fixed 👍

Issue #3350, opened 2020-07-30
Add pf-icon-bell


mcoker:
For use when we are using it with `pf-icon-attention-bell` interchangeably so that the icons align consistently.

PR #3351, opened 2020-07-30
fix(pficon): added pf-icon-bell

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3350


src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -71,7 +71,8 @@
     position: relative;
   }
 
-  .pf-icon-attention-bell {
+  .pf-icon-attention-bell,
```
christiemolloy, line 74: should you add a class for the icon and then put this icon inside of it?

PR #3353, opened 2020-07-31
feat(skeleton): add skeleton component

### LINKED ISSUE #3105: Skeleton loading

christiemolloy:
closes #3105 

Issue copied from old fork.

old PR #3320


src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,178 @@
...
+    content: "";
+  }
+
+  &::after {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: var(--pf-c-skeleton__element--after--ZIndex);
```
mcoker, line 66: do you need this?

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,178 @@
...
+    display: block;
+    width: 100%;
+    height: 100%;
+    content: "";
```
mcoker, line 70: this is fine, but we typically do `top: 0; right: 0; bottom: 0; left: 0;`. Since we don't have vars currently or if we used the coordinate properties, it doesn't really matter but wanted to mention it.

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,178 @@
...
+    width: var(--pf-c-skeleton__element--m-width-lg--Width);
+  }
+
+  &.pf-m-width-sm,
+  &.pf-m-width-md,
+  &.pf-m-width-lg {
+    --pf-c-skeleton__element--Height: none;
+
+    &::before {
+      padding-bottom: 100%;
```
mcoker, line 100: should this be a var? Then you could define the var under the modifier, and not need to nest `::before`, I think...

src/patternfly/components/Skeleton/skeleton.scss
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  &.pf-m-width-lg {
+    width: var(--pf-c-skeleton__element--m-width-lg--Width);
+  }
+
+  &.pf-m-width-sm,
+  &.pf-m-width-md,
+  &.pf-m-width-lg {
+    --pf-c-skeleton__element--Height: none;
```
mcoker, line 97: Another way of doing this would be to let width set width, and let height set height, and have `pf-m-circle` and `pf-m-square` unset the height and use `padding-bottom: 100%`

src/patternfly/components/Skeleton/skeleton.scss
christiemolloy: I removed width and height of 100%

src/patternfly/components/Skeleton/skeleton.scss
christiemolloy: that would mean having the modifier `pf-m-square` though.. not sure we need to add that

PR #3357, opened 2020-07-31
chore(repo): update stalebot ignored labels

### LINKED ISSUE #3356: chore(repo): update stalebot ignored labels
Add `enhancement` and `Feature` labels to the list of labels the stalebot tool will ignore, so that these issues never get marked stale & automatically closed out.

evwilkin:
closes #3356 

Add enhancement & Feature labels to stalebot ignore

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -12,6 +12,8 @@ exemptLabels:
   - security
   - accessibility
   - "breaking change :boom:"
+  - enhancement
+  - Feature
 
 # Label to use when marking as stale
 staleLabel: wontfix
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -9,7 +9,7 @@ daysUntilClose: 30
 # Issues with these labels will never be considered stale
 exemptLabels:
   - accessibility
-  - "breaking changes"
+  - "breaking change"
   - design approved
 
 # Label to use when marking as stale
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -8,12 +8,9 @@ daysUntilClose: 30
 
 # Issues with these labels will never be considered stale
 exemptLabels:
-  - pinned
-  - security
   - accessibility
-  - "breaking change :boom:"
-  - enhancement
-  - Feature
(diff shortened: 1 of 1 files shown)


mcoker:
Users often add these labels to issues they create since they're generic - do we want to allow for that? What if we had something like "design approved" that could be applied to an enhancement or feature issue (and maybe other things).

Also, scope creep, but can we also make these changes?
* remove "pinned" and "security" since core doesn't have those
* update the breaking changes label to "breaking changes" - we don't have the 💥

evwilkin:
@mcarrano follow-up from our earlier discussion - I'd like to get your feedback on @mcoker's points above on using a `design approved` label in place of adding `enhancement` and `Feature` labels to the ignore list for stalebot.

mcarrano:
@evwilkin @mcoker I'm OK with whatever works for you guys.

evwilkin:
@mcoker updated per your feedback above

.github/stale.yml
```diff
@@ -8,10 +8,9 @@ daysUntilClose: 30
 
 # Issues with these labels will never be considered stale
 exemptLabels:
-  - pinned
-  - security
   - accessibility
-  - "breaking change :boom:"
+  - "breaking changes"
```
mcoker, line 12: double check me, but pretty sure this is the label....

```suggestion
  - "breaking change"
```

evwilkin:
@mcoker you're right, updated from `breaking changes` to `breaking change`

.github/stale.yml
```diff
@@ -8,10 +8,9 @@ daysUntilClose: 30
 
 # Issues with these labels will never be considered stale
 exemptLabels:
-  - pinned
-  - security
   - accessibility
-  - "breaking change :boom:"
+  - "breaking change"
+  - design approved
```
mcoker, line 13: do we need quotes? `"breaking change"` has quotes but `design approved` doesn't.

evwilkin: There should not be quotes - removed these

PR #3359, opened 2020-08-03
fix(release): added github plugin

redallen:
Hopefully will leave a comment on PRs. Will also create a github release.


redallen:
Looks like I need to fix the token used to be patternfly-build instead of my account.

I'll do that after the org redesign release.

Issue #3368, opened 2020-08-03
Selected page header item styles overriding notification badge state styles


mcoker:
When the notification badge is in either the unread or attention state and is expanded via `pf-m-selected` on the parent header tools item, the notification badge state background isn't being shown.

Issue #3371, opened 2020-08-03
Ensure notification badge and notification drawer item state is used correctly in demos


mcoker:
For example, the drawer demo does not contain a notification drawer. Should we update that to be some other type of icon that could toggle a drawer open/closed?

The notification drawer demos all use the read notification badge state, but have unread (and potentially attention state) items. Let's ensure we're showing the correct items in the notification drawer, and the notification badge used in the demo reflects the drawer contents.

Also review anywhere else this may be an issue.

mcarrano:
@mcoker does this issue require any React follow-up?

mcoker:
@mcarrano this only updated our examples so the badge was appropriate based on the contents of the notification drawer. As long as the notification badge demos in react look ok to you, it should be fine - https://patternfly-react.surge.sh/components/notification-drawer/react-demos

We did include a demo for the attention state. Maybe the react demos could be updated to add the attention state when there is an unread attention state item?

PR #3372, opened 2020-08-03
fix(page): fixed notification badge state and header item conflict

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3368


mcoker:
@nicolethoen we have another issue here to review the usage of the notification badge in examples - https://github.com/patternfly/patternfly/issues/3371

PR #3355, opened 2020-07-31
feat(togglegroup): add toggle group component 

christiemolloy:
closes #2625 

new PR from new forked repo

old PR #3295

PR #3373, opened 2020-08-04
fix(togglegroup): add support for icon + text

christiemolloy:
Follow-up to #3355 to support icon + text in the toggle group button


src/patternfly/components/ToggleGroup/toggle-group.scss
```diff
@@ -79,3 +85,11 @@
     pointer-events: none;
   }
 }
+
+.pf-c-toggle-group__icon + .pf-c-toggle-group__text {
+  margin-left: var(--pf-c-toggle-group__icon--text--MarginLeft);
+}
+
```
mcoker, line 92: This looks good, though I wonder if we would ever want the space to be different depending on which comes first. It seems like we wouldn't, so you could also combine the selectors and apply a single `margin-left` declaration. wdyt?

christiemolloy: @mcoker yeah agreed i think we would want the spacing to be the same. what do you mean by "combine the selectors and apply a single margin-left declaration"?

Issue #3381, opened 2019-09-16
Add spinner to button component


matthewcarleton:
We've seen a couple use cases for this on the OpenShift side so it seems reasonable to offer it via PatternFly.

In order to provide this we'd need a button variation that has the spinner component included.

<img width="234" alt="Screen Shot 2019-09-16 at 11 21 07 AM" src="https://user-images.githubusercontent.com/287880/64965926-3ba53700-d874-11e9-8fa2-fb63f837d366.png">

I'm not sure if we'd also now provide a cancel action for this or if that would be handled on the product side. Worth a conversation though.

This is related to https://github.com/patternfly/patternfly-react/issues/2824

matthewcarleton:
@mcarrano I believe we'd run into spacing issues with that approach if I recall correctly.

PR #3382, opened 2020-08-05
feat(button): added progress button

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3381

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -151,17 +151,17 @@
   --pf-c-button__icon--m-end--MarginLeft: var(--pf-global--spacer--xs);
 
   // Progress
-  --pf-c-button__progress-icon--width: calc(var(--pf-global--icon--FontSize--md) + var(--pf-global--spacer--sm)); // matches medium spinner diameter plus a spacer
-  --pf-c-button__progress-icon--Opacity: 0;
-  --pf-c-button__progress-icon--TranslateY: -50%;
-  --pf-c-button__progress-icon--Top: 50%;
-  --pf-c-button__progress-icon--Left: var(--pf-global--spacer--md);
+  --pf-c-button__progress--width: calc(var(--pf-global--icon--FontSize--md) + var(--pf-global--spacer--sm)); // matches medium spinner diameter plus a spacer
+  --pf-c-button__progress--Opacity: 0;
+  --pf-c-button__progress--TranslateY: -50%;
+  --pf-c-button__progress--Top: 50%;
+  --pf-c-button__progress--Left: var(--pf-global--spacer--md);
   --pf-c-button--m-progress--TransitionProperty: padding;
   --pf-c-button--m-progress--TransitionDuration: var(--pf-global--TransitionDuration);
-  --pf-c-button--m-progress--PaddingRight: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width) / 2);
-  --pf-c-button--m-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width) / 2);
+  --pf-c-button--m-progress--PaddingRight: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress--width) / 2);
+  --pf-c-button--m-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress--width) / 2);
   --pf-c-button--m-in-progress--PaddingRight: var(--pf-global--spacer--md);
-  --pf-c-button--m-in-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width));
+  --pf-c-button--m-in-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress--width));
 
   position: relative;
   display: inline-block;
@@ -515,12 +515,12 @@
   }
 }
 
-.pf-c-button__progress-icon {
+.pf-c-button__progress {
   position: absolute;
-  top: var(--pf-c-button__progress-icon--Top);
-  left: var(--pf-c-button__progress-icon--Left);
+  top: var(--pf-c-button__progress--Top);
+  left: var(--pf-c-button__progress--Left);
   line-height: 1;
-  transform: translateY(var(--pf-c-button__progress-icon--TranslateY));
+  transform: translateY(var(--pf-c-button__progress--TranslateY));
 
   .pf-c-spinner {
     --pf-c-spinner--Color: currentColor;
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -150,6 +150,19 @@
   --pf-c-button__icon--m-start--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-button__icon--m-end--MarginLeft: var(--pf-global--spacer--xs);
 
+  // Progress
+  --pf-c-button__progress-icon--width: calc(var(--pf-global--icon--FontSize--md) + var(--pf-global--spacer--sm)); // matches medium spinner diameter plus a spacer
+  --pf-c-button__progress-icon--Opacity: 0;
+  --pf-c-button__progress-icon--TranslateY: -50%;
+  --pf-c-button__progress-icon--Top: 50%;
+  --pf-c-button__progress-icon--Left: var(--pf-global--spacer--md);
+  --pf-c-button--m-progress--TransitionProperty: padding;
+  --pf-c-button--m-progress--TransitionDuration: var(--pf-global--TransitionDuration);
+  --pf-c-button--m-progress--PaddingRight: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width) / 2);
+  --pf-c-button--m-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width) / 2);
+  --pf-c-button--m-in-progress--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-button--m-in-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width));
+
   position: relative;
   display: inline-block;
   padding: var(--pf-c-button--PaddingTop) var(--pf-c-button--PaddingRight) var(--pf-c-button--PaddingBottom) var(--pf-c-button--PaddingLeft);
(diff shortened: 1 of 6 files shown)


matthewcarleton:
> The new Spinner variation looks good to me. The only question I have is whether we need to add a button example of this to address the original issue or would this be more like a demo?
I don't think this PR fully addresses #3381 does it? The button component is not currently built to take the spinner as a drop in is?
https://recordit.co/P1cVD8O6zq

mcoker:
@matthewcarleton I ran this demo by @mcarrano yesterday prior to putting up the PR and he said it looked as expected -  https://codepen.io/mcoker/pen/abNbjpx

Looks like the difference is the spinner size - I'm using the small variation in this demo.

matthewcarleton:
> @matthewcarleton I ran this demo by @mcarrano yesterday prior to putting up the PR and he said it looked as expected - https://codepen.io/mcoker/pen/abNbjpx
> 
> Looks like the difference is the spinner size - I'm using the small variation in this demo.

ah interesting. Ya that small spinner felt too small to me. Has this had any visual design review? It could be worth getting it reviewed from that perspective. It feels a bit odd that we would require the user to only use the small spinner for this not to break.

mcarrano:
But wouldn't a larger spinner increase the height of the button? Seems like that would cause different problems. @mceledonia it would be great to have your input here, also.

matthewcarleton:
> But wouldn't a larger spinner increase the height of the button? Seems like that would cause different problems. @mceledonia it would be great to have your input here, also.

Yes anything other than the sm spinner would "break" this. IMO even the small spinner spacing looks a bit odd too.

mcoker:
Since the medium spinner is taller than the regular font-size, the button is taller with the medium spinner in it, just like a button would be with an increased font-size. 

I've updated the pen with a button variation that lets the icon element in the button be as tall as the button itself, without making the button bigger. I included an example with the medium spinner, and a 24px icon as examples - https://codepen.io/mcoker/pen/abNbjpx?editors=1100

mceledonia:
I think the solution @mcoker just posted looks great, we need to be careful that this is only utilized for spinner size and not for icons though. I think the medium size spinner looks a lot more natural there than the small size.

mcoker:
@mcarrano @mceledonia @matthewcarleton I went back and added this as a feature of the button component instead. The button now supports a "loading" variant of the icon element that you place a spinner inside of. Since this is a variant of the existing button icon element, you can align the spinner to the left or right of the button text and existing icons, and it uses the same spacer as as the existing icon element.

Added a bunch of examples of regular & small buttons at the bottom of the button examples to show what it looks like. If we go this route, I'll clean up those examples and only list a couple of buttons.

https://patternfly-pr-3382.surge.sh/documentation/core/components/button#loading

mcarrano:
The examples look good @mcoker . I only have a question about the intended usage of this. As I understand this, the idea would be that the user clicks the button and it triggers the spinner while an action takes place. The spinner would then disappear after the action completes. Do I have that right @matthewcarleton ? So my question is does the button grow horizontally to accommodate the spinner or is the space allocated and the text just pushes over (but the button remains a fixed width? Not sure exactly what to expect, but I am concerned about the button pushing other elements to the right when the spinner is inserted.

mcoker:
I've updated this codepen to show what it looks like when there is space allocated for the spinner, just click on the button to show the loader - https://codepen.io/mcoker/pen/abNbjpx

mcarrano:
I have to admit that it feels odd to have the empty space reserved. Is it possible for the label to be centered and then push over when the spinner is introduced? I realize this might need to be part of the JS implementation. Or alternatively, would we always recommend that the label be proceeded with an icon that is replaced by the spinner?

mceledonia:
@mcarrano @mcoker I agree about the space allocation being odd. I think I would prefer Matt Carrano's suggestion of the text centering then shifting around first, then if that is too difficult I'd go with the button itself shifting as shown in the "no space for spinner" version in the codepen. 

It is possible some transition animation could make that expansion event a bit more smooth and less jarring.

matthewcarleton:
could we do something like this? @mcoker 
https://xd.adobe.com/view/3d14f61d-109e-4551-9140-197dc60692e3-6847/
I think the key is in the layout to ensure things can make room for elements growing/shrinking right?

mceledonia:
That seems like a good solution, another idea is rather than adding space to the button itself we add it as padding to the container, so that it then grows into that padding when the button expands. But yes I think the main consideration is ensuring the layout doesn't shift around when the spinner appears.

matthewcarleton:
> That seems like a good solution, another idea is rather than adding space to the button itself we add it as padding to the container, so that it then grows into that padding when the button expands. But yes I think the main consideration is ensuring the layout doesn't shift around when the spinner appears.

ah so it expands to the right instead of the left and keeps the input the same? If that's what you mean I wonder if that would cause some odd layout issues when things need to align right so things in rows above/below need to be flush right with the  right of the button 🤔

mcoker:
Updated the pen with the centered layout. https://codepen.io/mcoker/pen/abNbjpx

matthewcarleton:
oh ya - that centred layout is pretty slick. So we would just animate the spinner coming in but the button itself wouldn't change?
The one thing that could still add some layout complexity is if the text in the button needs to change with the spinner. So a scenario where the button reads "Check" and when clicked it changes to "Checking" with the spinner.

mcoker:
> The one thing that could still add some layout complexity is if the text in the button needs to change with the spinner. So a scenario where the button reads "Check" and when clicked it changes to "Checking" with the spinner.

If the button can just change size to accommodate the text, that should work fine with the centering. I think it's fair to expect the button to change size with different content. Though we could do something like make the button as big as it will be with either the default and loading text by default. Added a "width change" example - https://codepen.io/mcoker/pen/abNbjpx?editors=1100

src/patternfly/components/Button/button.scss
```diff
@@ -455,6 +457,19 @@
...
+  &.pf-m-loading {
+    display: inline-block;
+    margin-top: var(--pf-c-button__icon--m-loading--MarginTop);
+    margin-bottom: var(--pf-c-button__icon--m-loading--MarginBottom);
+    line-height: 1;
+    vertical-align: middle;
+    transform: translateY(-10%);
+
+    .pf-c-spinner {
+      --pf-c-spinner--Color: currentColor;
```
mattnolting, line 470: 🔥

mcoker:
@mceledonia @mcarrano @matthewcarleton just wanted to follow up - is there anything else you'd like to see or do you feel like we're ready for an implementation?

mcarrano:
Looks good to me, except that I'm unclear how we are handling the spacing issue. Or is that left to the React component to adjust the spacing on_Click. I also wonder if we should rename the example from "Loading" to "Progress." All this is indicating is something in progress. Could be loading or otherwise.

mcoker:
If the demo looks good, then we would just have a flag in the button that says it's a progress/loading button, and that flag will create extra space in the component for the (visually hidden) spinner to show, then we'll add an additional state in the CSS for when the button is in the loading state that shows the spinner without changing the width of the button. That loading state would be managed in the react component and triggered by the developer.

matthewcarleton:
Is it just the wrong spinner being used in the last two?
Screen Shot 2020-08-20 at 11 37 19 AM

mcoker:
> Is it just the wrong spinner being used in the last two?

I think I broke the codepen :) @mcarrano @matthewcarleton @mceledonia updated the PR so the PR preview works - https://patternfly-pr-3382.surge.sh/documentation/core/components/button#loading

A "loader" button will always be as wide as it needs to be with the spinner, but the spinner is not visible by default. When the loader button is in the "loading" state, the spinner appears (no animation), and the button text animates to move over to make space for the spinner

Aug-24-2020 15-20-31

mcarrano:
Looks great @mcoker . Did you consider my question re naming of this? I thought that "Loading" was too specific and maybe call it "Progress" or "Processing?" That would account for any action.

mcoker:
@mcarrano sorry I missed that. Updated this to `pf-m-progress` to indicate it's a progress button, and `pf-m-in-progress` as the state when the icon shows.

src/patternfly/components/Button/examples/Button.md
```diff
@@ -295,3 +315,5 @@ Semantic buttons and links are important for usability as well as accessibility.
 | `.pf-m-small` | `.pf-c-button` | Modifies the button so that it has small font size. |
 | `.pf-m-aria-disabled` | `.pf-c-button` | Modifies a button to be visually disabled, yet is still focusable. |
 | `.pf-m-display-lg` | `.pf-c-button`, `pf-c-button.pf-m-link` | Modifies the button and link button for large display styling. For example, use this modifier to achieve "call to action" styles. |
+| `.pf-m-loader` | `.pf-c-button` | Indicates that the button supports the loading state. |
```
christiemolloy, line 318: I think this should be `pf-m-progress` and `pf-m-progressing`

src/patternfly/components/Button/examples/Button.md
```diff
@@ -279,6 +298,7 @@ Semantic buttons and links are important for usability as well as accessibility.
 | -- | -- | -- |
 | `.pf-c-button` | `<button>` |  Initiates a button. Always use it with a modifier class. **Required** |
 | `.pf-c-button__icon` | `<span>` | Initiates a button icon. |
+| `.pf-c-button__loader` | `<span>` | Initiates a button loader container. |
```
christiemolloy, line 301: and change the name here too, instead of `button__progress-icon` should it be `button__progress` the icon throws me off because its a spinner ant not an icon

christiemolloy (changes requested):

mcoker:
@christiemolloy updated

src/patternfly/components/Button/button.scss
```diff
@@ -151,17 +151,17 @@
   --pf-c-button__icon--m-end--MarginLeft: var(--pf-global--spacer--xs);
 
   // Progress
-  --pf-c-button__progress-icon--width: calc(var(--pf-global--icon--FontSize--md) + var(--pf-global--spacer--sm)); // matches medium spinner diameter plus a spacer
-  --pf-c-button__progress-icon--Opacity: 0;
-  --pf-c-button__progress-icon--TranslateY: -50%;
-  --pf-c-button__progress-icon--Top: 50%;
-  --pf-c-button__progress-icon--Left: var(--pf-global--spacer--md);
+  --pf-c-button__progress--width: calc(var(--pf-global--icon--FontSize--md) + var(--pf-global--spacer--sm)); // matches medium spinner diameter plus a spacer
```
christiemolloy, line 154:
```suggestion
  --pf-c-button__progress--Width: calc(var(--pf-global--icon--FontSize--md) + var(--pf-global--spacer--sm)); // matches medium spinner diameter plus a spacer
```

src/patternfly/components/Button/button.scss
```diff
@@ -151,17 +151,17 @@
...
+  --pf-c-button__progress--Opacity: 0;
+  --pf-c-button__progress--TranslateY: -50%;
+  --pf-c-button__progress--Top: 50%;
+  --pf-c-button__progress--Left: var(--pf-global--spacer--md);
   --pf-c-button--m-progress--TransitionProperty: padding;
   --pf-c-button--m-progress--TransitionDuration: var(--pf-global--TransitionDuration);
-  --pf-c-button--m-progress--PaddingRight: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width) / 2);
-  --pf-c-button--m-progress--PaddingLeft: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress-icon--width) / 2);
+  --pf-c-button--m-progress--PaddingRight: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress--width) / 2);
```
christiemolloy, line 161:
```suggestion
  --pf-c-button--m-progress--PaddingRight: calc(var(--pf-global--spacer--md) + var(--pf-c-button__progress--Width) / 2);
```

christiemolloy (changes requested):

christiemolloy (approved):
nice !

PR #3383, opened 2020-08-05
fix(page): add href to page-header-brand-link

### LINKED ISSUE #3332: UX Bug: PageHeader wraps logo in a tag without an href, CSS makes link a pointer

Logos in the nav bars are usually links, the PageHeader Component will make the logo a link without an href, and CSS added by Patternfly will make all `<a>` tags user the pointer.

https://github.com/patternfly/patternfly-react/blob/2e2676ce6bc34f5e20231ca066d26ea4c3ac95e4/packages/react-core/src/components/Page/PageHeader.tsx#L42

Not sure where the CSS is coming from, but this is the CSS making a mouse use a pointer on all `<a>` tags (even if they don't have an href):
```css
button, a {
    cursor: pointer;
}
```

My thought would be to only make the logo an `a` tag if there's an href. In this context an `<a>` tag without an `href` doesn't make much sense.

Alternatively the CSS selector could be changed to `a[href]`, instead of just `a`.

Use PageHeader component and don't specify a logo URL

Bug, it's creating a UX issue

jenny-s51:
Closes #3332

### DIFF
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -144,7 +144,7 @@ wrapperTag: div
       {{#> page-header-brand-toggle}}
         toggle
       {{/page-header-brand-toggle}}
-      {{#> page-header-brand-link}}
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
         Logo
       {{/page-header-brand-link}}
     {{/page-header-brand}}
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -15,7 +15,7 @@ wrapperTag: div
       {{#> page-header-brand-toggle}}
         toggle
       {{/page-header-brand-toggle}}
-      {{#> page-header-brand-link}}
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
         Logo
       {{/page-header-brand-link}}
     {{/page-header-brand}}
@@ -45,7 +45,7 @@ wrapperTag: div
   {{#> page-header}}
     {{!-- Brand --}}
     {{#> page-header-brand}}
-      {{#> page-header-brand-link}}
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
         Logo
       {{/page-header-brand-link}}
     {{/page-header-brand}}
@@ -77,7 +77,7 @@ wrapperTag: div
   {{#> page-header}}
     {{!-- Brand --}}
     {{#> page-header-brand}}
-      {{#> page-header-brand-link}}
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
         Logo
       {{/page-header-brand-link}}
     {{/page-header-brand}}
@@ -111,7 +111,7 @@ wrapperTag: div
       {{#> page-header-brand-toggle}}
         toggle
       {{/page-header-brand-toggle}}
-      {{#> page-header-brand-link}}
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
         Logo
       {{/page-header-brand-link}}
     {{/page-header-brand}}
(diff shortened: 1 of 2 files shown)


mcoker:
Thanks @jenny-s51! We should also add a note to the [page docs](https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/Page/examples/Page.md) for `.pf-c-page__header-brand-link` and add `<span>` under the "applied to" column, and expand on the description and say something like "Use a `<span>` if there is no link."

src/patternfly/components/Page/page-header-brand-link.hbs
```diff
@@ -1,4 +1,4 @@
-<a class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
+<a href="{{page-header-brand-link--href}}" class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
```
mattnolting, line 1:
```suggestion
<a href="{{#if page-header-brand-link--href}}{{page-header-brand-link--href}}{{else}}#{{/if}}{{page-header-brand-link--href}}" class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
```

mattnolting: What do you think about making `#` the default and allowing it to be updated if needed?

christiemolloy: Thats a good idea

christiemolloy: Should we also add a condition for if there is no link, then instead of an `<a>` its a `<span>`

src/patternfly/components/Page/page-header-brand-link.hbs
```diff
@@ -1,6 +1,6 @@
-<a href="{{page-header-brand-link--href}}" class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
+<{{#if page-header-brand-link--href}}a href="{{#if page-header-brand-link--href}}{{page-header-brand-link--href}}{{else}}#{{/if}}" {{else}}span {{/if}} class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
```
mcoker, line 1: won't this inner test will always evaluate true since it's the same test as the outer test?

jenny-s51: good point @mcoker - fixed

src/patternfly/components/Page/page-header-brand-link.hbs
```diff
@@ -1,6 +1,6 @@
-<a class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
+<{{#if page-header-brand-link--href}}a href="{{page-header-brand-link--href}}" {{else}}span {{/if}} class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
```
mcoker, line 1:
```suggestion
<{{#if page-header-brand-link--href}}a href="{{page-header-brand-link--href}}"{{else}}span{{/if}} class="pf-c-page__header-brand-link{{#if page-header-brand-link--modifier}} {{page-header-brand-link--modifier}}{{/if}}"
```

src/patternfly/components/Page/examples/Page.md
```diff
@@ -15,7 +15,7 @@ wrapperTag: div
       {{#> page-header-brand-toggle}}
```
mcoker, line 15: looks like there is one more we need to add the href attr to for a variation that was just added under main section variations - https://github.com/patternfly/patternfly/blob/582d77fd842592483768ceed275175b89bb533dc/src/patternfly/components/Page/examples/Page.md#main-section-variations

Issue #3375, opened 2020-08-04
Add width var to skeleton so it's easily themeable


mcoker:
We're declaring the `width` property for each width modifier. Let's introduce a var for that so that react can define an inline custom property to override `--Width`

mcoker:
closed via https://github.com/patternfly/patternfly/pull/3385

PR #3385, opened 2020-08-06
fix(skeleton): added width var, safari fix bug w/ transparent

mcoker:
https://github.com/patternfly/patternfly/issues/3375

found a bug where safari renders `transparent` as `rgba(0,0,0,0)`, a transparent based on black (`rgb(0,0,0)`) and the gradient was the wrong color. https://stackoverflow.com/questions/38391457/linear-gradient-to-transparent-bug-in-latest-safari


mcoker:
@jenny-s51 nice catch! Updated.

Issue #3376, opened 2020-08-04
Introduce Label group component


mcarrano:
Introduces a label group similar to a chip group. Designs here: https://marvelapp.com/prototype/5ajfj8f

cc @mcoker

Issue #3182, opened 2020-06-16
ChipGroups: close chip group button wraps at smaller widths


nicolethoen:
Reproducible in the React Toolbar documentation for the Toolbar with filters example.

![image](https://user-images.githubusercontent.com/26305082/84826508-08edc200-aff1-11ea-945c-3e4fd1c6dd84.png)

![image](https://user-images.githubusercontent.com/26305082/84826308-c330f980-aff0-11ea-884a-d7fe80e58f20.png)

mcoker:
cc @mcarrano

mcarrano:
@nicolethoen @mcoker Is it possible to make this so the chip group collapses when it runs out of space like this rather than wraps?

nicolethoen:
@mcarrano There is already a breakpoint that exists that will default to 'xl' but that the consumer can control that chooses when the chip groups all collapse.

mceledonia:
Just chiming in here - I think our solution should be in the realm of the X / clear actions always being given priority over chips, i.e. chips should start dropping/truncating rather than the close actions moving.

PR #3396, opened 2020-08-12
feat(label-group): added label group component, overflow label

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3376

@maryshak1996 let me know what you would like to do about a removable vertical label group

### DIFF
--- a/src/patternfly/components/LabelGroup/examples/LabelGroup.md
+++ b/src/patternfly/components/LabelGroup/examples/LabelGroup.md
@@ -260,7 +260,7 @@ cssPrefix: pf-c-label-group
       {{/label}}
     {{/label-group-list-item}}
     {{#> label-group-list-item}}
-      {{#> label label--modifier="pf-m-overflow"}}
+      {{#> label label--IsOverflow="true"}}
         3 more
       {{/label}}
     {{/label-group-list-item}}
@@ -313,7 +313,7 @@ cssPrefix: pf-c-label-group
       {{/label}}
     {{/label-group-list-item}}
     {{#> label-group-list-item}}
-      {{#> label label--modifier="pf-m-overflow"}}
+      {{#> label label--IsOverflow="true"}}
         3 less
       {{/label}}
     {{/label-group-list-item}}
--- a/src/patternfly/components/LabelGroup/examples/LabelGroup.md
+++ b/src/patternfly/components/LabelGroup/examples/LabelGroup.md
@@ -1,0 +1,375 @@
+---
+id: 'Label group'
+beta: true
+section: components
+cssPrefix: pf-c-label-group
+---
+
+## Examples
+### Basic
+```hbs
+{{#> label-group label-group--id="label-group-basic"}}
+  {{#> label-group-list label-group-list--attribute='aria-label="Label group list"'}}
+    {{#> label-group-list-item}}
+      {{#> label}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-blue"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 2
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-green"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 3
+      {{/label}}
+    {{/label-group-list-item}}
+  {{/label-group-list}}
+{{/label-group}}
+```
+
+### Overflow
+```hbs
+{{#> label-group label-group--id="label-group-overflow"}}
+  {{#> label-group-list label-group-list--attribute='aria-label="Label group list"'}}
+    {{#> label-group-list-item}}
+      {{#> label}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-blue"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 2
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-green"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 3
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--IsOverflow="true"}}
+        3 more
+      {{/label}}
+    {{/label-group-list-item}}
+  {{/label-group-list}}
+{{/label-group}}
+```
+
+### Overflow expanded
+```hbs
+{{#> label-group label-group--id="label-group-overflow-expanded"}}
+  {{#> label-group-list label-group-list--attribute='aria-label="Label group list"'}}
+    {{#> label-group-list-item}}
+      {{#> label}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-blue"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 2
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-green"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 3
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-cyan"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 3
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-purple"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 3
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--IsOverflow="true"}}
+        3 less
+      {{/label}}
+    {{/label-group-list-item}}
+  {{/label-group-list}}
+{{/label-group}}
+```
+
+### Category
+```hbs
+{{#> label-group label-group--id="label-group-category" label-group--modifier="pf-m-category"}}
+  {{#> label-group-label label-group-label--attribute=(concat 'id="' label-group--id '-label"')}}
+    Group label
+  {{/label-group-label}}
+  {{#> label-group-list label-group-list--attribute=(concat 'aria-labelledby="' label-group--id '-label"')}}
+    {{#> label-group-list-item}}
+      {{#> label}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-blue"}}
+        {{#> label-icon}}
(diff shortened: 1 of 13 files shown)


maryshak1996:
@mcoker looking good! 
the only thing that I think that we are missing is that grey outline around the vertical variants -- having that will make it logical to have the 'times-circle' at the end to clear all of the labels

mcoker:
@christiemolloy yeah I went back and forth on this on whether the `pf-c-label` should be a `button` if it's an overflow chip, or if it should share the same basic feature that makes the text in the button an interactive element when the label serves as a link. To keep it simple, I just extended the existing behavior to let the text apply to buttons, as well, which means interactive elements could also be added to an overflow label later if needed.

Here's the link behavior, and why the outline displays the way it does.

Screen Shot 2020-08-18 at 10 47 24 AM

Do you think the label itself would be better as a button? Another option here could be to change the outline offset.

christiemolloy:
> @christiemolloy yeah I went back and forth on this on whether the `pf-c-label` should be a `button` if it's an overflow chip, or if it should share the same basic feature that makes the text in the button an interactive element when the label serves as a link. To keep it simple, I just extended the existing behavior to let the text apply to buttons as well, which means interactive elements could also be added to an overflow label later if needed.
> 
> Here's the link behavior, and why the outline displays the way it does.
> 
> Screen Shot 2020-08-18 at 10 47 24 AM
> 
> Do you think the label itself would be better as a button? Another option here could be to change the outline offset.

Hmm I see the label component as one whole thing even though it has parts because to me the background color is part of the label text. So I see the parent element as the button. Since this is an overflow button i cant really see another button being added to the label, but if you think its safest to keep the button just to the text, then i think changing the outline-offset would be a good idea.

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -0,0 +1,402 @@
...
+beta: true
+section: components
+cssPrefix: pf-c-label-group
+---
+
+## Examples
+### Basic
+```hbs
+{{#> label-group label-group--id="label-group-basic"}}
+  {{#> label-group-list label-group-list--attribute='aria-label="Label group list"'}}
```
christiemolloy, line 12: Group of labels

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -0,0 +1,402 @@
...
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-blue"}}
+        {{#> label-icon}}
+          <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+        {{/label-icon}}
+        Label 2
+      {{/label}}
+    {{/label-group-list-item}}
+    {{#> label-group-list-item}}
+      {{#> label label--modifier="pf-m-green"}}
```
christiemolloy, line 30: should the last label group item have no margin-right
Screen Shot 2020-08-18 at 1 18 25 PM

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -0,0 +1,98 @@
+.pf-c-label-group {
```
christiemolloy, line 1: should there be truncation when the labels are too long ? cc: @mcarrano @mceledonia 
Screen Shot 2020-08-18 at 1 21 07 PM

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -0,0 +1,98 @@
...
+  // label group list
+  --pf-c-label-group__list--MarginBottom: calc(var(--pf-global--spacer--xs) * -1);
```
christiemolloy, line 3: same here

Screen Shot 2020-08-18 at 1 24 52 PM

src/patternfly/components/LabelGroup/examples/LabelGroup.md
christiemolloy: as a suggestion for the aria-label

christiemolloy:
@maryshak1996 what do you think about moving the close button for the vertical view to the top right corner ? just seems a little odd being at the very bottom

Screen Shot 2020-08-18 at 1 25 53 PM

mcarrano:
> > @christiemolloy yeah I went back and forth on this on whether the `pf-c-label` should be a `button` if it's an overflow chip, or if it should share the same basic feature that makes the text in the button an interactive element when the label serves as a link. To keep it simple, I just extended the existing behavior to let the text apply to buttons as well, which means interactive elements could also be added to an overflow label later if needed.
> > Here's the link behavior, and why the outline displays the way it does.
> > Screen Shot 2020-08-18 at 10 47 24 AM
> > Do you think the label itself would be better as a button? Another option here could be to change the outline offset.
> 
> Hmm I see the label component as one whole thing even though it has parts because to me the background color is part of the label text. So I see the parent element as the button. Since this is an overflow button i cant really see another button being added to the label, but if you think its safest to keep the button just to the text, then i think changing the outline-offset would be a good idea.

This is an unfortunate artifact of the heavy focus rectangle. I wouldn't change the structure, it really is the link and not the entire label that's focusable. Maybe add some padding so it does not look as bad?

mcoker:
> it really is the link and not the entire label that's focusable

@mcarrano thanks for the feedback. What about the overflow label? Is it the label that is focusable/clickable? Or is it the same as the screenshot above where the content in the label is clickable?

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -0,0 +1,98 @@
...
+  align-items: center;
+}
+
+.pf-c-label-group__list-item {
+  display: inline-flex;
+  margin-right: var(--pf-c-label-group__list-item--MarginRight);
+  margin-bottom: var(--pf-c-label-group__list-item--MarginBottom);
+}
+
+.pf-c-label-group__label {
```
christiemolloy, line 84: i think its confusing that this is called `label-group__label` , since its unlike the actual label component, maybe it could be called `text` ?

mcoker: I'm in favor of keeping it consistent unless there is a good reason not to. I think the element "label" applies to the group, so it's fine as the group's label. Alternatively I would also be in favor of this and the chip group renaming `__label` to `__title`, as that seems a little more appropriate. I don't think we should call it text, and that's usually just the text node of some other element (as in we would have `.pf-c-label-group__label > .pf-c-label-group__label-text`). WDYT, still want to change it?

src/patternfly/components/LabelGroup/examples/LabelGroup.md
mcoker: Sure, though for what it's worth, this matches the chip group. Do you think they need to be different?

mattnolting:
I know we don't have a feature in `pf-c-chip-group` for this, but we have an [existing bug](https://github.com/patternfly/patternfly/issues/3182) that could be addressed here and added to `pf-c-chip-group` later (or at the same time). 

We currently don't have a method to control when label and items stack or when the close button wraps. When multiple `group`s appear in columns, they look to be inconsistent. 

Possible addition
- If we wrap `pf-c-label-group__list` and `pf-c-label-group__close` with `pf-c-label-group__main`, we could add `pf-m-nowrap` or `pf-m-wrap` to `pf-c-label-group__main` (new wrapper) and handle multiple label groups per view consistently. 
- If we add `pf-m-stack` and `pf-m-inline` (or something like that) as modifiers for `pf-c-label-group` we could allow the user to control when they stack and when they appear inline. 

cc @maryshak1996 [existing bug](https://github.com/patternfly/patternfly/issues/3182)

@mcoker wdyt?

mcarrano:
> > it really is the link and not the entire label that's focusable
> 
> @mcarrano thanks for the feedback. What about the overflow label? Is it the label that is focusable/clickable? Or is it the same as the screenshot above where the content in the label is clickable?

I think it's always the content, right? The label is only interactive because you put a link inside of it. Does that make sense?

mcoker:
> I think it's always the content, right? The label is only interactive because you put a link inside of it. Does that make sense?

@mcarrano yeah... though maybe another way of asking is, is the "overflow label" a label with an action inside that closes the label group, or it an action/button that looks like a label?

src/patternfly/components/LabelGroup/examples/LabelGroup.md
christiemolloy: Yeah i think it would benefit the user because IMO its easier to comprehend when the screen reader is reading it

src/patternfly/components/LabelGroup/label-group.scss
christiemolloy: Okay yeah if it means updating the chip group too, that would be a breaking change, so keeping it as the group's label is ok with me

mcarrano:
Ah, I think that I misunderstood your earlier question @mcoker . Yes, to me the overflow label is not really a label but more like a button styled to look like a label so I would expect the focus to be on the entire element. I checked the chip group overflow and looks like that is how it is implemented.

src/patternfly/components/LabelGroup/examples/LabelGroup.md
mcoker: Nope, there is a negative right margin on the container. That allows labels in the middle of the list to touch the right edge of the container when the labels wrap.

src/patternfly/components/LabelGroup/label-group.scss
mcoker: @christiemolloy we added support for that after the fact, via a `pf-c-label__text` element, so that can be added to the chips. There are examples of it on https://www.patternfly.org/v4/documentation/core/components/label (the last label in each row). There is an open react issue (https://github.com/patternfly/patternfly-react/issues/4660) to add support for that via a prop, and in a breaking change release, we'll make that the default, which will match the chip. I added an issue for that update here - https://github.com/patternfly/patternfly/issues/3409

mcoker:
> @maryshak1996 what do you think about moving the close button for the vertical view to the top right corner ? just seems a little odd being at the very bottom

@christiemolloy we talked about this and I mocked that variation up, and @maryshak1996 said to keep the button at the bottom. Though I'm happy to revisit if what we have isn't a good solution.

mcoker:
@christiemolloy updated the overflow label to be a button.

mcoker:
@mattnolting good call! Looks like that would need some more exploration as to how to best handle. I would be in favor of adding this component as is, then following up on that in the next or another milestone since the component will be beta for a bit. WDYT?

mattnolting:
> @mattnolting good call! Looks like that would need some more exploration as to how to best handle. I would be in favor of adding this component as is, then following up on that in the next or another milestone since the component will be beta for a bit. WDYT?

WFM

mcoker:
> i think the last child shouldnt have a margin-bottom (esp for vertical), because if someone is using the label group and they put it next to text then they may not want the extra space

@christiemolloy there is a negative margin on the outer `__list` element, so if you look at the dimensions of the outer `pf-c-label-group` component, you'll see there is no extra space at the bottom. That said, since we don't need to support wrapping with the vertical variation, I pulled that negative margin out and disabled the margin on the last child like you suggested.

> in two of the vertical variations, it looks like the overflow button is a <span> instead of a button <button>

Nice! Updated.

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -0,0 +1,105 @@
...
+  --pf-c-label-group--m-category--PaddingBottom: var(--pf-global--spacer--xs);
+  --pf-c-label-group--m-category--PaddingLeft: var(--pf-global--spacer--sm);
+  --pf-c-label-group--m-category--BorderRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-label-group--m-category--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-label-group--m-category--BorderColor: var(--pf-global--BorderColor--300);
+  --pf-c-label-group--m-category--BackgroundColor: var(--pf-global--BackgroundColor--100);
+
+  // Label in group
+  --pf-c-label-group__label--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-label-group__label--MarginBottom: 0;
```
christiemolloy, line 18: is this neceassary ? i dont know if we'll ever want a marginbottom, but maybe we would ?

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -0,0 +1,105 @@
...
+  --pf-c-label-group--m-vertical__label--MarginBottom: var(--pf-global--spacer--sm);
+  --pf-c-label-group__label--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-label-group__label--MaxWidth: 18ch;
```
christiemolloy, line 21: in a future issue maybe we could make this a global var since we seem to be using the max-width for labels/text in groups a lot

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.pf-c-label-group,
+.pf-c-label-group__list {
+  display: inline-flex;
+  flex-wrap: wrap;
+  align-items: center;
+}
+
+.pf-c-label-group__list-item {
```
christiemolloy, line 85: should you also remove on the last item here too ?

src/patternfly/components/LabelGroup/label-group.scss
mcoker: Good idea. @mceledonia @mcarrano wdyt about that? Seems like a nice theming option.

src/patternfly/components/LabelGroup/label-group.scss
mcoker: Nope, I don't think so... since horizontal labels will wrap in a label group, we want any label that is at the right edge of the container to be flush with the container - not just the last one. So they all have the right margin, and the parent has a negative right margin to offset it.

src/patternfly/components/LabelGroup/label-group.scss
mcoker: I could go either way with it. Having it allows us to easily change it under `.pf-m-vertical` as opposed to `.pf-m-vertical pf-c-label-group__label { margin-bottom: ... }`

I'm surprised we don't have one currently for when the chip/label groups wrap under the label.

christiemolloy: ok lets keep it!

PR #3397, opened 2020-08-13
feat(menu): add menu component

### LINKED ISSUE #2950: Context menu component
Introduce a right-click context menu component that can be applied upon right click of any element. Can support multiple levels as noted in the design.

Designs: https://marvelapp.com/dhj94fe/screen/67940185  (see last page for spec).

cc @mcoker @LHinson

christiemolloy:
closes #2950


christiemolloy:
@mcarrano 

1) I agree about calling it just `menu` I wasnt sure if maybe that was going to collide with other types of menus that could exist on a page like in a navigation or something, but if this is the main menu going forward that would be less confusing than `context-menu`

2) Yes that toggle was for the drill down or fly-out. Agree that toggle could be misleading since toggle should be reserved for the actual toggle that goes with the menu. The arrow indicates that there are sub-menu items, so maybe we could call the arrow something else to do with expansion

3) Yes. I made a general container for just actions, you can add the modifier `pf-m-favorite` to get the favorite styles for the different colors. I guess my question would be, when you want to display favoriting, should the entire menu have the favorites on the right hand side, or could just a few of them? Im leaning towards leaving the favorite example and then adding a new example to highlight how you can have any action in the right hand side

4) Yes I forgot about the option multi-select, thanks for the reminder! Will add that in.

mcoker:
I think context menu is a little confusing. Menu works, there was a [`<menu>` html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu) that could also be defined as a [contextmenu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#Context_menu) that fits our usage.

There is a similar element role, [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role), but I believe that's specific for a list of options that are selectable, not just actionable.

doruskova:
@christiemolloy @mcarrano This looks great! I agree with the naming and also with the "With favorites" section to be renamed to something more general including an "action" word. Also, I think that the entire menu should have a "favorite" icon on the right side. My only question is about the alignment of these icons . 

- Default state: 

Snímek obrazovky 2020-08-17 v 12 50 50

- Hover state:

Snímek obrazovky 2020-08-17 v 12 37 36

The "favorite" icon is aligned to a center, but the other icons are aligned to a top along with the item name. At first sight, it looks that the icon is flying, because, for example, the "with toggle" section aligns the arrow icon to the top. I'd align the star icon to a top when the description is included. WDYT? 
Snímek obrazovky 2020-08-17 v 12 53 28

christiemolloy:
Thanks for the feedback @doruskova. About the alignment, that would be the best/easiest solution to move the action item to the top when there is a description. The other way im thinking would be to move all of those actions to the contianer on the right but we couldnt do that because those are tied to the link element which is the main left container. So ill try moving the action up and let me know what you think :)

mcoker:
@christiemolloy agree, the actions currently occupy the entire height of the item, with the icon top aligned.

Screen Shot 2020-08-17 at 9 27 20 AM
Screen Shot 2020-08-17 at 9 27 15 AM

christiemolloy:
update @mcarrano @doruskova @mcoker

doruskova:
@christiemolloy looks good to me :) thank you!

doruskova:
@mcarrano Sure thing! @christiemolloy [here's the mock](https://marvelapp.com/prototype/66b0059/screen/72079939) with multilevel-drill down menu which should be for both desktop and touch devices. 
I think that the skeleton of this component is the same as here. 
The mock I shared here has more parts: 
- default menu with one step back > the similar one with yours for responsive version
- menu with collapsible/hidden breadcrumbs (dropdown with breadcrumbs) 
- menu with action item at the bottom 
- empty states included

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,812 @@
...
+            From Git
+          {{/menu-item-text}}
+        {{/menu-item-main}}
+      {{/menu-item}}
+    {{/menu-list-item}}
+    {{#> menu-list-item}}
+      {{#> menu-item}}
+        {{#> menu-item-main}}
+          {{#> menu-item-icon}}
+            <i class="fas fa-layer-group" aria-hidden="true"></i>
```
mcoker, line 75: can we add `fa-fw` to these icons?

mcoker:
I think we'll want to align these?

Screen Shot 2020-08-18 at 10 17 14 AM

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,252 @@
...
+  --pf-c-menu__item-action--m-favorite--m-favorited__icon--Color: var(--pf-global--palette--gold-400);
+
+  padding-top: var(--pf-c-menu--PaddingTop);
+  padding-bottom: var(--pf-c-menu--PaddingBottom);
+  background-color: var(--pf-c-menu--BackgroundColor);
+  box-shadow: var(--pf-c-menu--BoxShadow);
+
+  &.pf-m-submenu {
+    position: absolute;
+    top: 0;
```
mcoker, line 84: should we use `calc(var(--pf-c-menu--PaddingTop) * -1)` here so they align?

Screen Shot 2020-08-18 at 10 17 14 AM

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,812 @@
...
+    {{#> menu-list-item}}
+      {{#> menu-item}}
+        {{#> menu-item-main}}
+          {{#> menu-item-text}}
+            Action 2
+          {{/menu-item-text}}
+        {{/menu-item-main}}
+      {{/menu-item}}
+    {{/menu-list-item}}
+    {{> divider}}
```
mcoker, line 336: needs to be an `li`

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,252 @@
...
+  --pf-c-menu__group-title--FontWeight: var(--pf-global--FontWeight--semi-bold);
+  --pf-c-menu__group-title--Color: var(--pf-global--Color--dark-200);
+
+  // description
+  --pf-c-menu__item-description--FontSize: var(--pf-global--FontSize--xs);
+  --pf-c-menu__item-description--Color: var(--pf-global--Color--200);
+
+  // icon
+  --pf-c-menu__item-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-menu__item-icon--Color: var(--pf-global--Color--dark-200);
```
mcoker, line 49: Are we sure the icon shouldn't just match the current text color? It does in the current dropdown implementation - https://www.patternfly.org/v4/documentation/core/components/dropdown#split-button-action

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,812 @@
...
+      {{#> menu-item menu-item--type="a" menu-item--attribute='href="#"'}}
+        {{#> menu-item-main}}
+          {{#> menu-item-text}}
+            Link not in group
+          {{/menu-item-text}}
+        {{/menu-item-main}}
+      {{/menu-item}}
+    {{/menu-list-item}}
+    {{> divider}}
+    {{#> menu-group}}
```
mcoker, line 364: this is a `section` and the parent is a `ul`

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,812 @@
...
+          {{/menu-item-action}}
+        {{/menu-list-item}}
+        {{#> menu-list-item}}
+          {{#> menu-item}}
+            {{#> menu-item-main}}
+              {{#> menu-item-text}}
+                Item 4
+              {{/menu-item-text}}
+              {{> menu-item-external-icon}}
+              {{> menu-item-select-icon}}
```
mcoker, line 533: This needs to behave similarly to the external link icon so that its appearance doesn't shift the menu around when it shows/hides. I think we would hide it visually like the external link icon, then show it when the item is `pf-m-selected`

Aug-18-2020 10-41-07

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,812 @@
...
+          {{> menu-item-action menu-item-action--IsFavorite="true"}}
+        {{/menu-list-item}}
+        {{#> menu-list-item}}
+          {{#> menu-item}}
+            {{#> menu-item-main}}
+              {{#> menu-item-text}}
+                Item 2
+              {{/menu-item-text}}
+              {{> menu-item-external-icon}}
+              {{> menu-item-select-icon}}
```
mcoker, line 612: is this a valid combo? can selectable items serve as something that opens in a new window?

Screen Shot 2020-08-18 at 12 29 56 PM

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,252 @@
...
+
+// List item
+.pf-c-menu__list-item {
+  position: relative;
+  display: flex;
+  color: var(--pf-c-menu__list-item--Color);
+  background-color: var(--pf-c-menu__list-item--BackgroundColor);
+
+  &:hover:not(.pf-m-disabled),
+  &:focus:not(.pf-m-disabled) {
```
mcoker, line 117: I didn't see any `tabindex`es on this element so you can't focus on it. Focus styles aren't changing the background color. This may be a use cases for `:focus-within`

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,252 @@
...
+    .pf-c-menu__item-action-icon {
+      font-size: var(--pf-c-menu__item-action--m-favorite__icon--FontSize);
+    }
+
+    --pf-c-menu__item-action-icon--Color: var(--pf-c-menu__item-action--m-favorite__icon--Color);
+
+    &.pf-m-favorited {
+      --pf-c-menu__item-action-icon--Color: var(--pf-c-menu__item-action--m-favorite--m-favorited__icon--Color);
+    }
+  }
```
mcoker, line 244: Can you reorder the things in this block so it's

```
// vars
// & selectors
// nested selectors
```

src/patternfly/components/Menu/menu.scss
christiemolloy: Yep thats a good idea

src/patternfly/components/Menu/menu.scss
christiemolloy: @mcarrano @mceledonia @doruskova what do you guys think ?

christiemolloy: This is the current dropdown:

Screen Shot 2020-08-18 at 1 59 26 PM

christiemolloy: **This is what i have in the menu:**
**default:**
Screen Shot 2020-08-18 at 2 09 34 PM
**hover/focus:**
Screen Shot 2020-08-18 at 2 09 38 PM

gdoyle1: @christiemolloy i think it makes sense to have it gray since it's a separate action and distinct from the menu item!

src/patternfly/components/Menu/examples/Menu.md
christiemolloy: based on conversation with matt, i think im going to update the body to use `<div>` and then you could nest multiple menu's inside of the body. that would fix this problem too.

src/patternfly/components/Menu/menu.scss
mcoker: Just to make sure we're referring to the same thing, this question is just about the icon on the left of the text. Here's an example of the icon when used with an item that also has an associated icon

what it looks like now
Screen Shot 2020-08-18 at 1 37 33 PM

what it would look like if the icon color matched the text color (like it does currently in the dropdown)
Screen Shot 2020-08-18 at 1 37 44 PM

christiemolloy: oh my bad i was comparing the wrong icons 😩 , @gdoyle1 what do you think about cokers example , i think it makes sense for them to be the same as text since they dont have any hover/focus states

src/patternfly/components/Menu/examples/Menu.md
christiemolloy: No i dont think so, i was just testing out how the spacing would look, but will remove the select icon!

src/patternfly/components/Menu/menu.scss
christiemolloy: nice! that works

src/patternfly/components/Menu/menu.scss
doruskova: @christiemolloy I agree with Michael, black makes more sense when the icon doesn't have any states.

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,837 @@
...
+            {{/menu-item-text}}
+          {{/menu-item-main}}
+        {{/menu-item}}
+      {{/menu-list-item}}
+    {{/menu-list}}
+  {{/menu-body}}
+{{/menu}}
+```
+
+### With-links
```
mcoker, line 287: You don't need to add hyphens to the titles.

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,837 @@
...
+{{#> menu}}
+  {{#> menu-body}}
+    {{#> menu-list}}
+      {{#> menu-list-item}}
+        {{#> menu-item menu-item--type="a" menu-item--attribute='href="#"'}}
+          {{#> menu-item-main}}
+            {{#> menu-item-text}}
+              Link 1
+            {{/menu-item-text}}
+            {{> menu-item-external-icon}}
```
mcoker, line 298: I wonder if we should remove the external link icon for at least one of these to show that it isn't necessary unless the icon indicates the link launches in a new window or is an external domain.

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,837 @@
...
+```hbs
+{{#> menu}}
+  {{#> menu-body}}
+    {{#> menu-group}}
+      {{#> menu-group-title}}
+        Favorites
+      {{/menu-group-title}}
+      {{#> menu-list}}
+        {{#> menu-list-item}}
+          {{#> menu-item menu-item--type="a"}}
```
mcoker, line 575: these are missing an `href`. You could use `--IsLink` or something to generate the element and attributes. You also need to add `type="button"` to the actions which could be done as an `--IsAction` or `{{# unless --IsLink}}`

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,837 @@
...
+            Description
+          {{/menu-item-description}}
+        {{/menu-item}}
+      {{/menu-list-item}}
+    {{/menu-list}}
+  {{/menu-body}}
+{{/menu}}
+```
+
+### Responsive-drill-down
```
mcoker, line 742: Do you want to leave this example or remove it until we support that?

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+  --pf-c-menu__item-action--m-favorite__icon--Color: var(--pf-global--disabled-color--200);
+  --pf-c-menu__item-action--m-favorite__icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-menu__item-action--m-favorite--m-favorited__icon--Color: var(--pf-global--palette--gold-400);
+
+  padding-top: var(--pf-c-menu--PaddingTop);
+  padding-bottom: var(--pf-c-menu--PaddingBottom);
+  background-color: var(--pf-c-menu--BackgroundColor);
+  box-shadow: var(--pf-c-menu--BoxShadow);
+
+  &.pf-m-flyout {
```
mcoker, line 88: I don't know if it's expensive to be adding and removing things from the DOM to hide/show the submenus but another approach to how this is built could be to render all of the markup and we hide with `display: none; visibility: hidden;` and react just toggles a class (`.pf-m-expanded`) that we then hide/show the menu with via CSS

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+    .pf-c-menu__sublist {
+      position: absolute;
+      top: var(--pf-c-menu--m-flyout__sublist--Top);
+      left: 100%;
+    }
+
+    .pf-c-menu__list-item {
+      &:hover,
+      &:focus-within {
+        color: unset;
```
mcoker, line 98: what does this do?

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+        background-color: unset;
+      }
+    }
+
+    .pf-c-menu__item {
+      &:hover,
+      &:focus-within {
+        color: var(--pf-c-menu__list-item--hover--Color);
+        background-color: var(--pf-c-menu__list-item--hover--BackgroundColor);
+      }
```
mcoker, line 108: why do you need to move the `:hover/focus-within` styles in a flyout menu? Assuming it's necessary, you probably need to scope the classes, or maybe add a class to a list-item that has a submenu and scope it to that. Here's what it looks like if you nest a list with actions in the flyout menu now

Screen Shot 2020-08-19 at 6 22 17 PM

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+  padding-top: var(--pf-c-menu--PaddingTop);
+  padding-bottom: var(--pf-c-menu--PaddingBottom);
+  background-color: var(--pf-c-menu--BackgroundColor);
+  box-shadow: var(--pf-c-menu--BoxShadow);
+
+  &.pf-m-flyout {
+    .pf-c-menu__sublist {
+      position: absolute;
+      top: var(--pf-c-menu--m-flyout__sublist--Top);
+      left: 100%;
```
mcoker, line 92: we probably also want the ability for the menu to flyout to the left. That can come later though if needed.

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// Header
+.pf-c-menu__header {
+  padding-top: var(--pf-c-menu__header--PaddingTop);
+  padding-right: var(--pf-c-menu__header--PaddingRight);
+  padding-bottom: var(--pf-c-menu__header--PaddingBottom);
+  padding-left: var(--pf-c-menu__header--PaddingLeft);
+
+  .pf-c-menu__item {
```
mcoker, line 125: do you have an example with an item in a header? Also assuming that's needed, could you just move the `__item--Padding` vars up under `__header`?

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+  &:hover:not(.pf-m-disabled),
+  &:focus-within:not(.pf-m-disabled) {
+    --pf-c-menu__list-item--Color: var(--pf-c-menu__list-item--hover--Color);
+    --pf-c-menu__list-item--BackgroundColor: var(--pf-c-menu__list-item--hover--BackgroundColor);
+
+    > .pf-c-menu__sublist > .pf-c-menu__list-item {
+      --pf-c-menu__list-item--Color: unset;
+      --pf-c-menu__list-item--BackgroundColor: none;
+    }
+
```
mcoker, line 156: Not sure what this does?

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,294 @@
...
+    .pf-c-menu__item-external-icon {
```
mcoker, line 157: you'll want to scope this class in case there are nested links in submenus

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -0,0 +1,837 @@
...
+        {{#> menu-item}}
+          {{#> menu-item-main}}
+            {{#> menu-item-text}}
+              Add storage
+            {{/menu-item-text}}
+          {{/menu-item-main}}
+        {{/menu-item}}
+      {{/menu-list-item}}
+      {{#> menu-list-item}}
+        {{#> menu-item}}
```
mcoker, line 182: if an item has a sublist, should the button that triggers it have `aria-expanded`?

src/patternfly/components/Menu/examples/Menu.md
christiemolloy: Yeah good point, i was following the designs in the marvel doc here: https://marvelapp.com/prototype/dhj94fe/screen/66701095 but ill wait til matt implements it completely

src/patternfly/components/Menu/examples/Menu.md
christiemolloy: updated the "links" and "favorites" examples to show only some of the list items displaying the external icon and added `target="_blank"` to menu items with the external icon, does that sound right ?

src/patternfly/components/Menu/menu.scss
christiemolloy: Yes the example was in the drill-down example where I nested the item inside of the header, I removed the padding from the item because the header has the padding, so i cant imagine that the item would need to have padding, what do you think? I guess we could also have a new container for the drill down in the header ? @mattnolting what do you think 

Screen Shot 2020-08-20 at 11 14 50 AM

src/patternfly/components/Menu/examples/Menu.md
mcoker: @mcarrano can you confirm what the intent of the "external" icon is?

Screen Shot 2020-08-20 at 10 17 14 AM

src/patternfly/components/Menu/menu.scss
christiemolloy: do you think it should go inside of `pf-c-menu__item` or `pf-c-menu__item-main` ?

src/patternfly/components/Menu/menu.scss
christiemolloy: i might be wrong, but i think from @mattnolting and I's conversation, we wanted to reuse `pf-c-menu__item-sublist` for every sublist, and apply `pf-m-flyout` to apply the styling for sublist if it should be positioned like a flyout,
are you suggesting to add `pf-m-flyout` to each `pf-c-menu__item-sublist` instead ? and have a `pf-m-expanded` class if its expanded otherwise its hidden ?

src/patternfly/components/Menu/menu.scss
```diff
@@ -0,0 +1,253 @@
...
+      pointer-events: none;
+    }
+  }
+}
+
+// Item
+.pf-c-menu__item {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
```
mattnolting, line 134:
```suggestion
  flex-basis: 100%;
  min-width: 0;
```

`width: 100%` here causes this:

Screen Shot 2020-08-20 at 2 00 03 PM

PR #3401, opened 2020-08-13
feat(datalist): add draggable variation

### LINKED ISSUE #2728: Introduce draggable rows to Data List
Want to allow the user to manually sort rows in a data list via drag and drop. This will be required to support column reordering and other potential use cases in the future. 

See designs for table column management here: https://docs.google.com/document/d/1QRCmAHcSrJ4uPlQsUiRLcezz9wo2zdIdAML1wBLTXYI/edit?usp=sharing

christiemolloy:
closes #2728


src/patternfly/components/Chip/chip.scss
```diff
@@ -12,6 +12,11 @@
   // overflow chip button
   --pf-c-chip--m-overflow__text--Color: var(--pf-global--primary-color--100);
 
+  // draggable chip
+  --pf-c-chip--m-draggable--BackgroundColor: var(--pf-global--BackgroundColor--200);
+  --pf-c-chip--m-draggable--BoxShadow: var(--pf-global--BoxShadow--sm);
+  --pf-c-chip__m-draggable--icon--FontSize: var(--pf-global--icon--FontSize--sm);
```
mcoker, line 18:
```suggestion
  --pf-c-chip--m-draggable__icon--FontSize: var(--pf-global--icon--FontSize--sm);
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -60,6 +68,16 @@
     }
   }
 
+  &.pf-m-draggable {
+    --pf-c-chip--BackgroundColor: var(--pf-c-chip--m-draggable--BackgroundColor);
+
+    box-shadow: var(--pf-c-chip--m-draggable--BoxShadow);
+
+    .pf-c-chip__icon {
+      font-size: var(--pf-c-chip__m-draggable--icon--FontSize);
```
mcoker, line 77:
```suggestion
      font-size: var(--pf-c-chip--m-draggable__icon--FontSize);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -79,6 +81,11 @@
   --pf-c-data-list__toggle-icon--Transition: .2s ease-in 0s;
   --pf-c-data-list__item--m-expanded__toggle-icon--Rotate: 90deg;
 
+  // Draggable icon
+  --pf-c-data-list__item--m-draggable-draggable-icon--Color: var(--pf-global--icon--Color--light);
```
mcoker, line 85:
```suggestion
  --pf-c-data-list__item--m-draggable__draggable-icon--Color: var(--pf-global--icon--Color--light);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
     box-shadow: var(--pf-c-data-list__item--m-selected--BoxShadow);
   }
 
+  // stylelint-disable
+  &.pf-m-draggable {
+    position: relative;
+    cursor: pointer;
+
+    .pf-c-data-list__item-draggable-icon {
+      color: var(--pf-c-data-list__item--m-draggable-draggable-icon--Color);
```
mcoker, line 254:
```suggestion
      color: var(--pf-c-data-list__item--m-draggable__draggable-icon--Color);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+      &:not(:last-child) {
+        --pf-c-data-list__item--BorderBottomWidth: 0;
+
+        + .pf-c-data-list__item {
+          border-top: var(--pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth) solid var(--pf-c-data-list__item--m-selectable--hover--item--BorderTopColor);
+        }
+      }
+
+      .pf-c-data-list__item-draggable-icon {
+        --pf-c-data-list__item--m-draggable-draggable-icon--Color: var(--pf-c-data-list__item--m-draggable--hover-draggable-icon--Color);
```
mcoker, line 270:
```suggestion
        --pf-c-data-list__item--m-draggable__draggable-icon--Color: var(--pf-c-data-list__item--m-draggable--hover__draggable-icon--Color);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+      }
+    }
+
+    &.pf-m-disabled {
+      pointer-events: none;
+
+      .pf-c-data-list__item-draggable-icon {
+        --pf-c-data-list__item--m-draggable-draggable-icon--Color: var(--pf-c-data-list__item--m-draggable--m-disabled-draggable-icon--Color);
```
mcoker, line 278:
```suggestion
        --pf-c-data-list__item--m-draggable__draggable-icon--Color: var(--pf-c-data-list__item--m-draggable--m-disabled__draggable-icon--Color);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+    &.pf-m-dragged {
+      .pf-c-data-list__item-row {
+        &::after {
+          position: absolute;
+          top: 0;
+          right: 0;
+          bottom: 0;
+          left: 0;
+          content: "";
+          background-color: var(--pf-global--disabled-color--200);
```
mcoker, line 291: this should be a component var

maryshak1996:
> I think this looks good although it's hard to evaluate without the JS implementation. My only question for @maryshak1996 and @mceledonia is whether the "draggable chip" should have a drop shadow. Seems like it should since it's something that overlays other content.

I definitely agree with you Matt -- the chip should always have the appearance of floating above the data list (so the hover would achieve that)

mcoker:
I wonder if we're going to end up with conflicting styles, like `pf-m-selectable` and `pf-m-draggable`. When an item is selectable, it shows a similar (but not the same) shadow. Selectable uses 2 shadows - `var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom)` - and draggable uses `var(--pf-global--BoxShadow--sm)`. The "selected" state also uses the same shadow. When you hover/focus on a selectable row, its shadow should appear over adjacent "selected" rows. Currently the draggable shadow on hover does not appear on top of selected rows. Should it? And should those shadows be the same? Assuming so, should we have a single modifier that creates the shadow on hover/focus that we use for both draggable and selectable?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+
+    &.pf-m-disabled {
+      pointer-events: none;
+
+      .pf-c-data-list__item-draggable-icon {
+        --pf-c-data-list__item--m-draggable-draggable-icon--Color: var(--pf-c-data-list__item--m-draggable--m-disabled-draggable-icon--Color);
+      }
+    }
+
+    &.pf-m-dragged {
```
mcoker, line 282: wdyt about `.pf-m-dragging` instead since it's more of a present tense action than past tense?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+          bottom: 0;
+          left: 0;
+          content: "";
+          background-color: var(--pf-global--disabled-color--200);
+          mix-blend-mode: screen;
+          opacity: .8;
+        }
+      }
+
+      .pf-c-chip.pf-m-draggable {
```
mcoker, line 297: just a reminder to pull this out

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
     box-shadow: var(--pf-c-data-list__item--m-selected--BoxShadow);
   }
 
+  // stylelint-disable
+  &.pf-m-draggable {
+    position: relative;
+    cursor: pointer;
```
mcoker, line 251: you can use a `grab` cursor here.

src/patternfly/components/Chip/chip.scss
```diff
@@ -88,3 +106,7 @@
   font-size: var(--pf-c-chip__text--FontSize);
   color: var(--pf-c-chip__text--Color);
 }
+
+.pf-c-chip__icon {
+  margin-right: var(--pf-c-chip__icon--MarginRight);
```
mcoker, line 111: I wonder if this should be `__icon + __text, __text + icon { margin-left... }`? That would allow the icon to be placed before or after the text, and also have a standalone `__icon`

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+
```
mcoker, line 252: If a user should be able to click anywhere in the item row to drag the row (and clicking on any links/buttons/checks/etc in the row would do nothing), then I wonder if we should draw an `::after` pseudo element here with a `z-index` so that it overlays the row and makes clicking anywhere on the row trigger a click on the pseudo element.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+          border-top: var(--pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth) solid var(--pf-c-data-list__item--m-selectable--hover--item--BorderTopColor);
+        }
+      }
+
+      .pf-c-data-list__item-draggable-icon {
+        --pf-c-data-list__item--m-draggable-draggable-icon--Color: var(--pf-c-data-list__item--m-draggable--hover-draggable-icon--Color);
+      }
+    }
+
+    &.pf-m-disabled {
```
mcoker, line 274: Is this used to indicate that a row isn't draggable?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +245,65 @@
...
+      .pf-c-data-list__item-row {
+        &::after {
+          position: absolute;
+          top: 0;
+          right: 0;
+          bottom: 0;
+          left: 0;
+          content: "";
+          background-color: var(--pf-global--disabled-color--200);
+          mix-blend-mode: screen;
```
mcoker, line 292: very cool! 👍

mcarrano:
> I wonder if we're going to end up with conflicting styles, like `pf-m-selectable` and `pf-m-draggable`. When an item is selectable/selected, it shows the same shadow. When you hover/focus on a selectable row, its shadow should appear over adjacent "selected" rows. Currently the draggable shadow on hover does not appear on top of selected rows. Should it? @maryshak1996
> 
> -- actually I was assuming this feature could apply to any data-list. Is that the end goal, or are we only supporting dragging rows in the [column management modal](https://patternfly-pr-3401.surge.sh/documentation/core/demos/table/column-management-modal)?

These are good questions, @mcoker . To answer your last question first, I do expect this could be applied to any data list. It was being driven by the need to reorder columns, but we definitely should not restrict to that.

It looks like the way this is implemented, I can mouse-down and drag anywhere in the row, is that right? I'm wondering if we should constrain this a bit and require the user to mouse over the drag icon to get this behavior. I too am concerned about conflicts. For example, If I have interactive elements inside of a row, when is the mouse event interpreted as the start of a drag vs a click event? @maryshak1996 @mceledonia what were your expectations here?

mcoker:
> I do expect this could be applied to any data list

@christiemolloy has done a great job with this so I think we could support that with a few more changes.

> I can mouse-down and drag anywhere in the row, is that right? I'm wondering if we should constrain this a bit and require the user to mouse over the drag icon to get this behavior. I too am concerned about conflicts. For example, If I have interactive elements inside of a row, when is the mouse event interpreted as the start of a drag vs a click event?

That is how it's currently implemented. If a draggable data-list can be re-ordered at any time (meaning the draggable icon is always there, and you can interact with the data-list items and drag rows) I think it makes more sense that clicking on the draggable icon is what triggers a draggable row.

If draggable rows is something that is enabled only when you want to drag rows (via an action or something), and it's expected that when that's enabled, all you can do with the data list is drag the rows, and you disable the ability to drag the rows when you're done, then we could make it so that clicking anywhere in the row will drag the row.

src/patternfly/components/DataList/data-list.scss
christiemolloy: the disabled modifier is a separate style, @maryshak1996 can you speak more to this ?

src/patternfly/components/DataList/data-list.scss
christiemolloy: yep thats a good idea

src/patternfly/components/Chip/chip.scss
christiemolloy: nicee!

christiemolloy:
"If draggable rows is something that is enabled only when you want to drag rows (via an action or something), and it's expected that when that's enabled, all you can do with the data list is drag the rows, and you disable the ability to drag the rows when you're done, then we could make it so that clicking anywhere in the row will drag the row."

^ @mcoker I think this is a good point. I would imagine that when you see the draggable icon in the data list you would be inside of a "draggable" mode where you cant actually interact with the data list, but instead you are just re-ordering the rows. @mcarrano is that correct?

src/patternfly/components/Chip/chip.scss
```diff
@@ -36,7 +36,7 @@
   --pf-c-chip__c-badge--MarginLeft: var(--pf-global--spacer--xs);
 
   // icon
-  --pf-c-chip__icon--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-chip__icon--MarginLefts: var(--pf-global--spacer--xs);
```
mattnolting, line 39:
```suggestion
  --pf-c-chip__icon--MarginLeft: var(--pf-global--spacer--xs);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +247,68 @@
     box-shadow: var(--pf-c-data-list__item--m-selected--BoxShadow);
   }
 
+  // stylelint-disable
+  &.pf-m-draggable {
+    position: relative;
+    cursor: grab;
+
```
mattnolting, line 254:
```suggestion

    &:active {
      cursor: grabbing;
    }
```

christiemolloy: i added this below, should i move it above hover ?

mattnolting: Yeah, that's probably better

src/patternfly/components/Chip/chip.scss
```diff
@@ -60,6 +68,16 @@
     }
   }
 
+  &.pf-m-draggable {
+    --pf-c-chip--BackgroundColor: var(--pf-c-chip--m-draggable--BackgroundColor);
+
```
mattnolting, line 73:
```suggestion
cursor: grab;
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -60,6 +68,16 @@
...
+    box-shadow: var(--pf-c-chip--m-draggable--BoxShadow);
+
```
mattnolting, line 75:
```suggestion

    &:active {
      cursor: grabbing;
    }
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -12,6 +12,11 @@
   // overflow chip button
   --pf-c-chip--m-overflow__text--Color: var(--pf-global--primary-color--100);
 
+  // draggable chip
+  --pf-c-chip--m-draggable--BackgroundColor: var(--pf-global--BackgroundColor--200);
+  --pf-c-chip--m-draggable--BoxShadow: var(--pf-global--BoxShadow--sm);
+  --pf-c-chip--m-draggable--icon--FontSize: var(--pf-global--icon--FontSize--sm);
```
mcoker, line 18:
```suggestion
  --pf-c-chip--m-draggable__icon--FontSize: var(--pf-global--icon--FontSize--sm);
```

src/patternfly/components/Chip/chip.scss
```diff
@@ -60,6 +68,16 @@
     }
   }
 
+  &.pf-m-draggable {
+    --pf-c-chip--BackgroundColor: var(--pf-c-chip--m-draggable--BackgroundColor);
+
+    box-shadow: var(--pf-c-chip--m-draggable--BoxShadow);
+
+    .pf-c-chip__icon {
+      font-size: var(--pf-c-chip--m-draggable--icon--FontSize);
```
mcoker, line 77:
```suggestion
      font-size: var(--pf-c-chip--m-draggable__icon--FontSize);
```

christiemolloy:
@maryshak1996 @mcarrano we added a button around the toggle icon, what do you think of the focus ring? I think it might be confusing to the user because it doesnt indicate that the whole row is selected, or do you think its ok? Otherwise we could look at trying to change focus to the parent when you click that child element.

Screen Shot 2020-08-28 at 10 07 07 AM

maryshak1996:
> @maryshak1996 @mcarrano we added a button around the toggle icon, what do you think of the focus ring? I think it might be confusing to the user because it doesnt indicate that the whole row is selected, or do you think its ok? Otherwise we could look at trying to change focus to the parent when you click that child element.
hummm I see what you mean with that, but I feel like it might be more confusing if the focus ring extended to the whole row. curious what @mcarrano thinks

mcarrano:
> @maryshak1996 @mcarrano we added a button around the toggle icon, what do you think of the focus ring? I think it might be confusing to the user because it doesnt indicate that the whole row is selected, or do you think its ok? Otherwise we could look at trying to change focus to the parent when you click that child element.

I see what you mean @christiemolloy . But it does look like individual elements contained in the row take focus as I tab through? In that case, this behavior might be consistent and helpful for keyboard users. What do you think @jessiehuff ?

src/patternfly/components/DataList/data-list-item-draggable-button.hbs
```diff
@@ -0,0 +1,6 @@
+<button class="pf-c-data-list__item-draggable-button{{#if data-list-item-draggable-button--modifier}} {{data-list-item-draggable-button--modifier}}{{/if}}"
+  {{#if data-list-item-draggable-button--attribute}}
+    {{{data-list--item-draggable-button--attribute}}}
```
mcoker, line 3:
```suggestion
    {{{data-list-item-draggable-button--attribute}}}
```

src/patternfly/components/DataList/data-list-item-draggable-icon.hbs
```diff
@@ -0,0 +1,6 @@
+<span class="pf-c-data-list__item-draggable-icon{{#if data-list-item-draggable-icon--modifier}} {{data-list-item-draggable-icon--modifier}}{{/if}}"
+  {{#if data-list-item-draggable-icon--attribute}}
+    {{{data-list--item-draggable-icon--attribute}}}
```
mcoker, line 3:
```suggestion
    {{{data-list-item-draggable-icon--attribute}}}
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +249,57 @@
     box-shadow: var(--pf-c-data-list__item--m-selected--BoxShadow);
   }
 
+  // stylelint-disable
+  &.pf-m-draggable {
+    position: relative;
+
+    .pf-c-data-list__item-draggable-button {
```
mcoker, line 256: does this need to be nested in `.pf-m-draggable`? Doesn't seem like it depends on that parent class for anything.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +249,57 @@
...
+      padding-right: var(--pf-c-data-list__item--m-draggable__draggable-button--PaddingRight);
+      padding-left: var(--pf-c-data-list__item--m-draggable__draggable-button--PaddingLeft);
+      margin-top: var(--pf-c-data-list__item--m-draggable__draggable-button--MarginTop);
+      margin-left: var(--pf-c-data-list__item--m-draggable__draggable-button--MarginLeft);
+      border: 0;
+
+      &:hover {
+        cursor: grab;
+
+        --pf-c-data-list__item--m-draggable__draggable-button__draggable-icon--Color: var(--pf-c-data-list__item--m-draggable__draggable-button--hover__draggable-icon--Color);
```
mcoker, line 266: need this color on `:focus`, too

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +249,57 @@
...
+
+      &.pf-m-disabled {
+        pointer-events: none;
+        cursor: none;
+
+        --pf-c-data-list__item--m-draggable__draggable-button__draggable-icon--Color: var(--pf-c-data-list__item--m-draggable__draggable-button--m-disabled__draggable-icon--Color);
+      }
+
+      .pf-c-data-list__item-draggable-icon {
+        color: var(--pf-c-data-list__item--m-draggable__draggable-button__draggable-icon--Color);
```
mcoker, line 281: you could set this on the `.pf-c-data-list__item-draggable-button` element. I doubt we need this element since it's basically just a plain button, but we could keep it, too, in case we need it.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -238,6 +249,57 @@
...
+      }
+    }
+
+    &.pf-m-ghost-row {
+      .pf-c-data-list__item-row {
+        &::after {
```
mcoker, line 287: This works, assuming we'll close any expanded content sections when you click on the draggable button. Though it could also apply to `.pf-c-data-list__item::after`, and it would overlay anything that's in the `__item`. 

An alternative to this could also be just to change the `opacity` on the item.

Or using a white as a background color and `opacity`, or just `rgba(255,255,255,0.8)` as a background color. 

Is there a benefit to using `mix-blend-mode` instead?

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -666,6 +666,81 @@ When a list item includes more than one block of content, it can be difficult fo
...
+        {{/data-list-cell}}
+      {{/data-list-item-content}}
+    {{/data-list-item-row}}
+  {{/data-list-item}}
+
+  {{#> data-list-item data-list-item--modifier="pf-m-draggable pf-m-ghost-row" data-list-item--attribute=(concat 'aria-labelledby="' data-list--id '-item3"')}}
+    {{#> data-list-item-row}}
+      {{#> data-list-item-control}}
+        {{> data-list-item-draggable-button}}
+        {{#> data-list-check checkbox--attribute=(concat 'name="' data-list--id '-check-action-check3" aria-labelledby="' data-list--id '-item3" checked disabled')}}{{/data-list-check}}
```
mcoker, line 704: This is probably more of a design question, but does the checkbox need to be disabled when you're dragging a row? It seems unnecessary, you can't click it if you're dragging anyways.

src/patternfly/components/DataList/data-list.scss
christiemolloy: yeah i was meaning to ask you about this. i agree it doesnt have to be inside of `pf-m-draggable`

src/patternfly/components/DataList/data-list.scss
christiemolloy: yeah I was thinking of keeping it for now in case anything changes with the interaction of the draggable icon

src/patternfly/components/DataList/examples/DataList.md
christiemolloy: yeah and maybe also a question for @jessiehuff depending on the keyboard interaction?

src/patternfly/components/DataList/data-list.scss
christiemolloy: im not sure of the benefit to using `mix-blend-mode` instead, so I think your idea is a nice simple, straight-forward approach!

src/patternfly/components/DataList/examples/DataList.md
jessiehuff: I think it might make sense for the checkbox to be disabled and therefore out of the tab order when they're in the middle of dragging. As @mcoker was saying, it's likely unnecessary, and potentially could just be an annoying tab stop for keyboard users.

jgiardino:
Thanks for sharing the demos, @jessiehuff! 

The keyboard interaction of the [first one](https://schne324.github.io/dragon-drop/demo/) seems more obvious and natural to me, and also seems to be a better match for the PF design. 

Some questions regarding our design:

- It might be worth considering how the component displays during the keyboard interaction. For example, when the drag button is pressed, how does the sighted keyboard only user know that it's pressed:
  image
- How does the sighted keyboard-only user know what keys can be used to move the list item and save the new position of the list item? Should the drag button provide some affordance to the keyboard user regarding how to interact with the element?
- Does it make sense to include in the design document what gets announced to the user during each drag or key press action? See the next comment regarding contents announced to the user...

Regarding the core implementation, the first [demo/example](https://schne324.github.io/dragon-drop/demo/) that @jessiehuff shared includes additional elements in the DOM:
- description that's associated with the drag button via aria-describedby
- aria-live elements in the dom to support screen reader users

Are these expected to be included in the PF drag-and-drop example? It would be good to capture what gets communicated and when via aria-live contents. Should these contents be included in the design document?

jgiardino:
I never answered this question in my previous reply:
> Some initial thoughts, I don't think we need the role="list" if it's on the <ul> as the implicit role of a <ul> is "list". Though, I see that it's on other examples so perhaps the redundancy doesn't matter too much. Maybe @jgiardino has some insight about why it's typically included here?

During testing on VO, the redundant `role="list"` resulted in VO announcing more information regarding semantics. In a simpler `<ul><li></li></ul>` dom structure, this isn't necessary. But as additional elements are added as children to `<li>` like we have in the Data list component, VO would stop communicating the semantics of the list without the redundant `role` attribute.

Note that this might not be the case anymore, and is worth testing again.

christiemolloy:
Thank you @jessiehuff and @jgiardino for all the very useful information!! Based on your recommendations I went ahead and added:
- Text at the beginning of the data list example that describes how to use the drag drop feature. This has an ID which is used for each of the draggable buttons for `aria-describedby`. Do you think this would be piece of text that we would want to say is required ? We could definitely create an element for that if thats the case.
- `aria-labelledby` to the draggable buttons to label them using the button and text IDs 
- `aria-live="assertive"` on the text element - this is what is done in [Dragon Drop](https://schne324.github.io/dragon-drop/demo/) 
- `aria-pressed` to the draggable button. I didnt add this directly to the code but i added it to the markdown. I wasn't sure if we include it in the Core code or if its just in React ?

What do you think about this ? is there anything i missed ?

jessiehuff (commented):
It's looking great so far!
- I think you might have an id missing? In testing it, I noticed that the checkbox seemed like it wasn't labeled, but when I looked at the html, it looks like it's just that the `aria-labelledby` isn't pointing at anything right now because the id doesn't exist. It also results in axe issues ("ARIA attributes must conform to valid values" and "Form elements must have labels"), but both look like they're issues with missing ids so that the aria-labelledby is pointing to an id that doesn't exist, one with the list item itself and one with the checkbox input.
- In terms of the `aria-live`, this made me also wonder where the right place to put the `aria-live` would be. Jenn and I discussed it, and typically, this is a div that exists in the DOM, where information that should be announced to the user is inserted. In the deque demo, these divs aren’t visible. And they are used to provide information about what is happening. There should be an extra div for our case as well, similar to the `<div id="draggable-help">` div but one that is visually hidden, and maybe just includes placeholder text in the core example stating that this is where information is communicated to the user about what is happening. In the current PR example, this should not be included inside the list items, like it is now.
- Looking at the deque demo, it is a little odd that `aria-pressed` isn't automatically added, I think it might make sense to put these as `aria-pressed="false"` in the example until a user presses the button, making it true. According to the specifications, leaving it undefined means that the button does not support being pressed, which isn’t the case here. 
- Your question on making `aria-describedby` required is a good question that is kind of open for discussion. Jenn and I discussed this too, and it's kind of new territory for us where we're using key commands that might otherwise not be obvious to the user. The same might be true for the drag button as well, although I feel like that’s a bit more standard. It could be argued that we use up/down arrow keys for other components without requiring an explanation on how to use the component, but all of these are in dropdown-like components, which are very much standard and familiar. Whereas the “toggle button to enable reordering” is kind of a new pattern. Perhaps design can weigh in on this? What do you think, @mcarrano? Do you think that we should require the text at the beginning of the data list example that describes how to use the drag drop feature? This has an ID which is used for each of the draggable buttons for aria-describedby. It might be something that's included as part of the component or part of the design documentation or maybe both?

jessiehuff:
@mcarrano and I were just discussing the `aria-describedby` here, and we think it might make sense to keep it optional, but perhaps we could highly recommend it in the accessibility section of the markdown. I'm not sure of the best wording for this, maybe something like "Highly recommended to give users instructions on how to interact with component".

jgiardino:
> @mcarrano and I were just discussing the aria-describedby here, and we think it might make sense to keep it optional, but perhaps we could highly recommend it in the accessibility section of the markdown.

Should we treat the aria-live region in the same way? They have similar purposes, but the aria-describedby text provides more instructional text and the aria-live region provides more real-time feedback.

Here are some examples of what the Dragon Drop demo provides in the aria-live region:
- When using drag and drop
  image
- When using the up/down arrow keys while the drag button is pressed
  image

If we want to include details about aria-live regions, too, here's a draft of a description:
"Recommended to give screen reader users live feedback about what's happening during interaction with the data list, both during drag and drop interactions and keyboard interactions."

Related to @jessiehuff's comment about the missing aria-labelledby ids, there are a few very nit-picky things that seem odd about the button:
- The label for the button is on the icon: `<i class="fas fa-grip-vertical" aria-label="Drag"></i>`. Should this be on the button itself? When this is implemented in react, icons are inline `<svg>`s. Moving the aria-label to the button would create more consistency between the core and react examples.
- The label for the button is "Drag", but should it be something that describes the user action, like "Reorder", and not the method of interaction? Saying "Drag" seems too similar to saying "Click here". Also, a screen reader user might be more likely to use the keyboard than the mouse to interact with the element.

christiemolloy:
Thanks for the feedback @jessiehuff  @jgiardino !!  i updated the branch with your suggestions:

- I updated all of the IDs they're now showing up - so the error should hopefully go away now
- Added the `aria-live` region as a div with the `screen-reader` class so that it is visually hidden / Made it highly recommended in the docs 
- added `aria-pressed="false"` to all the buttons
- kept the div at the top to communicate aria-describedby / made it highly recommended in the docs
- Moved the aria-label to the button instead of the icon and updated the label to say "Reorder"

jessiehuff (approved):
This looks fantastic, Christie!! Great job with this!!! 🥇

src/patternfly/components/Chip/chip.scss
```diff
@@ -30,6 +35,9 @@
   // chip badge
   --pf-c-chip__c-badge--MarginLeft: var(--pf-global--spacer--xs);
 
+  // icon
+  --pf-c-chip__icon--MarginLeft: var(--pf-global--spacer--xs);
```
mattnolting, line 39: This spacing looks tighter than the other examples. Should it be `--pf-c-chip__icon--MarginLeft: var(--pf-global--spacer--sm);`

mattnolting (commented):
I think this looks pretty good. Left a few comments. Also, I think with the addition of `.pf-c-data-list__cell-text`, we would add that to the other cells, but that can be done in a follow up.

mattnolting (approved):
LPTM! Nice work!

mcoker (approved):
Great work!

PR #3403, opened 2020-08-14
fix(toolbar-padding): updated toolbar padding

### LINKED ISSUE #3240: Toolbar padding enhancements

mattnolting:
fixes #3240

Initial PR here: https://github.com/patternfly/patternfly/pull/3317

### DIFF
--- a/src/patternfly/components/Toolbar/examples/Toolbar.md
+++ b/src/patternfly/components/Toolbar/examples/Toolbar.md
@@ -151,12 +151,38 @@ In some instances, it may be necessary to adjust spacing explicitly where items 
 {{/toolbar}}
 ```
 
-### Toolbar spacers
+```hbs title=Insets
+{{#> toolbar toolbar--modifier="pf-m-inset-none pf-m-inset-md-on-md pf-m-inset-2xl-on-lg" toolbar--id="toolbar-inset-example"}}
+  {{#> toolbar-content}}
+    {{#> toolbar-content-section}}
+      {{#> toolbar-group}}
+        {{#> toolbar-item}}
+          Item
+        {{/toolbar-item}}
+        {{#> toolbar-item}}
+          Item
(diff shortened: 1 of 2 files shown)


mcoker:
I was talking with @mcarrano about this and we think it's too risky to change the default padding as we don't want to introduce a visual bug for users using the toolbar currently.

We can offer the inset system as a way to get the padding, but wdyt about adding a variation that adds this padding, then updating the default padding later in a breaking change release?

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -151,12 +151,64 @@ In some instances, it may be necessary to adjust spacing explicitly where items
...
+    {{/toolbar-content-section}}
+  {{/toolbar-content}}
+{{/toolbar}}
+```
+
+### Toolbar spacers and insets
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
+| `.pf-m-page-insets` | `.pf-c-toolbar` | Modifies toolbar insets to match page section, table, page header or any other component whose inset shifts from `--pf-global--spacer--md` to `--pf-global--spacer--lg` at the `xl` breakpoint. |
```
mcoker, line 208: I imagine it's a good idea to reuse this modifier for any other component that needs this kind of variant for spacing. @christiemolloy wdyt about the name?

christiemolloy: yeah i think the naming works!

PR #3404, opened 2020-08-14
Feat add overflow menu to toolbar

### LINKED ISSUE #3213: Add overflow-menu to toolbar
- Create new toolbar demo that includes
  - `overflow-menu`s in the toolbar
Screen Shot 2020-06-23 at 5 16 28 PM
    to this
Screen Shot 2020-06-23 at 5 30 59 PM

  - show `toolbar` with a `table` and sticky pagination, and top pagination hidden 
Screen Shot 2020-06-23 at 5 29 58 PM

... more to come

mattnolting:
closes #3213


src/patternfly/demos/Toolbar/examples/Toolbar.md
```diff
@@ -7,15 +7,15 @@ import './Toolbar.css'
 
 ## Demos
 
-### Toolbar attribute value search filter desktop
+### Toolbar-attribute-value-search-filter-desktop
```
mcoker, line 10: you don't need these dashes

src/patternfly/components/Toolbar/toolbar-overflow-menu-example.hbs
```diff
@@ -1,5 +1,5 @@
-{{#> toolbar-item toolbar-item--modifier="pf-m-overflow-menu"}}
-  {{#> overflow-menu overflow-menu-button--aria-label="Expand overflow menu" overflow-menu--id=(concat toolbar--id '-overflow-menu') dropdown--IsExpanded=toolbar-overflow-menu-example--IsExpanded overflow-menu--modifier=toolbar-demo-overflow-menu--modifier overflow-menu--attribute=toolbar-demo-overflow-menu--attribute}}
+{{#> toolbar-item toolbar-item--modifier=(concat 'pf-m-overflow-menu ' toolbar-overflow-menu-example--modifier)}}
```
mcoker, line 1: for something like this you may call this something specific to the item, so that you can add this to other elements in the same hbs file. So something like `toolbar-overflow-menu-example-toolbar-item--modifier`

PR #3407, opened 2020-08-17
feat(splitter): add the splitter component

### LINKED ISSUE #3203: Introduce horizontal and vertical splitters
Introduce a horizontal and vertical splitter component that can be used with a drawer or similar component for adjusting the width or height of the target element.

Designs can be found in this deck: https://marvelapp.com/8cdg211/screen/69469192

cc @mcoker

christiemolloy:
closes #3203


christiemolloy:
@mcarrano you can have a dark variation of both horizontal and vertical - this is just all i showed! ill update the examples to avoid confusion

doruskova:
@christiemolloy It looks great to me. 
I'm just wondering if the dark version is dark enough. Do we want to reflect a dark mode or just have a darker splitter @mcarrano ?

mcarrano:
Not sure about this @doruskova Can you remind me why we are creating a dark version? What's the use case?If it's for dark theme, I think we should hold off until we address the entire dark theme question.

doruskova:
@mcarrano I didn't create a dark version for the splitter, because you mentioned that the dark theme is in progress and we should wait. As you said, I think we should wait for a dark theme to match this component:)

mcarrano:
In that case maybe we should remove the dark version example. Do you agree @christiemolloy @mcoker ?

christiemolloy:
@mcarrano I do think it would be beneficial to have another color for the splitter, because currently i think the bg-color for the splitter matches our grey default background color, so maybe there should be a bg-color we recommend using when the default bg-color is in use.

doruskova:
@christiemolloy I've updated design for the splitter and I used `#D2D2D2` for bg and `#151515` for lines. This should have better contrast than the previous one https://marvelapp.com/prototype/8cdg211/screen/69932685

christiemolloy:
updated @mcarrano @doruskova

doruskova (commented):
looks good to to me @christiemolloy

src/patternfly/components/Splitter/splitter.scss
```diff
@@ -0,0 +1,70 @@
...
+  --pf-c-splitter--m-vertical__handle--Bottom: 0;
+
+  // splitter handle line
+  --pf-c-splitter__handle-line--BackgroundColor: var(--pf-global--Color--100);
+  --pf-c-splitter__handle-line--Height: #{pf-size-prem(1px)};
+  --pf-c-splitter__handle-line--Width: #{pf-size-prem(12px)};
+  --pf-c-splitter--m-vertical__handle-line--Height: #{pf-size-prem(12px)};
+  --pf-c-splitter--m-vertical__handle-line--Width: #{pf-size-prem(1px)};
+
+  position: relative;
```
mattnolting, line 28: I think, in use, this will need to be absolutely positioned. Then `top`, `bottom`,`left`, and `right` would control where it appears in the component?

mattnolting, line 28: I also wonder if the clickable area should be larger (vertically in the horizontal splitter, horizontal in the vertical splitter). Though that may be something we can define.

src/patternfly/components/Splitter/examples/Splitter.css
```diff
@@ -0,0 +1,3 @@
+#ws-core-c-splitter-vertical .ws-preview-html {
+  height: 300px;
+}
```
mattnolting, line 3:
```suggestion
}

```

src/patternfly/components/Splitter/splitter.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="pf-c-splitter{{#if splitter--modifier}} {{splitter--modifier}}{{/if}} {{#if splitter--IsVertical}} pf-m-vertical{{/if}}"
```
mcoker, line 1:
```suggestion
<div class="pf-c-splitter{{#if splitter--modifier}} {{splitter--modifier}}{{/if}}{{#if splitter--IsVertical}} pf-m-vertical{{/if}}"
```

src/patternfly/components/Splitter/splitter.scss
```diff
@@ -0,0 +1,70 @@
+.pf-c-splitter {
+  // splitter
+  --pf-c-splitter--Height: var(--pf-global--spacer--sm);
```
mcoker, line 3: I don't know that we want to use a spacer var here. If someone were to theme the global spacer vars, would we expect that change to impact the splitter size?

src/patternfly/components/Splitter/splitter.scss
```diff
@@ -0,0 +1,70 @@
...
+  --pf-c-splitter--Width: 100%;
+  --pf-c-splitter--BackgroundColor: var(--pf-global--BorderColor--100);
+  --pf-c-splitter--Cursor: row-resize;
+  --pf-c-splitter--m-vertical--Height: 100%;
+  --pf-c-splitter--m-vertical--Width: var(--pf-global--spacer--sm);
```
mcoker, line 8: same re: using a spacer var for the splitter sizee

src/patternfly/components/Splitter/splitter.scss
```diff
@@ -0,0 +1,70 @@
...
+  --pf-c-splitter__handle--Left: 50%;
+  --pf-c-splitter--m-vertical__handle--Top: 50%;
+  --pf-c-splitter--m-vertical__handle--Right: 0;
+  --pf-c-splitter--m-vertical__handle--Left: 0;
+  --pf-c-splitter--m-vertical__handle--Bottom: 0;
+
+  // splitter handle line
+  --pf-c-splitter__handle-line--BackgroundColor: var(--pf-global--Color--100);
+  --pf-c-splitter__handle-line--Height: #{pf-size-prem(1px)};
+  --pf-c-splitter__handle-line--Width: #{pf-size-prem(12px)};
```
mcoker, line 24: do we need the `__handle` and `__handle-line` elements? Couldn't the 2 lines be drawn with pseudo elements?

src/patternfly/components/Splitter/splitter.scss
mcoker: Could be... though I think the component should style it because it may not be absolutely positioned. You could end up creating a container for the splitter instead. That might be a better approach so that you could do something like change `--pf-c-splitter--Width` and not need to make any sort of padding or other adjustments in the component that includes it, which could be tricky since `.pf-c-drawer` (as an example) can't read `--pf-c-splitter--Width`, and automatically create space for the splitter size.

mcoker: Here's one way https://codepen.io/mcoker/pen/VwaKxQV

christiemolloy: good point, 8px is pretty small. @jessiehuff do you think we should increase the width so that there is more target area for the user ?

christiemolloy: yep, i was thinking the layout would handle that - or should i make this `position: absolute`

src/patternfly/components/Splitter/splitter.scss
christiemolloy: great point - yeah i dont think so!

src/patternfly/components/Splitter/splitter.scss
mcoker: I'm in favor of allowing the context it's used to define the element's positioning, and is inline with how the rest of our components are positioned used in another component.

mattnolting: > Here's one way https://codepen.io/mcoker/pen/VwaKxQV

That's real slick!

PR #3410, opened 2020-08-19
fix(treeview): updates to fix tree

### LINKED ISSUE #3358: Tree view follow-ups
This seems like a bit of a UX problem, if you go to click on an item, you may accidentally click the action icon when you hover and it appears.

Also this icon only shows up on hover so it isn't keyboard accessible.

Jul-31-2020 18-28-11

We should also consider a different modifier for `__node.pf-m-active` since we reserve that class for the `:active` state. We can reuse `.pf-m-current` there if we also move that class up to the `__item`. That would mimic what we're doing in the navigation component.

christiemolloy:
closes #3358 

Follow-ups to the tree view component - based on Marie's comments here https://github.com/patternfly/patternfly/issues/3358#issuecomment-668483498 
@mcarrano @doruskova @mcoker @mattnolting 

1) I removed `pf-m-active` which was for the active item (bold text for first tree node). However this was getting confused with the "current" item which is in blue text. Based on Marie's suggestion, `pf-m-current` is now the only current/active modifier, and it is applied to the tree view node that was selected last.

2) As per Marie's redesign I removed "current" from the checkbox example. The checkbox example should be used to show filtering rather than selecting another item outside of the tree view. As per the designs [here](https://marvelapp.com/prototype/8cdg211/screen/71734415), React will need to add the "indeterminate" state for the parent checkboxes @kmcfaul 

3) When checkboxes are used for filtering, the toggle should also be visible. This created some issues with what should happen when you click on the node ? Should it toggle the children items or should it select the checked item ? For this case, I updated the example so that the node container is a div, there is a new toggle button element which is a `<button>` and there is a new toggle button main element which is also a `<button>` and thats where the check and text go, so if you click on that, it will check the checked element.

4) Actions now don't appear on hover, they are always visible. But I need some guidance with how the action icons should be styled for hover, focus ? should they have a background color, should the color be black ? @doruskova

### DIFF
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -16,7 +16,7 @@ beta: true
       {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application launcher"}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application 1"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
               {{> tree-view-node tree-view-node--text="Settings"}}
@@ -77,7 +77,7 @@ beta: true
       {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application launcher"}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application 1"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
               {{> tree-view-node tree-view-node--text="Settings"}}
@@ -130,36 +130,36 @@ beta: true
 {{#> tree-view}}
   {{#> tree-view-list tree-view-list--IsRoot="true"}}
     {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-      {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Application launcher"}}
+      {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Application launcher"}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Application 1"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
-              {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
             {{/tree-view-list-item}}
             {{#> tree-view-list-item}}
-              {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--modifier="pf-m-current" tree-view-node--text="Current"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--HasCheckboxSelected="true" tree-view-node--text="Loader"}}
             {{/tree-view-list-item}}
           {{/tree-view-list}}
         {{/tree-view-list-item}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Application 2"}}
+          {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Application 2"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-              {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
             {{/tree-view-list-item}}
-            {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-              {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Loader"}}
+            {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true" tree-view-node--text="Current"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true"}}
               {{#> tree-view-list newcontext}}
                 {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-                  {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Loading app 1"}}
+                  {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--HasCheckboxSelected="true" tree-view-node--text="Loading app 1"}}
                 {{/tree-view-list-item}}
                 {{#> tree-view-list-item}}
-                  {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 2"}}
+                  {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 2"}}
                 {{/tree-view-list-item}}
                 {{#> tree-view-list-item}}
-                  {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 3"}}
+                  {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 3"}}
                 {{/tree-view-list-item}}
               {{/tree-view-list}}
             {{/tree-view-list-item}}
@@ -168,13 +168,13 @@ beta: true
       {{/tree-view-list}}
     {{/tree-view-list-item}}
     {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-      {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Cost management"}}
+      {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Cost management"}}
     {{/tree-view-list-item}}
     {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-      {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Sources"}}
+      {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Sources"}}
     {{/tree-view-list-item}}
     {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-      {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="This is a really really really long folder name that overflows from the width of the container."}}
+      {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="This is a really really really long folder name that overflows from the width of the container."}}
     {{/tree-view-list-item}}
   {{/tree-view-list}}
 {{/tree-view}}
@@ -189,7 +189,7 @@ beta: true
       {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasFolderIcon="true" tree-view-node--text="Application launcher"}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--IsActive="true" tree-view-node--HasFolderIcon="true" tree-view-node--text="Application 1"}}
+          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasFolderIcon="true" tree-view-node--text="Application 1"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
               {{> tree-view-node tree-view-node--HasFolderIcon="true" tree-view-node--text="Settings"}}
@@ -251,7 +251,7 @@ beta: true
       {{/tree-view-node}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{#> tree-view-node tree-view-node--IsToggle="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{#> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application 1"}}
             {{#> tree-view-node-count}}
               {{#> badge badge--modifier="pf-m-read"}}
                 2
@@ -359,7 +359,7 @@ beta: true
       {{/tree-view-node}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-          {{#> tree-view-node tree-view-node--IsToggle="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{#> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application 1"}}
             {{#> tree-view-action tree-view-action--attribute='aria-label="Copy"'}}
               <i class="fas fa-clipboard" aria-hidden="true"></i>
             {{/tree-view-action}}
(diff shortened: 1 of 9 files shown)


mcarrano:
Looks good @christiemolloy , but I will defer to @doruskova on the details.

doruskova:
@christiemolloy it looks great! Thanks for the update. The hover and focus should have a black background color. You can see it here in the last example: https://marvelapp.com/prototype/8cdg211/screen/68897369

src/patternfly/components/TreeView/tree-view-node-check.hbs
```diff
@@ -2,5 +2,9 @@
   {{#if tree-view-node-check--attribute}}
     {{{tree-view-node-check--attribute}}}
   {{/if}}>
-  <input type="checkbox" aria-label="Tree view node check" />
+  {{#if tree-view-node--HasCheckboxSelected}}
```
mattnolting, line 5:
```suggestion
  {{#if tree-view-node-check--IsChecked}}
```

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -130,36 +130,36 @@ beta: true
...
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Application 1"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
-              {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
             {{/tree-view-list-item}}
             {{#> tree-view-list-item}}
-              {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--modifier="pf-m-current" tree-view-node--text="Current"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--HasCheckboxSelected="true" tree-view-node--text="Loader"}}
```
mattnolting, line 142:
```suggestion
              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--checkbox--IsChecked="true" tree-view-node--text="Loader"}}
```

src/patternfly/components/TreeView/tree-view-node.hbs
```diff
@@ -1,21 +1,32 @@
...
     {{#if tree-view-node--HasFolderIcon}}
       {{> tree-view-node-icon tree-view-node-icon--IsFolder="true"}}
     {{/if}}
-    {{#if tree-view-node--text}}
-      {{#> tree-view-node-text}}
-        {{{tree-view-node--text}}}
-      {{/tree-view-node-text}}
+    {{#if tree-view-node--HasCheckbox}}
+      {{#> tree-view-node-main}}
+        {{> tree-view-node-check}}
```
mattnolting, line 17:
```suggestion
        {{> tree-view-node-check tree-view-node-check--IsChecked=tree-view-node--checkbox--IsChecked}}
```

src/patternfly/components/TreeView/tree-view-node.hbs
```diff
@@ -1,21 +1,32 @@
...
+  <{{# if tree-view-node--type}}{{tree-view-node--type}}{{else}}button{{/if}} class="pf-c-tree-view__node{{#if tree-view-node--modifier}} {{tree-view-node--modifier}}{{/if}}"
     {{#if tree-view-node--attribute}}
       {{{tree-view-node--attribute}}}
     {{/if}}>
     {{#if tree-view-node--IsToggle}}
       {{> tree-view-node-toggle-icon}}
     {{/if}}
     {{#if tree-view-node--HasCheckbox}}
-      {{> tree-view-node-check}}
+      {{> tree-view-node-toggle}}
```
mattnolting, line 10: Is this correct? Is a toggle always present if `HasCheckbox`?

mcoker:
For the checkbox example, we're only showing expandable/collapsed nodes as checked. Would it make sense to show a couple of different kinds of these? Maybe a single non-expandable node checked, and an example of a parent node checked with all of its children checked? 
Screen Shot 2020-08-27 at 10 12 59 AM

mcoker:
Should `.pf-c-tree-view__node-text` be a `<label for="input-id">` when used with an `<input type="checkbox" id="input-id">`?

mcoker:
@christiemolloy looks like the action icon needs a color change on hover/focus

Screen Shot 2020-08-27 at 12 45 05 PM

Also, following the pattern we have in select favorites, I think the background color change on focus would move from `..pf-c-tree-view__node` to `.pf-c-tree-view__content` when there are actions in the row, so it would be `.pf-c-tree-view__content:hover, .pf-c-tree-view__content:focus-within { // background color change; }`. 

That would result in this when an item in a node is focused

Screen Shot 2020-08-27 at 12 52 08 PM

Instead of this where only the focused thing in the node getting a background color

Screen Shot 2020-08-27 at 12 52 02 PM

src/patternfly/components/TreeView/tree-view-node.hbs
christiemolloy: I think this is meant to be `{{> tree-view-node-check tree-view-node-check--IsChecked="true"}}` ?

mcoker:
Screen Shot 2020-08-27 at 12 57 13 PM

The background color change on hover doesn't match the clickable area - everywhere else we have this pattern (including the non-checkbox nodes in this component), if the background color changes like this, it indicates that clicking will do something on the item. And related, the target area for these items is really small. This is out of the scope of this issue though, so we can follow up if you'd prefer do that.

src/patternfly/components/TreeView/tree-view-node.hbs
christiemolloy: Nice find! I dont know why I wrote that, removing this and changing the line above to be `HasToggle` and then showing the toggle button

christiemolloy:
> Screen Shot 2020-08-27 at 12 57 13 PM
> 
> The background color change on hover doesn't match the clickable area - everywhere else we have this pattern (including the non-checkbox nodes in this component), if the background color changes like this, it indicates that clicking will do something on the item. And related, the target area for these items is really small. This is out of the scope of this issue though, so we can follow up if you'd prefer do that.

Yep great point. @doruskova how do you think this should be designed ? Seeing as we would have to remove the background on hover and the toggle button and check would be the only active items ?

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -224,25 +224,25 @@ beta: true
...
 
-### With-badges
+### With badges
 
 ```hbs
 {{#> tree-view}}
   {{#> tree-view-list tree-view-node--IsRoot="true"}}
     {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-      {{#> tree-view-node tree-view-node--IsToggle="true" tree-view-node--text="Application launcher"}}
+      {{#> tree-view-node tree-view-node--HasToggle="true" tree-view-node--HasToggle="true" tree-view-node--text="Application launcher"}}
```
mcoker, line 245: There are a handful of this duplication in this file

```suggestion
      {{#> tree-view-node tree-view-node--HasToggle="true" tree-view-node--text="Application launcher"}}
```

src/patternfly/components/TreeView/tree-view-node-check.hbs
```diff
@@ -2,8 +2,8 @@
   {{#if tree-view-node-check--attribute}}
     {{{tree-view-node-check--attribute}}}
   {{/if}}>
-  {{#if tree-view-node--HasCheckboxSelected}}
-    <input type="checkbox" checked aria-label="Tree view node check checked" />
+  {{#if tree-view-node--IsChecked}}
+    <input type="checkbox" id="check-{{{tree-view-node--id}}}" checked aria-label="Tree view node check checked" />
```
mcoker, line 6: Is the self closing `/` needed in our workspace? I don't think it is since it isn't JSX.

```suggestion
    <input type="checkbox" id="check-{{{tree-view-node--id}}}" checked aria-label="Tree view node check checked">
```

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -50,17 +50,19 @@ $pf-c-tree-view--MaxNesting: 10;
...
   padding-top: var(--pf-c-tree-view--PaddingTop);
   padding-bottom: var(--pf-c-tree-view--PaddingBottom);
 }
 
-// stylelint-disable selector-max-class, selector-max-compound-selectors
+// stylelint-disable
 .pf-c-tree-view__list-item {
   &.pf-m-expanded {
-    > .pf-c-tree-view__content > .pf-c-tree-view__node > .pf-c-tree-view__node-toggle-icon > i {
+    > .pf-c-tree-view__content > .pf-c-tree-view__node > .pf-c-tree-view__node-toggle > .pf-c-tree-view__node-toggle-icon > i {
```
mcoker, line 65: can you just transform `.pf-c-tree-view__node-toggle-icon`? We don't use `<i>`'s in react.

mcoker: looks like react is swapping out the icon. I think we could just handle it with CSS so they don't have to do that.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -137,22 +136,32 @@ $pf-c-tree-view--MaxNesting: 10;
...
+    --pf-c-tree-view__action-item--Color: var(--pf-c-tree-view__action-item--hover--Color);
+  }
+
   &:focus {
+    --pf-c-tree-view__action-item--Color: var(--pf-c-tree-view__action-item--focus--Color);
+
     background-color: var(--pf-c-tree-view__action--focus--BackgroundColor);
   }
+
+  .pf-c-tree-view__action-item {
```
mcoker, line 162: is there a reason this should be `-item` and not `-icon`, like you did with the toggle/toggle-icon and data-list draggable-button/draggable-button-icon? Same comment here about `color` belonging on the button, and the `-item/icon` class not necessary but we can keep it. It's useful in the toggle-icon because we can apply a transform to that.

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -112,54 +112,54 @@ beta: true
...
+          {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--id="node-5" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Application 2"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
+              {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--id="node-6" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
             {{/tree-view-list-item}}
             {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true" tree-view-node--text="Current"}}
-              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true"}}
+              {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--id="node-7" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--IsChecked="true"}}
```
mcoker, line 153: Are these checked checkboxes correct? Wouldn't loading app 1 and loader app 3 also be checked if the parent was? I think in this case the way it is, current would be indeterminate? Also should they all be either "loading" or "loader", instead of both, since it's a sequential list?

Screen Shot 2020-08-28 at 4 47 10 PM

doruskova:
> > Screen Shot 2020-08-27 at 12 57 13 PM
> > The background color change on hover doesn't match the clickable area - everywhere else we have this pattern (including the non-checkbox nodes in this component), if the background color changes like this, it indicates that clicking will do something on the item. And related, the target area for these items is really small. This is out of the scope of this issue though, so we can follow up if you'd prefer do that.
> 
> Yep great point. @doruskova how do you think this should be designed ? Seeing as we would have to remove the background on hover and the toggle button and check would be the only active items ?

I think that we should follow React implementation in this, so when you click on the whole item > expanded. And when you want to select the item > click on a checkbox. If it's a leaf node > highlight it with a blue color. Like this: 
Kapture 2020-08-31 at 11 13 50

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -112,54 +112,54 @@ beta: true
...
-          {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--IsActive="true" tree-view-node--text="Application 1"}}
+          {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--IsChecked="true" tree-view-node--text="Application 1"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
-              {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--IsChecked="true" tree-view-node--text="Settings"}}
             {{/tree-view-list-item}}
             {{#> tree-view-list-item}}
-              {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--modifier="pf-m-current" tree-view-node--text="Current"}}
+              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--IsChecked="true" tree-view-node--IsChecked="true" tree-view-node--text="Loader"}}
```
mattnolting, line 142:
```suggestion
              {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node-check--IsChecked="true" tree-view-node--text="Loader"}}
```

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -112,54 +112,54 @@ beta: true
...
+          {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Application 2"}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
-              {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
+              {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Settings"}}
             {{/tree-view-list-item}}
-            {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
-              {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Loader"}}
+            {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true" tree-view-node--text="Current"}}
+              {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--IsChecked="true"}}
```
mattnolting, line 153:
```suggestion
              {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node-check--IsChecked="true"}}
```

src/patternfly/components/TreeView/examples/TreeView.md
mattnolting: I think you can just pass the `tree-view-node-check--IsChecked` through `{{> tree-view-node}}`.

src/patternfly/components/TreeView/tree-view-node-check.hbs
```diff
@@ -2,5 +2,9 @@
   {{#if tree-view-node-check--attribute}}
     {{{tree-view-node-check--attribute}}}
   {{/if}}>
-  <input type="checkbox" aria-label="Tree view node check" />
+  {{#if tree-view-node--IsChecked}}
```
mattnolting, line 5:
```suggestion
  {{#if tree-view-node-check--IsChecked}}
```

src/patternfly/components/TreeView/examples/TreeView.md
christiemolloy: Nice spot! not sure why it was there twice

christiemolloy:
@mcoker @doruskova @mattnolting updated!

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -17,22 +17,21 @@ $pf-c-tree-view--MaxNesting: 10;
...
   --pf-c-tree-view__node--hover--BackgroundColor: var(--pf-global--BackgroundColor--200);
   --pf-c-tree-view__node--focus--BackgroundColor: var(--pf-global--palette--black-200);
 
   // toggle 90degon
-  --pf-c-tree-view__node-toggle-icon--PaddingRight: var(--pf-global--spacer--sm);
-  --pf-c-tree-view__node-toggle-icon--MinWidth: var(--pf-global--spacer--lg);
-  --pf-c-tree-view__node-toggle-icon--MinHeight: var(--pf-c-tree-view__node-toggle--icon--MinWidth);
+  --pf-c-tree-view__node-toggle--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-tree-view__node-toggle--MinWidth: var(--pf-global--spacer--lg);
+  --pf-c-tree-view__node-toggle--MinHeight: var(--pf-c-tree-view__node-toggle--icon--MinWidth);
```
mattnolting, line 29:
```suggestion

// Toggle icon
--pf-c-tree-view__node-toggle--Transition: var(--pf-global--Transition);
```

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -104,13 +103,13 @@ $pf-c-tree-view--MaxNesting: 10;
   margin-right: var(--pf-c-tree-view__node-check--MarginRight);
 }
```
mattnolting, line 105:
```
.pf-c-tree-view__node-toggle-icon {
  transition: var(--pf-c-tree-view__node-toggle--Transition);
}
```

mattnolting (commented):
Looking good. I think we need to add a transition to the icon rotate to be consistent w/other expand/collapse things. 

Did we get direction on hover color of action items?

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -112,54 +112,54 @@ beta: true
...
-                  {{> tree-view-node tree-view-node--IsToggle="true" tree-view-node--HasCheckbox="true" tree-view-node--text="Loading app 1"}}
+                  {{> tree-view-node tree-view-node--HasToggle="true" tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--HasCheckboxSelected="true" tree-view-node--text="Loader app 1"}}
                 {{/tree-view-list-item}}
                 {{#> tree-view-list-item}}
-                  {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 2"}}
+                  {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node-check--IsChecked="true" tree-view-node--text="Loader app 2"}}
                 {{/tree-view-list-item}}
                 {{#> tree-view-list-item}}
-                  {{> tree-view-node tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 3"}}
+                  {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true"}}
```
mcoker, line 162:
```suggestion
                  {{> tree-view-node tree-view-node--type="div" tree-view-node--HasCheckbox="true" tree-view-node--text="Loader app 3"}}
```

src/patternfly/components/TreeView/tree-view.scss
mcoker: Looks like it's `.2s ease-in 0s` in data-list and table, but in nav, it's `var(--pf-c-nav__toggle-icon--Transition)` with default easing.

mcoker (approved):
very nice!

mattnolting (approved):
LPTM!

mcoker:
@christiemolloy do you know if we need a react issue for any of the work in this PR? If so, do you mind opening an issue?

PR #3414, opened 2020-08-20
fix(docs): compatibility for new docs

redallen:
Add compatibility for new docs.
- All MD files have an `id`
- Demo names align with component names (new Demo tab coming to staging.patternfly.org)
- Fix example CSS selectors to align with new `id`s. Since the current Gatsby site uses file names instead, temporarily rename affected files with example CSS until the workspace uses the new theme.


mcoker:
@redallen do we need to get this in for the current release?

redallen:
Yes, please! That'd be great.

src/patternfly/components/AppLauncher/examples/application-launcher.md
```diff
@@ -207,7 +207,7 @@ import './AppLauncher.css'
 
 ### Favorites
 ```hbs
-{{#> app-launcher id="app-launcher-favorites" app-launcher--IsExpanded="true" app-launcher--IsGrouped="true"}}
+{{#> app-launcher app-launcher--IsExpanded="true" app-launcher--IsGrouped="true"}}
```
mcoker, line 210: Why do we need to remove the `id`? We reference that in examples, usually to build unique `id` attribute values in the HTML. You can see that a couple of lines down in the `{{#> form-control...` block, it's used in the `concat` function

src/patternfly/components/Nav/examples/Navigation.md
```diff
@@ -4,7 +4,7 @@ section: components
 cssPrefix: pf-c-nav
 ---
 
-import './Nav.css'
+import './Navigation.css'
```
mcoker, line 7: should this be uppercase? the other updates are lowercase.

mcoker: Oh nevermind.... is it only hyphenated example CSS that is lowercase?

mcoker:
@redallen if you rebase, you'll pull in splitter, menu, and label group. Also should we update the new thing scaffolding? (`node generate NewThing`)

Issue #3119, opened 2020-05-14
Table(compound expansion) mobile view issue on FF


christiemolloy:
To recreate the issue test on mobile in the latest version of FF. The button is getting a focus-within border that is unexpected. No other buttons in FF have any focus rings.

[FF-bug.mov.zip](https://github.com/patternfly/patternfly/files/4630704/FF-bug.mov.zip)

cc: @mattnolting @mcoker

PR #3421, opened 2020-08-24
fix(table): removed compound expansion active border

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3119

@christiemolloy I'm not seeing that, but I am seeing an `:active` border that doesn't seem to match any other interactive styles we have on elements like this. WDYT?

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -682,7 +682,6 @@
     }
   }
 
-  &:active,
   &:focus-within {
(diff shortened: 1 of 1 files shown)


mattnolting (commented):
I don't see the active ring in the grid version, but I do see it in table version.

Here is the surge.sh preview vs .org
ezgif-4-0995f7a8673b

mattnolting (approved):
LGTM 👍

Issue #3429, opened 2020-08-25
Select - focused menu items not styled correctly


nicolethoen:
On the typeahead and multitypeahead select variants, if the user uses the up/down keys to move the focus on the open menu, the items do not appear focused. The styles assume that the browser focus is moving to the menu and uses :focus and :focus-within. It should also be looking for the presence of the `pf-m-focus` class on the menu item.
<img width="606" alt="Screen Shot 2020-08-25 at 1 26 19 PM" src="https://user-images.githubusercontent.com/26305082/91207353-9d9e2b80-e6d6-11ea-9dad-3487325ce1c7.png">
<img width="437" alt="Screen Shot 2020-08-25 at 1 26 29 PM" src="https://user-images.githubusercontent.com/26305082/91207354-9e36c200-e6d6-11ea-98db-c09ff2d4cd80.png">

PR #3431, opened 2020-08-25
fix(select): added pf-m-focus support for favorite __menu-items

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3429

This also adds `:focus-within` support in the app launcher to match the select, so that when you focus on an element in a favorite menu item, the whole item's background changes. This will change the current keyboard navigation style in the app launcher from:
before

to:
after

This makes the app launcher match the select, and the general style of navigating any of these types of menus by keyboard.

### DIFF
--- a/src/patternfly/components/AppLauncher/app-launcher.scss
+++ b/src/patternfly/components/AppLauncher/app-launcher.scss
@@ -44,8 +44,6 @@
   --pf-c-app-launcher__menu-item--disabled--Color: var(--pf-global--Color--dark-200);
   --pf-c-app-launcher__menu-item--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-app-launcher__menu-item--m-link--PaddingRight: 0;
-  --pf-c-app-launcher__menu-item--m-link--hover--BackgroundColor: transparent;
-  --pf-c-app-launcher__menu-item--m-action--hover--BackgroundColor: transparent;
   --pf-c-app-launcher__menu-item--m-action--Color: var(--pf-global--disabled-color--200);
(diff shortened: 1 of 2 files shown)


mcoker:
@nicolethoen @tlabaj this PR updates the CSS to work with the way the react component adds `pf-m-focus` to `pf-c-select__menu-item`. However, in the favorites select menu, the parent `li.pf-c-select__menu-wrapper` is what we style on `:hover` and `:focus-within` to create the grey background behind all of the actions in that item row. We don't apply that styling to the `pf-c-select__menu-item` element like other menus with a single action per item (when the item is the action).

If it's straight forward to do so, could we move the `pf-m-focus` class from `pf-c-select__menu-item`s to the parent `li.pf-c-select__menu-wrapper` element in the react component when navigating by keyboard? If so, I'll need to update this PR. If it's not ideal to do that we can keep it the way it is now in this PR.

nicolethoen:
> @nicolethoen @tlabaj this PR updates the CSS to work with the way the react component adds `pf-m-focus` to `pf-c-select__menu-item`. However, in the favorites select menu, the parent `li.pf-c-select__menu-wrapper` is what we style on `:hover` and `:focus-within` to create the grey background behind all of the actions in that item row. We don't apply that styling to the `pf-c-select__menu-item` element like other menus with a single action per item (when the item is the action).
> 
> If it's straight forward to do so, could we move the `pf-m-focus` class from `pf-c-select__menu-item`s to the parent `li.pf-c-select__menu-wrapper` element in the react component when navigating by keyboard? If so, I'll need to update this PR. If it's not ideal to do that we can keep it the way it is now in this PR.

This would take a little change on the react side I think, but it shouldn't be problematic.

Issue #1228, opened 2019-01-14
Documentation updates about what elements can have hover styles


jgiardino:
As part of discussion for #1211, we thought it would be worth adding something like the following to one of our documents:

> When defining hover styles, these styles should be applied to the clickable element.

And/Or

> While the default styles applied to an element might not provide a visual indication of target area, the styles that display on hover should. To ensure that these styles accurately convey to the target area of an element where the user can click, hover styles should be applied to the clickable element of a component, and not to a larger wrapping element.

I'm not sure if this is a better fit for [Contribution](https://pf-next.com/contribution) or [Guidelines](https://pf-next.com/guidelines), or somewhere else.

christiemolloy:
@lareplog consult @jgiardino if you have any questions

mcoker:
cc @jessiehuff wdyt about this, and any idea where it could go if we want to add it?

I'll also mention that we follow this as a general rule, but there are exceptions where hover styles are defined by a wrapping element when it has interactive children that consume the available space in the parent. An example is [favorites in app launcher](https://www.patternfly.org/v4/documentation/core/components/applauncher#favorites) where the item background change on hover is triggered on the parent `li`.

jessiehuff:
@mcoker Are there any set of CSS guidelines that gets maintained for others contributing to core? I think the documents that Jenn provided above have gone away, and the only locations I can find that look similar are https://github.com/patternfly/patternfly#guidelines-for-css-development and https://pf4.patternfly.org/contribution. I'm open to whatever you'd like to do for this. If there are set of guidelines shared by core contributors where it makes sense to add this, I'm open to that, or we can close this issue. However you'd prefer to handle documenting the rules core contributors are following. 🙂

mcoker:
@jessiehuff either https://pf4.patternfly.org/guidelines or https://pf4.patternfly.org/contribution is where we would put something like this. I think it would best live under the guidelines page, maybe under https://pf4.patternfly.org/guidelines#nested-calc-functions.

jessiehuff:
@mcoker Sure, that works for me. 🙂

PR #3432, opened 2020-08-25
docs(guidelines): added section for hover styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/1228

### DIFF
--- a/src/site/pages/guidelines.md
+++ b/src/site/pages/guidelines.md
@@ -355,7 +355,7 @@ PatternFly developers should avoid nested `calc()` CSS functions until this bug 
 
 #### Hover styles
 
-While the default styles applied to an element might not provide a visual indication of target area, the styles that display on hover should. To ensure that these styles accurately convey to the target area of an element where the user can click, `:hover` styles should be applied to the clickable element of a component, and not to a larger wrapping element.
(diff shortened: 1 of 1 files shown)


src/site/pages/guidelines.md
```diff
@@ -353,6 +353,10 @@ There is currently a bug in cssnano ([issue #64 on postcss-calc](https://github.
 
 PatternFly developers should avoid nested `calc()` CSS functions until this bug is resolved and the package is updated in the [patternfly repository](https://github.com/patternfly/patternfly). If you're interested in following this issue, you can do so in [issue #1295 on patternfly](https://github.com/patternfly/patternfly/issues/1295).
 
+#### Hover styles
+
+While the default styles applied to an element might not provide a visual indication of target area, the styles that display on hover should. To ensure that these styles accurately convey to the target area of an element where the user can click, `:hover` styles should be applied to the clickable element of a component, and not to a larger wrapping element.
```
mattnolting, line 358:
```suggestion
While the default styles applied to an element might not provide a visual indication of target area, the styles that display on hover should. To ensure that these styles accurately convey the target area of an element where the user can click, `:hover` styles should be applied to the clickable element of a component, and not to a larger wrapping element.
```

mattnolting (approved):
LPTM 🎉

Issue #3405, opened 2020-08-17
Modal: Add option for top-positioning


mcarrano:

katierik:
@mcarrano This is not in the docs! Is that in progress?

garrett:
Modal top alignment is mentioned in the following locations:

https://www.patternfly.org/v4/components/modal/design-guidelines/#top-aligned-modal
https://www.patternfly.org/v4/components/modal/html-demos/#top-aligned
https://www.patternfly.org/v4/components/modal#top-aligned

HTML version doesn't have an example, except in the HTML demos, but it does have the CSS mentioned under usage; https://www.patternfly.org/v4/components/modal/html/#usage:

Class | Applied | Outcome
-- | -- | --
`.pf-m-align-top` | `.pf-c-modal-box` | Modifies for top alignment.

katierik:
@garrett Thanks for clarifying - the dev who was looking had been accidentally looking at an old version of the docs. Thanks :D

PR #3435, opened 2020-08-26
feat(modal): added align-top variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3405

After looking at this some more, I think introducing a dialog layout would be too much hassle than it's worth at this point. We can just reposition the modal within the bullseye for now, and that should work fine. In an upcoming breaking change release we can re-evaluate and refactor the modal-box just to be a box and move all of the positioning, max-widths/heights, etc from it to the modal to a dialog layout.

I also evened out the outer spacing for this variation so that the top/bottom/left/right spacing around the modal is all the same.

Screen Shot 2020-08-26 at 11 53 16 AM
Screen Shot 2020-08-26 at 11 53 11 AM

### DIFF
--- a/src/patternfly/components/ModalBox/examples/ModalBox.md
+++ b/src/patternfly/components/ModalBox/examples/ModalBox.md
@@ -179,10 +179,11 @@ A modal box is a generic rectangular container that can be used to build modals.
 | `.pf-c-modal-box` | `<div>` | Initiates a modal box. **Required** |
 | `.pf-c-button.pf-m-plain` | `<button>` | Initiates a modal box close button. |
 | `.pf-c-modal-box__header` | `<header>` | Initiates a modal box header. **Required** if using a `.pf-c-modal-box__title`. |
-| `.pf-c-modal-box__title` | `<h1>`,`<h2>`,`<h3>`,`<h4>`,`<h5>`,`<h6>`, `<div>` | Initiates a modal box title. |
-| `.pf-c-modal-box__description` | `<div>` | Initiates a modal box description. A modal title is **required** if using a modal description. |
+| `.pf-c-modal-box__title` | `<h1>`,`<h2>`,`<h3>`,`<h4>`,`<h5>`,`<h6>`, `<div>` | Initiates a modal box title. **Required** if using a modal description. |
+| `.pf-c-modal-box__description` | `<div>` | Initiates a modal box description. |
 | `.pf-c-modal-box__body` | `<div>` | Initiates a modal box body. |
 | `.pf-c-modal-box__footer` | `<footer>` | Initiates a modal box footer. |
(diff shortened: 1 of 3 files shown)


garrett:
I don't know how I can check this in Cockpit or even see an isolated demo page... but when looking at the code and the screenshots, it looks like the dialog modification in this PR is stretched to the content area? 

We don't need all of our dialogs to be as large as possible. Although, I guess, these larger fullscreen dialogs _might_ be useful for some — and perhaps even for us too, if we designed for it. (This might especially be the case if we had a way to toggle between normal and maximized mode, I guess?)

What we definitely need: Dialogs that can be positioned relative to the top, so:
1. when content changes it doesn't move all the widgets
2. dialogs are more properly balanced, as they're aligned downward due to being middle-aligned in an iframe (so the heading isn't accounted for, and it shoves everything further down)

Context of the issues we're facing with the currently center-aligned dialogs in Cockpit @ https://github.com/patternfly/patternfly-design/issues/890 (with screenshots)

mcoker:
@garrett if my guess at the markup is correct, this is what the top-aligned modal in this PR will look like in a patternfly page where the content in the main section is an `iframe` - https://codepen.io/mcoker/pen/NWNvpqp

I just added the top-aligned example from https://patternfly-pr-3435.surge.sh/documentation/core/demos/modal/top-aligned as inline code for the iframe, with a `<link>` to the stylesheet from that PR. How does that look?

garrett:
@mcoker: Oh! That would look nice. 

I was thrown off by the `isFullscreen` and the screenshots... which I guess are large because of the content.

(I wonder if there's a way to view branches of PF in Cockpit. Perhaps @KKoukiou or @martinpitt might know.)

mattnolting (approved):
LPTM 👍

ziccardi:
LGTM 👍

christiemolloy (commented):
i've seen some product designs that used patternfly in the past position their modal at the very top of the page without the margin. I could definitely see this being a use case, because right now it looks like its floating especially at xl size. do you think we should support this use case?

Screen Shot 2020-09-08 at 3 43 40 PM

mcoker:
cc @mcarrano wdyt about [ christie's comment](https://github.com/patternfly/patternfly/pull/3435#pullrequestreview-484443370), should we offer a top variation with no margin

mcarrano:
> i've seen some product designs that used patternfly in the past position their modal at the very top of the page without the margin. I could definitely see this being a use case, because right now it looks like its floating especially at xl size. do you think we should support this use case?

I'm inclined to say no, @mcoker @christiemolloy . But what do you think, @mceledonia ?

christiemolloy (approved):
lgtm! we can always add a variation for the modal to sit at the very top later

PR #3438, opened 2020-08-27
feat(drilldown): added drilldown menu

### LINKED ISSUE #2635: Need support for multi level drilldown menu
 Openshift Developer Console needs multi level dropdown menu in the topology component. 

Enhancement

 Openshift Developer Console ( Targetting for release version 4.4)

mattnolting:
fixes #2635 

Parent menus control placement of children. If `pf-m-drilled-in` is applied to `pf-c-menu`, the menu shifts to the left `transform: translateX(-100%)`, out of view. The nested, children menus are positioned `left: 100%;`, until `.pf-m-drilled-in` is applied, where then `transform: translateX(-100%)` is applied.

Immediate parent `li` of the current menu has `.pf-m-current-path` applied. `pf-m-current-path` hides sibling, non-current `li` s and sets their visibility to `hidden`. We need to use visibility here because JS will need to reference height of lists when showing/hiding them.


mattnolting:
@mcoker @christiemolloy examples have included JS to replicate React state mgmt. I'll remove that before final review.

mcarrano (commented):
This is looking good @mattnolting . My only critique is that when you drill into a new level, I believe that the header should show the name of the parent node rather than just the text "Back". To return up a level, the user would click the "<" icon. Does that make sense to others?

doruskova:
@mattnolting @mcarrano Agree with showing the current level and use the back arrow for going back. However, it looks great to me. :)

mcoker:
> To return up a level, the user would click the "<" icon.

@mcarrano do you mean only the arrow would be clickable, and not the entire item? (arrow + parent node name)

mcoker:
This looks great! I'll dig in some more, but some initial questions:

1) Do we need to animate the height? or could we just animate `translate/opacity` of menus coming in/out from the left/right? That would simplify the logic and increase the performance of the animation.
2) Not sure about the width… if we keep it, need to use `pf-size-prem()`. Did you try letting the width be determined by the  contents of the menu?
3) Mentioned this outside of the issue, but I think we should be specific about the properties we animate on which items. Currently we're animating "all" on everything and on items that I wonder if need animation, like `__menu-item`.

mattnolting:
> 1. Do we need to animate the height? or could we just animate `translate/opacity` of menus coming in/out from the left/right? That would simplify the logic and increase the performance of the animation.

To accomplish the left/right transition, I had to hide overflow, limiting the viewable children to the root dropdown menu width. Which means using the parent menu as the background for all child menus. Otherwise you would see child menus fading into view to the left/right of the initial dropdown. So it would be more of a fade in/fade out transition than a slide in/slide out transition. I couldn't find another way do that, also the dropshadow interferes with transitions when using the child menu's background. 

> 2. Not sure about the width… if we keep it, need to use `pf-size-prem()`. Did you try letting the width be determined by the  contents of the menu?

Since I used the width of the parent menu for all children, I had to set a base width. We could apply that css variable to the menu component making it easily changeable and set the base size w/`pf-size-prem()`. 

> 3. Mentioned this outside of the issue, but I think we should be specific about the properties we animate on which items. Currently we're animating "all" on everything and on items that I wonder if need animation, like `__menu-item`.

Yep, agree 👍 

Here's the prototype I created if you want to experiment without : https://codepen.io/bobcatTaco/pen/pogQbmY

mcarrano:
This is looking great. I love the animation @mattnolting ! The only thing I would change is what shows up in the breadcrumb header when you drill down. Rather than being just a back link, it would contain the title of the parent node that you are viewing the contents of. The user would need to click the "<" to go back up a level. See how the Move To feature works in Google Drive, for example. Here's what that looks like:

Screen Shot 2020-09-16 at 11 27 17 AM

 Only the back-arrow is interactive.

src/patternfly/components/Menu/menu-drilldown-example.hbs
```diff
@@ -0,0 +1,70 @@
...
+                  {{> menu-list-item newcontext menu-list-item--text="Reports"}}
+                  {{> menu-list-item newcontext menu-list-item--text="Policies"}}
+                  {{> menu-list-item newcontext menu-list-item--text="Systems"}}
+                {{/menu}}
+              {{/menu-list-item}}
+            {{/menu}}
+          {{/menu-list-item}}
+          {{#> menu-list-item menu-list-item--IsCurrentReset="true" menu-list-item--IsExpandable="true" menu-list-item--text="RBAC" menu-list-item--IsDrillDown="true"}}
+            {{#> menu menu--IsSubmenu="true" menu--IsDrilledInReset="true"}}
+              {{> menu-list-item newcontext menu-list-item--IsDrillUp="true" menu-list-item--text="Edit" menu-item--data-target=menu-drilldown-example--id}}
```
mcoker, line 31: do you still need `menu-item--data-target`?

src/patternfly/components/Menu/menu-item.hbs
```diff
@@ -6,6 +6,17 @@
   {{/if}}
   {{#if menu-item--attribute}}
     {{{menu-item--attribute}}}
+  {{/if}}
+  {{#if menu-item--data-target}}
```
mcoker, line 10: same here - looks like this is no longer used, is it still needed?

mcoker (changes requested):
This looks good to me! Can you update the transitions to be more specific?

src/patternfly/components/Menu/menu.scss
```diff
@@ -89,6 +111,62 @@
...
+      }
+    }
+
+    .pf-c-menu__list {
+      position: relative;
+      transition: transform var(--pf-c-menu--m-drilled-in__list--Transition--transform);
+    }
+
+    .pf-c-menu__list-item {
+      position: static;
```
mcoker, line 151: Looks like this is only `position: relative;` to begin with to support the flyout variation. At least that's the only other place we define `position` (it's `postion: absolute;` on nested `.pf-c-menu` elements, which should be relative to `.pf-c-menu__list-item`. I bet we could remove this, and add `.pf-c-menu__list-item { postion: relative; }` in the flyout block here - https://github.com/patternfly/patternfly/blob/2f4ad756a10f0bbd1aa9b0f712e018621c166e51/src/patternfly/components/Menu/menu.scss#L84-L90

src/patternfly/components/Menu/menu.scss
```diff
@@ -76,6 +82,22 @@
   --pf-c-menu__item-action--m-favorite__icon--FontSize: var(--pf-global--icon--FontSize--sm);
   --pf-c-menu__item-action--m-favorite--m-favorited__icon--Color: var(--pf-global--palette--gold-400);
 
+  // Drilldown
+  --pf-c-menu--m-drilldown--Width: 250px;
```
mcoker, line 86: do we need to define this? The menu will be absolutely positioned so it won't be full-width when it's used. Can we just let it grow like a regular dropdown menu, and have this set to `auto` by default, so that you can override the var if you want to set a width? I would be in favor of setting `--pf-c-menu--Width/Height` just on the parent component. There is a `maxHeight` prop on the react select component that sets `max-height: {maxHeight}; overflow: auto;`. I've argued that should be a core modifier that sets a reasonable max-height by default, then is overridable via `--pf-c-menu--Height`.

mcoker (commented):
Looks great. Just a couple of suggestions, neither have to happen to get this merged.

src/patternfly/components/Menu/menu.scss
mattnolting: Nice!

mattnolting:
@mcoker updated

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -825,8 +852,12 @@ import './Menu.css'
...
-| `.pf-m-submenu` | `.pf-c-menu` | Modifies the menu to be the submenu. |
 | `.pf-m-favorite` | `.pf-c-menu__item-action` | Modifies the menu item action to handle the favorite icon. |
 | `.pf-m-favorited` | `.pf-c-menu__item-action.pf-m-favorite` | Modifies the menu item action icon to be favorited. |
 | `.pf-m-selected` | `.pf-c-menu__item` | Modifies the menu item to be selected. |
-| `.pf-m-flyout` | `.pf-c-menu` | Modifies the menu so that all nested `.pf-c-menu` elements "flyout" |
+| `.pf-m-flyout` | `.pf-c-menu` | Modifies the menu so that all nested `.pf-c-menu` elements "flyout". |
+| `.pf-m-drilldown` | `.pf-c-menu` | Modifies the menu so that all nested `.pf-c-menu` elements "drill down". |
+| `.pf-m-current-path` | `.pf-c-menu.pf-m-drilldown .pf-c-menu__list-item` | Modifies the menu list item for current path state. |
+| `.pf-m-drilled-in` | `.pf-c-menu.pf-m-drilldown, .pf-c-menu.pf-m-drilldown .pf-c-menu` | Modifies the menu list for drilled in state. |
+| `--pf-c-menu--m-drilldown--Width: {width}` | `.pf-c-menu.pf-m-drilldown` | Modifies the width of the drilldown menu. The default value is `auto`. |
```
christiemolloy, line 862: should this just be a variable ? and maybe a modifier that specifies that you are modifying the width

christiemolloy (commented):
Not sure about this ... but if the parent menus/menu-items aren't visible should they be shown like this or should they have display: none

Screen Shot 2020-10-15 at 9 53 14 AM

christiemolloy (commented):
Looking at the JS interactivity, I noticed when I click on a menu item, it opens the next menu very quickly and I feel like I lose my position, could the keyboard focus / tabindex go straight to the top of the menu?

src/patternfly/components/Menu/menu.scss
```diff
@@ -29,12 +29,13 @@
   --pf-c-menu__item--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-menu__item--PaddingBottom: var(--pf-global--spacer--sm);
   --pf-c-menu__item--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-menu__item--OutlineOffset: -2px;
```
christiemolloy, line 32: can you make this pf-size-prem

src/patternfly/components/Menu/menu.scss
```diff
@@ -43,28 +44,34 @@
...
   --pf-c-menu__item-icon--MarginRight: var(--pf-global--spacer--sm);
 
-  // toggle icon
+  // Toggle icon
   --pf-c-menu__item-toggle-icon--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-menu__item-toggle-icon--PaddingLeft: var(--pf-global--spacer--sm);
 
-  // select icon
+  // Text + toggle
+  --pf-c-menu__item-text--c-menu__item-toggle-icon--MarginLeft: var(--pf-global--spacer--sm);
```
christiemolloy, line 59: why does this have `--c-menu` in it when the menu is already in the component

src/patternfly/components/Menu/menu.scss
```diff
@@ -87,6 +111,65 @@
...
+
+    // Nested menu
+    .pf-c-menu {
+      --pf-c-menu--BoxShadow: none;
+
+      position: absolute;
+      top: calc(var(--pf-c-menu--PaddingTop) * -1);
+      left: 100%;
+      width: 100%;
+      transition: transform var(--pf-c-menu--m-drilldown__menu--Transition--transform), visibility var(--pf-c-menu--m-drilldown__menu--Transition--visibility);
```
christiemolloy, line 144: i think you want to break out transform / visibility etc like we have sone in other components e.g. Nav

src/patternfly/components/Menu/menu.scss
```diff
@@ -76,6 +83,23 @@
...
   --pf-c-menu__item-action--m-favorite--m-favorited__icon--Color: var(--pf-global--palette--gold-400);
 
+  // Drilldown
+  --pf-c-menu--m-drilldown--Width: auto;
+  --pf-c-menu--m-drilldown--Height: auto;
+  --pf-c-menu--m-drilldown--Transition--transform: var(--pf-global--TransitionDuration);
+  --pf-c-menu--m-drilldown--Transition--height: var(--pf-global--TransitionDuration);
+
+  // Drilldown menu
+  --pf-c-menu--m-drilldown__menu--Transition--transform: var(--pf-global--TransitionDuration);
```
christiemolloy, line 93: The naming should include TransitionDuration and then assign it to transform below (check out Dropdown and Navigation)

src/patternfly/components/Menu/menu.scss
```diff
@@ -76,6 +83,23 @@
...
+  --pf-c-menu--m-drilldown__menu--Transition--visibility: var(--pf-global--TransitionDuration);
```
christiemolloy, line 94: why does visibility / transform both have the same value ?

src/patternfly/components/Menu/menu.scss
```diff
@@ -87,6 +111,65 @@
...
+      }
+      // stylelint-enable
+    }
+
+    // Nested menu
+    .pf-c-menu {
+      --pf-c-menu--BoxShadow: none;
+
+      position: absolute;
+      top: calc(var(--pf-c-menu--PaddingTop) * -1);
```
christiemolloy, line 141: this should be a variable declared at the top

src/patternfly/components/Menu/menu.scss
```diff
@@ -87,6 +111,65 @@
...
+      position: relative;
+      transition: transform var(--pf-c-menu--m-drilled-in__list--Transition--transform);
+    }
+
+    .pf-c-menu__list-item {
+      transition: visibility var(--pf-c-menu--m-drilldown__list-item--Transition--visibility);
+
+      // stylelint-disable selector-max-class
+      &.pf-m-current-path .pf-c-menu {
+        z-index: var(--pf-c-menu--m-drilled-in__c-menu__list-item--m-current-path__c-menu--ZIndex);
```
christiemolloy, line 161: not sure why this variable name needs the `c` before the menu

src/patternfly/components/Menu/menu.scss
```diff
@@ -213,6 +295,14 @@
   padding-left: var(--pf-c-menu__item-toggle-icon--PaddingLeft);
 }
 
+.pf-c-menu__item-text + .pf-c-menu__item-toggle-icon {
+  margin-left: var(--pf-c-menu__item-text--c-menu__item-toggle-icon--MarginLeft);
```
christiemolloy, line 299: could this be

```suggestion
  margin-left: var(--pf-c-menu__item-text__item-toggle-icon--MarginLeft);
```

christiemolloy (changes requested):

src/patternfly/components/Menu/examples/Menu.md
mattnolting: `--pf-c-menu--m-drilldown--Width` is set to auto be default, so the menu will be as wide as the first parent list. All nested menus will also be as wide as the first parent list because it provides the background/drop shadow styling. The width of the dropdown can be updated inline by modifying `--pf-c-menu--m-drilldown--Width`. Using this method means that devs can use any width value they want, rather than relying on a predefined class.

mattnolting:
> Not sure about this ... but if the parent menus/menu-items aren't visible should they be shown like this or should they have display: none
> 
> Screen Shot 2020-10-15 at 9 53 14 AM

Can't hide the parent and see it's children. I wrestled with this as well, which is why I'm using visibility.

src/patternfly/components/Menu/menu.scss
mattnolting: As in this? 

`--pf-c-menu--m-drilldown--c-menu--Transition: transform var(--pf-global--TransitionDuration), visibility var(--pf-global--TransitionDuration);`

`transition: var(--pf-c-menu--m-drilldown--c-menu--Transition);`

mcoker: I would probably keep the component var you have for the transition-duration, so a user can override that without having to override the entire `transition` prop value. 

```
transition: transform var(--pf-c-menu--m-drilldown__menu--TransitionDuration--transform), visibility var(--pf-c-menu--m-drilldown__menu--TransitionDuration--visibility);
```

That's how we did it in tabs https://github.com/patternfly/patternfly/blob/0d2c0bfb9210fa7c19ddd331fd40073b2f713bea/src/patternfly/components/Tabs/tabs.scss#L435

mattnolting:
> Looking at the JS interactivity, I noticed when I click on a menu item, it opens the next menu very quickly and I feel like I lose my position, could the keyboard focus / tabindex go straight to the top of the menu?

Great point, i'm not sure how to do this statically. I'd like to shift this to React.

mattnolting:
@christiemolloy @mcoker updated

src/patternfly/components/Divider/divider.hbs
```diff
@@ -2,6 +2,7 @@
 {{#if divider--attribute}}
   {{divider--attribute}}
 {{/if}}
+aria-hidden="true"
```
mcoker, line 5: did you mean to add this?

src/patternfly/components/Menu/menu.scss
```diff
@@ -76,6 +83,26 @@
...
+  --pf-c-menu--m-drilldown--Transition: transform var(--pf-c-menu--m-drilldown--TransitionDuration--transform), height var(--pf-c-menu--m-drilldown--TransitionDuration--height);
+
+  // Nested menu
+  --pf-c-menu--m-drilldown--c-menu--Top: calc(var(--pf-c-menu--PaddingTop) * -1);
+  --pf-c-menu--m-drilldown--c-menu--TransitionDuration--transform: var(--pf-global--TransitionDuration);
+  --pf-c-menu--m-drilldown--c-menu--TransitionDuration--visibility: var(--pf-global--TransitionDuration);
+  --pf-c-menu--m-drilldown--c-menu--Transition: transform var(--pf-c-menu--m-drilldown--c-menu--TransitionDuration--transform), visibility var(--pf-c-menu--m-drilldown--c-menu--TransitionDuration--visibility);
+
+  // Drilldown list
+  --pf-c-menu--m-drilldown__list--TransitionDuration--transform: transform var(--pf-global--TransitionDuration);
```
mcoker, line 100:
```suggestion
  --pf-c-menu--m-drilldown__list--TransitionDuration--transform: var(--pf-global--TransitionDuration);
```

mcoker (changes requested):
Just one thing to fix a broken transition, and a question about the divider.

src/patternfly/components/Divider/divider.hbs
mattnolting: I did, but maybe better addressed in another issue :)

christiemolloy (approved):
LGTM!!!

mcoker (approved):
NICE!!!

PR #3439, opened 2020-08-27
feat(Text): WIP utility class

### LINKED ISSUE #3246: Enhancement(Text): Add size & color mods to Text component
This request is to add size and color modifications to the Text component, stemming from a user request to be able to easily update the font color:

Screen Shot 2020-07-06 at 9 53 03 AM

jenny-s51:
Will close #3246


src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+
+@include pf-utility-builder($pf-u-font-size-options, $pf-global--breakpoint-list);
+
+
+// Font weight options
+$pf-u-font-weight-options: (
+  font-weight-light:              (font-weight var(--pf-global--FontWeight--light)),
+  font-weight-normal:             (font-weight var(--pf-global--FontWeight--normal)),
+  font-weight-bold:               (font-weight var(--pf-global--FontWeight--bold)),
+  font-weight-semi-bold:          (font-weight var(--pf-global--FontWeight--semi-bold)),
```
mcoker, line 23: I don't think we need semi-bold or overpass-*, as those are only for overpass, which we don't need to support in the utility.

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  font-color-300: (color var(--pf-global--Color--300)),
+  font-color-400: (color var(--pf-global--Color--400)),
+);
+
+@include pf-utility-builder($pf-u-font-color-options, $pf-global--breakpoint-list);
+
+
+// Background color options
+$pf-u-font-background-color-options: (
+
```
mcoker, line 44: this should be a separate "background-color" utility

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  /* Background color */
+  background-color-100:  (background-color var(--pf-global--BackgroundColor--100)),
+  background-color-200:  (background-color var(--pf-global--BackgroundColor--200)),
+
+  /*light theme*/
```
mcoker, line 49: maybe we just start with `background-color-100/200` and add the rest if we get specific requests for them.

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  background-color-dark-transparent-100:  (background-color var(--pf-global--BackgroundColor--dark-transparent-100)),
+  background-color-dark-transparent-200:  (background-color var(--pf-global--BackgroundColor--dark-transparent-200))
+);
+
+@include pf-utility-builder($pf-u-font-background-color-options, $pf-global--breakpoint-list);
+
+
+// Font layout options
+$pf-u-font-alignment-options: (
+  font-truncate: var(--pf-m-truncate),
```
mcoker, line 68: you might look at the accessibility utility for how to setup some of these since they will need a number of CSS properties.

mcoker: Also we have a mixin for this block of CSS @ https://github.com/patternfly/patternfly/blob/c4ab7f92b852d40b45cab7eb9bf45cae6adc3a52/src/patternfly/sass-utilities/mixins.scss#L149-L153

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  font-break-word: var(--pf-m-break-word),
```
mcoker, line 69: this would apply `word-break: break-word;`

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  font-nowrap: var(--pf-m-nowrap),
```
mcoker, line 70: `white-space: nowrap;`

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  font-wrap: var(--pf-m-wrap),
```
mcoker, line 71: `white-space: wrap;`. I think this would only be needed to undo `white-space: nowrap;`

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,77 @@
...
+  font-fit-content: var(--pf-m-fit-content)
```
mcoker, line 72: I'm not sure about `fit-content` here... since that's a value of `width/height` or a size function used to define grid tracks, I think that's more appropriate under the sizing utility. Though it doesn't work in FF yet so I don't know if we need it yet.

mcoker:
If you want to test the compiled CSS, you can run `npm run build-patternfly`, then look in `dist/utilities/[utility]/[utility.css]`

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,49 @@
...
+
+@include pf-utility-builder($pf-u-font-weight-options, $pf-global--breakpoint-list);
+
+
+// Font color options
+$pf-u-font-color-options: (
+  font-color-100: (color var(--pf-global--Color--100)),
+  font-color-200: (color var(--pf-global--Color--200)),
+  font-color-300: (color var(--pf-global--Color--300)),
+  font-color-400: (color var(--pf-global--Color--400)),
```
mcoker, line 33: Can you replace `font-color-` with `color-` so it matches the property name?

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
+---
+id: Text
+section: utilities
+---
+
+import './Text.css'
+
+## Examples
+
+### Sizing
```
mcoker, line 10: Can you call this Font size?

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+{{/text}}
+{{#> text text--modifier="pf-u-font-size-3xl"}}
+  3xl text
+{{/text}}
+{{#> text text--modifier="pf-u-font-size-4xl"}}
+  4xl text
+{{/text}}
+```
+
+### Font Weight
```
mcoker, line 39:
```suggestion
### Font weight
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+{{/text}}
+{{#> text text--modifier="pf-u-font-weight-normal"}}
+  Normal
+{{/text}}
+{{#> text text--modifier="pf-u-font-weight-bold"}}
+  Bold
+{{/text}}
+```
+
+### Font Color
```
mcoker, line 53:
```suggestion
### Color
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+{{/text}}
+{{#> text text--modifier="pf-u-color-300"}}
+  Font color 300
+{{/text}}
+{{#> text text--modifier="pf-u-color-400"}}
+  Font color 400
+{{/text}}
+```
+
+### Font Alignment
```
mcoker, line 70: Maybe rename this to align with the table's description of these modifiers? https://pf4.patternfly.org/documentation/core/components/table#controlling-text

We call it "text modifiers" in the data list but that seems a little generic of a label to use here https://pf4.patternfly.org/documentation/core/components/datalist#text-modifiers

```suggestion
### Controlling text
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+| ------------------------- | ---------- | ----------------------------- |
+| `.pf-u-font-size-xs`      | `*`        | Sets font-size to xs          |
+| `.pf-u-font-size-sm`      | `*`        | Sets font-size to sm          |
+| `.pf-u-font-size-md`      | `*`        | Sets font-size to md          |
+| `.pf-u-font-size-lg`      | `*`        | Sets font-size to lg          |
+| `.pf-u-font-size-xl`      | `*`        | Sets font-size to xl          |
+| `.pf-u-font-size-2xl`     | `*`        | Sets font-size to 2xl         |
+| `.pf-u-font-size-3xl`     | `*`        | Sets font-size to 3xl         |
+| `.pf-u-font-size-4xl`     | `*`        | Sets font-size to 4xl         |
+| `.pf-u-font-weight-light` | `*`        | Sets font-weight to light     |
```
mcoker, line 109:
```suggestion
| `.pf-u-font-weight-light` | `*`        | Sets font-weight to light     |
| `.pf-u-font-weight-normal` | `*`        | Sets font-weight to normal     |
| `.pf-u-font-weight-bold` | `*`        | Sets font-weight to bold     |
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+| `.pf-u-font-size-normal`  | `*`        | Sets font-weight to normal    |
```
mcoker, line 110:
```suggestion
| `.pf-u-font-weight-normal`  | `*`        | Sets font-weight to normal    |
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+| `.pf-u-font-size-bold`    | `*`        | Sets font-weight to bold      |
```
mcoker, line 111:
```suggestion
| `.pf-u-font-weight-bold`    | `*`        | Sets font-weight to bold      |
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+| `.pf-u-color-100`         | `*`        | Sets font-color to Color 100  |
+| `.pf-u-color-200`         | `*`        | Sets font-color to Color 200  |
+| `.pf-u-color-300`         | `*`        | Sets font-color to Color 300  |
+| `.pf-u-color-400`         | `*`        | Sets font-color to Color 400  |
+| `.pf-u-font-break-word`   | `*`        | Sets word-break to break-word |
```
mcoker, line 116:
```suggestion
| `.pf-u-text-break-word`   | `*`        | Sets word-break to break-word |
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+| `.pf-u-font-nowrap`       | `*`        | Sets white-space to nowrap    |
```
mcoker, line 117:
```suggestion
| `.pf-u-text-nowrap`       | `*`        | Sets white-space to nowrap    |
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,118 @@
...
+| `.pf-u-font-wrap`         | `*`        | Sets white-space to normal    |
```
mcoker, line 118:
```suggestion
| `.pf-u-text-wrap`         | `*`        | Sets white-space to normal    |
```

mcoker: On these, can we call the utilities `text-[variation]`, and update the CSS?

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,47 @@
...
+
+@include pf-utility-builder($pf-u-font-weight-options, $pf-global--breakpoint-list);
+
+
+// Font color options
+$pf-u-font-color-options: (
+  color-100: (color var(--pf-global--Color--100)),
+  color-200: (color var(--pf-global--Color--200)),
+  color-300: (color var(--pf-global--Color--300)),
+  color-400: (color var(--pf-global--Color--400)),
```
mcoker, line 33: I think we'll also want to add

* pf-global--icon--Color--light
* pf-global--icon--Color--dark
* pf-global--link--Color
* pf-global--active-color--[100-400]
* pf-global--disabled-color--[100-300]
* pf-global--primary-color--100
* pf-global--primary-color--200
* pf-global--secondary-color--100
* pf-global--default-color--100
* pf-global--default-color--200
* pf-global--default-color--300
* pf-global--success-color--100
* pf-global--success-color--200
* pf-global--info-color--100
* pf-global--info-color--200
* pf-global--warning-color--100
* pf-global--warning-color--200
* pf-global--danger-color--100
* pf-global--danger-color--200
* pf-global--danger-color--300

We may consider adding the palette vars and light/dark theme vars, but that can probably wait until someone requests them.

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,47 @@
...
+);
+
+@include pf-utility-builder($pf-u-font-color-options, $pf-global--breakpoint-list);
+
+// Font layout options
+$pf-u-font-alignment-options: (
+  font-break-word: (word-break break-word),
+  font-nowrap:     (white-space nowrap),
+  font-wrap:       (white-space normal),
```
mcoker, line 42: Were you able to get truncate working?

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,47 @@
...
+);
+
+@include pf-utility-builder($pf-u-font-alignment-options, $pf-global--breakpoint-list);
+
```
mattnolting, line 46: I'd like to see truncate added, it would look a little different.. I don't know how I feel about making all of the values `!important`.. @christiemolloy @mcoker wdyt?

```

.pf-u-text-truncate {
  min-width: 0 !important;
  max-width: 100% !important;
  overflow: hidden !important;
  text-overflow: ellipsis !important;
  white-space: nowrap !important;
  display: inline-block !important;
}
```

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.css
```diff
@@ -0,0 +1,10 @@
+#ws-core-u-backgroundcolor-white {
+  margin-bottom: var(--pf-global--spacer--lg);
+  background-color: var(--pf-global--BackgroundColor--100);
+
+}
+#ws-core-u-backgroundcolor-gray {
+  margin-bottom: var(--pf-global--spacer--lg);
+  background-color: var(--pf-global--BackgroundColor--200);
+
+}
```
mattnolting, line 10:
```suggestion
}

```

src/patternfly/utilities/BackgroundColor/BackgroundColor.scss
```diff
@@ -0,0 +1,12 @@
+// stylelint-disable
+
+// Background color options
+$pf-u-background-color-options: (
+  /* Background color */
+  background-color-100:  (background-color var(--pf-global--BackgroundColor--100)),
+  background-color-200:  (background-color var(--pf-global--BackgroundColor--200)),
+);
```
mattnolting, line 8: I think we should add these colors, as defined by design, to utility background colors:

Screen Shot 2020-09-16 at 3 48 58 PM

mattnolting (commented):
This is looking great @jenny-s51!

src/patternfly/utilities/Text/text.scss
mcoker: I think we would add at least what's in our mixin - white-space, overflow, text-overflow. You need those at a minimum, and that should handle most cases. I wouldn't add display though since we can't predictably change that, so I think it can be up to the user to make sure their element is block, kind of like needing a block element with something like a block/vertical margin.

Though I would add `!important` since utilities are designed to be added to anything and we never know what we're styling. They're basically like writing an inline style with a class. What's your hesitation?

src/patternfly/patternfly-addons.scss
```diff
@@ -9,3 +9,4 @@
 @import "./utilities/Float/float.scss";
 @import "./utilities/Sizing/sizing.scss";
 @import "./utilities/Spacing/spacing.scss";
+@import "./utilities/Text/text.scss";
```
mcoker, line 12: can you add the bg color utility here?

src/patternfly/utilities/Text/text.scss
mattnolting: > Though I would add !important since utilities are designed to be added to anything and we never know what we're styling. They're basically like writing an inline style with a class. What's your hesitation?

Good point, 🚢  it!

mattnolting:
What do you think about adding `font-size` utilities as well?

```
// Icon size options
$pf-u-icon-size-options: (
  icon-size-sm:           (font-size var(--pf-global--icon--FontSize--sm)),
  icon-size-md:           (font-size var(--pf-global--icon--FontSize--md)),
  icon-size-lg:           (font-size var(--pf-global--icon--FontSize--lg)),
  icon-size-xl:           (font-size var(--pf-global--icon--FontSize--xl)),
);

@include pf-utility-builder($pf-u-icon-size-options, $pf-global--breakpoint-list);
```

src/patternfly/utilities/Text/text.scss
jenny-s51: @mcoker just committed with @mattnolting 's suggested `pf-u-text-truncate` utility class:

```
.pf-u-text-truncate {
  min-width: 0 !important;
  max-width: 100% !important;
  overflow: hidden !important;
  text-overflow: ellipsis !important;
  white-space: nowrap !important;
}
```

Looks like the truncation aspect works, but the ellipsis is missing. I thought it was because the width is constrained here by percent rather than pixel, but after changing max-width to `100px` to test, the ellipsis is still MIA. Do you guys have any other ideas why this might be happening?

mattnolting: `pf-u-truncate` needs to be applied to the individual element (`div, p, h1-2-3..., etc`) rather than the parent div.

jenny-s51: @mattnolting totally makes sense - thank you!

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.css
```diff
@@ -0,0 +1,28 @@
+#ws-core-u-background-color-white {
```
mcoker, line 1: I don't think we need to set the background on the example container. I would be in favor of just using the utility class on the example like you have it and removing the CSS from the example container. Even when it's a white on white background, or black background with black text. It could be misleading as to what the utility does to present it differently. Maybe in some of the examples, we use a `pf-u-color` utility to also change the text color so it shows up against the background?

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Overview
+
+Background color utility
+
+### Usage
+
+| Class                             | Applied to | Outcome                            |
+| --------------------------------- | ---------- | ---------------------------------- |
+| `.pf-u-background-color-100`      | `*`        | Applies background-color 100.      |
```
mcoker, line 68: for these could you update the outcome to, assuming the SCSS will output the hex color

```suggestion
| `.pf-u-background-color-100`      | `*`        | Applies `--pf-global--BackgroundColor--100` (#{$pf-global--BackgroundColor--100})      |
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,128 @@
...
+{{#> text text--modifier="pf-u-color-400"}}
+  Font color 400
+{{/text}}
+```
+
+### Controlling text
+
+```hbs
+<h4><strong>Breakword</strong></h4>
+{{#> text text--modifier="pf-u-font-break-word"}}
```
mcoker, line 74: Need to update the utility name of the "controlling text" utilitys to `pf-u-text` like you have in the docs.

```suggestion
{{#> text text--modifier="pf-u-text-break-word"}}
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Overview
+
+Text utility
+
+### Usage
+
+| Class                      | Applied to | Outcome                       |
+| -------------------------- | ---------- | ----------------------------- |
+| `.pf-u-font-size-xs`       | `*`        | Sets font-size to xs          |
```
mcoker, line 110: same thing for these docs? wdyt?

```suggestion
| `.pf-u-font-size-xs`       | `*`        | Sets font size to `--pf-global--FontSize--sm` (#{$pf-global--FontSize--sm})         |
```

src/patternfly/utilities/Text/text.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="{{#if text--modifier}} {{text--modifier}}{{/if}}"
```
mcoker, line 1: we usually write these like this in case no `text--modifier` is passed, you won't end up with `class=""` in the markup

```suggestion
<div{{#if text--modifier}}class="{{text--modifier}}"{{/if}}
```

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,164 @@
...
+    white-space nowrap
+  ),
+  font-wrap: (
+    white-space normal
+  )
+);
+
+@include pf-utility-builder($pf-u-font-alignment-options, $pf-global--breakpoint-list);
+
+.pf-u-text-truncate > * {
```
mcoker, line 156: this should apply to the element the class is applied to like the other utility classes.

src/patternfly/utilities/Text/text.scss
```diff
@@ -0,0 +1,164 @@
...
+  color-200: (
+    color var(--pf-global--Color--200)
+  ),
+  color-300: (
+    color var(--pf-global--Color--300)
+  ),
+  color-400: (
+    color var(--pf-global--Color--400)
+  ),
+  icon-color-light: (
```
mcoker, line 62: there are a lot more colors here than are in the docs. Can you make sure the docs are updated to reflect all of the classes?

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.css
jenny-s51: @mcoker thank you for your feedback -- I initially had tried to just use the utility class on the example, but the background color gets overwritten by the `background-color` of `.ws-preview`, which is set to white, so I had to set it manually on each example container. If you have any ideas for another workaround I'd be happy to make any necessary modifications! what do you think would be the best way to go here?

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.md
jenny-s51: Markdown doesn't like this syntax 🙁

mcoker: Oh sorry, I was confusing mdx and SCSS. Just `--pf-global--BackgroundColor--100` is fine then!

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.css
mcoker: Oh you can just add a div in the example and give it the class of the utility class. Like this

Screen Shot 2020-10-19 at 1 52 43 PM

jenny-s51: oh excellent - done. thank you!

src/patternfly/utilities/BackgroundColor/examples/BackgroundColor.md
```diff
@@ -0,0 +1,93 @@
+---
+id: Background color
+section: utilities
+---
+
+import './BackgroundColor.css'
```
mcoker, line 6: Can you remove this line? It's causing the build to fail.

mcoker (approved):
thanks @jenny-s51!

Issue #3440, opened 2020-08-28
Page main sections shrinking unexpectedly in safari


mcoker:
In safari we have an issue with page main sections that are set to `display: flex` shrinking and floating on top of one another instead of stacking and overflowing the main container.

A simple repro is here where the page sections are set to `.pf-m-limit-width` (which sets `display: flex;`). https://codepen.io/mcoker/pen/JjXJKxz

The fix is to ensure the main section is set to `flex-shrink: 0; flex-basis: auto;` as outlined here https://github.com/philipwalton/flexbugs#flexbug-1

PR #3441, opened 2020-08-28
fix(page): fixed safari bug w/ sections shrinking shorter than content

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3440

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -387,11 +387,11 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
   &:last-of-type,
   &:only-child,
   &.pf-m-fill {
-    flex: 1;
+    flex-grow: 1;
   }
 
   &.pf-m-no-fill {
-    flex: 0 0 auto;
+    flex-grow: 0;
   }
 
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -438,6 +438,8 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
 .pf-c-page__main-breadcrumb,
 .pf-c-page__main-section,
 .pf-c-page__main-wizard {
+  flex-shrink: 0;
```
christiemolloy, line 441: trying to follow along here so this may be incorrect, but from your codepen you used `flex-basis: auto;` , i dont see that anywhere here do you need that too ?

christiemolloy (changes requested):

src/patternfly/components/Page/page.scss
mcoker: Ah yeah, `flex-basis` on those elements is already `auto`, so it's not necessary to set it specifically. The elements just need `flex-shrink: 0` with `flex-basis: auto`

christiemolloy (approved):
LGTM!

mattnolting (approved):
Nice!

PR #3445, opened 2020-08-31
fix(chipgroup): position removable button at the top

### LINKED ISSUE #3182: ChipGroups: close chip group button wraps at smaller widths

christiemolloy:
closes #3182

### DIFF
--- a/src/patternfly/components/ChipGroup/chip-group.scss
+++ b/src/patternfly/components/ChipGroup/chip-group.scss
@@ -13,9 +13,6 @@
   --pf-c-chip-group--m-category--BorderRadius: var(--pf-global--BorderRadius--sm);
   --pf-c-chip-group--m-category--BackgroundColor: var(--pf-global--BackgroundColor--200);
 
-  // Removable
-  --pf-c-chip-group--m-category--m-removable--PaddingRight: var(--pf-global--spacer--2xl);
-
   // Label in group
   --pf-c-chip-group__label--MarginRight: var(--pf-global--spacer--sm);
   --pf-c-chip-group__label--FontSize: var(--pf-global--FontSize--sm);
@@ -32,6 +29,9 @@
   --pf-c-chip-group__list-item--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-chip-group__list-item--MarginBottom: var(--pf-global--spacer--xs);
 
+  // Main
+  --pf-c-chip-group__main--MarginBottom: calc(var(--pf-global--spacer--xs) * -1);
+
   position: relative;
 
   &.pf-m-category {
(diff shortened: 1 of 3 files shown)


mcoker:
My only hesitation is with absolute positioning. If we chose to add another action, or allow a user to pass some other action or element that stays with the close button, we'll run into issues allocating space for those elements. @mcarrano @mceledonia do you see that as a possibility?

Also looking at these 3 patterns, what's the ideal wrapping behavior?

mcarrano:
I can't think of what additional elements we might add.

mattnolting:
@mcoker great point. I think that ideally the label and list would be in their own container, with `__actions` being a sibling. Something like: 
```
.pf-c-chip-group
  .pf-c-chip-group__main(.pf-m-stack.pf-m-inline-on-lg)
    .pf-c-chip-group__label
    .pf-c-chip-group__list
  .pf-c-chip-group__actions
    .pf-c-button(kebab)
    .pf-c-button(close)
    .pf-c-button(copy or something)
```
That would allow us to control when the label and list wrap or stack and allow extensibility. It would also add the necessary structure to handle when to stack labels and lists and when to present them inline, highlighted in the following SS.

Screen Shot 2020-09-03 at 8 57 56 AM

I think it's important for us to consider that, with most of our components, we can't really forecast use cases, so flexibility is crucial. I expect that our consumers will need an extended chip group with multiple actions, in addition to `close`.

We could introduce the new structure as a bug fix, without affect current behavior. @mcoker @christiemolloy wdyt?

mcoker:
@mcarrano do you have an opinion on the ideal wrapping behavior from the gifs above?

mcarrano:
@mcoker I personally like the second example with the title sitting above the chips. If we go this way and there is a Close button, I would expect to see it right aligned in the same row as the category title.

src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -26,6 +26,8 @@
   --pf-c-chip-group__list-item--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-chip-group__list-item--MarginBottom: var(--pf-global--spacer--xs);
 
+  position: relative;
```
mattnolting, line 29: don't need this anymore

src/patternfly/components/ChipGroup/examples/ChipGroup.md
```diff
@@ -118,198 +118,306 @@ cssPrefix: pf-c-chip-group
...
+        {{/chip}}
+      {{/chip-group-list-item}}
+    {{/chip-group-list}}
+  {{/chip-group-main}}
 {{/chip-group}}
 ```
 
 ### Chip group with categories removable
 ```hbs
 {{#> chip-group chip-group--modifier="pf-m-category" chip-group--id="Chip-group-with-categories-removable"}}
```
mattnolting, line 283:
```suggestion
{{#> chip-group chip-group--modifier="pf-m-category" chip-group--id="chip-group-with-categories-removable"}}
```

mattnolting (approved):
This looks great! Just a nit and an observation :)

src/patternfly/components/ChipGroup/examples/ChipGroup.md
```diff
@@ -118,198 +118,306 @@ cssPrefix: pf-c-chip-group
...
+            <i class="fas fa-times" aria-hidden="true"></i>
+          {{/button}}
+        {{/chip}}
+      {{/chip-group-list-item}}
+    {{/chip-group-list}}
+  {{/chip-group-main}}
+  {{> chip-group-close}}
+{{/chip-group}}
+
+Outdated - used for testing purposes.
```
mcoker, line 354:
```suggestion
<br><br>
#### Outdated - used for testing purposes.
```

mcoker (approved):
nice!!

PR #3457, opened 2020-09-04
feat(codeeditor): add code editor

### LINKED ISSUE #3419: Code editor/snippet (layout and styling)
Provide styling for elements that will wrap the Monaco code editor as defined in the POC here: https://github.com/patternfly/patternfly-react/issues/4514

Designs live here: https://marvelapp.com/prototype/b284chd  This should include elements that are added to the editor like tool buttons and language tab.

christiemolloy:
closes #3419


mcarrano (commented):
@christiemolloy This is looking good. Just a few questions, comments. Would also like @nicolethoen to take a look since she did the React PoC with the Monaco editor.

* I feel like the component should be called "Code snippet" vs "Code editor" because it supports both editing and read-only use cases.
* The `pf-c-code-editor__tab` presents an interesting question since that is actually presented by the editor itself. @nicolethoen will this work as currently structured?
* When the code snippet is displayed as read-only, it will take on different styling. Not sure if that's left to the editor or we need to apply the desired CSS. 
* I feel like the tool buttons should have tooltips. Would that be added in the core example or left to be coded in React?

src/patternfly/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -0,0 +1,140 @@
...
+## Examples
+### Default
+```hbs
+{{#> code-editor}}
+  {{#> code-editor-control}}
+    {{#> code-editor-action-group}}
+      {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Copy to clipboard"'}}
+        <i class="fas fa-copy" aria-hidden="true"></i>
+      {{/button}}
+      {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Download code"'}}
```
mcoker, line 17: Using an input group would get rid of the double border between buttons

Screen Shot 2020-09-08 at 5 44 16 PM

mcoker:
This is more for design, but these buttons feel a bit disjointed with the different border colors (both bottom and the top/right/left), and how the bottom border changes width on hover/focus

Screen Shot 2020-09-08 at 5 45 19 PM

src/patternfly/components/CodeEditor/code-editor-action-group-control.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-code-editor__action-group-control{{#if code-editor-action-group-control--modifier}} {{code-editor-action-group-control--modifier}}{{/if}}"
```
mcoker, line 1: What does this element do?

mcoker, line 1: what does this element do?

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+  --pf-c-code-editor__tab--before--BorderRightColor: var(--pf-global--BorderColor--100);
+  --pf-c-code-editor__tab--before--BorderBottomColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-code-editor__tab--before--BorderLeftColor: var(--pf-global--BorderColor--100);
+
+  width: 100%;
+}
+
+.pf-c-code-editor__control {
+  position: relative;
+  height: var(--pf-c-code-editor__control--Height);
```
mcoker, line 36: I'm not sure why you have the children of `__control` as `position: absolute`... that requires this `height`, and it also means that if the height of the children change for whatever reason, this `height` needs to change.

Couldn't this element just be a flex layout, with `align-items: flex-end`, and `margin-left: auto` on the tab element to push it to the right side? You shouldn't need a z-index on the children, either. You could just add `position: relative` to this element, the remove `position: relative` on the `__code` element. That will mean you should just use a regular border on the `__code` element instead of a pseudo element, but that's preferred anyways. Currently the border is overlapping the content, and you can't click into the `__code` element because the pseudo element is overlaid on top of it.

Screen Shot 2020-09-08 at 5 58 54 PM

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+
+.pf-c-code-editor__tab {
+  position: absolute;
+  right: 0;
+  z-index: var(--pf-c-code-editor__tab--ZIndex);
+  padding: var(--pf-c-code-editor__tab--PaddingTop) var(--pf-c-code-editor__tab--PaddingRight) var(--pf-c-code-editor__tab--PaddingBottom) var(--pf-c-code-editor__tab--PaddingLeft);
+  color: var(--pf-c-code-editor__tab--Color);
+  background-color: var(--pf-c-code-editor__tab--BackgroundColor);
+
+  &::before {
```
mcoker, line 81: you could use a border here instead of a pseudo element, too. This pseudo element is keeping you from being able to click on the content in the tab.

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+  --pf-c-code-editor__tab--before--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-code-editor__tab--before--BorderTopColor: var(--pf-global--BorderColor--100);
+  --pf-c-code-editor__tab--before--BorderRightColor: var(--pf-global--BorderColor--100);
+  --pf-c-code-editor__tab--before--BorderBottomColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-code-editor__tab--before--BorderLeftColor: var(--pf-global--BorderColor--100);
+
+  width: 100%;
+}
+
+.pf-c-code-editor__control {
```
mcoker, line 34: I'm not sure about the name `__control` since it's a container for the stuff at the top, including the tab. It might be better labeled `__header` maybe? Then you could rename `__action-group` to `__control` or `__controls`.

src/patternfly/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -0,0 +1,140 @@
...
+        <i class="fas fa-download"></i>
+      {{/button}}
+      {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Upload code"'}}
+        <i class="fas fa-upload"></i>
+      {{/button}}
+      {{#> code-editor-action-group-control}}
+        {{#> overflow-menu overflow-menu--id="overflow-menu-simple"}}
```
mcoker, line 52: Could you show an example that shows how the overflow menu is used before items have collapsed into a dropdown?

mcarrano:
> This is more for design, but these buttons feel a bit disjointed with the different border colors (both bottom and the top/right/left), and how the bottom border changes width on hover/focus
> 
> Screen Shot 2020-09-08 at 5 45 19 PM

This didn't bother me @mcoker but what do you think @gdoyle1 ?

src/patternfly/components/CodeEditor/code-editor-action-group-control.hbs
christiemolloy: It's a container around the overflow kebab element - I thought we should have a container in case we style the element later on. I renamed it `pf-c-code-editor__action-group-overflow`. wdyt?

src/patternfly/components/CodeEditor/code-editor.scss
christiemolloy: ok that sounds good to me

src/patternfly/components/CodeEditor/examples/CodeEditor.md
christiemolloy: what do you mean by this?

christiemolloy:
> @christiemolloy This is looking good. Just a few questions, comments. Would also like @nicolethoen to take a look since she did the React PoC with the Monaco editor.
> 
> * I feel like the component should be called "Code snippet" vs "Code editor" because it supports both editing and read-only use cases.

Updated to Code snippet

> * The `pf-c-code-editor__tab` presents an interesting question since that is actually presented by the editor itself. @nicolethoen will this work as currently structured?

I was thinking that the tab would sit on top of the container of the code snippet component and then the code editor would sit inside of the code section? or are you asking Nicole whether we will be able to get the information (e.g. HTML or JS) from the code editor?

> * When the code snippet is displayed as read-only, it will take on different styling. Not sure if that's left to the editor or we need to apply the desired CSS.

@nicolethoen do you know ?

> * I feel like the tool buttons should have tooltips. Would that be added in the core example or left to be coded in React?

We could add them here. I feel like they usually get added in React. @mcoker is there a rule here ?

gdoyle1:
> > This is more for design, but these buttons feel a bit disjointed with the different border colors (both bottom and the top/right/left), and how the bottom border changes width on hover/focus
> > Screen Shot 2020-09-08 at 5 45 19 PM
> 
> This didn't bother me @mcoker but what do you think @gdoyle1 ?

@mcarrano Doesn't bother me either - I think it's important to have some distinction there anyway! @mceledonia curious to hear your thoughts as well

@christiemolloy one thing I think we may want to add is an "empty state" for the language detector... I was thinking we could just have the code icon show with no text to the right of it, like this:
Screen Shot 2020-09-09 at 11 11 04 AM
Thoughts?

gdoyle1 (changes requested):
> > This is more for design, but these buttons feel a bit disjointed with the different border colors (both bottom and the top/right/left), and how the bottom border changes width on hover/focus
> > Screen Shot 2020-09-08 at 5 45 19 PM
> 
> This didn't bother me @mcoker but what do you think @gdoyle1 ?

@mcarrano Doesn't bother me either - I think it's important to have some distinction there anyway! @mceledonia curious to hear your thoughts as well

@christiemolloy One thing I think we may want to add is an "empty state" for the language detector... I was thinking we could just have the code icon show with no text to the right of it, like this:
Screen Shot 2020-09-09 at 11 11 04 AM
Thoughts?

Otherwise I think this looks great!!

mceledonia:
Yeah it doesn't bother me too much either. Another option would be to add some spacing between the editor and the buttons so they aren't attached. @gdoyle1 @mcoker

nicolethoen:
@mcarrano @christiemolloy 
> * I feel like the component should be called "Code snippet" vs "Code editor" because it supports both editing and read-only use cases.

I feel like Code snippet does not communicate the right concept either - a code snippet to me is usually something that is read only.
And if someone is desiring to put a 'code snippet' into their application, they may just wrap some text in `<code>` html tags. I think because of all the features - even if it's a read only variant of the editor - it's a little more intuitive that it'd be called a code editor. There may be something even better though. Naming is not my strength

> * The `pf-c-code-editor__tab` presents an interesting question since that is actually presented by the editor itself. @nicolethoen will this work as currently structured?

The tabs did not come with the code editor itself. I think it's good as is.

> * When the code snippet is displayed as read-only, it will take on different styling. Not sure if that's left to the editor or we need to apply the desired CSS.

I believe that the editor does not update the style in read only mode. We probably want to control that.

src/patternfly/components/CodeEditor/code-editor-action-group-control.hbs
mcoker: Oh so it's just a dropdown for overflow items, not necessarily an overflow menu component where you'll have items that collapse into the menu at a breakpoint?

src/patternfly/components/CodeEditor/examples/CodeEditor.md
mcoker: I was assuming this meant the overflow menu component was being used here, but it sounds like this is just a dropdown with overflow items in it?

src/patternfly/components/CodeSnippet/examples/CodeSnippet.md
```diff
@@ -0,0 +1,146 @@
...
+        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Download code"'}}
+          <i class="fas fa-download"></i>
+        {{/button}}
+        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Upload code"'}}
+          <i class="fas fa-upload"></i>
+        {{/button}}
+      {{/input-group}}
+    {{/code-snippet-controls}}
+    {{#> code-snippet-tab}}
+      <i class="fas fa-code"></i>
```
mcoker, line 27: Can you add a `__tab-icon` element for this?

src/patternfly/components/CodeSnippet/examples/CodeSnippet.md
```diff
@@ -0,0 +1,146 @@
...
+        {{/button}}
+        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Download code"'}}
+          <i class="fas fa-download"></i>
+        {{/button}}
+        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Upload code"'}}
+          <i class="fas fa-upload"></i>
+        {{/button}}
+      {{/input-group}}
+    {{/code-snippet-controls}}
+    {{#> code-snippet-tab}}
```
mcoker, line 26: I think this will need to be a flex layout as the white space between the icon and text may not always be preserved.

src/patternfly/components/CodeSnippet/code-snippet.scss
```diff
@@ -0,0 +1,69 @@
...
+
+  // control
+  --pf-c-code-snippet__header--Top: var(--pf-c-code-snippet__tab--BorderWidth);
+
+  width: 100%;
+}
+
+.pf-c-code-snippet__header {
+  position: relative;
+  top: var(--pf-c-code-snippet__header--Top);
```
mcoker, line 34: This creates a `1px` gap above the header - can you use a negative margin instead?

Though I wonder if we want to do that, or if we should just leave this element's positioning normal, then use a negative margin on the `__tab` and potentially the `__controls` element since @mceledonia mentioned maybe moving the controls. It seems like only the `__tab` is the element that should overlay on top of the editor.

src/patternfly/components/CodeSnippet/code-snippet.scss
```diff
@@ -0,0 +1,69 @@
...
+  border-color: var(--pf-c-code-snippet__code--BorderColor);
+}
+
+.pf-c-code-snippet__tab {
+  padding: var(--pf-c-code-snippet__tab--PaddingTop) var(--pf-c-code-snippet__tab--PaddingRight) var(--pf-c-code-snippet__tab--PaddingBottom) var(--pf-c-code-snippet__tab--PaddingLeft);
+  margin-left: auto;
+  color: var(--pf-c-code-snippet__tab--Color);
+  background-color: var(--pf-c-code-snippet__tab--BackgroundColor);
+  border: var(--pf-c-code-snippet__tab--BorderWidth) solid;
+  border-color: var(--pf-c-code-snippet__tab--BorderTopColor) var(--pf-c-code-snippet__tab--BorderRightColor) var(--pf-c-code-snippet__tab--BorderBottomColor) var(--pf-c-code-snippet__tab--BorderLeftColor);
```
mcoker, line 68: If you don't use a bottom border, you'll get clean edges where the `__code` and `__tab` borders overlap

This is what it looks like now

Screen Shot 2020-09-09 at 11 21 17 AM

And with no bottom border

Screen Shot 2020-09-09 at 11 22 17 AM

christiemolloy:
> @mcarrano @christiemolloy
> 
> > * I feel like the component should be called "Code snippet" vs "Code editor" because it supports both editing and read-only use cases.
> 
> I feel like Code snippet does not communicate the right concept either - a code snippet to me is usually something that is read only.
> And if someone is desiring to put a 'code snippet' into their application, they may just wrap some text in `<code>` html tags. I think because of all the features - even if it's a read only variant of the editor - it's a little more intuitive that it'd be called a code editor. There may be something even better though. Naming is not my strength
> 

I agree code snippet reminds me of a one line piece of code that wouldnt have the full capabilities of an editor. Another option could be "Code Container" since we really are just providing that and you could put any editor inside of the container right ? Otherwise I would be in favor of Code Editor, because read-only is just a state of the editor. @mcarrano

src/patternfly/components/CodeSnippet/code-snippet.scss
christiemolloy: Yeah if design thinks we should go with what michael suggested, then i could see us just adding it to the tabs. but for now i think they will both need it. i think it makes sense to add it separately to both tab and controls so we can be flexible with how we change it in the future ?

Screen Shot 2020-09-09 at 4 35 27 PM
Screen Shot 2020-09-09 at 4 35 32 PM

christiemolloy:
@nicolethoen @gdoyle1 @mcarrano @mcoker ready for a re-review

src/patternfly/components/CodeSnippet/code-snippet.scss
```diff
@@ -0,0 +1,132 @@
...
+
+.pf-c-code-snippet__header {
+  position: relative;
+  display: flex;
+  align-items: flex-end;
+}
+
+.pf-c-code-snippet__controls {
+  position: relative;
+  top: var(--pf-c-code-snippet__controls--Top);
```
mcoker, line 54: curious why you'd use `top` instead of a negative margin? This leaves space between the top of the element and whatever is before it, similar to using `translate()`. 

Screen Shot 2020-09-09 at 5 02 14 PM

src/patternfly/components/CodeSnippet/code-snippet.scss
```diff
@@ -0,0 +1,132 @@
...
+  background-color: var(--pf-c-code-snippet__main--BackgroundColor);
+  border: var(--pf-c-code-snippet__main--BorderWidth) solid;
+  border-color: var(--pf-c-code-snippet__main--BorderColor);
+}
+
+.pf-c-code-snippet__code {
+  position: relative;
+
+  &.pf-m-drag-hover {
+    &::before {
```
mcoker, line 86: can you swap the before and after so the background is the before, and the border is the after? You shouldn't need a `z-index` then, either.

src/patternfly/components/CodeSnippet/examples/CodeSnippet.md
```diff
@@ -0,0 +1,261 @@
...
+        {{/button}}
+        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Upload code"'}}
+          <i class="fas fa-upload"></i>
+        {{/button}}
+        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Save code"'}}
+          <i class="fas fa-save"></i>
+        {{/button}}
+      {{/input-group}}
+      {{#> code-snippet-controls-overflow}}
+        {{#> overflow-menu overflow-menu--id="overflow-menu-simple"}}
```
mcoker, line 218: I'm still not sure why we need the overflow menu component here - isn't this just a dropdown?

Assuming so, since it's outside of the input group, we have a double border between the input group and overflow control. Is that intended?

Screen Shot 2020-09-09 at 5 06 42 PM

Maybe instead of an input group, we just have `__controls-item` that wraps each button, then we do `__controls-item + __controls-item { margin-left: 1px; }` to fix the double border, and have a modifier of `pf-m-overflow` on the overflow item so we could position/style it differently. wdyt?

gdoyle1 (commented):
@christiemolloy this looks great!! Thank you.

Re the naming convo: @nicolethoen @christiemolloy @mcarrano I'm good with using "code editor" as the component name

gdoyle1 (approved):

mcarrano:
This is looking good @christiemolloy @mcoker . The only question I have is why we are showing the Upload button as disabled in the Read-only example. Are we saying that it doesn't make sense to upload a read-only code block? Maybe that makes sense. What do you think @gdoyle1 ?

Regarding the naming, after giving this more thought, yes, I agree that calling this component "Code editor" makes better sense. Let's change it back.

mattnolting (commented):
I expect that, at some point, the number of options and will grow and the tab content will collide. I like the idea of the `edit__controls` as a container for positioning its content, but I wonder if we should build in an overflow menu and set to to non-collapse by default. That would provide flexibility in mobile context and help future-proof. 

Screen Shot 2020-09-10 at 12 08 03 PM

gdoyle1:
> This is looking good @christiemolloy @mcoker . The only question I have is why we are showing the Upload button as disabled in the Read-only example. Are we saying that it doesn't make sense to upload a read-only code block? Maybe that makes sense. What do you think @gdoyle1 ?
> 
> Regarding the naming, after giving this more thought, yes, I agree that calling this component "Code editor" makes better sense. Let's change it back.

I think we wanted to show an example of a disabled button and when a code block is read-only, it wouldn't be edited so it made sense that the user could still download and copy, but not overwrite the code within it by uploading new code

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -0,0 +1,144 @@
...
+  // stylelint-enable selector-class-pattern
+}
+
+.pf-c-code-editor__code {
+  position: relative;
+  padding: var(--pf-c-code-editor__code--PaddingTop) var(--pf-c-code-editor__code--PaddingRight) var(--pf-c-code-editor__code--PaddingBottom) var(--pf-c-code-editor__code--PaddingLeft);
+  font-family: var(--pf-c-code-editor__code--FontFamily);
+  font-size: var(--pf-c-code-editor__code--FontSize);
+
+  pre {
```
mcoker, line 123: This should probably be a BEM element.

gdoyle1 (approved):
Looks great! Thanks @christiemolloy

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -0,0 +1,158 @@
...
+  --pf-c-code-editor__tab--BorderTopWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-code-editor__tab--BorderRightWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-code-editor__tab--BorderBottomWidth: 0;
+  --pf-c-code-editor__tab--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-code-editor__tab--BorderColor: var(--pf-global--BorderColor--100);
+
+  // tab text
+  --pf-c-code-editor__tab-icon--text--MarginLeft: var(--pf-global--spacer--sm);
+
+  width: 100%;
```
mcoker, line 46: do you need this?

src/patternfly/components/CodeEditor/code-editor.scss
mcoker: can you update this selector?

mattnolting (approved):
Very nice work!! It's perfect 🎉

mcoker (approved):
LPTM!

Issue #3459, opened 2020-09-08
Bottom drawer panel too tall on short viewports


mcoker:
The drawer panel has a flex-basis of 300px, so if the drawer is less than 300px tall, the panel height extends beyond the height of the drawer.

PR #3460, opened 2020-09-08
fix(drawer): kept drawer panel from extending beyond drawer height

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3459


mattnolting (approved):
Genius 👍

christiemolloy (approved):
LPTM!!

Issue #3461, opened 2020-09-08
Tables headers with help button not truncating


mcoker:
When a `th` has text + help button, the text modifiers are not working properly. For example, truncation does not work.

<img width="899" alt="Screen Shot 2020-09-08 at 1 48 43 PM" src="https://user-images.githubusercontent.com/35148959/92515937-065acd00-f1da-11ea-8f5f-f056621d08e5.png">

PR #3462, opened 2020-09-08
fix(table): enabled column headers with help button to truncate

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3461


jschuler (approved):
Thanks for finding and fixing this bug 👍

Issue #3465, opened 2020-09-08
Uneven horizontal padding in vertical label group


mcoker:
The right padding should match the left.

<img width="145" alt="Screen Shot 2020-09-08 at 3 47 45 PM" src="https://user-images.githubusercontent.com/35148959/92526212-b59fa000-f1ea-11ea-900c-74e8b3a0f952.png">

PR #3466, opened 2020-09-08
fix(label-group): made left/right padding match in vertical layout

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3465


christiemolloy (approved):
perfect

mattnolting (approved):
👍

PR #3469, opened 2020-09-09
fix(pficon): updated pficon data to fix svg rendering issues

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3468


mattnolting (approved):
👀  PTM

PR #3470, opened 2020-09-10
fix(button): allowed inline link text to wrap and use with span element

### LINKED ISSUE #3313: Bug: Inline Link Button Not working as expected in components

mcoker:
fixes #3313

### DIFF
--- a/src/patternfly/components/Hint/examples/Hint.md
+++ b/src/patternfly/components/Hint/examples/Hint.md
@@ -44,10 +44,7 @@ cssPrefix: pf-c-hint
 ```hbs
 {{#> hint}}
   {{#> hint-body}}
-    Welcome to the new documentation experience.
-    {{#> button button--modifier="pf-m-link pf-m-inline"}}
-      Learn more about the improved features.
-    {{/button}}
+    Welcome to the new documentation experience. <a href="#">Learn more about the improved features</a>.
   {{/hint-body}}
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Hint/examples/Hint.md
```diff
@@ -44,10 +44,7 @@ cssPrefix: pf-c-hint
 ```hbs
 {{#> hint}}
   {{#> hint-body}}
-    Welcome to the new documentation experience.
-    {{#> button button--modifier="pf-m-link pf-m-inline"}}
-      Learn more about the improved features.
-    {{/button}}
+    Welcome to the new documentation experience. <a href="#">Learn more about the improved features</a>.
```
mattnolting, line 47: 👍

mattnolting (approved):
LPTM!

christiemolloy (approved):
LPTM

Issue #3480, opened 2020-09-14
Top aligned modal bugs


mcoker:
1) Using a margin-top for the positioning is a bad idea because it will not work well with percentages, since a vertical margin is based off of the width of the containing block.
2) Max-width is breaking because it adjusts based on the top spacer value, so if the top spacer is very large, the max width is very small.
3) Max-height has the same problem as max-width.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/3481

PR #3481, opened 2020-09-14
fix(modal): refactored the top aligned variation so it works better

mcoker:
https://github.com/patternfly/patternfly/issues/3480

yay `min()` and `max()`! 

refactored this a bit to
* use `top` instead of `margin-top` since percentage based values for `margin-top` will not work as intended.
* set smart values for min/max-width so that, if the top-aligned spacer is < the default left/right/bottom gutter for a modal, the gutters will match the spacer, creating equal space around the modal. But if the spacer is > the default gutter sizes, the default gutters are applied.

### DIFF
--- a/src/patternfly/components/ModalBox/modal-box.scss
+++ b/src/patternfly/components/ModalBox/modal-box.scss
@@ -10,9 +10,9 @@
   --pf-c-modal-box--MaxHeight: calc(100% - var(--pf-global--spacer--2xl)); // MaxHeight ensures that the modal will not go off the screen and instead the body will scroll
   --pf-c-modal-box--m-align-top--spacer: var(--pf-global--spacer--sm);
   --pf-c-modal-box--m-align-top--xl--spacer: var(--pf-global--spacer--xl);
-  --pf-c-modal-box--m-align-top--MarginTop: var(--pf-c-modal-box--m-align-top--spacer);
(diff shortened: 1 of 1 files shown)


mattnolting (approved):
LGTM 🚢

src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -10,9 +10,9 @@
   --pf-c-modal-box--MaxHeight: calc(100% - var(--pf-global--spacer--2xl)); // MaxHeight ensures that the modal will not go off the screen and instead the body will scroll
   --pf-c-modal-box--m-align-top--spacer: var(--pf-global--spacer--sm);
   --pf-c-modal-box--m-align-top--xl--spacer: var(--pf-global--spacer--xl);
-  --pf-c-modal-box--m-align-top--MarginTop: var(--pf-c-modal-box--m-align-top--spacer);
```
christiemolloy, line 13: should you remove this variable at a breaking change ?

christiemolloy (changes requested):

src/patternfly/components/ModalBox/modal-box.scss
mcoker: Oh good catch, changed it back and left a note to rename it.

christiemolloy (approved):
LGTM!

PR #3482, opened 2020-09-15
feat(date-picker): added date-picker, flatpickr styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/180

to test the flatpickr styles, you can import the compiled `patternfly-date-picker.css` in the root into this POC page http://datepicker-in-progress.surge.sh/documentation/react/components/datepicker

@nicolethoen right now I'm just styling the flatpickr classes directly. We'll want to scope these styles so that we only style our date picker instance of flatpickr, in case someone uses flatpickr on their site for some other purpose. Can we add a class to the `.flatpickr-calendar` element, or wrap it in a container so I can target the styles to the class we add?

### DIFF
--- a/src/patternfly/patternfly-date-picker.scss
+++ b/src/patternfly/patternfly-date-picker.scss
@@ -4,6 +4,7 @@
     // calendar
     --pf-c-date-picker__flatpickr__calendar--Background: var(--pf-global--BackgroundColor--light-100);
     --pf-c-date-picker__flatpickr__calendar--Padding: var(--pf-global--spacer--md);
+    --pf-c-date-picker__flatpickr__calendar--open--ZIndex: var(--pf-global--ZIndex--sm);
 
     // month/year section
     --pf-c-date-picker__flatpickr__months--PaddingBottom: var(--pf-global--spacer--md);
@@ -38,6 +39,10 @@
     // flatpickr container
     background: var(--pf-c-date-picker__flatpickr__calendar--Background);
     padding: var(--pf-c-date-picker__flatpickr__calendar--Padding);
+
+    &.open {
+      z-index: var(--pf-c-date-picker__flatpickr__calendar--open--ZIndex);
+    }
   }
 
   .flatpickr-innerContainer {
--- a/src/patternfly/patternfly-date-picker.scss
+++ b/src/patternfly/patternfly-date-picker.scss
@@ -1,143 +1,145 @@
 // stylelint-disable
-.flatpickr-calendar {
-  // calendar
-  --pf-c-date-picker__flatpickr__calendar--Background: var(--pf-global--BackgroundColor--light-100);
-  --pf-c-date-picker__flatpickr__calendar--Padding: var(--pf-global--spacer--md);
+.pf-c-date-picker__calendar {
+  &.flatpickr-calendar {
+    // calendar
+    --pf-c-date-picker__flatpickr__calendar--Background: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__calendar--Padding: var(--pf-global--spacer--md);
 
-  // month/year section
-  --pf-c-date-picker__flatpickr__months--PaddingBottom: var(--pf-global--spacer--md);
-  --pf-c-date-picker__flatpickr__month--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
-  --pf-c-date-picker__flatpickr__month-select--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
-  --pf-c-date-picker__flatpickr__month-select--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
-  --pf-c-date-picker__flatpickr__month-select--year--MarginLeft: var(--pf-global--spacer--sm);
-  --pf-c-date-picker__flatpickr__month-arrow--Color: var(--pf-global--Color--dark-200);
-  --pf-c-date-picker__flatpickr__month-arrow--hover--Color: var(--pf-global--Color--dark-100);
-  --pf-c-date-picker__flatpickr__year--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    // month/year section
+    --pf-c-date-picker__flatpickr__months--PaddingBottom: var(--pf-global--spacer--md);
+    --pf-c-date-picker__flatpickr__month--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__month-select--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__month-select--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__month-select--year--MarginLeft: var(--pf-global--spacer--sm);
+    --pf-c-date-picker__flatpickr__month-arrow--Color: var(--pf-global--Color--dark-200);
+    --pf-c-date-picker__flatpickr__month-arrow--hover--Color: var(--pf-global--Color--dark-100);
+    --pf-c-date-picker__flatpickr__year--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+
+    // Weekdays section
+    --pf-c-date-picker__flatpickr__weekdays--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__weekdays--PaddingBottom: var(--pf-global--spacer--md);
+    --pf-c-date-picker__flatpickr__weekdays--MarginBottom: var(--pf-global--spacer--sm);
+    --pf-c-date-picker__flatpickr__weekdays--BorderBottom: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
+    --pf-c-date-picker__flatpickr__weekday--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+
+    // Calendar days
+    --pf-c-date-picker__flatpickr__calendar__day--Color: var(--pf-global--Color--dark-100);
+    --pf-c-date-picker__flatpickr__calendar__day--disabled--Color: var(--pf-global--disabled-color--200);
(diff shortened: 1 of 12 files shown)


mcarrano (commented):
This looks good @mcoker But I am not seeing any styling for the calendar. Should I be?

mcoker:
@mcarrano no, I created an add-on stylesheet that will need to be pulled in to our implementation of flatpickr on the react side. There is a way to take the styles I created and add them to Nicole's POC, but you have to do it manually in dev tools.

christiemolloy:
should the text truncate if its too large ?

Screen Shot 2020-09-22 at 2 04 53 PM

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
+// stylelint-disable
+.pf-c-date-picker__calendar {
+  &.flatpickr-calendar {
+    // calendar
+    --pf-c-date-picker__flatpickr__calendar--Background: var(--pf-global--BackgroundColor--light-100);
```
christiemolloy, line 5: do you need background on both of them ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__calendar--Padding: var(--pf-global--spacer--md);
```
christiemolloy, line 6: Should this be split into padding top, right, bottom, left ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__calendar--open--ZIndex: var(--pf-global--ZIndex--sm);
```
christiemolloy, line 7: do you also need to repeat the z-index ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+
+    // month/year section
+    --pf-c-date-picker__flatpickr__months--PaddingBottom: var(--pf-global--spacer--md);
+    --pf-c-date-picker__flatpickr__month--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
```
christiemolloy, line 11: why does this need a background color too ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__month--PaddingTop: 10px;
```
christiemolloy, line 12: can you ask design if theyd be ok with using 8px to fit in with the system ?

mcoker:
@christiemolloy if that's the month selector, I'm inclined to leave that as it is and allow flatpickr to manage how it presents. It seems OK with @nicolethoen's POC. We should also check other languages to make sure it manages the size as the strings in the select may be longer.

@nicolethoen wdyt, do you think there is a risk of the text in the month select overflowing the container like that in a way that flatpickr wouldn't be able to handle?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__month-select--BackgroundSize: .625em;
+    --pf-c-date-picker__flatpickr__month-select--BackgroundPosition: calc(100% - var(--pf-global--spacer--sm)) center;
+    --pf-c-date-picker__flatpickr__month-select--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__month-select--hover--BackgroundColor: var(--pf-c-date-picker__flatpickr__month-select--BackgroundColor);
+    --pf-c-date-picker__flatpickr__month-select--year--MarginLeft: var(--pf-global--spacer--sm);
+    --pf-c-date-picker__flatpickr__month-arrow--Width: 7px;
+    --pf-c-date-picker__flatpickr__month-arrow--Color: var(--pf-global--Color--dark-200);
+    --pf-c-date-picker__flatpickr__month-arrow--hover--Color: var(--pf-global--Color--dark-100);
+    --pf-c-date-picker__flatpickr__year--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
```
christiemolloy, line 25: I don't see the hover style for this and the select, is this to override the styles from flatpickr ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__year--FontWeight: var(--pf-global--FontWeight--normal);
+
+    // Weekdays section
+    --pf-c-date-picker__flatpickr__weekdays--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+    --pf-c-date-picker__flatpickr__weekdays--PaddingBottom: var(--pf-global--spacer--md);
+    --pf-c-date-picker__flatpickr__weekdays--MarginBottom: var(--pf-global--spacer--sm);
+    --pf-c-date-picker__flatpickr__weekdays--BorderBottom: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
```
christiemolloy, line 32: should you split up these vars into border width and color ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__weekday--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+
+    // Calendar days
+    --pf-c-date-picker__flatpickr__calendar__day--Color: var(--pf-global--Color--dark-100);
+    --pf-c-date-picker__flatpickr__calendar__day--disabled--Color: var(--pf-global--disabled-color--200);
+    --pf-c-date-picker__flatpickr__calendar__day--hover--BorderColor: var(--pf-global--BackgroundColor--light-300);
+    --pf-c-date-picker__flatpickr__calendar__day--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
+    --pf-c-date-picker__flatpickr__calendar__day--selected--hover--BorderColor: var(--pf-global--primary-color--100);
```
christiemolloy, line 40: why does the border need to be the same as the background color ? could you use a pseudo element or can you not change the html ?

src/patternfly/patternfly-date-picker.scss
```diff
@@ -0,0 +1,177 @@
...
+    --pf-c-date-picker__flatpickr__calendar__day--selected--hover--Background: var(--pf-global--primary-color--100);
+    --pf-c-date-picker__flatpickr__calendar__day--selected--hover--BorderColor: var(--pf-global--Color--light-100);
+    --pf-c-date-picker__flatpickr__calendar__day--today--BorderColor: var(--pf-global--primary-color--100);
```
christiemolloy, line 43: today was a little confusing what about currentday

christiemolloy (changes requested):

src/patternfly/patternfly-date-picker.scss
mcoker: Not sure what you mean by both - which other one are you referring to?

mcoker: Do you mean both in the `pf-c-date-picker` component's `__calendar` element, as well as here in these flatpickr overrides?

src/patternfly/patternfly-date-picker.scss
mcoker: Yes, the calendar has it's own `z-index` declaration with higher specificity than `.pf-c-date-picker__calendar`

src/patternfly/patternfly-date-picker.scss
mcoker: Flatpickr comes with a grey bg there 
Screen Shot 2020-09-22 at 5 52 11 PM

src/patternfly/patternfly-date-picker.scss
mcoker: This makes the month/year area align with the arrows beside it. It's not creating a visible 10px gap from a border (which wouldn't align with our system), so I don't think there is much value in using our spacer vars. This is what it looks like with 8px.

Screen Shot 2020-09-22 at 5 53 28 PM

and here it is with 10

Screen Shot 2020-09-22 at 5 53 35 PM

I think we should wait until design can look at it and then we can update the spacing if it doesn't look right.

src/patternfly/patternfly-date-picker.scss
mcoker: Yes, this is what it looks like without the overrides

Screen Shot 2020-09-22 at 6 01 12 PM

src/patternfly/patternfly-date-picker.scss
mcoker: This overrides the way flatpickr styles the selected day when you hover. Their design includes changing borders, so we're just overriding it.

src/patternfly/patternfly-date-picker.scss
mcoker: We can but the flatpickr library uses `.flatpickr-day.today`, so I'm just re-using their names. They may start using "current day" to represent something else along with "today", then we might end up with conflicts in the naming. Seems best to match what they have if we can.

mcoker:
@christiemolloy updated

nicolethoen:
> @nicolethoen wdyt, do you think there is a risk of the text in the month select overflowing the container like that in a way that flatpickr wouldn't be able to handle?
I think I agree with your instincts. I could test languages with long month names... anyone know any? lol

mcoker:
>  I could test languages with long month names... anyone know any?

If there were other month names injected, would that be something flatpickr supports and does (switches to another language for example), or would we allow users to pass other names to the select?

nicolethoen:
> > I could test languages with long month names... anyone know any?
> 
> If there were other month names injected, would that be something flatpickr supports and does (switches to another language for example), or would we allow users to pass other names to the select?

flatpickr allows for different locales and handles all the month/day/etc translation. I demonstrate that in my react example by setting the locale to France and you can see a french calendar

mcoker:
cool! I say we just leave it and if we run into issues, we can address them as they come up. I wouldn't expect the select text to overflow as I've reduced the font size of that element from what it was before.

christiemolloy (approved):
LGTM!!

PR #3483, opened 2020-09-15
feat(form-section): added form section

### LINKED ISSUE #3395: Enhancement: form section
Currently, `forms` don't include the concept of `section` s. In long forms with different sections, input groups run together visually. 

Suggestion:
- add `pf-c-form__section`
- add `pf-c-form__section-title`

Screen Shot 2020-08-12 at 5 53 56 PM

Example design: https://marvelapp.com/prototype/8g74j7h/screen/70315576

mattnolting:
closes #3395

### DIFF
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -64,6 +64,10 @@
   --pf-c-form__helper-text--m-warning--Color: var(--pf-global--warning-color--200);
   --pf-c-form__helper-text--m-error--Color: var(--pf-global--danger-color--100);
 
+  // Section
+  --pf-c-form__section--MarginBottom: var(--pf-global--spacer--xl);
+  --pf-c-form__section--Gap: var(--pf-global--gutter--md);
+
   display: grid;
   grid-gap: var(--pf-c-form--GridGap);
 
@@ -103,6 +107,15 @@
   &.pf-m-action {
     margin-top: var(--pf-c-form__group--m-action--MarginTop);
     overflow: hidden; // keeps the negative bottom margin bottom on .pf-c-form__actions from triggering overflow
+  }
+}
+
+.pf-c-form__section {
+  display: grid;
+  gap: var(--pf-c-form__section--Gap);
+
+  &:not(:last-child) {
+    margin-bottom: var(--pf-c-form__section--MarginBottom);
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Form/form.scss
```diff
@@ -106,6 +110,15 @@
   }
 }
 
+.pf-c-form__section {
+  display: grid;
+  gap: var(--pf-c-form__section--Gap);
+
+  &:not(:last-child) {
```
mcoker, line 117: It seems valid that all form control blocks could be sections, meaning you could end up with a block that looks like this, and you have too much space between the last section and the action group.

Screen Shot 2020-09-16 at 11 10 11 AM

Maybe this should be `&:not(:first-child) { margin-top... }`?

mattnolting: Updated

mcoker:
This LGTM! Assuming all of the form groups are in sections, this works great. Though currently since the section uses a margin top, you could have groups not in a section followed by a section, and get the section spacing. But you can't have a section followed by groups and get section spacing. You could if you did something like `.pf-c-form__section + .pf-c-form__group:not(.pf-m-action) { // section-margin-top }`.

WDYT? That rule has a high specificity, but we could apply it to `section + *`, then unset it on `section + section` & `section + group.pf-m-action`.

Screen Shot 2020-09-29 at 9 39 58 AM

src/patternfly/components/Form/form.scss
```diff
@@ -106,6 +110,16 @@
   }
 }
 
+.pf-c-form__section {
+  display: grid;
+  gap: var(--pf-c-form__section--Gap);
+
+  & + .pf-c-form__group:not(.pf-m-action),
+  & ~ & {
```
mcoker, line 118: is the expectation that if you have a regular form group followed by a section, that you would get the larger section spacing? if so, that isn't working currently. I think you would replace `& ~ &` with `&:not(:first-child)`?

src/patternfly/components/Form/form.scss
```diff
@@ -106,6 +110,16 @@
   }
 }
 
+.pf-c-form__section {
+  display: grid;
+  gap: var(--pf-c-form__section--Gap);
```
christiemolloy, line 115: does it need a gap and a marginTop? could you update the gap on the selector ?

christiemolloy (changes requested):

mcoker (approved):
LPTM!

christiemolloy (approved):
looks great

PR #3485, opened 2020-09-15
fix(card-and-divider-padding): added padding to divider + section

### LINKED ISSUE #2698: Update: `pf-c-card__body` padding
Screen Shot 2020-02-11 at 12 54 32 PM

`.pf-c-divider` in `.pf-c-card` doesn't work as expected. If we update `--pf-c-card__body--PaddingTop` to `0` only when it's a sibling of another `card__head`, `card__header`, `card__body` or `card__footer`, then `.pf-c-divider` will work as expected when added.

Would look something like this;
```
.pf-c-card__head,
.pf-c-card__header,
.pf-c-card__body,
.pf-c-card__footer {
  padding-top: var(--pf-c-card--first-child--PaddingTop);
  padding-right: var(--pf-c-card--child--PaddingRight);
  padding-bottom: var(--pf-c-card--child--PaddingBottom);
  padding-left: var(--pf-c-card--child--PaddingLeft);

  & + & {
    --pf-c-card--first-child--PaddingTop: 0;
  }
}
```

` --pf-c-card--first-child--PaddingTop` would be updated to `--pf-c-card--PaddingTop` during breakpoint change.

mattnolting:
closes #2698

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -15,6 +15,7 @@
   --pf-c-card--m-compact--child--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-card--m-compact--child--PaddingBottom: var(--pf-global--spacer--md);
   --pf-c-card--m-compact--child--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-card--m-compact--c-divider--child--PaddingTop: var(--pf-global--spacer--md);
   --pf-c-card--m-compact__title--not--last-child--PaddingBottom: var(--pf-global--spacer--sm);
   --pf-c-card--m-flat--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-card--m-flat--BorderColor: var(--pf-global--BorderColor--100);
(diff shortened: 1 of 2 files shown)


mcoker:
@mceledonia @mcarrano when you get a sec, would you mind looking at this? It adds a divider between card sections, and inserts the divider after the default padding between sections, then adds more padding. So it adds a divider + 16px spacing after a section.

I'm curious if that's how we want the spacing to be when there is a divider between sections, or should the divider go in between the existing space between sections with no addition to the existing spacing?

mcarrano:
Looks good to me @mcoker . But I'll defer to @mceledonia on your questions above.

mcoker:
ping @mceledonia

mattnolting:
@mceledonia ping, wdyt??

mceledonia:
This looks good to me, heads up any `compact` variants (e.g. card) should probably drop down to --md spacer.

mceledonia (approved):

src/patternfly/components/Card/card.scss
```diff
@@ -99,6 +102,15 @@
       transform: rotate(var(--pf-c-card--m-expanded__header-toggle-icon--Rotate));
     }
   }
+
+  > .pf-c-divider {
+    & + .pf-c-card__header,
```
mcoker, line 107: I don't think you need the `&`?

mcoker (approved):
LGTM! One nit about the `&` but it's fine as is.

Issue #3450, opened 2020-09-03
Add hover states for the unread and attention Notification badge


mcarrano:
The Notification badge should have hover states for the unread and attention states as see below.

![image (3)](https://user-images.githubusercontent.com/6619860/92162155-b5289300-edff-11ea-9157-550d5d9372a7.png)

This was an oversight in the original design.

cc @mceledonia

mceledonia:
Color spec for hovers:

Unread/blue: `$pf-color-blue-500` aka `$pf-global--primary-color--200`
Attention/red: `$pf-color-red-200` aka `$pf-global--danger-color--200`

PR #3488, opened 2020-09-15
fix(notification-badge): added hover/focus styles for unread/attention

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3450

@mcarrano @mceledonia this works, and the color changes applies to the notification badge in a dark area (masthead) and light area (the example page). Though I'll note that we've traditionally made things lighter on hover/focus in dark theme, and this makes them darker. Just want to make sure I got that right?

### DIFF
--- a/src/patternfly/components/NotificationBadge/notification-badge.scss
+++ b/src/patternfly/components/NotificationBadge/notification-badge.scss
@@ -35,10 +35,12 @@
   // Unread
   --pf-c-notification-badge--m-unread--Color: var(--pf-global--Color--light-100);
   --pf-c-notification-badge--m-unread--after--BackgroundColor: var(--pf-global--active-color--100);
+  --pf-c-notification-badge--m-unread--hover--after--BackgroundColor: var(--pf-global--primary-color--200);
 
   // Attention
   --pf-c-notification-badge--m-attention--Color: var(--pf-global--Color--light-100);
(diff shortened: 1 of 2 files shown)


mcarrano (commented):
@mcoker you make a really good point about the fact that with other masthead icons we go lighter, not darker on hover. I'll defer to @mceledonia on what we actually want here. The other thing I'm wondering is about the behavior of the icon in the default state. When placed in the masthead, this has a hover state which is to show the light gray background. But when the badge is selected there is no change in background but would there be for the Read and Critical badges? Kind of hard to tell without a full interactive demo.

Screen Shot 2020-09-16 at 9 58 17 AM

mcoker:
@mcarrano for all states (read, unread, attention), the hover, focus, and selected background colors are all the same. The read background updates to a light grey, the unread state updates to a darker blue, and the attention updates to a darker red.

mceledonia:
I think going darker on those is fine, it is an inconsistency but makes sense given the starting colors of unread/attention. Going brighter would reduce contrast with the white icon. In addition the other icons don't have states and we are still aligned with the normal default state lighter background on hover.

mattnolting (approved):
LGTM

christiemolloy (approved):
LPTM

mcarrano (approved):
Looks good then @mcoker .

PR #3495, opened 2020-09-21
feat(modalbox): add help button

### LINKED ISSUE #3474: Add Help button to modal header
Want to add a Help button to the header for modals per the request here: https://github.com/patternfly/patternfly-design/issues/843

@mceledonia can you add design mocks per our earlier discussions?

christiemolloy:
closes #3474 

@mcoker added a container that doesn't need `position: absolute`. 

I added this to the header which might affect people's content if they have multiple things in the header besides the close button.
```
  display: flex;
  justify-content: space-between;
```

I had to remove `flex: 0 0 auto` from the title which might also be a problem?


mcarrano (approved):
Looks good from my perspective. @mceledonia is the spacing what you expect?

mcoker:
@christiemolloy yeah this one is tricky. I wonder if the safest bet is going to be to absolutely position the button similar to the close button, though we will need to consider the DOM order of the button if we take that approach. Currently the close button comes before the modal content, and the space we add to create room for the close button is dependent upon that DOM order, and I'm not sure if it makes sense for the help button to come before the content like that. If you go that route, we may ask Jessie about the DOM order. 

Though if we can figure out a way to add this and/or the close button in a way where they are part of the layout instead of being absolutely positioned, that would be great.

This is the modal w/ description example from this PR:

Screen Shot 2020-09-22 at 1 19 11 PM

jessiehuff (commented):
So Jenn and I discussed this structure of this, and we think it'd probably be fine to allow the help button to appear first. When testing in Safari, it announces the `aria-label` of the modal first, role, number of children, then the element with focus (usually the close button but in this case it would be the help button). I have seen something different in Chrome where it reads the close button then the contents, but we usually try to base it off of Safari since VO is most frequently paired with Safari. It's always great to keep consistency so if it was an easy fix to have the close button first and then help button, that would work. But if it requires massive changes, I think it's ok to break the consistency and just keep the normal DOM order. It's more important that users are aware of all elements than keeping the consistency of focus.
Also, I was going to approve, but I think your last update altered the example. You might already be aware of it, but figured I'd point it out just in case. 🙂 
Screen Shot 2020-09-30 at 5 32 39 PM

src/patternfly/components/ModalBox/examples/ModalBox.md
```diff
@@ -179,11 +205,14 @@ A modal box is a generic rectangular container that can be used to build modals.
 | `.pf-c-modal-box` | `<div>` | Initiates a modal box. **Required** |
 | `.pf-c-button.pf-m-plain` | `<button>` | Initiates a modal box close button. |
 | `.pf-c-modal-box__header` | `<header>` | Initiates a modal box header. **Required** if using a `.pf-c-modal-box__title`. |
+| `.pf-c-modal-box__header-main` | `<div>` | Initiates a modal box header main container. **Required** when `pf-c-modal-box__header-help` is used. | 
 | `.pf-c-modal-box__title` | `<h1>`,`<h2>`,`<h3>`,`<h4>`,`<h5>`,`<h6>`, `<div>` | Initiates a modal box title. **Required** if using a modal description. |
 | `.pf-c-modal-box__description` | `<div>` | Initiates a modal box description. |
 | `.pf-c-modal-box__body` | `<div>` | Initiates a modal box body. |
 | `.pf-c-modal-box__footer` | `<footer>` | Initiates a modal box footer. |
+| `.pf-c-modal-box__header-help` | `<div>` | Initiates the help button container in the modal box header actions. |
```
mcoker, line 213: can you move this up under `__header-main`?

src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -41,10 +41,17 @@
...
 
   // Close
-  --pf-c-modal-box--c-button--Top: calc(var(--pf-global--spacer--lg) - var(--pf-global--spacer--form-element) + #{pf-size-prem(1px)}); // align top of button with top of text
+  --pf-c-modal-box--c-button--Top: calc(var(--pf-global--spacer--lg) + var(--pf-global--spacer--xs) - var(--pf-global--spacer--form-element) + #{pf-size-prem(1px)}); // align top of button with top of text
   --pf-c-modal-box--c-button--Right: var(--pf-global--spacer--md);
   --pf-c-modal-box--c-button--sibling--MarginRight: var(--pf-global--spacer--xl);
 
+  // Actions container
+  --pf-c-modal-box__actions--sibling--MarginRight: var(--pf-global--spacer--4xl);
+  --pf-c-modal-box__actions--m-2-actions--sibling--MarginRight: var(--pf-global--spacer--xl);
```
mcoker, line 50: do you need this `__actions` code anymore?

src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -94,13 +101,31 @@
...
 }
 
 .pf-c-modal-box__header {
+  display: flex;
+  flex-direction: column;
   padding-top: var(--pf-c-modal-box__header--PaddingTop);
   padding-right: var(--pf-c-modal-box__header--PaddingRight);
   padding-left: var(--pf-c-modal-box__header--PaddingLeft);
 
+  &.pf-m-row {
```
mcoker, line 125: I wonder if this should be an opt-in to restructure the `__header` element.

Something like

```
.pf-c-modal-box__header.pf-m-help {
 display: flex;
 flex-direction: row;
}
```

mcoker (approved):
Fantastic!!!

Issue #3451, opened 2020-09-03
Long alert descriptions exceed component's width


bdone:
an alert toast description with long strings looks like...

![image](https://user-images.githubusercontent.com/302272/92175758-3174a380-ee0b-11ea-9658-13b746657419.png)

this component (and others?) should default to something that doesn't exceed its own width. perhaps [like this ](https://codepen.io/bdone/pen/abNVdXP) ...

![image](https://user-images.githubusercontent.com/302272/92176003-a1832980-ee0b-11ea-85c0-6f2c9cfbe2ff.png)

bdone:
perhaps there is a utility class that helps here? there's some mention of them in https://github.com/patternfly/patternfly/issues/1534.

PR #3505, opened 2020-09-23
fix(alert): allowed long strings in description to wrap

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3451


mattnolting (approved):
Perfect!

PR #3508, opened 2020-09-23
fix(inputgroup): fix double line in invalid

### LINKED ISSUE #2732: Input group - form control invalid state has double border
This block is causing a double border between the invalid form control and the element that follows it:

https://github.com/patternfly/patternfly-next/blob/9d75002b14731c7d13d2ff531f0368981228770b/src/patternfly/components/InputGroup/input-group.scss#L41-L47

Screen Shot 2020-02-20 at 11 20 28 AM

christiemolloy:
closes #2732 

@mcoker can this be positioned as a bug so that I can remove the block and variable ?

### DIFF
--- a/src/patternfly/components/InputGroup/input-group.scss
+++ b/src/patternfly/components/InputGroup/input-group.scss
@@ -17,7 +17,7 @@
 
   // Form control
   --pf-c-input-group--c-form-control--invalid--ZIndex: var(--pf-global--ZIndex--xs);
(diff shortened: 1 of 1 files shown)


mcoker:
> can this be positioned as a bug so that I can remove the block and variable ?

Unless there is a benefit, I would just leave it to be on the safe side.

mcoker (approved):
Perfect!

mattnolting (approved):
LPTM!

PR #3509, opened 2020-09-23
feat(pagination): add sticky position to bottom

### LINKED ISSUE #3225: Create pagniation footer for use in Card and other view
Want to create a pagination footer to add to the bottom of a card or similar view that currently does not have a built in footer. See design below.

cc @mcoker

christiemolloy:
closes #3225 

Added sticky bottom pagination to the card view demo.


mcarrano (approved):
Looks great @christiemolloy

mattnolting (approved):
LPTM 🎉

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -112,6 +113,11 @@
       box-shadow: none;
     }
 
+    &.pf-m-sticky {
```
mcoker, line 116: Could we also make it top sticky in case someone wants to do that?

Also I realized I left `bottom: 0;` on the bottom pagination on the desktop view when the pagination isn't sticky. We should probably add `bottom: auto` there, then reset `bottom: 0` on `.pf-m-bottom.pf-m-sticky`.

christiemolloy:
Updated @mcoker @mattnolting @mcarrano added a top sticky variation to Pagination. Do you mind taking a look again. An example of the sticky top is in the Card View demo.

mcarrano:
I guess this is OK @christiemolloy . But it does raise a design question. What you've basically done is to make the toolbar above the card grid sticky. Is this what we expect? I would think the same would apply to a data list or table in a view like this. @mceledonia @gdoyle1 I'd be interested to hear your thoughts. Is this the default behavior we expect for full page views like this?

src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -23,14 +23,31 @@ cssPrefix: pf-c-pagination
 {{/pagination}}
 ```
 
-### Bottom
+### Top sticky
+```hbs isFullscreen
+{{#> pagination pagination--modifier="pf-m-sticky"}}
```
mcoker, line 28: can you make this `sticky-bottom` to match the page section variations?

src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -78,6 +95,7 @@ Note: `<button>` or `<a>` elements can be used in `.pf-c-pagination__nav-page-se
 | `.pf-m-bottom` | `.pf-c-pagination` | Modifies for bottom pagination component styles. |
 | `.pf-m-compact` | `.pf-c-pagination` | Modifies for compact pagination component styles. |
 | `.pf-m-static` | `.pf-c-pagination.pf-m-bottom` | Modifies bottom pagination to not be positioned sticky on mobile. |
+| `.pf-m-sticky` | `.pf-c-pagination.pf-m-bottom` | Modifies bottom pagination to be positioned sticky at all viewports. |
```
mcoker, line 98: how about something like this?

```suggestion
| `.pf-m-sticky` | `.pf-c-pagination` | Modifies the pagination to be sticky to its container. It will be sticky to the top of the container by default, and sticky to the bottom of the container when applied to `.pf-c-pagination.pf-m-bottom`. |
```

mcoker:
@christiemolloy found a few bugs and added a fix in this PR https://github.com/christiemolloy/patternfly-1/pull/8

Also, children of `__main` should be some sort of section. Currently the pagination in the card view demo is a child of `__main` - it should be added to a `__main-section.pf-m-no-padding`.

Though to make that sticky, we'll need to merge https://github.com/patternfly/patternfly/pull/3516

And I'm not sure that we want to show the top pagination as sticky in the card view demo? Or maybe that's what @mcarrano is asking above? Maybe a better place for it would be a table and/or data-list demo. That works better because then the pagination doesn't need to be in a section and the markup can be `__main-section > pagination.pf-m-sticky + table + pagination.pf-m-bottom.pf-m-sticky`. We could add a new example showing sticky top and/or bottom pagination, assuming that's a valid demo from design's perspective.

gdoyle1:
> I guess this is OK @christiemolloy . But it does raise a design question. What you've basically done is to make the toolbar above the card grid sticky. Is this what we expect? I would think the same would apply to a data list or table in a view like this. @mceledonia @gdoyle1 I'd be interested to hear your thoughts. Is this the default behavior we expect for full page views like this?

@mcarrano I'd agree that if you have the bottom pagination sticky, you wouldn't need the top to be sticky as well. I could see having the top toolbar sticky if there are actions in there and _no_ pagination. In this case though, agreed only bottom needs to be sticky

christiemolloy:
@mcarrano @mcoker @gina updated the Table basic demo and the Card view demo, let me know what you think about the top pagination in the table demo

mcarrano:
@christiemolloy I think these are fine as demos. It shows the flexibility we have. @gdoyle1 I'm wondering if this topic should be addressed as part of our design guidelines for pagination and/or toolbars. Now that we have this capability, I expect questions about when we should make these elements sticky or not. Thoughts?

src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -23,14 +23,31 @@ cssPrefix: pf-c-pagination
...
+### Top sticky
+```hbs isFullscreen
+{{#> pagination pagination--modifier="pf-m-sticky"}}
+  {{> pagination-total-items-content}}
+  {{> pagination-options-menu options-menu id="pagination-options-menu-top-example" options-menu--IsText="true"}}
+  {{> pagination-nav-content}}
+{{/pagination}}
+```
+
+### Bottom default
```
mcoker, line 35: I would probably just call this "Bottom" like it was before, and it will match the "Top" example.

gdoyle1:
> @christiemolloy I think these are fine as demos. It shows the flexibility we have. @gdoyle1 I'm wondering if this topic should be addressed as part of our design guidelines for pagination and/or toolbars. Now that we have this capability, I expect questions about when we should make these elements sticky or not. Thoughts?

Agreed we should address that in design guidelines! @mmenestr any way we could create an issue in pf-org about this?

mmenestr:
@gdoyle1 @mcarrano just to make sure i'm understanding, you want to add something to the pagination and toolbars guidelines to mention how you can make them sticky?

gdoyle1:
> @gdoyle1 @mcarrano just to make sure i'm understanding, you want to add something to the pagination and toolbars guidelines to mention how you can make them sticky?

@mmenestr Yes around toolbars as well as around full page demos! So I think if we have card views/table/list views we could mention it there as well

mcoker:
@gdoyle1 @mmenestr we ended up taking the sticky pagination demo out of the table demos since it had both top and bottom pagination sticky, and moved top and bottom sticky pagination to examples on the pagination component page - https://pf4.patternfly.org/components/pagination

Let us know if you want to update any of that.

PR #3517, opened 2020-09-24
feat(alert): add inline variation demo

### LINKED ISSUE #3282: Inline alert demo
Create a demo that shows how to display errors and alerts within a form. Should address the following.

    Associated with the form label versus the control (displaying above the inputs/controls)
    Associated with a group of controls (eg, a group of inline checkbox controls)
    An inline alert message - at the top of the form and/or the top of the page (underneath the header and description)

christiemolloy:
closes #3282

### DIFF
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -8,7 +8,7 @@
   --pf-c-form--m-horizontal__group-control--md--GridColumnWidth: 1fr;
 
   // limit width
-  --pf-c-form--m-limit-width--md--Width: #{pf-size-prem(500px)};
+  --pf-c-form--m-limit-width--MaxWidth: #{pf-size-prem(500px)};
 
   // Group label
   --pf-c-form--m-horizontal__group-label--md--PaddingTop: var(--pf-global--spacer--sm);
@@ -41,6 +41,7 @@
 
   // Group control
   --pf-c-form__group-control--m-inline--child--MarginRight: var(--pf-global--spacer--lg);
+  --pf-c-form__group-control__helper-text--MarginBottom: var(--pf-global--spacer--xs);
 
   // Actions
   --pf-c-form__actions--child--MarginTop: var(--pf-global--spacer--sm);
@@ -108,11 +109,7 @@
   }
 
   &.pf-m-limit-width {
-    width: 100%;
-
-    @media (min-width: $pf-global--breakpoint--md) {
-      width: var(--pf-c-form--m-limit-width--md--Width);
-    }
+    max-width: var(--pf-c-form--m-limit-width--MaxWidth);
   }
 }
 
@@ -125,6 +122,10 @@
 
 .pf-c-form__group-label {
   padding-bottom: var(--pf-c-form__group-label--PaddingBottom);
+
+  .pf-c-form__label + .pf-c-form__helper-text {
+    --pf-c-form__helper-text--MarginTop: 0;
+  }
 }
 
 .pf-c-form__label {
@@ -182,6 +183,12 @@
     > * {
       margin-right: var(--pf-c-form__group-control--m-inline--child--MarginRight);
     }
+  }
+
+  .pf-c-form__helper-text:first-child {
+    --pf-c-form__helper-text--MarginTop: 0;
+
+    margin-bottom: var(--pf-c-form__group-control__helper-text--MarginBottom);
   }
 }
 
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -7,6 +7,9 @@
   --pf-c-form--m-horizontal__group-label--md--GridColumnGap: var(--pf-global--spacer--md);
   --pf-c-form--m-horizontal__group-control--md--GridColumnWidth: 1fr;
 
+  // limit width
+  --pf-c-form--m-limit-width--md--Width: #{pf-size-prem(500px)};
+
   // Group label
   --pf-c-form--m-horizontal__group-label--md--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-form__group-label--PaddingBottom: var(--pf-global--spacer--sm);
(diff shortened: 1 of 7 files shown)


mcarrano (changes requested):
So I think that what is missing here is showing an alert below the label for the checkbox group. Can you confirm @mmenestr ?

mmenestr:
@christiemolloy thank you for doing this!!
@mcarrano is right, need to add an error state for the "Your experience" and "How can we contact you fields" (new pattern)!
I realize now that I didn't put any spacers or anything, so I just updated the mocks with that info!

Also small thing that's also my fault from the mocks that (which i've now updated), could you just make the spacing between the inline alert and the start of the horizontal form 24px instead of 48px? 

Otherwise looks perfect!

mcoker:
If we want to enforce this specific spacing between the inline alert and the form, I don't think we want to use a generic layout. I'm wondering if this inline alert should be an element of the form component. Then we can add the appropriate amount of spacing per form layout modifier.

And in the stacked demo, we're using the grid layout. Is that to impose a width? Seems like we might want to have a modifier to impose that width instead if that's what we're going for.

mattnolting (changes requested):
Looks like you can get rid of the layouts entirely and add `helper text` to last group..

christiemolloy:
@mcarrano @mmenestr @mcoker @mattnolting updated do you mind re-reviewing ? 

Screen Shot 2020-09-30 at 1 56 51 PM
Screen Shot 2020-09-30 at 1 56 55 PM

I originally omitted the helper text at the bottom because the spacing is off. Should I include it inside of the `pf-c-form-group-label` so that `pf-c-form-label` and `pf-c-form-helper-text` are siblings? and then write CSS for when they are next to each other?

src/patternfly/components/Form/form.scss
```diff
@@ -7,6 +7,9 @@
   --pf-c-form--m-horizontal__group-label--md--GridColumnGap: var(--pf-global--spacer--md);
   --pf-c-form--m-horizontal__group-control--md--GridColumnWidth: 1fr;
 
+  // limit width
+  --pf-c-form--m-limit-width--md--Width: #{pf-size-prem(500px)};
```
mattnolting, line 11: Should this be `--pf-c-form--MaxWidth: auto`? Then `--pf-c-form--MaxWidth` can be updated inline on the parent `form`.

src/patternfly/demos/Alert/examples/horizontal-form.hbs
```diff
@@ -0,0 +1,106 @@
...
+              Submit form
+            {{/button}}
+            {{#> button button--modifier="pf-m-secondary" button--IsReset="true"}}
+              Cancel
+            {{/button}}
+          {{/form-actions}}
+        {{/form-group-control}}
+      {{/form-group}}
+  {{/form}}
+{{/page-main-section}}
```
mattnolting, line 106:
```suggestion
{{/page-main-section}}

```

mmenestr:
Just looking at the preview site, and not the screenshot, there's a little too much spacing in the stacked form demo is. Should be no space between the label and error, and only 4px between the error and the checkboxes, instead of the 8px you have right now!

src/patternfly/demos/Alert/examples/stacked-form.hbs
```diff
@@ -0,0 +1,114 @@
...
+          {{/button}}
+          {{#> button button--modifier="pf-m-secondary" button--IsReset="true"}}
+            Cancel
+          {{/button}}
+        {{/form-actions}}
+      {{/form-group-control}}
+    {{/form-group}}
+  {{/form}}
+
+{{/page-main-section}}
```
mattnolting, line 114:
```suggestion
{{/page-main-section}}

```

src/patternfly/demos/Alert/examples/page-default.hbs
```diff
@@ -0,0 +1,87 @@
...
+            {{#> nav-link nav-link--href="#"}}
+              Tertiary nav item 5
+            {{/nav-link}}
+          {{/nav-item}}
+        {{/nav-list}}
+      {{/nav}}
+    {{/page-main-nav}}
+    {{> @partial-block}}
+  {{/page-main}}
+{{/page}}
```
mattnolting, line 87:
```suggestion
{{/page}}

```

src/patternfly/components/Form/form.scss
```diff
@@ -97,6 +106,14 @@
       }
     }
   }
+
+  &.pf-m-limit-width {
+    width: 100%;
```
mcoker, line 111: could this just be `max-width: 500px` with no need for a media query?

src/patternfly/components/Form/form.scss
```diff
@@ -97,6 +106,14 @@
...
+
+    @media (min-width: $pf-global--breakpoint--md) {
+      width: var(--pf-c-form--m-limit-width--md--Width);
+    }
+  }
```
mattnolting, line 116: If using an inline variable update, this block can go away.

src/patternfly/components/Form/form.scss
```diff
@@ -76,6 +79,12 @@
       }
     }
 
+    .pf-c-form__group {
+      &.pf-m-width-100 {
+        display: block;
+      }
+    }
```
mattnolting, line 86: Is this needed?

christiemolloy: its to undo the display: grid thats set, should i do something else ?

src/patternfly/components/Form/form.scss
```diff
@@ -108,6 +122,10 @@
 
 .pf-c-form__group-label {
   padding-bottom: var(--pf-c-form__group-label--PaddingBottom);
+
+  .pf-c-form__label + .pf-c-form__helper-text {
```
mcoker, line 126: do you need this or could you just set `--pf-c-form__helper-text--MarginTop: 0;` on `.pf-c-form__group-label`?

mcoker (approved):
very nice!!

mcarrano (approved):
Looks great @christiemolloy !

Issue #3513, opened 2020-09-24
Introduce extra-small empty state


mcarrano:
Create a new extra-small empty state, primarily for use in cards. Designs here: https://marvelapp.com/prototype/7c5cdhd/screen/72928590

PR #3519, opened 2020-09-25
feat(empty-state): added xs variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3513

### DIFF
--- a/src/patternfly/components/EmptyState/empty-state.scss
+++ b/src/patternfly/components/EmptyState/empty-state.scss
@@ -24,6 +24,10 @@
   --pf-c-empty-state--m-xs--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-empty-state--m-xs--PaddingBottom: var(--pf-global--spacer--md);
   --pf-c-empty-state--m-xs--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-empty-state--m-xs__icon--MarginBottom: var(--pf-global--spacer--md);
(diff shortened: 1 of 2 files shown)


mcarrano (changes requested):
The extra small empty space should not include an icon. See the designs here https://marvelapp.com/prototype/7c5cdhd/screen/72928590. Please reach out to @mmenestr if you have questions.

mcoker:
@mcarrano ah ok, I didn't see that as a requirement. Technically you can still use one, but I removed it from the example. Would we want to decrease the icon size if a user did include an icon?

mcarrano:
I think so. @mmenestr ?

mmenestr:
I agree that we should still give people the option to add an icon if they wanted to, although in most cases it may not fit. 
I just did a test of 24px vs 54px icon, and I think I would keep the icon at 54px since there is no in between size (based on PF icon sizing), and the 24px looks too small IMO.  One other thing is the spacing between the body and the buttons should be 16px not 32px! Otherwise, looks great!

image

mcoker:
Thanks @mmenestr forgot to update that spacing! Here is what it looks like with an icon and primary button, if someone were to put those in. 16px between all sections, except I'm retaining the 8px spacer between the primary button and secondary list, which is used in the other variations.

mattnolting (approved):
LGTM

mmenestr (approved):

christiemolloy (approved):
LGTM!!

mcarrano (approved):
Looks great. Thanks @christiemolloy !

Issue #3520, opened 2020-09-25
Date picker follow ups


mcoker:
Issue for core changes needed from reviews in the react PR - https://github.com/patternfly/patternfly-react/pull/4888

* trailing characters in the month select are getting cut off.

PR #3521, opened 2020-09-25
fix(date-picker): kept text in month select from getting cut off

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3520


christiemolloy (approved):
confirmed - Saturday is visible now

PR #3542, opened 2020-10-01
chore(build): don't upload artifacts

redallen:
Speeds up builds ~1m20s. We upload them to surge and no longer need to upload them in CI since they're rather large now.


mcoker (approved):
👍

Issue #3344, opened 2020-07-29
Stacked toolbar spacing bug for PF React toolbar component


fionayhlin:
Stacked toolbar example on the PF website [here](https://www.patternfly.org/v4/documentation/react/components/toolbar#stacked-example) has a spacing issue. 

<img width="1038" alt="Screen Shot 2020-07-29 at 1 40 31 PM" src="https://user-images.githubusercontent.com/49918197/88833936-13c77500-d1a1-11ea-95f7-c5ad67700022.png">

gdoyle1:
fyi @mattnolting this is what we were discussing in slack!

unlikelyzero:
Filed as a blocking issue against awx https://github.com/patternfly/patternfly/issues/3344

unlikelyzero:
ping

PR #3543, opened 2020-10-01
Feat add gap to toolbar

### LINKED ISSUE #3348: Add GAP to toolbar!

mattnolting:
closes #3348 #3344 

@mcarrano @mceledonia I set `row-gap` for direct children of `content-sections` to `spacer--md`. For items within groups, I set the `row-gap` to `spacer--xs`. Wdyt?

Also, I set toggle groups to **nowrap**, as they are a special use case and imo should never wrap as their behavior has a specific functionality.

Screen Shot 2020-10-01 at 3 10 43 PM

Screen Shot 2020-10-01 at 3 11 44 PM

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -13,6 +13,7 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   // Content
   --pf-c-toolbar__content--PaddingRight: var(--pf-global--spacer--md); // remove at breaking change
   --pf-c-toolbar__content--PaddingLeft: var(--pf-global--spacer--md); // remove at breaking change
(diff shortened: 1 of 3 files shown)


mcarrano (approved):
Looks good to me.

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -130,8 +134,10 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   --pf-c-toolbar--spacer: var(--pf-c-toolbar__group--spacer);
 
   display: flex;
+  flex-wrap: wrap;
```
mcoker, line 137: This seems like a pretty big change, we don't think this will be breaking? row-gap or not, seems like it will change the way items lay out in the toolbar. And some people won't have `row-gap` yet either.

mattnolting: You're right, we already have `pf-m-wrap`.

PR #3549, opened 2020-10-02
feat(gallery): add width variable

### LINKED ISSUE #3547: Enhancement: Gallery layout grid template minimum width
**Why?**
250px width is a generic value that should be updatable to accomodate different layout needs.

**Example**
Screen Shot 2020-10-02 at 8 29 15 AM

add `--pf-l-gallery--GridTemplateColumns--min: 250px` to define minimum width in the css, then React can allow for this variable to be updated inline

mattnolting:
closes #3547 

This PR assigns a variable to the gallery layout's `min` value in the `grid-template-columns` declaration. It also allows for this css variable to be updated responsively.

### DIFF
--- a/src/patternfly/layouts/Gallery/gallery.scss
+++ b/src/patternfly/layouts/Gallery/gallery.scss
@@ -1,28 +1,28 @@
 .pf-l-gallery {
   --pf-l-gallery--m-gutter--GridGap: var(--pf-global--gutter);
   --pf-l-gallery--GridTemplateColumns--min: 250px;
-  --pf-l-gallery--GridTemplateColumns--min--core: var(--pf-l-gallery--GridTemplateColumns--min);
-  --pf-l-gallery--GridTemplateColumns: repeat(auto-fill, minmax(var(--pf-l-gallery--GridTemplateColumns--min--core), 1fr)); // The 250px value is here just so that the grid has a width to use in the minmax() for laying out the gallery columns.
+  --pf-l-gallery--GridTemplateColumns--minmax--min: var(--pf-l-gallery--GridTemplateColumns--min);
+  --pf-l-gallery--GridTemplateColumns: repeat(auto-fill, minmax(var(--pf-l-gallery--GridTemplateColumns--minmax--min), 1fr)); // The 250px value is here just so that the grid has a width to use in the minmax() for laying out the gallery columns.
   --pf-l-gallery--GridTemplateRows: auto;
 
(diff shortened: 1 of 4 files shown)


src/patternfly/layouts/Gallery/gallery.scss
```diff
@@ -1,8 +1,30 @@
 .pf-l-gallery {
   --pf-l-gallery--m-gutter--GridGap: var(--pf-global--gutter);
-  --pf-l-gallery--GridTemplateColumns: repeat(auto-fill, minmax(250px, 1fr)); // The 250px value is here just so that the grid has a width to use in the minmax() for laying out the gallery columns.
+  --pf-l-gallery--GridTemplateColumns--min: 250px;
+  --pf-l-gallery--GridTemplateColumns--min--core: var(--pf-l-gallery--GridTemplateColumns--min);
```
mcoker, line 4: Love this change. Though what does `--core` mean here? Maybe we should structure it like the modal's top aligned spacer, and name the overridable thing something user friendly, and that maps to the var name that we build off of the selector/property name. So here, keep `--pf-l-gallery--GridTemplateColumns--min`, and change `--pf-l-gallery--GridTemplateColumns--min--core` to `--pf-l-gallery--GridTemplateColumns--minmax--min`?

mattnolting: Love it!

src/patternfly/layouts/Gallery/gallery.scss
```diff
@@ -1,8 +1,30 @@
...
+  --pf-l-gallery--GridTemplateColumns: repeat(auto-fill, minmax(var(--pf-l-gallery--GridTemplateColumns--min--core), 1fr)); // The 250px value is here just so that the grid has a width to use in the minmax() for laying out the gallery columns.
   --pf-l-gallery--GridTemplateRows: auto;
 
+  @media screen and (min-width: $pf-global--breakpoint--sm) {
```
mcoker, line 8: WDYT creating an SCSS function for this MQ block since it's a repeatable thing and it would be easy to make a mistake either writing/reviewing it?

mcoker (approved):
Very nice!!

src/patternfly/layouts/Gallery/gallery.scss
mattnolting: @mcoker check it out!

christiemolloy (approved):
Looks great!!

PR #3552, opened 2020-10-04
feat(add-gap-to-flex): added gap to flex

### LINKED ISSUE #3489: Add GAP to flex layout!
Flex gap is finally here, we need to add it to the flex layout. Should have optional row-gap modifiers.

mattnolting:
closes #3489

### DIFF
--- a/src/patternfly/layouts/Flex/examples/Flex.md
+++ b/src/patternfly/layouts/Flex/examples/Flex.md
@@ -769,6 +769,98 @@ Ordering - Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.
 {{/l-flex}}
 ```
 
+### Row gap
+```hbs
+{{#> l-flex l-flex--modifier="pf-m-row-gap" l-flex--attribute='style="--pf-l-flex--RowGap: var(--pf-global--spacer--lg);"'}}
+  {{#> l-flex l-flex--modifier="pf-m-row-gap" l-flex--attribute='style="--pf-l-flex--RowGap: var(--pf-global--spacer--sm);"'}}
+    {{#> l-flex-item}}
+      Set 1, Item A
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item B
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item C
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item D
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item E
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item F
+    {{/l-flex-item}}
+  {{/l-flex}}
+  {{#> l-flex l-flex--modifier="pf-m-row-gap" l-flex--attribute='style="--pf-l-flex--RowGap: var(--pf-global--spacer--lg); --pf-l-flex--RowGap-on-xl: var(--pf-global--spacer--2xl);"'}}
+    {{#> l-flex-item}}
+      Set 2, Item A
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 2, Item B
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 2, Item C
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 2, Item D
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 2, Item E
+    {{/l-flex-item}}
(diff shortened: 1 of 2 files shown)


mcoker:
@mattnolting this is cool! Can you add an example and the modifier to the docs?

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -21,6 +23,7 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
   display: var(--pf-l-flex--Display);
   flex-wrap: var(--pf-l-flex--FlexWrap);
   align-items: var(--pf-l-flex--AlignItems);
+  row-gap: var(--pf-l-flex--RowGap);
```
mcoker, line 26: I think this should probably be an opt in as it will add space between rows where there is none currently. I could see it being beneficial for sure, but likely to cause enough problems we should wait until a breaking change to make it the default. WDYT?

mattnolting:
@mcoker updated

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -769,6 +769,98 @@ Ordering - Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.
 {{/l-flex}}
 ```
 
+### Row gap
```
mcoker, line 772: Could we add an example that just shows `pf-m-row-gap` on its own? That could be a simple example. Then this example could show custom row gaps.

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -769,6 +769,98 @@ Ordering - Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.
...
+```hbs
+{{#> l-flex l-flex--modifier="pf-m-row-gap" l-flex--attribute='style="--pf-l-flex--RowGap: var(--pf-global--spacer--lg);"'}}
+  {{#> l-flex l-flex--modifier="pf-m-row-gap" l-flex--attribute='style="--pf-l-flex--RowGap: var(--pf-global--spacer--sm);"'}}
```
mcoker, line 775: since `--sm` is the default gap, I think it's confusing to see it set here

```suggestion
  {{#> l-flex l-flex--modifier="pf-m-row-gap"}}
```

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -769,6 +769,98 @@ Ordering - Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.
...
+      Set 1, Item D
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item E
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Set 1, Item F
+    {{/l-flex-item}}
+  {{/l-flex}}
+  {{#> l-flex l-flex--modifier="pf-m-row-gap" l-flex--attribute='style="--pf-l-flex--RowGap: var(--pf-global--spacer--lg); --pf-l-flex--RowGap-on-xl: var(--pf-global--spacer--2xl);"'}}
```
mcoker, line 795: this is responsive - it reads like the example under it has the responsive examples.

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -820,7 +913,8 @@ Ordering - Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.
 | `.pf-m-align-content-space-around{-on-[breakpoint]}` | `.pf-l-flex` | Modifies the flex layout align-content property to space-around. |
 | `.pf-m-align-left{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Resets the flex layout element margin-left property to 0. |
 | `.pf-m-align-right{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the flex layout element margin-left property to auto. |
-| `--pf-l-flex--item--Order{-on-[breakpoint]}: {order}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the flex layout element order property. |
+| `--pf-l-flex--RowGap{-on-[breakpoint]}: {height}` | `.pf-l-flex` | Modifies the flex layout `row-gap` value. |
```
mcoker, line 916: What about "size" here for a size unit?

```suggestion
| `--pf-l-flex--RowGap{-on-[breakpoint]}: {size}` | `.pf-l-flex` | Modifies the flex layout `row-gap` value. |
```

mcoker (approved):
lgtm!

Issue #3562, opened 2020-10-05
Modal alert updates


mcoker:
* "error" variant should be renamed to "danger"
* There is less space between the title icon and text in the react component due to stripping of white space. Title should be a flex layout when there is an icon present.

PR #3563, opened 2020-10-05
fix(modal-box): added icon variant, renamed error to danger

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3562


KKoukiou (approved):
Both reported issues where addressed! thanks for the quick reaction!

redallen (approved):

PR #3567, opened 2020-10-07
fix(progress): enabled static width for measure value

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3529

@kmcfaul adding you to see if the variation works with what you'd expect in the react component. Adds `pf-m-fixed-width` to the `__measure` element that fits any `0-100%` value. And that width is overridable if someone is using a non-percentage value for the measure and wants a custom width by setting `pf-m-fixed-width` and defining `--pf-c-progress__measure--m-fixed-width--Width: 25%` (where 25% is any size unit).

### DIFF
--- a/src/patternfly/components/Progress/examples/Progress.md
+++ b/src/patternfly/components/Progress/examples/Progress.md
@@ -152,7 +152,7 @@ cssPrefix: pf-c-progress
 {{#> progress
   progress__value="100"
   progress--modifier="pf-m-outside pf-m-lg"
-  progress__id="progress-outside-fixed-width-2-example"
+  progress__id="progress-outside-fixed-width-3-example"
   progress-measure--modifier="pf-m-fixed-width"
 }}
 {{/progress}}
--- a/src/patternfly/components/Progress/examples/Progress.md
+++ b/src/patternfly/components/Progress/examples/Progress.md
@@ -131,6 +131,33 @@ cssPrefix: pf-c-progress
 {{/progress}}
 ```
 
+### Outside fixed width measure
+```hbs
+{{#> progress
+  progress__value="1"
+  progress--modifier="pf-m-outside pf-m-lg"
+  progress__id="progress-outside-fixed-width-example"
+  progress-measure--modifier="pf-m-fixed-width"
+}}
+{{/progress}}
+<br>
+{{#> progress
+  progress__value="50"
+  progress--modifier="pf-m-outside pf-m-lg"
+  progress__id="progress-outside-fixed-width-2-example"
+  progress-measure--modifier="pf-m-fixed-width"
+}}
+{{/progress}}
+<br>
+{{#> progress
+  progress__value="100"
+  progress--modifier="pf-m-outside pf-m-lg"
+  progress__id="progress-outside-fixed-width-2-example"
+  progress-measure--modifier="pf-m-fixed-width"
+}}
+{{/progress}}
+```
+
 ### On single line
 ```hbs
 {{#> progress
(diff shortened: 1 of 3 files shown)


christiemolloy (changes requested):
I wonder if there could be a more fluid way of handling this sizing rather than having to add a modifier. Because the percentages are so dynamic a lot of users may not realize they need to add the variant until their percentages are large.

There also seems to be a lot of space now between the numbers and the progress bar , i wonder what design thinks about left aligning the text ?

Screen Shot 2020-10-20 at 9 58 25 AM

Screen Shot 2020-10-20 at 9 59 37 AM

mcoker:
> I wonder if there could be a more fluid way of handling this sizing rather than having to add a modifier. Because the percentages are so dynamic a lot of users may not realize they need to add the variant until their percentages are large.

@christiemolloy do you have suggestion or idea as to how to improve it? Not sure if I totally understand the concern.

> There also seems to be a lot of space now between the numbers and the progress bar , i wonder what design thinks about left aligning the text ?

Good point! The status text is currently right-aligned, but I think that's just so that the status is to the right of the progress bar when it's above the bar. Seems like when it's beside the bar, it would be left aligned. WDYT @mcarrano @mceledonia?

mcarrano:
> Good point! The status text is currently right-aligned, but I think that's just so that the status is to the right of the progress bar when it's above the bar. Seems like when it's beside the bar, it would be left aligned. WDYT @mcarrano @mceledonia?

Good question. I don't have a strong preference about this. To me the tradeoff is between having a consistent gap and jagged right edge or vice versa. @mceledonia what do you think?

mattnolting:
Could you update [grid-template-columns](https://github.com/patternfly/patternfly/pull/3567/files#diff-50dd7d2d78cd1db30743e5ef74d70f3f113dbb75fad8b9b19001707a4e3a0cd8L118) to achieve the desired result without the extra markup?

mattnolting (approved):
LPTM!

christiemolloy:
Maybe my comment is more so about the design documentation. Since percentages are dynamic, you could have 1% or 150%. If a user goes with the "On Single Line" and adds multiple of them, then it could end up like the case is here below. I feel like we should add documentation so that the user knows to use "Outside fixed width measure" when using multiple progress bars

Screen Shot 2020-12-02 at 2 48 58 PM

src/patternfly/components/Progress/examples/Progress.md
```diff
@@ -240,3 +267,4 @@ If this component is describing the loading progress of a particular region of a
 | `.pf-m-warning` | `.pf-c-progress` |  Changes the appearance of the progess component to indicate a warning state. |
 | `.pf-m-danger` | `.pf-c-progress` |  Changes the appearance of the progess component to indicate a danger (failure) state. |
 | `.pf-m-truncate` | `.pf-c-progress__description` | Modifies the description to display a single line and truncate any overflow text with ellipses. |
+| `.pf-m-fixed-width` | `.pf-c-progress.pf-m-outside .pf-c-progress__measure` | Modifies the measure element to be a fixed width that will hold 0-100%. Overridable by setting `--pf-c-progress__measure--m-fixed-width--Width`. |
```
christiemolloy, line 270: in the insights project i was in, there was some usage data that went to +1000%, maybe you could update this to support up to 1000% ?

christiemolloy (approved):
two comments but we could follow up on them the rest looks great

src/patternfly/components/Progress/examples/Progress.md
mcoker: I'm trying to target the most common use case, and the component's default max value is 100

Screen Shot 2020-12-02 at 1 56 45 PM

We can support 1000 by default, though that would mean that for anyone using a 0-100% scale, there will be extra space to accommodate 1000%. I see that as an edge case where a user should update `--pf-c-progress__measure--m-fixed-width--Width` - WDYT?

Although I do think I might add another example with a non-standard value that shows how to use the `--pf-c-progress__measure--m-fixed-width--Width` var.

mcoker:
@mcarrano thoughts on @christiemolloy's question? https://github.com/patternfly/patternfly/pull/3567#issuecomment-737457528

mcarrano:
I think @christiemolloy makes a good point about usage and addressing this in the design guidelines. We currently have an open documentation issue here and I will link this comment: https://github.com/patternfly/patternfly-org/issues/2119

src/patternfly/components/Progress/examples/Progress.md
christiemolloy: Yep if you could add an example that would be good

Issue #3572, opened 2020-10-07
Table - add examples showing expanded content with more than one cell with content


mcoker:
This is a follow up to the react PR: https://github.com/patternfly/patternfly-react/pull/4860

Let's add an example of this in core so we can be sure to test that kind of layout when making changes to expandable tables.

Examples are here - https://patternfly-react.surge.sh/components/table#compact-expandable

PR #3573, opened 2020-10-07
chore(table): added examples with multiple expandable cells

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3572

@jessiehuff we have `aria-controls` on the expandable table row toggle buttons and I don't see that attribute in the react component. Can you confirm which is correct?


mattnolting (approved):
LPTM!

Issue #3544, opened 2020-10-01
Data list - page jumps when dragging rows in chrome


mcoker:
related: https://github.com/patternfly/patternfly-react/pull/4852

adding `overflow-anchor: none;` fixes it. Let's add a class to the data list that's added while dragging rows that enables this property. We may also investigate different ways of moving the data list ghost row around in the react component that may not cause shifting like it is currently.

PR #3574, opened 2020-10-07
fix(data-list): added variation to avoid shift when item is dragged over

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3544

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -166,6 +166,10 @@
       font-size: var(--pf-c-data-list--m-compact__check--FontSize);
     }
   }
+
+  &.pf-m-drag-over {
+    overflow-anchor: none;
+  }
 }
 
 .pf-c-data-list,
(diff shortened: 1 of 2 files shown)


src/patternfly/components/DataList/data-list.scss
```diff
@@ -166,6 +166,10 @@
       font-size: var(--pf-c-data-list--m-compact__check--FontSize);
     }
   }
+
+  &.pf-m-drag-over {
```
christiemolloy, line 170: is this modifier being added to the HTML anywhere ? or is it just for react

christiemolloy (changes requested):

src/patternfly/components/DataList/data-list.scss
mcoker: It's just for react to apply once the drag event starts. You can see it being applied as an inline style now in the react component. This just assigns it to a class so react can apply that instead of an inline style.

mattnolting (approved):
LPTM!

christiemolloy (approved):
LGTM

Issue #3379, opened 2020-08-05
fix(togglegroup): investigate height of buttons.


christiemolloy:
Buttons in the toggle group are 37px tall, they should be 36px tall to match the button component.

<img width="371" alt="Screen Shot 2020-08-05 at 10 22 31 AM" src="https://user-images.githubusercontent.com/20118816/89424363-ae760580-d705-11ea-816b-a4cb25b9f2c9.png">

PR #3576, opened 2020-10-08
fix(togglegroup): updated button height to match other form elements

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3379


christiemolloy (approved):
LPTM!!

Issue #3578, opened 2020-10-08
Form control placeholder text color


mcoker:
The code below isn't working because `::placeholder` is a pseudo element, not a pseudo class. So it will inherit color, but any change needs to be applied via a `color` property declaration.

https://github.com/patternfly/patternfly/blob/2f4ad756a10f0bbd1aa9b0f712e018621c166e51/src/patternfly/components/FormControl/form-control.scss#L193-L195

PR #3579, opened 2020-10-08
fix(form-control): updated placeholder block so the color change applies

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3578

@mcarrano @mceledonia the css we had in place to change the placeholder text to black-600 wasn't working. This PR fixes that so the color is now black-600.

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -191,7 +191,7 @@ $pf-c-form-control--m-clock--Coordinates: "M256 8C119 8 8 119 8 256s111 248 248 
   // stylelint-enable
 
   &::placeholder {
-    --pf-c-form-control--Color: var(--pf-c-form-control--placeholder--Color);
+    color: var(--pf-c-form-control--placeholder--Color);
   }
 
   &:not(textarea) {


mcarrano (commented):
I'm still not seeing it @mcoker . Here's what I get:

Screen Shot 2020-10-08 at 4 23 43 PM

mcoker:
@mcarrano sorry I should have been more clear, this will just update the existing placeholder text styling in the form control component. You will either need to add ‘placeholder=“placeholder text”’ to a form control text input in an existing example using dev tools or you can see placeholder text in some of our examples. The phone number input in this form demo has placeholder text https://patternfly-pr-3579.surge.sh/components/form/html-demos#basic

mcarrano:
Let's discuss further @mcoker . The difficulty with select lists is what's the placeholder value. In the example above "Please choose" is really a placeholder and not a value in the list. So how do you declare placeholder text in the HTML?

mcoker:
@mcarrano this doesn't change select lists at all - the `placeholder` attribute works on text inputs and textareas. We had the color for placeholder text defined as black-600 in the form control component code, though it wasn't working correctly so that color wasn't being applied. What was being applied is the browser's default placeholder text color, which is very close to black-600. What that means is that the placeholder text in text inputs was a light grey that came from the browser's styling of placeholder text, and wasn't black-600 like we intended it to be.

That's the discrepancy that you saw in the demo where I styled select menus to appear as if they have placeholder text - I styled the selects to appear as if they had placeholder text by making the text black-600. But when comparing those to the typeahead select (which has placeholder text in a _text input_ in the toggle) the color was different because the select elements' color was black-600 and the typeahead placeholder text was the browser's default placeholder color.

What this PR does is fix text inputs and textareas so that the placeholder text color is black-600 instead of the browser's default placeholder text color... assuming that's intended. Places you can see the difference:

#### App launcher
* current (browser default placeholder color): https://www.patternfly.org/v4/components/application-launcher/html#favorites
* this PR (black-600): http://patternfly-pr-3579.surge.sh/components/application-launcher/html/#favorites

#### File upload
* current: https://www.patternfly.org/v4/components/file-upload/html#basic-file-upload
* this PR: http://patternfly-pr-3579.surge.sh/components/file-upload/html/#basic-file-upload

#### Search input
* current: https://www.patternfly.org/v4/components/search-input/
* this PR: http://patternfly-pr-3579.surge.sh/components/search-input/

#### Typeahead select
* current: https://www.patternfly.org/v4/components/select/#typeahead
* this PR: http://patternfly-pr-3579.surge.sh/components/select/#single-with-typeahead
 
I don't plan on changing anything with the select component until we finish investigating https://github.com/patternfly/patternfly-design/issues/775. This just fixes what appears to be a bug with text inputs and textareas.

Let me know if it's still unclear.

mcarrano:
OK. Yes, I was totally confused then @mcoker . I mistakenly thought that this PR was in response to the design issue you referenced above. So in that case I will go ahead and approve it.

mceledonia (approved):

mattnolting (approved):
LPTM 🌮

christiemolloy (approved):
NICE!!!

Issue #3554, opened 2020-10-05
Introduce expandable card


mcarrano:
Want to support an expandable card that allows a card to be minimized to only show information in the header row when collapsed and show the full height of the card when expanded.  See designs here: https://marvelapp.com/prototype/5e55293

cc @mcoker @gdoyle1 - Please add any further detail that is needed here.

PR #3586, opened 2020-10-09
feat(card): added expandable variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3554

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -26,6 +26,7 @@
   --pf-c-card__header-toggle--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-card__header-toggle--MarginBottom: calc(var(--pf-global--spacer--form-element) * -1);
   --pf-c-card__header-toggle--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-card__header-toggle-icon--Transition: var(--pf-global--Transition);
   --pf-c-card--m-expanded__header-toggle-icon--Rotate: 90deg;
   --pf-c-card__title--FontSize: var(--pf-global--FontSize--md);
   --pf-c-card__title--FontWeight: var(--pf-global--FontWeight--bold);
@@ -117,6 +118,7 @@
 
 .pf-c-card__header-toggle-icon {
   display: inline-block;
+  transition: var(--pf-c-card__header-toggle-icon--Transition);
 }
 
 .pf-c-card__title {
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -22,6 +22,11 @@
   --pf-c-card--child--PaddingRight: var(--pf-global--spacer--lg);
   --pf-c-card--child--PaddingBottom: var(--pf-global--spacer--lg);
   --pf-c-card--child--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-card__header-toggle--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-card__header-toggle--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-card__header-toggle--MarginBottom: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-card__header-toggle--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-card--m-expanded__header-toggle-icon--Rotate: 90deg;
   --pf-c-card__title--FontSize: var(--pf-global--FontSize--md);
   --pf-c-card__title--FontWeight: var(--pf-global--FontWeight--bold);
   --pf-c-card__title--not--last-child--PaddingBottom: var(--pf-global--spacer--md);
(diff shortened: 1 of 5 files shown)


gdoyle1 (approved):
LGTM, thanks @mcoker !!

mcarrano (commented):
Looks good @mcoker . Can other elements be added in the collapsed header as @gdoyle1 showed in her designs (labels in that case)?

mcoker:
@mcarrano the only thing I did here was create the expandable button and sections, so anything else that you could normally put beside the title (actions, labels, text, etc) can be done here. In the case of gina's design, I think the developer would just add their own layout (split, grid, etc) there to have a label group beside the card title.

mcarrano (approved):
OK. In that case I'm good with this @mcoker .

src/patternfly/components/Card/examples/Card.md
```diff
@@ -252,6 +252,48 @@ cssPrefix: pf-c-card
...
+    {{> card-header-toggle}}
+    {{#> card-actions}}
+      {{#> dropdown id=(concat card--id "-dropdown-kebab-right-aligned") dropdown--IsActionMenu="true" dropdown-toggle--modifier="pf-m-plain" dropdown--HasKebabIcon="true" aria-label="Actions"}}
+      {{/dropdown}}
+      <input type="checkbox" id="{{card--id}}-check" name="{{card--id}}-check" aria-labelledby="{{card--id}}-title">
+    {{/card-actions}}
+    {{#> card-title card-title--attribute=(concat 'id="' card--id '-title"')}}
+      Title
+    {{/card-title}}
+  {{/card-header}}
```
mattnolting, line 268: This example omits `{{#> card-expandable-content}}`, which I think is fine if we move in that direction, however all (or most) other `expandable` things are `hidden` when collapsed. It would be more consistent with other componentry if this where `display: none; visibility: hidden;` and `hidden`.

src/patternfly/components/Card/examples/Card.md
```diff
@@ -252,6 +252,48 @@ cssPrefix: pf-c-card
 {{/card}}
 ```
 
+### Expandable (collapsed)
+```hbs
+{{#> card card--id="card-expandable-example"}}
```
mattnolting, line 257:
```suggestion
{{#> card card--id="card-expandable-collapsed-example"}}
```

src/patternfly/components/Card/card.scss
```diff
@@ -87,12 +92,31 @@
...
+  }
+}
+
+.pf-c-card__header-toggle {
+  align-self: flex-start;
+  margin: var(--pf-c-card__header-toggle--MarginTop) var(--pf-c-card__header-toggle--MarginRight) var(--pf-c-card__header-toggle--MarginBottom) var(--pf-c-card__header-toggle--MarginLeft);
+}
+
+.pf-c-card__header-toggle-icon {
+  display: inline-block;
```
mattnolting, line 119: To be consistent with other expandable things:

```
  transition: var(--pf-c-card__header-toggle-icon--Transition);
```

src/patternfly/components/Card/card.scss
```diff
@@ -22,6 +22,11 @@
   --pf-c-card--child--PaddingRight: var(--pf-global--spacer--lg);
   --pf-c-card--child--PaddingBottom: var(--pf-global--spacer--lg);
   --pf-c-card--child--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-card__header-toggle--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-card__header-toggle--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-card__header-toggle--MarginBottom: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-card__header-toggle--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-card--m-expanded__header-toggle-icon--Rotate: 90deg;
   --pf-c-card__title--FontSize: var(--pf-global--FontSize--md);
```
mattnolting, line 30: --pf-c-card__header-toggle-icon--Transition: var(--pf-global--Transition);

src/patternfly/components/Card/examples/Card.md
mcoker: Yeah, I did that specifically here because the card CSS looks for adjacent elements to apply spacing. I'm in favor of setting up our examples like this as a way to indicate to react how the element should be hidden. WDYT @christiemolloy?

src/patternfly/components/Card/card.scss
mcoker: Good call, I forgot to add that.

christiemolloy (changes requested):
Should the spacing to the right of the toggle icon be the same as expandable - looks too big here ?
Screen Shot 2020-10-14 at 4 02 37 PM
Screen Shot 2020-10-14 at 4 02 45 PM

christiemolloy (commented):
Can you add the toggle icon to the left of the img in the header?

mcoker:
> Should the spacing to the right of the toggle icon be the same as expandable - looks too big here ?

@christiemolloy I don't think they need to match. The expandable is unique in that the toggle icon and text are all the expandable toggle. In the card, we're using a plain button next to the text. We could make them a bit closer, but then the clickable area of the title and toggle would touch, so it could be a problem if both the title and toggle were clickable

Existing spacing
Screen Shot 2020-10-14 at 3 18 15 PM

Space between toggle and title removed
Screen Shot 2020-10-14 at 3 18 30 PM

Showing the clickable area with the space removed
Screen Shot 2020-10-14 at 3 18 37 PM

mcoker:
> Can you add the toggle icon to the left of the img in the header?

@christiemolloy Yep, the toggle would go to the left of everything that normally goes in the card header. Or are you asking to add an example with that?

christiemolloy:
> > Can you add the toggle icon to the left of the img in the header?
> 
> @christiemolloy Yep, the toggle would go to the left of everything that normally goes in the card header. Or are you asking to add an example with that?

@mcoker yeah i think it would be good to see an example of that

mattnolting (approved):
LPTM!

src/patternfly/components/Card/examples/Card.md
mattnolting: @jessiehuff Can you see any accessibility issue here? ^^

christiemolloy (approved):
Looks great!!!

src/patternfly/components/Card/examples/Card.md
jessiehuff: @mattnolting I can see the pros and cons to this approach. I know that we've done this in other components like the dropdown, but I'm concerned that this might have caused the dropdown issue we noticed in JAWS. We created a workaround by adding additional key listeners to shift focus, but I suspect that since the content didn't exist before JAWS opened the dropdown, it didn't recognize it and announce it. So I suppose it's something to be on the lookout for when we use this approach. I'm guessing that whenever react implements it, we'll just need to test to make sure it doesn't have the same issue. @mcoker also had an idea of how to test out this theory as well to see if it was the culprit. 🙂

Issue #3587, opened 2020-10-12
Modal close button and content overlap


mcoker:
Need to add more space here to accommodate the close button.
<img width="1078" alt="Screen Shot 2020-10-12 at 2 44 52 PM" src="https://user-images.githubusercontent.com/35148959/95784505-92f52100-0c99-11eb-88b8-8ddc08bb7498.png">

PR #3588, opened 2020-10-12
fix(modal): increased space between close button and what preceedes it

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3587

Verified this is correct in the react component. Core component is a little off since the icon isn't 16px wide.

Before
Screen Shot 2020-10-12 at 3 12 01 PM

After
Screen Shot 2020-10-12 at 3 11 04 PM


christiemolloy (approved):
LGTM

Issue #1860, opened 2019-05-27
Add the ability to have favorites in a table view 


beanh66:
In OpenShift we have a need to allow users to pin or favorite certain items in a table view for quicker access when scanning long lists. The primary use case is for the projects table because users can have hundreds of these but typically only interact with a few at a time. There are other use cases within OpenShift to add this mechanism to tables as well, such as Roles. For this case, we would like to "favorite" a few important roles by default and then extend the behavior to allow users to add or remove from the favorites set by default.

After a number of design iterations, we landed on the following design for the use cases described above: http://openshift.github.io/openshift-origin-design/web-console/future-openshift/pinning/pinning

Seems like this would be worth contributing as a generic component for PF. @bmignano worked on the original designs and can help with a design contribution if desired @mcarrano!

mcarrano:
@beanh66 Thanks for submitting this. I think it will be a useful feature. Once this is slotted on the PF roadmap, I will be happy to work with @bmignano on the design. @LHinson @rachael-phillips please note.

rachael-phillips:
@mcarrano thanks for looping me in - I've added it to the component status unplanned section for the time being. Let's continue to discuss with @LHinson on where it should live

mcarrano:
@beanh66 @bmignano We are talking about moving forward with this and I just wanted to check in to see if the designs posted on May 27 are still the desired approach. If so, we can talk about next steps.

mcarrano:
Per today's PF Weekly mtg I will add this to our design backlog. @bmignano if you have bandwidth to work on this, I'd be glad to have you lead that effort. @LHinson @rachael-phillips let's consider where this can be slotted on the PF roadmap.

bmignano:
@mcarrano Happy to continue working on this!

mcarrano:
That's great @bmignano. I will assign this to you and set up some time for us to discuss next steps.

mcarrano:
@bmignano @mceledonia where do you feel like we stand with this following last week's review? I think we are in good shape excepting maybe a couple of minor loose ends. Is this something we can wrap up this week?

gdoyle1:
Hey @mcarrano @LHinson ! 
OpenShift is planning on adding favorites in table view and our "project selector" (which is a dropdown right now and _I think_ could use the same type of styling as favorited items in application launcher) for our 4.7 release in August! 
Do you think we could get this back on the roadmap for that timeline?

mcarrano:
[...]

PR #3594, opened 2020-10-13
feat(table): added favorites

mcoker:
fixes https://github.com/patternfly/patternfly/issues/1860


mcarrano (commented):
This looks good @mcoker . My only question is whether the sortable favorites column should be left aligned so that the stars align when there is a sortable header. That would be more consistent with other column headers. Thoughts?

mcoker:
@mcarrano yep I had a bug in the spacing, should be good now!

mcarrano (approved):
Thanks @mcoker . That looks great. @mceledonia can you also take a look?

christiemolloy (commented):
Should there be an aria relationship between the favorite star and the text in the column next to it?

christiemolloy (commented):
Screen Shot 2020-10-14 at 3 59 41 PM

should the star have less padding on the left and right side so that these are closer together

mcoker:
@christiemolloy I can see adding that. WDYT @jessiehuff? I set this up like the app-launcher/select where the label for the button is just "starred" and "not starred". Though the item text in the app-launcher/select come before the favorite button, and here the favorite button comes before any text cell in the row. Should we use an ID for the title of a row, and add that to `aria-labelledby` on the button so it reads like "{{title}} starred/not starred"?

That seems to be how the expandable table is setup in the react component. The button has a label of "details" and is `aria-labelled` by the id of the row title and the button. An aside, but the react examples do not have an element in the row with the ID that is being used with `aria-labelledby` in the expandable examples.

mcoker:
> should the star have less padding on the left and right side so that these are closer together

@christiemolloy right now the buttons (favorite and action) are adjacent to one another with no spacing. We could do that, but I think we would want to update the action, too, so that the target areas are the same for both of them. If design is OK with the way it is now, I'm in favor of leaving it.

src/patternfly/components/Table/table.scss
```diff
@@ -95,6 +95,16 @@
   --pf-c-table__check--input--MarginTop: #{pf-size-prem(4px)};
   --pf-c-table__check--input--FontSize: var(--pf-global--FontSize--md);
 
+  // Favorite
+  --pf-c-table--cell--m-favorite--Color: var(--pf-global--Color--light-300);
+  --pf-c-table__favorite--c-button--Color: var(--pf-global--Color--light-300);
```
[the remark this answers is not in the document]
mcoker: `--pf-c-table--cell--m-favorite--Color` is for the sortable table header button, and `--pf-c-table__favorite--c-button--Color` is for the body cell button.

src/patternfly/components/Table/table-grid.scss
```diff
@@ -380,7 +390,9 @@
   tr:not(.pf-c-table__expandable-row) {
     .pf-c-table__toggle,
     .pf-c-table__check,
+    .pf-c-table__favorite,
     .pf-c-table__action {
+      align-self: center;
```
mattnolting, line 395: When text wraps, these elements shift down 

Screen Shot 2020-10-15 at 10 27 52 AM

mattnolting (approved):
Fantastic work!!! 🥇

christiemolloy (approved):
Nice job!!! Looks great

jessiehuff:
> @christiemolloy I can see adding that. WDYT @jessiehuff? I set this up like the app-launcher/select where the label for the button is just "starred" and "not starred". Though the item text in the app-launcher/select come before the favorite button, and here the favorite button comes before any text cell in the row. Should we use an ID for the title of a row, and add that to `aria-labelledby` on the button so it reads like "{{title}} starred/not starred"?
> 
> That seems to be how the expandable table is setup in the react component. The button has a label of "details" and is `aria-labelled` by the id of the row title and the button. An aside, but the react examples do not have an element in the row with the ID that is being used with `aria-labelledby` in the expandable examples.

- I wouldn't recommend putting `aria-label` on the icon because it results in different examples between core and react, where core uses `<i>` and react uses `<svg>` for icons. Instead, I'd move the label to the button so that it's always on the same element.
- I'd recommend the approach you're considering similar to the expandable example: an `aria-label` to describe the action and an `aria-labelledby` to associate the row header with the action. This is primarily due to the row header coming after the button in the row and therefore not getting announced (at least in combinations tested). An alternative to this method would be to use only `aria-label` but state both the action name and the row name, but I think using `aria-labelledby` here makes sense.

mcoker:
@jessiehuff updated so the button is `aria-labelledby` the row title + itself.

jessiehuff (approved):
LGTM! 🙂

src/patternfly/components/Table/examples/Table.md
```diff
@@ -2168,12 +2168,279 @@ By default, truncation and wrapping settings do not affect the grid layout, but
...
+  {{/table-thead}}
 
-| Class | Applied to | Outcome |
+  {{#> table-tbody}}
+    {{#> table-tr table-tr--id="1"}}
+      {{#> table-td table-td--check="true"}}
+        <input type="checkbox" name="{{table--id}}-checkrow{{table-tr--id}}" aria-labelledby="{{concat table--id '-node' table-tr--id}}">
+      {{/table-td}}
+      {{> table-td table-td--favorite="true" table-td--IsFavorited="true"}}
+      {{#> table-td table-td--data-label="Repository name"}}
```
mattnolting, line 2205: Looks like we need a text wrapper `{{#> table-td--text}}`

Screen Shot 2020-10-20 at 5 00 39 PM

mcoker: Doh! Good catch. Just wrapped that in a `<div>` since that's what we're doing in most of the other examples with cells that have elements.

christiemolloy (approved):
LPTM!!!

PR #3596, opened 2020-10-15
feat(jump-links): add component

### LINKED ISSUE #3558: Scrollspy component
Introduce a new Scrollspy navigation component per the designs here: https://marvelapp.com/prototype/25f068g6 This should include both horizontal and vertical variants.

christiemolloy:
closes #3558


mcarrano (commented):
This is looking good @christiemolloy . Just a bit unclear what the 'Horizontal with background' example is intended to show. Is this just a centered version? Does the gray background become part of the component?

@lboehling @ajacobs21e would be good if you can also take a look.

ajacobs21e:
@lboehling did you want to keep the label all caps? I don't have a preference but if we do change it to sentence case we'll have to update PFE. 

@mcarrano The background is part of the component for the horizontal version. There would need to be some amount of padding around the component that's sticky and on top. So maybe do we just need the one version with a white background but then it's optional if you want to change to a different color background?
The vertical one has an option to put it on a card in case it passes across different background bands. It's a similar idea - basically there is some amount of space around the component that can't be interrupted by other elements.

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+    --pf-c-scrollspy__item-text--Color: var(--pf-c-scrollspy__link--focus__item-text--Color);
+    --pf-c-scrollspy__link--before--BorderTopWidth: var(--pf-c-scrollspy__link--focus--before--BorderTopWidth);
+    --pf-c-scrollspy__link--before--BorderLeftWidth: var(--pf-c-scrollspy__link--focus--before--BorderLeftWidth);
+    --pf-c-scrollspy__link--before--BorderColor: var(--pf-c-scrollspy__link--focus--before--BorderColor);
+  }
+
+  &::before {
+    position: absolute;
+    top: 0;
+    right: 0;
```
mattnolting, line 126: You don't need this declaration

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+
+  + .pf-c-scrollspy__link {
+    --pf-c-scrollspy__link--PaddingLeft: var(--pf-c-scrollspy__link--link--PaddingLeft);
+  }
+}
+
+// stylelint-disable
+.pf-c-scrollspy__item {
+  &.pf-m-current {
+    display: flex;
```
mattnolting, line 143: To increase clickable area, seems like it would be better to move `display: flex; to line [103](https://github.com/patternfly/patternfly/pull/3596/files#diff-32f68850123d67016f257f021955f786ca61f4bbe4e33f01ecd501e305399e2fR103) and set `.pf-c-scrollspy__link` to `flex-basis: 100%;` or `flex: 1 1 100%;`

Screen Shot 2020-10-20 at 3 29 15 PM

src/patternfly/components/Scrollspy/scrollspy.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-scrollspy{{#if scrollspy--modifier}} {{scrollspy--modifier}}{{/if}}"
```
mattnolting, line 1: Should be `nav` in order for VO to announce it as such and probably should always have a label, even if it's hidden.

mcarrano:
> The background is part of the component for the horizontal version. There would need to be some amount of padding around the component that's sticky and on top. So maybe do we just need the one version with a white background but then it's optional if you want to change to a different color background?
The vertical one has an option to put it on a card in case it passes across different background bands. It's a similar idea - basically there is some amount of space around the component that can't be interrupted by other elements.

Thanks for the explanation @ajacobs21e . I think that part of my confusion is the way the examples are shown. So [this example](https://patternfly-pr-3596.surge.sh/components/scrollspy#horizontal-with-background) called "horizontal with background" is actually showing 2 things, is that right - the scrollspy nav with a background and also centered?  @christiemolloy maybe we should either split into two examples or make it clear that you're combining two independent things into the same example.

src/patternfly/components/Scrollspy/scrollspy-item-text.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-scrollspy__item-text{{#if scrollspy-item-text--modifier}} {{scrollspy-item-text--modifier}}{{/if}}"
```
mcoker, line 1: needs to be a span since it's in a button

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+
+    .pf-c-scrollspy__list {
+      flex-direction: column;
+    }
+  }
+}
+
+.pf-c-scrollspy__list {
+  position: relative;
+  display: inline-flex;
```
mcoker, line 85: Right now the container is a block element, but this list is inline. Seems like we would either want both the container and list to be inline (and the component will only be as wide as it needs to be), or both to be block (and the container it's in can manage the width)

current container
Screen Shot 2020-10-21 at 10 57 41 AM

current list item
Screen Shot 2020-10-21 at 10 57 35 AM

if both were block
Screen Shot 2020-10-21 at 10 56 24 AM

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+  --pf-c-scrollspy--m-vertical__item--m-current__link--before--BorderLeftWidth: var(--pf-global--BorderWidth--lg);
+
+  // text
+  --pf-c-scrollspy__item-text--Color: var(--pf-global--Color--200);
+  --pf-c-scrollspy__link--hover__item-text--Color: var(--pf-global--Color--100);
+  --pf-c-scrollspy__link--focus__item-text--Color: var(--pf-global--Color--100);
+  --pf-c-scrollspy__item--m-current__item-text--Color: var(--pf-global--Color--100);
+
+  // label
+  --pf-c-scrollspy__label--PaddingTop: var(--pf-global--spacer--md);
```
mcoker, line 52: Do you need this? Shouldn't it come from the container it's in?

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+  --pf-c-scrollspy__label--PaddingBottom: var(--pf-global--spacer--md);
+
+  padding-right: var(--pf-c-scrollspy--PaddingRight);
+  padding-left: var(--pf-c-scrollspy--PaddingLeft);
+  background-color: var(--pf-c-scrollspy--BackgroundColor);
+
+  &.pf-m-background {
```
mcoker, line 59: Is this needed? Seems like the section it's in would define the background.

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+    --pf-c-scrollspy__list--PaddingLeft: 0;
+    --pf-c-scrollspy__list--before--BorderTopWidth: var(--pf-c-scrollspy--m-vertical__list--before--BorderTopWidth);
+    --pf-c-scrollspy__list--before--BorderLeftWidth: var(--pf-c-scrollspy--m-vertical__list--before--BorderLeftWidth);
+    --pf-c-scrollspy__link--focus--before--BorderTopWidth: var(--pf-c-scrollspy--m-vertical__link--focus--before--BorderTopWidth);
+    --pf-c-scrollspy__link--focus--before--BorderLeftWidth: var(--pf-c-scrollspy--m-vertical__link--focus--before--BorderLeftWidth);
+    --pf-c-scrollspy__item--m-current__link--before--BorderTopWidth: var(--pf-c-scrollspy--m-vertical__item--m-current__link--before--BorderTopWidth);
+    --pf-c-scrollspy__item--m-current__link--before--BorderLeftWidth: var(--pf-c-scrollspy--m-vertical__item--m-current__link--before--BorderLeftWidth);
+
+    .pf-c-scrollspy__list {
+      flex-direction: column;
```
mcoker, line 78: you could make this a var, too

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+
+.pf-c-scrollspy__list {
+  position: relative;
+  display: inline-flex;
+  padding-right: var(--pf-c-scrollspy__list--PaddingRight);
+  padding-left: var(--pf-c-scrollspy__list--PaddingLeft);
+  margin: 0;
+
+  &::before {
+    position: absolute;
```
mcoker, line 91: can you add `pointer-events: none;` so you aren't clicking on the pseudo element when you click the button?

src/patternfly/components/Scrollspy/examples/scrollspy.md
```diff
@@ -0,0 +1,362 @@
...
+          Inactive section
+        {{/scrollspy-item-text}}
+      {{/scrollspy-link}}
+    {{/scrollspy-item}}
+    {{#> scrollspy-item scrollspy-item--modifier="pf-m-current"}}
+      {{#> scrollspy-link}}
+        {{#> scrollspy-item-text}}
+          Active expandable section
+        {{/scrollspy-item-text}}
+        {{#> scrollspy-link scrollspy-link--modifier="pf-m-subsection"}}
```
mcoker, line 302: should this be in a new list?

src/patternfly/components/Scrollspy/scrollspy.scss
mcoker: Also right now the items' click area is different depending on the content, which we don't do in vertical lists AFAIK.

Screen Shot 2020-10-21 at 11 08 22 AM
Screen Shot 2020-10-21 at 11 08 13 AM

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+  padding-right: var(--pf-c-scrollspy--PaddingRight);
+  padding-left: var(--pf-c-scrollspy--PaddingLeft);
+  background-color: var(--pf-c-scrollspy--BackgroundColor);
+
+  &.pf-m-background {
+    --pf-c-scrollspy--BackgroundColor: var(--pf-c-scrollspy--m-background--BackgroundColor);
+  }
+
+  &.pf-m-center {
+    text-align: center;
```
mcoker, line 64: I would probably make the parent flex, and set `justify-content` to center the box, versus an inheritable property like `text-align` which may require resetting the alignment on nested children if we add more things that shouldn't be center aligned, since it applies to any child that is inline.

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+}
+
+.pf-c-scrollspy__list {
+  position: relative;
+  display: inline-flex;
+  padding-right: var(--pf-c-scrollspy__list--PaddingRight);
+  padding-left: var(--pf-c-scrollspy__list--PaddingLeft);
+  margin: 0;
```
mcoker, line 88: Is this needed?

src/patternfly/components/Scrollspy/scrollspy.scss
```diff
@@ -0,0 +1,168 @@
...
+
+  &::before {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    content: "";
+    border-color: var(--pf-c-scrollspy__list--before--BorderColor);
+    border-style: solid;
```
mcoker, line 98: you could combine these `border: solid var(--pf-c-scrollspy__list--before--BorderColor);`

mcoker:
I'm confused as to whether the grey line above the links should extend beyond the width of the items or not. Do we need to build in the ability for users to customize that line's width like we do with tabs? https://www.patternfly.org/v4/components/tabs/html#inset

Seems like there is a lot of overlap with tabs.

Screen Shot 2020-10-21 at 11 21 54 AM

Screen Shot 2020-10-21 at 11 22 27 AM

src/patternfly/components/Scrollspy/examples/scrollspy.md
```diff
@@ -0,0 +1,362 @@
...
+      {{/scrollspy-link}}
+    {{/scrollspy-item}}
+  {{/scrollspy-list}}
+{{/scrollspy}}
+```
+
+### Horizontal with label
+```hbs
+{{#> scrollspy}}
+  {{#> scrollspy-main}}
```
mcoker, line 70: Currently it doesn't look like `__main` is necessary, though I'm good with adding it. If we keep it, I wonder if it could just always be present, and set to `display: flex; flex-direction: column;`. That will also keep react styles happy since it will have CSS.

christiemolloy:
> > The background is part of the component for the horizontal version. There would need to be some amount of padding around the component that's sticky and on top. So maybe do we just need the one version with a white background but then it's optional if you want to change to a different color background?
> > The vertical one has an option to put it on a card in case it passes across different background bands. It's a similar idea - basically there is some amount of space around the component that can't be interrupted by other elements.
> 
> Thanks for the explanation @ajacobs21e . I think that part of my confusion is the way the examples are shown. So [this example](https://patternfly-pr-3596.surge.sh/components/scrollspy#horizontal-with-background) called "horizontal with background" is actually showing 2 things, is that right - the scrollspy nav with a background and also centered? @christiemolloy maybe we should either split into two examples or make it clear that you're combining two independent things into the same example.

Separated the examples into two and made the text capitalized

ajacobs21e:
@mcoker "whether the grey line above the links should extend beyond the width of the items or not"
do you mean this?
image

Since the vertical jump links have a leading and trailing line, I'd prefer if the horizontal ones do as well. It's a slight visual distinction from tabs. @lboehling is that good with you?

I have not heard of a need to have an inset option for jump links like tabs do. 

Are we calling this component jump links or scrollspy? PFE wants to call it jump links because "scrollspy" means something specific in code and this is a component that uses scrollspy but isn't equivalent.

src/patternfly/components/Scrollspy/scrollspy.scss
christiemolloy: what do you mean by the section its in would define the background ? like the list would ?

mcoker:
> Are we calling this component jump links or scrollspy? PFE wants to call it jump links because "scrollspy" means something specific in code and this is a component that uses scrollspy but isn't equivalent.

+1 for jump links for the same reason - that's what they are, scrollspy would be something that detects the scroll position and updates the links as active/inactive.

mcoker:
> I have not heard of a need to have an inset option for jump links like tabs do.

@ajacobs21e I think it would depend on how the jump links are used. Could the grey line extend to the left edge of a page section or card, and the text in the first link align with the content in the page section/card? That's one of the use cases for insets in tabs.

src/patternfly/components/Scrollspy/scrollspy.scss
mcoker: This looks like a page/header section that has scrollspy in it, as opposed to the whole grey box being the scrollspy component.

Screen Shot 2020-10-21 at 12 45 05 PM

src/patternfly/components/Scrollspy/scrollspy.scss
christiemolloy: when i removed this the current border wasnt being applied

src/patternfly/components/Scrollspy/scrollspy.scss
christiemolloy: @ajacobs21e what are your thoughts here ?

src/patternfly/components/Scrollspy/scrollspy.scss
christiemolloy: Screen Shot 2020-10-21 at 1 55 12 PM

This is what happens when i move the padding to the element , it shouldnt have this space above it only space above the label when the label is used. Should i move it back?

mcoker: Using the screenshot from https://github.com/patternfly/patternfly/pull/3596#discussion_r509496832, it looks like the padding above the scrollspy label would come from the section it's in, not from the scrollspy component. Typically we don't add outside space like that, and let the space be defined by the spacing of where it's used. That would mean just adding space to the bottom of the label (though I think it should be a margin, not padding)

ajacobs21e:
> > I have not heard of a need to have an inset option for jump links like tabs do.
> 
> @ajacobs21e I think it would depend on how the jump links are used. Could the grey line extend to the left edge of a page section or card, and the text in the first link align with the content in the page section/card? That's one of the use cases for insets in tabs.

I don't think so. I think the line would be aligned with the content. it should feel like part of the content rather than a wrapper around the content (like a tab).

src/patternfly/components/Scrollspy/scrollspy.scss
ajacobs21e: Yeah this is an example of a band that includes a logo, the jump links component, and a button. The whole band would need to be sticky and above the content (content scrolls beneath it).

src/patternfly/components/Scrollspy/scrollspy.scss
christiemolloy: this wasnt working for me when i tried making them vars - ill revisit if i have time

src/patternfly/components/Scrollspy/scrollspy.scss
ajacobs21e: Is that a combination of components or would the jumplinks component have a way to put other components in / on it?

mcoker: On our side, the grey box is a stick page section. Inside the grey box would be a split/flex/grid layout with the icon in left part of the layout, scrollspy in the center, and button in the right. So the scrollspy wouldn't need a grey background - it doesn't have a background and can be used in other things that have white or grey backgrounds.

lboehling:
> @lboehling did you want to keep the label all caps? I don't have a preference but if we do change it to sentence case we'll have to update PFE.

I prefer lowercase since that is more consistent with existing PF text styles. 

> I don't think so. I think the line would be aligned with the content. it should feel like part of the content rather than a wrapper around the content (like a tab).

I agree with @ajacobs21e about not needing an inset.

also +1 on calling it jump links.

src/patternfly/components/Scrollspy/examples/scrollspy.md
christiemolloy: yep just added those styles actually

mcarrano:
I OK with renaming this Jump links, however for the React implementation I was assuming that this would have scrollspy behaviors. Why would we ever want it to behave otherwise @ajacobs21e ?

ajacobs21e:
> I OK with renaming this Jump links, however for the React implementation I was assuming that this would have scrollspy behaviors. Why would we ever want it to behave otherwise @ajacobs21e ? Without that behavior, isn't this just a restyled nav bar?

@mcarrano It's a component that uses scrollspy. From Coker:
+1 for jump links for the same reason - that's what they are, scrollspy would be something that detects the scroll position and updates the links as active/inactive.

mattnolting (approved):
Looks great!

mcarrano:
@mcoker So for the React implementation would you then see scrollspy as something that is built into the Jump links component and can be enabled/disabled via a prop or as a utility class of some type that can be applied?

mcoker:
@mcarrano that sounds right to me! Scrollspy is a feature of jump links.

src/patternfly/components/JumpLinks/examples/JumpLinks.md
```diff
@@ -0,0 +1,306 @@
...
+    {{/jump-links-item}}
+    {{#> jump-links-item jump-links-item--modifier="pf-m-current"}}
+      {{#> jump-links-link}}
+        {{#> jump-links-item-text}}
+          Active expandable section
+        {{/jump-links-item-text}}
+      {{/jump-links-link}}
+        {{#> jump-links-list newcontext}}
+          {{#> jump-links-item}}
+            {{#> jump-links-link jump-links-link--modifier="pf-m-subsection"}}
```
mcoker, line 243: I don't know about having to add a `pf-m-subsection` to every nested link. Seems like we could handle that with nested link selectors in the CSS, similar to tree view.

src/patternfly/components/JumpLinks/jump-links-link.hbs
```diff
@@ -0,0 +1,6 @@
+<button class="pf-c-jump-links__link{{#if jump-links-link--modifier}} {{jump-links-link--modifier}}{{/if}}"
+  {{#if jump-links-link--attribute}}
```
mcoker, line 2:
```suggestion
  type="button"
  {{#if jump-links-link--attribute}}
```

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -0,0 +1,169 @@
+.pf-c-jump-links {
+  --pf-c-jump-links--PaddingRight: var(--pf-global--spacer--lg);
```
mcoker, line 2: dont need this

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -0,0 +1,169 @@
...
+  --pf-c-jump-links--PaddingLeft: var(--pf-global--spacer--lg);
```
mcoker, line 3: not needed

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -0,0 +1,169 @@
...
+    --pf-c-jump-links__list--PaddingBottom: var(--pf-c-jump-links--m-vertical__list--PaddingBottom);
+    --pf-c-jump-links__list--PaddingLeft: var(--pf-c-jump-links--m-vertical__list--PaddingLeft);
+    --pf-c-jump-links__list--before--BorderTopWidth: var(--pf-c-jump-links--m-vertical__list--before--BorderTopWidth);
+    --pf-c-jump-links__list--before--BorderLeftWidth: var(--pf-c-jump-links--m-vertical__list--before--BorderLeftWidth);
+    --pf-c-jump-links__link--focus--before--BorderTopWidth: var(--pf-c-jump-links--m-vertical__link--focus--before--BorderTopWidth);
+    --pf-c-jump-links__link--focus--before--BorderLeftWidth: var(--pf-c-jump-links--m-vertical__link--focus--before--BorderLeftWidth);
+    --pf-c-jump-links__item--m-current__link--before--BorderTopWidth: var(--pf-c-jump-links--m-vertical__item--m-current__link--before--BorderTopWidth);
+    --pf-c-jump-links__item--m-current__link--before--BorderLeftWidth: var(--pf-c-jump-links--m-vertical__item--m-current__link--before--BorderLeftWidth);
+    --pf-c-jump-links__list--FlexDirection: var(--pf-c-jump-links--m-vertical__list--FlexDirection);
```
mcoker, line 79: do you need this and the `flex-direction` declaration below?

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -0,0 +1,169 @@
...
+    content: "";
+    border-color: var(--pf-c-jump-links__link--before--BorderColor);
+    border-style: solid;
+    border-width: var(--pf-c-jump-links__link--before--BorderTopWidth) var(--pf-c-jump-links__link--before--BorderRightWidth) var(--pf-c-jump-links__link--before--BorderBottomWidth) var(--pf-c-jump-links__link--before--BorderLeftWidth);
+  }
+}
+
+// stylelint-disable
+.pf-c-jump-links__item {
+  list-style: none;
```
mcoker, line 144: is this necessary?

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -0,0 +1,169 @@
...
+}
+// stylelint-enable
+
+.pf-c-jump-links__link-text {
+  color: var(--pf-c-jump-links__link-text--Color);
+  text-align: left;
+}
+
+.pf-c-jump-links__label {
+  padding-bottom: var(--pf-c-jump-links__label--PaddingBottom);
```
mcoker, line 163: should this be a margin?

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -0,0 +1,169 @@
...
+      --pf-c-jump-links__link--before--BorderColor: var(--pf-c-jump-links__item--m-current__link--before--BorderColor);
+      --pf-c-jump-links__link-text--Color: var(--pf-c-jump-links__item--m-current__link-text--Color);
+    }
+  }
+}
+// stylelint-enable
+
+.pf-c-jump-links__link-text {
+  color: var(--pf-c-jump-links__link-text--Color);
+  text-align: left;
```
mcoker, line 159: is this needed?

src/patternfly/components/JumpLinks/jump-links.scss
christiemolloy: yeah, flex-direction: column needs to be set on the jump link and jump link list and it wasnt when i added the line on 78

src/patternfly/components/JumpLinks/jump-links.scss
christiemolloy: not anymore not that subsections is removed

src/patternfly/components/JumpLinks/jump-links.scss
mcoker: oh sorry, I thought they applied to the same element.

PR #3598, opened 2020-10-15
feat(actionlist): add component

### LINKED ISSUE #2021: Create action list component

christiemolloy:
closes #2021

### DIFF
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,0 +1,34 @@
+.pf-c-action-list {
+  --pf-c-action-list--Gap: var(--pf-global--spacer--2xl);
+  --pf-c-action-list--m-vertical--Display: inline-flex;
+  --pf-c-action-list--m-vertical--Gap: var(--pf-global--spacer--lg);
+
+  // Group
+  --pf-c-action-list__group--Gap: var(--pf-global--spacer--md);
+  --pf-c-action-list__group--m-vertical--Gap: var(--pf-global--spacer--sm);
+
+  display: flex;
+  gap: var(--pf-c-action-list--Gap);
+
+  &.pf-m-icons {
+    --pf-c-action-list__group--Gap: 0;
+  }
+
+  &.pf-m-vertical {
+    --pf-c-action-list--Gap: var(--pf-c-action-list--m-vertical--Gap);
+
+    display: inline-flex;
+    flex-direction: column;
+  }
+}
+
+.pf-c-action-list__group {
+  display: flex;
+  gap: var(--pf-c-action-list__group--Gap);
+
+  &.pf-m-vertical {
+    --pf-c-action-list__group--Gap: var(--pf-c-action-list__group--m-vertical--Gap);
+
+    flex-direction: column;
+  }
+}
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,7 +1,7 @@
 .pf-c-action-list {
   --pf-c-action-list--spacer: var(--pf-global--spacer--md);
   --pf-c-action-list__group-group--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
-  --pf-c-action-list__group-group--m-vertical--Gap: var(--pf-global--spacer--lg);
+  --pf-c-action-list__group-group--m-vertical--spacer: var(--pf-global--spacer--lg);
   --pf-c-action-list--m-vertical--spacer: var(--pf-global--spacer--sm);
 
   &.pf-m-icons {
@@ -35,7 +35,7 @@
     + .pf-c-action-list__item {
       --pf-c-action-list__group-group--spacer: 0;
 
-      margin-top: var(--pf-c-action-list__group-group--m-vertical--Gap);
+      margin-top: var(--pf-c-action-list__group-group--m-vertical--spacer);
       margin-left: 0;
     }
   }
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,6 +1,7 @@
 .pf-c-action-list {
   --pf-c-action-list--spacer: var(--pf-global--spacer--md);
-  --pf-c-action-list__group-group--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
+  --pf-c-action-list__group--group--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
+  --pf-c-action-list__group--item--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
 }
 
 .pf-c-action-list,
@@ -15,9 +16,12 @@
     --pf-c-action-list--spacer: 0;
   }
 
-  + .pf-c-action-list__group,
+  + .pf-c-action-list__group {
+    margin-left: var(--pf-c-action-list__group--group--spacer);
+  }
+
   + .pf-c-action-list__item {
-    margin-left: var(--pf-c-action-list__group-group--spacer);
+    margin-left: var(--pf-c-action-list__group--item--spacer);
   }
 }
 
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,28 +1,42 @@
 .pf-c-action-list {
-  --pf-c-action-list--Gap: var(--pf-global--spacer--md);
-  --pf-c-action-list--m-group--Gap: var(--pf-global--spacer--2xl);
-  --pf-c-action-list--m-group--m-vertical--Gap: var(--pf-global--spacer--lg);
-  --pf-c-action-list--m-vertical--Gap: var(--pf-global--spacer--sm);
-
-  &.pf-m-group {
-    gap: var(--pf-c-action-list--m-group--Gap);
-  }
+  --pf-c-action-list--spacer: var(--pf-global--spacer--md);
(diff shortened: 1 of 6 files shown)


mcarrano (dismissed):
Looks good @christiemolloy . But a couple of questions/comments. This is a bit confusing to me:

Screen Shot 2020-10-17 at 5 48 18 PM

Yes, those a two alternate spacings, but both show buttons for a wizard (next, back, and cancel). Maybe change the first example to something else like Save and Cancel? What do you think @mmenestr ?

Also, shouldnt the Cancel button be a link style button (blue)?

mmenestr:
@christiemolloy for some reason in my preview, there are no spaces anywhere?
image

But yes to what @mcarrano said! I agree that for the "Action list with cancel button: In modals, forms, data lists" example, "Save" and "Cancel" would make more sense - with just two buttons, with Save being primary, and Cancel being a link button. 

Otherwise, looks good - just can't see the spacing for wtv reason so I can't really say haha

christiemolloy:
Updated @mmenestr @mcarrano , let me know if you still cant see the spacing and ill send you screenshots

mcarrano (commented):
This looks much better @christiemolloy . I'm still not sure about the stacked examples. Is the Cancel button right-aligned? That doesn't seem right but not exactly sure what we are trying to illustrate here. I'll defer to @mmenestr .

mmenestr:
@mcarrano the button is left aligned, but since the cancel button is wider than "next" and "back" it seems at first glance like it's right aligned. The thinking behind that demo was to show what would happen in a mobile view if everything were to get stacked up (ie: how it would react responsively). And what would happen is that it would stack with a left-alignment so that the buttons are still left-aligned in the modal or card or wherever they may be.

I think it's ready to go now! Apart from just adding the page break between the stacked text, and the buttons so that the button appear below it and not in the same line

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,28 @@
+.pf-c-action-list {
+  --pf-c-action-list--Gap: var(--pf-global--spacer--md);
+  --pf-c-action-list--m-group--Gap: var(--pf-global--spacer--2xl);
+  --pf-c-action-list--m-group--m-vertical--Gap: var(--pf-global--spacer--lg);
+  --pf-c-action-list--m-vertical--Gap: var(--pf-global--spacer--sm);
+
+  &.pf-m-group {
+    gap: var(--pf-c-action-list--m-group--Gap);
```
mcoker, line 8: This comes with the caveat that it doesn't work in safari, and we need to support the latest version of Safari. I wonder if we should build it so it works cross-browser, but with vars that will easily update once gap is supported.

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,28 @@
+.pf-c-action-list {
+  --pf-c-action-list--Gap: var(--pf-global--spacer--md);
+  --pf-c-action-list--m-group--Gap: var(--pf-global--spacer--2xl);
+  --pf-c-action-list--m-group--m-vertical--Gap: var(--pf-global--spacer--lg);
+  --pf-c-action-list--m-vertical--Gap: var(--pf-global--spacer--sm);
+
+  &.pf-m-group {
```
mcoker, line 7: This requires all children of the action list to be groups. What if you want to have `action-list -> item + item + group + item`, with item spacing between items, and group spacing around groups?

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,42 @@
+.pf-c-action-list {
+  --pf-c-action-list--spacer: var(--pf-global--spacer--md);
+  --pf-c-action-list__group-group--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
+  --pf-c-action-list__group-group--m-vertical--Gap: var(--pf-global--spacer--lg);
```
mcoker, line 4: should this be called spacer too?

src/patternfly/components/ActionList/examples/ActionList.md
```diff
@@ -0,0 +1,235 @@
...
+        Next
+      {{/button}}
+    {{/action-list-item}}
+    {{#> action-list-item}}
+      {{#> button button--modifier="pf-m-secondary"}}
+        Back
+      {{/button}}
+    {{/action-list-item}}
+  {{/action-list-group}}
+  {{#> action-list-group action-list-group--modifier="pf-m-vertical"}}
```
mcoker, line 192: does this need `pf-m-vertical`? It only has one child.

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,42 @@
...
+  --pf-c-action-list--m-vertical--spacer: var(--pf-global--spacer--sm);
+
+  &.pf-m-icons {
```
mcoker, line 7: Seems like you should be able to apply this to groups, too?

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,42 @@
...
+    --pf-c-action-list--spacer: 0;
+  }
+}
+
+.pf-c-action-list,
+.pf-c-action-list__group {
+  display: flex;
+
+  > *:not(:last-child) {
```
mcoker, line 16: Can you nest an action-list in another action list? Seems like you wouldn't, you would just nest groups?

mcoker:
@christiemolloy I'm wondering if we need the vertical variation at all. As far as I'm aware, we don't have vertical lists of actions. The designs with stacking buttons are of the action list items wrapping to new lines, which I don't think we need to support for now until we can safely use `gap` and `margin-trim` since we decided we would just build a basic implementation of the action list that isn't responsive until we support those properties later. Until then, to avoid wrapping, a user could use an overflow menu or a flex layout that allows them to use the flex modifiers to switch to a column layout, change alignment, etc.

The benefit of the vertical layout would be that, with responsive modifiers that allow a user to specify a vertical or row layout at specified breakpoints, a user could force a "wrapping" layout at a breakpoint as opposed to waiting for items to wrap on their own. Though we could always add that later if we see it as something that's needed. It's a lot easier to add the things we need later than to remove things we don't need after they're implemented.

WDYT, is there value in offering both a horizontal and vertical layout? also curious what you think @matthewcarleton @mcarrano

matthewcarleton:
ya that sounds right to me @mcoker 
The long term plan of supporting `gap` covers the use case I saw for vertical stacking.

src/patternfly/components/ActionList/action-list.scss
christiemolloy: yep you would only nest groups, moving this outside of this double selector

christiemolloy: yep just groups, do you suggest changing this

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  &.pf-m-icons {
+    --pf-c-action-list--spacer: 0;
+  }
+
+  + .pf-c-action-list__group,
+  + .pf-c-action-list__item {
+    margin-left: var(--pf-c-action-list__group-group--spacer);
+  }
```
mattnolting, line 21:
```suggestion
  + .pf-c-action-list__group {
    margin-left: var(--pf-c-action-list__group--group--spacer);
  }
  + .pf-c-action-list__item {
    margin-left: var(--pf-c-action-list__group--item--spacer);
  }
```

mattnolting: I imagine we'll want these to be separated.

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -0,0 +1,23 @@
+.pf-c-action-list {
+  --pf-c-action-list--spacer: var(--pf-global--spacer--md);
+  --pf-c-action-list__group-group--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
```
mattnolting, line 3:
```suggestion
  --pf-c-action-list__group--group--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
  --pf-c-action-list__group--item--spacer: calc(var(--pf-global--spacer--2xl) - var(--pf-global--spacer--md));
```

src/patternfly/components/ActionList/examples/ActionList.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Overview
+
+### Usage
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-c-action-list` | `<div>` | Initiates the action list container. |
+| `.pf-c-action-list__item` | `<div>` | Initiates the action list item container. |
+| `.pf-c-action-list__group` | `<div>` | Initiates the action list group container. |
+| `.pf-m-icons` | `.pf-c-action-list` | Modifies the action list to support button icons. |
```
mcoker, line 138: I wonder if this should just be `pf-m-icon` and also apply to groups? That would match toolbar where we have an "icon group"

mcoker (approved):
LPTM!

mattnolting (approved):
LPTM 🎉

PR #3604, opened 2020-10-19
feat(touchspin): add component

### LINKED ISSUE #362: Touchspin
http://www.patternfly.org/pattern-library/widgets/#bootstrap-touchspin

christiemolloy:
closes #362

### DIFF
--- a/src/patternfly/components/Touchspin/examples/Touchspin.md
+++ b/src/patternfly/components/Touchspin/examples/Touchspin.md
@@ -1,0 +1,275 @@
+---
+id: Touchspin
+section: components
+cssPrefix: pf-c-touchspin
+beta: true
+---
+
+## Examples
+### Default
+```hbs
+{{#> touchspin}}
+  {{#> touchspin-main}}
+    {{#> input-group}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+      {{#> form-control controlType="input" input="true" form-control--attribute='type="text" value="90" id="textInput1" name="textInput1" aria-label="Input touchspin"'}}
+      {{/form-control}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control"}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-plus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+    {{/input-group}}
+  {{/touchspin-main}}
+{{/touchspin}}
+```
+
+### Active touchspin
+```hbs
+{{#> touchspin}}
+  {{#> touchspin-main touchspin-main--modifier="pf-m-active"}}
+    {{#> input-group}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+      {{#> form-control controlType="input" input="true" form-control--attribute='type="text" value="91" id="textInput2" name="textInput2" aria-label="Input touchspin"'}}
+      {{/form-control}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control pf-m-expanded"}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-plus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+    {{/input-group}}
+  {{/touchspin-main}}
+{{/touchspin}}
+```
+
+### With units and editing text
+```hbs
+{{#> touchspin}}
+  {{#> touchspin-main}}
+    {{#> input-group}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+      {{#> form-control controlType="input" input="true" form-control--attribute='type="text" value="90" id="textInput3" name="textInput3" aria-label="Input touchspin"'}}
+      {{/form-control}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control"}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-plus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+    {{/input-group}}
+  {{/touchspin-main}}
+  {{#> touchspin-units}}%{{/touchspin-units}}
+{{/touchspin}}
+<br/>
+{{#> touchspin}}
+  {{#> touchspin-main touchspin-main--modifier="pf-m-active"}}
+    {{#> input-group}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+      {{#> form-control form-control--modifier="pf-m-expanded" controlType="input" input="true" form-control--attribute='type="text" value="91" id="textInput4" name="textInput4" aria-label="Input touchspin"'}}
+      {{/form-control}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control"}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-plus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+    {{/input-group}}
+  {{/touchspin-main}}
+  {{#> touchspin-units}}%{{/touchspin-units}}
+{{/touchspin}}
+<br/>
+<br/>
+{{#> touchspin}}
+  {{#> touchspin-units}}${{/touchspin-units}}
+  {{#> touchspin-main}}
+    {{#> input-group}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+      {{#> form-control controlType="input" input="true" form-control--attribute='type="text" value="1.00" id="textInput5" name="textInput5" aria-label="Input touchspin"'}}
+      {{/form-control}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control"}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-plus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+    {{/input-group}}
+  {{/touchspin-main}}
+{{/touchspin}}
+<br/>
+{{#> touchspin}}
+  {{#> touchspin-units}}${{/touchspin-units}}
+  {{#> touchspin-main touchspin-main--modifier="pf-m-active"}}
+    {{#> input-group}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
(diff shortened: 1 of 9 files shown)


maryshak1996 (approved):

mcarrano (commented):
Looks good @christiemolloy . Just one question - should the input field be set to only take numeric characters or is this something that would be handled in the React implementation?

mattnolting (commented):
What do you think about adding a `width` value to `.pf-c-touchspin .pf-c-form-control` that's set to `auto` by default, that would allow React to optionally and dynamically update the input's width?

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+  --pf-c-touchspin--c-form-control--Color: var(--pf-global--disabled-color--100);
+  --pf-c-touchspin--c-form-control--MaxWidth: 6ch;
+
+  display: flex;
+  align-items: center;
+
+  .pf-c-form-control {
+    display: inline;
+    max-width: var(--pf-c-touchspin--c-form-control--MaxWidth);
+    color: var(--pf-c-touchspin--c-form-control--Color);
```
mcoker, line 25: I don't think you need this. Looking at the designs, the grey text looks to be placeholder text. When the input has a value, it's black text like a regular input.

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+  &.pf-m-active {
+    --pf-c-touchspin--c-form-control--Color: var(--pf-c-touchspin__main--m-active--c-form-control--Color);
+  }
+}
+
+.pf-c-touchspin__main + .pf-c-touchspin__units {
+  margin-left: var(--pf-c-touchspin__main__units--MarginLeft);
+}
+
+.pf-c-touchspin__units + .pf-c-touchspin__main {
```
mcoker, line 40: You can combine this with the previous selector like

```
.pf-c-touchspin__main + .pf-c-touchspin__units,
.pf-c-touchspin__units + .pf-c-touchspin__main
```

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+  margin-left: var(--pf-c-touchspin__main__units--MarginLeft);
```
mcoker, line 41:
```suggestion
  margin-left: var(--pf-c-touchspin__main--units--MarginLeft);
```

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+
+.pf-c-touchspin__control {
+  &.pf-m-threshold {
+    .pf-c-button {
+      color: var(--pf-c-touchspin__control--m-threshold--c-button--Color);
+      pointer-events: none;
+    }
+  }
+
+  .pf-c-touchspin__icon {
```
mcoker, line 52: You don't need to nest this in `__control`

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+  margin-left: var(--pf-c-touchspin__main__units--MarginLeft);
+}
+
+.pf-c-touchspin__units + .pf-c-touchspin__main {
+  margin-left: var(--pf-c-touchspin__main__units--MarginLeft);
+}
+
+.pf-c-touchspin__control {
+  &.pf-m-threshold {
+    .pf-c-button {
```
mcoker, line 46: can you just disable the button? React can do that once the threshold is met, I don't think we need a class for it.

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+
+  .pf-c-form-control {
+    display: inline;
+    max-width: var(--pf-c-touchspin--c-form-control--MaxWidth);
+    color: var(--pf-c-touchspin--c-form-control--Color);
+    text-align: right;
+  }
+}
+
+.pf-c-touchspin__main {
```
mcoker, line 30: Do you anticipate needing the `__main` element? I was assuming the grey/black text change would come from placholder/regular value text, and even if we kept the `pf-m-active` modifier, it could apply to the parent `pf-c-touchspin` component class, too.

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,55 @@
...
+  --pf-c-touchspin__control--m-threshold--c-button--Color: var(--pf-global--disabled-color--100);
+
+  // icon
+  --pf-c-touchspin__control-icon--FontSize: var(--pf-global--FontSize--xs);
+
+  // form control
+  --pf-c-touchspin--c-form-control--Color: var(--pf-global--disabled-color--100);
+  --pf-c-touchspin--c-form-control--MaxWidth: 6ch;
+
+  display: flex;
```
mcoker, line 19: should this be inline?

```suggestion
  display: inline-flex;
```

src/patternfly/components/Touchspin/examples/Touchspin.md
```diff
@@ -0,0 +1,275 @@
...
+        {{#> button button--modifier="pf-m-control" button--attribute='id="textAreaButton1"'}}
+          {{#> touchspin-icon}}
+            <i class="fas fa-minus" aria-hidden="true"></i>
+          {{/touchspin-icon}}
+        {{/button}}
+      {{/touchspin-control}}
+      {{#> form-control controlType="input" input="true" form-control--attribute='type="text" value="91" id="textInput2" name="textInput2" aria-label="Input touchspin"'}}
+      {{/form-control}}
+      {{#> touchspin-control}}
+        {{#> button button--modifier="pf-m-control pf-m-expanded"}}
```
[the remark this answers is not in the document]
mcoker: From looking at the mockups, I don't think we need expanded on the form control or buttons. That styling would just come from hover/focus states. I think that would be the most predictable since that's how the button/form controls work in other contexts.

src/patternfly/components/Touchspin/touchspin.scss
christiemolloy: @maryshak1996 could you take a look at this. From the designs my understanding was that the text color is light grey when the touchspin controls arent active / user isnt editing the text. Is the grey to represent placeholder text before the user inputs something or changes the control?

src/patternfly/components/Touchspin/examples/Touchspin.md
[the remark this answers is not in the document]
christiemolloy: Okay that sounds good to me

src/patternfly/components/Touchspin/touchspin.scss
christiemolloy: the reason i added the modifier class was because adding disabled on the button gives it a background color, and removes all the borders from the button. I thought this might be leaner instead of adding back styles, but let me know what you think!

mcoker: Ah!  I'm not sure why we would need to deviate from the default design here since the button is effectively disabled. I would be in favor of just letting the button be disabled so it's consistent with disabled buttons in the rest of patternfly.

mcoker: @mcarrano @maryshak1996 wdyt? Do the disabled form control buttons in the touchspin need to be presented differently than regular disabled form control buttons?

@mcarrano also made the comment that he could also see plain buttons used in this component, which wouldn't apply a background when disabled.

christiemolloy: This is how it would look, what do you think @maryshak1996 @mcarrano 

Screen Shot 2020-10-20 at 5 21 01 PM

mcarrano: I like it @christiemolloy . I think this makes it very clear that the buttons are disabled and is consistent with how we handle disabled button states elsewhere.

mcarrano: Do you also think we need an example with the entire control disabled (i.e. buttons and the field)?

christiemolloy: yep good idea ill add that!

src/patternfly/components/Touchspin/touchspin-units.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-touchspin__units{{#if touchspin-units--modifier}} {{touchspin-units--modifier}}{{/if}}"
```
mcoker, line 1: should this just be `__unit` or does it take multiple units?

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,43 @@
...
+  --pf-c-touchspin__main--c-input-group--MarginLeft: var(--pf-global--spacer--sm);
+
+  // control
+  --pf-c-touchspin__control--m-threshold--c-button--Color: var(--pf-global--disabled-color--100);
+
+  // icon
+  --pf-c-touchspin__icon--FontSize: var(--pf-global--FontSize--xs);
+
+  // form control
+  --pf-c-touchspin--c-form-control--MaxWidth: #{pf-size-prem(70px)};
```
mcoker, line 15: wdyt about using `size` instead of CSS `max-width`, and showing examples with `size` updated depending on the number of characters in the input - implying react would update it on the fly as the value changes? React could also have a `size` prop that sets it to something custom as a way to manage the width.

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,43 @@
...
+    text-align: right;
+  }
+}
+
+.pf-c-input-group + .pf-c-touchspin__units,
+.pf-c-touchspin__units + .pf-c-input-group {
+  margin-left: var(--pf-c-touchspin__main--c-input-group--MarginLeft);
+}
+
+.pf-c-touchspin__control {
```
mcoker, line 32: I don't think you need this block anymore? Also do you think we need the `__control` element? If not, I think we can set the font-size on the button as opposed to the icon.

src/patternfly/components/Touchspin/touchspin.scss
```diff
@@ -0,0 +1,43 @@
...
+  --pf-c-touchspin__icon--FontSize: var(--pf-global--FontSize--xs);
+
+  // form control
+  --pf-c-touchspin--c-form-control--MaxWidth: #{pf-size-prem(70px)};
+
+  display: inline-flex;
+  align-items: center;
+
+  .pf-c-form-control {
+    display: inline;
```
mcoker, line 21: is this needed?

src/patternfly/components/Touchspin/touchspin.scss
christiemolloy: I removed this block. I was thinking about still keeping the icon class + setting the font size there in case we make future changes ?

mcoker: That works. Typically I think we would apply font-size to the button to control the size of the thing in it, but the way you've done it works, too. I agree with keeping the icon element/class.

src/patternfly/components/Touchspin/touchspin.scss
christiemolloy: yes otherwise it fills the whole width

src/patternfly/components/Touchspin/touchspin.scss
christiemolloy: could you show me an example of using size. is it a property ? or a concept for the var name ?

mcoker: https://codepen.io/mcoker/pen/WNxoxeq

src/patternfly/components/Touchspin/examples/Touchspin.md
```diff
@@ -0,0 +1,140 @@
+---
+id: Touch spin
+section: components
+cssPrefix: pf-c-touchspin
```
mcoker, line 4: Is this called "Touchspin" or "Touch spin"? If it's two words, I think there should be a hyphen in the component name.

src/patternfly/components/Touchspin/examples/Touchspin.md
```diff
@@ -0,0 +1,140 @@
...
+    {{#> button button--modifier="pf-m-control" button--attribute='id="controlButton10" aria-label="Plus"'}}
+      {{#> touchspin-icon}}
+        <i class="fas fa-plus" aria-hidden="true"></i>
+      {{/touchspin-icon}}
+    {{/button}}
+  {{/input-group}}
+{{/touchspin}}
+```
+
+### With unit and left threshold reached
```
mcoker, line 69:
```suggestion
### With unit and lower threshold reached
```

src/patternfly/components/Touchspin/examples/Touchspin.md
```diff
@@ -0,0 +1,140 @@
...
+      {{#> touchspin-icon}}
+        <i class="fas fa-plus" aria-hidden="true"></i>
+      {{/touchspin-icon}}
+    {{/button}}
+  {{/input-group}}
+  {{#> touchspin-unit}}%{{/touchspin-unit}}
+{{/touchspin}}
+```
+
+### With unit and right threshold reached
```
mcoker, line 90:
```suggestion
### With unit and upper threshold reached
```

src/patternfly/components/Touchspin/examples/Touchspin.md
mcoker: Confirmed with @mcarrano it's "Touchspin"

src/patternfly/components/Touchspin/examples/Touchspin.md
```diff
@@ -0,0 +1,140 @@
+---
+id: Touch spin
```
mcoker, line 2:
```suggestion
id: Touchspin
```

src/patternfly/components/Touchspin/examples/Touchspin.md
christiemolloy: yep thanks for clarifying that was a mistake

Issue #3556, opened 2020-10-05
Dual list selector component


mcarrano:
Introduce a new dual list selector component per the designs here: https://marvelapp.com/prototype/12e47bhj

mcarrano:
I think that it would also make sense to consider the ability to drag and drop between lists as well as use the buttons. @mmenestr @mcoker what do you think? We did not consider this as part of the original design but we could leverage approach for data list. What are your thoughts? This can always be considered as an add-on.

mcoker:
@mcarrano sounds good to me!

mmenestr:
Yes that seems fine to me! And just to go deeper in the specifications, let's say you have 2 items selected, and you try to click and drag - do both items get moved over, or does only the one item your cursor is over get dragged?

mcoker:
cc @kmcfaul what do you think about @mmenestr's question above re: dragging more than one item at a time?

PR #3605, opened 2020-10-20
feat(dual-list-selector): added dual list selector component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3556

In the interest of time and a consideration on how the list is structured, I've only created the basic desktop version for now. In the next milestone we can add responsive design w/ breakpoint modifiers and variations of the menu/list types.

### DIFF
--- a/src/patternfly/components/DualListSelector/dual-list-selector-controls-item.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-controls-item.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-dual-list-selector__controls-item{{#if dual-list-selector-controls-item--modifier}} {{dual-list-selector-controls-item--modifier}}{{/if}}"
+  {{#if dual-list-selector-controls-item--attribute}}
+    {{dual-list-selector-controls-item--attribute}}
+  {{/if}}>
+  {{>@partial-block}}
+</div>
--- a/src/patternfly/components/DualListSelector/dual-list-selector-header.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-header.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-dual-list-selector__header{{#if dual-list-selector-header--modifier}} {{dual-list-selector-header--modifier}}{{/if}}"
+  {{#if dual-list-selector-header--attribute}}
+    {{dual-list-selector-header--attribute}}
+  {{/if}}>
+  {{>@partial-block}}
+</div>
--- a/src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
@@ -1,6 +1,6 @@
-<div class="pf-c-dual-list-selector__status-text{{#if dual-list-selector-status-text--modifier}} {{dual-list-selector-status-text--modifier}}{{/if}}"
+<span class="pf-c-dual-list-selector__status-text{{#if dual-list-selector-status-text--modifier}} {{dual-list-selector-status-text--modifier}}{{/if}}"
   {{#if dual-list-selector-status-text--attribute}}
     {{dual-list-selector-status-text--attribute}}
   {{/if}}>
   {{>@partial-block}}
-</div>
+</span>
--- a/src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-dual-list-selector__status-text{{#if dual-list-selector-status-text--modifier}} {{dual-list-selector-status-text--modifier}}{{/if}}"
+  {{#if dual-list-selector-status-text--attribute}}
+    {{dual-list-selector-status-text--attribute}}
+  {{/if}}>
+  {{>@partial-block}}
+</div>
--- a/src/patternfly/components/DualListSelector/dual-list-selector-tools.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-tools.hbs
@@ -1,0 +1,6 @@
+<div class="pf-c-dual-list-selector__tools{{#if dual-list-selector-tools--modifier}} {{dual-list-selector-tools--modifier}}{{/if}}"
+  {{#if dual-list-selector-tools--attribute}}
+    {{dual-list-selector-tools--attribute}}
+  {{/if}}>
+  {{>@partial-block}}
+</div>
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -104,11 +104,13 @@
 }
 
 .pf-c-dual-list-selector__status {
+  display: flex;
   grid-area: var(--pf-c-dual-list-selector__status--GridArea);
   margin-bottom: var(--pf-c-dual-list-selector__status--MarginBottom);
 }
 
 .pf-c-dual-list-selector__status-text {
+  flex-grow: 1;
   font-size: var(--pf-c-dual-list-selector__status-text--FontSize);
   color: var(--pf-c-dual-list-selector__status-text--Color);
 }
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -23,7 +23,8 @@
   --pf-c-dual-list-selector__tools-filter--tools-actions--MarginLeft: var(--pf-global--spacer--sm);
 
   // Menu
-  --pf-c-dual-list-selector__menu--BoxShadow: var(--pf-global--BoxShadow--md);
+  --pf-c-dual-list-selector__menu--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-dual-list-selector__menu--BorderColor: var(--pf-global--BorderColor--100);
   --pf-c-dual-list-selector__menu--MinHeight: #{pf-size-prem(200px)};
   --pf-c-dual-list-selector__menu--MaxHeight: #{pf-size-prem(320px)};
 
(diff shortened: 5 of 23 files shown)


mcarrano (commented):
The basic layout looks great @mcoker . Just a couple of comments/questions:

* regarding the example naming, what if we title the lefthand list "Available options" and the righthand list "Chosen options." You can then have "Available item selected" and "Chosen item selected." What do you think @mmenestr ?

* What if I want to move multiple items? Should we support checkbox selection in both lists? I know there was also discussion about allowing the list to be a hierarchical tree. Should we show that?

mmenestr (approved):

src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-dual-list-selector__status-text{{#if dual-list-selector-status-text--modifier}} {{dual-list-selector-status-text--modifier}}{{/if}}"
```
mattnolting, line 1:
```suggestion
<span class="pf-c-dual-list-selector__status-text{{#if dual-list-selector-status-text--modifier}} {{dual-list-selector-status-text--modifier}}{{/if}}"
```

mcoker: I added this text element anticipating there might be icons, buttons, etc that get added here. Do you think this should be a span, or should the `__status` parent be a flex layout, so we can easily place things before/after the text?

mattnolting:
Looks like the responsive layout needs some updates:
Screen Shot 2020-10-20 at 2 37 23 PM

mcoker:
@mattnolting yeah I'm holding off on the responsive breakpoints until next sprint, just so we can focus on getting this in. If there's time, we could add them this sprint.

src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
mattnolting: I would suggest both `__status { display: flex }` to support things before/after and `span.pf-c-dual-list-selector__status-text` to be consistent with other `__text` elements. Although, we could update `__status { display: flex }` when that becomes necessary.

mattnolting:
> @mattnolting yeah I'm holding off on the responsive breakpoints until next sprint, just so we can focus on getting this in. If there's time, we could add them this sprint.

My fault, I didn't read your description :|

src/patternfly/components/DualListSelector/dual-list-selector-status-text.hbs
mcoker: I'll go ahead and add `flex` since doing that after the fact is a breaking change since it will change the basic layout of all children of the `__status` element.

src/patternfly/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -0,0 +1,681 @@
...
+| `.pf-c-dual-list-selector__status-text` | `<span>` | Initiates a dual list selector pane selected status text. **Required** |
+| `.pf-c-dual-list-selector__menu` | `<div>` | Initiates a dual list selector pane menu container. **Required** |
+| `.pf-c-dual-list-selector__list` | `<ul>` | Initiates a dual list selector pane menu list. **Required** |
+| `.pf-c-dual-list-selector__list-item` | `<li>` | Initiates a dual list selector pane menu list item. **Required** |
+| `.pf-c-dual-list-selector__item` | `<button>`, `<div>` | Initiates a dual list selector pane menu item. **Required** |
+| `.pf-c-dual-list-selector__item-main` | `<span>` | Initiates a dual list selector pane menu item main container. **Required** |
+| `.pf-c-dual-list-selector__item-text` | `<span>` | Initiates a dual list selector pane menu item text. **Required** |
+| `.pf-c-dual-list-selector__controls` | `<div>` | Initiates the dual list selector controls. **Required** |
+| `.pf-c-dual-list-selector__controls-item` | `<div>` | Initiates the dual list selector controls item. **Required** |
+| `.pf-m-available` | `.pf-c-dual-list-selector__pane` | Defines a pain as the available list. |
```
mattnolting, line 679:
```suggestion
| `.pf-m-available` | `.pf-c-dual-list-selector__pane` | Defines a pane as the available list. |
```

src/patternfly/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -0,0 +1,681 @@
...
+| `.pf-m-chosen` | `.pf-c-dual-list-selector__pane` | Defines a pain as the chosen list. |
```
mattnolting, line 680:
```suggestion
| `.pf-m-chosen` | `.pf-c-dual-list-selector__pane` | Defines a pane as the chosen list. |
```

mattnolting (commented):
Very nice!

mceledonia:
One thing I missed when reviewing this before is the shadows on the boxes. Switching them to a border would be more consistent with our new shadows approach, since these aren't displaying on top of any other content.

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -0,0 +1,173 @@
...
+    "pane-menu controls pane-menu-c";
+  grid-template-columns:
+    minmax(var(--pf-c-dual-list-selector--GridTemplateColumns--pane--MinMax--min), var(--pf-c-dual-list-selector--GridTemplateColumns--pane--MinMax--max))
+    min-content
+    minmax(var(--pf-c-dual-list-selector--GridTemplateColumns--pane--MinMax--min), var(--pf-c-dual-list-selector--GridTemplateColumns--pane--MinMax--max));
+  grid-template-rows: repeat(3, auto) auto;
+}
+
+.pf-c-dual-list-selector__pane {
+  display: contents;
```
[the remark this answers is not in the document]
mcoker: Haha I'm pretty proud of that 🥳 It was the best way I found to align the controls with the list boxes.

Though since it's a little weird and we can't apply styles directly to `__pane` now, we could also make the list heights a static thing assignable via a variable, and make the controls element that height by sharing the variable, and bottom align the controls in the overall layout, then center align the buttons in the control element. You lose the fluid height adjustment of the lists, but I think that would be OK. WDYT?

mattnolting (approved):
Perfecto 🎉

christiemolloy (changes requested):
Screen Shot 2020-10-21 at 10 39 17 AM
 looks like the menu container should have a box-shadow

font-size of the menu items should be 14px (based on Marvel)

Screen Shot 2020-10-21 at 10 42 16 AM
Filter options should have a search icon

src/patternfly/components/DualListSelector/dual-list-selector-tools.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-dual-list-selector__tools{{#if dual-list-selector-tools--modifier}} {{dual-list-selector-tools--modifier}}{{/if}}"
```
christiemolloy, line 1: should this be `__toolbar`

christiemolloy (changes requested):

src/patternfly/components/DualListSelector/dual-list-selector-header.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-dual-list-selector__header{{#if dual-list-selector-header--modifier}} {{dual-list-selector-header--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
mcoker: `__header` in case we add something like a description, close button, etc (more than just the title).

`__title` to establish the title section, with the `__title-text` to designate the text - which allows us to add an icon or action or something else in `__title` that's adjacent to `__title-text`.

Seems like a repeatable structure we have in a lot of components.

src/patternfly/components/DualListSelector/dual-list-selector-controls-item.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-dual-list-selector__controls-item{{#if dual-list-selector-controls-item--modifier}} {{dual-list-selector-controls-item--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
mcoker: `item` is typically what we call a generic child of some parent like this, and helps differentiate between `__controls` and `__control` if we went with that naming since they're so similar. We use `control`/`controls` in other places - controls is what we used in code editor https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/CodeEditor/code-editor-controls.hbs

src/patternfly/components/DualListSelector/dual-list-selector-tools.hbs
mcoker: I don't think so, since toolbar is a component, and this isn't for a toolbar. `<Page>` uses `header-tools` in a similar way.

mcoker:
@christiemolloy updated! Matt/Michael said yesterday the box shadow should be a border.

mmenestr:
Just to pitch into the box convo, since I was originally using menu items in my dual list mocks, I used the Menu background component in sketch to have a base. So I'm not really tied to the shadow vs. a border, I think how it is right now is totally fine!

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -0,0 +1,178 @@
...
+  --pf-c-dual-list-selector__title-text--FontWeight: var(--pf-global--FontWeight--bold);
+
+  // Tools
+  --pf-c-dual-list-selector__tools--MarginBottom: var(--pf-global--spacer--md);
+  --pf-c-dual-list-selector__tools-filter--tools-actions--MarginLeft: var(--pf-global--spacer--sm);
+
+  // Menu
+  --pf-c-dual-list-selector__menu--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-dual-list-selector__menu--BorderColor: var(--pf-global--BorderColor--100);
+  --pf-c-dual-list-selector__menu--MinHeight: #{pf-size-prem(200px)};
```
christiemolloy, line 28: should this always match the value of this: `--pf-c-dual-list-selector--GridTemplateColumns--pane--MinMax--min`

christiemolloy (commented):
this looks really good - left one comment

christiemolloy (approved):
NICEEEEE

Issue #3612, opened 2020-10-22
Update description list auto-fit var names


mcoker:
These should have `--m-auto-fit` in the var name

https://github.com/patternfly/patternfly/blob/18605332688fdcf443460f090e5f392b979bdb1b/src/patternfly/components/DescriptionList/description-list.scss#L31-L32

PR #3613, opened 2020-10-22
fix(description-list): updated auto-fit var names

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3612


mattnolting (approved):
Great catch! LPTM!

PR #3621, opened 2020-10-23
chore(drawer): update demos, remove notification drawer demos

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3620

PR #3622, opened 2020-10-23
chore(notification-drawer): updated demos so badge reflects drawer items

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3371

@mceledonia @mcarrano this adds a read notification drawer demo, attention demo, and updates the other demo notification badges to unread. WDYT, does this look like you would expect? Also in the drawer demos, since there is no notification drawer in the drawer, should we update the header icon that toggles the drawer open/closed to be something other than the bell icon?

### DIFF
--- a/src/patternfly/components/NotificationDrawer/notification-drawer-basic-list.hbs
+++ b/src/patternfly/components/NotificationDrawer/notification-drawer-basic-list.hbs
@@ -1,5 +1,5 @@
 {{#> notification-drawer-list}}
-  {{#> notification-drawer-list-item notification-drawer-list-item--IsInfo="true"}}
+  {{#> notification-drawer-list-item notification-drawer-list-item--IsInfo="true" notification-drawer-list-item--IsRead=notification-drawer-basic-list--AllRead}}
     {{#> notification-drawer-list-item-header}}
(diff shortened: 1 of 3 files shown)


mcarrano (commented):
@mcoker I think that the Notification drawer demos look good. @mceledonia what do you think?

For the drawer demos, I agree that including the notification badge (bell) does not make sense if this is not a notification drawer demonstration. Seems like this could be pared done to just showing a horizontal and vertical drawer. Not sure if we even need a masthead icon, but what about attaching this to the Settings (gear) icon. I could see possibly putting settings in a drawer.

mcoker:
> Seems like this could be pared done to just showing a horizontal and vertical drawer. Not sure if we even need a masthead icon, but what about attaching this to the Settings (gear) icon. I could see possibly putting settings in a drawer.

@mcarrano made that update in the drawer demo PR - https://github.com/patternfly/patternfly/pull/3621

https://patternfly-pr-3621.surge.sh/components/drawer/html-demos

mcarrano (approved):
Looks good!

christiemolloy (approved):
nice!

PR #3629, opened 2020-10-29
Fix add gap to toolbar

### LINKED ISSUE #3348: Add GAP to toolbar!

mattnolting:
https://github.com/patternfly/patternfly/pull/3543

closes #3348 #3344 

@mcarrano @mceledonia I set `row-gap` for direct children of `content-sections` to `spacer--md`. For items within groups, I set the `row-gap` to `spacer--xs`. Wdyt?

Also, I set toggle groups to **nowrap**, as they are a special use case and imo should never wrap as their behavior has a specific functionality.

Screen Shot 2020-10-01 at 3 10 43 PM

Screen Shot 2020-10-01 at 3 11 44 PM

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -13,6 +13,7 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   // Content
   --pf-c-toolbar__content--PaddingRight: var(--pf-global--spacer--md); // remove at breaking change
   --pf-c-toolbar__content--PaddingLeft: var(--pf-global--spacer--md); // remove at breaking change
+  --pf-c-toolbar__content--m-chip-container--RowGap: var(--pf-global--spacer--md);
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -173,6 +178,11 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   &.pf-m-toggle-group {
     --pf-c-toolbar--spacer: var(--pf-c-toolbar__group--m-toggle-group--spacer);
 
+    &,
+    .pf-c-toolbar__group {
+      flex-wrap: nowrap;
```
mcoker, line 183: Is this necessary? Doesn't look like they're set to wrap to begin with.

mattnolting: Good catch!

mcoker (approved):
Very nice!

christiemolloy (approved):
nice!

PR #3632, opened 2020-10-30
feat(jumplinks): add subsections

### LINKED ISSUE #3616: Jump links - add subsections

christiemolloy:
closes #3616

### DIFF
--- a/src/patternfly/components/JumpLinks/jump-links.scss
+++ b/src/patternfly/components/JumpLinks/jump-links.scss
@@ -23,6 +23,7 @@
   --pf-c-jump-links__link--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-jump-links__link--PaddingBottom: var(--pf-global--spacer--md);
   --pf-c-jump-links__link--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-jump-links--c-jump-links--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-jump-links__link--OutlineOffset: calc(-1 * var(--pf-global--spacer--sm));
 
   // before
@@ -44,6 +45,10 @@
 
   // text
   --pf-c-jump-links__link-text--Color: var(--pf-global--Color--200);
+  --pf-c-jump-links__list--list-link-text--Color: var(--pf-global--Color--200);
+  --pf-c-jump-links__list--list-link--hover__link-text--Color: var(--pf-global--Color--100);
+  --pf-c-jump-links__list--list-link--focus__link-text--Color: var(--pf-global--Color--100);
+  --pf-c-jump-links__list--list-item--m-current__link-text--Color: var(--pf-global--Color--100);
   --pf-c-jump-links__link--hover__link-text--Color: var(--pf-global--Color--100);
   --pf-c-jump-links__link--focus__link-text--Color: var(--pf-global--Color--100);
   --pf-c-jump-links__item--m-current__link-text--Color: var(--pf-global--Color--100);
@@ -98,6 +103,20 @@
     border: solid var(--pf-c-jump-links__list--before--BorderColor);
     border-width: var(--pf-c-jump-links__list--before--BorderTopWidth) var(--pf-c-jump-links__list--before--BorderRightWidth) var(--pf-c-jump-links__list--before--BorderBottomWidth) var(--pf-c-jump-links__list--before--BorderLeftWidth);
   }
+
+  .pf-c-jump-links__list {
+    --pf-c-jump-links__list--PaddingTop: 0;
+    --pf-c-jump-links__list--PaddingBottom: 0;
+    --pf-c-jump-links__link--PaddingLeft: var(--pf-c-jump-links--c-jump-links--PaddingLeft);
+
+    .pf-c-jump-links__link-text {
+      color: var(--pf-c-jump-links__list--list-link-text--Color);
+    }
+
+    .pf-c-jump-links__item.pf-m-current {
+      --pf-c-jump-links__list--list-link-text--Color: var(--pf-c-jump-links__list--list-item--m-current__link-text--Color);
+    }
+  }
 }
 
 .pf-c-jump-links__link {
@@ -113,10 +132,12 @@
 
   &:hover {
     --pf-c-jump-links__link-text--Color: var(--pf-c-jump-links__link--hover__link-text--Color);
+    --pf-c-jump-links__list--list-link-text--Color: var(--pf-c-jump-links__list--list-link--hover__link-text--Color);
   }
 
   &:focus {
     --pf-c-jump-links__link-text--Color: var(--pf-c-jump-links__link--focus__link-text--Color);
+    --pf-c-jump-links__list--list-link-text--Color: var(--pf-c-jump-links__list--list-link--focus__link-text--Color);
     --pf-c-jump-links__link--before--BorderTopWidth: var(--pf-c-jump-links__link--focus--before--BorderTopWidth);
     --pf-c-jump-links__link--before--BorderLeftWidth: var(--pf-c-jump-links__link--focus--before--BorderLeftWidth);
     --pf-c-jump-links__link--before--BorderColor: var(--pf-c-jump-links__link--focus--before--BorderColor);
--- a/src/patternfly/components/JumpLinks/jump-links.scss
+++ b/src/patternfly/components/JumpLinks/jump-links.scss
@@ -108,14 +108,6 @@
     --pf-c-jump-links__list--PaddingTop: 0;
     --pf-c-jump-links__list--PaddingBottom: 0;
     --pf-c-jump-links__link--PaddingLeft: var(--pf-c-jump-links__list--list__link--PaddingLeft);
-
-    .pf-c-jump-links__link-text {
-      color: var(--pf-c-jump-links__list--list-link-text--Color);
-    }
-
-    .pf-c-jump-links__item.pf-m-current {
-      --pf-c-jump-links__list--list-link-text--Color: var(--pf-c-jump-links__list--list-item--m-current__link-text--Color);
-    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


redallen (approved):

mcarrano (commented):
Looks good to me. but I'll defer to @lboehling on spacing issues. @ajacobs21e perhaps you can also take a look.

ajacobs21e:
> Looks good to me. but I'll defer to @lboehling on spacing issues. @ajacobs21e perhaps you can also take a look.

What's the spacing issue?

mcarrano:
No issue @ajacobs21e I just wanted to get another set of eyes on this.

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -44,6 +45,10 @@
 
   // text
   --pf-c-jump-links__link-text--Color: var(--pf-global--Color--200);
+  --pf-c-jump-links__list--list-link-text--Color: var(--pf-global--Color--200);
+  --pf-c-jump-links__list--list-link--hover__link-text--Color: var(--pf-global--Color--100);
+  --pf-c-jump-links__list--list-link--focus__link-text--Color: var(--pf-global--Color--100);
+  --pf-c-jump-links__list--list-item--m-current__link-text--Color: var(--pf-global--Color--100);
```
mcoker, line 51: are these nested link colors needed?

mcoker (commented):
LGTM, just a question about some vars that look like they aren't needed.

mattnolting (approved):
LGTM, after you remove the unused vars!

ajacobs21e:
The spacing for the vertical jump links is different from my original design. I had 8px top and bottom spacers for 16px total spacing between the text of each link. It's double that. @lboehling was that intentional? 
https://www.sketch.com/s/f6a52c25-8657-4680-9d3e-4c09131e6ed0/a/exD82z

I also did not have a top spacer above the subsections to keep it closer to the parent section. I'd like to keep that. 

Here's an example of PFE jump links: https://access.redhat.com/documentation/en-us/red-hat-enterprise-linux/8/guide/8c44950e-29f0-4150-89a5-036f57bf775e
It's a long list and would get longer with 32px between each line of text. It's 12px in that example, it should be 16px though.

lboehling:
> The spacing for the vertical jump links is different from my original design. I had 8px top and bottom spacers for 16px total spacing between the text of each link. It's double that. @lboehling was that intentional?
https://www.sketch.com/s/f6a52c25-8657-4680-9d3e-4c09131e6ed0/a/exD82z

> I also did not have a top spacer above the subsections to keep it closer to the parent section. I'd like to keep that.

Agreed @ajacobs21e, there should be 8px above/below to total 16px between each item. Also agree with removing the top spacer above subsections.

christiemolloy:
Updated @ajacobs21e @lboehling 

Screen Shot 2020-11-10 at 4 41 56 PM

also removed vars @mcoker @mattnolting

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -98,6 +101,18 @@
     border: solid var(--pf-c-jump-links__list--before--BorderColor);
     border-width: var(--pf-c-jump-links__list--before--BorderTopWidth) var(--pf-c-jump-links__list--before--BorderRightWidth) var(--pf-c-jump-links__list--before--BorderBottomWidth) var(--pf-c-jump-links__list--before--BorderLeftWidth);
   }
+
+  .pf-c-jump-links__list {
+    --pf-c-jump-links__list--PaddingTop: 0;
+    --pf-c-jump-links__list--PaddingBottom: 0;
+    --pf-c-jump-links__link--PaddingTop: var(--pf-c-jump-links__list--list__link--PaddingTop);
```
mcoker, line 108: I thought we were only using `--` in `list--list` for a nested item that is a component. eg, `list--c-button`. This conflicts with the syntax we use for adjacent items, for styling `__list + __list` - the var syntax for that would be `list--list`. The same applies if we have an element adjacent to a component, too, eg `__list + c-button`, but I think we compromised on that for readability if I recall. Personally I think we should use `__` everywhere for children, whether it's an element of a component or separate component, so we leave room to style adjacent elements with `--` between the names in a var. wdyt?

mcoker:
With uneven top and bottom padding for the first subsection link, the touch area for the link is different, and the accent is uneven extending below, but not above the item.

Screen Shot 2020-11-10 at 4 28 43 PM

We could extend the blue accent up, but then it overlaps with the touchable area of the previous link

Screen Shot 2020-11-10 at 4 37 29 PM

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -98,6 +101,18 @@
     border: solid var(--pf-c-jump-links__list--before--BorderColor);
     border-width: var(--pf-c-jump-links__list--before--BorderTopWidth) var(--pf-c-jump-links__list--before--BorderRightWidth) var(--pf-c-jump-links__list--before--BorderBottomWidth) var(--pf-c-jump-links__list--before--BorderLeftWidth);
   }
+
+  .pf-c-jump-links__list {
+    --pf-c-jump-links__list--PaddingTop: 0;
```
mcoker, line 106: You could add `--pf-c-jump-links__list--before--BorderColor: transparent;` here to disable border from nested lists, though it isn't necessary since it's the same color and overlaps with the parent's border.

mcoker:
Can you add `pointer-events: none;` to `.pf-c-jump-links__link::before`?

src/patternfly/components/JumpLinks/jump-links.scss
christiemolloy: Ah yes sorry I always get confused with this syntax, agreed changed it to `__list`

christiemolloy:
> With uneven top and bottom padding for the first subsection link, the touch area for the link is different, and the accent is uneven extending below, but not above the item.
> 
> Screen Shot 2020-11-10 at 4 28 43 PM
> 
> We could extend the blue accent up, but then it overlaps with the touchable area of the previous link
> 
> Screen Shot 2020-11-10 at 4 37 29 PM

Added a negative margin top

src/patternfly/components/JumpLinks/jump-links.scss
mcoker: I think it's confusing, too, because we do it different ways. I'll add an agenda item for our next CSS meeting and we can go over whether it still makes sense to use `--c-component` for nested items, or if we can be consistent and use `__`.

mcoker:
Is it possible for the item with the nested sections to be current, and not any of the nested sections? I would expect that to be possible, and was assuming just that item would be marked current. Currently it looks like this if that item is made current, making the item and its children current:

Screen Shot 2020-11-11 at 2 49 27 PM

Also with the negative margin, the buttons now overlap. I'm not sure what the best solution is there. Maybe it's OK that they overlap? @mcarrano do you or anyone on the design side have feedback for how to best lay those out considering the uneven spacing on the first item in a nested section?

mcarrano:
> Is it possible for the item with the nested sections to be current, and not any of the nested sections? I would expect that to be possible, and was assuming just that item would be marked current. 

I don't think so @mcoker . To me this should work more like grouped, expandable navigation. If the parent node is opened, then the first subsection is selected by default. That seems consistent with the designs here: 

Screen Shot 2020-11-11 at 5 45 32 PM

> Also with the negative margin, the buttons now overlap. I'm not sure what the best solution is there. Maybe it's OK that they overlap? @mcarrano do you or anyone on the design side have feedback for how to best lay those out considering the uneven spacing on the first item in a nested section?

Yeah maybe we can live with that. I don't have an easy fix. If I click in that overlap region, do you know what would take precedent? Let's not hold up the works for this. We can always come back and address as a bug fix if needed.

mcoker:
@mcarrano Couldn't you have links like this where you anchor to "h2.link3" but you aren't yet anchored to "h3.sub-link1"? I guess I'm asking if the parent element for a sub-section acts as a label for the subsection (and isn't an anchor itself on the page), or if the parent for a sub-section can also be an anchor itself  that is separate from the first anchor in the sub-section.

I thought we had decided to only draw a blue accent beside the active item, whether it's in a sub-section or not, instead of having the accent span all of the links in a sub-section.

```
link1
link2
link3
    sub-link1
    sub-link2
```

that links to content like

```
h2.link1
content

h2.link2
content

h2.link3
content

h3.sub-link1
content

h3.sub-link2
content
...
```

mcarrano:
I guess you could @mcoker . I have to admit to not being clear about how we expect people to use nested jump links. Whether these are to denote heading levels or just a way to group sets of links. @ajacobs21e @lboehling do you have insight here? (see thread above)

mcoker (approved):
LGTM! We can follow up on the parent of a subsection's active styling in a follow up PR if needed.

ajacobs21e:
@mcoker @mcarrano Here is another use case:
image
Here the subsections are shown all the time and then one can be highlighted at a time.

I think the idea was that if the subsection grouping was collapsed, you would click on or scroll into the parent level to expose it, so we would keep the piece you clicked on highlighted until you click or scroll out of the section. I do not know if there was a specific example for this use case - Sam worked on this part.

If this collapsed / expanded thing is too tricky, I think we could just implement what it's in the image I just posted and maybe revisit the expansion idea.

mcarrano:
Yeah, my concern with the current design is that if you have a collapsed section, there is no affordance that there are subsections like there is for a tree view. I think of this like the TOC we have on our website where h3 headings are shown as subsections.

Screen Shot 2020-11-12 at 1 56 54 PM

Looks like that's the way this is implemented now @mcoker ?  If so, I'm good with it and we can enhance later if needed.

mcoker:
@mcarrano in that TOC would you ever want to make "Examples" or "Documentation" as active/current, without marking the whole menu as current - like the example on the left in @ajacobs21e's comment? You can't currently do that in the component. This is what it looks like if you mark the title for a subsection as active:

Screen Shot 2020-11-16 at 6 25 47 PM

If so, we can update the component to do that instead (you can't make an entire subsection current), or we could update it so that you can make either any individual link current, or an entire section.

Right now the "current" state applies to list items, and the structure is like this:

```
ul
    li <--- this makes this link current
        link - text
    li
        link - text
    li <--- since this is current, everything in this list item is marked current. No way to just make the link in this list item current.
        link - text
        ul
            li
                link - text
            li
                link - text
            li
                link - text
    li
        link - text
```

We can update it so that either a list item or a link can be current. Users would make the link current if they want that individual link to be current. They would make the list item current if they want to make everything in that list item current (including submenus)

```
ul
    li
        link - text <--- this would be current if this item were current
    li
        link - text
    li <--- this would be current if you want the whole subsection to be marked current
        link - text <--- this would be current if you only wanted "text" here to be current
        ul
            li
                link - text
            li
                link - text
            li
                link - text
    li
        link - text
```

mcarrano:
@mcoker I have to admit to being pretty confused about how subsections are intended to be used here. I saw this as just a way to create hierarchy for grouping links that are at the same heading level, i.e., a group of h3 headings that are are of a larger section starting with an h2. I think it's a bit confusing to highlight the entire section that might contain several separate jump links because it still does not tell me exactly where I am on a page (just that I'm within this larger section). But before we change anything, I'd like to hear from @ajacobs21e and @lboehling as they are closer to what the actual use cases are for this. How do we imagine subsections being used? The mocks above seem to illustrate some divergent behaviors.

I will mark the corresponding React issue "on hold" until we work through this.

Issue #3625, opened 2020-10-26
Update implementation for Date Picker


evwilkin:
See previous Core issues https://github.com/patternfly/patternfly/pull/3482 & https://github.com/patternfly/patternfly/issues/3256

See related React issue https://github.com/patternfly/patternfly-react/issues/5057

Date Picker was initially implemented using Flatpickr library, which provided some functionality but also required battling against some of their features to get what we need.

To avoid this moving forward, replace Flatpickr with custom implementation while maintaining features from existing implementation.

PR #3633, opened 2020-10-30
feat(calendar-month): added calendar month component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3625

**https://patternfly-pr-3633.surge.sh/components/calendar-month**


redallen (changes requested):
Not sure which line is causing this -- but there's a bug in an `id`:
```html
<span id="id&#x3D;&quot;calendar-month-basic-month-select&quot;-label" hidden>Choose one</span>
```

Aside from that this has everything I need!

redallen (commented):
This is more of a design question for @mcarrano : how should we display dates that cannot be selected because they are outside of a time range? Flatpickr grayed them out more than days for previous/future months:
image

mcarrano:
@redallen Yes, we probably should be graying those dates out, even though it's not shown that way in the design. @nicolethoen do you agree?

nicolethoen:
@mcarrano @redallen Yeah, there should be something about them that makes them look unavailable like we style other disabled form elements.

redallen (approved):
There's still some bugs in the rendered markup like `id="id&#x3D;&quot;calendar-month-range-month-select&quot;-toggle"` and `id&#x3D;&quot;calendar-month-range-month-select&quot;-label id&#x3D;&quot;calendar-month-range-month-select&quot;-toggle` but this has all the beautiful styles I need for React!

maryshak1996:
@mcoker this looks awesome! the only update that I would like to see from the design is to have the background color on the range go one shade of blue lighter (`--pf-global--palette--blue-50`).
I have a question as well: I'm not sure I'm understanding range unselected vs range selected -- what do you mean by these titles? I would have thought that 'range unselected' would start with the filled blue circle but would have the outline circle while user is hovering over the end date' and that 'range selected' would be the one to have two filled in circles. Maybe I'm misunderstanding what these titles are supposed to convey

mcoker:
@maryshak1996 updated the background color

> Maybe I'm misunderstanding what these titles are supposed to convey

Totally my fault! I had the example names mixed up. I renamed the examples and created another so that we show what range calendars look like when a start date is selected and the end date is unselected, and vice versa.

mcarrano (changes requested):
This looks good @mcoker But I think you have the range examples swapped. I believe the open circle is before the second date gets selected and it fills in after. Can you confirm @maryshak1996 ?

maryshak1996 (approved):

maryshak1996:
> This looks good @mcoker But I think you have the range examples swapped. I believe the open circle is before the second date gets selected and it fills in after. Can you confirm @maryshak1996 ?

Yup :) this is fixed now!

mcarrano (approved):

src/patternfly/components/CalendarMonth/calendar-month-grid-body.hbs
```diff
@@ -0,0 +1,6 @@
+<tbody class="pf-c-calendar-month__grid-body{{#if calendar-month-grid-body--modifier}} {{calendar-month-grid-body--modifier}}{{/if}}"
```
christiemolloy, line 1: should this be called `calendar-month__dates-body`

src/patternfly/components/CalendarMonth/calendar-month-grid-head.hbs
```diff
@@ -0,0 +1,6 @@
+<thead class="pf-c-calendar-month__grid-head{{#if calendar-month-grid-head--modifier}} {{calendar-month-grid-head--modifier}}{{/if}}"
```
christiemolloy, line 1: should this be called `calendar-month__days`

src/patternfly/components/CalendarMonth/calendar-month-header-nav.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-calendar-month__header-nav{{#if calendar-month-header-nav--modifier}} {{calendar-month-header-nav--modifier}}{{/if}}"
```
christiemolloy, line 1: could this be `calendar-month-control`

src/patternfly/components/CalendarMonth/calendar-month.scss
```diff
@@ -0,0 +1,226 @@
...
+  --pf-c-calendar-month--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-calendar-month--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-calendar-month--FontSize: var(--pf-global--FontSize--sm);
+
+  // header
+  --pf-c-calendar-month__header--MarginBottom: var(--pf-global--spacer--md);
+  --pf-c-calendar-month__header-year--Width: 8ch;
+  --pf-c-calendar-month__header-nav--MarginRight: 0;
+  --pf-c-calendar-month__header-nav--MarginLeft: 0;
+  --pf-c-calendar-month__header-nav--m-prev-month--MarginRight: var(--pf-global--spacer--sm);
```
christiemolloy, line 14: why does this need a bigger margin when the nav is for the prev-month?

src/patternfly/components/CalendarMonth/calendar-month.scss
```diff
@@ -0,0 +1,226 @@
...
+  --pf-c-calendar-month__header-nav--m-prev-month--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-calendar-month__header-nav--m-next-month--MarginRight: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-calendar-month__header-nav--m-next-month--MarginLeft: var(--pf-global--spacer--sm);
+
+  // thead
+  --pf-c-calendar-month__grid-head--BorderBottomSize: var(--pf-global--BorderWidth--sm);
```
christiemolloy, line 20:
```suggestion
  --pf-c-calendar-month__grid-head--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
```

src/patternfly/components/CalendarMonth/calendar-month.scss
```diff
@@ -0,0 +1,226 @@
...
+.pf-c-calendar-month__header-month {
+  flex-grow: 1;
+}
+
+.pf-c-calendar-month__header-year {
+  width: var(--pf-c-calendar-month__header-year--Width);
+}
+
+.pf-c-calendar-month__body {
+  display: flex;
```
christiemolloy, line 98: are these needed ?

christiemolloy (changes requested):
looking great just left a few comments

christiemolloy (commented):
Looking at the sketch link for the design - it looks like the border underneath the "days of the week" is lighter than what you have here?

src/patternfly/components/CalendarMonth/calendar-month-header-nav.hbs
mcoker: I went with nav because it's more specific than a generic control, since we may introduce other controls here. What about `__nav-control`? That would match pagination, and give us room to introduce other types of controls and even a `__nav` in the future if we want a container for the nav.

src/patternfly/components/CalendarMonth/calendar-month.scss
mcoker: Hmmm... not sure what you mean. The prev/next button shave the same margin from the month/year elements, and both have negative margins to align the arrow with the content in the box.

Screen Shot 2020-11-04 at 9 35 44 AM
Screen Shot 2020-11-04 at 9 35 28 AM

src/patternfly/components/CalendarMonth/calendar-month.scss
mcoker: It will keep the calendar part centered if for some reason the container is wider than the calendar, and will keep react styles happy by giving this element styles, but you're right - it's probably better to do it another way. Aligning everything in `pf-c-calendar-month` works better.

src/patternfly/components/CalendarMonth/calendar-month.scss
redallen: Not sure if this is relevant or not, but I'm putting this component inside a Popover which seems to already have some content padding: https://patternfly-react-pr-5026.surge.sh/components/date-picker

mcoker: @redallen would we not use the date picker for that? https://www.patternfly.org/v4/components/date-picker

redallen: That's the date picker I'm refactoring to not use `flatpickr` :)

mcoker: Oh I mean the date picker was built to hold any sort of calendar, not just flatpickr. Could we not continue to use the date picker component with the calendar month component inside of it?

redallen: Unfortunately the current date picker in React is built to hold _exactly_ `flatpickr` which creates its own version of a popover. I've transitioned to just using the `Popover` component in https://github.com/patternfly/patternfly-react/pull/5026 , but it comes with 16px of padding. I can find a way to remove that padding in favor of the `24 24 16 24` in this PR if needbe, just thought I'd let you know!

mcoker:
@christiemolloy updated the BEM elements after our discussion

mcoker:
> Looking at the sketch link for the design - it looks like the border underneath the "days of the week" is lighter than what you have here?

@christiemolloy I'll defer to @maryshak1996 for that. I went with our standard `--pf-global--BorderColor--100`.

src/patternfly/components/CalendarMonth/calendar-month.scss
```diff
@@ -0,0 +1,214 @@
...
+    bottom: 0;
+    left: 0;
+    content: "";
+    border: var(--pf-c-calendar-month__date--after--BorderWidth) solid var(--pf-c-calendar-month__date--after--BorderColor);
+  }
+
+  &,
+  &::after {
+    border-radius: var(--pf-c-calendar-month__date--BorderRadius);
+  }
```
mattnolting, line 186: I think the default focus ring is visually confusing:

Screen Shot 2020-11-03 at 4 09 35 PM

Perhaps we should disable default `:focus` with `outline: none;` since focus styles are explicitly applied below in the CSS.

@mcarrano wdyt?

mcoker: @mattnolting I agree. Also @jessiehuff wdyt about that?

christiemolloy (approved):
fantastic job on this!

mattnolting (approved):
Very nice work! 

I noticed this tiny bug in Safari, though not a blocker. Focus ring is obscured on tab focus. 

Screen Shot 2020-11-10 at 10 18 00 AM

src/patternfly/components/CalendarMonth/calendar-month.scss
jessiehuff: I'm good with this suggestion. The general recommendation is that if you're going to remove the native focus styles, that you must replace them with custom CSS that indicate focus. Jenn's told me that at the beginning of PF4, we took the stance of relying on browser styles for indicating focus when possible, and providing custom CSS when browser styles weren't available (i.e. the Switch component). But there are probably situations where native browser styles are present but not ideal for some reason, and this is probably a more gray, subjective area. I think in cases where we're doing a really custom component (i.e. not a native html element) that chances are there are more reasons in favor of using custom css for focus than native browser styles.

Issue #3635, opened 2020-11-02
Jump links keep focus state after clicking


mcoker:
When you click on a link, it receives browser focus. Then as the user scrolls and we manipulate `pf-m-current` dynamically via scrollspy, the clicked link maintains focus styles even though it is no longer the current link.

PR #3636, opened 2020-11-02
fix(jump-links): use focus-visible for focus styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3635


redallen (approved):
Works to fix the bug @mcarrano found in https://github.com/patternfly/patternfly-react/pull/5089

Issue #3255, opened 2020-07-07
In page wizard - wizard body should scroll, not the entire wizard


mcoker:
In our in page wizard demos, the page scrolls normally, so the entire wizard scrolls if the wizard content exceeds the available height. The wizard should fill the available space and any overflow in the navigation or body should scroll independently in those individual sections, and the footer should be sticky to the bottom of the body section.

Looks like `.pf-c-page__main-wizard` just needs `min-height: 0;` for it to work properly.

I didn't see a react demo for an in page wizard, so we should verify using `.pf-c-page__main-wizard` is the ideal way to include a wizard in the page. If it would require a user specify the wizard section instead of a regular section, and that might lead to inconsistencies, we may also consider adding `min-height: 0;` to `.pf-c-page__main-section` so that the wizard will work in a regular section.
  * That would remove the border we apply to `.pf-c-page__main-wizard`, but if we add `flex-shrink: 0;` to the divider (which seems reasonable), then a horizontal divider could be used there to create the border.

We need to create a follow-up issue in react to create an in page wizard demo once this is sorted out.

related - https://github.com/patternfly/patternfly-design/issues/888

matthewcarleton:
@mcarrano can we prioritize this again? We would benefit from it on the OpenShift side.

mcarrano:
@matthewcarleton this is currently on the roadmap for February 2021. Will that work?

matthewcarleton:
thanks @mcarrano that works - just confirming it hadn't slipped :)

DaoDaoNoCode:
@mcoker Hi, I tried to add `min-height: 0` to `.pf-c-page__main-wizard` or `.pf-c-page__main-section` to try to fix this, but it doesn't work. Do you know why this happens?

mcoker:
@DaoDaoNoCode try also adding `flex-shrink: 1;`. I have a codepen with a demo here - https://codepen.io/mcoker/pen/XWjvVBq

DaoDaoNoCode:
@mcoker It looks good! And it seems that we can also get rid of `.pf-c-page__main-wizard` right? I see you use `.pf-m-wizard` instead. I believe we can put a wizard component inside a normal page section. And I agree we still need more options about whether to using a divider or use a prop `type=wizard` in page section.

mcoker:
[...]

PR #3639, opened 2020-11-03
feat(jumplinks): add demo

### LINKED ISSUE #3560: Introduce jump links demo
Create a demo that shows placement of the jump links/scrollspy navigation inside of a page for both horizontal and vertical variants. See designs here: https://github.com/patternfly/patternfly-design/issues/792

christiemolloy:
closes #3560 

@mcoker @mattnolting what do you guys think about adding a new section in the page to handle the jump links? it looks like in these designs: https://www.sketch.com/s/e6c01dec-8567-436c-b633-eda045100a37/a/Zz2AnY the section needs no top or bottom padding 
Screen Shot 2020-11-03 at 3 36 25 PM

### DIFF
--- a/src/patternfly/demos/JumpLinks/examples/JumpLinks.md
+++ b/src/patternfly/demos/JumpLinks/examples/JumpLinks.md
@@ -1,0 +1,311 @@
+---
+id: Jump links
+section: components
+---
+
+## Demos
+### Vertical
+
+```hbs isFullscreen
+{{#> page page--id=id}}
+  {{#> skip-to-content skip-to-content--attribute=(concat 'href="#main-content-' page--id '"')}}
+    Skip to content
+  {{/skip-to-content}}
+  {{#> page-header}}
+    {{!-- Brand --}}
+    {{#> page-header-brand}}
+      {{#> page-header-brand-toggle}}
+        {{#> button button--modifier="pf-m-plain" button--attribute=(concat 'id="' page--id '-nav-toggle" aria-label="Global navigation" aria-expanded="true" aria-controls="' page--id '-primary-nav"')}}
+          <i class="fas fa-bars" aria-hidden="true"></i>
+        {{/button}}
+      {{/page-header-brand-toggle}}
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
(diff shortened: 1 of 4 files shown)


mcarrano (commented):
Looks good to me @christiemolloy , but what about responsive behavior? What do we expect to happen when we shrink the page down? I think this was addressed in the design.

@lboehling can you also take a look?

mattnolting:
@christiemolloy I can see this demo going two ways

`pf-c-page__main-section.pf-m-jump-links` or `pf-c-page__main-section.pf-m-no-padding.pf-m-inset-xx`

Regardless, I think we should consider adding `insets` to `pf-c-page__main-section`, which may be a separate issue. We could potentially add `insets` in another PR and consume them here...

mcoker:
Should the vertical jump links have the ability to remain sticky? Would we want to support something like this?

72051351-0e3a3880-32fe-11ea-8003-bb4b85685484

It also seems like you should be able to have the vertical jump links adjacent to main sections. In the demo, jump links are beside a content component. I was expecting a user to take an existing page with multiple sections, and have a jump links element positioned to the left or right, and come before all of the page sections it links to in the markup. Kind of like an inline drawer.

To be able to reposition and lay out the jumplinks to the top of the page and as a dropdown per the responsive designs, it seems like we'll need a page component layout to take a content section (the page sections) and jump links component that presents them side by side and allows for the jump links to be sticky or not, and allows the user to switch the layout to a responsive view at a prescribed breakpoint that would move the jump links above the content section, and switch the jump links layout to a mobile layout.

That's just how I envisioned it from looking at the designs, what do y'all think?

And yeah I think it's fair for the jump links to have it's own page section, which gives the user the ability to put it wherever they want. If it follows the react page component sections, `jumpLinks` or something like that would be the prop that a user passes to the page component and we insert it automatically (like we do with tertiary nav and breadcrumbs). Then we would have a `<PageJumpLinks />` component if the user constructed it themselves, like [<PageNavigation](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Page/PageNavigation.tsx) and [<PageBreadcrumb>](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Page/PageBreadcrumb.tsx).

mcarrano:
These are good questions @mcoker I've scheduled some time to discuss. I think this is very related to the request here: https://github.com/patternfly/patternfly/issues/3255 also.

Issue #3257, opened 2020-07-07
Allow for custom icon in alerts


mcarrano:
Want to allow the consumer to specify a custom icon within an alert. The icon must be a member of our existing icon set (i.e. PF icons or FA icons, only)

PR #3641, opened 2020-11-03
feat(alert): added examples to support custom alert icon

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3257

@mcarrano there are no changes in core outside of adding the examples. Any icon can be dropped into any of the alert variations and it will take on the properties of the existing supported icons. Was there anything else we need to support with custom icons?

### DIFF
--- a/src/patternfly/components/Alert/examples/Alert.md
+++ b/src/patternfly/components/Alert/examples/Alert.md
@@ -291,6 +291,27 @@ cssPrefix: pf-c-alert
 {{/alert}}
 ```
 
+### Custom icon
+```hbs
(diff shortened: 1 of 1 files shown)


mcarrano (commented):
Looks good @mcoker . I think that works.

@katierik is this what you were looking for when you filed this issue? https://github.com/patternfly/patternfly-design/issues/861

mcoker:
@mcarrano I verified with Katie that it looks as expected.

mcarrano (approved):
Sounds good. I approve then!

christiemolloy (approved):
LGTM!

PR #3656, opened 2020-11-10
feat(duallistselector): add tree view feat

### LINKED ISSUE #3615: Dual list selector - add menu enhancements
From https://marvelapp.com/prototype/12e47bhj, the menu needs to support checkboxes & expansion.

christiemolloy:
closes #3615

### DIFF
--- a/src/patternfly/components/DualListSelector/dual-list-selector-item-count.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-item-count.hbs
@@ -1,6 +1,6 @@
 <span class="pf-c-dual-list-selector__item-count{{#if dual-list-selector-item-count--modifier}} {{dual-list-selector-item-count--modifier}}{{/if}}"
   {{#if dual-list-selector-item-count--attribute}}
-    {{dual-list-selector-item-count--attribute}}
+    {{{dual-list-selector-item-count--attribute}}}
   {{/if}}>
-  {{>@partial-block}}
+  {{> @partial-block}}
 </span>
--- a/src/patternfly/components/DualListSelector/dual-list-selector-item-count.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-item-count.hbs
@@ -1,0 +1,6 @@
+<span class="pf-c-dual-list-selector__item-count{{#if dual-list-selector-item-count--modifier}} {{dual-list-selector-item-count--modifier}}{{/if}}"
+  {{#if dual-list-selector-item-count--attribute}}
+    {{dual-list-selector-item-count--attribute}}
+  {{/if}}>
+  {{>@partial-block}}
+</span>
--- a/src/patternfly/components/DualListSelector/dual-list-selector.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.hbs
@@ -1,6 +1,6 @@
 <div class="pf-c-dual-list-selector{{#if dual-list-selector--IsTreeview}} pf-m-tree-view{{/if}}{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
   {{#if dual-list-selector--attribute}}
-    {{dual-list-selector--attribute}}
+    {{{dual-list-selector--attribute}}}
   {{/if}}>
   {{>@partial-block}}
 </div>
--- a/src/patternfly/components/DualListSelector/dual-list-selector.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.hbs
@@ -1,4 +1,4 @@
-<div class="pf-c-dual-list-selector{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
+<div class="pf-c-dual-list-selector{{#if dual-list-selector--IsTreeview}} pf-m-tree-view{{/if}}{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
   {{#if dual-list-selector--attribute}}
     {{dual-list-selector--attribute}}
   {{/if}}>
--- a/src/patternfly/components/DualListSelector/dual-list-selector.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.hbs
@@ -1,4 +1,4 @@
-<div class="pf-c-dual-list-selector{{#if dual-list-selector--IsTreeview}} pf-m-tree-view{{/if}}{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
+<div class="pf-c-dual-list-selector{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
   {{#if dual-list-selector--attribute}}
     {{{dual-list-selector--attribute}}}
   {{/if}}>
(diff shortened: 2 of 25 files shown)


mcarrano (changes requested):
I'd expect behavior of tree list items to be identical to the Tree view here with items having a hover state: https://www.patternfly.org/v4/components/tree-view/html#with-checkboxes

src/patternfly/components/DualListSelector/dual-list-selector.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="pf-c-dual-list-selector{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
+<div class="pf-c-dual-list-selector{{#if dual-list-selector--IsTreeview}} pf-m-tree-view{{/if}}{{#if dual-list-selector--modifier}} {{dual-list-selector--modifier}}{{/if}}"
```
mcoker, line 1: Is this class necessary, or could we scope the style changes to `__list-item.pf-m-expandable`?

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -128,26 +166,31 @@
...
+}
+
+.pf-c-dual-list-selector__list-item {
 
-  &.pf-m-selected {
-    --pf-c-dual-list-selector__list-item--BackgroundColor: var(--pf-c-dual-list-selector__list-item--m-selected--BackgroundColor);
-    --pf-c-dual-list-selector__item-text--Color: var(--pf-c-dual-list-selector__list-item--m-selected__item-text--Color);
+  // stylelint-disable
+  &.pf-m-expanded {
+    .pf-c-dual-list-selector__item > .pf-c-dual-list-selector__item-main > .pf-c-dual-list-selector__item-toggle > .pf-c-dual-list-selector__item-toggle-icon > i {
```
mcoker, line 187: If this is to keep nested icons from rotating, you can just use this.

Note I'm styling `.pf-c-dual-list-selector__item-toggle-icon` instead of `i`, since that wouldn't rotate the icon in react. That also will require `display: inline-block;` on `__item-toggle-icon` so that it will transform, and an initial var for `--pf-c-dual-list-selector__item-toggle-icon--Transform: 0;`

```
&.pf-m-expanded {
  > .pf-c-dual-list-selector__item .pf-c-dual-list-selector__item-toggle-icon {
    --pf-c-dual-list-selector__item-toggle-icon--Transform: rotate(--pf-c-dual-list-selector__list-item--m-expanded__item-toggle-icon--Rotate);
  }
}
```

src/patternfly/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -676,6 +855,11 @@ cssPrefix: pf-c-dual-list-selector
 | `.pf-c-dual-list-selector__item-text` | `<span>` | Initiates a dual list selector pane menu item text. **Required** |
 | `.pf-c-dual-list-selector__controls` | `<div>` | Initiates the dual list selector controls. **Required** |
 | `.pf-c-dual-list-selector__controls-item` | `<div>` | Initiates the dual list selector controls item. **Required** |
+| `.pf-c-dual-list-selector__item-check` | `<span>` | Initiates the dual list selector item check. |
```
mcoker, line 858: can you move these `__item-` elements above `__controls` with the other `__item-` things?

src/patternfly/components/DualListSelector/dual-list-selector-item-count.hbs
```diff
@@ -0,0 +1,6 @@
+<span class="pf-c-dual-list-selector__item-count{{#if dual-list-selector-item-count--modifier}} {{dual-list-selector-item-count--modifier}}{{/if}}"
```
mcoker, line 1: Need some styles for this element - looks like it needs a left margin and a background color override similar to tree view

https://github.com/patternfly/patternfly/blob/2c6324d2ec18eff85b9baee3ed2ec1a0a1b9695d/src/patternfly/components/TreeView/tree-view.scss#L38

mcoker (commented):
Looks good @christiemolloy! Left some feedback.

I know we decided this isn't the tree view component, yet it shares a tree view/nestable layout. I think building it as it is now is fine for the time being, but I think we should consider the differences between the tree view component and the tree view list we're building here and see if...

* They're actually the same thing, and we should use tree view here.
* They aren't, but could be if we made tree view a little more flexible.
* They aren't, but they share something else that we could abstract out so the layout inside is shared
  * Like an underlying list structure that builds the nested layout but the tree view and dual list selector components individually define the interaction and functionality for the list.

christiemolloy:
> I'd expect behavior of tree list items to be identical to the Tree view here with items having a hover state: https://www.patternfly.org/v4/components/tree-view/html#with-checkboxes
> 
> cc @mmenestr

I spoke to Margot, and we discussed that the list items in the nested tree shouldn't be selectable because selecting them does nothing. The toggle expands the item and the checkbox is what selects the item to move it over... therefore we didnt think it needed a hover focus style. But let us know your thoughts

mmenestr:
Yeah I think after talking with Matt, and others I take back what I said about the hover state, and I would add it + change the badge color to be darker grey just like in tree view so that it doesn't "disappear" when it's in hover state.

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -176,3 +223,20 @@
   padding-right: var(--pf-c-dual-list-selector__controls--PaddingRight);
   padding-left: var(--pf-c-dual-list-selector__controls--PaddingLeft);
 }
+
+.pf-c-dual-list-selector__item-main {
+  display: flex;
+}
```
mattnolting, line 229: ```min-width: 0;```

mattnolting (changes requested):

mattnolting (approved):
Fantastic! LPTM 👍

src/patternfly/components/DualListSelector/dual-list-selector.hbs
christiemolloy: removed

mcoker (approved):

mmenestr:
I just saw that this was merged but there seems to be a potential bug, where you can select a row without selecting a checkbox? 
The multiple level dual lists shouldn't have a selection state on the row itself, it should just be checkbox selected or not! Alternatively though, I can see a case where the selection state changes the row visuals as well as the checkbox state - but that's not what seems to be shown in the example so I'm confused haha

If you look at the screenshot below, "Red" has the intended selected state bc it's just the checkbox selected, where as orange does not. Orange seems to be selecting like a single list dual list, but without the checkbox being checked? 
Just a little confused about that.
image

Also for this one, not sure what it's supposed to show? Are those rows supposedly selected or not selected? I'd imagine you're trying to show a selected state but the checkbox is empty. I feel like i'm missing something hahah
image

mcarrano:
@mcoker how should we handle this? It's scheduled to go to React in the next milestone. Should we just open a new core issue to make these updates before beginning the React work?

mcoker:
@christiemolloy was there a reason for those selection states in the tree menu examples?

Assuming we just need to update the examples so that the tree selection state only happens by the checkbox, I can make that update before the react work starts.

Issue #3648, opened 2020-11-05
Move splitter component into the drawer


mcoker:
The splitter is only intended for use in the drawer. Let's remove the component, and move it into the the drawer. It doesn't necessarily need to be styled yet until we look at how to best incorporate the splitter into the drawer so it can be used to resize the panel.

Issue #3610, opened 2020-10-22
Splitter demo


mcarrano:
Build a Splitter component demo according to the designs here: https://www.sketch.com/s/3fe9a72b-a9c3-4d12-a02d-c843c41128d4

Need to consider where the Splitter lives - inside or outside of the drawer. Does the splitter remain visible when the drawer is closed?

PR #3659, opened 2020-11-11
feat(drawer): added resizable drawer, moved splitter

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3648
fixes https://github.com/patternfly/patternfly/issues/3610

Removes the splitter component and adds it as an element of the drawer
Creates resizable drawers


mcarrano (commented):
I think this looks good @mcoker . It's a bit hard to get a sense of what this will look like visually when placed in a "real" context, but I guess we can make visual adjustments later if needed. @mceledonia can you also take a look?

mattnolting (approved):
LPTM! 🌮

mceledonia:
One last thing, inline example should also have the border on the other side as well. 
image

mceledonia (approved):
Looks good!

mattnolting:
LGTM!

mcarrano (approved):
Looks good

nicolethoen (commented):
It looks like the blue focus border on the last inline example is not going all the way around the splitter on chrome

src/patternfly/components/Drawer/examples/Drawer.md
```diff
@@ -242,3 +306,4 @@ import './Drawer.css'
 | `.pf-m-no-padding` | `.pf-c-drawer__body` | Modifies the element to remove padding. |
 | `.pf-m-no-background` | `.pf-c-drawer__section`, `.pf-c-drawer__content`, `.pf-c-drawer__panel` | Modifies the drawer body/panel background color to transparent. |
 | `.pf-m-width-{25, 33, 50, 66, 75, 100}{-on-[breakpoint]}` | `.pf-c-drawer__panel` | Modifies the drawer panel width. |
+| `.pf-m-resizable` | `.pf-c-drawer__panel` | Modifies the drawer panel to be resizable. |
```
christiemolloy, line 309: I think this could be mistaken as meaning the drawer can be different sizes, i think you should include that its resizable by using a splitter or handle or something a little more detailed.

mcoker:
@nicolethoen I noticed that, too. @mcarrano @mceledonia is that something you want us to update? It's because the splitter border overlaps. I can bring the focus outline in by 1px, which will look like this

Screen Shot 2020-11-11 at 3 20 45 PM

christiemolloy (changes requested):

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -278,6 +334,68 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
...
+  position: absolute;
+  top: var(--pf-c-drawer__splitter--Top);
+  right: var(--pf-c-drawer__splitter--Right);
+  bottom: var(--pf-c-drawer__splitter--Bottom);
+  left: var(--pf-c-drawer__splitter--Left);
+  display: none;
+  width: var(--pf-c-drawer__splitter--Width);
+  height: var(--pf-c-drawer__splitter--Height);
+  cursor: var(--pf-c-drawer__splitter--Cursor);
+  visibility: visible;
```
christiemolloy, line 347: why is this display: none and visibility: visible ?

christiemolloy (changes requested):

src/patternfly/components/Drawer/drawer.scss
mcoker: because I oof'd 👍

src/patternfly/components/Drawer/drawer-splitter.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="pf-c-drawer__splitter{{#if drawer-splitter--modifier}} {{drawer-splitter--modifier}}{{/if}}{{#if drawer-splitter--IsVertical}} pf-m-vertical{{/if}}"
```
tlabaj, line 1: @jessiehuff does this need an `aria-label`?

mcarrano:
>@nicolethoen I noticed that, too. @mcarrano @mceledonia is that something you want us to update? It's because the splitter border overlaps. I can bring the focus outline in by 1px, which will look like this

Sure. Do it.

christiemolloy (approved):
LGTM ship!!

src/patternfly/components/Drawer/drawer-splitter.hbs
jessiehuff: I think this is probably ok. When testing in VO, it announces as "horizontal splitter" or "vertical splitter" because of the aria attributes so I think this is probably fine.

PR #3661, opened 2020-11-11
Fix(tree-view): Fix indentation

### LINKED ISSUE #3532: Tree View: confusing indentation difference of expandable vs non-expandable items
I noticed something about the way the tree view handles indentation that indicates to me that there may be a CSS bug, but I'm not sure if it was an intentional design decision.

The `.pf-c-tree-view__list-item` is always positioned at the same horizontal position regardless of whether it is `.pf-m-expandable` or not. The `.pf-m-expandable` version has a ">" icon for the expand/collapse control, pushing the icon/checkbox/name/etc of the item farther to the right than its non-expandable siblings.

When we have a tree where some items are expandable and other sibling items are not, especially with checkboxes and icons, this leads to a confusing layout where it can appear that the expandable items are one level deeper than they are (or, non-expandable items can appear one level shallower than they are).

For example, I have a tree with data structured like this:

* All datacenters
  * Fake_DC
  * V2V_DC
    * Fake_Cluster
    * V2V_Cluster
      * esx13.v2v.bos.redhat.com
    * jortel

The TreeView component displays it like this:

Screen Shot 2020-09-29 at 4 36 49 PM

As you can see, any item with the expand/collapse arrow has its checkbox pushed to the right and aligned with its children, so it looks like V2V_DC is a child of Fake_DC when they are really siblings, and it looks like Fake_Cluster is a sibling of V2V-DC when it is really a child.

As a temporary solution, I added this CSS in our app:
```css
.pf-c-tree-view li.pf-c-tree-view__list-item:not(.pf-m-expandable) {
  padding-left: var(--pf-global--spacer--lg);
}
```

And now the tree looks like this:

Screen Shot 2020-09-29 at 4 52 16 PM

The checkboxes and icons are now aligned as I would expect based on the data structure.

@christiemolloy @mcarrano is this an issue we should fix or an intentional design decision?

christiemolloy:
closes #3532

### DIFF
--- a/src/patternfly/components/TreeView/tree-view.scss
+++ b/src/patternfly/components/TreeView/tree-view.scss
@@ -5,9 +5,8 @@ $pf-c-tree-view--MaxNesting: 10;
   --pf-c-tree-view--PaddingBottom: var(--pf-global--spacer--sm);
 
   // Link base
-  --pf-c-tree-view__node--indent--base: var(--pf-global--spacer--sm);
+  --pf-c-tree-view__node--indent--base: calc(var(--pf-global--spacer--lg) + var(--pf-global--spacer--sm));
   --pf-c-tree-view__node--nested-indent--base: calc(var(--pf-global--spacer--lg) + var(--pf-global--spacer--sm));
-  --pf-c-tree-view__item--m-expandable__node--nested-indent--base: var(--pf-global--spacer--lg);
 
   // Link
   --pf-c-tree-view__node--PaddingTop: var(--pf-global--spacer--sm);
@@ -72,6 +71,7 @@ $pf-c-tree-view--MaxNesting: 10;
 // stylelint-enable
 
 .pf-c-tree-view__node {
+  position: relative;
   display: flex;
   flex: 1 1;
   align-items: center;
@@ -110,12 +110,15 @@ $pf-c-tree-view--MaxNesting: 10;
 }
 
 .pf-c-tree-view__node-toggle {
+  position: absolute;
+  left: calc(var(--pf-c-tree-view__node--PaddingLeft) - var(--pf-global--spacer--lg) - var(--pf-global--spacer--lg));
   display: inline-flex;
   align-items: center;
   justify-content: center;
   min-width: var(--pf-c-tree-view__node-toggle--MinWidth);
   min-height: var(--pf-c-tree-view__node-toggle--MinHeight);
   padding-right: var(--pf-c-tree-view__node-toggle--PaddingRight);
+  transform: translateX(100%);
 }
 
 .pf-c-tree-view__node-text {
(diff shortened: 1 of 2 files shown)


src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -5,9 +5,8 @@ $pf-c-tree-view--MaxNesting: 10;
   --pf-c-tree-view--PaddingBottom: var(--pf-global--spacer--sm);
 
   // Link base
-  --pf-c-tree-view__node--indent--base: var(--pf-global--spacer--sm);
-  --pf-c-tree-view__node--nested-indent--base: var(--pf-global--spacer--lg);
-  --pf-c-tree-view__item--m-expandable__node--nested-indent--base: var(--pf-global--spacer--lg);
+  --pf-c-tree-view__node--indent--base: calc(var(--pf-global--spacer--lg) + var(--pf-global--spacer--sm));
+  --pf-c-tree-view__node--nested-indent--base: calc(var(--pf-global--spacer--lg) + var(--pf-global--spacer--sm));
```
mattnolting, line 9:
```suggestion
  --pf-c-tree-view__node--nested-indent--base: var(--pf-global--spacer--lg);
```

Looks like this has a bit too much space?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -110,12 +110,15 @@ $pf-c-tree-view--MaxNesting: 10;
 }
 
 .pf-c-tree-view__node-toggle {
+  position: absolute;
+  left: calc(var(--pf-c-tree-view__node--PaddingLeft) - var(--pf-global--spacer--lg) - var(--pf-global--spacer--lg));
```
mcoker, line 114: need a var for this

mcoker: Couldn't this just be `--pf-c-tree-view__node--PaddingLeft`, then use `transform: translateX(-100%);`?

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -110,12 +110,15 @@ $pf-c-tree-view--MaxNesting: 10;
...
   display: inline-flex;
   align-items: center;
   justify-content: center;
   min-width: var(--pf-c-tree-view__node-toggle--MinWidth);
   min-height: var(--pf-c-tree-view__node-toggle--MinHeight);
   padding-right: var(--pf-c-tree-view__node-toggle--PaddingRight);
+  transform: translateX(100%);
```
mcoker, line 121: need a var for this too

src/patternfly/components/TreeView/tree-view.scss
mcoker, line 114: Normally I would add a `top` value here, but it seems to be working without it. Maybe someone can explain why that works some day? If we wanted to add that, it should work with `top: var(--pf-c-tree-view__node--PaddingTop)`

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -110,12 +111,16 @@ $pf-c-tree-view--MaxNesting: 10;
 }
 
 .pf-c-tree-view__node-toggle {
+  position: absolute;
+  top: var(--pf-c-tree-view__node--PaddingTop);
```
mcoker, line 115: Can you create a var for this and set `--pf-c-tree-view__node-toggle--Top: var(--pf-c-tree-view__node--PaddingTop);`

```suggestion
  top: var(--pf-c-tree-view__node-toggle--Top);
```

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -110,12 +111,16 @@ $pf-c-tree-view--MaxNesting: 10;
...
+  left: calc(var(--pf-c-tree-view__node--PaddingLeft));
```
mcoker, line 116: same for this one, and you have an unused `calc()`

mcoker (approved):
LPTM!

mcoker (approved):

Issue #3668, opened 2020-11-12
Calendar month round 2


redallen:
1. [x] Focus style should be different from hover style. For now give hover a background. Alternatively give focus a non-circular border.
2. [x] `.pf-m-adjacent-month.pf-selected` needs better text contrast:
![image](https://user-images.githubusercontent.com/47335686/98970765-4e4de700-24de-11eb-8989-d192c4317254.png)
3. [ ] Add screenreader text for days of week.
    - Will be first implemented in https://github.com/patternfly/patternfly-react/pull/5026 and then can be ported
4. [ ] Make the date buttons `aria-labelledby` the month and year so it would announce something like "4 March 2020". cc @jessiehuff 
    - Will be first implemented in https://github.com/patternfly/patternfly-react/pull/5026 and then can be ported
5. [ ] Use Input group + Button for calendar icon
   - Already implemented in https://github.com/patternfly/patternfly-react/pull/5026

redallen:
6. Make `pf-m-focus` differentiable from `:focus` for tabbing to the table after interacting with header controls:
![what-is-focused](https://user-images.githubusercontent.com/47335686/98990357-5c106600-24f8-11eb-9c38-2dd5c447e3c0.gif)
(In this gif I'm confused that the "5" has focus before moving to the 6).

cc @mcarrano

mcarrano:
Looks right @redallen . Working with @mcoker on the hover and focus appearances now. But as I'm testing this, seems like I can get into places where the keyboard interaction breaks as I navigate to a new month. Having a hard time to consistently reproduce but you might need more testing here.

redallen:
I'm noticing that focus is no longer trapped when you click inside the Calendar month which may be the cause of what you're seeing. I'll see what I can do tomorrow about it.

Steps to reproduce would be helpful.

redallen:
Closing in favor of #3679.

PR #3669, opened 2020-11-12
fix(calendar-month): removed focus outline, fixed selected styling

mcoker:
part of https://github.com/patternfly/patternfly/issues/3668

* Removes default browser focus outline
* Fixes adjacent month selected styling


redallen (approved):
LGTM!

PR #3671, opened 2020-11-12
fix(calendar-month): differentiate hover/focus styles

mcoker:
part of https://github.com/patternfly/patternfly/issues/3668, differentiates hover and focus styles


mcarrano (approved):
Looks great @mcoker

PR #3691, opened 2020-11-19
feat(table): extended borderless to expandable

mattnolting:
**Previews start here:** https://patternfly-pr-3691.surge.sh/components/table#borderless

I added `.pf-c-table__control-row` for compound expandable table control rows, that allows for a border below that works with hover/active styles:

Screen Shot 2020-11-19 at 4 13 23 PM

Screen Shot 2020-11-19 at 4 13 30 PM

Screen Shot 2020-11-19 at 3 58 20 PM

I think we may want to keep some borders on `.pf-m-expandable` tables. 

@mceledonia @mcarrano Here are SS:

Screen Shot 2020-11-19 at 3 58 30 PM

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -649,7 +649,7 @@ When header cells are empty or they contain interactive elements, `<th>` should 
 
   {{#> table-tbody table-tbody--modifier="pf-m-expanded"}}
     {{#> table-tr table-tr--expanded="true"}}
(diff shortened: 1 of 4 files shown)


mcarrano:
@mattnolting so is what you are saying here that if we allow any table to be borderless then we no longer have separation between the bottom of the expanded table and the next row? That's my interpretation of what you show above. I agree that in this case, i.e. for an expanded row, we always need these borders even if the consumer chooses to make the table borderless. Does that make sense?

mceledonia (approved):

mcoker (approved):

mcarrano (approved):
Looks great!

Issue #3679, opened 2020-11-16
Calendar month round 3


redallen:
Now that https://github.com/patternfly/patternfly-react/pull/5026 is in React the HTML here should be updated. I also have one CSS nit remaining.

- [ ] CSS nit: if a cell has `.pf-m-selected` applied you cannot tell when its button has `:focus` applied since the borders overlap:
![image](https://user-images.githubusercontent.com/47335686/99288360-78b6e180-2809-11eb-9d51-bdaae99ef7d8.png)
[W3's date picker](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html) gets around this by having a different border for selected/focused.
- [x] Single textinput replaced with InputGroup. The `button.pf-c-button.pf-m-control` contains a OutlinedCalendarAltIcon instead of custom SVG
- [x] Title span added to month select for a11y
- [x] Added hidden day of week `<span>`s for long day names like Monday
- [x] Aria-label added to buttons in `<td>`s

PR #3696, opened 2020-11-20
fix(calendar-month): updated a11y per the react component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3679

> if a cell has .pf-m-selected applied you cannot tell when its button has :focus applied since the borders overlap

@mcarrano do you want to do something about this? We could use a blue shadow on focus or something like that?

Nov-20-2020 14-46-36

### DIFF
--- a/src/patternfly/components/DatePicker/examples/DatePicker.md
+++ b/src/patternfly/components/DatePicker/examples/DatePicker.md
@@ -24,21 +24,59 @@ import './DatePicker.css'
 ### Helper text
 ```hbs
 {{#> date-picker date-picker--id="helper-text" date-picker-helper-text--text="Select a date."}}
-  {{> form-control controlType="input" input="true" form-control--modifier="pf-m-icon pf-m-calendar" form-control--attribute=(concat 'type="text" value="2020-03-05" id="' date-picker--id '-input" name="' date-picker--id '-input" aria-label="Basic date picker example"')}}
+  {{#> input-group}}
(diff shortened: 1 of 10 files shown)


mcarrano:
I think this is a good solution @mcoker .

src/patternfly/components/DatePicker/examples/DatePicker.md
```diff
@@ -12,7 +12,12 @@ import './DatePicker.css'
 ### Basic
 ```hbs
 {{#> date-picker date-picker--id="basic"}}
-  {{> form-control controlType="input" input="true" form-control--modifier="pf-m-icon pf-m-calendar" form-control--attribute=(concat 'type="text" value="03/05/2020" id="' date-picker--id '-input" name="' date-picker--id '-input" aria-label="Basic date picker example"')}}
+  {{#> input-group}}
+    {{> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" value="03/05/2020" id="' date-picker--id '-input" name="' date-picker--id '-input" aria-label="Date picker"')}}
```
redallen, line 16: Can you change `value="03/05/2020"` to `value="2020-03-05"`? We also made that change upstream.

redallen (approved):
LGTM!!

mceledonia (approved):

mattnolting (approved):
LGTM! I tested w/VO and I the behavior is a little odd, but it's also consistent with google calendar and a few other JS calendars I tested. The `<td>` and `<button>` are focusable and announce different sets of info..

Issue #3704, opened 2020-11-24
Calendar month first row is taller than other rows


redallen:
The first row seems to be 48px tall while the others are 41px. This leads to the range classes looking a little funny for the first row:
![image](https://user-images.githubusercontent.com/47335686/100126391-e1c2d880-2e4b-11eb-8f06-d1c80e3a7b47.png)

Fixes this sprint would fix https://github.com/patternfly/patternfly-react/pull/5184 also.

PR #3705, opened 2020-11-24
fix(calendar-month): updated in-range bg to stay within cell padding

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3704


redallen (approved):
LGTM!
image

PR #3707, opened 2020-11-25
fix(jump-links): mark only top-level item as current

### LINKED ISSUE #3683: Jump links subsections - active item/section styling
Re: https://github.com/patternfly/patternfly/pull/3632#issuecomment-728528421

Right now there is no way to mark the top-level item with a subsection as current without marking the whole subsection current. Should we be able to do that? In what ways do we want to support active styling with subsections?

cc @mcarrano @lboehling @ajacobs21e

DaoDaoNoCode:
fixes #3683 .
Now when you set the top-level item as current, it will no longer mark the whole subsection under this item as current.

### DIFF
--- a/src/patternfly/components/JumpLinks/jump-links.scss
+++ b/src/patternfly/components/JumpLinks/jump-links.scss
@@ -10,7 +10,7 @@
   --pf-c-jump-links--m-vertical__list--PaddingLeft: 0;
   --pf-c-jump-links__list--FlexDirection: row;
   --pf-c-jump-links--m-vertical__list--FlexDirection: column;
-  --pf-c-jump-links__list--before--BorderColor: var(--pf-global--BorderColor--100);
+  --pf-c-jump-links__list--before--BorderColor: transparent;
   --pf-c-jump-links__list--before--BorderTopWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-jump-links__list--before--BorderRightWidth: 0;
   --pf-c-jump-links__list--before--BorderBottomWidth: 0;
@@ -157,9 +157,6 @@
     --pf-c-jump-links__link--before--BorderLeftWidth: var(--pf-c-jump-links__item--m-current__link--before--BorderLeftWidth);
     --pf-c-jump-links__link--before--BorderColor: var(--pf-c-jump-links__item--m-current__link--before--BorderColor);
     --pf-c-jump-links__link-text--Color: var(--pf-c-jump-links__item--m-current__link-text--Color);
-  }
-  &.pf-m-current > .pf-c-jump-links__list::before {
-    border-color: transparent;
   }
 }
 // stylelint-enable
--- a/src/patternfly/components/JumpLinks/jump-links.scss
+++ b/src/patternfly/components/JumpLinks/jump-links.scss
@@ -10,7 +10,7 @@
   --pf-c-jump-links--m-vertical__list--PaddingLeft: 0;
   --pf-c-jump-links__list--FlexDirection: row;
   --pf-c-jump-links--m-vertical__list--FlexDirection: column;
-  --pf-c-jump-links__list--before--BorderColor: transparent;
+  --pf-c-jump-links__list--before--BorderColor: var(--pf-global--BorderColor--100);
   --pf-c-jump-links__list--before--BorderTopWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-jump-links__list--before--BorderRightWidth: 0;
   --pf-c-jump-links__list--before--BorderBottomWidth: 0;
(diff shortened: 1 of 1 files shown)


src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -152,12 +152,15 @@
 
 // stylelint-disable
 .pf-c-jump-links__item {
-  &.pf-m-current {
+  &.pf-m-current > .pf-c-jump-links__link {
```
mcoker, line 155: 👍

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -152,12 +152,15 @@
...
     --pf-c-jump-links__link--before--BorderTopWidth: var(--pf-c-jump-links__item--m-current__link--before--BorderTopWidth);
     --pf-c-jump-links__link--before--BorderLeftWidth: var(--pf-c-jump-links__item--m-current__link--before--BorderLeftWidth);
     --pf-c-jump-links__link--before--BorderColor: var(--pf-c-jump-links__item--m-current__link--before--BorderColor);
     --pf-c-jump-links__link-text--Color: var(--pf-c-jump-links__item--m-current__link-text--Color);
   }
+  &.pf-m-current > .pf-c-jump-links__list {
+    margin-top: 0;
```
mcoker, line 162: This will cause the nested list to shift as the top-level item is made current.

Nov-30-2020 10-48-46

The left grey border on a nested list doesn't need to be there. If you can disable it somehow, that will fix the border overlap.

mcoker (changes requested):
Thanks @DaoDaoNoCode, it's looking good! One comment about the margin change for nested jump links, I think we'll need to try something else to fix that so that the layout doesn't shift.

src/patternfly/components/JumpLinks/jump-links.scss
DaoDaoNoCode: Fixed, thanks!

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -152,12 +152,15 @@
...
 .pf-c-jump-links__item {
-  &.pf-m-current {
+  &.pf-m-current > .pf-c-jump-links__link {
     --pf-c-jump-links__link--before--BorderTopWidth: var(--pf-c-jump-links__item--m-current__link--before--BorderTopWidth);
     --pf-c-jump-links__link--before--BorderLeftWidth: var(--pf-c-jump-links__item--m-current__link--before--BorderLeftWidth);
     --pf-c-jump-links__link--before--BorderColor: var(--pf-c-jump-links__item--m-current__link--before--BorderColor);
     --pf-c-jump-links__link-text--Color: var(--pf-c-jump-links__item--m-current__link-text--Color);
   }
+  &.pf-m-current > .pf-c-jump-links__list::before {
+    border-color: transparent;
```
mcoker, line 162: This does it! Although two things:

1. This only disables the border on nested lists when the parent item is current. While that fixes this particular issue, it would be great if we can just disable that border on any nested list.

2. We have a var for the list border - `--pf-c-jump-links__list--before--BorderColor`. You can manipulate that variable to change the border color as opposed to writing out a selector and targeting the `border-color` property. And you can redefine that variable on a parent selector as opposed to targeting the actual element it applies to, since css vars are inherited.

For example, if you have this code below - here's a [codepen](https://codepen.io/mcoker/pen/vYXOyZJ):

```
<style>
:root {
  --link-color: blue;
}

a {
  color: var(--link-color);
}
</style>

<p><a href="#">blue link</a></p>
<p class="foo"><a href="#">also a blue link</a></p>
```

And you want to change the link color to red when it's in `.foo`, you can write:

```
.foo {
  --link-color: red;
}
```

instead of having to write

```
.foo a {
  --link-color: red;
}
```

Here is a codepen for that - https://codepen.io/mcoker/pen/bGwdBRx

DaoDaoNoCode: Thanks! That makes sense.

src/patternfly/components/JumpLinks/jump-links.scss
```diff
@@ -10,7 +10,7 @@
   --pf-c-jump-links--m-vertical__list--PaddingLeft: 0;
   --pf-c-jump-links__list--FlexDirection: row;
   --pf-c-jump-links--m-vertical__list--FlexDirection: column;
-  --pf-c-jump-links__list--before--BorderColor: var(--pf-global--BorderColor--100);
+  --pf-c-jump-links__list--before--BorderColor: transparent;
```
mcoker, line 13: This removes the border all together 😄 We only want to remove from nested lists.

mcoker (approved):
LGTM! Thanks @DaoDaoNoCode!

Issue #296, opened 2018-06-05
Slider


LHinson:
Interaction design and React to first determine which slider will be used.

https://www.patternfly.org/pattern-library/forms-and-controls/slider/

#180 has open questions that also impact this issue. Need to determine where the CSS lives and if we are using a third party library.

LHinson:
Yes, hold tight while I found out more. I'll get back to you asap

maryshak1996:
Thanks @LHinson

LHinson:
@maryshak1996 After talking with Joachim, I think we need to first open an investigation story on React before we can proceed with visuals. Therefore we are going to put this one back on the backlog and hold off. For that reason, I'll also un-assign you for the time being. Apologies about that!

maryshak1996:
Hey @LHinson no worries! Thanks for the update!

maryshak1996:
@mcarrano @gdoyle1 I've put together a handful of different ways to look at (a) discrete sliders, (b) continuous sliders, and (c) codependent slider
I don't recall the codependent slider being in the requirements, but @bmignano and I a codependent slider that the kiali team had a need for, so there is at least some teams that would appreciate this being a part of the patternfly components. 
I'm curious as to your thoughts on if the various options that I'm presenting here for the discrete and continuous sliders are all variations that should be available in PF4, or if we should come to a consensus on have, for example, just top-aligned numeric indicators as opposed to right-aligned. I think that there are arguments to be made for having one for the sake of consistency, but I also think that there are times when the layout of a page might be more conducive to one alignment over another. Here's what I've got:
![screen shot 2019-03-04 at 11 46 25 am](https://user-images.githubusercontent.com/19824574/53814764-429ad680-3f2e-11e9-8bc8-45e10abb32bf.png)

gdoyle1:
@maryshak1996 @mcarrano  I think it would make sense to recommend using the designs that have the percentages to the right of the slider (Option 3 discrete, Option 1 continuous) and then if they don't have enough space, they can use the slider alone (Option 1 discrete, Option 1 continuous). I do think it is worth showing the codependent slider if we needed it beforehand too! Also would show that the slider is customizable if needed.

mcarrano:
I agree with @gdoyle1 . The other thing I've seen is to show a tooltip with the value when the user hovers over the slide handle. The only other thing I'll say is that the text field with the underscore looks a bit odd in this context (I know that's our standard text box). @mceledonia would be interested in your thoughts.

[...]

PR #3711, opened 2020-12-01
feat(slider): added slider component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/296

**SLIDER LINK - https://patternfly-pr-3711.surge.sh/components/slider**

Designs - https://www.sketch.com/s/6f81d099-0c11-407a-8962-5672df30dcf2/a/j5KAvW

WAI-ARIA example - https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html

### DIFF
--- a/src/patternfly/components/Slider/examples/Slider.md
+++ b/src/patternfly/components/Slider/examples/Slider.md
@@ -30,27 +30,6 @@ cssPrefix: pf-c-slider
     {{> slider-thumb}}
   {{/slider-main}}
 {{/slider}}
-
-<br>
-
-{{#> slider
-  slider--value-min="0"
-  slider--value-max="100"
-  slider--value-now="50"
-  slider--attribute='style="--pf-c-slider--value: 50%;"'
-  }}
-  {{#> slider-main}}
-    {{> slider-rail}}
-    {{#> slider-steps}}
-      {{> slider-step slider-step--Left="0%" slider-step--label="0%" slider-step--IsActive="true"}}
-      {{> slider-step slider-step--Left="25%" slider-step--IsActive="true"}}
-      {{> slider-step slider-step--Left="50%" slider-step--label="50%" slider-step--IsActive="true"}}
-      {{> slider-step slider-step--Left="75%"}}
-      {{> slider-step slider-step--Left="100%" slider-step--label="100%"}}
-    {{/slider-steps}}
-    {{> slider-thumb}}
-  {{/slider-main}}
-{{/slider}}
 ```
 
 ### Continuous
@@ -63,6 +42,22 @@ cssPrefix: pf-c-slider
   }}
   {{#> slider-main}}
     {{> slider-rail}}
+    {{> slider-thumb}}
+  {{/slider-main}}
+{{/slider}}
+
+{{#> slider
+  slider--value-min="0"
+  slider--value-max="100"
+  slider--value-now="50"
+  slider--attribute='style="--pf-c-slider--value: 50%;"'
+  }}
+  {{#> slider-main}}
+    {{> slider-rail}}
+    {{#> slider-steps}}
+      {{> slider-step slider-step--Left="0%" slider-step--label="0%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="100%" slider-step--label="100%"}}
+    {{/slider-steps}}
     {{> slider-thumb}}
   {{/slider-main}}
 {{/slider}}
--- a/src/patternfly/components/Slider/examples/Slider.md
+++ b/src/patternfly/components/Slider/examples/Slider.md
@@ -254,6 +254,7 @@ cssPrefix: pf-c-slider
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-slider` | `<div>` | Initiates the slider component. **Required** |
+| `.pf-c-slider__main` | `<div>` | Initiates the slider main element. **Required** |
 | `.pf-c-slider__rail` | `<div>` | Initiates the slider rail. **Required** |
 | `.pf-c-slider__rail-track` | `<div>` | Initiates the slider rail track. **Required** |
 | `.pf-c-slider__steps` | `<div>` | Initiates the slider steps. |
--- a/src/patternfly/components/Slider/examples/Slider.md
+++ b/src/patternfly/components/Slider/examples/Slider.md
@@ -175,7 +175,7 @@ cssPrefix: pf-c-slider
   }}
   {{#> slider-actions}}
     {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Minus"'}}
-      <i class="fas fa-minus" aria-hidden="true"></i>
+      <i class="fas fa-fw fa-minus" aria-hidden="true"></i>
     {{/button}}
   {{/slider-actions}}
   {{#> slider-main}}
@@ -184,7 +184,7 @@ cssPrefix: pf-c-slider
   {{/slider-main}}
   {{#> slider-actions}}
     {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Plus"'}}
-      <i class="fas fa-plus" aria-hidden="true"></i>
+      <i class="fas fa-fw fa-plus" aria-hidden="true"></i>
     {{/button}}
   {{/slider-actions}}
 {{/slider}}
@@ -209,7 +209,7 @@ cssPrefix: pf-c-slider
   {{/slider-main}}
   {{#> slider-actions}}
     {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Locked"'}}
-      <i class="fas fa-lock" aria-hidden="true"></i>
+      <i class="fas fa-fw fa-lock" aria-hidden="true"></i>
     {{/button}}
   {{/slider-actions}}
 {{/slider}}
@@ -234,7 +234,7 @@ cssPrefix: pf-c-slider
   {{/slider-main}}
   {{#> slider-actions}}
     {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Lock"'}}
-      <i class="fas fa-lock-open" aria-hidden="true"></i>
+      <i class="fas fa-fw fa-lock-open" aria-hidden="true"></i>
     {{/button}}
   {{/slider-actions}}
 {{/slider}}
--- a/src/patternfly/components/Slider/examples/Slider.md
+++ b/src/patternfly/components/Slider/examples/Slider.md
@@ -6,7 +6,7 @@ cssPrefix: pf-c-slider
 ---
 
 ## Examples
-### Basic
+### Discrete
 ```hbs
 {{#> slider
   slider--value-min="0"
@@ -264,3 +264,4 @@ cssPrefix: pf-c-slider
 | `.pf-c-slider__value` | `<div>` | Initiates the slider value. |
 | `.pf-c-slider__actions` | `<div>` | Initiates the slider actions. |
 | `.pf-m-floating` | `.pf-c-slider__thumb` | Modifies the slider value to float above the thumb. |
+| `--pf-c-slider--value` | `.pf-c-slider` | Applies appropriate slider styles based on the current value. **Required** |
--- a/src/patternfly/components/Slider/examples/Slider.md
+++ b/src/patternfly/components/Slider/examples/Slider.md
@@ -1,0 +1,266 @@
+---
+id: 'Slider'
+beta: true
+section: components
+cssPrefix: pf-c-slider
+---
+
+## Examples
+### Basic
+```hbs
+{{#> slider
+  slider--value-min="0"
+  slider--value-max="8"
+  slider--value-now="5"
+  slider--attribute='style="--pf-c-slider--value: 62.5%;"'
+  }}
+  {{#> slider-main}}
+    {{> slider-rail}}
+    {{#> slider-steps}}
+      {{> slider-step slider-step--Left="0%" slider-step--label="0" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="12.5%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="25%" slider-step--label="2" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="37.5%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="50%" slider-step--label="4" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="62.5%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="75%" slider-step--label="6"}}
+      {{> slider-step slider-step--Left="87.5%"}}
+      {{> slider-step slider-step--Left="100%" slider-step--label="8"}}
+    {{/slider-steps}}
+    {{> slider-thumb}}
+  {{/slider-main}}
+{{/slider}}
+
+<br>
+
+{{#> slider
+  slider--value-min="0"
+  slider--value-max="100"
+  slider--value-now="50"
+  slider--attribute='style="--pf-c-slider--value: 50%;"'
+  }}
+  {{#> slider-main}}
+    {{> slider-rail}}
+    {{#> slider-steps}}
+      {{> slider-step slider-step--Left="0%" slider-step--label="0%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="25%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="50%" slider-step--label="50%" slider-step--IsActive="true"}}
+      {{> slider-step slider-step--Left="75%"}}
+      {{> slider-step slider-step--Left="100%" slider-step--label="100%"}}
+    {{/slider-steps}}
+    {{> slider-thumb}}
+  {{/slider-main}}
+{{/slider}}
+```
+
+### Continuous
+```hbs
+{{#> slider
+  slider--value-min="0"
+  slider--value-max="100"
+  slider--value-now="50"
+  slider--attribute='style="--pf-c-slider--value: 50%;"'
+  }}
+  {{#> slider-main}}
+    {{> slider-rail}}
+    {{> slider-thumb}}
+  {{/slider-main}}
+{{/slider}}
+```
+
+### Value input
+```hbs
+{{#> slider
+  slider--value-min="0"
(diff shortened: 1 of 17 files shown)


mcarrano (changes requested):
These look good @mcoker . I'll defer to @maryshak1996 and @mceledonia on the visual details, but I did wonder if we can refactor the examples a bit. IMO, the continuous slider is the most common use case. I would either not have an example labeled "Basic" and just have Continuous and Discrete as in the design spec or change the basic example to show a continuous slider with labels at either end as to me this would be the most simple/common. What do you think?

maryshak1996 (approved):

maryshak1996:
> These look good @mcoker . I'll defer to @maryshak1996 and @mceledonia on the visual details, but I did wonder if we can refactor the examples a bit. IMO, the continuous slider is the most common use case. I would either not have an example labeled "Basic" and just have Continuous and Discrete as in the design spec or change the basic example to show a continuous slider with labels at either end as to me this would be the most simple/common. What do you think?

@mcarrano I don't have a real preference as to which variant should be displayed first, but I think that you bring up a good point about the naming. I think the Continuous and Discrete is more accurate than Continuous and Basic (basic implies that it's the default choice, but I don't know that we want to be making that type of preference suggestion).

src/patternfly/components/Slider/slider.scss
```diff
@@ -0,0 +1,203 @@
...
+  --pf-c-slider__thumb--focus--BoxShadow: var(--pf-c-slider__thumb--BoxShadow--base);
+  --pf-c-slider__thumb--active--BoxShadow:
+    var(--pf-c-slider__thumb--BoxShadow--base),
+    0 0 2px 5px var(--pf-global--active-color--200);
+
+  // value
+  --pf-c-slider__value--MarginLeft: var(--pf-global--spacer--md);
+  --pf-c-slider__value--c-form-control--width-base: 3.5ch;
+  --pf-c-slider__value--c-form-control--width-chars: 2;
+  --pf-c-slider__value--c-form-control--Width: calc(var(--pf-c-slider__value--c-form-control--width-base) + (var(--pf-c-slider__value--c-form-control--width-chars) * 1ch));
```
mattnolting, line 53: I think you need to add `+ (var(--pf-global--spacer--sm) * 2)` or the equivalent of `--pf-c-form-control--inset--base`. A 100% value doesn't fit for me

src/patternfly/components/Slider/examples/Slider.md
```diff
@@ -0,0 +1,266 @@
...
+```hbs
+{{#> slider
+  slider--value-min="0"
+  slider--value-max="100"
+  slider--value-now="50"
+  slider--attribute='style="--pf-c-slider--value: 50%;"'
+  }}
+  {{#> slider-actions}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Minus"'}}
+      <i class="fas fa-minus" aria-hidden="true"></i>
```
mattnolting, line 178:
```suggestion
      <i class="fas fa-minus fa-fw" aria-hidden="true"></i>
```

christiemolloy (commented):
When i drag the slider circle, should all the numbers highlight like this?
Screen Shot 2020-12-02 at 11 08 34 AM

src/patternfly/components/Slider/examples/Slider.md
```diff
@@ -0,0 +1,266 @@
...
+    {{/button}}
+  {{/slider-actions}}
+  {{#> slider-main}}
+    {{> slider-rail}}
+    {{> slider-thumb}}
+  {{/slider-main}}
+  {{#> slider-actions}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Plus"'}}
+      <i class="fas fa-plus" aria-hidden="true"></i>
```
mattnolting, line 187:
```suggestion
      <i class="fas fa-plus fa-fw" aria-hidden="true"></i>
```

src/patternfly/components/Slider/examples/Slider.md
```diff
@@ -0,0 +1,266 @@
...
+    {{#> slider-value slider-value--modifier="pf-m-floating"}}
+      {{#> input-group}}
+        {{> form-control controlType="input" input="true" form-control--attribute=(concat 'disabled type="number" value="' slider--value-now '" aria-label="Slider value input"')}}
+        {{#> input-group-text input-group-text--modifier="pf-m-plain"}}%{{/input-group-text}}
+      {{/input-group}}
+    {{/slider-value}}
+  {{/slider-main}}
+  {{#> slider-actions}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Locked"'}}
+      <i class="fas fa-lock" aria-hidden="true"></i>
```
mattnolting, line 212:
```suggestion
      <i class="fas fa-lock fa-fw" aria-hidden="true"></i>
```

christiemolloy (commented):
Is the slider thumb meant to visually look like this? It looks different to the design

Screen Shot 2020-12-02 at 11 10 08 AM
Screen Shot 2020-12-02 at 11 10 37 AM

src/patternfly/components/Slider/slider.scss
```diff
@@ -0,0 +1,203 @@
...
+  height: var(--pf-c-slider__rail-track--Height);
+
+  &::before {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    content: "";
+    background:
```
mattnolting, line 87: Beautiful! 🤩

christiemolloy:
Looks like the little squares should actually be circles (based off the design)

Screen Shot 2020-12-02 at 11 12 00 AM

Screen Shot 2020-12-02 at 11 12 09 AM

christiemolloy:
The design seems to have a small margin-left (in orange) which I don't see in the slider.

Screen Shot 2020-12-02 at 11 14 18 AM
Screen Shot 2020-12-02 at 11 14 22 AM

mattnolting (commented):
Excellent work! Left a few comments.

mcoker:
> When i drag the slider circle, should all the numbers highlight like this?
> Screen Shot 2020-12-02 at 11 08 34 AM

@christiemolloy probably since you're basically making a selection dragging a static element like that. Same thing happens with the drawer splitter. Once the react component handles the interactivity, that shouldn't happen.

Screen Shot 2020-12-02 at 10 14 32 AM

christiemolloy:
The color for the thumbnail seems to be too dark in the implementation compared to whats in the design. I cant see the Marvel doc so not sure what it should be

Screen Shot 2020-12-02 at 11 18 29 AM

Screen Shot 2020-12-02 at 11 18 32 AM

mcoker:
> Is the slider thumb meant to visually look like this? It looks different to the design
> 
> Screen Shot 2020-12-02 at 11 10 08 AM
> 
> Screen Shot 2020-12-02 at 11 10 37 AM

@maryshak1996 can you verify? I got the styling from 

Screen Shot 2020-12-02 at 10 19 42 AM

src/patternfly/components/Slider/slider.scss
mcoker: Thanks! And this should still work when we upgraydd the slider to select a range, we'll just introduce more stops.

mcoker:
> Looks like the little squares should actually be circles (based off the design)
> 
> Screen Shot 2020-12-02 at 11 12 00 AM
> 
> Screen Shot 2020-12-02 at 11 12 09 AM

Nice catch, I didn't notice that. @maryshak1996 can you verify the ticks should be circles instead or do you want to leave them as squares?

mcoker:
> The design seems to have a small margin-left (in orange) which I don't see in the slider.
> 
> Screen Shot 2020-12-02 at 11 14 18 AM
> 
> Screen Shot 2020-12-02 at 11 14 22 AM

I talked with @maryshak1996 about this when I began development, and we agreed the tick would be flush left with the track instead of having a small indent. Mary would you mind verifying this looks as you'd expect?

christiemolloy:
Are you going to show an example with the remove button in the slider: https://www.sketch.com/s/6f81d099-0c11-407a-8962-5672df30dcf2/a/JZ2Qx7

christiemolloy:
In the "Actions" examples, do you think it would be beneficial to show these examples under "codepedent sliders" where the text input sits next to the action button ? https://www.sketch.com/s/6f81d099-0c11-407a-8962-5672df30dcf2/a/JZ2Qx7

mcoker:
> The color for the thumbnail seems to be too dark in the implementation compared to whats in the design. I cant see the Marvel doc so not sure what it should be
> 
> Screen Shot 2020-12-02 at 11 18 29 AM
> 
> Screen Shot 2020-12-02 at 11 18 32 AM

It's blue-500 from the screenshot in https://github.com/patternfly/patternfly/pull/3711#issuecomment-737337117. I'll let @maryshak1996 verify.

mcoker:
> Are you going to show an example with the remove button in the slider: https://www.sketch.com/s/6f81d099-0c11-407a-8962-5672df30dcf2/a/JZ2Qx7

@mcarrano @maryshak1996 what do you think? The actions examples (lock, plus/minus) just show an action you could put into the slider. The lock and plus/minus are not specific parts of this component, only the actions. You could replace those actions with any other sort of action you want, so I only showed a couple of examples with different actions and didn't think it was necessary to show the remove button. Maybe that would go in a future demo with the slider used in the context of a form.

maryshak1996:
@christiemolloy those are good catches (clearly stuff I didn't catch while trying to multitask haha). 
I checked my designs and the thumbnail color is too dark (I made a typo on my spec). The circle and the blue hover/drag outline should both use `blue-400` @mcoker do you want me to update my spec, or are you cool with just updating it for your implementation?
And to the comment about the circles vs rectangles for the tick marks -- I think that, if we could get the main bar itself to have rounded edges then yes, circles is the way to go like this: 
Screen Shot 2020-12-02 at 11 31 58 AM

src/patternfly/components/Slider/examples/Slider.md
```diff
@@ -0,0 +1,267 @@
...
+cssPrefix: pf-c-slider
+---
+
+## Examples
+### Discrete
+```hbs
+{{#> slider
+  slider--value-min="0"
+  slider--value-max="8"
+  slider--value-now="5"
```
christiemolloy, line 14: why "now" instead of a word like "current"

src/patternfly/components/Slider/examples/Slider.md
```diff
@@ -0,0 +1,267 @@
...
+  slider--attribute='style="--pf-c-slider--value: 62.5%;"'
```
christiemolloy, line 15: whats this for

mcarrano:
> @mcarrano @maryshak1996 what do you think? The actions examples (lock, plus/minus) just show an action you could put into the slider. The lock and plus/minus are not specific parts of this component, only the actions. You could replace those actions with any other sort of action you want, so I only showed a couple of examples with different actions and didn't think it was necessary to show the remove button. Maybe that would go in a future demo with the slider used in the context of a form.

Agree that it's not necessary to show this example. As long as you can put any action there the current example is fine.

src/patternfly/components/Slider/examples/Slider.md
```diff
@@ -0,0 +1,267 @@
...
+  }}
+  {{#> slider-main}}
+    {{> slider-rail}}
+    {{#> slider-steps}}
+      {{> slider-step slider-step--Left="0%" slider-step--label="0" slider-step--IsActive="true"}}
```
christiemolloy, line 20: does the left value always have to be in % or could it be numerical

christiemolloy: Could the left value and the label equal the same here?

src/patternfly/components/Slider/slider.scss
```diff
@@ -0,0 +1,204 @@
...
+  --pf-c-slider__rail-track--before--fill--BackgroundColor--gradient-stop: var(--pf-c-slider--value);
+
+  // steps
+  --pf-c-slider__steps--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-slider__steps--Height: var(--pf-c-slider__steps--FontSize);
+
+  // step tick
+  --pf-c-slider__step-tick--Top: var(--pf-global--spacer--md);
+  --pf-c-slider__step-tick--Width: #{pf-size-prem(4px)};
+  --pf-c-slider__step-tick--Height: #{pf-size-prem(4px)};
```
christiemolloy, line 18: should the value for height be the variable for width since they're proportional

src/patternfly/components/Slider/slider.scss
```diff
@@ -0,0 +1,204 @@
...
+
+  // step label
+  --pf-c-slider__step-label--TranslateX: -50%;
+  --pf-c-slider__step-label--Top: calc(var(--pf-global--spacer--xl) + var(--pf-c-slider__rail-track--Height));
+  --pf-c-slider__step--first-child__step-label--TranslateX: 0;
+  --pf-c-slider__step--last-child__step-label--TranslateX: -100%;
+
+  // thumb
+  --pf-c-slider__thumb--Top: calc(var(--pf-c-slider__rail-track--Height) / 2 + var(--pf-global--spacer--md));
+  --pf-c-slider__thumb--Width: #{pf-size-prem(16px)};
```
christiemolloy, line 33: should the value for height be the variable for width since they're proportional

christiemolloy (changes requested):

src/patternfly/components/Slider/examples/Slider.md
mcoker: `aria-valuenow` - https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html#rps_label

Though it's just an internal handlebars var so we can call it whatever you'd like to change it.

src/patternfly/components/Slider/examples/Slider.md
mcoker: It defines:

1. the track fill gradient stops to create the blue fill color - https://github.com/patternfly/patternfly/blob/a296e201f88e918a7ac0e335ec65c95b8b9978d2/src/patternfly/components/Slider/slider.scss#L9
1. positions the thumb - https://github.com/patternfly/patternfly/blob/a296e201f88e918a7ac0e335ec65c95b8b9978d2/src/patternfly/components/Slider/slider.scss#L36
1. positions the floating value input that displays over the thumb - https://github.com/patternfly/patternfly/blob/a296e201f88e918a7ac0e335ec65c95b8b9978d2/src/patternfly/components/Slider/slider.scss#L57

Basically all parts of the UI that update as the value of the slider changes so that the react component only has to update a single var and the slider CSS handles the layout.

src/patternfly/components/Slider/examples/Slider.md
mcoker: The left position will always have to be a percentage even if the value of the step/tick/label is a different number, because it's used to declare the `left` property in the layout, so it will need to be a percentage.

src/patternfly/components/Slider/slider.scss
mcoker: It could, though it creates the potential breaking change where, if we make the track 10px tall and we want the tick to be 4x10, if we change `--pf-c-slider__step-tick--Height: var(--pf-c-slider__step-tick--Width);` to `--pf-c-slider__step-tick--Height: 10px;`, then since `--pf-c-slider__step-tick--Width` used to control `--pf-c-slider__step-tick--Height`, and it no longer does, that could be a breaking change. Similar to how we have hover/focus/active, and focus/active aren't defined by the hover var, even though they're all the same value.

Though in this case, I don't think it's likely, but it doesn't cost much to keep them independent.

src/patternfly/components/Slider/slider.scss
mcoker: Same idea as https://github.com/patternfly/patternfly/pull/3711#discussion_r534333133

mcoker:
@maryshak1996 @christiemolloy updated the thumb color, outline, active/filled tick colors, and made the ticks circles.

jessiehuff (approved):
This looks good to me! Great job on this!! 😄 🎉

src/patternfly/components/Slider/examples/Slider.md
christiemolloy: oh right, thats good with me didnt realize it was for aria

src/patternfly/components/Slider/slider.scss
christiemolloy: ok makes sense lets keep them

src/patternfly/components/Slider/slider-main.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-slider__main{{#if slider-main--modifier}} {{slider-main--modifier}}{{/if}}"
```
christiemolloy, line 1: i think you forgot to add __main to the docs

src/patternfly/components/Slider/slider.scss
```diff
@@ -0,0 +1,206 @@
...
+  --pf-c-slider__thumb--Height: #{pf-size-prem(16px)};
+  --pf-c-slider__thumb--Left: var(--pf-c-slider--value);
+  --pf-c-slider__thumb--BackgroundColor: var(--pf-global--primary-color--100);
+  --pf-c-slider__thumb--TranslateX: -50%;
+  --pf-c-slider__thumb--TranslateY: -50%;
+  --pf-c-slider__thumb--BorderRadius: var(--pf-global--BorderRadius--lg);
+  --pf-c-slider__thumb--BoxShadow--base:
+    0 0 0 2px #fff,
```
christiemolloy, line 42: instead of #fff should this use the variable we have for white

mcoker: doh!

mcoker:
@christiemolloy updated

src/patternfly/components/Slider/slider.scss
mattnolting: Screen Shot 2020-12-02 at 4 01 04 PM

src/patternfly/components/Slider/examples/Slider.md
mattnolting: Screen Shot 2020-12-02 at 4 01 55 PM

mattnolting (approved):
🚢  it! 🎉

christiemolloy (approved):
nice job turning this around so quickly

mcarrano (approved):
Looks great. Thanks @mcoker !

PR #3714, opened 2020-12-01
fix(card): allow card header to wrap

### LINKED ISSUE #3713: Card header doesn't wrap
When there's a long string in the header or items in the action area (such as buttons), a card's header does not wrap. In Cockpit, we sometimes have buttons in card action areas (instead of hiding common actions under a ⋮ menu and have hit this problem.

Thankfully, it's as simple as adding either `flex-wrap: wrap` changing `flex-direction: row` to `flex-flow: row wrap`.

garrett:
fixes #3713

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -115,7 +115,7 @@
 
 .pf-c-card__header {
   display: flex;
-  flex-direction: row;
+  flex-flow: row wrap;
   align-items: center;
 
   .pf-c-card__title {


mcoker:
@garrett thanks for the PR! With this change, our basic example with a long title and actions changes from:

Screen Shot 2020-12-01 at 9 25 04 AM

to

Screen Shot 2020-12-01 at 9 25 10 AM

That's not the intended design, and would be a breaking change. Can you elaborate more on the initial problem you're trying to solve in https://github.com/patternfly/patternfly/issues/3713? cc @mcarrano 

> When there's a long string in the header or items in the action area (such as buttons), a card's header does not wrap. In Cockpit, we sometimes have buttons in card action areas (instead of hiding common actions under a ⋮ menu and have hit this problem.

garrett:
We were using the actions area for buttons. In some cases, the buttons would shove the consent to be wider than the screen allowed, such as for very small screens and mobile devices. Instead, we needed the buttons to wrap under the title so that everything would fit on screen.

Short sumamry: non-wrapping content in the actions area force the card to be too wide. ([s stated in [the issue](https://github.com/patternfly/patternfly/issues/3713): We're using action buttons instead of the ⋮ menu.)

I had a suggested local Cockpit change, but @KKoukiou wanted me to upstream it. (Perhaps it might still be better as a local change, since the widgets we're using on are subtly different from the example PF widgets.)

mcoker:
@garrett gotcha! I'm sure you've considered it, but allow the `__actions` element to wrap could be an option, though then the actions wouldn't wrap under the title, and it sounds like that's what you're going for.

A local change sounds appropriate here. Though I can see a case for stacking the title + actions at a breakpoint in some cards, which is something we could introduce a modifier to do.

I'd be curious to hear what @mcarrano thinks about either the general layout or case for adding an enhancement allowing the card actions to wrap - would you be able to share designs that show your ideal layouts as the page resizes?

mcarrano:
Yes, I'd be interested in seeing the designs @garrett . Must the actions be in the header in this case or could they just be contained in the body of the card?

mcoker:
@garrett just checking in - did you want to leave this PR open and continue working toward a solution?

garrett:
Yep, let's keep working toward a solution.   This PR as-is doesn't work for the default case. Sorry.

I have been working on a few other things in the Cockpit codebase meanwhile, so I haven't gotten back to this yet. Apologies.

Right now, where we have this problem, we're doing a flex wrap, but it's not ideal. We could better solve it with a vertical gap on the flex, but WebKit currently does not support gap in anything other than grid. While it _works_, it's ugly in Safari and Epiphany. But I guess that might be acceptable enough, as this is basically all about handling what would otherwise be an errored state in the layout. 

In other words: 
1. By default it's too wide and the UI chopped off and/or runs off the page (depending on things, such as overflow)
2. Wrapping with bad spacing is preferable to breaking the page or chopping off elements
3. Wrapping with proper spacing, where available, is even better

So we could, in this case, consider using better spacing via row-gap in a flex as a "progressive enhancement" Chromium and Gecko browsers get, but not WebKit — yet. Thankfully, it's _supposedly_ going into Safari's technology preview, so it should land in the next major version. But that could be anywhere from _weeks_ to _months_ away. Meanwhile, I'm running Epiphany Technology Preview here and don't see [gap support in MDN demos](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) yet (except for grid), so I'm somewhat doubtful about a sooner-rather-than-later timeframe.

One alternative could be using grid in a (by default) 2D manner, but we probably wouldn't want the elements stretched to be the same width. Unless that width is basically 100% — then the heading and the actions would basically be stacked.

Of course, this should work with the checkbox + ⋮ menu by default in PF4, which this PR doesn't work well with, as well as the heading + action button(s) of varying length (without the checkbox and menu). And I've been talking about the latter (what Cockpit does), not the former (what PatternFly does by default). Ideally, we'd want something that works for both cases without having to resort to an option (or a custom override in Cockpit's codebase, which is what we currently have).

Since today's my last work day out of the year, I probably won't get around to this until January.

garrett:
Converted to draft to make it clear that this is a WIP preview and isn't to be merged. Hopefully either I or someone else can get something that works for all of us and update this PR with another preview, probably in January.

mcoker:
@garrett would the negative margin trick work here as a temporary solution until flex gap support is available? i.e., give the immediate descendants `__header` a `margin-bottom` set to the gap value you want, then give `__header` a negative margin equal to the gap value.

```
__header {
  --gap: var(--pf-global--spacer--sm);

  margin-bottom: calc(var(--gap) * -1);

  > * {
    margin-bottom: var(--gap);
  }
}
```

garrett:
@mcoker: Yeah, something like that could work! Good idea.

---

FWIW (probably to people who aren't @mcoker, as I'm guessing he knows this too): I would _ordinarily_ suggest we'd wrap it in something like `@supports (gap: 1rem) {}`, as `> *` is considered somewhat "expensive*" (as CSS parses from right to left)… except WebKit _thinks_ it supports gap already, as it supports "gap" in grid, but nowhere else. I don't think there's a way to target gap-in-flex with `@supports`, sadly.

I can't wait until all browsers fully support gap in flex & columns — and also subgrid. And then time passes and we can just use it all, everywhere, instead of having to resort to hacks still. I'm also glad we don't have to resort to many hacks these days! 🥳 

(* However, "expensively slow" in CSS terms is still extremely quick compared to most anything in JavaScript. :wink:)

References:
- The `@supports (gap) {}` problem, explained in more detail (for anyone interested): https://ishadeed.com/article/flexbox-gap/
- RTL selector matching in CSS, explained: https://stackoverflow.com/questions/5797014/why-do-browsers-match-css-selectors-from-right-to-left/5813672#5813672 (Basically: It's a quicker way to throw out everything an element _isn't_.)

---

Have happy end-of-year/start-of-next-year celebrations, everyone! 🎉

garrett:
1. This isn't stale, bot.

2. Safari finally supports gap in 14.1 on desktop and 14.5 on mobile (as of the end of April for both): https://caniuse.com/flexbox-gap, so that's all browsers. (Gap support for flex already landed a year ago in Chrome and nearly three years ago in Firefox.) So we could use gap in flex instead of hacky margins now.

mcarrano:
@garrett looks like discussion on this kind of fizzled out at the end of last year. If it's still a valid issue, perhaps you, @mcoker , and I should get together to discuss a path forward. Let me know if you want me to set something up.

garrett:
I think this is still something relevant to Cockpit, at least.

It fizzled out because we've been working on lots of other things (this is especially true for me, as I'm the only designer in Cockpit now) and we currently have a workaround in place for Cockpit itself... but it would still be great to have something upstream in PatternFly.

Hopefully we can move forward with this sometime soon.

mcarrano:
@garrett @mcoker I do want to revisit the solution here and question whether we can accomplish this without a change to the wrapping behavior. As I understand it, the Card has a header and a title area that can be separate (https://www.patternfly.org/v4/components/card/design-guidelines#elements). Could the custom actions live in the title area so they fall underneath the header? I don't think we intended that the header actions (checkbox and kabob) would be customizable to contain buttons or other elements. But if we do, we should rethink the design. @mceledonia any thoughts on this?

mceledonia:
@mcarrano I wouldn't be opposed to rethinking the design, I can see a more flexible header being useful especially considering dashboards. A while back I had mockups for an OpenShift dashboard card which featured a plain variant dropdown select menu for choosing different scopes for the data it was visualizing.

That being said I don't think there would be a need for much rethinking, the actions would behave as the current kebab/checkbox do today and cut into the title space, so having a wrap solution for titles that need 2+ lines makes sense to me!

mcoker:
@garrett to confirm, are you going for something like this? https://codesandbox.io/s/still-dust-26393?file=/style.css

One thing to note is that we don't add support for CSS features that aren't supported by the browsers we support. And we support the latest 2 major versions of the following browsers - https://github.com/patternfly/patternfly#browser-support

So we're holding off on using flex gap until Safari 15 is released. If we release this in PF before then, we will need to make it work without `gap`.

garrett:
@mcoker: Yep! That codesandbox demo looks like what we're doing in Cockpit.

Bummer about Safari 15 not being released yet. I guess it'll hopefully arrive soon? I saw talk about Safari 16 beta (IIRC?) on Twitter this week.

garrett:
Not stale. 

We were waiting for Safari 15 being released. As Safari 15 was released September 20, does this mean we can now use the gap property in PatternFly, even in flex? https://caniuse.com/?search=flex%20gap

If so, we could adjust this PR to use gap and possibly land it or something like it in the near future.

garrett:
Still not stale...

mcoker:
Closing in favor of https://github.com/patternfly/patternfly/pull/5972

PR #3716, opened 2020-12-01
feat(masthead): added masthead component

### LINKED ISSUE #3682: Masthead generic scaffolding
General structure
```
.pf-c-masthead
  .pf-c-masthead__toggle
  .pf-c-masthead__brand
  .pf-c-masthead__nav
  .pf-c-masthead__content
```

Modifiers
```
.pf-c-masthead.pf-m-{inline/stack} 
```
• `-split/inline` would be applied at a predefined breakpoint (in React)
• a react prop would be available to prohibit the default functionality like, `noSplit` or `stacked`

Can switch between stacked layout and horizontal

Reference: https://github.com/patternfly/patternfly/pull/3597

mattnolting:
closes #3682 

Component preview: https://patternfly-pr-3716.surge.sh/components/masthead
Page preview: https://patternfly-pr-3716.surge.sh/components/page/html-demos/untitled-example/


mceledonia (approved):

mcarrano (commented):
@mattnolting Not sure what the scope of the PR is intended to be, but seems like a good start towards the larger masthead refactoring we spoke about. I have a few questions:

* For the new Masthead component, it's not clear to me that we should offer the light variants as we don't currently use them and it might just be inviting inconsistency. @mceledonia @mcoker what do you think?

* For the demo, I'm wondering if you should show a more full featured demo, i.e. adding in the utility items and possible a demo that includes a context selector. Maybe that's out of scope here, but wanted to see what your thoughts are.

* Now that we've done this, it's not clear how this co-exists with the existing page layouts. Is the new masthead an opt-in? Are changes to the Page component needed in parallel to adopt this?

mattnolting:
@mcarrano great questions, thanks!

> For the new Masthead component, it's not clear to me that we should offer the light variants as we don't currently use them and it might just be inviting inconsistency. @mceledonia @mcoker what do you think?

Forward looking, the same features have been requested for other components (`tabs`, `toggle-group`, `nav`, etc). They aren't necessary for this PR.

> For the demo, I'm wondering if you should show a more full featured demo, i.e. adding in the utility items and possible a demo that includes a context selector. Maybe that's out of scope here, but wanted to see what your thoughts are.

I think that's outside of the scope for this PR. The demo is really just to fit the masthead in context and make some preliminary adjustments in preparation for other components. I can comment that remove/comment that code until we're ready to fit the `toolbar` and/or `horizontal-nav`.

> Now that we've done this, it's not clear how this co-exists with the existing page layouts. Is the new masthead an opt-in? Are changes to the Page component needed in parallel to adopt this?

Yes, it will need to be an opt in, much like composable tables. Additions will need to be made to page to support `masthead` in `page-header`, but all existing code and functionality remains unaltered.

src/patternfly/components/Masthead/masthead-brand.hbs
```diff
@@ -0,0 +1,15 @@
+<{{#if masthead-brand--type}}{{masthead-brand--type}}{{else}}a{{/if}} class="pf-c-masthead__brand{{#if masthead-brand--modifier}} {{masthead-brand--modifier}}{{/if}}"
+  {{#if masthead-brand--attribute}}
+    {{masthead-brand--attribute}}
```
mcoker, line 3: Just merged a PR to fix this in the new component template

```suggestion
    {{{masthead-brand--attribute}}}
```

src/patternfly/components/Masthead/masthead-brand.hbs
```diff
@@ -0,0 +1,15 @@
+<{{#if masthead-brand--type}}{{masthead-brand--type}}{{else}}a{{/if}} class="pf-c-masthead__brand{{#if masthead-brand--modifier}} {{masthead-brand--modifier}}{{/if}}"
+  {{#if masthead-brand--attribute}}
```
mcoker, line 2: nit - typically we put `--attribute` as the last thing in the inline attrs, so that whatever is passed there would override anything else the template might add. So the `--attribute` block would come after `--href` block.

src/patternfly/components/Masthead/masthead-content.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="pf-c-masthead__content{{#if masthead-content--modifier}} {{masthead-content--modifier}}{{/if}}"
+  {{#if masthead-content--attribute}}
+    {{masthead-content--attribute}}
```
mcoker, line 3:
```suggestion
    {{{masthead-content--attribute}}}
```

src/patternfly/components/Masthead/masthead-main.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-masthead__main{{#if masthead-main--modifier}} {{masthead-main--modifier}}{{/if}}"
+  {{#if masthead-main--attribute}}
+    {{masthead-main--attribute}}
```
mcoker, line 3:
```suggestion
    {{{masthead-main--attribute}}}
```

src/patternfly/components/Masthead/masthead.hbs
```diff
@@ -0,0 +1,9 @@
+<header class="pf-c-masthead{{#if masthead--modifier}} {{masthead--modifier}}{{/if}}"
+  {{#if masthead--attribute}}
+    {{masthead--attribute}}
```
mcoker, line 3:
```suggestion
    {{{masthead--attribute}}}
```

src/patternfly/components/Masthead/masthead.hbs
```diff
@@ -0,0 +1,9 @@
...
+  {{/if}}
+  {{#if masthead--id}}
```
mcoker, line 5: same nit as above, this would go before `--attribute`

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,200 @@
...
+
+  > :first-child {
+    padding-left: var(--pf-c-masthead--first-child--PaddingLeft);
+  }
+
+  > :last-child {
+    padding-right: var(--pf-c-masthead--last-child--PaddingRight);
+  }
+
+  @include pf-t-dark; // force the container to follow the dark theme
```
mcoker, line 98: We usually make this the first thing in a block, after any custom properties, but before any property declarations.

src/patternfly/components/Masthead/examples/masthead.md
```diff
@@ -0,0 +1,140 @@
...
+  {{#> masthead-main}}
+    {{#> masthead-brand}}
+      Logo
+    {{/masthead-brand}}
+  {{/masthead-main}}
+  {{> masthead-content}}
+{{/masthead}}
+```
+
+### Light variant
```
mcoker, line 61: Should this title be "Light 100 variant"? Or maybe we could rename the modifier to `pf-m-light` to align with the page section variation? Or just leave it as it is?

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,200 @@
...
+    --pf-c-masthead__main--Order: var(--pf-c-masthead--m-display-inline__main--Order);
+    --pf-c-masthead__main--FlexBasis: var(--pf-c-masthead--m-display-inline__main--FlexBasis);
+    --pf-c-masthead__main--BorderBottom: var(--pf-c-masthead--m-display-inline__main--BorderBottom);
+    --pf-c-masthead__main--PaddingRight: var(--pf-c-masthead--m-display-inline__main--PaddingRight);
+    --pf-c-masthead__main--PaddingLeft: var(--pf-c-masthead--m-display-inline__main--PaddingLeft);
+    --pf-c-masthead__content--PaddingRight: var(--pf-c-masthead--m-display-inline__content--PaddingRight);
+    --pf-c-masthead__content--PaddingLeft: var(--pf-c-masthead--m-display-inline__content--PaddingLeft);
+  }
+
+  > :first-child {
```
mcoker, line 90: I wonder if this is going to end up being fragile, like the table's first/last child selectors for insets. Seems like we could end up adding things that are hidden or shouldn't take on that spacing and run into issues.

mcoker (commented):
Looks really good! Nice and simple, but flexible and capable. Left some comments for review.

src/patternfly/components/Masthead/masthead.scss
mattnolting: Great point, let me work on that a bit!

src/patternfly/components/Masthead/masthead-brand.hbs
```diff
@@ -0,0 +1,15 @@
+<{{#if masthead-brand--type}}{{masthead-brand--type}}{{else}}a{{/if}} class="pf-c-masthead__brand{{#if masthead-brand--modifier}} {{masthead-brand--modifier}}{{/if}}"
+  {{#if masthead-brand--href}}
+    href="{{masthead-brand--href}}"
+  {{else}}
```
mcoker, line 4: This creates a problem if `masthead-brand--type` is passed, you can't remove `href`. An obvious solution isn't sticking out so I'd be in favor of just making this work for our default use case and update it later if we need to.

Also should the default be `href="/"`, so that it's a location? `#` will add that to the address bar, too.

src/patternfly/components/Masthead/masthead-content.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="pf-c-masthead__content{{#if masthead-content--modifier}} {{masthead-content--modifier}}{{/if}}"
+  {{#if masthead-content--attribute}}
+    {{{masthead-content--attribute}}}
+  {{/if}}>
+  {{#if @partial-block}}
+    {{> @partial-block}}
+  {{else}}
+    <span>Start content</span>
+    <span class="pf-u-ml-auto">End content</span>
```
mcoker, line 9: We probably shouldn't include utilities in the component examples, otherwise people might think it's part of the component. As long as this is just to show the size of the `__content` element, I think it's OK to remove it.

src/patternfly/components/Masthead/masthead-toggle.hbs
```diff
@@ -0,0 +1,16 @@
+<span class="pf-c-masthead__toggle{{#if masthead-toggle--modifier}} {{masthead-toggle--modifier}}{{/if}}"
+  {{#if masthead-toggle--attribute}}
+    {{{masthead-toggle--attribute}}}
+  {{/if}}>
+  {{#if @partial-block}}
+    {{> @partial-block}}
+  {{else}}
+    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Masthead toggle"'}}
```
mcoker, line 8: Technically this will need `aria-expanded`, too. I expect it might be fair for someone to come here and copy/paste this code into their project? I wonder if the `aria-label` should be "Global navigation", too.

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,234 @@
...
+  --pf-c-masthead--m-display-inline__content--c-nav--m-horizontal--MarginLeft: 0;
+
+  // main border
+  --pf-c-masthead__main--before--Right: calc(var(--pf-c-masthead--inset) * -1);
+  --pf-c-masthead__main--before--Left: calc(var(--pf-c-masthead--inset) * -1);
+  --pf-c-masthead__main--before--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-masthead__main--before--BorderBottomColor: var(--pf-global--palette--black-600);
+
+  // toggle
+  --pf-c-masthead__toggle--MarginRight: calc(var(--pf-global--gutter--md) / 2);
```
mcoker, line 83: Can you explain why we're using `--pf-global--gutter--md / 2`?

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,234 @@
...
+
+  // light 100
+  --pf-c-masthead--m-light-100--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-masthead--m-light-100__main--BorderBottomColor: var(--pf-global--palette--black-200);
```
mcoker, line 87: In the page component, turning a page section white is `.pf-m-light`. Then we have `.pf-m-color-scheme--light-300` in tabs, though that's arguably a different use case.

And we have a drawer issue this sprint (https://github.com/patternfly/patternfly/issues/3813) to enable the grey background on the panel/content. I'd like to see if we can create something consistent. I like this. Maybe just `pf-m-light` and `pf-m-light-200` to align with the page? cc @mcarrano @mceledonia

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,234 @@
...
+
+  // set layout to stack by default
+  @include pf-c-masthead--m-display-stack;
+
+  // set layout to inline at medium breakpoint
+  @media screen and (min-width: $pf-global--breakpoint--md) {
+    @include pf-c-masthead--m-display-inline;
+  }
+
+  @media screen and (min-width: $pf-global--breakpoint--xl) {
```
mcoker, line 101: if all you're doing is redefining a var, this should go up to the block where they're originally defined

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  &.pf-m-light-200 {
+    @include pf-t-light;
+
+    --pf-c-masthead--BackgroundColor: var(--pf-c-masthead--m-light-200--BackgroundColor);
+    --pf-c-masthead__main--BorderBottomColor: var(--pf-c-masthead--m-light-200__main--BorderBottomColor);
+  }
+
+  &.pf-m-page-insets {
```
mcoker, line 130: should this be the default, assuming the default use case of the masthead is in the page?

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// toggle
+.pf-c-masthead__toggle {
+  align-self: center;
+  order: 0;
+  margin-right: var(--pf-c-masthead__toggle--MarginRight);
+
+  .pf-c-button {
+    margin-left: calc(var(--pf-c-button--PaddingLeft) * -1);
```
mcoker, line 206: This works but usually we'll just offset the component element (`.pf-c-masthead__toggle`). The reasoning was, while this updates automatically if we change `--pf-c-button--PaddingLeft`, it breaks if we change `--pf-c-button--PaddingLeft` (we changed some var names and broke some rules like this). And it's more inline with keeping the styling for a component in its component. And for having a 1:1 with a style property to component var (ie, `margin-left: var(--pf-c-masthead__toggle--MarginLeft);` that's themable/overridable.

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,234 @@
...
+  }
+}
+
+// stylelint-disable no-duplicate-selectors
+.pf-c-masthead {
```
mcoker, line 211: 👍👍👍

src/patternfly/components/Page/page.scss
```diff
@@ -202,6 +206,17 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
   @media screen and (min-width: $pf-global--breakpoint--lg) {
     grid-template-columns: auto 1fr auto;
   }
+
+  .pf-c-masthead {
+    --pf-c-masthead--inset: var(--pf-c-page__header--c-masthead--inset);
```
mcoker, line 211: Oh I see what you did here in setting the insets. I would have set the page insets as default, or just used `.pf-m-page-insets` instead of setting the vars here, too. I would have set the toggle button font-size below in the masthead component, too, as the default. I'm curious what the use cases for the masthead are outside of our page or why you would want to change insets or font size?

src/patternfly/components/Masthead/examples/masthead.md
```diff
@@ -0,0 +1,204 @@
...
+```hbs
+{{#> masthead masthead--id="basic-masthead-with-horizontal-nav"}}
+  {{> masthead-toggle}}
+  {{#> masthead-main}}
+    {{#> masthead-brand}}
+      Logo
+    {{/masthead-brand}}
+  {{/masthead-main}}
+  {{#> masthead-content}}
+    <nav class="pf-c-nav pf-m-horizontal pf-m-scrollable" aria-label="Global">
```
mcoker, line 57: just a note, we'll probably want to remove this, or if we keep it, use HBS. My vote would be to remove, and we'll show it in a page/nav demo.

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,220 @@
...
+  --pf-c-masthead__content--MarginLeft: var(--pf-c-masthead--m-display-inline__content--MarginLeft);
+  --pf-c-masthead__content--c-nav--m-horizontal--MarginRight: var(--pf-c-masthead--m-display-inline__content--c-nav--m-horizontal--MarginRight);
+  --pf-c-masthead__content--c-nav--m-horizontal--MarginLeft: var(--pf-c-masthead--m-display-inline__content--c-nav--m-horizontal--MarginLeft);
+}
+
+.pf-c-masthead {
+  --pf-c-masthead--PaddingLeft: var(--pf-c-masthead--inset);
+  --pf-c-masthead--PaddingRight: var(--pf-c-masthead--inset);
+  --pf-c-masthead--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);
+  --pf-c-masthead--inset: var(--pf-global--spacer--lg);
```
mcoker, line 40: looks like you can remove this

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,220 @@
...
+  --pf-c-masthead__main--before--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-masthead__main--before--BorderBottomColor: var(--pf-global--palette--black-600);
+
+  // toggle
+  --pf-c-masthead__toggle--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-masthead__toggle--MarginLeft: calc(var(--pf-global--spacer--md) * -1); // offset by button padding
+
+  // light 100
+  --pf-c-masthead--m-light--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-masthead--m-light__main--BorderBottomColor: var(--pf-global--palette--black-200);
```
mcoker, line 88: We have a border-color var - do you think this is a good use of it?

```suggestion
  --pf-c-masthead--m-light__main--BorderBottomColor: var(--pf-global--BorderColor--300);
```

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,220 @@
...
+
+  // light 200
+  --pf-c-masthead--m-light-200--BackgroundColor: var(--pf-global--BackgroundColor--200);
+  --pf-c-masthead--m-light-200__main--BorderBottomColor: var(--pf-global--palette--black-300);
```
mcoker, line 92: Same here - would the border color var work?

```suggestion
  --pf-c-masthead--m-light-200__main--BorderBottomColor: var(--pf-global--BorderColor--100);
```

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,220 @@
...
+
+.pf-c-masthead {
+  --pf-c-masthead--PaddingLeft: var(--pf-c-masthead--inset);
+  --pf-c-masthead--PaddingRight: var(--pf-c-masthead--inset);
+  --pf-c-masthead--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);
+  --pf-c-masthead--inset: var(--pf-global--spacer--lg);
+
+  // insets
+  --pf-c-masthead--inset: var(--pf-global--spacer--md); // match page insets
+  --pf-c-masthead--xl--inset: var(--pf-global--spacer--lg); // match page insets
```
mcoker, line 44: This is never used...

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -0,0 +1,220 @@
...
+  --pf-c-masthead--m-display-stack__content--c-nav--m-horizontal--MarginRight: calc(var(--pf-c-masthead--inset) * -1);
+  --pf-c-masthead--m-display-stack__content--c-nav--m-horizontal--MarginLeft: calc(var(--pf-c-masthead--inset) * -1);
+
+  // inline
+  --pf-c-masthead--m-display-inline--FlexWrap: nowrap;
+  --pf-c-masthead--m-display-inline__main--Order: 0;
+  --pf-c-masthead--m-display-inline__main--FlexBasis: auto;
+  --pf-c-masthead--m-display-inline__main--PaddingTop: 0;
+  --pf-c-masthead--m-display-inline__main--PaddingBottom: 0;
+  --pf-c-masthead--m-display-inline__main--MarginRight: calc(var(--pf-global--gutter--md) / 2); // split gutter
```
mcoker, line 67: Why does this use the gutter global var over the large spacer? Or is it meant to be half of whatever the current inset is? (24px > 1200px viewport, 16px < 1200px)?

mcoker (changes requested):

src/patternfly/components/Page/page.scss
```diff
@@ -154,13 +154,19 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
...
   }
 
   // Wizard main section
   --pf-c-page__main-wizard--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
   --pf-c-page__main-wizard--BorderTopColor: var(--pf-global--BorderColor--100);
   --pf-c-page__main-wizard--BorderTopWidth: var(--pf-global--BorderWidth--sm);
 
+  // masthead
+  --pf-c-page__header--c-masthead--inset: var(--pf-global--spacer--md);
+  --pf-c-page__header--c-masthead--xl--inset: var(--pf-global--spacer--lg);
```
mcoker, line 167: are these page component insets still needed since the masthead default insets match the page?

src/patternfly/components/Page/page.scss
```diff
@@ -154,13 +154,19 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
...
+  --pf-c-page__header--c-masthead-toggle--c-button--FontSize: var(--pf-global--FontSize--2xl);
```
mcoker, line 168: should this just be the default? It seems reasonable as a default, most mobile toggles are pretty big. Or is there a reason we want the default to be something different? I would expect the masthead's default behavior to work like it should in the page component, but I may be unaware of other use cases for it.

mcoker (approved):

Issue #3718, opened 2020-12-01
Panel-left drawer content shifts when panel width changes


mcoker:
Reported in https://github.com/patternfly/patternfly-react/pull/5193

If you update `--pf-c-drawer__panel--FlexBasis` on the panel-left variation, the content shifts a little bit. Looks like we need to animate the `flex-basis` property. This doesn't work in Safari, but it's due to a bug, and is [fixed in TP 116](https://webkit.org/blog/11348/release-notes-for-safari-technology-preview-116/)

cc @kmcfaul

PR #3719, opened 2020-12-01
fix(drawer): kept content from shifting when panel resizes

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3718


mattnolting (approved):
LPTM

Issue #3437, opened 2020-08-27
Kebab has more built in padding than other toolbar icons


katierik:
The kebab is wider than the other toolbar icons. I assume it is because it is a dropdown. However it is causing problems in implementation as it is difficult to get proper spacing as defined in the Toolbar design documentation.

![image](https://user-images.githubusercontent.com/23172038/91466733-8c335b80-e85d-11ea-96e2-929a652fcc86.png)

![image](https://user-images.githubusercontent.com/23172038/91466761-96555a00-e85d-11ea-8443-6332646b77f2.png)

PR #3720, opened 2020-12-02
fix(dropdown): Increase spacer

rebeccaalpert:
So this bug was a little odd - the examples Katie noted were React examples, so those will need to be updated.

I updated the spacer to reflect what other buttons are using per advice from @mcarrano and @mceledonia. I was also told to use an overflow menu in the toolbar, but it looks like we're already using it for the Core examples. I think React will need to be updated to reflect that suggestion and fix the remaining issues there.

Fixes https://github.com/patternfly/patternfly/issues/3437.

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -1,9 +1,9 @@
 .pf-c-dropdown {
   // Toggle
   --pf-c-dropdown__toggle--PaddingTop: var(--pf-global--spacer--form-element);
-  --pf-c-dropdown__toggle--PaddingRight: var(--pf-global--spacer--sm);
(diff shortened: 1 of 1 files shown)


mcoker:
Nice find @rebeccaalpert! @mceledonia @mcarrano should we also update the plain options menu toggle to match?

mcoker:
@rebeccaalpert after talking with @mceledonia and @mcarrano we decided that the plain toggle variation in dropdown and options menu is what should be updated to match the plain button's width. Looking at it briefly, I think we'll want to add new vars for the plain toggle's left/right padding set to `--pf-global--spacer--md` (to match the button's padding), then set the toggles' base left/right padding vars to the new vars. It looks like we will also need to remove the `min-width` for those toggles. To keep it from being a breaking change, I think we can just set the existing min-width var to "0".

Issue #3722, opened 2020-12-03
Update accordion expanded content color


mcoker:
The accordion's expanded content color uses `var(--pf-global--secondary-color--100)` when it should use `var(--pf-global--Color--200)`.

PR #3723, opened 2020-12-03
fix(accordion): updated expanded content color to use correct global var

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3722


mattnolting (approved):
LPTM 👍

Issue #3735, opened 2020-12-07
Add patternfly-base-no-reset


mcoker:
This is for react apps that want to exclude the global reset styles, but don't want to import all of `patternfly-no-reset.css`.

PR #3736, opened 2020-12-07
feat(global): made variation of patternfly-base.css w/o global reset

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3735


redallen (approved):
Beautiful!

Issue #3675, opened 2020-11-13
Menu updates


mcoker:
In the 1st example, the 2nd disabled item has an aria attr in the class value.

* Review block elements that are children of `<button>` elements and convert to spans and apply block styling if needed.
* Remove selected icon on non-selected items to match the select component and react implementation.
* Investigate need for disabled class on the `<li>` for items
  * Should that only be on the interactive children?
  * If we leave it, does that impact things like flyout or drilldown where you have nested menus/multiple items in a single `<li>`?
* Ensure all disabled items have the appropriate disabled classes/attrs. 
  * For example, the disabled item in the [with description](https://pf4.patternfly.org/components/menu#with-description) only has the disable class on the `li` and not on the `button`.

Issue #3674, opened 2020-11-13
Favorite star inconsistencies


mcoker:
Reviewing app launcher, select, table, and menu, I found these discrepancies in the hover states (the focus styles match the hover styles):

App launcher:
* Hovering either a favorited or non-favorited star turns the star black.

Select and Table:
* Hovering over a non-favorited star turns the star black
* Hovering over a favorited star, the star remains gold

Menu:
* No hover styles

----

Also all of the favorite buttons when disabled do not change colors except for the select - the select component updates the start with the disabled text color, which looks a little odd.

<img width="892" alt="Screen Shot 2020-11-13 at 3 52 03 PM" src="https://user-images.githubusercontent.com/35148959/99124703-47030800-25c8-11eb-835f-8d6de6c6e84b.png">

@mcarrano @mceledonia can you confirm what these should be? Should we take this opportunity to also introduce a favorite component that is a button, the icon, and the hover, focus, and disabled states built in?

mcarrano:
Yes, we were recently discussing this. @mceledonia can you add the desired behavior here? I agree that we should create a Favorites component to use consistently. I'll look at getting this on the roadmap.

mceledonia:
@mcoker @mcarrano 

For all favorite interaction across all components - 

**Hover**
- Hovering an unfavorited star (gray) should result in a black star
- Hovering a favorited star (gold) should result in a darker gold star `$pf-color-gold-500`

Issue #2365, opened 2019-10-11
Application launcher enhancements


mcarrano:
Introduce an enhanced Application Launcher that adds the following functionality:

* Search/filter capability to filter the list of applications. Should behave like the current type-ahead select to filter as list of choices while typing.

* Marking items as favorites and exposing a list of favorites at the top of the menu.

**Design doc:** https://docs.google.com/document/d/1jSQV48VBeavFG37akVGojfb6RpOq6r_14ebsTKntm3k/edit?usp=sharing

**Sketch file:** https://sketch.cloud/s/ezjkj

maryshak1996:
@mcarrano here is my design doc for this
https://docs.google.com/document/d/1jSQV48VBeavFG37akVGojfb6RpOq6r_14ebsTKntm3k/edit?usp=sharing

mcarrano:
Looks great @maryshak1996 .  @beanh66 can you take a look and see if there are any issues before we hand off for development?

maryshak1996:
These links are in the above doc as well, but here is my design file:
https://sketch.cloud/s/ezjkj

PR #3744, opened 2020-12-10
fix(menu): cleaned up examples, code

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3675

* Clean up hbs/examples
* Add disabled examples under actions and favorites
* Change item-main and item-description to spans
* Disable item action if parent item is disabled
* Clean up external icon var names
* Use proper global vars
* Update disabled item action, toggle icon colors

@mcarrano can you verify that if an item is disabled that any children actions should also be disabled? I added some new disabled examples to review under expanded toggle, links, actions, and favorites.

### DIFF
--- a/src/patternfly/components/Menu/examples/Menu.md
+++ b/src/patternfly/components/Menu/examples/Menu.md
@@ -32,8 +32,8 @@ import './Menu.css'
           {{/menu-item-main}}
         {{/menu-item}}
       {{/menu-list-item}}
-      {{#> menu-list-item menu-list-item--modifier="pf-m-disabled"}}
-        {{#> menu-item menu-item--attribute="disabled"}}
+      {{#> menu-list-item menu-list-item--IsDisabled="true"}}
+        {{#> menu-item}}
           {{#> menu-item-main}}
             {{#> menu-item-text}}
             Disabled action
@@ -41,8 +41,8 @@ import './Menu.css'
           {{/menu-item-main}}
         {{/menu-item}}
       {{/menu-list-item}}
-      {{#> menu-list-item menu-list-item--modifier="pf-m-disabled"}}
-        {{#> menu-item menu-item--IsLink="true" menu-item--modifier='aria-disabled="true" tabindex="-1"'}}
+      {{#> menu-list-item menu-list-item--IsDisabled="true"}}
+        {{#> menu-item menu-item--IsLink="true"}}
           {{#> menu-item-main}}
             {{#> menu-item-text}}
             Disabled link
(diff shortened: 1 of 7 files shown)


mcarrano (commented):
@mcoker In answer to your question, I think it makes sense that disabling an item would also disable any actions, but not clear what all the use cases are here. For example, would it make sense for the user to ever favorite an item that is disabled? I don't know. @gdoyle1 any thoughts here since this was done to address your use case in OS?

The only thing I did notice in reviewing this is that the favorite actions don't have hover states (on top of the row hover). Shouldn't they?

mcoker:
> the favorite actions don't have hover states

@mcarrano yep, I noted that here in this issue - https://github.com/patternfly/patternfly/issues/3674

mcoker:
@mcarrano @mceledonia updated the favorites colors - https://github.com/patternfly/patternfly/issues/3674

@gdoyle1 ping about this question. Matt left some comments on it, too here - https://github.com/patternfly/patternfly/pull/3744#pullrequestreview-549237389

> can you verify that if an item is disabled that any children actions should also be disabled?

mattnolting (approved):
🤩 Love it, LPTM!

gdoyle1:
@mcoker @mcarrano I think it makes sense that if the menu item is disabled, so are its children. Not sure if these colors will get updated in your other fixes but the disabled star looks more prominent than the actionable ones
Screen Shot 2020-12-15 at 9 02 33 AM

mceledonia:
@gdoyle1 Good catch, I actually think those star colors (default and disabled) are backwards. We would want the disabled color to be lightest gray and the default color to be our regular gray-to-black on hover action icon states.

@mcoker @mattnolting @mcarrano

gdoyle1:
> @gdoyle1 Good catch, I actually think those star colors (default and disabled) are backwards. We would want the disabled color to be lightest gray and the default color to be our regular gray-to-black on hover action icon states.
> 
> @mcoker @mattnolting @mcarrano

I think we'll also have to update those colors for other areas where we used them... like application launcher?

mcoker:
@mceledonia @gdoyle1 it would be great if the favorites button shared the same colors as plain buttons, though as @gdoyle1 said, we have favorites other places with the same styles - app launcher, table, and select.

FWIW here are the app launcher designs (the first favorites implementation)  - https://www.sketch.com/s/e82c74e1-b739-403b-bace-00c303a2bcc8/a/JQE2A5 and the app launcher issue - https://github.com/patternfly/patternfly/issues/2365

mceledonia:
I would update these across the board, I think they should follow the same pattern as any other actionable icons, with the exception of the "favorited" state of course.

PR #3746, opened 2020-12-10
fix(dropdown): support dropdown menu alignment on different breakpoint

### LINKED ISSUE #3069: Add `.pf-c-dropdown__menu.pf-m-align-right-on-[breakpoint]` classes
In OpenShift, we have a case where the position of a dropdown changes based on the viewport size.  As a result, we need to be able to switch the position of the dropdown menu in order to accommodate the changing dropdown placement.  I added the necessary `.pf-m-align-right-on-md` to https://github.com/openshift/console/pull/5308/files#diff-0a2376512d2b9815c0e2ae0da8992c8cR242, but certainly other PatternFly users are going to have a similar need.

DaoDaoNoCode:
closes #3069 .
Now it's not only support `pf-m-align-right-on-[breakpoint]` but also support `pf-m-align-left-on-[breakpoint]`, so you can set your style like `pf-m-align-right-on-md pf-m-align-left-on-xl`.

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -1,3 +1,6 @@
+
+$pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl", "2xl");
+
 .pf-c-dropdown {
   // Toggle
   --pf-c-dropdown__toggle--PaddingTop: var(--pf-global--spacer--form-element);
@@ -355,8 +358,19 @@
   background-clip: padding-box;
   box-shadow: var(--pf-c-dropdown__menu--BoxShadow);
 
-  &.pf-m-align-right {
-    right: 0;
+  @each $breakpoint, $breakpoint-value in $pf-c-dropdown--breakpoint-map {
+    $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
+
+    @include pf-apply-breakpoint($breakpoint, $pf-c-dropdown--breakpoint-map) {
+      &.pf-m-align-right#{$breakpoint-name} {
+        right: 0;
+      }
+
+      // unset alignment set for .pf-m-align-right
+      &.pf-m-align-left#{$breakpoint-name} {
+        right: unset;
+      }
+    }
   }
 
   .pf-c-dropdown.pf-m-top & {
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -355,8 +357,19 @@
...
+    $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
+
+    @include pf-apply-breakpoint($breakpoint, $pf-c-dropdown--breakpoint-map) {
+      &.pf-m-align-right#{$breakpoint-name} {
+        right: 0;
+      }
+
+      // unset alignment set for .pf-m-align-right
+      &.pf-m-align-left#{$breakpoint-name} {
+        right: unset;
```
mcoker, line 370: `unset` works here. Though for what it's worth, `unset` needs to check to see if the property is inheritable (`right` is not inheritable), then the value will be `inherit` if it is, or `initial` if it isn't.

So you could set this to `initial` and save the browser that check. I'm assuming it's an extra check anyways...

But the `initial` value for `right` is `auto`, so you could just set this to `auto`, too.

mcoker (approved):
Looks perfect. Left one comment about using `auto` instead of `unset`, but it's fine as it is, too. Thanks @DaoDaoNoCode!!!

src/patternfly/components/Dropdown/dropdown.scss
mattnolting, line 370: I think I'd just set it to `auto` to avoid any unintended outcome.

mattnolting (commented):
Looks great @DaoDaoNoCode! Can you add an example of the responsive alignment switching and add the update to the docs?

src/patternfly/components/Dropdown/dropdown.scss
DaoDaoNoCode: Will do, thank you!

DaoDaoNoCode:
@mcoker @mattnolting Could you review this again? I have updated the docs, add an example and change `right: unset` to `right: auto`.

mattnolting (approved):
Perfect! Thanks @DaoDaoNoCode 👍

mcoker (approved):
Great work @DaoDaoNoCode!! 🏆

Issue #3706, opened 2020-11-24
Update touchspin to use input[type="number"] and rename.


mcoker:
Update the touchspin component's `<input type="text">` to `type="number"`, and hide the default browser controls. Also want to rename this component to "Number input."

PR #3748, opened 2020-12-10
fix(touchspin): renamed touchspin to number input, use number type

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3706

@jessiehuff do you think we should we add `min`, `max`, or anything like that?


src/patternfly/components/NumberInput/number-input-unit.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-number-input__unit{{#if number-input-unit--modifier}} {{number-input-unit--modifier}}{{/if}}"
```
mattnolting, line 1:
```suggestion
<span class="pf-c-number-input__unit{{#if number-input-unit--modifier}} {{number-input-unit--modifier}}{{/if}}"
```

src/patternfly/components/NumberInput/number-input-unit.hbs
```diff
@@ -0,0 +1,6 @@
...
+  {{#if number-input-unit--attribute}}
+    {{{number-input-unit--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
```
mattnolting, line 6:
```suggestion
</span>
```

mattnolting (approved):
LGTM! Looks like the `-unit` element needs updated 👍

jessiehuff:
Since `<input type="number">` corresponds to the default role `spinbutton`, aria attributes supported by the spinbutton role should work. However, there are native html attributes that can be used with an `<input type="number">`, and I would default to using those where they exist. [These are the native html attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Additional_attributes) that are supported with `<input type="number">` which does include `min` and `max`. While I don't think these are required, we could add them as an enhancement or to show their usage for consumers.

mcoker:
@jessiehuff updated the lower/upper threshold examples with `min` and `max`, and added an a11y table with those attrs documented.

mattnolting (approved):
LPTM

jessiehuff (approved):
LGTM! 🙂

PR #3749, opened 2020-12-11
feat(switch-radio-check): updated input height and grid

### LINKED ISSUE #3210: Switch, Radio, Check - improve wrapping behavior
Screen Shot 2020-06-23 at 9 23 22 AM

Wrapping should be an edge case, since the labels should only be a few words, but let's update it so that the control is top-aligned with the content when content wraps. That would look something like:

Screen Shot 2020-06-23 at 9 24 56 AM

mattnolting:
closes #3210 

To test, add a long text string to `__label`. 

Screen Shot 2020-12-11 at 9 12 49 AM

Screen Shot 2020-12-11 at 9 13 34 AM

Screen Shot 2020-12-11 at 9 14 02 AM

### DIFF
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -6,13 +6,15 @@
   --pf-c-check__label--FontSize: var(--pf-global--FontSize--md);
   --pf-c-check__label--LineHeight: var(--pf-global--LineHeight--sm);
   --pf-c-check__input--MarginTop: #{pf-size-prem(-3px)};
+  --pf-c-check__input--Height: calc(var(--pf-c-check__label--FontSize) * var(--pf-c-check__label--LineHeight));
+  --pf-c-check__input--MarginTop: calc(-1px * (var(--pf-c-check__label--LineHeight) * 1.25));
   --pf-c-check__description--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-check__description--Color: var(--pf-global--Color--200);
 
   display: grid;
   grid-template-columns: auto 1fr;
   grid-gap: var(--pf-c-check--GridGap);
-  align-items: center;
+  align-items: start;
   justify-items: start;
 }
 
@@ -24,6 +26,7 @@
 }
 
 .pf-c-check__input {
+  height: var(--pf-c-check__input--Height);
   margin-top: var(--pf-c-check__input--MarginTop);
 }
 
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Check/check.scss
```diff
@@ -6,13 +6,15 @@
   --pf-c-check__label--FontSize: var(--pf-global--FontSize--md);
   --pf-c-check__label--LineHeight: var(--pf-global--LineHeight--sm);
   --pf-c-check__input--MarginTop: #{pf-size-prem(-3px)};
+  --pf-c-check__input--Height: calc(var(--pf-c-check__label--FontSize) * var(--pf-c-check__label--LineHeight));
+  --pf-c-check__input--MarginTop: calc(-1px * (var(--pf-c-check__label--LineHeight) * 1.25));
```
mcoker, line 10: is there any significance to 1.25 here or is that just the number that worked the best?

src/patternfly/components/Switch/switch.scss
```diff
@@ -1,6 +1,7 @@
 .pf-c-switch {
   // Switch - the vars for this file don't follow our normal order so that the IE11 script can convert the vars correctly
   --pf-c-switch--FontSize: var(--pf-global--FontSize--md);
+  --pf-c-switch--ColumnGap: var(--pf-global--spacer--md);
```
mcoker, line 4: Using this in place of the padding works fine, though if someone had customized `--pf-c-switch__label--PaddingLeft`, they're going to have double space here now.

WDYT about leaving `--pf-c-switch__label--PaddingLeft` and setting `--pf-c-switch--ColumnGap: var(--pf-c-switch__label--PaddingLeft)`?

Or maybe better, though we haven't done this before - defining this var as `--pf-c-switch--ColumnGap: var(--pf-c-switch__label--PaddingLeft, var(--pf-global--spacer--md));` and removing `--pf-c-switch__label--PaddingLeft`? 

Both of those will mean `--pf-c-switch__label--PaddingLeft` is a reserved var for the user, and not something we can use until we can clean it up in a breaking change, but it's less likely to break a user if they have customized this var.

mcoker (approved):
Bang up job on this! One question about the switch var but I could go either way with it.

src/patternfly/components/Check/check.scss
mattnolting: Excellent question. There appears to be a discrepancy in how FF, Safari and Chrome calculate `line-height` + `font-size`. Chrome & Safari calculate `height` (`1.3 * 16px`) as `20px`, or `_floor`, whereas FF calculates `height` as `20.8px`, the actual value. Bumping the negative margin by `.25` nudges the input up slightly to accomodate the FF difference.

src/patternfly/components/Switch/switch.scss
mattnolting: Love it, updated!

mcoker (approved):
L🍕TM!

PR #3750, opened 2020-12-11
fix(datalist-alignment): updated alignment settings

### LINKED ISSUE #3412: Data list alignment
This is just an example. We should review the component as a whole cross-browser.

Firefox
Screen Shot 2020-08-19 at 5 41 21 PM

Safari
Screen Shot 2020-08-19 at 5 38 03 PM

Chrome
Screen Shot 2020-08-19 at 5 39 30 PM

mattnolting:
closes #3412 

Preview: https://patternfly-pr-3750.surge.sh/components/data-list

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -335,7 +335,6 @@
 
 .pf-c-data-list__item-draggable-button {
   display: flex;
-  flex-direction: column;
   padding: var(--pf-c-data-list__item-draggable-button--PaddingTop) var(--pf-c-data-list__item-draggable-button--PaddingRight) var(--pf-c-data-list__item-draggable-button--PaddingBottom) var(--pf-c-data-list__item-draggable-button--PaddingLeft);
   margin: var(--pf-c-data-list__item-draggable-button--MarginTop) 0 var(--pf-c-data-list__item-draggable-button--MarginBottom) var(--pf-c-data-list__item-draggable-button--MarginLeft);
   border: 0;
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -89,7 +89,6 @@
   --pf-c-data-list__item-draggable-button--PaddingTop: var(--pf-global--spacer--lg);
   --pf-c-data-list__item-draggable-button--PaddingBottom: var(--pf-global--spacer--lg);
   --pf-c-data-list__item-draggable-button--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
-  --pf-c-data-list__item-draggable-button-icon--Height: calc(var(--pf-c-data-list--FontSize) * var(--pf-c-data-list--LineHeight));
   --pf-c-data-list__item-draggable-button-icon--Color: var(--pf-global--icon--Color--light);
   --pf-c-data-list__item-draggable-button--m-disabled__draggable-icon--Color: var(--pf-global--disabled-color--200);
   --pf-c-data-list__item-draggable-button--hover__draggable-icon--Color: var(--pf-global--icon--Color--dark);
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -1,7 +1,7 @@
 @import "./data-list-grid.scss";
 
 .pf-c-data-list {
-  --pf-c-data-list--FontSize: 1em; // inherit font-size
+  --pf-c-data-list--FontSize: 1em; // inherit font-size, update this to global variable at breaking change
   --pf-c-data-list--LineHeight: var(--pf-global--LineHeight--md);
   --pf-c-data-list--BorderTopColor: var(--pf-global--BorderColor--300);
   --pf-c-data-list--BorderTopWidth: var(--pf-global--spacer--sm);
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -336,7 +336,7 @@
   align-items: flex-start;
   align-self: stretch;
   padding: var(--pf-c-data-list__item-draggable-button--PaddingTop) var(--pf-c-data-list__item-draggable-button--PaddingRight) var(--pf-c-data-list__item-draggable-button--PaddingBottom) var(--pf-c-data-list__item-draggable-button--PaddingLeft);
-  margin-top: var(--pf-c-data-list__item-draggable-button--MarginTop) 0 var(--pf-c-data-list__item-draggable-button--MarginBottom) var(--pf-c-data-list__item-draggable-button--MarginLeft);
+  margin: var(--pf-c-data-list__item-draggable-button--MarginTop) 0 var(--pf-c-data-list__item-draggable-button--MarginBottom) var(--pf-c-data-list__item-draggable-button--MarginLeft);
   border: 0;
 
   &:hover {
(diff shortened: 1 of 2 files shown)


src/patternfly/components/DataList/data-list.scss
```diff
@@ -56,7 +58,7 @@
   --pf-c-data-list__item-content--md--PaddingBottom: var(--pf-global--spacer--lg);
 
   // Cell
-  --pf-c-data-list__cell--PaddingTop: var(--pf-global--spacer--lg);
+  --pf-c-data-list__cell--PaddingTop: calc(var(--pf-global--spacer--lg) + (var(--pf-c-data-list--LineHeight) * 1px));
```
mcoker, line 61: You shouldn't need the nested calc. Order of operations will multiply first.

```suggestion
  --pf-c-data-list__cell--PaddingTop: calc(var(--pf-global--spacer--lg) + var(--pf-c-data-list--LineHeight) * 1px);
```

src/patternfly/components/DataList/data-list.scss
```diff
@@ -123,48 +130,49 @@
...
-    font-size: var(--pf-c-data-list--m-compact--FontSize);
-
     .pf-c-data-list__cell {
       --pf-c-data-list__cell--PaddingTop: var(--pf-c-data-list--m-compact__cell--PaddingTop);
       --pf-c-data-list__cell--PaddingBottom: var(--pf-c-data-list--m-compact__cell--PaddingBottom);
       --pf-c-data-list__cell--MarginRight: var(--pf-c-data-list--m-compact__cell--cell--MarginRight);
       --pf-c-data-list__cell--cell--PaddingTop: var(--pf-c-data-list--m-compact__cell-cell--PaddingTop);
     }
 
+    --pf-c-data-list--FontSize: var(--pf-c-data-list--m-compact--FontSize);
```
mcoker, line 166: nit - can you move these vars before the `.pf-c-data-list__cell` block so the vars come first, then styles?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -123,48 +130,49 @@
...
   --pf-c-data-list--m-compact__cell-cell--PaddingTop: 0;
   --pf-c-data-list--m-compact__cell-cell--md--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-data-list--m-compact__cell--cell--MarginRight: var(--pf-global--spacer--md);
-  --pf-c-data-list--m-compact__item-control--PaddingTop: var(--pf-global--spacer--sm);
+  --pf-c-data-list--m-compact__item-control--PaddingTop: calc(var(--pf-global--spacer--sm) + (var(--pf-c-data-list--LineHeight) * 1px));
   --pf-c-data-list--m-compact__item-control--PaddingBottom: 0;
   --pf-c-data-list--m-compact__item-control--MarginRight: var(--pf-global--spacer--md);
-  --pf-c-data-list--m-compact__item-action--PaddingTop: var(--pf-global--spacer--sm);
-  --pf-c-data-list--m-compact__item-action--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-data-list--m-compact__item-action--PaddingTop: 0;
```
mcoker, line 144: This is causing this kind of layout

Screen Shot 2020-12-14 at 6 12 48 PM

src/patternfly/components/DataList/data-list.scss
mcoker: This would be 24px + 1.5px = 25.5px. What's the significance of adding the line height as a pixel value?

src/patternfly/components/DataList/data-list.scss
```diff
@@ -339,19 +369,20 @@
     margin-right: var(--pf-c-data-list__item-action--not-last-child--MarginRight);
   }
 
-  // Offset action button
+  // offset action button
   .pf-c-data-list__action {
-    margin-top: var(--pf-c-data-list__action--MarginTop);
```
mcoker, line 344: We need to leave that so it isn't breaking.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -132,19 +140,38 @@
...
 
+  font-size: var(--pf-c-data-list--FontSize);
+  line-height: var(--pf-c-data-list--LineHeight);
   overflow-wrap: break-word;
   list-style-type: disc;
   border-top: var(--pf-c-data-list--BorderTopWidth) solid var(--pf-c-data-list--BorderTopColor);
 
   &.pf-m-compact {
+    --pf-c-data-list__check--FontSize: var(--pf-c-data-list--m-compact__check--FontSize);
+    --pf-c-data-list__action--MarginTop: var(--pf-c-data-list--m-compact__action--MarginTop);
```
mcoker, line 162: An aside `--pf-c-data-list__action--MarginTop` was removed, but left here, although we need to add it back.

mattnolting: updated

mcoker (approved):
So much better! 🔥

src/patternfly/components/DataList/data-list.scss
```diff
@@ -71,41 +73,47 @@
...
   --pf-c-data-list__toggle-icon--Transition: .2s ease-in 0s;
   --pf-c-data-list__item--m-expanded__toggle-icon--Rotate: 90deg;
 
-  // Draggable button/icon
+  // draggable button/icon
   --pf-c-data-list__item-draggable-button--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-data-list__item-draggable-button--PaddingRight: var(--pf-global--spacer--md);
-  --pf-c-data-list__item-draggable-button--MarginTop: calc(var(--pf-global--spacer--sm) * -1);
+  --pf-c-data-list__item-draggable-button--MarginTop: 0;
   --pf-c-data-list__item-draggable-button--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
```
mcoker, line 86: After looking at this again, I believe we're going to need the draggable button to have the target area of a button. Currently it isn't necessary since you can drag from anywhere in a draggable row and it will drag the row, but I believe we're going to update that so that you have to click on/drag the button.

Can you add a padding-top/bottom using the form-element spacer, and use a negative top/bottom margin (of it's top/bottom padding), and the button still be aligned? This will also mean the element should not be `display: flex` if possible, so that the line height of the button can make the button as big as it needs to be. Though you could continue to make the draggable-button-icon have a fixed width of the font-size * line-height, I just wonder if that's necessary.

mattnolting: Screen Shot 2020-12-16 at 9 43 37 AM

I was able to achieve this while still using `flex`.

mcoker: @mcarrano @mceledonia wdyt about this? Or would you prefer the button only take up the space it needs, like this? 
Screen Shot 2020-12-16 at 9 21 48 AM

mcarrano: I like the bigger target, personally.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -322,6 +367,7 @@
   }
 
   .pf-c-data-list__item-draggable-icon {
+    height: var(--pf-c-data-list__item-draggable-button--Height);
```
mcoker, line 370: doesn't look like this is used

mcoker (commented):
Good to go except the declaration with `--pf-c-data-list__item-draggable-button--Height`

src/patternfly/components/DataList/data-list.scss
mattnolting: updated 👍

PR #3754, opened 2020-12-14
fix(navigation): add example with scroll-snap attribute

### LINKED ISSUE #3101: Investigate adding scroll-snap where appropriate
An initial spike to investigate lists like tabs and horizontal/tertiary nav (maybe page sections, too?), let's try out subtle scroll snapping to see what that experience is like and share with design to see if it's something we'd like to add to patternfly.

cc @mcarrano

DaoDaoNoCode:
closes #3101 .
@mcoker I currently only add an example to use `scroll-snap` in horizontal navigation. Could you review it first?
By the way, what other components do you think is appropriate to add this feature? I will implement them in this PR.

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -226,6 +226,12 @@
   --pf-c-nav--c-divider--PaddingLeft: 0;
   --pf-c-nav--c-divider--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
 
+  // Scroll snap
+  --pf-c-nav__list--ScrollSnapTypeAxis: x;
+  --pf-c-nav__list--ScrollSnapTypeStrictness: proximity;
+  --pf-c-nav__list--ScrollSnapType: var(--pf-c-nav__list--ScrollSnapTypeAxis) var(--pf-c-nav__list--ScrollSnapTypeStrictness);
+  --pf-c-nav__item--ScrollSnapAlign: end;
+
   @media screen and (min-width: $pf-global--breakpoint--xl) {
     --pf-c-nav__link--PaddingRight: var(--pf-c-nav__link--xl--PaddingRight);
     --pf-c-nav__link--PaddingLeft: var(--pf-c-nav__link--xl--PaddingLeft);
@@ -251,7 +257,7 @@
       max-width: 100%;
       overflow-x: auto;
       white-space: nowrap;
-      scroll-snap-type: x proximity;
+      scroll-snap-type: var(--pf-c-nav__list--ScrollSnapType);
       -webkit-overflow-scrolling: touch;
 
       @include pf-overflow-hide-scroll;
@@ -259,7 +265,7 @@
 
     .pf-c-nav__item {
       display: flex;
-      scroll-snap-align: end;
+      scroll-snap-align: var(--pf-c-nav__item--ScrollSnapAlign);
     }
 
     .pf-c-nav__link {
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -254,15 +254,12 @@
       scroll-snap-type: x proximity;
       -webkit-overflow-scrolling: touch;
 
-      > .pf-c-nav__item {
-        scroll-snap-align: end;
-      }
-
       @include pf-overflow-hide-scroll;
     }
 
     .pf-c-nav__item {
       display: flex;
+      scroll-snap-align: end;
     }
 
     .pf-c-nav__link {
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -251,11 +251,11 @@
       max-width: 100%;
       overflow-x: auto;
       white-space: nowrap;
-      scroll-snap-type: x mandatory;
+      scroll-snap-type: x proximity;
       -webkit-overflow-scrolling: touch;
 
       > .pf-c-nav__item {
-        scroll-snap-align: start;
+        scroll-snap-align: end;
       }
 
       @include pf-overflow-hide-scroll;
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -251,19 +251,14 @@
       max-width: 100%;
       overflow-x: auto;
       white-space: nowrap;
+      scroll-snap-type: x mandatory;
       -webkit-overflow-scrolling: touch;
 
-      @include pf-overflow-hide-scroll;
-
-      &.pf-m-scroll-snap {
-        scroll-snap-type: x mandatory;
-
-        // stylelint-disable
-        > .pf-c-nav__item {
-          scroll-snap-align: start;
-        }
-        // stylelint-enable
+      > .pf-c-nav__item {
+        scroll-snap-align: start;
       }
+
+      @include pf-overflow-hide-scroll;
     }
 
     .pf-c-nav__item {
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -254,6 +254,16 @@
       -webkit-overflow-scrolling: touch;
 
       @include pf-overflow-hide-scroll;
+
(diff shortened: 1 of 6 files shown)


mcoker (commented):
Very cool @DaoDaoNoCode! Some thoughts. @mceledonia @mcarrano wdyt?

1. I would probably use `proximity` instead of `mandatory` so the effect is more subtle. This effect is just a nice to have here, and the UI doesn't really rely on the snap behavior always happening, so it's probably better to enable it when the scroll position is near a snap point.
1. I'm not sure whether the left or right of a nav item should be the anchor point. @mceledonia @mcarrano I made a demo here - https://codesandbox.io/s/elegant-framework-wbp2d?file=/style.css. Can you let me know which you prefer?
    * "mandatory" is a more aggressive snap behavior (always snaps into place), and 
    * "proximity" is a little looser (only snaps into place if the end scroll point is near the parent's edge.
    * "start" has the snap point on the left side of the list
    * "end" has it on the right
1. I'm not sure if we want to make this a variation or just update the default behavior. I think last we spoke, we would just do it since it isn't very intrusive.
1. As for other components, we have a tertiary nav, tabs, notification drawer, page sections, table/data-list (rows would snap on the Y axis). For now I think we could just update this, tertiary nav, and tabs since those are all horizontal lists with scroll buttons that also scroll the list.

mattnolting:
I vote **Proximity End** `x proximity`. Feels the most natural as I scroll to reveal the list items obscured to the right, then scan back left to the previous list item.

mceledonia:
Yeah I'm with @mattnolting on this one, scroll to reveal items on the right so prioritizing them for alignment makes sense.

src/patternfly/components/Nav/nav.scss
```diff
@@ -251,8 +251,13 @@
       max-width: 100%;
       overflow-x: auto;
       white-space: nowrap;
+      scroll-snap-type: x proximity;
       -webkit-overflow-scrolling: touch;
 
+      > .pf-c-nav__item {
+        scroll-snap-align: end;
```
mcoker, line 258: You can save some lines of CSS and reduce the specificity of this selector by adding the `scroll-snap-align: end;` declaration to this block - https://github.com/patternfly/patternfly/blob/9dfbd1a9e1ffa053067e0835814f0653503cf939/src/patternfly/components/Nav/nav.scss#L264-L266

DaoDaoNoCode: Fixed, thanks!

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -145,6 +145,14 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
       margin-left: 0;
       transform: translateX(0);
     }
+
+    .pf-c-tabs__list {
```
mcoker, line 149: I don't think this only needs to apply to `.pf-m-scrollable`. While the react component will add this class to any tabs list that has overflow content, so it's fine to style it this way... for a core user who doesn't have the scroll buttons enabled, and just uses the browser scrolling to scroll a tabs list, they should probably still get the snapping behavior.

Also you could change the axis to "y" in `.pf-m-vertical` tabs.

I'm also wondering if we should create vars for these values so a user can customize/disable them.

Something like

```
--pf-c-tabs__list--ScrollSnapTypeAxis: x;
--pf-c-tabs__list--ScrollSnapTypeBehavior: proximity;
--pf-c-tabs__list--ScrollSnapType: var(--pf-c-tabs__list--ScrollSnapTypeAxis) var(--pf-c-tabs__list--ScrollSnapTypeBehavior);
```

wdyt @mattnolting ?

DaoDaoNoCode: Thank you, where do you think is the best place to put the styles? Do you think I should create some class like `.pf-m-horizontal` here or we just simply apply this to `.pf-c-tabs__list` and `.pf-c-tabs__item` and override them when it comes to `.pf-m-vertical`?

mcoker: The default styles (not having `.pf-m-vertical`) are the horizontal styles, so you can just add them to `__list` and `__item`, then modify them under `.pf-m-vertical`

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -23,6 +23,11 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs--m-color-scheme--light-300__link--BackgroundColor: transparent;
   --pf-c-tabs--m-color-scheme--light-300__item--m-current__link--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
 
+  // Tab list
+  --pf-c-tabs__list--ScrollSnapTypeAxis: x;
+  --pf-c-tabs__list--ScrollSnapTypeBehavior: proximity;
```
mcoker, line 28: Sorry I suggested the wrong thing. According to the [snap spec](https://drafts.csswg.org/css-scroll-snap-1/#snap-strictness) we should call this "strictness"

```suggestion
  --pf-c-tabs__list--ScrollSnapTypeStrictness: proximity;
```

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -23,6 +23,11 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
+  --pf-c-tabs__list--ScrollSnapType: var(--pf-c-tabs__list--ScrollSnapTypeAxis) var(--pf-c-tabs__list--ScrollSnapTypeBehavior);
```
mcoker, line 29:
```suggestion
  --pf-c-tabs__list--ScrollSnapType: var(--pf-c-tabs__list--ScrollSnapTypeAxis) var(--pf-c-tabs__list--ScrollSnapTypeStrictness);
```

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -330,13 +336,15 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
   scroll-behavior: smooth;
+  scroll-snap-type: var(--pf-c-tabs__list--ScrollSnapType);
   -webkit-overflow-scrolling: touch;
 }
 
 // Tabs item
 .pf-c-tabs__item {
   display: flex;
   flex: none;
+  scroll-snap-align: end;
```
mcoker, line 347: Can you create a var for this, too? `--pf-c-tabs__item--ScrollSnapAlign: end;`

src/patternfly/components/Nav/nav.scss
```diff
@@ -251,13 +251,15 @@
       max-width: 100%;
       overflow-x: auto;
       white-space: nowrap;
+      scroll-snap-type: x proximity;
```
mcoker, line 254: Can you create the same type of vars for the nav component?

DaoDaoNoCode:
@mcoker I've made all the changes you mentioned above. Could you please review it again?

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -229,6 +235,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
     --pf-c-tabs__link--after--Top: 0;
     --pf-c-tabs__link--after--Bottom: 0;
     --pf-c-tabs__link--after--Right: auto;
+    --pf-c-tabs__list--ScrollSnapTypeAxis: y;
```
mcoker, line 238: Can you make a var for this? 

```
// near the top
--pf-c-tabs--m-vertical__list--ScrollSnapTypeAxis: y;

// here in this block
--pf-c-tabs__list--ScrollSnapTypeAxis: var(--pf-c-tabs--m-vertical__list--ScrollSnapTypeAxis);

mcoker (approved):
Great work! 🥇

mattnolting (approved):
LPTM

mattnolting (approved):

PR #3756, opened 2020-12-14
fix(wizard): added cancel button element to increase cancel spacer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/1894

Once the action list component is promoted, we can use that here for more flexibility.

### DIFF
--- a/src/patternfly/components/Wizard/examples/Wizard.md
+++ b/src/patternfly/components/Wizard/examples/Wizard.md
@@ -87,9 +87,11 @@ import './Wizard.css'
       {{#> button button--modifier="pf-m-secondary"}}
         Back
       {{/button}}
-      {{#> button button--modifier="pf-m-link"}}
-        Cancel
-      {{/button}}
+      {{#> wizard-footer-cancel}}
+        {{#> button button--modifier="pf-m-link"}}
+          Cancel
+        {{/button}}
+      {{/wizard-footer-cancel}}
     {{/wizard-footer}}
   {{/wizard-outer-wrap}}
 {{/wizard}}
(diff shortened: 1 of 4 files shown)


matthewcarleton:
Thanks for doing this @mcoker I'm leaning towards 16px instead of 32px (I think that's what @mceledonia suggested too). 
Screen Shot 2020-12-15 at 9 23 02 AM

mceledonia:
I'm thinking 24px or 16px should work well! I do like it being spaced a little further out than the primary and secondary action.

mattnolting:
> Thanks for doing this @mcoker I'm leaning towards 16px instead of 32px (I think that's what @mceledonia suggested too). Seeing 48px in the browser now feels a bit extreme ( I know I suggested it so my bad :) )

@mceledonia @matthewcarleton @mcoker WDYT about zeroing the left/right padding on the `cancel` button to avoid inadvertent clicks as @matthewcarleton [noted here](https://github.com/patternfly/patternfly/issues/1894#issue-452591166), with a 16px margin-left? 

Looks like this:

Screen Shot 2020-12-15 at 9 56 11 AM

Screen Shot 2020-12-15 at 9 56 18 AM

mcoker:
@matthewcarleton @mceledonia when we created the action list, the docs said the space should be 48px in the wizard, too - https://docs.google.com/presentation/d/1Kh8wGM27X5shhSyVpZbgjV3uidpOmlz-CLYWHhxEMEI/edit#slide=id.g6c2040c51b_2_17

Since the action list has a 48px for groups, I was expecting that to be used here once the component is promoted from beta.

Just want to confirm we want to deviate from the 48px spacing, and assuming so, whether we want to push this change back to the action list. I assume we'll also want to have cancel spacing other places, like the form component, so it would be good if the action list supports a unique cancel spacing so we can use that component instead, and also extend the ability to use groups in wizards, forms, etc.

mcarrano:
I agree that we should be consistent with what we defined for the Action list so we can later use that component. If we really want to revisit this, let's make the update there also, but personally I'm OK with the 48px spacing.

matthewcarleton:
ah ok - I won't rock the boat then. Happy to go with this @mcoker

mceledonia:
Agreed, given the context of the decision I think we're alright to leave it as-is.

matthewcarleton (approved):
LGTM!

mcarrano (approved):
Looks good!

mattnolting (approved):
I'd like to revisit spacing later, bc it looks too big to me, but the update is 💯

PR #3761, opened 2020-12-16
refactor(mixins): remove breakpoint map in pf-apply-breakpoint function

### LINKED ISSUE #3751: Apply breakpoint function update
We have a SCSS function `pf-apply-breakpoint`, which requires us to pass 2 parameters here `$breakpoint` and `$breakpoint-map`. However, we don't use `$breakpoint-map` in this function, so it's better to get rid of it.

Screen Shot 2020-12-11 at 3 02 47 PM

DaoDaoNoCode:
closes #3751 .
`$breakpoint-map` is not needed and I have removed it from the function and anywhere we call the function.


mattnolting (approved):
Nice work, can't believe I missed that! 🤯

mcoker (approved):
Nice find @DaoDaoNoCode!!

PR #3778, opened 2021-01-06
fix(drawer): fix splitter scroll off when scroll the panel

### LINKED ISSUE #3776: Resizable drawer: Splitter scrolls off screen when drawer body scrolls
@evwilkin commented on [Fri Dec 18 2020](https://github.com/patternfly/patternfly-react/issues/5284)

DaoDaoNoCode:
closes #3776 .
See the example here [https://www.patternfly.org/v4/components/drawer/html#resizable-bottom-panel](url). If the content in the panel is long enough to scroll, it will scroll the whole panel, thus the splitter will also be scrolled up and disappears.
So I add `overflow: auto` to `.pf-c-drawer__body` so it will only scroll the body in the panel content and will not scroll off the splitter.


mcoker:
This works great if you only have a single `__body` element. However, you can have multiple `__body` elements, meaning you could end up with multiple scrollbars. Here's an example with multiple `__body` elements - https://patternfly-react.surge.sh/components/drawer#stacked-content-body-elements

And here's what that looks like with each `__body` set to `overflow: auto;`.

You could try and see if `position: sticky` will work. And since this is an enhancement and the component is still beta, we could introduce a parent element to contain the `__body` elements, and set that element to `overflow: auto;`. That's less ideal as it requires a markup change, but it's an option.

DaoDaoNoCode:
@mcoker Updated. I add `.pf-c-drawer__panel-main` class and set the layout to flex. Could you please review it when you have time? Thank you.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -231,6 +233,13 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   overflow: auto;
 }
 
+.pf-c-drawer__panel-main {
+  @extend .pf-c-drawer__panel;
```
mcoker, line 237: This may work in this context, but we usually do not use `@extend` as it can cause unpredictable issues, particularly in a large project. It will extend any instance of `.pf-c-drawer__panel` anywhere in the stylesheet, including compound selectors. The first part of [this article](https://webinista.com/updates/dont-use-extend-sass/) covers it pretty well.

A better alternative could be to use a placeholder or `@mixin` for shared blocks like this, however in this instance I would suggest just adding `.pf-c-drawer__panel-main` to any `.pf-c-drawer__panel` styles that need to be shared.

mcoker: Here's a good example of styles from `.pf-c-drawer__panel` that apply to `.pf-c-drawer__panel-main` now, and you don't want any of them.

Screen Shot 2021-01-11 at 10 54 48 AM

The only rule in there that you want on `__panel-main` is `overflow: auto;` but that will come from this block when you add `__panel-main` to it:

https://github.com/patternfly/patternfly/blob/506ec62cc4b7c824f42ad6f14cd37a0a25538c68/src/patternfly/components/Drawer/drawer.scss#L225-L232

DaoDaoNoCode: @mcoker I see! I had never expected using `@extend` would cause such big problems. This is a great reference and thank you for pointing that out. I will fix it!

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -417,7 +412,20 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
...
+      > .pf-c-drawer__splitter {
+        right: var(--pf-c-drawer__panel--m-resizable--PaddingLeft);
+        flex-grow: 0;
+        flex-shrink: 0;
+      }
+
+      > .pf-c-drawer__panel-main {
+        flex-shrink: 1;
+        flex-grow: 1;
+        margin-left: calc(var(--pf-c-drawer__panel--m-resizable--PaddingLeft) * -1);
```
mcoker, line 427: A couple of things about this - first, just to note about syntax and structure, we typically don't do `calc()`s with a variable that do not belong to the property in the declaration in the CSS. We try as much as we can to do the calcs and refrences to other variables as the value of another variable that matches the property name in a declaration. As an example, for this we would do....

```
--pf-c-drawer__panel--m-resizable__panel-main--MarginLeft: calc(var(--pf-c-drawer__panel--m-resizable--PaddingLeft) * -1);

[...]

margin-left: var(--pf-c-drawer__panel--m-resizable__panel-main--MarginLeft:);
```

However, I don't think the `margin-left` here is necessary, or the `margin-right` on the left panel or `margin-top` on the bottom panel. It looks like you can remove the `padding-left`, `padding-right`, and `padding-top` from the default panel, left panel, and bottom panel now. That padding was used previously to create space for the absolutely positioned splitter element.

And since you can remove that, you shouldn't need negative margins on those elements, or the `right` and `bottom` you've added to the `.pf-c-drawer__splitter` element.

DaoDaoNoCode: @mcoker Thank you for you suggestions on `calc()` function, I will update it.
However, the reason I set `padding` and `margin` here is for the usage in React. When the `flex-basis` of the resizable panel is set to `0`, we need to use the `padding` to keep the splitter inside the drawer. Thus, we also need `margin` to keep the panel body in the right position. You may try to remove `padding` and `margin` mentioned above, but the splitter will be disappeared when you change the size of the panel to 0, then you will not be able to drag it back.
Or do you think there is a better way to keep the splitter inside the drawer?

src/patternfly/components/Drawer/drawer-example-resizable-panel.hbs
```diff
@@ -0,0 +1,12 @@
+{{#> drawer-panel drawer-panel--NoBody="true"}}
```
mcoker, line 1: This was bugging me that we have `drawer-panel--NoBody` set and are including `{{#> drawer-body}}` below. I changed the logic of `drawer-panel.hbs` a little to see if we can make that file a little more logical and added a PR for your branch here - https://github.com/DaoDaoNoCode/patternfly/pull/1. Let me know what you think, if you think that improves it or you'd like to handle it a different way.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -479,13 +479,9 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
     }
 
     > .pf-c-drawer__main > .pf-c-drawer__panel.pf-m-resizable {
-      padding-top: var(--pf-c-drawer--m-panel-bottom__panel--m-resizable--PaddingTop);
-      padding-left: 0;
+      flex-direction: var(--pf-c-drawer--m-panel-bottom__panel--m-resizable--FlexDirection);
```
mcoker, line 482: We try to reassign a var if we can instead of overriding the property. However, CSS vars cascade. So you can introduce them at the highest level selector where you need the change. It looks like you might be able to make this variable re-assignment a few lines up, underneath `.pf-c-drawer.pf-m-panel-bottom`

```suggestion
      --pf-c-drawer__panel--m-resizable--FlexDirection: var(--pf-c-drawer--m-panel-bottom__panel--m-resizable--FlexDirection);
```

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -443,12 +445,9 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
     }
 
     > .pf-c-drawer__main > .pf-c-drawer__panel.pf-m-resizable {
-      padding-right: var(--pf-c-drawer--m-panel-left__panel--m-resizable--PaddingRight);
-      padding-left: 0;
-
       > .pf-c-drawer__splitter {
-        --pf-c-drawer__splitter--Right: 0;
-        --pf-c-drawer__splitter--Left: auto;
+        order: 2;
```
mcoker, line 449: We try to use the first value needed to make the change, so `order: 1;` should work since nothing else in this flex grouping has an `order` 

```suggestion
        order: 1;
```

mcoker (commented):
Looking good @DaoDaoNoCode! Left a few bits of feedback, let me know what you think.

src/patternfly/components/Drawer/drawer-example-resizable-panel.hbs
DaoDaoNoCode: @mcoker Thank you for your suggestions! Actually I had no idea about how to organize the logic at first but it's much more clear to refer to your commit. I will update this and other examples that using `drawer-panel--NoBody` in the doc.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -460,6 +459,7 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   .pf-c-drawer.pf-m-panel-bottom {
     --pf-c-drawer--m-expanded__panel--BoxShadow: var(--pf-c-drawer--m-expanded--m-panel-bottom__panel--BoxShadow);
     --pf-c-drawer__panel--MaxHeight: 100%;
+    --pf-c-drawer__panel--FlexBasis--min: var(--pf-c-drawer__splitter--Height);
```
mcoker, line 462: Can you create a variable for the value used here, except move it up to the top of the component where the variables are set?

```
--pf-c-drawer--m-panel-bottom__panel--FlexBasis--min: var(--pf-c-drawer__splitter--Height);
```

And then here on this line, change it to:

```
--pf-c-drawer__panel--FlexBasis--min: var(--pf-c-drawer--m-panel-bottom__panel--FlexBasis--min);
```

That will create a variable that users can modify (`--pf-c-drawer--m-panel-bottom__panel--FlexBasis--min`) in their application by setting the code below.

```
.pf-c-drawer {
  --pf-c-drawer--m-panel-bottom__panel--FlexBasis--min: // some custom value;
}
```

Otherwise they would have to set it via:

```
.pf-c-drawer.pf-m-panel-bottom { 
  --pf-c-drawer__panel--FlexBasis--min: // some custom value;
}
```

mcoker:
@DaoDaoNoCode looks like the primary detail demos aren't working as expected - https://patternfly-pr-3778.surge.sh/demos/primary-detail/html-demos/

DaoDaoNoCode:
@mcoker Hmm... It's weird. I saw the bug at first, but the bug disappears now without updating anything. Is it because at that time the building was not finished?

mcoker (approved):
Very nice! Great work @DaoDaoNoCode 🥇

mattnolting (commented):
This is looking good @DaoDaoNoCode. The drop shadow disappears at narrow width and I'm not sure why. Any ideas?

Screen Shot 2021-01-13 at 10 00 26 AM


I found that bumping [`--pf-c-drawer__splitter--m-vertical--Width: #{pf-size-prem(9px)};` ](https://github.com/patternfly/patternfly/pull/3778/files#diff-835c6b836c04b3ceeadcff9a9cd802d43883b6a57f36325f80fa07fb6c1b30d8L73)to `#{pf-size-prem(24px)}` retains the dropshadow.

DaoDaoNoCode:
@mattnolting Actually I have no idea about why the shadow disappears when the panel width is small. This issue happens before I make these changes.
@mcoker Do you have any thoughts? Is it designed to do so or is this a bug?

mcoker:
> @mcoker Do you have any thoughts? Is it designed to do so or is this a bug?

Looks like it's because of our use of negative spread in `$pf-global--BoxShadow--lg-left`

https://github.com/patternfly/patternfly/blob/1a69538ff9a8a473d9261c050992640a87779efe/src/patternfly/sass-utilities/scss-variables.scss#L112

The element will need to be twice that size for the shadow to show up. I think we can create a follow up issue for that. Maybe we can move the `box-shadow` to another element that has a minimum width. We could also consider using a custom shadow. I would also love to not transition that property because it's an expensive property to animate. Do you mind creating a follow up issue for that?

mattnolting (approved):
LPTM! Fantastic work @DaoDaoNoCode 🎉

DaoDaoNoCode:
Create a follow up issue here #3795 for panel box shadow problem.

Issue #3771, opened 2020-12-18
Search input: attribute-value search field


mcarrano:
Enhance the Search input component to support advanced attribute-value search use cases that allow the user to enter a query string directly in the input field or fill out a search form. Designs: https://marvelapp.com/prototype/276i4086

PR #3783, opened 2021-01-08
feat(search-input): added advanced search

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3771

### DIFF
--- a/src/patternfly/components/SearchInput/search-input.scss
+++ b/src/patternfly/components/SearchInput/search-input.scss
@@ -30,12 +30,28 @@
   --pf-c-search-input__utilities--c-button--PaddingRight: var(--pf-global--spacer--xs);
   --pf-c-search-input__utilities--c-button--PaddingLeft: var(--pf-global--spacer--xs);
 
+  // Menu
+  --pf-c-search-input__menu--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-search-input__menu--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-search-input__menu--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-search-input__menu--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-search-input__menu--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
+  --pf-c-search-input__menu--BoxShadow: var(--pf-global--BoxShadow--md);
+  --pf-c-search-input__menu--Top: calc(100% + var(--pf-global--spacer--xs));
+  --pf-c-search-input__menu--ZIndex: var(--pf-global--ZIndex--sm);
+  --pf-c-search-input--m-top__menu--Top: 0;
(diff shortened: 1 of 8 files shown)


mcarrano (commented):
Looks great @mcoker . I'm just wondering why the close/clear button disappeared in the expanded state. Was that an oversight?

mcoker:
> the close/clear button disappeared in the expanded state

@mcarrano I was going from the designs where, in the collapsed state the text input has a real value of the search params, but in the expanded state, the text input just has a `placeholder` value.

Screen Shot 2021-01-08 at 10 30 13 AM

However, after reading the design notes, it sounds like the text input and form will always be in sync? Meaning I should update the expanded example so that the text input value matches what's in the form?

Screen Shot 2021-01-08 at 10 36 16 AM

mcarrano:
Ah, I had forgotten that @mcoker . I think that what confused me was that even thought the text style of the field changes to the placeholder style and the clear button is hidden, I can still type into the text field when the form is open. Do you think that typing into the field should be disabled when the form is open?

mcoker:
> Do you think that typing into the field should be disabled when the form is open?

I'm not sure, would the advanced search always be searching attribute/value pairs, or could it be used (for example) to contain something like checkboxes/radios to filter search categories? I guess I could see each product handling it a different way depending on what the text input does versus what the advanced form does, and whether it makes sense to disable the text input, keep the text input and advanced search form in sync, leave the text input editable and the form active, etc.

mcarrano:
That's a fair point. It's just a bit off the way it is now where I can edit text in the field but not clear the field when the form is open. Maybe the clear button should just remain at all times?

mcoker:
@mcarrano updated the examples so that both have a value in the text intput, and the value is different from the placeholder text (the value and placholder were the same in the previous examples/design).

https://patternfly-pr-3783.surge.sh/components/search-input#advanced-search

mcarrano (approved):
Looks good to me @mcoker

src/patternfly/components/SearchInput/search-input.scss
```diff
@@ -113,3 +129,20 @@
   display: flex;
   align-items: center;
 }
+
+.pf-c-search-input__menu {
```
mattnolting, line 133: Should we use `pf-c-menu` here?

mattnolting (commented):
This looks great! Only question is, since we've created the `pf-c-menu` component, seems like this example is a candidate for its use.

src/patternfly/components/SearchInput/search-input.scss
mcoker: Definitely could be, though since the menu is beta, and may be for a while, using it here would make the search input beta until the menu is promoted. Since this is really just a generic box and doesn't need the stuff the menu provides (items, groups, etc), I just made this something in the search input.

mattnolting: wfm

mattnolting (approved):
Fantastic! LPTM 🚀

Issue #3784, opened 2021-01-08
Drawer - inline bottom panel layout issues


mcoker:
From https://github.com/patternfly/patternfly-react/issues/5289

The inline/bottom drawer panel doesn't behave properly, it doesn't transition from the bottom and the panel takes up layout when collapsed.

cc @kmcfaul

PR #3785, opened 2021-01-08
fix(drawer): fixed inline bottom panel layout issues

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3784


DaoDaoNoCode (approved):
Looks great to me! Thanks @mcoker . This would also probably fix https://github.com/patternfly/patternfly-react/issues/5288 in the React repo. I will mention this PR under that issue, too.

mattnolting (commented):
Looks great. I wonder if the transition, before `md` breakpoint should be from right to left, like the other drawers variants or if from bottom is the desired effect @mcarrano? Otherwise, LPTM

Issue #3787, opened 2021-01-09
Resizable drawer splitter disappears at mobile breakpoint


mcoker:
Currently if you resize a drawer and the react component sets a manual drawer width, the drawer remains resized on at the mobile breakpoint, but the splitter disappears. After talking to @mcarrano the drawer should not be resizable at the mobile breakpoint.

PR #3788, opened 2021-01-09
fix(drawer): limited drawer resize to desktop breakpoint

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3787

This adds a var `--pf-c-drawer__panel--m-resizable--FlexBasis` that can be used anywhere we want it to be able to override the panel `flex-basis`. It also moves the inline var from the panel to the parent drawer.

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -26,6 +26,7 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --pf-c-drawer__panel--TransitionDuration: var(--pf-global--TransitionDuration);
   --pf-c-drawer__panel--TransitionProperty: margin, transform, box-shadow, flex-basis;
   --pf-c-drawer__panel--m-resizable--PaddingLeft: var(--pf-c-drawer__splitter--m-vertical--Width);
+  --pf-c-drawer__panel--m-resizable--FlexBasis: initial;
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Drawer/drawer.scss
```diff
@@ -130,12 +131,12 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   }
 
   @media screen and (min-width: $pf-global--breakpoint--xl) {
-    --pf-c-drawer__panel--FlexBasis: var(--pf-c-drawer__panel--xl--FlexBasis);
+    --pf-c-drawer__panel--FlexBasis: var(--pf-c-drawer__panel--m-resizable--FlexBasis, var(--pf-c-drawer__panel--xl--FlexBasis));
```
mattnolting, line 134: Should this be:

```suggestion
    --pf-c-drawer__panel--FlexBasis: var(--pf-c-drawer__panel--m-resizable--FlexBasis, var(--pf-c-drawer__panel--xl--FlexBasis, var(--pf-c-drawer__panel--md--FlexBasis)));
```

mcoker: It could be, though I don't think it needs to be. The fallback to `--md` would be needed if `--xl` were an optional var, but since `--pf-c-drawer__panel--xl--FlexBasis` is set by the component, it should always be true. wdyt?

mattnolting (approved):
LPTM!

DaoDaoNoCode (approved):
Looks great to me. Thanks @mcoker !

Issue #3623, opened 2020-10-26
Add placeholder text style for Form Select


mcarrano:
Want to provide placeholder styled text for the Form Select similar to the placeholder styling for other form elements (e.g. Text inputs). 

Design recommendation:

![image.png](https://images.zenhubusercontent.com/5ce816524e6c594dc1b0df4d/1f5adede-609f-4d4f-9d3c-c4489849b62e)

PR #3790, opened 2021-01-11
feat(form-control): added placeholder variation for form select

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3623

After chatting with @kmcfaul the most flexible way to do this is to dynamically toggle a class based off of the chosen option.


mattnolting (approved):
LPTM!

mcarrano (approved):
Looks good to me. Thanks @mcoker !

PR #3793, opened 2021-01-12
fix(card): introduce large card variant

### LINKED ISSUE #3670: Introduce Large (marketing) card variant
Introduce a large card variant per the design here:

image

This will satisfy marketing use cases. Along with this want to introduce a modifier to add rounded corners to any card.

@gdoyle1 can you provide spec for intended corner radius?

DaoDaoNoCode:
closes #3670

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -24,7 +24,7 @@
   --pf-c-card--m-display-lg--child--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-card--m-display-lg--child--PaddingLeft: var(--pf-global--spacer--xl);
   --pf-c-card--m-display-lg--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--lg);
   --pf-c-card--m-flat--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-card--m-flat--BorderColor: var(--pf-global--BorderColor--100);
   --pf-c-card--m-rounded--BorderRadius: var(--pf-global--BorderRadius--sm);
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -17,14 +17,14 @@
   --pf-c-card--m-compact--child--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-card--m-compact--c-divider--child--PaddingTop: var(--pf-global--spacer--md);
   --pf-c-card--m-compact__title--not--last-child--PaddingBottom: var(--pf-global--spacer--sm);
-  --pf-c-card--m-large__body--FontSize: var(--pf-global--FontSize--lg);
-  --pf-c-card--m-large__footer--FontSize: var(--pf-global--FontSize--lg);
-  --pf-c-card--m-large--first-child--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-card--m-large--child--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-card--m-large--child--PaddingBottom: var(--pf-global--spacer--xl);
-  --pf-c-card--m-large--child--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-card--m-large--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-card--m-large__title--not--last-child--PaddingBottom: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg__body--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-card--m-display-lg__footer--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-card--m-display-lg--first-child--PaddingTop: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--child--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--child--PaddingBottom: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--child--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--xl);
   --pf-c-card--m-flat--BorderWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-card--m-flat--BorderColor: var(--pf-global--BorderColor--100);
   --pf-c-card--m-rounded--BorderRadius: var(--pf-global--BorderRadius--sm);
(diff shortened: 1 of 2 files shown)


mcoker:
@mcarrano @mceledonia should this follow the variation naming we introduced for the tile and CTA button (`.pf-m-display-lg`)?

src/patternfly/components/Card/examples/Card.md
```diff
@@ -347,9 +377,11 @@ A card is a generic rectangular container that can be used to build other compon
...
+| `.pf-m-large` | `.pf-c-card` | Creates a large variation of the card component that involves larger font sizes and spacing. This variation is for marketing use cases. |
 | `.pf-m-no-fill` | `.pf-c-card__body` | Sets a `.pf-c-card__body` to not fill the available space in `.pf-c-card`. `.pf-m-no-fill` can be added to multiple card bodies. |
 | `.pf-m-hoverable` | `.pf-c-card` | Modifies the card to include hover styles on `:hover`. |
 | `.pf-m-selectable` | `.pf-c-card` | Modifies a selectable card so that it is selectable. |
 | `.pf-m-selected` | `.pf-c-card.pf-m-selectable` | Modifies a selectable card for the selected state. |
-| `.pf-m-flat` | `.pf-c-card` | Modifies the card to have a border instead of a shadow. `.pf-m-flat` is for use in layouts where cards are against a white background.
+| `.pf-m-flat` | `.pf-c-card` | Modifies the card to have a border instead of a shadow. `.pf-m-flat` is for use in layouts where cards are against a white background. |
+| `.pf-m-rounded` | `.pf-c-card` | Modifies the card to for the rounded corner. |
```
mcoker, line 386:
```suggestion
| `.pf-m-rounded` | `.pf-c-card` | Modifies the card to have rounded corners. |
```

src/patternfly/components/Card/card.scss
```diff
@@ -17,8 +17,17 @@
...
   --pf-c-card--m-compact--c-divider--child--PaddingTop: var(--pf-global--spacer--md);
   --pf-c-card--m-compact__title--not--last-child--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-card--m-large__body--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-card--m-large__footer--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-card--m-large--first-child--PaddingTop: var(--pf-global--spacer--xl);
+  --pf-c-card--m-large--child--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-card--m-large--child--PaddingBottom: var(--pf-global--spacer--xl);
+  --pf-c-card--m-large--child--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-card--m-large--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
+  --pf-c-card--m-large__title--not--last-child--PaddingBottom: var(--pf-global--spacer--lg);
```
mcoker, line 27: Looking at these designs, it looks like the header/title should have the `--xl` spacer.

Screen Shot 2021-01-12 at 2 12 05 PM

In the card, the title and header are somewhat synonymous as far as the spacing is concerned. The header is just used to wrap the title if you have other things you want to put adjacent to the title. In the examples in the design, there is a "title" but that looks to just be a generic title, not necessarily the card's title element. Do you see it differently?

DaoDaoNoCode: @mcoker Sorry for not noticing that! Thank you, will update.

mcarrano:
@mcoker I'm OK to call this "Large" for lack of a better name unless @mceledonia or @ajacobs21e have a better idea.

ajacobs21e:
Large sounds good to me. The only other thing I could think of is something like 'web' or 'marketing' but they're too specific.

mcoker:
@mcarrano @ajacobs21e thanks! Just to clarify, the variation is currently called `pf-m-large`. This is inline with other components, like empty state or modal, with basic size variations.

However, when we introduced the CTA button, we decided to introduce a `pf-m-display-[name]` variation (https://github.com/patternfly/patternfly/pull/3214#issuecomment-655722194) - https://pf4.patternfly.org/components/button#call-to-action, and to use that for variants of a different presentational use, like web/marketing. In theory, we could have a `pf-m-large` variant of a card that is just a general "large" card, and still have a `pf-m-display-lg` as a separate variant that is for web/marketing use. 

We also used `pf-m-display-lg` as a variation for the tile (https://github.com/patternfly/patternfly/pull/3229) - https://pf4.patternfly.org/components/tile#stacked-tiles-large

Following that, I'm assuming it makes sense to call this variation `pf-m-display-lg` since it's specifically for web/marketing, and not just another size variant for the card? Since naming things is always hard, I just want to make sure we're all on the same page and in agreement.

mceledonia:
@mcoker That makes sense to me in terms of being consistent. I do think it would make sense to just have more generic large variations without implying a use case with the name, that seems like something better suited for documentation. I would say given that the `display` naming convention already exists, we should probably be consistent with it for now.

mcoker (approved):
LGTM!

mattnolting (approved):
Perfect! Well done @DaoDaoNoCode! 🚢  it!

mcarrano (commented):
Looks good to me, but @mceledonia can you also take a look?

As an aside, I don't understand why the card variants/examples don't show in the right-hand POC for this page as they do for other components.

mcoker:
> As an aside, I don't understand why the card variants/examples don't show in the right-hand POC for this page as they do for other components.

@redallen do you know why that is? Seems like if you have more than 23 h3's, all of them disappear.

redallen:
@mcoker Bump theme-patternfly-org and this should be fixed. I think the limit used to be 25 items total per design.

mceledonia:
@mcarrano @gdoyle1 I am definitely late to the party here, but reviewing this I got confused by the usage of the title section. In the design examples above, there are various examples where the card header is in the body section, when I would expect it to be in the title section. I was always under the impression that the title/header section of the card was primarily for this purpose. I see the use case for an icon/image there, shown in another example, but I can see this getting confusing for users when trying to understand what to use the title section for (why not the actual title/header?). 

image

In other words, I would expect the title section to be used for titles/headers/images, and the font size of the text inside that area to reflect that (larger than body font as seen in designs)

I first noticed this when I noticed the mismatch of font sizes in the implementation of header/body. The header section has a 16px font size and the body has 18px, which doesn't match the designs at first glance.

Any insight on the decision to use headers in the body section as seen above?

gdoyle1:
@mceledonia I think the problem was that the spacing between the header and body was changing throughout the examples that the web team had. There was a new-ish part of the card, which was optional, which is the image/video. And then sometimes a title was placed in but didn't span across the whole card, and the spacing between the title and body often changed.

I'm good with going with whatever you suggest to standardize them all. I think it would just require us to make sure the spacer used btwn title and body is consistent

src/patternfly/components/Card/card.scss
```diff
@@ -17,8 +17,17 @@
...
   --pf-c-card--m-compact--c-divider--child--PaddingTop: var(--pf-global--spacer--md);
   --pf-c-card--m-compact__title--not--last-child--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-card--m-display-lg__body--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-card--m-display-lg__footer--FontSize: var(--pf-global--FontSize--lg);
+  --pf-c-card--m-display-lg--first-child--PaddingTop: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--child--PaddingRight: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--child--PaddingBottom: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--child--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
+  --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--xl);
```
mcoker, line 27:
```suggestion
  --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--lg);
```

mcoker: @DaoDaoNoCode confirmed with @mceledonia that everything should have an `xl` spacer except the space below the title, that should be `lg`

mcoker (approved):

mattnolting (approved):
LPTM🎉

PR #3794, opened 2021-01-12
fix(card-demo): add empty state card into the demo

### LINKED ISSUE #3687: Add empty state card to Card View demo
Add a New card/empty state card to the Card View demo here: https://www.patternfly.org/v4/components/card/html-demos

Designs: https://marvelapp.com/prototype/f869h61/screen/74368154

DaoDaoNoCode:
closes #3687 .
Thanks @mmenestr ! We have a fourth variation for the empty state card here https://marvelapp.com/prototype/f869h61/screen/74368154. It contains the title and removes the empty state body.

### DIFF
--- a/src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
+++ b/src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
@@ -3,9 +3,9 @@
       {{#> bullseye}}
         {{#> empty-state empty-state--modifier="pf-m-xs"}}
           <i class="fas fa-plus-circle pf-c-empty-state__icon"></i>
-          {{#> card-title}}
-            <p id="{{card--id}}-check-label">To get started, add a new group to your page.</p>
-          {{/card-title}}
+          {{#> title titleType="h2" title--modifier="pf-m-md"}}
+            To get started, add a new group to your page.
+          {{/title}}
           {{#> empty-state-secondary}}
             {{#> button button--modifier="pf-m-link"}}
               Add group
--- a/src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
+++ b/src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
@@ -1,4 +1,19 @@
 {{#> gallery gallery--modifier="pf-m-gutter"}}
+    {{#> card card--id="card-empty-state" card--modifier="pf-m-hoverable pf-m-compact"}}
+      {{#> bullseye}}
+        {{#> empty-state empty-state--modifier="pf-m-xs"}}
+          <i class="fas fa-plus-circle pf-c-empty-state__icon"></i>
+          {{#> card-title}}
+            <p id="{{card--id}}-check-label">To get started, add a new group to your page.</p>
+          {{/card-title}}
+          {{#> empty-state-secondary}}
+            {{#> button button--modifier="pf-m-link"}}
+              Add group
+            {{/button}}
+          {{/empty-state-secondary}}
+        {{/empty-state}}
+      {{/bullseye}}
+    {{/card}}
     {{#> card card--id="card-1" card--modifier="pf-m-hoverable pf-m-compact"}}
       {{#> card-header}}
         <img src="/assets/images/pf-logo-small.svg" alt="PatternFly logo">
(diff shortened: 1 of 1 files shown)


mmenestr:
@mcarrano do you remember what our conclusion was for where we should add this card? As the first card, or as the last card in the table? I think we had said first card so that it was always visible on the first page but I can't remember now. 

@DaoDaoNoCode I'd also make the "addition card" the same size as the rest of the cards in the table!

src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
```diff
@@ -180,4 +180,17 @@
         Expose REST services and their APIs using Swagger specification.
       {{/card-body}}
     {{/card}}
+    {{#> card card--id="card-11" card--modifier="pf-m-compact pf-l-bullseye"}}
```
mcoker, line 183: While it looks like this works, we can't always rely on adding `.pf-l-bullseye` to an existing component, since there may be overlapping styles, a product may add the card styles after the bullseye styles, we may make an update to the card or bullseye that breaks the other component, etc. Also the bullseye layout is a separate component in react. So here we would just use `{{#> bullseye}}` as a child of `{{#> card}}`

DaoDaoNoCode:
@mmenestr The size of the card in this demo depends on the largest card size on that row. Thus, I would agree on setting the empty state card as the first.

src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
```diff
@@ -180,4 +180,17 @@
...
+      {{#> empty-state empty-state--modifier="pf-m-xs"}}
+        <i class="fas fa-plus-circle pf-c-empty-state__icon"></i>
+        {{#> title titleType="h1" title--modifier="pf-m-md"}}
```
mcoker, line 186: It probably makes more sense here to mimic the title in the other cards. So you can use `{{#> card-title}}<p> To get started...</p>{{/card-title}}`

src/patternfly/demos/Card/card-view-demo-template-gallery.hbs
mcoker, line 183: @mmenestr should this card be hoverable like the others?

mmenestr: Hmm I want to say yes, to keep the card interaction the same as usual? But are the other cards hoverable with the idea that they're clickable, or is it just an added visual indication for the user to know where their cursor is?

mcoker: @mmenestr I'm assuming the latter since they aren't selectable or anything like that. The react demo is interactive, and it just seems like a visual enhancement - https://patternfly-react.surge.sh/components/card/react-demos/card-view/

mmenestr: In that case I think let's just keep it consistent!

mcarrano:
@DaoDaoNoCode @mmenestr I don't think we ever reached a conclusion about whether the add card should be first or last in the grid. But I'm OK to make it first if it makes the demo work better. I also feel like this card should have a hover state if other cards do. I guess you could argue that it's interactive by having the add action associated with is. But more so it just feels wrong that it doesn't have a hover effect currently.

mcoker (approved):
LGTM! 🍕

mmenestr:
Just noticed one spacing thing, is it expected for there to be an extra 8px underneath the header? This is in addition to the 16px that already exists between elements
image

DaoDaoNoCode:
@mmenestr Yes, I use `{{#> card-title}}` here so there is a 8px bottom padding.
@mcoker What do you think? Should I change this back to `{{#> title titleType="h1" title--modifier="pf-m-md"}}` as it shows in the empty state examples?

mcoker:
@DaoDaoNoCode oh! 🤦‍♂️ I'm sorry, I told you wrong. What you had before is correct, the title is a part of the empty state, not the card. Though the heading level should probably  be an `<h2>`. Seems like all of the card titles should be an `<h2>` - @jessiehuff wdyt?

mmenestr:
Looks great now, thank you!!

mmenestr (approved):

PR #3798, opened 2021-01-14
fix(list): add new list variants

### LINKED ISSUE #3743: Add new List variants
Add new variants of the List component to include horizontal separators and icons.

Designs: https://marvelapp.com/prototype/7gc36d1

DaoDaoNoCode:
closes #3743 
@mcoker @mattnolting When the text is multiline, I could only align the icon/logo in the center. Do you have any idea about how to align it to the top of the text? Because if I simply align it to the top, the icon will not be on the same height with the text when the text is single line.

### DIFF
--- a/src/patternfly/components/List/examples/List.md
+++ b/src/patternfly/components/List/examples/List.md
@@ -77,6 +77,89 @@ cssPrefix: pf-c-list
 {{/list}}
 ```
 
+### With horizontal rules
+```hbs
+{{#> list list--type="ul" list--modifier="pf-m-plain pf-m-bordered"}}
+  <li>Donec blandit a lorem id convallis.</li>
+  <li>Integer in volutpat libero.</li>
+  <li>Donec a diam tellus.</li>
+  <li>Aenean nec tortor orci.</li>
+  <li>Vivamus maximus ultricies pulvinar.</li>
+{{/list}}
+```
+
+### With small icons
+```hbs
+{{#> list list--type="ul" list--modifier="pf-m-plain"}}
+  <li>
+    {{#> list-item-main}}
+      {{#> list-item-icon}}
+        <i class="pf-icon pf-icon-catalog" aria-hidden="true"></i>
+      {{/list-item-icon}}
+      {{#> list-item-text}}
+        List item one
+      {{/list-item-text}}
+    {{/list-item-main}}
+  </li>
+  <li>
+    {{#> list-item-main}}
+      {{#> list-item-icon}}
+        <i class="fas fa-key" aria-hidden="true"></i>
+      {{/list-item-icon}}
+      {{#> list-item-text}}
+        List item two
+      {{/list-item-text}}
+    {{/list-item-main}}
+  </li>
+  <li>
+    {{#> list-item-main}}
+      {{#> list-item-icon}}
+        <i class="pf-icon pf-icon-monitoring" aria-hidden="true"></i>
+      {{/list-item-icon}}
+      {{#> list-item-text}}
+        List item three
+      {{/list-item-text}}
+    {{/list-item-main}}
+  </li>
+{{/list}}
+```
+
+### With large icons
+```hbs
+{{#> list list--type="ul" list--modifier="pf-m-plain"}}
+  <li>
+    {{#> list-item-main}}
+      {{#> list-item-icon list-item-icon--modifier="pf-m-display-lg"}}
+        <i class="pf-icon pf-icon-catalog" aria-hidden="true"></i>
+      {{/list-item-icon}}
+      {{#> list-item-text}}
+        List item one
+      {{/list-item-text}}
+    {{/list-item-main}}
+  </li>
+  <li>
+    {{#> list-item-main}}
+      {{#> list-item-icon list-item-icon--modifier="pf-m-display-lg"}}
+        <i class="fas fa-key" aria-hidden="true"></i>
+      {{/list-item-icon}}
+      {{#> list-item-text}}
+        List item two
+      {{/list-item-text}}
+    {{/list-item-main}}
+  </li>
+  <li>
+    {{#> list-item-main}}
+      {{#> list-item-icon list-item-icon--modifier="pf-m-display-lg"}}
+        <i class="pf-icon pf-icon-monitoring" aria-hidden="true"></i>
+      {{/list-item-icon}}
+      {{#> list-item-text}}
+        List item three
+      {{/list-item-text}}
+    {{/list-item-main}}
+  </li>
+{{/list}}
+```
+
 ## Documentation
 ### Overview
 Non-inline lists can be nested up to any level.
@@ -87,3 +170,5 @@ Non-inline lists can be nested up to any level.
 | `.pf-c-list` | `<ul>, <ol>` | Initiates a list. **Required**  |
 | `.pf-m-inline` | `.pf-c-list` | Displays list items inline. |
 | `.pf-m-plain` | `.pf-c-list` |  Removes the list marker and base indentation. |
+| `.pf-m-bordered` | `pf-c-list` | Add horizontal divider between items in a list. |
+| `.pf-m-display-lg` | `pf-c-list-item-icon` | Show large icons or logos. |
--- a/src/patternfly/components/List/examples/List.md
+++ b/src/patternfly/components/List/examples/List.md
@@ -79,7 +79,7 @@ cssPrefix: pf-c-list
 
 ### With horizontal rules
 ```hbs
-{{#> list list--type="ul" list--modifier="pf-m-plain pf-m-bordered"}}
+{{#> list list--modifier="pf-m-plain pf-m-bordered"}}
   <li>Donec blandit a lorem id convallis.</li>
   <li>Integer in volutpat libero.</li>
   <li>Donec a diam tellus.</li>
@@ -90,10 +90,10 @@ cssPrefix: pf-c-list
 
 ### With small icons
 ```hbs
-{{#> list list--type="ul" list--modifier="pf-m-plain"}}
+{{#> list list--modifier="pf-m-plain"}}
   {{#> list-item}}
     {{#> list-item-icon}}
-      <i class="fas fa-book-open" aria-hidden="true"></i>
+      <i class="fas fa-book-open fa-fw" aria-hidden="true"></i>
     {{/list-item-icon}}
     {{#> list-item-text}}
       List item one
@@ -101,7 +101,7 @@ cssPrefix: pf-c-list
   {{/list-item}}
   {{#> list-item}}
     {{#> list-item-icon}}
-      <i class="fas fa-key" aria-hidden="true"></i>
+      <i class="fas fa-key fa-fw" aria-hidden="true"></i>
     {{/list-item-icon}}
     {{#> list-item-text}}
       List item two
@@ -109,7 +109,7 @@ cssPrefix: pf-c-list
   {{/list-item}}
   {{#> list-item}}
     {{#> list-item-icon}}
-      <i class="fas fa-desktop" aria-hidden="true"></i>
+      <i class="fas fa-desktop fa-fw" aria-hidden="true"></i>
     {{/list-item-icon}}
     {{#> list-item-text}}
       List item three
@@ -120,10 +120,10 @@ cssPrefix: pf-c-list
 
 ### With large icons
 ```hbs
-{{#> list list--type="ul" list--modifier="pf-m-plain pf-m-lg"}}
+{{#> list list--modifier="pf-m-plain pf-m-lg"}}
   {{#> list-item}}
     {{#> list-item-icon}}
-      <i class="fas fa-book-open" aria-hidden="true"></i>
+      <i class="fas fa-book-open fa-fw" aria-hidden="true"></i>
     {{/list-item-icon}}
     {{#> list-item-text}}
       List item one
@@ -131,7 +131,7 @@ cssPrefix: pf-c-list
   {{/list-item}}
   {{#> list-item}}
     {{#> list-item-icon}}
-      <i class="fas fa-key" aria-hidden="true"></i>
+      <i class="fas fa-key fa-fw" aria-hidden="true"></i>
     {{/list-item-icon}}
     {{#> list-item-text}}
       List item two
@@ -139,7 +139,7 @@ cssPrefix: pf-c-list
   {{/list-item}}
   {{#> list-item}}
     {{#> list-item-icon}}
-      <i class="fas fa-desktop" aria-hidden="true"></i>
+      <i class="fas fa-desktop fa-fw" aria-hidden="true"></i>
     {{/list-item-icon}}
     {{#> list-item-text}}
       List item three
--- a/src/patternfly/components/List/examples/List.md
+++ b/src/patternfly/components/List/examples/List.md
@@ -120,7 +120,7 @@ cssPrefix: pf-c-list
 
 ### With large icons
 ```hbs
-{{#> list list--modifier="pf-m-plain pf-m-lg"}}
+{{#> list list--modifier="pf-m-plain pf-m-icon-lg"}}
   {{#> list-item}}
     {{#> list-item-icon}}
       <i class="fas fa-book-open fa-fw" aria-hidden="true"></i>
@@ -159,4 +159,4 @@ Non-inline lists can be nested up to any level.
 | `.pf-m-inline` | `.pf-c-list` | Displays list items inline. |
 | `.pf-m-plain` | `.pf-c-list` |  Removes the list marker and base indentation. |
 | `.pf-m-bordered` | `pf-c-list` | Add horizontal divider between items in a list. |
-| `.pf-m-lg` | `pf-c-list` | Show all the icons or logos in the list large. |
+| `.pf-m-icon-lg` | `pf-c-list` | Show all the icons or logos in the list large. |
--- a/src/patternfly/components/List/examples/List.md
+++ b/src/patternfly/components/List/examples/List.md
@@ -158,5 +158,5 @@ Non-inline lists can be nested up to any level.
 | `.pf-c-list` | `<ul>, <ol>` | Initiates a list. **Required**  |
 | `.pf-m-inline` | `.pf-c-list` | Displays list items inline. |
 | `.pf-m-plain` | `.pf-c-list` |  Removes the list marker and base indentation. |
-| `.pf-m-bordered` | `pf-c-list` | Add horizontal divider between items in a list. |
-| `.pf-m-icon-lg` | `pf-c-list` | Show all the icons or logos in the list large. |
+| `.pf-m-bordered` | `.pf-c-list` | Add horizontal divider between items in a list. |
+| `.pf-m-icon-lg` | `.pf-c-list` | Show all the icons or logos in the list large. |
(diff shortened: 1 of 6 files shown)


mcoker:
> Do you have any idea about how to align it to the top of the text?

@DaoDaoNoCode in that case, you may try `align-items:  baseline;` - though that may align the icon too low relative to the text. If that's the case, you can try adding a top margin. We use a value like `#{pf-size-prem(6px)}`, which calls a the SCSS function `pf-size-prem()` and will convert the value passed to a `rem` unit.

src/patternfly/components/List/examples/List.md
```diff
@@ -77,6 +77,89 @@ cssPrefix: pf-c-list
...
+  <li>Aenean nec tortor orci.</li>
+  <li>Vivamus maximus ultricies pulvinar.</li>
+{{/list}}
+```
+
+### With small icons
+```hbs
+{{#> list list--type="ul" list--modifier="pf-m-plain"}}
+  <li>
+    {{#> list-item-main}}
```
mcoker, line 95: Just want to leave a thought that I don't think this is necessary, though I can see it being useful considering the way we've structured other lists like menu/dropown/select/etc, tree view, dual list selector, etc. 

Another way to do it would be to add the class `.pf-m-icon` to the `li`, and that changes the `li` to `display: flex; ...`.

I think I like the way you did it better, since if we want to add a "description" (small grey text under the main item text), we'll just add another element and reuse this structure as opposed to needing a class like `pf-m-description`.

src/patternfly/components/List/list.scss
```diff
@@ -1,11 +1,26 @@
...
+  --pf-c-list--m-bordered--li--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-list--m-bordered--li--BorderBottomColor: var(--pf-global--BorderColor--100);
+  --pf-c-list--m-bordered--li--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
+
+  // icon
+  --pf-c-list__item-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-list__item-icon--Color: var(--pf-global--icon--Color--light);
+  --pf-c-list__item-icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-list__item-icon--m-display-lg--MarginRight: var(--pf-global--spacer--md);
+  --pf-c-list__item-icon--m-display-lg--FontSize: var(--pf-global--icon--FontSize--lg);
```
mcoker, line 22: If this is just a larger sized icon for general use, this can probably just be `pf-m-lg`. If it's specifically for marketing use the `pf-m-display-lg` modifier is appropriate.

Also is it the case that if a user opts in to large icons, that should apply to the entire list? Or could you mix and match small and large icons?

Can you confirm @maryshak1996?

src/patternfly/components/List/list.scss
```diff
@@ -1,11 +1,26 @@
...
   --pf-c-list--nested--MarginTop: var(--pf-global--spacer--sm);
   --pf-c-list--nested--MarginLeft: var(--pf-global--spacer--sm);
   --pf-c-list--ul--ListStyle: var(--pf-global--ListStyle);
   --pf-c-list--li--MarginTop: var(--pf-global--spacer--sm);
+
+  // inline
   --pf-c-list--m-inline--li--MarginRight: var(--pf-global--spacer--lg);
 
+  // bordered
+  --pf-c-list--m-bordered--li--PaddingBottom: var(--pf-global--spacer--sm);
```
mcoker, line 13: Two questions about how the borders are drawn:

Screen Shot 2021-01-14 at 5 59 41 PM

Since we create borders between items, if you have a nested list in a parent list item, there is no border between the parent item and the first nested item. Is that intended? Are bordered lists valid when nested like this?

Also since we're using `border`, the border doesn't extend underneath the list marker.

@maryshak1996 can you confirm this looks as expected?

src/patternfly/components/List/list.scss
```diff
@@ -43,4 +75,11 @@
       }
     }
   }
+
+  &.pf-m-bordered {
+    > li:not(:last-child) {
```
mcoker, line 80: One thing about this is that it will create a fairly high specificity for this rule. There are a few ways you can improve that.

```
> li {
  border-bottom...
> li:last-child {
  border-bottom: 0;
```

Or you can rewrite that the same way with the universal selector, since the only valid child of a list is the `li` element, and the universal selector has 0 specificity.

```
> {
  border-bottom...
> :last-child {
  border-bottom: 0;
```

You can also write this using a [lobotomized owl selector](https://css-tricks.com/lobotomized-owls/), though it will change the way borders appear with nested lists.

```
> * + * {
  border-top...
```

mcoker (commented):
@DaoDaoNoCode great job on this! Left some feedback for review.

src/patternfly/components/List/examples/List.md
mattnolting: Agree w/@mcoker, I think you'd be better served to add `.pf-m-icon` to `li`. I'd use `display: grid` with `column-gap` to easily maintain alignment.

src/patternfly/components/List/list-item-main.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-list__item-main{{#if list-item-main--modifier}} {{list-item-main--modifier}}{{/if}}"
```
mattnolting, line 1: Don't think you need this element

src/patternfly/components/List/list.scss
mattnolting: +1 for universal, direct descendant 

```
> * {
  border-bottom...
> :last-child {
  border-bottom: 0;
```

src/patternfly/components/List/list.scss
```diff
@@ -22,6 +37,23 @@
...
   }
 
+  .pf-c-list__item-main {
+    display: flex;
+    align-items: center;
+  }
+
+  .pf-c-list__item-icon {
+    display: flex;
+    margin-right: var(--pf-c-list__item-icon--MarginRight);
```
mattnolting, line 47: Icon width can vary slightly, a more bulletproof implementation would be to use `grid` and `column-gap` instead of `margin-right`

mattnolting (commented):
Great work, @DaoDaoNoCode! Left a few comments.

src/patternfly/components/List/list.scss
DaoDaoNoCode: I am not sure whether we could use borders on nested list, but I am intended not to set border between the parent item and the first nested item. Waiting for designer's check.

src/patternfly/components/List/examples/List.md
mcoker: @mattnolting what do you think about needing to add a modifier for other things that will use a flex layout here? I think if we're stopping at icons, `pf-m-icon` would be fine. But I can see adding favorites, descriptions, actions, etc (like the other list components) - will we need a modifier for each? If we opt in to the markup @DaoDaoNoCode had previously, you can just place elements (icon, description, etc) in the list and it will lay itself out. Also if we went that route, I would be in favor of placing a margin/padding where we need it instead of using gap so we can just exclude items from the grid if we don't want them as opposed to needing to manage a class or something that removes the `gap`. wdyt?

src/patternfly/components/List/list.scss
maryshak1996: Hey guys, yes this looks as expected! Dropping the divider line between the parent and first child helps tie parent and child together to remind user of the relationship between the two (I feel like I'm talking about a family reunion right now not UX....)

DaoDaoNoCode: @maryshak1996 Thank you for checking. BTW, do you think whether the border should extend underneath the list marker or not?

maryshak1996: @DaoDaoNoCode I don't think that we want to extend it out under the markers, no. it looks a bit too heavy if it is.

src/patternfly/components/List/examples/List.md
```diff
@@ -77,6 +77,77 @@ cssPrefix: pf-c-list
 {{/list}}
 ```
 
+### With horizontal rules
+```hbs
+{{#> list list--type="ul" list--modifier="pf-m-plain pf-m-bordered"}}
```
mcoker, line 82:
```suggestion
{{#> list list--modifier="pf-m-plain pf-m-bordered"}}
```

You shouldn't need to specify `list--type`, its a `<ul>` by default - https://github.com/patternfly/patternfly/blob/bc15396391eb0414d45b51629b79b26610b14b7f/src/patternfly/components/List/list.hbs#L1

src/patternfly/components/List/examples/List.md
```diff
@@ -77,6 +77,77 @@ cssPrefix: pf-c-list
...
+  <li>Integer in volutpat libero.</li>
+  <li>Donec a diam tellus.</li>
+  <li>Aenean nec tortor orci.</li>
+  <li>Vivamus maximus ultricies pulvinar.</li>
+{{/list}}
+```
+
+### With small icons
+```hbs
+{{#> list list--type="ul" list--modifier="pf-m-plain"}}
```
mcoker, line 93:
```suggestion
{{#> list list--modifier="pf-m-plain"}}
```

src/patternfly/components/List/list.scss
```diff
@@ -22,6 +40,28 @@
...
+    display: flex;
+    min-width: var(--pf-c-list__item-icon--MinWidth);
+    margin-top: var(--pf-c-list__item-icon--MarginTop);
+    margin-right: var(--pf-c-list__item-icon--MarginRight);
+    font-size: var(--pf-c-list__item-icon--FontSize);
+    color: var(--pf-c-list__item-icon--Color);
+  }
+
+  &.pf-m-lg {
+    .pf-c-list__item-icon {
```
mcoker, line 57: Custom properties are inherited, so you shouldn't need to define this selector, and just apply it to `.pf-c-list.pf-m-lg`

src/patternfly/components/List/list.scss
```diff
@@ -43,4 +83,16 @@
...
   }
+
+  &.pf-m-bordered {
+    > * {
+      padding-bottom: var(--pf-c-list--m-bordered--li--PaddingBottom);
+      border-bottom: var(--pf-c-list--m-bordered--li--BorderBottomWidth) solid var(--pf-c-list--m-bordered--li--BorderBottomColor);
+    }
+
+    > :last-child {
+      padding-bottom: 0;
```
mcoker, line 94: This is fine, but typically we try to redefine properties via custom properties. So:

```
--pf-c-list--m-bordered--li--PaddingBottom: 0;
--pf-c-list--m-bordered--li--BorderBottomWidth: 0;
```

src/patternfly/components/List/list.scss
```diff
@@ -43,4 +83,16 @@
...
+      border-bottom: none;
```
mcoker, line 95: FYI a shorter way to write this would be `border: 0;` or `border-bottom: 0;`, they're both valid.

src/patternfly/components/List/list.scss
```diff
@@ -1,11 +1,29 @@
...
+  // inline
   --pf-c-list--m-inline--li--MarginRight: var(--pf-global--spacer--lg);
 
+  // bordered
+  --pf-c-list--m-bordered--li--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-list--m-bordered--li--BorderBottomColor: var(--pf-global--BorderColor--100);
+  --pf-c-list--m-bordered--li--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
+
+  // icon
+  --pf-c-list__item-icon--MinWidth: var(--pf-global--FontSize--xs);
```
mcoker, line 18: The min-width should match whatever the font-size is so that this fits a square icon or one that is more narrow than it is wide. All of the icons in react are squares, so that is what we're targeting mainly. If a user has icons that are wider than they are tall, they will need to update `--pf-c-list__item-icon--MinWidth` on the list.

src/patternfly/components/List/list.scss
```diff
@@ -1,11 +1,29 @@
...
+  --pf-c-list__item-icon--MarginTop: #{pf-size-prem(6px)};
+  --pf-c-list__item-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-list__item-icon--Color: var(--pf-global--icon--Color--light);
+  --pf-c-list__item-icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-list--m-lg__item-icon--MinWidth: var(--pf-global--FontSize--3xl);
```
mcoker, line 23: The same comment here about the `min-width` - it should match what the font-size is so that the space provided works for a square icon or one more narrow than it is tall.

src/patternfly/components/List/list.scss
```diff
@@ -22,6 +40,28 @@
...
+  .pf-c-list__item-icon {
+    display: flex;
+    min-width: var(--pf-c-list__item-icon--MinWidth);
+    margin-top: var(--pf-c-list__item-icon--MarginTop);
+    margin-right: var(--pf-c-list__item-icon--MarginRight);
+    font-size: var(--pf-c-list__item-icon--FontSize);
+    color: var(--pf-c-list__item-icon--Color);
+  }
+
+  &.pf-m-lg {
```
mcoker, line 56: I'm curious what you think @mcarrano @maryshak1996 about a list component with the variation `pf-m-lg` as a way to make the icons in the list larger. This would be the large list variation. I'm wondering if this should be `pf-m-large-icon` or something more specific like that, and reserve `pf-m-lg` for making a larger variation of the list itself (bigger font, more padding, etc).

src/patternfly/components/List/examples/List.md
```diff
@@ -77,6 +77,77 @@ cssPrefix: pf-c-list
...
+  {{#> list-item}}
+    {{#> list-item-icon}}
+      <i class="fas fa-book-open" aria-hidden="true"></i>
```
mcoker, line 96: in instances like this where we want to ensure all of the icons have the same width, font awesome has a utility class `.fa-fw` that will give an icon a fixed width. You can apply that to all of the icons that you want to have a fixed width relative to each other.
```suggestion
      <i class="fas fa-book-open fa-fw" aria-hidden="true"></i>
```

DaoDaoNoCode:
@mcoker Updated all. Thank you for the suggestions!

src/patternfly/components/List/examples/List.md
```diff
@@ -87,3 +158,5 @@ Non-inline lists can be nested up to any level.
 | `.pf-c-list` | `<ul>, <ol>` | Initiates a list. **Required**  |
 | `.pf-m-inline` | `.pf-c-list` | Displays list items inline. |
 | `.pf-m-plain` | `.pf-c-list` |  Removes the list marker and base indentation. |
+| `.pf-m-bordered` | `pf-c-list` | Add horizontal divider between items in a list. |
```
mcoker, line 161: It might be a little confusing that we call it a horizontal divider but don't use the divider component. This is a little simpler. 

```suggestion
| `.pf-m-bordered` | `pf-c-list` | Adds a border between list items. |
```

mcoker (approved):
This looks great! Just want to get clarification on the large icon variation name before committing.

src/patternfly/components/List/list.scss
```diff
@@ -22,6 +40,26 @@
     list-style: var(--pf-c-list--ul--ListStyle);
   }
 
+  .pf-c-list__item {
+    display: flex;
+  }
+
+  .pf-c-list__item-icon {
+    display: flex;
```
mcoker, line 48: Looks like this needs `flex-shrink: 0;`, otherwise it will shrink if there is a lot of content in the `item-text` element.

mcoker: Also needs `align-items: start;` or it will stretch whatever is in it vertically. You can see this if you put a non-square image in the `__item-icon` element, or put a square icon in and give it a modified `width`

mcoker: Actually, looking at this some more re: `align-items`, I'm not sure why this element has `display: flex;` set. If it's because the icons are too tall without `display: flex;`, it's because `__item-icon` will have a default line-height of `1.5`, which with a `16px` font, will make the height of the element `24px`. Setting `display: flex;` will give the box block formatting, removing the extra space created by `line-height`. A way to remove that extra height would be to set `line-height: 1;`, so the `line-height` will always match the current `font-size`. Try that and see if you can remove `display: flex;`, and then you won't need `align-items`, either.

src/patternfly/components/List/list.scss
mcarrano: I agree @mcoker

src/patternfly/components/List/list.scss
```diff
@@ -1,11 +1,29 @@
...
+
+  // icon
+  --pf-c-list__item-icon--MinWidth: var(--pf-global--icon--FontSize--sm);
+  --pf-c-list__item-icon--MarginTop: #{pf-size-prem(6px)};
+  --pf-c-list__item-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-list__item-icon--Color: var(--pf-global--icon--Color--light);
+  --pf-c-list__item-icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-list--m-lg__item-icon--MinWidth: var(--pf-global--icon--FontSize--lg);
+  --pf-c-list--m-lg__item-icon--MarginRight: var(--pf-global--spacer--md);
+  --pf-c-list--m-lg__item-icon--FontSize: var(--pf-global--icon--FontSize--lg);
```
mattnolting, line 25:
```suggestion
  --pf-c-list--m-icon-lg__item-icon--MinWidth: var(--pf-global--icon--FontSize--lg);
  --pf-c-list--m-icon-lg__item-icon--MarginRight: var(--pf-global--spacer--md);
  --pf-c-list--m-icon-lg__item-icon--FontSize: var(--pf-global--icon--FontSize--lg);
```

src/patternfly/components/List/list.scss
mattnolting: I'd go with, `.pf-m-icon-lg` to add a specific name and to be consistent w/`.pf-m-display-lg`.

mattnolting (commented):
Great work! Left a few comments.

src/patternfly/components/List/list.scss
DaoDaoNoCode: @mcoker It works, thank you!

DaoDaoNoCode:
@mcoker Do you have any idea about why I always fails the lint test in CircleCI?

mcoker:
> Do you have any idea about why I always fails the lint test in CircleCI?

It's a CSS size error - those are safe to ignore. They're more warnings than an error. Sometimes rebasing can help, I think?

src/patternfly/components/List/examples/List.md
mcoker, line 161:
```suggestion
| `.pf-m-bordered` | `.pf-c-list` | Add horizontal divider between items in a list. |
```

src/patternfly/components/List/examples/List.md
```diff
@@ -87,3 +158,5 @@ Non-inline lists can be nested up to any level.
...
+| `.pf-m-icon-lg` | `pf-c-list` | Show all the icons or logos in the list large. |
```
mcoker, line 162:
```suggestion
| `.pf-m-icon-lg` | `.pf-c-list` | Show all the icons or logos in the list large. |
```

DaoDaoNoCode: Nice catch! Updated, thanks.

mcoker (approved):
L🍕TM!!

mcarrano (approved):
Looks good to me. Thanks @DaoDaoNoCode

mattnolting (approved):
Great work @DaoDaoNoCode, thanks!

Issue #3804, opened 2021-01-20
Resizable drawer var enhancements


mcoker:
From talking with @kmcfaul about the react features we're supporting around sizing, I think the CSS can help out with those and be more performant and easier for the react component to manage.

The react component will allow a user to set a base size (that may be different than our `pf-m-width*` modifiers). It also allows the user to specify a minimum and maximum size for the drawer. I think that means we need to:

* Add ability to set a custom width. Bonus points for widths at different breakpoints
* Add ability to define min and max widths via vars and use the CSS to keep it within bounds (`clamp()` should allow for a min/main/max size to use with `flex-basis).
* Allow both `--pf-c-drawer__panel--FlexBasis` and `--pf-c-drawer--m-resizable__panel--FlexBasis` to override the width prop (doesn't currently)
* Make sure the var overrides exist on the elements that make the most sense. This probably means keeping all of these vars on `__panel`

PR #3805, opened 2021-01-21
fix(drawer): changes to vars used to redefine panel flex-basis

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3804

Introduces these vars as a means to control the drawer width:

| -- | -- | -- |
| `--pf-c-drawer__panel--md--FlexBasis--min` | `.pf-c-drawer__panel` | Defines the drawer panel minimum size. |
| `--pf-c-drawer__panel--md--FlexBasis` | `.pf-c-drawer__panel` | Defines the drawer panel size. |
| `--pf-c-drawer__panel--md--FlexBasis--max` | `.pf-c-drawer__panel` | Defines the drawer panel maximum size. |

These are all applied > the medium breakpoint, so they do not apply to the mobile layout, and they're applied via `flex-basis: clamp(--pf-c-drawer__panel--md--FlexBasis--min, --pf-c-drawer__panel--md--FlexBasis, --pf-c-drawer__panel--md--FlexBasis--max)`. 

The var `--pf-c-drawer__panel--FlexBasis` still serves to define `flex-basis` on `.pf-c-drawer__panel`, and can be used to override the panel with on all viewport widths. So it works like it did before, we're just dropping the `--m-resizable--FlexBasis` vars we added recently in favor of a single `--md--FlexBasis` var that has an initial value by the component and can be overridden by the consumer, our modifiers, and the resizable drawer that modifies the panel size on the fly.

The react component's `min/maxSize` props can set the `--min` and `--max` vars inline. The `--min` var also has an initial value in our component of the splitter's width (or height in a panel-bottom variation) on resizable drawers so the panel is always at least as wide/tall as the splitter.

And all of these vars are applied to `.pf-c-drawer__panel`.

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -27,7 +27,6 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --pf-c-drawer--m-panel-bottom__panel--md--MinHeight: 50%;
   --pf-c-drawer--m-panel-bottom__panel--xl--MinHeight: #{pf-size-prem(300px)};
   --pf-c-drawer--m-panel-bottom__panel--xl--FlexBasis: #{pf-size-prem(300px)};
-  --pf-c-drawer__panel--m-resizable--FlexBasis: initial; // falls back by default
   --pf-c-drawer__panel--m-resizable--FlexDirection: row;
   --pf-c-drawer__panel--m-resizable--md--FlexBasis--min: var(--pf-c-drawer__splitter--m-vertical--Width);
   --pf-c-drawer--m-panel-bottom__panel--m-resizable--FlexDirection: column;
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -12,24 +12,26 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --pf-c-drawer__content--ZIndex: var(--pf-global--ZIndex--xs);
 
   // Panel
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Drawer/drawer.scss
```diff
@@ -12,24 +12,26 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
...
+  --pf-c-drawer__panel--FlexBasis: 100%;
+  --pf-c-drawer__panel--md--FlexBasis--min: 0%;
+  --pf-c-drawer__panel--md--FlexBasis: 50%;
+  --pf-c-drawer__panel--md--FlexBasis--max: 100%;
+  --pf-c-drawer__panel--xl--MinWidth: #{pf-size-prem(450px)};
+  --pf-c-drawer__panel--xl--FlexBasis: #{pf-size-prem(450px)};
+  --pf-c-drawer--m-panel-bottom__panel--md--MinHeight: 50%;
+  --pf-c-drawer--m-panel-bottom__panel--xl--MinHeight: #{pf-size-prem(300px)};
+  --pf-c-drawer--m-panel-bottom__panel--xl--FlexBasis: #{pf-size-prem(300px)};
+  --pf-c-drawer__panel--m-resizable--FlexBasis: initial; // falls back by default
```
DaoDaoNoCode, line 30: This variable is not used in any class, do we still need to keep it here?

mcoker: Nope, good find! Removed.

DaoDaoNoCode (approved):
@mcoker Thanks, LPTM! 😎

mattnolting (approved):
This is awesome! LPTM 🎉

PR #3807, opened 2021-01-22
fix(jump-links): added vertical jump links demos

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3560


DaoDaoNoCode (approved):
@mcoker Looks perfect! Approved. 👏

mcarrano (commented):
These look great @mcoker . The only thing I'm questioning is what the horizontal jump links should do as we shrink the viewport. When you have here works for a limited number of links, but would break if more links were added. Perhaps at the mobile breakpoint these should collapse into the dropdown header like the vertical links do? I don't think the designs explicitly addressed this. @ajacobs21e what do you expect?

ajacobs21e:
@mcarrano @mcoker The design uses a disclosure from the Web UX Kit, which is like a single accordion
https://www.sketch.com/s/00432cc6-fd0c-454f-bdd6-c9720ceebbdc/p/jump-links

mcarrano:
You need to give us access to this on Sketch Cloud @ajacobs21e

mattnolting (approved):
Wow, don't see this everyday.

Screen Shot 2021-02-09 at 10 54 38 AM

LPTM, nice work!!

mcoker:
@mcarrano wdyt, does this look good or do you think it still needs more work?

mcarrano (approved):
This looks great @mcoker !

PR #3816, opened 2021-01-27
feat(description-list): added help text support

### LINKED ISSUE #3808: Definition list: Add option to create an underlined label for more info
Want to add the option to declare underlined labels that will open more info when the user clicks on them.

Designs: https://marvelapp.com/prototype/7j26ab6

mattnolting:
closes #3808

### DIFF
--- a/src/patternfly/components/DescriptionList/description-list-text.hbs
+++ b/src/patternfly/components/DescriptionList/description-list-text.hbs
@@ -1,5 +1,5 @@
-<{{#if description-list-text--type}}{{description-list-text--type}}{{else}}div{{/if}} class="pf-c-description-list__text{{#if description-list-text--modifier}} {{description-list-text--modifier}}{{/if}}{{#if description-list-text--IsClickable}} pf-m-clickable{{/if}}"
-  {{#if description-list-text--IsClickable}}
+<{{#if description-list-text--type}}{{description-list-text--type}}{{else}}div{{/if}} class="pf-c-description-list__text{{#if description-list-text--modifier}} {{description-list-text--modifier}}{{/if}}{{#if description-list-text--IsHelp}} pf-m-help-text{{/if}}"
+  {{#if description-list-text--IsHelp}}
     role="button"
     type="button"
     tabindex="0"
--- a/src/patternfly/components/DescriptionList/description-list-text.hbs
+++ b/src/patternfly/components/DescriptionList/description-list-text.hbs
@@ -1,4 +1,9 @@
-<{{#if description-list-text--type}}{{description-list-text--type}}{{else}}div{{/if}} class="pf-c-description-list__text{{#if description-list-text--modifier}} {{description-list-text--modifier}}{{/if}}"
+<{{#if description-list-text--type}}{{description-list-text--type}}{{else}}div{{/if}} class="pf-c-description-list__text{{#if description-list-text--modifier}} {{description-list-text--modifier}}{{/if}}{{#if description-list-text--IsClickable}} pf-m-clickable{{/if}}"
+  {{#if description-list-text--IsClickable}}
+    role="button"
+    type="button"
+    tabindex="0"
+  {{/if}}
   {{#if description-list-text--attribute}}
     {{{description-list-text--attribute}}}
   {{/if}}>
--- a/src/patternfly/components/DescriptionList/description-list.scss
+++ b/src/patternfly/components/DescriptionList/description-list.scss
@@ -31,11 +31,12 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "md", "lg",
   --pf-c-description-list--m-auto-fit--GridTemplateColumns--min: #{pf-size-prem(250px)};
   --pf-c-description-list--m-auto-fit--GridTemplateColumns--minmax--min: var(--pf-c-description-list--m-auto-fit--GridTemplateColumns--min);
 
-  // clickable text
-  --pf-c-description-list__text--m-clickable--TextDecorationColor: var(--pf-global--palette--black-500);
-  --pf-c-description-list__text--m-clickable--hover--TextDecorationColor: var(--pf-global--palette--black-900);
-  --pf-c-description-list__text--m-clickable--TextDecorationThickness: var(--pf-global--BorderWidth--sm);
-  --pf-c-description-list__text--m-clickable--TextDecorationOffset: #{pf-size-prem(4px)};
+  // help text
+  --pf-c-description-list__text--m-help-text--TextDecorationColor: var(--pf-global--BorderColor--200);
+  --pf-c-description-list__text--m-help-text--TextDecorationThickness: var(--pf-global--BorderWidth--sm);
+  --pf-c-description-list__text--m-help-text--TextDecorationOffset: #{pf-size-prem(4px)};
+  --pf-c-description-list__text--m-help-text--hover--TextDecorationColor: var(--pf-global--Color--100);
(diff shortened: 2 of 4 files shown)


src/patternfly/components/DescriptionList/description-list.scss
```diff
@@ -31,6 +31,12 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "md", "lg",
   --pf-c-description-list--m-auto-fit--GridTemplateColumns--min: #{pf-size-prem(250px)};
   --pf-c-description-list--m-auto-fit--GridTemplateColumns--minmax--min: var(--pf-c-description-list--m-auto-fit--GridTemplateColumns--min);
 
+  // clickable text
+  --pf-c-description-list__text--m-clickable--TextDecorationColor: var(--pf-global--palette--black-500);
+  --pf-c-description-list__text--m-clickable--hover--TextDecorationColor: var(--pf-global--palette--black-900);
```
mcoker, line 36:
```suggestion
  --pf-c-description-list__text--m-clickable--hover--TextDecorationColor: var(--pf-global--Color--100);
```

src/patternfly/components/DescriptionList/description-list.scss
```diff
@@ -31,6 +31,12 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "md", "lg",
   --pf-c-description-list--m-auto-fit--GridTemplateColumns--min: #{pf-size-prem(250px)};
   --pf-c-description-list--m-auto-fit--GridTemplateColumns--minmax--min: var(--pf-c-description-list--m-auto-fit--GridTemplateColumns--min);
 
+  // clickable text
+  --pf-c-description-list__text--m-clickable--TextDecorationColor: var(--pf-global--palette--black-500);
```
mcoker, line 35:
```suggestion
  --pf-c-description-list__text--m-clickable--TextDecorationColor: var(--pf-global--BorderColor--200);
```

src/patternfly/components/DescriptionList/description-list.scss
```diff
@@ -81,6 +87,21 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "md", "lg",
...
+.pf-c-description-list__text {
+  &.pf-m-clickable {
+    text-decoration: underline;
+    cursor: pointer;
+    text-decoration-style: dashed;
+    text-decoration-thickness: var(--pf-c-description-list__text--m-clickable--TextDecorationThickness);
+    text-underline-offset: var(--pf-c-description-list__text--m-clickable--TextDecorationOffset);
+    text-decoration-color: var(--pf-c-description-list__text--m-clickable--TextDecorationColor);
+
+    &:hover {
```
mcoker, line 99: Looks like the design also calls for a focus state.

src/patternfly/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -9,7 +9,12 @@ cssPrefix: pf-c-description-list
 
 ### Default
 ```hbs
-{{> description-list__example description-list--title="Default DL" description-list--header="Test"}}
+{{> description-list__example description-list--title="Default DL"}}
+```
+
+### Clickable term
+```hbs
+{{> description-list__example description-list--title="Clickable term" description-list-term--TextIsClickable="true"}}
```
mcoker, line 17: You could also just define `description-list-text--IsClickable="true"` here and you wouldn't need to set that var again on `{{#> description-list-text}}`

src/patternfly/components/DescriptionList/description-list.scss
```diff
@@ -81,6 +87,21 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "md", "lg",
   }
 }
 
+.pf-c-description-list__text {
+  &.pf-m-clickable {
```
mcoker, line 91: I wonder if "clickable" is too broad, and could conflict with term that's clickable in some other way but isn't underlined because  there isn't more info. The queue is more that it has more info than being interactive. I was thinking `.pf-m-help` since we use "help" in the table and form for the help icon popovers, but that could conflict with those icons if we add them. Maybe `.pf-m-help-text`? It should also make sense if we use it other places, which I imagine we might as an alternative to a bunch of help icons.

src/patternfly/components/DescriptionList/examples/DescriptionList.md
mattnolting: Yes, that is true, but since we're using `__text` in both `__term` and `__description`, I had to isolate `__term > __text` from `__description > __text` so they're not all clickable.

mcoker: Ah, that makes sense!

mcoker (approved):
LPTM!!

src/patternfly/components/DescriptionList/description-list-text.hbs
```diff
@@ -1,4 +1,9 @@
-<{{#if description-list-text--type}}{{description-list-text--type}}{{else}}div{{/if}} class="pf-c-description-list__text{{#if description-list-text--modifier}} {{description-list-text--modifier}}{{/if}}"
+<{{#if description-list-text--type}}{{description-list-text--type}}{{else}}div{{/if}} class="pf-c-description-list__text{{#if description-list-text--modifier}} {{description-list-text--modifier}}{{/if}}{{#if description-list-text--IsHelp}} pf-m-help-text{{/if}}"
+  {{#if description-list-text--IsHelp}}
```
mcoker, line 2: 👍

src/patternfly/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -109,6 +115,7 @@ cssPrefix: pf-c-description-list
 | `.pf-c-description-list__term` | `<dt>` | Initiates a description list component term. **Required** |
 | `.pf-c-description-list__description` | `<dd>` | Initiates a description list component description. **Required** |
 | `.pf-c-description-list__text` | `<span>`, `<div>` | Initiates a description list component text element. Use a `<span>` when a child of `.pf-c-description-list__term`. **Required** |
+| `.pf-m-help-text` | `.pf-c-description-list__text` | Modifies the description list component text to be clickable. |
```
mcoker, line 118: Oh just noticed this. Maybe something more help text related in the description, like "Modifies the description list component text to indicate there is more info available for the term text." Or less wordy but less consistent with the others, "Indicates there is more info available for the description list component term text."

mcarrano (approved):
Looks good to me. @maryshak1996 can you also check to make sure this is consistent with your design intent? Thanks.

jessiehuff (approved):
I think this looks good, it's keyboard and VO accessible from what I can tell. My only concerns are more down the road with react. Do we expect our customers to manually add a lot of the interaction that would typically be expected with a button element or do we know the expectations for that? When we used the [inline link as span button](https://www.patternfly.org/v4/components/button#inline-link-as-span) example, we had to warn customers that if they use it, they have to consider all of the additional button interactions that are the browser default. Is that what we'd do here too or would that be something we'd be trying to handle for them?

maryshak1996 (approved):

mcarrano:
Good question @jessiehuff . The intent is to restrict this to only displaying pop-up help on an element (i.e. opening a Popover). For the React component, maybe we would bake that in? So upon clicking the label we will open a Popover. You can put whatever you want in the popover, but it always just displays a popover. What do you think @tlabaj ?

Issue #3810, opened 2021-01-22
Update Label styles


mcarrano:
Update styles of Label component to work better on a gray background.

Designs: https://marvelapp.com/prototype/553965i

PR #3817, opened 2021-01-28
fix(label): updated colors to work better on a gray background

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3810

### DIFF
--- a/src/patternfly/components/Label/label.scss
+++ b/src/patternfly/components/Label/label.scss
@@ -10,8 +10,8 @@
   --pf-c-label--FontSize: var(--pf-global--FontSize--sm);
 
   // Border base styles
-  --pf-c-label__content--before--BorderWidth: 0;
-  --pf-c-label__content--before--BorderColor: transparent;
+  --pf-c-label__content--before--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-label__content--before--BorderColor: var(--pf-global--BorderColor--100);
 
   // Outline
   --pf-c-label--m-outline--BackgroundColor: var(--pf-global--BackgroundColor--100);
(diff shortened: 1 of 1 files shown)


mcarrano (commented):
These look great to me. But I will defer to @maryshak1996 for final design approval.

maryshak1996:
@mcoker this is looking great!
A couple minor things though (mostly just some tweaks to the text colors on the outline versions, so that they look more vibrant, while still keeping AA compliance):

- For the text color in the orange outline label, can we use `var(--pf-color--orange--150)`?
- For the text color in the red outline label, can we use `var(--pf-color--red--100)`?
- For the text color in the blue outline label, can we use `var(--pf-color--blue--400)`?
- For the text color in the green outline label, can we use `var(--pf-color--green--500)`?
- For the text color in the cyan outline label, can we use `var(--pf-color--cyan--400)`?
- For the text color in the purple outline label, can we use `var(--pf-color--purple--500)`?

And one update to the filled labels:

- For the grey filled label, the background should use `var(--pf-global--BackgroundColor--150)` instead of `var(--pf-global--BackgroundColor--200)`

mcoker:
> For the text color in the orange outline label, can we use var(--pf-color--orange--150)?

@maryshak1996 we don't have orange 150 - should I add a new color? Also just an observation, with the text color update, all of the labels except orange and cyan are changing the text color to match the icon color. If we were to follow suit with orange and cyan, the text would be orange-300 (instead of orange-150) and cyan-300 (instead of cyan-400)

mcoker:
@maryshak1996 we also don't have backgroundcolor-150, only 100 and 200:

https://github.com/patternfly/patternfly/blob/db4e7370e459a95cf15a283d7923a97e027e478e/src/patternfly/sass-utilities/scss-variables.scss#L26-L27

maryshak1996:
> @maryshak1996 we don't have orange 150 - should I add a new color? Also just an observation, with the text color update, all of the labels except orange and cyan are changing the text color to match the icon color. If we were to follow suit with orange and cyan, the text would be orange-300 (instead of orange-150) and cyan-300 (instead of cyan-400)

Woops I made a typo here -- I meant 500, not 150! 

> @maryshak1996 we don't have orange 150 - should I add a new color? Also just an observation, with the text color update, all of the labels except orange and cyan are changing the text color to match the icon color. If we were to follow suit with orange and cyan, the text would be orange-300 (instead of orange-150) and cyan-300 (instead of cyan-400)

I think that the issue with making them match is that the shades we are using for orange and cyan in the icons don't quite pass for accessibility for text color (but they pass accessibility for graphical elements and are a bit more vibrant, which is why we want to keep the icons in the shades that they are, and just accept having a very slight color difference between the text and icon)

maryshak1996:
> > For the grey filled label, the background should use var(--pf-global--BackgroundColor--150)
> 
> @maryshak1996 we also don't have backgroundcolor-150, only 100 and 200:
> 
> https://github.com/patternfly/patternfly/blob/db4e7370e459a95cf15a283d7923a97e027e478e/src/patternfly/sass-utilities/scss-variables.scss#L26-L27

Sorry I meant, `--pf-global--palette--black-150` -- can we use that as the background color for the label? this just gives a bit more contrast when over a grey background

maryshak1996 (approved):
Looks amazing! Thanks for making those color updates!

mattnolting (approved):
LPTM 🚢

Issue #3813, opened 2021-01-25
Drawer: support optional gray background


mcarrano:
Allow the Drawer to optionally have a light gray background as shown: 
![image](https://user-images.githubusercontent.com/6619860/105729724-82d91880-5efb-11eb-94c4-8bf367461009.png)

Design issue: https://github.com/patternfly/patternfly-design/issues/935

PR #3819, opened 2021-01-28
fix(drawer): added light-200 variation for grey background

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3813

@mcarrano @mceledonia I added this variation as `pf-m-light-200`. I'd like to be consistent as we use a variation like this going forward. Our page sections are `--pf-global--BackgroundColor-200` by default, and we use `pf-m-light` as the variation name to turn it white, so we have that precedent to work with. I'm thinking we can keep `pf-m-light` to represent white/`--pf-global--BackgroundColor-100`, and use `pf-m-light-200` for `--pf-global--BackgroundColor-200`. That will also allow us to apply this to other components in the same way. Sound OK?

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -5,10 +5,12 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 .pf-c-drawer {
   // Section
   --pf-c-drawer__section--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-drawer__section--m-light-200--BackgroundColor: var(--pf-global--BackgroundColor--200);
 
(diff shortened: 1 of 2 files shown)


mcarrano (changes requested):
@mcoker I don't have a strong feeling about the naming so I'm good with your approach.

I'm wondering if we can change this example. The specific ask was to have an inline gray drawer panel with a white content area. See the mock here: https://github.com/patternfly/patternfly/issues/3813 Could we create an example that is closer to this? It's the way I see this being used. Thanks.

mcoker:
@mcarrano updated

mcarrano (approved):
Looks great @mcoker

mattnolting (approved):
Beautiful!! LPTM 🌮

Issue #3770, opened 2020-12-18
Login: Add option to show/hide password


mcarrano:
Add the ability to show or hide the password entered into the Password field on the Login form.

![Show-hide password.png](https://images.zenhubusercontent.com/5ce816524e6c594dc1b0df4d/671a4301-0129-4a85-aeb2-ccaf08af8bef)

Show icon: fas fa-eye
Hide icon: fas fa-eye-slash

PR #3820, opened 2021-01-28
feat(login): added support for hide/show password

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3770

### DIFF
--- a/src/patternfly/components/Login/__login-footer.hbs
+++ b/src/patternfly/components/Login/__login-footer.hbs
@@ -1,0 +1,20 @@
+{{#> login-footer}}
+  <p>This is placeholder text only. Use this area to place any information or introductory message about your application that may be relevant to users.</p>
+  {{#> list list--modifier="pf-m-inline"}}
+    <li>
+      <a href="#">
+        Terms of use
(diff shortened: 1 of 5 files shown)


mcarrano (commented):
Looks good to me @mcoker . Is this what you expected @ajacobs21e ?

My only comment is whether the names of the examples should be swapped, i..e. when the password is hidden, the action is to "show password" and vice versa.

mcoker:
> My only comment is whether the names of the examples should be swapped

Agree, I wasn't sure, whatever you think is more intuitive. It's kind of confusing since "show password" as a description of the action's function has the password hidden, and "hide password" has the password shown. We could rename them "password hidden" and "password shown"?

ajacobs21e:
I agree with Matt. The name is about the action not the end state.
Show password: password becomes shown (goes from hashed to text)
Hide password: password becomes hidden (goes from text to hashed)

mcoker:
@mcarrano @ajacobs21e updated

mattnolting (approved):
Nice! LPTM! 🎉

Issue #3311, opened 2020-07-20
Update check and radio components so that they do not need a label


mcoker:
Let's allow these components to only contain the input. Currently the grid layout relies on 2 columns with a `grid-gap` that creates space when there is only 1 column. This is so we can style checks/radios elements in the check/radio component, and then use those in places like bulk select where there is only a check and no label.

Issue #3288, opened 2020-07-14
Table checkbox UX is not correct


karelhala:
When you hover over a checkbox in table the cursor remains the same (`default`) whereas it should change to a hand (`pointer`). Also the target area for checkbox is incorrect it should be slightly larger than just the checkbox so users has easier time clicking in the area.

karelhala:
ping @mcoker @mattnolting

bclarhk:
I believe industry standard click target is still 44x44. Feel free to ping me with any questions!

mcoker:
cc @jessiehuff @mcarrano we've discussed the clickable area of checkboxes (and radios) in the past. I created a POC PR (https://github.com/patternfly/patternfly/pull/2745) with custom designed checkboxes and radios that increased the target area - https://patternfly-next-pr-2745.surge.sh/documentation/core/components/check

Do we want to consider adding that in? I think the concern was that it could be a breaking change since the elements take up more space and we can't guarantee it won't impact layouts adversely. There are other ways of increasing the target area without changing the checkbox/radio size, too, like this https://codepen.io/mcoker/pen/XWXPRww. In that example, the grey background indicates the clickable area, and you can change its size by modifying the `--inset` value in the top/middle (CSS) pane.

mcarrano:
I know this has come up before and I would be in favor of updating this as long as we can accomplish it in a non-breaking way. I can add this to the roadmap for further investigation.

jgiardino:
I just posted a [comment in a similar issue](https://github.com/patternfly/patternfly/issues/865#issuecomment-663162700) regarding WCAG requirements (TLDR, there aren't any).

@mcoker -  I don't get the larger target area or background color in Firefox in this pen >>> https://codepen.io/mcoker/pen/XWXPRww

Something else to consider is to focus only on checkboxes that display in tables and datalists (i.e. checkboxes that aren't accompanied by `<label>` elements), and wrap the checkbox with a `<label>`. This would allow us to control the size of the clickable area by controlling the size of the label. This seems a bit hacky, but is the best suggestion I have for a major shortcoming of modern web development. If we did this, we might want to move the aria-label from the  `<input>` to a sr-only span inside the label (because it seems wrong to use `<label>` without including an actual label in it). Here's a codepen that illustrates this: https://codepen.io/jenngiardino/pen/Bajbprj

mcoker:
>  I don't get the larger target area or background color in Firefox in this pen

[...]

PR #3821, opened 2021-01-29
feat(radio, check): added standalone variation, cursor styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3288
fixes https://github.com/patternfly/patternfly/issues/3311

@mcarrano @mattnolting @mceledonia this adds a "pf-m-standalone" variation to the check and radio components for use when the radio/check is just a single input with no label/description. Previously, the components required a label. This PR updates our beta components that use a regular checkbox (tree view, dual list selector) and card view demo checkboxes to use the standalone variation of the checkbox component instead of regular checkboxes.

I tested using the standalone checkbox/radio components in our existing non-beta components (table, data list, split button dropdown toggles) and it works fine, though you could argue it might be breaking if users are relying on the existing input that's used in those components. It's probably OK, but I wouldn't want to inadvertently introduce a layout problem there if users have custom styles for the existing inputs that don't transfer to the new radio/checkbox component. So to address the pointer style issue from #3288 in existing components that have regular checkboxes , I've added `pointer: cursor;` styles to those inputs. And in a breaking change, we can update table, data list, etc to use the checkbox/radio components.

Re: the target area/size for inputs, I think we should explore that in a separate issue. I'm thinking to be able to do that consistently/reliably, we'll want to use custom checkbox/radio elements.

### DIFF
--- a/src/patternfly/components/Check/examples/Check.md
+++ b/src/patternfly/components/Check/examples/Check.md
@@ -63,7 +63,7 @@ cssPrefix: pf-c-check
 ### Standalone input
 ```hbs
 {{#> check check--modifier="pf-m-standalone"}}
-  {{#> check-input check-input--attribute='id="check-standalone-input" name="check-standalone-input"'}}{{/check-input}}
+  {{#> check-input check-input--attribute='id="check-standalone-input" name="check-standalone-input" aria-label="Standalone input"'}}{{/check-input}}
 {{/check}}
 ```
 
--- a/src/patternfly/components/Check/examples/Check.md
+++ b/src/patternfly/components/Check/examples/Check.md
@@ -60,6 +60,13 @@ cssPrefix: pf-c-check
 {{/check}}
 ```
 
+### Standalone input
+```hbs
+{{#> check check--modifier="pf-m-standalone"}}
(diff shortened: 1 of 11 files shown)


mcarrano (approved):
I like it @mcoker !

src/patternfly/components/Check/examples/Check.md
```diff
@@ -78,4 +85,5 @@ If you extend this component or modify the styles of this component, then make s
 | `.pf-c-check__input` | `<input type="checkbox">` |  Initiates a check input. **Required**  |
 | `.pf-c-check__label` | `<label>`, `<span>` |  Initiates a label. **Required**  |
 | `.pf-c-check__description` | `<div>` |  Initiates a check description. |
-| `.pf-m-disabled` | `.pf-c-check__label` |  Initiates a disabled style for labels. **Required when input is disabled** |
+| `.pf-m-standalone` | `.pf-c-radio` |  Modifies the check component for use with a standalone `<input type="checkbox">`. **Required when there is no label** |
```
mattnolting, line 88:
```suggestion
| `.pf-m-standalone` | `.pf-c-check` |  Modifies the check component for use with a standalone `<input type="checkbox">`. **Required when there is no label** |
```

mcoker:
@mattnolting updated

mattnolting (approved):
Love it, 🚢  it!

PR #3822, opened 2021-01-30
fix(wizard): set in page wizard page section to shrink so footer/nav are sticky

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3255

@mcarrano this adds a top box-shadow to the wizard footer when the wizard is 1) used in a modal, 2) used in the special page component section currently needed to add the wizard to the page. Those are the places where we should be able to guarantee the wizard footer remains sticky, so a shadow seems appropriate. I didn't add it globally to any wizard footer since, if the footer doesn't remain sticky as it should, the shadow will look out of place. So this means we may end up with some wizard footers that have a shadow, and ones that don't, even if the wizard footer is sticky since the wizard wasn't used in a modal or the special page section. That seems like a fair compromise since the wizard footers don't have a shadow currently, so nothing changes there. WDYT about that? Should we just wait and add a shadow to the footer in a breaking change to avoid a potential inconsistency with some footers having a shadow, and some not?

### DIFF
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -205,7 +205,7 @@ This component provides the basic chrome for a page, including sidebar, header, 
 | `.pf-c-page__main-nav` | `<section>` |   Creates a container to nest the navigation component in the main page area. |
 | `.pf-c-page__main-breadcrumb` | `<section>` |   Creates a container to nest the breadcrumb component in the main page area. |
 | `.pf-c-page__main-section` | `<section>` |  Creates a section container in the main page area. **Note: The last/only `.pf-c-page__main-section` element will grow to fill the availble vertical space. You can change this behavior using `.pf-m-fill` and `.pf-m-no-fill`, which are documented below.**  |
-| `.pf-c-page__main-wizard` | `<section>` | Creates a container to nest the wizard component in the mian page area. |
+| `.pf-c-page__main-wizard` | `<section>` | Creates a container to nest the wizard component in the main page area. |
 | `.pf-c-page__main-body` | `<div>` | Creates the body section for a page section. **Required when using `.pf-m-limit-width` on `.pf-c-page__main-section`** |
 | `.pf-c-page__main-group` | `<div>` | Creates the group of `.pf-c-page__main-*` sections. Can be used in combination with `.pf-m-sticky-[top/bottom]` to make multiple sections sticky. |
 | `.pf-c-page__drawer` | `<div>` |  Creates a container for the drawer component when placing the main page element in the drawer body. |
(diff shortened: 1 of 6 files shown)


DaoDaoNoCode (approved):
Thanks @mcoker , codes LPTM!

And I agree that we should only add box shadow on wizard in pages and modals. For basic examples, adding a box shadow on footer seems a little bit weird for me when I apply `var(--pf-global--BoxShadow--sm-top)` to it.

Screen Shot 2021-02-01 at 5 18 51 PM

There is a small shadow on the right side. When I apply `var(--pf-global--BoxShadow--md-top)` to it, the shadow on the right disappears, while the top shadow seems too large.

Screen Shot 2021-02-01 at 5 24 11 PM

In my opinion, currently adding box shadow only to the places where we need may be the best choice to avoid breaking changes.

mcarrano:
@mcoker @DaoDaoNoCode I have a couple of questions about this. I thought that the original goal was to make the left-hand steps panel as well as the footer sticky so only the body scrolls. From the title of this PR, looks like it only addresses the footer? From the examples I can't really tell what happens to the siderbar.

Also, I basically agree with you regarding the application of the shadow. However it's visually odd that the footer has a shadow but the left panel does not. Should they have the same effect? I don't know. @mceledonia I'd like to get your opinion here, also.

mcoker:
@mcarrano updated the description - the problem was that the in page wizard was growing too tall, when it needs to shrink to conform with the main page area height. I didn't make any adjustments to the nav - just fixed the wizard's page section so that it shrinks to only fill the available space.

I also updated the wizard examples/demos with a basic form to fill the content area and allow us to text the overflow/scrolling behavior. Does that look OK or would you prefer leave those examples empty?

mcarrano (commented):
Thanks @mcoker . Adding the form inside the wizard body makes it much easier to see whaty is going on. But I still have the same question for @mceledonia about the box shadow.

mattnolting (commented):
Looks great, one question. Should we set `min-height` on `.pf-c-wizard__inner-wrap` for this scenario?

Screen Shot 2021-02-09 at 10 41 50 AM

Or would nobody ever use a browser this way, which is entirely possible..

mcoker:
> Should we set min-height on .pf-c-wizard__inner-wrap for this scenario?

I wondered the same thing... it seems like a safe add, though I also wonder if we should just leave it to the user to make sure there is enough room for the wizard to show the content. @mcarrano wdyt?

mcarrano:
@mcoker @mattnolting I think setting a `min-height` makes sense. So what would happen in the scenario about then?

mcoker:
@mcarrano the footer actions would be pushed down and not sticky if the viewport height was less than the min-height. Once the viewport height was greater than the min-height, the footer would become sticky again.

mcarrano:
Makes sense to me @mcoker .

mcoker:
@mattnolting updated the `min-height`

mattnolting (approved):
Beautiful, LPTM 🎉

Issue #3825, opened 2021-02-01
.pf-c-modal-box__header contents can overlay .pf-c-modal-box__body contents in Safari


rhamilto:
This is because `.pf-c-modal-box__header` is allowed to shrink.  Adding `flex-shrink: 0` to `.pf-c-modal-box__header` resolves the bug.

Bug without fix:
<img width="771" alt="Screen Shot 2021-02-01 at 8 51 19 AM" src="https://user-images.githubusercontent.com/895728/106471335-8296cb00-646f-11eb-90dd-e3a24c0c18f1.png">

Bug with fix:
<img width="772" alt="Screen Shot 2021-02-01 at 8 51 31 AM" src="https://user-images.githubusercontent.com/895728/106471363-89254280-646f-11eb-90ff-1d92657d2b4e.png">

[First reported in OpenShift console.](https://bugzilla.redhat.com/show_bug.cgi?id=1922015)

PR #3826, opened 2021-02-01
fix(modal-box): updated modal-box__header to not shrink

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3825


mattnolting (approved):
LPTM!

PR #3835, opened 2021-02-02
feat(table): added hoverable, selected rows

### LINKED ISSUE #2371: Add selectable row variant to DataTable
Want to support selectable rows in a DataTable separate from the checkbox select such that clicking on the body of a row will select it. This should be implemented as a new variant that can be consumed. Default appearance and behavior remains unchanged. It must be possible to enable row selection independent of checkbox selection and row expansion (i.e. it can exist with or without these separate features.

The selectable hover styles shouldn't be tied to only selectable rows - we should call it "hoverable" or something generic that can apply to other use cases where the row is interactive. From the conversation in https://github.com/patternfly/patternfly/issues/3273

See designs below:

Mock Ups: https://marvelapp.com/69dc1eg/screen/62345709
Mock up video

Background doc for context: https://docs.google.com/document/d/1gt561Ez-8j4-9WidAlQ2ZR12pvrvGD_i_S-eL4EFZdY/edit#heading=h.h7ibgfekw7rm

cc @LHinson @rachael-phillips @mcoker

mattnolting:
closes #2371


mcarrano (commented):
I think this is right @mattnolting . 

@mceledonia and @lboehling it would also be good if you can take a look. One think I wasn't sure about was whether the selected row should also have a hover effect.

mcoker:
* Do we want the accent colors to match the data-list? The data list shows both selectable and expandable rows as having a dark blue accent, but if a selectable row (whether it is selected or not) is expanded, it has the light blue accent.

The data list for reference:
Screen Shot 2021-02-09 at 5 52 24 PM

Screen Shot 2021-02-09 at 5 51 44 PM

* Should this bottom shadow exist on the parent row, between the parent and expanded content? It exists in the design, but the design also has a cursor over the parent row indicating it's being hovered. Even still, I don't think the parent row should have a shadow... only the parent/expanded block would have a shadow. Or is this an older design from when the parent row had a drop shadow between it and the expanded content?

Screen Shot 2021-02-09 at 5 53 27 PM


I'm not sure about this... it seems like we would want to show the shadow around the row you're hovering, even if adjacent rows are selected, since the shadow indicates (to me anyways) the interactive thing. For example, in this case if you clicked the row, you would unselect _that specific row_, but the shadow indicates clicking would unselect both rows.

Screen Shot 2021-02-09 at 5 57 48 PM

I opened https://github.com/patternfly/patternfly/issues/3847 to address a couple of bugs I found in the data-list. We can update it with anything else we do in this table implementation that we want to update the data list to match.

src/patternfly/components/Table/table.scss
```diff
@@ -464,6 +464,149 @@
       border-left-width: 0;
     }
   }
+
+  // tr hoverable
+  --pf-c-table--tr--m-hoverable--BoxShadow: none;
```
mcoker, line 469: Can you move these vars to the top?

src/patternfly/components/Table/table.scss
```diff
@@ -464,6 +464,149 @@
...
+    }
+
+    & + tr.pf-m-selected,
+    & + tr.pf-m-selected:hover {
+      box-shadow: var(--pf-c-table--tr--m-selected--m-selected--BoxShadow);
+    }
+  }
+
+  // tbody hoverable
+  --pf-c-table--tbody--m-hoverable--BoxShadow: none;
```
mcoker, line 537: Same here - can you move them to the top?

src/patternfly/components/Table/table.scss
```diff
@@ -464,6 +464,149 @@
...
+      --pf-c-table--tbody--m-hoverable--BackgroundColor: var(--pf-c-table--tbody--m-hoverable--focus--BackgroundColor);
+    }
+
+    &:active {
+      --pf-c-table--tbody--m-hoverable--BoxShadow: var(--pf-c-table--tbody--m-hoverable--active--BoxShadow);
+      --pf-c-table--tbody--m-hoverable--BackgroundColor: var(--pf-c-table--tbody--m-hoverable--active--BackgroundColor);
+    }
+  }
+
+  --pf-c-table--tbody--m-expanded--m-selected--BorderColor: var(--pf-global--active-color--100);
```
mcoker, line 584: Same here - move to the top?

mcoker (commented):
This looks really good! I have a few questions about the design and parity with the data list.

mcoker:
Looks like the mobile state needs some work, although we can follow up in the next sprint if needed. Looks like expandable still works fine.

Screen Shot 2021-02-10 at 1 01 32 PM

src/patternfly/components/Table/table.scss
```diff
@@ -464,6 +511,125 @@
       border-left-width: 0;
     }
   }
+
+  // tr hoverable
+  tr.pf-m-hoverable {
+    cursor: pointer;
+    background-color: var(--pf-c-table--tr--m-hoverable--BackgroundColor);
```
mcoker, line 518: could this just be white? with a comment to make it the default for `tr` in a breaking change? Or is there a reason we only set it on hover/focus/active?

mcoker (approved):
Looks good! Created a follow up issue to address some style bugs

https://github.com/patternfly/patternfly/issues/3852

Issue #3627, opened 2020-10-28
Adding Toggle button component variation


mcarrano:

mcarrano:
Introduce Toggle button as a new component or as a new variant of Button per the designs here: https://docs.google.com/document/d/1GyVgog_SbErqbYyF08_N2rfLDUgwAVbdKP5udBjdxYI/edit?usp=sharing For this initial implementation, include only variants listed as part of the "Simple toggle." This covers the most common dropdown and select list use cases.

PR #3845, opened 2021-02-09
feat(menu-toggle): added menu toggle component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3627

Didn't create the clear button example because it's unclear what that toggle is. That implies the outer element is something other than a `<button>`, since the clear button is also a `<button>`. Is that example a typeahead? If I can get clarification on that, I can try and add it, or we can follow up and add it as a later enhancement.


src/patternfly/components/Toggle/toggle.hbs
```diff
@@ -0,0 +1,25 @@
...
+  {{else}}
+    aria-expanded="false"
+  {{/if}}
+  {{~#if toggle--IsDisabled}}
+    disabled
+  {{/if}}
+  {{#if toggle--attribute}}
+    {{{toggle--attribute}}}
+  {{/if}}>
+  {{>@partial-block}}
```
mattnolting, line 24:
```suggestion
  {{> @partial-block}}
```

src/patternfly/components/Toggle/toggle.scss
```diff
@@ -0,0 +1,181 @@
...
+
+  // border
+  --pf-c-toggle--after--BorderTopWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--after--BorderRightWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--after--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--after--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--after--BorderTopColor: var(--pf-global--BorderColor--300);
+  --pf-c-toggle--after--BorderRightColor: var(--pf-global--BorderColor--300);
+  --pf-c-toggle--after--BorderBottomColor: var(--pf-global--BorderColor--200);
+  --pf-c-toggle--after--BorderLeftColor: var(--pf-global--BorderColor--300);
```
mattnolting, line 19: Can you split the gray border and the accent border with `::before` (being the gray border) and `::after` (being the accent border) like we've done in other places, table, tabs, etc, to avoid a bevel if user updates `border-width`?

Screen Shot 2021-02-09 at 10 12 19 AM

mattnolting (commented):
Looking good, just a few comments 👍

mattnolting:
This works really well in `.pf-c-dropdown` in conjunction w/`.pf-c-menu` as well, nice job 👍

mcarrano (commented):
This looks like a good first step @mcoker . I agree with keeping the version with the clear button out of the mix for now. I'm also wondering if we should call this something other than "Toggle" since we already have a toggle button group. Maybe it's a "Dropdown button?"

@tlabaj @jschuler @kmcfaul can you also take a look at this? It's directly related to this React issue and the result of a convo we had about composable dropdown menus last October - https://github.com/patternfly/patternfly-react/issues/5073 Will this support that composable dropdown approach?

mcoker:
> I'm also wondering if we should call this something other than "Toggle" since we already have a toggle button group. Maybe it's a "Dropdown button?"

+1! Toggle is confusing with the toggle group as a separate thing. I think dropdown button could work, though if we stick with separate components for dropdown and select, it could be confusing to use a dropdown toggle with a select menu. Maybe "Menu toggle"?

src/patternfly/components/Toggle/toggle.scss
mcoker: Updated

mcoker:
@mattnolting updated the base/accent borders to be separate.

@mcarrano wdyt about the naming?

mcarrano:
I like calling it a "menu toggle" as that will keep it grouped with the Menu component in the nav.

src/patternfly/components/Toggle/toggle.scss
```diff
@@ -0,0 +1,194 @@
...
+
+  // border base
+  --pf-c-toggle--before--BorderTopWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--before--BorderRightWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--before--BorderBottomWidth: 0;
+  --pf-c-toggle--before--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-toggle--before--BorderTopColor: var(--pf-global--BorderColor--300);
+  --pf-c-toggle--before--BorderRightColor: var(--pf-global--BorderColor--300);
+  --pf-c-toggle--before--BorderBottomColor: transparent;
+  --pf-c-toggle--before--BorderLeftColor: var(--pf-global--BorderColor--300);
```
mattnolting, line 19: I think this could just be `--pf-c-toggle--before--BorderWidth` and `--pf-c-toggle--before--BorderColor` as we assume that the `::before` is a simple border and the `::after` is fancier.

mattnolting (commented):
Another border comment, but aside from that it LGTM. What do you think about `Menu control`? Also, I think "Menu toggle" works well.

src/patternfly/components/Toggle/toggle.scss
mcoker: It could... my only hesitation is with breaking changes where, if we change one of the border colors down the road and need to separate the vars (or just add a new var to define one of the sides), since `--pf-c-toggle--before--BorderColor` would no longer control all border colors, it could be considered breaking. Similar to how we separate hover/focus/active into separate vars, even if they're all the same. Using a common var sets us up for the role a var plays in a component to define something as no longer defining that thing. Though I don't feel strongly about it, we can't avoid some of those situations, though writing vars separately like this is one way to try and avoid that scenario.

mattnolting (approved):
L🌮 TM!

PR #3846, opened 2021-02-09
feat(table): added tree view support

### LINKED ISSUE #3809: Introduce a Tree Table component
Introduce a new component that allows tabular data to be displayed in a hierarchical tree according to the following requirements:

* Rows can be nested n levels deep (no enforced limitation).
* Rows can be either folder nodes or leaf nodes, where folder nodes contain nested children that may be either folder nodes or leaf nodes. Either folder nodes or leaf nodes may have additional attributes or actions displayed in columns with the constraint that all items have the same (or a subset of) the same set of attributes.
* Folder nodes may optionally have a leading icon.
* Checkbox selection is supported for selecting one or more nodes. Selection rules should work as for the Tree View component.
* Filtering is supported to limit the number of items displayed and will work as for a Tree View. The user may filter on any attribute/column.
* Column sorting is not required.

Designs: https://marvelapp.com/prototype/fj23h11

mattnolting:
closes #3809 

The a11y issues don't align w/https://www.w3.org/TR/wai-aria-practices/examples/treegrid/treegrid-1.html specification. If you run axe on the w3 example, you get the same error. 

Preview: https://patternfly-pr-3846.surge.sh/components/table/#tree-table

Note: the position of `toggle` and `check` were swapped to be consistent with other components, `Treeview` `Expandable Table` etc. and to resolve odd spacing in the absence of `toggle`.


mcarrano:
@mattnolting This is looking good. I just had two issues/questions:

1- What is driving the width of the first column? I seems like it's much wider than needed for the content and causes columns 2, 3, and 4 to be really squished.

2- I'm not sure that the alignment of the Level 2 Leaf checkboxes is correct. Shouldn't they be indented?

@nicolethoen perhaps you should also take a look to ensure this will support the React implementation.

DaoDaoNoCode:
@mcarrano Hi, I have a small question about the design. In this tree view table design, the checkbox is on the leftmost in a row. However, we can see other examples like [TreeView](https://www.patternfly.org/v4/components/tree-view/html#with-checkboxes) or [Compact Expandable Table](https://www.patternfly.org/v4/components/table/html#compact-expandable), where checkbox is on the right of the expand icon. Should we make this consistent in this example? If not, the space between checkbox and title in child nodes looks a little weird as you mentioned above.
Screen Shot 2021-02-09 at 2 04 15 PM

mcarrano (approved):
This looks great @mattnolting !

mattnolting:
@DaoDaoNoCode updated the position of `toggle` and `check`

src/patternfly/components/Table/examples/Table.md
```diff
@@ -1416,6 +1416,44 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
...
+```
+
+### Tree table with checkboxes
+```hbs
+{{> table-tree-view--basic table--id="tree-table-with-checkboxes" table--gridsss="true" tree-view--HasCheckboxes="true" table--attribute='aria-label="This is a simple tree table, with checkboxes example"'}}
+```
+
+### Tree table with checkboxes, icons here
+```hbs
+{{> table-tree-view--basic table--id="tree-table-with-checkboxes-icons" table--gridsss="true" tree-view--HasCheckboxes="true" table-tree-view--HasIcons="true" table--attribute='aria-label="This is a simple tree table, with checkboxes example"'}}
```
DaoDaoNoCode, line 1431:
```suggestion
{{> table-tree-view--basic table--id="tree-table-with-checkboxes-icons" table--gridsss="true" tree-view--HasCheckboxes="true" table-tree-view--HasIcons="true" table--attribute='aria-label="This is a simple tree table, with checkboxes and icons example"'}}
```

DaoDaoNoCode (approved):
Looks great, thanks @mattnolting ! Only a small problem in the doc. Could you update the `aria-label`? Except this, the other parts are perfect!

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  // check
+  --pf-c-table__tree-view-main--c-table__check--MarginRight: var(--pf-global--spacer--md);
+
+  // icon
+  --pf-c-table__tree-view-icon--MinWidth: var(--pf-global--FontSize--md);
+  --pf-c-table__tree-view-icon--MarginRight: var(--pf-global--spacer--sm);
+
+  // toggle
+  --pf-c-table--m-tree-view__toggle--Width: #{pf-size-prem(42px)}; // based off of the expected width of the toggle button
```
mcoker, line 17: The react button will be different. The core fa icon is 10px, so 16px (button padding) * 2 + 10px is 42px. The react icon will be 48px since the icon is 16px wide. Should we do this the way we do it in tree view and dual list selector, where nested tree view items have consistent padding, and the toggle button is absolutely positioned (to take it out of the layout) and `translateY()`'d to the left 100%? That will maintain aligned nested content, and allow for minor variation in toggle width.

mcoker:
Compared to tree view, the nested expand  icons don't align with the start of the parent content.
Screen Shot 2021-02-10 at 3 47 10 PM

Here's tree view for comparison

Screen Shot 2021-02-10 at 3 49 25 PM

mcoker (approved):
LGTM! A couple of follow ups to look into next sprint - https://github.com/patternfly/patternfly/issues/3853

PR #3860, opened 2021-02-12
fix(drawer): fixed bottom variant issues in safari

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3857

`flex-basis: 100%` on `__content` isn't working in a vertical layout because its parent `__main` doesn't have a height.


evwilkin (approved):
LGTM - thanks! 🎉

evwilkin:
Looks like there are a11y issues related to tree table

mattnolting (approved):
Perfecto 🌮

Issue #3861, opened 2021-02-12
Tree table is causing the a11y test to fail


mcoker:
We added some `aria` attrs to support tree view (following the [w3 tree grid example](https://www.w3.org/TR/wai-aria-practices/examples/treegrid/treegrid-1.html)), and the a11y tests are failing. They fail on the w3 example, too.

PR #3862, opened 2021-02-12
fix(build): set a11y test to temporarily ignore aria-allowed-attr

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3861

just temporarily ignoring this attr and we can enable next sprint once we sort out why it's failing.


mattnolting (approved):
🚢  it!

PR #3869, opened 2021-02-17
fix(release-notes): increment new core version number correctly

nicolethoen:


redallen (approved):
Take it easy while you spend the next hour bumping things :)

Issue #3876, opened 2021-02-18
Login show/hide password example - icons missing aria-hidden


mcoker:
The `fa-eye` and `fa-eye-slash` icons on the login examples page are missing `aria-hidden="true"`

https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/Login/examples/Login.md

PR #3877, opened 2021-02-18
chore(login): added aria-hidden to hide/show password example fa icons

mcoker:
Small bug from last sprint. Fixes https://github.com/patternfly/patternfly/issues/3876


evwilkin (approved):
LGTM

PR #3878, opened 2021-02-22
fix(options-menu, dropdown): normalized plain toggle size

mcoker:
This makes the plain toggles match the plain button size.

fixes https://github.com/patternfly/patternfly/issues/3437

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -6,7 +6,7 @@ $pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "
   --pf-c-dropdown__toggle--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-dropdown__toggle--PaddingBottom: var(--pf-global--spacer--form-element);
   --pf-c-dropdown__toggle--PaddingLeft: var(--pf-global--spacer--sm);
-  --pf-c-dropdown__toggle--MinWidth: var(--pf-global--target-size--MinWidth);
+  --pf-c-dropdown__toggle--MinWidth: 0;
   --pf-c-dropdown__toggle--FontSize: var(--pf-global--FontSize--md);
   --pf-c-dropdown__toggle--FontWeight: var(--pf-global--FontWeight--normal);
   --pf-c-dropdown__toggle--Color: var(--pf-global--Color--100);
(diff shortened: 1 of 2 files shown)


evwilkin (approved):
LGTM

mattnolting (commented):
Ah, the `toolbar`!

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -6,7 +6,7 @@ $pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "
   --pf-c-dropdown__toggle--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-dropdown__toggle--PaddingBottom: var(--pf-global--spacer--form-element);
   --pf-c-dropdown__toggle--PaddingLeft: var(--pf-global--spacer--sm);
-  --pf-c-dropdown__toggle--MinWidth: var(--pf-global--target-size--MinWidth);
+  --pf-c-dropdown__toggle--MinWidth: 0;
```
mattnolting, line 9: Looks like you need to keep the `var(--pf-global--target-size--MinWidth);` value as the button is the correct height now, but narrower than before. 

Screen Shot 2021-03-02 at 8 43 56 AM

The `toolbar` spacing issue seems to be a result of a misconfiguration in the `toolbar` demos. Right now, these buttons are part of an `overflow menu` whereas they should just be in an `toolbar icon group`. 

Screen Shot 2021-03-02 at 8 54 24 AM

To correct, we could A) move the overflow menu with the kebab into the `icon button group`, B) negative margin an `overflow menu` that appears next to an `icon button group` or C) use the built in margining to space the overflow menu. I think I'd most support option B, as it seems to be the most flexible solution. Because the icon buttons `pf-m-plain` their padding provides the expected visual spacing, but that may cause and issue when they're a sibling to any other `button` when it is exposed from an overflow menu.

mcoker: > Looks like you need to keep the var(--pf-global--target-size--MinWidth); value as the button is the correct height now, but narrower than before.

I think that's OK, this update basically just updates the plain toggles to match the plain button sizes, so if you have a plain button and a plain dropdown toggle, they'll be the same size in react anyways since our FA icons are a variable width. To make them the same in core, I think all of the plain actions (buttons/toggles) would need some sort of min-width - even then you'd have some variation since some core icons are wider than they are tall.

mattnolting (approved):
WFM, LAPTM! 🌮

PR #3879, opened 2021-02-22
feat(slider): added support for disabled

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3864

### DIFF
--- a/src/patternfly/components/Slider/examples/Slider.md
+++ b/src/patternfly/components/Slider/examples/Slider.md
@@ -235,12 +235,69 @@ cssPrefix: pf-c-slider
 {{/slider}}
 ```
 
+### Disabled
+```hbs
+{{#> slider
+  slider--IsDisabled="true"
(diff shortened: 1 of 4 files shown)


mcarrano (approved):
Looks good to me. @maryshak1996 @mceledonia what do you think?

mcarrano:
@KKoukiou just wanted to let you know that a core PR is up. Once this is merged I think it will be ready to implement in the React component.

maryshak1996 (approved):

mattnolting (approved):
This looks great! My only question is, should the `disabled` cursor  be `not-allowed`? @mcarrano @maryshak1996?

maryshak1996:
@mattnolting that's a really good point -- I do think that having the cursor change would make it clearer that a slider is intentionally not interactive, rather than just 'not working'

mcarrano:
I like it. Thanks for adding the cursor disabled feature!

Issue #3880, opened 2021-02-23
Spacing between switch and label seems too large


mfrances17:
Received comments during a keycloak end-of-sprint demo, some from UXD design team members, that the space between the switch and its label seemed too large and to verify that it was following Patternfly standards. We verified that we are using the standard space and have not change it. Before changing the CSS of the component on our end, figured it made sense to see if it should be changed directly in PF since it will impact consistency among products.

Keycloak issue that I filed: 
https://github.com/keycloak/keycloak-admin-ui/issues/383

Existing keycloak layout:
![image](https://user-images.githubusercontent.com/39063664/108860878-a2c43080-75bc-11eb-81c4-52bb298ff6ef.png)

Patternfly example:
![image](https://user-images.githubusercontent.com/39063664/108861448-3ac21a00-75bd-11eb-8ca3-d7db87955b5e.png)

As you can see in the above example, it becomes even more apparent when the switch is placed horizontally with other components. Some of the comments were that the 'Enabled' label looked like it was its own component, not tied to the switch at all, that it appeared 'floating' in between the two other components, etc.

mcoker:
cc @mcarrano @mceledonia

mceledonia:
It is quite large, and seems off center as well. I think we could drop this down 1 spacer (e.g. from lg to md) and also see if there's any way we can get the text aligned with the center of the switch.

mcarrano:
I'd be in favor of updating this. @mcoker do you see the risk of a visual breaking change if we reduce the spacing?

mcoker:
@mcarrano nope, seems fine to update. It actually looks like a bug - there are 2 medium spacers creating that space, where it looks like there should only be 1.

mcarrano:
Sounds good. Hopefully a quick fix?

PR #3882, opened 2021-02-23
fix(switch): adjust space between switch and label

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3880

Looks like this was a bug introduced in https://github.com/patternfly/patternfly/pull/3749/


evwilkin (approved):
LGTM!

mcarrano (approved):
Definitely looks better! What do you think @mceledonia ?

mattnolting (approved):
LGTM 👍

Issue #3837, opened 2021-02-03
Textarea Autosize/Autoexpand


andrewballantyne:
Would like to request that we could get the Textarea component (https://www.patternfly.org/v4/components/text-area) to get an auto-size ability when the user hits enter, it expands to contain all the text. Ideally we'd want it to shrink again if the user removes lines.

This is an ask for React PF, but was suggested to log here first and see where the logic needs to go first. 

cc @mcarrano 

Thoughts on design:

* starts at the default size
    * would be useful if this was configurable (ie `rows` or `min-height`?) 
* when the user reaches the end of the height of the area, the next 'enter' grows the box to support the new line
* when the user removes a line, it shrinks back down (until it hits the "default" size)

mcarrano:
@mcoker @tlabaj can you take a look at this request? It came up in the OpenShift meeting today. I wasn't sure what would be required to enable this behavior since I know that we are currently using a browser native control for the text area.

cc @evwilkin

mcoker:
From the core side, I think we would just want to add variable support for manipulating the textarea height since there are no vars in place currently. That would support defining initial size/min-height (and optionally a max-height), and a var the react component can set inline that manipulates the textarea element height.

mcarrano:
Thanks @mcoker . I will get this on the PF roadmap then, starting in Core. Are you needing this in any specific timeframe @andrewballantyne ?

andrewballantyne:
@mcarrano 4.8 of OCP. April 9th is Feature Freeze, so ideally before that.

mcarrano:
Thanks @andrewballantyne . I will try to get this into the queue for our 2020.04 (end of March) release.

andrewballantyne:
Is this going to make its way as a configurable Prop in the React component? @mcoker / @mcarrano

Issue #2251, opened 2019-09-11
Textarea: Add modifiers for resize properties


christiemolloy:
This issue is related to an issue in React: https://github.com/patternfly/patternfly-react/issues/2839

There are two parts:
1) Make a modifier for the text area component that sets `resize` to just be on the vertical axis. Also add a modifier for the horizontal axis. We could also have a modifier for resize none, but that would be up to design.

2) Currently in Core, we are able to horizontally resize the TextArea component outside of its parent container. Resizing should only be limited to within its parent constraints. See screenshot here:

<img width="832" alt="Screen Shot 2019-09-11 at 10 48 59 AM" src="https://user-images.githubusercontent.com/20118816/64708110-11253980-d482-11e9-872e-069af57b2f3b.png">

rachael-phillips:
We story-pointed on the enhancement - step 1 of this issue

PR #3883, opened 2021-02-23
fix(form-control): added vars for textarea width/height for resizing

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3837

Allows textarea to be resized via `--pf-c-form-control--textarea--Width` and `--pf-c-form-control--textarea--Height`

@tlabaj can you let me know if this is enough to support autosize/resize?


mattnolting (commented):
Looks great, can we add a `max-width: 100%` as well? When resizing, the fluidity of width makes the textarea feel unmaintained... 

Screen Shot 2021-02-25 at 8 50 39 AM

mcoker:
@mattnolting when we initially introduced the resize modifiers in https://github.com/patternfly/patternfly/issues/2251#issuecomment-540640274, we decided not to impose any limitations on the ability to resize the width/height with the default textarea. It could also be considered a breaking change to do that at this point.

mattnolting (approved):
LGTM 👍

tlabaj:
@mcoker yes, this should be sufficient to autosize the textarea based on the scrollWidth/scrollHeight .

Issue #3885, opened 2021-02-24
Split does not wrap


manstis:
Should the content of a `Split` exceed the width of the `Split` itself the children overflow vs wrap.

Is this the expected behaviour? Should applications take remedial steps?

We are tracking work here: https://issues.redhat.com/browse/FAI-423

![FAI-423](https://user-images.githubusercontent.com/497544/109032815-ee123800-76bd-11eb-85a6-7eb92112f214.png)

`@patternfly/patternfly: 4.70.2`
`@patternfly/react-core: 4.84.4`

mcoker:
@manstis the flex and level layouts are set to wrap by default, so it makes sense to me that this layout should wrap by default, too. I have a PR adding a `pf-m-wrap` variation to the layout, and a note proposing to update the default behavior in our next breaking change release. https://github.com/patternfly/patternfly/pull/3887

Until this feature is supported, you could use the flex layout, which supports wrapping and controlling the layout/flex properties at different breakpoints.

manstis:
Thank-you @mcoker. For the time being I can add `flex-wrap: wrap` to my use of `Split`.

manstis:
Thank-you @mcoker

PR #3887, opened 2021-02-24
feat(split): added variation to allow children to wrap

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3885


mattnolting (approved):
Perfect 🎉

Issue #3865, opened 2020-05-04
Accordion for web use cases


ajacobs21e:
Proposed variant for the accordion for a larger marketing / web use case. 
![Accordion specs](https://user-images.githubusercontent.com/6415742/81011743-75887500-8e26-11ea-9b7f-159483a74d56.png)

Text sizes, icon size, and padding are increased. 
Added divider lines between each toggle. 
Optional CTA shown. 

Another change is we are not adding the font weight increase and change to link blue color for selected or hover states. With the larger text size the change to medium weight font is overwhelming. We also feel the blue highlight is sufficient to show selected state, like for tabs or data list rows. 

Accordion demo: https://www.sketch.com/s/f6a52c25-8657-4680-9d3e-4c09131e6ed0/p/accordion
Accordion components: https://www.sketch.com/s/f377ffc2-c43b-4ba6-8504-6396047f96db/p/accordion

cc @coreyvickery

PR #3888, opened 2021-02-25
feat(accordion): added display-lg, support multiple bodies

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3865

* Adds web/marketing variation
* Moves the pseudo elements around a little to support the border overlap
* Adds support for multiple `__expanded-content-body` elements

@ajacobs21e @mcarrano can you also verify the spacing between adjacent sections is correct in the regular/fixed example? https://patternfly-pr-3888.surge.sh/components/accordion#fixed

### DIFF
--- a/src/patternfly/components/Accordion/accordion-expanded-content.hbs
+++ b/src/patternfly/components/Accordion/accordion-expanded-content.hbs
@@ -3,7 +3,5 @@
   {{#if accordion-expanded-content--attribute}}
     {{{accordion-expanded-content--attribute}}}
   {{/if}}>
-  {{#> accordion-expanded-content-body}}
-    {{> @partial-block}}
-  {{/accordion-expanded-content-body}}
+  {{> @partial-block}}
 </{{#if accordion--IsDefinitionList}}dd{{else}}div{{/if}}>
--- a/src/patternfly/components/Accordion/accordion-toggle.hbs
+++ b/src/patternfly/components/Accordion/accordion-toggle.hbs
@@ -1,4 +1,10 @@
 <{{#if accordion--IsDefinitionList}}dt>{{else}}h3>{{/if}}<button class="pf-c-accordion__toggle{{#if accordion-toggle--IsExpanded}} pf-m-expanded{{/if}}{{#if accordion-toggle--modifier}} {{accordion-toggle--modifier}}{{/if}}"
+  type="button"
+  {{#if accordion-toggle--IsExpanded}}
+   aria-expanded="true"
+  {{else}}
+   aria-expanded="false"
+ {{/if}}
  {{#if accordion-toggle--attribute}}
    {{{accordion-toggle--attribute}}}
  {{/if}}>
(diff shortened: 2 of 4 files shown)


mcarrano (commented):
Looks good to me @mcoker . I'll defer to @ajacobs21e regarding the spacing and font sizes for the new Large variant. Not sure how to answer your question regarding the "fixed" variant. Looks OK to me but perhaps @mceledonia should also take a look.

mceledonia:
Looks alright to me @mcarrano @ajacobs21e !

ajacobs21e:
LGTM! https://www.sketch.com/s/00432cc6-fd0c-454f-bdd6-c9720ceebbdc/a/OvQza4

mcoker:
> Not sure how to answer your question regarding the "fixed" variant.

Currently the accordion component only accepts a single body element per accordion item. So all of the content in the accordion item goes in a single accordion item body. This PR introduces the ability to pass multiple body elements to an accordion item, so just wanted to verify the 8px spacer between body items in the regular accordion (not the large variation introduced here) looks OK.

mcarrano:
Ah, I see @mcoker . I think that the 8px spacer looks fine in this example. So would the spacing always be 8px or could I define a larger spacing if I wanted. Is the 8px consistent with the spacing that we use on a standard form?

mcoker:
> would the spacing always be 8px or could I define a larger spacing if I wanted

The spacing between accordion body elements would always be 8px (or whatever space we define here). If you wanted different spacing, you could always use a stack or flex or other layout in a single accordion body and define your own spacing. 

> Is the 8px consistent with the spacing that we use on a standard form?

The spacer between form rows is 16px. I used an 8px spacer here since the spacer in the large/website variation is 16px, but we could use that as the non-large spacer, too.

current/8px spacer
Screen Shot 2021-03-01 at 2 22 12 PM


16px spacer
Screen Shot 2021-03-01 at 2 22 59 PM

mattnolting (commented):
Looks great, can we separate `display-lg` and `pf-m-border`?

mcoker:
> Looks great, can we separate display-lg and pf-m-border?

@ajacobs21e @mcarrano wdyt? Should the larger presentation and the borders between the items be separate variations? That implies that you could use the bordered variation with a regular accordion, and could use the large/web variation without a border - is that what you were thinking @mattnolting?

mattnolting:
@mcoker yessir, that's exactly it!

mcarrano:
@mcoker @mattnolting that makes sense to me. @mceledonia can you see any reason we shouldn't allow borders on the standard/regular accordion headers?

mcoker:
@mcarrano @mattnolting separated bordered & large variations, added a regular bordered example and updated the large example to include the bordered variation.

mcarrano (approved):
Looks great @mcoker . I actually like this much better with the borders on both regular and large variants.

@doruskova let's make sure to add this option to the Sketch kit, also. Can you open an issue? Thanks.

mattnolting (approved):
Fanstatic! 🚢  it!

mcoker:
>  I actually like this much better with the borders on both regular and large variants.

@mattnolting++

PR #3890, opened 2021-02-26
fix(table): fixed treeview alignment

### LINKED ISSUE #3853: Tree table follow ups
From https://github.com/patternfly/patternfly/pull/3846

* Instead of `--pf-c-table--m-tree-view__toggle--Width: #{pf-size-prem(42px)};`, we should consider matching the tree view component:

https://github.com/patternfly/patternfly/blob/2c25d962c626871e8242a984dcf8cf6ace8ef5ad/src/patternfly/components/TreeView/tree-view.scss#L8

* Let's consider absolutely positioning the nested toggles the way we did in tree view and dual list selector. If a user were using a bigger toggle than we allocated space for in the grid column, the toggle width would grow away from the row content it precedes instead of toward it, potentially overlapping the content. The way the component is setup now, there is no space between the toggle and checkmark, so it would be easy for them to overlap. Or maybe that approach isn't ideal and we should update tree view and dual list selector to match the table?

mattnolting:
closes #3853

### DIFF
--- a/src/patternfly/components/Table/table-tree-view.scss
+++ b/src/patternfly/components/Table/table-tree-view.scss
@@ -1,49 +1,77 @@
 $pf-c-tree-view--MaxDepth: 10;
 
 .pf-c-table {
+  --pf-c-table__tree-view-main--indent--base: calc(var(--pf-global--spacer--md) * 2 + var(--pf-c-table__tree-view-icon--MinWidth)); // based off of the expected width of the toggle button
+  --pf-c-table__tree-view-main--nested-indent--base: calc(var(--pf-c-table__tree-view-main--indent--base) - var(--pf-global--spacer--md)); // nested spacing that removes the toggle button's left padding, so the icon aligns with text on the node above it
+
   // main
-  --pf-c-table__tree-view-main--GridTemplateColumns: var(--pf-c-table--m-tree-view__toggle--Width) max-content 1fr;
-  --pf-c-table__tree-view-main--indent--base: var(--pf-global--spacer--lg);
-  --pf-c-table__tree-view-main--MarginLeft: calc(var(--pf-global--spacer--md) * -1); // based off of the expected width of the toggle button
+  --pf-c-table__tree-view-main--PaddingLeft: 0;
+  --pf-c-table__tree-view-main--MarginLeft: calc(var(--pf-c-table--cell--first-last-child--PaddingLeft) * -1);
 
   // check
-  --pf-c-table__tree-view-main--c-table__check--MarginRight: var(--pf-global--spacer--md);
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -1,49 +1,77 @@
...
 
 .pf-c-table {
+  --pf-c-table__tree-view-main--indent--base: calc(var(--pf-global--spacer--md) * 2 + var(--pf-c-table__tree-view-icon--MinWidth)); // based off of the expected width of the toggle button
+  --pf-c-table__tree-view-main--nested-indent--base: calc(var(--pf-c-table__tree-view-main--indent--base) - var(--pf-global--spacer--md)); // nested spacing that removes the toggle button's left padding, so the icon aligns with text on the node above it
+
   // main
-  --pf-c-table__tree-view-main--GridTemplateColumns: var(--pf-c-table--m-tree-view__toggle--Width) max-content 1fr;
-  --pf-c-table__tree-view-main--indent--base: var(--pf-global--spacer--lg);
-  --pf-c-table__tree-view-main--MarginLeft: calc(var(--pf-global--spacer--md) * -1); // based off of the expected width of the toggle button
+  --pf-c-table__tree-view-main--PaddingLeft: 0;
```
mcoker, line 8: I wonder if this should be `--pf-global--spacer--sm` > xl breakpoint so that the button icon aligns with the column header text

Current > 1200px
Screen Shot 2021-03-01 at 1 51 53 PM

< 1200px aligns since the table and button padding are 16px
Screen Shot 2021-03-01 at 1 52 01 PM

mcoker (commented):
LGTM, just one comment about the icon alignment on desktop.

mattnolting:
@mcoker updated! I used the cell padding var so it's always relative to that value.

mcoker (commented):
LGTM! Can you uncomment the top examples?

mattnolting:
@mcoker updated!

mcarrano (commented):
I guess this is OK, and the reasoning behind it makes sense, although I have a hard time seeing what this fixes from the original version. @mattnolting can you just summarize what this is changing/fixing?

mcoker (approved):
🥳

mcoker:
> can you just summarize what this is changing/fixing?

@mcarrano the tree toggle was a fixed pixel width so that we could use that width to assign the proper amount of indentation for nested rows, and maintain alignment with the toggles and text. This PR just removes that fixed width and uses an estimation of the toggle width based off of the toggle button's padding and icon min-width, which is more flexible than the fixed width and aligns with the way we have managed the toggle width and indentation in the tree view and dual list selector components. Visually you probably won't notice a change.

Issue #3863, opened 2021-02-15
Search input: add auto-complete variant


mcarrano:
Add optional auto-complete to the Search input component.

Designs here: https://docs.google.com/document/d/1mFOuhoAJNBSWKHZKwz8T6OVk71OWNPhfAuVF1Jw3BAk/edit?usp=sharing

mcarrano:
@mcoker I've opened this issue so we can pull it into the 2021.03 milestone. We are still reviewing this design, but expect to finalize this week.

mcoker:
@mcarrano checking to see if this is OK to begin working on? Also I'm assuming the advanced search box can be open at the same time, and the auto-complete box would open on top?

mcarrano:
Yes, that's correct.

PR #3892, opened 2021-03-02
feat(search-input): added autocomplete

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3863

@mcarrano 

* This does not add the hint/grey text that finishes the search term in the text input when there is a single option in the autocomplete box. That part isn't obvious how to implement since the text there is a regular text input and you can't style text like that in a text input. If that feature is a requirement, I'd like to have a follow-up issue to investigate how to best do that.
* We have the option to use the menu component here. I didn't implement it that way, and just made this list a feature of the search input component since we wouldn't be able to promote the search input component until we promoted the menu component. If that's OK, or we anticipate wanting to support features of the menu component in this list (descriptions, favorites, etc), I can update this component to use the menu component instead.

### DIFF
--- a/src/patternfly/components/SearchInput/examples/SearchInput.css
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.css
@@ -1,11 +1,8 @@
-#ws-core-c-search-input-advanced-search-expanded {
+#ws-core-c-search-input-advanced-search-expanded,
+#ws-core-c-search-input-advanced-search-expanded-with-autocomplete {
   height: 550px;
 }
 
 #ws-core-c-search-input-autocomplete {
   height: 250px;
-}
-
-#ws-core-c-search-input-autocomplete-final-option {
-  height: 150px;
 }
--- a/src/patternfly/components/SearchInput/examples/SearchInput.css
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.css
@@ -6,3 +6,7 @@
 #ws-core-c-search-input-autocomplete {
   height: 250px;
 }
+
+#ws-core-c-search-input-autocomplete-last-option-hint {
+  height: 150px;
+}
--- a/src/patternfly/components/SearchInput/examples/SearchInput.css
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.css
@@ -1,3 +1,11 @@
 #ws-core-c-search-input-advanced-search-expanded {
   height: 550px;
 }
+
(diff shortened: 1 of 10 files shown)


mattnolting (approved):
This looks great, although it feels to me like we should use the menu. Could we keep this dedicated menu with the assumption that later we could promote the `menu` and replace?

mcarrano (commented):
This looks good @mcoker . I don't see a reason to pull in the menu component here. This should never be anything more than a single select list. I'm more concerned that we can't complete the auto-complete text styling. Do you have any idea how others are doing this? Maybe needs more research. We could still support the basic functionality without it. Thoughts?

mcoker:
> I don't see a reason to pull in the menu component here. This should never be anything more than a single select list. 

@mcarrano OK that works, though if we end up wanting to support descriptions or other things, it's a breaking change to use the menu component here instead. We would need to align that update with a breaking change release. That's OK?

> Do you have any idea how others are doing this? 

Do you have an example of this being used somewhere? Preferably on a desktop device. I checked amazon, search engines, and a handful of places and I couldn't find an instance of the hint text appearing in a search text input field. I can think of a couple of ways off the top of my head to do it though it isn't a very straight forward approach.

The easiest way might be to layer text inputs on top of one another, so that a text input underneath the active/top one has the full search term in hint styling (grey text) and that text shows through as you're typing in the top text input.

Another way could be to use `contenteditable` with a regular `<div>` that the user types in, which will mimic a text input, and to have react manage showing the hint text after the cursor to complete the word.

mcarrano:
@mcoker yes, I see where this can get really complicated. I did some quick research and most auto-complete components do not support hints like this. But I found two references:

https://www.npmjs.com/package/react-autocomplete-hint

https://reactjsexample.com/a-react-component-for-autocomplete-hint/

See if these give you any ideas. If this becomes too unwieldy, I'd consider pulling the requirement and perhaps just supporting a keyboard shortcut to auto-complete the term when there is only one remaining option.

mcoker:
@mcarrano thanks for that! The component used in those pages is the same, but looks like they layer inputs on top of one another like I mentioned above. I'll give that a shot here.

mcoker:
@mcarrano @mattnolting @jessiehuff added support for hint text.

mcarrano (approved):
Thanks for figuring out how to implement the hint text @mcoker . This looks good to me!

Issue #3896, opened 2021-03-03
Breadcrumb - allow items to be buttons instead of links


mcoker:
Primary use case is to support menu drilldown breadcrumbs. This has already been implemented in react, we just need docs and examples. After making this update, we should update the breadcrumbs in the menu examples. Reference react for the correct usage.

PR #3901, opened 2021-03-03
fix(breadcrumb): added docs, example for items as buttons

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3896


mattnolting (approved):
LPTM! 🥇

PR #3914, opened 2021-03-06
feat(tabs): added tabs demos

### LINKED ISSUE #3700: Tabs demo
Implement a demo of tabs usage per the designs here: https://marvelapp.com/prototype/5fba0cj

mattnolting:
closes #3700 

What the PR does
- Adds tab demos. 
  - Design here: https://marvelapp.com/prototype/5fba0cj/screen/73740947 
  - Previews here: https://patternfly-pr-3914.surge.sh/components/tabs/html-demos
- Adds full height card: https://patternfly-pr-3914.surge.sh/components/card#full-height-card
  - Currently, when a `card` is a child of a `layout__item`, the `card` is not the full height of the parent `__item`. `pf-m-fill` allows the user to set the height of the `card` to 100%.
- ~Updates the `sidebar` component. Currently, `sidebar` height is set to `100%`, but `sidebar__main` also needs to be `height: 100%` to accomplish [this type of vertical layout](https://patternfly-pr-3914.surge.sh/demos/tabs/html-demos/vertical-tabs/)~
- Extends various `hbs` templates to accept more customization
- Adds `pf-c-page-main__tabs` section. 
- Adds `pf-m-page-insets` to `pf-c-tabs`
- ~Adds `padding` to `pf-c-sidebar`~


mcarrano (commented):
These look great @mattnolting .

@gdoyle1 can you also give this a look?

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -250,6 +253,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
 
   // Vertical
   &.pf-m-vertical {
+    --pf-c-tabs--Width: var(--pf-c-tabs--m-vertical--Width);
```
mcoker, line 256: Do you think this could be breaking? Here's a probable use case where adding the width could break the layout. Though I agree this should have a block layout and consume the width of whatever it's in, I wonder if this could break people.

mcoker:
One thing I've not been certain about is the spacing for horizontal tabs in a page section. An issue is that we have specific sections with unique spacing for breadcrumbs, tertiary nav, and our generic page section container. 

This tabs demo uses our generic page section container with the padding removed. It works great if it follows another generic page section container (with the page title, or a paragraph of text, for example), since that page section will have bottom padding separating that content from the tabs.
Screen Shot 2021-03-10 at 4 33 01 PM

However the tertiary nav and breadcrumbs sections do not have bottom padding, so if tabs (which uses the generic page section with no padding) follows those elements, there is no spacing between them.

Screen Shot 2021-03-10 at 4 30 49 PM

Screen Shot 2021-03-10 at 4 35 00 PM

If we expect there to be some sort of page section with content before the tabs, I think this is fine. Another approach would be to introduce a special page section for tabs (similar to the section we have for breadcrumbs and tertiary nav) that has the padding removed by default, and maybe adds some top padding if it follows the breadcrumbs or tertiary nav.

I'm fine leaving it the way it is, but wanted to get your thoughts @mcarrano @mattnolting

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -3,6 +3,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
 
 .pf-c-tabs {
   --pf-c-tabs--inset: 0;
```
mcoker, line 5: wdyt about adding `.pf-m-page-insets` to tabs the way we have that for toolbar? Also I wonder if that should be `pf-m-inset-page`?

mcoker:
For the primary-detail demo, should it span the width of the viewport? The regular primary-detail demos do. Here's what this one looks like on a large screen. @mcarrano wdyt?

Screen Shot 2021-03-10 at 4 45 06 PM

mcoker:
@mcarrano in the design for tabs in the primary detail panel, the blue underline extends to the edge of the panel and the text is aligned with the left edge of the panel content.

Screen Shot 2021-03-10 at 4 48 40 PM


We have insets for tabs, but it's for the tab itself. The distance from the left edge of the blue underline to the text is static (16px/md spacer). To align the text in the tab (16px indent) with the panel contents (24px indent), we use an 8px spacer as the inset for the tabs.... but that creates an 8px gap. Is that OK? Or should we change that? We could also potentially add a variation for the inset of the tab text for a design like this.

Screen Shot 2021-03-10 at 4 47 48 PM

FWIW, this is what it looks like if the tab inset matches the panel inset (24px/lg spacer).

Screen Shot 2021-03-10 at 4 48 07 PM

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,813 @@
...
+                __tabs-item--attribute=(concat 'aria-controls="' tabs--id '-cluster-installation-panel"')}}
+              {{> __tabs-item
+                __tabs-item--id="subscription-support"
+                __tabs-item--aria-label="Subscription and support"
+                __tabs-item--text="Subscription and support"
+                __tabs-item--attribute=(concat 'aria-controls="' tabs--id '-subscription-support-panel"')}}
+            {{/tabs-list}}
+          {{/tabs}}
+        {{/sidebar-panel}}
+        {{#> sidebar-content sidebar-content--modifier="pf-u-p-lg"}}
```
mcoker, line 768: It would be great if we can avoid using a utility class in the demo. Maybe the sidebar should have padding like the drawer?

mcoker: Technically you could use a page section. The only thing about page sections is that they expect to be in a flex layout as flex children. Putting them in `.pf-c-page__main-group` is a way to introduce them somewhere other than a direct child of `.pf-c-page__main`. Though I think the sidebar should probably have padding and that seems like a suitable approach.

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,813 @@
...
+          {{#> tab-content tab-content--IsActive="true" tab-content--attribute=(concat 'aria-labelledby="' tab-content--id '-installation-link" id="' tab-content--id '-installation-panel"')}}
+            {{#> content}}
+              <h2>Installation content</h2>
+              <p>This document is a guide for preparing a new AWS account for use with OpenShift. It will help prepare an account tot crea a single cluster and provide insight for adjustments which may be needed for additional cluster.</p>
```
mcoker, line 772:
```suggestion
              <p>This document is a guide for preparing a new AWS account for use with OpenShift. It will help prepare an account to create a single cluster and provide insight for adjustments which may be needed for additional cluster.</p>
```

src/patternfly/demos/Tabs/templates/tabs--page-grid-view.hbs
```diff
@@ -0,0 +1,84 @@
+{{#> grid grid--modifier="pf-m-gutter pf-u-h-100"}}
```
mcoker, line 1: `pf-u-h-100` doesn't appear to do anything here.

src/patternfly/components/Card/card.scss
```diff
@@ -125,6 +126,10 @@
     }
   }
 
+  &.pf-m-fill {
```
mcoker, line 129: It doesn't seem like this is a requirement for the PR. Also looks like you're using `pf-u-h-100` in a handful of other places. I'm curious if we need this, or that use of the utility classes, or if we could make the demo designs more simple since as far as I can tell the use of `pf-m-fill` and the utility classes aren't essential to the tab layout in any way.

@mcarrano wdyt? This adds a variation to the card component that tells it to fill the available height since some of the cards in these demos fill the space like that. And then we're using utility classes in place of custom CSS for some other parts of the layout, like making these boxes below take up a bunch of vertical height using a height utility class since they don't have any content and don't take up height naturally.

Generally we avoid using utility classes in the demos since utility classes don't come with PF by default, and can be a heavy import to include them (which is why we don't include them by default), so we don't want to make it seem like a feature relies upon them being included.

I'm curious if it would be OK to alter the demo design a little so we don't have to use the utilities and add a new card variation. It would mean some of the boxes don't fill all of the height the way they do now, and maybe we don't use a grid layout in a layout like below and just remove the cards on the right.

Screen Shot 2021-03-10 at 5 13 19 PM

mcoker (changes requested):
Great work on this! Left some comments. Looks great, other than a couple of things I'd like to consider.

1) Changing the tabs width may be a breaking change. I think they should take up the full width, I'm just concerned about someone using tabs now and how changing the width could impact them. Maybe we could introduce that as a variation?

2) I'm wondering if the use of utility classes and the new `pf-m-fill` variation on the card are necessary. Typically we demo what PF components can do using the existing variations, and add an enhancement if it's a requirement for the feature. Seems like those classes are just to support parts of the demo that aren't necessary and could be left to a product to implement if that's what they wanted. Curious what @mcarrano thinks.

src/patternfly/components/Card/card.scss
mcarrano: Perhaps try to simplify this and we can take a look at it. I understand the issue and agree that this demo should focus on the tabs. Other elements on the page are not so relevant as long as the final result looks presentable.

mcarrano:
@mceledonia @gdoyle1 See @mcoker 's question about the alignment of tabs in primary-detail. I really don't have a strong opinion about this other than to say I'd prefer to to introduce any new variants to make this work.

src/patternfly/components/Card/card.scss
mattnolting: `pf-m-fill` is necessary for cards inside layout `__item` s as they don't fill the height of their containers. I used it here in the demo context.

mcoker: We could also put content in either of the cards on the right to give them height (or just remove them), and put content in the tab content to give it height, since to make them as tall as they are here with no/empty content requires a new variation to the card.

mattnolting: This is the sitch I'm describing. We need a `full height` option if using layout `item` wrappers. Although they aren't necessary (which we've noted), to make cards full height if you are using them, say for `Gallery` or `Grid`:

Before 

Screen Shot 2021-03-12 at 9 48 06 AM

After

Screen Shot 2021-03-12 at 9 52 32 AM

Before

Screen Shot 2021-03-12 at 9 51 29 AM

After

Screen Shot 2021-03-12 at 9 51 46 AM

Before:
 This is the most pertinent example. **`pf-l-grid__item` is necessary to span grid columns.In that case, the card has no way of consuming full height of its parent, unless using a `card modifier` or a `utility class / pf-u-h-100`.**

Screen Shot 2021-03-12 at 9 51 16 AM

After

Screen Shot 2021-03-12 at 9 52 03 AM

gdoyle1:
@mcoker @mcarrano @mattnolting 

1) Tabs in page section -- we don't expect anyone to use tabs without a page section, so I think it's fine! We'll have that detail in the design guidelines as well, that the tabs should be placed underneath a title on the page.

2) Tabs in primary detail -- Tabs can stay as you have them (so they won't fill the whole space of the panel) - we are allowing people to choose whatever they'd want in that case (default vs filled I mean). I actually do think that the tabs look better with the 24px margin on the sides so it aligns with the content. I'd like to hear what @mceledonia thinks as well though!

src/patternfly/components/Tabs/tabs.scss
mattnolting: Yeah, I suppose it could be. All we need to change here is `flex-grow: 1`. I'll revert.

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
...
+
+### Tables and tabs, auto width tabs
+```hbs isFullscreen
+{{#> tabs--page-wrapper tabs--page-wrapper--id="table-tabs-w-secondary-example"}}
+  {{#> page-main-section page-main-section--IsLimitWidth="true" page-main-section--modifier="pf-m-light"}}
+    {{#> title title--modifier="pf-m-2xl pf-u-mt-md"}}
+      Nodes
+    {{/title}}
+  {{/page-main-section}}
+  {{#> page-main-tabs page-main-tabs--IsLimitWidth="true" page-main-tabs--modifier="pf-m-light pf-m-tabs"}}
```
mcoker, line 402: You've got `.pf-m-light.pf-m-tabs` on a bunch of these - do you need those? Looks like you set the background color on `.pf-c-page__main-tabs` and that modifier isn't defined in the CSS anywhere.

src/patternfly/components/Page/examples/Page.md
```diff
@@ -205,6 +208,7 @@ This component provides the basic chrome for a page, including sidebar, header,
 | `.pf-c-page__main-nav` | `<section>` |   Creates a container to nest the navigation component in the main page area. |
 | `.pf-c-page__main-breadcrumb` | `<section>` |   Creates a container to nest the breadcrumb component in the main page area. |
 | `.pf-c-page__main-section` | `<section>` |  Creates a section container in the main page area. **Note: The last/only `.pf-c-page__main-section` element will grow to fill the availble vertical space. You can change this behavior using `.pf-m-fill` and `.pf-m-no-fill`, which are documented below.**  |
+| `.pf-c-page__main-tabs` | `<section>` | Creates a container to nest the tabs component in the main page area. |
```
mcoker, line 211: Is this specific section needed? Per @gdoyle1's [comment ](https://github.com/patternfly/patternfly/pull/3914#issuecomment-797554356) it doesn't look like we need that page section, and can just use `.pf-c-page__main-section` instead.

> Tabs in page section -- we don't expect anyone to use tabs without a page section, so I think it's fine! We'll have that detail in the design guidelines as well, that the tabs should be placed underneath a title on the page.


The problem with it is that we can't guarantee users will use the right page section to wrap the tabs. `.pf-c-page__main-[nav/breadcrumbs]` work because the tertiary nav and breadcrumbs are props on the `<Page>` component, so react can put the nav/breadcrumbs in the appropriate page section for the user. Since we don't know where tabs will go on the page, I don't think we can support adding tabs to a page via a prop, which means users will need to know to put it in `.pf-c-page__main-tabs.` I just wonder if it will lead to inconsistencies.

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
...
+      {{/content}}
+    {{/tab-content}}
+  {{/page-main-section}}
+{{/tabs--page-wrapper}}
+```
+
+### Tables and tabs, auto width tabs
+```hbs isFullscreen
+{{#> tabs--page-wrapper tabs--page-wrapper--id="table-tabs-w-secondary-example"}}
+  {{#> page-main-section page-main-section--IsLimitWidth="true" page-main-section--modifier="pf-m-light"}}
```
mcoker, line 397:
```suggestion
  {{#> page-main-section page-main-section--modifier="pf-m-light"}}
```

src/patternfly/demos/Tabs/examples/Tabs.md
mcoker, line 402:
```suggestion
  {{#> page-main-tabs page-main-tabs--modifier="pf-m-light pf-m-tabs"}}
```

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
...
+          __tabs-item--attribute=(concat 'aria-controls="' tabs--id '-nodes-panel"')}}
+        {{> __tabs-item
+          __tabs-item--id="node-connectors"
+          __tabs-item--aria-label="node-connectors"
+          __tabs-item--text="Node connectors"
+          __tabs-item--attribute=(concat 'aria-controls="' tabs--id '-node-connectors-panel"')}}
+      {{/tabs-list}}
+    {{/tabs}}
+  {{/page-main-tabs}}
+  {{#> page-main-section page-main-section--IsLimitWidth="true" page-main-section--modifier="pf-m-no-padding pf-m-light" tab-content--id=(concat tabs--page-wrapper--id '-tabs')}}
```
mcoker, line 419:
```suggestion
  {{#> page-main-section page-main-section--modifier="pf-m-no-padding pf-m-light" tab-content--id=(concat tabs--page-wrapper--id '-tabs')}}
```

mcoker: This will match the demo above it. The drawer area should be full width.

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
+---
+id: 'Tabs'
+section: demos
```
mcoker, line 3: so it shows up as the demos tab on the component, not in the demos section in the vertical nav

```suggestion
section: components
```

mattnolting:
@mcoker updated

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
...
+
+### Open tabs
+
+```hbs isFullscreen
+{{#> tabs--page-wrapper tabs--page-wrapper--id="open-tabs-example"}}
+  {{> page-template-breadcrumb page-template-breadcrumbs--IsPods="true"}}
+  {{#> page-main-section page-main-section--IsLimitWidth="true" page-main-section--modifier="pf-m-light"}}
+    {{> tabs--pod-page-title}}
+  {{/page-main-section}}
+  {{#> page-main-tabs page-main-tabs--IsLimitWidth="true" page-main-tabs--modifier="pf-m-light pf-m-tabs"}}
```
mcoker, line 16: Do you need that? Looks like you set `background-color` on `.pf-c-page__main-tabs`. If not, it's in other spots in the demos, too.

```suggestion
  {{#> page-main-tabs page-main-tabs--IsLimitWidth="true" page-main-tabs--modifier="pf-m-tabs"}}
```

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
...
+          {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close dialog"'}}
+            <i class="fas fa-times" aria-hidden="true"></i>
+          {{/button}}
+          {{#> modal-box-header}}
+            {{#> modal-box-title modal-box-title--attribute=(concat 'id="' example-wrapper--id '-modal-title"')}}
+              Modal title
+            {{/modal-box-title}}
+          {{/modal-box-header}}
+          {{#> modal-box-body modal-box-body--attribute=(concat 'id="' example-wrapper--id '-modal-description"')}}
+            {{#> grid grid--modifier="pf-m-gutter"}}
```
mcoker, line 482: Why are we using grid here but flex in https://patternfly-pr-3914.surge.sh/components/tabs/html-demos/nested-tabs/?

src/patternfly/demos/Tabs/examples/Tabs.md
```diff
@@ -0,0 +1,814 @@
...
+          __tabs-item--attribute=(concat 'aria-controls="' tabs--id '-support-panel"')}}
+      {{/tabs-list}}
+    {{/tabs}}
+  {{/page-main-tabs}}
+  {{#> page-main-section page-main-section--IsLimitWidth="true" tab-content--id=(concat tabs--page-wrapper--id '-tabs') page-main-section--modifier="pf-m-light"}}
+    {{#> tab-content tab-content--attribute=(concat 'aria-labelledby="' tab-content--id '-new-link" id="' tab-content--id '-new-panel"')}}
+      What's new content
+    {{/tab-content}}
+    {{#> tab-content tab-content--IsActive="true" tab-content--attribute=(concat 'aria-labelledby="' tab-content--id '-get-started-link" id="' tab-content--id '-get-started-panel"')}}
+      {{#> grid grid--modifier="pf-m-gutter"}}
```
mcoker, line 633: Also grid but flex in another example that looks like the same layout as far as I can tell.

src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -62,7 +61,6 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
     --pf-c-sidebar__main--FlexDirection: var(--pf-c-sidebar__main--md--FlexDirection);
     --pf-c-sidebar__panel--BoxShadow: none;
     --pf-c-sidebar__panel--FlexBasis: var(--pf-c-sidebar__panel--md--FlexBasis);
-    --pf-c-sidebar__main--AlignItems: var(--pf-c-sidebar__main--md--AlignItems);
```
mcoker, line 65: This breaks the sidebar sticky panel

mcoker (changes requested):

mcoker (approved):
L🥳TM!

PR #3922, opened 2021-03-09
feat(brand): added picture + srcset

### LINKED ISSUE #3666: Enhancement: Brand
Enhance `pf-c-brand` to use `<picture>` element along with responsive images `srcset`. 

https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

mattnolting:
closes #3666 

[Brand preview](https://patternfly-pr-3922.surge.sh/components/brand/)

What this PR does:
- Adds `picture` element as `pf-c-brand` option
- Adds `srcset` element option
- Adds responsive `width` variable
- Adds responsive `height` variable


mcarrano (commented):
Looks fine to me @mattnolting . Can you explain what the goal is for making this change?

src/patternfly/components/Brand/brand.scss
```diff
@@ -1,4 +1,14 @@
-// stylelint-disable
+$pf-c-brand--breakpoint-map: build-breakpoint-map();
+
 .pf-c-brand {
+  --pf-c-brand--Width--base: auto;
+  --pf-c-brand--Height--base: auto;
+
+  display: flex;
```
mcoker, line 7: This will be breaking with the existing component since it will make the inline `img` have block formatting. Maybe just limit this to `.pf-m-picture`?

mattnolting:
> Looks fine to me @mattnolting . Can you explain what the goal is for making this change?

It allows the image switching between viewport sizes. 

This

Screen Shot 2021-03-23 at 9 11 37 PM

To this

Screen Shot 2021-03-23 at 9 11 50 PM

mattnolting:
@mcoker updated

src/patternfly/components/Brand/brand.scss
```diff
@@ -1,4 +1,17 @@
-// stylelint-disable
+$pf-c-brand--breakpoint-map: build-breakpoint-map();
+
 .pf-c-brand {
+  --pf-c-brand--Width--base: auto;
+  --pf-c-brand--Height--base: auto;
+
+  width: var(--pf-c-brand--Width--base);
+  max-width: 100%;
```
mcoker, line 8: I'm wondering if we should also move this in to `&.pf-m-picture`. I'm trying to think of how this might break someone, and if they have applied `.pf-c-brand` to something other than an `img`, and had custom CSS loaded before patternfly, this might change their layout. An `img` should already get it from https://github.com/patternfly/patternfly/blob/51955caf5d1dba384a7cfa1dd454d20f1caf5d1c/src/patternfly/base/_globals.scss#L65-L73

WDYT about moving it in to `.pf-m-picture`, and leaving a comment to move it in the next breaking change release?

src/patternfly/components/Brand/brand.scss
```diff
@@ -1,4 +1,17 @@
...
+  height: var(--pf-c-brand--Height--base);
+
+  @include pf-build-css-variable-stack("--pf-c-brand--Width--base", "--pf-c-brand--Width", $pf-c-brand--breakpoint-map);
+  @include pf-build-css-variable-stack("--pf-c-brand--Height--base", "--pf-c-brand--Height", $pf-c-brand--breakpoint-map);
+
+  &.pf-m-picture {
+    display: flex;
```
mcoker, line 15: Could this be `inline-flex`? `<picture>` is inline.

src/patternfly/components/Brand/brand.scss
mattnolting: WFM

src/patternfly/components/Brand/brand.scss
mattnolting: Yep

mattnolting:
@mcoker updated

mcoker (approved):
LPTM!!

Issue #3904, opened 2021-03-03
Expandable wizard substeps


mcarrano:
Add the option to show/hide substeps in a Wizard inside of an expandable section. Designs here: https://marvelapp.com/prototype/142h0f9e

PR #3927, opened 2021-03-15
feat(wizard): added expandable subsections

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3904

### DIFF
--- a/src/patternfly/components/Wizard/examples/Wizard.md
+++ b/src/patternfly/components/Wizard/examples/Wizard.md
@@ -186,6 +186,184 @@ import './Wizard.css'
 {{/wizard}}
 ```
 
+### Expandable collapsed
+```hbs isFullscreen
+{{#> wizard wizard--id="wizard-basic"}}
+  {{#> wizard-header}}
+    {{#> button button--modifier="pf-m-plain pf-c-wizard__close" button--attribute='aria-label="Close"'}}
+      <i class="fas fa-times" aria-hidden="true"></i>
+    {{/button}}
+    {{#> title title--modifier="pf-m-3xl pf-c-wizard__title"}}Wizard title{{/title}}
+    {{#> wizard-description}}
+      Here is where the description goes
+    {{/wizard-description}}
+  {{/wizard-header}}
+  {{#> wizard-toggle}}
+    {{#> wizard-toggle-list}}
+        {{#> wizard-toggle-list-item}}
+          {{#> wizard-toggle-num}}2{{/wizard-toggle-num}}
+          Configuration
+          {{> wizard-toggle-separator}}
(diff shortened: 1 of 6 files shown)


mcarrano (commented):
Looks good, but I had a couple of questions.

* Should the caret styling be different than the expandable step label? Is that consistent with what we do for accordions for example?
* When the steps are expanded, I'm wondering if there should be a hover effect.

Thoughts @maryshak1996 ?

maryshak1996:
@mcarrano @mcoker I think that the ideal interaction for for expand collapse would be to mimic the caret coloring we see in the expandable data list (https://www.patternfly.org/v4/components/data-list#expandable) -- so the caret should be grey, but then when the parent step is hovered on, it would turn black.

mcoker:
@maryshak1996 @mcarrano updated. A side note, I wonder if we should audit our expandable components and see if any should be updated. As an example, looking at accordion and expandable section, the arrow is always black - but data-list, table, and card mimic the plain button styling.

mcarrano (approved):
This looks fine to me. Are you good with it @maryshak1996 ?

As for other components, I'm betting we have several inconsistencies, but not willing to go down that road now :)

src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -186,6 +186,184 @@ import './Wizard.css'
...
+        {{/button}}
+      {{/wizard-footer-cancel}}
+    {{/wizard-footer}}
+  {{/wizard-outer-wrap}}
+{{/wizard}}
+```
+
+### Expandable expanded
+```hbs isFullscreen
+{{#> wizard wizard--id="wizard-basic"}}
```
mattnolting, line 280: Need new `id`

mattnolting (commented):
This is looking good! What do you think about wrapping all `.pf-c-wizard__nav-link` text with `.pf-c-wizard__nav-link-text`, then you wouldn't need the `.pf-m-expandable` class.

And can you generate screenshots for your new examples?

src/patternfly/components/Wizard/examples/Wizard.md
mcoker: updated!

mcoker:
> What do you think about wrapping all .pf-c-wizard__nav-link text with .pf-c-wizard__nav-link-text, then you wouldn't need the .pf-m-expandable class.

@mattnolting the modifier is an opt-in to `display: flex;` on `.pf-c-wizard__nav-link`, since adding it would break core/non-react users.

mattnolting (approved):
LGTM!

maryshak1996 (approved):

Issue #3898, opened 2021-03-03
Add active state for Buttons


mcarrano:
Add active state styling to the button component per the designs here: https://marvelapp.com/prototype/14210469/screen/76778608

PR #3928, opened 2021-03-16
feat(button): added active state for buttons, links

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3898

@mcarrano @doruskova since we're updating the inline link styles to remove the underline on `:active`, I'm assuming we'll also want to update the global link styles to match? I've done that in this PR, though I wonder if that will be a breaking change as it could seem unexpected and users may have styled other parts of their product to mimic PF links, and now their active styling may not match ours.

### DIFF
--- a/src/patternfly/base/_globals.scss
+++ b/src/patternfly/base/_globals.scss
@@ -120,6 +120,10 @@
       --pf-global--link--Color: var(--pf-global--link--Color--hover);
       --pf-global--link--TextDecoration: var(--pf-global--link--TextDecoration--hover);
     }
+
+    &:active {
+      --pf-global--link--TextDecoration: var(--pf-global--link--TextDecoration--active);
+    }
   }
 
   button,
--- a/src/patternfly/base/_variables.scss
+++ b/src/patternfly/base/_variables.scss
@@ -205,6 +205,7 @@
   --pf-global--link--Color--dark--hover: #{$pf-global--link--Color--dark--hover};
   --pf-global--link--TextDecoration: #{$pf-global--link--TextDecoration};
   --pf-global--link--TextDecoration--hover: #{$pf-global--link--TextDecoration--hover};
+  --pf-global--link--TextDecoration--active: #{$pf-global--link--TextDecoration--active};
 
   // Borders
   --pf-global--BorderWidth--sm: #{$pf-global--BorderWidth--sm};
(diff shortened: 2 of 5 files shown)


mcarrano (commented):
Good question about the global link style @mcoker . Looking at this raises some additional design issues for me, also:

* I like all of the buttons except the Warning variant. This is the only place that the label flips from black to white. I'm questioning if we really want that. I'm guessing that's done to maintain contrast for a11y compliance, but does that really matter in this state which is more transient?
* Do we need to also consider 'active' states for plain/icon button and control buttons? Looks like we currently are not.

Thoughts @doruskova @mceledonia ?

mcoker:
> Do we need to also consider 'active' states for plain/icon button and control buttons? Looks like we currently are not.

@mcarrano this would likely imply additional work like updating the menu toggles and components where we have a plain action that isn't the plain button component variation. And updating the control variation might imply we update form control elements, too, to match?

mceledonia:
I don't think we need them on the plain actions, but at the same time I question the purpose of having this on some actionable items and not others. Is there a clear design rationale for these, and if so does it apply to all actions? If so, we probably should apply this to everything, but could handle the low hanging fruit like buttons now and clean up the rest later on.

@mcarrano @mcoker I definitely agree about the text flipping on the warning variant, I can't imagine we need that state to be accessible given how transient it is.

doruskova:
Agree that we don't need it on the plain actions (for now) and agree about the text color for the warning button (confused from the dark theme, haha). However, I'd start with this state for buttons and then we can do the rest later because I think it'll need more analysis.

mattnolting (approved):
LGTM. @mcarrano @doruskova is the intention of the design to replace browser default focus ring? I know we've talked about removing focus rings before, but it seems that we are providing enough visual indication that `button`s specifically are `focused/active/hovered`.

doruskova:
@matthewcarleton The focus ring will still be there. The intention was to differentiate hover and active/pressed state because they had the same visual style.

mcarrano:
@christianvogt we currently have this PR open that addresses this issue that you filed a while ago: https://github.com/patternfly/patternfly-design/issues/907  It would add active states to Button components, only. We placed it on hold due to concerns about creating inconsistencies in the behavior of components should we do this only for buttons but not other elements.

While I think we all agree that active states make sense, our preference is to address this as part of a larger visual refresh. I wanted to check to see if there is a specific problem you were needing to address in the short term (and what that is) or if this can wait to be addressed at a future date? Please advise.

christianvogt:
@mcarrano my request was for active states for all interactive elements. I have no requirement for active states on buttons or all components in general. I would like to see the active state adopted for all components in the future.

PR #3929, opened 2021-03-16
fix(dropdown): reverted badge menu to dropdown menu

### LINKED ISSUE #3923: Update badge dropdown toggle (and breadcrumb and menu examples) to use dropdown menu
Currently those examples use the menu component for the menu, and we haven't come up with a plan for how to use the menu component with toggles, existing menus (dropdown, select, etc). The requirements for the badge/breadcrumb dropdown work fine with the existing dropdown menu, so let's revert that menu back to the standard dropdown menu until we have a bigger plan for how to use the menu component and have time to plan and integrate that work.

mattnolting:
closes #3923


mcoker (changes requested):
Looks good, but looks like we also need to update the menu used in the menu component drilldown breadcrumbs examples. https://patternfly-pr-3929.surge.sh/components/menu#drilldown-with-breadcrumbs--level-3

mattnolting:
@mcoker updated!

mcoker (approved):
L🥔TM!

Issue #3902, opened 2021-03-03
Add optional item counts to Select menu


mcarrano:
Want to add optional item counts to select menus. Designs here: https://marvelapp.com/prototype/c18hjf1/screen/76910983

This is useful primarily for checkbox multi-select menus, but should not be restricted to that.

PR #3931, opened 2021-03-16
feat(select): added support for item count

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3902

@mcarrano I wasn't sure the best way to show the counts, so I added a new example with different types of menu items that have a count. Let me know if that works.


mcarrano (changes requested):
A couple of comments @mcoker .

* While the item counts can be used with any menu, they make the most sense with a checkbox select for faceted filtering. Can you change the example to a checkbox select?
* For items that are showing the blue checkmark to indicate selection, I think it's odd to have the checkmark pushing the count to the left. Was this out design intent @mmenestr ? I'd prefer to see the count values in a vertically aligned column. See https://patternfly-pr-3931.surge.sh/components/select#counts
* Not sure if this is just an artifact of the workspace, but when I shrink the screen, the counts on the right hand edge are getting cut off.
* @mcoker while we are in here, would it be possible to fix this page so that the TOC items properly show in the right-hand jump menu to make it easy to actually find example? If not here, can we create a separate PR to do that?
*

mmenestr:
I would agree with everything Matt said, this makes more sense for a checkbox select so I would definitely include an example of that.
And in terms of the placement in a multi select with the blue checkmark, it's true that it looks weird when the counts aren't aligned. But if not all options have a count, nor a checkmark, is it possible to make it so that things don't automatically just right align, and they stay in a specific "column"? Ie: counts would be right aligned, and checkmarks would show up in a "column" to the left whether there's a number count or not for that option - like this: 

image

mcoker:
@mmenestr unfortunately due to limitations in CSS we can't have a layout where the count acts as a column. If the count column were a fixed width, we could, but I'm assuming it would need to be a variable width based on the count value. 

Using a fixed column, we could do something like this

Screen Shot 2021-03-18 at 4 29 02 PM

Or truncation with "..."

Screen Shot 2021-03-18 at 4 30 20 PM

mmenestr:
Ok i mean honestly that seems fine to me, especially since we don't envision most people using this component for something like this, and more the checkbox select! What do you think @mcarrano?

mcarrano (approved):
Agree with @mmenestr . This example works much better. Thanks @mcoker !

mcoker:
@mcarrano @mmenestr do y'all want to support the count being used in a regular/non-checkbox select? If we go with one of the truncated layouts above, I might need to build this differently.

mcarrano:
I don't see that as a use case. Do you @mmenestr ?

mmenestr:
Yeah I don't really see one either - I think as long as it can be used in a checkbox select, and optionally in the catalog view component as well, it should be good to go!

mcoker:
> Yeah I don't really see one either - I think as long as it can be used in a checkbox select, and optionally in the catalog view component as well, it should be good to go!

@mmenestr this PR adds the count as an element of the select menu. Looks like the catalog view component uses the check component with a custom "item count" element. So this PR does not add functionality to or impact the catalog view's usage of the item count in the check component.

Should we move this count feature to the check component instead, and then consume the check's count in the select menu, catalog view, and anywhere else the check component is used?

mmenestr:
I feel like that might actually make more sense, since that would allow it to be be used across different components! What do you think @mcarrano? The whole idea was to keep this use case presented in the same way across components - with the use case being showing the number of items that would match a certain "filter" selection, whether that be in a select list or a catalog view.

mcarrano:
> Should we move this count feature to the check component instead, and then consume the check's count in the select menu, catalog view, and anywhere else the check component is used?

@mcoker I think that makes some sense, although the functionality is pretty specific to checkboxes when used in the context of a menu. I think that the real problem is that we don't have a generic menu component that is used either in a dropdown or in a catalog view sidebar.

Do we just pull in the checkbox component now when we build this menu? If so, then maybe it does make sense to shift the functionality there to give us a consistent approach. Thoughts?

mcoker:
> I think that makes some sense, although the functionality is pretty specific to checkboxes when used in the context of a menu.

My thoughts are that the radio/check's primary function is in a form. A count could belong, or maybe some sort of extra info associated with the element (where you could place a count, but it isn't limited to being a count), but particularly a right-aligned count, I agree that it seems like a specific use of the radio/check in a menu/list.

> Do we just pull in the checkbox component now when we build this menu?

In catalog view, yes. It's a custom built menu (catalog view CSS that doesn't come from core) that include the PF components:

* Text input - top filter search
* A somewhat random use of the form components fieldset element which doesn't do much
* Check component for the menu items
* Button link variation for the "show [n] more" link

Otherwise the titles, spacing, counts, icons, etc are all part of the catalog view's custom CSS.

> I think that the real problem is that we don't have a generic menu component that is used either in a dropdown or in a catalog view sidebar.

Yep, though that seems fair since you can put a lot of things in a select menu item that you likely wouldn't want to use in the catalog view list, and vice versa. It could get tricky trying to have a single menu that works in both use cases, unless we knew we would keep the catalog view menu pretty simple. 

Thinking about it some more, I don't know that making the count a feature of the check/radio component would work if we want to place the select's selected checkmark icon to the left of the count. The check component is only used where the text is in a select menu item. The checkmark icon is part of the outer container. We would need to take the checkmark out of the parent container and move it inside the check component to allow the count to display to the right of the checkmark icon. We can do that, it just seems more to me that the selected checkmark icon and count (and any other icons/things we display in a select menu item like that) are part of the select component. So assuming that the list used in the catalog view and list in the select/dropdown/etc menu are different, it seems like we would make the counts a feature of both the select and catalog view lists.

mmenestr:
Oh wait, when you say check component, do you mean the one with the blue checkmark to the far right? I thought you meant just the checkbox select component where you have the checkbox on the left of the option, and then just the number count on the far right. Thats what would be used in catalog views, not the blue checkmark component

mcoker:
@mmenestr oh sorry, that's probably confusing. By check component, I just mean the interactive checkbox/text pair. The selected blue check icon is a part of the select component, not the check component.

mcarrano:
@mcoker In reading all this, I think we should keep this enhancement localized to the Select component. It avoids a lot of potential complications. The catalog view sidebar is something we need more discussion about, I think. Is that something that is even officially supported right now in PatternFly? I know that different teams have done custom implementations and I think we'd like to have a common implementation that can be shared. But that's really a separate issue, I think.

mcoker:
> The catalog view sidebar is something we need more discussion about, I think. Is that something that is even officially supported right now in PatternFly?

@mcarrano it's a react extension, so I'm not sure what that means re: official support. As far as the html/css it's built with, it's mostly custom (doesn't come from core), with the PF components I mentioned above that are used. The titles, spacing, things like the icon below are all custom elements that only exist in this extension and don't come from core.

Screen Shot 2021-03-22 at 4 07 54 PM

mattnolting (commented):
Looks great! What do you think about changing the name of `-count` to something more generic like `-secondary` with the same styling? Then, in the event where the user wants to add something else (thinking a `beta label` or a `info popover` as examples), that would go inside `secondary content`.

mcoker:
> Looks great! What do you think about changing the name of -count to something more generic like -secondary with the same styling? Then, in the event where the user wants to add something else (thinking a beta label or a info popover as examples), that would go inside secondary content.

@mattnolting that could work, though I'm probably in favor of leaving the count since it's a specific element and could potentially get different styling, spacing, etc than a more generic container, but adding that generic secondary container if users ask for more flexibility with putting stuff in the menu item. wdyt? cc @mcarrano

mattnolting (approved):
LPTM!

Issue #1094, opened 2018-12-11
Feature Request: Secondary/Tertiary Danger Button


TimoSta:
**Current State**
[Patternfly 4 buttons](https://pf-next.com/components/Button/examples/) currently come in different designs, each with different semantics.

This includes the following:
- primary
- secondary
- tertiary
- danger

**Requested Feature**
The current set of options does not allow buttons with are both _secondary_ or _tertiary_, and _danger_.

We would like to correctly model a button which is not a primary action on the page and will rarely be used, but is destructive and should be marked as such.

**Example Use Case**
A project overview page in an app gives access to project header info, and allows the user to execute a bunch of actions, including deleting the project. This option will rarely be used:

![image](https://user-images.githubusercontent.com/5621996/49821457-7b207f00-fd48-11e8-9208-aa6399e961e3.png)

After some (highly unscientific) user testing, it seems that the big red button draws too much attention away from the rest of the page, considering that it should not be the focus.

**Potential Solution**
A combination of the _secondary_ and _danger_ style might be appropriate:

![image](https://user-images.githubusercontent.com/5621996/49821676-0c8ff100-fd49-11e8-92fe-a2b32ed87775.png)

---

**Please tell me why my interpretation of the PF4 design system is horribly wrong and how we should approach this use case instead** :slightly_smiling_face:

dgutride:
@mcarrano @kybaker  - can you guys weigh in on this request or how you'd like to see these?

kybaker:
@dgutride @rachael-phillips @mcarrano I think this is a valid usecase we should accommodate. Having a secondary destructive button allows us to have multiple levels severity in the same screen e.g. account delete and suspend. 

@rachael-phillips Could you create a story to add this post beta?

ryelo:
@rachael-phillips Can we also include a danger "link"? So a button that doesn't have any borders. 

It was presented to us in mockups, so it would be awesome to be included.

rachael-phillips:
@ryelo that sounds like a good idea to me. @kybaker @mcarrano what are your thoughts on the danger link suggestion?

rachael-phillips:
Also, can you tell me more about your timeline for these requests and what priority this is for you?

TimoSta:
@rachael-phillips We are planning to ship the first version of our product around April, so it would be nice to have it in place by then.

That said, it is a visual issue and does not negatively impact any functionality, so it probably can't be considered the highest of priorities.

[...]

PR #3932, opened 2021-03-16
feat(button): added secondary/link danger variations

mcoker:
fixes https://github.com/patternfly/patternfly/issues/1094

this adds active states to the new variations that match https://github.com/patternfly/patternfly/pull/3928

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -7,6 +7,7 @@
   --pf-c-button--LineHeight: var(--pf-global--LineHeight--md);
   --pf-c-button--FontWeight: var(--pf-global--FontWeight--normal);
   --pf-c-button--FontSize: var(--pf-global--FontSize--md);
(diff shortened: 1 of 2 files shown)


mcarrano (commented):
Looks good @mcoker . But for consistency should we also have a "small" version of the secondary danger button?

mcarrano (approved):
Looks great. Thanks @mcoker !

mceledonia (approved):
Thanks @mcoker looks great!

mattnolting (approved):
Looks great! One thing I might consider is that with increasing number of variations, we might separate them by example. There are enough variants at this point, I get a little lost looking at all the buttons in the same space without labels to distinguish between them.

mcoker:
@mattnolting agree, I actually did that in this PR then reverted it since it wasn't super clear what the best structure would be. Created an issue for that here - https://github.com/patternfly/patternfly/issues/3946

Issue #3906, opened 2021-03-03
Add new Copy to Clipboard variants


mcarrano:
Want to enhance the Clipboard Copy component to add new variants for single line borderless and inline option. The goal is to provide light-weight, read-only versions of the components that can include additional actions per this design issue. https://github.com/patternfly/patternfly-design/issues/1009  

@lwigh can you attach a link to your final design mocks?

lwrigh:
Here is the Marvel link to the inline and single line mocks: https://marvelapp.com/prototype/g39724e The first mock features the step by step interaction points that are out of context of the Quick Starts. Let me know if any other additional states need to be documented.

mcarrano:
Thanks for posting @lwrigh 

@mcoker please take a look at this when you get a chance. I wasn't sure if the Inline and Single line were really different. Seems like the only difference in the design is the height of the gray background. Not sure what drives that height, but I think the idea is that when placed within a paragraph of text, this shouldn't need to change or increase the text spacing.

mcoker:
@lwrigh I see a feature of the inline variation is the ability to click on the code and copy to clipboard.

<img width="274" alt="Screen Shot 2021-03-08 at 5 58 43 PM" src="https://user-images.githubusercontent.com/35148959/110397940-640c8c80-8038-11eb-875b-64cbe11b0b34.png">

But also if there is overflow text, clicking on the text shows a text cursor and allows the user to scroll/select the text in the input like a regular read-only text input (like the read-only inputs here - https://pf4.patternfly.org/components/clipboard-copy#expandable)

<img width="612" alt="Screen Shot 2021-03-08 at 5 58 52 PM" src="https://user-images.githubusercontent.com/35148959/110398024-91593a80-8038-11eb-9db8-334c63ce2e19.png">

How would you differentiate those two features? When would clicking on one copy the text, and clicking on the other focuses the text and lets you scroll in the text input?

Also @mcarrano it seems like for consistency, we might want to support that basic feature (clicking the text to copy it) in the current clipboard copy component?

mcarrano:
@mcoker That's a good point about enabling a mouse event on the field. @lwrigh do you feel that's an important part of the design? Would it be OK to require the user to click the copy button, only, to initiate the action?

lwrigh:
> @lwrigh I see a feature of the inline variation is the ability to click on the code and copy to clipboard.
> 
[...]

PR #3933, opened 2021-03-17
feat(clipboard-copy): added inline variant

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3906

### DIFF
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-actions-item.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-actions-item.hbs
@@ -1,0 +1,6 @@
+<span class="pf-c-clipboard-copy__actions-item{{#if clipboard-copy-actions-item--modifier}} {{clipboard-copy-actions-item--modifier}}{{/if}}"
+  {{#if clipboard-copy-actions-item--attribute}}
+    {{{clipboard-copy-actions-item--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</span>
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-actions.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-actions.hbs
@@ -1,0 +1,6 @@
+<span class="pf-c-clipboard-copy__actions{{#if clipboard-copy-actions--modifier}} {{clipboard-copy-actions--modifier}}{{/if}}"
+  {{#if clipboard-copy-actions--attribute}}
+    {{{clipboard-copy-actions--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</span>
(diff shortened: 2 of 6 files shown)


mcarrano (approved):
This looks great @mcoker .

christianvogt:
If you put two code block immediately adjacent they end up looking like this. Can you tell what is being copied by each?

image

lwrigh:
> If you put two code block immediately adjacent they end up looking like this. Can you tell what is being copied by each?
> 
> image
> 
> Do we need a border? a hover state? a less then line-height background?

I think adding a white border or adding a bit of white space at the beginning and end of the inline code snippet would do the trick if there's one right after the other.

mcoker:
Hey @christianvogt! If you made that change in the PR, the space won't show up because our workspace removes the whitespace between inline elements, so as long as an application doesn't remove that whitespace, the a space character will exist between two inline clipboard copy elements like this. In our examples, you would need to add a `&nbsp;` between them so that the whitespace isn't removed.

Screen Shot 2021-03-17 at 12 15 00 PM

christianvogt:
@mcoker yes I forgot to add explicit spacing between the two blocks. But do we still need to separate the line backgrounds vertically?

image

Also, when using `<code>`. Should the outer `div` changed to `code`? or should the inner `span` (`<code class="pf-c-clipboard-copy__text"`) be changed?

mcoker:
> Also, when using <code>. Should the outer div changed to code? or should the inner span (<code class="pf-c-clipboard-copy__text") be changed?

We could use `<code>` although a user could put anything they want in there, so it's not always going to be code. If we did support that as the default element or as a `component` variant to specify the element used to wrap the text, we would need additional styling to override the monospace font the browser automatically adds when using `<code>`. @jessiehuff WDYT?

christianvogt:
I'm not saying to have it as the default. But I want to be able to use the `code` element as part of this component. This will need to be supported in the react-component. We will want the monospaced for our use case.

My question was more directly to the implementation and whether we should replace the outer `div` or the inner `span` when we want to use the `code` element.

mcoker:
> I want to be able to use the code element as part of this component. This will need to be supported in the react-component. We will want the monospaced for our use case.

@mcarrano @lwrigh is that a variation we should support in core? In order to get the PF monospaced font stack with a `<code>` block, we will need to add support for that in the clipboard copy component.

> whether we should replace the outer div or the inner span when we want to use the code element.

@christianvogt the inner `span.pf-c-clipboard-copy__text` element since you wouldn't want to include the copy button or any additional actions.

mcarrano:
@mcoker I think it makes sense to allow using a `<code>` block inside this component. Is that something that can be doe as part of this PR?

mcoker:
@christianvogt @lwrigh @mcarrano added the code variant here - https://patternfly-pr-3933.surge.sh/components/clipboard-copy#inline

lwrigh:
That's looking good to me! My only comment is is there any way to include an automatic space after the code snippet ends in the long string that wraps and the in sentence one?

mcarrano:
Looks good to me, too.

mcoker:
>  is is there any way to include an automatic space after the code snippet ends in the long string that wraps and the in sentence one?

@lwrigh yep, we made a workspace change that removed it so I updated the example to add it back. Let me know if it looks OK now.

lwrigh:
That looks good to me!

mattnolting (approved):
LPTM!

christianvogt:
@mcoker @lwrigh i just wanted to point out that i've had some inquiries about how the background of inline code blocks touch the line above making it harder to distinguish separate blocks. This can be seen in the screenshots posted in this PR.
Is there a chance we can address this or are we set on the final design?

mcoker:
@christianvogt @lwrigh sorry for the delay on this, there is a PR up with the change - https://github.com/patternfly/patternfly/pull/4296 if you'd like to leave your feedback.

Issue #3907, opened 2021-03-03
Introduce Code Block component


mcarrano:
Code Block is a new component for presenting a snippet of read-only code. It is lighter weight than the Code Editor in both appearance and code footprint. It supports optional actions but not editing. 

Design issue. patternfly/patternfly-design#1009

@lwigh can you attach a link to your final design mocks?

lwrigh:
Here is the Marvel link to the code block mocks: https://marvelapp.com/prototype/c2i36cd The first mock features the step by step interaction points that are out of context of the Quick Starts. Let me know if any other additional states need to be documented.

mceledonia:
@christianvogt @mcarrano @lwrigh I do like the float-right solution christian has there. It scales well for large/small whereas the proposed horizontal layout ends up using more space than it needs to in certain circumstances. I don't think we considered this as an option before, it could be a good evolution of the proposal.

mceledonia:
Could we potentially have the layout shift to float-right when there's ample room for it, looking at this from a responsive perspective? So a larger breakpoint could shift to float-right to not add the additional horizontal space when it's not needed.

christianvogt:
While quick starts in a resizable drawer may begin with a smaller available width, it is just the first use case driving this component. We need to consider this could be used by others in main content areas where there is ample space available.

After all this is going to be a generic component.

mcoker:
@lwrigh I have a PR for the component up here. I've gone with what's in the marvel designs for now. https://github.com/patternfly/patternfly/pull/3937

lwrigh:
@mcoker That's looking good to me! Upon hover there should be a tooltip that tells the user what the action does, should that just be noted in the design documentation? I can see that being an optional part of the component depending on what the user needs from the component.

PR #3937, opened 2021-03-19
feat(code-block): added code block component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3907

### DIFF
--- a/src/patternfly/components/CodeBlock/code-block-code.hbs
+++ b/src/patternfly/components/CodeBlock/code-block-code.hbs
@@ -1,0 +1,6 @@
+<code class="pf-c-code-block__code{{#if code-block-code--modifier}} {{code-block-code--modifier}}{{/if}}"
+  {{#if code-block-code--attribute}}
+    {{{code-block-code--attribute}}}
+  {{/if}}>
+  {{~>@partial-block~}}
+</code>
--- a/src/patternfly/components/CodeBlock/code-block-code.hbs
+++ b/src/patternfly/components/CodeBlock/code-block-code.hbs
@@ -2,5 +2,5 @@
   {{#if code-block-code--attribute}}
     {{{code-block-code--attribute}}}
   {{/if}}>
-  {{~>@partial-block~}}
+  {{>@partial-block}}
 </code>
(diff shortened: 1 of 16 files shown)


mcarrano (commented):
Looks good to me @mcoker . You @lwigh?

src/patternfly/components/CodeBlock/code-block-code.hbs
```diff
@@ -0,0 +1,6 @@
+<code class="pf-c-code-block__code{{#if code-block-code--modifier}} {{code-block-code--modifier}}{{/if}}"
+  {{#if code-block-code--attribute}}
+    {{{code-block-code--attribute}}}
+  {{/if}}>
+  {{~>@partial-block~}}
```
mattnolting, line 5: Did you mean to add `~`?

src/patternfly/components/CodeBlock/code-block.scss
```diff
@@ -0,0 +1,37 @@
...
+  --pf-c-code-block__content--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-code-block__pre--FontFamily: var(--pf-global--FontFamily--monospace);
+  --pf-c-code-block__pre--FontSize: var(--pf-global--FontSize--sm);
+
+  background-color: var(--pf-c-code-block--BackgroundColor);
+}
+
+.pf-c-code-block__header {
+  display: flex;
+  width: 100%;
```
mattnolting, line 17: Don't think you need this.
```suggestion
```

mattnolting (commented):
This looks awesome, great work! Just need to add `white-space` controls, and it's GTG!

Screen Shot 2021-03-23 at 2 09 17 PM

mattnolting (approved):
LPTM! 🎉

lwrigh:
The demo looks great to me!

mcarrano:
This looks good @mcoker . My only question is whether the gap between the bottom of the text and the expansion widget needs to be so wide. Is that just a function of the block contents or is that built-in spacing?

mcoker:
@mcarrano That's an artifact of whitespace management in our examples, @redallen is working on a fix for us. This is what the expand button will look like when we can better manage the whitespace.

Screen Shot 2021-03-24 at 2 28 16 PM
Screen Shot 2021-03-24 at 2 30 24 PM

mcarrano (approved):
Thanks for the clarification @mcoker . I approve then!

sahil143:
There should not be a need to put <br /> tag on each line inside <pre/>. It should handle the formatting itself.

christianvogt:
@mcoker I agree with @sahil here. If we're displaying a code block, I don't want to have to replace new lines with a break tag.

redallen:
The reason that the `<br>` tags are there is a workspace bug. #3956 should fix it.

Issue #3874, opened 2021-02-18
validated="error" styles for Select component 


jgiardino:
PF currently does not support a `validated='success' | 'warning' | 'error' | 'default'` on the Select component, whereas it does for the Form Select component, as shown below. Is it possible to add similar styles for the Select component that we have for the Form Select component?

<img width="824" alt="image" src="https://user-images.githubusercontent.com/21063328/108427843-c0377a00-720b-11eb-8c75-c222e5c4e6e8.png">

For my use case, I'm primarily interested in the error styles, but given that these are similar components, maybe all validated states would be useful.

cc @christiemolloy

mcoker:
cc @mcarrano

mcarrano:
@jgiardino I don't see any reason we couldn't support that, do you @mcoker ? But I am curious if there's a reason to not use the Form Select. Does your use case involve data entry into a form? I believe we were expecting that you would use a form select there vs a Select more for things like filters. @mmenestr also wanted to see if making this addition would conflict with anything we say about select usage in our guidelines.

mcoker:
Sounds good to me, seems well suited for the select and potentially the options menu, too.

jgiardino:
Here are more details about our use case:
In our implementation of filters, we have a couple of cases the user could encounter that we need to inform the user about. We have a limit on how many filters can be applied at one time. We also validate on the free-form text input for "Name" which doesn't allow certain characters. I didn't see any examples explaining how to handle validation in the context of the PF filter design guideline. So our approach for now is to display a validation message in a tooltip, and use the validated prop on the input. But we don't have a validated prop on the select component (where we're using the checkbox variant of Select for multiselection). 

Let me know if there's another approach that you'd recommend for this use case, and if that approach is currently supported with the PF components.

mcarrano:
Thanks for providing that additional detail @jgiardino . So I think we should get this on the roadmap then. Is there a particular time from you require this for? And remind me what the project is.

@mmenestr is this something we should also address in the filter guidelines or do you have other guidance regarding these error states? It's true that it's not something we really considered.

jgiardino:
This would be for the app services bundle we're working on for clouddot. We're expecting to have an external eval available in April.

mmenestr:
I think this would be fine to add! Had never considered this before either but I definitely see the value in it for those situations. It might be worth adding a short "validation" section to the filter guidelines to address this, but I would hold off until it's implemented!

jgiardino:
[...]

PR #3940, opened 2021-03-19
feat(select): added invalid state

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3874

### DIFF
--- a/src/patternfly/components/Select/examples/Select.md
+++ b/src/patternfly/components/Select/examples/Select.md
@@ -19,6 +19,13 @@ import './Select.css'
 {{/select}}
 ```
 
+### Invalid
+```hbs
+{{#> select id="select-invalid" select--IsInvalid="true"}}
+  Filter by birth month
+{{/select}}
+```
+
 ### Single expanded
 ```hbs
 {{#> select id="select-single-expanded" select--IsExpanded="true"}}
--- a/src/patternfly/components/Select/select-toggle-status-icon.hbs
+++ b/src/patternfly/components/Select/select-toggle-status-icon.hbs
@@ -1,0 +1,6 @@
+<span class="pf-c-select__toggle-status-icon{{#if select-toggle-status-icon--modifier}} {{select-toggle-status-icon--modifier}}{{/if}}"
+  {{#if select-toggle-status-icon--attribute}}
+    {{{select-toggle-status-icon--attribute}}}
+  {{/if}}>
+  {{>@partial-block}}
+</span>
--- a/src/patternfly/components/Select/select-toggle.hbs
+++ b/src/patternfly/components/Select/select-toggle.hbs
@@ -10,6 +10,9 @@
     {{#if aria-label}}
       aria-label="{{aria-label}}"
     {{/if}}
+    {{#if select--IsInvalid}}
+      aria-invalid="true"
+    {{/if}}
     {{/unless}}
   {{#if select-toggle--attribute}}
     {{{select-toggle--attribute}}}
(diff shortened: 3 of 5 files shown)


jgiardino:
This looks great! Thanks for adding this!

mmenestr:
LGTM - although should there be some explanatory text somewhere too? How would a user know _what_ is wrong exactly when they make a selection? @jgiardino can't remember if this was discussed in the original issue now

jgiardino:
> should there be some explanatory text somewhere too? How would a user know what is wrong exactly when they make a selection?

Looking at all PF examples, the Form component shows the error helper text that you're describing, versus the form input components like form control and form select. None of our current form examples use the Select component, so it's worth asking if we should show an example like that. 

However, in my use case, this is used for filtering, so the normal Form layout doesn't work there. Instead of displaying the error text inline, we display it in a tooltip. Maybe this would be worth showing in the React example for Select? @mcarrano What would you expect here?

mcarrano (commented):
Thanks @mcoker . This looks good to me. But is the user able to choose any of the validation styles to apply here, error, warning, or success?  Is this what you were expecting @jgiardino ?

mcoker:
@mcarrano the valid/warning/error helper text comes from the form component, so a user would place a form select, custom select, text input, textarea, etc in a form, and upon either form or field validation, the form component shows the error/helper text. Similarly, there are no uses of the form helper text in the [form control component examples](https://www.patternfly.org/v4/components/form-control), either. But you do see it on the [form component examples](https://www.patternfly.org/v4/components/form#invalid)

I believe there are a couple of components that have that same helper text built in (date picker is one), so it seems like a valid use case to allow/move the helper text to be used outside of the context of the form. I believe we discussed this at some point and talked about maybe making helper text an inline text variation to the alert component, if I recall?

mcarrano:
Interesting question about application of the helper text. I do see it applied in the React example for Form Select here: https://www.patternfly.org/v4/components/form-select#validated  I guess that I didn't realize use of the helper text was restricted to only when components are used inside of a form. I don't think users would make that distinction. That said, I'm not sure if we need to add it for the more general case. 

@jgiardino can you say a bit more about your use case? Would you want to have helper text below the field here instead of relying on a tooltp? I will say that it's a bit unexpected to get error states returned by a select at all since it implies that we are giving users invalid options to select from. Under what circumstances do we do that?

jgiardino:
> can you say a bit more about your use case? Would you want to have helper text below the field here instead of relying on a tooltp? I will say that it's a bit unexpected to get error states returned by a select at all since it implies that we are giving users invalid options to select from. Under what circumstances do we do that?

We're using it in a filter. We have a maximum number of filters that can be defined at one time, and so the error case happens when the user has reached this max. Although with @mcarrano's point above, maybe error isn't the right variant for this, and a warning is more appropriate?:
> it's a bit unexpected to get error states returned by a select at all since it implies that we are giving users invalid options to select from.

Here are examples of what we have now. Ideally, the tooltip would display for the toggle, but we haven't made that update yet. In the case of using Select to specify a filter, we keep selected options as enabled, and disable the non-selected options. This let's a user clear a filter from the menu but not select additional options. 

image

image

For comparison, here is a filter option that is a text input. This is behavior that we'd like to duplicate for the Select component.

image

mcarrano:
Thanks for that additional detail @jgiardino . Given what you are showing here, not sure if adding helper text below the field is the right approach. What do you think @mmenestr ?

mmenestr:
I think as long as there is some sort of explanation the user can have access to (in this case a tooltip) then this should be good to go!

mattnolting (approved):
Perfect!

Issue #3941, opened 2021-03-22
Remove use of pf-m-expandable from form field groups


mcoker:
This class was removed from the CSS but is still in the examples.

PR #3942, opened 2021-03-22
chore(form): removed used of pf-m-expandable from field groups

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3941


mattnolting (approved):
LPTM!

Issue #3944, opened 2021-03-22
Check/radio body and description elements are invalid HTML when a child of `<label>`


mcoker:
When the parent `.pf-c-[check/radio]` elements are a `<label>`, we need the children to be phrasing content so that the HTML is valid. Since those elements are grid children, they'll have block formatting without any additional changes as a `<span>`.

PR #3945, opened 2021-03-22
chore(check/radio): updated body/description elements to be spans

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3944


mattnolting (approved):
LPTM! 🌮

PR #3956, opened 2021-03-26
fix(workspace): fix whitespace and code block

redallen:
As promised @mcoker . Hope you enjoyed your PTO!


mcoker (approved):
Love it!! The [expandable code block examples](https://patternfly-pr-3956.surge.sh/components/code-block#expandable) have too much space in them, but I'm fine with it. Looks like everything works, we'll just need to modify those examples.

Issue #3957, opened 2021-03-29
Update icon implementation in Icons page


mmenestr:
It's been brought to my attention that the icons size guidance in PF does not match the icon sizes in our sketch kit. 
Sketch kit: 
![image](https://user-images.githubusercontent.com/51165119/112858922-58811400-9080-11eb-9247-26aef6877eda.png)

PF icons: 
![image](https://user-images.githubusercontent.com/51165119/112858939-5e76f500-9080-11eb-866f-aa949ae1a4a6.png)

These two issues give a bit more background:
1.  https://github.com/patternfly/patternfly-design-kit/issues/162
2. https://github.com/patternfly/patternfly-design/issues/787

It seems that the sketch kit sizing is the correct one (12px, 16px, 24px), @mcoker do you have any suggestions as to what we should do here? Would there be a potential problem with updating the icon sizing in the PF guidelines?

cc: @mcarrano @doruskova

mcoker:
@mmenestr The problem with updating the icon size recommendation in the guidelines is that the sizes reference our global variables, so we would ideally update the guidelines by updating our global vars. And I wonder if changing the global var values to match the sketch kit would be a breaking change. Changing the sizes could cause alignment/layout issues.

Here is a list of everywhere we use the global icon size variables. 

```
Dropdown/dropdown.scss:  --pf-c-dropdown__menu-item-icon--Width: var(--pf-global--icon--FontSize--lg);
Dropdown/dropdown.scss:  --pf-c-dropdown__menu-item-icon--Height: var(--pf-global--icon--FontSize--lg);
OptionsMenu/options-menu.scss:  --pf-c-options-menu__menu-item-icon--FontSize: var(--pf-global--icon--FontSize--sm);
Alert/alert.scss:  --pf-c-alert__icon--FontSize: var(--pf-global--icon--FontSize--md);
Chip/chip.scss:  --pf-c-chip--m-draggable__icon--FontSize: var(--pf-global--icon--FontSize--sm);
Tile/tile.scss:  --pf-c-tile__icon--FontSize: var(--pf-global--icon--FontSize--md);
Tile/tile.scss:  --pf-c-tile__header--m-stacked__icon--FontSize: var(--pf-global--icon--FontSize--lg);
Tile/tile.scss:  --pf-c-tile--m-display-lg__header--m-stacked__icon--FontSize: var(--pf-global--icon--FontSize--xl);
[...]

PR #3964, opened 2021-03-31
fix(global): updated sm/md icon size global vars

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3957

### DIFF
--- a/src/patternfly/sass-utilities/scss-variables.scss
+++ b/src/patternfly/sass-utilities/scss-variables.scss
@@ -182,8 +182,8 @@ $pf-global--BorderRadius--lg:        30em !default; // This is a sufficiently la
 // Icons
 $pf-global--icon--Color--light: $pf-color-black-600 !default;
 $pf-global--icon--Color--dark: $pf-color-black-900 !default;
-$pf-global--icon--FontSize--sm: pf-font-prem(10px) !default;
-$pf-global--icon--FontSize--md: pf-font-prem(18px) !default;
+$pf-global--icon--FontSize--sm: pf-font-prem(12px) !default;
+$pf-global--icon--FontSize--md: pf-font-prem(16px) !default;
 $pf-global--icon--FontSize--lg: pf-font-prem(24px) !default;
 $pf-global--icon--FontSize--xl: pf-font-prem(54px) !default;
 


gdoyle1:
@mcoker did this update the icons in the demos and components too? Wondering if I can tell the difference compared to what is implemented today

mcoker:
@gdoyle1 I updated the list of elements in our components that would be impacted by changing the small and medium icon sizes in this comment - https://github.com/patternfly/patternfly/issues/3957#issuecomment-810604433

Any of those elements/icons will get the size change when they're used in component examples, demos, etc. I can provide more details on where exactly those icons are in the components if you'd like. You can compare the preview URL in this PR with what's on https://pf4.patternfly.org/ to see how the change compares to what's in the component today.

gdoyle1:
Thanks @mcoker - I just went through them all. The only ones I noticed a _major_ difference in were the icons for the select menus and then the spinner sizes.. but I think the changes are fine! I'd like to get @mceledonia 's review as well when he's back from PTO before merging this.

gdoyle1 (approved):

mcoker:
@mcarrano @mceledonia do you have any feedback? Curious if we want to consider making this change now, or queue it up for a breaking change release?

mceledonia:
I think we might as well update these now, they are smaller so shouldn't be visually breaking. @mcarrano @mcoker

mcoker:
> they are smaller so shouldn't be visually breaking

The small variable increases from 10px to 12px

mcarrano:
I'd like to hold off on making this change. Looks to be low risk, but I'm wary to introduce some type of unexpected spacing or alignment effect in product. @mcoker I leave it to you as to how we want to handle this PR, but we should not merge.

I've opened https://github.com/patternfly/patternfly-design-kit/issues/371 to avoid the documentation issue by only refering to icons by size and not pixel dimensions letting these be driven by the CSS variables.

mcoker:
> I'd like to hold off on making this change. Looks to be low risk, but I'm wary to introduce some type of unexpected spacing or alignment effect in product.

@mcarrano added the breaking change label to the original issue and will close this PR. We can redo it when we're ready to make the update.

Issue #3953, opened 2021-03-25
Add View more action to Select menus


mcarrano:
View more option should be added at the bottom of menus in the Select and Menu components.

Final design: https://marvelapp.com/prototype/c49ebi3/screen/77975699

PR #3968, opened 2021-03-31
feat(select): added view more and loading support to menu

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3953

@mcarrano @mmenestr  are we just adding the "view more" item and loading spinner, or the menu footer, too? 

----

This PR also aims to make sure that the list of items has top/bottom padding. A good example of this is an update made in the context selector to add this padding to the top and bottom of the list between the search filter and footer.

Screen Shot 2021-04-05 at 7 19 50 PM
Screen Shot 2021-04-05 at 7 19 46 PM

That update has been made in the select and menu components, too.


mcarrano (changes requested):
@mcoker When the "View more" action is added at the bottom of the menu, I don't think it should take the same hover styling as regular menu items but just be treated like a link button. Do you agree @mmenestr ?

As for the footer, yes I think we wanted both options. The footer should be exactly the same as we use here: https://www.patternfly.org/v4/components/context-selector/html#with-footer The difference would be that the footer is sticky where the View more item would scroll with the menu. Does that make sense?

While the immediate need is to have these features in Select, I also think they should be added to the composable Menu component for consistency.

mmenestr:
Agreed, I would expect it to behave like a regular link button - but still would expect it to be aligned/placed the way it is currently!

mcoker:
@mcarrano @mmenestr updated based on y'alls feedback. Only outstanding question is the text alignment on the footer.

mcoker:
After chatting with @gdoyle1 and @mmenestr, we left aligned the footer and updated the example to include an inline link button.

gdoyle1:
LGTM @mcoker - will we open another issue to update the button alignment in the context selector as well?

mcoker:
@gdoyle1 yep! Added an issue here, feel free to edit/comment as you see fit -https://github.com/patternfly/patternfly/issues/3971

mcarrano:
This looks great @mcoker . Can I just ask for one small change? Can change the footer action in the Footer example from "View more" to just "Action?" I think that will better convey the idea that you can place any action in the footer. I think that using the action inline with the menu will be the preferred way to implement a View more behavior.

Also, how do you want to replicate these changes to Menu? Add to this PR or open a separate PR?

mmenestr:
Looks great to me too, thanks @mcoker !

mcoker:
> Can I just ask for one small change? Can change the footer action in the Footer example from "View more" to just "Action?" I think that will better convey the idea that you can place any action in the footer. I think that using the action inline with the menu will be the preferred way to implement a View more behavior.

@mcarrano yep, updated!

Also re: the inline item, currently this is a "load" variation of a select item (paired with the "loading" variation with the spinner). Would you ever anticipate using this item to perform some other action? If so I'm wondering if I should call the variation something more generic.

Here's an example, though I don't suspect we would use the inline item to add something new, that seems more appropriate for a footer action.

Screen Shot 2021-04-05 at 10 27 44 AM

>  how do you want to replicate these changes to Menu?

I'll add it to this PR now since the changes are mostly approved.

mcarrano:
> Also re: the inline item, currently this is a "load" variation of a select item (paired with the "loading" variation with the spinner). Would you ever anticipate using this item to perform some other action? If so I'm wondering if I should call the variation something more generic.

I don't think so. My take is that any other action (like 'Add new item') should be placed in the sticky footer and not inline with the menu. Do you agree @mmenestr ?

mcoker:
@mcarrano @mmenestr @mceledonia updated the menu component. Also made sure there is space above/below the list of options in the select, menu, and context selector item lists, and that space remains when there is a search filter at the top, or a footer at the bottom of the list. That space is best seen in the [context selector in this PR](https://patternfly-pr-3968.surge.sh/components/context-selector)

Screen Shot 2021-04-05 at 7 19 46 PM

Screen Shot 2021-04-05 at 7 19 50 PM

You can see in the [pf.org context selector examples](https://www.patternfly.org/v4/components/context-selector/html) that this spacing around the list of items doesn't exist.

Screen Shot 2021-04-05 at 7 29 40 PM

Screen Shot 2021-04-05 at 7 29 36 PM

mcarrano (approved):
Looks great @mcoker

src/patternfly/components/Menu/menu.scss
```diff
@@ -121,10 +124,16 @@
   // Drilled in
   --pf-c-menu--m-drilled-in--c-menu__list-item--m-current-path--c-menu--ZIndex: var(--pf-global--ZIndex--2xl);
 
+  // Footer
+  --pf-c-menu__footer--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-menu__footer--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-menu__footer--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-menu__footer--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-menu__footer--BoxShadow: var(--pf-global--BoxShadow--sm-top);
+  --pf-c-menu__footer--MarginBottom: calc(var(--pf-global--spacer--sm) * -1);
```
mattnolting, line 133: Looks like this is causing a scrollbar to appear

Screen Shot 2021-04-14 at 9 26 25 AM

Don't think you need it if you're removing padding from `pf-c-menu`

mattnolting (commented):
This is looking great. Just seeing the scrollbar issue.

mmenestr:
It also looks like there's something off about the footer spacing? The spacing above and below the link button should be equal.
When I inspect it looks like the bottom spacing falls off of the footer - not sure how to explain it haha

image

src/patternfly/components/Menu/menu.scss
mcoker: @mattnolting updated!

mattnolting (approved):
LPTM!

mcoker:
@mmenestr thanks for catching that, the preview is updated if you want to have another look!

mmenestr:
Looks good!

Issue #3971, opened 2021-04-01
Left align context selector footer action(s)


mcoker:
After discussing with @gdoyle1 and @mmenestr, we want to left align the content in the context selector footer, and recommend that if someone places a link variation of the button in the footer, they use the inline link variation.

@mcarrano do you think we should add this as a new variation so that it isn't a breaking change, then update the default alignment in a future breaking change release?

mcarrano:
Hmmm... That's a really good question. Yeah, I don't want to fit people with an unexpected change but also think a new variation feels like overkill. Maybe we would add an `alignment` prop, but keep the default right-aligned in the Context Selector only? What do you think @gdoyle1 @mmenestr ?

gdoyle1:
I'd agree to not create a new variation since it's just the button alignment... I also wonder how many people are using the footer with action button in the context selector? @mcarrano @mcoker the reason I bring that up is just to see how much of a breaking change it would be for users??

mcoker:
I mention the variation because we did that with another component (modal, I think?) when its actions were right aligned and we wanted to make them left aligned. The context selector is likely nowhere near as commonly used as the modal. @redallen are you able to see how many uses there are of the `<ContextSelector>`'s `footer` prop?

redallen:
I don't see any apps that directly use `import { ContextSelector } from '@patternfly/react-core';`. I only found one small [shared library](https://github.com/rh-uxd/appservices-patternfly/tree/master/packages/appservices-patternfly-crossconsole/src/components/CrossNavContextSelector) by @mfrances17 and @dlabaj that seems to reimplement patternfly-react's `<ContextSelector>` but it doesn't use `footer` at all.

mcoker:
@mcarrano @gdoyle1 sounds like it's probably safe to update, though it may still be used somewhere else - depends on how much of a breaking change we consider the layout change to be if it is in use.

mcarrano:
I think it is a very minor change and most people won't care. It's certainly not as controversial as moving the modal buttons. Do you agree @gdoyle1 ?

gdoyle1:
Yes I agree @mcarrano @mcoker

PR #3978, opened 2021-04-07
fix(context-selector): left aligned footer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3971


mcarrano (approved):
This looks great @mcoker . Will this require any React update or is this strictly a CSS level change?

mcoker:
@mcarrano the alignment comes for free, but the action in the react workspace example is a link button and that should be updated to be an inline link - https://patternfly-react.surge.sh/components/context-selector#with-footer

mattnolting (approved):
Perfect!

Issue #3980, opened 2021-04-07
Expandable card - toggle icon not rotating


mcoker:
Looks like it was accidentally removed in https://github.com/patternfly/patternfly/pull/3914/files#diff-f77b58a7cf3d9657d084dec7fcc8ac27020c93e9dedea7d718cc122ccf48cd83L40

PR #3981, opened 2021-04-07
fix(card): fixed expandable toggle rotation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3980


mattnolting (approved):
LPTM! 🍔

Issue #3983, opened 2021-04-08
Bug: mobile view of filter panel


jgiardino:
In the primary-detail example — https://www.patternfly.org/v4/demos/primary-detail/react-demos/primary-detail-full-page/ — a css issue exists when the viewport is resized to show the filter icon that displays a panel with filter options, where the data list contents behind the filter panel become visible when hovering the data list item.

Steps to reproduce:
1. Navigate to https://www.patternfly.org/v4/demos/primary-detail/react-demos/primary-detail-full-page/
2. Resize the window until the fillter toolbar components are replaced with a filter icon
3. Click the icon to display the filter components
4. Move the mouse cursor to hover over the first data list item

Result: The filter panel is still open, but the data list item displays on top of the panel on hover. 
![image](https://user-images.githubusercontent.com/21063328/113951529-b67ccc80-97e1-11eb-9556-f2f54fd25c98.png)

Expected behavior: the filter panel should remain visible until I click to dismiss the panel.
![image](https://user-images.githubusercontent.com/21063328/113951501-ad8bfb00-97e1-11eb-8e28-2b8240fb96b4.png)

mcoker:
@jgiardino nice catch! The expanded content should be using a higher z-index, `--pf-global--ZIndex--sm`, which is what we use for menus that pop out from a toggle.

PR #3986, opened 2021-04-08
fix(toolbar): changed expand content zindex to sm to match other menus

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3983


mattnolting (approved):
LPTM! 🌮

jgiardino:
Thanks for fixing this so fast! I'm not really sure how to test this without being able to show the filter menu on top of a data view. I couldn't find any html demos with this exact display state. But I have faith that it works as expected.

mcoker:
@jgiardino you can inspect the toolbar component in the demo you linked to in the [original issue](https://github.com/patternfly/patternfly/issues/3983), and find the declaration `--pf-c-toolbar__expandable-content--ZIndex: var(--pf-global--ZIndex--xs);` in the var list, and change the value from `--pf-global--ZIndex--xs` to `--pf-global--ZIndex--sm`. Then do the hover test you described to reproduce the problem.

Or you can also assign `--pf-c-toolbar__expandable-content--ZIndex: var(--pf-global--ZIndex--sm);` as an inline var on the `.pf-c-toolbar` element.

Issue #3987, opened 2021-04-08
Select typeahead text input height should match form control elements


mcoker:
The form control in the select typeahead toggle is shorter than a regular form control text input since its top/bottom borders have been removed. The time picker uses a select typeahead with the toggle button on the right removed, making the typeahead toggle defined by the form control element, making it shorter than other form controls, dropdown toggles, etc.

PR #3988, opened 2021-04-08
fix(select): updated typeahead text input height to match form control

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3987

this will allow us to remove the toggle button in a typeahead select and the select toggle height will remain the same height (36px)


mattnolting (approved):
LGTM! 👍

Issue #790, opened 2018-10-05
Universal Focus State


mceledonia:
Designing a focus state pattern which works across all components. The focus state visual design should be independent of the component's other states such as hover or active.

[Focus State PF4.png.zip](https://github.com/patternfly/patternfly-next/files/2483817/Focus.State.PF4.png.zip)

LHinson:
Just realized there is issue #786 @jgiardino states: 

> One of the WCAG items we need to meet is to clearly indicate focus ([WCAG 2.4.7](https://www.w3.org/WAI/WCAG21/quickref/#focus-visible)).
> 
> Our original approach has been to rely on the browser defaults for indicating focus. But currently there are a couple of issues with this:
> 
> - Our current styles override the default styles provided by Firefox for some elements, like buttons
> - Some components might not be able to rely on browser default styles. For example, in a Switch component that's based on a native html checkbox, we're hiding the checkbox which is the element that receives focus, which means the focus is not visible to the user.
> 
> 
> P.S. WCAG link above includes links to additional information about how to meet this criteria.

mceledonia:
Universal focus state visual design

[Focus State PF4.png.zip](https://github.com/patternfly/patternfly-next/files/2483817/Focus.State.PF4.png.zip)

guillaumevincent:
> in the design I shared, border is `#007BBA` and the background is also `#007BBA` at 15% transparency

from @mceledonia on slack

andybraren:
Just dropping a note here that whenever this gets implemented it should be tested with the Tabs component. Discussion about this starts [here](https://github.com/patternfly/patternfly-next/pull/868#issuecomment-436062717).

mceledonia:
Giving this a bump @rachael-phillips @mcoker design work here should be done, might need to have some discussion around how this behaves with certain components such as tabs and input groups, but overall I think we are set to implement.

mceledonia:
Want to note here the last discussion that took place on this topic:

We would like to shift directions here, going with browser default focus states when applicable, and our custom focus state design only when the browser default state does not work. We reached this conclusion after a better understanding that keyboard users are far more accustomed to and expect browser default focus states. As this is an accessibility focused feature, we want to accommodate that use case as best as we can by providing an expected experience.

mceledonia:
bump @rachael-phillips @jgiardino @mcarrano @mcoker @LHinson

jgiardino:
Thanks for recapping that here, @mceledonia!

[...]

PR #3991, opened 2021-04-12
fix(input-group): added z-index

### LINKED ISSUE #3926: Input group focus rings are truncated
Input groups join widgets right next to each other. In the default state, that's fine. However, when one of the widgets in the group gets focus, the ring is chopped off for everything but the last child.

Screenshots from https://www.patternfly.org/v4/components/input-group (with focus):

Screenshot from 2021-03-15 13-13-14

Screenshot from 2021-03-15 13-12-53

Screenshot from 2021-03-15 13-12-58

Screenshot from 2021-03-15 13-13-01

mattnolting:
closes #3926 

Added `z-index` to `:focus`

### DIFF
--- a/src/patternfly/components/InputGroup/input-group.scss
+++ b/src/patternfly/components/InputGroup/input-group.scss
@@ -33,7 +33,8 @@
   }
   // stylelint-enable
 
-  > :focus {
+  > :focus,
+  > :focus-within {
     z-index: var(--pf-c-input-group--child--ZIndex);
   }
 
--- a/src/patternfly/components/InputGroup/input-group.scss
+++ b/src/patternfly/components/InputGroup/input-group.scss
@@ -1,6 +1,7 @@
 .pf-c-input-group {
   // Input group
   --pf-c-input-group--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-input-group--child--ZIndex: var(--pf-global--ZIndex--xs);
 
   // Input group text
   --pf-c-input-group__text--FontSize: var(--pf-global--FontSize--md);
@@ -30,8 +31,11 @@
   > * + * {
     margin-left: -1px;
   }
-
   // stylelint-enable
+
+  > :focus {
+    z-index: var(--pf-c-input-group--child--ZIndex);
+  }
 
   .pf-c-form-control {
     &[aria-invalid="true"] {


garrett:
This isn't fixed here. I just tested the preview page for input groups @ https://patternfly-pr-3991.surge.sh/components/input-group and got the following:

Screenshot from 2021-04-12 16-11-11
Screenshot from 2021-04-12 16-11-08
Screenshot from 2021-04-12 16-11-04
Screenshot from 2021-04-12 16-11-01
Screenshot from 2021-04-12 16-10-51
... additional groups are also affected, but the above demonstrates it enough.

Note the truncation of the focus ring on one or both sides, depending on the group's neighboring widgets.

garrett (changes requested):
Thanks for the effort, but this isn't fixed.

I just tested the preview page for input groups @ https://patternfly-pr-3991.surge.sh/components/input-group and got the following:

Screenshot from 2021-04-12 16-11-11
Screenshot from 2021-04-12 16-11-08
Screenshot from 2021-04-12 16-11-04
Screenshot from 2021-04-12 16-11-01
Screenshot from 2021-04-12 16-10-51
... additional groups are also affected, but the above demonstrates it enough.

Note the truncation of the focus ring on one or both sides, depending on the group's neighboring widgets.

mattnolting:
@mceledonia @mcarrano Hey guys, talked to @mcoker and we agreed that the safest approach to solving the focus ring issue is to define it explicitly (overwrite default browser styles) then apply a negative offset so that it always appears inside the inputs. Can you provide a focus style?

mceledonia:
@mattnolting Check out #1824 for our last focus style, that will be a good place to start in terms of a design.

jessiehuff (approved):
LGTM! I think your approach here makes sense, especially given your point that the default styling across browsers can be inconsistent.

mcoker (changes requested):
We discussed and decided using the xs z-index is likely a more sustainable/manageable fix.

mattnolting:
@mcoker @garrett updated

mcoker (approved):
L🔥TM!

garrett (approved):
Looks great now! :+1: The focus ring is always above the other elements, not hidden/truncated as before.

Tested in Firefox stable, Firefox nightly (which has different default widgets now), Chrome stable, and WebKit. The focus ring is shown as it should be. (Although it's styled differently everywhere, but that's another issue. :wink:)

Thanks!

mattnolting:
@garrett Yw! 👍  We'll work towards universal focus states ([here if your interested](https://github.com/patternfly/patternfly/issues/790))

PR #3993, opened 2021-04-12
fix(toolbar): added explicit text-overflow

### LINKED ISSUE #3449: Toolbar filters are narrower in Chrome
Toolbar filters are narrower in Chrome than either Safari or Firefox. This becomes obvious when using long placeholder text, which forces an ellipsis. I was able to replicate the issue here: https://www.patternfly.org/v4/documentation/core/components/toolbar#toggle-group , using placeholder="Filter by exact username" 

Here's an example from the RBAC UI:
Screen Shot 2020-09-03 at 2 08 04 PM

mattnolting:
closes #3449 

It appears to solve the `overflow ellipsis` setting `text-overflow: ellipsis;` to `::placeholder && [placeholder]` works. WDYT??

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -201,6 +201,11 @@ $pf-c-form-control--m-clock--Coordinates: "M256 8C119 8 8 119 8 256s111 248 248 
     color: var(--pf-c-form-control--placeholder--Color);
   }
 
+  &::placeholder,
+  &[placeholder] {
+    text-overflow: ellipsis;
+  }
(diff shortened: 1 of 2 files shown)


epwinchell:
Hi @mattnolting Here's what I'm seeing in Safari, Chrome and FF:

Screen Shot 2021-04-12 at 4 38 39 PM

mattnolting:
@epwinchell I believe this issue will be resolved with this PR and [3994](https://github.com/patternfly/patternfly/pull/3994)

epwinchell (approved):

Issue #3996, opened 2021-04-12
Button - revert variations to define regular properties instead of custom properties


mcoker:
In this PR to add the secondary/link danger button variations, we switched button variations from defining `color` and `background-color` properties to defining custom properties instead. This can cause issues if a button variation is overriding existing CSS for a button that defines regular properties - our button variations will need to define regular properties in order to properly override that existing CSS.

PR #3997, opened 2021-04-12
fix(button): reverted variations to regular property declarations

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3996

From https://github.com/patternfly/patternfly/pull/3932/files

* changes the variations back to defining `color` and `background-color`
* removes the initial `--pf-c-button--Color/BackgroundColor` vars that were added in #3932
* moves the danger variation to come after link
* nests the danger variations together

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -194,11 +194,9 @@
   font-size: var(--pf-c-button--FontSize);
   font-weight: var(--pf-c-button--FontWeight);
   line-height: var(--pf-c-button--LineHeight);
-  color: var(--pf-c-button--Color, inherit);
   text-align: center;
   white-space: nowrap;
   user-select: none;
-  background-color: var(--pf-c-button--BackgroundColor);
   border: 0;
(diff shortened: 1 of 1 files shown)


evwilkin:
@mcoker is there any chance this could break people the other way, if users are overriding the custom properties?

mattnolting (approved):
Nice! LGTM 🔥

mcoker:
> @mcoker is there any chance this could break people the other way, if users are overriding the custom properties?

The only concern I have is that in the last release, I added `--pf-c-button--Color` and `--pf-c-button--BackgroundColor`, which are set to default values. Those vars were basically placeholders on our side just to be overridden by other variations. 
 This PR removes them. So if between the last release and this release, a user has opted in to defining them, they would no longer work. However, if a user has defined them, the definition shouldn't do anything as all buttons require to be used with a variation, and each variation defines `color` and `background-color`, so it seems safe to remove them as part of putting the button properties back to the way they were before the PR that caused these issues.

jcaianirh (approved):

PR #4011, opened 2021-04-16
feat(page): added center mod

### LINKED ISSUE #3984: Enhancement: add center modifier to page__section
Some layouts, that don't have a sidebar, need to be aligned center. 

Example: 
Screen Shot 2021-04-08 at 8 15 01 AM

mattnolting:
closes #3984


src/patternfly/components/Page/page.scss
```diff
@@ -411,6 +411,12 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
       flex: 1;
       max-width: var(--pf-c-page--section--m-limit-width--MaxWidth);
     }
+
+    &.pf-m-center {
+      > .pf-c-page__main-body {
+        margin: 0 auto;
```
mcoker, line 417: See any reason it couldn't just be `margin: auto;`?

```suggestion
        margin: auto;
```

mattnolting: Nope. Updated.

src/patternfly/components/Page/page.scss
```diff
@@ -411,6 +411,12 @@ $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl",
       flex: 1;
       max-width: var(--pf-c-page--section--m-limit-width--MaxWidth);
     }
+
+    &.pf-m-center {
+      > .pf-c-page__main-body {
+        margin: auto;
```
mcoker, line 417: Also is there a reason we wouldn't use align-items here since the parent is a flex container? And wdyt about calling the variation `pf-m-align-[dir]` similar to the dropdown/menu alignment? Would probably be more consistent.

src/patternfly/components/Page/examples/Page.md
```diff
@@ -175,6 +175,33 @@ wrapperTag: div
...
+      {{#> page-header-brand-link page-header-brand-link--href="#"}}
+        Logo
+      {{/page-header-brand-link}}
+    {{/page-header-brand}}
+    {{#> page-header-tools}}
+      header-tools
+    {{/page-header-tools}}
+  {{/page-header}}
+  {{#> page-main}}
+    {{#> page-main-section page-main-section--modifier="pf-m-align-center"}}
```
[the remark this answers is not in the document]
mattnolting: Center alignment without limited width

src/patternfly/demos/Page/examples/Page.md
```diff
@@ -36,5 +36,123 @@ wrapperTag: div
...
+        {{#> card-title}}
+          <h2>Card title</h2>
+        {{/card-title}}
+        {{#> card-body}}
+          <p>Page section width limited, filled, centered. Full width card.</p>
+        {{/card-body}}
+      {{/card}}
+    {{/page-main-section}}
+    {{> divider}}
+    {{#> page-main-section page-main-section--modifier="pf-m-align-center"}}
```
mcoker, line 74: does this do anything since it's not width-limited?

```suggestion
    {{#> page-main-section}}
```

mcoker (changes requested):

mattnolting:
@mcoker wdyt about moving `display: flex; flex-direction: column` to the parent sections? This would allow `pf-m-align-center` to always center a non-width-limited section. This would work really well with w/`pf-c-width-limiter` too. 

```
.pf-c-page__main-nav,
.pf-c-page__main-breadcrumb,
.pf-c-page__main-tabs,
.pf-c-page__main-section,
.pf-c-page__main-wizard {
  display: flex;
  flex-direction: column;

  &.pf-m-limit-width {
    padding: 0;

    > .pf-c-page__main-body {
      flex: 1;
      max-width: var(--pf-c-page--section--m-limit-width--MaxWidth);
    }
  }
} 
```

mcoker:
@mattnolting I'm hesitant to do that because of the way users might be using the `__main` and `__main-section` elements now, so introduced it only with the `pf-m-limit-width` class as an opt-in. I think long term that makes sense, I just worry it will break users introducing it now.

What use case does the basic centering via `&.pf-m-limit-width { align-items: center; }` not fulfill?

src/patternfly/demos/Page/examples/Page.md
mcoker: Just following up on this comment. Otherwise it looks good to me.

src/patternfly/demos/Page/examples/Page.md
```diff
@@ -36,5 +36,123 @@ wrapperTag: div
...
+        {{#> card-body}}
+          <p>Page section width limited, filled, centered. Full width card.</p>
+        {{/card-body}}
+      {{/card}}
+    {{/page-main-section}}
+    {{> divider}}
+    {{#> page-main-section}}
+      {{#> grid grid--modifier="pf-m-gutter pf-m-all-6-col-on-md pf-m-all-3-col-on-xl"}}
+        {{#> grid-item grid-item--modifier="pf-m-12-col"}}
+          <h2 class="pf-u-text-align-center">(Centered page section)</h2>
```
mcoker, line 77: This section isn't centered - should we center it or modify this title?

mcoker (approved):
L🍕TM!!

Issue #4014, opened 2021-04-16
Tree table rows should maintain base indentation when not expandable


mcoker:
The last row here should maintain its indentation relative to other expandable rows.

<img width="853" alt="Screen Shot 2021-04-16 at 10 03 45 AM" src="https://user-images.githubusercontent.com/35148959/115052595-bd6fa300-9ea3-11eb-9ee1-03af25cd5995.png">

PR #4015, opened 2021-04-16
fix(table): fixed indendation for non-expandable tree table rows

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4014


mattnolting (approved):
WFM 👍

Issue #4013, opened 2021-04-16
Divider disappears in certain flex conditions


srambach:
Under certain conditions (inside a column flex container) the divider can disappear. Adding `flex-shrink:0` to the `<hr>`prevents it from shrinking to nothing. This should be added to the divider component by default to prevent the problem.

PR #4016, opened 2021-04-16
fix(divider): updated to not shrink in flex layouts

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4013


srambach (approved):
Looks great! Thanks!

PR #4029, opened 2021-04-28
feat(log-viewer): added log viewer

### LINKED ISSUE #3811: Log viewer component
Provide toolbar layout and core support for the React log viewer POC/component.

Designs here: https://marvelapp.com/prototype/bia3j2d

React issue: https://github.com/patternfly/patternfly-react/issues/5341

mattnolting:
closes #3811

Preview: https://patternfly-pr-4029.surge.sh/components/log-viewer
Designs here: https://marvelapp.com/prototype/bia3j2d
React issue: https://github.com/patternfly/patternfly-react/issues/5341

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown-toggle.hbs
+++ b/src/patternfly/components/Dropdown/dropdown-toggle.hbs
@@ -10,6 +10,10 @@
 
   {{#if dropdown-toggle--IsPlain}}
     aria-label="Actions"
+  {{/if}}
+
+  {{#if dropdown-toggle--attribute}}
+    {{{dropdown-toggle--attribute}}}
   {{/if}}>
 
   {{#if dropdown-toggle--IsPlain}}
--- a/src/patternfly/components/LogViewer/examples/LogViewer.css
+++ b/src/patternfly/components/LogViewer/examples/LogViewer.css
@@ -1,0 +1,25 @@
+#ws-core-c-log-viewer-responsive {
+  height: 400px;
+}
+
+#ws-core-c-log-viewer-responsive .ws-preview-html {
+  height: 100%;
+}
+
+#ws-core-c-log-viewer-with-popover-open .ws-preview-html {
+  position: relative;
+}
+
+#ws-core-c-log-viewer-with-popover-open .pf-c-popover {
+  display: none;
+  position: absolute;
+  z-index: 9999;
+  top: 52px;
+  right: 78px;
+}
+
+@media screen and (min-width: 992px) {
+  #ws-core-c-log-viewer-with-popover-open .pf-c-popover {
+    display: block;
+  }
+}
--- a/src/patternfly/components/LogViewer/examples/LogViewer.md
+++ b/src/patternfly/components/LogViewer/examples/LogViewer.md
@@ -1,0 +1,133 @@
+---
+id: 'Log viewer'
+beta: true
+section: components
+cssPrefix: pf-c-log-viewer
+---
+
+import './LogViewer.css';
+
+## Examples
+
+### Basic
+
+```hbs
+{{#> log-viewer log-viewer--id="log-viewer-basic-example" log-viewer--attribute='style="--pf-c-log-viewer__index--Width: 75px"'}}
+  {{> __log-viewer-toolbar menu--IsHidden="true"}}
+  {{> __log-viewer-main}}
+{{/log-viewer}}
+```
+
+### With line numbers
+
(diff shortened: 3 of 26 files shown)


mattnolting:
@arivepr Updated

nicolethoen (approved):
it definitely looks beautiful! 
I'm happy to work with you and Alexander to incorporate these.

mcarrano (commented):
This is looking good @mattnolting . Just a couple of questions...

* I wasn't exactly sure about this example

Screen Shot 2021-04-29 at 11 28 32 AM

Does this represent what the user would see as the result of a find? If so, can we highlight matching strings vs whole rows? Also, wasn't sure where the lime green color comes from.

* For the responsive layout, the way elements collapse looks good. Should we have an example with the search field opened?

mcarrano:
@mattnolting after consulting with @mceledonia we decided to use the following colors to represent "found" items. All rows that match are highlighted using `Gold-50` and the current selected row is highlighted with `Gold-100`. Here's what that should look like:

Logviewer-find

nicolethoen:
@mcarrano What is the use case for selecting a logger row? I can understand hover styles being used to help make reading the log viewer a little easier, and the highlighting of found search results, but what is a 'selected row' representing?

mcarrano:
@nicolethoen In the find feature, you may have hits on multiple rows. But selected on mean the one you are currently focuses on. The up/down arrows in the Search input let you move that focus. "Selected" might not be the best term, but not sure what to call that.

kmcfaul:
Maybe `matched row` would make sense for that modifier then?

Everything looks good to me, just looks like the highlighted colors need to be updated to the golds.

jessiehuff (commented):
This looks pretty awesome, @mattnolting!!

I think the structure looks pretty good, and I like the approach. The only aspect that I'm not positive I understand the exact interaction yet is the search feature and found/selected results. Is it going to behave more like the Typeahead Select where using the arrow keys just changes visual focus and not keyboard focus? (When you're in typeahead, browser focus stays in the input so that you can continue typing/interacting with the input.) Is that similar to how this would be handled?

Also, currently every log item has a tabIndex which makes me wonder if there will ever be interactive content in the log viewer. Will it always be static items or will there ever be the expectation to interact with log viewer items?

DaoDaoNoCode (commented):
@mattnolting Looks great, nice work! A small question, why we need both `With drilldown expanded` and `With dropdown expanded` examples? I know the first one is for mobile view and the second one is for desktop view, but they will not show at the same time, right? Will it be better to use a single example named like `With dropdown/drilldown menu open` instead of these 2 to show dropdown menu on desktop and show drilldown menu on mobile? It will more clear to me, because currently the `With drilldown expanded` example is the same as the basic example on large screen width, and `With dropdown expanded` example is the same as the basic example on small screen width. I cannot differentiate them from basic example at the first glance, which may also confuse users. Just an advice! Other parts look perfect, thanks!

DaoDaoNoCode:
BTW, I found a small issue on `Search Input`. In the design, the left icon is "search previous" and the right icon is "search next".

Screen Shot 2021-05-03 at 3 36 16 PM

However, in the implement of `Search Input`, they are different.

Screen Shot 2021-05-03 at 3 31 27 PM

Which makes this example look weird IMO.
@mattnolting @mcarrano WDYT? The problem is not in this PR, should we create an issue to track and fix this in `Search Input` component to make it the same as the design? Thanks.

mcarrano:
Good catch @DaoDaoNoCode . Yes that looks like a bug in the HTML example. In the React example they are correct:

Screen Shot 2021-05-03 at 5 00 46 PM

mattnolting:
Great feedback @DaoDaoNoCode! I've combined the examples with expanded dropdowns per your suggestion. 

@mcarrano @kmcfaul I've updated `pf-m-selected` to `pf-m-match` and `pf-m-current`. 

@mcarrano I've updated the hover style to match `pf-m-match` and @jessiehuff  removed `tabindex` on `pf-c-log-viewer__list-item`.

The `pf-c-search-input` navigation buttons have been updated as well, which will affect both `pf-c-search-input` and `pf-c-log-viewer`.

nicolethoen (approved):

DaoDaoNoCode (approved):
LPTM!! Thanks @mattnolting 👏

kmcfaul (approved):

dlabrecq (commented):
LGTM

PR #4033, opened 2021-04-30
refactor(menu): modified scroll behavior

### LINKED ISSUE #3975: Select - when using a scrollable list, search filter and footer should remain fixed
This is what the menu looks like when using `maxHeight` with the select component. The whole box scrolls instead of just the list of items.

Screen Shot 2021-04-05 at 7 22 43 PM

The search filter should probably remain static. And with the upcoming footer enhancement from https://github.com/patternfly/patternfly/pull/3968, the list should scroll between the filter and footer, like the context selector (below)

Screen Shot 2021-04-05 at 7 24 02 PM

wdty @mcarrano?

### LINKED ISSUE #4046: Update menu component scrolling
We added basic scrolling support to the menu component in https://github.com/patternfly/patternfly/pull/3999, but then commented it out in https://github.com/patternfly/patternfly/pull/3968 once we realized the list of menu items will need to scroll independently of the menu header and footer elements and the work in #3999 wouldn't allow for that.

Let's refactor that scrolling work so that the list of items can scroll while the header and footer remain static.

cc @mattnolting

mattnolting:
closes #4046


src/patternfly/components/Menu/menu.scss
```diff
@@ -6,7 +6,7 @@
   --pf-c-menu--BoxShadow: var(--pf-global--BoxShadow--md);
   --pf-c-menu--MinWidth: 100%;
   --pf-c-menu--Width: auto;
-  --pf-c-menu--MaxHeight: auto;
+  --pf-c-menu--MaxHeight: auto; // remove at breaking change
```
mcoker, line 9: This component is beta, so you can remove the var if you need to - but looks like it's still being used?

src/patternfly/components/Menu/menu.scss
```diff
@@ -246,6 +253,10 @@
     }
   }
   // stylelint-enable
+  .pf-c-divider {
+    --pf-c-select-menu--c-divider--MarginTop: 0;
```
mcoker, line 257: Not sure what these do since they're select component styles? Also would you mind leaving a comment to update that var name in a breaking change release?

mcoker: Oh nevermind, I see the example above. I was expecting the fix for https://github.com/patternfly/patternfly/issues/3975 to be in the select component. This fix pairs the menu component with the select component, and as far as I'm aware we haven't decided how we're going to use the menu component with the toggle and paired with other components yet, so it's probably something we should discuss to make sure we're taking the right approach.

Also I forgot to open an issue after we commented out the previous menu scrolling enhancement once we realized that it didn't allow the menu items to scroll while leaving a header or footer static. Here's that issue - https://github.com/patternfly/patternfly/issues/4046

mcarrano (commented):
@mattnolting looks great in Menu. For Select, is there an example where this behavior is shown?

mcoker (approved):
LGTM! 🥳

mattnolting:
> @mattnolting looks great in Menu. For Select, is there an example where this behavior is shown?

@mcarrano No, not until the `menu` component is promoted out of beta.

mcoker:
@mattnolting this looks good to me, but there is a conflict now.

src/patternfly/components/Menu/menu.scss
```diff
@@ -6,8 +6,6 @@
   --pf-c-menu--BoxShadow: var(--pf-global--BoxShadow--md);
   --pf-c-menu--MinWidth: 100%;
   --pf-c-menu--Width: auto;
-  --pf-c-menu--MaxHeight: auto;
-  --pf-c-menu--ZIndex: var(--pf-global--ZIndex--sm);
```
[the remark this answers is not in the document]
mattnolting: Nope, that was an oversight

mcoker (approved):
LGTM!

PR #4043, opened 2021-05-05
feat(masthead): added dropdown and toolbar support

### LINKED ISSUE #2808: Header dropdown styling
This issues should be addressed directly after, or with #2801

- Dropdowns should be full height of header

> I wonder if we should just make the dropdown enhancement, then the page enhancements once we finish the overall masthead story? Some questions:
> 
> * If someone is using another dropdown in the header, do we want the bottom border styles applied?
> * Would we want the underline styles to apply to the existing user dropdown without the avatar?
> * Can we handle the bottom border better so that the sides aren't angled?
> * Since we can include this in the breaking release, do we want to support using the avatar outside of the dropdown menu anymore?
> 
> Here's a screenshot of another dropdown in the header with the underline styles, also showing the angled sides.
> 
> Screen Shot 2020-03-10 at 5 50 09 PM

Line 170
```
  // Dropdown
  --pf-c-page__header--c-dropdown__toggle--before--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
  --pf-c-page__header--c-dropdown__toggle--before--hover--BorderBottomColor: var(--pf-global--active-color--400);

  // Toggle
  --pf-c-header--c-dropdown__toggle--before--BorderColor: transparent;
  --pf-c-header--c-dropdown__toggle--before--BorderWidth: 0;
  --pf-c-header--c-dropdown__toggle--before--BorderBottomColor: transparent;
```

Line 215-233
```

  .pf-c-dropdown {
    .pf-c-avatar {
      --pf-c-page__header-tools--c-avatar--MarginLeft: 0;
    }
  }

  .pf-c-dropdown__toggle {
    --pf-c-dropdown__toggle--BorderTopColor: transparent;
    --pf-c-dropdown__toggle--BorderRightColor: transparent;
    --pf-c-dropdown__toggle--BorderBottomColor: transparent;
    --pf-c-dropdown__toggle--BorderLeftColor: transparent;
    --pf-c-dropdown__toggle--hover--BorderBottomColor: var(--pf-c-page__header--c-dropdown__toggle--before--hover--BorderBottomColor);

    &::before {
      border-bottom-width: var(--pf-c-page__header--c-dropdown__toggle--before--BorderBottomWidth);
    }
  }

```

### LINKED ISSUE #3657: Menu enhancements, fit in dropdown
Since we'll be using the `menu` inside `dropdown`s, `menu` needs some additional contextual CSS support. `pf-c-dropdown > pf-c-menu` needs to be `position: absolute` and respond to `pf-m-top` and `pf-m-right`.

Also, menu needs to replicate dropdown `pf-m-text` https://www.patternfly.org/v4/components/dropdown/html#dropdown-with-image-and-text

mattnolting:
closes #2808 

@mcoker @mcarrano This PR is designed to support the next step, which is fitting all the toolbar buttons and toggle group items in masthead toolbar.


src/patternfly/components/Masthead/masthead.scss
```diff
@@ -127,19 +157,91 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
...
+    --pf-c-context-selector__toggle--BorderRightColor: transparent;
+    --pf-c-context-selector__toggle--BorderLeftColor: transparent;
+    --pf-c-context-selector__toggle--expanded--BorderBottomColor: var(--pf-c-masthead--c-context-selector__toggle--expanded--before--BorderBottomColor);
+  }
+
+  .pf-c-context-selector__toggle::before {
+    border-width: 0;
+    border-bottom-width: var(--pf-c-masthead--c-context-selector__toggle--expanded--before--BorderBottomWidth);
+  }
+
```
mcoker, line 224: With this approach, we'll need to maintain a list of components/elements that need to be modified here. There are a lot of specific styles for those components being restyled in the masthead. I'm wondering if it makes more sense to maintain a variation of those components instead?

Also this will apply to all dropdowns/context selectors in the masthead - maybe that's good? Or maybe we should target these a little more specifically? I'd be interested to hear your thoughts on that.

mcarrano:
@mattnolting What is the scope of this? Looks like it adds the full height dropdown item for use as a content selector? Anything else I should specifically be looking for? For this example, https://patternfly-pr-4043.surge.sh/components/masthead/html-demos#basic, not sure the styling is exactly right. 

@mceledonia can you take a look? I think that we intended there to be vertical borders on either side of the dropdown that stretched over the height of the masterhead, but I could be mistaken. Can we compare against your masthead mocks?

PR #4045, opened 2021-05-05
chore(card): put expandable right aligned toggle in regular DOM order

### LINKED ISSUE #3924: Card actions element margin
Looks like the header toggle margin vars were applied to the card actions element - https://github.com/patternfly/patternfly/pull/3586/files#diff-f77b58a7cf3d9657d084dec7fcc8ac27020c93e9dedea7d718cc122ccf48cd83R135

Card actions should have its own vars. Also the negative top/bottom margin only works with card actions if there is a plain button/dropdown. If there is only a checkbox or regular/primary button, for example, the alignment of the actions element is off. And it does not need a margin-right.

### LINKED ISSUE #4044: Card - move toggle in DOM for right aligned toggles
https://github.com/patternfly/patternfly-react/pull/5723#pullrequestreview-649247293

> So this one's a bit odd with keyboard. The tab order (and DOM structure) don't really match the visual presentation which makes for a kind of jumpy experience. Is there a way that we can put the toggle in the DOM after the actions? Then the DOM order, tab order, and visual order would all match. I'm not sure if this is something we'd have do do in core though?

mattnolting:
closes #4044

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -36,6 +36,12 @@
   --pf-c-card__header-toggle--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-card__header-toggle--MarginBottom: calc(var(--pf-global--spacer--form-element) * -1);
   --pf-c-card__header-toggle--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
+  --pf-c-card__actions--MarginTop: 0;
+  --pf-c-card__actions--MarginRight: 0;
+  --pf-c-card__actions--MarginBottom: 0;
+  --pf-c-card__actions--MarginLeft: 0;
+  --pf-c-card__actions__dropdown--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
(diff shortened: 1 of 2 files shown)


evwilkin:
👀  watching for https://github.com/patternfly/patternfly-react/pull/5723#pullrequestreview-649247293

src/patternfly/components/Card/card.scss
```diff
@@ -213,6 +223,11 @@
   + .pf-c-card__footer {
     padding: 0;
   }
+
+  .pf-c-dropdown {
```
mcoker, line 227: I think moving the margin that offsets the alignment from the actions container to an individual component is going to cause inconsistencies. You could use other plain things here, too - button, options menu, etc, and then the alignment would be different.

As an example, this is what it looks like currently if you use a dropdown and a button

Screen Shot 2021-05-05 at 10 32 28 AM

And here is what that looks like with this change3

Screen Shot 2021-05-05 at 10 32 20 AM

mcoker (changes requested):

mcoker (approved):

PR #4052, opened 2021-05-06
fix(drawer): made panel main/body elements fill height on mobile

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4051


evwilkin (approved):
LGTM - thanks!

Issue #3779, opened 2021-01-06
Table demos missing border between table and pagination


mcoker:
On any demo where the table + pagination is in a card, the last table row's bottom border is not showing:

<img width="1168" alt="Screen Shot 2021-01-06 at 3 30 20 PM" src="https://user-images.githubusercontent.com/35148959/103825786-20e36c80-503b-11eb-95fe-67e6bd1997ca.png">

Here is an example - https://www.patternfly.org/v4/components/table/html-demos/sortable/

PR #4055, opened 2021-05-07
feat: fixed missing border in table demos (#3779)

agagancarczyk:


agagancarczyk:
hi @mcoker this PR is ready for review. Thanks

mcoker:
Hi @agagancarczyk, thanks for the PR! Is there an issue number that references the original problem? From the looks of it, this just removes the card that wraps the toolbar/table/pagination in the table demos? cc @mcarrano wdyt?

Also can you rebase this PR? Looks like there are some unnecessary card changes showing up in the file changes.

mcarrano:
Would like to better understand what problem this is solving.

mcoker:
Hi @agagancarczyk just wanted to follow up and see if you were still working on this PR?

agagancarczyk:
hi @mcoker I have now rebased this PR - please review. @mcarrano this is fixing a missing border between table and pagination in the table demos https://github.com/patternfly/patternfly/issues/3779. Thanks

mcarrano (approved):
Looks good to me. Thanks @agagancarczyk !

mcoker (changes requested):
Thanks for updating the issue!

While removing the card fixes the demos, including a table + pagination in a card is still a valid use case, so the root issue remains.

mcoker:
It's really odd that the border disappears in the card component like that!  From looking at it a little, it looks to be an issue specific to Chrome, and is some sort of problem with `display: flex` on the card, and the table's bottom border (coming from the last `<tr>`) overflowing the table - then the natural stacking context of the pagination places it on top of the table row's border, hiding it.

A way you can test this is to any of the following. Each will show the border:

1. remove `display: flex` from the card. This shows that it has something to do with the way flex handles the space between the table and pagination.
2. add `z-index: 1;` to the table. Shows that the border is there and is being hidden by pagination.
3. remove the border on the `<tr>` and place it on the `<table>` instead. Shows that a regular border on the table works fine, it must have something to do with a border specifically on a `<tr>` when the table is part of a flex layout?

I'm wondering if there is a workaround, or maybe a chrome bug filed that explains the issue with a potential workaround?

agagancarczyk:
Thank you @mcoker yes, this is strange behavior. I wasn't able to find any solution explaining this behavior. Would you like to close this PR for now? Many thanks

mcoker:
@agagancarczyk sounds good, if it comes up again we can investigate further. Thanks for looking into it!

PR #4054, opened 2021-05-07
fix(toggle-group): updated colors, borders, removed light variant

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3951

PR #4059, opened 2021-05-10
fix(toggle-group): moved borders to before pseudo element

mcoker:
small bug caught after merging https://github.com/patternfly/patternfly/pull/4054


mattnolting (approved):
Perfect! ⭐

PR #4067, opened 2021-05-13
feat(content): added support for visited link styling

### LINKED ISSUE #3726: Visited style for links
Requesting a design exploration for a visited style for links. Possibly also for other elements (navigation, tabs, buttons, etc?)
@mceledonia

mattnolting:
closes #3726 

Preview: https://patternfly-pr-4067.surge.sh/components/text

### DIFF
--- a/src/patternfly/base/_globals.scss
+++ b/src/patternfly/base/_globals.scss
@@ -120,6 +120,10 @@
       --pf-global--link--Color: var(--pf-global--link--Color--hover);
       --pf-global--link--TextDecoration: var(--pf-global--link--TextDecoration--hover);
     }
+
+    &.pf-m-visited:visited {
+      color: var(--pf-global--link--Color--visited);
+    }
   }
 
   button,
--- a/src/patternfly/base/_globals.scss
+++ b/src/patternfly/base/_globals.scss
@@ -120,10 +120,6 @@
       --pf-global--link--Color: var(--pf-global--link--Color--hover);
       --pf-global--link--TextDecoration: var(--pf-global--link--TextDecoration--hover);
     }
-
-    &.pf-m-visited:visited {
-      color: var(--pf-global--link--Color--visited);
-    }
   }
 
   button,
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -110,12 +110,14 @@
     }
   }
 
-  &.pf-m-visited,
-  .pf-m-visited {
+  // stylelint-disable selector-no-qualifying-type
+  &.pf-m-visited a,
+  a.pf-m-visited {
     &:visited {
-      color: var(--pf-c-content--a--visited--Color);
+      --pf-c-content--a--Color: var(--pf-c-content--a--visited--Color);
     }
   }
+  // stylelint-enable
 
   li + li {
     margin-top: var(--pf-c-content--li--MarginTop);
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -64,6 +64,7 @@
   --pf-c-content--a--TextDecoration: var(--pf-global--link--TextDecoration);
   --pf-c-content--a--hover--Color: var(--pf-global--link--Color--hover);
   --pf-c-content--a--hover--TextDecoration: var(--pf-global--link--TextDecoration--hover);
+  --pf-c-content--a--visited--Color: var(--pf-global--link--Color--visited);
 
   // Blockquote
   --pf-c-content--blockquote--PaddingTop: var(--pf-global--spacer--md);
@@ -106,6 +107,13 @@
     &:hover {
       --pf-c-content--a--Color: var(--pf-c-content--a--hover--Color);
       --pf-c-content--a--TextDecoration: var(--pf-c-content--a--hover--TextDecoration);
+    }
+  }
+
+  &.pf-m-visited,
+  .pf-m-visited {
+    &:visited {
+      color: var(--pf-c-content--a--visited--Color);
     }
   }
 
--- a/src/patternfly/components/Content/examples/Content.md
+++ b/src/patternfly/components/Content/examples/Content.md
@@ -119,5 +119,5 @@ This component is an exception to the variable system since we style type select
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rythm and typographic treatment to html elements. |
+| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rhythm and typographic treatment to html elements. |
 | `.pf-m-visited` | `.pf-c-content`, `<a>` | Modifies all links in a content block to include visited styles. Can also be applied to a single link in a content block. |
--- a/src/patternfly/components/Content/examples/Content.md
+++ b/src/patternfly/components/Content/examples/Content.md
@@ -63,6 +63,9 @@ cssPrefix: pf-c-content
 <p>Quisque at semper enim, eu hendrerit odio. Etiam auctor nisl et
   <em>justo sodales</em> elementum. Maecenas ultrices lacus quis neque consectetur, et lobortis nisi molestie.</p>
 <hr>
+<h3>Visited link example</h3>
+<a class="pf-m-visited" href="#">Visited link</a>
+<hr>
 <p>Sed sagittis enim ac tortor maximus rutrum. Nulla facilisi. Donec mattis vulputate risus in luctus. Maecenas vestibulum interdum
   commodo.
 </p>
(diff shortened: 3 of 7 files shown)


mcarrano (commented):
Looks good to me @mattnolting .

@mceledonia @ajacobs21e can you also verify that this is what you expect?

ajacobs21e:
LGTM

mcoker:
@mcarrano with this change, say a user has a link in a data list cell, like the link seen in this demo - https://patternfly-pr-4067.surge.sh/components/data-list/html/expandable/

That link is not in a content component, because it's just text and a link and there is no need for the formatting that the content component provides. And then in the expanded content, they have a couple of paragraphs with links within the content component to provide paragraph formatting. The regular cell link will not have visited styling, and the links in the expanded content will - would this not be considered breaking? Also it may be really hard for a user to track down why this visited styling is being applied inconsistently, too.

Screen Shot 2021-05-13 at 9 24 12 AM

mcarrano:
@mcoker I see how this can be potentially confusing, but I think that in most cases a link in a list or table would be more navigational. I guess that someone could fix this by wrapping the cell text in a Text (content) component, correct? 

I'd be OK to live with this. But what do you think @ajacobs21e @mceledonia ?  I guess we could also define a new Text variant that was something like "Documentation block" that picked up this styling and give the consumer more choice. But it feels like unnecessary complexity. I did want to float this out to other designers at next week's PF design share meeting so maybe we hold off on merging until then?

mcoker:
> I see how this can be potentially confusing, but I think that in most cases a link in a list or table would be more navigational. I guess that someone could fix this by wrapping the cell text in a Text (content) component, correct?

@mcarrano I guess my point is just that users could easily have links in their app both inside and outside of the content component that they want to have visited styling, and pushing this could give inconsistent visited styling, requiring updates to fix - which is typically how we define a breaking change.

Also it's not obvious to users that visited styling comes from the content component, so if a user needs to fix an inconsistency brought in from this PR, it could be tricky for a user to track down why this is happening and how to fix it.

Given these potential issues, I'm fine committing the change if you and the rest of the team are, I just want to make sure we've considered this since it could be a potential issue.

ajacobs21e:
@mcoker is it possible to add in a visited link style as an option? So no existing links would get this treatment (so nothing existing breaks), but it could be applied to existing and future links if the dev/des wants to?

mcoker:
@ajacobs21e yep, it could be an opt-in with the content component - basically the change we made here, except you would apply a modifier class to the content component to get the styling.

And if we expect this to be a global style at some point, it probably makes sense to introduce this as a global var (`--pf-global--link--visited--Color`), then reference that var in the content component.

mcarrano:
@mcoker if we can do this via a modifier class, then I'm good with that. How would that be exposed in the documentation? Would we create an example that shows the modifier class applied?

mcoker:
@mcarrano yep that sounds right to me. The class would be in the component docs table, and we can have an example of it in use.

mcarrano:
@mcoker I like that approach. It protects us from those who might not like this change (for whatever reason). @mattnolting what are your thoughts? Are you OK to rework this as suggested?

mattnolting:
> @mcoker I like that approach. It protects us from those who might not like this change (for whatever reason). @mattnolting what are your thoughts? Are you OK to rework this as suggested?

@mcarrano That works for me. I think we would want the ability to apply `.pf-m-visited-link` to `.pf-c-content` as well as individual `a` tags. @mcoker does that sound right to you?

mcoker:
@mattnolting yep, sounds good

mattnolting:
@mcoker @mcarrano updated

mcarrano (approved):
Looks good to me. Thanks @mattnolting !

src/patternfly/components/Content/content.scss
```diff
@@ -109,6 +110,13 @@
     }
   }
 
+  &.pf-m-visited,
+  .pf-m-visited {
+    &:visited {
+      color: var(--pf-c-content--a--visited--Color);
+    }
+  }
+
```
mcoker, line 119: This compiles to

```
.pf-c-content.pf-m-visited:visited,
.pf-c-content .pf-m-visited:visited {
  color: var(--pf-c-content--a--visited--Color);
}
```

Which won't target all `:visited` links in a content block with `.pf-m-visited`

I think maybe you meant to do this?

```
.pf-c-content {
  &.pf-m-visited :visited,
  .pf-m-visited:visited {
      color: var(--pf-c-content--a--visited--Color);
  }
}
```

Which compiles to:

```
.pf-c-content.pf-m-visited :visited, /* targets links in a block with .pf-m-visited */
.pf-c-content .pf-m-visited:visited { /* targets a link with .pf-m-visited */
  color: var(--pf-c-content--a--visited--Color);
}
```

Another way would be to group the styles with the `a` element above, it seems more appropriate there:

```
    @at-root .pf-c-content.pf-m-visited a,
    &.pf-m-visited {
        &:visited {
            --pf-c-content--a--Color: var(--pf-c-content--a--visited--Color);
        }
    }
```

which compiles to:

```
.pf-c-content.pf-m-visited a:visited,
.pf-c-content a.pf-m-visited:visited {
  --pf-c-content--a--Color: var(--pf-c-content--a--visited--Color);
}
```

@mcarrano @mceledonia can you remind me if the eventual intent is to have all text links (that aren't part of a component, like nav, tabs, dropdown items, etc) have visited styles, and we'll make that update in a breaking change release? If so, we should probably also add the ability to make a single outside of the content component have visited styles by adding that modifier to the global `a` rules in https://github.com/patternfly/patternfly/blob/78eb519250b9aaf53391f366688f92f1d420c0ed/src/patternfly/base/_globals.scss#L114-L123

src/patternfly/components/Content/examples/Content.md
```diff
@@ -116,4 +119,5 @@ This component is an exception to the variable system since we style type select
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rythm and typographic treatment to html elements |
+| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rythm and typographic treatment to html elements. |
+| `.pf-m-visited-link` | `.pf-c-content`, `<a>` | Modifies a visited link styling to `--pf-color-purple-600`. |
```
mcoker, line 123: fixed the class name and also made the description less specific - if we update the color, we won't have to update this description.

```suggestion
| `.pf-m-visited` | `.pf-c-content`, `<a>` | Modifies all links in a content block to include visited styles. Can also be applied to a single link in a content block. |
```

mcoker (changes requested):

src/patternfly/components/Content/content.scss
mcarrano: @mcoker I don't know under what circumstances you would want to apply visited styles in places other than the content component, but it probably wouldn't hurt to allow that modifier to be globally available just in case someone wants to use it.

mcoker: @mcarrano it makes it a lot easier on our end if we limit these styles just to the content component. Then we don't have to worry about the overlapping styles that will result from global `:visited` styles interfering with component link elements that shouldn't have a `:visited` state - like nav, tabs, etc. If the eventual intent is to apply visited styles to links globally though, it's probably good to go ahead and introduce this variation for global links.

mcarrano: I'm good to keep this localized to the content component for now.

src/patternfly/base/_globals.scss
```diff
@@ -120,6 +120,10 @@
       --pf-global--link--Color: var(--pf-global--link--Color--hover);
       --pf-global--link--TextDecoration: var(--pf-global--link--TextDecoration--hover);
     }
+
+    &.pf-m-visited:visited {
```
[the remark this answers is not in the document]
mcoker: per @mcarrano's comment, we can remove this and just keep it in the content component - https://github.com/patternfly/patternfly/pull/4067/files#r635556709

src/patternfly/components/Content/examples/Content.md
```diff
@@ -116,4 +119,5 @@ This component is an exception to the variable system since we style type select
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rythm and typographic treatment to html elements |
+| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rythm and typographic treatment to html elements. |
```
mcoker, line 122: nit...

```suggestion
| `.pf-c-content` | `<div>`, `<section>`, or `<article>` | Generates vertical rhythm and typographic treatment to html elements. |
```

mcoker (changes requested):

mcoker (approved):
L🎉TM!

Issue #3959, opened 2021-03-30
card: when using first level title fonts and there is a button in the cards actions, alignment breaks  


KKoukiou:
This screenshot is with the the xl font, notice the button and title are not center aligned in the header wrapper element. 
![Screen Shot 2021-03-30 at 11 11 29](https://user-images.githubusercontent.com/14921356/112966121-36ec5f00-914a-11eb-9e2d-4e8f2138b251.png)

mcoker:
@KKoukiou interesting problem! We keep the actions aligned to the top of the card header for when the title text wraps, so that may always be a bit of an issue depending on the size of the thing in the card header that the actions are beside.

<img width="845" alt="Screen Shot 2021-04-06 at 1 43 37 PM" src="https://user-images.githubusercontent.com/35148959/113762370-21dd7600-96de-11eb-9301-5de9cba3bf25.png">

However it looks like in this case, it's the negative margin on the actions container that is causing an issue. We use a negative top/bottom margin on the cards container so that card headers with and without actions will have the same height and align with one another (https://codepen.io/mcoker/pen/RwKZeqw), but it looks like that's throwing the alignment off when the header container is taller than the actions (due to a large card title or whatever).

A temporary fix could be to set `.pf-c-card__actions { margin-top: 0; }`.

@mcarrano can we add this to a future milestone?

mcoker:
@KKoukiou there is a PR up here - https://github.com/patternfly/patternfly/pull/4071. When you get a chance, can you let me know if that will work for you?

PR #4071, opened 2021-05-14
feat(card): added variation to remove actions negative margin offset

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3959

There is an obvious problem with the way the actions element's negative margin 1) uses the wrong vars and 2) doesn't work in a lot of cases, like the problem in #3959. We have an issue to clean this up (https://github.com/patternfly/patternfly/issues/3924) but it will be a breaking change.


KKoukiou (approved):
Looks great, the title is not presenter center aligned with the button.

mattnolting (approved):
Nice, Perfecto!

Issue #4017, opened 2021-04-19
Add Rich Helper Text for Form elements


mcarrano:
Currently we support simple helper text for form fields that include text strings only. Want to allow for more complex structures that would support the requirement to provide real-time validation to be applied to passwords or other places where the user entered a string into a standard text input (like naming an object).

Designs: https://marvelapp.com/prototype/6jehce7/screen/78511266

mcarrano:
There may be some a11y concerns with this. Please add details @jessiehuff .

PR #4089, opened 2021-05-20
feat(helper-text): added helper text component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4017

Preview - https://patternfly-pr-4089.surge.sh/components/helper-text

Adds 2 types of helper texts:

### Static
This is the default helper text. With static helper text, the text color and icon color are the same. By default the text and optional icon are black. Indeterminate is grey, warning is yellow, success is green, invalid is red.

### Dynamic
This is an opt in to the helper text component. By default a dynamic helper text and icon are black. Intermediate is grey, warning has yellow icon and grey text, success has green icon and grey text, and invalid has red icon and black text.

A helper text component is a block made up of helper text items. There can be one or more items, and if there are multiple items in a helper text block, there is a `xs` spacer between items. A block of multiple items can be [regular divs](https://patternfly-pr-4089.surge.sh/components/helper-text/#multiple-static) (in the case of regular helper text that isn't necessarily a list), or an [HTML list](https://patternfly-pr-4089.surge.sh/components/helper-text/#dynamic-list) (particularly useful in dynamic/validation helper text).

### DIFF
--- a/src/patternfly/components/HelperText/helper-text.scss
+++ b/src/patternfly/components/HelperText/helper-text.scss
@@ -1,5 +1,5 @@
 .pf-c-helper-text {
-  --pf-c-helper-text--Gap: var(--pf-global--spacer--sm);
+  --pf-c-helper-text--Gap: var(--pf-global--spacer--xs);
   --pf-c-helper-text--FontSize: var(--pf-global--FontSize--sm);
 
   // item
--- a/src/patternfly/components/HelperText/helper-text.scss
+++ b/src/patternfly/components/HelperText/helper-text.scss
@@ -3,11 +3,16 @@
   --pf-c-helper-text--FontSize: var(--pf-global--FontSize--sm);
 
   // item
-  --pf-c-helper-text__item--Color: var(--pf-global--Color--100);
-  --pf-c-helper-text__item--m-indeterminate--Color: var(--pf-global--Color--200);
-  --pf-c-helper-text__item--m-warning--Color: var(--pf-global--warning-color--100);
-  --pf-c-helper-text__item--m-success--Color: var(--pf-global--success-color--100);
-  --pf-c-helper-text__item--m-invalid--Color: var(--pf-global--danger-color--100);
+  --pf-c-helper-text__item-icon--Color: var(--pf-global--Color--100);
+  --pf-c-helper-text__item-text--Color: var(--pf-global--Color--100);
+  --pf-c-helper-text__item-icon--m-indeterminate--Color: var(--pf-global--Color--200);
+  --pf-c-helper-text__item-text--m-indeterminate--Color: var(--pf-global--Color--200);
+  --pf-c-helper-text__item-icon--m-warning--Color: var(--pf-global--warning-color--100);
+  --pf-c-helper-text__item-text--m-warning--Color: var(--pf-global--warning-color--200);
+  --pf-c-helper-text__item-icon--m-success--Color: var(--pf-global--success-color--100);
+  --pf-c-helper-text__item-text--m-success--Color: var(--pf-global--success-color--200);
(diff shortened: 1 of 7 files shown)


mcarrano (commented):
This looks good @mcoker . My only feedback is that I think the yellow/warning version should use the darker gold text color for better contrast/readability. @mceledonia @doruskova @ajacobs21e can you also take a look?

doruskova (commented):
Agree with @mcarrano. The warning text should use `--pf-global--warning-color--200` because the light yellow is not accessible. However, it looks great! Thank you!

mcoker:
@mcarrano @doruskova thanks! Looking at the alert, we use `success/warning/danger-color--100` for the icon and `success/warning/danger-color--200` for the text. Should I use that here? 

In the form component, the error helper text and icon both are `danger-color--100`, however I'm not sure that's the correct implementation. We do not have a unique icon color for the form component helper text for any of the states. This is what it looks like currently - the success color is `success-color--200`, warning is `warning-color--200`.

Screen Shot 2021-05-24 at 11 05 24 AM


Here is this helper text PR with icon/text color that matches the alert component with `success/warning/danger-color--100` for the icon and `success/warning/danger-color--200` for the text.

Screen Shot 2021-05-24 at 11 06 08 AM

mcoker:
@mcarrano great, those color updates are in the PR now.

mcoker:
@mcarrano btw I called this component "Helper text" - is that OK or do you want to name it "Rich helper text"?

mcarrano:
I think that the name is fine @mcoker . It's more likely people will look for it under this name.

mcarrano (approved):
Looks good.

doruskova:
@mcoker Looks great!

mattnolting (changes requested):
Hey @mcoker, this is looking great! Couple of questions

- Not sure I understand the static/dynamic concept. It looks like they could be used interchangeably, based on messaging and/or intent. 

In this example, I would assume both helper texts are dynamic, but I wouldn't inherently know which variant to use. 

Screen Shot 2021-05-25 at 9 18 45 AM

I think this could be solved by either changing `dynamic` to another name or adding documentation explaining when to use which variant. 

- Looks like you need to add standard documentation at the end of `HelperText.md`

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -0,0 +1,83 @@
...
+    --pf-c-helper-text__item-icon--Color: var(--pf-c-helper-text__item-icon--m-invalid--Color);
+    --pf-c-helper-text__item-text--Color: var(--pf-c-helper-text__item-text--m-invalid--Color);
+    --pf-c-helper-text--m-dynamic__item-icon--Color: var(--pf-c-helper-text--m-dynamic--m-invalid__item-icon--Color);
+    --pf-c-helper-text--m-dynamic__item-text--Color: var(--pf-c-helper-text--m-dynamic--m-invalid__item-text--Color);
+  }
+
+  &.pf-m-dynamic {
+    --pf-c-helper-text__item-text--Color: var(--pf-c-helper-text--m-dynamic__item-text--Color);
+    --pf-c-helper-text__item-icon--Color: var(--pf-c-helper-text--m-dynamic__item-icon--Color);
+  }
```
mattnolting, line 73: 🤩

mcoker:
@mattnolting updated the docs

mcoker:
@mattnolting I chatted with @mcarrano offline about the confusion using the dynamic modifier, and didn't have an obvious solution. I'm fine with merging it as is and following up if we need to. wdyt?

mattnolting (approved):
LAPTM 😄  Let's follow up w/naming the variant.

PR #4090, opened 2021-05-21
refactor(menu): changed translate to left

### LINKED ISSUE #4062: Menu component positioned with popper transitions unexpectedly
Re: https://github.com/patternfly/patternfly-react/pull/5722

https://patternfly-react-pr-5722.surge.sh/demos/composable-menu/#drilldown-menu-with-toggle

Popper is being used to position the menu relative to the toggle, and it uses `translate()` for positioning. We have a `transition` for `transform` that is being applied and causing the menu to transition the first time it is loaded.

Let's review that implementation and see if we need to transition `transform`, and if so, make sure popper's positioning via `transform` isn't interfering with our use of of that property in the CSS. We may also see if it's better for popper to use a wrapping element for positioning so it doesn't interfere with our component transitions.

Also as an aside, we're transitioning `height`, too https://github.com/patternfly/patternfly/blob/722442b2f6f76fba82df89c92362264515c9a31e/src/patternfly/components/Menu/menu.scss#L111

Does that need to move to `__content` since that element's height is being animated?

mattnolting:
closes #4062

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -13,6 +13,9 @@
   --pf-c-menu--m-flyout__list-item--m-menu-top__menu--Bottom: calc(var(--pf-c-menu__list--PaddingTop) * -1);
   --pf-c-menu--m-flyout__list-item--m-menu-left__menu--Right: 100%;
 
+  // Content
+  --pf-c-menu__content--Height: auto;
+
   // Divider
   --pf-c-menu__list--c-divider--MarginTop: var(--pf-global--spacer--sm);
   --pf-c-menu__list--c-divider--MarginBottom: var(--pf-global--spacer--sm);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Menu/menu.scss
```diff
@@ -173,23 +169,37 @@
   &.pf-m-drilldown {
     display: flex;
     flex-direction: column;
-    transition: var(--pf-c-menu--m-drilldown--Transition);
```
mcoker, line 176: I believe this change will fix the original issue. PopperJS is applying `transform: translate3d(...)` to the outer `.pf-c-menu` to position the menu relative to the toggle, and this transition on `.pf-c-menu.pf-m-drilldown` was causing that position to transition in. You can see it if you add `className="pf-m-drilldown"` to one of the `<Menu>` components in the [composable menu demos](https://www.patternfly.org/v4/demos/composable-menu). 

Unless there's another reason to transition `left` on the inner menus, I think transitioning `translate()` is better because it's more performant.

src/patternfly/components/Menu/menu.scss
```diff
@@ -173,23 +169,37 @@
...
+
+    .pf-c-menu,
+    .pf-c-menu__list {
+      transition: var(--pf-c-menu--m-drilldown--Transition);
```
mcoker, line 175: Should this var be renamed?

mcoker (changes requested):

src/patternfly/components/Menu/menu.scss
mcoker: Also an aside, I found this in the popperJS docs - looks like you may be able to change the way popper positions the menu to stick with `top` and `left` positioning if we still have conflicting styles - https://popper.js.org/docs/v2/modifiers/compute-styles/#gpuacceleration.

mcoker (approved):
L🔥TM!

PR #4091, opened 2021-05-21
fix(drawer): added panel min-width

### LINKED ISSUE #3795: Resizable drawer panel box shadow disappears when panel size is small
There should be a box shadow for drawer panel

Screen Shot 2021-01-13 at 4 16 50 PM

However, when the panel size is very small (likely `flex-basis` less than 2%), the shadow will disappear.

Screen Shot 2021-01-13 at 4 18 41 PM

----

Also with inline drawers, one of the splitter borders disappears when you resize the panel to 0, and the x-axis scrollbar appears.

Screen Shot 2021-02-04 at 12 22 15 PM

mattnolting:
closes #3795


DaoDaoNoCode (approved):
Look 😎 to me, thanks @mattnolting !

mcoker (changes requested):
Looks great! Can you also add support for a `min-height` (and undo the `min-width`) on the `.pf-m-panel-bottom` variation? Looks like we have the same problem with the shadow there.

mcoker (approved):
LPTM! 🏆

PR #4095, opened 2021-05-24
fix(context-selector): added plain text variant and example

### LINKED ISSUE #3954: Add "plain" variant of Context Selector
Add a Plain variant of the Context selector component. Also update the examples to show context selectors on both dark and light backgrounds. Design issue for reference: https://github.com/patternfly/patternfly-design/issues/934

bitwiseor:
fixes #3954

### DIFF
--- a/src/patternfly/components/ContextSelector/context-selector-toggle.hbs
+++ b/src/patternfly/components/ContextSelector/context-selector-toggle.hbs
@@ -1,4 +1,4 @@
-<button class="pf-c-context-selector__toggle{{#if context-selector-toggle--modifier}} {{context-selector-toggle--modifier}}{{/if}}"
+<button class="pf-c-context-selector__toggle{{#if context-selector-toggle--IsPlain}} pf-m-plain{{/if}}{{#if context-selector-toggle--modifier}} {{context-selector-toggle--modifier}}{{/if}}"
   {{#if context-selector--IsExpanded}}
     aria-expanded="true"
   {{else}}
--- a/src/patternfly/components/ContextSelector/context-selector-toggle.hbs
+++ b/src/patternfly/components/ContextSelector/context-selector-toggle.hbs
@@ -1,4 +1,4 @@
-<button class="pf-c-context-selector__toggle{{#if context-selector-toggle--IsPlain}} pf-m-plain{{/if}}{{#if context-selector-toggle--modifier}} {{context-selector-toggle--modifier}}{{/if}}"
+<button class="pf-c-context-selector__toggle{{#if context-selector-toggle--IsText}} pf-m-text{{/if}}{{#if context-selector-toggle--IsPlain}} pf-m-plain{{/if}}{{#if context-selector-toggle--modifier}} {{context-selector-toggle--modifier}}{{/if}}"
   {{#if context-selector--IsExpanded}}
     aria-expanded="true"
   {{else}}
@@ -7,5 +7,10 @@
   {{#if context-selector-toggle--attribute}}
     {{{context-selector-toggle--attribute}}}
   {{/if}}>
+  {{#if context-selector-toggle--icon}}
+      {{#> context-selector-toggle-icon}}
+        {{context-selector-toggle--icon}}
+      {{/context-selector-toggle-icon}}
+    {{/if}}
   {{> @partial-block}}
 </button>
--- a/src/patternfly/components/ContextSelector/context-selector-toggle.hbs
+++ b/src/patternfly/components/ContextSelector/context-selector-toggle.hbs
@@ -7,10 +7,5 @@
   {{#if context-selector-toggle--attribute}}
     {{{context-selector-toggle--attribute}}}
   {{/if}}>
-  {{#if context-selector-toggle--icon}}
-      {{#> context-selector-toggle-icon}}
-        {{context-selector-toggle--icon}}
-      {{/context-selector-toggle-icon}}
-    {{/if}}
   {{> @partial-block}}
 </button>
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -18,6 +18,11 @@
   --pf-c-context-selector__toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-context-selector__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-context-selector__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
+  --pf-c-context-selector__toggle--m-plain--Color: var(--pf-global--Color--100);
+  --pf-c-context-selector__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
+  --pf-c-context-selector__toggle--m-plain--disabled--Color: var(--pf-global--disabled-color--200);
+  --pf-c-context-selector__toggle--m-plain--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-context-selector__toggle--m-plain--PaddingLeft: var(--pf-global--spacer--sm);
 
   // Toggle text
   --pf-c-context-selector__toggle-text--FontSize: var(--pf-global--FontSize--md);
@@ -115,6 +120,23 @@
       --pf-c-context-selector__toggle--BorderBottomColor: var(--pf-c-context-selector__toggle--expanded--BorderBottomColor);
 
       border-bottom-width: var(--pf-c-context-selector__toggle--expanded--BorderBottomWidth);
+    }
+  }
+
+  &.pf-m-plain {
+    --pf-c-context-selector__toggle--PaddingRight: var(--pf-c-context-selector__toggle--m-plain--PaddingRight);
+    --pf-c-context-selector__toggle--PaddingLeft: var(--pf-c-context-selector__toggle--m-plain--PaddingLeft);
+
+    // display: inline-block;
+    color: var(--pf-c-context-selector__toggle--m-plain--Color);
+
+    // remove at breaking change
+    > * {
+      line-height: var(--pf-c-context-selector__toggle--m-plain--child--LineHeight);
+    }
+
+    &::before {
+      border: 0;
     }
   }
 
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -18,7 +18,7 @@
   --pf-c-context-selector__toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-context-selector__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-context-selector__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
-  --pf-c-context-selector__toggle--m-plain--Color: var(--pf-global--Color--100);
+  --pf-c-context-selector__toggle--m-plain--Color: var(--pf-global--Color--200);
   --pf-c-context-selector__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
   --pf-c-context-selector__toggle--m-plain--disabled--Color: var(--pf-global--disabled-color--200);
   --pf-c-context-selector__toggle--m-plain--PaddingRight: var(--pf-global--spacer--sm);
@@ -69,6 +69,11 @@
   --pf-c-context-selector__menu-list-item--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-context-selector__menu-list-item--hover--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
   --pf-c-context-selector__menu-list-item--disabled--Color: var(--pf-global--Color--dark-200);
+
+  // Menu item icon
+  --pf-c-context-selector__menu-item-icon--Color: var(--pf-global--active-color--100);
+  --pf-c-context-selector__menu-item-icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-context-selector__menu-item-icon--PaddingLeft: var(--pf-global--spacer--lg);
 
   position: relative;
   display: inline-block;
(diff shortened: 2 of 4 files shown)


bitwiseor:
This is actually a plain text variant; still need to create the plain variant!

mcoker (changes requested):
Leaving a review as a test

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -144,6 +199,15 @@
   box-shadow: var(--pf-c-context-selector__menu--BoxShadow);
 }
 
+.pf-c-context-selector__menu-item-icon {
```
mcoker, line 202: Is this element used anywhere?

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -132,6 +174,19 @@
   }
 }
 
+.pf-c-context-selector__toggle-button-icon {
```
mcoker, line 177: Is this element used anywhere?

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -132,6 +174,19 @@
...
+  position: relative;
+}
+
+.pf-c-context-selector__toggle-icon {
+  margin-right: var(--pf-c-context-selector__toggle-icon--MarginRight);
+  margin-left: var(--pf-c-context-selector__toggle-icon--MarginLeft);
```
mcoker, line 183: It looks like these margin styles are already included here - https://github.com/patternfly/patternfly/pull/4095/files#diff-f6fa17cc8d90aa2f9022c472197706093dbe9f9330d00d4ff2d2dd5892e0e450R163-R166

However, I think moving them out of the toggle class selector is a great idea! There is no need for those styles to be nested the way they are. Can you remove those lines in favor of this block you added?

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -132,6 +174,19 @@
...
+
+  .pf-c-context-selector.pf-m-top.pf-m-expanded & {
```
mcoker, line 185: We don't have a `.pf-m-top` class for context selector, so you don't need this block.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -18,6 +18,11 @@
   --pf-c-context-selector__toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-context-selector__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-context-selector__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
+  --pf-c-context-selector__toggle--m-plain--Color: var(--pf-global--Color--200);
+  --pf-c-context-selector__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
+  --pf-c-context-selector__toggle--m-plain--disabled--Color: var(--pf-global--disabled-color--200);
+  --pf-c-context-selector__toggle--m-plain--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-context-selector__toggle--m-plain--PaddingLeft: var(--pf-global--spacer--sm);
```
mcoker, line 25: Let's use the `--md` spacer for these `--m-plain--Padding` vars. This will be used with `.pf-m-plain` on it's own. Then let's create 2 new vars for `--m-plain--m-text--Padding[Right/Left]` that use the `--sm` spacer.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -118,6 +128,38 @@
     }
   }
 
+  &.pf-m-plain {
+    --pf-c-context-selector__toggle--PaddingRight: var(--pf-c-context-selector__toggle--m-plain--PaddingRight);
+    --pf-c-context-selector__toggle--PaddingLeft: var(--pf-c-context-selector__toggle--m-plain--PaddingLeft);
+
+    &:not(.pf-m-text) {
+      justify-content: center;
```
mcoker, line 136: Let's add `display: inline-block` here for the `.pf-m-plain` (non-text) variant, then we can remove `justify-content` since that property doesn't apply to inline-block elements.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -118,6 +128,38 @@
...
+    &:hover,
+    &:active,
+    &.pf-m-active,
+    &:focus,
+    .pf-c-context-selector.pf-m-expanded > & {
+      --pf-c-context-selector__toggle--m-plain--Color: var(--pf-c-context-selector__toggle--m-plain--hover--Color);
+    }
+
+    &.pf-m-disabled,
+    &:disabled {
```
mcoker, line 158: Technically we don't support a disabled context selector, since the use case for the context selector is that it would always be enabled, so I don't think you need this block.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -118,6 +128,38 @@
...
+      width: auto;
+      color: var(--pf-c-context-selector__toggle--m-plain--Color);
+    }
+
+    &::before {
+      border: 0;
+    }
+
+    .pf-c-context-selector__toggle-button-icon {
+      line-height: var(--pf-c-context-selector__toggle--LineHeight);
```
mcoker, line 146: Doesn't look like you need this - the var `--pf-c-context-selector__toggle--LineHeight` doesn't exist either. 

Doesn't look like we need it in the options menu, either. Would you mind creating an issue to remove this line from the options menu component and add a "breaking change" label (since we're removing a variable). https://github.com/patternfly/patternfly/blob/005fa69c06705950c767354cd9d667d3d90a75bc/src/patternfly/components/OptionsMenu/options-menu.scss#L174

src/patternfly/components/ContextSelector/examples/context-selector.css
```diff
@@ -5,3 +5,7 @@
 #ws-core-c-context-selector-with-footer {
   min-height: 380px;
 }
+
+#ws-core-c-context-selector-plain-with-text {
+  min-height: 380px;
```
mcoker, line 10: Since this is the same as the other menu heights, we can combine all of the selectors to one - 
```
#ws-core-c-context-selector-basic, #ws-core-c-context-selector-with-footer, #ws-core-c-context-selector-plain-with-text {
  min-height: 380px;
}
```

src/patternfly/components/ContextSelector/context-selector-toggle.hbs
```diff
@@ -7,5 +7,10 @@
   {{#if context-selector-toggle--attribute}}
     {{{context-selector-toggle--attribute}}}
   {{/if}}>
+  {{#if context-selector-toggle--icon}}
```
mcoker, line 10: Doesn't look like this var is ever used - do you need it?

src/patternfly/components/ContextSelector/examples/context-selector.md
```diff
@@ -53,6 +53,53 @@ import './context-selector.css'
 {{/context-selector}}
 ```
 
+### Plain with text
```
mcoker, line 56: very nice!!

mcoker (changes requested):
Great job with this!! Left some comments for you to review. It looks like we're close to having `.pf-m-plain` supported on its own, so I left a couple of comments to see if we can get that variation working, though we don't need an example of it.

One last thing for that support is we'll need to support `.pf-m-plain` and `.pf-m-text` on the parent component element, too (`.pf-c-context-selector`) so that we can control the width of the component. A `.pf-c-context-selector.pf-m-plain:not(.pf-m-text)` context selector should set `--pf-c-context-selector--Width: auto;` so that the context selector doesn't have a fixed width so the toggle and menu can only be the size they need to be based on the contents.

src/patternfly/components/ContextSelector/context-selector-toggle.hbs
bitwiseor: Nope, I'll remove it! Left over from trying to get the plain variant with kebab icon.

bitwiseor:
I've adjusted my code on this branch based on feedback after initial creation of this PR

src/patternfly/components/ContextSelector/context-selector.scss
mcoker: If this isn't being used (looks like it isn't) can you remove this block and any associated vars?

src/patternfly/components/ContextSelector/context-selector.scss
mcoker: If this isn't being used, can you remove it, too?

src/patternfly/components/ContextSelector/context-selector.scss
mcoker: Looks like the other block for `.pf-c-context-selector__toggle-icon` is still in the CSS and doesn't need to be since you've duplicated it here.

src/patternfly/components/ContextSelector/context-selector.scss
mcoker: If this isn't being used, can you remove this block?

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -18,6 +18,13 @@
   --pf-c-context-selector__toggle--active--BorderBottomColor: var(--pf-global--active-color--100);
   --pf-c-context-selector__toggle--expanded--BorderBottomWidth: var(--pf-global--BorderWidth--md);
   --pf-c-context-selector__toggle--expanded--BorderBottomColor: var(--pf-global--active-color--100);
+  --pf-c-context-selector__toggle--m-plain--Color: var(--pf-global--Color--200);
+  --pf-c-context-selector__toggle--m-plain--hover--Color: var(--pf-global--Color--100);
+  --pf-c-context-selector__toggle--m-plain--disabled--Color: var(--pf-global--disabled-color--200);
```
mcoker, line 23: Looks like you removed the block that uses this var, but I think we'll still want that for the `.pf-m-plain:not(.pf-m-text)` variation, since the plain/icon/kebab variation has a different text/icon color when disabled.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -18,6 +18,13 @@
...
+  --pf-c-context-selector__toggle--m-plain--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-context-selector__toggle--m-plain--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-context-selector__toggle--m-plain--m-text--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-context-selector__toggle--m-plain--m-text--PaddingLeft: var(--pf-global--spacer--sm);
```
mcoker, line 27: These vars are great, but doesn't look like you're using them. Can you use them to redevine the toggle's left/right padding when it has `.pf-m-plain.pf-m-text`?

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -118,6 +130,30 @@
     }
   }
 
+  &.pf-m-plain {
+    --pf-c-context-selector__toggle--PaddingRight: var(--pf-c-context-selector__toggle--m-plain--PaddingRight);
+    --pf-c-context-selector__toggle--PaddingLeft: var(--pf-c-context-selector__toggle--m-plain--PaddingLeft);
+
+    display: inline-block;
```
mcoker, line 137: This should only apply to the non-text plain toggle - can you move it below to the `&:not(.pf-m-text)` selector below? It's causing the toggle to lose its flex layout, so it looks like this currently 
Screen Shot 2021-06-04 at 3 08 53 PM

mcoker (changes requested):
Aside from the plain disabled color needing to be updated, the plain variation looks great now! And the plain text variation needs a couple of small updates, too. 

Also I missed this before, but can you add a new entry to the component's documentation for the `.pf-m-plain.pf-m-text` modifier? We group the modifiers by the element they apply to - starting with the outer-most element (start with `.pf-c-context-selector`, then move to  `.pf-c-context-selector__toggle`), and we group modifiers that apply to the same element together. You can include this modifier before or after `.pf-m-active`.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -118,9 +130,35 @@
     }
   }
 
-  .pf-c-context-selector__toggle-icon {
-    margin-right: var(--pf-c-context-selector__toggle-icon--MarginRight);
-    margin-left: var(--pf-c-context-selector__toggle-icon--MarginLeft);
+  &.pf-m-plain {
+    --pf-c-context-selector__toggle--PaddingRight: var(--pf-c-context-selector__toggle--m-plain--m-text--PaddingRight);
```
mcoker, line 134: We typically try to have the vars match the CSS selectors, so we wouldn't want `--m-text` in the var name if it isn't in a `.pf-m-text` selector.

What you could do is move the `--pf-c-context-selector__toggle--m-plain--PaddingRight/Left` vars below up here, then write a nested `&.pf-m-text` selector and move these vars into that block.

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -118,9 +130,35 @@
...
+
+    &::before {
+      border: 0;
+    }
+
+    &:hover,
+    &:active,
+    &.pf-m-active,
+    &:focus,
+    .pf-c-context-selector.pf-m-expanded > & {
```
mcoker, line 159: These styles do not need to apply to `.pf-m-text`. You may be able to move this entire block into the `&:not(.pf-m-text)` selector above?

src/patternfly/components/ContextSelector/context-selector.scss
mcoker: Looks like this block is still here - I think it's safe to remove since I don't see an element with this class used anywhere in the component?

mcoker (changes requested):
Visually everything looks great! Just a couple of small updates to the code left to review.

mcoker (approved):
Fantastic job!! 🥳

mcoker:
Sorry missed one thing! Can you add an entry to the documentation on the examples page? Something like

```
.pf-m-plain.pf-m-text | .pf-c-context-selector__toggle | Modifies the context selector toggle for plain text styles.
```

PR #4097, opened 2021-05-25
feat(label): added editable label

### LINKED ISSUE #3741: Add editable Label Group
Add a new variant of the Label Group component to allow for creating new labels and adding them to the group. Here is an example of how this should work:

mattnolting:
closes #3741 

What's going on in this PR?
- Updates the `.pf-c-inline-edit` component to allow free-form editing
- Adds example of `.pf-c-inline-edit` in `.pf-c-label`
- Adds editable `pf-c-label-group` example

- This approach is based on gmail's add email address in new email approach which uses a `textarea` to the right of the email (label) list. 

- To achieve the desired interaction, using an `input` as the editable content element becomes very hard to style as the width of an input is defined by `size` rather than implicit width. To control, we'd have to calculate the width of the input's content and apply to size, which would likely lead to unexpected presentational changes.


gdoyle1 (changes requested):
Not sure if this change should be made here or not... but after I edit the label, when I click enter, I expected the label to turn back to non-editable... but it makes a line break instead. Can we update it to just turn it back to non-editable? @mattnolting

mcarrano (commented):
I'm a little confused by this example @mattnolting . https://patternfly-pr-4097.surge.sh/components/label/#editable Looks like there are two ways to edit a label? Either in context or active? Why is that the case? Happy to walk through this live if that will make more sense.

mattnolting:
@mcarrano @gdoyle1 I've updated the examples to be more accurate. We'll need JS/React to control label actions, but the examples now show various states.

gdoyle1:
@mattnolting I'm not sure if it's just me but when I try typing in at the editable state, my text sometimes moves to the search bar input (I'm not clicking on anything other than characters)

https://user-images.githubusercontent.com/30870144/119693532-5f2ace00-be1a-11eb-86c6-e89874c90a38.mov

mattnolting:
> @mattnolting I'm not sure if it's just me but when I try typing in at the editable state, my text sometimes moves to the search bar input (I'm not clicking on anything other than characters)

@gdoyle1 It's a workspace issue https://github.com/patternfly/patternfly-org/pull/2505

redallen:
@gdoyle1 Until this PR bumps theme-patternfly-org avoid pressing "s" or "/" in the label. @mattnolting found a way to outsmart @mturley and I...

mturley:
I knew that hotkey would be too good to be true somehow... Thanks for the fix @redallen

src/patternfly/components/Label/examples/Label.md
```diff
@@ -528,6 +528,30 @@ import './Label.css'
 {{/label}}
 ```
 
+### Editable label behavior must be handled with JavaScript.
+* `.pf-c-inline-edit__editable-text` OnClick event should:
```
mcoker, line 532:
```suggestion
* `.pf-c-inline-edit__editable-text` onClick event should:
```

src/patternfly/components/Label/examples/Label.md
```diff
@@ -528,6 +528,30 @@ import './Label.css'
...
+  * Remove `.pf-m-editable-active` from `.pf-c-label`
+  * Remove `contenteditable="true"` from `.pf-c-inline-edit__editable-text`
+
+### Editable
+``` hbs
+{{#> label label--id="editable-label" label--IsEditable="true" label--isRemovable="true" label--modifier="pf-m-blue"}}
+  Editable label
+{{/label}}
+
+{{#> label label--id="editable-label-active" label--IsEditable="true" label--IsEditableActive="true" label--modifier="pf-m-blue pf-m-active"}}
```
mcoker, line 550: do you need this?

```suggestion
{{#> label label--id="editable-label-active" label--IsEditable="true" label--IsEditableActive="true" label--modifier="pf-m-blue"}}
```

src/patternfly/components/Label/examples/Label.md
```diff
@@ -545,3 +569,5 @@ import './Label.css'
 | `.pf-m-red` | `.pf-c-label` | Modifies the label to have red colored styling. |
 | `.pf-m-purple` | `.pf-c-label` | Modifies the label to have purple colored styling. |
 | `.pf-m-cyan` | `.pf-c-label` | Modifies the label to have cyan colored styling. |
+| `.pf-m-editable` | `.pf-c-label` | Modifies label for editable styles. |
+| `.pf-m-active` | `.pf-c-editable` | Modifies editable label for active styles. |
```
mcoker, line 573:
```suggestion
| `.pf-m-editable-active` | `.pf-c-editable` | Modifies editable label for active styles. |
```

src/patternfly/components/Label/label.scss
```diff
@@ -247,6 +252,18 @@
     }
   }
 
+  &.pf-m-editable {
+    &.pf-m-editable-active {
```
mcoker, line 256: With the class name having editable in it, I wonder if these need to be nested or if we could style the classes independently.

mcoker (changes requested):

src/patternfly/components/Label/label.scss
```diff
@@ -134,6 +134,11 @@
   --pf-c-label__c-button--PaddingBottom: var(--pf-global--spacer--xs);
   --pf-c-label__c-button--PaddingLeft: var(--pf-global--spacer--sm);
 
+  // Editable
+  --pf-c-label--m-editable--m-active--BackgroundColor: transparent;
```
mcoker, line 138:
```suggestion
  --pf-c-label--m-editable--m-editable-active--BackgroundColor: transparent;
```

src/patternfly/components/Label/label.scss
```diff
@@ -134,6 +134,11 @@
...
+  --pf-c-label--m-editable--m-active__content--before--BorderWidth: 0;
```
mcoker, line 139:
```suggestion
  --pf-c-label--m-editable--m-editable-active__content--before--BorderWidth: 0;
```

src/patternfly/components/Label/label.scss
```diff
@@ -134,6 +134,11 @@
...
+  --pf-c-label--m-editable--m-active__content--before--BorderColor: transparent;
```
mcoker, line 140:
```suggestion
  --pf-c-label--m-editable--m-editable-active__content--before--BorderColor: transparent;
```

src/patternfly/components/Label/label.scss
```diff
@@ -247,6 +252,16 @@
     }
   }
 
+  &.pf-m-editable-active {
+    --pf-c-label--BackgroundColor: var(--pf-c-label--m-editable--m-active--BackgroundColor);
```
mcoker, line 256:
```suggestion
    --pf-c-label--BackgroundColor: var(--pf-c-label--m-editable--m-editable-active--BackgroundColor);
```

src/patternfly/components/Label/label.scss
```diff
@@ -247,6 +252,16 @@
...
+    --pf-c-label__content--before--BorderWidth: var(--pf-c-label--m-editable--m-active__content--before--BorderWidth);
```
mcoker, line 257:
```suggestion
    --pf-c-label__content--before--BorderWidth: var(--pf-c-label--m-editable--m-editable-active__content--before--BorderWidth);
```

src/patternfly/components/Label/label.scss
```diff
@@ -247,6 +252,16 @@
...
+    --pf-c-label__content--before--BorderColor: var(--pf-c-label--m-editable--m-active__content--before--BorderColor);
```
mcoker, line 258:
```suggestion
    --pf-c-label__content--before--BorderColor: var(--pf-c-label--m-editable--m-editable-active__content--before--BorderColor);
```

mcoker (changes requested):
This looks good, just need to rename a few vars. Though looks like some extra files were pulled in to this PR and shouldn't be included in the changed files.

mattnolting:
@mcoker updated

gdoyle1 (approved):
LGTM! Thank you @mattnolting

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -117,3 +139,20 @@
   margin-bottom: var(--pf-c-label-group__close--MarginBottom);
   margin-left: var(--pf-c-label-group__close--MarginLeft);
 }
+
+.pf-c-label-group__textarea {
```
mcoker, line 143: Can you add this one to the docs?

src/patternfly/components/Label/examples/Label.md
```diff
@@ -545,3 +569,5 @@ import './Label.css'
 | `.pf-m-red` | `.pf-c-label` | Modifies the label to have red colored styling. |
 | `.pf-m-purple` | `.pf-c-label` | Modifies the label to have purple colored styling. |
 | `.pf-m-cyan` | `.pf-c-label` | Modifies the label to have cyan colored styling. |
+| `.pf-m-editable` | `.pf-c-label` | Modifies label for editable styles. |
+| `.pf-m-editable-active` | `.pf-c-editable` | Modifies editable label for active styles. |
```
mcoker, line 573:
```suggestion
| `.pf-m-editable-active` | `.pf-c-label.pf-m-editable` | Modifies editable label for active styles. |
```

mcoker (changes requested):
looks great! just a couple of docs updates.

mcarrano (approved):
> @mcarrano @gdoyle1 I've updated the examples to be more accurate. We'll need JS/React to control label actions, but the examples now show various states.

Thanks @mattnolting . This makes more sense now.

mcoker (approved):
This is really cool! Great work 🎉

Issue #4101, opened 2021-05-25
Flyout menus still trigger on hover when item is disabled 


mcoker:
If a menu item is disabled and has a flyout, the flyout menu still shows when you hover over the item.

<img width="999" alt="Screen Shot 2021-05-25 at 4 07 18 PM" src="https://user-images.githubusercontent.com/35148959/119568680-5e8b2c80-bd73-11eb-8a9c-4b5477df9a94.png">

PR #4102, opened 2021-05-25
fix(menu): disabled pointer events on disabled list item

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4101

@mattnolting can you think of a reason we wouldn't just disable pointer events on the top level `li` instead of disabling them on the children?


mattnolting (approved):
Perfect!

mattnolting:
> @mattnolting can you think of a reason we wouldn't just disable pointer events on the top level `li` instead of disabling them on the children?

@mcoker Yes, I think this is a better approach

PR #4108, opened 2021-05-26
feat(dropdowns): added dropdown, context selector full height mods

mattnolting:
closes #4109

PR #4110, opened 2021-05-26
feat(masthead): added dropdown and context selector styling

### LINKED ISSUE #4105: Update masthead context selector / dropdown border colors
Masthead context selector / dropdown border color spec

Screen Shot 2021-05-26 at 10 57 07 AM

mattnolting:
closes #4105

### DIFF
--- a/src/patternfly/components/Masthead/masthead.scss
+++ b/src/patternfly/components/Masthead/masthead.scss
@@ -110,16 +110,19 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --pf-c-masthead--m-light-200__main--BorderBottomColor: var(--pf-global--BorderColor--100);
 
   // Dropdown
-  --pf-c-masthead--c-dropdown__toggle--BorderTopColor: transparent;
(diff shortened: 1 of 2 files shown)


mcoker:
Looks like there is a conflict in this PR.

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -173,6 +185,22 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
     display: flex;
     align-items: center;
   }
+
+  .pf-c-dropdown__toggle {
```
mcoker, line 189: If these style changes only apply to the full height variation, do you think it makes sense to add `.pf-m-full-height` to the selectors?

Also curious what you think about making these border changes the default in the component variations, assuming if you're using a full height toggle, you don't want the top border, and maybe you want the larger bottom border width?

Can I see these changes in use in this PR or do we need to merge https://github.com/patternfly/patternfly/pull/4108 first? This is what the current demo looks like

Screen Shot 2021-06-11 at 2 16 37 PM

mcoker:
implemented in https://github.com/patternfly/patternfly/pull/4211

Issue #4111, opened 2021-05-26
Wizard expandable nav item toggle icon rotation/padding issue


mcoker:
The nav item toggle icon has left/right padding, that when rotated 90deg, becomes top/bottom padding. That padding should stay left/right padding.

PR #4112, opened 2021-05-26
fix(wizard): fixed expandable nav item toggle rotation/padding issue

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4111

just separates the icon element that rotates from the toggle element that has padding for the toggle icon


mattnolting (approved):
LPTM 🏑

PR #4118, opened 2021-06-01
fix(release-notes): added 2021.07 release notes, updated screenshots

mcoker:


nicolethoen:
@mcoker this is good - can we just push out the date like maybe two days since I bet the release wont be complete until then at the earliest :)

mcoker:
@nicolethoen updated

nicolethoen (approved):

Issue #4053, opened 2021-05-07
Table + Toolbar: Search bar floats above the header when scrolled down and in focus


imjoyjean:
**Description of issue**:
When the Search bar of the table is in focus and you scroll down the table, the search bar floats on top of the tabs and header area. May need some css fixes here. 

**Screenshot**: 
![image](https://user-images.githubusercontent.com/57908474/117467424-4913a700-af19-11eb-8740-e811f98c84ee.png)

PR #4128, opened 2021-06-04
fix(page): increase sticky page section z-index

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4053

Uses the z-index higher than our menus, but lower than overlays (backdrop, modal, etc)


bitwiseor (approved):
Looks good to me! 💯

mattnolting (approved):
L💯 TM!

Issue #4131, opened 2021-06-07
Content component visited link color


mcoker:
The color change isn't working - the browser restricts properties that can be defined for the `:visited` pseudo class, and looks like setting a custom property will not work.

PR #4132, opened 2021-06-07
fix(content): changed visited to declare color property

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4131

also wrapped the link in a `<p>` to improve the spacing and removed the value from `href`, which is how [MDN does it](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited)


srambach (commented):
LGTM 👍

srambach (approved):
LGTM 👍

mcoker:
@tlabaj changes in this PR are needed for the visited color to work properly.

Issue #4092, opened 2021-05-21
Label with isOverflowLabel does not support color variants


jerolimov:
## Describe the issue. What is the expected and unexpected behavior?

The [Label](https://www.patternfly.org/v4/components/label) supports a button variant with the prop `isOverflowLabel`. But when enable it, it breaks the `color` prop.

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**

tl;dr: https://codesandbox.io/s/patternfly-isoverflow-color-label-issue-iqej8

(The Label looks fine on CodeSandbox, but the color issue could be checked here as well.)

**When using the label in different variants with a color the label looks like expected:**

Code:
```tsx
import { Label } from '@patternfly/react-core';

export default function LabelIssue() {
  let i = 1;
  return (
    <div>
      <Label>Label {i++}</Label>
      <Label variant="filled">Label {i++}</Label>
      <Label variant="outline">Label {i++}</Label>

      <Label color="purple">Label {i++}</Label>
      <Label color="purple" variant="filled">Label {i++}</Label>
      <Label color="purple" variant="outline">Label {i++}</Label>
    </div>
  );
}
```

Visual result:
![image](https://user-images.githubusercontent.com/139310/119141376-77c76c80-ba45-11eb-84dd-8ca8058a6564.png)

No hover effect (as expected):
![Peek 2021-05-21 15-11](https://user-images.githubusercontent.com/139310/119142445-cb868580-ba46-11eb-8544-1a078fc7b66f.gif)

**But when adding `isOverflowLabel` and `onClick` prop (only the first one changes the UI, it also switches the internal component from a `span` to a `button`) the colors are missed:**

Code:
```tsx
import { Label } from '@patternfly/react-core';

export default function LabelIssue() {
  const onClick = (event: React.MouseEvent<HTMLButtonElement>) => console.log('onClick', event);
  let i = 1;
  return (
    <div>
      <Label isOverflowLabel onClick={onClick}>Label {i++}</Label>
      <Label isOverflowLabel onClick={onClick} variant="filled">Label {i++}</Label>
      <Label isOverflowLabel onClick={onClick} variant="outline">Label {i++}</Label>

      <Label isOverflowLabel onClick={onClick} color="purple">Label {i++}</Label>
      <Label isOverflowLabel onClick={onClick} color="purple" variant="filled">Label {i++}</Label>
[...]

PR #4136, opened 2021-06-08
chore(label): replaced overflow with truncate in truncation examples

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4092


mcarrano (approved):
Looks good. Thanks @mcoker .

PR #4137, opened 2021-06-08
fix(overflow-menu): added aria-expanded to examples

### LINKED ISSUE #3773: Overflow menu examples missing aria-expanded
The examples are missing `aria-expanded`. It looks like there is some logic for this in https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/OverflowMenu/overflow-menu-control.hbs referencing `overflow-menu-dropdown-button-aria-expanded` but doesn't appear to be working.

bitwiseor:
fixes #3773 

also added aria-expanded to the toolbar and tabs examples that use overflow menus

### DIFF
--- a/src/patternfly/demos/Tabs/templates/tabs--table-overflow-menu.hbs
+++ b/src/patternfly/demos/Tabs/templates/tabs--table-overflow-menu.hbs
@@ -1,5 +1,5 @@
 {{#> overflow-menu dropdown-menu--modifier="pf-m-align-right" overflow-menu--id=tabs--table-overflow-menu--id}}
-  {{#> overflow-menu-control overflow-menu-button--aria-label="Dropdown for tabs table" overflow-menu-dropdown-button-aria-expanded="false"}}
+  {{#> overflow-menu-control overflow-menu-button--aria-label="Dropdown for tabs table" overflow-menu-dropdown-button-aria-expanded="true"}}
     {{#> overflow-menu-dropdown-item}}
       Action Link
     {{/overflow-menu-dropdown-item}}
--- a/src/patternfly/demos/Tabs/templates/tabs--table-overflow-menu.hbs
+++ b/src/patternfly/demos/Tabs/templates/tabs--table-overflow-menu.hbs
@@ -1,7 +1,15 @@
 {{#> overflow-menu dropdown-menu--modifier="pf-m-align-right" overflow-menu--id=tabs--table-overflow-menu--id}}
-  {{#> overflow-menu-control overflow-menu-button--aria-label="Dropdown for tabs table" overflow-menu-dropdown-button-aria-expanded="true"}}
-    {{#> overflow-menu-dropdown-item}}
-      Action Link
-    {{/overflow-menu-dropdown-item}}
-  {{/overflow-menu-control}}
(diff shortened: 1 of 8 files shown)


mcoker (changes requested):
This looks really good! Though looks like this example should be `aria-expanded="true"`? https://patternfly-pr-4137.surge.sh/components/toolbar#expanded-elements

Also there is an expanded overflow menu dropdown here that needs `aria-expanded="true"` too? https://patternfly-pr-4137.surge.sh/components/tabs/html-demos/tables-and-tabs/

mcoker (changes requested):
Thinking about this some more, I think we could cut down on the code a bit by setting `overflow-menu-dropdown-button-aria-expanded` by default, then only specifying it when the value needs to be something other than the default. 

One way to do that would be to use the `{{#> wrapper}}` trick we discussed, and set  `overflow-menu-dropdown-button-aria-expanded="true"` on that block (since it looks like `aria-expanded="true"` is the most common) - then only set `overflow-menu-dropdown-button-aria-expanded` explicitly when the value is false.

You could add `{{#> wrapper}}` around the markup in `overflow-menu.hbs`, then remove any instance of `overflow-menu-dropdown-button-aria-expanded="true"` in the examples/demos hbs. 

Also looking at `overflow-menu-control.hbs`, there is already a var in use called `overflow-menu-button--aria-label` on that button - I think we could rename `overflow-menu-dropdown-button-aria-expanded` to `overflow-menu-button--aria-expanded` to match.

src/patternfly/demos/Tabs/templates/tabs--table-overflow-menu.hbs
```diff
@@ -1,7 +1,15 @@
 {{#> overflow-menu dropdown-menu--modifier="pf-m-align-right" overflow-menu--id=tabs--table-overflow-menu--id}}
-  {{#> overflow-menu-control overflow-menu-button--aria-label="Dropdown for tabs table"}}
-    {{#> overflow-menu-dropdown-item}}
-      Action Link
-    {{/overflow-menu-dropdown-item}}
-  {{/overflow-menu-control}}
+  {{#if dropdown--IsExpanded}}
```
mcoker, line 2: This looks good. My question about moving the `{{#if/else}}` block in the `toolbar-overflow-menu-example.hbs` template is basically asking if we can update that template's conditional to just wrap `{{#> overflow-menu-control}}` as you've done here so we have less repeated code.

mcoker (changes requested):
Looking great, just a few comments!

mcoker (changes requested):
Looks like there is one last spot on the tables and tabs demos - the toolbar has an overflow menu that is collapsed but has `aria-expanded="true"`

https://patternfly-pr-4137.surge.sh/components/tabs/html-demos/tables-and-tabs/
https://patternfly-pr-4137.surge.sh/components/tabs/html-demos/tables-and-tabs-auto-width-tabs/

I think you'll find that in https://github.com/patternfly/patternfly/blob/8c2993c87074407f6ebd74b6b6591dfbced907b2/src/patternfly/components/Toolbar/toolbar--template.hbs#L75-L79

mcoker (approved):
Looks great!! 🏆

Issue #4139, opened 2021-06-09
Helper text - rename invalid state to error


mcoker:
This is consistent with our existing use of helper text in form and date picker.

PR #4140, opened 2021-06-09
fix(helper-text): renamed invalid state to error

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4139


srambach (approved):
LGTM 👍

PR #4143, opened 2021-06-10
feat(expandable-section): added large variation

### LINKED ISSUE #3952: Add disclosure panel variant for Expandable section
Add a new Disclosure panel variant for expandable sections. This will be optimized for use to display marketing oriented content. See designs here: https://www.sketch.com/s/f6a52c25-8657-4680-9d3e-4c09131e6ed0/p/disclosure

srambach:
This adds a display-lg variant to the expandable section. 
Note: pending font size and max-width confirmation from design
@ajacobs21e 

fixes #3952

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section.scss
+++ b/src/patternfly/components/ExpandableSection/expandable-section.scss
@@ -23,7 +23,7 @@
 
   // Content
   --pf-c-expandable-section__content--MarginTop: var(--pf-global--spacer--md);
-  --pf-c-expandable-section__content--MaxWidth: inherit;
+  --pf-c-expandable-section__content--MaxWidth: auto;
 
   // Large
   --pf-c-expandable-section--m-display-lg--BoxShadow: var(--pf-global--BoxShadow--md);
@@ -37,7 +37,6 @@
   --pf-c-expandable-section--m-display-lg__content--PaddingLeft: var(--pf-global--spacer--lg);
   --pf-c-expandable-section--m-display-lg__content--MaxWidth: var(--pf-global--breakpoint--md);
   --pf-c-expandable-section--m-display-lg--after--BackgroundColor: transparent;
-  --pf-c-expandable-section--m-display-lg--after--Top: 0;
   --pf-c-expandable-section--m-display-lg--after--Width: var(--pf-global--BorderWidth--lg);
   --pf-c-expandable-section--m-display-lg--m-expanded--after--BackgroundColor: var(--pf-global--primary-color--100);
 
@@ -115,7 +114,6 @@
 
 .pf-c-expandable-section__content {
   max-width: var(--pf-c-expandable-section__content--MaxWidth);
-  padding-top: var(--pf-c-expandable-section__content--PaddingTop);
   padding-right: var(--pf-c-expandable-section__content--PaddingRight);
   padding-bottom: var(--pf-c-expandable-section__content--PaddingBottom);
   padding-left: var(--pf-c-expandable-section__content--PaddingLeft);
--- a/src/patternfly/components/ExpandableSection/expandable-section.scss
+++ b/src/patternfly/components/ExpandableSection/expandable-section.scss
@@ -26,7 +26,7 @@
   --pf-c-expandable-section__content--MaxWidth: auto;
 
   // Limit Width
-  --pf-c-expandable-section--m-limit-width__content--MaxWidth: #{pf-size-prem(700px)};
+  --pf-c-expandable-section--m-limit-width__content--MaxWidth: #{pf-size-prem(750px)};
 
   // Large
   --pf-c-expandable-section--m-display-lg--BoxShadow: var(--pf-global--BoxShadow--md);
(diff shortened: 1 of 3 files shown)


src/patternfly/components/ExpandableSection/expandable-section.scss
```diff
@@ -23,16 +23,60 @@
...
+  --pf-c-expandable-section--m-display-lg--BoxShadow: var(--pf-global--BoxShadow--md);
+  --pf-c-expandable-section--m-display-lg__toggle--PaddingTop: var(--pf-global--spacer--md);
+  --pf-c-expandable-section--m-display-lg__toggle--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-expandable-section--m-display-lg__toggle--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-expandable-section--m-display-lg__toggle--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-expandable-section--m-display-lg__content--MarginTop: 0;
+  --pf-c-expandable-section--m-display-lg__content--PaddingRight: var(--pf-global--spacer--lg);
+  --pf-c-expandable-section--m-display-lg__content--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-expandable-section--m-display-lg__content--PaddingLeft: var(--pf-global--spacer--lg);
+  --pf-c-expandable-section--m-display-lg__content--MaxWidth: var(--pf-global--breakpoint--md);
```
mcoker, line 38: I'm not sure that we would want to use a breakpoint var here (we have only used breakpoint vars in `@media` queries). For example, if a user updated the breakpoint value to be something else, would we expect the content width in this component to match? This max-width likely won't match an actual media query breakpoint on the page since the expandable section will likely be in a page section and/or card or something which has its own padding, meaning the viewport media query will take effect before the expandable content max-width (set to the same breakpoint size) is reached.

If it's just supposed to limit the width to improve readability, we have a similar [`max-width` set on the form component](https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/Form/form.scss#L11) (it's `500px`) available as an opt-in as `.pf-m-limit-width`. I wonder if we should do something like that? And if we might benefit from a global variable we use for things like this for consistency.

wdyt @ajacobs21e @mcarrano @srambach?

src/patternfly/components/ExpandableSection/expandable-section.scss
```diff
@@ -23,16 +23,60 @@
...
+  --pf-c-expandable-section--m-display-lg--after--BackgroundColor: transparent;
+  --pf-c-expandable-section--m-display-lg--after--Top: 0;
```
mcoker, line 40: doesn't look like this is used

mcoker (changes requested):
Looks great @srambach!!! Left a few comments for review.

src/patternfly/components/ExpandableSection/expandable-section.scss
srambach: My understanding was that it was for readability. I've moved the `max-width` from the large variant to the `.pf-m-limit-width` modifier with a value of 750px (since that's what is in the spec).

mcoker (commented):
Code looks great to me.

I see @ajacobs21e said the title and CTA are optional, but I'm still not totally clear if we should add a title and footer element to the component for consistency, or if that's just totally up to the user to put whatever they want in there. I'm assuming the latter, in which case I might just put regular text in the body in our example so it's clear the title and everything is not part of the component.

Also want to confirm with @ajacobs21e and @mcarrano that the `pf-m-limit-width` variation as an opt-in to limit the content width is the way to go, or if it should be the default. 

And do we want to call these examples "disclosure" or "large"? In the accordion and tile, we call the `pf-m-display-lg` variation "large", but in the button we call it "CTA", so we aren't consistent in naming the examples "large".

mcarrano:
@mcoker looks like the name "Disclosure" comes from the original request here: https://github.com/patternfly/patternfly-design/issues/874  I'd like to stick with that since it's about more than just larger text/spacing but also adds a border to give it a different look from the basic expandable section.

I guess that I also agree with making the width limitation an opt-in feature if that's consistent with what we do elsewhere.

ajacobs21e:
@mcarrano do we have other instances of components that we limit the width of either the component or the text within a component?

mcarrano:
Good question @ajacobs21e . Do you know @mcoker ?  I think that in most cases the text fills the width of the container, whatever that is.

mcoker:
@ajacobs21e 

The form component has a `.pf-m-limit-width` variation that limits the form width to 500px - https://www.patternfly.org/v4/components/form#limit-width

Not a component, but a page section has an opt-in to limit its width to 1200px. You can see that on our page component demos - https://www.patternfly.org/v4/components/page/html-demos/

mcoker:
@srambach chatted with @mcarrano and confirmed we can leave the `.pf-m-limit-width` modifier as it is, but that we do want to update the disclosure examples to just include the same body text as the rest of the expandable section examples, so it's clear the title, button, etc are not part of that variation of the component.

mcoker (approved):
🥳 🎉

PR #4145, opened 2021-06-11
feat(login): adds a password strength demo

### LINKED ISSUE #4018: Login page - password strength meter demo
Create a new Login page demo that shows show to implement a password strength meter according to the designs here: https://marvelapp.com/prototype/6jehce7/screen/78511266

srambach:
Adds an example login page showing interactive password strength using helper text

Fixes #4018

### DIFF
--- a/src/patternfly/components/Login/examples/Login.md
+++ b/src/patternfly/components/Login/examples/Login.md
@@ -191,7 +191,6 @@ wrapperTag: div
   {{/login-container}}
 {{/login}}
 ```
-
 ## Documentation
 ### Usage
 | Class | Applied to | Outcome |
--- a/src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
+++ b/src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
@@ -10,14 +10,13 @@ section: demos
 ```hbs
   {{#> form form--id="password-strength-demo--initial"}}
     {{#> form-group}}
-      {{#> form-group-label}}
-        {{#> split split--modifier="pf-m-gutter"}}
-          {{#> split-item split-item--modifier="pf-m-fill"}}
-            {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
-          {{/split-item}}
-          {{#> split-item}}
-          {{/split-item}}
-        {{/split}}
+      {{#> form-group-label form-group-label-info="true"}}
+        {{#> form-group-label-main}}
+              {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+        {{/form-group-label-info}}
       {{/form-group-label}}
       {{#> form-group-control}}
         {{#> input-group}}
@@ -52,14 +51,13 @@ section: demos
 ```hbs
   {{#> form  form--id="password-strength-demo--invalid"}}
     {{#> form-group}}
-      {{#> form-group-label}}
-        {{#> split split--modifier="pf-m-gutter"}}
-          {{#> split-item split-item--modifier="pf-m-fill"}}
-            {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}} Password{{/form-label}}
-          {{/split-item}}
-          {{#> split-item}}
-          {{/split-item}}
-        {{/split}}
+      {{#> form-group-label form-group-label-info="true"}}
+        {{#> form-group-label-main}}
+              {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+        {{/form-group-label-info}}
       {{/form-group-label}}
       {{#> form-group-control}}
         {{#> input-group}}
@@ -94,20 +92,19 @@ section: demos
 ```hbs
   {{#> form  form--id="password-strength-demo--weak"}}
     {{#> form-group}}
-      {{#> form-group-label}}
-        {{#> split split--modifier="pf-m-gutter"}}
-          {{#> split-item split-item--modifier="pf-m-fill"}}
-            {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}} Password{{/form-label}}
-          {{/split-item}}
-          {{#> split-item}}
-            {{#> helper-text}}
-              {{#> helper-text-item helper-text-item--modifier="pf-m-error"}}
-                {{> helper-text-item-icon helper-text-item-icon--type="exclamation-circle"}}
-                {{#> helper-text-item-text}}Weak{{/helper-text-item-text}}
-              {{/helper-text-item}}
-            {{/helper-text}}
-          {{/split-item}}
-        {{/split}}
+      {{#> form-group-label form-group-label-info="true"}}
+        {{#> form-group-label-main}}
+              {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+          {{#> helper-text}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-error"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="exclamation-circle"}}
+              {{#> helper-text-item-text}}Weak{{/helper-text-item-text}}
+            {{/helper-text-item}}
+          {{/helper-text}}
+        {{/form-group-label-info}}
       {{/form-group-label}}
       {{#> form-group-control}}
         {{#> input-group}}
@@ -142,20 +139,19 @@ section: demos
 ```hbs
   {{#> form  form--id="password-strength-demo--strong"}}
     {{#> form-group}}
-      {{#> form-group-label}}
-        {{#> split split--modifier="pf-m-gutter"}}
-          {{#> split-item split-item--modifier="pf-m-fill"}}
-            {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}} Password{{/form-label}}
-          {{/split-item}}
-          {{#> split-item}}
-            {{#> helper-text}}
-              {{#> helper-text-item helper-text-item--modifier="pf-m-success"}}
-                {{> helper-text-item-icon helper-text-item-icon--type="check-circle"}}
-                {{#> helper-text-item-text}}Strong{{/helper-text-item-text}}
-              {{/helper-text-item}}
-            {{/helper-text}}
-          {{/split-item}}
-        {{/split}}
+      {{#> form-group-label form-group-label-info="true"}}
+        {{#> form-group-label-main}}
+          {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+          {{#> helper-text}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-success"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="check-circle"}}
+              {{#> helper-text-item-text}}Strong{{/helper-text-item-text}}
+            {{/helper-text-item}}
+          {{/helper-text}}
+        {{/form-group-label-info}}
       {{/form-group-label}}
       {{#> form-group-control}}
         {{#> input-group}}
@@ -188,4 +184,4 @@ section: demos
 
 ## Documentation
 
-This demo implements a password strength meter. It uses multiple helper text items in the form helper text area below the input field to indicate whether the password meets validity criteria. Once the validity criteria are met, it also places helper text above the input field to indicate the strength of the password.
+This demo implements a password strength meter. It uses multiple helper text items in the form helper text area below the input field to indicate whether the password meets validity criteria. Once the validity criteria are met, it also places helper text in the info area above the input field to indicate the strength of the password.
--- a/src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
+++ b/src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
@@ -61,7 +61,7 @@ section: demos
       {{/form-group-label}}
       {{#> form-group-control}}
         {{#> input-group}}
-          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$RedHat78" placeholder="Password"')}}
+          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'required type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$RedHat78" placeholder="Password"')}}
           {{/form-control}}
           {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Show password button for password input"'}}
             <i class="fas fa-eye-slash" aria-hidden="true"></i>
--- a/src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
+++ b/src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
@@ -19,29 +19,31 @@ section: demos
           {{/split-item}}
         {{/split}}
       {{/form-group-label}}
-      {{#> input-group}}
-          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="" placeholder="Password"')}}
-        {{/form-control}}
-        {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Show password button for password input"'}}
-          <i class="fas fa-eye-slash" aria-hidden="true"></i>
-        {{/button}}
-      {{/input-group}}
-      {{#> form-helper-text form-helper-text--type="div"}}
-        {{#> helper-text helper-text--type="ul" helper-text-item--type="li"}}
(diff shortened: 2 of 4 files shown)


src/patternfly/components/Login/examples/Login.md
```diff
@@ -192,6 +192,83 @@ wrapperTag: div
...
+              {{/button}}
+            {{/input-group}}
+            {{#> helper-text helper-text--modifier="pf-u-mt-sm"}}
+              {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-success"}}
+                {{> helper-text-item-icon helper-text-item-icon--type="check"}}
+                {{#> helper-text-item-text}}Must be at least 14 characters{{/helper-text-item-text}}
+              {{/helper-text-item}}
+              {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-invalid"}}
+                {{> helper-text-item-icon helper-text-item-icon--type="times"}}
+                {{#> helper-text-item-text}}Cannot contain any variation of the work "redhat"{{/helper-text-item-text}}
```
mcoker, line 243:
```suggestion
                {{#> helper-text-item-text}}Cannot contain any variation of the word "redhat"{{/helper-text-item-text}}
```

mcoker (changes requested):
This looks great!! A couple of questions for @mcarrano 

* This is built as a demo - there were no enhancements to the login component to support this PR. The way the helper text that is adjacent to the label when it's above the text input is using a split layout in the demo. We may get variation in use in products if the way we put helper text adjacent to the label isn't built into the form or login component itself. This would also need to be built differently if a horizontal form were used. This seems OK, but I wanted to confirm whether the intent was for it to be purely a demo or if we want to build support for helper text by the form label into the form or login component.
* We typically do not use beta components in other component examples/demos, so we don't give the impression the component is ready to use. It's definitely a login page demo, but I wonder if it belongs under the helper text demos since it's beta?

@srambach the ultimate goal for the helper text component is to replace the current helper text elements we have in individual components. In this demo, the helper text component is placed after the text input with a margin from a utility class. I think we could go ahead and use an element in the form that has the proper spacing for helper text in relation to a form field, and place the helper text component inside of it. Looking at the [`.pf-c-form__helper-text` element](https://github.com/patternfly/patternfly/blob/aab4c0135acb584584edb0f0607bcef0e442091d/src/patternfly/components/Form/form.scss#L277-L303), that seems generic enough (maybe except for the `font-size` declaration) to be able to support putting the helper text component inside. The `font-size` won't conflict here though since the helper text component's `font-size` is the same. WDYT?

srambach:
@mcoker there seems to be a problem here because `.pf-c-form__helper-text` creates a `<p>` tag, but you cannot nest the `pf-c-helper-text` with its `<div>`s inside a p tag.

mcarrano:
> This is built as a demo - there were no enhancements to the login component to support this PR. The way the helper text that is adjacent to the label when it's above the text input is using a split layout in the demo. We may get variation in use in products if the way we put helper text adjacent to the label isn't built into the form or login component itself. This would also need to be built differently if a horizontal form were used. This seems OK, but I wanted to confirm whether the intent was for it to be purely a demo or if we want to build support for helper text by the form label into the form or login component.

I think this is OK. I had thought at times whether we should define a unique Password component that includes all of these capabilities, but seems like that's overkill. Do you agree @mcoker ? I think we can provide guidelines around usage of the password strength in the the design guidelines to drive consistent usage.

> We typically do not use beta components in other component examples/demos, so we don't give the impression the component is ready to use. It's definitely a login page demo, but I wonder if it belongs under the helper text demos since it's beta?

I see your point, @mcoker , but I don't think anyone would look for this there. Is it possible to mark this specific example/demo as Beta since it uses a beta component?

ajacobs21e:
This isn't part of the login page. It would be for an account creation form, not a login form once the password is already created. 
The strength on the top right (weak, etc) shouldn't be shown if one of the password requirements below has failed. Only after all are checked would we then show the password strength, because it doesn't matter if it's the strongest password in the world if it doesn't meet a requirement.

mcarrano:
That's a great point @ajacobs21e . I had wrongly been thinking of this as associated with the password field on Login. In that case, I think we should have two separate demos. One that is associated with the Helper Text component and shows how to use that for field validation and separately to demonstrate how to add password strength. I'm not sure where we add the later. @mcoker is this just a demo of how to add this to a form field? Do you think we would create something unique called Password strength and add to the Demo section?

mcoker:
@mcarrano yep, we're just demoing how to use the helper text with a form field. I think the password strength demo could be a unique demo, especially since it's tied to multiple components. And since it's in it's own section, we could create a few examples showing the various states.

mcoker:
@srambach looking at https://www.patternfly.org/v4/components/form, all of the helper text is a `<div>` and it's coming from `<FormGroup>` - https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Form/FormGroup.tsx.

Though there is a `<FormHelperText>` component that generates a `<p>` - https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Form/FormHelperText.tsx

If the preferred use of helper text in a form is to use the `helperText*` props on `<FormGroup>`, I'm in favor of changing `<FormHelperText>` to a `<div>` to allow for a more flexible element to put helper text in, in case users are using it to define their helper text. It seems like it would be good to be consistent in the element we use in `<FormGroup>` and `<FormHelperText>`.

For the purposes of this PR, I think it's fair to assume `.pf-c-form__helper-text` can be either a `<p>` or `<div>` and we could use a `<div>` here.

I'm also curious from @kmcfaul, @tlabaj, and @jessiehuff on that change, as well as how they would envision using the helper text component to add helper text to a form field.

tlabaj:
@mcoker Right now the react component accepts any React node for the `helperText` prop.  The consumer can pass a string if the wish.  In that case the helper text would not be wrapped in `<p>` tag. ([see example here](https://www.patternfly.org/v4/components/form#basic)).
If they they pass a `<FormHelperText>` node, they would get the `<p>` tag, ([see example here](https://www.patternfly.org/v4/components/form#invalid))

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Examples
+
+### Initial state
+```hbs
+  {{#> form}}
+    {{#> form-group}}
+      {{#> split split--modifier="pf-m-gutter"}}
+        {{#> split-item split-item--modifier="pf-m-fill"}}
+          {{#> form-label form-label--attribute='for="invalid-login-demo-form-password"' required="true"}}Password{{/form-label}}
```
mcoker, line 15: One thing we do a lot is give the `{{#> form}}` block a `form--id="helper-text-invalid"`, then do `form-label--attribute=(concat 'for="' form--id '-password"')`, so we can use that anywhere we need a unique string that's used multiple places in an example, then use that same code in each example with a unique `form--id`.

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,175 @@
...
+        {{/split-item}}
+        {{#> split-item}}
+        {{/split-item}}
+      {{/split}}
+      {{#> input-group}}
+        {{#> form-control controlType="input" input="true"  form-control--attribute='type="search" id="invalid-login-demo-form-password" name="invalid-login-demo-form-password" aria-label="Password input" value="" placeholder="Password"'}}
```
mcoker, line 21: `type="search"` should be `type="text"`?

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,175 @@
...
+            {{> helper-text-item-icon helper-text-item-icon--type="check"}}
+            {{#> helper-text-item-text}}Must include at least 3 of the following: lowercase letters, uppercase letters, numbers, symbols{{/helper-text-item-text}}
+          {{/helper-text-item}}
+        {{/helper-text}}
+      {{/form-helper-text}}
+    {{/form-group}}
+  {{/form}}
+```
+
+### Valid, but weak password
```
mcoker, line 85: nit - just to match the next example

```suggestion
### Valid, weak password
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,175 @@
...
+section: demos
+---
+
+## Examples
+
+### Initial state
+```hbs
+  {{#> form}}
+    {{#> form-group}}
+      {{#> split split--modifier="pf-m-gutter"}}
```
mcoker, line 13: Looking at this split a little more, I don't know that we can do this in react. The `<FormGroup>` component takes a `label` for the label, and its children are passed into the form group control element. I'm wondering if we need to add a new form element for info to be displayed beside the label like this, and potentially support for what that would look like in a horizontal form layout. wdyt?

mcoker (commented):
Just a few small comments, though a bigger issue is the use of the split to have the helper text beside the form label. Left a comment about that. If we can't figure out a solution, I'd be fine with just omitting that part of the demo (the text beside the label for the strength text) and following up with that next sprint.

mcarrano (approved):
This looks god to me now from a design perspective. Thanks @srambach !

jessiehuff (commented):
This looks pretty good! My only concern is that I believe this helper text will end up being dynamic so screen reader users will need to be updated with this information the same way that sighted users are. I imagine that it would follow the same pattern as our [invalid form example](https://www.patternfly.org/v4/components/form#invalid). The only difference that I wonder about is the password strength text.

srambach:
> This looks pretty good! My only concern is that I believe this helper text will end up being dynamic so screen reader users will need to be updated with this information the same way that sighted users are. I imagine that it would follow the same pattern as our [invalid form example](https://www.patternfly.org/v4/components/form#invalid). The only difference that I wonder about is the password strength text.

@jessiehuff do you think that `aria-live="polite"` should be added to the info by default here, or do you think that should ultimately be handled by using dynamic helper text?

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+        {{/input-group}}
+        {{#> form-helper-text form-helper-text--type="div"}}
+          {{#> helper-text helper-text--type="ul" helper-text-item--type="li"}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-indeterminate"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="minus"}}
+              {{#> helper-text-item-text}}Must be at least 14 characters{{/helper-text-item-text}}
+            {{/helper-text-item}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-indeterminate"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="minus"}}
+              {{#> helper-text-item-text}}Cannot contain any variation of the work "redhat"{{/helper-text-item-text}}
```
mcoker, line 37:
```suggestion
              {{#> helper-text-item-text}}Cannot contain any variation of the word "redhat"{{/helper-text-item-text}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+        {{/input-group}}
+        {{#> form-helper-text form-helper-text--type="div"}}
+          {{#> helper-text helper-text--type="ul" helper-text-item--type="li"}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-success"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="check"}}
+              {{#> helper-text-item-text}}Must be at least 14 characters{{/helper-text-item-text}}
+            {{/helper-text-item}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-error"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="times"}}
+              {{#> helper-text-item-text}}Cannot contain any variation of the work "redhat"{{/helper-text-item-text}}
```
mcoker, line 78:
```suggestion
              {{#> helper-text-item-text}}Cannot contain any variation of the word "redhat"{{/helper-text-item-text}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+        {{/input-group}}
+        {{#> form-helper-text form-helper-text--type="div"}}
+          {{#> helper-text helper-text--type="ul" helper-text-item--type="li"}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-success"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="check"}}
+              {{#> helper-text-item-text}}Must be at least 14 characters{{/helper-text-item-text}}
+            {{/helper-text-item}}
+            {{#> helper-text-item helper-text-item--modifier="pf-m-dynamic pf-m-success"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="check"}}
+              {{#> helper-text-item-text}}Cannot contain any variation of the work "redhat"{{/helper-text-item-text}}
```
mcoker, line 125:
```suggestion
              {{#> helper-text-item-text}}Cannot contain any variation of the word "redhat"{{/helper-text-item-text}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
```
mcoker, line 172:
```suggestion
              {{#> helper-text-item-text}}Cannot contain any variation of the word "redhat"{{/helper-text-item-text}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+        {{#> form-group-label-main}}
+              {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+        {{/form-group-label-info}}
+      {{/form-group-label}}
+      {{#> form-group-control}}
+        {{#> input-group}}
+            {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="" placeholder="Password"')}}
```
mcoker, line 23:
```suggestion
            {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'required type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="" placeholder="Password"')}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+        {{#> form-group-label-main}}
+              {{#> form-label form-label--attribute=(concat 'for="' form--id '-password"') required="true"}}Password{{/form-label}}
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+        {{/form-group-label-info}}
+      {{/form-group-label}}
+      {{#> form-group-control}}
+        {{#> input-group}}
+          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$RedHat78" placeholder="Password"')}}
```
mcoker, line 64:
```suggestion
          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'required type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$RedHat78" placeholder="Password"')}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+            {{#> helper-text-item helper-text-item--modifier="pf-m-error"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="exclamation-circle"}}
+              {{#> helper-text-item-text}}Weak{{/helper-text-item-text}}
+            {{/helper-text-item}}
+          {{/helper-text}}
+        {{/form-group-label-info}}
+      {{/form-group-label}}
+      {{#> form-group-control}}
+        {{#> input-group}}
+          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$Can3Read" placeholder="Password"')}}
```
mcoker, line 111:
```suggestion
          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'required type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$Can3Read" placeholder="Password"')}}
```

src/patternfly/demos/PasswordStrength/examples/PasswordStrength.md
```diff
@@ -0,0 +1,187 @@
...
+            {{#> helper-text-item helper-text-item--modifier="pf-m-success"}}
+              {{> helper-text-item-icon helper-text-item-icon--type="check-circle"}}
+              {{#> helper-text-item-text}}Strong{{/helper-text-item-text}}
+            {{/helper-text-item}}
+          {{/helper-text}}
+        {{/form-group-label-info}}
+      {{/form-group-label}}
+      {{#> form-group-control}}
+        {{#> input-group}}
+          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$Can8Read3Pass@Word" placeholder="Password"')}}
```
mcoker, line 158:
```suggestion
          {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'required type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="Marie$Can8Read3Pass@Word" placeholder="Password"')}}
```

mcoker (changes requested):
LGTM! Just a couple of small things.

mcoker (approved):
LGTM! Thanks!!

PR #4159, opened 2021-06-17
fix(table): added background color to draggable ghost row

### LINKED ISSUE #4158: Draggable table row missing background color
The ghost row needs a background color so it covers rows behind it.

mcoker:
fixes #4158


srambach (approved):
Looks fine.

PR #4164, opened 2021-06-17
feat(sass): use dart sass

### LINKED ISSUE #4096: sass: / is deprecated for division
https://github.com/sass/dart-sass/releases/tag/1.33.0

https://sass-lang.com/documentation/breaking-changes/slash-div

If `sass` version 1.33.0 or newer is installed, the console is absolutely flooded by "/ is deprecated for division" errors when patternfly builds.

redallen:
Closes #4096

- Upgrade workspace to dart-sass.
  - New compile time is up to 19.29s from 14.82s but I think that's acceptable for better compatibility (pure JS, no binaries need installing) and to support the next generation of sass.
  - Use clever `div.import.scss` workaround to avoid deprecation warnings while supporting old `node-sass`

Our current CI already makes sure that `node-sass` works: https://github.com/patternfly/patternfly/blob/master/.circleci/config.yml#L114


mcoker (approved):
🥳👍

PR #4179, opened 2021-06-21
fix(contextext-selector-dropdown): changed display type

### LINKED ISSUE #4178: fix(context-selector,dropdown): full-height variant should be display: inline-flex

mattnolting:
closes #4178


mcoker (approved):
👍👍

PR #4180, opened 2021-06-21
feat(search-input): added variant with submit button

### LINKED ISSUE #4036: Add search input variant with submit button
Want to add a new variant of the search input component that has an explicit submit button to the right of the search field. See designs here: https://github.com/patternfly/patternfly-design/issues/1030

bitwiseor:
fixes #4036 

Added a variant of the SearchInput component with a submit button, also added an example on the component page for SearchInput.

### DIFF
--- a/src/patternfly/components/SearchInput/examples/SearchInput.md
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.md
@@ -28,6 +28,11 @@ import './SearchInput.css'
 {{> search-input search-input--placeholder="Find by name" search-input--value="John Doe" search-input--count="1 / 3" search-input--IsNavigable="true" search-input--IsFirstMatch="true"}}
 ```
 
+### With submit button
+```hbs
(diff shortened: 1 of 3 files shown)


mcoker (changes requested):
Perfect!! Great job. One change that's a pre-existing issue is I noticed the search button is `type="button"`. Looks like we're using `type="submit"` in the react component, and that seems more appropriate. Would you mind updating the button in [search-input-search-button.hbs](https://github.com/patternfly/patternfly/blob/da311aa862ab8ea76ebb2018ba813b48515e685b/src/patternfly/components/SearchInput/search-input-search-button.hbs) to be `type="submit"`? Looks like we offer that via `button--IsSubmit` in [button.hbs](https://github.com/patternfly/patternfly/blob/e6045c0b7ae3280b25bd8596cf4da29d2e4b67f4/src/patternfly/components/Button/button.hbs#L2-L3)

bitwiseor:
Updated so the search input search button is of type "submit"!

mcoker (approved):
🥳👍

Issue #4056, opened 2021-05-07
Add expandable inline alert


mcarrano:
Add a new Alert variation for an expandable inline alert.

Designs: https://marvelapp.com/prototype/9ihajjd

gunishmatta:
I want to work on this issue.

mcoker:
Hi @gunishmatta, thanks for offering to contribute! I've assigned it to you and am here if you have any questions. You can find development instructions in the README - https://github.com/patternfly/patternfly#development

Let me know if you'd like an introduction to how the repo is structured, the dev process, etc. You can reach out to me on the [PatternFly slack](https://slack.patternfly.org/) at @mcoker.

mcoker:
@gunishmatta just wanted to check in and see if you still have time to work on this issue, and if so, do you have an ETA? We are hoping to get this feature into a release soon if possible. Thanks!

gunishmatta:
I will try to finish by this weekend

mcoker:
@gunishmatta sounds great, thank you!

PR #4183, opened 2021-06-22
feat(alert): added expandable inline alert

gunishmatta:
Added an inline expandable alert. 
If any issues are there in the code please provide me feedback and I will try to fix them.

fixes https://github.com/patternfly/patternfly/issues/4056


src/patternfly/components/Alert/alert.scss
```diff
@@ -119,22 +125,54 @@
...
 
 .pf-c-alert__icon {
-  grid-area: icon;
   display: flex;
+  align-items: center;
   margin-top: var(--pf-c-alert__icon--MarginTop);
   margin-right: var(--pf-c-alert__icon--MarginRight);
   font-size: var(--pf-c-alert__icon--FontSize);
   color: var(--pf-c-alert__icon--Color);
+  grid-area: icon;
```
mcoker, line 166: nit - we use stylelint to lint our styles, including the order of CSS properties, though looks like the `grid-` properties' orders are not linted. I think it makes more sense to keep the `grid-` properties at the top near the `display` property, similar to how flex properties need to be grouped near the top. WDYT?

src/patternfly/components/Alert/examples/Alert.md
```diff
@@ -54,7 +56,56 @@ cssPrefix: pf-c-alert
...
+    {{/button}}
+  {{/alert-action-group}}
+
+  {{#> alert-title}}
+    {{#> screen-reader}}Success alert:{{/screen-reader}}
+ Success alert title
+  {{/alert-title}}
+  {{#> alert-description}}
+        This is some description for this alert that is super long. This is some description for this alert that is super long. This is some description for this alert that is long.
+        Some additional information:
```
mcoker, line 101: nit - could you turn this into a list? An easy wan to do that would be to wrap everything in a `{{#> content}}` block, then use `<p>`'s for the blocks of text, and an `<ol>` for this list - the `.pf-c-content` component will space/format everything nicely.

src/patternfly/components/Alert/alert-toggle.hbs
```diff
@@ -0,0 +1,4 @@
+  {{#> button button--modifier="pf-m-plain"}}
```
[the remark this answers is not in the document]
mcoker: Could you also add `aria-expanded={{#if alert--IsExpanded}}true{{else}}false{{/if}}` here, and set `alert--IsExpanded="true"` on the example that is expanded? You can see an example of this in the [expandable section toggle](https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs), though it's written a little differently.

src/patternfly/components/Alert/alert.scss
mcoker: If you agree, maybe we should move the `grid-template-areas/columns` applied to `.pf-c-alert` to the top, too.

mcoker (changes requested):
Thanks for the PR @gunishmatta!! Overall it looks really good and is very close, there are just some fine tuning adjustments that we'll need to address having to do with formatting and consistency with other components in the library.

Let me know if you have any feedback, and reach out if you have any questions or would like to see an example of a change I've requested or anything like that.

mcoker:
@gunishmatta 2 things I forgot to mention. Can you revert `package-lock.json` back to what's on master in this PR? Typically we do not merge changes to that file unless they correspond with a change we've made to `package.json`.

And it looks like there is something that needs to be addressed in the `test_a11y` job, you can see details in the list of CI tests. Here's the error I see

```
17/801    /components/alert/html/variations
violations: aria-valid-attr-value, duplicate-id-aria
```

This is probably due to the lack of `alert--id` defined on the examples, which should be an easy fix!

mcoker:
Hi @gunishmatta just wanted to check in and see if you're still working on this PR and if there is anything I can do to help. Also feel free to ignore the comments that start with "nit" - those aren't necessary to fix to get this feature added to the alert component.

mcoker:
@gunishmatta thanks again for the contribution. We're hoping to get this change in with an upcoming release, so I'm going to go ahead and close this PR for the time being and open a new one so we can move it along. I'm happy to ping you in the new PR if you would like, and I would love to work with you to find another issue for you if you're interested in more contributions!

mcoker:
@gunishmatta I put up a PR for this change here if you would like to review! https://github.com/patternfly/patternfly/pull/4268

PR #4200, opened 2021-07-01
docs(button): updated accessibility table

### LINKED ISSUE #4193: Accessibility table for Button should have aria-label applied to .pf-c-button.pf-m-plain
In [the accessibility table for the Button component](https://www.patternfly.org/v4/components/button/html#accessibility), the aria-label attribute says it should be applied to `.pf-m-plain`. This needs to be updated, as the aria-label for a Button should be applied to `.pf-c-button.pf-m-plain`.

Screen Shot 2021-06-28 at 9 19 47 AM

bitwiseor:
fixes #4193


mcoker (approved):
Nice!

PR #4209, opened 2021-07-08
feat: drop '?#iefix' for font face definitions

jonkoops:
Removes the `?#iefix` suffix for loading `.eot` fonts, this is a workaround for Internet Explorer 8 and older to load [ensure fonts are loaded properly](https://stackoverflow.com/questions/14207425/css-font-face-iefix). This is no longer needed as this browser is not supported by Patternfly.

The motivation for removing this is that it causes [issues](https://github.com/evanw/esbuild/issues/826) with more modern bundlers such as Snowpack (which uses `esbuild` under the hood.

### DIFF
--- a/src/patternfly/assets/fonts/overpass-mono-webfont/overpass-mono.css
+++ b/src/patternfly/assets/fonts/overpass-mono-webfont/overpass-mono.css
@@ -3,7 +3,7 @@
 @font-face {
   font-family: 'overpass-mono';
   src: url('overpass-mono-light.eot');
-  src: url('overpass-mono-light.eot?#iefix') format('embedded-opentype'),
+  src: url('overpass-mono-light.eot') format('embedded-opentype'),
        url('overpass-mono-light.woff2') format('woff2'),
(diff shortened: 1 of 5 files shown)


jonkoops:
Sidenote, it might be a good idea to drop all font formats besides [WOFF](https://caniuse.com/woff) and [WOFF2](https://caniuse.com/woff2), as these are supported by all browsers in the compatibility list. If we don't care about Internet Explorer, then WOFF2 is supported by all browsers.

jenny-s51 (approved):
LGTM @jonkoops

redallen (approved):
I love `esbuild`! It looks like there's already an upstream fix for this, but I don't see the harm since we don't support IE anymore anyways.

jonkoops:
> I love `esbuild`! It looks like there's already an upstream fix for this, but I don't see the harm since we don't support IE anymore anyways.

Yeah, it looks like it hasn't tricked down yet to Snowpack, but that could be us.

mcoker (approved):
LGTM!

PR #4210, opened 2021-07-09
feat(fonts): dropped support for legacy fonts

jonkoops:
This change removes font formats that exist to provide support for older browsers and leaves only the WOFF (for Internet Explorer 11) and WOFF2 (everything else) formats. These formats are enough to provide [full support](https://css-tricks.com/snippets/css/using-font-face/#practical-level-of-browser-support) to the browsers in the support target as outlined in the readme.

I've been a bit conservative and kept WOFF (non 2) around, but WOFF2 is supported by all [the major browsers](https://caniuse.com/woff2). Let me know if we want to also drop this and I will make the change.

The motivation for this PR is to allow the us (Keycloak team) to compile our application using Snowpack, which at this point cannot resolve font paths that include a `#` symbol (which was used for SVG fonts). Incidentally I also resolved a bug where the SVG font source for FontAwesome was incorrectly referred to using `#fontawesome` instead of `#fontawesome-free`.

This PR also enables WOFF2 support for the Red Hat fonts, which for some reason only loaded the WOFF versions, even though the WOFF2 files are present in the directory.


jonkoops:
@jenny-s51 @redallen @mcoker Could I get your opinions on this change?

redallen (approved):
I'm very happy to see this change. I'm curious why WOFF seems to be what browsers load (see https://patternfly.org) instead of WOFF2, but since you're keeping WOFF around it doesn't matter much.

jenny-s51 (approved):

jonkoops:
> I'm very happy to see this change. I'm curious why WOFF seems to be what browsers load (see https://patternfly.org) instead of WOFF2, but since you're keeping WOFF around it doesn't matter much.

The reason for this is that the WOFF2 variants of the fonts were not included for the RedHat fonts (see my comment on the PR). I have however [fixed this](https://github.com/patternfly/patternfly/pull/4210/files#diff-5483908e451094cacdd1510fc18b8f03e6999203407e08667a30741fc684db1eR12) in this change, and you can see the WOFF2 files loading in the network tab when you check the [preview on Surge](https://patternfly-pr-4210.surge.sh):

image

I think that we can actually ditch WOFF as well and just use WOFF2, but I would like your input on that.

redallen:
I think it wise to keep WOFF around for now because:
1. We unofficially support IE11
2. We might destroy caches that have cached the existing WOFF files

The only compelling reason to remove it is lower size on disk of our assets. However, I haven't heard anyone complain about disk space.

mcoker (approved):
LGTM, thanks @jonkoops!

PR #4222, opened 2021-07-15
fix(a11y): fix for accessibility issue in nav section title

agagancarczyk:
hi @mcoker, while working on the accessibility in a new keycloak-admin-ui, I have noticed accessibility issue "Hadings should not be empty" in the Nav component in the PF lib as follows. I used axe devtools chrome extension by deque to find this issue.
keycloak-a11-issue

 I think I have found a fix for the issue by simply adding aria-label to nav section title as follows: 
adding-aria-label-1

This fix will help us to remove the accessibility issue with the following result: 
after-adding-aria-label-result-2

Please let me know if this fix will fix the issue or if you think there is an alternative way to fix it. Many thanks!

### DIFF
--- a/src/patternfly/components/Nav/examples/Navigation.md
+++ b/src/patternfly/components/Nav/examples/Navigation.md
@@ -40,7 +40,7 @@ import './Navigation.css'
 ```hbs
 {{#> nav nav--attribute='aria-label="Global"'}}
   {{#> nav-section nav-section--attribute='aria-labelledby="grouped-title1"'}}
-    {{#> nav-section-title nav-section-title--attribute='id="grouped-title1"'}}
+    {{#> nav-section-title nav-section-title--attribute='id="grouped-title1" aria-label="grouped-title1"'}}
       Section title 1
     {{/nav-section-title}}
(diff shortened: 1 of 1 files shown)


mcoker:
Hey @agagancarczyk! It looks like the issue is that you have an empty `<h2>` as the group title for that nav section. Adding an `aria-label` to the examples in this PR doesn't do anything since the titles in these examples have text.

Currently it looks like our grouped navigation relies upon the title being present. We could update the component to support no title.

wdyt @mcarrano? If using grouped navigation, should all nav groups have a clear title? Or could we support nav sections (spaced apart) with or without titles?

agagancarczyk:
Thank you @mcoker , supporting no title would do the trick here if agreed by @mcarrano .

mcarrano:
I do think that we envisioned that groups would always have a title. Any reason why we can't provide a group title @agagancarczyk ? Otherwise what is the goal of grouping items?

mcoker:
Looks like this PR is stale - feel free to re-open if you'd like to revisit!

PR #4223, opened 2021-07-15
fix(a11y): fix for accessibility issue in nav link

agagancarczyk:
hi @mcoker , while working on the accessibility in a new keycloak-admin-ui, I have noticed the accessibility issue "Links with the same name have a similar purpose" in the Nav component in the PF lib as follows. This is because we have the same link "clients" in the nav and breadcrumbs. I used axe devtools chrome extension by deque to find this issue.

keycloak-issue-nav-item

I think I have found a potential fix for the issue by simply adding aria-label to nav link as follows:
patternfly-nav-link-fix

This fix will help us to remove the accessibility issue with the following result:
keycloak-after-fix-2

Please let me know if this fix will fix the issue or if you think there is an alternative way to fix it. Many thanks!

### DIFF
--- a/src/patternfly/components/Nav/examples/Navigation.md
+++ b/src/patternfly/components/Nav/examples/Navigation.md
@@ -13,22 +13,22 @@ import './Navigation.css'
 {{#> nav nav--attribute='aria-label="Global"'}}
   {{#> nav-list}}
     {{#> nav-item}}
-      {{#> nav-link nav-link--href="#"}}
+      {{#> nav-link nav-link--href="#" nav-link--attribute='aria-label="default-nav-link"'}}
         Link 1
       {{/nav-link}}
     {{/nav-item}}
     {{#> nav-item}}
-      {{#> nav-link nav-link--href="#" nav-link--current="true"}}
+      {{#> nav-link nav-link--href="#" nav-link--current="true" nav-link--attribute='aria-label="default-nav-link"'}}
         Current link
       {{/nav-link}}
     {{/nav-item}}
     {{#> nav-item}}
-      {{#> nav-link nav-link--href="#"}}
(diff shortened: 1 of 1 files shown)


mcoker:
@agagancarczyk sorry for the delay on this! @jessiehuff wdyt about this? I'm not seeing this axe error on our examples, though it is erroring about the patternfly logo links in the header and footer, and and the "open example in new window" links under each example. I just spot checked the navigation component examples/demos, it may be appearing on other pages though.

mcoker:
ping @jessiehuff do you think this is something we should look at? I wasn't seeing the warning/error on the pages I tested.

jessiehuff:
@agagancarczyk Based on what I can see in the screenshots, I think the issue is that two of the links within your page have the same text but likely have a different href. Deque explains this rule in [this article](https://dequeuniversity.com/rules/axe/4.1/identical-links-same-purpose). 

For example, if someone has "Read more" links throughout their page, a screen reader user has no way of distinguishing between these two links when they're out of context in a screen reader menu. You can add an `aria-label` that gives more context like `aria-label="Read more about Seminole tax hike"` and label the other link `aria-label="Read more about Seminole's new baby mayor"` (using Deque's example) but the `aria-label` should essentially add the context that's missing. In your case, I can't tell where the other "Clients" link it might be pointing to is, but I'd recommend something maybe with more context than "nav-link". With that label, a screen reader will completely override the "Clients" text with that label and have no idea what that link is when they just hear "nav-link". 

Let me know if you'd like me to dig in further or take a look at your UI. I'm happy to take a deeper look and help with any of these details. :)

mcoker:
Closing for now. Feel free to open it again if we need to revisit this issue.

PR #4226, opened 2021-07-15
feat(popover): add position options

### LINKED ISSUE #3930: Enhancement: Popover should support "diagonal" positioning

Popover only supports cardinal positioning, i.e. NSEW (top, bottom, ...); and the "flip" support also only supports flipping along these cardinal directions. If the Popover is associated with an element in a corner of a viewport, none of these orientations works. For example, if the Popover is associated with an element in the bottom left of the viewport, then Top positioning will result in the popover being partially, possibly even mostly, off-screen to the left; and Right positioning will result in the popover being partially off-screen to the bottom; etc.

This could I think be easily addressed by allowing for NW/NE/SE/SW positioning; and flipping in this fashion as well.

Enhancement. No work-around is possible at the moment, except avoiding the problematic viewport locations altogether :(

RedHat ACM, Kubernetes Kui

srambach:
This adds modifiers `.pf-m-start` and `.pf-m-end` to the popover in order to adjust the arrow when the position of the popover is at the start (top for left and right; left for top and bottom) and the end (bottom for left and right; right for top and bottom). 

Fixes #3930

### DIFF
--- a/src/patternfly/components/Popover/popover.scss
+++ b/src/patternfly/components/Popover/popover.scss
@@ -29,6 +29,8 @@
   --pf-c-popover__arrow--m-left--TranslateX: 50%;
   --pf-c-popover__arrow--m-left--TranslateY: -50%;
   --pf-c-popover__arrow--m-left--Rotate: 45deg;
+  --pf-c-popover__arrow--m-start--Left: var(--pf-c-popover__arrow--Width);
+  --pf-c-popover__arrow--m-start--Top: var(--pf-c-popover__arrow--Height);
 
   // Close
   --pf-c-popover--c-button--MarginLeft: var(--pf-global--spacer--sm);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Popover/popover.scss
```diff
@@ -94,6 +96,36 @@
       transform: translateX(var(--pf-c-popover__arrow--m-right--TranslateX)) translateY(var(--pf-c-popover__arrow--m-right--TranslateY)) rotate(var(--pf-c-popover__arrow--m-right--Rotate));
     }
   }
+
+  &.pf-m-left.pf-m-start,
+  &.pf-m-right.pf-m-start {
+    .pf-c-popover__arrow {
```
mcoker, line 102: If we setup base vars for the arrow's position, we can remove this `.pf-c-popover__arrow` selector and set the vars with the `.pf-m-[direction]` selectors.

Also re: having `.pf-m-left.pf-m-start` vs `.pf-m-top-left`, one argument for the latter is the specificity of the rules will be lower. But seems like that would be more code since you'd need 8 selectors instead of 4.

wdyt @mattnolting?

mcoker (commented):
Looks great! Just one bit of feedback.

src/patternfly/components/Popover/popover.scss
srambach: FWIW, I looked at how specific it already was, and what the element is, and my decision was that one more modifier wasn't that significant. This seemed cleaner to implement as well as to understand. But I see how it could go either way.

mattnolting: I think `.pf-m-top/bottom-left/right` is consistent with the way CSS defines specific props, like `border-top-left-radius`. That seems a bit more straightforward to me.

srambach: So @mcoker and @mattnolting, you want `.pf-m-top/bottom/left/right-top/bottom/left/right` and not `.pf-m-top/bottom/left/right-start/end` like popper has, is that correct?

mcoker (approved):
LGTM!

mattnolting (approved):
Nice! LPTM!980987

PR #4237, opened 2021-07-26
fix(wizard): switched ol to span in wizard toggle

### LINKED ISSUE #2202: Address nested <ol> in the wizard toggle <button>
Only [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) is permitted in the [<button>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element, so we need to address the `ol.pf-c-wizard__toggle-list` in `button.pf-c-wizard__toggle`.

Should we just remove the list and use `<span>`s to style the toggle text instead?

cc @jgiardino @jessiehuff

bitwiseor:
fixes #2202 

Needs double checking; I'm not sure where in my dev server to find the relevant demos to verify that this is correct 😨

### DIFF
--- a/src/patternfly/components/Wizard/wizard-toggle-list.hbs
+++ b/src/patternfly/components/Wizard/wizard-toggle-list.hbs
@@ -1,6 +1,6 @@
-<ol class="pf-c-wizard__toggle-list{{#if wizard-toggle-list--modifier}} {{wizard-toggle-list--modifier}}{{/if}}"
+<span class="pf-c-wizard__toggle-list{{#if wizard-toggle-list--modifier}} {{wizard-toggle-list--modifier}}{{/if}}"
   {{#if wizard-toggle-list--attribute}}
     {{{wizard-toggle-list--attribute}}}
   {{/if}}>
   {{> @partial-block}}
-</ol>
+</span>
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Wizard/wizard-toggle-list.hbs
```diff
@@ -1,6 +1,6 @@
-<ol class="pf-c-wizard__toggle-list{{#if wizard-toggle-list--modifier}} {{wizard-toggle-list--modifier}}{{/if}}"
+<span class="pf-c-wizard__toggle-list{{#if wizard-toggle-list--modifier}} {{wizard-toggle-list--modifier}}{{/if}}"
```
mcoker, line 1: This looks good, but looks like the children of this element (`.pf-c-wizard__toggle-list-item`) are still `<li>`s, which is invalid in a `<span>`.

mcoker (changes requested):
> I'm not sure where in my dev server to find the relevant demos to verify that this is correct

@bitwiseor you can look at any of the examples or demos on https://patternfly-pr-4237.surge.sh/components/wizard.

The change is in the top nav toggle. You'll want to verify the layout matches what's currently in main on patternfly.org both in the collapsed and expanded states. 

The layout looks good to me! Just one comment about the element of the lists direct children.

mcoker (changes requested):
Looks good! Can you also update the docs to list the new elements?

Screen Shot 2021-07-27 at 3 32 34 PM

mcoker (approved):
LGTM! @jessiehuff are there any a11y concerns with this change from an `ol` to `span`s?

jessiehuff (approved):
So I tested in VO, JAWS, and NVDA, and all of them sound the same as what's currently in org so it seems ok to me. I did notice that in NVDA the wizard step number and the wizard step button are being read as two separate buttons for some reason, but I don't think it's related to this PR since when I tested org I got the same result.

PR #4240, opened 2021-07-26
fix(form-group): updated form group role

### LINKED ISSUE #4124: Form groups for checkboxes/radios have wrong markup
When grouping checkboxes and radios, the markup for `pf-c-form__group` incorrectly uses a `<label>` element to define the group's "heading." These types of groupings should use the `<fieldset>` element and their "headings" should be a `<legend>` rather than a `<label>`.

- This issue directly affects accessibility as well.
- This issue can be seen at the [PF demo site](https://www.patternfly.org/v4/components/form#limit-width)
- [More info on using fieldsets](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes) for grouping checkboxes and radios

image

CC: @jessiehuff

mattnolting:
fixes #4124 

Fix based on the guidance provided here: [associating-related-controls-with-wai-aria](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria)

Using the `<fieldset>` element with grid worked for formatting, although I'm not sure if it loses semantic value in doing so. It looks like, based on [this description](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#styling_with_css), that it's acceptable to use. What didn't work well was the `<legend>` element, which is only display block. Using this breaks the `required` and/or `helper` concept layouts.

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -46,7 +46,7 @@ cssPrefix: pf-c-form
       {{#> form-control controlType="textarea" form-control--attribute=(concat 'type="text" id="' form--id form-group--id '" name="' form--id form-group--id '" aria-label="Textarea example"')}}{{/form-control}}
     {{/form-group-control}}
   {{/form-group}}
-  {{#> form-group form-group--id="-checkbox"}}
+  {{#> form-group form-group--IsCheckGroup="true" form-group--id="-checkbox"}}
     {{#> form-group-label form-group-label--modifier="pf-m-no-padding-top"}}
       {{#> form-label}}
         Label (no top padding)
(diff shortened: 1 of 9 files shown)


mattnolting:
@markcaron

mcoker (commented):
> What didn't work well was the <legend> element, which is only display block. Using this breaks the required and/or helper concept layouts.

This works for me. One question - from https://www.w3.org/WAI/tutorials/forms/grouping/#radio-buttons, it says

> Radio button groups should always be grouped using <fieldset>

And from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints#labeling_with_aria, it looks like there is a specific `role="radiogroup"` for radio groups. Do you know if that's needed to treat radio groups differently? All of the examples I found using that role are in combination with custom `role="radio"` elements that aren't `<input>` elements.

Also an aside, our use of radio buttons on https://patternfly-pr-4240.surge.sh/components/form/html-demos/ seems inappropriate. IMO those should be checkboxes, and we should show and example of a radio group where only one option can be selected. WDYT @mcarrano?

mcarrano:
> Also an aside, our use of radio buttons on https://patternfly-pr-4240.surge.sh/components/form/html-demos/ seems inappropriate. IMO those should be checkboxes, and we should show and example of a radio group where only one option can be selected. WDYT @mcarrano?

Good catch @mcoker I totally agree.

mcoker:
@mattnolting can you update the demos with an example of a checkbox group and a radio button group in this PR? We can do that in a separate issue, too, if you'd prefer.

mattnolting:
@mcoker sure, updated

mcarrano:
New demo looks great. Thanks for clarifying these examples @mattnolting @mcoker

mcoker:
Leaving a comment about what we discussed offline - moving the radio group into the basic demo under the checkbox group, I don't think we need a separate radio group example.

mattnolting:
@mcarrano I moved the radio group to the basic example

mcarrano (approved):
This looks good to me. Thanks @mattnolting .

mcoker (approved):
LGTM!

jessiehuff (approved):
LGTM! 😄 (Also, I tested it in VO, JAWS, and NVDA and all of them identify it correctly as a group.)

markcaron:
For what it's worth, the ARIA role "radiogroup" seems like a good workaround for the traditional fieldset here. As long as we're using a heading for the group's heading rather than a label (like in the original issue).

https://dequeuniversity.com/library/aria/radio-and-radio-group

We should test screen readers, though, because ARIA roles don't always perform consistently (and often depend on specific application).

https://a11ysupport.io/tests/tech__aria__aria-required-radiogroup

markcaron:
Whoops! Overlooked @jessiehuff's testing. Glad it passed!

Nice work!

PR #4246, opened 2021-07-28
fix(helper-text): updated static and dynamic to use same icons

### LINKED ISSUE #4205: Update icons in Helper text component
Wnat to update the icons used in Helper text so that static and dynamic helper text use the same icons. Revised mockups are here: https://marvelapp.com/prototype/6jehce7/screen/80602884

bitwiseor:
fixes #4205 

updated static demo icons to match dynamic demo icons (specifically default and indeterminate)

### DIFF
--- a/src/patternfly/components/HelperText/examples/HelperText.md
+++ b/src/patternfly/components/HelperText/examples/HelperText.md
@@ -43,14 +43,14 @@ cssPrefix: pf-c-helper-text
 ```hbs
 {{#> helper-text}}
   {{#> helper-text-item}}
-    {{> helper-text-item-icon helper-text-item-icon--type="info"}}
+    {{> helper-text-item-icon helper-text-item-icon--type="minus"}}
     {{#> helper-text-item-text}}This is default helper text{{/helper-text-item-text}}
   {{/helper-text-item}}
 {{/helper-text}}
 
 {{#> helper-text}}
   {{#> helper-text-item helper-text-item--modifier="pf-m-indeterminate"}}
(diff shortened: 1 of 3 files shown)


mcoker:
@mcarrano can you confirm the actual change needed for this? 

Looking at the react component, the icons are the same - is that what we should be matching?

Screen Shot 2021-07-28 at 3 50 19 PM
Screen Shot 2021-07-28 at 3 50 13 PM

mcarrano:
@mcoker No. The React component has inconsistent icons. The original problem was that we can different sets of icons in each example. The goal was to use the same icons for error, success, etc. Looks like the design is saying to use the filled circle versions and not the open versions for everything. Does that help?

bitwiseor:
Should they be matching the icons in [the static with custom icons demo](https://www.patternfly.org/v4/components/helper-text#static-with-custom-icons), but with filled circle versions for default and indeterminate as well?

mcoker:
@bitwiseor sorry for the confusion. I met with @mcarrano and we determined these should be the icons used for both static and dynamic

Screen Shot 2021-07-28 at 4 24 55 PM

We'll also want to make sure everywhere the component is used is updated. That should be:

* Helper text examples page
* Helper text demos page
* Password Strength demo

mcoker (approved):
LGTM!! 🥳

mcoker:
@doruskova can you confirm the icons here are what you were expecting to see? https://patternfly-pr-4246.surge.sh/components/helper-text

doruskova:
@mcoker Yes! They're correct. Thank you!

PR #4256, opened 2021-07-30
feat(notification-drawer) Add task list and tabs to the notification drawer

### LINKED ISSUE #4186: Enhance Notification drawer component to display tasks
Want to add tasks to the notification drawer per the designs here: https://marvelapp.com/prototype/g8idd54/screen/79645478 . After making this enhancement the component should be able to display a list of alerts, tasks, or both. Also consider whether this component should be renamed since it should be possible to use with or without the drawer.

srambach:
This adds a task list to the notification drawer, and tabs to separate the alert list from the task list.
Known issues: 
- the progress bar cannot (afaik) be single line and not have a space to the right side for the measure
- I question the alignment of the kebab with the toggle caret in the version with multiple actions
- The spinner shows an a11y error for the wrong role, but this also happens in the spinner component

Fixes #4186


src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -97,6 +134,49 @@
...
+  // Task group toggle action
+  --pf-c-notification-drawer__task-group-toggle-action--MarginRight: var(--pf-global--spacer--md);
+
+  // Task group toggle count
+  --pf-c-notification-drawer__task-group-toggle-count--MarginRight: var(--pf-global--spacer--md);
+
+  // Task group toggle icon
+  --pf-c-notification-drawer__task-group-toggle-icon--MarginRight: var(--pf-global--spacer--md);
+  --pf-c-notification-drawer__task-group-toggle-icon--Color: var(--pf-global--Color--200);
+  --pf-c-notification-drawer__task-group-toggle-icon--Transition: .2s ease-in 0s;
```
mcoker, line 172: Can we update this (and the notification item icon transition) to use `var(--pf-global--Transition)`? We've since started using that for these icon toggle transitions.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -128,11 +208,18 @@
   margin-left: auto;
 }
 
+.pf-c-notification-drawer__tabs {
+  background-color: var(--pf-c-notification-drawer__header--BackgroundColor);
```
mcoker, line 212: Since this isn't part of the header, can this be it's own var? Also if we were to ever separate the background color vars for these two in the future, the way it is here would technically be a breaking change.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -208,6 +300,76 @@
...
+
+.pf-c-notification-drawer__task-list-item-hover-action {
+  display: var(--pf-c-notification-drawer__task-list-item-hover-action--Display);
+  margin-right: var(--pf-c-notification-drawer__task-list-item-hover-action--MarginRight);
+}
+
+.pf-c-notification-drawer__task-list-item-status {
+  margin-left: auto;
+  color: var(--pf-c-notification-drawer__task-list-item-status--Color);
+  text-align: end;
```
mcoker, line 362: Is this necessary if it's always pushed to the right? Also it's totally fine if you want to use a logical property, but we don't really use them anywhere as far as I'm aware, except for flex/grid position.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -289,17 +476,79 @@
...
+// Task group toggle
+.pf-c-notification-drawer__task-group-toggle {
+  display: grid;
+  align-items: center;
+  width: 100%;
+  grid-template-columns: 1fr auto auto;
+  padding: var(--pf-c-notification-drawer__task-group-toggle--PaddingTop) var(--pf-c-notification-drawer__task-group-toggle--PaddingRight) var(--pf-c-notification-drawer__task-group-toggle--PaddingBottom) var(--pf-c-notification-drawer__task-group-toggle--PaddingLeft);
+  background-color: var(--pf-c-notification-drawer__task-group-toggle--BackgroundColor);
+  border: solid var(--pf-c-notification-drawer__task-group-toggle--BorderColor);
+  border-width: 0 0 var(--pf-c-notification-drawer__task-group-toggle--BorderBottomWidth) 0;
```
mcoker, line 488: Typically this and the line above are just written as `border-bottom: var(--pf-c-notification-drawer__task-group-toggle--BorderBottomWidth) solid var(--pf-c-notification-drawer__task-group-toggle--BorderBottomColor);`

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -289,17 +476,79 @@
...
+
+.pf-c-notification-drawer__task-group-header {
+  grid-column: 1 / 2;
+  grid-row: 1 / 2;
 }
 
-.pf-c-notification-drawer__group-toggle-icon {
-  margin-right: var(--pf-c-notification-drawer__group-toggle-icon--MarginRight);
-  color: var(--pf-c-notification-drawer__group-toggle-icon--Color);
-  transition: var(--pf-c-notification-drawer__group-toggle-icon--Transition);
```
mcoker, line 300: Removing these 3 vars will be a breaking change. If you wanted to share the vars, which seems safe here, you would want to use these existing ones.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -289,17 +476,79 @@
...
+}
+
+.pf-c-notification-drawer__group-toggle-icon,
+.pf-c-notification-drawer__task-group-toggle-icon {
+  margin-right: var(--pf-c-notification-drawer__task-group-toggle-icon--MarginRight);
+  color: var(--pf-c-notification-drawer__task-group-toggle-icon--Color);
+  transition: var(--pf-c-notification-drawer__task-group-toggle-icon--Transition);
 
-  .pf-c-notification-drawer__group.pf-m-expanded & {
-    transform: rotate(var(--pf-c-notification-drawer__group--m-expanded__group-toggle-icon--Rotate));
```
mcoker, line 303: Same for this, breaking to remove it.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -289,17 +476,79 @@
   word-break: break-word;
 }
 
-.pf-c-notification-drawer__group-toggle-count {
-  margin-right: var(--pf-c-notification-drawer__group-toggle-count--MarginRight);
-  margin-left: auto;
+// Task group toggle
+.pf-c-notification-drawer__task-group-toggle {
+  display: grid;
+  align-items: center;
```
mcoker, line 482: This causes the toggle and count to be centered vertically:

Screen Shot 2021-08-04 at 5 34 56 PM

That doesn't match the existing notification group styling:

Screen Shot 2021-08-04 at 5 33 38 PM

src/patternfly/components/NotificationDrawer/notification-drawer-basic-task-list-hover-action-progress.hbs
```diff
@@ -0,0 +1,67 @@
...
+
+  {{#> notification-drawer-task-list-item notification-drawer-task-list-item--IsHoverable="true"}}
+    {{#> notification-drawer-task-list-item-title}}
+      hover action with progress bar status
+    {{/notification-drawer-task-list-item-title}}
+    {{#> notification-drawer-task-list-item-hover-action}}
+      Cancel
+    {{/notification-drawer-task-list-item-hover-action}}
+    {{#> notification-drawer-task-list-item-status}}
+      {{#> progress
```
mcoker, line 11: This is creating extra space to the right because the single line variation has a grid-gap, expecting the value to be on the right. I wonder if we need a variation for a progress without a value, or if this should include a value? wdyt @mcarrano?

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-toggle-header.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="pf-c-notification-drawer__task-group-toggle-header{{#if notification-drawer-task-group-toggle-header--modifier}} {{notification-drawer-task-group-toggle-header--modifier}}{{/if}}">
```
mcoker, line 1: this is missing the `--attribute` conditional

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-toggle.hbs
```diff
@@ -0,0 +1,11 @@
+<button class="pf-c-notification-drawer__task-group-toggle{{#if notification-drawer-task-group-toggle--modifier}} {{notification-drawer-task-group-toggle--modifier}}{{/if}}"
```
mcoker, line 1: This markup is invalid. The children of a button should be `<span>` elements, and you can't have a link in a button.

Screen Shot 2021-08-04 at 5 42 51 PM

src/patternfly/components/NotificationDrawer/notification-drawer-task-list-item-hover-action.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="pf-c-notification-drawer__task-list-item-hover-action{{#if notification-drawer-task-list-item-hover-action--modifier}} {{notification-drawer-task-list-item-hover-action--modifier}}{{/if}}"
+  {{#if notification-drawer-task-list-item-hover-action--attribute}}
+    {{{notification-drawer-task-list-item-hover-action--attribute}}}
+  {{/if}}>
+  <button class="pf-c-button pf-m-inline pf-m-link" type="a">{{> @partial-block}}</button>
+
+
```
mcoker, line 7: nit - some extra lines here.

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-toggle-action.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-notification-drawer__task-group-toggle-action{{#if notification-drawer-task-group-toggle-action--modifier}} {{notification-drawer-task-group-toggle-action--modifier}}{{/if}}"
+  {{#if notification-drawer-task-group-toggle-action--attribute}}
+    {{{notification-drawer-task-group-toggle-action--attribute}}}
+  {{/if}}>
+  <a class="pf-c-button pf-m-inline pf-m-link">{{> @partial-block}}</a>
```
mcoker, line 5: Looks like "action" gets passed to these - curious why it's an `<a>`? Also if we keep this as the butotn component, can we use the `{{#> button}}` hbs to generate this element?

mcoker (changes requested):
Great work on all of this! I left some feedback.

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
```diff
@@ -105,6 +105,332 @@ cssPrefix: pf-c-notification-drawer
 {{/notification-drawer}}
 ```
 
+### With tabs and tasks with hover actions
```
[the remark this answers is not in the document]
srambach: Fixed the typo so it should be now.

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-list.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-notification-drawer__task-group-list{{#if notification-drawer-task-group-list--modifier}} {{notification-drawer-task-group-list--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
srambach: It should be now that the typo is fixed.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -97,6 +134,49 @@
...
+  --pf-c-notification-drawer__task-group-toggle--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-notification-drawer__task-group-toggle--BorderColor: var(--pf-global--BorderColor--100);
+  --pf-c-notification-drawer__task-group-toggle--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-notification-drawer__task-group-toggle--OutlineOffset: #{pf-size-prem(-4px)};
+
+  // Task group toggle header
+  --pf-c-notification-drawer__task-group-toggle-header--AlignItems: baseline;
+
+  // Task group toggle title
+  --pf-c-notification-drawer__task-group-toggle-title--MarginRight: var(--pf-global--spacer--md);
```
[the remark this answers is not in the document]
srambach: > Also just to verify, the spinner and action element should be part of the task list group and not just something like a generic container for users to put whatever they want there?

@mcarrano I believe we said the status would be tied to the task statuses, so I assumed that would be a specific part of the component. Is this correct?
@mcoker @mcarrano If action is not necessarily an action, what should a more generic container be called?

src/patternfly/components/NotificationDrawer/notification-drawer.scss
srambach: @mcoker The border-width is turning off the other borders that would be coming from the button styling. This is taken from how the `.pf-c-notification-drawer__group-toggle` is done. Do you have a better way that also handles the other borders?

src/patternfly/components/NotificationDrawer/notification-drawer.scss
srambach: The title in the task group toggle and the existing group toggle for alerts is actually clamped at 1 line as far as I can tell?

The center alignment  provided the alignment shown in the [mockup](https://marvelapp.com/prototype/g8idd54/screen/79182156) when there is a timestamp below the title, and the count and toggle icon are shown as center aligned vertically. I asked @mcarrano and if the title does actually go to multiple lines, we could top align instead.

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-toggle.hbs
srambach: @mcoker This is mirroring the existing group toggle, which includes `div`s. So I guess they all should be updated?

The link in the design for the task group toggle further complicates matters. This renders in Chrome, FF, and Safari, but I agree it is invalid according to the spec. What is your suggestion here?

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-toggle-action.hbs
srambach: @mcoker This is related to the issue above of the action link nested inside a button. I think we could use `{{#> button-link }}` but it wasn't creating the underline on hover, which I thought was better to have since there's also an action on clicking the entire toggle. Anyway we need to resolve how to handle the link inside the toggle first.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
mcoker: Gotcha, wasn't thinking about the button's default borders.

src/patternfly/components/NotificationDrawer/notification-drawer-task-group-toggle.hbs
mcoker: > This is mirroring the existing group toggle, which includes divs. So I guess they all should be updated?

Yeah I think so, or we can create a follow up issue to do the notification drawer items since it's out of scope for this PR. Up to you.

I think in this case, we might make the group toggle a div, and handle the click event in the react component, and keep any other clicks events on clickable elements in the toggle from bubbling up. WDYT @kmcfaul @nicolethoen?

mcoker:
https://github.com/patternfly/patternfly/issues/4186 was closed

PR #4262, opened 2021-08-03
feat(alert): add plain inline alert

### LINKED ISSUE #4220: Alert: add "plain" variant
Add a plain variant of the Alert component. This is designed as a light-weight inline alert that can be placed any where in the page. Unlike the full inline alert, it would not have other actions associated with it. See https://marvelapp.com/prototype/852jaj9/screen/75866265 for an example of what appearance and placement might look like.

srambach:
Adds a plain variation on the inline alert.

Fixes #4220

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -73,8 +73,12 @@
   --pf-c-alert--m-inline--m-info--BackgroundColor: var(--pf-global--palette--blue-50);
 
   // Inline plain
-  --pf-c-alert--m-inline--m-plain--BorderTopColor: transparent;
+  --pf-c-alert--m-inline--m-plain--BorderTopWidth: 0;
   --pf-c-alert--m-inline--m-plain--BackgroundColor: transparent;
+  --pf-c-alert--m-inline--m-plain--PaddingTop: 0;
+  --pf-c-alert--m-inline--m-plain--PaddingRight: 0;
+  --pf-c-alert--m-inline--m-plain--PaddingBottom: 0;
+  --pf-c-alert--m-inline--m-plain--PaddingLeft: 0;
 
   @include pf-t-light; // This component always needs to be light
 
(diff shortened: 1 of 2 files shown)


mcoker (commented):
Looks great! Just a nit, but if you set `--pf-c-alert--BorderTopWidth: 0` instead of `--pf-c-alert--BorderTopColor: transparent`, the border width will not be part of the overall alert height. Currently there is an extra `--pf-global--BorderWidth--md` on the top compared to the bottom.

Also from the design, it looks like the padding around it is also removed since there is no longer a box around it, and so that the left/right edges align with the left/right edge of whatever comes before and after it. Can you confirm @mcarrano?

Here is what it looks like now with content above/below it:

Screen Shot 2021-08-03 at 1 09 07 PM

And this shows the alert padding creating the space:

Screen Shot 2021-08-03 at 1 09 13 PM

This is what it looks like with the padding removed:

Screen Shot 2021-08-03 at 1 08 51 PM

And the alert box:

Screen Shot 2021-08-03 at 1 08 58 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -119,6 +127,16 @@
     --pf-c-alert--BoxShadow: var(--pf-c-alert--m-inline--BoxShadow);
     --pf-c-alert--BackgroundColor: var(--pf-c-alert--m-inline--BackgroundColor);
   }
+
+  &.pf-m-plain {
+    border-top-width: var(--pf-c-alert--m-inline--m-plain--BorderTopWidth);
```
mcoker, line 132: Will this work?

```suggestion
    --pf-c-alert--BorderTopWidth: var(--pf-c-alert--m-inline--m-plain--BorderTopWidth);
```

mcarrano (commented):
This looks good @srambach . Just a couple of questions....

- I assume that the plain styling could be applied to any alert type, i.e. info, warning, etc.
- Does the text pick up the appropriate color per alert level, e.g. red text for an error alert? 

If the answer is yes for those, then I think we're good.

srambach:
> This looks good @srambach . Just a couple of questions....
> 
> * I assume that the plain styling could be applied to any alert type, i.e. info, warning, etc.
> * Does the text pick up the appropriate color per alert level, e.g. red text for an error alert?
> 
> If the answer is yes for those, then I think we're good.

@mcarrano Yes, it does apply to any and the color comes along - you can see that it's the blue info color rather than the teal default now. Do you want an example for each? Or is this ok?

src/patternfly/components/Alert/alert.scss
srambach: 🤪

mcoker (approved):
LGTM!!

mcarrano:
> @mcarrano Yes, it does apply to any and the color comes along - you can see that it's the blue info color rather than the teal default now. Do you want an example for each? Or is this ok?

Thanks for clarifying @srambach . I don't know if we need examples of all of these or it's fine as is. What do you think @mcoker ? Do you think we should have a new example called "Plain types" that show them all or do you think it is clear enough to just folder this into the Inline types as it is now?

mcoker:
@mcarrano good question. I don't think we need to show the different status types from a development side, since the plain variation doesn't do anything with those types. Though we could show the types, just to communicate the plain variation works with all types. I do think it gets lost where it is now, I didn't see it the first time I looked at the page. I think "types" with the alert examples are the different status types. I might either expect it to be under variations, or as a new example. I would probably list it as a new example, "Inline plain", with using the success variation, similar to the custom icon example.

mcarrano (approved):
I like this much better. Thanks @srambach !

mcoker (approved):

PR #4271, opened 2021-08-06
fix(workspace): show fullscreen example readonly code

redallen:
Accidentally was hiding on https://www.patternfly.org/v4/components/background-image


redallen:
Closing in favor of #4270

Issue #3642, opened 2020-11-03
Review text color and background color utility classes


mcoker:
Currently we don't have much of a strategy around the colors and background colors offered in the new text and background color utility classes. I'd like to take some time and consider what variations we should allow for colors and background colors, and track that in this issue.

Some thoughts from a short conversation with design:

* Text colors should be limited to the status colors we apply to text in components with status (eg, success alert title text)
* Icons can be any text and status color. Does that mean we have icon-color and text-color classes?
* Background colors are limited to just global backgrounds. What about status backgrounds - inline alert backgrounds for example.

PR #4283, opened 2021-08-10
[WIP] Text and background color utilities

srambach:
This adds in some text color and background color utilities. There are a few issues and questions to work out.
- Link, Active, and Primary text colors were already listed in the documentation and implemented, but not in any examples. Do we want to show them?
- Icons use the same colors as text, but there is also a variable for a light and dark icon. These were implemented but I've added an example for these.
- I've put light text/dark backgrounds into examples marked "inverse" and pointed to them from the documentation. 
- I've added some additional documentation that we could use to address the purpose and caveats of using these as customizations.

Note this is WIP. I'm sending it out so that we can discuss direction. @mcarrano @mcoker @mceledonia 
#3642

### DIFF
--- a/src/patternfly/utilities/BackgroundColor/BackgroundColor.scss
+++ b/src/patternfly/utilities/BackgroundColor/BackgroundColor.scss
@@ -19,9 +19,25 @@ $pf-u-background-color-options: (
   ),
   background-color-dark-400: (
     background-color var(--pf-global--BackgroundColor--dark-400)
+  ),
+  background-color-danger: (
+    background-color var(--pf-global--palette--red-50)
+  ),
+  background-color-default: (
+    background-color var(--pf-global--palette--cyan-50)
+  ),
+  background-color-info: (
+    background-color var(--pf-global--palette--blue-50)
+  ),
+  background-color-success: (
+    background-color var(--pf-global--palette--green-50)
+  ),
+  background-color-warning: (
+    background-color var(--pf-global--palette--gold-50)
   )
 );
 
 @include pf-utility-builder($pf-u-background-color-options, $pf-global--breakpoint-list);
 
 // stylelint-enable
+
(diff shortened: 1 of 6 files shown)


src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -66,6 +64,81 @@ import './Text.css'
...
+<div class="pf-u-background-color-dark-400">
+  {{#> text text--modifier="pf-u-color-light-100"}}
+    Font color light 100
+  {{/text}}
+  {{#> text text--modifier="pf-u-color-light-200"}}
+    Font color light 200
+  {{/text}}
+</div>
+```
+### State/status text colors
```
mcoker, line 82: Nit - to match the bg utility example name

```suggestion
### Status and state text colors
```

src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -66,6 +64,81 @@ import './Text.css'
...
+### Disabled text colors
+```hbs
+{{#> text text--modifier="pf-u-disabled-color-100"}}
+  Font color disabled 100
+{{/text}}
+{{#> text text--modifier="pf-u-disabled-color-200"}}
+  Font color disabled 200
+{{/text}}
+{{#> text text--modifier="pf-u-disabled-color-300"}}
+  Font color disabled 300
```
mcoker, line 130: Question of where this belongs. This global var is used as a background-color with one exception as a border-color for tertiary nav (that we're deprecating). @mceledonia do you know the intent of this color and where it should belong?

src/patternfly/utilities/BackgroundColor/BackgroundColor.scss
```diff
@@ -19,9 +19,25 @@ $pf-u-background-color-options: (
   ),
   background-color-dark-400: (
     background-color var(--pf-global--BackgroundColor--dark-400)
+  ),
+  background-color-danger: (
+    background-color var(--pf-global--palette--red-50)
```
mcoker, line 24: Since we're using these as background colors for states, I wonder if these should be global vars. All of the colors for text and background colors currently map to global vars. @mceledonia wdyt? It wouldn't necessarily have to be a part of this PR though.

mcoker (changes requested):
Looks great! Left some comments.

> Link, Active, and Primary text colors were already listed in the documentation and implemented, but not in any examples. Do we want to show them?

Yeah, I would be in favor of having an example for each class for all of the utilities.

Issue #4255, opened 2021-07-30
Introduce Back to Top component


mcarrano:
Introduce a new component called Back to Top.  This will be a button that floats on top of content for long scrolling pages and navigates the user back to the top of the page in a single click.

Designs

![image](https://user-images.githubusercontent.com/6619860/127695654-261f784c-2310-406b-94a4-d1f6bb86deac.png)

Design issue: https://github.com/patternfly/patternfly-design/issues/970

PR #4291, opened 2021-08-11
feat(back-to-top): added back to top component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4255

example: https://patternfly-pr-4291.surge.sh/components/back-to-top
demo: https://patternfly-pr-4291.surge.sh/components/back-to-top/html-demos/basic/
  * The demo just shows the button placement in context of the page. Doesn't show any sort of interaction, which will all likely be handled via JS in the react component. 

Some questions:

* Will there be some kind of show/hide fade animation?
* It's easy to add this to a simple page that has a single scrollbar for the main content area with a bunch of content on it that scrolls - the button is just positioned at the bottom/right of the main content area. It's more difficult to add it to these types of pages:
  * [overflow scroll](https://www.patternfly.org/v4/components/page/html-demos/overflow-scroll/)
    * In this scenario, would the button be in the scrollable page section, and clicking would scroll that page section back to the top?
    * We'll likely need to introduce a new scroll wrapper. 
    * There is this technique, though it's a bit hacky... I'd want to test it more - https://codesandbox.io/s/fervent-stonebraker-txoy3?file=/style.css
  * [sticky bottom section](https://www.patternfly.org/v4/components/page/html-demos/sticky-section-bottom/)
    * By default, since the button is at the bottom/right of the main content area, it's probably going to overlap the sticky section at the bottom. 
    * JS could detect the height of any sticky sections like that and try to offset the button. That can get really tricky though because there may be a page section that's sticky, but may not always be sticky to the bottom if, for example, it isn't the last section on the page. That may also be more parsing and making assumptions about the children to the page than we're comfortable doing.

Do we want to try and support those use cases? Or would we just support a basic page that scrolls? We can also just start out with a super basic implementation and change it as needed.

cc @mcarrano @mceledonia @bgliwa01

### DIFF
--- a/src/patternfly/components/BackToTop/back-to-top.hbs
+++ b/src/patternfly/components/BackToTop/back-to-top.hbs
@@ -1,0 +1,11 @@
+<div class="pf-c-back-to-top{{#if back-to-top--modifier}} {{back-to-top--modifier}}{{/if}}"
+  {{#if back-to-top--attribute}}
+    {{{back-to-top--attribute}}}
+  {{/if}}>
+  {{#> button-link button-link--attribute=(concat 'href="#' back-to-top--target '"') button-link--modifier="pf-m-primary"}}
+    Back to top
+    {{#> button-icon button-icon--modifier="pf-m-end"}}
+      <i class="fas fa-angle-up" aria-hidden="true"></i>
+    {{/button-icon}}
+  {{/button-link}}
+</div>
--- a/src/patternfly/components/BackToTop/back-to-top.scss
+++ b/src/patternfly/components/BackToTop/back-to-top.scss
@@ -17,6 +17,11 @@
   position: absolute;
   right: var(--pf-c-back-to-top--Right);
   bottom: var(--pf-c-back-to-top--Bottom);
+  display: none;
+
+  &.pf-m-visible {
+    display: block;
+  }
 
   .pf-c-button {
     --pf-c-button--FontSize: var(--pf-c-back-to-top--c-button--FontSize);
--- a/src/patternfly/components/BackToTop/back-to-top.scss
+++ b/src/patternfly/components/BackToTop/back-to-top.scss
@@ -17,10 +17,9 @@
   position: absolute;
   right: var(--pf-c-back-to-top--Right);
   bottom: var(--pf-c-back-to-top--Bottom);
-  display: none;
 
-  &.pf-m-visible {
-    display: block;
+  &.pf-m-hidden {
+    display: none;
   }
 
   .pf-c-button {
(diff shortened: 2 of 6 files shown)


mattnolting (approved):
Perfecto 👍

bgliwa01:
lgtm! In terms of implementation - the component should only appear after the user has scrolled, so if the scroll bar is at the top this component shouldn't be shown.

srambach:
LGTM, and good questions above in the description.

mcarrano (commented):
Looks good @mcoker . Just a couple of questions.

- Does this have a drop shadow? I can't really tell, but the shadow looked more prominent in the designs.
- I agree with @bgliwa01 that the component should only appear after the user scrolls. Is that what you meant by animation? Would that be implemented here or require React?

Thanks for pointing out all the edge cases. I am OK with just introducing the basic version for now and addressing these situations if/when they come up. Do you agree?

mcoker:
@mcarrano 

The button is using the `--pf-global--BoxShadow--md-bottom` shadow. I agree it's hard to see though.

Current shadow

Screen Shot 2021-08-17 at 9 45 35 AM

The "large" shadow variant - want to use this instead?

Screen Shot 2021-08-17 at 9 46 08 AM

> the component should only appear after the user scrolls. Is that what you meant by animation? Would that be implemented here or require React?

It's a 2 part question - if it hides/shows on scroll, we'll add a class that allows toggling the visibility. React will handle toggling that class based on how far the user has scrolled. Then there is a question of how the button should appear - should it just appear (no animation), or should it have some sort of fade or other effect when it's hidden/shown?

mcarrano:
@bgliwa01 @mceledonia what do you think about the above two questions?

bgliwa01:
@mcarrano I think the larger shadow looks better. In terms of how the button appears, I think it would be okay to have no animation but being able to see it in a demo would be very helpful to decide.

mcoker:
@mcarrano @bgliwa01 updated the shadow. Also added a class to show the component (it's hidden by default). There is no animation. Once we see it in context in a react demo, we can add an animation if needed.

mcarrano:
That sounds good @mcoker . But I'm seeing a couple of issues with the demo.

- Missing screen shot.
- When I open the full screen demo, Back to top does not appear.

mcoker:
@mcarrano updated it to be visible by default, then hidden with a modifier. I think that makes more sense, and it fixes the demo. The demo screenshot will be updated after code freeze with the PR that updates the release notes.

srambach (approved):
LGTM 👍

mcarrano (approved):
Looks great. Thanks @mcoker !

Issue #3966, opened 2021-03-31
Inline clipboard copy - leave whitespace between wrapping rows of text


mcoker:
Re: https://github.com/patternfly/patternfly/pull/3933#issuecomment-811286752

<img width="845" alt="Screen Shot 2021-03-31 at 1 38 38 PM (1)" src="https://user-images.githubusercontent.com/35148959/113203318-86f22100-9231-11eb-9608-0abd48a529c1.png">

PR #4296, opened 2021-08-12
fix(clipboard-copy): allowed gap b/w wrapping lines in inline variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/3966

examples - https://patternfly-pr-4296.surge.sh/components/clipboard-copy#inline-compact-in-sentence

This only changes the default inline variation with text that wraps inline with other text. It does not change the block variation of inline clipboard copy - that variation looks like it will need a breaking change to rework the html/css to allow for gaps between wrapping lines. Likely something like this:

```
div.pf-c-clipboard-copy.pf-m-inline // set to block
  div.pf-c-clipboard-copy__main // set to inline with styles .pf-c-clipboard-copy.pf-m-inline has now (bgcolor, left/right padding, white-space, etc)
    span.pf-c-clipboard-copy__text
    span.pf-c-clipboard-copy__actions
...
```

### DIFF
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy.scss
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy.scss
@@ -17,8 +17,8 @@
   --pf-c-clipboard-copy__expandable-content--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
 
   // Inline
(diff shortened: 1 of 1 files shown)


mceledonia (approved):
Makes sense to me! Thanks for the update. I think for block formatting we can leave it with the background by default, but if the gap is needed it sounds like the regular inline copy can go in a div to achieve that.

mcarrano (approved):
This looks good to me. Thanks @mcoker !

srambach (approved):
LGTM 👍

PR #4304, opened 2021-08-17
chore(split): added wrappable demo

### LINKED ISSUE #4167: Add example of split layout with wrap modifier
We're missing an example of `.pf-l-split.pf-m-wrap`. We can copy the react example.

srambach:
Adds a demo for a wrappable split layout.

Fixes #4167


mcoker (approved):
LGTM!

Issue #4315, opened 2021-08-19
Remove log viewer border in dark mode


mcoker:
Per discussion with @DaoDaoNoCode and @mcarrano, the log viewer doesn't need a border in dark mode.

PR #4316, opened 2021-08-19
fix(log-viewer): removed border from log viewer in dark theme

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4315


DaoDaoNoCode (approved):
Looks 😎 to me, thanks for the quick fix @mcoker !!

Issue #4323, opened 2021-08-23
Change the style of log viewer component


DaoDaoNoCode:
Based on the refactor of log viewer in [PatternFly React](https://github.com/patternfly/patternfly-react/pull/6201), the design in the core repo should also be modified to align with the design in React repo.
Here is the todo list:
1. Get rid of `flex: 1 1` in class `.pf-c-log-viewer__main`.
2. Get rid of the gaps between elements in the log viewer, which makes them look like inline elements.

mcoker:
@DaoDaoNoCode just to confirm, do you mean the space between the header and the main element?

DaoDaoNoCode:
@mcoker Oh sorry, I mean the space between the lines in the main element. https://patternfly-react-pr-6201.surge.sh/extensions/log-viewer/#with-search See this example, there is no space between every line after the update in React repo.

PR #4327, opened 2021-08-24
fix(log-viewer): updated to support dynamic size list

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4323


DaoDaoNoCode (approved):
@mcoker Looks great to me, thanks for the fix!

Issue #4288, opened 2021-08-11
Form demo - bug?


mmenestr:
Was looking at helper text [HTML examples](https://www.patternfly.org/v4/components/form/html#help-text) and noticed inconsistencies in spacing. 

The bottom helper text example is inconsistent with the rest in two ways:
1. The spacing between the input field and the helper text is bigger than in the others (and in my mind, wrong)
2. The last example has an icon in front of the helper text, while the other examples do not.

Wondering if this is a bug, or an intentional thing? 

<img width="830" alt="Screen Shot 2021-08-11 at 4 09 18 PM" src="https://user-images.githubusercontent.com/51165119/129096243-10d39b84-af4c-407e-aec3-01b032171dbf.png">

I also noticed the added icon in the helper text in these examples... is this a new thing? How come some of the examples have it and some don't? Seems redundant when it's already present in the input field itself.

![image](https://user-images.githubusercontent.com/51165119/129096575-e3d3a835-a0e7-4ad7-a3ad-cb23c41b2cce.png)

cc: @arburka

mcoker:
> The spacing between the input field and the helper text is bigger than in the others (and in my mind, wrong)

@mmenestr that is a bug, @mcarrano can we add this issue to an upcoming milestone?

> I also noticed the added icon in the helper text in these examples... is this a new thing? How come some of the examples have it and some don't? Seems redundant when it's already present in the input field itself.

That's been a feature of the helper text element for a while. It was originally added to support use of helper text in the [login component](https://www.patternfly.org/v4/components/login-page/html/invalid/). This example on the form page is just demonstrating the ability to add an icon - I'm not sure about the redundancy of the icon also in the form field. It's visible in the react examples, too - https://www.patternfly.org/v4/components/form. @mcarrano do you think we should change the examples?

mcarrano:
@mcoker adding this issue to the next sprint. Regarding the helper text icon. Yeah, it seem redundant and contrary to our guidance. Can we just remove that from the examples and then we can follow up with a React issue?

mcoker:
> Can we just remove that from the examples and then we can follow up with a React issue?

@mcarrano yep. Though do you want to remove the icon or the status in the input fields? I'd like to keep an icon in one of the examples if we can so it's visible and not just referenced in the docs. Maybe we could add an icon to regular (black text) helper text that doesn't have a state associated with it?

PR #4329, opened 2021-08-24
fix(form-control): removed extra space under textarea

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4288

Not sure if I want to make this change now or in a breaking change. It seems like a safe enough change, I just wonder if it could cause some unexpected issue - and the form control elements are commonly used. Beyond fixing the extra space from the original issue, it fixes this inconsistency that currently exists under the element between firefox and either chrome or safari.

wdyt @mattnolting @srambach 

chrome
Screen Shot 2021-08-24 at 5 47 11 PM

safari
Screen Shot 2021-08-24 at 5 47 20 PM

firefox
Screen Shot 2021-08-24 at 5 47 31 PM

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -180,9 +180,11 @@ cssPrefix: pf-c-form
     {{/form-group-control}}
   {{/form-group}}
   {{#> form-group form-group--id="info"}}
-    {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"')}}
-      Information
-    {{/form-label}}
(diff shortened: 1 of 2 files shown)


srambach (approved):
LGTM, nice find.

mattnolting (approved):
Update looks good to me, although the example seems to be malformed. Looks like the label needs a `pf-c-form__group-label` wrapper. 

Screen Shot 2021-09-08 at 11 42 58 AM

Screen Shot 2021-09-08 at 11 42 47 AM

mcoker:
@mattnolting nice catch! updated

Issue #4146, opened 2021-06-14
Update sidebar split and stack variants to horizontal and vertical


mcoker:
From this thread in the description list PR implementing horizontal and vertical `orientation` prop values

https://github.com/patternfly/patternfly-react/pull/5885#discussion_r648663409

mcoker:
@mcarrano can we add this to the next milestone? It's an easy change and will need to be made in both core and react in the same sprint.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/4332

PR #4332, opened 2021-08-25
fix(sidebar): changed split/stack to horizontal/vertical

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4146

### DIFF
--- a/src/patternfly/components/Sidebar/examples/Sidebar.md
+++ b/src/patternfly/components/Sidebar/examples/Sidebar.md
@@ -32,7 +32,7 @@ import './Sidebar.css'
   {{/sidebar-panel}}
   {{#> sidebar-content}}
     {{#> content}}
-      <p>Adds a gutter between the panel and content when in the split layout.</p>
+      <p>Adds a gutter between the panel and content when in the horizontal layout.</p>
       <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse dapibus nulla id augue dictum commodo. Donec mollis arcu massa, sollicitudin venenatis est rutrum vitae. Integer pulvinar ligula at augue mollis, ac pulvinar arcu semper. Maecenas nisi lorem, malesuada ac lectus nec, porta pretium neque. Ut convallis libero sit amet metus mattis, vel facilisis lorem malesuada. Duis consectetur ante sit amet magna efficitur, a interdum leo vulputate.</p>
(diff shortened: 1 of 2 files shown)


srambach:
This seems a little confusing when compared to the drawer, which uses `.pf-m-vertical` and `.pf-m-horizontal` to describe the splitter orientation, which would be the opposite of how it is used here for the orientation of the content and sidebar. Where you already have a `.pf-m-panel-right` modifier, what about somehow making this consistent with how drawer uses `.pf-m-panel-left` and `.pf-m-panel-bottom` modifiers?

mcoker:
> This seems a little confusing when compared to the drawer, which uses .pf-m-vertical and .pf-m-horizontal to describe the splitter orientation, which would be the opposite of how it is used here for the orientation of the content and sidebar.

If I understand correctly, I think that's going to be the case since the splitter is horizontal when used in a vertical (top/bottom) drawer layout, and it's vertical when used in a horizontal (left/right) drawer layout. I would expect it to be used the same here, too - vertical splitter with horizontal/left/right sidebar, horizontal splitter with vertical/top/bottom sidebar.

> Where you already have a .pf-m-panel-right modifier, what about somehow making this consistent with how drawer uses .pf-m-panel-left and .pf-m-panel-bottom modifiers?

Are you suggesting the default orientation of this layout and the drawer match? (panel-right by default?) Or dropping horizontal/vertical and just using side (top, left, right, bottom) modifiers? Or maybe something else?

srambach:
I see what you mean about the "splitter" orientation. I think this is confusing because what you think of as a vertical sidebar here is using the horizontal modifier, because it's actually modifying the _layout_. This is so similar in concept with the drawer, I think as much consistency as possible would be good, so yeah, maybe the top/bottom/left/right would be more clear?
image

mcoker:
@mcarrano wdyt about the naming for this? Currently we're the variations for this component "horizontal" and "vertical" depending on whether the panel/content are side by side or the panel is stacked on top of the content. To @srambach's point, this can be confusing since we call it a "sidebar" component, and in the horizontal variation, the sidebar is vertical - and vice versa.

mcarrano:
@mcoker @srambach I really don't know what to say. Has anyone else complained that this is currently confusing? Was there a problem with "stacked" and "split"? I agree horizontal and vertical are confusing terms here.

mcoker:
@mcarrano I haven't heard any feedback about it. We were just trying to be consistent in our prop/value naming after implementing vertical/horizontal description list groups - https://github.com/patternfly/patternfly-react/pull/5885#pullrequestreview-680119708

I'm fine with leaving it as split and stack, too, since "vertical" can be confusing here. We use split and stack in other contexts, too - split/stack layouts, "stack" tile header, and "stack" form groups for stacking multiple form control elements.

mcoker:
Closing this for now since this isn't solving an obvious problem and could lead to more confusion.

PR #4333, opened 2021-08-25
fix(tabs): moved text into expandable toggle

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4174

I left legacy examples since we need to support both variations of the markup until a breaking change. I'll open a breaking change PR to remove the legacy support if this PR is merged.

### DIFF
--- a/src/patternfly/components/Tabs/examples/Tabs.md
+++ b/src/patternfly/components/Tabs/examples/Tabs.md
@@ -218,6 +218,30 @@ import './Tabs.css'
 {{/tabs}}
 ```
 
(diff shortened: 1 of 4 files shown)


mattnolting (changes requested):
Looks good, need to control long strings

Screen Shot 2021-09-08 at 11 30 18 AM

mcoker:
@mattnolting good catch, updated so the text wraps and the toggle button is aligned to the top to match the existing wrap behavior.

mattnolting (approved):
Great work, LPTM!

PR #4361, opened 2021-09-06
feat(dual-list-selector): add support for disabled items

KKoukiou:
Needed for: https://github.com/patternfly/patternfly-react/issues/6250

### DIFF
--- a/src/patternfly/components/DualListSelector/dual-list-selector-item.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-item.hbs
@@ -1,12 +1,14 @@
 {{#> dual-list-selector-list-item-row}}
-  <{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--IsDisabled}} pf-m-disabled{{/if}}{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
-    {{#unless dual-list-selector-item--IsDisabled}}
-        {{#if dual-list-selector-item--type}}
-          tabindex="0"
-        {{/if}}
+  <{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
+    {{#unless dual-list-selector-list-item--IsDisabled}}
+      {{#if dual-list-selector-item--type}}
+        tabindex="0"
+      {{/if}}
     {{/unless}}
-    {{#if dual-list-selector-item--IsDisabled}}
-      disabled
+    {{#if dual-list-selector-list-item--IsDisabled}}
+      {{#unless dual-list-selector-item--type}}
+        disabled
+      {{/unless}}
     {{/if}}
     {{#unless dual-list-selector-item--type}}
       type="button"
--- a/src/patternfly/components/DualListSelector/dual-list-selector-item.hbs
+++ b/src/patternfly/components/DualListSelector/dual-list-selector-item.hbs
@@ -1,7 +1,12 @@
 {{#> dual-list-selector-list-item-row}}
-  <{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
-    {{#if dual-list-selector-item--type}}
-      tabindex="0"
+  <{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--IsDisabled}} pf-m-disabled{{/if}}{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
+    {{#unless dual-list-selector-item--IsDisabled}}
+        {{#if dual-list-selector-item--type}}
+          tabindex="0"
+        {{/if}}
+    {{/unless}}
+    {{#if dual-list-selector-item--IsDisabled}}
+      disabled
     {{/if}}
     {{#unless dual-list-selector-item--type}}
       type="button"
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -53,7 +53,7 @@ $pf-c-dual-list-selector__item--MaxNesting: 10;
   --pf-c-dual-list-selector__item-text--Color: var(--pf-global--Color--100);
   --pf-c-dual-list-selector__list-item-row--m-selected__text--Color: var(--pf-global--active-color--100);
   --pf-c-dual-list-selector__list-item-row--m-selected__text--FontWeight: var(--pf-global--FontWeight--bold);
-  --pf-c-dual-list-selector__list-item-row--m-disabled__text--Color: var(--pf-global--disabled-color--100);
+  --pf-c-dual-list-selector__list-item--m-disabled__item-text--Color: var(--pf-global--disabled-color--100);
 
   // Status
   --pf-c-dual-list-selector__status--MarginBottom: var(--pf-global--spacer--sm);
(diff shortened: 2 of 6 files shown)


src/patternfly/components/DualListSelector/dual-list-selector-item.hbs
```diff
@@ -1,4 +1,4 @@
-<{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--HasCheck}} pf-m-check{{/if}}{{#if dual-list-selector-item--IsSelected}} pf-m-selected{{/if}}{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
+<{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--HasCheck}} pf-m-check{{/if}}{{#if dual-list-selector-item--IsSelected}} pf-m-selected{{/if}}{{#if dual-list-selector-item--IsDisabled}} pf-m-disabled{{/if}}{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
```
mcoker, line 1: When the item is a `<button>`, you should only need to add a `disabled` attribute, then all you need to do is style the text via `:disabled`.

But this item can also be a `<div>` for both expandable and checkbox items. Maybe in those cases, it's appropriate to add `.pf-m-disabled` here, remove the `tabindex` from this element, add `disabled` to the check,  and style the toggle and text via `.pf-m-disabled` as you've done here.

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -219,6 +220,15 @@ $pf-c-dual-list-selector__item--MaxNesting: 10;
     }
   }
 
+  &.pf-m-disabled {
+    .pf-c-dual-list-selector__item-text {
+      --pf-c-dual-list-selector__item-text--Color: var(--pf-c-dual-list-selector__item--m-disabled__text--Color);
+    }
+
+    pointer-events: none;
+    cursor: none;
```
mcoker, line 229: If you add `pointer-events: none;`, you shouldn't need to style the `cursor`

mcoker (changes requested):
Thanks for the PR @KKoukiou!!! Left some comments. Also added @mcarrano to review visually.

KKoukiou:
> This looks fine to me. However I'm not quite sure how this addresses [patternfly/patternfly-react#6250](https://github.com/patternfly/patternfly-react/issues/6250) which seems to be asking to disable the entire control. Would the idea be that `isDisabled` would just set all items to disabled so you couldn't select anything? I guess this works, but not sure when that's useful.

It's much more flexible to be able to disable each of the items, I can see such a scenario for cockpit for example, where some of the items couldn't be used. If you want to disable all you can still do that by disabling each of them separately.

mcarrano (approved):
LGTM!

mcoker:
@KKoukiou do you need to get this in for this release? If so we have a code freeze today. If we can update the styles/examples to use `:disabled` and remove the `cursor` styles, we can merge it and follow up on disabled expandable/check items.

KKoukiou:
@mcoker no it's not urgent from my side.

src/patternfly/components/DualListSelector/dual-list-selector.scss
KKoukiou: I copied this from src/patternfly/components/DataList/data-list.scss - we can fix this there as well then.

KKoukiou:
@mcoker I am not sure if I did the needed changes, let me know.

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -259,6 +260,14 @@ $pf-c-dual-list-selector__item--MaxNesting: 10;
   text-align: left;
   cursor: pointer;
   border: 0;
+
+  &.pf-m-disabled {
+    .pf-c-dual-list-selector__item-text {
+      --pf-c-dual-list-selector__item-text--Color: var(--pf-c-dual-list-selector__list-item-row--m-disabled__text--Color);
```
mcoker, line 266: You can just set this var under `&.pf-m-disabled` - should be no need to scope it to `&.pf-m-disabled .pf-c-dual-list-selector__item-text`

src/patternfly/components/DualListSelector/dual-list-selector-item.hbs
```diff
@@ -1,7 +1,12 @@
 {{#> dual-list-selector-list-item-row}}
-  <{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
-    {{#if dual-list-selector-item--type}}
-      tabindex="0"
+  <{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item{{#if dual-list-selector-item--IsDisabled}} pf-m-disabled{{/if}}{{#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
```
mcoker, line 2: For this block I think you want to add some checks around `pf-m-disabled` and the `disabled` attr. A button should have `disabled` and a div/non-button should have `pf-m-disabled`.

```
<{{#if dual-list-selector-item--type}}{{dual-list-selector-item--type}}{{else}}button{{/if}} class="pf-c-dual-list-selector__item
  {{~#if dual-list-selector-item--IsDisabled}}
    {{~#if dual-list-selector-item--type}} pf-m-disabled{{/if}} // add pf-m-disabled if it's a disabled div/non-button
  {{/if}}
  {{~#if dual-list-selector-item--modifier}} {{dual-list-selector-item--modifier}}{{/if}}"
  {{#unless dual-list-selector-item--IsDisabled}}
    {{#if dual-list-selector-item--type}}
      tabindex="0"
    {{/if}}
  {{/unless}}
  {{#if dual-list-selector-item--IsDisabled}}
    {{#unless dual-list-selector-item--type}}
      disabled // add disabled attr if it's a disabled button
```

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -259,6 +260,14 @@ $pf-c-dual-list-selector__item--MaxNesting: 10;
   text-align: left;
   cursor: pointer;
   border: 0;
+
+  &.pf-m-disabled {
```
mcoker, line 264: add `:disabled` to this:

```
&.pf-m-disabled, // non-buttons
&:disabled { // buttons
```

mcoker (changes requested):
Thanks for the updates @KKoukiou!! Left some feedback. Looks like you're close to supporting both buttons and expandable items (`div`s). Would you mind adding an example of a disabled expandable item? We will also need to disable all nested items of an expandable item.

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -53,6 +53,7 @@ $pf-c-dual-list-selector__item--MaxNesting: 10;
   --pf-c-dual-list-selector__item-text--Color: var(--pf-global--Color--100);
   --pf-c-dual-list-selector__list-item-row--m-selected__text--Color: var(--pf-global--active-color--100);
   --pf-c-dual-list-selector__list-item-row--m-selected__text--FontWeight: var(--pf-global--FontWeight--bold);
+  --pf-c-dual-list-selector__list-item-row--m-disabled__text--Color: var(--pf-global--disabled-color--200);
```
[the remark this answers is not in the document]
mcarrano: @mdoruskova @mceledonia can you confirm what the correct color for disabled menu items should be?

KKoukiou:
@mcoker thanks a lot for the hints! Updated according to your comments. Sorry for the initial mess - I am not very much in touch with this repo.

mcoker (changes requested):
Just realized we will need a disabled class at least on the list-item-row element since that's where we change the background color of the item on hover. I wanted to run through the changes to make sure it will work so I issued a PR to your branch here if you'd like to see what I did - https://github.com/KKoukiou/patternfly/pull/1

It also:
* disables the check for disabled items
* applies a disabled color to an expandable toggle icon
* limits the `disabled` attribute to items that are buttons
* moves `pf-m-disabled` to a top level `__list-item` element

KKoukiou:
> Just realized we will need a disabled class at least on the list-item-row element since that's where we change the background color of the item on hover. I wanted to run through the changes to make sure it will work so I issued a PR to your branch here if you'd like to see what I did - [KKoukiou#1](https://github.com/KKoukiou/patternfly/pull/1)
> 
> It also:
> 
>     * disables the check for disabled items
> 
>     * applies a disabled color to an expandable toggle icon
> 
>     * limits the `disabled` attribute to items that are buttons
> 
>     * moves `pf-m-disabled` to a top level `__list-item` element

Ah cool :) As I am really strugling with CSS I feel I am abusing your review to learn here. thanks for fixing it for me!

mcoker (approved):

srambach (approved):
LGTM, thanks for the docs update.

PR #4367, opened 2021-09-09
fix(progress-stepper): fixed global color var name

mcoker:


redallen (approved):
Good work, detective! :mag_right:

PR #4375, opened 2021-09-15
fix(radio,checkbox,form): Correct label alignment

### LINKED ISSUE #3868: form: vertical alignment issues
1. There are issues with the checkboxes and radio buttons not being properly aligned with their textual labels. This is slightly different per browser.
2. Labels for form elements are down too low (or are elements too high?)

Included are screenshots that show these issues, from Firefox:

In Cockpit:

Screenshot from 2021-02-17 13-00-25

Screenshot from 2021-02-17 13-00-16

On the PatternFly website:

Screenshot from 2021-02-17 13-08-19

Screenshot from 2021-02-17 13-07-27

Expect results would be:
- Checkbox and radio buttons should look vertically centered to the first line of text
- Labels and form elements should be aligned to baseline

srambach:
This does some math to calculate the distance from the top of the element to the baseline so that the horizontal form labels and also the checkbox/radio button labels can be properly aligned vertically.
Sadly, in Firebox, aligning the label/input to "baseline" doesn't give the desired behavior on multi-line inputs.

Old and new alignment here:
image

Fixes #3868

### DIFF
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -5,8 +5,8 @@
   --pf-c-check__label--FontWeight: var(--pf-global--FontWeight--normal);
   --pf-c-check__label--FontSize: var(--pf-global--FontSize--md);
   --pf-c-check__label--LineHeight: var(--pf-global--LineHeight--sm);
-  --pf-c-check__input--Height: calc(var(--pf-c-check__label--FontSize) * var(--pf-c-check__label--LineHeight));
-  --pf-c-check__input--MarginTop: calc(-1px * (var(--pf-c-check__label--LineHeight) * 1.25));
+  --pf-c-check__input--Height: var(--pf-c-check__label--FontSize);
+  --pf-c-check__input--MarginTop: calc(((var(--pf-c-check__label--FontSize) * var(--pf-c-check__label--LineHeight)) - var(--pf-c-check__input--Height)) / 2);
   --pf-c-check__description--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-check__description--Color: var(--pf-global--Color--200);
   --pf-c-check__body--MarginTop: var(--pf-global--spacer--sm);
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -11,8 +11,32 @@
   --pf-c-form--m-limit-width--MaxWidth: #{pf-size-prem(500px)};
 
   // Group label
-  --pf-c-form--m-horizontal__group-label--md--PaddingTop: var(--pf-global--spacer--sm);
+
+  // Terminology and a note on vertical alignment of the group labels:
+  // Content area height = Font size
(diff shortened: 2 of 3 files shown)


garrett (approved):
@srambach: This is **_wonderful!_**

I took a few screenshots on Fedora Linux 35 (beta):

## Firefox

Screenshot from 2021-09-16 16-21-44


## GNOME Web (WebKit, basically Linux's Safari)

Screenshot from 2021-09-16 16-23-21

 
## Chromium

Screenshot from 2021-09-16 16-24-33

---

It looks _perfect_ in all the browsers except for the "Name" field in Chromium. It's fine in Information and looks good in the options as well, so I think there's a half pixel rounding error somewhere, where it rounds inconsistently in some cases.

It's definitely an improvement and fixes it everywhere (except for that pixel rounding in Chrome for whatever reason... I'm guessing a minor browser or font rendering bug), so I wholeheartedly approve!

garrett:
I have a Windows VM I'm launching right now to try to see what happens in browsers in Windows versus Linux, just as a third point of comparison (assuming you're using a Mac, since you have Safari examples).

srambach:
> I have a Windows VM I'm launching right now to try to see what happens in browsers in Windows versus Linux, just as a third point of comparison (assuming you're using a Mac, since you have Safari examples).

Yes, I'm on a mac. Thank you for the additional testing!

garrett:
## Windows Firefox

Screenshot from 2021-09-16 16-59-19

Oddly, it's all off a pixel... some up, some down.

## Windows Edge (Chromium)

Screenshot from 2021-09-16 16-59-31

Chrome has the same off-by-one on the first input ("Name"), similar to Linux.

---

It seems Windows renders a little different from Linux and Mac. Still, this is already a lot better than what we currently have!

srambach:
Here's MS Edge Canary on my Mac.
image

garrett:
Oh, interesting! It seems like Chromium has fixed the rendering issue then? I'll try newer versions here on Linux too.

Also, the Edge I was testing is Chromium edge, just to be clear (not the old EdgeHTML Edge).

garrett:
## Windows Chrome Canary
Screenshot from 2021-09-16 17-50-13

It looks like someone fixed the rounding error in somewhere between Chromium stable (and Edge Beta on Linux) and Chromium Canary builds!

## Windows Firefox Nightly

Screenshot from 2021-09-16 17-51-51

Firefox on Windows is just weird for some reason.

---

OK, assuming the rounding error we see fixed in Canary builds of Chromium also show up in Linux (which renders otherwise the same as the other platforms it seems), that would mean that you've achieved perfect placement of the labels in all browsers except Firefox on Windows specifically, right?

I'd suggest we go forward with this and just ignore Firefox on Windows. It's only off by a pixel, which is still better than being off by two. And everything else (including Firefox on other platforms) can benefit from the better calculation!

Thanks again! :+1:  :tada:

mcoker (approved):
LGTM! 🤯

Issue #4358, opened 2021-09-03
Update selected states for Cards and Tiles


mcarrano:
Want to update selected states for cards and titles to make less confused with focus state. See design proposal here: https://github.com/patternfly/patternfly-design/issues/1042

PR #4380, opened 2021-09-17
fix(card): updated hoverable/selectable/selected styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4358


mcarrano:
@matthewcarleton wondered if you can take a look at this. Some time ago you filed the original design issue that prompted us to rethink this. The proposed solution is a very different visual approach from what we do now and I wanted to see what you thought and how this change might impact you from a product perspective. The best way to see this would be in this demo where the first card is selected and other cards display the new hover state. https://patternfly-pr-4380.surge.sh/demos/primary-detail/html-demos/primary-detail-card-view-expanded/ There is an animation when you mouse down on a card to select it. Let us know your thoughts.

matthewcarleton:
@mcarrano ya I think this works, the only thing I'd point out is if you are hovering over a card that is cut off by the viewport you don't see the hover effect.

mcarrano:
That's a really good point @matthewcarleton . Thoughts @mcoker @mceledonia ? Maybe we still retain the shadow together with the lower border on-Hover?

mcarrano:
I like that a lot better. What do others think?

PR #4383, opened 2021-09-20
feat(banner): add link style

### LINKED ISSUE #4360: Add special link styling for banner text
Want to introduce special styling for links when placed inside of a banner. See https://github.com/patternfly/patternfly-design/issues/1029 for discussion and design proposal.

srambach:
This adds styles for inline links in the banner.
@mceledonia @lboehling I don't know if we want to show all these examples or not, but I put them there for you to take a look at. Thanks!

Fixes #4360

### DIFF
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -57,4 +57,18 @@
     z-index: var(--pf-c-banner--m-sticky--ZIndex);
     box-shadow: var(--pf-c-banner--m-sticky--BoxShadow);
   }
+
+  .pf-c-button.pf-m-link.pf-m-inline {
+    --pf-c-button--FontWeight: var(--pf-global--FontWeight--semi-bold);
+    --pf-c-button--m-link--Color: var(--pf-c-banner--Color);
+    --pf-c-button--m-link--m-inline--hover--Color: var(--pf-c-banner--Color);
+    --pf-c-button--disabled--Color: var(--pf-c-banner--Color);
+
+    &:disabled {
+      --pf-c-button--m-link--m-inline--hover--TextDecoration: none;
+
+      pointer-events: auto;
+      cursor: not-allowed;
+    }
+  }
 }
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -70,20 +70,20 @@
     color: var(--pf-c-banner--link--Color);
 
     &:hover {
-      color: var(--pf-c-banner--link--hover--Color);
+      --pf-c-banner--link--Color: var(--pf-c-banner--link--hover--Color);
     }
 
     // stylelint-disable selector-no-qualifying-type
     &.pf-m-disabled {
-      color: var(--pf-c-banner--link--disabled--Color);
+      --pf-c-banner--link--Color: var(--pf-c-banner--link--disabled--Color);
+
       text-decoration: none;
-      pointer-events: auto;
       cursor: not-allowed;
     }
     // stylelint-enable selector-no-qualifying-type
   }
 
-  .pf-c-button.pf-m-link {
+  .pf-m-inline {
     --pf-c-button--FontWeight: var(--pf-c-banner--link--FontWeight);
     --pf-c-button--m-link--Color: var(--pf-c-banner--link--Color);
     --pf-c-button--m-link--m-inline--hover--Color: var(--pf-c-banner--link--hover--Color);
@@ -93,7 +93,6 @@
     &.pf-m-disabled {
       --pf-c-button--m-link--m-inline--hover--TextDecoration: none;
 
-      pointer-events: auto;
       cursor: not-allowed;
     }
   }
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -15,10 +15,12 @@
   }
 
   // banner link variables
-  --pf-c-banner--link--FontWeight: var(--pf-global--FontWeight--semi-bold);
   --pf-c-banner--link--Color: var(--pf-c-banner--Color);
+  --pf-c-banner--link--TextDecoration: underline;
   --pf-c-banner--link--hover--Color: var(--pf-c-banner--Color);
+  --pf-c-banner--link--hover--FontWeight: var(--pf-global--FontWeight--semi-bold);
   --pf-c-banner--link--disabled--Color: var(--pf-c-banner--Color);
+  --pf-c-banner--link--disabled--TextDecoration: none;
 
   // modifier variables
   --pf-c-banner--m-info--BackgroundColor: var(--pf-global--palette--blue-200);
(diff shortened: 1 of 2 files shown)


lboehling:
looks great!!

src/patternfly/components/Banner/banner.scss
```diff
@@ -57,4 +57,18 @@
     z-index: var(--pf-c-banner--m-sticky--ZIndex);
     box-shadow: var(--pf-c-banner--m-sticky--BoxShadow);
   }
+
+  .pf-c-button.pf-m-link.pf-m-inline {
```
mcoker, line 61: Looks like the styles added are limited to the inline link button component - is that intended @mcarrano or should a user be able to add a regular link, too?

src/patternfly/components/Banner/banner.scss
```diff
@@ -57,4 +57,18 @@
...
+    --pf-c-button--FontWeight: var(--pf-global--FontWeight--semi-bold);
+    --pf-c-button--m-link--Color: var(--pf-c-banner--Color);
```
mcoker, line 63: I would create vars for these values instead of reusing the banner text color and using a global var for the font-weight. Then they're customizable by the user via a var without having to write out the `.pf-c-button.pf-m-link.pf-m-inline` selector and redefining the button vars.

Maybe something like

```
// top of scss
--pf-c-banner--link--FontWeight: var(--pf-global--FontWeight--semi-bold);
--pf-c-banner--link--Color: var(--pf-c-banner--Color);
--pf-c-banner--link--hover--Color: var(--pf-c-banner--Color);
--pf-c-banner--link--disabled--Color: var(--pf-c-banner--Color);

// here
--pf-c-button--FontWeight: var(--pf-c-banner--link--FontWeight);
--pf-c-button--m-link--Color: var(--pf-c-banner--link--Color);
...

File: src/patternfly/components/Banner/banner.scss (hunk 1 above)
Comment (mcarrano) [reply src/patternfly/components/Banner/banner.scss:61]: I was assuming a regular link.

File: src/patternfly/components/Banner/banner.scss (hunk 2 above)
Comment (srambach) [reply src/patternfly/components/Banner/banner.scss]: I can remove .pf-m-inline, but  `.pf-c-button.pf-m-link` from the button component will override the variables if I don't create more specificity here.

File: src/patternfly/components/Banner/banner.scss (hunk 1 above)
Comment (srambach) [reply src/patternfly/components/Banner/banner.scss]: I added that. @mcarrano which examples do you want to show? Do we need all these? I put them there to make sure everything works but I'm not sure if we need to show them all as examples or not.

Comment (mcarrano) [reply src/patternfly/components/Banner/banner.scss]: @srambach Yes, I do feel like we have too many examples here. What's the difference between a link and an inline link? I think that the Call to action example should be removed as it's not something we recommend that banners be used for. I'm on the fence about whether we need to allow for both inline text links and link buttons. The banner was originally designed to hold short text messages like an alert. It's reasonable that you might include a text link in that message. A button implies to me some action that probably does not belong on a banner. Put I'm OK to leave it there as there might be a use case I'm not considering.

@mceledonia @lboehling I also want to get your input/feedback about the styling. Looks like the current approach is to make the link text bold and then underline on hover. I'm not sure that bold text affords a link to me. Would just underlining the link be more clear (in it's default state)?

Comment (srambach) [reply src/patternfly/components/Banner/banner.scss]: Also, although I was able to make the not-allowed cursor on the disabled button link, there are good a11y reasons not to try to make a link look disabled. Also I had to turn pointer events back on to do the cursor change for the link button, which isn't really good to do. So maybe we forget about that part?
And if we don't want to support button links, I can either take that out or leave it in just in case, and remove the examples. 
I'll wait to hear about about the bold/underline.

File: src/patternfly/components/Banner/banner.scss [hunk 3]
```diff
@@ -57,4 +64,37 @@
     z-index: var(--pf-c-banner--m-sticky--ZIndex);
     box-shadow: var(--pf-c-banner--m-sticky--BoxShadow);
   }
+
+  a {
+    font-weight: var(--pf-c-banner--link--FontWeight);
+    color: var(--pf-c-banner--link--Color);
+
+    &:hover {
+      color: var(--pf-c-banner--link--hover--Color);
```
mcoker, line 73:
```suggestion
      --pf-c-banner--link--Color: var(--pf-c-banner--link--hover--Color);
```

src/patternfly/components/Banner/banner.scss
```diff
@@ -57,4 +64,37 @@
...
+    }
+
+    // stylelint-disable selector-no-qualifying-type
+    &.pf-m-disabled {
+      color: var(--pf-c-banner--link--disabled--Color);
```
mcoker, line 78:
```suggestion
      --pf-c-banner--link--Color: var(--pf-c-banner--link--disabled--Color);
```

src/patternfly/components/Banner/banner.scss
```diff
@@ -57,4 +64,37 @@
...
+      text-decoration: none;
+      pointer-events: auto;
+      cursor: not-allowed;
+    }
+    // stylelint-enable selector-no-qualifying-type
+  }
+
+  .pf-c-button.pf-m-link {
```
mcoker, line 86: targeting `.pf-m-inline` is probably more ideal here since that's what we want to style. I suspect we want to leave the plain link button variant alone, as we would the other button variants like primary, secondary, tertiary, etc. my previous comment was just about being able to target the inline button links using link vars instead of always being tied to the banner color, which you've done here.

src/patternfly/components/Banner/examples/Banner.md
```diff
@@ -36,6 +36,64 @@ cssPrefix: pf-c-banner
...
+  {{/button}}
+{{/banner}}
+<br>
+{{#> banner banner--modifier="pf-m-danger"}}
+  A long danger banner with both text and {{#> button button--IsInlineLinkSpan="true" button--attribute='tabindex="0"' button--modifier="pf-m-link pf-m-inline"}}
+    long button text that needs to be a span so that it will wrap inline with the text around it.
+  {{/button}}
+{{/banner}}
+<br>
+{{#> banner banner--modifier="pf-m-info"}}
```
mcoker, line 87: looks like @mcarrano wants to remove this example

mcoker (changes requested):
LGTM! Just a couple of small things.

src/patternfly/components/Banner/examples/Banner.md
srambach: @mcoker @mcarrano Do we actually need most of these examples? I wanted to make sure they all worked, but should we simply add the one preferred method (which one?) of using a link within the banner into one or all of the existing examples?

mcarrano: @srambach @mcoker I would be in favor of that. Maybe just update the existing examples to include a link in the message text. That way you would see what the link looks like for each color background.

src/patternfly/components/Banner/banner.scss
```diff
@@ -57,4 +66,45 @@
...
+    &.pf-m-disabled {
+      --pf-c-banner--link--Color: var(--pf-c-banner--link--disabled--Color);
+      --pf-c-banner--link--TextDecoration: var(--pf-c-banner--link--disabled--TextDecoration);
+
+      cursor: not-allowed;
+    }
+    // stylelint-enable selector-no-qualifying-type
+  }
+
+  .pf-m-inline {
```
mcoker, line 90: This needs to be scoped so it doesn't apply styles to some other component with `.pf-m-inline`

mcoker (changes requested):

mcoker:
@mcarrano ready for re-review

mcarrano (approved):
Looks good to me. Thanks for making these changes @srambach .

mcoker (approved):
LGTM!!!

PR #4389, opened 2021-09-21
fix(compact-tree): addressed followup issues

### LINKED ISSUE #4277: Compact tree view follow-ups
* Need to add documentation for `.pf-c-tree-view__node-content`
* Not sure why the guide pseudo element needs a `z-index`. It makes sense that it would probably need one, but I disabled it in dev tools and as far as I can tell, it looks the same. Also if we're using a `z-index`
  * Not sure why we need `--pf-c-tree-view--m-guides--guide-zindex--base` since it's only used once. Seems like we could just define the `z-index` value in `--pf-c-tree-view--m-guides--guide--ZIndex`.
  * `--pf-global--ZIndex--sm` is the `z-index`. That's going to potentially conflict with component menus (dropdown, select, etc). I don't see any `z-index`es (indices? indexii?) defined anywhere else, so seems like we should be able to use `--pf-global--ZIndex--xs`, which is typically for manipulating the stacking context of adjacent types of things.

mattnolting:
closes #4277


mcoker (approved):
LPTM!!

Issue #4027, opened 2021-04-26
request: list select, similar to <select multiple>


garrett:
We've been using some list select widgets, implemented with <select multiple> for some time in Cockpit. In most cases, we only have one item selected at a time. In others, we allow multiple selections.

garrett:
GitHub is acting funky and stripped all formatting and images, making the feature request hard to read and understand. I'll try to correct here.

It looked like this before:
![114021483-0c4c8580-9871-11eb-90f9-257c8f73ef8d](https://user-images.githubusercontent.com/10246/116102516-106d0580-a6af-11eb-9139-1c2f69300281.png)

Some of the devs used the _menu_ component (which is wrong, but has the functionality we're looking for):
![114021588-25553680-9871-11eb-8d85-b4d58145b1a7](https://user-images.githubusercontent.com/10246/116102511-0fd46f00-a6af-11eb-9828-08ec00a81f4b.png)

I mad some CSS to override the style a bit, to make it look like what a list select widget might look like:
![116096952-fed52f00-a6a9-11eb-9e4d-05aa1523264c](https://user-images.githubusercontent.com/10246/116102518-106d0580-a6af-11eb-8c99-ea138ed152b4.png)

KKoukiou:
Some of the devs; /me hiding

garrett:
I wasn't pointing fingers. :point_left: :point_up: :point_right:  :wink:

garrett:
I'd imagine that we should have the default version without the filter, but have an option to have a filter above it too, like pictured above.

It would also be similar to the new treeview, but without the tree. I'm not sure how much this could inherit from the menu or the treeview, but perhaps we can build it off of the parts of one or both of them? _Or_ make a special mode in the treeview that would do this selection too?

mcoker:
@garrett have you looked at the simple list component? https://www.patternfly.org/v4/components/simple-list

cc @mcarrano

garrett:
Oh, I hadn't considered simple list, but it does have these issues that make it not quite right either:

- font size is `sm` instead of the normal `md`
- it wasn't obvious that it should be used for a scrolling a long list of items
- there's no frame around it indicating that it's a region (especially important when scrolling)
- there isn't a way to ask for a search area above it for filtering
- the style where a selected item becomes bold is a bit different from the rest of PF? (various selected items get one or more of the following: background highlight, check next to them, and/or bold)

[...]

PR #4392, opened 2021-09-21
feat(menu): added scrollable and plain modifiers

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4027

### DIFF
--- a/src/patternfly/components/Menu/examples/Menu.md
+++ b/src/patternfly/components/Menu/examples/Menu.md
@@ -101,48 +101,20 @@ import './Menu.css'
 {{/menu}}
 ```
 
-### With expanded toggle
+### Scrollable
 ```hbs
-{{#> menu}}
+{{#> menu menu--modifier="pf-m-scrollable"}}
   {{#> menu-content}}
-    {{#> menu-list}}
-      {{#> menu-list-item}}
-        {{#> menu-item}}
-          {{#> menu-item-main}}
-            {{#> menu-item-text}}
-              From Git
-            {{/menu-item-text}}
-            {{> menu-item-toggle-icon}}
-          {{/menu-item-main}}
-        {{/menu-item}}
-      {{/menu-list-item}}
-      {{#> menu-list-item menu-list-item--IsDisabled="true"}}
-        {{#> menu-item}}
-          {{#> menu-item-main}}
-            {{#> menu-item-text}}
-              Container image
-            {{/menu-item-text}}
-            {{> menu-item-toggle-icon}}
-          {{/menu-item-main}}
-        {{/menu-item}}
-      {{/menu-list-item}}
-      {{#> menu-list-item}}
-        {{#> menu-item}}
-          {{#> menu-item-main}}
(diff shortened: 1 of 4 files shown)


mcarrano (commented):
I think that the basic plain version without the border looks great. It does get a little weird with the search and the footer when you remove the outer boarder. What do you think @garrett . Maybe we would remove all borders/dividers? But then the scroll bar floats.

@mceledonia can you also take a look?

mceledonia:
It looks good to me, for the search/footer version I'm thinking given that it will appear in some other context it should work out. Do you have any examples of where that would be used? I'm having a tough time visualizing it.

mcoker:
@mcarrano @mceledonia, @garrett has an example in the original issue - https://github.com/patternfly/patternfly/issues/4027#issuecomment-826896486

Ideally the menu will be in some kind of context box where it will make sense if the menu is scrollable. The only thing I think looks weird is that the footer currently always has a box shadow. If the menu isn't scrollable, it looks like this:

Screen Shot 2021-09-21 at 12 38 33 PM

We could probably solve that by creating a variation for a scrollable menu, which we aren't doing now. A currently user makes the menu scrollable by passing a height to the content area instead of opting into a scrollable variation. If we had a scrollable variation, we could remove the shadow on the footer and use a border there instead, both in this plain variation as well as the regular menu.

Here's what the regular menu looks like now with a footer when the content isn't scrollable:

Screen Shot 2021-09-21 at 12 40 34 PM

And here's what my proposed change would look like:

Screen Shot 2021-09-21 at 12 40 58 PM

This would also allow us to set a reasonable default max-height for the scrollable menu, and the user could override that if they want something different. Currently if they want a scrollable menu, they have to set the max-height manually (there is no default).

cc @mattnolting wdyt?

mcarrano:
@mcoker I agree that the single line divider makes more sense than the shadow. Interested to hear what others think.

garrett:
For what it's worth, here's how our override looks in Cockpit, in the (filterable) language selector dialog:

Screenshot 2021-09-22 at 10-29-16 Overview - garrett Bolt

Screenshot 2021-09-22 at 10-29-31 Overview - garrett Bolt

We don't have extra borders around anything. It's styled like a search input plus a scrollable, selectable region (with a border around it, otherwise it would look broken).

mcoker:
@mcarrano updated the component to:

* Add a "scrollable" variation - by default when the variation is enabled, the menu height is 300px. Added new examples of that.
* Made the menu footer have a top border by default instead of a box shadow.
* When the menu is set to the scrollable variation, the footer has a top shadow and bottom border.

mcarrano:
I think this looks much better @mcoker . Only one small nit. Shouldn't the divider above the scrollable list also have a shadow in this example? https://patternfly-pr-4392.surge.sh/components/menu/#plain-scrollable-with-search-and-footer  What do you think @mceledonia ?

mcarrano (approved):
Please disregard my last comment. This is good to go. Thanks @mcoker !

srambach (approved):
Looks great!

Issue #4396, opened 2021-09-22
Preserve white space in log viewer text


mcoker:
From @DaoDaoNoCode, we need to preserve white space in the log entries to be able to accurately determine the width of each row. Also some log entries are multiple lines and look weird if the white space isn't preserved.
<img width="1286" alt="Screen Shot 2021-09-22 at 4 06 14 PM" src="https://user-images.githubusercontent.com/35148959/134422115-5d0b9a93-4b9b-4e95-80fe-317f42d5ca78.png">

PR #4397, opened 2021-09-22
fix(log-viewer): preserved white space in log viewer text

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4396

@redallen when you have an html element in an example block, it renders a new line before the element. Is there a way we can fix that? It's messing up the formatting in some of these examples.
Screen Shot 2021-09-22 at 4 56 54 PM

### DIFF
--- a/src/patternfly/components/LogViewer/log-viewer-string.hbs
+++ b/src/patternfly/components/LogViewer/log-viewer-string.hbs
@@ -1,6 +1,0 @@
-<span class="pf-c-log-viewer__string{{#if log-viewer-string--modifier}} {{log-viewer-string--modifier}}{{/if}}"
-  {{#if log-viewer-string--attribute}}
-    {{{log-viewer-string--attribute}}}
-  {{/if}}>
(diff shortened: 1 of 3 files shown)


DaoDaoNoCode (approved):
LGTM, thanks @mcoker !

gautamkrishnar (approved):

redallen:
@mcoker Unfortunately there's no option for that in https://github.com/Prettyhtml/prettyhtml#readme . We could of course fork it and implement such an option, but some other formatting might suffer.

mcoker:
@redallen bummer. I tried adding `<!-- prettier-ignore -->` in the examples and including `src/patternfly/LogViewer` in `.prettierignore` (though I'm guessing that's not the right path?) - didn't seem to work. Is there a way we can exclude a block or file from formatting?

redallen:
@mcoker Sure, I could add something like that!

redallen:
Here's a PR with it: https://github.com/patternfly/patternfly/pull/4401

Issue #4376, opened 2021-09-15
Implement styles for DragDrop component


nicolethoen:
Follow up work for https://github.com/patternfly/patternfly-react/pull/6258

redallen:
There are 2 components, `<div className="draggable">` and `<div className="droppable">`. Interactions and DOM structure can be found [here.](http://patternfly-react.surge.sh/components/drag-and-drop)

Droppables have 2 states with box shadows, valid and invalid.

Draggables also have 2 states with box shadows, valid and invalid. Invalid should have `cursor: 'not-allowed'`. There is an additional 3rd state of being dragged. Non-dragging items have animations for moving as the user drags over them. The dragging item has another animation for returning home after an invalid drop.

You can see the various [`setStyle` tags implemented in React](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/DragDrop/Draggable.tsx#L130) to see what exactly is being used now. Some things like draggable's `transform: translate(mouseX, mouseY)` and `background: getDefaultBackground()` will have to stay, but the rest should be classes!

mcarrano:
Here are the designs for reference: https://marvelapp.com/prototype/h198be4/screen/81623894

PR #4398, opened 2021-09-22
feat(drag-drop): added drag drop

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4376

#### draggable
* `.pf-c-draggable` is a draggable thing
* `.pf-m-dragging` when it's being dragged
* `.pf-m-drag-outside` when its out of bounds

#### droppable
* `.pf-c-droppable` is a droppable thing
* `.pf-m-dragging` when it's being dragged over
* `.pf-m-drag-outside` when the dragged thing is out of bounds

### DIFF
--- a/src/patternfly/patternfly-drag-drop.scss
+++ b/src/patternfly/patternfly-drag-drop.scss
@@ -1,4 +1,5 @@
 .pf-c-draggable {
+  --pf-c-draggable--Cursor: auto;
   --pf-c-draggable--m-dragging--Cursor: grabbing;
   --pf-c-draggable--m-dragging--BoxShadow: var(--pf-global--BoxShadow--md);
   --pf-c-draggable--m-dragging--after--BorderWidth: var(--pf-global--BorderWidth--sm);
@@ -6,9 +7,12 @@
   --pf-c-draggable--m-drag-outside--Cursor: not-allowed;
   --pf-c-draggable--m-drag-outside--after--BorderColor: var(--pf-global--danger-color--100);
 
+  cursor: var(--pf-c-draggable--Cursor);
+
   &.pf-m-dragging {
+    --pf-c-draggable--Cursor: var(--pf-c-draggable--m-dragging--Cursor);
+
     position: relative;
-    cursor: var(--pf-c-draggable--m-dragging--Cursor);
     background-color: var(--pf-c-draggable--m-dragging--BackgroundColor); // intentionally empty to be set if doesn't exist
     box-shadow: var(--pf-c-draggable--m-dragging--BoxShadow);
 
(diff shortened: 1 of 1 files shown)


redallen (approved):
Very good first pass! Hopefully it works without needing more changes :crossed_fingers:

mcoker:
@redallen this should work fine as a standalone thing and with the data-list to test out. Though not sure how it's going to work used with other components, but we can figure that out. I meant to include this codepen link which has the SCSS in it as a demo https://codepen.io/mcoker/pen/gORjzaE?editors=1100

src/patternfly/patternfly-drag-drop.scss
```diff
@@ -0,0 +1,63 @@
+.pf-c-draggable {
+  --pf-c-draggable--m-dragging--Cursor: grabbing;
```
mattnolting, line 2: Should you set `--pf-c-draggable--Cursor: grab;` by default, then update to `grabbing`?

src/patternfly/patternfly-drag-drop.scss
```diff
@@ -0,0 +1,63 @@
...
+  --pf-c-draggable--m-dragging--BoxShadow: var(--pf-global--BoxShadow--md);
+  --pf-c-draggable--m-dragging--after--BorderWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-draggable--m-dragging--after--BorderColor: var(--pf-global--active-color--100);
+  --pf-c-draggable--m-drag-outside--Cursor: not-allowed;
+  --pf-c-draggable--m-drag-outside--after--BorderColor: var(--pf-global--danger-color--100);
+
```
mattnolting, line 8:
```suggestion

cursor: var(--pf-c-draggable--Cursor);

```

src/patternfly/patternfly-drag-drop.scss
mattnolting:
```suggestion
  --pf-c-draggable--Cursor: grab;
```

mattnolting (commented):
Nice work! Left a few comments.

src/patternfly/patternfly-drag-drop.scss
mcoker: I wasn't thinking so, unless it's a thing in the draggable element that is only draggable (like the draggable icon). Because I imagine you'd still want your normal cursor for regular stuff in the draggable item (links, text, etc), since that's probably the most common interaction with the draggable item. It would change to dragging once we've detected you probably want to drag the thing. wdyt @redallen

src/patternfly/patternfly-drag-drop.scss
mcoker: Added this but set it to auto per https://github.com/patternfly/patternfly/pull/4398/files#r716069162.

PR #4399, opened 2021-09-23
fix(menu): updated flyout button to anchor

### LINKED ISSUE #4306: Examine flyout menu parent element
re: https://github.com/patternfly/patternfly/pull/4301#discussion_r691216222

We're using a button for the flyout menu parent element, and looks like we should be using an anchor - https://www.w3.org/WAI/tutorials/menus/flyout/#use-parent-as-toggle

mattnolting:
closes #4306 

Based on the direction provided by [W3.org flyout guidance](https://www.w3.org/WAI/tutorials/menus/flyout/), a flyout parent should be an anchor. If the flyout control should be a link, the icon should control toggle and text should be clickable.

### DIFF
--- a/src/patternfly/components/Menu/examples/Menu.md
+++ b/src/patternfly/components/Menu/examples/Menu.md
@@ -119,6 +119,150 @@ import './Menu.css'
 {{/menu}}
 ```
 
+### With flyout with link
+```hbs
+{{#> menu menu--modifier="pf-m-flyout"}}
+  {{#> menu-content}}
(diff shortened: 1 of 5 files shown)


mcoker:
@mattnolting talking to @jessiehuff about this and she said

> I think I'd like to talk to Matt about when he gets back. I'm curious of his reasoning since I don't see a massive difference on this with using anchor vs. button. Technically the role `menuitem` we put on them will override it anyway so I'm not quite sure of the benefit.

PR #4401, opened 2021-09-23
chore(build): add noFormat

redallen:
Just add `noFormat` to your example code block.

``````
```hbs noFormat
a <span>b</span> c
```
``````


mcoker:
Does this disable formatting of the entire example? Ideally we would be able to disable the formatting on a single block or file or something like that.

redallen:
Yes, it's for the entire example. I think you can use `<!--prettyhtml-ignore-->` for single nodes, although I haven't tested it.

mcoker:
Closing in favor of using `<!--prettyhtml-ignore-->` in one off cases since disabling formatting is generally a bad practice.

PR #4402, opened 2021-09-24
fix(icons): copied unicodes from PF3 icons, make new icon unicodes persist

### LINKED ISSUE #3626: Update pficon font unicode to match PF3 font
To support designers using the [unicode from the PF3 icon font](https://www.patternfly.org/v3/styles/icons/), 
let's update the unicode in the PF4 icon font to match the PF3 font, so that they don't have to update the unicode for each character when upgrading from to the PF4 font.

evwilkin:
Closes #3626 

This PR:
- Pulls forward PF3 unicodes to PF4 icons
  - Aligns the [PF4 PatternFly icons](https://pf4.patternfly.org/icons) with their [unicode assignments from PF3](https://www.patternfly.org/v3/styles/icons/) (for the icons that match across PF versions), by adding a new `existingIconCodes.json` file to track these pairings.
  - Adds the `fixedCodepoints` option to `iconfontCss` which checks the new `existingIconCodes.json` file when assigning unicodes to each icon and uses the assigned unicode from that file rather than creating a new one for existing icons.
    - Note: this feature doesn't seem to be mentioned in their [docs](https://github.com/backflip/gulp-iconfont-css#api) but was [added here](https://github.com/backflip/gulp-iconfont-css/pull/30/files).

- Enforces that unicodes for both new & existing icons persist once they are assigned
  - Adds the `firstGlyph` option to `iconfontCss` which indicates the starting unicode to assign for any new icons.
    - This is calculated by the `nextCodePoint` and `nextUnicodeHex` variables which simply find the largest existing icon unicode and increment it by 1.
    - If more than 1 new icon is added, the plugin automatically continues to increment by 1 for the next unicode assignment.
  - Writes the full list of icon unicode assignments & corresponding codepoints back to the new `existingIconCodes.json` file, thereby persisting any new icons added.
    - This triggers off of the `glyphs` event emitted by the `gulp-iconfont` plugin.


scripts/gulp/icons.js
```diff
@@ -14,10 +19,14 @@ function pfIconFont() {
     .pipe(
       iconfontCss({
         fontName: pficonFontName,
-        path: 'scss',
+        path: path.join(process.cwd(), '/scripts/gulp/icons_template.scss'),
```
redallen, line 22: Use `path.join(__dirname, 'icons_template.scss'` so the cwd can be anywhere.

scripts/gulp/icons.js
```diff
@@ -27,6 +36,23 @@ function pfIconFont() {
...
+      const newIconsObj = glyphs.reduce((iconsObj, glyph) => {
+        const { name, unicode } = glyph;
+        const codePoint = unicode[0].codePointAt(0);
+        iconsObj[name] = codePoint.toString(16).toUpperCase();
+        iconsObj.codePoints.push(codePoint);
+        return iconsObj;
+      }, { codePoints: [] });
+
+      fs.writeFile(
+        path.join(process.cwd(), '/src/icons/existingIconCodes.json'),
```
redallen, line 51: Use `path.join(__dirname, '../../src/icons/existingIconCodes.json'` so the cwd can be anywhere.

scripts/gulp/icons.js
```diff
@@ -27,6 +36,23 @@ function pfIconFont() {
...
+        JSON.stringify(newIconsObj),
+        err => err && console.log(err)
```
redallen, line 53: Remove this line. If there's an error it will already be printed AND it should crash the program as well.

src/icons/existingIconCodes.json
```diff
@@ -0,0 +1,250 @@
+{
+  "codePoints": [
```
redallen, line 2: Can't this simply be `"lastCodePoint": 59665`?

src/patternfly/assets/pficon/pficon.scss
```diff
@@ -21,376 +25,376 @@
 }
 
 @function pf-icon-char($filename) {
-	$char: "";
+	$char: '';
 
 	@if $filename == add-circle-o {
-		$char: "\E001";
+		$char: '\E61B';
```
redallen, line 31: I'm not sure if anyone is relying on the existing chars. I'm comfortable making this change as long as you'll be able to fix any potential problems consumers have in the weeks after this is merged.

scripts/gulp/icons.js
```diff
@@ -2,8 +2,13 @@ const { src, dest } = require('gulp');
 const iconfont = require('gulp-iconfont');
 const iconfontCss = require('gulp-iconfont-css');
 const generateIcons = require('../../src/icons/generateIcons.js');
+const existingIconCodes = require('../../src/icons/existingIconCodes.json');
+const fs = require('fs');
+const path = require('path');
 
 const pficonFontName = 'pficon';
+const nextCodePoint = Math.max(...existingIconCodes.codePoints) + 1;
```
redallen, line 10: Can't this just be `existingIconCodes.lastCodePoint + 1`?

redallen (changes requested):
I'm not a big fan of having two sources of truth for code points in `existingIconCodes.json` and `pficon.scss`. Is there a way to just use one or the other?

scripts/gulp/icons.js
evwilkin: FYI it looks like the program fails to run if I remove this line completely (node complains that a callback is expected here), but I'll remove the console.log.

redallen: Ah, just use `fs.writeFileSync` instead!

evwilkin:
@redallen I've removed the 2 sources of truth so that `pficons.scss` is the sole source of truth now.  Existing icon names & corresponding values are parsed from that file before the `gulp-iconfont` and `gulp-iconfont-css` plugins run, and then that same `pficons.scss` file is updated with any changes (aka new icons) when the font is generated.

Note:
- I did try using `sass.renderSync` to parse `pficons.scss` but it was breaking as soon as it hit the undefined variable `#{$pf-global--fonticon-path}`.
- Needing a way to cut that variable out of the parsing (that @font-face rule doesn't contain any relevant icon unicode info) I fell back to `readFileSync` and otherwise mostly the same string manipulation to map the icon names & unicodes from the SASS style rules.

redallen (approved):
LGTM! Thanks for the single source of truth :)

bdellasc:
Thanks for the follow-up @evwilkin! Looks great

evwilkin:
@redallen I've updated this PR to write both Font Awesome && pf-icon names/unicodes to JSON files within `dist` in order to use them when adding new "Unicode" column to the docs in https://github.com/patternfly/patternfly-org/pull/2736

I reorganized the code to add a few shared functions to handle the parsing/file writing at the top, then tackled the Font Awesome icons, then most of the pf-icon gulp code at the bottom is the same.

Issue #4403, opened 2021-09-29
Move drag drop into component


mcoker:
> could you make it a component? that way react-styles can pick it up and we can integrate with JS classnames instead of hardcoded strings
> it's a component in React

PR #4404, opened 2021-09-29
chore(drag-drop): moved drag drop styles into component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4403


redallen (approved):
Thanks!

Issue #4374, opened 2021-09-09
Support key-value filtering


mcarrano:
some sort of input with menu that can be used to support key-value filter use cases for selecting multiple search terms. See the proposed design here: https://marvelapp.com/prototype/15dbe4c9/screen/81845426 

@nicolethoen @kmcfaul @tlabaj please review and let me if you see any issues and how best to proceed.

@jessiehuff would also love your input from a a11y perspective.

nicolethoen:
This likely should be a composable input - to be paired with a menu to build an attr-value advanced search. The core implementation does not need to include the menu. It could build upon a preexisting input component? like form control?

PR #4408, opened 2021-09-30
feat(text-input-group): added component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4374

### DIFF
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -23,7 +23,7 @@ cssPrefix: pf-c-text-input-group
 ```hbs
 {{#> text-input-group text-input-group--modifier="pf-m-icon"}}
   {{#> text-input-group-row}}
-      {{#> text-input-group-main}}
+    {{#> text-input-group-main}}
       {{#> text-input-group-text}}
         {{> text-input-group-icon}}
         {{> text-input-group-text-input}}
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -1,0 +1,77 @@
+---
+id: 'Text input group'
+beta: true
+section: components
+cssPrefix: pf-c-text-input-group
+---
+
+## Examples
+### Basic
+```hbs
+{{#> text-input-group text-input-group--id="basic"}}
+  {{#> text-input-group-row}}
+    {{#> text-input-group-main}}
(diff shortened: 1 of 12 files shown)


mcarrano (commented):
This looks good @mcoker . I just have a question about what the responsive behavior should be. Looks like right now it just clips when I shrink the screen. Expect some type of truncation. Will React need to handle that?

mcoker:
@mcarrano in order for the text to truncate with CSS, text needs to overflow the container. In this case, the chip group is overflowing the container, so the text in the chips doesn't truncate. We could try some other sort of effect.

mcarrano:
So I guess that the Chip group needs to handle the overflow then?

nicolethoen (approved):
I think the basic html structure makes sense and should be fairly straight forward to bring into react. Responsive behaviors are a good question. But if we can put breakpoint mods on various sub components inside a text-input-group, then it may work out no problem

src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
```diff
@@ -0,0 +1,77 @@
...
+    {{/text-input-group-main}}
+  {{/text-input-group-row}}
+{{/text-input-group}}
+```
+
+### Utilities and icon
+```hbs
+{{#> text-input-group text-input-group--modifier="pf-m-icon"}}
+  {{#> text-input-group-row}}
+      {{#> text-input-group-main}}
```
mattnolting, line 26:
```suggestion
    {{#> text-input-group-main}}
```

src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
```diff
@@ -0,0 +1,77 @@
...
+      {{#> text-input-group-text}}
+        {{> text-input-group-icon}}
```
mattnolting, line 28: Looks like this partial is missing.

mattnolting (commented):
This looks great. I do wonder if you need `.pf-c-text-input-group__row` or those styles can be applied directly to `.pf-c-text-input-group`.

Looks like your

mcoker:
> I do wonder if you need .pf-c-text-input-group__row or those styles can be applied directly to .pf-c-text-input-group.

@mattnolting it's modeled off of the existing search input, and the row class is there to separate the search input stuff from the menu that popped out of the bottom of the component. I don't think it's necessary here, though I left it since I imagine that container could be useful or allow it to be styled more flexibly. Though since it's beta, we can just add it later if we end up needing it.

Updated that and fixed the missing partial.

mattnolting (approved):
LPTM 🚀  Ship it!!

Issue #4412, opened 2021-10-01
Update use of date element


mcoker:
`<date>` should be `<time>`

```
src/patternfly//demos/Tabs/examples/Tabs.md:                <date>Oct 15, 1:51 pm</date>
src/patternfly//demos/Tabs/examples/Tabs.md:                      <date>Oct 15, 1:51 pm</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 11:02 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 11:13 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 10:59 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 10:45 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 11:02 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 11:13 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 10:59 am</date>
src/patternfly//demos/Card/examples/Card.md:            <date class="pf-u-color-200 pf-u-font-size-sm">Jun 17, 10:45 am</date>
```

mcoker:
fixed via https://github.com/patternfly/patternfly/pull/4423/files

PR #4423, opened 2021-10-06
fix(global): Updated use of date element (#4412)

bonjo7:
Updated use of date element from `date` to `time` ->  https://github.com/patternfly/patternfly/issues/4412


mcoker (approved):
LGTM! Thanks @bonjo7!

PR #4425, opened 2021-10-07
feat(card): added hoverable/selectable/selected-raised

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4358

**Note:** triggering "selected" state on `:active` (mouse down) for demo purposes only. Will update that prior to merging.


mcarrano:
@mcoker @srambach I'm looking at the original issue (https://github.com/patternfly/patternfly/issues/4358) and realizing that this change is intended to apply to both Cards and Tiles. Should we have a separate PR for Tiles or add it here?

srambach:
> @srambach @mcarrano I've updated the flat card so it looks like this now

I'm fine with it as long as design is, just wanted to check all the options.

evwilkin:
@mceledonia 👆 WDYT?

PR #4448, opened 2021-10-15
feat(table): added nested headers

### LINKED ISSUE #4365: Table: Add nested/grouped column headers
Add the option to have grouped columns in a table. Designs here: https://marvelapp.com/prototype/5d421g0

mattnolting:
closes #4365


mcoker:
@jessiehuff would you have time to look at this? One thing I noticed is the tabindex with the buttons in the nested header examples seems off, though I'm not sure what we can do about given the markup.

mcoker:
@jessiehuff tabbing doesn't start in the first column like you might(?) expect since all of the top level headers are in a single row, and the nested headers are in a separate row.
Oct-20-2021 13-55-41

mattnolting:
@mcoker updated

PR #4451, opened 2021-10-18
fix(toolbar): Do not hide labels in collapsed state

### LINKED ISSUE #4284: Toolbar hides labels in collapsed state
Using example: https://www.patternfly.org/v4/components/toolbar/react/stacked-example/
Changing browser size so toolbar is collapsed it looks like this:
Screenshot from 2021-08-11 14-40-10

Labels for all selects are not displayed. This seems odd, as for someone who only sees this collapsed state (like mobile users) would have hard time understanding what these selects do.

Removing [THIS](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Toolbar/toolbar.scss#L338) rule makes it look like this, which seems much more user friendly:
Screenshot from 2021-08-11 14-40-30

I would even send PR with removal of this one rule, but I feel like I am not understanding why it was introduced in the first place.

marusak:
Fixes #4284

See #4284 for before and after.

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -99,6 +99,10 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   --pf-c-toolbar__item--m-label--spacer: var(--pf-c-toolbar__item--spacer);
   --pf-c-toolbar__item--m-label--FontWeight: var(--pf-global--FontWeight--bold);
 
+  // Label in expanded content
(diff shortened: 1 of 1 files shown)


mcoker:
Hi @marusak just wanted to check in and see if my suggestion made sense? Let me know if you need any additional information.

marusak:
Sorry for taking too long to update this PR. Your suggestions work like a charm and it indeed looks much better! Thank you :)

mcoker (approved):
LGTM! @mcarrano would you mind re-reviewing?

mcarrano (approved):
Looks great @mcoker @marusak !

Issue #4435, opened 2021-10-11
Request for a nowrap log viewer text style


DaoDaoNoCode:
Currently, Log Viewer at React side is planning to support not-wrapping lines of the log viewer, thus, we need a classname to support `nowrap` style on the text of the log viewer.

PR #4455, opened 2021-10-19
feat(log-viewer): added nowrap variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4435

### DIFF
--- a/src/patternfly/components/LogViewer/log-viewer.scss
+++ b/src/patternfly/components/LogViewer/log-viewer.scss
@@ -13,7 +13,6 @@
 
   // Scroll container
   --pf-c-log-viewer__scroll-container--Height: #{pf-size-prem(600px)};
-  --pf-c-log-viewer__scroll-container--Width: 100%;
   --pf-c-log-viewer__scroll-container--PaddingTop: var(--pf-global--spacer--sm);
   --pf-c-log-viewer__scroll-container--PaddingBottom: var(--pf-global--spacer--sm);
 
@@ -24,14 +23,8 @@
   --pf-c-log-viewer--m-line-numbers__main--before--BackgroundColor: var(--pf-global--BorderColor--100);
 
   // List
-  --pf-c-log-viewer__list--Width: auto;
   --pf-c-log-viewer__list--Height: auto;
-
-  // List item
-  --pf-c-log-viewer__list-item--Right: 0;
-  --pf-c-log-viewer__list-item--Left: 0;
-  --pf-c-log-viewer__list-item--Width: auto;
-  --pf-c-log-viewer--m-nowrap__list-item--Left: var(--pf-c-log-viewer__index--Width);
+  --pf-c-log-viewer--m-line-numbers__list--Left: var(--pf-c-log-viewer__index--Width);
 
   // Index
   --pf-c-log-viewer__index--Display: none;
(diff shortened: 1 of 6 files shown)


DaoDaoNoCode:
@mcoker Thanks for the quick fix! Could you hold this PR for a while? Currently, the line numbers will be scrolled along with the text when we don't wrap lines, this is because of the design of the log viewer, I need to discuss more with the OpenShift team about the design. Thanks!

src/patternfly/components/LogViewer/log-viewer.scss
```diff
@@ -21,11 +24,14 @@
   --pf-c-log-viewer--m-line-numbers__main--before--BackgroundColor: var(--pf-global--BorderColor--100);
 
   // List
-  // --pf-c-log-viewer__list--PaddingTop: var(--pf-global--spacer--sm);
-  // --pf-c-log-viewer__list--PaddingBottom: var(--pf-global--spacer--sm);
+  --pf-c-log-viewer__list--Width: 100%;
```
DaoDaoNoCode, line 27: Is it better to set the width to `unset` or something similar here? If I set to 100%, there are still some problems when scrolling to the right.

Screen Shot 2021-11-08 at 1 10 27 PM

src/patternfly/components/LogViewer/log-viewer.scss
```diff
@@ -85,6 +96,24 @@
...
+
+    .pf-c-log-viewer__list {
+      position: absolute;
+      right: var(--pf-c-log-viewer__list-item--Right);
+      left: var(--pf-c-log-viewer--m-nowrap__list-item--Left);
+      width: var(--pf-c-log-viewer__list-item--Width);
+      overflow: auto;
+    }
+
+    .pf-c-log-viewer__index {
```
DaoDaoNoCode, line 110: Is it OK to apply these styles to the whole component, but not only the `pf-m-nowrap`? The index should always be fixed and set left as `0` no matter the text is wrapped or not.

src/patternfly/components/LogViewer/log-viewer.scss
```diff
@@ -85,6 +96,24 @@
     word-break: break-all;
   }
 
+  &.pf-m-nowrap {
+    --pf-c-log-viewer__text--WhiteSpace: var(--pf-c-log-viewer--m-nowrap__text--WhiteSpace);
+
+    .pf-c-log-viewer__list {
```
DaoDaoNoCode, line 102: Also, the same as `.pf-c-log-viewer__index`, could you also please apply these styles directly to the `.pf-c-log-viewer__list`? The style when wrapping lines is incorrect currently, but after taking this class and `.pf-c-log-viewer__index` out of the `.pf-m-nowrap`, the problem could be solved.

Screen Shot 2021-11-08 at 1 18 19 PM

src/patternfly/components/LogViewer/log-viewer.scss
```diff
@@ -21,11 +24,14 @@
...
+  --pf-c-log-viewer__list--Width: 100%;
+  --pf-c-log-viewer__list--Height: auto;
 
-  --pf-c-log-viewer__scroll-container--PaddingTop: var(--pf-global--spacer--md);
-  --pf-c-log-viewer__scroll-container--PaddingBottom: var(--pf-global--spacer--md);
+  // List item
+  --pf-c-log-viewer__list-item--Right: 0;
+  --pf-c-log-viewer__list-item--Left: 0;
+  --pf-c-log-viewer__list-item--Width: 100%;
+  --pf-c-log-viewer--m-nowrap__list-item--Left: var(--pf-c-log-viewer__index--Width);
```
DaoDaoNoCode, line 34: If I understand correctly, this should be applied to the list but not the list-item, right? Also, this one should be the same no matter when the line is wrapped or not. It could be different when there is no line numbers, so my suggestion is that this could be set to `0` in `.pf-c-log-viewer__list` and changed to `var(--pf-c-log-viewer__index--Width)` in `.pf-m-line-numbers` modifier.

DaoDaoNoCode (changes requested):
This looks great! However, there are still some styles problems in the preview, I put up some suggestions, hope these could make all the styles work perfectly, thank you! @mcoker

mcoker:
@DaoDaoNoCode sorry, looks like I forgot to push some updates. Would you mind checking again?

DaoDaoNoCode (approved):
LPTM, thanks a lot! @mcoker

Issue #4378, opened 2021-09-15
New: Composable collapsable panel


nicolethoen:
We need a composable component that is designed to be a custom content panel which overlays other content.

For example, the advanced search form found under `SearchInput` `React-Demos` on the website.
The form is placed in something that looks like a `Menu` but a menu is designed to have `MenuItems` as children. We need a more flexible, general use panel than a `Menu`

PR #4456, opened 2021-10-19
feat(panel): added panel component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4378
follow up from https://github.com/patternfly/patternfly/pull/4440

### DIFF
--- a/src/patternfly/components/Panel/examples/Panel.md
+++ b/src/patternfly/components/Panel/examples/Panel.md
@@ -82,6 +82,17 @@ cssPrefix: pf-c-panel
 {{/panel}}
 ```
 
+### Bordered
+```hbs
+{{#> panel panel--modifier="pf-m-bordered"}}
+  {{#> panel-main}}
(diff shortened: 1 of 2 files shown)


mcoker:
@mcarrano @srambach I believe @mceledonia was involved in that decision? It's in the other menus like that - select, context selector, menu, etc. Typically we use a shadow to communicate that something is layered on top of some other thing, including elements we make sticky.

Personally I like the border better since, when something is sticky, for example, we don't know if it's actually overlaying content ("stuck") or just a normal element on the page (depending on the user's scroll position), so the element will have a shadow when it's sticky (but not "stuck") and when it's "stuck". A border would work in both scenarios, you just won't get the indication of overlay/depth that the shadow provides.

You can end up in the same situation with the panel/menu/select/etc when the component is set to "scrollable" (and the footer has a shadow) but there isn't a scrollbar since the scrollbar only appears once the scrollable component reaches a specified max height.

Issue #4422, opened 2021-10-06
Reduce spacing for popover and tooltip


mcarrano:
~~Want to reduce outer padding around content for Popover and Tooltip to 16px.~~

Talked to @mceledonia and he clarified that we need to update the tooltip to have 8px padding all around. The popover is OK as it is.

He would also like to decrease the size of the tooltip's arrow by 50%.

See https://github.com/patternfly/patternfly-design/issues/1081 for discussion.

nicolethoen:
closed by https://github.com/patternfly/patternfly/pull/4462

PR #4462, opened 2021-10-21
fix(tooltip): updated spacing, arrow

bonjo7:
Updated content padding to 16px using variable `--pf-global--spacer--lg` #4422


mcoker:
Hey @bonjo7 thanks for the PR! After reading the design issue, I followed up with the designer and the description in https://github.com/patternfly/patternfly/issues/4422 is inaccurate. I updated the issue's description but basically we just want to

* reduce the tooltip's padding to 8px all around (`var(--pf-global--spacer--sm)`)
* leave the popover's padding as it is already (undo the changes in this PR)
* reduce the tooltip arrow size by 50%

Would you mind making those changes in this PR?

PR #4468, opened 2021-10-21
fix(icons): moved unicodes json write to src so it's bundled with build

### LINKED ISSUE #4466: Unicodes json not included in npm package
The work done in https://github.com/patternfly/patternfly/pull/4402 towards https://github.com/patternfly/patternfly/issues/3626 correctly publishes `fa-unicodes.json` but does not publish `pf-unicodes.json` because it's only written when `pfIconFont` gulp plugin is run manually, which is not part of the build process.

These json files or a combined file should be written to `src` and committed, so they are packaged and shipped for use in Org.  The functionality to update/add new icons can stay within `pfIconFont` as it will overwrite this updated file when needed.

evwilkin:
Closes #4466 

This PR combines `fa-unicodes.json` and `pf-unicodes.json` into one `iconUnicodes.json` file and writes it to the `src` directory so it is bundled in the build process.

Previously `fa-unicodes.json` was getting written to `dist` as desired because it was executed at top of file, but `pf-unicodes` was not as it is contained within a function that is not called at build time.


evwilkin:
@mcoker it does add a new file (unicodes.json) which needs to be committed, so I believe we do need a new release.

Issue #4021, opened 2021-04-21
Upgrade Red Hat Font


garrett:
There has been a new release of the Red Hat Font. It includes a bunch of useful things for everyone using PatternFly:

1. Monospace version
2. Tabular figures (fixed width numbers in the proportional version; great for UIs)
3. Variable fonts
4. Light weights
5. Lots of little fixes

Changelog: https://github.com/RedHatOfficial/RedHatFont/blob/master/CHANGELOG.md
GitHub repo: https://github.com/RedHatOfficial/RedHatFont
Demo website: https://redhatofficial.github.io/RedHatFont/

mcoker:
cc @mcarrano

mcarrano:
We will put this on the roadmap for further investigation. In general, I agree this is something we should do.

PR #4476, opened 2021-10-27
feat(fonts): updated red hat font

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4021

* Adds all of the current [RedHatFont](https://github.com/RedHatOfficial/RedHatFont/tree/master/webfonts) files to the repo
* Creates opt-in fonts for:
  * `.pf-m-redhat-updated-font` - updates the base text/display font families we use to the latest. We'll have some users test this font and see if they run into any issues, then tentatively update the default font in PF to these fonts
  * `.pf-m-redhatmono-font` - updates our mono stack to use `RedHatMono`
  * `.pf-m-redhatVF-font` - updates the text/display/mono fonts to use the RH variable font version of these fonts
  * `.pf-m-tabular-nums` - enables `font-variant-numeric: tabular-nums;` on any element. Add it to `body` for a global opt-in.

**Note:** The `woff` format has been dropped from https://github.com/RedHatOfficial/RedHatFont/tree/master/webfonts. If opting in to any of these font updates, the `woff` format has been removed. This should only impact IE users, which we don't support.


garrett:
Hooray! Thanks!

I'm checking to see if there's a way I can test this PR in Cockpit itself.

Also: How does one use the tabular numbers with PatternFly? Is it enabled by default (as it's PF provides UI anyway) or parts of PF (like lists, tables, progress bars, and other widgets), or does it need opt-in also? Or does it need custom CSS for this still?

(Tabular numbers is a feature for using monospaced numbers in sans serif font so the UI doesn't shift when the numbers change. This is especially useful for making sure numbers line up properly and progress percentages don't bounce around on every update. It's common in UI fonts and is a new feature for the RedHatText font.)

mattnolting (approved):
LPTM 👍

dlabrecq (approved):

mcoker:
@mcarrano @mceledonia wdyt about the tabular numbers with this new font? I imagine this will be a really popular feature. Mark Pedrotti is also looking for this enhancement.

As @garrett says, numbers in a block will use the monospace font so they're all the same width.

Current:
Screen Shot 2021-11-22 at 2 23 07 PM

Using tabular-nums
Screen Shot 2021-11-22 at 2 23 47 PM

Should that be the default, either for all numbers, or in specific components? It will cause the numbers to take up more space, so it could shift layouts, which could be seen as breaking. We can also add this to the text utility class so it could be declared on a block with a patternfly class.

mceledonia:
Yeah I would be hesitant to force this considering it adds width and could break component layouts. I do like the idea of it being a utility class and allowing folks to choose when/where this is used. 

I'll have to look more into tabular vs standard numbers soon, to be honest I'm not too well versed in when to use one versus the other.

garrett:
You should use tabular numbers when:

- Numbers should line up
  - especially with decimal points, when aligned to the right, and displayed vertically; this is often seen in tables
- When numbers may change a lot
  - especially percentages in progress bars, as the number can quickly change, which causes shifts with proportional numbers
- With time
  - when minutes and especially seconds can cause a shift
  - when multiple times are used next to each other, especially vertically

So tabular numbers would benefit tables, progress bars, and time-related widgets the most.

mceledonia:
That all makes sense to me, thanks for the quick info @garrett. Also just got a chance to scan your linked article. Tabular figures are pretty cool!

We should definitely use them for those components, in terms of how we implement them are there any preferences? Going with opt-in will eliminate breaking things but could lead to inconsistency across products, though I'm not sure how disruptive the inconsistency will be with something like this. If it was disruptive/noticeable it might end up prompting a change from the product teams anyways.

mcoker:
> Going with opt-in will eliminate breaking things but could lead to inconsistency across products, though I'm not sure how disruptive the inconsistency will be with something like this.

@mceledonia I don't think it's a significant difference, most people probably won't notice. Maybe we could have some products test the opt-in, and if all goes well, we could consider making it the default. Last I spoke with @mcarrano, that's how we're planning to handle this font update as a whole - test the opt-in in products, then update the default font if all goes well.

mcarrano:
I agree that this really helps the scanability of tabular data, but I wouldn't want to make this the default for fear of unintended side effects. Let's stick with offering this as an opt-in for now, but promote the new tabular font-face as the preferred choice to use when displaying tabular data.

@doruskova @mmenestr please take note of this as I think it's something we'll want to expose and recommend to designers through the Sketch kit and design documentation.

mmenestr:
Create this design guideline issue https://app.zenhub.com/workspaces/design-doc-workspace-5f08c64d6a6ef2001d56a458/issues/patternfly/patternfly-org/2760

mattnolting (approved):
LPTM 🔥  +1 to tabular numbers global opt-in. I doubt it would change any application's layout unless to improve them.

mcoker:
@mattnolting added `.pf-m-tabular-nums` as a common class.

mattnolting:
@mcoker 🔥🔥🔥

PR #4484, opened 2021-11-01
feat(TextInputGroup): added disabled styling

### LINKED ISSUE #4482: Add disabled styling to text-input-group
Per patternfly/patternfly-react#6524 text input group requires disabled styling similar to that of [Text Input's disabled styling](https://www.patternfly.org/v4/components/text-input/#disabled).

wise-king-sullyman:
Closes #4482

### DIFF
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -19,10 +19,10 @@ cssPrefix: pf-c-text-input-group
 
 ### Disabled
 ```hbs
-{{#> text-input-group text-input-group--id="disabled" text-input-group--modifier='pf-m-disabled'}}
+{{#> text-input-group text-input-group--id="disabled" text-input-group--IsDisabled="true"}}
   {{#> text-input-group-main}}
     {{#> text-input-group-text}}
-      {{> text-input-group-text-input aria-label="disabled text input group example" text-input-group-text-input--attribute='disabled' text-input-group--value='Disabled'}}
+      {{> text-input-group-text-input text-input-group--value='Disabled'}}
     {{/text-input-group-text}}
   {{/text-input-group-main}}
 {{/text-input-group}}
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -19,10 +19,10 @@ cssPrefix: pf-c-text-input-group
 
 ### Disabled
 ```hbs
-{{#> text-input-group text-input-group--id="disabled" text-input-group--IsDisabled="true"}}
+{{#> text-input-group text-input-group--id="text-input-group-disabled" text-input-group--IsDisabled="true"}}
   {{#> text-input-group-main}}
     {{#> text-input-group-text}}
-      {{> text-input-group-text-input text-input-group--value='Disabled'}}
+      {{> text-input-group-text-input}}
     {{/text-input-group-text}}
   {{/text-input-group-main}}
 {{/text-input-group}}
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -17,6 +17,17 @@ cssPrefix: pf-c-text-input-group
 {{/text-input-group}}
 ```
 
+### Disabled
+```hbs
+{{#> text-input-group text-input-group--id="disabled" text-input-group--modifier='pf-m-disabled'}}
+  {{#> text-input-group-main}}
+    {{#> text-input-group-text}}
+      {{> text-input-group-text-input aria-label="disabled text input group example" text-input-group-text-input--attribute='disabled' text-input-group--value='Disabled'}}
+    {{/text-input-group-text}}
+  {{/text-input-group-main}}
+{{/text-input-group}}
+```
+
 ### Utilities and icon
 ```hbs
 {{#> text-input-group}}
--- a/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
@@ -2,6 +2,9 @@
   type="text"
   value="{{text-input-group--value}}"
   aria-label="Type to filter"
+  {{#if text-input-group--IsDisabled}}
+    disabled
+  {{/if}}
   {{#if text-input-group-text-input--attribute}}
     {{text-input-group-text-input--attribute}}
   {{/if}}>
--- a/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
@@ -3,10 +3,7 @@
   value="{{#if text-input-group--IsDisabled}}Disabled{{else}}{{text-input-group--value}}{{/if}}"
   aria-label="Type to filter"
   {{#if text-input-group--IsDisabled}}
-    disabled
-  {{/if}}
-  {{#if text-input-group--IsDisabled}}
-    aria-label="Disabled text input group example input"
+    disabled aria-label="Disabled text input group example input"
   {{/if}}
   {{#if text-input-group-text-input--attribute}}
     {{text-input-group-text-input--attribute}}
--- a/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
@@ -1,9 +1,12 @@
 <input class="pf-c-text-input-group__text-input{{~#if text-input-group-text-input--modifier}} {{text-input-group-text-input--modifier}}{{/if}}"
   type="text"
-  value="{{text-input-group--value}}"
+  value="{{#if text-input-group--IsDisabled}}Disabled{{else}}{{text-input-group--value}}{{/if}}"
   aria-label="Type to filter"
   {{#if text-input-group--IsDisabled}}
     disabled
+  {{/if}}
+  {{#if text-input-group--IsDisabled}}
+    aria-label="Disabled text input group example input"
   {{/if}}
   {{#if text-input-group-text-input--attribute}}
     {{text-input-group-text-input--attribute}}
--- a/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
@@ -3,7 +3,8 @@
   value="{{#if text-input-group--IsDisabled}}Disabled{{else}}{{text-input-group--value}}{{/if}}"
   aria-label="Type to filter"
   {{#if text-input-group--IsDisabled}}
-    disabled aria-label="Disabled text input group example input"
+    disabled
+    aria-label="Disabled text input group example input"
   {{/if}}
   {{#if text-input-group-text-input--attribute}}
     {{text-input-group-text-input--attribute}}
--- a/src/patternfly/components/TextInputGroup/text-input-group.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group.hbs
@@ -1,16 +1,12 @@
 <div class="pf-c-text-input-group
-  {{#if text-input-group--modifier}}
-    {{text-input-group--modifier}}
-  {{/if}}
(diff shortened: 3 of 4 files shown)


src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -35,12 +38,28 @@
...
   --pf-c-text-input-group__utilities--c-button--PaddingLeft: var(--pf-global--spacer--xs);
 
+  // input disabled style
+  --pf-c-text-input-group--m-disabled--Color: var(--pf-global--disabled-color--100);
+  --pf-c-text-input-group--m-disabled--BackgroundColor: var(--pf-global--disabled-color--300);
+  --pf-c-text-input-group--m-disabled--BorderColor: transparent;
+
   position: relative;
   display: flex;
+  color: var(--pf-c-text-input-group--Color);
```
mcoker, line 48: Can you give this property a default value? You could also set `--pf-c-text-input-group--Color`, but you would have to set to `--pf-global--Color--100` or something instead of `inherit` since `inherit` is a special word and would apply to the custom property instead of being passed as the value to some other property when the custom property is used.

```suggestion
  color: var(--pf-c-text-input-group--Color, inherit);
```

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -35,12 +38,28 @@
...
   &:hover {
     --pf-c-text-input-group__text--after--BorderBottomColor: var(--pf-c-text-input-group--hover__text--after--BorderBottomColor);
   }
+
+  &.pf-m-disabled {
+    --pf-c-text-input-group--Color: var(--pf-c-text-input-group--m-disabled--Color);
+    --pf-c-text-input-group__icon--Color: var(--pf-c-text-input-group--m-disabled--Color);
+    --pf-c-text-input-group--BackgroundColor: var(--pf-c-text-input-group--m-disabled--BackgroundColor);
+
+    cursor: not-allowed;
```
mcoker, line 60: typically we will use `pointer-events: none;` here, which will disable any kind of interaction within the component.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -35,12 +38,28 @@
...
+    border-color: var(--pf-c-text-input-group--m-disabled--BorderColor);
```
mcoker, line 61: We try to override vars whenever possible. You'll want to disable both the default grey border and the bottom accent border. You could do that via:


```
--pf-c-text-input-group__text--before--BorderWidth: 0;
--pf-c-text-input-group__text--after--BorderBottomWidth: 0;
```

src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
```diff
@@ -17,6 +17,17 @@ cssPrefix: pf-c-text-input-group
 {{/text-input-group}}
 ```
 
+### Disabled
+```hbs
+{{#> text-input-group text-input-group--id="disabled" text-input-group--modifier='pf-m-disabled'}}
+  {{#> text-input-group-main}}
+    {{#> text-input-group-text}}
+      {{> text-input-group-text-input aria-label="disabled text input group example" text-input-group-text-input--attribute='disabled' text-input-group--value='Disabled'}}
```
mcoker, line 25: * There isn't an `aria-label` arg on that hbs template but there should be.
* You'll also want to add `tabindex="-1"` to disable tabbing to it or anything in it.
* You could also do `text-input-group--IsDisabled` on the parent hbs template, and just have that set all the classes/attrs/labels you will always set if it's disabled. That makes it easier to use going forward.

mcoker (approved):
Looks really good! A few bits of feedback.

src/patternfly/components/TextInputGroup/text-input-group.hbs
```diff
@@ -1,6 +1,16 @@
-<div class="pf-c-text-input-group{{#if text-input-group--modifier}} {{text-input-group--modifier}}{{/if}}"
+<div class="pf-c-text-input-group
+  {{#if text-input-group--modifier}}
```
mcoker, line 2:
```suggestion
  {{~#if text-input-group--modifier}}
```

src/patternfly/components/TextInputGroup/text-input-group.hbs
```diff
@@ -1,6 +1,16 @@
...
+    {{text-input-group--modifier}}
+  {{/if}}
+  {{#if text-input-group--IsDisabled}}
```
mcoker, line 5:
```suggestion
  {{~#if text-input-group--IsDisabled}}
```

mcoker: The `~` will remove any leading white-space to keep extra spaces from showing up in the class list. Also, we typically have the `--modifiers` block last, after any other classes we're adding.

src/patternfly/components/TextInputGroup/text-input-group.hbs
```diff
@@ -1,6 +1,16 @@
...
+    pf-m-disabled
+  {{/if}}"
   {{#if text-input-group--attribute}}
     {{{text-input-group--attribute}}}
+  {{/if}}
+  {{#if text-input-group--IsDisabled}}
+    aria-label="Disabled text input group example"
+    tabindex="-1"
```
mcoker, line 13: I'm actually not sure if this needs `tabindex="-1"`. I tested adding a button to the component and it still allowed me to tab to it, even though the parent component had `tabindex="-1"`. Since this element isn't focusable to begin with, I don't think we need to add this when it's disabled. Sorry for the wrong suggestion :)

src/patternfly/components/TextInputGroup/text-input-group.hbs
```diff
@@ -1,6 +1,16 @@
...
+    {{text-input-group--modifier}}
+  {{/if}}
+  {{#if text-input-group--IsDisabled}}
+    pf-m-disabled
+  {{/if}}"
   {{#if text-input-group--attribute}}
     {{{text-input-group--attribute}}}
+  {{/if}}
+  {{#if text-input-group--IsDisabled}}
+    aria-label="Disabled text input group example"
```
mcoker, line 12: @jessiehuff can you confirm whether we need an `aria-label` on the outer container or just on the text input inside? I'm assuming the text input since this is designed to mimic the existing TextInput component.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -35,12 +38,31 @@
...
     --pf-c-text-input-group__text--after--BorderBottomColor: var(--pf-c-text-input-group--hover__text--after--BorderBottomColor);
   }
+
+  &.pf-m-disabled {
+    --pf-c-text-input-group--Color: var(--pf-c-text-input-group--m-disabled--Color);
+    --pf-c-text-input-group__icon--Color: var(--pf-c-text-input-group--m-disabled--Color);
+    --pf-c-text-input-group--BackgroundColor: var(--pf-c-text-input-group--m-disabled--BackgroundColor);
+
+    pointer-events: none;
+    border-color: var(--pf-c-text-input-group--m-disabled--BorderColor);
```
mcoker, line 61: This element doesn't have a border currently, so it shouldn't need this?

src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
```diff
@@ -2,6 +2,9 @@
   type="text"
   value="{{text-input-group--value}}"
```
mcoker, line 3: You could do `{{#if text-input-group--IsDisabled}}Disabled{{else}}{{text-input-group--value}}{{/if}}`

src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
```diff
@@ -17,6 +17,17 @@ cssPrefix: pf-c-text-input-group
 {{/text-input-group}}
 ```
 
+### Disabled
+```hbs
+{{#> text-input-group text-input-group--id="disabled" text-input-group--IsDisabled="true"}}
```
mcoker, line 22: We will typically add a little context to the ID to avoid any potential conflicts if this is used as an `id` somewhere, since an `id` can only be used once on a page.

```suggestion
{{#> text-input-group text-input-group--id="text-input-group-disabled" text-input-group--IsDisabled="true"}}
```

mcoker (changes requested):
Coming together nicely! Left some feedback

src/patternfly/components/TextInputGroup/text-input-group.hbs
wise-king-sullyman: I'll go ahead and move this label to just the input itself.

mcarrano (commented):
@wise-king-sullyman the styling looks fine. But in comparing with https://www.patternfly.org/v4/components/text-input#disabled I notice that the standard disabled text field also has a cursor change, but the text input group does not.

wise-king-sullyman:
@mcarrano I had originally used `cursor: not-allowed` on this component to copy the disabled experience you're referring to from `text-input`, but @mcoker suggested that I instead use `pointer-events: none` so that any interaction with any items inside of the disabled `text-input-group` would be prevented.

I can change back to using `cursor: not-allowed` to enable the cursor change, but this will exclude me from being able to use `pointer-events: none` and require natively disabling any interactive elements (other than the input) passed into the `text-input-group` when setting it as disabled.

[Button](https://www.patternfly.org/v4/components/button/) and [Dropdown](https://www.patternfly.org/v4/components/dropdown) are a couple of examples where the cursor does not change when disabled. 

Ultimately this comes down to a decision between making things simpler from a developer perspective, or nicer from a user perspective. I'm happy to go in either direction you prefer but thought you may want this additional context.

src/patternfly/components/TextInputGroup/text-input-group.hbs
jessiehuff: I agree, the `aria-label` makes the most sense on the text input like the TextInput component so that when the user gets to the input, it hears the label. That's where the screen reader would expect to interact with it. :)

mcarrano (approved):
@wise-king-sullyman Thanks for the explanation about why we should not change the cursor when the text input group is disabled. I agree to keep it as is.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -35,12 +38,29 @@
...
+
+  &.pf-m-disabled {
+    --pf-c-text-input-group--Color: var(--pf-c-text-input-group--m-disabled--Color);
+    --pf-c-text-input-group__icon--Color: var(--pf-c-text-input-group--m-disabled--Color);
+    --pf-c-text-input-group--BackgroundColor: var(--pf-c-text-input-group--m-disabled--BackgroundColor);
+
+    pointer-events: none;
+
+    --pf-c-text-input-group__text--before--BorderWidth: 0;
+    --pf-c-text-input-group__text--after--BorderBottomWidth: 0;
```
mcoker, line 62: Would you mind moving these vars to the top of this block? We follow the format of

```
.selector {
  // vars
  --var: whatever;

  // properties
  property: whatever;

  // combined selectors
  &.pf-m-whatever[:not(.something)] {
    // stuff
  }

  // combinators, descendent selectors, etc
  .some-descendent {
    // stuff
  }
  ~ .some-sibling {
    // stuff
  }
}
```

mcoker (changes requested):
Looks awesome! Just a couple of nits and it's ready to go.

src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
```diff
@@ -1,7 +1,10 @@
 <input class="pf-c-text-input-group__text-input{{~#if text-input-group-text-input--modifier}} {{text-input-group-text-input--modifier}}{{/if}}"
   type="text"
-  value="{{text-input-group--value}}"
+  value="{{#if text-input-group--IsDisabled}}Disabled{{else}}{{text-input-group--value}}{{/if}}"
   aria-label="Type to filter"
+  {{#if text-input-group--IsDisabled}}
+    disabled aria-label="Disabled text input group example input"
```
mcoker, line 6: One last nit! Can you put these on 2 lines?

```suggestion
    disabled
    aria-label="Disabled text input group example input"
```

mcoker (changes requested):

mcoker (approved):
Fantastic!!! 🏆

PR #4486, opened 2021-11-02
fix(nav): fixed toggle icon rotation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4485

@mattnolting I think ideally we would handle the icon rotation by redefining vars instead of setting the `transform` property like we're doing now, but since we're treating changing properties as breaking changes, this seemed like the easiest fix. Lemme know what you think!


mattnolting (approved):
Nice, LPTM!

PR #4489, opened 2021-11-04
fix(table): fix pf-m-truncate alignment

### LINKED ISSUE #4488: fix(table): update sticky columns to support flex and truncation

mattnolting:
closes #4488 

Using `pf-l-flex` as a child of table cells, in conjunction with `.pf-m-truncate` and icons prohibits the cell from restricting width. This PR

- Removes `pf-l-flex` 
- Add support for vertical alignment of icons in `pf-m-truncate`
- Allows for text to truncate when width is set

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -552,6 +552,14 @@
     --pf-c-table--cell--Overflow: hidden;
     --pf-c-table--cell--TextOverflow: ellipsis;
     --pf-c-table--cell--WhiteSpace: nowrap;
+
+    // stylelint-disable selector-class-pattern
+    // align icon middle when truncated
+    > .pficon,
+    > .fas {
+      vertical-align: middle;
+    }
+    // stylelint-enable
   }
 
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -552,6 +552,14 @@
     --pf-c-table--cell--Overflow: hidden;
     --pf-c-table--cell--TextOverflow: ellipsis;
     --pf-c-table--cell--WhiteSpace: nowrap;
+
+    // stylelint-disable selector-class-pattern
+    // align icon middle when truncated
+    > .pficon,
+    > .fas {
+      vertical-align: middle;
```
mcoker, line 560: Do you think we can make this assumption? And could it move icons someone might be using today in a way they wouldn't expect? Also this won't do anything with react SVGs (which are likely the most common use case) or another icon a user might be using.

Could we remove the icons from our example, and leave sorting this out with custom CSS on the user?

mcoker (approved):
L🎉TM!

PR #4490, opened 2021-11-04
fix(form-select): match padding with select

### LINKED ISSUE #4387: "form select" padding is not consistent with PatternFly "select"
When set to auto width, the dropdown icon is too close to the text:
Screenshot 2021-09-21 at 10-31-29 PatternFly 4 • Form select

It looks like it's set to 24 pixels padding to the right, which _includes_ the icon, padding to the right, and padding to the left:
Screenshot from 2021-09-21 10-31-33

Here's the non-form PatternFly select, also set to auto width:
Screenshot 2021-09-21 at 10-31-07 PatternFly 4 • Select

Meanwhile, the same icon in the PF select is 24px + 8px to the right + 16px to the left. That's 48px total, twice of what the form select.
Screenshot from 2021-09-21 10-31-12

---

Here's a comparison, with PF select on top and the PF's styled form select (multiple times, under the first one) aligned to different locations:

Screenshot from 2021-09-21 11-13-21

Here's a difference with the form select being at 25%:

select-comparison

We can see that vertical padding is fine, text is fine, and even the dropmarker to the right side is fine.

The problems:

- calculation of the padding on the right to change the width between the text and dropdown icon (a little too much)
- padding on the left between the border and the text (noticeably not enough)

---

For context: We encountered this issue in Cockpit-Podman @ https://github.com/cockpit-project/cockpit-podman/issues/755. There are additional screenshots on that issue, but hopefully the ones here communicate the issue enough.

srambach:
This adjusts the padding on the form component select so that it matches the PatternFly select component.
Screen shot of Chrome, FF, and Safari below. @garrett I didn't find that the same values you provided gave me the correct alignment (but the selector was very helpful), so if you could check this out that would be great. 

To see this, you'll need to inspect and put width:auto on the select form elements.

image

One other note: the placeholder text doesn't seem to respect the padding when using width: auto. If you put `&nbsp;` in the placeholder you can fake it... 😬

fixes #4387

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -148,7 +148,8 @@ $pf-c-form-control--m-clock--Coordinates: "M256 8C119 8 8 119 8 256s111 248 248 
   --pf-c-form-control--m-clock--BackgroundUrl: #{pf-bg-svg($pf-c-form-control--m-clock--Coordinates, $svg-color: $pf-c-form-control--m-clock--Color)};
 
   // Select -- rename __select to --select in a breaking change release
-  --pf-c-form-control__select--PaddingRight: var(--pf-global--spacer--lg);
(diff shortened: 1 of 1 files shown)


mcoker (approved):
LGTM!!!

PR #4494, opened 2021-11-05
fix(menu-toggle): added rounded corners to primary

### LINKED ISSUE #4493: fix(menu-toggle): primary style should match dropdown primary style
Dropdown and design guidance feature rounded corners whereas menu toggle does not.

**Dropdown**

Screen Shot 2021-11-05 at 9 48 51 AM

**Menu toggle**

Screen Shot 2021-11-05 at 9 48 46 AM

**Design guidance**

Screen Shot 2021-11-05 at 9 48 58 AM

mattnolting:
closes #4493

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -1,5 +1,5 @@
 .pf-c-menu-toggle {
-  --pf-c-menu-toggle--base--BorderRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-menu-toggle--BorderRadius: 0;
   --pf-c-menu-toggle--PaddingTop: var(--pf-global--spacer--form-element);
   --pf-c-menu-toggle--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-menu-toggle--PaddingBottom: var(--pf-global--spacer--form-element);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -130,6 +131,8 @@
     --pf-c-menu-toggle--active--BackgroundColor: var(--pf-c-menu-toggle--m-primary--active--BackgroundColor);
     --pf-c-menu-toggle--m-expanded--Color: var(--pf-c-menu-toggle--m-primary--m-expanded--Color);
     --pf-c-menu-toggle--m-expanded--BackgroundColor: var(--pf-c-menu-toggle--m-primary--m-expanded--BackgroundColor);
+
+    border-radius: var(--pf-c-menu-toggle--m-primary--BorderRadius);
```
mcoker, line 135: wdyt about adding this to the base toggle and using a var for the primary variation? then we can add that for other variations if we add them.

mcoker (commented):
just one question

src/patternfly/components/MenuToggle/menu-toggle.scss
mcoker: oh sorry what I meant was like:

```
.pf-c-menu-toggle {
  --pf-c-menu-toggle--BorderRadius: 0;
  --pf-c-menu-toggle--m-primary--BorderRadius: var(--pf-global--BorderRadius--sm);

  border-radius: var(--pf-c-menu-toggle--BorderRadius);

  &.pf-m-primary {
    --pf-c-menu-toggle--BorderRadius: var(--pf-c-menu-toggle--m-primary--BorderRadius);
  }
}
```

Then we can just update the `--BorderRadius` var any time we need to introduce a border-radius to a variation instead of defining the property per variation.

mcoker (approved):
LPTM!!

PR #4499, opened 2021-11-08
fix(dual-list-selector): align a11y to react

### LINKED ISSUE #4413: Dual List Selector: Update markup to match react
`. pf-c-dual-list-selector__item` used to be a `button` element, but because the `li` tag has the `role='option'` the  screen readers expect the `li` to have the focus and event handling attached.  
So in react, we changed the `button` to a `span` and when using the keyboard to navigate the items, the browser focus is placed on the `li.pf-c-dual-list-selector__list-item` instead, so the `focus` styles which were previously on the button will now need to be applied to the `li`.

srambach:
Fixes #4413 and also a few more inconsistencies not mentioned in the issue. 

Important notes:
This PR is currently failing A11y checks due to a nested interactive component where it's not allowed in the draggable version. 
The React component currently has a span there instead of a button for the drag handle - but this was a stop gap measure and the plan is to put it back. I could remove the button temporarily here, but changing it to a span (as the React component has) means that other a11y issues arise such as `aria-label` and `aria-labeledby` is not allowed on a span.

### DIFF
--- a/src/patternfly/components/DualListSelector/examples/DualListSelector.md
+++ b/src/patternfly/components/DualListSelector/examples/DualListSelector.md
@@ -38,7 +38,7 @@ cssPrefix: pf-c-dual-list-selector
       {{/dual-list-selector-status-text}}
     {{/dual-list-selector-status}}
     {{#> dual-list-selector-menu}}
-      {{#> dual-list-selector-list dual-list-selector-list--LabelledBy=(concat dual-list-selector-pane--id '-status-text')}}
+      {{#> dual-list-selector-list}}
         {{#> dual-list-selector-list-item}}
           {{> dual-list-selector-item dual-list-selector-item--text="Item1"}}
         {{/dual-list-selector-list-item}}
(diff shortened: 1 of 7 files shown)


src/patternfly/components/DualListSelector/examples/DualListSelector.md
```diff
@@ -33,12 +33,12 @@ cssPrefix: pf-c-dual-list-selector
       {{/dual-list-selector-tools-actions}}
     {{/dual-list-selector-tools}}
     {{#> dual-list-selector-status}}
-      {{#> dual-list-selector-status-text}}
+      {{#> dual-list-selector-status-text dual-list-selector-status-text--id=(concat dual-list-selector-pane--id '-status-text')}}
```
mcoker, line 36: Instead of doing this in each HBS block, could we just generate the status text `id` and `aria-labelledby` directly in the HBS partial? We could do that by setting this in the status text partial (beware, untested code ahead)

`id="(concat dual-list-selector--id '-' dual-list-selector-pane--id '-status-text')"`

and this in the panel partial

 `aria-labelledby="(concat dual-list-selector--id '-' dual-list-selector-pane--id '-status-text')"`

mcoker (changes requested):
Looks good to me! Just a couple of nits/questions, also you probably don't need these declarations now

https://github.com/patternfly/patternfly/blob/77ac5fba069f19223135f1c256138cd4e0f42d50/src/patternfly/components/DualListSelector/dual-list-selector.scss#L269-L270

mcoker (approved):
🥳

Issue #4359, opened 2021-09-03
Non-actionable card variant


mcarrano:
Introduce a new non-actionable design variant according to the designs here: https://marvelapp.com/prototype/6765ag3

mceledonia:
@mcarrano In order for this to align with the new selectable card designs the bar would need to be shifted to the bottom, just wanted to give a heads up!

PR #4500, opened 2021-11-08
feat(card): added non-selectable card, updated card view demo

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4359

* Adds a `pf-m-non-selectable-raised` variant
* Updates the card view demo to use the new selectable-raised variant instead of the old hoverable variant, and add a selected and non selectable card

example - https://patternfly-pr-4500.surge.sh/components/card#non-selectable
card view demo - https://patternfly-pr-4500.surge.sh/demos/card-view/html-demos/card-view/

FYI, here is what the card view demo looks like with flat cards:
Screen Shot 2021-11-08 at 3 45 41 PM

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -61,6 +61,10 @@
   --pf-c-card--m-selectable-raised--m-selected-raised--before--Transition: transform .25s linear;
   --pf-c-card--m-selectable-raised--m-selected-raised--before--TranslateY: calc(var(--pf-c-card--m-selectable-raised--m-selected-raised--TranslateY) * -1);
   --pf-c-card--m-selectable-raised--m-selected-raised--before--ScaleY: 2;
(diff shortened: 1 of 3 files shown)


mcarrano (commented):
This looks good @mcoker . Except that for the Selectable card, we seem to have lost the animation when mouse-down on the card.

mcoker:
> This looks good @mcoker . Except that for the Selectable card, we seem to have lost the animation when mouse-down on the card. 

I had initially only enabled the selected styling on mouse down to demonstrate the selected animation in core, since the selected animation will normally trigger when the react component adds the selected class. When I put up this PR, I realized I accidentally left that in, so I removed it. However, I verified that the selected styling still triggers with the selected class as expected after this update.

mcarrano (approved):
Thanks for the explanation @mcoker . Looks good then!

mattnolting (approved):
LGTM!

PR #4502, opened 2021-11-09
feat(truncate): added truncate component

### LINKED ISSUE #4129: Request for truncation component
We have come across the need for truncation in the middle of a string in OpenShift to fit in a given amount of space. It would be helpful to have a component that truncates in the middle of the string, but only when there isn’t enough space. If truncated, it should automatically add a tooltip. The number of chars can vary to ensure it fills as much available space as possible.

mattnolting:
closes #4129 

A couple of notes: 
- Only firefox currently supports `text-overflow: <string>`, all other browser hide overflow text regardless of character placement (characters could be hidden in the middle). We could control this programmatically, but could result in performance issues. 
- No modifiers are needed, the css recognizes `start` and `end` elements and logically controls where truncation occurs. 
- Text strings are scalable ie there's no `width` control other than that included by container.

### DIFF
--- a/src/patternfly/components/Truncate/examples/Truncate.md
+++ b/src/patternfly/components/Truncate/examples/Truncate.md
@@ -38,28 +38,6 @@ The truncate component contains to child elements, `.pf-c-truncate__start` and `
 </div>
 ```
 
-### In text
-```hbs
-<div style="width: 420px">
-  {{#> content}}
-    <p>
-      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Cras at lacus quis urna auctor sagittis sed vitae tellus. Morbi fringilla purus leo, molestie eleifend enim convallis in. Nam justo justo, interdum at efficitur a, posuere at mauris.
-      {{#> truncate truncate--id="basic-middle-of-line-truncation-example" truncate--modifier="pf-u-default-color-200"}}
-        {{> truncate-start truncate-start--text="Vestibulum interdum risus et enim"}}
-        {{> truncate-end truncate-end--text="faucibus, sit amet molestie est accumsan.&lrm;"}}
-      {{/truncate}} Cras tincidunt rhoncus massa, ac consectetur purus luctus nec. Vestibulum imperdiet turpis sed elit facilisis, at ultrices eros tempor. Curabitur porta dolor malesuada accumsan varius. Quisque ut accumsan nulla. Curabitur ac libero est.
-    </p>
-    <p>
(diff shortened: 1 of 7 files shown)


mcarrano (commented):
This is looking good @mattnolting . Just a couple of comments...

* For the middle truncation, I don't think we want a space after the "..." Usually this is used to truncate something like a long name in a list or table and it will be confusing to have the space.
* I'm not sure that middle truncation inside of a paragraph/text block would ever make sense as this is only really useful for things like filenames, server names or similar objects where seeing the start and end of the string is most important. Having the capability is fine, but I worry that the example might be misleading.

nicolethoen (approved):
LGTM after addressing Matt's questions.
It even works for the screen reader :)

mcoker (commented):
This is really cool! The clipping of the middle of characters is odd though.
Screen Shot 2021-11-10 at 5 45 25 PM



And the RTL trick (ellipsis at the beginning of the string) doesn't work on Safari :( I ran into that in another issue recently. Maybe we don't offer that variation? From the issue:

> I'm unsure if we really need an ellipsis at the start. The primary use cases are for middle and end.



Screen Shot 2021-11-10 at 5 54 21 PM

mcarrano:
@mattnolting @mcoker Interestingly I'm see a different result for middle truncation between Firefox and Chrome.

Firefox:
Screen Shot 2021-11-11 at 10 47 28 AM

Chrome:
Screen Shot 2021-11-11 at 10 48 01 AM

I'm also a little unclear how this works from looking at the code. Looks like we're passing two different strings? I would just expect to pass a single string and have the component figure out where the middle is. Maybe I'm not understanding how this works.

Less concerned about the Safari issue. Maybe we either don't support leading truncation or just live with the bug?

mcoker:
> different result for middle truncation between Firefox and Chrome

@mcarrano Firefox is smart enough to not clip part of a character, so that space is where a new character will show once there is enough room for it. The chrome screenshot has cut off the left side of the "t".

> I would just expect to pass a single string and have the component figure out where the middle is.

I believe this would require JS to monitor the containing box and look for overflow, and replace characters in the middle accordingly?

mcoker:
met with @mcarrano and due to the gotchas with this approach, we're going to put this PR on hold and begin the react work and see if we can use this or maybe go with a more JS oriented approach particularly with middle truncation.

mattnolting:
Standard, middle, and start truncation solutions works in 
- Firefox
- Opera 
- Edge (go figure)

Standard only truncation fallback for Safari.

mcoker (approved):
LGTM just a docs nit

Issue #4503, opened 2021-11-09
Log viewer line index background color is broken on dark theme


DaoDaoNoCode:
<img width="839" alt="Screen Shot 2021-11-09 at 2 34 13 PM" src="https://user-images.githubusercontent.com/37624318/140992299-535b64bd-d4ad-4c83-9cf6-e93a288ae473.png">

PR #4506, opened 2021-11-10
fix(log-viewer): update way dark theme is applied

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4503


DaoDaoNoCode (approved):
LGTM, thanks! @mcoker

PR #4508, opened 2021-11-10
fix(dropdown): applied primary styling to split

### LINKED ISSUE #4130: isPrimary does not work for Split Button
Setting the DropdownToggle to use primary button styling does not seem to work for a Split Button (Dropdown component), even though DropdownToggle already seems to support isPrimary for a regular Dropdown.

It looks like a regular Split Button:
image

Would instead expect it to look something like a regular Dropdown with a primary toggle, like:
https://www.patternfly.org/2021.06/components/dropdown/#primary-toggle
image

A minimal example to reproduce: https://codesandbox.io/s/hopeful-platform-lrj0z?file=/index.js (not sure if anyone can access this, so here's the code):
```
import ReactDOM from "react-dom";
import "@patternfly/react-core/dist/styles/base.css";

import * as React from "react";
import {
  Dropdown,
  DropdownToggle,
  DropdownToggleAction
} from "@patternfly/react-core";

class PrimarySplitButton extends React.Component {
  render() {
    const dropdownToggle = (
      <DropdownToggle
        isPrimary
        key="split-button-toggle-id"
        splitButtonVariant="action"
        splitButtonItems={[
          <DropdownToggleAction key="split-button-toggle-action">
            Default action
          </DropdownToggleAction>
        ]}
      />
    );
    return <Dropdown toggle={dropdownToggle} />;
  }
}

ReactDOM.render(<PrimarySplitButton />, document.getElementById("root"));
```

Believe this is a bug. Fix would be targeting Patternfly 4.

mattnolting:
closes #4130

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -56,6 +56,18 @@ $pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "
   --pf-c-dropdown__toggle--m-split-button__toggle-check__input--TranslateY: #{pf-size-prem(-1px)};
   --pf-c-dropdown__toggle--m-split-button__toggle-text--MarginLeft: var(--pf-global--spacer--sm);
 
+  // Split button
+  --pf-c-dropdown__toggle--m-split-button--m-primary--child--BorderRadius: var(--pf-global--BorderRadius--sm);
+  --pf-c-dropdown__toggle--m-split-button--m-primary--child--BackgroundColor: var(--pf-global--primary-color--100);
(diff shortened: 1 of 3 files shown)


mcoker (changes requested):
LGTM, although hovering over the arrow gives this styling
Screen Shot 2021-11-10 at 4 52 32 PM

@mcarrano @mceledonia can you also review please?

mcarrano (commented):
I guess this is what was requested. But it looks really odd now that I see it. What do you think @mceledonia ?

mcoker:
@mattnolting - met with @mcarrano and looks like we can just remove the example with the checkbox. The checkbox shouldn't be used with the primary variation.

mcoker (approved):

PR #4513, opened 2021-11-11
feat(menu): added nav variant

### LINKED ISSUE #4510: Update nav flyout menu to work with popper
re: https://github.com/patternfly/patternfly-react/pull/6562

The flyout menu needs to live outside of the sidebar since the sidebar hides its overflow. We will need to:

* Move the current flyout menu styles in the nav component into the menu component under a modifier (`.pf-m-flyout-nav`?) - or maybe there is a better solution?
* Add a modifier to nav flyout items that applies hover styles when the nav item has a flyout menu open (`.pf-m-current-path` borrowed from menu, or maybe just `.pf-m-hover`?)

mattnolting:
closes #4510 

example is for pre-merge demo only


nicolethoen (approved):
That's a lot of CSS magic! But I think the HTML all looks good to me. It's in a popper but the css does not need popper dependent. Very cool to be able to skin a menu like that

mcoker (commented):
LGTM! A few thoughts:

* I could see other uses of the menu in the nav that aren't flyout - should we call this `pf-m-nav-flyout`? or `pf-m-flyout-nav`?
* I think you could create this popper example you have as something we leave in,  just remove the popper box and have the example page nav.css either position the menu itself or some dummy div that wraps the menu relative to the nav.
* I think the nav item that opens the flyout is going to need some class to give it hover styles for when the flyout is open since it needs to appear hovered, but you aren't hovering anything when interacting with the flyout since it's outside of the nav component.

mcoker:
@mattnolting I updated the flyout examples and added a hover class for the nav link. Feel free to merge if it looks good to you.

mcoker (approved):

PR #4520, opened 2021-11-15
chore(card): put selects into demos

### LINKED ISSUE #4471: Change dropdowns in dashboard card demos to Selects
The dropdown menus in deashboard card demos were incorrectly implemented as Dropdowns. They should be Selects without borders (plain variants). This applies to the following demos:

Utilization card 3: https://www.patternfly.org/v4/components/card/html-demos#utilization-card-3
Trend card 1: https://www.patternfly.org/v4/components/card/html-demos#trend-card-1
Log view: https://www.patternfly.org/v4/components/card/html-demos#log-view
Events view: https://www.patternfly.org/v4/components/card/html-demos#events-view

srambach:
Fixes #4471 

Note: The width limiter was removed because it causes an overflow in the expanded menu.

### DIFF
--- a/src/patternfly/components/Select/examples/Select.md
+++ b/src/patternfly/components/Select/examples/Select.md
@@ -46,6 +46,7 @@ The single select should be used when the user is selecting an option from a lis
 | `.pf-c-select__menu-item` | `<li>` |  Initiates the items in the custom select dropdown menu. |
 | `.pf-c-select__menu-item-icon` | `<i>` |  Initiates the selected item icon. |
 | `.pf-m-top` | `.pf-c-select` |  Modifies the select menu to display above the toggle. |
+| `.pf-m-align-right` | `.pf-c-select` |  Modifies the select menu to display right aligned to the toggle. |
 | `.pf-m-active` | `.pf-c-select` | Forces display of the active state of the toggle. |
 
 ## States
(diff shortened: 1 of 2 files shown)


src/patternfly/demos/Card/examples/Card.md
```diff
@@ -753,7 +753,9 @@ import './Card.css'
         {{/title}}
       {{/card-title}}
       {{#> card-actions}}
-        {{> dropdown dropdown--id=(concat card--id '-dropdown') dropdown-toggle--text="Filter" dropdown-menu--modifier="pf-m-align-right"}}
+        {{#> select id=(concat card--id '-select-dropdown')}}
```
mcoker, line 756: You can align these menus the same way as dropdowns - https://github.com/patternfly/patternfly/blob/a67b16945bec19cae3916d9688fe1e9c6cba2576/src/patternfly/components/Select/select.scss#L463-L465

Looks like it isn't in the select docs - could you add it to this section under `.pf-m-top`? https://pf4.patternfly.org/components/select#usage

mcoker (changes requested):

src/patternfly/demos/Card/examples/Card.md
srambach: Yup, just found that. Fixed the docs.

src/patternfly/components/Select/examples/Select.md
```diff
@@ -46,6 +46,7 @@ The single select should be used when the user is selecting an option from a lis
 | `.pf-c-select__menu-item` | `<li>` |  Initiates the items in the custom select dropdown menu. |
 | `.pf-c-select__menu-item-icon` | `<i>` |  Initiates the selected item icon. |
 | `.pf-m-top` | `.pf-c-select` |  Modifies the select menu to display above the toggle. |
+| `.pf-m-align-right` | `.pf-c-select` |  Modifies the select menu to display right aligned to the toggle. |
```
mcoker, line 49:
```suggestion
| `.pf-m-align-right` | `.pf-c-select__menu` |  Modifies the select menu to display right aligned to the toggle. |
```

mcoker (changes requested):

mcoker (approved):

Issue #4354, opened 2021-08-26
Progress button completion state


ajacobs21e:
I have a use case that is either an extension of the progress button or a new component. 
I would like to use the progress button to show that an action is happening in the background, and then change the button state to show that the action has completed. This could possibly also show that the action has failed, although that's not needed for my use case.

My example:
![progress button](https://user-images.githubusercontent.com/6415742/131010562-31781e00-0c50-479c-a58d-f00e1c13b4fe.png)
User goes to log in to their Red Hat account and is prompted to link their company IdP to their Red Hat account. During this process, change the button to a progress button to show the linking is happening (and also disable the button so the user can't submit multiple times), then show that the link is complete for a few seconds before redirecting to the user's destination.
In this use case I'm using the Brand primary button color.
I can't use a toast on the next step because at that point the user has left the SSO app. It would require adding code to every system that uses Red Hat SSO which isn't viable. Another option would be to show an interstitial page or a toast on this page before it redirects. 

Is this something that can be achieved with existing components + js?

ajacobs21e:
@mcarrano sorry I missed this. I don't have a specific deadline. Is this something that can still be made into a demo please? Thanks!

mcarrano:
Yes. This is currently on our roadmap for January.

PR #4528, opened 2021-11-17
docs(button): adds demo for progress button completion state

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4354

### DIFF
--- a/src/patternfly/demos/Button/examples/Button.md
+++ b/src/patternfly/demos/Button/examples/Button.md
@@ -1,6 +1,5 @@
 ---
 id: 'Button'
-beta: true
 section: components
 cssPrefix: pf-d-button
 ---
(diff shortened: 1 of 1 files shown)


ajacobs21e:
@mcoker The button part looks good, thank you!

@mcarrano is there a standard set of dummy data we use for demos? I don't think we should use "Red Hat login" or my name in the example.

mcoker:
@ajacobs21e I changed it to something more generic - does this look ok?

Screen Shot 2021-11-17 at 4 27 19 PM

ajacobs21e:
@mcoker yes that's better! Can you make username a default input and put ***** in the password field so it looks like the form has been completed please?

mcarrano (commented):
@mcoker I don't have any comments beyond what @ajacobs21e has already flagged. But I am curious why we need the beta feature heading on the Demo page. Is that needed for something that's just a demo?

mcoker:
@mcarrano @ajacobs21e updated. The beta flag is just part of the new demo template - removed it.

ajacobs21e:
LGTM! @mcarrano are you ok with the button text "link account and log in"? There's probably something more generic we could use but I am blanking on a better suggestion. It probably doesn't matter.

mcarrano (approved):
I think that the example text is fine @ajacobs21e . This looks good to me! Thanks @mcoker .

ajacobs21e (approved):

Issue #4280, opened 2021-03-31
pf-c-table: why do we need to much padding in the nested tables? 


KKoukiou:
I recently noticed that we have a local override locally to revert this rule for nested tables paddings.

https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/Table/table.scss#L224
 
With your paddings our nested tables look like this:
![Screen Shot 2021-03-31 at 12 06 46](https://user-images.githubusercontent.com/14921356/113128181-d3c9fd80-9219-11eb-93c7-0a5c21e7ab84.png)

Reverting this rule (which is what we do in Cockpit) makes the nested tables looks like this:

![Screen Shot 2021-03-31 at 12 10 06](https://user-images.githubusercontent.com/14921356/113128449-1986c600-921a-11eb-913b-160da5edf527.png)

Design is definitely not my field, but it looks like the first variant has too much left padding. Do we really need this rule?

PR #4529, opened 2021-11-17
fix(table): removed extra padding from nested table cells

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4280

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -763,9 +763,7 @@ When header cells are empty or they contain interactive elements, `<th>` should 
     {{/table-tr}}
 
     {{#> table-tr table-tr--expandable="true" table-tr--IsExpanded="true"}}
-      {{> table-td table-td--IsEmpty="true"}}
-      {{> table-td table-td--IsEmpty="true"}}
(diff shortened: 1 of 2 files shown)


mcarrano:
Is this the example I should be looking at @mcoker ? https://patternfly-pr-4529.surge.sh/components/table#expandable-with-nested-table-example  I thought the intent was to remove padding, but in the example looks like the nested table has even more padding.

mattnolting (approved):
LPTM

mcoker:
@mcarrano yep that's a new example showing a simple table nested in expandable content. That extra padding is coming from the expanded content element. Updated it to remove all of that padding and just have standard padding. Another example to look at is compound expansion, since that includes a nested table - https://patternfly-pr-4529.surge.sh/components/table/#compound-expansion-example

If we have a better use case/design for a nested table I can use to modify the example I added, let me know and I'll update it.

mcarrano:
Yes, this looks much better @mcoker . Is this what you were expecting @KKoukiou ?

KKoukiou:
> Yes, this looks much better @mcoker . Is this what you were expecting @KKoukiou ?

Indeed! The new example is what I would expect! Thanks for fixing this!

PR #4531, opened 2021-11-19
feat(demos): adds a password generator demo

### LINKED ISSUE #4340: Password generator demo
In addition to the password strength indicator, this would be an enhancement to the input field that allows for a menu to appear once selecting the field (similar to typeahead) that presents a randomly generated value. We are going with a different way of approaching this for the time being in OCM, but it'd greatly improve the feature if we were able to do what I have below! My guess is that this would be best to add to the password strength component that was recently designed. No real rush on the timeline here - the sooner it's released, the sooner we can update our feature to use it!

Screen Shot 2021-06-07 at 10 15 47 AM

Screen Shot 2021-06-07 at 10 16 02 AM

srambach:
This adds a new demo showing a form input field in combination with a menu to show a suggested password and button to regenerate a new password.

Mockup here: https://marvelapp.com/prototype/c7j2b6b/screen/79266406

Fixes #4340

### DIFF
--- a/src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
+++ b/src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
@@ -30,7 +30,7 @@ section: demos
           {{#> menu-content}}
               {{#> menu-list}}
                 {{#> menu-list-item}}
-                  {{#> menu-item menu-item--modifier="pf-m-selected"}}
+                  {{#> menu-item}}
                     {{#> menu-item-main}}
                       {{#> menu-item-text}}
                         Use suggested password: fqu9kKe676JmKt2
--- a/src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
+++ b/src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
@@ -1,6 +1,5 @@
 ---
 id: 'Password generator'
-beta: false
 section: demos
 ---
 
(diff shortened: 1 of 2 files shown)


src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
```diff
@@ -0,0 +1,57 @@
+---
+id: 'Password generator'
+beta: false
```
mcoker, line 3: You could just remove this line.

src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
```diff
@@ -0,0 +1,57 @@
...
+          {{> form-group-label-help}}
+        {{/form-group-label-main}}
+        {{#> form-group-label-info}}
+        {{/form-group-label-info}}
+      {{/form-group-label}}
+      {{#> form-group-control}}
+        {{#> input-group}}
+            {{#> form-control controlType="input" input="true"  form-control--attribute=(concat 'required type="text" id="' form--id '-password" name="' form--id '-password" aria-label="Password input" value="" placeholder="Password"')}}
+          {{/form-control}}
+          {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Show password button for password input"'}}
```
mcoker, line 26: Should we set the password to hidden by default? I think that's how I would expect it. Also I think the design has the icon wrong. If the password is hidden w/ `type="password"`, the icon should be the regular eye. And if it's visible w/ `type="text"`the icon should be the eye with the line through it, as you have it here.

Also this should probably just say "Show password" as the action the button performs. Also looks like we should update the password strength demos. This will match the login page demo with the show/hide password button - https://www.patternfly.org/v4/components/login-page/react/showhide-password/

src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
```diff
@@ -0,0 +1,57 @@
...
+            <i class="fas fa-eye-slash" aria-hidden="true"></i>
+          {{/button}}
+        {{/input-group}}
+        {{#> menu}}
+          {{#> menu-content}}
+            {{#> menu-group}}
```
mcoker, line 32: Do we need to use a group here?

mcoker (changes requested):

src/patternfly/demos/PasswordGenerator/examples/PasswordGenerator.md
```diff
@@ -0,0 +1,54 @@
...
+          {{/form-control}}
+          {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Show password"'}}
+            <i class="fas fa-eye" aria-hidden="true"></i>
+          {{/button}}
+        {{/input-group}}
+        {{#> menu}}
+          {{#> menu-content}}
+              {{#> menu-list}}
+                {{#> menu-list-item}}
+                  {{#> menu-item menu-item--modifier="pf-m-selected"}}
```
mcoker, line 33: Is `pf-m-selected` needed?

```suggestion
                  {{#> menu-item}}
```

mcoker (changes requested):
Looks fantastic!! Just one thing

mcoker (approved):
🥳

Issue #4524, opened 2021-11-15
page: demo: the icon group in the toolbar in the masthead top is not clear when focused / selected


KKoukiou:
I am following https://www.patternfly.org/v4/components/page/react-demos/sticky-section-group/ for adapting cockpit's top panel to PF4.

I think when the background is dark, like here, we need some classes which will outline / highlight the elements which are focused/active. 
In this example, the icons on the top, when focused, have almost no difference for the user.

mcoker:
cc @mcarrano wdyt? The attention bell should probably be a notification badge, assuming it should toggle a notification drawer.

mcarrano:
Yes, that's true @mcoker . I checked some of the other React demos and looks like they all have the same problem except for the Notification drawer demo that uses the correct badge. I don't see this in the core demos though. I'm wondering if we should just remove the notification icon from all demos except for the ones directly related to Notification drawer. Thoughts?

But I'm not really sure that addresses the issue that @KKoukiou raises. It's true that the focus/hover effect for these icons is very subtle and hard to notice. But it's been that way since we introduced PF4. @mceledonia any thoughts?

mceledonia:
Yeah we need some focus states here and I would argue the active state is the drawer/menu itself, but not opposed to keeping the hover styling while active. 

For focus states, is there a reason we aren't applying our usual focus states on these icons?

PR #4533, opened 2021-11-19
fix(global): removed custom firefox focus styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4524

These custom focus styles were to fix a problem with buttons being different sizes in firefox, however that has since been fixed. This restores the default focus styles for buttons. @mcarrano @mceledonia do you see this as a breaking change? It's definitely a bug fix as far as I'm concerned, but it will change the default focus style for buttons in FF so it could result in an unexpected visual change for FF users.

### DIFF
--- a/src/patternfly/base/_globals.scss
+++ b/src/patternfly/base/_globals.scss
@@ -127,22 +127,6 @@
     cursor: pointer;
   }
 
-  button,
-  [type="button"],
-  [type="reset"],
-  [type="submit"] {
-    // Remove the inner border and padding in Firefox.
-    &::-moz-focus-inner {
-      padding: 0;
-      border-style: none;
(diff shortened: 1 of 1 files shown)


mcarrano (commented):
@mcoker I am OK with this change, and it does fix the issue with focus styles in the masthead. What do you think @mceledonia ? Do you see any issues with making this change?

mcoker:
@mcarrano yep, you can test it here - https://patternfly-pr-4533.surge.sh/components/masthead/html-demos/advanced-integration-with-menu-options/

srambach (approved):
LGTM 👍

mceledonia:
LGTM, no issues here!

PR #4536, opened 2021-11-23
feat(form): add the ability for custom breakpoints

KKoukiou:


src/patternfly/components/Form/examples/Form.md
```diff
@@ -600,3 +617,4 @@ cssPrefix: pf-c-form
 | `.pf-m-inline` | `.pf-c-form__group-control` | Modifies form group children to be inline (this is primarily for radio buttons and checkboxes). |
 | `.pf-m-stack` | `.pf-c-form__group-control` | Modifies form group children to be stacked with space between children. |
 | `.pf-m-expanded` | `.pf-c-form__field-group` | Modifies an expandable field group for the expanded state. |
+| `.pf-m-grid-sm`, .pf-m-grid-md`, `.pf-m-grid-lg`, `.pf-m-grid-xl`, `.pf-m-grid-2xl` | `.pf-c-form` | Changes horizontal layout to vertical, at suffixed breakpoint. By default, pf-c-global--breakpoint--md is used |
```
mcoker, line 620: It looks to me like you're adding support for modifying when the horizontal layout is applied. Typically we would do that via `.pf-m-horizontal{-on-[breakpoint]}`.

Seems like you could do it with something like this:

```
// Add this to top of form.scss - creates a map of the breakpoint suffixes to use
$pf-c-form--m-horizontal--breakpoint-map: build-breakpoint-map("sm", "md", "lg", "xl", "2xl");

// updates the mixin name to something more unique to avoid namespace conflicts
@mixin pf-c-form--m-horizontal {

    // default horizontal styles applied at medium breakpoint
    @media screen and (min-width: $pf-global--breakpoint--md) {
      @content;
    }

    // custom xs style since this isn't in our maps - maps are based on mobile first design. This may not even be a valid option based on design but I guess it doesn't hurt to add.
    &-on-xs {
        @media screen and (min-width: $pf-global--breakpoint--xs) {
          @content;
        }
    }

    // loop to build the rest of the breakpoints
    @each $breakpoint, $breakpoint-value in $pf-c-form--m-horizontal--breakpoint-map {
        &-on-#{$breakpoint} {
            @include pf-apply-breakpoint($breakpoint) {
               @content
            }
        }
    }
}

// you could place `@include pf-c-form--m-horizontal {}` in the existing `&.pf-m-horizontal` block like this:
 .pf-c-form {
     &.pf-m-horizontal {
        @include pf-c-form--m-horizontal {
            /* grid styles */
        }
     }
 }
```

mcoker (changes requested):
Hey @KKoukiou thanks for the PR! I think this is a great add. Left a comment with some suggestions. Let me know what you think!

src/patternfly/components/Form/form.scss
```diff
@@ -1,3 +1,29 @@
+$pf-c-form--m-horizontal--breakpoint-map: build-breakpoint-map("sm", "md", "lg", "xl", "2xl");
+
+// updates the mixin name to something more unique to avoid namespace conflicts
```
mcoker, line 3: Can you remove these comments?

src/patternfly/components/Form/examples/Form.md
```diff
@@ -588,7 +605,7 @@ cssPrefix: pf-c-form
 | `.pf-c-form__field-group-header-description` | `<div>` | Initiates the form field group description. |
 | `.pf-c-form__field-group-header-actions` | `<div>` | Initiates the form field group actions container. |
 | `.pf-c-form__field-group-body` | `<div>` | Initiates the form field group body. |
-| `.pf-m-horizontal` | `.pf-c-form` | Modifies form for a horizontal layout. |
+| `.pf-m-horizontal{-on-[sm, md, lg, xl, 2xl]}` | `.pf-c-form` | Modifies the form for a horizontal layout. |
```
mcoker, line 608: adding xs and updating the description 

```suggestion
| `.pf-m-horizontal{-on-[xs, sm, md, lg, xl, 2xl]}` | `.pf-c-form` | Modifies the form for a horizontal layout at an optional breakpoint. The default breakpoint is `-md`. |
```

src/patternfly/components/Form/examples/Form.md
```diff
@@ -67,6 +67,23 @@ cssPrefix: pf-c-form
 {{/form}}
 ```
 
+### Form custom breakpoints for horizontal layout
```
mcoker, line 70: Matches the example above it a little better

```suggestion
### Horizontal layout at a custom breakpoint
```

mcoker (changes requested):
Looks great! Just a few nits.

src/patternfly/components/Form/form.scss
KKoukiou: removed :) Thanks for the patience reviewing this!

mcoker (approved):
🥳

Issue #4501, opened 2021-11-09
Update stylelint to v14.0.1


mcoker:
The VSCode plugin is no longer compatible with v13 - https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint

Looks like latest is v14.0.1 https://www.npmjs.com/package/stylelint

PR #4537, opened 2021-11-23
fix(stylelint): update stylelint, deps

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4501

@dgutride @evwilkin after updating stylelint, I the build was failing with something using postcss-scss. I updated postcss to latest and it's fixed. I'm not sure where all postcss is used - do you think that's ok?


mcoker:
@rhamilto - this PR updates postcss in PF core from `^7.0.18` to`^8.3.11`. I wasn't sure if y'all compile PF SCSS yourselves in OpenShift - do you anticipate this causing any issues with OS?

rhamilto:
> @rhamilto - this PR updates postcss in PF core from `^7.0.18` to`^8.3.11`. I wasn't sure if y'all compile PF SCSS yourselves in OpenShift - do you anticipate this causing any issues with OS?

We do.  But we're already using [^8.2.13](https://github.com/openshift/console/blob/f1b7d2a37bcea66481f677d9ae03c75741295e8d/frontend/package.json), so I doubt it'll be a problem.  But cc-ing @spadgett, @kdoberst, @sg00dwin in case anything weird crops up after we adopt this version.

nicolethoen (approved):

PR #4542, opened 2021-11-29
feat(dropdown): added plain text variant

### LINKED ISSUE #4282: Create plain text variant for dropdown component
Similar to the options menu, it's a plain toggle with text (not an icon, which is what `.pf-m-plain` is for currently)

mattnolting:
closes #4282


mcoker (changes requested):
Looks great! Just one nit, and can you also add a height to the plain text example so the expanded menu doesn't overflow?

PR #4546, opened 2021-12-01
fix(table): added label to checkbox

### LINKED ISSUE #4355: Increase click target for selectable tables
The click target for selectable tables is very small - 13x13 px
image

Vs the space that the select takes is much larger - 53px wide here
image

Source: https://www.patternfly.org/v4/components/table#composable-selectable

After observing users struggle to hit the click target on console.redhat.com, I am requesting an enhancement to increase the click target for selectable tables to 44x44 px, which is the WCAG AAA recommendation (https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#dfn-pointer-input)

mattnolting:
closes #4355 

Updates start here: https://patternfly-pr-4546.surge.sh/components/table#checkboxes-and-actions-example

Each example including a  checkbox is updated

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -864,11 +864,7 @@
 
 // stylelint-disable
 // Minimize width
-.pf-c-table .pf-c-table__check {
-  --pf-c-table--cell--MinWidth: calc(var(--pf-c-table--cell--PaddingRight) + #{pf-size-prem(14px)} + var(--pf-c-table--cell--PaddingLeft)); // set min-width to padding + checkbox width
-  --pf-c-table--cell--Width: 1%;
-}
-
+.pf-c-table .pf-c-table__check,
 .pf-c-table .pf-c-table__toggle,
 .pf-c-table .pf-c-table__action,
 .pf-c-table .pf-c-table__favorite,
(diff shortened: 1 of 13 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -902,9 +912,19 @@
...
   --pf-c-table--cell--FontSize: var(--pf-c-table__check--input--FontSize);
 
-  > input:not([disabled]) {
+  label:not([disabled]),
+  input:not([disabled]) {
     cursor: pointer;
   }
+
+  label {
+    position: absolute;
```
mcoker, line 921: Just a thought - would it make more sense to have the same padding as the table cell with negative margins equal to that padding? Then you wouldn't have to manually set the cell width.

mcoker (commented):
LGTM! Just one question. Also assuming we'll want to do this with the radio buttons, but we can open a separate issue.

src/patternfly/components/Table/table.scss
mattnolting: Yeah, I think that's a good point. Let me try that.

wise-king-sullyman:
It seems like [tree table with checkbox](https://patternfly-pr-4546.surge.sh/components/table/#tree-table-with-checkboxes) and tree table with checkbox and icons still require clicking directly on the checkbox to me.

src/patternfly/components/Table/table.scss
```diff
@@ -902,9 +908,16 @@
...
 
-  > input:not([disabled]) {
+  label:not([disabled]),
+  input:not([disabled]) {
     cursor: pointer;
   }
+
+  label {
+    display: block;
+    padding: inherit;
```
mcoker, line 918: 🔥

mattnolting:
> It seems like [tree table with checkbox](https://patternfly-pr-4546.surge.sh/components/table/#tree-table-with-checkboxes) and tree table with checkbox and icons still require clicking directly on the checkbox to me.

@wise-king-sullyman Ah yes, duly noted. I believe updating the checkbox here requires a design update as there isn't enough space to accommodate 44px width. I've opened https://github.com/patternfly/patternfly/issues/4554 to address this issue.

mcoker (approved):
L😎TM!

Issue #4521, opened 2021-11-15
Add multiple file upload


mcarrano:
Want to add the capability for multi-file upload. Design here: https://marvelapp.com/prototype/69jc198/screen/82770742 Need to decide if this is a new component of just a new variant of the existing file upload.

PR #4548, opened 2021-12-01
feat(multiplefileupload): added component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4521

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
+++ b/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
@@ -3,23 +3,21 @@
     {{{expandable-section-toggle--attribute}}}
   {{/if}}
   {{#if expandable-section--IsExpanded}}
-  aria-expanded="true"
+    aria-expanded="true"
   {{else}}
-  aria-expanded="false"
+    aria-expanded="false"
   {{/if}}
   {{#if expandable-section--IsDetached}}
     aria-controls="{{expandable-section--id}}-content"
-  {{/if}}
-  >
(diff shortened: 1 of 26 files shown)


mcarrano (commented):
This is looking good @mcoker . Just a couple comments...
* Can we rename this to something more like "File upload - multiple?" That way it will be easier to find in the navigation when someone looks for file upload.
* Is what I am seeing here for the horizontal the actual fixed width? Is that consistent with the design?
* Shouldn't there be a layout that includes one or more progress bars that show status of the upload in progress?

mcoker:
@mcarrano looks like you reviewed before the build was finished. Check again and let me know what you think.

Also there are 2 bugs outstanding. From 
Screen Shot 2021-12-01 at 5 32 24 PM

I'm not sure how to present a long file name and align the progress percentage text (eg, "35%"). That percentage text is aligned bottom by default in the progress component. Should I make an update?

Long text in the progress area causes the horizontal/fixed width layout width to increase. Still working to address that.

mcarrano:
Another solution would be to just truncate the filename if it exceeds the available space. Seems like this would be a pretty rare occurrence.

mcarrano (approved):
This looks great @mcoker I know there are a couple of loose ends. Feel free to open a follow-up issue if needed and we'll look to move this to React in the next milestone.

mattnolting (approved):
> This looks great @mcoker I know there are a couple of loose ends. Feel free to open a follow-up issue if needed and we'll look to move this to React in the next milestone.

+1 LGTM

Issue #4469, opened 2021-10-25
Update editable label with new structure and design


jessiehuff:
After testing in react, I think it makes the most sense to update the structure of editable label to use more semantic elements (toggling between button and input). However, the changing structure means we also need to update the styling. 

[React PR](https://github.com/patternfly/patternfly-react/pull/6472)

[Design issue](https://github.com/patternfly/patternfly-design/issues/1073) for upcoming updates

mcarrano:
@srambach regarding design changes, this is the direction that we are headed:

<img width="279" alt="Screen Shot 2021-11-17 at 2 27 40 PM" src="https://user-images.githubusercontent.com/6619860/142269949-34e9f9a3-8912-4e8e-99be-df454b49d31c.png">

For editable lables, we want to add a dotted underline below the text similar to what we do in definition lists. Let me know if you have questions or need a mre detailed visual spec.

srambach:
@mcarrano What happens when the label becomes active? Does the underline go away, or change? 
Two other things I notice; the offset from the text is smaller in the image above compared to the description list setting. And the underline does not extend all the way to the X; it is only under the text. Is that ok?

![image](https://user-images.githubusercontent.com/19825616/143948562-5d66c340-ac8d-4f47-bd95-12774c80118a.png)

mcarrano:
@srambach I think that the spacing you have is fine. I agree that it makes sense to be consistent with what we did for Description list. I also agree that the underline should not extend all the way to the X. Good question about what happens when editing is active. I hadn't expected the underline to remain after entering edit mode, but maybe it should. What do you think @mceledonia ?

mceledonia:
@gonoraj Did you have any design proposals/mocks for the active state here?

gonoraj:
@mceledonia @mcarrano  I had thought the active state would stay the same, with the blue box appearing for the browser's focus state, but I had made possible options incorporating the label while editing as well. I'm not sure what your thoughts are on this or if it's feasible (I also just added this to [this doc](https://docs.google.com/document/d/1ReoW10DzrrWxdAnzwuXbeTMdGXpFfiIAE9JHu-43o4U/edit?usp=sharing) if you prefer to discuss there). I also agree that I didn't expect the underline to remain after entering edit mode.
![active-state](https://user-images.githubusercontent.com/33614237/144121815-ef00dc8b-ef8c-4f04-89a3-12044795047d.jpg)

mcoker:
[...]

PR #4551, opened 2021-12-01
fix(label): adjust style of editable

srambach:
This fixes the styling of the editable label, but the structure is in the inline edit component and will be addressed in a future PR.

#4469 

[design](https://docs.google.com/document/d/1ReoW10DzrrWxdAnzwuXbeTMdGXpFfiIAE9JHu-43o4U/edit)

[React PR](https://github.com/patternfly/patternfly-react/pull/6472)

### DIFF
--- a/src/patternfly/components/Label/label-editable-text.hbs
+++ b/src/patternfly/components/Label/label-editable-text.hbs
@@ -1,8 +1,10 @@
 <{{#if label-editable-text--type}}{{label-editable-text--type}}{{else if label--IsEditableActive}}input{{else}}button{{/if}} class="pf-c-label__editable-text{{#if label-editable-text--modifier}} {{label-editable-text--modifier}}{{/if}}"
-  id="{{label--id}}-editable-text"{{#if label--IsEditableActive}} 
-  type="text" 
-  value="{{~> @partial-block}}"{{else}} 
-  currvalue="{{~> @partial-block}}"{{/if}}
+  id="{{label--id}}-editable-text"
+  {{#if label--IsEditableActive}} 
+    type="text" 
+    value="{{~> @partial-block}}"{{else}} 
+    currvalue="{{~> @partial-block}}"
+  {{/if}}
   {{#if label-editable-text--aria-label}}
     aria-label="{{label-editable-text--aria-label}}"
   {{else}}
@@ -14,5 +16,3 @@
 {{#unless label--IsEditableActive}}
   {{> @partial-block}}
 </{{#if label-editable-text--type}}{{label-editable-text--type}}{{else}}button{{/if}}>{{/unless}}
-
-
--- a/src/patternfly/components/Label/label-editable-text.hbs
+++ b/src/patternfly/components/Label/label-editable-text.hbs
@@ -1,5 +1,8 @@
 <{{#if label-editable-text--type}}{{label-editable-text--type}}{{else if label--IsEditableActive}}input{{else}}button{{/if}} class="pf-c-label__editable-text{{#if label-editable-text--modifier}} {{label-editable-text--modifier}}{{/if}}"
-  id="{{label--id}}-editable-text"{{#if label--IsEditableActive}} type="text" value="{{~> @partial-block}}"{{else}} currvalue="{{~> @partial-block}}"{{/if}}
+  id="{{label--id}}-editable-text"{{#if label--IsEditableActive}} 
+  type="text" 
+  value="{{~> @partial-block}}"{{else}} 
+  currvalue="{{~> @partial-block}}"{{/if}}
   {{#if label-editable-text--aria-label}}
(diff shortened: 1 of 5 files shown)


srambach:
After discussions, this removes the inline edit component from the editable label. This needs to be verified against https://github.com/patternfly/patternfly-react/pull/6472/files to make sure the structure still works.

For completeness, I added a compact editable variant example for verification by @mceledonia. This shows a 1px underline offset. @mcarrano please let me know if we want to keep this as an example after verifying the design.

mceledonia:
Thanks Sarah that looks good to me!

mcarrano:
Yes, this looks great @srambach !

src/patternfly/components/Label/label-editable-text.hbs
```diff
@@ -0,0 +1,15 @@
+<{{#if label-editable-text--type}}{{label-editable-text--type}}{{else if label--IsEditableActive}}input{{else}}button{{/if}} class="pf-c-label__editable-text{{#if label-editable-text--modifier}} {{label-editable-text--modifier}}{{/if}}"
+  id="{{label--id}}-editable-text"{{#if label--IsEditableActive}} type="text" value="{{~> @partial-block}}"{{else}} currvalue="{{~> @partial-block}}"{{/if}}
```
mcoker, line 2: Is there a reason this isn't multi-line?

```suggestion
  id="{{label--id}}-editable-text"
  {{#if label--IsEditableActive}}
    type="text"
    value="{{~> @partial-block}}"
  {{else}}
    currvalue="{{~> @partial-block}}"
  {{/if}}
```

src/patternfly/components/Label/label.scss
```diff
@@ -277,6 +313,13 @@
     }
   }
 
+  .pf-c-label__editable-text {
+    @include pf-text-overflow;
+
+    max-width: var(--pf-c-label__editable-text--MaxWidth);
+    border-width: var(--pf-c-label__editable-text--BorderWidth);
```
mcoker, line 320: this is fine but I think it's also fine just to use `border: 0;` since I wouldn't expect it to be a themeable thing.

mcoker (changes requested):
Looking good! Left some comments. Also per the design, the text color should still change per color when it's editable. Are you able to just update the `__content` element to be a `<button>` (similar to how `__content` is an `<a>` in the link variation)? Then you wouldn't need the new element/class.

src/patternfly/components/Label/label-editable-text.hbs
```diff
@@ -0,0 +1,15 @@
...
+    aria-label="{{label-editable-text--aria-label}}"
+  {{else}}
+    aria-label="Editable text"
+  {{/if}}
+  {{#if label-editable-text--attribute}}
+    {{{label-editable-text--attribute}}}
+  {{/if}}{{#if label--IsEditableActive}} /{{/if}}>
+{{#unless label--IsEditableActive}}
+  {{> @partial-block}}
+</{{#if label-editable-text--type}}{{label-editable-text--type}}{{else}}button{{/if}}>{{/unless}}
```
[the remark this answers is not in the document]
srambach: It's inside an #unless on line 11; are you saying to combine the unless and if? 
```  {{#if label-editable-text--attribute}}
    {{{label-editable-text--attribute}}}
  {{/if}}{{#if label--IsEditableActive}} />{{else}}>
  {{> @partial-block}}
</{{#if label-editable-text--type}}{{label-editable-text--type}}{{else}}button{{/if}}>{{/if}}```

src/patternfly/components/Label/label-editable-text.hbs
srambach: just an artifact. I'll fix it.

src/patternfly/components/Label/label.scss
srambach: I wondered if someday the active state would want a border other than the focus ring. 🤷

src/patternfly/components/Label/label-editable-text.hbs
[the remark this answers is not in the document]
mcoker: Oh! I missed that, this looks good.

mcoker (approved):

jessiehuff (commented):
This looks awesome - thanks for working on it!!! 😄  Do we need to update the documentation above it?

srambach:
> This looks awesome - thanks for working on it!!! 😄 Do we need to update the documentation above it?

@jessiehuff Good catch 🤦‍♀️ Can you take another look please? I'm not totally familiar with what's happening in the JS so if you could help with the explanation that would be great!

jessiehuff (approved):
I took a closer look, and I think all of this is actually correct. My bad! 😄   I was still thinking about how we're not using `contenteditable` now, but everything you included I think we still do.

PR #4560, opened 2021-12-06
fix(truncate): fixed fontsize var value

mcoker:
react build is failing because `--pf-c-truncate--FontSize` is referenced but not defined

### DIFF
--- a/src/patternfly/components/Truncate/truncate.scss
+++ b/src/patternfly/components/Truncate/truncate.scss
@@ -1,5 +1,5 @@
 .pf-c-truncate {
-  --pf-c-truncate--FontSize: var(--pf-c-truncate--FontSize, 1rem);
+  --pf-c-truncate--FontSize: 1rem;
 
   display: inline-flex;
   flex-wrap: nowrap;


src/patternfly/components/Truncate/truncate.scss
```diff
@@ -1,5 +1,5 @@
 .pf-c-truncate {
-  --pf-c-truncate--FontSize: var(--pf-c-truncate--FontSize, 1rem);
+  --pf-c-truncate--FontSize: 1rem;
```
mattnolting, line 2: Should the value be `em` instead of `rem`?

mcoker: I don't think so, since `1em` will be `0` here, which doesn't seem like what we're going for?  https://github.com/patternfly/patternfly/blob/dc6cf07d69cbb80026dca7fe43dc28704aabea64/src/patternfly/components/Truncate/truncate.scss#L34-L40

mcoker: Maybe it should be this instead?

```
.pf-c-truncate {
  --pf-c-truncate__text--FontSize: var(--pf-global--FontSize--md);
}

.pf-c-truncate__text {
  font-size: var(--pf-c-truncate__text--FontSize); 
}

evwilkin (approved):
LGTM - tested and this prevented react-tokens build from failing.  Thanks!

Issue #4538, opened 2021-11-23
Audit stylelintrc


mcoker:
We haven't looked at stylelint in a couple of years, and there are a lot of new features in the new version introduced in https://github.com/patternfly/patternfly/pull/4537

PR #4567, opened 2021-12-13
chore(stylelint): cleaned up stylelint rules/sass

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4538

This updates our rules to match the new versions of stylelint and our extends/plugins. A lot of our rules are now defaults in the plugins, so removed those rules. We do have some [invalid @import rules](https://stylelint.io/user-guide/rules/list/no-invalid-position-at-import-rule) that I disabled, but they're in some of our base scss files and I'd like to be careful with those and update them separately to make sure the changes aren't breaking. Created an issue for that here - https://github.com/patternfly/patternfly/issues/4573


srambach (approved):
LGTM 🤷‍♀️

PR #4568, opened 2021-12-14
feat(card): added draggable variation

mcoker:
POC for https://github.com/patternfly/patternfly-design/issues/1106


mcoker:
@mcarrano @mmenestr should we keep this POC PR open or want to close it?

mmenestr:
As in, should we have an example of a card with an indication that it's draggable?

mcoker:
@mmenestr actually I just dug up the original [design issue](https://github.com/patternfly/patternfly-design/issues/1106) and looks like we have an issue to create this feature as part of a [drag and drop demo](https://github.com/patternfly/patternfly/issues/4605). I don't think there is any reason to keep this POC PR around anymore. If we want to use any of the code in this PR to create a draggable card, we can do that in a new PR.

Issue #4564, opened 2021-12-09
Add optional row striping to Table


mcarrano:
Want to add the option to "stripe" or add a background fill to alternate rows in a table. Designs here: https://marvelapp.com/prototype/2ccbf976  We want to use Black-100 to provide a light gray background.

PR #4569, opened 2021-12-15
feat(table): added striped rows

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4564

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -1517,7 +1517,7 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
 
 ### Hoverable and selected example
 ```hbs
-{{#> table table--id="table-expandable-hoverable" table--grid="true" table--modifier="pf-m-grid-lg" table--expandable="true" table--attribute='aria-label="Expandable table example"'}}
+{{#> table table--id="table-hoverable" table--grid="true" table--modifier="pf-m-grid-lg" table--attribute='aria-label="Hoverable and selectable table example"'}}
   {{#> table-thead}}
     {{#> table-tr table-tr--index="thead"}}
       {{> table--check table--check--IsThead="true"}}
@@ -1545,7 +1545,7 @@ Note: To apply padding to `.pf-c-table__expandable-row`, wrap the content in `.p
 
 ### Expandable, hoverable, and selected example
 ```hbs
-{{#> table table--id="table-tbody-expandable-hoverable" table--grid="true" table--modifier="pf-m-grid-lg" table--expandable="true" table--attribute='aria-label="Expandable table example"'}}
+{{#> table table--id="table-expandable-hoverable" table--grid="true" table--modifier="pf-m-grid-lg" table--expandable="true" table--attribute='aria-label="Expandable and hoverable table example"'}}
   {{#> table-thead}}
     {{#> table-tr table-tr--index="thead"}}
       {{> table-td table-td--IsEmpty="true"}}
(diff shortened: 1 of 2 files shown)


mcoker:
@mcarrano here's a POC of all of the tables with row striping. Some tables may not be able to be styled automatically via CSS since what we can do is style "every other" (odd or even) row, or every other `tbody` element's primary row, depending on how the table is constructed. This works best with simple tables with a single `tbody`. Where it doesn't work as well is compound expansion (that example is broken currently), expandable rows (though it seems to work OK), and a user composed table that might look like this:

```
table parent
  tbody
    trow
    trow
    trow
  tbody
    trow
    trow
    trow
  tbody
    trow
    trow
    trow
```

In this scenario we would need to know to style the odd rows in the first tbody, even rows in the second tbody, and odd rows again in the third tbody, which we can't do with just CSS.

For the example above and other custom/non-standard structures needing striping, we could provide different classes to stripe both odd and even rows and leave it to the user to figure out where to use the classes.

We could see if react can try and figure out which rows to apply a class to, and apply a class to each row that gets the striped background color. That's probably a lot of logic though, I'm guessing.

Or we could limit this to use in simple tables with a flat/known row structure, and have a manual striped row class if a product has a more complex table and wants to add striping manually. wdyt?

cc @mattnolting - any other ideas?

mcarrano:
@mcoker I can see where this could get complicated in certain configurations. Maybe we do limit the usage to simple tables.

I did also want to ask about these examples for hoverable/selectable rows: https://patternfly-pr-4569.surge.sh/components/table#hoverable-and-selected It looks like there's a different hover appearance for white vs gray rows or is that my imagination?

Also, are these filled rows using BLACK-100? The look darker than the design mocks.

mcoker:
>  It looks like there's a different hover appearance for white vs gray rows

@mcarrano nothing changed there regarding the hover behavior. When you hover over a row, a shadow appears above/below that row, so that shadow is more prominent when you hover a grey background row (since the shadow appears on the white rows above/below it) compared to hovering a white background row (since the shadow appears on grey rows above/below it, less visible since the shadow is grey)

> are these filled rows using BLACK-100?

Nope! I assumed it was background-color--200 (black-200) - updated it to be black-100.

mcarrano:
Hmmmm... After the last update, I'm not seeing the row striping at all.

mcoker:
@mcarrano oops, had a typo - updated.

mcarrano:
Yep. That looks good now.

mattnolting (commented):
@mcoker @mcarrano I took a stab at some creative and tricky CSS styling based on `tbody` odd/even row combinators, the result was overly complex and unreliable. 

> @mcoker I can see where this could get complicated in certain configurations. Maybe we do limit the usage to simple tables.

Seems to me that you've outlined the best and most reliable path forward above.

mcarrano (commented):
@mcoker Looks good. But just a couple questions/requests.

* I think we agreed that it is OK to restrict this to simple tables. Is it possible to also support tables with simple expansion as in the design mocks?
* Can we include documentation that calls out exactly what table variations support this (and what does not)? I can see someone trying to apply this solution to a complex table and wondering why it does not work.

mcoker:
@mcarrano yep, added a blurb under the main striped section header. I called out what is supported, implying variations not listed are not supported (which keeps us from having to maintain a list as we add variations). Let me know if that's OK or if you'd like me to be more explicit.

Expandable was supported, so added an example of that under the striped section.

mcarrano (approved):
Looks great @mcoker . Thanks!

src/patternfly/components/Table/table.scss
```diff
@@ -303,6 +303,11 @@
   // Subhead
   --pf-c-table__subhead--Color: var(--pf-global--Color--200);
 
+  // Striped
+  --pf-c-table--m-striped--n-child: odd;
+  --pf-c-table--m-striped-even--n-child: even;
```
srambach, line 308: Are these used?

src/patternfly/components/Table/examples/Table.md
```diff
@@ -3636,6 +3633,481 @@ For sticky columns to function correctly, the parent table's width must be contr
...
+      {{#> table-td table-td--data-label="Workspaces"}}
+        5
+      {{/table-td}}
+      {{#> table-td table-td--data-label="Last commit"}}
+        2 days ago
+      {{/table-td}}
+    {{/table-tr}}
+  {{/table-tbody}}
+
+  {{#> table-tbody table-tbody--modifier="pf-m-striped-even"}}
```
srambach, line 3962: Could the text in the cells be modified to be a little more obvious/explanatory that there are rowgroups causing the need for `pf-m-striped-even`?

srambach (commented):
Just a couple of questions. Nice job.

mcarrano (approved):
Looks good to me!

mattnolting (approved):
L🍕 TM!

Issue #4566, opened 2021-12-13
Expandable section - provide an option to indent contents


jgiardino:
With some layouts, having the contents of an expandable section left-aligned with the arrow and not the text makes it hard to convey visual hierarchy and grouping of elements. 

For example, we want to replace section headers with an expandable section component. But our previous layout presents these section headers with a bigger/bolder font than the expandable section. When we replace the section headers with the expandable section, we lose the visual prominence that our previous section headers had. 

If there were an option to indent contents inside an expandable section, this would help to convey the visual hierarchy of the contents on the screen. 

cc @lboehling

lclay2:
Thanks for creating this @jgiardino

PR #4571, opened 2021-12-15
feat(expandable-section): added indented variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4566

### DIFF
--- a/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
+++ b/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
@@ -71,6 +71,18 @@ cssPrefix: pf-c-expandable-section
 {{/stack}}
 ```
 
+
+### Indented
+```hbs
+{{#> expandable-section expandable-section--IsExpanded="true" expandable-section--IsIndented="true"}}
+  {{#> expandable-section-toggle}}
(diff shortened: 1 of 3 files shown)


mcarrano (approved):
This looks good @mcoker . I can definitely see how this is useful. Is this something that requires React follow up (assume yes)? And if so could you open an issue? 

I can try to get this into the January milestone but we are already pretty overloaded. @jgiardino do you have any specific scheduling requirements?

mcoker:
@mcarrano here's the react issue - https://github.com/patternfly/patternfly-react/issues/6705

I can do the react work once this PR is merged

jgiardino:
These updates look great! Thanks for adding this so quickly.

@mcarrano We are hoping to move do usability testing as soon as we're able to get a couple more updates in place. I don't know yet how many sprints are needed to add that. One is adding expandable sections, which would be simple, but our other task might be a little more complicated. So 2 or 3 sprints is my best guess. 

I'm not opposed to using the class name that provides this until it's supported in the react component, but I can follow up with our UI dev team to see if this is not an option.

srambach:
LGTM, and much needed imho :-)

srambach (approved):

PR #4574, opened 2021-12-16
fix(build): make reloading on change to mardown files work again

seanforyou23:
This PR makes the live-reloading work again for markdown files. It loos like the `!(demos)` part of the glob pattern may have been being treated differently in certain environments, so I broke down the directories it looks like that statement was trying to handle in a one liner into multiple statements. Less fancy but more readable/maintainable and less finicky.

[edit] lockfile changes made from simply installing with `yarn install` makes changes to the dependency tree that prevent the test suite and build from working correctly. I've tried this with node@{14,16,17} and all give the same result. I will look for a fix for this in a separate work item, just wanted to note because it prevents me from being able to verify these changes locally (and other basic things like run the build/test suite). Raised https://github.com/patternfly/patternfly/issues/4576 for this


mcoker (approved):

seanforyou23:
Took a bit of poking around to see all the impacts of dismantling references like this, but I think we've gotten to a point where things are working and the changes make sense. I'm still learning how `sourceMD` is used but it seems as long as we send each of these matched sets of example resources to the `html` resource group/directory they get picked up by the rest of the build pipeline. Hoping that's all it takes.. please let me know if you see any other issues!

Issue #4585, opened 2022-01-05
Log viewer updates


mcoker:
* Remove list overflow so horizontal scrollbar shows up at bottom
* Move line number border to list so it doesn't stay fixed and stays between the index and log entry text
* Set `user-select: none;` on the `__index` so it isn't selectable
* Add support for footer and make sure it's light theme - to support https://www.patternfly.org/v4/extensions/log-viewer#with-footer-component

PR #4587, opened 2022-01-05
fix(log-viewer): update scroll, add footer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4585

@DaoDaoNoCode I left the border as a `::before` so all the vars still work as they did before in case someone has modified those.


DaoDaoNoCode (approved):
@mcoker Looks perfect to me, thanks!

PR #4588, opened 2022-01-06
fix(text input group): updated to only apply margin if utilities not empty

### LINKED ISSUE #4514: Text input group - empty utilities element causes extra space
`<TextInputGroupUtilities>` is an exposed component, so it may be empty. This can be seen here - https://patternfly-react-pr-6581.surge.sh/components/text-input-group/react-demos#key-value-filtering

Screen Shot 2021-11-11 at 4 25 50 PM

An empty utilities element shouldn't create extra space in the layout.

evwilkin:
closes #4514 

Updated `.pf-c-text-input-group__utilities` margins to only apply if element is not empty.

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -162,10 +162,12 @@
 .pf-c-text-input-group__utilities {
   display: flex;
   align-items: center;
+  margin-right: var(--pf-c-text-input-group__utilities--MarginRight);
+  margin-left: var(--pf-c-text-input-group__utilities--MarginLeft);
 
-  &:not(:empty) {
-    margin-right: var(--pf-c-text-input-group__utilities--MarginRight);
-    margin-left: var(--pf-c-text-input-group__utilities--MarginLeft);
+  &:empty {
+    --pf-c-text-input-group__utilities--MarginRight: 0;
+    --pf-c-text-input-group__utilities--MarginLeft: 0;
   }
 
   > * + * {
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -162,8 +162,11 @@
 .pf-c-text-input-group__utilities {
   display: flex;
   align-items: center;
-  margin-right: var(--pf-c-text-input-group__utilities--MarginRight);
-  margin-left: var(--pf-c-text-input-group__utilities--MarginLeft);
+
+  &:not(:empty) {
+    margin-right: var(--pf-c-text-input-group__utilities--MarginRight);
+    margin-left: var(--pf-c-text-input-group__utilities--MarginLeft);
+  }
 
   > * + * {
     margin-left: var(--pf-c-text-input-group__utilities--child--MarginLeft);


src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -162,8 +162,11 @@
 .pf-c-text-input-group__utilities {
   display: flex;
   align-items: center;
-  margin-right: var(--pf-c-text-input-group__utilities--MarginRight);
-  margin-left: var(--pf-c-text-input-group__utilities--MarginLeft);
+
+  &:not(:empty) {
```
mcoker, line 166: nice!! 🤩 This is great, but another way to do it that keeps the specificity lower would be to leave the existing declarations, then add `&:empty { --pf-c-text-input-group__utilities--MarginRight/Left: 0; }`. That might be a little easier to work with so we don't have to deal with specificity issues if we add other rules that need to modify the utilities margins.

mcoker (changes requested):

mcoker (approved):
LGTM, though it's worth noting this approach is dependent on the app being able to control whitespace since whitespace in the `__utilities` element will keep `:empty` styles from applying. If the app can't control the whitespace, they can use a comment like

```
<TextInputGroupUtilities><!--

--></TextInputGroupUtilities>
```

I see in the [MDN :empty docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) this note, which looks promising, but doesn't apply currently:

> Note: In Selectors Level 4 the `:empty` pseudo-class was changed to act like `:-moz-only-whitespace`, but no browser currently supports this yet.

@evwilkin @tlabaj do you know if we're able to trim the leading/trailing whitespace from the children passed to `<TextInputGroupUtilities />`?

nicolethoen:
Documenting a slack conversation here:

Evan
hey @Titani & @Austin Sullivan wanted to get your thoughts on Coker’s feedback on TextInputGroupUtilities 
We’re updating to only apply margin around this component if it’s empty (using :empty pseudo class) to avoid it taking up any space when it’s empty
This solution doesn’t work if there’s any whitespace (such as empty string) in the component
I’m unclear if we should look at trimming whitespace in a React follow-up and move forward with this :empty Core update, or if there’s no need to render this component unless there’s something entered in the TextInputGroup (similar to the SearchInput clear icon here) in which case we may not need any changes in Core.
LMK what you think, I’m happy to move forward either way but am a little less familiar with how TextInputGroup && TextInputGroupUtilities are being used and don’t want to miss a use case - thanks!

Austin
Speaking strictly about removing whitespace: IMO having the removal of whitespace for children as a responsibility of TextInputGroupUtilities feels off to me.
If there needs to be no whitespace for it to truly appear empty I feel like that responsibility falls more on the consumer by nature of its composable design.

Evan
That makes sense to me.  In that case, it sounds like a quick solution would just be to cancel out any Core changes and update our example to follow this advice?
```
<TextInputGroupUtilities>
  {showClearButton && (
    <Button variant="plain" onClick={clearInput} aria-label="Clear button and input">
to
{showClearButton && (
  <TextInputGroupUtilities>
    <Button variant="plain" onClick={clearInput} aria-label="Clear button and input">
```

Austin
I was just typing a followup to say that I feel like that’s a better approach :laughing:
I can go ahead and update those examples as part of my text input group issues PR. That’s probably how I should’ve approached it from the outset.

Evan 
Okay thanks for calling that out @mcoker if we update our example to show consumers that the TextInputGroupUtilities should only render if it contains something (and thus should have the padding), do you think it’s okay to close out https://github.com/patternfly/patternfly/issues/4514 - or do you still see a need to make these styling changes to the component?

mcoker 
@Evan Wilkinson awesome! Let’s close it, less bits and bytes in the stylesheet is always a good thing

Evan 
Sounds good to me

Austin Sullivan 
I do think that rather than only conditionally rendering TextInputGroupUtilities the pattern we probably want to demonstrate is to have a separate conditional for the button as well, and maybe an additional button on at least one example.
Reason being that I think if we only demonstrate conditionally rendering TextInputGroupUtilities as a whole I imagine we’ll get issues from people confused about how to implement showing one button but not others if they have multiple utilities.
I could be totally wrong on that though and defer to those of you with more experience on this topic.

^Per the above discussion I believe we are opting for a react solution rather than this core solution? @evwilkin is that all correct?

evwilkin:
@nicolethoen thanks for documenting - that is correct!
- @wise-king-sullyman has replaced the need for this core fix by updating the React example to show how to conditionally render this component [here](https://github.com/patternfly/patternfly-react/pull/6774#issuecomment-1015602363) 
- Confirmed w/@mcoker closing this PR & Core issue out in favor of note above - no code changes needed to core.

PR #4589, opened 2022-01-06
fix(table): fixed sort/fit-content width

### LINKED ISSUE #4535: Nested columns do not provide spacing for sort icon
Nested columns look great except when the sort icon is included after the column header

image

See how icon overlaps with column divider.

mattnolting:
closes #4535 

@mcarrano Looks like this issue applies to all sortable table `th`s, not just nested `th`s. `max-width` should be `none` rather than `fit-content`. Update tested in FF, Chrome and Safari.


mcarrano (approved):
Looks good to me.

srambach:
LGTM 🎉

mcoker (approved):
L🍕TM!

PR #4591, opened 2022-01-06
fix(login page): aligned login box to top

### LINKED ISSUE #4540: Login dialog footer doesn't stay at the bottom when the right hand side is bigger
Having more text on the right hand side of a login dialog, doesn't keep the "footer" in inside the login dialog at the bottom of the component:

image

I would expect that the "main" section expands in this case, and the footer (and band) is still aligned with the bottom.

evwilkin:
Closes #4540 

This PR sets `align-self: start` on `.pf-c-login__main` to avoid it stretching to fill height if the content beside it is taller than it is.

### DIFF
--- a/src/patternfly/components/Login/login.scss
+++ b/src/patternfly/components/Login/login.scss
@@ -165,6 +165,7 @@
 }
 
 .pf-c-login__main {
+  align-self: start;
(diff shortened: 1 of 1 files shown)


mcoker (approved):
⭐️⭐️⭐️⭐️⭐️

PR #4595, opened 2022-01-07
fix(table): move expand all toggle in demo

### LINKED ISSUE #4366: Table demo: move expand all control from toolbar to table header
Per [this design issue](https://github.com/patternfly/patternfly-design/issues/1071) want to update the demo here to locate the "expand all" control to the table header. https://www.patternfly.org/v4/components/table/html-demos#expandable.

Design proposal looks like this:

image

Want to implement Proposal 2.

srambach:
This moves the toggle from the toolbar to the table header in the Table expandable examples and expandable table demo.
Since the toolbar component also had demos showing the toggle in the toolbar over a table, these were removed. Per @mcarrano we might want to add a similar expandable data list demo in the future. However, since it was no longer used anywhere at this time, I removed the toolbar "expand all" handlebars template. This could be left in for future use if desired.

Fixes #4366


mcarrano (approved):
This looks great. Thanks @srambach !

mmenestr:
The expansions look great! Just have one thing I noticed that's unrelated, but I see that in the HTML demos, when there are bulk selects in the table toolbar, there are also bulk selects in the table header. Shouldn't there just be one or the other, since they do the same thing - preferably just the one in the toolbar? See image for what I mean! 

image

srambach:
@mmenestr I also noticed that but I left it because I'd already expanded the scope of the issue 😄 Which do you think it should be? Maybe this should be a new issue?

src/patternfly/components/Table/templates/table--toggle-all.hbs
```diff
@@ -0,0 +1 @@
+{{#if table-tr--expanded}}{{#> table-td table-tr--expanded="true" table-td--toggle="true" table-td--button--attribute=(concat 'id="' table--id '-expandable-toggle-' table-tr--index '" aria-label="Collapse all" ')}}{{/table-td}}{{else}}{{#> table-td table-td--toggle="true" table-td--button--attribute=(concat 'id="' table--id '-expandable-toggle-' table-tr--index '" aria-label="Expand all" ')}}{{/table-td}}{{/if}}
```
mcoker, line 1: Does this need to be on one line or can we make it multi-line so it's easier to read?

src/patternfly/demos/Table/table-collapsed-table.hbs
```diff
@@ -1,7 +1,7 @@
 {{#> table table--id="table-collapsed" table--grid="true" table--modifier="pf-m-grid-lg" table--expandable="true" table--attribute='aria-label="Collapsed table example"'}}
```
mcoker, line 1: looks like you can remove this file. Looks like it's only used in the "Toolbar expand all (collapsed)" toolbar demo that's removed in this PR.

src/patternfly/demos/Table/table-expanded-table.hbs
```diff
@@ -1,7 +1,7 @@
 {{#> table table--id=(concat page--id '-table') table--grid="true" table--modifier="pf-m-grid-xl" table--expandable="true" table--attribute='aria-label="Expanded table example"'}}
```
mcoker, line 1: You should be able to remove this, too. Looks to only be used in the "Toolbar expand all (expanded)" toolbar demo removed in this PR.

mcoker (approved):
Looks fantastic! A few comments for review, none require any changes though.

src/patternfly/components/Table/templates/table--toggle-all.hbs
srambach: It was making an extra line break unless I put it all on one line for some reason. I'd personally prefer readable code and live with an extra line break but I figured I'd get called out on it 😝

mcoker: Yah as long as it doesn't break the HTML or make the example code look bad - like a line break or extra space in the log viewer's text in a `<pre>` tag causes the text in the example to look bad) - I'm in favor of readable code.

Issue #4596, opened 2022-01-07
Dark log viewer button style issues


mcoker:
We have some old "dark theme" button styles in our generic dark theme element/component SASS that is causing issues in the log viewer. 

https://github.com/patternfly/patternfly/blob/125a5a7171512b46599608c1d09aea88fe73d346/src/patternfly/sass-utilities/placeholders.scss#L20-L41

Going to go back to only applying dark theme to the log viewer's `__main` element until we can remove those styles. Also noticed a small position issue with the divider between the line count/index and log entry text.

PR #4597, opened 2022-01-07
fix(log-viewer): updated way dark theme is applied, adjusted border

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4596


DaoDaoNoCode (approved):
Look good to me, thanks! @mcoker

PR #4599, opened 2022-01-10
fix(truncate): resolved spacing issues

mattnolting:
[React issue 6713](https://github.com/patternfly/patternfly-react/pull/6713) raises several questions about the truncate component. There is only one way to truncate with CSS per character, and that is with `text-overflow: ellipsis` or `text-overflow: {string}` (currently only supported in FF). Calculating string widths per non-monospace font character is a heavy JS lift alone, and the truncate component will appear multiple times in tables, data lists, etc those browser calcs will be exponential and lead to performance issues. 

The current implementation takes advantage of `text-overflow: ellipsis` by setting the font size of the duplicate ellipsis to `0` which essentially hides it. `display: flex` doesn't respond to truncated text when it occurs at the beginning of a string and something in the `text-overflow` functionality considers individual character width resulting in: 

> @mattnolting @mcarrano It looks like when the truncation is in the middle, there is sometimes a noticeable large space to the right of the ellipses. On the left, the ellipses are always right next to last character before them. Is this intentional? image

## This update changes the way middle truncation is presented

Instead of cutting a string in half and adding the ellipsis the exact middle of that string, we cut the last 5-10 characters off the string and present them as the second piece of truncated text.

trim

After further contextual testing I learned a couple of things: 
* `display: inline-flex` does not function well in (at least) table and data-list, which makes in an inviable solution. Updating to `display: inline-grid` works very well and is more stable within other components and standard block level elements.
* Passing `font-size` from parent to child is fragile and will lead to unforeseen issues.

With this updated solution we can:
* Mimic the middle truncation functionality by slicing the last 5-10 characters off the string and passing to `pf-c-truncate__end`. This allows `pf-c-truncate__start` to handle truncation alone.
* We no longer need much of the conditional CSS in the current implementation.
* We eliminate or limit (to no more than one space) the gapping that occurs in the current implementation.
* We can remove the `pf-c-truncate__text` wrapper and avoid passing parent text `font-size` to child text. Which eliminates the need for additional JS logic. 

Outline of what will need to be updated in React implementation:
* Remove `pf-c-truncate__text` wrapper.
* For middle truncated text, slice the last 5-10 characters off the string and add to `pf-c-truncate__end`, rather than slicing the string equally into parts `__start` and `__end`. 

Example CSS has also been updated.

### DIFF
--- a/src/patternfly/components/Truncate/examples/Truncate.css
+++ b/src/patternfly/components/Truncate/examples/Truncate.css
@@ -1,0 +1,9 @@
+.pf-c-truncate--example {
+  width: 320px;
+  resize: horizontal;
+  overflow: auto;
+  min-width: 161px;
+  max-width: 100%;
+  padding: var(--pf-global--spacer--md);
+  border: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
+}
--- a/src/patternfly/components/Truncate/examples/Truncate.md
+++ b/src/patternfly/components/Truncate/examples/Truncate.md
@@ -5,6 +5,8 @@ section: components
 cssPrefix: pf-c-truncate
 ---
 
+import './Truncate.css'
+
 ## Examples
 
 ### Notes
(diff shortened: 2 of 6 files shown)


mcarrano (approved):
Looks good @mattnolting . Seems to be a bit better in avoiding the gapping.

mcoker (approved):
Great job! Should we mention in the docs that start truncation doesn't work in safari?

tlabaj (approved):
Looks great!  Thanks for the update.

PR #4600, opened 2022-01-11
fix(table): removed cols/colgroups

### LINKED ISSUE #4584: Remove `colgroup` and `col` references from nested table headers
After further review, `colgroup` and `col` references are unused and offer no semantic value. Their implementation is also convoluted. 

Remove here:
- https://www.patternfly.org/v4/components/table/html#nested-column-headers-and-expandable-rows
- https://www.patternfly.org/v4/components/table/html#nested-column-headers (update example ID)

mattnolting:
Closes #4584 

Removes col/colgroups references from examples


mcoker (approved):
L🤙TM!

PR #4601, opened 2022-01-11
fix(focus): fixed focus styles for interactable elems

### LINKED ISSUE #4534: dropdown: Masthead dropdowns have a residual focus/active line after dismissing the menu with a mouse
When a dropdown is in the masthead, a line should show up when hovering and when the menu is active. When the menu is dismissed, the line should disappear. However, it doesn't. (This should not affect keyboard focus, however.)

Here's a video to show the issue:

https://user-images.githubusercontent.com/10246/142893848-05514356-6a22-4850-b1dc-af08a15c46eb.mp4
(The line is actually blue, not grey. There's a glitch when I recorded it. It's blue on-screen.)

The fix, which I'm adding to Cockpit right now as a workaround, is:

```scss
// In masthead dropdowns, hide line when focus-activated by a mouse
// But still show the line when a keyboard focuses or a mouse hovers or clicks
.pf-c-masthead .pf-c-dropdown:not(.pf-m-expanded) > .pf-c-dropdown__toggle:focus:not(:focus-visible):not(:hover):not(:active)::before {
    border-color: transparent;
}
```

Yes, this has a lot of modifiers. Hopefully the comment helps to walk through. Most importantly, this does not affect keyboard focus.

This would impact https://github.com/patternfly/patternfly/issues/3844, as the style would be altered, but the technique to undo the state would be the same.

mattnolting:
fixes #4534 

This PR changes the default behavior of dropdown, context-selector, menu toggle and select. If element `:focus` + `mouseout`, focus/active/hover styles are reverted to default state.

### DIFF
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -267,8 +267,6 @@
     $namespace: #{$namespace} + ":not(.pf-m-expanded):focus:not(:focus-visible):not(:hover):not(:active)::" + #{$position};
   }
 
-  @debug #{$namespace};
-
   .#{$namespace} {
     @content;
   }
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -254,3 +254,22 @@
     }
   }
 }
+
+@mixin pf-reset-accent-focus-border($namespace, $toggle: $namespace + "__toggle", $root: true, $before: false, $position: "after") {
+  @if $before == true {
+    $position: "before";
+  }
+
+  // construct selector
+  @if $root == false {
+    $namespace: #{$namespace} + ":not(.pf-m-expanded) > ." + #{$toggle} + ":focus:not(:focus-visible):not(:hover):not(:active)::" + #{$position};
+  } @else {
+    $namespace: #{$namespace} + ":not(.pf-m-expanded):focus:not(:focus-visible):not(:hover):not(:active)::" + #{$position};
+  }
+
+  @debug #{$namespace};
+
+  .#{$namespace} {
+    @content;
+  }
+}
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -260,11 +260,13 @@
     $position: "before";
   }
 
+  $focus-root: ":focus:where(:not(:focus-visible):not(:hover):not(:active))::";
+
   // construct selector
   @if $root == false {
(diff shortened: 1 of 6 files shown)


src/patternfly/sass-utilities/mixins.scss
```diff
@@ -254,3 +254,22 @@
...
+  }
+
+  // construct selector
+  @if $root == false {
+    $namespace: #{$namespace} + ":not(.pf-m-expanded) > ." + #{$toggle} + ":focus:not(:focus-visible):not(:hover):not(:active)::" + #{$position};
+  } @else {
+    $namespace: #{$namespace} + ":not(.pf-m-expanded):focus:not(:focus-visible):not(:hover):not(:active)::" + #{$position};
+  }
+
+  @debug #{$namespace};
```
srambach, line 270: Did you mean to leave this here?

mattnolting: Nope, good catch

mcoker (changes requested):
I love this change, my only hesitation is the change in specificity with the `:focus` styles. My inclination is that it's probably fine, but we've had issues with changing selectors/specificity and some apps' overrides no longer applying because they're targeting the override via specific selectors and not CSS vars. And until we have better docs that have been communicated well around how to make customizations using CSS vars, we've communicated that we would avoid changing selectors like this.

I put together a codepen that shows this, and also uses the `:where()` selector to reduce the specificity - https://codepen.io/mcoker/pen/BawbebY. That may work though we may need to use a custom selector per component based on the original focus style specificity. Just an idea.

These focus styles seem safe to update and I wouldn't expect much of a conflict with existing overrides, though these are some pretty common components, and I don't believe this as a high priority issue in this instance. Given that, I'm curious y'alls thoughts.

mattnolting:
@mcoker updated

srambach:
What is a value of having both `--base` and `--reset`? It looks like `--reset` is always the value of `--base` unless there wasn't a `--base` to start with?

mattnolting:
> What is a value of having both `--base` and `--reset`? It looks like `--reset` is always the value of `--base` unless there wasn't a `--base` to start with?

@srambach `--reset` allows for the variable to be accessed contextually, with a less specific selector, like

```
.pf-c-masthead .pf-c-dropdown {
  --pf-c-dropdown__toggle--before--BorderBottomColor--reset: {new-color};
}
```

srambach (approved):
LGTM

mattnolting:
After discussing w/@mcoker, `:focus-visible` at scale in favor of `:focus` will create the desired effect on intractable elements.

PR #4603, opened 2022-01-12
feat(DescriptionList): add icon variant

### LINKED ISSUE #4353: Allow Icon on DescriptionListTerm
I think it would be great if you could do `<DescriptionListTerm icon={<AnIcon />}>...` in the same way you can do `<ListItem icon={<AnIcon />}>` and have it format correctly.  It feels like they have similar use-cases where some visual indicator of the term can be helpful for quick identification.

Screenshot_2021-08-16_15-40-22

That is an example where I've just put an icon in a span next to the `DescriptionListTerm`.  It's OK, but something feels not quite right and I'm not sure what; which is why it would be nice if PF did this automatically :)

Thanks!

srambach:
This adds support for an icon in the term of a description list. 
Note that per design, at the smallest width, the text wraps under the icon, but at the small breakpoint the text changes to be indented past the icon.

Fixes #4353

### DIFF
--- a/src/patternfly/components/DescriptionList/description-list-term.hbs
+++ b/src/patternfly/components/DescriptionList/description-list-term.hbs
@@ -3,7 +3,6 @@
     {{#if description-list-term--attribute}}
       {{{description-list-term--attribute}}}
     {{/if}}>
-
       {{> @partial-block}}
   </dt>
 {{/wrapper}}
--- a/src/patternfly/components/DescriptionList/description-list-term.hbs
+++ b/src/patternfly/components/DescriptionList/description-list-term.hbs
@@ -2,7 +2,6 @@
   {{#if description-list-term--attribute}}
     {{{description-list-term--attribute}}}
   {{/if}}>
-  {{#> description-list-text description-list-text--type="span" description-list-text--IsHelp=description-list-term--TextIsHelp}}
+
     {{> @partial-block}}
-  {{/description-list-text}}
 </dt>
--- a/src/patternfly/components/DescriptionList/description-list-term.hbs
+++ b/src/patternfly/components/DescriptionList/description-list-term.hbs
@@ -1,7 +1,9 @@
-<dt class="pf-c-description-list__term{{#if description-list-term--modifier}} {{description-list-term--modifier}}{{/if}}"
-  {{#if description-list-term--attribute}}
-    {{{description-list-term--attribute}}}
-  {{/if}}>
+{{#> wrapper description-list-text--type="span"}}{{!-- in a term, always make the text a span --}}
+  <dt class="pf-c-description-list__term{{#if description-list-term--modifier}} {{description-list-term--modifier}}{{/if}}"
+    {{#if description-list-term--attribute}}
+      {{{description-list-term--attribute}}}
+    {{/if}}>
 
-    {{> @partial-block}}
-</dt>
+      {{> @partial-block}}
+  </dt>
+{{/wrapper}}
--- a/src/patternfly/components/DescriptionList/description-list.scss
+++ b/src/patternfly/components/DescriptionList/description-list.scss
@@ -19,23 +19,20 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "sm", "md",
   --pf-c-description-list--m-compact--ColumnGap: var(--pf-global--spacer--sm);
 
   // term
+  --pf-c-description-list__term--Display: inline;
+  --pf-c-description-list__term--sm--Display: flex;
   --pf-c-description-list__term--FontWeight: var(--pf-global--FontWeight--bold);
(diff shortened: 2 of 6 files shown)


src/patternfly/components/DescriptionList/description-list.scss
```diff
@@ -23,6 +23,20 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "sm", "md",
...
 
+  // icon
+  --pf-c-description-list__term-icon--MinWidth: var(--pf-global--icon--FontSize--sm);
+  --pf-c-description-list__term-icon--MarginRight: var(--pf-global--spacer--sm);
+  --pf-c-description-list__term-icon--FontSize: var(--pf-global--icon--FontSize--sm);
+  --pf-c-description-list__term-icon--Color: var(--pf-global--icon--Color--light);
+
+  // term text
+  --pf-c-description-list__term-text--Display: inline;
+  --pf-c-description-list__term-text--sm--Display: flex;
```
mcoker, line 34: If someone has passed multiple elements as the term, switching this to a `flex` layout is going to put those elements in a flex row. I think we need to preserve the `__term-text` element's layout to avoid a breaking change since that's what the children of `<DescriptionListTerm />` are currently passed to. 

https://github.com/patternfly/patternfly-react/blob/b30c7c3f64a443451b47ca5ab30a8a13de521bdd/packages/react-core/src/components/DescriptionList/DescriptionListTerm.tsx#L17-L19

2 thoughts on that. Can we move the icon element to be adjacent to the text element? Otherwise, we could add an opt-in class like `.pf-m-icon` to the text element if there is an icon, then create a new element for the text content, though I think the previous approach would be simpler.

mcoker (changes requested):
Looks good! Left a comment about changing an existing element to a flex layout and how that might break people currently.

Also the icons don't seem well aligned with the text.
Screen Shot 2022-01-12 at 2 24 43 PM

FWIW, I know this would differ from the list component, but I think the icons look better if they match the text font-size (14px in this case):
Screen Shot 2022-01-12 at 2 23 27 PM

FWIW, this is the list component if we allowed the icons to match the text size (16px in that case)
Screen Shot 2022-01-12 at 2 27 53 PM

src/patternfly/components/DescriptionList/description-list.scss
srambach: @mceledonia @mcarrano Can you verify what you want for the icon size given the discussion in the issue?

mceledonia: Agreed with Coker, let's go with 14px font size!

src/patternfly/components/DescriptionList/description-list__example.hbs
```diff
@@ -13,23 +15,29 @@
...
+      {{/description-list-text}}
     {{/description-list-term}}
     {{#> description-list-description}}
       example
     {{/description-list-description}}
   {{/description-list-group}}
   {{#> description-list-group}}
     {{#> description-list-term}}
-      Pod selector
+      {{#> description-list-text description-list-text--type="span"}}
```
mcoker, line 38: This works, but something we've done in other templates is wrap all of the contents of the `description-list-text` partial with a call to a partial that doesn't exist (so it doesn't render an element) to always include the context in that template. We use `{{#> wrapper}}` as the partial name. You can see an example of that here - https://github.com/patternfly/patternfly/blob/89fa7ba92f6953b048f1d95a7cced1a46b8a896b/src/patternfly/components/Table/Tree-table/table-tree-view--basic.hbs#L27-L30

Here's what we would add to the `description-list-text` partial
```
{{#> wrapper description-list-text--type="span"}}
  [original hbs contents go here]
{{/wrapper}}
```

Then we can remove `description-list-text--type="span"` (and more importantly don't have to include it in the future) for all uses of `description-list-text`  in `description-list-term`

mcoker (approved):
LGTM! One small nit, but it's fine the way it is, too!

src/patternfly/components/DescriptionList/description-list-term.hbs
```diff
@@ -1,8 +1,9 @@
...
-  {{#> description-list-text description-list-text--type="span" description-list-text--IsHelp=description-list-term--TextIsHelp}}
-    {{> @partial-block}}
-  {{/description-list-text}}
-</dt>
+{{#> wrapper description-list-text--type="span"}}{{!-- in a term, always make the text a span --}}
+  <dt class="pf-c-description-list__term{{#if description-list-term--modifier}} {{description-list-term--modifier}}{{/if}}"
+    {{#if description-list-term--attribute}}
+      {{{description-list-term--attribute}}}
+    {{/if}}>
+
```
mcoker, line 6: nit - blank line here

mcoker (approved):
LGTM!!! One super small nit.

mcarrano (approved):
Looks good to me. Thanks @srambach !

Issue #4512, opened 2021-11-11
JumpLinks or Drawer?: build demo for a page with jumplinks and a drawer


nicolethoen:
requested by @jgiardino and @srambach

nicolethoen:
https://codesandbox.io/s/objective-visvesvaraya-qso4q?file=/index.js

PR #4608, opened 2022-01-13
chore(drawer): added demos with jumplinks

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4512

drawer collapsed - https://patternfly-pr-4608.surge.sh/components/drawer/html-demos/drawer-with-jump-links
drawer expanded - https://patternfly-pr-4608.surge.sh/components/drawer/html-demos/expanded-drawer-with-jump-links

### DIFF
--- a/src/patternfly/demos/Drawer/examples/Drawer.md
+++ b/src/patternfly/demos/Drawer/examples/Drawer.md
@@ -46,5 +46,15 @@ wrapperTag: div
 {{/inline}}
 ```
 
+### Drawer with jump links
+```hbs isFullscreen
+{{> drawer-jump-links drawer-jump-links--id="drawer-jump-links"}}
+```
+
+### Expanded drawer with jump links
+```hbs isFullscreen
+{{> drawer-jump-links drawer-jump-links--id="drawer-expanded-jump-links" drawer-panel--IsOpen="true"}}
+```
+
 ## Documentation
 This demo implements the drawer in context of the page component.
--- a/src/patternfly/demos/Drawer/examples/drawer-jump-links.hbs
+++ b/src/patternfly/demos/Drawer/examples/drawer-jump-links.hbs
@@ -1,0 +1,28 @@
+{{> page-template
+      page-template--id=drawer-jump-links--id
+}}
+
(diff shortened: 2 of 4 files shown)


mattnolting (commented):
Implementation is 👍 . I see what you're going for in the example structure, it's cool. However, in this example the code is repeated and would need to be repeated in further examples that could share this content. I think it would more straightforward to put the custom code in it's own partial and `*inline` it like you have here or just wrap with `drawer-demo-default` and set the custom flag on that template.

srambach:
The jumplinks aren't sticky here. I think the goal was to show how jumplinks can work while also having a drawer on the page. Does this need to be a React demo instead?

mcarrano:
What's the status of this one @mcoker ? Looks like it's been sitting idle for a while.

mcoker:
@mcarrano I'm going to try and address this milestone.

wise-king-sullyman:
Will this PR be able to make it in this milestone? It is a blocker for https://github.com/patternfly/patternfly-react/issues/6582 which I'm slated to work on next milestone.

mcoker:
@wise-king-sullyman I don't think so, it's lower priority so it's likely going to move into the next milestone.

mcarrano (commented):
Looks OK to me. But I will pass as I'm not exactly sure what we were intending to demo here.

srambach:
Section ids are `-first`, `-second` etc - is that intended or should there be a prefix?

Since the jumplinks highlighting (showing the current section) does not work without JS, there should be text along with the demo explaining this - otherwise it just looks like our demo is broken.

`pf-c-drawer__content` is the scrollable container here that needs to be used as the scrollable selector by the scrollspy. It would be better if there were a unique id on this element because if there is more than one `pf-c-drawer__content` for some reason, the scrollspy will not work properly.

Finally, I would like to see a link to this demo from the jumplinks demo page. I don't see a need to duplicate it, but it's quite likely that someone trying to make this work would be looking at the jumplinks component, not at the drawer component.

mcoker:
@srambach updated!

srambach (approved):
🎉 Great!

PR #4616, opened 2022-01-17
feat(Text input group): added autocomplete ghosting

### LINKED ISSUE #4558: Add autocomplete ghosting feature to TextInputGroup
similar to what is done in SearchInput

evwilkin:
Closes #4558 

This PR adds a new example for text input group using the existing `--pf-c-text-input-group__text-input--m-hint--Color` variable to show ghosting for autocomplete suggestion.

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
@@ -6,16 +6,16 @@
   type="text"
   {{#if text-input-group--IsDisabled}}
     value="Disabled"
+    disabled
+    aria-label="Disabled text input group example input"
   {{else if text-input-group-text-input--hint-text}}
     value="{{text-input-group-text-input--hint-text}}"
   {{else}}
     value="{{text-input-group--value}}"
   {{/if}}
-  aria-label="Type to filter"
-  {{#if text-input-group--IsDisabled}}
-    disabled
-    aria-label="Disabled text input group example input"
-  {{/if}}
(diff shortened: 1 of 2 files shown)


mattnolting (approved):
L🍕 TM

src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs
```diff
@@ -1,11 +1,25 @@
...
   type="text"
-  value="{{#if text-input-group--IsDisabled}}Disabled{{else}}{{text-input-group--value}}{{/if}}"
+  {{#if text-input-group--IsDisabled}}
+    value="Disabled"
+  {{else if text-input-group-text-input--hint-text}}
+    value="{{text-input-group-text-input--hint-text}}"
+  {{else}}
+    value="{{text-input-group--value}}"
+  {{/if}}
   aria-label="Type to filter"
```
mcoker, line 14: Can we omit this if it's a hint input? It won't do anything if it's `aria-hidden`. Looks like the react component is already doing that in the search input. We'll need to update the search input, too - we can do that here or in a separate issue if you'd prefer.

```suggestion
  {{#if text-input-group-text-input--hint-text }}
    aria-label="Type to filter"
  {{/unless}}
```

mcoker (approved):
Fabulous! A couple of nits, neither are blockers if we want to follow up later.

evwilkin:
@mcoker made your suggested changes - will need a separate follow up for SearchInput, I took a look but wasn't quite sure what needed to change.

PR #4617, opened 2022-01-18
feat(masthead): added horizontal nav demo

### LINKED ISSUE #4522: Add nav to masthead example/demo
> I’ve tested horizontal nav within masthead. It works as expected. If I add a toolbar to  pf-c-masthead__content, the nav shrinks and the toolbar takes up most of the space, though.
Would be great to have an example / snippet on patternfly.org which shows a masthead with toggle, brand, horizontal nav and toolbar! (

mattnolting:
closes #4522 

@mcarrano @mceledonia can you please review [Masthead with horizontal nav](https://patternfly-pr-4617.surge.sh/components/masthead/html-demos#horizontal-nav)? A couple of things in question, bearing in mind that space becomes much more limited in this iteration. 

* Do we want to show this demo with context selector included? It has to live in the same container as `__brand` to be accommodated. 
* Should the nav be full height?
* Any design details that I've missed...

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.hbs
+++ b/src/patternfly/components/Dropdown/dropdown.hbs
@@ -1,6 +1,6 @@
 <div class="pf-c-dropdown{{#if dropdown--IsExpanded}} pf-m-expanded{{/if}}{{#if dropdown--modifier}} {{dropdown--modifier}}{{/if}}"
   {{#if dropdown--attribute}}
-    {{dropdown--attribute}}
+    {{{dropdown--attribute}}}
   {{/if}}>
 
   {{!-- if content is passed --}}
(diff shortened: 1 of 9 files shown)


mattnolting:
Designs:
https://marvelapp.com/prototype/5cjajdg/screen/63256957
https://marvelapp.com/prototype/4d9chi8/screen/67194841
https://marvelapp.com/prototype/5cjajdg/screen/63256945

mcarrano (changes requested):
@mattnolting Yes, there are a few things that should change here:

* We wouldn't recommend using this together with a vertical nav, so I think you should take that out.
* I'd be in favor of removing the context selector here. I don't think that we'd recommend having both that and a horizontal menu live together in the masthead.
* I believe that our design intent was to have the menu fill the full height of the masthead (I'm guessing that's what you mean by "full height).

@mceledonia do you agree with the above? Am I missing anything?  The bottom line is that we rarely use a horizontal primary navigation and would only recommend it in a very simple use case which is why you don't see it often in product.

mcarrano (approved):
This looks great. Thanks @mattnolting !

srambach (approved):
LGTM

mcoker (approved):
L🔥TM!

PR #4636, opened 2022-01-21
fix(ProgressStepper): remove extra space below

### LINKED ISSUE #4633: progress stepper includes extra margin at the bottom
In the [vertical variation of the progress stepper](https://www.patternfly.org/v4/components/progress-stepper#vertical-with-descriptions), margin-bottom is applied to all steps, including the last one. This results in the progress stepper having extra space at the bottom, which isn't expected. Usually, PF components that define space between items will [not include this space after the last item](https://github.com/patternfly/patternfly/search?q=%3Alast-child).

Can the progress stepper be updated to not include the bottom margin after the text of the last step? 

image

cc @cfullam1

srambach:
Fixes #4633 by explicitly setting a 0 margin after the last step.

https://patternfly-pr-4636.surge.sh/components/progress-stepper#vertical-with-descriptions

### DIFF
--- a/src/patternfly/components/ProgressStepper/progress-stepper.scss
+++ b/src/patternfly/components/ProgressStepper/progress-stepper.scss
@@ -207,8 +207,8 @@
     --pf-c-progress-stepper__step-icon--Color: var(--pf-global--info-color--100);
   }
 
-  &:last-child .pf-c-progress-stepper__step-main {
-    margin-bottom: 0;
+  &:last-child {
+    --pf-c-progress-stepper__step-main--MarginBottom: 0;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/patternfly/components/ProgressStepper/progress-stepper.scss
```diff
@@ -206,6 +207,10 @@
   &.pf-m-info {
     --pf-c-progress-stepper__step-icon--Color: var(--pf-global--info-color--100);
   }
+
+  &:last-child .pf-c-progress-stepper__step-main {
```
mcoker, line 211: Could this also be written this way with less specificity?

```
.pf-c-progress-stepper__step:last-child {
    --pf-c-progress-stepper__step-main--MarginBottom: var(--pf-c-progress-stepper__step--last-child__step-main--MarginBottom);
}
```

We also often don't create a var for something like this that we wouldn't expect to be customized, so your declaration could also be `--pf-c-progress-stepper__step-main--MarginBottom: 0;`, and remove the var. But the way you've done it is fine, too.

srambach: I removed the variable, but you can't remove specificity by removing step-main because then you aren't targeting what is causing the extra space.

mcoker: Would this not work? https://codesandbox.io/s/dank-morning-bc5r4?file=/style.css

srambach: Ah, yes, if you set the variable instead of margin-bottom directly. 🤪

mcoker (approved):
L🥳TM!

Issue #4638, opened 2022-01-24
compact labels have wrong height inside a label group


eliranmal:
### description

when putting `Label` components inside a `LabelGroup`, and marking only some of them - **but not all** - with the compact modifier, their layout is not affected by the compact state, and the label's box remains at full height.

this issue is opened to handle a higher-level issue ([patternfly/react #6815][1]), which originates in the underlying CSS rules.

### steps to reproduce

1. create a label group, and put some labels inside
1. assign the `pf-m-compact` modifier to one of the labels

### expected result

the modified label should have a compact layout, with lower height.

### actual result

the label remains at its full height (but its text does get smaller).

[1]: https://github.com/patternfly/patternfly-react/issues/6815

eliranmal:
@nicolethoen / @evwilkin - can you please assign me to this issue?

i opened it following another issue on the react repo, thinking it should be handled here.
i already have a working fix, just want to tie it to this issue and open a PR.

eliranmal:
thanx!

eliranmal:
#### discovery

- any label group item is a flexbox, and has default alignment for its items.
- as result, nested labels are aligned with the initial rule: `align-items: normal`, which [behaves like `stretch` for flex items][1].

#### suggested solution

apply `align-items: center` to label group items as a sensible default, and declare it as a new component-level property, to allow overrides (e.g. `--pf-c-label-group__list-item--AlignItems`).


[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/align-items

mcarrano:
@eliranmal see my comment on https://github.com/patternfly/patternfly-react/issues/6815. I think this was based on a misunderstanding of how the label group was intended to work and should be closed, I think. Do you agree>

eliranmal:
@mcarrano, yes, indeed the original issue was based on a misunderstanding (or misuse).  
i think it may still be a good idea to fix this (for fool-proofing?), but i understand that i can't anticipate many of the breaking changes this change may introduce, and there are other concerns that are not met by it.

i can simply omit the CSS changes, and the mixed-labels example from the docs, and leave in just the new compact label examples.

eliranmal:
*note:* this issue was not actually fixed (but resolved by updating the docs), so maybe it should be labeled with a `wontfix` (?)

PR #4639, opened 2022-01-24
fix(label-group): added compact examples

eliranmal:
#### content

- in label group items, explicitly align flex items to center to avoid stretching their height
- add examples for compact labels variants to the docs

###### off topic

- fix headings in label group docs

#### notes

- this fix applies to compact labels, when _they are mixed_ with non-compact labels in the same group
- the impl' assigns `center` as a hard-coded value for alignment, as i couldn't find any corresponding global property (and refrained from using utils) - please _let me know_ if i missed something.

fixes: #4638


src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -554,7 +554,148 @@ In addition to the JavaScript management of [editable labels](/components/label#
 {{/label-group}}
 ```
 
-### Accessibility
+### Compact labels
+```hbs
+{{#> label-group label-group--id="label-group-basic"}}
```
mcoker, line 559:
```suggestion
{{#> label-group label-group--id="label-group-compact"}}
```

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -554,7 +554,148 @@ In addition to the JavaScript management of [editable labels](/components/label#
...
+        {{/label}}
+      {{/label-group-list-item}}
+    {{/label-group-list}}
+  {{/label-group-main}}
+{{/label-group}}
+```
+
+### Mixed labels (compact / default)
+```hbs
+{{#> label-group label-group--id="label-group-basic"}}
```
mcoker, line 593:
```suggestion
{{#> label-group label-group--id="label-group-mixed-compact-default"}}
```

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -554,7 +554,148 @@ In addition to the JavaScript management of [editable labels](/components/label#
...
+        {{/label}}
+      {{/label-group-list-item}}
+    {{/label-group-list}}
+  {{/label-group-main}}
+{{/label-group}}
+```
+
+### Compact labels, overflow
+```hbs
+{{#> label-group label-group--id="label-group-overflow"}}
```
mcoker, line 627:
```suggestion
{{#> label-group label-group--id="label-group-compact-overflow"}}
```

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -554,7 +554,148 @@ In addition to the JavaScript management of [editable labels](/components/label#
...
+        {{/label}}
+      {{/label-group-list-item}}
+    {{/label-group-list}}
+  {{/label-group-main}}
+{{/label-group}}
+```
+
+### Compact labels, vertical
+```hbs
+{{#> label-group label-group--id="label-group-vertical" label-group--modifier="pf-m-vertical"}}
```
mcoker, line 666:
```suggestion
{{#> label-group label-group--id="label-group-compact-vertical" label-group--modifier="pf-m-vertical"}}
```

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -574,7 +715,7 @@ In addition to the JavaScript management of [editable labels](/components/label#
 | `.pf-c-label-group__textarea` | `<textarea>` | Initiates the textarea element in the label group. **Required when label group is editable** |
 | `.pf-c-label-group__label` | `<span>` | Initiates the label to be used in the label group. |
 | `.pf-c-label-group__close` | `<div>` | Initiates the container used for the button to remove the label group. |
-| `.pf-m-editable` | `.pf-c-label-group` | Modifies the label group to support editable styling. |
 | `.pf-c-button` | `.pf-c-label-group__close <button>` | Initiates the button used to remove the label group. |
+| `.pf-m-editable` | `.pf-c-label-group` | Modifies the label group to support editable styling. |
```
mcoker, line 719: 👍👍

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -554,7 +554,148 @@ In addition to the JavaScript management of [editable labels](/components/label#
...
+          {{/label-icon}}
+          Label 3
+        {{/label}}
+      {{/label-group-list-item}}
+    {{/label-group-list}}
+  {{/label-group-main}}
+{{/label-group}}
+```
+
+## Accessibility
```
mcoker, line 698: Nice catch! We're actually missing a `## Documentation` heading here, and the existing a11y/usage headings should remain `h3`'s - do you mind updating that? Here's an example - https://www.patternfly.org/v4/components/alert/html -- https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Alert/examples/Alert.md

mcoker (changes requested):
@eliranmal thanks! @mcarrano @mmenestr wdyt of this mixed compact/regular example? https://patternfly-pr-4639.surge.sh/components/label-group/#mixed-labels-compact--default

We haven't considered that with the current code. Is that something we want to support, and assuming so should the labels be center aligned like they are in the PR? FWIW here's how it wraps - 

Screen Shot 2022-01-24 at 5 12 19 PM

And this is what it looks like now if you mix compact and regular labels in a label group

Screen Shot 2022-01-24 at 5 15 01 PM

Screen Shot 2022-01-24 at 5 14 54 PM

eliranmal:
> We haven't considered that with the current code. Is that something we want to support, and assuming so should the labels be center aligned like they are in the PR?

my decision to have it centered was quite arbitrary. we could also decide, of course, that the default is aligned to the bottom (technically, `flex-end`).

src/patternfly/components/LabelGroup/examples/LabelGroup.md
eliranmal: oh yes, that makes much sense. i'll update it to match

src/patternfly/components/LabelGroup/examples/LabelGroup.md
eliranmal: thanx, good catch with these ids!

eliranmal:
if we're concerned about backwards compatibility (even though the current situation looks a bit off), we could condition these rules with a component scope property (so `.pf-c-label-group.pf-m-compact`).

this can eventually be tied to the the PF/react impl' to make it more lightweight /  loosely coupled.

i'm assuming we can do this safely as the component is still in beta.

mcoker:
> if we're concerned about backwards compatibility (even though the current situation looks a bit off), we could condition these rules to be mutually inclusive with a component-scope property (so only apply it when .pf-c-label-group.pf-m-compact).

It should be fine, the way it looks now definitely seems like a bug.

mcarrano:
@eliranmal @mcoker It may be that this PR is fine if it improves the layout of mixed labels. However, I don't think that mixing label types (regular and compact) in the same group is something we would recommend. @eliranmal is there a use case that is coming up in product where this is necessary? In general, the compact labels where only intended to be used in places where space might be limited, like in a table cell or similar place.

eliranmal:
thanx for the input @mcarrano!

it looks like there is demand for this use case (see [the issue that originated this change][1], up at the patternfly-react repo), where the user reported that they're trying to use mixed labels (i'm not sure about the context, or why are they doing that).



[1]: https://github.com/patternfly/patternfly-react/issues/6815

mcarrano:
@eliranmal Thanks for forwarding me that link. It was very helpful.

My reading of the original issue is not that the desire was to mix label types, but rather that the author expected that applying the `isCompact` prop would force all labels nested in the group to be forced to compact, which it does not do. I'm not clear on what exactly that prop is for and whether this is a valid bug or if the purpose of this prop needs to be communicated better. @tlabaj @nicolethoen can you explain more about what the intent here is in the React component?

eliranmal:
> My reading of the original issue is not that the desire was to mix label types, but rather that the author expected that applying the `isCompact` prop would force all labels nested in the group to be forced to compact, which it does not do.

indeed, it seems that this is the case.

> I'm not clear on what exactly that prop is for and whether this is a valid bug or if the purpose of this prop needs to be communicated better.

looking at the source, this prop seem to be there for the sole purpose of making the overflow button compact (see my comment on that, there, suggesting it should be renamed to `isOverflowButtonCompact`, to clarify its intent). at the moment, it surely does not behave like other component-level props. 
however, this group-level prop may be used in the future as the base for a broader feature-set.




[1]: https://github.com/patternfly/patternfly-react/issues/6815#issuecomment-1019327848

mcarrano:
@tlabaj @nicolethoen Thoughts?

eliranmal:
thanx @mcarrano - i also asked the original poster about the intended behavior from their POV ([here][1]).


https://github.com/patternfly/patternfly-react/issues/6815#issuecomment-1022408057

mcoker:
Hey @eliranmal! re: https://github.com/patternfly/patternfly/issues/4638#issuecomment-1027440947

> i think it may still be a good idea to fix this (for fool-proofing?), but i understand that i can't anticipate many of the breaking changes this change may introduce, and there are other concerns that are not met by it.

Agree on both fronts.

> i can simply omit the CSS changes, and the mixed-labels example from the docs, and leave in just the new compact label examples.

This would be perfect! If you can do it by EOD Thursday, we'll get it in this upcoming release! 💪

eliranmal:
thanks @mcoker!

saw this comment too late, so next release it is :shipit:

eliranmal:
i made the discussed changes.

as there was no change to behavior (only documentation), should we mark the original issue as `wontfix`?

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -554,6 +554,115 @@ In addition to the JavaScript management of [editable labels](/components/label#
...
+          {{/label-icon}}
+          Label 3
+        {{/label}}
+      {{/label-group-list-item}}
+    {{/label-group-list}}
+  {{/label-group-main}}
+{{/label-group}}
+```
+
+## Documentation
```
mcoker, line 664: 👍👍

mcoker (approved):
Looks great, thanks!!!

PR #4640, opened 2022-01-25
fix(TableDemo): remove bulk selector from header

### LINKED ISSUE #4598: Remove duplicate bulk select in table demos
Some of the full page table demos have 2 bulk selects instead of just 1 (one in the table toolbar, and one in the table header).

I propose that we remove the bulk selects in the table header, in favor of the ones in the table toolbar.

These are the demos that would need to be changed:
• https://www.patternfly.org/v4/components/table/html-demos#basic
• https://www.patternfly.org/v4/components/table/html-demos#sortable
• https://www.patternfly.org/v4/components/table/html-demos#sortable
• https://www.patternfly.org/v4/components/table/html-demos#compact
• https://www.patternfly.org/v4/components/table/html-demos#static-bottom-pagination
• https://www.patternfly.org/v4/components/table/html-demos#column-management-modal
• https://www.patternfly.org/v4/components/table/html-demos#sticky-header

cc: @mcarrano @srambach

srambach:
Fixes #4598 

This also replaces a couple of empty THs with TDs to fix an accessibility error (empty THs are not allowed).

https://patternfly-pr-4640.surge.sh/components/table/html-demos#basic
Bulk selector removed from the header of  the basic, sortable, expandable, compact, static bottom pagination, column management modal, and sticky header demos.
The data list example just changes the empty TH to a TD for a11y, but I believe should still have the bulk selector in the table header since it's a selector for the secondary table. https://patternfly-pr-4640.surge.sh/components/data-list/html-demos#expandable-demo

### DIFF
--- a/src/patternfly/demos/DataList/data-list-expandable-compact-table.hbs
+++ b/src/patternfly/demos/DataList/data-list-expandable-compact-table.hbs
@@ -22,8 +22,8 @@
       {{#> table-th table-th--attribute='scope="col"' table-th--icon="true"}}
         Icons
       {{/table-th}}
-      {{#> table-th}}{{/table-th}}
-      {{#> table-th}}{{/table-th}}
+      {{> table-td table-td--IsEmpty="true"}}
+      {{> table-td table-td--IsEmpty="true"}}
     {{/table-tr}}
   {{/table-thead}}
   {{#> table-tbody}}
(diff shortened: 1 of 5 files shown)


mmenestr:
this all LGTM!

mmenestr (approved):

mcoker (approved):
🎉

Issue #4224, opened 2021-07-15
Remove up and down arrows for pagination direct input box


vjwilson:
In reference to this component, https://www.patternfly.org/v4/components/pagination

<img width="767" alt="Screen Shot 2021-07-15 at 11 19 29 AM" src="https://user-images.githubusercontent.com/715729/125813306-9698ef50-3754-4a86-b2b9-38e51adfa670.png">

Those up/down arrow buttons in the <input type="number"...> type are hard to see, hard to click, less accessible, and cause more confusion than they are worth.

I think we should remove them from design, docs, and implementations like partternfly-react.

For an example of why those arrows should be removed, see this article from the Gov.UK team:
https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

mcoker:
cc @mcarrano @jessiehuff wdyt?

mcarrano:
My understanding is that those spin buttons are imposed by the browser because this is a number input. Is that right? @jessiehuff is there an a11y reason to have them there? I don't feel like they really add value from a usability perspective since the user can always use the "<" or ">" to go forward or back a page. I'd be OK to remove them as long as it doesn't have other bad side effects to do so.

mcoker:
> My understanding is that those spin buttons are imposed by the browser because this is a number input. Is that right?

@mcarrano yep. We removed them in the [number input](https://www.patternfly.org/v4/components/number-input) because we added our own increment buttons. You could make the same argument to remove them in pagination since we have the  "<" and ">" buttons like you mentioned.

mcarrano:
@jessiehuff any reason not to remove these?

mcarrano:
@jessiehuff ^^^

jessiehuff:
I think it'd be fine to just remove the visual arrows similar to the [Number Input component](https://www.patternfly.org/v4/components/number-input/). I can't think of any downsides to doing so. 👍

vjwilson:
@mcoker @evwilkin Would one of your remove the bot's *wontfix* label and keep this issue open?

mcarrano:
Sorry this got lost in the shuffle. Just added to our roadmap for 2022.

PR #4641, opened 2022-01-25
feat(pagination): remove number input arrow visibility

jeffpuzzo:
Closes: #4224

To test, navigate to the Pagination examples and verify when focusing on any page select input box that up/down arrows are not visible and you are still able to use keyboard up/down arrows to increment/decrement the count.

### DIFF
--- a/src/patternfly/components/Pagination/pagination.scss
+++ b/src/patternfly/components/Pagination/pagination.scss
@@ -37,7 +37,7 @@ $pf-c-pagination--breakpoint-map: build-breakpoint-map();
   --pf-c-pagination__nav-page-select--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--child--MarginRight: var(--pf-global--spacer--xs);
-  --pf-c-pagination__nav-page-select--c-form-control--width-base: calc(var(--pf-global--spacer--sm) * 2);
+  --pf-c-pagination__nav-page-select--c-form-control--width-base: calc(var(--pf-global--spacer--sm) * 2 + var(--pf-global--BorderWidth--sm) * 2);
   --pf-c-pagination__nav-page-select--c-form-control--width-chars: 2;
   --pf-c-pagination__nav-page-select--c-form-control--Width: calc(var(--pf-c-pagination__nav-page-select--c-form-control--width-base) + (var(--pf-c-pagination__nav-page-select--c-form-control--width-chars) * 1ch));
 
--- a/src/patternfly/components/Pagination/pagination.scss
+++ b/src/patternfly/components/Pagination/pagination.scss
@@ -37,7 +37,7 @@ $pf-c-pagination--breakpoint-map: build-breakpoint-map();
   --pf-c-pagination__nav-page-select--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--PaddingRight: var(--pf-global--spacer--md);
   --pf-c-pagination__nav-page-select--child--MarginRight: var(--pf-global--spacer--xs);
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Pagination/pagination.scss
```diff
@@ -255,6 +255,19 @@ $pf-c-pagination--breakpoint-map: build-breakpoint-map();
     }
   }
 
+  // hide number input up/down arrow buttons
+  input[type="number"] {
```
mcoker, line 259: Can you test adding these styles under the `.pf-c-form-control` selector below? That selector should target this element without the need to define `input[...]`

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -255,6 +255,19 @@ $pf-c-pagination--breakpoint-map: build-breakpoint-map();
...
+    // stylelint-disable
+    &::-webkit-inner-spin-button,
+    &::-webkit-outer-spin-button {
+    // stylelint-enable
+      appearance: none;
```
mcoker, line 264: This is great! We do this in number input using an older technique. Any interest in updating that as part of this PR, too? If not we can make that a separate issue:

https://github.com/patternfly/patternfly/blob/d57498cdc0946095da6f605e95067df610f52787/src/patternfly/components/NumberInput/number-input.scss#L21-L29

mcoker (changes requested):
This looks really good! Left a couple of comments.

Also one more update needed - we calculate the width of that input with 2 variables:
* `--pf-c-pagination__nav-page-select--c-form-control--width-chars` - the number of characters the input should show.
* `--pf-c-pagination__nav-page-select--c-form-control--width-base` - the base width of the input including the estimated width of the arrows that were removed in this PR.

Now that the arrows are gone, we should update the calculation that creates space for them in the width of that input. Here's the block where those vars are defined: 

https://github.com/patternfly/patternfly/blob/d57498cdc0946095da6f605e95067df610f52787/src/patternfly/components/Pagination/pagination.scss#L40-L42

FWIW we also do this in the number input, which may be good as a reference:

https://github.com/patternfly/patternfly/blob/d57498cdc0946095da6f605e95067df610f52787/src/patternfly/components/NumberInput/number-input.scss#L8-L11

mcoker (commented):
This looks great, appreciate the DRY approach, and it matches the number input. Though I am noticing that there isn't always space for the number of characters to show via the `--width-chars` in both components. For example, "20" doesn't show correctly:

Screen Shot 2022-01-26 at 1 54 28 PM

Screen Shot 2022-01-26 at 1 54 32 PM

Screen Shot 2022-01-26 at 1 54 23 PM

Screen Shot 2022-01-26 at 1 54 19 PM

Seems like we should look into that a bit and make a small adjustment - do you agree @mcarrano @mmenestr?

Also this is going to change the visual width of the input, and if a user was either depending on the arrows and/or customized the input width, this could cause a visual issue for them. Just want to verify that's OK @mcarrano @mmenestr

mcarrano:
@mcoker Interestingly, I don't see the problem in Firefox, but I do in Chrome. I definitely think we need to address this, although I'm wary of adding more width since we could have cases where a pagination control is used in a toolbar with tight spacing. But is it true that overall we are reducing the with of the input by this PR? In that case maybe add some width back?

jeffpuzzo:
>

I will say that I didn't recognize this in the react example equivalent for pagination because it seems we're automatically setting the number count to what the max is, so entering 20 out of 4 wasn't an option there, so it was definitely overlooked.

Thanks to the help of @mcoker who pointed out the input border was causing this issue, and for providing the calculation itself, I think these should be all set now:
image
image

mcoker (approved):
Awesome work @jeffpuzzo!! So much better!

mcarrano (approved):
Looks great @jeffpuzzo !

PR #4642, opened 2022-01-26
feat(table): updated tree table checkbox padding

### LINKED ISSUE #4554: Update checkbox spacing in tree table
https://patternfly-pr-4546.surge.sh/components/table/#tree-table-with-checkboxes

@mceledonia @mcarrano to increase clickable area of checkboxes in tree tables, spacing would need to be added to layout.

Screen Shot 2021-12-02 at 11 03 23 AM

mattnolting:
closes #4554

### DIFF
--- a/src/patternfly/components/Table/table-tree-view.scss
+++ b/src/patternfly/components/Table/table-tree-view.scss
@@ -9,8 +9,8 @@ $pf-c-tree-view--MaxDepth: 10;
   --pf-c-table__tree-view-main--MarginLeft: calc(var(--pf-c-table--cell--PaddingLeft) * -1);
 
   // check
+  --pf-c-table__tree-view-main--c-table__check--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-table__tree-view-main--c-table__check--PaddingLeft: var(--pf-global--spacer--sm);
-  --pf-c-table__tree-view-main--c-table__check--MarginRight: var(--pf-global--spacer--md); // update at breaking change
 
   // icon
   --pf-c-table__tree-view-icon--MinWidth: var(--pf-global--FontSize--md);
@@ -60,13 +60,12 @@ $pf-c-tree-view--MaxDepth: 10;
   }
 
   > .pf-c-table__check {
-    align-self: stretch;
-    margin: calc(var(--pf-c-table--cell--PaddingTop) * -1) 0 calc(var(--pf-c-table--cell--PaddingBottom) * -1) 0;
+    margin-right: var(--pf-c-table__tree-view-main--c-table__check--MarginRight);
   }
 
   > .pf-c-table__check label {
-    height: 100%;
-    padding: var(--pf-c-table--cell--PaddingTop) var(--pf-c-table__tree-view-main--c-table__check--MarginRight) var(--pf-c-table--cell--PaddingBottom) var(--pf-c-table__tree-view-main--c-table__check--PaddingLeft);
+    padding-right: var(--pf-c-table__tree-view-main--c-table__check--PaddingRight);
+    padding-left: var(--pf-c-table__tree-view-main--c-table__check--PaddingLeft);
     margin: 0;
   }
 }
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -59,7 +60,14 @@ $pf-c-tree-view--MaxDepth: 10;
   }
 
   > .pf-c-table__check {
-    margin-right: var(--pf-c-table__tree-view-main--c-table__check--MarginRight);
```
mcoker, line 62: I think this will break core users who don't get the markup change. We should probably keep all the positioning stuff limited to the `<label>`, which is basically our opt-in for the change.

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -59,7 +60,14 @@ $pf-c-tree-view--MaxDepth: 10;
...
+    align-self: stretch;
+    margin: calc(var(--pf-c-table--cell--PaddingTop) * -1) 0 calc(var(--pf-c-table--cell--PaddingBottom) * -1) 0;
+  }
+
+  > .pf-c-table__check label {
+    height: 100%;
```
mcoker, line 68: This could be a little weird that the target area is so tall, and doesn't match the expandable toggle:

Screen Shot 2022-01-26 at 2 28 51 PM

expandable toggle

Screen Shot 2022-01-26 at 2 28 55 PM

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -59,7 +60,14 @@ $pf-c-tree-view--MaxDepth: 10;
...
+    padding: var(--pf-c-table--cell--PaddingTop) var(--pf-c-table__tree-view-main--c-table__check--MarginRight) var(--pf-c-table--cell--PaddingBottom) var(--pf-c-table__tree-view-main--c-table__check--PaddingLeft);
```
mcoker, line 69: FWIW I ran the change by @mceledonia and we agreed to keep the existing spacing, but add an extra 8px between the expandable toggle and checkbox. This change increases the space between the checkbox and text an additional 8px, too. It may be fine - wdyt @mceledonia @mmenestr 

Additional 8px spacing in this PR

Screen Shot 2022-01-26 at 2 27 33 PM

Maintaining the existing space between the checkbox and text

Screen Shot 2022-01-26 at 2 27 49 PM

mcoker (changes requested):
Looks good! Left a couple of comments.

src/patternfly/components/Table/table-tree-view.scss
mcoker: @mceledonia @mmenestr @mcarrano can you review this spacing change?

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -9,7 +9,8 @@ $pf-c-tree-view--MaxDepth: 10;
   --pf-c-table__tree-view-main--MarginLeft: calc(var(--pf-c-table--cell--PaddingLeft) * -1);
 
   // check
-  --pf-c-table__tree-view-main--c-table__check--MarginRight: var(--pf-global--spacer--sm);
```
mcoker, line 12: Removing this will break core users since they don't have the new `<label>`

Screen Shot 2022-02-01 at 11 04 50 AM

mcoker (changes requested):

mcoker:
> It looks like the spacing between the checkbox and text is unchanged.

@mcarrano you're right, looks like the last update addressed that.

src/patternfly/components/Table/table-tree-view.scss
mattnolting: @mcoker Whoops, updated.

mcoker (approved):
Perfecto!

Not related to the PR, but would you mind also adding a "Tree table" section header and maybe updating the example titles so the tree table section matches the others? Most of the others follow the format of "Tree table" -> "Tree table example", "Tree table with checkboxes example", etc.

Right now it looks like it's part of the "
Screen Shot 2022-02-02 at 1 41 28 PM
hoverable and selected" section

mmenestr:
@mcoker I like the added spacing better than without the added spacing, though i'm not seeing any added spacing in the staging preview? 

image

mcoker (approved):
🚀

Issue #4643, opened 2022-01-26
Table - striped tr example has unnecessary class


mcoker:
http://www.patternfly.org/v4/components/table/html/striped-tr-example

The table doesn't need `.pf-m-striped` since the example shows how to add classes to the rows directly.

PR #4644, opened 2022-01-26
chore(table): removed unnecessary class from striped tr example

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4643


srambach (approved):
✅ LGTM

PR #4649, opened 2022-01-31
feat(nav): added variation to fix section spacing

### LINKED ISSUE #4602: Bug: navigation section margins
* Remove `.pf-c-nav__section + .pf-c-nav__section` `margin-top`, apply to `section-title`
* Add optional title text to `aria-label` 
* Add example to core

mattnolting:
fixes #4602 

@mcarrano @mceledonia I added a divider between sections to separate them and an `aria-label` to sections without titles. Lmk if you this presentation is what you expect.

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -1078,7 +1078,7 @@
     --pf-c-nav__section--MarginTop: var(--pf-c-nav__section--section--MarginTop);
   }
 
-  &.pf-m-no-margin-top {
+  &.pf-m-no-title {
     --pf-c-nav__section--MarginTop: 0;
   }
 }
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -1074,8 +1074,8 @@
   // Divider
   --pf-c-nav--c-divider--MarginBottom: 0;
 
-  & + & {
-    --pf-c-nav__section--MarginTop: var(--pf-c-nav__section--section--MarginTop);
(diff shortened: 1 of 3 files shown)


mcarrano (commented):
Looks fine to me. but I'll defer to you @mceledonia .

src/patternfly/components/Nav/nav.scss
```diff
@@ -1060,8 +1060,8 @@
   // Divider
   --pf-c-nav--c-divider--MarginBottom: 0;
 
-  & + & {
-    --pf-c-nav__section--MarginTop: var(--pf-c-nav__section--section--MarginTop);
+  & + & .pf-c-nav__section-title {
+    margin-top: var(--pf-c-nav__section--section--MarginTop); // update at breaking change
```
mcoker, line 1064: This works, but can break someone who had `& + & { margin-top: [custom]; }` - they'll get a double margin. This is what has broken the openshift folks in the past. I think it's probably fine but nav isn't something we want to break. Do you think there is a different way?

Just an aside, this will be a great use of `:has()` when it lands!!

```
& + &:has(.pf-c-nav__section-title) {
  --pf-c-nav__section--MarginTop: var(--pf-c-nav__section--section--MarginTop);
}
```

mcoker (changes requested):

src/patternfly/components/Nav/nav.scss
mattnolting: @mcoker updated

mcoker (approved):
🔥🔥

PR #4651, opened 2022-01-31
feat(table): added overflow menu examples

### LINKED ISSUE #4607: Table demo - add example using overflow-menu instead of dropdown + extra buttons
related to: https://github.com/patternfly/patternfly-react/issues/6776#issuecomment-1012226609

mattnolting:
closes #4607

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -975,6 +975,10 @@
   vertical-align: middle;
 }
 
+.pf-c-table__action {
+  text-align: right;
+}
+
 // Inline edit
 .pf-c-table__inline-edit-action {
   --pf-c-table--cell--PaddingLeft: 0;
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -975,6 +975,10 @@
   vertical-align: middle;
 }
 
+.pf-c-table__action {
+  text-align: right;
```
mcoker, line 979: I agree we should push the stuff to the right, but this could cause 2 problems

* `text-align` is inherited, so it is going to apply to all children. looks like we set `text-align` in our menu items since we use buttons and those are center aligned by default.
* It could cause an unexpected layout change users might need to address when they pull in this version. As common/integral as the table is, I think we should wait for a breaking change to change the alignment.

wdyt?

mcoker (commented):
Looks fantastic - one comment about `text-align`

src/patternfly/components/Table/table.scss
mattnolting: @mcoker wfm, captured here: https://github.com/patternfly/patternfly/issues/4661

mcoker (approved):
🥳

PR #4653, opened 2022-02-01
fix(label): improve click area of editable label

### LINKED ISSUE #4626: Update mark up for Editable labels
> I don't think this always meets minimum click size standards.

There are a couple of things we can do about that. If you look at the link labels, the `__content `element is what becomes the link and we use a `::before pseudo` element attached to` __content` to fill the entire width/height of the label so that the whole label will trigger a click of the label (except for the close button if that is present, since the close button sits on top of the `__content `element).

Are we able to move the `button` element to `__content `- and potentially remove the need for `__editable-text`? That would make the whole label clickable. Alternatively, we could use the same `::before` pseudo element trick that we use on `__content` on the `__editable-text`element, too, doing the same thing. Either of those would make a single letter label's click target about the size of the close button.

> Also, double clicking in some areas outside the labels can cause text within to highlight.

I'm seeing that with regular labels, too, unless you're referring to something different. As a chronic double-clicker on everything as a nervous habit kind of person, I think that's just how the browser works - double clicking around something usually selects text within it.

This is a follow up issue in response to this [react PR](https://github.com/patternfly/patternfly-react/pull/6472)

srambach:
Fixes #4626 

Moves `__content` to the button to make a larger click area.


src/patternfly/components/Label/label-content.hbs
```diff
@@ -1,11 +1,13 @@
-<{{# if label-content--IsLink}}a{{else}}span{{/if}} class="pf-c-label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
+{{# if label--IsEditable}}
+  {{#> label-editable-text}}{{> @partial-block}}{{/label-editable-text}}
+{{else}}
+<{{# if label-content--IsLink}}a{{else if label--IsEditable}}button{{else}}span{{/if}} class="pf-c-label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
```
mcoker, line 4: looks like you could take out the `label--IsEditable` vars in this conditional block since if the var is set, this part isn't included.

src/patternfly/components/Label/label-content.hbs
```diff
@@ -1,11 +1,13 @@
...
   {{/if}}
   {{#if label-content--attribute}}
     {{{label-content--attribute}}}
   {{/if}}>
-  {{#if label--IsEditable}}
-    {{#> label-editable-text}}{{> @partial-block}}{{/label-editable-text}}
-  {{else}}{{~> @partial-block}}{{/if}}
-</{{# if label-content--IsLink}}a{{else}}span{{/if}}>
+  {{~> @partial-block}}
+</{{# if label-content--IsLink}}a{{else if label--IsEditable}}button{{else}}span{{/if}}>
```
mcoker, line 12: same here for `label--IsEditable`

src/patternfly/components/Label/label-editable-text.hbs
```diff
@@ -1,4 +1,4 @@
-<{{#if label-editable-text--type}}{{label-editable-text--type}}{{else if label--IsEditableActive}}input{{else}}button{{/if}} class="pf-c-label__editable-text{{#if label-editable-text--modifier}} {{label-editable-text--modifier}}{{/if}}"
+<{{#if label-editable-text--type}}{{label-editable-text--type}}{{else if label--IsEditableActive}}input{{else}}button{{/if}} class="pf-c-label__content pf-c-label__editable-text{{#if label-editable-text--modifier}} {{label-editable-text--modifier}}{{/if}}"
```
mcoker, line 1: Typically we don't share BEM elements like this `pf-c-label__content pf-c-label__editable-text`.

Is it necessary for this to have the `__editable-text` class, or could we just target it (including in react) by the `__content` class when the parent is `.pf-m-editable`?

Also I wonder if this logic should be built in to the `__content` conditional? wdyt?

src/patternfly/components/Label/label.scss
```diff
@@ -204,6 +212,9 @@
     --pf-c-label--m-outline__content--before--BorderColor: var(--pf-c-label--m-outline--m-green__content--before--BorderColor);
     --pf-c-label--m-outline__content--link--hover--before--BorderColor: var(--pf-c-label--m-outline--m-green__content--link--hover--before--BorderColor);
     --pf-c-label--m-outline__content--link--focus--before--BorderColor: var(--pf-c-label--m-outline--m-green__content--link--focus--before--BorderColor);
+    --pf-c-label--m-editable__content--before--BorderColor: var(--pf-c-label--m-green--before--BorderColor);
```
mcoker, line 215: looks like these green vars are missing `__content` ie, `--pf-c-label--m-green__content--before--BorderColor`

mcoker (changes requested):
LGTM! Only real issue is the green borders aren't working.

src/patternfly/components/Label/label.scss
```diff
@@ -122,7 +122,9 @@
 
   // Content
   --pf-c-label__content--Color: var(--pf-global--Color--100);
+  --pf-c-label__content--MaxWidth: initial;
```
mcoker, line 125: Do you know if the `max-width` default for a flex item is `none`? That's what `initial` initial is for `max-width`.  https://developer.mozilla.org/en-US/docs/Web/CSS/max-width

I ask because `min-width` for a flex item is `min-content`, which is different from the `initial` value.

mcoker (commented):
Looks great! Just one question about the `max-width` default for a flex item. Curious if setting it to `initial` is going to cause a change in the `max-width` behavior.

src/patternfly/components/Label/label.scss
srambach: I changed it to 100%, which is max-width for pf-flex children. Also that seems to make more sense in general, right?

mcoker (approved):
LGTM!

PR #4654, opened 2022-02-02
feat(theme): added dark theme

mcoker:
**Note:** This is not production ready and we are not advising products to adopt the dark theme stylesheet at this time. This feature is being released to test in specific use cases to assess the feasibility of adding a dark theme to PatternFly.

--

Adds `patternfly-theme-dark.css`
* This stylesheet is intended to be loaded after all PatternFly styles (base, components, layouts, etc), but before any existing overrides.
* Dark theme styles are enabled by adding the class `pf-theme-dark` to the `<html>` tag.
* The dark theme overrides use the selector `:where(.pf-theme-dark)` so that the use of the additional class to target dark theme overrides does not increase the specificity of the overrides.
  * Generally speaking, this works better for:
    * Making small adjustments to components for dark theme styles without introducing new specificity issues.
    * Ensuring most existing overrides/customizations on top of PatternFly should still continue to work as long as they're loaded after the dark theme styles from PatternFly.


evwilkin (approved):
LGTM - thanks @mcoker! 🎉

DaoDaoNoCode (commented):
Screen Shot 2022-02-10 at 11 29 07 AM
Looks like the demos are broken, any ideas?

mcoker:
> Looks like the demos are broken, any ideas?

@DaoDaoNoCode oh good catch. That's a separate issue. Just rebased - should fix it.

evwilkin:
> > Looks like the demos are broken, any ideas?
> 
> @DaoDaoNoCode oh good catch. That's a separate issue. Just rebased - should fix it.

Looks like the rebase fixed this issue 👏  though FYI the preview link is missing a screenshot for the first `Page` demo now (https://patternfly-pr-4654.surge.sh/components/page/html-demos#basic)

mcoker:
@evwilkin awesome! I regenerate all screenshots in the release notes PR, so that should be fine.

src/patternfly/patternfly-theme-dark.scss
```diff
@@ -7,3 +7,9 @@ $pf-theme-dark-class: ".pf-theme-dark" !default;
 :where(#{$pf-theme-dark-class}) {
   @include pf-theme-dark($pf-theme-dark-class);
 }
+
+/* stylelint-disable */
+.ws-preview,
+.ws-example {
```
DaoDaoNoCode, line 13: Maybe also add `.ws-code` here? So the doc will look nicer.

mcoker: Oh I'm just adding these temporarily and was planning on taking them out. Ideally I think we would put dark theme workspace changes in the patternfly-org repo where the workspace/website theme styles are hosted.

Issue #4621, opened 2022-01-18
Update card demo actions to use plain toggle


mcoker:
Anywhere we have an action in the top-right like this, let's use a plain toggle instead of the regular boxed toggle.

https://www.patternfly.org/v4/components/card/html-demos/#trend-card-1
<img width="461" alt="Screen Shot 2022-01-18 at 10 17 46 AM" src="https://user-images.githubusercontent.com/35148959/149976072-50d908d3-780c-4729-9880-ef380d5580e0.png">

Issue #4553, opened 2021-12-02
Better align card actions with title


mcoker:
re: https://github.com/patternfly/patternfly-react/pull/6639#pullrequestreview-819106228

See if we can better align actions in cards with different title font sizes

mcoker:
@mcarrano @mmenestr @mceledonia can I get your opinion on our ideal card title and action alignment from the conversation in https://github.com/patternfly/patternfly-react/pull/6639#pullrequestreview-819106228? 

By default, actions are top-aligned with the card title. Since we expect most actions to be "plain", we move (offset) the action position up by 6px (the amount of the action's top padding), so that the top of the title text aligns with the top of the button text/icon. This is what's seen in the demos currently. It works better with the default card title size, but looks odd with large text.

<img width="554" alt="Screen Shot 2022-01-17 at 5 35 53 PM" src="https://user-images.githubusercontent.com/35148959/149847780-86611a09-412d-46a7-8b34-02266f335df3.png">

In cases where the user isn't using a plain action, we have a variation for the action that removes the offset, so that the top of the action itself (ie, a primary button) aligns with the top of the title text.

<img width="441" alt="Screen Shot 2022-01-17 at 5 34 37 PM" src="https://user-images.githubusercontent.com/35148959/149847716-b8fa7871-9383-41b0-9f3d-156b83da0185.png">

Based off of knowing the default behavior, what do you think is the ideal alignment here? And is that alignment an option/default we want to make in the PF card component, or just something to demonstrate here in the demos?

mmenestr:
@mcoker I think what you have in those examples makes sense. I wonder if we could just say "horizontally center align" header and button every time, which would basically go along with what you have there.

So if the header text height is smaller than the button height, you'll have parts of the button "stick out" equally above and below. And if the header text height is the same as the entire button height, then there's perfect alignment. Hopefully that makes sense?

I also think it would make sense to make it some sort of default in the card component + something to show in the demo. Though we'd also want to keep in mind which buttons we'd want to encourage people to use in cards vs. not, and decide which buttons we'd want to show in a demo based on that.

mcarrano:
[...]

PR #4667, opened 2022-02-04
chore(card): updated card demo select to be plain, fixed alignment

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4553
fixes https://github.com/patternfly/patternfly/issues/4621

@mcarrano I just added custom CSS to change the card action/title position so they're aligned to baseline.

### DIFF
--- a/src/patternfly/demos/Card/examples/Card.md
+++ b/src/patternfly/demos/Card/examples/Card.md
@@ -744,6 +744,8 @@ import './Card.css'
 
 ### Utilization card 3
 ```hbs
+<b>Note:</b> Custom CSS is used in this demo to align the card title and select toggle text to <code>baseline</code> alignment.
+<br><br>
 {{#> gallery gallery--modifier="pf-m-gutter" gallery--attribute='style="--pf-l-gallery--GridTemplateColumns--min: 360px;"'}}
   {{#> card card--id="utilization-card-3-example"}}
     {{#> card-header card-header--modifier="pf-u-align-items-flex-start"}}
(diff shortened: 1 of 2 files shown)


mcarrano (approved):
I think that these look much better. Thanks @mcoker

srambach:
Looks good. The workspace tweak should be documented in a note near the examples using it for someone who would like to replicate this to know that they need to use a layout or custom CSS to make this line up properly.

mcoker:
@srambach thanks, updated!

srambach (approved):
I have to admit I'm struggling with the wording of the note, as it implies that you've set `vertical-align:baseline;` somewhere rather than tweaking the padding. And since this is in workspace CSS, a reader will need to do some work to find it in the inspector. I know it's a minor thing, but it's bugging me that workspace code is being used to fix the demo as opposed to fixing the framework around the demo or showing something invisible like we do in the layouts. That said, I don't have a better solution so I'll approve. 😕  🤷‍♀️

srambach:
@mcoker Yeah - I thought about inline styles..if you can do that I think that would a good solution.

mcoker:
@srambach updated

src/patternfly/demos/Card/examples/Card.md
```diff
@@ -959,8 +963,8 @@ import './Card.css'
           Workload
         </span>
       {{/l-flex}}
-      {{#> card-actions}}
-        {{#> select id=(concat card--id '-select-dropdown') select-menu--modifier="pf-m-align-right"}}
+      {{#> card-actions card-actions--modifier="pf-m-no-offset" card-actions--attribute='style="padding-top: 3px;"'}}
```
srambach, line 966: Was this supposed to be 1 px?

srambach (approved):
Awesome, I like this much better. Just checking on one adjustment. Thanks!

src/patternfly/demos/Card/examples/Card.md
mcoker: doh!

PR #4669, opened 2022-02-07
feat(code-editor): added header content section

### LINKED ISSUE #4658: Add layout style for functional button in the CodeEditor header
Based on the design here https://marvelapp.com/prototype/15bb10j4/screen/81521392, we are going to add a demo to support the 'View shortcuts' feature, and it would be nice if the layout style of the button could be supported here.

mattnolting:
closes #4658 

Preview: http://localhost:8001/components/code-editor#with-optional-header-content-and-keyboard-shortcuts

### DIFF
--- a/src/patternfly/components/CodeEditor/code-editor.scss
+++ b/src/patternfly/components/CodeEditor/code-editor.scss
@@ -96,14 +96,6 @@
   padding-left: var(--pf-c-code-editor__header-main--PaddingLeft);
 }
 
-.pf-c-code-editor__keyboard-shortcuts {
-  margin-left: auto;
-
-  & + .pf-c-code-editor__tab {
-    margin-left: 0;
-  }
-}
-
 .pf-c-code-editor__main {
   position: relative;
   background-color: var(--pf-c-code-editor__main--BackgroundColor);
@@ -160,7 +152,7 @@
   display: flex;
   align-items: center;
   padding: var(--pf-c-code-editor__tab--PaddingTop) var(--pf-c-code-editor__tab--PaddingRight) var(--pf-c-code-editor__tab--PaddingBottom) var(--pf-c-code-editor__tab--PaddingLeft);
-  margin-left: auto;
+  margin-left: auto; // remove at breaking change
   color: var(--pf-c-code-editor__tab--Color);
   background-color: var(--pf-c-code-editor__tab--BackgroundColor);
   border-color: var(--pf-c-code-editor__tab--BorderColor);
--- a/src/patternfly/components/CodeEditor/code-editor.scss
+++ b/src/patternfly/components/CodeEditor/code-editor.scss
@@ -27,6 +27,10 @@
   --pf-c-code-editor__code-pre--FontSize: var(--pf-global--FontSize--sm);
   --pf-c-code-editor__code-pre--FontFamily: var(--pf-global--FontFamily--monospace);
 
+  // content
+  --pf-c-code-editor__content--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-code-editor__content--PaddingLeft: var(--pf-global--spacer--sm);
+
   // tab
   --pf-c-code-editor__tab--BackgroundColor: var(--pf-global--BackgroundColor--100);
   --pf-c-code-editor__tab--Color: var(--pf-global--Color--200);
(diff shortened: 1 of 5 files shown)


DaoDaoNoCode (approved):
The style looks great to me! Thanks, @mattnolting. However, I am not sure whether it's good to call it `pf-c-code-editor__content` because it may cause some confusion with the main content of the code editor. But I cannot think of a better classname, either. 🤔

mcoker:
@DaoDaoNoCode agree, maybe `header-content`. Or `header-main` to match the card's header for the container to put whatever you want in.

src/patternfly/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -149,6 +149,52 @@ cssPrefix: pf-c-code-editor
...
+      {{/button}}
+      {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Download code"'}}
+        <i class="fas fa-download"></i>
+      {{/button}}
+      {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Upload code"'}}
+        <i class="fas fa-upload"></i>
+      {{/button}}
+    {{/code-editor-controls}}
+    {{#> code-header-main}}
+      {{#> l-flex}}
```
mcoker, line 168: Typically we would just put some placeholder text here so it doesn't imply the flex layout is part of or styled by the code editor. wdyt about doing that here and either creating a demo, or just showing @DaoDaoNoCode how to add it.

mcoker (approved):
Looks great! Just one comment, but it's fine as is.

src/patternfly/components/CodeEditor/examples/CodeEditor.md
mattnolting: Yeah, I thought about that. My reasoning was that this addition didn't necessarily warrant a demo, but it is a demo. @DaoDaoNoCode you good w/me adding the section and you using a flex layout within `__header-main`?

mcarrano (approved):
Looks good to me. Thanks @mattnolting !

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -86,6 +90,13 @@
   }
 }
 
+.pf-c-code-editor__header-main {
+  flex-grow: 1;
+  align-self: center;
```
mcoker, line 95: I'm not sure that I would expect this. By default it would do something like this if the tab text font were bigger (just as an example)
Screen Shot 2022-02-08 at 6 48 45 PM

I think I would expect it to be `flex-end` like the other elements, and if the children of that element need a different alignment, that could be done with a layout.

I realize it will make the example look like this, but I think that's ok... wdyt @mcarrano?

Screen Shot 2022-02-08 at 6 52 51 PM

mcarrano:
> I realize it will make the example look like this, but I think that's ok... wdyt @mcarrano?

I'm OK with that if you guys think this makes more sense. So to make sure I understand, we would need to include a layout within the React demo to get the button alignment we want for the "View shorcuts" button? Does that work for you @DaoDaoNoCode ?

DaoDaoNoCode:
@mcarrano I am OK with it, but I want to make sure of one thing. Do we need to create a demo for `View shortcuts` button or do we wrap that button into the Code Editor component and only provide the user with a prop like `isShortcutsVisible`? The shortcuts are always the same and we should not let the user define it and create the button/popover.

mattnolting:
@DaoDaoNoCode as these aren't natively supported features, you'll need to manage presentation. I can help after merge.

@mcoker updated

mcoker (approved):

mcoker:
@mcarrano @DaoDaoNoCode would we always want the button to be in the same position in all code editors? Assuming so, and that we don't want users to change the button type, text, etc, or the popover contents, I'm wondering if this should be a structural element. If we don't care where they put it, but want consistency with the button/popover, maybe we could create a `<CodeEditorShortcuts />` component that bundles the button/popover, and users can put it in their own layout wherever they want inside (or outside I guess?) the code editor component.

DaoDaoNoCode:
@mcoker Maybe there is no need to create a component? We could follow the patterns of other features like upload/download and use a prop to set it. WDYT?

mcoker (approved):
Looks great! One nit is the icon isn't really aligned with the text and it's making the button be 39px tall instead of 36px, but that's a pficon problem and out of the scope of this change.

PR #4670, opened 2022-02-07
feat(menu): added aria attributes

### LINKED ISSUE #4655: Add appropriate aria attributes to `pf-c-menu`
Provide direction for [6681](https://github.com/patternfly/patternfly-react/issues/6881)

====================================================================

## `role="menubar"` is used incorrectly in menu implementation.

@jessiehuff @mcoker 

Screen Shot 2022-02-02 at 11 26 20 AM

https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#menubar

```
menubar (role)§
A presentation of menu that usually remains visible and is usually presented horizontally.

The menubar role is used to create a menu bar similar to those found in Windows, Mac, and Gnome desktop applications. A menu bar is used to create a consistent set of frequently used commands. Authors SHOULD ensure that menubar interaction is similar to the typical menu bar interaction in a desktop graphical user interface.

To be keyboard accessible, authors SHOULD manage focus of descendants for all instances of this role, as described in Managing Focus.

Elements with the role menubar have an implicit aria-orientation value of horizontal.
```
Looks like the issue was introduced here: https://github.com/patternfly/patternfly-react/pull/5789/files#diff-b36dd79f43392929bd33775b355ed31e3307fc83c7f18bf61b897fee3d4d4dc2R17

Screen Shot 2022-02-02 at 11 30 13 AM

Also, looks like we have a `role` conflict on the child `<li role="none">`. `role="none"` is correct, however this role [cannot be focusable (either natively or using tabindex)](https://dequeuniversity.com/rules/axe/4.2/presentation-role-conflict). 

Screen Shot 2022-02-02 at 11 32 13 AM

```
Ensure that each element with role="none" or role="presentation" does not have one of the following characteristics:

A global ARIA attribute
Is focusable (either natively or using tabindex)
```

Visible in all menu implementations

Bug, not blocking

PF next release

### In menu
```
.pf-c-menu
  .pf-c-menu__list(role="menu")
    .pf-c-menu__list-item(role="none") // remove tabindex
      .pf-c-menu__item(role="menuitem")
```

### In nav
```
.pf-c-nav.pf-m-horizontal
  .pf-c-menu(role="menubar")
    .pf-c-menu__list(role="menu")
      .pf-c-menu__list-item(role="none") // remove tabindex
        .pf-c-menu__item(role="menuitem")
```

mattnolting:
closes #4655

### DIFF
--- a/src/patternfly/components/Menu/examples/Menu.md
+++ b/src/patternfly/components/Menu/examples/Menu.md
@@ -1711,7 +1711,10 @@ import './Menu.css'
 ### Accessibility
 | Attribute | Applied | Outcome |
 | -- | -- | -- |
+| `role="menu"` | `.pf-c-menu__list` | Declares `.pf-c-menu__list` as a menu. |
 | `disabled` | `button.pf-c-menu__item` | When the menu item uses a button element, indicates that it is unavailable and removes it from keyboard focus. |
+| `role="menuitem"` | `.pf-c-menu__item` | Assigns `.pf-c-menu__item` as an option in a set of choices contained by a menu. |
+| `role="none"` | `.pf-c-menu__list-item` | Removes semantic meaning from `.pf-c-menu__list-item`. |
 | `aria-disabled="true"` | `a.pf-c-menu__item` | When the menu item uses a link element, removes it from keyboard focus. |
 | `tabindex="-1"` | `a.pf-c-menu__item` | When the menu item uses a link element, removes it from keyboard focus. |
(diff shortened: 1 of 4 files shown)


mcoker:
@nicolethoen @jessiehuff what do you think would be the best way to test this change? Should we create a react issue so we can test it there and make changes as needed? Assuming we make a change, we'll hold off on this PR so it can go in the same release with the react change.

nicolethoen:
I think that the axe tests passing in core is almost as good as we can do here. I agree with holding off on the PR until the react PR gets merged, hopefully as soon as code freeze ends. The corresponding change will be tested with VO on the react side so we will know if we need to make more updates there. But getting a thumbs up on the interact-able react side is likely a good sign for this.

PR #4672, opened 2022-02-07
fix(nav): updated nav menu styling

### LINKED ISSUE #4656: Update navigation + menu styling
@mcarrano 
> You can see where the header text is slightly heavier and the bottom border is thicker. I realize that these may be core styling issues that were missed in the original review, and therefore am copying in @mattnolting to comment. @mceledonia would also be good to get your input since (Becca) the original designer is no longer with the company.

* Update `pf-c-menu__item-text` font-weight for back button
* Update `pf-c-divider` under back button weight
* Update back chevron positioning in navigation according to: https://www.figma.com/proto/AJi4XmCkrdGEK532YmbAqH/tertiary-nav?node-id=54%3A2191&scaling=min-zoom&page-id=53%3A2106

mattnolting:
closes #4656

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -329,6 +329,8 @@
   --pf-c-nav--c-menu__list-item--focus-within--Color: var(--pf-global--Color--light-100);
   --pf-c-nav--c-menu__list-item--hover--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
   --pf-c-nav--c-menu__list-item--focus-within--BackgroundColor: var(--pf-global--BackgroundColor--dark-200);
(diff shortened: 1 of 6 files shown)


mcarrano (approved):
This looks much better. Thanks @mattnolting !

src/patternfly/components/Nav/nav.scss
```diff
@@ -489,6 +491,18 @@
     }
   }
 
+  .pf-c-menu__list-item:where(.pf-m-drill-up) {
```
mcoker, line 494: Curious what `:where()` is used for here - is it to maintain specificity or something else?

mcoker (approved):
LGTM!

Issue #4666, opened 2022-02-04
Custom style for TextInputGroup within a MenuToggle


jeffpuzzo:
**Objective:** Leverage Menu's components to achieve a similar visual representation of a typeahead select.

**Issue Description:**
Using a TextInput inside of a MenuToggle leads to excess padding and a border that is undesired around that input. 

**Possible Solution:**
Create a custom style/class for a borderless input specifically for use inside of a MenuToggle that has negative margins to offset the padding MenuToggle already has.

Once completed the following can be worked on:
https://github.com/patternfly/patternfly-react/issues/6888

nicolethoen:
@jeffpuzzo does the text input group need to go inside a menu toggle? Could you just append a menu to a text input group?

One of the reasons I ask is because I believe a menu toggle is a button - and you cannot have buttons inside buttons, so the controls of the text input group would not be valid inside a menu toggle as it is currently defined.

PR #4673, opened 2022-02-08
feat(menu-toggle/text-input-group): update menu-toggle for text inputs & add plain text input group

jeffpuzzo:
**Closes:** #4666

### DIFF
--- a/src/patternfly/components/MenuToggle/examples/MenuToggle.md
+++ b/src/patternfly/components/MenuToggle/examples/MenuToggle.md
@@ -188,7 +188,7 @@ import './MenuToggle.css'
 
 ### Split button (checkbox)
 ```hbs
-{{#> menu-toggle menu-toggle--id="split-button-checkbox-disabled-example" menu-toggle--IsDisabled="true" menu-toggle--IsSplitButton="true"}}
+{{#> menu-toggle menu-toggle--id="split-button-checkbox-disabled-example" menu-toggle--IsDisabled="true" menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true"}}
   {{> menu-toggle--check menu-toggle--check--IsStandalone="true"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
@@ -199,7 +199,7 @@ import './MenuToggle.css'
 
 &nbsp;
 
-{{#> menu-toggle menu-toggle--id="split-button-checkbox-example" menu-toggle--IsSplitButton="true"}}
+{{#> menu-toggle menu-toggle--id="split-button-checkbox-example" menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true"}}
   {{> menu-toggle--check menu-toggle--check--IsStandalone="true"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
@@ -210,7 +210,7 @@ import './MenuToggle.css'
 
 &nbsp;
 
-{{#> menu-toggle menu-toggle--id="split-button-checkbox-expanded-example" menu-toggle--IsExpanded="true" menu-toggle--IsSplitButton="true"}}
+{{#> menu-toggle menu-toggle--id="split-button-checkbox-expanded-example" menu-toggle--IsExpanded="true" menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true"}}
   {{> menu-toggle--check menu-toggle--check--IsStandalone="true"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
@@ -222,7 +222,7 @@ import './MenuToggle.css'
 
 ### Split button (checkbox with toggle text)
 ```hbs
-{{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-disabled-example" menu-toggle--IsSplitButton="true" menu-toggle--IsDisabled="true"}}
+{{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-disabled-example" menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true" menu-toggle--IsDisabled="true"}}
   {{> menu-toggle--check menu-toggle--check--text="10 selected" check--IsDisabled="true"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
@@ -233,7 +233,7 @@ import './MenuToggle.css'
 
 &nbsp;
 
-{{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-example" menu-toggle--IsSplitButton="true"}}
+{{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-example" menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true"}}
   {{> menu-toggle--check menu-toggle--check--text="10 selected"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
@@ -244,7 +244,7 @@ import './MenuToggle.css'
 
 &nbsp;
 
-{{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-expanded-example" menu-toggle--IsSplitButton="true" menu-toggle--IsExpanded="true"}}
+{{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-expanded-example" menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true" menu-toggle--IsExpanded="true"}}
   {{> menu-toggle--check menu-toggle--check--text="10 selected"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
@@ -257,7 +257,7 @@ import './MenuToggle.css'
 
 ### Split button, primary
 ```hbs
-{{#> menu-toggle menu-toggle--modifier="pf-m-primary" menu-toggle--id="split-button-checkbox-primary-disabled-example" menu-toggle--IsSplitButton="true" menu-toggle--IsDisabled="true"}}
+{{#> menu-toggle menu-toggle--modifier="pf-m-primary" menu-toggle--id="split-button-checkbox-primary-disabled-example"  menu-toggle--IsDiv="true" menu-toggle--IsSplitButton="true" menu-toggle--IsDisabled="true"}}
   {{> menu-toggle--check menu-toggle--check--text="10 selected" check--IsDisabled="true"}}
   {{#> menu-toggle-button menu-toggle-button--IsToggle="true"}}
     {{#> menu-toggle-controls}}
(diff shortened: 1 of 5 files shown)


src/patternfly/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -250,6 +250,27 @@ import './MenuToggle.css'
 {{/menu-toggle}}
 ```
 
+### Typeahead select
```
mcoker, line 253: I'd just call it "typeahead" not to tie it to particular usage

```suggestion
### Typeahead
```

src/patternfly/components/MenuToggle/menu-toggle.hbs
```diff
@@ -1,4 +1,4 @@
-<button class="pf-c-menu-toggle
+<div class="pf-c-menu-toggle
```
mcoker, line 1: This element should still be a button by default, just needs to be a div for this version and other versions down the road, like split button, which we'll add at some point. Typically the way we do this is via `{{#if menu-toggle--type}}{{menu-toggle--type}}{{else}}button{{/if}}` for the opening and close tags, which leaves the default as it is (doesn't break any current examples) and allows you to set `menu-toggle--type="div"` on the typeahead example. Here's the select toggle as an example - https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Select/select-toggle.hbs


```suggestion
<{{#if menu-toggle--type}}{{menu-toggle--type}}{{else}}button{{/if}} class="pf-c-menu-toggle
```

src/patternfly/components/MenuToggle/menu-toggle.hbs
```diff
@@ -15,7 +15,6 @@
     pf-m-expanded
   {{/if}}
   {{#if menu-toggle--modifier}} {{menu-toggle--modifier}}{{/if}}"
-  type="button"
```
mcoker, line 18: we'll want to leave this since it should still be a `<button>` by default, and a button is `type="submit"` by default, which we don't want. Though we should wrap it in something like `{{#unless menu-toggle--type}}`, which will unset it if you specify an element other than a `<button>` here. The select toggle does the same

https://github.com/patternfly/patternfly/blob/871aa484334fe6d14cb9f4a479188df97b67e6f6/src/patternfly/components/Select/select-toggle.hbs#L5-L7

src/patternfly/components/MenuToggle/menu-toggle.hbs
```diff
@@ -28,4 +27,4 @@
     {{{menu-toggle--attribute}}}
   {{/if}}>
   {{> @partial-block}}
-</button>
+</div>
```
mcoker, line 30:
```suggestion
</{{#if menu-toggle--type}}{{menu-toggle--type}}{{else}}button{{/if}}>
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -121,6 +121,7 @@
   font-size: var(--pf-c-menu-toggle--FontSize);
   line-height: var(--pf-c-menu-toggle--LineHeight);
   color: var(--pf-c-menu-toggle--Color);
+  cursor: pointer;
```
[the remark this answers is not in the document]
mcoker: I think this is fine. We do the same with the select toggle 👍

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -276,6 +277,14 @@
 
     height: 100%;
   }
+
+  .pf-c-text-input-group {
+    margin:
```
mcoker, line 282: This works, but it's probably better if we create local component vars for this. So this at the top of the scss

```
--pf-c-menu-toggle--c-text-input-group--MarginTop: calc(var(--pf-c-menu-toggle--PaddingTop) * -1);
--pf-c-menu-toggle--c-text-input-group--MarginRight: calc(var(--pf-c-menu-toggle--PaddingRight) * -1);
--pf-c-menu-toggle--c-text-input-group--MarginBottom: calc(var(--pf-c-menu-toggle--PaddingBottom) * -1);
--pf-c-menu-toggle--c-text-input-group--MarginLeft: calc(var(--pf-c-menu-toggle--PaddingLeft) * -1);
```

Then in this block:

```
margin: var(--pf-c-menu-toggle--c-text-input-group--MarginTop) var(--pf-c-menu-toggle--c-text-input-group--MarginRight) var(--pf-c-menu-toggle--c-text-input-group--MarginBottom) var(--pf-c-menu-toggle--c-text-input-group--MarginLeft);
```

This creates menu toggle component vars to restyle the text input group margin if a product wanted to do that, so that they don't have to write CSS.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -61,6 +61,17 @@
 
     pointer-events: none;
   }
+
+  &.pf-m-plain {
+    --pf-c-text-input-group--backgroundcolor: transparent;
```
mcoker, line 66: css vars are case sensitive

```suggestion
    --pf-c-text-input-group--BackgroundColor: transparent;
```

mcoker (changes requested):
Looks really good! Left some comments - you'll also want to set the toggle to look focused when you've focused on an element in the toggle. You can do that by combining `:focus` here with `:focus-within` - https://github.com/patternfly/patternfly/blob/871aa484334fe6d14cb9f4a479188df97b67e6f6/src/patternfly/components/MenuToggle/menu-toggle.scss#L230

like

```
&:focus,
&:focus-within
```

mcoker (changes requested):
@jeffpuzzo apologies, something else I just noticed is we're going to need a button toggle in the menu toggle as a manual button to toggle the menu open. This will replace the current caret icon, which is just an icon since the menu toggle itself is normally the button. The typeahead select has something similar.

What this might look like is
* Replace `pf-c-menu-toggle__controls` with a new div `pf-c-menu-toggle__button`
* Create a plain `<Button>` there to be the toggle
* Move the `aria-expanded` attribute from `div.pf-c-menu-toggle` to the button you created in the previous step
* Apply some negative margins to the `pf-c-menu-toggle__button` element to fix the spacing<sup>[1]</sup>
* Add some more styles to the text input group when it's in the menu toggle<sup>[2]</sup>
* I'm sure I forgot something so let me know if something doesn't work!

1.
```
margin-top: calc(var(--pf-global--spacer--form-element) * -1);
margin-bottom: calc(var(--pf-global--spacer--form-element) * -1);
margin-right: calc(var(--pf-c-menu-toggle--PaddingRight) * -1);
}
```

2.
```
--pf-c-text-input-group__utilities--MarginRight: 0;
--pf-c-menu-toggle--c-text-input-group--MarginRight: 0;
```

src/patternfly/components/MenuToggle/menu-toggle-button.hbs
```diff
@@ -0,0 +1,16 @@
+{{#if menu-toggle--IsExpanded}}
```
mcoker, line 1: Can you wrap the contents of this partial in a menu toggle BEM element? It's generally bad practice for a component (BEM block) to also be a BEM element somewhere else. 


```
<div class="pf-c-menu-toggle__button{{#if menu-toggle-button--modifier}} {{menu-toggle-button--modifier}}{{/if}}"
  {{#if menu-toggle-button--attribute}}
    {{{menu-toggle-button--attribute}}}
  {{/if}}>
  // contents of this partial go here
</div>
```

src/patternfly/components/MenuToggle/menu-toggle.hbs
```diff
@@ -15,7 +15,9 @@
     pf-m-expanded
   {{/if}}
   {{#if menu-toggle--modifier}} {{menu-toggle--modifier}}{{/if}}"
-  type="button"
+  {{#unless menu-toggle--type}}
+    type="button"
+  {{/unless}}
   {{#if menu-toggle--IsExpanded}}
```
mcoker, line 21: Looks like we need to move this into the `{{#unless menu-toggle--type}}` conditional above - `aria-expanded` should only be on the plain button you've added.

mcoker (changes requested):

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -314,7 +322,7 @@
     height: 100%;
   }
 
-  &.pf-m-split-button {
+  @mixin menuToggleDiv {
```
mcoker, line 325: I don't know that you need/want all these styles to apply to the typeahead - it's probably just the first 2 lines

```
    --pf-c-menu-toggle__toggle-icon--MarginRight: 0;

    padding: 0; // pass padding to children
```

The other styles for the children I think are too specific to the split button to be considered shared and to share the same vars - keeping in mind it may be breaking to give the typeahead its own vars later as either the typeahead or split button change.

Also instead of using a mixin, I'd group the selectors you want to have these common styles - that's DRYer than this approach since sass will actually spit out this grouping of styles twice.

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -327,6 +335,21 @@
         padding-right: var(--pf-c-menu-toggle--m-split-button--first-child--PaddingRight);
       }
     }
+  }
+
+  &.pf-m-input-group {
```
mcoker, line 340: I'd probably just call this typeahead. It would be good to have a modifier for the typeahead variant. Input group also is a little confusing since we have both an input group and a text input group, and I think these styles are more applicable to the use of the text input group with typeahead than, say, any time the input group is in the menu toggle.

mcoker (changes requested):
Looking good! Left some comments, also looks like you've got some out of bounds action going on here with the arrows 

Screen Shot 2022-03-17 at 5 52 03 PM

src/patternfly/components/MenuToggle/menu-toggle.hbs
```diff
@@ -14,17 +14,22 @@
...
     pf-m-expanded
   {{/if}}
-  {{~#if menu-toggle--IsAction}}
-    pf-m-action
-  {{/if}}
-  {{#if menu-toggle--IsSplitButton}}
-    pf-m-split-button
-    {{~#if menu-toggle--IsDisabled}}
-      pf-m-disabled
+  {{#if menu-toggle--IsDiv}}
```
mcoker, line 17: nit - while the test is valid, I don't think this is necessary?

```suggestion
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -314,6 +322,17 @@
     height: 100%;
   }
 
+  &.pf-m-typeahead {
+    padding: var(--pf-c-menu-toggle--PaddingTop) var(--pf-c-menu-toggle--PaddingRight) var(--pf-c-menu-toggle--PaddingBottom) var(--pf-c-menu-toggle--PaddingLeft);
```
mcoker, line 326: this is the default, so it isn't doing anything. With the text-input-group's negative margin, I don't think you need it.

mcoker (changes requested):
couple of things - otherwise, lgtm!!

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -150,6 +150,12 @@
   --pf-c-menu-toggle__button__controls--MarginRight: var(--pf-global--spacer--sm);
   --pf-c-menu-toggle__button__controls--MarginLeft: var(--pf-global--spacer--sm);
 
+  // Input group margins
+  --pf-c-menu-toggle--c-text-input-group--MarginTop: calc(var(--pf-c-menu-toggle--PaddingTop) * -1);
```
mcoker, line 154:
```suggestion
  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginTop: calc(var(--pf-c-menu-toggle--PaddingTop) * -1);
```

mcoker (changes requested):
Just a couple of comments! 🧟‍♂️🔫

mcoker (changes requested):

mcoker (approved):
L🔥TM! Great work on this.

Issue #4586, opened 2022-01-05
Make sure that all full page demos are created from a common template


mcarrano:
Want to make sure that we are using a common template for all full page demos. Identify common elements like masthead, navigation, breadcrumbs, drawers, etc. so that all demos have a common look and feel for consistency.

PR #4674, opened 2022-02-08
chore(demos): created common template for full page demos

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4586

----

Updated the page and table demos to make sure it works and not turn this into a huge PR. Assuming the general structure works, we can follow up and update the rest.

This creates a bunch context parameters and templates that we can define as inline locally per example to override whatever part of the structure we want. I think it's pretty similar to the react component where you can set like `<Page header={Header} sidebar={Sidebar}>` and define your custom content externally.

Some thoughts:
* I think supporting a directory structure for these files would be good, but our workspace doesn't rebuild changed templates in nested dirs so I didn't do that here. I'd like to down the road though.
* The inline blocks could call includes to keep the example code shorter, but it's already shorter.
* Write some docs around usage.
* with ^, this could be its own demo/component that we don't include on the website.


srambach (approved):
Seems like a great approach. As you say, we'll want some good documentation on how to build demos. 🎉 👍

mattnolting (approved):
Nice work, LPTM!

PR #4683, opened 2022-02-10
chore(masthead): added masthead template

mattnolting:
closes #4579 
closes #4686 

Masthead template applied to Page and Masthead demos.

PR #4689, opened 2022-02-11
chore(context-selector): fixed broken include paths

mcoker:
fixes the context selector demos after an include was removed from https://github.com/patternfly/patternfly/pull/4683/


mattnolting (approved):
Nice catch, thanks!

PR #4696, opened 2022-02-21
feat(menu): added checkbox to menu

### LINKED ISSUE #4663: Menu: Add checkbox menu
Add a checkbox menu, so we can build a checkbox select using the menu and toggle.

cc @mcarrano

mattnolting:
closes #4663


src/patternfly/components/Menu/menu-check-input.hbs
```diff
@@ -0,0 +1,5 @@
+<input class="pf-c-menu__check-input{{#if menu-check-input--modifier}} {{menu-check-input--modifier}}{{/if}}"
```
mcoker, line 1: I wonder if instead of this, we could have a `__check` element that holds a standalone check component. IMO it would be great if we could use the check component everywhere so all checkbox enhancements, themes, etc are available in components. Seems like updating all of our use of single checkboxes in components would be a good breaking change update.

src/patternfly/components/Menu/menu-label.hbs
```diff
@@ -0,0 +1,17 @@
+<label class="pf-c-menu__label{{#if menu-label--modifier}} {{menu-label--modifier}}{{/if}}"
```
mcoker, line 1: I wonder if we could make the `__item` element the `label`, which would be more inline with how the other menu items work where the `__item` is the clickable thing. Is there a reason that wouldn't work? Something like this:
Screen Shot 2022-02-25 at 1 26 58 PM

mattnolting: Aside: have to add `role="menuitem"` to `.pf-c-menu__list-item` containing a checkbox.

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -1713,7 +1732,7 @@ import './Menu.css'
 | -- | -- | -- |
 | `role="menu"` | `.pf-c-menu__list` | Declares `.pf-c-menu__list` as a menu. |
 | `disabled` | `button.pf-c-menu__item` | When the menu item uses a button element, indicates that it is unavailable and removes it from keyboard focus. |
-| `role="menuitem"` | `.pf-c-menu__item` | Assigns `.pf-c-menu__item` as an option in a set of choices contained by a menu. |
+| `role="menuitem"` | `.pf-c-menu__item`, `.pf-c-menu__list-item (checkbox)` | Assigns `.pf-c-menu__item` as an option in a set of choices contained by a menu. |
```
mcoker, line 1735: seems like we would put the checkbox description outside of the backticks, or maybe call it out in the description cell?

```suggestion
| `role="menuitem"` | `.pf-c-menu__item`, `.pf-c-menu__list-item` (checkbox) | Assigns `.pf-c-menu__item` as an option in a set of choices contained by a menu. |
```

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -1731,9 +1750,11 @@ import './Menu.css'
 | `.pf-c-menu__content` | `<div>` | Initiates the menu content. Use for lists. **Required** |
 | `.pf-c-menu__list` | `<ul>` | Initiates the menu list. **Required** |
 | `.pf-c-menu__list-item` | `<li>` | Initiates the menu list item. **Required** |
-| `.pf-c-menu__item` | `<button>`, `<a>`, `<div>` | Initiates the menu item. **Required** |
+| `.pf-c-menu__item` | `<button>`, `<a>`, `<div>`, `<label>` | Initiates the menu item. **Required** |
 | `.pf-c-menu__item-main` | `<span>` | Initiates the menu item main container. **Required** |
 | `.pf-c-menu__item-text` | `<span>` | Initiates the menu item text. **Required** |
+| `.pf-c-menu__item-text` | `<span>` | Initiates the menu item text. **Required** |
```
mcoker, line 1756: dupe

```suggestion
```

src/patternfly/components/Menu/menu-item.hbs
```diff
@@ -1,17 +1,23 @@
...
+      {{/if}}
+    {{else}}
+      type="button"
+      {{#if menu-list-item--IsDisabled}}
+        disabled
+      {{/if}}
     {{/if}}
+    role="menuitem"
+  {{/unless}}
+  {{#if menu-list-item--IsCheckbox}}
```
mcoker, line 18: just a nit - if you do `{{#unless menu-list-item--IsCheckbox}}...{{/unless}}{{#if menu-list-item--IsCheckbox}}...{{/if}}`, that could be rewritten as `{{#if menu-list-item--IsCheckbox}}...{{else}}...{{/if}}`

src/patternfly/components/Menu/menu-item.hbs
```diff
@@ -23,5 +29,20 @@
       aria-expanded="false"
     {{/if}}
   {{/if}}>
-  {{> @partial-block}}
-</{{# if menu-item--IsLink}}a{{else if menu-item--type}}{{menu-item--type}}{{else}}button{{/if}}>
+  {{#if menu-list-item--IsCheckbox}}
+    {{#> menu-item-main}}
+      {{#> menu-item-check}}
+        {{#> check check--type="span" check--IsDisabled=menu-list-item--IsDisabled}}
```
mcoker, line 35: We should probably use the standalone modifier, though it seems to throw off the alignment a bit... wdyt?

```suggestion
        {{#> check check--type="span" check--modifier="pf-m-standalone" check--IsDisabled=menu-list-item--IsDisabled}}
```

src/patternfly/components/Menu/menu.scss
```diff
@@ -648,6 +655,15 @@
   margin-right: var(--pf-c-menu__item-icon--MarginRight);
 }
 
+// Check
+.pf-c-menu__item-check {
+  margin-right: var(--pf-c-menu__item-check--MarginRight);
+
+  .pf-c-check {
+    grid-template-columns: auto; // update pf-c-check grid-template-columns to grid-auto-columns: 1fr at breaking change
```
mcoker, line 663: if you use the standalone modifier, you shouldn't need this

mcoker (changes requested):
Love it! Left some comments, using the standalone modifier is the important one.

src/patternfly/components/Menu/menu.scss
mattnolting: I initially added `pf-m-standalone`, but that mod shifted the checkbox vertically. Are you seeing that?

mattnolting: I thought one fewer modifier and this `grid-template-columns` update was better than adding a modifier, then adding a margin update to that modifier.

mcoker: Gotcha, tho this is one of the use cases the standalone modifier was created for. When we have a breaking change, it would be good if we could replace our use of regular checkboxes/radios (like in selectable tables) with the standalone check/radio component. I haven't looked at why the alignment is off - is it something we could fix by modifying the standalone variation or modifying the `__item-check` element you added here? If you still think it's better as is, we can leave it and revisit later since this component is beta.

mcoker (approved):
L🎃TM - just one comment about the check component standalone variation. Not a blocker tho!

src/patternfly/components/Menu/menu.scss
mattnolting: @mcoker updated!

Issue #4687, opened 2022-02-10
Page - update centered content example


mcarrano:
**Is this a new component or an extension of an existing one?**
Page
**Describe the feature**
Update the [example for centered content](https://www.patternfly.org/v4/components/page/html#centered-content) here so it is more intuitive. See comment by @mcoker : https://github.com/patternfly/patternfly-react/pull/6893#issuecomment-1033961372 I believe this would make the example easier to understand.

PR #4698, opened 2022-02-23
chore(page): update wording in centered examples

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4687


mcarrano (commented):
@mcoker This is fine, but I thought the goal was to change the example to wrap the content in a card as discussed here: https://github.com/patternfly/patternfly-react/pull/6893#issuecomment-1033202022

mcoker:
@mcarrano when adding it, I realized adding a card makes it more of a demo, and we have a centered demo that can show centering with and without the card. Though it is a pretty simple - I'll add something about it being in a card to show how it works and that the card isn't part of the feature.

mcoker:
@mcarrano updated

mcarrano (approved):
Looks great. Thanks @mcoker !

mattnolting (approved):
Looks great, thanks!!

PR #4701, opened 2022-02-24
feat(menu): added image support

### LINKED ISSUE #4635: Dropdown and menu toggle updates
LGTM! Just one thing. I'm inclined to keep what's in this PR and we make a couple of changes in core to make the element that will hold an icon or image in the toggle work better.

_Originally posted by @mcoker in https://github.com/patternfly/patternfly-react/pull/6807#pullrequestreview-858808039_

mattnolting:
closes #4635 

@mcoker the positioning of the image is vertically centered with these updates, but the padding is reduced. Wondering if we should add a modifier `.pf-m-avatar` to add a bit of padding top/bottom, although it might be fine @mceledonia @mcarrano? 

Menu toggle - Icon: https://patternfly-pr-4701.surge.sh/components/menu-toggle#icon
Menu toggle - Avatar: https://patternfly-pr-4701.surge.sh/components/menu-toggle#with-image-and-text


mcoker (changes requested):
Nice, looks good. One nit about the docs, isn't necessary, but I'd say the description should either say "icon" to match the BEM element class or "icon/image" to denote you can use whatever you want there. 

Also looks like we have this CSS for the menu toggle image, but it's never used. I think we can remove that: `--pf-c-menu-toggle__image--MarginRight: var(--pf-global--spacer--sm);`

mcarrano:
> Wondering if we should add a modifier .pf-m-avatar to add a bit of padding top/bottom

Looks fine to me, although we might see it better if you used this styling for the avatar: https://www.patternfly.org/v4/components/avatar/html#bordered---light

mcoker:
I think you'll run into the padding thing with any larger icon/image. @mcarrano here it is with the updated avatar
Screen Shot 2022-02-24 at 4 27 27 PM

We may also combine these examples since it's just a variation of the same feature?

mcoker (approved):
LGTM!

mcarrano (approved):
Looks great. Thanks @mattnolting !

PR #4702, opened 2022-02-24
fix(select): fix active focus states bottom border

### LINKED ISSUE #4632: Select toggle focus/expanded state issue
The select toggle has a 1px bottom border on hover/focus/active/expanded, while the rest of our menu toggles have 1px hover, 2px focus/active/expanded. The select toggle should be updated to match the other toggles.

Select:
Screen Shot 2022-01-20 at 10 47 14 AM

Options menu (for comparison):

Screen Shot 2022-01-20 at 10 49 30 AM

cc @mcarrano @mmenestr seems like this might be a regression. Can you confirm the toggle treatment should match?

srambach:
Fixes #4632 

Used individual width variables instead of the combined width variable that's marked for removal at breaking change.

### DIFF
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -15,7 +15,7 @@
   --pf-c-select__toggle--before--BorderRightWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-select__toggle--before--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-select__toggle--before--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-select__toggle--before--BorderWidth: var(--pf-c-select__toggle--before--BorderTopWidth) var(--pf-c-select__toggle--before--BorderRightWidth) var(--pf-c-select__toggle--before--BorderBottomWidth) var(--pf-c-select__toggle--before--BorderLeftWidth); // remove at breaking change
+  --pf-c-select__toggle--before--BorderWidth: initial; // remove at breaking change
   --pf-c-select__toggle--before--BorderTopColor: var(--pf-global--BorderColor--300);
   --pf-c-select__toggle--before--BorderRightColor: var(--pf-global--BorderColor--300);
   --pf-c-select__toggle--before--BorderBottomColor: var(--pf-global--BorderColor--200);
@@ -292,6 +292,8 @@
   }
 
   &::before {
+    --pf-c-select__toggle--before--BorderWidth-base: var(--pf-c-select__toggle--before--BorderTopWidth) var(--pf-c-select__toggle--before--BorderRightWidth) var(--pf-c-select__toggle--before--BorderBottomWidth) var(--pf-c-select__toggle--before--BorderLeftWidth); // remove at breaking change
+
     position: absolute;
     top: 0;
     right: 0;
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Select/select.scss
```diff
@@ -300,7 +300,7 @@
     content: "";
     border-color: var(--pf-c-select__toggle--before--BorderTopColor) var(--pf-c-select__toggle--before--BorderRightColor) var(--pf-c-select__toggle--before--BorderBottomColor) var(--pf-c-select__toggle--before--BorderLeftColor);
     border-style: solid;
-    border-width: var(--pf-c-select__toggle--before--BorderWidth);
+    border-width: var(--pf-c-select__toggle--before--BorderTopWidth) var(--pf-c-select__toggle--before--BorderRightWidth) var(--pf-c-select__toggle--before--BorderBottomWidth) var(--pf-c-select__toggle--before--BorderLeftWidth);
```
mcoker, line 303: Hmm... technically this is breaking because a user could set `--pf-c-select__toggle--before--BorderWidth` to some custom width, and it wouldn't work after this change.  

What you have here is ideal, though we should probably try and avoid the edge case of a breaking change if we can.

The problem is we declared `--pf-c-select__toggle--before--BorderWidth` at the top level component scope using `--pf-c-select__toggle--before--BorderBottomWidth`, then we're redeclaring `--pf-c-select__toggle--before--BorderBottomWidth` in the toggle scope and never redeclaring `--pf-c-select__toggle--before--BorderWidth`.

To fix that problem above, we either need to:

1. Redeclare `--pf-c-select__toggle--before--BorderWidth` in the toggle scope wherever we set any of the border-bottom-widths.

```
  --pf-c-select__toggle--before--BorderWidth: var(--pf-c-select__toggle--before--BorderTopWidth) var(--pf-c-select__toggle--before--BorderRightWidth) var(--pf-c-select__toggle--before--BorderBottomWidth) var(--pf-c-select__toggle--before--BorderLeftWidth);
}
```

That wouldn't be breaking, but we end up with extra code. 

2. Move the lines where we redeclare `--pf-c-select__toggle--before--BorderBottomWidth` to the top level component scope. e.g., Take the code [in this block](https://github.com/patternfly/patternfly/blob/47dfeb3563962ec0452cd1ccc47efe1585d141a1/src/patternfly/components/Select/select.scss#L322-L325) and either change the selector to `@at-root .pf-c-select.pf-m-expanded {}` or nest it under `.pf-c-select` and change  the selector to `&.pf-m-expanded {}`.

This doesn't change the specificity so basic overrides should still work, though it does change the selector for the rules, so that could be breaking, too.

3. Move our declaration of `--pf-c-select__toggle--before--BorderWidth` into the toggle scope, but still let it be overridden by the consumer. We could do that by moving it and renaming it (e.g., `--BorderWidth-base`), then declaring `--pf-c-select__toggle--before--BorderWidth: initial;` at the component scope so that it will trigger a fallback. Then in the toggle, set the `::before`'s border width to `border-width: var(--pf-c-select__toggle--before--BorderWidth, var(--pf-c-select__toggle--before--BorderWidth-base));`. That allows `--pf-c-select__toggle--before--BorderWidth` to be set by the consumer to override the border width, but keeps our declarations of the border width vars in the toggle scope so our existing state overrides still work. And we can clean that up in a breaking change.

Here's a demo with the 3 ways commented out: https://codesandbox.io/s/cool-haze-osi3f9?file=/style.css

mcoker (changes requested):
Just one comment - this change may be OK, but if we can make it without removing the use of `--BorderWidth`, that would be ideal.

mcoker (approved):
It's beautiful!!! 😍

Screen Shot 2022-02-28 at 3 19 08 PM

PR #4706, opened 2022-02-28
fix(masthead): updated toggle borders

### LINKED ISSUE #4693: Dropdowns in masthead are missing default bottom border
Spotted in this PR when the dropdown is in an input group - https://github.com/patternfly/patternfly/pull/4691

mattnolting:
closes #4693

### DIFF
--- a/src/patternfly/components/Masthead/masthead.scss
+++ b/src/patternfly/components/Masthead/masthead.scss
@@ -241,13 +241,16 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   .pf-c-dropdown {
     --pf-c-dropdown__toggle--before--BorderTopColor: var(--pf-c-masthead--c-dropdown__toggle--before--BorderTopColor);
     --pf-c-dropdown__toggle--before--BorderRightColor: var(--pf-c-masthead--c-dropdown__toggle--before--BorderRightColor);
-    --pf-c-dropdown__toggle--before--BorderBottomColor: var(--pf-c-masthead--c-dropdown__toggle--before--BorderBottomColor);
     --pf-c-dropdown__toggle--before--BorderLeftColor: var(--pf-c-masthead--c-dropdown__toggle--before--BorderLeftColor);
 
     &.pf-m-full-height {
       --pf-c-dropdown__toggle--before--BorderRightColor: var(--pf-c-masthead--c-dropdown--m-full-height__toggle--before--BorderRightColor);
       --pf-c-dropdown__toggle--before--BorderLeftColor: var(--pf-c-masthead--c-dropdown--m-full-height__toggle--before--BorderLeftColor);
     }
+  }
+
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Masthead/masthead.scss
```diff
@@ -250,6 +249,10 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
     }
   }
 
+  :where(:not(.pf-c-input-group)) > .pf-c-dropdown {
```
mcoker, line 252: do you need the child combinator? also could these styles not just apply to the full height variation, or is there a use case of a regular/non-full-height dropdown that needs the bottom border removed?

```suggestion
  :where(:not(.pf-c-input-group)) .pf-c-dropdown {
```

mattnolting: > do you need the child combinator?

Yes, bc `--pf-c-dropdown__toggle--before--BorderBottomColor` belongs to `pf-c-dropdown`, although I couldn't solve needing `>` in the selector. Removing it negates the intended effect. Any ideas?

> also could these styles not just apply to the full height variation, or is there a use case of a regular/non-full-height dropdown that needs the bottom border removed?

@mcoker Here's the original design,  https://marvelapp.com/prototype/5cjajdg/screen/63256958 and there are more iterations showing the context selector styled the same way. As far as I know, this is the intended styling. Also, we can't use `pf-m-plain` as it removes the `active/focus` borders, maybe we need another modifier...

mcoker: > Removing it negates the intended effect

Ah, I see. It works like that, I guess it just introduces some fragility where the dropdown needing to be a direct child of the input group. 

Hmm... just to double check, @mmenestr @mcarrano @mceledonia can you confirm the user dropdown design? It currently looks like this:

Screen Shot 2022-03-01 at 6 25 12 PM

I'm wondering if it should look like one of these two:

Screen Shot 2022-03-01 at 6 28 06 PM
Screen Shot 2022-03-01 at 6 27 48 PM

Which would match the style of other menus like that in the masthead

Screen Shot 2022-03-01 at 6 28 53 PM

> maybe we need another modifier

Maybe so... otherwise we might need to add a bunch of exceptions to the styles like we are here, which could potentially get complicated down the road.

mmenestr:
@mcoker Makes sense to me for them all to be consistent! The only thing is that the example would be inconsistent with what's currently in product but I think that's ok, people can always take in the change later!

mceledonia:
Agreed, this should be the same.

mcoker (changes requested):
Looks great! Just need to add the full height modifier to the user dropdown now https://patternfly-pr-4706.surge.sh/components/masthead/html-demos/advanced-integration-with-menu-options/

mcoker (approved):
⭐️⭐️⭐️⭐️⭐️

Issue #4688, opened 2022-02-10
Wizard - description uses p tag, which heavily restricts description content


starpit:
**Describe the problem**
The WizardDescription component uses `<p>` as its wrapper for the content. This precludes the use of ... well almost everything... as description content. This is at odds with the type spec, which specifies that the `description` field can be any `React.ReactNode`.

For example, if you use a `<p>` or `<div>` as your `description`, you will get these console errors:

> `validateDOMNesting(...): <div> cannot appear as a descendant of <p>.`

**How do you reproduce the problem?**
https://codesandbox.io/s/nervous-pine-vmpzq

**Expected behavior**
I should be able to use any `React.ReactNode` as a `description`.

**Is this issue blocking you?**
List the workaround if there is one.

**Screenshots**
n/a

**What is your environment?**
 - OS: any
 - Browser: any
 - Version: 4.192.15

**What is your product and what release date are you targeting?**

**Any other information?**

PR #4708, opened 2022-02-28
chore(wizard): update docs/hbs for description to allow div

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4688

The plan is to add a `component` prop in the react component that will let it be either a `<p>` (default) or a `<div>` - then update it to be a `<div>` by default in our next breaking change.


srambach (approved):
LGTM 👍

Issue #4646, opened 2022-01-28
Sticky column has higher z-index than select


rvsia:
We are using table sticky column with a table filter, but sticky table has higher z-index than the select, so the select is hidden behind the first columns:

![Screenshot 2022-01-28 at 15 44 42](https://user-images.githubusercontent.com/32869456/151567808-5662df2c-7696-40e9-bf83-d568badfee9a.png)

`pf-c-dropdown__menu`  `var(--pf-c-dropdown__menu--ZIndex);` `200` (same applies for `pf-c-select__menu`)
`pf-c-table__sticky-column` `var(--pf-c-table__sticky-column--ZIndex);` `300`

**How it should look**

![Screenshot 2022-01-28 at 15 53 30](https://user-images.githubusercontent.com/32869456/151568489-ea20bdd0-4c97-4fc3-80c3-a4d149687b5f.png)

rvsia:
The same issue applies for table actions dropdown

![image](https://user-images.githubusercontent.com/32869456/153183645-8312e5a3-b2af-4749-b2f9-907946d6ab52.png)

mcoker:
@rvsia the reason for that is because we don't want elements that will scroll behind a sticky column/header to pop out on top of the sticky column/header. We're trying to avoid something like this:

https://user-images.githubusercontent.com/35148959/153511237-7628814b-eb32-4b25-bb12-777eb9f9f8a0.mp4

I believe in these cases if you know you want the menu to appear on top of the elements around it, you can use the `menuAppendTo` prop to append the element to the `<body>` element.

Another way to do it could be to assign a higher `z-index` to those particular elements, or a lower `z-index` to the element adjacent to the menu that you want the menu to appear on top of, creating a stacking context. For example, using a `z-index` that is lower than the dropdown menu (eg, `z-index: 199`) on the entire table that you want the dropdown in the toolbar to appear on top of will create a stacking context for the table, and preserve any existing `z-index`es in the table.

Will one of those solutions work?

mcoker:
@rvsia Thanks, great suggestions! We'll be working on this next sprint so hopefully we can improve on what's there currently.

PR #4709, opened 2022-02-28
docs(table): added docs around z-index conflicts with use of sticky columns/headers

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4646

Is this too wordy? I imagine we'll want to add something like this to the react docs, too.

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -2970,7 +2970,13 @@ Add a [width modifier](#width-modifiers) to `thead th` to limit string length or
 
 ## Sticky table modifiers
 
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Table/examples/Table.md
```diff
@@ -2970,6 +2970,8 @@ Add a [width modifier](#width-modifiers) to `thead th` to limit string length or
...
 
+Sticky table headers and columns have a higher z-index than the z-index used for menus (dropdown, select, etc). The intent is that the contents of a scrollable table will scroll under the sticky header/column, including any expanded menus. However, there may be use cases where a menu needs to appear on top of a sticky header/column, such as a menu placed inside of a toolbar above a table with a sticky header. In these cases it's advised to append the menu to an element outside of the table (e.g., the  <code>body</code> element) so that the menu has a higher stacking context than - and can appear on top of - sticky header/columns as well as appear outside of any scrollable content in the table. Typically this is achieved via the <code>menuAppendTo</code> prop in non-composable react components. Another approach may be to manipulate the z-index of the menu manually. In the case where the menu is outside of the table, another approach may be to assign the entire table a z-index lower than the z-index of the menu, creating a lower stacking context for the table compared to the menu outside of the table, preserving the stacking context of the elements inside of the table.
```
srambach, line 2973: What about breaking up the paragraph a bit with a list of the options? 
I'd also add "Note:" and set off the z-index and a couple other tweaks. IDK if the first option should be called "recommended" ?

Note: Sticky table headers and columns have a higher `z-index` than the `z-index` used for menus (dropdown, select, etc). The intent is that the contents of a scrollable table will scroll under the sticky header/column, including any expanded menus. However, there may be use cases where a menu needs to appear on top of a sticky header/column, such as a menu placed inside of a toolbar above a table with a sticky header. 
There are a few ways this can be handled:
- Use the <code>menuAppendTo</code> prop in non-composable react components to append the menu to an element outside of the table (e.g., the <code>body</code> element) so that the menu has a higher stacking context than - and can appear on top of - sticky header/columns as well as appear outside of any scrollable content in the table. (Recommended) 
- Manipulate the z-index of the menu manually.
- In the case where the menu is outside of the table, another approach may be to assign the entire table a z-index lower than the `z-index` of the menu. This creates a lower stacking context for the table compared to the menu outside of the table, while preserving the stacking context of the elements inside of the table.

mattnolting (approved):
Love it, LPTM 🚢

Issue #4652, opened 2022-01-18
FormSelect: Arrow icon is stored in Data: and doesn't appear


AlexisLessard:
**Describe the issue. What is the expected and unexpected behavior?**
On our app, we activated _Content-Security-Policy_. However, we noticed that after that, the arrow in our `FormSelect` dropdowns weren't showing up anymore

![image](https://user-images.githubusercontent.com/10947093/149984032-34d0eedf-c5c3-4109-931b-5b9eb13c837e.png)

The following warning was appearing in the chrome devtools console:

Refused to load the image '<URL> xmlns='<URL>' viewBox='0 0 320 512'%3E%3Cpath fill='%23urrentColor' d='M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z'/%3E%3C/svg%3E' because it violates the following Content Security Policy directive: "img-src 'self'".

The problem is fixed when changing the `image-src` section in our index.html to this:   `img-src 'self' data:;`, As far as I've red, it is impossible to narrow the filter down anymore than that.

Problem is that apparently, **this is an insecure setting**, according to the [Mozzila documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src). 

**Please provide the steps to reproduce. Feel free to link CodeSandbox or another tool.**
Content Security Tag in our index.html:
```
<meta
      http-equiv="Content-Security-Policy"
      content="
        script-src 'unsafe-inline' https: 'nonce-someNonce' 'strict-dynamic';
        script-src-elem 'self';
        object-src 'none';
        base-uri 'none';
        require-trusted-types-for 'script';
        img-src 'self';
        manifest-src 'self';
        connect-src <urls>;
      "
    />
```

Using a `FormSelect` inside of a `FormGroup`:
```
<FormGroup
      label={t(`${creationProjetOpenshiftPrefix}.roles`)}
      fieldId="projectRoleBindings"
    >
          <FormSelect
            width="50%"
            aria-label="projectRoleDropdown"
            value={roleBindingToSet.role}
            onChange={(roleToSet) => {
              setRoleBindingToSet({
                ...roleBindingToSet,
                role: roleToSet as Roles,
              });
            }}
          >
            {Object.values(Roles).map((role) => (
[...]

PR #4711, opened 2022-03-01
feat(form-control): added icon sprite variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4652

----

This is one way of doing it. Another could be to host separate images for each icon but this approach seems cleaner. Some thoughts...

* Something like this would get rid of the hurdles we run into around needing to embed SVG data in a data URI.
* Another option could be to enhance the text input group to support states/icons like this, and suggest using that instead.
* If we end up going this route, I think we would want to be able to import global vars into the SVG - right now I'm using hex values for the colors.
* I'm not 100% that an asset like this is served from the repo like this in an application. I think my main question is - if we update this asset in a release, is it updated for anyone using it? Or might someone copy this asset somewhere in their site structure to serve it - then need to re-copy when we release a new version?


mattnolting (approved):
Awesome.

PR #4713, opened 2022-03-01
feat(menu-toggle): added split button

### LINKED ISSUE #4612: MenuToggle: Need split button variant - like Dropdown
in order to implement all dropdown variants use composable solutions, we need a split button menu toggle like
https://www.patternfly.org/v4/components/dropdown/html#split-button-checkbox
https://www.patternfly.org/v4/components/dropdown/html#split-button-checkbox-with-toggle-text
https://www.patternfly.org/v4/components/dropdown/html#split-button-action
https://www.patternfly.org/v4/components/dropdown/html#split-button-primary-action

Note: It doesn't need to be a MenuToggle variant - It could be a button variant or a layout of some kind. It just needs to be a composable stand alone solution for a split button

mattnolting:
closes #4612 

There are some differences between this and dropdown.
- The target are for `pf-c-menu-toggle__button` is larger than that of `pf-c-dropdown`. Ideally, it would be at least `44px` wide, but that looks a bit odd. 
- `pf-c-dropdown` doesn't support `pf-m-secondary`, `menu-toggle` does and presents an example.
- The interactions are the same

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -111,6 +111,43 @@
   --pf-c-menu-toggle--m-full-height--focus--after--BorderBottomWidth: var(--pf-global--BorderWidth--xl);
   --pf-c-menu-toggle--m-full-height--active--after--BorderBottomWidth: var(--pf-global--BorderWidth--xl);
 
+  // Split button
+  --pf-c-menu-toggle--m-split-button--BackgroundColor: var(--pf-global--BackgroundColor--100);
+
+  // Split button, child
+  --pf-c-menu-toggle--m-split-button--child--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-menu-toggle--m-split-button--child--disabled--Color: var(--pf-global--Color--dark-200);
+  --pf-c-menu-toggle--m-split-button--child--disabled--BackgroundColor: var(--pf-global--disabled-color--300);
+
+  // Split button, action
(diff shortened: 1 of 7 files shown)


src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -276,6 +313,159 @@
...
+      --pf-c-menu-toggle--m-split-button--m-action--child--BorderLeftColor: transparent;
+
+      &::before,
+      &::after {
+        content: none;
+      }
+    }
+
+    // disabled styles for children
+    > [disabled],
```
mcoker, line 450:
```suggestion
    > :disabled,
```

mcoker (commented):
Looks fantastic. Only thing I'm not 100% on is the click area on the checkbox here.

Screen Shot 2022-03-02 at 2 22 45 PM
Screen Shot 2022-03-02 at 2 22 41 PM

----

In the dropdown we just reduce the padding on the sides that touch

Screen Shot 2022-03-02 at 2 23 54 PM

Screen Shot 2022-03-02 at 2 23 49 PM

----

Meaning we might do something like

Screen Shot 2022-03-02 at 2 25 54 PM

Screen Shot 2022-03-02 at 2 25 57 PM

----

Or this if we want to match what's currently in the dropdown - it removes the margin on the controls element when it's in the `__button`

Screen Shot 2022-03-02 at 2 26 31 PM

Screen Shot 2022-03-02 at 2 26 27 PM

mcoker (approved):

PR #4714, opened 2022-03-01
chore(form): improve accessible label on form elements in examples/demos

### LINKED ISSUE #4453: Improve accessibility of form help buttons in examples/demos
Currently all the demos and examples of a field level help button are read as "More info." A user navigating form controls could get many items that are just "more info" so it would be nice to show how to include the label so that it says something like "more info for name."

@jessiehuff @nicolethoen @jgiardino WDYT?

srambach:
Fixes #4453 

This adds `aria-described` by on the form field "more info" buttons  and improves the `aria-label` text so that core matches patternfly-react.

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -13,7 +13,7 @@ cssPrefix: pf-c-form
       {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"') required="true"}}
         Name
       {{/form-label}}
-      {{> form-group-label-help}}
+      {{> form-group-label-help form-group-label-help--aria-label="More information for name field" form-group-label-help--aria-describedby=(concat form--id form-group--id)}}
     {{/form-group-label}}
     {{#> form-group-control}}
       {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form--id form-group--id '" name="' form--id form-group--id '" required')}}{{/form-control}}
@@ -40,7 +40,7 @@ cssPrefix: pf-c-form
       {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"')}}
         Information
       {{/form-label}}
-      {{> form-group-label-help}}
+      {{> form-group-label-help form-group-label-help--aria-label="More information for information field" form-group-label-help--aria-describedby=(concat form--id form-group--id)}}
     {{/form-group-label}}
     {{#> form-group-control}}
       {{#> form-control controlType="textarea" form-control--attribute=(concat 'type="text" id="' form--id form-group--id '" name="' form--id form-group--id '" aria-label="Textarea example"')}}{{/form-control}}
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Form/examples/Form.md
```diff
@@ -75,7 +75,7 @@ cssPrefix: pf-c-form
       {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"') required="true"}}
         Name
       {{/form-label}}
-      {{> form-group-label-help}}
+      {{> form-group-label-help form-group-label-help--aria-label="More information for name field" form-group-label-help--aria-describedby=(concat form--id form-group--id)}}
```
mcoker, line 78: this is unrelated, but I noticed this example and the regular horizontal example both use `form--id="form-horizontal"` - wold you mind updating this one to be unique?

src/patternfly/components/Form/examples/Form.md
```diff
@@ -255,7 +255,7 @@ cssPrefix: pf-c-form
   {{#> form-group form-group--id="-label1"}}
     {{#> form-group-label}}
       {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"') required="true"}}Label 1{{/form-label}}
-      {{> form-group-label-help}}
+      {{> form-group-label-help form-group-label-help--aria-label="More information for Label 1 field" form-group-label-help--aria-describedby=(concat form--id form-group--id)}}
```
mcoker, line 258: nit - looks like you're mixing the case of some labels. they should probably all just be sentence case

```suggestion
      {{> form-group-label-help form-group-label-help--aria-label="More information for label 1 field" form-group-label-help--aria-describedby=(concat form--id form-group--id)}}
```

mcoker (changes requested):
Looks great, but it looks like there are a bunch of other places we use this help button that may need updating.

I wonder if instead of a manual update, we could use the `form-group--id` field better. We would need to take the `-` out of it and either insert the hyphen manually when it's used to create an `id`, `for`, or whatever - ie, change code like `id="' form--id form-group--id '"` to `id="' form--id '-' form-group--id '"`. We could also just leave the hyphen out - pretty sure it's just to generate `id` and `for`, and as long as it's unique, doesn't matter to me what it is. Tho it's more readable with the hyphens. That may be even more complicated though.

src/patternfly/components/Form/examples/Form.md
```diff
@@ -13,7 +13,7 @@ cssPrefix: pf-c-form
       {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"') required="true"}}
         Name
       {{/form-label}}
-      {{> form-group-label-help}}
+      {{> form-group-label-help form-group-label-help--aria-label="More information for name field" form-group-label-help--aria-describedby=(concat form--id form-group--id)}}
```
mattnolting, line 16: Looks like https://github.com/patternfly/patternfly/pull/4714/files#diff-604e9154a4176f06e862bcb8493641aea5f98be03d3ea0e9935ca8935b229fdaR11 needs to be something like `{{#> form-group form-group--id=(concat form--id "-name")}}` or added to `form-group.hbs`

srambach:
@mcoker I added the capability to use the `form-group--id` in the .hbs. However when I started converting all the examples, I noticed that there are quite a few where the id would not result in an understandable name when read. (e.g. clientid and some that are further hyphenated because of nesting) Instead of creating inconsistency in the examples, I left all using the original method of passing in the entire help text string. Without having a specified string, it will fall back to use "More information for `form-group--id`" and then just the text "More information."

mattnolting (approved):
LPTM! 🎉

mcoker (approved):
Perfect!

PR #4715, opened 2022-03-02
feat(description-list): added description list demo

### LINKED ISSUE #1921: Description list demo
https://marvelapp.com/prototype/6e72b59

Create a demo to show how a definition list can be built in PatternFly (using the Description list component).

---

From the design looks like we can consider creating 4 things:
* Basic demo - not sure how useful that is? Might serve as a good baseline for other demos.
* Editable demo - this could also go in the inline edit component. We had an example for an editable `<dl>` that's commented out - probably makes sense to re-enable that with the new description list component. Then we could simply show the list term/item as editable with the edit icon as shown in the designs.
* Description list in a drawer - unless this shows something unique, I don't think this is useful. Pretty much anything can go in a drawer.
* Complex content demo - the demos with tables in the list item are probably good to show.

mattnolting:
closes #1921 

@mcarrano Inline edit isn't yet available via React outside of `pf-c-table`, so I think it's best to add those demo features when it's available globally.

### DIFF
--- a/src/patternfly/demos/DescriptionList/examples/DescriptionList.md
+++ b/src/patternfly/demos/DescriptionList/examples/DescriptionList.md
@@ -404,7 +404,7 @@ cssPrefix: pf-d-description-list
                   {{/description-list-text}}
                 {{/description-list-term}}
                 {{#> description-list-description}}
-                  {{#> table table--id="service-address" table--grid="true" table--modifier="pf-m-grid-md pf-m-compact" table--attribute='aria-label="Service address" style="--pf-c-table--BackgroundColor: transparent"'}}
+                  {{#> table table--id="service-address" table--grid="true" table--modifier="pf-m-grid-md pf-m-compact" table--attribute='aria-label="Service address"'}}
                     {{#> table-thead}}
                       {{#> table-tr}}
                         {{#> table-th table-th--attribute='scope="col"'}}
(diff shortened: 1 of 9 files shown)


mcarrano (commented):
This is looking good @mattnolting . I agree regarding the edit in-place example. I do have a couple of question for @mceledonia , however.

1- When a description list is placed in a page as in these demos, should it be on a white or gray background? The designs show white, however to get that, would we place this inside a card?

2- The drawer example feels a bit dense to me vs the designs, but I can't say why. Anything different that you would recommend @mceledonia ?

Here is the design link: https://marvelapp.com/prototype/6e72b59/screen/85219622

mceledonia (approved):

mattnolting:
@mcarrano @mceledonia I've updated the demos to show several content color combinations.

src/patternfly/demos/DescriptionList/examples/DescriptionList.md
```diff
@@ -0,0 +1,507 @@
...
+                {{/description-list-description}}
+              {{/description-list-group}}
+              {{#> description-list-group}}
+                {{#> description-list-term}}
+                  {{#> description-list-text}}
+                    Service port mapping
+                  {{/description-list-text}}
+                {{/description-list-term}}
+                {{#> description-list-description}}
+                  {{#> table table--id="service-port" table--grid="true" table--modifier="pf-m-grid-md pf-m-compact" table--attribute='aria-label="Service address" style="--pf-c-table--BackgroundColor: transparent"'}}
```
mcoker, line 446:
```suggestion
                  {{#> table table--id="service-port" table--grid="true" table--modifier="pf-m-grid-md pf-m-compact" table--attribute='aria-label="Service address"'}}
```

src/patternfly/demos/DescriptionList/examples/DescriptionList.md
```diff
@@ -0,0 +1,507 @@
...
+          {{#> grid-item}}
+            {{#> description-list}}
+              {{#> description-list-group}}
+                {{#> description-list-term}}
+                  {{#> description-list-text}}
+                    Service address
+                  {{/description-list-text}}
+                {{/description-list-term}}
+                {{#> description-list-description}}
+                  {{#> table table--id="service-address" table--grid="true" table--modifier="pf-m-grid-md pf-m-compact" table--attribute='aria-label="Service address" style="--pf-c-table--BackgroundColor: transparent"'}}
```
mcoker, line 407:
```suggestion
                  {{#> table table--id="service-address" table--grid="true" table--modifier="pf-m-grid-md pf-m-compact" table--attribute='aria-label="Service address"'}}
```

mcoker (approved):
LPTM!! Do you need to set the table bg color? Doesn't seem like it's necessary since it's against a white background.

PR #4721, opened 2022-03-03
feat(dashboard): added dashboard demo

### LINKED ISSUE #4659: Dashboard demo
Create a dashboard demo page that shows how to use cards to create responsive dashboard layouts. Should include layouts for a full page dashboard and for a dashbar as defined in the design mocks here: https://marvelapp.com/prototype/d642i6g/screen/84550919  Can use placeholder images for charts where applicable as we do for card demos. These should be replaced with actual charts as we move to React.

mattnolting:
closes #4659 

Preview: https://patternfly-pr-4721.surge.sh/demos/dashboard/html-demos/basic/


mcarrano (commented):
I think this looks great @mattnolting . The way this card responds as I shrink the screen is a little odd.

Screen Shot 2022-03-03 at 5 16 43 PM

But that was the only issue I saw. Would like to have @mceledonia and @doruskova also take a look.

doruskova:
@mattnolting @mcarrano It looks great! I think that it works as expected. Just wanted to ask if the section "Recommendation by category" will be an image or not? It doesn't have to look exactly the same. What about https://www.patternfly.org/v4/charts/pie-chart#basic-with-right-aligned-legend or is there any other technical reason?

https://user-images.githubusercontent.com/32618096/156745066-66132621-d980-4431-a14f-d005830ac47c.mp4

mattnolting:
> @mattnolting @mcarrano It looks great! I think that it works as expected. Just wanted to ask if the section "Recommendation by category" will be an image or not? It doesn't have to look exactly the same. What about https://www.patternfly.org/v4/charts/pie-chart#basic-with-right-aligned-legend or is there any other technical reason?

@doruskova Great question, Core doesn't have access to charts, so we have to use images as placeholders. The React version will include live, interactive charting.

mattnolting:
> I think this looks great @mattnolting . The way this card responds as I shrink the screen is a little odd.
> 
> Screen Shot 2022-03-03 at 5 16 43 PM
> 
> But that was the only issue I saw. Would like to have @mceledonia and @doruskova also take a look.

@mcarrano nice catch, updated

mcarrano (approved):
Loos good to me. Thanks @mattnolting !

nicolethoen:
It's so beautiful

I'm sure products are going to eat this up. Which is the only reason i hesitate to give it a pass even though tab order is not quite what i'd expect.

it feels like a case of 'how strict do we want to be?'. The main concern is keyboard or screen reader users getting lost or disoriented by navigation order not matching presentation order. But the groups are well labelled, so that will definitely help.

So, yeah.... i'm not sure. @jessiehuff, anything else you would add?

src/patternfly/demos/Card/templates/card-template-status.hbs
```diff
@@ -0,0 +1,143 @@
...
+        {{#> notification-drawer-group notification-drawer-group--IsExpanded=card-demo--notication-drawer--IsOpen}}
+          {{#> notification-drawer-group-toggle}}
+            {{#> notification-drawer-group-toggle-title}}
+              {{#> l-flex}}
+                {{#> notification-drawer-group-toggle-title}}
+                  {{#> l-flex l-flex--modifier="pf-m-space-items-sm" l-flex--attribute='style="row-gap: var(--pf-global--spacer--xs);"'}}
+                    {{#> l-flex-item l-flex-item--modifier="pf-m-spacer-md"}}
+                      <span>Notifications</span>
+                    {{/l-flex-item}}
+                    {{#> label label--id="default-red-icon" label--modifier="pf-m-red"}}
```
mcoker, line 102: These labels could be in a label group. That's how it's setup in the horizontal grid card demo - https://www.patternfly.org/v4/components/card/html-demos#horizontal-grid-collapsed

mcoker (approved):
This is really great!!! Love all of the intricacies with the layouts, multiple `<dd>`s in the `<dl>`, notification drawer group... good stuff!

Issue #4726, opened 2022-03-04
Bug - Menu toggle - self referencing var failing react build


mcoker:
https://github.com/patternfly/patternfly/blob/3009a0813ebe7fe7a75a66d15ed84fca4c60cda9/src/patternfly/components/MenuToggle/menu-toggle.scss#L340

PR #4727, opened 2022-03-04
fix(menu-toggle): fixed self referencing var

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4726


evwilkin (approved):
LGTM - thanks for the quick fix!

evwilkin (approved):
LGTM

PR #4728, opened 2022-03-09
fix(build): updated build to include themes directory

### LINKED ISSUE #4700: Bug - Dark Theme - themes directory isn't included in PatternFly package.
**Describe the problem**

Users who want to @import `.scss` encounter a compilation error, as the themes directory is missing in package.

evwilkin:
Closes #4700 

This PR makes the build process aware of the new `themes` directory so it is included in the `dist` directory to be shipped as part of the package.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,9 +13,10 @@ const { build } = require('theme-patternfly-org/scripts/cli/build');
 const sassFiles = [
   './src/patternfly/patternfly*.scss',
   './src/patternfly/base/patternfly*.scss',
-  './src/patternfly/{components,layouts,patterns,utilities,themes}/**/*.scss',
-  './src/patternfly/components/**/themes/**/*.scss',
-  '!./src/patternfly/**/_all.scss'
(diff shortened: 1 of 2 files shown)


nicolethoen (approved):

nicolethoen (changes requested):
I'm noticing when I try to build that all the new dark theme related .css files are empty...

mattnolting:
Confirmed, I too am seeing empty `.css` files

evwilkin:
@nicolethoen @mattnolting thanks for the callout - I've updated to remove the empty CSS files from being generated.

FYI - I confirmed these files were producing empty CSS files correctly, as the source SASS files only contain mixins & SASS variable declarations but no actual style rules to compile down into CSS.  Each of these SASS files all roll up into the `src/patternfly/patternfly-theme-dark.scss` file which is compiled into `patternfly-theme-dark.css`, and all these theme `.scss` files are still shipped in the package.

nicolethoen (approved):

mcoker (approved):
LGTM!

mattnolting (approved):
LPTM 👍

PR #4730, opened 2022-03-16
refactor(search-input): converted to text-input-group

### LINKED ISSUE #4614: Refactor Search Input to use Text Input Group under the hood

mattnolting:
closes #4614

### DIFF
--- a/src/patternfly/components/Badge/badge.hbs
+++ b/src/patternfly/components/Badge/badge.hbs
@@ -2,5 +2,10 @@
   {{#if badge--attribute}}
     {{{badge--attribute}}}
   {{/if}}>
-  {{> @partial-block}}
+  {{#if badge--text}}
+    {{badge--text}}
+  {{/if}}
+  {{#if @partial-block}}
+    {{> @partial-block}}
+  {{/if}}
 </span>
--- a/src/patternfly/components/Badge/badge.hbs
+++ b/src/patternfly/components/Badge/badge.hbs
@@ -4,8 +4,7 @@
   {{/if}}>
   {{#if badge--text}}
     {{badge--text}}
-  {{/if}}
-  {{#if @partial-block}}
+  {{else if @partial-block}}
     {{> @partial-block}}
   {{/if}}
 </span>
--- a/src/patternfly/components/SearchInput/examples/SearchInput.md
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.md
@@ -82,14 +82,14 @@ import './SearchInput.css'
   {{/text-input-group-main}}
   {{#> text-input-group-utilities}}
     {{> badge badge--modifier="pf-m-read" badge--text="1 / 3"}}
-    {{#> text-input-group-element-group}}
+    {{#> input-group}}
       {{#> button button--modifier="pf-m-plain" button--attribute='disabled aria-label="Next"'}}
         <i class="fas fa-angle-up fa-fw" aria-hidden="true"></i>
       {{/button}}
       {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Next"'}}
         <i class="fas fa-angle-down fa-fw" aria-hidden="true"></i>
       {{/button}}
-    {{/text-input-group-element-group}}
+    {{/input-group}}
     {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Clear input"'}}
       <i class="fas fa-times fa-fw" aria-hidden="true"></i>
     {{/button}}
--- a/src/patternfly/components/SearchInput/examples/SearchInput.md
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.md
@@ -311,7 +311,6 @@ import './SearchInput.css'
   {{#> text-input-group-main text-input-group-main--modifier="pf-m-icon"}}
     {{#> text-input-group-text}}
       {{> text-input-group-icon}}
-      {{> text-input-group-text-input text-input-group-text-input--hint-text="app" }}
       {{> text-input-group-text-input}}
     {{/text-input-group-text}}
   {{/text-input-group-main}}
--- a/src/patternfly/components/SearchInput/examples/SearchInput.md
+++ b/src/patternfly/components/SearchInput/examples/SearchInput.md
@@ -151,7 +151,7 @@ import './SearchInput.css'
 {{#> panel panel--modifier="pf-m-raised"}}
   {{#> panel-main}}
     {{#> panel-main-body}}
-      {{#> form form--id="text-input-group-advanced-search-input-form-example"}}
+      {{#> form form--id="text-input-group-advanced-search-input-expanded-legacy-form-example"}}
         {{#> form-group form-group--id="-username"}}
           {{#> form-group-label}}
             {{#> form-label form-label--attribute=(concat 'for="' form--id form-group--id '"')}}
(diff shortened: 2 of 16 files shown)


nicolethoen (commented):
It's either confusing or helpful - not sure which - to have two search inputs in each example. 
I _think_ it's unnecessary to keep the old search inputs that use the old search input classes. There are not going to be the equivalent react implementations. React will use Text input group completely under the hood. 

I think you may want to either call out the use of a text input group under the hood in the docs and/or list the text input group class names in the api docs rather than the search input classes.

mattnolting:
@nicolethoen Oh yes, I'm sorry I should've explained. I'll remove the old version once we've had a chance to compare old vs new.

src/patternfly/components/SearchInput/examples/SearchInput.md
```diff
@@ -11,31 +11,140 @@ import './SearchInput.css'
...
+{{#> text-input-group}}
+  {{#> text-input-group-main text-input-group-main--modifier="pf-m-icon"}}
+    {{#> text-input-group-text}}
+      {{> text-input-group-icon}}
+      {{> text-input-group-text-input text-input-group--value="John Doe"}}
+    {{/text-input-group-text}}
+  {{/text-input-group-main}}
+  {{#> text-input-group-utilities}}
+    {{> badge badge--modifier="pf-m-read" badge--text="1 / 3"}}
+    {{#> text-input-group-element-group}}
```
mcoker, line 85: this could be an input group, too, since it's super generic

src/patternfly/components/Badge/badge.hbs
```diff
@@ -2,5 +2,10 @@
   {{#if badge--attribute}}
     {{{badge--attribute}}}
   {{/if}}>
-  {{> @partial-block}}
+  {{#if badge--text}}
+    {{badge--text}}
+  {{/if}}
+  {{#if @partial-block}}
```
mcoker, line 8: just a nit, but if the plan is to use `--text` or `@partial-block`, and not together, having `@partial-block` as an `else` condition will keep this partial from spitting stuff out if it's included as a child of a parent HBS partial that takes `@partial-block` that is intended for the parent. Written the way it is in the PR we'd need to do `{{#> badge badge--text="foo"}}{{/badge}}`. With an `else` condition here we could just do `{{> badge badge-text="foo"}}`

nicolethoen:
> @nicolethoen is the goal in react to keep `<SearchInput>` the way it is and suggest folks use `<TextInputGroup>` going forward for a search input? Or are we going to rework the search input to use text input group under the hood?

The later - we are going to rework the search input to use the text input group under the hood.

mcoker:
> The later - we are going to rework the search input to use the text input group under the hood.

@nicolethoen @mcarrano I guess I'm just wondering where these examples should go. Using the text input group from this PR, core wouldn't have a search input if we aren't supporting the old search input markup. Long term I'm assuming we wouldn't list the search input component for core, and would instead list these "search input" examples under the text input group component examples/demos, and react would have the search input component - similar to core's form control which is used for react's `<TextInput/TextArea/FormSelect>`?

We could do that now, and add a banner or note saying we're deprecating that component in core in favor of the text input group.

Or we could update these examples so the text input group is used for all of them, then rename the existing search input component examples to "legacy" as we've done for other components?

mcarrano:
I don't the idea of getting rid of the component "search input group" even if it is actually something built from a text input group. I think Search input group is something people will look for. If Search input group becomes something more like a demo, that's OK, but I still want it listed in the nav.

src/patternfly/components/SearchInput/examples/SearchInput.md
```diff
@@ -112,6 +305,66 @@ import './SearchInput.css'
     {{/search-input-menu-list}}
   {{/search-input-menu}}
 {{/search-input}}
+
+<div class="ws-example-wrapper" style="margin-top: 200px;">
+{{#> text-input-group text-input-group--id="autocomplete" text-input-group--value="app"}}
+  {{#> text-input-group-main text-input-group-main--modifier="pf-m-icon"}}
+    {{#> text-input-group-text}}
+      {{> text-input-group-icon}}
+      {{> text-input-group-text-input text-input-group-text-input--hint-text="app" }}
```
mcoker, line 314: Since this is just autocomplete, you shouldn't need to render the hint text input. The [react search input autocomplete demo](https://www.patternfly.org/v4/components/search-input/react-demos#search-with-autocomplete) only renders it when hint text is shown, and the [core search input autocomplete](https://www.patternfly.org/v4/components/search-input/html#autocomplete) core [advanced expanded autocomplete search input and text input group examples](https://patternfly-pr-4730.surge.sh/components/search-input) don't render it either.

mcoker (commented):
This LGTM, just one little nit about rendering the hint text input, though it isn't a blocker - we can follow up on it.

@mcarrano @mmenestr what do you think about the example presentation here with search input example first, and text input group second? At a minimum, maybe we should yank the existing search input examples out and label them "legacy"? Or maybe just put sub-headings in the examples that identify one or both of them as search input and/or text input group?

src/patternfly/components/TextInputGroup/text-input-group-control.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-text-input-group__control{{#if text-input-group-control--modifier}} {{text-input-group-control--modifier}}{{/if}}"
```
mcoker, line 1: Oh, I forgot to mention - is this used? Doesn't look like it, unless I missed it.

src/patternfly/components/TextInputGroup/text-input-group-group.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="pf-c-text-input-group__group{{#if text-input-group-group--modifier}} {{text-input-group-group--modifier}}{{/if}}"
```
mcoker, line 1: Could you add this to the text input group docs?

nicolethoen:
The search input on the react side comes with a lot of functionality wired up for the consumer. I am open to combining them - but if we want a consistent experience across products, and we are able to ship the SearchInput implementation of the TextInputGorup, I feel like that's pretty powerful.

mcoker (changes requested):

mcarrano (changes requested):
> @mcarrano @mmenestr what do you think about the example presentation here with search input example first, and text input group second? At a minimum, maybe we should yank the existing search input examples out and label them "legacy"? Or maybe just put sub-headings in the examples that identify one or both of them as search input and/or text input group?

Yes, I think we should yank the existing search input examples and either label them legacy or not show them at all. Is there a a reason someone might still need these? I agree that the current presentation is confusing.

mcoker:
>  Is there a a reason someone might still need these?

Probably just as a reference if that's what they're currently using and can't/don't want to use the text input group. I would still be in favor of eventually just including the search input examples (on the core side anyways) under the text input group examples and/or demos, since it doesn't fit our format of a component. For example, we list all of the component docs/elements/classes/etc at the bottom of the component page, and if we did that here with the text input group, we would be listing text input group docs - which already exist on the text input group component page.

src/patternfly/components/TextInputGroup/text-input-group-group.hbs
mattnolting: I would, but there don't seem to be any text input group docs to add to. Feels like a followup to document all of text input group.

mattnolting:
@mcoker @mcarrano Moved the legacy examples into a legacy block

nicolethoen (approved):

mcoker (approved):
LGTM! 🚀

Issue #4729, opened 2022-03-11
Bug - Accordion - links at bottom of bordered accordion items not clickable


nicolethoen:
**Describe the problem**
When using the bordered accordion, an anchor tag at the bottom of the accordion content body is not clickable.

**How do you reproduce the problem?**
Codesandbox to reproduce
https://codesandbox.io/s/amazing-moser-sr3ibb?file=/index.tsx

**Is this issue blocking you?**
It seems that an anchor tag, when adding the `pf-c-button` and `pf-m-link` classes, it does work. So buttons as links work, but not just vanilla anchor tags.

**What is your environment?**
latest version of PF (4.183.1)

**What is your product and what release date are you targeting?**
Customer portal
https://access.redhat.com/support/cases

Reported on behalf of @adithyaakrishna

PR #4740, opened 2022-03-21
fix(accordion): made links at bottom of bordered item clickable

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4729

Added a link in the first bordered example that wasn't clickable before.

### DIFF
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -140,6 +140,7 @@
             right: 0;
             bottom: 0;
             left: 0;
+            pointer-events: none;
             content: "";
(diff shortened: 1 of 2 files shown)


evwilkin:
@mcoker this looks good to me, I see the content within the bordered content area (aka the new link) is clickable now whereas it wasn't without the `pointer-events` rule updated.

I did notice this example has no "Item four", do we want to update that as part of this PR?  If not just LMK and I'll approve 👍

mcoker:
@evwilkin nice catch! updated.

nicolethoen:
@adithyaakrishna - wanted to be sure this update for your issue is now in a PR :)

evwilkin (approved):
LGTM

Issue #4725, opened 2022-03-01
Tree view: Checkbox variant has misleading hover state


nicolethoen:
The checkbox variant works differently than other Tree view variants because for all the others, clicking anywhere in the row expands or collapses the children. If the row is displaying a hover state, and you click, the row will toggle.

For checkbox variants, if you click on the expand/collapse button, the row toggles. If you click on the checkbox OR the text, the checkbox is selected. If you click on the space to the right of the text, nothing will happen. Nothing happens BUT the hover styles are still there (along with a pointer cursor) making the user think that something will happen if they click... 

I believe in this case, the row should expand/collapse if the user is clicking in the empty space beside the text - like it would for the non-checkbox variants.

![Mar-01-2022 15-53-48](https://user-images.githubusercontent.com/26305082/156247238-e06e7345-bfd8-4c70-bda7-0d69aa36e875.gif)

mcoker:
With this issue and https://github.com/patternfly/patternfly/issues/4724, we decided to update tree view nodes so that on any row that is selectable and has a toggle (checkbox or the "tree navigation" from #4724), only the toggle arrow and anything to the left of it should be clickable to expand the row, and anything to the right (except for actions) should select the row. Also as part of this update, we should update the toggle icon to be color-200 by default and color-100 when hovered/focused/expanded. Use a plain button there if that makes sense.

Issue #4724, opened 2022-03-04
TreeView - Ability to select parent nodes without changing expand state


daggmano:
For our use case, we need to be able to trigger onSelect for any node (leaf or parent) without changing the expand state. A parent node can then be expanded using the caret to the left of the item. Is this possible, or can this be requested as an improvement?

mcarrano:
@daggmano Thanks for raising this issue. This has been raised before and we are planning to address. It's on our roadmap for Q2. Let us know if you have any specific dates when this is needed.

daggmano:
@mcarrano As with most things, we needed it yesterday ;) For now I've managed to get around it using jQuery and a custom `onSelect` method, and also added the option to have 'disabled' items (specific format of the `id` field and CSS matching to attribute substrings), but it's a hack. Would be great to have this baked in, so very glad it's on the roadmap.

PR #4744, opened 2022-03-22
fix(tree-view): fixed non clickable areas in node

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4725

---

@nicolethoen I'm wondering if this is best handled with JS and leaving the click handler on the node element - it's disabled when the node has a checkbox: https://github.com/patternfly/patternfly-react/blob/0ae1c2fe74b54b12236dcc7d64f9b3827d457d8d/packages/react-core/src/components/TreeView/TreeViewListItem.tsx#L187

I'm guessing the reason it's disabled is because a click on the check/text will bubble up to the node, triggering the node to be expanded? If so, could we use `stopPropagation` on the input and [`label.pf-c-tree-view__text`](https://github.com/patternfly/patternfly-react/blob/0ae1c2fe74b54b12236dcc7d64f9b3827d457d8d/packages/react-core/src/components/TreeView/TreeViewListItem.tsx#L187)? 

I think that's my preferred way of doing it. I can't have a button (or interactive element) wrap the checkbox + label, so to handle this with markup, I would need to add another element that comes after the check + label to make that area clickable, and I don't think we want to do that. WDYT? Also I'm tagging you since you created the issue - feel free to tag another JS dev if you'd like.

Also I fixed this - if the node has actions, the margin between the node and actions isn't clickable. This orange spot: 

Screen Shot 2022-03-21 at 8 04 16 PM

And fixed this - the orange spot between the check and text isn't clickable:

Screen Shot 2022-03-21 at 8 17 40 PM

### DIFF
--- a/src/patternfly/components/TreeView/tree-view.scss
+++ b/src/patternfly/components/TreeView/tree-view.scss
@@ -11,9 +11,10 @@ $pf-c-tree-view--MaxNesting: 10;
   // Node
   --pf-c-tree-view__node--PaddingTop--base: var(--pf-global--spacer--sm);
   --pf-c-tree-view__node--PaddingTop: var(--pf-c-tree-view__node--PaddingTop--base);
-  --pf-c-tree-view__node--PaddingRight: var(--pf-global--spacer--sm);
+  --pf-c-tree-view__node--PaddingRight: var(--pf-global--spacer--lg);
   --pf-c-tree-view__node--PaddingBottom: var(--pf-global--spacer--sm);
   --pf-c-tree-view__node--PaddingLeft: var(--pf-c-tree-view__node--indent--base);
+  --pf-c-tree-view__node--last-child--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-tree-view__node--Color: var(--pf-global--Color--100);
   --pf-c-tree-view__node--m-current--Color: var(--pf-global--link--Color);
   --pf-c-tree-view__node--m-current--FontWeight: var(--pf-global--FontWeight--bold);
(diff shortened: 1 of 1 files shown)


nicolethoen (approved):
LGTM and makes sense.
I think the only thing I'll need to double check is whether any of the aria attributes need to be modified or moved around to be sure they are attached to the correct inter-actable elements. It's likely hard to know that without playing with it on the React side....

mcoker:
@nicolethoen cool!

@mcarrano @mmenestr 2 questions:

* Currently the expand arrow is always black - should we update that to change colors like the table, tree table, data list, etc? (`color-200` default, `color-100` hover/focus/expanded)
* With this change, and the suggestion about the expand arrow colors, should we add a visual affordance that clicking on the checkbox/label will select the check, and clicking anywhere else in the row will toggle the row's expanded state? Changing the arrow color when the row will expand/collapse (and not changing the color when hovering the checkbox/label) is one affordance. Should there be something else? 
  * I have the same question about the other issue where you can select the node without expanding it - https://github.com/patternfly/patternfly/issues/4724
  * One option could be to make the text blue when hovering/focusing on a part of the node that will select the node (not expand it), and (optionally) turning the text blue/bold when the node is selected, like we do with current/non-expandable nodes.
  * 

#### Hover
Screen Shot 2022-03-22 at 2 07 09 PM

#### Selected
Screen Shot 2022-03-22 at 2 07 27 PM

#### Existing "current" styling
Screen Shot 2022-03-22 at 2 40 29 PM

mcarrano:
> Yes, would probably be best to match the grey/black styling for the arrow! So grey by default, and black on hover/click. But we should do that for all of tree view then, not just the tree view with checkbox.

I agree. Sounds like this was an earlier oversight.

> This one is a bit more tough. I don't think we should turn the text blue on hover, because idk that we do that anywhere else, but turning the text blue on select could be a good idea - and would match what we do in the other tree view variations. My only concern is when many many things are selected- would that be way too much blue? @mcarrano what do you think?

We had discussed this previously in design and decided not to provide any additional affordance as we do not elsewhere in our menus. See https://github.com/patternfly/patternfly-design/issues/1124#issuecomment-1058313622

mattnolting (approved):
LPTM 👍

mcoker:
After discussing with @mmenestr @nicolethoen @mcarrano we're going to close this and restructure the tree view a bit, including the clickable area for the checkbox variant to be inline with the work we'll do with tree navigation in https://github.com/patternfly/patternfly/issues/4724. Outlined that in a comment in the original issue - https://github.com/patternfly/patternfly/issues/4725#issuecomment-1076785582

PR #4750, opened 2022-03-23
feat(menu-toggle): updated typeahead variant layout/spacing

jeffpuzzo:
Follow-up to #4673, where the hierarchy of elements has been changed in split button and typeahead variants to be `controls -> button -> icon` versus what was previously `button -> controls -> icon`. Padding and margins were adjusted accordingly.

Associated issue: #4666

React issue that needs typeahead toggle: https://github.com/patternfly/patternfly-react/issues/6888

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -151,10 +151,8 @@
   --pf-c-menu-toggle__button__controls--MarginLeft: var(--pf-global--spacer--sm);
 
   // Typeahead
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginTop: calc(var(--pf-c-menu-toggle--PaddingTop) * -1);
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginRight: calc(var(--pf-c-menu-toggle--PaddingRight) * -1);
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginBottom: calc(var(--pf-c-menu-toggle--PaddingBottom) * -1);
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginLeft: calc(var(--pf-c-menu-toggle--PaddingLeft) * -1);
+  --pf-c-menu-toggle--m-typeahead--c-menu-toggle__button--PaddingRight: var(--pf-global--spacer--md);
+  --pf-c-menu-toggle--m-typeahead--c-menu-toggle__button--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-menu-toggle--m-typeahead--c-text-input-group__utilities--c-button--PaddingRight: var(--pf-global--spacer--sm);
 
   position: relative;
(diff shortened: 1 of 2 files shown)


src/patternfly/components/MenuToggle/menu-toggle-controls.hbs
```diff
@@ -1,6 +1,13 @@
...
   {{#if menu-toggle-controls--attribute}}
     {{{menu-toggle-controls--attribute}}}
-  {{/if}}>
+  {{/if}}
+  {{#if menu-toggle--IsDisabled}}
+    disabled
+  {{/if}}
+  {{#if menu-toggle-button--IsDisabled}}
+    disabled
+  {{/if}}
```
mcoker, line 10: You shouldn't need this - doesn't look like it's getting styles from the `disabled` attribute, and `disabled` isn't valid on a `<span>`. It's only valid on [form elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) like inputs, text areas, buttons, etc.

```suggestion
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -119,7 +119,8 @@
   --pf-c-menu-toggle--m-split-button--child--disabled--Color: var(--pf-global--Color--dark-200);
   --pf-c-menu-toggle--m-split-button--child--disabled--BackgroundColor: var(--pf-global--disabled-color--300);
   --pf-c-menu-toggle--m-split-button--first-child--PaddingRight: var(--pf-global--spacer--sm);
-  --pf-c-menu-toggle--m-split-button--last-child--PaddingLeft: 0;
+  --pf-c-menu-toggle--m-split-button--last-child--PaddingLeft: var(--pf-global--spacer--md);
+  --pf-c-menu-toggle--m-split-button--last-child--PaddingRight: var(--pf-global--spacer--md);
```
mcoker, line 123: This is now targeting the controls element. This makes the button inside tiny -

Screen Shot 2022-03-28 at 5 29 17 PM
Screen Shot 2022-03-28 at 5 29 22 PM

I'm wondering if we need to use the controls element in this variation. I think the controls is just to separate the toggle icon from the text, though styling a generic controls element instead of the toggle icon would allow us to put other things in there besides the toggle icon and keep the spacing tied to a generic container. Doesn't seem like that applies in this variation where the only thing in the button is a single icon, since it doesn't need to be spaced apart from anything.

@mattnolting wdyt?

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -151,10 +152,7 @@
   --pf-c-menu-toggle__button__controls--MarginLeft: var(--pf-global--spacer--sm);
 
   // Typeahead
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginTop: calc(var(--pf-c-menu-toggle--PaddingTop) * -1);
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginRight: calc(var(--pf-c-menu-toggle--PaddingRight) * -1);
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginBottom: calc(var(--pf-c-menu-toggle--PaddingBottom) * -1);
-  --pf-c-menu-toggle--m-typeahead--c-text-input-group--MarginLeft: calc(var(--pf-c-menu-toggle--PaddingLeft) * -1);
+  --pf-c-menu-toggle--m-typeahead--c-menu-toggle__controls--PaddingRight: var(--pf-global--spacer--md);
```
mcoker, line 155: This spacer should be on the last action. Shouldn't be necessary if we add the padding back to the button inside of it.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -188,5 +188,7 @@
   .pf-c-button {
     --pf-c-button--PaddingRight: var(--pf-c-text-input-group__utilities--c-button--PaddingRight);
     --pf-c-button--PaddingLeft: var(--pf-c-text-input-group__utilities--c-button--PaddingLeft);
+
+    display: flex;
```
mcoker, line 192: Why do you need this? It throws off the button size - it should fill the height of the component.

Screen Shot 2022-03-28 at 5 46 01 PM

Screen Shot 2022-03-28 at 5 46 08 PM

mcoker (changes requested):

src/patternfly/components/TextInputGroup/text-input-group.scss
jeffpuzzo: The icon was more center-aligned with display: flex.  There is a very slight difference, maybe 1px worth. If you toggle the property on and off using the PR surge link I think you'll notice the same. Can you confirm?

nicolethoen:
@mcoker I took another look at this, and even though the questions were not answered directly, Jeff did address each of your comments. Would you and @mattnolting like to give this another review? I am happy to make updates if more are necessary.

mattnolting (approved):
LGTM 💪

mcoker (approved):

Issue #4747, opened 2022-03-23
Add custom critical icon to a PF icons set


doruskova:
Regarding this issue https://github.com/patternfly/patternfly-design/issues/1138
I created an icon that presents a critical total risk value for Insights Advisor. 

- SVG file here [critical-risk.svg.zip](https://github.com/patternfly/patternfly/files/8331998/critical-risk.svg.zip)
- Updated Icon exporter [icon-exporter.sketch.zip](https://github.com/patternfly/patternfly/files/8332007/icon-exporter.sketch.zip)
- AI file with an icon [critical-risk-icon.ai.zip](https://github.com/patternfly/patternfly/files/8332012/critical-risk-icon.ai.zip)
<img width="174" alt="Snímek obrazovky 2022-03-23 v 11 14 28" src="https://user-images.githubusercontent.com/32618096/159676132-71068e3d-67bd-4bda-ab4d-f885208a05fa.png">

@mcarrano @mceledonia I need someone who can review it and let me know that the SVG file is good to go so that I can open the PR to update Icon exporter and patternfly-icons.zip

mcarrano:
@doruskova following up on our conversation today, I think we should also add the other risk icons being used from font awesome (i.e., important, moderate, and any others) to our pf-icon set so they show up on the Icons page. Can you identify what those are and we should be able to handle as part of this issue. Those will also need to be added to the official PF icon set as part of the icon exporter sketch file. Thanks.

doruskova:
@mcarrano Here's the list of mentioned icons from Fontawesome :) and updated Icon exporter with a zip file. It includes all icons that we discussed. 

SVGs for 3 other icons  - [SVGs.zip](https://github.com/patternfly/patternfly/files/8350797/SVGs.zip)
Updated Icon exporter - [icon-exporter.sketch.zip](https://github.com/patternfly/patternfly/files/8351535/icon-exporter.sketch.zip)
Updated zip file - [patternfly-icons.zip](https://github.com/patternfly/patternfly/files/8352095/patternfly-icons.zip)
- Important total risk - [Fontawesome link](https://fontawesome.com/v5/icons/angle-double-up?s=solid)
- Moderate total risk - [Fontawesome link](https://fontawesome.com/v5/icons/equals?s=solid)
- Low total risk - [Fontawesome link](https://fontawesome.com/v5/icons/angle-double-down?s=solid)

mcarrano:
[...]

PR #4758, opened 2022-03-25
feat(icon): added critical-risk pficon

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4747


mcarrano (commented):
@mcoker other than seeing that you did add this to the files attached in this PR, is there anything else I can or should do to review this. So far, looks good as far as I can tell.

mcoker:
@mcarrano just verify it looks OK on the icon page - https://patternfly-pr-4758.surge.sh/icons

mcarrano:
@mcoker OK. Forgot about that page. Yes, I see it and it looks good. But I do have another question- In your comment here: https://github.com/patternfly/patternfly/issues/4747#issuecomment-1079276419, you noted that the other three icons were already present in the icon set. If that's the case, how come they don't show up on the Icons page referenced above?

mcoker:
@mcarrano the icons on that page are from the "pficon" icon font - the other icons are part of Font Awesome solid. I could probably update that page to include our bundled FAS icons, too, if you think that would be helpful.

mcarrano:
@mcoker that's probably not necessary since it's not a page that appears publicly on the website. I will approve.

mcarrano (approved):
Thanks, @mcoker !

doruskova:
@mcoker Thank you a lot!

PR #4765, opened 2022-04-01
feat(divider): add horizontal/vertical breakpoint support to divider

### LINKED ISSUE #4720: Divider - add breakpoint support for horizontal/vertical change
Extension of divider 

**Describe the feature**
Several of our layout orientations can be changed at breakpoints, dividers need the ability to adjust as well

Default (horizontal):

Screen Shot 2022-03-03 at 11 58 52 AM

(flex shifting from `row` to `column`)
Divider interrupts the layout shift

Screen Shot 2022-03-03 at 11 58 59 AM

Vertical mod: 
Dividers are invisible

Screen Shot 2022-03-03 at 11 59 10 AM

Screen Shot 2022-03-03 at 11 59 17 AM

**Any other information?**

jenny-s51:
Closes #4720


mcoker (changes requested):
The vertical modifier looks good! But we're going to need a horizontal modifier, too, so a user can switch between vertical and horizontal at the supported breakpoints.

jenny-s51:
Thank you @mcoker ! I've updated this PR with a horizontal modifier and demo.

src/patternfly/components/Divider/divider.scss
```diff
@@ -67,6 +55,25 @@ $pf-c-divider--spacer-map: build-spacer-map("none", "xs", "sm", "md", "lg", "xl"
...
+
+        &::after {
+          flex-basis: calc(var(--pf-c-divider--m-vertical--after--FlexBasis) - var(--pf-c-divider--after--Inset));
+          width: var(--pf-c-divider--m-vertical--after--Width);
+        }
+      }
+
+      &.pf-m-horizontal#{$breakpoint-name} {
+        display: var(--pf-c-divider--Display);
+        flex-direction: row;
```
mcoker, line 75: While this works, I think we're going to need to reset all of the properties that the vertical modifier is changing, just so there are no issues with the way it behaves currently. If you look at the [Horizontal on lg example](https://patternfly-pr-4765.surge.sh/components/divider#horizontal-on-lg), the divider container is too tall because its `height` is set to `inherit`, since that came from `.pf-m-vertical`, and wasn't reset with `.pf-m-horizontal`.

Often what we do in this case is create `--base` vars for the default/base (horizontal) style properties, then assign those vars as the values for any default vars. We will also create placeholder vars for properties that the vertical variation changes, even if we didn't have to declare anything for that property initially in the default divider. We may do that for `height`, for example.

```
--pf-c-divider--Display--base: flex;
--pf-c-divider--Height--base: auto;
--pf-c-divider--Display: var(--pf-c-divider--Display--base);
--pf-c-divider--Height: var(--pf-c-divider--Height--base);
--pf-c-divider--m-horizontal--Display: inline-flex;
// --pf-c-divider--m-horizontal--Height: inherit; - you actually can't do this - `inherit` is a special word and will try to inherit `--pf-c-divider--m-horizontal--Height`, not evaluate to `inherit` where `var(--pf-c-divider--m-horizontal--Height)` is used

display: var(--pf-c-divider--Display); // you don't actually need this, just showing as an example. The line `@include pf-hidden-visible(var(--pf-c-divider--Display));` declares this for us.

&.pf-m-vertical[-on-{breakpoint}] {
  --pf-c-divider--Display: var(--pf-c-divider--m-vertical--Display);
  --pf-c-divider--Height: var(--pf-c-divider--m-vertical--Height, inherit); // means if `--pf-c-divider--m-vertical--Height` isn't defined, which it isn't, fall back to `inherit`. This allows the user to specify `--pf-c-divider--m-vertical--Height`, too, if they want.
}

&.pf-m-horizontal[-on-{breakpoint}] {
  --pf-c-divider--Display: var(--pf-c-divider--Display--base);
  --pf-c-divider--Height: var(--pf-c-divider--Height--base);
}
```

(This is unverified, so it may have errors but hopefully conveys the point!)

mcoker (changes requested):

src/patternfly/components/Divider/divider.scss
jenny-s51: Thank you for your feedback here @mcoker ! I've been investigating. Per your suggestion, I tried setting these base variables. It seems like the height of the container is coming from https://github.com/patternfly/patternfly/pull/4765/files#diff-7f3eddc914cec5fd7f6b03001bc43bd31c2ca2de891a5b6dc90e5565c0fe11b2R2 -- where `height` must be set to `3rem` in order for the vertical divider to render.

Does your suggested method still apply? Changing `--pf-c-divider--Height` like this doesn't seem to affect the height of `ws-example` once there's a switch from vertical to horizontal. Am I understanding correctly? What would you recommend?

mcoker: @jenny-s51 I think the height on the `.ws-preview` container is fine, we just don't want the horizontal divider's height to be set to `inherit` like it's set with a vertical divider. The `.ws-preview` having a fixed height with a horizontal divider is OK though. One thing that I think may help the example(s) that switch between vertical/horizontal using breakpoint modifiers would be to see if we can remove `display: flex` from the `.ws-preview` container in `Divider.css`. I did that locally and the vertical divider looked OK on my end, but I didn't test in other browsers, so you'd want to test that in FF, Safari, and Chrome if you give that a shot.

src/patternfly/components/Divider/examples/Divider.css
```diff
@@ -1,7 +1,5 @@
-#ws-core-c-divider-vertical .ws-preview-html,
-#ws-core-c-divider-vertical-inset-medium .ws-preview-html,
-#ws-core-c-divider-vertical-md-inset-no-inset-on-md-lg-inset-on-lg-sm-inset-on-xl .ws-preview-html {
+[id*="ws-core-c-divider-vertical"],
+[id*="ws-core-c-divider-horizontal"] .ws-preview-html {
   height: 3rem;
-  display: flex;
   align-items: center;
-}
+}
```
mattnolting, line 5:
```suggestion
}

```

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -64,7 +74,7 @@ The divider renders as an `<hr>` by default. It is possible to make the divider
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-c-divider` | `<hr>`, `<li>`, `<div>` | Defines the divider component. |
-| `.pf-m-vertical` | `.pf-c-divider` | Modifies the divider component from horizontal to vertical. This modifier requires that the parent has an explicit or implicit height, or has a flex or grid based layout parent. |
+| `.pf-m-vertical{-on-[breakpoint]}` | `.pf-c-divider` | Modifies the divider component to vertical. This modifier requires that the parent has an explicit or implicit height, or has a flex or grid based layout parent. Modifies the divider back to horizontal at an optional breakpoint. |
 | `.pf-m-inset-{none, sm, md, lg, xl, 2xl, 3xl}{-on-[sm, md, lg, xl, 2xl]}` | `.pf-c-divider` | Modifies divider padding/inset to visually match padding of other adjacent components. |
```
mattnolting, line 78: Looks like we need to add documentation for the horizontal variant as well.

mattnolting:
@jenny-s51 since you're switching between two layouts, I thought I'd mention this approach. https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Masthead/masthead.scss#L4 It's a new approach to setting default layouts settings with a mixin and switching between alternate layouts.

src/patternfly/components/Divider/examples/Divider.md
jenny-s51: Updated!

jenny-s51:
Thank you for your review and feedback @mattnolting ! I incorporated your comments into this PR and updated the scss loop using the [mixin approach that you linked -- super helpful](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Masthead/masthead.scss#L4)! 

When you have a moment, please kindly let me know if these changes are what you had in mind 🙂

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,16 +1,40 @@
...
+
+  &::after {
+    flex-basis: calc(var(--pf-c-divider--m-vertical--after--FlexBasis) - var(--pf-c-divider--after--Inset));
+    width: var(--pf-c-divider--m-vertical--after--Width);
+  }
+}
 
-  // end - update during breaking change
 
-  --pf-c-divider--Display: flex;
```
mcoker, line 13: Looks like we lost the ability to set this and change the divider display, which would be a breaking change. We can work on it tomorrow and try and get it in.

mcoker (changes requested):

src/patternfly/components/Divider/divider.scss
jenny-s51: Thank you @mcoker , that sounds like a good plan!

src/patternfly/components/Divider/examples/Divider.md
```diff
@@ -7,64 +7,88 @@ cssPrefix: pf-c-divider
 import './Divider.css'
 
 ## Examples
+
 ### hr
+
 ```hbs
-{{> divider}}
+{{> divider divider--modifier="pf-m-horizontal"}}
```
mcoker, line 14: I don't think we want to use `.pf-m-horizontal` on its own anywhere since it isn't necessary. We also use the divider throughout the site in other components/demos that wouldn't have `.pf-m-horizontal`, so it could be confusing.

You could even argue excluding `.pf-m-horizontal` as a class on its own by wrapping this line in a conditional that only includes the media query code if `($breakpoint != "base")` https://github.com/patternfly/patternfly/blob/7ab7699daf915fdb158dd78c5af38837daa38f6b/src/patternfly/components/Divider/divider.scss#L87-L90

mcoker (changes requested):

src/patternfly/components/Divider/divider.scss
```diff
@@ -1,64 +1,102 @@
...
+  }
+}
+
+.pf-c-divider {
+  // Base
+  --pf-c-divider--BorderWidth--base: var(--pf-global--BorderWidth--sm);
+  --pf-c-divider--BorderColor--base: var(--pf-c-divider--BackgroundColor); // update at breaking change
+  --pf-c-divider--Height: var(--pf-c-divider--BorderWidth--base); // remove at breaking change
+  --pf-c-divider--BackgroundColor: var(--pf-global--BorderColor--100); // remove at breaking change
+  --pf-c-divider--after--Height: var(--pf-c-divider--Height);
```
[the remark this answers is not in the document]
jenny-s51: Nice catch, good call 👍

src/patternfly/components/Divider/examples/Divider.md
jenny-s51: Sounds good - updated!

mcoker (approved):
LGTM!!! Great work!!

PR #4773, opened 2022-04-06
fix(form-control): fixed placeholder variation menu item color

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4772

When a form select has the placeholder modifier to turn the selected item text grey in a collapsed select menu, in chrome on linux (and FF on mac, probably other places, too), the placeholder color is applying to the items in the expanded menu, too, and they should not.

These screenshots are from FF on https://www.patternfly.org/v4/components/form-select

#### Non-placeholder - the non-disabled items are black text
Screen Shot 2022-04-06 at 3 22 40 PM

#### Placeholder - all of the items are set to the placeholder color
Screen Shot 2022-04-06 at 3 22 19 PM

### DIFF
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -44,7 +44,7 @@ import './FormControl.css'
   <option value="Other">Other</option>
 {{/form-control}}
 <br><br>
-{{#> form-control controlType="select" form-control--modifier="pf-m-placeholder" form-control--attribute='id="select-standard" name="select-standard" aria-label="Standard select example"'}}
+{{#> form-control controlType="select" form-control--modifier="pf-m-placeholder" form-control--attribute='id="select-placeholder-enabled" name="select-placeholder-enabled" aria-label="Placeholder enabled select example"'}}
   <option value="" selected>Please choose</option>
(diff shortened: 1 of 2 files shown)


mcarrano (commented):
Not sure if I'm looking at the right thing, but seems like all of the items still have placeholder styling.

Screen Shot 2022-04-06 at 6 33 53 PM

mcoker:
@mcarrano are you looking at this example? https://patternfly-pr-4773.surge.sh/components/form-control#select

It looks OK to me 

Screen Shot 2022-04-06 at 5 43 10 PM

mcarrano:
I think I was confused because your description references Form Select.

evwilkin:
@mcoker I'm seeing the issue described fixed (tested in FF on mac) but the selected text remains greyed out, is that correct?
Screen Shot 2022-04-07 at 9 26 45 AM

mcoker:
> the selected text remains greyed out, is that correct?

@evwilkin yep, the HTML select doesn't have native support for placeholder styling, so we added a class that needs to be applied manually depending on the selection. You can see that in the react example - https://www.patternfly.org/v4/components/form-select#validated

mattnolting (approved):
LGTM 🚢

Issue #4770, opened 2022-04-06
Bug - Card - Dropdowns in a selected card appear behind unselected cards


wise-king-sullyman:
**Describe the problem**
Dropdowns in a selected card appear behind unselected cards.

**How do you reproduce the problem?**
1. go to the selectable cards React example: https://www.patternfly.org/v4/components/card#selectable
2. select the first card
3. click on the kebab to open the dropdown

**Expected behavior**
I would expect the dropdown to be in front of any other cards so that it is fully readable regardless of the selection state of the card.

**Is this issue blocking you?**
Not blocking, but it may present an issue with finding a long-term solution for patternfly/patternfly-react#6798 depending on the changes needed.

**Screenshots**
![image](https://user-images.githubusercontent.com/7115053/162040208-138fd820-de87-4819-affe-a42232068666.png)

**What is your environment?**
 - OS: macOS 11.6.5
 - Browser: Chrome, Firefox, and Safari
 - Version: I'm unsure which version this is asking about, PF version? Current versions for each browser?

**What is your product and what release date are you targeting?**
N/A
**Any other information?**

PR #4780, opened 2022-04-07
fix(card): fix stacking context issue with selectable raised cards

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4770

@wise-king-sullyman this is a fun one! There are certain properties in CSS that create a thing called a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). In short, an element with a stacking context is basically the base z-layer for "positioned" (the `position` property being set to anything other than the default, which is `static) elements. The property here that is creating a new stacking context is `transform` on `.pf-c-card.pf-m-selectable-raised`. It's the property that moves the selected card up a bit to "raise" it. https://github.com/patternfly/patternfly/blob/6e04cbd09b4e9bc7f09fc9a39831ca9d64f773d7/src/patternfly/components/Card/card.scss#L201

What that means is the dropdown menu's `z-index` in a selected card is positioned relative to the card it's in, instead of whatever it was positioned relative to before (probably the `html` element). And since cards that come after the selected card will appear on top of cards that come before it, that means the expanded dropdown from the selected card appears behind cards that come after it. 

There are a few ways we could fix that. I'm giving the selected card a `z-index` we've reserved for this kind of use case, where we have simple adjacent elements that need a low `z-index` to appear on top of one another. Since that `z-index` is lower than the dropdown menu's default `z-index`, that means that a dropdown from a non-selected card that comes before a selected card will still appear on top of the selected card, since the menu's `z-index` is higher.

Another way to fix it could be to use `top` or `bottom` to "raise" the card (`top: -1em` is the same as `transform: translateY(-1em)`), which would resolve it since `top` and `bottom` don't create new stacking contexts. The downsize is that we animate the "raised" effect, and animating the `transform` property is quite a bit more performant than `top` or `bottom`. 

#### Before
Screen Shot 2022-04-07 at 4 54 35 PM

#### After
Screen Shot 2022-04-07 at 4 53 48 PM

#### After w/ the selected card after one with a dropdown expanded
Screen Shot 2022-04-07 at 4 55 45 PM

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -57,6 +57,7 @@
   --pf-c-card--m-selectable-raised--m-selected-raised--TranslateY: var(--pf-c-card--m-selectable-raised--m-selected-raised--TranslateY--base);
   --pf-c-card--m-flat--m-selectable-raised--m-selected-raised--TranslateY: calc(var(--pf-c-card--m-selectable-raised--m-selected-raised--TranslateY--base) + var(--pf-c-card--m-flat--BorderWidth));
   --pf-c-card--m-rounded--m-selectable-raised--m-selected-raised--TranslateY: calc(var(--pf-c-card--m-selectable-raised--m-selected-raised--TranslateY--base) + var(--pf-c-card--m-rounded--BorderRadius));
+  --pf-c-card--m-selectable-raised--m-selected-raised--ZIndex: var(--pf-global--ZIndex--xs);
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (approved):
This is interesting!

The approach here makes sense based on what you've explained. Thanks for getting this figured out so quickly!

mattnolting (approved):
LPTM 🚀

Issue #4769, opened 2022-04-06
Bug - [Form] - max-width for forms should be 880px


mcarrano:
**Describe the problem**
It was discovered that the max-with for the form component (if set) is currently 500px but the design guidance is 880px. Seems like this was a disconnect at some point.

Here is a Slack thread about this:

[Maria Agaphontzev](https://app.slack.com/team/UHBKEFHDY)  [4:19 AM](https://patternfly.slack.com/archives/C4FM977N0/p1649233157571419)
When using forms the design guidelines say that max width should usually be 880px, but when using isWidthLimited the max width is 500px. shouldnt these values be the same?
https://www.patternfly.org/v4/components/form/design-guidelines#forms-on-a-page
PatternFly docsPatternFly docs
[PatternFly 4 • Form](https://www.patternfly.org/v4/components/form/design-guidelines#forms-on-a-page)
PatternFly is Red Hat's open source design system. It consists of components, documentation, and code for building enterprise applications at scale.

[mcarrano](https://app.slack.com/team/U2NT3LSF6)  [9:04 AM](https://patternfly.slack.com/archives/C4FM977N0/p1649250273407479)
[@Maria Agaphontzev](https://patternfly.slack.com/team/UHBKEFHDY) 880px sounds right. I don't know where the 500px value comes from. [@mcoker](https://patternfly.slack.com/team/U8SGWKL15) do you have any insight here?

[mcoker](https://app.slack.com/team/U8SGWKL15)  [10:44 AM](https://patternfly.slack.com/archives/C4FM977N0/p1649256281027879)
[@mcarrano](https://patternfly.slack.com/team/U2NT3LSF6) I’m not sure, it was added as 500px in the original PR:
PR - https://github.com/patternfly/patternfly/issues/3282
Issue - https://github.com/patternfly/patternfly/pull/3517
Design issue - https://github.com/patternfly/patternfly-design/issues/862
[10:45](https://patternfly.slack.com/archives/C4FM977N0/p1649256315221859)
Here’s a codepen that will toggle between 500 and 880px, to see the difference - https://codepen.io/mcoker/pen/vYpdowP
CodePenCodePen
[vYpdowP](https://codepen.io/mcoker/pen/vYpdowP)
... (51 kB)
https://codepen.io/mcoker/pen/vYpdowP[](https://codepen.io/mcoker/pen/vYpdowP)

[...]

PR #4782, opened 2022-04-08
fix(form): changes width limited form max width from 500 to 800px

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4769


mcarrano (approved):
Looks fine @mcoker . Can't think if there's anything we should test for.

mcoker:
@mcarrano I think it's OK unless you think it's worth running this by any teams to see if it's going to have a negative impact.

mcarrano:
@mcoker I think it will be hard to get pre-approval for this, but we should definitely communicate this out via release notes and a possible mention during the next community meeting. If someone did not like the new wider default, I assume they could override the default with a different value?

Issue #4767, opened 2022-04-05
Data list - update expandable demo include expand/collapse all in toolbar


nicolethoen:
Follow up to: https://github.com/patternfly/patternfly-react/issues/6933

Update the full page expandable demo to include an expand/collapse button in the toolbar.

PR #4784, opened 2022-04-08
chore(data-list): updated expandable demo to include expand/collapse all in toolbar

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4767

### DIFF
--- a/src/patternfly/demos/DataList/examples/DataList.md
+++ b/src/patternfly/demos/DataList/examples/DataList.md
@@ -62,6 +62,7 @@ wrapperTag: div
       {{> toolbar-template
             toolbar--id="data-list-expandable-example-toolbar"
             toolbar-template--HasToggleGroup="true"
+            toolbar-template--HasExpandAll="true"
             toolbar-template--HasBulkSelect="true"
             toolbar-template--HasSearchFilter="true"
             toolbar-template--HasOverflowMenu="true"
--- a/src/patternfly/demos/Toolbar/toolbar-template.hbs
+++ b/src/patternfly/demos/Toolbar/toolbar-template.hbs
@@ -1,6 +1,7 @@
   {{!--
 Options: [
   toolbar-template--HasToggleGroup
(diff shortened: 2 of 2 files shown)


src/patternfly/demos/DataList/examples/DataList.md
```diff
@@ -62,6 +62,7 @@ wrapperTag: div
       {{> toolbar-template
             toolbar--id="data-list-expandable-example-toolbar"
             toolbar-template--HasToggleGroup="true"
+            toolbar-template--HasExpandAll="true"
```
mattnolting, line 65: 👍 👍

src/patternfly/demos/Toolbar/toolbar-template.hbs
```diff
@@ -25,6 +26,13 @@ Options: [
       {{!-- toggle group --}}
       {{#if toolbar-template--HasToggleGroup}}
         {{#> toolbar-group toolbar-group--modifier="pf-m-toggle-group pf-m-show-on-xl"}}
+          {{#if toolbar-template--HasExpandAll}}
```
mattnolting, line 29: I think you'll want to move this outside of the `toggle-group` for two reasons. 1) I _think_ the expand all button should remain visible when `toggle-group` hides its contents and 2) You would need to consider it in the expandable content section when `toggle-group` moves it there. 

@mmenestr @mcarrano does this sound right to you? Would you expect the expand all button to remain visible when filters are hidden?

mcoker: @mattnolting good call, updated!

mmenestr: What do you mean when filters are hidden? Like if there's no toolbar?

mattnolting (approved):
LPTM 👍

mcarrano (approved):
Looks great. Thanks @mcoker !

nicolethoen (approved):
LGTM!

PR #4785, opened 2022-04-11
feat(button): added selected state

### LINKED ISSUE #4664: Add selected state for masthead items
We have this support in the current page header that allows a button in the header to have a "selected state". We should add support for that in the masthead, too.

The existing style can be seen in this notification drawer demo
https://www.patternfly.org/v4/components/notification-drawer/html-demos/expanded-read/

cc @jeffpuzzo @mattnolting

mattnolting:
closes #4664

### DIFF
--- a/src/patternfly/components/Button/examples/Button.md
+++ b/src/patternfly/components/Button/examples/Button.md
@@ -124,6 +124,15 @@ import './Button.css'
 {{/button}}
 ```
 
+### Selected
+```hbs
+{{#> button button--modifier="pf-m-plain pf-m-selected" button--attribute='aria-label="Notifications"'}}
+  {{#> notification-badge}}
+    <i class="pf-icon-bell" aria-hidden="true"></i>
+  {{/notification-badge}}
+{{/button}}
+```
+
 ### Disabled
 ```hbs
 {{#> button button--modifier="pf-m-primary" button--attribute="disabled"}}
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Button/examples/Button.md
```diff
@@ -383,6 +392,7 @@ Semantic buttons and links are important for usability as well as accessibility.
 | `.pf-m-block` | `.pf-c-button` | Creates a block level button. |
 | `.pf-m-control` | `.pf-c-button` | Modifies for control styles. **Note:** This modifier should only be used when using buttons in the Input Group or Clipboard Copy components. |
 | `.pf-m-expanded` | `.pf-c-button.pf-m-control` | Modifies a control button for the expanded state. |
+| `.pf-m-selected` | `.pf-c-button` | Modifies a control button for the selected state. |
```
mcoker, line 395: @mcarrano @mmenestr @mceledonia is there a guideline for when to use "selected" instead of "expanded"?

mmenestr: Idk of a guideline - will defer to @mcarrano on that one, but I want to say that "selected" makes more sense in this case because it's a technically just a button that's been selected, regardless of what it leads to.

mcarrano:
@mcoker I don't know that there's any effective difference between "selected' and "expanded." I remember that a while ago we discussed whether buttons needed an "active" state. Is this the same thing or different.

I'm also not sure what I am looking at here. I see the example for Selected as a white button on a black background. This is for use in a dark background like tha masthead, correct? Does this also get a hover? Would we need an equivalent selected button state for buttons on a light background?

mattnolting:
@mcarrano Since we're supporting dark theme universally now, we'll need a spec for dark/light mode as a button anywhere could serve as this drawer toggle, and any interaction states you want included. We need to avoid contextual styling (masthead specific) in favor of variant/modifier application, as the former approach creates unpredictable dependencies. 

> Would we need an equivalent selected button state for buttons on a light background?

Yes, I believe this will be necessary for all design enhancements in the future.

mcarrano:
@mceledonia can you look at this also? Do we have the button states defined as @mattnolting outlines or do we need to create a design issue to examine this further?

src/patternfly/components/Button/examples/Button.md
```diff
@@ -124,6 +124,15 @@ import './Button.css'
 {{/button}}
 ```
 
+### Selected
+```hbs
```
mcoker, line 128: Let's mark this as beta so we can change it.

```suggestion
```hbs isBeta
```

src/patternfly/components/Button/examples/Button.md
mcoker, line 395: I'm assuming this can work with any kind of button?

```suggestion
| `.pf-m-selected` | `.pf-c-button` | Modifies a button for the selected state. |
```

mcoker (changes requested):
This works for the masthead, but if a selected button is a general variant, it should start in light theme. For now, we could add the dark theme styles for it in the `%pf-t-dark` extend to work with our existing "dark" style grouping, which should make it work in masthead. Or we could just build it into masthead for now until we're able to fully support a dark theme and sort the styles out better.

mcoker:
closing in favor of https://github.com/patternfly/patternfly/pull/5048

Issue #4757, opened 2022-03-22
Add support Tab close and new Tab buttons


christianvogt:
**Describe the issue. What is the expected and unexpected behavior?**
OpenShift Console is creating a terminal view with the ability to add and remove tabs. It uses the `Tab` component along with `TabTitleIcon` in the title.

There is no clean way to support the addition of a button in the title. We can easily add our own `Button`, minus the padding. However I feel like the addition of a close button should be part of the react component framework.

Additionally, support for a button to the right of the tabs that is always visible in order to `add new tab` should be provided. Today we can get around this by adding an additional tab and make it function as a button and not a content tab. However if many tabs are present, this tab can scroll off screen. Ideally such a button would become sticky on the right side and not get hidden.

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**
enhancement

**What is your product and what release version are you targeting?**
OpenShift 4.11

mcarrano:
@christianvogt @abhinandan13jan what is your expectation for naming the new tab when it is added? Should the new tab name just be set by the consumer? I would expect so, but want to make sure there's no expectation that the user can edit the tab names.

christianvogt:
@mcarrano The new tab action should signal the parent component to render a new tab. Therefore tab name will be controlled by the parent component.

Editing of tab names via an inline editor may be requested in future but is unnecessary atm. I believe the rendering component has control over the display of the title and can render a customized editable tab title if they choose to right now.

mcoker:
@christianvogt I have a PR here if you'd like to leave a review - https://github.com/patternfly/patternfly/pull/4787

PR #4787, opened 2022-04-11
feat(tabs): added close button and add new tab

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4757

There is currently no styling or anything associated with the "add new tab" tab - it's just a regular tab with an icon. @mcarrano @mmenestr wdyt?

### DIFF
--- a/src/patternfly/components/Tabs/__tabs-item.hbs
+++ b/src/patternfly/components/Tabs/__tabs-item.hbs
@@ -1,19 +1,9 @@
 {{#> tabs-item tabs-item--IsCurrent=__tabs-item--current}}
   {{#> tabs-link tabs-link--attribute=(concat __tabs-item--attribute ' id="' tabs--id '-' __tabs-item--id '-link"')}}
     {{#if __tabs-list--HasIcons}}
-      {{#if __tabs-item--NoText}}
-        {{> tabs-item-icon
-          tabs-item-icon--name=__tabs-item--icon-name
-          tabs-item-icon--attribute='aria-label="{{__tabs-item--aria-label}}"'
-          }}
-      {{else}}
-        {{> tabs-item-icon
-          tabs-item-icon--name=__tabs-item--icon-name
-          tabs-item-icon--attribute='aria-hidden="true"'
-          }}
-      {{/if}}
+      {{> tabs-item-icon tabs-item-icon--name=__tabs-item--icon-name}}
     {{/if}}
-    {{#unless __tabs-list--NoText}}
+    {{#unless __tabs-item--NoText}}
       {{#> tabs-item-text}}
         {{{__tabs-item--text}}}
       {{/tabs-item-text}}
--- a/src/patternfly/components/Tabs/__tabs-item.hbs
+++ b/src/patternfly/components/Tabs/__tabs-item.hbs
@@ -1,5 +1,5 @@
-{{#> tabs-item tabs-item--IsCurrent=__tabs-item--current}}
-  {{#> tabs-link tabs-link--attribute=(concat __tabs-item--attribute ' id="' tabs--id '-' __tabs-item--id '-link"')}}
+{{#> tabs-item tabs-item--IsCurrent=__tabs-item--current tabs-item--IsAdd=__tabs-item--IsAdd tabs-item--modifier=__tabs-item--modifier tabs-item--attribute=__tabs-item--attribute}}
+  {{#> tabs-link tabs-link--attribute=(concat __tabs-link--attribute ' id="' tabs--id '-' __tabs-item--id '-link"')}}
     {{#if __tabs-list--HasIcons}}
       {{> tabs-item-icon tabs-item-icon--name=__tabs-item--icon-name}}
     {{/if}}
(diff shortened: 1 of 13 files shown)


mcarrano (commented):
@mcoker thanks for sharing this. I feel like the Close button is a little heavy visually. I know that's our standard icon button styling, but maybe we need something different here? I have a similar feeling about the Add (+) button. @mceledonia can you take a look at this?

mceledonia:
@mcarrano @mcoker Yeah, I think we can drop down the icon size to small for both close and add, then keep the same padding shown above. Here's a quick mockup of how that looks:

image

christianvogt:
IMO the padding to the left and right of the `x` button seems off as the gap on the right is smaller. But maybe that's ok?
image


I was testing the addition of the close button to several other variants and there seems to be issues.
image
image

mceledonia:
I'm ok with them being different (left vs right padding) but the padding on the right does feel rather tight on the box variation due to the border being more visible. Any reason not to have separate right-hand padding between box and filled variations? On box I would propose bumping it up a spacer to better balance against the more visible shape of the tab.

mcoker:
@mceledonia here is the change with extra padding on the right in the box variation

Screen Shot 2022-04-12 at 1 51 47 PM

FWIW, here's that change but applying it to the close button globally. I think this looks OK too.

Screen Shot 2022-04-12 at 1 53 36 PM

And this is the same change, but using equal left/right padding on the close button, which separates the text and close button a bit.

Screen Shot 2022-04-12 at 1 48 47 PM

wdyt?

mceledonia:
@mcoker I actually think it looks fine on both, thanks for trying it out. These look great to me!

mmenestr:
The only comment I have (and maybe it's just me) but visually, the text and x icon don't look like they're aligned - ie: the x icon sits a little higher than the text and it feels off to me? Idk if that's something that can be changed though. Looks like they are vertically center aligned when I use inspect which in theory is correct, but wonder what it would look like if the X was a few pixels lower, to bottom align with the text?

mcarrano:
@mcoker I like this much better now. I think the question that @mmenestr raised is an interesting one. I hadn't noticed it until I read her comment. Is that a function of the X icon being enclosed in a button?

Also, I wondered if we could change the last example to include both the Close button and the Add button together to see what this will really look like. I can't see a case where we would support adding new tabs without the ability to remove them.

mcarrano:
@mcoker thanks for adding the additional examples. It does highlight one problem with how this is implemented, however. Looks like the Add button is just another tab that means it gets hidden when the tabs overflow and horizontal scrolling is exposed. I'd expect the Add button to always be visible. The way Firefox handles this is a good model I think.

Screen Shot 2022-04-13 at 9 42 04 AM

Note that the Add button exists outside of the tab group following the scroll button so it is always available. Can we do something like that?

mcoker:
@mcarrano yep we can, though I'll note that I asked about that specifically in one of the planning meetings and we decided it should just be another tab at the end of the list, contained within the scroll buttons - you would need to scroll to the end of the list to see the add tab button. Just want to mention that in case there was a specific reason for that - or maybe I miscommunicated/misunderstood our previous convo 😁

Also assuming we move it out of the list and scroll buttons (when present), should it still look like a regular tab (including the basic tab hover/focus styling) or should it have a more generic style?

mmenestr:
That looks better @mcoker thank you!
And I kind of agree with Matt that it would maybe be better to move it outside - I don't remember the tab conversation so I can't say why we didn't say that before but, one thing I like about moving it outside, apart from it not getting lost, is that there's also a clearer visual distinction between the end of a tab and the start of the + tab. Right now there's no line or anything between them which feels weird to me.

christianvogt:
I highlighted the use of the tab as the add button in the original defect https://github.com/patternfly/patternfly/issues/4757

If we can solve the issue of it being hidden while still being present adjacent to the last opened tab (instead of right justified), that would be my preference.

mcarrano:
> @mcarrano yep we can, though I'll note that I asked about that specifically in one of the planning meetings and we decided it should just be another tab at the end of the list, contained within the scroll buttons

Yeah, sorry about that @mcoker . I may have said that without realizing what the implications of that would be.

> Also assuming we move it out of the list and scroll buttons (when present), should it still look like a regular tab (including the basic tab hover/focus styling) or should it have a more generic style?

I think the later, because it would be wrong, IMO, for this to be styled as a tab, but not behave like the other tabs. Do you agree @mceledonia @mmenestr ?

> If we can solve the issue of it being hidden while still being present adjacent to the last opened tab (instead of right justified), that would be my preference.

I agree

mceledonia (approved):
Latest looks good!

mcarrano:
@mcoker the screen grab above looks great. For some reason, the preview still feels like it has more padding for the scroll button, or is that my imagination?

Screen Shot 2022-04-13 at 4 29 15 PM

mcarrano (approved):
Everything looks good. Thanks @mcoker !

mattnolting (approved):
Nice work! LPTM

PR #4809, opened 2022-04-26
fix(description-list): fix text wrap

### LINKED ISSUE #4768: Long strings in description list term overflow
re: https://github.com/stackrox/stackrox/pull/1197

Whenever the term or description element is flex/grid, the text element in it needs `min-width: 0;`. Probably safe just to set that by default on the text element?

https://codesandbox.io/s/gifted-dewdney-nkjes5?file=/index.html

srambach:
Fixes #4768 

Fixes text overflowing in data list text (when there are no spaces in the text).


mattnolting (approved):
LPTM 👍

PR #4810, opened 2022-04-27
chore(primary-detail): updated card view demo to use new selectable styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4792


evwilkin (approved):
LGTM

PR #4823, opened 2022-05-02
fix(DataList): selectable item states

### LINKED ISSUE #3847: Selectable data list bugs
There should be no shadow between adjacent selected rows.

Screen Shot 2021-02-09 at 6 02 55 PM

There is currently a thin grey border showing between adjacent selected rows when you hover over one.

Screen Shot 2021-02-09 at 6 02 40 PM

eliranmal:
# content

###### `DataList` component

- fix selected items' undesired dividing border on hover
- remove selected items' box shadow, to avoid showing in between items
- place the `&.pf-m-selected` selector as a child of `&.pf-m-selectable` (may break usage, not API)

###### `DataList` example / _**Selectable rows**_

- update item text to reflect the selected state

###### `DataList` example / _**Selectable expandable rows**_

- remove redundant item prop from the list element (`isSelectable`)
- add `isSelected` where needed to reflect the example's intention. this fixes _expanded & selected_ items' color
- add `isSelectable` where missing (that is - anywhere `isSelected` already exists), to enable items' intended selected state

# screenshots

**note:** interaction states are demonstrated on the first list item.

### [data-list / selectable rows][1]

| state | before | after |
| --- | --- | --- |
| - | selectable-rows-before | selectable-rows-after |
| hover | selectable-rows-item-hover-before | selectable-rows-item-hover-after |
| active | selectable-rows-item-active-before | selectable-rows-item-active-after |

### [data-list / selectable rows][1] (stacked layout)

| state | before | after |
| --- | --- | --- |
| - | selectable-rows-stacked-before | selectable-rows-stacked-after |
| hover | selectable-rows-stacked-item-hover-before | selectable-rows-stacked-item-hover-after |
| active | selectable-rows-stacked-item-active-before | selectable-rows-stacked-item-active-after |

### [data-list / selectable expandable rows][2]

| before | after |
| --- | --- |
| selectable-expandable-rows-before | selectable-expandable-rows-after |

# ticketing

fixes #3847

[1]: https://www.patternfly.org/v4/components/data-list/html/#selectable-rows
[2]: https://www.patternfly.org/v4/components/data-list/html/#selectable-expandable-rows

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -27,8 +27,6 @@
   --pf-c-data-list__item--m-expanded--m-selectable--before--BackgroundColor: var(--pf-global--active-color--300);
   --pf-c-data-list__item--BorderBottomColor: var(--pf-global--BorderColor--300);
   --pf-c-data-list__item--BorderBottomWidth: #{pf-size-prem(8px)};
-  --pf-c-data-list__item--m-selectable--hover--item--BorderTopColor: var(--pf-c-data-list__item--BorderBottomColor);
-  --pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth: var(--pf-c-data-list__item--BorderBottomWidth);
   --pf-c-data-list__item--sm--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-data-list__item--sm--BorderBottomColor: var(--pf-global--BorderColor--100);
 
@@ -238,13 +236,21 @@
 
   &::before {
     position: absolute;
-    top: var(--pf-c-data-list__item--before--Top);
+    top: 0;
     bottom: 0;
     left: 0;
     width: var(--pf-c-data-list__item--before--Width);
     content: "";
     background-color: var(--pf-c-data-list__item--before--BackgroundColor);
     transition: var(--pf-c-data-list__item--before--Transition);
+  }
+
+  &:first-of-type::before {
+    top: var(--pf-c-data-list__item--before--Top);
+  }
+
+  &:last-of-type::before {
+    bottom: var(--pf-c-data-list__item--before--Top);
   }
 
   &.pf-m-selectable {
@@ -256,28 +262,12 @@
 
       position: relative;
       z-index: var(--pf-c-data-list__item--m-selected--ZIndex);
-
-      &:hover,
-      &:focus {
-        margin-bottom: var(--pf-c-data-list__item--BorderBottomWidth);
-        border-bottom-width: 0;
-      }
     }
 
     &:hover,
     &:focus {
       position: relative;
       z-index: var(--pf-c-data-list__item--m-selectable--hover--ZIndex);
-
-      &:not(.pf-m-selected):not(:last-child) {
-        --pf-c-data-list__item--BorderBottomWidth: 0;
-
-        // stylelint-disable
-        + .pf-c-data-list__item {
-          border-top: var(--pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth) solid var(--pf-c-data-list__item--m-selectable--hover--item--BorderTopColor);
-        }
-        // stylelint-enable
-      }
     }
 
     &:hover {
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -257,7 +257,8 @@
       position: relative;
       z-index: var(--pf-c-data-list__item--m-selected--ZIndex);
 
-      &:hover {
+      &:hover,
+      &:focus {
         margin-bottom: var(--pf-c-data-list__item--BorderBottomWidth);
         border-bottom-width: 0;
       }
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -251,6 +251,18 @@
     cursor: pointer;
     outline-offset: var(--pf-c-data-list__item--m-selectable--OutlineOffset);
 
+    &.pf-m-selected {
+      --pf-c-data-list__item--before--BackgroundColor: var(--pf-c-data-list__item--m-selected--before--BackgroundColor);
+
+      position: relative;
+      z-index: var(--pf-c-data-list__item--m-selected--ZIndex);
+
+      &:hover {
+        margin-bottom: var(--pf-c-data-list__item--BorderBottomWidth);
+        border-bottom-width: 0;
+      }
+    }
+
     &:hover,
     &:focus {
       position: relative;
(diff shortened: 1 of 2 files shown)


src/patternfly/components/DataList/data-list.scss
```diff
@@ -251,6 +251,18 @@
     cursor: pointer;
     outline-offset: var(--pf-c-data-list__item--m-selectable--OutlineOffset);
 
+    &.pf-m-selected {
+      --pf-c-data-list__item--before--BackgroundColor: var(--pf-c-data-list__item--m-selected--before--BackgroundColor);
+
+      position: relative;
+      z-index: var(--pf-c-data-list__item--m-selected--ZIndex);
+
+      &:hover {
```
srambach, line 260: I think this will need to also have :focus because when I use keyboard focus, I can still see the break in the border. 
image

srambach (commented):
Hi, @eliranmal - thank you for the detailed explanation of changes. I really appreciate having that.

I left one comment on the code about the :focus state. Other than that, I'd just like to hear if @mcoker is ok with the increase in specificity and if so, the rest looks fine to me.

src/patternfly/components/DataList/data-list.scss
```diff
@@ -280,14 +292,6 @@
     }
   }
 
-  &.pf-m-selected {
-    --pf-c-data-list__item--before--BackgroundColor: var(--pf-c-data-list__item--m-selected--before--BackgroundColor);
-
-    position: relative;
-    z-index: var(--pf-c-data-list__item--m-selected--ZIndex);
-    box-shadow: var(--pf-c-data-list__item--m-selected--BoxShadow);
```
mcoker, line 288: We still need to retain the shadow, as that's a pattern we have in other components, too. I think this change might do it?

Add this var

```
--pf-c-data-list__item--item--m-selected--BoxShadow: var(--pf-global--BoxShadow--sm-bottom);
```

And then use it here

```
diff --git a/src/patternfly/components/DataList/data-list.scss b/src/patternfly/components/DataList/data-list.scss
index b531d184..0acec449 100644
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -286,6 +286,10 @@
     position: relative;
     z-index: var(--pf-c-data-list__item--m-selected--ZIndex);
     box-shadow: var(--pf-c-data-list__item--m-selected--BoxShadow);
+
+    & + & {
+      --pf-c-data-list__item--m-selected--BoxShadow: var(--pf-c-data-list__item--item--m-selected--BoxShadow);
+    }
   }
 
   &.pf-m-ghost-row {
```

src/patternfly/components/DataList/data-list.scss
mcoker: Hmm, this removes the bottom border on hover, which I think looks a bit off since the border is removed when you hover the top selected row, but it appears if you hover the bottom selected row.

Hovering the top selected row:
Screen Shot 2022-07-05 at 6 32 20 PM

Hovering the bottom selected row:
Screen Shot 2022-07-05 at 6 32 25 PM

For this little gap, I'm thinking we might need some input from design. We're probably going to need to make a decision about how far the blue border spans for both selected and expanded rows, and for when you have adjacent rows that are expanded (not selected) and selected.

mcoker (changes requested):

mcoker:
@eliranmal after talking with @mmenestr and @mceledonia about the gap between selected rows on hover/focus we decided:

* To keep that gap all of the time (not just on hover/focus) as a way to provide a visual separation between rows. That means this would end up looking like this (note that the first and last rows cover the data-list's top and bottom border)
  * Screen Shot 2022-07-06 at 11 59 05 AM
* We also want to apply this to expandable rows so that there is a gap between expandable rows, but no gap between the expandable row content and the expandable row that triggered the expanded content to show. Like this 
  * Screen Shot 2022-07-06 at 12 11 54 PM

This would apply to both data-lists and tables (these screenshots are of a table). Hopefully we could make the update to both components in the same sprint for consistency's sake. Lemme know what you think!

mcoker:
@eliranmal just wanted to check in and make sure you saw my latest feedback - let me know if there's anything I can do to help!

eliranmal:
hey @mcoker , those notifications fell through as i focused on the ROSA development for OCM... sorry about not updating.

i don't think i need anything else, i'll get to it. thanx 👍

src/patternfly/components/DataList/data-list.scss
eliranmal: > this will need to also have :focus

good catch. adding `:focus` to that selector works pretty well

eliranmal: about the gap with the selected row bottom-border; my intention was to have the top border server as a consistent cue for the user for the _selected_ state.

i can re-add it to the last item, and try to get a feel about how it impact the experience.

for sure we're gonna need some help from designers :)

eliranmal:
ok, i saw the design updates from your discussion.

that renders my previous comments obsolete, but it's perfect for impl'! thanks

i'll follow those guidelines (on keeping a gap on these rows), and see if i need to update any more components to support the table (i presume not, as a _table_ must be a derivation of a _data list_..)

mcoker:
@eliranmal sounds great, thanks!

eliranmal:
ok, i got some progress (see [this change][1]) - now list items always have fixed spacing between them.

2 things left to do:

- [ ] test and impl' a distinction for grouped (expanded) rows, to glue them back together
- [ ] test the box-shadow on hover, for regressions


[1]: 910313d319cac887bcddd96c6684aa2dbc840b97

eliranmal:
> then, all i've left to do is to test the box-shadow on hover, for regressions

done - found one regression (which is arguable a bug), when hovering the top and bottom items:

formerly, the top item's shadow would not bleed out from the list area, so the top outer-shadow would be lacking/clipped.  
currently, both the top and the bottom items show their border on both their tops and bottoms.

@mcoker - i think that's ready for re-review. please let me know if there's any requirement i missed, or caused some other regression.

cheers :)

src/patternfly/components/DataList/data-list.scss
```diff
@@ -27,8 +27,6 @@
   --pf-c-data-list__item--m-expanded--m-selectable--before--BackgroundColor: var(--pf-global--active-color--300);
   --pf-c-data-list__item--BorderBottomColor: var(--pf-global--BorderColor--300);
   --pf-c-data-list__item--BorderBottomWidth: #{pf-size-prem(8px)};
-  --pf-c-data-list__item--m-selectable--hover--item--BorderTopColor: var(--pf-c-data-list__item--BorderBottomColor);
-  --pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth: var(--pf-c-data-list__item--BorderBottomWidth);
```
mattnolting, line 31: These would technically be breaking changes. Looks like you could change to:
```
  --pf-c-data-list__item--m-selectable--hover--item--BorderTopColor: transparent;	
  --pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth: 0;	
```

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -800,8 +800,8 @@ When a list item includes more than one block of content, it can be difficult fo
 
 ### Selectable expandable rows
 ```hbs
-{{#> data-list data-list--id="data-list-selectable-expandable-rows" data-list-item--IsSelectable="true" data-list--attribute='aria-label="Selectable, expandable data list example"'}}
-  {{#> data-list-item data-list-item--id="item-1" data-list-item--expanded="true"}}
+{{#> data-list data-list--id="data-list-selectable-expandable-rows" data-list--attribute='aria-label="Selectable, expandable data list example"'}}
```
mattnolting, line 803: 👍

mattnolting (changes requested):
Thanks for this contribution, @eliranmal! Left a couple of suggestions.

PR #4827, opened 2022-05-03
feat(search-input): added collapsible variant

### LINKED ISSUE #4705: Search input - add dark and masthead variants
What is the existing component, if any? Search Input

**Describe the feature**
Want to add new variants of the Search input to be used either in the masthead or above the primary nav.

The first variant is just a dark version of the existing search input that can be used above the navigation or anywhere else a search input is required on a dark background.

The second variant is intended for use in the masthead and includes both dark styling and the ability to collapse the input to an icon. This behavior can be found on the PatternFly website.

See https://marvelapp.com/prototype/hgh0dj1/screen/85392551 for final design.

**Any other information?**

mattnolting:
closes #4705 

This PR does a couple of things.
 - Illustrates an approach to setting dark theme styles on a nested component
 - Adds `plain` variant to `input group`
 - Adds `plain` variant to `text-input-group`
 - Adds `bordered` variant to `text-input-group`
 - Adds `hidden` variants to `input-group` and `text-input-group`
 - Updates the name of the dark theme mixins as to avoid conflict with other possible theme names

@mcarrano These examples use dark theme styles and standard componentry, rather than a context specific variant of the `input-group/text-input-group`. There are several reasons for this. 
- Drive interactive/visual consistency through componentry. 
- Avoid duplicate code and css bloat.
- Avoid unnecessarily complicating context specific css for specific components
- Preserves `pf-m-light` variant for masthead (although this won't work correctly in globally set dark mode, which is another issue to address)

It also applies dark theme standards to components used in the masthead. This is really a byproduct of using dark theme styles in masthead.

**Before:**

Screen Shot 2022-05-03 at 11 02 39 AM

**After:**

Screen Shot 2022-05-03 at 11 02 55 AM


mcarrano (commented):
@mattnolting I think this looks good. I'd like @mceledonia and @doruskova to also take a look and see if they have any problems with your implementation. I was curious why the new text input group - collapsed examples come first before the Basic example. Seems inconsistent with our usual approach.

mattnolting:
> I was curious why the new text input group - collapsed examples come first before the Basic example. Seems inconsistent with our usual approach.

@mcarrano So they're easy for you to find/see. I'll move them before final PR preparation.

doruskova:
@mcarrano  Looks great! I just wanted to ask if we need a light variant for it. I know that you @mattnolting mentioned that you need a design for a light variant. However, we don't have any demo or something that shows a light masthead.

mattnolting:
@doruskova no need for a light theme with this approach! Thanks

doruskova:
@mattnolting Thank you! Nice work :)

mcarrano (approved):
Thanks for checking this @doruskova . Looks good to go then!

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -247,6 +255,13 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   .pf-c-nav {
     align-self: stretch;
   }
+
+  &:where(:not(.pf-m-light):not(.pf-m-light-200)) {
```
mcoker, line 259: would this work?

```suggestion
  &:where(:not([class*="pf-m-light"])) {
```

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -49,6 +49,13 @@
   color: var(--pf-c-text-input-group--Color, inherit);
   background-color: var(--pf-c-text-input-group--BackgroundColor);
 
+  &.pf-m-hidden,
+  .pf-m-hidden,
+  &[hidden],
+  [hidden] {
```
mcoker, line 55: Same here and for `.pf-m-hidden` above - can/should these be limited to direct descendents so it doesn't conflict or do something it shouldn't with some random nested child?

mcoker (changes requested):
Sorry it's taken so long to look at this! This looks really cool.

I mostly just reviewed the code and glanced at the demo, and will look again at the demo tomorrow, but wanted to give this feedback.

Also, just looking at the demo - typically when we hide/show something, especially using `hidden`, the react component will render that element on the fly. I'm wondering if we need all that functionality built in to the CSS, or if in the collapsed state we simply show the button, and in the expanded state, we just show the text input group. The hidden/visible stuff would be a good amount of work to implement on the react side, too.

mcoker (changes requested):

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -362,3 +361,18 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
 @include pf-theme-dark {
   @include pf-theme-dark-masthead;
 }
+
+// set masthead to dark theme by default
+// ignore dark theme if light variants are called
+@import "../../base/themes/dark/variables";
```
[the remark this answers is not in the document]
mcarrano: That's an interesting problem. Are there risks associated with that? I feel like this feature is somewhat a 'nice to have' so I'm wary about introducing something that causes downstream residual problems.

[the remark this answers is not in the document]
mcoker: > Are there risks associated with that?

@mcarrano currently our dark theme component vars aren't setup to be themed as easily as the light theme vars. And as we adjust things in dark theme (look/feel/colors, CSS structure, etc), we may introduce problems with whatever dark theme styles are being pulled into masthead. I suppose a user could put something in the masthead that doesn't look the way we intend it to, too, if there isn't full dark theme support for that component. In this PR it looks like we're importing all of the dark theme global var overrides into the masthead component to set the base styles for the masthead to come from the dark theme global styles.

mcoker:
Per our discussion, let's see if we can avoid relying on the new dark theme styles which aren't fully released, tested, etc, and see if we can use the existing [`pf-t-[light/dark]` @mixins](https://github.com/patternfly/patternfly/blob/0b89bb099b24f6cbc46a6bbd9deb7b596a3f4371/src/patternfly/sass-utilities/mixins.scss#L190-L200) to build this.

mcarrano (approved):
Looks good. Thanks @mattnolting !

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -243,6 +243,10 @@ $pf-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   .pf-c-nav {
     align-self: stretch;
   }
+
+  .pf-c-button.pf-m-plain {
+    @include pf-t-dark; // for all plain buttons, especially inside other components (.pf-c-input-group), for dark mode
```
mcoker, line 248: The mixin also sets `color` by default since we typically use this on blocks that inherit `color`, and `color` may not be redefined on things within the block - meaning the global var change wouldn't apply. So if the thing we apply it to uses a color other than the default, we need to pass that. https://github.com/patternfly/patternfly/blob/9ab78a81b9d6d63cd7795fa82b1ff673abba5332/src/patternfly/sass-utilities/mixins.scss#L190-L200

So here the button's default color is `Color--100` and it should be `Color--200`

```suggestion
    @include pf-t-dark(--pf-global--Color--200); // for all plain buttons, especially inside other components (.pf-c-input-group), for dark mode
```

However since the plain button defines `color`, maybe a better way to do this would be to update the extend so you can omit the `color` property from being set. That's a better way of addressing it than re-assigning the global var here in case the plain button color ever changes.

mcoker (changes requested):
Looks awesome! Just a couple of nits - the plain button's color is off a little bit and we should move the example under the search input examples.

src/patternfly/components/InputGroup/examples/InputGroup.md
```diff
@@ -115,4 +115,5 @@ When using the `.pf-c-input-group` always ensure labels are used outside the inp
 | -- | -- | -- |
 | `.pf-c-input-group` | `<div>` |  Initiates the input group. **Required** |
 | `.pf-c-input-group__text` | `<span>` |  Initiates the input group text. This can be used to show text, radio, icons, or check boxes. |
+| `.pf-m-plain` | `.pf-c-input-group` | Removes the background from the input group. |
```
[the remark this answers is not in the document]
mattnolting: I agree, seems to me that input group should definitely have a transparent background.

src/patternfly/components/InputGroup/input-group.scss
```diff
@@ -27,6 +27,10 @@
   width: 100%;
   background-color: var(--pf-c-input-group--BackgroundColor);
 
+  &.pf-m-plain {
+    --pf-c-input-group--BackgroundColor: transparent; // update at breaking change
```
mcoker, line 31: wdyt about this? just so we aren't left wondering what we agreed to update it to?  😅

```suggestion
    --pf-c-input-group--BackgroundColor: transparent; // make default at breaking change
```

mcoker (approved):
LPTM! One thing about the breaking change comment, but up to you.

PR #4828, opened 2022-05-03
feat(Label): add button for adding new labels

### LINKED ISSUE #4756: [Label Group] - editable label group
Want to introduce the ability for users to add new labels to a label group component.

Design issue: https://github.com/patternfly/patternfly-design/issues/1108
Final designs: [Marvel ](https://marvelapp.com/prototype/2d90c04g/screen/85343000)and [Design doc](https://docs.google.com/document/d/1XfP0mgrIrTSUTTClCMl3hvU_CVE14jJmVv-LEOXdfuY/edit#)

thatblindgeye:
Closes #4756 

Convenience links:

[Label componeont](https://patternfly-pr-4828.surge.sh/components/label#add-label)
[LabelGroup component](https://patternfly-pr-4828.surge.sh/components/label-group#add-label)

This PR:

- Adds a parameter for an add label
- Adjusts the CSS so that the add label shares the same styles with the overflow button (per convo with Matt and Margot)
- Adds simple examples of what the add label would look like in both the Label and LableGroup components

One thing to consider with how I updated the CSS is that within the Label page, the variables aren't exposed below the doc table for the `pf-m-add` selector, and within the CSS itself this selector is also using variables with "overflow" in their name. I believe an alternative would be to copy+paste any styles for the overflow and change them to reflect this add label (` --pf-c-label--m-add__content--Color` and so on). Since the overflow and add labels are both assumed to always share the same styles, though, I'm wondering whether if at some point down the line these variables would be updated to be less specific to the overflow (though this still doesn't resolve this possible issue in the present)?


mcarrano (commented):
This looks good @thatblindgeye . Do you think we need an example that shows the Add label and overflow buttons existing together? Not sure, but wanted to raise the question.

thatblindgeye:
@mcarrano I'm not sure if such an example is *needed*, but I also wouldn't be opposed to adding one in just to show how both might look co-existing with one another. @mcoker what are your thoughts?

src/patternfly/components/Label/examples/Label.md
```diff
@@ -596,24 +603,36 @@ This style of label is used to indicate overflow within a label group.
 {{#> label label--id="compact-editable-label-active" label--modifier="pf-m-compact pf-m-blue" label--IsEditable="true" label--IsEditableActive="true"}}Compact editable active{{/label}}
 ```
 
+### Add label
+
+This style of label is used to add new labels to a label group.
+
+```hbs isBeta
+{{#> label label--id="add" label--IsAdd="true"}}
```
mcoker, line 611: Just a nit but I don't think this ID is necessary. Doesn't look necessary on a bunch of other labels, either, though we don't have to update those here.

```suggestion
{{#> label label--IsAdd="true"}}
```

src/patternfly/components/Label/examples/Label.md
```diff
@@ -596,24 +603,36 @@ This style of label is used to indicate overflow within a label group.
...
-| `.pf-m-blue` | `.pf-c-label` | Modifies the label to have blue colored styling. |
-| `.pf-m-green` | `.pf-c-label` | Modifies the label to have green colored styling. |
-| `.pf-m-orange` | `.pf-c-label` | Modifies the label to have orange colored styling. |
-| `.pf-m-red` | `.pf-c-label` | Modifies the label to have red colored styling. |
-| `.pf-m-purple` | `.pf-c-label` | Modifies the label to have purple colored styling. |
-| `.pf-m-cyan` | `.pf-c-label` | Modifies the label to have cyan colored styling. |
-| `.pf-m-editable` | `.pf-c-label` | Modifies label for editable styles. |
-| `.pf-m-editable-active` | `.pf-c-label.pf-m-editable` | Modifies editable label for active styles. |
+
+| Class                        | Applied to                  | Outcome                                                                                                                                                                                                                  |
```
mcoker, line 620: Nit - this kind of spacing is hard to maintain manually when making updates to the docs - or everyone making changes will need the same formatter - so we just use single spaces in the docs tables to keep it simple for all contributors.

src/patternfly/components/Label/label.hbs
```diff
@@ -1,4 +1,4 @@
-<{{# if label--type}}{{label--type}}{{else if label--IsOverflow}}button{{else}}span{{/if}} class="pf-c-label{{#if label--IsOverflow}} pf-m-overflow{{/if}}{{#if label--modifier}} {{label--modifier}}{{/if}}{{#if label--IsEditable}} pf-m-editable{{/if}}{{#if label--IsEditableActive}} pf-m-editable-active{{/if}}"
+<{{# if label--type}}{{label--type}}{{else if label--IsOverflow}}button{{else if label--IsAdd}}button{{else}}span{{/if}} class="pf-c-label{{#if label--IsOverflow}} pf-m-overflow{{/if}}{{#if label--IsAdd}} pf-m-add{{/if}}{{#if label--modifier}} {{label--modifier}}{{/if}}{{#if label--IsEditable}} pf-m-editable{{/if}}{{#if label--IsEditableActive}} pf-m-editable-active{{/if}}"
   {{#if label--attribute}}
```
mcoker, line 2: I just noticed this label is missing `type="button"` when it's a `<button>` element (default is `type="submit"` which we don't want). Could you add a conditional in here that tests for `IsOverflow/Add` and adds `type="button"`?

mcoker (changes requested):
Really great job @thatblindgeye!!! Just a few nits, and we discussed that the add button should (probably?) not be part of the label group list as it isn't an item in the group list - it's more like a control for the group. This will be inline with the add button we recently added to tabs, too.

src/patternfly/components/Label/examples/Label.md
thatblindgeye: Ah I think maybe Prettier got to this before I disabled it for the workspace. I noticed the SCSS file also had some formatting changed but turned Prettier off and reverted that, this one snuck by me.

thatblindgeye:
> Really great job @thatblindgeye!!! Just a few nits, and we discussed that the add button should (probably?) not be part of the label group list as it isn't an item in the group list - it's more like a control for the group. This will be inline with the add button we recently added to tabs, too.

@mcoker As mentioned in our convo, I'm leaning towards this being something for next milestone possibly? I know it was mentioned when we sat to discuss this that getting the add button to be more inline but also outside the label group list might be tricky, so it may just require a little more research/discussion.

mcoker (approved):
LGTM!! Great work!! 🤩

mcarrano (approved):
Looks good. Thanks @thatblindgeye !

PR #4830, opened 2022-05-04
chore(spinner): reverse the spin of legacy spinner

### LINKED ISSUE #3803: Update spinner to spin in opposite direction
This was added to the new SVG variation (https://github.com/patternfly/patternfly/pull/3690) but should be applied to the non-SVG variation, too.

srambach:
Fixes #3803 to make the legacy CSS spinner go clockwise, as the newer SVG spinner does.


mcoker (approved):
Awesome!!

PR #4836, opened 2022-05-06
fix(chip group): fixes an overflow problem

### LINKED ISSUE #4685: Clear button mispositioned in Select with chips
When using the `Select` component with chips it is possible that the 'Clear all' button is misaligned and overlaps the chips.

1. Go to the 'Multiple with Render Custom Chip Group' [example](https://www.patternfly.org/v4/components/select#multiple-with-render-custom-chip-group) in the Patternfly documentation
2. Select 'Florida', 'New Jersey' and 'North Carolina'
3. Set the width on the element with class `ws-example` to `200px`

<p align="center">
  A screenshot showing the misaligned button
</p>

It's a bug, non-blocking but not very user-friendly.

This issue was found in the [Keycloak admin console](https://github.com/keycloak/keycloak-admin-ui/pull/930#issuecomment-900801166).

```bash
npm list --depth=0 | grep patternfly
├── @patternfly/patternfly@4.125.3
├── @patternfly/react-core@4.147.0
├── @patternfly/react-icons@4.11.8
├── @patternfly/react-table@4.29.37
```

srambach:
Fixes #4685 
The overlap in the select shown in the related issue is caused by the chips not ellipsing when the width of the container is too narrow to fit them. This is caused by flex objects having a default min-width of auto rather than 0.
The overflow would also happen to all chip groups besides in the typeahead select (if they were too narrow) and this fixes that as well.
To test, look at chip groups and typeahead select and manually set a max-width on the container to be very narrow.

### DIFF
--- a/src/patternfly/components/Chip/chip.scss
+++ b/src/patternfly/components/Chip/chip.scss
@@ -43,6 +43,7 @@
   position: relative;
   display: inline-flex;
   align-items: center;
+  min-width: 0;
   padding: var(--pf-c-chip--PaddingTop) var(--pf-c-chip--PaddingRight) var(--pf-c-chip--PaddingBottom) var(--pf-c-chip--PaddingLeft);
   list-style: none;
   background-color: var(--pf-c-chip--BackgroundColor);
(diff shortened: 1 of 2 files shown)


mcoker (approved):
LGTM!!

Issue #4838, opened 2022-04-30
Bug - Dropdown - overlapped by sidebar


jowenn:
**Describe the problem**
I have  Dropdown within a toolbar in a a Masthead and a page sidebar/navigation panel. If the panel is open, the dropdown opens behind the sidebar.
--pf-c-page__sidebar--ZIndex: var(--pf-global--ZIndex--sm);
--pf-c-dropdown__menu--ZIndex: var(--pf-global--ZIndex--sm);
![menu_behind_sidebar](https://user-images.githubusercontent.com/5236596/166100702-a471c1ed-497a-4495-ae59-4aa4eddd53b6.jpg)

**How do you reproduce the problem?**
Provide steps to reproduce. A [codesandbox](https://codesandbox.io/s/nifty-mendeleev-yh97pg ) demonstrating the problem is appreciated.

**Expected behavior**
The menu should be completely visible

**Is this issue blocking you?**

If I change --pf-c-dropdown__menu--ZIndex: var(--pf-global--ZIndex--sm); to --pf-c-dropdown__menu--ZIndex: var(--pf-global--ZIndex--md); it works, but I think it should workout of the box.

**Screenshots**
If applicable, add screenshots to help explain the issue.

**What is your environment?**
 - OS: macOS Big Sur 11.6.5
 - Browser Chrome
 - Version Version 100.0.4896.127 (Offizieller Build) (x86_64)

**What is your product and what release date are you targeting?**

**Any other information?**

PR #4839, opened 2022-05-09
fix(page): add masthead to header/nav/main stacking context order

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4838


mattnolting (approved):
LPTM! 🎉

PR #4842, opened 2022-05-10
fix(charts): added dark theme tooltip border styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4841

PR #4843, opened 2022-05-10
fix(release-notes): added release notes for 2022.06

mcoker:
I'd still like to merge this charts PR and add it to the release notes - https://github.com/patternfly/patternfly/pull/4842

We can go ahead and merge this PR, then update the release notes with the charts PR when that merges, or hold off on this PR until the charts PR is merged.


nicolethoen (approved):
LGTM - only curious if the update to label should include a reference to the fact that it's enabling the Editable Label Group in case people are looking for that?

PR #4847, opened 2022-05-10
fix(drawer): removed min-height

### LINKED ISSUE #4797: Drawer body bottom padding
re: https://github.com/patternfly/patternfly/pull/4778/files#r848962284

Before removing `min-height: 0`, let's test and make sure this won't inadvertently cause a visual breaking change if we remove it, or see if there is possibly another way to fix the root issue.

mattnolting:
Fixes #4797 

As far as I can tell, this update not only fixed the `__body` padding issue, but also fixes an overlay issue caused by `min-height`. 

**Before** 

Screen Shot 2022-05-10 at 7 48 45 PM

**After**

Screen Shot 2022-05-10 at 7 49 20 PM

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -345,7 +345,6 @@ $pf-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 
 // Modified content children
 .pf-c-drawer__body {
-  min-height: 0;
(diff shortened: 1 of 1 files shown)


mcoker:
@mattnolting nice! Are you able to test this in console just to make sure any use of drawers still looks ok? IIRC, we added this around some support of notification drawer there.

mattnolting:
@mcoker It breaks the scrollability in OSC, although I do believe this to be an OpenShift specific issue as their page structure isn't quite a aligned w/PF. We'll need to update OpenShift before merging.

**Before:**

Screen Shot 2022-05-24 at 11 02 55 AM

**After:**

Screen Shot 2022-05-24 at 11 02 30 AM

mcoker:
closing PR for now - issue is marked for a breaking change release.

Issue #3409, opened 2020-08-19
Enable text truncation on label as default


mcoker:
Currently truncation is enabled by adding a `__text` container to the text. Let's update the label component so the `__text` container is used by default, which will match the chip component.

mcoker:
Closing in favor of https://github.com/patternfly/patternfly/issues/4795

PR #4849, opened 2022-05-11
fix(labels): updated label truncation

### LINKED ISSUE #4795: Bug - Label - Labels break out of container with long strings of text
**Describe the problem**
Labels exceed their container with long strings of text.

https://codesandbox.io/s/musing-villani-wtq9p3?file=/index.js

**Expected behavior**
Labels should truncate by default. The truncate variant should be updated to limit the default width of 100%. 

Yes, but can work around

mattnolting:
Closes #4795 

This PR 
- adds `.pf-m-auto-width` support to `.pf-c-label`, which sets `--pf-c-label__text--MaxWidth` to 100%. We would remove this at breaking change and make this the default. 
- adds `.pf-c-label__text` to labels by default, which allows the label text to truncate by default
- sets `.pf-c-label-group` and childrens' `max-width`s to ensure they don't exceed their containers

### DIFF
--- a/src/patternfly/components/Label/examples/Label.md
+++ b/src/patternfly/components/Label/examples/Label.md
@@ -12,36 +12,48 @@ import './Label.css'
 
 ```hbs
 {{#> label label--id="default-grey"}}
-  Grey
+  {{#> label-text}}
+    Grey
+  {{/label-text}}
 {{/label}}
 
 {{#> label label--id="default-grey-icon"}}
   {{#> label-icon}}
     <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
   {{/label-icon}}
-  Grey icon
+  {{#> label-text}}
+    Grey icon
+  {{/label-text}}
 {{/label}}
 
 {{#> label label--id="default-grey-close" label--isRemovable="true"}}
-  Grey removable
+  {{#> label-text}}
+    Grey removable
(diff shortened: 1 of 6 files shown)


mcoker (changes requested):
After making this change, is the intent that we would update the react component with this? Then remove the auto width modifier in a breaking change to make this the default behavior (satisfying https://github.com/patternfly/patternfly/issues/3409)?

Assuming so

* Adding `max-width: 100%` to the label is going to change core users' labels (who won't get the `__text` and `.pf-m-auto-width` update), by chopping off the extra text:
  * Screen Shot 2022-05-25 at 1 10 16 PM
* Moving the label children currently passed to `__content` to `__text` instead could have some unintended consequences since `__content` is currently a flex layout and `__text` is not. If this were an opt-in, that would be OK since you have the same issue between the default label and `isTruncated`, but `isTruncated` is an opt-in, so you have the opportunity to fix anything that happens when you add it. If this is the default, then any problem it caused would be considered breaking.
  * I suppose you could update `__text` to be a flex layout/align center with `.pf-m-auto-width`? Doesn't seem ideal, but maybe it would be OK.

mcoker:
Confirmed with @mattnolting that this will be best addressed in a breaking change release and that OpenShift is where the problem was spotted, and they have a workaround.

PR #4850, opened 2022-05-13
fix(tabs): added responsive bp to vert tabs

### LINKED ISSUE #3915: Tabs: add responsive support for vertical tabs
Need to consider responsive, vertical tabs. Vertical tab implementation will be primarily dependent on the `sidebar` component. The `sidebar` component switches from `vertical` to `horizontal` direction at `md` breakpoint. We either need to A) set vertical tabs orientation to take effect at the same breakpoint or B) allow user to set the breakpoint at which direction changes.

tabs

mattnolting:
Closes #3915 

This PR
- Adds demos for vertical tabs in sidebar component
- Adds demos for vertical, full height tabs in sidebar component
- Adds `.pf-m-full-height` to the sidebar component
- Wraps `.pf-c-tabs.pf-m-vertical` and variants in a `md` media query

Design files:
- https://marvelapp.com/prototype/dj6cag1/screen/87992106
- https://marvelapp.com/prototype/5fba0cj/screen/73740954

Previews:
- https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-tabs/
- https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-box-tabs
- https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-tabs-full-height/

### DIFF
--- a/src/patternfly/components/Sidebar/examples/Sidebar.md
+++ b/src/patternfly/components/Sidebar/examples/Sidebar.md
@@ -161,11 +161,11 @@ import './Sidebar.css'
 | `.pf-c-sidebar__panel` | `<div>` | Initiates the sidebar panel element. **Required** |
 | `.pf-c-sidebar__content` | `<div>` | Initiates the sidebar content element. **Required** |
 | `.pf-m-gutter` | `.pf-c-sidebar` | Modifies the sidebar component to add a gutter between the panel and content. |
-| `.pf-m-full-height` | `.pf-c-sidebar` | Modifies the sidebar and its children to full height of their container. |
 | `.pf-m-stack` | `.pf-c-sidebar` | Modifies the sidebar to stack the panel on top of the content. |
 | `.pf-m-split` | `.pf-c-sidebar` | Modifies the sidebar to position the panel and content side by side. |
 | `.pf-m-panel-right` | `.pf-c-sidebar` | Modifies the sidebar to place the panel to the right of the content. |
 | `.pf-m-sticky` | `.pf-c-sidebar__panel` | Modifies the panel to be sticky to the top of the layout. |
+| `.pf-m-full-height` | `.pf-c-sidebar__panel, .pf-c-sidebar__content [@md breakpoint]`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__panel`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__content` | Modifies the panel or content elment to be full height of its parent. |
 | `.pf-m-static` | `.pf-c-sidebar__panel` | Modifies the panel to be positioned statically. |
 | `.pf-m-width-{default, 25, 33, 50, 66, 75, 100}{-on-[breakpoint]}` | `.pf-c-sidebar__panel` | Modifies the panel width at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). **Note:** does not apply when the panel is stacked on top of the content. |
 | `.pf-m-no-background` | `.pf-c-sidebar`, `.pf-c-sidebar__panel, .pf-c-sidebar__content` | Modifies the element to have a transparent background. |
--- a/src/patternfly/components/Sidebar/examples/Sidebar.md
+++ b/src/patternfly/components/Sidebar/examples/Sidebar.md
@@ -161,6 +161,7 @@ import './Sidebar.css'
 | `.pf-c-sidebar__panel` | `<div>` | Initiates the sidebar panel element. **Required** |
 | `.pf-c-sidebar__content` | `<div>` | Initiates the sidebar content element. **Required** |
 | `.pf-m-gutter` | `.pf-c-sidebar` | Modifies the sidebar component to add a gutter between the panel and content. |
+| `.pf-m-full-height` | `.pf-c-sidebar` | Modifies the sidebar and its children to full height of their container. |
 | `.pf-m-stack` | `.pf-c-sidebar` | Modifies the sidebar to stack the panel on top of the content. |
 | `.pf-m-split` | `.pf-c-sidebar` | Modifies the sidebar to position the panel and content side by side. |
 | `.pf-m-panel-right` | `.pf-c-sidebar` | Modifies the sidebar to place the panel to the right of the content. |
--- a/src/patternfly/components/Sidebar/examples/Sidebar.md
+++ b/src/patternfly/components/Sidebar/examples/Sidebar.md
@@ -165,7 +165,7 @@ import './Sidebar.css'
 | `.pf-m-split` | `.pf-c-sidebar` | Modifies the sidebar to position the panel and content side by side. |
 | `.pf-m-panel-right` | `.pf-c-sidebar` | Modifies the sidebar to place the panel to the right of the content. |
 | `.pf-m-sticky` | `.pf-c-sidebar__panel` | Modifies the panel to be sticky to the top of the layout. |
-| `.pf-m-full-height` | `.pf-c-sidebar__panel, .pf-c-sidebar__content [@md breakpoint]`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__panel`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__content` | Modifies the panel or content elment to be full height of its parent. |
+| `.pf-m-full-height` | `.pf-c-sidebar__panel, .pf-c-sidebar__content [@md breakpoint]`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__panel`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__content` | Modifies the panel or content element to be full height of its parent. |
 | `.pf-m-static` | `.pf-c-sidebar__panel` | Modifies the panel to be positioned statically. |
 | `.pf-m-width-{default, 25, 33, 50, 66, 75, 100}{-on-[breakpoint]}` | `.pf-c-sidebar__panel` | Modifies the panel width at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). **Note:** does not apply when the panel is stacked on top of the content. |
 | `.pf-m-no-background` | `.pf-c-sidebar`, `.pf-c-sidebar__panel, .pf-c-sidebar__content` | Modifies the element to have a transparent background. |
--- a/src/patternfly/components/Sidebar/sidebar.scss
+++ b/src/patternfly/components/Sidebar/sidebar.scss
@@ -61,6 +61,16 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
 
   // Full height
   --pf-c-sidebar__main--child--m-full-height--AlignSelf: stretch;
+  --pf-c-tabs--m-vertical__list--before--BorderWidth: 0;
+
+  .pf-c-tabs {
+    --pf-c-tabs--m-vertical__list--before--BorderRightWidth: 0;
+
+    &.pf-m-box {
+      --pf-c-tabs__link--before--BorderRightWidth: 0;
+      --pf-c-tabs--m-box__link--last-child--before--BorderRightWidth: 0;
+    }
+  }
 
   @media (min-width: $pf-global--breakpoint--md) {
     --pf-c-sidebar__main--FlexDirection: var(--pf-c-sidebar__main--md--FlexDirection);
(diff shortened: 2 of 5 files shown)


src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -88,6 +88,16 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
     }
   }
 
+  &.pf-m-full-height {
+    --pf-c-sidebar__main--AlignItems: stretch;
```
mcoker, line 92: This is going to break the sticky behavior by default. I wonder if `stretch` should be the default, and we move sticky to a `[panel/content]-body` element that holds the stuff in those elements.

src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -88,6 +88,16 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
...
+
+    height: 100%;
+
+    .pf-c-sidebar__main {
```
mcoker, line 96: just a nit, you could also remove the `height: 100%` above and change this selector to this. maybe looks cleaner? could just be me, so up to you :)

```
&,
.pf-c-sidebar__main {
  height: 100%;
}
```

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -304,152 +304,156 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
-
-    display: inline-flex;
-    flex-direction: column;
-    height: 100%; // If not a flex child, set height
-    padding: 0; // Because vertical variant has no scroll buttons, reset padding
-    overflow: visible; // remove in breaking change release
-
-    .pf-c-tabs__list {
-      position: relative;
+  @media screen and (min-width: $pf-global--breakpoint--md) {
```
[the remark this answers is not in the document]
mcarrano: I feel like it's hard to answer this, in general. I think that answer depends on how many tabs you have. The examples in the demos work well because there are only two tabs. If there were more tabs, then changing to the expandable menu might be a better solution.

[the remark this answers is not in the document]
mmenestr: How do expandable vertical tabs work? When opened, does it push down the content, does it overlay it, or does it push it to the right? I'm assuming it pushes down the content when expanded, and clicking on a tab would change the content underneath, and you'd have to collapse the tabs or scroll down to see the content for that tab right?

If that's the case then yeah, I don't think that's too bad of an experience! It's almost like a navigation that you collapse and expand as needed.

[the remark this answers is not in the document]
mcoker: > I'm assuming it pushes down the content when expanded, and clicking on a tab would change the content underneath, and you'd have to collapse the tabs or scroll down to see the content for that tab right?

@mmenestr yep! https://www.patternfly.org/v4/components/tabs#vertical-expandable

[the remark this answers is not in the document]
mcoker: @mattnolting looks like expandable tabs also rely on `.pf-m-vertical`, so wrapping the media query around `.pf-m-vertical` breaks the ability for expandable to work.

I think we should group `.pf-m-vertical` and `.pf-m-expandable` for the base vertical styling, like we do in jump links. wdyt?

https://github.com/patternfly/patternfly/blob/dd3439944fa37e1bb97131c4a81b277a8eb78ed1/src/patternfly/components/JumpLinks/jump-links.scss#L89-L90

mattnolting:
@mcarrano The only demos that showcase the expandable, vertical tabs are the new demos in this PR. They are present in examples only. 

- https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-tabs/
- https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-tabs-full-height/

mcarrano:
@mattnolting I agree that these don't feel right. Frankly, I'm not sure exactly what the use case is for vertical tabs. It was a feature I think we added for parity with RH marketing sites. Maybe we should put this on hold until we can take a look in design?

mcoker:
@mattnolting ah, I see, thanks for the screenshots. I wonder if part of the problem there is that there are vertical tabs style variations applying to expandable tabs when it probably doesn't make sense for that to happen. Here's what the expandable tabs demo currently looks like

Screen Shot 2022-05-24 at 4 08 26 PM

Assuming that's correct, maybe we update expandable tabs to:
* Always use the default (don't apply "box" or "box-light" styles) when tabs are expandable. Any box or other styling like that will apply when the tabs are no longer expandable.
  * Or we could keep the box styling if the user specifies that, and remove the right vertical border? And maybe the inset? It just looks a little odd
  * Screen Shot 2022-05-24 at 4 13 54 PM
* Keep the inset fixed, or maybe even remove it
  * Here is what it looks like with the inset removed: 
    * Screen Shot 2022-05-24 at 4 16 31 PM
  * And with it set to 2xl - assuming the user set a 2xl inset, seems like we wouldn't want that when the list is expandable
    * Screen Shot 2022-05-24 at 4 16 49 PM
* We need to do something about the toggle button. Currently its position is offset so that the arrow in the button is left aligned with the tabs text.
  * That's creating this lack of space/padding around the toggle button:
    * Screen Shot 2022-05-24 at 4 19 35 PM
  * This is the actual toggle button area with the red background showing the toggle button and how its position is offset to align the arrow with the tab items:
    * Screen Shot 2022-05-24 at 4 20 09 PM

mcarrano:
OK. I've opened https://github.com/patternfly/patternfly-design/issues/1165 to shift this to design.

src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -231,6 +231,14 @@ import './Tabs.css'
 {{/tabs}}
 ```
 
+### Vertical, box, expanded
```
mcoker, line 234: total nit - We have "Vertical", "Box vertical", and "Vertical, box, expanded"

I wonder if that should be
* "Vertical", "Box vertical", and "Box vertical expanded"
* "Vertical", "Vertical box", and "Vertical box expanded"

Just looking at the rest of the examples, I would assume the latter. We could probably even take "horizontal" off of the example names that have them as that's the default, and prefix all of the vertical examples with "Vertical"?

The exception would be "horizontal overflow", but I dunno if that's the best description for that anyways. The default overflow is horizontal overflow, too, since we don't have vertical overflow. "horizontal overflow" has the overflow tab. Maybe that should be called "Overflow tab"? Just some thoughts, not to hold up or impact this PR.

mmenestr:
This looks good, and I agree with @mcoker on the re-naming !

src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -88,6 +88,15 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
     }
   }
 
+  &.pf-m-full-height {
+    --pf-c-sidebar__main--AlignItems: stretch;
```
mcoker, line 92: I'm not sure if this is the var we want to change. Currently since it comes before `.pf-m-split`, it's getting overridden because the split layout sets this, too, so it can switch from stretch to start at a medium breakpoint.

I wonder if `pf-m-full-height` might make more sense on the panel and content, so you can make one or the other full height? It's worth noting that when the panel is full height, it is longer sticky, so maybe it makes sense to be able to make the content full height for whatever reason, so the panel can still be sticky? Or this could come after the stack and split blocks, and maybe just set `--pf-c-sidebar--m-split__main--AlignItems: stretch` since that's what I believe this is targeting.

src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -148,6 +157,20 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
     --pf-c-sidebar__panel--Position: static;
     --pf-c-sidebar__panel--Top: auto;
   }
+
+  > .pf-c-tabs {
+    --pf-c-tabs--m-vertical--MaxWidth: auto; // when tabs are part of the sidebar, unset max width
```
mcoker, line 162: This was also set in the tabs component. Did you mean to just make these modifications here locally in the sidebar component? I know some people are using tabs in the sidebar now, even though the sidebar is beta. I wonder if it makes sense for this to be a variation? And/or the sidebar panel has width modifiers - maybe we should add an `auto` width modifier for things like this with a predefined width? Seems like adding both variations would make it the most flexible and be useful. Ultimately this should just be the default though, like jump links.

src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -148,6 +157,20 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
...
+
+    &.pf-m-vertical.pf-m-box {
+      --pf-c-tabs__link--before--BorderRightColor: transparent;
+      --pf-c-tabs--m-vertical__list--before--BorderColor: transparent;
```
mcoker, line 166: These borders really don't make sense when the tabs are expandable. I wonder if we could find out how people are using the expandable tabs list currently? I imagine the tabs shouldn't have a width constraint or the borders or anything, then we wouldn't have to recreate the borders here.

mcoker (changes requested):
Nice! I left some comments... it looks like there are some potential breaking changes, though I'm wondering if those are intentional as they're bug fixes? I reviewed most everything, but wanted to get clarification on the things that may be breaking first.

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -16,15 +16,26 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs--m-page-insets--inset: var(--pf-global--spacer--md); // make this the default inset at breaking change
   --pf-c-tabs--m-page-insets--xl--inset: var(--pf-global--spacer--lg); // make this the default inset at breaking change
 
+  // Expandable
+  --pf-c-tabs--m-expandable__toggle--MarginTop: var(--pf-global--spacer--sm);
```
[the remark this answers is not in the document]
mattnolting: I suppose it _could_ be, but it looks broken currently and this matches the design.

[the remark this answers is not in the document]
mcoker: I agree, but if folks are using it now, I think it's going to move things around unexpectedly. This and the unconstrained width and some other changes. Maybe we could have a quick check-in with design and go over the changes and see if we need some sort of opt-in, or if we can go ahead and change the layout as bug fixes?

srambach:
Should the vertical line be visible on https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-tabs/ ?

src/patternfly/components/Sidebar/examples/Sidebar.md
```diff
@@ -165,6 +165,7 @@ import './Sidebar.css'
 | `.pf-m-split` | `.pf-c-sidebar` | Modifies the sidebar to position the panel and content side by side. |
 | `.pf-m-panel-right` | `.pf-c-sidebar` | Modifies the sidebar to place the panel to the right of the content. |
 | `.pf-m-sticky` | `.pf-c-sidebar__panel` | Modifies the panel to be sticky to the top of the layout. |
+| `.pf-m-full-height` | `.pf-c-sidebar__panel, .pf-c-sidebar__content [@md breakpoint]`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__panel`, `.pf-c-sidebar.pf-m-split .pf-c-sidebar__content` | Modifies the panel or content elment to be full height of its parent. |
```
srambach, line 168: typo "element"

mattnolting:
> Should the vertical line be visible on https://patternfly-pr-4850.surge.sh/components/tabs/html-demos/vertical-tabs/ ?
> 
> You can (not!) see it here: https://user-images.githubusercontent.com/19825616/189410702-840a68a5-af20-447a-8798-104167553282.mp4

@mcarrano wdyt?

mcarrano:
@mattnolting this demo feels kind of odd to me. I don't think that we would recommend using vertical tabs inside of a card like that. @mmenestr @doruskova do we have a mockup for a full page example of using vertical tabs that can be used here instead?

mmenestr:
@mcarrano I don't personally. I guess we'd usually use jump links in this kind of scenario on a long scrolling form, instead of vertical tabs. I honestly can't think of a scenario where I'd recommend vertical tabs over horizontal tabs so I'm not sure how I'd go about demo-ing it in a "realistic" way

doruskova:
@mcarrano OpenShift is using vertical tabs here http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/user-preferences/ and Satellite is using them here https://marvelapp.com/prototype/1121a8de/screen/72413287

mattnolting:
> @mcarrano OpenShift is using vertical tabs here http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/user-preferences/ and Satellite is using them here https://marvelapp.com/prototype/1121a8de/screen/72413287

@doruskova Thanks for the refs! Looks like OS does not have a vertical divider and satellite does. Which is correct?

doruskova:
@mattnolting The correct one is with a vertical divider. OpenShift has a divider as well. http://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/user-preferences/

srambach:
I'm not sure my original question was clear - I was asking about line at the smallest breakpoint, where you get hover and selection left border but they aren't connected by the faint gray line. It seems like the answer was yes to vertical line, but I don't see that connecting line at the smallest breakpoint. If that's the desired outcome, it's fine with me but it seemed off.
image

mattnolting:
@mcoker @srambach I've updated the vertical, expandable tabs, box variant support. The extended borders are drawn by the sidebar component.

srambach:
I'm not sure if it was resolved above, but this alignment looks off to me. If it's ok with design, then no problem. 
image

mmenestr:
@srambach which example is that? But you're right I think - looking at the expanded tab example the arrow is like slightly to the right of the 'left line' that borders the tabs

Issue #2702, opened 2020-02-12
Spike: investigate solutions for displaying icons in validated text inputs


jeff-phillips-18:
**Describe the issue. What is the expected and unexpected behavior?**

Setting the validated field on a `TextInput` should show the appropriate icon to the far right per Patternfly core:

https://pf4.patternfly.org/documentation/core/components/form#help-text
![image](https://user-images.githubusercontent.com/11633780/74340996-fe869780-4d74-11ea-811e-0b6384c3eed8.png)

Instead it is not showing:

https://patternfly-react.surge.sh/documentation/react/components/form#validated
![image](https://user-images.githubusercontent.com/11633780/74341455-c03da800-4d75-11ea-90ce-fad6fa01ed5d.png)

**Is this a bug or enhancement? If this issue is a bug, is this issue blocking you or is there a work-around?**

Bug, no work around.

**What is your product and what release version are you targeting?**

Openshift Console, likely 4.5

cc @serenamarie125 @christianvogt

mcoker:
Looks like this is a CSS problem and only occurs when the form control's value is selected from autofill. Chrome is applying a `background` with `!important`, that overrides the background image we use for the icon.

mcoker:
A potential fix - https://codepen.io/mcoker/pen/rNVegEX

jeff-phillips-18:
@mcoker The work around you had no longer works. You can see the problem in https://codesandbox.io/s/autofill-background-issues-2wrejy?file=/src/styles.css

Do you have any suggestions?

jerolimov:
This is the style that is now added by Chrome:

![image](https://user-images.githubusercontent.com/139310/161988652-8362f22b-68bb-4c34-8aa3-f01dfe0f3bfc.png)

Changing the CSS rules from the previous fix - https://codepen.io/mcoker/pen/rNVegEX from `-webkit-autofill` to `-internal-autofill-selected` unfortunately doesn't work.

srambach:
As noted above, known solutions trying to maintain a background image on an input while Chrome autocomplete removes it have all stopped working. At this point, the only known solution is to restructure the HTML to add the image as a separate element rather than a background image on the input. 
For now, we'll add a note on the form field component and mark this to be fixed in a breaking change.

nicolethoen:
The TextInput was designed for use in a form element. The TextInputGroup was not - and does not currently have a concept of validation states built into it.

[...]

PR #4856, opened 2022-05-16
chore(form field): note autocomplete limitation

### LINKED ISSUE #4858: Bug - Form input - status icons disappear on autofill
**Describe the problem**
Add documentation to provide a workaround to input fields that have status icons, which get overwritten when Chrome autofill is used.

This is related to #2702, which will be addressed in a breaking change.

srambach:
Fixes #4858 

Adds documentation relating to #2702, which will be fixed in a future breaking change.

Adds a note on form field component saying that form fields with status icons should turn off autocomplete or use dynamic helper text to indicate status because of a bug/limitation in webkit.

### DIFF
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -8,6 +8,9 @@ import './FormControl.css'
 
 ## Examples
 ### Input
+
+**Note:** In webkit browsers, inputs with status icons that are autocompleted will have their icons removed by the user agent stylesheet. To ensure that status will still be shown if the input is filled via autocomplete, turn off autocomplete with `autocomplete="off"` or use [dynamic helper text](/components/helper-text/html-demos) instead.
+
 ```hbs
 {{> form-control controlType="input" input="true" form-control--attribute='type="text" value="Standard" id="input-standard" aria-label="Standard input example"'}}
 <br><br>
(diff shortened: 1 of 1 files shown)


mcarrano:
@srambach I think this is fine, however I'm wondering if we need this documentation on the React component page also since that's how most developers consume PatternFly components. Thoughts @mcoker @nicolethoen @tlabaj ?

srambach:
@mcarrano As soon as we agree on whether this is accepted documentation I can make a follow-on issue for patternfly-react.

jerolimov:
> To ensure that status will still be shown if the input is filled via autocomplete, turn off autocomplete

When following this recommendation the user doesn't have the option to autocomplete the field. Sure, the icon is always visible, but "the input is filled via autocomplete" will never happen then, right? :thinking:

srambach:
> > To ensure that status will still be shown if the input is filled via autocomplete, turn off autocomplete
> 
> When following this recommendation the user doesn't have the option to autocomplete the field. Sure, the icon is always visible, but "the input is filled via autocomplete" will never happen then, right? 🤔

How about this instead @jerolimov?
If the field does not need to use autocomplete, turn it off with `autocomplete="off"` to avoid the problem. Otherwise, use [dynamic helper text](/components/helper-text/html-demos) instead to ensure that the status will still be visible if the field is autocompleted.

src/patternfly/components/FormControl/examples/FormControl.md
```diff
@@ -8,6 +8,9 @@ import './FormControl.css'
 
 ## Examples
 ### Input
+
+**Note:** In webkit browsers, inputs with status icons that are autocompleted will have their icons removed by the user agent stylesheet. To ensure that status will still be shown if the input is filled via autocomplete, turn off autocomplete with `autocomplete="off"` or use [dynamic helper text](/components/helper-text/html-demos) instead.
```
mcoker, line 12: Does it need to be dynamic, or could helper text with a status icon also work?

mcoker (commented):
lgtm! Just one comment.

mcoker (approved):
LGTM!

PR #4857, opened 2022-05-17
fix(table): added tr border to sticky nested table header

### LINKED ISSUE #4846: Bug: Nested table header breaks with sticky header
When scrolling starts in the table the headers that do not have nested subheadings appear broken out above the normal heading.
Screen Shot 2022-03-07 at 12 41 51 PM

An example was prepared and can be found here https://codesandbox.io/s/vibrant-goldstine-1t2b68?file=/index.tsx

The nested table header is used on the OpenShift Network Observability operator. This is in Tech preview now and scheduled for full release with 4.11.

mattnolting:
closes #4846 

https://patternfly-pr-4857.surge.sh/components/table#nested-column-headers
https://patternfly-pr-4857.surge.sh/components/table#nested-column-headers-sticky-header

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -3556,7 +3556,7 @@ For sticky columns to function correctly, the parent table's width must be contr
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-m-nested-column-header` | `<thead>` | Modifies a table header to handle nested header cells. |
-| `.pf-m-border` | `.pf-m-sticky-header > .pf-m-nested-column-header > tr` | Modifies the last row in a sticky header with nested column headers to display as a border. |
+| `.pf-m-border-row` | `.pf-m-sticky-header > .pf-m-nested-column-header > tr` | Modifies the last row in a sticky header with nested column headers to display as a border. |
 | `.pf-m-border-right` | `<th>`, `<td>` | Modifies a table cell to show a right border. |
 | `.pf-m-border-left` | `<th>`, `<td>` | Modifies a table cell to show a left border. |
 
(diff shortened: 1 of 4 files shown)


kmcfaul:
It looks like in the preview the new example is rendering the hbs as text instead of visually.

I've got a react update in place locally for when this PR goes in as well.

src/patternfly/components/Table/table-tr.hbs
```diff
@@ -26,14 +27,20 @@
...
+  {{/unless}}
   {{#if table-tr--IsHoverable}}
     tabindex="0"
   {{/if}}
   {{#if table-tr--attribute}}
     {{{table-tr--attribute}}}
   {{/if}}>
-  {{> @partial-block}}
+  {{#if table-tr--IsBorderRow}}
+    {{> table-td table-td--IsEmpty="true" table-td--attribute='colspan="100%" aria-hidden="true" tabindex="-1"'}}
```
mcoker, line 42: > `colspan="100%"`

Is that valid? From MDN, looks like this needs to be an integer https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#attributes

src/patternfly/components/Table/examples/Table.md
```diff
@@ -3473,67 +3473,79 @@ For sticky columns to function correctly, the parent table's width must be contr
...
+      {{> table-tr--nested-column-header table-tr--index="10"}}
+    {{/table-tbody}}
+  {{/table}}
+</div>
+```
+
+### Nested column headers, sticky header
+```hbs
+<div class="pf-c-scroll-inner-wrapper">
+  {{#> table table--id="table-nested-column-headers-sticky-header-example" table--modifier="pf-m-grid-md pf-m-sticky-header" table--grid="true" table--attribute='aria-label="This is a nested column header table example"'}}
```
mcoker, line 3495: nit - 

```suggestion
  {{#> table table--id="table-nested-column-headers-sticky-header-example" table--modifier="pf-m-grid-md pf-m-sticky-header" table--grid="true" table--attribute='aria-label="This is a sticky nested column header table example"'}}
```

mcoker (commented):
LGTM! 2 small comments.

mcoker (approved):
L🍕TM!

Issue #3256, opened 2020-07-07
Styling the date/time picker


mcoker:
Let's investigate the work @nicolethoen has put together so far with the date and time picker, and see what is involved in styling it as close as we can to the proposed design - the best approach to take to styling it (can we modify the class structure, or do we apply overrides)
* are there any established best practices to styling it already out there
* how would we introduce the styling into PF since this will be a bit different
* what are the limitations to our ability to style based on the intended design, and are we able to find a compromise 

Designs - https://docs.google.com/document/d/1hlS0R1l7gnoe-YDu1SEHpUSrf6WMuZPejR5AwYr0a20/edit
Library - https://flatpickr.js.org
Original feature issue - https://github.com/patternfly/patternfly/issues/180

A [POC was begun](https://github.com/patternfly/patternfly-react/issues/2010) - the code for a preliminary/experimental Date and Time picker can be found on [this branch](https://github.com/nicolethoen/patternfly-react/tree/add_date_time_picker).

For the sake of experimenting, a very basic implementation of flatpickr with patternfly loaded in is running in [this sandbox.](https://codesandbox.io/s/silly-feynman-o0vh3)

cc @mcarrano

mcarrano:
@mcoker if any design questions come up while I am away next week, please reach out to @maryshak1996 who worked on the visual styling in the attached design doc.

mcoker:
@mcarrano looks like the picker is well marked up with flexible classes, and is easily styled. The text input can be styled like our form control component. The colors, padding, etc are all customizable. From looking at some of the [markup we've created](https://github.com/nicolethoen/patternfly-react/blob/add_date_time_picker/packages/react-core/src/components/DateTimePicker/TimePicker.tsx), we can use the markup to allow us to customize it in most of the ways we'll need.

One thing I wasn't sure about was our design for the date-range selection, but I just mocked up our basic design from [flatpickr's range calendar example](https://flatpickr.js.org/examples/#range-calendar) in dev tools and looks like that should work no problem:

<img width="320" alt="Screen Shot 2020-08-06 at 3 39 22 PM" src="https://user-images.githubusercontent.com/35148959/89581710-0a6f8580-d7fd-11ea-8b7e-e1e4eb302591.png">

Some things I'm not sure about:

In our mockups, we have dropdowns for the month and year selection in the calendar

[...]

PR #4862, opened 2022-05-18
fix(CalendarMonth): improve layout of nav controls

### LINKED ISSUE #4760: Calendar month layout improvements
1. There is a margin between the prev/next month buttons that is not clickable, and if the margin is removed, the icon aligns better with the first/last columns in the calendar grid
1. There is a double border between the month select and year text input. Those should be in an input group.

#### Before
Screen Shot 2022-03-28 at 10 37 24 AM
Screen Shot 2022-03-28 at 10 37 28 AM

#### After
Screen Shot 2022-03-28 at 10 39 09 AM
Screen Shot 2022-03-28 at 10 39 15 AM

thatblindgeye:
Closes #4760 

[Calendar Month preview build](https://patternfly-pr-4862.surge.sh/components/calendar-month)

### DIFF
--- a/src/patternfly/components/CalendarMonth/calendar-month.scss
+++ b/src/patternfly/components/CalendarMonth/calendar-month.scss
@@ -80,6 +80,16 @@
 .pf-c-calendar-month__header-nav-control {
   margin-right: var(--pf-c-calendar-month__header-nav-control--MarginRight);
   margin-left: var(--pf-c-calendar-month__header-nav-control--MarginLeft);
+
+  &.pf-m-prev-month {
+    --pf-c-calendar-month__header-nav-control--MarginRight: var(--pf-c-calendar-month__header-nav-control--m-prev-month--MarginRight);
+    --pf-c-calendar-month__header-nav-control--MarginLeft: var(--pf-c-calendar-month__header-nav-control--m-prev-month--MarginLeft);
+  }
+
+  &.pf-m-next-month {
+    --pf-c-calendar-month__header-nav-control--MarginRight: var(--pf-c-calendar-month__header-nav-control--m-next-month--MarginRight);
+    --pf-c-calendar-month__header-nav-control--MarginLeft: var(--pf-c-calendar-month__header-nav-control--m-next-month--MarginLeft);
+  }
 }
 
 .pf-c-calendar-month__header-month {
--- a/src/patternfly/components/CalendarMonth/calendar-month.scss
+++ b/src/patternfly/components/CalendarMonth/calendar-month.scss
@@ -11,10 +11,10 @@
   --pf-c-calendar-month__header-year--Width: 8ch;
   --pf-c-calendar-month__header-nav-control--MarginRight: 0;
   --pf-c-calendar-month__header-nav-control--MarginLeft: 0;
-  --pf-c-calendar-month__header-nav-control--m-prev-month--MarginRight: var(--pf-global--spacer--sm);
-  --pf-c-calendar-month__header-nav-control--m-prev-month--MarginLeft: calc(var(--pf-global--spacer--md) * -1);
-  --pf-c-calendar-month__header-nav-control--m-next-month--MarginRight: calc(var(--pf-global--spacer--md) * -1);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/CalendarMonth/calendar-month.scss
```diff
@@ -80,16 +80,6 @@
 .pf-c-calendar-month__header-nav-control {
   margin-right: var(--pf-c-calendar-month__header-nav-control--MarginRight);
   margin-left: var(--pf-c-calendar-month__header-nav-control--MarginLeft);
-
-  &.pf-m-prev-month {
-    --pf-c-calendar-month__header-nav-control--MarginRight: var(--pf-c-calendar-month__header-nav-control--m-prev-month--MarginRight);
```
mcoker, line 85: we actually still need these - it's part of preserving the CSS var API - if we remove the use of them in the code, they no longer work if a user has defined them with a custom margin. That's effectively why we preserve the variables<sup>*</sup>, so they can be used.

<sup>*</sup> I think a better approach would be to remove the variable declaration we've marked for removal, and just use "0" (or whatever the variable has been set to) as a fallback, and mark this for removal. We could also do the same with `--pf-c-calendar-month__header-nav-control--Margin[Left/Right]` since those are no longer necessary, either. That would save let us remove some unnecessary code. Though that's a separate issue, we could do it here - wdyt @mattnolting @srambach?

mcoker (changes requested):
Looks great, just one thing! about still needing to use the vars we're going to remove. What we don't want to break is the ability for a user to define the var on their end to provide a custom margin. I left a larger question there, too, though that's a separate issue - no reason the answer should hold up this PR if you wanna go ahead and merge this.

src/patternfly/components/CalendarMonth/calendar-month.scss
srambach: I like the falback solution @mcoker 
If there isn't one already, should we agree on a standard comment adjacent so that we can easily find them to remove?

mcarrano:
>  I did notice that there is now a little less space for the month/year controls because the left/right buttons previously had negative outer margins, pushing them outside of the component layout a bit, creating more space for the controls.

@mcoker I don't think that will be a problem, at least not for English. But @thatblindgeye pointed out to me that currently the core and React versions of the calendar use different icons to page through the months. The React version actually looks truer to the original designs. Do you know why core is different?

mmenestr:
Agreed, I don't think the spacing in the demo is problematic!

mcoker:
> Do you know why core is different?

@mcarrano from the design doc in the date/time picker core issue, the calendar has the icons used in core - https://github.com/patternfly/patternfly/issues/3256

mcarrano:
> @mcarrano from the design doc in the date/time picker core issue, the calendar has the icons used in core - https://github.com/patternfly/patternfly/issues/3256

Hmmm... That's interesting. The Sketch Kit has the arrow icons so somewhere along the way this got changed. I could live with either icon version. @mmenestr sounds like you prefer the left/right caret? @mceledonia do you have any thoughts? Compare https://www.patternfly.org/v4/components/calendar-month vs https://www.patternfly.org/v4/components/calendar-month/html

mceledonia:
The caret icon is more in line with our typical usage, normally we use the arrow icons for more navigational purposes and the carets for flipping through content. For example the [pagination](https://www.patternfly.org/v4/components/pagination) carets versus the [call to action button](https://www.patternfly.org/v4/components/button#call-to-action)

src/patternfly/components/CalendarMonth/calendar-month.scss
mattnolting: @mcoker I think that, in general, this is a better approach than setting vars marked for deprecation to `0`. @thatblindgeye Not a blocker, but feel free to make that update!

mattnolting (approved):
LPTM!

mmenestr:
@mcarrano just opened this issue: https://github.com/patternfly/patternfly-org/issues/2973

mcoker (approved):
🥳

PR #4868, opened 2022-05-23
feat(utilities): add font family utilities

### LINKED ISSUE #4764: Add font family utility class
re: https://github.com/patternfly/patternfly/issues/4763

We're missing a font-family utility class. Considering we have a text, heading, and monospace font, and these variations in a variable font, it seems like it would be useful to be able to specify a font family on a block.

srambach:
Adds utility classes for font families san serif, heading san serif, and monospace, and variable fonts of each of those. Breakpoint modifiers are not included.

Fixes #4764

### DIFF
--- a/src/patternfly/utilities/Text/text.scss
+++ b/src/patternfly/utilities/Text/text.scss
@@ -1,5 +1,29 @@
 // stylelint-disable
 
+// Font family options
+$pf-u-font-family-options: (
+  font-family-sans-serif: (
+    font-family var(--pf-global--FontFamily--sans-serif)
+  ),
+  font-family-heading-sans-serif: (
+    font-family var(--pf-global--FontFamily--sans-serif)
+  ),
+  font-family-monospace: (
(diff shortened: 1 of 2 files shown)


src/patternfly/utilities/Text/text.scss
```diff
@@ -1,5 +1,29 @@
 // stylelint-disable
 
+// Font family options
+$pf-u-font-family-options: (
+  font-family-sans-serif: (
+    font-family var(--pf-global--FontFamily--sans-serif)
+  ),
+  font-family-heading-sans-serif: (
+    font-family var(--pf-global--FontFamily--sans-serif)
```
mcoker, line 9:
```suggestion
    font-family var(--pf-global--FontFamily--heading--sans-serif)
```

mcoker (changes requested):
Fantastic! Just one thing.

mcoker (approved):
LGTM!

PR #4869, opened 2022-05-24
feat(tree-view): add selectable/expandable, improve click areas

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4724
fixes https://github.com/patternfly/patternfly/issues/4725

Makes the node a `div` when the node is selectable (including nodes with checkboxes), and makes the toggle a `button`. This will require a react update to only allow a selectable node to expand when clicking on the node-toggle, otherwise clicking on the node will select it.

This also changes the icon color on hover/focus/active when clicking will expand the node
* hover/focus/active anywhere on a non-selectable node
* hover/focus/active on the toggle on a selectable node

In the original issue, we said that clicking anywhere to the left of a selectable/expandable toggle would also toggle the row, but decided against that since it isn't necessarily intuitive, and it would also mean that if you focused on a nested toggle, it would draw a focus outline around all of the area to the left of the toggle, which doesn't seem ideal or a pattern we have anywhere else as far as I'm aware. @mcarrano @mmenestr let me know if you want to change that. Right now the icon changes color when you're hovered/focused on the part of the node that will expand it.

### DIFF
--- a/src/patternfly/components/TreeView/tree-view-node.hbs
+++ b/src/patternfly/components/TreeView/tree-view-node.hbs
@@ -1,7 +1,5 @@
 {{#if tree-view-node--HasCheckbox}}
   {{> __tree-view-node tree-view-node--type="label"}}
-{{else if tree-view-node--IsSelectable}}
-  {{> __tree-view-node tree-view-node--type="div"}}
 {{else}}
   {{> __tree-view-node}}
 {{/if}}
@@ -13,6 +11,7 @@
     {{/if}}
     {{#if tree-view-node--HasCheckbox}}
       for="check-{{{tree-view-node--id}}}"
+      id="label-{{{tree-view-node--id}}}"
     {{/if}}
     {{#if tree-view-node--attribute}}
       {{{tree-view-node--attribute}}}
--- a/src/patternfly/components/TreeView/tree-view-node.hbs
+++ b/src/patternfly/components/TreeView/tree-view-node.hbs
@@ -1,27 +1,37 @@
-<{{# if tree-view-node--type--IsDiv}}div{{else}}button{{/if}} class="pf-c-tree-view__node{{#if tree-view-node--modifier}} {{tree-view-node--modifier}}{{/if}}"
-  {{#if tree-view-node--type--IsDiv}}
-    tabindex="0"
-  {{/if}}
-  {{#if tree-view-node--attribute}}
-    {{{tree-view-node--attribute}}}
-  {{/if}}>
-  {{#> tree-view-node-container}}
-    {{#if tree-view-list-item--IsExpandable}}
-      {{> tree-view-node-toggle}}
+{{#if tree-view-node--HasCheckbox}}
+  {{> __tree-view-node tree-view-node--IsDiv="true"}}
+{{else if tree-view-node--IsSelectable}}
+  {{> __tree-view-node tree-view-node--IsDiv="true"}}
+{{else}}
+  {{> __tree-view-node}}
+{{/if}}
+
+{{#* inline "__tree-view-node"}}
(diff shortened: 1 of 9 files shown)


mmenestr:
This looks good! Just have one comment; I notice that clicking anywhere to the right of the text does nothing, which makes sense. The only thing is that the cursor still retains the "can click" state. Is that because of the hover state that remains, despite hovering on a technically non clickable area?

I'm wondering if the cursor should change when a user is on top of any "non-clickable" area?

mcoker:
> This looks good! Just have one comment; I notice that clicking anywhere to the right of the text does nothing, which makes sense. The only thing is that the cursor still retains the "can click" state. Is that because of the hover state that remains, despite hovering on a technically non clickable area?

@mmenestr good point. Since this is a core demo, node selection isn't going to be fully accurate since we have click event handlers from react that will update tree view nodes when you click to select them.

Assuming you're specifically talking about the checkbox nodes, I was going to leave it up to the react component to capture the click event anywhere on the node - except for the toggle button - and mark the row as selected (toggle the checkbox "checked" state). However I just pushed an update that makes the entire node the `<label>` associated with the checkbox, so clicking anywhere on the node (except for the toggle button) will toggle the checkboxes checked state. Curious what @mattnolting and @tlabaj think of that, if there is any downside to that. I suspect it's fine since non-checkbox nodes are just a big `<button>`, which basically behaves the same as the `<label>` here.

mattnolting (commented):
Looks like you have some duplicate id's, but other than that, the html is valid and this works well. The only other thing I noticed is a checkbox flicker on toggle click. 

Screen Recording 2022-05-25 at 2 08 18 PM

mcoker:
@mattnolting updated!

> The only other thing I noticed is a checkbox flicker on toggle click.

Hmm... that's how a checkbox behaves by default when it's `:active` because you clicked on something else inside the `<label>` - https://codepen.io/mcoker/pen/vYdWjpm. Though that is not ideal, even if it is the default behavior. I'm not sure how to get around that.

I guess that begs the question of whether the button belongs in the `<label>` or not.

If so - can we live with that flicker until we can hopefully figure out a way to address it, if possible?

If not - we could move the `<label>` back to the text, and make the parent node a `<div>`, which would allow us to either:
1. Have JS handle the click event on the `<div>` node to select the row and leave everything as it is.
2. Add a `::before` to the `<label>` node-text that spans the entire node to allow clicking anywhere in the node to trigger a click on the `<label>`.
  * Give the `::before` a negative `z-index` to push it behind the toggle and whatever else may be in the label, and add a benign `transform` or something to the parent node element to give it a stacking context so that the negative `z-index` doesn't doesn't go behind the node parent. 

If we go with 2 above, we could probably do the same to regular tree view nodes and change the node parent from a `<button>` to a `<div>`, and add a `<button>` inside the node that wraps the node content and is adjacent to the toggle (the toggle could also be a `<button>` now), and use a pseudo element on the node content `<button>` to capture the click event of the entire node.

WDYT?

mcoker:
Also I think this is going to require accompanying react changes, so while this issue's progress can stay on track for the 2020.07 milestone, I don't think we'll merge/release it until 2020.08. cc @mcarrano

src/patternfly/components/TreeView/tree-view-node.hbs
```diff
@@ -1,27 +1,41 @@
...
+        {{#> tree-view-node-text}}
+          {{{tree-view-node--text}}}
+        {{/tree-view-node-text}}
+      {{/if}}
+      {{#if @partial-block}}
+        {{> @partial-block}}
+      {{/if}}
+    {{/tree-view-node-container}}
+  </{{# if tree-view-node--type}}{{tree-view-node--type}}{{else}}button{{/if}}>
+{{/inline}}
```
mattnolting, line 41:
```suggestion
{{/inline}}

```

mattnolting (approved):
LGTM, as long as we can have the `:active` state of the `checkbox` controlled by JS when the nested button is clicked.

mmenestr (approved):

mmenestr (approved):

PR #4876, opened 2022-05-26
feat(tabs): add expandable overflow styling

### LINKED ISSUE #4800: Tabs - introduce horizontal tab overflow
Enhancement to Tabs

**Describe the feature**
Want to support horizontal tab overflow without scrolling per this design: https://marvelapp.com/prototype/hi2b48a/section/1568195 We should consider whether this should be implemented as a variant of horizontal tabs or more generally as an Overflow component as was proposed here: https://github.com/patternfly/patternfly/issues/1625

wise-king-sullyman:
Closes #4800

Convenience link: https://patternfly-pr-4876.surge.sh/components/tabs#horizontal-overflow

### DIFF
--- a/src/patternfly/components/Tabs/__tabs-list.hbs
+++ b/src/patternfly/components/Tabs/__tabs-list.hbs
@@ -32,18 +32,13 @@ __tabs-list--IsCloseDisabled: bool
     __tabs-item--aria-label="Users"
     __tabs-item--text="Users"
     }}
-  {{#> wrapper
+  {{> __tabs-item
+    __tabs-item--current=(inverse __tabs-list--IsOverflowSelected)
     __tabs-item--id="containers"
     __tabs-item--icon-name="fas fa-box"
     __tabs-item--aria-label="Containers"
     __tabs-item--text="Containers"
     }}
-    {{#if __tabs-list--IsOverflowSelected}}
-      {{> __tabs-item }}
-    {{else}}
-      {{> __tabs-item __tabs-item--current="true"}}
-    {{/if}}
-  {{/wrapper}}
   {{> __tabs-item
     __tabs-item--id="database"
     __tabs-item--icon-name="database"
--- a/src/patternfly/components/Tabs/__tabs-list.hbs
+++ b/src/patternfly/components/Tabs/__tabs-list.hbs
@@ -6,6 +6,7 @@ __tabs-list--HasIcons: bool
 __tabs-list--IsScrollable: bool
 __tabs-list--IsOverflow: bool
 __tabs-list--IsOverflowExpanded: bool
+__tabs-list--IsOverflowSelected: bool
 __tabs-list--IsShort: bool
 __tabs-list--IsLong: bool
 __tabs-item--NoText: bool
@@ -31,13 +32,22 @@ __tabs-list--IsCloseDisabled: bool
     __tabs-item--aria-label="Users"
     __tabs-item--text="Users"
     }}
-  {{> __tabs-item
+  {{#if __tabs-list--IsOverflowSelected}}
+      {{> __tabs-item
+    __tabs-item--id="containers"
+    __tabs-item--icon-name="fas fa-box"
+    __tabs-item--aria-label="Containers"
+    __tabs-item--text="Containers"
+    }}
+  {{else}}
+      {{> __tabs-item
     __tabs-item--current="true"
     __tabs-item--id="containers"
     __tabs-item--icon-name="fas fa-box"
     __tabs-item--aria-label="Containers"
     __tabs-item--text="Containers"
     }}
+  {{/if}}
   {{> __tabs-item
     __tabs-item--id="database"
     __tabs-item--icon-name="database"
@@ -113,6 +123,7 @@ __tabs-list--IsCloseDisabled: bool
     {{/if}}
     {{#if __tabs-list--IsOverflow}}
       {{#> wrapper
+        __tabs-item--current=__tabs-list--IsOverflowSelected
         __tabs-item--id="more"
         __tabs-item--icon-name="more"
         __tabs-item--aria-label="Show more tabs"
--- a/src/patternfly/components/Tabs/__tabs-list.hbs
+++ b/src/patternfly/components/Tabs/__tabs-list.hbs
@@ -32,22 +32,18 @@ __tabs-list--IsCloseDisabled: bool
     __tabs-item--aria-label="Users"
     __tabs-item--text="Users"
     }}
-  {{#if __tabs-list--IsOverflowSelected}}
-      {{> __tabs-item
+  {{#> wrapper
     __tabs-item--id="containers"
     __tabs-item--icon-name="fas fa-box"
     __tabs-item--aria-label="Containers"
     __tabs-item--text="Containers"
     }}
-  {{else}}
-      {{> __tabs-item
-    __tabs-item--current="true"
-    __tabs-item--id="containers"
-    __tabs-item--icon-name="fas fa-box"
-    __tabs-item--aria-label="Containers"
-    __tabs-item--text="Containers"
-    }}
-  {{/if}}
+    {{#if __tabs-list--IsOverflowSelected}}
+      {{> __tabs-item }}
+    {{else}}
+      {{> __tabs-item __tabs-item--current="true"}}
+    {{/if}}
+  {{/wrapper}}
   {{> __tabs-item
     __tabs-item--id="database"
     __tabs-item--icon-name="database"
--- a/src/patternfly/components/Tabs/__tabs-list.hbs
+++ b/src/patternfly/components/Tabs/__tabs-list.hbs
@@ -112,24 +112,16 @@ __tabs-list--IsCloseDisabled: bool
         }}
     {{/if}}
     {{#if __tabs-list--IsOverflow}}
-      {{#if __tabs-list--IsOverflowExpanded}}
-        {{> __tabs-item
-          __tabs-item--id="more"
-          __tabs-item--icon-name="more"
-          __tabs-item--aria-label="Show more tabs"
-          __tabs-item--text="More"
-          __tabs-item--IsOverflow="true"
-          __tabs-item--IsOverflowExpanded="true"
+      {{#> wrapper
+        __tabs-item--id="more"
+        __tabs-item--icon-name="more"
+        __tabs-item--aria-label="Show more tabs"
+        __tabs-item--text="More"
+        __tabs-item--IsOverflow="true"
+        __tabs-item--IsOverflowExpanded=__tabs-list--IsOverflowExpanded
         }}
-      {{else}}
-        {{> __tabs-item
-          __tabs-item--id="more"
-          __tabs-item--icon-name="more"
-          __tabs-item--aria-label="Show more tabs"
-          __tabs-item--text="More"
-          __tabs-item--IsOverflow="true"
-        }}
-      {{/if}}
+        {{> __tabs-item}}
+      {{/wrapper}}
     {{/if}}
   {{/unless}}
 {{/tabs-list}}
--- a/src/patternfly/components/Tabs/__tabs-list.hbs
+++ b/src/patternfly/components/Tabs/__tabs-list.hbs
@@ -4,6 +4,8 @@ Tabs list content options
 ------------------------------------
 __tabs-list--HasIcons: bool
 __tabs-list--IsScrollable: bool
+__tabs-list--IsOverflow: bool
+__tabs-list--IsOverflowExpanded: bool
 __tabs-list--IsShort: bool
 __tabs-list--IsLong: bool
 __tabs-item--NoText: bool
@@ -108,6 +110,26 @@ __tabs-list--IsCloseDisabled: bool
         __tabs-item--aria-label="files"
         __tabs-item--text="Files"
         }}
+    {{/if}}
+    {{#if __tabs-list--IsOverflow}}
+      {{#if __tabs-list--IsOverflowExpanded}}
+        {{> __tabs-item
+          __tabs-item--id="more"
+          __tabs-item--icon-name="more"
+          __tabs-item--aria-label="Show more tabs"
+          __tabs-item--text="More"
+          __tabs-item--IsOverflow="true"
+          __tabs-item--IsOverflowExpanded="true"
+        }}
+      {{else}}
+        {{> __tabs-item
+          __tabs-item--id="more"
+          __tabs-item--icon-name="more"
+          __tabs-item--aria-label="Show more tabs"
+          __tabs-item--text="More"
+          __tabs-item--IsOverflow="true"
+        }}
+      {{/if}}
     {{/if}}
   {{/unless}}
 {{/tabs-list}}
--- a/src/patternfly/components/Tabs/examples/Tabs.md
+++ b/src/patternfly/components/Tabs/examples/Tabs.md
@@ -36,6 +36,20 @@ import './Tabs.css'
 {{/tabs}}
 ```
 
+### Horizontal overflow
+```hbs
+{{#> tabs tabs--id="horizontal-overflow-example"}}
+  {{> __tabs-list __tabs-list--NoScrollButtons="true" __tabs-list--IsOverflow="true"}}
+{{/tabs}}
+```
+
+### Horizontal overflow expanded
+```hbs
+{{#> tabs tabs--id="horizontal-overflow-expanded-example"}}
+  {{> __tabs-list __tabs-list--NoScrollButtons="true" __tabs-list--IsOverflow="true" __tabs-list--IsOverflowExpanded="true"}}
+{{/tabs}}
+```
+
 ### Accessibility
 
 | Attribute | Applied to | Outcome |
--- a/src/patternfly/components/Tabs/examples/Tabs.md
+++ b/src/patternfly/components/Tabs/examples/Tabs.md
@@ -36,6 +36,20 @@ import './Tabs.css'
 {{/tabs}}
 ```
 
+### Accessibility
+
+| Attribute | Applied to | Outcome |
+| -- | -- | -- |
+| `disabled` | `.pf-c-tabs__scroll-button` | Indicates that a scroll button is disabled, when at the first or last item of a list. **Required when disabled** |
+| `aria-hidden="true"` | `.pf-c-tabs__scroll-button` | Hides the icon from assistive technologies.**Required when not scrollable** |
+
+### Usage
+
+| Class | Applied to | Outcome |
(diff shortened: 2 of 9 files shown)


src/patternfly/components/Tabs/tabs-item-toggle-icon.hbs
```diff
@@ -0,0 +1,7 @@
+<span class="pf-c-tabs__item-toggle-icon">
+  {{#if tabs-item-toggle-icon--IsExpanded}}
```
mcoker, line 2: instead of swapping the icon source, we can use a single icon and handle the expanded state change in CSS (`transform: rotate()`). That allows us to transition the toggle rotation, too. You should be able to find an example of this in expandable section, table, data list, tree view, etc.

src/patternfly/components/Tabs/tabs-item-toggle-icon.hbs
```diff
@@ -0,0 +1,7 @@
...
+    <i class="fas fa-angle-down fa-sm" aria-hidden="true"></i>
+  {{else}}
+    <i class="fas fa-angle-right fa-sm" aria-hidden="true"></i>
```
mcoker, line 5: This will tie the icon size to markup, so if that ever changes, we can't just push a CSS change - devs will need to update the class (and however that would be done on the react-icons side, too) which can lead to inconsistencies and is a pain to update. For core we have to consider not breaking users who don't use react, so the less style we tie to markup the better.

Also if the size of the icon is tied to a class, and we change that class, that's technically breaking as someone may have applied a style to the icon using the class that changed (`.fa-sm`).

You can target `.pf-c-tabs__item-toggle-icon` in the CSS and set the font-size to sm.

```suggestion
    <i class="fas fa-angle-right fa-sm" aria-hidden="true"></i>
```

src/patternfly/components/Tabs/tabs-item-toggle-icon.hbs
mcoker: Also looks like the icon color changes on `:hover` (also `:focus` and `:active`), and expanded. The other components with rotation examples should show this, too, except for tree view - that's updated in an open PR.

src/patternfly/components/Tabs/tabs-item.hbs
```diff
@@ -9,7 +9,10 @@
     {{/if}}
   {{/if}}
   {{~#if tabs-item--IsAdd}} pf-m-add{{/if}}
+  {{~#if tabs-item--IsOverflow}} pf-m-overflow{{/if}}
+  {{~#if tabs-item--IsOverflowExpanded}} pf-m-expanded{{/if}}
   {{~#if tabs-item--modifier}} {{tabs-item--modifier}}{{/if}}"
+  {{~#if tabs-item--IsOverflowExpanded}} aria-expanded="true"{{/if}}
```
mcoker, line 15: typically `aria-expanded` goes on the thing that toggles the expansion, so I'd move this to the button.

src/patternfly/components/Tabs/tabs-item.hbs
```diff
@@ -9,7 +9,10 @@
     {{/if}}
   {{/if}}
   {{~#if tabs-item--IsAdd}} pf-m-add{{/if}}
+  {{~#if tabs-item--IsOverflow}} pf-m-overflow{{/if}}
+  {{~#if tabs-item--IsOverflowExpanded}} pf-m-expanded{{/if}}
```
mcoker, line 13: I think these classes are OK on the item, though you could argue that they belong on the tab button (either move them or duplicate them). This is the safest approach, since it allows us to either style the item or the tab button in the overflow/expanded state, but it will add an extra layer of specificity to any of the button rules if we have to target it like `.pf-c-tabs__item.pf-m-[overflow/expanded] .pf-c-tabs__link {}` instead of just being able to do `.pf-c-tabs__link.pf-m-[overflow/expanded]`. Though we can keep what you have and maintain specificity at "20" if the changes to an overflow/expanded link happen with vars on `.pf-c-tabs__item.pf-m-[overflow/expanded]`. So I'd say keep it the way it us but curious your thoughts, also cc @mattnolting

src/patternfly/components/Tabs/__tabs-list.hbs
```diff
@@ -109,6 +111,26 @@ __tabs-list--IsCloseDisabled: bool
...
+      {{else}}
+        {{> __tabs-item
+          __tabs-item--id="more"
+          __tabs-item--icon-name="more"
+          __tabs-item--aria-label="Show more tabs"
+          __tabs-item--text="More"
+          __tabs-item--IsOverflow="true"
+        }}
+      {{/if}}
+    {{/if}}
```
mcoker, line 133: fun fact! You can define a non-existent (or empty) `.hbs` partial (we use `{{#> wrapper}}`, which doesn't exist) and set parameters that will be inherited by its children, and since the `.hbs` file doesn't exist/do anything, it doesn't produce extra markup.

Also you can set `__tabs-item--IsOverflowExpanded` conditionally with the bool already set that indicates if its expanded (`__tabs-list--IsOverflowExpanded`)

So you could do something like 

```suggestion
    {{#if __tabs-list--IsOverflow}}
      {{#> wrapper
        __tabs-item--id="more"
        __tabs-item--id="more"
        __tabs-item--icon-name="more"
        __tabs-item--aria-label="Show more tabs"
        __tabs-item--text="More"
        __tabs-item--IsOverflow="true"
        __tabs-item--IsOverflowExpanded=__tabs-list--IsOverflowExpanded
        }}
          {{> __tabs-item}}
        {{/wrapper}}
    {{/if}}
```

src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -50,6 +50,20 @@ import './Tabs.css'
...
+{{/tabs}}
+```
+
+### Horizontal overflow expanded
+```hbs isBeta
+{{#> tabs tabs--id="horizontal-overflow-expanded-example"}}
+  {{> __tabs-list __tabs-list--NoScrollButtons="true" __tabs-list--IsOverflow="true" __tabs-list--IsOverflowExpanded="true"}}
+{{/tabs}}
+```
+
```
mcoker, line 66: Can you also include an example of what it looks like when the overflow tab is selected (the result of selecting an item from the menu). From the design:

Screen Shot 2022-05-27 at 11 38 45 AM

mcoker (changes requested):
Looks really good! Left some comments for review.

src/patternfly/components/Tabs/tabs-item-toggle-icon.hbs
wise-king-sullyman: I see the color change when expanded now, but I'm not seeing the color change on `:hover`, `:focus`, and `:active`. How can you tell it's changing in those states?

mcoker: @wise-king-sullyman you should see it when you hover with your mouse, focus with either the mouse or the keyboard, and `:active` is when you hold down your mouse on it.

Also dev tools have ways to trigger these states. This is chrome - you click the `:hov` button at the top/right of the styles pane. 

Screen Shot 2022-05-27 at 3 26 31 PM

src/patternfly/components/Tabs/tabs-item.hbs
wise-king-sullyman: To be completely honest I don't have the expertise to have a strong definitive opinion.

I can say that the reason I defaulted to just applying these classes to the tabs item itself is that from a React perspective with applying these classes I think I would *expect* that the classes would be applied directly to the list item.

But since disabled is applied to the button element within the tab currently I think that having them on the button instead would also make sense.

src/patternfly/components/Tabs/__tabs-list.hbs
wise-king-sullyman: That is fun!

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -531,6 +540,31 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
+  &.pf-m-expanded {
+    --pf-c-tabs__item-toggle-icon--Color: var(--pf-global--Color--100);
+    --pf-c-tabs__item-toggle-icon--Rotate: var(--pf-c-tabs__item--m-expanded__toggle-icon--Rotate);
+  }
+
+  &:hover {
+    --pf-c-tabs__item-toggle-icon--Color: var(--pf-c-tabs__item--hover--Color);
+  }
+
+  &:active,
```
mcoker, line 553: Fun fact with the order of `:hover`, `:focus`, and `:active`! The order should always be hover -> focus -> active, since that's the order of events. If you style `:hover` after either `:focus` or `:active`, when using a mouse, you would never see the `:focus` or `:active` styles since both of those states require the mouse be on (hovering) the element. And if you style `:focus` after `:active`, you would never see the `:active` since with a mouse or keyboard, an `:active` element also has `:focus`.

So we just need to swap the order of focus and active here.

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -531,6 +540,31 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
+    --pf-c-tabs__item-toggle-icon--Color: var(--pf-c-tabs__item--active--Color);
+  }
+
+  &:focus {
+    --pf-c-tabs__item-toggle-icon--Color: var(--pf-c-tabs__item--focus--Color);
+  }
+}
+
+.pf-c-tabs__item-toggle-icon {
+  font-size: small;
```
mcoker, line 564:
```suggestion
  font-size: var(--pf-c-tabs__item-toggle-icon--FontSize);
```

also you'll need to add `--pf-c-tabs__item-toggle-icon--FontSize: var(--pf-global--FontSize--sm)`

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -578,6 +612,7 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   position: relative;
   display: flex;
   flex: 1;
+  align-items: center;
```
mcoker, line 615: While this _should_ be ok... who knows if it could cause an issue with something someone is doing. Another way to do it would be to add `align-self: center` to the toggle icon element, assuming this declaration is to center that icon.

mcoker (changes requested):

src/patternfly/components/Tabs/__tabs-list.hbs
```diff
@@ -29,13 +32,18 @@ __tabs-list--IsCloseDisabled: bool
...
     __tabs-item--id="containers"
     __tabs-item--icon-name="fas fa-box"
     __tabs-item--aria-label="Containers"
     __tabs-item--text="Containers"
     }}
+    {{#if __tabs-list--IsOverflowSelected}}
+      {{> __tabs-item }}
+    {{else}}
+      {{> __tabs-item __tabs-item--current="true"}}
+    {{/if}}
```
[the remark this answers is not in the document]
mcoker: Yep, you can set it to something that doesn't exist - typically we use `reset`, so `__tabs-list--IsOverflowSelected=reset`. https://github.com/patternfly/patternfly/blame/main/src/patternfly/components/Form/examples/Form.md#L346

You can also declare `newcontext` on a partial and it will unset any inherited props. https://github.com/patternfly/patternfly/blame/main/src/patternfly/components/Menu/examples/Menu.md#L172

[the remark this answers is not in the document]
wise-king-sullyman: I think we might have had a bit of miscommunication about the reason I went with the implementation I had here, or there was some connection that I just wasn't making with your suggestions.

I did rewrite this in a better way in [44b78c5](https://github.com/patternfly/patternfly/pull/4876/commits/44b78c530e9952ec375ca568a9f99ef20441f4a0) with the aid of a simple helper I wrote to inverse a boolean though. Let me know what you think!

wise-king-sullyman:
@mmenestr something like this?

image

mmenestr:
Yes!! That's perfect, thank you :)

mcoker (changes requested):

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -658,6 +683,20 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
   }
+
+  &.pf-m-expanded {
+    --pf-c-tabs__link-toggle-icon--Color: var(--pf-c-tabs__link--m-expanded__toggle-icon--Color);
+    --pf-c-tabs__link-toggle-icon--Rotate: var(--pf-c-tabs__link--m-expanded__toggle-icon--Rotate);
+  }
+}
+
+.pf-c-tabs__link-toggle-icon {
+  align-self: center;
```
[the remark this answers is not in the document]
wise-king-sullyman: Yep that seems to work to me!

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -156,6 +156,15 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs__add--c-button--PaddingBottom: var(--pf-c-tabs__link--PaddingBottom);
   --pf-c-tabs__add--c-button--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
 
+  // Overflow menu toggle icon
+  --pf-c-tabs__item-toggle-icon--Color: var(--pf-global--disabled-color--100);
+  --pf-c-tabs__item-toggle-icon--Transition: .2s ease-in 0s;
+  --pf-c-tabs__item-toggle-icon--Rotate: 0;
+  --pf-c-tabs__item--m-expanded__toggle-icon--Rotate: 90deg;
+  --pf-c-tabs__item--hover--Color: var(--pf-global--link--Color--hover);
```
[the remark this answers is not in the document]
wise-king-sullyman: Ah yes sorry, hadn't updated this one because I didn't know if design was going to respond. Will address in just a second.

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -156,6 +156,18 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
 
+  // Overflow menu toggle icon
+  --pf-c-tabs__link-toggle-icon--Color: var(--pf-global--Color--200);
+  --pf-c-tabs__link-toggle-icon--Transition: .2s ease-in 0s;
+  --pf-c-tabs__link-toggle-icon--Rotate: 0;
+  --pf-c-tabs__link-toggle-icon--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-tabs__link-toggle-icon--PaddingTop: 3px;
+  --pf-c-tabs__link--m-expanded__toggle-icon--Color: var(--pf-global--Color--100);
+  --pf-c-tabs__link--m-expanded__toggle-icon--Rotate: 90deg;
+  --pf-c-tabs__item--hover--Color: var(---pf-global--Color--100);
```
mcoker, line 167: typo here - but is there a reason this hover state is on the `__item` and not the `__link`? Typically when we change a style when you're hovered/focused/active on something interactive - or otherwise interacting with the interactive thing, that the style change is visual feedback for) - we try to put the style on that thing if possible so as not to accidentally trigger a style change when clicking or whatever won't trigger a change on the interactive thing.

```suggestion
  --pf-c-tabs__item--hover--Color: var(--pf-global--Color--100);
```

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -156,6 +156,18 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
   --pf-c-tabs__add--c-button--PaddingBottom: var(--pf-c-tabs__link--PaddingBottom);
   --pf-c-tabs__add--c-button--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
 
+  // Overflow menu toggle icon
+  --pf-c-tabs__link-toggle-icon--Color: var(--pf-global--Color--200);
+  --pf-c-tabs__link-toggle-icon--Transition: .2s ease-in 0s;
+  --pf-c-tabs__link-toggle-icon--Rotate: 0;
+  --pf-c-tabs__link-toggle-icon--FontSize: var(--pf-global--FontSize--sm);
+  --pf-c-tabs__link-toggle-icon--PaddingTop: 3px;
```
mcoker, line 164: I looks like you can remove this

mcoker (changes requested):

src/patternfly/components/Tabs/tabs.scss
wise-king-sullyman: No reason other than me being unsure which things should be on the `__link` vs the `__item`. Will update.

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -688,6 +674,19 @@ $pf-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
...
+  }
+
+  &:focus {
+    --pf-c-tabs__link-toggle-icon--Color: var(--pf-c-tabs__link--focus--Color);
+  }
+
+  &:active,
+  &.pf-m-active {
+    --pf-c-tabs__link-toggle-icon--Color: var(--pf-c-tabs__link--active--Color);
+  }
```
mcoker, line 689: Could you move this up so it comes after the existing hover/focus/active state styles? https://github.com/patternfly/patternfly/blob/71925cdb08028f786a060befdfcdfb4585a7de32/src/patternfly/components/Tabs/tabs.scss#L622-L640

You could just add these styles to that block, though it will add them to the `.pf-m-action` items, too, and I don't think it needs to be there since you wouldn't want to trigger this color change if hovered over some other action in a tab item (like the close button).

Could you also rename the vars so it's

```suggestion
  &:hover {
    --pf-c-tabs__link-toggle-icon--Color: var(--pf-c-tabs__link--hover__toggle-icon--Color);
  }

  &:focus {
    --pf-c-tabs__link-toggle-icon--Color: var(--pf-c-tabs__link--focus__toggle-icon--Color);
  }

  &:active,
  &.pf-m-active {
    --pf-c-tabs__link-toggle-icon--Color: var(--pf-c-tabs__link--active__toggle-icon--Color);
  }
```

mcoker (changes requested):

mcoker (approved):
🥳

Issue #4877, opened 2022-05-26
Dark theme - copy files from new SCSS location


mcoker:
We moved a lot out of `themes/dark` to `base/themes/dark` and `sass-utilities/themes/dark` so that any stylesheet that has a dark theme version/overrides has `./themes/dark/` equivalent. For example: `base/_variables.scss` -> `base/themes/dark/_variables.scss`.

We just need to make sure those SCSS files are copied to the dist dir for consumers, our own build processes, etc.

PR #4864, opened 2022-05-19
fix(theme): import dark theme per component, add new global var

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4867
fixes https://github.com/patternfly/patternfly/issues/4873
fixes https://github.com/patternfly/patternfly/issues/4844

PR #4878, opened 2022-05-26
fix(theme): copy scss files to dist

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4877

for reference, here is the restructure that moved some of these files so they were no longer copied https://github.com/patternfly/patternfly/pull/4864/files


mattnolting (approved):
👍 LPTM

nicolethoen (approved):

PR #4879, opened 2022-05-26
fix(chip-group): fixed premature wrapping

### LINKED ISSUE #4874: Bug - Chip group - chips wrapping prematurely
https://www.patternfly.org/v4/components/chip-group/html/#simple-inline-chip-group-overflow

Looks like it was a result of https://github.com/patternfly/patternfly/pull/4836

mattnolting:
closes #4874 

Looks like the `max-width` on `.pf-c-chip-group__list` is the issue. 

**Before:** 

Screen Shot 2022-05-26 at 3 16 03 PM

**After:**

Screen Shot 2022-05-26 at 3 17 04 PM

**Very narrow:** The list still remains confined by the wrapper without the need for `max-width`. 

Screen Shot 2022-05-26 at 3 17 15 PM


mcoker (approved):
lgtm!!

PR #4884, opened 2022-06-01
fix(select): allow chip group to grow

### LINKED ISSUE #4883: Bug - Select - Multiple with Chip wraps
**Describe the problem**
Select with chips wraps making the component big

Select 2 or more items in the Multiple select

**Expected behavior**
The height of the component should not change and the "chips" should inline.

**Screenshots**
image

**What is your environment?**
 - OS: Fedora
 - Browser Chromium
 - Version 102.0.5005.61

srambach:
By adding `flex-grow:1` to a chip group in the select typeahead toggle, the chip group can grow so that it doesn't wrap as soon. However, there is a trade off because that does mean that the text input will be pushed further over now that the chip group container can grow. This is mainly noticeable on very wide screens, but in general the select should be within a form or something else that limits the select to a reasonable width, so this seems acceptable.
@tlabaj @mnolting @edewit @mcarrano @mmenestr 

Fixes #4883

### DIFF
--- a/src/patternfly/components/ChipGroup/chip-group.scss
+++ b/src/patternfly/components/ChipGroup/chip-group.scss
@@ -55,7 +55,6 @@
   flex-wrap: wrap;
   align-items: center;
   min-width: 0;
-  max-width: 100%;
 }
 
 .pf-c-chip-group__list-item {
--- a/src/patternfly/components/Select/select.scss
+++ b/src/patternfly/components/Select/select.scss
@@ -421,6 +421,7 @@
 
   // a chip group needs a bottom margin to make some space between the chip group and typeahead input box if it wraps
   .pf-c-chip-group {
+    flex-grow: 1;
     margin-top: var(--pf-c-select__toggle-wrapper--c-chip-group--MarginTop);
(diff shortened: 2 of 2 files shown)


mmenestr:
Is the amount of room/spacing between the end of the chip group, and the beginning of the placeholder text just a side effect of having this flex layout? And is there a way to reduce that spacing so that the chip group layout "limits" are closer to the last chip within the chip group? Let me know if this makes no sense :)

Screen Shot 2022-06-01 at 7 32 16 PM

srambach:
> Is the amount of room/spacing between the end of the chip group, and the beginning of the placeholder text just a side effect of having this flex layout? And is there a way to reduce that spacing so that the chip group layout "limits" are closer to the last chip within the chip group? Let me know if this makes no sense :)
> 
Yes - that's exactly it. By letting the chip group grow wider, now it's allowed to grow according to the flex rules of the container - so it's pushing the text input over because they share.
If the text input is given a higher `flex-grow` number than the chip group, then it will take up proportionately more of the space. You can see how this works in this video, where I keep increasing the value for the input.

mmenestr:
So if the chips in the chip group reaches the outer limit of the box it's in, is that when wrapping happens and the dropdown ends up taking 2 lines?

PR #4888, opened 2022-06-02
feat(a11y): add new screen reader element class

### LINKED ISSUE #4886: Accessibility - class similar to `pf-screen-reader`
Followup for https://github.com/patternfly/patternfly-react/pull/7474

https://github.com/patternfly/patternfly-react/pull/7474#pullrequestreview-992177438

image

We need a class to use that, instead of `position: fixed` and no width or height, is positioned `absolute` and full width and height to be used in things like a selectable `data-list` with a checkbox or radio.

wise-king-sullyman:
Closes #4886

### DIFF
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -22,14 +22,15 @@
 
 .pf-screen-reader {
   @include pf-u-screen-reader; // for use with assistive technologies
-}
 
-.pf-m-full-size {
-  @include pf-u-full-size;
-}
+  &.pf-m-full-size {
+    width: 100%;
+    height: 100%;
+  }
 
-.pf-m-absolute {
-  @include pf-u-absolute;
+  &.pf-m-absolute {
+    position: absolute;
+  }
 }
 
 .pf-m-tabular-nums {
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -24,8 +24,12 @@
   @include pf-u-screen-reader; // for use with assistive technologies
 }
 
-.pf-screen-reader-full-size {
-  @include pf-u-screen-reader-full-size;
+.pf-m-full-size {
+  @include pf-u-full-size;
+}
+
+.pf-m-absolute {
+  @include pf-u-absolute;
 }
 
 .pf-m-tabular-nums {
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -24,6 +24,10 @@
   @include pf-u-screen-reader; // for use with assistive technologies
 }
 
+.pf-screen-reader-full-size {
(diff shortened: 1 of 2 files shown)


mattnolting (approved):
LPTM 🎉

src/patternfly/sass-utilities/mixins.scss
```diff
@@ -126,6 +126,18 @@
   border: 0;
 }
 
+@mixin pf-u-screen-reader-full-size {
```
mcoker, line 129: LGTM, we could have this as a modifier, and potentially a fixed/absolute modifier. There may be a use case when you want something to be full size but be able to break out of its relatively positioned parent and relative to some other element instead - you could use `position: fixed` then, and add CSS that creates a stacking context to the parent you want it positioned relative to - https://codepen.io/mcoker/pen/XWZGapY

Then we could use `.pf-u-screen-reader.pf-m-full-size` (width/height 100%) and `.pf-u-screen-reader.pf-m-absolute` (position absolute) instead of unique classes for different variations.

Just a thought, I'm probably overthinking it 😂

wise-king-sullyman: That sounds like a great idea!

wise-king-sullyman:
@mcoker does [fd3eb36](https://github.com/patternfly/patternfly/pull/4888/commits/fd3eb36c88350f719495521e97e0a8b52614a3ad) match what you had in mind?

mcoker (approved):
LGTM. I think it's a bit random that we have these classes. I'm wondering if we should tie them to `.pf-screen-reader`, which would be more inline with how we use modifiers. It could go in the screen reader mixin, too, instead of defining new mixins.

WDYT @mattnolting?

src/patternfly/base/_common.scss
```diff
@@ -24,6 +24,14 @@
   @include pf-u-screen-reader; // for use with assistive technologies
 }
```
mattnolting, line 25:
```suggestion

  .pf-m-full-size {
    width: 100%;
    height: 100%;
  }
  
  &.pf-m-absolute {
    position: absolute;
  }
}
```

src/patternfly/base/_common.scss
```diff
@@ -24,6 +24,14 @@
...
 
+.pf-m-full-size {
+  @include pf-u-full-size;
+}
+
+.pf-m-absolute {
+  @include pf-u-absolute;
+}
+
```
mattnolting, line 34:
```suggestion
```

src/patternfly/base/_common.scss
mcoker: @mattnolting nice agree, thought did you mean to suggest `&.pf-m-full-size`?

mcoker (approved):
🥸👍

PR #4807, opened 2022-04-26
Remove phantom scrollbar in drilldown menu

srambach:
Fixes #4748 

I think this fixes it without breaking the scrollable menu but it is difficult to test within core. Here's a codepen with the fix, and the animation speed has been slowed so that you can see it. You can also remove a comment in index.tsx to test with the menu scrollable. https://codesandbox.io/s/drilldown-menu-phantom-scrollbar-fix-s674xf?file=/styles.css

This PR also removes an extra transition duration that was probably left in accidentally during development.

PR #4892, opened 2022-06-03
fix(Menu): fix scrollbar on flyout variant

### LINKED ISSUE #4871: Bug - Menu - flyout submenus cause horizontal scroll
**Describe the problem**
The flyout variant of the composable menu demo, as well as the flyout variant of the Navigation component, causes a horizontal scrollbar to appear when opening the flyout submenus. See video below.

https://user-images.githubusercontent.com/70952936/170264504-ca9d6d27-9b90-4307-9c18-1159f4f1ca59.mov

1. Go to [composable menu flyout example](https://www.patternfly.org/v4/demos/composable-menu/#composable-flyout) and [Navigation flyout example](https://www.patternfly.org/v4/components/navigation#flyout)
2. Hover over the items that have submenus (for Navigation, you need to go 2 levels deep) and notice that a scrollbar appears

**Expected behavior**
The submenus behave more like in the below video:

https://user-images.githubusercontent.com/70952936/170265050-24022798-82be-411a-9bb5-275b8089472d.mov

**Any other information?**
In the video directly above, the behavior was achieved by disabling the `position: relative` CSS rule in dev tools on selector `.pf-c-menu__list-item`. If this were the solution for this issue then this rule would have to be disabled only on this selector when part of a flyout menu (disabling this rule outright causes some slight shifting in the drilldown variant menu items I noticed).

thatblindgeye:
Closes #4871 

It looks like PR #4807 fix for the drilldown variant affected the flyout variant. I moved the fix introduced in that PR so that it only applies to the `pf-m-drilldown` class. Using the Codepen linked in this previous PR, I tweaked it a little to show how this fix *should* work (as I also don't believe I can test this solely in Core).

[Drilldown/flyout overflow fix codepen](https://codesandbox.io/s/drilldown-flyout-overflow-fix-s6k07v?file=/styles.css:50-127)

One point I could use clarification on is whether we would still need `overflow-x: hidden;` on line 245. If not, then I could instead replace that with `overflow: hidden` for the same result of this proposed solution it looks like.

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -241,8 +241,7 @@
 
     > .pf-c-menu__content {
       flex-grow: 1;
-      overflow: visible hidden;
-      overflow-x: hidden;
+      overflow: hidden;
       transition: var(--pf-c-menu--m-drilldown__content--Transition);
     }
 
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -196,8 +196,6 @@
   box-shadow: var(--pf-c-menu--BoxShadow);
 
   .pf-c-menu__content {
-    overflow: visible hidden;
-
     & & {
       overflow: visible;
     }
@@ -243,6 +241,7 @@
 
     > .pf-c-menu__content {
       flex-grow: 1;
+      overflow: visible hidden;
       overflow-x: hidden;
       transition: var(--pf-c-menu--m-drilldown__content--Transition);
     }


mattnolting (approved):
Nice! LPTM 🚀

thatblindgeye:
@mattnolting per @srambach comment above, I updated the PR to remove that extraneous `overflow-x` and just used `overflow: hidden`. I updated the codesandbox to reflex this and the behavior remains the same.

srambach (approved):
👍

mattnolting (approved):
LPTM 👍

PR #4895, opened 2022-06-09
feat(description-list): added display and card mods

### LINKED ISSUE #4699: DescriptionList - alternate font styles
Description list

**Describe the feature**
We'd like to use the semantics provided by the description list component to surface the contents shown in the following example, but using a font treatment similar to the styles shown below. The request is to provide modifier of the description list that provides styles similar to what's shown below. 

image

cc'ing @lboehling

mattnolting:
closes #4699

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -85,6 +85,10 @@
   --pf-c-card--m-display-lg--child--PaddingLeft: var(--pf-global--spacer--xl);
   --pf-c-card--m-display-lg--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
   --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--lg);
+  --pf-c-card--m-description-list-group--PaddingTop: var(--pf-c-card--first-child--PaddingTop);
+  --pf-c-card--m-description-list-group--PaddingRight: var(--pf-c-card--child--PaddingRight);
+  --pf-c-card--m-description-list-group--PaddingBottom: var(--pf-c-card--child--PaddingBottom);
+  --pf-c-card--m-description-list-group--PaddingLeft: var(--pf-c-card--child--PaddingLeft);
 
   // Flat
   --pf-c-card--m-flat--BorderWidth: var(--pf-global--BorderWidth--sm);
@@ -113,6 +117,10 @@
   flex-direction: column;
   background-color: var(--pf-c-card--BackgroundColor);
   box-shadow: var(--pf-c-card--BoxShadow);
+
+  &.pf-m-description-list-group {
+    padding: var(--pf-c-card--m-description-list-group--PaddingTop) var(--pf-c-card--m-description-list-group--PaddingRight) var(--pf-c-card--m-description-list-group--PaddingBottom) var(--pf-c-card--m-description-list-group--PaddingLeft);
+  }
 
   &.pf-m-hoverable {
     &:hover {
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -85,10 +85,6 @@
   --pf-c-card--m-display-lg--child--PaddingLeft: var(--pf-global--spacer--xl);
   --pf-c-card--m-display-lg--c-divider--child--PaddingTop: var(--pf-global--spacer--xl);
   --pf-c-card--m-display-lg__title--not--last-child--PaddingBottom: var(--pf-global--spacer--lg);
-  --pf-c-card--m-description-list-group--PaddingTop: var(--pf-c-card--first-child--PaddingTop);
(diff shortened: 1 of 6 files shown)


mcarrano (commented):
Thanks for creating these examples @mattnolting . 

@jgiardino @mceledonia can you take a look and see if this is what you expect? Since we didn't create an explicit design for this, it's hard for me to say what these examples should look like, but seems like the implementation is pretty flexible Thanks.

src/patternfly/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -117,6 +117,33 @@ Column fill will modify the default placement of description list groups to fill
...
 ```
 
+## Display and card variants
+
+### Display variant
+```hbs
+{{> description-list__example description-list--modifier="pf-m-display-lg pf-m-2-col-on-lg" description-list--title="Card DL"}}
+```
+
+### Group as card
```
srambach, line 127: What do you mean by "group"?

src/patternfly/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -117,6 +117,33 @@ Column fill will modify the default placement of description list groups to fill
...
+```hbs
+{{> description-list__example description-list--modifier="pf-m-2-col-on-lg" description-list-group--IsCard="true" description-list--title="Card DL"}}
+```
+
+### Display and card variant
+```hbs
+{{> description-list__example description-list-group--IsCard="true" description-list--modifier="pf-m-display-lg pf-m-2-col-on-lg" description-list--title="Card DL"}}
+```
+
+### Display and card variant, 3 col on lg
```
srambach, line 137: I'd spell out this heading text

srambach (commented):
I was going to say the use of a card is not a variant, but a demo... but since there's custom styling in that case I guess it isn't. At a minimum I think we need something in the usage section about making the card variant.
Also, if I saw this on a design, I'd likely first think it's some kind of card variant, so maybe there needs to be something over on the cards page to point you here.

src/patternfly/components/DescriptionList/examples/DescriptionList.md
mattnolting: Our description list `<dd>` and `<dt>` examples are wrapped by `.pf-c-description-list__group`. Do you think I should change the verbiage?

srambach (approved):
Docs are a lot clearer, thanks! LGTM 👍

src/patternfly/components/DescriptionList/description-list.scss
```diff
@@ -107,17 +114,29 @@ $pf-c-description-list--breakpoint-map: build-breakpoint-map("base", "sm", "md",
       // stylelint-enable
     }
   }
+
+  &.pf-m-display-lg {
+    .pf-c-description-list__description {
+      font-size: var(--pf-c-description-list--m-display-lg__description--FontSize);
+    }
```
mcoker, line 121: This works but you could add 

```
.pf-c-description-list__description {
  font-size: var(--pf-c-description-list__description--FontSize, inherit);
}
```

Then update this block to 

```suggestion
  --pf-c-description-list__description--FontSize: var(--pf-c-description-list--m-display-lg__description--FontSize);
```

src/patternfly/components/Card/card.scss
```diff
@@ -110,6 +114,10 @@
   background-color: var(--pf-c-card--BackgroundColor);
   box-shadow: var(--pf-c-card--BoxShadow);
 
+  &.pf-m-description-list-group {
```
mcoker, line 117: Hmm... this is fine, but I wonder if this should be either:

1. If this is specific to the description list, move these styles into the description list. The `--pf-c-card...Padding` vars should work fine since you have `.pf-c-card` on the element the modifier applies to.
  * Going forward, if the DL card styles grow, it might seem odd to bundle a bunch of extra description list specific styles when someone is just using the card.
1. A generic thing for a card without children, with padding. Maybe `.pf-m-padding`? Seems redundant since that's what `<div class="pf-c-card"><div class="pf-c-card__body">...</div></div>` does now.
1. Styled in the DL as `.pf-c-description-list > .pf-c-card`. That seems inline with how we style other things like that - like the toolbar in the masthead. Given the markup constraints to get this to work with a DL, seems like selector should be fine?

mcoker (commented):
Nice! I love the approach.

src/patternfly/components/Card/examples/Card.md
```diff
@@ -503,6 +503,10 @@ import './Card.css'
 {{/card}}
 ```
 
+## Description list group wrapper card variant
+
+Cards can be used as [description list group wrappers](../description-list#description-list-group-wrapper-as-card). Using cards in this way applies the card body padding directly to the card and enables the use of card modifiers within description list groups.
+
```
mcoker, line 509:
```suggestion
```

mcoker (changes requested):
Just one nit on the card docs

mcoker (approved):
nice!!

PR #4901, opened 2022-06-09
Feat progress stepper 4776

### LINKED ISSUE #4776: Progress stepper - responsive behavior for horizontal progress stepper
What is the existing component, if any?
- Existing component: https://www.patternfly.org/v4/components/progress-stepper

**Describe the feature**
- When the screen size is constrained the horizontal progress stepper would change to the vertical orientation at a specific breakpoint. The same as what happens for other components such as table and data list...
https://www.patternfly.org/v4/components/progress-stepper#vertical-with-descriptions

mattnolting:
closes #4776


srambach:
I'm seeing a horizontal scroll in some cases when it flips.

https://user-images.githubusercontent.com/19825616/174153749-6f25dae9-ece0-420b-8382-2c70029e336a.mp4

srambach:
Normal (not compact) vertical orientation needs some design as well. This is how it looks today if specified center and vertical, but center align didn't automatically go vertical so it wouldn't happen by default as it does now.
image

mattnolting:
> Normal (not compact) vertical orientation needs some design as well. This is how it looks today if specified center and vertical, but center align didn't automatically go vertical so it wouldn't happen by default as it does now. image

Great point @srambach! Both compact vertical and centered vertical will need design support. Adding to the description.

mcarrano:
@mattnolting Thanks for calling this out. I see how the center aligned versions are a problem. Maybe we should just expect that to switch to left aligned when it goes vertical. Is that possible? For the compact version I think I'm missing what the problem is. Looks OK to me when I preview it.

mcoker:
In the PR preview, this is what I'm seeing that differs from the screenshots above:

#### Compact vertical (looks ok)
Screen Shot 2022-06-16 at 4 43 36 PM

#### Compact vertical center
Screen Shot 2022-06-16 at 4 45 00 PM

mcoker (changes requested):
This looks fantastic to me. A couple of nits about the docs and the design issues you called out. As far as I can tell, it's just the center aligned variation that needs attention. Sounds like @mcarrano is suggesting we limit `.pf-m-center` to `.pf-m-horizontal`, which works for me. Just to whip something up quickly, we could also look at something like:

Screen Shot 2022-06-16 at 6 04 11 PM
Screen Shot 2022-06-16 at 6 02 55 PM

mcarrano (approved):
I think this looks fine now with the latest changes. Thanks @mattnolting !

src/patternfly/components/ProgressStepper/progress-stepper.scss
```diff
@@ -1,18 +1,134 @@
...
+  --pf-c-progress-stepper__step-connector--before--BorderRightColor: var(--pf-c-progress-stepper--m-vertical__step-connector--before--BorderRightColor);
+  --pf-c-progress-stepper__step-connector--before--BorderBottomWidth: var(--pf-c-progress-stepper--m-vertical__step-connector--before--BorderBottomWidth);
+  --pf-c-progress-stepper__step-connector--before--BorderBottomColor: var(--pf-c-progress-stepper--m-vertical__step-connector--before--BorderBottomColor);
+  --pf-c-progress-stepper__step-connector--before--Transform: var(--pf-c-progress-stepper--m-vertical__step-connector--before--Transform);
+  --pf-c-progress-stepper__step-main--MarginTop: var(--pf-c-progress-stepper--m-vertical__step-main--MarginTop);
+  --pf-c-progress-stepper__step-main--MarginRight: var(--pf-c-progress-stepper--m-vertical__step-main--MarginRight);
+  --pf-c-progress-stepper__step-main--MarginBottom: var(--pf-c-progress-stepper--m-vertical__step-main--MarginBottom);
+  --pf-c-progress-stepper__step-main--MarginLeft: var(--pf-c-progress-stepper--m-vertical__step-main--MarginLeft);
```
mcoker, line 19: small nit - we should remove this on the center/vertical variation

src/patternfly/components/ProgressStepper/progress-stepper.scss
```diff
@@ -264,6 +358,16 @@
 // Step main
 .pf-c-progress-stepper__step-main {
   margin: var(--pf-c-progress-stepper__step-main--MarginTop) var(--pf-c-progress-stepper__step-main--MarginRight) var(--pf-c-progress-stepper__step-main--MarginBottom) var(--pf-c-progress-stepper__step-main--MarginLeft);
+
+  // Draw a new border for vertical alignment using step main
+  .pf-c-progress-stepper__step:not(:last-of-type) > &::before {
+    position: absolute;
+    top: calc(100% + var(--pf-c-progress-stepper__step-main--MarginTop));
+    left: calc(50% - var(--pf-c-progress-stepper__step-connector--before--BorderRightWidth));
```
mcoker, line 366: This is a bit off-center. Part of it is because of the left margin on step-main, but then we only want to move it back by half of its width.

```suggestion
    left: calc(50% - calc(var(--pf-c-progress-stepper__step-connector--before--BorderRightWidth) / 2))
```

mcoker (changes requested):

mcoker (approved):
LPTM!!

PR #4902, opened 2022-06-13
feat(card): add styling to cards when their hidden input is focused

### LINKED ISSUE #4835: Card - hidden inputs used by screen readers should appear to focus the associated card
What is the existing component, if any?

Card

**Describe the feature**
A clear and concise description of the new feature. What is the expected behavior?

In patternfly/patternfly-react#7144 hidden inputs are being added to selectable card components to improve a11y in a manner similar to how our [switch component](https://www.patternfly.org/v4/components/switch) works. It would be beneficial for some indicative styling to be applied to the card that is associated with a hidden input when that input receives focus.

What the switch icon looks like when focused on a hidden input (the third switch from the top is currently focused):
image

**Any other information?**

wise-king-sullyman:
Closes #4835

I didn't add a new demo for this but can if needed.

With this PR a card will look like the middle card in the below mockup when the hidden input associated with it is focused.

image

An alternative approach to this that I considered was to just apply the hover styles when the input is focused, but since the hover style is different for non-legacy vs legacy I think this would need to happen in the next breaking change release if it's decided as the best path forward.


src/patternfly/components/Card/card.scss
```diff
@@ -380,6 +384,10 @@
   font-size: var(--pf-c-card__footer--FontSize);
 }
 
+input:focus + .pf-c-card {
```
mcoker, line 387: This selector is likely OK, though it is a bit generic and has the potential to cause an issue. Can we give it a class? 

Also toggling a `border` on a normal element like that will cause the content to shift a bit. Curious if you've tested it and seen if it shifts by the size of the border width? Typically we'll use a pseudo element for stuff like that, looks like we could use `::after` here.

Also I'm not completely familiar with this feature. Is this input is for screen readers only - not keyboard users?

mcoker (changes requested):

src/patternfly/components/Card/card.scss
wise-king-sullyman: Correct screen readers only, keyboard users that aren't using a screen reader won't be able to reach the hidden input.

And good points, I'll give it a class and use `::after` as I do see it shift a bit when the border is applied now that you mention it.

mcoker (approved):
LGTM! We should add documentation for this element at some point. Want to add that now, or later after we've sorted out that everything works like we want on the react side?

wise-king-sullyman:
@mcoker how would you want to see it documented? An example added to core?

mcoker:
@wise-king-sullyman oh sorry, documenting the element/class in the card docs table - https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Card/examples/Card.md#documentation

An example would be good at some point, too.

src/patternfly/components/Card/card-sr-input.hbs
```diff
@@ -0,0 +1,8 @@
+<input
+  type="checkbox"
+  tabindex="-1"
+  class="pf-c-card__sr-input pf-screen-reader
+    {{#if card-sr-input--modifier}}
+      {{card-sr-input--modifier}}
+    {{/if}}"
+/>
```
mcoker, line 8: nit - this would align more with how a standard hbs element template is setup. You could put the modifier stuff on it's own lines if you want, just use the `~` to remove the leading whitespace, like `{{~#if card-sr-input--modifier}}`

also in core, html5 doesn't require the `/` on self closing elements

```suggestion
<input class="pf-c-card__sr-input pf-screen-reader{{#if card-sr-input--modifier}} {{card-sr-input--modifier}}{{/if}}"
  type="checkbox"
  tabindex="-1"
  {{#if card-sr-input--attribute}}
    {{{card-sr-input--attribute}}}
  {{/if}}>
```

src/patternfly/components/Card/examples/Card.md
```diff
@@ -270,6 +270,22 @@ import './Card.css'
 {{/card}}
 ```
 
+### Selectable with a hidden input for improved screen reader accessibility
+```hbs
+{{> card-sr-input }}
```
mcoker, line 275: nit 

```suggestion
{{> card-sr-input}}
```

src/patternfly/components/Card/examples/Card.md
```diff
@@ -509,6 +525,7 @@ A card is a generic rectangular container that can be used to build other compon
 | `.pf-c-card__actions` | `<div>` | Creates an actions element to be used in the card header. |
 | `.pf-c-card__header-main` | `<div>` | Creates a wrapper element to be used in the card header when using an image, logo, or text. |
 | `.pf-c-card__expandable-content` | `<div>` | Creates the expandable card's expandable content. |
+| `.pf-c-card__sr-input` | `<input>` | Creates an input which, when focused, makes a following `.pf-c-card` appear focused. |
```
mcoker, line 528: would this suggestion be accurate? I think "focused" would imply normal focus.

```suggestion
| `.pf-c-card__sr-input` | `<input>` | Creates an input which, when has screen reader focus, makes a following `.pf-c-card` appear focused. |
```

wise-king-sullyman: Hmm, I think this is accurate as there's nothing about the class that makes the element inherently unfocusable.

mcoker (approved):
Nice!!!

PR #4905, opened 2022-06-14
Adds height breakpoints for sticky top and bottom modifiers

### LINKED ISSUE #3660: sticky header/footer: breakpoint improvements for increased font size and mobile
In the PatternFly-Cockpit sync up call, we talked about the sticky header & footer component (#3516).

I brought up the how this can be both good and bad for mobile and @jgiardino mentioned this also affects all browser sizes when fonts are increased.

Browsers don't support component-based sizing in CSS right now, so we'd have to try to guess at what best to do.

PatternFly could, by default, have breakpoints where sticky is enabled for larger horizontal _and_ vertical breakpoints (with font size taken into account for the vertical breakpoints). And then have an override option for someone who knows what they're doing to enable sticky for mobile too (in the cases where the author of the app _knows_ the content won't be more than 1 line tall, for example).

(We're already doing something like this in Cockpit, IIRC; but it's in our own custom components. It would be great to have this in PF itself, by default.)

srambach:
This adds a set of vertical breakpoints that are used for sticky top and bottom modifiers on page components. [Demo for a sticky breadcrumb](https://patternfly-pr-4905.surge.sh/components/page/html-demos#sticky-breadcrumb-on-medium)
Fixes #3660 

As part of this, the functions and mixins for creating utilities were updated to allow creation of vertical breakpoint utilities.

There was previous discussion about whether to try to create two-dimensional media queries. After some thought working through it with this [codepen](https://codepen.io/srambach/pen/dydKQaV?editors=0100), I feel that it's a lot of overhead for not that much advantage. There are some cases that can't really be handled with separate height and width media queries, but I honestly think that would be a sign of a too-complicated design. [This article](https://www.kooslooijesteijn.net/blog/multidimensional-responsive-design-5-tips) helped me to believe that as well.

### DIFF
--- a/src/patternfly/demos/Page/examples/Page.md
+++ b/src/patternfly/demos/Page/examples/Page.md
@@ -30,6 +30,15 @@ wrapperTag: div
 }}
 ```
 
+### Sticky breadcrumb on medium
+```hbs isFullscreen
+{{> page-template
+      page-template--id="page-demo-sticky-top-breadcrumb"
+      page-template-gallery--IsLongGallery="true"
+      page-template-breadcrumb--modifier="pf-m-sticky-top-on-md-height"
+}}
+```
+
 ### Sticky section group
 ```hbs isFullscreen
 {{> page-template
--- a/src/patternfly/demos/Page/examples/Page.md
+++ b/src/patternfly/demos/Page/examples/Page.md
@@ -31,7 +31,7 @@ wrapperTag: div
 ```
 
 ### Sticky breadcrumb on medium
-```hbs isFullscreen
+```hbs isFullscreen isBeta
 {{> page-template
       page-template--id="page-demo-sticky-top-breadcrumb"
       page-template-gallery--IsLongGallery="true"
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -97,7 +97,7 @@
 // Example individual utility:
 // @include pf-utility-builder(flex-fill flex "1 1 auto", $pf-global--breakpoint-list);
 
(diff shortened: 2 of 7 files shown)


src/patternfly/sass-utilities/mixins.scss
```diff
@@ -53,13 +90,14 @@
...
 // responsive, including all breakpoints    @include pf-utility-builder($sass-map, $pf-global--breakpoint-list)
+// responsive height breakpoints            @include pf-utility-builder($sass-map, $pf-global--height-breakpoint-list, 'height')
 
 // ## Passing individual utilities values
 // ===============================================================================================
 // Example individual utility:
 // @include pf-utility-builder(flex-fill flex "1 1 auto", $pf-global--breakpoint-list);
 
-@mixin pf-utility-builder($props, $breakpoints: null) {
+@mixin pf-utility-builder($props, $breakpoints: null, $direction:"width") {
```
mcoker, line 100:
```suggestion
@mixin pf-utility-builder($props, $breakpoints: null, $direction: "width") {
```

mcoker (changes requested):

mattnolting (approved):
LGTM!!

src/patternfly/demos/Page/examples/Page.md
```diff
@@ -30,6 +30,15 @@ wrapperTag: div
 }}
 ```
 
+### Sticky breadcrumb on medium
+```hbs isFullscreen
```
mcoker, line 34:
```suggestion
```hbs isFullscreen isBeta
```

mcoker (changes requested):

mcoker (approved):

Issue #4889, opened 2022-06-02
Bug - Card component - Cards presented on the same background color in dark theme mode loose their edge contrast


sg00dwin:
**Describe the problem**
Cards presented on the same background color, which happens within `<PageSection variant={PageSectionVariants.light}>`, have the same background `--pf-global--BackgroundColor--100`, which is fine on light theme, but in dark theme mode they loose their edge contrast because the box shadow is barely perceptible.

OpenShift console presents cards within `<PageSection variant={PageSectionVariants.light}>` in several locations.

**How do you reproduce the problem?**
https://codesandbox.io/s/whie-worker-hzvqws-hzvqws

**Expected behavior**
It would help to have increase edge contrast between the card and its page section in dark theme.

_Option 1_ 
Set card background to a slightly different color than the current `--pf-global--BackgroundColor--100`
eg. `:where(.pf-theme-dark) .pf-c-card {--pf-c-card--BackgroundColor: var(--pf-global--BackgroundColor--dark-300)`

_Option 2_
Add a 1px border to `.pf-c-card` 
eg. `:where(.pf-theme-dark) .pf-c-card {border: 1px solid var(--pf-global--palette--black-1000)}`

_Option 3_
Increase the card boxshadow size to `--pf-global--BoxShadow--md`

**Screenshots**
Provided APIs are presented using the card component
<img width="1252" alt="os-dark-card" src="https://user-images.githubusercontent.com/1874151/171700795-b57d12d4-e089-4427-a739-102b20372318.png">

<img width="1397" alt="os-dark-card2" src="https://user-images.githubusercontent.com/1874151/171700810-79cc587c-c0e9-4058-ba2d-31b13f89acba.png">

**What is your product and what release date are you targeting?**
OpenShift 4.11

**Any other information?**

nicolethoen:
@mceledonia and @mcoker could you take a look at this?

PR #4913, opened 2022-06-17
fix(card): force dk theme dk section card bg color

srambach:
Previous PR #4889 changed the box shadow of cards on dark background. However, design would like all cards on a dark section to have the same background. 

@mceledonia In some cases, the pf-m-non-selectable-raised card is a different color. Is this the desired result? They are a different color in the normal (light) theme.

### DIFF
--- a/src/patternfly/components/Card/themes/dark/card.scss
+++ b/src/patternfly/components/Card/themes/dark/card.scss
@@ -1,7 +1,8 @@
 @import "../../../../sass-utilities/themes/dark/all";
 
 @mixin pf-theme-dark-card() {
-  .pf-c-card {
+  .pf-c-card,
+  .pf-c-card.pf-m-non-selectable-raised {
     --pf-c-card--BoxShadow: var(--pf-global--BoxShadow--md);
 
     // Hoverable/selectable raised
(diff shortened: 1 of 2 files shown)


srambach:
Screenshots:

src/patternfly/components/Card/themes/dark/card.scss
```diff
@@ -10,4 +11,8 @@
     --pf-c-card--m-selectable-raised--focus--BoxShadow: var(--pf-global--BoxShadow--lg);
     --pf-c-card--m-selectable-raised--active--BoxShadow: var(--pf-global--BoxShadow--lg);
   }
+
+  .pf-c-page__main-section[class*="pf-m-dark-"] .pf-c-card {
```
mcoker, line 15: This has the potential of styling cards we don't intend on styling, like if they're used in some other component and not directly against the background color of the main section. That said, I hadn't made any sort of considerations for the `.pf-m-dark-` page section backgrounds with the dark theme work, so I wonder how often they're used and if this is likely to even be a problem.

mcoker (approved):

srambach:
New screenshots with different section variants.
image

Issue #4931, opened 2022-06-28
stylelint errors with currentColor case


mcoker:
The linter is failing here due to the mixed case

```
src/patternfly/components/Button/button.scss
 619:28  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case

src/patternfly/components/JumpLinks/jump-links.scss
 69:42  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case

src/patternfly/components/MenuToggle/menu-toggle.scss
 362:35  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case

src/patternfly/components/Table/table.scss
  180:44  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case
  181:51  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case
  182:51  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case
  183:52  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case
 1177:58  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case

src/patternfly/components/Tabs/tabs.scss
 129:36  ✖  Expected "currentColor" to be "currentcolor"  value-keyword-case
```

PR #4932, opened 2022-06-28
fix(global): changed currentColor to currentcolor

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4931


srambach:
Ironically, the build fails lint because it thinks it should be currentColor 🙄 Do we need an update to one thing or another?

PR #4934, opened 2022-06-29
build(docs): update docs framework

jschuler:
Co-locates docs related content under /patternfly-docs
Fixes CSS ordering in the docs build


mcoker (approved):

src/patternfly/base/_fonts.scss
```diff
@@ -11,7 +11,7 @@
     url("#{$pf-global--font-path}/RedHatDisplay/RedHatDisplay-Regular.woff") format("woff");
   font-style: normal;
   font-weight: 300;
-  text-rendering: optimizeLegibility;
```
wise-king-sullyman, line 14: I see that @mcoker had a different PR up for the currentColor -> currentcolor issue, but the MDN docs have this value in camel case and I don't see an issue or anything related to this change. I'm assuming this is related to the same stylelint errors mentioned in #4931?

wise-king-sullyman (commented):
I have one question about the optimizeLegibility change, other than that this all looks great!

src/patternfly/base/_fonts.scss
mcoker: @wise-king-sullyman good callout. Yep, with this PR `optimizeLegibility` fails, too

```
src/patternfly/base/_fonts.scss
 14:19  ✖  Expected "optimizeLegibility" to be "optimizelegibility"  value-keyword-case
```

It's safe to update as it isn't case sensitive, and then we're using lower case everywhere as far as I'm aware.

wise-king-sullyman (approved):

PR #4965, opened 2022-07-12
refactor(toolbar): updated toolbar to support resize observer

### LINKED ISSUE #3348: Add GAP to toolbar!

### LINKED ISSUE #4947: Toolbar - convert to composable & resize observer supported
Refactor toolbar to be composable and use resize observer
* Bulk Selector size changes based on selection + Search input + Filters input + Primary Button (hides at breakpoint) + actions menu icon + any other icons + pagination + the page primary navigation can change the toolbar width. Too many variables to rely just on breakpoints and have it work like expected. Need resize observers and smartly collapse toolbar items based on actual toolbar width.
* Set priority, based on available space
* Ex: collapse three filters into one, grouped select
* Separate filter group collapsing functionality. Make it incremental / can add things to expandable content section at various stages/size availability etc
* Filters and drop down actions allow sub menu items (beta component). Need to also be able to disable items when shown and have a tooltip why they are disabled.

mattnolting:
closes #4947


nicolethoen (commented):
I'm finally reviewing this!

I think the expandable content will be placed in a popper now when expanded. So I don't know if you need to have multiple hidden or empty expandable contents to represent having multiple toggle groups in a toolbar. I don't know how you'd usually demonstrate these things though now...

Issue #4970, opened 2022-05-14
Bug - [Notification Drawer] - [Sticky page sections do not stay put]


jowenn:
**Describe the problem**
My basic layout (that works as expected) is:
```
 <StrictMode>
     {environmentConfig.env_warning_banner?(<Banner variant="warning">{environmentConfig.env_warning_banner}</Banner>):undefined}
     <Suspense fallback="loading">
        <Translation> <!- from react-i18next -->
            <BrowserRouter>
                <Page header={header}
                    sidebar={<PageSidebar isNavOpen={isNavOpen} nav={<PageNav />}/>}
                     >
                       <PageSection sticky="top">TEST</PageSection>
        					<PageSection>
                               very long content

  header is a <Masthead> and PageNav is essentially <><ContextSelector/><Nav/>
```

  With this structure it works as expected and the sticky="top" header stays on top and the section with the long content scrolls, that works almost as expected (if the very long content is a table, the scrolling is visually strange, but that's another topic).

  Now I have tried to add a notification drawer, so the new structure is:
```
   <StrictMode>
     {environmentConfig.env_warning_banner?(<Banner variant="warning">{environmentConfig.env_warning_banner}</Banner>):undefined}
     <Suspense fallback="loading">
        <Translation> <!- from react-i18next -->
            <BrowserRouter>
                <Page header={header}
                    sidebar={<PageSidebar isNavOpen={isNavOpen} nav={<PageNav />} notificationDrawer={notificationDrawer}
                          isNotificationDrawerExpanded={this.state.isNotificationDrawerOpen} />}
                     >
                       <PageSection sticky="top">TEST</PageSection>
        					<PageSection>
                               very long content

notificationDrawer is:
const notificationDrawer=(
            <NotificationDrawer>
                <NotificationDrawerHeader onClose={this.onNotificationDrawerClose}>&nbsp;</NotificationDrawerHeader>
                <NotificationDrawerBody></NotificationDrawerBody>
            </NotificationDrawer>
        );
```
 After that change, the sticky header does not stay sticky anymore, and the whole cntent of the page (in this case both sections are scrolling)

My workaround is to add this class to my own stylesheet
```
.pf-c-drawer__body {
    display:flex;
    flex-direction:column;
    flex-basis:0%;
    flex-grow:1;
    flex-shrink:1;
}
```
[...]

PR #4971, opened 2022-07-13
fix(drawer): fix sticky page sections when in drawer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4970


srambach (approved):
LGTM

PR #4980, opened 2022-07-18
fix(menu): removed hidden-visible usage

### LINKED ISSUE #4870: Bug - Menu - drilldown variant has hidden submenus accessible via VO
**Describe the problem**
For the drilldown variant of composable menus (as well as the Navigation drilldown variant), traversing the menu via VO + right arrow key causes a submenu to be focused despite it meant to be hidden.

1. Go to [Composable Menu drildown example](https://www.patternfly.org/v4/demos/composable-menu/#composable-drilldown-menu)
2. Turn on Voice Over
3. Open the menu and traverse it via the VO key + right arrow key, and notice how the first menu item "Start rollout" has its submenu focused despite it meant to be hidden

**Expected behavior**
Traversing via VoiceOver using VO + arrow keys does not cause what should be a hidden menu to receive focus, and instead the next menu item should receive focus.

**Any other information?**
Looking at the computed styles of these submenus, it looks like the visibility property has two sources trying to apply:

computed properties for drilldown

Going to the second source and disabling the `--pf-hidden-visible--Visibility` variable from the below screenshot resolves this issue:

styles for drilldown

Replacing `visibility: hidden` with `display: none` on the `.pf-c-menu.pf-m-drilldown .pf-c-menu__list-item:not(.pf-m-current-path) .pf-c-menu` selector also resolved this issue, but it doesn't work as well on the Navigation drilldown variant (clicking a link within the menu then trying to open the drilldown menu item causes the entire menu to disappear).

This could also possibly be resolved via React, but based on the above info it seems like it might be more a Core issue.

mattnolting:
closes #4870 

After testing, the removal of `pf-m-hidden` and `pf-m-visible` modifiers has no affect on the `drilldown` variant for `.pf-c-menu__list-item`. Neither class is present in the DOM or used in any of the interaction states. It's safe to remove.


mcoker (commented):
TIL about this https://codepen.io/mcoker/pen/BarWpjy! I didn't know that was possible.

Traditionally we've required both `display: none` and `visibility: hidden` on an element to hide it across all the things we cared about - this will remove `visibility` from the hidden/visible classes on list items if users are hiding/showing them for other reasons. Is that OK?

Does `display: none` on the nested/hidden menus not work? Changing it in the react component like Eric mentioned seems to work - the parent menu height is still computed correctly.

mattnolting:
@mcoker after some more testing, the issue appears to be with `&:not(.pf-m-current-path) .pf-c-menu` not also being to to `display: none`. Try this solution out @thatblindgeye and let me know what you think.

mattnolting:
@mcoker @thatblindgeye There's also an issue with `pf-c-divider` being counted in child DOM elements, so VO announces the second `li` after a divider as "{title} item 3 of 6" instead of "{title} 2 of 5". I think we either need to `aria-hidden="true"` this, as we did with the nested table column header, or hide otherwise.

mcoker (approved):
🥳

thatblindgeye (approved):
Tested on both the React menu drilldown and nav drilldown examples and it looks to work great! 🎉  I did initially notice at first on the nav drilldown when trying to apply the styles to the appropriate selector that the second level menu (after clicking the "Item 1" item) wasn't displaying, but after trying again I wasn't able to replicate the issue.

PR #4981, opened 2022-07-18
fix(table): removed truncate from sticky hrs

### LINKED ISSUE #4882: Bug - Table - Sticky column table example has all `th` cells set to truncate
**Describe the problem**
Sticky column table example has all `th` cells set to truncate. 

Inspect code https://www.patternfly.org/v4/components/table/html#sticky-column

**Expected behavior**
Ideally, the first two `th` cells would be set to `pf-m-fit-content` and the rest of the cells would have no modifier at all. 

No

mattnolting:
closes #4882


mcoker (approved):
LPTM!

PR #4983, opened 2022-07-18
feat(number-input): added support for status icon

### LINKED ISSUE #4833: Status controls for number input component
### Requirement
The RHOSAK design team has come across a use case that requires a warning state for the number input field (in this case, the field label is Partitions). 

A user can create multiple Partitions using this form. When they reach the Partition limit, we want to display a warning state but not truly enforce the limit (i.e. allow the user to proceed with the number that is over the limit and not disable the field).

### Design samples
Expected_Form_Sample

Here's a PoC created by Michael Coker: [Code-Sandbox-Mockup](https://codesandbox.io/s/hungry-nightingale-y19h9x?file=/index.html)

srambach:
This increases the overall width of the number input so that if a status icon is put onto the form control, the number will not be cut off, and the width will not change, causing the right side control to jump. 
Since this increases the size, it uses a modifier `.pf-m-has-icon` to indicate that a status icon may be contained within and the extra space should be added. This modifier can be removed in a breaking change.
https://patternfly-pr-4983.surge.sh/components/number-input

Fixes #4833

### DIFF
--- a/src/patternfly/components/NumberInput/examples/NumberInput.md
+++ b/src/patternfly/components/NumberInput/examples/NumberInput.md
@@ -130,7 +130,7 @@ cssPrefix: pf-c-number-input
 
 ### With status
 ```hbs
-{{#> number-input number-input--id="number-input-default" number-input--modifier="pf-m-status"}}
+{{#> number-input number-input--id="number-input-status" number-input--modifier="pf-m-status"}}
   {{#> input-group}}
     {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Minus"'}}
       {{#> number-input-icon}}
@@ -147,7 +147,7 @@ cssPrefix: pf-c-number-input
   {{/input-group}}
 {{/number-input}}
 <br/><br/>
-{{#> number-input number-input--id="number-input-default" number-input--modifier="pf-m-status"}}
+{{#> number-input number-input--id="number-input-status-warning" number-input--modifier="pf-m-status"}}
     {{#> input-group}}
     {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Minus"'}}
       {{#> number-input-icon}}


src/patternfly/components/NumberInput/number-input.scss
```diff
@@ -7,12 +7,24 @@
...
+      calc(
+        var(--pf-c-number-input--c-form-control--width-base) + var(--pf-c-number-input--c-form-control--width-chars) * 1ch
+      ) + var(--pf-c-number-input--c-form-control--width-icon)
+    );
 
   display: inline-flex;
   align-items: center;
 
+  // This modifier can be removed in a breaking change and the extra space created can always be included.
+  &.pf-m-has-icon {
```
mcoker, line 24: I would probably either call this icon or status. I looked and we don't have any other classes with `pf-m-[has/is]`.

mcoker (changes requested):
Other than the base `--width-icon` needing a unit and a small comment about the class name, LGTM!

mcarrano (approved):
This looks good to me. Thanks @srambach !

src/patternfly/components/NumberInput/examples/NumberInput.md
```diff
@@ -128,6 +128,43 @@ cssPrefix: pf-c-number-input
 {{/number-input}}
 ```
 
+### With status
+```hbs
+{{#> number-input number-input--id="number-input-default" number-input--modifier="pf-m-status"}}
```
mcoker, line 133:
```suggestion
{{#> number-input number-input--id="number-input-status" number-input--modifier="pf-m-status"}}
```

src/patternfly/components/NumberInput/examples/NumberInput.md
```diff
@@ -128,6 +128,43 @@ cssPrefix: pf-c-number-input
...
+    {{/form-control}}
+    {{#> button button--modifier="pf-m-control" button--attribute='aria-label="Plus"'}}
+      {{#> number-input-icon}}
+        <i class="fas fa-plus" aria-hidden="true"></i>
+      {{/number-input-icon}}
+    {{/button}}
+  {{/input-group}}
+{{/number-input}}
+<br/><br/>
+{{#> number-input number-input--id="number-input-default" number-input--modifier="pf-m-status"}}
```
mcoker, line 150:
```suggestion
{{#> number-input number-input--id="number-input-status-warning" number-input--modifier="pf-m-status"}}
```

mcoker (changes requested):
looks awesome!! Just a couple of nits with the `--id` param, which is creating `name` attributes on the text inputs with the same values between examples.

mcoker (approved):
🥳

Issue #4935, opened 2022-06-29
Form inputs - Provide styles for read-only inputs that are visually similar to description list


jgiardino:
**Is this a new component or an extension of an existing one?**
Form and form inputs

**Describe the feature**
We'd like to have a different style for read-only inputs that are visually similar to how property values display in a [description list](https://www.patternfly.org/v4/components/description-list) instead of being similar to how disabled inputs display. 

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
For Kafka topics, we worked with @lboehling and @srambach to create [custom css](https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui/blob/481894163351cddd41487ef7f0c75aa71c69e6b4/src/components/TextWithLabelPopover/TextWithLabelPopover.css) to create the UI shown in the following example. For this use case, most properties are read-only but a few are editable. We created custom css for read-only inputs for the following reasons:
- The original design used similar styles as the description list. But when the user is in the context of a form, providing the values in a read-only input is more accessible (e.g. the 3rd property is read-only, but is still focusable).
- There was concern that inputs that appeared visually disabled would imply there was some action a user could take to enable them. In this scenario, the properties are always read-only.
 
![image](https://user-images.githubusercontent.com/21063328/176474033-e206c8bc-dd48-4669-ab97-fefaa42cf3cf.png)

**Any other information?**
There are other projects that would like to reuse this variation. One of the projects would use this style in a view-mode of a page where the user can click Edit to switch to edit-mode. We still want the visual presentation that is shown above for the Kafka topic, but there are a few additional use cases for this project:
- The form is generated dynamically from a schema. Being able to apply a modifier at the Form level would be preferred. But also note for the use case described above, we need also need to be able to apply this modifier at the input level. 
- In our original implementation of these styles, we only had to deal with text inputs. But the other project that has forms generated dynamically could have other input types. These other types include the PF Select component and boolean/checkbox values.

mmenestr:
@nicolethoen talked to other designers around this and we all agreed that the solution proposed here is better than what we have today for a read-only field. So the proposal is to **replace** the current read-only styling, in favor of the more "detail list" styling. Thinking it should still have a border and classic input styling but it not be visible if that makes sense.

I just want to bring this up at design share in 2 weeks for a sanity check before we move forward, but after that it should be good to be added to a milestone.

nicolethoen:
@mcarrano could we add this to an upcoming milestone?

[...]

PR #4984, opened 2022-07-18
fix(form-control): style changes to readonly inputs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4935


mcoker:
This was just to test, we decided to create a plain variation that removes the border, background, and padding and basically presents the input as text.

PR #4986, opened 2022-07-19
fix(dropdown): added menu position combinator

### LINKED ISSUE #4861: Bug - Select/Dropdown/ContextSelector - position property affects popper flipping
**Describe the problem**
The Select, Dropdown, and Context Selector components all use Popper under the hood when they aren't appended inline, which by default should allow the menus of these components to flip vertically when reaching a boundary. Due to the menu's having `position: absolute` in their styling, this does not work as intended (the flipping occurs at the incorrect time and the position is off when it does).

The selectors that have this property are:

- Context selector component: `.pf-c-context-selector__menu`
- Dropdown component: `.pf-c-dropdown__menu` (this also had a `min-width` property added in the React fix linked below)
- Select component: `.pf-c-select__menu`

See React PR https://github.com/patternfly/patternfly-react/pull/7434 for a current React fix

Go the the following component pages and apply `menuAppendTo="parent"` to the returned component in an example:

[Context Selector](https://www.patternfly.org/v4/components/context-selector/)
[Dropdown](https://www.patternfly.org/v4/components/dropdown)
[Select](https://www.patternfly.org/v4/components/select)

Video of the behavior in action:

https://user-images.githubusercontent.com/70952936/169051335-b39ef48e-c44c-46e7-8299-d8821add202b.mov

**Expected behavior**
See the video below of how the composable menu demos behave:

https://user-images.githubusercontent.com/70952936/169051721-c69caa51-fd86-40d8-88c6-f3bd4fca4399.mov

N/A

**Screenshots**
N/A

**What is your environment?**
 - OS: macOS Monterey 12.3.1
 - Browser Chrome
 - Version 101.0.4951.64

N/A

**Any other information?**
N/A

mattnolting:
closes #4861 

Possible approach. When popper is used, a duplicate component is added to the DOM. By using an adjacent sibling combinator, we can target non-popper menus.

### DIFF
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -299,8 +299,9 @@
   background-clip: padding-box;
   box-shadow: var(--pf-c-context-selector__menu--BoxShadow);
 
-  &:only-child {
-    position: initial;
+  &:only-child,
+  &.pf-m-static {
+    position: static;
   }
 }
 
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -302,6 +302,9 @@
   &:only-child,
   &.pf-m-static {
     position: static;
+    top: auto;
+    z-index: auto;
+    min-width: auto;
   }
 }
 
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -304,7 +304,7 @@
     position: static;
     top: auto;
     z-index: auto;
-    min-width: auto;
+    min-width: min-content;
   }
 }
 
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -291,6 +291,7 @@
 .pf-c-context-selector__menu {
   @include pf-t-light;
 
+  position: absolute;
   top: var(--pf-c-context-selector__menu--Top);
   z-index: var(--pf-c-context-selector__menu--ZIndex);
   min-width: 100%;
@@ -298,8 +299,8 @@
   background-clip: padding-box;
   box-shadow: var(--pf-c-context-selector__menu--BoxShadow);
 
-  .pf-c-context-selector__toggle + & {
-    position: absolute;
+  &:only-child {
+    position: initial;
   }
 }
 
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -299,7 +299,6 @@
   background-clip: padding-box;
   box-shadow: var(--pf-c-context-selector__menu--BoxShadow);
 
-  &:only-child,
   &.pf-m-static {
     position: static;
     top: auto;
--- a/src/patternfly/components/ContextSelector/context-selector.scss
+++ b/src/patternfly/components/ContextSelector/context-selector.scss
@@ -65,7 +65,6 @@
   --pf-c-context-selector__menu-list--PaddingTop: var(--pf-c-context-selector__menu--PaddingTop); // reference spacer directly at breaking change
   --pf-c-context-selector__menu-list--PaddingBottom: var(--pf-global--spacer--sm);
 
-
   // Menu item
   --pf-c-context-selector__menu-list-item--Color: var(--pf-global--Color--dark-100);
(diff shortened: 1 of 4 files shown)


mcoker (commented):
Hmm.... this works but I wonder if changing this style to be based off of the markup order is either going to cause problems with the way someone's using it today or how we might want to use it in the future. I'm curious if this should be a variation instead, or some way to opt-in to the change. An opt-in way could be to do `&:only-child { position: revert; ... }`

WDYT about a modifier that assumes the menu is positioned some other way, and removes the properties needed for absolute positioning? `top`, potentially `z-index`, `width/min-width`. We have another issue to handle the way popper applies `z-index` to menus when it positions them versus the component positioning them.

@thatblindgeye the issue that requires the `min-width` fix is because the parent element that popper positions has a fixed `width` that appears to match the width of the toggle. For dropdowns, that's incorrect - the menu should be at least as wide as the toggle, but can grow to be bigger if content allows. More accurately, popper could assign `min-width: {{fixed value of toggle}}` and get the correct behavior. I'm not sure if that `width` is used for anything, like calculating when the menu is out-of-bounds or something?

Here you can see how the popper's fixed width doesn't match the content inside:
Screen Shot 2022-07-19 at 6 50 32 PM

mcoker:
I created a doc from our meeting about some options on how to approach this. https://docs.google.com/document/d/1wnEQEfzVbhawk5ZKo-2O1fD-E63P5AJBL7etffXdTxA/edit?usp=sharing

mattnolting:
@mcoker Great, thanks for the document. Based on the current implementation of popper (duplicate dropdowns), it seems that we can use both the  `&:only-child` pseudo selector and a `.pf-m-static` modifier. The menu width issue is a popper implementation bug, and setting `--pf-c-dropdown__menu--MinWidth` to `--pf-c-dropdown__toggle--MinWidth` results in the dropdown menu `min-width: 0;`, which undoes the menu being at least as wide as the toggle. Seems like we need to update the popper inline styling.

mattnolting:
@mcoker @thatblindgeye @wise-king-sullyman updated

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -631,6 +631,14 @@ $pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "
   background: var(--pf-c-dropdown__menu--BackgroundColor);
   background-clip: padding-box;
   box-shadow: var(--pf-c-dropdown__menu--BoxShadow);
+
+  &:only-child,
+  &.pf-m-static {
+    position: static;
+    top: auto;
+    z-index: auto;
+    min-width: auto;
```
thatblindgeye, line 640: Tested these styles out in React locally by replacing the current inline style that uses `position: revert; min-width: min-content` on `pf-c-dropdown__menu`, and they do work when `popperMatchesTriggerWidth={false}` is passed into Popper. Doing that may result in the following, though:

Dropdown toggle wider than content

Changing `min-width` to `min-content` from these styles (without needing `popperMatchesTriggerWidth={false}`) looks to more closely match how the "inline" dropdown looks. Would `auto` be better to use since this is a modifier class that could be used outside of just these dropdown-like components?

thatblindgeye (approved):
🎉 Looks good!

wise-king-sullyman (approved):
🚀

src/patternfly/components/ContextSelector/context-selector.scss
```diff
@@ -299,6 +298,14 @@
   background-color: var(--pf-c-context-selector__menu--BackgroundColor);
   background-clip: padding-box;
   box-shadow: var(--pf-c-context-selector__menu--BoxShadow);
+
+  &:only-child,
```
mcoker, line 302: do we still need this if we're going to add the modifier in react? I was thinking the class would let us get around needing to use this markup dependent selector.

src/patternfly/components/Dropdown/dropdown.scss
mcoker: Just wanted to mention that another solution we discussed was if popper could use `min-width` instead of `width` with `poppermatchesTriggerWidth` since we don't always want the width to match (I think that would be updated here - https://github.com/patternfly/patternfly-react/blob/8bcbea20cbe53d6d5ba472a0a94b40860ff5e15c/packages/react-core/src/helpers/Popper/Popper.tsx#L252-L267), but I don't see a downside to use `min-width: min-content` on the static menu, either. Since dropdown item text doesn't wrap, this will fix other conditions where there is overflow in the menu from long menu items.

mcoker (commented):
Just one comment! Not sure if there is a reason to keep the nth-child selector or if we can get rid of it.

mcoker (approved):
🤩

Issue #4987, opened 2022-07-19
Form control - incorrect disabled text color


mcoker:
Looks like the disabled text color var isn't applying properly.

PR #4988, opened 2022-07-19
fix(form-control): fixed disabled text color

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4987

fixed the root issue by passing the form control base text color - also updated the disabled state to set `color` so it isn't overridden by other states currently setting `color`.


mcarrano (commented):
Thanks for updating @mcoker . I feel like this can even be a shade lighter, but not sure if that will pass a11y standards. @mceledonia what do you think? I'll defer to your opinion.

mcoker:
@mcarrano one consideration might be that if we update this text color, we would also want to consider updating the other things that have form control styling like menu toggles and control buttons.

mcarrano:
Good point. So maybe we leave this alone then.

srambach (approved):
LGTM 👍

mattnolting (approved):
LGTM 🎉

Issue #4781, opened 2022-04-04
Consider aligning the popover close with the end of content


chloenayon:
We've noticed that the close button on a popover does not align with the end of the popover content.
This is consistent with the examples of popovers in the [documentation](https://www.patternfly.org/v4/components/popover/):

![Screenshot from 2022-04-04 18-06-25](https://user-images.githubusercontent.com/11710246/161585769-1553c222-e899-42b3-982c-f6e2628d84fa.png)

How about aligning the button with the end of the content?

![popover-align](https://user-images.githubusercontent.com/11710246/161585458-8b0cb30d-e8fc-444d-a806-05cf23323889.png)

mcarrano:
We will get this onto our roadmap. Currently scheduling into Q3.

PR #4989, opened 2022-07-20
fix(popover): align close button with the end of content

blfetche:
@mcoker 

applied following:
`--pf-c-popover--c-button--Right: calc(var(--pf-c-popover__content--PaddingRight) - var(--pf-global--spacer--md));`

Initially tried using 
`var(--pf-c-button--PaddingRIght) `but did not work, in its place I used `var(--pf-global--spacer--md)`

image


blfetche:
> Did the FontWeight changes come in by accident or are they intended? Otherwise, LGTM 👍 Also, if you add "Fixes #4781" in the description, it will tie them together automatically.

Hmm, I was positive I had committed my prior changes to another branch before committing to this one. Thank you.

Issue #4990, opened 2022-07-21
Selectable tree view bugs


mcoker:
* Checkbox and selectable toggle icons should be black when expanded.
* Examples are not apply the current class to nodes that are selectable.
* If you apply `.pf-m-current` to a selectable node and hover the toggle, it turns blue. It should turn black instead.
* All checkbox nodes have `.pf-m-selectable` - only the expandable nodes need that

PR #4991, opened 2022-07-21
fix(treeview): selectable/expandable node bugs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4990

### DIFF
--- a/src/patternfly/components/Check/check-input.hbs
+++ b/src/patternfly/components/Check/check-input.hbs
@@ -1,5 +1,8 @@
 <input class="pf-c-check__input{{#if check-input--modifier}} {{check-input--modifier}}{{/if}}"
   type="checkbox"
+  {{#if check-input--IsChecked}}
+    checked
+  {{/if}}
   {{#if check-input--attribute}}
     {{{check-input--attribute}}}
   {{/if}}
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -701,7 +701,7 @@ beta: true
             {{/tree-view-list-item}}
             {{#> tree-view-list-item}}
               {{#> tree-view-content}}
-                {{> tree-view-node tree-view-node--text="Current"}}
+                {{> tree-view-node tree-view-node--text="Options"}}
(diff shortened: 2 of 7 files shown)


mmenestr:
This looks good if it's just making the expanded nodes black, and the collapsed nodes grey?

mmenestr (approved):

mcoker:
@mmenestr yep, the only other visual change is

> If you apply `.pf-m-current` to a selectable node and hover the toggle, it turns blue. It should turn black instead.

You can see that works in this example - https://patternfly-pr-4991.surge.sh/components/tree-view#with-selectable-expandable-nodes

You can see the bug it fixes in this example in @kmcfaul's PR - https://patternfly-react-pr-7714.surge.sh/components/tree-view#with-separate-selection-and-expansion
* Select an expandable row by clicking on the text, then hover over the toggle button and note the arrow turns blue instead of black.

mcoker:
Updated the expandable/selectable non-checkbox node to be a `div` with a `button` toggle element inside per @kmcfaul's feedback. Also worked on the aria-labels a bit. It should read "select [node text]" now for the non-checkbox expandable/selectable node and checkbox treeview's checkbox inputs. And it updates the selectable/expandable and checkbox toggles' labels to read "toggle [node text]"

thatblindgeye (commented):
Looks good! I am noticing an issue when navigating via VO in the checkbox example.

The items that only have a checkbox (no expandable toggle) seem to have inaccurate labeling. In Application 1, Settings gets announced as "Select loader settings, level 3", for example. The same applies for Application 1 > Loader, and Application 2 > Settings (the second one).

The checkbox's `aria-labelledby` are pulling in the `text-` ID from the selectable/expandable example rather than their own example.

mcoker:
@thatblindgeye ah hah! Sorry for the delay, I was trying to wrap my head around some non-repetitive logic to set a flag to enable the text `id` when appropriate and ended up using our `concat` helper to basically serve as an `or` operator. Hopefully that works!

thatblindgeye (approved):
🎉 Looks good! I created https://github.com/patternfly/patternfly/issues/5010 as a followup to the issue we discussed

PR #4995, opened 2022-07-21
fix(text-input-group): fix chip/filter wrapping

### LINKED ISSUE #4885: Bug - Text input group - too many chips is unreadable
**Describe the problem**
The chips in the chip group passed to the text input group are given the property `flex-wrap: nowrap;` which causes them all to squish into one line and become impossible to read.
Screen Shot 2022-06-01 at 11 50 42 AM

**Expected behavior**
Chips wrapping and causing the text input group to grow in height is probably preferable to the chips being completely unreadable. Users can still prevent this by using the `show more/less` chip group feature.

srambach:
This removes the rule preventing the chip group from wrapping within a text input group, and changes the min-width from 0 to fit-content so that the chips don't get too small to read. This isn't foolproof at extremely small widths but mostly alleviates the problem.
Also fixed a duplicate id in the example causing an axe error.

Fixes #4885

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -20,9 +20,9 @@
   --pf-c-text-input-group__main--ColumnGap: var(--pf-global--spacer--sm);
 
   // Chip group
-  --pf-c-text-input-group__c-chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
-  --pf-c-text-input-group__c-chip-group__main--PaddingRight: var(--pf-global--spacer--xs);
-  --pf-c-text-input-group__c-chip-group__main--PaddingBottom: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group--c-chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group--c-chip-group__main--PaddingRight: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group--c-chip-group__main--PaddingBottom: var(--pf-global--spacer--xs);
 
   // Text input
   --pf-c-text-input-group__text-input--PaddingTop: var(--pf-global--spacer--form-element);
@@ -98,9 +98,9 @@
   }
 
   .pf-c-chip-group__main {
-    padding-top: var(--pf-c-text-input-group__c-chip-group__main--PaddingTop);
-    padding-right: var(--pf-c-text-input-group__c-chip-group__main--PaddingRight);
-    padding-bottom: var(--pf-c-text-input-group__c-chip-group__main--PaddingBottom);
+    padding-top: var(--pf-c-text-input-group--c-chip-group__main--PaddingTop);
+    padding-right: var(--pf-c-text-input-group--c-chip-group__main--PaddingRight);
+    padding-bottom: var(--pf-c-text-input-group--c-chip-group__main--PaddingBottom);
   }
 }
 
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -167,7 +167,6 @@
   position: relative;
   width: 100%;
   min-width: var(--pf-c-text-input-group__text-input--MinWidth);
-  height: fit-content;
   padding: var(--pf-c-text-input-group__text-input--PaddingTop) var(--pf-c-text-input-group__text-input--PaddingRight) var(--pf-c-text-input-group__text-input--PaddingBottom) var(--pf-c-text-input-group__text-input--PaddingLeft);
   border: 0;
 
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -16,11 +16,13 @@
   // Main
   --pf-c-text-input-group__main--first-child--not--text-input--MarginLeft: var(--pf-global--spacer--sm);
   --pf-c-text-input-group__main--m-icon__text-input--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-text-input-group__main--RowGap: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__main--ColumnGap: var(--pf-global--spacer--sm);
 
   // Chip group
-  --pf-c-text-input-group__chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
-  --pf-c-text-input-group__chip-group__main--PaddingRight: var(--pf-global--spacer--xs);
-  --pf-c-text-input-group__chip-group__main--PaddingBottom: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__c-chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__c-chip-group__main--PaddingRight: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__c-chip-group__main--PaddingBottom: var(--pf-global--spacer--xs);
 
   // Text input
   --pf-c-text-input-group__text-input--PaddingTop: var(--pf-global--spacer--form-element);
@@ -84,7 +86,7 @@
   display: flex;
   flex: 1;
   flex-wrap: wrap;
-  gap: var(--pf-global--spacer--xs) var(--pf-global--spacer--sm);
+  gap: var(--pf-c-text-input-group__main--RowGap) var(--pf-c-text-input-group__main--ColumnGap);
   min-width: 0;
 
   &.pf-m-icon {
@@ -96,9 +98,9 @@
   }
 
   .pf-c-chip-group__main {
-    padding-top: var(--pf-c-text-input-group__chip-group__main--PaddingTop);
-    padding-right: var(--pf-c-text-input-group__chip-group__main--PaddingRight);
-    padding-bottom: var(--pf-c-text-input-group__chip-group__main--PaddingBottom);
+    padding-top: var(--pf-c-text-input-group__c-chip-group__main--PaddingTop);
+    padding-right: var(--pf-c-text-input-group__c-chip-group__main--PaddingRight);
+    padding-bottom: var(--pf-c-text-input-group__c-chip-group__main--PaddingBottom);
   }
 }
 
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -121,7 +121,6 @@
   grid-template-columns: 1fr;
   grid-template-areas: "text-input";
   flex: 1;
-  align-items: center;
 
   &::before,
   &::after {
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -120,6 +120,7 @@
   grid-template-columns: 1fr;
   grid-template-areas: "text-input";
   flex: 1;
+  align-items: center;
 
   &::before,
   &::after {
@@ -165,6 +166,7 @@
   position: relative;
   width: 100%;
   min-width: var(--pf-c-text-input-group__text-input--MinWidth);
+  height: fit-content;
   padding: var(--pf-c-text-input-group__text-input--PaddingTop) var(--pf-c-text-input-group__text-input--PaddingRight) var(--pf-c-text-input-group__text-input--PaddingBottom) var(--pf-c-text-input-group__text-input--PaddingLeft);
   border: 0;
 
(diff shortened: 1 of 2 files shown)


mcarrano (commented):
@srambach Looks like this works. I don't know if we'd want a prop to enable the wrapping behavior or leave this as the default. @nicolethoen what do you think?

nicolethoen:
I like this as the default IMO. I'm not sure what the alternative would be.
The react implementation, by default will likely hide most of the chips in a 'x more' chip.

nicolethoen (approved):

mcoker:
I like this as the default, too, though we did decide to have the horizontal filter overflow scroll as a single line as opposed to wrap like this (and typeahead select) when we created the component. Just want to verify that's OK. Here's the original issue where we added the text input group component, which funny enough shows a different way of showing overflow chips (in a popover) - https://github.com/patternfly/patternfly/issues/4374

I believe this is the design issue that we used to create the initial component https://github.com/patternfly/patternfly-design/issues/1088

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -103,9 +103,10 @@
     }
   }
 
-  .pf-c-chip-group,
-  .pf-c-chip-group__list {
-    flex-wrap: nowrap;
+  .pf-c-chip-group li,
```
mcoker, line 106: Is there an `li` in chip group that's outside of `__list`? Curious why you need this and the line below it.

Also could you target `.pf-c-chip-group__list-item` instead of `li`?

And now that we're allowing these to wrap, I'm not sure why we need to update `min-width` - I wouldn't think they would shrink if they can wrap instead. Disabling this `min-width: fit-content` style in dev tools doesn't break anything as far as I can tell, though I'm probably missing something 🤓

mcoker (changes requested):
Nice! I left a comment, and also looks like we need some kind of padding or something because the chips are touching the top/bottom edge of the parent when the chips wrap. Also looks like there is an overflow property we might be able to get rid of.

Jul-21-2022 15-15-02

mcarrano (approved):
Looks good from a design perspective. Thanks @srambach !

mcoker (changes requested):
This looks great! The only thing I see now is that the typeahead input doesn't wrap and gets stretched.

Screen Shot 2022-07-25 at 5 06 20 PM
 
I'm assuming we want to mimic the select typeahead - @mcarrano @mmenestr can you confirm? The text input basically just has a minimum width that it will shrink down to, and once met, it wraps to a new line and occupies that entire line.

Screen Shot 2022-07-25 at 5 03 38 PM

Screen Shot 2022-07-25 at 5 03 27 PM

mcarrano:
Agree with @mmenestr

srambach:
As @mcoker notes, the original design called for horizontal scroll. If we just change the `min-width` of a chip, then they don't excessively shrink down, and the scroll kicks in. Do you want to go that direction?

https://user-images.githubusercontent.com/19825616/181033437-44258aa5-35e6-459b-86f7-4fb0f0549df6.mp4

nicolethoen:
My initial reaction is that - as is - there are very few visual clues that the list of chips is scrollable and that there are some chips out of view

srambach:
> My initial reaction is that - as is - there are very few visual clues that the list of chips is scrollable and that there are some chips out of view

I agree 100% that the scroll is not obvious. Letting it wrap is also not great because it could shift the entire layout. 
This example is also not showing a chip to collapse the chip group back down (which is in the react example). Just throwing out an idea - what if instead, that overflow chip stayed sticky, and maybe it somehow has something indicate there's more (an icon, a gradient, an ellipse, ?) Sort of this idea? https://css-tricks.com/text-fade-read-more/

mcarrano:
I think that's an interesting idea @srambach . We do something similar with horizontal navigation menus. But how does the user interact to scroll it? Is it a keyboard accessible solution?

nicolethoen:
I still think that since we have a 'view more' feature of these chips, that it's okay if things wrap or take up more space when they are expanded. Is that considered a breaking change?

srambach:
An idea with the "more" chip stuck to the end of the chip group and using a gradient to indicate there is something to scroll. https://codepen.io/srambach/pen/eYMMoYb?editors=1100

mcoker:
@mcarrano ^ - @srambach's codepen is what I was going to create last time we discussed this issue.

mcarrano:
@srambach @mcoker I like this a lot. @mmenestr @mceledonia perhaps also take a look.

mmenestr:
I like it too! I just wonder if there's a need to add a gradient in this case when there's already a "8/View more" button. I would expect there to be a gradient when the view more button has been clicked, to give a visual indication that you can scroll, but not before!

srambach:
In the codepen example, to make the overflow chip sticky, I've selected the last chip. However, if there isn't an overflow chip, it would just select the last one to make sticky. Unfortunately, the `.pf-m-overflow` class is on the button, but it's the chip that needs to be targeted to make sticky. 
So, a few options:
A. rely on always having the overflow chip if there's going to be scroll (I'm not sure if this is feasible because the react component creates it when needed?)
B. add a class to the chip indicating it should stick
C. use `:has(.pf-m-overflow)` to target the chip (:has is not well supported yet)

C is great but doesn't yet have support. And I'm not sure whether A works. WDYT @mcoker @nicolethoen ?

mcoker:
> B. add a class to the chip indicating it should stick

@srambach do you mean add a class to the chip group item (`.pf-c-chip-group__list-item`)? If so, I'm in favor of that. seems like a useful class if we need the chip group to be aware it's an overflow chip so it could be spaced differently or something like that.

If so, we should probably create a follow up and add that to the label group (and add a class for the "add" chip)?

srambach:
Based on a discussion with @mcarrano @mmenestr @nicolethoen and @mcoker, we decided to:

Implement a scroll variant of a chip group, which does not wrap but remains on one line with a gradient indicating when there is something scrolled off to one side or the other.
- This variant could be used in a text input group when it must remain on one line (e.g. in a toolbar) **However, the preferred solution is to use the toolbar with filters component.**
- Follow-on React task needed to handle adding an "n more" chip that automatically adjusts how many chips are hidden based on the available space. Clicking on that overflow chip would remove the overflow chip and then all chips would be available via the horizontal scroll. There is no reverse (show less) action.

When the default (wrapping) version of the chip group is used in a text input group, allow the text input to wrap below the chip group. 
- As a future task, implement an "add chip" feature to put the text input within the chip group to wrap with the chips.

Sound correct?

mcoker:
> Implement a scroll variant of a chip group, which does not wrap but remains on one line with a gradient indicating when there is something scrolled off to one side or the other.

👍

> When the default (wrapping) version of the chip group is used in a text input group, allow the text input to wrap below the chip group.

IIRC, this would be a follow up for later, too. Text input group only supports horizontal scrolling for now.

srambach:
I've tried to get scroll shadows to work with the chip group, but because of the background color of chips, I've been unable to ([codepen showing why](https://codepen.io/srambach/pen/eYMobBb)). 
I propose for now that we let the[ text input wrap below the chip group](https://patternfly-pr-4995.surge.sh/components/text-input-group/#filters), which is allowed to wrap and the chips now won't need to get too narrow to read. If a better solution is found in the future, we can go back and revisit.
WDYT @mcarrano @mmenestr ?

mcoker:
@srambach would it still be an option to add the shadows via a class on the react side, depending on which direction has overflow?

mmenestr:
@srambach I think that's fine!!

srambach:
After further investigation with @mcoker and @mattnolting, I got the shadows above the chips, but the shadow is misplaced when a close button is added. Using :has() this problem can be solved without causing a breaking change in the html, but that is not widely implemented yet.
This codepen shows some of the issues. Viewing in FF (no support for :has) you can see the problem with the shadow location when there is a close button.
If we want to create a modifier that indicates there is a close button, that could be used to adjust instead of using :has() but honestly I don't feel all this is a stable enough solution to warrant adding it.
I'd like to suggest we accept [the solution you can see now in the preview](https://patternfly-pr-4995.surge.sh/components/text-input-group), wrapping the chip group and wrapping the text input below the chip group.
@mcarrano @mmenestr

mmenestr:
@srambach I'd be fine with that!

mcarrano:
Me too.

mcoker (commented):
There's a bit of a vertical `gap` creating a little too much space between the chips and the text input.

Screen Shot 2022-09-22 at 7 09 00 PM

I'm not sure what all is intended to wrap there, but I'm wondering if it should just be a `row-gap` anyways.

Also just an idea, but `display: contents` on the chip parent elements would allow us to display the chips and text input inline, and the text input will shrink down to it's min-width, and then wrap and consume the entire next row.

Screen Shot 2022-09-22 at 7 15 51 PM

mattnolting (approved):
Idk about the explicit values for padding that @mcoker mentioned, but LGTM 👍

mcarrano:
I'm thinking inline like the second example, but don't feel strongly. @mmenestr what do you think?

mmenestr:
I think inline in the second example looks good, but is there a minimum width that it should be before it just goes below? Because I feel like in situations where the text input is tiny it might look weird!

srambach:
> I think inline in the second example looks good, but is there a minimum width that it should be before it just goes below? Because I feel like in situations where the text input is tiny it might look weird!

I think it's a 12 character min-width, so that should be fine. I'll update the PR.

srambach:
One thing. Using `display:contents` means that in Safari, the chip group is no longer announced. This is fixed in Chrome and FF, so the chip group is read the same in those browsers. @nicolethoen @jessiehuff do you have any thoughts about the tradeoff and support here?

Currently how it's read in Safari:

https://user-images.githubusercontent.com/19825616/192053463-8ea9c7d7-4ca1-4f79-a748-0aeb4661ba0a.mp4

With `display:contents` in Safari:

https://user-images.githubusercontent.com/19825616/192053551-6c5d5b4b-a374-4062-8946-ab100dfcf68a.mp4

mcoker:
@srambach that _may_ have been fixed in Safari 16. They do have this in the release notes under a11y - https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes

> Added accessibility support for display:contents.

This is what I'm seeing

Sep-23-2022 16-14-34

Tho it's worth noting that our workspaces stopped working on Safari (including pre-v16 for me anyways) in the last release, though they've since been updated. If everything is working OK for you and you're not on 16, you may need to pull down the latest changes in core and `yarn install` the latest patternfly-docs package with the fix.

Also it's fine as-is, this kind of approach seems a bit fragile anyways, though it does seem useful if we want to put something outside of a wrapping flex row inline with the flex items in that row. This might be a good place to test it since text input group is currently beta.

nicolethoen (approved):
I can confirm it appears to work in Safari 16 :)
LGTM

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -17,6 +17,11 @@
   --pf-c-text-input-group__main--first-child--not--text-input--MarginLeft: var(--pf-global--spacer--sm);
   --pf-c-text-input-group__main--m-icon__text-input--PaddingLeft: var(--pf-global--spacer--xl);
 
+  // Chip group
+  --pf-c-text-input-group__chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__chip-group__main--PaddingRight: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__chip-group__main--PaddingBottom: var(--pf-global--spacer--xs);
```
mcoker, line 23:
```suggestion
  --pf-c-text-input-group--c-chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
  --pf-c-text-input-group--c-chip-group__main--PaddingRight: var(--pf-global--spacer--xs);
  --pf-c-text-input-group--c-chip-group__main--PaddingBottom: var(--pf-global--spacer--xs);
```

mcoker (changes requested):
Looks perfect! Just a couple of var updates.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -16,6 +16,13 @@
   // Main
   --pf-c-text-input-group__main--first-child--not--text-input--MarginLeft: var(--pf-global--spacer--sm);
   --pf-c-text-input-group__main--m-icon__text-input--PaddingLeft: var(--pf-global--spacer--xl);
+  --pf-c-text-input-group__main--RowGap: var(--pf-global--spacer--xs);
+  --pf-c-text-input-group__main--ColumnGap: var(--pf-global--spacer--sm);
+
+  // Chip group
+  --pf-c-text-input-group__c-chip-group__main--PaddingTop: var(--pf-global--spacer--xs);
```
mcoker, line 23: With nested components, the common format is `--c-[component]` when its used as a child - https://www.patternfly.org/v4/developer-resources/global-css-variables#component-variables

mcoker (approved):

Issue #4881, opened 2022-05-31
Truncation/disclosure box with inline "show [more/less]" action


mcoker:
Originally requested by @simrandhaliw, they're looking for some sort of truncation or disclosure that gives you the ability to expand/collapse content with an inline action like this "show [more/less]" button.

<img width="1498" alt="Screen Shot 2022-05-31 at 8 59 34 AM" src="https://user-images.githubusercontent.com/35148959/171231254-6b346467-30a6-4f10-8efd-34e9a1224ce0.png">
<img width="1498" alt="Screen Shot 2022-05-31 at 8 59 38 AM" src="https://user-images.githubusercontent.com/35148959/171231257-e67b88e7-acff-40a3-a443-77c95e2f1734.png">

cc @mcarrano @mmenestr

PR #5001, opened 2022-07-26
feat(expandable-section): added truncate variant

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4881

### DIFF
--- a/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
+++ b/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
@@ -8,8 +8,7 @@ cssPrefix: pf-c-expandable-section
 ### Hidden
 ```hbs
 {{#> expandable-section}}
-  {{#> expandable-section-toggle}}
-  {{/expandable-section-toggle}}
+  {{> expandable-section-toggle}}
   {{#> expandable-section-content}}
     This content is visible only when the component is expanded.
   {{/expandable-section-content}}
@@ -19,8 +18,7 @@ cssPrefix: pf-c-expandable-section
 ### Expanded
 ```hbs
 {{#> expandable-section expandable-section--IsExpanded="true"}}
-  {{#> expandable-section-toggle}}
-  {{/expandable-section-toggle}}
+  {{> expandable-section-toggle}}
   {{#> expandable-section-content}}
     This content is visible only when the component is expanded.
   {{/expandable-section-content}}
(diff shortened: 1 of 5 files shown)


mmenestr (approved):
LGTM!

mattnolting (approved):
LPTM!

mcarrano (commented):
@mcoker This looks good, but I just have a couple of comments/questions:

* Now that I see this, I'm questioning whether we should have the caret icon in front of the button (change from what I said before). When I see that, I immediately think that the content will expand below. Maybe we should remove that and just have the text? What do you think @mmenestr ?
* I don't really like the naming of this. What is we call it "Truncated text area" or "Expandable text block"?

mcoker:
> Now that I see this, I'm questioning whether we should have the caret icon in front of the button

@mcarrano I agree, though it's worth noting that it will have a smaller clickable target area if we use the inline link button (no padding), which would remove the left padding and allow the button text to be aligned to the left of the expandable text.

> When I see that, I immediately think that the content will expand below.

It's also worth noting that I flipped the caret icon when expanded, since the default use case would be to expand content above the toggle. This matches our pre-existing detached toggle example. https://patternfly-pr-5001.surge.sh/components/expandable-section#detached-toggle

> What is we call it "Truncated text area" or "Expandable text block"?

Sure, though the variation name is currently "truncate" (`pf-m-truncate`) - do you think that's confusing, or is it just the example name? "text area" also somewhat implies a `<textarea>` (a multi-line form input). What about "Truncated text"?

mcarrano:
> Sure, though the variation name is currently "truncate" (pf-m-truncate) - do you think that's confusing, or is it just the example name? "text area" also somewhat implies a <textarea> (a multi-line form input). What about "Truncated text"?

I'm concerned that people might confuse this with the Truncate component. Maybe "Truncated text" is OK. @mmenestr what do you think?

mmenestr:
@mcarrano I don't mind with caret or without, like @mcoker said, the direction of the arrow matches the direction of the text so I don't think it really matters where it's pointing.. but I get the appeal of just having a link button as well.

In terms of the naming, could we maybe do a combo, and call it something like "Truncate expansion" (?) haha

mcoker:
@mcarrano @mmenestr an issue with the expandable section toggle now is that it has its own button and doesn't use our button component. If we want to support an inline link there, I'd like to look at enhancing that so a regular button component can be used.

Also a side note - should the caret be our secondary color by default, and the primary color on hover/focus/active/expanded?

srambach (approved):
LGTM 👍

mcoker:
per @mcarrano, updated the truncate toggle to remove the icon and extra spacing.

The toggle text is still darker when the text is expanded - is that good or should it stay the same between truncated/expanded? @mmenestr @mcarrano 

Screen Shot 2022-08-01 at 6 35 17 PM

Default variation for comparison

Screen Shot 2022-08-01 at 6 35 30 PM

mcarrano (approved):
This looks good @mcoker . Thanks for making these updates. I am fine with the "Show less" text being darker for consistency with the other variants.

mmenestr (approved):

PR #5002, opened 2022-07-26
fix(menu): removed footer bottom border when scrollable

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5006

Putting this up for review. We added that bottom border when we created the plain variation of the menu in this PR - https://github.com/patternfly/patternfly/pull/4392. The problem is that the bottom border seems unnecessary in both the default component (when the menu has a box shadow) and the plain variation (box shadow is removed), and creates a double border if you put the menu in something with a border, like a flat card - https://codesandbox.io/s/agitated-sanderson-xjy56g?file=/index.html

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -182,8 +182,6 @@
   --pf-c-menu__footer--BoxShadow: none;
   --pf-c-menu__footer--after--BorderTopWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-menu__footer--after--BorderTopColor: var(--pf-global--BorderColor--100);
-  --pf-c-menu__footer--after--BorderBottomWidth: 0;
-  --pf-c-menu__footer--after--BorderBottomColor: var(--pf-global--BorderColor--100);
   --pf-c-menu--m-scrollable__footer--BoxShadow: var(--pf-global--BoxShadow--sm-top);
   --pf-c-menu--m-scrollable__footer--after--BorderTopWidth: 0;
   --pf-c-menu--m-scrollable__footer--after--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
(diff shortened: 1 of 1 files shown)


mcarrano (commented):
I think this looks fine. Is it possible that we just did this because the example [here](https://patternfly-pr-5002.surge.sh/components/menu/#plain-scrollable-with-search-and-footer) looks wrong without a border?  Would I always need to use this inside a card then? Is there a situation where I wouldn't want this inside of a card?

mcoker:
> Is it possible that we just did this because the example [here](https://patternfly-pr-5002.surge.sh/components/menu/#plain-scrollable-with-search-and-footer) looks wrong without a border? Would I always need to use this inside a card then? Is there a situation where I wouldn't want this inside of a card?

@mcarrano maybe? Even with the bottom border, I don't think that example looks good. I agree with @mceledonia's comment here https://github.com/patternfly/patternfly/pull/4392#issuecomment-924164804

> It looks good to me, for the search/footer version I'm thinking given that it will appear in some other context it should work out.

> Is there a situation where I wouldn't want this inside of a card?

It doesn't seem like you'd ever want this menu just floating like that, I'm assuming it would be placed in something that gives it defined edges. Not necessarily always in a card, but somewhere that creates the edge, where the existing border would otherwise need to be removed.

Another thing we could do is create a bordered variation of the menu that creates a border instead of a drop shadow, though I believe we considered that before and agreed the flat card would serve that use case.

mcarrano:
@mcoker I think everything you say makes sense. But I'm still a little reluctant to take this out for fear that it breaks something. @mceledonia is on PTO until next week so maybe we can table this until he returns? Personally the double bottom border did not bother me so much visually.

mmenestr (approved):

srambach (approved):

mattnolting (approved):
LPTM! 🥇

Issue #5009, opened 2022-07-29
Page - some "sticky-on" sections have incorrect bottom padding


mcoker:
main-nav padding doesn't work with `sticky-on-*`

https://github.com/patternfly/patternfly/blob/5a2e36f9f6db8323951a60438fe0cd1b7157a628/src/patternfly/components/Page/page.scss#L545-L555

main-breadcrumbs padding doesn't work with `sticky-on-*`

https://github.com/patternfly/patternfly/blob/5a2e36f9f6db8323951a60438fe0cd1b7157a628/src/patternfly/components/Page/page.scss#L575-L587

main-tabs doesn't work with either (this one may be tricky)

PR #5011, opened 2022-07-29
fix(page): fixed sticky-on section padding

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5009

@mmenestr @mcarrano 

This happens when the section switches between sticky and not. It happened before (a sticky section had more padding than it would if it was not sticky), but it wasn't as visible since the sticky modifier was either on or off - it didn't change dynamically as the viewport height changed.

Jul-29-2022 11-50-58
Jul-29-2022 11-52-01

I wanted to check to see if you think this is an issue. If so, I think we could solve it a couple of ways...

1) Change the padding on the breadcrumb and nav sections so that there is top/bottom padding all of the time (regardless of whether it's sticky or not) that looks appropriate both when the section is sticky and not. That's likely a breaking change as who knows how that might interfere with ways users are using these sections - for example, if they have dividers between sections currently, changing the section padding could move the divider in a way that it doesn't look right.

2) Add this padding to any section that is set to be sticky at some breakpoint, regardless of whether it's actually sticky or not. ie, a breadcrumb section that's sticky on the medium height breakpoint would always have that extra padding, including below the medium breakpoint when it isn't sticky.

**edit:** I'll add that I think #1 would probably be a better long term solution. We're going to run into the same issue using the resize observer where it dynamically applies the sticky class via JS depending on the viewport size. If we go with #2 above, we'll need a unique way to allow the resize observer to attach a class to the section that it's observing that updates the section padding independently from the class that makes the section sticky.

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -548,9 +548,16 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
   padding-left: var(--pf-c-page__main-nav--PaddingLeft);
   background-color: var(--pf-c-page__main-nav--BackgroundColor);
 
-  &.pf-m-sticky-top,
-  .pf-c-page__main-group.pf-m-sticky-top &:last-child {
-    padding-bottom: var(--pf-c-page__main-nav--m-sticky-top--PaddingBottom);
+  // Responsive height modifiers for sticky top/bottom
+  @each $breakpoint, $breakpoint-value in $pf-c-page--height-breakpoint-map {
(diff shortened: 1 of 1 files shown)


mmenestr:
I feel like it's not that big of an issue - and it would only happen if the sticky behavior was applied right? Since that shadow would always be present, I personally feel like it's fine.

mcoker:
@mmenestr yep, when the section is made sticky, you get the shadow and extra padding at the same time.

srambach:
My only comment is that now there's a visible shift to what's below when it's applied - is that desired?

srambach (approved):
Assuming the shift is ok, code looks fine to me.

PR #4928, opened 2022-06-25
docs(docs): fixed duplicate IDs in example docs

evwilkin:
Closes #4927 

This PR resolves duplicate IDs throwing a11y errors identified in https://github.com/patternfly/patternfly-org/issues/2920

PR #5015, opened 2022-08-01
chore(a11y): updated data list and toolbar duplicate IDs follow up

### LINKED ISSUE #5036: Follow-up to duplicate IDs
Additional updates were meant to be included in https://github.com/patternfly/patternfly/pull/4928 but were not committed into that merged code.

DataList:

https://github.com/patternfly/patternfly/pull/4928#discussion_r931157589
I think it would be flagged as an issue - but the aria-labels in the expandable examples are being generated incorrectly. For example, the first expandable content in the expandable compact example has an aria-label= data-list-expandable-compact Primary content details

Toolbar:

https://github.com/patternfly/patternfly/pull/4928#discussion_r930541205
https://github.com/patternfly/patternfly/pull/4928#discussion_r930541466

evwilkin:
Closes #5036 

Additional updates meant to be included in #4928 but not committed into that merged code.

DataList:
- https://github.com/patternfly/patternfly/pull/4928#discussion_r931157589
- > I think it would be flagged as an issue - but the aria-labels in the expandable examples are being generated incorrectly. For example, the first expandable content in the expandable compact example has an aria-label= `data-list-expandable-compact Primary content details`

Toolbar:
- https://github.com/patternfly/patternfly/pull/4928#discussion_r930541205
- https://github.com/patternfly/patternfly/pull/4928#discussion_r930541466


src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -228,7 +228,7 @@ When a list item includes more than one block of content, it can be difficult fo
       {{/data-list-item-content}}
       {{> data-list-item-action}}
     {{/data-list-item-row}}
-    {{#> data-list-expandable-content data-list-expandable-content--attribute=(concat 'id="' data-list--id '-content1" aria-label="' data-list--id ' Primary content details"')}}
+    {{#> data-list-expandable-content data-list-expandable-content--attribute=(concat 'id="' data-list--id '-content1" aria-label="' data-list--id '-primary-content-details"')}}
```
mcoker, line 231: hrrmm.. Since this is an aria-label it's just going to read the string "data-list-expandable-primary-content-details".

The react examples have an aria-label like "First expandable content details".

We could do something like that, or since these are all expandable, the first cell has text wrapped with an `id` that describes the row that we use as part of `aria-labelldby` with the toggle button.

We could update these lines to do the same, something like:

```suggestion
    {{#> data-list-expandable-content data-list-expandable-content--attribute=(concat 'id="' data-list--id '-content1" aria-label="Expandable content details" aria-labelledby="' data-list--id '-' data-list-item--id ' ' data-list--id '-content1"')}}
```

^ untested code

Just an aside, we could clean up these attributes quite a bit so we only have to define an id on the data list and each item, then just use `-toggle` (expandable button), `-title` (title text in primary cell), `-content` (expandable content), etc to create/reference the `id`'s for things, which we (hopefully) never have to change with updates like these.

evwilkin: The untested code above doesn't quite work, it ends up grabbing all of the "Lorem ipsum" text and using it in the aria-labelledby.  I tried a few approaches but didn't have any plain readable text to pull from across examples to create a one-size-fits-all solution, so ended up writing out unique aria-labels for each instance.

evwilkin:
```
0: "Expandable primary content details"
1: "Expandable secondary content details"
2: "Expandable tertiary content details"
3: "Expandable compact primary content details"
4: "Expandable compact secondary content details"
5: "Expandable compact tertiary content details"
6: "Expandable nested primary content details"
7: "Nested row 1 details"
8: "Nested row 2 details"
9: "Nested row 3 details"
10: "Expandable nested secondary content details"
11: "Expandable nested tertiary content details"
12: "Flex modifier expandable primary content details"
13: "Selectable expandable row primary content details"
14: "Selectable expandable row secondary content details"
15: "Selectable expandable row tertiary content details"
16: "Selectable expandable row quaternary content details"
```

mcoker (approved):
🤩

PR #5018, opened 2022-08-02
feat(page): adds full-height modifier

### LINKED ISSUE #4822: Base CSS - Revisit requirement for consumer to declare height on html/body/app-root elements
Currently, PatternFly apps need to add base CSS to ensure the page component will properly take 100% of the browser height. A question came up with regard to why this can't be handled inside PatternFly CSS itself.

From a discussion on #patternfly-core slack with @mcoker :

> [@mturley]
> Hey, question for core CSS folks. @vojtechszocs brought up an interesting point in [this PR comment](https://github.com/openshift/dynamic-plugin-sdk/pull/74/files#r860089316) : the patternfly-react-seed has this custom CSS to make page layouts work correctly:
> ```css
> html, body, #app {
>   height: 100%;
> }
> ```
> is there a reason this is not included in the base styles of PatternFly itself? Is it simply because we don't want to bake in the assumption that the consumer will use #app as their main container id? (we could maybe use a class then?)

> [@mcoker]
> Hey! Yep, that’s effectively it. Maybe more broadly, we need everything from `<body>` to `<div class="pf-c-page">` to have height declared, and that varies per product. I think I’ve most commonly seen it as `#root`, too.
> It’s to get around an annoying issue in some screens where something simple like `height: 100vh` on the page component doesn’t work.
> There is a cool new feature that (`:has()`) that could select all ancestors of `.pf-c-page` and give it a height, though it’s really new and isn’t supported in all of the major browsers yet. Only works in Safari at the moment.
> https://codepen.io/mcoker/pen/eYVOjQE
> I think it’s worth looking at again and see if there is a better way to handle that.

srambach:
This adds a modifier to the page component that sets the height using viewport units to eliminate the need to ensure that all elements above the page are set to 100% height.

Fixes #4822

### DIFF
--- a/src/patternfly/demos/Page/examples/Page.md
+++ b/src/patternfly/demos/Page/examples/Page.md
@@ -13,7 +13,7 @@ wrapperTag: div
 
 ### Full height page
 Using the `.pf-m-full-height` modifier class on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other ancestors of `.pf-c-page`, have height set to 100%.
-```hbs isFullscreen
+```hbs isFullscreen isBeta
 {{> page-template page-template--id="page-demo-full-height" page-template--modifier="pf-m-full-height"}}
 ```
 
--- a/src/patternfly/demos/Page/examples/Page.md
+++ b/src/patternfly/demos/Page/examples/Page.md
@@ -12,7 +12,7 @@ wrapperTag: div
 ```
 
 ### Full height page
-Using the `pf-m-full-height` modifier on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other tags above `.pf-c-page`, have height set to 100%.
+Using the `.pf-m-full-height` modifier class on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other ancestors of `.pf-c-page`, have height set to 100%.
 ```hbs isFullscreen
 {{> page-template page-template--id="page-demo-full-height" page-template--modifier="pf-m-full-height"}}
 ```
@@ -116,4 +116,4 @@ Using the `pf-m-full-height` modifier on the page component eliminates the need 
 ```
 
 ## Documentation
-To make the page component take up the full height of the viewport, it is recommended to add `height: 100%;` to all ancestor elements of the page component.
+To make the page component take up the full height of the viewport, it is recommended to add `height: 100%;` to all ancestor elements of the page component. Alternatively, use the `.pf-m-full-height` modifier class on the page component.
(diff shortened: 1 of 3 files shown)


src/patternfly/demos/Page/examples/Page.md
```diff
@@ -11,6 +11,12 @@ wrapperTag: div
 {{> page-template page-template--id="page-demo-basic"}}
 ```
 
+### Full height page
+Using the `pf-m-full-height` modifier on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other tags above `.pf-c-page`, have height set to 100%.
```
mcoker, line 15:
```suggestion
Using the `.pf-m-full-height` modifier on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other ancestors of `.pf-c-page`, have height set to 100%.
```

mcoker (changes requested):
🤗 so cool! Still testing but saw this one bit in the docs.

src/patternfly/demos/Page/examples/Page.md
mcoker: Oh sorry I meant to give more context here. Wanted to say maybe use `.` with the class and/or say "modifier _class_" just so it's clear it's a class. Also that we use "ancestors" on the react side when referring to the elements above page, that seems a little clear. https://www.patternfly.org/v4/components/page/react-demos

> To make the page take up the full height, it is recommended to set the height of all ancestor elements up to the page component to 100%

mcoker (changes requested):
Woooo! This is great. I was testing on mobile safari and saw the notorious `100vh` bug, but dug in a little and looks like it's because that's used in the website/workspace CSS 🤦‍♂️ - https://github.com/patternfly/patternfly-org/blob/532c94559691a461aaa248851e150744b15f01ab/packages/documentation-framework/global.css#L2-L6

```
#root,
#ws-router,
#ws-page {
  height: 100vh;
}
```

I'd love to merge this, then update the workspaces/website to test it out. I'm thinking we would
* replace `height: 100vh` with `100%` on those elements just for consistency with what we recommend to users
* add `.pf-m-full-height` to the page component that serves the docs.

One thing this would break is with our current demo for how to put a banner above and below the page component, but I'm wondering if we could do something better there instead. For reference, here's the [core banner demo](https://www.patternfly.org/v4/components/banner/html-demos/topbottom/) and the [react banner demo](https://www.patternfly.org/v4/components/banner/react-demos/top-and-bottom-banner/)

mcoker (approved):
🚀

mattnolting (commented):
I think this is looking great. I would suggest adding a `.pf-c-page__top-container` and `.pf-c-page__bottom-container` for miscellaneous items (banners) that appear outside of standard page elements. Then we could move banners into those elements for corresponding demos. 

@mcoker wdyt?

mattnolting (approved):
LGTM! 🎉

mturley:
Awesome!

Issue #4845, opened 2022-05-10
Checkbox - Add the ability to indicate that a checkbox is required


mcarrano:
**Is this a new component or an extension of an existing one?**
Checkbox

**Describe the feature**
The may be cases where you want to indicate that a checkbox must be set to proceed, e.g. when the user must acknowledge terms and conditions to open an account. In this case, just as for labeled form fields, it should be possible to place a red asterisk to the right of a checkbox label.

mcarrano:
@blfetche we did not create design mocks for this, but if you need advise on spacing or related visuals, I'd suggest checking with @mceledonia . This would mainly be for the case of a standalone checkout that must be checked to move forward with a process. Otherwise checkboxes are naturally optional by definition.

blfetche:
> @blfetche we did not create design mocks for this, but if you need advise on spacing or related visuals, I'd suggest checking with @mceledonia . This would mainly be for the case of a standalone checkout that must be checked to move forward with a process. Otherwise checkboxes are naturally optional by definition.

Ok thank you.

PR #5025, opened 2022-08-05
feat(checkbox): checkbox required indicator added

blfetche:
Areas Updated for #4845 

image

image

image
I think this one may be incorrect as it doesn't create the required indication but, instead styles it. Please let me know if it should be removed. Thanks.


src/patternfly/components/Check/check-label.hbs
```diff
@@ -3,4 +3,7 @@
     {{{check-label--attribute}}}
   {{/if}}>
   {{> @partial-block}}
+  {{#if required}}
```
mcoker, line 6: We generally try to namespace all of the hbs params to the component and/or hbs file where they're used. In this case, since it's the check component is either required or not, that param could go on the main `{{#> check}}` block, like `{{#> check check--IsRequired}}`.

Then you can use `{{#if check--IsRequired}} {{> check-label-required}}{{/if}}` here and `{{#if check--IsRequired}} required{{/if}}` as an attribute in the check input hbs, similar to how we add the `checked` attribute: https://github.com/patternfly/patternfly/blob/9ab78a81b9d6d63cd7795fa82b1ff673abba5332/src/patternfly/components/Check/check-input.hbs#L3-L5

src/patternfly/components/Check/check.scss
```diff
@@ -11,6 +11,11 @@
   --pf-c-check__description--Color: var(--pf-global--Color--200);
   --pf-c-check__body--MarginTop: var(--pf-global--spacer--sm);
 
+  // Required checkbox
+  --pf-c-form__label-required--MarginLeft: var(--pf-global--spacer--xs);
```
mcoker, line 15: just a nit - these var names should match the elements they apply to.

```suggestion
  --pf-c-check__label-required--MarginLeft: var(--pf-global--spacer--xs);
```

src/patternfly/components/Check/examples/Check.md
```diff
@@ -102,14 +110,18 @@ If you extend this component or modify the styles of this component, then make s
 | Attribute | Applied to | Outcome |
 | -- | -- | -- |
 | `disabled` | `<input type="checkbox">` | Indicates that the element is unavailable and removes it from keyboard focus. **Required when input is disabled** |
+| `required` | `<input type="checkbox">` | Indicates that the element is required. |
```
mcoker, line 113: This is a pre-existing thing, but these attrs can refer to the component class instead of the html element. the only place we need to mention the html element is below when we're documenting the component class.

```suggestion
| `disabled` | `.pf-c-check__input` | Indicates that the element is unavailable and removes it from keyboard focus. **Required when input is disabled** |
| `required` | `.pf-c-check__input` | Indicates that the element is required. |
```

src/patternfly/components/Check/examples/Check.md
```diff
@@ -102,14 +110,18 @@ If you extend this component or modify the styles of this component, then make s
...
 | `.pf-c-check` | `<div>`, `<label>` |  Initiates the check component. **Required**  |
 | `.pf-c-check__input` | `<input type="checkbox">` |  Initiates a check input. **Required**  |
 | `.pf-c-check__label` | `<label>`, `<span>` |  Initiates a label. **Required**  |
+| `.pf-c-check__label-required` | `<span>` |  Initiates a required indicator. |
 | `.pf-c-check__description` | `<span>` |  Initiates a check description. |
 | `.pf-c-check__body` | `<span>` |  Initiates a check body. |
 | `.pf-m-standalone` | `.pf-c-check` |  Modifies the check component for use with a standalone `<input type="checkbox">`. **Required when there is no label** |
 | `.pf-m-disabled` | `.pf-c-check__label` |  Modifies the check component for the disabled state. **Required when input is disabled** |
+
+
```
mcoker, line 127: Looks like there are some extra lines at the end of this file. We should just have the one newline character at the end of the last word in the file like mentioned above.

mcoker (changes requested):
Great job, this looks really good! Just a few small things. Also I called it out about changing the hbs parameter `required` to `check--IsRequired` but you'll need to add the `required` attribute to the text intput, too.

src/patternfly/components/Check/check.scss
blfetche: Thank you, should have caught this.

src/patternfly/components/Check/check-label.hbs
blfetche: Can we discuss it tomorrow? I just want to make sure I understand correctly. Thanks.

src/patternfly/components/Check/check-label.hbs
```diff
@@ -3,4 +3,7 @@
     {{{check-label--attribute}}}
   {{/if}}>
   {{> @partial-block}}
+{{#if check--IsRequired}} 
+  {{> check-label-required}}
+{{/if}}
 </{{#if check-label--type}}{{check-label--type}}{{else}}label{{/if}}>
```
mcoker, line 9: Just one nit! Need to indent this block since it's nested

```suggestion
  {{> @partial-block}}
  {{#if check--IsRequired}} 
    {{> check-label-required}}
  {{/if}}
</{{#if check-label--type}}{{check-label--type}}{{else}}label{{/if}}>
```

mcoker (changes requested):
Just one small thing, otherwise looks perfect!

mattnolting (approved):
LGTM, great work @blfetche!!

mcoker (approved):
Woooo!!! 😎👍

PR #5029, opened 2022-08-10
feat(button): adds a count to the link button

### LINKED ISSUE #5012: Button - Need new variant that retains state for opening and closing a drawer
A request has been made to create a button variant to be used as a drawer toggle. Contingent on completion of: https://github.com/patternfly/patternfly-design/issues/1173

srambach:
This adds a count as an optional component of the button. Design feels this should only be used with the link button (and so only that variant is shown as an example), but for completeness we've made sure the other variants display adequately. 
Direct link: https://patternfly-pr-5029.surge.sh/components/button#link-button-with-a-count
The other variants are shown here:
image

Other decisions per @mmenestr and @mceledonia:
- The badge color does not change on hover
- Disabling the button causes the badge (even if blue/unread) to display as read/gray.

Fixes #5012

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -282,7 +282,7 @@
       --pf-c-button--m-primary--BackgroundColor: var(--pf-c-button--m-primary--active--BackgroundColor);
     }
 
-    &:not(:disabled):not(.pf-m-disabled) .pf-c-badge.pf-m-unread {
+    .pf-c-badge.pf-m-unread {
       border: var(--pf-c-button--m-primary__c-badge--BorderWidth) solid var(--pf-c-button--m-primary__c-badge--BorderColor);
     }
   }
@@ -576,9 +576,10 @@
     color: var(--pf-c-button--disabled--Color);
     background-color: var(--pf-c-button--disabled--BackgroundColor);
 
-    & .pf-c-badge {
+    .pf-c-badge {
       --pf-c-badge--m-unread--Color: var(--pf-c-button--disabled__c-badge--Color);
       --pf-c-badge--m-unread--BackgroundColor: var(--pf-c-button--disabled__c-badge--BackgroundColor);
+      --pf-c-button--m-primary__c-badge--BorderWidth: 0;
     }
   }
 
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -192,6 +192,13 @@
   --pf-c-button--m-in-progress--m-plain__progress--Left: 50%;
   --pf-c-button--m-in-progress--m-plain__progress--TranslateX: -50%;
 
+  // Count
+  --pf-c-button__count--MarginLeft: var(--pf-global--spacer--sm);
+  --pf-c-button--disabled__c-badge--Color: var(--pf-global--Color--dark-100);
+  --pf-c-button--disabled__c-badge--BackgroundColor: var(--pf-global--BackgroundColor--200);
+  --pf-c-button--m-primary__c-badge--BorderColor: var(--pf-global--BorderColor--300);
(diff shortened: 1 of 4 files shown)


mmenestr:
Wondering whether we should show the read or the unread state, or both for the link button.. @mcarrano what do you think?
Otherwise looks great !!

src/patternfly/components/Button/button.scss
```diff
@@ -274,6 +281,10 @@
       --pf-c-button--m-primary--Color: var(--pf-c-button--m-primary--active--Color);
       --pf-c-button--m-primary--BackgroundColor: var(--pf-c-button--m-primary--active--BackgroundColor);
     }
+
+    &:not(:disabled):not(.pf-m-disabled) .pf-c-badge.pf-m-unread {
```
mcoker, line 285: it looks like you might be able to drop the specificity here, if you just set `--pf-c-button--m-primary__c-badge--BorderWidth: 0` below in your `[disabled selectors] .pf-c-badge {}` block

```suggestion
    .pf-c-badge.pf-m-unread {
```

src/patternfly/components/Button/button.scss
```diff
@@ -564,6 +575,11 @@
 
     color: var(--pf-c-button--disabled--Color);
     background-color: var(--pf-c-button--disabled--BackgroundColor);
+
+    & .pf-c-badge {
+      --pf-c-badge--m-unread--Color: var(--pf-c-button--disabled__c-badge--Color);
+      --pf-c-badge--m-unread--BackgroundColor: var(--pf-c-button--disabled__c-badge--BackgroundColor);
```
mcoker, line 581: Looks like you can probably just remove the disabled border here instead of using `:not()` above to create the border.

```suggestion
      --pf-c-badge--m-unread--BackgroundColor: var(--pf-c-button--disabled__c-badge--BackgroundColor);
      --pf-c-button--m-primary__c-badge--BorderWidth: 0;
```

mcoker (changes requested):
Looks great, left some comments. Also I'm assuming the baseline alignment is good @mmenestr @mcarrano? The text in the button aligns with the text in the badge, so the badge itself is shifted down a bit vertically and isn't center aligned.

src/patternfly/components/Button/button.scss
srambach: @mmenestr says it should only be used on the link button, however, if someone did put it on another button there were some bad things happening with contrast. So a minimal amount of css puts a guardrail there. I added a note in the documentation saying only to use it with link buttons.

srambach:
> Looks great, left some comments. Also I'm assuming the baseline alignment is good @mmenestr @mcarrano? The text in the button aligns with the text in the badge, so the badge itself is shifted down a bit vertically and isn't center aligned.

@mceledonia checked out the alignment and was ok with that.

mcarrano:
This looks great @srambach . Can you just update the example to show an instance of the button with an unread as well as a read badge? Thanks.

mcoker (approved):
LGTM. FWIW the disabled control button has the same background we're using for the disabled badge background 

Screen Shot 2022-08-11 at 7 22 30 PM

mcarrano (approved):
Thanks for adding the unread example @srambach . This looks good to me. I'm not so concerned about the disabled styling. We can address that down the road if it's an issue. @mmenestr do you agree?

srambach:
@mcarrano @mmenestr @mceledonia  Here are the buttons in light and dark, with the control button added. Let me know if you'd like any changes.
image

mmenestr:
Yeah, i'd say i'm not too concerned about the disabled styling either.
For the dark mode primary button - could we make the outline of the badge white as well in the unread state? It's hard to see the grey outline on that blue background!

Otherwise LGTM!

srambach:
image

mmenestr (approved):

Issue #5028, opened 2022-08-04
base-no-reset.css overrides the base html font size


warmbowski:
```
html {
  font-size: unset !important;
}
```

`base.css` and `base-no-reset.css` contains what seems like a pretty egregious misuse of `!important` by overriding the base font-size for the html element with an `unset`. This seems especially odd to do in the "no-reset" css file, which by definition shouldn't reset that element. Ideally, this should really have an escape hatch using css variables for folks who aren't using @patternfly/react-* generally for their web site. For my use, I am integrating one particular react component (log-viewer) into an established site that uses MUI react lib and it's theming generally. All base font sizes are set in 'rem' units, so messing with the base html element font-size cascades into messing up all base font sizes. 

Can you add a variable to override in future versions in the base css files, and/or take completely out of the no-reset css?

Marc19:
We are experiencing the same issue. All of the fonts are affected and it's currently blocking us.
@mcoker Do you have an idea when is this issue going to be tackled? In the interim, is there some way you can think of to prevent this behavior? @warmbowski Have you figured out something? Thanks!

warmbowski:
Yeah, basically we worked around it by using React Helmet in our app to add a style tag with the css to override it with a `!important`.

```
html { font-size: 14px !important }

```

Less than ideal, but it works.

mcoker:
Hey @Marc19 and @warmbowski, thanks for the issue. Just wanted to check in and say I'm hoping to have a look at this tomorrow and see if there is anything we can do. I agree this is something you should be able to opt out of easily.

You can see some context in the comments as to why that style was added, and that it's able to be removed by setting `$pf-global--unset-root-font-size: false;` and compiling the PF SCSS, if that's an option currently.

https://github.com/patternfly/patternfly/blob/9ab78a81b9d6d63cd7795fa82b1ff673abba5332/src/patternfly/base/_common.scss#L12-L21

PR #5032, opened 2022-08-11
fix(global): added custom property to override root font-size

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5028

This PR allows you to set a custom `font-size` with this rule that was added to PF4 to override PF3's `font-size` on `html`

https://github.com/patternfly/patternfly/blob/9ab78a81b9d6d63cd7795fa82b1ff673abba5332/src/patternfly/base/_common.scss#L12-L21

An issue with updating this rule is that it's so fundamental, if we introduce a change that inadvertently causes an issue with the way someone is using PF now, it would likely impact the entire site/app.

The solution in this PR is probably the simplest approach, though it does require a value passed for the `font-size`, which may be a limitation since you can't allow `html` to `inherit` its `font-size` from some other dynamic/unknown style. To override the `font-size` with this PR, you would just declare the custom property like so:

```
:root {
  --pf-global--root--FontSize: [custom value];
}
```

Would this work @Marc19 and @warmbowski? I like this because it doesn't add any bloat, and is overridable like any of our other global vars. 

## Other options

1. Update the current selector that unsets the `font-size` currently to:

    ```
    html:where(:not(.pf-no-root-font-size)) {
      font-size: var(--pf-global--root--FontSize, unset) !important;
    }
    ```
    * This would allow an opt-out at the class level to disable the declaration entirely and inherit from some other system without having to explicitly set the `font-size`. This seems like the best of both worlds - you can opt out of the declaration entirely with a class on the `html` element, and override the size with a custom value if preferred.
      * My only hesitation with this is that `:where()` was introduced relatively recently (though it's within our officially supported browsers) and may break some users, which we always try to avoid if we can unless we're fixing a critical bug and there is no other workaround. We could write a fallback with `@supports not selector(:where(...)) {}` with the original rule, but `@supports selector` support is similar to `:where()`, with safari at least.
1. Add another compiled stylesheet that omits the rule by setting `$pf-global--unset-root-font-size: false;`, similar to [patternfly-base-no-reset.scss](https://github.com/patternfly/patternfly/blob/main/src/patternfly/patternfly-base-no-reset.scss). This would also be fine, but I'd like to avoid adding another compiled stylesheet if possible since it has the potential to get a little complicated to combine different features in these custom compiled variations.

cc @srambach @mattnolting

### DIFF
--- a/src/patternfly/base/_base.scss
+++ b/src/patternfly/base/_base.scss
@@ -1,11 +1,13 @@
 // stylelint-disable no-invalid-position-at-import-rule
 
+// remove in breaking change
 @if $pf-global--load-pf-3 {
   @import url("https://fonts.googleapis.com/css?family=Open+Sans");
   @import url("https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.37.10/css/patternfly.min.css");
   @import url("https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.37.10/css/patternfly-additions.min.css");
 }
 
+// remove pf3 shield styles in breaking change
 @import "shield-inheritable";
 @import "shield-noninheritable";
 
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -1,4 +1,5 @@
 // Set common reset styles for patternfly components
+// remove in breaking change. this should come from globals, and can be opted out
 [class*="pf-c-"] {
   &,
   &::before,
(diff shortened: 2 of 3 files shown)


Marc19:
Hey @mcoker, I think this should work for us 👍 
Thanks for addressing this!

mattnolting (approved):
Adding a passive variable plus a fallback consistent with the current implementation seems totally reasonable to me. Option 1 FTW.

mcoker:
@warmbowski - just wanted to check and see if this solution will work for you. FYI we have code freeze EOD today, so if we can get it merged it will be in our next release.

warmbowski:
This works for me. Thanks!

PR #5039, opened 2022-08-17
feat(table): added max-height to outerscroll

### LINKED ISSUE #5038: Table OuterScrollContainer - add max height support
Table OuterScrollContainer 

**Describe the feature**

In response to: https://github.com/patternfly/patternfly-react/issues/7425

@kmcfaul The codesandbox implementation is correct (except that we wouldn't need a wrapping div w/height with the following) and the behavior is what I would expect to see. The `OuterScrollContainer` is set to `display: flex; direction: column;` and will adjust to the parent height for whatever it contains, but it also requires a `min-height` so that this doesn't happen. 

Screen Shot 2022-07-28 at 10 24 25 AM

**We could update to something like this:**

```
    --pf-c-scroll-outer-wrapper--MinHeight: 16rem;
    --pf-c-scroll-outer-wrapper--MaxHeight: 100%;

    display: flex;
    flex-direction: column;
    max-width: 100%;
    min-height: min(100%, var(--pf-c-scroll-outer-wrapper--MinHeight));
    max-height: min(100%, var(--pf-c-scroll-outer-wrapper--MaxHeight));
    overflow: hidden;
```

**Where you can specify the min and maximum height inline on `OuterScrollContainer`** 

```
<div class="pf-c-scroll-outer-wrapper" style="--pf-c-scroll-outer-wrapper--MinHeight: 6rem; --pf-c-scroll-outer-wrapper--MaxHeight: 12rem;">
  {toolbar} 
  <div class="pf-c-scroll-inner-wrapper">  
    {scrollable things}
  </div>
</div>
```

mattnolting:
closes #5038

### DIFF
--- a/src/patternfly/components/Table/table-scrollable.scss
+++ b/src/patternfly/components/Table/table-scrollable.scss
@@ -48,7 +48,6 @@
   display: flex;
   flex-direction: column;
   max-width: 100%;
-  height: 100%;
   min-height: var(--pf-c-scroll-outer-wrapper--MinHeight);
   max-height: var(--pf-c-scroll-outer-wrapper--MaxHeight);
   overflow: hidden;
--- a/src/patternfly/components/Table/table-scrollable.scss
+++ b/src/patternfly/components/Table/table-scrollable.scss
@@ -43,12 +43,14 @@
 
 .pf-c-scroll-outer-wrapper {
   --pf-c-scroll-outer-wrapper--MinHeight: #{pf-size-prem(400px)};
+  --pf-c-scroll-outer-wrapper--MaxHeight: 100%;
 
   display: flex;
   flex-direction: column;
   max-width: 100%;
+  height: 100%;
   min-height: var(--pf-c-scroll-outer-wrapper--MinHeight);
-  max-height: 100%;
+  max-height: var(--pf-c-scroll-outer-wrapper--MaxHeight);
   overflow: hidden;
 }
 


src/patternfly/components/Table/table-scrollable.scss
```diff
@@ -43,12 +43,14 @@
 
 .pf-c-scroll-outer-wrapper {
   --pf-c-scroll-outer-wrapper--MinHeight: #{pf-size-prem(400px)};
+  --pf-c-scroll-outer-wrapper--MaxHeight: 100%;
 
   display: flex;
   flex-direction: column;
   max-width: 100%;
+  height: 100%;
```
mcoker, line 51: did you mean to add this? I don't see reference to it and looks like it could change an existing layout if the `min-height` was unset to allow a table to grow to the `max-height`, then start scrolling. 

min-height auto

Screen Shot 2022-08-18 at 5 56 29 PM

^ w/ height: 100%

Screen Shot 2022-08-18 at 5 56 33 PM

From what I tested, `height: 100%` doesn't seem necessary (at least in the codesandbox @kmcfaul created), but I don't know all of the use cases.

mattnolting: > did you mean to add this?

@mcoker oops, nope

mcoker (approved):
L🎃TM!

mcoker:
@mattnolting @kmcfaul do we want to add a prop or anything for the min/max-heights, or just leave them as CSS vars? If y'all want to add a prop, would you mind creating an issue for that in the patternfly-react repo? Unless one exists already - I looked and didn't see one.

PR #5048, opened 2022-08-19
feat(notification-badge): added expanded state

### LINKED ISSUE #4664: Add selected state for masthead items

mcoker:
fixes #4664

### DIFF
--- a/src/patternfly/components/NotificationBadge/examples/NotificationBadge.md
+++ b/src/patternfly/components/NotificationBadge/examples/NotificationBadge.md
@@ -12,87 +12,15 @@ import './NotificationBadge.css'
 
 ```hbs
 <div class="pf-t-dark">
-{{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Notifications"'}}
-  {{#> notification-badge notification-badge--modifier="pf-m-read"}}
-    <i class="pf-icon-bell" aria-hidden="true"></i>
-  {{/notification-badge}}
-{{/button}}
-
-{{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Unread notifications"'}}
-  {{#> notification-badge notification-badge--modifier="pf-m-unread"}}
-    <i class="pf-icon-bell" aria-hidden="true"></i>
-  {{/notification-badge}}
-{{/button}}
-
-{{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Attention notifications"'}}
-  {{#> notification-badge notification-badge--modifier="pf-m-attention"}}
-    <i class="pf-icon-attention-bell" aria-hidden="true"></i>
-  {{/notification-badge}}
-{{/button}}
-
-<br><br>
-
-{{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Tasks"'}}
-  {{#> notification-badge notification-badge--modifier="pf-m-read"}}
-    <i class="pf-icon-task" aria-hidden="true"></i>
-  {{/notification-badge}}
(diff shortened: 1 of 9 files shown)


mcarrano (approved):
Looks good to me @mcoker . Do we need any follow up to start using this in out masthead demos and examples?

mcoker:
@mcarrano yep, that's probably a good idea. I can go through the core demos and make any updates needed in this PR, or we can make that a separate task. I already updated the notification drawer demos in this PR which _I think_ is all that needs to be updated. Then we'll need a react issue to do the same.

We may also consider removing the notification bell from the default masthead used in full-page demos, and only including it when clicking on it opens a notification drawer. Otherwise the basic full-page demo masthead would just include generic actions and basic functionality (a dropdown or whatever would trigger a menu, but with generic items).  WDYT?

mcarrano:
That all sounds good @mcoker . Maybe open follow-up issues and I'll pull them into a future milestone?

src/patternfly/components/NotificationBadge/examples/NotificationBadge.md
```diff
@@ -52,6 +52,50 @@ import './NotificationBadge.css'
 </div>
 ```
 
+### Expanded
+
+```hbs
+<div class="pf-t-dark">
+{{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Notifications" aria-expanded="true"'}}
```
mattnolting, line 59: Were you using `aria-expanded="true"` here instead of `button--IsExpanded="trye"` to avoid applying `pf-m-expanded` to the button itself?

src/patternfly/components/NotificationBadge/notification-badge.scss
```diff
@@ -108,9 +111,15 @@
     color: var(--pf-c-notification-badge--m-attention--Color);
 
     &:hover {
-      --pf-c-notification-badge--after--BackgroundColor: var(--pf-c-notification-badge--m-attention--hover--after--BackgroundColor);
+      --pf-c-notification-badge--m-attention--after--BackgroundColor: var(--pf-c-notification-badge--m-attention--hover--after--BackgroundColor);
```
mattnolting, line 114: 👍

mattnolting (commented):
LGTM 👍 Just a couple comments.

src/patternfly/components/NotificationBadge/examples/NotificationBadge.md
mcoker: Yeah, it's only supported on the control button, and noticed the notification badge button in the masthead was using the expanded class, so updated that just to be `aria-expanded`, too. Looking at the react component, I think we'll want to add that there, too - https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/NotificationBadge/NotificationBadge.tsx

srambach:
Just checking - is this the intended dark theme color (especially for the attention one)? Just seems off to me. 🤷 
image

mcoker:
@srambach good callouts. I have an issue here for the attention icon color - https://github.com/patternfly/patternfly/issues/5049. For now I've updated it to be black to match our danger button.

The read icon should be white (or whatever the hover color is) when expanded. Updated that, too!

srambach (approved):
LGTM 🔔

PR #5052, opened 2022-08-22
chore(build): updated to documentation-framework@1.2.5

mcoker:
I noticed the docs build was failing in https://github.com/patternfly/patternfly/pull/5037:

> './node_modules/@patternfly/documentation-framework/components/example/example.js\n'
> "Module not found: Error: Can't resolve '@patternfly/react-core/next' in '/home/runner/work/patternfly/patternfly/node_modules/@patternfly/documentation-framework/components/example'\n"

So I added `@patternfly/react-core` as a dependency, too, since it wasn't before, and all looks good.

Is that the right fix for this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@patternfly/patternfly-a11y": "4.3.1",
     "@patternfly/react-charts": "6.84.8",
     "@patternfly/react-code-editor": "4.72.8",
+    "@patternfly/react-core": "4.234.1",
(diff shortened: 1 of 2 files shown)


jeffpuzzo (approved):

jeffpuzzo:
As mentioned in slack, you can also try to update an existing dependency like `react-table` to the latest so that a later version of react-core is installed.

PR #5059, opened 2022-08-26
feat(Inputs): add plain read only variant

### LINKED ISSUE #4935: Form inputs - Provide styles for read-only inputs that are visually similar to description list

thatblindgeye:
Closes #4935

### DIFF
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -18,6 +18,8 @@ import './FormControl.css'
 <br><br>
 {{> form-control controlType="input" input="true" form-control--attribute='readonly type="text" value="Readonly" id="input-readonly" aria-label="Readonly input example"'}}
 <br><br>
+{{> form-control controlType="input" input="true" form-control--modifier="pf-m-plain" form-control--attribute='readonly type="text" value="Readonly plain" id="input-readonly" aria-label="Readonly input example"'}}
+<br><br>
 {{> form-control controlType="input" input="true" form-control--attribute='disabled type="text" value="Disabled" id="input-disabled" aria-label="Disabled input example"'}}
 <br><br>
 {{> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute='type="text" value="Success" id="input-success" aria-label="Success state input example"'}}
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -18,7 +18,7 @@ import './FormControl.css'
 <br><br>
 {{> form-control controlType="input" input="true" form-control--attribute='readonly type="text" value="Readonly" id="input-readonly" aria-label="Readonly input example"'}}
 <br><br>
-{{> form-control controlType="input" input="true" form-control--modifier="pf-m-plain" form-control--attribute='readonly type="text" value="Readonly plain" id="input-readonly" aria-label="Readonly input example"'}}
+{{> form-control controlType="input" input="true" form-control--modifier="pf-m-plain" form-control--attribute='readonly type="text" value="Readonly plain" id="input-readonly-plain" aria-label="Readonly plain input example"'}}
 <br><br>
 {{> form-control controlType="input" input="true" form-control--attribute='disabled type="text" value="Disabled" id="input-disabled" aria-label="Disabled input example"'}}
 <br><br>
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -127,6 +127,10 @@ Standard
 Readonly
 {{/form-control}}
 <br><br>
+{{#> form-control controlType="textarea" form-control--modifier='pf-m-plain' form-control--attribute='readonly name="textarea-plain-readonly" id="textarea-plain-readonly" aria-label="Plain readonly textarea example"'}}
+Readonly
+{{/form-control}}
+<br><br>
 {{#> form-control controlType="textarea" form-control--attribute='disabled name="textarea-disabled" id="textarea-disabled" aria-label="Disabled textarea example"'}}
 Disabled
 {{/form-control}}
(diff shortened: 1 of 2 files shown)


src/patternfly/components/FormControl/form-control.scss
```diff
@@ -243,6 +248,12 @@ $pf-c-form-control--m-clock--Coordinates: "M256 8C119 8 8 119 8 256s111 248 248
     }
   }
 
+  &[readonly].pf-m-plain {
```
mcoker, line 251: You can move this up as the last child under the existing `&[readonly]` and add `&.pf-m-plain`. It'll generate the same selector, and group these styles all under `readonly`.

mcoker (changes requested):
This looks really good! A couple of comments. The first bit is all I think that needs to be fixed.

* There is a bottom-border/padding change that happens on focus that seems to still be happening. You can only see the border if you focus in dev tools or disable the outline:
  * Aug-26-2022 15-56-31

* This may be OK, but since you're setting the border width to 0, the plain control is a different height than non-plain. So if you swap between plain and non-plain, you're gonna get some layout shift action (text will shift because of the top-border change). You might look at setting the border to `transparent` instead. The disabled control does that, and just sets the `border-color` property to avoid having to set all four color vars https://github.com/patternfly/patternfly/blob/2da6d2170f2806947734303e9eeecd32e537f8e3/src/patternfly/components/FormControl/form-control.scss#L280
  * Screen Shot 2022-08-26 at 3 46 49 PM
  * Screen Shot 2022-08-26 at 3 47 03 PM

That may be OK since design is good with the layout shift - though if we're going to maintain the vertical/block/y-axis control padding with this plain variation, I'm wondering if we should keep the same overall size of the element, so the shift when switching from plain to non-plain is minimal and only from the horizontal/inline/x-axis padding.

Here are some gifs of 3 options:

### The PR now - slight vertical shift from top/bottom borders changing
Aug-26-2022 16-32-35

### No vertical shift
Aug-26-2022 16-32-48

### Lots of shift - the input has no padding/height/borders when plain
Aug-26-2022 16-32-40

thatblindgeye:
@mcoker Made suggested changes! For now I opted to update things so that the input should only shift horizontally when being updating from `pf-m-plain` to non-plain/non-readonly. @mcarrano @jgiardino do either of you have a preference in terms of the clips posted in the comment above showing some different shifting amounts?

mcarrano:
@thatblindgeye I agree with your judgement to opt for only shifting horizontally when shifting between plain and bordered inputs.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -241,6 +245,17 @@ $pf-c-form-control--m-clock--Coordinates: "M256 8C119 8 8 119 8 256s111 248 248
...
+
+    &.pf-m-plain {
+      --pf-c-form-control--readonly--BackgroundColor: var(--pf-c-form-control--readonly--m-plain--BackgroundColor);
+      --pf-c-form-control--inset--base: var(--pf-c-form-control--readonly--m-plain--inset--base);
+
+      border-color: transparent;
+
+      &:focus {
+        --pf-c-form-control--focus--BorderBottomColor: transparent;
+      }
```
mcoker, line 257:
```suggestion
```

Doesn't look like you need this...
1. The specificity of `.pf-c-form-control[readonly]:not(.pf-m-success):not([aria-invalid=true]):focus` (whew 😅) overrides this var declaration, but that doesn't really matter since...
2. Where you declared [`border-color: transparent;`](https://github.com/patternfly/patternfly/blob/1a8ce3dc99f3554c3069d299afc56f186d8bfb63/src/patternfly/components/FormControl/form-control.scss#L253), you're no longer referencing the custom property/var for the `.pf-m-plain` border-color, and since the [original `:focus` selector](https://github.com/patternfly/patternfly/blob/1a8ce3dc99f3554c3069d299afc56f186d8bfb63/src/patternfly/components/FormControl/form-control.scss#L266) to update the border color updates a var, your declaration of the `border-color` property will also apply to `:focus`.

mcoker (changes requested):
This looks money! 

* One comment about some CSS you can probably remove
* We just met about this and decided we should support this with the `<textarea>`, too, and it seems to work like a charm already. Can you also add an example of a plain/readonly textarea? Updated the react issue to add support for that to `<TextArea>`, too - https://github.com/patternfly/patternfly-react/issues/7930
* Can you add a line in the docs for `.pf-m-plain`? Probably something like:

```
| `.pf-m-plain` | `textarea[readonly].pf-m-form-control`, `input[readonly].pf-m-form-control` | Modifies a `readonly` `<input>` or `<textarea>` to present as normal text. |
```

src/patternfly/components/FormControl/examples/FormControl.md
```diff
@@ -125,6 +127,10 @@ Standard
 Readonly
 {{/form-control}}
 <br><br>
+{{#> form-control controlType="textarea" form-control--modifier='pf-m-plain' form-control--attribute='readonly name="textarea-plain-readonly" id="textarea-plain-readonly" aria-label="Plain readonly textarea example"'}}
+Readonly
```
mcoker, line 131:
```suggestion
Readonly plain
```

mcoker (changes requested):
Just one small nit!

mcoker (approved):
🥸
👔👍

Issue #5056, opened 2022-08-26
Tabs - tab buttons need type="button", are currently type="submit"


mcoker:
re: https://github.com/patternfly/patternfly-react/issues/7887

PR #5060, opened 2022-08-27
fix(tabs): fixed button type for tab link (#5056)

gautamkrishnar:
Fixes issues with tabs when used with form
fixes: #5056

### DIFF
--- a/src/patternfly/components/Tabs/tabs-link.hbs
+++ b/src/patternfly/components/Tabs/tabs-link.hbs
@@ -1,4 +1,4 @@
-<{{#if tabs-link--isLink}}a{{else}}button{{/if}} class="pf-c-tabs__link
+<{{#if tabs-link--isLink}}a{{else}}button type="button"{{/if}} class="pf-c-tabs__link
   {{~#if tabs-link--IsAriaDisabled}}
     pf-m-aria-disabled
   {{/if}}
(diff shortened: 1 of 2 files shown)


mcoker (commented):
Perfect! Thanks @gautamkrishnar 👍👍

Just poking around, I found another spot that needs this update - the horizontal scroll buttons. Would you mind adding that to this PR? The code is here:

https://github.com/patternfly/patternfly/blob/e28b057d806a74695e92e27b083dd737f22c1739/src/patternfly/components/Tabs/tabs-scroll-button.hbs#L2-L16

We should be adding `tabs-scroll-button--attribute` after all of the other hard coded attributes though, so the update would probably look something like this:

```
  type="button"
  {{#if tabs-scroll-button--IsDisabled}} 
    disabled 
  {{/if}} 
  {{#if tabs-scroll-button--IsHidden}} 
    aria-hidden="true" 
  {{/if}} 
  {{#if tabs-scroll-button--IsLeft}} 
    aria-label="Scroll left" 
  {{/if}} 
  {{#if tabs-scroll-button--IsRight}} 
    aria-label="Scroll right" 
  {{/if}}
  {{#if tabs-scroll-button--attribute}} 
    {{{tabs-scroll-button--attribute}}} 
  {{/if}}> 
```

gautamkrishnar:
@mcoker done ✅

mcoker (approved):
Nice! 😎 Thanks @gautamkrishnar!!

PR #5066, opened 2022-08-31
fix(page): set max-height on full height page

srambach:
This adds `max-height: 100%;` to the `pf-m-full-height` page modifier to ensure that if a full height page is contained within another container, it does not grow larger than the available space.

To test, go to the [Banner Top/Bottom HTML demo](https://patternfly-pr-5066.surge.sh/components/banner/html-demos/topbottom/) and add `.pf-m-full-height` to the `pf-c-page` element.


mcoker (approved):
LPTM!

Issue #5067, opened 2022-08-31
Log viewer - allow customization of line number column based on number of lines


mcoker:
re: https://github.com/patternfly/patternfly-react/issues/7876

add CSS var support for a prop that allows you to specify the number of digits to show in the line number column.

PR #5068, opened 2022-09-01
feat(log-viewer): support for custom line number col size

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5067

@DaoDaoNoCode I created a new class for this so we don't change or break the existing line number column width/padding or the existing vars. If you think it's fine just to update, I can remove the need for the class.

Also do you see any issue with moving the font-family/size properties from the index + text elements up to the parent list?

### DIFF
--- a/src/patternfly/components/LogViewer/examples/LogViewer.md
+++ b/src/patternfly/components/LogViewer/examples/LogViewer.md
@@ -19,7 +19,15 @@ import './LogViewer.css';
 
 ### With line numbers
 ```hbs
-{{#> log-viewer log-viewer--id="log-viewer-line-number-example" log-viewer--HasLineNumbers="true" log-viewer--attribute='style="--pf-c-log-viewer__index--Width: 75px"' log-viewer--aria-label="Log viewer with line numbers"}}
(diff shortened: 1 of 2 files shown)


src/patternfly/components/LogViewer/examples/LogViewer.md
```diff
@@ -132,7 +140,9 @@ import './LogViewer.css';
 | `.pf-m-wrap-text` | `.pf-c-log-viewer` | Modifies the log viewer text to wrap. |
 | `.pf-m-nowrap` | `.pf-c-log-viewer` | Modifies the log viewer text to not wrap. |
 | `.pf-m-line-numbers` | `.pf-c-log-viewer` | Modifies the log viewer to display line numbers. |
+| `.pf-m-line-number-chars` | `.pf-c-log-viewer` | Modifies the log viewer allow for a custom line number column size. Use with  |
```
DaoDaoNoCode, line 143: Looks like the description is not finished

DaoDaoNoCode (approved):
@mcoker It LGTM! Left a small comment, thanks for the quick fix!

DaoDaoNoCode (approved):
LGTM (again)!

srambach (approved):
Nice way of preserving backward compatibility for the font family and size variables.

Issue #5013, opened 2022-07-29
Banner - add a new icon variant


mcarrano:
**Is this a new component or an extension of an existing one?**
Banner

**Describe the feature**
Want to add a new variant of the banner that allows for an icon in front of the text. This is an a11y enhancement that will provide an additional affordance besides color to allow color blind users to detect status. We should support the usual standard status icons along with the ability to provide a custom icon. 

Designs: https://marvelapp.com/prototype/17i95gf4/screens

mcoker:
@mcarrano we don't currently format the text in the banner in any way aside from adjust the text color based on the variant. We could support an icon in the layout, but then we would need to be a bit more opinionated about the layout than we are currently. This could also be a demo where we put the icon and text in a flex layout and use the flex layout to manage the spacing/alignment.

PR #5075, opened 2022-09-06
feat(banner): added support/docs for status icons and text

blfetche:
Applies to update feature for issue #5013

### DIFF
--- a/src/patternfly/components/Banner/examples/Banner.md
+++ b/src/patternfly/components/Banner/examples/Banner.md
@@ -13,26 +13,26 @@ cssPrefix: pf-c-banner
 
 <br>
 
-{{#> banner banner--modifier="pf-m-info"}}
-  Info banner 
+{{#> banner banner--modifier="pf-m-blue"}}
+  Blue banner 
 {{/banner}}
 
 <br>
 
-{{#> banner banner--modifier="pf-m-danger"}}
-  Danger banner
+{{#> banner banner--modifier="pf-m-red"}}
+  Red banner
 {{/banner}}
 
 <br>
 
-{{#> banner banner--modifier="pf-m-success"}}
-  Success banner
+{{#> banner banner--modifier="pf-m-green"}}
+  Green banner
 {{/banner}}
 
 <br>
 
-{{#> banner banner--modifier="pf-m-warning"}}
-  Warning banner
+{{#> banner banner--modifier="pf-m-gold"}}
+  Gold banner
 {{/banner}}
 ```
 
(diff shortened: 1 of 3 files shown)


blfetche:
@mcoker can you give this draft PR a look over? Would like to make sure I am approaching it accurately. I am wondering if mulitple HBS files makes sense or should I place the flex layout under banner.hbs with an if statement that looks for modifier pf-m-icon to enable it? Thanks

blfetche:
image

mcoker (changes requested):
Hey, this looks pretty good! Though we're not really opinionated on what goes into the banner, so this should just be a demo that creates the different color/status variations you've shown here, but just uses a basic flex layout and an icon to space the icon and text.

So you shouldn't need to make any sorts of updates to the banner component itself - no CSS or handlebars or anything, you would just add a demo (or demos?) and call it/them "Banner with status" - or maybe individual demos like "Banner with [success/warning/error/etc] status icon" or "[Success/Warning/Error/etc] banner".

@mcarrano wdyt about how to present these as demos? We could also add a single demo that presents like an example (just inline, not a full page thing) that shows "Banners with status" one after another like the existing [basic banner example](https://www.patternfly.org/v4/components/banner#basic), then incorporate one into a demo?

And similar to an alert or modal (with a status icon variation), you might also include some screen reader text like "Warning/Error/etc banner". Here's the alert as an example

Screen Shot 2022-09-06 at 5 55 09 PM

mcarrano:
> @mcarrano wdyt about how to present these as demos? We could also add a single demo that presents like an example (just inline, not a full page thing) that shows "Banners with status" one after another like the existing [basic banner example](https://www.patternfly.org/v4/components/banner#basic), then incorporate one into a demo?

I'm not sure. I like the way the current example looks as we have "Banner with links" and "Banner with icons" as examples of different things you can do. And then maybe leave the demo page to be full page examples like it is now that show placement and context?  Maybe we update the Basic demo to include the icon since that's a more accessible solution? @thatblindgeye what do you think? I believe you originated this issue.

mcoker:
>  I like the way the current example looks as we have "Banner with links" and "Banner with icons" as examples of different things you can do.

@mcarrano I agree, though the existing component examples show features of the banner - the banner with links exists since we style links uniquely in the banner because they appear differently than links in the page content. Technically the banner icon won't be a feature of the banner (like links are), since we decided just to leave it up to users to specify an icon, manage spacing, etc - which is why we were going to add it as a demo. Though that's a technicality on our end, and I'm not sure that anyone looking at our examples would care about that.

Just thinking out loud, we could:
* do as you've suggested, and add an icons example, then optionally add/update a demo with icons
* update the basic example just to include the default banner (no variation - uses the default styling), then add a new "variations" or "status" example that shows the success/error/etc status variations - and those include the icons and optional sr-only text (screen reader text that would say "Success banner" and announce that before the banner text)
* add new demos that add icons (and optional sr-only text) for each of the variations, maybe titled "Default status", "Success status", "Info status", etc...
* From my comment:
  * > We could also add a single demo that presents like an example (just inline, not a full page thing) that shows "Banners with status" one after another like the existing [basic banner example](https://www.patternfly.org/v4/components/banner#basic), then incorporate one into a demo?

I'm curious what @thatblindgeye thinks about that, too, and if we should add some sr-only text

thatblindgeye:
@mcarrano @mcoker based on Michael's comment above with some options, I think I would personally lean towards the option of "update the basic example just to include the default banner...". While users can ultimately add (or omit) whatever content from a banner, if a banner is being used to convey some sort of status/severity (especially if color is primarily being used to visually indicate status; I imagine users wouldn't be using variants _solely_ to change color?) it'd be better to provide examples that provide improved accessibility.

That would also require updating the "Banner with links" example, though since two banners in that example are already using the "default" variant that may not be a major issue.

The default banner variant is kind of the wild card I feel like, though, since that's a banner where an icon might be more optional... If we did go down the path of adding a "status/variant" example and updating the "basic" example, we could have two "default" variant banners with one having an icon I suppose:

Basic banner example with two default variants

I think sr-only text would work well here as well! Even if the text gets visually rendered due to the CSS not being applied, I don't think it would be super intrusive/confusing/verbose to users. Using the Alert example with CSS styles disabled as an example:

Alert without CSS

mcoker:
I'm not 100% on when/where a banner is used. I recall it being added for a pretty specific use case. @mcarrano to @thatblindgeye's question - do you know if the color variations are always used to indicate status? And if it's not necessarily a status, the grey/default (and/or maybe blue/info?) variation is used? Or could they be used just because the color pops or works more for the app/layout, not necessarily to indicate "success" or "danger" status if a green or red banner is used?

mcoker:
@mcarrano gotcha! in that case, seems like either adding an example or demo with [all of the banners stacked](https://github.com/patternfly/patternfly/pull/5075#issuecomment-1237614442), with documentation like "When a banner is used to convey status, you should [do all of these things...]", icons, and sr-only text called something like "Banner status" would be useful, then we could add a demo or update one of the existing ones to show it in context with the page. Though unless there is something unique about a full-page demo that we can't show in a stacked example/demo, I don't know that a full-page demo is necessary. wdyt? @thatblindgeye

mcoker:
@blfetche my apologies, I forgot to mention earlier - looks like this branch may have been created from your tree view PR. If you look through the commit history and files changed, you'll see changes from the tree view in this PR:

https://github.com/patternfly/patternfly/pull/5075/commits
https://github.com/patternfly/patternfly/pull/5075/files

blfetche:
Thanks for the catch @mcoker

mcoker:
> Mainly because right now, the HTML modifiers and React variant prop uses things like "danger" and "success", which make it seem like the intent is that these are mainly meant for status (though can be used solely for color).

@thatblindgeye yep! I was going to mention that we could rename the classes to just refer to the color, like the label, though that would be a breaking change. We could also add new color classes and leave the old legacy ones around until a breaking change, then remove them. That would clear up some of that confusion, though I dunno if it's that big of a deal.

It might look something like this

Screen Shot 2022-09-08 at 11 22 52 AM

mcarrano:
I agree that this would be better. We don't want people to confuse banners with alerts. The banner is not limited to notifying the user about error conditions so consumers might use color in a variety of ways.

mcoker:
@mcarrano should we go ahead do what's in [this comment](https://github.com/patternfly/patternfly/pull/5075#issuecomment-1240942480) and:

* update the current status modifier/prop names to reference colors instead (like the label)
  * we'll keep the existing status names around until a breaking change, at which point we'll remove them
* add a new example like in the screenshot above

mcarrano:
Yes, I think so. Do you agree @mcoker @blfetche ?

mcoker:
@mcarrano yep!

blfetche:
Sounds like a plan, thank you all for the input!

blfetche:
@mcoker @mcarrano @thatblindgeye 
Applied the above edits to reflect what was discussed. Please review when you can, thanks.

image

src/patternfly/components/Banner/examples/Banner.md
```diff
@@ -51,36 +51,109 @@ cssPrefix: pf-c-banner
...
   {{/button}}
 {{/banner}}
 ```
 
+### Status
+When a banner is used to convey status, it is advised to add an icon that also conveys status visually, and to provide screen reader text using ` .pf-screen-reader ` or an equivalent.
+
+```hbs
+{{#> banner}}
+  {{#> l-flex l-flex--modifier="pf-m-space-items-xs"}}
```
mcoker, line 80: Hmm... after looking at this in the PR, I think `sm` spacer would work better.

```suggestion
  {{#> l-flex l-flex--modifier="pf-m-space-items-sm"}}
```

Before:
Screen Shot 2022-09-09 at 5 54 55 PM


After:
Screen Shot 2022-09-09 at 5 54 46 PM

src/patternfly/components/Banner/examples/Banner.md
```diff
@@ -51,36 +51,109 @@ cssPrefix: pf-c-banner
...
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Warning banner
+    {{/l-flex-item}}
+  {{/l-flex}}
+{{/banner}}
+```
+
 ## Documentation
 Add a modifier class to the default banner to change the presentation: `.pf-m-info`, `.pf-m-danger`, `.pf-m-success`, or `.pf-m-warning`.
```
mcoker, line 149:
```suggestion
Add a modifier class to the default banner to change the presentation: `.pf-m-blue`, `.pf-m-red`, `.pf-m-green`, or `.pf-m-gold`.
```

mcoker (changes requested):
Awesome! Thanks for doing that so fast, too! 🚀

Just a couple of small comments.

thatblindgeye (commented):
This looks great! Awesome job on this. I think the only nitpick I have is that it could be worth adding a description to the "Basic" example, stating something along the lines of, "Banners can be styled with one of 5 different colors. A basic banner should only be used when the banner color does not represent status or severity." Mainly my worry is the code from the first example being grabbed without looking further down the page at the "Status" example.

@blfetche @mcoker what do you think? We do already have the description on the "Status" example which covers when you should use it, so adding one to the "Basic" example may be a bit redundant admittedly.

mcarrano (commented):
This looks great @blfetche . I agree with @thatblindgeye that it might be worth adding an additional note about not using the Basic configuration for anything reflecting status.

blfetche:
Thanks all for the feedback please see below, I have made the updates to my local environment. 

image

mcoker (approved):
🥳

thatblindgeye (approved):
🎉 Awesome job again on this!

mcarrano (approved):
Looks great!

PR #5076, opened 2022-09-06
fix(breakpoints): added breakpoint override

### LINKED ISSUE #4511: breakpoints are incorrect when page body is within an iframe
Breakpoints are based on the browser's width (and [similar to Bootstrap and other frameworks](https://ricostacruz.com/til/css-media-query-breakpoints)). Side navigation is common within PatternFly.

However, when the content of the page itself is in an iframe and the shell of the page is outside the iframe, with navigation on the side, the breakpoints are incorrect and the widgets in the page do not change when they should.

---

Here's a table which subtracts:

  | PF4 default | adjusted for sidebar outside of iframe
-- | -- | --
xs | 0 | 0
sm | 576 | 236
md | 768 | 428
lg | 992 | 652
xl | 1200 | 860
2xl | 1450 | 1110

So we've been using the following:

```scss
// Redefine grid breakpoints
$pf-global--breakpoint--xs: 0 !default;
$pf-global--breakpoint--sm: 236px !default;
$pf-global--breakpoint--md: 428px !default;
$pf-global--breakpoint--lg: 652px !default;
$pf-global--breakpoint--xl: 860px !default;
$pf-global--breakpoint--2xl: 1110px !default;
```

But it would be nice if PatternFly supported breakpoints for page content that is in frames instead, so we don't have to calculate it as a one-off hack.

---

Here's what it looks like visually...

The iframe is a complete document by itself. Everything outside (what we call the "shell", which contains navigation) isn't taken into account:

Screenshot from 2021-11-11 15-35-05

Here's the complete page and the iframe:

Screenshot from 2021-11-11 15-35-13

Sometimes (but rarely) our pages are even outside of the shell, so it would be wonderful if we could somehow set a variable to adjust this on the top of the document, like `.pf-m-breakpoint-minus-sidebar` or something similar. And, also noteworthy: mobile doesn't have side navigation.

As media queries do not support CSS custom properties, SASS variables need to be used. We could, however, have breakpoints wrapped inside the media query, like this pseudocode: 

```scss

@mixin breakpoint-md {
    /* Normal breakpoint code here */
}

@media screen and (max-width: $pf-global--breakpoint--md) {
  .pf-c-page:not(.pf-m-breakpoint-minus-sidebar) {
    @include breakpoint-md;
  }
}

@media screen and (max-width: $pf-global--breakpoint--md - $pf-c-page__sidebar--Width) {
  .pf-c-page.pf-m-breakpoint-minus-sidebar {
    @include breakpoint-md;
  }
}
```

This should be a way to include support for breakpoints without affecting current pages. (In other words, it would be an opt-in change only and not a breaking change.)

---

Related conversations at Cockpit: 

My original comment talking about the difference in Cockpit @ https://github.com/cockpit-project/cockpit/pull/13474#issuecomment-589045090 & https://github.com/cockpit-project/cockpit/pull/13474#issuecomment-591930416.

Also related: https://github.com/patternfly/patternfly-react/issues/3815

mattnolting:
closes #4511 

Added optional compiled base stylesheet and optional breakpoint variable override. 

Rather than rewrite all of our screen based media-queries, I've included a breakpoint override for `.scss` users and a compiled base stylesheet for `.css` only configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 When you install PatternFly 4, the package includes:
 
 - a single file for the entire compiled library: `node_modules/@patternfly/patternfly/patternfly.css`
+- a single file for the entire compiled library, with page based breakpoints: `node_modules/@patternfly/patternfly/patternfly-page-width-breakpoints.css`
 - individual files with each component compiled separately: `node_modules/@patternfly/patternfly/<ComponentName>/styles.css`
 - a single file for the entire library's source (SASS): `node_modules/@patternfly/patternfly/patternfly.scss`
 - individual files for each component's source (SASS): `node_modules/@patternfly/patternfly/<ComponentName>/styles.scss`
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 When you install PatternFly 4, the package includes:
 
 - a single file for the entire compiled library: `node_modules/@patternfly/patternfly/patternfly.css`
-- a single file for the entire compiled library, with page based breakpoints: `node_modules/@patternfly/patternfly/patternfly-page-width-breakpoints.css`
+- a single file for the entire compiled library, with page based breakpoints: `node_modules/@patternfly/patternfly/patternfly-full-page-breakpoints.css`
 - individual files with each component compiled separately: `node_modules/@patternfly/patternfly/<ComponentName>/styles.css`
 - a single file for the entire library's source (SASS): `node_modules/@patternfly/patternfly/patternfly.scss`
 - individual files for each component's source (SASS): `node_modules/@patternfly/patternfly/<ComponentName>/styles.scss`
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -71,7 +71,7 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 
   // Sidebar
   --pf-c-page__sidebar--ZIndex: var(--pf-global--ZIndex--sm);
-  --pf-c-page__sidebar--Width: #{pf-size-prem($pf-global--sidebar--Width)};
+  --pf-c-page__sidebar--Width: var(--pf-global--sidebar--Width);
   --pf-c-page__sidebar--BackgroundColor: var(--pf-global--BackgroundColor--dark-300);
   --pf-c-page__sidebar--m-light--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
   --pf-c-page__sidebar--m-light--BorderRightWidth: var(--pf-global--BorderWidth--sm);
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -71,7 +71,7 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 
   // Sidebar
   --pf-c-page__sidebar--ZIndex: var(--pf-global--ZIndex--sm);
-  --pf-c-page__sidebar--Width: #{pf-size-prem(290px)};
+  --pf-c-page__sidebar--Width: #{pf-size-prem($pf-global--sidebar--Width)};
   --pf-c-page__sidebar--BackgroundColor: var(--pf-global--BackgroundColor--dark-300);
   --pf-c-page__sidebar--m-light--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
   --pf-c-page__sidebar--m-light--BorderRightWidth: var(--pf-global--BorderWidth--sm);
--- a/src/patternfly/patternfly-base-breakpoint-minus-sidebar.scss
+++ b/src/patternfly/patternfly-base-breakpoint-minus-sidebar.scss
@@ -1,0 +1,8 @@
+$pf-global--exclude-sidebar-in-breakpoints: true;
+
+@import "./sass-utilities/all";
+@import "base/variables";
+@import "base/fonts";
+@import "base/base";
+@import "base/themes";
+@import "base/icons";
--- a/src/patternfly/patternfly-base-breakpoint-minus-sidebar.scss
+++ b/src/patternfly/patternfly-base-breakpoint-minus-sidebar.scss
@@ -1,8 +1,0 @@
-$pf-global--exclude-sidebar-in-breakpoints: true;
-
-@import "./sass-utilities/all";
-@import "base/variables";
-@import "base/fonts";
-@import "base/base";
-@import "base/themes";
-@import "base/icons";
--- a/src/patternfly/sass-utilities/scss-variables.scss
+++ b/src/patternfly/sass-utilities/scss-variables.scss
@@ -14,7 +14,7 @@ $pf-global--disable-fontawesome: false !default;
 $pf-global--load-pf-3: false !default;
 $pf-global--enable-shield: false !default;
 $pf-global--unset-root-font-size: true !default;
-$pf-global--page-width-breakpoints: false !default;
+$pf-global--full-page-breakpoints: false !default;
 
 // Patternfly uses the default root size of the browser, it assumes a 16px base
 // since most browser root size is 16. This variables is just to make the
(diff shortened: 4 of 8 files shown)


src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -167,6 +168,16 @@ $pf-global--breakpoint--lg: 992px !default;
 $pf-global--breakpoint--xl: 1200px !default;
 $pf-global--breakpoint--2xl: 1450px !default;
 
+// Redefine grid breakpoints
```
srambach, line 171: add "for breakpoint within iframe support"

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -14,6 +14,7 @@ $pf-global--disable-fontawesome: false !default;
 $pf-global--load-pf-3: false !default;
 $pf-global--enable-shield: false !default;
 $pf-global--unset-root-font-size: true !default;
+$pf-global--exclude-sidebar-in-breakpoints: false !default; // use this variable to enable page breakpoint minus sidebar width. this is for breakpoint within iframe support
```
srambach, line 17: Nit, can you add capitals and a period at the end?

srambach (commented):
Just some small wording things in comments. But also, these are magic numbers based on the width of the sidebar - is it worth somehow documenting here or at the sidebar that there's a relationship, so that if a redesign changes the width, there will be a trail to fix these numbers?

mcoker:
> But also, these are magic numbers based on the width of the sidebar - is it worth somehow documenting here or at the sidebar that there's a relationship, so that if a redesign changes the width, there will be a trail to fix these numbers?

@srambach good point. IMO the sidebar width might be a good case for a global variable, then we could reference that variable in the SCSS. Although (and I could be thinking of this wrong) - since the existing breakpoints are based off of the existing sidebar width, that seems to imply that without the current sidebar width, that's the default/ideal breakpoints for the components themselves. So if we updated the nav width, we would likely only want to update the existing breakpoints - does that sound right?

src/patternfly/patternfly-base-breakpoint-minus-sidebar.scss
```diff
@@ -0,0 +1,8 @@
+$pf-global--exclude-sidebar-in-breakpoints: true;
```
mcoker, line 1: What's the use case for this file? This builds `patternfly-base.css` with the var set, but this stylesheet doesn't include media queries. It's just the global vars and such - https://unpkg.com/browse/@patternfly/patternfly@4.213.0/patternfly-base.css

Also if you wanted to build patternfly-base, you could do it like this by just importing the existing `patternfly-base.scss` like https://github.com/patternfly/patternfly/blob/67698913b57c147dfd6d517e6b79fe13e7979b70/src/patternfly/patternfly-base-no-reset.scss

Or if you wanted to create a `patternfly.css` with these changes, you'd just import `patternfly.scss` - https://github.com/patternfly/patternfly/blob/67698913b57c147dfd6d517e6b79fe13e7979b70/src/patternfly/patternfly-no-reset.scss

mcoker (changes requested):
Nice and simple! Left a few comments.

Also there is a question of how this gets pulled into the react components, since I think most folks are just importing patternfly-base, then the react component is pulling in the CSS for each component.

Offering a `patternfly-[changed-breakpoints].css` as a whole stylesheet solution, and leaving the compilation of individual components with updated breakpoints up to the user is one option. I'm wondering if there are any issues we'll run into with react apps being able to use the styles. wdyt?

srambach:
> > But also, these are magic numbers based on the width of the sidebar - is it worth somehow documenting here or at the sidebar that there's a relationship, so that if a redesign changes the width, there will be a trail to fix these numbers?
> 
> @srambach good point. IMO the sidebar width might be a good case for a global variable, then we could reference that variable in the SCSS. Although (and I could be thinking of this wrong) - since the existing breakpoints are based off of the existing sidebar width, that seems to imply that without the current sidebar width, that's the default/ideal breakpoints for the components themselves. So if we updated the nav width, we would likely only want to update the existing breakpoints - does that sound right?

Yeah... true. 🤯 Anyway, there is a relationship there so it would be good to document it somehow.

srambach:
> This is good, but another way to name this could be like full width breakpoints or something, which doesn't tie this to a sidebar in the naming. It could be for a single page that doesn't have a sidebar or even something like single use of a component in some other design system.

+1 to that. I was trying to think of a simpler name

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -167,6 +168,16 @@ $pf-global--breakpoint--lg: 992px !default;
...
 $pf-global--breakpoint--2xl: 1450px !default;
 
+// Redefine grid breakpoints
+@if $pf-global--exclude-sidebar-in-breakpoints {
+  $pf-global--breakpoint--xs: 0;
+  $pf-global--breakpoint--sm: 236px;
+  $pf-global--breakpoint--md: 428px;
+  $pf-global--breakpoint--lg: 652px;
+  $pf-global--breakpoint--xl: 860px;
+  $pf-global--breakpoint--2xl: 1110px;
```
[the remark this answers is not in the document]
mattnolting: I attempted to create a slick mixin, but it wasn't playing well with sass variables and resetting them. I went with a function that will allow a user to offset any variable value and reset it, globally or locally.

src/patternfly/patternfly-base-breakpoint-minus-sidebar.scss
mattnolting: Updated

README.md
```diff
@@ -8,6 +8,7 @@
 When you install PatternFly 4, the package includes:
 
 - a single file for the entire compiled library: `node_modules/@patternfly/patternfly/patternfly.css`
+- a single file for the entire compiled library, with page based breakpoints: `node_modules/@patternfly/patternfly/patternfly-page-width-breakpoints.css`
```
mcoker, line 11: Out of scope, but I wonder if we should also include `patternfly-base.css` - that's required to use with the individual component CSS.

@srambach what do you think about "page width breakpoints" as a name? I think I referred to it as "full width" in my other comment.

src/patternfly/components/Page/page.scss
```diff
@@ -71,7 +71,7 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 
   // Sidebar
   --pf-c-page__sidebar--ZIndex: var(--pf-global--ZIndex--sm);
-  --pf-c-page__sidebar--Width: #{pf-size-prem(290px)};
+  --pf-c-page__sidebar--Width: #{pf-size-prem($pf-global--sidebar--Width)};
```
mcoker, line 74: I think a global var for this is probably a good thing for this reason - the sidebar width is probably a useful value to have outside of the context of the page component considering things like media queries or other SCSS vars may want to reference it. And this works, I have a few thoughts though, curious what you think @mattnolting @srambach.

* I wonder if it should be a CSS var, too, though I wouldn't want to introduce a problem/breaking change with this update, since this is such an integral part of the overall layout.
  * I think the way you've done it here is probably the least intrusive, but I'm thinking you could also add a CSS var without any risk of something breaking. That would be more consistent, and could be used for other uses where a CSS var can be referenced, but isn't within `.pf-c-page` - like the `:root` scope. wdyt?
* Since the current media queries are based off of the sidebar at a `rem` width (using `pf-size-prem()`), I'm wondering if the global var should have a `rem` value so the new media queries are based off of that rem value. Probably wouldn't make a difference, but I'm guessing might be more accurate if someone as modified `$pf-global--space-size-root`? https://github.com/patternfly/patternfly/blob/b6fdfb59baeeca10a5942c66c3ae208379a20a3f/src/patternfly/sass-utilities/functions.scss#L10-L12
  * Update this to update to `$pf-global--sidebar--Width: pf-size-prem(290px) !default;`? https://github.com/patternfly/patternfly/blob/4b338526c8349c2c224d3e951ca5815aa81269e9/src/patternfly/sass-utilities/scss-variables.scss#L157

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -151,6 +152,10 @@ $pf-global--spacer--form-element: pf-size-prem(6px) !default;
 $pf-global--gutter: $pf-global--spacer--md !default;
 $pf-global--gutter--md: $pf-global--spacer--lg !default;
 
+
+// Pade sidebar width
```
mcoker, line 156: nit

```suggestion
// Page sidebar width
```

README.md
srambach: I think I like full width better than page width.

src/patternfly/sass-utilities/scss-variables.scss
srambach, line 156: "Page" 😄  
✨ for adding this as an offset

mattnolting: 😆

README.md
mcoker: @mattnolting wdyt about the name?

mattnolting:
@mcoker @srambach Updated! Let's see if this works for you. I had to add a function to revert the sidebar `rem` value back to px value to offset for the `px` based breakpoints.

src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -332,3 +336,12 @@ $pf-global--spacer-map: (
   "3xl": "var(--pf-global--spacer--3xl)",
   "4xl": "var(--pf-global--spacer--4xl)"
 );
+
+// Modified grid breakpoints
+@if $pf-global--full-page-breakpoints {
+  $pf-global--breakpoint--sm: pf-offset-variable-value($pf-global--breakpoint--sm, pf-size-ppx($pf-global--sidebar--Width));
```
mcoker, line 342: I think you can use `calc()` here, too? Would this work? https://codepen.io/mcoker/pen/RwyGWay

```suggestion
  $pf-global--breakpoint--sm: calc(#{$pf-global--breakpoint--sm} - #{$pf-global--sidebar--Width});
```

mattnolting: I could, but you had mentioned creating a reusable mixin/function here earlier

README.md
```diff
@@ -8,6 +8,7 @@
 When you install PatternFly 4, the package includes:
 
 - a single file for the entire compiled library: `node_modules/@patternfly/patternfly/patternfly.css`
+- a single file for the entire compiled library, with page based breakpoints: `node_modules/@patternfly/patternfly/patternfly-full-page-breakpoints.css`
```
srambach, line 11: I know this is the name we agreed to, and it's fine - but here or somewhere can we document "use this version if you are embedding {the page? Patternfly components} in an iframe, or {any other use cases?}"

srambach (approved):
Just some documentation needed -  maybe wherever that "beta" tag goes :)

src/patternfly/sass-utilities/scss-variables.scss
mattnolting: @mcoker `calc` doesn't work bc the units differ `px` vs `em`

mattnolting:
Closing this in favor of https://github.com/patternfly/patternfly/issues/5111

PR #5077, opened 2022-09-06
fix(progress-stepper): fix width and long words wrap

### LINKED ISSUE #5024: Progress stepper - long strings in title/description don't wrap
These blocks should allow long strings to break, so the layout doesn't break. It would also be great if the layout didn't break this way if this does happen, too.

Screen Shot 2022-08-05 at 4 32 53 PM

Screen Shot 2022-08-05 at 4 36 49 PM

----

@mmenestr @mcarrano should the steps always be the same width/distance apart in a horizontal layout? I looked through the design issue (https://github.com/patternfly/patternfly-design/issues/879) and didn't see any mention of that, and didn't see any text wrapping/long text in the mockups. 

Here's an example:

Screen Shot 2022-08-05 at 4 49 30 PM

Or would we want it to look like this:

Screen Shot 2022-08-05 at 4 51 23 PM

srambach:
Fixes #5024 

This sets `min-width:0` so that the steps will maintain even widths, and allows wrapping of long words without spaces.

Screen shots of all examples with long text for convenience:
image

### DIFF
--- a/src/patternfly/components/ProgressStepper/progress-stepper.scss
+++ b/src/patternfly/components/ProgressStepper/progress-stepper.scss
@@ -125,8 +125,6 @@ $pf-c-progress-stepper--breakpoint-map: build-breakpoint-map();
   --pf-c-progress-stepper--m-center__step-main--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-progress-stepper--m-center__step-main--MarginLeft: var(--pf-global--spacer--xs);
   --pf-c-progress-stepper--m-center__step-main--TextAlign: center;
-  --pf-c-progress-stepper--m-center__step-title--TextAlign: auto;
-  --pf-c-progress-stepper--m-center__step-description--TextAlign: auto;
   --pf-c-progress-stepper--m-center__step-description--MarginRight: 0;
   --pf-c-progress-stepper--m-center__step-description--MarginLeft: 0;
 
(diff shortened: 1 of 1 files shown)


src/patternfly/components/ProgressStepper/progress-stepper.scss
```diff
@@ -124,8 +124,9 @@ $pf-c-progress-stepper--breakpoint-map: build-breakpoint-map();
   --pf-c-progress-stepper--m-center__step-connector--JustifyContent: center;
   --pf-c-progress-stepper--m-center__step-main--MarginRight: var(--pf-global--spacer--xs);
   --pf-c-progress-stepper--m-center__step-main--MarginLeft: var(--pf-global--spacer--xs);
-  --pf-c-progress-stepper--m-center__step-title--TextAlign: center;
-  --pf-c-progress-stepper--m-center__step-description--TextAlign: center;
+  --pf-c-progress-stepper--m-center__step-main--TextAlign: center;
+  --pf-c-progress-stepper--m-center__step-title--TextAlign: auto;
+  --pf-c-progress-stepper--m-center__step-description--TextAlign: auto;
```
mcoker, line 129: since the progress stepper is beta, I don't know that we need to keep the vars for title and description assuming they aren't needed/not used. you could always leave the var in the declaration with the default, and just remove these at the top level component selector scope.

src/patternfly/components/ProgressStepper/progress-stepper.scss
```diff
@@ -413,6 +419,7 @@ $pf-c-progress-stepper--breakpoint-map: build-breakpoint-map();
   font-size: var(--pf-c-progress-stepper__step-description--FontSize);
   color: var(--pf-c-progress-stepper__step-description--Color);
   text-align: var(--pf-c-progress-stepper__step-description--TextAlign);
+  overflow-wrap: anywhere;
```
mcoker, line 422: do you need this? or could it inherit [overflow-wrap from __main](https://github.com/patternfly/patternfly/blob/c44c23f9dc3dd755d048793621ca56b88d654fb8/src/patternfly/components/ProgressStepper/progress-stepper.scss#L374)? also assuming `break-word` will work, but I think `anywhere` was only needed on the title because the button is finicky. Or we could update `__main` to use `anywhere`?

mcoker (changes requested):

src/patternfly/components/ProgressStepper/progress-stepper.scss
srambach: 👍  `overflow-wrap:anywhere` seems to play nicely with soft breaks so I'm putting it on `__main`. I updated the screen shots and included a hyphen to show wrapping.

mcoker (approved):
awesome! 🤩

mattnolting (approved):
LGTM. We've used `overflow-wrap: break-word;` in our other components, but this seems to work just as well 👍

srambach:
> LGTM. We've used `overflow-wrap: break-word;` in our other components, but this seems to work just as well 👍

@mattnolting `overflow-wrap: anywhere` was needed in order for the helper text buttons to wrap.

Issue #4918, opened 2022-06-21
Wizard with drawer cleanup


mcoker:
Couple of follow ups to the demo

https://github.com/patternfly/patternfly/pull/4896#issuecomment-1162389578

in https://patternfly-pr-4896.surge.sh/components/wizard/html-demos/with-drawer-in-page/, the "open drawer" link should only be present when the drawer is closed.

I also noticed we limit the page width of the wizard section when it's in the page. This is what our current demos look like:

<img width="2331" alt="Screen Shot 2022-06-21 at 4 45 40 PM" src="https://user-images.githubusercontent.com/35148959/174902639-462d104f-6731-46b0-ab13-72a91573ffb9.png">

One solution could be to allow the wizard section to be full width, then limit the width of the form itself

<img width="2331" alt="Screen Shot 2022-06-21 at 4 46 30 PM" src="https://user-images.githubusercontent.com/35148959/174902729-c0d911c4-56f7-4081-a71c-72b3183918bf.png">

wdyt @mmenestr @mcarrano @mceledonia?

mmenestr:
I think that makes sense @mcoker (to allow the wizard section to be full width, but limit the width of the form itself )! And I would imagine that if there was a drawer in the wizard, the drawer would still be aligned with the right part of the window, and the drawer itself would stay the same width no matter how the window is resized. So as you resize the window, the form would be more covered or less covered by the drawer depending on the width of the window.

PR #5078, opened 2022-09-06
chore(wizard): clean up wizard demos

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4918

* adds docs for `.pf-m-limit-width` and `--pf-c-form--m-limit-width--MaxWidth` on wizard component page
* moves the limit-width modifier in wizard demos from the page section to the form
  * from this
    * Screen Shot 2022-09-06 at 5 39 40 PM
  * to this
    * Screen Shot 2022-09-06 at 5 39 17 PM
* Adds a `wizard--id` since the `id`'s and `<label for>`'s and hwhat-not in the wizard form template are based off of `wizard--id`, so they're currently not working like we want them to.
* Updates the "With drawer, in page" demo to collapse the drawer
* Adds a new "With drawer, in page, expanded" demo to expand the drawer and hides the "Open drawer" button


mattnolting (approved):
Updates LPTM 👍 Looks like AXE testing returns some errors on multiple main elements. Just something to note for a follow up issue.

PR #5083, opened 2022-09-11
refactor(tabs): updated demos

### LINKED ISSUE #5058: Tabs - rework Tab demos
Want to rework these demos to better clarify design intent. See https://marvelapp.com/prototype/9766ieg for requested changes. The Comment field includes text to add to help better explain what these demos are for.

mattnolting:
closes #5058

### DIFF
--- a/src/patternfly/demos/Page/page-template-title.hbs
+++ b/src/patternfly/demos/Page/page-template-title.hbs
@@ -11,7 +11,7 @@
       {{~#if page-template-title--description}}
         {{page-template-title--description}}
       {{else}}
-        Click any project card to view Tabs within Modals.
+        This is a full page demo.
       {{/if}}
     </p>
   {{/content}}
--- a/src/patternfly/demos/Page/page-template-title.hbs
+++ b/src/patternfly/demos/Page/page-template-title.hbs
@@ -1,6 +1,18 @@
 {{#> page-main-section page-main-section--IsLimitWidth="true" page-main-section--modifier=(concat 'pf-m-light ' page-template-title--modifier)}}
   {{#> content}}
-    <h1>Main title</h1>
-    <p>This is a full page demo.</p>
+    <h1>
+      {{~#if page-template-title--title}}
+        {{page-template-title--title}}
+      {{else}}
+        Main title
+      {{/if}}
+    </h1>
+    <p>
+      {{~#if page-template-title--description}}
+        {{page-template-title--description}}
+      {{else}}
+        Click any project card to view Tabs within Modals.
+      {{/if}}
+    </p>
   {{/content}}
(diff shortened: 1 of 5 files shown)


src/patternfly/demos/Page/page-template-title.hbs
```diff
@@ -1,6 +1,18 @@
...
+        {{page-template-title--title}}
+      {{else}}
+        Main title
+      {{/if}}
+    </h1>
+    <p>
+      {{~#if page-template-title--description}}
+        {{page-template-title--description}}
+      {{else}}
+        Click any project card to view Tabs within Modals.
```
srambach, line 14: This text is coming into other demos where it's not correct e.g. https://github.com/patternfly/patternfly/pull/5083/files#

srambach (commented):
Just the one thing where demo text is carried across demos.

src/patternfly/demos/Tabs/tabs--page-grid-view.hbs
```diff
@@ -10,7 +10,7 @@
         {{#> wrapper newcontext tabs--page-grid-view--id=(concat tabs--page-grid-view--id "-subtabs")}}
           {{#> l-flex l-flex--modifier="pf-m-column"}}
             {{#> l-flex-item}}
-              {{#> tabs tabs--IsSecondary="true" tabs--id=tabs--page-grid-view--id}}
+              {{#> tabs tabs--id=tabs--page-grid-view--id}}
```
mcoker, line 13: On the [nested tabs design](https://marvelapp.com/prototype/9766ieg/screen/88398077) the callout is to add a bottom border to the secondary tabs in the card, but this change makes them primary tabs instead - which adds a border, but also increases the font-size. Was it decided that those tabs should be primary tabs? Or should they remain secondary and we just add a border? cc @mcarrano @mmenestr

mmenestr:
A few comments:
- https://patternfly-pr-5083.surge.sh/components/tabs/html-demos/nested-tabs/: the tabs inside the card should be secondary (ie: the smaller font ones) 
- https://patternfly-pr-5083.surge.sh/components/tabs/html-demos/nested-unindented-tabs/: these tabs should also be secondary tabs, and the headers in the cards shouldn't be so bold
- https://patternfly-pr-5083.surge.sh/components/tabs/html-demos/tables-and-tabs/ row kebab menu being opened by default like that looks a little weird. Wouldn't show it. Otherwise looks great!!

Otherwise looks good, thanks for doing this!

mcoker:
@mmenestr updated, except for the card titles. I'd argue that if those are actually card titles, we use what the card provides by default since we try to use components as they are in our demos as much as possible. Unless there is another reason they shouldn't be bold?

PR #5088, opened 2022-09-13
fix(sidebar): fixed demos

### LINKED ISSUE #4908: Bug - Sidebar panel - sticky variant jumps on scroll within drawer
**Describe the problem**
When the sidebar panel is within a drawer, scrolling causes the panel to scroll a small amount before sticking. This can be seen in the [jump links in drawer HTML demo](https://www.patternfly.org/v4/components/jump-links/html-demos/jump-links-in-drawer/)

Comment https://github.com/patternfly/patternfly-react/pull/7520#discussion_r897265688 points out a couple of potential solutions.

**Expected behavior**
The sidebar panel, when inside a drawer component, doesn't scroll up when the page scrolls.

mattnolting:
closes #4908

### DIFF
--- a/src/patternfly/demos/Drawer/examples/drawer-jump-links.hbs
+++ b/src/patternfly/demos/Drawer/examples/drawer-jump-links.hbs
@@ -5,18 +5,22 @@
 {{#*inline "page-template-main-content"}}
   {{#> drawer drawer--id=(concat page-template--id '-drawer') jump-links-template--id=(concat page-template--id '-jump-links')}}
     {{#> drawer-main}}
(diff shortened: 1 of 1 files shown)


mcoker (approved):
lgtm! long term I think the component should support padding in the content and panel and probably multiple body elements, or we should generalize content sections like this so a "page section" and "[drawer/card/sidebar] body" elements are all common/shared and can be used anywhere that kind of box or list of boxes is needed. Created an issue for that here - https://github.com/patternfly/patternfly/issues/5131

PR #5093, opened 2022-09-16
feat(page): added multiple body support

### LINKED ISSUE #4982: Page sidebar - better support for custom content
From a request on slack:

> Hello!  I have an app following the typical Patternfly layout, with a PageSidebar theme="dark" nav={Navigation} where Navigation contains the usual Nav and NavList.  What I'd like to do is put a copyright notice down at the bottom of the sidebar.  The problem is, if I put children inside the PageSidebar they are ignored, and if I put them in Navigation they are already inside a div (or some other element) that only has the height of the menu.  Does anyone know a good way to do this?

I created this sandbox https://codesandbox.io/s/festive-haibt-81j4cn?file=/index.js

The page sidebar isn't currently setup to accept custom content without a fair amount of modifications. Should we support more types of custom content/components in the sidebar?

mattnolting:
closes #4982

### DIFF
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -25,7 +25,7 @@ import './Page.css'
       header-tools
     {{/page-header-tools}}
   {{/page-header}}
-  {{#> page-sidebar}}
+  {{#> page-sidebar page-sidebar-body--modifier="pf-m-nav"}}
     Navigation
   {{/page-sidebar}}
   {{#> page-main}}
@@ -77,7 +77,7 @@ import './Page.css'
 {{/page}}
 ```
 
-### Using flex layout
+### Multiple sidebar body elements, padding, and fill
 ```hbs
 {{#> page}}
   {{#> page-header}}
@@ -94,66 +94,13 @@ import './Page.css'
     {{/page-header-tools}}
   {{/page-header}}
   {{#> page-sidebar page-sidebar--ExcludeSidebarBody="true"}}
-    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-fill"}}
-      {{#> l-flex l-flex--modifier="pf-m-column pf-u-h-100"}}
-        {{#> l-flex-item}}
-          pf-c-nav
-        {{/l-flex-item}}
-        {{#> l-flex-item}}
-          other content
-        {{/l-flex-item}}
-        {{#> l-flex-item l-flex-item--modifier="pf-m-column pf-u-mt-auto"}}
-            footer content
-        {{/l-flex-item}}
-      {{/l-flex}}
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-nav"}}
+      Navigation
     {{/page-sidebar-body}}
-  {{/page-sidebar}}
-  {{#> page-main}}
-    {{#> page-main-section page-main-section--modifier="pf-m-light"}}
-    {{/page-main-section}}
-  {{/page-main}}
-{{/page}}
-```
-
-### Multiple sidebar body elements and page insets
-```hbs
-{{#> page}}
-  {{#> page-header}}
-    {{#> page-header-brand}}
-      {{#> page-header-brand-toggle}}
-        toggle
-      {{/page-header-brand-toggle}}
-      {{#> page-header-brand-link page-header-brand-link--href="#"}}
-        Logo
-      {{/page-header-brand-link}}
-    {{/page-header-brand}}
-    {{#> page-header-tools}}
-      header-tools
-    {{/page-header-tools}}
-  {{/page-header}}
-  {{#> page-sidebar page-sidebar--ExcludeSidebarBody="true"}}
-    {{#> page-sidebar-body page-sidebar-body--modifier=""}}
-      pf-c-nav
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-fill pf-m-inset"}}
+      inset content
     {{/page-sidebar-body}}
-    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-fill"}}
-      {{#> l-flex l-flex--modifier="pf-m-column"}}
-        {{#> l-flex-item}}
-          other content
-        {{/l-flex-item}}
-        {{#> l-flex-item}}
-          footer content
-        {{/l-flex-item}}
-      {{/l-flex}}
-    {{/page-sidebar-body}}
-  {{/page-sidebar}}
-  {{#> page-sidebar page-sidebar--ExcludeSidebarBody="true"}}
-    {{#> page-sidebar-body}}
-      pf-c-nav
-    {{/page-sidebar-body}}
-    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-page-insets"}}
-      other content
-    {{/page-sidebar-body}}
-    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-align-bottom pf-m-page-insets"}}
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-inset"}}
       footer content
     {{/page-sidebar-body}}
   {{/page-sidebar}}
@@ -213,7 +160,7 @@ import './Page.css'
       header-tools
     {{/page-header-tools}}
   {{/page-header}}
-  {{#> page-sidebar}}
+  {{#> page-sidebar page-sidebar-body--modifier="pf-m-nav"}}
     Navigation
   {{/page-sidebar}}
   {{#> page-main}}
@@ -246,7 +193,7 @@ import './Page.css'
       header-tools
     {{/page-header-tools}}
   {{/page-header}}
-  {{#> page-sidebar}}
+  {{#> page-sidebar page-sidebar-body--modifier="pf-m-nav"}}
     Navigation
   {{/page-sidebar}}
   {{#> page-main}}
@@ -348,22 +295,24 @@ This component provides the basic chrome for a page, including sidebar, header, 
 | `.pf-m-selected` | `.pf-c-page__header-tools-item` | Modifies a header tools item to indicate that the button inside is in the selected state. |
 | `.pf-m-expanded` | `.pf-c-page__sidebar` |  Modifies the sidebar for the expanded state. |
 | `.pf-m-collapsed` | `.pf-c-page__sidebar` |  Modifies the sidebar for the collapsed state. |
-| `.pf-m-page-insets` | `.pf-c-page__sidebar-body` | Modifies a sidebar body padding/inset to visually match padding of page elements. |
+| `.pf-m-inset` | `.pf-c-page__sidebar-body` | Modifies a sidebar body left/right inset to visually match padding of page elements. |
+| `.pf-m-no-inset` | `.pf-c-page__sidebar-body` | Removes a sidebar body left/right inset. |
+| `.pf-m-padding` | `.pf-c-page__sidebar-body` | Modifies a sidebar body top/bottom padding. |
+| `.pf-m-no-padding` | `.pf-c-page__sidebar-body` | Removes a sidebar body top/bottom padding. |
+| `.pf-m-nav` | `.pf-c-page__sidebar-body` | Modifies a sidebar body top/bottom padding to accomodate a navigation component. |
 | `.pf-m-light` | `.pf-c-page__sidebar` |  Modifies the sidebar the light variation. **Note: for use with a light themed nav component** |
 | `.pf-m-light` | `.pf-c-page__main-section` | Modifies a main page section to have a light theme. |
 | `.pf-m-dark-200` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a dark transparent background. |
 | `.pf-m-dark-100` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a darker transparent background. |
 | `.pf-m-light-200` | `.pf-c-page__main-wizard` | Modifies a wizard page section to have a light 200 theme. |
-| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section`, `.pf-c-page__sidebar-body` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
-| `.pf-m-no-padding` | `.pf-c-page__sidebar-body` | Removes padding from the main page section body. |
 | `.pf-m-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Modifies the main page section to add padding back in at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). Should be used with pf-m-no-padding. |
-| `.pf-m-fill` | `.pf-c-page__main-section`, `.pf-c-page__sidebar-body` | Modifies a main page section to grow to fill the available vertical space. |
+| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
+| `.pf-m-fill` | `.pf-c-page__main-section`, `.pf-c-page__sidebar-body` | Modifies the element to grow to fill the available space. |
 | `.pf-m-no-fill` | `.pf-c-page__main-section` | Modifies a main page section to not grow to fill the available vertical space. |
 | `.pf-m-hidden{-on-[breakpoint]}` | `.pf-c-page__header-tools-group`, `.pf-c-page__header-tools-item` | Hides a header tools group or item at an optional breakpoint, or hides it at all [breakpoints](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes) with `.pf-m-hidden`. |
 | `.pf-m-visible{-on-[breakpoint]}` | `.pf-c-page__header-tools-group`, `.pf-c-page__header-tools-item` | Shows a header tools group or item at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-limit-width` | `.pf-c-page__main-section` | Modifies a page section to limit the `max-width` of the content inside. |
 | `.pf-m-align-center` | `.pf-c-page__main-section.pf-m-limit-width` | Modifies a page section body to align center. |
-| `.pf-m-align-bottom` | `.pf-c-page__sidebar-body` | Modifies a page section body to align bottom.
 | `.pf-m-sticky-top{-on-[breakpoint]-height}` | `.pf-c-page__main-*` | Modifies a section/group to be sticky to the top of its container at an optional height breakpoint. |
 | `.pf-m-sticky-bottom{-on-[breakpoint]-height}` | `.pf-c-page__main-*` | Modifies a section/group to be sticky to the bottom of its container at an optional height breakpoint. |
 | `.pf-m-shadow-bottom` | `.pf-c-page__main-*` | Modifies a section/group to have a bottom shadow. |
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -95,13 +95,15 @@ import './Page.css'
   {{/page-header}}
   {{#> page-sidebar page-sidebar--ExcludeSidebarBody="true"}}
     {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-fill"}}
-      pf-c-nav
-      {{#> l-flex l-flex--modifier="pf-m-column"}}
+      {{#> l-flex l-flex--modifier="pf-m-column pf-u-h-100"}}
+        {{#> l-flex-item}}
+          pf-c-nav
+        {{/l-flex-item}}
         {{#> l-flex-item}}
           other content
         {{/l-flex-item}}
-        {{#> l-flex-item}}
-          footer content
+        {{#> l-flex-item l-flex-item--modifier="pf-m-column pf-u-mt-auto"}}
+            footer content
         {{/l-flex-item}}
       {{/l-flex}}
     {{/page-sidebar-body}}
@@ -113,7 +115,7 @@ import './Page.css'
 {{/page}}
 ```
 
-### Multiple sidebar body elements
+### Multiple sidebar body elements and page insets
 ```hbs
 {{#> page}}
   {{#> page-header}}
@@ -142,6 +144,17 @@ import './Page.css'
           footer content
         {{/l-flex-item}}
       {{/l-flex}}
+    {{/page-sidebar-body}}
+  {{/page-sidebar}}
+  {{#> page-sidebar page-sidebar--ExcludeSidebarBody="true"}}
+    {{#> page-sidebar-body}}
+      pf-c-nav
+    {{/page-sidebar-body}}
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-page-insets"}}
+      other content
+    {{/page-sidebar-body}}
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-align-bottom pf-m-page-insets"}}
+      footer content
     {{/page-sidebar-body}}
   {{/page-sidebar}}
   {{#> page-main}}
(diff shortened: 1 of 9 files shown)


src/patternfly/demos/Page/examples/Page.md
```diff
@@ -17,6 +17,59 @@ Using the `.pf-m-full-height` modifier class on the page component eliminates th
 {{> page-template page-template--id="page-demo-full-height" page-template--modifier="pf-m-full-height"}}
 ```
 
+### Multiple body elements
```
srambach, line 20: I'd expect to see multiple `pf-c-page__main-body` with this description. Should it be "Multiple sidebar body elements" ? Are multiple body elements valid, and if so should we show them?

src/patternfly/components/Page/examples/Page.md
```diff
@@ -71,6 +73,37 @@ wrapperTag: div
 {{/page}}
 ```
 
+### Multiple body elements
```
srambach, line 76: I think it would be clearer to name this "Multiple sidebar body elements"

mattnolting:
> This satisfies the immediate request, but I wonder if we could just open it up so that someone can put in whatever kind of layout they want (and let flex or whatever align things), and use sidebar bodies as children if they want, since we already have layouts that will do this.

@mcoker The flex example illustrates the nuance involved in using flex to provide basic padding/spacing provided by page insets, which isn't something the flex layout does. As you can see, it's def doable, but pretty inefficient. I think providing multiple `body` elements (with added page-inset mod) in conjunction with layouts is the way to go.

mattnolting:
@mcoker rebased and ready

src/patternfly/components/Page/page.scss
```diff
@@ -453,6 +475,15 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
       width: 100%;
     }
   }
+
+  &.pf-m-no-padding {
+    --pf-c-page__sidebar-body--PaddingTop: 0;
+    --pf-c-page__sidebar-body--PaddingBottom: 0;
+  }
+
+  &.pf-m-align-bottom {
```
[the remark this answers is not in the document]
mattnolting: @mcoker see following comment

src/patternfly/components/Page/page.scss
```diff
@@ -386,13 +395,16 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 
 // Sidebar
 .pf-c-page__sidebar {
+  z-index: var(--pf-c-page__sidebar--ZIndex);
+  display: flex;
+  flex-direction: column;
```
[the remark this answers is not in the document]
mattnolting: Moved to breaking change release

srambach (approved):
👍

src/patternfly/sass-utilities/placeholders.scss
```diff
@@ -6,6 +7,10 @@
   --pf-global--link--Color: var(--pf-global--link--Color--dark);
   --pf-global--link--Color--hover: var(--pf-global--link--Color--dark--hover);
   --pf-global--BackgroundColor--100: var(--pf-global--BackgroundColor--light-100);
+
+  .pf-c-divider {
+    --pf-c-divider--after--BackgroundColor: var(--pf-global--BorderColor--100);
```
mcoker, line 12: I'd love to remove all of the component specific stuff from these blocks if we can. This is going to change the color on any divider in a `pf-t-dark` block, but leave `BorderColor--100` alone. Dividers in the new dark theme are closer to this, but that comes from `--pf-global--BorderColor--100`, because (as far as I know from design) dividers are a single line border using the border color. 

So using `.pf-t-dark` we'll get something like this in tabs, where a divider wouldn't match the tabs bottom border

Screenshot 2023-01-17 at 7 05 10 PM

This is what it looks like currently

Screenshot 2023-01-17 at 7 04 27 PM

and fwiw, this is what it looks like in the new dark theme

Screenshot 2023-01-17 at 7 08 19 PM

I guess it's worth considering if we should update `--pf-global--BorderColor--100` here instead of creating an exception.

src/patternfly/components/Page/examples/Page.md
```diff
@@ -77,6 +77,93 @@ import './Page.css'
 {{/page}}
 ```
 
+### Using flex layout
```
mcoker, line 80: do we need this example? it's more of a demo, and you have a demo for it, too

src/patternfly/components/Page/examples/Page.md
```diff
@@ -77,6 +77,93 @@ import './Page.css'
...
+    {{/page-sidebar-body}}
+  {{/page-sidebar}}
+  {{#> page-main}}
+    {{#> page-main-section page-main-section--modifier="pf-m-light"}}
+    {{/page-main-section}}
+  {{/page-main}}
+{{/page}}
+```
+
+### Multiple sidebar body elements and page insets
```
mcoker, line 118: same here

src/patternfly/components/Page/examples/Page.md
```diff
@@ -261,19 +348,22 @@ This component provides the basic chrome for a page, including sidebar, header,
...
 | `.pf-m-light` | `.pf-c-page__sidebar` |  Modifies the sidebar the light variation. **Note: for use with a light themed nav component** |
 | `.pf-m-light` | `.pf-c-page__main-section` | Modifies a main page section to have a light theme. |
 | `.pf-m-dark-200` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a dark transparent background. |
 | `.pf-m-dark-100` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a darker transparent background. |
 | `.pf-m-light-200` | `.pf-c-page__main-wizard` | Modifies a wizard page section to have a light 200 theme. |
-| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
+| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section`, `.pf-c-page__sidebar-body` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
```
mcoker, line 357:
```suggestion
| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
```

src/patternfly/components/Page/examples/Page.md
```diff
@@ -261,19 +348,22 @@ This component provides the basic chrome for a page, including sidebar, header,
...
+| `.pf-m-no-padding` | `.pf-c-page__sidebar-body` | Removes padding from the main page section body. |
 | `.pf-m-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Modifies the main page section to add padding back in at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). Should be used with pf-m-no-padding. |
-| `.pf-m-fill` | `.pf-c-page__main-section` | Modifies a main page section to grow to fill the available vertical space. |
+| `.pf-m-fill` | `.pf-c-page__main-section`, `.pf-c-page__sidebar-body` | Modifies a main page section to grow to fill the available vertical space. |
```
mcoker, line 360: just updated to remove the element specific wording. Or we could add something in about the sidebar, too.

```suggestion
| `.pf-m-fill` | `.pf-c-page__main-section`, `.pf-c-page__sidebar-body` | Modifies the element to grow to fill the available space. |
```

src/patternfly/components/Page/page.scss
```diff
@@ -76,6 +78,8 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
   --pf-c-page__sidebar--m-light--BackgroundColor: var(--pf-global--BackgroundColor--light-100);
   --pf-c-page__sidebar--m-light--BorderRightWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-page__sidebar--m-light--BorderRightColor: var(--pf-global--BorderColor--100);
+  --pf-c-page__sidebar--m-light--c-divider--after--BackgroundColor: var(--pf-global--BorderColor--300);
```
mcoker, line 81: @mcarrano @mceledonia @mmenestr - do we need to continue to support the light theme vertical nav with v5 or is that something we could drop? https://www.patternfly.org/v4/components/navigation/html#default-in-light-mode

src/patternfly/components/Page/page.scss
```diff
@@ -91,9 +95,14 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
   --pf-c-page__sidebar-body--m-menu--BorderTopColor: var(--pf-global--BackgroundColor--dark-200);
   --pf-c-page__sidebar-body--m-menu--BorderTopWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-page__sidebar-body--m-menu--c-context-selector--BorderBottomColor: var(--pf-global--BackgroundColor--dark-200);
+  --pf-c-page__sidebar-body--m-page-insets--inset: var(--pf-global--spacer--md); // make this the default inset at breaking change
+  --pf-c-page__sidebar-body--m-page-insets--xl--inset: var(--pf-global--spacer--lg); // make this the default inset at breaking change
```
mcoker, line 99: what are the breaking change comments for?

src/patternfly/demos/Page/examples/Page.md
```diff
@@ -17,6 +17,58 @@ Using the `.pf-m-full-height` modifier class on the page component eliminates th
 {{> page-template page-template--id="page-demo-full-height" page-template--modifier="pf-m-full-height"}}
 ```
 
+### Using flex layout instead of multiple body elements
```
mcoker, line 20: If you want to merge this demo (assuming it isn't just for the PR to test), we should update it so it's clear this applies to the sidebar. I'm not sure that we need it as long as we open up the body elements to take any children.

src/patternfly/demos/Page/examples/Page.md
```diff
@@ -17,6 +17,58 @@ Using the `.pf-m-full-height` modifier class on the page component eliminates th
...
+        {{/l-flex-item}}
+      {{/l-flex}}
+    {{/page-sidebar-body}}
+  {{/page-sidebar}}
+{{/inline}}
+```
+
+### Multiple sidebar body elements
+```hbs isFullscreen
+{{> page-template page-template--id="multiple-body-elements-demo"}}
```
mcoker, line 45: the demo above matches this demo's existing id, FYI

```suggestion
{{> page-template page-template--id="multiple-sidebar-body-elements-demo"}}
```

src/patternfly/demos/Page/examples/Page.md
```diff
@@ -17,6 +17,58 @@ Using the `.pf-m-full-height` modifier class on the page component eliminates th
...
+    {{> divider}}
+    {{#> page-sidebar-body}}
+      {{> page-template-sidebar-nav}}
+    {{/page-sidebar-body}}
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-page-insets"}}
+      {{#> content}}
+        <p>Custom sidebar content</p>
+      {{/content}}
+    {{/page-sidebar-body}}
+    {{#> page-sidebar-body page-sidebar-body--modifier="pf-m-align-bottom pf-m-page-insets"}}
```
mcoker, line 63: I'm still not sure why we need an `align-bottom` class if body elements can fill, and you can put whatever you want in a body element. Also I wonder if these elements should act like page sections - the only/last child will fill, and we have fill/no-fill support - that's how you'd do a footer in the page main content area, which is already pretty familiar using the page and cards.

mcoker (commented):
looking good! left some comments.

src/patternfly/components/Page/examples/Page.md
mattnolting: No, it's demonstrative of the alternate integration method that is the user generating their own layout

src/patternfly/components/Page/examples/Page.md
```diff
@@ -261,14 +295,19 @@ This component provides the basic chrome for a page, including sidebar, header,
...
 | `.pf-m-dark-100` | `.pf-c-page__main-section` |  Modifies a main page section to have a dark theme and a darker transparent background. |
 | `.pf-m-light-200` | `.pf-c-page__main-wizard` | Modifies a wizard page section to have a light 200 theme. |
-| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Modifies the main page section to add padding back in at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). Should be used with pf-m-no-padding. |
-| `.pf-m-fill` | `.pf-c-page__main-section` | Modifies a main page section to grow to fill the available vertical space. |
+| `.pf-m-no-padding`, `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
```
mcoker, line 309:
```suggestion
| `.pf-m-no-padding{-on-[breakpoint]}` | `.pf-c-page__main-section` | Removes padding from the main page section at an optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
```

src/patternfly/components/Page/page.scss
```diff
@@ -421,8 +431,11 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
     @include pf-t-light;
 
     --pf-c-page__sidebar--BackgroundColor: var(--pf-c-page__sidebar--m-light--BackgroundColor);
+  }
 
-    border-right: var(--pf-c-page__sidebar--m-light--BorderRightWidth) solid var(--pf-c-page__sidebar--m-light--BorderRightColor);
+  .pf-m-page-insets {
+    padding-right: var(--pf-c-page__sidebar-body--m-page-insets--PaddingRight);
+    padding-left: var(--pf-c-page__sidebar-body--m-page-insets--PaddingLeft);
   }
```
mcoker, line 439: Looks like you can remove this. I don't see var support for it.

src/patternfly/components/Page/page.scss
```diff
@@ -453,6 +466,35 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
...
+
+  &.pf-m-nav {
+    --pf-c-page__sidebar-body--PaddingTop: var(--pf-c-page__sidebar-body--m-nav--PaddingTop);
+    --pf-c-page__sidebar-body--PaddingBottom: var(--pf-c-page__sidebar-body--m-nav--PaddingBottom);
+  }
+
+  &.pf-m-inset {
+    padding-right: var(--pf-c-page__sidebar-body--m-inset--PaddingRight);
+    padding-left: var(--pf-c-page__sidebar-body--m-inset--PaddingLeft);
```
mcoker, line 477: should we add padding vars to `__body` so we can redeclare those - like top/bottom padding?

src/patternfly/components/Page/page.scss
```diff
@@ -453,6 +469,19 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
       width: 100%;
     }
   }
+
+  &.pf-m-no-padding {
```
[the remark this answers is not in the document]
mcoker: still have this question ^

src/patternfly/components/Page/page.scss
```diff
@@ -453,6 +466,35 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
...
+    padding-right: 0;
+    padding-left: 0;
+  }
+
+  &.pf-m-no-padding {
+    padding-top: 0;
+    padding-bottom: 0;
+  }
+
+  &.pf-m-padding {
```
mcoker, line 490: if this is the default, what's the use case for this class? I would probably think a sidebar body element has no padding by default from the way you've set everything up here.

src/patternfly/components/Page/page.scss
```diff
@@ -453,6 +466,35 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
...
+    padding-top: var(--pf-c-page__sidebar-body--PaddingTop);
+    padding-bottom: var(--pf-c-page__sidebar-body--PaddingBottom);
+  }
+
+  &.pf-m-fill {
```
mcoker, line 495: wdyt about setting last-child to grow by default, and adding fill/no-fill - like page sections?

mcoker (commented):
This looks good - left some comments, curious your feedback. And I like the structure, but I'll just mention the primary use case for the sidebar is for vertical nav, but the sidebar body element needs a modifier to provide the nav padding. It's fine that way, I just wonder if there is a way for the sidebar body to work for the nav by default, without being confusing with the different types of padding. I'm also wondering if we need that specific padding at the top of the sidebar for nav - can we either remove it or leave it there and use it for everything? IMO it looks like we might be able to remove it?

The context selector is a good example of when we do not want that top padding. wdyt @mceledonia @mmenestr 

Screenshot 2023-02-14 at 1 30 21 PM
Screenshot 2023-02-14 at 1 30 14 PM

Here it is with the nav top padding removed

Screenshot 2023-02-14 at 1 32 47 PM
Screenshot 2023-02-14 at 1 32 36 PM
Screenshot 2023-02-14 at 1 32 07 PM
Screenshot 2023-02-14 at 1 32 04 PM

src/patternfly/components/Page/page.scss
mattnolting: Just the inverse if padding needs to be reapplied

src/patternfly/components/Page/page.scss
```diff
@@ -453,6 +462,43 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
...
+
+  &.pf-m-no-vertical-padding {
+    --pf-c-page__sidebar-body--PaddingTop: 0;
+    --pf-c-page__sidebar-body--PaddingBottom: 0;
+  }
+
+  &.pf-m-vertical-padding {
+    padding-top: var(--pf-c-page__sidebar-body--PaddingTop);
+    padding-bottom: var(--pf-c-page__sidebar-body--PaddingBottom);
+  }
```
mcoker, line 489: I imagine these should probably have their own variables?

mcoker (changes requested):
After mentioning to @mceledonia that the top padding on the sidebar was making this trickier than it should be, he mentioned it's just a spacer similar to the top/bottom padding on a menu. In the menu component, that padding is on the menu list - it's the list of items that has spacing above/below it. I'm thinking since that spacing is specific to the nav component, it seems like we could move that spacing to the nav list the same way. Default sidebar body elements could have no padding, and we have modifiers for padding, if that's needed. Personally I think we could add support for insets and end there, since that seems common enough to want to align with the content in a sidebar. If a user wants to add vertical padding, they could do that via gaps/spacers in the layout they use to stack things on top of one another. Then we can remove the `pf-m-nav` modifier, too.

Just glancing at the nav component, it seems like we could remove that nav-specific spacing from nav body, then:
* Add the padding-top and padding-bottom to `.pf-c-nav__list`
  * Remove it from `.pf-c-nav__subnav` (expandable nav)
* Sort out this `margin-top` on `.pf-c-nav__item`. It only applies to grouped items in `.pf-c-nav__section` and is creating space between nav items, and the gap above the items, which should now be handled by the padding on `.pf-c-nav__list`. That could be a flex column with a gap or easiest might be to update that rule to be `&:not(:first-child) {}`? https://github.com/patternfly/patternfly/blob/b51b987d9d82d4e80f278c8f75c6d2df96c916f2/src/patternfly/components/Nav/nav.scss#L737

The only visual change that makes as far as I can tell is it reduces the space between a group title and the top of the sidebar, when the title is the first thing in the sidebar. But IMO it looks more intentional this way? wdyt @mceledonia @mmenestr @mcarrano 

With this update to remove the extra space at the top of the sidebar
Screenshot 2023-03-01 at 7 53 36 PM
Screenshot 2023-03-01 at 7 53 40 PM

What it looks like now on v4
Screenshot 2023-03-01 at 7 54 02 PM
Screenshot 2023-03-01 at 7 54 02 PM

mcarrano:
> The only visual change that makes as far as I can tell is it reduces the space between a group title and the top of the sidebar, when the title is the first thing in the sidebar. But IMO it looks more intentional this way? wdyt @mceledonia @mmenestr @mcarrano

I'm OK with this if others are.

mceledonia:
I'm good with this as well!

src/patternfly/components/Page/page.scss
```diff
@@ -1,9 +1,14 @@
...
 $pf-c-page--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl", "2xl");
 $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl", "2xl");
+// stylelint-enable
 
 // URL.com/guidelines#layout
 .pf-c-page {
   --pf-c-page--BackgroundColor: var(--pf-global--BackgroundColor--light-300);
+  --pf-c-page--inset: 0;
+  --pf-c-page--m-page-insets--inset: var(--pf-global--spacer--md);
+  --pf-c-page--m-page-insets--xl--inset: var(--pf-global--spacer--lg);
```
mcoker, line 11: I like that we have a higher order var for this, as it lets us use this other places - page sections, masthead, etc. But I don't know if this is entirely accurate. The page does have an inset without `.pf-m-page-insets`.

As you have it now, you'll need a class on a thing that doesn't use page insets to add the page inset. I think it might make more sense to be something like this - wdyt?

```
--pf-c-page--inset: var(--pf-global--spacer--md);
--pf-c-page--xl--inset: var(--pf-global--spacer--lg);

--pf-c-page__[thing]--m-page-insets--PaddingLeft: var(--pf-c-page--inset);
--pf-c-page__[thing]--m-page-insets--PaddingRight: var(--pf-c-page--inset);

@media (min-width: [xl breakpoint]) {
  --pf-c-page--inset: var(--pf-c-page--xl--inset);
}

.pf-c-page__[thing] {
  &.pf-m-page-insets {
    --pf-c-page__[thing]--PaddingLeft: var(--pf-c-page__[thing]--m-page-insets--PaddingLeft);
    --pf-c-page__[thing]--PaddingRight: var(--pf-c-page__[thing]--m-page-insets--PaddingRight);
  }
}
```

src/patternfly/demos/Page/page-template-sidebar.hbs
```diff
@@ -1,38 +1,18 @@
...
-          {{/nav-link}}
-        {{/nav-item}}
-      {{/nav-list}}
-    {{/nav}}
+    {{#> page-sidebar-body}}
+      {{> page-template-sidebar-nav}}
+    {{/page-sidebar-body}}
+    {{#> page-sidebar-body}}
+      {{> page-template-sidebar-nav}}
+    {{/page-sidebar-body}}
```
mcoker, line 16: Did you mean to have 2 of these? 

Screenshot 2023-03-02 at 11 15 04 AM

mcoker (changes requested):

src/patternfly/demos/Page/page-template-sidebar.hbs
mattnolting: Nope, good catch

src/patternfly/components/Page/page.scss
```diff
@@ -427,8 +438,12 @@ $pf-c-page--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
...
-  padding-bottom: var(--pf-c-page__sidebar-body--PaddingBottom);
+  padding-right: var(--pf-c-page__sidebar-body--PaddingRight);
+  padding-left: var(--pf-c-page__sidebar-body--PaddingLeft);
+
+  &:last-child {
+    flex-grow: 1;
+  }
 
   &.pf-m-menu {
     --pf-c-page__sidebar-body--PaddingTop: var(--pf-c-page__sidebar-body--m-menu--PaddingTop);
```
thatblindgeye, line 449: Doesn't look like this and the below PaddingBottom vars will do anything since they're not applied to a property.

thatblindgeye: Push an update to remove these lines and the CSS vars that weren't being used.

PR #5103, opened 2022-09-19
fix(slider): fixed excessive wrapping of labels

### LINKED ISSUE #5053: Bug - word-break can break words unexpectedly
Reported in slack, nesting absolutely positioned elements will break every character using `word-break: break-word;`

https://codesandbox.io/s/muddy-wave-1gy49s?file=/index.html

Looks like `break-word` has been deprecated, too. https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

srambach:
Fixes #5053 

This fixes excessive (every character) wrapping resulting from an absolutely positioned element within another.
However, note that 
- long words without breaks won't break, and could cause horizontal overlap.
- normal word breaks could cause vertical overlap if there's something directly below
- There are ways to fix this ([more info here](https://codesandbox.io/s/slider-word-wrap-fl44nh?file=/style.css)) but require a fixed min or max width.

### DIFF
--- a/src/patternfly/components/Slider/slider.scss
+++ b/src/patternfly/components/Slider/slider.scss
@@ -165,6 +165,7 @@
 .pf-c-slider__step-label {
   position: absolute;
   top: var(--pf-c-slider__step-label--Top);
+  word-break: normal;
   transform: translateX(var(--pf-c-slider__step-label--TranslateX));
 }
 


mcoker (approved):
LGTM. All of the things you mentioned are pre-existing.

srambach:
> LGTM. All of the things you mentioned are pre-existing.

Right - just wanted to be clear those conditions will still happen. Thanks for clarifying.

mattnolting (approved):
LPTM! 🎉

PR #5104, opened 2022-09-20
feat(Button): added progress support for inline link

### LINKED ISSUE #4994: Button - add progress support to inline link
Need a plain version of the progress button (i.e. link button styling) to support inline use as in situations shown here: https://marvelapp.com/prototype/b3aaaed

thatblindgeye:
Closes #4994 

[Button progress examples](https://patternfly-pr-5104.surge.sh/components/button#progress)

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -116,8 +116,8 @@
   --pf-c-button--m-link--m-inline--PaddingRight: 0;
   --pf-c-button--m-link--m-inline--PaddingBottom: 0;
   --pf-c-button--m-link--m-inline--PaddingLeft: 0;
-  --pf-c-button--m-inline--m-link__progress--Left: var(--pf-global--spacer--xs);
-  --pf-c-button--m-link--m-inline--m-in-progress--PaddingLeft: calc(var(--pf-c-button--m-inline--m-link__progress--Left) + 1rem + var(--pf-global--spacer--sm));
+  --pf-c-button--m-link--m-inline__progress--Left: var(--pf-global--spacer--xs);
+  --pf-c-button--m-link--m-inline--m-in-progress--PaddingLeft: calc(var(--pf-c-button--m-link--m-inline__progress--Left) + 1rem + var(--pf-global--spacer--sm));
 
   // Link danger
   --pf-c-button--m-link--m-danger--BackgroundColor: transparent;
@@ -374,7 +374,7 @@
 
     &.pf-m-inline {
       --pf-c-button--FontSize: var(--pf-c-button--m-link--m-inline--FontSize);
-      --pf-c-button__progress--Left: var(--pf-c-button--m-inline--m-link__progress--Left);
+      --pf-c-button__progress--Left: var(--pf-c-button--m-link--m-inline__progress--Left);
 
       display: inline;
       padding: var(--pf-c-button--m-link--m-inline--PaddingTop) var(--pf-c-button--m-link--m-inline--PaddingRight) var(--pf-c-button--m-link--m-inline--PaddingBottom) var(--pf-c-button--m-link--m-inline--PaddingLeft);
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -112,12 +112,12 @@
   --pf-c-button--m-link--m-inline--FontSize: inherit;
   --pf-c-button--m-link--m-inline--hover--TextDecoration: var(--pf-global--link--TextDecoration--hover);
   --pf-c-button--m-link--m-inline--hover--Color: var(--pf-global--link--Color--hover);
-  --pf-c-button--m-inline--m-link--PaddingTop: 0;
-  --pf-c-button--m-inline--m-link--PaddingRight: 0;
-  --pf-c-button--m-inline--m-link--PaddingBottom: 0;
-  --pf-c-button--m-inline--m-link--PaddingLeft: 0;
-  --pf-c-button--m-inline--m-link--m-in-progress--PaddingLeft: calc(var(--pf-global--spacer--xs) + 1rem + var(--pf-global--spacer--sm));
+  --pf-c-button--m-link--m-inline--PaddingTop: 0;
+  --pf-c-button--m-link--m-inline--PaddingRight: 0;
+  --pf-c-button--m-link--m-inline--PaddingBottom: 0;
+  --pf-c-button--m-link--m-inline--PaddingLeft: 0;
   --pf-c-button--m-inline--m-link__progress--Left: var(--pf-global--spacer--xs);
+  --pf-c-button--m-link--m-inline--m-in-progress--PaddingLeft: calc(var(--pf-c-button--m-inline--m-link__progress--Left) + 1rem + var(--pf-global--spacer--sm));
 
   // Link danger
   --pf-c-button--m-link--m-danger--BackgroundColor: transparent;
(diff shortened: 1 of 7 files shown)


mcarrano (commented):
This looks good @thatblindgeye . My only question is what's the difference between "inline loading" and "inline loader?" Looks like there is a spacing difference between the text and spinner?

src/patternfly/components/Button/button.scss
```diff
@@ -370,11 +370,14 @@
...
 
       display: inline;
-      padding: 0;
       text-align: left;
       white-space: normal;
       cursor: pointer;
 
+      &:not(.pf-m-in-progress) {
+        padding: 0;
+      }
```
mcoker, line 379: This is going to give the in-progress button padding on all sides, which I don't think we want. Also it will increase the specificity of the padding. So if someone had a local override in their app for something like adding a small left/right padding to these buttons via `.pf-c-button.pf-m-link.pf-m-inline { padding: 0 .5rem; }`, that will no longer work since the original compiled selector changed to `.pf-c-button.pf-m-link.pf-m-inline:not(.pf-m-in-progress)`, which has higher specificity than their rule. This isn't always a huge deal since we encourage using vars to override properties, but there is no var where we set `padding: 0`, and the button is a really common component, so it would be good to avoid changing the specificity if we can. Though I don't think you need to do this anyways, you could probably just add a `padding-left` with a var for the inline link progress button.

Here's a screenshot of what it looks like with this change in context of a paragraph with text around it, which is often how inline link buttons are used.

Sep-20-2022 16-54-06

I added this CSS to the `pf-m-in-progress` button 

```
    padding: 0; // overrides the padding added from this PR
    --pf-c-button__progress--Left: var(--pf-global--spacer--xs); // a super small left spacer to move the progress spinner to the right a bit - otherwise it looks a little too close to the text to the left of the button
    padding-left: calc(var(--pf-global--spacer--xs) + 1rem + var(--pf-global--spacer--sm)); // left padding for the spinner - xs spacer that equals the `__progress--Left` value used above + 1rem (the size of the spinner) + a sm spacer to create space between the spinner and the button text
```

This is what it looks like after

Sep-20-2022 17-20-27

I'm guessing something like that might work?

@mcarrano what I have above will shift any text to the right of the inline link button over a bit to make room for the spinner when it appears - as you can see in the screenshot above. Is that OK? It doesn't happen with non-inline buttons because those buttons have default left/right padding that we can shift around to make room for the spinner SVG without changing the overall width of the button. The inline link button doesn't have left/right padding by default.

mcoker (changes requested):
Left a comment for review.

Also I think we should use a `1em` sized spinner since an inline link button's font-size adapts to whatever the `font-size` is where it's used. 

I'm guessing it might make sense to add `--pf-c-spinner--diameter: 1em;` to the spinner in these examples, and use that in the react component by default, unless a user passes `spinnerSize`, then use whatever they pass. Though we don't currently pass `spinnerSize` - it's hardcoded in the button as the `md` size - https://github.com/patternfly/patternfly-react/blob/69c858f3d65e2fc0dd16e557ffdefaad2a706ff3/packages/react-core/src/components/Button/Button.tsx#L170

If we go that route, we may also add a `spinnerDiameter` for a custom size. Though maybe there's a better way to do all of that?

> Looks like there is a spacing difference between the text and spinner?

+1 to @mcarrano's comment, which is just presenting weird since the plain button's "in-progress" is showing before the inline link's "progress" (not "in-progress"), so it looks like the plain progress spinner is part of that inline link progress. Maybe we should just insert a line break or two between the variations in the progress example?

What it looks like now:

Screen Shot 2022-09-20 at 5 33 30 PM

With line breaks:

Screen Shot 2022-09-20 at 5 36 13 PM

thatblindgeye:
@mcoker 

> Also I think we should use a 1em sized spinner since an inline link button's font-size adapts to whatever the font-size is where it's used.
>
> I'm guessing it might make sense to add `--pf-c-spinner--diameter: 1em;` to the spinner in these examples, and use that in the react component by default, unless a user passes `spinnerSize`, 

Would the end result then be that only the inline link, in progress button has a `1em` sized spinner (via the `diameter` var) and the non-inline link buttons retain a default `md` size? Or would `1em` be the default for any progress button, with the ability to override it via the size modifier classes on the spinner itself?

> I added this CSS to the pf-m-in-progress button

I wasn't sure if you had added that code in dev tools or the scss file (trying in the scss file myself, adding that code to the `.pf-m-in-progress` block results in all progress buttons losing padding). Doing something similar in the scss file, after reverting the updates I made in this PR I added the follow block:

```
// within the `&.pf-m-inline` block at line 369
&.pf-m-in-progress {
  --pf-c-button__progress--Left: var(--pf-global--spacer--xs);

  padding-left: calc(var(--pf-global--spacer--xs) + 1rem + var(--pf-global--spacer--sm));
}
```

Which results in the behavior shown in your last video above. This would add another level of specificity, but it wouldn't affect the `.pf-c-button.pf-m-link.pf-m-inline` selector like this PR currently does.

Could also do the inverse by adding a `&.pf-m-link.pf-m-inline` block within `.pf-m-in-progress`, which that might depend on a) where in the CSS order we'd want this, and b) whether there's convention of where to place these sorts of blocks (is it more styling for an in progress button that is also an inline link, or for an inline link button that is also in progress).

src/patternfly/components/Button/button.scss
mcarrano: @mcoker I think that's fine. I really don't see this being used in a block of text like that anyway. Most of the places this might be used would be within a form or similar place where using a standard bordered button will be too heavy visually.

mcoker:
> Would the end result then be that only the inline link, in progress button has a 1em sized spinner (via the diameter var) and the non-inline link buttons retain a default md size? Or would 1em be the default for any progress button, with the ability to override it via the size modifier classes on the spinner itself?

Just the inline link I think, it's probably a more valid use case since its font-size changes based on the context on its own using [inherit](https://github.com/patternfly/patternfly/blob/9e3c7120f9d1485d70a5d08ad29cb4608cc27e0b/src/patternfly/components/Button/button.scss#L112), and the regular button's font is explicit using the [default font size](https://github.com/patternfly/patternfly/blob/9e3c7120f9d1485d70a5d08ad29cb4608cc27e0b/src/patternfly/components/Button/button.scss#L9). Also it may be breaking to change that on the existing button progress spinner.

Your code there looks great, the extra specificity is fine since it's new and I _think_ that's the best way to target that element<sup>1</sup>. It's changing specificity that would be breaking.

<sup>1</sup> - You could try this below? (untested, but works in my head?)

#### Padding
* change the inline-link padding to like `padding: var(--pf-c-button--m-inline--m-link--PaddingTop) var(--pf-c-button--m-inline--m-link--PaddingRight) var(--pf-c-button--m-inline--m-link--PaddingBottom) var(--pf-c-button--m-inline--m-link--PaddingLeft)`
  * or you could do `padding: 0 0 0 var(--pf-c-button--m-inline--m-link--PaddingLeft);` just to support this. If you're gonna have one var, seems like you should probably have all for customization's sake.
* set those vars with a value of `0`
* Then in `.pf-c-button.pf-m-in-progress`, set `--pf-c-button--m-inline--m-link--PaddingLeft: var(--pf-c-button--m-inline--m-link--m-in-progress--PaddingLeft);`

So the specificity for that would be:
* `.pf-c-button.pf-m-in-progress` instead of `.pf-c-button.pf-m-link.pf-m-inline.pf-m-in-progress`

#### Progress left
* Set `--pf-c-button__progress--Left: var(--pf-c-button--m-inline--m-link__progress--Left);` in `.pf-c-button.pf-m-inline.pf-m-link` instead of `.pf-c-button.pf-m-inline.pf-m-link`.

thatblindgeye:
@mcoker Made the suggested changes! I had to tweak a couple of handlebars files to get the inline link progress rendered without the `pf-m-md` class on the spinner.

mcoker (approved):
nice job @thatblindgeye!!

Issue #5106, opened 2022-09-21
Selectable node text should be button to fix react keyboard focus


mcoker:
per https://github.com/patternfly/patternfly-react/issues/7759

Trying out making the node text element a natively selectable element to address the keyboard focus issue.

This update:
1. Makes the node text a `<button>`
1. Adds a pseudo element to the node text `<button>` that occupies the entire row
1. Adds a `z-index` to the selectable node's toggle button so it stacks on the text button's pseudo element

Those last 2 changes are likely not needed to handle the keyboard focus issue, but it does make it so that clicking anywhere in the node triggers a click on the node text button. I'm not sure if that would simplify the way we're handling click events and tab focus and all of that between the parent node and child toggle and text elements.

cc @kmcfaul

PR #5107, opened 2022-09-21
fix(tree-view): made selectable node text a button to fix keyboard a11y

mcoker:
Makes the text element a `<button>` when a node is selectable and expandable.

convenience link - https://patternfly-pr-5107.surge.sh/components/tree-view#with-selectable-expandable-nodes

---

fixes https://github.com/patternfly/patternfly/issues/5106


kmcfaul (approved):

srambach (approved):
Seems ok for kb a11y. (screenreader isn't so great but this doesn't change that)

PR #5118, opened 2022-09-23
fix(form-group): fixed help wrapping

### LINKED ISSUE #5063: Bug - Form Group - Labels with icons should be on the same line
**Describe the problem**
There needs to be a way to keep the label and the icon on the same line, and Patternfly currently doesn't support this with its current Form styling. 

This issue is described in detail in our Keycloak UI https://github.com/keycloak/keycloak-ui/issues/2073

It can be reproduced in Keycloak UI with the following steps: 
1. Run Keycloak 19: `docker run --rm -p 8080:8080 -e KEYCLOAK_ADMIN=admin -e KEYCLOAK_ADMIN_PASSWORD=admin [quay.io/keycloak/keycloak:latest](http://quay.io/keycloak/keycloak:latest) start-dev`
2. Go to `Clients` from the left-side menu
3. Create a new client
4. Select created client, go to `Authorization` > `Export` tab
5. Observe that icon is on its own on the next line

**Expected behavior**
The labels with icons should be on the same line. 

Yes, please see https://github.com/keycloak/keycloak-ui/issues/2073

**Screenshots**
Screenshot 2022-08-31 at 17 19 32

**What is your environment?**
 - OS: macOS Monterey v12.5 
 - Browser Chrome Version 104.0.5112.101 (Official Build) (x86_64)

Keycloak UI

**Any other information?**

mattnolting:
fixes #5063

### DIFF
--- a/scripts/gulp/html.js
+++ b/scripts/gulp/html.js
@@ -20,6 +20,10 @@ hbsInstance.registerHelper('concat', (...params) => {
   return params.join('');
 });
 
+hbsInstance.registerHelper('removeWhiteSpaceForPrettier', () => {
+  return 'removeWhiteSpaceForPrettier';
+})
+
 hbsInstance.registerHelper('ifEquals', (arg1, arg2, options) => {
   return (arg1 === arg2) ? options.fn(this) : options.inverse(this);
 });
@@ -83,7 +87,9 @@ function compileMD0(srcFiles) {
                 let html = hbsInstance.compile(node.value)({});
                 html = prettyhtml(html)
                   .contents.replace(/class /g, '')
-                  .replace(/ class>/g, '>');
+                  .replace(/ class>/g, '>')
+                  .replace(/[\n\t\s\r]+&nbsp;[\n\t\s\r]+/g, '&nbsp;')
+                  .replace(/[\n\t\s\r]+removeWhiteSpaceForPrettier[\n\t\s\r]+/g, '');
                 node.lang = 'html';
                 node.value = html;
               } catch (error) {
--- a/scripts/gulp/html.js
+++ b/scripts/gulp/html.js
@@ -88,7 +88,6 @@ function compileMD0(srcFiles) {
                 html = prettyhtml(html)
                   .contents.replace(/class /g, '')
                   .replace(/ class>/g, '>')
-                  .replace(/[\n\t\s\r]+&nbsp;[\n\t\s\r]+/g, '&nbsp;')
                   .replace(/[\n\t\s\r]+removeWhiteSpaceForPrettier[\n\t\s\r]+/g, '');
                 node.lang = 'html';
                 node.value = html;
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -5,23 +5,6 @@ cssPrefix: pf-c-form
 ---
 
 ## Examples
-### Label updates
-```hbs
-{{#> form form--id="form-vertical"}}
-  {{#> form-group form-group--id="-name"}}
-    {{#> form-group-label}}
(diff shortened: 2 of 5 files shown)


mattnolting:
@srambach @mcoker This is a model of the necessary updates for required and help wrapping. @DaoDaoNoCode is looking at a way to remove generated whitespace and returns in the workspace html. The changes are outlined here https://github.com/patternfly/patternfly/issues/5063#issuecomment-1256311218. 

> The solution to this issue is to omit returns in the html and use `&nbsp;` instead of a standard space (currently rendered in the markup).
> 
> Instead of
> 
> ```
> <div class="pf-c-form__group-label">
>   <label class="pf-c-form__label">
>     <span class="pf-c-form__label-text">Label with very long text lorem ipsum</span>
>     <span class="pf-c-form__label-required" aria-hidden="true"> *</span>
>   </label> 
>   <button>
>     <svg>
>   </button>
> </div>
> ```
> 
> It needs to be written and rendered as
> 
> ```
> <div class="pf-c-form__group-label"><label class="pf-c-form__label"><span class="pf-c-form__label-text">Label with very long text lorem ipsum</span>&nbsp;<span class="pf-c-form__label-required" aria-hidden="true">{remove space here}*</span>&nbsp;<span {change button to span, display: inline}><svg></span></label></div>
> ```
> 
> A couple of things to note
> 
> * Button will not render an `inline` display value. https://html.spec.whatwg.org/multipage/rendering.html#button-layout, Therefore a `<span>` should be used in its place with `role="button"`
> * There is a space before the required asterisk. This should be removed and a non-breaking space `&nbsp;`, before the element, should be used in its place.
> * Help button should move to inside the `__label` element

scripts/gulp/html.js
```diff
@@ -83,7 +87,9 @@ function compileMD0(srcFiles) {
                 let html = hbsInstance.compile(node.value)({});
                 html = prettyhtml(html)
                   .contents.replace(/class /g, '')
-                  .replace(/ class>/g, '>');
+                  .replace(/ class>/g, '>')
+                  .replace(/[\n\t\s\r]+&nbsp;[\n\t\s\r]+/g, '&nbsp;')
```
mcoker, line 91: I'm not super hip to regex, but if I'm not mistaken, this removes any char that creates space around `&nbsp`'s? If so, it's valid to have spaces around `&nbsp` so I'm not sure we want to remove them with no way to preserve them. https://codepen.io/mcoker/pen/yLjzQLO

DaoDaoNoCode: You are right, maybe I can leave this to @mattnolting because I have already created `removeWhiteSpaceForPrettier` helper which he can use before and after `&nbsp;` where he wants to remove all the spaces around it so it will be more specific and will not influence the whole app.

mattnolting: @mcoker I removed this line and used `{{removeWhiteSpaceForPrettier}}` where needed. Ready for review.

src/patternfly/components/Form/examples/Form.md
```diff
@@ -4,6 +4,23 @@ section: components
 cssPrefix: pf-c-form
 ---
 
+## Examples
```
DaoDaoNoCode, line 7: Could you let me know why you add another `Example` section here?

srambach (approved):
🦸 great job, and props to @DaoDaoNoCode for stepping into help on this also.

src/patternfly/components/Form/examples/Form.md
mcoker: Removed it

mcoker:
@srambach rebased and made a couple of updates in [this commit](https://github.com/patternfly/patternfly/pull/5118/commits/b9dc166f12b09730e659c8249a654433faace71e). Would you mind reviewing again?

srambach (approved):
LGTM 👍

Issue #5119, opened 2022-09-23
Update outdated dependencies


mcoker:
Running [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) shows a lot of dependencies that are out of date.

Can we update some/all of these? I imagine some would likely need to be part of a major release if they're used to compile CSS - like cssnano (v4 to v5) or sass (minor version bump - might be ok?)

And some likely need to be kept in sync with other repos - like react, maybe prettier, surge?

Just glancing at it, I'd really like to update commitlint, stylelint, and prettier.

cc @jschuler @nicolethoen 

```
$ ncu
Using yarn
Checking /Users/cmichael/repos/patternfly/package.json
[====================] 53/53 100%

 @commitlint/cli                                   ^13.1.0  →       ^17.1.2
 @commitlint/config-conventional                   ^13.1.0  →       ^17.1.0
 @octokit/rest                                    ^16.40.1  →       ^19.0.4
 @patternfly/documentation-framework                1.2.27  →        1.2.30
 cheerio                                       ^1.0.0-rc.3  →  ^1.0.0-rc.12
 commander                                          ^5.1.0  →        ^9.4.0
 cssnano                                            ^4.0.0  →       ^5.1.13
 eslint                                             ^7.0.0  →       ^8.23.1
 eslint-config-standard                            ^16.0.0  →       ^17.0.0
 eslint-config-standard-jsx                        ^10.0.0  →       ^11.0.0
 eslint-config-standard-react                      ^11.0.0  →       ^11.0.1
 eslint-plugin-import                              ^2.24.2  →       ^2.26.0
 eslint-plugin-node                                ^11.0.0  →       ^11.1.0
 eslint-plugin-promise                              ^5.0.0  →        ^6.0.1
 eslint-plugin-react                                ^7.0.0  →       ^7.31.8
 glob                                               ^7.1.2  →        ^8.0.3
 gulp-cli                                           ^2.2.0  →        ^2.3.0
 gulp-iconfont                                     ^10.0.3  →       ^11.0.1
 gulp-postcss                                       ^8.0.0  →        ^9.0.1
 gulp-rename                                        ^1.4.0  →        ^2.0.0
 gulp-sourcemaps                                    ^2.6.5  →        ^3.0.0
 husky                                              ^1.3.1  →        ^8.0.1
 lint-staged                                        ^9.2.5  →       ^13.0.3
 node-plop                                         ^0.18.0  →       ^0.31.0
[...]

PR #5123, opened 2022-09-23
chore(deps): updated commitlint, stylelint, prettier

mcoker:
resolves some of https://github.com/patternfly/patternfly/issues/5119

made breaking change notes to fix using `:not()` as changing `:not(.foo):not(.bar)` to `:not(.foo, .bar)` will change the specificity from 20 to 10. From https://developer.mozilla.org/en-US/docs/Web/CSS/:not

> _The specificity of the :not() pseudo-class is replaced by the specificity of the most specific selector in its comma-separated argument of selectors; providing the same specificity as if it had been written [:not(:is(argument))](https://developer.mozilla.org/en-US/docs/Web/CSS/:is)_


nicolethoen (approved):
this seems good to me

srambach:
Are the @import url() changes meant to be part of this PR?

mcoker:
@srambach yep, the `url()` function isn't needed with an `@import` statement, the URL can just be passed as a string.

srambach:
> @srambach yep, the `url()` function isn't needed with an `@import` statement, the URL can just be passed as a string.

k, just checking it didn't slip in from another since it wasn't mentioned in the description or issue.

srambach (approved):

PR #5130, opened 2022-09-28
docs(accordion): gave scrollable region keyboard access in fixed examples

### LINKED ISSUE #5124: Accordion examples - scrollable region must have keyboard access
The [Fixed example](https://www.patternfly.org/v4/components/accordion/html#fixed) has an expanded accordion with a scrollbar that cannot receive keyboard focus.

RuiBrian:
Closes #5124

Set the `tabindex` property to 0 so that the scroll region is focusable (and thus can be accessed via keyboard)


src/patternfly/components/Accordion/accordion-expanded-content.hbs
```diff
@@ -1,4 +1,4 @@
-<{{#if accordion--IsDefinitionList}}dd{{else}}div{{/if}} class="pf-c-accordion__expanded-content{{#if accordion-expanded-content--IsExpanded}} pf-m-expanded{{/if}}{{#if accordion-expanded-content--IsFixed}} pf-m-fixed{{/if}}{{#if accordion-expanded-content--modifier}} {{accordion-expanded-content--modifier}}{{/if}}"
+<{{#if accordion--IsDefinitionList}}dd{{else}}div{{/if}} tabindex=0 class="pf-c-accordion__expanded-content{{#if accordion-expanded-content--IsExpanded}} pf-m-expanded{{/if}}{{#if accordion-expanded-content--IsFixed}} pf-m-fixed{{/if}}{{#if accordion-expanded-content--modifier}} {{accordion-expanded-content--modifier}}{{/if}}"
```
thatblindgeye, line 1: Rather than applying `tabindex=0` to every content component, we should apply it only when `accordion-expanded-content--IsFixed` is true. Or alternatively we could apply it in the example itself rather than in the handlebars. @mcoker do you have any preference? Ideally we would only want the tabindex when a scrollbar is present, which may not always be the case when the accordion content `isFixed`, but it might depend on if we anticipate showing an example where `isFixed` is applied but there isn't a scrollbar?

thatblindgeye (changes requested):
Had one change request below.

Also, after looking at the preview build, there's a couple of other things we should do for the example that has the scrollable accordion content per the PatternFly content on [scrollable elements](https://www.patternfly.org/v4/accessibility/product-development-guide#scrollable-elements)

- In this handlebars file, we should apply a `role="region"` attribute to the accordion content when `accordion-expanded-content--IsFixed` is true, unless `accordion--IsDefinitionList` is true (in which case I don't think we want the region role applied since the content would have the semantics of a `<dd>` element already).
- In the examples Markdown file, an aria label or aria-labelledby should be added to the accordion content with the scrollbar. Since we already have the accordion toggle acting as a label of sorts for the content, using `aria-labelledby` might be better here. We'd just need to give the accordion toggle an ID, then pass that ID as a value to the content's `aria-labelledby`.

    Using aria-label can work as well, since we could instead just pass in a string like `aria-label="Item two content"` to the accordion content.

src/patternfly/components/Accordion/accordion-expanded-content.hbs
mcoker: @thatblindgeye good question. It's most likely true that if we're using a fixed accordion in core, the content shown will likely be scrollable (with a scrollbar visible), but it isn't always necessarily the case.

Ideally I think we could have a conditional in this partial that tests for the presence of a parameter indicating there is scrollable content:

```
{{#if accordion-expanded-content--HasScrollbar}}
  tabindex="0"
{{/if}}
```

That could be set 1 of a few ways:
1. Individually on expandable content elements that have a scrollbar
1. On the parent `{{#> accordion}}` partial, which will set it to true for all expandable content elements
1. We could tie it to `accordion-expanded-content--IsFixed` by setting `accordion-expanded-content--HasScrollbar=accordion-expanded-content--IsFixed`, which is functionally equivalent to changing the conditional above to test for `accordion-expanded-content--IsFixed`. Through some trickery, that could be set by default, then we could set `accordion-expanded-content--HasScrollbar=reset` on individual sections that don't have a scrollbar.

The downside to 1 and 2 is that it requires us to set the tabindex manually per accordion section, and it isn't guaranteed we'll add it consistently. Given that accordions in core are static, it's most likely a fixed demo/example will have scrollable content, so I'd opt for 3. And maybe if we're showing a fixed accordion with multiple sections expanded and not all have scrollbars, we'll have the option to remove the `tabindex`.

src/patternfly/components/Accordion/accordion-expanded-content.hbs
```diff
@@ -1,5 +1,13 @@
 <{{#if accordion--IsDefinitionList}}dd{{else}}div{{/if}} class="pf-c-accordion__expanded-content{{#if accordion-expanded-content--IsExpanded}} pf-m-expanded{{/if}}{{#if accordion-expanded-content--IsFixed}} pf-m-fixed{{/if}}{{#if accordion-expanded-content--modifier}} {{accordion-expanded-content--modifier}}{{/if}}"
   {{#unless accordion-expanded-content--IsExpanded}}hidden{{/unless}}
+  {{#if accordion-expanded-content--IsFixed}}
+    {{#unless accordion--IsDefinitionList}}
+      role="region"
```
mcoker, line 5: @thatblindgeye according to the docs you linked:

> Additionally, when adding tabIndex in this scenario, add role='region' and an aria-label that describes the scrollable element or an aria-labelledby pointing to a heading with a descriptive, accessible name.

Is it true that the region/label should also only be present when there is a visible scrollbar? If so, I think we need to move that into the conditional that checks for `accordion-expanded-content--HasNoScrollbar`.

thatblindgeye: I think that would make sense. In React for the Page component as an example, you *could* pass in `hasOverflowScroll` which applies a tabindex even when there is no scrollbar, but ideally you'd only want to include it when there is an actual scrollbar.

thatblindgeye (commented):
This is looking great! Reading through @mcoker comments again, it looks like the only thing left to do would be to add `accordion-expanded-content--HasNoScrollbar="true"` in examples where there is no scrollbar (which I believe would be all but the one accordion content that does have a scrollbar).

mcoker (approved):
Great job! Once @thatblindgeye's comment about that attribute is resolved, LGTM! @thatblindgeye do we need to add a11y docs to the tables on the core pages, or is that handled by the a11y docs on the website now?

thatblindgeye:
@mcoker The a11y docs should be covered on the website now. When I updated the Accordion a11y docs in org I ported over the Core a11y info. Which I now realize I never removed that info from Core. Since there are several components that have had their a11y docs updated, including porting over that Core info, I can open a PR to remove that when I return from PTO if you'd want?

mcoker:
Sounds great, thanks @thatblindgeye!!

thatblindgeye (approved):

PR #5148, opened 2022-10-07
chore(wizard): updated button order

### LINKED ISSUE #5110: Button order for the Wizard is counter intuitive/not following established convention
I would like to put the discussion around the button order for the Wizard forward as an issue. It seems this was brought up in https://forum.patternfly.org/t/wizard-button-placement/ but never got a conclusion. I believe the

```
[ Next ]  [ Back ]   Cancel
```

is very counter intuitive when used with a desktop application. it goes against the conventions and guidelines that exist for Windows and macOS:

  * https://docs.microsoft.com/en-us/windows/win32/controls/wizards
  * https://developer.apple.com/design/human-interface-guidelines/macos

I made a demo recording for the desktop application we are developing for CodeReady Containers, and we are using Patternfly to get a consistent look and feel. We were hoping for feedback on the functionality, but instead it started to center around the first page of the Wizard why this button order was chosen. It took a while to confirm this was 'by design' and default from Patternfly:

 * https://www.patternfly.org/v4/components/wizard/design-guidelines/

I would prefer to re-use existing and established components... but I also want people to be able to rely on a mental model that they already have from the desktop they use daily.

blfetche:
Changed the button order in an effort to improve user experience and common mental models when using wizards.

closes #5110

### DIFF
--- a/src/patternfly/components/Wizard/examples/Wizard.md
+++ b/src/patternfly/components/Wizard/examples/Wizard.md
@@ -80,8 +80,7 @@ import './Wizard.css'
         {{> __wizard-form}}
       {{/wizard-main}}
     {{/wizard-inner-wrap}}
-    {{#> wizard-footer}}
-    {{/wizard-footer}}
+    {{> wizard-footer}}
   {{/wizard-outer-wrap}}
 {{/wizard}}
 ```
@@ -158,8 +157,7 @@ import './Wizard.css'
         {{> __wizard-form}}
       {{/wizard-main}}
     {{/wizard-inner-wrap}}
-    {{#> wizard-footer}}
-    {{/wizard-footer}}
+    {{> wizard-footer}}
   {{/wizard-outer-wrap}}
 {{/wizard}}
 ```
@@ -214,8 +212,7 @@ import './Wizard.css'
               {{/drawer-body}}
             {{/drawer-panel}}
           {{/drawer-main}}
-          {{#> wizard-footer}}
-          {{/wizard-footer}}
+          {{> wizard-footer}}
         {{/drawer}}
       {{/wizard-main}}
     {{/wizard-inner-wrap}}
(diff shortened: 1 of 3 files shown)


mcoker (commented):
Nice! LGTM, though I poked around and looks like every instance of a wizard footer in the workspace includes all of these buttons the exact same way. I think that might be a good case for adding them directly to the wizard-footer template, and replace `{{> @partial-block}}` with these buttons in that template.

https://github.com/patternfly/patternfly/blob/b0eca7bb80c043ec9b2da216b7520201265526e8/src/patternfly/components/Wizard/wizard-footer.hbs#L5

Then in all of the examples/demos we can replace the `{{#> wizard-footer}}...{{/wizard-footer}}` blocks with `{{> wizard-footer}}`. wdyt about making that update?

mcoker (approved):
LGTM! Thanks!!

blfetche:
@mcarrano 

https://patternfly-pr-5148.surge.sh/

image

mcarrano:
Yep. That's what we want. Look good to you, @mceledonia @mmenestr ?

mmenestr:
Looks good button wise! One thing I did notice is that only the "with drawer" demo has the footer with a shadow... the other don't have a shadow to make it obvious it's sticky. Is that normal?

mcoker:
@mmenestr good call out. Whether the footer is sticky or not technically depends on the overall height of the wizard being limited in some regard, so that if there is a lot of wizard step content, the content overflows and scrolls instead of the wizard just growing in height to accommodate the amount of content in a step. When we created the wizard, we supported it in the modal and page content, and specifically limit the wizard height so that overflow will scroll - so we only applied the footer shadow if the wizard was in a modal or the page content, since those were cases where we could guarantee that the footer would be sticky. We've since added a case for the wizard in a drawer, too.

Our examples are all just standalone wizards (except for the drawer example), so there is no shadow there by default. If you check the demos, you'll see there is a shadow since the wizard is used in the context of the page and wizard.

We have a breaking change task to just apply the shadow by default (which would add shadows to standalone examples), implying the user should ensure the wizard height is limited so that steps with a lot of content scroll so that the footer is always visible. You can see that commented here in the component SCSS https://github.com/patternfly/patternfly/blob/3db0726e3397bd7c7a5b4fde5d820b791db3213f/src/patternfly/components/Wizard/wizard.scss#L552-L558

mmenestr:
Ok perfect @mcoker, in that case this looks good to me!

Issue #4972, opened 2022-07-13
Skip to content and button style conflict


mcoker:
`.pf-c-skip-to-content` is applied directly to `.pf-c-button`, and both have `position` set to different values (absolute vs relative), so if button styles are loaded after the skip to content styles, the skip to content button is set to `position: relative`

Generally speaking we try and avoid sharing component styles on a single element. The skip to content component should probably just be a container that positions itself and can use `:focus-within` to manage the `left` position to bring it into view.

nicolethoen:
Closed by #5153

PR #5153, opened 2022-10-07
fix(skip-to-content): separate pf-c-skip-to-content from pf-c-button

nicolethoen:
Closes: #4972 

Attempt to see if this creates a new patternfly-five/patternfly alpha release

### DIFF
--- a/src/patternfly/components/SkipToContent/skip-to-content.hbs
+++ b/src/patternfly/components/SkipToContent/skip-to-content.hbs
@@ -1,8 +1,8 @@
-<div class="pf-c-skip-to-content">
-  <a class="pf-c-button pf-m-primary{{#if skip-to-content--modifier}} {{skip-to-content--modifier}}{{/if}}"
-    {{#if skip-to-content--attribute}}
-      {{{skip-to-content--attribute}}}
-    {{/if}}>
+<div class="pf-c-skip-to-content{{#if skip-to-content--modifier}} {{skip-to-content--modifier}}{{/if}}"
+  {{#if skip-to-content--attribute}}
+    {{{skip-to-content--attribute}}}
+  {{/if}}>
+  {{#> button-link button-link--attribute=(concat 'href="' skip-to-content--href '"') button-link--modifier="pf-m-primary"}}
     {{> @partial-block}}
-  </a>
+  {{/button-link}}
 </div>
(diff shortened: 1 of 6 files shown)


src/patternfly/components/SkipToContent/skip-to-content.hbs
```diff
@@ -1,6 +1,8 @@
...
-  {{> @partial-block}}
-</a>
+<div class="pf-c-skip-to-content">
+  <a class="pf-c-button pf-m-primary{{#if skip-to-content--modifier}} {{skip-to-content--modifier}}{{/if}}"
+    {{#if skip-to-content--attribute}}
+      {{{skip-to-content--attribute}}}
+    {{/if}}>
+    {{> @partial-block}}
+  </a>
+</div>
```
mcoker, line 8: This looks good as it is, but there is one small cleanup we could make in this PR, or we can open another issue and handle that separately, just lemme know.

The `--modifier` and `--attribute` attrs should ideally go on the component level thing in the hbs file, so here we would move them to the new `div.skip-to-content` element. That means we'll need to use something other than `skip-to-content--attribute='href="#target"'` to give the link an `href`.

We also have a `button-link` partial that generates a link using the button component. One way could be to do it this way below, creating a `skip-to-content--href` param to pass the URL in as:

```suggestion
<div class="pf-c-skip-to-content{{#if skip-to-content--modifier}} {{skip-to-content--modifier}}{{/if}}"
  {{#if skip-to-content--attribute}}
    {{{skip-to-content--attribute}}}
  {{/if}}>
  {{#> button-link button-link--attribute=(concat 'href="' skip-to-content--href '"') button-link--modifier="pf-m-primary"}}
    {{> @partial-block}}
  {{/button-link}}
</div>
```

Then anywhere we're using `{{#> skip-to-content [attr param with href]}}` just use `{{#> skip-to-content skip-to-content--href="#target"}}`

Looks like we would want to update these spots:

https://github.com/patternfly/patternfly/blob/e362c95792e234592911881cb3c818177c023fd5/src/patternfly/demos/NotificationDrawer/notification-drawer-demo.hbs#L2-L4

https://github.com/patternfly/patternfly/blob/e362c95792e234592911881cb3c818177c023fd5/src/patternfly/demos/Page/page-template-skip-to-content.hbs#L1-L3

https://github.com/patternfly/patternfly/blob/e362c95792e234592911881cb3c818177c023fd5/src/patternfly/components/SkipToContent/examples/SkipToContent.md#L10-L12

https://github.com/patternfly/patternfly/blob/e362c95792e234592911881cb3c818177c023fd5/src/patternfly/demos/Page/page-demo-expandable-nav.hbs#L2-L4

mcoker (changes requested):
This looks awesome! Looks like there is a syntax error in the markup keeping some of the demos from rendering.

mcoker (approved):
🥳

PR #5157, opened 2022-10-10
fix(Slider): reorder slider thumb and slider input elements

### LINKED ISSUE #5121: Bug - Slider: tab order incorrect when input is above thumb
**Describe the problem**
The tab order is incorrect when the value input is placed above the thumb. Rather than tabbing to the input then the thumb slider, focus goes to the thumb slider and then the value input.

Related to https://github.com/patternfly/patternfly-react/issues/8053

RuiBrian:
Closes #5121

Swapped slider thumb and slider input elements so that the tab order is correct (i.e., tab to input first, then slider).


thatblindgeye (approved):

thatblindgeye:
@RuiBrian Just wanted to let you know https://github.com/patternfly/patternfly/pull/5172 has been opened with some discussion that is related to this PR.

mcoker:
@thatblindgeye and I discussed and I think a consideration for this should be that the floating input is currently more like a popover/tooltip attached to the thumb. I can see updating this input to use popper with a floating/plain popover (without the box), and then the input may be before or after the thumb in any direction when the orientation of the input flips. Conceptually, I think the thumb is the primary adjustment on the slider, and the input is a representation of, and a way to change, the thumb's value. So I think there is a strong use case for the thumb remaining first in the tab order.

RuiBrian:
I see, that makes sense. Should we go ahead and close this PR then?

thatblindgeye:
@nicolethoen @jessiehuff for visibility, Michael's comments regarding the Slider are above. Would either of you agree with keeping the slider as-is as well (similar to the Label Group PR), or is the Slider *just* different enough to warrant a markup update?

nicolethoen:
Leaving it as is sounds fine to me.

PR #5161, opened 2022-10-11
feat(progress): added helper text

### LINKED ISSUE #4915: [File Upload Multiple] - Allow to control display message for file to upload
File Upload Multiple - https://www.patternfly.org/v4/components/file-upload---multiple

Using the component with `customFileHandler` prop allows us to have fine control of the file to upload as we need to validate them against a XML schema and also make sure there weren't previously uploaded. Meanwhile we need to be able to display a specific message against a file when an error occurs.

Our current work around is to display an error message above the component but that's far from ideal especially when several files are on error.

Would it be possible to have, similarly to `progressValue` and `progressVariant` props, a `progressMessage` (or equivalent) ? Allowing to completely manage the `MultipleFileUploadStatusItem` ?

A typical error message would look like this:
image

wise-king-sullyman:
Closes #4915 

Link directly to added example: https://patternfly-pr-5161.surge.sh/components/progress/#helper-text

### DIFF
--- a/src/patternfly/components/Progress/progress.scss
+++ b/src/patternfly/components/Progress/progress.scss
@@ -12,6 +12,7 @@
   --pf-c-progress__bar--before--Opacity: .2; // one-off change in opacity to allow progress bar to automatically coordinate with the indicator color
   --pf-c-progress__indicator--Height: var(--pf-c-progress__bar--Height);
   --pf-c-progress__indicator--BackgroundColor: var(--pf-c-progress__bar--before--BackgroundColor);
+  --pf-c-progress__helper-text--MarginTop: calc(var(--pf-global--spacer--xs) - var(--pf-global--spacer--md));
 
   // Modifier variables
   --pf-c-progress--m-success__bar--BackgroundColor: var(--pf-global--success-color--100);
@@ -199,6 +200,12 @@
   left: 0;
   height: var(--pf-c-progress__indicator--Height);
   background-color: var(--pf-c-progress__indicator--BackgroundColor);
+}
+
+.pf-c-progress__helper-text {
+  grid-column: 1 / 3;
+  grid-row: 3 / 4;
+  margin-top: var(--pf-c-progress__helper-text--MarginTop);
 }
 
 // stylelint-disable no-invalid-position-at-import-rule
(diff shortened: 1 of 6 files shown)


src/patternfly/components/Progress/progress.scss
```diff
@@ -12,6 +12,7 @@
   --pf-c-progress__bar--before--Opacity: .2; // one-off change in opacity to allow progress bar to automatically coordinate with the indicator color
   --pf-c-progress__indicator--Height: var(--pf-c-progress__bar--Height);
   --pf-c-progress__indicator--BackgroundColor: var(--pf-c-progress__bar--before--BackgroundColor);
+  --pf-c-progress__helper-text--MarginTop: calc(var(--pf-global--spacer--xs) - var(--pf-global--spacer--md));
```
mcoker, line 15: We haven't done this sort of thing much and I think it's worth considering how this should scale with the overall progress grid gap. We'll have the same question for the form component and anywhere else we do this.

The grid gap for the progress layout is represented by `--pf-c-progress--GridGap: var(--pf-global--spacer--md);` (`16px`). The helper text margin-top is `-12px` (whatever the gap is minus `12px`), so the gap above the helper text is `4px`, which is the `--xs` spacer value, which matches the form helper text 👍

So that means that if a user updates `--pf-c-progress--GridGap` to `24px` (`--lg` spacer), the gap above the helper text will be `12px`. And if the progress gap is `32px`, the help text is `20px`, and so on.

However, if the user updates the progress gap to `8px`, the helper text is now `-4px`, which is likely not what we want. Exhibits A and B:

Screen Shot 2022-10-11 at 1 05 56 PM

Screen Shot 2022-10-11 at 12 58 07 PM

I guess the question is, is there a proportion to what the helper text gap should be? For example, if the overall gap is `16px` now, and the helper text gap is `4px`, should the helper text be 25% of the overall gap? This may work, but it will give us spacers that don't exist in our spacing system. For example, a `24px` overall gap would create a `6px` gap for the helper text, and `6px` isn't part of our main spacer vars (we have a special spacer for form control elements we use there though)

Or should the helper text gap be separate from the overall gap - meaning it's `4px` regardless whatever the overall gap is - and it's implied that they should be updated separately? Something like `calc(var(--overall-gap) * -1 + var(--helper-text-gap))`?

curious your thoughts @mcarrano @mmenestr @mceledonia

mcoker (commented):
LGTM! I left a couple of non-blocking comments. If the help text gap isn't something we can solve quickly, we can follow up in another issue.

src/patternfly/components/Progress/progress.scss
mmenestr: @mcoker Hmm I'd probably say they should be updated separately and the gap between the progress bar and the helper text should remain 4px no matter what.

mcoker (approved):
Perfect!

PR #5169, opened 2022-10-13
fix(fonts): removed Overpass font support

### LINKED ISSUE #4219: Remove Overpass font support
When we switched the default font to the [RedHatFont](https://github.com/RedHatOfficial/RedHatFont), we created an opt-in to continue using the old font, Overpass, by using `.pf-m-overpass-font`. Let's remove this in the next breaking change release as it appears it is not in use as far as we're aware.

jenny-s51:
Closes #4219


mcoker (approved):
So nice!! LGTM!

mcoker:
For reference for anyone reviewing, here is the original issue that added the overpass opt-in - https://github.com/patternfly/patternfly/pull/2955/

mcoker (approved):
Love it! So nice to see all of this removed. Great job 👍👍

Issue #5122, opened 2022-09-23
Bug - Label group: vertical category with overflow produces wrong tab order


thatblindgeye:
**Describe the problem**
The React implementation of the label group has the [Vertical category overflow removable](https://www.patternfly.org/v4/components/label-group#vertical-category-overflow-removable) example, which has an incorrect tab order based on its visual order. When tabbing, focus goes to the truncated category name, the "2 more" button, then back to the top for the remove button.

While Core doesn't have this example, the CSS would most likely need to be updated slightly so that the markup can have the "remove" button come before the list of labels in DOM order for the vertical variant, while still appearing horizontally to the right of the category name.

**How do you reproduce the problem?**
Provide steps to reproduce. A [codepen](http://codepen.io) or [codesandbox](https://codesandbox.io/) demonstrating the problem is appreciated.

**Expected behavior**
Tabbing would cause the focus order to be category name > remove button > overflow button.

thatblindgeye:
Closing per conversation in #5172

PR #5172, opened 2022-10-13
fix(LabelGroup): updated tab order to match visual order

thatblindgeye:
WIP for #5122 

Had a conversation with @mcoker regarding this and #5157. The tldr is that this can be a bit of a gray area, where both the visual order and the logical/expected order both make sense.

The following link should link to the "Vertical category removable, updated markup" example which shows a potential fix for the linked issue. The example immediately below it, "Vertical category removable", is the behavior seen in our React repo.
[Label group](https://patternfly-pr-5172.surge.sh/components/label-group#vertical-category-removable-updated-markup)

For label group with truncated label + overflow button + close button, the visual order seems to be more left to right, top to bottom, e.g. the label should receive focus, then the close button (visually to the right of the label), then the overflow. However:

- If you view it more as a column then top to bottom, left to right also makes sense. E.g., the label receives focus, then the overflow (within the same "column" as the label, then over to the next "column" the close button receives focus.
- The horizontal label group tab order currently is setup in a way that tab order + visual order just happen to match, but updating the tab order for the vertical label group could be confusing since you'd be going from label > overflow > close button, to label > close button > overflow

[Placing the interactive elements in an order that follows sequences and relationships within the content](https://www.w3.org/WAI/WCAG21/Techniques/general/G59) may be relevant.

@jessiehuff @nicolethoen your input would be appreciated

### DIFF
--- a/src/patternfly/components/LabelGroup/examples/LabelGroup.md
+++ b/src/patternfly/components/LabelGroup/examples/LabelGroup.md
@@ -438,13 +438,50 @@ cssPrefix: pf-c-label-group
 {{/label-group}}
 ```
 
-### Vertical category removable
+### Vertical category removable, updated markup
 ```hbs
 
+{{#> label-group label-group--id="label-group-vertical-category-removable-2" label-group--modifier="pf-m-vertical pf-m-category"}}
+  {{#> label-group-main}}
+    {{#> label-group-header}}
+      {{#> label-group-label label-group-label--attribute=(concat 'id="' label-group--id '-label"  tabindex="0"')}}
+        Group label with really long text
+      {{/label-group-label}}
+      {{> label-group-close}}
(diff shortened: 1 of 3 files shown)


thatblindgeye:
@mcoker @jessiehuff Nicole and I had a quick chat regarding this update, and we're both leaning towards keeping things as-is per the reasons stated above (may cause confusion if we update the tab order for vertical only and either train of thought makes sense; Nicole also brought up a good point about how having the close button in the middle may interrupt the flow for screen readers).

Would both of you agree with that?

mcoker:
@thatblindgeye sounds good to me!

thatblindgeye:
Spoke with Jessie, and based on the discussion above for now I'll close this PR and issue (as well as the React followup).

Issue #5188, opened 2022-10-24
Dropdown - static variation not working with top aligned menu


mcoker:
`.pf-m-static` should effectively disable the dropdown component's CSS positioning of the menu. This is for use when either the dropdown menu should appear and be part of the page layout (push any content after it down), or with some other positioning tool (eg, popper).

`.pf-m-top` relies upon the default CSS positioning and uses `translateY` to reposition the menu. Currently `.pf-m-static` doesn't disable the `.pf-m-top` positioning.

PR #5189, opened 2022-10-24
fix(dropdown): disabled top positioning when using static variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5188

@thatblindgeye this disables the dropdown's `.pf-m-top` modifier from doing anything with the dropdown menu when `.pf-m-static` is set. That was probably just an oversight when we added `.pf-m-static`. You can test it by adding the classes manually in the core workspace.

### DIFF
--- a/src/patternfly/components/OptionsMenu/examples/options-menu.md
+++ b/src/patternfly/components/OptionsMenu/examples/options-menu.md
@@ -200,6 +200,7 @@ import './options-menu.css'
 | `.pf-m-top` | `.pf-c-options-menu` | Modifies to display the menu above the toggle. |
 | `.pf-m-align-right` | `.pf-c-options-menu__menu` | Modifies to display the menu aligned to the right edge of the toggle |
 | `.pf-m-expanded` | `.pf-c-options-menu` |  Modifies for the expanded state. |
+| `.pf-m-static` | `.pf-c-options-menu__menu` | Modifies to position the menu statically to support custom positioning. |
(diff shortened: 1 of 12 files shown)


thatblindgeye (approved):
Tested by adding `pf-m-top` and `pf-m-static` in the appropriate elements in both the preview build and the live site, and things look good!

mcoker:
@thatblindgeye ah! I went through all of the components we have menus for and added the variation and docs. Also reset all of the inset position properties (top, right, bottom, left) with `.pf-m-static`. Technically they won't do anything since they require `position` to be anything but `static` to work, but if a user set the menu to `position: relative` or something, for a simple custom change, any of those inset positions that were still set would start working and that's probably not preferred with `.pf-m-static`.

thatblindgeye (approved):

src/patternfly/components/OptionsMenu/examples/options-menu.md
```diff
@@ -200,6 +200,7 @@ import './options-menu.css'
 | `.pf-m-top` | `.pf-c-options-menu` | Modifies to display the menu above the toggle. |
 | `.pf-m-align-right` | `.pf-c-options-menu__menu` | Modifies to display the menu aligned to the right edge of the toggle |
 | `.pf-m-expanded` | `.pf-c-options-menu` |  Modifies for the expanded state. |
+| `.pf-m-static` | `.pf-c-options-menu__menu` | Modifies to position the menu statically to support custom positioning. |
```
srambach, line 203: I think the wording on the menu item is slightly clearer, "Modifies the menu to be statically positioned to support custom positioning." Though I don't love positioned-positioning but I can't come up with anything better.

PR #5191, opened 2022-10-25
fix(context-selector): updated static props

mcoker:
follow up from https://github.com/patternfly/patternfly/pull/5189


mattnolting (approved):
LPTM 👍

PR #5201, opened 2022-10-31
fix(toolbar): added center and baseline alignments

### LINKED ISSUE #5030: Toolbar - explore alignment; error validation throws off alignment
Toolbar

**Describe the feature**
Explore adding a modifier for this alignment issue. 

image

image

mattnolting:
closes #5030 

This PR adds:
- `.pf-m-align-items-center` to `.pf-c-toolbar__content-section` and `.pf-c-toolbar__group`
- `.pf-m-align-self-center` to `.pf-c-toolbar__group` and `.pf-c-toolbar__item`
- `.pf-m-align-self-baseline` to `.pf-c-toolbar__group` and `.pf-c-toolbar__item`

### DIFF
--- a/src/patternfly/components/Toolbar/examples/Toolbar.md
+++ b/src/patternfly/components/Toolbar/examples/Toolbar.md
@@ -665,9 +665,7 @@ As the toolbar component is a hybrid layout and component, some of its elements 
 | `.pf-m-align-right{-on-[breakpoint]}` | `.pf-c-toolbar > *` | Modifies toolbar element to align right, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-align-left{-on-[breakpoint]}` | `.pf-c-toolbar > *` | Modifies toolbar element to align left, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-align-items-center` | `.pf-c-toolbar__content-section`, `.pf-c-toolbar__group` | Modifies toolbar element to vertically align children to center. |
-| `.pf-m-align-items-baseline` | `.pf-c-toolbar__content-section`, `.pf-c-toolbar__group` | Modifies toolbar element to vertically align children to center. |
 | `.pf-m-align-self-center` | `.pf-c-toolbar__group`, `.pf-c-toolbar__item` | Modifies toolbar element to vertically align self to center. |
-| `.pf-m-align-self-baseline` | `.pf-c-toolbar__group`, `.pf-c-toolbar__item` | Modifies toolbar element to vertically align self to baseline. |
 | `.pf-m-label` | `.pf-c-toolbar__item` | Modifies toolbar item to label. |
 | `.pf-m-chip-container` | `.pf-c-toolbar__content`, `.pf-c-toolbar__group` | Modifies the toolbar element for applied filters layout. |
 | `.pf-m-overflow-container` | `.pf-c-toolbar__item`, `.pf-c-toolbar__group` | Modifies the toolbar element to hide overflow and respond to available space. Used for horizontal navigation. |
--- a/src/patternfly/components/Toolbar/examples/Toolbar.md
+++ b/src/patternfly/components/Toolbar/examples/Toolbar.md
@@ -665,7 +665,9 @@ As the toolbar component is a hybrid layout and component, some of its elements 
 | `.pf-m-align-right{-on-[breakpoint]}` | `.pf-c-toolbar > *` | Modifies toolbar element to align right, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-align-left{-on-[breakpoint]}` | `.pf-c-toolbar > *` | Modifies toolbar element to align left, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-align-items-center` | `.pf-c-toolbar__content-section`, `.pf-c-toolbar__group` | Modifies toolbar element to vertically align children to center. |
+| `.pf-m-align-items-baseline` | `.pf-c-toolbar__group` | Modifies toolbar group to vertically align children to baseline. |
 | `.pf-m-align-self-center` | `.pf-c-toolbar__group`, `.pf-c-toolbar__item` | Modifies toolbar element to vertically align self to center. |
+| `.pf-m-align-self-baseline` | `.pf-c-toolbar__group`, `.pf-c-toolbar__item` | Modifies toolbar element to vertically align self to baseline. |
 | `.pf-m-label` | `.pf-c-toolbar__item` | Modifies toolbar item to label. |
 | `.pf-m-chip-container` | `.pf-c-toolbar__content`, `.pf-c-toolbar__group` | Modifies the toolbar element for applied filters layout. |
 | `.pf-m-overflow-container` | `.pf-c-toolbar__item`, `.pf-c-toolbar__group` | Modifies the toolbar element to hide overflow and respond to available space. Used for horizontal navigation. |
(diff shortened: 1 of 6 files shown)


mattnolting:
@mcoker rebased and ready

srambach (approved):

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -208,9 +214,17 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
 
   @include pf-hidden-visible(var(--pf-c-toolbar__group--Display));
 
-  align-items: center;
+  align-items: var(--pf-c-toolbar__group--AlignSelf);
```
mcoker, line 217: This is a mismatch of the property and var name - is that intentional?

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -208,9 +214,17 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
...
   margin-right: var(--pf-c-toolbar--spacer);
 
+  &.pf-m-align-self-center {
+    --pf-c-toolbar__group--AlignSelf: center;
```
mcoker, line 221: related to the other comment, the class is `align-self` but it's going to align its children. If the goal is to set group children alignment at the group level, I think we should use a `__group--AlignItems` var and `pf-m-align-items` modifier.

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -9,13 +9,18 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   --pf-c-toolbar--RowGap: var(--pf-global--spacer--lg);
   --pf-c-toolbar--PaddingTop: var(--pf-global--spacer--md);
   --pf-c-toolbar--PaddingBottom: var(--pf-global--spacer--md);
+  --pf-c-toolbar--AlignItems--base: baseline;
+  --pf-c-toolbar--AlignSelf--base: auto;
```
mcoker, line 13: I wonder if we should have different `--base` vars for the things we allow alignments for. Just because some things may not have a default of `align-items: baseline` or `align-self: auto`. I'm also not 100% on why we have `--base` vars since we don't have responsive support where you want to reference/return to whatever the default is with a modifier. If we introduced that and needed `--base` vars, we could do that without a breaking change.

mcoker (commented):
Looks good! Left some comments. Also curious if we should update these to use the new vars

https://github.com/patternfly/patternfly/blob/78f6c2de4b0023cd5783ee138009c1ec367befeb/src/patternfly/components/Toolbar/toolbar.scss#L437

https://github.com/patternfly/patternfly/blob/78f6c2de4b0023cd5783ee138009c1ec367befeb/src/patternfly/components/Toolbar/toolbar.scss#L170-L175

mattnolting:
@mcoker updated

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -13,9 +13,12 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
   // Item
   --pf-c-toolbar__item--Display: block;
   --pf-c-toolbar__item--MinWidth--base: auto;
+  --pf-c-toolbar__item--AlignSelf: baseline;
```
mcoker, line 16: should this be `auto`? otherwise, setting `align-items` on the group doesn't seem like it will ever work, since `align-self` on the items will always override that.

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -664,6 +664,10 @@ As the toolbar component is a hybrid layout and component, some of its elements
 | `.pf-m-visible{-on-[breakpoint]}` | `.pf-c-toolbar__content`, `.pf-c-toolbar__content-section`, `.pf-c-toolbar__item`, `.pf-c-toolbar__group` | Modifies toolbar element to be shown, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-align-right{-on-[breakpoint]}` | `.pf-c-toolbar > *` | Modifies toolbar element to align right, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 | `.pf-m-align-left{-on-[breakpoint]}` | `.pf-c-toolbar > *` | Modifies toolbar element to align left, at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
+| `.pf-m-align-items-center` | `.pf-c-toolbar__content-section`, `.pf-c-toolbar__group` | Modifies toolbar element to vertically align children to center. |
+| `.pf-m-align-items-baseline` | `.pf-c-toolbar__content-section`, `.pf-c-toolbar__group` | Modifies toolbar element to vertically align children to center. |
```
mcoker, line 668:
```suggestion
| `.pf-m-align-items-baseline` | `.pf-c-toolbar__content-section`, `.pf-c-toolbar__group` | Modifies toolbar element to vertically align children to baseline. |
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -208,9 +212,26 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
...
+  align-self: var(--pf-c-toolbar__group--AlignSelf);
   margin-right: var(--pf-c-toolbar--spacer);
 
+  &.pf-m-align-items-center {
+    --pf-c-toolbar__group--AlignItems: center;
+  }
+
+  &.pf-m-align-items-baseline {
+    --pf-c-toolbar__group--AlignItems: baseline;
+  }
```
mcoker, line 225: Since this is the default, is there ever a use case for this class?

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -357,7 +386,16 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
...
+  align-items: var(--pf-c-toolbar__content-section--AlignItems);
   width: 100%;
+
+  &.pf-m-align-items-center {
+    --pf-c-toolbar__content-section--AlignItems: center;
+  }
+
+  &.pf-m-align-items-baseline {
+    --pf-c-toolbar__content-section--AlignItems: baseline;
+  }
```
mcoker, line 398: Same question here as for the group - since it's the default, what's the use case for this class?

mcoker (changes requested):

mattnolting:
@mcoker added baseline mods back as children can inherit variable update

mcoker (commented):
Looks good, everything works well.

One potential issue - if everything is set to baseline by default, baseline is going to try and align with stuff around it, meaning "baseline" could result in "center" alignment (visually) if the other items in the layout are center aligned. Is that expected? https://codepen.io/mcoker/pen/QWxOJZQ

I wonder if the alignment properties should inherit. I don't think I would expect them to. Looking at the flex layout, the alignment mods redefine the regular property, not custom property, so the changes aren't inherited. That might be the most clean, tho if we want to manage everything with custom properties, we could use a `--base` var to set the default within each element that takes an alignment selector. Here's an example - https://codepen.io/mcoker/pen/jOKaQXx?editors=1100

Or in this PR, something like:

```
.pf-c-toolbar {
  --pf-c-toolbar__group--AlignItems--base: baseline;
}
.pf-c-toolbar__group {
  --pf-c-toolbar__group--AlignItems: var(--pf-c-toolbar__group--AlignItems--base);
  align-items: var(--pf-c-group__AlignItems);
  &.pf-m-center {
    --pf-c-toolbar__group--AlignItems: center;
  }
}
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -1,7 +1,7 @@
 // Don't remove this magic comment. See gulpfile.js.
 // @import "../../sass-utilities/all";
 $pf-c-toolbar--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl", "2xl");
-$pf-c-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
+$pf-c-toolbar--spacer-map: build-spacer-map();
```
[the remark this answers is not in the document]
mattnolting: No, not purposeful

mcoker (approved):
LGTM 🥳

srambach:
Two questions - what is the criteria for putting modifiers in the top section? Some are there but not all.
Second, is it worth making a demo of this, since it's not necessarily obvious how this alignment is useful in the case of error messages?

mattnolting:
> Two questions - what is the criteria for putting modifiers in the top section? Some are there but not all. Second, is it worth making a demo of this, since it's not necessarily obvious how this alignment is useful in the case of error messages?

@srambach I think the current alignment is a misimplementation. Center alignment would be the edge case for anything that doesn't share a height with selects, buttons, text inputs, etc. If we add a demo, it should highlight center alignment, rather than baseline. I would expect for that to be necessary for custom implementations, like including an icons, logos, text, etc.

Screenshot 2022-12-16 at 10 10 13 AM

Screenshot 2022-12-16 at 10 11 27 AM

srambach:
Is it expected that the "toolbar with validation desktop" example doesn't wrap and overflows? It happens at a relative wide width. (around 1245px)
Also, could you fix the title of that and the other desktop demo to be "on desktop" or something like that?
image

srambach (approved):
Nice job 👍

PR #5207, opened 2022-11-02
fix(pagination): added support for insets

### LINKED ISSUE #5090: Pagination - support ability to specify insets
Bottom pagination has padding defined that does not exist on the top/non-bottom pagination.

https://github.com/patternfly/patternfly/blob/fd83f75e967a6c0304cd911d6aaf5cd4d8b1294f/src/patternfly/components/Pagination/pagination.scss#L174

That really only works if the top pagination's toolbar is set to use page-insets. While page insets aren't set by default, the dev should set page insets on the toolbar when used in a page layout with a table/data-list. FWIW, if isn't set, then it looks like this by default. Note that the content on the left of the toolbar also isn't aligned.

Screen Shot 2022-09-13 at 5 27 16 PM

As reported by a user on slack, with toolbar insets removed, it looks like this. 

Screen Shot 2022-09-13 at 5 23 02 PM

That seems like a valid layout and there is no way to modify the pagination's insets currently. I think we should extend the ability to change the insets as needed.

blfetche:
closes #5090

### DIFF
--- a/src/patternfly/components/Pagination/examples/Pagination.md
+++ b/src/patternfly/components/Pagination/examples/Pagination.md
@@ -127,7 +127,19 @@ import './Pagination.css'
 {{/pagination}}
 ```
 
+### Inset
+```hbs
+{{#> pagination pagination--modifier="pf-m-inset-none pf-m-inset-md-on-md pf-m-inset-2xl-on-lg"}}
+  {{> pagination-total-items-content}}
+  {{> pagination-options-menu pagination-options-menu--id="pagination-options-menu-top-example"}}
+  {{> pagination-nav-content pagination-nav--aria-label="Pagination nav - top example"}}
+{{/pagination}}
+```
+
 ## Documentation
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-c-pagination` | Modifies pagination horizontal padding at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 Note: `<button>` or `<a>` elements can be used in `.pf-c-pagination__nav-page-select`.
 
 ### Accessibility
--- a/src/patternfly/components/Pagination/examples/Pagination.md
+++ b/src/patternfly/components/Pagination/examples/Pagination.md
@@ -137,21 +137,13 @@ import './Pagination.css'
 ```
 
 ## Documentation
-| Class | Applied to | Outcome |
-| -- | -- | -- |
-| `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-c-pagination` | Modifies pagination horizontal padding at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
 Note: `<button>` or `<a>` elements can be used in `.pf-c-pagination__nav-page-select`.
 
 ### Accessibility
 | Attribute | Applied to | Outcome |
 | -- | -- | -- |
 | `aria-label`  | `.pf-c-pagination__nav` |  Provides an accessible name for pagination navigation element. **Required** |
-
-## Pagination nav input
-
-| Attribute | Applied to | Outcome |
-| -- | -- | -- |
-| `type="number"` | `.pf-c-pagination` > `.pf-c-form-control` | Defines a field as a number. **Required** |
+| `type="number"` | `.pf-c-pagination__nav-page-select` > `.pf-c-form-control` | Defines a field as a number. **Required** |
 | `value` | `.pf-c-pagination__nav-page-select` > `.pf-c-form-control` | Provides initial integer value. **Required** |
 | `min` | `.pf-c-pagination__nav-page-select` > `.pf-c-form-control` | Provides minimum integer value. **Required** |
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Pagination/pagination.scss
```diff
@@ -293,6 +300,17 @@ $pf-c-pagination--breakpoint-map: build-breakpoint-map();
         --pf-c-pagination__total-items--Display: var(--pf-c-pagination--m-display-full__total-items--Display);
         --pf-c-pagination__total-items--Visibility: var(--pf-c-pagination--m-display-full__total-items--Visibility);
       }
+
+      /* &.pf-m-inset-md#{$breakpoint-name} {
+        padding-left: var(--pf-global--spacer--md);
+        padding-right: var(--pf-global--spacer--md);
+      } */
```
mcoker, line 307: Is this still needed?

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -1,6 +1,11 @@
 $pf-c-pagination--breakpoint-map: build-breakpoint-map();
+$pf-c-pagination--variable-map: build-spacer-map("none", "md", "xl");
```
mcoker, line 2: This should probably support all of the spacers

```suggestion
$pf-c-pagination--variable-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
```

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -1,6 +1,11 @@
...
 
 .pf-c-pagination {
```
mcoker, line 4: We should also add support for `.pf-m-page-insets`. You can look at the tabs component for a good example of how to do that. Here's the idea:

1. define the base page inset var and xl viewport inset var https://github.com/patternfly/patternfly/blob/1ead5fd152aa9ff95020036e1c8708cf01328205/src/patternfly/components/Tabs/tabs.scss#L15-L17
2. add an xl media query and update the base inset var to be set to the xl inset var
  https://github.com/patternfly/patternfly/blob/1ead5fd152aa9ff95020036e1c8708cf01328205/src/patternfly/components/Tabs/tabs.scss#L175-L178
3. then add a `.pf-m-page-insets` block where you set the padding to the page inset var.
  https://github.com/patternfly/patternfly/blob/1ead5fd152aa9ff95020036e1c8708cf01328205/src/patternfly/components/Tabs/tabs.scss#L467-L469

src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -127,7 +127,19 @@ import './Pagination.css'
 {{/pagination}}
 ```
 
+### Inset
+```hbs
+{{#> pagination pagination--modifier="pf-m-inset-none pf-m-inset-md-on-md pf-m-inset-2xl-on-lg"}}
```
mcoker, line 132: This example isn't currently working since the `2xl` inset isn't supported. If you update `$pf-c-pagination--variable-map` to support all of the spacers as mentioned in another comment, this example will work great!

src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -127,7 +127,19 @@ import './Pagination.css'
...
+  {{> pagination-total-items-content}}
+  {{> pagination-options-menu pagination-options-menu--id="pagination-options-menu-top-example"}}
+  {{> pagination-nav-content pagination-nav--aria-label="Pagination nav - top example"}}
+{{/pagination}}
+```
+
 ## Documentation
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-inset-{none, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-c-pagination` | Modifies pagination horizontal padding at optional [breakpoint](/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
```
mcoker, line 142: Can you move this line down into the "Usage" table? You can add it under `.pf-m-sticky` to group it with the `.pf-c-pagination` modifiers. Also if you add page insets, can you document that line in the table, too?

```suggestion
```

src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -127,7 +127,19 @@ import './Pagination.css'
...
 Note: `<button>` or `<a>` elements can be used in `.pf-c-pagination__nav-page-select`.
 
 ### Accessibility
```
mcoker, line 145: The docs here are a bit off, which is a pre-existing thing. Would you mind moving all of the stuff from the "Pagination nav input" table into this "accessibility" table? And  in `type="number" | .pf-c-pagination > .pf-c-form-control | Defines a field as a number. Required`, change `.pf-c-pagination` to `.pf-c-pagination__nav-page-select`.

mcoker (changes requested):

src/patternfly/components/Pagination/examples/Pagination.md
blfetche: Sure!

src/patternfly/components/Pagination/pagination.scss
blfetche: Removed

blfetche:
@mcoker all edits applied do we need line 110 in pagination.scss?

    --pf-c-pagination__scroll-button--Width: var(--pf-c-pagination__scroll-button--xl--Width);

srambach (approved):
LGTM, nice job

mcoker (approved):

srambach (approved):

PR #5211, opened 2022-11-02
fix(SearchInput/TextInputGroup): remove browser clear icons

### LINKED ISSUE #5210: Bug - Inputs: inconsistencies across browsers for search types
**Describe the problem**
When using an `<input type="search" />` (such as is being done in some React PRs, see https://github.com/patternfly/patternfly-react/pull/8040 and https://github.com/patternfly/patternfly-react/pull/8293, the latter reverted the type for now though), the input gets rendered differently.

In Chromium/Safari, an "x" icon will render inside the input, which can be clicked to clear it. In Firefox, however, no such icon gets rendered.

Rather than just using `type="text"` for search inputs, we should use `type="search"` since it'd be more semantically correct, and depending on the assistive tech it would get announced as a search text input.

A followup to this would also involve using the correct `type` for inputs/allowing the correct types to be used. Some notes:

- The dual list selector PR above added a `type` prop to the SearchInput component, which it may be better to only allow `type="search"` since we have other input components to pass a `type` into.
- The TextInput and TextInputGroupMain components in React defaults to the "text" type, but also allows a "search" type
    - For TextInputGroupMain, `type="text"` is hardcoded in a native input when the `hint` prop is passed
- Core and React have inconsistency in the markup of the SearchInput component. Core uses `pf-c-search-input` classes, while React uses the TextinputGroupMain under the hood

The following list shows currently where an input is being used, what `type` is being used, and whether a native input or one of our component's is being used. Most of this pertains to React implementation, but should be relatively 1:1 with Core (Core also has several checkbox inputs in Table demos)
<details>
<summary>Input uses cases in PatternFly</summary>

- Currently using type="text":
    - Label (for editable variant; native input element)
    - AdvancedSearchMenu (TextInput component; passes "text" into the `type` prop)
    - TextArea (only in examples; the `type` could be omitted)
    - TimePicker (TextInput component; passes "text" into the `type` prop)
    - PasswordStrength demo (TextInput component)
    - DateTimePicker demo (TextInput component; passes "text" into the `type` prop)
    - all 3 HelperText demos (TextInput for all 3; passes "text" into the `type` prop)
    - TableTextInput example (TableTextInput component; inline-edit extension)
    - FileUploadDemo (FileUpload component; for the demo-app in react-integration)
    - EditableTextCell (TextInput component; in react-table)
    - ViewToolbar (TextInput; in react-topology)
    - ClipboardCopy (uses TextInput's default of "text")
    - DatePicker (uses TextInput's default of "text")
    - FileUploadField (uses TextInput's default of "text")
    - TextInput examples (passes in "text" to `type` prop)
    - PasswordGenerator demot (TextInput; uses the default for the `type` prop)
    - TextInputGroupMain examples (the "disabled" example passes "text" into the `type` prop, the other examples do not; Core only uses "text" in examples)
    - ComposableTypeaheadSelect and ComposableMultiTypeaheadSelect (TextInputGroupMain; uses the default `type` prop of "text")
    - AttributeFiltering and AutoCompleteSearch demos for TextinputGroup (TextInputGroupMain; uses the default `type` prop of "text")
    - SelectMultiTypeahead and SelectTypeahead examples for Select Next (TextInputGroupMain; uses the default `type` prop of "text")
    - SearchInput (TextInputGroupMain; uses default `type` prop of "text", but can be overridden; Core only uses "text" in examples)
- Currently using type="search":
    - ApplicationLauncher (native input element; the above linked PR would change this)
    - ContextSelector (TextInput component)
    - DualListSelector (SearchInput element, see PR linked above)
    - TreeViewSearch (native input element)
    - PrimaryDetail (TextInput component, multiple times)
    - ComposableApplicationLauncher demo (TextInput component)
    - ComposableContextSelector demo (TextInput component)
    - MenuDemo (TextInput component; for the demo-app in react-integration)
    - ToolbarDemo (TextInput component; for the demo-app in react-integration)
- Currently using type="number":
    - CalendarMonth (TextInput component)
    - NumberInput (TextInput component)
    - Pagination > Navigation (native input element)
    - Slider (TextInput component)
    - FormDemo (FormGroup component; for the demo-app in react-integration)
    - TopologyPackage (TextInput component; in react-topology)
    - Tooltip examples
- Currently using type="email":
    - ModalWithForm example (TextInput component)
- Currently using type="checkbox":
    - Card (native input element)
    - Checkbox (native input element)
    - DataListCheck (native input element)
    - DropdownToggleCheckbox (native input element)
    - DualListSelectorTreeItem (native input element)
    - MenuToggleCheckbox (native input element)
    - SelectOption (native input element)
    - Switch (native input element)
    - TreeViewListItem (native input element)
- Currently using type="radio"
    - DataListItem (native input element)
    - Radio (native input element)
- Currently using multiple types
    - Select (uses "text" and "search" in different parts of the React code; native input element for all)
    - Button demo (uses "password" and "text"; TextInput component for both)
    - InputGroupDemo (uses multiple types, all on Textinput component; for the demo-app in react-integration)
    - SelectColumn (can use either "checkbox" or "radio"; native input element)
    - TextInputGroupMain (can use either "text" or have type passed in; native input element)
    - Form examples (uses multiple type; all using either FormGroup or TextInput components; FormGroup may be using `type` incorrectly)
    - InputGroup examples (uses multiple types, all using TextInput component)
    - LoginForm (uses either "text" or "password"; TextInput component)

</details>

**Screenshots**

Chrome (similar in Safari):
Chrome search input

Firefox:
Firefox search input

thatblindgeye:
closes #5210

I added the rules to both files because Core uses `pf-c-search-input` classes for the Search Input, but React uses Text Input Group under the hood (not sure if there's an issue open about making this more consistent, I didn't come across one). 

This [How to Remove “X” icon from search input field or input type search](https://medium.com/@rion.mrk/how-to-remove-x-icon-from-search-input-field-or-input-type-search-db3c808405fb) article is what I came across.

To test (use Chromium or Safari):

1. Go to [search input preview build](https://patternfly-pr-5211.surge.sh/components/search-input)
2. Update the first example so the `input` has `type="search"`
3. Enter text in the input and notice how an icon does not get rendered by the browser
4. Do steps 2-3 on the [HTML search input](https://www.patternfly.org/v4/components/search-input/html) on the live site, and notice how an icon does get rendered

### DIFF
--- a/src/patternfly/components/SearchInput/search-input.scss
+++ b/src/patternfly/components/SearchInput/search-input.scss
@@ -122,6 +122,13 @@
   padding: var(--pf-c-search-input__text-input--PaddingTop) var(--pf-c-search-input__text-input--PaddingRight) var(--pf-c-search-input__text-input--PaddingBottom) var(--pf-c-search-input__text-input--PaddingLeft);
   border: 0;
 
+  &[type="search"]::-webkit-search-decoration,
+  &[type="search"]::-webkit-search-cancel-button,
+  &[type="search"]::-webkit-search-results-button,
+  &[type="search"]::-webkit-search-results-decoration {
+    display: none;
+  }
+
   &,
   &.pf-m-hint {
     grid-area: text-input;
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -157,6 +157,13 @@
   padding: var(--pf-c-text-input-group__text-input--PaddingTop) var(--pf-c-text-input-group__text-input--PaddingRight) var(--pf-c-text-input-group__text-input--PaddingBottom) var(--pf-c-text-input-group__text-input--PaddingLeft);
   border: 0;
 
+  &[type="search"]::-webkit-search-decoration,
+  &[type="search"]::-webkit-search-cancel-button,
+  &[type="search"]::-webkit-search-results-button,
+  &[type="search"]::-webkit-search-results-decoration {
+    display: none;
+  }
+
   &,
   &.pf-m-hint {
     grid-area: text-input;


mcoker:
This looks great. I opened an issue to replace our use of the search input with the text input group. https://github.com/patternfly/patternfly/issues/5212

On this, we show examples of search inputs without without our clear button (a use case may be when there is not enough space to show it) - in those cases, would we still want to remove the browser generated clear button? Even if it's inconsistent, I know some people use particular browsers for their UI with form elements like this.

Related, what about the form control (`<TextInput>`)? Seems like the same use case, except the text input and search input components take an `onClear` event that will create the clear button, and the form control doesn't - it would need to be manually paired with a clear button in an input group  - https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/FormControl/examples/FormControl.md#icon-sprite

thatblindgeye:
> On this, we show examples of search inputs without without our clear button (a use case may be when there is not enough space to show it) - in those cases, would we still want to remove the browser generated clear button? Even if it's inconsistent, I know some people use particular browsers for their UI with form elements like this.

Yeah it's tricky for when the clear button provided by PatternFly may need to be omitted due to spacing. Some possiblities:

- Remove the browser clear button + adjust the styles for text input group so that at smaller widths, everything renders correctly (right now `pf-c-text-input-group` at around ~175px width seems to be about as small as you can go before the clear button for SearchInput starts creeping visually into the input (image as at 150px width):
    Search input with 150 pixel width
- Have these rules as part of a class that can be applied by the consumer. This could then be added in React only when the search input's clear button is rendered, though in Core it would more be up to the consumer to ensure the dual clear button behavior doesn't occur
- Only have the browser rendered button removed completely (what this PR currently does)
- Other (if there are other ideas out there)

At a minimum I think it'd be ideal to at least prevent the browser clear + our clear from both rendering (which option 2 would help resolve). _Personally_ I would opt to also prevent the browser one from rendering and having the TextInputGroup's/SearchInput's used, since it would provide everyone (regardless of browser) a means to immediately clear the input, make things more consistent across browsers, and make input elements more consistent (rather than only some inputs having a browser rendered clear button, and others not having one). That said, it could very easily skirt the line of being something nobody really notices, or a significant change for users.

> Related, what about the form control (<TextInput>)? Seems like the same use case, except the text input and search input components take an onClear event that will create the clear button, and the form control doesn't - it would need to be manually paired with a clear button in an input group 

Depending what route we'd want to go down, we could also add these styles to `pf-c-form-control`.

mcoker:
> At a minimum I think it'd be ideal to at least prevent the browser clear + our clear from both rendering (which option 2 would help resolve). Personally I would opt to also prevent the browser one from rendering and having the TextInputGroup's/SearchInput's used, since it would provide everyone (regardless of browser) a means to immediately clear the input, make things more consistent across browsers, and make input elements more consistent (rather than only some inputs having a browser rendered clear button, and others not having one).

@thatblindgeye I agree, #2 seems like a good option. An issue is a simple `<TextInput>` - I think that's pretty commonly used, and to get a close button, you'd need to use the `<TextInputGroup>` or something more complex. Personally, I'd be in favor of making the `<TextInputGroup>` the default way of creating a text input as long as we're adding all of these customizations. It gives us the structure/ability to customize and attach all of this stuff to the text input that you can't do with a simple `<input type="text">` element. wdyt @srambach @mattnolting?

mcoker:
Re: a class, I think it could go a couple of ways.

* A class whose job is to remove the clear icon, eg `.pf-m-no-clear-icon`.
* A `.pf-m-close` class on the text input group, assuming that's our suggested way of using a text input to create a clear button (moving away from a simple input group with a `<TextInput>` + `<Button>`). This class is a bit more versatile, and could apply the styling to remove the browser's clear button and manage the presence of the PF clear button/layout in other ways that work better for when the text input group has a clear button attached to the input. For example, maybe some layout rework to handle the narrow viewport/overlap bug you mentioned above. My vote would be for this.

Then we could maybe add a class to the `<TextInput>` that removes the browser icon when either a user has used a simple `<InputGroup>` to create a clear button, or for some other use.

srambach:
If I read it correctly, then this is removing the browser clear (which appears in safari/chrome but not in FF) in favor of manually adding a clear button? In this particular case, I'd favor consistency within a browser over consistency across browsers. Altering default browser behavior can be risky with no guarantee of future behavior.

mattnolting:
In testing w/VO, I'm not seeing a huge benefit to updating `type` to `search`, especially bc the field is described by an `aria-label`. It seems most reasonable, given all the extras provided by the affected components (close button, result count, etc), to use only `type="text"` and update the `aria-label` to something like `Search by name`. Best to avoid, wherever possible, changing specific browser default functionality, as future updates create unpredictable results.  @mcoker @srambach wdyt?

thatblindgeye:
>If I read it correctly, then this is removing the browser clear (which appears in safari/chrome but not in FF) in favor of manually adding a clear button? In this particular case, I'd favor consistency within a browser over consistency across browsers. Altering default browser behavior can be risky with no guarantee of future behavior.

@srambach That's a really good point! Maybe instead of tinkering with any of this, we can just recommend having our clear button rendered when necessary via examples and/or documentation?

>In testing w/VO, I'm not seeing a huge benefit to updating type to search, especially bc the field is described by an aria-label. It seems most reasonable, given all the extras provided by the affected components (close button, result count, etc), to use only type="text" and update the aria-label to something like Search by name. Best to avoid, wherever possible, changing specific browser default functionality, as future updates create unpredictable results.

@mattnolting Yeah with VO the only difference I think is "search text field" being announced when type="search" vs "edit text" (NVDA and JAWS for me I didn't notice any difference). aria-labels could help differentiate, and the React examples do use "Search input" as the default aria-label. Based on JAWS and NVDA, at least for me, not signifying the type of input like VO does, I think the only pros I can think of using type="search" is 1) it's slightly more accurate (though in 2/3 of screen readers that accuracy doesn't seem to do much), and 2) it may replace the typical "Enter" symbol on touch keyboards with a magnifying glass.

There have already been some updates in React that update instances of Search Input to use `type="search"`, so depending where we'd land those may have to be reverted.

thatblindgeye:
Closing per discussion in the linked issue.

Issue #5139, opened 2022-10-03
Add popover for enabled tab


yannnz:
### **Requirement**
The RHOAD design team has come across a use case that requires a popover in an enabled tab to provide additional description of the tab itself. The use case here is the Rulesets tab. 

There are two proposed options to have the popover popped up: hovering or clicking the question-circle icon. 

### **Design samples**
![2](https://user-images.githubusercontent.com/10469467/193600352-fb76a1e1-f263-42f8-af72-625f1e129384.png)

PR #5213, opened 2022-11-03
feat(tabs): added help popover, refactored tab action

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5139

Convenience link - https://patternfly-pr-5213.surge.sh/components/tabs#tab-item-actions

* refactors the close and help buttons to be generic tab actions - so we can add more things like this, or open it up to users to add their own actions. Renames the close class and vars, but this is beta in both core and react so should be OK. I can rework that if needed and build help as its own class w/ vars.
* Decreases the spacing a bit after the tab text if it has an action. This lets me easier adjust the spacing between the help button and tab text to be less than the close button and help text. I can rework this if needed.
  * before
    www patternfly org_v4_
  * after
    localhost_8001_components_tabs
* updates the examples page a bit to add titles for the sections and groups a bunch of common handlebars partial attrs to be defined on a single block level instead of per hbs partial call

Here are some outlines that show the actual spacing per clickable thing in the tab

* Help
  localhost_8001_components_tabs (1)

* Close
  localhost_8001_components_tabs (2)

* Help and close
  localhost_8001_components_tabs (3)

### DIFF
--- a/src/patternfly/components/Button/button.hbs
+++ b/src/patternfly/components/Button/button.hbs
@@ -17,6 +17,12 @@
   {{#if button--IsInlineLinkSpan}}
     role="button"
   {{/if}}
+  {{#if button--aria-label}}
+    aria-label="{{button--aria-label}}"
+  {{/if}}
+  {{#if button--IsDisabled}}
+    disabled
+  {{/if}}
   {{#if button--attribute}}
     {{{button--attribute}}}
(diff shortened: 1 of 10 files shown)


kmcfaul (approved):
lgtm just needs to pass the build

mmenestr (approved):
LGTM also!

Issue #5091, opened 2022-09-14
Notification drawer - menus are cut off unnecessarily 


mcoker:
<img width="1214" alt="Screen Shot 2022-09-14 at 9 43 44 AM" src="https://user-images.githubusercontent.com/35148959/190186972-83bfc8d3-e951-41ea-be5c-04c721dd75e6.png">

It seems like the `__body` element needs to grow by default since it also scrolls its overflow. I'm wondering if this has always been a problem, or if `flex-grow` was removed for some other reason and adding it back might cause an issue?

nicolethoen:
Closed by #5214

PR #5214, opened 2022-11-03
fix(notification-drawer): set body to fill available vertical space

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5091


mattnolting (approved):
LPTM!

PR #5216, opened 2022-11-04
refactor(about-modal): moved to modal-box

### LINKED ISSUE #4526: Refactor About Modal to use the Modal component
- use the Modal component to wrap the About Modal (so that the contents of the about modal is more like the wizard which will fill a modal)
- There will be an about modal in-line in the core examples, and a new about modal demo which places the about modal in a Modal component
- the width/height restraints should be removed and retained in the modal

This is similar to how the wizard is used in the modal

Note: the react examples with combine the Modal with the About modal contents and ship it at one

mattnolting:
closes #4526

### DIFF
--- a/src/patternfly/components/AboutModalBox/about-modal-box.scss
+++ b/src/patternfly/components/AboutModalBox/about-modal-box.scss
@@ -4,7 +4,6 @@
   --pf-c-about-modal-box--Height: 100%;
   --pf-c-about-modal-box--lg--Height: #{pf-size-prem(762px)}; // Height is optimized for the exact height desired
   --pf-c-about-modal-box--Width: 100%;
-  --pf-c-about-modal-box--MaxWidth: #{pf-size-prem(1232px)};
   --pf-c-about-modal-box--sm--grid-template-columns: 5fr 1fr;
   --pf-c-about-modal-box--lg--grid-template-columns: 1fr .6fr;
 
@@ -107,7 +106,6 @@
     "header header"
     "content content";
   width: var(--pf-c-about-modal-box--Width);
-  max-width: var(--pf-c-about-modal-box--MaxWidth);
   height: var(--pf-c-about-modal-box--Height);
   overflow-x: hidden;
   overflow-y: auto;
@@ -127,7 +125,6 @@
 
     grid-template-columns: var(--pf-c-about-modal-box--lg--grid-template-columns);
     grid-template-rows: max-content max-content auto;
-    max-width: var(--pf-c-about-modal-box--lg--MaxWidth);
   }
 }
 
--- a/src/patternfly/components/AboutModalBox/about-modal-box.scss
+++ b/src/patternfly/components/AboutModalBox/about-modal-box.scss
@@ -2,30 +2,8 @@
   // Component variables
   --pf-c-about-modal-box--BackgroundColor: var(--pf-global--palette--black-1000); // Modal uses a non-standard background color
   --pf-c-about-modal-box--BoxShadow: 0 0 100px 0 rgba(255, 255, 255, .05);  // This is a special one-off glow for the about modal
-  --pf-c-about-modal-box--ZIndex: var(--pf-global--ZIndex--xl);
-  --pf-c-about-modal-box--Height: 100%;
-  --pf-c-about-modal-box--lg--Height: #{pf-size-prem(762px)}; // Height is optimized for the exact height desired
-  --pf-c-about-modal-box--lg--MaxHeight: calc(100% - var(--pf-global--spacer--xl)); // MaxHeight ensures that the modal will not go off the screen and instead the content will scroll
-  --pf-c-about-modal-box--Width: 100vw;
-  --pf-c-about-modal-box--lg--Width: calc(100% - (var(--pf-global--spacer--3xl) * 2)); // Modal should be full width with a gutter on either side until it hits a max-width
-  --pf-c-about-modal-box--lg--MaxWidth: #{pf-size-prem(1232px)};
-  --pf-c-about-modal-box--PaddingTop: var(--pf-global--spacer--xl);
-  --pf-c-about-modal-box--PaddingRight: var(--pf-global--spacer--xl);
-  --pf-c-about-modal-box--PaddingBottom: var(--pf-global--spacer--xl);
-  --pf-c-about-modal-box--PaddingLeft: var(--pf-global--spacer--xl);
-  --pf-c-about-modal-box--sm--PaddingTop: var(--pf-global--spacer--3xl);
-  --pf-c-about-modal-box--sm--PaddingRight: var(--pf-global--spacer--3xl);
-  --pf-c-about-modal-box--sm--PaddingBottom: var(--pf-global--spacer--3xl);
-  --pf-c-about-modal-box--sm--PaddingLeft: var(--pf-global--spacer--3xl);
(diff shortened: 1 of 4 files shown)


src/patternfly/components/AboutModalBox/examples/AboutModalBox.md
```diff
@@ -6,11 +6,10 @@ cssPrefix: pf-c-about-modal-box
 
 ## Examples
 ### Basic
-```hbs isFullscreen
-{{#> about-modal-box about-modal-box--attribute='aria-labelledby="about-modal-title"'}}
+```hbs
```
mcoker, line 9: This should probably still be full-screen. There is some wonky parallax action going on with the background image as you scroll the example page.

```suggestion
```hbs isFullscreen
```

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -2,31 +2,9 @@
   // Component variables
   --pf-c-about-modal-box--BackgroundColor: var(--pf-global--palette--black-1000); // Modal uses a non-standard background color
   --pf-c-about-modal-box--BoxShadow: 0 0 100px 0 rgba(255, 255, 255, .05);  // This is a special one-off glow for the about modal
```
mcoker, line 4: @mceledonia do you see any reason to keep this box shadow on the about modal around? it's a 100px spread, white w/ 50% opacity shadow. It's barely visible, and if we're moving the the about modal into the modal component, the modal has its own box shadow.

The current about modal w/ 100px shadow

Screen Shot 2022-11-07 at 5 10 32 PM

No shadow, just to compare

Screen Shot 2022-11-07 at 5 10 36 PM

using the modal component's default shadow

Screen Shot 2022-11-07 at 5 11 25 PM

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -2,31 +2,9 @@
...
-  --pf-c-about-modal-box--ZIndex: var(--pf-global--ZIndex--xl);
-  --pf-c-about-modal-box--Height: 100%;
-  --pf-c-about-modal-box--lg--Height: #{pf-size-prem(762px)}; // Height is optimized for the exact height desired
-  --pf-c-about-modal-box--lg--MaxHeight: calc(100% - var(--pf-global--spacer--xl)); // MaxHeight ensures that the modal will not go off the screen and instead the content will scroll
```
mcoker, line 8: Hmmm... removing these heights is going to make the about modal shrink to fit its contents. Maybe that's a good thing? We had this use case when we moved the wizard into the modal component, too, and we ended up giving it a height when it's inside of a modal. https://github.com/patternfly/patternfly/blob/7857f7ed377cc40bdf7721a1c9b710b7f34258a0/src/patternfly/components/Wizard/wizard.scss#L207-L211

This is what the about modal looks like now on org with minimal content:

Screen Shot 2022-11-07 at 5 21 24 PM

And this is what it looks like in this PR:

Screen Shot 2022-11-07 at 5 21 52 PM

@mcarrano @mceledonia @mmenestr wdyt?

mcoker (commented):
Looks good, have a couple of questions for design.

src/patternfly/components/AboutModalBox/about-modal-box.scss
mceledonia: Nope good catch, let's drop it!

src/patternfly/components/AboutModalBox/about-modal-box.scss
mceledonia: I prefer having the negative space even with limited content, it provides a nice balance which gets skewed when the height is set to fit.

mcarrano: Agree

mcoker: @mceledonia @mcarrano will this component always be inside of a modal, or do you see a use case where it may be outside of a modal - either now or in the future with some other kind of component that can present it?

mcarrano: I suppose you could have an "About page," but I don't really see a need for this. A simple modal should suffice in most cases. @mceledonia @mmenestr do you agree?

mcoker:
@mattnolting can you rebase this against the v5 branch and update the base branch to v5?

src/patternfly/components/AboutModalBox/about-modal-box.scss
mattnolting: Height retained

mattnolting:
@mcoker rebased and ready

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -148,7 +128,6 @@
     grid-template-columns: var(--pf-c-about-modal-box--lg--grid-template-columns);
     grid-template-rows: max-content max-content auto;
     max-width: var(--pf-c-about-modal-box--lg--MaxWidth);
```
mcoker, line 130:
```suggestion
```

looks like you can remove this, it's currently breaking the `max-width` on lg screens. Although I think this begs the question of whether the about modal should manage its width at all, or if it should just consume the available space and be controlled by the modal. I'd lean toward the latter - wdyt @srambach @mceledonia @mcarrano?

fwiw if you fix this and let the `max-width` apply, it looks like this in a modal with the modal width set to 50% of the viewport.

Screen Shot 2022-11-09 at 12 24 05 PM

mcoker (changes requested):

srambach (approved):
LGTM.

src/patternfly/components/AboutModalBox/about-modal-box.scss
mcarrano: Not sure I completely understand the question @mcoker . What's there now looks good to me. The model should have a max-width IMO. Maybe I'm struggling to differentiate between Modal and About modal.

mattnolting: @mcoker updated

mcoker: @mcarrano sorry, it's definitely confusing. The _about modal_ has a custom width and height set, since it is it's own implementation of a modal. With this update, I was proposing we remove the _about modal_'s width, and just let it fill the size of whatever container it's in, which would mean the _modal_ component is used to manage the _about modal_'s width. The _modal_ has width variations for small, medium, large, and a "custom" width.

src/patternfly/demos/AboutModal/about-modal-template.hbs
```diff
@@ -1,47 +1,45 @@
...
-            Trademark and copyright information here
-          {{/about-modal-box-strapline}}
-        {{/about-modal-box-content}}
-      {{/about-modal-box}}
-    {{/bullseye}}
-  {{/backdrop}}
-{{/about-modal}}
+{{#> modal-template}}
+  {{#> modal-box modal-box--modifier="pf-m-lg" modal-box--attribute='aria-labelledby="about-modal-title"'}}
+    {{#> about-modal-box about-modal-box--attribute='aria-labelledby="about-modal-title"'}}
```
mcoker, line 3:
```suggestion
    {{#> about-modal-box}}
```

mcoker (changes requested):
Nice! Just one small update.

mcoker (approved):

PR #5220, opened 2022-11-07
fix(icons): moved svg classes to _icons.scss

### LINKED ISSUE #5218: Bug - pf-svg classes not included in react-styles
**Describe the problem**
The CI build is failing in https://github.com/patternfly/patternfly-react/pull/5275 due to [these missing css classes ](https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/patternfly-icons.scss#L4-L30) which do not get pulled into `react-styles`.

This is because `patternfly-base.scss` doesn't include them, as they're not in the import chain.

https://patternfly-react-pr-5275.surge.sh/components/button#variant-examples
The button variants that include icons are oversized due to the missing styles.

jenny-s51:
Fixes bug that was preventing these css classes from being pulled into `patternfly.css`. Closes #5218 

Before:
Screen Shot 2022-11-07 at 2 13 00 PM

After:
Screen Shot 2022-11-07 at 2 08 52 PM


mcoker (approved):
LGTM! I still wonder if we need these, and where they should live. The main problem as I see it is that the sizes don't align with core. We have icon size global vars that could be used instead, though they don't match these sizes. If they used the same vars/sizes, if a user customized their icon size vars, PF's use of icons in components would match their one-off uses of icons with size modifiers easier. For example, if we use a large icon for something in a component (core's font awesome or react's react-icons), and they use the large modifier here, if they update `--pf-whatever--icon--lg--size`, it will update everything.

Since we're breaking stuff, should they align? 

https://github.com/patternfly/patternfly/blob/aa5cdbd57fd58096332933d9d7af60dc5586f3f3/src/patternfly/base/_variables.scss#L235-L238

https://github.com/patternfly/patternfly/blob/aa5cdbd57fd58096332933d9d7af60dc5586f3f3/src/patternfly/sass-utilities/scss-variables.scss#L193-L196

FWIW we'll be updating the icon font sizes with the global var rework, we can reconsider this then.

PR #5221, opened 2022-11-07
feat(sidebar): added padding and border to sidebar

### LINKED ISSUE #5131: Sidebar - panel/content should support padding
From https://github.com/patternfly/patternfly/pull/5088/files, we're adding page sections in the sidebar to create padding - while that works, we should probably rethink that. 

The sidebar should support padding/insets with body elements like similar components (Drawer)

Seems like having a vertical divider between sections is also going to be a common enough use case that the component should likely support that as well.

mattnolting:
closes #5131

### DIFF
--- a/src/patternfly/components/Sidebar/examples/Sidebar.md
+++ b/src/patternfly/components/Sidebar/examples/Sidebar.md
@@ -208,7 +208,7 @@ import './Sidebar.css'
 | `.pf-c-sidebar__main` | `<div>` | Initiates the sidebar main element. **Required** |
 | `.pf-c-sidebar__panel` | `<div>` | Initiates the sidebar panel element. **Required** |
 | `.pf-c-sidebar__content` | `<div>` | Initiates the sidebar content element. **Required** |
-| `.pf-m-border` | `.pf-c-sidebar__panel`, `.pf-c-sidebar__content` | Modifies the sidebar component to add a border between the panel and content. **Note:** Does not work with `.pf-c-sidebar__panel.pf-m-static`. |
+| `.pf-m-border` | `.pf-c-sidebar__main` | Modifies the sidebar component to add a border between the panel and content. |
 | `.pf-m-padding` | `.pf-c-sidebar__panel`, `.pf-c-sidebar__content` | Modifies the sidebar component to add a padding to panel and/or content. |
 | `.pf-m-gutter` | `.pf-c-sidebar` | Modifies the sidebar component to add a gutter between the panel and content. |
 | `.pf-m-stack` | `.pf-c-sidebar` | Modifies the sidebar to stack the panel on top of the content. |
--- a/src/patternfly/components/Sidebar/examples/Sidebar.md
+++ b/src/patternfly/components/Sidebar/examples/Sidebar.md
@@ -151,6 +151,54 @@ import './Sidebar.css'
 {{/sidebar}}
 ```
 
+### Border
+```hbs
+{{#> sidebar sidebar--modifier="pf-m-gutter" sidebar--HasBorder=true}}
+  {{#> sidebar-panel}}
+    Sidebar panel
+  {{/sidebar-panel}}
+  {{#> sidebar-content}}
+    {{#> content}}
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Sidebar/examples/Sidebar.md
```diff
@@ -29,7 +29,7 @@ import './Sidebar.css'
   {{#> sidebar-panel}}
     Sidebar panel. Adding some extra content so that the gap between the panel and content area is better illustrated in this example.
   {{/sidebar-panel}}
-  {{#> sidebar-content}}
+  {{#> sidebar-content sidebar-content--modifier="pf-m-bordersss"}}
```
srambach, line 32: Do you mean `pf-m-border` on line 16 and 32?

mattnolting: Oof, nope! Good catch, thanks

mattnolting: @srambach updated!

mcoker (commented):
The padding looks great. The border is triggering unnecessary scroll in all directions because of the negative offset applied based on the gutter size.

I think there are a handful of ways to handle this, and mentioned some offline. AFAIK this isn't a pressing issue (I opened it), and I think it might be worth reconsidering the approach and doing so with v5 since I think we'll likely change the gutter system anyways and probably need to rework this anyways. The same goes for the padding, potentially. WDYT @mcarrano, do you see a need to get this in for v4 (or some part of it - like the padding and/or border between panel and content), or would it be ok to defer to v5?

srambach:
I'm still seeing a phantom scrollbar I think? 

https://user-images.githubusercontent.com/19825616/208736781-640d9ee5-20b0-494a-8597-7e790377f6ab.mp4

srambach (approved):
Examples look good now 👍

wise-king-sullyman (approved):

thatblindgeye (approved):

src/patternfly/components/Sidebar/sidebar.scss
```diff
@@ -152,11 +181,24 @@ $pf-c-sidebar__panel--list--width: ("default", 25, 33, 50, 66, 75, 100);
...
+
   &.pf-m-no-background {
     --pf-c-sidebar__content--BackgroundColor: transparent;
   }
+
+  & + .pf-c-sidebar__panel {
+    --pf-c-sidebar__panel--Order: 2;
+  }
+
+  :where(&:first-child) {
```
mcoker, line 199: I wonder if it would be more flexible just to have the panel always on the left, and if you want it on the right (either by the class modifier or in the markup), you use `.pf-m-panel-right`. Looking at it again, we allow the panel on the right when the panel is first in the markup, but we don't allow the panel on the left when the panel is last in the markup. Using `.pf-m-panel-right` alone to switch the order would allow for that. wdyt?

src/patternfly/components/Sidebar/examples/Sidebar.md
```diff
@@ -160,6 +208,8 @@ import './Sidebar.css'
 | `.pf-c-sidebar__main` | `<div>` | Initiates the sidebar main element. **Required** |
 | `.pf-c-sidebar__panel` | `<div>` | Initiates the sidebar panel element. **Required** |
 | `.pf-c-sidebar__content` | `<div>` | Initiates the sidebar content element. **Required** |
+| `.pf-m-border` | `.pf-c-sidebar__panel`, `.pf-c-sidebar__content` | Modifies the sidebar component to add a border between the panel and content. **Note:** Does not work with `.pf-c-sidebar__panel.pf-m-static`. |
```
mcoker, line 211:
```suggestion
| `.pf-m-border` | `.pf-c-sidebar__main` | Modifies the sidebar component to add a border between the panel and content. |
```

mcoker (changes requested):

srambach (approved):

mcoker (approved):

Issue #5141, opened 2022-09-21
Bug - Radio - Radio body width is limited when the TextInput is set as the body


DaoDaoNoCode:
**Describe the problem**
When you set `TextInput` as the body of Radio, the width will be limited, and it's not stretching like when you set some text as the body.

**How do you reproduce the problem?**
Check this codesandbox example https://codesandbox.io/s/loving-keller-6odpsp

**Expected behavior**
The `TextInput` in the body could extend and has the full width.

**Is this issue blocking you?**
No, but I may need to overwrite some PF styles, any better workaround?

**Screenshots**
<img width="843" alt="Screen Shot 2022-09-21 at 10 05 08 AM" src="https://user-images.githubusercontent.com/37624318/191525892-61a72e07-c95b-493f-8e48-382c9ac32e22.png">

**What is your product and what release date are you targeting?**
OpenDataHub

**Any other information?**

mcoker:
Hey @DaoDaoNoCode, I've added this to our major release milestone to fix on the CSS side as I think it will be a breaking change. For now can you apply some custom CSS to fix that? I updated your codesandbox, which just tells the radio body element to span the full width of the layout via `justify-self: stretch` - https://codesandbox.io/s/goofy-smoke-u7vcjj?file=/style.css

---

For more context when we work on this, we have `[align/justify]-items: start` on the parent grid, which is keeping the grid items from growing to allow the items' children to grow. This may be on purpose, but we should investigate to see if they're needed and update as needed so we can support the use case above.

https://github.com/patternfly/patternfly/blob/02bda75f8f22a9e230be216ed2e6e8ffd5b48cc3/src/patternfly/components/Check/check.scss#L19-L23

https://github.com/patternfly/patternfly/blob/02bda75f8f22a9e230be216ed2e6e8ffd5b48cc3/src/patternfly/components/Radio/radio.scss#L16-L20

DaoDaoNoCode:
@mcoker I see, Thanks for the suggestion!

PR #5233, opened 2022-11-11
fix(check/radio): set description/body to fill available width

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5141


srambach (approved):
LGTM 👍

PR #5235, opened 2022-11-14
feat(Wizard): review step component styles

jeffpuzzo:
Relates to: https://github.com/patternfly/patternfly-react/issues/7916

Result of changes inside of a patternfly-react example:
Mixed expandable and non-expandable (with expanded state):
image

Mixed expandable and non-expandable (with collapsed state):
image

All non-expandable
image

All expandable and in collapsed state:
image

All expandable and in expanded state:
image


mcarrano (approved):
Looks great. Thanks for making all these adjustments @jeffpuzzo !

PR #5242, opened 2022-11-16
fix(formcontrol): fixed form control select dark theme icon

### LINKED ISSUE #5215: Bug - [Form Control Select] - [Icon is incorrect in dark mode]
**Describe the problem**
A clear and concise description of the problem. Which components are affected?

1. Visit https://pf4.patternfly.org/components/form-control/#select
2. Switch to dark mode and see that the select icon changes into a checkmark

We observe this problem in cockpit with the following versions:

```
    "@patternfly/patternfly": "4.219.2",
    "@patternfly/react-core": "4.258.3",
    "@patternfly/react-icons": "4.92.10",
    "@patternfly/react-styles": "4.91.10",
    "@patternfly/react-table": "4.111.33",
```

**Expected behavior**

It should show a white triangle, such in other selects in dark mode

We have a workaround in Cockpit.

**Screenshots**

**What is your environment?**
 - Chrome 107.0.5304.87-1

Cockpit

evwilkin:
Closes #5215 

Previously we were hard coding the incorrect icon in for `--pf-c-form-control__select--BackgroundUrl` in dark theme, but the icon does not change - only the color needs to change.  This PR uses the same `pf-bg-svg` SASS function in dark theme but passes in the third argument (`$svg-color`) to override the color.

Note: The color looks correct now, but I'm unclear of if there's a better variable to use (or new one to create) for this usage.

### DIFF
--- a/src/patternfly/components/FormControl/themes/dark/form-control.scss
+++ b/src/patternfly/components/FormControl/themes/dark/form-control.scss
@@ -2,7 +2,7 @@
 
 @mixin pf-theme-dark-form-control() {
   .pf-c-form-control {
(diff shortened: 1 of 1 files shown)


mcoker (approved):
Looks perfect to me! FWIW, it's because of this:

https://github.com/patternfly/patternfly/blob/5eee883c9aadfa4834798de2ee3c00dfa8a37a6a/src/patternfly/sass-utilities/functions.scss#L18-L22

`currentColor` will not work with a `background-image` as far as I can tell, so that default doesn't make sense. Also we're forcing `%23` (`#`) at the beginning of the `fill` attribute, so if no color is passed, it creates `fill="%23urrentColor` 🤦‍♂️ Might be a fun one to try and fix but hopefully we can get away from these background images on form elements entirely.

PR #5243, opened 2022-11-16
chore(flex,toolbar): use gap for space-items

srambach:
fixes https://github.com/patternfly/patternfly/issues/3348

This changes pf-m-space-items to use gap instead of margins. Individual spacing is still done with margins. The two are not compatible, so I'll need to update the docs as well if this is what we want to do.

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -16,6 +16,8 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
 
   // Group
   --pf-c-toolbar__group--Display: flex;
+  --pf-c-toolbar__group--RowGap: 0;
+  --pf-c-toolbar__group--ColumnGap: 0;
 
   // Sticky
   --pf-c-toolbar--m-sticky--ZIndex: var(--pf-global--ZIndex--xs);
@@ -210,6 +212,8 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
 
   align-items: center;
   margin-right: var(--pf-c-toolbar--spacer);
+  row-gap: var(--pf-c-toolbar__group--RowGap);
+  column-gap: var(--pf-c-toolbar__group--ColumnGap);
 
   // Button group
   &.pf-m-button-group {
@@ -508,11 +512,10 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
     @include pf-apply-breakpoint($breakpoint) {
       @each $spacer, $spacer-value in $pf-c-toolbar--spacer-map {
         .pf-m-space-items-#{$spacer}#{$breakpoint-name} {
-          > * {
-            --pf-c-toolbar--spacer: #{$spacer-value};
-          }
+          --pf-c-toolbar__group--RowGap: #{$spacer-value};
+          --pf-c-toolbar__group--ColumnGap: #{$spacer-value};
 
-          > :last-child {
+          > * {
             --pf-c-toolbar--spacer: 0;
           }
         }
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -210,6 +212,8 @@ $pf-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl"
 
   align-items: center;
   margin-right: var(--pf-c-toolbar--spacer);
+  row-gap: var(--pf-c-toolbar__group--RowGap);
+  column-gap: var(--pf-c-toolbar__group--ColumnGap);
```
mcoker, line 216: You could combine these these as:

```suggestion
  gap: var(--pf-c-toolbar__group--RowGap) var(--pf-c-toolbar__group--ColumnGap);
```

We do that for other shorthand properties like `border`,`padding`, `margin`, etc

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -21,6 +23,7 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
   display: var(--pf-l-flex--Display);
   flex-wrap: var(--pf-l-flex--FlexWrap);
   align-items: var(--pf-l-flex--AlignItems);
+  gap: var(--pf-l-flex--RowGap, --pf-l-flex--ColumnGap);
```
mcoker, line 26: did you mean to do it this way? 

```suggestion
  gap: var(--pf-l-flex--RowGap) var(--pf-l-flex--ColumnGap);
```

The way it's written now, `--pf-l-flex--ColumnGap` is a fallback for a single value being passed to `gap` (a single value will always evaluate as a `row-gap`), and since it isn't wrapped in `var()`, it's just an invalid value. You can see here that setting `--pf-l-flex--RowGap: initial`, which will force the fallback value to be used, results in no gap at all.

Screenshot 2022-11-18 at 7 08 11 PM

mcoker (commented):
This works for establishing a gap, but breaks when a layout with `.pf-m-space-items` has children with `.pf-m-spacer`. From the docs:

> Applying .pf-m-space-items-{size} to .pf-l-flex will override css variable values for direct descendants, excluding last child. This spacing can be overridden for direct descendant with .pf-m-spacer-{size}.

Off the top of my head, I'm not sure how you would use a `row-gap` in combination with margins on the children to offset the parent row gap, specifically for layouts where 1) items wrap and 2) when the item spacer is less than the row gap (giving the item a negative margin), without the item will go outside of bounds at the edge where it wraps.

This is where something like `:has()` would be useful - https://codepen.io/mcoker/pen/JjZOwNX

src/patternfly/layouts/Flex/flex.scss
srambach: Yeah, I understand fallbacks; I just made a mistake.

mattnolting (commented):
I think there we could preserve the individual spacing modifiers and enable `gap`. I was thinking something like this.

## Add modifiers
- `.pf-m-row-gap{-spacer-size}`
- `.pf-m-column-gap{-spacer-size}`
- `.pf-m-gap{-spacer-size}` 

## HTML/CSS demos and react implications
- `--row-gap: 0` set by default and applied to `groups` and `content`
- `--col-gap: 0` set by default and applied to `groups` and `content`
- `.pf-m-row-gap`
  - adds row gap
- `.pf-m-column-gap` - **React:** adds `.pf-m-space-items-none`, unless otherwise specified
  - adds column gap
- `.pf-m-gap` - **React:** adds `.pf-m-space-items-none`, unless otherwise specified
  - adds both row & column gap
    - `--row-gap: var(spacer)`
    - `--col-gap: var(spacer)`
    - row and col gap vars can be updated individually

mcoker:
closing in favor of https://github.com/patternfly/patternfly/pull/5516

PR #5248, opened 2022-11-18
feat(Backdrop): adjusted color of spinner component

### LINKED ISSUE #5174: Color variant of the spinner when on backdrop

thatblindgeye:
Closes #5174 

Included a full screen example to show the update + to match React, but can delete it if it's not necessary. The spinner does slightly change color when toggling the dark theme on and off as well.

### DIFF
--- a/src/patternfly/components/Backdrop/backdrop.scss
+++ b/src/patternfly/components/Backdrop/backdrop.scss
@@ -10,6 +10,10 @@
   width: 100%;
   height: 100%;
   background-color: var(--pf-c-backdrop--BackgroundColor);
+
+  .pf-c-spinner {
+    --pf-c-spinner--Color: var(--pf-global--primary-color--light-100);
+  }
 }
 
 .pf-c-backdrop__open {
--- a/src/patternfly/components/Backdrop/examples/Backdrop.md
+++ b/src/patternfly/components/Backdrop/examples/Backdrop.md
@@ -6,8 +6,19 @@ cssPrefix: pf-c-backdrop
 
 ## Examples
 ### Basic
+
 ```hbs isFullscreen
 {{#> backdrop}}
+{{/backdrop}}
+```
+
+### With spinner
+
+```hbs isFullscreen
+{{#> backdrop}}
+  {{#> bullseye}}
+    {{#> spinner spinner--IsSvg="true"}}Loading...{{/spinner}}
+  {{/bullseye}}
 {{/backdrop}}
 ```
 


src/patternfly/components/Backdrop/backdrop.scss
```diff
@@ -10,6 +10,10 @@
   width: 100%;
   height: 100%;
   background-color: var(--pf-c-backdrop--BackgroundColor);
+
+  .pf-c-spinner {
```
mcoker, line 14: Hmm.... this may be a tricky one. This will style a spinner when it is a child of backdrop, and currently we're placing the modal _inside_ a backdrop (modal and about modal are all I see we're putting in a backdrop). That means that if you have a spinner in a modal, it's going to take on this color.

Screenshot 2022-11-18 at 3 17 10 PM

This may be more long term, but the backdrop `z-index` is 400, modal/about modal are 500, alert group and skip to content are 600. The backdrop and modal combo isn't something I've looked at in a while and don't expect one to be a child of the other. I think it's worth reconsidering that relationship and where backdrop should live. That would allow you to put stuff in backdrop like this if it's designed to be part of the backdrop. Or you could have backdrop on its own if it's just to provide a layer of shade between the page and stuff that shows up on top of it like a modal.

thatblindgeye: An alternative could be instead of placing this change in Backdrop, putting it in Spinner by setting the `--pf-c-spinner--Color` to this lighter color based on a new class, then updating the backdrop examples to use that new class. This would just require adding a new prop in React, though, which might depend if we want to introduce a new prop just for this very specific use case.

I'd agree with what you said re: "I think it's worth reconsidering that relationship and where backdrop should live".

mcoker: @thatblindgeye nice, yeah that's one option. I believe have a dark variant of some other components for uses like this.

Another could be that we do something like create an element of the backdrop (`.pf-c-backdrop__content`) that is designed to put content against the backdrop, and we use the [`pf-t-dark` placeholder](https://github.com/patternfly/patternfly/blob/5eee883c9aadfa4834798de2ee3c00dfa8a37a6a/src/patternfly/sass-utilities/placeholders.scss#L11-L18) to change the vars of anything in that box, which would change the text, border, primary, etc colors to their dark theme equivalents. The only downside to that is that the backdrop isn't a very dark box, so it may not always work that well for anything else. This is what that would look like, using a `.pf-t-dark` class that will make those var changes on an element, though we don't advise anyone use this, it's just for examples AFAIK. 
Nov-21-2022 16-32-11

Another could be to push back on this use case and see if there is a better way to indicate loading status - when @mcarrano brought this issue up, he mentioned that if there isn't a clear solution, we may spend some time reconsidering how to indicate loading status. Looking at the original [issue that added the backdrop w/ spinner example](https://github.com/patternfly/patternfly-react/issues/5008), I'm wondering if it came from a use case, or if we added it on our own.

mcarrano: I don't think we should spend a ton of time on this. My thought was that this would be a simple fix. But if not, yes there are other approaches we can recommend. This was not a pattern that we intended to support, but because an example showed up in the React repo, it was hard to dismiss the request for this.

mcoker: @thatblindgeye great points.

> Could that potentially cause unwanted color changes?

I was thinking when you use this element in the backdrop, that's for stuff you want to load as part of the backdrop. Like the spinner here. I don't think you'd use it with anything else - introducing that element would also go along with us moving everything that is currently in the backdrop (like the modal) out of it. You nailed it with the specificity of a style using a direct descendent combinator, tho to support a specific use case of the spinner in the backdrop, we could create a `.pf-c-backdrop__loading` element and put the spinner in that and style it.

@mcarrano  I guess it's hard to build support for a general use case of putting anything directly on top of the spinner like that. If there's a good use case for a spinner against the backdrop as a pattern we want to support, we could create support for a "loading" backdrop or something like that - that would be pretty simple.

thatblindgeye: >I was thinking when you use this element in the backdrop, that's for stuff you want to load as part of the backdrop. Like the spinner here. I don't think you'd use it with anything else - introducing that element would also go along with us moving everything that is currently in the backdrop (like the modal) out of it.

I am liking this idea a bit more... At least testing in one of the HTML demos from my PR build, moving the `pf-l-bullseye` out of the backdrop, giving it similar properties as backdrop (position fixed, top/left/right of 0, and adding a z-index) does work, and would allow the idea of a backdrop__content element with a recommendation that only loading components should go in there or something? The only thing there is, would we need to adjust the styles for our layouts so that if `pf-l-[layout name]` is a sibling of backdrop, it's given the `z-index: var(--pf-global--ZIndex--xl)` so that it's "higher" than the backdrop (and I assume the position/top/left/etc properties)?

On the one hand, I can see a spinner on a backdrop having some use, if an action was initiated but whatever will appear has to load/be fetched. A spinner on an empty backdrop I think can arguably look "nicer" than an empty modal with a spinner or other loading component. On the other hand, we could also remove the spinner example and advise that content should only be placed in the backdrop if it's ready to be rendered; any sort of loading spinner should appear on the page before the backdrop and other content is rendered which should avoid the color contrast issue. We could still reevaluate the relationship of the backdrop and modal, though, but it at least wouldn't be tied to this issue.

mcarrano:
@thatblindgeye I just had a discussion about this with @mcoker . We decided to put this on hold for now. We want to investigate further why this example was added to React in the first place since it does seem to violate best practices for use of the spinner and the backdrop. We may end up taking a step back to create a better approach that can be implemented as a demo and remove the current React example.

mcoker:
Closing for now.

PR #5255, opened 2022-11-22
chore(button): removed test code from #5135

mcoker:
Removes some hbs test code that was committed in https://github.com/patternfly/patternfly/pull/5135/. Used `git checkout [prev commit hash]` on `[components/demos]/button`


mattnolting (approved):
LPTM 👍

PR #5259, opened 2022-11-28
fix(tabs): remove pagination, expose tags

### LINKED ISSUE #5227: fix(tabs): remove pagination in tables example
The main purpose of this Tabs demo is to showcase the behavior/look of Tabs - there is not enough data to necessitate pagination here. If users would like to see pagination demos, they can look at our `Pagination` component or `Table` examples.

Following a recent discussion regarding updates to demos with tables... since design uses pre-made templates, often the mocks get made using the "kitchen sink" approach to the toolbars. This results in a lot of toolbar features/actions that do not support the main purpose of the demo.

This is specifically an issue on the react side -- implementing every little toolbar feature would add a lot of verbose and unrelated logic to the react demo. We don't want to puff it up with lines and lines of extra code that doesn't support what is actually being demoed.

 This issue is to remove pagination in order to match the updates in https://github.com/patternfly/patternfly-react/pull/8286/files

jenny-s51:
Closes #5227 . Also removed some data to match the current react workspace.


mattnolting (approved):
LGTM 👍

mcoker (approved):
LGTM! I think the tags in the [react demo](https://www.patternfly.org/v4/components/tabs/react-demos/tables-and-tabs/) are better - `tag 1`, `tag 2`, `tag 3`, `2 more` vs `Tag 1`, `Different tag 1`, `Another different tag 1`...

Can we update those in this PR or would you like me to open a separate issue?

jenny-s51 (commented):
Good call @mcoker, agreed ! Updated 👍

mcoker (approved):
🔥

Issue #4926, opened 2022-06-24
Bulk Selection - Enhancement for loading state


katierik:
**Is this a new component or an extension of an existing one?**
This is an extension of the existing Bulk Selection component

**Describe the feature**
We have found that in some cases, the back-end can be slow to respond when the user is selecting a large number of items at the same time. This leaves a delay between the user clicking the "select" action, and seeing results.
This proposes to add a loading state using the existing spinner.

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
![bulk loading](https://user-images.githubusercontent.com/23172038/175625862-c84f09cd-27d4-4612-abf1-ef3b8f78446c.jpg)

This is currently being built as a reusable component by a member of the Insights Frontend team for console.redhat.com

nicolethoen:
[I did create a codesandbox](https://codesandbox.io/s/youthful-volhard-808re3) to demonstrate that it should be feasible - but it would need some style updates to format the spinner correctly.

this should also be something enabled in the MenuToggle component.

nicolethoen:
@mmenestr @mcarrano are there any design guidelines or recommendations related to this?

PR #5269, opened 2022-12-01
feat(dropdown): adds loading state for split button checkbox

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4926

disables the checkbox while in progress to keep the check/label from accepting events while the spinner is shown.

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -62,6 +62,10 @@ $pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "
   // toggle button
   --pf-c-dropdown__toggle-button--Color: var(--pf-global--Color--100);
 
+  // toggle progress
+  --pf-c-dropdown__toggle-progress--Visibility: hidden;
+  --pf-c-dropdown__toggle-progress--c-spinner--diameter: .8em; // one off sizing to roughly match the checkbox size
+
   // split buttons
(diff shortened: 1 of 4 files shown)


mcarrano (approved):
Looks good to me. Thanks @mcoker !

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -74,6 +78,7 @@ $pf-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "
   --pf-c-dropdown__toggle--m-split-button--m-action--child--PaddingRight: var(--pf-global--spacer--sm);
   --pf-c-dropdown__toggle--m-split-button--m-action__toggle-button--MarginRight: calc(-1 * var(--pf-global--BorderWidth--sm));
   --pf-c-dropdown__toggle--m-split-button__toggle-check__input--TranslateY: #{pf-size-prem(-1px)};
+  --pf-c-dropdown__toggle--m-split-button__toggle-check__input--Visibility: auto;
```
thatblindgeye, line 81: Should the value be `visible` instead since `auto` isn't a valid value for visibility? In the end it does get set to visible so may not be a huge deal.

mcoker: @thatblindgeye nice catch, thanks! Updated to `unset` since `visible` will set the checkbox to be visible if some parent element were set to `visiblity: hidden`. `unset` will allow `visibility` to continue to be inherited by default.

thatblindgeye (approved):

jenny-s51 (approved):
Looks good and works as expected @mcoker! 👍

PR #5270, opened 2022-12-01
Feat(flex): Add modifiers for gap to flex component

### LINKED ISSUE #5288: Flex - add gap to flex layout
**Describe the feature**
Update the layout to use gap; maintain the ability to individually space items. Add default gap of -lg for row/column.

srambach:
Fixes #5288

### DIFF
--- a/src/patternfly/layouts/Flex/examples/Flex.md
+++ b/src/patternfly/layouts/Flex/examples/Flex.md
@@ -284,7 +284,7 @@ The CSS approach, by keeping specificity low on base class properties and resett
 {{/l-flex}}
 <br>
 <h3>
-  Column gap using <code>.pf-m-column-gap</code>
+  Column gap using <code>.pf-m-column-gap-sm</code>
 </h3>
 {{#> l-flex l-flex--modifier="pf-m-column-gap-sm"}}
   {{#> l-flex-item}}
@@ -326,7 +326,7 @@ The CSS approach, by keeping specificity low on base class properties and resett
 {{/l-flex}}
 <br>
 <h3>
-  Gap inherits, but can be overridden when nested - <code>.pf-l-gap-xl > .pf-l-gap-xs</code>
+  Gap can be overridden when nested - <code>.pf-l-gap-xl > .pf-l-gap-xs</code>
 </h3>
 {{#> l-flex l-flex--modifier="pf-m-gap-xl"}}
   {{#> l-flex l-flex--modifier="pf-m-gap-xs"}}
@@ -352,7 +352,7 @@ The CSS approach, by keeping specificity low on base class properties and resett
 <br>
 
 <h3>
-  Row gap with individually spaced items - <code>.pf-m-row-gap-xs > pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}</code>. Using <code>.pf-m-spacer</code> is not recommended with column gap.
+  Row gap with individually spaced items - <code>.pf-m-row-gap-xs > .pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}</code>. Using <code>.pf-m-spacer</code> is not recommended on flex layouts that wrap.
 </h3>
 {{#> l-flex l-flex--modifier="pf-m-row-gap-xs"}}
   {{#> l-flex-item}}
@@ -388,9 +388,7 @@ The CSS approach, by keeping specificity low on base class properties and resett
 {{/l-flex}}
 ```
 ### Usage
-Use gap modifiers to space flex items evenly. If individual spacing of items along the main axis is needed, the [flex spacer system](#spacer) can be used instead of row gap.
-
-Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) with gap or column gap is not recommended, as in some cases, it could cause elements to extend past the parent's boundary along the main axis.
+Gap is recommended when there is a consistent gap between items. If individual items need unique gaps, use the [flex spacer system](#spacing) instead. Using the spacer system (<code>.pf-m-spacer</code>) along an axis that uses gap is not recommended, as in some cases, it could cause elements to extend past the boundary of the parent element.
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
@@ -1040,11 +1038,11 @@ Ordering - Ordering can be applied to nested <code>.pf-l-flex</code> and <code>.
 | `.pf-m-align-left{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Resets the flex layout element margin-left property to 0. |
 | `.pf-m-align-right{-on-[breakpoint]}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the flex layout element margin-left property to auto. |
 | `--pf-l-flex--item--Order{-on-[breakpoint]}: {order}` | `.pf-l-flex > .pf-l-flex`, `.pf-l-flex__item` | Modifies the flex layout element order property. |
-| `.pf-m-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-row-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-column-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap, row gap, or column gap. Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) will override gap. |
+| `.pf-m-gap{-none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-row-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-column-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap, row gap, or column gap. Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) will override gap. |
 
 ### Spacer system
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-m-spacer-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex`, `.pf-l-flex > .pf-l-flex__item` |  Modifies a nested flex layout or a flex item spacing. |
-| `.pf-m-item-space-items-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` |  Modifies the flex layout direct descendant spacing. |
+| `.pf-m-spacer{-none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex`, `.pf-l-flex > .pf-l-flex__item` |  Modifies a nested flex layout or a flex item spacing. |
+| `.pf-m-item-space-items{-none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` |  Modifies the flex layout direct descendant spacing. |
--- a/src/patternfly/layouts/Flex/examples/Flex.md
+++ b/src/patternfly/layouts/Flex/examples/Flex.md
@@ -326,7 +326,7 @@ The CSS approach, by keeping specificity low on base class properties and resett
 {{/l-flex}}
 <br>
 <h3>
-  Gap can be overridden when nested - <code>.pf-l-gap-xl > .pf-l-gap-xs</code>
+  Gap can be overridden when nested - <code>.pf-m-gap-xl > .pf-m-gap-xs</code>
 </h3>
 {{#> l-flex l-flex--modifier="pf-m-gap-xl"}}
   {{#> l-flex l-flex--modifier="pf-m-gap-xs"}}
--- a/src/patternfly/layouts/Flex/examples/Flex.md
+++ b/src/patternfly/layouts/Flex/examples/Flex.md
@@ -305,6 +305,24 @@ The CSS approach, by keeping specificity low on base class properties and resett
   {{#> l-flex-item}}
     Flex item
   {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
 {{/l-flex}}
 <br>
 <h3>
@@ -376,7 +394,9 @@ Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) with gap or
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-m-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-row-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-column-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap, row gap, or column gap. |
+| `.pf-m-gap{-none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap. |
+| `.pf-m-row-gap{-none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a row gap. |
+| `.pf-m-column-gap{-none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a column gap. |
 
 ### Layout modifiers
 ```hbs
--- a/src/patternfly/layouts/Flex/examples/Flex.md
+++ b/src/patternfly/layouts/Flex/examples/Flex.md
@@ -6,275 +6,6 @@ cssPrefix: pf-l-flex
 
 import './Flex.css'
 
-### Using gap
-```hbs
-<h3>
-  Gap using <code>.pf-m-gap</code>
-</h3>
-{{#> l-flex l-flex--modifier="pf-m-gap-xs"}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-  {{#> l-flex-item}}
-    Flex item
-  {{/l-flex-item}}
-{{/l-flex}}
-<br/>
-<h3>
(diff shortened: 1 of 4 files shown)


src/patternfly/layouts/Flex/flex.scss
```diff
@@ -10,6 +10,10 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
   --pf-l-flex--m-row--AlignItems: baseline;
   --pf-l-flex--m-row-reverse--AlignItems: baseline;
   --pf-l-flex--item--Order: 0;
+  // stylelint-disable length-zero-no-unit
```
mcoker, line 13: looks like you added the unit - do you still need the stylelint rules?

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -6,6 +6,275 @@ cssPrefix: pf-l-flex
...
+  {{#> l-flex-item l-flex-item--modifier="pf-m-spacer-2xl"}}
+    Item - 2xl
+  {{/l-flex-item}}
+  {{#> l-flex-item l-flex-item--modifier="pf-m-spacer-3xl"}}
+    Item - 3xl
+  {{/l-flex-item}}
+{{/l-flex}}
+<br>
+<h3>
+  Space items will override column gap - <code>.pf-m-gap-xs.pf-m-space-items-xl</code>
```
mcoker, line 209: What's the use case for using them together? On the main axis, it continues to use gap, but offsets the gap with the margin specified in the `pf-m-space-items` class, so you end up with negative margins when the gap is larger than the margin here, too.

Screenshot 2022-12-08 at 6 35 36 PM

mcoker (commented):
Very cool! 😎 Left some comments.

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -6,6 +6,275 @@ cssPrefix: pf-l-flex
...
+  {{#> l-flex l-flex--modifier="pf-m-gap-xs"}}
+    {{#> l-flex-item}}
+      Flex item
+    {{/l-flex-item}}
+  {{/l-flex}}
+{{/l-flex}}
+<br>
+
+<h3>
+  Gap with individually spaced items - <code>.pf-m-gap-xs > pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}</code>.
```
mcoker, line 173: oops! Forgot to submit this comment in my previous review

This works if the spacer is larger than the gap, but if the gap is larger than the spacer, the item will have a negative margin, causing it to go out of bounds.

Screenshot 2022-12-08 at 6 32 48 PM

srambach:
I removed the extra examples and changed the spacer example to show use with row gap rather than gap. I also added a note clarifying that using spacer classes with gap/column gap is not recommended.

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -197,6 +197,185 @@ The CSS approach, by keeping specificity low on base class properties and resett
...
+    {{/l-flex-item}}
+    {{#> l-flex-item}}
+      Flex item
+    {{/l-flex-item}}
+  {{/l-flex}}
+{{/l-flex}}
+<br>
+
+<h3>
+  Row gap with individually spaced items - <code>.pf-m-row-gap-xs > pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}</code>. Using `.pf-m-spacer` is not recommended with column gap.
```
mcoker, line 337: nit. 

```suggestion
  Row gap with individually spaced items - <code>.pf-m-row-gap-xs > pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}</code>. Using <code>.pf-m-spacer</code> is not recommended with column gap.
```

also fwiw the class `.ws-code` formats these code blocks a bit better, though I don't know what the solution is there. I noticed in our a11y docs we're hard coding those classes when we write out the element names, which doesn't seem ideal. Maybe we need to target elements with better selectors to style these elements when we have to write out the tag, or maybe a way to be able to include markdown in these blocks so it gets added automagically? https://github.com/patternfly/patternfly-org/blob/main/packages/v4/patternfly-docs/content/accessibility/card/card.md

Added this org issue to track - https://github.com/patternfly/patternfly-org/issues/3311

mcoker, line 337: nit. 

```suggestion
  Row gap with individually spaced items - <code>.pf-m-row-gap-xs > pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}</code>. Using <code>.pf-m-spacer</code> is not recommended with column gap.
```

also fwiw the class `.ws-code` formats these code blocks a bit better, though I don't know what the solution is there. I noticed in our a11y docs we're hard coding those classes when we write out the element names, which doesn't seem ideal. Maybe we need to target elements with better selectors to style these elements when we have to write out the tag, or maybe a way to be able to include markdown in these blocks so it gets added automagically? https://github.com/patternfly/patternfly-org/blob/main/packages/v4/patternfly-docs/content/accessibility/card/card.md

Added this org issue to track - https://github.com/patternfly/patternfly-org/issues/3311

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -197,6 +197,185 @@ The CSS approach, by keeping specificity low on base class properties and resett
...
+    Item - 3xl
+  {{/l-flex-item}}
+{{/l-flex}}
+```
+### Usage
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-row-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-column-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap, row gap, or column gap. |
+
+Note: Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) with gap or column gap is not recommended, as it could cause elements to extend past the parent's boundary in the inline direction.
```
mcoker, line 377: nit - I think that's probably more clear, since the main axis refers to the flex direction.

```suggestion
Note: Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) with gap or column gap is not recommended, as it could cause elements to extend past the parent's boundary along the main axis.
```

mcoker (commented):
I think this is great and the best approach as far as I'm aware, too. The only thing I'm uncertain about is potential confusion about when to use spacers vs gaps. It may just be clarifying the reasons you'd choose one over another better.

Ideally in a perfect world (`<cough>using :has()</cough>` - _looking at you Firefox_) we could detect the use of an individual spacer on a flex child and update the parent's use of `gap` appropriately for the user. Then we would effectively maintain the current API with a parent spacer and child spacers, and allow us to use gap when appropriate.

If I understand the current state, the main uses cases for gap are:
* There is an equal amount of space between the flex children and you expect those children to wrap.
* The gap is used on the cross axis.

You would use the existing spacing system (which only impacts the main-axis) if:
* You need individual spacers on the children and you don't want to nest extra layouts/groups (unique gaps on nested layouts/groups could provide a unique spacer similar to the spacer system)

Where it gets confusing is using gap with spacers on the main axis. Technically it's fine to use gap on the parent with individual spacers on the children, if:
  * Your layout will not wrap.
  * Your layout will wrap, and the child spacers are larger than the parent gap.

However, in those cases my guess is you'd just use the spacer system because using `gap` with the spacer system doesn't really get you anything over just using the spacer system as far as I'm aware. The only benefit of combining gap with spacers (that I can think of anyways) is that gap would let you could set a common `gap` for both row and column gap with a single value (and if you want to change it, you change a single value), instead of needing to set a spacer on the main axis and gap on the cross axis. 

I'm wondering if we need to support the use case of combining them - and maybe more specifically, if we need to show an example of combining the two, or if it's introducing confusion and would be more clear just to recommend gap for a common spacer (or nested things with gaps for unique spacers), and spacers to support individual child spacing, and note the pros and cons of each approach with recommended use cases. wdyt?

mattnolting:
Love it! I think the potential conflicting gap/spacer values issue is an acceptable risk and that developers should understand how to control the two if they choose to use gap and individual spacers together. 

The most common use cases I can see are:
- Components like chip group/input filters/badges/buttons/etc using a common `.pf-m-gap` where row/column gaps remain consistent as their count changes, and users can expect wrapping. 
- Any instance where `row-gap` is needed (I think this actually applies to all instances), because anything can wrap in the flex layout. 

My only observation is that, I wonder if `row-gap` should be set be default?? I haven't seen an instance where the current implementation looks correct when flex items wrap.

mcoker:
@srambach one update we plan on making is using a common gutter/inset (lg/24px) for the page chrome and components, instead of the dynamic one (md/16px small screens, lg/24px large screens), so we could update to use 24px as the default spacer. @mceledonia can you confirm?

srambach:
> @srambach one update we plan on making is using a common gutter/inset (lg/24px) for the page chrome and components, instead of the dynamic one (md/16px small screens, lg/24px large screens), so we could update to use 24px as the default spacer. @mceledonia can you confirm?

@mcoker you are talking about column gap, correct? Also, this is something we'll want to make as a common gap token rather than having -lg in many places, yes?

srambach:
Added default gaps and changed the default spacer for spacer/space-items to be -lg
It seems a little weird that there's a row gap by default but default spacing on the inline axis is via the margin.
WDYT? @mattnolting @mcoker

mcoker:
@srambach agree, I was wondering if they could both just use gap, too. Whenever there is a consistent spacer for the layout (like the default spacer), I think it should always be a gap.

The only reason I can think of not to do that would be that it is less breaking to leave it as is, since updating to use gap by default will require users review their layouts and update the parent to use the spacer system if any of the children have spacers. It is something code mods may help out with on the react side. Though it is a breaking change release, so... 

There is also an argument to not have a default spacer. None of the other layouts do - they're all enabled with a `gutter` variation.

srambach:
I propose we stick with this for now, and add a follow on issue to rearrange the flex layout so that gap is the main way that the gutter is created.

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -197,6 +197,187 @@ The CSS approach, by keeping specificity low on base class properties and resett
...
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
```
mcoker, line 307: Would you mind adding a few more so the row wraps by default and the column gap is visible? The first two examples use 12 items, that would probably make sense to keep them consistent.

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -197,6 +197,187 @@ The CSS approach, by keeping specificity low on base class properties and resett
...
+{{/l-flex}}
+```
+### Usage
+Use gap modifiers to space flex items evenly. If individual spacing of items along the main axis is needed, the [flex spacer system](#spacer) can be used instead of row gap.
+
+Using the spacer system (`.pf-m-spacer` or `.pf-m-item-space-items`) with gap or column gap is not recommended, as in some cases, it could cause elements to extend past the parent's boundary along the main axis.
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-row-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-column-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap, row gap, or column gap. |
```
mcoker, line 379: I think these should be unique rows in the table for `gap`, `row-gap`, and `column-gap`.

Also since you don't have to specify a size, I think the class suffix should be `.pf-m-gap{-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}}`, or something like that, just denoting that `.pf-m-[row/column-]gap` can be used on its own. @mattnolting wdyt, I feel like you know the curly vs regular braces syntax better than I do

mcoker, line 379: Needs the breakpoint link.

```suggestion
| `.pf-m-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-row-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}`, `.pf-m-column-gap-{none, xs, sm, md, lg, xl, 2xl}{-on-[breakpoint]}` | `.pf-l-flex` | Creates a gap, row gap, or column gap at optional [breakpoint](https://www.patternfly.org/v4/developer-resources/global-css-variables#breakpoint-variables-and-class-suffixes). |
```

src/patternfly/layouts/Flex/flex.scss
mcoker: Not sure if they're needed or not, but wanted to mention again. Looks like they should be able to be removed?

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -160,6 +167,24 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
...
+        }
+      }
+
+      &.pf-m-column-gap#{$breakpoint-name} {
+        --pf-l-flex--ColumnGap: var(--pf-l-flex--RowGap-base);
+
+        > * {
+          --pf-l-flex--spacer: 0;
+        }
+      }
```
mcoker, line 186: There is no `.pf-m-row-gap` - should there be? I think that might be a good way to get the default spacer when used with a breakpoint suffix, eg `.pf-m-row-gap-xs.pf-m-row-gap-on-md` would use an xs gap by default, and the default on `md` screens and larger. If the default row gap is set by default in the layout, I don't know that we need `.pf-m-row-gap` on its own.

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -312,4 +337,53 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
     }
   }
 }
+
+// .pf-m-gap-{size}{-on-breakpoint}
+@each $breakpoint, $breakpoint-value in $pf-l-flex--breakpoint-map {
+  $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
```
mcoker, line 343: similar to `$breakpoint-name` here, if you have a conditional for `$spacer` that only includes the `-#{$spacer}` if the key isn't "base", you can remove the [classes that set `.pf-m-gap/row-gap/column-gap`](https://github.com/patternfly/patternfly/blob/a9848b35565359c9bf7f1dddd73078b3252c7ec9/src/patternfly/layouts/Flex/flex.scss#L170-L186) and move those into this block so it will generate `.pf-m-gap[-[sizes][-on-[breakpoints]]`

mcoker (changes requested):
There is a conflict with the column layout and row-gap

Screenshot 2022-12-14 at 3 37 54 PM

--

Feel free to add these to another issue or however you'd like to handle it, happy to approve whenever so we can iterate on it if needed.

src/patternfly/layouts/Flex/examples/Flex.md
srambach: This isn't on any of the individual modifiers throughout this page because there are so many, but that link is mentioned at the top of the page in a breakpoint section.

src/patternfly/layouts/Flex/flex.scss
srambach: done

src/patternfly/layouts/Flex/examples/Flex.md
srambach: Done. I think the column modifier works correctly now with gap. I think long term there are some improvements we could do to use gap by default for both directions.

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -311,5 +332,75 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
...
+        }
+        &.pf-m-gap-#{$spacer}#{$breakpoint-name} {
+          --pf-l-flex--spacer-base: 0;
+          --pf-l-flex--RowGap: var(#{map-get($pf-l-flex--variable-map, $spacer-value)});
+          --pf-l-flex--ColumnGap: var(#{map-get($pf-l-flex--variable-map, $spacer-value)});
+
+          > * {
+            --pf-l-flex--spacer: 0;
+          }
+        }
```
mcoker, line 359: With all of these loops, having both `.pf-m-[row/column-]gap#{$breakpiont-name}` and `.pf-m-[row/column-]gap-#{$spacer}#{$breakpoint-name}` is causing the `.pf-m-[row/column-]gap#{$breakpiont-name}` block to be duplicated in each iteration of the spacer map loop. I _think_ you could just move that block outside/before the spacer loop.

Here are a couple of screenshots.

Screenshot 2023-01-04 at 9 15 37 PM
Screenshot 2023-01-04 at 9 15 16 PM

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -197,6 +197,207 @@ The CSS approach, by keeping specificity low on base class properties and resett
...
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+  {{#> l-flex-item}}
+    Flex item
+  {{/l-flex-item}}
+{{/l-flex}}
+<br>
+<h3>
+  Gap inherits, but can be overridden when nested - <code>.pf-l-gap-xl > .pf-l-gap-xs</code>
```
mcoker, line 329: I'm probably not reading it right, but not sure what "gap inherits" is referring to. The `xl` gap only applies to the parent layout. If you remove the nested gap (no space modifer, uses the default spacer system), or just use `.pf-m-gap`, the default spacers are used.

mcoker, line 329:
```suggestion
  Gap inherits, but can be overridden when nested - <code>.pf-m-gap-xl > .pf-m-gap-xs</code>
```

mcoker (changes requested):

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -311,5 +332,77 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
       }
     }
   }
+
+
+  // .pf-m-gap-{size}{-on-breakpoint}
+  @each $breakpoint, $breakpoint-value in $pf-l-flex--breakpoint-map {
```
mattnolting, line 338: If you were to move line 338 - 406 above line 320, the spacer could override gap settings.

mattnolting (changes requested):

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -10,9 +10,15 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
   --pf-l-flex--m-row--AlignItems: baseline;
   --pf-l-flex--m-row-reverse--AlignItems: baseline;
   --pf-l-flex--item--Order: 0;
+  --pf-l-flex--RowGap-base: var(--pf-global--spacer--lg);
```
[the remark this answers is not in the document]
mcoker: Confirmed with @mceledonia we're going to drop the 16px -> 24px gutter change between small -> large screens, so a 24px gap seems suitable

mattnolting (changes requested):
Just a few nits, otherwise this looks great!🥇

src/patternfly/layouts/Flex/flex.scss
```diff
@@ -295,6 +316,77 @@ $pf-l-flex--variable-map: build-variable-map("pf-l-flex--spacer", $pf-l-flex--sp
...
+    }
+  }
+
+  // .pf-m-column-gap-{size}{-on-breakpoint}
+  @each $breakpoint, $breakpoint-value in $pf-l-flex--breakpoint-map {
+    $breakpoint-name: if($breakpoint != "base", -on-#{$breakpoint}, "");
+
+    @include pf-apply-breakpoint($breakpoint) {
+      &.pf-m-column-gap#{$breakpoint-name} {
+        --pf-l-flex--ColumnGap: var(--pf-l-flex--RowGap-base);
```
mcoker, line 371: typo?

```suggestion
        --pf-l-flex--ColumnGap: var(--pf-l-flex--ColumnGap-base);
```

mcoker (changes requested):
One typo, I think anyways :) Once that's fixed, I'm inclined to file another issue for the stuff below as a follow up to move this feature along so other components and the react work can begin. I think the API is fine, the CSS just needs some adjustin'. wdyt?

--

The columns are still off in some cases. I found a couple of places, but would like to test more.
* column gaps set the spacer var to "0", but spacer is used between rows for column flex layouts
  Screenshot 2023-01-17 at 5 05 08 PM
* There is a difference between columns if the children are regular items versus flex layouts because of the way the space is calculated, and the value of `--pf-l-flex--RowGap` - it's set to "0" for `.pf-m-column`, but is reset to `--base` on a nested `.pf-l-flex`.
  Screenshot 2023-01-17 at 5 11 37 PM

thatblindgeye (approved):

mcoker:
closing in favor of https://github.com/patternfly/patternfly/pull/5516

PR #5280, opened 2022-12-08
feat(card): improve card header

### LINKED ISSUE #3924: Card actions element margin

### LINKED ISSUE #4958: Card - rework header to be more flexible
The card header is a bit complicated to use currently.

Specifically, it's a flex layout, so its children will currently be a part of that layout. This works fine with something simple like a title + actions, but can be a problem with other types of elements in the header (images for example, or just a few spans). To resolve that we added `__header-main`, but it's currently a bit unintuitive as to when to use `__header-main`, including on the react side since the card is basically entirely composable.

It would probably be good to have `__header-main` as a required element of the card header as a place the consumer can pass the title, images, etc as regular elements, and lay them out as they please. That would also allow react to pass children of `<CardHeader>` to `__header-main` behind the scenes, making it more intuitive for users.

Can the main element be a structurally required element of the header?
Maybe the flex layout flexes the main, the actions, and the icon.

Related:
* https://github.com/patternfly/patternfly/issues/3713
* https://github.com/patternfly/patternfly/issues/3924

srambach:
Fixes #4958 by making the `.pf-c-card__header-main` required if there is a `.pf-c-card__header`.

This creates a container so that other items and layouts can be within the header without disturbing the expand icon and actions. Some styling that was specific to title in the header was moved to header-main.

Fixes #3924 by using the correct margin variables.

Pending a design change, it does not directly address having actions wrap below the header main content as requested in #3714, but that would be possible by placing the actions into a layout as desired in the header main.

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -106,7 +106,6 @@
   // Toggle right
   --pf-c-card__header--m-toggle-right--toggle--MarginRight: calc(var(--pf-global--spacer--form-element) * -1);
   --pf-c-card__header--m-toggle-right--toggle--MarginLeft: var(--pf-global--spacer--xs);
-  --pf-c-card__header--m-toggle-right--actions--MarginRight: 0;
 
   // Card input
   --pf-c-card__input--focus--BorderWidth: var(--pf-global--BorderWidth--md);
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -329,7 +329,7 @@
   align-self: flex-start;
   order: 1;
   padding-left: var(--pf-c-card__actions--PaddingLeft);
-  margin: var(--pf-c-card__actions--MarginTop) auto var(--pf-c-card__actions--MarginBottom) auto;
+  margin: var(--pf-c-card__actions--MarginTop) 0 var(--pf-c-card__actions--MarginBottom) auto;
 
   > * + * {
     margin-left: var(--pf-c-card__actions--child--MarginLeft);
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -14,6 +14,9 @@
   --pf-c-card__footer--FontSize: var(--pf-global--FontSize--md);
   --pf-c-card__actions--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-card__actions--child--MarginLeft: var(--pf-global--spacer--sm);
+  --pf-c-card__actions--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-card__actions--MarginRight: var(--pf-global--spacer--xs);
+  --pf-c-card__actions--MarginBottom: calc(var(--pf-global--spacer--form-element) * -1);
 
   // Expandable
   --pf-c-card__header-toggle--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
@@ -296,17 +299,17 @@
     --pf-c-card__header-toggle--MarginLeft: var(--pf-c-card__header--m-toggle-right--toggle--MarginLeft);
 
     .pf-c-card__actions {
-      --pf-c-card__header-toggle--MarginRight: var(--pf-c-card__header--m-toggle-right--actions--MarginRight);
+      --pf-c-card__actions--MarginRight: var(--pf-c-card__header--m-toggle-right--actions--MarginRight);
     }
 
     .pf-c-card__header-toggle {
       order: 2;
     }
-
-    .pf-c-card__title {
-      flex: 1;
-    }
   }
+}
+
+.pf-c-card__header-main {
+  flex: 1;
 }
 
 .pf-c-card__header-toggle {
@@ -331,7 +334,7 @@
   align-self: flex-start;
   order: 1;
   padding-left: var(--pf-c-card__actions--PaddingLeft);
-  margin: var(--pf-c-card__header-toggle--MarginTop) var(--pf-c-card__header-toggle--MarginRight) var(--pf-c-card__header-toggle--MarginBottom) auto;
+  margin: var(--pf-c-card__actions--MarginTop) var(--pf-c-card__actions--MarginRight) var(--pf-c-card__actions--MarginBottom) auto;
 
   > * + * {
     margin-left: var(--pf-c-card__actions--child--MarginLeft);
@@ -344,8 +347,8 @@
   }
 
   &.pf-m-no-offset {
-    --pf-c-card__header-toggle--MarginTop: 0;
-    --pf-c-card__header-toggle--MarginBottom: 0;
+    --pf-c-card__actions--MarginTop: 0;
+    --pf-c-card__actions--MarginBottom: 0;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Card/card.scss
```diff
@@ -14,6 +14,9 @@
   --pf-c-card__footer--FontSize: var(--pf-global--FontSize--md);
   --pf-c-card__actions--PaddingLeft: var(--pf-global--spacer--md);
   --pf-c-card__actions--child--MarginLeft: var(--pf-global--spacer--sm);
+  --pf-c-card__actions--MarginTop: calc(var(--pf-global--spacer--form-element) * -1);
+  --pf-c-card__actions--MarginRight: var(--pf-global--spacer--xs);
```
mattnolting, line 18: I think this should be set to `0` by default and updated to `--xs` when `.pf-m-right`

src/patternfly/components/Card/card.scss
```diff
@@ -296,19 +299,19 @@
     --pf-c-card__header-toggle--MarginLeft: var(--pf-c-card__header--m-toggle-right--toggle--MarginLeft);
 
     .pf-c-card__actions {
-      --pf-c-card__header-toggle--MarginRight: var(--pf-c-card__header--m-toggle-right--actions--MarginRight);
+      --pf-c-card__actions--MarginRight: var(--pf-c-card__header--m-toggle-right--actions--MarginRight);
```
mattnolting, line 302: Looks like this variable doesn't exist, but the result is expected as `.pf-c-card__header-toggle` has a left margin update for `.pf-c-card__header.pf-m-toggle-right`, therefore this looks like it's unnecessary.

src/patternfly/demos/Card/examples/Card.md
```diff
@@ -981,23 +1007,25 @@ import './Card.css'
...
-            <span class="pf-u-color-200">
-              Storage capacity
-            </span>
-          {{/l-flex}}
-        {{/l-flex-item}}
-        {{#> l-flex-item l-flex-item--modifier="pf-m-flex-1"}}
-          {{> card-demo--chart}}
-        {{/l-flex-item}}
-      {{/l-flex}}
+      {{#> card-header-main card-header-main}}
```
mattnolting, line 1010: Dupe `.hbs` filename.

mattnolting (commented):
Looks good 👍, noted a few things.

src/patternfly/components/Card/card.scss
srambach: Are you sure? And do you mean .pf-m-toggle-right? This looks the same as in production to me.

src/patternfly/components/Card/card.scss
srambach: Which variable are you saying doesn't exist?

src/patternfly/components/Card/card.scss
mattnolting: It looks errant to me. The right margin causes actions to be misaligned. I assume this is a remnant of an older layout?

Screenshot 2022-12-16 at 9 50 23 AM

Screenshot 2022-12-16 at 9 52 55 AM

@mceledonia wdyt?

src/patternfly/components/Card/examples/Card.md
```diff
@@ -430,9 +438,11 @@ import './Card.css'
       {{> card--dropdown}}
       {{> card--check}}
     {{/card-actions}}
+    {{#> card-header-main}}
```
mattnolting, line 441: Nit: next code block could be indented

src/patternfly/components/Card/examples/Card.md
```diff
@@ -518,7 +532,7 @@ A card is a generic rectangular container that can be used to build other compon
 | `.pf-c-card__header-toggle` | `<div>` | Creates the expandable card toggle. |
 | `.pf-c-card__header-toggle-icon` | `<span>` | Creates the expandable card toggle icon. |
 | `.pf-c-card__actions` | `<div>` | Creates an actions element to be used in the card header. |
-| `.pf-c-card__header-main` | `<div>` | Creates a wrapper element to be used in the card header when using an image, logo, or text. |
+| `.pf-c-card__header-main` | `<div>` | Creates a wrapper element to be used in the card header when using an image, logo, or text. **Required if `.pf-c-card__header` has content outside a card header toggle or card header actions** |
```
mattnolting, line 535:
```suggestion
| `.pf-c-card__header-main` | `<div>` | Creates a wrapper element to be used in the card header when using an image, logo, or text. **Required if `.pf-c-card__header` has content outside of a card header toggle or card header actions** |
```

src/patternfly/components/Card/card.scss
```diff
@@ -331,7 +329,7 @@
   align-self: flex-start;
   order: 1;
   padding-left: var(--pf-c-card__actions--PaddingLeft);
-  margin: var(--pf-c-card__header-toggle--MarginTop) var(--pf-c-card__header-toggle--MarginRight) var(--pf-c-card__header-toggle--MarginBottom) auto;
+  margin: var(--pf-c-card__actions--MarginTop) auto var(--pf-c-card__actions--MarginBottom) auto;
```
mattnolting, line 332: Should this be `0` instead of auto?

```suggestion
  margin: var(--pf-c-card__actions--MarginTop) 0 var(--pf-c-card__actions--MarginBottom) auto;
```

src/patternfly/components/Card/examples/Card.md
srambach: Is that North Carolinese? ;-P

mattnolting: No telling, could be any number of dialects

src/patternfly/components/Card/card.scss
srambach: @mattnolting is this correct? 
image

srambach:
> wdyt about using gap and a `.pf-m-wrap` modifier on `.pf-c-card__header`? One concern with that might be with the expandable variation since you don't want things to wrap under the toggle.

I'm not sure if you are asking for a code change here or suggesting another way to do it?

mcoker:
@srambach Just to support the basic use case from @garrett's issue in https://github.com/patternfly/patternfly/issues/3713 (PR https://github.com/patternfly/patternfly/pull/3714). We can follow up on that in another issue though, since we likely need design feedback like you mentioned in this PR's description.

mcoker (approved):
LGTM. A couple of follow up considerations, but we can file another issue for them, however you'd like to go forward with it.

The level layout in this card demo looks different now that the `__main` element is set to flex grow. Does it need to be set to grow? Or should we update the demo? https://patternfly-pr-5280.surge.sh/components/card/html-demos#horizontal-grid-collapsed

We can use `gap` for the card actions element to space its children.

mattnolting (approved):
Great work! LPTM 🚢

mcoker (approved):

Issue #5263, opened 2022-11-29
Icons - Add "processor" icon


mcarrano:
Adds a new icon for "Processor". Pending completion of https://github.com/patternfly/patternfly-design/issues/1210

yichen1yu:
![data-processor](https://user-images.githubusercontent.com/108347329/208143913-465fd14a-31c2-47a5-a39e-10265e25e418.svg)

yichen1yu:
Svg filed attached. Feel free to let me know if there's any question. Thanks!

PR #5284, opened 2022-12-16
feat(icon): added data-processor icon

mcoker:
**preview link: https://patternfly-pr-5284.surge.sh/icons**

fixes https://github.com/patternfly/patternfly/issues/5263


mcarrano (commented):
Looks good to me. @yichen1yu can you also verify? 

@mcoker My only question is about naming. I was calling this `data-processor` and you have it here as just `processor` See Org issue: https://github.com/patternfly/patternfly-org/issues/3326. Which should it be? @yichen1yu do you have a preference? My only caution with using the name "processor" is that people could confuse this with the hardware processor as in a CPU. Thoughts?

mcoker:
@mcarrano thanks for the org issue link! Re-read the design issue and looks like this is complimentary to data-sink and data-source, so I went ahead and updated the icon name to data-processor.

yichen1yu:
@mcarrano @mcoker Yes it should be data-processor as Michael pointed out we already have data-sink and data-source. I guess we just follow that pattern

yichen1yu:
@mcoker I tried this link https://patternfly-pr-5284.surge.sh/icons and found there shows no icon on the "processor" and didn't find a "data-processor" in the column. Would mind pointing me where to see the preview? Thanks!!

yichen1yu:
@mcoker Thank you. Looks good!

mcarrano (approved):
LGTM. Thanks @mcoker !

Issue #5247, opened 2022-11-18
Chip group - update main element


mcoker:
This can likely be the default, and we can remove this example - https://www.patternfly.org/v4/components/chip-group/html#legacy-chip-group-examples-without-main-element

nicolethoen:
Already completed

PR #5293, opened 2022-12-20
chore(ChipGroup): use gap instead of item margins

### LINKED ISSUE #5289: Chip group - add gap for spacing
**Describe the feature**
Use gap for spacing chips. Maintain existing spacing.

srambach:
Fixes #5289 

Removes margins within chip group in favor of using gap for spacing.
Some comments included to mark items that can be removed if legacy chip group (without the main container) is removed.
I left a slight difference in spacing between category label and the chips when they wrap because I suspect it was due to lack of row spacing or just oversight? @mceledonia can you let me know which is preferred?
image

### DIFF
--- a/src/patternfly/components/ChipGroup/chip-group.scss
+++ b/src/patternfly/components/ChipGroup/chip-group.scss
@@ -1,11 +1,24 @@
 .pf-c-chip-group {
   @include pf-t-light; // force the container follow the light theme
 
-  // Chip group list
-  --pf-c-chip-group__list--MarginBottom: calc(var(--pf-global--spacer--xs) * -1);
-  --pf-c-chip-group__list--MarginRight: calc(var(--pf-global--spacer--xs) * -1);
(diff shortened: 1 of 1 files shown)


mattnolting:
I'm in favor of removing the legacy chip group ✋

srambach:
> I'm in favor of removing the legacy chip group ✋

@mcoker WDTY?

mcoker (approved):
LPTM! 🎯

mcoker:
@mattnolting @srambach sorry just saw the comment about the legacy structure - yep, we have an issue to remove it, we should be using the `__main` element all the time https://github.com/patternfly/patternfly/issues/5247

mattnolting (approved):
LPTM! 🚀

srambach:
@mcoker ok, then is this good to go and we'll remove the legacy group separately? You can assign me that if you want.

PR #5296, opened 2022-12-20
fix(label-group): use gap instead of item margins

### LINKED ISSUE #5294: LabelGroup - change from margins to using gap
**Describe the feature**
Use gap for spacing labels within the label group. Maintain existing spacing.

srambach:
Fixes #5294 

Removes margins within label group in favor of using gap for spacing.
Like in the [chip group PR](https://github.com/patternfly/patternfly/pull/5293), there is a slight change to the space between category and label when it wraps - @mceledonia please let me know what spacing is to spec.

### DIFF
--- a/src/patternfly/components/LabelGroup/label-group.scss
+++ b/src/patternfly/components/LabelGroup/label-group.scss
@@ -1,7 +1,21 @@
 .pf-c-label-group {
-  // List
-  --pf-c-label-group__list--MarginBottom: calc(var(--pf-global--spacer--xs) * -1);
-  --pf-c-label-group__list--MarginRight: calc(var(--pf-global--spacer--xs) * -1);
+  // Label group - spaces main with the close button
+  --pf-c-label-group--RowGap: var(--pf-global--spacer--sm);
+  --pf-c-label-group--ColumnGap: 0;
+  --pf-c-label-group-m-vertical--RowGap: var(--pf-global--spacer--sm);
+  --pf-c-label-group-m-vertical--ColumnGap: var(--pf-global--spacer--sm);
+
+  // Main - spaces the category label with the list
+  --pf-c-label-group__main--RowGap: var(--pf-global--spacer--xs);
+  --pf-c-label-group__main--ColumnGap: var(--pf-global--spacer--sm);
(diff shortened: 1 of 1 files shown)


mattnolting (approved):
Yay `gap`!! LPTM 👍

src/patternfly/components/LabelGroup/label-group.scss
```diff
@@ -110,31 +115,31 @@
...
 
   max-width: var(--pf-c-label-group__label--MaxWidth);
-  margin-right: var(--pf-c-label-group__label--MarginRight);
-  margin-bottom: var(--pf-c-label-group__label--MarginBottom);
   font-size: var(--pf-c-label-group__label--FontSize);
 }
 
 .pf-c-label-group__close {
+  display: flex;
+  align-self: start;
```
mcoker, line 142: I think it's fine, but it's worth noting that this shifts the close button up by a pixel or two?

This PR:
patternfly-pr-5296 surge sh_components_label-group_html_category-removable_

Currently on v4:
pf4 patternfly org_components_label-group_html_category-removable_

And that's from this change to move the close button to the top instead of being vertically centered when labels wrap. I didn't see that called out in the issue, so just wanted to make sure that change is good, too:

This PR:
pf4 patternfly org_components_label-group_html_category-removable_ (1)

Currently on v4:
patternfly-pr-5296 surge sh_components_label-group_html_category-removable_ (1)

cc @mceledonia @mmenestr

mcoker (commented):
Looks awesome! One small nit - just want to verify it's good from design before merging.

src/patternfly/components/LabelGroup/label-group.scss
mmenestr: I personally think moving it to the top right is the right choice!
I'm confused though because you say you moved the close button to the top, but image is showing that that is the current version?

srambach: Just to clarify - this PR maintains the close button at the top right that was changed in a previous PR.
image

mcoker (approved):
L🚀TM!

PR #5307, opened 2023-01-06
fix(empty-state): cleaned up CSS

### LINKED ISSUE #5138: Empty state - clean up CSS
* Looks like we can comb through the CSS and make general updates around variable usage and general maintenance of specificity - like creating a var for the body font-size that can be manipulated at the size variation selector specificity https://github.com/patternfly/patternfly/blob/02bda75f8f22a9e230be216ed2e6e8ffd5b48cc3/src/patternfly/components/EmptyState/empty-state.scss#L79-L81
* We're styling the title font-size - we should remove that and see if there is still a need to manage the font-size there - https://github.com/patternfly/patternfly/blob/02bda75f8f22a9e230be216ed2e6e8ffd5b48cc3/src/patternfly/components/EmptyState/empty-state.scss#L92-L94
  * We may consider adding a title element to the empty state for the title, and/or for a place to use the title component, instead of styling the title component when it's a direct descendent.
* Similar to ^, we're styling `.pf-c-button.pf-m-primary` when it's a direct descendent and that can be problematic. We should probably remove that, and just use the `__primary` empty state element for the primary action (which may not always be a primary button)
* Consider using the small button modifier in the xs empty state if we want to have small actions, instead of styling all buttons, which could be problematic and potentially break if we add button variations that set font-size with a higher specificity than this rule - https://github.com/patternfly/patternfly/blob/02bda75f8f22a9e230be216ed2e6e8ffd5b48cc3/src/patternfly/components/EmptyState/empty-state.scss#L60-L62
* Use gap here - https://github.com/patternfly/patternfly/blob/02bda75f8f22a9e230be216ed2e6e8ffd5b48cc3/src/patternfly/components/EmptyState/empty-state.scss#L117-L129
* Update the icon so that we're always using a container for the icon, and the icon is placed as a child of the icon container. This will keep the empty state class on the container instead of needing to manipulate a class on the icon itself. Currently the react empty statue component needs an icon passed as a component by default, and the empty state class is applied to the icon directly which we would like to avoid.
  * related - https://github.com/patternfly/patternfly-design/issues/1198

srambach:
Fixes #5138 
[Empty state page](https://patternfly-pr-5307.surge.sh/components/empty-state)

- Body text font size is properly controlled with variables
- The icon must be contained in the icon container
- Title size is driven by the modifier on the title rather than empty state overrides
- Button size is determined by the button modifier rather than empty state overrides
- __primary/__secondary containers are now required if there are primary/secondary actions
- __primary and __secondary containers use flex/gap to space contents
- Rearranged variables for clarity
- Removed Patternfly "4" from example text

### DIFF
--- a/src/patternfly/components/EmptyState/empty-state-icon.hbs
+++ b/src/patternfly/components/EmptyState/empty-state-icon.hbs
@@ -1,14 +1,13 @@
-{{#if @partial-block}}
   <div class="pf-c-empty-state__icon{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
     {{#if empty-state-icon--attribute}}
       {{{empty-state-icon--attribute}}}
     {{/if}}>
-    {{> @partial-block}}
-  </div>
-{{else}}
-<i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}} pf-c-empty-state__icon{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
-  aria-hidden="true"
-  {{#if empty-state-icon--attribute}}
-    {{{empty-state-icon--attribute}}}
-  {{/if}}></i>
-{{/if}}
+    {{#if @partial-block}}
+      {{> @partial-block}}
+    {{else}}
+      <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
+      aria-hidden="true"
+      {{#if empty-state-icon--attribute}}
+        {{{empty-state-icon--attribute}}}
+      {{/if}}></i>
+  </div>{{/if}}
--- a/src/patternfly/components/EmptyState/empty-state-icon.hbs
+++ b/src/patternfly/components/EmptyState/empty-state-icon.hbs
@@ -1,13 +1,14 @@
-  <div class="pf-c-empty-state__icon{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
+<div class="pf-c-empty-state__icon{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
+  {{#if empty-state-icon--attribute}}
+    {{{empty-state-icon--attribute}}}
+  {{/if}}>
+  {{#if @partial-block}}
+    {{> @partial-block}}
+  {{else}}
+    <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
+    aria-hidden="true"
     {{#if empty-state-icon--attribute}}
       {{{empty-state-icon--attribute}}}
-    {{/if}}>
-    {{#if @partial-block}}
-      {{> @partial-block}}
-    {{else}}
-      <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
-      aria-hidden="true"
-      {{#if empty-state-icon--attribute}}
-        {{{empty-state-icon--attribute}}}
-      {{/if}}></i>
-  </div>{{/if}}
+    {{/if}}></i>
+  {{/if}}
+</div>
--- a/src/patternfly/components/EmptyState/empty-state-icon.hbs
+++ b/src/patternfly/components/EmptyState/empty-state-icon.hbs
@@ -5,10 +5,7 @@
   {{#if @partial-block}}
     {{> @partial-block}}
   {{else}}
-    <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
-    aria-hidden="true"
-    {{#if empty-state-icon--attribute}}
-      {{{empty-state-icon--attribute}}}
-    {{/if}}></i>
+    <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}"
+    aria-hidden="true"></i>
   {{/if}}
 </div>
--- a/src/patternfly/components/EmptyState/empty-state.scss
+++ b/src/patternfly/components/EmptyState/empty-state.scss
@@ -22,6 +22,15 @@
   --pf-c-empty-state--m-xl__icon--MarginBottom: var(--pf-global--spacer--xl);
   --pf-c-empty-state--m-xl__icon--FontSize: #{pf-size-prem(100px)};
 
+  // Title text
+  --pf-c-empty-state__title-text--FontFamily: var(--pf-global--FontFamily--heading--sans-serif);
+  --pf-c-empty-state__title-text--FontSize: var(--pf-global--FontSize--xl);
+  --pf-c-empty-state__title-text--FontWeight: var(--pf-global--FontWeight--normal);
+  --pf-c-empty-state__title-text--LineHeight: var(--pf-global--LineHeight--md);
+  --pf-c-empty-state--m-xs__title-text--FontSize: var(--pf-global--FontSize--md);
+  --pf-c-empty-state--m-xl__title-text--FontSize: var(--pf-global--FontSize--4xl);
+  --pf-c-empty-state--m-xl__title-text--LineHeight: var(--pf-global--LineHeight--sm);
+
   // Body
   --pf-c-empty-state__body--MarginTop: var(--pf-global--spacer--md);
   --pf-c-empty-state__body--Color: var(--pf-global--Color--200);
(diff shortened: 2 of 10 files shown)


src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -88,25 +87,26 @@
 
 .pf-c-empty-state__content {
   max-width: var(--pf-c-empty-state__content--MaxWidth);
-
-  > .pf-c-title.pf-m-lg {
-    font-size: var(--pf-c-empty-state__content--c-title--m-lg--FontSize);
```
mcoker, line 93: This is good, but I'm wondering if we should use the title component here, or have an empty state title element of the empty state component instead (`.pf-c-empty-state__title`).

FWIW, this override came from a design request to change the empty state's title font-size from lg to xl along with another change (I think with an opt-in to a different font). Since we're using the title component here, and the size is tied to a modifier class, an HTML change has to happen to get that change, and we try to avoid that for core users since they'll have the wrong font size until they change their HTML. HTML changes needed to fix things from a release are considered breaking - though I wouldn't necessarily say it's "breaking" to use a lg font here instead of the intended xl. But since we wanted everyone to get this font size change from lg to xl , we styled the title's large modifier to use the xl font - which is just a messy compromise. In a number of other components, we've removed use of the title component and replaced it with an element of the component, when the title's styling is tied more to the component and its variations than just to the user's preference of any sort of title. A user would still have the option of passing a `<Title>` in if they wanted to.

What I was thinking is a block for the header, body, and footer/actions, and using a consistent container class for blocks of actions.

```
empty-state
  empty-state__header (new)
    empty-state__icon
    empty-state__title (new)
      this can be any node - text, a layout, a title component, whatever
  empty-state__body
  empty-state__footer (new)
    empty-state__actions (new - at least one required)
    empty-state__actions (new - for any secondary/add'l actions)
```

wdyt?

src/patternfly/components/EmptyState/examples/EmptyState.md
```diff
@@ -191,12 +241,12 @@ cssPrefix: pf-c-empty-state
...
-| `.pf-c-empty-state__body` | `<div>` |  Creates the empty state body content. You can have more than one `.pf-c-empty-state__body` elements. |
-| `.pf-c-button.pf-m-primary` | `<button>` |  Creates the primary action button. |
-| `.pf-c-empty-state__primary` | `<div>` |  Container for primary actions. Can be used in lieu of using `.pf-c-button.pf-m-primary`. |
-| `.pf-c-empty-state__secondary` | `<div>` |  Container secondary actions. |
+| `.pf-c-empty-state__header` | `<div>` |  Creates the header container. **Required** |
+| `.pf-c-empty-state__title` | `<div>` |  Creates the empty state title container. |
+| `.pf-c-empty-state__icon` | `<div>` |  Creates the empty state icon container. |
+| `.pf-c-empty-state__body` | `<div>` |  Creates the empty state body content. There can be more than one `.pf-c-empty-state__body` elements. |
+| `.pf-c-empty-state__footer` | `<div>` |  Container for actions. **Required** |
+| `.pf-c-empty-state__actions` | `<div>` |  Container for actions. **Required** |
```
mcoker, line 249: Just a nit - these have the same description, which is accurate. I wonder if the footer should be more inline with the header and body desc? Like "Creates the footer container" - wdyt?

mcoker (commented):
This is awesome!! One nit about the docs.

I think my only other question is - should the empty state have its own title element with a size/weight/etc that is managed by the component - assuming that title can be overridden by the consumer? Or will we continue to show the title in our examples using the title component, but that's more of a demo since all of the title styles come from the title component. One consideration around that - if the empty state title is part of the empty state component, we can adjust the font properties of the title in the CSS and everyone gets the change (like adjusting the font-size size for a particular empty state variation). If the title styling comes from the title component, any/most change(s) we'd like to see to the font properties of an empty state title (for a particular variation or whatever) will need to be handled by the user, since the title component styles will be tied to the markup (eg, we get a design request to "use a "large" title instead of "extra large" title for [whatever] empty state variation"). I suppose this is also a question for design - @mcarrano @mmenestr @mceledonia

src/patternfly/components/EmptyState/empty-state-icon.hbs
```diff
@@ -1,14 +1,13 @@
...
-{{/if}}
+    {{#if @partial-block}}
+      {{> @partial-block}}
+    {{else}}
+      <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
+      aria-hidden="true"
+      {{#if empty-state-icon--attribute}}
+        {{{empty-state-icon--attribute}}}
+      {{/if}}></i>
+  </div>{{/if}}
```
mattnolting, line 13:
```suggestion
  </div>{{/if}}
  
```

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -3,38 +3,43 @@
...
+
+  // Footer
+  --pf-c-empty-state__footer--RowGap: var(--pf-global--spacer--sm);
+  --pf-c-empty-state__footer--MarginTop: var(--pf-global--spacer--xl);
+  --pf-c-empty-state--m-xs__footer--MarginTop: var(--pf-global--spacer--md);
+
+  // Actions
+  --pf-c-empty-state__actions--RowGap: var(--pf-global--spacer--xs);
+  --pf-c-empty-state__actions--ColumnGap: var(--pf-global--spacer--xs);
+  --pf-c-empty-state__actions--MarginTop: 0;
```
mattnolting, line 42: Is this being used?

mattnolting (commented):
This is great! 

I agree w/@mcoker here
> I think my only other question is - should the empty state have its own title element with a size/weight/etc that is managed by the component - assuming that title can be overridden by the consumer? Or will we continue to show the title in our examples using the title component, but that's more of a demo since all of the title styles come from the title component.

Otherwise LGTM!

src/patternfly/components/EmptyState/examples/EmptyState.md
```diff
@@ -9,85 +9,113 @@ cssPrefix: pf-c-empty-state
...
-  {{#> empty-state-secondary}}
+
+  {{#> empty-state-footer}}
+    {{#> empty-state-actions}}
+    {{#> button button--modifier="pf-m-primary"}}
+      Primary action
+    {{/button}}
+  {{/empty-state-actions}}
+
+  {{#> empty-state-actions}}
```
mattnolting, line 118:
```suggestion
    {{#> empty-state-actions}}
      {{#> button button--modifier="pf-m-primary"}}
        Primary action
      {{/button}}
    {{/empty-state-actions}}

    {{#> empty-state-actions}}
```

src/patternfly/components/EmptyState/examples/EmptyState.md
```diff
@@ -101,82 +129,104 @@ cssPrefix: pf-c-empty-state
...
     {{/button}}
     {{#> button button--modifier="pf-m-link"}}
-      In the secondary
+      In the second
     {{/button}}
     {{#> button button--modifier="pf-m-link"}}
       Action area
     {{/button}}
-  {{/empty-state-secondary}}
+  {{/empty-state-actions}}
```
mattnolting, line 137:
```suggestion
    {{/empty-state-actions}}
```

src/patternfly/components/EmptyState/examples/EmptyState.md
```diff
@@ -101,82 +129,104 @@ cssPrefix: pf-c-empty-state
...
 
 ### Large
 ```hbs
 {{#> empty-state empty-state--modifier="pf-m-lg"}}
-  {{#> empty-state-icon}}{{/empty-state-icon}}
-  {{#> title titleType="h1" title--modifier="pf-m-lg"}}
-    Empty state
-  {{/title}}
+  {{#> empty-state-header}}
+    {{#> empty-state-icon}}{{/empty-state-icon}}
```
mattnolting, line 146:
```suggestion
    {{> empty-state-icon}}
```

src/patternfly/components/EmptyState/examples/EmptyState.md
```diff
@@ -101,82 +129,104 @@ cssPrefix: pf-c-empty-state
...
-  {{#> empty-state-body}}
-    This represents an the empty state pattern in PatternFly 4. Hopefully it's simple enough to use but flexible enough to meet a variety of needs.
-  {{/empty-state-body}}
-  {{#> button button--modifier="pf-m-primary"}}
-    Primary action
-  {{/button}}
-{{/empty-state}}
-```
+  {{#> empty-state-header}}
+    {{#> empty-state-icon}}{{/empty-state-icon}}
```
mattnolting, line 192:
```suggestion
    {{> empty-state-icon}}
```

mattnolting (commented):
Just a few nits

mattnolting (approved):
LPTM 😄

src/patternfly/components/EmptyState/empty-state-icon.hbs
```diff
@@ -1,14 +1,14 @@
...
-{{/if}}
+  {{/if}}>
+  {{#if @partial-block}}
+    {{> @partial-block}}
+  {{else}}
+    <i class="fas fa-{{#if empty-state-icon--type}} fa-{{empty-state-icon--type}}{{else}}cubes{{/if}}{{#if empty-state-icon--modifier}} {{empty-state-icon--modifier}}{{/if}}"
+    aria-hidden="true"
+    {{#if empty-state-icon--attribute}}
+      {{{empty-state-icon--attribute}}}
+    {{/if}}></i>
```
mcoker, line 12: I don't think we need this, we don't typically pass attrs to FA icons. Tho if you want to keep it, the param name will need to be unique - currently it's the same as the param used for the `.pf-c-empty-state__icon` element.

mcoker (changes requested):
Just one small docs nit, otherwise lgtm!

mcoker (approved):
🚀

PR #5314, opened 2023-01-10
feat(v5): merge main into v5

mcoker:


nicolethoen (approved):
This LGTM
I will note that the commit history listed in this PR is much longer than it needs to be - most of those commits have already been merged into v5, and that's why the number of files changed is only 5. I merged a PR right before shut down with updated v5 with most of these commits.

mcoker:
@nicolethoen yep, looks like v5 is currently 52 commits behind main - https://github.com/patternfly/patternfly/compare/v5...main

good to merge this one?

mcoker:
@nicolethoen ✅

Issue #5222, opened 2022-11-08
Bug - CalendarMonth - Fix spacing of year to not truncate


phill-85:
**Describe the problem**
The year input field for the CalendarMonth component appears to use a standard HTML input element with the type number. Using Brave on Ubuntu 22.10 & Windows 10 this gives the little up and down arrows to the right of the input. Using Brave on Android 13 this behaves like a line edit with the numeric keyboard.

**How do you reproduce the problem?**
View demo at https://www.patternfly.org/v4/components/calendar-month

**Expected behavior**
I think this would provide better design consistency, and usability, if the NumberInput component was used instead. The + & - controls will provide better usability, especially on mobile where to change to the next or previous year will require only 1 touch instead of 3.

phill-85:
Alternatively / Additionally, |< & >| buttons could be added to the CalendarMonth component to change the year.

mcarrano:
We looked at this. While we don't expect to use a Number input here, we do see the issue and will investigate other solutions.

nicolethoen:
Our list of supported browsers at this time can be found here: https://www.patternfly.org/v4/get-started/about#supported-browsers

phill-85:
Also the same in Chrome Version 107.0.5304.110 (Official Build) (64-bit), on mobile and desktop.

Screenshot from Chrome:
![image](https://user-images.githubusercontent.com/4619659/202048677-3f662c5a-5a9a-44d4-b8f7-1e2166c6208d.png)

nicolethoen:
The arrows appearing in the input was something we left intentionally. But in some browsers it's causing the year to truncate, so that's not ideal.

PR #5316, opened 2023-01-10
fix(calendar-month): updated year width to not truncate

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5222

You can see the year truncate if you set the year to "0000" in chrome. This just bumps the size a bit for cases where that might trigger.


thatblindgeye (approved):

mattnolting (approved):
LPTM 🌮

srambach (approved):
👍

PR #5317, opened 2023-01-11
fix(wizard): move wizard footer below steps/nav

### LINKED ISSUE #5183: Wizard demo - footer extends below steps/nav
https://www.patternfly.org/v4/components/wizard/html-demos/with-drawer-in-page-expanded/

Screen Shot 2022-10-18 at 3 39 36 PM

The footer should only be under the wizard content. Also a side note, there is an extra scrollbar in this example.

The react example looks correct - https://www.patternfly.org/v4/components/wizard/react-demos/in-page-with-drawer-and-informational-step/

mattnolting:
closes #5183


mcoker (approved):
lgtm!

PR #4840, opened 2022-05-09
fix(tooltip): added dark theme border

mcoker:
fixes https://github.com/patternfly/patternfly/issues/4771

PR #5318, opened 2023-01-11
fix(tooltip-popover): updated bg color, border, box shadow

### LINKED ISSUE #5158: Tooltip/popover - dark theme background/border/shadow
After reviewing with @mceledonia we decided that the popover and tooltips should present like menus in dark theme.

* background-color-300
* no border
* bring back the box shadow

Chart tooltips have borders but that's OK, the tooltip is an SVG path and adding a box shadow seems like it isn't very straight forward.

mattnolting:
closes #5158

### DIFF
--- a/src/patternfly/components/Popover/themes/dark/popover.scss
+++ b/src/patternfly/components/Popover/themes/dark/popover.scss
@@ -2,16 +2,10 @@
 
 @mixin pf-theme-dark-popover() {
   .pf-c-popover {
-    --pf-c-popover--BoxShadow: none;
-    --pf-c-popover__arrow--BoxShadow: none;
+    --pf-c-popover__content--BackgroundColor: var(--pf-global--BackgroundColor--300);
     --pf-c-popover__arrow--BackgroundColor: var(--pf-c-popover__content--BackgroundColor);
     --pf-c-popover--m-default__title-text--Color: var(--pf-global--default-color--200);
     --pf-c-popover--m-info__title-text--Color: var(--pf-global--info-color--100);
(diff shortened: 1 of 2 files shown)


srambach (approved):
👍

mcoker:
@mceledonia @lboehling when we added the border around the tooltip in dark theme (https://github.com/patternfly/patternfly/pull/4840/files) we had to increase the size of the little arrow so it would still appear as an arrow and have a visible border. This is what it looks like now:

Light theme:
Screenshot 2023-01-20 at 3 50 47 PM


Dark theme:
Screenshot 2023-01-20 at 3 50 54 PM

----

Without the border and the arrow size in light/dark theme the same size, this is what dark theme looks like

Screenshot 2023-01-20 at 3 48 42 PM

----

FWIW, this is what light and dark theme look like if they both use the larger arrow size from dark theme when it had a border

Light theme:
Screenshot 2023-01-20 at 3 46 17 PM

Dark theme:
Screenshot 2023-01-20 at 3 44 39 PM

wdyt? The arrow feels a bit small to me, but that may just be me. It may also be because the popover's arrow is the same size between dark and light theme. I'm wondering if we can keep the changes between themes to just the color, which would happen at the global var level with the change in background color.

Current popover: https://pf4.patternfly.org/components/popover/html/top/
Popover in @mattnolting's PR that removes the border: https://patternfly-pr-5318.surge.sh/components/popover/html/top/

mcoker (approved):
LGTM! Chatted with @mceledonia about the arrow and opened a follow up issue - https://github.com/patternfly/patternfly/issues/5358

PR #5328, opened 2023-01-12
refactor(background-image): updated bg image

### LINKED ISSUE #5282: Login and About modal - Update background image to use lightweight SVG
Want to update background images used in Login and About Modal components per the designs here: https://github.com/patternfly/patternfly-design/issues/1202#issuecomment-1349585686

The goal is to replace the bitmaps with lighter weight SVGs that can be easily replaced by projects. The default SVG images will be shipped as placeholders.

@mceledonia can you add the SVG file here to be used for this implementation?

mattnolting:
closes #5282 

Previews:
https://patternfly-pr-5328.surge.sh/components/about-modal/
https://patternfly-pr-5328.surge.sh/components/background-image/
https://patternfly-pr-5328.surge.sh/components/login-page/


mcoker (commented):
A few comments. I also wonder if we should build in some image replacement at different viewport sizes? Kinda like the brand component supports.

mcarrano (commented):
This looks great. My only critique is that the login page looks a bit odd in dark mode since there is not much is any differentiation between the login form and the background.

Screenshot 2023-02-02 at 1 50 29 PM

Not sure what to do about that, but feels like the space occupied by the login form shoukd be better defined. Thoughts @mceledonia ?

mcarrano (approved):
This looks great @mattnolting . Thanks for making those changes.

src/patternfly/components/BackgroundImage/background-image.scss
```diff
@@ -1,47 +1,22 @@
...
-
-.pf-c-background-image__filter {
-  display: block;
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: -1;
+  width: 100%;
+  height: 100%;
+  content: "";
```
thatblindgeye, line 16: This could be removed.

thatblindgeye (changes requested):

srambach (commented):
AFAICT, there is no documentation about how a product would replace the PatternFly logo with their own since it's no longer a separate element and is a background image instead. I assume the React implementation would provide a prop for it, but it really should also be documented in core since this is a common customization that will happen.

thatblindgeye (approved):
Looks good other than the above comment from @srambach.

Wondering if for the HTML implementation whether it'd make more sense to just add a level 3 heading to the "Documentation" section, more explicitly stating "In order to use a custom image do XYZ" (I imagine just giving the CSS var a new `url('...')` value).

Could also show an example, but that might require an example for the other components that use BackgroundImage internally.

Issue #5320, opened 2023-01-12
promote beta components for v4


nicolethoen:
Banner (html)
Description list (html)
Helper text demo (html)
Page demos (html)
Password strength demo (html)

PR #5329, opened 2023-01-12
fix(beta): cleaned up beta tags for v4

nicolethoen:
Closes: https://github.com/patternfly/patternfly/issues/5320

### DIFF
--- a/src/patternfly/demos/Page/examples/Page.md
+++ b/src/patternfly/demos/Page/examples/Page.md
@@ -13,7 +13,7 @@ wrapperTag: div
 
 ### Full height page
 Using the `.pf-m-full-height` modifier class on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other ancestors of `.pf-c-page`, have height set to 100%.
-```hbs isFullscreen isBeta
+```hbs isFullscreen
 {{> page-template page-template--id="page-demo-full-height" page-template--modifier="pf-m-full-height"}}
 ```
 
@@ -37,7 +37,7 @@ Using the `.pf-m-full-height` modifier class on the page component eliminates th
 ```
 
 ### Sticky breadcrumb on medium
-```hbs isFullscreen isBeta
+```hbs isFullscreen
 {{> page-template
       page-template--id="page-demo-sticky-top-breadcrumb"
       page-template-gallery--IsLongGallery="true"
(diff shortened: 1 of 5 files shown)


src/patternfly/demos/Page/examples/Page.md
```diff
@@ -13,7 +13,7 @@ wrapperTag: div
 
 ### Full height page
 Using the `.pf-m-full-height` modifier class on the page component eliminates the need to ensure that the `<html>` and `<body>` tags, and any other ancestors of `.pf-c-page`, have height set to 100%.
-```hbs isFullscreen isBeta
+```hbs isFullscreen
```
mcoker, line 16: @srambach can you think of any reason this should still be beta? Assuming it's promoted, do you think it would be OK to make this the default page component height and remove this opt-in for v5?

mcoker (commented):
LGTM, one question for @srambach

src/patternfly/demos/Page/examples/Page.md
srambach: I think it could definitely leave beta. 
If it were the default, would that break the federated apps? I don't think they are placing a whole page into the content area but I'm not positive. What would you call the modifier to remove it if necessary?

mcoker: Agree, it could definitely stay beta. I'm not sure who all is using it and how well it works. Re: the federated apps, I don't know either - I imagine that is something we could test/verify? @nicolethoen was there another reason we want to promote this feature for v4 before releasing v5?

FWIW, since the class applies a [max height of 100%][(https://github.com/patternfly/patternfly/blob/3e41bd9ad94783c02f7c10c77330a9bf8ff94f31/src/patternfly/components/Page/page.scss#L215), I would expect it to be OK nested within other sections. 

Here is a page component nested in the content of another with both using `.pf-m-full-height`. I added `height: 100%` to the page section that holds the nested page component: https://codepen.io/mcoker/pen/VwBrejE

And here is the same without the `.pf-m-full-height` modifiers, but leaving `height: 100%` on the parent page section for the nested page so scrolling works properly: https://codepen.io/mcoker/pen/WNKXrjr

I'm not sure if that helps or is what you were getting at though?

nicolethoen: Im happy to leave the beta tags here

mcoker (approved):

Issue #5240, opened 2022-11-16
Tab - help action should have same spacing as other actions


kmcfaul:
Spacing for help and other tab actions should be the same.

In react, it's worth noting that a user could potentially opt in to uniform spacing by omitting the css modifier altogether. I'm unsure if that would change anything else aside from spacing though.

PR #5330, opened 2023-01-13
fix(tabs): removed help action specific styling

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5240

The help action feature is marked as beta in core and react, safe to remove. This will need a react follow up to remove the custom class.


mmenestr (approved):

mattnolting (approved):
LPTM 👍

PR #5339, opened 2023-01-18
chore(base): remove breaking styles from _common.scss

### LINKED ISSUE #5261: Bug - Remove outdated styles in _common.scss
**Describe the problem**

Follow up to #5218 
Towards https://github.com/patternfly/patternfly-react/pull/5275

Tested on a local branch with the changes in https://github.com/patternfly/patternfly-react/pull/5275. The icon sizes are correct now when we import the svg classes following the changes in #5218 . However some outdated styles in `_common.scss` are now causing the PF site to look like this when built:

Screen Shot 2022-11-16 at 3 47 32 PM

After investigating, removing these outdated styles seems to fix it. https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/_common.scss#L2-L11

These are already expected to be removed in the breaking-change release based on the comments.

jenny-s51:
Closes #5261

### DIFF
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -1,26 +1,4 @@
 // Set common reset styles for patternfly components
-// remove in breaking change. this should come from globals, and can be opted out
-[class*="pf-c-"] {
-  &,
-  &::before,
-  &::after {
-    padding: 0;
-    margin: 0;
-    background-color: transparent;
-  }
-}
-
-// remove in breaking change
-@if $pf-global--unset-root-font-size {
-  // Since PF3 sets root font size to 10px, we need to unset it.
-  // This doesn't affect PF3.
-  // https://github.com/twbs/bootstrap/blob/v3.4.0/less/scaffolding.less#L23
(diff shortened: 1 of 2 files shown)


src/patternfly/base/_common.scss
```diff
@@ -1,26 +1,4 @@
...
-  &::before,
-  &::after {
-    padding: 0;
-    margin: 0;
-    background-color: transparent;
-  }
-}
-
-// remove in breaking change
-@if $pf-global--unset-root-font-size {
```
mcoker, line 14: we can also remove this https://github.com/patternfly/patternfly/blob/14e45911682ee81e822cecc02c22cb380a98c05b/src/patternfly/sass-utilities/scss-variables.scss#L15

mcoker (changes requested):
so nice!!! one small change to remove a sass var we no longer need.

jenny-s51 (commented):
Thank you @mcoker ! Updated 🙂

mcoker (approved):
nice 😎

Issue #5348, opened 2023-01-24
Code block - classes in docs are incorrect


mcoker:
the docs are referencing chip-group instead of code-block - https://github.com/patternfly/patternfly/blob/4979b534998855b4036fd33371d19eba7fb21a6a/src/patternfly/components/CodeBlock/examples/CodeBlock.md?plain=1#L127-L137

mcoker:
closed by https://github.com/patternfly/patternfly/pull/5349

PR #5349, opened 2023-01-24
chore(code-block): updated incorrect class names in docs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5348


srambach (approved):
Good catch.

Issue #5352, opened 2023-01-25
Label - editable label shows close button


mcoker:
I think this is an unexpected change from https://github.com/patternfly/patternfly/pull/5209. 

Specifically this block was removed:

https://github.com/patternfly/patternfly/pull/5209/files#diff-2c450c181adeb511f6c0e5b9bdaab0cc05d7ec94487293b9cac0329ff2e38a7eL375-L378

```diff
-
-    .pf-c-button {
-      visibility: hidden; // use visibility hidden to retain positioning
-    }
```

`visibility` is set there on purpose (not along with `display: none`) to hide the button but keep its layout on the page.

@thatblindgeye can you confirm?

thatblindgeye:
Is there a specific example showing the bug? I'm not sure if I'm noticing it.

Looking at the latest releases on Org, Core and React example for an editable Label look similar to a few releases back. For both it looks like the close button just doesn't get rendered when `pf-m-editable-active` is applied.

mcoker:
@thatblindgeye ah sorry, it took me a minute to find the example, too! You can see it here - https://pf5.patternfly.org/components/label-group/#editable-labels-label-active-dynamic-label-group

> For both it looks like the close button just doesn't get rendered when pf-m-editable-active is applied.

Ah hah, yep you're right. I think we just need to remove the close button from that editable label example linked above. @mattnolting can you confirm? Or is there a reason we would want to leave the close button there, but hide it with `visibility: hidden`? I put a PR up with that change - https://github.com/patternfly/patternfly/pull/5354

thatblindgeye:
Closed via #5354

PR #5354, opened 2023-01-26
chore(label-group): removed close button from editable active example

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5352

### DIFF
--- a/src/patternfly/components/LabelGroup/examples/LabelGroup.md
+++ b/src/patternfly/components/LabelGroup/examples/LabelGroup.md
@@ -514,7 +514,7 @@ In addition to the JavaScript management of [editable labels](/components/label#
         {{/label}}
       {{/label-group-list-item}}
       {{#> label-group-list-item}}
-        {{#> label label--modifier="pf-m-blue pf-m-active" label--id=(concat label-group--id '-editable-label-active') label--isRemovable="true" label--IsEditable="true" label--IsEditableActive="true"}}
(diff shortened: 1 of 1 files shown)


src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -514,7 +514,7 @@ In addition to the JavaScript management of [editable labels](/components/label#
         {{/label}}
       {{/label-group-list-item}}
       {{#> label-group-list-item}}
-        {{#> label label--modifier="pf-m-blue pf-m-active" label--id=(concat label-group--id '-editable-label-active') label--isRemovable="true" label--IsEditable="true" label--IsEditableActive="true"}}
+        {{#> label label--modifier="pf-m-blue pf-m-active" label--id=(concat label-group--id '-editable-label-active') label--IsEditable="true" label--IsEditableActive="true"}}
```
thatblindgeye, line 517: It looks like the `pf-m-active` class isn't doing anything for the label component here. In React, that class doesn't get applied anywhere for Label or LabelGroup. Same would apply to line 583. Should this class be something to add to the React implementation, or removed from Core?

mcoker: Nice find! Looks like it was added here by accident, it doesn't seem to do anything. https://github.com/patternfly/patternfly/pull/4097/files

thatblindgeye (approved):

PR #5355, opened 2023-01-26
chore: updated components to use standalone checkbox/radio

### LINKED ISSUE #3830: Update components with single check/radio input to use check/radio component
From https://github.com/patternfly/patternfly/pull/3821, we have existing components that include a regular `<input type="checkbox/radio">` that should use the standalone variation of the check/radio component in our next breaking change release.

thatblindgeye:
Closes #3830

### DIFF
--- a/src/patternfly/components/Check/check-input.hbs
+++ b/src/patternfly/components/Check/check-input.hbs
@@ -3,9 +3,9 @@
   {{#if check-input--IsChecked}}
     checked
   {{/if}}
+  {{#if check-input--IsDisabled}}
+    disabled
+  {{/if}}
   {{#if check-input--attribute}}
     {{{check-input--attribute}}}
-  {{/if}}
-  {{#if check--IsDisabled}}
-    disabled
   {{/if}}>
--- a/src/patternfly/components/Check/check.hbs
+++ b/src/patternfly/components/Check/check.hbs
@@ -2,5 +2,8 @@
   {{#if check--attribute}}
     {{{check--attribute}}}
   {{/if}}>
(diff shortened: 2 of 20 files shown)


mcoker (commented):
Nice! Just a couple of nits/questions - curious your feedback. The PR is fine as-is though, we can follow up on those in another issue (or not at all).

Also wdyt about these two? From the react component perspective, would it be beneficial to use our check component for uses like this - ease of creating, attaching events, etc? It's probably easier to leave them as they are from core's perspective. If we use the check component, we will probably need to bump the specificity of the existing CSS that references the elements.

https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Switch/switch-input.hbs

https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Card/card-sr-input.hbs

thatblindgeye:
>Also wdyt about these two? From the react component perspective, would it be beneficial to use our check component for uses like this - ease of creating, attaching events, etc? It's probably easier to leave them as they are from core's perspective. If we use the check component, we will probably need to bump the specificity of the existing CSS that references the elements.

I was considering both of these when going through things that had a sort of standalone checkbox/radio. They both felt independent enough from our checkbox component to not really *need* to be updated. It would probably depend if we'd need/want any of the styling coming from any of the `.pf-c-check...` classes applying to the Switch or hidden Card input for any reason. In React there are some props already on Checkbox that are being applied to the `input` elements, but the same thing might apply re: the `pf-c-check` classes.

mcoker (approved):
Nice work!!

PR #5356, opened 2023-01-26
chore(chip-group): require main and remove legacy example

### LINKED ISSUE #5335: Chip group - require use of main element
re: https://github.com/patternfly/patternfly/pull/5293

remove [legacy examples](https://pf4.patternfly.org/components/chip-group) and any other examples and CSS associated with supporting chip group without the `__main` element.

srambach:
Fixes #5335 

- `.pf-c-chip-group__main` container is now required
- Removes the `--pf-c-chip-group__close--MarginLeft` variable
- Removes the legacy chip group example


src/patternfly/components/ChipGroup/chip-group.scss
```diff
@@ -8,7 +8,7 @@
   --pf-c-chip-group--PaddingBottom: 0;
   --pf-c-chip-group--PaddingLeft: 0;
   --pf-c-chip-group--RowGap: var(--pf-global--spacer--sm);
-  --pf-c-chip-group--ColumnGap: var(--pf-global--spacer--sm);
+  --pf-c-chip-group--ColumnGap: 0;
```
mcoker, line 11: it's fine to leave, but this isn't used, so we could remove it, too.

mcoker (approved):
LGTM!

mcoker (approved):
LGTM!

src/patternfly/components/ChipGroup/chip-group.scss
srambach: I think for flexibility it's good to have (it is used on line 43 if that's what you meant)

PR #5364, opened 2023-02-02
fix(label): allow text to truncate

### LINKED ISSUE #4795: Bug - Label - Labels break out of container with long strings of text

mattnolting:
This PR: Closes #4795 

### Variables

Adds:
* `--pf-c-label--MaxWidth`
* `--pf-c-label__text--m-truncate--MaxWidth`


src/patternfly/components/Label/label.scss
```diff
@@ -410,12 +415,15 @@
   @include pf-text-overflow;
 
   max-width: var(--pf-c-label__text--MaxWidth);
+
+  &.pf-m-truncate {
```
mcoker, line 419: Is the issue no longer to truncate labels by default?

src/patternfly/components/Label/label.hbs
```diff
@@ -1,10 +1,46 @@
...
+  {{/ifEquals}}
+  {{#ifEquals label--color "gold"}}
+    pf-m-gold
+  {{/ifEquals}}
+  {{#if label--modifier}} {{label--modifier}}{{/if}}"
+  {{#if label--IsOverflow}}
+    type="button"
+  {{else if label--IsAdd}}
+    type="button"
+  {{/if}}
```
mcoker, line 34: could also be written as

```suggestion
  {{#if (concat label--IsOverflow label--IsAdd)}}
    type="button"
  {{/if}}
```

mcoker (commented):
Seems like there was a modifier added to truncate, but the issue was to truncate labels by default. Did that change? The labels that say the text should truncate do not truncate
Screenshot 2023-02-16 at 5 53 03 PM

src/patternfly/components/Label/label.scss
mattnolting: @mcoker This issue is still to truncate labels by default, but do so when the label runs out of room rather than at `{x}ch` width. 

Screenshot 2023-02-22 at 3 38 23 PM

Perhaps the predefined width truncation `pf-m-truncate` modifier would be implemented by simply updating `--pf-c-label__text--MaxWidth`.

src/patternfly/components/Label/label.hbs
mattnolting: Woot!

src/patternfly/components/Label/label-editable-content.hbs
```diff
@@ -1,9 +1,10 @@
...
   id="{{label--id}}-editable-content"
-  {{#if label--IsEditableActive}} 
-    type="text" 
-    value="{{~> @partial-block}}"{{else}} 
-    currvalue="{{~> @partial-block}}"
+  {{#if label--IsEditableActive}}
+    type="text"
+    value="{{label-editable-content--value}}"
+  {{else}}
+    currvalue="{{label-editable-content--value}}"
```
mcoker, line 7: out of scope but I don't understand what this does? It's in the docs, but it isn't a valid html attr and isn't used in react. I don't think you need it - you'd just update the button text with the input value.

https://github.com/patternfly/patternfly/blob/bef73445bf939b7a022ba297a7697a3c555821ef/src/patternfly/components/Label/examples/Label.md?plain=1#L89

src/patternfly/components/LabelGroup/examples/LabelGroup.md
```diff
@@ -14,25 +14,25 @@ cssPrefix: pf-c-label-group
       {{#> label-group-list-item}}
         {{#> label}}
           {{#> label-icon}}
-            <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i>
+            <i class="fas fa-fw fa-info-circle" aria-hiddent></i>
```
mcoker, line 17: typo?

src/patternfly/components/Label/examples/Label.md
```diff
@@ -699,7 +123,7 @@ This style of label is used to add new labels to a label group.
 | `.pf-c-label` | `<span>`, `<button>` | Initiates a label. Without a color modifier, the label's default style is grey. Use a color modifier to change the label color. Use a `<button>` if the label is an overflow label used in the label group. **Required** |
 | `.pf-c-label__content` | `<span>`, `<a>`, `<button>` | Iniates a label content. Use as an `<a>` element if the label serves as a link. Use a `<button>` if the label serves as an action. **Required** |
 | `.pf-c-label__icon` | `<span>` | Initiates a label icon. |
-| `.pf-c-label__text` | `<span>` | Initiates label text. |
+| `.pf-c-label__text` | `<span>` | Initiates label text. **Required** |
 | `.pf-c-label__editable-text` | `<button>`, `<input>` | Initiates editable label text. See the [editable](#editable) example for details about handling behavior in Javascript.|
 | `.pf-m-outline` | `.pf-c-label` | Modifies label for outline styles. |
 | `.pf-m-compact` | `.pf-c-label` | Modifies label for compact styles. |
```
mcoker, line 129: need to add `.pf-m-truncate`

src/patternfly/components/Label/label.scss
mcoker: got it. Should this class be on `.pf-c-label` tho?

mcoker (changes requested):
A couple of comments, but otherwise LGTM. We do need to update the examples in some way since there is a label in each example block that says it will truncate. I don't know if we add `.pf-m-truncate` to those examples, or have a new example block that shows truncation? Seems like the former is the easiest and matches what we have now. 

Also added @mceledonia @mcarrano @mmenestr to double check the proposed change is good. This will allow labels to be as big as they need to be based on their content, and will only truncate if the label runs out of space in the layout.

Screenshot 2023-02-27 at 7 43 35 PM

mcarrano (commented):
@mattnolting @mcoker Just a couple of questions to clarify what is happening here. I'm reading this as follows. Please confirm that this is correct.

By default, labels will grow to the width of their content unless or until they are constrained by the layout or container that they live in. If there is not enough room to show all the content, that content will truncate. We're also adding a modifier to allow setting a max width of the label, in which case the content will truncate if it exceeds that length.

Do I understand this correctly? If so, I am good with this but agree that we should show this somehow in our examples.

mmenestr:
I agree with @mcarrano ^ LGTM

mcoker (approved):

PR #5370, opened 2023-02-03
feat(popover): add header, title, close elements

### LINKED ISSUE #4123: Introduce header and title elements for popover
Similar to the modal, the popover should have a header and title element instead of just using the title component.

srambach:
Fixes #4123 

Breaking changes:
- Adds `pf-c-popover__close` around the close button (required if there is a close button)
- Adds `pf-c-popover__header` and `pf-c-popover__title` to all variations (was previously only on icon/alert variations)
- Moves h1 from `pf-c-popover__title` to the `pf-c-popover__title-text`
- Moves the screen reader text in the alert variant to within `pf-c-popover__title-text`

Also removes and renames corresponding css variables.

### DIFF
--- a/src/patternfly/components/Popover/examples/Popover.md
+++ b/src/patternfly/components/Popover/examples/Popover.md
@@ -437,6 +437,7 @@ A popover is used to provide contextual information for another component on cli
 | `aria-modal="true"` | `.pf-c-popover` | Tells assistive technologies that the windows underneath the current popover are not available for interaction. **Required**|
 | `aria-label="Close"` | `.pf-c-button` | Provides an accessible name for the close button as it uses an icon instead of text. **Required**|
 | `aria-hidden="true"` | Parent element containing the page contents when the popover is open. | Hides main contents of the page from screen readers. The element with `.pf-c-popover` must not be a descendent of the element with `aria-hidden="true"`. For more info, see [trapping focus](/accessibility/product-development-guide#trapping-focus). **Required** |
+| `.pf-screen-reader` | `.pf-c-popover__title-text` | Adds text to be read saying the alert status before the title. If `.pf-c-popover__title-text` is not used to create a title, this must be added within the title. **Required for popovers that are alerts** |
 
 ### Usage
 | Class | Applied to | Outcome |
--- a/src/patternfly/components/Popover/examples/Popover.md
+++ b/src/patternfly/components/Popover/examples/Popover.md
@@ -11,12 +11,18 @@ import './Popover.css'
 ```hbs isFullscreen
 {{#> popover popover--modifier="pf-m-top" popover--attribute='aria-labelledby="popover-top-header" aria-describedby="popover-top-body"'}}
   {{#> popover-content}}
-    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
-      <i class="fas fa-times" aria-hidden="true"></i>
-    {{/button}}
-    {{#> title titleType="h1" title--modifier="pf-m-md" title--attribute='id="popover-top-header"'}}
-      Popover header
-    {{/title}}
+    {{#> popover-close}}
+      {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
+        <i class="fas fa-times" aria-hidden="true"></i>
+      {{/button}}
+    {{/popover-close}}
+    {{#> popover-header}}
+      {{#> popover-title popover-title--attribute='id="popover-top-header"'}}
+        {{#> popover-title-text}}
+          Popover header
+        {{/popover-title-text}}
+      {{/popover-title}}
+    {{/popover-header}}
     {{#> popover-body popover-body--attribute='id="popover-top-body"'}}
       Popovers are triggered by click rather than hover. Click again to close.
     {{/popover-body}}
@@ -31,12 +37,18 @@ import './Popover.css'
 ```hbs isFullscreen
 {{#> popover popover--modifier="pf-m-right" popover--attribute='aria-labelledby="popover-right-header" aria-describedby="popover-right-body"'}}
   {{#> popover-content}}
-    {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
-      <i class="fas fa-times" aria-hidden="true"></i>
-    {{/button}}
-    {{#> title titleType="h1" title--modifier="pf-m-md" title--attribute='id="popover-right-header"'}}
-      Popover header
-    {{/title}}
+    {{#> popover-close}}
+      {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
+        <i class="fas fa-times" aria-hidden="true"></i>
+      {{/button}}
+    {{/popover-close}}
+    {{#> popover-header}}
+      {{#> popover-title popover-title--attribute='id="popover-right-header"'}}
+        {{#> popover-title-text}}
+          Popover header
+        {{/popover-title-text}}
+      {{/popover-title}}
+    {{/popover-header}}
     {{#> popover-body popover-body--attribute='id="popover-right-body"'}}
       Popovers are triggered by click rather than hover. Click again to close.
     {{/popover-body}}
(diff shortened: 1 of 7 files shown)


mcoker (changes requested):
LGTM. Need to update https://github.com/patternfly/patternfly/blob/8ac660dade16889e9e1803e8816e19574aa56dba/src/patternfly/demos/Card/templates/card-demo--popover.hbs#L2-L17

That's used here - 
https://patternfly-pr-5370.surge.sh/components/card/html-demos#status-card-expanded-with-popover

And https://github.com/patternfly/patternfly/blob/8ac660dade16889e9e1803e8816e19574aa56dba/src/patternfly/components/LogViewer/examples/LogViewer.md?plain=1#L82-L102

That's used here - https://patternfly-pr-5370.surge.sh/extensions/log-viewer#with-popover-open

src/patternfly/components/Popover/popover-title-text.hbs
```diff
@@ -1,6 +1,22 @@
-<span class="pf-c-popover__title-text{{#if popover-title-text--modifier}} {{popover-title-text--modifier}}{{/if}}"
+<{{#if popover-title--type}}{{popover-title--type}}{{else}}h1{{/if}} class="pf-c-popover__title-text{{#if popover-title-text--modifier}} {{popover-title-text--modifier}}{{/if}}"
   {{#if popover-title-text--attribute}}
     {{{popover-title-text--attribute}}}
   {{/if}}>
+    {{#if popover--IsAlert}}
+    <span class="pf-u-screen-reader">
```
mcoker, line 6: If we allow the consumer to pass their own title in, would it then be up to them to include the sr text in their title to keep it in the heading element used for the title? We're including the sr text in the heading for them now:

* https://github.com/patternfly/patternfly-react/blob/01c2451bcb6af1dc7767d86e6884c509b6b41891/packages/react-core/src/components/Popover/PopoverHeader.tsx#L39-L45
* https://github.com/patternfly/patternfly-react/blob/01c2451bcb6af1dc7767d86e6884c509b6b41891/packages/react-core/src/components/Modal/ModalBoxTitle.tsx#L56-L69

I wonder if we could include the sr text in the title wrapper instead, then we'll include it regardless the title used.

src/patternfly/components/Popover/popover.scss
```diff
@@ -215,9 +208,12 @@
...
+.pf-c-popover__header {
+  margin-bottom: var(--pf-c-popover__header--MarginBottom);
+}
+
 .pf-c-popover__title {
   flex: 0 0 auto;
-  margin-bottom: var(--pf-c-popover__title--MarginBottom);
   font-family: var(--pf-c-popover__title--FontFamily);
   font-size: var(--pf-c-popover__title--FontSize);
   line-height: var(--pf-c-popover__title--LineHeight);
```
mcoker, line 219: Are we updating this the same way as empty state to move the title text styling to the title-text element (that can be replaced with a `<Title>` or whatever to opt-out of our styling)? 

There is also a `.pf-m-icon` class modifier on `__title` that creates a flex layout. Since the title text will be in `title-text` now, I think we can just make this element a flex layout by default.

src/patternfly/components/Popover/popover-title-text.hbs
srambach: @mcoker I don't know if that still works ok. Could this be a follow on issue so that this can move forward?

mcoker: @srambach yeah, though since it can effect the way the title updates are made in react (and other components with this structure), I wonder if we can decide now. If not, I think we can merge it and follow up. @thatblindgeye or @nicolethoen do you have an opinion on this?

We're updating the structure (in other components, too) so that the "title" element is just a container (`<div>`) for the title to be used. In that element, you can use the "title-text" element that PF provides to get the preferred element (heading, div, etc) and title text styling that comes from design, or you can use a `<Title>` (or whatever) component instead if you want to manage your own title.

Currently if the popover has an alert status, we include sr-only text in the popover title heading element. Since a user can potentially opt-out of our new title heading element (title-text) and use their own title, we can only predictably add sr-only text to the heading element if they use our title-text element. Seems like we could go a couple of ways with that.

1. Document that alert status popovers (or other components with this structure) need sr-only text in the title heading element, and we include that in the title-text element. If a user is providing their own title, they'll need to be sure and include the sr-only text themselves. This is probably my preference.
2. Move the sr-only text out of the heading element, and as the first thing in the title container (`<div>`). I'm not sure of the implications of that since the text is no longer in the heading element.

thatblindgeye: I think I would agree with your preference, @mcoker, that we document it and if a consumer wants to use their own custom thing, it's up to them to make sure they take any additional considerations. At least with VO, having it as part of the `title-text` has it announced as a single text content ("Default alert: Default popover title"), whereas moved outside of they become more separated.

One thing unrelated to this conversation is the `pf-u-screen-reader` class being applied. I believe we're trying to update usage of that class to just `pf-screen-reader`

src/patternfly/components/Popover/examples/Popover.md
```diff
@@ -51,12 +63,18 @@ import './Popover.css'
...
+    {{#> popover-close}}
       {{#> button button--modifier="pf-m-plain" button--attribute='aria-label="Close"'}}
         <i class="fas fa-times" aria-hidden="true"></i>
       {{/button}}
-      {{#> title titleType="h1" title--modifier="pf-m-md" title--attribute='id="popover-bottom-header"'}}
-        Popover header
-      {{/title}}
+    {{/popover-close}}
+    {{#> popover-header}}
+      {{#> popover-title popover-title--attribute='id="popover-bottom-header"'}}
```
mcoker, line 72: @thatblindgeye wdyt about this `id` (used as the value for `aria-labelledby` on the outer popover element) - it used to be on the heading element, but this change is moving it to the "title" `<div>`. Is that ok? Or should it preferably be on the heading element? If the latter, I think the same isue about the sr-only text applies here, too. If we prefer the `id` go on the heading itself, and a user provides their own heading, it would be up to them to add the `id`.

src/patternfly/components/Popover/examples/Popover.md
```diff
@@ -357,19 +437,20 @@ A popover is used to provide contextual information for another component on cli
...
 | `.pf-c-popover__arrow` | `<div>` |  Creates an arrow pointing towards the element the popover describes. **Required** |
 | `.pf-c-popover__content` | `<div>` |  Creates the content area of the popover. **Required** |
-| `.pf-c-button` | `<button>` |  Positions the close icon in the top-right corner of the popover. **Required** |
-| `.pf-c-popover__header` | `<header>` | Creates the popover header |
-| `.pf-c-popover__title` | `<h1>`,`<h2>`,`<h3>`,`<h4>`,`<h5>`,`<h6>`,`<div>` | Creates the popover title |
-| `.pf-c-title` | `<h1>`,`<h2>`,`<h3>`,`<h4>`,`<h5>`,`<h6>` |  Initiates a title. Always use it with a modifier class. See the [title component](/components/title) for more info.|
-| `.pf-c-popover__title-icon` | `<span>` | Creates the popover title icon |
-| `.pf-c-popover__title-text` | `<span>` | Creates the popover title text |
+| `.pf-c-popover__close` | `<div>` |  Positions the close icon in the top-right corner of the popover. **Required if there is a close button** |
+| `.pf-c-button` | `<button>` |  Creates the close button for the popover. |
```
mcoker, line 449: I don't think we need this.

src/patternfly/components/Popover/examples/Popover.md
```diff
@@ -357,19 +437,20 @@ A popover is used to provide contextual information for another component on cli
 | `aria-modal="true"` | `.pf-c-popover` | Tells assistive technologies that the windows underneath the current popover are not available for interaction. **Required**|
 | `aria-label="Close"` | `.pf-c-button` | Provides an accessible name for the close button as it uses an icon instead of text. **Required**|
 | `aria-hidden="true"` | Parent element containing the page contents when the popover is open. | Hides main contents of the page from screen readers. The element with `.pf-c-popover` must not be a descendent of the element with `aria-hidden="true"`. For more info, see [trapping focus](/accessibility/product-development-guide#trapping-focus). **Required** |
+| `.pf-screen-reader` | `.pf-c-popover__title-text` | Adds text to be read saying the alert status before the title. If `.pf-c-popover__title-text` is not used to create a title, this must be added within the title. **Required for popovers that are alerts** |
```
mcoker, line 440: It would go on a span in title-text. Here's the alert a11y docs for comparison.

```suggestion
| `.pf-screen-reader` | `.pf-c-popover__title-text <span>` | Adds text to be read saying the alert status before the title. If `.pf-c-popover__title-text` is not used to create a title, this must be added within the title. **Required for popovers that are alerts** |
```

mcoker (approved):
LGTM. I left a few comments, none are blockers. Feel free to file a follow up issue if you think any of them should be addressed.

PR #5371, opened 2023-02-03
fix(modal): add wrapper to close button

### LINKED ISSUE #5362: Modal - add __close element to wrap the close button
**Describe the feature**
The close button should be wrapped in a __close element so that it can be styled more closely so that it doesn't have unintended side effects.

srambach:
Fixes #5362 

Breaking change:
- Adds the `pf-c-modal-box__close` wrapper around the close button. Associated CSS variables are renamed.

### DIFF
--- a/src/patternfly/components/ModalBox/examples/ModalBox.md
+++ b/src/patternfly/components/ModalBox/examples/ModalBox.md
@@ -10,11 +10,7 @@ import './ModalBox.css'
 ### Basic
 ```hbs isFullscreen
 {{#> modal-box modal-box--attribute='aria-labelledby="modal-title" aria-describedby="modal-description"'}}
-  {{#> modal-box-close}}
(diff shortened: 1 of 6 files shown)


mcoker (changes requested):
The change looks good, but there are a lot of demos that also need to be updated. And I wonder if we can include the close button by default (or not) with an attribute? If an attribute doesn't make sense, `{{> modal-box-close}}` could include everything needed, as long as the icons/aria-labels for the close buttons are all the same.

mcoker (changes requested):
Just poking at the code a little, we have modals in these demos, too:

### Tabs
* https://patternfly-pr-5371.surge.sh/components/tabs/html-demos/modal-tabs/

### Table
* https://patternfly-pr-5371.surge.sh/components/table/html-demos/column-management-modal/

mcoker (approved):
lgtm!

PR #5377, opened 2023-02-07
fix(dependencies): update all dependencies to latest versions

nicolethoen:
Closes: #5119 

Note: It converts the gulp scripts to use esModules rather than CommonJS. This is because a number of the newest versions of dependencies are not compatible with CommonJS. Likely we will need to do the same in pf-react.


mcoker (approved):
NICE! 🥳

PR #5380, opened 2023-02-10
fix(stylelint): cleanup stylelint-scss

mcoker:
Removes [stylelint-config-recommended-scss](https://github.com/stylelint-scss/stylelint-config-recommended-scss) in favor of [stylelint-config-standard-scss](https://github.com/stylelint-scss/stylelint-config-standard-scss) per the recommended config's readme:

> This config is extended by the [stylelint-config-standard-scss](https://github.com/stylelint-scss/stylelint-config-standard-scss). That config is better suited to most users as it also turns on the stylistic rules in the [stylelint-scss plugin pack](https://github.com/stylelint-scss/stylelint-scss).

Updates some rules to match, including uniquely namespacing our global var and mixin names. Also removes `stylelint-scss` as it's already bundled.


src/patternfly/base/_globals.scss
```diff
@@ -137,5 +137,5 @@
 @import "./themes/dark/globals";
 
 @include pf-theme-dark($pf-theme-dark-prefix: ":root") {
-  @include globals;
+  @include pf-_globals;
```
srambach, line 140: I don't love this syntax. Is there a way around having `-_`?

srambach (commented):
I am not a fan of the `-_` in a few file names because it looks like a mistake. Is there a way around doing that?
Also, I can't verify the build file changes but other than those two things, looks ok to me.

src/patternfly/base/_globals.scss
mcoker: Agree, updated it to have a `pf-theme-dark` prefix and drop the `_` in the mixin name.

srambach (approved):
👍 thanks, that looks better.

PR #5387, opened 2023-02-16
fix(Progress): prevented empty desc and status from rendering

### LINKED ISSUE #4328: Progress - Don't render top progress description and status elements when empty
Currently if a progress description and/or status is not passed, and either would go above the progress bar, we're rendering empty elements. Those should not be needed if they aren't used.

thatblindgeye:
Closes #4328

### DIFF
--- a/src/patternfly/components/Progress/progress.hbs
+++ b/src/patternfly/components/Progress/progress.hbs
@@ -24,9 +24,12 @@
   {{#if progress__description}}
     {{> progress-description}}
   {{/if}}
-  {{#unless progress--no-measure}}
-    {{> progress-status}}
-  {{/unless}}
+  {{#if progress--no-measure}}
+    {{#if (concat progress--danger progress--success progress--warning)}}
+      {{> progress-status}}
+    {{/if}}
+    {{else}} {{> progress-status}}
+  {{/if}}
   {{> progress-bar}}
   {{#if progress__helper-text}}{{> progress-helper-text}}{{/if}}
   {{> @partial-block}}
--- a/src/patternfly/components/Progress/progress.hbs
+++ b/src/patternfly/components/Progress/progress.hbs
@@ -21,8 +21,12 @@
   {{#if progress--attribute}}
     {{{progress--attribute}}}
   {{/if}}>
-  {{> progress-description}}
-  {{> progress-status}}
+  {{#if progress__description}}
+    {{> progress-description}}
+  {{/if}}
+  {{#unless progress--no-measure}}
+    {{> progress-status}}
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Progress/progress.hbs
```diff
@@ -21,8 +21,12 @@
...
     {{{progress--attribute}}}
   {{/if}}>
-  {{> progress-description}}
-  {{> progress-status}}
+  {{#if progress__description}}
+    {{> progress-description}}
+  {{/if}}
+  {{#unless progress--no-measure}}
+    {{> progress-status}}
+  {{/unless}}
```
mcoker, line 29: I think we only want to remove the status element if there is no measure and no status icon. This update removes the icon from this example, which I don't think we want to do. https://patternfly-pr-5387.surge.sh/components/progress#failure-without-measure

This is the current v4 example that has an empty `__status` element - https://pf4.patternfly.org/components/progress#without-measure

Screenshot 2023-02-16 at 3 24 19 PM

Another small thing is that there is a margin on the status icon when there is no measure. What might be better is to turn `__status` into `display: flex; justify-content: flex-end; align-items: flex-start; gap: var(--whatever-spacer-that-margin-uses)`. That will also remove any space created by whitespace between the measure and status icon, too, and let the gap serve as the only source of space between those elements.

mcoker (changes requested):

src/patternfly/components/Progress/progress.hbs
thatblindgeye: Added the styling to __status, and also removed the `status-icon__MarginLeft` var since this is a breaking change issue and it would no longer serve any purpose.

The conditional rendering should be working as intended now.

src/patternfly/components/Progress/progress.hbs
```diff
@@ -21,8 +21,15 @@
...
-  {{> progress-status}}
+  {{#if progress__description}}
+    {{> progress-description}}
+  {{/if}}
+  {{#if progress--no-measure}}
+    {{#if (concat progress--danger progress--success progress--warning)}}
+      {{> progress-status}}
+    {{/if}}
+    {{else}} {{> progress-status}}
+  {{/if}}
```
mcoker, line 32: nit 

```suggestion
  {{#if progress--no-measure}}
    {{#if (concat progress--danger progress--success progress--warning)}}
      {{> progress-status}}
    {{/if}}
  {{else}}
    {{> progress-status}}
  {{/if}}
```

mcoker (approved):
LGTM! One small formatting change, not a blocker.

PR #5389, opened 2023-02-16
fix(Wizard): cleaned up CSS

### LINKED ISSUE #5143: Wizard - clean up CSS
A couple of things I spotted, but there may be other improvements, too

* Wizard close button class is on the plain button component - we should have a separate wizard element for the close button that the plain button component is placed inside.
* The wizard title class is on the title component - same as above. We may also consider not using the title component there, which ties the title markup to the presentation. For example, if we want to update the font size of the wizard title, that's a markup change - it maybe be better if we could handle that in CSS.

(check for other instances in the wizard as well)

thatblindgeye:
Closes #5143

Also updated a Wizard example to use the new markup of Empty State


src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -12,10 +12,12 @@ import './Wizard.css'
 ```hbs isFullscreen
 {{#> wizard wizard--id="wizard-basic"}}
   {{#> wizard-header}}
-    {{#> button button--modifier="pf-m-plain pf-c-wizard__close" button--attribute='aria-label="Close"'}}
-      <i class="fas fa-times" aria-hidden="true"></i>
-    {{/button}}
-    {{#> title title--modifier="pf-m-3xl pf-c-wizard__title"}}Wizard title{{/title}}
+    {{> wizard-close}}
+    {{#> wizard-title}}
```
mattnolting, line 16: When I see a repeated, standard blocks like this (esp `component__title-text`), I try to add a `.hbs` param to the `title` partial. So, instead of needing to repeat the `wizard-title-text` block, you could just say `{{> wizard-title wizard-title-text--value="My title"}}` and save a couple lines of code for each example. If it needs to be updated later, all the html config lives in one place instead of 5 or 100 or however many times it's used. 

So in the `title` partial:

```
<div class="pf-c-wizard__title{{#if wizard-title--modifier}} {{wizard-title--modifier}}{{/if}}"
  {{#if wizard-title--attribute}}
    {{{wizard-title--attribute}}}
  {{/if}}>
  {{#if wizard-title-text--value}}
    {{#> wizard-title-text}}
      {{{wizard-title-text}}}
    {{/wizard-title-text}}
  {{/if}}
  {{> @partial-block}}
</div>
```

src/patternfly/components/Wizard/wizard-title.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="pf-c-wizard__title{{#if wizard-title--modifier}} {{wizard-title--modifier}}{{/if}}"
+  {{#if wizard-title--attribute}}
+    {{{wizard-title--attribute}}}
+  {{/if}}>
+  {{#if wizard-title-text--value}}
+    {{#> wizard-title-text}}
+      {{{wizard-title-text--value}}}
+    {{/wizard-title-text}}
+  {{/if}}
```
srambach, line 9: It looks like if there was a partial block and `wizard-title-text--value` we'd get both - do we want that?

thatblindgeye: Good catch; I can't imagine we would. Just pushed an update to resolve that.

mcoker (changes requested):
LGTM! Though these demos also need to be updated (the first 5) - https://patternfly-pr-5389.surge.sh/components/wizard/html-demos

src/patternfly/components/Wizard/wizard-title.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="pf-c-wizard__title{{#if wizard-title--modifier}} {{wizard-title--modifier}}{{/if}}"
+  {{#if wizard-title--attribute}}
+    {{{wizard-title--attribute}}}
+  {{/if}}>
+  {{#if wizard-title-text--value}}
+    {{#> wizard-title-text}}
+      {{{wizard-title-text--value}}}
+    {{/wizard-title-text}}
+  {{else if @partial-block}}
```
srambach, line 9: This could probably just be an `else`, right?

thatblindgeye: Yep! I can open a followup issue for some other components that could use a similar update (Badge being one)

srambach: @thatblindgeye I'm sorry - I've learned that it's actually better to use the `{{else if @partial-block}}` so that you can use the syntax `{{> thing}}` rather than having to use an open and close `{{#> thing}}{{/thing}}`

mcoker (approved):
LGTM other than @srambach's comment about making this an `else if` again

https://github.com/patternfly/patternfly/blob/9c544388cf3ed518699fc4a433786d8e7a0c2812/src/patternfly/components/Wizard/wizard-title.hbs#L9

thatblindgeye:
@mcoker @srambach added back in the `else if`!

mcoker (approved):

PR #5393, opened 2023-02-17
feat(table): added support for sticky column on right

### LINKED ISSUE #5391: Bug - Table - Sticky column `th` needs right value
**Describe the problem**
Sticky column `th` needs right value to support sticky positioning on either side. 

https://www.patternfly.org/v4/components/table/html#sticky-column

**Expected behavior**
React users update variable `--pf-c-table__sticky-column--Right` instead of `right: {inline-value}`

No

mattnolting:
closes #5391

This PR adds `right: var(--pf-c-table__sticky-column--Right, revert);` to support righthand sticky column. If the table cell is the first child, `--pf-c-table__sticky-column--Left` is set to `0`. If the table cell is the last child, `--pf-c-table__sticky-column--Right` is set to `0`. There are no other instances in which either value should be `0`.

### DIFF
--- a/src/patternfly/components/Table/table-scrollable.scss
+++ b/src/patternfly/components/Table/table-scrollable.scss
@@ -1,7 +1,7 @@
 .pf-c-table {
   --pf-c-table__sticky-cell--MinWidth--base: #{pf-size-prem(200px)};
   --pf-c-table__sticky-cell--MinWidth: var(--pf-c-table__sticky-cell--MinWidth--base);
-  --pf-c-table__sticky-cell--ZIndex: var(--pf-global--ZIndex--md);
+  --pf-c-table__sticky-cell--ZIndex: var(--pf-global--ZIndex--xs);
   --pf-c-table__sticky-cell--Top: 0;
   --pf-c-table__sticky-cell--Right: auto;
   --pf-c-table__sticky-cell--Left: auto;
--- a/src/patternfly/components/Table/table-scrollable.scss
+++ b/src/patternfly/components/Table/table-scrollable.scss
@@ -1,55 +1,50 @@
 .pf-c-table {
-  --pf-c-table__sticky--cell-min-width--base: #{pf-size-prem(200px)};
-  --pf-c-table__sticky--cell--MinWidth: var(--pf-c-table__sticky--cell-min-width--base);
-  --pf-c-table__sticky--cell--ZIndex: var(--pf-global--ZIndex--md);
-  --pf-c-table__sticky--cell--Top: 0;
-  --pf-c-table__sticky--cell--BackgroundColor: var(--pf-global--BackgroundColor--100);
-  --pf-c-table__sticky--cell--m-border-right--before--BorderRightWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-table__sticky--cell--m-border-right--before--BorderRightColor: var(--pf-global--BorderColor--100);
-  --pf-c-table__sticky--cell--m-border-left--before--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-table__sticky--cell--m-border-left--before--BorderLeftColor: var(--pf-global--BorderColor--100);
-  --pf-c-table--m-sticky-header--thead__sticky--cell--ZIndex: calc(var(--pf-c-table__sticky--cell--ZIndex) + 1);
-  --pf-c-table__sticky-left--Left: 0;
-  --pf-c-table__sticky-right--Right: 0;
+  --pf-c-table__sticky-cell--MinWidth--base: #{pf-size-prem(200px)};
+  --pf-c-table__sticky-cell--MinWidth: var(--pf-c-table__sticky-cell--MinWidth--base);
+  --pf-c-table__sticky-cell--ZIndex: var(--pf-global--ZIndex--md);
+  --pf-c-table__sticky-cell--Top: 0;
+  --pf-c-table__sticky-cell--Right: auto;
+  --pf-c-table__sticky-cell--Left: auto;
+  --pf-c-table__sticky-cell--BackgroundColor: var(--pf-global--BackgroundColor--100);
+  --pf-c-table__sticky-cell--m-border-right--before--BorderRightWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-table__sticky-cell--m-border-right--before--BorderRightColor: var(--pf-global--BorderColor--100);
+  --pf-c-table__sticky-cell--m-border-left--before--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
+  --pf-c-table__sticky-cell--m-border-left--before--BorderLeftColor: var(--pf-global--BorderColor--100);
+  --pf-c-table__sticky-cell--m-right--Right: 0;
+  --pf-c-table__sticky-cell--m-left--Left: 0;
+  --pf-c-table--m-sticky-header__sticky-cell--ZIndex: calc(var(--pf-c-table__sticky-cell--ZIndex) + 1);
 
-  .pf-c-table__sticky-left,
-  .pf-c-table__sticky-right {
+  .pf-c-table__sticky-cell {
+    --pf-c-table--cell--Overflow: visible;
+    --pf-c-table--m-sticky-header--cell--ZIndex: var(--pf-c-table--m-sticky-header__sticky-cell--ZIndex);
+
     position: sticky;
-    z-index: var(--pf-c-table__sticky--cell--ZIndex);
-    min-width: var(--pf-c-table__sticky--cell--MinWidth);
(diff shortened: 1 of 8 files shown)


src/patternfly/components/Table/table-scrollable.scss
```diff
@@ -13,12 +12,21 @@
...
+    right: var(--pf-c-table__sticky-column--Right, revert);
+    left: var(--pf-c-table__sticky-column--Left, revert);
     z-index: var(--pf-c-table__sticky-column--ZIndex);
     min-width: var(--pf-c-table__sticky-column--MinWidth);
     background-color: var(--pf-c-table__sticky-column--BackgroundColor);
     background-clip: padding-box;
 
+    &:where(:first-child) {
+      --pf-c-table__sticky-column--Left: 0;
+    }
```
mcoker, line 24: Just another option. I'm not tied to either way, but figured I'd mention it

----
I'm not super familiar with sticky tables, but is it a use case that a sticky column could not be the first child of the row? Either the previous children are hidden in some way or maybe aren't sticky, but some child in the middle of the row is? Here is an example with the third column sticky- https://codesandbox.io/s/great-christian-69uy1g?file=/index.htmlhttps://codesandbox.io/s/lucid-ardinghelli-7fnkro?file=/style.css

Another use case is where multiple columns are sticky, with regular positioned columns in-between. Here is an example with the first and third columns sticky - https://codesandbox.io/s/great-christian-69uy1g?file=/index.html

Those will break with this update, since it's going into main.

It looks like the first sticky (not necessarily first child) should be `left: 0` by default, and any subsequent sticky (left) columns should have a left value set - assuming so, I wonder if it's better with a default of `left: 0`. I know that will mess up a right sticky column, since that will need to be unset, but with the page component, we have `pf-m-sticky-top` and `pf-m-sticky-bottom`. I'm wondering if maybe we could have left and right sticky, or maybe just `.pf-m-sticky-column` is left, and `pf-m-sticky-column-right` is right? Either now (to avoid breaking things), or for the long-term. I'm thinking it could be both, but I'm curious your opinion.

Looking at the react API, we set this for positioning:

#### First sticky column
```
isStickyColumn
```

#### Sticky siblings
```
isStickyColumn 
stickyLeftOffset="120px" // default is "0" from CSS
```

With this change as it is in the PR, I'm assuming we might do something like this?

#### First sticky column (on either side)
```
// Left
// ----
isStickyColumn 
// stickyLeftOffset="0" - would be needed if first sticky column isn't first-child

// Right
// -----
isStickyColumn 
// stickyRightOffset="0" - would be needed if first sticky column isn't last-child
```

#### Sticky siblings
```
// Left
// ----
isStickyColumn 
stickyLeftOffset="120px" // default is unset

// Right
// ----
isStickyColumn 
stickyRightOffset="120px" // default is unset
```

If we left the existing way it works as it is, and enabled `isStickyColumnRight` (or whatever), that could unset `left`, and set `right: 0;` and you'd have this API, and the codesandboxes I posted above would work as is.

#### First sticky column
```
// Left
// ----
isStickyColumn 

// Right
// -----
isStickyColumnRight
```

#### Sticky siblings
```
// Left
// ----
isStickyColumn 
stickyOffset="120px" // --pf-c-table__sticky-column--offset applies to left or right based on the modifier class

// Right
// ----
isStickyColumnRight
stickyOffset="120px" // --pf-c-table__sticky-column--offset applies to left or right based on the modifier class
```

To keep it from being breaking, we could keep `stickyLeftOffset` as it is and remove in v5, if that seems like an ok approach.

src/patternfly/components/Table/table-scrollable.scss
```diff
@@ -13,12 +12,21 @@
 
   .pf-c-table__sticky-column {
```
mcoker, line 13: unrelated - does this need to be scoped in `.pf-c-table?`

mcoker (commented):
Left a comment - I'm wondering if the move to targeting the first-last child might break someone.

src/patternfly/components/Table/table-scrollable.scss
mattnolting: Yes, it needs a higher specificity

src/patternfly/components/Table/table-scrollable.scss
mattnolting: Since React is resetting left/right variables, I think we can move this to v5 `pf-c-table__sticky-left/right`

thatblindgeye (approved):

src/patternfly/components/Table/table-scrollable.scss
```diff
@@ -1,46 +1,57 @@
...
-  --pf-c-table__sticky-column--Top: 0;
-  --pf-c-table__sticky-column--BackgroundColor: var(--pf-global--BackgroundColor--100);
-  --pf-c-table__sticky-column--m-border-right--before--BorderRightWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-table__sticky-column--m-border-right--before--BorderRightColor: var(--pf-global--BorderColor--100);
-  --pf-c-table__sticky-column--m-border-left--before--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-table__sticky-column--m-border-left--before--BorderLeftColor: var(--pf-global--BorderColor--100);
-  --pf-c-table--m-sticky-header--thead__sticky-column--ZIndex: calc(var(--pf-c-table__sticky-column--ZIndex) + 1);
-
-  .pf-c-table__sticky-column {
+  --pf-c-table__sticky--cell-min-width--base: #{pf-size-prem(200px)};
```
mcoker, line 2: Looking at our other components, if the "base" var is modifying a css property name (versus a concept, like "spacer"), it would be written as

```suggestion
  --pf-c-table__sticky--cell--MinWidth--base: #{pf-size-prem(200px)};
```

src/patternfly/components/Table/table-scrollable.scss
mcoker: In the CSS, I see `__sticky-left, __sticky-right` grouped with common styles, then both classes modify the same `left` property later in different selectors, and the vars are named `--pf-c-table__sticky--cell` - that says to me that sticky is the element, the sticky direction is the modifier, and then the vars could match:

```
[...]
--pf-c-table__sticky-header--Left: auto;
--pf-c-table__sticky-header--Right: auto;
--pf-c-table__sticky-header--m-left--Left: 0;
--pf-c-table__sticky-header--m-right--Right: 0;
[...]

.pf-c-table__sticky-header {
  // common styles
  position: var(--pf-c-table__sticky-header--Position);
  left: var(--pf-c-table__sticky-header--Left); // or just omit this
  right: var(--pf-c-table__sticky-header--Right); // or just omit this
  [...]
  
  &.pf-m-left {
    --pf-c-table__sticky-header--Left: var(--pf-c-table__sticky-header--m-left--Left); // or style as `left: var(--pf-c-table__sticky-header--m-left--Left);` if `left` is omitted on the parent.
  }

  &.pf-m-right {
    --pf-c-table__sticky-header--Right: var(--pf-c-table__sticky-header--m-right--Right); // or style as `right: var(--pf-c-table__sticky-header--m-right--Right);` if `right` is omitted on the parent.  }
}
```

That also implies a sticky element can be either variant, which I don't see why it couldn't be - in the same way sections and things are sticky when you reach them in scrollable regions - https://codesandbox.io/s/recursing-tristan-co5vc4?file=/fonts.css

Or are there other reasons you think they should be separate BEM elements?

mattnolting:
@mcoker Updated. `sticky-header` is being used to define a `thead`, so `sticky-cell` is more descriptive. LMK your thoughts.

src/patternfly/components/Table/table-scrollable.scss
```diff
@@ -1,43 +1,49 @@
...
-  --pf-c-table__sticky-column--m-border-right--before--BorderRightColor: var(--pf-global--BorderColor--100);
-  --pf-c-table__sticky-column--m-border-left--before--BorderLeftWidth: var(--pf-global--BorderWidth--sm);
-  --pf-c-table__sticky-column--m-border-left--before--BorderLeftColor: var(--pf-global--BorderColor--100);
-  --pf-c-table--m-sticky-header--thead__sticky-column--ZIndex: calc(var(--pf-c-table__sticky-column--ZIndex) + 1);
-
-  .pf-c-table__sticky-column {
+  --pf-c-table__sticky-cell--MinWidth--base: #{pf-size-prem(200px)};
+  --pf-c-table__sticky-cell--MinWidth: var(--pf-c-table__sticky-cell--MinWidth--base);
+  --pf-c-table__sticky-cell--ZIndex: var(--pf-global--ZIndex--md);
+  --pf-c-table__sticky-cell--Top: 0;
```
mcoker, line 5: doesn't look like this is used

mcoker (approved):
Just one nit - LGTM!

Issue #5264, opened 2022-11-29
Refactor click/select affordances for cards


mcarrano:
Follow up to https://github.com/patternfly/patternfly-design/issues/1211

### Updates to the card behavior/interactions:

1. Selectable cards:
**Updates**: There is a hover on the entire body of the card via mouse and keyboard. We are introducing a background color for more visual affordance on the hover, focus, and selected state. Clicking anywhere on the card would select the card and there is now more visual affordance to the selected state of the card with the addition of the background color and blue border around the body of the card. This mirrors Carbon's interaction for tiles.
![Selectable Card@1x](https://user-images.githubusercontent.com/98424339/213583757-ae226d0d-f0c0-405f-9d8e-ea801f90928e.png)

2. Single-selectable cards (new):
There is a hover on the entire body of the card via mouse and keyboard. We are introducing a background color for more visual affordance on the hover, focus, and selected state. Clicking anywhere on the card would select the card and there is now more visual affordance to the selected state of the card with the addition of the background color and blue border around the body of the card.
![Single Selectable Card@1x](https://user-images.githubusercontent.com/98424339/213583777-c13fa034-bd9d-4b6a-b315-b305572a7a47.png)

3. Clickable cards:
There is a hover on the entire body of the card via mouse and keyboard. We are introducing a background color for more visual affordance on the hover, focus, and selected state. Clicking anywhere on the card would trigger the action: ex. opening the primary-detail drawer, navigating to a new page, etc. **Updates**: There is now more visual affordance to the "clicked" state of the card with the addition of the background color and blue border around the body of the card.
![Clickable Card@1x](https://user-images.githubusercontent.com/98424339/213583818-a319803f-e7ad-47e9-ad31-30d3ae1bc8aa.png)
![Clickable Card w primary detail](https://user-images.githubusercontent.com/98424339/211892745-243edd4b-1012-400d-aecc-d36363767ffa.png)

4. Selectable & clickable cards:
Moving away from nested interactions and requiring separate affordances when there are multiple actions within a card. The body of the card being a trigger to an action will be reserved for when there is a single action within the card, like the above examples. 
There is a hover on the card title and checkbox via mouse and keyboard. Clicking on the title will trigger the action of either opening a primary-detail drawer, navigating to a new page, etc. If opening a primary-detail drawer, the card will appear with a blue border to indicate that the card has been clicked. To select the card for a global action, the user would need to click the checkbox. Below are examples of the different states in a white and gray background.
[...]

PR #5397, opened 2023-02-21
feat(Table/DataList): refactored visual affordances

### LINKED ISSUE #5347: Refactor click/select/expand affordances for data lists and tables
Follow up to https://github.com/patternfly/patternfly-design/issues/1211

### Updates to the table rows behavior/interactions:

1. Selectable table rows: (same behavior/interaction as current)
There is a hover on the checkbox via mouse and keyboard. Selection is done by clicking the checkbox and the visual affordance is the checkbox appears to show the row is selected. There is no hover on the row itself.

2. Expanding table rows:
There is a hover on the caret via mouse and keyboard. Expansion is done by clicking the caret. There is no hover on the row itself. **Updates**: Once the row expands, the visual affordance will no longer be the left blue border to indicate expansion, instead it will be the removal of the bottom border of the parent row and the growing of the row into the expanded row content. The expanded row content will also be indented to be in line with the first column content. 
Expandable-3

3. Clickable table rows [**primary-detail view only**]:
There is a hover on the row via mouse and keyboard (drop shadow and blue border). The row can be clicked to produce the action of opening the primary-detail drawer. **Updates**: Once the primary-detail drawer opens, the visual affordance will be updated to be a thicker left blue border (double the width of current) so it can be clearly seen in both cases where the table has content padding and where it doesn't.
Primary-detail no padding
Clickable-3

**Note that while these mocks are specifically for tables, the same changes should be applied to clickable/expandable data lists.**

thatblindgeye:
Closes #5347 

- Updated "hoverable"/"selectable" verbiage to "clickable". This includes class names, example text, etc. (several updates made will simply be lines with this update)
- To provide the additional ask of the border for indented datalist items, I removed some left padding and then added a left margin to the `pf-c-data-list` element instead.

Question regarding these components scss files: would the `*-grid.scss` files also need to be updated? It didn't look like the files were interfering with updates made in the `<component>.scss` file, but wanted to double check

### DIFF
--- a/src/patternfly/components/DataList/data-list.scss
+++ b/src/patternfly/components/DataList/data-list.scss
@@ -19,16 +19,15 @@
   --pf-c-data-list__item--m-expanded--before--BackgroundColor: var(--pf-global--active-color--100);
   --pf-c-data-list__item--m-selected--before--BackgroundColor: var(--pf-global--active-color--100);
   --pf-c-data-list__item--m-selected--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
-  --pf-c-data-list__item--m-selectable--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
-  --pf-c-data-list__item--m-selectable--hover--ZIndex: calc(var(--pf-c-data-list__item--m-selected--ZIndex) + 1);
-  --pf-c-data-list__item--m-selectable--hover--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
-  --pf-c-data-list__item--m-selectable--focus--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
-  --pf-c-data-list__item--m-selectable--active--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
-  --pf-c-data-list__item--m-expanded--m-selectable--before--BackgroundColor: var(--pf-global--active-color--300);
+  --pf-c-data-list__item--m-clickable--OutlineOffset: calc(-1 * var(--pf-global--spacer--xs));
+  --pf-c-data-list__item--m-clickable--hover--ZIndex: calc(var(--pf-c-data-list__item--m-selected--ZIndex) + 1);
+  --pf-c-data-list__item--m-clickable--hover--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
+  --pf-c-data-list__item--m-clickable--focus--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
+  --pf-c-data-list__item--m-clickable--active--BoxShadow: var(--pf-global--BoxShadow--sm-top), var(--pf-global--BoxShadow--sm-bottom);
   --pf-c-data-list__item--BorderBottomColor: var(--pf-global--BorderColor--300);
   --pf-c-data-list__item--BorderBottomWidth: #{pf-size-prem(8px)};
-  --pf-c-data-list__item--m-selectable--hover--item--BorderTopColor: var(--pf-c-data-list__item--BorderBottomColor);
-  --pf-c-data-list__item--m-selectable--hover--item--BorderTopWidth: var(--pf-c-data-list__item--BorderBottomWidth);
+  --pf-c-data-list__item--m-clickable--hover--item--BorderTopColor: var(--pf-c-data-list__item--BorderBottomColor);
+  --pf-c-data-list__item--m-clickable--hover--item--BorderTopWidth: var(--pf-c-data-list__item--BorderBottomWidth);
   --pf-c-data-list__item--sm--BorderBottomWidth: var(--pf-global--BorderWidth--sm);
   --pf-c-data-list__item--sm--BorderBottomColor: var(--pf-global--BorderColor--100);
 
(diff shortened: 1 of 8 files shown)


mcoker:
Just a side comment - I wonder if we should call this kind of thing "actionable" instead of "clickable." I think people get it regardless, but actionable seems to be a better description and isn't tied to a device interaction. Would also apply to @srambach's card PR - https://github.com/patternfly/patternfly/issues/5264

mcarrano (changes requested):
@thatblindgeye for the most part this looks good. There were just a couple of nits that I have that are listed below:

* For both the data list and table, clickable should not be confused with selection. You've updated the terminology, and that's good. But the examples still confuse the two. When clickable rows are used, only one row should be active at any time. To select multiple rows, the checkbox selection should be used. I know this is more of a JS behavior, but would be good if the examples can be changed to reflect this.

* for the nested data list, it feels odd to have the double border at the end of the expanded section as in the screen shot. Maybe the last nested item does not need a bottom border? Open to suggestions.

Screenshot 2023-02-21 at 3 53 03 PM

thatblindgeye:
> for the nested data list, it feels odd to have the double border at the end of the expanded section as in the screen shot. Maybe the last nested item does not need a bottom border? Open to suggestions.

Yeah I agree visually it does look kind of odd. On the one hand it *does* kind of act as a signifier that it's the end of the item, but on the other hand removing that indented border at the end and just using that wider border above "Secondary content" does, too.

Comparing the above screenshot with the following where that last indented border is removed:

Nested datalist expansion without bottom border

mcarrano:
I like this better :)

thatblindgeye:
@mcarrano Related to @mcoker comment above, would there be any preference to:

- using "actionable" over "clickable" as the verbiage and class name
- if we want to separate "selectable" from "clickable"/"actionable", and if selection is more meant for the checkbox approach, updating the classname from `pf-m-selected` to `pf-m-active` for the left border "selected" styling (this would also be a question for Michael)

mcarrano:
@thatblindgeye these are all good points. In this case, perhaps "actionable" and "active" are better terms than "clickable" and "clicked?"  @mmenestr and @edonehoo as first-class wordsmithers, what do you think? Is the word "actionable" to describe a card, list item, or table item that can be clicked to trigger some related event better than "clickable?" In the past we referred to these cards or rows as selectable, but I'm trying to avoid confusion with things that can be selected via a checkbox. Thoughts?

mmenestr:
@mcarrano I actually feel like clickable might make more sense only because checkbox selection is usually related to "actions". In other words, "actionable" is more broad in that both clicking on a row, or selecting a row via checkboxes lead to actions or being able to perform an action on the table.

mcoker:
I don't think we need to solve this in this PR, but my only concern with the name "active" is a potential conflict as we add states to things that have this "active" state as well as an `:active` state (buttons and things, when the thing is being pressed on via mouse, keyboard, or whatever). Technically a clickable/actionable row could have an `:active` state, too. Thinking in terms of tokens or just how to refer to that, you may want `$button--active--color` or say "active button color" and need to distinguish between those two states.

FWIW, we use "current" in nav, tabs, simple-list, wizard (steps nav), for when a single thing in a list of things has been chosen. I've wondered if maybe we should piggyback on existing distinctions for these names, like referring to the `aria-` attrs - `aria-[selected, multiselectable, pressed, current, checked, expanded, etc]`

mcarrano:
> FWIW, we use "current" in nav, tabs, simple-list, wizard (steps nav), for when a single thing in a list of things has been chosen. I've wondered if maybe we should piggyback on existing distinctions for these names, like referring to the aria- attrs - aria-[selected, multiselectable, pressed, current, checked, expanded, etc]

Yeah, maybe "current" make sense since these clickable items are really intended to be used like navigation.

thatblindgeye:
@mcoker @mcarrano for now I updated the examples to just use "Clicked" to differentiate it from a checkbox selected. "Current" might work, though, especially as in "currently clicked/active/etc". 

From an AT standpoint, I'm trying to think what would make the most sense as far as an ARIA attribute. If the "clickable" functionality is only meant to be for a primary detail view, then `aria-expanded` might make sense; that might require differentiate between "inline expandsion" and "drawer expansion". `aria-current` seems like it worked okay in a quick test, announcing whichever datalist item as "current item".

The functionality in a table throws a bit of a wrench, though, since a table has been a little more difficult in making accessible when it comes to being "clickable" in the past.

mcarrano (approved):
This looks good. Thanks for making these changes @thatblindgeye !

srambach:
Am I reading the issue correctly that for clickable rows, "There is a hover on the row via mouse and keyboard (drop shadow and blue border)" or has that been revised?

srambach (approved):
LGTM 🥇

src/patternfly/components/DataList/data-list.scss
```diff
@@ -465,11 +458,26 @@
...
+      border-bottom: 0;
+    }
+
+    .pf-c-data-list__item-row {
+      --pf-c-data-list__item-row--PaddingLeft: 0;
+    }
+
+    .pf-c-data-list__expandable-content-body {
+      --pf-c-data-list__expandable-content-body--PaddingLeft: 0;
+    }
```
mcoker, line 479: Just looking at this change from another PR - I wonder if this could be re-written as:

```
.pf-c-data-list { // this is the main selector at the root of this stylesheet
  & { // compiles to .pf-c-data-list .pf-c-data-list. I'm not sure you have to specify it's a data-list in an expandable content body, as I think that's the only use case of a nested data-list
    --pf-c-data-list__item-row--PaddingLeft: 0; // this var and the one below can be set at the component level as far as I can tell
    --pf-c-data-list__expandable-content-body--PaddingLeft: 0;

    margin-left: var(--pf-c-data-list--data-list--MarginLeft); // update the var name to indicate it's nested

    .pf-c-data-list__item:last-child {
      --pf-v5-c-data-list__item--BorderBottomWidth: 0; // if we were to do something like change this from a `border` on the last item to a border on a pseudo element of the last item or something, we would re-use this `--BorderBottomWidth` var and this override would still work.
    }    
```

PR #4717, opened 2022-03-03
fix(table): right align table action cells

mattnolting:
closes [#4716](https://github.com/patternfly/patternfly/issues/4716)

Preview: 
- https://patternfly-pr-4717.surge.sh/components/table#overflow-menu-usage-mobile 

To view update, remove one of the overflow menu buttons

Screenshot 2023-02-15 at 11 28 10 AM

PR #5399, opened 2023-02-27
fix(table): reverts "right align table action cells"

mcoker:
Reverts patternfly/patternfly#4717

#4717 was an old PR and accidentally opened against main, needs to come out since this is a breaking change intended for v5


mattnolting (approved):
LPTM 👍

Issue #5361, opened 2023-01-04
Bug - [FormControl] - [readonly elements in dark mode need to change background similar to disabled]


KKoukiou:
**Describe the problem**
readonly elements in dark mode have the same background color like plain ones.

**Expected behavior**
readonly dark mode styling for form control elements to be similar to disabled, like it is in light theme. 

**Is this issue blocking you?**
no

**Screenshots**
![Screen Shot 2023-01-04 at 14 40 56](https://user-images.githubusercontent.com/14921356/210567632-a2dfaf0a-1710-49a8-8377-3f314775c320.png)

PR #5402, opened 2023-02-28
fix(form-control): updated readonly, disabled colors

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5361

@mceledonia @lboehling this updates the form control/text input. What about other things - buttons (specifically control buttons, which match form control styles), menu toggles - anything else? 

FWIW here is a readonly and disabled text input after our disabled buttons:

Screenshot 2023-02-27 at 7 06 45 PM

And some menu toggles, with a readonly and disabled text input after the disabled menu toggle example

Screenshot 2023-02-27 at 7 07 27 PM


mcarrano (approved):
I think this looks good now. I actually wish we had a similar differentiation in light theme. What do you think @mceledonia @lboehling ?

PR #5408, opened 2023-03-02
fix(banner): replaced status with color

### LINKED ISSUE #5162: Banner - remove status variants in favor of colors
Similar to labels, banner colors can be used without conveying danger/warning/success/etc status. We recently added the color variants to the component, so we can remove the status variants.

Note - keep the status examples, we're just removing the status classes.

thatblindgeye:
Closes #5162


mcoker (approved):
nice!

PR #5409, opened 2023-03-02
WIP feat(card): changes card affordances

srambach:
Very, very, very WIP. If I don't come back from vacation, have fun with it.


mcoker:
Just a couple of quick examples, though I didn't spend much time on it so hopefully it's helpful.

Using a label in the card, this is one way https://codepen.io/mcoker/pen/QWVOWxp
* The link in the card points to this Kitty Giraudel article, which outlines the basic technique - https://kittygiraudel.com/2022/04/02/accessible-cards/
* `:focus-visible` keeps focus styling on keyboard interaction, not clicks
* I think we would want to use the check component there, and it's not super clear how to do that currently. We could use it as it is, and just apply all the styles to the check component label. Or we could try and support a `<input class="pf-c-check pf-m-standalone">` to get the checkbox, which seems like a useful enhancement, and then have a `pf-c-card__[label-element]` element for the label we're styling, which keeps them as siblings and allows the label to be styled by the checkbox state.

Another way might be to add a `tabindex` to the card and manage the selected state styling via a class. A label could still be used on the checkbox directly if we want something like the ability to focus on the outer card wrapper as well as focus on the check input in the card to select the card. Here's a basic demo of what that might look like - https://codepen.io/mcoker/pen/bGxYvWd

srambach:
Status: Using `.pf-c-card__label` which is the label of an input to handle the affordances.
Known issues:
- clickable cards don't quite have keyboard focus correct
- lots of specificity that needs to be reduced
- lots of variable sharing right now between the variants - leave or split up?
- can probably use some variable fallbacks
- hbs can be improved to better pass in options
- probably many other things

note: using the inspector and trying to check the focus can cause weird flashes of background color and odd borders that don't appear when not inspecting.

tiyiprh:
@srambach This is looking great! As @mcarrano pointed out, the clickable-selectable interaction is partially correct. The hover state should be on the card title and the checkbox but not on the whole card body. When the title is selected, the card would display a clicked state which mirrors the selectable and clickable cards (darker background and blue border). When the card is selected, by clicking the checkbox, its selected state would just be the checked checkbox (no background or border).

srambach:
> @srambach This is looking great! As @mcarrano pointed out, the clickable-selectable interaction is partially correct. The hover state should be on the card title and the checkbox but not on the whole card body. When the title is selected, the card would display a clicked state which mirrors the selectable and clickable cards (darker background and blue border). When the card is selected, by clicking the checkbox, its selected state would just be the checked checkbox (no background or border).

@mcarrano @tiyiprh I updated the component - want to take another look? I added a `pf-m-current` modifier for the card that is clicked so that you can see what it would look like - the React is going to need to take the click on the text and make the card "clicked," i.e. current. WDYT?

tiyiprh:
@srambach Hi Sarah, thanks for making those updates - I think we are almost there! The "clicked" card interaction and states via clicking on the title looks correct. The selecting of the card should be done by clicking on the checkbox only, not the whole body of the card. This is different from the selectable only or clickable only cards. Here is an updated image showing the hover states of what can be clicked on for the selectable and clickable card. The focus states should also be on the title and checkbox only, not the whole card.

Selectable   Clickable Card

thatblindgeye (changes requested):
Some initial comments below. We should also add in the `pf-m-disabled` class to the docs table.

Regarding the clickable AND selectable implementation... Overall I think it works well, but from an a11y standpoint it's tricky to convey something as selected via input as well as have a separate "selection" via clicking. It's even more tricky because the clickable only example uses a hidden input and you can technically click anywhere on the card , while the clickable and selectable example only the card title can be clicked for that "clickable" aspect.

@mcarrano for the clickable/selectable implementation, is the intent only for that to be used for a primary detail view where clicking it causes a drawer or something similar to expand open?

I want to mull that over a bit more, but wanted to share some initial thoughts about the clickable/selectable cards.

mcarrano:
> @mcarrano for the clickable/selectable implementation, is the intent only for that to be used for a primary detail view where clicking it causes a drawer or something similar to expand open?

Yes, that's mostly the case @thatblindgeye . We struggled a lot with this in design and finally arrived at the current solution as the best option among a set of less than optimal choices. If you have a better idea of how to handle this, I'm open to ideas.

thatblindgeye (approved):
We'll have to see how interaction goes in the React implementation, but this is looking good so far. If anything we can follow up if we notice any issues that arise.

@mcarrano One thing is the difference between clickable only and clickable+selectable. Clickable only using the hidden input and label works to make the whole card clickable (if it should be a link it's iffy; from tinkering around we should be able to provide a link that is conveyed to AT, and in React I think we could trigger a link opening when clicking the label). For clickable+selectable we're using the card title only as the clickable aspect since we can't put a hidden input in there (that'd be too confusing for users I imagine to have a visible checkbox *and* hidden radio being announced to them).

Considering a11y issues, this may be one of the better approaches to go about it, but having  a clickable only and clickable+selectable, with the clickable aspect being able to navigate to a link or trigger an action for both types of cards, is difficult to pin down with how different the interactions are intended to be. For the clickable only we'd also have to recommend that no other interactive content is placed inside the card (links, buttons, dropdown, etc).

mcarrano (approved):
This looks great. I think that you have it all correct now @srambach . @tiyiprh perhaps just give one last look to also confirm?

> For the clickable only we'd also have to recommend that no other interactive content is placed inside the card (links, buttons, dropdown, etc).

I also agree with the comment by @thatblindgeye . Someplace we need to make it clear that when the entire card is clickable, it behaves like a button and you would not place other interactive elements inside of it. If that is a requirement, then the clickable/selectable is the preferred pattern to follow so that each interactable items takes focus separately.

tiyiprh:
This all looks good to me! Thanks for making those updates! 

I also agree with the rest of the comments above by @thatblindgeye and @mcarrano regarding no other interactive content being placed inside the card and that a clickable/selectable pattern should be used if that is necessary.

src/patternfly/components/Card/card--hidden-input.hbs
```diff
@@ -0,0 +1,6 @@
+{{#> radio radio--modifier="pf-m-standalone"}}
+  {{#> radio-input radio-input--modifier="pf-c-card__hidden-input pf-c-card__input pf-screen-reader" radio-input--attribute=(concat 'id="' card--id '-sr-only-radio" name="' card--id '-sr-only-radio" aria-labelledby="' card--id '" ' card--hidden-input--IsDisabled )}}{{/radio-input}}
```
mcoker, line 2: Is it necessary to have this as `.pf-c-radio__input.pf-c-card__input.pf-c-card__hidden-input`? We've avoided using multiple component classes between components on a single element, and have backed out the handful of places where we have done that, including some in this v5 release. For example, separating `.pf-c-skip-to-content` and `.pf-c-button` - https://github.com/patternfly/patternfly/issues/4972

src/patternfly/components/Card/card.scss
```diff
@@ -108,57 +136,128 @@
...
-      box-shadow: var(--pf-c-card--m-hoverable--hover--BoxShadow);
-    }
+  //////
+  // SELECTABLE CARDS, CLICKABLE CARDS
+  &.pf-m-selectable,
+  &.pf-m-clickable {
+    isolation: isolate;
+
+    // box-shadow will come from the card__label::before instead of the card so remove this so there's isn't a double shadow
+    box-shadow: none;
```
mcoker, line 156: nit - we aren't consistent with this, but comments about a specific property should be inline with the property and it allows you to maintain the spacing

```suggestion
    isolation: isolate;
    box-shadow: none; // box-shadow will come from the card__label::before instead of the card so remove this so there's isn't a double shadow
```

src/patternfly/components/Card/card.scss
```diff
@@ -108,57 +136,128 @@
...
   display: flex;
   flex-direction: column;
   background-color: var(--pf-c-card--BackgroundColor);
   box-shadow: var(--pf-c-card--BoxShadow);
 
-  &.pf-m-hoverable {
-    &:hover {
-      box-shadow: var(--pf-c-card--m-hoverable--hover--BoxShadow);
-    }
+  //////
```
mcoker, line 149: nit - do you need this?

```suggestion
```

src/patternfly/components/Card/card.scss
```diff
@@ -108,57 +136,128 @@
...
     }
+  }
 
-    &:focus {
-      box-shadow: var(--pf-c-card--m-selectable--focus--BoxShadow);
+  // Hover the card (label)
+  .pf-c-card__label:hover {
+    --pf-c-card--BackgroundColor: var(--pf-c-card--m-selectable--hover--BackgroundColor);
+    --pf-c-card--BoxShadow: var(--pf-c-card--m-selectable--hover--BoxShadow);
+  }
```
mcoker, line 179: Can you move all of these selectors out of `.pf-c-card`? That will reduce the specificity. Doesn't seem like they need to be nested here.

src/patternfly/components/Card/card.scss
```diff
@@ -108,57 +136,128 @@
...
+
+  // Selected card
+  .pf-c-card__input:where(:checked) ~ .pf-c-card__label,
+  &.pf-m-selected {
```
mcoker, line 183: This implies needing to be nested in `.pf-c-card`, but you could either make this `.pf-c-card.pf-m-selected` or put an `@at-root` at-rule before the `.pf-c-card__input:where(...` line, which will compile to `.pf-c-card__input:where(...` instead of `.pf-c-card .pf-c-card__input:where(...`

src/patternfly/components/Card/card--hidden-input.hbs
mcoker: Actually just poking around, is `.pf-c-card__hidden-input` relevant/needed? doesn't seem to do anything and isn't documented.

mcoker (approved):
A few nits, all can be follow ups if you think they're needed. The important one we may want to address is sharing card and radio component classes on a single element, since that may impact the way it's built in react if we end up changing it.

src/patternfly/components/Card/card--hidden-input.hbs
srambach: @mcoker  To clarify, are you suggesting changing from:
```
    div.pf-c-card__actions
      div.pf-c-radio.pf-m-standalone
          input.pf-c-radio__input.pf-c-card__input.pf-screen-reader
          label.pf-c-radio__label.pf-c-card__label
```
to something like:
```
    div.pf-c-card__actions
      div.pf-c-radio.pf-m-standalone
          div.pf-c-card__input
            input.pf-c-radio__input.pf-screen-reader
          div.pf-c-card__label
            label.pf-c-radio__label
```

srambach: Ok - before you've been reluctant to select one component from another app I wasn't sure what you meant.

srambach:
@mcoker I think I addressed most of your concerns, definitely rearranged the structure as you asked. I think now we can plan to tweak and improve in the future but let React get started?

mcoker (approved):
🚀

thatblindgeye (approved):

Issue #5175, opened 2022-10-07
Potential namespace conflict with "default" in status prop enums


mcoker:
Technically "default" is a status indicated by `.pf-m-default` (cyan), but "default" is also a special word in many of our enums that applies whatever the prop's default is (as if the prop were undefined). It seems like we usually (though not always) have a way to provide a value to set the prop to undefined?

We have status props that are optional (meaning you should be able to set the prop to it's default as if it were undefined), but also need "default" as a status value (to apply `.pf-m-default`), and they can't use the same name. Some examples

### Popover
https://github.com/patternfly/patternfly-react/blob/b44f6c31e0ab822af9626fd16912d4eb6ba01f3e/packages/react-core/src/components/Popover/Popover.tsx#L49

### Notification drawer list item
https://github.com/patternfly/patternfly-react/blob/b44f6c31e0ab822af9626fd16912d4eb6ba01f3e/packages/react-core/src/components/NotificationDrawer/NotificationDrawerListItem.tsx#L19

### Modal/Modal content
https://github.com/patternfly/patternfly-react/blob/b44f6c31e0ab822af9626fd16912d4eb6ba01f3e/packages/react-core/src/components/Modal/Modal.tsx#L68

https://github.com/patternfly/patternfly-react/blob/b44f6c31e0ab822af9626fd16912d4eb6ba01f3e/packages/react-core/src/components/Modal/ModalContent.tsx#L75

### Icon
https://github.com/patternfly/patternfly-react/blob/b44f6c31e0ab822af9626fd16912d4eb6ba01f3e/packages/react-core/src/components/Icon/Icon.tsx#L20

### Alert

This one is a little more tricky, and kind of. `<Alert>` on its own without a status modifier class provides _default status_ styling (cyan). However, there is no `.pf-m-default` class to specify default status styling. It's implied that if you want to convey "default" status, don't use a modifier. This is because there is no non-status alert (just regular/black text/icons). I don't know if there should be? That said, it's a bit inconsistent with some of our other components where we offer optional size variations where whatever the default size is (no size variation) is also offered as a specific variation. I think that's useful from a general theming perspective. It also lets us update the default size while still offering whatever the previous default was as an option.

So `variant="default"` undoes any status variation/class, but also provides default status styling since that's the default from the core CSS.

I guess we could
* leave this as is
[...]

PR #5418, opened 2023-03-08
fix(global): rename default status color to custom

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5175

Also makes a status variation required for the alert component. Previously the alert with no variant would give you the "default", aka "custom" status styling. Now the alert without a variation is unstyled and you have to add `.pf-m-custom`.


src/patternfly/components/Alert/alert.scss
```diff
@@ -119,6 +124,13 @@
   border-top: var(--pf-c-alert--BorderTopWidth) solid var(--pf-c-alert--BorderTopColor);
   box-shadow: var(--pf-c-alert--BoxShadow);
 
+  &.pf-m-custom {
+    --pf-c-alert--BorderTopColor: var(--pf-c-alert--m-custom--BorderTopColor);
+    --pf-c-alert__icon--Color: var(--pf-c-alert--m-custom__icon--Color);
+    --pf-c-alert__title--Color: var(--pf-c-alert--m-custom__title--Color);
+    --pf-c-alert--m-inline--BackgroundColor: var(--pf-c-alert--m-inline--m-custom--BackgroundColor);
```
thatblindgeye, line 131: Doesn't look like `--pf-c-alert--m-inline--m-custom--BackgroundColor` is defined anywhere

src/patternfly/components/Popover/popover-title-text.hbs
```diff
@@ -4,7 +4,7 @@
   {{/if}}>
     {{#if popover--IsAlert}}
       {{#> screen-reader}}
-      {{#if popover--IsDefaultAlert}}
+      {{#if popover--IsCustomAlert}}
         Default
```
thatblindgeye, line 8: Similar to above, would we want this text to be "Custom" instead?

thatblindgeye (changes requested):

thatblindgeye (commented):
Looks good! Only other thing is the screenshot for the "custom alert popover" example needs to be updated.

mcoker:
@thatblindgeye great catches! The full-page example screenshots are updated prior to a release, usually with the versioned commit that adds the release notes. Though you can run the job and push to this PR if you'd like! The instructions are at https://github.com/patternfly/patternfly/#update-screenshots Note that it may update more screenshots than this one since that job isn't run with each change and there have been a lot of changes since the last run.

src/patternfly/components/NotificationDrawer/examples/NotificationDrawer.md
srambach, line 1: Line 127 - Should add '.pf-m-custom' as an acceptable modifier (or just remove the list and simply say always to use it with a state modifier)

srambach (approved):
:+1 LGTM

Issue #5313, opened 2023-01-10
Remove WOFF fonts


jonkoops:
Older font formats were removed under #4210, but the decision was made to keep the WOFF files as Internet Explorer was [unofficially supported](https://github.com/patternfly/patternfly/pull/4210#issuecomment-877334727). With a new major version coming up, it might be time to drop these font files to reduce the bundled size of applications.

Issue #2493, opened 2019-12-04
Suggested separation of normalize vs. base theme styles


starryeyez024:
In the _globals.scss file, there exists both a set of normalize styles (above line 90) and a set of PatternFly opinionated styles (though arguably the box-sizing styles should be in the normalize section). If these could be split out into separate @if statements, that would make extending the parts you need easier. :) 
https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/_globals.scss#L90

PR #5435, opened 2023-03-20
fix(global): remove woff, use where with globals, cleanup

mcoker:
* Fixes https://github.com/patternfly/patternfly/issues/5313 - removes support for woff font files
* Fixes https://github.com/patternfly/patternfly/issues/2493 - separates reset from normalize styles that can be excluded independently
* Adds `:where()` around global (normalize and reset) styles. This removes the specificity from these rules, so they will not fight for specificity with any selector styles with a specificity > 0 - from our components or any user overrides.
  * This would be really useful with the old "shield" styles that applied `[class*=pf-c-], [class*=pf-c-]::before, [class*=pf-c-]::after { padding/margin: 0; background: transparent; }`, but those were removed for v5 in another PR. However, I think this may be good practice in general for PF and users to be able to easily override any base/global styles without having to fight for specificity or put styles in an appropriate layer (in the future).
  * A potential problem I see here is if a user loads some third party library that styles `h1, a, body, etc` and loads PF afterward - our styles will currently override those styles (assuming our specificity is >= and the rules match 1:1), and they will not if we wrap our styles in `:where()`. Though I think that kind of conflict is something that should be handled on a case by case basis when importing multiple global styles as there are likely multiple conflicts like that regardless of using `:where()` 
* Removes bootstrap variables file
* Removes card and button styles from light/dark placeholder. The button styles were never intended to stay there, and the card style is too specific to an old concept of a dashboard layout.
* Renames the pf-*-no-reset.scss files to no-globals, since reset has a more specific meaning.


srambach:
I see a difference in masthead -
image

srambach:
There's a slightly different background on dark theme card demos
image

mcoker:
> There's a slightly different background on dark theme card (actually I think many dark theme) demos

@srambach that's from removing the `background-color` from `<body>` since those full page examples are on a blank page.

> I see a difference in masthead -

That's to be expected, the old styles were from previous design iterations on buttons.

srambach (approved):
OK, looks good then!

srambach:
Actually though - that primary button doesn't pass the contrast check.

mcoker:
@srambach opened https://github.com/patternfly/patternfly/issues/5436 as a follow-up

mattnolting (approved):
LPTM 🔥

PR #5440, opened 2023-03-24
Add calendar month to date and time group

mcarrano:
Want to move the Calendat month component into the Date and Time group with other like components.


nicolethoen (approved):

nicolethoen:
Should we move `Timestamp` into the `Date and time` section while we are doing this?

Issue #5321, opened 2023-01-12
remove beta tags for v5


nicolethoen:
Banner (html)
Charts: Patterns
Date picker
Description list (html)
Dropdown: bulk select loading state
Dropdown - next
Helper text demo (html)
Icon
LoginPage: header utilities
MenuItem: hasCheck
Page demos (html)
Password strength demo (html)
Progress stepper
removeFindDomNode
SearchInput
Select next
Select: isCreateSelectOptionObject
Select: loadingVariant
Select: isInputFilterPersisted
Tabs: dynamic
Tabs: horizontal overflow
Tabs: actions
Tabs: close button props
Text area: icon sprite
Text input group
Tile
Truncate
Wizard next

PR #5441, opened 2023-03-27
fix(docs): promote various components/features out of beta for v5

nicolethoen:
Closes: https://github.com/patternfly/patternfly/issues/5321
 
Date picker
 Dropdown: bulk select loading state
 Icon
 Log viewer
 Page demos (html)
 Progress stepper
 SearchInput
 Tabs: dynamic
 Tabs: horizontal overflow
 Tabs: actions
 Tabs: close button props
 Text input group
 Tile
 Truncate

### DIFF
--- a/src/patternfly/components/DatePicker/examples/DatePicker.md
+++ b/src/patternfly/components/DatePicker/examples/DatePicker.md
@@ -1,6 +1,5 @@
 ---
 id: 'Date picker'
-beta: true
 section: components
 subsection: date-and-time
 cssPrefix: pf-c-date-picker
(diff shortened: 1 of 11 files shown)


mcarrano (approved):
Looks good. So is the assumption that everything that was previously beta will now be promoted as we transition to v5?

nicolethoen:
Not everything? I will draft up a list of what will be left to discuss with leads in the next week or two

srambach:
> @srambach my vote is to make that change, it should work better for everyone. We would also want to update any docs about needing `height: 100%`, like https://github.com/patternfly/patternfly-react/blob/ab1d27353a3aed7cc3c2b83c4278edc66a63c2de/packages/react-core/src/demos/Page.md?plain=1#L28

Added a core issue here: https://github.com/patternfly/patternfly/issues/5448
And react here: https://github.com/patternfly/patternfly-react/issues/8873

PR #5443, opened 2023-03-27
feat(helper-text): make icon required for status

### LINKED ISSUE #5438: Helper text - make status icon required for status variants
Warning, success, and error should include the status icon by default. We'll want to:

* Update the helper text component examples and demos to add those icons.
* Update the docs to indicate the icon is required with a status variant.
* Update any other component examples/demos that use helper text to ensure they have an icon if a status is set
  * Looks like we already have params for `--Is[Status]` and also `--IsDynamic` (icons are also required for dynamic items), and icons are included with `--HasIcon`, so I'm thinking we can just test for `--Is[Status], --IsDynamic, --HasIcon`, and include the icon automatically if any are set. One way to do that is with the concat helper like `{{#if (concat --IsError --IsWarning --IsSuccess --IsDynamic --HasIcon) }} {{> icon}} {{/if}}`. Then we just need to review our uses of helper text in all of the component docs and make sure we're indicating status via those params.

srambach:
Fixes #5438 

Whenever there is a status on helper text, it now adds the icon in the handlebars.
Removes `helper-text--HasIcon` in uses of helper-text where there is already a status specified, as it is no longer needed
Removes the separate example in Form.md for helper text with an icon, since it's always there
In FormControl.md, the explanatory text at the beginning is updated, but the examples do not include the helper text, but points to the Form examples to show how.

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -186,19 +186,7 @@ cssPrefix: pf-c-form
     {{#> form-group-control}}
       {{#> form-control controlType="input" input="true" form-control--modifier="pf-m-success" form-control--attribute=(concat 'value="This is a valid comment"' 'type="text" id="' form-group--id '" name="' form-group--id '" aria-describedby="' form-group--id '-helper"')}}
       {{/form-control}}
-      {{> form-helper-text helper-text--value='This is helper text for success input.' helper-text-item--IsSuccess=true}}
-    {{/form-group-control}}
-  {{/form-group}}
-  {{#> form-group form-group--id=(concat form--id '-info')}}
-    {{#> form-group-label}}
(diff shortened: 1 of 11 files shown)


thatblindgeye (changes requested):
In addition to the file comments below:

- There are a couple of template files in the demos/Alert directory (horizontal-form and stacked-form) where helper-text-item--HasIcon should be able to be removed (something like IsError is being passed to trigger the icon already)
- For the "Dynamic" and "Dynamic lists" examples for HelperText, the helper-text-item--HasIcon should be able to be removed

thatblindgeye (approved):

srambach:
Closing in favor of a new approach to keep icons in the form controls. https://github.com/patternfly/patternfly/issues/5469

Issue #5466, opened 2023-04-10
Number input - add helper text area for status


srambach:
**Is this a new component or an extension of an existing one?**
Number input

**Describe the feature**
Removing the background status icon from a text input means the number input will no longer display an icon indicating status. This issue is to add an area in the number input for helper text to be placed to show the status.

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
![image](https://user-images.githubusercontent.com/19825616/230974942-6eec1c62-4f45-49fe-85fc-c708eab0ccdd.png)

**Any other information?**
See https://github.com/patternfly/patternfly/issues/5439 for related information.

PR #5467, opened 2023-04-10
feat(form control): remove background icons

### LINKED ISSUE #5439: Remove status icons from form control components
The problem with those icons, since they are background images to browser rendered form elements:
* https://github.com/patternfly/patternfly/issues/2702 - browsers stomp on custom background styles for these elements in ways we can't override
* https://github.com/patternfly/patternfly/issues/4652 - content security policies disallow data URIs as the source for images in markup
* https://github.com/patternfly/patternfly/issues/4900 - due to the way we import sass vars, the default/light theme form control icon is the wrong color currently.

Met with the team and we decided the icon in the form control component should be moved out of that element and into the helper text instead. This update should remove the use of background images from the form control component.
  * The exception here may be the form select arrow, since it is also created from a FA icon as a background image. I think we should remove it and get away from trying to style that directly on a browser rendered element until there is better support for customizing these elements, or we suggest the use of our select component.

Related:
Here's the core issue for updating the helper text component - https://github.com/patternfly/patternfly/issues/5438
And the related react issue - https://github.com/patternfly/patternfly-react/issues/8863

srambach:
This removes icons from the background of form control elements. Status icons will be replaced with required icons in helper text from this pr https://github.com/patternfly/patternfly/pull/5443

Native `select` elements were being restyled with a background image to replace the caret. This replaces the caret with a styled :after element but now requires a wrapper around the select to accomplish this.

The icon, clock, and calendar variants are removed. The time picker should be updated to use a form control group.

There is a related issue to replace the icon on the text input within the number input with helper text https://github.com/patternfly/patternfly/issues/5466

Fixes #5439

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -179,7 +179,9 @@
     border-bottom-width: var(--pf-c-form-control--m-warning--BorderBottomWidth);
   }
 
-  @at-root select#{&} {
+  // Wrapper to style a select input
+  // Replace with &:has(select) when there is enough support
+  &.pf-c-form-control-select {
     --pf-c-form-control--PaddingRight: var(--pf-c-form-control__select--PaddingRight);
     --pf-c-form-control--PaddingLeft: var(--pf-c-form-control__select--PaddingLeft);
 
@@ -199,7 +201,20 @@
       border-right: 5px solid transparent;
       border-left: 5px solid transparent;
       transform: translate(-100%, -50%);
-   }
+    }
+
+    &.pf-m-disabled {
+      --pf-c-form-control--BackgroundColor: var(--pf-c-form-control--disabled--BackgroundColor);
+      --pf-c-form-control--Color: var(--pf-c-form-control--disabled--Color);
+      --pf-c-form-control--BorderBottomColor: var(--pf-c-form-control--disabled--BorderColor);
+    }
+
+    &.pf-m-error {
+      --pf-c-form-control--BorderBottomColor: var(--pf-c-form-control--invalid--BorderBottomColor);
+
+      padding-bottom: var(--pf-c-form-control--invalid--PaddingBottom); // Can't redefine --pf-c-form-control--PaddingBottom b/c the hover padding bottom uses the focus padding var to compute it's padding
+      border-bottom-width: var(--pf-c-form-control--invalid--BorderBottomWidth);
+    }
 
     // Firefox's select text has additional padding
     // stylelint-disable-next-line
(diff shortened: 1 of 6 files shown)


srambach:
> If we're going to wrap the select to support an icon, why wouldn't we also wrap the text input and textarea and leave the status icons there? If the elements can have wrappers, we can just include icons as we normally do. AFAIK the status icons were fine as they were, aside from all of the complications that came with them being background images.

The issue description says that the team decision was to use helper text for status icons. The wrapper was only added as a stop-gap with a recommendation to use the select component instead. There are better ways to accomplish the icon paired with input using the input group.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -85,133 +58,35 @@ $pf-c-form-control--m-clock--Coordinates: "M256 8C119 8 8 119 8 256s111 248 248
...
-  --pf-c-form-control--m-icon-sprite__select--BackgroundSize: var(--pf-c-form-control--FontSize);
-  --pf-c-form-control--m-icon-sprite__select--BackgroundPositionX: calc(100% - var(--pf-global--spacer--md) + 7px);
-  --pf-c-form-control--m-icon-sprite__select--success--BackgroundPosition: calc(100% - var(--pf-global--spacer--md) + 1px - var(--pf-global--spacer--lg));
-  --pf-c-form-control--m-icon-sprite__select--m-warning--BackgroundPosition: calc(100% - var(--pf-global--spacer--md) - var(--pf-global--spacer--lg) + #{pf-size-prem(1px)});
-  --pf-c-form-control--m-icon-sprite__select--invalid--BackgroundPosition: calc(100% - var(--pf-global--spacer--md) - var(--pf-global--spacer--lg));
-
   // This component always needs to be light
   @include pf-t-light("--pf-c-form-control--Color");
 
+  position: relative;
```
thatblindgeye, line 84: Would this be needed here since this is being set on `pf-c-form-control-select` already?

thatblindgeye (commented):
In addition to the file comment below, one nit/alternative (wouldn't block over it) would be instead of having `.pf-c-form-control-select` added to a `.pf-c-form-control` element, having it be a separate wrapper element:

```
<div class="pf-c-form-control-select"> // or pf-c-form-control-wrapper?
    <element class="pf-c-form-control" />
</div>
```

Then using them in examples like:

```
{{#> form-control-select}}
    {{#> form-control }}
        ...content
    {{/form-control}}
{{/form-control-select}}
```

Might depend on whether the `pf-c-form-control` styles should apply to this wrapper element as well (color, height, etc). Otherwise based on conversation in the issue/if using this wrapper to provide a native HTML `select` our caret icon only, this is looking nice!

srambach:
Closed in favor of a new approach detailed in https://github.com/patternfly/patternfly/issues/5469

PR #5472, opened 2023-04-12
fix(deps): use react + org alphas, clean up gulp usage

nicolethoen:


nicolethoen:
let me spend a little more time with this

nicolethoen:
So, there is a known doc-framework issue when it comes to hot reloading new changes.
For now, until it's fixed, the work around is:
- once your localhost is up and running the workspace on port 8001, delete the built 'public' directory. 
Then you can develop as normal and updates should hot reload as they save + compile.

I think we should get this merged ASAP, so @mattnolting can rebase and then we can generate new screenshots of all examples (once his branch is building) and compare them to what's currently on v5. 

This PR updates all screenshots for the workspace, but does not generate them for ALL examples. I'll do that separately now that the docs-framework version is up to date with this PR.

patternfly-docs/generated/components/alert-group/html/toast-alert-group.png
mcoker, line 1: Hmm, it looks like the text/line-height is shifting a bit in this example, but I can't reproduce it in any browser between v4 -> v5 -> this PR build....

mcoker (approved):
LGTM. Left a comment about text shifting and causing a screenshot to be regenerated. Not sure what's causing that?

srambach (approved):
Build/hot reload seems to work ok but fixing the public folder thing sooner than later would be much appreciated!

PR #5477, opened 2023-04-12
Combine Alert and AlertGroup docs

mcarrano:
Combines the documentation for the Alert and Alert Group components onto a single page called Alert.


mcoker (approved):

mcoker:
@mcarrano this should be good now. Can you review? I'm not able to add you as a reviewer since you created the PR.

mcarrano:
Yep. Looks good to me @mcoker . Thanks for pushing those changes. 

I was talking to @nicolethoen a bit about the CSS variable tables and I guess we'll need to find a separate solution for passing an array to pull in variables for multiple components. @nicolethoen do we need an issue for that? Where will that work need to take place?

srambach (approved):
The Alert overview section that was in v4 seems to be gone in v5 and I just wanted to make sure we don't need an overview for each section? Otherwise it looks good assuming the CSS variables issue can be resolved.

mcarrano:
@srambach Are you talking about this text? https://www.patternfly.org/v4/components/alert/html#overview  I don't think there was an intent to remove it, so may have just gotten lost in the copy/paste process. Should I add it back?

nicolethoen:
Here's the issue for tracking the need for CSS vars.
https://github.com/patternfly/patternfly-org/issues/3497

srambach (approved):
👍

PR #5478, opened 2023-04-13
fix(form): add transparent background to form group help label

KKoukiou:
This fixes https://github.com/patternfly/patternfly-react/issues/8951


mcoker (approved):
Nice! Thanks @KKoukiou

Issue #5198, opened 2022-10-28
Bug - Menu/Menu toggle - clean up guidance on aria-label usage


nicolethoen:
Right now, the Menu docs read as below:
![Screen Shot 2022-10-28 at 2 47 51 PM](https://user-images.githubusercontent.com/26305082/198710211-f14bf7c4-af38-4b85-a648-b175c4d57ef7.png)

In fact, whether or not there is a favorite modifier on the action, there should always be an aria-label on a pf-c-menu__item-action-icon because there is no visible descriptive text labelling the icon.

Additionally, right now the MenuToggle docs include this usage of `aria-label`:
![Screen Shot 2022-10-28 at 3 47 49 PM](https://user-images.githubusercontent.com/26305082/198720456-22dc3931-acad-49c4-b41c-f753310d384d.png)

It's not the fact that `pf-m-plain` modifier is on the MenuToggle that requires the addition of an `aria-label`. Instead, it can be updated to note that an aria-label should be placed on the `pf-c-menu-toggle` whenever there is no text visible in the menu toggle. Often, plain modifiers are used when the menu toggle only contains an icon.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/5483

PR #5483, opened 2023-04-14
fix: added clarification to Accessiblity sections of Menu and Menu to…

nzeager:
…ggle documentation

This fix is for [issue #5198](https://github.com/patternfly/patternfly/issues/5198). Let me know if I need to make any adjustments to the wording.


nicolethoen (approved):

thatblindgeye:
It looks like a few days after the original issue was created, a couple PRs went in in our org repo to add accessibility docs for Menu and MenuToggle (https://github.com/patternfly/patternfly-org/pull/3256 and https://github.com/patternfly/patternfly-org/pull/3265 respectively).

Rather than adding that documentation in the HTML page for the components, we should remove the "Accessibility" section so that there's a single source of truth for that documentation in Org. It'd be worth double checking the documentation in Org, though, to make sure the linked issue is resolved. 

@nicolethoen wdyt? There's probably some other components that need their a11y docs removed from this repo after having them added/updated in Org; I can go through and create an issue for that separate from this if we'd want this PR to remove that documentation from Menu and MenuToggle.

nicolethoen:
@thatblindgeye 
I think for now, just in case we don't get the a11y docs written super soon, we should merge this so that the information is documented somewhere. 
In general, most core docs include some information about accessibility related attributes. It may only be necessary to move that info to an accessibility tab if it's sporadic, inconsistent, or incorrect. That may be a future documentation update to consider, but likely not high priority at this time. I'd rather an overabundance of a11y documentation than the alternative.

nicolethoen (approved):

thatblindgeye (approved):

PR #5506, opened 2023-04-21
chore(SearchInput): removed component

### LINKED ISSUE #5212: Remove search input classes/component
In core, the search input component is still beta, and is used in these places. I'm assuming these should be updated to use the text input group, and the search input group component can be removed?

* Menu
* Application launcher
* Log viewer - [we have CSS support for it](https://github.com/patternfly/patternfly/blob/e12e8d5cf82cbb9e0d49a619974d07565dcf68e3/src/patternfly/components/LogViewer/log-viewer.scss#L136-L138)
* Select
* Dual list selector
* Page and masthead demos - masthead drilldown
* Context selector
* Tree view
* Toolbar - [toolbar-item-context-selector.hbs](https://github.com/patternfly/patternfly/blob/e12e8d5cf82cbb9e0d49a619974d07565dcf68e3/src/patternfly/components/Toolbar/toolbar-item-context-selector.hbs#L11)
* Text input group - [text-input-group-clear.hbs](https://github.com/patternfly/patternfly/blob/e12e8d5cf82cbb9e0d49a619974d07565dcf68e3/src/patternfly/components/TextInputGroup/text-input-group-clear.hbs) - doesn't appear to be used anywhere

thatblindgeye:
Closes #5212


mcoker (changes requested):
LGTM! Couple of nits:

* Looks like we need to remove https://github.com/patternfly/patternfly/blob/246534b256b958d487f003b6da86c8e526e49363/src/patternfly/sass-utilities/component-namespaces.scss#L206-L207
* I don't think any of these `#ws-core-c-search-input[...]` id's exist anywhere in the text input group docs? https://github.com/patternfly/patternfly/blob/v5/src/patternfly/components/TextInputGroup/examples/TextInputGroup.css

srambach (approved):
lgtm. Created an issue for a masthead demo where the search input is displaying incorreclty outside the masthead, but that exists in the demo already. (https://github.com/patternfly/patternfly/issues/5557)

mcoker (approved):
💪

PR #5507, opened 2023-04-21
feat(Nav): added top and bottom padding to nav list

### LINKED ISSUE #5406: Nav - add sidebar padding to nav ul
Nav

**Describe the feature**
[A clear and concise description of the new feature. What is the expected behavior?](https://github.com/patternfly/patternfly/pull/5093#pullrequestreview-1320868058)

@mcoker
> Just glancing at the nav component, it seems like we could remove that nav-specific spacing from nav body, then:

> Add the padding-top and padding-bottom to .pf-c-nav__list
Remove it from .pf-c-nav__subnav (expandable nav)
Sort out this margin-top on .pf-c-nav__item. It only applies to grouped items in .pf-c-nav__section and is creating space between nav items, and the gap above the items, which should now be handled by the padding on .pf-c-nav__list. That could be a flex column with a gap or easiest might be to update that rule to be &:not(:first-child) {}?

**Any other information?**

thatblindgeye:
Closes #5406 

Looks like styles for `.pf-c-nav__item` are being applied twice, which is pre-existing behavior. From the Core v5 surge:

Core v5 nav item styling from devtools

From this PR's surge:

PR nav item styling from devtools

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -993,6 +993,8 @@
 
 // Subnav
 .#{$nav}__subnav {
+  --#{$nav}__list--PaddingTop: 0;
+  --#{$nav}__list--PaddingBottom: 0;
   --#{$nav}__link--PaddingTop: var(--#{$nav}__subnav__link--PaddingTop);
   --#{$nav}__link--PaddingRight: var(--#{$nav}__subnav__link--PaddingRight);
   --#{$nav}__link--PaddingBottom: var(--#{$nav}__subnav__link--PaddingBottom);
@@ -1021,11 +1023,6 @@
     --#{$nav}__link--PaddingLeft: var(--#{$nav}__subnav__subnav__link--PaddingLeft);
     --#{$nav}__subnav--PaddingLeft: var(--#{$nav}__subnav__subnav--PaddingLeft);
   }
-
-  > .#{$nav}__list {
-    --#{$nav}__list--PaddingTop: 0;
-  }
-
 
   // remove at breaking change - not used
   &.pf-m-flyout {
@@ -1100,7 +1097,7 @@
     --#{$nav}__section--MarginTop: 0;
   }
 
-  > .#{$nav}__list {
+  &:not(:last-child) {
     --#{$nav}__list--PaddingBottom: 0;
   }
 }
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -30,7 +30,7 @@
 
   // List
   --#{$nav}__list--PaddingTop: var(--#{$pf-global}--spacer--sm);
-  --#{$nav}__list--PaddingBottom: var(--#{$pf-global}--spacer--md);
+  --#{$nav}__list--PaddingBottom: var(--#{$pf-global}--spacer--sm);
 
   // Item
   --#{$nav}__item--MarginTop: 0;
@@ -178,7 +178,7 @@
   --#{$nav}--m-horizontal-subnav__link--m-current--after--BorderLeftWidth: var(--#{$pf-global}--BorderWidth--sm);
 
   // Subnav
-  --#{$nav}__subnav--PaddingBottom: var(--#{$pf-global}--spacer--md);
+  --#{$nav}__subnav--PaddingBottom: var(--#{$pf-global}--spacer--sm);
   --#{$nav}__subnav--xl--PaddingLeft: var(--#{$nav}__link--PaddingLeft);
   --#{$nav}__subnav__link--MarginTop: 0;
   --#{$nav}__subnav__link--PaddingTop: var(--#{$pf-global}--spacer--sm);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Nav/nav.scss
```diff
@@ -1000,7 +1012,6 @@
   --#{$nav}--c-divider--PaddingLeft: var(--#{$nav}__subnav--c-divider--PaddingLeft);
 
   max-height: var(--#{$nav}__subnav--MaxHeight);
-  padding-bottom: var(--#{$nav}__subnav--PaddingBottom);
```
mcoker, line 1003: Looks like you also need to remove this var - https://github.com/patternfly/patternfly/blob/171eb5cf7841f9612dce372cf045a72f41676433/src/patternfly/components/Nav/nav.scss#L181

mcoker (changes requested):
@thatblindgeye this looks great! Just one comment about a var that needs to be removed. Though there are some visual differences between this PR and v4, but I'm not sure if we just want to match v4, or review those differences. Questions for @mceledonia or whoever you think is best to answer:

* This PR adds space around the nested expanded list - do we want to keep that or match PF4?
  * Here is PF4 - no space between the top level expandable item and expanded list
  Screenshot 2023-05-09 at 5 21 52 PM

  * Here is this PR - note the space around the expanded list creating space between the expandable item and the list
  Screenshot 2023-05-09 at 5 22 12 PM

  * Here is what it looks like if I remove the top spacer on the list (to match PF4 - no space between the expandable item and the expanded list), and change the bottom padding on the list to small, to match the normal list top-padding I suggested in another comment.

----

* This PR introduces extra space between sections - wdyt?
  * PF4
  Screenshot 2023-05-09 at 5 24 35 PM

  * This PR
  Screenshot 2023-05-09 at 5 24 47 PM

src/patternfly/components/Nav/nav.scss
andrew-ronaldson: I agree with @mcoker's observations and with these changes we should be good to go.

thatblindgeye:
@mcoker I believe I captured your comments, let me know if I missed anything! I had to add in another var + styling for the subnav to maintain the spacing between items in the Expandable demo.

src/patternfly/components/Nav/nav.scss
```diff
@@ -1010,6 +1022,10 @@
     --#{$nav}__subnav--PaddingLeft: var(--#{$nav}__subnav__subnav--PaddingLeft);
   }
 
+  > .#{$nav}__list {
+    --#{$nav}__list--PaddingTop: 0;
```
mcoker, line 1026: This works, but I believe you can just set this under https://github.com/patternfly/patternfly/blob/4659b8c6202d1e5fe07cfe3093557ef30997f45e/src/patternfly/components/Nav/nav.scss#L995

src/patternfly/components/Nav/nav.scss
```diff
@@ -1083,6 +1099,10 @@
   &.pf-m-no-title {
     --#{$nav}__section--MarginTop: 0;
   }
+
+  > .#{$nav}__list {
+    --#{$nav}__list--PaddingBottom: 0;
```
mcoker, line 1104: This is gonna remove the bottom spacing from all lists in a section, removing the spacer from the bottom of the sidebar like this:
Screenshot 2023-05-16 at 7 28 46 PM

You could probably do this a couple of ways.

* something like `__section:not(:last-child) { __list--PaddingBottom: 0; }` and target the bottom padding on lists other than the last section.
* just reduce the section + section margin-top to include whatever the nav list padding is
* you might try using a `margin-bottom` on the nav list instead (and heck - `margin-top`, too!)... and [margin-collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) will remove the lists's bottom margin when it touches the section + section margin. Not sure if that would cause any other issues but might be worth trying.

mcoker (changes requested):
one nit, an issue with the bottom padding on lists in groups, and I just noticed the third-level expandable (nested subnavs) has an unnecessary bottom padding.

Screenshot 2023-05-16 at 7 35 20 PM


I _think_ we can just remove the top and bottom padding from a list in `__subnav`? If we want to keep the extra space below the list at the end of a `__subnav`, we can just adjust the `__subnav` bottom padding.

existing spacing in PR
Screenshot 2023-05-16 at 7 45 33 PM

removing the bottom padding, too, from a subnav list
Screenshot 2023-05-16 at 7 45 22 PM

Existing spacing
Screenshot 2023-05-16 at 7 42 01 PM

with the bottom padding removed
Screenshot 2023-05-16 at 7 41 27 PM

wdyt @mceledonia @andrew-ronaldson

mcoker (approved):
LGTM! I found another spacer that's off from the way we approached the group spacing. I wonder if one of the other options might work better. I filed an issue so we can work on that separately - https://github.com/patternfly/patternfly/issues/5592. Feel free to grab it if you'd like @thatblindgeye!

PR #5510, opened 2023-04-24
build(dev-server): enable hot reloads of helpers

### LINKED ISSUE #5509: Enable hot reloads of helpers
**Describe the feature**
Currently the dev server isn't able to perform hot reloads when handlebars helpers are changed, which necessitates restarting server many times as you're trying to make or improve a helper. This makes working on helpers extremely tedious.

Ideally we would be able to have the dev server hot reload when helpers are changed.

wise-king-sullyman:
Closes #5509

Also refactors helpers to be exported from their own file, with a utility that handles helper registration in their previous location. This has the additional benefit of helping to separate concerns a bit and tries make the approach to helpers a little cleaner.

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -1,10 +1,11 @@
 import fs from'fs';
+import path from 'path';
 import gulp from'gulp';
 import rimraf from'rimraf';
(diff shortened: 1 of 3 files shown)


mcoker (approved):
Looks great! 🚀 I poked around and everything I tested seems to work like I'd expect, the server start time is dramatically reduced, and the hbs helpers file hot reloading is 🔥

PR #5516, opened 2023-04-27
feat(flex): added gap support

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5288

I believe the feature is in a good spot to review. I haven't added official examples/documentation/etc yet because I'd like to test the feature first and see if there need to be changes.  Cliff's notes:
* The "spacer" system (margin used to space items on the main axis) is maintained as the default for the flex layout without any sort of modifier classes.
* A gap (either `row-gap` or `column-gap`) is added by default on the cross axis when using the spacer system on the main axis
  * This means a `row-gap` has been added as the default (no other modifiers), so along with the spacer system on the main axis, there is now space between wrapping rows
* Using gap on the same axis as a defined spacer system will override the spacer system
* There is a default value for the spacer/gap used for columns, and one for the spacer/gap used for rows. Whether you're using space-items/spacers, gap/row-gap/column-gap, or a flex row or column, without any sort of size modifier for your spacer/gap/[row/column]-gap, you will always get a default spacer between adjacent items on the X-axis and items on the Y-axis.
* `pf-m-row-gap` adds the default row-gap for the flex layout. You can add `-[size]` and `-on[breakpoint]` for more flexy goodness.
* `pf-m-column-gap` ^ same story
* `pf-m-gap` adds the default column gap and row gap, which are different. You're opting into the flex layout's default gaps. From there if you specify `-[size]`, that size is used for both column and row gap.
* I changed the default column spacer from `md` to `lg` - every other layout uses the `lg` spacer, and that is the new page chrome gutter, too, so that seemed reasonable.
* Our sass loops generate a bunch of redundant code, so I tried to group selector rules for things that can be grouped in these loops. This can apply to the other loops, too, if we'd like to try that and cut down on compiled code. Basically creating a block like this for each breakpoint 
    ```scss
    @media (min-width: 1200px)
      .pf-l-flex:is(.pf-m-column-gap-on-xl, .pf-m-column-gap-none-on-xl, .pf-m-column-gap-sm-on-xl, .pf-m-column-gap-md-on-xl, .pf-m-column-gap-lg-on-xl, .pf-m-column-gap-xl-on-xl, .pf-m-column-gap-2xl-on-xl, .pf-m-column-gap-3xl-on-xl, .pf-m-column-gap-4xl-on-xl) > * {
          --pf-l-flex--spacer--column: 0;
      }
    ```
    * instead of
    ```scss
    @media (min-width: 1200px)
      .pf-l-flex.pf-m-column-gap-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-none-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-sm-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-md-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-lg-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-xl-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-2xl-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-3xl-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
      .pf-l-flex.pf-m-column-gap-4xl-on-xl {
        --pf-l-flex--spacer--column: 0;
      }
    ```


srambach:
I think this looks good - couple of questions.
- Before, we'd talked about overriding gap with individual spacers, but ended up deciding that wasn't reasonable, right?
- I'd been assuming `pf-m-column` would swap row and column gap but this doesn't attempt to do that, correct? I agree because that simplifies things considerably.

mcoker:
> Before, we'd talked about overriding gap with individual spacers, but ended up deciding that wasn't reasonable, right?

Yeah, though if it's something we still want to consider, I can fork this work and try and support that. Just some additional thoughts - I think the issue with children extending outside of their container is going to be a difficult caveat to document and for users to find. It will likely pop up as a bug, since the docs would have to be something like you can only use item spacers with gap if 1) the flex layout doesn't wrap, or 2) the item spacers are greater than the gap size. With `margin-trim` is on the horizon, I can't think of a downside to using the spacing system on the main axis for item spacers if we also convert the item margins to logical properties - meaning, down the road at least, I don't see the benefit of using gap with spacers. Seems like using one or the other for now is reasonable. 

For science, I whipped up a demo of the flex layout with gap and spacers (creating negative margins on the items) and tested with `margin-trim` in safari (supported in 16.4), and looks like it only _trims_ excess margin, not negative margins. I wonder if there might be a property for that in the future, or future support in the [margin-trim spec](https://w3c.github.io/csswg-drafts/css-box-4/#margin-trim)?

> I'd been assuming pf-m-column would swap row and column gap but this doesn't attempt to do that, correct? I agree because that simplifies things considerably.

Yep. Since `column/row-gap` maintain their role regardless the direction, seems like our spacer system should follow suit. I imagine that's also how users expect it to work, too.

mcoker:
@mceledonia @lboehling I updated the default spacer for the flex layout to `--pf-global--spacer--lg` - which is the universal page gutter spacer we're moving toward, as well as the default spacer used in our other layouts (grid, stack, split, gallery).
* Is that OK?
* By default, we're also adding a spacer between rows, too, when rows wrap. Should that be the same as the column spacer, or a different value? Here is a demo with the [row and column gaps the same](https://codepen.io/mcoker/pen/bGmRaMe?editors=1100) and here is a [small row gap and large column gap](https://codepen.io/mcoker/pen/gOBRvpM?editors=1100).

srambach (approved):
👍

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -7,39 +7,34 @@ cssPrefix: pf-v5-l-flex
...
-- `.pf-m-column` direct descendants (not last child): `margin-bottom: 16px`.
-- `.pf-m-column` nested `.pf-v5-l-flex` containers (not last child): `margin-bottom: 16px`.
-
-## Features
-
-- `.pf-v5-l-flex` is infinitely nestable and can be used to group items within.
-- `.pf-m-spacer-{xs,sm,md,lg,xl,2xl,3xl}` can be applied to parent or direct children and changes the spacer value for only the element to which it is applied. Responsive spacers can be used by appending `{-on-[breakpoint]}` to `.pf-m-spacer-{size}`. Example: `.pf-m-spacer-lg-on-xl`.
-- `.pf-m-space-items-{xs,sm,md,lg,xl,2xl,3xl}` can be applied to `.pf-v5-l-flex` only and changes the spacing of direct children only. Responsive spacers can be used by appending `{-on-[breakpoint]}` to `.pf-m-space-items-{size}`. Example: `.pf-m-space-items-lg-on-xl`.
+### Spacing
+The flex layout provides to ways of spacing its direct children.
```
thatblindgeye, line 13:
```suggestion
The flex layout provides two ways of spacing its direct children.
```

src/patternfly/layouts/Flex/examples/Flex.md
```diff
@@ -7,39 +7,34 @@ cssPrefix: pf-v5-l-flex
...
+- [Spacing system](#spacing-system)
+  - The spacing system applies a margin between the flex layout's direct children to create space between items along the main axis only. The benefit of the spacing system is that it allows for variable spacing between direct children. Use the spacing system when individual flex items along the main axis require a different spacer than sibling items.
+- [Gap spacing](#gap-spacing)
+  - Gap spacing uses flex `gap` to space the flex layout's direct children, and can be applied to space rows (`row-gap`), columns (`column-gap`), or both (`gap`). The benefit of gap spacing is that item wrapping is improved and improved item spacing that works better with CSS flex's logical layout properties. E.g., spacing in RTL layouts that rely on logical properties is improved. Use the gap system when all direct children should use the same spacer for rows, columns, or both. **Note** using `gap` along the main axis will override any other spacing applied using the spacing system.
```
thatblindgeye, line 17: Nit: feel like the "Note" gets kind of lost in the shuffle since this bullet is so lengthy:

```suggestion
  - Gap spacing uses flex `gap` to space the flex layout's direct children, and can be applied to space rows (`row-gap`), columns (`column-gap`), or both (`gap`). The benefit of gap spacing is that item wrapping is improved and improved item spacing that works better with CSS flex's logical layout properties. E.g., spacing in RTL layouts that rely on logical properties is improved. Use the gap system when all direct children should use the same spacer for rows, columns, or both. 
  - **Note** using `gap` along the main axis will override any other spacing applied using the spacing system.
```
Though both this and the Spacing sub-bullet could be broken up into several sub-bullets as well.

thatblindgeye (commented):
Quick typo fix and non-blocking nit below, otherwise looks good

PR #5517, opened 2023-04-27
chore(page): default page to full viewport height

### LINKED ISSUE #5448: Page - make pf-m-full-height the default
We added a modifier `.pf-m-full-height` to eliminate the need to add `height: 100%` to all ancestor elements. At the breaking change, we want to make this the default.

Question: do we need a modifier to opt out? (4/18/23 let's not for now and can add it later if needed)

**Any other information?**
Documentation like this will also need to be updated https://github.com/patternfly/patternfly-react/blob/ab1d27353a3aed7cc3c2b83c4278edc66a63c2de/packages/react-core/src/demos/Page.md?plain=1#L28

srambach:
Fixes #5448 
Removes `.pf-m-full-height` for a page in lieu of defaulting the page to be 100vh/dvh with a max-height so it doesn't overrun the space it's allocated if it's within another layout.

This will need a follow-on issue on .org to limit the width of all the page examples (this is currently done just for a couple of the examples).

### DIFF
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -254,7 +254,7 @@ import './Page.css'
 
 ## Documentation
 ### Overview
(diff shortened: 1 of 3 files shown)


mcoker (approved):
👍

Issue #5265, opened 2021-03-29
Make icon sizes consistent across the design system


mmenestr:
It's been brought to my attention that the icons size guidance in PF does not match the icon sizes in our sketch kit. 
Sketch kit: 
![image](https://user-images.githubusercontent.com/51165119/112858922-58811400-9080-11eb-9247-26aef6877eda.png)

PF icons: 
![image](https://user-images.githubusercontent.com/51165119/112858939-5e76f500-9080-11eb-866f-aa949ae1a4a6.png)

These two issues give a bit more background:
1.  https://github.com/patternfly/patternfly-design-kit/issues/162
2. https://github.com/patternfly/patternfly-design/issues/787

It seems that the sketch kit sizing is the correct one (12px, 16px, 24px), @mcoker do you have any suggestions as to what we should do here? Would there be a potential problem with updating the icon sizing in the PF guidelines?

cc: @mcarrano @doruskova

PR #5525, opened 2023-05-02
fix(icons): update global icon font sizes

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5265

----

@mceledonia @andrew-ronaldson @mmenestr @lboehling 

A couple of questions
* By default, SVGs and font-based icons (font awesome, pficons in core) will inherit their size based on whatever the text size is where the icon is used. So their default size is `1em` basically. **What is the use case for these icon specific sizes?** As in, when would you want to use an icon-specific size versus making the icon match whatever the text size is? I can see if you have a standalone icon in a thing like a big icon in a card, or like the empty state icon. But if you have an icon beside a title (or any other kind of text)... would you ever want that icon to be a different size than the title? Should it be an option for the icon to match whatever the text size is?
  * One thing I'm thinking about is how our current global font-size vars for text are `--xs: 12px`, `--sm: 14px`, and  `--md: 16px`. But our icon font-sizes are `--sm: 12px` and `--md: 16px`. So if you wanted to make an icon match the 14px global font-size, you can't do that using icon size global vars. Should you be able to do that? 
* Our icon component currently sizes the icon inside of it using the default font-size (16px) when there is no size modifier passed. You can then override the icon size using one of the values provided by the icon component - those values are based off of the global vars for icon sizes. Personally I think the icon component should deliver an icon as `1em` by default, if no size modifier is passed, so it's as big/small as the font-size is where the icon is used. I imagine that's the most common use case? WDYT?

### DIFF
--- a/src/patternfly/components/Icon/icon.scss
+++ b/src/patternfly/components/Icon/icon.scss
@@ -1,40 +1,39 @@
 // @debug $icon; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
 .#{$icon} {
-  // Size
-  --#{$icon}--Width: var(--#{$pf-global}--icon--FontSize--md);
-  --#{$icon}--Height: var(--#{$pf-global}--icon--FontSize--md);
-
-  // When inline, use the parent's size
+  // Sizes
+  --#{$icon}--Width: 1em;
+  --#{$icon}--Height: 1em;
+  --#{$icon}--m-sm--Width: var(--#{$pf-global}--icon--FontSize--sm);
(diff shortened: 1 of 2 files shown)


mceledonia:
Hey Michael, we discussed this on the design leads call and agreed that your proposal 

> I think the icon component should deliver an icon as 1em by default, if no size modifier is passed, so it's as big/small as the font-size is where the icon is used.

...sounds great!

andrew-ronaldson:
I agree with @mceledonia on this too. Proposal sounds good to me.

srambach (approved):
LGTM 👍

mceledonia (approved):

Issue #5395, opened 2023-02-20
Update static icon color inside component that use an embedded icon


mcarrano:
This follows up on https://github.com/patternfly/patternfly-design/issues/1195

Want to update the text colors of icons used in the following components from gray to black so the gray icon color is reserved for interactive elements, only.

Search input

![image](https://user-images.githubusercontent.com/6619860/220192399-7ee35e8d-2d70-4982-a4d3-33136e79f49d.png)

Change the gray search icon to black.

Breadcrumb

![image](https://user-images.githubusercontent.com/6619860/220192458-11db8621-db0c-478e-bade-6dacba6d93e3.png)

Change the caret separating the nodes to black.

Timepicker

![image](https://user-images.githubusercontent.com/6619860/220192626-fa62121e-bc58-48a5-a751-008142ce972e.png)

Change the clock icon to black.

Issue #5469, opened 2023-04-11
Form controls - remove use of background images


srambach:
**Is this a new component or an extension of an existing one?**
This applies to form controls that are using the native elements for text input, select, and text area.

**Describe the feature**
For [various reasons](https://github.com/patternfly/patternfly/issues/5439) the method of using a background image on the control to display an icon needs to be changed. Per leads meeting of 4/11/23, a decision was made to add a wrapper element around the native elements, and to add another child element able to hold multiple icons.  The wrapper element will receive the PatternFly styling and the HTML elements themselves will have appearance:none to remove browser-specific styling.

This will require follow on React issues for the text input, text area, and form select components.
The Text Input Group and Select components will remain as is.

nicolethoen:
related to https://github.com/patternfly/patternfly/issues/2702

PR #5526, opened 2023-05-02
fix(icons): update static icon colors

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5395

**note:** time picker (in react) is built using the form control component, but @srambach should fix that via https://github.com/patternfly/patternfly/issues/5469 so I left that component alone here to avoid any sort of conflicts.

----

@mcarrano @mceledonia I noticed a handful of other places - do we want to update any of these? 
* Empty state icons - https://www.patternfly.org/v4/components/empty-state/html
* Multiple file upload icon - https://www.patternfly.org/v4/components/file-upload---multiple/html
* List component icons - https://www.patternfly.org/v4/components/list#with-icons


thatblindgeye (approved):
Will take another look if the other components mentioned are updated, but looks good and the icon colors update as expected when swapping between dark and light mode.

mmenestr:
@mcoker can't believe I missed those, but I guess yes, those should be updated as well! The only one I am iffy about is the empty state because it might look really harsh, but at the same time there's not good reason why it shouldn't follow the guideline. Would you be able to show me an example before changing everything? :D 

Also, I see the changes in breadcrumb, but don't see it in the search input?

mcoker:
@mmenestr yep, screenshots below! Re: the search input, sorry I should have mentioned that is the [text input group component](https://patternfly-pr-5526.surge.sh/components/text-input-group#utilities-and-icon-with-placeholder-text) in core. The search input component in react is built from the text input group component. The core "search input" component page is being removed.

Screenshot 2023-05-03 at 2 01 43 PM
Screenshot 2023-05-03 at 2 01 39 PM
Screenshot 2023-05-03 at 2 01 14 PM
Screenshot 2023-05-03 at 2 01 08 PM
Screenshot 2023-05-03 at 2 01 04 PM
Screenshot 2023-05-03 at 2 01 00 PM

mcoker:
@mmenestr for the multiple file upload, I updated the main "file upload" icon color as well as the individual file icons in the progress list - but I left the spinner grey. Is that what you'd expect?

Screenshot 2023-05-03 at 9 41 10 PM

mmenestr:
That looks good @mcoker ! Does the spinner in the upload move? Shouldn't it be blue actually? or is it just a static icon? and if it's static then in theory that should also be black yes

srambach:
Just asking - tiles also have icons but seem to follow a different model so just checking you didn't want those to match.

mmenestr:
@srambach tiles use black icons in their default state so I think they pretty much do follow this model, other than the fact that they become blue when the tile is selected, but I think that's fine!

mcoker:
> Does the spinner in the upload move? Shouldn't it be blue actually? or is it just a static icon? and if it's static then in theory that should also be black yes

@mmenestr it's static - updated! https://patternfly-pr-5526.surge.sh/components/file-upload/multiple-file-upload#file-upload-status

srambach (approved):
👍

Issue #5529, opened 2023-05-04
About modal box, background image: remove default image, update docs


mcoker:
Re: https://github.com/patternfly/patternfly/issues/4930

The use of relative paths to images using `$pf-global--image-path` is problematic. With the existing v5 updates to [about modal](https://github.com/patternfly/patternfly/pull/5328/), [background image](https://github.com/patternfly/patternfly/pull/5328/), and [form control](https://github.com/patternfly/patternfly/issues/5469), these relative image paths no longer need to be defined and the about modal and background image can be updated to not provide an image in the CSS by default.

Issue #4930, opened 2022-06-28
Bug - core - Ambiguous url('../../assets...)


rp-:
**Describe the problem**
ParcelCSS errors, because patternfly doesn't use absolute url() in css.
```
@parcel/transformer-css: Ambiguous url('../../assets/images/status-icon-sprite.svg#success') in custom property. Relative paths are resolved from the location the var() is used, not where the
 custom property is defined. Use an absolute URL instead

  /home/rp/prj/node_modules/@patternfly/react-styles/css/components/FormControl/form-control.css:122:63
    121 |   --pf-c-form-control--textarea--invalid--BackgroundPositionY: var(--pf-c-form-control--PaddingLeft);
  > 122 |   --pf-c-form-control--m-icon-sprite--success--BackgroundUrl: url("../../assets/images/status-icon-sprite.svg#success");
  >     |                                                               ^
    123 |   --pf-c-form-control--m-icon-sprite--m-warning--BackgroundUrl: url("../../assets/images/status-icon-sprite.svg#warning");
    124 |   --pf-c-form-control--m-icon-sprite--invalid--BackgroundUrl: url("../../assets/images/status-icon-sprite.svg#invalid");

  💡 Replace with: url(/node_modules/@patternfly/react-styles/css/assets/images/status-icon-sprite.svg#success)
```

**How do you reproduce the problem?**
Try to use patternfly with Parcel >= 2.4.0

**Expected behavior**
Parcel correctly building

**Is this issue blocking you?**
Work around is to use parcel < 2.4.0 or correctly activate PostCSS as alternative

**Screenshots**
-

**What is your environment?**
 - OS: archlinux
 - Browser chrome
 - Version 102

**What is your product and what release date are you targeting?**
patternfly 2022.7

jamestalton:
https://parceljs.org/languages/css/#url()
> Note: Only [absolute paths](https://parceljs.org/features/dependency-resolution/#absolute-specifiers) may be used within CSS custom properties, not relative paths. This is because url() references in custom properties are resolved from the location where the var() is used, not where the custom property is defined. This means that the custom property could resolve to different URLs depending on which file it is used in. To resolve this ambiguity, use absolute paths when referencing URLs in custom properties.

jamestalton:
Parcel makes it easy to create UI libraries that support both ES6 modules and legacy CommonJS.
It would be wonderful to use Parcel with PatternFly but it is blocked by this issue.

mcoker:
[...]

PR #5530, opened 2023-05-04
fix(about-modal, background-image): updated to not ship default background image

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5529, fixes https://github.com/patternfly/patternfly/issues/4930

* No longer shipping the PF logo as the default for about modal and background image in the CSS, with approval from @mceledonia 
* Updated docs for ^ to note the var needed to set the background images
* Removed `$pf-global--image-path` as it's no longer necessary
* Updated form control sprite paths to no longer use the global var path so the SCSS will compile - those paths will be removed in https://github.com/patternfly/patternfly/issues/5469


mcarrano (approved):
Looks good to me!

thatblindgeye (approved):

mcoker:
@thatblindgeye @srambach rebased with the v5 versioning work.

srambach (approved):
🔥

PR #5533, opened 2023-05-04
fix(wizard): adjust z-indexes so menus show with overflow

### LINKED ISSUE #4794: Wizard - cannot append select menus to parent in a wizard
setting the Select `menuAppendTo` prop to be the parent - both cause issues with overflow and stacking context. This does work in a plain modal.

[This codesandbox demonstrates the side effect of appending to parent.](https://codesandbox.io/s/bold-worker-6ggke3?file=/index.js)

[This additional sandbox demonstrates an attempted work around of appending the menu to the wizard main body - not ideal](https://codesandbox.io/s/objective-hugle-84nzc6?file=/index.js)

note: both of the code sandboxes above use the menu + menuToggle rather than a select. Cost management needs the select to work for now

The footer's shadow currently needs to appear on top of the content of the wizard. If it can be drawn differently, then the menu does not need to be under the footer.

srambach:
Fixes #4794 

Adjusts z-indexes of wizard components as demonstrated in https://codesandbox.io/s/elated-jerry-1xl9m5?file=/fonts.css

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -166,7 +166,7 @@
   --#{$wizard}__outer-wrap--MinHeight: #{pf-size-prem(250px)};
 
   // Main
---#{$wizard}__main--ZIndex: auto;
+  --#{$wizard}__main--ZIndex: auto;
 
   // Body
   --#{$wizard}__main-body--PaddingTop: var(--#{$pf-global}--spacer--md);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -166,7 +166,7 @@
   --#{$wizard}__outer-wrap--MinHeight: #{pf-size-prem(250px)};
 
   // Main
-  --#{$wizard}__main--ZIndex: var(--#{$pf-global}--ZIndex--xs);
+--#{$wizard}__main--ZIndex: auto;
```
mcoker, line 169: nit - indent

```suggestion
  --#{$wizard}__main--ZIndex: auto;
```

mcoker (approved):
LGTM. It's worth noting that this doesn't currently work with menus that have a positioned parent within the wizard main area, like dropdown or select (the parent component class selector is positioned). It does work with the menu component in a standard wizard, since the menu's position is relative to the outer wrap element.

Issue #5523, opened 2023-05-02
Audit dark theme & address issues across all v5 components


nicolethoen:
@andrew-ronaldson has begun a doc to track issues discovered during his audit:
https://docs.google.com/document/d/1cvhnDIN2e93dwZvXDxrHNtFvSVPz-Hu-YgzI3XMO5kM/edit

Meeting planned for 5/5 to complete audit and work through fixes.

PR #5518, opened 2023-04-27
chore(input-group): update documentation

srambach:
Fixes #5437 
Corrects the documentation for the updated structure of input-group to include input-group__text

PR #5535, opened 2023-05-05
fix(theme): fix dark theme bugs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5523

@srambach @thatblindgeye also ~~updated docs for the input group~~ (oops, just saw https://github.com/patternfly/patternfly/pull/5518), and the use of input group in the slider.

@mceledonia @andrew-ronaldson the updates we discussed today should be in place. I also added the floating background color to the "raised" panel variation - is that OK? I believe the intention of that variation is to be used for stuff you may want to attach to a menu toggle. In the future, I believe our plan is to always have the raised panel variation wrap the menu when toggling a menu with the menu toggle - then we'll remove the header/filter/search and the footer from the menu component, and build that into the panel's header/footer instead. Here is a use of the panel in this PR - it holds the form below this text input group:  https://patternfly-pr-5535.surge.sh/components/text-input-group/#search-input-group-advanced-search-expanded


mcarrano (changes requested):
I thought that a good way to check this out would be to look at some of our full screen demos. But now I'm seeing that when demos open in a new window, they no longer inherit dark theme even it that's set in the workspace. Is this expected? I expect that if dark theme is set at the workspace level, demos as well as embedded examples will inherit that theme.

mcoker:
@mcarrano yep, that's how the dark theme toggle has always worked. It only sets it on the current page. If you refresh the page or open an example in a new window, you have to set it again. I believe @kmcfaul is working a way for the site/toggle to remember your preference.

mcarrano (commented):
Regarding the full screen demos, I realize that is unrelated to this PR. However I feel like it's something we should try to fix with the website/workspace as it will be surprising if we cannot preview these demos in light or dark.

For this PR, I will defer to @mceledonia and @andrew-ronaldson for design review, although in taking a quick look, seems like some things are still unresolved. I'm especially seeing this for Brand and components that use the brand like cards. Is there a fix there?

Screenshot 2023-05-09 at 2 07 51 PM

srambach (approved):
LGTM

mcoker:
Rebased with the latest from v5. Note that the dark theme toggle on full page examples isn't working, I put up a fix for that here and will rebase this PR once that fix is published to the docs framework package - https://github.com/patternfly/patternfly-org/pull/3532

For now you can manually enable dark theme on those pages by by adding `class="pf-v5-theme-dark"` to the `<html>` element

nicolethoen:
@mcoker should be able to rebase and get the fixed theme switcher

andrew-ronaldson (approved):
Went through the demo and the changes look good. Thanks @mcoker

mcarrano (commented):
@mcoker This looks good, but I just have a few questions about things that I'm seeing.

* Is it possible to swap the brand image in examples where we are looking at the black on white PatternFly brand placeholder? I'm guessing that occurs because we have black text on a transparent background. See Brand and Card examples specifically.

* I'm noticing that outlined label borders have a weird pixelation in dark theme.
Screenshot 2023-05-15 at 5 05 42 PM
Gray borders seem OK. Anything that can be done about that?

* I see that the background of the User menu changes when I switch to dark theme. Is that what we expect or want?

Screenshot 2023-05-15 at 5 07 39 PM

@andrew-ronaldson @mceledonia interested in your thoughts

srambach:
There's a contrast failure on the [read-only code editor](https://patternfly-pr-5535.surge.sh/components/code-editor/html/read-only/). 
image

srambach:
File upload drag over states fail contrast - should the background be a darker gray? https://patternfly-pr-5535.surge.sh/components/file-upload/multiple-file-upload

srambach:
File upload drag over states fail contrast - should the background be a darker gray? https://patternfly-pr-5535.surge.sh/components/file-upload/multiple-file-upload

mattnolting (approved):
LGTM

mcoker:
@srambach @mcarrano created a follow-up issue for the remaining bugs https://github.com/patternfly/patternfly/issues/5586

PR #5556, opened 2023-05-10
feat(deps): introduce stylelint-use-logical-spec stylelint plugin

### LINKED ISSUE #5555: [all] - [RTL _ Use logical properties and values for controlling the layout]
Physical dimensions and directions are described left to right and top to bottom, while their logical counterparts are described start to end and inline or block.

Using the latter has the advantage that UIs will be corrently auto-adjusted for RTL languages. Patternfly so far uses the physical properties.

https://www.w3.org/TR/css-logical-1/

This will be fix of a big percentage of problems that Patternfly has regarding RTL support. 

You can use the stylelint plugin [0] to autoconvert a big part of the fixes. 

I suggest that you first: 

* [ ] stop introducing properties the 'old' way
* [ ] convert the existing codebase little by little or how you feel confortable to use the logical properties

Note that this should haveno impact on LTR users, while massively improving the experience of RTL.

[0] https://www.npmjs.com/package/stylelint-use-logical-spec

KKoukiou:
Add this new linter check when linting new code.

This enforces the usage of Logical Properties and Values in CSS.

More info @ https://www.npmjs.com/package/stylelint-use-logical-spec

Partially Fixes #5555


mcoker:
Hey @KKoukiou thanks for the PR. I ended up adding this plugin to per your PR to this changeset of a first pass at some RTL changes - https://github.com/patternfly/patternfly/pull/5778

https://github.com/patternfly/patternfly/blob/56b4f3fe8772e77130cc94d6d46ff3dd38d63cea/package.json#L93

OK to close this PR?

PR #5558, opened 2023-05-10
Chore table selectors 5547

### LINKED ISSUE #5549: Update Table descendent elements have versioned classnames

mattnolting:
closes #5549


src/patternfly/components/Table/examples/Table.md
```diff
@@ -125,7 +125,9 @@ import './Table.css'
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-v5-c-table` | `<table>` | Initiates a table element. **Required** |
-| `.pf-v5-c-table__caption` | `<caption>` | Initiates a table caption. |
+| `.pf-v5-c-table__tr` | `<th>` | Initiates a table row. **Required** |
```
mcoker, line 128: Doesn't look like we're always adding this class - should it be marked as required?

```suggestion
| `.pf-v5-c-table__tr` | `<tr>` | Initiates a table row. **Required** |
```

srambach (approved):
Looks good afaict 👍

thatblindgeye (approved):

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -205,7 +205,7 @@ $pf-v5-c-tree-view--MaxDepth: 10;
     }
   }
 
-  td:not(.#{$table}__tree-view-title-cell) {
+  td:where(#{$table}__tr):not(.#{$table}__tree-view-title-cell) {
```
mcoker, line 208:
```suggestion
  td:where(.#{$table}__tr):not(.#{$table}__tree-view-title-cell) {
```

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -245,10 +245,10 @@ $pf-v5-c-tree-view--MaxDepth: 10;
     }
   }
 
-  tr.pf-m-tree-view-details-expanded {
+  tr:where(.#{$table}__tr).pf-m-tree-view-details-expanded {
     padding-bottom: var(--#{$table}--m-tree-view-grid--m-tree-view-details-expanded--PaddingBottom);
 
-    td[data-label] {
+    td:where(#{$table}__td)[data-label] {
```
mcoker, line 251:
```suggestion
    td:where(.#{$table}__td)[data-label] {
```

src/patternfly/components/Table/table-tree-view.scss
```diff
@@ -289,7 +289,7 @@ $pf-v5-c-tree-view--MaxDepth: 10;
   }
 
   @for $i from 2 through $pf-v5-c-tree-view--MaxDepth {
-    tr[aria-level="#{$i}"] {
+    tr:where(#{$table}__tr)[aria-level="#{$i}"] {
```
mcoker, line 292:
```suggestion
    tr:where(.#{$table}__tr)[aria-level="#{$i}"] {
```

src/patternfly/components/Table/table-grid.scss
```diff
@@ -172,12 +172,12 @@
   }
 
   // Thead
-  thead {
+  > thead {
     display: none;
   }
```
mcoker, line 177: Not sure if this is by design or not, but all nested tables currently take on a responsive/grid layout if they're in a table with a responsive/grid layout, even when the nested table doesn't have a `.pf-m-grid*` class. Limiting the scope of selectors like this breaks this selector's role a nested table's responsive layout. For example, in this compound expansion example (https://patternfly-pr-5558.surge.sh/components/table#compound-expansion), the `thead` stuff is no longer hidden:

Screenshot 2023-05-15 at 8 26 32 PM

I wonder if maybe there are a few things we can do:
1. Add a BEM class for all things - don't use child/sibling selectors as a means to tie an element to a version. That should allow for a 1:1 replacement? So for this selector, update `thead` to `thead:where(.pf-c-table__thead)` and don't use `>`. IMO that would be the easiest, and most consistent. It would also be more clear if everything had a class, since we have a mix/match of this kind of thing`tr:where($table__tr)` and `$table > tr` and it isn't super obvious as to why. I'm often having to look at the compiled CSS to track down whether a style is tied to a versioned selector.
2. Try to fix these selectors so they behave the way they did before. I think this selector would either 1) need a `thead:where($table__thead)`? Which I guess makes me think #1 might just be easier.
3. Try and limit the scope of responsive styles so nested tables aren't impacted. IMO that would be a good change regardless. Right now you can't have a non-responsive table layout nested in a responsive layout parent table. 

Other ideas or thoughts?

src/patternfly/components/Table/table-tree-view.scss
mcoker, line 208:
```suggestion
  td:where(#{$table}__td):not(.#{$table}__tree-view-title-cell) {
```

mcoker (approved):
🥳

PR #5565, opened 2023-05-16
fix(toolbar): fixed vertical alignment issue

### LINKED ISSUE #5532: Bug - Toolbar - examples with selects misaligned on firefox
Seems to be an issue with how MenuToggles are rendering

Screenshot 2023-04-27 at 7 59 39 AM
Screenshot 2023-04-27 at 7 59 43 AM

### LINKED ISSUE #5612: Bug - Masthead - new alignment throws of toggle
**Describe the problem**
When updating #5557, I noticed that the dropdown toggle shifting up

`align-items: flex-start`
Screenshot 2023-05-24 at 1 48 08 PM

After updating the base alignment in masthead
Screenshot 2023-05-24 at 1 48 17 PM

mattnolting:
closes #5532 closes #5612

Apparently Firefox calculates baseline differently than other browsers. 

Here's an example including this update: https://codesandbox.io/s/relaxed-smoke-qfxw7r?file=/index.html

### DIFF
--- a/src/patternfly/components/Masthead/masthead.scss
+++ b/src/patternfly/components/Masthead/masthead.scss
@@ -139,6 +139,7 @@ $pf-v5-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
   --#{$masthead}--c-menu-toggle--m-full-height--before--BorderBottomColor: transparent;
 
   // Toolbar
+  --#{$masthead}--c-toolbar--AlignItems--base: center;
   --#{$masthead}--c-toolbar__content--PaddingRight: 0;
   --#{$masthead}--c-toolbar__content--PaddingLeft: 0;
   --#{$masthead}--c-toolbar__expandable-content--PaddingTop: var(--#{$pf-global}--spacer--md);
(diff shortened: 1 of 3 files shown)


mcoker (changes requested):
Looks like this is still causing an issue in FF, and potentially introduced a new issue to chrome/safari - https://patternfly-pr-5565.surge.sh/components/toolbar/html-demos/toolbar-with-validation-on-desktop/

mcoker (approved):
This seems to fix the alignment issues! 

A couple of thoughts:
* Do you want to update the item alignment classes in the same manner? https://github.com/patternfly/patternfly/blob/3925ac563a360c1ef8356246d4865ac6b198baaf/src/patternfly/components/Toolbar/toolbar.scss#L315-L321
* It's getting very flex-layout-ey but the reason we have `.pf-m-align-[items/self]-baseline`, even though the default was already baseline, was if you applied `.pf-m-align-[items/self]-center` to a _thing_, and you wanted some of the _things in that thing_ to maintain their default/baseline alignment, we provide a class for that. Now that the default is `flex-start`, should we provide a `.pf-m-align-flex-start` on things for when you use baseline/center on the thing's parent?

mattnolting:
@mcoker updated the fix. Added `align-self/items: flex-start/center/baseline` to flexbox elements and `align-self: flex-start/center/baseline` to non-flexbox (`toolbar__item`s).

mcoker (approved):
LPTM! One nit, I'd rename the `flex-start` modifier classes to `start` - we'll move there eventually as we have full browser support for removing `flex-` from those logical values, and the modifier name doesn't tie us through the wording in the name to a flex layout. We can switch to grid or whatever, and "start" will still mean the start of whatever direction it flows. wdyt?

mattnolting:
@mcoker updated!

mcoker (approved):

PR #5579, opened 2023-05-17
fix(table): updated table paddings and templates

### LINKED ISSUE #5576: Table bugs
follow up from https://github.com/patternfly/patternfly/pull/5558

* Tree with no children or indentation example - spacing is off, looks like tbody missing BEM class.
* Compact example - thead is taller than it needs to be due to an issue with the checkbox td cell in the `thead`.

mattnolting:
closes #5576 

@mcoker 

> Tree with no children or indentation example 
Fix: https://patternfly-pr-5579.surge.sh/components/table#tree-with-no-children-or-indentation

> Compact example - thead is taller than it needs to be due to an issue with the checkbox td cell in the thead.
Fix: https://patternfly-pr-5579.surge.sh/components/table#compact-example

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -1305,7 +1305,6 @@
   // Thead
   thead:where(.#{$table}__thead) {
     th:where(.#{$table}__th),
-    td:where(.#{$table}__td),
     .#{$table}__toggle {
       --#{$table}--cell--PaddingTop: var(--#{$table}--m-compact__th--PaddingTop);
       --#{$table}--cell--PaddingBottom: var(--#{$table}--m-compact__th--PaddingBottom);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -1305,7 +1305,6 @@
   // Thead
   thead:where(.#{$table}__thead) {
     th:where(.#{$table}__th),
-    td:where(.#{$table}__td),
```
mcoker, line 1308: Works for me! I'll just mention that this looks to be due to the checkbox cell specifically since the checkbox has a different `font-size` than the rest of the cells (it's font size * line height makes the cell taller than other 1-line `<th>` cells). So this fixes it, but I dunno if we ever want to put something else in a `thead > td` and have that cell have the same padding as the rest of the `th` cells.

Another approach could be to target something about `.pf-c-table__check` in a compact table > `thead`, or just in a compact table in general. It looks like the presence of the table check element is making all compact table rows taller than they would be if the check wasn't there or used the same font-size as the compact table cell text. Here are 2 duplicate compact tables - one with checkboxes and one without (cc @andrew-ronaldson @mceledonia)

Screenshot 2023-05-17 at 9 18 06 PM

mcoker (approved):
Left one comment - lemme know what you think, we can follow up if needed.

Issue #5584, opened 2023-05-17
Layouts - update layouts that use flex and support gutters to use gap


mcoker:
Looks like this is level, split, and stack. These should use gap instead of margins.

PR #5585, opened 2023-05-17
fix(layouts): update level, split, stack to use gap

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5584


mattnolting (approved):
:ship it! LPTM

Issue #5596, opened 2023-05-18
Bug - Cards - clickable cards are showing a radio button that should be hidden


srambach:
**Describe the problem**
Clickable cards are showing a radio button. This is a regression from when the feature was added in https://patternfly-pr-5409.surge.sh/components/card#clickable Here you can see the radio button is hidden.
Current: https://pf5.patternfly.org/components/card#clickable

**Expected behavior**
The design was not to show a radio button (or checkbox) https://github.com/patternfly/patternfly/issues/5264

srambach:
Looks like there was a merge problem with the variable versioning
```
<input
              class="pf-v5-c-radio__input {{pfv &#x27;unset-prefix&#x27;}}screen-reader"
              type="radio"
              id="card-clickable-example-sr-only-radio"
              name="card-clickable-example-sr-only-radio"
              aria-labelledby="card-clickable-example"
 />
```

PR #5597, opened 2023-05-18
fix(card): fixed screen-reader class on clickable card radio

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5596

### DIFF
--- a/patternfly-docs/site/pages/accessibility-guide.md
+++ b/patternfly-docs/site/pages/accessibility-guide.md
@@ -120,7 +120,7 @@ The [WCAG 2.0 techniques](https://www.w3.org/TR/WCAG20-TECHS/Overview.html#conte
         <i class="..." aria-hidden="true"></i>
       </button>
       ```
-      - For non-interactive icons, include `.pf-screen-reader` text near the icon. Depending on the component, the `.pf-screen-reader` text might not be a direct sibling to the icon element. For example, in the alert component, the icon label text is adjacent to the message. This way, when `role="alert"` is added to `.pf-v5-c-alert__body` for dynamically displayed alerts, the type of message is announced along with the message text.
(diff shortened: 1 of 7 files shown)


srambach:
This is putting the version on the screen-reader class but it shouldn't have one, should it?
```
            <input
              class="pf-v5-c-radio__input **pf-v5-screen-reader**"
              type="radio"
              id="card-clickable-example-sr-only-radio"
              name="card-clickable-example-sr-only-radio"
              aria-labelledby="card-clickable-example"
            />
```
https://patternfly-pr-5597.surge.sh/components/card#clickable

mcoker:
@srambach not sure if I understand.  If you're referring to `pf-v5-screen-reader`, AFAIK that does need to be prefixed - for v6 or whatever, we may change the way we apply those styles and wouldn't want conflicts with the old v5 and new v6 screen-reader class styles.

srambach:
I thought the utility `pf-u-scrnee-reader` was prefixed, but `pf-screen-reader` was not? https://patternfly-pr-5597.surge.sh/utility-classes/accessibility

mcoker:
If you're asking if it is currently prefixed, it's defined here:

https://github.com/patternfly/patternfly/blob/474883c523da6bf7b5bbc44dc415b2d15429565c/src/patternfly/base/_common.scss#L3

And you can see it applied in the preview

Screenshot 2023-05-18 at 12 45 42 PM

srambach (approved):
🤷‍♀️ OK - Then we also need to fix up the A11y documentation. And it looks like a couple other places have incorrect documentation. (calendar, popover)

mcoker:
@srambach ah I gotcha. Also banner and input-group's docs reference `.pf-screen-reader`. I imagine those doc updates got lost in the shuffle as I don't believe the class was prefixed to begin with, then we added the prefix. Though I'd like @mattnolting to confirm whether the global classes should have the prefix or not. I'm assuming they should for the reason I gave above (along with any other kind of [global/common class](https://github.com/patternfly/patternfly/blob/v5/src/patternfly/base/_common.scss) we have like that), but Matt may know more about it than I do.

mattnolting:
@mcoker @srambach We concluded that, because the method will potentially change in the future, `.pf-screen-reader` should be versioned. It's correct that the docs should be updated.

mcoker:
Thanks @mattnolting! PR should be ready to review, the docs are updated.

mattnolting (commented):
LPTM 👍

mattnolting (approved):

Issue #5599, opened 2023-05-18
Flex - xs gap spacer not working


mcoker:
https://github.com/patternfly/patternfly-react/pull/9143#pullrequestreview-1433380716

PR #5600, opened 2023-05-18
fix(flex): added xs gap spacer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5599


nicolethoen (approved):

thatblindgeye (approved):
Flex with extra small gap

🎉

PR #5607, opened 2023-05-22
fix(Nav): updated spacing for grouped nav

### LINKED ISSUE #5592: Nav - group spacing is off
Small bug left from https://github.com/patternfly/patternfly/pull/5507

Screenshot 2023-05-17 at 8 40 11 PM

thatblindgeye:
Closes #5592 

@mcoker @andrew-ronaldson Tried comparing examples/demos that had screenshots in the previous PR to the local build for this update. Seemed like the local build matched what was expected visually, but some spacing between screenshots was pretty close to call for me.

It seems like the issue was more tied to a section that also had `pf-m-no-title`, so the following also seemed to work:

```
  &.pf-m-no-title {
    --#{$nav}__list--PaddingTop: 0;
    --#{$nav}__section--MarginTop: 0;
  }
```

### DIFF
--- a/src/patternfly/components/Nav/examples/Navigation.md
+++ b/src/patternfly/components/Nav/examples/Navigation.md
@@ -89,7 +89,7 @@ import './Navigation.css'
 ### Grouped nav, no titles
 ```hbs
 {{#> nav nav--attribute='aria-label="Global"'}}
-  {{#> nav-section nav-section--modifier="pf-m-no-title" nav-section--attribute='aria-label="Section one"'}}
+  {{#> nav-section nav-section--attribute='aria-label="Section one"'}}
     {{#> nav-list}}
       {{#> nav-item}}
         {{#> nav-link nav-link--href="#"}}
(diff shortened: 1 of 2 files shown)


mattnolting:
I think removing the [margin](https://github.com/patternfly/patternfly/pull/5607/files#diff-98e414994743ef6a80d6a54353a2d0df03e8af105f92236dd1ba848e28c8add7R265) from `pf-c-nav--c-divider` and [margin-top](https://github.com/patternfly/patternfly/blob/4eeffe17067be041a5732aa8bb539f827d19526b/src/patternfly/components/Nav/nav.scss#L206)  from `pf-c-nav__section` and [applying margin directly to](https://github.com/patternfly/patternfly/blob/v5/src/patternfly/components/Nav/nav.scss#L224) `__section + __section` would do it.

thatblindgeye:
@mattnolting doing that (or at least reverting the updates in this PR and making the above updates; didn't add any margin-bottom to __section + __section) results in this:

image

Where the first "Link 3" doesn't have any spacing between itself and the divider.

thatblindgeye (commented):
This looks nice with the updates! One thing is the `no titles no margin` example spacing:

image

Would we want to still update this example to be "Grouped nav, no titles, no divider"?

mattnolting (approved):
Nice work! LPTM 👍

PR #5609, opened 2023-05-23
fix(form-control): update form-control in other components

### LINKED ISSUE #5566: Form control - update in demos
As a follow up to #5469, there are uses of form controls elsewhere in demos that now need to be fixed for the new structure.

```[tasklist]
### Places that form control is used and may need to be updated
- [ ] Alert horizontal and stacked forms
- [ ] Calendar - no update needed
- [ ] Clipboard copy
- [ ] Datepicker
- [ ] File upload
- [ ] Form
- [ ] Inline edit
- [ ] Input group
- [ ] Login
- [ ] Number input
- [ ] Pagination
- [ ] Select
- [x] Slider - fixed in https://github.com/patternfly/patternfly/pull/5590
- [ ] Text input group
- [ ] Basic forms
- [ ] Helper text
- [ ] Modal
- [ ] Password generator
- [ ] Password strength
- [ ] Toolbar
- [ ] Wizard form
```

srambach:
Fixes #5566
Form control height was previously removed, but is necessary in some cases - however this causes a slight difference elsewhere. 

Alert horizontal and stacked forms
Calendar - no update needed
Clipboard copy
Datepicker
File upload - must set form control option (readonly) with the form-element hbs parameter
Form
Inline edit - also removed a large block of commented code that is no longer needed
Input group
Login
Number input
Pagination
Select
Slider - fixed in https://github.com/patternfly/patternfly/pull/5590
Text input group - no change necessary
Basic forms
Helper text
Modal
Password generator
Password strength
Toolbar
Wizard form

### DIFF
--- a/src/patternfly/components/Form/examples/Form.md
+++ b/src/patternfly/components/Form/examples/Form.md
@@ -17,7 +17,7 @@ cssPrefix: pf-v5-c-form
       {{> form-group-label-help form-group-label-help--aria-label="More information for name field" form-group-label-help--aria-describedby=form-group--id}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '" required')}}{{/form-control}}
+      {{#> form-control controlType="input" input="true" form-control--IsRequired='true' form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '"')}}{{/form-control}}
     {{/form-group-control}}
   {{/form-group}}
 {{/form}}
@@ -33,7 +33,7 @@ cssPrefix: pf-v5-c-form
       {{/form-label}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '" required')}}{{/form-control}}
+      {{#> form-control controlType="input" input="true" form-control--IsRequired='true' form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '"')}}{{/form-control}}
     {{/form-group-control}}
   {{/form-group}}
   {{#> form-group form-group--id=(concat form--id '-info')}}
@@ -79,7 +79,7 @@ cssPrefix: pf-v5-c-form
       {{> form-group-label-help form-group-label-help--aria-label="More information for name field" form-group-label-help--aria-describedby=(concat form-group--id)}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '" required')}}{{/form-control}}
+      {{#> form-control controlType="input" input="true" form-control--IsRequired='true' form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '"')}}{{/form-control}}
     {{/form-group-control}}
   {{/form-group}}
 {{/form}}
@@ -96,7 +96,7 @@ cssPrefix: pf-v5-c-form
         {{/form-label}}
       {{/form-group-label}}
       {{#> form-group-control}}
-        {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form-section--id '-input" name="' form-section--id '-input" required')}}{{/form-control}}
+        {{#> form-control controlType="input" input="true" form-control--IsRequired='true' form-control--attribute=(concat 'type="text" id="' form-section--id '-input" name="' form-section--id '-input"')}}{{/form-control}}
       {{/form-group-control}}
     {{/form-group}}
     {{#> form-group}}
@@ -106,7 +106,7 @@ cssPrefix: pf-v5-c-form
         {{/form-label}}
       {{/form-group-label}}
       {{#> form-group-control}}
-        {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form-section--id '-input-2" name="' form-section--id '-input-2" required')}}{{/form-control}}
+        {{#> form-control controlType="input" input="true" form-control--IsRequired='true' form-control--attribute=(concat 'type="text" id="' form-section--id '-input-2" name="' form-section--id '-input-2"')}}{{/form-control}}
       {{/form-group-control}}
     {{/form-group}}
   {{/form-section}}
@@ -121,7 +121,7 @@ cssPrefix: pf-v5-c-form
         {{/form-label}}
       {{/form-group-label}}
       {{#> form-group-control}}
-        {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'type="text" id="' form-section--id '-input" name="' form-section--id '-input" required')}}{{/form-control}}
+        {{#> form-control controlType="input" input="true" form-control--IsRequired='true' form-control--attribute=(concat 'type="text" id="' form-section--id '-input" name="' form-section--id '-input"')}}{{/form-control}}
       {{/form-group-control}}
     {{/form-group}}
     {{#> form-group}}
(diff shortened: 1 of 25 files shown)


src/patternfly/components/Select/select-toggle-typeahead.hbs
```diff
@@ -1,13 +1,15 @@
...
+    form-control--attribute=(concat 'type="text" id="' select--id '-typeahead" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
   {{/form-control}}
 {{else}}
   {{#if select--IsInvalid}}
     {{#> form-control controlType="input"
       input="true"
+      form-control--IsError='true'
       form-control--modifier="pf-v5-c-select__toggle-typeahead"
       form-control--attribute=(concat 'type="text" id="' select--id '-typeahead" aria-invalid="true" value="Invalid" aria-label="Type to filter" placeholder="' select-typeahead--Placeholder '"')}}
     {{/form-control}}
```
thatblindgeye, line 15: Duplicate error icons are being rendered in the Invalid example for Select typeahead with this addition since select-toggle is setting an icon as well.

thatblindgeye (changes requested):
Couple comments below. Also from the following comment on the React PR that  updated FormControls to use the new wrapper, seems like we should remove the `form-control--IsPlaceholder` from line 16 of `FormControl.md` examples (should be the second input in the Input section) https://github.com/patternfly/patternfly-react/pull/9132#discussion_r1197967968

src/patternfly/demos/Alert/alert-template-horizontal-form.hbs
```diff
@@ -28,7 +28,7 @@
       {{/form-label}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control controlType="input" input=true form-control--attribute=(concat 'type="text" value="patternfly@patternfly.com" id="' form-group--id '" name="' form-group--id '" required')}}{{/form-control}}
+      {{#> form-control controlType="input" input=true form-control--IsError="true" form-control--IsRequired="true" form-control--attribute=(concat 'type="text" value="patternfly@patternfly.com" id="' form-group--id '" name="' form-group--id )}}{{/form-control}}
```
mcoker, line 31:
```suggestion
      {{#> form-control controlType="input" input=true form-control--IsRequired="true" form-control--attribute=(concat 'type="text" value="patternfly@patternfly.com" id="' form-group--id '" name="' form-group--id )}}{{/form-control}}
```

src/patternfly/demos/Alert/alert-template-stacked-form.hbs
```diff
@@ -16,7 +16,7 @@
       {{/form-label}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control controlType="input" input="true" form-control--attribute=(concat 'required type="text" id="' form-group--id '" name="' form-group--id '" aria-invalid="true" aria-describedby="' form--id '-helper"')}}
+      {{#> form-control controlType="input" input="true" form-control--IsError="true" form-control--attribute=(concat 'required type="text" id="' form-group--id '" name="' form-group--id '" aria-invalid="true" aria-describedby="' form--id '-helper"')}}
```
mcoker, line 19:
```suggestion
      {{#> form-control controlType="input" input="true" form-control--IsError="true" form-control--IsRequired="true" form-control--attribute=(concat 'type="text" id="' form-group--id '" name="' form-group--id '" aria-invalid="true" aria-describedby="' form--id '-helper"')}}
```

src/patternfly/demos/Alert/alert-template-stacked-form.hbs
```diff
@@ -39,7 +39,7 @@
         State of residence
       {{/form-label}}
     {{/form-group-label}}
-    {{#> form-control controlType="select" form-control--attribute='required aria-invalid="true" id="select-group-error" name="select-group-error" aria-label="Error state select group example"'}}
+    {{#> form-control controlType="select" form-control--IsError="true" form-control--attribute='required aria-invalid="true" id="select-group-error" name="select-group-error" aria-label="Error state select group example"'}}
```
mcoker, line 42:
```suggestion
    {{#> form-control controlType="select" form-control--IsError="true" form-control--IsRequired="true" form-control--attribute='aria-invalid="true" id="select-group-error" name="select-group-error" aria-label="Error state select group example"'}}
```

mcoker (changes requested):
if `form-control--IsError` is set along every use of `aria-invalid="true"`, would we update the form-control hbs to add that attribute like it does with readonly, required, etc?

Our use of `.pf-m-placeholder` (`form-control--IsPlaceholder`) is inconsistent, and I'm not sure your original intent. Technically `.pf-m-placeholder` doesn't do anything on a non-select, since we style `::placeholder`, but our [basic placeholder text input example](https://github.com/patternfly/patternfly/blob/62d313773fcbcff6135791813f50e04f2f838643/src/patternfly/components/FormControl/examples/FormControl.md?plain=1#L16) uses `form-control--IsPlaceholder`

I mostly just reviewed the code and only spot checked a couple of things in the browser, but I noticed this number input example doesn't have an icon - https://patternfly-pr-5609.surge.sh/components/number-input#with-status

srambach:
@mcoker:
> if form-control--IsError is set along every use of aria-invalid="true", would we update the form-control hbs to add that attribute like it does with readonly, required, etc?

In the interest of getting this in today, I didn't go through and do that yet but I can do it after today, as it's just a handlebars/md change.

> We've previously decided to not progressively enhance like this since it can lead someone to think things look OK across the browsers we support if it looks OK in theirs. Probably not a big deal, but I'm not sure why this is needed and what it would look like for FF since that doesn't support :has().
> 
> I'm also wondering if we could get away from this height all together though. Setting the height in this PR is causing form controls in chrome to be inconsistent heights, depending on whether they have a status or not, due to the difference in the border size.

I had initially taken out the height, but saw a small difference in pagination. I took it back out. 

> Now that we have a div wrapping these controls, I think we should move the borders from the inputs to the parent div if possible - that will simplify the CSS. The paddings, and maybe other things, are calc()s based on the border size that could all be removed since borders on pseudo elements will not impact the layout.

Not sure what you mean here - the border is already on the `div.pf-v5-c-form-control`. Do you mean to move them to a pseudo element of that div?

thatblindgeye (approved):

mcoker:
> Not sure what you mean here - the border is already on the div.pf-v5-c-form-control. Do you mean to move them to a pseudo element of that div?

@srambach doh! yep, I meant to say pseudo elements on the div.

src/patternfly/components/Form/examples/Form.md
```diff
@@ -160,7 +160,7 @@ cssPrefix: pf-v5-c-form
       {{/form-label}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control form-control--modifier="pf-m-warning" controlType="input" input="true" form-control--attribute=(concat 'required type="text" id="' form-group--id '" name="' form-group--id '" aria-describedby="' form-group--id '-helper"')}}
+      {{#> form-control form-control--IsWarning='true' controlType="input" input="true" form-control--attribute=(concat 'required type="text" id="' form-group--id '" name="' form-group--id '" aria-describedby="' form-group--id '-helper"')}}
```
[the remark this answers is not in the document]
mcoker: Looks like you missed this one ^

Also looks like there are some other places we aren't using `required` and `form-control--IsRequired` (sets `required` and `.pf-m-required`) consistently. It would take a while to track them all down, but all of the `--IsError` examples on the main form control examples page are only setting `required`, and I counted 67 other spots in this PR where we're setting `form-control--IsRequired`. It doesn't look like `.pf-m-required` does anything, so I don't know that we need that class. If not, we could just remove this line - https://github.com/patternfly/patternfly/blob/03a2d78de7fd154883025ab39e34a6d693a32516/src/patternfly/components/FormControl/form-control.hbs#L5

Then, either setting `required` or `form-control--IsRequired` would do the same exact thing, which would clean up the markup in our examples, and we can clean up the way we're writing the hbs some other time.

src/patternfly/demos/Alert/alert-template-horizontal-form.hbs
```diff
@@ -28,7 +28,7 @@
       {{/form-label}}
     {{/form-group-label}}
     {{#> form-group-control}}
-      {{#> form-control controlType="input" input=true form-control--IsError="true" form-control--IsRequired="true" form-control--attribute=(concat 'type="text" value="patternfly@patternfly.com" id="' form-group--id '" name="' form-group--id )}}{{/form-control}}
+      {{#> form-control controlType="input" input='true' form-control--IsRequired='true' form-control--attribute=(concat 'type="text" value="patternfly@patternfly.com" id="' form-group--id '" name="' form-group--id '"')}}{{/form-control}}
```
mcoker, line 31: Just a comment, I see you added single quotes around an unquoted `true` and changed double quotes to single quotes. For these boolean things, they don't need quotes at all, and typically we use double quotes to wrap param values except values that will contain double quotes (most commonly like `--attribute='something="whatever"'`)

mcoker (approved):
LGTM - left some comments, feel free to open follow ups if you think they're needed.

PR #5611, opened 2023-05-24
fix(toolbar): updated positioning of expanded content

### LINKED ISSUE #5557: Bug - Masthead - Search input in demo is outside of masthead
**Describe the problem**
This demo shows the search input going across the page instead of tucked into the masthead.
https://pf5.patternfly.org/components/masthead/html-demos/with-toggle-group-filters-expandable-content-expanded-mobile/

mattnolting:
closes #5557 

The layout shift was caused by adding `row-gap` https://github.com/patternfly/patternfly/blame/13ed13afc9c20557d88d90533b2fdb9af646ca71/src/patternfly/components/Toolbar/toolbar.scss#L383

So I moved the expandable content background to a `::before` pseudo.

### DIFF
--- a/src/patternfly/components/Masthead/masthead.scss
+++ b/src/patternfly/components/Masthead/masthead.scss
@@ -147,7 +147,6 @@ $pf-v5-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
   --#{$masthead}--c-toolbar__expandable-content--PaddingRight: var(--#{$masthead}--inset);
   --#{$masthead}--c-toolbar__expandable-content--PaddingBottom: var(--#{$pf-global}--spacer--md);
   --#{$masthead}--c-toolbar__expandable-content--PaddingLeft: var(--#{$masthead}--inset);
-  --#{$masthead}--c-toolbar__expandable-content--before--Top: calc(var(--#{$masthead}--c-toolbar__expandable-content--PaddingTop));
   --#{$masthead}--c-toolbar__expandable-content--before--BorderTopWidth: var(--#{$pf-global}--BorderWidth--sm);
   --#{$masthead}--c-toolbar__expandable-content--before--BorderTopColor: var(--#{$masthead}--item-border-color--base);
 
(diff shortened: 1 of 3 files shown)


mcoker (commented):
I'm not too familiar with all the fancy ways the toolbar can be used, but I'm wondering why we removed the expandable content element's top padding and offset its `top` position to create space in https://github.com/patternfly/patternfly/commit/9bf639285f994882414790a38e3e76219b6e08f6? As far as I can tell, that element is always absolutely positioned, so a `gap` on the parent shouldn't apply to it since it's removed from the layout. Even if it isn't removed from the layout (not absolutely positioned), I think using a negative margin equal to the gap size to offset it would work great since this element will never wrap.

thatblindgeye (commented):
Other than above, just wanted to confirm that the demo should look like this:
Toolbar mobile example in PR


Compared to v4 where the search input has the light background and spans across the entire width of the toolbar (rather than visually positioned far off to the right):

v4 mobile toolbar example

Might be unrelated and wouldn't block this over it (probably a quick demo update as a followup), but it looks like we're incorrectly applying `pf-m-fill` to the `.pf-v5-c-input-group__item` that contains the "Name" `.pf-v5-c-select`, then the other `pf-v5-c-input-group__item` should have `pf-m-fill` and the `pf-v5-c-text-input-group` shouldn't have the `width: auto` styling.

mattnolting:
@mcoker @thatblindgeye This should be added to `.pf-c-masthead` instead of `.pf-c-toolbar`, as this functionality only occurs there. 


> Other than above, just wanted to confirm that the demo should look like this: Toolbar mobile example in PR

@thatblindgeye That's right, the expanded content doesn't shift page layout. 

> Might be unrelated and wouldn't block this over it (probably a quick demo update as a followup), but it looks like we're incorrectly applying pf-m-fill to the .pf-v5-c-input-group__item that contains the "Name" .pf-v5-c-select, then the other pf-v5-c-input-group__item should have pf-m-fill and the pf-v5-c-text-input-group shouldn't have the width: auto styling.

@thatblindgeye That would be demo update 👍

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -190,10 +190,10 @@ $pf-v5-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
   }
 
   .#{$toolbar} {
+    --#{$toolbar}--BackgroundColor: var(--#{$masthead}--c-toolbar--BackgroundColor);
     --#{$toolbar}--AlignItems--base: var(--#{$masthead}--c-toolbar--AlignItems--base);
     --#{$toolbar}__content--PaddingRight: var(--#{$masthead}--c-toolbar__content--PaddingRight);
     --#{$toolbar}__content--PaddingLeft: var(--#{$masthead}--c-toolbar__content--PaddingLeft);
-    --#{$toolbar}__expandable-content--PaddingTop: var(--#{$masthead}--c-toolbar__expandable-content--PaddingTop);
```
thatblindgeye, line 196: Not a blocker, but does this toolbar var need to be removed or could it be used on line 210 instead?

thatblindgeye (approved):

mcoker (approved):
🔥

PR #5619, opened 2023-05-25
fix(Pagination): updated styles to target menu toggle

### LINKED ISSUE #5610: Pagination - update reference to deprecated options menu
Options menu is used for the pagination dropdown, and we style the options menu directly in pagination. That no longer works now that the options menu is deprecated and we're using a menu/menu-toggle. We need to update that to either

1. Style the menu/menu-toggle instead
2. If possible and time allows, I think the ideal update would be to not style another component directly, and move styles to the pagination component and potentially introduce options to the menu/menu-toggle for this use case

thatblindgeye:
Closes #5610 

Will need a followup to rework the styling to not target MenuToggle directly inside Pagination.


mcoker (approved):
Looks like you've still got a couple of vars to update, not blocking though if we want to update that later, just lemme know!

https://github.com/patternfly/patternfly/blob/60840f559d2afbdab6d0bd463e6ec1ddc27f15ba/src/patternfly/components/Pagination/pagination.scss#L80-L81

mcoker (approved):
💯

PR #5623, opened 2023-05-25
fix(theme): fix dark theme global var specificity

### LINKED ISSUE #5621: Dark theme style selector specificity
https://github.com/patternfly/patternfly/pull/5598/files#diff-735d6082e3c81d5c30b7bea7871588a4d944b65fd492d4096da6b672ec248bdeL299-R300

Looks like in this change, we changed the dark theme class selector from `:where(.pf-theme-dark) .pf-c-component { // overrides }` to `:root:where(.pf-theme-dark) .pf-c-component { // overrides }` which is increasing the specificity of overrides and breaking some of the styles by adding `:root`. The prefix to the dark theme selector should not be set by default, and only be included if one is set.

mattnolting:
closes #5621

### DIFF
--- a/src/patternfly/base/_variables.scss
+++ b/src/patternfly/base/_variables.scss
@@ -1,4 +1,4 @@
-:root {
+:where(:root) {
   // Colors
   // Global palette
   --#{$pf-global}--palette--black-100: #{$pf-v5-color-black-100};
--- a/src/patternfly/patternfly-charts.scss
+++ b/src/patternfly/patternfly-charts.scss
@@ -8,7 +8,7 @@
   }
 }
 
-:root {
+:where(:root) {
   // Chart colors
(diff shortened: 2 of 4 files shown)


mcoker:
Dark theme is still broken because base css vars are set on `:root`, and we aren't setting the target class for the dark theme css var overrides, so the overrides are `:where(.pf-dark-theme-class)` (lower specificity). The easiest solution is probably just to wrap our `:root` styles in `:where()`, and use `:where(.pf-dark-theme-class` across the board for overrides. That makes our base styles for our global vars have no specificity, so they're easily overrideable like our global/reset styles are now, which seems to make sense. I went ahead and made that update and will merge it to fix the dark theme styles. Feel free to open another issue if you think there is a better way to do that!

Issue #5272, opened 2022-12-01
Calendar month has padding that keeps it from aligning with content around it


mcoker:
looking at the [calendar month examples](https://www.patternfly.org/v4/components/calendar-month), the calendar component has padding around it that will provide space around it wherever it is used. Typically our components don't have padding like that and extend to the edges of wherever they're placed[1], if that were the case here, the thing the calendar is placed inside of would provide that padding instead of the calendar creating the padding itself.

You can see here how the calendar content does not align with the content above it because of the calendar's built-in padding.

<img width="586" alt="Screenshot 2022-12-01 at 4 55 47 PM" src="https://user-images.githubusercontent.com/35148959/205176846-e6cb9ca8-c84f-4da0-8465-55e26209bac9.png">

<img width="530" alt="Screenshot 2022-12-01 at 4 59 26 PM" src="https://user-images.githubusercontent.com/35148959/205176973-2fe5680c-07a8-4a36-b10e-f71e48b239ab.png">

#### Here is what it looks like if the calendar's default padding is removed

<img width="527" alt="Screenshot 2022-12-01 at 4 59 53 PM" src="https://user-images.githubusercontent.com/35148959/205177286-9c73a555-23be-4e65-b246-13d1c3eb54a7.png">

#### Here is what the calendar looks like in the date picker popover
Note that the popover's built-in padding has been removed (using a no-padding variation), and the calendar component's space is creating the padding.

<img width="751" alt="Screenshot 2022-12-01 at 4 52 54 PM" src="https://user-images.githubusercontent.com/35148959/205176217-30a4dbb9-c164-4bd2-8868-f4e4e18bc4f7.png">

#### Here is what it looks like if the calendar component's padding is removed, and the popover creates the padding

<img width="639" alt="Screenshot 2022-12-01 at 4 53 19 PM" src="https://user-images.githubusercontent.com/35148959/205176156-5d1fe0d9-9d2b-484f-8edf-21e5274b53b6.png">

One gotcha with the padding is that the calendar month's built-in padding has a larger top padding than bottom padding, to help balance the top/bottom whitespace visually since there is extra padding below each date. That extra space is so we can draw a circle around the date when it's selected. So if none of the dates in the bottom row are selected, it looks like the spacing is imbalanced. Here's what it looks like with a date selected to illustrate what's going on there.

[...]

PR #5624, opened 2023-05-25
fix(CalendarMonth): update bg color

### LINKED ISSUE #5615: Bug - Calendar month - Styling of calendar does not work for the Date Range example
**Describe the problem**
In Calendar month, range indication not visible when selecting a range of dates in dark theme.

How it looks:

Screenshot 2023-05-24 at 4 10 33 PM

How it looks in light theme

Screenshot 2023-05-24 at 4 10 19 PM

jenny-s51:
Closes #5615

### DIFF
--- a/src/patternfly/components/CalendarMonth/calendar-month.scss
+++ b/src/patternfly/components/CalendarMonth/calendar-month.scss
@@ -1,4 +1,5 @@
 .#{$calendar-month} {
+  --#{$calendar-month}--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
   --#{$calendar-month}--PaddingTop: var(--#{$pf-global}--spacer--lg);
   --#{$calendar-month}--PaddingRight: var(--#{$pf-global}--spacer--lg);
   --#{$calendar-month}--PaddingBottom: var(--#{$pf-global}--spacer--md);
@@ -68,6 +69,7 @@
   flex-direction: column;
   padding: var(--#{$calendar-month}--PaddingTop) var(--#{$calendar-month}--PaddingRight) var(--#{$calendar-month}--PaddingBottom) var(--#{$calendar-month}--PaddingLeft);
   font-size: var(--#{$calendar-month}--FontSize);
+  background-color: var(--#{$calendar-month}--BackgroundColor);
 }
 
 .#{$calendar-month}__header {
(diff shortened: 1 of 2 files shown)


mattnolting (approved):
LPTM 👍

mcoker:
Adding @mcarrano and @mceledonia  - just want to make sure the original issue is with the background color of the range not having enough contrast. Another potential issue with the react date range example from the screenshot is that the "end date" has a circle around the date and isn't blue, even though the example is supposed to indicate a start/end date range - just want to note that it's because the end date has a forced "hover" state applied for the purpose of the example. Here's what that example looks like if we remove the hover state

Screenshot 2023-05-25 at 9 51 01 PM

This PR adds the primary background color to the default theme calendar month component, and the "floating" background color to the dark theme calendar month. I want to say we talked about this at some point and that the calendar month isn't limited to use in the popover (popover uses the floating background color), so we decided remove the calendar month background all together so it can be used anywhere and its background  comes from wherever it's used (the background was removed in this PR - https://github.com/patternfly/patternfly/commit/bf43ac007cb53e76b6541e9812842fce2d38f27d#diff-854e9254d40eeb86f001f44ff5dbe8fe7b890d609dd314ea357eefd54eec0899L4). Another update to allow the popover to be used anywhere is this issue to [remove the calendar month's built-in padding](https://github.com/patternfly/patternfly/issues/5272), and the calendar month could be placed in a card, for example, as a fixed element on the page, and the calendar month's background would take on the background of the card.

Screenshot 2023-05-25 at 11 01 30 PM

Otherwise it would look something like this if the calendar month used the floating background and was placed in a card.

Screenshot 2023-05-25 at 11 01 50 PM

Another option here could be to change the range background color so it works on any of our background colors? Currently the range background uses the secondary background color.

mcarrano (commented):
This is definitely better, but I see what you mean about the background @mcoker. Not sure what the best solution for that would be. In the card example, perhaps we could add a border around the calendar? But I understand that would be a design change and probably require a new prop. Maybe we could live with this in the short term and fix post-v5.

For the end date, I'm not clear why we are forcing the hover. Why does this look OK in light mode? @mceledonia @andrew-ronaldson thoughts?

mcarrano (approved):
Let's run with this... Thanks @jenny-s51 and @mcoker !

mcoker (approved):
⭐️

Issue #5386, opened 2023-02-15
Deprecate page header


nicolethoen:
All docs should mark page header and its subcomponents as deprecated. 
the page header example on the page docs should not be removed but should be marked deprecated.
All other demos through out the workspace should be updated to use the masthead and not page header.

nicolethoen:
Class names in the docs table will need to be moved to the new deprecated page

Issue #5292, opened 2022-12-20
Options menu - Combine with Select and deprecate


mcarrano:
Want to deprecate the existing Options Menu component in favor of including this functionality as a type of Select. Work requested includes:

* Adding an Options menu example to Select.
* Deprecating the existing Options menu

Issue #5291, opened 2022-12-19
Context selector - Deprecate in favor of composable demo


mcarrano:
Deprecate the existing component in favor of the composable context selector demo.

Issue #5290, opened 2022-12-19
Application Launcher - deprecate 


mcarrano:
Want to deprecate the existing application launcher in favor of the composable application launcher demo.

PR #5626, opened 2023-05-25
feat(deprecations): deprecate PageHeader and add banner to other deprecated components

nicolethoen:
Closes https://github.com/patternfly/patternfly/issues/5290
Closes https://github.com/patternfly/patternfly/issues/5291
Closes https://github.com/patternfly/patternfly/issues/5292
Closes https://github.com/patternfly/patternfly/issues/5386

Also updates the CSS vars tables for Chip, Alert, Label docs
And removes the word legacy from Tabs and Nav examples - uses deprecated instead


nicolethoen:
Hold off on reviewing this -  I think I have an alternate - less aggressive solution.

nicolethoen:
Pause on reviewing until I can make a change in the doc-framework to resolve a11y issues.

mcarrano (commented):
@nicolethoen this looks good, but I am a bit confused about whether the `deprecated` labels should be showing up in the nav. I'm also not sure why Dontent selector has a `demo` label but not `deprecated` Is that demo also considered deprecated? It's not clear what I'm looking at.

src/patternfly/demos/Nav/examples/Nav.md
```diff
@@ -44,8 +44,8 @@ section: components
 }}
 ```
 
-### Legacy tertiary nav
-```hbs isFullscreen
+### tertiary nav
```
thatblindgeye, line 47: Sentence case:
```suggestion
### Tertiary nav
```

src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -92,24 +92,24 @@ import './Tabs.css'
 {{/tabs}}
 ```
 
-### Vertical expandable example (legacy)
-```hbs
+### Vertical expandable example (deprecated)
```
thatblindgeye, line 95: We probably don't need the `(deprecated)` anymore with the new label being applied.

Could we also add the deprecated label to the Card examples marked deprecated?

thatblindgeye (changes requested):
For the deprecated banner, the "about page" link leads to a 404.

I imagine the "demo" label being applied to Context Selector in the side nav might be due there being no plain "HTML" page for the component?

src/patternfly/components/Tabs/examples/Tabs.md
nicolethoen: for now I left it because there were other examples with the same name without it. We can do a follow up in the future to rename these examples in a better way.

Issue #5627, opened 2023-05-26
Form control - disabled and readonly dark theme bugs


mcoker:
dark styles are not targeting new var names for readonly/disabled

srambach:
The readonly dark versions of the form-control and code editor are showing insufficient contrast - looks like it should still have the normal text color.

Also, looks like the hover causes the bottom border to disappear on the form control text input and text area.

PR #5628, opened 2023-05-26
fix(form-control): fixed dark theme readonly/disabled styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5627


mcarrano (changes requested):
@mcoker this looks much better now. Just one small problem. The Read-only field should retain the bottom border like other fields (except disabled). It does in light theme but not visible in dark.

@andrew-ronaldson and @mceledonia can you also take a look?

andrew-ronaldson (approved):
LGTM @mcoker

thatblindgeye (approved):

PR #5641, opened 2023-06-01
fix(form-control): move borders to pseudo element, remove extra css

mcoker:
Updates the form control to move the borders to pseudo elements, remove all of the custom height and padding `calc()`s that were needed to offset the borders since they were part of the component layout.

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -1,88 +1,81 @@
 .#{$form-control} {
-  // Component
   --#{$form-control}--ColumnGap: var(--#{$pf-global}--spacer--sm);
   --#{$form-control}--Color: var(--#{$pf-global}--Color--100);
   --#{$form-control}--FontSize: var(--#{$pf-global}--FontSize--md);
   --#{$form-control}--LineHeight: var(--#{$pf-global}--LineHeight--md);
   --#{$form-control}--Resize: none;
-  --#{$form-control}--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
-  --#{$form-control}--BorderTopColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderRightColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderBottomColor: var(--#{$pf-global}--BorderColor--200);
-  --#{$form-control}--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderRadius: 0;
+  --#{$form-control}--after--BorderTopWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderRightWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderLeftWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderTopColor: var(--#{$pf-global}--BorderColor--300);
+  --#{$form-control}--after--BorderRightColor: var(--#{$pf-global}--BorderColor--300);
+  --#{$form-control}--after--BorderBottomColor: var(--#{$pf-global}--BorderColor--200);
+  --#{$form-control}--after--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
   --#{$form-control}--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
   --#{$form-control}--Width: 100%;
-  --#{$form-control}--Height: calc(var(--#{$form-control}--FontSize) * var(--#{$form-control}--LineHeight) + var(--#{$form-control}--BorderWidth) * 2 + var(--#{$form-control}--PaddingTop) + var(--#{$form-control}--PaddingBottom)); // Needed for various browsers that compute height of form elements differently
 
   // padding
-  // Note for top/bottom padding, subtracting the border height so the height of the input is consistent with buttons and dropdowns that use the same spacing but draw their borders with pseudo elements
   --#{$form-control}--inset--base: var(--#{$pf-global}--spacer--sm);
-  --#{$form-control}--PaddingTop: calc(var(--#{$pf-global}--spacer--form-element) - var(--#{$pf-global}--BorderWidth--sm));
-  --#{$form-control}--PaddingBottom: calc(var(--#{$pf-global}--spacer--form-element) - var(--#{$pf-global}--BorderWidth--sm));
+  --#{$form-control}--PaddingTop: var(--#{$pf-global}--spacer--form-element);
+  --#{$form-control}--PaddingBottom: var(--#{$pf-global}--spacer--form-element);
   --#{$form-control}--PaddingRight: var(--#{$form-control}--inset--base);
   --#{$form-control}--PaddingLeft: var(--#{$form-control}--inset--base);
 
   // hover
-  --#{$form-control}--hover--BorderBottomColor: var(--#{$pf-global}--primary-color--100);
+  --#{$form-control}--hover--after--BorderBottomColor: var(--#{$pf-global}--primary-color--100);
 
   // focus
-  --#{$form-control}--focus--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--md);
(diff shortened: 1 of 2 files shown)


thatblindgeye (approved):
🎉

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -1,88 +1,81 @@
...
-  --#{$form-control}--BorderTopColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderRightColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderBottomColor: var(--#{$pf-global}--BorderColor--200);
-  --#{$form-control}--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderRadius: 0;
+  --#{$form-control}--after--BorderTopWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderRightWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderLeftWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
```
thatblindgeye, line 11: One small thing, doesn't look like this `after--BorderWidth` var is being used anywhere

srambach:
FWIW, the focus ring is narrower than it was previously - but now you can see the status color bottom border. It also removes the little artifact at the corner, which is nice.
image

srambach:
I think on line 90, `--#{$form-control}--textarea--Height` can also be removed?

Similar to the previous comment, the focus ring is also cut off when next to something, like in the date picker. 
image

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -161,65 +160,51 @@
...
       --#{$form-control}--inset--base: var(--#{$form-control}--m-readonly--m-plain--inset--base);
-
-      border-color: transparent;
+      --#{$form-control}--after--BorderStyle: none;
     }
   }
 
   &:hover {
-    --#{$form-control}--BorderBottomColor: var(--#{$form-control}--hover--BorderBottomColor);
+    --#{$form-control}--after--after--BorderBottomColor: var(--#{$form-control}--hover--after--BorderBottomColor);
```
srambach, line 180: Search and replace oops? (--after--after)

srambach (commented):
Generally this looks good but for a few nits and the focus ring question. 
Also, just for discussion - using `--after` is consistent with our model, but when I was refactoring the cards we talked about using more logical names to keep future things non-breaking and perhaps ease token use. E.g., if we kept the border variables as is, we could move them between the wrapper or the pseudo-element or whatever clever thing we came up with, without causing a breaking change. WDYT?

mcoker:
@srambach thanks for the review! Should have addressed everything but the focus ring. Dark theme vars also needed to be updated, moved the `:focus` styles to `:focus-within`, and fixed focus styles appearing incorrectly in the readonly input.

Re: the focus ring... a few thoughts. I'll note that from glancing, the focus ring looks the same in menu toggle and control buttons, which have the same look.
* The lack of focus ring date picker is from the input group. The input group moves any `* + *` child to the left by 1px, to prevent double-borders when putting form-things side by side. So if you focus on anything that has a sibling after it in the input group layout, the sibling component will be on top of the focused component. We had this style previously to address that, we could add it back. wdyt? cc @mattnolting 
 
  https://github.com/patternfly/patternfly/blob/7ddc2de5cfb96f204e62730878a9e77bed64b998/src/patternfly/components/InputGroup/input-group.scss#L43-L46
* One solution could be to adjust the `outline-offset` - that would move the focus ring inside - it would need to be inside now because the textarea controls overflow with `overflow: auto;` and that will hide the outline if it goes outside of the box.
* An easier solution could be to manage the stacking context:
  * move the border to `::before` and add `position: relative` or `isolation: isolate` to the input/textarea/select
  * keep the borders on `::after` and add our xs z-index on the form element. I'd probably prefer this as a solution - ideally I think the grey borders for the form control should be on the `::before` and the status/bottom border should be on `::after` so they can be drawn separately and not create an angled corner like they do now. This is how the menu toggle is setup currently.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -1,88 +1,80 @@
...
-  --#{$form-control}--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
-  --#{$form-control}--BorderRadius: 0;
+  --#{$form-control}--after--BorderTopWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderRightWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderLeftWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$form-control}--after--BorderTopColor: var(--#{$pf-global}--BorderColor--300);
+  --#{$form-control}--after--BorderRightColor: var(--#{$pf-global}--BorderColor--300);
+  --#{$form-control}--after--BorderBottomColor: var(--#{$pf-global}--BorderColor--200);
+  --#{$form-control}--after--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
```
mattnolting, line 14:
```suggestion
  --#{$form-control}--before--BorderTopWidth: var(--#{$pf-global}--BorderWidth--sm);
  --#{$form-control}--before--BorderRightWidth: var(--#{$pf-global}--BorderWidth--sm);
  --#{$form-control}--before--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
  --#{$form-control}--before--BorderLeftWidth: var(--#{$pf-global}--BorderWidth--sm);
  --#{$form-control}--before--BorderTopColor: var(--#{$pf-global}--BorderColor--300);
  --#{$form-control}--before--BorderRightColor: var(--#{$pf-global}--BorderColor--300);
  --#{$form-control}--before--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
  --#{$form-control}--before--BorderLeftColor: var(--#{$pf-global}--BorderColor--300);
  
  --#{$form-control}--after--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
  --#{$form-control}--after--BorderBottomColor: var(--#{$pf-global}--BorderColor--200);
```

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -130,9 +122,16 @@
   line-height: var(--#{$form-control}--LineHeight);
   resize: var(--#{$form-control}--Resize);
   background-color: var(--#{$form-control}--BackgroundColor);
-  border: var(--#{$form-control}--BorderWidth) solid;
-  border-color: var(--#{$form-control}--BorderTopColor) var(--#{$form-control}--BorderRightColor) var(--#{$form-control}--BorderBottomColor) var(--#{$form-control}--BorderLeftColor);
-  border-radius: var(--#{$form-control}--BorderRadius);
+
+  &::after {
```
mattnolting, line 126:
```suggestion
  &::before,
  &::after {
```

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -130,9 +122,16 @@
...
+    position: absolute;
+    inset: 0;
+    pointer-events: none;
+    content: "";
+    border-color: var(--#{$form-control}--after--BorderTopColor) var(--#{$form-control}--after--BorderRightColor) var(--#{$form-control}--after--BorderBottomColor) var(--#{$form-control}--after--BorderLeftColor);
+    border-style: var(--#{$form-control}--after--BorderStyle, solid);
+    border-width: var(--#{$form-control}--after--BorderTopWidth) var(--#{$form-control}--after--BorderRightWidth) var(--#{$form-control}--after--BorderBottomWidth) var(--#{$form-control}--after--BorderLeftWidth);
```
mattnolting, line 133: Then split these between `::before, ::after`

mattnolting (changes requested):
I think moving the [borders to a pseudo](https://patternfly-pr-5641.surge.sh/components/forms/form-control#input) is a good idea, but typically we add gray, static borders to `::before` and accent/hover borders to `::after` to avoid bevelled borders

Screenshot 2023-06-01 at 2 43 03 PM



Screenshot 2023-06-01 at 2 43 03 PM

srambach:
I don't hate moving the outline-offset in about 2px. That gives the full focus ring, the width is consistent with other focus rings other than the ones you mention (which then could be updated as well), and still shows the status bottom border color. Do you know of any problems it creates?

srambach:
If the text area is resized, the focus ring isn't stretching properly. 
image

srambach (approved):
👍

mattnolting (approved):
Thanks for updating, LPTM!!

Issue #5622, opened 2023-05-25
Bug - Switch - Look of disabled switch is not quite right in dark mode


mcarrano:
**Describe the problem**
When the switch is disabled, we are mixing dark and light elements. Seems like the text and toggle should both either be light or dark.

<img width="282" alt="Screenshot 2023-05-25 at 3 09 33 PM" src="https://github.com/patternfly/patternfly/assets/6619860/cc8198fc-5e6a-4fe3-901d-e75d4babac88">

Let's add disabled switches with icons to the disabled example block in core to match the screenshot from the react workspace.

mcarrano:
Thoughts @mceledonia ?

PR #5651, opened 2023-06-07
fix(core): darkened circle icon to match dark theme

ashley-o0o:
Adjusted the disabled switch in dark mode so that the colors match the other dark mode switches. Also added extra switches to include option with check.
Screenshot 2023-06-07 at 10 30 48 AM

Closes https://github.com/patternfly/patternfly/issues/5622

### DIFF
--- a/src/patternfly/components/Switch/examples/Switch.md
+++ b/src/patternfly/components/Switch/examples/Switch.md
@@ -105,7 +105,7 @@ cssPrefix: pf-v5-c-switch
 {{#> switch switch--attribute='for="switch-with-icon-disabled-1"'}}
   {{#> switch-input switch-input--id="switch-with-icon-disabled-1" switch-input--attribute='disabled checked'}}{{/switch-input}}
   {{#> switch-toggle}}
-    {{#> switch-toggle-icon-disabled}}{{/switch-toggle-icon-disabled}}
+    {{#> switch-toggle-icon}}{{/switch-toggle-icon}}
   {{/switch-toggle}}
 {{/switch}}
 <br/>
@@ -114,7 +114,7 @@ cssPrefix: pf-v5-c-switch
   {{#> switch-input switch-input--id="switch-with-icon-disabled-2" switch-input--attribute='disabled'}}
   {{/switch-input}}
   {{#> switch-toggle}}
-    {{#> switch-toggle-icon-disabled}}{{/switch-toggle-icon-disabled}}
+    {{#> switch-toggle-icon}}{{/switch-toggle-icon}}
   {{/switch-toggle}}
 {{/switch}}
 ```
--- a/src/patternfly/components/Switch/examples/Switch.md
+++ b/src/patternfly/components/Switch/examples/Switch.md
@@ -108,7 +108,7 @@ cssPrefix: pf-v5-c-switch
 {{#> switch switch--attribute='for="switch-with-icon-disabled-1"'}}
   {{#> switch-input switch-input--id="switch-with-icon-1" switch-input--attribute='disabled checked'}}{{/switch-input}}
   {{#> switch-toggle}}
-    {{#> switch-toggle-icon}}{{/switch-toggle-icon}}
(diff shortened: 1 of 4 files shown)


ashley-o0o:
What: Closes [#5622 ](https://github.com/patternfly/patternfly/issues/5622)

src/patternfly/components/Switch/examples/Switch.md
```diff
@@ -100,6 +100,26 @@ cssPrefix: pf-v5-c-switch
   {{#> switch-label switch-label--id="switch-disabled-2-on" switch-label--modifier="pf-m-on" switch-label--attribute='aria-hidden="true"'}}Message when on{{/switch-label}}
   {{#> switch-label switch-label--id="switch-disabled-2-off" switch-label--modifier="pf-m-off" switch-label--attribute='aria-hidden="true"'}}Message when off{{/switch-label}}
 {{/switch}}
+
+
+
+<br/>
+<br/>
+{{#> switch switch--attribute='for="switch-with-icon-disabled-1"'}}
+  {{#> switch-input switch-input--id="switch-with-icon-1" switch-input--attribute='disabled checked'}}{{/switch-input}}
```
mcoker, line 109: The value here needs to match the value in the `for` attribute above.

```suggestion
  {{#> switch-input switch-input--id="switch-with-icon-disabled-1" switch-input--attribute='disabled checked'}}{{/switch-input}}
```

src/patternfly/components/Switch/examples/Switch.md
```diff
@@ -100,6 +100,26 @@ cssPrefix: pf-v5-c-switch
...
+  {{#> switch-toggle}}
+    {{#> switch-toggle-icon}}{{/switch-toggle-icon}}
+  {{/switch-toggle}}
+{{/switch}}
+<br/>
+<br/>
+{{#> switch switch--attribute='for="switch-with-icon-disabled-2"'}}
+  {{#> switch-input switch-input--id="switch-with-icon-2" switch-input--attribute='disabled'}}
```
mcoker, line 117: Same here - this needs to match the `for` attribute above

```suggestion
  {{#> switch-input switch-input--id="switch-with-icon-disabled-2"" switch-input--attribute='disabled'}}
```

mcoker (changes requested):
Good job on this! This matches the check and the circle, but like we discussed offline, it's probably best if we retain the disabled color of the circle and update the check to match.

Because we're looking to get this in ASAP, the way I would probably go about this would be to:
* Declare a new variable at the end of this list for the check icon that we will use to manipulate the color of the check icon when the switch is in the disabled state, and give it the same value as the disabled circle color  https://github.com/patternfly/patternfly/blob/2f153a8479371d82b6e547b0e66ff322333bad02/src/patternfly/components/Switch/switch.scss#L9-L13
  * new variable - `--#{$switch}__input--disabled__toggle-icon--Color: var(--#{$pf-global}--palette--black-150);`
* In this disabled CSS block, in between the switch toggle selector and `cursor` declaration, add a new declaration for the check icon color to be defined by the disabled icon variable we defined in the previous step (Just a nit - you probably need an empty line after the new variable and the `cursor` declaration) https://github.com/patternfly/patternfly/blob/2f153a8479371d82b6e547b0e66ff322333bad02/src/patternfly/components/Switch/switch.scss#L123-L124 
  * new declaration - `--#{$switch}__toggle-icon--Color: var(--#{$switch}__input--disabled__toggle-icon--Color);`
* In the dark theme stylesheet, add a new declaration for the disabled check icon color that matches the disabled circle color https://github.com/patternfly/patternfly/blob/2f153a8479371d82b6e547b0e66ff322333bad02/src/patternfly/components/Switch/themes/dark/switch.scss#L4-L9
  * new declaration - `--#{$switch}__input--disabled__toggle-icon--Color: var(--#{$pf-global}--disabled-color--100);`

mcoker (approved):
LGTM!

mceledonia (approved):

mattnolting (approved):
LPTM 👍

Issue #5649, opened 2023-06-06
Drawer, nav - update to examples


mcoker:
Follow ups from react issues

Drawer - https://github.com/patternfly/patternfly-react/issues/9190
* Change "light-200" in example title to "secondary" in https://pf5.patternfly.org/components/drawer/#panel-with-light-200-background
* Update the docs from "Modifies the drawer body/panel background color to light grey." to something like "Modifies the body/panel to use the secondary background color."

Nav - https://github.com/patternfly/patternfly-react/pull/9223
* Mark any of the light examples as deprecated
  * https://pf5.patternfly.org/components/navigation#default-in-light-mode
  * https://pf5.patternfly.org/components/navigation#expanded-in-light-mode
  * https://pf5.patternfly.org/components/navigation/html-demos#light-theme-sidebar-and-nav

PR #5657, opened 2023-06-08
fix: updated some text, and added deprecated tags where needed

ashley-o0o:
Changed "light-200" in text to "secondary." Also added deprecated tags where needed.
closes: [#5649](https://github.com/patternfly/patternfly/issues/5649)


mcoker (approved):
🥳👍

Issue #5653, opened 2023-06-08
Bug - select - Accessibility text incorrect


phill-85:
**Describe the problem**
`<span id="select-checkbox-counts-label" hidden>Choose one</span>`  in the example (as well as other examples on that page) at https://pf5.patternfly.org/components/menus/select#checkbox-with-counts is singular, but the component has the ability for multiple choices to be selected.

**How do you reproduce the problem?**
See example code for the link above

**Expected behavior**
I'm no expert in accessibility,  but it would seem that it should it be something like "Choose one or more" or "Check multiple"?

**Is this issue blocking you?**
No

**What is your product and what release date are you targeting?**
Which ever vesion of the spec(?) is current at pf5.patternfly.org

**Any other information?**
![image](https://github.com/patternfly/patternfly/assets/4619659/3ac1ed61-3f89-4c9a-9e01-9a47a4371fb2)

PR #5663, opened 2023-06-12
chore(select): fixed text in HTML example to show multiple checks can be selected

ashley-o0o:
Fixed some of the hidden labels in the HTML examples on Select page. Some were saying to choose one when multiple could be selected. Adjusted to reflect that.

Closes: [#5653](https://github.com/patternfly/patternfly/issues/5653)

### DIFF
--- a/src/patternfly/components/Button/examples/Button.md
+++ b/src/patternfly/components/Button/examples/Button.md
@@ -403,6 +403,7 @@ Semantic buttons and links are important for usability as well as accessibility.
 | `.pf-m-secondary` | `.pf-v5-c-button` | Modifies for secondary styles. |
 | `.pf-m-tertiary` | `.pf-v5-c-button` | Modifies for tertiary styles. |
 | `.pf-m-danger` | `.pf-v5-c-button` | Modifies for danger styles. |
+| `.pf-m-warning` | `.pf-v5-c-button` | Modifies for warning styles. |
 | `.pf-m-link` | `.pf-v5-c-button` | Modifies for link styles. This button has no background or border and is styled as a link. This button would commonly appear in a form and may include an icon. |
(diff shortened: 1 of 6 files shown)


srambach:
Nice fix! ⭐ I see one inconsistency that didn't get covered, but I'm not sure if it's really supposed to be multiple:
This example shows multiple items selected, but uses "Choose one" - I'm not sure whether this should have the label fixed or remove the multiple selections, since the next example shows multiple selection with checkboxes. https://patternfly-pr-5663.surge.sh/components/menus/select#item-description

srambach:
This is great - one thing though I think it might be better to make the one that is selected to be one of the options that has a description - that way it demonstrates the alignment with the description. What do you think?

mcoker:
@ashley-o0o +1 to @srambach's comment, I think showing how the checkmark aligns with an item with a description would be a nice add for this example.

mcoker:
@ashley-o0o looks like you've got a couple of conflicts. In https://github.com/patternfly/patternfly/pull/5667, the `src/patternfly/components/Select/examples/*` files were moved to `src/patternfly/components/Select/deprecated/`, so your changes to those files are conflicting. Let me know if you need help resolving that 👍

thatblindgeye (changes requested):
This is looking nice! Just had a couple other examples that'd be worth updating a bit further:

- [Single with typeahead expanded](https://patternfly-pr-5663.surge.sh/components/menus/select#single-with-typeahead-expanded) has the hidden text as "New", which makes the listbox announced as "New listbox..." rather than "Choose a state listbox...".
- "Single with typeahead expanded and selected" and "Multi with typeahead (chip group collapsed)" examples have a similar issue, where the text of the input is being used as the hidden label text

Something else I'm noticing that I can open a follow-up for is adding `aria-selected="false"` to any items that aren't selected, as well as updating the documentation wherever `aria-selected="true"` is documented. With VoiceOver right now even non-selected items are announced as "[item text], selected".

thatblindgeye (approved):
I opened the above-mentioned followup issue

mcoker (approved):

Issue #5672, opened 2023-06-16
Bug - Charts - dark theme chart stylesheet vars have wrong prefix


mcoker:
Dark theme chart vars are being compiled with a `pf-v5-c-chart` prefix and it should be `pf-v5-chart`

PR #5673, opened 2023-06-16
fix(charts): fixed dark theme chart var prefix

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5672

### DIFF
--- a/src/patternfly/base/themes/dark/_chart-globals.scss
+++ b/src/patternfly/base/themes/dark/_chart-globals.scss
@@ -1,7 +1,7 @@
 @import "../../../sass-utilities/colors";
 @import "../../../sass-utilities/scss-variables";
 // stylelint-disable scss/dollar-variable-pattern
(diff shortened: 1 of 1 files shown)


evwilkin (approved):
LGTM

srambach:
It looks like PF4 uses `pf-c-chart` so why wouldn't we want `pf-v5-c-chart`?

mcoker:
@srambach where are you seeing `pf-c-chart`? This is what I'm seeing on the `v4` branch:

* CSS vars - https://github.com/patternfly/patternfly/blob/v4/src/patternfly/patternfly-charts.scss
* SCSS vars - https://github.com/patternfly/patternfly/blob/v4/src/patternfly/base/_chart-globals.scss
* dark SCSS vars - https://github.com/patternfly/patternfly/blob/v4/src/patternfly/base/themes/dark/_chart-globals.scss

srambach (approved):
Let's leave it matching how it was in PF4 - looks good.

Issue #5677, opened 2023-06-20
Bug - RHD bold is italic and shouldn't be


mcoker:
RHD + bold font weight is italic and shouldn't be 
<img width="1194" alt="Screenshot 2023-06-20 at 11 12 45 AM" src="https://github.com/patternfly/patternfly/assets/35148959/70d7d42c-0193-4a48-807c-83f42ceeefab">

PR #5678, opened 2023-06-20
fix(font): fixed RHD bold incorrect font mapping to italic font

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5677


mattnolting (approved):
LPTM 👍

PR #5680, opened 2023-06-21
fix(dark-theme-placeholders): @mcoker fixed the placeholders

### LINKED ISSUE #5618: Duplicate pf-t-dark placeholder styles
dupes - https://unpkg.com/browse/@patternfly/patternfly@5.0.0-alpha.64/base/patternfly-themes.css

This file is generating duplicate `.pf-t-light/dark` blocks - https://github.com/patternfly/patternfly/blob/v5/src/patternfly/base/patternfly-themes.scss

This is due to the way the build/sass compilation process imports our global SCSS vars here - https://github.com/patternfly/patternfly/blob/be98c51370b03d8823587a44c79c0c7a94331bec/scripts/gulp/sass.mjs#L28

That import is ignored for `src/patternfly/patternfly[-*].scss` files, since those files import their own global SCSS. We also have stylesheets that import their own global SCSS in `src/patternfly/base/patternfly-*.scss`, so the compile script should probably be updated to ignore those as well?
https://github.com/patternfly/patternfly/blob/be98c51370b03d8823587a44c79c0c7a94331bec/src/patternfly/base/patternfly-themes.scss#L1

----

dupes - https://unpkg.com/browse/@patternfly/patternfly@5.0.0-prerelease.4/patternfly.css

If you search for something like `:where(.pf-v5-theme-dark) .pf-v5-c-page__main-section[class*=pf-m-dark-]`, you'll see the dupes. This is because each dark theme stylesheet imports [dark theme SCSS](https://github.com/patternfly/patternfly/blob/8aa654bdc03f6746c2d67b616bda04349d84e74d/src/patternfly/components/Alert/themes/dark/alert.scss#L1), and this [placeholders block](https://github.com/patternfly/patternfly/blob/main/src/patternfly/sass-utilities/themes/dark/placeholders.scss) is getting duplicated. This was pre-existing and uncaught in v4.

mattnolting:
fixes #5618

### DIFF
--- a/src/patternfly/sass-utilities/scss-variables.scss
+++ b/src/patternfly/sass-utilities/scss-variables.scss
@@ -1,4 +1,5 @@
 @import './init'; // initialize versioning
+$pf-v5-global--placeholders-loaded: false;
 
 // This file is for initial variable values.
 // All variables should have a valid css value,
--- a/src/patternfly/patternfly-base.scss
+++ b/src/patternfly/patternfly-base.scss
@@ -4,3 +4,5 @@
 @import "base/base";
 @import "base/themes";
 @import "base/icons";
+
+$pf-v5-global--placeholders-loaded: true;
--- a/src/patternfly/sass-utilities/themes/dark/placeholders.scss
+++ b/src/patternfly/sass-utilities/themes/dark/placeholders.scss
@@ -1,5 +1,7 @@
-%pf-v5-theme-dark--t-dark {
-  .#{$button} {
-    --#{$button}--m-primary--BackgroundColor: var(--#{$pf-global}--primary-color--300);
+@if not $pf-v5-global--placeholders-loaded {
+  %pf-v5-theme-dark--t-dark {
+    .#{$button} {
+      --#{$button}--m-primary--BackgroundColor: var(--#{$pf-global}--primary-color--300);
+    }
   }
 }


src/patternfly/sass-utilities/scss-variables.scss
```diff
@@ -1,4 +1,5 @@
 @import './init'; // initialize versioning
+$pf-v5-global--placeholders-loaded: false;
```
mcoker, line 2: Thinking about this a little more, wdyt about localizing everything just to the placeholders file? Seems a little odd for this to be with the rest of our other global vars. I dropped this in place, and it seems to work. 

```diff
diff --git a/src/patternfly/sass-utilities/themes/dark/placeholders.scss b/src/patternfly/sass-utilities/themes/dark/placeholders.scss
index 1d3f1307..10ee60de 100644
--- a/src/patternfly/sass-utilities/themes/dark/placeholders.scss
+++ b/src/patternfly/sass-utilities/themes/dark/placeholders.scss
@@ -1,5 +1,12 @@
-%pf-v5-theme-dark--t-dark {
-  .#{$button} {
-    --#{$button}--m-primary--BackgroundColor: var(--#{$pf-global}--primary-color--300);
+$pf-v5-global--placeholder-loaded: false !default;
+
+// Prevents loading this file multiple times in a single compilation
+@if not $pf-v5-global--placeholder-loaded {
+  %pf-v5-theme-dark--t-dark {
+    .#{$button} {
+      --#{$button}--m-primary--BackgroundColor: var(--#{$pf-global}--primary-color--300);
+    }
   }
 }
+
+$pf-v5-global--placeholder-loaded: true;
```

Issue #5613, opened 2023-05-24
Bug - Brand - Brand image used in examples cannot be seen in dark theme


mcarrano:
**Describe the problem**
The PatternFly Brand placeholder used for the Brand examples and some card examples is not visible in dark theme. 

<img width="862" alt="Screenshot 2023-05-24 at 3 56 00 PM" src="https://github.com/patternfly/patternfly/assets/6619860/d5f69463-dcbd-4f8a-a5ba-cfe2b969fcd4">

<img width="834" alt="Screenshot 2023-05-24 at 3 58 21 PM" src="https://github.com/patternfly/patternfly/assets/6619860/ff8a615b-a597-4290-83fc-7038c1d7c642">

Is it possible to apply styles to this to update the colors of the SVG when viewed in dark mode? We could also potentially replace this with a graphic that works in either theme.

cc @mceledonia @andrew-ronaldson

PR #5681, opened 2023-06-21
fix(brand+card): update examples to display in dark theme

jenny-s51:
Towards #5613


mcoker (commented):
Thanks @jenny-s51! My initial thought is that I'm not sure if we want to go down the road of introducing an enhancement to individual components to support toggling between the base and dark themes. Initially I was thinking we would add local workspace styles to the examples to toggle any images we want to swap between themes, by targeting the image or whatever it is using the dark theme class selector. But I'm curious what others think. @srambach @mattnolting 

```
.some-image--dark, // hide dark image by default
.pf-v5-theme-dark .some-image { // hide light image when dark theme is set
  display: none;
}

.pf-v5-theme-dark .some-image--dark { // show dark image when dark theme is set
  display: revert;
}
```

jenny-s51:
Thank you @mcoker for your feedback on this - it sounds like we want to take a different approach here. Just want to confirm... does [this change](https://github.com/patternfly/patternfly/issues/5801) need to go in before we can make any subsequent updates to this PR?

mcoker:
@jenny-s51 I don't think so, we could just update the examples that use brand to use the local style method described in https://github.com/patternfly/patternfly/pull/5681#pullrequestreview-1522981635 and we can follow up with the brand enhancement later and yank out the local styles.

jenny-s51 (commented):
@mattnolting @mcoker thanks again for all your great feedback on this! The dark/light switching behavior now seems to be working as expected. If this PR looks good to you guys, we may be able to just merge this in and close https://github.com/patternfly/patternfly/issues/5613 since the original issue has been addressed. 

Were there any other changes you either of you wanted to see added to this PR?

PR #5693, opened 2023-06-27
Penta theming/token hackathon

srambach:
Preview link - https://patternfly-pr-5693.surge.sh/patterns/hackathon/html-demos/hackathon-demo-app/


andrew-ronaldson (commented):
Minor nits and more questions for @mceledonia and @lboehling 
In the mocks of the table there is a spacer between the icons and the counts. I don't see anything in the CSS so do we need to add something? 
Screenshot 2023-06-29 at 2 32 19 PM

The other small one is the hover state of the toolbar icons. The mock ups have the spacer even around the icon but the demo seems to be more narrow on the left & right padding. Is that just because of the line height adding some extra height?

Screenshot 2023-06-29 at 2 32 31 PM

mceledonia:
Yeah I noticed these as well as a few other small nits I had Coker work out afterwards. For an ideal representation:
- The icon buttons could have either a sm or xs spacer for left/right padding (if they already have one, bump it up to the next)
- The icon/counts could have a sm spacer between them
- Another thing I noticed is the dark theme border hover color was specced incorrectly and is one step darker than it should be, it should cascade like...

`—pf-global—border—color—control—hover` > `—pf-global—color—brand—200` > `$pf-color-blue-300`

^ Sorry about that one! It was an error on our (my) side when writing the spec.

lboehling:
One thing i noticed after looking at the demo again is that there is no padding to the left of the page section when the nav is closed, I think this is a flaw in how we spec-d out the designs for page component. Can we remove the right 24px padding from the side bar and add the 24px padding the the left of the page instead so that there is still padding to the left of the page sections when the nav is closed? Open to other solutions too if there's a better way to get that padding to remain when nav is closed! 
Screenshot 2023-07-11 at 2 51 18 PM

thatblindgeye:
@andrew-ronaldson @lboehling @mceledonia pushed updates based on above feedback. Can you confirm if things are in a state that you were expecting?

andrew-ronaldson (approved):
I think this is looking good. @lboehling @mceledonia any thoughts

lboehling (approved):
looks amazing!!! 🤩

mcoker:
Successfully pulled into react and stood up in a demo via https://github.com/patternfly/patternfly-react/issues/9357. Great work on this everyone!!! 🥳🚀

PR #5696, opened 2023-06-29
feat(Button): extended inline styling

### LINKED ISSUE #5351: Button - expand "inline" support to other variants
The problem - we have super small buttons across components that are either
* not using the PF button component
* are using the PF button component but the padding/size is customized in each component

Some examples:
* [form help popover button](https://www.patternfly.org/v4/components/form#basic) - doesn't use PF button
* [label close button](https://www.patternfly.org/v4/components/label#filled) - uses a modified PF plain button
* [text input group close/utility button(s)](https://www.patternfly.org/v4/components/text-input-group/html#utilities-and-icon-with-placeholder-text) - uses a modified PF plain button
* [tabs help/close actions](https://www.patternfly.org/v4/components/tabs#help-and-close-actions) - uses a modified PF plain button

Looking at the [WCAG spec on target size](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html), there are cases for buttons that do not have minimum target sizes with a specific callout for help icons or icons that flow with sentences and could wrap with the text and throw off the size/spacing of the text around it.

Right now, `.pf-m-inline` only applies to "link" buttons, and it basically makes a link button look like a link/anchor. I don't know if we would try and expand that variation, or possibly create a new variant, I think at a minimum, there is a case for an inline plain button where the font-size is inherited and either padding is removed or minimal padding is added so that it doesn't exceed the line-height of adjacent text of the same size.

cc @srambach @mceledonia

thatblindgeye:
Closes #5351 

@mceledonia @mcoker This draft currently only extends inline support to the plain button. I did try seeing what similar styling might look like for other button variants, for example the primary variant:

Inline primary example

One thing to note is that with the Tabs component's action buttons (close and help), the icon becomes a little bigger than it currently is (can be fixed by adding a font size CSS var for pf-v5-c-tabs__item-action), and the outline on focus gets a little overlapped on the top and bottom:

Tabs action button outline

 We could use the current outline offset in the Tabs css (-0.1875rem), though that might make it feel cramped using `--pf-v5-global--FontSize--xs` (worse if the font size is custom and smaller than this):

Tabs action buttons outline with offset

This is also assuming we'd want to move the button styling out of the Tabs scss. If not, then it's a bit simpler to keep the `pf-v5-c-tabs__item-action .pf-v5-c-button` block and add the outline offset there.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -144,12 +144,6 @@
   --#{$button}--m-plain--disabled--Color: var(--#{$pf-global}--disabled-color--200);
   --#{$button}--m-plain--disabled--BackgroundColor: transparent;
 
-    // Plain compact button
-    --#{$button}--m-plain--m-compact--PaddingTop: 0;
-    --#{$button}--m-plain--m-compact--PaddingRight: 0;
-    --#{$button}--m-plain--m-compact--PaddingBottom: 0;
-    --#{$button}--m-plain--m-compact--PaddingLeft: 0;
-
   // Control Button
   --#{$button}--m-control--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
   --#{$button}--m-control--Color: var(--#{$pf-global}--Color--100);
@@ -565,13 +559,8 @@
     color: var(--#{$button}--m-plain--Color);
     background-color: var(--#{$button}--m-plain--BackgroundColor);
 
-    &.pf-m-compact {
-      --#{$button}--PaddingTop: var(--#{$button}--m-plain--m-compact--PaddingTop);
-      --#{$button}--PaddingRight: var(--#{$button}--m-plain--m-compact--PaddingRight);
-      --#{$button}--PaddingBottom: var(--#{$button}--m-plain--m-compact--PaddingBottom);
-      --#{$button}--PaddingLeft: var(--#{$button}--m-plain--m-compact--PaddingLeft);
-  
-      line-height: 1;
+    &.pf-m-no-padding {
+      padding: 0;
     }
 
     &:hover {
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -143,12 +143,12 @@
   --#{$button}--m-plain--active--Color: var(--#{$pf-global}--Color--100);
   --#{$button}--m-plain--disabled--Color: var(--#{$pf-global}--disabled-color--200);
   --#{$button}--m-plain--disabled--BackgroundColor: transparent;
-  --#{$button}--m-plain--m-inline--FontSize: inherit;
-  --#{$button}--m-plain--m-inline--PaddingTop: 0;
-  --#{$button}--m-plain--m-inline--PaddingRight: var(--#{$pf-global}--spacer--sm);;
-  --#{$button}--m-plain--m-inline--PaddingBottom: 0;
-  --#{$button}--m-plain--m-inline--PaddingLeft: var(--#{$pf-global}--spacer--sm);;
-  --#{$button}--m-plain--m-inline--LineHeight: normal;
+
+    // Plain compact button
+    --#{$button}--m-plain--m-compact--PaddingTop: 0;
+    --#{$button}--m-plain--m-compact--PaddingRight: 0;
+    --#{$button}--m-plain--m-compact--PaddingBottom: 0;
+    --#{$button}--m-plain--m-compact--PaddingLeft: 0;
 
   // Control Button
   --#{$button}--m-control--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
@@ -565,13 +565,13 @@
     color: var(--#{$button}--m-plain--Color);
     background-color: var(--#{$button}--m-plain--BackgroundColor);
 
-    &.pf-m-inline {
-      padding-block-start: var(--#{$button}--m-plain--m-inline--PaddingTop);
-      padding-block-end: var(--#{$button}--m-plain--m-inline--PaddingBottom);
-      padding-inline-start: var(--#{$button}--m-plain--m-inline--PaddingLeft);
-      padding-inline-end: var(--#{$button}--m-plain--m-inline--PaddingRight);
-      font-size: var(--#{$button}--m-plain--m-inline--FontSize);
-      line-height: var(--#{$button}--m-plain--m-inline--LineHeight);
+    &.pf-m-compact {
+      --#{$button}--PaddingTop: var(--#{$button}--m-plain--m-compact--PaddingTop);
+      --#{$button}--PaddingRight: var(--#{$button}--m-plain--m-compact--PaddingRight);
+      --#{$button}--PaddingBottom: var(--#{$button}--m-plain--m-compact--PaddingBottom);
+      --#{$button}--PaddingLeft: var(--#{$button}--m-plain--m-compact--PaddingLeft);
+  
+      line-height: 1;
     }
 
     &:hover {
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -553,6 +559,12 @@
     color: var(--#{$button}--m-plain--Color);
     background-color: var(--#{$button}--m-plain--BackgroundColor);
 
+    &.pf-m-inline {
+      padding: var(--#{$button}--m-plain--m-inline--PaddingTop) var(--#{$button}--m-plain--m-inline--PaddingRight) var(--#{$button}--m-plain--m-inline--PaddingBottom) var(--#{$button}--m-plain--m-inline--PaddingLeft);
```
srambach, line 563: Do you want to go ahead and make this logical properties?

thatblindgeye: Do we plan on updating CSS vars similarly at any point, e.g. instead of `--component--PaddingTop` => `--component--PaddingBlockStart` or something along those lines? Probably makes more sense to keep the `PaddingTop` etc var names to be less confusing, but curious if that's on the roadmap in regards to RTL.

srambach: @coker and I talked about that - we were thinking to keep the variables as is because that would be a sweeping breaking change.

src/patternfly/components/Button/button.scss
```diff
@@ -143,6 +143,12 @@
   --#{$button}--m-plain--active--Color: var(--#{$pf-global}--Color--100);
   --#{$button}--m-plain--disabled--Color: var(--#{$pf-global}--disabled-color--200);
   --#{$button}--m-plain--disabled--BackgroundColor: transparent;
+  --#{$button}--m-plain--m-inline--FontSize: inherit;
```
mcoker, line 146: `inherit` (and `unset` and `initial`) can't be used as the value of a CSS var, and have that value come out of the `var()` function that refers to it. You'll need to use `inherit` as the fallback to a CSS var in the `var()` function. From there you can trigger the fallback by setting the var's value to `initial`. Technically `inherit` will trigger the fallback if one is set, too, but `initial` is the designated value to always trigger the fallback. 
You can read about that here - https://www.w3.org/TR/css-variables-1/#example-ee865e1b

Your PR is working because in `font-size: var(--#{$button}--m-plain--m-inline--FontSize);`, the custom property in the `var()` function is invalid, so `var()` is returning `initial` as the value... and the `initial` value for `font-size` is `inherit`.

mcoker (commented):
Nice! I left some comments, but I wonder if we should look at this overall. I think there are probably 2 use cases here:
1) Use for a small/compact plain button outlined in https://github.com/patternfly/patternfly/issues/5351
2) An "inline" button that adapts its size to the context where it's used. 

I think this is more along the lines of an inline button, but the left/right padding set to "sm" doesn't seem ideal as the text scales. If we wanted the padding to scale, it could be based off of the font-size (`1em`). Though if we followed the non-inline plain button as a guide for that padding, its padding is `16px`, so the left/right would just be `1em`, which is too big as the button icon size scales down.

There is a comment in the original issue around discussing a product need for this https://github.com/patternfly/patternfly/issues/5351#issuecomment-1493133772

@nicolethoen do you recall that conversation? I looked in the office hours agenda before that comment and skimmed a couple of the recordings and didn't see that discussion.

src/patternfly/components/Button/button.scss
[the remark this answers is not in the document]
thatblindgeye: Ah interesting! 🤯  The codepen definitely helps show it

thatblindgeye:
> Use for a small/compact plain button

@mcoker that's something I had thought about, but it seemed less of an "inline" button and more of a "small" variant of a plain button. Maybe we offer both a plain+inline button as well as a small/compact+plain button variant, with the latter applying the smaller font size as well as a smaller padding? Throwing something together locally, here's what that could look like (keeping the same `var(--#{$button}--m-small--FontSize` font-size applied, then just updating the paddings so top/bottom gets the xs spacer and left/right gets sm, but could go a bit smaller all around):

Normal plain button vs small plain button

src/patternfly/components/Button/button.scss
```diff
@@ -143,6 +143,12 @@
...
+  --#{$button}--m-plain--m-inline--PaddingTop: 0;
+  --#{$button}--m-plain--m-inline--PaddingRight: var(--#{$pf-global}--spacer--sm);;
+  --#{$button}--m-plain--m-inline--PaddingBottom: 0;
+  --#{$button}--m-plain--m-inline--PaddingLeft: var(--#{$pf-global}--spacer--sm);;
+  --#{$button}--m-plain--m-inline--LineHeight: normal;
```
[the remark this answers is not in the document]
thatblindgeye: Updating from `normal` to `var(--#{$pf-global}--LineHeight--md)` causes a slight difference in height between the button and whatever text it might be next to, with the button height slightly exceeding the height of the text next to it.

With the global var:

Text height next to plain inline button with global var
Plain inline button height with global var


With the `normal` keyword:

Text height next to plain inline button with normal keyword
Plain inline button with normal keyword

Screenshots are from Firefox but Chrome has similar height differences, which get more significant as the font size increases. Though this could possibly depend how exactly a plain+inline button would actually be used... wdyt?

mcoker:
@thatblindgeye that all seems on point to me. We call those buttons "small" now. I wonder if this is a case for a "compact" button variant, assuming there could be a meaningful distinction between small and compact.

I know I created the original issue for this change, but I'm not totally clear on the role of an inline button. And I'm wondering if a compact button would negate any immediate need for an inline variant. From what I can tell, the padding/math in your inline screenshots looks good to me though.

I'm interested in the user's comment from the issue around their need for this feature, too, as well as any design input. cc @andrew-ronaldson @lboehling 

Side note, I'd love to see the "small" buttons in a separate example block called "small", they get lost in the variations block now.

thatblindgeye:
If we wanted to differentiate between "small" and "compact", "small" could refer to the font size while "compact" refer to both the font size *and* padding (or just padding?).

That all sounds good, though, will wait for some feedback from everyone else. Depending where we end up going with this PR I'll also move those small buttons to a separate example as part of this.

nicolethoen:
> @nicolethoen do you recall that conversation? I looked in the office hours agenda before that comment and skimmed a couple of the recordings and didn't see that discussion.

I don't - I looked at which office hours he'd be referring to - the one on March 15. And it must have been a question posed after the recording was stopped 🙁

andrew-ronaldson (approved):
Looks good to me but defer to Coker and Sarah on any outstanding code changes.

thatblindgeye:
@andrew-ronaldson @mcoker this can use another look after latest updates. I replaced the previous inline styling with a simpler compact styling as that was along the lines of the initial use-case. 

I also moved the small buttons to their own example per convo above and to align with the React examples.

mcoker (approved):
💪🚀👍

andrew-ronaldson:
Happy days. Looks great!

src/patternfly/components/Button/examples/Button.md
```diff
@@ -406,6 +411,7 @@ Semantic buttons and links are important for usability as well as accessibility.
 | `.pf-m-warning` | `.pf-v5-c-button` | Modifies for warning styles. |
 | `.pf-m-link` | `.pf-v5-c-button` | Modifies for link styles. This button has no background or border and is styled as a link. This button would commonly appear in a form and may include an icon. |
 | `.pf-m-plain` | `.pf-v5-c-button` | Modifies for icon styles. This button has no background or border, uses a standard text color, and is used for `.pf-m-plain` icon buttons such as close, expand, kebab, etc. |
+| `.pf-m-no-padding` | `.pf-v5-c-button.pf-m-plain` | Modifies a plain button to remove padding. Should only be used when the button is inline within a sentence or block of text. Multiple plain buttons without padding should not be placed adjacent to one another without spacing between them. |
```
srambach, line 414: I'd make the second sentence a full sentence - "This modifier should only be used when the button is inline within a sentence or block of text."
And then maybe simplify the last sentence? I don't love this wording but was trying to make it what TO do rather than what not to do.
"Adjacent plain buttons without padding should always have spacing between them."

srambach (approved):
👍

PR #5700, opened 2023-06-30
fix(toolbar): fixed vertical alignment issue

### LINKED ISSUE #5688: Bug - [ToolbarItem] - [ToolbarItem label is not centered]
**Describe the problem**

After upgrading from an old PF5 alpha release to a new one the `<ToolbarItem variant="label">` is no longer centered.

```
-    "@patternfly/patternfly": "5.0.0-alpha.53",
-    "@patternfly/react-core": "5.0.0-alpha.98",
+    "@patternfly/patternfly": "5.0.0-alpha.69",
+    "@patternfly/react-core": "5.0.0-alpha.127",
```

The problem can be seen on the [pf5 demo website](https://pf5.patternfly.org/components/toolbar#stacked-on-desktop) the label [used to be centered](https://www.patternfly.org/v4/components/toolbar#stacked-example).

Our [pixel tests show](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18823-20230626-090824-316dfa91-fedora-38/pixeldiff.html#TestTerminal-testBasic-header) a clear difference:

**Expected behavior**

The label is centered.

Applied a css workaround:

```
div.pf-v5-c-toolbar__group {
  align-items: center;
}
```

**Screenshots**

image

Cockpit.

**Any other information?**

This might be related, although I am not 100% sure https://github.com/patternfly/patternfly/pull/5565 what I can see is that it seems to have had `align-items: center` before.

mattnolting:
closes #5688 

- Utilizing `margin-top` [in the check](https://github.com/patternfly/patternfly/compare/main...mattnolting:fix-toolbar-5688?expand=1#diff-2ebe9a69a8f32a342f05d24c6a666599c84c77547d41adc52d63dd35b11870a8L267) and [radio](https://github.com/patternfly/patternfly/compare/main...mattnolting:fix-toolbar-5688?expand=1#diff-601dac91bce44671e202bba80bb2f3654e8051ff8cc0b0f0aa93164449c74601L11) cause the height of the parent to increase when in context. Apply `inline-{layout}` to `.ws-preview-html` and inspect the parent to see the difference. 

**Before update**
Screenshot 2023-06-30 at 2 55 41 PM

**After update**
Screenshot 2023-06-30 at 2 55 04 PM

- Apparently, FF doesn't like [`flex`ing buttons](https://bugzilla.mozilla.org/show_bug.cgi?id=984869#c2)

- Setting items to `align-items: baseline` fixes vertical alignment for labels

https://codesandbox.io/s/frosty-shannon-4m7nny?file=/style.css


src/patternfly/components/Radio/radio.scss
```diff
@@ -8,7 +8,7 @@
   --#{$radio}__label--FontSize: var(--#{$pf-global}--FontSize--md);
   --#{$radio}__label--LineHeight: var(--#{$pf-global}--LineHeight--sm);
   --#{$radio}__input--Height: var(--#{$radio}__label--FontSize);
-  --#{$radio}__input--MarginTop: calc(((var(--#{$radio}__label--FontSize) * var(--#{$radio}__label--LineHeight)) - var(--#{$radio}__input--Height)) / 2);
+  --#{$radio}__input--TranslateY: calc(((var(--#{$radio}__label--FontSize) * var(--#{$radio}__label--LineHeight)) - var(--#{$radio}__input--Height)) / 2);
```
srambach, line 11: We really should be including comments in the code to document stuff like that so that in the future we can more easily know what's going on and why this was necessary.

mcoker (commented):
From the [issue initially reported with `baseline` alignment](https://github.com/patternfly/patternfly/issues/5532) (why we moved to `start`), looks like that example is still a little off in FF and chrome - https://patternfly-pr-5700.surge.sh/components/toolbar/#selected-filters-on-desktop-not-responsive

Screenshot 2023-07-05 at 7 07 29 PM

Screenshot 2023-07-05 at 7 07 24 PM

FWIW if you update the standalone variant from `inline-grid` to `inline-block`, the bulk select box alignment looks better. That likely improves the layout of putting other things, like regular text or whatever, adjacent to the standalone input. Not sure if there are other implications of that or if that's a good fix, but figured I would mention it.

Also this needs to change to `--TranslateY` - https://github.com/mattnolting/patternfly/blob/a0cc480d72c53b4a8ffe0bc7b14797620d523fbc/src/patternfly/components/MenuToggle/menu-toggle.scss#L368

mattnolting:
> From the https://github.com/patternfly/patternfly/issues/5532 (why we moved to start), looks like that example is still a little off in FF and chrome

@mcoker I think that's to be expected as the height was being pushed off by margin. LMK what you think now.

src/patternfly/components/Radio/radio.scss
```diff
@@ -38,7 +38,7 @@
 
 .#{$radio}__input {
   height: var(--#{$radio}__input--Height);
-  margin-top: var(--#{$radio}__input--MarginTop);
+  transform: translate(var(--#{$radio}__input--TranslateY));
```
srambach, line 41: This should be `translateY`
Also, I'd suggest adding the same comment here to explain why we must manually nudge this with translate.

src/patternfly/components/Check/check.scss
```diff
@@ -52,7 +52,7 @@
 
 .#{$check}__body {
   grid-column: 2;
-  margin-top: var(--#{$check}__body--MarginTop);
+  transform: translateY(var(--#{$check}__body--TranslateY));
```
srambach, line 55:
```suggestion
  transform: translateY(var(--#{$check}__body--TranslateY)); // inline-block causes a misalignment that this fixes
```
(something like that; i.e. why the heck are we manually adjusting this)

thatblindgeye:
One thing I'm noticing with the update from `inline-grid` to `inline-block` is on a split button MenuToggle:

image

vs what it looks like in staging:

image

The other split button examples look fine since they include text, it's just the split button checkbox without any text that now looks off

src/patternfly/components/Check/check.scss
```diff
@@ -8,10 +8,10 @@
   --#{$check}__label--FontSize: var(--#{$pf-global}--FontSize--md);
   --#{$check}__label--LineHeight: var(--#{$pf-global}--LineHeight--sm);
   --#{$check}__input--Height: var(--#{$check}__label--FontSize);
-  --#{$check}__input--MarginTop: calc(((var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight)) - var(--#{$check}__input--Height)) / 2);
+  --#{$check}__input--TranslateY: calc(((var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight)) - var(--#{$check}__input--Height)) / 2);
   --#{$check}__description--FontSize: var(--#{$pf-global}--FontSize--sm);
   --#{$check}__description--Color: var(--#{$pf-global}--Color--200);
-  --#{$check}__body--MarginTop: var(--#{$pf-global}--spacer--sm);
+  --#{$check}__body--TranslateY: var(--#{$pf-global}--spacer--sm);
```
mcoker, line 14: This is pushing the body outside of bounds since using `translate()` won't impact the layout of things around it - for the body, we want any space above the it to be part of the layout so stuff is moved around accordingly. Is there a reason this update is needed?

Screenshot 2023-07-06 at 3 16 09 PM

src/patternfly/components/Select/select.scss
```diff
@@ -264,7 +264,7 @@
 
 .#{$select}__toggle {
   position: relative;
-  display: flex;
+  display: inline-flex;
```
mcoker, line 267: Is this needed?

src/patternfly/components/Check/check.scss
```diff
@@ -25,9 +25,9 @@
   &.pf-m-standalone {
     --#{$check}--GridGap: 0;
     --#{$check}__input--Height: auto;
-    --#{$check}__input--MarginTop: 0;
+    --#{$check}__input--TranslateY: 0;
 
-    display: inline-grid;
+    display: inline-block;
```
mcoker, line 30: I'm assuming we'd want to make this to the radio, too, for consistency?

mcoker (commented):
menu toggle split button checkboxes look off

Screenshot 2023-07-06 at 3 18 46 PM

Dropdown split button changed a little, but IMO looks fine, maybe better than it did before

Screenshot 2023-07-06 at 3 32 48 PM


Table checkboxes are aligned a little low

Screenshot 2023-07-06 at 3 27 14 PM

mcoker (commented):
Looks like the bulk select split button dropdown is still off by a pixel - https://patternfly-pr-5700.surge.sh/components/toolbar/#selected-filters-on-desktop-not-responsive

From messing around in dev tools, I might try this:
https://github.com/patternfly/patternfly/blob/443f457c07431712e384dc8f36516797b01edea2/src/patternfly/components/Dropdown/dropdown.scss#L346-L348

* add `align-self: stretch` - this matches menu toggle split button. I think the menu toggle would work fine here, btw.
* remove `display: flex;` and `align-items: center;`
  * If you modify `display` here, be sure to check the progress example, that may need some attention - https://patternfly-pr-5700.surge.sh/components/menus/dropdown#split-button-progress-checkbox

https://github.com/patternfly/patternfly/blob/443f457c07431712e384dc8f36516797b01edea2/src/patternfly/components/Dropdown/dropdown.scss#L360
* remove `transform: translateY(var(--pf-v5-c-dropdown__toggle--m-split-button__toggle-check__input--TranslateY))`
* remove that var ^ `--pf-v5-c-dropdown__toggle--m-split-button__toggle-check__input--TranslateY`


Also the check + label alignment doesn't seem to work well on Safari. I'm also a little concerned this may not look OK on older versions of safari... I don't have a way to reliably test that.

Screenshot 2023-07-06 at 4 31 11 PM

src/patternfly/components/Select/select.scss
mattnolting: Yes, this resolves an alignment issue related to FF disliking `flex`ing buttons.

Before

Screenshot 2023-07-07 at 7 39 13 AM


After

Screenshot 2023-07-07 at 7 39 30 AM

mattnolting:
@mcoker @srambach @thatblindgeye 

Here's what this update does

• Standardizes our approach to`check/radio` alignment
• Dries out the complex margin calcs
• Looks more consistent
• Resolves height issues cause by top margin 

@mcoker Looks like we were using margining to prevent vertical centering when labels wrap. 

**With centered aligment:**

Screenshot 2023-07-07 at 8 26 42 AM

**With baseline alignment:**

Screenshot 2023-07-07 at 8 25 52 AM

I applied height and `align-self: center` to `--#{$check/$radio}__input`, but that caused an issue with Safari and looked great in Chrome/FF. I remove the height and set `--#{$check/$radio}__input` to `align-self: baseline` and that looked great in Chrome/FF, but not Safari. Given the unique bugs FF presents (also doesn't like buttons being `flex`'d), I believe allowing the exception to `-moz-transform: translateY(--#{$radio}__input--TranslateY);` is the least intrusive and most logical approach to proper alignment.

src/patternfly/components/Dropdown/dropdown.scss
```diff
@@ -637,6 +636,7 @@ $pf-v5-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 // Loading spinner
 .#{$dropdown}__toggle-progress {
   position: absolute;
+  left: var(--#{$dropdown}__toggle--m-split-button--child--PaddingLeft);
```
mattnolting, line 639: @mcoker This needs to be consistently aligned with padding, ie doesn't need a unique variable.

thatblindgeye (commented):
Other than what was mentioned on Slack regarding the Radio component in the preview surge, things are looking good. I am noticing that when adding a checkbox to a Toolbar Item, the alignment is still off (just added to one of the stacked examples):

image

Applying one of the `align` modifier classes resolves it, but considering in v4 it was centered by default I imagine we'd want to keep that behavior for v5.

srambach (approved):
The menu toggle checkbox is a skosh lower but I'm ok with it in the interest of getting this in. Is this alignment ok for the non-full height menu toggle with the avatar?

image

thatblindgeye (commented):
Awesome job (and persistence!) on this.  Just had one comment below, but it could be a followup if needed

thatblindgeye (approved):

src/patternfly/components/Check/check.scss
```diff
@@ -18,17 +17,20 @@
   --#{$check}__label-required--FontSize: var(--#{$pf-global}--FontSize--sm);
   --#{$check}__label-required--Color: var(--#{$pf-global}--danger-color--100);
 
-  display: grid;
+  display: inline-grid;
```
mcoker, line 20: The check and radio components are expected to be block/full-width elements by default. I have a feeling this is going to have some other repercussions. For example, this will introduce a regression from this update - https://github.com/patternfly/patternfly/issues/5141

Why is it needed for them to be inline?

src/patternfly/components/Radio/radio.scss
```diff
@@ -6,26 +6,24 @@
...
+  --#{$radio}__label--LineHeight: var(--#{$pf-global}--LineHeight--md);
+  --#{$radio}__input--TranslateY: #{pf-size-prem(5px)}; // FF needs an offset
   --#{$radio}__input--first-child--MarginLeft: #{pf-size-prem(1px)};
   --#{$radio}__input--last-child--MarginRight: #{pf-size-prem(1px)};
   --#{$radio}__description--FontSize: var(--#{$pf-global}--FontSize--sm);
   --#{$radio}__description--Color: var(--#{$pf-global}--Color--200);
   --#{$radio}__body--MarginTop: var(--#{$pf-global}--spacer--sm);
 
-  display: grid;
+  display: inline-grid;
```
mcoker, line 17: Same comment here as the checkbox - it's expected this component will have normal block formatting.

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -165,7 +165,7 @@
 
   position: relative;
   display: inline-flex;
-  align-items: center;
+  align-items: baseline;
```
mcoker, line 168: This could have some unexpected repercussions, too, based on what all is going inside of a menu toggle... curious this is needed?

Here is the big avatar menu toggle example - https://patternfly-pr-5700.surge.sh/components/menus/menu-toggle#with-avatar-and-text-full-height
Screenshot 2023-07-10 at 5 22 10 PM

It may just be that these menus are missing `.pf-m-full-height` (per the example title) but we'd still run the risk of a taller menu toggle (due to the content) having misaligned things in it.

mcoker (changes requested):
This looks really good! Left some questions. If we need to leave the change of the default alignment for the menu toggle, I think we'll need to test that well once it lands in react since they have all of the cool examples and demos of menu toggle in the dropdown/select/etc - stuff like text input groups in menu toggles and other stuff that may shift the height of the toggle.

Also, looks like split button menu toggles in FF look like they need the translate() removed 

Screenshot 2023-07-10 at 5 45 46 PM

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -15,15 +15,13 @@ $pf-v5-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2
   --#{$toolbar}--item--RowGap--base: var(--#{$pf-global}--spacer--xs); // shared row-gap for items and groups
 
   // Item
-  --#{$toolbar}__item--Display: block;
+  --#{$toolbar}__item--Display: flex;
   --#{$toolbar}__item--MinWidth--base: auto;
-  --#{$toolbar}__item--AlignSelf: auto;
-  --#{$toolbar}__item--AlignItems: var(--#{$toolbar}--AlignItems--base);
+  --#{$toolbar}__item--AlignSelf: start;
```
mcoker, line 20:
```suggestion
  --#{$toolbar}__item--AlignSelf: flex-start;
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -6,7 +6,7 @@ $pf-v5-c-toolbar--spacer-map: build-spacer-map("none", "sm", "md", "lg");
 $pf-v5-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
 
 .#{$toolbar} {
-  --#{$toolbar}--AlignItems--base: flex-start;
+  --#{$toolbar}--AlignItems--base: start;
```
mcoker, line 9: This should probably stay as `flex-start` until safari 17, it only got support in 15.4.

```suggestion
  --#{$toolbar}--AlignItems--base: flex-start;
```

mcoker (changes requested):

src/patternfly/components/MenuToggle/menu-toggle.scss
mcoker, line 168: It looks like we're using `align-items: center` on the dropdown toggle - I'm curious if you know what breaks when we use that here in the dropdown? It seems like we would (almost?) always want center aligned items. I know we've gone back and forth on this, but I'm curious where we landed and if we might be able to center align these items again, or if maybe the dropdown needs to be updated to match since they're so similar.

mcoker (changes requested):
Left one comment about align-items in menu toggle. Also looks like:

* Checkbox/radio alignment is still off in safari
* Toolbar labels are still off, since they're inheriting alignment from the parent. We should be able to target those specifically with `.pf-m-label`

mcoker (commented):
A couple of things that might be good to look at if it take 15 minutes to fix:
* Split dropdown spinner vertical alignment is different different with and without toggle text
* Split button menu toggle - check with text label is too wide and has extra space on the right. It seems related to width: 100% - at least removing that fixes it. It might be good to track down why, but it could be a local fix in the menu toggle though I wonder if it will continue to pop up like this other places.

A couple more that we can either live with and/or follow up
* There is space above/below split button toggle-check that isn't clickable (it should probably align stretch)
  * This is pre-existing in dropdown, but a regression in menu-toggle
* Split button menu toggle can go to multiple lines with enough content, and check/toggle icon are center aligned in pf4 and top aligned in pf5. Ideally at least the toggle icon would be center aligned, and I think we should get design to chime in on the ideal layout for that.

src/patternfly/components/Check/check.scss
```diff
@@ -18,32 +16,27 @@
...
   grid-template-columns: auto 1fr;
   grid-gap: var(--#{$check}--GridGap);
+  width: 100%;
```
mcoker, line 22: I'm not sure if this is the root cause, but the check in a split button menu toggle is wider than it needs to be (wider than the content) and it's pushing the caret icon over...

What it looks like now
Screenshot 2023-07-14 at 1 09 14 PM

Without `width: 100%`
Screenshot 2023-07-14 at 1 09 19 PM

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -347,6 +348,7 @@
 
   &.pf-m-split-button {
     padding: 0; // pass padding to children
+    align-items: baseline;
```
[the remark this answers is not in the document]
mcoker: Also the menu toggle button/icon are going to be baseline aligned now instead of center. That's _probably_ alright since I don't know how often this would come up, and the typeahead menu toggle has its own alignment that aligns the toggle button to the center (and that's used for typeahead select). Tho if we can center align it, that would probably be good so this doesn't pop up as a bug later. It was center aligned in PF4, too.

PF4
Screenshot 2023-07-14 at 1 27 11 PM

This PR
Screenshot 2023-07-14 at 1 27 17 PM

mcoker (commented):
Left some comments. In order of priority, I think we should look at:
* dropdown__toggle-check is no longer a flex layout, so that's creating extra space between the check and text and will cause wrapping layout issues if the toggle runs out of room for the text. **this is the only one I think would be good to sort out now if possible**. The rest are likely edge cases
* split button menu toggle's icon should ideally be center aligned.
* stretch align the split button toggle's check label so the whole toggle is clickable
* see why the split button toggle's checkbox + label is pushing the toggle arrow icon over in FF. That's going to cause a misalignment with that arrow (pushed too far to the right) compared to other toggles

srambach (commented):
Just some naming nits 😅

mcoker (changes requested):
This looks awesome! Just a couple of nits, but I think we need to fix the menu toggle typeahead button, and the masthead alignment is off.

Screenshot 2023-07-17 at 6 15 49 PM

mcoker (approved):
💪🚀👍

PR #5703, opened 2023-07-03
fix(form-control): change divs to spans

### LINKED ISSUE #5701: Bug - Form control - form control element wrapper should be span/phrasing content
Currently it's a `<div>` and that limits where it can be used since that is flow content, and often a form control needs to go in phrasing content, like a `<label>`. Let's update that to a `<span>` and see if there are any further updates needed when changing the element type.

srambach:
Fixes #5701 
Changes all the divs in a form control to be spans to allow a form control to be inside phrasing content.


mcoker (approved):
lgtm!

Issue #5704, opened 2023-07-03
Clean up a11y testing hitting bad URLs


nicolethoen:
Due to some of the IA updates, and deprecation of some components for v5, the logic for determining which URLs to hit with the a11y testing needs to be updated.

<img width="1403" alt="Screenshot 2023-07-03 at 2 23 18 PM" src="https://github.com/patternfly/patternfly/assets/26305082/05949c9b-c8b2-4e14-9c99-89319c725b06">

PR #5705, opened 2023-07-03
chore(a11y): update logic for determining URLs to test

nicolethoen:
Closes: https://github.com/patternfly/patternfly/issues/5704

### DIFF
--- a/patternfly-a11y.config.js
+++ b/patternfly-a11y.config.js
@@ -10,9 +10,22 @@ const urls = Object.keys(fullscreenRoutes)
       return key;
     } else {
       const path = fullscreenRoutes[key].path
-      if (path.match(/\/demos\/.*\/html-demos$/g)) {
+      if (path.match(/\/patterns\/.*\/html-demos$/g)) {
         return path.replace(/\/html-demos$/, '');
       } else {
+        if (
+          path.includes('/application-launcher/') |
+          path.includes('/options-menu/') |
+          path.includes('/dropdown/') |
+          path.includes('/select/')
+        ) {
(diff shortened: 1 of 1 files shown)


patternfly-a11y.config.js
```diff
@@ -10,9 +10,22 @@ const urls = Object.keys(fullscreenRoutes)
       return key;
     } else {
       const path = fullscreenRoutes[key].path
-      if (path.match(/\/demos\/.*\/html-demos$/g)) {
+      if (path.match(/\/patterns\/.*\/html-demos$/g)) {
         return path.replace(/\/html-demos$/, '');
       } else {
+        if (
```
mcoker, line 16: Do we need to add context selector here? BTW - not sure if this is an issue but the demo page shows up first now, maybe that's intentional since the demos are still valid - they would just ideally be built using menu/menu toggle. Since core isn't going to have a context selector component, I don't know if the context selector demos belong there though. Or maybe the nav link needs a "demo" badge, like password strength/generator?

mcoker (approved):
LGTM - just one thing, seems like we might be missing context selector in this group?

nicolethoen:
so, context-selector has html demos as its default tab. Its case in handled on line 13 of this file and its the only deprecated component which has any demos at all.

the context-selector demos use the deprecated context-selector. I think that's the only place where that's the case, because generally, we've rewritten old demos to use recommended/non-deprecated components, so we _do_ want them to appear before the deprecated component docs. The context-selector docs is the exception to that rule. Tab order is defined in the documentation-framework.

Issue #5230, opened 2022-11-10
Badge - Update examples to add screenreader text


thatblindgeye:
**Is this a new component or an extension of an existing one?**
Badge

**Describe the feature**
From https://github.com/patternfly/patternfly-org/pull/3277#issuecomment-1309131369, adding text via the `pf-screen-reader` class to help convey context of the badges better (rather than just through color)

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
N/A

**Any other information?**
N/A

PR #5708, opened 2023-07-05
chore(badge): added screenreader to unread badges

ashley-o0o:
closes: [#5230](https://github.com/patternfly/patternfly/issues/5230)


thatblindgeye (commented):
This looks good! Only couple things are:

- There are a couple of unrelated commits as part of the PR
- I'd defer to @mcoker regarding the formatting of the handlebars. Other instances of the screen-reader handlebars as it on a separate line from other content, but it might make more sense in those contexts

thatblindgeye (approved):
🎉 Thanks for making the update!

Issue #5463, opened 2023-04-05
Bug - Wizard - add type="button" to pf-c-wizard__nav-link


tlabaj:
**Describe the problem**
A clear and concise description of the problem. Which components are affected?
Wizard

**How do you reproduce the problem?**
"When a Wizard is used inside a form, clicking on a nav item causes the form to submit."
see react issue https://github.com/patternfly/patternfly-react/issues/8875

PR #5709, opened 2023-07-05
chore(wizard): fixed button type for wizard

ashley-o0o:
Fixed the button type so it no longer submits when clicked.
closes: [#5463](https://github.com/patternfly/patternfly/issues/5463)


mcoker (approved):
LGTM! 🥳

Issue #5385, opened 2023-02-15
Bug - lists: update roles to match React


thatblindgeye:
**Describe the problem**
Followup to https://github.com/patternfly/patternfly/pull/5357 (and possibly https://github.com/patternfly/patternfly-react/issues/8453)

The following components need roles updated to match React:

- Tabs
- Dropdown
- OptionsMenu
- SearchInput (the list for autocomplete examples)
- OverflowMenu
- ProgressStep may need `role="listitem"` added to it due to the `display:contents` CSS
- ApplicationLauncher and ContextSelector (I had updated them to `role="list"`, but should be menu role; the items will also need `role="menuitem"` added)

Components listed here and previously updated in 5357 could also be updated to have aria-labels added in in the examples.

PR #5753, opened 2023-07-11
fix(role): update roles to match React

ashley-o0o:
closes: [#5385](https://github.com/patternfly/patternfly/issues/5385)

### DIFF
--- a/src/patternfly/components/AppLauncher/deprecated/application-launcher.md
+++ b/src/patternfly/components/AppLauncher/deprecated/application-launcher.md
@@ -13,10 +13,10 @@ import './application-launcher.css'
 ```hbs
 {{#> app-launcher app-launcher--id="application-launcher-collapsed"}}
   {{#> app-launcher-menu}}
-    <li role="menuitem">{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link{{/app-launcher-menu-item}}</li>
-    <li role="menuitem">{{#> app-launcher-menu-item app-launcher-menu-item--type="button"}}Action{{/app-launcher-menu-item}}</li>
+    <li role="none">{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link{{/app-launcher-menu-item}}</li>
+    <li role="none">{{#> app-launcher-menu-item app-launcher-menu-item--type="button"}}Action{{/app-launcher-menu-item}}</li>
     {{> divider divider--type="li"}}
-    <li role="menuitem">{{#> app-launcher-menu-item app-launcher-menu-item--modifier="pf-m-disabled" app-launcher-menu-item--attribute='href="#" aria-disabled="true" tabindex="-1"'}}Disabled link{{/app-launcher-menu-item}}</li>
+    <li role="none">{{#> app-launcher-menu-item app-launcher-menu-item--modifier="pf-m-disabled" app-launcher-menu-item--attribute='href="#" aria-disabled="true" tabindex="-1"'}}Disabled link{{/app-launcher-menu-item}}</li>
   {{/app-launcher-menu}}
 {{/app-launcher}}
 ```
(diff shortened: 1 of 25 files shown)


src/patternfly/components/ContextSelector/__context-selector-menu-menu.hbs
```diff
@@ -1,70 +1,70 @@
 {{#> context-selector-menu-menu}}
-  <li>
+  <li role="menuitem">
```
thatblindgeye, line 2: For this file we'll want to update the `li` element to `role="none"`, then within the `context-selector-menu-list-item.hbs` component we want to add the `role="menuitem"`

src/patternfly/components/AppLauncher/deprecated/application-launcher.md
```diff
@@ -13,10 +13,10 @@ import './application-launcher.css'
 ```hbs
 {{#> app-launcher app-launcher--id="application-launcher-collapsed"}}
   {{#> app-launcher-menu}}
-    <li>{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link{{/app-launcher-menu-item}}</li>
-    <li>{{#> app-launcher-menu-item app-launcher-menu-item--type="button"}}Action{{/app-launcher-menu-item}}</li>
+    <li role="menuitem">{{#> app-launcher-menu-item app-launcher-menu-item--attribute='href="#"'}}Link{{/app-launcher-menu-item}}</li>
```
thatblindgeye, line 16: Similar to the above comment, we'd want to place this role on `app-launcher-menu-item` instead, then have this `li` element be `role="none"`

src/patternfly/components/Tabs/tabs-item.hbs
```diff
@@ -11,6 +11,7 @@
   {{~#if tabs-item--IsAdd}} pf-m-add{{/if}}
   {{~#if tabs-item--IsOverflow}} pf-m-overflow{{/if}}
   {{~#if tabs-item--modifier}} {{tabs-item--modifier}}{{/if}}"
+  role="presentation"
```
thatblindgeye, line 14: The `tabs-link` component will also need `role="tab"` passed in

thatblindgeye (changes requested):
Just some quick updates needed below to resolve some a11y errors. 

I'm also noticing that the deprecated Select has an a11y issue in the [loading menu item example](https://staging.patternfly.org/components/menus/select/html-deprecated#loading-menu-item) (linking to the staging site since it's pre-existing), since the `<li>` that contains the spinner doesn't have a `role="option"` child. This wasn't caught in React because the spinner is only present momentarily, whereas the HTML example it's persistent to showcase it.

We could do what the new Select is doing and update the markup so that the spinner is wrapped inside a `.pf-v5-c-select__menu-item` component, which would require React followup and possibly be slightly more significant a markup change. @mcoker what do you think? It feels more like a bug that needs resolving even though it's the deprecated select.

mcoker:
@thatblindgeye that sounds fine to me. From the sound of it, it doesn't seem like a breaking change on the core side, but would change the markup in the react component. I'm assuming that's OK outside of a breaking change release?

thatblindgeye:
For both Core and React it would only be changing the markup in that specific example (in React it's not something we provide via prop or anything, it's just an example showing how a consumer could set up a "view more" sort of loading item). So either way it doesn't sound like it'd be breaking to the component. cc @tlabaj just to confirm, though

tlabaj:
> For both Core and React it would only be changing the markup in that specific example (in React it's not something we provide via prop or anything, it's just an example showing how a consumer could set up a "view more" sort of loading item). So either way it doesn't sound like it'd be breaking to the component. cc @tlabaj just to confirm, though

@thatblindgeye I agree, this would not be breaking.

thatblindgeye (changes requested):
Below are the comments I mentioned during our Meet, let me know if you have any questions!

I am noticing some other a11y issues being flagged by axe, specifically examples that are utilizing things like favorite buttons (issue being that it's causing a parent with `role="menu"` to have children with in invalid role), but this is an existing issue in both repos and would be out of scope of this PR.

thatblindgeye (changes requested):
Awesome job with these updates 🎉 Just had one more quick fix below (plus a question for the leads)

src/patternfly/components/OptionsMenu/options-menu-menu-item.hbs
```diff
@@ -1,5 +1,5 @@
-<li>
-  <button class="{{pfv}}options-menu__menu-item{{#if options-menu-menu-item--modifier}} {{options-menu-menu-item--modifier}}{{/if}}"
+<li role="none">
+  <button role="option" class="{{pfv}}options-menu__menu-item{{#if options-menu-menu-item--modifier}} {{options-menu-menu-item--modifier}}{{/if}}"
```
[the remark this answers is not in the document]
mcoker: @thatblindgeye that would be fine with me. If the react markup is good, I think we should update the core markup since it's just documentation and less disruptive to users.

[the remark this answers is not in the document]
tlabaj: @mcoker @thatblindgeye I do not think we should update the react implantation unless it is a critical bug reported by a consumer since the component is deprecated.

mcoker (approved):
🥸👍

Issue #5258, opened 2022-11-28
Bug - Chip group: align with React markup


thatblindgeye:
**Describe the problem**
Followup from https://github.com/patternfly/patternfly-org/pull/3297#issuecomment-1329215470

The markup/attributes in Core don't matchup with React, which has `role="group"` on the chip group so an aria-label can be applied, and aria-hidden set in Core for the category name prevents the close button from using it as part of its accessible name via `aria-labelledby`.

PR #5765, opened 2023-07-13
fix(chip): added role and aria-label to match react

ashley-o0o:
closes: [#5258](https://github.com/patternfly/patternfly/issues/5258)


mcoker (approved):
🥳👍

Issue #5767, opened 2023-07-18
Bug - Toolbar - Filter/chip "clear" link button misaligned


mcoker:
<img width="895" alt="Screenshot 2023-07-18 at 12 37 42 PM" src="https://github.com/patternfly/patternfly/assets/35148959/c1d146cd-1e7b-4dee-b465-286ccde26366">

We can likely fix this in CSS, but I'm noticing a markup/class difference between the chip/filter examples in core and react. A CSS fix (likely) means we need to ensure we're recommending the right markup so we can make sure the CSS is targeted correctly.

Core uses `.pf-v5-c-toolbar__content.pf-m-chip-container` (outlined below) and react does not.

https://github.com/patternfly/patternfly/blob/4e1911f119b681468e61efd0bf6f0c869bbf03a4/src/patternfly/components/Toolbar/toolbar.scss#L481-L507

@mattnolting can you confirm the right markup and if there is an easy CSS fix that can target chip elements?

mattnolting:
@mcoker this markup works, but `pf-m-chip-container` can also be applied to `{{#> toolbar-group}}` instead of `{{#> toolbar-item}}`. I think that `.pf-m-inline` would do well top margin/padding of `--pf-v5-global--spacer--form-element` to match the other form elements. We can also create generic containers for inline elements...

PR #5769, opened 2023-07-18
fix(toolbar): revert alignment of chip/filter layout

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5767

This just reverts the alignment to what it was before. The label alignment isn't great, but we don't have any open issues against that, so I don't think it's something we need to fix for this release. Assuming it reverts the layout properly, we shouldn't need to test it extensively or worry about inadvertent side effects.


mattnolting (approved):
LPTM 👍

srambach (approved):

thatblindgeye (approved):

PR #5771, opened 2023-07-19
chore(tree-view): added default flag to max nesting var

### LINKED ISSUE #5452: TreeView - Override MaxNesting

`TreeView`

**Describe the feature**

Currently the tree view maxes out at 10 levels. Yes, sometimes we need more.

In the .scss file, it is defined as:

```scss
$pf-c-tree-view--MaxNesting: 10;
```

See: https://github.com/patternfly/patternfly/blob/7ddc2de5cfb96f204e62730878a9e77bed64b998/src/patternfly/components/TreeView/tree-view.scss#LL1-L1C33

It would be great if this could be overridden. To my understanding, it should be possible with:

```scss
$pf-c-tree-view--MaxNesting: 10 !default;
```

mattnolting:
closes #5452


srambach (approved):
🚀

mcoker (approved):
🎃👍

thatblindgeye (approved):

PR #5778, opened 2023-07-24
feat(global): added RTL support

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5791 and also implements logical properties across the library.

#### The changes:
* Search and replace
  * `padding-[top, right, bottom, left]`
  * `margin-[top, right, bottom, left]`
  * `text-align: [left/right]]`
  * `[top/right/bottom/left]`
  * `border-[...]-[top/right/bottom/left]`
* Manual conversion of shorthand to longhand values (eg, `margin: 1 2 3` to `margin-block-start: 1; margin-inline-end: 2; margin-block-end: 3; margin-inline-start: 2;`)
* Re-ordered a bunch of property declarations manually. I think stylelint fixed a few, but I moved most by hand
* Used shorthand property declaration for spacing utility x/y classes.

Some build related changes:
* Adds backstop screenshot visual regression tool<sup>[1]</sup>
* Adds `use-logical-spec` stylelint rule - currently set to always. This is a good baseline that there are no non-logical properties defined anywhere. We may need to modify that if we add directional properties back when RTL testing.

[1] - The TL;DR of backstop:
* Tests against the prod build (`yarn build && yarn serve` - listens on `localhost:5000`)
* Tests every example and demo - grabbed the URLs from our existing screenshot tool's method of grabbing example/demo URLs
* Tests on 3 viewports (mobile, tablet, desktop)
* It does not have any screenshots stored currently - adding screenshots in this PR to use as baseline to test against crashed the "files changed" tab since it adds a ton of images. Maybe that could be added at a separate time.
* Total testing time for all URLs viewports is around 20-30 minutes, I believe
* All of ^ is configurable. You can test specific URLs/components, viewports, test against the dev server on `:8001`, etc

Questions for @nicolethoen @wise-king-sullyman @tlabaj 
* We have issues of taking a screenshot before the page has fully rendered. Is there some sort of event triggered on the page when everything has loaded that we could use to add a class to the page (eg, `.window-load`)? If so we could configure the screenshot tool to wait until that class is present before grabbing a screenshot.
* I wonder if we could include the [styles that disable animations/transitions](https://github.com/patternfly/patternfly/blob/a6993607b9df861de250c0f5ec05d6b8f4541b61/backstop_data/engine_scripts/onReady.js#L3) (and here they are in the [docs framework package](https://github.com/patternfly/patternfly-org/blob/07f44ec1c8a33f5ae4d559dde30e7a5a8fb118c3/packages/documentation-framework/scripts/writeScreenshots.js#L13)) as styles included from the docs framework if some thing is set - a query string parameter (`?disable-animations1), a cookie, a media query (puppeteer can set [media features](https://pptr.dev/api/puppeteer.page.emulatemediafeatures)), etc...


srambach:
Quick half-first-pass of things to address:
- jumplinks border
- caret rotation/direction
- on the workspace the HTML code doesn't show up
- progress stepper spacing and vertical line
- slider - ball is off and ticks are off
- switch - ball is misplaced
- breadcrumbs - reverse the caret
- dual list selector - carets, sorting icons, search input
- form control spacing with icons
- text input group search icon
- tree view general wonkiness
- Utility classes - do we rename/add logical properties?

backstop_data/engine_scripts/cookies.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "domain": ".www.yourdomain.com",
```
srambach, line 3: Is this file supposed to be customized for PatternFly?

mcoker: Not this file, they provide a lot of example/templates for different configs/scripts you might want to use when testing. This would set a cookie with some value you might need for chromium to access a test URL

srambach (approved):
I'm ok with getting this merged since the diffs look good and we'll follow quickly with a visit to each component for checking RTL. I'd love to see a 👍 from @nicolethoen or @dgutride on the backstop stuff.

tlabaj:
> Questions for @nicolethoen @wise-king-sullyman @tlabaj
> 
> * We have issues of taking a screenshot before the page has fully rendered. Is there some sort of event triggered on the page when everything has loaded that we could use to add a class to the page (eg, `.window-load`)? If so we could configure the screenshot tool to wait until that class is present before grabbing a screenshot.

@mcoker I believe this is what you are looking for
https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event

nicolethoen (approved):
No objections from me!
I opened followup issues to track tech-debt
https://github.com/patternfly/patternfly/issues/5848
https://github.com/patternfly/patternfly-org/issues/3703

PR #5782, opened 2023-07-26
chore(menu-toggle): updated demos to use menu-toggle

### LINKED ISSUE #5764: Update toolbar examples/demos to use `menu-toggle` instead of `dropdown`
Our existing toolbar demos include the dropdown, which has been deprecated. Those examples/demos need to use the `pf-v5-c-menu-toggle` instead.

mattnolting:
closes #5764


thatblindgeye (changes requested):
The ID's for the toolbar examples will need to be updated to be unique as well.

Some other things I noticed which may be out of scope, in which case we can do a followup if necessary

- There's still some usage of the deprecated dropdown, such as [Selected filters on mobile](https://patternfly-pr-5782.surge.sh/components/toolbar#selected-filters-on-mobile-filters-collapsed-selected-filters-summary-visible) (kebab button). 
- The "filter" plain toggle in some examples is just using the PF button instead of a MenuToggle; this is both Core and React, though cc @mcoker

thatblindgeye (approved):

src/patternfly/components/Panel/panel.hbs
```diff
@@ -1,4 +1,7 @@
 <div class="{{pfv}}panel{{#if panel--modifier}} {{panel--modifier}}{{/if}}"
+  {{#if panel--IsPopper}}
+    style="position: absolute; top: calc(100% + 2px); z-index: var(--pf-v5-global--ZIndex--sm);"
+  {{/if}}
```
mcoker, line 4: IMO we shouldn't include this as an inline style. It's probably better as a local demo style, just so it doesn't seem like it's part of the component.

src/patternfly/components/Toolbar/toolbar-item.hbs
```diff
@@ -2,6 +2,9 @@
   {{#if toolbar-item--id}}
     id="{{toolbar-item--id}}"
   {{/if}}
+  {{#if toolbar-item--IsRelative}}
+    style="position: relative;"
+  {{/if}}
```
mcoker, line 7: Same here about this local style - can it be moved to local demo styles if it's necessary? Also could we just not show expanded menus in core since the tools used to position those no longer lives in the CSS?

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -395,6 +395,10 @@ $pf-v5-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2
   &:last-child {
     --#{$toolbar}--spacer: 0;
   }
+
+  .#{$panel} {
+    --#{$panel}--MinWidth: 100%;
+  }
```
mcoker, line 401: Can you point link to the examples where this is used? I'm wondering if we need to scope this style further so it doesn't do anything we don't intend.

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -450,6 +454,10 @@ $pf-v5-c-toolbar--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2
 
 // Expandable content
 .#{$toolbar}__expandable-content {
+  .#{$menu-toggle} {
+    width: 100%;
+  }
```
mcoker, line 459: Same here - can you link to where this is seen? I'm wondering if this needs to be further scoped to not do something unexpected.

mcoker (commented):
Massive! Just a few comments.

src/patternfly/components/Toolbar/toolbar.scss
mattnolting: **With toggle at 100%**

Screenshot 2023-10-05 at 1 35 10 PM

mcoker:
Here's the [backstop report](https://github.com/patternfly/patternfly/files/12826864/backstop-errors.pdf)
* Split button menu toggles - it's known/expected those are wider in the menu toggle than in dropdown
* Menu toggles in the masthead have a transparent background - I think that's how they were originally intended to look anyways
* Extra space in menu items with checkboxes - left a comment/code suggestion to address that
* Toolbar filter toggle icon changed from grey to black - left a comment to address that
* There is extra space between toolbar expandable content filter labels the stuff below them - left a comment

src/patternfly/components/Toolbar/toolbar.scss
mcoker: Ah, thanks! I think that should be a separate issue and we can see how to handle that. Filed an isssue here - https://github.com/patternfly/patternfly/issues/6008

I think we used to get this for free because a select was used there, and the select toggle width is 100%? I'm wondering if:
* there is ever a use case for a menu toggle that isn't a select - meaning we might not want to style all menu toggles this way
* the react examples have a fixed width assigned to the filters in the expanded content - https://www.patternfly.org/components/toolbar. Is that on purpose? If they should be full-width, should that be something the react `<Select>` component sets by default? Or should we encourage the use of `.pf-m-full-width` on the menu toggle for those filters in our examples (change the fixed width to the full width modifier)

src/patternfly/components/Toolbar/toolbar.scss
mcoker: Ping - didn't see a reply here. Similar to the menu toggle comment, I'm curious if this is fixing a known bug and if this might do something we don't want it to since it has a broad scope. If it's a bug fix/enhancement, can we move it into another issue?

mcoker (changes requested):
Screenshots look good! Left 2 comments about the SCSS changes that could have an unintended impact. I filed a separate issue for the [menu toggles in toolbar expanded content](https://github.com/patternfly/patternfly/issues/6008) - I'd like to have core/react review that and see what the best solution is. And left a comment about a panel style that was added - can you review and lemme know what you think?

mcoker (approved):
🎃👍
👖
🛼

Issue #5786, opened 2023-07-31
Bug - MenuItem - descriptions are wrapping on character not word


nicolethoen:
descriptions are wrapping on character not word
This is different than how the v4 select components handles wrapping.

![Screenshot 2023-07-31 at 9 44 11 AM](https://github.com/patternfly/patternfly/assets/26305082/d3864b95-5d41-4fe5-aaf1-4b899988c3f8)

reported by James Talton & Ansible

PR #5789, opened 2023-07-31
fix(menu): keep description from wrapping mid-word

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5786


srambach (approved):
👍

Issue #5792, opened 2023-08-02
Dependencies are out of date


mcoker:
the `@patternfly/[...]` dependencies are pulling in alpha versions, and there is a dependency tree error that aborts any sort of updates to npm packages

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: @patternfly/patternfly@0.0.0-development
npm ERR! Found: @patternfly/react-code-editor@5.0.0-alpha.116
npm ERR! node_modules/@patternfly/react-code-editor
npm ERR!   dev @patternfly/react-code-editor@"5.0.0-alpha.116" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer @patternfly/react-code-editor@"5.0.0-prerelease.13" from @patternfly/documentation-framework@5.0.7
npm ERR! node_modules/@patternfly/documentation-framework
npm ERR!   dev @patternfly/documentation-framework@"5.0.7" from the root project
```

PR #5793, opened 2023-08-02
chore: update dependencies to latest

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5792

* runs `ncu -u`
* disables stylelint's `"media-query-no-invalid"` - it's recommended to turn off in SCSS https://stylelint.io/user-guide/rules/media-query-no-invalid/
* removes `stylelint-value-no-unknown-custom-properties` - fails a dependency check and doesn't appear to be used
* Updates import line for glob - v10 removed default exports, and now only has named exports https://github.com/isaacs/node-glob/blob/main/changelog.md#1000
* Same^ for sass - https://sass-lang.com/documentation/breaking-changes/default-export/


nicolethoen:
I don't know why it'd suddenly be an issue when it wasn't before - but if you upload the node version here (and on the other github actions) https://github.com/patternfly/patternfly/blob/main/.github/workflows/pr-preview.yml#L29 to node: '18' it will likely resolve it.

nicolethoen (approved):

Issue #5797, opened 2023-08-03
npm script to deep clean the repo


mcoker:
Often when there are significant changes, we need to run `yarn clean && git clean -dfx` to really get the repo to a clean state. And if we have a modified package.json before we run that, we often also have to run `yarn install` prior, so the command looks like `yarn install && yarn clean && git clean -dfx`. 

I have this as a local script I run called "superclean" that I've shared with the team that some folks use. I think it would be useful to add to the repo so it's accessible for everyone.

PR #5798, opened 2023-08-03
chore: add superclean npm script

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5797

Usage when you've pulled in significant changes to package.json and want to start with a fresh repo:

```
yarn superclean
yarn install
yarn start
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "build:pficons": "gulp pfIcons",
     "build:pficonfont": "gulp pfIconFont",
     "clean": "gulp clean",
+    "superclean": "npm install && npm run clean && git clean -dfx",
     "dev": "gulp develop",
     "develop": "gulp develop",
     "format": "prettier --trailing-comma es5 --no-semi --single-quote --write src/**/*.js",


package.json
```diff
@@ -12,6 +12,7 @@
     "build:pficons": "gulp pfIcons",
     "build:pficonfont": "gulp pfIconFont",
     "clean": "gulp clean",
+    "superclean": "npm install && npm run clean && git clean -dfx",
```
nicolethoen, line 15: I don't think you need to run an install first? unless there is a goal in doing so that i'm unaware of? But the git clean is going to immediately remove what you just installed.

mcoker: I'm not sure why, but if there are significant package.json files, often if you don't run `npm install` first, you'll get a bunch of errors running `yarn clean` and/or `git clean -dfx`. @srambach ran into it today when pulling in the latest dependency updates - running `yarn install` before `yarn clean && `git clean -dfx` addressed it.

nicolethoen (approved):

Issue #5804, opened 2023-08-04
Bug - Dropdown - start value has mixed support


jschuler:
In a small create-react-app application, I was seeing this warning originating from Dropdown.css:

```
autoprefixer: start value has mixed support, consider using flex-start instead
```

I believe it is referring to this property:
https://github.com/patternfly/patternfly/blob/eb1fdc9d4ed0ab191452972892749f02dc946608/src/patternfly/components/Dropdown/dropdown.scss#L193

PR #5807, opened 2023-08-07
fix: updated use of flex start/end to be more compatible

mcoker:
The element it looks like autoprefixer is warning about was introduced in v5, but it looks like we have the same issue in a few other components, too. autoprefixer likely isn't catching them, I'm guessing, because either they aren't using/processing those component styles and/or the offending values are used in custom properties (css vars) and I don't know if autoprefixer can go as far as to trace the value of vars back through the CSS and see if the value is used on/in a flex layout.

IMO this isn't a showstopper since it's pre-existing in other places, and AFAIK no one has flagged it on any of the other components. But we should at least update it for good measure since it is also in violation of the browser support we currently have documented - Safari only supports the current value in versions >= 15.4.

----

fixes https://github.com/patternfly/patternfly/issues/5804

Dropdown:
* The first use of `start` was introduced in v5 - https://github.com/patternfly/patternfly/commit/d3033930db2ca1d4a83960f0e0794ac4551f1abb
* The second use was introduced 3 years ago - https://github.com/patternfly/patternfly/commit/ddb4ff3f6900b8a8065cae89a3feee4bd41fee7f

Label group:
* Introduced in v5 - https://github.com/patternfly/patternfly/commit/763f34722f0bc8bdded6ce41ddb268638db7ebf1

Progress stepper:
* Introduced 2 years ago - https://github.com/patternfly/patternfly/commit/92af87a40838eb9cf80858c2d20f599af6e9d313

Sidebar:
* Introduced 3 years ago - https://github.com/patternfly/patternfly/commit/9690373637ee2b5b202c7ed90a3945feec3a470d

### DIFF
--- a/src/patternfly/components/Dropdown/dropdown.scss
+++ b/src/patternfly/components/Dropdown/dropdown.scss
@@ -190,7 +190,7 @@ $pf-v5-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 
   position: relative;
   display: inline-flex;
-  align-items: start;
+  align-items: flex-start;
   max-width: 100%;
 
   .#{$divider} {
@@ -755,7 +755,7 @@ $pf-v5-c-dropdown--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg"
 
     &.pf-m-description {
       flex-direction: column;
-      align-items: start;
+      align-items: flex-start;
     }
 
     .#{$dropdown}__menu-item-main {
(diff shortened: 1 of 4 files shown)


nicolethoen (approved):

srambach (approved):
I don't think it's going to matter here but something to remember is that start and flex-start act slightly differently for some cases in RTL languages. #TIL https://csslayout.news/whats-the-difference-between-the-alignment-values-of-start-flex-start-and-self-start/

Issue #5805, opened 2023-08-04
Bug - Options Menu - docs failing a11y testing


nicolethoen:
It's disrupting PR builds / CI

```
147/265   /components/menus/options-menu
violations: aria-required-children, aria-required-parent, list, listitem
```

Seems to be related to changes introduced in https://github.com/patternfly/patternfly/pull/5753
and revealed once a11y tests were cleaned up in https://github.com/patternfly/patternfly/pull/5705

Issue #5796, opened 2023-08-03
Update markup for multi selectable Options Menu


thatblindgeye:
**Describe the enhancement or change**
Followup to comment https://github.com/patternfly/patternfly/pull/5753#discussion_r1267272383

Core markup should be updated to match React for the [multiple options example](https://www.patternfly.org/components/menus/options-menu/html-deprecated#multiple-options)

In Core the markup is an unordered list with nested unordered lists for each "group":
![image](https://github.com/patternfly/patternfly/assets/70952936/7aaedf06-c04a-4abd-87f5-e69175f85ca7)

In React it's `section` elements for each "group" and an unordered list in each section:
![image](https://github.com/patternfly/patternfly/assets/70952936/953a2c41-0adc-4f58-8485-24bbec0108e3)

**Is this request originating from a Red Hat product team? If so, which ones and is there any sort of deadline for this enhancement?**

**Any other information?**

PR #5809, opened 2023-08-08
chore(options-menu): updated markup to match react component

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5796
fixes https://github.com/patternfly/patternfly/issues/5805


mattnolting (approved):
LPTM 🚢 it!

PR #5843, opened 2023-08-14
feat(Tokens): add initial global tokens

srambach:
This PR adds three token files - one for the color palette, one for default theme tokens, and one for dark theme tokens. These are exported from the Figma Penta variables and styles file and processed with Style Dictionary. More information about what is happening at https://github.com/srambach/patternfly-style-dictionary

Tokens files are imported into `_variables.scss` but not used anywhere.

For the moment, dark tokens are wrapped in a selector so that they are only defined for `.pf-v5-theme-dark`. 

The tokens are in the order they are spit out for now.


src/patternfly/base/tokens/_tokens-dark.scss
```diff
@@ -0,0 +1,296 @@
+/**
+ * Do not edit directly
+ * Generated on Mon, 14 Aug 2023 18:11:22 GMT
+ */
+
+ //  stylelint-disable custom-property-pattern
```
mcoker, line 6: If you wanna add support for this pattern, here's what you'd do it

https://github.com/patternfly/patternfly/blob/47689a42e7557e2967ae3d3974ddaa7fcf251064/.stylelintrc#L42

I imagine you'd just do something like change `v5` to `(v5|t)-(global|[lcud])`

mcoker (approved):
LGTM

src/patternfly/base/_variables.scss
```diff
@@ -290,3 +290,7 @@
 @include pf-v5-theme-dark {
   @include pf-v5-theme-dark-variables;
 }
+
+@import "./tokens/tokens-palette";
+@import "./tokens/tokens-default";
+@import "./tokens/tokens-dark";
```
mcoker, line 296: Oh an extra line after this last line will make the little red icon go away. We generally add an extra line at EOF to all things, so if you were to import this file and another file right after it it would import as...

```
[...]
@import "./tokens/tokens-palette";
@import "./tokens/tokens-default";
@import "./tokens/tokens-dark";
// other file import would start here
```

insetead of

```
[...]
@import "./tokens/tokens-palette";
@import "./tokens/tokens-default";
@import "./tokens/tokens-dark";// other file import would start here
```

mcoker (approved):

Issue #5842, opened 2023-08-11
Bug - Chip - add margin for badge when used in text element


mcoker:
To fix https://github.com/patternfly/patternfly-react/issues/9483

TL;DR - we need to add back `--pf-c-chip__c-badge--MarginLeft: var(--pf-global--spacer--xs);`, and have it target the badge when it's a child of `.pf-[v]-c-chip__text`.

----

In core the badge is adjacent to `.pf-[v]-c-chip__text`, and in react the badge is a child of `.pf-[v]-c-chip__text`. Fun fact, this markup discrepancy goes back to the original implementation of a badge in the chip from 2019 via https://github.com/patternfly/patternfly/issues/1302 and https://github.com/patternfly/patternfly-react/pull/1245

For v5, core updated the CSS to add a `column-gap` to the parent of the badge + text, and removed the left margin from the badge when it's a child of a chip - https://github.com/patternfly/patternfly/pull/5360/files. This broke the badges used in the react chip component.

Talking with @tlabaj we want to 1) continue supporting core and react's markup (the change outlined above should allow that), 2) add a prop to the react chip component for a `count` or something similar that places the badge in the chip and matches core's markup, and 3) in our next breaking change, remove this left margin we're adding back from the CSS.

PR #5844, opened 2023-08-14
fix(chip): add margin for badge when used in text element

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5842

I copied the markup from the [react chip example with a badge](https://www.patternfly.org/components/chip/#chip-variants) into the core workspace for this PR

Screenshot 2023-08-14 at 6 33 45 PM


thatblindgeye (approved):

tlabaj (approved):
LGTM

PR #5860, opened 2023-08-24
fix: added base, wizard, pagination RTL styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5839
fixes https://github.com/patternfly/patternfly/issues/5840

----

## Global
* Adds a `pf-v5-rtl($transform)` mixin that manages 2 things so far:
  * Handles selector(s) we'll use to wrap RTL styles.
    * Currently works on `.ws-dir-rtl, .#{$pf-prefix}m-dir-rtl, [dir="rtl"]`.
  * Can return helpful, common blocks of CSS with a preferred way of doing some RTL transform by passing a `$transform`. 
    * Added a "flip-inline" transform that applies `scale: -1 1`
* Adds `.pf-v5-m-rtl-flip-inline` as a helper class to flip something horizontally (inline), and gets its code from ^ `pf-v5-rtl($transform: "flip-inline")`
* Adds `.pf-v5-m-dir-rtl` to apply `direction: rtl;` to something
* Adds `.pf-v5-m-dir-ltr` to apply `direction: ltr;` to something if you don't want it to flip in RTL. This is what keeps the dark theme and RTL switches in the bottom right of the full page examples from switching to RTL when the whole document changes to RTL.

## Wizard
* Added  `.pf-v5-m-rtl-flip-inline` to the wizard icon bem element to flip the icon in the "finished" example. **I'll take this out prior to merging, so it isn't confused as something that's required for use in the component.** It should probably go on the icon itself anyways, but it was easier to add it to the wizard icon bem element for this example.
* Added `pf-v5-rtl("flip-inline")` to the mobile nav breadcrumb arrows

## Pagination
* Added `pf-v5-rtl("flip-inline")` to the next/prev arrows

## Follow-ups
* I'd like to update the workspace RTL button to use either `dir="rtl"` (probably preferred, as I think this is our main use case) or `.pf-v5-m-dir-rtl` on full page examples. Also `.pf-v5-m-dir-ltr` on the dark theme/RTL switcher box. Then we can remove the current `.ws-dir-[rtl/ltr]` classes from the workspace code.
* Doesn't have to happen, but I'd also like to add a toggle in the masthead on the main workspace view (like the dark theme switcher) that can toggle `.pf-v5-m-dir-rtl` on the example blocks, which would let us quickly scan RTL examples in the example list page. That should be an easy add. It could eventually toggle the whole page, too, not the example blocks, if we wanted to test the workspace this way.
  * We could probably just add a plain action in the masthead that toggles between the FA [align-left](https://fontawesome.com/icons/align-left?f=classic&s=solid) and [align-right](https://fontawesome.com/icons/align-right?f=classic&s=solid) icons?

### DIFF
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -311,7 +311,7 @@
 }
 
 @mixin pf-v5-rtl($transform: null) {
-  @at-root :where(.ws-dir-rtl, [dir="rtl"]) & {
+  @at-root :where(.ws-dir-rtl, .#{$pf-prefix}m-dir-rtl, [dir="rtl"]) & {
     @if $transform == 'flip-inline' {
       scale: -1 1;
     }
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -310,11 +310,27 @@
   }
 }
 
+// Applies custom styles to support RTL
 @mixin pf-v5-rtl($transform: null) {
-  @at-root :where(.ws-dir-rtl, .#{$pf-prefix}m-dir-rtl, [dir="rtl"]) & {
-    @if $transform == 'flip-inline' {
-      scale: -1 1;
+  // resets property used to inverse transform values calc()s in RTL
+  @if $transform == "transform-flip-revert" {
+    --#{$pf-global}--rtl--transform--flip: 1;
+  } @else {
+    // establishes the selectors used to target RTL
+    // remove .ws-dir-rtl before releasing RTL
+    @at-root :where(.ws-dir-rtl, .#{$pf-prefix}m-dir-rtl, [dir="rtl"]) #{&} {
+      // sets property used in calcs to inverse transform values
+      @if $transform == "transform-flip" {
+        --#{$pf-global}--rtl--transform--flip: -1;
+      }
+
+      // flips something horizontally/inline. relies upon scale/scale() not already being used for the element
+      @if $transform == 'flip-inline' {
+        scale: -1 1;
+      }
+
+      // any custom CSS to apply within the selectors that target RTL
+      @content
     }
-    @content
   }
 }
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -309,3 +309,12 @@
     }
   }
 }
+
+@mixin pf-v5-rtl($transform: null) {
+  @at-root :where(.ws-dir-rtl, [dir="rtl"]) & {
+    @if $transform == 'flip-inline' {
+      scale: -1 1;
+    }
+    @content
+  }
+}
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -26,13 +26,20 @@
 
 // RTL helpers
 .#{$pf-prefix}m-dir-rtl {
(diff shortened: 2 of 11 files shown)


src/patternfly/sass-utilities/mixins.scss
```diff
@@ -309,3 +309,12 @@
     }
   }
 }
+
+@mixin pf-v5-rtl($transform: null) {
+  @at-root :where(.ws-dir-rtl, .#{$pf-prefix}m-dir-rtl, [dir="rtl"]) & {
```
thatblindgeye, line 314: Checked and looks to work well when applying `dir` attribute, `ws-dir-rtl` class, and the modifier class to either the `html` tag or the base component tag 👍🏼

thatblindgeye (changes requested):
This looks nice 💪🏼

For the Wizard expandable icon, I'd expect that when collapsed it would be facing to the left when rtl is enabled. By default we have the nav link text followed by the toggle icon facing right (away from the link text), but with RTL enabled we have the toggle icon facing right (towards the link text) followed by the link text. This would follow how the tree view dual list selector is styled, where the icon is facing towards the item text in both LTR and RTL.

Can we just add in the `@include pf-v5-rtl("flip-inline");` in the nav-link-toggle-icon block around line ~536 to resolve that?

mcoker:
@thatblindgeye nice find, updated!

wise-king-sullyman:
It looks like the `of`s in pagination are off? 
image

wise-king-sullyman:
Also no idea if this is actually an issue, but it's interesting to me that the text itself doesn't change, but the period moves around in the text here. Is that expected? 
image

src/patternfly/sass-utilities/mixins.scss
```diff
@@ -309,3 +309,28 @@
     }
   }
 }
+
+// Applies custom styles to support RTL
+@mixin pf-v5-rtl($transform: null) {
+  // resets property used to inverse transform values calc()s in RTL
+  @if $transform == "transform-flip-revert" {
```
thatblindgeye, line 316: Only nit is the naming of this particular value. Just wondering if "transform-flip(-revert)" conveys exactly what it's for, or if it's too close to "flip-inline"... "transform-modifier(-revert)" or "transform-multiplier(-revert)" maybe?

thatblindgeye (approved):
🚀 Only have a nit below, but ti isn't blocking for this PR.

thatblindgeye:
@wise-king-sullyman I feel like those are expected? If I render just the text `Lorem ipsum.` or `1-10 of 36` in a plain div, applying rtl to it it's the same result.

mcoker:
> Also no idea if this is actually an issue, but it's interesting to me that the text itself doesn't change, but the period moves around in the text here. Is that expected? image

@wise-king-sullyman I noticed that, too, and it appears to just be the way punctuation is handled between LTR/RTL. I don't think there's anything we can do about that, and I think it changes based on your language.

mcoker:
> It looks like the `of`s in pagination are off? image

Since those are just strings in core, I think that would look right in the correct language.

thatblindgeye:
Granted I'm viewing it after having the page translated to English, but when I go to the BBC site for Arabic, things look similar there where the punctuation is at the "start" (to us), and instead of "6 hours ago" an article reads "hours ago 6" (which would align to the Pagination x of y text).

mattnolting (approved):
Let's merge.

PR #5866, opened 2023-08-28
chore(Alert,NotificationDrawer): added additional rtl styling

### LINKED ISSUE #5817: Implement RTL: notification drawer

### LINKED ISSUE #5818: Implement RTL: alert and hint

thatblindgeye:
Closes #5817 
closes #5818 

For both Alert and Notification drawer, the truncation is slightly different between Chromium and Firefox when viewing an example in a new window (viewing the examples normally on the component page things like fine). In Chromium a line clamp of 1 seems to cause the first word to be cut off, while a line clamp greater than 1 seems to truncate okay (same screenshot from the alert/hint issue linked above):

Alert title truncation with RTL direction in Chrome

In Firefox, it seems to be the reverse; line clamp of 1 truncates okay, while line clamp greater than 1 causes the first word of the last line to be cut off:

Alert title truncation with RTL direction in Firefox

Since it seems to more have this affect if the Alert is wide enough it may not be a huge issue. This could also just be an issue with the Core examples, as applying `dir="rtl"` to the `html` tag in the React example the truncation looks fine (though that was also without any of the other RTL styling updates from Core).

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -199,7 +199,7 @@
 }
 
 .#{$alert}__toggle-icon {
-  @include pf-v5-rtl('flip-inline');
+  @include pf-v5-mirror-inline-on-rtl;
   
   display: inline-block;
   transition: var(--#{$alert}__toggle-icon--Transition);
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -199,6 +199,8 @@
 }
 
 .#{$alert}__toggle-icon {
+  @include pf-v5-rtl('flip-inline');
+  
   display: inline-block;
   transition: var(--#{$alert}__toggle-icon--Transition);
   transform: rotate(var(--#{$alert}__toggle-icon--Rotate));
--- a/src/patternfly/components/NotificationDrawer/notification-drawer.scss
+++ b/src/patternfly/components/NotificationDrawer/notification-drawer.scss
@@ -307,6 +307,8 @@
 }
 
 .#{$notification-drawer}__group-toggle-icon {
+  @include pf-v5-rtl('flip-inline');
+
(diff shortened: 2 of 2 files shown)


mcoker:
Hmm... I'm probably missing something, but these examples between chrome and firefox look the same to me:

Screenshot 2023-08-28 at 5 05 21 PM


Screenshot 2023-08-28 at 5 05 42 PM

src/patternfly/components/Alert/alert.scss
```diff
@@ -199,6 +199,8 @@
 }
 
 .#{$alert}__toggle-icon {
+  @include pf-v5-rtl('flip-inline');
```
mcoker, line 202:
```suggestion
  @include pf-v5-mirror-inline-on-rtl;
```

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -307,6 +307,8 @@
 }
 
 .#{$notification-drawer}__group-toggle-icon {
+  @include pf-v5-rtl('flip-inline');
```
mcoker, line 310:
```suggestion
  @include pf-v5-mirror-inline-on-rtl;
```

mcoker (changes requested):
LGTM! Just need to rebase and update to the new mixin name.

thatblindgeye:
@mcoker Ah yep, it's because I have my browsers at an increased default zoom level. Lowering both to 100% results in similar truncation between the two.

mcoker (approved):
🚀

srambach:
FWIW, this has the same deal as I'm seeing on the card expansion toggle icon - it spins when going from LTR to RTL because of the transition timing provided for the rotation. If we can't find a workaround quickly, I think we should move ahead as is and follow up with a fix for all these cases.

srambach (approved):
Looks good - just the one comment about the transition timing causing an extra little flip.

PR #5869, opened 2023-08-28
fix(calendar-month): rtl support

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5831


thatblindgeye (approved):
Looks good! Tried looking into the expectation for RTL calendars and it seems like what this PR does would be accurate. This [stack exchange post](https://ux.stackexchange.com/questions/112887/in-an-rtl-language-in-what-direction-should-calendar-days-go) may be an interesting read

PR #5873, opened 2023-08-29
chore(Nav, Menu): apply RTL fixes

### LINKED ISSUE #5761: Implement RTL: masthead and navigation
Apply necessary tweaks to masthead and navigation.

srambach:
Fixes up carets on navigation and scrollable navigation buttons. Fixed menu as part of drilldown navigation.
Navigation demos appear broken but should be fixed by Sidebar RTL fixes.
No fixes needed for masthead(?)

Fixes #5761

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -282,7 +282,11 @@
       // target first list in menu
       > .#{$menu}__content > .#{$menu}__list,
       > .#{$menu}__list {
-        transform: translateX(-100%);
+        @include pf-v5-bidirectional-style(
+          $prop: transform,
+          $ltr-val: translateX(-100%),
+          $rtl-val: translateX(100%)
+        );
       }
       // stylelint-enable
     }
@@ -299,8 +303,11 @@
 
       // stylelint-disable selector-max-class
       &.pf-m-drilled-in {
-        transform: translateX(-100%);
-      }
+        @include pf-v5-bidirectional-style(
+          $prop: transform,
+          $ltr-val: translateX(-100%),
+          $rtl-val: translateX(100%)
+        );      }
       // stylelint-enable
     }
 
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -285,8 +285,8 @@
         @include pf-v5-bidirectional-style(
           $prop: transform,
           $ltr-val: translateX(-100%),
-          $rtl-val: translateX(100%)
-        );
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Menu/menu.scss
```diff
@@ -282,7 +282,11 @@
       // target first list in menu
       > .#{$menu}__content > .#{$menu}__list,
       > .#{$menu}__list {
-        transform: translateX(-100%);
+        @include pf-v5-bidirectional-style(
+          $prop: transform,
+          $ltr-val: translateX(-100%),
+          $rtl-val: translateX(100%)
```
mcoker, line 288: For these places you're using `pf-v5-bidirectional-style` where you want the inverse of some value in RTL, if you use the [`pf-v5-calc-invserse` function](https://github.com/patternfly/patternfly/blob/447630c3449372332ee1628268397f16f5ffa965/src/patternfly/sass-utilities/functions.scss#L121-L123), that will allow the translate to go back to normal if this component is used in a block set to [`.pf-v5-m-dir-ltr`](https://github.com/patternfly/patternfly/blob/447630c3449372332ee1628268397f16f5ffa965/src/patternfly/base/_common.scss#L36-L42) since the inverse value is tied to `--#{$pf-global}--inverse--multiplier`

The example we have of this is with the switches in the bottom right of the full page examples. The little circles in the switch use `translateX()`, and use `pf-v5-calc-invserse` to control their position in RTL. The doc is RTL, but that box is set to `.pf-v5-m-dir-ltr`, which resets `--#{$pf-global}--inverse--multiplier`, so the `translateX()` is reset.

mcoker (changes requested):
I think the RTL bidir-style arg needs to use the inverse calc function. Otherwise LGTM!

Just noting that the menu item-toggle-icon needs to be mirrored per this example, but I'll update it in a menu PR. 

https://patternfly-pr-5873.surge.sh/components/navigation/html/nav-with-flyout/

mcoker:
@srambach actually if you can mirror that menu icon, I think that's the only other thing that needs to be done for the menu. With the 2 other changes you made in this PR, it LGTM. It's this element https://github.com/patternfly/patternfly/blob/447630c3449372332ee1628268397f16f5ffa965/src/patternfly/components/Menu/menu.scss#L688-L692

The breadcrumb is also used in the drilldown menu and doesn't look right, but there is a PR that just needs one more review https://github.com/patternfly/patternfly/pull/5870

thatblindgeye (approved):
I didn't notice anything in Masthead that needed updating either

mcoker (approved):

PR #5874, opened 2023-08-29
fix(rtl): fixes truncate rtl

### LINKED ISSUE #5763: Implement RTL: avatar, brand, background image, backdrop, divider, truncate, icon
Apply mixin created in https://github.com/patternfly/patternfly/issues/5758 to the page and login page components.

mattnolting:
closes #5763


src/patternfly/components/Icon/icon.scss
```diff
@@ -42,6 +42,12 @@
   width: var(--#{$icon}--Width);
   height: var(--#{$icon}--Height);
 
+  @include pf-v5-mirror-inline-on-rtl;
+
+  &.pf-m-dir-ltr {
+    direction: ltr;
+  }
```
mcoker, line 49: I don't think we want to reverse anything by default, and we have the global modifier that can be applied to reverse an icon (or anything else).

src/patternfly/components/Truncate/truncate.scss
```diff
@@ -29,6 +29,18 @@
...
+    @include pf-v5-bidirectional-style (
+      $prop: 'direction',
+      $ltr-val: 'ltr',
+      $rtl-val: 'rtl'
+    );
+
+    &::before {
+      content: '.';
+    }
+  }
```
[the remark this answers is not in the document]
mattnolting: Good point. I move `.pf-m-file-extension` to `.pf-v5-c-truncate__file-extension` so it can be inserted in `__start` or `__end`

src/patternfly/components/Truncate/examples/Truncate.md
```diff
@@ -14,27 +14,55 @@ The truncate component contains two child elements, `.pf-v5-c-truncate__start` a
...
+  {{/truncate}}
+</div>
+```
+
+### File extension middle
+```hbs
+<div class="pf-v5-c-truncate--example">
+  {{#> truncate truncate--id='middle-of-line-truncation-file-extension-example'}}
+    {{> truncate-start truncate-start--text='redhat_logo_black_and_white_reversed'}}
+    {{> truncate-end truncate-end--text='white_reversed' truncate-file-extension--text='zip'}}
```
thatblindgeye, line 56: We'll just need to remove one of the "white_reversed" instances, believe it'd be this one:

```suggestion
    {{> truncate-start truncate-start--text='redhat_logo_black_and_'}}
    {{> truncate-end truncate-end--text='white_reversed' truncate-file-extension--text='zip'}}
```

thatblindgeye (commented):
(possibly) last comment below. Would be worth getting other input on the new file extension examples; it might be because they're in English that they may be positioned oddly (mainly the "file extension middle" example)

src/patternfly/components/Truncate/truncate.scss
```diff
@@ -23,12 +23,21 @@
...
-  direction: ltr;
+
+  @include pf-v5-bidirectional-style {
+    $prop: direction;
+    $ltr-value: ltr;
+    $rtl-value: rtl;
+  }
 }
 
 // safari not supported
```
mcoker, line 43: wdyt about also updating this bidir style for the safari override?

src/patternfly/components/Truncate/truncate-start.hbs
```diff
@@ -9,3 +9,6 @@
     {{> @partial-block}}
   {{/if}}
 </span>
+{{#if truncate-file-extension--text}}
+  {{removeWhiteSpaceForPrettier}}{{> truncate-file-extension}}
+{{/if}}
```
mcoker, line 14:
```suggestion
```

src/patternfly/components/Truncate/truncate-file-extension.hbs
```diff
@@ -0,0 +1,10 @@
+<span class="{{pfv}}truncate__file-extension{{#if truncate-file-extension--modifier}} {{truncate-file-extension--modifier}}{{/if}}"
```
mcoker, line 1: I think you can delete this file

src/patternfly/components/Truncate/truncate-end.hbs
```diff
@@ -9,3 +13,6 @@
     {{> @partial-block}}
   {{/if}}
 </span>
+{{#if truncate-file-extension--text}}
+  {{removeWhiteSpaceForPrettier}}{{> truncate-file-extension}}
+{{/if}}
```
mcoker, line 18:
```suggestion
```

src/patternfly/components/Truncate/truncate-end.hbs
```diff
@@ -1,4 +1,8 @@
-<span class="{{pfv}}truncate__end{{#if truncate-end--modifier}} {{truncate-end--modifier}}{{/if}}"
+<span class="{{pfv}}truncate__end
+  {{#if isFileExtension}}
+    pf-m-file-extension
+  {{/if}}
+  {{#if truncate-end--modifier}} {{truncate-end--modifier}}{{/if}}"
```
mcoker, line 5:
```suggestion
<span class="{{pfv}}truncate__end{{#if truncate-end--modifier}} {{truncate-end--modifier}}{{/if}}"  
```

mcoker (changes requested):

mcoker (approved):
LGTM!

src/patternfly/components/BackgroundImage/background-image.scss
```diff
@@ -18,6 +18,11 @@
   background-color: var(--#{$background-image}--BackgroundColor);
   background-image: var(--#{$background-image}--BackgroundImage);
   background-repeat: no-repeat;
-  background-position: var(--#{$background-image}--BackgroundPosition);
   background-size: var(--#{$background-image}--BackgroundSize);
+
+  @include pf-v5-bidirectional-style(
+    $prop: background-position,
+    $ltr-val: var(--#{$background-image}--BackgroundPosition),
+    $rtl-val: bottom left
```
mcoker, line 26: I wonder if we should use this on the `--#{$background-image}--BackgroundPosition` declaration instead? This update opts you out of the var to control the background position. On this line https://github.com/patternfly/patternfly/blob/b985207b7f7b861d9b9301d22d3f7d32ee6b2465/src/patternfly/components/BackgroundImage/background-image.scss#L6


we'd update to
```
@include pf-v5-bidirectional-style(
    $prop: --#{$background-image}--BackgroundPosition,
    $ltr-val: bottom right,
    $rtl-val: bottom left
```

wdyt @srambach @thatblindgeye

thatblindgeye: That could make sense. Either way there's always the chance it may not work as intended for a consumer.

- If we leave it as the PR currently is, the background image may be covered by some other element(s) when the p background position is updated
- If we update the var instead of the property, it may prevent the background image from being positioned based on RTL, which could cause it to be covered by other elements that are swapped to the other side of the screen

Updating the var may be the better guesstimate for us to make; at that point it'd be on the consumer to ensure they're customizing the var correctly for LTR and RTL.

thatblindgeye (approved):

thatblindgeye (approved):

Issue #5850, opened 2023-08-16
Update image used in horizontal card demo


mcoker:
https://www.patternfly.org/components/card/html-demos#horizontal-split

Update to a lightweight, PF branded image - SVG if possible.

PR #5875, opened 2023-08-30
chore(card): updated horiz demo image

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5850

should help with the screenshot tool generating false positives since the screenshot sometimes fires before the image has finished downloading


srambach (approved):
LGTM!

PR #5876, opened 2023-08-30
fix(clipboard-copy): added rtl styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5832

### DIFF
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy.scss
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy.scss
@@ -66,6 +66,8 @@
 }
 
 .#{$clipboard-copy}__toggle-icon {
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):
Looks good!

Just for context for future reference, based on the convo and testing during our group Meet, we're expecting the icon and text to "swap" in the [inline in a sentence example](https://patternfly-pr-5876.surge.sh/components/clipboard-copy#inline-compact-in-sentence) when the text is in an RTL based language.

srambach:
FWIW, this also exhibits the little flip of the caret caused by the transition timing for the rotation applying to the mirroring of the icon. I'm good with going ahead as is and finding a fix for it as a follow up.

PR #5877, opened 2023-08-30
fix: added tabs and jump links rtl styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5824


src/patternfly/components/Tabs/tabs.scss
```diff
@@ -700,6 +700,8 @@ $pf-v5-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
 }
 
 .#{$tabs}__link-toggle-icon {
+  @include pf-v5-mirror-inline-on-rtl;
```
thatblindgeye, line 703: We'll also need this on line ~493 for the expandable examples.

thatblindgeye (changes requested):
Also noticing that we'll need to update some styling in the org repo, as the `.ws-toc` styling there is using shorthand margin and padding which causes an issue with the jump links on component pages.

LTR:

image


RTL:

image

I'll open an issue for this in Org

thatblindgeye (approved):

Issue #5829, opened 2023-08-11
Implement RTL: tooltip and popover


srambach:

thatblindgeye:
We could also investigate whether the classnames being applied (e.g. pf-m-top, pf-m-left, etc) should be updated in a breaking change to more logical naming.

PR #5882, opened 2023-08-31
fix(popover, tooltip): added rtl styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5829

Adds support for the top/left/right/bottom/etc modifiers as logical equivalents since those are a little more clear. Uses `:is()` so it should actually decrease the size of those stylesheets with the additional modifiers. Also adds support for `.pf-m-text-align-start` on tooltip content.

I think we'll want these modifiers eventually, though we don't have to add them, or add prop vals for them in react as part of this release, but we could. And if we keep them, we could also update our examples to use them by default, as the preferred way to set the popover/tooltip direction. Thoughts?

### DIFF
--- a/src/patternfly/components/Popover/examples/Popover.css
+++ b/src/patternfly/components/Popover/examples/Popover.css
@@ -1,10 +1,4 @@
 .ws-core-c-popover .ws-preview-html {
-  position: relative;
- }
-
- .ws-core-c-popover .pf-v5-c-popover {
-   position: absolute;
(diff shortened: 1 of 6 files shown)


thatblindgeye (approved):
🔥 

Personally I feel like it could be worth updating things to use the more logical-property named classes at some point. In RTL the `pf-m-right` popover/tooltip would look like the `pf-m-left` one in LTR, while `inline-end` feels more consistent between both?

PR #5883, opened 2023-08-31
fix(tree-view): added rtl styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5822


src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -444,6 +446,12 @@ $pf-v5-c-tree-view--MaxNesting: 10 !default;
   background-color: var(--#{$tree-view}__node-toggle--BackgroundColor);
   border: 0;
   transform: translateX(var(--#{$tree-view}__list-item__list-item__node-toggle--TranslateX));
```
srambach, line 448: This line can be removed since you have the conditional one below, right?

srambach (approved):
👍

PR #5889, opened 2023-08-31
fix(forms): updated rtl styles

### LINKED ISSUE #5811: Implement RTL: Form, Text input group, input group, form control, checkbox, radio, number input

mattnolting:
closes #5811

### DIFF
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -311,6 +311,16 @@ $pf-v5-c-form--m-horizontal--breakpoint-map: build-breakpoint-map("sm", "md", "l
   border: 0;
   transform: translateY(var(--#{$form}__group-label-help--TranslateY));
 
+  @include pf-v5-bidirectional-style(
+    $prop: display,
+    $ltr-val: inline,
+    $rtl-val: inline-flex,
+  );
+
+  @include pf-v5-rtl {
+    --#{$form}__group-label-help--TranslateY: 0;
+  };
+
   &:hover {
     --#{$form}__group-label-help--Color: var(--#{$form}__group-label-help--hover--Color);
   }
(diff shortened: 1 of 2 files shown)


thatblindgeye (changes requested):
- `.pf-v5-c-form__field-group-toggle-icon` will just need the mirror inline mixin added to mirror the expandable icon.
- The [autocomplete example for Text Input Group](https://patternfly-pr-5889.surge.sh/components/text-input-group#autocomplete-last-option-hint) has the highlighted text positioned incorrectly with RTL active, though wondering if this is more due to it being in English than anything else:
    image
- This would probably be more an example update in Core and more a component update in React, but the right arrow icon for the [Search Input Group examples](https://patternfly-pr-5889.surge.sh/components/text-input-group#search-input-group-with-submit-button) should be reversed for RTL

Agreed on the other components not needing further update

mcoker (approved):
💪🎃

mcoker:
@thatblindgeye

> The [autocomplete example for Text Input Group](https://patternfly-pr-5889.surge.sh/components/text-input-group#autocomplete-last-option-hint) has the highlighted text positioned incorrectly with RTL active, though wondering if this is more due to it being in English than anything else:

Tested this in RTL and it behaves mostly as we'd expect.... except, at least in arabic anyways, the last letter(s) of a string get some extra embellishment, so it doesn't really work well? Here I've typed a partial and full word in each text input and separated them so you can see how the end of the word looks different based on whether it's the partial or full word

Screenshot 2023-08-31 at 4 14 32 PM


> This would probably be more an example update in Core and more a component update in React, but the right arrow icon for the [Search Input Group examples](https://patternfly-pr-5889.surge.sh/components/text-input-group#search-input-group-with-submit-button) should be reversed for RTL

Agree, unless we had a class or something to target that input group item as the search button, I don't think we want to put it in the CSS. We can just use the `.pf-v5-m-mirror-on-rtl` helper class here https://github.com/patternfly/patternfly-react/blob/52cbf11e7e407d70d674a160da7f432d197c5f57/packages/react-core/src/components/SearchInput/SearchInput.tsx#L386

PR #5897, opened 2023-09-05
chore(Card): remove extraneous class

### LINKED ISSUE #5896: Card demo - extraneous class modifier
In the card demos, `pf-v5-m-mirror-inline-rtl` was added but isn't needed.

srambach:
Fixes #5896


srambach:
Mistake, not needed.

PR #5903, opened 2023-09-05
fix(nav): updated example css

### LINKED ISSUE #5899: RTL visual regression - navigation and tabs
Part of https://github.com/patternfly/patternfly/issues/5898

Navigation - there was an update to the local example CSS to style the nav background, and doesn't look like it's working as expected

Tabs - seems to be a spacing issue around tabs with help/popover buttons

mattnolting:
Fixes #5899 

• Looks like the [third level expandable example](https://www.patternfly.org/components/navigation/html#expandable-third-level) had an [explicit background color set](https://github.com/patternfly/patternfly/commit/447630c3449372332ee1628268397f16f5ffa965#diff-c0c17bbec61392998ccf9e8532dc141d0f007e6bc152355119b66317925f671cR1). I'm not entirely sure how backstop is drawing and comparing examples, but it doesn't seem to be appearing in localhost. Added the example css declaration back to navigation.css.

• Tabs are fine. Backstop seems to be throwing an error bc of scroll position. 

**.org**
Screenshot 2023-09-05 at 3 01 34 PM

**localhost**
Screenshot 2023-09-05 at 3 01 41 PM

### DIFF
--- a/src/patternfly/components/Nav/examples/Navigation.css
+++ b/src/patternfly/components/Nav/examples/Navigation.css
@@ -1,4 +1,4 @@
-.ws-core-c-navigation:not([id*="tertiary"]):not([id*="light-mode"]) .ws-preview-html {
+.ws-core-c-navigation:not([id*="tertiary"]):not([id*="light-mode"]) .pf-v5-c-nav{
   background-color: var(--pf-v5-global--BackgroundColor--dark-300);
 }
 


src/patternfly/components/Nav/examples/Navigation.css
```diff
@@ -1,4 +1,5 @@
-.ws-core-c-navigation:not([id*="tertiary"]):not([id*="light-mode"]) .ws-preview-html {
+.ws-core-c-navigation:not([id*="tertiary"]):not([id*="light-mode"]) .ws-preview-html,
+#ws-core-c-navigation-expandable-third-level .pf-v5-c-nav {
```
mcoker, line 2: Nice! I overlooked those selectors. Looks like there are 2 more we may need to add? https://github.com/patternfly/patternfly/commit/447630c3449372332ee1628268397f16f5ffa965

However, I don't see why we're adding those that way... all of the navs that aren't targeting `.pf-v5-c-nav`, if the viewport is shorter than the nav height, if you scroll down, [the background is missing](https://www.patternfly.org/components/navigation/html/expanded/). I wonder if we could just update this block one of a couple of ways https://github.com/patternfly/patternfly/blob/825d0af0912528722280460f17c82acf5b66c241/src/patternfly/components/Nav/examples/Navigation.css#L1-L3

Either update the selector to target all nav elements:
```
.ws-core-c-navigation:not([id*="tertiary"]):not([id*="light-mode"]) .pf-v5-c-nav
```

Or nix the `pf-v5-u-h-100` utility class on those blocks with an override like:
```
 .ws-core-c-navigation:not([id*="tertiary"]):not([id*="light-mode"]) .ws-preview-html {
   height: auto !important;
   background-color: var(--pf-v5-global--BackgroundColor--dark-300); 
 } 
```

Just an aside, I don't think we need `pf-v5-u-h-100` anymore. That shouldn't be necessary since we added the 100vh default height to `.pf-c-page`. We can also nix the `height: 100%` from all of its ancestors (`#root, #ws-router, #ws-page`)

Screenshot 2023-09-05 at 2 33 25 PM

mcoker (commented):
Left a comment - not a blocker, though I do think if we're going to keep the approach in this PR, we should add the other two selectors that look like they also had this style previously. Curious your thoughts on the other approaches. I don't think we can take the h-100 utility class of the example block now, though we could do that as a separate issue later.

mcoker (approved):

Issue #5898, opened 2023-09-05
RTL visual regressions


mcoker:
PDF of the visual regressions -> [BackstopJS Report.pdf](https://github.com/patternfly/patternfly/files/12526744/BackstopJS.Report.pdf)

```[tasklist]
### Tasks
- [ ] https://github.com/patternfly/patternfly/issues/5899
- [ ] https://github.com/patternfly/patternfly/issues/5900
- [ ] https://github.com/patternfly/patternfly/issues/5901
```

PR #5915, opened 2023-09-06
chore: update screenshots

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5898

Failures from this run: [backstop-fail.pdf](https://github.com/patternfly/patternfly/files/12543137/backstop-fail.pdf)

* The first image with the card text is a false positive
* The split cards with backgrounds images are expected, the image was updated to a lightweight SVG
* All of the alerts with the line misplaced look like issues from the previous run, the images from this PR look correct
* Text rendering in the modals are false positives
* Horizontal nav differences are with the scroll position at different viewports
* Same ^ for the tabs
* All of the nav examples are to be expected, they were updated to apply a background to the nav, not the parent example block
* Popper is a false positive, probably something with the way we changed the position of the example to use flex like the bullseye, not custom absolute positioning
* Skeleton is to be expected, the examples were updated to remove an unnecessary nested skeleton that changed one of the shapes.
* Pagination is to be expected, we fixed the markup in the toolbar demo pagination and it corrected the spacing
* Tooltips are to be expected, padding was added around the examples to show the little pointer/caret
* Truncate examples are to be expected, we updated the example text
* Flex examples are to be expected, I ran this report against the prod build on port 5000 - the previous report was against the dev build on port 8001, and the examples with a `.` in the example name don't seem to load in puppeteer on port 8001. Below are the URLs that fail on port 8001
  * http://localhost:5000/layouts/flex/html/aligning-nested-columns-with-.pf-m-align-self-baseline-example/
  * http://localhost:5000/layouts/flex/html/aligning-nested-columns-with-.pf-m-align-self-center-example/
  * http://localhost:5000/layouts/flex/html/aligning-nested-columns-with-.pf-m-align-self-stretch-example/
  * http://localhost:5000/layouts/flex/html/alignment-using-.pf-m-align-self-flex-end-example/


mattnolting (approved):
LPTM 👍

PR #5923, opened 2023-09-12
fix(progress): updated indicator direction

### LINKED ISSUE #5919: Bug - Progress bar - bar should progress LTR even in RTL
**Describe the problem**
Even in RTL, because the bar is showing time and tracking with numbers, the bar on the progress bar should fill in from left to right.
The title should remain on the right in RTL.
The number% should remain as is (on the left) in RTL.

mattnolting:
closes #5919

### DIFF
--- a/src/patternfly/components/Progress/progress.scss
+++ b/src/patternfly/components/Progress/progress.scss
@@ -186,8 +186,8 @@
 
   &::before {
     position: absolute;
-   inset-block-start: 0;
-   inset-inline-start: 0;
+    inset-block-start: 0;
+    inset-inline-start: 0;
     width: 100%;
     height: 100%;
     content: "";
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Progress/progress.scss
```diff
@@ -200,10 +200,11 @@
 // It is assumed that the width of the progress__indicator is set inline or via Javascript
 .#{$progress}__indicator {
   position: absolute; // position absolute to bring above the semi-transparent bar overlay
- inset-block-start: 0;
- inset-inline-start: 0;
+  inset-block-start: 0;
+  inset-inline-start: 0;
```
mcoker, line 204: If this doesn't need to flip, we can just use `top: 0; left: 0;` and we won't need to add `direction`. Or you could just use `inset: 0;` since the inline `width` will limit the width. That's probably preferred because I think the linter will currently complain if you use a non-logical property.

src/patternfly/components/Progress/progress.scss
```diff
@@ -186,8 +186,8 @@
 
   &::before {
     position: absolute;
-   inset-block-start: 0;
-   inset-inline-start: 0;
+    inset-block-start: 0;
+    inset-inline-start: 0;
     width: 100%;
     height: 100%;
```
mcoker, line 192: you could replace all of this with `inset: 0;`

mcoker (commented):
This works but seems like we could just not use logical properties over forcing `direction` - wdyt?

thatblindgeye (approved):

Issue #5927, opened 2023-09-12
Code editor - determine layout in RTL


mcoker:
Looking at a bunch of coding tutorials in other languages, the applications used have a LTR interface and all of the code is LTR, so I'm thinking it should be LTR by default?

PR #5928, opened 2023-09-12
fix(code-editor): make code box and header LTR

mcoker:
This updates the `__code` element to use `dir="ltr"` so it will always be LTR. Everything else adapts to RTL as normal.

fixes https://github.com/patternfly/patternfly/issues/5927

### DIFF
--- a/src/patternfly/components/CodeEditor/code-editor-code.hbs
+++ b/src/patternfly/components/CodeEditor/code-editor-code.hbs
@@ -1,4 +1,4 @@
-<code class="{{pfv}}code-editor__code{{#if code-editor-code--modifier}} {{code-editor-code--modifier}}{{/if}}"
+<code dir="ltr" class="{{pfv}}code-editor__code{{#if code-editor-code--modifier}} {{code-editor-code--modifier}}{{/if}}"
   {{#if code-editor-code--attribute}}
     {{{code-editor-code--attribute}}}
   {{/if}}>
--- a/src/patternfly/components/CodeEditor/code-editor.scss
+++ b/src/patternfly/components/CodeEditor/code-editor.scss
@@ -135,6 +135,11 @@
   // stylelint-enable selector-class-pattern
 }
 
+.#{$code-editor}__code,
+.#{$code-editor}__header {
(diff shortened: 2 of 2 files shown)


wise-king-sullyman (approved):

thatblindgeye (approved):
Looks good, just a couple of comments:

- I could see reasoning for allowing the header to be affected by RTL, and only making the `__code` element always kept LTR. Not a blocker as we can see what feedback we get if anything.
- We could add a `dir` attribute to examples instead to avoid using CSS to dictate the direction, though with this we don't have to worry about consumers not doing so themselves for the HTML library (with React it could at least be hardcoded into the component). Not a blocker either.

mcoker:
> I could see reasoning for allowing the header to be affected by RTL, and only making the __code element always kept LTR. Not a blocker as we can see what feedback we get if anything.

I agree. That would certainly be simpler. For the sake of keeping it simple, if we don't have a strong preference one way or another, I'd be in favor of not trying to force LTR styling in the header. wdyt?

mcoker:
> We could add a `dir` attribute to examples instead to avoid using CSS to dictate the direction, though with this we don't have to worry about consumers not doing so themselves for the HTML library (with React it could at least be hardcoded into the component). Not a blocker either.

Agree - when I created the LTR/RTL classes, I wasn't keeping in mind that you can apply `dir` to elements in the page. It seems like in most cases, `dir` should be used. The only downside is that if we use the `dir` attribute (markup change) is that if we change that value, react users should be good since we can update the react component, but core users won't get the change without a markup change - so we need to continue to support the previous `dir` value for core users, otherwise it may be a breaking change.

thatblindgeye:
>  I'd be in favor of not trying to force LTR styling in the header. wdyt?

I'd be good with that!

And good point re: the dir attribute. I think the styling vs markup topic happened before with RTL (though maybe briefly?), but either way using the styling here isn't a blocker for me.

tlabaj (approved):

thatblindgeye (approved):

srambach (approved):
!tcefreP

PR #5931, opened 2023-09-13
feat(menu-toggle): add in work from hackathon

srambach:
Menu toggle plus some nits from previous PRs for hackathon stuff


mcoker (approved):
LGTM, just one bit about some extra vars and icon color stuff, but we can follow up if needed.

Issue #5958, opened 2023-09-21
Update screenshots


mcoker:
With the q3 milestone issues wrapped up, let's update the screenshots and give a sanity check that nothing unexpected changed.

Issue #5957, opened 2023-09-21
Badge - icon/toggle icon support


mcoker:
A badge is sometimes used as a the toggle for a dropdown menu (in breadcrumbs, when multiple breadcrumb items are collapsed to a menu - the badge represents the number of collapsed items), and there is a small toggle/caret icon inside of the badge that is basically unstyled (no support for an icon/toggle icon in the badge). In core, the arrow is too close to the badge text. It looks better in react because the SVG created by react has a little bit of whitespace around it that the core icon does not. The space between the badge text and the toggle icon just comes from a whitespace character in the HTML.

Core:
<img width="272" alt="Screenshot 2023-09-21 at 5 17 00 PM" src="https://github.com/patternfly/patternfly/assets/35148959/3a0d8f31-c6f5-45a5-9342-af17fc64d04c">

React:
<img width="374" alt="Screenshot 2023-09-21 at 5 18 41 PM" src="https://github.com/patternfly/patternfly/assets/35148959/4ce9d88d-3f47-4c48-9287-c45eefc54ea5">

It seems like if we want to continue to use the menu toggle with this kind of badge inside of it, we should probably add support for either an icon and/or a toggle icon that can hold the icon and add a consistent spacer there between core and react. Either that or add support to the menu toggle that will create a badge and toggle icon, and the toggle icon can come from the menu toggle component. If we needed to do something like control the toggle/caret icon color based on whether the menu toggle is hovered/focused/expanded, we would probably want that toggle icon to come from the menu toggle component instead of the badge.

At any rate, here's what the badge would look like with a specific spacer for that icon:

Core:
xs spacer
<img width="257" alt="Screenshot 2023-09-21 at 5 24 47 PM" src="https://github.com/patternfly/patternfly/assets/35148959/3d0b17a3-dcf1-44fa-a4dc-afa6795cc7a4">

sm spacer
<img width="282" alt="Screenshot 2023-09-21 at 5 25 00 PM" src="https://github.com/patternfly/patternfly/assets/35148959/1716045d-59ae-4dcf-b151-52c44097b77e">

React:
xs spacer
<img width="372" alt="Screenshot 2023-09-21 at 5 25 40 PM" src="https://github.com/patternfly/patternfly/assets/35148959/5804e992-50ad-4b4d-b10b-66d6837fb1c3">

sm spacer
[...]

Issue #5945, opened 2023-09-18
Toolbar - mirror expand all icon in RTL


mcoker:
Follow up to:
* https://github.com/patternfly/patternfly/pull/5937
* https://github.com/patternfly/patternfly-react/pull/9605

And should ideally be done with https://github.com/patternfly/patternfly-react/issues/9641 during a breaking change release.

PR #5959, opened 2023-09-21
chore(screenshots): update visual regression screenshots

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5958

Failure report: [BackstopJS Report.pdf](https://github.com/patternfly/patternfly/files/12693937/BackstopJS.Report.pdf)

* Buttons - we updated the plain and added the "no padding plain" examples
* Clipboard copy - we added a "Code" example that matches react, and serves as an RTL example since that code should not change directions
* Data list - the "expand all" icon in the toolbar changed width a bit because it's now wrapped in the icon component so we can flip it in RTL. Just an example change since it's our formal recommendation until we can update the toolbar to flip that icon for us automatically. Here's the issue to do that - https://github.com/patternfly/patternfly/issues/5945
* Drilldown menu breadcrumb/badge - spacing difference from updating that badge to be in the menu toggle instead of the dropdown component. Our code matches the react code, but our icon looks a little tight there as it is. Created a follow up issue to look at that - https://github.com/patternfly/patternfly/issues/5957
* Horizontal nav - just a scroll position difference, not sure why that happens, but it isn't an issue
* Tabs - ^same


thatblindgeye (approved):

srambach (approved):
🚀 🚀

PR #5960, opened 2023-09-22
chore(screenshots): updated full page screenshots

mcoker:
@nicolethoen @wise-king-sullyman @tlabaj I ran the full-page screenshot update for core, but ran into a couple of issues

~~* The RTL button in the bottom/right of full page examples is showing up in the screenshots and shouldn't since that's a workspace specific setting, and not enabled on org~~
* I changed the padding around the switches in the bottom/right

Those changes are marking every full page screenshot we have as updated/different. I reverted both of those changes above for these screenshots so I could get a cleaner diff to see if anything unexpected changed, and all looks good. If we merge this PR, I'll back out the padding change at the very least.

~~Re: the RTL button in the screenshots, how should we proceed there? I could see doing this a few ways:~~

* ~~Core (and react, I'm assuming) wouldn't run this screenshot job, and org would run it instead. That may be the best since it could just be part of the release process, which is when this job is run anyways.~~
* ~~Org could run the job and those screenshots show up on org, but core/react could run it, too, and when you're in a core/react workspace, the core/react screenshots show up (a more accurate screenshot for the.~~
* ~~Core/react could run the job and could have some special CSS (or pull in the org workspace settings) so we create screenshots that look appropriate for showing up on org.~~

Update - org takes their own screenshots, so these screenshots are fine with the RTL button in them.


nicolethoen:
@mcoker these screenshots will only show up in teh core workspace, not on org. so it shouldn't be a problem that the RTL switcher is in these screenshots

PR #5962, opened 2023-09-25
fix(Toolbar): increase workspace container height

### LINKED ISSUE #5772: Bug - Toolbar - example needs min-height
**Describe the problem**
The toolbar example https://pf5.patternfly.org/components/toolbar/html-demos#toolbar-attribute-value-single-select-filter-on-desktop needs to have the min-height applied because the open dropdown is overlapping the content below.

image

srambach:
Fixes #5772 
Went ahead and did it since stalebot was on it's tail!
Adds min height for the 4th toolbar demo.

### DIFF
--- a/src/patternfly/demos/Toolbar/examples/Toolbar.css
+++ b/src/patternfly/demos/Toolbar/examples/Toolbar.css
@@ -5,6 +5,10 @@
   min-height: 420px;
 }
 
+#ws-html-demos-c-toolbar-toolbar-attribute-value-single-select-filter-on-desktop {
+  min-height: 320px;
+}
+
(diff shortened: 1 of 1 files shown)


thatblindgeye (commented):
Should we also update "Toolbar attribute value checkbox select filter on desktop":

image

mcoker (approved):
LGTM tho agree with @thatblindgeye about the [attr value select filter example](https://patternfly-pr-5962.surge.sh/components/toolbar/html-demos/#toolbar-attribute-value-checkbox-select-filter-on-desktop)

thatblindgeye (approved):

PR #5965, opened 2023-09-25
feat(table): apply tokens from hackathon

srambach:
Tokens on table; minimal changes to toolbar.
Adds shadow tokens that will need to be broken out because they'll come from Figma styles (not variables) for now.
Next up will be to straighten out the token files (and update) but for now I've added them into the section with the font styles.

### DIFF
--- a/src/patternfly/base/tokens/_tokens-font.scss
+++ b/src/patternfly/base/tokens/_tokens-font.scss
@@ -62,4 +62,33 @@
   --pf-t--global--border--width--control--default: var(--pf-t--global--border--width--100);
   --pf-t--global--border--width--control--hover: var(--pf-t--global--border--width--100);
   --pf-t--global--border--width--control--active: var(--pf-t--global--border--width--200);
+
+  // box-shadow
+  --pf-t--global--box-shadow--sm: 0px, 1px, 2px, 0px rgba(0, 0, 0, 0,0,0, 16%);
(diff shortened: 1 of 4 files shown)


src/patternfly/base/tokens/_tokens-font.scss
```diff
@@ -62,4 +62,33 @@
   --pf-t--global--border--width--control--default: var(--pf-t--global--border--width--100);
   --pf-t--global--border--width--control--hover: var(--pf-t--global--border--width--100);
   --pf-t--global--border--width--control--active: var(--pf-t--global--border--width--200);
+
+  // box-shadow
+  --pf-t--global--box-shadow--sm: 0px, 1px, 2px, 0px rgba(0, 0, 0, 0,0,0, 16%);
+  --pf-t--global--box-shadow--sm--top: 0px, -1px, 2px, 0px rgba(0, 0, 0, 16%);
```
mcoker, line 68: This may be known since you're still working on these, but we need to remove the commas between the values and it's probably a good time to go ahead and switch to `rgb()` since it supports the alpha channel and `rgba()` is deprecated (though it still works fine). 

```suggestion
  --pf-t--global--box-shadow--sm--top: 0px -1px 2px 0px rgb(0 0 0 / 16%);
```

mcoker (approved):
LGTM at a glance - left a note about box-shadow but that may be something for later?

mcoker (approved):

PR #5969, opened 2023-10-02
fix(check-radio): added aria-labelledby to check/radio

### LINKED ISSUE #5803: Bug - Checkbox/Radio - Description/paragraph text not accessible to screenreaders
**Describe the problem**
Checkboxes that include additional text below the label are not implemented in a way that the additional text is announced via screenreaders. For example: https://www.patternfly.org/components/forms/checkbox#checkbox-with-description

1. Navigate to https://www.patternfly.org/components/forms/checkbox/html
2. Open the HTML view 
3. The `<input>` element for the checkbox does not include `aria-describedby` referencing the additional description text provided for the checkbox

Alternatively, if you use a screenreader to navigate the form, and Tab through the form elements, focus would land on the checkbox and only announce the checkbox label.

**Expected behavior**
An `aria-describedby` is provided for the checkbox input so that when it receives focus, a screenreader will announce the checkbox label, followed by the additional text.

mattnolting:
closes #5803

### DIFF
--- a/src/patternfly/components/Check/check-input.hbs
+++ b/src/patternfly/components/Check/check-input.hbs
@@ -3,8 +3,10 @@
   {{#if check-description--text}}
     aria-describedby="{{check--id}}-description"
   {{/if}}
-  id="{{check--id}}-input"
-  name="{{check--id}}-input"
+  {{#if check--id}}
+    id="{{check--id}}-input"
+    name="{{check--id}}-input"
+  {{/if}}
   {{#if check--IsStandalone}}
     aria-label="{{#if check--standalone--label}}{{check--standalone--label}}{{else}}Standalone check{{/if}}"
   {{/if}}
--- a/src/patternfly/components/Check/check-input.hbs
+++ b/src/patternfly/components/Check/check-input.hbs
@@ -1,18 +1,10 @@
 <input class="{{pfv}}check__input{{#if check-input--modifier}} {{check-input--modifier}}{{/if}}"
   type="checkbox"
-  {{#ifAny check-description--text check--HasDescription check-body--text check--HasBodyText}}
-    aria-describedby="
-    {{#ifAny check-description--text check--HasDescription}}
-      {{check--id}}-description
-    {{/ifAny}}
-    {{#ifAny check-body--text check--HasBodyText}}
-      {{check--id}}-body
-    {{/ifAny}}"
-  {{/ifAny}}
-  {{#if check--id}}
-    id="{{check--id}}-input"
-    name="{{check--id}}-input"
+  {{#if check-description--text}}
+    aria-describedby="{{check--id}}-description"
   {{/if}}
+  id="{{check--id}}-input"
+  name="{{check--id}}-input"
   {{#if check--IsStandalone}}
     aria-label="{{#if check--standalone--label}}{{check--standalone--label}}{{else}}Standalone check{{/if}}"
   {{/if}}
--- a/src/patternfly/components/Check/check-input.hbs
+++ b/src/patternfly/components/Check/check-input.hbs
@@ -1,5 +1,21 @@
 <input class="{{pfv}}check__input{{#if check-input--modifier}} {{check-input--modifier}}{{/if}}"
   type="checkbox"
+  {{#ifAny check-description--text check--HasDescription check-body--text check--HasBodyText}}
+    aria-describedby="
+    {{#ifAny check-description--text check--HasDescription}}
+      {{check--id}}-description
+    {{/ifAny}}
+    {{#ifAny check-body--text check--HasBodyText}}
+      {{check--id}}-body
+    {{/ifAny}}"
+  {{/ifAny}}
+  {{#if check--id}}
+    id="{{check--id}}-input"
+    name="{{check--id}}-input"
+  {{/if}}
+  {{#if check--IsStandalone}}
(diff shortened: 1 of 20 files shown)


src/patternfly/components/Check/examples/Check.md
```diff
@@ -60,34 +60,28 @@ cssPrefix: pf-v5-c-check
 
 ### With description
 ```hbs
-{{#> check}}
-  {{#> check-input check-input--attribute='id="check-with-description" name="check-with-description"'}}{{/check-input}}
-  {{#> check-label check-label--attribute='for="check-with-description"'}}Check with description{{/check-label}}
-  {{#> check-description}}
-    Single-tenant cloud service hosted and managed by Red Hat that offers high-availability enterprise-grade clusters in a virtual private cloud on AWS od GCP.
-  {{/check-description}}
+{{#> check check--id="check-with-description" check--HasDescription=true}}
```
mcoker, line 63: Is it feasible to update this to

```
{{> check
  check--id="check-with-description"
  check-description--text="Single-tenant cloud service hosted and managed by Red Hat that offers high-availability enterprise-grade clusters in a virtual private cloud on AWS od GCP."
  check-label--text="Check with description"}}
```

and instead of having to specify `check--HasDescription`, just test for `check-description--text`? I have that question for dropping `check--HasBodyText` and testing for `check-body--text`.

src/patternfly/components/Check/check-input.hbs
```diff
@@ -1,5 +1,21 @@
 <input class="{{pfv}}check__input{{#if check-input--modifier}} {{check-input--modifier}}{{/if}}"
   type="checkbox"
+  {{#ifAny check-description--text check--HasDescription check-body--text check--HasBodyText}}
+    aria-describedby="
+    {{#ifAny check-description--text check--HasDescription}}
+      {{check--id}}-description
+    {{/ifAny}}
+    {{#ifAny check-body--text check--HasBodyText}}
+      {{check--id}}-body
+    {{/ifAny}}"
```
mcoker, line 10: I wonder if we should...
1. Only have `aria-describedby` for either description if it exists, or a body if there is no description - instead of both.
2. Only use `aria-describedby` for the description, and not the body. The description should be something short/sweet and an actual description. The body could be a bunch of nested form controls or who knows what.

src/patternfly/components/Check/examples/Check.md
```diff
@@ -8,97 +8,62 @@ cssPrefix: pf-v5-c-check
...
+{{> check check-input--id="check-checked" check-label--text="Check checked"}}
 ```
 
 ### Label wrapping input
 ```hbs
-{{#> check check--type="label" check--attribute='for="check-label-wrapping-input"'}}
-  {{#> check-input check-input--attribute='id="check-label-wrapping-input" name="check-label-wrapping-input"'}}{{/check-input}}
-  {{#> check-label check-label--type="span"}}Check label wraps input{{/check-label}}
-{{/check}}
+{{> check check-input--id="check-label-wrapping-input" check-label--text="Check label wraps input" check--IsLabelWrapped=true}}
```
mcoker, line 26: This example is generating `id="-input"` on the parent check container since that just looks for `check--id` but you're setting `check-input--id` on all of the examples https://github.com/patternfly/patternfly/blob/99476bba3b3433847d6c68f83dffb9a943b29803/src/patternfly/components/Check/check.hbs#L8-L10

Looks like the radio `isLabelWrapped` example works because you're setting `radio--id` on that specific example instead of `radio-input--id`

I'm not sure I understand the id-helper hbs entirely, but at a glance I'd probably expect setting the check--id on the outer container and being able to generate all of the inner id's from check--id? If I get the idea of the helper, I wonder if this would work:

```
{{> check check--id="some-id"}}

// check.hbs
{{#> check--context check-input--id=(ternary check-input--id (concat check-input--id "-input") (concat check--id "-input)}}
  // all the existing stuff in check.hbs
{{/check-context}}
```

mcoker (changes requested):
Just the one ID is off. I left a comment, but looks like it may be as simple as changing `check-input--id` to `check--id` on that one example

src/patternfly/components/Check/examples/Check.md
mattnolting: @mcoker Since this is a workspace issue, why don't we create a followup?

src/patternfly/components/Check/examples/Check.md
```diff
@@ -8,97 +8,62 @@ cssPrefix: pf-v5-c-check
 ## Examples
 ### Basic
 ```hbs
-{{#> check}}
-  {{#> check-input check-input--attribute='id="check-basic" name="check-basic"'}}{{/check-input}}
-  {{#> check-label check-label--attribute='for="check-basic"'}}Check{{/check-label}}
-{{/check}}
+{{> check check--id="basic-example-example" check-label--text="Check"}}
```
mcoker, line 11: Did you mean to add an extra "example"?

```suggestion
{{> check check--id="basic-example" check-label--text="Check"}}
```

mcoker (approved):
[Backstop report](https://github.com/patternfly/patternfly/files/12836036/check-radio-backstop.pdf)

Looks good, but the disabled radio example wasn't rendering and the param used to check a radio wasn't working. Pushed an update for those two

PR #5980, opened 2023-10-09
chore(tokens): makes prefix for token version

srambach:
This adds a prefix for the tokens (resulting in --pf-t-v1--tokenname), adds the -v1 into the tokens, and adds the prefix variable to uses of the tokens.

Note: I did not use the prefix variable within the token files, because that should be added by the token dictionary. If we add it in core, then a v2 set of tokens could be marked as v1, etc.


srambach:
Let's defer this to later to make it easier to autofill tokens - but we'll definitely do something like this as clean up after.
One thing to consider is using `--pft-v1--` rather than `--pf-t-v1--`

PR #5987, opened 2023-10-11
feat(Accordion): added left-aligned toggle variant

### LINKED ISSUE #5426: Accordion - variant with toggle on the left
An extension to the [accordion component](https://www.patternfly.org/v4/components/accordion).

**Describe the feature**
Add the possibility to display the accordion toggle on the left side of the header, as in the [expandable section component](https://www.patternfly.org/v4/components/expandable-section), rather than on the right as is the default now.

My quick mockup:
Screenshot 2023-01-30 at 18-19-26 🚧 New test results CI Dashboard

thatblindgeye:
Closes #5426 

For context, I went with a medium spacer since it looked closer to the spacing between the toggle icon and toggle text when the text is truncated:

image

Opted to place the modifier on the top level `pf-v5-c-accordion` element since having that modifier on the individual `pf-v5-c-accordion__toggle` elements, there's the possibility to mix and match left and right aligned accordion toggles within the same accordion which isn't ideal. Having it on the top level element would sort of prevent that as the styling wouldn't be correct for the default accordion toggle positioning (the toggle icon would no longer be positioned at the end of the toggle)

### DIFF
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -19,6 +19,8 @@
   --#{$accordion}__toggle--after--Width: var(--#{$pf-global}--BorderWidth--lg);
   --#{$accordion}__toggle--m-expanded--after--BackgroundColor: var(--#{$pf-global}--primary-color--100);
   --#{$accordion}__toggle--BackgroundColor: transparent;
+  --#{$accordion}__toggle--JustifyContent: space-between;
+  --#{$accordion}__toggle--ColumnGap: 0;
 
   // accordion toggle text
   --#{$accordion}__toggle-text--MaxWidth: calc(100% - var(--#{$pf-global}--spacer--lg));
@@ -84,6 +86,11 @@
   @include pf-v5-t-light;
 
   background-color: var(--#{$accordion}--BackgroundColor);
+
+  &.pf-m-start {
+    --#{$accordion}__toggle--JustifyContent: start;
+    --#{$accordion}__toggle--ColumnGap: var(--#{$pf-global}--spacer--md);
+  }
 
   &.pf-m-display-lg {
     --#{$accordion}__toggle--PaddingTop: var(--#{$accordion}--m-display-lg__toggle--PaddingTop);
@@ -161,8 +168,9 @@
 .#{$accordion}__toggle {
   position: relative;
   display: flex;
+  column-gap: var(--#{$accordion}__toggle--ColumnGap);
   align-items: center;
-  justify-content: space-between;
+  justify-content: var(--#{$accordion}__toggle--JustifyContent);
   width: 100%;
   padding-block-start: var(--#{$accordion}__toggle--PaddingTop);
   padding-block-end: var(--#{$accordion}__toggle--PaddingBottom);
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -22,6 +22,10 @@
   --#{$accordion}__toggle--JustifyContent: space-between;
   --#{$accordion}__toggle--ColumnGap: 0;
 
+  // Accordion toggle toggle-start modifier
+  --#{$accordion}--m-toggle-start__toggle--JustifyContent: start;
+  --#{$accordion}--m-toggle-start__toggle--ColumnGap: var(--#{$pf-global}--spacer--md);
+
   // accordion toggle text
   --#{$accordion}__toggle-text--MaxWidth: calc(100% - var(--#{$pf-global}--spacer--lg));
   --#{$accordion}__toggle--hover__toggle-text--Color: var(--#{$pf-global}--link--Color);
@@ -87,9 +91,9 @@
 
   background-color: var(--#{$accordion}--BackgroundColor);
 
-  &.pf-m-start {
-    --#{$accordion}__toggle--JustifyContent: start;
-    --#{$accordion}__toggle--ColumnGap: var(--#{$pf-global}--spacer--md);
+  &.pf-m-toggle-start {
+    --#{$accordion}__toggle--JustifyContent: var(--#{$accordion}--m-toggle-start__toggle--JustifyContent);
+    --#{$accordion}__toggle--ColumnGap: var(--#{$accordion}--m-toggle-start__toggle--ColumnGap);
   }
 
   &.pf-m-display-lg {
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Accordion/accordion.scss
```diff
@@ -161,8 +168,9 @@
 .#{$accordion}__toggle {
   position: relative;
   display: flex;
+  column-gap: var(--#{$accordion}__toggle--ColumnGap);
   align-items: center;
-  justify-content: space-between;
+  justify-content: var(--#{$accordion}__toggle--JustifyContent);
```
mcoker, line 173: 👌

src/patternfly/components/Accordion/examples/Accordion.md
```diff
@@ -294,6 +295,62 @@ cssPrefix: pf-v5-c-accordion
 {{/accordion}}
 ```
 
+### Toggle icon at start
+
+```hbs
+{{#> accordion accordion--IsStartAligned="true"}}
```
mcoker, line 301: It's a new thing, but we can now use `true` and `false` as booleans instead of the string (which returns true for the hbs param regardless the value (eg `param="[true/false/chocolate chip cookie]"` all return true) 🥳

```suggestion
{{#> accordion accordion--IsStartAligned=true}}
```

src/patternfly/components/Accordion/examples/Accordion.md
```diff
@@ -294,6 +295,62 @@ cssPrefix: pf-v5-c-accordion
...
+    {{#> accordion-toggle-icon}}{{/accordion-toggle-icon}}
+    {{#> accordion-toggle-text}}Item three{{/accordion-toggle-text}}
+  {{/accordion-toggle}}
+  {{#> accordion-expandable-content}}
+    {{#> accordion-expandable-content-body}}
+      This text is hidden
+    {{/accordion-expandable-content-body}}
+  {{/accordion-expandable-content}}
+
+  {{#> accordion-toggle accordion-toggle--IsExpanded="true"}}
```
mcoker, line 332: Just in case you want to try and update any of these other pre-existing params to use `true`, one sneaky little thing to look for is to make sure we aren't relying on the string value to do something somewhere - eg, to do something like below. There's no harm in just leaving them the way they are now, either.

```
# Some .hbs file
<button
  aria-expanded="{{accordion-toggle--IsExpanded}}"
  ...
```

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -85,6 +87,11 @@
 
   background-color: var(--#{$accordion}--BackgroundColor);
 
+  &.pf-m-start {
```
mcoker, line 90: I might name this something a little more scoped, like `.pf-m-toggle-start`?

src/patternfly/components/Accordion/accordion.scss
```diff
@@ -85,6 +87,11 @@
...
+    --#{$accordion}__toggle--JustifyContent: start;
+    --#{$accordion}__toggle--ColumnGap: var(--#{$pf-global}--spacer--md);
```
mcoker, line 92: I'd create new local component vars for these values and use those vars here. That will allow someone to change the column-gap, for example, by doing this

```
.pf-v5-c-accordion {
  --#{$accordion}--m-toggle-start__toggle--ColumnGap: 42px; // or whatever their custom column gap is
}
```

As opposed to what's in this PR, they'd need to do

```
.pf-v5-c-accordion.pf-m-start {
  --#{$accordion}--m-toggle-start__toggle--ColumnGap: 42px; // or whatever their custom column gap is
}
```

This keeps all the themable vars scoped to the base component class, so they can make all of their var overrides in `.pf-v5-c-accordion` and not having to mess with additional selectors.

```suggestion
    --#{$accordion}__toggle--JustifyContent: var(--#{$accordion}--m-toggle-start__toggle--JustifyContent);
    --#{$accordion}__toggle--ColumnGap: var(--#{$accordion}--m-toggle-start__toggle--ColumnGap);
```

mcoker, line 92: Good call on this column gap. Looks like right now the thing that creates space between the text and icon is a `max-width` on the text. I'm not sure why that's there, and a margin or something wasn't used instead? (Here's the original PR for this change, I didn't see anything that stood out as to why -https://github.com/patternfly/patternfly/pull/1357)

Not sure if it makes any difference, but now that you have the column gap, you can probably set `--pf-v5-c-accordion__toggle-text--MaxWidth: 0;` in the the start aligned selector. I don't think you need a local component var for that since it isn't something we'd want anyone to change.

You may also leave a comment (or create an issue) to apply this column-gap to the default toggle and remove the max-width.

src/patternfly/components/Accordion/examples/Accordion.md
```diff
@@ -294,6 +295,62 @@ cssPrefix: pf-v5-c-accordion
 {{/accordion}}
 ```
 
+### Toggle icon at start
+
+```hbs
```
mcoker, line 300: Not sure if we need it, but you could add `isBeta` so we can iterate on it for a bit if we need to make any changes or anticipate it being a beta prop in react. I probably wouldn't since it's so simple, but totally your call.

mcoker (commented):
This is good to merge as is, but I left some comments, a little verbose just to knowledge share some 😅 

If we aren't going to mark it beta, I'd be interested in your thoughts on the `.pf-m-start` class name, since we wouldn't be able to change it after the fact.

src/patternfly/components/Accordion/examples/Accordion.md
thatblindgeye: 🤯 such a small thing but it feels like a game changer to me 😆

thatblindgeye:
@mcoker made updates based on above. I didn't set the maxWidth var to 0 within the start modifier selector since that causes the text element to have a width of 0; could update it to 100%, but since it doesn't really have much effect/it'd be replaced I didn't include it here. If we want to for now I can push another update, though.

src/patternfly/components/Accordion/accordion.scss
thatblindgeye: https://github.com/patternfly/patternfly/issues/6013

mcoker (approved):
@thatblindgeye doh! re: the max-width, I meant to say whatever the default is - `none` or reconfigure it a bit to use `revert`. But the way it is now is fine, too, since it doesn't appear to do anything.

🎃

srambach (approved):

PR #5991, opened 2023-10-12
feat(notificationBadge): refactor with tokens

srambach:
Simplifies some things that were marked for breaking change releases as well as adding tokens.
Convenience link: https://patternfly-pr-5991.surge.sh/components/notification-badge

Fixes #5990

PR #5992, opened 2023-10-12
fix(tokens): add dark theme blend mode token

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5993

### DIFF
--- a/src/patternfly/base/tokens/_tokens-font.scss
+++ b/src/patternfly/base/tokens/_tokens-font.scss
@@ -4,7 +4,7 @@
 // The ability to make font variables is reportedly coming by the end of the year
 
 :root {
-// Base tokens for fonts
+  // Base tokens for fonts
   --pf-t--global--font--family--100: redhattext;
   --pf-t--global--font--family--200: redhatdisplay;
   --pf-t--global--font--family--300: redhatmono;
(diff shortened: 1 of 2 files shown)


thatblindgeye:
@mcoker it looks like it's fixed with dark theme enabled, but the issue now occurs when dark theme is **disabled**, but there's a dark background behind the badge like the examples now show. Hard to see but there's a single attention badge that I've hovered over above the 3 badges on the bottom:

image

(also the badges being pushed up on top of one another)

Since we have full page demos that place a notification badge inside of a masthead with a dark background regardless of theme, it'd probably cause similar issues there; even if we update the background color for a masthead, do we want to assume a notification badge in light theme will always be placed on top of a lighter background, and the opposite for dark theme?

srambach (approved):
These changes look ok to me - I think the overlapping etc are because notification badge is in a weird state without the other PR.

srambach:
> Since we have full page demos that place a notification badge inside of a masthead with a dark background regardless of theme, it'd probably cause similar issues there; even if we update the background color for a masthead, do we want to assume a notification badge in light theme will always be placed on top of a lighter background, and the opposite for dark theme?
@thatblindgeye The masthead isn't going to be dark when in light (default) theme any more.

thatblindgeye (approved):
May be worth still investigating the blend mode when a badge is on a dark background in light them/vice versa in the future, but not a blocker for this PR

mcoker:
@thatblindgeye the overlapping is coming from the plain button not having any padding, so the notification badge's red/blue/transparent surface popping out of the button and overlapping adjacent buttons. It's pre-existing, you can see it here - https://pf6.patternfly.org/components/notification-badge/

I fixed the padding, and when paired with @srambach's PR https://github.com/patternfly/patternfly/pull/5991/ (after a little tinkering), it fixes things.

The light theme hover problem now was coming from this global token definition in the button, but I pushed an update to remove it in this PR now that this PR declares that token globally in the base PF CSS - https://github.com/patternfly/patternfly/blob/f12a2fb9c83ebbdb60c498b10ea54393b2993da3/src/patternfly/components/Button/button.scss#L150-L151

PR #6021, opened 2023-10-30
feat(page): apply tokens

### LINKED ISSUE #6003: Apply tokens to page

srambach:
Fixes #6003 

A couple of TODOs left waiting for tokens to be added next time.
Should we drop the overflow scroll demo? Is that modifier still useful?
The "sticky horizontal subnav" example will also need to be tweaked depending on the horizontal nav resolution.


mcoker (approved):
🦃👍

Issue #6024, opened 2023-10-31
Bug - v6 clickable table issues


mcoker:
There is an overlay for the clickable table row that is messing up the styles, and preventing stuff underneath it from being clickable in Safari. It's also failing the CI cypress tests in react - https://github.com/patternfly/patternfly-react/pull/9774

mcoker:
closed via #6025

PR #6025, opened 2023-10-31
fix(table): fix clickable overlay

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6024

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -252,7 +252,6 @@
   --#{$table}__tr--m-clickable--BoxShadow--top: var(--#{$table}__tr--BoxShadow--top--base);
   --#{$table}__tr--m-clickable--BackgroundColor: transparent;
   --#{$table}__tr--m-clickable--BoxShadow: none;
-  --#{$table}__tr--m-clickable--BorderRadius: var(--pf-t--global--border--radius--small);
   --#{$table}__tr--m-clickable--OutlineOffset: calc(-1 * var(--#{$pf-global}--spacer--xs));
 
   // TODO Shadow tokens START HERE AGAIN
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):
Looks good, assuming we want to get rid of that border radius completely on hover. Not a blocker since we can wait to see how this update works when pulled into the draft React pr currently up, but I am wondering if other instances of setting `position: relative'` on an element and then `position: absolute` on the ::before will have similar issues. 

For example, when I click one of the buttons in the [compound expansion HTML example](https://www.patternfly.org/components/table/html#compound-expansion-example) on org (with the text "234") and move the cursor away, the focus styling remains. Doing the asme in this PR's surge, however, doesn't retain the focus styling, and the compound-expansion-toggle element is using a similar position + `::before` styling.

tlabaj (approved):
Looks good in Safari now.

mcoker:
>  I am wondering if other instances of setting position: relative' on an element and then position: absolute on the ::before will have similar issues.

Nice callout! For sure, it's something you have to take into consideration when using a pseudo element that way. Sometimes it's OK, if the whole thing is clickable, for example. Like using a pseudo element on a button, for example. There are a number of ways you can use this combo and have the pseudo element not block everything underneath it, I created this codepen with a few ways - https://codepen.io/mcoker/pen/NWorqvp

Re: the [compound expansion HTML example](https://www.patternfly.org/components/table/html#compound-expansion-example) issue, I'm not seeing it - is this it or are you referring to something else? 


https://github.com/patternfly/patternfly/assets/35148959/b676eb77-67e5-4f02-99c2-71521f6f6b63

thatblindgeye:
@mcoker Maybe I was looking at a different example/got tabs mixed up. Could have sworn I was seeing some odd behavior before, but not able to reproduce now. If I notice it again I'll ping you, though.

PR #6030, opened 2023-11-01
feat(hint): apply tokens

### LINKED ISSUE #5997: Add tokens to Hint

srambach:
Fixes #5997


lboehling:
this looks great! I'm noticing that the borders radius is using the large token instead of medium. can we update it to use the medium border radius (16px)?

mcoker (approved):
🦃👍

PR #6033, opened 2023-11-03
feat(nav): added tokens, new design spec

### LINKED ISSUE #6000: Consume tokens in: Navigation

mattnolting:
Fixes #6000


srambach:
In pages, you can see that we've got a large spacer on the inline-start of the page sidebar AND on the navigation. I think the padding should be removed from the page sidebar, WDYT?

mattnolting:
> In pages, you can see that we've got a large spacer on the inline-start of the page sidebar AND on the navigation. I think the padding should be removed from the page sidebar, WDYT?

@srambach Yep, I agree 👍

src/patternfly/components/Nav/nav-item.hbs
```diff
@@ -1,14 +1,23 @@
...
+  {{/unless}}"
+
   {{#if nav-item--attribute}}
     {{{nav-item--attribute}}}
   {{/if}}>
-  {{> @partial-block}}
+  {{#if nav-link--text}}
+    {{> nav-link}}
+  {{/if}}
+  {{#if @partial-block}}
```
srambach, line 20: Just curious about the reason to put the partial-block inside a conditional

andrew-ronaldson:
Agree with the comments above on the spacer

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,276 +1,113 @@
...
+  --#{$nav}--m-horizontal--PaddingBlockEnd: var(--pf-t--global--spacer--sm);;
+  --#{$nav}--m-horizontal__list--RowGap: 0;
+  --#{$nav}--m-horizontal__list--ColumnGap: var(--pf-t--global--spacer--xs);
+  --#{$nav}--m-horizontal__link--AlignItems: center;
+  --#{$nav}--m-horizontal__link--WhiteSpace: nowrap;
+  --#{$nav}--m-horizontal__link--row-gap--offset: 0;
+  --#{$nav}--m-horizontal__link--column-gap--offset: calc(var(--#{$nav}__list--ColumnGap) / 2 * -1); // inline start/end offset by half row gap to consume space between items to prevent flicker
+
+  // Horinzontal subnav
+  --#{$nav}--m-horizontal--m-subnav--BorderBlockStart--width: var(--pf-t--global--border--color--200);
```
srambach, line 67: Here you have the width and color mixed up. Also, I think the color should be `--pf-t--global--border--color--default`.

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,276 +1,113 @@
...
-  --#{$nav}__subnav__subnav__link--PaddingLeft: var(--#{$pf-global}--spacer--md);
-  --#{$nav}__subnav__subnav__link--FontSize: var(--#{$pf-global}--FontSize--xs);
-  --#{$nav}__item--m-expanded__subnav--MaxHeight: 100%;
-  --#{$nav}__subnav--c-divider--PaddingRight: var(--#{$pf-global}--spacer--lg);
-  --#{$nav}__subnav--c-divider--PaddingLeft: var(--#{$pf-global}--spacer--lg);
+  --#{$nav}__link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--alt--hover);
+  --#{$nav}__link--focus--BackgroundColor: var(--pf-t--global--background--color--action--plain--alt--focus);
+  --#{$nav}__link--active--BackgroundColor: var(--pf-t--global--background--color--action--plain--alt--selected);
+  --#{$nav}__link--m-current--BackgroundColor: var(--pf-t--global--background--color--action--plain--alt--selected);
+  --#{$nav}__link--Transition: var(--#{$pf-global}--Transition);
```
srambach, line 44: Can you leave a `// TODO` here about using a transition token once available?

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,276 +1,113 @@
...
+  --#{$nav}__toggle--PaddingLeft: var(--pf-t--global--spacer--sm);
+  --#{$nav}__toggle--FontSize: var(--pf-t--global--icon--size--md);
+  --#{$nav}__toggle--Color: var(--pf-t--global--icon--color--subtle);
+  --#{$nav}__toggle--hover--Color: var(--pf-t--global--text--color--regular);
+  --#{$nav}__toggle--focus--Color: var(--pf-t--global--text--color--regular);
+  --#{$nav}__toggle--active--Color: var(--pf-t--global--text--color--regular);
+  --#{$nav}__toggle--TranslateY: calc((var(--#{$nav}__link--LineHeight) * var(--#{$nav}__link--FontSize) / 2 ) - 50%); // find height of single label, divide by two, offset by 50% of own height
 
   // Toggle icon
   --#{$nav}__toggle-icon--Transition: var(--#{$pf-global}--TransitionDuration);
```
srambach, line 106: Can you leave a `// TODO` here about using a transition token once available?

srambach (commented):
Couple of minor things, but looks really good! Maybe you can walk us through the handlebars helpers sometime?

lboehling:
this is looking great 🎉, thank you @mattnolting!! a few of comments --

- I agree with the spacing comments above. can you remove the left and right padding from main nav sections, so it just has the top/bottom padding? like this 
Screenshot 2023-11-15 at 10 22 11 AM

- Can you remove the right padding from nested the nav section so it extends full width on the right like this?
Screenshot 2023-11-15 at 10 23 02 AM

- Can we bump up group header text to body-md (14px)? In the designs I also had it set to small (12px) but i think that it reads better larger now that i see it in build! I'll need to update the design for this too. 

- The horizontal sub nav is showing within the page content area. Should it show below the masthead or should I update design to make it look better when placed in the page section?
Screenshot 2023-11-15 at 10 28 42 AM

src/patternfly/components/Nav/nav-item.hbs
mattnolting: Great question, the conditional allows for additional content in the `nav__item`. Without the conditional, partial content is required, which then requires a closing tag 

`{{#> nav-item nav-link--text='my-content'}}{{/nav-item}}` 

vs 

`{{> nav-item nav-link--text='my-content'}}`)

mattnolting:
@lboehling 

> I agree with the spacing comments above. can you remove the left and right padding from main nav sections, so it just has the top/bottom padding? like this

Done. Thanks @srambach 👍 

> Can you remove the right padding from nested the nav section so it extends full width on the right like this?

Done. 

Screenshot 2023-11-16 at 11 38 03 AM

> Can we bump up group header text to body-md (14px)? In the designs I also had it set to small (12px) but i think that it reads better larger now that i see it in build! I'll need to update the design for this too.

Done.

Screenshot 2023-11-16 at 11 38 46 AM

> The horizontal sub nav is showing within the page content area. Should it show below the masthead or should I update design to make it look better when placed in the page section?

Screenshot 2023-11-16 at 11 26 45 AM

@lboehling I think it would be great to see both examples. The example you're referring to is controlled by `page` rather than `nav`, so we'd need to consider how main/subnav is presented within page context. I'd also like to see `nav/subnav` is the masthead in both mobile and desktop formats. 


Screenshot 2023-11-16 at 11 47 25 AM

Screenshot 2023-11-16 at 11 43 33 AM

Should `scroll-buttons` have interaction states (hover/focus/active) other than text color? 

Should `scroll-buttons` have a `box-shadow` when the list exceeds the available space? 

Something like this: 

Screenshot 2023-11-16 at 11 55 43 AM

andrew-ronaldson (commented):
Screenshot 2023-12-16 at 8 41 00 AM
Not sure if this was addressed above but the subnav content has a padding-inline-start of 24px but the designs are 16px.

andrew-ronaldson (approved):
Looks great, thanks!

src/patternfly/components/Nav/examples/Navigation.md
```diff
@@ -8,6 +8,91 @@ import './Navigation.css'
...
+      {{> nav-item nav-link nav-link--text='User management'}}
+      {{> nav-item nav-link nav-link--text='Administration'}}
+      {{> nav-item nav-link nav-link--text='Import YAML'}}
+      {{> nav-item nav-link nav-link--text='Help & support'}}
+    {{/nav-list}}
+  {{/nav-section}}
+{{/nav}}
+```
+
+## New Expandable
```
srambach, line 37:
```suggestion
### New Expandable
```
Hierarchy in the headings here - but also, what should these really be called? (probably not "New" right?)

srambach (commented):
The carets seem to be broken in the preview?

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,1210 +1,415 @@
...
-  // Divider
-  --#{$nav}--c-divider--MarginTop: var(--#{$pf-global}--spacer--sm);
-  --#{$nav}--c-divider--MarginBottom: var(--#{$pf-global}--spacer--sm);
-  --#{$nav}--c-divider--PaddingRight: 0;
-  --#{$nav}--c-divider--PaddingLeft: 0;
-  --#{$nav}--c-divider--BackgroundColor: var(--#{$pf-global}--BackgroundColor--dark-200);
-
-  // Scroll snap
+:root,
+:where(.#{$nav}) {
```
thatblindgeye, line 2: Other components have been updated to root + just the component selector here:

```suggestion
:root,
.#{$nav} {
```

Do we want to shift to utilizing `:where` here?

src/patternfly/components/Nav/nav.scss
```diff
@@ -1,1210 +1,415 @@
...
-  --#{$nav}__link--hover--Color: var(--pf-t--global--text--color--subtle);
-  --#{$nav}__link--focus--Color: var(--pf-t--global--text--color--subtle);
-  --#{$nav}__link--active--Color: var(--pf-t--global--text--color--subtle);
-  --#{$nav}__link--m-current--Color: var(--pf-t--global--text--color--regular);
-  --#{$nav}__link--BackgroundColor: transparent;
-  --#{$nav}__link--hover--BackgroundColor: var(--pf-t--global--background--color--primary--default);
-  --#{$nav}__link--focus--BackgroundColor: var(--pf-t--global--background--color--primary--default);
-  --#{$nav}__link--active--BackgroundColor: var(--pf-t--global--background--color--primary--default);
-  --#{$nav}__link--m-current--BackgroundColor: var(--pf-t--global--background--color--primary--default);
-  --#{$nav}__link--OutlineOffset: calc(var(--#{$pf-global}--spacer--xs) * -1);
```
thatblindgeye, line 67: This would be worth adding back in, as right now the focus outline gets cut off in horizontal examples:

image

src/patternfly/components/Nav/examples/Navigation.md
mattnolting: Ah, yes. Those examples are for demo purposes. They'll be removed before merge.

src/patternfly/components/Nav/nav.scss
mattnolting: The core team talked about this earlier this week. Using `:where()` here resolves some variables conflicts while allowing `vars()` to be updated/themed without wrapping in the `.pf-v5-c-component` selector.

thatblindgeye (approved):

src/patternfly/components/Button/button.scss
```diff
@@ -271,7 +274,15 @@
...
-  border-radius: var(--#{$button}--BorderRadius);
+  border-start-start-radius: var(--#{$button}--start-start--Radius--BorderRadius);
+  border-start-end-radius: var(--#{$button}--start-end--Radius--BorderRadius);
+  border-end-start-radius: var(--#{$button}--end-start--Radius--BorderRadius);
+  border-end-end-radius: var(--#{$button}--end-end--Radius--BorderRadius);
+
+  // apply line-height to all children, including fa fonts
+  > * {
+    line-height: var(--#{$button}--LineHeight);
+  }
```
mcoker, line 285: Curious what this is needed for? Looks like the horizontal nav arrow icons align OK without it.

src/patternfly/components/Button/button.scss
```diff
@@ -611,6 +622,12 @@
       }
     }
   }
+
+  &.pf-m-RTL-icon-static {
+    .#{$button}__icon {
+      @include pf-v5-mirror-inline-on-rtl
+    }
+  }
```
mcoker, line 630: This works but so far we tried not to have RTL modifiers for everything that accepts an icon, and instead use either the mixins in core or react has been using `<Icon shouldMirrorRTL />`. On the react side I'd expect this to be setup as

```
<Button variant="plain">
  <Icon shouldMirrorRTL><AngleRightIcon /></Icon>
</Button>
```

I'd probably do the same on the core side. You should always be able to pass the icon component to a button as the `icon` prop or as a child of a plain button. That also prevents any confusion around the global class `.pf-m-mirror-inline-rtl` and this local `.pf-m-RTL-icon-static` class, which are named differently, but do the same thing.

mcoker (commented):
LGTM, the only thing I'd like to look at before merging is the button enhancement to have its own class to mirror/flip its icon, since that's adding a feature to the react button. IIRC, last we spoke about it we wanted to try and use `<Icon shouldMirrorRTL>` anywhere there isn't a core class we can target to apply a mixin and you can pass the icon to something.

mcoker (changes requested):
Oh forgot to leave this comment - the new button vars you've introduced are not initially defined in the button component. Not sure if that's on purpose or not, but at a minimum the border-radius vars need to be added back since the button has lost its rounded corners https://patternfly-pr-6033.surge.sh/components/button

mcoker (approved):

Issue #5716, opened 2023-07-05
Consume tokens: Chip, Label, and Badge


srambach:
[Figma design for labels](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A31&mode=design&t=doZDRUaxE4vsn5dD-1)

[Figma design for badge](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A5&mode=design&t=doZDRUaxE4vsn5dD-1)

Note that **chips will be deprecated** and labels should be used instead. https://github.com/patternfly/patternfly/issues/5343
```[tasklist]
### Tasks
- [x] Tokenize chip
- [x] Tokenize badge
- [ ] https://github.com/patternfly/patternfly/issues/6181
```

Chip will not be pentified - but will removed in favor of Label. The React Chip component will deliver a Label under the hood.

PR #6037, opened 2023-11-07
fix(badge): adds tokens, dropdown toggle icon

mcoker:
towards #5716
fixes https://github.com/patternfly/patternfly/issues/5957


srambach (approved):
This looks good - the badge right now is not tall enough, but it's because the line-height values are currently switched between body and heading text. This PR is good as it stands and the line-height can be fixed elsewhere. (or here if you want)

mfrances17 (commented):
LGTM

PR #6042, opened 2023-11-08
feat(modal): add tokens to modal and about modal

### LINKED ISSUE #5707: Consume tokens in: Modal and About Modal
[Figma design for Modal](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A34&mode=design&t=doZDRUaxE4vsn5dD-1)

[Figma design for About modal](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A18&mode=design&t=doZDRUaxE4vsn5dD-1)

srambach:
Fixes #5707 

NOTE: The about modal image currently does not switch over in dark mode. We'll need a solution for this. Right now I've commented out a couple of lines that were being used for sizing the background image in favor of using "cover" but if we use a different background image we may want these back again. @lboehling what ideas do you have for the image, and what is the expectation as the modal is resized?

ALSO, this corrects a previous mistake in the line-height tokens, where the semantic tokens for regular text and heading line-height were switched.


lboehling:
this is looking awesome 🎉 

here's an svg for the bkg image 
PentaBkgImg
If we use the gray-40 fill color for both shapes and put them both at 15% opacity, it looks good on both our primary and secondary background colors in light and dark theme i think?

The image is resizing as i would expect in the PR!

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -1,14 +1,12 @@
 // @debug $about-modal-box; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
-.#{$about-modal-box} {
+:root {
   // Component variables
   --#{$about-modal-box}--BackgroundImage: none;
-  --#{$about-modal-box}--BackgroundColor: var(--#{$pf-global}--palette--black-1000); // Modal uses a non-standard background color
+  --#{$about-modal-box}--BackgroundColor: var(--pf-t--global--background--color--floating--default);
```
mcoker, line 6: We could also remove this if the intent from design is that this comes from the modal. Similar to the rounded corners and box-shadow. The design would even be a lot more clear if the about modal were created inside of the modal component, so it's clear what parts of the design are from what components.

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -1,14 +1,12 @@
...
   --#{$about-modal-box}--BackgroundPosition: bottom right;
-  --#{$about-modal-box}--BackgroundSize--min-width: 200px;
-  --#{$about-modal-box}--BackgroundSize--width: 60%;
-  --#{$about-modal-box}--BackgroundSize--max-width: 600px;
-  --#{$about-modal-box}--BackgroundSize: clamp(var(--#{$about-modal-box}--BackgroundSize--min-width), var(--#{$about-modal-box}--BackgroundSize--width), var(--#{$about-modal-box}--BackgroundSize--max-width));
+  --#{$about-modal-box}--BackgroundSize--width: 100%;
+  --#{$about-modal-box}--BackgroundSize: contain;
```
mcoker, line 9: We should nail down what the intent is with the background image - @lboehling @andrew-ronaldson do we have a design guideline for what kind of image to use here? And is there any reason  If you put a wide image in now, with `contain`, it will fill the whole box - https://codepen.io/mcoker/pen/PoVpVqQ

Doesn't need to hold up this PR, just something to consider. Also the about modal and background image component CSS is basically the same. The login page uses the background image component - seems like the about modal could, too, and we could just remove the background image specific styles from the about modal?

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -54,58 +52,53 @@
...
 
   // Close Button
-  --#{$about-modal-box}__close--c-button--Color: var(--#{$pf-global}--Color--100);
-  --#{$about-modal-box}__close--c-button--FontSize: var(--#{$pf-global}--FontSize--xl);
-  --#{$about-modal-box}__close--c-button--BorderRadius: var(--#{$pf-global}--BorderRadius--lg);
-  --#{$about-modal-box}__close--c-button--Width: calc(var(--#{$about-modal-box}__close--c-button--FontSize) * 2); // Height and width are 2x font-size
-  --#{$about-modal-box}__close--c-button--Height: calc(var(--#{$about-modal-box}__close--c-button--FontSize) * 2);
-  --#{$about-modal-box}__close--c-button--BackgroundColor: var(--#{$pf-global}--palette--black-1000); // Close button uses a non-standard background color
-  --#{$about-modal-box}__close--c-button--hover--BackgroundColor: #{rgba($pf-v5-color-black-1000, .4)};
+  --#{$about-modal-box}__close--c-button--Color: var(--pf-t--global--icon--color--regular);
```
mcoker, line 55: You shouldn't need this - it will come from the button.

src/patternfly/components/AboutModalBox/about-modal-box.scss
```diff
@@ -219,23 +213,8 @@
...
-    height: var(--#{$about-modal-box}__close--c-button--Height);
     padding: 0;
     font-size: var(--#{$about-modal-box}__close--c-button--FontSize);
     color: var(--#{$about-modal-box}__close--c-button--Color);
-    background-color: var(--#{$about-modal-box}__close--c-button--BackgroundColor);
-    border-radius: var(--#{$about-modal-box}__close--c-button--BorderRadius);
-
-    &:hover {
-      --#{$about-modal-box}__close--c-button--BackgroundColor: var(--#{$about-modal-box}__close--c-button--hover--BackgroundColor);
-    }
```
mcoker, line 232: I think you can get rid of all of this custom CSS aside from the font-size. Everything else should just come from the plain button. All of this is because it is a circle shape in PF4/5.

src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -1,81 +1,88 @@
...
 
+.#{$modal-box} {
   position: relative;
   z-index: var(--#{$modal-box}--ZIndex);
   display: flex;
   flex-direction: column;
   width: var(--#{$modal-box}--Width);
   max-width: var(--#{$modal-box}--MaxWidth);
   max-height: var(--#{$modal-box}--MaxHeight);
+  overflow: hidden;
```
mcoker, line 83: If you have a tall thing in the modal and no modal body (the only element that allows you to scroll overflow currently), it will just run off the screen. I would probably set this to `auto` instead. 

This may cause a double scrollbar as a tall modal with a tall body element is in a super short viewport, but we can revisit that. I don't know that it would be any worse than it is now. We could also remove the overflow scroll from the modal body and make the footer sticky to the bottom.

mcoker (approved):
LGTM, left some comments. The only thing I might go ahead and update is setting the modal to `overflow: auto`, but that can be a follow up. Lemme know if you'd rather merge and follow up or update this PR.

src/patternfly/components/AboutModalBox/about-modal-box.scss
srambach: It wasn't clear to me whether this background color would simply be the modal background color or if them being the same was a coicidence. @lboehling ?

src/patternfly/components/AboutModalBox/about-modal-box.scss
srambach: I took out the sizing because it's fairly specific to the image anyway and this one needs to span the full height or it will look odd. And there will need to be guidelines on making the image since it needs to handle the switch to dark theme.
I do think we should probably just use the background component here?

srambach:
@lboehling says the description is not actually part of an about modal, so @mcoker please let me know if there are any additional things - I fixed the close button and spacing.

mcoker (approved):
🤖👍

PR #6031, opened 2023-11-02
feat(text, title, list): apply tokens

srambach:
Fixes #5730 

PR right before vacation so... 🤪 

Also adds modifiers for heading styles to the title component.

PR #6051, opened 2023-11-10
feat(global): remove invalid selector crashing inspector

srambach:
`:where()` with a pseudo element is not valid, and this was causing the inspector in Chrome 119 and Edge to crash. This changes the selector to remove `:where`. The only side effect is increasing the specificity of that particular rule.

### DIFF
--- a/src/patternfly/base/_globals.scss
+++ b/src/patternfly/base/_globals.scss
@@ -33,10 +33,8 @@
 // Normalize
 @if $pf-v5-global--enable-normalize {
   *,
-  :where(
-    *::before,
-    *::after
-  ) {
+  *::before,
+  *::after {
     box-sizing: border-box;
   }
 


evwilkin (approved):
LGTM to fix the bug - though would still defer to @mattnolting or @mcoker on any consequences of the specificity increase you called out 👍

mcoker:
LGTM but this change is also in https://github.com/patternfly/patternfly/pull/6031. That one is ready to merge, too. Do you prefer to keep the change there or in this PR?

mcoker (approved):

srambach:
@mcoker Either way - just wanted to get this fix into v6 so it's not slowing people down.

mcoker:
Addresses in https://github.com/patternfly/patternfly/pull/6031

PR #6056, opened 2023-11-10
feat(EmptyState): apply tokens

### LINKED ISSUE #5733: Consume tokens: Empty state
[Figma design for empty state](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A23&mode=design&t=doZDRUaxE4vsn5dD-1)

srambach:
Fixes #5733 

This also adds status modifiers for the empty state per conversation with @andrew-ronaldson. These set the icon's color to the appropriate status color.

### DIFF
--- a/src/patternfly/components/EmptyState/empty-state.scss
+++ b/src/patternfly/components/EmptyState/empty-state.scss
@@ -1,14 +1,14 @@
 // @debug $empty-state; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
-.#{$empty-state} {
-  --#{$empty-state}--PaddingTop: var(--#{$pf-global}--spacer--xl);
-  --#{$empty-state}--PaddingRight: var(--#{$pf-global}--spacer--xl);
-  --#{$empty-state}--PaddingBottom: var(--#{$pf-global}--spacer--xl);
-  --#{$empty-state}--PaddingLeft: var(--#{$pf-global}--spacer--xl);
-  --#{$empty-state}--m-xs--PaddingTop: var(--#{$pf-global}--spacer--md);
-  --#{$empty-state}--m-xs--PaddingRight: var(--#{$pf-global}--spacer--md);
-  --#{$empty-state}--m-xs--PaddingBottom: var(--#{$pf-global}--spacer--md);
-  --#{$empty-state}--m-xs--PaddingLeft: var(--#{$pf-global}--spacer--md);
+:root {
+  --#{$empty-state}--PaddingTop: var(--pf-t--global--spacer--xl);
+  --#{$empty-state}--PaddingRight: var(--pf-t--global--spacer--xl);
+  --#{$empty-state}--PaddingBottom: var(--pf-t--global--spacer--xl);
+  --#{$empty-state}--PaddingLeft: var(--pf-t--global--spacer--xl);
+  --#{$empty-state}--m-xs--PaddingTop: var(--pf-t--global--spacer--md);
+  --#{$empty-state}--m-xs--PaddingRight: var(--pf-t--global--spacer--md);
+  --#{$empty-state}--m-xs--PaddingBottom: var(--pf-t--global--spacer--md);
+  --#{$empty-state}--m-xs--PaddingLeft: var(--pf-t--global--spacer--md);
 
   // Content
   --#{$empty-state}__content--MaxWidth: none;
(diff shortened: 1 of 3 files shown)


src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -17,40 +17,49 @@
...
-  --#{$empty-state}--m-xs__body--FontSize: var(--#{$pf-global}--FontSize--sm);
-  --#{$empty-state}--m-xs__body--MarginTop: var(--#{$pf-global}--spacer--md);
-  --#{$empty-state}--m-xl__body--FontSize: var(--#{$pf-global}--FontSize--xl);
-  --#{$empty-state}--m-xl__body--MarginTop: var(--#{$pf-global}--spacer--lg);
+  --#{$empty-state}__body--MarginTop: var(--pf-t--global--spacer--md);
+  --#{$empty-state}__body--Color: var(--pf-t--global--text--color--subtle);
+  --#{$empty-state}--body--FontSize: var(--pf-t--global--font--size--body--lg);
+  --#{$empty-state}--m-xs__body--FontSize: var(--pf-t--global--font--size--body);
+  --#{$empty-state}--m-xs__body--MarginTop: var(--pf-t--global--spacer--md);
+  --#{$empty-state}--m-xl__body--FontSize: var(--pf-t--global--font--size--body--lg);
```
mattnolting, line 49:
```suggestion
  --#{$empty-state}--m-xl__body--FontSize: var(--pf-t--global--font--size--heading--md);
```

Design spec shows `20px` font-size here. Should it be `--pf-t--global--font--size--heading--md`?

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -17,40 +17,49 @@
   --#{$empty-state}--m-lg__content--MaxWidth: #{pf-size-prem(600px)};
 
   // Icon
-  --#{$empty-state}__icon--MarginBottom: var(--#{$pf-global}--spacer--lg);
-  --#{$empty-state}__icon--FontSize: var(--#{$pf-global}--icon--FontSize--xl);
-  --#{$empty-state}__icon--Color: var(--#{$pf-global}--icon--Color--light);
-  --#{$empty-state}--m-xs__icon--MarginBottom: var(--#{$pf-global}--spacer--md);
-  --#{$empty-state}--m-xl__icon--MarginBottom: var(--#{$pf-global}--spacer--xl);
+  --#{$empty-state}__icon--MarginBottom: var(--pf-t--global--spacer--lg);
```
mattnolting, line 20: Could we update this to `--#{$empty-state}__content--RowGap: var(--pf-t--global--spacer--md);`

Then remove the margins from individual elements?

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -93,6 +102,26 @@
...
+  &.pf-m-info {
+    --#{$empty-state}__icon--Color: var(--#{$empty-state}--m-info__icon--Color);
+  }
+
+  &.pf-m-custom {
+    --#{$empty-state}__icon--Color: var(--#{$empty-state}--m-custom__icon--Color);
+  }
 }
 
 .#{$empty-state}__content {
```
mattnolting, line 127:
```suggestion
.#{$empty-state}__content {
  row-gap: var(--#{$empty-state}__content--RowGap);
```

mattnolting (changes requested):
Looking good!

@srambach looks like `pf-m-link` padding is off as well, should we open a separate issue?

Screenshot 2023-11-21 at 9 53 40 AM

src/patternfly/components/EmptyState/empty-state.scss
srambach: @lboehling I suspect we still want a body text here? Can you confirm?

src/patternfly/components/EmptyState/empty-state.scss
srambach: @mattnolting Can you tell me more about what you expect here? __content isn't a flex container so RowGap wouldn't apply.

src/patternfly/components/EmptyState/empty-state.scss
srambach: (@lboehling said to use Body Large - a change in Figma wasn't merged yet)

src/patternfly/components/EmptyState/empty-state.scss
mattnolting: I would expect our `content` containers to have a base layout, so they're flexible enough to include additional info. Any specific sub-components would then have padding to match the design spec. 

Screenshot 2023-11-21 at 3 23 27 PM

In this example, the container provides base spacing of `spacer--md` for rows, where `pf-v5-c-empty-state__body` provides bottom padding.

mattnolting: Also, I'd make `pf-v5-c-empty-state__body` `display: grid`

srambach:
NOTE: XL variation has a 2xl icon, which is in Figma but has not yet been brought over to our token variables with adjusted icon sizing.

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -93,10 +102,31 @@
...
   max-width: var(--#{$empty-state}__content--MaxWidth);
+  max-width: var(--#{$empty-state}__content--RowGap);
```
mcoker, line 129: did you mean to add this? the width constraints for the size variations isn't working with this in.

mcoker (commented):
Just one comment, otherwise 👍👍

src/patternfly/components/EmptyState/empty-state.scss
srambach: I did not! I will fix it.

mcoker (approved):
L🎄TM!

thatblindgeye (approved):
Looks good, only nit is we could add the empty state icon to the "extra small" example, not a blocker here.

Just to confirm @srambach, the "extra large" example having the incorrect icon size is expected until updated icon sizing from Figma is brought over?

srambach:
> Looks good, only nit is we could add the empty state icon to the "extra small" example, not a blocker here.
Ah - yes I think the old standard was not to include an icon possibly. I'll add the icon and change the text to say it's optional (per @lboehling )
> 
> Just to confirm @srambach, the "extra large" example having the incorrect icon size is expected until updated icon sizing from Figma is brought over?
It should have :) but I rebased to get it and we need to work out an icon size thing... stay tuned.

srambach:
@thatblindgeye I added the icon to the XS variant, and I added a new token that's now in Figma for a 96px, 3xl icon.

thatblindgeye (approved):

PR #6057, opened 2023-11-13
fix(toolbar-action-list-overflow-menu): added tokens, new design spec

### LINKED ISSUE #5713: Consume tokens in: Toolbar and Action list
[Figma design for Toolbar](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A49&mode=design&t=doZDRUaxE4vsn5dD-1)

```[tasklist]
### Tasks
- [ ] https://github.com/patternfly/patternfly/pull/6057
```

mattnolting:
closes #5713 

https://codepen.io/bobcatTaco/pen/MWxjpVr?editors=1100

Changes affect:
- https://patternfly-pr-6057.surge.sh/components/action-list/
- https://patternfly-pr-6057.surge.sh/components/overflow-menu/
- https://patternfly-pr-6057.surge.sh/components/toolbar
- https://patternfly-pr-6057.surge.sh/components/toolbar/html-demos/

### DIFF
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,24 +1,19 @@
-// @debug $action-list; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
-
-:where(:root),
-:where(.#{$action-list}) {
+:where(:root) {
   --#{$action-list}--RowGap: var(--pf-t--global--spacer--sm);
-  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--sm);
-  --#{$action-list}--Padding: var(--pf-t--global--spacer--sm);
-  --#{$action-list}--inset: var(--#{$action-list}--Padding);
-  --#{$action-list}--LineHeight: var(--pf-t--global--font--line-height--body);
-  --#{$action-list}--FontSize: var(--pf-t--global--font--size--body--md);
-  --#{$action-list}--BackgroundColor: var(--pf-t--global--background--color--primary--default);
+  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--2xl);
+
+  // * Action list group
+  --#{$action-list}__group--ColumnGap: var(--pf-t--global--spacer--md);
 
   // * Action list icons
-  --#{$action-list}--m-icons--ColumnGap: 0;
+  --#{$action-list}--m-icons--ColumnGap: var(--pf-t--global--spacer--sm);
 }
 
 // - Action list - Action list group
 .#{$action-list},
 .#{$action-list}__group {
   display: flex;
-  align-items: center;
+  align-items: start;
 }
 
 // - Action list
@@ -26,13 +21,19 @@
   row-gap: var(--#{$action-list}--RowGap);
   column-gap: var(--#{$action-list}--ColumnGap);
 
+  // - Action list icons
   &.pf-m-icons {
-    column-gap: var(--#{$action-list}--m-icons--ColumnGap);
+    column-gap: var(--#{$action-list}--m-icons--ColumnGap); // update prop: val to prevent potential inheritance conflicts
   }
 }
 
 // - Action list group
 .#{$action-list}__group {
   row-gap: var(--#{$action-list}__group--RowGap, var(--#{$action-list}--RowGap));
-  column-gap: var(--#{$action-list}__group--ColumnGap, var(--#{$action-list}--ColumnGap));
+  column-gap: var(--#{$action-list}__group--ColumnGap);
+
+  // - Action list group icons
+  &.pf-m-icons {
+    column-gap: var(--#{$action-list}--m-icons--ColumnGap); // update prop: val to prevent potential inheritance conflicts
+  }
 }
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,29 +1,38 @@
 // @debug $action-list; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
-.#{$action-list} {
-  --#{$action-list}--m-icon--spacer: 0;
-  --#{$action-list}--child--spacer-base: var(--#{$pf-global}--spacer--md);
-  --#{$action-list}--group--spacer-base: var(--#{$pf-global}--spacer--2xl);
+:where(:root),
+:where(.#{$action-list}) {
+  --#{$action-list}--RowGap: var(--pf-t--global--spacer--sm);
+  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--sm);
+  --#{$action-list}--Padding: var(--pf-t--global--spacer--sm);
+  --#{$action-list}--inset: var(--#{$action-list}--Padding);
+  --#{$action-list}--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$action-list}--FontSize: var(--pf-t--global--font--size--body--md);
+  --#{$action-list}--BackgroundColor: var(--pf-t--global--background--color--primary--default);
+
+  // * Action list icons
+  --#{$action-list}--m-icons--ColumnGap: 0;
 }
 
+// - Action list - Action list group
 .#{$action-list},
 .#{$action-list}__group {
-  --#{$action-list}--child--spacer: var(--#{$action-list}--child--spacer-base);
-  --#{$action-list}--group--spacer: var(--#{$action-list}--group--spacer-base);
-
   display: flex;
   align-items: center;
+}
 
-  > * + * {
-    margin-inline-start: var(--#{$action-list}--child--spacer);
-  }
-
-  > * + .#{$action-list}__group,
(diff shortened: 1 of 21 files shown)


andrew-ronaldson:
The preview seems to be broken on this one?

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -1,209 +1,84 @@
...
+  --#{$toolbar}--PaddingBlock: var(--pf-t--global--spacer--sm);
+  --#{$toolbar}--PaddingInline: var(--pf-t--global--spacer--sm);
+  --#{$toolbar}--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$toolbar}--FontSize: var(--pf-t--global--font--size--body--default);
+
+  // * Toolbar item
+  --#{$toolbar}__item--ColumnGap: var(--pf-t--global--spacer--sm);
+
+  // * Toolbar expandable content
+  --#{$toolbar}__expandable-content--ZIndex: var(--#{$pf-global}--ZIndex--xs); // TODO: update when z-index is tokenized
```
srambach, line 24: There are tokens for zindex now (also line 30)

srambach (commented):
Need to remove the test toolbar, and there are zindex tokens now, then I can't see anything else obvious, so I think we can move ahead! 🚗

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -1,29 +1,38 @@
...
-    margin-inline-start: var(--#{$action-list}--group--spacer);
-  }
+// - Action list
+.#{$action-list} {
+  row-gap: var(--#{$action-list}--RowGap);
+  column-gap: var(--#{$action-list}--ColumnGap);
 
   &.pf-m-icons {
-    --#{$action-list}--child--spacer: var(--#{$action-list}--m-icon--spacer);
+    column-gap: var(--#{$action-list}--m-icons--ColumnGap);
```
thatblindgeye, line 30: Should this instead be:

```suggestion
    --#{$action-list}--ColumnGap: var(--#{$action-list}--m-icons--ColumnGap);
```

src/patternfly/components/OverflowMenu/overflow-menu.scss
```diff
@@ -1,89 +1,51 @@
...
 
-  // Icon button group
+  // - Overflow menu icon button group
   &.pf-m-icon-button-group {
-    --#{$overflow-menu}--spacer: var(--#{$overflow-menu}__group--m-icon-button-group--spacer);
-
-    > * {
-      --#{$overflow-menu}--spacer: var(--#{$overflow-menu}__group--m-icon-button-group--space-items);
-    }
+    column-gap: var(--#{$overflow-menu}__group--m-icon-button-group--ColumnGap, var(--#{$overflow-menu}--ColumnGap));
```
thatblindgeye, line 38: Do we need the fallback since the group--m-icon--button-group--columngap is defined?

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -1,29 +1,38 @@
...
   }
 }
+
+// - Action list group
+.#{$action-list}__group {
+  row-gap: var(--#{$action-list}__group--RowGap, var(--#{$action-list}--RowGap));
+  column-gap: var(--#{$action-list}__group--ColumnGap, var(--#{$action-list}--ColumnGap));
```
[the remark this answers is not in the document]
mcoker: Asked @lboehling about this and she said the space between groups (or an item and a group) should be 2xl

src/patternfly/components/ActionList/action-list.scss
```diff
@@ -1,29 +1,38 @@
 // @debug $action-list; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
-.#{$action-list} {
-  --#{$action-list}--m-icon--spacer: 0;
-  --#{$action-list}--child--spacer-base: var(--#{$pf-global}--spacer--md);
-  --#{$action-list}--group--spacer-base: var(--#{$pf-global}--spacer--2xl);
+:where(:root),
+:where(.#{$action-list}) {
+  --#{$action-list}--RowGap: var(--pf-t--global--spacer--sm);
+  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--sm);
```
mcoker, line 6: per the design, looks like this should be a medium spacer

```suggestion
  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--md);
```

src/patternfly/components/OverflowMenu/overflow-menu.scss
```diff
@@ -1,89 +1,51 @@
...
 }
 
-// Group
+// - Overflow menu group
 .#{$overflow-menu}__group {
-  --#{$overflow-menu}--spacer: var(--#{$overflow-menu}__group--spacer);
-
-  display: flex;
-  align-items: center;
+  column-gap: var(--#{$overflow-menu}__group--ColumnGap, var(--#{$overflow-menu}--ColumnGap));
```
thatblindgeye, line 29: Might be similar to Michael's comment(s) above? I think one worry about doing this without defining the var is that it doesn't get documented on Org, so you have to inspect the page to see that `--#{$overflow-menu}__group--ColumnGap` isn't being applied to the column gap.

Can be a follow up if we would want to change it, but lmk if there's another reason for doing this vs defining --#{$overflow-menu}__group--ColumnGap and other vars at the top

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -1,209 +1,84 @@
...
+
+@if $pf-v5--include-toolbar-breakpoints {
+  $pf-v5-c-toolbar--breakpoint-map: build-breakpoint-map('base', 'sm', 'md', 'lg', 'xl', '2xl');
+}
 
-  // Filter group
-  --#{$toolbar}__group--m-filter-group--spacer: var(--#{$toolbar}__group--spacer);
-  --#{$toolbar}__group--m-filter-group--space-items: 0;
+:root,
+[data-theme="#{$toolbar}"] {
```
thatblindgeye, line 12: Is this meant to be here, or should be the `:where` pseudo?

src/patternfly/components/OverflowMenu/overflow-menu.scss
mattnolting: Nice catch! We don't.

src/patternfly/components/ActionList/action-list.scss
mattnolting: In past iterations of modifier implementation, updating variable vs property, especially where inheritance could be a potential issue, can result in unwanted results. Since this is a layout variable, seems best to update the property rather than var.

src/patternfly/components/OverflowMenu/overflow-menu.scss
mattnolting: Doesn't look like the workspace is currently generating css variables correctly as is, so that requires a followup. 

@thatblindgeye Great point, in this case, I agree with you. In other cases, where something like `menu` shares `row-gap` values, I much prefer applying this fallback method.

src/patternfly/components/Toolbar/toolbar.scss
mattnolting: This is intentional in that, `:where(.component-root)` adds the component scope to each instance which overrides any variable updates provided by a parent component. @mcoker, @srambach and I talked about this last week and concluded that `[data-theme="#{$component}"]`, where components could be potentially nested, adds an optional, non-root component namespaced "reset" option separate from the component selector. This allows us to `reset` any component/child component when needed.

thatblindgeye (commented):
Just a question related to resolved comment above. Seems like there could be other instances of using fallbacks where we could (should?) instead define vars and not need a fallback elsewhere in the various scss files.

Action list for example has line 32 `row-gap: var(--#{$action-list}__group--RowGap, var(--#{$action-list}--RowGap));`, 

> In other cases, where something like menu shares row-gap values, I much prefer applying this fallback method.

Would you mind providing a quick example of this? Not sure if it's me just not understanding or it still feeling like a Monday morning (or both) 😆

mattnolting:
@thatblindgeye Using `menu` as an example, item, header, group-title, breadcrumb, and footer share `menu__item`'s padding values. We could list these by instance, but the approach essentially does exactly that. It eliminates four lines of code for each use. In this example, ~ 36 lines of code are redundant. 

Screenshot 2024-01-16 at 12 06 20 PM

Screenshot 2024-01-16 at 11 53 42 AM

Screenshot 2024-01-16 at 11 54 16 AM

Screenshot 2024-01-16 at 11 53 59 AM

Screenshot 2024-01-16 at 11 54 21 AM

Screenshot 2024-01-16 at 11 54 34 AM

Screenshot 2024-01-16 at 11 54 39 AM

Does that address your question?

thatblindgeye (approved):
Ah gotcha! Thanks for the explanation, 💯% answers it

srambach (approved):
😵‍💫

PR #6060, opened 2023-11-14
feat(breadcrumb): apply tokens

kmcfaul:
First pass at this!

I couldn't find an analog for `$pf-v5-global--spacer--form-element` in the new design tokens (equates to 6px), should I use `--pf-t--global--spacer--sm` (8px) and subtract 2 inline or is there another token I'm overlooking?

I also noticed that `--pf-t--global--link--text-decoration` was also set to `--pf-t--global--text-decoration--200` (same as hover state) and it should probably be `--pf-t--global--text-decoration--100` instead? I adjusted it in this PR but am unsure if I'm right and if so whether that's the right location to fix it.

Beyond those token specific questions there are some differences I'm noticing between the css and figma that I'm less sure about editing (such as figma using padding over margin, and line height ending up different despite using the analogous token) so would love some feedback

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -2,25 +2,25 @@
 
 .#{$breadcrumb} {
   // Breadcrumb item
-  --#{$breadcrumb}__item--FontSize: var(--#{$pf-global}--FontSize--sm);
-  --#{$breadcrumb}__item--LineHeight: var(--#{$pf-global}--LineHeight--sm);
-  --#{$breadcrumb}__item--MarginRight: var(--#{$pf-global}--spacer--sm);
+  --#{$breadcrumb}__item--FontSize: var(--pf-t--global--font--size--body--sm);
+  --#{$breadcrumb}__item--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$breadcrumb}__item--MarginRight: var(--pf-t--global--spacer--sm);
 
   // Breadcrumb divider
-  --#{$breadcrumb}__item-divider--Color: var(--#{$pf-global}--icon--Color--dark);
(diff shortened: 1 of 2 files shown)


kmcfaul:
@srambach @lboehling @andrew-ronaldson Want to confirm that the `--pf-t--global--link--text-decoration` is meant to be `none` and not `underline` (it was set to underline and this PR changes it to none so as to match the figma).

srambach:
I think the difference in spacing is from the caret icon, which is square in Figma but is just the size of the icon here. We just talked about how this should be fixed by adoption of the new icon set, I believe - @mcoker @lboehling is that correct?

That said, I also think the flex container could be changed here to use gap rather than using a margin on each item. @kmcfaul is that something you'd feel comfortable tackling? We can do a follow-on if not.

kmcfaul:
@srambach Opened https://github.com/patternfly/patternfly/issues/6097 to follow up with the gap

andrew-ronaldson:
> @srambach @lboehling @andrew-ronaldson Want to confirm that the `--pf-t--global--link--text-decoration` is meant to be `none` and not `underline` (it was set to underline and this PR changes it to none so as to match the figma).

In the Figma file I have a underline on the hover state of the interactive breadcrumb elements.

srambach (approved):
🍞

src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -1,36 +1,38 @@
 // @debug $breadcrumb; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
-.#{$breadcrumb} {
+:root, .#{$breadcrumb} {
```
thatblindgeye, line 3: Should this be just `:root`?

thatblindgeye (commented):
Just one question below, otherwise looks good

PR #6063, opened 2023-11-15
feat(ExpandableSection): added tokens

### LINKED ISSUE #5729: Consume tokens: Accordion and Expandable section
[Figma design for accordion](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A3&mode=design&t=doZDRUaxE4vsn5dD-1)

[Figma design for expandable section](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A24&mode=design&t=doZDRUaxE4vsn5dD-1)

thatblindgeye:
Work towards #5729 

[Expandable section preview](https://patternfly-pr-6063.surge.sh/components/expandable-section)

Still need to go through examples/documentation and make any necessary updates. 

Some comments/clarification needed:

- ~~For the toggle text, Figma uses a `global--color--brand--default` token, which I've applied in this PR. Previously we were using a `--#{$pf-global}--link--Color`, so just curious if we should be using the brand token or the `--pf-t--global--text--color--link--default` token.~~ EDIT: per design we should use the brand token here
- ~~For the disclosure variant, the background on hover changes. Is the intent that it also changes when expanded, or should the background only change on hover when not expanded? If the former, then would need input whether the border radius should remain as currently this is what it looks like on hover + expanded (image is showing the focus state which is similar to hover, hover just doesn't have the black outline):~~ EDIT: Per design, trying out applying the border radius only to top left and top right of the toggle for the disclosure variant when expanded
  image

- ~~In the designs the detached variant looks like it's meant to be encompassed by a single expandable section wrapper (based on the disclosure + detached design), however the current implementation has the toggle and content in their own wrappers, inside of stack layouts. Should this still be the case, or are we wanting to/should we move towards more of a "toggle position: top or bottom" setup? cc @mcoker since it looks like you originally added the detached example.~~ EDIT: not supporting a detached + disclosure variant
- Does the `pf-m-display-lg` class/verbiage still make sense or should that be updated to `pf-m-disclosure` or some other class?

### DIFF
--- a/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
+++ b/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
@@ -17,7 +17,17 @@ cssPrefix: pf-v5-c-expandable-section
 
 ### Expanded
 ```hbs
-{{#> expandable-section expandable-section--IsExpanded="true"}}
+{{#> expandable-section expandable-section--IsExpanded=true}}
+  {{> expandable-section-toggle}}
+  {{#> expandable-section-content}}
+    This content is visible only when the component is expanded.
+  {{/expandable-section-content}}
+{{/expandable-section}}
+```
+
+### Indented
+```hbs
+{{#> expandable-section expandable-section--IsExpanded=true expandable-section--IsIndented=true}}
   {{> expandable-section-toggle}}
   {{#> expandable-section-content}}
     This content is visible only when the component is expanded.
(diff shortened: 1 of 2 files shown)


srambach (approved):
LGTM

mcoker (approved):
LGTM! 🐸

PR #6065, opened 2023-11-16
feat(Popover): applied tokens

### LINKED ISSUE #5732: Consume tokens: Tooltip and Popover
```[tasklist]
### Tasks
- [ ] Tooltip tokens
- [ ] Popover tokens
```

[Figma designs for tooltip and popover](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A40&mode=design&t=boxqVg1hn5hO0weV-1)

thatblindgeye:
Work towards #5732 

[Popover preview](https://patternfly-pr-6065.surge.sh/components/popover)

In addition to the below issues, will leave some comments for clarification in the files

- ~~The color contrast of the Warning and Custom variants colors fails in light mode; the text color is too light against the white background.~~ EDIT: use default/regular text color
- ~~The no padding variant ends up with the text visually outside of the container, due to the border radius (maybe just an example update? or a non-issue since we mention that it's intended for content that has its own padding and should be up to the consumer).~~ EDIT: This will be up to the consumer (could update the example)
  Popover without padding


src/patternfly/components/Popover/popover.scss
```diff
@@ -2,37 +2,40 @@
...
+  --#{$popover}__content--PaddingLeft: var(--pf-t--global--spacer--md);
+  --#{$popover}__content--BorderRadius: var(--pf-t--global--border--radius--medium);
 
   // Arrow
-  --#{$popover}__arrow--Width: var(--#{$pf-global}--arrow--width-lg);
-  --#{$popover}__arrow--Height: var(--#{$pf-global}--arrow--width-lg);
-  --#{$popover}__arrow--BoxShadow: var(--#{$pf-global}--BoxShadow--lg);
-  --#{$popover}__arrow--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
+  --#{$popover}__arrow--Width: #{pf-size-prem(25px)};
+  --#{$popover}__arrow--Height: #{pf-size-prem(25px)};
```
thatblindgeye, line 36: There's no arrow tokens, so I just took the value of `--#{$pf-global}--arrow--width-lg` which is what Tooltip has for similar CSS vars

src/patternfly/components/Popover/popover.scss
```diff
@@ -47,30 +50,31 @@
...
 
   // Header
-  --#{$popover}__header--MarginBottom: var(--#{$pf-global}--spacer--sm);
+  --#{$popover}__header--MarginBottom: var(--pf-t--global--spacer--sm);
 
   // Title text
-  --#{$popover}__title-text--LineHeight: var(--#{$pf-global}--LineHeight--md);
-  --#{$popover}__title-text--FontFamily: var(--#{$pf-global}--FontFamily--heading);
-  --#{$popover}__title-text--FontSize: var(--#{$pf-global}--FontSize--md);
+  --#{$popover}__title-text--LineHeight: var(--pf-t--global--font--line-height--heading);
```
thatblindgeye, line 61: Could use another pair of eyes, but it looks like the alignment of the icon and title text is a little off with this line height, whereas if I used the `...--body` line height it looks more in line. Which line height should be used here?

src/patternfly/components/Popover/popover.scss
```diff
@@ -47,30 +50,31 @@
   --#{$popover}__arrow--m-left--Rotate: 45deg;
 
   // Close
-  --#{$popover}__close--Top: calc(var(--#{$popover}__content--PaddingTop) - var(--#{$pf-global}--spacer--form-element)); // align top of button with top of text
-  --#{$popover}__close--Right: calc(var(--#{$popover}__content--PaddingRight) - var(--#{$pf-global}--spacer--md)); // align right of content
-  --#{$popover}__close--sibling--PaddingRight: var(--#{$pf-global}--spacer--2xl);
+  --#{$popover}__close--Top: calc(var(--#{$popover}__content--PaddingTop) - var(--pf-t--global--spacer--sm)); // align top of button with top of text
+  --#{$popover}__close--Right: calc(var(--#{$popover}__content--PaddingRight)); // align right of content
```
thatblindgeye, line 54: For the Top var, there is no form element token so I'm using the sm spacer token for now. If there's a better alternative I can update.

For the Right var, I just removed the calculation as with it (after updating to the new spacer--md token) the button looked too far to the right:

Popover close button aligned further to the right

src/patternfly/components/Popover/popover.scss
thatblindgeye: Update to same size as Tooltip

src/patternfly/components/Popover/popover.scss
thatblindgeye: Update the title text to body--lg and remove the title text line height

thatblindgeye:
Updated the no padding example slightly, applying custom CSS to a div passed to the popover body and giving it a bit more padding than normal just to have it standout.

mcoker (approved):
Looks like the arrow position needs to be adjusted, and another nit about the example - otherwise LGTM! 🐷👍

src/patternfly/components/Popover/popover.scss
```diff
@@ -45,32 +41,37 @@
   --#{$popover}__arrow--m-left--TranslateX: 50%;
   --#{$popover}__arrow--m-left--TranslateY: -50%;
   --#{$popover}__arrow--m-left--Rotate: 45deg;
+  --#{$popover}__arrow--m-inline-top--Top: var(--pf-t--global--border--radius--medium);
+  --#{$popover}__arrow--m-inline-bottom--Bottom: var(--pf-t--global--border--radius--medium);
+  --#{$popover}__arrow--m-block-left--Left: var(--pf-t--global--border--radius--medium);
+  --#{$popover}__arrow--m-block-right--Right: var(--pf-t--global--border--radius--medium);
```
thatblindgeye, line 47: Not 100% sure how I feel about the naming so open to other ideas. Could also just apply the border radius token directly where the top/right/bottom/left vars are reassigned rather than adding new css vars, but since we may want to do something similar in Tooltip it would make sense to do the same here.

srambach (commented):
Just 1.5 comments 😆

src/patternfly/components/Popover/popover.scss
```diff
@@ -45,32 +41,37 @@
...
   // Header
-  --#{$popover}__header--MarginBottom: var(--#{$pf-global}--spacer--sm);
+  --#{$popover}__header--MarginBottom: var(--pf-t--global--spacer--sm);
 
   // Title text
-  --#{$popover}__title-text--LineHeight: var(--#{$pf-global}--LineHeight--md);
-  --#{$popover}__title-text--FontFamily: var(--#{$pf-global}--FontFamily--heading);
-  --#{$popover}__title-text--FontSize: var(--#{$pf-global}--FontSize--md);
+  --#{$popover}__title-text--Color: var(--pf-t--global--text--color--regular);
+  --#{$popover}__title-text--FontFamily: var(--pf-t--global--font--family--heading);
```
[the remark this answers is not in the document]
lboehling: yeah the popover title should be using semibold body-large text!

srambach (approved):
⭐

PR #6070, opened 2023-11-20
fix(tokens): adjust font-weight base for headings

srambach:
Result of a discussion with @lboehling
Bold body font weight should be 500.
Heading font weight - for now keep both base tokens at 700. They might adjust in the future.
Remove any uses of the semantic bold heading token to date in favor of the (regular) heading font weight token.


mcoker (approved):

mattnolting (approved):
👍

PR #6071, opened 2023-11-20
feat(form): tokens for form, form control, radio, check

srambach:
I want to get this up for design to look at - @lboehling @andrew-ronaldson 
This has the change to the font-weight for bold body text to be 500.

Note: 
- Input groups and number input aren't included here
- I'm still not sure the labels are the right font, but body text 12pt 500 weight matches figma
- There is a crazy calculation for form group labels that I translated to tokens, but I don't think it's quite right, and in figma, the labels are just top aligned to the form control. Opinions?


lboehling:
thanks sarah! i'm not seeing font weight 500 come through bold at all for the form labels, it seems to match the regular weight. Also now that i'm seeing it live, i wonder if we should use the default body size for the labels. they look larger in figma 🤷‍♀️ i can update design to match if we want to bump that up. otherwise i think this looks great!!

andrew-ronaldson:
Thanks for this @srambach. Couple notes:
Looks like the grid-gap between check/radio and label is set to 16px. I think it should be `var(--global-spacer-sm, 8px)`
Also, I think these items should be aligned vertically in the middle.

srambach:
@andrew-ronaldson Thanks - not sure what happened there. Should be fixed now.

srambach:
Closing for new v6 version

Issue #6072, opened 2023-11-21
Penta - use variable font, create opt-out for legacy/non-variable font


mcoker:
Design is using `font-weight: 500` for bold body text, which is based off of 500 weight in the variable font, and that update is landing in https://github.com/patternfly/patternfly/pull/6070/files and https://github.com/patternfly/patternfly/pull/6071/files. Need to update the defaults to use the variable font, and create an opt-in to use the non-variable font and font-weights.

mcoker:
Closed via https://github.com/patternfly/patternfly/pull/6073

PR #6073, opened 2023-11-21
fix(font): use variable font, add opt-in for legacy font

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6072

I didn't add any base vars for the legacy font families/weights. It's a global opt-in, design should never need to use it - it's just for products if they aren't ready to change their font-weights to use VF, and imagine we'll drop this support with the next major release so I don't think we need to make room for it outside of overriding the existing vars. LMK if that sounds alright!

This probably won't look right until https://github.com/patternfly/patternfly/pull/6070/ goes in, since #6070 updates the bold body text token to use `500`. @lboehling if https://github.com/patternfly/patternfly/pull/6070/ looks OK, we can merge that and rebase this PR so you can see the change.

### DIFF
--- a/src/patternfly/base/_common.scss
+++ b/src/patternfly/base/_common.scss
@@ -18,10 +18,14 @@
 }
 
 // Variable font opt-in
-.#{$pf-prefix}m-vf-font {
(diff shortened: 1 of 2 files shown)


srambach (approved):
🔠 👍

PR #6081, opened 2023-11-28
feat(banner): refactor with tokens

### LINKED ISSUE #6082: Banner tokens

wise-king-sullyman:
Closes #6082

[Figma design for banner](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A8&mode=design&t=doZDRUaxE4vsn5dD-1)

Convenience link: https://patternfly-pr-6081.surge.sh/components/banner


lboehling:
can you make sure that the text and icon colors are using the on-status text color tokens? text--color--status--on-danger, on-success, etc. and icon--color--status--on-danger, on-success, etc.? this should fix this issue. it won't match up to the figma file since the figma file has updated brand colors that haven't been pulled into code yet, but updating these tokens will make sure it's accessible once that's pulled in.

srambach (commented):
LOL, I had the same comment as @lboehling and @thatblindgeye but forgot to submit the review!

srambach (approved):
My only question is for design - whether we want to rearrange the order of the examples and documentation sections to emphasize the status variants over the individual color variants.
Otherwise, ⭐ 🚀

thatblindgeye (approved):
Looks good! Some of the color contrasts seem a little low for me personally, but they aren't failing AA in aXe in Chrome so I wouldn't block + we can always revisit if we receive similar comments from consumers/users

PR #6089, opened 2023-11-30
feat(Pagination): applied tokens

### LINKED ISSUE #6083: Pagination tokens

thatblindgeye:
Closes #6083 

[Pagination preview](https://patternfly-pr-6089.surge.sh/components/pagination)

Most other Penta work relies on other components being updated.
Some notes:

- The PaginationOptionsMenu component file needs to be updated to use MenuToggle, as that is what is used in React since OptionsMenu is deprecated (we're also not targeting options menu within Pagination css which causes the toggle to not become `display: none` and instead causes that + the total items text to display at smaller viewports, screenshot below). Created a followup https://github.com/patternfly/patternfly/issues/6094
  Pagination total items and toggle being incorrectly displayed together

  - That said, I'm wondering whether we should use Select in the React implementation instead of a custom Menu that acts like a Select cc @tlabaj 
- This was discussed during happy hour yesterday 11/29, but currently the top and bottom variants of Pagination have slightly different layouts/styles, especially at smaller viewports. The top variant displays only the total items text (no toggle, no buttons; at least this is the intent), while the bottom variant has the toggle + next/previous buttons (no first/last buttons). The bottom variant also looks to have an incorrect layout that may need to be updated or at least discussed:
  Bottom pagination with possibly incorrect layout
  - So the questions are: do we want the top and bottom pagination to have slightly different style/layout, especially at smaller viewports, or should they be the same unless the consumer decides via the modifier classes? Secondly, was the original intent for the bottom pagination at smaller viewports to be all on one line, like below (maybe not center aligned, though?)
  Bottom pagination styled to have all elements on one line

### DIFF
--- a/src/patternfly/components/Pagination/pagination-options-menu.hbs
+++ b/src/patternfly/components/Pagination/pagination-options-menu.hbs
@@ -1,18 +1,23 @@
-{{#> options-menu options-menu--IsPlainWithText="true" options-menu--HasToggleIcon="true" options-menu-toggle-button--aria-label="Items per page" options-menu--modifier=pagination-options-menu--modifier options-menu--id=pagination-options-menu--id}}
-  {{#> options-menu-toggle}}
-    {{#> options-menu-toggle-text}}
-      <b>1 - 10</b>&nbsp;of&nbsp;<b>{{#if pagination--IsIndeterminate}}many{{else}}36{{/if}}</b>
-    {{/options-menu-toggle-text}}
-  {{/options-menu-toggle}}
-  {{#> options-menu-menu}}
-    {{#> options-menu-menu-item}}
-      5 per page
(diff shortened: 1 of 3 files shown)


thatblindgeye:
The inset example also is not working correctly, may be related to https://github.com/patternfly/patternfly/issues/6093

mcoker (approved):
LGTM - just one comment about removing the page-insets variation, though that can also be a follow up.

tlabaj:
> * That said, I'm wondering whether we should use Select in the React implementation instead of a custom Menu that acts like a Select cc @tlabaj

@thatblindgeye The Select is just a wrapped menu.  What do you the benefit is of using a Select instead of the menu here?

mcoker (approved):
LGTM!

PR #6100, opened 2023-12-04
feat(SimpleList): apply tokens

### LINKED ISSUE #6099: Simple list tokens

kmcfaul:
Closes #6099


src/patternfly/components/SimpleList/simple-list-item-link.hbs
```diff
@@ -1,4 +1,4 @@
-<{{#if simple-list-item-link--IsLink}}a{{else}}button{{/if}} class="{{pfv}}simple-list__item-link{{#if simple-list-item-link--modifier}} {{simple-list-item-link--modifier}}{{/if}}"
+<{{#if simple-list-item-link--IsLink}}a{{else}}button{{/if}} class="{{pfv}}simple-list__item-link{{#if simple-list-item-link--IsLink}} pf-m-link{{/if}}{{#if simple-list-item-link--modifier}} {{simple-list-item-link--modifier}}{{/if}}"
```
mcoker, line 1: 🔥

FYI sometimes we'll write it this way just 'cause it's more readable. Note the `~`'s, which removes any whitespace (new line chars in this case) before it. Up to you if you want to update it.

```suggestion
<{{#if simple-list-item-link--IsLink}}a{{else}}button{{/if}} class="{{pfv}}simple-list__item-link
  {{~#if simple-list-item-link--IsLink}} pf-m-link{{/if}}
  {{~#if simple-list-item-link--modifier}} {{simple-list-item-link--modifier}}{{/if}}"
```

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -1,38 +1,37 @@
...
+  --#{$simple-list}__item-link--PaddingBottom: var(--pf-t--global--spacer--xs);
+  --#{$simple-list}__item-link--PaddingLeft: var(--pf-t--global--spacer--md);
+  --#{$simple-list}__item-link--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
+  --#{$simple-list}__item-link--Color: var(--pf-t--global--text--color--subtle);
+  --#{$simple-list}__item-link--FontSize: var(--pf-t--global--font--size--body);
+  --#{$simple-list}__item-link--m-current--Color: var(--pf-t--global--text--color--regular);
+  --#{$simple-list}__item-link--m-current--BackgroundColor: var(--pf-t--global--background--color--action--plain--selected);
+  --#{$simple-list}__item-link--hover--Color: var(--pf-t--global--text--color--subtle);
+  --#{$simple-list}__item-link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
+  --#{$simple-list}__item-link--MixBlendMode: var(--pf-t--global--background--color--action--plain--hover--blend);
```
thatblindgeye, line 16: Just want to double check, but is it intended to apply this var on the mix-blend-mode property to make the background darker than what the designs show for focus/hover/current?

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -1,38 +1,37 @@
 // @debug $simple-list; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
-.#{$simple-list} {
+:root, .#{$simple-list}__item-link {
```
thatblindgeye, line 3: @mcoker was there some styling issue that this solves? I'm not noticing any difference when removing the item-link selector from here, and just wondering whether we'd want to have CSS vars that aren't related to the item-link being defined within its scope?

thatblindgeye (commented):
Just a couple questions below. Also, should the disabled styling be added still? Mainly wondering because if so then in the designs the background color shown for a disabled item would make the text difficult/impossible to read.

src/patternfly/components/SimpleList/simple-list.scss
mcoker: @thatblindgeye yep, it will cause the background color to adjust based on the color of whatever is behind the item. So if you updated the workspace examples to have the same background color as the current simple list item, the item's background would adjust so it's visible instead of effectively disappearing.

Screenshot 2023-12-08 at 11 28 09 AM
Screenshot 2023-12-08 at 11 28 17 AM
Screenshot 2023-12-08 at 11 29 04 AM
Screenshot 2023-12-08 at 11 29 18 AM

mcoker (approved):
LGTM other than the selector update on the component vars @thatblindgeye mentioned

PR #6103, opened 2023-12-04
chore(tokens): updates tokens from figma

srambach:
This updates the token files with newly generated ones.
Note: Border width tokens are now coming from Figma rather than being manually made and stuffed into the font tokens file.


mcoker (approved):
👍👍

mcoker (approved):
🦃👍

Issue #6062, opened 2023-11-14
Bug - [Form] - Forms make Dropdown components not truncate


Gkrumbach07:
**Describe the problem**
A clear and concise description of the problem. Which components are affected?
This is affecting Dropdown, Select, really any inline component that inside a Form. The problem is that Forms use a grid layout which messes with the width of the form. So in our case, a dropdown value was very very long and needed to be truncated, but becuase it was in a Form, it extended past the bounds.

here is an example from the patternfly docs for dropdown with only a Form added
![image](https://github.com/patternfly/patternfly/assets/12587674/2ab0e5d1-9e6b-4095-b635-8b30af276e97)

**How do you reproduce the problem?**
I have also reproduced the problem with bear min css in this sandbox
https://codepen.io/gkrumbach07/pen/abXyprL?editors=1100

**Expected behavior**
I expect it to truncate like it does not in a Form

**Is this issue blocking you?**
What we had to do was use Truncate component for the dropdown toggle text. this introduced other bugs becuase Truncate has a min-width set which creates a unwanted right padding when the text is short. we fixed this issue with custom css.

**What is your environment?**
 - OS: OSX 14
 - Chrome
 - Version 119.0.6045.123 (Official Build) (arm64)

**What is your product and what release date are you targeting?**

**Any other information?**

wise-king-sullyman:
We think this is resolvable with our current release. @nicolethoen @mcoker will investigate and get back to you!

PR #6109, opened 2023-12-06
fix(form): keep form groups/controls from growing too wide

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6062


srambach (approved):
LGTM 👍

Issue #6007, opened 2023-10-18
Bug - checkbox/radio follow-ups


mcoker:
[BackstopJS.Report.pdf](https://github.com/patternfly/patternfly/files/12977313/BackstopJS.Report.pdf)

* example should be checked - https://core-staging.patternfly.org/components/forms/checkbox#checked
* clickable/selectable input misaligned - https://core-staging.patternfly.org/components/card/#clickable-and-selectable
* second radio should say "Radio disabled checked" - https://core-staging.patternfly.org/components/forms/radio/html/disabled/
* split button checkbox misaligned
  * https://core-staging.patternfly.org/components/menus/menu-toggle/html/split-button-checkbox-with-toggle-text/
  * https://core-staging.patternfly.org/components/menus/menu-toggle/html/split-button-primary/
  * https://core-staging.patternfly.org/components/menus/menu-toggle/html/split-button-secondary/

PR #6110, opened 2023-12-06
fix(card/menu-toggle/check): alignment bugs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6007

Fixes everything except https://core-staging.patternfly.org/components/card/#clickable-and-selectable

Looking at that, here are some issues aside from the misalignment:
* The presence of the checkbox is changing the layout/spacing compared with a card without it (delete the checkbox and watch the layout change)
* The checkbox is using the standalone variation, but the label element is present - the standalone variant should only have the checkbox input as a child.

In those examples, probably the simplest way to fix that is to remove `.pf-m-standalone`? That isn't ideal, but it addresses the issues mentioned above. 

Longer term, I wonder if those checkboxes (in all of the selectable/clickable card variants) can be updated to the standalone variant with the `<label>` wrapping the input, which you can see in the [standalone example in core](https://core-staging.patternfly.org/components/forms/checkbox#standalone-input).

wdyt @srambach? You're probably most familiar with the clickable/selectable cards.

### DIFF
--- a/src/patternfly/components/Card/card--check.hbs
+++ b/src/patternfly/components/Card/card--check.hbs
@@ -1,6 +1,6 @@
-{{#> check check--modifier="pf-m-standalone"}}
-  {{#> check-input check-input--attribute=(concat 'id="' card--id '-check" name="' card--id '-check" aria-labelledby="' card--id ' "')}}{{/check-input}}
-  {{#if card--IsSelectable}}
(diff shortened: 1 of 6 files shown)


mcoker:
@srambach updated and added a custom style to remove the small grid gap that shows up. https://github.com/patternfly/patternfly/pull/6110/commits/3a517dce2dc1fd0a7aea2df35d3d31d4dfa6d8f9

In the hbs, I added `--is

Created an issue to clean this up

PR #6111, opened 2023-12-06
chore(tokens): updating usage of new font tokens

srambach:
This builds on the update of the font tokens coming from Figma. It removes font size tokens from the non-generated font token file, except for heading sizes xs-2xl, which haven't been added to figma. These are still used in a few components when it's not clearly a particular heading level (h tag).


mcoker (approved):
🤖👍

PR #6118, opened 2023-12-06
feat(HelperText): apply tokens

andrew-ronaldson:
Closed https://github.com/patternfly/patternfly/issues/6195

### DIFF
--- a/src/patternfly/components/HelperText/helper-text.scss
+++ b/src/patternfly/components/HelperText/helper-text.scss
@@ -6,31 +6,24 @@
   // item
   --#{$helper-text}__item-icon--Color: var(--pf-t--global--icon--color--regular);
   --#{$helper-text}__item-text--Color: var(--pf-t--global--text--color--regular);
-  --#{$helper-text}__item--FontWeight: var(--pf-t--global--font--weight--body);
+  --#{$helper-text}__item-text--FontWeight: var(--pf-t--global--font--weight--body);
   --#{$helper-text}__item-icon--m-indeterminate--Color: var(--pf-t--global--icon--color--subtle);
   --#{$helper-text}__item-text--m-indeterminate--Color: var(--pf-t--global--text--color--subtle);
   --#{$helper-text}__item-icon--m-warning--Color: var(--pf-t--global--icon--color--status--warning--default);
-  --#{$helper-text}__item-text--m-warning--Color: var(--pf-t--global--text--color--regular);
-  --#{$helper-text}__item--m-warning--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$helper-text}__item-text--m-warning--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$helper-text}__item-icon--m-success--Color: var(--pf-t--global--icon--color--status--success--default);
-  --#{$helper-text}__item-text--m-success--Color: var(--pf-t--global--text--color--regular);
-  --#{$helper-text}__item--m-success--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$helper-text}__item-text--m-success--FontWeight: var(--pf-t--global--font--weight--body--bold);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -45,29 +50,35 @@
...
   display: flex;
+  font-weight: var(--#{$helper-text}__item--FontWeight);
+
 
   &.pf-m-indeterminate {
     --#{$helper-text}__item-icon--Color: var(--#{$helper-text}__item-icon--m-indeterminate--Color);
     --#{$helper-text}__item-text--Color: var(--#{$helper-text}__item-text--m-indeterminate--Color);
     --#{$helper-text}--m-dynamic__item-icon--Color: var(--#{$helper-text}--m-dynamic--m-indeterminate__item-icon--Color);
-    --#{$helper-text}--m-dynamic__item-text--Color: var(--#{$helper-text}--m-dynamic--m-indeterminate__item-text--Color);
+    --#{$helper-text}--m-dynamic__item-text--Color: var(--#{$helper-text}--m-dynamic--m-warning__item-text--Color);
```
mcoker, line 60: is this intentional?

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -45,29 +50,35 @@
 
 .#{$helper-text}__item {
   display: flex;
+  font-weight: var(--#{$helper-text}__item--FontWeight);
```
mcoker, line 53: While I doubt this would be an issue... I think this belongs on `__item-text` since the text is what we want bolded. Say we added something like sub-text later down the road, or a plain action menu, those and any of their children would be bold from this declaration when we probably don't want that as the default.

mcoker (commented):
Just a couple of comments. An aside, we can probably remove all of the text color vars, and just keep the one for indeterminate text to make it grey/subtle, but that can happen later, too.

src/patternfly/components/HelperText/helper-text.scss
andrew-ronaldson: Updated these lines to add the __item-text and removed the unused text-color stuff

mcoker (approved):
LGTM. Question for you and @lboehling - is there a need for dynamic helper text now? Or do we simply have helper text that's shown the way it is here? That's another bit we can remove if it's no longer a way we intend helper text to be used.

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -1,36 +1,34 @@
...
+  --#{$helper-text}--m-dynamic--m-indeterminate__item-text--Color: var(--pf-t--global--text--color--subtle);
+  --#{$helper-text}--m-dynamic--m-warning__item-icon--Color: var(--pf-t--global--icon--color--status--warning--default);
+  --#{$helper-text}--m-dynamic--m-warning__item-text--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$helper-text}--m-dynamic--m-success__item-icon--Color: var(--pf-t--global--icon--color--status--success--default);
+  --#{$helper-text}--m-dynamic--m-error__item-icon--Color: var(--pf-t--global--icon--color--status--danger--default);
 
   // icon
-  --#{$helper-text}__item-icon--MarginRight: var(--#{$pf-global}--spacer--xs);
-
+  --#{$helper-text}__item-icon--MarginRight: var(--pf-t--global--spacer--xs);
```
thatblindgeye, line 29: @mcoker would we want to replace this with a gap on helper-itext__item as a followup instead? May not be viable if we want to consider the possibility of adding other elements that can be rendered inside a `helper-text__item`, but adding anything extra to a helper text may not be ideal

thatblindgeye (approved):
Just a comment below. As part of a followup, since all helper text text is now the same color (instead of red for danger, etc) we should also update the "static" example so that it doesn't use status verbiage; ideally we'd probably want to recommend using icons if a helper text is meant to convey status.

PR #6124, opened 2023-12-08
chore(ExpandableSection): updated Penta docs and focus/hover styling

### LINKED ISSUE #6104: ExpandableSection - clean up docs and styling for Penta
Followup to previous Penta work. ExpandableSection should have instances of the `pf-m-detached` class removed from docs and examples as it is no longer needed.

Additionally, styling for the `color` property should be updated to use the `pf-v5-c-expandable-section__toggle-text` selector, including updating and CSS var names. This will expose the class for use in React again, which will require a followup to revert an update made in [React's core bump PR](https://github.com/patternfly/patternfly-react/pull/9836/files#diff-113a13bb145136c71a6c7bf0f9c9376f89a3fc7f6e9d99f001d6434de4952b6d) (where the class name is being semi-hardcoded )

A nice to have as part of this but not required would also be updating focus/hover/active styling to remove active styles (they aren't being utilized) and condense hover/focus styling a bit, similar to an update made in https://github.com/patternfly/patternfly/pull/6058 where we only declare a hover var and use an `is(:hover, :focus)` selector

**Any other information?**

thatblindgeye:
Closes #6104 

Also removed the active class/styling, which will require removing the isActive prop in React, as it seemed like it wasn't really necessary. If there was a use-case originally for adding it I can readd it back.

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section.scss
+++ b/src/patternfly/components/ExpandableSection/expandable-section.scss
@@ -21,7 +21,7 @@
   // Toggle text
   --#{$expandable-section}__toggle-text--Color: var(--pf-t--global--color--brand--default);
   --#{$expandable-section}__toggle-text--hover--Color: var(--pf-t--global--color--brand--hover);
-  --#{$expandable-section}__toggle--m-expanded__toggle-text--Color: var(--pf-t--global--color--brand--hover);
+  --#{$expandable-section}--m-expanded__toggle-text--Color: var(--pf-t--global--color--brand--hover);
 
 
   // Content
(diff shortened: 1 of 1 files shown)


src/patternfly/components/ExpandableSection/expandable-section.scss
```diff
@@ -23,6 +18,12 @@
   --#{$expandable-section}--m-expanded__toggle-icon--Rotate: 90deg;
   --#{$expandable-section}--m-expanded__toggle-icon--m-expand-top--Rotate: -90deg;
   
+  // Toggle text
+  --#{$expandable-section}__toggle-text--Color: var(--pf-t--global--color--brand--default);
+  --#{$expandable-section}__toggle-text--hover--Color: var(--pf-t--global--color--brand--hover);
+  --#{$expandable-section}__toggle--m-expanded__toggle-text--Color: var(--pf-t--global--color--brand--hover);
```
mcoker, line 24: Just a small nit, `pf-m-expanded` applies to `$expandable-section/.pf-v5-c-expandable-section`, so the `--m-expanded` bit would go on that element in the var name. Also no need for `__toggle` in there.

```suggestion
  --#{$expandable-section}--m-expanded__toggle-text--Color: var(--pf-t--global--color--brand--hover);
```

mcoker (approved):
🦾🤖

PR #6127, opened 2023-12-11
fix(buttons): add tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5998

I updated the examples page to (hopefully) make it easier to see the different variations - I'll back those updates out prior to merge


srambach (commented):
Couple of quick questions just from the preview:
- Do we want any space around the inline link for the focus ring? It's really tight.
- Inline disabled links aren't legible in dark theme - I _think_ it's ultimately getting the on-disabled color rather than the disabled text color?

thatblindgeye (commented):
In addition to the inline link disabled issue mentioned above, the contrast for the other disabled/aria-disabled buttons is failing in Firefox for both light and dark themes due to the background color

mcoker:
@srambach good catch, updated the disabled inline link text/icon colors and added a small outline-offset.

@thatblindgeye I'll leave that for @andrew-ronaldson and @lboehling. As long as the usage of vars all looks correct, I'd say the component should be fine to merge even if we update the contrast of those colors later.

src/patternfly/components/Button/button.scss
```diff
@@ -198,13 +227,14 @@
...
   --#{$button}__progress--Left: var(--pf-t--global--spacer--md);
+  --#{$button}__progress--Color: var(--#{$button}__icon--Color);
   --#{$button}--m-progress--TransitionProperty: padding;
   --#{$button}--m-progress--TransitionDuration: var(--#{$pf-global}--TransitionDuration);
   --#{$button}--m-progress--PaddingRight: calc(var(--pf-t--global--spacer--md) + var(--#{$button}__progress--width) / 2);
   --#{$button}--m-progress--PaddingLeft: calc(var(--pf-t--global--spacer--md) + var(--#{$button}__progress--width) / 2);
   --#{$button}--m-in-progress--PaddingRight: var(--pf-t--global--spacer--md);
   --#{$button}--m-in-progress--PaddingLeft: calc(var(--pf-t--global--spacer--md) + var(--#{$button}__progress--width));
-  --#{$button}--m-in-progress--m-plain--Color: var(--pf-t--global--color--brand--default);
+  --#{$button}--m-in-progress--m-plain--Color: var(--#{$spinner}--Color);
```
thatblindgeye, line 237: Nit: do we want to rely on Spinner's styles here since the other progress button's don't? Just wondering if it'd be confusing that a plain progress button's spinner color would change when the spinner Color var is updated, but none of the other progress button's do. May also be the case that updating the progress--Color var would change the other progress button's spinner color, but not the plain progress button.

thatblindgeye (approved):
✅ Had a comment below, but not really blocking. The above comment regarding contrasts could also be a followup.

src/patternfly/components/Button/button.scss
mcoker: @thatblindgeye great question, I just went with the best I could tell based off of the designs.

@andrew-ronaldson per the figma designs, the spinners in progress buttons look to have the same color as an icon in that button would have. So for example, the spinner component in figma uses the color-brand-default token, but in a link button, the spinner uses icon-color-brand-default, which is the same color that an icon in a link button uses.

In v5, we support spinners in plain buttons and the spinner color is just whatever the spinner component color is. So for that support in penta, I'm setting the plain button's spinner color to the spinner default color (color-brand-default). WDYT? I didn't see a plain progress button in figma, though I could have overlooked it if you know of one.

srambach:
> Loving the buttons! The only thing that I noticed is the icon button. I remember talking to Lucia about a min-width to match the height so it would be more squared off. Screenshot 2023-12-20 at 7 49 07 AM

I'd second this for the purposes of target size.

srambach (approved):
Aside from Andrew's question about min-width and Eric's about the progress spinner color, everything looks good to me. 🚨

mcoker:
@andrew-ronaldson updated! Right now the min-width is set to this math formula based off of the plain button's styles:

`1em * {button's line-height} + top padding + bottom padding`. Hopefully there's a better way to do this, but that will make the element a square by default, regardless the font-size/padding.

mcoker:
@srambach @thatblindgeye updated the min-width, control variation left/right padding, and confirmed with @andrew-ronaldson the plain spinner color should come from the spinner component (we don't want to set it to anything).

thatblindgeye (approved):

mcoker:
@srambach updated the inline link text properties to inherit. Here's an inline link button in a title component

Screenshot 2023-12-20 at 5 35 47 PM

srambach (approved):
I love the way you've used the fallbacks on the inline-link. ✨

PR #6130, opened 2023-12-11
feat(DataList): apply tokens

kmcfaul:
with new v6 branch. old PR: #6107


src/patternfly/components/DataList/data-list.scss
```diff
@@ -467,7 +410,12 @@
 // expandable content
 .#{$data-list}__expandable-content {
   max-height: var(--#{$data-list}__expandable-content--MaxHeight);
+  margin-block-end: 16px;
+  margin-inline-start: 16px;
+  margin-inline-end: 16px;
```
thatblindgeye, line 415: Can't recall if this was gone over during the Penta call, but should these properties be assigned variables instead of hardcoded values?

thatblindgeye (commented):
Also the `themes` directory inside of DataList can be deleted

srambach (commented):
Minor color questions

mcoker (approved):

PR #6141, opened 2023-12-12
fix(banner): add missing custom modifier

### LINKED ISSUE #6140: Penta - Banner is missing custom style modifier
Followup to https://github.com/patternfly/patternfly/pull/6081

It seems that I missed adding a modifier for the custom variant of the banner.

wise-king-sullyman:
Closes #6140 

Convenience link: https://patternfly-pr-6141.surge.sh/components/banner/#status

### DIFF
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -90,6 +90,11 @@
     --#{$banner}--Color: var(--#{$banner}--m-info--Color);
   }
 
+  &.pf-m-custom {
+    --#{$banner}--BackgroundColor: var(--#{$banner}--m-custom--BackgroundColor);
+    --#{$banner}--Color: var(--#{$banner}--m-custom--Color);
+  }
+
   &.pf-m-red {
     --#{$banner}--BackgroundColor: var(--#{$banner}--m-red--BackgroundColor);
     --#{$banner}--Color: var(--#{$banner}--m-red--Color);


srambach (approved):
Good catch!

src/patternfly/components/Banner/banner.scss
```diff
@@ -90,6 +90,11 @@
     --#{$banner}--Color: var(--#{$banner}--m-info--Color);
   }
 
+  &.pf-m-custom {
+    --#{$banner}--BackgroundColor: var(--#{$banner}--m-custom--BackgroundColor);
+    --#{$banner}--Color: var(--#{$banner}--m-custom--Color);
```
srambach, line 95: I believe all these colors should be using the "on-" tokens - e.g. `--global-text-color-status-on-custom-default`

wise-king-sullyman: I'm probably misunderstanding something, but I assign those tokens to the modifier vars [higher in the file](https://github.com/patternfly/patternfly/blob/12db7dae8e8c7a01be8be6d78484d46406a4c9e3/src/patternfly/components/Banner/banner.scss#L41). Do I need to redo something about how I did that?

srambach: No I'm just an idiot today. Sorry, trying to do too many things at once.

srambach (approved):
For reals, looks great.

mcoker (approved):
🍰

Issue #5738, opened 2023-07-05
Consume tokens: Switch, Slider, Spinner, Skeleton


srambach:
[Figma design for slider](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A45&mode=design&t=boxqVg1hn5hO0weV-1)

[Figma design for skeleton and spinner](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A44&mode=design&t=boxqVg1hn5hO0weV-1)

```[tasklist]
### Tasks
- [ ] https://github.com/patternfly/patternfly/pull/6153
- [ ] https://github.com/patternfly/patternfly/pull/6157
- [ ] https://github.com/patternfly/patternfly/pull/6154
- [ ] https://github.com/patternfly/patternfly/pull/6158
```

PR #6154, opened 2023-12-18
feat(Spinner): consumed Penta tokens

thatblindgeye:
Work towards #5738

### DIFF
--- a/src/patternfly/components/Spinner/spinner.scss
+++ b/src/patternfly/components/Spinner/spinner.scss
@@ -1,23 +1,34 @@
 // @debug $spinner; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
+:root,
 .#{$spinner} {
-  --#{$spinner}--diameter: var(--#{$pf-global}--icon--FontSize--xl);
+  --#{$spinner}--diameter: var(--pf-t--global--icon--size--2xl);
   --#{$spinner}--Width: var(--#{$spinner}--diameter);
   --#{$spinner}--Height: var(--#{$spinner}--diameter);
-  --#{$spinner}--Color: var(--#{$pf-global}--primary-color--100);
+  --#{$spinner}--Color: var(--pf-t--global--icon--color--brand--default);
+  --#{$spinner}--AnimationDuration: 1.4s;
+  --#{$spinner}--AnimationTimingFunction: linear;
+  --#{$spinner}--StrokeWidth: 10;
 
+  // path
+  --#{$spinner}__path--StrokeWidth: var(--#{$spinner}--StrokeWidth);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Spinner/spinner.scss
```diff
@@ -1,23 +1,34 @@
...
   // sizes
-  --#{$spinner}--m-sm--diameter: var(--#{$pf-global}--icon--FontSize--sm);
-  --#{$spinner}--m-md--diameter: var(--#{$pf-global}--icon--FontSize--md);
-  --#{$spinner}--m-lg--diameter: var(--#{$pf-global}--icon--FontSize--lg);
-  --#{$spinner}--m-xl--diameter: var(--#{$pf-global}--icon--FontSize--xl);
-
+  --#{$spinner}--m-sm--diameter: var(--pf-t--global--icon--size--md);
+  --#{$spinner}--m-md--diameter: var(--pf-t--global--icon--size--lg);
+  --#{$spinner}--m-lg--diameter: var(--pf-t--global--icon--size--xl);
+  --#{$spinner}--m-xl--diameter: var(--pf-t--global--icon--size--2xl);
```
thatblindgeye, line 21: Do we still need this modifier class since this is the default size of a spinner?

srambach: I lean to the side of leaving the modifier for completeness but I wouldn't fight about it.

srambach (approved):
🍭

src/patternfly/components/Spinner/spinner.scss
```diff
@@ -1,23 +1,34 @@
 // @debug $spinner; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
+:root,
 .#{$spinner} {
-  --#{$spinner}--diameter: var(--#{$pf-global}--icon--FontSize--xl);
+  --#{$spinner}--diameter: var(--pf-t--global--icon--size--2xl);
   --#{$spinner}--Width: var(--#{$spinner}--diameter);
   --#{$spinner}--Height: var(--#{$spinner}--diameter);
-  --#{$spinner}--Color: var(--#{$pf-global}--primary-color--100);
+  --#{$spinner}--Color: var(--pf-t--global--icon--color--brand--default);
```
mcoker, line 8: This seems like the right value but in figma it's defined as the border--color--brand token. cc @lboehling @andrew-ronaldson do you know which is right?

src/patternfly/components/Spinner/spinner.scss
mcoker: IMO the "inline" spinner should be the default, then you would opt-in to a size variation. It seems odd that the default spinner is so large.

mcoker (approved):

PR #6155, opened 2023-12-18
feat(background-image): apply tokens

### LINKED ISSUE #6145: Background image tokens
@lboehling is going to look into making a background image that doesn't need to span top to bottom so that it can live in the lower right corner like the current one does. This makes it more resilient when resizing the window to different aspect ratios.

srambach:
Fixes #6145


mcoker (approved):

mcoker (approved):
LGTM! Do you also want to use this in the about modal? https://patternfly-pr-6155.surge.sh/components/about-modal/html/basic/

andrew-ronaldson (approved):
oooooooo, fancy!
Approved!

PR #6157, opened 2023-12-18
feat(Slider): consumed Penta tokens

thatblindgeye:
Work towards #5738

### DIFF
--- a/src/patternfly/components/Slider/slider.scss
+++ b/src/patternfly/components/Slider/slider.scss
@@ -60,9 +60,9 @@
   
   // value
   --#{$slider}__value--MarginLeft: var(--pf-t--global--spacer--md);
-  --#{$slider}__value--c-form-control--width-base: 3.5ch;
+  --#{$slider}__value--c-form-control--width-base: calc(var(--#{$form-control}--PaddingLeft) + var(--#{$form-control}--PaddingRight) + #{pf-size-prem(20px)}); // form control base width + number input spinner width
   --#{$slider}__value--c-form-control--width-chars: 3;
-  --#{$slider}__value--c-form-control--Width: calc(var(--#{$slider}__value--c-form-control--width-base) + (var(--#{$slider}__value--c-form-control--width-chars) * 1ch));
+  --#{$slider}__value--c-form-control--Width: calc(var(--#{$slider}__value--c-form-control--width-base) + var(--#{$slider}__value--c-form-control--width-chars) * 1ch);
   --#{$slider}__value--m-floating--TranslateX: -50%;
   --#{$slider}__value--m-floating--TranslateY: -100%;
   --#{$slider}__value--m-floating--Left: var(--#{$slider}--value);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Slider/slider.scss
```diff
@@ -1,87 +1,90 @@
...
   --#{$slider}__thumb--hover--BoxShadow: var(--#{$slider}__thumb--BoxShadow--base);
   --#{$slider}__thumb--focus--BoxShadow: var(--#{$slider}__thumb--BoxShadow--base);
   --#{$slider}__thumb--active--BoxShadow:
     var(--#{$slider}__thumb--BoxShadow--base),
-    0 0 2px 5px var(--#{$pf-global}--active-color--200);
-  --#{$slider}__thumb--before--Width: var(--#{$pf-global}--target-size--MinWidth);
-  --#{$slider}__thumb--before--Height: var(--#{$pf-global}--target-size--MinHeight);
-  --#{$slider}__thumb--before--BorderRadius: var(--#{$pf-global}--BorderRadius--lg);
+    0 0 2px 5px var(--pf-t--global--color--nonstatus--blue--clicked);
+  --#{$slider}__thumb--before--Width: 44px;
```
srambach, line 55: My only question is whether we want to keep using the global target size here (maybe just for now) and I would guess we might want to make this a dimension token in the future?

srambach (approved):

andrew-ronaldson (commented):
I saw the height of the rail is `0.25rem` which technically matches the token but wanted to see if we are explicitly defining the variables as `px` or `rem`.

andrew-ronaldson (approved):

src/patternfly/components/Slider/slider.scss
```diff
@@ -1,87 +1,90 @@
...
+  --#{$slider}__thumb--Top: calc(var(--#{$slider}__rail-track--Height) / 2 + var(--pf-t--global--spacer--md));
   --#{$slider}__thumb--Width: #{pf-size-prem(16px)};
   --#{$slider}__thumb--Height: #{pf-size-prem(16px)};
   --#{$slider}__thumb--Left: var(--#{$slider}--value);
-  --#{$slider}__thumb--BackgroundColor: var(--#{$pf-global}--primary-color--100);
+  --#{$slider}__thumb--BackgroundColor: var(--pf-t--global--color--brand--default);
   --#{$slider}__thumb--TranslateX: -50%;
   --#{$slider}__thumb--TranslateY: -50%;
-  --#{$slider}__thumb--BorderRadius: var(--#{$pf-global}--BorderRadius--lg);
+  --#{$slider}__thumb--BorderRadius: var(--pf-t--global--border--radius--large);
```
mcoker, line 46: I think a circle border radius would be useful - cc @lboehling. The value would be "50%" and it will make any square a circle, regardless of size.

mcoker (approved):

Issue #5711, opened 2023-07-05
Consume tokens in: Panel, Drawer, Sidebar, Tab content


srambach:
[Figma design for drawer](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A15&mode=design&t=doZDRUaxE4vsn5dD-1)

```[tasklist]
### Tasks
- [ ] https://github.com/patternfly/patternfly/pull/6160
- [ ] https://github.com/patternfly/patternfly/pull/6183
- [ ] https://github.com/patternfly/patternfly/pull/6178
- [x] #6161
```

PR #6160, opened 2023-12-19
feat(Panel): consumed Penta tokens

thatblindgeye:
Work towards #5711 

Right now this just updates vars to use tokens since the Drawer designs in Figma aren't totally 1:1 for what a Panel is.

Should we remove the bordered and raised styling for a panel, and follow more closely the inline, full page Drawer designs? That looks to have no border and always has a box shadow around it, but I'm not sure if we would always want the panel to have a box shadow.

### DIFF
--- a/src/patternfly/components/Panel/panel.scss
+++ b/src/patternfly/components/Panel/panel.scss
@@ -1,53 +1,59 @@
 // @debug $panel; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 
+:root,
 .#{$panel} {
   --#{$panel}--Width: auto;
   --#{$panel}--MinWidth: auto;
   --#{$panel}--MaxWidth: none;
   --#{$panel}--ZIndex: auto;
-  --#{$panel}--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
+  --#{$panel}--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$panel}--BoxShadow: none;
 
   // border
   --#{$panel}--before--BorderWidth: 0;
-  --#{$panel}--before--BorderColor: var(--#{$pf-global}--BorderColor--100);
+  --#{$panel}--before--BorderColor: var(--pf-t--global--border--color--default);
+
+  // secondary modifier
+  --#{$panel}--m-secondary--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
 
   // bordered
-  --#{$panel}--m-bordered--before--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$panel}--m-bordered--before--BorderWidth: var(--pf-t--global--border--width--regular);
 
   // raised
-  --#{$panel}--m-raised--BoxShadow: var(--#{$pf-global}--BoxShadow--md);
-  --#{$panel}--m-raised--ZIndex: var(--#{$pf-global}--ZIndex--sm);
-  --#{$panel}--m-raised--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
+  --#{$panel}--m-raised--BoxShadow: var(--pf-t--global--box-shadow--md);
+  --#{$panel}--m-raised--ZIndex: var(--pf-t--global--Zindex--sm);
+  --#{$panel}--m-raised--BackgroundColor: var(--pf-t--global--background--color--primary--default);
 
   // header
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Panel/panel.scss
```diff
@@ -1,53 +1,59 @@
...
+  --#{$panel}__footer--PaddingTop: var(--pf-t--global--spacer--md);
+  --#{$panel}__footer--PaddingRight: var(--pf-t--global--spacer--md);
+  --#{$panel}__footer--PaddingBottom: var(--pf-t--global--spacer--md);
+  --#{$panel}__footer--PaddingLeft: var(--pf-t--global--spacer--md);
   --#{$panel}__footer--BoxShadow: none;
 
   // scrollable
   --#{$panel}--m-scrollable__main--MaxHeight: #{pf-size-prem(300px)};
   --#{$panel}--m-scrollable__main--Overflow: auto;
   --#{$panel}--m-scrollable__footer--BoxShadow: 0 #{pf-size-prem(-5px)} #{pf-size-prem(4px)} #{pf-size-prem(-4px)} #{rgba($pf-v5-color-black-1000, .16)}; // insets the shadow so it doesn't show on left/right sides
```
thatblindgeye, line 53: Is there a token we want to use for this hardcoded rgba?

andrew-ronaldson:
Looks like there is some dark theme tokens shenanigans. 
Screenshot 2023-12-19 at 12 38 43 PM

thatblindgeye:
@andrew-ronaldson ah that should be covered by the Drawer portion of the linked issue. This PR only covers the [Panel component](https://patternfly-pr-6160.surge.sh/components/panel). There are some Toolbar examples that use a panel to contain Menu components where the text isn't working properly in Dark mode, but that should be resolved by the Menu penta work.

src/patternfly/components/Panel/panel.scss
srambach: I'm not sure we'll even want a box-shadow there? If so, @andrew-ronaldson do you want to pick a palette color to replace it since this color will go away, and then @thatblindgeye we can put in a todo or follow up issue so we don't lose this?

srambach (commented):
This looks good other than that variable you called out.

src/patternfly/components/Panel/panel.scss
```diff
@@ -1,53 +1,59 @@
...
-  --#{$panel}--m-bordered--before--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
+  --#{$panel}--m-bordered--before--BorderWidth: var(--pf-t--global--border--width--regular);
 
   // raised
-  --#{$panel}--m-raised--BoxShadow: var(--#{$pf-global}--BoxShadow--md);
-  --#{$panel}--m-raised--ZIndex: var(--#{$pf-global}--ZIndex--sm);
-  --#{$panel}--m-raised--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
+  --#{$panel}--m-raised--BoxShadow: var(--pf-t--global--box-shadow--md);
+  --#{$panel}--m-raised--ZIndex: var(--pf-t--global--Zindex--sm);
+  --#{$panel}--m-raised--BackgroundColor: var(--pf-t--global--background--color--primary--default);
```
mcoker, line 25: I believe the intention for this variation is to act as a panel that would float on top of other stuff, like a dropdown or whatever. Assuming that's true, we'll want to use the [floating background color token](https://github.com/patternfly/patternfly/blob/d244d9f50027a16269548cc204b4e6edfc162dac/src/patternfly/base/tokens/_tokens-default.scss#L240).

The v5 panel was using the floating background color but only in dark theme since that's where it was added - https://github.com/patternfly/patternfly/blob/48c4dc9c315d08989d30b20488d932b77be1172f/src/patternfly/components/Panel/themes/dark/panel.scss#L5

src/patternfly/components/Panel/panel.scss
mcoker: fwiw, with our [box shadow tokens](https://github.com/patternfly/patternfly/blob/d244d9f50027a16269548cc204b4e6edfc162dac/src/patternfly/base/tokens/_tokens-font.scss#L59-L73) we're just using black (`rgba(0 0 0 / 16%)`)

mcoker (commented):
LGTM, though I do think the raised variation should use the floating background token.

Is there a design for panel? I didn't see it in figma.

mcoker (approved):
reviewed w/ @srambach and updated the box shadow and raised background color

PR #6161, opened 2023-12-19
feat(TabContent): consumed Penta tokens

thatblindgeye:
Work towards #5711

### DIFF
--- a/src/patternfly/components/TabContent/examples/TabContent.md
+++ b/src/patternfly/components/TabContent/examples/TabContent.md
@@ -37,18 +37,18 @@ cssPrefix: pf-v5-c-tab-content
 {{/tab-content}}
 ```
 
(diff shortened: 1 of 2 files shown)


srambach (approved):
Preview is still down, but the code looks 💯

andrew-ronaldson (approved):
Ship it!

mcoker (approved):
LGTM! Just to confirm - the tab content previously did not have a background color specified by default, so it would be transparent and the background would come from whatever it's placed inside of. Now the default background is our primary background color, with the secondary variation for a secondary background color.

Is that the intention, or should it just be transparent by default? @andrew-ronaldson

mcoker:
Verified with @andrew-ronaldson that we should just use a transparent background by default.

PR #6167, opened 2023-12-19
chore(menu): add tokens

### LINKED ISSUE #6162: Consume tokens: Menu
Menu

[Figma design for menus](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A16&mode=design&t=doZDRUaxE4vsn5dD-1)

### LINKED ISSUE #6218: Workspace - add support for "and" operator in tests
We have "or" support via the [`ifAny` helper](https://github.com/patternfly/patternfly/blob/fd18c81bf84f67feff5b9941a75da89a3c104a4a/scripts/helpers.mjs#L43-L49) but AFAIK we don't have "and" support. An `ifAll` helper seems like it could be useful.

mattnolting:
closes #6162 
closes #6218

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -11,7 +11,8 @@
   --#{$breadcrumb}__item-divider--FontSize: var( --pf-t--global--font--size--body--sm);
 
   // Breadcrumb link
-  --#{$breadcrumb}__link--PaddingInline: var(--pf-t--global--spacer--sm); // use a mutable value for alignment control
+  --#{$breadcrumb}__link--PaddingInlineStart: var(--pf-t--global--spacer--sm); // use a mutable value for alignment control
+  --#{$breadcrumb}__link--PaddingInlineEnd: var(--pf-t--global--spacer--sm); // use a mutable value for alignment control
   --#{$breadcrumb}__link--Color: var(--pf-t--global--text--color--link--default);
   --#{$breadcrumb}__link--TextDecoration: var(--pf-t--global--link--text-decoration);
   --#{$breadcrumb}__link--hover--Color: var(--pf-t--global--text--color--link--hover);
@@ -54,7 +55,7 @@
   list-style: none;
 
   &:first-child {
-    margin-inline-start: calc(var(--#{$breadcrumb}__link--PaddingInlineStart, var(--#{$breadcrumb}__link--PaddingInline)) * -1); // left align first item in list with list bounds while retaining link padding
+    margin-inline-start: calc(var(--#{$breadcrumb}__link--PaddingInlineStart) * -1); // left align first item in list with list bounds while retaining link padding
   }
 
   &:not(:last-child) {
@@ -74,8 +75,8 @@
 
 // Breadcrumb link
 .#{$breadcrumb}__link {
-  padding-inline-start: var(--#{$breadcrumb}__link--PaddingInlineStart, var(--#{$breadcrumb}__link--PaddingInline)); // use a mutable value for alignment control
-  padding-inline-end: var(--#{$breadcrumb}__link--PaddingInlineEnd, var(--#{$breadcrumb}__link--PaddingInline)); // use a mutable value for alignment control
+  padding-inline-start: var(--#{$breadcrumb}__link--PaddingInlineStart); // use a mutable value for alignment control
+  padding-inline-end: var(--#{$breadcrumb}__link--PaddingInlineEnd); // use a mutable value for alignment control
   font-size: inherit;
   font-weight: var(--#{$breadcrumb}__link--FontWeight);
   line-height: inherit;
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -1,6 +1,5 @@
-// @debug $breadcrumb; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
-
-:root, .#{$breadcrumb} {
+:where(:root),
+:where(.#{$breadcrumb}) {
   // Breadcrumb item
   --#{$breadcrumb}__item--FontSize: var(--pf-t--global--font--size--body--sm);
   --#{$breadcrumb}__item--LineHeight: var(--pf-t--global--font--line-height--body);
@@ -12,6 +11,7 @@
   --#{$breadcrumb}__item-divider--FontSize: var( --pf-t--global--font--size--body--sm);
 
   // Breadcrumb link
+  --#{$breadcrumb}__link--PaddingInline: var(--pf-t--global--spacer--sm); // use a mutable value for alignment control
   --#{$breadcrumb}__link--Color: var(--pf-t--global--text--color--link--default);
   --#{$breadcrumb}__link--TextDecoration: var(--pf-t--global--link--text-decoration);
   --#{$breadcrumb}__link--hover--Color: var(--pf-t--global--text--color--link--hover);
@@ -53,6 +53,10 @@
   white-space: nowrap; // Keeps the item-divider and __link text on the same line
   list-style: none;
 
+  &:first-child {
+    margin-inline-start: calc(var(--#{$breadcrumb}__link--PaddingInlineStart, var(--#{$breadcrumb}__link--PaddingInline)) * -1); // left align first item in list with list bounds while retaining link padding
+  }
+
   &:not(:last-child) {
     margin-inline-end: var(--#{$breadcrumb}__item--MarginRight);
   }
@@ -70,6 +74,8 @@
 
 // Breadcrumb link
 .#{$breadcrumb}__link {
+  padding-inline-start: var(--#{$breadcrumb}__link--PaddingInlineStart, var(--#{$breadcrumb}__link--PaddingInline)); // use a mutable value for alignment control
+  padding-inline-end: var(--#{$breadcrumb}__link--PaddingInlineEnd, var(--#{$breadcrumb}__link--PaddingInline)); // use a mutable value for alignment control
   font-size: inherit;
   font-weight: var(--#{$breadcrumb}__link--FontWeight);
   line-height: inherit;
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -54,8 +54,8 @@
   white-space: nowrap; // Keeps the item-divider and __link text on the same line
   list-style: none;
 
-  &:first-child {
-    margin-inline-start: calc(var(--#{$breadcrumb}__link--PaddingInlineStart) * -1); // left align first item in list with list bounds while retaining link padding
+  &:first-child:has(.#{$breadcrumb}__link) {
+    --#{$breadcrumb}__link--PaddingInlineStart: 0; // remove padding from first child if link
   }
 
   &:not(:last-child) {
--- a/src/patternfly/components/Menu/examples/Menu.md
+++ b/src/patternfly/components/Menu/examples/Menu.md
@@ -50,7 +50,7 @@ import './Menu.css'
           {{/menu-item-main}}
         {{/menu-item}}
       {{/menu-list-item}}
-      {{#> menu-list-item menu-list-item--IsAriaDisabled="true"}}
+      {{#> menu-list-item menu-list-item--IsAriaDisabled=true}}
         {{#> menu-item}}
           {{#> menu-item-main}}
             {{#> menu-item-text}}
@@ -59,8 +59,8 @@ import './Menu.css'
           {{/menu-item-main}}
         {{/menu-item}}
       {{/menu-list-item}}
-      {{#> menu-list-item menu-list-item--IsAriaDisabled="true"}}
-        {{#> menu-item menu-item--IsLink="true"}}
+      {{#> menu-list-item menu-list-item--IsAriaDisabled=true}}
+        {{#> menu-item menu-item--IsLink=true}}
           {{#> menu-item-main}}
             {{#> menu-item-text}}
             Aria-disabled link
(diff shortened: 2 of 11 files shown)


src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -12,6 +12,7 @@
   --#{$breadcrumb}__item-divider--FontSize: var( --pf-t--global--font--size--body--sm);
 
   // Breadcrumb link
+  --#{$breadcrumb}__link--PaddingInlineStart: revert; // use a mutable value for alignment control
```
srambach, line 15: Were changed to this file meant to be in this PR?

srambach (commented):
Couple of code questions and a couple of visual questions -

Should the split button toggles be pill shaped? Also, [this one](https://patternfly-pr-6167.surge.sh/components/menus/menu-toggle#split-button-checkbox) doesn't have the border radius all around on the disabled version.

I think the highlighting isn't working right on disabled menu items with actions/favs?

https://github.com/patternfly/patternfly/assets/19825616/a6b97863-89d7-441d-8cf3-2da6d94dac4f

src/patternfly/components/Menu/menu.scss
```diff
@@ -1,252 +1,216 @@
...
-  --#{$menu}--m-drilldown--c-menu--Top: 0;
-  --#{$menu}--m-drilldown--c-menu--TransitionDuration--transform: var(--#{$pf-global}--TransitionDuration);
-  --#{$menu}--m-drilldown--c-menu--Transition: transform var(--#{$menu}--m-drilldown--c-menu--TransitionDuration--transform);
-
-  // Drilldown content
-  --#{$menu}--m-drilldown__content--TransitionDuration--height: var(--#{$pf-global}--TransitionDuration);
-  --#{$menu}--m-drilldown__content--TransitionDuration--transform: var(--#{$pf-global}--TransitionDuration);
+  --#{$menu}__item-action--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
+  --#{$menu}__item-action--focus--BackgroundColor: var(--pf-t--global--background--color--action--plain--focus);
+  --#{$menu}__item-action--m-selected--BackgroundColor: var();
```
srambach, line 99: Did you mean to do this?

src/patternfly/components/Breadcrumb/breadcrumb.scss
thatblindgeye: Looks like this is causing the focus styles to look like this:

image

Disabling this inline start padding causes the link text to be more centered horizontally in the focus ring

src/patternfly/components/Menu/menu.scss
```diff
@@ -1,252 +1,232 @@
...
+  // * Menu list item
+  --#{$menu}__list-item--Color: var(--pf-t--global--text--color--regular);
+  --#{$menu}__list-item--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
+  --#{$menu}__list-item--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
+  --#{$menu}__list-item--active--BackgroundColor: var(--pf-t--global--background--color--action--plain--clicked);
+  --#{$menu}__list-item--m-disabled--Color: var(--pf-t--global--text--color--disabled);
+  --#{$menu}__list-item--m-danger--Color: var(--pf-t--global--text--color--status--danger--default);
+  --#{$menu}__list-item--m-load__item--Color: var(--pf-t--global--text--color--link--default);
+  --#{$menu}__list-item--has--menu-action--PaddingRight: var(--pf-t--global--spacer--sm);
+  --#{$menu}__list-item--child--not-active--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
```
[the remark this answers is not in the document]
mattnolting: This prevents parent siblings from changing background on hover

src/patternfly/components/Menu/menu.scss
```diff
@@ -1,252 +1,232 @@
...
-  // Search
-  --#{$menu}__search--PaddingTop: var(--#{$pf-global}--spacer--md);
-  --#{$menu}__search--PaddingRight: var(--#{$pf-global}--spacer--md);
-  --#{$menu}__search--PaddingBottom: var(--#{$pf-global}--spacer--md);
+  // * Menu breadcrumb
+  --#{$menu}__breadcrumb--FontSize: var(--pf-t--global--font--size--body--sm);
+
+  // * Menu search
   --#{$menu}__search--PaddingLeft: var(--#{$pf-global}--spacer--md);
+  --#{$menu}__search--PaddingRight: var(--#{$pf-global}--spacer--md);
```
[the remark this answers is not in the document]
mattnolting: I assumed that to be errant. 

Screenshot 2024-01-16 at 11 27 48 AM

andrew-ronaldson:
When I click on the menu examples the borders disappear? 
Screenshot 2024-01-18 at 11 33 52 AM

mattnolting:
> When I click on the menu examples the borders disappear? Screenshot 2024-01-18 at 11 33 52 AM

@andrew-ronaldson Same thing happens on [pf6](https://patternfly-pr-6167.surge.sh/components/menus/menu-toggle) That's part of the menu toggle, separate from menu. I'm updating menu toggle now :)

srambach:
I'm seeing some weirdness with flyout and drilldown menus - the child menu seems to go to all hovered after you are on it and off again...or something like that... Both with mouse and tab focus.

https://github.com/patternfly/patternfly/assets/19825616/9bbe2570-c07b-4cbf-b527-d74e90a79af4

src/patternfly/components/Menu/examples/Menu.md
```diff
@@ -217,7 +217,7 @@ import './Menu.css'
           {{/menu-item-main}}
         {{/menu-item}}
       {{/menu-list-item}}
-      {{#> menu-list-item}}
+      {{#> menu-list-item menu-list-item--IsGroup=true}}
```
thatblindgeye, line 220: Axe dev tools is throwing an error because of this on the flyout examples. We shouldn't need the group role on these `li`'s

thatblindgeye (changes requested):

src/patternfly/components/Menu/menu.scss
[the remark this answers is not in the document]
thatblindgeye: Just wanted to check that this is still needed? Tried removing where this var is being applied locally and at least for the Flyout and Drilldown examples it didn't seem to make a difference

thatblindgeye (approved):
Had a followup question above from a previous comment, but otherwise ✔️

src/patternfly/components/Menu/menu.scss
```diff
@@ -489,298 +385,217 @@
...
+    background-color: var(--#{$menu}__list-item--child--not-active--BackgroundColor);
   }
 
-  &.pf-m-disabled {
-    --#{$menu}__list-item--focus-within--BackgroundColor: transparent;
-
-    pointer-events: none;
-  }
+  &:has(.#{$menu}__item:hover, .#{$menu}__item:focus) {
+    --#{$menu}__list-item--Color: var(--#{$menu}__list-item--hover--Color);
```
mcoker, line 406: `--#{$menu}__list-item--hover--Color` is undefined, though I'm not sure what `--#{$menu}__list-item--Color` does. Setting it doesn't appear to change any styles in the menu as far as I could tell - do you know where it's used?

src/patternfly/components/Menu/menu.scss
```diff
@@ -489,298 +385,217 @@
...
-    cursor: pointer;
+  &.pf-m-selected {
+    --#{$menu}__item-select-icon--Color: var(--#{$menu}__item-select-icon--m-selected--Color);
   }
+
 }
 
+// - Menu item select icon
+.#{$menu}__item-select-icon {
+  color: var(--#{$menu}__item-select-icon--Color, var(--#{$menu}--link--icon--Color));
```
mcoker, line 483: I don't know if using `--#{$menu}--link--icon--Color` is best here, since the icon is just blue, and not a link. Also I see `--#{$menu}--link--icon--Color` is used on the external link icon, and I don't know that those share a concept for a shared var, they seem to just both be blue? WDYT?

src/patternfly/components/Menu/menu.scss
```diff
@@ -1,252 +1,205 @@
...
+  --#{$menu}--RowGap: var(--pf-t--global--spacer--sm);
+  --#{$menu}--Width: auto;
+  --#{$menu}--MinWidth: auto;
+  --#{$menu}--PaddingBlock: var(--pf-t--global--spacer--sm);
+  --#{$menu}--BackgroundColor: var(--pf-t--global--background--color--floating--default);
+  --#{$menu}--BoxShadow: var(--pf-t--global--box-shadow--md);
+  --#{$menu}--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$menu}--OutlineOffset: calc(var(--pf-t--global--border--width--control--default) * -3);
+  --#{$menu}--ZIndex: var(--pf-t--global--Zindex--sm);
+  --#{$menu}--disabled--Color: var(--pf-t--global--text--color--disabled);
```
mcoker, line 13: It seems like `--#{$menu}--disabled--Color` is specific to plain button styling - can we update the var name to be more specific?

mcoker (commented):
Left some comments, though we should probably go through it and see what can be deferred so we can get this component through.

And just a general comment about using fallback vars as the value for a property declaration, and the first var in the `var()` function left undefined for customization... I love the concept, and think we should be doing something like that, but I think we should try and be intentional and consistent if that's the direction we want to go. \

Currently though, it can be hard to know what the role of the primary and any fallback vars are, and when to write them that way, since we haven't established any norms around it. And we'll need to discuss what the definition of a breaking change is, specifically around maintaining the role/function of CSS vars (and not changing that function between breaking changes), since the more we share the role of vars, the more difficult it is to introduce any change in the relationship of shared vars without changing what the function of those vars in the component.

I'd be in favor of just using regular vars as we have in PF5 for the time being, and discuss this along with how we're going to define breaking changes in PF6 after alpha, and introduce any changes once we're confident in the approach and it fits with our overall breaking change strategy. WDYT? cc @srambach

src/patternfly/components/Menu/menu.scss
```diff
@@ -489,298 +385,217 @@
...
-
-  padding-block-start: var(--#{$menu}__list--PaddingTop);
-  padding-block-end: var(--#{$menu}__list--PaddingBottom);
-
-  > .#{$divider} {
-    margin-block-start: var(--#{$menu}__list--c-divider--MarginTop);
-    margin-block-end: var(--#{$menu}__list--c-divider--MarginBottom);
-  }
+// - Menu list
+.#{$menu}__list :where(.#{$divider}:is(li)) {
```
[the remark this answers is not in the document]
mattnolting: Because the divider behaves differently contextually, this makes sense. When in a list, adopt the list formatting, otherwise be full width.

src/patternfly/components/Menu/menu.scss
```diff
@@ -489,298 +385,217 @@
...
+  --#{$breadcrumb}__item--FontSize: var(--#{$menu}__breadcrumb--FontSize);
 
-.#{$menu}__footer {
-  position: relative;
-  padding-block-start: var(--#{$menu}__footer--PaddingTop);
-  padding-block-end: var(--#{$menu}__footer--PaddingBottom);
-  padding-inline-start: var(--#{$menu}__footer--PaddingLeft);
-  padding-inline-end: var(--#{$menu}__footer--PaddingRight);
-  box-shadow: var(--#{$menu}__footer--BoxShadow);
+  z-index: var(--#{$menu}--ZIndex);
```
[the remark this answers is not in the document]
mattnolting: Yes, it's because the breadcrumb dropdown appears behind the menu.

andrew-ronaldson (approved):
approved. thanks

srambach:
> I'd be in favor of just using regular vars as we have in PF5 for the time being, and discuss this along with how we're going to define breaking changes in PF6 after alpha, and introduce any changes once we're confident in the approach and it fits with our overall breaking change strategy. WDYT? cc @srambach

I agree on this - and I'd want to document the scheme of how we'd use undefined variables with fallbacks for future maintainers.

srambach:
Nit - these demos are escaping their workspace area
https://patternfly-pr-6167.surge.sh/components/menus/menu/#with-flyout-menu-top
https://patternfly-pr-6167.surge.sh/components/menus/menu/#with-flyout-menu-left-top

mattnolting:
> Nit - these demos are escaping their workspace area https://patternfly-pr-6167.surge.sh/components/menus/menu/#with-flyout-menu-top https://patternfly-pr-6167.surge.sh/components/menus/menu/#with-flyout-menu-left-top

@srambach examples updated!

src/patternfly/components/Menu/menu.scss
```diff
@@ -1,252 +1,236 @@
...
+}
+
+// - Menu hidden
+[class*="#{$menu}"]:is([hidden]) {
+  display: none;
+}
+
+// - Menu
+.#{$menu} {
+  // TODO: Explore this fallback method
```
srambach, line 194: I think this fallback method is also up in breadcrumb? Can you revert that back also for now?

srambach (approved):
👨‍🍳 fully cooked!

src/patternfly/components/Menu/menu-item-action.hbs
```diff
@@ -1,18 +1,20 @@
...
-    aria-disabled="true"
-  {{/if}}
-  {{#if menu-item-action--IsFavorited}}
-    aria-label="Starred"
-  {{else if menu-item-action--IsFavorite}}
-    aria-label="Not starred"
-  {{/if}}
+<div class="{{pfv}}menu__item-action
+  {{setModifiers
+    menu-list-item--IsAriaDisabled='pf-m-aria-disabled'
```
mcoker, line 3: Do you need this on both the `__item-action` container and the action inside?

src/patternfly/components/Menu/menu.scss
```diff
@@ -1,252 +1,236 @@
...
+  // padding-block-start: var(--#{$menu}--PaddingBlockStart, var(--#{$menu}--PaddingBlock));
+  // padding-block-end: var(--#{$menu}--PaddingBlockEnd, var(--#{$menu}--PaddingBlock));
+  row-gap: var(--#{$menu}--RowGap);
   width: var(--#{$menu}--Width);
   min-width: var(--#{$menu}--MinWidth);
+  padding-block-start: var(--#{$menu}--PaddingBlockStart);
+  padding-block-end: var(--#{$menu}--PaddingBlockEnd);
+  overflow: hidden;
+  color: var(--#{$menu}--Color);
```
mcoker, line 203: Is setting color needed?

src/patternfly/components/Menu/menu.scss
```diff
@@ -489,298 +417,217 @@
...
-// List item
+// - Menu list item
 .#{$menu}__list-item {
-  @include pf-v5-hidden-visible(var(--#{$menu}__list-item--Display));
-
-  position: relative;
+  align-items: start;
+  min-width: 0; // allow list item to shrink for text overflow
   color: var(--#{$menu}__list-item--Color);
   background-color: var(--#{$menu}__list-item--BackgroundColor);
```
mcoker, line 431: Are either of these needed?

mcoker (commented):
Left some comments, though in v5, hovering/focusing anything in a menu list item (the main item and/or any actions in it) would trigger the background change on the list item, and I see that was removed here. In checking with @lboehling, it looks like we want to keep the behavior from v5, and that impacts a lot of my feedback. Let's regroup on Monday and see what the best path forward is to getting menu updated for alpha.

src/patternfly/components/Menu/menu.scss
mattnolting: `background-color` is needed, `color` is not

src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -53,6 +54,10 @@
   white-space: nowrap; // Keeps the item-divider and __link text on the same line
   list-style: none;
 
+  &:first-child {
+    margin-inline-start: calc(var(--#{$breadcrumb}__link--PaddingInlineStart) * -1); // left align first item in list with list bounds while retaining link padding
+  }
```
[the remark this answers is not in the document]
mattnolting: Currently, breadcrumbs are misaligned when link is present in the first child. 

Screenshot 2024-02-03 at 12 10 18 PM

Removing padding from the first link resolves this issue:

Screenshot 2024-02-03 at 12 11 48 PM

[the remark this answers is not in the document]
mcoker: Oh I was asking why the padding was added in the first place. It doesn't exist in v6 currently. This last example is misaligned but that's because the browser/agent styles are adding padding to the breadcrumb link when it's a button. IMO that's just a bug we overlooked since we don't want the default `button` element styles - we'd just remove that (set breadcrumb__link to `padding: 0`). 

Screenshot 2024-02-05 at 2 57 40 PM

This is what I'm seeing in this PR currently

Screenshot 2024-02-05 at 3 01 21 PM

src/patternfly/components/Menu/menu.scss
```diff
@@ -489,298 +417,217 @@
...
 
+  // - Menu item loading
   &.pf-m-loading {
-    --#{$menu}__list-item--hover--BackgroundColor: transparent;
-    --#{$menu}__list-item--focus-within--BackgroundColor: transparent;
-
     justify-content: center;
-    padding-block-start: var(--#{$menu}__list-item--m-loading--PaddingTop);
     overflow: hidden; // prevents spinner rotation from overflowing
+    pointer-events: none;
```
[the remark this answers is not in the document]
tlabaj: @mcoker If you are referring to the spinner, it is currently a span with not `onClick` callback.  At this pont I would not anticipate a use case where we would click on it.

I believe the `tabindex="-1"` is correct.  We do not typically tab through our menu items.  We have been navigating them with the up and down arrow keys.  @thatblindgeye please correct me if I am wrong.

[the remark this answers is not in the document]
thatblindgeye: That's correct regarding the tabindex. We do update the tabindex to 0 in React when an item actually receives focus via the arrow keys (just so the previously focused item will receive focus again upon entering the Menu, unless the menu was closed), but otherwise should be -1.

The/one of the only times we'd need to tab to content in a Menu is if there's actions in a footer/outside the actual menu item list.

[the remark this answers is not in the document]
mcoker: > If you are referring to the spinner, it is currently a span with not onClick callback. At this pont I would not anticipate a use case where we would click on it.

@tlabaj I'm referring to the `button.pf-v5-c-menu__item` that wraps it - you can see it in the screenshot below

Screenshot 2024-02-05 at 3 40 31 PM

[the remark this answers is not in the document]
tlabaj: @mcoker I see. Even thought the spinner is rendered in a button, the `onClick` is only called when the "View more" button is clicked and not the spinner.  If we wanted to allow for that we would need to modify the code.  I would not consider ding that right now unless we had a request for it.  At that point I would probably get UX's opinion on the interaction.

mattnolting:
@lboehling 

Updates:
- Favorited button hover color 
- List item, dark mode hover `mix-blend-mode`
- Removed `button` role from loading `<li>`

srambach:
It doesn't look like selected menu items are showing the check when selected? https://patternfly-pr-6167.surge.sh/components/menus/menu/#option-multi-select

lboehling:
Noticed a few more things! 
- Flyout menus look like they're getting the action plain hover across the entire menu bkg, these should have the same bkg (floating) as the main menu. 
Screenshot 2024-02-09 at 11 25 42 AM

- Drilldown levels 2-X aren't getting the blend mode applied to make them lighter (1st level drilldowns all look correct though)
Screenshot 2024-02-09 at 11 25 55 AM

- the scrolling menu examples have a weird rounded top corner and inset on the first item of the list
Screenshot 2024-02-09 at 11 26 05 AM

mcoker (approved):
LGTM! I'll open a couple of follow-up issues for higher and lower priority updates.

Issue #6170, opened 2023-12-21
Button - bugs and support for span element


mcoker:
Bugs with the no padding variation of plain buttons:
* shouldn't have min-width
* shouldn't have background color

Also we need to be able to use `<span>` on the no padding plain buttons (instead of `<button>`) for when a button flows in text, but we want to be able to keep the button from wrapping on its own.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6171

PR #6171, opened 2023-12-21
fix(button): better support for span button, fix plain bugs

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6170


srambach (approved):
🎄

PR #6177, opened 2024-01-02
feat(menu-toggle): added tokens, new design spec

### LINKED ISSUE #6163: Consume tokens: Menu toggle
Menu toggle

[Figma design for menu toggles](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A16&mode=design&t=doZDRUaxE4vsn5dD-1)

mattnolting:
closes #6163 

Menu toggle: https://patternfly-pr-6177.surge.sh/components/menus/menu-toggle

Text input group will be addressed in followup [6257](https://github.com/patternfly/patternfly/issues/6257)


andrew-ronaldson:
Looks like this split expanded state is using `var(--pf-v5-c-menu-toggle--Color)` .  Can we point to `--pf-t--global--text--color--on-brand--clicked`
 Screenshot 2024-01-25 at 5 48 04 PM

srambach:
In this example, https://patternfly-pr-6177.surge.sh/components/menus/menu-toggle/#plain should the 3rd one show as expanded? And should the 5th one be marked as disabled?

lboehling (changes requested):
Thanks Matt!! looking great. I noticed a few things -- 

- Hover and expanded states for the regular toggles should be using `border—color—hover` and `border—color—clicked` respectively, not brand--hover/clicked.

- Plain with text toggle looks like it needs the mix blend mode token so the hover shows on dark theme. 

- Split action has a blue divider in between on the default state, that should probably just point to border--color--default. image

- Split action and split secondary only have one hover state in the PR, but the action and dropdown toggle should have separate hover/clicked interactions for both of those.

Otherwise looks good! thank you!

thatblindgeye (commented):
- @lboehling  For the split button toggles, should each variant have separate hover styling for the action and the toggle button? The [primary split button action](https://patternfly-pr-6177.surge.sh/components/menus/menu-toggle/#split-button-primary-action) for example only changes the style on hover for the button you're hovering. The default and secondary don't (at least I couldn't really tell)
- For the [typeahead](https://patternfly-pr-6177.surge.sh/components/menus/menu-toggle/#typeahead), the outline for the arrow toggle button gets a little cutoff/overlapped by I assume the text input group border (screenshot below). This is more a nit because I barely noticed it, but I know we've noticed similar issues in the past where the outline gets cutoff so just wanted to note it
    image
- This can be a followup, but it'd be nice to update examples so that we state what toggle is in what state. The primary and secondary examples do this, with the toggle text being "Collapsed" etc, but the plain toggle, split button and such examples aren't as clear unless you dig through the code.

mattnolting:
> * @lboehling  For the split button toggles, should each variant have separate hover styling for the action and the toggle button? The [primary split button action](https://patternfly-pr-6177.surge.sh/components/menus/menu-toggle/#split-button-primary-action) for example only changes the style on hover for the button you're hovering. The default and secondary don't (at least I couldn't really tell)

@thatblindgeye I had the same question. I believe for the Alpha release, we're going to stick with the current interaction

> * For the [typeahead](https://patternfly-pr-6177.surge.sh/components/menus/menu-toggle/#typeahead), the outline for the arrow toggle button gets a little cutoff/overlapped by I assume the text input group border (screenshot below). This is more a nit because I barely noticed it, but I know we've noticed similar issues in the past where the outline gets cutoff so just wanted to note it

@thatblindgeye There are several other changes required for this. In the interest of the Alpha release, I created this followup to address text input group [followup 6257](https://github.com/patternfly/patternfly/issues/6257)

>   image
> * This can be a followup, but it'd be nice to update examples so that we state what toggle is in what state. The primary and secondary examples do this, with the toggle text being "Collapsed" etc, but the plain toggle, split button and such examples aren't as clear unless you dig through the code.

@thatblindgeye I also left many `TODO`s for menu toggle. https://github.com/patternfly/patternfly/pull/6177/files#diff-62015d4798b534885d858ec2db1ee853e67d6eedcfde5142cd4f4364e88cbd52R3-R8 are a few

src/patternfly/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -160,18 +160,20 @@ import './MenuToggle.css'
 
 ### Plain
 ```hbs
-{{#> menu-toggle menu-toggle--IsPlain="true" menu-toggle--attribute='aria-label="Actions"'}}
+{{> menu-toggle menu-toggle--UsesButton=true menu-toggle--IsPlain=true menu-toggle--IsDiv=true}}
+&nbsp;
+{{#> menu-toggle menu-toggle--IsPlain=true menu-toggle--attribute='aria-label="Actions"'}}
```
thatblindgeye, line 165: Do we plan to support both a div wrapper and button wrapper for instances like this? With the div wrapper the focus styling ends up not working correctly, since those styles end up on the `div.pf-ct-c-menu-toggle`. There's still an outline, but the background isn't changing like I'd expect (at least in dark mode).

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -1,173 +1,173 @@
...
-  --#{$menu-toggle}--m-plain--focus--Color: var(--#{$pf-global}--Color--100);
-  --#{$menu-toggle}--m-plain--active--Color: var(--#{$pf-global}--Color--100);
-  --#{$menu-toggle}--m-plain--disabled--Color: var(--#{$pf-global}--disabled-color--200);
-  --#{$menu-toggle}--m-plain--m-expanded--Color: var(--#{$pf-global}--Color--100);
+  --#{$menu-toggle}--hover__toggle-icon--Color: var(--pf-t--global--icon--color--regular);
+
+  // * Menu toggle active
+  --#{$menu-toggle}--active--BackgroundColor: var(--pf-t--global--background--color--control--default);
+  --#{$menu-toggle}--active--BorderWidth: var(--pf-t--global--border--width--button--active);
+  --#{$menu-toggle}--active--BorderColor: var(--pf-t--global--border--color--clicked);
```
thatblindgeye, line 37: @lboehling is there a use case for active/clicked styling of a menu toggle that would differ from its expanded state?

thatblindgeye (commented):
Above sounds good. Few more things I noticed which if they wouldn't be covered by the todo's you've noted, we can add on as a followup. Don't think anything stands out as blocking to me otherwise so I'd be good giving it the ✅ 

- The split button primary: it may not be immediately apparent that there is a checkbox that can be clicked when it's checked:
    image

src/patternfly/components/MenuToggle/examples/MenuToggle.md
mattnolting: Not both, transitionally we need to support `button.pf-v5-c-menu-toggle__button`, but should be using `.pf-v5-c-button`.

srambach (approved):
🏎️

thatblindgeye (approved):
My comments above can be addressed as a followup, can we add those comments to the issue you created to track followups needed? Otherwise lgtm

andrew-ronaldson (approved):
Good stuff! Thanks

Issue #6091, opened 2023-11-30
Modal - Add support for status modifiers to -pf-v5-c-modal__title container


tlabaj:
**Is this a new component or an extension of an existing one?**
Modal

**Describe the feature**
We are creating a composable Modal.  Right now status modifiers (e.g. pf-m-info) are applied to the `pf-v5-c-modal-box` div.  With the composable approach, the `pf-v5-c-modal-box` will not know about the title so we need to able to the modifiers to  `-pf-v5-c-modal__title container`

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
Include screenshots or links to Marvel or other mockups.

**Any other information?**
This is nof the next modal that is being delivered in PF5

PR #6180, opened 2024-01-04
fix(modal): add support for status modifiers on title

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6091

### DIFF
--- a/src/patternfly/components/ModalBox/examples/ModalBox.md
+++ b/src/patternfly/components/ModalBox/examples/ModalBox.md
@@ -289,6 +289,28 @@ import './ModalBox.css'
 {{/modal-example}}
 ```
 
+### Danger alert with modifier on modal title
+The status modifier classes can be applied directly to the modal title element, instead of the parent modal.
+
+```hbs isFullscreen
+{{#> modal-example modal-example--id="danger-alert-title"}}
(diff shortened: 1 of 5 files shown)


srambach (approved):
Looks good - is it worth saying that if the modifier is placed on the title, it will override the modifier on the modal box? I know the react component won't do this, but if someone does it, it might be confusing. Or not, I'm fine if you don't think it's necessary.
Will we phase out allowing the modifier on the modal box?

tlabaj (approved):
LGTM

mcoker:
> The status modifier classes can be applied directly to the modal title element, instead of the parent modal.

@srambach probably not, just from the wording suggesting it goes on one or the other (and not both)? But I could go either way, though I may need some help wording that properly 😅

> Will we phase out allowing the modifier on the modal box?

Yeah, assuming this update works out for the next modal, it would be great to just go with a single approach. There is still a possibility adding the modifier to the title like this is too narrowly scoped (like we also want to draw a status color border around the modal or something like that), but I think we have some time to test that out before the next component is promoted and the current one is deprecated. I went back and added a beta tag to the new example so we can change it if we need to.

WDYT - good to merge or want to discuss any more updates?

srambach (approved):
🚀

PR #6183, opened 2024-01-05
feat(Drawer): consumed Penta tokens

thatblindgeye:
Work towards #5711


src/patternfly/components/Drawer/examples/Drawer.md
```diff
@@ -169,14 +168,27 @@ import './Drawer.css'
 {{/drawer}}
 ```
 
-### Modified panel padding
+### Modified panel body padding
```
thatblindgeye, line 171: Is this a valid use of the pf-m-padding class that we want to showcase? I updated the example to show one way to use the class on panel body content (previously this example wasn't any different from a basic drawer example), but in React we don't have a prop to apply the class on a DrawerPanelBody (though the class could be manually passed in if a consumer knew it could be used).

srambach:
I think the border radius is on the wrong side (switched) for the left and right drawers?
image

srambach:
For the resizeable panel, it would be good to keep the border that separates the sizer from the text, same as in the inline resizeable.

srambach (approved):
Corners and splitter look great now!

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -270,22 +264,21 @@ $pf-v5-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   z-index: var(--#{$drawer}__content--ZIndex);
   flex-basis: var(--#{$drawer}__content--FlexBasis);
   order: 0;
-
-  // background-color: var(--#{$drawer}__content--BackgroundColor);
+  background-color: var(--#{$drawer}__content--BackgroundColor);
```
mcoker, line 267: @lboehling noticed in some of the full page examples, we've lost the rounded corners around the page main content area. For example, from https://patternfly-pr-6183.surge.sh/components/drawer/html-demos/expanded-bottom/

Screenshot 2024-01-12 at 4 34 48 PM

The rounded corners are on `.pf-v5-c-page__main`, but in these full page examples, `.pf-v5-c-page__main` is a child of `.pf-v5-c-drawer__content`, which has this background color added and keeps the corners on page__main from showing up. Opened https://github.com/patternfly/patternfly/issues/6225 to try and address that, since we probably need to test a bit and see what the ideal solution is there.

andrew-ronaldson: Thanks for filing this issue @mcoker 
Lots of products are using drawer in this way so definitely important to fix.

andrew-ronaldson (approved):
Apart from the full page example Coker pointed out I think this first round is ready. Thanks!

mcoker (approved):
&nbsp;🤖👍
`NICE`

PR #6187, opened 2024-01-05
feat(Card): apply tokens

### LINKED ISSUE #6186: Consume tokens in card

srambach:
Fixes #6186 

One known problem - [this demo](http://localhost:8001/components/card/html-demos#status-card-expanded-with-popover) shows a popover that would need to be attached differently in React so that it can escape the bounding box. @mcoker should we remove this demo, or is there a way to fix it in core?


andrew-ronaldson (approved):
LGTM, thanks Sarah!

mcoker (approved):
🥳🍕

PR #6199, opened 2024-01-08
feat(CodeEditor): apply tokens

kmcfaul:
Todos/questions - 
- ~~Header controls look a little off when updated to icon buttons. I put in a couple overrides but it's not quite there. Adjusting the padding/flex properties to match Figma better causes some weird alignment issues with the backgrounds/border. It also seems like we may still need hover styling for the icon buttons?~~
- ~~View shortcuts button hover is covering/removing the header border~~
- Need to figure out how to update the docs to use icon buttons and new pf-m-plain modifier (will probably be a follow up)

Update: should be matching figma.


src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -70,62 +87,63 @@
   }
 }
 
-.#{$code-editor}__controls {
+.#{$code-editor}__header-content {
   display: flex;
+  flex: 1;
+  align-items: flex-end;
```
thatblindgeye, line 93: @lboehling do we want the header content to continue being aligned at the bottom like this:

image

That's how it has been in org. This is what it looks like more centered:

image

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -1,54 +1,72 @@
...
-  --#{$code-editor}__controls--c-button--m-control--focus--Color: var(--#{$pf-global}--Color--100);
-  --#{$code-editor}__controls--c-button--m-control--disabled--after--BorderBottomColor: var(--#{$pf-global}--BorderColor--100);
-
-  // Header
-  --#{$code-editor}__header--before--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
-  --#{$code-editor}__header--before--BorderBottomColor: var(--#{$pf-global}--BorderColor--100);
+  --#{$code-editor}__controls--PaddingInlineStart: var(--pf-t--global--spacer--sm);
+  --#{$code-editor}__controls--Gap: var(--pf-t--global--spacer--xs);
+  --#{$code-editor}__controls--c-button--m-control--Color: var(--pf-t--global--icon--color--regular);
+  --#{$code-editor}__controls--c-button--m-control--disabled--after--BorderBottomColor: var(--pf-t--global--border--color--default);
```
thatblindgeye, line 8: There can be removed since they're not being applied anywhere (unless we do need some overrides to the buttons at all)

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -70,62 +87,63 @@
...
-      &::after {
-        border-block-end-color: var(--#{$code-editor}__controls--c-button--m-control--disabled--after--BorderBottomColor);
-      }
-    }
+  &.pf-m-plain {
+    --#{$code-editor}__header--BackgroundColor: var( --#{$code-editor}__header--m-plain--BackgroundColor);
   }
 }
 
+.#{$code-editor}__controls {
```
thatblindgeye, line 104: Might be worth adding flex-wrap to this, for when  the amount of control buttons causes the container to shove everything off screen:

image

src/patternfly/components/CodeEditor/code-editor.scss
lboehling: yeah, let's center it! good catch!

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -1,54 +1,70 @@
...
-  --#{$code-editor}__header--before--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
-  --#{$code-editor}__header--before--BorderBottomColor: var(--#{$pf-global}--BorderColor--100);
+  --#{$code-editor}__controls--PaddingInlineStart: var(--pf-t--global--spacer--sm);
+  --#{$code-editor}__controls--Gap: var(--pf-t--global--spacer--xs);
+
+  // header
+  --#{$code-editor}__header--before--BorderBottomWidth: var(--pf-t--global--border--width--control--default);
+  --#{$code-editor}__header--before--BorderBottomColor: var(--pf-t--global--border--color--default);
+  --#{$code-editor}__header--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
+  --#{$code-editor}__header--m-plain--BackgroundColor: var(--pf-t--global--background--color--primary--default);
```
thatblindgeye, line 12: Nit: since these are being applied to the header-content element, we could rename the vars

kmcfaul:
Renamed vars, cleaned up padding on header-content, added padding to code block

thatblindgeye (approved):

andrew-ronaldson (approved):
LGTM

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -70,62 +87,66 @@
...
+  &.pf-m-plain {
+    --#{$code-editor}__header-content--BackgroundColor: var( --#{$code-editor}__header-content--m-plain--BackgroundColor);
   }
 }
 
+.#{$code-editor}__controls {
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--#{$code-editor}__controls--Gap);
+  align-items: flex-start;
```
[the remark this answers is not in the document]
mcoker: After talking with @lboehling & @andrew-ronaldson, we don't want the header to grow in height like this, though I don't know that we need to do anything special to keep that from happening... we probably just need to keep the actions on the left from wrapping?

srambach (approved):
LGTM

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -70,62 +87,65 @@
...
+  }
+}
+
 .#{$code-editor}__main {
   position: relative;
   color: var(--#{$code-editor}__main--Color, inherit);
   background-color: var(--#{$code-editor}__main--BackgroundColor);
   border: var(--#{$code-editor}__main--BorderWidth) solid;
   border-color: var(--#{$code-editor}__main--BorderColor);
+  border-radius: 0 0 var(--#{$code-editor}__tab--BorderRadius) var(--#{$code-editor}__tab--BorderRadius);
```
mcoker, line 144: Looks like we have `__tab--BorderRadius` defined on `__main`, `__header-content`, and `__tab`. That means to change any of them, a user will set `--#{$code-editor}__tab--BorderRadius`.

If we decide to give main (as an example) a unique border-radius via `--#{$code-editor}__main--BorderRadius`, then `--#{$code-editor}__tab--BorderRadius` will no longer control `__main`'s border-radius. If a user themed  `--#{$code-editor}__tab--BorderRadius`, that will no longer style `__main`, which is technically a breaking change. 

That logic and what we consider a breaking change may change as we nail down the definition of what what a breaking change is in core now that we have such a rich set of vars to style everything, but for now would you mind creating unique vars for `__main`, `__header-content`, and `__tab`?

mcoker (approved):
LGTM! Left a couple of comments, neither are blockers. If you agree anything should change, we can open a follow-up if issue for post-alpha you'd prefer.

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -70,62 +87,65 @@
...
+  align-items: flex-end;
+  padding-block-start: var(--#{$code-editor}__header-content--PaddingBlockStart);
+  padding-block-end: var(--#{$code-editor}__header-content--PaddingBlockEnd);
+  padding-inline-start: var(--#{$code-editor}__header-content--PaddingInlineStart);
+  padding-inline-end: var(--#{$code-editor}__header-content--PaddingInlineEnd);
+  background-color: var(--#{$code-editor}__header-content--BackgroundColor);
+  border-radius: var(--#{$code-editor}__tab--BorderRadius) 0 0 0;
+
+  &.pf-m-plain {
+    --#{$code-editor}__header-content--BackgroundColor: var( --#{$code-editor}__header-content--m-plain--BackgroundColor);
```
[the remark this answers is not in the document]
kmcfaul: I moved the plain modifier up to `__header`. I think the background color needs to remain on the `header-content` though, because moving it up to the header was creating a grey background when it shouldn't (like when there's no header only the language tab).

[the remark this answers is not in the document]
mcoker: @kmcfaul thanks, perfect!

mcoker (approved):
🐸👍

PR #6203, opened 2024-01-09
feat(Tile): add tokens

### LINKED ISSUE #6197: Consume tokens in tiles

srambach:
Fixes #6197 

Note: A11y checks fail for duplicate ID, which is inside the inline SVG. @mcoker @mnolting @lucia do you have other ideas/a better way to include the SVG?


mcoker (commented):
re: the dupe id's, looks like you just need something like this on the 3 tile examples with the PF svgs

`tile--id="stacked-pf-logo-[default/selected/disabled]"`

mcoker:
Also in the other svg's we can remove a bunch of stuff:

https://github.com/patternfly/patternfly/blob/db5c739a6c4c3fdd0dbd8ec62c9814c74e730a34/src/patternfly/components/Tile/tile-img-openshift.hbs#L1-L4
Only needs to be:

```
<svg width="6.65em" height="1em" viewBox="0 0 1616.7 243.1" style="enable-background:new 0 0 1616.7 243.1;">
```

https://github.com/patternfly/patternfly/blob/db5c739a6c4c3fdd0dbd8ec62c9814c74e730a34/src/patternfly/components/Tile/tile-img-insights.hbs#L1-L2

Only needs to be:

```
<svg width="2.59em" height="1em" viewBox="0 0 140 54">
```

mcoker (approved):
A couple of nits, but LGTM! If it's easy, it would also be nice to figure out what's going on with the PF logo SVG, it's wider than it needs to be. All can be follow ups if you'd prefer!

Screenshot 2024-01-11 at 6 17 08 PM

lboehling:
lgtm! thanks sarah!  mentioned on the group review call that we should consider a follow up issue for making tiles a variant of the card component.

thatblindgeye (approved):

PR #6206, opened 2024-01-10
feat(Tabs): consumed Penta tokens

### LINKED ISSUE #6188: Tokens: Tabs

thatblindgeye:
Closes #6188 

Couple of comments:

- ~The default tab hover/focus background color doesn't seem to work correctly in dark mode. Can be tested by hovering a tab in the first example in light mode (notice the abckground color change), then try in dark mode~ EDIT: resolved after discussion with Lucia and updating some tokens
- The Figma design shows the `More >` component acting similar to a scroll button, with tab items being inside a scrollable container and having a left border to separate it. Functionally this component is meant to replace the scrolling behavior, so currently I did not apply any left border to separate it from the other tab items (in React if you select an item from this More dropdown, it'll then render more like a proper tab item).
- Fixed an issue in the actions examples where tabs actions weren't being properly disabled/aria-disabled.

There's also an odd visual bug within the [help action examples](https://patternfly-pr-6206.surge.sh/components/tabs/#tab-item-actions) that I can't quite pinpoint, specifically on the "Close button example" and "Help button example". The issue only seems to be happening on the main examples page, though; if I open the examples in a new window, this visual bug doesn't occur.

image

The action button ends up visually cut off when focused. This also occurs if I add another action, except the issue remains on the first action while the second action is fine:

image
image

The "Help and close button example" doesn't have this issue:

image


thatblindgeye:
The visual bug mentioned above looks to be possibly related to several things:

- to the `mix-blend-mode`. Removing the new property introduced in the latest commit (line 643 of the Tabs.scss file) and disabling the `--pf-t--global--background--color--action--plain--hover--blend` on root resolves it.
- disabled `overflow-x: auto` on the tabs list element resolves it (or changing the value)
- Adding a z-index to the tabs list resolves it

srambach (approved):
Aside from the mystery dimension that part of the tab goes into, I can't see any problems! 👍

andrew-ronaldson (approved):
LGTM, thanks

lboehling:
lgtm!

Issue #6181, opened 2024-01-04
Tokenize label


mcoker:

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6208

PR #6208, opened 2024-01-11
fix(label): add tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6181

--

TODO - still need to add states (success, error, etc)


andrew-ronaldson (commented):
These look great.
I saw the beta label has this hover state. @lboehling any thoughts? Screenshot 2024-01-12 at 11 06 07 AM

mcoker:
@andrew-ronaldson looks like it's in a plain button, which looks to just be just for the tooltip? I created an org issue to fix that - https://github.com/patternfly/patternfly-org/issues/3859

srambach:
The close button just slightly overlaps the text - is that ok or do we want a little more space? 
image

The focus outline in conjunction with the label shape is a little weird overall but I understand why and don't have any better ideas. :-) 
image

srambach (approved):
Other than the focus outline, which can be adjusted later if we want, looks 👌

andrew-ronaldson (approved):
Perfecto!

PR #6211, opened 2024-01-11
feat(SimpleFileUpload): consumed Penta tokens

### LINKED ISSUE #6209: Update single file upload to use tokens

thatblindgeye:
Closes #6209 

- Removed usage of the form helper text since it seems like from the design the helper text should be tied to File Upload, not Form
- Added a helper text example that makes the text area aria label "Uploaded file content" and tied the helper text to the Upload button. Also made these updates to the "with error" example. As a follow up we should consider making similar changes to other examples, as well as:
  - possibly updating the text input aria label to "File name"; right now it just duplicates the placeholder or value attribute
  - remove the aria-describedby from the text input, as currently it being described by the upload button doesn't provide adequate description

For the "with error" example, is the intent that if a file upload has a type/size restriction, it must be placed in a Form component? Or is this example more like a demo of showing how one could place a file upload inside a form? Right now React only uses Form components for a similar "restrictive" example, but it would seem like a non-restrictive file upload could/should also be placed inside a Form for some sort of submission as well.

### DIFF
--- a/src/patternfly/components/FileUpload/file-upload.scss
+++ b/src/patternfly/components/FileUpload/file-upload.scss
@@ -16,9 +16,6 @@
   --#{$file-upload}--m-drag-hover--BorderColor: var(--pf-t--global--icon--color--brand--default);
   --#{$file-upload}--m-drag-hover--BorderStyle: dashed;
 
-  // File select > Input group
-  --#{$file-upload}__file-select__c-input-group--ColumnGap: var(--pf-t--global--spacer--xs);
-
   // File select > Button
   --#{$file-upload}__file-select__c-button--m-control--OutlineOffset: calc(-1 * var(--pf-t--global--spacer--xs));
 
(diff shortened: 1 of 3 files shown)


lboehling (approved):
LGTM!!

srambach (approved):
🌟 Looks great to me

andrew-ronaldson (approved):
Happy happy joy joy!

kmcfaul (approved):

src/patternfly/components/FileUpload/file-upload.scss
```diff
@@ -1,65 +1,60 @@
...
-      }
     }
   }
 }
 
 .#{$file-upload}__file-select {
+  .#{$input-group} {
+    column-gap: var(--#{$file-upload}__file-select__c-input-group--ColumnGap);
+  }
+
```
mcoker, line 57: If you rebase, I believe you can remove this now?

```suggestion
```

mcoker (changes requested):
Just one nit!

mcoker (approved):
🔥

Issue #6215, opened 2024-01-11
Bug - Button - plain action background buttons need mix blend mode


mcoker:
Couple of bugs:
* Link button background isn't visible on hover if used against secondary background color. Looks like any elements that use the plain action background colors also need the mix-blend-mode token set
* "buttons as links" are presented inline and not honoring block padding.

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6216

PR #6216, opened 2024-01-11
fix(button): fix link background, add base display prop

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6215


srambach (approved):
Fab!!

PR #6219, opened 2024-01-11
feat(DualListSelector): apply tokens

### LINKED ISSUE #5741: Consume tokens: Dual list selector
[Figma design for dual list selector](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A22&mode=design&t=boxqVg1hn5hO0weV-1)

kmcfaul:
Closes #5741

Think I've got it in enough shape for a first pass. The padding around the toggle icon has changed and I'm unsure if it's matching the figma properly. ~~I'm also not sure about how to go about styling the text input group at the top of DualListSelector. We don't currently override any text input group css, and I could add in that for the border radius/background etc. I was I wondering if text input will be done separately, or if we do need overrides in DualListSelector.~~

Edit - per conversation, the text input group that acts as the search input for DLS should be updated separately and no overrides for DLS are needed.

### DIFF
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -65,6 +65,7 @@ $pf-v5-c-dual-list-selector__item--MaxNesting: 10;
   // Controls
   --#{$dual-list-selector}__controls--PaddingRight: var(--pf-t--global--spacer--md);
   --#{$dual-list-selector}__controls--PaddingLeft: var(--pf-t--global--spacer--md);
+  --#{$dual-list-selector}__controls--Gap: var(--pf-t--global--spacer--sm);
 
   // Toggle
   --#{$dual-list-selector}__item-toggle--PaddingRight: var(--pf-t--global--spacer--sm);
@@ -295,8 +296,13 @@ $pf-v5-c-dual-list-selector__item--MaxNesting: 10;
 }
 
 .#{$dual-list-selector}__controls {
+  display: flex;
+  flex-direction: column;
   grid-area: var(--#{$dual-list-selector}__controls--GridArea);
+  gap: var(--#{$dual-list-selector}__controls--Gap);
+  align-items: center;
   align-self: center;
+  justify-content: center;
   padding-inline-start: var(--#{$dual-list-selector}__controls--PaddingLeft);
   padding-inline-end: var(--#{$dual-list-selector}__controls--PaddingRight);
 }
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -1,7 +1,7 @@
 // @debug $dual-list-selector; // check your variable names located in src/patternfly/sass-utilities/component-namespaces.scss
 $pf-v5-c-dual-list-selector__item--MaxNesting: 10;
 
-.#{$dual-list-selector} {
+:where(:root), :where(.#{$dual-list-selector}) {
   // Grid
   --#{$dual-list-selector}__header--GridArea: pane-header;
   --#{$dual-list-selector}__tools--GridArea: pane-tools;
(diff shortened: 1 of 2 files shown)


srambach:
I think it needs a small spacer between those controls in the middle.

kmcfaul:
Updated the controls styling to match figma (making it a flex) and include the gap. LMK if any of those properties aside from gap should also be a variable instead of directly declared.

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -16,79 +16,77 @@ $pf-v5-c-dual-list-selector__item--MaxNesting: 10;
...
-  --#{$dual-list-selector}__list-item-row--BackgroundColor: transparent;
-  --#{$dual-list-selector}__list-item-row--hover--BackgroundColor: var(--#{$pf-global}--BackgroundColor--200);
-  --#{$dual-list-selector}__list-item-row--focus-within--BackgroundColor: var(--#{$pf-global}--BackgroundColor--200);
-  --#{$dual-list-selector}__list-item-row--m-selected--BackgroundColor: var(--#{$pf-global}--BackgroundColor--200);
-  --#{$dual-list-selector}__list-item--m-ghost-row--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
+  --#{$dual-list-selector}__list-item-row--FontSize: var(--pf-t--global--font--size--sm);
+  --#{$dual-list-selector}__list-item-row--BackgroundColor:  var(--pf-t--global--background--color--action--plain--default);
+  --#{$dual-list-selector}__list-item-row--hover--BackgroundColor: var(--pf-t--global--background--color--primary--hover);
+  --#{$dual-list-selector}__list-item-row--BorderRadius: var(--pf-t--global--border--radius--tiny);
+  --#{$dual-list-selector}__list-item-row--focus-within--BackgroundColor: var(--pf-t--global--background--color--primary--hover);
```
thatblindgeye, line 40: We should be able to remove this var and just reuse the hover var forboth hover and focus styles

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -226,6 +225,10 @@ $pf-v5-c-dual-list-selector__item--MaxNesting: 10;
 
       font-weight: var(--#{$dual-list-selector}__list-item-row--m-selected__text--FontWeight);
     }
+
+    &:hover {
+      --#{$dual-list-selector}__list-item-row--BackgroundColor: var(--#{$dual-list-selector}__list-item-row--hover--BackgroundColor);
+    }
```
thatblindgeye, line 231: Might be a duplicate of the style on line 216

kmcfaul: It's needed here as well for `pf-m-selected` rows to also get hover styling.

andrew-ronaldson (approved):
Cool bean. thanks

thatblindgeye (approved):

srambach (commented):
It looks like there should be another md spacer above the menus and the controls according to Figma.

Also, there's a very slight weird spot in the corner of the top item when focused - what do you think about pulling in the focus ring with outline-offset 2 or 4px? @andrew-ronaldson @lboehling 

image

srambach:
One other thing - the disabled controls have the dark button disabled styling, but in Figma the disabled controls are dimmed. @andrew-ronaldson @lboehling do you want the disabled button styling to be overridden here?

lboehling:
@srambach i like the 2px offset for the focus state! (middle option in the screenshot). 

and the disabled icon button styling is getting updated to just use the disabled icon color with no background, so that should be resolved once that change gets in 👍

kmcfaul:
Added the title text size, margin top for the padding between the header and panes, and the outline offset. I left the outline offset as a direct declaration for now because 2px isn't one of the spacer sizes, but lmk if it should equate to a different token or be declared as a variable.

srambach (approved):
Looks good! 📜📜

Issue #6226, opened 2024-01-13
Label - add status variants


mcoker:
error, success, etc

wise-king-sullyman:
Closed by #6227

PR #6227, opened 2024-01-13
fix(label): add status variants

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6226

* Adds status variants
* Renames `label--isRemovable` to `label--IsRemovable`
* Adds an `id` to the label text element when a label is removable, since the close button currently references a non-existent `id` in `aria-labelledby` via something like `<button id="close-btn" arial-label="Remove" aria-labelledby="close-btn label-text">`
* Moves all of the color hbs blocks used between the filled and outline examples into a single hbs file.

1/23/24
* Adds `.pf-m-filled` for the filled styling, which gives us a modifier for filled-specific styling'.
  * We'll need to use `.pf-m-filled` or `.pf-m-outline` for the variant on all labels except for the add and overflow labels as those are structured a bit differently currently. 
* Updates non-status labels to use a generic (cube) icon for the label icon since the old info icon is now uses specifically for info labels.


andrew-ronaldson (commented):
Apologies if I missed a conversation. Should the status lables have the status color on the icon or is it just inheriting the colour from the text? In Figma the icons all have the status color. 

From a design guidelines perspective I think we should recommend using a status label with the icons so they stand out from the non status versions. @lboehling do you have any thoughts here?

thatblindgeye (commented):
Unless it'd be better as a followup, can we update the status labels to use appropriate icons? Right now they're all using the info icon, when success should have a checkmark, warning the triangle `!`, etc.

src/patternfly/components/Label/label--variants.hbs
```diff
@@ -0,0 +1,93 @@
...
+      label--id=(concat label--variants--id '-status-truncated-removable')
+      label-text--value=(concat label--variants--title ' removable, max-width truncation customization')}}
+{{else}}
+  {{> label
+      label--id=(concat label--variants--id '-default')
+      label-text--value=label--variants--title}}
+
+  {{> label
+      label--id=(concat label--variants--id '-icon')
+      label-icon--value="info-circle"
```
thatblindgeye, line 41: Just a nit: since we're showing status labels that would use this icon, we might want to avoid using a "statusy" icon outside a status context. WdYT?

thatblindgeye (approved):
✅  As a followup we could add screen reader text to the status labels, but it might be worth evaluating where else we may want to do that in other components and/or if it makes sense to always render it vs rendering it only if the visible text doesn't convey that status well.

andrew-ronaldson (approved):
Rockstar. LGTM

srambach (approved):
LGTM. Kind of a bummer to have to require `.pf-m-filled` but makes sense. 🏷️

lboehling (changes requested):
looks good! I am noticing that the filled background color for the status labels are referencing a border color token instead of the global/color/status/[success/danger/warning/etc.] tokens

mcoker:
@lboehling great catch!! Looks like all of the filled background color tokens were wrong 🤦‍♂️ Updated, you can see the tokens changed here - https://github.com/patternfly/patternfly/pull/6227/commits/ea51c909cf61e5b2ec91d319b07e6f26f4aaad88

FWIW I just did a search/replace for `--BackgroundColor: var(--pf-t--global--border--color` with `--BackgroundColor: var(--pf-t--global--color`

lboehling:
perfect! thanks @mcoker!!

Issue #6202, opened 2024-01-09
Toggle group: consider mobile presentation


srambach:
@andrew-ronaldson As a followup, we need to consider mobile presentation. Here's an example of toggle group items overflowing and how `toggle-group` can handle this scenario. 

<img width="365" alt="Screenshot 2024-01-06 at 11 04 36 AM" src="https://github.com/patternfly/patternfly/assets/5385435/cffd90a9-f775-4b6c-822a-3b56d18581fd">

**Things to consider**
- [x] Adjacent selected item border. When textual content is less consistent, adjacent selected items appear run together. Should we include border updates here??
- [ ] When toggle items extend past toggle group boundaries, the overflow is currently hidden (default). If we move the border to `toggle-group` instead of items, we can drop top/bottom borders on individual items, and conditionally drop left/right border for `:first/:last-child`. This results in the content being scrollable while presenting design accurately. 
- [ ] When `toggle-group` is a child of `display: flex/grid` its width grows. Should add `width: min-content` to `toggle-group` as well as `white-space: nowrap` to `toggle-group__button`. 

**Toggle group as a child of `display: flex/grid`**

<img width="605" alt="Screenshot 2024-01-06 at 11 35 32 AM" src="https://github.com/patternfly/patternfly/assets/5385435/b1908ec8-e5e4-4f27-9d19-ec462630836a">

**Toggle group button set to `white-space: nowrap`**

<img width="379" alt="Screenshot 2024-01-06 at 11 33 30 AM" src="https://github.com/patternfly/patternfly/assets/5385435/dcb3a42d-10d0-4ba5-9dcf-0279c3642889">

https://codesandbox.io/p/sandbox/infallible-meadow-msz34w?file=%2Fstyle.css%3A12%2C111

_Originally posted by @mattnolting in https://github.com/patternfly/patternfly/issues/6190#issuecomment-1879756168_

andrew-ronaldson:
Design recommendations here https://github.com/patternfly/patternfly-design/issues/1279

PR #6239, opened 2024-01-18
feat(ToggleGroup): add border between like toggles

srambach:
This addresses the adjacent borders in #6202 (but not the mobile is too-wide and cut off problems)

This adds a token that has already been added in Figma, and fixes the block comment on that file that lint complains about.
It splits the border into parts so that the left side can be set to the right color to divide two blocks of the same color. Fallbacks are used so that if the individual border colors aren't set, it falls back to the overall border color.


mcoker (approved):
lgtm!

lboehling (approved):
lgtm!! thank you!

PR #6251, opened 2024-01-25
feat(Wizard): apply tokens

### LINKED ISSUE #5742: Consume tokens: Wizard
[Figma design for wizard](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A51&mode=design&t=boxqVg1hn5hO0weV-1)

srambach:
Fixes #5742 

Uses the link button for styling the wizard nav links, although there are a few overrides.

There is now a line above the footer in all cases; no shadow when sticky. (per design)

There's a TODO to calculate an offset but I think that can wait for another day...


src/patternfly/components/Wizard/wizard.scss
```diff
@@ -27,122 +28,116 @@
   }
 
   // Close
-  --#{$wizard}__close--Top: calc(var(--#{$pf-global}--spacer--lg) - var(--#{$pf-global}--spacer--form-element));
+  --#{$wizard}__close--Top: calc(var(--pf-t--global--spacer--lg) - var(--#{$pf-global}--spacer--form-element));
   --#{$wizard}__close--Right: 0;
```
thatblindgeye, line 32: Would we want to update this from 0? Right now at < xl viewports the button is right up against the edge of the wizard header, which it may just be the context of the example as it would look fine in the context of an actual page most likely.

image

thatblindgeye (commented):
This is a followup/docs thing I noticed, but some examples need to be updated to either include the toggle icon for the second step or make that second step not have substeps:

image

thatblindgeye (approved):

andrew-ronaldson:
Screenshot 2024-02-08 at 12 01 08 PM
The selected step needs the action/plain/clicked/ background change, please and thanks

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -27,122 +28,128 @@
...
-  --#{$wizard}__toggle--BoxShadow: var(--#{$pf-global}--BoxShadow--md-bottom);
-  --#{$wizard}__toggle--PaddingTop: var(--#{$pf-global}--spacer--lg);
-  --#{$wizard}__toggle--PaddingRight: var(--#{$pf-global}--spacer--md);
-  --#{$wizard}__toggle--PaddingBottom: var(--#{$pf-global}--spacer--lg);
-  --#{$wizard}__toggle--PaddingLeft: calc(var(--#{$pf-global}--spacer--md) + var(--#{$wizard}__nav-link--before--Width) + var(--#{$pf-global}--spacer--sm));
-  --#{$wizard}__toggle--m-expanded--BorderBottomWidth: var(--#{$pf-global}--BorderWidth--sm);
-  --#{$wizard}__toggle--m-expanded--BorderBottomColor: var(--#{$pf-global}--BorderColor--100);
-  --#{$wizard}--m-in-page__toggle--xl--PaddingLeft: calc(var(--#{$pf-global}--spacer--xl) + var(--#{$wizard}__nav-link--before--Width) + var(--#{$pf-global}--spacer--sm));
+  --#{$wizard}__toggle--BackgroundColor: var(--pf-t--global--background--color--primary--default);
+  --#{$wizard}__toggle--ZIndex: var(--pf-t--global--Zindex--xs);
```
mcoker, line 96: Toggle has a box shadow that isn't showing because it's behind the wizard nav - increasing the z-index on the toggle gets it to show. It's worth noting that v5 has this same z-index setup and the box shadow on the toggle does show, so I'm wondering if something else has changed the stacking order of toggle + nav on mobile?

mcoker (commented):
* in page wizards - only the content should scroll, header and footer should be sticky (though they don't need to be position sticky necessarily)
* step number background disappears on hover/focus in dark mode
* disabled step text color changes slightly on hover in dark mode

src/patternfly/components/Wizard/wizard.scss
mcoker: Confirmed this is pre-existing - it happens when the drawer is in the main section. It's probably best fixed by assigning a z-index to the main section that's lower than the toggle, which should keep the toggle on top of all children in the main section, regardless their z-index. However, the main section has no z-index on purpose so its children (like menus) that have z-indexes can pop out on top of the footer. I'm not sure how to get around that.

Another option could be to use a border instead of a shadow on the toggle, similar to the update I think we made for the wizard footer. There is already a border in place of the shadow on the toggle when the nav is expanded. wdyt @lboehling? It would look like this:

This is the current shadow:
Screenshot 2024-02-13 at 9 48 45 PM

With a border instead:
Screenshot 2024-02-13 at 9 51 07 PM

And the pre-existing border when the nav is expanded:
Screenshot 2024-02-13 at 9 49 59 PM

PR #6263, opened 2024-01-30
feat(action-list): added icon group variable update

### LINKED ISSUE #6262: Action list - update column-gap var for icon modifier
Action list

**Describe the feature**
We should update the column-gap var for action-list__group when action-list has the `pf-m-icon` modifier, that way we can update all groups in an action list at once rather than having to apply the modifier to each group

**Any other information?**

mattnolting:
closes #6262

### DIFF
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -23,6 +23,8 @@
 
   // - Action list icons
   &.pf-m-icons {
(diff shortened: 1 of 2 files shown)


thatblindgeye (approved):

srambach:
If we have groups inside an action list that is marked as `.pf-m-icons`, the wider space between groups is reduced to the icon spacing. Is that the intent? @lboehling @andrew-ronaldson @mattnolting If so, looks ok, but I don't remember the whole discussion.

image

andrew-ronaldson (approved):
Approved!

Issue #6266, opened 2024-01-30
Bug - Label - editable active has wrong cursor


mcoker:
re: https://github.com/patternfly/patternfly-react/pull/10037#pullrequestreview-1846078689

Has pointer cursor, should be reset to the browser's default for whatever element is in the label

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6267

PR #6267, opened 2024-01-30
fix(label): fixed editable active cursor

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6266

### DIFF
--- a/src/patternfly/components/Label/label.scss
+++ b/src/patternfly/components/Label/label.scss
@@ -166,6 +166,7 @@
   // Clickable
   --#{$label}--m-clickable--hover--BorderWidth: 0;
   --#{$label}--m-clickable--hover--BorderColor: transparent;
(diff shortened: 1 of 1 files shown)


kmcfaul (approved):

src/patternfly/components/Label/label.scss
```diff
@@ -208,6 +209,7 @@
   // Content
   --#{$label}__content--MaxWidth: 100%;
   --#{$label}__content--Gap: var(--pf-t--global--spacer--xs);
+  --#{$label}__content--Cursor: auto;
```
thatblindgeye, line 212: Noticing an issue caused by this for the "add label", since the pf-v5-c-label__content class is on a div in that case instead of the button element:


https://github.com/patternfly/patternfly/assets/70952936/22dcb0b3-23ac-4b39-998c-2ba12e9a7f60

thatblindgeye (approved):

Issue #6270, opened 2024-01-31
Bug - Divider - disappears in a flex layout


mcoker:
re: https://github.com/patternfly/patternfly-react/issues/9919#issuecomment-1892721578

From glancing at it, looks like we need to add some flex properties as defaults (likely to the base divider for horiz/vertical) to keep it visible, like `flex-shrink: 0`

wise-king-sullyman:
Closed by #6271

PR #6271, opened 2024-01-31
fix(divider): support use in flex layout

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6270


mattnolting (approved):
Bravo 👏👏👏 LPTM!

PR #6281, opened 2024-02-05
chore(tokens): update tokens and add chart tokens

srambach:
This updates the tokens last updated in mid-December.
It adds the token file for charts, but does _not_ pull it in anywhere yet.
It updates stylelint to allow for the token pattern to include `--pf-t--chart--`


mcoker (approved):
LGTM tho looks like there are a bunch of var name changes that need to be updated in the components - want to do that in a separate issue?

srambach:
Redoing this export.

Issue #5723, opened 2023-07-05
Consume tokens: Tree view


srambach:
[Figma designs for tree view](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A50&mode=design&t=boxqVg1hn5hO0weV-1)

wise-king-sullyman:
Closed by #6284

PR #6284, opened 2024-02-08
fix(tree-view): add tokens

mcoker:
**[convenience link to tree view component](https://patternfly-pr-6284.surge.sh/components/tree-view)**

fixes https://github.com/patternfly/patternfly/issues/5723

* Just CSS updates
* Left a to-do about maybe using the button component for the toggle icon, but maybe not - it isn't actually a `<button>` element except for in a checkbox tree-view, otherwise it's just an icon inside of the tree view node, and the node is a `<button>`.
* Not sure about the spacing between the action and the right edge of the node here - wdyt @lboehling @andrew-ronaldson? It's a similar layout to the menu component, but the spacing is different in figma.

Screenshot 2024-02-08 at 10 00 52 PM


srambach (approved):
I don't see any problems, good questions posed in the description. 🌴

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -1,77 +1,70 @@
...
-  --#{$tree-view}__node-toggle-icon--MinWidth: var(--#{$pf-global}--FontSize--md);
-  --#{$tree-view}__node-toggle-icon--Transition: transform var(--#{$pf-global}--TransitionDuration) var(--#{$pf-global}--TimingFunction);
-  --#{$tree-view}__node-toggle-button--PaddingTop: var(--#{$pf-global}--spacer--form-element);
-  --#{$tree-view}__node-toggle-button--PaddingRight: var(--#{$pf-global}--spacer--md);
-  --#{$tree-view}__node-toggle-button--PaddingBottom: var(--#{$pf-global}--spacer--form-element);
-  --#{$tree-view}__node-toggle-button--PaddingLeft: var(--#{$pf-global}--spacer--md);
-  --#{$tree-view}__node-toggle-button--MarginTop: calc(var(--#{$pf-global}--spacer--form-element) * -1);
-  --#{$tree-view}__node-toggle-button--MarginBottom: calc(var(--#{$pf-global}--spacer--form-element) * -1);
+  --#{$tree-view}__node-toggle-icon--MinWidth: var(--pf-t--global--icon--size--body--default);
+  --#{$tree-view}__node-toggle-icon--MinWidth: var(--pf-t--global--icon--size--body--default);
```
wise-king-sullyman, line 45: Duplicate var?

wise-king-sullyman (changes requested):

Issue #5728, opened 2023-07-05
Consume tokens: Table


srambach:
[Figma design for (datalist and) table](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/Penta-Component-Library?type=design&node-id=3%3A12&mode=design&t=doZDRUaxE4vsn5dD-1)

mcoker:
GTG on table for alpha 🚀

PR #6294, opened 2024-02-08
Chore table padding

mattnolting:
relates to #5728


src/patternfly/components/Table/table.scss
```diff
@@ -808,14 +786,15 @@
...
-  padding-block-start: var(--#{$table}--cell--PaddingTop);
-  padding-block-end: var(--#{$table}--cell--PaddingBottom);
-  padding-inline-start: var(--#{$table}--cell--PaddingLeft);
-  padding-inline-end: var(--#{$table}--cell--PaddingRight);
+  width: auto;
+  padding-block-start: var(--#{$table}__button--PaddingTop);
+  padding-block-end: var(--#{$table}__button--PaddingBottom);
+  padding-inline-start: var(--#{$table}__button--PaddingLeft);
+  padding-inline-end: var(--#{$table}__button--PaddingRight);
+  margin-inline-start: calc(var(--#{$table}__button--PaddingLeft) * -1);
```
mcoker, line 797: Bold strategy, but let's see how it plays out! ☄️

src/patternfly/components/Table/table.scss
```diff
@@ -954,39 +924,35 @@
...
-    margin-inline-start: calc(var(--#{$table}--cell--PaddingLeft) * -1);
-    margin-inline-end: calc(var(--#{$table}--cell--PaddingRight) * -1);
 
+    thead & {
+      vertical-align: bottom;
+    }
+
+    tbody & {
+      vertical-align: top;
+    }
```
mcoker, line 939: just a thought, but you could probably add `vertical-align: top` as the base style [under the display declaration](https://github.com/patternfly/patternfly/blob/a01f949cbef42c7bc6a7cf7e9b3fc64099189fc5/src/patternfly/components/Table/table.scss#L931) and just change that in `thead`?

src/patternfly/components/Table/table.scss
```diff
@@ -954,39 +924,35 @@
...
   }
 
-  @at-root .#{$table}__thead .#{$check}.pf-m-standalone {
-    --#{$check}__input--TranslateY: var(--#{$table}__thead__check--input--TranslateY);
+  // TODO: remove label wrapper at breaking change
+  label {
+    display: contents;
```
mcoker, line 944: I wonder if this will cause an a11y issue since last I read, display: contents still isn't a great thing for elements with semantic value specifically around remaining in the a11y tree properly.

mcoker (approved):
LGTM! Left some comments as I was looking at the code but nothing that needs action, feel free to address later if you think any of the comments are relevant.

Issue #6299, opened 2024-02-09
Nav - horizontal updates


mcoker:
* Scroll button disabled styles are not working - looks like the plain buttons in that example wrap the icon in the button__icon element, which isn't needed and is making the icon the wrong color.
* Horizontal subnav page section needs updates
* Horizontal subnav demo doesn't have horizontal subnav? https://pf6.patternfly.org/components/navigation/html-demos/horizontal-subnav/

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6302

PR #6302, opened 2024-02-10
fix(nav): updates to horizontal subnav

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6299

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -117,7 +117,7 @@ $pf-page-v5--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   // Main section
   --#{$page}__main-section--MarginTop: var(--pf-t--global--spacer--md);
   --#{$page}__main-section--PaddingTop: var(--pf-t--global--spacer--lg);
-  --#{$page}__main-section--PaddingRight: calc(var(--pf-t--global--spacer--lg) - var(--#{$page}__main--BorderWidth)); // subtract the border of the main section
+  --#{$page}__main-section--PaddingRight: calc(var(--pf-t--global--spacer--lg) - var(--#{$page}__main--BorderWidth)); // subtract the border of the main section - TODO, see if we can remove the need for this calc(), or possibly move it down to the property definition so a user can theme this without a calc()
   --#{$page}__main-section--PaddingBottom: 0;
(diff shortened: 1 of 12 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -151,14 +151,11 @@ $pf-page-v5--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
   // Main section horizontal subnav
-  --#{$page}__main-subnav--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
-  --#{$page}__main-subnav--BorderTopWidth: var(--pf-t--global--border--width--divider--default);
-  --#{$page}__main-subnav--BorderTopColor: var(--pf-t--global--border--color--default);
-  --#{$page}__main-subnav--BorderLeftWidth: 0;
-  --#{$page}__main-subnav--BorderLeftColor: var(--pf-t--global--border--color--default);
-  --#{$page}__sidebar--main__main-subnav--BorderLeftWidth: var(--pf-t--global--border--width--divider--default);
-  --#{$page}__sidebar--m-collapsed--main__main-subnav--BorderLeftWidth: 0;
-
+  --#{$page}__main-subnav--BackgroundColor: var(--pf-t--global--background--color--primary--default);;
```
srambach, line 154: semisemi :)

tlabaj:
Should the button have the `pf-m-disabled` modifier?  It does in react hen `isDisabled` is applied.

image

mcoker:
@tlabaj only when the button isn't a `<button>` element. For `<button>` elements, we can just style via the `disabled` attribute (`:disabled { ... }`). But when the element is an `<a>` (https://www.patternfly.org/components/button#links-as-buttons) or `<span>` (https://www.patternfly.org/components/button#inline-link-as-span), the `disabled` attribute isn't valid, so we style it to look disabled via the class `pf-m-disabled`

srambach:
One other comment - the focus rings on the scroll buttons are a bit cut off and weird, but that's definitely fine for a follow up issue.

mcoker:
Screenshots of a nav demo with horizontal subnav with scrollbars

Screenshot 2024-03-05 at 6 26 53 PM
Screenshot 2024-03-05 at 6 26 43 PM

srambach (approved):
Looks fabulous! Love the focus 💍

Issue #6250, opened 2024-01-24
Bug - Accordion, Button, Horizontal Nav Overflow - Mix blend modes dropped


lboehling:
@thatblindgeye noticed in [this issue](https://github.com/patternfly/patternfly-react/pull/10033#pullrequestreview-1838876252) that the hover states on accordions weren't appearing as expected in dark mode. Manually applying the mix blend mode made the hover states appear normally. 

Looks like the mix blend mode tokens have fallen off a number of components that previously had them defined in the [v6 core staging site](https://pf6.patternfly.org/components/button). Noticing it missing on plain and link buttons, accordion toggles, and horizontal overflow buttons.

srambach:
@lboehling @mcoker can this be closed now that we are using the opacity in the token rather than a separate mix blend mode?

PR #6306, opened 2024-02-13
fix(accordion): address item background not showing

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6250


mcoker:
Closing for now, going to try another method for this kind of background.

PR #6310, opened 2024-02-13
feat(CodeEditor): add border-start-end-radius if header-content is the last child

### LINKED ISSUE #6309: CodeEditor - v6 - header-content needs border-start-end-radius
Because in react the language label is optional, the header-content also needs the right side border radius to maintain consistency.

kmcfaul:
Closes #6309


mcoker (approved):
🚀

PR #6321, opened 2024-02-14
feat(menu-toggle): remove extra div, reorder examples

### LINKED ISSUE #6318: Menu toggle follow up for alpha
* Looks like a feature was added where the menu toggle is just a div and it nests a button component inside that we should remove for alpha. If we want to support that, we should talk about the best way to go about it, but not include it for alpha. There are plain menu toggle examples that show that feature - https://pf6.patternfly.org/components/menus/menu-toggle/#plain
* Plain disabled menu toggle should be updated to look like the plain disabled button

srambach:
Fixes #6318 

Removes the extra div around some examples.
Reorders within example sets so that the toggles are always collapsed, expanded, disabled
Styles disabled plain version to match the disabled plain button.

### DIFF
--- a/src/patternfly/components/MenuToggle/examples/MenuToggle.md
+++ b/src/patternfly/components/MenuToggle/examples/MenuToggle.md
@@ -160,19 +160,13 @@ import './MenuToggle.css'
 
 ### Plain
 ```hbs
-{{> menu-toggle menu-toggle--UsesButton=true menu-toggle--IsPlain=true menu-toggle--IsDiv=true}}
-&nbsp;
 {{#> menu-toggle menu-toggle--IsPlain=true menu-toggle--attribute='aria-label="Actions"'}}
   <i class="fas fa-ellipsis-v" aria-hidden="true"></i>
 {{/menu-toggle}}
 &nbsp;
-{{> menu-toggle menu-toggle--UsesButton=true menu-toggle--IsPlain=true menu-toggle--IsDiv=true menu-toggle--IsExpanded=true}}
-&nbsp;
 {{#> menu-toggle menu-toggle--IsPlain=true menu-toggle--IsExpanded=true menu-toggle--attribute='aria-label="Actions"'}}
   <i class="fas fa-ellipsis-v" aria-hidden="true"></i>
 {{/menu-toggle}}
-&nbsp;
-{{> menu-toggle menu-toggle--UsesButton=true menu-toggle--IsPlain=true menu-toggle--IsDiv=true menu-toggle--IsDisabled=true}}
 &nbsp;
 {{#> menu-toggle menu-toggle--IsPlain="true" menu-toggle--IsDisabled="true" menu-toggle--attribute='aria-label="Actions"'}}
   <i class="fas fa-ellipsis-v" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


src/patternfly/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -161,15 +161,15 @@ import './MenuToggle.css'
 ### Plain
 ```hbs
 {{#> menu-toggle menu-toggle--IsPlain=true menu-toggle--attribute='aria-label="Actions"'}}
```
mcoker, line 163: I haven't tried it, but the hbs looks like it might be wired up so you can write it this way?

```suggestion
{{> menu-toggle menu-toggle--IsPlain=true menu-toggle--attribute='aria-label="Actions"' menu-toggle--icon="ellipsis-v"}}
```

mcoker (approved):
LGTM! Just one nit. Also we'll need to poke around for plain menu toggles and make sure they have the icon wrapper. If we use the suggested code in my nit comment, that will let us include/exclude the icon wrapper behind the scenes without needing to go through the code and include/exclude the nested `{{> menu-toggle-icon ...}}` child that's in the PR now.

mcoker:
@thatblindgeye @tlabaj this PR updates the plain menu toggle to include the the icon wrapper `pf-v5-c-menu-toggle__icon`. That lets us assign icon styling to that class that's consistent between plain and non-plain menu toggles, compared to what we do now which is style that class in non-plain menu toggles and style the whole menu toggle with icon styles for plain toggles. One issue with the way we do it now is that not all plain toggles contain just an icon - the breadcumb dropdown, for example, is a [plain toggle with a badge](https://github.com/patternfly/patternfly-react/blob/08a6b1aaaf6769b657ed674f4bb741b530265c7d/packages/react-core/src/components/Breadcrumb/examples/BreadcrumbDropdown.tsx#L69-L76).

With this update, looking at the update we'll need to make to the menu toggle component in react, I'm guessing that might mean that an icon is always passed as the `icon` prop, instead of passed as children for plain toggles? Do you see any issues with that change? FWIW, I'd like to make the same update to button - we'll add a wrapper for button text, and text will always go in the text container, and an icon will always go in the icon container, for all button variants.

src/patternfly/components/MenuToggle/examples/MenuToggle.md
srambach: @mcoker the additional changes you made for this in the examples/demos all look good.

thatblindgeye:
Just wanted to show a screenshot from React after adding the icon wrapper around the plain toggle's svg icon and updating the styles to remove the margin and line-height from that wrapper:

Plain Menu toggle with icon wrapper around icon

Just to check that it looks as expected in React.

As for any necessary updates in React, it may depend how much we want to restrict passing anything as children to MenuToggle. Right now a user could pass whatever: "Toggle text", `<div class="fancy">ToggleText</div>`, `<SomeIcon />`, etc.

So we could update prop descriptions and examples to convey that we recommend passing *all* icons to the `icon` prop and only text content (whether it be wrapped in some other element or not) as `children`, but that wouldn't really stop consumers from still passing icons as children. Which maybe that's on the consumer to update to get the intended styling; we can provide a codemod to at least warn consumers about this sort of update so they're aware.

Alternatively we could restrict the type that can be passed to children to just a string or creating a new "text" prop to replace children, neither of which sound all to intuitive or great.

Here's what it may look like passing the icon to the `icon` prop isntead of children; note that I had to manually delete the `__controls` content and place the `pf-m-plain` class in dev tools due to how we're rendering the MenuToggle (doing everything in the example code results in nothing being rendered for a plain toggle since its expecting children to render).

Plain menu toggle with icon passed to icon prop

Depending where we land on this, I can use the open React issue for Penta MenuToggle updates to add this wrapper and make any other necessary updates if we want.

Issue #6319, opened 2024-02-14
Menu - follow ups for alpha


mcoker:
from https://github.com/patternfly/patternfly/pull/6167#pullrequestreview-1866449856

* [Menu item focus isn't working](https://github.com/patternfly/patternfly/pull/6167#discussion_r1489741757)
* [Menu item hover/focus color is different between menus](https://github.com/patternfly/patternfly/pull/6167#discussion_r1489815352)
* [Take all but `__list`, `__item`, and `__group` out of the grouped hidden-visible blocks](https://github.com/patternfly/patternfly/pull/6167#discussion_r1489777617)
* [Fix for/id mismatch in the checkbox menu examples and remove the input pseudo element styling](https://github.com/patternfly/patternfly/pull/6167#discussion_r1489751339)
* [Remove `:active` styles](https://github.com/patternfly/patternfly/pull/6167#discussion_r1489745076)

PR #6322, opened 2024-02-15
fix(menu): updates for alpha

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6319

I didn't fix this issue with the [item hover/focus color being different between menus](https://github.com/patternfly/patternfly/pull/6167#discussion_r1489815352) - I believe it has to do with the blend mode and we're going to try an alternative to the blend mode for alpha, which will hopefully resolve whatever's going on there

Also left and fixed the for/id attributes (instead of just removing them) on the labels with nested checkboxes because that matches the react component.


src/patternfly/components/Menu/menu.scss
```diff
@@ -472,9 +471,10 @@
     padding-inline-end: var(--#{$menu}__list-item--has--menu-action--PaddingInlineEnd);
   }
 
+  &:focus-within,
```
kmcfaul, line 474: I think we need `pf-m-focus` back for typeahead select to manually show visual focus on an item, because actual focus remains on the input.

kmcfaul (approved):

src/patternfly/components/Menu/menu.scss
```diff
@@ -560,13 +567,6 @@
 // - Menu check
 .#{$menu}__item-check .#{$check} {
   --pf-v5-c-check__input--TranslateY: none;
```
mattnolting, line 569:
```suggestion
  --pf-v6-c-check__input--TranslateY: none;
```

mattnolting (changes requested):
Nice work, several version updates are needed for examples.

mattnolting (approved):

Issue #6305, opened 2024-02-13
Make sure all the tokens we're using are defined


mcoker:
We may be using renamed/outdated tokens in our components, so we should check for that and update where needed. One way could be to get a list of tokens used in all components and search those in our tokens stylesheets. Eg,

```
for i in `grep -riIho "\-\-pf-t--[a-z0-9_-]*" src/patternfly/components/ | sort | uniq`; do
  if grep -riq -- $i src/patternfly/base/tokens/; then
    echo "FOUND: $i";
  else
    echo "NOT FOUND: $i";
  fi;
done
```

----

If it's useful, there is also a script available from a PR that renamed some tokens - https://github.com/patternfly/patternfly/pull/6282#issuecomment-1934944287

```
// manually created this file at ~/replace (eg, /Users/mcoker/replace)
pf-t--global--border--width--control--active|pf-t--global--border--width--control--clicked
pf-t--global--border--width--button--active|pf-t--global--border--width--button--clicked
pf-t--global--border--width--divider--active|pf-t--global--border--width--divider--clicked
pf-t--global--color--status--unread--default--active|pf-t--global--color--status--unread--default--clicked
pf-t--global--icon--size--body--lg|pf-t--global--icon--size--font--body--lg
pf-t--global--icon--size--body--default|pf-t--global--icon--size--font--body--default
pf-t--global--icon--size--body--sm|pf-t--global--icon--size--font--body--sm
pf-t--global--icon--size--heading--h6|pf-t--global--icon--size--font--heading--h6
pf-t--global--icon--size--heading--h5|pf-t--global--icon--size--font--heading--h5
pf-t--global--icon--size--heading--h4|pf-t--global--icon--size--font--heading--h4
pf-t--global--icon--size--heading--h3|pf-t--global--icon--size--font--heading--h3
pf-t--global--icon--size--heading--h2|pf-t--global--icon--size--font--heading--h2
pf-t--global--icon--size--heading--h1|pf-t--global--icon--size--font--heading--h1
```

```bash
for i in `cat ~/replace`; do
  OLD=`echo $i | cut -d'|' -f1`; # stores the old token name in $OLD
  NEW=`echo $i | cut -d'|' -f2`; # stores the new token name in $NEW
[...]

PR #6328, opened 2024-02-17
fix(vars): fix invalid token names

mcoker:
Part of https://github.com/patternfly/patternfly/issues/6305, but I think we should leave that issue open and do this again once core code changes are complete.


srambach (approved):
Changes look good

src/patternfly/components/Login/__login-main-footer.hbs
```diff
@@ -1,29 +1,29 @@
 {{#> login-main-footer}}
   {{#> login-main-footer-links}}
     {{#> login-main-footer-links-item}}
-      {{#> login-main-footer-links-item-link login-main-footer-links-item-link--attribute='aria-label="Log in with Google"'}}
-        <svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 488 512"><path d="M488 261.8C488 403.3 391.1 504 248 504 110.8 504 0 393.2 0 256S110.8 8 248 8c66.8 0 123 24.5 166.3 64.9l-67.5 64.9C258.5 52.6 94.3 116.6 94.3 256c0 86.5 69.1 156.6 153.7 156.6 98.2 0 135-70.4 140.8-106.9H248v-85.3h236.1c2.3 12.7 3.9 24.9 3.9 41.4z"/></svg>
-      {{/login-main-footer-links-item-link}}
+      {{#> button button--IsPlain=true button--aria-label="Log in with Google"}}
```
thatblindgeye, line 4: Could be a followup, but with this update the icon color for the "log in with..." buttons no longer works in dark theme:

Log in buttons with dark text in dark theme

Looks like it's because we're no longer targeting the `fill` property for the icon color here after removing the `login-main-footer-links-item-link` component + styling.

thatblindgeye (commented):
Looks good. Had a comment below which could be a followup, but depends whether we'd want to add anything back in that was removed in this PR

src/patternfly/components/Login/__login-main-footer.hbs
mcoker: nice! I meant to add `fill="currentColor"` to those SVGs. updated

https://github.com/patternfly/patternfly/pull/6328/commits/ab289581b94910efeeddcf72c25e7cd4b7165d6f

https://patternfly-pr-6328.surge.sh/components/login-page/html/basic/

thatblindgeye (approved):

PR #6330, opened 2024-02-18
fix(check): update standalone styling

mattnolting:
Applies to #6294 

This PR fixes standalone padding vertical alignment. The standalone variant lacks consideration in context. This code takes a line height and font size to calculate a height, then uses default alignment of center to position vertically, but will respond to alternative alignments.


thatblindgeye (commented):
The spacing looks off in tables with checkboxes with this update:

image

Is this intended? Otherwise lgtm

kmcfaul (approved):

mcoker (approved):

PR #6335, opened 2024-02-20
feat(tokens): update tokens, add opacity, box-shadow tokens

srambach:
Tokens updated
- Box shadow tokens from figma added
- Box shadows built from constituent tokens
- Clicked changed to use opacity rather than blend mode
- Some additional color changes

Components:
- Mix blend mode removed
- Accordion `--selected` token changed to `--clicked`
- Fixed some missing semicolons in button

### DIFF
--- a/src/patternfly/base/tokens/_tokens-charts.scss
+++ b/src/patternfly/base/tokens/_tokens-charts.scss
@@ -1,6 +1,6 @@
 // /**
 //  * Do not edit directly
-//  * Generated on Tue, 06 Feb 2024 18:33:23 GMT
+//  * Generated on Tue, 20 Feb 2024 01:08:05 GMT
 //  */
 
 :root {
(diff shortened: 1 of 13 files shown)


mcoker (approved):
LGTM! Tested all of the components that use the `background--color--plain` tokens and they look fantastic. All of the shadow tokens are rendering as expected.

Though I did notice that we still have these dark theme overrides that look like they can be removed. They are what you see if you switch to dark theme, but if you uncheck any of them in dev tools, the shadow seen now comes from the dark theme token changes to these new tokens @lboehling added and it looks identical to these overrides - https://github.com/patternfly/patternfly/blob/9c5b73edff22f5898ace6fd52523e0d69df2e5aa/src/patternfly/base/tokens/_tokens-font.scss#L149-L158

@srambach if you don't mind confirming, and if it's ok to remove them, we can do that as a follow up.

PR #6375, opened 2024-03-01
chore(table): remove shadows

mattnolting:
references #5728


srambach (approved):
👍  this does remove all box shadows. I'd guess you'll want some of the removed selectors back for the hover states' background colors but if so they're always in history I guess.

mcoker (approved):
🚀

Issue #5751, opened 2023-07-11
Bug - FormSelect - Applying "pf-m-placeholder" causes all root level options to grey out on chrome


phill-85:
**Describe the problem**
Setting `pf-m-placeholder` on the wrapping span of a `FormSelect component` causes all the options not in optgroups to grey out on chrome.

The first image is chrome (note option 5 & 6 are grayed out), the second one is firefox.
![image](https://github.com/patternfly/patternfly/assets/4619659/92a7d031-9ab0-4144-ba90-a59d89cf69e2)
![image](https://github.com/patternfly/patternfly/assets/4619659/7c5a2008-697a-46ec-8aa9-70a07c6f9bc3)

**What is your environment?**
 - OS: Ubuntu 23.04
 - Browser 114.0.5735.198 (Official Build) (64-bit) & Brave browser

mcoker:
Hey @phill-85! Thanks for reporting this issue. I'm unable to reproduce this using chome and brave. Can you share a codepen or codesandbox that replicates the bug you're seeing? Or link to an example on one of our documentation sites if you're seeing it there, too?

FWIW, @srambach has a PR up that changes some code that looks like it could be related, I'm curious what you're seeing in that PR preview. Here's the PR - https://github.com/patternfly/patternfly/pull/5752, and here is a link to the form select from that PR - https://patternfly-pr-5752.surge.sh/components/forms/form-control#select

phill-85:
> FWIW, @srambach has a PR up that changes some code that looks like it could be related, I'm curious what you're seeing in that PR preview. Here's the PR - #5752, and here is a link to the form select from that PR - https://patternfly-pr-5752.surge.sh/components/forms/form-control#select

On that link you supplied ^ removing the `pf-m-placeholder` from the span wrapping the select with the `id="select-selectable-placeholder"` causes all the options to change from grey to black. See: https://drive.google.com/file/d/1u1qYtju0EpR4voJz16HHD1MIRlh9kti1/view?usp=sharing for a video.

It doesn't appear to affect child `option's` of an `optgroup`

phill-85:
I've had a play and I don't know if this is fixable...

i changed `--pf-v5-c-form-control--m-placeholder--Color` to red. The style inspector shows the `.pf-v5-c-form-control.pf-m-placeholder > select` rule as overridden on the option tag. 

Even adding `!important` didn't make a difference.

See that attached screenshot.

[...]

Issue #5695, opened 2023-06-28
Bug - [Select/FormSelect] - [Expanded menu is in white theme not dark (regression)]


jelly:
**Describe the problem**

We noticed this when upgrading from 53 to 98 alpha release and can reproduce it on the PF5 [demo website](https://pf5.patternfly.org/components/forms/form-control)

![image](https://github.com/patternfly/patternfly/assets/67428/3eb24240-b0eb-47a2-a9a8-8df2029b5373)

**How do you reproduce the problem?**

1. Go to https://pf5.patternfly.org/components/forms/form-control
2. Enable dark theme
3. Open any of the dropdown menu's.

**Expected behavior**

Dropdown menu should be dark

**Is this issue blocking you?**
List the workaround if there is one.

**Screenshots**
If applicable, add screenshots to help explain the issue.

**What is your product and what release date are you targeting?**

Cockpit

jelly:
This actually seems to be mostly a FormSelect issue or well, where `select` is used, previously the select had a `class="pf-v5-c-form-control"` and with the new version it does not as it's now a child of a div with that class.

jelly:
This was previously overridden by patternfly, but as the select now lacks the `pf-v5-c-form-control` class the css rules no longer apply.

jelly:
Re-testing with https://patternfly-react-v5.surge.sh/components/forms/form-select makes it seem that this is Chrome(ium) only.

![image](https://github.com/patternfly/patternfly/assets/67428/db5ec41f-2eba-49e9-b908-9b2720ade8a8)

![image](https://github.com/patternfly/patternfly/assets/67428/8d92c582-916b-4860-8e7c-f08cdd8bafe6)

mcoker:
@jelly I don't have chromium installed, but checking in chrome and chrome canary, I'm not able to reproduce what you have in the screenshot. The native `<select>` menu is taking on whatever the OS preference is for dark/light theme, regardless the setting for dark/light mode in patternfly. I confirmed this is the case in our latest pre-release version of PF as well as the version prior to the update that wrapped the `<select>` in a div. Is this only happening on Chromium?

jelly:
I just spun up a Windows VM with Edge and it shows as this. So I didn't set the OS preference to dark theme but only toggled the radio button in the PatternFly website.


![image](https://github.com/patternfly/patternfly/assets/67428/2a283494-61cf-49ec-8e03-e97adac4696d)

garrett:
[...]

PR #5531, opened 2023-05-04
feat(form-control): move background icons to element

srambach:
Fixes #5469 #4900 #2702

Note: examples using form control still need to be updated

This changes from using a background image on a form control to using a wrapper around the form control with associated utility container for icon, status icons, and toggle icon.

note: There is some existing code for placeholder text color that doesn't seem to quite work right and the existing feature is not quite as I'd expect so I'd like to discuss @mcoker

PR #6397, opened 2024-03-05
fix(FormControl): updated select styling in light and dark themes

thatblindgeye:
Towards https://github.com/patternfly/patternfly/issues/5695 and #5751 

Need to test this update in other OS/browser combos to ensure nothing visually breaks.

Would most likely need a v6 update as well if we go forward with this.

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -93,7 +93,7 @@
   --#{$form-control}--textarea--Width: var(--#{$form-control}--Width);
   --#{$form-control}--textarea--Height: auto;
 
-  // Form control icon 
+  // Form control icon
   --#{$form-control}__icon--PaddingTop: var(--#{$pf-global}--spacer--form-element);
   --#{$form-control}__icon--Color: var(--#{$pf-global}--icon--Color--light);
   --#{$form-control}__icon--m-status--Color: var(--#{$pf-global}--icon--Color--light);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/FormControl/form-control.scss
```diff
@@ -273,13 +273,18 @@
       --#{$form-control}--PaddingRight: calc(var(--#{$form-control}__select--PaddingRight) - 1px);
       --#{$form-control}--PaddingLeft: calc(var(--#{$form-control}__select--PaddingLeft) - 4px);
     }
+
+    // We need this property for certain Linux distros/non-Mac OS where dark theme styles aren'y applied properly.
```
thatblindgeye, line 277:
```suggestion
    // We need this property for certain Linux distros/non-Mac OS where dark theme styles aren't applied properly.
```

mcoker:
For reference, we dug up an old issue that reported this same bug a couple of years ago and was fixed, and that fix was later removed.

https://github.com/patternfly/patternfly/issues/4772
https://github.com/patternfly/patternfly/pull/4773/files

Looks like it probably [broke in this PR](https://github.com/patternfly/patternfly/pull/5531) where we changed  `color: var(--#{$form-control}--placeholder--child--Color);` to `--#{$form-control}--Color: var(--#{$form-control}--m-placeholder--child--Color);`

mcoker (approved):
Tested Chrome, FF, Safari, and Edge on my mac - the backgrounds and item colors all look good (unchanged)

thatblindgeye:
Checked Windows + Chrome/FireFox/Edge and from what I can tell nothing breaks either. Can confirm that the bug from #5751 does affect Windows + Chrome as well.

srambach (approved):
LGTM thank you!

Issue #6273, opened 2024-01-31
Utility classes - update to support RTL


mcoker:
There are some utility classes that could be updated with logical values. We should look over all of them, but two that stand out are

https://github.com/patternfly/patternfly/blob/7bff19089077f5f61631735b5bc7b190257d90d2/src/patternfly/utilities/Alignment/alignment.scss#L2-L7

https://github.com/patternfly/patternfly/blob/7bff19089077f5f61631735b5bc7b190257d90d2/src/patternfly/utilities/Float/float.scss#L3-L6

srambach:
Related issue: https://github.com/patternfly/patternfly/issues/6315

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6400

PR #6400, opened 2024-03-06
fix(utilities): rtl, penta updates

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6273 and drops in a few other easy penta changes


srambach (approved):
Looks good 👍  Thanks for doing this one.

PR #6410, opened 2024-03-11
fix(page): add container around main element

### LINKED ISSUE #6225: Page - main content rounded borders not always visible
...

srambach:
This adds a container around the main element so that the border-rounding and other styling can be applied to the container, while the main element handles layout of the sections.

Fixes #6225

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -11,7 +11,8 @@ $pf-v5-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 
   // Content
   --#{$drawer}__content--FlexBasis: 100%;
-  --#{$drawer}__content--BackgroundColor: var(--pf-t--global--background--color--primary--default);
+  --#{$drawer}__content--BackgroundColor: transparent;
+  --#{$drawer}__content--m-primary--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$drawer}__content--m-secondary--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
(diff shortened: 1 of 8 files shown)


mcoker (commented):
Just one nit

mcoker (approved):

mcoker (changes requested):
One more bit - now that `{{#> page-main}}` includes this globally, we can remove the container from this template https://github.com/patternfly/patternfly/blob/6ffe50d3b8697bd7cda6f05599d39ecff81567b7/src/patternfly/demos/Page/page-template-main.hbs#L1-L5

Also just to confirm, with a drawer in the page (like a [notification drawer](https://patternfly-pr-6410.surge.sh/components/drawer/html-demos/expanded/)) the grid area for the main page stuff is `.pf-c-page__drawer` which is not currently wrapped in the main-container - should it be, or is it good the way it is?

srambach:
> One more bit - now that `{{#> page-main}}` includes this globally, we can remove the container from this template
> 
if one is good, two are better, no? 🤪 
> https://github.com/patternfly/patternfly/blob/6ffe50d3b8697bd7cda6f05599d39ecff81567b7/src/patternfly/demos/Page/page-template-main.hbs#L1-L5
> 
> Also just to confirm, with a drawer in the page (like a [notification drawer](https://patternfly-pr-6410.surge.sh/components/drawer/html-demos/expanded/)) the grid area for the main page stuff is `.pf-c-page__drawer` which is not currently wrapped in the main-container - should it be, or is it good the way it is?
Hmmmmm. That would fix a problem where a drawer content with a background escapes the rounded corners, but also cause the drawer panel to be contained inside the rounded corners, which is not the intention I think.

mcoker (approved):

PR #6413, opened 2024-03-12
chore(table): updated all thead > td to th

### LINKED ISSUE #6272: Bug - Table - accessibility rules have changed, all thead cells should be `<th>`s
**Describe the problem**
Accessibility rules have changed, `<td>`s in `thead`s are no longer acceptable. 

Test with lighthouse, other accessibility tools, and/or screen reader

**Expected behavior**
Columns are described clearly and concisely with text and without altering table layout

- [x] Replace all `thead > td`s (blank visually) to `<th><span class="pf-v5-screen-reader">{descriptive text}{/span></th>`
- [x] For non-text content (select all checkbox, expand all buttons), `<th aria-label="Row check">{interactive element such as input or button with its own aria-label}</th>`
- [x] Update `:empty` table cell styling to be something like 
```
:is(.pf-v5-c-table__th, .pf-v5-c-table__td) {
  &:has(.pf-v5-screen-reader),
  &:empty {
    width: auto
    min-width: auto;
    padding: 0;
  }
}
```

mattnolting:
closes #6272 
addresses [#10057](https://github.com/patternfly/patternfly-react/issues/10057)

### DIFF
--- a/src/patternfly/components/Table/TableCells/table-cell-check.hbs
+++ b/src/patternfly/components/Table/TableCells/table-cell-check.hbs
@@ -2,10 +2,10 @@
 {{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
 {{!-- TODO: automate attribution --}}
 
-{{#> table-td table-td--IsCheckbox=true}}
+{{#> table-cell table-cell--IsCheckbox=true}}
   {{> check
       check--IsStandalone=true
       check--IsLabelWrapped=true
       check--id=(ternary table-cell-check--id table-cell-check--id (dasherize table--id 'checkrow' table-tr--index 'check'))
     }}
-{{/table-td}}
+{{/table-cell}}
--- a/src/patternfly/components/Table/TableCells/table-cell-toggle.hbs
+++ b/src/patternfly/components/Table/TableCells/table-cell-toggle.hbs
@@ -2,7 +2,8 @@
 {{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
 {{!-- TODO: automate attribution --}}
 
-{{#> table-td table-td--IsToggle=true}}
+{{!-- {{#> table-td table-td--IsToggle=true}} --}}
+{{#> table-cell table-cell--IsToggle=true}}
   {{#> button
     button--IsSmall=table--IsCompact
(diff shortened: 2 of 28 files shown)


src/patternfly/components/Table/TableCells/table-cell-toggle.hbs
```diff
@@ -2,7 +2,8 @@
 {{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
 {{!-- TODO: automate attribution --}}
 
-{{#> table-td table-td--IsToggle=true}}
+{{!-- {{#> table-td table-td--IsToggle=true}} --}}
```
thatblindgeye, line 5: This and line 21 comments were left in

src/patternfly/components/Table/TableCells/table-cell-check.hbs
```diff
@@ -2,10 +2,10 @@
 {{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
 {{!-- TODO: automate attribution --}}
 
-{{#> table-td table-td--IsCheckbox=true}}
+{{#> table-cell table-cell--IsCheckbox=true}}
```
thatblindgeye, line 5: Similar thing for the table-cell-toggle file, but the `th` element will need an aria-label of its own when the only content is an interactive, non-text content. "Row select" in this case and "Row expansion" for the toggle, or something like that (I think I used those in my React PR)

thatblindgeye (commented):
Looks like some examples are broken with this update (tree tables for example).

Also noticing some instances of td still in thead, where the `td` elements in the column contain links: "Compound expansion example", "Compact example", "Compact expandable example", "Clickable and expandable example", "Borderless with compound expansion example", "Nested column headers and expandable rows" (this last one is also missing the empty class)

mattnolting:
There's too much going on in this PR. Will be opening a new, simpler PR in favor. The `.hbs` templates do need to be refactored/deleted where applicable, but that can be done post v6 release.

thatblindgeye:
Agreed. I think the big thing was getting something in React done since there was a somewhat larger issue of element transformation happening unexpectedly. As long as we keep the Core followup for the effort on the radar post v6 it's not totally necessary to get this into Core urgently

Issue #6390, opened 2024-03-05
Bug - [Masthead] - Hamburger disappears in Light Variant


ChrisLee5000:
**Describe the problem**
Masthead-> Light and Light 200 Variant.
The hamburger is white on white (invisible) when hovering, and in default mode is of a very low contrast.

**How do you reproduce the problem?**
https://www.patternfly.org/components/masthead/html/light-variant/
Visit that example

**Expected behavior**
The hamburger should be of a high contrast black color

**Is this issue blocking you?**
Override the class

**What is your environment?**
 - OS: Windows 10
 - Browser Chrome
 - Version 121.0.6167.185

mcoker:
Hey @ChrisLee5000! Can you share your use case of using the light variant of the masthead? A link or just a screenshot would be great.

ChrisLee5000:
Hi mcoker, I can't share my application, but the usage and bug is as shown in the example. I guess what I'm expecting is the hamburger should be a higher contrast color when using the light variant.

mcoker:
@ChrisLee5000 sorry for the delay in getting back. And I'm sorry you've run into this bug, this is a pretty big oversight on our part. That said, we discussed [the fix I put up](https://github.com/patternfly/patternfly/pull/6418) as a group and would prefer seeing if we can give you a workaround. The reasons being, 1) this bug has been there for over a year, so if anyone is using the light variant of the masthead, they've likely already worked around it. Pushing a fix at this point would likely be disruptive to any existing workarounds. 2) The light variant of the masthead was added without a use case and isn't currently recommended by our design teams for use in PF apps. 3) We're currently working to release a new major version of PF later this year, which includes a new look/feel from design, and this variant is going to be removed, so it isn't something we continue to plan on supporting.

The override would look like this:

```
.pf-v5-c-masthead .pf-v5-c-button.pf-m-plain {
  --pf-v5-global--Color--100: var(--pf-v5-global--Color--dark-100);
  --pf-v5-global--Color--200: var(--pf-v5-global--Color--dark-200);
  --pf-v5-global--BorderColor--100: var(--pf-v5-global--BorderColor--dark-100);
  --pf-v5-global--primary-color--100: var(--pf-v5-global--primary-color--dark-100);
  --pf-v5-global--link--Color: var(--pf-v5-global--link--Color--dark);
  --pf-v5-global--link--Color--hover: var(--pf-v5-global--link--Color--dark--hover);
[...]

PR #6418, opened 2024-03-14
fix(masthead): fix plain button color in light variants

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6390

While it's a simple fix, I'm hesitant to make this change since:
* The style has been in PF since 8/2022 so folks may have worked around it, and our fix may conflict with theirs https://github.com/patternfly/patternfly/pull/4827
* It could have unintended side effects. It will change any plain button inside of the masthead, which is a wide spread change since it includes nested components that may have their own backgrounds where this change isn't wanted. 
* The masthead is a very prominent part of the page, so any unwanted change has a high impact.

From this comment, it looks like the light variants weren't part of the original design or intended to be used? https://github.com/patternfly/patternfly/pull/3716#pullrequestreview-542902494 @lboehling @andrew-ronaldson do you know if we support and encourage use of this variant? Just wondering how widespread a change might be and what might be needed to test. We also have some analytics that might give us an idea of how commonly this variant is used.

### DIFF
--- a/src/patternfly/components/Masthead/masthead.scss
+++ b/src/patternfly/components/Masthead/masthead.scss
@@ -174,17 +174,21 @@ $pf-v5-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
   background-color: var(--#{$masthead}--BackgroundColor);
 
   &.pf-m-light {
-    @include pf-v5-t-light;
-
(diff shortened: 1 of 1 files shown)


mcoker:
closed https://github.com/patternfly/patternfly/issues/6390

PR #6423, opened 2024-03-14
chore(table): adjusted plain button fitting

### LINKED ISSUE #5728: Consume tokens: Table

mattnolting:
closes #5728 

Because the plain button has `min-width: 37px`, browsers implicitly create an icon wrapper that resolves to `21px` width: padding-left - padding-right (37px - 8px - 8px = 21px) or 21px in this case. `.fas` icons misalign because their `line-height` is set to 1, which overrides PatternFly set `line-height`s, causing more vertical alignment inconsistencies. Using `flexbox` removes the icon from inline context, allowing it to respond to alignment and justification props. Furthermore, passing a `.fas` icon (`line-height: inherit`) allows it to retain proper height, matching that of its container.

**Inline context:**

Screenshot 2024-03-01 at 5 08 02 PM

**Block context**

Screenshot 2024-03-01 at 5 08 31 PM

**Flexbox context**

Screenshot 2024-03-01 at 5 08 55 PM

As you can see, `inline` and `block` contexts slightly misalign icons vertically and horizontally, and each are slightly different from each other.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -253,8 +253,6 @@
   --#{$table}__thead--m-nested-column-header__tr--PaddingBottom: var(--pf-t--global--spacer--md);
 
   // * Table subhead
-  --#{$table}__subhead--PaddingLeft: var(--pf-t--global--spacer--sm);
-  --#{$table}__subhead--PaddingRight: var(--pf-t--global--spacer--sm);
   --#{$table}__subhead--Color: var(--pf-t--global--text--color--subtle);
 
   // * Table subhead button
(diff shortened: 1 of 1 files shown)


mcoker:
From the meeting yesterday, we're just looking at the horizontal alignment aligning the left edge of the star between the header and body. This PR seems focused on lots of other things that I think we should look at separately, ideally with actual use cases and as a group we can look at whether it's an issue, how far reaching the issue and fix are, and if we can prioritize it with the other work for penta.

However, this PR does address that alignment we discussed, and the key part is you used a min-width in the table button that matches the min-width in the button. If we fix the first/last-child padding bug, the fix in the PR is 1) min-width in the table button for its icon, 2) center the icon. I dropped that in a codepen here - https://codepen.io/mcoker/pen/XWQjvpG

Since this isn't critical, I'd like to bring this back to design and just outline this issue since it has to do with the min-width/aspect we're using for plain actions (requested by design) and we'll probably have this kind of thing pop up a lot. And explore some other options, too.

Issue #6055, opened 2023-11-10
[Label] - allow clickable labels to be disabled


Gkrumbach07:
**Is this a new component or an extension of an existing one?**
This is using the Label component with `onClick` defined.

**Describe the feature**
Because labels can now be clickable and act as buttons, they should also be able to have a disabled state with the accompanying styling.

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
[Code sandbox](https://codesandbox.io/s/modest-sanderson-5qpgdt?file=/fonts.css) provided by @mcoker 
![image](https://github.com/patternfly/patternfly/assets/12587674/cc5f56e4-6243-4d76-8e49-79bdc453215a)

**Any other information?**
This was talked about in a slack thread [here](https://patternfly.slack.com/archives/C293LQ36J/p1699429646350539) where we had a list of labels and a clickable label that routed to a new page. In some case that label button needed to be disabled. We didn't want to apply any custom styling so we instead had to use a link variant button to sort of match the non disabled label.

wise-king-sullyman:
Hello @Gkrumbach07, thanks for submitting this issue but I don't think we're quite understanding the use case.

Would a Button be a better fit here? Or put another way, is there a reason that a Button can't be used in place of a clickable label for your use case?

Gkrumbach07:
@wise-king-sullyman We tried a button but it ended up looking too bulky for the table cell we had it in.

This is the UX we have now
![image](https://github.com/patternfly/patternfly/assets/12587674/96ffbcd8-5443-4647-8452-77045be8135e)

The table cell displays a label list with a button to essentially add more to the list. But we wanted the ability to disable this button, and the small version of a button was still not the same font size as the label. Our disabled solution is using a button, but we had to manually make the font size `xs` to match the labels.

Ideally we would like to just be able to disable the label if it is clickable, becuase that was our original design

wise-king-sullyman:
Thanks @Gkrumbach07! We've slated this for work in our Q1 milestone

PR #6424, opened 2024-03-16
fix(label): added support for disabled labels

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6055

* Moves the template hbs file to the root label component dir since the workspace doesn't auto-compile-n-reload on files nested in subdirs.
* Sets the close button color in a disabled label to the same color as the disabled text, since by default a disabled close button color is the same color as the disabled label background.
* Just setup to work on regular (filled/outline) labels - not the overflow or add. It's easy enough to add support for those, or we could add that support if/when it comes up. wdyt @lboehling @andrew-ronaldson?

### DIFF
--- a/src/patternfly/components/Label/label-content.hbs
+++ b/src/patternfly/components/Label/label-content.hbs
@@ -6,7 +6,7 @@
     href="#"
     {{#if label--IsDisabled}}
       tabindex="-1"
-      aria-hidden="true"
+      aria-disabled="true"
     {{/if}}
   {{/if}}
   {{#if label-content--IsAction}}
--- a/src/patternfly/components/Label/label-content.hbs
+++ b/src/patternfly/components/Label/label-content.hbs
@@ -4,6 +4,10 @@
 <{{#if label-content--IsLink}}a{{else if label-content--IsAction}}button{{else}}span{{/if}} class="{{pfv}}label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
   {{#if label-content--IsLink}}
     href="#"
+    {{#if label--IsDisabled}}
+      tabindex="-1"
+      aria-hidden="true"
+    {{/if}}
   {{/if}}
   {{#if label-content--IsAction}}
     type="button"
--- a/src/patternfly/components/Label/label-content.hbs
+++ b/src/patternfly/components/Label/label-content.hbs
@@ -1,13 +1,19 @@
 {{#if label--IsEditable}}
   {{#> label-editable-content}}{{> @partial-block}}{{/label-editable-content}}
 {{else}}
-<{{#if label-content--IsLink}}a{{else}}span{{/if}} class="{{pfv}}label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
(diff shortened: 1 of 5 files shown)


srambach:
@mcoker looks like a11y tests are failing on 
92/266    /components/label
violations: duplicate-id-aria

src/patternfly/components/Label/label-template-variants.hbs
```diff
@@ -42,6 +55,22 @@
     label-text--attribute='style="--pf-v5-c-label__text--MaxWidth: 38ch"'
     label-icon--value="info-circle"}}
 
+{{> label
+    label--isRemovable=true
+    label--id=(concat label-template-variants--id '-link-removable')
```
srambach, line 60: Probably want to add -disabled to the id?

src/patternfly/components/Label/label-template-variants.hbs
```diff
@@ -42,6 +55,22 @@
...
+    label--IsDisabled=true
+    label-text--value=(concat label-template-variants--title ' link removable (disabled)')
+    label-content--IsLink=true
+    label-icon--value="info-circle"}}
+
+{{> label
+    label--isRemovable=true
+    label--id=(concat label-template-variants--id '-clickable-removable')
```
srambach, line 68: Probably want to add -disabled to the id?

srambach (commented):
Other than the duplicate ids, looks great! 🏷️

srambach (approved):
🚀

src/patternfly/components/Label/label-template-variants.hbs
```diff
@@ -42,6 +55,22 @@
     label-text--attribute='style="--pf-v5-c-label__text--MaxWidth: 38ch"'
     label-icon--value="info-circle"}}
 
+{{> label
+    label--isRemovable=true
+    label--id=(concat label-template-variants--id '-link-disabled')
+    label--IsDisabled=true
+    label-text--value=(concat label-template-variants--title ' link removable (disabled)')
+    label-content--IsLink=true
+    label-icon--value="info-circle"}}
```
thatblindgeye, line 64: Similar to [my comment in the v6 PR ](https://github.com/patternfly/patternfly/pull/6445#discussion_r1532104508)about disabled link labels.

src/patternfly/components/Label/label-content.hbs
```diff
@@ -1,13 +1,23 @@
 {{#if label--IsEditable}}
   {{#> label-editable-content}}{{> @partial-block}}{{/label-editable-content}}
 {{else}}
-<{{#if label-content--IsLink}}a{{else}}span{{/if}} class="{{pfv}}label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
+<{{#if label-content--IsLink}}a{{else if label-content--IsAction}}button{{else}}span{{/if}} class="{{pfv}}label__content{{#if label-content--modifier}} {{label-content--modifier}}{{/if}}"
   {{#if label-content--IsLink}}
     href="#"
+    {{#if label--IsDisabled}}
+      tabindex="-1"
+      aria-hidden="true"
```
thatblindgeye, line 9: Might be better to use aria-disabled here instead. Just thinking if it would be confusing for there to be a "close" button for something that is hidden from AT.

mcoker: 🤦‍♂️ totally meant for that to be `aria-disabled`, nice catch.

thatblindgeye (approved):

Issue #6242, opened 2024-01-19
Menu toggle - support for validation/status


mcoker:
Similar to https://www.patternfly.org/components/menus/select/react-deprecated#validated-selections

PR #6434, opened 2024-03-19
fix(menu-toggle): added support for validation/status

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6242

### DIFF
--- a/src/patternfly/components/MenuToggle/examples/MenuToggle.md
+++ b/src/patternfly/components/MenuToggle/examples/MenuToggle.md
@@ -565,6 +565,21 @@ import './MenuToggle.css'
 {{/menu-toggle}}
 ```
 
+### Status
+```hbs isBeta
+{{> menu-toggle menu-toggle--IsSuccess=true menu-toggle--text="Success"}}
+
+<br/><br/>
+
+{{> menu-toggle menu-toggle--IsWarning=true menu-toggle--text="Warning"}}
+
+<br/><br/>
+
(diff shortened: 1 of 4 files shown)


andrew-ronaldson (approved):
Perfecto!

srambach (approved):
Looks great 👍

thatblindgeye (approved):
May not really pertain to the new example added to MenuToggle, but wherever a status toggle is used elsewhere in followups we should probably be sure examples/demos include some sort of helper text noting what the warning/error is.

mcoker:
@thatblindgeye great point! I debated adding this hidden `aria-live` block I found in the deprecated react select  https://github.com/patternfly/patternfly-react/blob/488cf77f807ccc04a6d67d1a539f82e35ecb23ed/packages/react-core/src/deprecated/components/Select/examples/Select.md?plain=1#L519-L521

But I wasn't sure if that was the best method, and if that belongs in the non-interactive core examples. WDYT?

thatblindgeye:
For the non-interactive example you added it may not really be necessary. Adding a description to the example would probably be a better option, just to have verbiage mentioning when/why to provide a helper text for warning/danger status toggles. I think it was planned that Core examples would eventually have descriptions added to them similar to what was done to several React components' examples, but it would be more beneficial to add a description sooner than later. I'd be good with that as a followup, though.

mcoker:
@thatblindgeye nice, sounds good to me! Created this issue - https://github.com/patternfly/patternfly/issues/6436

PR #6435, opened 2024-03-19
fix(Nav): updated var placement for nav flyout menus

### LINKED ISSUE #6327: Bug - Nav - flyout menus are broken
Looks like it was probably introduced with the penta menu PR
* https://github.com/patternfly/patternfly/pull/6167 
* https://patternfly-pr-6167.surge.sh/components/navigation#nav-with-flyout

The commit before that (menu toggle), flyout menu looks good
* https://github.com/patternfly/patternfly/pull/6177
* https://patternfly-pr-6177.surge.sh/components/navigation#nav-with-flyout

thatblindgeye:
Closes #6327

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -129,7 +129,7 @@
     flex-grow: 1;
   }
 
-  + .#{$menu} {
+  .#{$menu} {
     --#{$menu}--MinWidth: 100%;
   }
 }
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -104,8 +104,6 @@
 
 // - Nav root
 .#{$nav} {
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Nav/nav.scss
```diff
@@ -130,6 +128,10 @@
   &.pf-m-fill {
     flex-grow: 1;
   }
+
+  + .#{$menu} {
```
thatblindgeye, line 132: This is due to how the markup is setup in Core - a nav with a menu sibling. In React, though, the menu is a descendant of nav.

If the React markup is the correct one, I can update the nav flyout example in core to match and update this to remove this sibling combinator and just target any menu decendant of nav.

thatblindgeye: Updated to nsst the menu inside nav per off-GH discussion

mcoker (approved):
🚢

Issue #6442, opened 2024-03-19
Add disabled state for interactive labels


mcoker:
Follow up for the feature added to main https://github.com/patternfly/patternfly/issues/6055

thatblindgeye:
Closed by #6445

PR #6445, opened 2024-03-19
fix(label): added disabled styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6442

### DIFF
--- a/src/patternfly/components/Label/label--variants.hbs
+++ b/src/patternfly/components/Label/label--variants.hbs
@@ -2,37 +2,58 @@
   {{> label
       label--id=(concat label--variants--id '-status-default')
       label-text--value=(concat label--variants--title)}}
+
   {{> label
       label--IsRemovable=true
       label--id=(concat label--variants--id '-status-removable')
       label-text--value=(concat label--variants--title ' removable')}}
+
   {{> label
       label-content--IsLink=true
       label--id=(concat label--variants--id '-status-link')
       label-text--value=(concat label--variants--title ' link')}}
+
   {{> label
       label-content--IsButton=true
       label--id=(concat label--variants--id '-status-button')
       label-text--value=(concat label--variants--title ' button')}}
+
   {{> label
       label--IsRemovable=true
       label-content--IsLink=true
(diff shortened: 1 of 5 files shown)


andrew-ronaldson (approved):
i think it looks good. Thanks

src/patternfly/components/Label/label--variants.hbs
```diff
@@ -2,37 +2,58 @@
...
       label--id=(concat label--variants--id '-status-truncated-removable')
       label-text--value=(concat label--variants--title ' removable, max-width truncation customization')
       label-text--attribute='style="--pf-v5-c-label__text--MaxWidth: 28ch"'}}
+
+  {{> label
+      label--IsRemovable=true
+      label--IsDisabled=true
+      label-content--IsLink=true
+      label--id=(concat label--variants--id '-status-link-removable-disabled')
+      label-text--value=(concat label--variants--title ' link removable (disabled)')}}
```
thatblindgeye, line 49: I don't think we should show link labels being disabled (at least not the whole label). While you can't click the link to activate it, you can still activate/focus it via keyboard. Also not sure when it may make sense to disable an `<a>` element, and at that point it's not really a link but rather just text.

If this variant could be updated so that only the "remove" button is disabled, but the rest of the label has its normal styling (and the link works on mouse click) it could be worth keeping in, but otherwise probably better to just remove it.

mcoker: Great feedback! Just for reference from our convo offline, adding for now since we have this as an established (anti-?)pattern already in the library, and will investigate a better approach to this via [the spike issue you created](https://github.com/patternfly/patternfly/issues/6448). If we can fit this change in the timeline for penta, that would be great!

mcoker:
@thatblindgeye would you mind re-reviewing when you get a chance? Added `aria-hidden` and `tabindex="-1"` (matches the button and other places we're disabling links/non-buttons), which I just missed when making this update initially.

thatblindgeye:
@mcoker left a comment on the other label PR, but did you mean to use aria-disabled instead to match the button as links example?

srambach (approved):
👍

PR #6447, opened 2024-03-19
feat(wizard): add wizard step status

### LINKED ISSUE #5142: Wizard - support for step status
React has custom styles for an invalid step that we should bring back to core.

Here's the example https://patternfly-react-v6.surge.sh/components/wizard/#step-error-status.

srambach:
Fixes #5142 

This adds an icon element and styles it to appear where the step number currently appears (hiding the step number if so). 
A new modifier is added for danger status.
A new full page example is added to wizard.

The size of the step number was adjusted to match the xl icon size per @lboehling

### DIFF
--- a/src/patternfly/components/Wizard/examples/Wizard.md
+++ b/src/patternfly/components/Wizard/examples/Wizard.md
@@ -85,7 +85,7 @@ import './Wizard.css'
 
 ### Nav expanded (mobile)
 ```hbs isFullscreen
-{{#> wizard wizard--IsExpanded="true"}}
+{{#> wizard wizard--id="wizard-nav-expanded" wizard--IsExpanded="true"}}
   {{#> wizard-header}}
     {{> wizard-close}}
     {{> wizard-title wizard-title-text--value="Wizard title"}}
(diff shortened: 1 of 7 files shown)


src/patternfly/components/Wizard/wizard-nav-link-icon.hbs
```diff
@@ -0,0 +1,6 @@
+<span class="{{pfv}}wizard__nav-link-icon{{#if wizard-nav-link-icon--modifier}} {{wizard-nav-link-icon--modifier}}{{/if}}"
```
mcoker, line 1: I might rename this from icon to status-icon, just so we can leave room for a step icon that we can target uniquely from the status icon and not run into a namespace issue. WDYT?

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -48,6 +48,14 @@
   --#{$wizard}__nav-link--m-disabled--Color: var(--pf-t--global--text--color--disabled);
   --#{$wizard}__nav-link--m-disabled--BackgroundColor: transparent;
 
+  // Nav link status icon
+  --#{$wizard}__nav-link-icon--Color: var(--pf-t--global--icon--color--regular);
+  --#{$wizard}__nav-link--m-danger--nav-link-icon--Color: var(--pf-t--global--icon--color--status--danger--default);
```
mcoker, line 53: Technically I think you'd use `--m-danger__nav-link-icon` (since it's a nested element) but since the var starts at `__nav-link` I think you'd also be alright just calling this `__icon` if you wanted to shorten the var.

```suggestion
  --#{$wizard}__nav-link--m-danger__icon--Color: var(--pf-t--global--icon--color--status--danger--default);
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -454,6 +462,15 @@
     }
   }
 
+  &.pf-m-danger {
+    --#{$wizard}__nav-link-status-icon--Color: var(--#{$wizard}__nav-link--m-danger__nav-link-status-icon--Color);
+
+    &::before {
+      // display: none;
+      opacity: 0;
+    }
```
thatblindgeye, line 471: Did we want to use the opacity or display property here?

thatblindgeye (commented):
This is intended to replace how React currently renders error icons in the [step error status example](https://www.patternfly.org/components/wizard#step-error-status) I assume?

src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -459,6 +459,157 @@ import './Wizard.css'
...
+    {{> wizard-close}}
+    {{> wizard-title wizard-title-text--value="Wizard title"}}
+    {{#> wizard-description}}
+      Here is where the description goes
+    {{/wizard-description}}
+  {{/wizard-header}}
+  {{#> wizard-toggle}}
+    {{#> wizard-toggle-list}}
+      {{#> wizard-toggle-list-item wizard-toggle-list-item--modifier='pf-m-danger'}}
+        {{#> wizard-toggle-status-icon}}<i class="fas fa-exclamation-circle" aria-hidden="true"></i>{{/wizard-toggle-status-icon}}
```
thatblindgeye, line 550: Similar to above comment that was resolved, not a blocker but could render the icon inside `wizard-toggle-status-icon` rather than needing to pass it in examples. Maybe even rendering the status icon inside of `toggle-list-item` when isDanger=true?

src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -459,6 +459,157 @@ import './Wizard.css'
...
+    {{> wizard-title wizard-title-text--value="Wizard title"}}
+    {{#> wizard-description}}
+      Here is where the description goes
+    {{/wizard-description}}
+  {{/wizard-header}}
+  {{#> wizard-toggle}}
+    {{#> wizard-toggle-list}}
+        {{#> wizard-toggle-list-item}}
+          {{#> wizard-toggle-num}}2{{/wizard-toggle-num}}
+          Configuration
```
thatblindgeye, line 476: Think this list item needs the status icon passed in as well. On wider viewport the 2nd step has an error icon, but on smaller viewport the toggle doesn't show the error icon. That or the 2nd step for desktop view should have the error icon removed (so that only the 1st step has an error).

Wider viewport:

image

Smaller viewport:

image

thatblindgeye (approved):
We'll need a followup in React to adjust the error styling there to match, but otherwise looks good

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -478,6 +509,26 @@
...
+  padding-block-start: var(--#{$wizard}__nav-link--PaddingBlockStart);
+  padding-block-end: var(--#{$wizard}__nav-link--PaddingBlockEnd);
+  padding-inline-start: var(--#{$wizard}__nav-link--PaddingInlineStart);
+  padding-inline-end: var(--#{$wizard}__nav-link--PaddingInlineEnd);
+  background-color: var(--#{$wizard}__nav-link--BackgroundColor);
+  border: none;
+  border-radius: var(--#{$wizard}__nav-link--BorderRadius);
+}
+
+.#{$wizard}__nav-link-status-icon {
```
mcoker, line 525: One thing to note about the font awesome status icons is that the warning-triangle icon is wider than the icons we use for success/danger, so if we put a warning icon by a step name, that step text will be pushed over farther than the step text above/below it. It's not a huge issue since the icons in react are square, and most users will use those, but if a user supplied their own icon that wasn't a square, that may cause an issue.

I wonder if this element should share the shape size we define for the count circles, or we should reconsider using absolute positioning/translate here like it was before. Or possibly use subgrid for the number/status icon, though worth noting the column width would shift for all steps if a wider status icon was added to a step.

Screenshot 2024-04-18 at 5 07 06 PM
Screenshot 2024-04-18 at 5 12 14 PM
Screenshot 2024-04-18 at 5 12 22 PM

src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -459,6 +459,155 @@ import './Wizard.css'
 {{/wizard}}
 ```
 
+### Error on step
+```hbs isFullscreen
+{{#> wizard wizard--id="wizard-basic"}}
```
mcoker, line 464:
```suggestion
{{#> wizard wizard--id="wizard-error-on-step"}}
```

src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -459,6 +459,155 @@ import './Wizard.css'
...
+      {{/wizard-main}}
+    {{/wizard-inner-wrap}}
+    {{> wizard-footer}}
+  {{/wizard-outer-wrap}}
+{{/wizard}}
+```
+
+### Nav expanded with error (mobile)
+```hbs isFullscreen
+{{#> wizard wizard--IsExpanded="true"}}
```
mcoker, line 538: Looks like we're missing a `wizard--id` here and in the OG expanded example https://github.com/patternfly/patternfly/blob/dc3604d18826769957b3a1a6271ca095453fd333/src/patternfly/components/Wizard/examples/Wizard.md?plain=1#L88

Would you mind adding one here and in the other example? That's used for a few things, including the form that gets added to these wizards

Screenshot 2024-04-18 at 5 23 13 PM

src/patternfly/components/Wizard/wizard-nav-link-main.hbs
```diff
@@ -0,0 +1,6 @@
+<span class="{{pfv}}wizard__nav-link-main{{#if wizard-nav-link-main--modifier}} {{wizard-nav-link-main--modifier}}{{/if}}"
```
mcoker, line 1: can you add this element to the component docs?

src/patternfly/components/Wizard/examples/Wizard.md
```diff
@@ -510,4 +660,5 @@ import './Wizard.css'
 | `.pf-m-expanded` | `.pf-v6-c-wizard__nav-item` | Modifies a nav item for the expanded state. |
 | `.pf-m-current` | `.pf-v6-c-wizard__nav-link` | Modifies a step link for the current state. **Required** |
 | `.pf-m-disabled` | `.pf-v6-c-wizard__nav-link` | Modifies a step link for the disabled state. |
+| `.pf-m-danger` | `.pf-v6-c-wizard__nav-link` | Modifies a step link to indicate danger status. |
```
mcoker, line 663: looks like you can also add it to the list-item in the mobile nav

```suggestion
| `.pf-m-danger` | `.pf-v6-c-wizard__nav-link`, `pf-v6-c-wizard__toggle-list-item` | Modifies a step link to indicate danger status. |
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -57,44 +64,54 @@
...
   --#{$wizard}__toggle-list-item--not-last-child--MarginRight: var(--pf-t--global--spacer--sm);
   --#{$wizard}__toggle-list-item--MarginBottom: var(--pf-t--global--spacer--xs);
-
+  --#{$wizard}__toggle-list-item--Gap: var(--pf-t--global--spacer--sm);
+
+  // Toggle status icon
+  --#{$wizard}__nav-link-status-icon--LineHeight: 1;
+  --#{$wizard}__toggle-status-icon--Color: var(--pf-t--global--icon--color--regular);
+  --#{$wizard}__toggle--m-danger__nav-link-status-icon--Color: var(--pf-t--global--icon--color--status--danger--default);
+  --#{$wizard}__toggle-status-icon--Top: calc(var(--pf-t--global--spacer--xs) / 2);
```
mcoker, line 107: I could be wrong here, but it looks like this use of `spacer--xs` doesn't correspond with alignging the status-icon with something else that uses `spacer--xs` as a top spacer... this calc just returns 2px which looks to be about the adjustment that works visually to center the icon? If so, I _might_ just use `2px` here instead? It's really a nit, but if a user (or we as PF) themed/customized `spacer--xs` for a compact theme or whatever, I don't think we would want this value to change necessarily. I think the same comment applies to https://github.com/patternfly/patternfly/blob/dc3604d18826769957b3a1a6271ca095453fd333/src/patternfly/components/Wizard/wizard.scss#L82 

Kind of a philosophical question around using a token because it gives the look you're going for, versus it being the right token given the token's meaning. Since our spacer tokens are so generic, I don't think it matters, but figured I'd leave a comment for light reading  😅

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -478,6 +509,26 @@
...
+  display: flex;
+  flex-grow: 1;
+  justify-content: space-between;
+  padding-block-start: var(--#{$wizard}__nav-link--PaddingBlockStart);
+  padding-block-end: var(--#{$wizard}__nav-link--PaddingBlockEnd);
+  padding-inline-start: var(--#{$wizard}__nav-link--PaddingInlineStart);
+  padding-inline-end: var(--#{$wizard}__nav-link--PaddingInlineEnd);
+  background-color: var(--#{$wizard}__nav-link--BackgroundColor);
+  border: none;
+  border-radius: var(--#{$wizard}__nav-link--BorderRadius);
```
mcoker, line 522: should we rename these?

```suggestion
  padding-block-start: var(--#{$wizard}__nav-link-main--PaddingBlockStart);
  padding-block-end: var(--#{$wizard}__nav-link-main--PaddingBlockEnd);
  padding-inline-start: var(--#{$wizard}__nav-link-main--PaddingInlineStart);
  padding-inline-end: var(--#{$wizard}__nav-link-main--PaddingInlineEnd);
  background-color: var(--#{$wizard}__nav-link-main--BackgroundColor);
  border: none;
  border-radius: var(--#{$wizard}__nav-link-main--BorderRadius);
```

mcoker (commented):
woot! Left some comments, lemme know what ya think!

src/patternfly/components/Wizard/wizard.scss
srambach: @mcoker Can we defer this pending a decision about the new icon set?

mcoker: Yep, good plan 👍

mcoker (approved):
LGTM! 🚀

PR #6451, opened 2024-03-21
fix(check): changed height to min-height

### LINKED ISSUE #6450: Bug - Check - standalone variant height in menu-toggle is unhappy
**Describe the problem**
This is an issue with 
- [ ] Patternfly 6  

Standalone check height needs to be min-height

Screenshot 2024-03-21 at 10 08 23 AM

mattnolting:
closes #6450

### DIFF
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -4,7 +4,7 @@
 .#{$check} {
   --#{$check}--GridGap: var(--pf-t--global--spacer--sm) var(--pf-t--global--spacer--sm);
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Check/check.scss
```diff
@@ -4,7 +4,7 @@
 .#{$check} {
   --#{$check}--GridGap: var(--pf-t--global--spacer--sm) var(--pf-t--global--spacer--sm);
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
-  --#{$check}--Height: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
+  --#{$check}--MinHeight: 1lh;
```
mcoker, line 7: Can't use this quite yet - I think it was added in Safari 16.(4?).

mcoker (approved):
Just the one nit!

PR #6455, opened 2024-03-22
chore(ci): replace stale bot with stale action

### LINKED ISSUE #6005: Stalebot - update to stale github action
We use stalebot but it's no longer maintained - https://github.com/apps/stale

It's recommended to use the stale action instead - https://github.com/actions/stale

We should also add a "stale" label and configure the action to set that label on issues identified as stale. The action also allows us to apply that label manually as another way to mark an issue as stale.

wise-king-sullyman:
Closes #6005


.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-stale: 60
+          days-before-close: 30
+          exempt-issue-labels: 'accessibility,breaking change,design approved, pinned'
```
mcoker, line 14: nit

```suggestion
          exempt-issue-labels: 'accessibility,breaking change,design approved,pinned'
```

mcoker (approved):

.github/workflows/stale.yml
wise-king-sullyman: Great catch!

PR #6461, opened 2024-03-25
feat(table-button): added table button plain

### LINKED ISSUE #6437: Table - sort button
Table sort button 

**Describe the feature**
Needs a `pf-m-plain` variant that matches the button variant. The variant belongs to `pf-v6-c-table__button` (@mcoker) as they have yet to be updated to `menu-toggle`s.

Screenshot 2024-03-19 at 11 08 48 AM

mattnolting:
closes #6437

### DIFF
--- a/src/patternfly/components/Table/table-favorite-sort.hbs
+++ b/src/patternfly/components/Table/table-favorite-sort.hbs
@@ -1,8 +1,12 @@
 {{#> action-list action-list--modifier='pf-m-icons'}}
-  {{#> button button--IsPlain=true button--aria-label='Favorite all'}}
-      <i class="fas fa-star" aria-hidden="true"></i>
-  {{/button}}
-  {{#> button button--IsPlain=true button--aria-label='Sort favorites'}}
-    {{> table-sort-indicator}}
-  {{/button}}
+  {{#> action-list-item}}
+    {{#> button button--IsPlain=true button--aria-label='Favorite all'}}
+        <i class="fas fa-star" aria-hidden="true"></i>
+    {{/button}}
+  {{/action-list-item}}
+  {{#> action-list-item}}
+    {{#> button button--IsPlain=true button--aria-label='Sort favorites'}}
+      {{> table-sort-indicator}}
+    {{/button}}
+  {{/action-list-item}}
 {{/action-list}}
--- a/src/patternfly/components/Table/table-favorite-sort.hbs
+++ b/src/patternfly/components/Table/table-favorite-sort.hbs
@@ -1,0 +1,18 @@
+{{#> action-list action-list--modifier='pf-m-icons'}}
+  {{#>button
+      button--IsPlain=true
+      button--id=(dasherize table--id 'favorite-button' table-tr--index)
+      button--aria-label=(ternary table-td--IsFavorited 'Starred' 'Not starred')
+      button--aria-labelledby=(dasherize table--id 'node' table-tr--index table--id 'favorite-button' table-tr--index)
+    }}
+      <i class="fas fa-star" aria-hidden="true"></i>
+  {{/button}}
+  {{#>button
+      button--IsPlain=true
+      button--id=(dasherize table--id 'sort-button' table-tr--index)
+      button--aria-label=(ternary table-td--IsFavorited 'Starred' 'Not starred')
+      button--aria-labelledby=(dasherize table--id 'node' table-tr--index table--id 'favorite-button' table-tr--index)
+    }}
+    {{> table-sort-indicator}}
+  {{/button}}
+{{/action-list}}
--- a/src/patternfly/components/Table/table-favorite-sort.hbs
+++ b/src/patternfly/components/Table/table-favorite-sort.hbs
@@ -1,18 +1,8 @@
 {{#> action-list action-list--modifier='pf-m-icons'}}
-  {{#>button
-      button--IsPlain=true
-      button--id=(dasherize table--id 'favorite-button' table-tr--index)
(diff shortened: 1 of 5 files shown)


mcoker:
@kaylachumley for this action for [this table favorite column header action](https://patternfly-pr-6461.surge.sh/components/table/html/favorites-sortable-example/)

Screenshot 2024-05-06 at 10 21 29 AM

1. Just to confirm, should this be a single action that clicking on it just sorts the column? Or should this be a "split button" (2 actions)  where the star icon would favorite all/none rows and the sort icon would sort the column? We do not currently have a "split" variation of the button component, so this would be a new enhancement. 
3. If this is a single icon, should we add a variation of the button that supports 2 (or more) icons and no text? That would be an enhancement, too.

andrew-ronaldson:
I wonder if we can do two plain buttons inside a wrapper. Kind of like a split button minus the extra styling. Not sure this example is widely used so this may be enough for a first pass. Here is a mock.
https://github.com/patternfly/patternfly/assets/28658548/08639792-53ea-4b65-acea-6b08751dd9e3

mattnolting:
> I wonder if we can do two plain buttons inside a wrapper. Kind of like a split button minus the extra styling. Not sure this example is widely used so this may be enough for a first pass. Here is a mock. https://github.com/patternfly/patternfly/assets/28658548/08639792-53ea-4b65-acea-6b08751dd9e3

@andrew-ronaldson that wfm. @mcoker @srambach wdyt?

mcoker:
> @andrew-ronaldson that wfm. @mcoker @srambach wdyt?

👍👍

andrew-ronaldson:
> If it's two different things - star everything and also sort on the starred items, that seems way better. If you can't star everything at once, it shouldn't be a button and then you just have the sort button.

Agreed with Sarah above.

srambach:
I'm seeing broken handlebars at https://patternfly-pr-6461.surge.sh/components/table#favorites-sortable-example

src/patternfly/components/Table/table-favorite-sort.hbs
```diff
@@ -0,0 +1,18 @@
...
+      button--id=(dasherize table--id 'favorite-button' table-tr--index)
+      button--aria-label=(ternary table-td--IsFavorited 'Starred' 'Not starred')
+      button--aria-labelledby=(dasherize table--id 'node' table-tr--index table--id 'favorite-button' table-tr--index)
+    }}
+      <i class="fas fa-star" aria-hidden="true"></i>
+  {{/button}}
+  {{#>button
+      button--IsPlain=true
+      button--id=(dasherize table--id 'sort-button' table-tr--index)
+      button--aria-label=(ternary table-td--IsFavorited 'Starred' 'Not starred')
```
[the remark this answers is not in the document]
thatblindgeye: Yeah, I'd expect the first button to just have `aria-label="Not starred"` and the second to have something like `aria-label="Sort favorites"` (or something), without any aria-labelledby on either.

Tried throwing something together quickly in React to mimic this a bit, and something like `Sort favorites in X order`, where X is the opposite of the current sort order e.g. if currently sorting favorites in ascending order, "Sort favorites in descending order" could work, too, as clicking the button may not convey anything has happened. Currently clicking a sortable header button doesn't convey that either, though, so just a standard "Sort column" aria-label may be better while we decide how to handle this pre-existing issue.

src/patternfly/components/Table/table-favorite-sort.hbs
```diff
@@ -0,0 +1,18 @@
+{{#> action-list action-list--modifier='pf-m-icons'}}
+  {{#>button
```
thatblindgeye, line 2: Is this missing the action-list-item component around the buttons?

andrew-ronaldson (approved):
Rubber stamp of approval!

src/patternfly/components/Table/table-favorite-sort.hbs
mattnolting: EagleEyeBlindGuy 😆 I'll add `.action-list__item`

thatblindgeye (approved):

src/patternfly/components/Table/table-favorite-sort.hbs
thatblindgeye: 🦅

srambach (approved):
Looks ok now to me ⏯️

src/patternfly/components/Table/table.scss
```diff
@@ -593,6 +593,14 @@
       border-block-end: 0;
     }
   }
+
+  .#{$action-list}.pf-m-icons {
+    --#{$action-list}--m-icons--ColumnGap: 0;
+  }
```
mcoker, line 599: This needs to be better scoped - it compiles to `.pf-v6-c-table .pf-v6-c-action-list.pf-m-icons`, which will apply to any use of `.pf-v6-c-action-list.pf-m-icons` anywhere in a table.

If all this is doing is putting 2 things side by side in a flex layout, could it just be a local style to the table?

mcoker (changes requested):

src/patternfly/components/Table/table.scss
mattnolting: I added action list and modified the spacer as you suggested. Also left a comment regarding the icon- group spacing being inconsistent with all other icon groups. I disagreement that it's scoped poorly.

mcoker (approved):
🚀

mcoker:
@mattnolting would you mind opening a react issue for this change with a tl;dr of the changes needed in react?

PR #6464, opened 2024-03-26
chore(tokens): updated form element spacers to tokens

### LINKED ISSUE #6363: Add form element semantic spacer
Add a form element semantic spacer (the equivalent of `--#{$pf-global}--spacer--form-element`) and set it to the small spacer token, and update any existing references of `--#{$pf-global}--spacer--form-element` to the new token.

Then as a follow up for later (probably?) update any form-esque components that are already using the small spacer to use the new spacer token.

thatblindgeye:
Closes #6363 

Only updated TODOs/existing form element vars to tokens. Started to update instances of t-shirt spacers, but that may depend on naming (for example, plain buttons would need a new token added, but does `--pf-t--global--spacer--button--plain--default`/`--pf-t--global--spacer--button--plain--compact` make sense?). Also using the same spacer on different components could cause different spacing than possibly intended (had placed a spacer on the Text Input Group left/right padding as well as a Drawer close button's padding or margin, and it results in one of them having significant difference than current spacing; couldn't really think of a good name for a new token other than maybe "--enlarged" or "--extra-compact" and such)

### DIFF
--- a/src/patternfly/base/tokens/_tokens-font.scss
+++ b/src/patternfly/base/tokens/_tokens-font.scss
@@ -145,9 +145,5 @@
     var(--pf-t--global--box-shadow--blur--lg) 
     var(--pf-t--global--box-shadow--spread--lg)
     var(--pf-t--global--box-shadow--color--lg);
-
-  // Spacers
-  --pf-t--global--spacer--control--vertical--compact: var(--pf-t--global--spacer--100);
-  --pf-t--global--spacer--control--horizontal--compact: var(--pf-t--global--spacer--200);
 }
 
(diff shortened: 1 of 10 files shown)


andrew-ronaldson (approved):
I think this looks good...🤞

srambach:
This seems to have some differences - the alert is getting taller and then the text aligns oddly
image

And the card is taller than it used to be
image

And form control is definitely off :-(

srambach (approved):
I see a teensy shift from before in card and in where the X is on alerts, but it seems to be lined up with the text so afaict it's good!

Issue #6366, opened 2024-02-28
Update all references to v5-v6


mattnolting:
- [x] .scss variables need to be resolved
  - [x] mixins
  - [x] functions
  - [x] icons
- [x] Utilities
- [x] Layouts
- [x] Markdown
  - [x] References to css variables
  - [x] References to selectors
- [x] Investigate icons

### Tackling this as:
- [x] Breakpoint mixins
- [x] RTL mixins
- [x] Other mixins
- [x] #6481 
- [x] #6483
- [x] #6494

PR #6468, opened 2024-03-27
chore(mixins): change version prefix on breakpoint mixins and variables

srambach:
Updates v5 -> v6 on breakpoint mixins and variables.

One part of #6366, more to come.


mcoker (approved):
👍

PR #6480, opened 2024-03-28
fix(TreeView): improved a11y experience

### LINKED ISSUE #6478: Bug - TreeView - Improve a11y experience
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

When Eric and I were reviewing the Tree View, we noticed some improvements that could be made ([summary of conversation here](https://github.com/patternfly/patternfly-org/pull/3916)).

- Mention `aria-multiselectable` is needed when the tree view is intended to have multiple items selected. So that'd be more a docs update in Core
- Add `aria-selected` to tree view items when selection made (we should probably look into whether `aria-checked` should be applied instead of `aria-selected` for checkbox tree views)
- Add aria-labeling to gives context to what the tree view is of/for (either `aria-labelledby` to a heading or `aria-label`)

**Expected behavior**
To match WAI ARIA tree view: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/

thatblindgeye:
Closes #6478

### DIFF
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -12,7 +12,7 @@ cssPrefix: pf-v5-c-tree-view
   {{#> tree-view-list tree-view-list--IsRoot="true" tree-view-list--AriaLabel="Tree View single selectable example"}}
     {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
       {{#> tree-view-content}}
-        {{> tree-view-node tree-view-node--text="Application launcher"}}
+        {{> tree-view-node tree-view-node--text="Application launcher" tree-view-node--IsFocusable=true}}
       {{/tree-view-content}}
       {{#> tree-view-list newcontext}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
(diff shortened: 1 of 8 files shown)


src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -863,29 +960,16 @@ cssPrefix: pf-v5-c-tree-view
...
 ### Compact, no background
 ```hbs
-{{> tree-view--base tree-view--base--modifier="pf-m-compact pf-m-no-background"}}
+{{> tree-view--base tree-view-list--AriaLabel="Compact no background example" tree-view--base--modifier="pf-m-compact pf-m-no-background"}}
 ```
 
 ## Documentation
 
-### Accessibility
-
```
thatblindgeye, line 877: This section/table no longer needed with the a11y doc update being made in org.

jessiehuff (approved):
Makes sense to me, but definitely defer to @mcoker's opinion!

thatblindgeye:
@jessiehuff @mcoker The latest commit updates tabindices to match what is currently mentioned in the treeview a11y docs update PR. Basically only the first focusable item in each treeview example has a tabindex of 0, all other focusable elements have tabindex of -1. I've also removed the tabindex off of the `li[role="treeitem"]` element; that had matched WAI treeview examples, but our implementation is a bit different and React wasn't updating those tabindices.

The intent is that only a single element inside the treeview will ever have a 0 tabindex, and arrow keys must be used to navigate the treeview (updating the tabindices as you navigate through). The React keyboard navigation logic will need to be updated to work more properly (the PR I have up currently helps improve it).

The issue with having complicated tree views where there may be action items, checkboxes separate from the expansion, or selection separate from the expansion, is that it isn't clear at all that those things exist/are separate especially for screen reader. Similar to our Menu examples where there may be secondary actions, a user may not know they can navigate horizontally in addition to vertically. That, and Left/Right Arrow are expected to have specific functionality depending on the expanded state of the currently focused tree view node, which may be confusing/frustrating if the expectation is that "Right arrow should expand this parent node" but focus is instead moving from the expand toggle to the node text button.

 A basic tree view is probably easy enough to traverse and understand, but beyond that some sort of visible text description/instruction may need to be included. For example, something like:

image

What do you both think? Maybe this is something that requires a separate, more in depth discussion, but wanted to propose the above in any case.

jessiehuff (approved):

mcoker (approved):
🚀

Issue #6489, opened 2024-03-29
Sticky toolbars need a background color


mcoker:
Scrolling content behind toolbar is visible.

<img width="634" alt="Screenshot 2024-03-29 at 2 24 05 PM" src="https://github.com/patternfly/patternfly/assets/35148959/b24e0969-d973-442f-8203-f45a7e460d14">
<img width="325" alt="Screenshot 2024-03-29 at 2 23 57 PM" src="https://github.com/patternfly/patternfly/assets/35148959/f7369334-86ee-4581-926d-31b454b4d56f">

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6490

PR #6490, opened 2024-03-29
fix(toolbar): added sticky background, addl bg variants

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6489


srambach (approved):
🚀 🔥

PR #6499, opened 2024-04-01
fix(tokens): update ZIndex to z-index

### LINKED ISSUE #6498: Tokens - update to correct ZIndex to z-index
Just to be consistent, it looks like the pf-t--global--Zindex--[sizes] tokens should use a lower-case property name with a hyphen (pf-t--global--z-index), similar to the box-shadow tokens - pf-t--global--box-shadow

This has been done in Figma, so just needs to be pulled through.

srambach:
Pulls latest tokens and as a result, all tokens with "Zindex" are now "z-index"

Fixes #6498


mcoker (approved):
🥳

Issue #6519, opened 2024-04-05
Update dependencies for v6


mattnolting:
Mainly interested in the updated version of the sass package, but we can also see if there are other packages that look safe to bump.

PR #6523, opened 2024-04-05
chore(build): update packages

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6519

The deets:
* Updates all dependencies to latest using `ncu -u`
* Found a problem with the gulp major version bump from 4.x -> 5.x, so reverted the gulp and gulp-cli versions for this PR
  * Opened https://github.com/patternfly/patternfly/issues/6525 to do the gulp update separately
* Removes the gulp lintCSS scripts - that checks for 2 things, which we can do better than that script is doing
  * Checks for valid use of CSS functions compared to a static list of valid function names in the gulpfile. We just don't need that AFAIK, stylelint can do that for us, and furthermore functions will not work if invalid
  * Checks for CSS comments in the compiled CSS. This is good, but we actively maintain our stylelint rules, and stylelint-css has a rule for that, [comment-no-loud](https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/comment-no-loud/README.md) that I added to our [stylelint config](https://github.com/patternfly/patternfly/blob/c719a599770eca443d80739ceb1d51bba749ab80/.stylelintrc#L81)


mattnolting (approved):
Wonderful! 🚢 👍

wise-king-sullyman (approved):
🎸

PR #6538, opened 2024-04-09
chore(pf): updated @import to @use

### LINKED ISSUE #6518: @import -> @use/@forward
```[tasklist]
### Tasks
- [ ] https://github.com/patternfly/patternfly/issues/6519
- [ ] https://github.com/patternfly/patternfly/issues/6520
- [ ] https://github.com/patternfly/patternfly/issues/6524
- [ ] https://github.com/patternfly/patternfly/issues/6535
```

### LINKED ISSUE #6524: organize @forwards

mattnolting:
closes #6518 
closes #6524


src/patternfly/patternfly-no-globals.scss
```diff
@@ -1,4 +1,5 @@
-$pf-v6-global--enable-reset: false;
-$pf-v6-global--enable-normalize: false;
-
-@import "./patternfly";
+@use './config/helpers' with (
+  $pf-v6-global--enable-reset: false,
+  $pf-v6-global--enable-normalize: false
+);
+@use './patternfly-base.scss';
```
mcoker, line 5:
```suggestion
@use './patternfly.scss';
```

src/patternfly/patternfly-addons.scss
```diff
@@ -1,13 +1,11 @@
...
+@use './utilities/Accessibility/accessibility';
+@use './utilities/Alignment/alignment';
+@use './utilities/BackgroundColor/background-color';
+@use './utilities/BoxShadow/box-shadow';
+@use './utilities/Display/display';
+@use './utilities/Flex/flex';
+@use './utilities/Float/float';
+@use './utilities/Sizing/sizing';
+@use './utilities/Spacing/spacing';
+@use './utilities/Text/text';
```
mcoker, line 11: can you just import the dir?

```suggestion
@use './utilities';
```

src/patternfly/utilities/_index.scss
```diff
@@ -0,0 +1,10 @@
+@forward './Accessibility/accessibility';
+@forward './Alignment/alignment';
+@forward './BackgroundColor/background-Color';
```
mcoker, line 3:
```suggestion
@forward './BackgroundColor/background-color';
```

mcoker (changes requested):
NICE! Just a couple of nits I noticed looking at the code. Also when I run the dist build, it's taking about twice as long (~40s compared to ~20s before), any idea what's causing that?

src/patternfly/patternfly-no-globals.scss
mattnolting: https://sass-lang.com/dart-sass/#java-script-library the [sass package compiler ](https://sass-lang.com/dart-sass/#java-script-library) is slower than [using dart sass compiler](https://sass-lang.com/dart-sass/#dart-library). I'm happy to update the yarn script and test https://docs.google.com/document/d/1wd_ppsKREAibZlrhn5v3-BlvJgD71s-vBWYIQ31aok0/edit#heading=h.bsad7uswhrt.

mcoker (changes requested):
ran the visual regression tool and found a few things that need to be updated:
* [`.fa-fw` still references `div()`](https://github.com/patternfly/patternfly/blob/666bee3d38cd562c5588428a2b2659692490a706/src/patternfly/assets/fontawesome/_variables.scss#L12), so that class isn't working. That can probably just use `calc()` like the [update you made to the FA `-stack` class](https://github.com/patternfly/patternfly/blob/666bee3d38cd562c5588428a2b2659692490a706/src/patternfly/assets/fontawesome/_stacked.scss#L11)
* The whole `base/tokens` dir stylesheets are being included in patternfly-base, and that contains the chart tokens. For the time being anyways, we don't want to include the chart styles in PF by default. That could change, but let's stick with the way it's setup now and there are [no chart global vars in patternfly.css](https://unpkg.com/browse/@patternfly/patternfly@latest/patternfly.css)
* `_variables.scss/patternfly-variables.scss` was removed, and I think it should be added back for a couple of reasons
  * We still have global vars that are not tokens and that would be a great place to include them. It's the 2 blocks that reference `pf-v6-set-inverse()` from that removed file - we'll need those back. They're a var that is `1` in LTR and `-1` in RTL that we use with things like `translate()` on the x-axis to use a negative value in RTL since `translate()` doesn't have logical position support.
    * We have an RTL var that was removed in this update (that was in `_variables.scss`) that is breaking some RTL styles. 
  * React-tokens (and maybe other react scripts) parse this stylesheet and unless there is a strong business case to remove this file, I'd like to remove the barrier for needing devs to figure out that part of react-tokens before being able to bump to this version of core.
* [patternfly-no-globals.scss](https://github.com/patternfly/patternfly/blob/666bee3d38cd562c5588428a2b2659692490a706/src/patternfly/patternfly-no-globals.scss) is empty. 

FWIW, these are the updates I made locally to get the visual regression job to pass.

```diff
diff --git a/src/patternfly/assets/fontawesome/_variables.scss b/src/patternfly/assets/fontawesome/_variables.scss
index 020d11782..0b94bf439 100644
--- a/src/patternfly/assets/fontawesome/_variables.scss
+++ b/src/patternfly/assets/fontawesome/_variables.scss
@@ -9,7 +9,7 @@ $fa-version:                  "5.6.3" !default;
 $fa-border-color:             #eee !default;
 $fa-inverse:                  #fff !default;
 $fa-li-width:                 2em !default;
-$fa-fw-width:                 div(20em, 16);
+$fa-fw-width:                 calc(20em / 16);
 
 // Convenience function used to set content property
 @function fa-content($fa-var) {
diff --git a/src/patternfly/base/_index.scss b/src/patternfly/base/_index.scss
index 6d231e9e1..f071145ef 100644
--- a/src/patternfly/base/_index.scss
+++ b/src/patternfly/base/_index.scss
@@ -7,4 +7,4 @@
 @forward './patternfly-fa-icons';
 @forward './patternfly-pf-icons';
 @forward './patternfly-svg-icons';
-@forward './tokens';
+@forward './patternfly-variables';
diff --git a/src/patternfly/patternfly-no-globals.scss b/src/patternfly/patternfly-no-globals.scss
index 9a20d1d97..5cbccdee4 100644
--- a/src/patternfly/patternfly-no-globals.scss
+++ b/src/patternfly/patternfly-no-globals.scss
@@ -2,3 +2,4 @@
   $pf-v6-global--enable-reset: false,
   $pf-v6-global--enable-normalize: false
 );
+@use './patternfly' as *;
```

With the contents of `patternfly-variables.scss` of:

```scss
@use '../sass-utilities' as *;
@use "./tokens/workspace-overrides" as *;
@use "./tokens/tokens-font" as *;
@use "./tokens/tokens-palette" as *;
@use "./tokens/tokens-default" as *;
@use "./tokens/tokens-dark" as *;

:where(:root) {
  @include pf-v6-set-inverse(false);
}

@include pf-v6-rtl {
  @include pf-v6-set-inverse;
}
```

Issue #6515, opened 2024-04-05
SPIKE Animation - revisit and validate approach


srambach:
**Describe the feature**
Review previous research on handling animation with respect for the user's motion preferences.
Determine a structure for handling the `prefers-reduced-motion` media query on all animations, possibly including a variable-based "kill switch" to turn off all animations completely (useful for automated testing). This may either be the method shown in the codepen below or another scheme.

**Any other information?**
[Best practices doc](https://docs.google.com/document/d/1U_wEEKLGznFe1HR6zEJbTrx3-j2bA0jj-I7GWihifpU/edit)
[Codepen demonstrating using css variables to control animation](https://codepen.io/srambach/pen/GRGJyoL?editors=1100)

srambach:
Closing issue, as work to implement will continue in #6516

Issue #6369, opened 2024-02-29
Bug - CSS, Table (and probably much more of PatternFly) - performance issues due to unspecific key CSS selectors (those greedy ones at the rightmost side of selectors, especially `*`)


garrett:
**Describe the problem**

PatternFly's CSS has performance problems, especially with table widget CSS included on a page. Having a thousand rows takes _several_ seconds due to CSS selection. This affects all tables to some degree, especially with page complexity (inside and outside of the table too).

Background: We're currently rebuilding a file manager for Cockpit. (One exists, but it has been nearly abandoned. So we've been making a replacement that's more modern, uses PatternFly instead of just looking like it is using PatternFly, uses React instead of jQuery, etc.) Our initial version was done by a Google Summer of Code student, and some of the core team are looking at improving it now.

We noticed we are hitting all kinds of performance issues. I looked at it and found several problems. We're addressing all those. One in particular is with the table component's CSS.

I made an HTML mockup based on the table HTML from PatternFly's example and intended to have it switch back and forth between an icon view and a details view. But when I only had a table listing a thousand elements and added PatternFly CSS, it took several seconds to render the details view instead of being instant.

I looked into the table CSS and noticed that it's using a `*` glob operator for CSS instead of specifying classes... and it does this multiple times. When you have a table that can grow up to a thousand or more elements, it all adds up.

Here's a snippet from what I saw when peeking in Firefox's dev tools:

![image](https://github.com/patternfly/patternfly/assets/10246/3cad09e5-d3ff-43ef-b306-f4a40c04b55a)

 You can see it uses `> *` in a few places (as well as `:first-child`, which isn't nearly as bad, but still not ideal). As it's [children of the `tr`, those can only be `th` and `td` according to `tr`'s content model](https://html.spec.whatwg.org/multipage/tables.html#the-tr-element) (except for non-visible elements like `script` and `template`, which CSS doesn't need to touch). Even `:where(th, td)` should be much better. This is because [CSS matches from the right to the left](https://stackoverflow.com/questions/5797014/why-do-browsers-match-css-selectors-from-right-to-left/5813672#5813672), so those lines are saying "look at every element on the page (including children and parents of the wanted elements as well as everything else), then see if it has a parent of a th" instead of first narrowing it down to `td` and `th` elements.

An old, but still relevant blog post on the topic:
https://www.stevesouders.com/blog/2009/06/18/simplifying-css-selectors/

[...]

PR #6542, opened 2024-04-16
feat(button): sample animation on button and alert group

srambach:
As part of #6515 spike, this shows a few things:
- motion tokens added. These exist in Figma but were copied over manually for now
- also adds and uses a test duration of 1s so you can see what's going on
- Button has a transition so on hover you can see the color slowly change (1s). This transition has no reduced motion alternative.
- Alert group item has a new class `.pf-m-remove`. This class causes the  item to be removed by sliding off to the right (translateX), or for reduced-motion, it fades (opacity).
- All transitions are removed when the class `.pf-m-no-motion` is applied to a parent. (no matter what the reduced motion setting is) This is intended to be used for testing purposes.

To test:
- Access the prefers-reduced-motion option through Settings -> Accessibility -> Display -> Reduce motion
- Add and remove `.pf-m-remove` on the alert group item
- Add and remove `.pf-m-no-motion` on the body or other parent tag

### DIFF
--- a/src/patternfly/base/_variables.scss
+++ b/src/patternfly/base/_variables.scss
@@ -12,3 +12,35 @@
 @include pf-v6-rtl {
   @include pf-v6-set-inverse;
 }
+
+// stylelint-disable
+.pf-m-no-motion * {
+  // simplest way of turning off all animation
+  // disadvantage: takes the wildcard selector
+  // transition: none !important;
+  // animation: none !important;
+}
+// styleline-enable
+
+.pf-m-no-motion {
+  // instead we can set all (or just semantic?) motion tokens to 0
+  // disadvantage - need to catch any new tokens created
+  // note - relies on always using tokens
+  --pf-t--global--delay--400: 0s;
+  --pf-t--global--delay--300: 0s;
+  --pf-t--global--delay--200: 0s;
+  --pf-t--global--delay--100: 0s;
+  --pf-t--global--duration--300: 0s;
+  --pf-t--global--duration--200: 0s;
+  --pf-t--global--duration--100: 0s;
+  --pf-t--global--duration--000: 0s;
+  --pf-t--global--motion--delay--long: 0s;
+  --pf-t--global--motion--delay--default: 0s;
+  --pf-t--global--motion--delay--short: 0s;
+  --pf-t--global--motion--delay--none: 0s;
+  --pf-t--global--motion--duration--long: 0s;
+  --pf-t--global--motion--duration--default: 0s;
+  --pf-t--global--motion--duration--short: 0s;
+  // added just for testing purposes
+  --pf-t--global--motion--duration--test: 0s;
+}
--- a/src/patternfly/base/tokens/_tokens-default.scss
+++ b/src/patternfly/base/tokens/_tokens-default.scss
@@ -4,6 +4,25 @@
 //  */
 
 :root {
+  // motion tokens manually added for testing
+  --pf-t--global--delay--400: 7000ms;
+  --pf-t--global--delay--300: 100ms;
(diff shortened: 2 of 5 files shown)


andrew-ronaldson:
https://github.com/patternfly/patternfly/assets/28658548/0317f9ea-5b63-4b55-98b3-fe8272333eb3

thatblindgeye (commented):
This is looking pretty nice ✨ 

Some questions/things I noticed, which if it isn't applicable right now/if any of this is already known feel free to ignore:

- Noticing for the Alert, the no-motion modifier class doesn't cause the fade in/out effect that enabling the "prefers reduced motion" setting in the OS.
- For buttons, is it intended that borders would have the same effect on hover as backgrounds? Secondary and tertiary don't look to have any sort of effect on hover (other than the abrupt change in color)
- is the no-motion class intended purely for testing on our part, or do we intend for consumers to be able to utilize it as well so that users could manually enable/disable motion outside of an OS setting?

srambach:
@thatblindgeye I'm finally getting around to revisiting this!

> * Noticing for the Alert, the no-motion modifier class doesn't cause the fade in/out effect that enabling the "prefers reduced motion" setting in the OS.
This is by design - that modifier is intended to kill all motion/animation/transitions so that automated testing won't be confused. In contrast, yes, prefers-reduced-motion might still have some things happening like a fade instead of a slide.

> * For buttons, is it intended that borders would have the same effect on hover as backgrounds? Secondary and tertiary don't look to have any sort of effect on hover (other than the abrupt change in color)
There wasn't a final design - this was just noodling around.

> * is the no-motion class intended purely for testing on our part, or do we intend for consumers to be able to utilize it as well so that users could manually enable/disable motion outside of an OS setting?
TBD? Maybe it's useful for their automated testing but probably not as a user setting.

src/patternfly/base/_variables.scss
```diff
@@ -12,3 +12,35 @@
...
+  --pf-t--global--motion--delay--long: 0s;
+  --pf-t--global--motion--delay--default: 0s;
+  --pf-t--global--motion--delay--short: 0s;
+  --pf-t--global--motion--delay--none: 0s;
+  --pf-t--global--motion--duration--long: 0s;
+  --pf-t--global--motion--duration--default: 0s;
+  --pf-t--global--motion--duration--short: 0s;
+  // added just for testing purposes
+  --pf-t--global--motion--duration--test: 0s;
+}
```
srambach, line 46: @mattnolting @mcoker I'd be interested in your thoughts about the two methods of turning off all motion here. (this isn't reduced motion - it's turning everything off for automated testing purposes)

mattnolting: In [#6369](https://github.com/patternfly/patternfly/issues/6369#issue-2160813742), Garrett notes performance impacts of using global selectors. At the least, I'd say `> *` is nonviable. Second approach 👍 

> You can see it uses > * in a few places (as well as :first-child, which isn't nearly as bad, but still not ideal). As it's [children of the tr, those can only be th and td according to tr's content model](https://html.spec.whatwg.org/multipage/tables.html#the-tr-element) (except for non-visible elements like script and template, which CSS doesn't need to touch). Even :where(th, td) should be much better. This is because [CSS matches from the right to the left](https://stackoverflow.com/questions/5797014/why-do-browsers-match-css-selectors-from-right-to-left/5813672#5813672), so those lines are saying "look at every element on the page (including children and parents of the wanted elements as well as everything else), then see if it has a parent of a th" instead of first narrowing it down to td and th elements.

mcoker:
@srambach do you want to keep this open or is it ok to close?

PR #6543, opened 2024-04-16
chore(Badge/MenuToggle): updated styling

### LINKED ISSUE #6463: Penta Badge/Menu Toggle - update styling when only badge is passed to Menu toggle
Followup to https://github.com/patternfly/patternfly/pull/6430, which contains followup change to be made/looked into

**Any other information?**

thatblindgeye:
Closes #6463 

Currently I'm applying some additional styling to a `pf-m-small` MenuToggle rather than only when it's inside a breadcrumb dropdown. I can update to only change that styling when the MenuToggle is inside the breadcrumb dropdown, but was curious if it would make sense to adjust other styling for a `pf-m-small` Menu Toggle rather than only the top/bottom padding.

Also note that the `pf-m-text` modifier on MenuToggle no longer actually does anything, but we've kept it on the React component.

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -26,12 +26,12 @@
 
   // breadcrumb dropdown
   --#{$breadcrumb}__menu-toggle--MarginBlockStart: calc(var(--pf-t--global--spacer--control--vertical--default) * -1);
-  --#{$breadcrumb}__menu-toggle--MarginInlineEnd: calc(var(--#{$breadcrumb}__item--MarginInlineEnd) * -1);
+  --#{$breadcrumb}__menu-toggle--MarginInlineEnd: var(--pf-t--global--spacer--xs);
   --#{$breadcrumb}__menu-toggle--MarginBlockEnd: calc(var(--pf-t--global--spacer--control--vertical--default) * -1);
-  --#{$breadcrumb}__menu-toggle--MarginInlineStart: calc(var(--#{$breadcrumb}__item-divider--MarginInlineEnd) * -1);
+  --#{$breadcrumb}__menu-toggle--MarginInlineStart: var(--pf-t--global--spacer--xs);
   --#{$breadcrumb}__dropdown--c-menu-toggle--PaddingBlockStart: 0;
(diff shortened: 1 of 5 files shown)


andrew-ronaldson:
The menu toggle example looks good here
Screenshot 2024-04-30 at 12 52 48 PM

The breadcrumb example crops the hover/focus states to fit
Screenshot 2024-04-30 at 12 53 44 PM 

is this caused by the breadcrumbs height?

thatblindgeye:
@andrew-ronaldson that's from setting the padding top/bottom on the MenuToggle in that particular instance to 0 to try and get it closer to being contained in the outer breadcrumb element. I can remove those style overrides which would result in the following:

image
image

thatblindgeye:
@andrew-ronaldson Does this look more inline with what we'd want?

image

I updated the `--#{$breadcrumb}__dropdown--MarginInline[Start|End]` var to an xs spacer, instead of it creating a negative margin based on `calc(var(--#{$breadcrumb}__item--MarginInlineEnd) * -1)`

andrew-ronaldson:
Looks good with that update @thatblindgeye

andrew-ronaldson (approved):
looks dope!

srambach (approved):
🏅 Looks good to me.

src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -116,6 +118,8 @@
   margin-inline-end: var(--#{$breadcrumb}__menu-toggle--MarginInlineEnd);
 
   .#{$menu-toggle} {
+    padding-block-start: var(--#{$breadcrumb}__menu-toggle--PaddingBlockStart);
+    padding-block-end: var(--#{$breadcrumb}__menu-toggle--PaddingBlockEnd);
     line-height: var(--#{$breadcrumb}__menu-toggle--c-menu-toggle--LineHeight);
   }
```
mcoker, line 124: discussed in the working section that we can keep the padding as long as it it doesn't change the shape of the outer breadcrumb box. also looks like `line-height` is the same with or without this declaration, so I think we can nix this whole block!

mcoker (changes requested):
Just one small nit!

mcoker (approved):

Issue #6367, opened 2024-02-28
Log viewer - apply tokens/Penta


srambach:
Design issue is https://github.com/patternfly/patternfly-design-kit/issues/699

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6578

PR #6578, opened 2024-04-26
fix(log-viewer): add tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6367

Updates the log viewer styling. The menus at the top should be updated in https://github.com/patternfly/patternfly/pull/6566

One thing I didn't see called out in the design was the "matched" highlight. Is the text color in light theme OK? Highlight text is just using the default text color.

Screenshot 2024-04-25 at 8 42 07 PM

The highlight color with text in dark theme isn't accessible though. I see there is a mix-blend-mode in figma for the highlight color. WDYT @lboehling? Should I make an update to the CSS? I'm wondering if we should have text color tokens for use with the highlight background colors?

Screenshot 2024-04-25 at 8 41 54 PM

### DIFF
--- a/src/patternfly/components/LogViewer/log-viewer.scss
+++ b/src/patternfly/components/LogViewer/log-viewer.scss
@@ -6,56 +6,56 @@
   --#{$log-viewer}--m-line-numbers__index--Display: inline;
 
   // Header
-  --#{$log-viewer}__header--MarginBottom: var(--#{$pf-global}--spacer--sm);
+  --#{$log-viewer}__header--MarginBottom: var(--pf-t--global--spacer--sm);
 
   // Main
-  --#{$log-viewer}__main--BackgroundColor: var(--#{$pf-global}--BackgroundColor--100);
-  --#{$log-viewer}__main--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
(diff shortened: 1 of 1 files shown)


srambach (approved):
Changes look good. I agree there should be a text-on-highlight token. I'm fine with merging and doing a follow up issue for that.

thatblindgeye (commented):
Agreed on the above re: button spacing and highlight color.

Additionally, do we still need the "Dark" example, pf-m-dark styling in the scss, and documentation on the pf-m-dark class?

mcoker:
@andrew-ronaldson @lboehling I went ahead and merged this since updating the global vars to tokens is one of the things currently blocking react from pulling in the latest core - https://github.com/patternfly/patternfly/issues/6587. I'll open a follow-up for any remaining updates.

> Also, I think the icon buttons on the right need an xs spacer so the buttons aren't touching. It is more noticeable with the hover states.

This may be known, but wanted to mention that there is no space between the actions in the figma design, and that spacing comes from the ["icon button" toolbar group](https://staging-v6.patternfly.org/components/toolbar/html#group-types) which has no spacing between the actions. Still want me to update it?

Also to @thatblindgeye's question, should we keep the dark variation?

PR #6588, opened 2024-04-30
feat(Page): remove deprecated page header

### LINKED ISSUE #6563: Remove deprecated page header
Depends on https://github.com/patternfly/patternfly-react/issues/10140 - we don't want to remove the styles before it's removed in react, or the react build will fail when it pulls in a new version of core.

----

React is removing the deprecated page header so we can remove any associated styles once that work is complete. Anything `.pf-v-c-page__header` should be fair game  🚀

https://github.com/patternfly/patternfly/blob/6b5615b0be4370a59da25bbdafbfa845700063c3/src/patternfly/components/Page/page.scss#L13-L66

kmcfaul:
Closes #6563.

There was a single variable that looked like it was still in use for masthead, so I left that in. LMK if that block of code is also something we want to remove, or if we should keep it + rename the variable?


src/patternfly/components/Page/page.scss
```diff
@@ -199,119 +152,11 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 }
 
 // Header
-.#{$page}__header,
 .#{$page} > .#{$masthead} {
   z-index: var(--#{$page}__header--ZIndex);
```
kmcfaul, line 156: This header variable I left in because it looks like its used by masthead

mcoker: Thanks for calling that out! At a minimum, we should probably rename that var. Following our var naming scheme, `--#{$page}--c-masthead--ZIndex`. Though I have a related question for anyone interested (@srambach @mattnolting @thatblindgeye @tlabaj).

As a follow-up, WDYT about having a structural element for the page masthead/header area (`.pf-[v]-c-page__header`) that we use as the container for the header area (the masthead would go inside of it). It would:
* Be the element for the header area grid item in page, so page would be:
  * Page (parent grid layout)
    * page__header (holds masthead)
    * page__sidebar (holds navigation)
    * page__main (holds page sections)
* Get the z-index for the header area
* Could be a generic container if a user didn't pass a masthead to it or we offer some other type of element that can go up there

mcoker (approved):
Nice, LGTM! Pushed a couple of other small places I found references to the old page header.

src/patternfly/components/Page/page.scss
thatblindgeye: Off the top of my head I wouldn't be opposed. It would match the structure we're using for the other elements like you outlined above at the very least. If it's strictly a container wrapper for structure it may not cause any/much concern? Only thing I could think of is someone using the wrapper for their header content, but outside of Masthead; basically having content that should be in a `header` element but isn't.

mattnolting (approved):
LPTM! 🚀

src/patternfly/components/Page/page.scss
mcoker: @thatblindgeye good point. I was thinking the masthead could just be a `<div>` by default, and `page__header` would be the `<header>` element, so whatever went in there would be in `<header>` by default.

PR #6593, opened 2024-04-30
feat(all): rename variables to use logical direction names

### LINKED ISSUE #6389: Update var names used for logical properties
In v5, we updated CSS properties with logical property equivalents for padding, margin, borders, insets, etc. To avoid that being a breaking change, we left the CSS var names unchanged, so we have a lot of places in the code where the CSS property and var names do not match. For example https://github.com/patternfly/patternfly/blob/7df2e174cc259386641610cf1dc5bec1364b3f92/src/patternfly/components/Button/button.scss#L213-L216

For v6, let's update these to match. We should consider doing it in a single effort and using screenshots to validate the changes, similar to the workflow used in v5 to switch to logical properties.

srambach:
Fixes #6389

This renames variables with top/right/bottom/left in the name and have been repurposed to apply to inline/block start/end properties (supporting RTL)

Visual regression in https://drive.google.com/drive/u/0/folders/1Wb2OAB7rO_e-17jxLSGLuTtAm7g-WJUA
Note: I reverted all log viewer changes as @coker is going to do that in the log-viewer PR to avoid lots of conflicts.
Remaining visual regression failures are disabled button with a badge (fixed since the baseline screens), text formatting (also fixed since the baseline screens), and a leftover dropdown toggle.

### DIFF
--- a/src/patternfly/components/AboutModalBox/about-modal-box.scss
+++ b/src/patternfly/components/AboutModalBox/about-modal-box.scss
@@ -22,34 +22,34 @@
   );
 
   // Brand
-  --#{$about-modal-box}__brand--PaddingTop: var(--pf-t--global--spacer--2xl);
-  --#{$about-modal-box}__brand--PaddingRight: var(--pf-t--global--spacer--xl);
-  --#{$about-modal-box}__brand--PaddingLeft: var(--pf-t--global--spacer--xl);
-  --#{$about-modal-box}__brand--PaddingBottom: var(--pf-t--global--spacer--xl);
-  --#{$about-modal-box}__brand--sm--PaddingRight: var(--pf-t--global--spacer--3xl);
-  --#{$about-modal-box}__brand--sm--PaddingLeft: var(--pf-t--global--spacer--3xl);
-  --#{$about-modal-box}__brand--sm--PaddingBottom: var(--pf-t--global--spacer--3xl);
+  --#{$about-modal-box}__brand--PaddingBlockStart: var(--pf-t--global--spacer--2xl);
+  --#{$about-modal-box}__brand--PaddingInlineEnd: var(--pf-t--global--spacer--xl);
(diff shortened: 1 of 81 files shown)


mcoker:
@srambach sorry, I forgot to update the log viewer in https://github.com/patternfly/patternfly/pull/6578 🤦‍♂️ So I pushed that change to your PR along with a few other tweaks after reviewing. Here are the commits:

https://github.com/patternfly/patternfly/pull/6593/commits/35e3b2546c94dac37f617b669bcf15e9b97bc496
* Updates log viewer

https://github.com/patternfly/patternfly/pull/6593/commits/2197e02ee48c6edf839e290b3fbd9a2d8106149e
* A typo in app launcher - doesn't matter since we're deprecating app launcher, but it threw off the next commit which is how I noticed it

https://github.com/patternfly/patternfly/pull/6593/commits/026f8b330ffdfa95b1aabe525dd1710528b2c38f
* Looks like the `inset-[...]` property vars were missing the `Inset[...]` prefix, so added that

https://github.com/patternfly/patternfly/pull/6593/commits/c4fe2f19ebf9441fa0801497729173f0cc4ab44b
* A few random logical property updates I spotted along the way. Also adds back a missing var from dual list selector that as far as I can tell looks like a bug. It adds the missing margin before the item-count badge - this screenshot is from v6 currently
 
Screenshot 2024-04-30 at 9 55 47 PM

And after this update

Screenshot 2024-04-30 at 10 14 43 PM

srambach:
Thanks @mcoker. Progress had one tiny mistake, which I fixed. New regression tests at https://drive.google.com/file/d/1aPfWvC8Te9P9ecRRS95SDrkt6DDVdSLc/view?usp=sharing which show the button badge, text indentation, log viewer, and one card where it thinks the icons didn't render, but when I check it, they are there. And it did not show up before. So I'm thinking that is a fluke of timing.

mcoker (approved):
NICE!! LGTM 🚀

PR #6598, opened 2024-05-01
chore(tokens): import tokens without :root selector in token files

### LINKED ISSUE #6105: Import tokens without selector in token file
Currently, the tokens are generated into a file that includes the `:root` selector. @mcoker suggests we remove that so that those tokens could be imported into different contexts, not always in the `:root` context.

So [once that is done](https://github.com/patternfly/design-tokens/issues/9), the change here would be to move the import of those tokens to within the correct selectors - in base/_variables.scss we'd import within `:root` for the default, palette, and font tokens, and within ` :root:where(.pf-v5-theme-dark)` for the dark tokens.

wise-king-sullyman:
Closes #6105


src/patternfly/base/patternfly-variables.scss
```diff
@@ -2,11 +2,16 @@
...
-@use "./tokens/tokens-dark" as *;
+@use "./tokens/tokens-default" as default;
+@use "./tokens/tokens-dark" as dark;
 
 :where(:root) {
   @include pf-v6-set-inverse(false);
+  @include default.pf-v6-tokens;
+}
+
+:root:where(.pf-v6-theme-dark) {
```
mcoker, line 13: I wonder if this selector would work?

```suggestion
:where(.pf-v6-theme-dark) {
```

mcoker (commented):
Just so we have a record of it, as discussed offline:
* We can likely drop `@use "./tokens/workspace-overrides" as *;`
* It would be great to follow this same schema on the custom token file we have at `@use "./tokens/tokens-font" as *;` (wrap it in a mixin, `@include` it here)
  * Just a side note, we should probably rename that file 😅 Maybe `tokens-local` or `tokens-custom`? wdyt @srambach 
* Update `tokens-pallette` in the same way we did with tokens-default/dark
* Just ensure we're importing everything in the same order it was imported previously

mcoker (approved):
FABULOUS!! Thank you @wise-king-sullyman! 🫵🪨

PR #6600, opened 2024-05-01
fix(docs): add tabindex for scrollable examples

### LINKED ISSUE #6599: Bug - scrollable examples missing tabindex
Follow-up to website a11y report findings logged in https://github.com/patternfly/patternfly-org/issues/3979

Examples in the following Core component pages have scrollable areas that cannot be accessed with keyboard; React examples manually add `tabindex=0` and we should do the same in Core:

- [Panel](https://www.patternfly.org/components/panel#scrollable)
- [Sidebar](https://www.patternfly.org/components/sidebar#sticky-panel)
- [Table](https://www.patternfly.org/components/table/html#sticky-header)

evwilkin:
Closes #6599 

This PR fixes the a11y errors thrown by examples with scrollable content that wasn't focusable via keyboard:

- Panel:  adds `tabindex` manually to two scrollable examples to match React examples.

- Sidebar: 
  - Change example CSS from only `height: 200px` for all examples to `max-height: 200px` only for scrollable examples for consistency with React examples (is there a better way to do this other than example IDs?)
  - Set `tabindex` to these scrollable examples

- Table: use scrollable table template for sticky header table in place of one-off solution, this both fixes the focus issue and better aligns with the React examples.


mcoker (approved):
🥳

PR #6607, opened 2024-05-02
chore(Pagination): refactored to use MenuToggle instead of OptionsMenu

### LINKED ISSUE #6094: Pagination - refactor to use menu toggle/menu instead of options menu
Followup to https://github.com/patternfly/patternfly/pull/6089 

Currently core is still using options menu components for the `pagination-options-menu.hbs` file. We need to refactor it to use menutoggle/menu, similar to React

**Any other information?**

thatblindgeye:
Closes #6094

[Pagination preview](https://patternfly-pr-6607.surge.sh/components/pagination)


src/patternfly/components/Pagination/examples/Pagination.md
```diff
@@ -131,13 +120,12 @@ import './Pagination.css'
...
 {{#> pagination pagination--id="pagination-inset" pagination--modifier="pf-m-inset-none pf-m-inset-md-on-md pf-m-inset-2xl-on-lg"}}
   {{> pagination-total-items-content}}
-  {{> pagination-options-menu pagination-options-menu--id=(concat pagination--id '-options-menu')}}
+  {{> pagination-menu-toggle pagination-menu-toggle--id=(concat pagination--id '-menu-toggle')}}
   {{> pagination-nav-content pagination-nav--aria-label="Pagination nav - inset example"}}
 {{/pagination}}
 ```
 
 ## Documentation
-Note: `<button>` or `<a>` elements can be used in `.pf-v6-c-pagination__nav-page-select`.
```
thatblindgeye, line 140: This wasn't accurate, as the element refers to the number input field for page select. We also don't allow customizing the elements for buttons used elsewhere in the component in React (menu toggle should be a button and the next/previous controls should be buttons)

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -104,13 +107,12 @@ $pf-v6-c-pagination--variable-map: build-spacer-map("none", "sm", "md", "lg", "x
...
     .#{$pagination}__page-menu {
       position: absolute;
-      display: block;
+      display: var(--#{$pagination}--m-bottom__page-menu--Display);
     }
 
     .#{$pagination}__nav {
       display: flex;
       flex-basis: 100%;
-      justify-content: space-between;
```
thatblindgeye, line 113: This was causing an issue where visual focus wasn't matching with tab order. The following screenshot makes it seems like pressing Tab would make the "previous" button on the left be focused first, then the menu toggle, but the menu toggle receives focus first then the "previous" button does. I can update this styling further (do we want it to look more like compact pagination where everything is pushed to the right, or do we want all controls centered?), but just wanted to remove this for now to at least fix the issue.

Bottom Pagination at breakpoint, showing a visual order that doesn't match with the actual tab order

kaylachumley (approved):
looks good! nice job

src/patternfly/components/Pagination/pagination.scss
mcoker: Hmm.... I don't see a specific design in figma for bottom pagination on mobile - did you talk with design about this change? If not, IMO this is probably better as a separate issue to address the a11y of the tab order and update the design/layout.

For reference, this is what it looks like now
Screenshot 2024-05-22 at 5 31 22 PM

If we go with something like this, it's probably worth looking at the CSS some more to see if we can not use absolute positioning for the menu since the way that's written currently is specific to the old layout.

mcoker (changes requested):
LGTM, just a few nits, and larger comment around the layout change for bottom/mobile pagination. Just want to make sure design approves, and if there is more work to be done, probably doing that in a separate issue.

src/patternfly/components/Pagination/pagination.scss
thatblindgeye: @kaylachumley @lboehling wdyt? My screenshot is the existing design, Michael's screenshot is what it looks like with my updates in this PR. If we'd want something different from what is in this PR I can revert and open a followup to tackle this.

lboehling: I think the first example is a better interaction for mobile use -- separating the front and back arrows more is better for touch interaction/a more typical layout for mobile!

mcoker: @thatblindgeye other than this comment, the PR looks good to merge 👍

thatblindgeye: Opened https://github.com/patternfly/patternfly/issues/6731 as a followup

lboehling (approved):
rad!

srambach (approved):
📃 LGTM

mcoker (approved):
🏆

PR #6608, opened 2024-05-02
feat(Wizard): set flexbasis with height var

### LINKED ISSUE #6457: Wizard - setting height does not work when inside of a modal
Re: https://github.com/patternfly/patternfly-react/issues/10144

React has a `height` prop for `<Wizard>` that applies an inline style of `style="height: [value of height prop]"` to control the height. This works fine, except for when the wizard is the direct child of a modal. In a modal, the wizard height is managed by `flex-basis` https://github.com/patternfly/patternfly/blob/c1e50c4c674df83fd730088d8d73fea85e61d339/src/patternfly/components/Wizard/wizard.scss#L216-L220

A proposal as a way to fix this bug and make it easier for the user to manage the wizard height is to:
* Update the react component to set `--#{$wizard}--Height: [value of height prop]` instead
* Update the wizard in core to set both `height` and `flex-basis` (when in a modal) via `--#{$wizard}--Height`

That should allow the `height` prop in react to manage the wizard height both inside and outside of a modal.

kmcfaul:
Closes #6457

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -1,7 +1,7 @@
 @use '../../sass-utilities' as *;
 
 :where(:root, .#{$wizard}) {
-  --#{$wizard}--height: initial;
+  --#{$wizard}--Height: initial;
   --#{$wizard}--Height--base: 100%;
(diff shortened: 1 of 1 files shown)


mcoker (approved):
LGTM!

thatblindgeye (approved):

PR #6611, opened 2024-05-02
chore(Wizard): resolved scroll issue and used ActionList for wizard buttons

### LINKED ISSUE #6314: Wizard follow ups
From https://github.com/patternfly/patternfly/pull/6251#pullrequestreview-1879050606

* in page wizards - only the wizard content (nav and main) should scroll. Has to do with `align-self: start` being added to `$page__main`.
* Use the action list in the footer - right now the actions in the footer have margins, including a margin-bottom, so the bottom padding of the footer takes into account the action margin, which we shouldn't need to do?

thatblindgeye:
Closes #6314

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -314,7 +314,8 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 
 // Main & Drawer
 .#{$page}__main-container {
-  align-self: start;
+  display: flex;
+  flex-direction: column;
   max-height: var(--#{$page}__main-container--MaxHeight);
   margin-block-start: 0;
   margin-inline-start: var(--#{$page}__main-container--MarginInlineStart);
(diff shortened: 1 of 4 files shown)


mcoker (changes requested):
Nice! I think we can now update the footer just to be a regular box since the action list provides the button layout we need. We can take off all the flex stuff and child styling. Another thing to look for is the footer's bottom padding currently accounts for the bottom margin we're adding to children (buttons) in the footer, so that padding will likely need to be updated, too.

https://github.com/patternfly/patternfly/blob/c3f6c8175a7de2e643ec02102d32520b298e06ce/src/patternfly/components/Wizard/wizard.scss#L513-L533

thatblindgeye:
@mcoker Updated, but just to note In React, a consumer would still be able to create a custom footer that doesn't use ActionList (see [custom footer example](https://www.patternfly.org/components/wizard#custom-footer)). In those scenarios, I assume we'd just recommend either using Action List or utilizing our flex layouts?

mcoker (approved):
🥳🍕

PR #6614, opened 2024-05-03
fix(MenuToggle): updated borderradius and filter padding

### LINKED ISSUE #6536: Menu toggle bugs
bugs found in https://github.com/patternfly/patternfly-react/issues/10011
* https://github.com/patternfly/patternfly-react/issues/10011#issuecomment-2045942434
* https://github.com/patternfly/patternfly-react/issues/10011#issuecomment-2045959999

thatblindgeye:
Closes #6536

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -6,11 +6,10 @@
   --#{$text-input-group}--BorderWidth: var(--pf-t--global--border--width--control--default);
   --#{$text-input-group}--BorderRadius--base: var(--pf-t--global--border--radius--small);
 
-  // Actual logical properties don't exist for border-radius, but the naming fits the intent
-  --#{$text-input-group}--BorderBlockStartInlineStartRadius: var(--#{$text-input-group}--BorderRadius--base);
-  --#{$text-input-group}--BorderBlockStartInlineEndRadius: var(--#{$text-input-group}--BorderRadius--base);
-  --#{$text-input-group}--BorderBlockEndInlineEndRadius: var(--#{$text-input-group}--BorderRadius--base);
-  --#{$text-input-group}--BorderBlockEndInlineStartRadius: var(--#{$text-input-group}--BorderRadius--base);
+  --#{$text-input-group}--BorderStartStartRadius: var(--#{$text-input-group}--BorderRadius--base);
+  --#{$text-input-group}--BorderStartEndRadius: var(--#{$text-input-group}--BorderRadius--base);
+  --#{$text-input-group}--BorderEndEndRadius: var(--#{$text-input-group}--BorderRadius--base);
+  --#{$text-input-group}--BorderEndStartRadius: var(--#{$text-input-group}--BorderRadius--base);
 
   // Border hover
   --#{$text-input-group}--m-hover--BorderColor: var(--pf-t--global--border--color--hover);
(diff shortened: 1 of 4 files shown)


thatblindgeye:
@mcoker @andrew-ronaldson This doesn't include any update for the "double border" comment mentioned in the original React issue. I wasn't sure what the consensus was on that. We could remove the native focus `outline` on the text input group's text input when it's inside of a typeahead, expanded MenuToggle, which should work with how this sort of variant is setup in React for typeaheads (the input itself never loses actual focus). Something along the lines of this:


https://github.com/patternfly/patternfly/assets/70952936/8b521066-c235-4f5b-a00a-10aa23f239e9

andrew-ronaldson (approved):
I think this is looking good

mcoker:
Discussed offline, the text input group borders were changed in a way that the `.pf-m-plain` modifier is no longer working as expected. Also noticed that there is no border change on focus. A proposal for an update that could address this is:
* Move the border from `.pf-[v]-text-input-group::before` back to `.pf-[v]-text-input-group__text::before`
* Remove any references to `.pf-[v]-text-input-group__text::after` (no longer needed)

2 more nice-to-haves for text input group would be:
* Add an example for the `.pf-m-plain` variation
* Focus styles aren't working correctly. I would probably group some focus styles with [the existing `:hover` styles](https://github.com/patternfly/patternfly/blob/07530abda9d48d533dbe09e2e9dde16f160c11d3/src/patternfly/components/TextInputGroup/text-input-group.scss#L66-L68)
  * Something that mimics the v5 styling, maybe something like this?
    ```
    @at-root .#{$text-input-group}__text:focus-within,
    &:hover {
      --#{$text-input-group}--BorderColor: var(--#{$text-input-group}--m-hover--BorderColor);
     }
    ```

thatblindgeye:
@mcoker addressed comments above, however:

> Focus styles aren't working correctly

If we would want the focus styles to match hover style, we'd probably need to focus-within both the TextInputGroup and MenuToggle (though maybe just for typeahead variant). Using just focus works when the text input element receives focus to change the outer `pf-m-c-text-input-group` border color, but focusing any utilities in a text input group results in the border color not changing still. Same goes for the typeahead menu toggle.

Could be a question for design. If it's any consolation, while it may look odd to not have the outermost border color change, it also may not be necessary since everything else (should) have its own focus styles.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -52,15 +57,16 @@
   width: 100%; // when child of flex layout, take up space
   color: var(--#{$text-input-group}--Color, inherit);
   background-color: var(--#{$text-input-group}--BackgroundColor);
-  border-radius: var(--#{$text-input-group}--BorderRadius);
+  border-start-start-radius: var(--#{$text-input-group}--BorderStartStartRadius);
+  border-start-end-radius: var(--#{$text-input-group}--BorderStartEndRadius);
+  border-end-start-radius: var(--#{$text-input-group}--BorderEndStartRadius);
+  border-end-end-radius: var(--#{$text-input-group}--BorderEndEndRadius);
 
   &::before {
```
mcoker, line 65: Can this be removed now?

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -99,13 +105,13 @@
   }
 
   > :first-child:not(.#{$text-input-group}__text) {
+    margin-block-start: var(--#{$text-input-group}__main--first-child--not--text-input--MarginBlockStart);
+    margin-block-end: var(--#{$text-input-group}__main--first-child--not--text-input--MarginBlockEnd);
```
mcoker, line 109: I think this runs the risk of causing the text input group component to potentially be too tall depending on what that first-child is. The height needs to match other controls/form elements - button, menu toggle, text input, etc. That height is 37px in v6 (it was 36px in v5), and the text input group with the label group is 38px tall in this PR. If you put this text input group in a filter group next to another control (like a button), they'll be different heights.

In v5, we added padding to the chip group main element when it's in text input group, I wonder if that might work here, too? https://www.patternfly.org/components/text-input-group/html#filters

mcoker (commented):
I think we can get rid of `::after` in the selector list here https://github.com/patternfly/patternfly/blob/d523df4e1badea9b9cdf47b2cfa92703f44e9bb2/src/patternfly/components/TextInputGroup/text-input-group.scss#L89

mcoker (approved):
🫵🪨

Issue #6623, opened 2024-05-07
Update v6 dependencies


mcoker:
Update these `6.x-alpha.x` deps

https://github.com/patternfly/patternfly/blob/0c361370f5377616d887c823cce5c748b44e13c2/package.json#L48-L52

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6624

PR #6624, opened 2024-05-07
chore(deps): update pf deps

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6623


srambach (commented):
One thing - if you look at a full page demo and  click the RTL switch, the switches disappear. It looks like it's because the z-index variable `var(--pf-v6-global--ZIndex--2xl)` is now undefined.

Otherwise, I don't see any other issues.

srambach (approved):
🚗

mcoker:
The `z-index` var @srambach mentioned should be fixed once we finish updating docs-framework styles to use tokens

PR #6636, opened 2024-05-09
fix(breakpoints): add css variables for breakpoints

### LINKED ISSUE #6613: Add CSS vars for breakpoints
We removed the CSS vars for breakpoints since we can't use them in `@media` queries, and left only the scss vars. We need to add those back since the react repo uses the CSS vars as a means to get the breakpoint values to do fancy pants scripty things like resize observer, and I imagine products may, too.

https://github.com/patternfly/patternfly-react/blob/029351e0fa402c80a6199e2bd4b28e8c06f17562/packages/react-core/src/components/Page/Page.tsx#L4C8-L4C26

https://github.com/patternfly/patternfly-react/blob/029351e0fa402c80a6199e2bd4b28e8c06f17562/packages/react-core/src/components/Page/Page.tsx#L163-L197

srambach:
This adds back css variables for breakpoints as tokens. They should match the scss variables, but the scss variables are still required for use in media queries. 

Fixes #6613

### DIFF
--- a/src/patternfly/base/tokens/tokens-charts-dark.scss
+++ b/src/patternfly/base/tokens/tokens-charts-dark.scss
@@ -1,6 +1,6 @@
 
 // Do not edit directly
-// Generated on Tue, 07 May 2024 18:55:09 GMT
+// Generated on Thu, 09 May 2024 17:55:04 GMT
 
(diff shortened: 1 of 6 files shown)


mcoker (approved):
🚀

Issue #5949, opened 2023-09-19
Consume tokens in: Charts


nicolethoen:
Design issue: https://github.com/patternfly/patternfly-design/issues/1275

Determine what changes are needed to consume tokens in Charts styles:

https://github.com/patternfly/patternfly/blob/v6/src/patternfly/patternfly-charts.scss
https://github.com/patternfly/patternfly/blob/v6/src/patternfly/patternfly-charts-theme-dark-unversioned.scss
https://github.com/patternfly/patternfly/blob/v6/src/patternfly/patternfly-charts-theme-dark.scss

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6644

PR #6533, opened 2024-04-08
chore(global): remove old global sass vars

mcoker:

PR #6644, opened 2024-05-10
fix(charts): add tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5949

There is an issue with some of the charts tokens not being generated because of their "string" type not supported in the exporter, but that should be fixed soon. This PR only references tokens that are generated, so we'll need a follow up once those tokens are available.

I don't know that there is much to verify for now<sup>[1]</sup> since I don't believe we can pull this change into react yet. Here is the original PR that took out the charts scss/global vars if anyone wants to reference that - https://github.com/patternfly/patternfly/pull/6533/files

1. Just to clarify, this PR and core should be good to pull into react as-is, but I believe pulling into react is currently blocked by https://github.com/patternfly/patternfly-react/pull/10345. So there isn't anything visual we can validate with this PR beyond scanning the code changes.

### DIFF
--- a/src/patternfly/patternfly-charts.scss
+++ b/src/patternfly/patternfly-charts.scss
@@ -11,5 +11,411 @@ $chart: #{$pf-prefix} + 'chart';
 }
 
 :where(:root) {
-  // TODO update with tokens
+  // Chart colors
+  // blue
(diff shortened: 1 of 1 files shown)


srambach (approved):
Looks good on manual inspection 👀

lboehling (approved):
lgtm!! 📊 🪙 🙌

PR #6657, opened 2024-05-14
chore(table): updated clickable expandable example

### LINKED ISSUE #6439: Table - clickable and expanded - `pf-m-selected` needs to be remove, `pf-m-expanded` needs to be remove from all rows
Screenshot 2024-03-19 at 11 21 48 AM

mattnolting:
closes #6439

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -574,7 +574,7 @@
       background-color: var(--#{$table}__tbody--m-selected--BackgroundColor);
       outline-offset: var(--#{$table}__tbody--m-selected--OutlineOffset);
 
-      .#{$table}__tr.pf-m-expanded:not(.pf-v6-c-table__expandable-row) {
+      .#{$table}__tr.pf-m-expanded:not(.#{$table}__expandable-row) {
         border: none;
       }
     }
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -572,6 +573,10 @@
     &.pf-m-selected {
       background-color: var(--#{$table}__tbody--m-selected--BackgroundColor);
       outline-offset: var(--#{$table}__tbody--m-selected--OutlineOffset);
+
+      .#{$table}__tr.pf-m-expanded:not(.pf-v6-c-table__expandable-row) {
+        border: none;
+      }
```
mcoker, line 579: What is this needed for? I don't see any changes when toggling it in dev tools.

```suggestion
      .#{$table}__tr.pf-m-expanded:not(.#{$table}__expandable-row) {
        border: none;
      }
```

mcoker (commented):
1 question about a style - if we keep it, just need to update to use the sass namespace var for table.

src/patternfly/components/Table/table.scss
mattnolting: @mcoker Thank you, that is correct

mcoker (approved):

PR #6675, opened 2024-05-20
feat(button): add animation for hover and clicked

### LINKED ISSUE #6589: Buttons - Animating transitions on all buttons variants.
**Goal**
Adding subtle transition on buttons between default and hover states. 
[Link to Figma annotations](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/branch/YIxfglww1fRdQGclApv1nm/PatternFly-6%3A-Components?type=design&node-id=23613%3A1180&mode=design&t=zeOzDxjxoTCXWQLl-1).

Follow up work could include a Pressed state. Will create another issue when needed.

srambach:
Fixes #6589 

There is no reduced-motion variation for this one.
Figma specs: https://www.figma.com/design/VMEX8Xg2nzhBX8rfBx53jp/branch/YIxfglww1fRdQGclApv1nm/PatternFly-6%3A-Components?m=auto&node-id=23613%3A1181&t=yaDSp7SyRyKeEd2A-1

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -25,7 +25,6 @@
   --#{$button}--hover--BorderWidth: var(--pf-t--global--border--width--button--hover);
   --#{$button}--hover--TextDecoration: none;
   
-
   // Clicked
   --#{$button}--m-clicked--BackgroundColor: transparent;
   --#{$button}--m-clicked--BorderColor: transparent;
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -15,17 +15,24 @@
   --#{$button}--BorderWidth: var(--pf-t--global--border--width--button--default);
   --#{$button}--BorderRadius: var(--pf-t--global--border--radius--pill);
   --#{$button}--TextDecoration: none;
+  --#{$button}--TransitionDuration: var(--pf-t--global--motion--duration--fade--default);
+  --#{$button}--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
 
   // Hover
   --#{$button}--hover--BackgroundColor: transparent;
   --#{$button}--hover--BorderColor: transparent;
   --#{$button}--hover--BorderWidth: var(--pf-t--global--border--width--button--hover);
   --#{$button}--hover--TextDecoration: none;
+  --#{$button}--hover--TransitionDuration: var(--pf-t--global--motion--duration--fade--default);
+  --#{$button}--hover--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
+  
 
   // Clicked
   --#{$button}--m-clicked--BackgroundColor: transparent;
   --#{$button}--m-clicked--BorderColor: transparent;
   --#{$button}--m-clicked--BorderWidth: var(--pf-t--global--border--width--button--clicked);
+  --#{$button}--m-clicked--TransitionDuration: var(--pf-t--global--motion--duration--fade--short);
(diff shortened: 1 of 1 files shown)


srambach:
@andrew-ronaldson should this same animation also be on menu toggles?

andrew-ronaldson (approved):
sweet moves, ace!

src/patternfly/components/Button/button.scss
```diff
@@ -294,6 +301,9 @@
   border-start-end-radius: var(--#{$button}--BorderStartEndRadius, var(--#{$button}--BorderRadius));
   border-end-start-radius: var(--#{$button}--BorderEndStartRadius, var(--#{$button}--BorderRadius));
   border-end-end-radius: var(--#{$button}--BorderEndEndRadius, var(--#{$button}--BorderRadius));
+  transition-timing-function: var(--#{$button}--TransitionTimingFunction);
+  transition-duration: var(--#{$button}--TransitionDuration);
+  transition-property: background-color;
```
mcoker, line 306: Just a nit and you may know this but the shorthand version would be

```
transition: background-color var(--#{$button}--TransitionDuration) var(--#{$button}--TransitionTimingFunction);
```

srambach: Since the property isn't a variable I wasn't sure if the shorthand version was advisable or not? I can change it if you think it's better.

srambach: @mcoker I realized color needed to be transformed as well, so I left this as is since the shorthand doesn't take multiple properties (unless we use `all`)

kaylachumley (approved):
LGTM 💯

mcoker (commented):
LGTM! Left a few nit comments.

Also maybe this question has already been asked or is obvious, but in figma the annotations say _"on button click..."_. Just want to point out that for the button, we have "clicked" styles for the `.pf-m-clicked` variation, but that might be different from just clicking on a button. The `.pf-m-clicked` variation is for something like a button that's used to open a drawer. We'll apply the clicked variation to indicate that the drawer is expanded, and remove it when the drawer is closed. Just plain ol' clicking on a button triggers something called `:active`, which happens when you hold down your mouse button (or whatever pointer device or the enter/space key on the keyboard). Here's a codepen showing an `:active` transition - https://codepen.io/mcoker/pen/WNBROpE. I don't know that it means anything in terms of this PR, since the hover/focus and clicked transitions are all the same, but wanted to make sure we were referring to the same thing.

mcoker:
Oh I meant to leave one more comment. In terms of allowing users to theme/customize the button animations, I think this PR is good, but it looks to assume the following which I just wanted to confirm is 👍👍

* The properties to transition (text color, background color, border color, border width) are not intended to be themed since we don't have a var for `transition-property`.
* The timing function and duration can be themed, but it's a single timing function and a single duration and it applies to all of the properties listed above. For example, we don't intend for background color and border width to have different timing functions or durations.

src/patternfly/components/Button/button.scss
```diff
@@ -303,6 +309,8 @@
     border: var(--#{$button}--BorderWidth) solid;
     border-color: var(--#{$button}--BorderColor);
     border-radius: inherit;
+    transition-timing-function: inherit;
+    transition-duration: inherit;
```
mcoker, line 313: Just to confirm - since the initial value of `transition-property` is "all", leaving that out here will set that. Was that intended? If not, `transition: inherit;` should do the trick.

mcoker (commented):
Looks good! Just one question 👍

src/patternfly/components/Button/button.scss
srambach: eh, not really 😄  I'll fix it.

mcoker (approved):
🚀

PR #6678, opened 2024-05-21
fix(Toolbar): Added hidden class to content element

### LINKED ISSUE #6627: Toolbar - hidden class not working on content element
React relies on `.pf-[v]-c-toolbar__content.pf-m-hidden` https://github.com/patternfly/patternfly-react/issues/10342

Looks like that was removed in https://github.com/patternfly/patternfly/pull/6057

More info below

----

Specifically that class came from the use of `@include pf-v5-hidden-visible` on `.pf-[v]-c-toolbar__content`. There were a bunch of elements in the toolbar that had `@include pf-v5-hidden-visible` in the element's styles that added `.pf-m-hidden` support to those elements, and that was rewritten to group/combine all of the element selectors that needed it and include `@include pf-v5-hidden-visible` once in the grouped selector - except `.pf-[v]-c-toolbar__content` was excluded from the group. Below are screenshots of that change from https://github.com/patternfly/patternfly/pull/6057

`@include pf-v5-hidden-visible` removed from `.pf-[v]-c-toolbar__content` (and the rest of the individual toolbar elements that included it)
Screenshot 2024-05-08 at 12 15 43 PM

`@include pf-v5-hidden-visible` was added back to a group selector that doesn't include `.pf-[v]-c-toolbar__content`
Screenshot 2024-05-08 at 12 16 28 PM

One way of addressing this could be to add `.pf-[v]-c-toolbar__content` to the grouped selector, but it's worth noting that the new grouped selector also includes `flex-wrap: wrap`, so we would want to ensure that's a safe rule to add to `.pf-[v]-c-toolbar__content` before adding it https://github.com/patternfly/patternfly/blob/0747e621c7680b13dff50d86f7db22f6c35edda4/src/patternfly/components/Toolbar/toolbar.scss#L66-L75

tlabaj:
closes: #6627


src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -63,7 +63,8 @@ $pf-v6--include-toolbar-breakpoints: true !default;
   --#{$toolbar}__item--m-label--FontWeight: var(--pf-t--global--font--weight--body--bold);
 }
 
-// - Toolbar content - Toolbar group - Toolbar item - Toolbar group label container - Toolbar group label group
+// - Toolbar content - Toolbar content section - Toolbar group - Toolbar item - Toolbar group label container - Toolbar group label group
+.#{$toolbar}__content,
```
mcoker, line 67: `pf-v6-hidden-visible()` generates `.pf-m-hidden` and `.pf-m-visible` classes for selectors its used with, and takes an argument for the value of the `display` property used to show the selector(s) that is used with `.pf-m-visible`.  Here's some fake code to show what I mean:

```
.pf-c-component,
.pf-c-another-component {
  @include pf-v6-hidden-visible(flex);
}

// compiles to this CSS
.pf-c-component.pf-m-hidden,
.pf-c-another-component.pf-m-hidden {
  display: none;
}

.pf-c-component.pf-m-visible,
.pf-c-another-component.pf-m-visible {
  display: flex; // "flex" is the argument passed to pf-v6-hidden-visible() above
}
```

All of the selectors in this block where `.#{$toolbar}__content` was added are flex layouts, so they can all be grouped this way. However, `.#{$toolbar}__content` is a grid layout, so it can't be used in this group. We can see that it is `display: grid` here https://github.com/patternfly/patternfly/blob/9f6d47c2a203ebd5e3867516e9f4f84fd77e859a/src/patternfly/components/Toolbar/toolbar.scss#L79-L83

So in short we just need to create a new block for `.#{$toolbar}__content` to use `pf-v6-hidden-visible()`. Something like

```
.#{$toolbar}__content {
  @include pf-v6-hidden-visible(grid);
}
```

Then remove the `display: grid` declaration on `.#{$toolbar}__content` here, since `pf-v6-hidden-visible()` will set it https://github.com/patternfly/patternfly/blob/9f6d47c2a203ebd5e3867516e9f4f84fd77e859a/src/patternfly/components/Toolbar/toolbar.scss#L82

But we'll want to make sure `.#{$toolbar}` is still set to `display: grid;` and both `.#{$toolbar}` and `.#{$toolbar}__content` are set to `position: relative`.

mcoker (changes requested):
Left a comment about what `pf-v6-hidden-visible()` does that hopefully clarifies a change needed to get this hooked up.

Also for the `.history/*` files showing up, that's the first I've seen of `.history` files making it into a PR but looks like it's in [react's .gitignore](https://github.com/patternfly/patternfly-react/blob/991c7b90aa4282ce01888cad7a9ab6941ed30feb/.gitignore#L26), so feel free to add it to core's as part of this PR!

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -74,11 +74,14 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
 
+.#{$toolbar}__content {
+  @include pf-v6-hidden-visible(grid);
+}
+
 // - Toolbar - Toolbar content
 .#{$toolbar},
 .#{$toolbar}__content {
   position: relative;
-  display: grid;
```
mcoker, line 81: Since this removes `display: grid;` from `.#{$toolbar}`, too, we'll need to add that back. You can probably just add it as the first `.#{$toolbar}` style here

https://github.com/patternfly/patternfly/blob/c7a6d8b0234a64c478bbb49578847d0511031dca/src/patternfly/components/Toolbar/toolbar.scss#L88-L89

mcoker (changes requested):
Looks good, just one small nit!

mcoker (approved):
🚀🚀

Issue #6683, opened 2024-05-22
Toolbar - spacer/margin classes typo


mcoker:
From @thatblindgeye 

> For this comment: https://github.com/patternfly/patternfly-react/issues/10387#issuecomment-2123582803 , if we still plan to utilize margin on toolbar components, looks like the modifier class is missing a hyphen between the prefix and the spacer. pf-m-margin-inline-startlg applies styling, while pf-m-margin-inline-start-lg doesn't

mcoker:
No longer needed, these modifiers will be removed

PR #6684, opened 2024-05-22
fix(toolbar): typo in inline margin modifier classes

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6683


thatblindgeye (approved):

mcoker:
No longer needed, we're going to remove these modifiers https://github.com/patternfly/patternfly-react/pull/10418#pullrequestreview-2074362005

Issue #6685, opened 2024-05-22
Pull update to chart/global/label/fill token to fix dark theme chart labels


mcoker:
Details in https://github.com/patternfly/patternfly/issues/6681

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6686

PR #6686, opened 2024-05-22
fix(charts): update dark label color token

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6685

~~@lboehling just want to make sure the changes to the hex values in the palette token file are intentional? Looks to just change a handful of shorter hex values to the long/full value (eg `#06c` to `#0066cc`)~~ nevermind, core has a linter rule that complains about long hex values that can be shorter. Let the linter fix it for now, we may consider aligning on that format in the future


srambach (approved):
🐐

Issue #6643, opened 2024-05-10
Bug - Table - Adding `screenReaderText` to a empty `Th` element changes column width


mvollmer:
Adding `screenReaderText` to a empty `Th` element changes the `min-width` property of the element, which might change column width.

We have a table with a empty `Th` for the first column. That column contains icons:

![image](https://github.com/patternfly/patternfly/assets/3228183/db5f10d8-2fdb-4ede-bc4e-0a586dc3bb1c)

The code for the header is pretty straighforward:

```
                    <Thead>
                        <Tr>
                            { show_icons && <Th /> }
                            <Th>{_("ID")}</Th>
                            <Th>{_("Type")}</Th>
                            <Th>{_("Location")}</Th>
                            <Th className="storage-size-column-header">{_("Size")}</Th>
                            <Th />
                        </Tr>
                    </Thead>
```

But we have started to receive this warning recently:

> Th: Table headers must have an accessible name. If the Th is
intended to be visually empty, pass in screenReaderText. If the Th
contains only non-text, interactive content such as a checkbox or
expand toggle, pass in an aria-label.

The correct thing to do seems to be to add `screenReaderText="Category"` to make the first column more accessible. This however results in a changed layout:

![image](https://github.com/patternfly/patternfly/assets/3228183/c1d46053-8c84-463b-9db3-fd894c490f04)

The first column is now much wider than it needs to be.

This happens because Patternfly has a CSS rule that sets the `min-width` property of elements inside `thead` to some largish value, except when the element is empty according to the `:empty` pseudo-class, in which case the `min-width` property is zero. Our layout relies on this exception. 

Giving `screenReaderText` a value will add a invisible element into the `Th`, which causes the "min-width: 0" exception to no longer apply.

I had expected that adding `screenReaderText` does not influence the layout. We will have to use `aria-label` instead, I guess.

This is an issue with 
- [x] Patternfly 5  
- [ ] Patternfly 6  
- [ ] other

mattnolting:
> Adding screenReaderText to a empty Th element changes the min-width property of the element, which might change column width.

> This happens because Patternfly has a CSS rule that sets the min-width property of elements inside thead to some largish value, except when the element is empty according to the :empty pseudo-class, in which case the min-width property is zero. Our layout relies on this exception.

Correct, which means this declaration will be updated to accommodate screen reader text.

[...]

PR #6689, opened 2024-05-23
fix(table): updated thead td to th

### LINKED ISSUE #6272: Bug - Table - accessibility rules have changed, all thead cells should be `<th>`s

mattnolting:
closes #6272 
closes https://github.com/patternfly/patternfly/issues/6643

[Backstop Report](https://drive.google.com/file/d/1X98JYLHM_U5YtSaLwGvrg-8aX6j4krD-/view?usp=sharing)

## 8/26 Update

**Header shift is due to `pf-v6-c-table__cell-empty` and is expected**

[Backstop report](https://drive.google.com/file/d/11u1r8QELIo1m9e2a5Y8AAiFoKYNW5Wwz/view?usp=sharing)

Favorites example shouldn't be sorted, this diff is expected

Screenshot 2024-08-26 at 12 44 03 PM

This PR creates a partial container for special table cells. The internals of the partials remain unchanged while the wrapping element detects `IsThead` and returns `th` if true, `td` if false. Empty cells' default screen reader text is set to "Action" (the most common use case) and optionally updated text via `table-cell--text`. 

This shift is due to updating action empty `th` to `pf-v6-c-table__cell-empty`

Screenshot 2024-08-20 at 12 34 35 PM

This  and all other search input shifts are due to button updates unrelated to this PR

Screenshot 2024-08-20 at 12 36 42 PM

### DIFF
--- a/src/patternfly/components/Table/TableCells/table-cell.hbs
+++ b/src/patternfly/components/Table/TableCells/table-cell.hbs
@@ -1,0 +1,16 @@
+{{!-- TODO: update `table-td` to `table-cell` --}}
+{{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
+{{!-- TODO: automate attribution --}}
+
+{{#> config table-cell--tag=(ternary IsThead 'th' 'td')}}
+
+<{{table-cell--tag}} class="{{concat (pfv) 'table__' table-cell--tag ' ' (concat (pfv) 'table' table-cell--classname) ' ' table-cell--modifier}}{{ternary table-cell--IsEmpty (concat (pfv) 'table__cell-empty') null}}"
+  {{{ternary table-cell--aria-label (concat 'aria-label="' table-cell--aria-label '"') null}}}
+  {{{ternary table-cell--attribute table-cell--attribute null}}}>
+
+  {{#if @partial-block}}
+    {{> @partial-block}}
+  {{/if}}
+</{{table-cell--tag}}>
+
+{{/config}}
--- a/src/patternfly/components/Table/TableCells/table-cell.hbs
+++ b/src/patternfly/components/Table/TableCells/table-cell.hbs
@@ -4,9 +4,22 @@
 
 {{#> config table-cell--tag=(ternary IsThead 'th' 'td')}}
 
-<{{table-cell--tag}} class="{{concat (pfv) 'table__' table-cell--tag ' ' (concat (pfv) 'table' table-cell--classname) ' ' table-cell--modifier}}"
+<{{table-cell--tag}} class="{{concat (pfv) 'table__' table-cell--tag}}
+  {{setModifiers
+    table-cell--classname=(concat (pfv) 'table' table-cell--classname)
+    table-cell--IsFavorite=(concat (pfv) 'table__favorite')
+    table-cell--IsSortable=(concat (pfv) 'table__sort')
+    table-cell--modifier=table-cell--modifier
+  }}"
   {{{ternary table-cell--aria-label (concat 'aria-label="' table-cell--aria-label '"') null}}}
-  {{{ternary table-cell--attribute table-cell--attribute null}}}>
+  {{{ternary table-cell--attribute table-cell--attribute null}}}
+  {{{ternary IsThead 'role="columnheader" scope="col"' null}}}
+  {{{ternary table-cell--IsSortable (concat 'aria-sort="' (ternary table-cell--IsAsc 'ascending' (ternary table-cell--desc 'descending' 'none')) '"') null}}}>
+
+
+  {{#if (hasAll table-cell--IsFavorite table-cell--IsSortable)}}
+    {{> table-favorite-sort}}
+  {{/if}}
 
   {{#if @partial-block}}
     {{> @partial-block}}
--- a/src/patternfly/components/Table/TableCells/table-cell.hbs
+++ b/src/patternfly/components/Table/TableCells/table-cell.hbs
@@ -4,7 +4,7 @@
 
 {{#> config table-cell--tag=(ternary IsThead 'th' 'td')}}
 
-<{{table-cell--tag}} class="{{concat (pfv) 'table__' table-cell--tag ' ' (concat (pfv) 'table' table-cell--classname) ' ' table-cell--modifier}}{{ternary table-cell--IsEmpty (concat (pfv) 'table__cell-empty') null}}"
+<{{table-cell--tag}} class="{{concat (pfv) 'table__' table-cell--tag ' ' (concat (pfv) 'table' table-cell--classname) ' ' table-cell--modifier}}"
   {{{ternary table-cell--aria-label (concat 'aria-label="' table-cell--aria-label '"') null}}}
   {{{ternary table-cell--attribute table-cell--attribute null}}}>
 
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -344,7 +344,6 @@ These classes can be used to ensure that the table changes between the tabular a
       {{#> table-td table-td--data-label="Pull requests"}}
         25
       {{/table-td}}
-      <!-- TODO: update overflow menu with menu/menu toggle  -->
       {{> table-cell-overflow-menu}}
     {{/table-tr}}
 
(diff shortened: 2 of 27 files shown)


src/patternfly/components/Table/examples/Table.md
```diff
@@ -526,7 +526,7 @@ These classes can be used to ensure that the table changes between the tabular a
 {{#> table table--id="table-single-select-radio" table--IsGrid=true table--modifier="pf-m-grid-lg" table--attribute='aria-label="This is single select table with radio inputs"'}}
   {{#> table-thead}}
     {{#> table-tr}}
-      {{> table-cell-empty}}
+      {{> table-cell-empty table-cell--text='Radio select'}}
```
thatblindgeye, line 529: Not really a blocker, just depends how strict we want to be about parity between Core and React examples. React uses "Row select" and similar verbiage for these types of empty cells.

src/patternfly/components/Table/TableCells/table-cell.hbs
```diff
@@ -0,0 +1,11 @@
+{{!-- TODO: update `table-td` to `table-cell` --}}
+{{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
+{{!-- TODO: automate attribution --}}
+
+{{#> config table-cell--tag=(ternary IsThead 'th' 'td')}}
+
+<{{table-cell--tag}} class="{{concat (pfv) 'table__' (ternary isThead 'th' 'td') ' ' (concat (pfv) 'table' table-cell--type)}}" {{#if table-cell--attribute}}{{{table-cell--attribute}}}{{/if}}>
+  {{> @partial-block}}
+</{{table-cell--tag}}>
```
thatblindgeye, line 9: We also need to be able to add an aria-label to the `th` element, in instances where there is visible content but no visible text content (checkbox, plain icon button, etc). The checkbox example in React has `<th aria-label="Row select" ....`, while the checkbox inside the th has its own aria-label to denote the action.

Core examples that would need an aria-label on the outer `th` element:

- Checkboxes and actions example
- Expandable example (both the "toggle all" button and "select all" checkbox)
- Expandable with set width columns example
- Expandable with nested table example (from what I can tell, the aria-labelledby on the "toggle all rows" button could be removed as it's not pointing to anything valid/doesn't need to. The aria-labelledby on the individual row toggles could be removed and the aria-label updated to e.g. "Toggle row 1", or the aria-labelledby needs to point to a valid element as they currently do not. This isn't a blocker for this PR, though)
- Compact example
- Compact expandable example
- Clickable example
- Clickable and expandable example
- Borderless example
- Borderless compact example
- Borderless expandable example
- Width modifiers examples
- Nested column headers and expandable rows
- Favorites examples
- Favorites sortable example (would also apply to your other PR for favorite sortable)
- Striped expandable table example

thatblindgeye (changes requested):

src/patternfly/components/Table/TableCells/table-cell.hbs
```diff
@@ -0,0 +1,11 @@
+{{!-- TODO: update `table-td` to `table-cell` --}}
+{{!-- TODO: automate `td/th` based up `IsThead/IsTbody` and make overridable --}}
+{{!-- TODO: automate attribution --}}
+
+{{#> config table-cell--tag=(ternary IsThead 'th' 'td')}}
```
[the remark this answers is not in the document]
mattnolting: `config` is a wrapping agent with no corresponding `.hbs` partial.  `table-thead.hbs` and `table-tbody.hbs` assign [`IsThead`](https://github.com/patternfly/patternfly/blob/v6/src/patternfly/components/Table/table-thead.hbs#L3) and [`IsTbody`](https://github.com/patternfly/patternfly/blob/1f6e12340d166db38e994e5e62de002f3d795f7f/src/patternfly/components/Table/table-tbody.hbs#L3) respectively.

```
{{#>
  configd
  IsBody=true
}}````

**Review by thatblindgeye — APPROVED** [review]

**Comment by mcoker** [comment]
Attaching the backstop report - [BackstopJS Report.pdf](https://github.com/user-attachments/files/16603370/BackstopJS.Report.pdf)

Looks like there may at least be an issue with icons rendering in menu toggles. Here's a toolbar example that's rendering an empty toggle

Screenshot 2024-08-13 at 12 45 20 PM

**Comment by mcoker** [comment]
@thatblindgeye would you mind reviewing this again? Particularly with the last changeset - https://github.com/patternfly/patternfly/pull/6689/commits/68bd944fad9251671371240be84c270627566bba

File: src/patternfly/components/Table/table.scss [hunk 4]
```diff
@@ -466,8 +466,8 @@
 
   // - Table truncate - Table wrap - Table nowrap - Table fit content - Table wrap - Table break word
   :where([class*='#{$table}']),
-  :where(&) > :is(thead, tbody) {
-    @at-root .#{$table} > thead,
+  :where(&) > :is(thead, tbody),
+  &.#{$table}__thead {
```
mcoker, line 470: This compiles to

```
.pf-v6-c-table :where([class*=pf-v6-c-table]),
:where(.pf-v6-c-table) > :is(thead, tbody),
.pf-v6-c-table.pf-v6-c-table__thead
```

Just want to verify that's correct - particularly the last line? Curious when we have `.pf-v6-c-table.pf-v6-c-table__thead` if so.

src/patternfly/demos/Toolbar/toolbar-template.hbs
```diff
@@ -32,7 +32,7 @@ Options: [
       {{!-- toggle group --}}
       {{#if toolbar-template--HasToggleGroup}}
         {{#> toolbar-group toolbar-group--IsToggleGroup=true toolbar-group--modifier="pf-m-show-on-xl"}}
-          {{> toolbar-toggle}}
+          {{> toolbar-toggle menu-toggle-icon='filter'}}
```
mcoker, line 35: Did you mean to leave this in?

```suggestion
          {{> toolbar-toggle}}
```

mcoker (commented):
Left a couple of comments and attaching the backstop report: 
[table-8.19.24.pdf](https://github.com/user-attachments/files/16667212/table-8.19.24.pdf)

Mostly it looks like these two things:
* column widths changed (even when there are no empty headers, actions, or truncated test)
* some `<th>` cells are no longer truncated by default

Are those to be expected?

If there are more updates, would you mind attaching visual regressions and leave a comment explaining the errors?

thatblindgeye (commented):
Other than the a11y failures this looks good

thatblindgeye (approved):

mcoker (commented):
@mattnolting LGTM, just want to confirm the changes in the backstop report are good. 
[BackstopJS Report.pdf](https://github.com/user-attachments/files/16767349/BackstopJS.Report.pdf)

Particularly this change on the [sortable favorites table example](https://patternfly-pr-6689.surge.sh/components/table/html/favorites-sortable-example/) which changes the favorite button from a sorted state to not

Screenshot 2024-08-27 at 1 30 36 PM

src/patternfly/components/Table/table.scss
```diff
@@ -1089,10 +1092,6 @@
   .#{$table}__thead & {
     border-block-end: 0;
   }
-
-  &.pf-m-border-row {
-    border-block-end: var(--#{$table}--border-width--base) solid var(--#{$table}--BorderColor);
-  }
```
mcoker, line 1095: Just want to make sure this is intentional? If so can you give a summary of why? Looks like we'll need a react update for this, too.

src/patternfly/components/Table/table.scss
```diff
@@ -1089,10 +1092,6 @@
...
 }
```
mcoker, line 1095:
```suggestion

  &.pf-m-border-row {
    border-block-end: var(--#{$table}--border-width--base) solid var(--#{$table}--BorderColor);
  }
}
```

mcoker (approved):
🎃👍

PR #6692, opened 2024-05-23
fix(docs): update backstop selector & reference images

### LINKED ISSUE #6602: Visual regression updates
* @wise-king-sullyman  added a [`.page-loaded` class to `<body>`](https://github.com/patternfly/patternfly-org/pull/3789/files) once a full page example loads that we should use as backstop's `readySelector` https://github.com/patternfly/patternfly/blob/cd1e552eede3f3b55f4f40a2cce5ce830b48b08c/backstop.js#L12
* We could look at adding a screenshot job for dark theme! That may be a separate issue unless it's really straight forward.

evwilkin:
Closes #6602

This PR updates the `readySelector` for backstop visual regression testing to the `.page-loaded` class which is ideally added after the page and images for each example have fully loaded.

This PR also updates the reference images for backstop.


mcoker (approved):
Awesome!! Thanks for handling this on org and core!! And now you're a screenshot ninja🥷

PR #6698, opened 2024-05-28
fix(Table): remove duplicate bottom border on mobile collapsed expandable rows

### LINKED ISSUE #6556: Table - remove extra border on mobile collapsed expandable rows
Per Kayla's comment below, remove the extra 1px border under collapsed expandable rows.  

Additional info from dupe https://github.com/patternfly/patternfly/issues/6667#issue-2299104805

> Double bottom-border on non-expanded rows in expandable tables. Here's where the borders are coming from
> 
> https://github.com/patternfly/patternfly/blob/c11387c8b1dea7e573beb0a7f5c4c12a51c1080a/src/patternfly/components/Table/table.scss#L1060-L1062
> 
> https://github.com/patternfly/patternfly/blob/c11387c8b1dea7e573beb0a7f5c4c12a51c1080a/src/patternfly/components/Table/table-grid.scss#L185-L188
> 
> Screenshot 2024-05-15 at 7 54 16 PM

evwilkin:
Closes #6556 

This PR removes the bottom border on collapsed expandable rows on mobile, which is causes a double border to appear.

### DIFF
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -295,10 +295,6 @@
     --#{$table}__tr--responsive--PaddingInlineStart: var(--#{$table}__tr--responsive--nested-table--PaddingInlineStart);
 
     border: 0;
-
-    tr:where(.#{$table}__tr):not(.#{$table}__expandable-row) + tr:where(.#{$table}__tr):not(.#{$table}__expandable-row) {
-      --#{$table}__tr--responsive--PaddingBlockStart: 0;
-    }
   }
 
   // - Table grid compound expansion toggle
--- a/src/patternfly/components/Table/table-grid.scss
+++ b/src/patternfly/components/Table/table-grid.scss
@@ -49,6 +49,9 @@
   --#{$table}__tbody--after--BorderInlineStartWidth: 0;
   --#{$table}__tbody--after--BorderColor: var(--pf-t--global--border--color--clicked);
 
+  // * Table body responsive
+  --#{$table}__tbody--responsive--m-expandable--BorderBlockEndWidth: var(--pf-t--global--border--width--divider--default);
+
   // * Table tr responsive
   --#{$table}__tr--responsive--border-width--base: var(--pf-t--global--border--width--divider--default);
   --#{$table}__tr--responsive--last-child--BorderBlockEndWidth: var(--#{$table}__tbody--responsive--border-width--base);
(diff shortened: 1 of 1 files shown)


kaylachumley (approved):
looks good!

src/patternfly/components/Table/table-grid.scss
```diff
@@ -50,7 +50,7 @@
   --#{$table}__tbody--after--BorderColor: var(--pf-t--global--border--color--clicked);
 
   // * Table tr responsive
-  --#{$table}__tr--responsive--border-width--base: var(--pf-t--global--border--width--divider--default);
+  --#{$table}__tr--responsive--border-width--base: 0;
   --#{$table}__tr--responsive--last-child--BorderBlockEndWidth: var(--#{$table}__tbody--responsive--border-width--base);
```
mattnolting, line 54:
```suggestion
  --#{$table}__tr--responsive--last-child--BorderBlockEndWidth: 0;
```

src/patternfly/components/Table/table-grid.scss
```diff
@@ -50,7 +50,7 @@
   --#{$table}__tbody--after--BorderColor: var(--pf-t--global--border--color--clicked);
 
   // * Table tr responsive
-  --#{$table}__tr--responsive--border-width--base: var(--pf-t--global--border--width--divider--default);
+  --#{$table}__tr--responsive--border-width--base: 0;
```
mattnolting, line 53:
```suggestion
  --#{$table}__tr--responsive--border-width--base: var(--pf-t--global--border--width--divider--default);
```

mattnolting (commented):
This update removes more borders than I believe are intended. 

**Before:**

Screenshot 2024-06-01 at 8 46 54 AM

**After:**

Screenshot 2024-06-01 at 8 47 01 AM

evwilkin:
Suggestions above are correct that the previous code changed more than the intended target, but the proposed solution looks to revert the original problem this PR aims to solve.

The issue is that:
- expandable tables contain tbody that always has a bottom border
- Child tr which has a bottom border when the tbody is not expanded (the double border)

There are several rules similar to this case, but all focus on the tr whereas we need to focus on the class on the parent `table` and go from there so I have added a new style rule.  Note - I believe this new rule can replace [the rule below it](https://github.com/patternfly/patternfly/pull/6698/files#diff-1c73508a6dac9de0a0052bcd77f9193e0f8d95a4d5cfef221569b6b97af1e649R186), which doesn't look to apply any changes as the nesting is causing the table component selector to be added twice?

src/patternfly/components/Table/table-grid.scss
```diff
@@ -177,6 +177,11 @@
     }
   }
 
+  // Remove border on tr inside non-expanded tbody in of expandable tables
+  &.pf-m-expandable tbody:not(.pf-m-expanded) tr {
+    border-block-end: 0;
+  }
+
```
mattnolting, line 184: This does what we want, but something like this would be more straightforward. Disable all `<tr>` `border-block-end`. Initially, this logic applied to a mix and match approach to expandable table (tbody.expandable next to a regular row). However, [React's guidance](https://www.patternfly.org/components/table#expandable) places all expandable table content in a loop, within a `Tbody`. I believe it's safe to take this approach. 

**Add this variable to `:root` (`.#{$table}[class*="pf-m-grid"]` in this case) to avoid potential conflicts**

```
--#{$table}__tbody--responsive--m-expandable--BorderBlockEndWidth: var(--pf-t--global--border--width--divider--default);
```

```
  &.pf-m-expandable {
    --#{$table}__tr--BorderBlockEndWidth: 0; // nested table rows have no border

    .#{$table}__tbody {
      border-block-end: var(--#{$table}__tbody--responsive--border-width) solid var(--#{$table}--responsive--BorderColor);
    }
  }
```

You should then be able to remove [lines 190-199](https://github.com/patternfly/patternfly/pull/6698/files#diff-1c73508a6dac9de0a0052bcd77f9193e0f8d95a4d5cfef221569b6b97af1e649R190-R199) and [lines 201-212](https://github.com/patternfly/patternfly/pull/6698/files#diff-1c73508a6dac9de0a0052bcd77f9193e0f8d95a4d5cfef221569b6b97af1e649R201-R212), but because `pf-v-table` is very complex, we should run a visual regression test before merging.

evwilkin: @mattnolting I've made these changes but noticed I also needed to specifically target `.#{$table}__tbody.pf-m-expanded` to apply as needed so added that specific rule to what you laid out above 👍

mattnolting (commented):
Wonderful. Last thing, I believe you can remove [lines 299 - 300](https://github.com/patternfly/patternfly/pull/6698/files#diff-1c73508a6dac9de0a0052bcd77f9193e0f8d95a4d5cfef221569b6b97af1e649R299-R301). Looks like the padding between rows is uneven. @kaylachumley Can you confirm?

**Before**

Screenshot 2024-06-25 at 10 13 31 AM

**After**

Screenshot 2024-06-25 at 10 38 06 AM

kaylachumley:
yep! That looks better. can confirm to remove rows 299-300! @mattnolting

mattnolting (approved):
LPTM 🤖

PR #6705, opened 2024-05-29
fix(card): move borders to pseudoelement

### LINKED ISSUE #6234: Card - refactor border
Look at moving the border from the card to a pseudoelement.
Update the "revert" properties, which are not currently working.

evwilkin:
Closes #6234 

This PR:
- Moves Card borders to `::before` pseudoelement
- Removes two unneeded `revert` rules

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -115,6 +115,7 @@
   &::before {
     position: absolute;
     inset: 0;
+    pointer-events: none;
     content: "";
     border: var(--#{$card}--BorderColor) var(--#{$card}--BorderStyle) var(--#{$card}--BorderWidth);
     border-radius: inherit;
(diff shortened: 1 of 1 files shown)


evwilkin:
@srambach I ended up removing 2 unneeded `revert` rules but the remaining two do look to work as expected compared to the Figma designs.  If I missed or misunderstood anything with this just lmk!

srambach:
I _think_ those reverts were there because on clickable and selectable cards, there should not be a border when the checkbox is checked, only when it's pf-m-current. You can see the difference on https://patternfly-pr-6705.surge.sh/components/card/#clickable-and-selectable vs. https://pf6.patternfly.org/components/card#clickable-and-selectable (click the checkbox on the first card)

evwilkin:
@srambach you're right, I was testing this incorrectly locally so thanks for pointing this out.  I don't see the background color rule having any effect, but I added back both removed rules out as maybe the background is affected outside of the use case in the example?  

The border revert rules are definitely needed to prevent the border color & size changes on focus-visible/selected. 👍

srambach (approved):
Thanks! 🎴

srambach (commented):
Actually, one sec - I'm seeing a flash when the checkbox is checked now https://patternfly-pr-6705.surge.sh/components/card/#selectable

PR #6708, opened 2024-05-29
fix(hint): remove extra hint actions space

### LINKED ISSUE #6116: Bug - Hint - actions are changing the layout/spacing between hint sections
The presence of an action is creating extra space between sections if the action is taller than the text beside it. This is more pronounced in v6 since the font size there is smaller, but the action spacing is about the same.

Screenshot 2023-12-06 at 2 24 48 PM
Screenshot 2023-12-06 at 2 24 07 PM

evwilkin:
Closes #6116 

This PR mirrors the negative margin-start on hint actions to the margin-bottom, to remove extra spacing added in the hint by the hint actions.

### DIFF
--- a/src/patternfly/components/Hint/examples/Hint.md
+++ b/src/patternfly/components/Hint/examples/Hint.md
@@ -90,3 +90,4 @@ cssPrefix: pf-v6-c-hint
 | `.pf-v6-c-hint__body` | `<div>` | Initiates the hint body element. |
 | `.pf-v6-c-hint__footer` | `<div>` | Initiates the hint footer element. |
 | `.pf-v6-c-hint__actions` | `<div>` | Initiates the hint actions element. |
+| `.pf-m-no-offset` | `.pf-v6-c-hint__actions` | Removes the negative vertical margins on the actions element intended to align the action content with the hint title. |
--- a/src/patternfly/components/Hint/examples/Hint.md
+++ b/src/patternfly/components/Hint/examples/Hint.md
@@ -79,6 +79,21 @@ cssPrefix: pf-v6-c-hint
 {{/hint}}
 ```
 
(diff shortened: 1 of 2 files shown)


mcoker (commented):
LGTM though in hint, we apply the negative margin to a plain menu toggle directly, and in the card we apply it to the actions container with a `.pf-m-no-offset` modifier to remove it. IMO I think consistency with the card would be better because
* allow any sort of plain thing (like a plain button) to get this styling without having to add components/variations to the selector list of what gets a negative margin
* consistency makes a more familiar API and hints seem to effectively be cards without all the bells-n-whistles
* keeps the hint styles on hint elements instead of needing to override other components. related, since this styles a menu toggle as a child of the hint actions, if there were a nested menu toggle, or a menu toggle anywhere in the actions box other than the top-level menu toggles in our examples, it would get the styling, too, which likely isn't wanted

wdyt @evwilkin @srambach?

evwilkin:
@mcoker makes sense to me, thanks for the reference - pushed these updates referencing Card as the example to follow 👍

mcoker (approved):
LGTM, we could add an example, but that can always come later. Would that be a quick add @evwilkin? Using a primary button in place of the plain button would match the [card example](https://www.patternfly.org/components/card/html#actions-with-no-offset).

mattnolting (approved):
Very nice! LGTM 👍 

Screenshot 2024-05-30 at 2 39 10 PM

evwilkin:
@mcoker thanks for the suggestion ✅ 

- [https://patternfly-pr-6708.surge.sh/components/hint#actions-with-no-offset](https://patternfly-pr-6708.surge.sh/components/hint#actions-with-no-offset)

PR #6711, opened 2024-05-29
fix(tokens): rename file and add no-motion modifier

### LINKED ISSUE #6516: Animation - implement scaffolding for handling motion preferences
**Describe the feature**
Building from #6515, implement the scaffolding (necessary added files, global variables, etc.) needed to use the method for handling `prefers-reduced-motion` and a way to turn off all animation for automated testing. One or more of the existing animations can be modified to use and validate the scaffolding.

### LINKED ISSUE #6682: Rename the custom/local tokens stylesheet
Right now this is called `tokens-font.scss` since it initially held font styles that were not defined in figma, but that file has grown to include a lot of custom/local tokens. Naming things is never fun - what about `tokens-local`? cc @srambach @lboehling 

https://github.com/patternfly/patternfly/blob/v6/src/patternfly/base/tokens/tokens-font.scss

srambach:
This renames the tokens-font file to be tokens-local since it evolved to include more than just font - e.g. the transition token override for turning off motion completely is added there in this PR.

Fixes #6516 
Fixes #6682

### DIFF
--- a/src/patternfly/base/tokens/tokens-local.scss
+++ b/src/patternfly/base/tokens/tokens-local.scss
@@ -146,7 +146,23 @@
     var(--pf-t--global--box-shadow--spread--lg)
     var(--pf-t--global--box-shadow--color--lg);
 
-  // Transitions
+  // Motion
+  .pf-m-no-motion {
+    // Set all motion tokens to 0 for testing purposes
+    --pf-t--global--delay--400: 0ms;
+    --pf-t--global--delay--300: 0ms;
+    --pf-t--global--delay--200: 0ms;
+    --pf-t--global--delay--100: 0ms;
+    --pf-t--global--duration--600: 0ms;
+    --pf-t--global--duration--500: 0ms;
+    --pf-t--global--duration--400: 0ms;
+    --pf-t--global--duration--300: 0ms;
+    --pf-t--global--duration--200: 0ms;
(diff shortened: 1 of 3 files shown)


src/patternfly/base/tokens/tokens-local.scss
```diff
@@ -146,7 +146,23 @@
     var(--pf-t--global--box-shadow--spread--lg)
     var(--pf-t--global--box-shadow--color--lg);
 
-  // Transitions
+  // Motion
+  .pf-m-no-motion {
+    // Set all motion tokens to 0 for testing purposes
```
mcoker, line 151: From the comment that it's for testing, this may not be relevant, but I think a good place for this would be [patternfly-common.scss](https://github.com/patternfly/patternfly/blob/493564c35a6768d4c23bd7f68bcd92e71bde66c2/src/patternfly/base/patternfly-common.scss). And just a thought, but if this should stomp on all animations, I wonder if we should consider adding `!important` to these declarations?

mcoker (approved):
🍰

src/patternfly/base/tokens/tokens-local.scss
srambach: I debated about where to put it. I think you're right. I'll move it and add the `!important`

mcoker (approved):
🎃👍

Issue #6248, opened 2024-01-24
V6 Bug - [Nav] - remove `pf-m-expandable` from examples


tlabaj:
**Describe the problem**

`pf-m-expandable` no longer exists in Nav style sheet.  It should be removed form example code.

**How do you reproduce the problem?**
You can see the modifier in examples with expandable nav (https://pf6.patternfly.org/components/navigation/#expanded)

**Expected behavior**
A clear and concise description of the expected behavior.

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6727

PR #6727, opened 2024-05-30
chore(nav): remove references to pf-m-expandable

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6248

#### Expandable
* removes `.pf-m-expandable`
* removes `nav-item--expandable` - it's never used
* removes dupes of `nav-item--IsExpandable`

#### Expanded
* removes `nav-item--IsExpanded` - this is never set, it's only checked for
* renames `nav-item--expanded` to `nav-item--IsExpanded` - more consistent with the other attributes

### DIFF
--- a/src/patternfly/components/Nav/examples/Navigation.md
+++ b/src/patternfly/components/Nav/examples/Navigation.md
@@ -603,7 +603,6 @@ The navigation system relies on several different sub-components:
 | `.pf-m-full-width` | `.pf-v6-c-nav` | Modifies nav for to full width of parent. |
 | `.pf-m-flyout` | `.pf-v6-c-nav__item` | Modifies nav item for the flyout variation. |
 | `.pf-m-scrollable` | `.pf-v6-c-nav` | Modifies nav for the scrollable state. |
-| `.pf-m-expandable` | `.pf-v6-c-nav__item` | Modifies for the expandable state. |
 | `.pf-m-expanded` | `.pf-v6-c-nav__item` | Modifies for the expanded state. |
 | `.pf-m-current` | `.pf-v6-c-nav__link` | Modifies for the current state. |
 | `.pf-m-hover` | `.pf-v6-c-nav__link` | Modifies for the hover state. |
--- a/src/patternfly/components/Nav/examples/Navigation.md
+++ b/src/patternfly/components/Nav/examples/Navigation.md
@@ -135,7 +135,7 @@ import './Navigation.css'
 ```hbs
 {{#> nav nav--attribute='aria-label="Global"'}}
   {{#> nav-list}}
-    {{#> nav-item nav-item--IsExpandable="true" nav-item--expanded="true" nav-item--current="true"}}
+    {{#> nav-item nav-item--IsExpandable="true" nav-item--IsExpanded="true" nav-item--current="true"}}
       {{#> nav-link nav-link--href="#" nav-link--attribute='id="expandable-example1"'}}
         Link 1 (current and expanded example)
       {{/nav-link}}
@@ -157,7 +157,7 @@ import './Navigation.css'
         {{/nav-item}}
       {{/nav-subnav}}
     {{/nav-item}}
-    {{#> nav-item nav-item--IsExpandable="true" nav-item--expanded="true"}}
+    {{#> nav-item nav-item--IsExpandable="true" nav-item--IsExpanded="true"}}
       {{#> nav-link nav-link--href="#" nav-link--attribute='id="expandable-example2"'}}
         Link 2 (expanded, but not current example)
       {{/nav-link}}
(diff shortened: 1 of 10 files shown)


src/patternfly/components/Nav/nav-item.hbs
```diff
@@ -1,15 +1,13 @@
...
     nav-item--IsDrilldown='pf-m-drilldown'
     nav-item--IsFlyout='pf-m-flyout'
     nav-item--modifier=nav--modifier
   }}
   {{#unless nav-item--IsFlyout}}
-    {{#ifAny nav-item--expanded nav-item--IsExpanded}}
+    {{#if nav-item--IsExpanded}}
       pf-m-expanded
-    {{/ifAny}}
+    {{/if}}
```
mattnolting, line 10: You can turn this into a one-liner, without a closing wrapper with 

```
{{ternary nav-item--IsExpanded 'pf-m-expanded' null}}
```

src/patternfly/components/Nav/nav-link.hbs
```diff
@@ -13,12 +13,12 @@
   {{#if nav-item--IsFlyout}}
     aria-haspopup="true"
   {{/if}}
-  {{#ifAny nav-item--IsFlyout nav-item--IsDrilldown nav-item--IsExpandable nav-item--IsExpandable}}
-    {{#ifAny nav-item--expanded nav-item--IsExpanded}}
+  {{#ifAny nav-item--IsFlyout nav-item--IsDrilldown nav-item--IsExpandable}}
+    {{#if nav-item--IsExpanded}}
```
mattnolting, line 17: Also, here you can use the `ternary` helper to dry out this block. That'll removing 4 lines of code. If we do this regularly, we can reduce `.hbs` partials file sizes significantly and eliminate the need for closing tags (`if/ifAny/etc`). It's also reduces complexity by improving scanability. Single quotes denote a string literal since the returned value is `'true/false'`.  

```
aria-expanded="{{ternary nav-item--IsExpanded 'true' 'false'}}"
```

vs 

```
{{#if nav-item--IsExpanded}}
  aria-expanded="true"
{{else}}
  aria-expanded="false"
{{/if}}
```

src/patternfly/components/Nav/examples/Navigation.md
```diff
@@ -135,7 +135,7 @@ import './Navigation.css'
 ```hbs
 {{#> nav nav--attribute='aria-label="Global"'}}
   {{#> nav-list}}
-    {{#> nav-item nav-item--IsExpandable="true" nav-item--expanded="true" nav-item--current="true"}}
+    {{#> nav-item nav-item--IsExpandable="true" nav-item--IsExpanded="true" nav-item--current="true"}}
```
mattnolting, line 138: Technically these settings are [boolean values](https://handlebarsjs.com/guide/expressions.html#helpers) rather than strings. I don't believe we need to change all instances, but it would be nice to omit quotes as we update code.

src/patternfly/components/Nav/nav-subnav.hbs
```diff
@@ -1,13 +1,13 @@
 <section class="{{pfv}}nav__subnav{{#if nav-subnav--modifier}} {{nav-subnav--modifier}}{{/if}}"
   aria-labelledby="{{ternary nav-subnav--aria-labelledby nav-subnav--aria-labelledby nav-item--id}}"
-  {{ternary (concat nav-item--expanded nav-item--IsExpanded) '' 'hidden'}}
+  {{ternary nav-item--IsExpanded '' 'hidden'}}
```
mattnolting, line 3:
```suggestion
  {{ternary nav-item--IsExpanded null 'hidden'}}
```

src/patternfly/demos/DataList/data-list-page-nav.hbs
```diff
@@ -1,6 +1,6 @@
 {{#> nav nav--attribute=(concat 'id="' page--id '-primary-nav" aria-label="Global"')}}
   {{#> nav-list}}
-    {{#> nav-item nav-item--IsExpandable="true" nav-item--expanded="true" nav-item--current="true"}}
+    {{#> nav-item nav-item--IsExpandable="true" nav-item--IsExpanded="true" nav-item--current="true"}}
```
mattnolting, line 3: Would you mind updating `nav-item--current` to `nav-item--IsCurrent` as well?

```suggestion
    {{#> nav-item nav-item--IsExpandable="true" nav-item--IsExpanded="true" nav-item--IsCurrent="true"}}
```

mattnolting (commented):
This looks great! I left some feedback that I hope will be helpful and informative. It would be awesome to enhance handlebars documentation in the near future to memorialize suggestions such as these 😄

src/patternfly/components/Nav/examples/Navigation.md
mattnolting, line 138:
```suggestion
    {{#> nav-item nav-item--IsExpandable=true nav-item--IsExpanded=true nav-item--current=true}}
```

Technically these settings are [boolean values](https://handlebarsjs.com/guide/expressions.html#helpers) rather than strings. I don't believe we need to change all instances, but it would be nice to omit quotes as we update code.

mattnolting (approved):
LGTM 🇹🇭

Issue #6669, opened 2024-05-16
Charts - update string values with tokens


mcoker:
with https://github.com/patternfly/design-tokens/pull/55, there will be some additional charts tokens that weren't generated previously that need to be plugged into the styles

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6728

PR #6728, opened 2024-05-31
fix(charts): update strings to tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6669

uses the tokens that were imported from https://github.com/patternfly/design-tokens/commit/35e3214ab64b6b6a1fb313c34d46dfd3503765b4


srambach (approved):
Looks great!  📊

Issue #6693, opened 2024-05-23
Charts - update color var names to reflect token name


mcoker:
Some of the color names changed for v6 and the charts vars (and react charts references to the corresponding vars in react-tokens) should be updated to match the new color names. High level, the changes are:
* cyan -> teal
* gold -> yellow

The CSS vars that need to be updated are:

https://github.com/patternfly/patternfly/blob/378c9b11d56555ac5754a79266c1dcd03717edaa/src/patternfly/patternfly-charts.scss#L43-L48

https://github.com/patternfly/patternfly/blob/378c9b11d56555ac5754a79266c1dcd03717edaa/src/patternfly/patternfly-charts.scss#L336-L341

https://github.com/patternfly/patternfly/blob/378c9b11d56555ac5754a79266c1dcd03717edaa/src/patternfly/patternfly-charts.scss#L57-L62

https://github.com/patternfly/patternfly/blob/378c9b11d56555ac5754a79266c1dcd03717edaa/src/patternfly/patternfly-charts.scss#L343-L348

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6735

PR #6735, opened 2024-06-04
fix(charts): renamed cyan to teal, gold to yellow

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6693


thatblindgeye (approved):
This looks good. Is there a plan to update other tokens still using the old color names? E.g. in the tokens default file we still have `--pf-t--global--color--nonstatus--gold--300: var(--pf-t--color--yellow--40);`

mcoker:
@thatblindgeye nice callout, opened https://github.com/patternfly/patternfly/issues/6742 for that change (depends on https://github.com/patternfly/design-tokens/issues/62)

PR #6737, opened 2024-06-04
fix(truncate): updated white-space

### LINKED ISSUE #6260: Bug - Truncate - when position "middle" splits at a space, that space disappears
This:

```
<Truncate content="XXXXXXXXXX 1234 4567"
                   position="middle"
                   trailingNumChars={4} />
```

will produce this:

image

Note that the space between the two "4" characters has disappeared.

mattnolting:
closes #6260

### DIFF
--- a/src/patternfly/components/Truncate/truncate.scss
+++ b/src/patternfly/components/Truncate/truncate.scss
@@ -16,7 +16,7 @@
 .#{$truncate}__end {
   overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
+  white-space: pre;
 }
 
 .#{$truncate}__start {


mcoker:
If we're using the example code from https://github.com/patternfly/patternfly/issues/6260 then this is what their rendered code looks like

```
<span class="pf-v6-c-truncate"><span class="pf-v6-c-truncate__start" style="
">XXXXXXXXXX         1234 </span><span class="pf-v6-c-truncate__end">4567</span></span>
```

Looks like the issue is that the missing space is at the end of `__start`, which is getting collapsed since the parent is inline-grid.

`white-space: pre;` is one option, but would need to apply to `__start` to fix the bug above, not `__end`. However, this will also preserve all white-space, which I don't think should be the default for the truncate component.

Screenshot 2024-06-04 at 8 16 52 PM

I could see this as either a variation in core to preserve all white-space, or possibly a react enhancement that could convert any spaces at the end of `__start` and beginning of `__end` to `&nbsp;` maybe? cc @tlabaj wdyt?

mattnolting:
> I could see this as either a variation in core to preserve all white-space, or possibly a react enhancement that could convert any spaces at the end of __start and beginning of __end to &nbsp; maybe? cc @tlabaj wdyt?

@mcoker that was part of [this update](https://github.com/patternfly/patternfly/commit/825d0af0912528722280460f17c82acf5b66c241#diff-4f2fd92c013c3801ba17754994c92c98806b7c53a32f9e6aacf87d9189f5d01dR27) and was discussed as an option. 

> white-space: pre; is one option, but would need to apply to __start to fix the bug above, not __end. However, this will also preserve all white-space, which I don't think should be the default for the truncate component.

@mcoker  `__end` needs it for the [middle variant](https://patternfly-pr-6737.surge.sh/components/truncate#middle)

> <span class="pf-v6-c-truncate"><span class="pf-v6-c-truncate__start" style="
">XXXXXXXXXX         1234 </span><span class="pf-v6-c-truncate__end">4567</span></span>

This doesn't seem like a real use case. If it is, there's likely an error, which isn't part of this issue or our responsibility to govern. 

@evwilkin @srambach WDYT?

mcoker:
@mattnolting were you going to update the parent components (or the children) to have `white-space: pre`?

mcoker (approved):
L🍕TM!

mattnolting:
@mvollmer Does this solution work for you? Removing multiple whitespaces would require JS. That can be addressed in a followup if necessary.

Issue #6739, opened 2024-06-04
Menu - make drill down menu respect prefers-reduced-motion


srambach:
**Describe the feature**
The drill-down menu currently uses a slide to show the drilled-in menus. This is potentially problematic and should have a `prefers-reduced-motion` transition that uses opacity (probably) instead of translateX.

Issue #6590, opened 2024-04-30
Menu toggle - Animating state transitions and menu panel


andrew-ronaldson:
**Goal**
Adding transitions between Default > Hover > Clicked states. This also includes the behaviour of the menu panel that is triggered by the menu toggle button.

Link to [Figma annotations](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/branch/J0HlG83xKgYoT2pMpCl4SY/PatternFly-6%3A-Components?type=design&node-id=22732%3A26347&mode=design&t=Mmwgp55YUkvDrHUz-1)

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6738

PR #6740, opened 2024-06-04
feat(menu): add transitions for menu and items in menu

srambach:
Adds transitions for the menu items, and for the menu itself. Note that for the menu as a whole, popper is controlling it being shown and hidden, so this PR uses !important to override the settings of Popper. Also note that the current implementation of Popper doesn't allow for the menu to slide, but simply transitions opacity.

A [follow-on issue](https://github.com/patternfly/patternfly/issues/6739) should address the motion currently happening in the drill-down menu.

Implements part of #6590


andrew-ronaldson (approved):
You've done it again! Remarkable

mattnolting (approved):
LPTM!

mcoker (approved):
NICEEEE!!!! 🍕🥷🐢

Issue #6045, opened 2023-11-08
Find a way to round corners that doesn't use border-radius


srambach:
**Describe the enhancement or change**
To apply rounded corners with `border-radius`, we also need to add `overflow:hidden`, or else something square inside the container will bump out of the corner. And hiding the overflow means that things like dropdowns will not extend outside the border of the container.

Therefore, we would like to find a solution that rounds the corners as desired without requiring `overflow:hidden`.

Possible solution along these lines: `clip-path: inset(45% 0% 33% 10% round 10px)`

thatblindgeye:
Possibly could be split into 2 issues:

- Find elements where a wrapper is needed, similar to the Page component having a wrapper around main (Card may be one)
- Find an alternative to overflow hidden (dropdowns in Cards will get clipped e.g.)

thatblindgeye:
In terms of scenarios where a container that has rounded corners has overflow scrollbar, adding a new wrapper similar to page may be the best bet to not have the overflow scrollbar "overflow" the edge of the container/cause the rounded corners to be lost beneath the scrollbar.

Customizing the `::-webkit-scrollbar` and `::-webkit-scrollbar-thumb` styles wouldn't really be an option because 1) it would only apply to Chromium browsers, 2) would differ from expected styling (especially if we're not updating the scrollbar in *every* PatternFly space), and 3) it still wouldn't totally solve the issue as from what I can tell, we can't adjust the scrollbar/thumb itself positioning wise so it would still cover the rounded corners.

`overflow: hidden`/clip sort of works in terms of preventing the scrollbar from covering the rounded corners, but if the scrollable area is very large, the scroll thumb gets smaller to the point where it may end up being hidden/difficult to click/look a little "off/funky" depending on the size of the border-radius (on top of issues with menu content then being cut off). The following image is from About Modal:

![Scrollbar thumb being partially hidden by an overflow property of hidden on the scrollable container](https://github.com/patternfly/patternfly/assets/70952936/83c516e5-b901-4769-b84f-294bc2766e9e)

While here is the Menu component where the rounded scrollbar doesn't look too bad even as the thumb gets smaller (About modal has a border radius of 24px while menu only has 6px):

![image](https://github.com/patternfly/patternfly/assets/70952936/7ace3c15-fcc8-4b76-b450-14bb191ec878)

[...]

Issue #6694, opened 2024-05-28
EPIC: patternfly.css performance issues


mattnolting:
# TLDR (from #6369)
patternfly.css has performance issues. These issues are due to global selectors `*`, `:has()`, `:is()`, `:where()`, `:first/last-child`, references to elements rather than selectors, and complex selector chains. Here is our current [performance profile](https://docs.google.com/spreadsheets/d/1NuHHMuCDu4mmxmRkj-MJYqDFXbOyv5phddtBZAaHdOQ/edit#gid=0). It's clear that these issues need to be addressed to optimize our css and through several updates to our codebase, we can do just that. 

## Update: 8/19/24

As I was recorded performance metrics, I found interesting results. While attempting to compare metrics (`match attempts/count` increase/decrease).

![Image](https://github.com/user-attachments/assets/683c9db8-59d9-4d61-a93e-8c204cf70a6f)

I realized that that the selectors I was looking for were no longer available. 

![Image](https://github.com/user-attachments/assets/5f2e6b6e-8e2e-4702-9852-d4b58d8f65fa)

The comparison I was looking for didn't exist because it was completely eliminated. 

By removing `:where(:root, .component)` alone,  `match attempts` went from 6,516,114 to 541,945, resulting in a total reduction of 5,974,169 `match attempts`. 

## Update: 7/30/24

There appear to be two, core causes for performance issues. As noted in @garrett's [issue](https://github.com/patternfly/patternfly/issues/6369), our approach to floating pseudo elements, global selectors, html elements, etc has a significant impact upon First Contentful Paint, Largest Contentful Paint, and Speed Index. The second, devtools based issue is part of our workspace JS.  

## Related content
* https://blogs.windows.com/msedgedev/2023/01/17/the-truth-about-css-selector-performance/
* https://stackoverflow.com/questions/5797014/why-do-browsers-match-css-selectors-from-right-to-left/5813672#5813672
* https://www.stevesouders.com/blog/2009/06/18/simplifying-css-selectors/
* https://webperf.tips/tip/layout-thrashing/

### Pseudo selectors
* `:has()` is bad, don't use it ever unless applied directly to class or id
* floating `:where()` pseudos are gobbling up memory. `:where(:root)` tells browser engines to look at every element on the page. Meaning it's the same as `*` selector. 
* `:is()` adds quite a lot of page load time and should be avoided
* remove `:first/last-child` psuedos

### Complex selectors
[...]

PR #6741, opened 2024-06-04
feat(Panel): added Penta styles

### LINKED ISSUE #6732: Panel - Penta styling
Panel

**Describe the feature**
It doesn't look like Panel has any Penta-y looking styling applied to it on  both [V6 core staging](https://pf6.patternfly.org/components/panel) as well as React. cc @andrew-ronaldson @lboehling @kaylachumley want to check if this is intentional or if it got overlooked?

**Any other information?**

thatblindgeye:
Closes #6732 

Worth noting that the scrollable example currently looks off. I can apply similar styling as the [scrollable menu example](https://pf6.patternfly.org/components/menus/menu#scrollable) (to make the scrollbar look like it has a curve to it), but may ultimately depend on how we want to proceed re: https://github.com/patternfly/patternfly/issues/6045

### DIFF
--- a/src/patternfly/components/Panel/panel.scss
+++ b/src/patternfly/components/Panel/panel.scss
@@ -110,13 +110,6 @@
   padding-block-end: var(--#{$panel}__main-body--PaddingBlockEnd);
   padding-inline-start: var(--#{$panel}__main-body--PaddingInlineStart);
   padding-inline-end: var(--#{$panel}__main-body--PaddingInlineEnd);
-
-  &:has(.#{$panel}__menu) {
-    --#{$panel}__main-body--PaddingBlockStart: 0;
-    --#{$panel}__main-body--PaddingBlockEnd: 0;
-    --#{$panel}__main-body--PaddingInlineStart: 0;
-    --#{$panel}__main-body--PaddingInlineEnd: 0;
-  }
 }
 
 .#{$panel}__footer {
--- a/src/patternfly/components/Panel/panel.scss
+++ b/src/patternfly/components/Panel/panel.scss
@@ -4,10 +4,9 @@
   --#{$panel}--Width: auto;
   --#{$panel}--MinWidth: auto;
   --#{$panel}--MaxWidth: none;
-  --#{$panel}--ZIndex: auto;
   --#{$panel}--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$panel}--BoxShadow: none;
-  --#{$panel}--BorderRadius: var(--pf-t--global--spacer--md);
+  --#{$panel}--BorderRadius: var(--pf-t--global--border--radius--small);
 
   // border
   --#{$panel}--before--BorderWidth: 0;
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Panel/panel.scss
```diff
@@ -7,6 +7,7 @@
   --#{$panel}--ZIndex: auto;
   --#{$panel}--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$panel}--BoxShadow: none;
+  --#{$panel}--BorderRadius: var(--pf-t--global--spacer--md);
```
mcoker, line 10: if we want to match menu, it uses the small border radius

```suggestion
  --#{$panel}--BorderRadius: var(--pf-t--global--border--radius--small);
```

mcoker (commented):
I might suggest these changes, noting where the suggestion comes from for consistency. I'd like @lboehling or @andrew-ronaldson to confirm the design related changes

* https://github.com/patternfly/patternfly/blob/24cb8c672f35e58cf88a922de37da7434afc76a3/src/patternfly/components/Panel/panel.scss#L20
  * I wonder if this should match card, which uses `--pf-t--global--border--width--box--default`
* We've removed `z-index` from things that popper controls (menus, tooltips, popovers, etc), so I'd be in favor of removing it here, too. `z-index` is on things that PF CSS positions like drawers and sticky things https://github.com/patternfly/patternfly/blob/24cb8c672f35e58cf88a922de37da7434afc76a3/src/patternfly/components/Panel/panel.scss#L24
* Menu uses a lg spacer for left/right padding on things
* Menu has a 16px visible top/bottom padding for its search/header and footer sections, though it comes from an 8px gap on the top-level menu component, 8px top/bottom padding on the outer menu wrapper, and 8px top/bottom margin on the search and footer elements. I suppose we can match that, though I'd probably leave off the 8px top/bottom padding on the outer panel component - I believe that's specific to the inner part of a list. Though I wonder if no gap on the parent would be better and we just use 16px (md) top/bottom padding on the header/footer and menu body sections. One way I think of menu is a use case @nicolethoen has mentioned, and I think 16px top/bottom padding, 24px left/right padding works everywhere except menu, which could just drop into the panel or go into a panel__menu wrapper that has no padding:
    ```
    ---------
    | header
    ----------
    | [menu component, maybe with a panel__menu wrapper]
    ----------
    | footer
    ----------
    ```
* Footer uses a custom box-shadow, menu uses `--pf-t--global--box-shadow--md--top` https://github.com/patternfly/patternfly/blob/24cb8c672f35e58cf88a922de37da7434afc76a3/src/patternfly/components/Panel/panel.scss#L53
*

src/patternfly/components/Panel/panel.scss
thatblindgeye: Oop yeah this should be the border--radius token. @lboehling do we want to match Menu or keep the medium border radius per discussion during working session?

lboehling: i agree with @mcoker's points above! i'm down to match this with menu!

thatblindgeye:
@mcoker @lboehling made updates per above. I added a demo to show a Menu inside a Panel that could use some eyes. It feels like we would want to remove the box-shadow that Menu has by default if it's inside a Panel?

mcoker:
> I added a demo to show a Menu inside a Panel that could use some eyes. It feels like we would want to remove the box-shadow that Menu has by default if it's inside a Panel?

@thatblindgeye the plain menu should work nicely for that

lboehling (commented):
gorgeous!

srambach (approved):
👨‍🍳 looks great :)

src/patternfly/components/Panel/panel.scss
```diff
@@ -107,6 +110,13 @@
   padding-block-end: var(--#{$panel}__main-body--PaddingBlockEnd);
   padding-inline-start: var(--#{$panel}__main-body--PaddingInlineStart);
   padding-inline-end: var(--#{$panel}__main-body--PaddingInlineEnd);
+
+  &:has(.#{$panel}__menu) {
```
mcoker, line 114: I wonder if we might handle this a little different for a couple of reasons
* there is a [potential performance concern around using `:has()` ](https://github.com/patternfly/patternfly/issues/6694)
* in other components, the structure is usually something like `component__block` with `component__block-body` as a child (or children if you have multiple sections). So I'd probably expect `__menu` to be adjacent to `__main-body`

Since the react component is composable, wdyt about updating this so `__menu` is adjacent to `__main-body`? So in react it would look something like below, though I'm guessing if you're using this as a menu, you probably wouldn't use `__main-body`, you'd just have `__menu` in the main area?

```
  <Panel>
    <PanelHeader>Header content</PanelHeader>
    <Divider />
    <PanelMain tabIndex={0}>
      <PanelMainBody>
        a sentence or two
      </PanelMainBody>
      <PanelMenu>
        /* menu component */
      </PanelMenu>
      <PanelMainBody>
        a sentence or two
      </PanelMainBody>
    </PanelMain>
    <PanelFooter>Footer content</PanelFooter>
  </Panel>
```

thatblindgeye: Updated the demo to just remov ethe MainBody, but the panel menu is still rendered inside the PanelMain component

src/patternfly/demos/Panel/Panel.md
```diff
@@ -0,0 +1,83 @@
...
+id: Panel
+section: components
+---
+
+## Demos
+
+### With a menu
+
+```hbs
+{{#> panel}}
```
mcoker, line 11: super duper nit, would you mind updating this demo to the riased variant? I imagine that's the most common use case. 

```suggestion
{{#> panel panel--modifier="pf-m-raised"}}
```

mcoker (approved):

PR #6743, opened 2024-06-04
fix(icon): use icon status color tokens

### LINKED ISSUE #6674: Icon component - color variants should use icon colors
In the icon component, the modifiers for color are currently using global color tokens rather than the global **icon** color tokens.

srambach:
Fixes #6674 

Uses icon status colors instead of the test status colors.


mcoker (approved):
👍 LGTM!

PR #6746, opened 2024-06-05
fix(action-list): updated icon spacer to xs

### LINKED ISSUE #6745: Action list - update icon group spacer to xs
Per https://www.figma.com/design/VMEX8Xg2nzhBX8rfBx53jp?node-id=6780-15839&m=dev#809628669

Update action
Screenshot 2024-06-05 at 2 03 04 PM

 list, action group spacing to `xs`

mattnolting:
closes #6745

### DIFF
--- a/src/patternfly/components/ActionList/action-list.scss
+++ b/src/patternfly/components/ActionList/action-list.scss
@@ -1,14 +1,14 @@
 @use '../../sass-utilities' as *;
 
 :where(:root, .#{$action-list}) {
-  --#{$action-list}--RowGap: var(--pf-t--global--spacer--sm);
-  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--2xl);
+  --#{$action-list}--RowGap: var(--pf-t--global--spacer--gap--group--vertical);
+  --#{$action-list}--ColumnGap: var(--pf-t--global--spacer--gap--group-to-group--horizontal);
 
   // * Action list group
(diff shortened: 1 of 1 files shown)


mcoker (approved):
👍

mcoker (changes requested):
We have some semantic spacers being handled in another issue, so I wasn't sure if they were ready to be used yet, but confirmed with @lboehling that they're good to go 🚀

Up to you if you want to go ahead and update the rest of the spacers in this component. If so I'm thinking we can update the following - @lboehling can you confirm this is correct?
* Default row gap (for when actions wrap) - `--#{$action-list}--RowGap` https://github.com/patternfly/patternfly/blob/1f6e12340d166db38e994e5e62de002f3d795f7f/src/patternfly/base/tokens/tokens-default.scss#L440 
* Default column gap (space between groups and/or standalone items not in a group) - `--#{$action-list}--ColumnGap` https://github.com/patternfly/patternfly/blob/1f6e12340d166db38e994e5e62de002f3d795f7f/src/patternfly/base/tokens/tokens-default.scss#L438
* Default group column gap (space between actions in a group) `--#{$action-list}__group--ColumnGap` - https://github.com/patternfly/patternfly/blob/1f6e12340d166db38e994e5e62de002f3d795f7f/src/patternfly/base/tokens/tokens-default.scss#L436

mcoker (approved):
L🎃TM!

Issue #6747, opened 2024-06-06
Base stylesheet cleanup


mcoker:
Discussed the role of [tokens-local.scss](https://github.com/patternfly/patternfly/blob/v6/src/patternfly/base/tokens/tokens-local.scss) and agreed we can use it as an override to variables from figma, so we can update it to load after the rest of the tokens files since it currently is loaded before.

Also we no longer need [workspace-overrides.scss](https://github.com/patternfly/patternfly/blob/v6/src/patternfly/base/tokens/workspace-overrides.scss).

And I noticed we're loading normalize before reset, and seems like that should be the other way around. Reset strips default styles, normalize adds more opinionated styles back.

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6748

Issue #6679, opened 2024-05-21
Dark theme - support for color-scheme


mcoker:
Looks like this got lost in the v5/v6 shuffle https://github.com/patternfly/patternfly/blob/d044b2fe342af5e201b151ad09b4ddc9198e1b04/src/patternfly/base/themes/dark/_globals.scss#L3

mcoker:
closed via https://github.com/patternfly/patternfly/pull/6748

PR #6748, opened 2024-06-06
fix(global): base stylesheets cleanup

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6747
fixes https://github.com/patternfly/patternfly/issues/6679

@srambach @mattnolting changes discussed from today's working session, except I also moved reset.scss to come before normalize.scss. Seems like a change we would want to make - wdyt? I'll also run visual regression tests to make sure nothing unexpected changed from that.

### DIFF
--- a/src/patternfly/base/_index.scss
+++ b/src/patternfly/base/_index.scss
@@ -1,7 +1,7 @@
 // @use '../sass-utilities' as *;
 
-@forward './normalize';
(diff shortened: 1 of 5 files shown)


mattnolting (approved):
LPTM 🌡️

srambach (approved):
🌟

Issue #6749, opened 2024-06-06
promote beta components


nicolethoen:
https://docs.google.com/spreadsheets/d/1uZb4gQ02qBHmJiV_LqdfhAD4zNZIbrXPmqhigBYlVPk/edit#gid=0

PR #5868, opened 2023-08-28
fix(nav): added optional text wrapper for nav text

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5857

PR #6759, opened 2024-06-10
fix(docs): promote various components/features out of beta

nicolethoen:
Closes: https://github.com/patternfly/patternfly/issues/6749

### DIFF
--- a/src/patternfly/components/Accordion/examples/Accordion.md
+++ b/src/patternfly/components/Accordion/examples/Accordion.md
@@ -339,7 +339,7 @@ cssPrefix: pf-v6-c-accordion
 
 ### Toggle icon at start
 
-```hbs isBeta
+```hbs
 {{#> accordion accordion--IsStartAligned=true}}
   {{#> accordion-item}}
     {{#> accordion-toggle}}
--- a/src/patternfly/components/Button/examples/Button.md
+++ b/src/patternfly/components/Button/examples/Button.md
@@ -176,7 +176,7 @@ Sed hendrerit nisi in cursus maximus. Ut malesuada nisi turpis, in condimentum v
 ```
 
 ### Counts
-```hbs isBeta
+```hbs
 {{#*inline "button-counts"}}
 {{#> button button--modifier="pf-m-primary" button--attribute='aria-label="View 7 issues"'}}
   View issues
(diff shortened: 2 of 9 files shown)


mcoker (commented):
Just to confirm:
* We're leaving drag/drop as beta? It's in the spreadsheet but it's a special color, so just wanted to double check that we're leaving it.
* Looks like the label group "add" example is no longer beta, but unless I missed it, I don't see it in the spreadsheet.
* nav item icons is no longer beta - I added that feature/beta label for v6 but not sure if it's necessary. The motivation to make that change is to the vertical nav's ability to collapse down to an icon-only view, so having nav icons as beta would allow us to make changes to that element we might not otherwise be able to, until the vertical nav feature is complete. But the icon element in nav is super simple, so it's probably not necessary. @andrew-ronaldson wdyt?

nicolethoen:
ah - yeah the Nav Link Text was on the spread sheet and from what I could tell, that seemed to be the same as Nav Item icons? I didn't look into it too long so if i removed the label erroneously, I can add it back 👍🏻 but we might need to double check that the appropriate react example is also marked beta? I don't think there is currently parity here, the react example is still called nav link text (i think).

I think I assumed that the label group "add" example was all part of the same editable label group concept. Is it separate? Should it stay in beta?

mcoker:
@nicolethoen gotcha! Some feedback below but tl;dr - the PR looks good, we shouldn't need a beta label for nav icons or add labels.

---

"nav link text" was added as an opt-in last year (https://github.com/patternfly/patternfly/pull/5868) for when non-string text is used in a nav link, though it also is needed for the vertical-nav-collape-to-an-icon feature, so technically they're separate enhancements. That said, I brought this up in the working session and I don't think anyone (including me) think the nav item icon needs to be beta, it's really simple, and any "breaking" changes to the nav icon needed to support collapable-vertical-nav can likely happen behind a class (opt-in) either on the icon itself or behind the class on the nav/sidebar to collapse it.

re: editable label, we added [editable labels](https://github.com/patternfly/patternfly/pull/4097) on 5/27/2021 and ["add" label](https://github.com/patternfly/patternfly/pull/4828) on 5/4/22, so technically they're separate. But we've had the add label for 2 years so and I brought it up in working session and no one anticipates needing to change it in a way that's breaking, so removing the beta label seems like a good move!

mcoker (approved):
```
   🌑
🚀
```

nicolethoen:
@mcoker can we merge this?

PR #6766, opened 2024-06-10
fix(page): fix section spacing

### LINKED ISSUE #6631: [PageSection] - change padding for adjacent page sections with the same background color
`PageSection` is an existing component.

**Describe the feature**
`PageSection` has multiple variants with different colored backgrounds. The padding all around each section is currently `24px`. When adjacent sections have different color backgrounds, the spacing works well. But when adjacent sections have the same background, it looks like a lot of blank space. This has come up in discussions with our UX where the request was to reduce the spacing between adjacent sections.

While it is easy enough to remove the top padding with an override, us devs want to adhere to the PF prescribed layouts. Therefore this is a request to formally adopt such a layout change in the `PageSection` component.

image

image
vs
image

**Any other information?**

srambach:
Fixes #6631 

Moves padding/margin from the page__main-body to the page__main-section.
This requires page__main-body to be required.
Refactors the side margin spacing that had a TODO on it.
Also decreases the height of the examples when they are on the example page so that each one isn't taking up 100vh!

NOTE: To see what it looks like, it currently has 16px (md spacer) in as the space between rather than the 24 that was spec'd.

Backstop report with 24px: https://drive.google.com/file/d/1or9X7tbWaN3TFv41rKexfmS1oV_invdi/view?usp=sharing

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -428,8 +428,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 }
 
 .#{$page}__main-section {
-  display: flex;
-  flex-direction: column;
   gap: var(--#{$page}__main-section--RowGap);
   padding-block-start: var(--#{$page}__main-section--PaddingBlockStart);
   padding-block-end: var(--#{$page}__main-section--PaddingBlockEnd);
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -78,12 +78,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 
   // Limit width
   --#{$page}--section--m-limit-width--MaxWidth: calc(#{pf-size-prem(2000px)} - var(--#{$page}__sidebar--Width));
-
-  // Page main body
-  --#{$page}__main-body--MarginBlockStart: 0;
-  --#{$page}__main-body--MarginBlockEnd: 0;
-  --#{$page}__main-body--PaddingInlineEnd: 0;
-  --#{$page}__main-body--PaddingInlineStart: 0;
   
   // Sticky
   --#{$page}--section--m-sticky-top--ZIndex: var(--pf-t--global--z-index--md);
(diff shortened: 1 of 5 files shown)


andrew-ronaldson (approved):
We looked at this on a call and I think it looks good!

lboehling (approved):
i think this is looking good! do we need follow up issues to update our demos to use the new page-main-body sections?

srambach:
> i think this is looking good! do we need follow up issues to update our demos to use the new page-main-body sections?

The update to the handlebars _should_ be doing this.

src/patternfly/components/Page/page.scss
```diff
@@ -247,7 +254,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   &.pf-m-limit-width {
     display: flex;
     flex-direction: column;
```
mcoker, line 256: Since body is required now, wdyt about moving these styles to this block so all of the main-[whatever] sections (anything that has `__main-body` as children) have this style?

src/patternfly/components/Page/page.scss
```diff
@@ -247,7 +254,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
-    padding: 0;
 
     > .#{$page}__main-body {
       flex: 1;
```
mcoker, line 259: I wonder if we can remove this? It's causing this issue of the parent section fills the available space, the `main-body` elements will all grow to fill the space

Screenshot 2024-06-14 at 10 45 58 AM

mcoker (commented):
This looks great! I'm assuming we want to include `__main-body` by default in all of the other types of page sections? With the possible exception of `__main-group` since it takes other sections as children. But I suppose we could add it there, too, to be consistent, if there are no side effects.

With v5, if a page had `.pf-m-limit-width`, we conditionally rendered `__main-body` and moved the padding from the outer section to the `__main-body`. WDYT about updating to just let the padding always remain on the outer section, and never move into the `__main-body`? That would allow us to simplify the CSS quite a bit. It would mean that within `__main-body` on a width limited section, there would be no padding-inline-end, but that should be fine. It would create a slight difference in the available space for content in a width limited section compared to v5 (more space in v6 since the padding-inline-end isn't on the inner `__main-body` with the `max-width`), but that seems fine to me.

mcoker:
Followup for the add'l stuff - https://github.com/patternfly/patternfly/issues/6801

PR #6775, opened 2024-06-11
fix(alert): update title tokens

andrew-ronaldson:


lboehling:
lookin gooood but i'm noticing the icons are little out of alignment 
image

srambach:
> lookin gooood but i'm noticing the icons are little out of alignment 

Yeah - there's a little adjustment on the icon that we can change. Line 32 `--#{$alert}__icon--MarginBlockStart: #{pf-size-prem(4px)};`

srambach:
On popover, the close button overlaps the body text a little now and the baseline of the text isn't aligned with the close button. @mcoker do you think we need a header wrapper?

https://github.com/patternfly/patternfly/assets/19825616/b77ba09b-17ee-4e13-9440-40213d2d3328

mcoker:
@andrew-ronaldson @lboehling looks like the popover close button is misaligned from this style failing from `--pf-t--global--spacer--control--vertical--compact` not being defined.
 https://github.com/patternfly/patternfly/blob/bd110273efa5e6365a8b0cc42c3de24784a98e8d/src/patternfly/components/Popover/popover.scss#L50

We can resolve it by changing the `calc()` to use `--pf-t--global--spacer--action--vertical--plain` as the offset to move the button up (by it's own top padding) - assuming that is the correct semantic token for the vertical padding on plain buttons. @lboehling can you verify that's the correct token for a plain button's top padding? Here's what the updated `calc()` would look like:

```
calc(var(--#{$popover}__content--PaddingBlockStart) - var(--pf-t--global--spacer--action--vertical--plain);
```

That doesn't have to happen in this PR though, if we want to get this through and follow up on that button - up to you @andrew-ronaldson 

As an aside, a couple of questions from looking at this:
* do we need to add `--pf-t--global--spacer--control--vertical--compact` back, or any like it?
* I noticed we have these tokens - assuming they're still valid, should these be named `--action` instead of `--button`? https://github.com/patternfly/patternfly/blob/625dcc9e49a0baf1867f40eb7c02e7c0adb93b05/src/patternfly/base/tokens/tokens-default.scss#L271-L272

src/patternfly/components/Alert/alert.scss
```diff
@@ -34,9 +34,9 @@
   --#{$alert}__icon--FontSize: var(--pf-t--global--icon--size--md);
 
   // Title
-  --#{$alert}__title--FontFamily: var(--pf-t--global--font--family--heading);
-  --#{$alert}__title--FontSize: var(--pf-t--global--font--size--heading--xs);
-  --#{$alert}__title--FontWeight: var(--pf-t--global--font--weight--heading);
+  --#{$alert}__title--FontFamily: var(--pf-t--global--font--family--body);
+  --#{$alert}__title--FontSize: var(--pf-t--global--font--size--body--default);
```
mcoker, line 38: It's fine to have these as component vars, especially if we want to open the font-family and font-size to be themed, but technically we don't need them, since they should inherit both of these styles from the global styles. In particular, I'd say we drop `--#{$alert}__title--FontFamily` unless you want users to customize it. We only set the font-family to `--pf-t--global--font--size--body--default` in 2 other places:
* Multiple file upload "separator" text (the word "or" here - https://patternfly-pr-6775.surge.sh/components/file-upload/multiple-file-upload#basic) https://github.com/patternfly/patternfly/blob/625dcc9e49a0baf1867f40eb7c02e7c0adb93b05/src/patternfly/components/MultipleFileUpload/multiple-file-upload.scss#L44
  * That's because the word "or" is part of the title, and the title uses the `--heading` font-family, so we need to reset it for the "or" text.
* Popover title text, which I'd advocate removing unless we expect the font-family to be themed. My guess as to why it's set is similar to what I'm assuming the change in this PR is. There was a declaration for `font-family` to set it to the `--heading` font, so we updated it to use `--body` when that isn't necessary since it will be `--body` by default from PF's global styles

src/patternfly/components/Alert/alert.scss
```diff
@@ -34,9 +34,9 @@
...
+  --#{$alert}__title--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$alert}__title--LineHeight: var(--pf-t--global--font--line-height--heading);
```
mcoker, line 40: We should remove this `line-height--heading` var if we don't expect users to theme the font-family. It was only there before because we always declare line-height use `line-height--heading` any time we use the `font--family--heading` token. PF global styles set `line-height--body` on everything by default.

src/patternfly/components/Hint/hint.scss
```diff
@@ -13,9 +13,9 @@
   --#{$hint}--Color: var(--pf-t--global--text--color--regular);
 
   // Hint Title
-  --#{$hint}__title--FontFamily: var(--pf-t--global--font--family--heading);
-  --#{$hint}__title--FontSize: var(--pf-t--global--font--size--heading--xs);
-  --#{$hint}__title--FontWeight: var(--pf-t--global--font--weight--heading);
+  --#{$hint}__title--FontFamily: var(--pf-t--global--font--family--body);
+  --#{$hint}__title--FontSize: var(--pf-t--global--font--size--body--default);
+  --#{$hint}__title--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$hint}__title--LineHeight: var(--pf-t--global--font--line-height--heading);
```
mcoker, line 19: Same here about removing font-family and line-height unless we expect users to theme the font-family for hint titles. We can also technically remove `--#{$hint}__title--FontSize: var(--pf-t--global--font--size--body--default);` since the component will inherit the default body font-size from our global styles, though it's totally fine to leave it if we expect users to theme the font-size, which seems totally fair.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -62,9 +62,9 @@
   --#{$notification-drawer}__list-item-header-icon--MarginInlineEnd: var(--pf-t--global--spacer--sm);
 
   // List item header title
-  --#{$notification-drawer}__list-item-header-title--FontFamily: var(--pf-t--global--font--family--heading);
-  --#{$notification-drawer}__list-item-header-title--FontSize: var(--pf-t--global--font--size--heading--xs);
-  --#{$notification-drawer}__list-item-header-title--FontWeight: var(--pf-t--global--font--weight--heading--bold);
+  --#{$notification-drawer}__list-item-header-title--FontFamily: var(--pf-t--global--font--family--body);
+  --#{$notification-drawer}__list-item-header-title--FontSize: var(--pf-t--global--font--size--body--default);
+  --#{$notification-drawer}__list-item-header-title--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$notification-drawer}__list-item-header-title--LineHeight: var(--pf-t--global--font--line-height--heading);
```
mcoker, line 68: Same comment here as in alert, hint, and popover.

src/patternfly/components/Popover/popover.scss
```diff
@@ -57,7 +57,7 @@
   // Title text
   --#{$popover}__title-text--Color: var(--pf-t--global--text--color--regular);
   --#{$popover}__title-text--FontFamily: var(--pf-t--global--font--family--body);
-  --#{$popover}__title-text--FontSize: var(--pf-t--global--font--size--body--lg);
+  --#{$popover}__title-text--FontSize: var(--pf-t--global--font--size--body--default);
```
mcoker, line 60: Same here as alert, hint, and notification drawer.

mcoker (commented):
```
Left some feedback for your perusal good sir  🥸
```

src/patternfly/components/Alert/alert.scss
```diff
@@ -29,15 +29,12 @@
 
   // Icon
   --#{$alert}__icon--Color: var(--pf-t--global--icon--color--regular);
-  --#{$alert}__icon--MarginBlockStart: #{pf-size-prem(4px)};
+  --#{$alert}__icon--MarginBlockStart: #{pf-size-prem(2px)};
```
srambach, line 32: I'm not sure what else changed, but 4 looks right to me now. 🙈

srambach (approved):
Alignments look great now

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -62,10 +62,7 @@
   --#{$notification-drawer}__list-item-header-icon--MarginInlineEnd: var(--pf-t--global--spacer--sm);
 
   // List item header title
-  --#{$notification-drawer}__list-item-header-title--FontFamily: var(--pf-t--global--font--family--heading);
-  --#{$notification-drawer}__list-item-header-title--FontSize: var(--pf-t--global--font--size--heading--xs);
-  --#{$notification-drawer}__list-item-header-title--FontWeight: var(--pf-t--global--font--weight--heading--bold);
-  --#{$notification-drawer}__list-item-header-title--LineHeight: var(--pf-t--global--font--line-height--heading);
```
mcoker, line 68: looks like we removed the `__list-item-header-title` (title in an individual notification item) font vars, then updated properties for `__header-title`. Was that intentional, or did you mean to make property updates to this block? https://github.com/patternfly/patternfly/blob/b12c46a93ee106e5e77c98e049760209089d8deb/src/patternfly/components/NotificationDrawer/notification-drawer.scss#L228-L232

src/patternfly/components/Popover/popover.scss
```diff
@@ -47,7 +47,7 @@
   --#{$popover}__arrow--m-block-right--InsetInlineEnd: var(--pf-t--global--border--radius--medium);
 
   // Close
-  --#{$popover}__close--InsetBlockStart: calc(var(--#{$popover}__content--PaddingBlockStart) - (var(--pf-t--global--spacer--control--vertical--compact) * 1.5)); // align top of button with top of text
+  --#{$popover}__close--InsetBlockStart: calc(var(--#{$popover}__content--PaddingBlockStart) - var(--pf-t--global--spacer--action--vertical--plain)); // align top of button with top of text
```
mcoker, line 50: From the last meeting we had with @thatblindgeye and @lboehling, we're going to use `--pf-t--global--spacer--control--vertical--[...]` for the top/bottom padding of all buttons/controls, so since this var references the top padding of an action, I'd update this to

```suggestion
  --#{$popover}__close--InsetBlockStart: calc(var(--#{$popover}__content--PaddingBlockStart) - var(--pf-t--global--spacer--control--vertical--default)); // align top of button with top of text 
```

src/patternfly/components/Popover/popover.scss
```diff
@@ -56,9 +56,9 @@
 
   // Title text
   --#{$popover}__title-text--Color: var(--pf-t--global--text--color--regular);
-  --#{$popover}__title-text--FontFamily: var(--pf-t--global--font--family--body);
-  --#{$popover}__title-text--FontSize: var(--pf-t--global--font--size--body--lg);
   --#{$popover}__title-text--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$popover}__title-text--FontSize: var(--pf-t--global--font--size--heading--xs);
```
mcoker, line 60: Since this is using a body font, I'd think we wouldn't want to use the heading font-size? Technically it should work just fine, but AFAIK we aren't mixing the body and heading tokens. The equivalent in the body font-size tokens would be `--pf-t--global--font--size--body--lg`.

mcoker (changes requested):
Just a few nits!

src/patternfly/components/NotificationDrawer/notification-drawer.scss
mcoker, line 68: Looks like we're still using this `--LineHeight` here, which we can remove https://github.com/patternfly/patternfly/blob/16cf11c1174c62157e2480faed8c1945a5bb65b2/src/patternfly/components/NotificationDrawer/notification-drawer.scss#L230

mcoker (approved):
🐝utiful!!!

lboehling:
pretty pretty!!!

Issue #6765, opened 2024-06-10
Navigation: Text-decoration: none on hover state


andrew-ronaldson:
The nav-items are assuming the global a:hover underline styling. I'd like to add a specific property to nav-item/link to remove that underline styling.

![Screenshot 2024-06-10 at 4 39 02 PM](https://github.com/patternfly/patternfly/assets/28658548/a7dc67ea-a351-40e3-87ea-5ab7f2875b34)

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6787

PR #6787, opened 2024-06-13
fix(nav): remove underline on hover

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6765

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -215,6 +215,7 @@
   line-height: var(--#{$nav}__link--LineHeight, inherit); // default to nav line height if no custom value is set
   color: var(--#{$nav}__link--Color);
   text-align: start;
+  text-decoration: none;
   background-color: var(--#{$nav}__link--BackgroundColor);
   border: none;
   border-radius: var(--#{$nav}__link--BorderRadius);
@@ -229,7 +230,6 @@
   &.pf-m-hover,
   &:focus {
     color: var(--#{$nav}__link--hover--Color);
-    text-decoration: none;
     background-color: var(--#{$nav}__link--hover--BackgroundColor);
   }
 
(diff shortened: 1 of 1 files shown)


srambach (approved):
Delineful!

mattnolting (approved):
Looks like we apply `text-decoration: none;` to the interactive element itself within several components (icon, jump-links, menu-item, simple-list-item-link, tab-link) and on `:hover/:focus` in others (breadcrumb-link, label, toggle-group). I prefer setting on the element itself, but we use both, so not a blocker. WDYT about moving to the element's default state?

andrew-ronaldson (approved):
You've crossed the line from awesome to super hero!

mattnolting:
Also, should you use the semantic token `--pf-t--global--link--text-decoration` instead of `text-decoration: none`?

mcoker:
> Also, should you use the semantic token --pf-t--global--link--text-decoration instead of text-decoration: none?

@mattnolting I don't think so, reason being `--pf-t--global--link--text-decoration` is the token/setting for a link's default text decoration. If you were to set that to something custom in an app, we wouldn't want that styling to apply to nav/other (more-button-looking links in components) on hover/focus.

mcoker:
> Looks like we apply text-decoration: none; to the interactive element itself within several components (icon, jump-links, menu-item, simple-list-item-link, tab-link) and on :hover/:focus in others (breadcrumb-link, label, toggle-group). I prefer setting on the element itself, but we use both, so not a blocker. WDYT about moving to the element's default state?

@mattnolting nice! We should probably move the rule to the base element. If someone declared `--pf-t--global--link--text-decoration` to underline regular text links, we wouldn't want that to persist to nav and components like this that we don't want to present like text links 👍👍

srambach (approved):
LPTM 🚀

Issue #6730, opened 2024-06-02
Bug - [Progress Stepper] - [Description of step is misaligned when using the Popover]


ChrisLee5000:
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [ ] Patternfly 6  
- [ ] other

Please give a clear and concise description of the problem. Which components are affected?

**How do you reproduce the problem?**

When you use a popover on the step, the description below is misaligned by a few pixels.

**Expected behavior**
The step title is left-aligned with the description, as is the case with the non popover step variant.

**Is this issue blocking you?**
I might try to override the popover title to push it left.

**Screenshots**
![Capture](https://github.com/patternfly/patternfly/assets/162269742/073b32a1-b67c-4121-aeed-b145f12bda64)

**What is your environment?**
 - OS: Win 10
 - Browser Chrome
 - Version Latest

**What is your product and what release date are you targeting?**

**Any other information?**

PR #6791, opened 2024-06-13
fix(progress-stepper): remove padding from help-text title button

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6730

scoped this to `__step-title.pf-m-help-text` (when we make the title a `<button>`) so it has less of a chance of impacting any existing user overrides.


thatblindgeye (approved):

srambach (approved):
That looks good, thank you

PR #6802, opened 2024-06-17
feat(TextInputGroup): added validation

### LINKED ISSUE #6654: Text input group - support for validation
Text input group is basically a super-deluxe version of a [text input](https://www.patternfly.org/components/forms/form-control#input) that supports actions and hwhat-not adjacent to the text input, so it should support validation like the more minimalist regular text input does.

One potential thing to look for is that the text input group currently puts all of the extra stuff it supports beside the `<input type="text">` in the component, which is visible when you focus on the `<input>` in the text input group:

Screenshot 2024-05-14 at 3 41 41 PM

Design has requested that the validation icon be placed within the `<input>` to match the regular text input validation icon, which implies a somewhat more complicated layout than if the icon were beside the `<input>`. Here's the current text input validation icon, which is how it should look in the text input group, too.

Screenshot 2024-05-14 at 3 41 22 PM

### LINKED ISSUE #6796: Text input group - disabled lost border-radius
In https://github.com/patternfly/patternfly/pull/6614/files we moved the border-radius from the component wrapper and put it on `__text` - that may have something to do with it?

thatblindgeye:
Closes #6654 , closes #6796

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group-template-status.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-template-status.hbs
@@ -1,0 +1,7 @@
+{{#if text-input-group-main--IsSuccess }}
+  <i class="fas fa-check-circle"></i>
+{{else if text-input-group-main--IsWarning }}
+  <i class="fas fa-exclamation-triangle"></i>
+{{else if text-input-group-main--IsError }}
+  <i class="fas fa-exclamation-circle"></i>
+{{/if}}
--- a/src/patternfly/components/TextInputGroup/text-input-group-template-status.hbs
+++ b/src/patternfly/components/TextInputGroup/text-input-group-template-status.hbs
@@ -1,7 +1,7 @@
-{{#if text-input-group-main--IsSuccess }}
-  <i class="fas fa-check-circle"></i>
-{{else if text-input-group-main--IsWarning }}
-  <i class="fas fa-exclamation-triangle"></i>
-{{else if text-input-group-main--IsError }}
-  <i class="fas fa-exclamation-circle"></i>
+{{#if text-input-group--IsSuccess }}
+  <i aria-hidden="true" class="fas fa-check-circle"></i>
+{{else if text-input-group--IsWarning }}
+  <i aria-hidden="true" class="fas fa-exclamation-triangle"></i>
+{{else if text-input-group--IsError }}
+  <i aria-hidden="true" class="fas fa-exclamation-circle"></i>
 {{/if}}
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -94,10 +94,6 @@
     border-end-end-radius: var(--#{$text-input-group}__text--BorderEndEndRadius);
   }
 
-  &:hover {
-    --#{$text-input-group}--BorderColor: var(--#{$text-input-group}--m-hover--BorderColor);
-  }
-
   &.pf-m-disabled {
     --#{$text-input-group}--Color: var(--#{$text-input-group}--m-disabled--Color);
     --#{$text-input-group}__icon--Color: var(--#{$text-input-group}--m-disabled__icon--Color);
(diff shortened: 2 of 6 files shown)


andrew-ronaldson (approved):
This looks good to me. One unrelated nit is that we talked about moving the X button to flex-start instead of center. Could we squeeze that in or make its own issue

Screenshot 2024-06-26 at 9 25 24 AM

mcoker:
@thatblindgeye if you rebase this PR, the a11y errors should go away since we removed the context selector component.

mcoker:
@thatblindgeye @andrew-ronaldson do we want badges and other actions in the utilities element to be centered vertically all the time for both single-line and multi-line text input group layouts? Or should everything be vertically centered in a single-line text input group, but be start aligned (vertically centered with the first line of content) in a mult-line text input group?

andrew-ronaldson:
If the children of the utilities container were aligned to top would we need extra margin around the badge to make it visually centered with the X button. 
Screenshot 2024-06-27 at 11 14 54 AM

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -66,6 +80,7 @@
   &.pf-m-disabled {
     --#{$text-input-group}--Color: var(--#{$text-input-group}--m-disabled--Color);
     --#{$text-input-group}__icon--Color: var(--#{$text-input-group}--m-disabled--Color);
+    --#{$text-input-group}__icon--m-status--Color: var(--#{$text-input-group}--m-disabled--Color);
```
mcoker, line 83: We should probably update these to use the disabled icon color token

src/patternfly/components/TextInputGroup/text-input-group-template-status.hbs
```diff
@@ -0,0 +1,7 @@
+{{#if text-input-group-main--IsSuccess }}
+  <i class="fas fa-check-circle"></i>
+{{else if text-input-group-main--IsWarning }}
+  <i class="fas fa-exclamation-triangle"></i>
+{{else if text-input-group-main--IsError }}
+  <i class="fas fa-exclamation-circle"></i>
+{{/if}}
```
mattnolting, line 7:
```suggestion
{{/if}}

```

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -89,10 +104,48 @@
...
+    border-end-end-radius: var(--#{$text-input-group}__text--BorderEndEndRadius);
+  }
+  
+  &:where(.pf-m-icon, .pf-m-succes, .pf-m-warning, .pf-m-error) {
+    --#{$text-input-group}__text--Position: relative;
+  }
+
   &.pf-m-icon {
     --#{$text-input-group}__text-input--PaddingInlineStart: var(--#{$text-input-group}__main--m-icon__text-input--PaddingInlineStart);
   }
```
mattnolting, line 125:
```suggestion

    display: inline-flex;
    align-items: center;
    justify-content: center;
    min-width: calc(var(--#{$text-input-group}__LineHeight) * --#{$text-input-group}__FontSize));
  }
```

mattnolting: Add these to `:where(:root)`

```
  --#{$text-input-group}__LineHeight): var(--pf-t--global--font--line-height--body);
  --#{$text-input-group}__FontSize: var(--pf-t--global--font--size--body--default);
```

mattnolting: Add these to `.#{$text-input-group}`

```
  line-height: var(--#{$text-input-group}__LineHeight);
  font-size: var(--#{$text-input-group}__FontSize);
```

mattnolting (changes requested):
Great work @thatblindgeye! Only thing I'm noticing is icon positioning. Following comments will address. 

**Before:**

Screenshot 2024-07-02 at 3 31 06 PM

**Update:**

Screenshot 2024-07-02 at 3 30 50 PM

mcoker (changes requested):
Looks fantastic. I left some comments and am also curious about a couple of things:
* We're drawing the border on the outer text input group wrapper from  the child `__main` element. That works now, but if we need to add `position: relative` to `__main`, we'll no longer be able to create the border from `__main`. Can we move both the status modifiers and the visible border to the outer `.pf-v6-c-text-input-group` element?
* What's the role of `.pf-m-status`? One thing is I see we're adding selectors like `.thing:where(.pf-m-success, .pf-m-error, ...)`... we could require `.pf-m-status` to be used with any kind of status on an element like `class="pf-m-status pf-m-error"`, and then instead of listing all of the classes in our selector, we could just target `.pf-m-status`.

andrew-ronaldson:
Changes look good Eric! I will close the related issue for the border-radius

mcoker (approved):
🐳

PR #6804, opened 2024-06-17
feat(alert-group): animate addition and removal of alerts

### LINKED ISSUE #6592: Alerts - Animating the toast alerts behaviours
Goal
Follow up to design issue https://github.com/patternfly/patternfly-design/issues/1292

Adding a small jiggle on the menu toggle if it does not pass validation. If this animation is possible, we can create more issues for the other form elements that use validation.

Link to [Figma annotations](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/branch/ORqbrfhmIEkTwphXeX3Iq2/PatternFly-6%3A-Components?type=design&node-id=23637%3A4127&mode=design&t=YQ6fhYv6fAJLn8JT-1)

srambach:
This add animation on the alert group item when adding/removing an item.

For expediency, for now there are no additional component variables - motion tokens are used directly in the transitions.
One other TODO remains: a toast alert that times out should leave in the same way, but have a slower duration than an alert that is dismissed.

To test, go to the [toast alert full page example](https://patternfly-pr-6804.surge.sh/components/alert/html/toast-alert-group/). Add `.pf-m-offstage-top` to simulate the starting condition. Remove it to see the alert appear (slide in from the top). To simulate removal, add the `.pf-m-offstage-right` class and see it slide to the right.

To test reduced motion, turn on "Reduce motion" in the Display settings (MacOS). When this option is on, only opacity should transition and nothing should slide.

Fixes #6592

### DIFF
--- a/src/patternfly/components/Alert/alert-group.scss
+++ b/src/patternfly/components/Alert/alert-group.scss
@@ -94,6 +94,7 @@
     // make the item have no height and position it up above
     grid-template-rows: 0fr;
     margin-block: 0;
+    overflow: hidden;
     opacity: 0;
     transform: translateY(-100%);
 
@@ -102,7 +103,6 @@
       min-height: 0;
       padding-block-start: 0;
       padding-block-end: 0;
-      overflow: hidden;
       border-width: 0;
     }
   }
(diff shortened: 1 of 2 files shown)


andrew-ronaldson (approved):
Glorious

mcoker (approved):
lgtm! We can clean up the vars in a follow-up.

Issue #6245, opened 2024-01-22
Button - use gap instead of left/right margin for icon button


srambach:
**Is this a new component or an extension of an existing one?**
On the icon button, we currently use `pf-m-start` and `pf-m-end` to apply a margin to the left or right of the icon to ensure proper space between the icon and text. If we use gap instead, the gap can collapse if there is no text. In addition, we can probably remove the two modifier classes since gap will work in both cases.

**Any other information?**
React is using an icon button for the notification badge and if there is no text, the icon is currently mis-aligned because of the extra space added. (React requires one of the modifiers)

@wise-king-sullyman

mattnolting:
https://github.com/patternfly/patternfly/pull/6167/files/88dd1182d2456ddc1750a186db004e40759ce642#r1468135792

Issue #5944, opened 2023-09-18
Bug - Button - Icons dont seem to have appropriate padding before button text


nicolethoen:
![Screenshot 2023-09-18 at 12 07 36 PM](https://github.com/patternfly/patternfly/assets/26305082/4e8b4834-5ae1-4d98-b2b2-10d36f9390f1)

[view in codesandbox](https://codesandbox.io/s/ancient-browser-9mhx4r)

The space between the icon and the button text appears too small.

reported by Ansible (James Talton)

srambach:
Note: this appears to happen because of a missing space character between the text and icon. The plan is to wrap the text in a container and apply a flex layout to the button.

PR #6806, opened 2024-06-18
fix(button): add icon/text wrapper

mcoker:
fixes https://github.com/patternfly/patternfly/issues/5944
fixes https://github.com/patternfly/patternfly/issues/6245

~~[BackstopJS Report-6.18.24.pdf](https://github.com/user-attachments/files/15891164/BackstopJS.Report-6.18.24.pdf)~~

Backstop visual regression report - [backstop-new-new-6.18.24.pdf](https://github.com/user-attachments/files/15892839/backstop-new-new-6.18.24.pdf)

[First commit](https://github.com/patternfly/patternfly/pull/6806/commits/ff4c7e2822c0298283b10dc4333a9b1788050338) is the CSS change if you want to look at just that. [Second commit](https://github.com/patternfly/patternfly/pull/6806/commits/1ee915881aa07f924420c4040ea39cbeea83641e) is the hbs updates.

I only updated the button component examples/demos (not in any of the other  but the backstop report looks good. A follow up would be to update the use of button in other component examples/demos.

TL;DR of the hbs change creates a similar API to the proposed react component change. The `{{#> button}}`'s `@partial-block` (or "children" in react) will always be the button text. Any time an icon is used, we'll pass it as an hbs attr/prop. With the change in hbs, that means all the other stuff (like count or whatever) will also need to be added via an hbs attr/prop.

### DIFF
--- a/src/patternfly/components/Badge/badge.hbs
+++ b/src/patternfly/components/Badge/badge.hbs
@@ -1,4 +1,5 @@
 <span class="{{pfv}}badge
+  {{#if badge--IsRead}}pf-m-read {{/if}}
   {{#if badge--IsUnread}}pf-m-unread {{/if}}
   {{#if badge--IsDisabled}}pf-m-disabled {{/if}}
   {{#if badge--modifier}} {{badge--modifier}}{{/if}}"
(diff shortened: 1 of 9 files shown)


srambach (approved):
OK, looks good with the little inline fix! 😹

thatblindgeye (approved):

andrew-ronaldson (approved):
it took my breath away.

srambach (approved):
Pushes all the right buttons

PR #6807, opened 2024-06-20
fix(Wizard): updated styling for wizard body

### LINKED ISSUE #6809: Bug - Wizard v6 - add styling for wizard main/body
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Pre-followup to https://github.com/patternfly/patternfly-react/pull/10637. React was previously allowing WizardBody to be omitted by passing `body={null}` to WizardStep. The linked PR removes that option, but we needed additional styling to have the content fill its container.

thatblindgeye:
Closes #6809

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -539,8 +539,8 @@
 .#{$wizard}__main {
   z-index: var(--#{$wizard}__main--ZIndex);
   display: flex;
-  flex: 1 1 auto;
   flex-direction: column;
+  flex-grow: 1;
   overflow-x: hidden;
   overflow-y: auto;
   word-break: break-word;
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -538,7 +538,9 @@
 
 .#{$wizard}__main {
   z-index: var(--#{$wizard}__main--ZIndex);
+  display: flex;
   flex: 1 1 auto;
+  flex-direction: column;
   overflow-x: hidden;
   overflow-y: auto;
   word-break: break-word;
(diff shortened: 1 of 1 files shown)


mcoker:
@srambach tl;dr this is to support the drawer placed as a child of `wizard__main -> wizard__main-body.pf-m-no-padding`. Just want to make sure that adding `display: flex/column` to `__main` and `flex-grow: 1` to `wizard__main-body:last-child` doesn't break anything else.

mcoker (approved):
🎃👍

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -538,7 +538,9 @@
 
 .#{$wizard}__main {
   z-index: var(--#{$wizard}__main--ZIndex);
+  display: flex;
   flex: 1 1 auto;
```
mattnolting, line 542: I think you'll want to remove this line as default `flex` values resolve to:

```
flex-grow: 0;
flex-shrink: 1;
flex-basis: auto;
```

And all you need to update is `:last-child` if that's an agreeable solution.

mattnolting (changes requested):

src/patternfly/components/Wizard/wizard.scss
thatblindgeye: At least for the flex property that should still be needed so that the wizard main element fills whatever space to match the current behavior when a drawer is rendered inside wizard main.

@mcoker can't recall what the discussion was, but was flex-grow on the `:last-child` necessary? Believe testing these styles in my React PR omitting the flex-grow on last child doesn't change anything (though i may just not be using the combo of elements to notice a difference)

mattnolting (approved):
LPTM 🔥

Issue #6803, opened 2024-06-17
Wizard - add bottom border to wizard header


mcoker:
From https://github.com/patternfly/patternfly-react/issues/10583, looks like we're missing the bottom border on the wizard header.

cc @kaylachumley

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6820

PR #6820, opened 2024-06-23
fix(wizard): added border to wizard header

andrew-ronaldson:
Closes https://github.com/patternfly/patternfly/issues/6803


mcoker (approved):
LGTM tho one question - should we use `border--width--regular` or `border--width--divider`? And do we need to update any of these borders to match?

https://github.com/patternfly/patternfly/blob/a09e0be720e6d233d23bfdfa484a5987cff7dfa3/src/patternfly/components/Wizard/wizard.scss#L15C3-L15C90

https://github.com/patternfly/patternfly/blob/a09e0be720e6d233d23bfdfa484a5987cff7dfa3/src/patternfly/components/Wizard/wizard.scss#L96

https://github.com/patternfly/patternfly/blob/a09e0be720e6d233d23bfdfa484a5987cff7dfa3/src/patternfly/components/Wizard/wizard.scss#L131

https://github.com/patternfly/patternfly/blob/a09e0be720e6d233d23bfdfa484a5987cff7dfa3/src/patternfly/components/Wizard/wizard.scss#L171

mcoker (approved):
🐳

PR #6823, opened 2024-06-24
fix(Switch): updated a11y by removing dynamic labeling

### LINKED ISSUE #6347: Bug - Switch - Inaccessible design pattern and implementation
**Describe the problem**
The Switch component does not meet the following accessibility requirement: 
[Success Criterion 2.5.3 Label in Name](https://www.w3.org/TR/WCAG21/#label-in-name) (Level A)

This requirement states that the visible label and the accessible label should either match or have the visible label used as the beginning of the accessible label. The reason for this is that some screen reader users are sighted users, and seeing text on the screen that does not match the text they hear presents a usability issue. 

Slightly different issues are present in the react vs HTML examples, but both are problematic. 

- HTML example — The `aria-labelledby` value always references the `true` label, but then the switch state is `false`, the visible label and the accessible name do not match
- React example — The `aria-labeledby` value always references the visible label. This means that as the state is switched from `true` to `false` the label also changes to the opposite meaning. For example, the switch either says "visible is true" or "hidden is false" resulting in the false state being communicated with a double negative. 

**Expected behavior**
Since the switch is implemented using a checkbox form input to manage the selection state, it should follow the same rules that we use when using checkboxes: 

- the label should not change based on the selection state
- the label should always describe the behavior or value when the checked state is true

thatblindgeye:
Closes #6347

### DIFF
--- a/src/patternfly/components/Switch/switch-input.hbs
+++ b/src/patternfly/components/Switch/switch-input.hbs
@@ -6,8 +6,6 @@
   {{/if}}
   {{#if aria-labelledby}}
     aria-labelledby="{{aria-labelledby}}"
-  {{/if}}
-  {{#if aria-labelledby}}
   {{else if aria-label}}
     aria-label="{{aria-label}}"
   {{else}}
--- a/src/patternfly/components/Switch/switch-input.hbs
+++ b/src/patternfly/components/Switch/switch-input.hbs
@@ -1,5 +1,6 @@
 <input class="{{pfv}}switch__input{{#if switch-input--modifier}} {{switch-input--modifier}}{{/if}}"
   type="checkbox"
+  role="switch"
   {{#if switch-input--id}}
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Switch/switch-input.hbs
```diff
@@ -1,18 +1,17 @@
...
   type="checkbox"
+  role="switch"
   {{#if switch-input--id}}
     id="{{switch-input--id}}"
   {{/if}}
   {{#if aria-labelledby}}
     aria-labelledby="{{aria-labelledby}}"
   {{/if}}
   {{#if aria-labelledby}}
+  {{else if aria-label}}
```
mcoker, line 11: not sure if you meant to have these two with no output?

mcoker (approved):
L🐸TM!!!

Issue #6822, opened 2024-06-24
Design review bugs: Beta components


andrew-ronaldson:
Clean up tasks for beta.

```[tasklist]
### Tasks
- [x] Modal title not using heading font
- [x] Drawer static/inline use primary not floating background
- [x] html demo “Drawer with jump links” to “Collapsed drawer with jump links”
- [x] Drawer. update the shadow to use a medium size (currently large in design & dev)
- [x] Add semantic spacers to action items/groups (core & design)
- [x] Inline edit. Update inline edit action groups to use new semantic gap spacers
```

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6825

Issue #6756, opened 2024-06-10
Alert group spacing update


lboehling:
- Alert group should have an 8px spacer between alerts to [match design](https://www.figma.com/design/VMEX8Xg2nzhBX8rfBx53jp/PatternFly-6%3A-Components?node-id=1110-2821&t=rewJ5ITy0heuNwCC-4)
- Looks like there's an override on the inline links that removes the underline on hover
- Some alerts have 8px above the actions, others have 4px. Let's update so the spacing above the actions is always 8px.
- Update spacing between actions to new semantic global/spacer/gap/action-to-action/default token

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6825

PR #6825, opened 2024-06-25
fix(beta bugs): minor component updates

andrew-ronaldson:
Closes https://github.com/patternfly/patternfly/issues/6822 and https://github.com/patternfly/patternfly/issues/6756

Updates to:

- Drawer background and shadow
- Drawer demo title
- Modal titles
- menu toggle :active m-active mods
- Inline edit action spacers
- Empty state action group spacers
- Alerts spacers and inline links


andrew-ronaldson:
I added the some extra alert updates from https://github.com/patternfly/patternfly/issues/6756

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -310,14 +310,6 @@
     --#{$menu-toggle}__toggle-icon--Color: var(--#{$menu-toggle}--expanded__toggle-icon--Color);
   }
 
-  &:is(:active, .pf-m-active) {
```
srambach, line 313: I think all the`--active--` variables can be removed then since they aren't used? And the `--m-primary--active` and `--m-secondary--active` then as well.

src/patternfly/components/InlineEdit/inline-edit.scss
```diff
@@ -2,14 +2,14 @@
 
 :where(:root, .#{$inline-edit}) {
   // Group
-  --#{$inline-edit}__group--item--MarginInlineEnd: var(--pf-t--global--spacer--sm);
+  --#{$inline-edit}__group--item--MarginInlineEnd: var(--pf-t--global--spacer--gap--action-to-action--plain);
```
lboehling, line 5: this should just be --pf-t--global--spacer--gap--action-to-action--default

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -55,8 +55,8 @@
   --#{$empty-state}--m-xs__footer--MarginBlockStart: var(--pf-t--global--spacer--md);
 
   // Actions
-  --#{$empty-state}__actions--RowGap: var(--pf-t--global--spacer--sm);
-  --#{$empty-state}__actions--ColumnGap: var(--pf-t--global--spacer--md);
+  --#{$empty-state}__actions--RowGap: var(--pf-t--global--spacer--action--vertical--default);
```
lboehling, line 58: this token is meant for the padding within actions, should probably use a gap token here instead... maybe --pf-t--global--spacer--gap--group--vertical?
```suggestion
  --#{$empty-state}__actions--RowGap: var(--pf-t--global--spacer--gap--group--vertical);
```

lboehling (changes requested):
thanks andrew!! left a couple comments! otherwise looks good!!

srambach (approved):
🤩 Awesome!

src/patternfly/components/Alert/alert.scss
```diff
@@ -236,8 +236,6 @@
   padding-block-start: var(--#{$alert}__action-group--PaddingBlockStart);
 
   > .#{$button} {
-    --#{$button}--m-link--m-inline--hover--TextDecoration: none;
-
     &:not(:last-child) {
       margin-inline-end: var(--#{$alert}__action-group__c-button--not-last-child--MarginInlineEnd);
     }
```
mcoker, line 241: Nit but if this is just intended to hold actions, it would be a lot better as a flex layout (or using the action list component)
* it would make the spacer we're using here as a margin (`--pf-t--global--spacer--gap--action-to-action--default`) the actual space between actions, instead of what it is now, which is the spacer + a space character
    Screenshot 2024-07-01 at 4 13 45 PM
* we could add a row gap
* it will let the actions touch the right edge of the container if they wrap - they don't right now because there will always be a margin to the right of an action
* We can remove all button styling from this block

src/patternfly/components/EmptyState/empty-state.scss
```diff
@@ -55,8 +55,8 @@
   --#{$empty-state}--m-xs__footer--MarginBlockStart: var(--pf-t--global--spacer--md);
 
   // Actions
-  --#{$empty-state}__actions--RowGap: var(--pf-t--global--spacer--sm);
-  --#{$empty-state}__actions--ColumnGap: var(--pf-t--global--spacer--md);
+  --#{$empty-state}__actions--RowGap: var(--pf-t--global--spacer--gap--group--vertical);
+  --#{$empty-state}__actions--ColumnGap: var(--pf-t--global--spacer--gap--action-to-action--default);
 }
```
mcoker, line 61: I'm wondering if we should we update the empty state footer to use the gap-to-gap spacer? It's the spacer between action groups in the footer https://github.com/patternfly/patternfly/blob/7bd4dd2a1eb955b621ac3bafc46265b4d782ade8/src/patternfly/components/EmptyState/empty-state.scss#L53

What it looks like now with a small spacer. Doesn't really differentiate between the groups
Screenshot 2024-07-01 at 4 22 15 PM

And with the group-to-group spacer
Screenshot 2024-07-01 at 4 23 18 PM

If not, assuming the goal of the small spacer is to keep a tight group like the primary and other actions are part of the same group, should we also update it to use `--pf-t--global--spacer--gap--group--vertical` as you've done here?

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -34,11 +34,6 @@
   --#{$menu-toggle}--hover--BorderColor: var(--pf-t--global--border--color--hover);
   --#{$menu-toggle}--hover__toggle-icon--Color: var(--pf-t--global--icon--color--regular);
 
-  // * Menu toggle active
-  --#{$menu-toggle}--active--BackgroundColor: var(--pf-t--global--background--color--control--default);
-  --#{$menu-toggle}--active--BorderWidth: var(--pf-t--global--border--width--action--clicked);
-  --#{$menu-toggle}--active--BorderColor: var(--pf-t--global--border--color--clicked);
```
mcoker, line 40: 🥳

mcoker (approved):
LGTM! Left some nit comments, lemme know if you want to merge as-is or follow up in this PR or create an issue for later.

src/patternfly/components/EmptyState/empty-state.scss
lboehling: I like this update. Within the link button group, it should get gap--group--vertical, but if the primary action is in a different group, it should get group-to-group.

lboehling (approved):
Looks good! I just made a comment agreeing with Coker's suggestion to use a group-to-group spacer for the empty state gap between the primary and link button action groups.

mattnolting (approved):
Beautiful 🤩

kaylachumley (approved):
Looks good! I agree with lucia and coker on the group to group spacer.

Issue #6829, opened 2024-06-26
Log viewer - remove log viewer styles from core


wise-king-sullyman:
**Describe the enhancement or change**
Now that https://github.com/patternfly/react-log-viewer/pull/70 has landed log viewer will hold its styles directly and core no longer needs to do so (in v6).

PR #6831, opened 2024-06-26
fix(log-viewer): remove log viewer from core

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6829


srambach:
Do we still want the card log view demo?

mcoker:
@srambach I imagine so, I think the [card log view demo](https://patternfly-pr-6831.surge.sh/components/card/html-demos#log-view) is something different, and looks to pair with this [card events view demo](https://pf6.patternfly.org/components/card/html-demos#events-view).

srambach (approved):
Just checking! Looks good otherwise.

PR #6832, opened 2024-06-27
feat(page): add fill modifier for page container

### LINKED ISSUE #6653: Page - allow main container to fill
**Describe the feature**
In V6, Page and PageSection components don't fill the page because pf-v6-c-page__main-container does not fill the available height (nor its child pf-v6-c-page__main).

Let's give the option with a `.pf-m-fill` on the main container to fill height and also look at whether any of the other possible components that can go there also need a modifier to fill vertical space.

@jeff-phillips-18 

(clarification: the page should not fill unless we add `.pf-m-fill`)

srambach:
Closes #6653 

This changes the main container not to fill unless it either a) has `.pf-m-fill` or b) it contains a section with `.pf-m-fill` 

Note: I'm not sure we really need `.pf-m-no-fill` now but I left it because it might be useful in some instances?

Also, I removed the flex-grow on last child because that causes it to take space even if another section has fill. I don't think the last section should automatically fill. There's one case I see that might have to be adjusted:
```
pf-c-main-container.pf-m-fill
  pf-c-main-section 
  pf-c-main-section pf-m-secondary
```
Here the secondary section wouldn't fill the space, but a `.pf-m-fill` could be added to that section.
Alternatively, if we automatically fill the last section, if you didn't want it to take extra space when another section has a fill, then it would need `.pf-m-no-fill`

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -423,7 +423,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 .#{$page}__main-section,
 .#{$page}__main-group,
 .#{$page}__main-wizard {
-  &:last-child,
   &:only-child,
   &.pf-m-fill {
     flex-grow: 1;
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -324,10 +324,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   border-radius: var(--#{$page}__main-container--BorderRadius);
 
   &.pf-m-fill,
-  &:has(.#{$page}__main-section.pf-m-fill,
-    .#{$page}__main-group.pf-m-fill,
-    .#{$page}__main-wizard,
-    .#{$drawer}) {
+  &:has(.#{$page}__main-wizard) {
     align-self: stretch;
   }
 }
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -313,6 +313,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 .#{$page}__main-container {
   display: flex;
   flex-direction: column;
+  align-self: start;
   max-height: var(--#{$page}__main-container--MaxHeight);
   margin-block-start: 0;
   margin-inline-start: var(--#{$page}__main-container--MarginInlineStart);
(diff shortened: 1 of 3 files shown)


mcoker:
I wonder what we consider to be the primary use case(s) for the main section to fill the page height in v6, and if that can guide wither the last section should continue to fill by default or not? I want to say this was the result of user feedback (we tagged Jeff in https://github.com/patternfly/patternfly/issues/6653) but I don't remember the specific ask.

src/patternfly/components/Page/page.scss
```diff
@@ -321,6 +322,13 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   background: var(--#{$page}__main-container--BackgroundColor);
   border: var(--#{$page}__main-container--BorderWidth) solid var(--#{$page}__main-container--BorderColor);
   border-radius: var(--#{$page}__main-container--BorderRadius);
+
+  &.pf-m-fill,
+  &:has(.#{$page}__main-section.pf-m-fill,
+    .#{$page}__main-group.pf-m-fill,
+    .#{$page}__main-wizard.pf-m-fill) {
```
mcoker, line 329: I suspect we will want the main content to fill if:
* There is a wizard in the page body area at all (it wouldn't necessarily need `.pf-m-fill`, it would just fill by default). It was a goal in v5 for the wizard footer to remain static regardless the height of the wizard content, and the wizard designs for v6 look the same. We got that by having the wizard fill the main container area and set the wizard content to scroll.
* There is a drawer attached to the main container. In that case, I'd probably expect the `__main-container` to fill but no page sections fill by default - page sections will still look like they normally do, but the drawer panel can expand over the full height of the page content area. Assuming that's still what we want in v6, that could be a reason to not make page sections fill the height by default.

srambach: OK - I changed it so that any wizard will make the main container fill, and a drawer inside will as well - bottom drawer won't though. WDYT?

evwilkin (approved):
LGTM - thanks @srambach!

mcoker (commented):
🐽

mcoker (approved):

src/patternfly/components/Page/page.scss
```diff
@@ -415,7 +421,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 .#{$page}__main-section,
 .#{$page}__main-group,
 .#{$page}__main-wizard {
-  &:last-child,
   &:only-child,
```
mcoker, line 424: Sorry didn't catch this before, but I'm thinking we probably don't want `:only-child` to fill, just like `:last-child` - you'd need to apply `.pf-m-fill` manually if you want something to fill. wdyt @srambach?

srambach: I agree. Updated it to remove only-child as well.

Issue #6744, opened 2024-06-04
Icon component doesn't have option to use default icon color


mcoker:
By default and via the `.pf-m-inline` modifier, the icon component just inherits `color`. For v6, we have icon specific color tokens, and the base color is https://github.com/patternfly/patternfly/blob/07530abda9d48d533dbe09e2e9dde16f160c11d3/src/patternfly/base/tokens/tokens-default.scss#L423

I wonder if it would be better for the default use of the icon component (no modifier) would give you 1) the default icon color and 2) the default icon size (below), and the `.pf-m-inline` variation inherits color and size?

https://github.com/patternfly/patternfly/blob/07530abda9d48d533dbe09e2e9dde16f160c11d3/src/patternfly/base/tokens/tokens-default.scss#L612

PR #6850, opened 2024-07-03
fix(icon): update default to use icon color/size tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6744


srambach (approved):
Fabulous!

Issue #6032, opened 2023-11-03
Bug - [Password] - [Password strength breaks on horizontal forms]


KKoukiou:
**Describe the problem**
The password strength does not wrap correctly on horizontal forms.

**Screenshots**
![Screen Shot 2023-11-03 at 16 03 25](https://github.com/patternfly/patternfly/assets/14921356/d8a62088-0c12-4f74-b096-30adddce1acc)

**What is your product and what release date are you targeting?**
Anaconda

mcoker:
cc @mmenestr @andrew-ronaldson this is a result of the "additional label info" layout breaking in a horizontal form - it's only supported in the non-horizontal form currently. https://www.patternfly.org/components/forms/form#form-group-with-additional-label-info

FWIW if we just put the add'l info under the label in a horizontal form, this is what it would look like.

preview link - https://codesandbox.io/s/bold-mountain-9kygkq?file=/index.html

<img width="445" alt="Screenshot 2023-11-07 at 1 40 55 PM" src="https://github.com/patternfly/patternfly/assets/35148959/0391db6f-2bbc-4801-8aff-2f543bcbe67a">

mcoker:
cc @andrew-ronaldson @lboehling WDYT about this issue?

PR #6853, opened 2024-07-03
fix(form): support label addl info in horizontal form

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6032 (for main/v5)

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,5 +108,6 @@
   "publishConfig": {
     "registry": "https://registry.npmjs.org/",
     "access": "public"
-  }
+  },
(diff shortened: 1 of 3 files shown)


andrew-ronaldson (approved):
sweet moves, ace!

evwilkin (approved):
LPTM! 🎉

mcoker:
@srambach updated the text

Though I noticed the add'l info on a regular vs horizontal form was different, so I addressed that so they match. The horizontal form styles only apply above the medium breakpoint, and I was applying the horizontal add'l info styles at all breakpoints, so I moved them to only apply on the medium breakpoint. cc @andrew-ronaldson 

Screenshot 2024-07-09 at 1 51 36 PM

Screenshot 2024-07-09 at 1 51 34 PM

srambach (approved):

Issue #6577, opened 2024-04-25
Bug - Notification drawer - Gap and alignment


mattnolting:
**Describe the problem**
This is an issue with 
- [ ] The [menu toggle updates](https://github.com/patternfly/patternfly/pull/6566) resulted in an alignment shift in button alignment. Alignment needs to be adjusted for text and action

![Screenshot 2024-05-20 at 2 58 44 PM](https://github.com/patternfly/patternfly/assets/5385435/199d231d-768f-4a84-869a-ec9bc2a3421c)

- [ ] Layout is still using `margin-block-end` and should be updated to `row-gap` at lines [223](https://github.com/patternfly/patternfly/blob/v6/src/patternfly/components/NotificationDrawer/notification-drawer.scss#L223) and [251](https://github.com/patternfly/patternfly/blob/v6/src/patternfly/components/NotificationDrawer/notification-drawer.scss#L251)

from: https://github.com/patternfly/patternfly/pull/6566#discussion_r1578413166

PR #6854, opened 2024-07-03
fix(notification-drawer): fix action alignment, add no-offset mod

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6577


andrew-ronaldson (approved):
Stupendous!

evwilkin:
Two questions:
1. The initial alignment problem from the issue looks fixed, but is the `.pf-m-no-offset` example as intended?  The action button looks to be top-aligned with the header text, but due to the size it looks to overlap down to the line of text below:
Screenshot 2024-07-05 at 9 58 28 AM

2. Also wanted to check if the second note in the original issue description about replacing `margin-block-end` with `row-gap` needs to be addressed before that issue is closed out or if that should be disregarded?

srambach (approved):
@evwilkin it looks like it was already changed to use gap.
I agree that the example with the primary button actually looks better without the `f-m-no-offset` because then the text aligns. But the modifier is working as expected, so if needed, looks good.

image

evwilkin (approved):
Based on @srambach's notes, LGTM 👍

PR #6857, opened 2024-07-07
chore(fonts): update font names and remove extra tokens

### LINKED ISSUE #6790: Update @font-face font-family to match token format from figma
Per https://github.com/patternfly/design-tokens/pull/63#issuecomment-2165972078

Update the [`RedHat*` @font-face declarations](https://github.com/patternfly/patternfly/blob/c67594075421a8e431a942092e9eb36833afe275/src/patternfly/base/_fonts.scss#L2) to have spaces to match https://github.com/patternfly/design-tokens/blob/39cac09a6c5d71b555e83fb5f6573b536c040e7a/packages/module/build/css/tokens-default.scss#L72-L74

srambach:
Fixes #6790 

Updates font names to match tokens (include spaces)
Also removes font tokens from the local token file since they are now in the default token file from Figma.

Note: There are two local tokens for font weight that are used but not in Figma - `--pf-t--global--font--weight--body--bold` and `--pf-t--global--font--weight--heading--bold`. These point to base tokens that also don't exist. (see diff for details) @lboehling should these semantic font weight tokens exist and should they point to existing base font weigh tokens?

### DIFF
--- a/src/patternfly/base/tokens/tokens-local.scss
+++ b/src/patternfly/base/tokens/tokens-local.scss
@@ -1,15 +1,9 @@
 // LOCAL TOKENS
 
 @mixin pf-v6-tokens {
-  // Missing base font tokens
-  --pf-t--global--font--weight--body--100: 400; // not currently used
-  --pf-t--global--font--weight--body--200: 500;
-  --pf-t--global--font--weight--heading--100: 700; // not currently used
-  --pf-t--global--font--weight--heading--200: 700;
-
-  // Missing semantic font tokens
-  --pf-t--global--font--weight--body--bold: var(--pf-t--global--font--weight--body--200);
-  --pf-t--global--font--weight--heading--bold: var(--pf-t--global--font--weight--heading--200);
+  // TODO Missing semantic font tokens not getting exported from Figma properly
+  --pf-t--global--font--weight--body--bold: var(--pf-t--global--font--weight--200);
+  --pf-t--global--font--weight--heading--bold: var(--pf-t--global--font--weight--400);
 
   // TODO Maintaining these in addition to the h1-h6 sizes
   --pf-t--global--font--size--heading--xs: var(--pf-t--global--font--size--md);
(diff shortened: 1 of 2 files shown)


src/patternfly/base/tokens/tokens-local.scss
```diff
@@ -1,29 +1,14 @@
-// FONT TOKENS
-// Manually added from hackathon values
-// Currently not being exported from Figma because they are styles rather than variables
-// The ability to make font variables is reportedly coming by the end of the year
+// LOCAL TOKENS
```
mcoker, line 1: 🥸👍

lboehling:
@srambach @mcoker the bold weights tokens are in figma, I wonder why they're not exporting? 
image

srambach:
> @srambach @mcoker the bold weight tokens are in figma, I wonder why they're not exporting? image

Ah...I think it's because there's a value for `font--weight--body` but then `--bold` is nested inside. I don't think the parser picks that up. I think that's one reason we added `--default` to most things.

Here's how bold is getting nested inside the body/heading
```
      "weight": {
        "body": {
          "description": "Use to define the default weight for body text",
          "type": "number",
          "value": "{global.font.weight.100}",
          "bold": {
            "description": "Use to define the bold weight for body text, often used to field labels or to add emphasis.",
            "type": "number",
            "value": "{global.font.weight.200}"
          }
        },
        "heading": {
          "description": "Use to define the default weight for heading text",
          "type": "number",
          "value": "{global.font.weight.300}",
          "bold": {
            "description": "Use to define the bold weight for heading text, often used to add emphasis.",
            "type": "number",
            "value": "{global.font.weight.400}"
          }
        }
      },
```

IDK if @evwilkin has any knowledge of making the dictionary handle that? Or we add `--default` I guess.

srambach:
@mcoker I reassigned those missing bold tokens to the correct base tokens. However, those and the text-decoration tokens are still used in the code instead of the newer ones coming from figma. I'd like to do that as a followup.

mcoker (approved):

Issue #6771, opened 2024-06-11
Update gutter size in our layouts


lboehling:
It looks like our default/universal gutter was changed to our large spacer/24px in v6 at some point. I think we should keep it the same value as in v5 (medium spacer/16px). this will make our layouts play better with the page section padding updates that @srambach is working on. 

does this need a new token or update to existing token @mcoker?

mcoker:
@lboehling hmm... I think it depends on what we're referring to as a gutter. If we're talking about the left/right padding/inset on page chrome things that should all align masthead, vertical nav, page sections, and potentially the left/right padding on components that are intended to span the full width of the page main content area, IMO that's a separate token. We traditionally (in v5) referred to that as "page inset". You can set an "inset" on lots of things and it just modifies the left/right padding of something using our global spacers, but "page inset" always mapped to whatever the page chrome inset was.

Or are you also referring to things like the gaps in a gallery/flex/grid layout that have a "gutter"? For example the [gallery gutter example](https://pf6.patternfly.org/layouts/gallery#with-gutter) where the "gutter" is the purple gaps here

<img width="914" alt="Screenshot 2024-06-27 at 3 38 36 PM" src="https://github.com/patternfly/patternfly/assets/35148959/5576a0e5-0aba-4316-9079-32fd89075da6">

lboehling:
@mcoker yep, I'm referring to the gaps in the gallery/flex/grid layouts that have a gutter! in v5 they were set to our global gutter variable which is 16px/md spacer. I noticed it increased to 24px in v6 and also noticed it just uses a generic spacer since we don't have an equivalent global gutter token.

PR #6863, opened 2024-07-09
fix(layouts): updating gutters tokens

andrew-ronaldson:
Closes https://github.com/patternfly/patternfly/issues/6771

One question I have is the Flex layout update. Not sure if this is intended @lboehling ?

`  --#{$flex}--spacer--column--base: var(--pf-t--global--spacer--gutter--default); // default spacer/gap for columns`

That trickles down as the base for other spacers. It was using `spacer--lg` before the update.

### DIFF
--- a/src/patternfly/layouts/Flex/flex.scss
+++ b/src/patternfly/layouts/Flex/flex.scss
@@ -13,7 +13,7 @@ $pf-v6-l-flex--variable-map: build-variable-map("#{$pf-prefix}l-flex--spacer", $
   --#{$flex}--m-row--AlignItems: baseline;
   --#{$flex}--m-row-reverse--AlignItems: baseline;
   --#{$flex}--item--Order: 0;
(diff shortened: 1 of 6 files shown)


mcoker:
The flex layout has always kinda been it's own thing, though I don't know why it wouldn't match the other layouts. I give this update 👍👍 and a 🐸 for good measure.

Just some more info - the one thing that's different about the flex layout is it has the column gap (16px per this PR) and row gap (8px via `spacer--sm`) by default, and they're completely, independently adjustable. The other layouts do not have a gap/gutter unless you add the modifier `.pf-m-gutter`, and you just get the fixed gutter.

mcoker (approved):
```
(rawr)
      🦖
```

mattnolting (approved):
Well done LPTM 🐸

lboehling (approved):
nice!!! yes agree, i think flex should match the other layouts

PR #6868, opened 2024-07-09
fix(table-tree-view): reset table tree view th header cell min/max-width, width , and display

### LINKED ISSUE #6865: Table - tree table first column heading truncated
**Please describe the issue**
On a tree table, the first column's heading is truncated on narrow viewports.
In v5, this heading was omitted completely.

https://pf6.patternfly.org/components/table/#tree-table-basic

image

mattnolting:
fixes #6865

### DIFF
--- a/src/patternfly/components/Table/table-tree-view.scss
+++ b/src/patternfly/components/Table/table-tree-view.scss
@@ -56,10 +56,6 @@ $pf-v6-c-tree-view--MaxDepth: 10;
       --#{$table}__tree-view-main--PaddingInlineStart: calc(var(--#{$table}__tree-view-main--nested-indent--base) * #{$i - 1} + var(--#{$table}__tree-view-main--indent--base));
     }
   }
-
-  .#{$table}.pf-m-tree-view .#{$table}__thead {
-    display: none;
-  }
 }
 
 // - Table tree vew main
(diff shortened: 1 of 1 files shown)


mcoker:
adding @lboehling and @andrew-ronaldson. @mattnolting please fill in any missing details, but I believe you made a change to the table in a previous PR to keep the table column header visible for the first column in a tree table on mobile. This PR just fixes a visual bug with that, but I don't think design had reviewed the initial change, so I want to make sure this is how design wants to present the table.

I think the reasoning for leaving the column header visible is because, in a tree table, the first column includes the toggle (a checkbox if it's selectable) and the node/toggle/column text, so we don't move the column header down into the display for that cell like we do other cells in the row.

Some screenshots:

**Desktop view**
Screenshot 2024-07-09 at 8 41 48 PM

**Mobile view**
Screenshot 2024-07-09 at 8 40 58 PM

**Mobile view in a page**
Screenshot 2024-07-09 at 8 41 58 PM

And for comparison, here is v5 since it didn't include the header

**Desktop view**
Screenshot 2024-07-09 at 8 51 51 PM


**Mobile view**
Screenshot 2024-07-09 at 8 51 55 PM

mcoker (approved):
LPTM!

Issue #6883, opened 2024-07-12
Pull in latest font-weight tokens, update styles


mcoker:
Need to pull in and fix/sort out:
* `--pf-t--global--font--weight--[body/heading]` changed to `--pf-t--global--font--weight--[body/heading]--default`
* `--pf-t--global--font--weight--[body/heading]--bold` now comes from figma, so we can remove our local instances

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6884

PR #6884, opened 2024-07-12
fix(tokens): pull in changes to font weight tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6883

A backstop run (using the desktop viewport size) before/after these changes showed 0 differences

### DIFF
--- a/src/patternfly/base/normalize.scss
+++ b/src/patternfly/base/normalize.scss
@@ -18,7 +18,7 @@
   :where(body) {
     font-family: var(--pf-t--global--font--family--body);
     font-size: var(--pf-t--global--font--size--body--default);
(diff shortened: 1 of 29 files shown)


thatblindgeye (commented):
This looks good, only thing is should the bold token be removed from the local file?

https://github.com/patternfly/patternfly/blob/b9519a8f84eebfee1109e90d015c434ff3864644/src/patternfly/base/tokens/tokens-local.scss#L5-L6

mcoker:
@thatblindgeye d'oh!! Yep, thanks for catching that, didn't check in that change for some reason. Updated 👍

thatblindgeye (approved):

PR #6668, opened 2024-05-16
feat(MenuToggle): split action toggle button modifier

kmcfaul:
Closes #6534.

Adds `pf-m-split-toggle-button` modifier, for use when a split action checkbox toggle has toggle text (not a checkbox label).

PR #6886, opened 2024-07-15
feat(MenuToggle): split action toggle text modifier

### LINKED ISSUE #6778: Menu toggle - add pf-m-text support to split toggle button
Follow up to https://github.com/patternfly/patternfly/pull/6668 but for v6

* Add `.pf-m-text` to `.pf-v6-c-menu-toggle__button`
* We can [fix the double label issue](https://github.com/patternfly/patternfly/issues/6776) in this issue, too

cc @kmcfaul

kmcfaul:
Closes #6778.

Same changes as https://github.com/patternfly/patternfly/pull/6668 but for v6 branch.

### DIFF
--- a/src/patternfly/components/MenuToggle/examples/MenuToggle.md
+++ b/src/patternfly/components/MenuToggle/examples/MenuToggle.md
@@ -249,7 +249,7 @@ import './MenuToggle.css'
 {{/menu-toggle}}
 ```
 
-### Split button (checkbox with toggle text)
+### Split button (checkbox with label)
 ```hbs
 {{#> menu-toggle menu-toggle--id="split-button-checkbox-with-toggle-text-example" menu-toggle--IsSplitButton="true"}}
   {{> menu-toggle--check check-label--text="10 selected"}}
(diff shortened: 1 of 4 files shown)


src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -135,6 +135,7 @@
   --#{$menu-toggle}__button--PaddingInlineStart: var(--pf-t--global--spacer--sm);
   --#{$menu-toggle}__button--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$menu-toggle}__button--MinWidth: calc(var(--#{$menu-toggle}--FontSize) * var(--#{$menu-toggle}--LineHeight) + var(--#{$menu-toggle}--PaddingBlockStart) + var(--#{$menu-toggle}--PaddingBlockEnd));
+  --#{$menu-toggle}__button--m-text--PaddingInlineStart: var(--#{$pf-global}--spacer--xs);
```
evwilkin, line 138:
```suggestion
  --#{$menu-toggle}__button--m-text--PaddingInlineStart: var(--pf-t--global--spacer--xs);
```

To match rules above?  It looks like the `#{$pf-global}` isn't used here in V6, though core devs can confirm if this is needed

src/patternfly/components/MenuToggle/examples/MenuToggle.md
```diff
@@ -643,6 +682,7 @@ import './MenuToggle.css'
 | `.pf-v6-c-menu-toggle__button` | `<button>` | Initiates the menu toggle button. |
 | `.pf-m-split-button` | `.pf-v6-c-menu-toggle` | Modifies the menu toggle component for the split button variation. |
 | `.pf-m-action` | `.pf-v6-c-menu-toggle.pf-m-split-button` | Modifies the menu toggle component for the action, split button variation. |
+| `.pf-m-text` | `.pf-v5-c-menu-toggle__button` | Modifies the menu toggle component split button variation with text. |
```
evwilkin, line 685:
```suggestion
| `.pf-m-text` | `.pf-v6-c-menu-toggle__button` | Modifies the menu toggle component split button variation with text. |
```

src/patternfly/components/MenuToggle/menu-toggle.scss
kmcfaul: Oh you are correct, will update.

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -135,6 +135,7 @@
   --#{$menu-toggle}__button--PaddingInlineStart: var(--pf-t--global--spacer--sm);
   --#{$menu-toggle}__button--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$menu-toggle}__button--MinWidth: calc(var(--#{$menu-toggle}--FontSize) * var(--#{$menu-toggle}--LineHeight) + var(--#{$menu-toggle}--PaddingBlockStart) + var(--#{$menu-toggle}--PaddingBlockEnd));
+  --#{$menu-toggle}__button--m-text--PaddingInlineStart: var(--pf-t--global--spacer--xs);
```
mcoker, line 138: I think we may need to increase this spacer for v6. Here are a few ideas. @lboehling wdyt? Just for reference, this is a split button toggle with a checkbox as one thing in the toggle, then text (with the caret) as the other thing in the toggle.

What it looks like now - the check is too close to the text, it's closer than the caret so it seems like the checkbox is linked to the text

Screenshot 2024-07-16 at 8 55 29 PM

If we bumped the left padding on the text from xs to sm - it's still not terribly clear that the checkbox is not associated with the text since it's about equal distance compared to the caret.

Screenshot 2024-07-16 at 8 52 15 PM

If we bumped the spacer from xs to md - seems like too much space

Screenshot 2024-07-16 at 8 52 37 PM

A little longer of an update but the caret in menu toggle doesn't actually have a spacer around it. The caret container has a `min-width` that I think we should remove and just use spacers around the caret instead. IIRC we also want the inset on the right of the toggle (distance from the right edge of the menu toggle to the right side of the caret) to match the inset on the left side (distance from the left edge of the menu toggle to the start of the text). Assuming that's all good, here is the min-width on the caret removed, the `text-to-element` gap token between the toggle text and caret icon, and a small spacer between the check and the text. I don't know if this is any better than the first screenshot updated to use the small spacer?

Screenshot 2024-07-16 at 8 54 30 PM

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -476,6 +477,14 @@
   &:has(.#{$menu-toggle}__controls) {
     padding-inline-start: 0;
   }
+
+  &.pf-m-text {
+    --#{$menu-toggle}--m-split-button--last-child--PaddingLeft: var(--#{$menu-toggle}__button--m-text--PaddingInlineStart);
```
mcoker, line 482: `--#{$menu-toggle}--m-split-button--last-child--PaddingLeft` isn't defined or used anywhere - is there a reason we need it or can we remove it?

mcoker (commented):
Left some comments, mostly just interested in the spacing so want to get design's feedback there. If needed, we can merge this and follow up and adjust the spacing later.

src/patternfly/components/MenuToggle/menu-toggle.scss
kmcfaul: I think this must have been something in v5 that got changed/removed in v6, so will remove the line.

src/patternfly/components/MenuToggle/menu-toggle.scss
kmcfaul: Leaving as is for now pending design feedback.

srambach (approved):
I agree - spacing adjustments can be a follow up. Everything else looks 👍

src/patternfly/components/MenuToggle/menu-toggle.scss
mcoker: Per design, we can update this to a small spacer for now, and design is going to look at this type of layout and see if there are further improvements. Design issue is - https://github.com/patternfly/patternfly-design/issues/1320

```suggestion
  --#{$menu-toggle}__button--m-text--PaddingInlineStart: var(--pf-t--global--spacer--sm);
```

mcoker (changes requested):
Just one change from our convo w/ design

srambach (approved):
👍

mcoker (approved):

PR #6893, opened 2024-07-16
fix(toolbar): changed toolbar `inline-padding` default to zero, updated toolbar demo examples

### LINKED ISSUE #6889: Toolbar default inset
In reviewing the visual regression screenshots, we noticed the toolbar has default left/right padding that keeps the first/last items in the toolbar from aligning with the left/right edge of content above/below it.

At a glance, this was the case not only in the masthead (the first item should start at the same point that the page content starts, and the last item should be flush right with the edge of the main content below) but in the places we checked where the toolbar is above tables and data-lists.

mattnolting:
closes #6889 

* updated `toolbar-group--modifier="pf-m-icon-button-group"` to `toolbar-group--IsActionGroupPlain=true`
* updated `toolbar-group--modifier="pf-m-filter-group"` to `toolbar--IsFilterGroup=true`
* updated `toolbar-group--modifier="pf-m-icon-button-group"` to `toolbar-group--IsActionGroupPlain=true`
* removed `page-main-section--modifier='pf-m-no-padding` from demos/examples
* updated demo padding
* set toolbar padding inline to zero

### DIFF
--- a/src/patternfly/components/Masthead/masthead.scss
+++ b/src/patternfly/components/Masthead/masthead.scss
@@ -74,7 +74,7 @@ $pf-v6-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
 .#{$masthead} {
   .#{$toolbar} {
     --#{$toolbar}--Width: var(--#{$masthead}--c-toolbar--Width);
-    --#{$toolbar}__content--PaddingBlock: var(--#{$masthead}--c-toolbar--PaddingBlock);
+    --#{$toolbar}--PaddingBlockEnd: 0;
     --#{$toolbar}__content--PaddingInline: var(--#{$masthead}--c-toolbar--PaddingInline);
     --#{$toolbar}__content--MinWidth: 0;
   }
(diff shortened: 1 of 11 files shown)


mcoker:
Backstop report - https://drive.google.com/file/d/1bBRkLeCs9K7PyNbOc0M5xSWy6YWlJRwP/view?usp=sharing

mcoker (approved):
LGTM!

mcoker:
@mattnolting there are 2 conflicts. I resolved them locally (to run the screenshots) by checking in your changes for `src/patternfly/demos/Masthead/masthead-template-content-icon-group.hbs` though the lines look identical so I don't think it matters, and checking in the existing v6 code for `src/patternfly/demos/Toolbar/toolbar-template.hbs` because v6 has `{{#> toolbar-group toolbar-group--IsActionGroupPlain=true toolbar-group--modifier="pf-m-align-end"}}` and yours is missing `toolbar-group--IsActionGroupPlain=true`, even though the toolbar items in that group are plain buttons.

srambach (approved):
👍 looks good!

PR #6897, opened 2024-07-17
fix(page): update page section padding

mcoker:
Here's the backstop report. 5 failures, the multiple file upload ones look like false positives. I don't see a difference in the first one, and the second one looks like the RH font didn't fully load before taking the screenshot.

[BackstopJS Report.pdf](https://github.com/user-attachments/files/16273015/BackstopJS.Report.pdf)


mattnolting (approved):
LGTM. I can't find more references to `page-main-nav` in the codebase. 🚢

srambach (approved):
🙉 🥝

Issue #6801, opened 2024-06-17
Page - section grouping follow-ups


mcoker:
re: https://github.com/patternfly/patternfly/pull/6766

* ~~https://github.com/patternfly/patternfly/pull/6766#discussion_r1640385195~~
  * ~~Remove any references for `__main-list`~~
* ~~https://github.com/patternfly/patternfly/pull/6766#discussion_r1640387344~~
  * ~~Remove any references for `__main-nav`, and/or any sort of tertiary nav stuff (deprecated && removed)~~
* https://github.com/patternfly/patternfly/pull/6766#pullrequestreview-2119397104 and https://github.com/patternfly/patternfly/pull/6766#discussion_r1640028898
  * Support`__main-body` as a child for all of the page sections except `__main-group`
  * Probably just support `pf-m-limit-width` on all, too, except `__main-width`? I think there are only a couple of elements that don't support it, and they could for consistency.
  * ~~Try keeping padding on the outer `__main-[section type]` and not moving it down into `__main-body`. Should allow us to remove some CSS that's in there currently.~~

PR #6902, opened 2024-07-17
feat(page): remove deprecated main list, main nav, tertiary nav

srambach:
Removes __main-list and __main-nav (along with deprecated tertiary nav)
Partially satisfies #6801


mcoker (approved):
L🍌TM!

PR #6926, opened 2024-07-25
fix(Table): favorite icon styles

### LINKED ISSUE #6925: Table favorite - favorite icon button missing styles
Favorite icons (stars) are not yellow even when clicked (when `.pf-m-favorited` is applied)

Screenshot 2024-07-25 at 16 40 40

https://staging-v6.patternfly.org/components/table/html/#favorites-examples

adamviktora:
Closes  #6925

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -786,10 +786,10 @@
   }
 
   &.pf-m-favorited .#{$button} {
-    --#{$button}--m-plain--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
+    --#{$button}--m-plain__icon--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
 
     &:is(:hover, :focus) {
-      --#{$button}--m-plain--hover--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
+      --#{$button}--hover__icon--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
     }
   }
 }


src/patternfly/components/Table/table.scss
```diff
@@ -787,9 +787,11 @@
 
   &.pf-m-favorited .#{$button} {
     --#{$button}--m-plain--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
+    --#{$button}--m-plain__icon--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
 
     &:is(:hover, :focus) {
       --#{$button}--m-plain--hover--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
+      --#{$button}--hover__icon--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
```
mcoker, line 794: Nice!! You should be able to just replace the existing vars with yours - can you test this and see if it works as expected? The plain icon in this case only has an icon in it, so we should only need to set the `__icon` vars.

```suggestion
    --#{$button}--m-plain__icon--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);

    &:is(:hover, :focus) {
      --#{$button}--hover__icon--Color: var(--#{$table}__favorite--m-favorited--c-button--Color);
```

adamviktora: Yes, works fine even when replaced

mcoker (approved):
LPTM!

PR #6944, opened 2024-07-31
chore(css): minify compiled css

### LINKED ISSUE #6941: chore(workspace): minify css in workspace

mattnolting:
closes #6941 

CSS is currently un-minified in Core workspace, resulting in lagging pageload speed. This update minifies workspace CSS and improves pageload speed for development.

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -3,7 +3,7 @@ import path from 'path';
 import gulp from'gulp';
 import rimraf from'rimraf';
 import { copyFA, copySource, copyAssets, copyDocs, watchCopyDocs } from'./scripts/gulp/copy.mjs';
-import { compileSASS, minifyCSS, watchSASS } from'./scripts/gulp/sass.mjs';
+import { compileSASS, minifyPF, minifyPFAddons, watchSASS } from'./scripts/gulp/sass.mjs';
 import { pfIconFont as definedPfIconFont, pfIcons as definedPfIcons } from'./scripts/gulp/icons.mjs';
 import { compileHBS, compileMD, watchHBS, watchMD, watchHelpers } from'./scripts/gulp/html.mjs';
 import { generateSnippets } from'./scripts/gulp/snippets.mjs';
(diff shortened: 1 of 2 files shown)


gulpfile.mjs
```diff
@@ -104,10 +104,11 @@ async function startWebpackDevServer() {
 
 const buildSrc = parallel(compileSrcSASS, series(compileSrcHBS, compileSrcMD));
 const buildDocs = series(buildSrc, copyDocs);
+console.log(buildDocs);
```
nicolethoen, line 107: Should clean up the left over console.log statement

nicolethoen (commented):
I have a vague memory of css minification causing css errors in product. I would have been years ago and I don't remember the details. I'm just calling it out here in case we start getting notes about weird behaviors to remember that there could be a chance those sorts of issues could come up again.

srambach (commented):
If I'm understanding this right, this adds in minifying PatternFly add-ons, and both that file and the main PatternFly.css file are written to dist - whereas before we wrote the minified file as a separate.min.css. Is that correct?

I'm curious as to the magnitude of the size difference and whether it's worth risking potential problems mentioned by @nicolethoen

mcoker:
Inline with @nicolethoen's comment, while this type of change may be fine,  I'd be concerned with what kinds of issues users could have, particularly since we've already released beta. There are a lot of unknowns I think we'd want to clarify.

* Do we know what kinds of issues users had before?
* Could this interfere with any other post-processing users run our css through?
* Why not leave non-minified CSS?
* Assuming this change is intended to apply to PF's shipped CSS, is there a plan for the various ways users import our CSS? Like the basic react setup of the modified base styles grabbed from core, and the import of individual component styles.

That said, the issue title is "chore(workspace): minify css in workspace" - is the goal just to update the CSS in the workspace? If so, that could be done just for docs framework. I don't see an issue with adding `patternfly-addons.min.css` (but leaving the existing stylesheet), and updating the docs framework css imports to reference the minified versions.

@nicolethoen I'm not sure if this is the same issue you are thinking of, but at some point I think cssnano had some kind of issue minifying certain combos of css properties used in other css properties, and we had to rewrite some of our CSS rules to get around it.

nicolethoen:
> I'm not sure if this is the same issue you are thinking of, but at some point I think cssnano had some kind of issue minifying certain combos of css properties used in other css properties, and we had to rewrite some of our CSS rules to get around it.

yep that sounds like it! cssnano sounds like the tool we wrestled with and we definitely needed workarounds back then

PR #6945, opened 2024-08-05
fix(TextInputGroup): updated inline padding of icons

### LINKED ISSUE #6930: Text input group inset bugs with icons and validation
We updated the menu toggle's inline padding, but icons on the left still start at the old spacer and should probably be updated. Maybe as simple as updating the `__icon`'s `inset-inline-start` to `var(--pf-t--global--spacer--control--horizontal--default)`

Also validation icons at the end and the close button may need to be adjusted? cc @andrew-ronaldson @kaylachumley @lboehling 
* Per @andrew-ronaldson in working session, the icons at the end should also have the same inset as everything at the start.

Screenshot 2024-07-26 at 2 28 13 PM

thatblindgeye:
Closes #6930 

Couple of notes;
- The gap between the status icon and utilities is slightly larger now.
  This PR: 
  image
  
  Staging:
  image

- Did we want padding at the end of the utilities, to create more of a visual gap? Right now when the close button is hovered there's not much of a gap:
  image

  But if we used the same spacing as the icons it may create too much of a visual gap when the button isn't focused:
  image

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -21,6 +21,7 @@
   --#{$text-input-group}__main--first-child--not--text-input--MarginInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--plain) / 2);
   --#{$text-input-group}__main--m-icon__text-input--PaddingInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
   --#{$text-input-group}--status__text-input--PaddingInlineEnd: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
+  --#{$text-input-group}--utilities--status__text-input--PaddingInlineEnd: calc(var(--pf-t--global--spacer--sm) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
   --#{$text-input-group}__main--RowGap: var(--pf-t--global--spacer--xs);
   --#{$text-input-group}__main--ColumnGap: var(--pf-t--global--spacer--xs);
 
@@ -139,6 +140,7 @@
 
   &:has(> .#{$text-input-group}__utilities) {
     --#{$text-input-group}__icon--m-status--InsetInlineEnd: var(--#{$text-input-group}--utilities--icon--m-status--InsetInlineEnd);
+    --#{$text-input-group}--status__text-input--PaddingInlineEnd: var(--#{$text-input-group}--utilities--status__text-input--PaddingInlineEnd);
   }
 }
 
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -20,7 +20,7 @@
   // Main
   --#{$text-input-group}__main--first-child--not--text-input--MarginInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--plain) / 2);
   --#{$text-input-group}__main--m-icon__text-input--PaddingInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
-  --#{$text-input-group}--status__text-input--PaddingInlineEnd: var(--pf-t--global--spacer--xl);
+  --#{$text-input-group}--status__text-input--PaddingInlineEnd: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
   --#{$text-input-group}__main--RowGap: var(--pf-t--global--spacer--xs);
   --#{$text-input-group}__main--ColumnGap: var(--pf-t--global--spacer--xs);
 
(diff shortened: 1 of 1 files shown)


mcoker:
Here's a screenshot of some text input groups with the utilities/close icon inset by the control/horizontal spacer

Screenshot 2024-08-05 at 10 40 46 AM

And here is the same, but with the utilities/close offset by the close button's horizontal padding, pushing it back to the right by the close button's right padding, which will make the close button's icon ("x") visually inset to the control/horizontal spacer, but on hover, you can see the actual button's right edge is closer to the right edge of the text input group

Screenshot 2024-08-05 at 10 42 00 AM

mcoker:
From 8.5.24 working session w/ @lboehling 
* Close button is fine on the right edge, @lboehling likes the hover/focus background to be flush with the right edge of the TIG. Assuming plain actions will always be in utilities (as our recommendation anyways), that should work fine
* The text-input right padding when there is a status icon is [set to the 'xl' spacer](https://github.com/patternfly/patternfly/blob/031ffb660f1b663d72d01b7de3060d702c058484/src/patternfly/components/TextInputGroup/text-input-group.scss#L23). We should update that to match the left padding when there is an icon, if possible, which would be something like `--pf-t--[spacer-control-horizontal` + `icon-size` + `text-to-element` like this https://github.com/patternfly/patternfly/blob/031ffb660f1b663d72d01b7de3060d702c058484/src/patternfly/components/TextInputGroup/text-input-group.scss#L22
* If the TIG has utilities, we should reduce the right padding of the text-input to something reasonable to separate text/icons/whatever in the text input from the utilities to the right of it. Probably a small spacer would work well?

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -49,8 +49,9 @@
 
   // Icon
   --#{$text-input-group}__icon--FontSize: var(--pf-t--global--icon--size--font--body--default);
-  --#{$text-input-group}__icon--InsetInlineStart: var(--pf-t--global--spacer--sm);
-  --#{$text-input-group}__icon--m-status--InsetInlineEnd: var(--pf-t--global--spacer--sm);
+  --#{$text-input-group}__icon--InsetInlineStart: var(--pf-t--global--spacer--control--horizontal--default);
+  --#{$text-input-group}__icon--m-status--InsetInlineEnd: var(--pf-t--global--spacer--control--horizontal--default);
+  --#{$text-input-group}__c-utilities__icon--m-status--InsetInlineEnd: var(--pf-t--global--spacer--control--horizontal--compact);
```
mcoker, line 54: Just a couple of nits on this
* The syntax for this var is flexible since it's a pretty uncommon pattern. Though typically we reserve the `c-[whatever]` naming when we reference another component (eg `c-button`). We'll use `--[component/BEM element]` (eg `--c-button` or `--utilities`) as a general identifier (not necessarily a child or direct child, just in there somewhere) so I wonder if the suggestion below is better?
* Also not sure if `--pf-t--global--spacer--control--horizontal--compact)` is the right spacer. Reason being, if a person themed their compact control padding, I wouldn't expect this spacing to update? Since it doesn't necessarily need to conform to an outer button/control padding to match other controls, I think we could just use the `--spacer--sm` token here. WDYT?

```suggestion
  --#{$text-input-group}--utilities--icon--m-status--InsetInlineEnd: var(--pf-t--global--spacer--control--horizontal--compact);
```

mcoker (commented):
Everything looks great! Just a couple of nits on the naming and token used in one spot. Lemme know what you think!

andrew-ronaldson:
Screenshot 2024-08-12 at 10 20 19 AM
Noticed this example has a broken status icon. When I inspect the warning example above it has an inset: var(--pf-v6-c-text-input-group__icon--m-status--InsetInlineEnd) but it says that same var is undefined on the error example

thatblindgeye:
@andrew-ronaldson should be fixed now, forgot to update the other instance of the CSS var that was renamed in the latest update

andrew-ronaldson (approved):
sweet moves, ace!

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -20,7 +20,7 @@
   // Main
   --#{$text-input-group}__main--first-child--not--text-input--MarginInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--plain) / 2);
   --#{$text-input-group}__main--m-icon__text-input--PaddingInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
-  --#{$text-input-group}--status__text-input--PaddingInlineEnd: var(--pf-t--global--spacer--xl);
+  --#{$text-input-group}--status__text-input--PaddingInlineEnd: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--pf-v6-c-text-input-group__icon--FontSize) + var(--pf-t--global--spacer--gap--text-to-element--default));
```
mcoker, line 23: This is awesome. But when the close button is present, we adjust the status button's inset from `--pf-t--global--spacer--control--horizontal--default` to `--pf-t--global--spacer--sm` with the other change in this PR. That leaves a little too much padding between the status icon and text in the text input.

Screenshot 2024-08-12 at 9 34 31 PM

WDYT about creating a variation of this var that swaps `--pf-t--global--spacer--control--horizontal--default` with `--pf-t--global--spacer--sm`? Then it looks like this, with the visual spacer being the text-to-element spacer:

Screenshot 2024-08-12 at 9 31 54 PM

mcoker (commented):
Looks awesome, just one nit comment that I think should be pretty straight forward?

src/patternfly/components/TextInputGroup/text-input-group.scss
thatblindgeye: Pushed, lemme know if that's what you had in mind. FWIW updating the existing var so that it starts as `calc(var(--#{$text-input-group}__icon--m-status--InsetInlineEnd)...` also worked, though I imagine if a consumer is updating that specific var we wouldn't want that to affect the spacing between status icon and text input.

mcoker (approved):
NICE!

PR #6947, opened 2024-08-05
feat(charts): remove deleted red tokens

kmcfaul:
Towards: https://github.com/patternfly/patternfly-react/issues/10582

The red charts tokens were removed for v6, the red-orange tokens are now meant to be used for danger statuses in charts.


mcoker (commented):
Nice! Looks like we're referencing one more token that doesn't exist -  `--pf-t--chart--global--label--stroke--width`. Would you mind also updating https://github.com/patternfly/patternfly/blob/18b788373565ab5a5c6e8f84eb2f00331fa4ed0f/src/patternfly/patternfly-charts.scss#L98 to 

```
--#{$chart}-global--label--stroke--Width: 0;
```

mcoker (approved):
Beautiful! Thanks @kmcfaul

Issue #6948, opened 2024-08-05
Card - update clickable variations


mcoker:
* Rename clickable to actionable - just match react
* Update clickable (aka actionable) to use `<button>` for actionable, and an `<a>` for when it's a link

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6949

PR #6949, opened 2024-08-06
fix(card): update clickable to use button and anchor

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6948

Ran visual regressions and only the new/renamed examples failed (lack of a reference image to compare against)

Changes:
* Updates the selectors in the existing selectable/clickable blocks to group the radio/check stuff in `:is()` so they're in a single selector instead of multiple (should be no visual change from that)
* Adds the new button/link element class (`.#{$card}__clickable-action`) in the `:is()` groups where it's needed, just a couple of them

To test this in a local react dev environment:
* Check this branch out in a clone of the core repo somewhere
* Run `yarn build-patternfly`
  * This builds the dist package that we put on npm for `@patternfly/patternfly` at `./dist` in the core repo
* Rename the existing `@patternfly/patternfly` dir in the react repo where your dev server is running
  * assuming you're at the root of the patternfly-react repo, run `mv node_modules/\@patternfly/patternfly node_modules/\@patternfly/patternfly.bak`
* Create a symlink of the core/dist dir to `node_modules/@patternfly/patternfly` in your react repo
  * Assuming you're still at the root of the patternfly-react repo, and the core repo is a sibling of the patternfly-react repo, run `ln -s ../patternfly/dist node_modules/\@patternfly/patternfly`. You can also specify the full paths to those dirs if that's more clear, like `ln -s /path/to/the/core-repo/dist /path/to/the/patternfly-react-repo/node_modules/\@patternfly/patternfly`

You may need to restart the react dev server for these changes to take effect? If you want to make modifications to the core CSS, you can just edit files in the core repo (currently checked out to this PR's branch), and when you're done, just run `yarn build-patternfly`, which will publish the changes to the ./dist dir. When that's done, the react dev server should auto-reload with the core changes.

### DIFF
--- a/src/patternfly/components/Card/card--actionable.hbs
+++ b/src/patternfly/components/Card/card--actionable.hbs
@@ -1,5 +1,12 @@
 {{#if card--actionable--IsLink}}
-  <a class="{{pfv}}card__clickable-action{{#if card--actionable--IsDisabled}} pf-m-disabled{{/if}}" {{#if card--actionable--IsDisabled}}tabindex="-1"{{/if}} href="{{ternary card--actionable--href card--actionable--href '#'}}"></a>
+  <a class="{{pfv}}card__clickable-action
+    {{~#if card--actionable--IsDisabled}} pf-m-disabled{{/if}}"
+    {{#if card--actionable--IsDisabled}}tabindex="-1"{{/if}}
+    aria-labelledby="{{card--id}}"
+    href="{{ternary card--actionable--href card--actionable--href '#'}}"></a>
 {{else}}
-  <button {{#if card--actionable--IsDisabled}}disabled{{/if}} type="button" class="{{pfv}}card__clickable-action"></button>
+  <button class="{{pfv}}card__clickable-action"
+    type="button"
+    aria-labelledby="{{card--id}}"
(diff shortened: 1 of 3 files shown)


thatblindgeye (commented):
Will try linking the content of this PR to React locally shortly, just wanted to leave  a couple quick comments. In addition to the file comment below, it looks like the focus styling is a bit odd.

When focusing the actionable (button) card, the corners have that faint "being overlapped" effect we've seen elsewhere. In light theme:

image

Looks worse in dark theme:

image


The actionable (link) cards don't seem to have any focus styling, though. Lastly looks like the "Actionable secondary style" example would need the update to use buttons or anchors.

thatblindgeye (approved):

srambach:
Just a question - on https://patternfly-pr-6949.surge.sh/components/card/#actionable-button the current (clicked) card changes border color on hover, whereas the other selected variants do not. I'm not sure which is more correct; I think we did choose not to change the selected hover color but maybe it's better to.

mcoker:
Nice catch @srambach! Updated 🐸👍

srambach (approved):
Hover looks good! 🎈

Issue #6951, opened 2024-08-07
Content - allow for font-size (and other customization) on individual elements


mcoker:
Follow up from: https://github.com/patternfly/patternfly-org/issues/4169#issuecomment-2269656181

The `.{$content}--[element]` classes currently don't respond to customizing `--#{$content}--FontSize`, but should.  And we can add a new/beta "editorial" modifier that will change the default font-size to the large font-size token.

mcoker:
fixed by https://github.com/patternfly/patternfly/pull/6952

PR #6952, opened 2024-08-07
fix(content): added editorial styling

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6951

Adds `font-size/line-height/color` to all of the `.pf-v6-c-content--[element]` classes, since they were previously only set on `.pf-v6-c-content` before.
Adds `.pf-m-editorial` as a modifier that can go on on `.pf-v6-c-content` or `.pf-v6-c-content--[element]` and
* Changes the default font-size from `body--default` to `body--lg`
* Changes the `<small>` from `body--sm` to `body--default`


srambach:
This looks 👌

lboehling (approved):
amazing!! I'm going to open up a follow-up design guidelines issue to add some more guidance on when/where to use editorial! thank you!!

Issue #6907, opened 2024-07-18
Review banner link color/styles


mcoker:
We're overriding styles for regular links and inline-link buttons in the banner component, and I'm wondering if we need to continue to for v6 and if there are any changes we need to make. cc @lboehling @andrew-ronaldson

andrew-ronaldson:
@andrew-ronaldson

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6953

Issue #6871, opened 2024-07-10
Bug - AboutDialog - Content gets wrapped when it doesn't need to be


jeff-phillips-18:
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  

Using the `AboutDialog` with long text and long labels for the items causes wrapping but leaves ample blank space to the right which could be used to limit the amount of text being wrapped

**How do you reproduce the problem?**
Add lots of text in the text content. Add TextListItems with very long labels

**Expected behavior**
Text should wrap when necessary but use the full width of the about dialog as necessary

**Is this issue blocking you?**
Yes without using a css override

**Screenshots**
![image](https://github.com/patternfly/patternfly/assets/11633780/89669ddc-e5d5-4641-84d9-d0fbfa2cc734)

**What is your product and what release date are you targeting?**
RHOAI

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6953

jeff-phillips-18:
This is still an issue

Issue #6729, opened 2024-05-31
Bug - Timestamp - Font color & size don't use the default variables


jgiardino:
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Please give a clear and concise description of the problem. Which components are affected?

**How do you reproduce the problem?**
It looks like the example in the design document shows the contents using the default font color, but in the HTML and React components, the font color is gray. The font size is also small rather than the default medium.

**Expected behavior**
I would expect contents to use the default font color and size to be applied by default, and to allow consumers to decide if different styles are needed for their specific designs.

**Is this issue blocking you?**
The difference in presentation results in confusion within our team. Our designs are currently reflecting what's shown in the react/html examples. If this isn't expected from a PF design perspective, it's possible that the end result would not match what PF design would recommend. 

**Screenshots**
![image](https://github.com/patternfly/patternfly/assets/21063328/d692837c-78a4-4611-8b7f-52954d14c93a)

**What is your product and what release date are you targeting?**
RHOAI, June 17

mcoker:
cc @andrew-ronaldson @lboehling

andrew-ronaldson:
@mcoker want me to take a look at this in core?

mcoker:
@andrew-ronaldson yes sir, thank you!

mcoker:
Closed by https://github.com/patternfly/patternfly/pull/6953

PR #6953, opened 2024-08-08
fix(components): various fixes for banner, timestamp, and content dl

andrew-ronaldson:
Closes https://github.com/patternfly/patternfly/issues/6907, https://github.com/patternfly/patternfly/issues/6729 and https://github.com/patternfly/patternfly/issues/6871

### DIFF
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -159,14 +159,10 @@
 
   .#{$button}.pf-m-inline {
     --#{$button}--m-link--Color: var(--#{$banner}--link--Color);
-    --#{$button}--m-link--m-inline--hover--Color: var(--#{$banner}--link--hover--Color);
+    --#{$button}--m-link--hover--Color: var(--#{$banner}--link--hover--Color);
     --#{$button}--disabled--Color: var(--#{$banner}--link--disabled--Color);
 
     text-decoration: var(--#{$banner}--link--TextDecoration);
-
-    &:hover {
-      --#{$banner}--link--Color: var(--#{$banner}--link--hover--Color);
-    }
 
     &:disabled,
     &.pf-m-disabled {
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -20,9 +20,7 @@
   --#{$banner}--link--Color: var(--#{$banner}--Color);
   --#{$banner}--link--TextDecoration: underline;
   --#{$banner}--link--hover--Color: var(--#{$banner}--Color);
-  --#{$banner}--link--hover--FontWeight: var(--pf-t--global--font--weight--body--bold);
-  --#{$banner}--link--disabled--Color: var(--#{$banner}--Color);
-  --#{$banner}--link--disabled--TextDecoration: none;
+  --#{$banner}--link--disabled--Color: var(--pf-t--global--text--color--disabled);
 
   // modifier variables
   --#{$banner}--m-sticky--ZIndex: var(--pf-t--global--z-index--md);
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Content/content.scss
```diff
@@ -327,7 +327,7 @@
     grid-template-columns: 1fr;
 
     @media screen and (min-width: $pf-v6-global--breakpoint--sm) {
-      grid-template: auto / auto 1fr;
+      grid-template: auto / 12ch 1fr;
       grid-row-gap: var(--#{$content}--dl--RowGap);
       grid-column-gap: var(--#{$content}--dl--ColumnGap);
```
mcoker, line 332: The column gap for `<dl>`'s is 2xl and the row gap is md. Do we want to update that? Some thoughts:
* In figma, the description list default column gap is sm, and row gap is lg. The compact description list column gap is sm, and row gap is md
* Currently our description list component's column and row gaps match what's in figma
* We have gap spacer tokens here - https://github.com/patternfly/patternfly/blob/bdd2cc6ea5d1d66773e74935441880783c121df1/src/patternfly/base/tokens/tokens-default.scss#L467-L474
  * For the default `<dl>` we could use `gap--text-to-element` and `gap--group-to-group--vertical`? I'm not 100% those are the right tokens though.
  * There are no compact variants of gap tokens.
  * If we update to gap spacer tokens, should we also update the description list component?

src/patternfly/components/Banner/banner.scss
```diff
@@ -20,9 +20,7 @@
   --#{$banner}--link--Color: var(--#{$banner}--Color);
```
mcoker, line 20: If you hover over the "inline link button" link in the blue banner, the text color changes to blue.

I think you can just make these changes:
* Remove this `:hover` block for inline buttons - we should be able to handle this with the next point, and just set button vars. https://github.com/patternfly/patternfly/blob/728f578a7f74398df41f6920e1e4ae28976bf0b8/src/patternfly/components/Banner/banner.scss#L167-L169
* On this line, change `--#{$button}--m-link--m-inline--hover--Color` to `--#{$button}--m-link--hover--Color`. The previous var no longer exists, and this should set the button's hover color - https://github.com/patternfly/patternfly/blob/728f578a7f74398df41f6920e1e4ae28976bf0b8/src/patternfly/components/Banner/banner.scss#L162

src/patternfly/components/Timestamp/timestamp.scss
```diff
@@ -2,13 +2,14 @@
 
 :where(:root, .#{$timestamp}) {
```
mcoker, line 3: Super nit, would you mind adding a new timestamp to the existing list (or whatever you think is best) that shows the help text styling? Doesn't look like we have an example of that anywhere in the core docs.

You can get it by adding `timestamp--IsHelp=true` to the outer `{{#> timestamp}}` so it's `{{#> timestamp timestamp--IsHelp=true}}`

mcoker (commented):
Looks good! Just a wee bit of feedback, nothing is blocking though if you'd prefer merge this as-is and follow up. Just lemme know if you'd prefer do that.

mcoker (commented):
LGTM! Just want to confirm that we don't want to update the column/row gap for the content component `<dl>`? IMO it looks like there is too much space between term and description. Here's what it is currently:

Screenshot 2024-08-19 at 6 54 41 PM

And here's the [horizontal description list component](https://patternfly-pr-6953.surge.sh/components/description-list/#horizontal)'s column/row gap:

Screenshot 2024-08-19 at 6 54 59 PM
Screenshot 2024-08-19 at 6 54 53 PM

mcoker (approved):
LPTM! Backstop report if you're interested. [zandrew-the-true-king-of-the-north.psd.v2-final.ammend.client-approved.pdf](https://github.com/user-attachments/files/16716845/zandrew-the-true-king-of-the-north.psd.v2-final.ammend.client-approved.pdf)

Some notes about the report:

* Skip to content diff is just spacing from the content component `<dl>` on that example
* Label group diff - font (not)awesome didn't load
* Card with the missing kebab - icon font didn't load
* Wizard example with missing reference file - I think that's my oopsie from a previous visual regression report

srambach (approved):
✨ looks great!

PR #6956, opened 2024-08-08
chore(backstop): update screenshots, urls

mcoker:
We reviewed the errors as a group, all were from updates to menu toggle text.

Updated the list of URLs after that, the only new URLs added were with menu-toggle (9 of them, renamed titles so the URLs changed) so all of those are in this report [backstop-8.8.24.pdf](https://github.com/user-attachments/files/16553192/backstop-8.8.24.pdf)


srambach (approved):
🦝

mattnolting (approved):
LPTM 👍

Issue #6960, opened 2024-08-12
Page sidebar - add reduced motion default transition


srambach:
**Describe the enhancement or change**
The sidebar uses a slide transition, but this is a big enough area sliding that we should create a default that does not slide for prefers-reduced-motion.

Issue #6959, opened 2024-08-12
Animation: change hard coded transitions to use motion tokens


srambach:
**Describe the enhancement or change**
There are hard coded transitions in: accordion, clipboard copy, data list, expandable section, notification drawer, table grid, tabs, and switch.
These should be updated to use motion tokens.

```[tasklist]
### Tasks
- [ ] Clipboard copy
- [ ] Data list
- [ ] Table grid
```

PR #6961, opened 2024-08-12
feat(motion): replace old local animation tokens with new motion tokens

### LINKED ISSUE #6936: Clean up old transitions
**Please describe the issue**
We had a few pre-existing animations (transitions) on a few elements. We need to clean these up so that they use animation tokens and are correctly targeted not to affect things that are not intended.

E.g. An old sidebar transition is causing the sidebar to change between light/dark modes slowly when if really should just be targeted at sliding in/out of view.

Anything using [these variables](https://github.com/patternfly/patternfly/blob/9df42bd68daf4766ec26b32f4054a83f0af45d73/src/patternfly/base/tokens/tokens-local.scss#L118) is a candidate for updating

srambach:
This removes the old local tokens for animation and replaces them with motion tokens.

There is a follow-on issue to [replace hard-coded transitions with tokens](https://github.com/patternfly/patternfly/issues/6959).

I've also put a [TODO in sidebar and drilldown menu](https://github.com/patternfly/patternfly/issues/6960) because we need to add a default for prefers-reduced-motion since these two are using a slide.

Fixes #6936


kaylachumley (commented):
LGTM!

mattnolting (commented):
These updates look great! I'm noticing that `page` and `wizard` transitions don't seem to have an affect. Are you seeing the same?

srambach:
> These updates look great! I'm noticing that `page` and `wizard` transitions don't seem to have an affect. Are you seeing the same?

I see the sidebar slide on the page, and on the wizard, the caret rotates. Those are the only ones on those components, right? What are you expecting to see?

mattnolting:
> I see the sidebar slide on the page, and on the wizard, the caret rotates. Those are the only ones on those components, right? What are you expecting to see?

I was looking for the sidebar itself to transition, which is the wrong thing to be looking at 😆

mattnolting (approved):
Love it, looks great! 🎉

mcoker (approved):
🦧🍌

Issue #6962, opened 2024-08-12
Menu toggle - add support for placeholder text


mcoker:
When used as a select toggle, we need to support placeholder styling. Here's the deprecated select component example with `.pf-m-placeholder` on the toggle - https://www.patternfly.org/components/menus/select/html-deprecated#placeholder-collapsed

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6963

PR #6963, opened 2024-08-12
fix(menu-toggle): added support for placeholder, fixed disabled icon color

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6962

* adds `.pf-m-placeholder` on the menu toggle, which changes the text color to the placeholder color token
* Icon colors were mostly set to `inherit` (matched text color), so this wires up/fixes vars to use icon color tokens 

Backstop report has no failures other than this random one, which looks like a viewport size/timing glitch
[BackstopJS Report.pdf](https://github.com/user-attachments/files/16606725/BackstopJS.Report.pdf)


srambach (approved):
LGTM, thanks for the new example. 🛤️

thatblindgeye (approved):

PR #6972, opened 2024-08-13
fix(card): remove unused modifier and old logo from demo

### LINKED ISSUE #6249: Bug - [V6 Card] - Update demos
**Describe the problem**
A clear and concise description of the problem. Which components are affected?
The "horizontal split" demo still has the card `pf-m-flat` modifier.  This should be removed.
Tis demo also still has the old logo.

rebeccaalpert:
Fixes https://github.com/patternfly/patternfly/issues/6249

I fixed the [Card demo](https://patternfly-pr-6972.surge.sh/components/card/html-demos#horizontal-split) as outlined in the issue. I also noticed the [examples](https://patternfly-pr-6972.surge.sh/components/card) for Cards also had old logos - I took a look at fixing those as well here. I updated [Primary-Detail](https://patternfly-pr-6972.surge.sh/patterns/primary-detail#primary-detail-card-view-expanded) and the [Card view](https://patternfly-pr-6972.surge.sh/patterns/card-view) patterns as well since it allowed me to remove an old logo image.

Since I modified some demos, we'll have to update some screenshots I think, but I'm not sure if I do those as part of a PR or not, so I didn't. Please let me know if I should!

There are some other old images in the assets folder that look like they are in use - I would love to go through and remove/update those too, but I feel like this is already large.


mcoker (approved):
Awesome!! Here is a visual regression report showing what changed to any examples/demos with "card" in the name from this PR:

[BackstopJS Report.pdf](https://github.com/user-attachments/files/16607259/BackstopJS.Report.pdf)
[BackstopJS Report-dark.pdf](https://github.com/user-attachments/files/16607260/BackstopJS.Report-dark.pdf) - this one I ran in reverse order, so the images are out of order 😅

One little thing is the logo file used in https://patternfly-pr-6972.surge.sh/components/card/html/with-only-image-in-head/ (and other examples) isn't ideal in dark theme because the text doesn't change colors.

We have a variation of that image that uses PF's text-color tokens for the "Patternfly" text that we use in the masthead. We could use that in these card demos, too. https://github.com/patternfly/patternfly/blob/v6/src/patternfly/components/Masthead/masthead-image.hbs

The only issue with masthead-image.hbs is it relies on `masthead--id` as a handlebars parameter either on or as an ancestor of the include for this partial (`{{> masthead-image}}`).  So if you were to use that in these card examples, you'd have to specify `masthead--id` somewhere in those examples for that SVG's ID to work properly and not generate duplicate IDs on the page. There are a bunch of ways you could do it, but the simplest for the time being would probably be - if the `{{#> card ...}}` examples that would include masthead-image.hbs have a `card--id="..."` on the outer `{{#> card}}`, then you could replace the logos in those cards with this: `{{> masthead-image masthead--id=card--id}}`. There is also a hard-coded `height` attribute on the svg in masthead-image.hbs - if that doesn't work in the card examples, you could do something like change `height="37px"` to `height="{{ternary masthead-image--height masthead-image--height "37px"}}"`, then you could pass a custom height via `{{> masthead-image masthead--id=card--id masthead-image--height="1234px"}}`

mcoker:
@rebeccaalpert left a comment but LGTM! Your comments about running the screenshots job and auditing the other images in the assets folder would be great, though I agree this PR is already pretty big and it would be good to do that in a follow-up. If you wanted to work on the stuff in my comment above, that could be a follow-up, too. Just lemme know how you'd prefer to handle it!

rebeccaalpert:
Happy to do the masthead-image.hbs sub in this PR. If it's ok, I'll do the screenshots and looking through the other images separately so it's less of a pain for folks to review.

rebeccaalpert:
The swap for masthead-image should be all set!

rebeccaalpert:
Follow-on for additional images is here: https://github.com/patternfly/patternfly/pull/6976.

mcoker (approved):
Fabulous!!

Issue #6973, opened 2024-08-14
Backstop - make it easier to update list of URLs


mcoker:
Currently it's a manual process to generate and manage a list of URLs to test against. Ideally the list of URLs can be generated on the fly.

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6974

PR #6974, opened 2024-08-14
chore(backstop): build list of urls on the fly

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6973

Builds the list of URLs on the fly based on the site routes. This is probably a good baseline for the test to run. From here, we may want to be able to create a hook to add options to certain URLs, or maybe maintain a static list of additional URLs with add'l options (like a custom delay, selector to hover/click on, etc)


srambach (approved):
Looks great - I copied that specific url with advanced scenarios over to the regression testing doc for future reference.

PR #6976, opened 2024-08-14
fix(LoginPage): replace old PF logo with new one

rebeccaalpert:
This is a follow-on from https://github.com/patternfly/patternfly/pull/6972. I had noticed some old brand images in use, as well as a bunch of what appear to be unused images.

I'm sorry about the huge number of files here - happy to break this up, but it's mostly deletions. I went through and looked for each file name in the codebase. I'm not sure why we have two separate image folders, but we do, and there were some duplicates there as well.


srambach (commented):
I think there are a few chart images that need to stay - they are used in /demos/Card/templates/card-demo--chart.hbs

rebeccaalpert:
> I think there are a few chart images that need to stay - they are used in /demos/Card/templates/card-demo--chart.hbs

@srambach - It seemed like it was using the ones in https://github.com/patternfly/patternfly/tree/v6/src/patternfly/assets/images to me, and it looked like they were all duplicated in the folder https://github.com/patternfly/patternfly/tree/v6/static/assets/images. I removed the ones in static/assets/images and left the ones in src/patternfly/assets/images. I have no idea why there are two assets folders though! I'm happy to add them back if there's some reason we need both.

srambach:
> > I think there are a few chart images that need to stay - they are used in /demos/Card/templates/card-demo--chart.hbs
> 
> @srambach - It seemed like it was using the ones in https://github.com/patternfly/patternfly/tree/v6/src/patternfly/assets/images to me, and it looked like they were all duplicated in the folder https://github.com/patternfly/patternfly/tree/v6/static/assets/images. I removed the ones in static/assets/images and left the ones in src/patternfly/assets/images. I have no idea why there are two assets folders though! I'm happy to add them back if there's some reason we need both.

Got it - my mistake! I don't know why there are two folders; @mcoker might remember. I agree it would be best to consolidate them.

mcoker:
I'll have to verify, but off the top of my head, I believe /static/assets/images is part of the core repo, but not included in our NPM package (`@patternfly/patternfly`). The stuff in `/src/patternfly/assets/images` is included in our NPM package - https://unpkg.com/browse/@patternfly/patternfly@alpha/assets/images/

For these images we use in our workspace, I'm in favor of moving everything to `/static/assets/images`, as well as removing any we don't use. That will reduce our bundle size, which is good. Some considerations:
* Are any of those images in our NPM package used by pf-react, pf-org, or any other repos like extensions? I'm pretty sure there is a script in the pf-react repo's build that copies assets from core to an accessible place in the pf-react repo - that might be a good place to check.
* Does core reference any images in the CSS we ship? Meaning the default presentation of some component might be broken if we exclude some image - like background images in the background image component, login component, or about modal component. Or avatar default images or anything like that? I'm pretty sure that isn't the case, but not 100%.

rebeccaalpert:
Let me do some research @srambach and @mcoker and see if anyone else is consuming these then! I'll migrate everything we're keeping that isn't in use to /static/assets/images.

rebeccaalpert:
**Okay, so I went through v6 for these three:**
1. Core
    * Doesn't seem to have any usage of `url(` in .scss files except for what appear to be icons and fonts
    * Did another text search in Core for every non-duplicate image we're removing and didn't find any usage in Core, including in .scss files.
2. React
    * Did a text search for every file I'm removing - the only hits were files that appeared to be in assets folders in React, images hosted elsewhere (https://raw.githubusercontent.com/patternfly/patternfly-react/main/packages/react-integration/demo-app-ts/src/assets/images/pfbg_992.jpg), or an image in the tables package
3. Org
    * Did a text search for every file I’m removing - only found link to https://pf4.patternfly.org/assets/images/pf-logo-small.svg, which we may need to update, but shouldn’t be affected by file removal on the v6 branch

**Extensions (pulled from master branch, repos linked at https://miro.com/app/board/uXjVMgbEaCg=/)**
4. Virtual assistant 
  * No hits on text search
5. Data view 
  * No hits on text search
7. Log viewer
  * No hits on text search
9. React catalog view
  * No hits on text search
11. React console
  * No hits on text search
13. React topology
 * No hits on text search
15. React component groups
  * No hits on text search
17. React user feedback
  * No hits on text search
19. Quickstarts
  * No hits on text search
21. React virtualized extension
  * No hits on text search

rebeccaalpert:
As a side note, It's looking to me like both folders wind up in the package and/or dist. I'm seeing https://github.com/patternfly/patternfly/blob/v6/static/assets/images/pf-c-brand__logo-base.jpg in https://unpkg.com/browse/@patternfly/patternfly@6.0.0-alpha.219/assets/images/, even though it's not in https://github.com/patternfly/patternfly/tree/v6/src/patternfly/assets/images, as an example. However, it does look like hbs files don't get picked up in static/assets/images, while they seem to work ok in patternfly/assets/images.

mattnolting (approved):
Great job! This looks good to me, @mcoker fwiw we should consider renaming the workspace directory to something more descriptive, like `workspace/images`. 👍

mcoker (approved):
LGTM, backstop report for reference 
[alpert-assets.pdf](https://github.com/user-attachments/files/17139620/alpert-assets.pdf)

PR #6978, opened 2024-08-14
chore(Card): added aria-disabled to link cards

### LINKED ISSUE #6979: Card - add aria-disabled for actionable link cards
Actionable only cards that contain an anchor element for its action should have aria-disabled when the card is disabled. We prevent keyboards from accessing the anchor element, but other AT can access it so we need to convey that it's a disabled link.

**Any other information?**

thatblindgeye:
Closes #6979


mcoker (approved):
NICE!

Issue #6977, opened 2024-08-14
Slider: Update semantic tokens to support dark theme


lboehling:
Slider component doesn't have enough contrast between the track and ticks in dark theme. Need to update the following semantic tokens in core: 

track: adjust unfilled bkg color to use color--nonstatus--gray--default token
tick: adjust filled tick color to use icon--color--on-brand
tick: adjust disabled tick color to use icon--color--on-disabled
tick: adjust tick border radius to border--radius--sharp
tick: adjust tick width to .15rem

end result should look like this: 
<img width="820" alt="image" src="https://github.com/user-attachments/assets/3a613230-a7c1-4fe7-aa12-ae6608732ed1">

Design issue for reference -- 
https://github.com/patternfly/patternfly-design/issues/1316

Note -- will need to pull in the latest tokens from figma in order to see the nonstatus gray bkg token color changes

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6981

PR #6981, opened 2024-08-15
fix(slider): adjust styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6977 and pulls in the latest tokens from https://github.com/patternfly/design-tokens/pull/82

@lboehling  the only difference I see in the between what's in the PR and in the referenced issue is the disabled/unfilled tick color (updated to `--pf-t--global--icon--color--on-disabled`). In the design screenshot, the tick color is lighter than the background, but `--pf-t--global--icon--color--on-disabled` is a darker color. Not sure if anything needs to be updated there?

### DIFF
--- a/src/patternfly/components/Slider/slider.scss
+++ b/src/patternfly/components/Slider/slider.scss
@@ -8,7 +8,7 @@
   --#{$slider}__rail--PaddingBlockStart: var(--pf-t--global--spacer--md);
   --#{$slider}__rail--PaddingBlockEnd: var(--pf-t--global--spacer--md);
   --#{$slider}__rail-track--Height: #{pf-size-prem(4px)};
-  --#{$slider}__rail-track--before--base--BackgroundColor: var(--pf-t--global--border--color--default);
+  --#{$slider}__rail-track--before--base--BackgroundColor: var(--pf-t--global--color--nonstatus--gray--default);
   --#{$slider}__rail-track--before--fill--BackgroundColor: var(--pf-t--global--border--color--hover);
   --#{$slider}__rail-track--before--BorderRadius: var(--pf-t--global--border--radius--tiny);
   --#{$slider}__rail-track--before--fill--BackgroundColor--gradient-stop: var(--#{$slider}--value);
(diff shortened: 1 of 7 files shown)


src/patternfly/components/Slider/slider.scss
```diff
@@ -19,12 +19,12 @@
 
   // step tick
   --#{$slider}__step-tick--InsetBlockStart: var(--pf-t--global--spacer--md);
-  --#{$slider}__step-tick--Width: #{pf-size-prem(4px)};
+  --#{$slider}__step-tick--Width: 0.15rem;
   --#{$slider}__step-tick--Height: #{pf-size-prem(4px)};
-  --#{$slider}__step-tick--BackgroundColor: var(--pf-t--global--icon--color--subtle);
+  --#{$slider}__step-tick--BackgroundColor: var(--pf-t--global--icon--color--on-disabled);
```
lboehling, line 24: now that i'm thinking about this more, we should probably use `--pf-t--global--icon--color--nonstatus--on-gray--default` here. It will still look darker than the design screenshot in the issue (I didn't update the tick color token in the screenshot!).

lboehling (changes requested):
looks great, thank u!! left one suggestion re: the on-disabled tick color.

src/patternfly/components/Slider/slider.scss
mcoker: Updated 🚀 https://patternfly-pr-6981.surge.sh/components/slider

thatblindgeye (commented):
Not necessarily a blocker, but for the "Actions" example, is the slider itself meant to be disabled along with the input? Second Slider in the below screenshot. Otherwise lgtm

Three sliders showing various actions

mcoker:
@thatblindgeye nice catch! Looking at what's in v5 for react as how that should behave, looks like the slider and value input are disabled when the action is set to the lock. Updated that example to disable the slider.

thatblindgeye (approved):
🐝 -utiful

lboehling (approved):
perfect!! thanks for the quick work on this!!

srambach (approved):
🛝

Issue #6238, opened 2024-01-18
Notification badge - remove from core


mcoker:
Notification badge has been replaced with [stateful buttons](https://pf6.patternfly.org/components/button#stateful), so we can drop this component in core, and make mention in release notes that the stateful button should be used instead. Include a blurb about how to replace  notification badge with stateful buttons in the release notes.

React will maintain the notification badge as a unique component, but just use the stateful button under the hood.

PR #6983, opened 2024-08-16
fix(notification-badge): removed component, updated hbs partial

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6238

Moves the old notification-badge.hbs to the button component and updates the logic, so no examples needed to be updated. We can always update that at some point if it's confusing.

[Backstop report](https://drive.google.com/file/d/17RdaZBKk-on-Xw7nDD0KYEdxak8UQlLZ/view?usp=sharing) - the majority of this is a padding update to plain menu toggles, and other content around the toggles shifting slightly. It also includes the logo updates @rebeccaalpert made. If you scroll to the notification drawer examples, you'll see the notification badge update. If this report looks good, I'll add the screenshots to the PR.


thatblindgeye (approved):
LGTM. Not a blocker since like mentioned we can revisit, but since it looks like the notification badge is only being called in one masthead hbs file, updating the file/component name to something like a "button--stateful" or "button--notification-badge" template would work with minimal update.

mattnolting (approved):
LPTM! 🩴

Issue #6918, opened 2024-07-24
Masthead - expanded content layout bug


mcoker:
https://pf6.patternfly.org/components/masthead/html-demos/with-toggle-group-filters-expandable-content-expanded-mobile/

Expanded content doesn't span the full width

<img width="514" alt="Screenshot 2024-07-24 at 1 39 52 PM" src="https://github.com/user-attachments/assets/73767607-1628-48d0-930a-b555e2e22305">

mcoker:
From working session, it may be as simple as making this update

Change this:
https://github.com/patternfly/patternfly/blob/50ad63b98c51baae1fdcebe1557c77e62d575efd/src/patternfly/components/Toolbar/toolbar.scss#L148-L151

To this (to add `.pf-v6-c-toolbar.pf-m-static` to the declaration so it gets `position: static`, too)
```
 &.pf-m-static { 
  :is(&, .#{$toolbar}__content) { 
     position: static; 
   } 
```

Then updating this block:
https://github.com/patternfly/patternfly/blob/50ad63b98c51baae1fdcebe1557c77e62d575efd/src/patternfly/components/Toolbar/toolbar.scss#L261-L263

To add:
`inset-inline-start: 0;`

I can't remember if we needed to add `inset-inline-end: 0;`, too, and/or remove `width: 100%` from the expandable content element.

mcoker:
closed by https://github.com/patternfly/patternfly/pull/6986

PR #6986, opened 2024-08-16
fix(toolbar): fix expanded content position in masthead

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6918

[BackstopJS Report.pdf](https://github.com/user-attachments/files/16641892/BackstopJS.Report.pdf) with `filter="toolbar|masthead"`

The 3 failures are the last 3 in the report, and seem to match the layout I see for these examples v5

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -146,6 +146,7 @@ $pf-v6--include-toolbar-breakpoints: true !default;
 
   //  - Toolbar static
   &.pf-m-static {
+    &,
     .#{$toolbar}__content {
       position: static;
     }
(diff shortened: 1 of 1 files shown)


srambach (commented):
Just one question - the padding is getting reset to 0 above 62rem [here](https://github.com/patternfly/patternfly/blob/2b97f6f972f885c490d7b2c7a82e271b88c27985/src/patternfly/components/Toolbar/toolbar.scss#L277) - is this by mistake? The box-shadow is never actually removed so was it trying to look more inline but something is still off?

srambach (approved):
After discussion - all good!
The example in question is a mobile expandable toolbar, and at desktop width the expanded content does not have padding around the toolbar. Currently there really isn't a core example showing a solution for this at desktop width, but there's really not a good use case for this. If it comes up, we can further consider how to handle the padding around expanded content at desktop width.

mcoker:
@srambach thanks for the meeting notes!

Issue #6725, opened 2024-05-30
remove `*` global selectors where applicable


mattnolting:
The use of global selectors has a significant impact upon performance. These globals need to be updated to targeted selectors.

mcoker:
closing for now until this is planned

PR #6988, opened 2024-08-19
fix(grid): removed global selectors

mattnolting:
Applies to #6725


mcoker:
Discussed with the team and it's probably best to revisit performance as a whole at some point, possibly in a sprint/milestone dedicated to focusing on browser performance. Closing for now given the PR is stale and this would introduce a breaking change.

PR #6995, opened 2024-08-21
fix(is): updated is pseudos

### LINKED ISSUE #6720: update floating `:is(.selector, .selector)`, wherever used, to comma separated list
The use of `:is()` is having a significant impact upon performance. Floating references to `:is()` need to be update to comma separated list.

Visual regression testing should accompany this issue

mattnolting:
closes #6720 

This PR eliminates ~ 150k unnecessary match attempts

Screenshot 2024-08-21 at 9 25 46 AM

Screenshot 2024-08-21 at 10 48 48 AM

[Backstop Report](https://drive.google.com/file/d/1iopLKQ--ReMzp__TLGo4XBRHK3935vJR/view?usp=sharing)

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -127,7 +127,8 @@
   // SELECTABLE CARDS
   &.pf-m-selectable {
     // disables the gap since these have an empty label
-    .#{$card}__selectable-actions :is(.#{$check}, .#{$radio}) {
+    .#{$card}__selectable-actions .#{$check},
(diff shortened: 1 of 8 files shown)


thatblindgeye (approved):

mcoker:
Discussed as a team and I think we need to first evaluate the performance implications of using `:is()` and determine whether we need to modify code guidelines with stricter rules around the use of certain selectors. This would be good to revisit at some point when we have time to focus on browser performance. We can revisit this PR then, if needed.

Issue #6998, opened 2024-08-22
Expandable section - toggle should use link button, or link button styles


mcoker:
Reviewing the expandable section w/ @lboehling the original intent was for expandable section to use the link button. The button mostly looks like a link button now but doesn't always maintain the shape of the link button and doesn't have the link button background.

mcoker:
closed by https://github.com/patternfly/patternfly/pull/7000

PR #7000, opened 2024-08-22
fix(expandable-section): use link button, update styles

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6998

Visual regression report - [expandable-section.pdf](https://github.com/user-attachments/files/16702893/expandable-section.pdf)

Just a draft to see if it's something we want to get in this sprint. Cleans up some styles, too.

The react changes would be:
* Make `__toggle` a `<div>` instead of a custom button
* Use a link `<Button>` in `__toggle` and update old `__toggle` logic to apply to the `<Button>` instead


lboehling (approved):
very nice!! ✨

srambach (approved):
Love removing extra styling :)

PR #7004, opened 2024-08-26
fix(global-selectors): fix * global selectors

### LINKED ISSUE #6725: remove `*` global selectors where applicable

mattnolting:
closes #6725

### DIFF
--- a/src/patternfly/base/normalize.scss
+++ b/src/patternfly/base/normalize.scss
@@ -59,10 +59,12 @@
     iframe,
     object,
     audio,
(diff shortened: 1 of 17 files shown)


mcoker:
I think there are a lot of good changes in this PR, but it looks like a lot/most will be breaking at this point. This is a good one to evaluate either in a future performance sprint/milestone or in our next breaking change release.

PR #7005, opened 2024-08-26
feat(tokens): add temporary color token for debugging use

### LINKED ISSUE #7003: Tokens - create temporary placeholder token for use in codemods
Create a local token set to the pink highlight color that can be used in code mods to set unknown color tokens to an otherwise unused color so that fixes will stand out.

suggested name: `--pf-t--temp--dev--tbd`
suggested value: #BD11E0 (this does not and should not exist in the color palette) This is the color currently used in design annotations.

srambach:
Fixes #7003 

I disabled the linter for the token pattern - I could alter the pattern instead but thought this was better for this 1-off.

@wise-king-sullyman @kmcfaul

### DIFF
--- a/src/patternfly/base/tokens/tokens-local.scss
+++ b/src/patternfly/base/tokens/tokens-local.scss
@@ -122,4 +122,12 @@
 
   // List bullet/marker
   --pf-t--global--list-style: disc outside;
+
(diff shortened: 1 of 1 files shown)


mcoker (approved):
🚀

PR #7008, opened 2024-08-27
chore(demo-updates): wrapped flex children in flex-item

mattnolting:
closes


mcoker:
Looks like this may have been a follow up to this PR that removes `*` from the flex styles - https://github.com/patternfly/patternfly/pull/6989. Since https://github.com/patternfly/patternfly/pull/6989 was closed, closing this one for now and we can revisit if needed.

PR #7016, opened 2024-08-28
fix(global): remove where from global/component var blocks

mcoker:
This PR is the result of reviewing other issues/PR's and taking the first step of removing `:where()` from the selector that wraps all component/layout vars, which was identified as having a negative impact on performance. That just leaves `:root, .pf-[version + component]` as the selector that wraps component vars.

From there, we may also try some other steps to evaluate performance gains (if any) and risks to see if there are other updates to make like this for v6:
* Removing `:root` from the selector list
* Removing `:where()` from global reset styles

[BackstopJS Report.pdf](https://github.com/user-attachments/files/16803615/BackstopJS.Report.pdf)

### DIFF
--- a/src/patternfly/base/patternfly-variables.scss
+++ b/src/patternfly/base/patternfly-variables.scss
@@ -4,14 +4,14 @@
 @use "./tokens/tokens-dark" as dark;
 @use "./tokens/tokens-local" as local;
 
-:where(:root) {
+:root {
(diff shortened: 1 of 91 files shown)


srambach (approved):
🐳

mattnolting (approved):
🤩🤩🤩🤩🤩🤩🤩🤩🤩🤩 LAPTM!!!

PR #7027, opened 2024-08-29
chore(page): fix demo id, fixes masthead logo

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7028

Old & busted: https://pf6.patternfly.org/components/table/html-demos/sticky-header-and-multiple-columns/
New hotness: https://patternfly-pr-7027.surge.sh/components/table/html-demos/sticky-header-and-multiple-columns/


evwilkin (approved):
L🔥TM

PR #7035, opened 2024-09-03
fix(fa-wildcard): converted @import to @use, removed *

### LINKED ISSUE #6508: Remove wildcard from font awesome icon classes
Each of the FA classes is prefixed with `*`, which we should remove if we can. You can see them in our compiled CSS - https://unpkg.com/browse/@patternfly/patternfly@5.3.0-prerelease.5/patternfly-base.css

And it comes from this file, specifically this block https://github.com/patternfly/patternfly/blob/dba5b2890adeaaf3fab513c20cdba91168ac96f1/src/patternfly/base/_fa-icons.scss#L29-L31

IIRC, this is a workaround because we want to import a stylesheet for those classes, but the import can't happen inside of a conditional.

mattnolting:
closes #6508 

[Backstop Report](https://drive.google.com/file/d/1Y_4T4E4pXn3SQTqqyx4u37sgiSsXX_G8/view?usp=sharing)

## Performance metrics

**Before**

Screenshot 2024-09-09 at 6 47 16 AM

**After**

Screenshot 2024-09-09 at 6 51 04 AM


mcoker (changes requested):
Unless I'm missing something, this doesn't conditionally include FA? You should be able to disable FA entirely with `$pf-v6-global--disable-fontawesome: true;` and include FA via the CDN links with `$pf-v6-global--enable-fontawesome-cdn: true;`

mattnolting:
> Unless I'm missing something, this doesn't conditionally include FA? You should be able to disable FA entirely with `$pf-v6-global--disable-fontawesome: true;` and include FA via the CDN links with `$pf-v6-global--enable-fontawesome-cdn: true;`

@mcokerI think you need to expand the changed files view, the conditional hasn't changed 👍

mcoker:
@mattnolting this block is what used to conditionally include the FA stylesheets, but was removed in this PR -  https://github.com/patternfly/patternfly/blob/b11cb3ed512a0a0df22e4eac9250b785b2ceb29e/src/patternfly/base/patternfly-fa-icons.scss#L31-L33

Also when I set `$pf-v6-global--disable-fontawesome: true;`, the FA stylesheets are still included in the compiled CSS

mcoker (approved):
Two questions, but otherwise LGTM. Gonna merge and we can follow-up if needed on the questions.

PR #7044, opened 2024-09-04
Disabled dry run.

dlabaj:
Disable dry run so we can release to npm again.


mcoker (approved):
I don't know the implications/risks of this change, but approving in case you just need an approval to merge 👍

Issue #7034, opened 2024-09-03
Button - link buttons with just an icon are not the correct width


mcoker:
We have a min-width on `.pf-m-plain` buttons that maintains a proper shape when a narrow image is used, but any button without a visible, default boundary (like link buttons) are "plain" (but not `.pf-m-plain`) and use the plain button spacers, so they need the same update. The goal of this style per design is so we do not have buttons that are taller than they are wide.

It should be a safe update to move the min-width off of plain buttons and just apply to the button component as a whole. Pill and control buttons should not be impacted since they have enough inline padding to maintain a proper shape with little/narrow content, although if we (or users) change pill/control inline padding to something smaller, we would still want to maintain a min-width so the button isn't taller than it is wide.

Approved this basic idea with design, but we'll need to run visual regressions and make sure it doesn't introduce some other bug.

PR #7045, opened 2024-09-04
fix(button): add min-width

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7034

[backstop-button-min-width.pdf](https://github.com/user-attachments/files/16882480/backstop-button-min-width.pdf)

There is a sub-pixel size difference with the clipboard copy buttons for some reason when adding the `min-width`. Not sure why, but not concerned about it either 🐸


evwilkin (approved):
LGTM

PR #7051, opened 2024-09-09
feat(FormControl): refine textarea styles

### LINKED ISSUE #7013: Bug - Text area - content overflows focus ring
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Put enough text into the text area to cause it to grow vertically. While the text area is focused, you can see that that the focus ring overlaps the content in the vertical direction.

**Expected behavior**
Ideally, the text should be contained within the focus ring area, which is inside the boundary of the text input. 
Note that the auto-resizing version of the text area does not seem to have this issue.

**Screenshots**
image

kmcfaul:
Closes #7013

Overrides padding styles for formcontrol textareas, to account for the outline offset. I put the new values into variables, but they could be calc'd inline in the modifier as well, so LMK which place would be preferred.

### DIFF
--- a/src/patternfly/components/FormControl/examples/FormControl.md
+++ b/src/patternfly/components/FormControl/examples/FormControl.md
@@ -119,43 +119,43 @@ cssPrefix: pf-v6-c-form-control
 
 ### Textarea
 ```hbs
-{{#> form-control controlType="textarea" form-control--attribute='name="textarea-standard" id="textarea-standard" aria-label="Standard textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--attribute='name="textarea-standard" id="textarea-standard" aria-label="Standard textarea example"'}}
 Standard
 {{/form-control}}
 <br>
-{{#> form-control controlType="textarea" form-control--IsReadonly="true" form-control--attribute='name="textarea-readonly" id="textarea-readonly" aria-label="Readonly textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--IsReadonly="true" form-control--attribute='name="textarea-readonly" id="textarea-readonly" aria-label="Readonly textarea example"'}}
 Readonly
 {{/form-control}}
 <br>
-{{#> form-control controlType="textarea" form-control--IsPlain='true' form-control--IsReadonly="true" form-control--attribute='name="textarea-readonly-plain" id="textarea-readonly-plain" aria-label="Readonly plain textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--IsPlain='true' form-control--IsReadonly="true" form-control--attribute='name="textarea-readonly-plain" id="textarea-readonly-plain" aria-label="Readonly plain textarea example"'}}
 Readonly plain
 {{/form-control}}
 <br>
-{{#> form-control controlType="textarea" form-control--IsDisabled="true" form-control--attribute='name="textarea-disabled" id="textarea-disabled" aria-label="Disabled textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--IsDisabled="true" form-control--attribute='name="textarea-disabled" id="textarea-disabled" aria-label="Disabled textarea example"'}}
 Disabled
 {{/form-control}}
 <br>
-{{#> form-control controlType="textarea" form-control--IsSuccess="true" form-control--attribute='name="textarea-success" id="textarea-success" aria-label="Success state textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--IsSuccess="true" form-control--attribute='name="textarea-success" id="textarea-success" aria-label="Success state textarea example"'}}
 Success
 {{/form-control}}
 <br>
-{{#> form-control controlType="textarea" form-control--IsWarning="true" form-control--attribute='name="textarea-warning" id="textarea-warning" aria-label="Warning state textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--IsWarning="true" form-control--attribute='name="textarea-warning" id="textarea-warning" aria-label="Warning state textarea example"'}}
 Warning
 {{/form-control}}
 <br>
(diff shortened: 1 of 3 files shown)


mcoker:
@kmcfaul nice! I like the overall approach. I have some feedback on the vars - naming and making sure we have easy padding vars to adjust for the textarea, but before going into the specifics, I'd love to get @mattnolting and @thatblindgeye to approve the overall approach.

src/patternfly/components/FormControl/examples/FormControl.md
```diff
@@ -119,43 +119,43 @@ cssPrefix: pf-v6-c-form-control
 
 ### Textarea
 ```hbs
-{{#> form-control controlType="textarea" form-control--attribute='name="textarea-standard" id="textarea-standard" aria-label="Standard textarea example"'}}
+{{#> form-control controlType="textarea" form-control--IsTextarea="true" form-control--attribute='name="textarea-standard" id="textarea-standard" aria-label="Standard textarea example"'}}
```
thatblindgeye, line 122: This could always be a followup, but would be good to just leverage the `controlType` to determine text-area things rather than also needing to pass IsTextarea

thatblindgeye (approved):
I think the approach should be good. It keeps the consistent focus styling as the other form control elements, prevents the text from 'escaping' the focus outlined area, and the resize handle in the bottom right isn't obscured.

thatblindgeye (approved):

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -90,6 +90,10 @@
   // Textarea
   --#{$form-control}--textarea--Width: var(--#{$form-control}--Width);
   --#{$form-control}--textarea--Height: auto;
+  --#{$form-control}--textarea--PaddingBlockStart: calc(-1 * var(--#{$form-control}--OutlineOffset));
+  --#{$form-control}--textarea--PaddingBlockEnd: calc(-1 * var(--#{$form-control}--OutlineOffset));
+  --#{$form-control}--textarea--PaddingInlineEnd: calc(-1 * var(--#{$form-control}--OutlineOffset));
+  --#{$form-control}--textarea--PaddingInlineStart: calc(-1 * var(--#{$form-control}--OutlineOffset));
```
mcoker, line 96:
```suggestion
  --#{$form-control}--textarea--PaddingBlockStart--offset: calc(-1 * var(--#{$form-control}--OutlineOffset));
  --#{$form-control}--textarea--PaddingBlockEnd--offset: calc(-1 * var(--#{$form-control}--OutlineOffset));
  --#{$form-control}--textarea--PaddingInlineEnd--offset: calc(-1 * var(--#{$form-control}--OutlineOffset));
  --#{$form-control}--textarea--PaddingInlineStart--offset: calc(-1 * var(--#{$form-control}--OutlineOffset));
```

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -174,6 +175,28 @@
...
+    &.pf-m-warning,
+    &.pf-m-error {
+      --#{$form-control}--m-status--PaddingInlineEnd--offset: calc(var(--#{$form-control}__icon--FontSize) + var(--#{$form-control}--ColumnGap)); 
+    }
+
+    > textarea {
+      padding-block-start: calc(var(--pf-t--global--spacer--control--vertical--default) - var(--#{$form-control}--textarea--PaddingBlockStart));
+      padding-block-end: calc(var(--pf-t--global--spacer--control--vertical--default) - var(--#{$form-control}--textarea--PaddingBlockEnd));
+      padding-inline-start: calc(var(--#{$form-control}--inset--base) - var(--#{$form-control}--textarea--PaddingInlineStart));
+      padding-inline-end: calc(var(--#{$form-control}--inset--base) - var(--#{$form-control}--textarea--PaddingInlineEnd) + var(--#{$form-control}--m-status--PaddingInlineEnd--offset, 0));
```
mcoker, line 194:
```suggestion
      padding-block-start: calc(var(--#{$form-control}--PaddingBlockStart) - var(--#{$form-control}--textarea--PaddingBlockStart--offset));
      padding-block-end: calc(var(--#{$form-control}--PaddingBlockEnd) - var(--#{$form-control}--textarea--PaddingBlockEnd--offset));
      padding-inline-start: calc(var(--#{$form-control}--PaddingInlineStart) - var(--#{$form-control}--textarea--PaddingInlineStart--offset));
      padding-inline-end: calc(var(--#{$form-control}--PaddingInlineEnd) - var(--#{$form-control}--textarea--PaddingInlineEnd--offset) + var(--#{$form-control}--m-status--PaddingInlineEnd--offset, 0));  
```

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -174,6 +175,28 @@
     text-overflow: ellipsis;
   }
 
+  &.pf-m-textarea {
+    padding-block-start: var(--#{$form-control}--textarea--PaddingBlockStart);
+    padding-block-end: var(--#{$form-control}--textarea--PaddingBlockEnd);
+    padding-inline-start: var(--#{$form-control}--textarea--PaddingInlineStart);
+    padding-inline-end: var(--#{$form-control}--textarea--PaddingInlineEnd);
```
mcoker, line 182:
```suggestion
    padding-block-start: var(--#{$form-control}--textarea--PaddingBlockStart--offset);
    padding-block-end: var(--#{$form-control}--textarea--PaddingBlockEnd--offset);
    padding-inline-start: var(--#{$form-control}--textarea--PaddingInlineStart--offset);
    padding-inline-end: var(--#{$form-control}--textarea--PaddingInlineEnd--offset);
```

mcoker (commented):
This looks great. Left a couple of suggestions in the code, but here's what I was thinking. The `--#{$form-control}--textarea--Padding[...]` vars look like they could be vars to set a custom textarea padding (from their naming), but that isn't what those vars do. Lemme know what you think of this:

* Rename the `--#{$form-control}--textarea--Padding[...]` vars to `--#{$form-control}--textarea--Padding[...]--offset`. Just makes it clear these aren't the vars to set the textarea padding.
* Update this block https://github.com/patternfly/patternfly/blob/58edf60127133d982d8b1b64430d501a538998be/src/patternfly/components/FormControl/form-control.scss#L190-L194
  * Replace the first vars in those `calc()`s to reference the `--#{$form-control}--Padding[...]` vars. Then the textarea is still referencing the same `--pf-v6-c-form-control--Padding` vars all of the form control elements use.

PR #7053, opened 2024-09-10
fix(menu-toggle): fixed alignment, remove display: flex from __icon

### LINKED ISSUE #6975: MenuToggle - Bug in Toolbar alignment when icon is applied
**Please describe the issue**
To reproduce:

1. Go to https://staging-v6.patternfly.org/components/toolbar/#with-filters
2. Add `icon={<FilterIcon />}` to `MenuToggle` in Status dropdown

**Are there visuals for this issue? Please provide screenshots**
Screenshot 2024-08-14 at 11 44 38 AM

**Could you point to a branch or draft PR where this issue exists? Please provide a link to the code**
https://github.com/jenny-s51/odh-dashboard/pull/10/files

cc @mcoker

mattnolting:
closes #6975 

# TLDR:
`.#{$menu-toggle}__icon` should not be `display: flex`.

### DIFF
--- a/src/patternfly/components/MenuToggle/examples/MenuToggle.md
+++ b/src/patternfly/components/MenuToggle/examples/MenuToggle.md
@@ -566,7 +566,7 @@ To add a label to a split toggle that will be linked to the toggle button, pass 
 ### With avatar and text
 ```hbs
 {{#> menu-toggle}}
-  {{#> menu-toggle-icon}}
+  {{#> menu-toggle-icon menu-toggle-icon--HasAvatar=true}}
     {{> avatar avatar--modifier="pf-m-sm"}}
   {{/menu-toggle-icon}}
   {{#> menu-toggle-text}}
@@ -580,7 +580,7 @@ To add a label to a split toggle that will be linked to the toggle button, pass 
 &nbsp;
 
 {{#> menu-toggle menu-toggle--IsExpanded="true"}}
-  {{#> menu-toggle-icon}}
+  {{#> menu-toggle-icon menu-toggle-icon--HasAvatar=true}}
     {{> avatar avatar--modifier="pf-m-sm"}}
   {{/menu-toggle-icon}}
   {{#> menu-toggle-text}}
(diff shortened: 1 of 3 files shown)


mcoker (commented):
LGTM, but these examples with the avatar are off - [BackstopJS Report.pdf](https://github.com/user-attachments/files/16949598/BackstopJS.Report.pdf)


I don't think it's a huge deal, but it looks like it's the result of an `` being used there - it aligns differently, and image add a little space below. And there is a pre-existing issue that with our smallest avatar, the menu toggle is too tall, since it's bigger than any of our existing icon sizes - I've brought this particular issue up with @lboehling.

I wonder if a straight forward fix would be to use some unique styling for the container for an avatar? Maybe with a negative top/bottom margin so it can be larger than a regular icon and not change the overall menu toggle height, and it can be a flex layout and the image/icon inside can be centered vertically?

wdyt @andrew-ronaldson @lboehling?

mcoker:
From today's working session:
* Remove the `__icon` min-height https://github.com/patternfly/patternfly/blob/3c51b9fb8587ca589ea3a5b36b270688a1a80f3a/src/patternfly/components/MenuToggle/menu-toggle.scss#L508
* Add a `.pf-m-avatar` modifier we can use specifically for avatars (since that's a super common use case) with a negative top/bottom margin to allow it to expand into the button's padding. I used the `--xs` spacer in working session, but we may also consider using `calc(var(--whatever-the-button-padding-is) / 2)` to always split the difference for variations of the menu toggle that update the padding.
* Update menu toggles with an avatar in them to use that modifier

Then we'll want to create a react issue and see whether that can go in 👍

mcoker (changes requested):
The avatar looks good but looks like you added `__icon` back to the block with `display: flex` and that breaks the original issue of the menu toggle being misaligned with an icon. I verified by putting the CSS from this PR into the react workspace and duplicating the bug outlined in https://github.com/patternfly/patternfly/issues/6975. It's also worth noting that in the PR (or currently on v6), a FA icon and SVG icon cause menu toggles to be misaligned, but removing `display: flex` from `__icon` takes care of that, too. I created a simple codepen here to test a couple of things in case it's helpful. It links to the CSS in this PR - https://codepen.io/mcoker/pen/RwXboOv

Also since you removed `min-height`, you can remove this var now https://github.com/patternfly/patternfly/blob/19567b7347dbc946dc3d7e15de1b3171222b23cc/src/patternfly/components/MenuToggle/menu-toggle.scss#L52

srambach (approved):
Seems to look ok - regression tests are too far out of date to check meaningfully, but code looks good.

kmcfaul (approved):

mcoker (approved):
L🐳TM!

PR #7060, opened 2024-09-11
fix(global): removed root selector from component vars, fix var refs

mcoker:
As noted by @srambach, removing `:root` from the component var selectors drastically improves dev tools performance in Firefox. This PR makes that update, and fixes a handful of places we've cross-referenced component vars in other components since vars were previously available from the `:root` scope. 

Here is the backstop report - [backstop-remove-root.pdf](https://github.com/user-attachments/files/16971491/backstop-remove-root.pdf)

Everything is from `.pf-v6-c-radio.pf-m-standalone` now having a `min-height` because those were previously referencing `--#{$check}--Height` for their height, which didn't exist. They now reference `--#{$radio}--m-standalone--MinHeight`. It looks like the presence of a standalone check or radio is making a single line of regular content taller than it would be without it, which ideally that wouldn't happen. Here are a few examples to show that - https://codepen.io/mcoker/pen/bGPZyNB. That can be a follow-up IMO.

The last 3 screenshots in the backstop report are false positives from what I can tell. 

FWIW here is the quick and dirty script I put together to find instances of component vars referenced in stylesheets outside of the component the var is for. This returns more matches than we need to update, since most of these are used correctly.

```
for i in `find src/patternfly/components -name *.scss`; do
  echo "" && echo $i && COMPVAR=$(grep 'include pf-root' $i | sed -e 's/^.*\$//g' | sed -e 's/).*//g');
  grep -Eo '\-\-#\{\$[^}]+\}' $i | grep -vi "\$$COMPVAR";
done
```

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -260,7 +260,7 @@
   --#{$button}__icon--m-end--MarginInlineStart: 0;
 
   // Progress
-  --#{$button}__progress--width: calc(var(--#{$spinner}--m-md--diameter) + var(--pf-t--global--spacer--sm)); // matches medium spinner diameter plus a spacer
+  --#{$button}__progress--width: calc(var(--pf-t--global--icon--size--lg) + var(--pf-t--global--spacer--sm)); // matches medium spinner diameter plus a spacer
   --#{$button}__progress--Opacity: 0;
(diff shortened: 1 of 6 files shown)


mattnolting (approved):
LPTM 🪨

mattnolting (approved):
LPTM 🪨

PR #7062, opened 2024-09-11

mcoker:
Backstop report - [backstop-9.11.24.pdf](https://github.com/user-attachments/files/16964907/backstop-9.11.24.pdf)

Reasons for failures:
* There is a sub-pixel shift in input groups with a control button, result of adding `min-width` to the base button component.
* Form control icon spacing is from semantic gap spacer updates
  * The select status icons have moved over quite a bit and looks like too much space, wdyt @lboehling @andrew-ronaldson? Here's a screenshot below, and here's a link to those examples - https://patternfly-pr-7062.surge.sh/components/forms/form-control#select
      Screenshot 2024-09-11 at 9 18 03 AM
* Form sections - the gap between label/control pairs in sections was updated to match the gap on the base form when the pairs are not in a section
* Dual list selector - gap between the arrow plain actions in the middle was updated
* Description list
  * Column gap between dt/dd pairs was updated
  * Column gap between dt and dd in horizontal layout was updated
* Copy to clipboard - gap between actions was updated
* Card actions - gap between actions was updated
* Accordion - space between toggle icon and text was updated to use text-to-element spacer
* Form - inline control group (like radios/checkboxes in a horizontal row) gap was updated


evwilkin (approved):
LGTM based on your summary of changes in issue description - everything flagged looks to map back to those updates 👍

Issue #7057, opened 2024-09-10
Pagination - audit improvements


kmcfaul:
Add top padding to pagination. Also after discussing with design, the spacer we'll try is group--vertical, and we'll try it for both pagination and the toolbar's bottom padding.

PR #7065, opened 2024-09-11
fix(pagination): adjust block padding

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7057

Visual regression report - [backstop-toolbar-pagination-padding.pdf](https://github.com/user-attachments/files/16970719/backstop-toolbar-pagination-padding.pdf)

This updates both the toolbar's bottom padding and the bottom pagination's top padding to use `var(--pf-t--global--spacer--gap--group--vertical)`

### DIFF
--- a/src/patternfly/components/Pagination/pagination.scss
+++ b/src/patternfly/components/Pagination/pagination.scss
@@ -32,25 +32,27 @@ $pf-v6-c-pagination--variable-map: build-spacer-map("none", "sm", "md", "lg", "x
   --#{$pagination}--m-sticky--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$pagination}--m-sticky--BoxShadow: var(--pf-t--global--box-shadow--sm--bottom);
   --#{$pagination}--m-sticky--PaddingBlockStart: var(--pf-t--global--spacer--md);
-  --#{$pagination}--m-sticky--PaddingInlineEnd: var(--pf-t--global--spacer--lg);
   --#{$pagination}--m-sticky--PaddingBlockEnd: var(--pf-t--global--spacer--md);
   --#{$pagination}--m-sticky--PaddingInlineStart: var(--pf-t--global--spacer--lg);
-  --#{$pagination}--m-sticky--ZIndex: 100; // TODO: replace with z-index token
+  --#{$pagination}--m-sticky--PaddingInlineEnd: var(--pf-t--global--spacer--lg);
+  --#{$pagination}--m-sticky--ZIndex: var(--pf-t--global--z-index--xs);
   --#{$pagination}--m-sticky--InsetBlockStart: 0;
 
   // bottom
   --#{$pagination}--m-bottom--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$pagination}--m-bottom--BoxShadow: var(--pf-t--global--box-shadow--sm--top);
-  --#{$pagination}--m-bottom--InsetBlockEnd: 0;
-  --#{$pagination}--m-bottom--PaddingInlineEnd: var(--pf-t--global--spacer--lg);
-  --#{$pagination}--m-bottom--PaddingInlineStart: var(--pf-t--global--spacer--lg);
   --#{$pagination}--m-bottom--m-sticky--BoxShadow: var(--pf-t--global--box-shadow--sm--top);
-  --#{$pagination}--m-bottom--PaddingBlockStart: 0;
-  --#{$pagination}--m-bottom--m-sticky--PaddingBlockStart: 0;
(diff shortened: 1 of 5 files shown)


src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -13,7 +13,7 @@ $pf-v6--include-toolbar-breakpoints: true !default;
 @include pf-root($toolbar) {
   --#{$toolbar}--RowGap: var(--pf-t--global--spacer--md);
   --#{$toolbar}--PaddingBlockStart: 0;
-  --#{$toolbar}--PaddingBlockEnd: var(--pf-t--global--spacer--md);
+  --#{$toolbar}--PaddingBlockEnd: var(--pf-t--global--spacer--gap--group--vertical);
```
lboehling, line 16: looking through the backstop report and i'm wondering if we should leave this a md spacer. Toolbar starts looking really cramped in examples that have a divider beneath the toolbar
Screenshot 2024-09-12 at 4 02 46 PM

evwilkin:
Two quick questions:
1. It looks like this only touches pagination and no longer updates Toolbar (based on Lucia's PR feedback above), if so can we update the title/description to avoid confusion?
2. Does [line 157](https://github.com/patternfly/patternfly/pull/7065/files#diff-456d4c35448af34e645b16d479973485d68365f781ccdadcf497eb0388c8037cR157) do anything?  I see the m-bottom PaddingBlockStart variable assigned a new value, but [just below](https://github.com/patternfly/patternfly/pull/7065/files#diff-456d4c35448af34e645b16d479973485d68365f781ccdadcf497eb0388c8037cL154) it looks like the actual `padding-block-start` rule references a different variable (m-sticky-paddingblockstart) so I'm not sure if that first reassignment is having any effect?

mcoker:
@evwilkin great feedback! On the var question, bottom pagination is a bit tricky. Here's the rundown of how it works:
* By default, bottom pagination is sticky (no modifier or anything, it's just sticky) on small screens, then becomes normal (aka static, but not with a modifier or anything, it's just static) on large screens
* You can apply `.pf-m-sticky` which will make it sticky on all viewports, but really only has an impact on large screens since the small screen default is already sticky
* You can apply `.pf-m-static` in the same way - static/not-sticky on all viewports, but really only impacts small screens since on large screens it's static by default

src/patternfly/components/Pagination/pagination.scss
```diff
@@ -45,6 +45,10 @@ $pf-v6-c-pagination--variable-map: build-spacer-map("none", "sm", "md", "lg", "x
   --#{$pagination}--m-bottom--PaddingInlineEnd: var(--pf-t--global--spacer--lg);
   --#{$pagination}--m-bottom--PaddingInlineStart: var(--pf-t--global--spacer--lg);
   --#{$pagination}--m-bottom--m-sticky--BoxShadow: var(--pf-t--global--box-shadow--sm--top);
+  --#{$pagination}--m-bottom--PaddingBlockStart: 0;
+  --#{$pagination}--m-bottom--m-sticky--PaddingBlockStart: 0;
+  --#{$pagination}--m-bottom--md--PaddingBlockStart: var(--pf-t--global--spacer--gap--group--vertical);
+  --#{$pagination}--m-bottom--m-static--PaddingBlockStart: var(--pf-t--global--spacer--gap--group--vertical);
```
thatblindgeye, line 51: Super🦸🏼 nit, but these new vars could be bumped up a line so that all the PaddingBlockStart vars are grouped together

thatblindgeye (approved):
Nothing blocking below

mcoker:
@evwilkin oops, I replied prematurely above. Nice catch, I see what you mean and yep that needs to be updated 👍

evwilkin (approved):

mcoker:
@srambach @thatblindgeye from the feedback above, I took another look at the padding vars. We basically have a static layout and a sticky layout. Sticky is less than the `--md` breakpoint and with the `.pf-m-sticky` modifier, and static is greater than `--md` and with `.pf-m-static`.

I took the route of creating 4 sets of padding vars:
* `--Padding` - this would use the sticky values since this is a small screen, no modifier layout
* `--md--Padding` - this would use static values
* `--m-sticky--Padding` - sticky, duh
* `--m-static--Padding` - ⚡️

Taking a stab at a simpler way to expose vars to users, and less vars on our end, I just created `--m-sticky--Padding` and `--m-static--Padding` and am using those instead. So default padding uses `--m-sticky`, > than `--md` breakpoint uses `--m-static`, and obviously the modifier classes use their corresponding vars. WDYT? I'm inclined to stick to our existing way of defining vars and have 4 separate sets, but figured I'd try this and get y'alls thoughts.

srambach (approved):
Looks good - I like the variables. 🎋

Issue #6834, opened 2024-06-27
Ensure scrollbars are presenting as expected


thatblindgeye:
**Describe the enhancement or change**
Followup to investigation/discussion in https://github.com/patternfly/patternfly/issues/6045#issuecomment-2192500304

We should ensure that scrollbars are being included in any overflow: hidden or clip path on containers to achieve the effect that the scrollbars are being rounded as well. This will include the Page component where we currently visually have the scrollbar offset from the container.

**Is this request originating from a Red Hat product team? If so, which ones and is there any sort of deadline for this enhancement?**

**Any other information?**

mcoker:
@andrew-ronaldson @lboehling I'm good closing this issue since we updated the scrollbar styling. Do y'all want to keep it around? Maybe to maybe do an audit/review at some point?

andrew-ronaldson:
Good to close!

PR #7068, opened 2024-09-12
fix(page): adjust main container and sidebar scrollbars

mcoker:
towards https://github.com/patternfly/patternfly-design/issues/1315

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -15,6 +15,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   // Sidebar
   --#{$page}__sidebar--ZIndex: var(--pf-t--global--z-index--sm);
   --#{$page}__sidebar--Width: #{pf-size-prem(290px)};
+  --#{$page}__sidebar--xl--Width: auto;
   --#{$page}__sidebar--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
   --#{$page}__sidebar--BoxShadow: none;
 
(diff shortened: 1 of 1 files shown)


andrew-ronaldson (commented):
I checked this in Chrome, firefox and Safari. Safari is the only one that keeps the scrollbar background-color. Would this `::-webkit-scrollbar-track:` work to tackle the safari issue.

Other than that I think this looks great and Safari is a lower priority browser

mcoker:
@andrew-ronaldson yep, the controls for safari scrollbar colors are a little different, and seem to be finicky? Here's a codepen with the overrides in it. https://codepen.io/mcoker/pen/LYKvOgW

You can see in safari that the sidebar scrollbar styles are working, but not on the page main container. If we update the `.pf-v6-c-page__main-container` to be `display: block` instead of `display: flex`, the scrollbar styles work? IMO that shouldn't impact the scrollbar styles at all.

Also, I'm assuming if we were to introduce custom scrollbar styles, we would do it for all scrollbars - not just the sidebar/page container?

I went ahead and took out the scrollbar changes from this PR - wdyt about merging these changes and following up on the scrollbar stuff, either for final or post-v6, but that would let us get these other changes in now.

lboehling (approved):
beautiful!!

thatblindgeye (approved):
Would this go towards https://github.com/patternfly/patternfly/issues/6834 at all, too?

andrew-ronaldson (approved):
This looks good in the main container. The floating track next to the nav feels awkward but can be addressed your followup issue. Thanks!

srambach (approved):
📜 Looks good, nice work with design

PR #7070, opened 2024-09-12
fix(content): removed floating is

mattnolting:
related to #6720 

# TLDR:

This PR removes `:is(.pf-v6-content)` selectors.

## Performance impact

Match attempt reductions: 1651673 - 1551338 = **140335**
Performance increase of: **9%**

**Before:** 
Screenshot 2024-09-12 at 11 07 15 AM

Screenshot 2024-09-12 at 11 04 46 AM

Screenshot 2024-09-12 at 11 04 22 AM

**After:**

Screenshot 2024-09-12 at 10 57 14 AM

Screenshot 2024-09-12 at 10 56 56 AM

Screenshot 2024-09-12 at 10 56 46 AM

[BackstopJS Report is.pdf](https://github.com/user-attachments/files/16980343/BackstopJS.Report.is.pdf)

### DIFF
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -99,7 +99,7 @@
   --#{$content}--hr--BackgroundColor: var(--pf-t--global--border--color--default);
 }
 
-[class*="#{$content}"] {
+.#{$content} {
   font-size: var(--#{$content}--FontSize);
   line-height: var(--#{$content}--LineHeight);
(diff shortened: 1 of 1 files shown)


mcoker:
This will also be a good one to evaluate as a team to see what the performance implications are around using `:is()` and determine if we need to be stricter with our coding guidelines when using `:is()`. That way we can be intentional and consistent about it across PatternFly. Closing for now, but can reference as needed.

PR #7074, opened 2024-09-13
fix(content): actually remove root, update content selector

mcoker:
Looks like I forgot to commit removing `:root` in https://github.com/patternfly/patternfly/pull/7060 🤦‍♂️ And there was an update to the content component I was still working out, so made that, too. 

Here's the [backstop report](https://github.com/user-attachments/files/16987350/BackstopJS.Report.pdf). From what I can tell, everything is a false positive or related to previous updates.

### DIFF
--- a/src/patternfly/components/Content/content.scss
+++ b/src/patternfly/components/Content/content.scss
@@ -1,6 +1,6 @@
 @use '../../sass-utilities' as *;
 
-@include pf-root($content) {
+@include pf-root('[class*="#{$content}"]', false) {
   // Body
   --#{$content}--MarginBlockEnd: var(--pf-t--global--spacer--md);
   --#{$content}--LineHeight: var(--pf-t--global--font--line-height--body);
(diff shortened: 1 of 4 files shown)


src/patternfly/components/Content/content.scss
```diff
@@ -1,6 +1,6 @@
 @use '../../sass-utilities' as *;
 
-@include pf-root($content) {
+@include pf-root(':root', false) {
```
mcoker, line 3: Since we support `.pf-v6-c-content--[all content elements]` as a standalone class, outside of `.pf-v6-c-content` now, these vars need to be accessible outside of `.pf-v6-c-content`. Originally I had changed this to `[class*="pf-v6-c-content"]` but that won't work with user overrides since we'll be resetting the variables for each of those classes. So maybe content can be an exception and continue to scope its vars to `:root`?

mattnolting (approved):
This LGTM 💷

srambach (approved):
🫚 lgtm, let's roll

Issue #7080, opened 2024-09-13
Token cleanup, update visual regressions


mcoker:
There were some tokens renamed from the latest design-tokens export in https://github.com/patternfly/patternfly/pull/7075/

Namely these:

```diff
-  --pf-t--global--spacer--gap--group-to-group--horizontal: var(--pf-t--global--spacer--2xl);
-  --pf-t--global--spacer--gap--group-to-group--vertical: var(--pf-t--global--spacer--lg);
+  --pf-t--global--spacer--gap--group-to-group--horizontal--compact: var(--pf-t--global--spacer--sm);
+  --pf-t--global--spacer--gap--group-to-group--horizontal--default: var(--pf-t--global--spacer--2xl);
+  --pf-t--global--spacer--gap--group-to-group--vertical--compact: var(--pf-t--global--spacer--md);
+  --pf-t--global--spacer--gap--group-to-group--vertical--default: var(--pf-t--global--spacer--lg);
+  --pf-t--global--spacer--gap--text-to-element--compact: var(--pf-t--global--spacer--xs);
```

We should:
* Update group-to-group--horizontal/vertical to add `--default`
* Remove the `--compact` custom spacers we added to [tokens-local](https://github.com/patternfly/patternfly/blob/263fc87840e57e75f8b0af784758770d325e6ccd/src/patternfly/base/tokens/tokens-local.scss#L127-L130) since they come from figma now
* Update visual regression screenshots

PR #7081, opened 2024-09-13
fix(global): token cleanup, update visual regression screens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7080


lboehling (approved):
looks good! thanks @mcoker

PR #7086, opened 2024-09-17
chore(Alert): updated toggle and close styling per audit

### LINKED ISSUE #7054: Alert - audit improvements
> should the left expand button have an adjustment to the margin to shift it left?
> Close button size - why is it small?

From doc, left expand button margin needs an adjustment and close button size may need to be adjusted as well.

thatblindgeye:
Closes #7054 

Toggle before update:
Alert expand toggle currently shifted too far to the right

Toggle after update:
Alert expand toggle shifted to the left

Close button before update:
Alert close button currently with smaller clickable area

Close button after update:
Alert close button with a larger clickable area

Do we ever intend for consumers to ever be able to have multiple actions in an Alert, not just a close button? Wondering if having the action button be larger would cause issues if that's the case.

### DIFF
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -22,7 +22,7 @@
   --#{$alert}__toggle--MarginBlockStart: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
   --#{$alert}__toggle--MarginBlockEnd: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
   --#{$alert}__toggle--MarginInlineStart: calc(-1 * var(--pf-t--global--spacer--action--horizontal--plain--default));
-  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--action--horizontal--plain--default);
+  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--sm);
 
   // Toggle icon
   --#{$alert}__toggle-icon--Rotate: 0;
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -21,8 +21,8 @@
   // Toggle
   --#{$alert}__toggle--MarginBlockStart: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
   --#{$alert}__toggle--MarginBlockEnd: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
-  --#{$alert}__toggle--MarginInlineStart: calc(-1 * var(--pf-t--global--spacer--sm));
-  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--sm);
+  --#{$alert}__toggle--MarginInlineStart: calc(-1 * var(--pf-t--global--spacer--action--horizontal--plain--default));
+  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--action--horizontal--plain--default);
 
   // Toggle icon
   --#{$alert}__toggle-icon--Rotate: 0;
@@ -227,7 +227,6 @@
   margin-block-start: var(--#{$alert}__action--MarginBlockStart);
   margin-block-end: var(--#{$alert}__action--MarginBlockEnd);
   margin-inline-end: var(--#{$alert}__action--MarginInlineEnd);
-  transform: translateY(var(--#{$alert}__action--TranslateY));
 }
 
 .#{$alert}__action-group {
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -193,9 +193,7 @@
 }
 
 .#{$alert}__icon {
-  display: flex;
   grid-area: icon;
-  margin-block-start: var(--#{$alert}__icon--MarginBlockStart);
   margin-inline-end: var(--#{$alert}__icon--MarginInlineEnd);
   font-size: var(--#{$alert}__icon--FontSize);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Alert/alert.scss
```diff
@@ -21,6 +21,7 @@
   // Toggle
   --#{$alert}__toggle--MarginBlockStart: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
   --#{$alert}__toggle--MarginBlockEnd: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
+  --#{$alert}__toggle--MarginInlineStart: calc(-1 * var(--pf-t--global--spacer--sm));
```
mcoker, line 24: any reason not to use the semantic spacer here? That way the style continues to work if we update horizontal padding for buttons.

```suggestion
  --#{$alert}__toggle--MarginInlineStart: calc(-1 * var(--pf-t--global--spacer--action--horizontal--plain--default));
```

mcoker:
> Do we ever intend for consumers to ever be able to have multiple actions in an Alert, not just a close button? Wondering if having the action button be larger would cause issues if that's the case.

I think the comment was just around the button not being the right height from the line-height you fixed, so I'd expect any add'l actions to be the same size as the close button is now.

src/patternfly/components/Alert/alert.scss
```diff
@@ -226,10 +228,6 @@
   margin-block-end: var(--#{$alert}__action--MarginBlockEnd);
   margin-inline-end: var(--#{$alert}__action--MarginInlineEnd);
   transform: translateY(var(--#{$alert}__action--TranslateY));
-
-  > .#{$button} {
-    --#{$button}--LineHeight: 1;
```
mcoker, line 231: Nice, not sure why that was there to begin with.

src/patternfly/components/Alert/alert.scss
```diff
@@ -226,10 +228,6 @@
   margin-block-end: var(--#{$alert}__action--MarginBlockEnd);
   margin-inline-end: var(--#{$alert}__action--MarginInlineEnd);
   transform: translateY(var(--#{$alert}__action--TranslateY));
```
mcoker, line 230: wdyt about removing this? It's causing the close button to be off-center. My guess is it had something to do with the button being smaller? From this PR, here's what it looks like with/without the transform

Before
Screenshot 2024-09-17 at 3 31 43 PM

After
Screenshot 2024-09-17 at 3 31 55 PM

And while I'm here, I don't know why we have `display: flex` and `margin-block-start` on `.pf-v6-c-alert__icon`. Taking both of those off aligns the icon as expected with the text because both the icon and text are inline with the same font-size and line-height.

That's scope creep though, so I can create a follow-up unless you want to give it a shot in this PR.

mcoker (commented):
LGTM, but I'd like to use the semantic spacer for the toggle margin if possible. Left another comment, but I can add that as a follow-up if you don't want to add anything additional to this PR.

src/patternfly/components/Alert/alert.scss
```diff
@@ -21,7 +21,8 @@
   // Toggle
   --#{$alert}__toggle--MarginBlockStart: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
   --#{$alert}__toggle--MarginBlockEnd: calc(-1 * var(--pf-t--global--spacer--control--vertical--default));
-  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--sm);
+  --#{$alert}__toggle--MarginInlineStart: calc(-1 * var(--pf-t--global--spacer--action--horizontal--plain--default));
+  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--action--horizontal--plain--default);
```
mcoker, line 25: I think this one should just use a regular spacer - or do you think it should use the action spacer? If so can you add why?

```suggestion
  --#{$alert}__toggle--MarginInlineEnd: var(--pf-t--global--spacer--sm);
```

mcoker (commented):
Just one comment but otherwise LGTM! Here's the visual regression report from any example URL with "alert"

[BackstopJS Report.pdf](https://github.com/user-attachments/files/17050155/BackstopJS.Report.pdf)
[BackstopJS Report.pdf](https://github.com/user-attachments/files/17050155/BackstopJS.Report.pdf)
[BackstopJS Report.pdf](https://github.com/user-attachments/files/17050155/BackstopJS.Report.pdf)

src/patternfly/components/Alert/alert.scss
thatblindgeye: Would we want this to also be the horizontal spacer for the same reason as using it for the MarginInlineStart var?

Actually wondering if it would make sense to reverse the tokens, so that the sm token is used in the calc for MarginInlineStart, and the horizontal-plain one is used for MarginInlineEnd. If the horizontal token were customized, would we want that to affect the InlineStart spacing? It could cause the toggle to look like it's "breaking out" of the alert container (below screenshot is using the horizontal plain token on InlineStart only with an override of 24px, using sm token on InlineEnd):

image


Versus switching the tokens:

image

Though maybe this is also a case for not using the horizontal token for either var.

thatblindgeye: Discussed off GitHub. We're gonna go with reverting the MarginInlineEnd to the sm spacer, with the intent being that additional overrides may need to be taken (e.g. overriding the `--#{$alert}--PaddingInlineStart` var).

mcoker (approved):
LPTM! Thanks @thatblindgeye! 🥳

andrew-ronaldson (approved):
This if fan-freaking-tastic!

srambach (approved):
🚨 Looks good to me.

PR #7087, opened 2024-09-17
fix(CodeEditor): set explicit 100% or 90% heights on divs

### LINKED ISSUE #7195: Bug - Code editor - 100% height not working
Context is in https://github.com/patternfly/patternfly-react/issues/11013

rebeccaalpert:
We're using a third-party dependency for the PF React CodeEditor. This dependency states that by default, it will be 100% of the parent. Unfortunately, this is not what happens in PF React. The docs say the default is 100%. However, in PF React, you have to set a fixed height or rely on the height of the code. 

You can see this issue if you go to https://patternfly-react-v6.surge.sh/components/code-editor and remove `height="400px"` from the first demo.

It appears the third-party dependency by itself still defaults to 100%: https://codesandbox.io/p/sandbox/competent-jepsen-gz9dzf?workspaceId=f5599198-6a68-44b3-85bd-1caed5609909. I had a go at trying to replicate this in PatternFly React and I was able to get it to work locally if I removed the default `height: ''` prop in React and set these heights as `style={{}}` attributes. I don't think PF React will let me merge anything with `style={{}}` attributes, so I'm hoping I can make the change directly here.

The 90% height takes into account the header we place on top of the editor. If we use 100%, the CodeEditor always overflows the parent since the third-party Monaco Editor tries to take the full 100%. This way it is 100% of 90% of 100%. 🪦 

We're trying to place the CodeEditor in a modal in virtual-assistant, and it would be a huge help if we could rely on getting a default height. 

This is related to https://github.com/patternfly/patternfly-react/issues/11013.

Fixes https://github.com/patternfly/patternfly/issues/7195

### DIFF
--- a/src/patternfly/components/CodeEditor/code-editor-container.hbs
+++ b/src/patternfly/components/CodeEditor/code-editor-container.hbs
@@ -1,0 +1,3 @@
+<div role="presentation" tabIndex="0" class="{{pfv}}code-editor-container">
+  {{> @partial-block}}
+</div>
--- a/src/patternfly/components/CodeEditor/code-editor-container.hbs
+++ b/src/patternfly/components/CodeEditor/code-editor-container.hbs
@@ -1,3 +1,3 @@
-<div role="presentation" tabIndex="0" class="{{pfv}}code-editor-container">
+<div role="presentation" tabindex="0" class="{{pfv}}code-editor__container">
   {{> @partial-block}}
 </div>
--- a/src/patternfly/components/CodeEditor/code-editor.scss
+++ b/src/patternfly/components/CodeEditor/code-editor.scss
@@ -83,6 +83,12 @@
   }
 }
 
+.#{$code-editor}__container {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+}
+
 .#{$code-editor}__header {
   position: relative;
   display: flex;
--- a/src/patternfly/components/CodeEditor/code-editor.scss
+++ b/src/patternfly/components/CodeEditor/code-editor.scss
@@ -71,8 +71,15 @@
 }
 
 .#{$code-editor} {
+  display: flex;
+  flex-direction: column;
+
   &.pf-m-read-only {
     --#{$code-editor}__main--BackgroundColor: var(--#{$code-editor}--m-read-only__main--BackgroundColor);
+  }
+
+  &.pf-m-full-height {
+    height: 100%;
   }
 }
 
@@ -144,7 +151,8 @@
 }
 
 .#{$code-editor}__main {
-  position: relative;
+  position: relative; 
+  flex-grow: 1;
   color: var(--#{$code-editor}__main--Color, inherit);
   background-color: var(--#{$code-editor}__main--BackgroundColor);
   border: var(--#{$code-editor}__main--BorderWidth) solid;
@@ -170,6 +178,7 @@
 
 .#{$code-editor}__code {
   position: relative;
+  height: 100%;
   padding-block-start: var(--#{$code-editor}__code--PaddingBlockStart);
   padding-block-end: var(--#{$code-editor}__code--PaddingBlockEnd);
   padding-inline-start: var(--#{$code-editor}__code--PaddingInlineStart);
--- a/src/patternfly/components/CodeEditor/examples/CodeEditor.css
+++ b/src/patternfly/components/CodeEditor/examples/CodeEditor.css
@@ -1,0 +1,3 @@
+#ws-core-c-code-editor-with-full-height-modifier .ws-preview-html {
+  height: 400px;
+}
--- a/src/patternfly/components/CodeEditor/examples/CodeEditor.md
+++ b/src/patternfly/components/CodeEditor/examples/CodeEditor.md
@@ -113,6 +113,64 @@ cssPrefix: pf-v6-c-code-editor
 {{/code-editor}}
 ```
 
+
+### With optional code editor container 
+This is an extra container used in React to prevent event propagation if upload is enabled or there is a provided empty state.
+
+```hbs 
+{{#> code-editor}}
+  {{#> code-editor-container}}
+  {{#> code-editor-header}}
+    {{#> code-editor-header-content}}
+      {{#> code-editor-controls}}
+        {{> code-editor-controls--buttons}}
+      {{/code-editor-controls}}
+    {{/code-editor-header-content}}
+    {{#> code-editor-tab}}
+      {{#> code-editor-tab-icon}}
+        <i class="fas fa-code"></i>
+      {{/code-editor-tab-icon}}
+      {{#> code-editor-tab-text}}
+        HTML
+      {{/code-editor-tab-text}}
+    {{/code-editor-tab}}
+  {{/code-editor-header}}
+  {{#> code-editor-main}}
+    {{#> code-editor-upload}}
+      {{#> empty-state}}
+        {{#> empty-state-header}}
+          {{> empty-state-icon empty-state-icon--type="code"}}
+          {{#> empty-state-title}}
+            {{#> empty-state-title-text}}
+              Start editing
+            {{/empty-state-title-text}}
+          {{/empty-state-title}}
+        {{/empty-state-header}}
+
+        {{#> empty-state-body}}
+          Drag a file here or browse to upload.
+        {{/empty-state-body}}
+
+        {{#> empty-state-footer}}
+          {{#> empty-state-actions}}
+            {{#> button button--IsPrimary=true}}
+              Browse
+            {{/button}}
+          {{/empty-state-actions}}
+          {{#> empty-state-actions}}
+            {{#> button button--IsLink=true}}
+              Start from scratch
+            {{/button}}
+          {{/empty-state-actions}}
+        {{/empty-state-footer}}
+      {{/empty-state}}
+    {{/code-editor-upload}}
+  {{/code-editor-main}}
+  {{/code-editor-container}}
+{{/code-editor}}
+```
+
+
 ### With full height modifiers
 
 ```hbs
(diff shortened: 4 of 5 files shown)


rebeccaalpert:
I'm putting up the PF React stuff I have here in case that better illustrates what I'm trying to do in this weird PR: https://github.com/patternfly/patternfly-react/pull/11014.

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -145,12 +147,13 @@
 
 .#{$code-editor}__main {
   position: relative;
+  height: 90%;
```
srambach, line 150: My only question would be whether we can calculate a better height here. We could use calc(100% - 45px), or break the 45px into 37px plus the 4px inline padding - but there is the possibility that contents in the header wrap and increase the height. So a calc or 90% are both approximations, but the calc should be accurate more often? I wonder if @mcoker has any better ideas?

rebeccaalpert: Thank you both! I pulled this in as inline styles in the React repo and it seems to work, so I'll adjust this PR to match.

rebeccaalpert: https://patternfly-react-pr-11014.surge.sh/components/code-editor#with-actions (from the PR with inline styles in React) seems ok, but I'm not sure what I need to be looking for beyond looking at it next to https://react-staging.patternfly.org/components/code-editor#with-actions.

mcoker: Cool! Re: the actions example, that works but if you open that example and another one (like the [basic example](https://patternfly-react-pr-11014.surge.sh/components/code-editor/react/basic/)) and make your browser really tall, you'll see that the actions example doesn't _appear_ to fill the height of the viewport. The code editor component wrapper does fill the height, but the `__main` element inside does not because the wrapper flex layout isn't being applied. Maybe that's alright, or good enough for now?

A couple of ways off of the top of my head that we could fix it...
* I'd look into why the div with `role="presentation"` is there in the first place. Looks like it's [this div element in the code editor component](https://github.com/patternfly/patternfly-react/blob/ab1735b515616effcc6bdff25ce301e15cf5e764/packages/react-code-editor/src/components/CodeEditor/CodeEditor.tsx#L663-L668). And looks like it adds `.pf-m-loading`, which does nothing in that context... but we have `.pf-m-loading` in 2 places, menu and file upload, so I'm guessing this was a `<FileUpload>` component, or a hard coded version of it, at some point?
  * Right now that div wraps and accepts a drag over event over the code editor's header and main elements, which is all that goes into a code editor. That says to me, you may be able to wrap the entire `<CodeEditor>` in `<FileUpload>` and get the same thing? And that won't break the code editor's flex layout to manage the `__header` and `__main` elements.
  * Maybe whatever functionality we're adding to that `<div>` in react, we could just add to the `.pf-v6-c-code-editor` div instead? That will maintain the flex layout, too.
* If we want to keep that `<div>` in the previous bullet, we can add a class to it (like `.pf-v6-c-code-editor__[whatever its job is]`) - then in this PR, you could add `.pf-v6-c-code-editor__[whatever] { display: flex; flex-direction: column; }`

rebeccaalpert: Got it, thank you. That makes sense - I can see the difference now on fullscreen.

We're using `react-dropzone` for upload functionality in both `<FileUpload>` and `<CodeEditor>`. It's effectively a wrapper in both components. I'm guessing that's why that loading class is there, but I'm not 100% sure. We also use `react-dropzone` for upload button functionality in Code Editor, as well as drag and drop.

Let me see what happens to those examples if I add that extra class first - and if that doesn't work, I'll try to dig into that `.pf-m-loading` div second.

rebeccaalpert: If I wrap everything that doesn't have upload enabled in an extra div with `display:flex; flex-direction: column`, it definitely fixes the other examples, but it breaks the one I need (the full-height example). Maybe we could add a class on the Core side and conditionally add it on the React side if a height is passed in? 

Screenshot 2024-09-18 at 11 35 50 AM

rebeccaalpert: Updated to use modifier classes per our last discussion! Thanks for thinking through this with me.

src/patternfly/components/CodeEditor/code-editor-container.hbs
```diff
@@ -0,0 +1,3 @@
+<div role="presentation" tabIndex="0" class="{{pfv}}code-editor-container">
```
mcoker, line 1: the component classes follow BEM so we can make "container" an element in code editor, which will be `.pf-v6-c-code-editor__container`. Also just lowercases the `tabindex` since it's lowercase in plain HTML.

```suggestion
<div role="presentation" tabindex="0" class="{{pfv}}code-editor__container">
```

src/patternfly/components/CodeEditor/code-editor.scss
```diff
@@ -71,9 +71,16 @@
...
+  flex-direction: column;
+
   &.pf-m-read-only {
     --#{$code-editor}__main--BackgroundColor: var(--#{$code-editor}--m-read-only__main--BackgroundColor);
   }
+
+  &.pf-m-full-height {
+    height: 100%;
+  }
 }
```
mcoker, line 85: Do we need styles for `#{$code-editor}__container`? IIRC from our convos, I'd expect it to be:

```suggestion
.#{$code-editor}__container {
  display: flex;
  flex-direction: column
  flex-grow: 1;
}
```

`display: flex; flex-direction: column;` - makes it a flex layout like the main code editor wrapper so that `__header` and `__main` are flex children, which makes this rule work https://github.com/patternfly/patternfly/blob/5006992758d980967cdcde91213692eda8cedeb9/src/patternfly/components/CodeEditor/code-editor.scss#L155

`flex-grow: 1;` - tells `__container` to fill the height of the code editor wrapper. Without this, the code editor won't fill the available space if you use `.pf-m-full-height` on the code editor wrapper.

src/patternfly/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -113,6 +113,99 @@ cssPrefix: pf-v6-c-code-editor
...
+  {{/code-editor-main}}
+  {{/code-editor-container}}
+{{/code-editor}}
+```
+
+
+### With full height modifiers
+
+```hbs
+{{#> code-editor code-editor--modifier="pf-m-full-height"}}
```
mcoker, line 177: For this example, can you add a local style to modify the height of the example so that the code editor grows to fill the available space? 

You'll want to create a new file here (in the examples dir) called `CodeEditor.css` and then add `import './CodeEditor.css'` to the top of this MD file - below the metadata and above the first heading, ~L6. Here's an example of this in our [table demo markdown](https://github.com/patternfly/patternfly/blob/5006992758d980967cdcde91213692eda8cedeb9/src/patternfly/demos/Table/examples/Table.md?plain=1#L7) file.

Then in `CodeEditor.css` add something like this:

```
#ws-core-c-code-editor-with-full-height-modifiers .ws-preview-html {
  height: 400px;
}

**Review by mcoker — COMMENTED** [review]
Looks rad! Left a few comments, lemme know if any of that doesn't make sense 😁

**Comment by rebeccaalpert** [comment]
Hopefully all set! Thank you @mcoker.

File: src/patternfly/components/CodeEditor/code-editor-container.hbs [hunk 5]
```diff
@@ -0,0 +1,3 @@
+<div role="presentation" tabindex="0" class="{{pfv}}code-editor-container">
```
mcoker, line 1: This will update the markup to match the CSS class

```suggestion
<div role="presentation" tabindex="0" class="{{pfv}}code-editor__container">
```

src/patternfly/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -113,6 +115,99 @@ cssPrefix: pf-v6-c-code-editor
...
+        {{/empty-state-footer}}
+      {{/empty-state}}
+    {{/code-editor-upload}}
+  {{/code-editor-main}}
+  {{/code-editor-container}}
+{{/code-editor}}
+```
+
+
+### With full height modifiers
```
mcoker, line 176: Just a nit but looks like there is only 1 modifier? 

```suggestion
### With full height modifier
```

src/patternfly/components/CodeEditor/examples/CodeEditor.css
```diff
@@ -0,0 +1,3 @@
+#ws-core-c-code-editor-with-full-height-modifiers .ws-preview-html {
```
mcoker, line 1: If you update the example name, we'll need to update this id

```suggestion
#ws-core-c-code-editor-with-full-height-modifier .ws-preview-html {
```

mcoker (changes requested):
woot! Just a few small comments. Also a small nit - I noticed in dev tools `height: 100%` on `.pf-v6-c-code-editor__code` isn't working because `<code>` is an inline element by default. But according to our docs and in react, that element is a `<div>`, and that's a block element and will fix the `height: 100%` style not being applied. Would you mind updating this `<code>` element to a `<div>`?

https://github.com/patternfly/patternfly/blob/83265b8488057e62d6fa2be92f4423698c5f3772/src/patternfly/components/CodeEditor/code-editor-code.hbs#L1-L6

rebeccaalpert:
Should be all set!

src/patternfly/components/CodeEditor/examples/CodeEditor.md
```diff
@@ -113,6 +115,99 @@ cssPrefix: pf-v6-c-code-editor
...
+        code goes here
+      {{/code-editor-code-pre}}
+    {{/code-editor-code}}
+  {{/code-editor-main}}
+{{/code-editor}}
+```
+
 ## Documentation
 
 ### Usage
```
srambach, line 213: Could you add the `pf-m-full-height` modifier to the documentation section please?

mcoker: @rebeccaalpert can you add the `__container` element to the docs, too?

mcoker (approved):
LGTM other than the comments to add `.pf-m-full-height` and the `__container` element to the docs table!

src/patternfly/components/CodeEditor/examples/CodeEditor.md
rebeccaalpert: You got it! All set @srambach and @mcoker!

mcoker (approved):
LPTM! Great job on this @rebeccaalpert ⭐️

PR #7090, opened 2024-09-18
fix(tokens): update box-shadows to only be on the variant side

mcoker:


andrew-ronaldson (approved):
I think you've done a great job here! 👑

lboehling (approved):
got surge to work for me! checked across all of our examples and these updates are looking goooood! thanks @mcoker!!

Issue #7036, opened 2024-09-03
Buttons: Fix transition on secondary and tertiary buttons


andrew-ronaldson:
The transition between the Default and hover states seems clunky on secondary and tertiary buttons.

PR #7096, opened 2024-09-21
fix(buttons): border transition update

andrew-ronaldson:
fixes https://github.com/patternfly/patternfly/issues/7036

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -50,6 +50,7 @@
   // Secondary
   --#{$button}--m-secondary--Color: var(--pf-t--global--text--color--brand--default);
   --#{$button}--m-secondary--BorderColor: var(--pf-t--global--border--color--brand--default);
+  --#{$button}--m-secondary--TransitionDuration: var(var(--pf-t--global--motion--duration--fade--short));
   --#{$button}--m-secondary__icon--Color: var(--pf-t--global--icon--color--brand--default);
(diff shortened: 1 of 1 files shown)


srambach (approved):
Spectacular!  ✨

src/patternfly/components/Button/button.scss
```diff
@@ -50,6 +50,7 @@
   // Secondary
   --#{$button}--m-secondary--Color: var(--pf-t--global--text--color--brand--default);
   --#{$button}--m-secondary--BorderColor: var(--pf-t--global--border--color--brand--default);
+  --#{$button}--m-secondary--TransitionDuration: var(--pf-t--global--motion--duration--fade--short);
```
mattnolting, line 53: Nice 👍. Instead of reassigning css vars, I suggest using a base transition duration for buttons, like:

```
...
--#{$button}--TransitionDuration: var(--pf-t--global--motion--duration--fade--short);
}

.button {
  transition-duration: var(--#{$button}--TransitionDuration);
```

We're working to pivot away from reassigning css variables as this approach affects performance implications and using `prop: val;` definitions is more readable in devtools.

mcoker (approved):
LGTM 👍

src/patternfly/components/Button/button.scss
mcoker: > We're working to pivot away from reassigning css variables as this approach affects performance implications and using prop: val; definitions are more readable in devtools.

We can talk about this as a team and see when/if it makes sense to do that. It's always something we can do as a non-breaking follow-up once we have a strategy for it 👍

PR #7101, opened 2024-09-24
fix(Scrollbar): style scrollbar color

### LINKED ISSUE #7085: Spike - scrollbar styles
re: https://github.com/patternfly/patternfly-design/issues/1315 and https://github.com/patternfly/patternfly/pull/7068

Design would like to investigate customizing the scrollbar styles in PF for v6. We can start with updating the track to "transparent" and the thumb to `--pf-t--global--border--color--default`. Some thoughts from testing:

* `scrollbar-color` [works everywhere except for Safari](https://caniuse.com/?search=scrollbar-color), and seemed to work fine in FF and Chrome on my Mac. We would also want to test other OS's and make sure things look as expected.
* Safari supports a [whole mess of pseudos for styling scrollbars](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar), notably `::-webkit-scrollbar-track` and `::-webkit-scrollbar-thumb` for our purposes.
  * These styles are somewhat finicky. I used this codepen for testing in another PR - https://codepen.io/mcoker/pen/LYKvOgW. Note that in Safari, the main content scrollbar is unstyled, and the first block in the CSS pane can be uncommented, showing that `display: block` instead of `display: flex` allows the scrollbar styles to work. 
* If `::-webkit-scrollbar-*` doesn't work as expected, would it be fair for that to be a FF/Chrome feature only/progressive enhancement (cc @andrew-ronaldson)? And we could either continue working on Safari to figure out the right combination of styles to get the change in Safari, or wait for Safari to implement `scrollbar-color`. 
  * Who knows how far out this is in Safari. Recently [scrollbar-gutter](https://bugs.webkit.org/show_bug.cgi?id=258743) and [scrollbar-width](https://bugs.webkit.org/show_bug.cgi?id=277518) was added to Safari Tech Preview, which is promising, but looks like there may be a blocker for [scrollbar-color](https://bugs.webkit.org/show_bug.cgi?id=231590).
* Is the expectation that this is a global style in PF, so any scrollbar will have custom colors? Or are we specifically targeting prominent ones like the sidebar and page main content? I'm assuming it's a global style for all scrollbars, but want to confirm the intent with design.

cc @andrew-ronaldson @lboehling @srambach - any additional thoughts/notes to add?

srambach:
Fixes #7085 

To test - Look at any scrollbar. Try light and dark mode, turn scrollbar preference to always/while scrolling, and try different browsers.

### DIFF
--- a/src/patternfly/base/normalize.scss
+++ b/src/patternfly/base/normalize.scss
@@ -6,6 +6,22 @@
   *::before,
   *::after {
     box-sizing: border-box;
+    scrollbar-color: var(--pf-t--global--border--color--default) transparent;
+  }
+    
+  // styling scrollbar color in Safari requires setting max-width and max-height
+  *::-webkit-scrollbar-thumb {
+    background-color: var(--pf-t--global--border--color--default);
+    border-radius: var(--pf-t--global--spacer--xs);
+  }
+
+  *::-webkit-scrollbar-track {
+    background-color: transparent;
+  }
+
+  *::-webkit-scrollbar {
+    max-width: var(--pf-t--global--spacer--sm);
+    max-height: var(--pf-t--global--spacer--sm);
   }
 
   :where(


dlabaj (approved):
LGTM.  Tested on Fedora and Ubuntu using both Firefox and Chrome.

mcoker (approved):
LPTM! Tested chrome/FF/safari and looks fantastic everywhere 🥳

Issue #7127, opened 2024-09-27
Page main container with pf-m-fill doesn't work in a drawer


mcoker:
Reported by @jeff-phillips-18 

If the `<Page>` has a `notificationDrawer` component set, then setting `isContentFilled` is not having an effect.

PR #7130, opened 2024-09-30
fix(page): fixed fill/scroll with notification drawer

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7127

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -417,11 +417,6 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   }
 }
 
-// Page level drawer - move to page component in breaking change
-.#{$drawer}__body > .#{$page}__main {
-  height: 100%;
-}
-
 .#{$drawer}__splitter {
   position: relative;
   display: none;
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -475,6 +475,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   flex-shrink: 0;
 }
 
-.#{$page}__drawer > .#{$drawer} {
-    flex: 1 0 auto;
+.#{$page}__drawer .#{$page}__main-container {
+  align-self: revert;
+
+  &.pf-m-fill {
+    flex-grow: 1;
+  }
 }
--- a/src/patternfly/demos/NotificationDrawer/notification-drawer-template.hbs
+++ b/src/patternfly/demos/NotificationDrawer/notification-drawer-template.hbs
@@ -11,7 +11,7 @@
   {{#> page-drawer}}
     {{#> drawer}}
       {{#> drawer-main}}
-        {{#> drawer-content}}
+        {{#> drawer-content drawer-content--NoBody=true}}
           {{#> page-main}}
             {{> page-template-breadcrumb}}
             {{> page-template-title}}


src/patternfly/components/Drawer/drawer.scss
```diff
@@ -417,11 +417,6 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   }
 }
 
-// Page level drawer - move to page component in breaking change
-.#{$drawer}__body > .#{$page}__main {
-  height: 100%;
-}
-
```
mcoker, line 424: Old v5 style, doesn't match the v6 structure.

src/patternfly/components/Page/page.scss
```diff
@@ -475,6 +475,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   flex-shrink: 0;
 }
 
-.#{$page}__drawer > .#{$drawer} {
-    flex: 1 0 auto;
+.#{$page}__drawer .#{$page}__main-container {
```
mcoker, line 478: There should only be one `.#{$page}__main-container` in `.#{$page}__drawer` unless anyone can think of a situation where that isn't the case? cc @srambach @mattnolting @dlabrecq

src/patternfly/components/Page/page.scss
```diff
@@ -475,6 +475,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
+  align-self: revert;
```
mcoker, line 479: default for `.#{$page}__main-container` is `align-self: start` which pushes that element to the top of the page component's grid layout. Since `.#{$page}__main-container` goes in `.#{$drawer}__content` (a flex column), that pushes `__main-container` to the left. This can just be the default/stretch.

src/patternfly/components/Page/page.scss
```diff
@@ -475,6 +475,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
+
+  &.pf-m-fill {
+    flex-grow: 1;
```
mcoker, line 482: Default for `__main-container.pf-m-fill` is `align-self: stretch` which works great in the page grid layout. Since this is in a flex column, it can just grow.

src/patternfly/demos/NotificationDrawer/notification-drawer-template.hbs
```diff
@@ -11,7 +11,7 @@
   {{#> page-drawer}}
     {{#> drawer}}
       {{#> drawer-main}}
-        {{#> drawer-content}}
+        {{#> drawer-content drawer-content--NoBody=true}}
```
mcoker, line 14: Just removes `drawer__body`, so `page__main-container` is a child of `drawer__content`, a flex column. Allows us to manage the height/alignment of `page__main-container` better

src/patternfly/components/Drawer/drawer.scss
tlabaj: Is the drawer body container being removed completely?

mcoker: Not from the component, just from the page component structure when the page-main-container/page-main elements are a child of the drawer content. AFAIK the only way to get that structure in react is via `<Page notificationDrawer={...}>`. So we would just remove the wrapper element `<DrawerContentBody>` here - https://github.com/patternfly/patternfly-react/blob/5decd6a8bf9632288abdbc98d3cac1902303c04e/packages/react-core/src/components/Page/Page.tsx#L349

tlabaj: Oh, it looks like it is just in relation to the page and does not impact the drawer component.

mattnolting (approved):
Great work, LPTM 🐰

srambach (commented):
🏎️ Go! Looks awesome, good detecting

srambach (approved):
🏎️ Go! Looks awesome, good detecting

PR #7134, opened 2024-10-01
fix(table): fixed thead checkbox fitting

### LINKED ISSUE #7132: Checkbox in empty state get cut off
Screenshot 2024-09-30 at 4 29 24 PM

Table `tbody` check cells create space for the header check cell. When the body cells aren't present (empty state), the `thead` check truncates.

mattnolting:
closes #7132 

# TLDR:
`table__check` cells are supposed to be set as `max-width: none`. In this case, `max-width` is being overridden by [`.pf-v6-c-table tr:where(.pf-v6-c-table__tr) > :where(th, td)`](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Table/table.scss#L322)

Convenience links:
https://patternfly-pr-7134.surge.sh/components/table/html-demos/loading-state-demo/
https://patternfly-pr-7134.surge.sh/components/table/html-demos/empty-state/

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -716,7 +716,7 @@
   --#{$table}--cell--MinWidth: 0;
   --#{$table}--cell--Width: 1%;
 
-  max-width: fit-content;
+  max-width: none;
 }
 // stylelint-enable
 
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -716,7 +716,7 @@
   --#{$table}--cell--MinWidth: 0;
   --#{$table}--cell--Width: 1%;
 
-  max-width: none;
+  max-width: fit-content;
 }
 // stylelint-enable
 
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -715,6 +715,8 @@
 .#{$table} .#{$table}__draggable {
   --#{$table}--cell--MinWidth: 0;
   --#{$table}--cell--Width: 1%;
+
+  max-width: none;
 }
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -715,6 +715,8 @@
 .#{$table} .#{$table}__draggable {
   --#{$table}--cell--MinWidth: 0;
   --#{$table}--cell--Width: 1%;
+
+  max-width: none;
```
srambach, line 719: WDYT about using something like fit-content or min-content? I'm not sure it's any different functionally but worries me less 🙃

mattnolting: Works for me! @srambach updated

mcoker: Looks like "fit-content" doesn't work in FF, though "none" does

mcoker (commented):
The checkbox isn't visible at all in FF, though FWIW that was pre-existing. If there are no concerns with using `max-width: none`, it looks like that works in FF, too.

mattnolting:
@mcoker updated `.#{$table}__check` to `max-width: none`

mcoker (approved):
🍗

mcoker:
[BackstopJS Report](https://github.com/user-attachments/files/17740253/BackstopJS.Report.pdf) for `yarn backstop:test --filter="table"`. The widened labels are from another PR (filed an issue to fix here - https://github.com/patternfly/patternfly/issues/7215)

src/patternfly/components/Table/table.scss
srambach: 👍🏻

srambach (approved):

PR #7136, opened 2024-10-01
chore(menu-toggle): remove reference to pf-m-action

### LINKED ISSUE #7135: Menu-toggle - remove docs for modifier pf-m-action
**Please describe the issue**
https://github.com/patternfly/patternfly/pull/7100/files#top removed the modifier `.pf-m-action` but it still exists in the documentation, so it should be removed.

srambach:
We removed `pf-m-action` from the menu toggle but missed removing it from the hbs and docs, so this cleans that up.

Fixes #7135

### DIFF
--- a/src/patternfly/components/MenuToggle/examples/MenuToggle.md
+++ b/src/patternfly/components/MenuToggle/examples/MenuToggle.md
@@ -536,7 +536,7 @@ Shown with default, primary, and secondary styling
 <br/>
 <br/>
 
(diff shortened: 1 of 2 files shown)


mcoker (approved):
LGTM! As part of this, we should probably also:
* Remove the use of `menu-toggle--IsAction` since that no longer does anything
* Rename and clean up the split toggle examples since a lot of them reference "action" if they were a `.pf-m-action`-specific toggle

Any interest in doing that here? If not we can create a follow-up.

mcoker (approved):
L🎄TM!

Issue #7137, opened 2024-10-02
Menu toggle - split button alignment issues with SVGs


mcoker:
Spotted while reviewing https://github.com/patternfly/patternfly-react/pull/11096, the menu toggles in these examples are too tall:
* https://react-staging.patternfly.org/components/menus/menu-toggle#split-toggle-with-checkbox-and-toggle-text
* https://react-staging.patternfly.org/components/menus/menu-toggle#split-toggle-with-action

And you can see the divider doesn't extend the height of the toggle 
<img width="602" alt="Screenshot 2024-10-01 at 8 24 31 PM" src="https://github.com/user-attachments/assets/c76686e0-b957-429e-a7e4-32259e004ad1">

PR #7138, opened 2024-10-02
fix(menu-toggle): split button alignment/spacing

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7137

Ran regression tests against menu-toggle, toolbar, table, and data-list and everything passed except the split button toggle in the toolbar in this example is misaligned now - https://patternfly-pr-7138.surge.sh/components/data-list/html-demos/expandable-demo/

Looks like it's misaligned when the toolbar items are in a toolbar__group, which uses baseline alignment. If the items are not in a toolbar__group, alignment is fine since the parent is toolbar__content-section which uses "start" alignment.

I'd say we hold off on this until post-v6 and spend some time cleaning it up.

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -115,7 +115,7 @@
   // Split button, controls, check
   --#{$menu-toggle}__button--BackgroundColor: transparent;
   --#{$menu-toggle}__button--Gap: var(--pf-t--global--spacer--gap--text-to-element--default);
-  --#{$menu-toggle}__button--AlignSelf: baseline;
+  --#{$menu-toggle}__button--AlignSelf: center;
   --#{$menu-toggle}__button--PaddingInlineStart: var(--pf-t--global--spacer--control--horizontal--plain);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -411,7 +411,7 @@
   &.pf-m-primary,
   &.pf-m-secondary {
     // Reduce the padding before/after the border unless it's an icon/check-only element
-    .#{$menu-toggle}__button:not(:has(.#{$menu-toggle}__toggle-icon:only-child)),
+    .#{$menu-toggle}__button:not(:has(.#{$menu-toggle}__controls:only-child)),
```
mcoker, line 414: This was just targeting the wrong element. Needs to look for `__controls` as an only child (meaning no `__text` sibling). It's effectively the same style as this one, but in `:not()` to target the inverse https://github.com/patternfly/patternfly/blob/689a71891bac6823c3dc2bd1406ea713f7037e3d/src/patternfly/components/MenuToggle/menu-toggle.scss#L486

mcoker:
Closing for now, @sg00dwin is looking to fix the alignment issue. I'll open a separate PR for the fix in this PR for the inline padding issue.

Issue #6933, opened 2024-07-30
Bug - [MenuToggle] - [dark theme background color is different from normal Button]


jelly:
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [ ] Patternfly 6  
- [ ] other

We use a MenuToggle and a Button next to each other and in a dark theme the background color of the MenuToggle is different from the Button.

![image](https://github.com/user-attachments/assets/b80849cc-d70a-47f7-85f5-ee9f8450247e)

**How do you reproduce the problem?**

I can't get codesandbox to work with a dark theme so the issue basically is:

```
<>
                  <MenuToggle
                    variant="secondary"
                    icon={<StarIcon />}
                    ref={toggleRef}
                  />
                    <Button
                      variant="plain"
                      icon={<CheckIcon  />}
                    />
</>
```

**Expected behavior**

I expect both background colors to be consistent and the same colour.

**Is this issue blocking you?**

No, we will apply a CSS workaround.

**Screenshots**

See above.

**What is your environment?**
 - OS: Linux
 - Browser chrome

**What is your product and what release date are you targeting?**

Cockpit-files

**Any other information?**

mcoker:
Thanks for the bug @jelly. If you have a working override, I think that's the best solution for this. We're code frozen and in the final stage of testing for the v5.4 release, and that will likely be our last v5 release until we release v6.

garrett:
Still a problem

PR #7140, opened 2024-10-03
fix(menu-toggle): sets secondary menu-toggle background-color transparent, except for disabled state

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/6933

The V5 dark theme Menu toggle secondary variant `background-color` was incorrect. This sets it to transparent

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -64,6 +64,9 @@
   --#{$menu-toggle}--m-secondary--Color: var(--#{$pf-global}--primary-color--100);
   --#{$menu-toggle}--m-secondary--BorderRadius: var(--#{$pf-global}--BorderRadius--sm);
   --#{$menu-toggle}--m-secondary--BackgroundColor: transparent;
+  --#{$menu-toggle}--m-secondary--hover--BackgroundColor: transparent;
+  --#{$menu-toggle}--m-secondary--focus--BackgroundColor: transparent;
+  --#{$menu-toggle}--m-secondary--active--BackgroundColor: transparent;
   --#{$menu-toggle}--m-secondary--before--BorderWidth: var(--#{$pf-global}--BorderWidth--sm);
   --#{$menu-toggle}--m-secondary--hover--before--BorderWidth: var(--#{$pf-global}--BorderWidth--md);
   --#{$menu-toggle}--m-secondary--focus--before--BorderWidth: var(--#{$pf-global}--BorderWidth--md);
@@ -72,6 +75,7 @@
   --#{$menu-toggle}--m-secondary--hover--before--BorderColor: var(--#{$pf-global}--primary-color--100);
   --#{$menu-toggle}--m-secondary--focus--before--BorderColor: var(--#{$pf-global}--primary-color--100);
   --#{$menu-toggle}--m-secondary--active--before--BorderColor: var(--#{$pf-global}--primary-color--100);
+  --#{$menu-toggle}--m-secondary--m-expanded--BackgroundColor: transparent;
   --#{$menu-toggle}--m-secondary--m-expanded--Color: var(--#{$pf-global}--primary-color--100);
   --#{$menu-toggle}--m-expanded__toggle--m-secondary--before--BorderColor: var(--#{$pf-global}--primary-color--100);
   --#{$menu-toggle}--m-expanded__toggle--m-secondary--before--BorderWidth: var(--#{$pf-global}--BorderWidth--md);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -246,6 +246,8 @@
   &.pf-m-secondary {
     --#{$menu-toggle}--Color: var(--#{$menu-toggle}--m-secondary--Color);
     --#{$menu-toggle}--m-expanded--Color: var(--#{$menu-toggle}--m-secondary--m-expanded--Color);
+    
+    background-color: var(--#{$menu-toggle}--m-secondary--BackgroundColor);
```
srambach, line 250: Our model has been to assign the modifier variable to the base variable for the property, although we've had a little bit of discuss around getting away from this. For now I think I'd prefer to stay consistent with the model.

srambach (commented):
Good catch! Just one question about using our base variable. Thanks!

src/patternfly/components/MenuToggle/menu-toggle.scss
sg00dwin: Sounds good, I've updated the pr. This required the `hover` `focus` `active` and `expanded` variables to be explicitly set.  Let me know if this isn't what's preferred.

sg00dwin: One item I noticed after pushing my changes is github is recognizing a [newline addition at the end of file](https://github.com/patternfly/patternfly/pull/7140/files#diff-62015d4798b534885d858ec2db1ee853e67d6eedcfde5142cd4f4364e88cbd52R639). 

My VS code settings for `Insert Final Newline` is not checked so I'm not sure why that's being added. Any ideas? 

@mcoker would you know? Do the PF devs use a shared [EditorConfig](https://editorconfig.org/) file, that I should use, for consistency?

mcoker: We don't have a shared editor config, but we do put an EOF at the end of files by default, unless there is a specific reason not to. A reason not to might be something like you want to control/remove the whitespace in the content of some element, and the content comes from a few `.hbs` files being included in one another, so you need to strip the `.hbs` files of their whitespace. But that case is pretty few and far between.

It's not a big deal to omit the newline, but if possible you should include it just for consistency's sake. It also can make the markup (or whatever the file is) more readable if there is a new line at the end of it. Otherwise you may get something like this:

```
// foo.hbs
<div>
  <span>foo</span>
</div> // no newline at EOF here
```

```
// bar.hbs
<div>
  <span>bar</span>
</div>
```

```
// foobar.hbs - the rendered HTML from the 2 files above included in some component example
<div>
  <span>foo</span>
</div><div> // these are on the same line from the lack of a newline at EOF in foo.hbs
  <span>bar</span>
</div>
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -280,6 +289,11 @@
       --#{$menu-toggle}--m-secondary--before--BorderColor: var(--#{$menu-toggle}--m-expanded__toggle--m-secondary--before--BorderColor);
       --#{$menu-toggle}--m-secondary--before--BorderWidth: var(--#{$menu-toggle}--m-expanded__toggle--m-secondary--before--BorderWidth);
     }
+
+    &:disabled,
+    &.pf-m-disabled {
+      --#{$menu-toggle}--BackgroundColor: var(--#{$menu-toggle}--disabled--BackgroundColor);
+    }
```
mcoker, line 296: As far as I can tell, this block isn't doing anything. Do you know if it's needed for any particular variation/use of the secondary toggle?

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -622,4 +636,4 @@
 
 @include pf-v5-theme-dark {
   @include pf-v5-theme-dark-menu-toggle;
-}
+}
```
mcoker, line 639:
```suggestion
}

```

mcoker (commented):
LGTM! Just one question about the disabled block, and a nit to add the newline back to the end. It's not a big deal but it would be great to get clarification around the disabled block prior to merging.

src/patternfly/components/MenuToggle/menu-toggle.scss
sg00dwin: You are correct, it's not needed after switching the `hover` `focus` `active` and `expanded` variables to be explicitly set.

mcoker (approved):
NICE!

srambach (approved):
L 🎃 TM!

PR #7149, opened 2024-10-16
Chart tooltips

### LINKED ISSUE #7148: Chart tooltips
Chart tooltips don't have rounded corners for PF v6

dlabrecq:
Charts should have rounded corners for PF v6

Fixes https://github.com/patternfly/patternfly/issues/7148

Screenshot 2024-10-16 at 4 26 39 PM

### DIFF
--- a/src/patternfly/patternfly-charts.scss
+++ b/src/patternfly/patternfly-charts.scss
@@ -293,10 +293,10 @@ $chart: #{$pf-prefix} + 'chart';
   --#{$chart}-threshold--stroke--Width: 1.5;
 
   // Tooltip
-  --#{$chart}-tooltip--corner-radius: 0;
+  --#{$chart}-tooltip--corner-radius: 6;
   --#{$chart}-tooltip--pointer-length: 10;
   --#{$chart}-tooltip--Fill: var(--#{$chart}-global--Fill--Color--200);
-  --#{$chart}-tooltip--flyoutStyle--corner-radius: 0;
+  --#{$chart}-tooltip--flyoutStyle--corner-radius: 6;
   --#{$chart}-tooltip--flyoutStyle--stroke--Width: 0;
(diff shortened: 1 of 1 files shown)


src/patternfly/patternfly-charts.scss
```diff
@@ -293,10 +293,10 @@ $chart: #{$pf-prefix} + 'chart';
   --#{$chart}-threshold--stroke--Width: 1.5;
 
   // Tooltip
-  --#{$chart}-tooltip--corner-radius: 0;
+  --#{$chart}-tooltip--corner-radius: 6;
```
mcoker, line 296: Not sure if you have a good way to do this on the react side, but I wonder if you could use the `--pf-t--global--border--radius--small` token here (and for `--#{$chart}-tooltip--flyoutStyle--corner-radius` below), and strip the unit in react-charts when you reference `chart_tooltip_corner_radius.value` and `chart_tooltip_flyoutStyle_corner_radius.value`?

FWIW, here's the tooltip style that references that same token
https://github.com/patternfly/patternfly/blob/b93029abee9117da890c05f219412e3c423d57b4/src/patternfly/components/Tooltip/tooltip.scss#L16

mcoker (approved):
LGTM! Left a comment in case it's helpful. ~~Feel free to merge whenever you think it's ready or~~ re-request a review if you push updates and you want another review 👍

src/patternfly/patternfly-charts.scss
dlabrecq: I don't think it's a good idea to include units here -- Victory cannot use them. This would unnecessarily introduce a mix of numbers and strings in react-tokens. Not only would be forced to later strip units, but anyone using a custom tooltip would need to do the same.

dlabaj (approved):
LGTM

Issue #7177, opened 2024-10-24
Tile - should be deprecated


tlabaj:
We deprecated the react implementation of tile.  Can we deprecate the HTML version.

PR #7178, opened 2024-10-24
chore(tile): mark tile as deprecated

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7177


tlabaj (approved):

evwilkin (changes requested):
To work as intended we'll want to move this file into a `deprecated` directory at `/components/Tile/deprecated/`

[Current logic in the documentation-site](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-site/patternfly-docs/patternfly-docs.source.js#L67-L71) looks for Core examples within a `deprecated` directory, similar to how ApplicationLauncher was in v5:
- https://github.com/patternfly/patternfly/tree/v5/src/patternfly/components/AppLauncher/deprecated

This changes the tab name to "HTML-deprecated" and unblocks the "deprecated" label in the sidenav [when not combined with a "React" or "Core" tab](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/components/sideNav/sideNav.js#L94-L95) (in the case that a deprecated implementation also has a newer implementation documented). 

FYI - The frontmatter here will enable the deprecation alert at the top of the Core docs page.

mcoker:
@evwilkin ah hah, thanks!! Updated - https://patternfly-pr-7178.surge.sh/components/tile

evwilkin (approved):
🍕

PR #7182, opened 2024-10-28
chore(github): update issue templates

srambach:
Removes alpha bug template and modifies the other templates to assign types rather than labels.


mcoker (approved):
L🎅TM!

PR #7186, opened 2024-10-31

mcoker:
Only 1 valid update - the first one ("pf-core__components_toolbar_html_sticky-toolbar_0_document_2_desktop.png"), updated with the new shadows that don't go beyond other edges.

Everything else is a false positive from random artifacts around breadcrumbs that we haven't figured out yet.

[BackstopJS Report.pdf](https://github.com/user-attachments/files/17589428/BackstopJS.Report.pdf)


srambach (approved):
Looks 🦃 to me. I like that the new screenshots seem to remove the artifacts.

Issue #7105, opened 2024-09-24
Datalist -  There is not spacing above the action buttons when datalist is compact


tlabaj:
**Please describe the issue**

There is no spacing above the action buttons when the Data list has the `pf-m-compact` modifier applied.

**Are there visuals for this issue? Please provide screenshots**

https://core-staging.patternfly.org/components/data-list#compact

![Screenshot 2024-09-24 at 4 14 12 PM](https://github.com/user-attachments/assets/0e99ed64-5320-43af-aa0a-5f0e2b72bdde)

mcoker:
Per @lboehling
* Compact data lists should use small/compact actions, so we should update our examples
* The data list actions element can be updated to use the compact button's vertical spacer

It will probably look like this, but design's feedback is "it still is pretty tight but i'd argue that if you need a whole bunch of buttons in your data list, you probably shouldn't be using a compact variant"

![Screenshot 2024-09-24 at 4 22 33 PM](https://github.com/user-attachments/assets/a77846d8-6021-4af1-8b31-9e57d19ddf55)

PR #7189, opened 2024-11-01
fix(data-list): add spacing to compact data-list buttons

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/7105

- Update DataList compact examples to use small buttons
- Add top and bottom padding to small buttons

Backstop tests
https://drive.google.com/drive/folders/1aKh3N42bPbMjsfkTTbGOrQr8zUO_YHYn?usp=sharing

**Before**
Screenshot 2024-11-01 at 6 13 52 PM

**After**
Screenshot 2024-11-01 at 6 13 32 PM

**Before**
Screenshot 2024-11-01 at 6 51 25 PM

**After**
Screenshot 2024-11-01 at 6 50 01 PM

### DIFF
--- a/src/patternfly/components/DataList/data-list-item-action.hbs
+++ b/src/patternfly/components/DataList/data-list-item-action.hbs
@@ -6,6 +6,7 @@
     {{> @partial-block}}
   {{else}}
     {{> menu-toggle
+      menu-toggle--IsSmall=data-list--IsCompact
       menu-toggle--IsPlain=true
       menu-toggle--HasKebab=true
       menu-toggle--id=(dasherize data-list--id data-list-item--id 'menu-toggle')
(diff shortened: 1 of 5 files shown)


src/patternfly/components/DataList/data-list-item-action.hbs
```diff
@@ -6,6 +6,7 @@
     {{> @partial-block}}
   {{else}}
     {{> menu-toggle
+      menu-toggle--IsSmall=data-list--IsCompact
```
mcoker, line 9: 👍

mcoker (commented):
LGTM!! Can you also update the actions in this example to use the small variation? https://patternfly-pr-7189.surge.sh/components/data-list/html/expandable-compact/

One other thing, I love the hbs `data-list--IsCompact` parameter - totally optional or could be a follow up, but it would be great if that could set `.pf-m-compact` as well. Otherwise it's a little confusing that a `.pf-m-compact` data-list also needs `data-list--IsCompact` to attempt to resize its buttons. As far as I can tell there are only 4 compact data lists in our docs that would need to be updated.

srambach (approved):
Looks great! 👍🏻

mcoker (approved):
🔥

Issue #7181, opened 2024-10-25
Bug - [Expandable section demo] - [Expandable text with character truncation demo is truncated incorrectly]


sg00dwin:
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

https://staging.patternfly.org/components/expandable-section/react-demos 
Text section truncation should be limited to first 10 characters. 

**Any other information?**
Related change that effects this issue. https://github.com/patternfly/patternfly/pull/6737

sg00dwin:
Close issue. It was determined this bug is related to unintended returns within the [react demo text content](https://github.com/patternfly/patternfly-react/issues/11225).

PR #7192, opened 2024-11-05
fix(expandable section): updated white-space when using truncate with and expandable section

sg00dwin:
fixes: https://github.com/patternfly/patternfly/issues/7181

Examples using ` <Truncate content={content} style={{ maxWidth: '30ch' }} />`

Related pr https://github.com/patternfly/patternfly/pull/6737 sets `.#{$truncate}__start, .#{$truncate}__end`  to have `white-space: pre`

**Before**
Screenshot 2024-11-05 at 9 41 12 AM

**After**
Screenshot 2024-11-05 at 9 36 29 AM

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section.scss
+++ b/src/patternfly/components/ExpandableSection/expandable-section.scss
@@ -76,6 +76,11 @@
       // stylelint-enable
       overflow: hidden;
     }
+
+    .#{$truncate}__start,
+    .#{$truncate}__end {
+      white-space: nowrap;
+    }
   }
 }
 


mattnolting (commented):
🦅 👁️ Nice catch, Steve! 

# TLDR:
Truncation is baked into expandable section. `.pf-c-expandable-section.pf-m-truncate` shouldn't be used in conjunction with `.pf-c-truncate`, as they serve different purposes. If single line start/middle/end string truncation is needed, `pf-v6-truncate` would be used **without** `.pf-c-expandable-section.pf-m-truncate`. 

## There are several of things happening here: 

1) Truncation is [baked into expandable section](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/ExpandableSection/expandable-section.scss#L36) and set at `-webkit-line-clamp: 3`, suggesting that `pf-v6-c-expandable-section` and `pf-v6-c-trunate` components treat truncation differently. This example is obscure as it attempts to demo unique design patterns. One being truncation after line wrapping (3), the other being inline truncation. 

2) `pf-v6-truncate` preserves white space, disallows line-wrapping, and honors `<br>/returns` in text blocks. The supplied text contains returns, which will be interpreted as intentional by browsers. 

[Screenshot 2024-11-18 at 11 36 56 AM](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)

AFAIK, support doesn't exist to both preserve whitespace and prohibit text wrapping at returns. 

3) [Expandable text with character truncation](https://www.patternfly.org/components/expandable-section/react-demos/#expandable-text-with-character-truncation) applies `max-width: '10ch';` to content. 

Screenshot 2024-11-18 at 11 47 20 AM

This further confuses the example and should be removed.

sg00dwin:
Closing this since it was determined the [react expandable section demo](https://github.com/patternfly/patternfly-react/issues/11225) text content contains returns causing the incorrect behavior.

PR #7194, opened 2024-11-05
fix: address early a11y issues + clean up

nicolethoen:
addresses following a11y issues flagged by patternfly-a11y upgrade:

- card demo has invalid aria-controls values (the aria-controls on a table's expandable row toggle does not match the id of the expandable content)
- multiple table examples have invalid aria-controls values (the aria-controls on a table's expandable row toggle does not match the id of the expandable content)

Also cleans up:

- the styles on the card demo - status card expanded with popover example
- `role="row"` always being rendered (rather than only conditionally rendered), preventing a lot of `[Object, object]` appearing in the rendered html markup in table examples.


mcoker (approved):
L🎄TM

Issue #7199, opened 2024-11-06
Table draggable rows beta label not showing correctly


mcoker:
On https://www.patternfly.org/components/table/html/#draggable-rows

From https://github.com/patternfly/patternfly/pull/7119/files

"isBeta" renders as text, when it should be a badge

![Image](https://github.com/user-attachments/assets/5db1e3ce-4d20-4862-ba20-a0e946ddfb6c)

PR #7200, opened 2024-11-06
chore(table): fix draggable beta label

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7199


nicolethoen:
👀 it looks like the a11y tests are actually not turned off in the CI. do you want me to do that until they can all be addressed?

mcoker:
@nicolethoen ah darn. We can hold off on this PR for a minute - I'll spend maybe an hour this morning and see if I can fix those other a11y errors.

Issue #7190, opened 2024-03-20
[Stack/Split] - Easier interface to use


andrewballantyne:
**Is this a new component or an extension of an existing one?**
What is the existing component, if any? `Stack`, `Split` (maybe this extends to `Flex`) 

**Describe the feature**
A clear and concise description of the new feature. What is the expected behavior?

Can we get a smaller barrier to entry to using Stack and Split where the Item variant is not needed? Officially supporting a more simple interface.

```
<Stack hasGutter>
  <MyItem />
  <MyOtherItem />
</Stack>
```

instead of the more verbose:

```
<Stack hasGutter>
  <StackItem>
    <MyItem />
  </StackItem>
  <StackItem>
    <MyOtherItem />
  </StackItem>
</Stack>
```

The only use-case I see for using a `StackItem` is for `isFilled` -- not sure if there is a better way to get around that. Maybe that's your use-case for `StackItem`? 

**Are there visuals for this feature? If applicable, please include examples for each state and for varying widths**
Include screenshots or links to Marvel or other mockups.

**Any other information?**

mcoker:
Hey @andrewballantyne, sorry about that, we're starting to work on these types of issues now that v6 is released so I went ahead and put up a PR for the change if you'd like to leave a review - https://github.com/patternfly/patternfly/pull/7201

PR #7201, opened 2024-11-07
fix(split/stack): allow pf-m-fill to apply to any direct child

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7190

### DIFF
--- a/src/patternfly/layouts/Split/examples/Split.md
+++ b/src/patternfly/layouts/Split/examples/Split.md
@@ -87,13 +87,13 @@ import './Split.css'
 
 ## Documentation
 ### Overview
-The split layout is designed to position items horizontally, with one item filling the available horizontal space.
+The split layout is designed to position items horizontally.
 
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-v6-l-split` | `<div>`, `<section>`, or `<article>` | Initiates the split layout. |
-| `.pf-v6-l-split__item` | `<div>` | Initiates a split item. **Required** |
+| `.pf-v6-l-split` | `*` | Initiates the split layout. |
+| `.pf-v6-l-split__item` | `*` | Initiates a split item. |
 | `.pf-m-gutter` | `.pf-v6-l-split` | Adds space between children by using the globally defined gutter value. |
 | `.pf-m-wrap` | `.pf-v6-l-split` | Sets the split layout to wrap. |
-| `.pf-m-fill` | `.pf-v6-l-split__item` | Specifies which item(s) should fill the avaiable horizontal space. |
+| `.pf-m-fill` | `.pf-v6-l-split > *`, `.pf-v6-l-split__item` | Specifies which item(s) should fill the avaiable horizontal space. |
--- a/src/patternfly/layouts/Split/split.scss
+++ b/src/patternfly/layouts/Split/split.scss
@@ -12,10 +12,11 @@
   &.pf-m-wrap {
     flex-wrap: wrap; // make default in a breaking change release
   }
-}
 
-.#{$split}__item.pf-m-fill {
-  flex-grow: 1;
+  @at-root .#{$split}__item.pf-m-fill,
+  & > .pf-m-fill {
+    flex-grow: 1;
+  }
 }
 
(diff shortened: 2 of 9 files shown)


src/patternfly/layouts/Split/examples/Split.md
```diff
@@ -87,13 +87,13 @@ import './Split.css'
 
 ## Documentation
 ### Overview
-The split layout is designed to position items horizontally, with one item filling the available horizontal space.
+The split layout is designed to position items horizontally.
```
mcoker, line 90: AFAIK there is no expectation that any (or just one) item(s) need to fill the space.

src/patternfly/layouts/Split/examples/Split.md
```diff
@@ -87,13 +87,13 @@ import './Split.css'
...
 
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-v6-l-split` | `<div>`, `<section>`, or `<article>` | Initiates the split layout. |
-| `.pf-v6-l-split__item` | `<div>` | Initiates a split item. **Required** |
+| `.pf-v6-l-split` | `*` | Initiates the split layout. |
```
mcoker, line 95: This can be any element, `<Split />` in pf-react allows a `component` prop to let you use any element here.

src/patternfly/layouts/Split/examples/Split.md
```diff
@@ -87,13 +87,13 @@ import './Split.css'
...
+| `.pf-v6-l-split__item` | `*` | Initiates a split item. |
```
mcoker, line 96: This is not required - prior to this PR, it was only required for `.pf-m-fill`, but with this PR it isn't required at all.

src/patternfly/layouts/Split/split.scss
```diff
@@ -12,10 +12,11 @@
   &.pf-m-wrap {
     flex-wrap: wrap; // make default in a breaking change release
   }
-}
 
-.#{$split}__item.pf-m-fill {
-  flex-grow: 1;
+  @at-root .#{$split}__item.pf-m-fill,
+  & > .pf-m-fill {
```
mcoker, line 17: Now this rule renders as

```
.pf-v6-l-stack__item.pf-m-fill,
.pf-v6-l-stack > .pf-m-fill {
    flex-grow: 1;
}
```

andrewballantyne:
Is there a react rendering of this that I can test out my expectations @mcoker ?

mcoker:
@andrewballantyne sure! Here's a basic [`<Split />` codesandbox](https://codesandbox.io/p/sandbox/competent-keller-393jrh?file=%2Findex.js%3A12%2C19) and [`<Stack />` codesandbox](https://codesandbox.io/p/sandbox/heuristic-chatterjee-stpd99?file=%2Findex.js%3A15%2C38) with
* the CSS from this PR imported as `pf-item-update.css`
* updated the examples a little to show passing other children to the layout
* using `isFilled` on an item
* mimicking `isFilled` on a `<div>` (passing `className="pf-m-fill"`)

andrewballantyne:
Looks good to me @mcoker -- that's what I was looking for. I can't edit the sandboxes (don't have an account) but visually inspecting the examples looks to fit what I was after. Thanks!

srambach (approved):
Looks good - one question, would you want to adjust the custom workspace styling (that adds the dashed outline) to also work without the _item classes?
Also, is this change useful on the level layout?

mcoker:
@srambach sure I don't see why not. Added `.pf-m-fill` to level - just to confirm, was that what you were asking? We'll need a react follow up to add a prop for that. Also updated the level examples so the boxes are different heights since one of the things the level does is vertically center its items.

srambach (commented):
I think in discussions we decided we don't need to add a `pf-m-fill` on the level layout, but keep the local workspace styling to show that `__item` isn't required.

mcoker:
@srambach backed out the level layout's `.pf-m-fill` change https://github.com/patternfly/patternfly/pull/7201/commits/a2bfe807f8b21d1d92c0373e665fcfa61f6f59d6

Issue #7072, opened 2024-09-12
DPO Typography Updates -- Content/Title Component updates? 


lboehling:
Made a few typography updates per recent DPO decisions. Tokens will be available once the latest figma tokens are pulled in via [#87](https://github.com/patternfly/design-tokens/issues/87). Decided to bump our 22px font size back up to 24px, and reduce the font weight in `global--font--weight--heading--default.` The weights and sizes should be handled by base token updates. 

Want to double check those are pulled in correctly. The DPO group also decided that only the H1 (or largest title if no H1 is present) on the page should be bolded. Our current titles/headings are all set to `global--font--weight--heading--default.`  We have `global--font--weight--heading--bold` that should be used to enable that. I'm not sure if we can just update that token for our H1 headings, or if it should be a modifier on the title component? 

Will create a follow-up design guidelines/typography page content issue. 

FYI @mcoker @andrew-ronaldson

mcoker:
@lboehling @andrew-ronaldson some thoughts:

* The content component styles headings based on heading level, so an `h1` always gets certain styles. You can't have an `h1` that looks like an `h4`, for example.
* The title component styles headings based off of the size modifier/class used. So you can have an `h1` that looks like an `h4`.
* The largest title may not always be an `h1`, depending on the markup on the rest of the page, so we can't just style by the element used.
* The largest title on the page may not always be a specific size (I'm assuming this is the case, please correct me if I'm wrong, or that isn't the design intent). Maybe on one page you have some kind of Super-Cool-Awesome-Hero-Box heading that's ginormous, and on another page, the main title is smaller, but both are the largest/most-prominent on their respective pages.

That says to me that we likely need some sort of modifier class for the content component and the title component that you can apply to whatever heading level and whatever size variation that says "This is my page title" and applies a bold font-weight. We could also use that class to apply any additional styles we decide later we want to add to whatever the "page title" is. wdyt?

PR #7212, opened 2024-11-11
feat(title/content): add page title variation

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7072

* Adds `.pf-m-page-title` to the content component and title component
* Can be applied to any heading level/size and it will override the font-weight
* Updates our standard full-page demo template to add the page title modifier to the main heading
  * Some demos may not use that template and would need to be updated manually
  * We could add the page-title modifier to the "Hello world" heading in main content component examples - wdyt @lboehling @andrew-ronaldson?

Links:
https://patternfly-pr-7212.surge.sh/components/content
https://patternfly-pr-7212.surge.sh/components/title#page-title
https://patternfly-pr-7212.surge.sh/components/page/html-demos/basic/
https://patternfly-pr-7212.surge.sh/components/data-list/html-demos/basic/

### DIFF
--- a/src/patternfly/components/Content/content--element.hbs
+++ b/src/patternfly/components/Content/content--element.hbs
@@ -1,7 +1,10 @@
 <{{content--element--type}} class="{{#unless content--element--ExcludeClass}}{{pfv}}content--{{content--element--type}}{{/unless}}
-  {{~#if content--element--IsEditorial}} pf-m-editorial{{/if}}
-  {{~#if content--element--IsPlain}} pf-m-plain{{/if}}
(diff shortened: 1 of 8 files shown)


srambach (approved):
LGTM. Will we want an org followup to apply/clean up the page titles?

mcoker:
@srambach I reckon so! I'd defer to @lboehling since she understands the usage requirements better than I do.

lboehling (approved):
This looks great! Thank you, @mcoker 🤩 I think adding it to the "Hello world" title in the example makes sense. I also +1 for a follow-up issue to make sure the styles on the org are cleaned up/updated.

lboehling (approved):
lgtm!!

PR #7213, opened 2024-11-12
fix(menu toggle): correct spacing for split menu-toggle__toggle-icon that contains svg for icon

sg00dwin:
Fixes https://github.com/patternfly/patternfly/issues/7137
The react `MenuToggle` component uses an `<svg>` for the `menu toggle` icon, instead of an `<i>` used by core.
Removing `display: flex` from `pf-v6-c-menu-toggle__toggle-icon` because it causes a bug  where the `svg` `CaretDownIcon` is misaligned with the text and the button height to grow taller.

Before and after
Screenshot 2024-11-12 at 1 42 43 PM

https://patternfly-pr-7213.surge.sh/components/menus/menu-toggle#split-toggle

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -500,14 +500,6 @@
   flex-wrap: nowrap;
 }
 
-// - Menu toggle controls - Menu toggle icon
-.#{$menu-toggle}__controls,
-.#{$menu-toggle}__toggle-icon {
-  display: flex;
-  align-items: center;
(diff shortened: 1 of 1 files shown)


mcoker (approved):
LPTM! Also love the simplicity of this change. Sometimes these alignment issues turn into huge changesets with lots of sweeping changes. I really appreciate narrowing this down to a straight forward, easy fix 👍

mcoker:
Also meant to mention that I ran backstop and all is good. Also made this change in the react verified it fixes the issue we raised.

PR #7214, opened 2024-11-13

mcoker:
Ran when reviewing https://github.com/patternfly/patternfly/pull/7213

* Not sure what's up with the first draggable rows example. I confirmed what's on staging matches the local server. The screenshots for that one are likely quite out of date because it had an improper use of `isBeta` in the title that you can still see on org (https://www.patternfly.org/components/table/html/draggable-rows-example-isbeta/) and that was fixed yesterday (https://github.com/patternfly/patternfly/pull/7200), so the backstop screenshots for that example have been saved under the filename with "isBeta" in the title up until yesterday, so this backstop run is probably comparing against a very old version of that example.
* Code editor changes are from changing the `__code` element from `<code>` to a `<div>` to match react and our own documentation - `<code>` preserved whitespace/newlines in the example and the `<div>` does not.

Everything else looks to be intentional example changes, new/renamed examples lacking a reference image, and random false positives.

[BackstopJS Report.pdf](https://github.com/user-attachments/files/17735093/BackstopJS.Report.pdf)


srambach:
There are lots of lemur messages at the end of the pdf - did they not load?
Also, the noise is back, and some of them - like pf-core__components_menus_menu-toggle_html_splittoggle_0_document_0_mobile.png - seem like they have an excessive amount of jankiness. Though maybe it's only because I'm viewing it zoomed in? Do you see it on the original?

srambach (approved):
Looks 🦃 to me

Issue #7215, opened 2024-11-13
Bug - Label - issue with min-width and short labels


mcoker:
Looks like as part of https://github.com/patternfly/patternfly/pull/7176, we added a min-width of "3em" to labels, but labels can be < 3 characters. We have an example here - https://core-staging.patternfly.org/components/table/html/nested-column-headers-example/

And for comparison, this is what it should look like - https://www.patternfly.org/components/table/html/nested-column-headers-example/

cc @sg00dwin

PR #7217, opened 2024-11-14
fix(label): update label min-width to match height

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/7215
Set label `min-width` to equal height

### DIFF
--- a/src/patternfly/components/Label/label.scss
+++ b/src/patternfly/components/Label/label.scss
@@ -6,7 +6,7 @@
   --#{$label}--PaddingBlockEnd: var(--pf-t--global--spacer--xs);
   --#{$label}--PaddingInlineStart: var(--pf-t--global--spacer--sm);
   --#{$label}--MaxWidth: 100%;
-  --#{$label}--MinWidth: 3em;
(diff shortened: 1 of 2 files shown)


srambach (approved):
This is fine, but I do wonder if we really ever want them to be narrower than they are tall? (ie a circle?) Maybe something to ask design and **circle** back on 😆

sg00dwin:
> This is fine, but I do wonder if we really ever want them to be narrower than they are tall? (ie a circle?) Maybe something to ask design and **circle** back on 😆

I reached out to Andrew and Lucia and got the thumbs up  to maintain a `min-width` so that a label will not shrink less than it's height and maintain a circle

src/patternfly/components/Label/label.scss
```diff
@@ -203,6 +203,7 @@
   --#{$label}--m-compact--PaddingInlineStart: var(--pf-t--global--spacer--sm);
   --#{$label}--m-compact--FontSize: var(--pf-t--global--font--size--body--sm);
   --#{$label}--m-compact--m-editable--TextUnderlineOffset: #{pf-size-prem(1px)};
+  --#{$label}--m-compact--MinWidth: calc(var(--#{$label}--m-compact--FontSize) * var(--pf-t--global--font--line-height--body));
```
srambach, line 206: Should compact use the same formula, adding on the block paddings just in case someone adjusts those variables to something other than zero?

mcoker (approved):
LPTM!

Issue #7001, opened 2024-08-22
Bug - [ToggleGroupItems] - [ToggleGroupItems is in foreground when used in Wizard]


avitova:
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  

When using ToggleGroupItems in a Wizard on a screen with a more narrow size, the Item that is `selected` is kept on the screen over the list of steps in Wizard.

**How do you reproduce the problem?**
[...]

PR #7218, opened 2024-11-15
fix: fixes z-index issue with wizard navigation.

dlabaj:
Fixes issue #7001 by setting the z-index correctly for the wizard navigation.

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -126,7 +126,7 @@
   --#{$wizard}__toggle--m-expanded__toggle-icon--Rotate: 180deg;
 
   // Nav
-  --#{$wizard}__nav--ZIndex: var(--pf-t--global--z-index--xs);
+  --#{$wizard}__nav--ZIndex: var(--pf-t--global--z-index--s);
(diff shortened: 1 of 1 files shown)


srambach (approved):
🧙🏻‍♂️ lgtm

mcoker:
I believe the reason it was using the other z-index was so that the shadow from the nav didn't sit on top of the footer when expanded on mobile and the nav extends the height of the wizard and has a scrollbar. For reference:

This is the shadow when it doesn't extend the full height of the wizard, lovely and perfect

Screenshot 2024-11-15 at 3 57 09 PM

And here's what it looks like currently on org when the nav extends the full height. Note the shadow is hidden underneath the footer. That was by design when we created the wizard.

Screenshot 2024-11-15 at 3 56 56 PM

And here's what it looks like in this PR

Screenshot 2024-11-15 at 3 50 05 PM

Looks OK to me personally but would love design's feedback. WDYT @lboehling or @andrew-ronaldson? Is this OK or think we should look at it some more and either try to hide the shadow again or make some other kind of update?

andrew-ronaldson:
I think hide the shadow like the Pre-PR screenshot shows.

PR #7224, opened 2024-11-18
fix(Page): limit width on section and example cleanup

### LINKED ISSUE #6801: Page - section grouping follow-ups

### LINKED ISSUE #7216: Bug - Navigation - Long labels don't wrap
**Describe the problem**
This is an issue with PatternFly 6. The navigation "container" gets wider when you expand nav sections with long nav item labels, because they aren't wrapping.

On [PatternFly.org](https://www.patternfly.org/), expand the "Ux writing" section when your browser is fullscreen or almost fullscreen. This can also be seen under "PatternFly AI > Chatbot". Note that this doesn't seem to occur for me in smaller windows (screenshot below).

**Expected behavior**
The navigation container should stay the same size. The longer navigation labels should wrap.

No. 

**Screenshots**
Default/expected size:
</img>

Size grows with nav label:
</img>

When I make my browser window small, the label does wrap as expected: 
</img>

**What is your environment?**
Mac OS, Chrome

srambach:
`.pf-m-limit-width` was not working at XL viewport because there was a calc() involving `auto`. Instead of a calc, I manually set the max-width value for the XL breakpoint. EDIT: in light of #7216, I kept the calc but changed `auto` to be the same width for XL as it was for the other breakpoints. This means the sidebar no longer stretches, and so the items will now wrap properly.

Added the main-subnav page section to the sections that can be width limited.

The `.pf-m-fill` example wasn't showing the effect because `.pf-m-fill` wasn't added to the page main container, so this has been fixed and documented.

The sidebar example didn't have the correct modifier for insets/no inset and the footer section was preventing the fill section from filling.

Changed the selector in the local css to catch the in-page examples that are inline and limit their height so that the examples page isn't super tall.

[Page examples](https://patternfly-pr-7224.surge.sh/components/page)

Fixes #6801 
Fixes #7216

### DIFF
--- a/src/patternfly/components/Page/page-main-container.hbs
+++ b/src/patternfly/components/Page/page-main-container.hbs
@@ -1,4 +1,4 @@
-<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}}{{page-main-container--modifier}}{{/if}}{{#if page-main-container--IsFilled}} pf-m-fill{{/if}}" tabindex="-1"
+<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}} {{page-main-container--modifier}}{{/if}}{{#if page-main-container--IsFilled}} pf-m-fill{{/if}}" tabindex="-1"
   {{#if page-main-container--attribute}}
     {{{page-main-container--attribute}}}
   {{/if}}>
--- a/src/patternfly/components/Page/page-main-container.hbs
+++ b/src/patternfly/components/Page/page-main-container.hbs
@@ -1,4 +1,4 @@
-<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}} {{page-main-container--modifier}}{{/if}}" tabindex="-1"
+<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}}{{page-main-container--modifier}}{{/if}}{{#if page-main-container--IsFilled}} pf-m-fill{{/if}}" tabindex="-1"
   {{#if page-main-container--attribute}}
     {{{page-main-container--attribute}}}
   {{/if}}>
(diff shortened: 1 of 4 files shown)


srambach:
[Backstop report](https://drive.google.com/file/d/1KFUyLDzP1uYIIsTkGawlSceLzDD6-0Vs/view?usp=sharing)

A lot of them are rending on the logo
Some legit differences in page and navigation that looks as intended
Differences in datalist that are unrelated
File upload wrapping difference?

src/patternfly/components/Page/page-main-container.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}} {{page-main-container--modifier}}{{/if}}" tabindex="-1"
+<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}}{{page-main-container--modifier}}{{/if}}{{#if page-main-container--IsFilled}} pf-m-fill{{/if}}" tabindex="-1"
```
mcoker, line 1: Not sure if you meant to remove that space?

```suggestion
<div class="{{pfv}}page__main-container{{#if page-main-container--modifier}} {{page-main-container--modifier}}{{/if}}{{#if page-main-container--IsFilled}} pf-m-fill{{/if}}" tabindex="-1"
```

mcoker (commented):
Just one little space nit, otherwise LGTM!

mcoker (approved):
🥷🤘

PR #7238, opened 2024-11-26
chore(stalebot): update stale.yml

### LINKED ISSUE #7188: Update stalebot to use new labels/types
Update stalebot to use new labels/types.

srambach:
Remove accessibility and design approved from the exemptions for stalebot.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -11,7 +11,7 @@ jobs:
         with:
           days-before-stale: 60
           days-before-close: -1
-          exempt-issue-labels: 'accessibility,breaking change,design approved,pinned'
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):
This just prevents the "stale" label from being applied now correct, since we don't auto-close issues any more?

wise-king-sullyman (approved):

srambach:
> This just prevents the "stale" label from being applied now correct, since we don't auto-close issues any more?

@thatblindgeye we don't close them anymore; they just get a "stale" label. So now `accessibility` would get the stale label, but won't be closed. (and we changed the label to `a11y` anyway)

mcoker (approved):
🐸

PR #7244, opened 2024-12-03
fix(table): targeted control-row sort button margining

mattnolting:
closes #7220 

TLDR: Issue is being caused by table sort vertical alignment (should be `vertical-align: top`) in tbody and negative margining. 

[Backstop report](https://drive.google.com/file/d/1vVn1BdRJeiG6sThWNFXL42A-GS9PCUNj/view?usp=sharing)

Because tables may contain elements with varying `line-height`/`font-size`, all `th/td` cells, whether in `thead` or `tbody` require a consistent and explicit `line-height` to properly align vertically.

PR #7050, opened 2024-09-09
fix(table): removed first/last child padding, added inset

mattnolting:
closes #7032  
closes [#11058](https://github.com/patternfly/patternfly/issues/7112)

Updated [Table Iso Backstop Report](https://drive.google.com/file/d/1aj7jx_rWbyy5eCK0738ZdcKh_AGa_bBx/view?usp=sharing)
Updated [Full Backstop Report](https://drive.google.com/file/d/1NKGM1Q2wz6V3Buyvj4w4Cx6BKRAuOf6R/view?usp=sharing)

Examples: https://patternfly-pr-7050.surge.sh/components/table

# TLDR:
Floating `:first/:last-child` pseudos result in a huge number of selector matching attempts. Attaching these pseudos to a selector results in a massive reduction in matching attempts. 

The cell padding and button fitting updates are the result of a somewhat hidden bug. `thead` row heights are affected erratically by using negative margining on `table__button` cells. See below:

**Before**

Screenshot 2024-09-17 at 10 56 47 AM

Screenshot 2024-09-17 at 10 56 05 AM

Screenshot 2024-09-17 at 10 53 34 AM

Screenshot 2024-09-17 at 10 52 39 AM

**After**

Screenshot 2024-09-17 at 10 57 35 AM

Screenshot 2024-09-17 at 10 57 23 AM
Screenshot 2024-09-17 at 10 57 10 AM
Screenshot 2024-09-17 at 10 56 58 AM

~~Table :first/:last-child css declarations are relics that should be standardized. As per a recent conversation w/ @lboehling, :first/:last-child th/td padding definitions can/should be simplified, while inset updates should be optional.~~

~~Also, toggles, checkboxes, radios, actions, etc do not need special spacing.~~

These need to be updated

## Performance impact

This PR 
* Reduces match attempts by **4,361,984**
* Improves performance by 2%
* Reduces blocking time by 50ms

**Before**

Screenshot 2024-09-10 at 11 57 31 AM

**After**

Screenshot 2024-09-10 at 11 49 02 AM

**Before**

Screenshot 2024-09-10 at 12 03 46 PM

**After**

Screenshot 2024-09-10 at 12 05 26 PM

[Backstop report](https://drive.google.com/file/d/1XSd0422s0tgllaC3dVdIir6kcamoFDiU/view)

PR #7246, opened 2024-12-04
fix(table): isolated sort button negative margins to thead

### LINKED ISSUE #7220: Bug - Table - misalignment when only some headers are sortable
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Table with sortable headers

Provide steps to reproduce. A [codepen](http://codepen.io) or [codesandbox](https://codesandbox.io/) demonstrating the problem is appreciated.

Go to the [React nested expandable table example](https://www.patternfly.org/components/table/#expandable-with-nested-table) and make only one of the header cells sortable. You can just pass `sort={{columnIndex: 0}}` to one of the `<Th` components that has a visible text label, do this for both the outermost table and the nested table.

Notice that the text for the sortable header button is misaligned compared to the NON-sortable header cells (see screenshot below):

Image

**Expected behavior**
All cells would be veritcally aligned regardless of the amount of sortable header cells

**Screenshots**

**What is your environment?**

This was found during updates being made to RHOAI for v6 migration.

**Any other information?**

mattnolting:
Closes #7220 

Quick fix for table sort buttons in nested tables. Eventually we'll need to address the issues detailed in #7244 and #7050, but this will solve our immediate issue.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -406,6 +406,10 @@
     --#{$table}--cell--TextOverflow: ellipsis;
     --#{$table}--cell--WhiteSpace: nowrap;
 
+    .#{$table}__sort .#{$table}__button {
+      margin-block-end: calc(var(--#{$table}__button--PaddingBlockEnd) * -1);
+    }
+
     // stylelint-disable
     &.pf-m-nested-column-header {
       button:where(.#{$button}) {


mcoker (approved):
🐝🫵🍵🫃🌕
(Bee-you-tea-full?? It's a work in progress...)

thatblindgeye (approved):
🤩

PR #7248, opened 2024-12-04
fix(menu-toggle): add variable for menu toggle plain border radius for MUI theme support

sg00dwin:
Fixes https://github.com/patternfly/patternfly/issues/7223


mcoker (approved):
🚀

Issue #7245, opened 2024-12-04
Backstop - configure to use light theme by default


mcoker:
There was a wonderful update to docs framework that enables dark theme automatically based on your OS preference (https://github.com/patternfly/patternfly-org/pull/4389/) but our visual regression testing needs to target light/dark theme specifically, and not based on OS preference.

Default should be light theme, and dark theme should be enabled via `yarn backstop:[test/approve]:dark` via `backstop --dark` https://github.com/patternfly/patternfly/blob/19ce9765f6ce4cec0b273a5336333398f4a65f20/package.json#L11

mcoker:
[Auto detect light/dark theme was disabled](https://github.com/patternfly/patternfly-org/pull/4424) for the time being in docs framework so this fix is no longer needed

PR #7279, opened 2025-01-09
chore(backstop): test light theme by default

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7245


dlabaj (approved):

mcoker:
Looks like the auto-detect light/dark theme was disabled in docs framework via https://github.com/patternfly/patternfly-org/pull/4424 so this fix is not needed at the moment.

Issue #7293, opened 2025-01-13
Bug - Form - Invalid CSS in `form.scss`


logonoff:
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Please give a clear and concise description of the problem. Which components are affected?

In this line of CSS, `none` is not a valid option for `background-color`

https://github.com/patternfly/patternfly/blob/fdfba00df2a534af46dba6a4218ee02acdec6e2a/src/patternfly/components/Form/form.scss#L265

**How do you reproduce the problem?**
Provide steps to reproduce. A [codepen](http://codepen.io) or [codesandbox](https://codesandbox.io/) demonstrating the problem is appreciated.

Use the `Form` component and then run a CSS validator

**Expected behavior**
A clear and concise description of the expected behavior.

The rule should either be removed or the value should be changed to `transparent`

**Is this issue blocking you?**
List the workaround if there is one.

No

**Screenshots**
If applicable, add screenshots to help explain the issue.

![Image](https://github.com/user-attachments/assets/5f349861-5869-4cda-b24a-8f67c0457b1c)

**What is your environment?**
 - OS: Fedora 41
 - Browser: Firefox
 - Version: 134.0

**What is your product and what release date are you targeting?**

OpenShift Management Console, targeting 4.19

PR #7301, opened 2025-01-14
fix(Form): remove invalid css in form.scss

logonoff:
Closes https://github.com/patternfly/patternfly/issues/7293

PR #7302, opened 2025-01-14
fix(form): remove broken label selection pseudo style

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7293

This rule has been in the stylesheet since the component was created. The comment when it was added was to hide the label selection on click, but I don't see that as an issue in any of the browsers we support, and we probably want the default behavior anyways.


mcoker:
Dupe of https://github.com/patternfly/patternfly/pull/7301/

PR #7304, opened 2025-01-14
feat(Page): remove extra margin on small viewports

### LINKED ISSUE #7300: Bug - Page  - Margins are too wide on mobile viewports
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

The margins on the `Page` are too wide on mobile, making content hard to read

Provide steps to reproduce. A [codesandbox](https://codesandbox.io/s/serverless-cherry-q9t3f ) demonstrating the problem is appreciated.

Load a PF6 page (e.g., the [dashboard demo](https://www.patternfly.org/patterns/dashboard/html-demos/basic/)) on your phone

**Expected behavior**

The margins on the main container are removed

**Screenshots**

**What is your environment?**
 - OS: Android
 - Browser: Firefox
 - Version: 134.0

OpenShift Management Console 4.19.0

**Any other information?**

logonoff:
Closes https://github.com/patternfly/patternfly/issues/7300

Before:
image

After:
image

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -61,7 +61,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}--masthead--main-container--GridArea: sidebar / sidebar / main / main; // no sidebar
   --#{$page}__main-container--MaxHeight: calc(100% - var(--pf-t--global--spacer--lg));
   --#{$page}__main-container--BackgroundColor: var(--pf-t--global--background--color--primary--default);
-  --#{$page}__main-container--MarginInlineStart: var(--#{$page}--inset); 
+  --#{$page}__main-container--MarginInlineStart: var(--#{$page}--inset);
   --#{$page}__main-container--MarginInlineEnd: var(--#{$page}--inset);
   --#{$page}__main-container--BorderRadius: var(--pf-t--global--border--radius--medium);
   --#{$page}__main-container--BorderWidth: #{pf-size-prem(4px)}; // TODO Change to be a page outline token
@@ -80,7 +80,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 
   // Limit width
   --#{$page}--section--m-limit-width--MaxWidth: calc(#{pf-size-prem(2000px)} - var(--#{$page}__sidebar--Width));
-  
+
   // Sticky
   --#{$page}--section--m-sticky-top--ZIndex: var(--pf-t--global--z-index--md);
   --#{$page}--section--m-sticky-top--BoxShadow: var(--pf-t--global--box-shadow--sm--bottom);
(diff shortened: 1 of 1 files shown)


andrew-ronaldson:
@lboehling how do you feel about removing the margins. I wonder if this should also reduce the padding on the page component?

lboehling (changes requested):
I think this looks good overall! 

one suggestion and one comment:

The sharp borders on the bottom of page content area look strange when the page body isn't filling the full height of the page. I wonder if on mobile we should default to/suggest having the page content area fill the height to avoid this. 
image
wdyt @srambach @andrew-ronaldson ? 

And I think the padding in the page component is fine for now. If we reduced it in the page body area on small viewports, we'd also want to reduce the page chrome padding on small viewports. We actually did this before in v5 but I remember it caused a lot of issues so we backed that decision out/decided to keep the padding consistent. @mcoker probably remembers more details on that!

mcoker:
@andrew-ronaldson @lboehling yeah any time we change the page section padding, we need to change the padding in some other components. Tables are a good example. If you remove page section padding around a table so the table row dividers span the width of the content area, since the page section padding and table padding are the same, the content aligns fine.

Screenshot 2025-03-04 at 10 12 08 AM

If we change the page section padding from 24px to 16px, but don't change the table padding, you can see that the content isn't aligned:

Screenshot 2025-03-04 at 10 12 56 AM

Changing the page section padding at a specific viewport just introduces some fragility because the padding of other components will likely also need to change, and it doesn't always make sense for the component padding to change between viewports based on whether the table, for example, is laid out as it is in my last screenshot above, or just a child of something in the page content area and doesn't span the width of the main content area.

mcoker (changes requested):
Thanks for the PR @logonoff! I chatted with @lboehling and @andrew-ronaldson about this and we wanted to make a few changes below the md breakpoint:

* The `__main-container` element should fill the viewport height below the md breakpoint (you should be able to set it to `align-self: stretch`)
* Remove all border-radius (radii?) from the `__main-container` element

Outside of that, can you create a CSS vars for these new values so that folks can customize the variable and change the styles in this viewport size? What that might look like here is (with `xs` as the "xs" breakpoint, which is "0")

```scss
// add these at the top of the component styles with the other __main-container vars 
--#{$page}__main-container--xs--MarginInlineStart: 0;
--#{$page}__main-container--xs-MarginInlineEnd: 0;
--#{$page}__main-container--xs--MaxHeight: 100%;
--#{$page}__main-container--xs--BorderRadius: 0;

// Update the styles in the breakpoint block you added
@media ... {
  --#{$page}__main-container--MarginInlineStart: var(--#{$page}__main-container--xs--MarginInlineStart);
  --#{$page}__main-container--MarginInlineEnd: var(--#{$page}__main-container--xs-MarginInlineEnd);
  --#{$page}__main-container--MaxHeight: var(--#{$page}__main-container--xs--MaxHeight);
  --#{$page}__main-container--BorderRadius: var(--#{$page}__main-container--xs--BorderRadius);
}
```

mcoker (commented):
Love this change! Thank you @logonoff. Here is a small visual regression report of page, table, and navigation component examples/demos. Note, not _everything_ in the report is from this change, so ignore the stuff that doesn't look related to this change.

[Visual regression report (PDF)](https://github.com/user-attachments/files/19168680/backstop-3.10.25.pdf)


@lboehling @andrew-ronaldson question around the small white border we add to the main content - should we also remove that border on the xs/sm breakpoints? I'm assuming so, but want to verify.

Here are a couple of screenshots that show the border. Note that the page section background and the content (when in a no-padding page section) do not touch the edge of the viewport.

Screenshot 2025-03-10 at 1 32 01 PM
Screenshot 2025-03-10 at 1 31 51 PM

mcoker (approved):
Confirmed with design that we'll leave the border 🚀

PR #7305, opened 2025-01-14
fix(Alert): create alert group variables for toast animation

### LINKED ISSUE #6812: Alert - add parameters for animation
Toast alert animation was added but right now the parameters for animation are hard coded as tokens - this issue should create alert-group component variables to allow customization.

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/6812

### DIFF
--- a/src/patternfly/components/Alert/alert-group.scss
+++ b/src/patternfly/components/Alert/alert-group.scss
@@ -11,6 +11,60 @@
   --#{$alert-group}--m-toast--MaxWidth: #{pf-size-prem(600px)};
   --#{$alert-group}--m-toast--ZIndex: var(--pf-t--global--z-index--2xl);
 
+  // Alert group item addition reduced motion
+  --#{$alert-group}--m-toast__item--TransitionDuration--opacity--default: var(--pf-t--global--motion--duration--fade--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--opacity--default: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--Transition--opacity--default: var(--#{$alert-group}--m-toast__item--TransitionDuration--opacity--default) 
+    var(--#{$alert-group}--m-toast__item--TransitionTimingFunction--opacity--default) 
+    0s;
+
+  // Alert group item addition
+  --#{$alert-group}--m-toast__item--TransitionDuration--opacity: var(--pf-t--global--motion--duration--fade--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--opacity: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--TransitionDelay--opacity: var(--pf-t--global--motion--delay--none);
+  --#{$alert-group}--m-toast__item--TransitionDuration--transform: var(--pf-t--global--motion--duration--slide-in--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--transform: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--TransitionDelay--transform: var(--pf-t--global--motion--delay--none);
+  --#{$alert-group}--m-toast__item--TransitionDuration--grid-template-rows: var(--pf-t--global--motion--duration--slide-in--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--grid-template-rows: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--TransitionDuration--margin-block: var(--pf-t--global--motion--duration--slide-in--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--margin-block: var(--pf-t--global--motion--timing-function--decelerate);
+
+  // Alert addition
+  --#{$alert-group}--m-toast__item--c-alert--TransitionDuration: var(--pf-t--global--motion--duration--slide-in--default);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Alert/alert-group.scss
```diff
@@ -71,22 +126,32 @@
   // Reduced motion by default
   // transparency change only
   transition: 
-  opacity var(--pf-t--global--motion--duration--fade--default) var(--pf-t--global--motion--timing-function--decelerate) 0s;
-  transform: translateX(0) translateY(0);
+    opacity var(--#{$alert-group}--m-toast__item--Transition--opacity--default);
```
srambach, line 128: I think here `var(--#{$alert-group}--m-toast__item--Transition--opacity--default)` already includes "opacity" so you could take it out here - that seems to be preventing the opacity transition as well. Or maybe it would be better to take opacity out of the variable - wdyt?

src/patternfly/components/Alert/alert-group.scss
```diff
@@ -11,6 +11,61 @@
   --#{$alert-group}--m-toast--MaxWidth: #{pf-size-prem(600px)};
   --#{$alert-group}--m-toast--ZIndex: var(--pf-t--global--z-index--2xl);
 
+  // Alert group item addition reduced motion
+  --#{$alert-group}--m-toast__item--TransitionDuration--opacity--default: var(--pf-t--global--motion--duration--fade--opacity--default);
```
srambach: I think this token should be `--pf-t--global--motion--duration--fade--default`, so then this variable isn't defined and that's ultimately causing the opacity change not to happen in reduced motion.

srambach (commented):
I think I found a couple of little things keeping the default fade from working in reduced-motion. 

I wonder if we could do away with some variables if we move the media query into the variables, but I'm not sure if it comes out any simpler or not, or if we'd want those variables for customization (my inclination is that customizing them is not necessary). Maybe it's worth chatting about since we are kind of breaking new ground using the reduced-motion preference. @mcoker

src/patternfly/components/Alert/alert-group.scss
sg00dwin: Yes, it was part of the problem of why the opacity transition wasn't happening at reduced motion. I agree it's best to remove it from the variable so that it works properly.

sg00dwin:
@srambach I've updated these changes, so that at reduced-motion, `pf-m-offstage-right` now correctly shows a transition to `opacity: 0` then it will collapse with no slide. To prevent the alert from moving while it's fading out, I used timing delays on the `grid-template-rows` `margin-block` `padding` and `border-width` properties.

[Here's a recording of it in real time and slowed down](https://drive.google.com/file/d/1zguaocrSKl4uBWrAF8_WFb-DWWHlf8NI/view?usp=drive_link).

src/patternfly/components/Alert/alert-group.scss
```diff
@@ -11,6 +11,60 @@
...
+  --#{$alert-group}--m-toast__item--TransitionDuration--opacity--default: var(--pf-t--global--motion--duration--fade--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--opacity--default: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--Transition--opacity--default: var(--#{$alert-group}--m-toast__item--TransitionDuration--opacity--default) 
+    var(--#{$alert-group}--m-toast__item--TransitionTimingFunction--opacity--default) 
+    0s;
+
+  // Alert group item addition
+  --#{$alert-group}--m-toast__item--TransitionDuration--opacity: var(--pf-t--global--motion--duration--fade--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--opacity: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--TransitionDelay--opacity: var(--pf-t--global--motion--duration--slide-in--default);
```
srambach: Based on #7341 can you set this delay to 0?

src/patternfly/components/Alert/alert-group.scss
```diff
@@ -11,6 +11,60 @@
...
+  --#{$alert-group}--m-toast__item--TransitionDuration--transform: var(--pf-t--global--motion--duration--slide-in--default);
+  --#{$alert-group}--m-toast__item--TransitionTimingFunction--transform: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$alert-group}--m-toast__item--TransitionDelay--transform: var(--pf-t--global--motion--duration--slide-in--default);
```
srambach: Based on #7341 can you set this delay to 0?

srambach (commented):
I was going to start #7341 to fix an issue we noticed in https://github.com/patternfly/patternfly-react/pull/11495 but I realized it would be better to address here. Left 2 teensy changes that should fix the animation as intended. (i.e. on the transition in from the top, the item will not wait until the "hole" opens before sliding into the slot; it all happens at the same time)

sg00dwin:
Removed transition delays which addresses https://github.com/patternfly/patternfly/issues/7341

srambach (commented):
Hey, I noticed one more bug that would be good to fix here too if you don't mind - I think the `pf-v6-c-alert-group__item` needs to have a `min-height: 0;` Without it, you can see that there's still a bit of space when you take a toast alert offstage right.

https://github.com/user-attachments/assets/15014b61-f7e5-43f4-89e7-7452a43ca10b

sg00dwin:
@srambach Added the `min-height: 0` to prevent the alert from adding space once it's removed.

https://github.com/user-attachments/assets/f601b6f8-c826-4ef6-a6ce-c6e679b04604

srambach (approved):
Looks great! Thank you!

mcoker (approved):
Sweet! Good job on these.

PR #7319, opened 2025-01-29
fix(toolbar): remove double padding when in main section

### LINKED ISSUE #7332: Extra spacing below toolbar
Image

As seen at https://quickstarts.netlify.app/?quickstart=getting-started-with-quick-starts

cc: @nicolethoen, @kmcfaul

logonoff:
closes https://github.com/patternfly/patternfly/issues/7332

before:
before

after:
after


sg00dwin:
After discussing with @mcoker, the Toolbar component intentionally has a bottom padding set which we don't want to remove.

We are looking into option to add a modifier to the Toolbar component, `hasNoPadding` which would set an additional class to remove that bottom padding.

logonoff:
obsoleted by https://github.com/patternfly/patternfly-react/pull/11648

PR #7322, opened 2025-02-05
fix(menu): enable pointer-events on aria-disabled items

### LINKED ISSUE #7253: Bug - MenuItem - pointer-events on aria-disabled prevents tooltips
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Related to React issue https://github.com/patternfly/patternfly-react/issues/11327. Essentially the `pointer-events: none` style prevents the React Tooltip component from working as expected for mouse -- hovering an aria-disabled MenuItem that has a tooltip does not trigger the tooltip, but focusing it via keyboard does trigger it.

Per @mcoker , Tabs and Buttons support the pf-m-aria-disabled class without applying the pointer-events style on them; MenuItems seems to be an outlier.

Go to a React Menu example (basic menu example) and hover the aria-disabled menu item. Notice a tooltip does not trigger.

Then try focusing the aria-disabled item via keyboard - notice the tooltip DOES trigger.

**Expected behavior**
The tooltip triggers via mouse or keyboard.

Workaround would be adding a custom style to override the pointer events.

**Screenshots**

**What is your environment?**

**Any other information?**

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7253

A couple of notes

* This keeps the background from showing up on focus for an aria-disabled item. This is different than v5, but the hover background with disabled text was more readable in v5 than it is in v6. This also now matches lack of background change on aria-disabled buttons.
* Updates the external icon var in a disabled item to 1) fix the var name since it was incorrect, and 2) change the color to transparent so the icon doesn't show on hover/focus for an aria-disabled item. The color change to transparent also applies to the select/check icon.

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -175,11 +175,21 @@
   &:is(.pf-m-disabled, :disabled, .pf-m-aria-disabled, [aria-disabled="true"]) {
     --#{$menu}__item--Color: var(--#{$menu}__item--m-disabled--Color);
     --#{$menu}__item-toggle-icon--Color: var(--#{$menu}--icon--disabled--Color);
-    --#{$menu}__item-external-icon--Color: var(--#{$menu}--icon--disabled--Color);
+    --#{$menu}__item-external--Color: transparent;
+    --#{$menu}__item-select-icon--Color: transparent;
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):

srambach (approved):
LGTM

PR #7336, opened 2025-02-12
feat(tabs): animate background, current border

### LINKED ISSUE #7250: Animation: Tabs - Add transitions on selected and hovered tabs
Follow up to https://github.com/patternfly/patternfly-design/issues/1331

For this issue, we'll just add the transition to default (non-boxed) horizontal tabs. We can have follow ups for vertical tabs and other variants later.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7250

### DIFF
--- a/src/patternfly/components/Tabs/tabs-link.hbs
+++ b/src/patternfly/components/Tabs/tabs-link.hbs
@@ -17,7 +17,7 @@
       disabled
     {{/if}}
   {{/if}}
-  {{!--
+  
   onClick="
     event.preventDefault();
     ((e) => {
@@ -42,7 +42,7 @@
         {{/if}}
       }
     )()"
-  --}}
+ 
   {{~#if tabs-link--IsAriaDisabled}}
     aria-disabled="true"
   {{/if}}
--- a/src/patternfly/components/Tabs/tabs-link.hbs
+++ b/src/patternfly/components/Tabs/tabs-link.hbs
@@ -17,6 +17,7 @@
       disabled
     {{/if}}
   {{/if}}
+  {{!--
   onClick="
     event.preventDefault();
     ((e) => {
@@ -41,6 +42,7 @@
         {{/if}}
       }
     )()"
+  --}}
   {{~#if tabs-link--IsAriaDisabled}}
     aria-disabled="true"
   {{/if}}
--- a/src/patternfly/components/Tabs/tabs-link.hbs
+++ b/src/patternfly/components/Tabs/tabs-link.hbs
@@ -17,6 +17,22 @@
       disabled
     {{/if}}
   {{/if}}
+  onClick="
+    ((e) => {
+      const el = this,
+            li = this.closest('li'),
+            ul = this.closest('ul'),
+            tabs = this.closest('.pf-v6-c-tabs'),
+            left = Math.round(li.offsetLeft),
+            width = Math.round(li.offsetWidth);
+
+      ul.querySelectorAll('li').forEach(function(el) {
+        el.classList.remove('pf-m-current')});
+        li.classList.add('pf-m-current');
+        tabs.style.setProperty('--pf-v6-c-tabs--link-accent--start', left + 'px');
+        tabs.style.setProperty('--pf-v6-c-tabs--link-accent--width', width + 'px');
+      }
+    )()"
   {{~#if tabs-link--IsAriaDisabled}}
     aria-disabled="true"
   {{/if}}
--- a/src/patternfly/components/Tabs/tabs-link.hbs
+++ b/src/patternfly/components/Tabs/tabs-link.hbs
@@ -25,13 +25,20 @@
             ul = this.closest('ul'),
             tabs = this.closest('.pf-v6-c-tabs'),
             left = Math.round(li.offsetLeft),
-            width = Math.round(li.offsetWidth);
+            top = Math.round(li.offsetTop),
+            width = Math.round(li.offsetWidth),
+            height = Math.round(li.offsetHeight);
 
       ul.querySelectorAll('li').forEach(function(el) {
         el.classList.remove('pf-m-current')});
         li.classList.add('pf-m-current');
-        tabs.style.setProperty('--pf-v6-c-tabs--link-accent--start', left + 'px');
-        tabs.style.setProperty('--pf-v6-c-tabs--link-accent--width', width + 'px');
+        {{#if tabs--IsVertical}}
+          tabs.style.setProperty('--pf-v6-c-tabs--link-accent--start', top + 'px');
+          tabs.style.setProperty('--pf-v6-c-tabs--link-accent--length', height + 'px');
+        {{else}}
+          tabs.style.setProperty('--pf-v6-c-tabs--link-accent--start', left + 'px');
+          tabs.style.setProperty('--pf-v6-c-tabs--link-accent--length', width + 'px');
+        {{/if}}
       }
     )()"
   {{~#if tabs-link--IsAriaDisabled}}
--- a/src/patternfly/components/Tabs/tabs-link.hbs
+++ b/src/patternfly/components/Tabs/tabs-link.hbs
@@ -18,6 +18,7 @@
     {{/if}}
   {{/if}}
   onClick="
+    event.preventDefault();
     ((e) => {
       const el = this,
             li = this.closest('li'),
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -3,6 +3,20 @@
 $pf-v6-c-tabs--breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg", "xl", "2xl");
 $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl");
 
+@property --#{$tabs}--link-accent--length {
+  syntax: "<length>";
+  inherits: true;
+  // stylelint-disable-next-line length-zero-no-unit
+  initial-value: 0px;
+}
+
+@property --#{$tabs}--link-accent--start {
+  syntax: "<length>";
+  inherits: true;
+  // stylelint-disable-next-line length-zero-no-unit
+  initial-value: 0px;
+}
+
 @include pf-root($tabs) {
   --#{$tabs}--inset: 0;
   --#{$tabs}--Width: auto;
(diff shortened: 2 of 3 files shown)


andrew-ronaldson:
This looks good to me. Is there ongoing work or good for review?

thatblindgeye (approved):
From a React perspective this sounds feasible. Looking into the existing code for Tabs stuff, I'm thinking we'd just leverage existing things for the most part:

- we're already using `componenDidUpdate` to check if the children length changes (covers adding/removing tabs to recalculate the value of these vars)
- We could use that ^ lifecycle to also check if the active tab changes and likewise recalculate the value of the vars
- We may need to add some logic within the Tab component to check if its content changes and recalculate the vars as well

Preventing the animation on initial render may be tricky depending if the animation would occur immediately, but as discussed if some modifier class were to be added to determine when to animate then we should be able to calculate the initial var values on the first render, then add whatever class to "enable" animations.

kaylachumley (approved):
i like it!

andrew-ronaldson:
Do we need a separate issue for vertical tabs?

andrew-ronaldson:
I can see a little blue nubbin' on the tabs now and the blue underline seems to be thicker with this last update.

https://github.com/user-attachments/assets/d40774f3-4078-42a1-b691-987c54414612

kaylachumley:
I am also seeing the same nubbin' as well

mcoker:
@andrew-ronaldson @kaylachumley sorry about that, it's not quite ready for review 😅 Aside from the nubbin' action, does the vertical tabs animation look as intended?

andrew-ronaldson:
> Aside from the nubbin' action, does the vertical tabs animation look as intended?

Absofruitly!
Except for the line thickness. Otherwise legendary

srambach:
It looks like the old border is still showing here before the sliding one comes in on the examples from Tab Item Actions down.

https://github.com/user-attachments/assets/d742ebc1-34d2-4a12-ac7b-db5a0721dc50

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -733,3 +781,32 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
...
+      border: 0 solid var(--#{$tabs}--link-accent--color);
+      border-block-end-width: var(--#{$tabs}--link-accent--BorderBlockEndWidth);
+      border-inline-start-width: var(--#{$tabs}--link-accent--BorderInlineStartWidth);
+      transition:
+        --#{$tabs}--link-accent--length var(--#{$tabs}--link-accent--TransitionDuration--length) var(--#{$tabs}--link-accent--TransitionTimingFunction--length),
+        --#{$tabs}--link-accent--start var(--#{$tabs}--link-accent--TransitionDuration--start) var(--#{$tabs}--link-accent--TransitionTimingFunction--start),
+    }
+  }
+
+  .#{$tabs}.pf-m-no-animate-current {
```
srambach: I'm not sure about this name 😄 This turns off the slide on a case by case basis? Do we need this?

mcoker: For the time being, this is the class we'll add to tabs that keeps the animation from happening on initial render. Just for documentation purposes, here were the options from our slack thread about it

> There are probably a few ways we can do that:
> * React can keep the tabs from rendering until they set the vars for the current item, then render the tabs component. Eric said that might be tricky though.
> * Core can add a class like .pf-m-accent-no-animate that disabled the animation, React can add that, set the vars, then remove the class. That's easy.
> * I looked into another way of doing it. I think I can expose 2 sets of vars the react component could set - one set of vars animates and the other doesn't. React could set the vars that don't animate first, then on subsequent clicks, they set the vars that do animate

Feel free to suggest a better name! `.pf-m-no-animate` is too broad IMO since we may need a class like this for other animations, so I threw on "current", but I'm thinking `.pf-m-no-animate-accent` would be more appropriate. wdyt?

mcoker:
> It looks like the old border is still showing here before the sliding one comes in on the examples from Tab Item Actions down.

@srambach d'oh, thanks! I forgot @sg00dwin mentioned that bug in one of our meetings. Fixed 👍

src/patternfly/components/Tabs/tabs.scss
srambach: I'm curious what scenario you can think of where we'd want to stop one animation but not others. My instinct is to use a more generic name like `.pf-m-no-animation` or `.pf-m-pause-animation` or even the `.pf-m-no-motion` that we already use.

mcoker: Most of our animations wouldn't need to be stopped like this because they either load with the initial component styles or after some sort of interaction or event. React needs to trigger the "current" animation logic manually (set the 2 vars for accent start and width) when the component mounts just to get the initial border style to render. None of our other animations require anything like that AFAIK.

That said, with a "no-motion" class (as in the user is opting out of the animation), if the goal is to not use the animation at all, I think that will need to do something different because if you simply disable the animation (duration set to 0 or whatever), you won't have an accent at all since the accent is created in a whole separate way depending on whether it's animated or not. That type of class will need to trigger the accent to be created on individual links instead of the list.

srambach: Tried to think of what's bothering me about this one...
- the grammar 😛 
- tokens use "motion" rather than "animate" to be more generic
- it's affecting accent variables so I think I'd prefer it to be named for that rather than current
- it's purpose is to freeze during load/initialization so I'm wondering if it could be named somehow for that state rather than no-animate

WDYT about something like `pf-m-initializing-accent` ?

mcoker: Love it! 🚀

andrew-ronaldson:
The blue line has disappeared! Thanos snapped away! Ahhhhh

mcoker:
@andrew-ronaldson yeah 2 things:
* I commented out the JS snippets I added to trigger the blue border on click. I don't think we want to include that in the core workspace because we don't want to convey that the borders need that snippet to work. 
* Because the borders now rely on react in order to show, they don't show in core. I'm not particularly happy about that, but I'm also not sure how to best address that.

However, we won't be able to merge this PR in core before code freeze, because without the react work being done, the border would disappear for everyone. So I think we have some time to sort out how we should make the borders appear in core, or if we leave them off (similar to how we don't show expanded menus in core since that's react functionality)

src/patternfly/components/Tabs/tabs-link.hbs
```diff
@@ -41,6 +42,7 @@
         {{/if}}
       }
     )()"
+  --}}
```
srambach: did you want to remove this yet?

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -733,3 +781,35 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
...
+      inset-inline-start: var(--#{$tabs}--link-accent--InsetInlineStart, var(--#{$tabs}--link-accent--start));
+      width: var(--#{$tabs}--link-accent--Width, var(--#{$tabs}--link-accent--length));
+      height: var(--#{$tabs}--link-accent--Height, var(--#{$tabs}--link-accent--length));
+      content: "";
+      border: 0 solid var(--#{$tabs}--link-accent--color);
+      border-block-end-width: var(--#{$tabs}--link-accent--BorderBlockEndWidth);
+      border-inline-start-width: var(--#{$tabs}--link-accent--BorderInlineStartWidth);
+      transition:
+        --#{$tabs}--link-accent--length var(--#{$tabs}--link-accent--TransitionDuration--length) var(--#{$tabs}--link-accent--TransitionTimingFunction--length),
+        --#{$tabs}--link-accent--start var(--#{$tabs}--link-accent--TransitionDuration--start) var(--#{$tabs}--link-accent--TransitionTimingFunction--start),
```
srambach:
```suggestion
        --#{$tabs}--link-accent--start var(--#{$tabs}--link-accent--TransitionDuration--start) var(--#{$tabs}--link-accent--TransitionTimingFunction--start)
```

andrew-ronaldson:
Could we have a selected tab in the examples to show the states or is it entirely in react?

src/patternfly/components/Tabs/tabs.scss
srambach: I'm wondering about breaking this up rather than using the shorthand. It is a little harder to make sure the timings are lined up - but also, do we need to have separate timing function and duration variables, or could we have one for the entire effect - that we'd always want them to be the same since it's one motion.

srambach (commented):
But I liked "tubtab" 😉 

Also, wasn't sure if you were ready to remove the commented out JS or not.

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -576,6 +623,7 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   text-decoration-line: none;
   background-color: var(--#{$tabs}__link--BackgroundColor);
   border-radius: var(--#{$tabs}__link--BorderRadius);
+  transition: background-color var(--#{$tabs}__link--TransitionDuration--background-color) var(--#{$tabs}__link--TransitionTimingFunction--background-color);
```
srambach, line 628: Do you want to separate this rather than using the shorthand?

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -733,3 +779,35 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
...
+      border-inline-start-width: var(--#{$tabs}--link-accent--BorderInlineStartWidth);
+      transition-timing-function: var(--#{$tabs}--link-accent--TransitionTimingFunction);
+      transition-duration: var(--#{$tabs}--link-accent--TransitionDuration);
+      transition-property: --#{$tabs}--link-accent--length, --#{$tabs}--link-accent--start, width;
+    }
+  }
+
+  .#{$tabs}.pf-m-initializing-accent {
+    --#{$tabs}--link-accent--TransitionDuration--length: 0;
+    --#{$tabs}--link-accent--TransitionDuration--start: 0;
```
srambach: Should these two lines now just be `--#{$tabs}--link-accent--TransitionDuration: 0;`?

src/patternfly/components/Tabs/tabs.scss
mcoker: Doh! Yep - fixed.


https://github.com/user-attachments/assets/a81d860e-2676-439c-89fe-566333edbdb1

srambach (approved):
L 📁 TM

kaylachumley (commented):
oops, missed this PR! I went to review but the animated tabs isnt working for me... maybe user error. I tried a new browser and nada! is the animated tabs working for you? @mcoker

mcoker:
@kaylachumley since the border depends on javascript I have to manually enable it in this PR, but will want to take it out prior to merging this PR since we don't include javascript like this in the core examples. I've re-enabled it so you can see it on all of the examples now - https://patternfly-pr-7336.surge.sh/components/tabs

kaylachumley:
looks awesome

andrew-ronaldson (approved):
🤩

PR #7345, opened 2025-02-13
feat(menu-toggle): menu toggle with danger status animation

### LINKED ISSUE #6591: Animation: Menu toggle - Validation
**Goal**
Follow up to https://github.com/patternfly/patternfly/issues/6590

Adding a small jiggle on the menu toggle if it does not pass validation. If this animation is possible, we can create more issues for the other form elements that use validation.

Link to [Figma annotations](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/branch/J0HlG83xKgYoT2pMpCl4SY/PatternFly-6%3A-Components?type=design&node-id=22732%3A26347&mode=design&t=Mmwgp55YUkvDrHUz-1)

Link to [prototype](https://www.figma.com/proto/VMEX8Xg2nzhBX8rfBx53jp/branch/J0HlG83xKgYoT2pMpCl4SY/PatternFly-6%3A-Components?type=design&node-id=23118-27297&t=NUaOc6etggp2gIAn-0&scaling=min-zoom&page-id=3%3A16)

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/6591

[Codesandbox](https://codesandbox.io/p/sandbox/awesome-cdn-flsfyv) showing behavior.

Note this [design prototype](https://www.figma.com/proto/VMEX8Xg2nzhBX8rfBx53jp/branch/J0HlG83xKgYoT2pMpCl4SY/PatternFly-6%3A-Components?type=design&node-id=23118-27297&t=NUaOc6etggp2gIAn-0&scaling=min-zoom&page-id=3%3A16) includes the `HelperText` component which fades in with the menu-toggle danger selection. That helper-text fade in transition code changes aren't currently included in these changes. I didn't know if it should be or if it would be a follow on. Because in my sandbox demo I only applied an css opacity transition that keys off of `pf-v6-c-helper-text__item.pf-m-error` which would imply all helper-text with the error class would transition, even if it's not associated with a `menu-toggle`

Update: 
These changes now include additions to fade in the helper text for `pf-m-error` status

### DIFF
--- a/src/patternfly/components/HelperText/helper-text.scss
+++ b/src/patternfly/components/HelperText/helper-text.scss
@@ -16,6 +16,8 @@
   --#{$helper-text}__item-text--m-success--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$helper-text}__item-icon--m-error--Color: var(--pf-t--global--icon--color--status--danger--default);
   --#{$helper-text}__item-text--m-error--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$helper-text}__item-icon--m-error--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--fade--default);
+  --#{$helper-text}__item-icon--m-error--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
 
   // dynamic
   --#{$helper-text}--m-dynamic__item-icon--Color: var(--pf-t--global--icon--color--regular);
@@ -71,6 +73,15 @@
     --#{$helper-text}__item-text--FontWeight: var(--#{$helper-text}__item-text--m-error--FontWeight);
     --#{$helper-text}__item-icon--Color: var(--#{$helper-text}__item-icon--m-error--Color);
     --#{$helper-text}--m-dynamic__item-icon--Color: var(--#{$helper-text}--m-dynamic--m-error__item-icon--Color);
+
+    transition:
+      opacity 
+        var(--#{$helper-text}__item-icon--m-error--TransitionDuration--Opacity)
+        var(--#{$helper-text}__item-icon--m-error--TransitionTimingFunction--Opacity);
+    
+    @starting-style {
+      opacity: 0;
+    }
   }
 
   &.pf-m-dynamic {
--- a/src/patternfly/components/HelperText/helper-text.scss
+++ b/src/patternfly/components/HelperText/helper-text.scss
@@ -16,8 +16,8 @@
   --#{$helper-text}__item-text--m-success--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$helper-text}__item-icon--m-error--Color: var(--pf-t--global--icon--color--status--danger--default);
   --#{$helper-text}__item-text--m-error--FontWeight: var(--pf-t--global--font--weight--body--bold);
-  --#{$helper-text}__item-icon--m-error--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--fade--default);
-  --#{$helper-text}__item-icon--m-error--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
+  --#{$helper-text}__item--m-error--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--fade--default);
+  --#{$helper-text}__item--m-error--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
 
   // dynamic
   --#{$helper-text}--m-dynamic__item-icon--Color: var(--pf-t--global--icon--color--regular);
@@ -74,13 +74,18 @@
     --#{$helper-text}__item-icon--Color: var(--#{$helper-text}__item-icon--m-error--Color);
     --#{$helper-text}--m-dynamic__item-icon--Color: var(--#{$helper-text}--m-dynamic--m-error__item-icon--Color);
 
-    transition:
-      opacity 
-        var(--#{$helper-text}__item-icon--m-error--TransitionDuration--Opacity)
-        var(--#{$helper-text}__item-icon--m-error--TransitionTimingFunction--Opacity);
+    animation-name: --#{$helper-text}-item-fade-in;
+    animation-duration: var(--#{$helper-text}__item--m-error--TransitionDuration--Opacity);
+    animation-timing-function: var(--#{$helper-text}__item--m-error--TransitionTimingFunction--Opacity); 
     
-    @starting-style {
-      opacity: 0;
+    @keyframes --#{$helper-text}-item-fade-in {
+      from {
+        opacity: 0;
+      }
+
+      to {
+        opacity: 1;
+      }
     }
   }
 
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -152,6 +152,8 @@
 
   // Status icon
   --#{$menu-toggle}__status-icon--Color: var(--pf-t--global--icon--color--regular);
+  --#{$menu-toggle}__status-icon--m--danger--Opacity: var(--pf-t--global--motion--duration--icon--default);
+  --#{$menu-toggle}__status-icon--m--danger--TransitionTimingFunction: var(--pf-t--global--motion--duration--icon--default);
 
   // Success
   --#{$menu-toggle}--m-success--BorderColor: var(--pf-t--global--border--color--status--success--default);
@@ -164,6 +166,8 @@
   // Danger
   --#{$menu-toggle}--m-danger--BorderColor: var(--pf-t--global--border--color--status--danger--default);
   --#{$menu-toggle}--m-danger__status-icon--Color: var(--pf-t--global--icon--color--status--danger--default);
+  --#{$menu-toggle}--m-danger--TransitionDuration--Transform: var(--pf-t--global--motion--duration--md);
+  --#{$menu-toggle}--m-danger--TransitionTimingFunction--Transform: var(--pf-t--global--motion--timing-function--default);
 
   // Placeholder
   --#{$menu-toggle}--m-placeholder--Color: var(--pf-t--global--text--color--placeholder);
@@ -337,6 +341,24 @@
   &.pf-m-danger {
     --#{$menu-toggle}--BorderColor: var(--#{$menu-toggle}--m-danger--BorderColor);
     --#{$menu-toggle}__status-icon--Color: var(--#{$menu-toggle}--m-danger__status-icon--Color);
+    
+    transform: translate3d(0, 0, 0);
+    animation-name: menu-toggle-m-danger-motion;
+    animation-duration: var(--#{$menu-toggle}--m-danger--TransitionDuration--Transform);
+    animation-timing-function: var(--#{$menu-toggle}--m-danger--TransitionTimingFunction--Transform);
+    animation-fill-mode: both;
+
+    .#{$menu-toggle}__status-icon {
+      color: var(--#{$menu-toggle}__status-icon--Color, inherit);
+      transition: 
+        opacity
+          var(--#{$menu-toggle}__status-icon--m--danger--Opacity)
+          var(--#{$menu-toggle}--m-danger--TransitionTimingFunction--Transform);
+      
+      @starting-style  {
+        opacity: 0;
+      }
+    }
   }
 
   &.pf-m-placeholder {
@@ -353,6 +375,18 @@
     &,
     .#{$button} {
       pointer-events: none;
+    }
+  }
+}
+
+@media (prefers-reduced-motion: no-preference) {
+  @keyframes menu-toggle-m-danger-motion {
+    33% {
+      transform: translate3d(-2px, 0, 0);
+    }
+
+    66% {
+      transform: translate3d(3px, 0, 0);
     }
   }
 }
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -152,8 +152,8 @@
 
   // Status icon
   --#{$menu-toggle}__status-icon--Color: var(--pf-t--global--icon--color--regular);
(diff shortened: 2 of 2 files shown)


andrew-ronaldson:
This looks good. If it's not too much right now could we do the fade as well so we don't have to go back with another issue?

kaylachumley (approved):
looking great 🪄 
+1 for andrews comment about incorporating the fade now if possible

andrew-ronaldson (approved):
excellent! Thanks!

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -16,6 +16,8 @@
   --#{$helper-text}__item-text--m-success--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$helper-text}__item-icon--m-error--Color: var(--pf-t--global--icon--color--status--danger--default);
   --#{$helper-text}__item-text--m-error--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$helper-text}__item-icon--m-error--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--fade--default);
+  --#{$helper-text}__item-icon--m-error--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
```
sg00dwin:
```suggestion
  --#{$helper-text}__item--m-error--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--fade--default);
  --#{$helper-text}__item--m-error--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
```

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -71,6 +73,15 @@
     --#{$helper-text}__item-text--FontWeight: var(--#{$helper-text}__item-text--m-error--FontWeight);
     --#{$helper-text}__item-icon--Color: var(--#{$helper-text}__item-icon--m-error--Color);
     --#{$helper-text}--m-dynamic__item-icon--Color: var(--#{$helper-text}--m-dynamic--m-error__item-icon--Color);
+
+    transition:
+      opacity 
+        var(--#{$helper-text}__item-icon--m-error--TransitionDuration--Opacity)
+        var(--#{$helper-text}__item-icon--m-error--TransitionTimingFunction--Opacity);
```
sg00dwin, line 80: rename

sg00dwin:
```suggestion
        var(--#{$helper-text}__item--m-error--TransitionDuration--Opacity)
        var(--#{$helper-text}__item--m-error--TransitionTimingFunction--Opacity);
```

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -71,6 +73,15 @@
...
+    
+    @starting-style {
+      opacity: 0;
+    }
```
sg00dwin, line 84: Switch to using `keyframes` because `@starting-style` isn't fully supported by Firefox

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -337,6 +341,28 @@
...
+        color: var(--#{$menu-toggle}__status-icon--Color, inherit);
+        opacity: 1;
+        transition: 
+          opacity 
+            var(--#{$menu-toggle}__status-icon--m-danger--TransitionDuration--Opacity)
+            var(--#{$menu-toggle}__status-icon--m-danger--TransitionDuration--Opacity);
+        // stylelint-disable max-nesting-depth
+        @starting-style {
+          opacity: 0;
+        }
```
sg00dwin, line 362: Switch to using keyframes because @starting-style isn't fully supported by Firefox

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -337,6 +341,33 @@
   &.pf-m-danger {
     --#{$menu-toggle}--BorderColor: var(--#{$menu-toggle}--m-danger--BorderColor);
     --#{$menu-toggle}__status-icon--Color: var(--#{$menu-toggle}--m-danger__status-icon--Color);
+
+    @media screen and (prefers-reduced-motion: no-preference) {
+      transform: translate3d(0, 0, 0);
```
mcoker: Just a note that you can change this to `translateX()`. You could also go ahead and put a var here, then just modify the var in the animation `@keyframes` if you wanted.

```suggestion
      transform: translate3d(0, 0, 0);
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -337,6 +341,33 @@
...
+      animation-name: menu-toggle-m-danger-motion;
+      animation-duration: var(--#{$menu-toggle}--m-danger--AnimationDuration--Transform);
+      animation-timing-function: var(--#{$menu-toggle}--m-danger--AnimationTimingFunction--Transform);
+      animation-fill-mode: both;
+      
+      .#{$menu-toggle}__status-icon {
+        color: var(--#{$menu-toggle}__status-icon--Color, inherit);
+        animation-name: menu-toggle-status-icon-fade-in;
```
mcoker: to avoid conflicts with different versions of patternfly loaded together, let's use the component var with the version in it

```suggestion
        animation-name: #{$menu-toggle}-status-icon-fade-in;
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -337,6 +341,33 @@
...
+        animation-duration: var(--#{$menu-toggle}__status-icon--m-danger--TransitionDuration--Opacity);
+        animation-timing-function: var(--#{$menu-toggle}__status-icon--m-danger--TransitionTimingFunction--Opacity);
+        
+        // stylelint-disable max-nesting-depth
+        @keyframes menu-toggle-status-icon-fade-in {
```
mcoker:
```suggestion
        @keyframes #{$menu-toggle}-status-icon-fade-in {
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -357,6 +388,18 @@
   }
 }
 
+@media (prefers-reduced-motion: no-preference) {
+  @keyframes menu-toggle-m-danger-motion {
```
mcoker:
```suggestion
  @keyframes #{$menu-toggle}-m-danger-motion {
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -337,6 +341,33 @@
   &.pf-m-danger {
     --#{$menu-toggle}--BorderColor: var(--#{$menu-toggle}--m-danger--BorderColor);
     --#{$menu-toggle}__status-icon--Color: var(--#{$menu-toggle}--m-danger__status-icon--Color);
+
+    @media screen and (prefers-reduced-motion: no-preference) {
+      transform: translate3d(0, 0, 0);
+      animation-name: menu-toggle-m-danger-motion;
```
mcoker:
```suggestion
      animation-name: #{$menu-toggle}-m-danger-motion;
```

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -71,6 +73,20 @@
     --#{$helper-text}__item-text--FontWeight: var(--#{$helper-text}__item-text--m-error--FontWeight);
     --#{$helper-text}__item-icon--Color: var(--#{$helper-text}__item-icon--m-error--Color);
     --#{$helper-text}--m-dynamic__item-icon--Color: var(--#{$helper-text}--m-dynamic--m-error__item-icon--Color);
+
+    animation-name: helper-text-item-fade-in;
+    animation-duration: var(--#{$helper-text}__item--m-error--TransitionDuration--Opacity);
+    animation-timing-function: var(--#{$helper-text}__item--m-error--TransitionTimingFunction--Opacity); 
+    
+    @keyframes helper-text-item-fade-in {
```
mcoker:
```suggestion
    @keyframes #{$helper-text}-item-fade-in {
```

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -71,6 +73,20 @@
     --#{$helper-text}__item-text--FontWeight: var(--#{$helper-text}__item-text--m-error--FontWeight);
     --#{$helper-text}__item-icon--Color: var(--#{$helper-text}__item-icon--m-error--Color);
     --#{$helper-text}--m-dynamic__item-icon--Color: var(--#{$helper-text}--m-dynamic--m-error__item-icon--Color);
+
+    animation-name: helper-text-item-fade-in;
```
mcoker:
```suggestion
    animation-name: #{$helper-text}-item-fade-in;
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -357,6 +388,18 @@
   }
 }
 
+@media (prefers-reduced-motion: no-preference) {
+  @keyframes menu-toggle-m-danger-motion {
+    33% {
+      transform: translate3d(-2px, 0, 0);
```
mcoker:
```suggestion
      transform: translateX(-2px);
```

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -357,6 +388,18 @@
...
+    }
+
+    66% {
+      transform: translate3d(3px, 0, 0);
```
mcoker:
```suggestion
      transform: translateX(3px);
```

mcoker (changes requested):
looks good! Just a few things we discussed on slack

src/patternfly/components/MenuToggle/menu-toggle.scss
srambach: On these transforms - won't `translate3d` be hardware accelerated and `translateX` won't? Or is that old/fake news?

src/patternfly/components/HelperText/helper-text.scss
```diff
@@ -16,6 +16,8 @@
   --#{$helper-text}__item-text--m-success--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$helper-text}__item-icon--m-error--Color: var(--pf-t--global--icon--color--status--danger--default);
   --#{$helper-text}__item-text--m-error--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$helper-text}__item--m-error--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--fade--default);
+  --#{$helper-text}__item--m-error--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
```
srambach, line 20: I think we've been doing lower case for the suffix `--opacity` - I'm fine either way though.

src/patternfly/components/MenuToggle/menu-toggle.scss
mcoker: Since the animation is on a `transform` function, it should already be performant. What I remember about the `translate3d()` (IIRC it was adding `translateZ(0)`, which is benign) deal was adding a transform function to an element would put the element on the GPU layer, so then you could animate other things (other than `translate`) and possibly be more performant? Not sure if it was fake news or not, smells like it, like too good to be true or _"Doctors hate this one CSS trick!"_ 😂

srambach: Yeah I found a number of sites/articles/blogs saying to use `translate3d` or `translateZ` because it would force it onto the gpu, but none of the usual suspects.

src/patternfly/components/HelperText/helper-text.scss
sg00dwin: I went with uppercase since, in this file, all the other variables with names used `--PropertyName`. But I noticed it was different in other files.

srambach: I don't think our guidelines cover that case 😄

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -357,6 +388,25 @@
   }
 }
 
+// Register the property type for the custom property to be animatable
+@property --#{$menu-toggle}--m-danger--TranslateX {
+  syntax: "<length>";
+  inherits: false;
+  initial-value: 0;
+}
```
srambach, line 395: 🔥 Neat!
I wonder if we are not putting this up with the other variables, i.e. "private" then if we want to go ahead and start using a `--_` notation?

mcoker: @srambach this is because we're animating the `--#{$menu-toggle}--m-danger--TranslateX` (custom) property, so it needs a syntax/type that is animatable. I like it, but I don't think it matters if we do it this way or if we just animate `tranform: translateX()`. Do you have an opinion?

Some pros/cons/thoughts off the top of my head
* As we discussed in standup, this lets you just animate the property instead of having to write out a whole `transform` function, which could be easier to maintain down the road given that `transform` is shorthand for all of its functions. So that's a plus.
  * Another way to not have to use `@property` and not have to redefine every transform function any time we change something in one of the transform functions would be to just use [`translate: var(...)`](https://developer.mozilla.org/en-US/docs/Web/CSS/translate). The individual transform functions have been supported as regular properties for a while, we just don't use most (or all?) of ours that way.
* I'm not sure if animating a custom property this way gets the same performance boost as animating a transform function, even though the custom property is used in a transform function.
* This scopes the var globally, er to `:root` I presume - not scoped to the component class. No real downside to that I reckon, but it's different.

I could go either way with putting the `_` at the beginning of the var. I'm all for it since I brought it up after seeing it used with the RHDS folks, but I think we should probably make sure it's something we want to do before we release like that since, even though our intent is for it to be an internal var users don't use, it won't be until we feel we've done our due diligence to communicate that to users. Just meaning if we use the `_` and decide that is a bad idea for whatever reason, we won't be able to just change it unless we've spammed everyone that those types of vars aren't protected from breaking changes. So I would probably hold off for now, but I care 2% and could be easily persuaded in the other direction!

andrew-ronaldson (approved):
Swish. 3 pointer. Touchdown!

mcoker (approved):
LPTM!

PR #7351, opened 2025-02-17
fix(toolbar): reset toolbar bottom padding to 0

### LINKED ISSUE #7332: Extra spacing below toolbar

sg00dwin:
Fixes https://github.com/patternfly/patternfly/issues/7332

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -13,7 +13,7 @@ $pf-v6--include-toolbar-breakpoints: true !default;
 @include pf-root($toolbar) {
   --#{$toolbar}--RowGap: var(--pf-t--global--spacer--md);
   --#{$toolbar}--PaddingBlockStart: 0;
-  --#{$toolbar}--PaddingBlockEnd: var(--pf-t--global--spacer--md);
+  --#{$toolbar}--PaddingBlockEnd: 0;
   --#{$toolbar}--PaddingInlineStart: 0;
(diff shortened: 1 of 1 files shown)


mcoker (commented):
Discussed this with @sg00dwin offline. tl;dr of the convo:
* Introducing this change as is would be a breaking change for folks since it would remove the space below all toolbars currently.
* The current usage of the toolbar relies on the bottom padding to space whatever follows it, and toolbars are almost always followed by something that the toolbar controls. This is how the v5 toolbar worked, too, except the v5 toolbar also had top padding that we removed for v6. Ideally, I think we would get rid of this bottom padding, too, and have a good layout solution for spacing the toolbar and whatever comes after it, but removing the default bottom padding will likely need to be done in our next breaking change.
* We could introduce this change behind a modifier (like `.pf-m-no-padding` in core and `hasNoPadding` in react), which I think would be a good idea. But we'll want to run that by design to make sure they have no objections to allowing users to remove the default padding in favor of the space being created some other way. @andrew-ronaldson @lboehling wdyt?
* The quickstarts layout issue in https://github.com/patternfly/patternfly/issues/7332 (that this PR resolves) is pretty nuanced - @sg00dwin is going to follow up on that.

sg00dwin:
@mcoker I opened a [new pr in the quickstarts repo](https://github.com/patternfly/patternfly-quickstarts/pull/372) to address the layout issue we spoke about.

Issue #6601, opened 2024-05-01
Icon buttons: Create motions for Favorite, Settings and stateful buttons


andrew-ronaldson:
Follow up to [design issue](https://github.com/patternfly/patternfly-design/issues/1298)
Figma prototype can be found [here](https://www.figma.com/file/VMEX8Xg2nzhBX8rfBx53jp/branch/1e5dyLys8Y7wBp2tPSwP1B/PatternFly-6%3A-Components?type=design&node-id=23621-551&mode=design&t=nPAKD02uDTHkG3b9-0)

### This issue includes
- [ ] Favorite button (Star icon)
- [x] Settings button (Cog icon)
- [ ] Stateful buttons (Bell)

PR #7368, opened 2025-02-25
feat(Button): add notify modifier to button for microanimation

srambach:
This is for the notification badge part of #6601 

A modifier `.pf-m-notify` gets added to a `.pf-m-button` in order to cause the icon portion of the button to have the ringing motion. 

https://github.com/patternfly/patternfly-react/issues/10624 will need to add the class, wait for the animation to complete, and then remove it, and also manage if a new notification arrives before the animation finishes from the previous notification.

This also requires React to update the notification badge to put the icon into the icon container rather than the text container, following core's structure.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -279,8 +279,8 @@
   --#{$button}__icon--MarginInlineEnd: 0;
   --#{$button}__icon--m-start--MarginInlineEnd: 0;
   --#{$button}__icon--m-end--MarginInlineStart: 0;
-  --#{$button}__icon--AnimationDuration--notify: var(--pf-t--global--duration--600);
-  --#{$button}__icon--AnimationTimingFunction--notify: var(--pf-t--global--motion--timing-function--default);
+  --#{$button}--m-notify__icon--AnimationDuration--notify: var(--pf-t--global--duration--600);
+  --#{$button}--m-notify__icon--AnimationTimingFunction--notify: var(--pf-t--global--motion--timing-function--default);
 
   // Progress
   --#{$button}__progress--width: calc(var(--pf-t--global--icon--size--lg) + var(--pf-t--global--spacer--sm)); // matches medium spinner diameter plus a spacer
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -279,6 +279,8 @@
   --#{$button}__icon--MarginInlineEnd: 0;
   --#{$button}__icon--m-start--MarginInlineEnd: 0;
   --#{$button}__icon--m-end--MarginInlineStart: 0;
+  --#{$button}__icon--AnimationDuration--notify: var(--pf-t--global--duration--600);
+  --#{$button}__icon--AnimationTimingFunction--notify: var(--pf-t--global--motion--timing-function--default);
```
mcoker: Technically, this is how we would normally write out this var. I'm cool leaving it the way you have it, since we have an identifier as `--notify` on the end. WDYT? 

```suggestion
  --#{$button}--m-notify__icon--AnimationDuration--notify: var(--pf-t--global--duration--600);
  --#{$button}--m-notify__icon--AnimationTimingFunction--notify: var(--pf-t--global--motion--timing-function--default);
```

src/patternfly/components/Button/button.scss
```diff
@@ -705,6 +707,12 @@
       }
     }
   }
+
+  &.pf-m-notify .#{$button}__icon {
+    animation-name: #{$button}-icon-notify;
+    animation-duration: var(--#{$button}__icon--AnimationDuration--notify);
+    animation-timing-function: var(--#{$button}__icon--AnimationTimingFunction--notify);
```
mcoker: If you made the other suggestion, we'd also want to update this

```suggestion
    animation-duration: var(--#{$button}--m-notify__icon--AnimationDuration--notify);
    animation-timing-function: var(--#{$button}--m-notify__icon--AnimationTimingFunction--notify);
```

srambach:
For convenience - here's the animation on the top of a demo page.

https://github.com/user-attachments/assets/a6292c86-5fa9-4983-a0cf-5e7c028b1941

src/patternfly/components/Button/button.scss
```diff
@@ -738,3 +746,17 @@
...
+    transform: rotate(30deg);
+  }
+
+  66% {
+    transform: rotate(-60deg);
+  }
+
+  100% {
+    transform: rotate(0);
+  }
```
mcoker: Just a nit but since it starts at `rotate(0)`, you shouldn't need this last part

```suggestion
```

mcoker (approved):
It's good as-is but left a few nit comments in case you wanted to change a couple of things I noticed.

PR #7375, opened 2025-03-04
fix(menu): enable menu scroll with grouped items

### LINKED ISSUE #7256: Bug - Menu - scrollable menus with groups aren't scrollable
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Noticed on the Select component in React, but when a Menu has groups of menu lists, the pf-m-scrollable class isn't working as expected -- no scrollbar is present. Looks like it's because of the following codeblock; `overflow-y` should be set to auto by line 383 (which should be setting the scrollbar on the menu__content wrapper where all menu groups would be children of), but is being overridden by this code block:

https://github.com/patternfly/patternfly/blob/d3213a69149a204bf614ec02af4d9464589f196b/src/patternfly/components/Menu/menu.scss#L385-L393

Go to the [select with grouped items example](https://www.patternfly.org/components/menus/select#with-grouped-items) and add the `isScrollable` prop to the Select component. Add several more SelectOptions as well, then open the toggle and notice no scrollbar appears.

**Expected behavior**
A menu is scrollable when there are menu groups.

Possibly, it might be fixable with a temporary custom style to re-override the `overflow-y` on the menu__content wrapper.

**Screenshots**

**What is your environment?**

Reported by RHOAI: https://issues.redhat.com/browse/RHOAIENG-16867

**Any other information?**

sg00dwin:
Fixes https://github.com/patternfly/patternfly/issues/7256

**After**
Screenshot 2025-03-04 at 7 49 22 AM

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -399,7 +399,6 @@
       z-index: var(--#{$menu}--ZIndex);
       margin-block-start: calc(var(--#{$menu}--RowGap) * -1);
       margin-block-end: calc(var(--#{$menu}--RowGap) * -1);
-      overflow: hidden; // hide scrollbars
     }
 
     .#{$menu}__list {
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -394,7 +394,6 @@
 
     // offset row-gap when scrollbar is present
     .#{$menu}__content {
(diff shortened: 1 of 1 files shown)


mcoker (approved):
LGTM!

srambach (approved):
👍🏻 
Unless we want to take out that TODO on 397?

PR #7381, opened 2025-03-05
chore(tab-content): move tab content under tabs

### LINKED ISSUE #7328: TabContent - update documentation to be part of Tabs docs
Currently TabContent is rendered as a separate component on Org for Core from the rest of the Tabs component(s). Per discussion from planning topics on 1/28/2025, we should update documentation for now and we can update location later. cc @mcoker 

**Any other information?**

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7328

I left the tabs example titles alone in case someone saved/bookmarked them. Lemme know if you'd prefer update any of them (for example "Default" to "Default tabs", or under tab content "Secondary" to "Secondary tab content").


thatblindgeye (approved):
Personally it's probably fine for now re: titling. The titles are unique on the page so it should e fine. 

Eventually it might be good to rethink the naming convention for examples and the structures, but waiting until we get the new docs framework up would be better since we may have more flexibility with how we can structure things (e.g. maybe something like `## Tab examples  > ### Basic > the example itself` if there's really only 1 example (rather than "Examples" > "Default" > "Default tab example" >  the example itself), and `## Tab examples > ### Overflow > #### Beginning of list > the example itself` etc).

But again, not something to do here either way, just thinking out loud.

andrew-ronaldson (approved):
Good to go with @thatblindgeye points

PR #7396, opened 2025-03-10
feat(card): enable card header with actions to wrap modifier

### LINKED ISSUE #3713: Card header doesn't wrap

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/3713


mcoker (approved):
LPTM!

Venefilyn:
Awesome. Thanks @sg00dwin and @mcoker!

PR #7397, opened 2025-03-11
fix(global): banner, data-list, menu-toggle bugs

### LINKED ISSUE #7297: Review visual regression report
There are a number of changes in this report. Most look OK but we should review them and see if any need action:

[backstop-1.13.25.pdf](https://github.com/user-attachments/files/18402928/backstop-1.13.25.pdf)

[Full backstop report](https://drive.google.com/file/d/1d6d6keeZ1_JSPQxI7Sbl8GOpzdPfdG7k/view?usp=sharing)

* Menu toggle - [plain button border-radius](https://github.com/patternfly/patternfly/blob/ed0b31fb7eb774c3a94fb0178ae52b1c596bbe5c/src/patternfly/components/MenuToggle/menu-toggle.scss#L134) should be set to border-radius-small token 
* Banner inline link button color is blue, should be same as banner text color
* Ask design if horizontal nav masthead logo should occupy sidebar width
* Review sidebar/content margin - looks like it changed in this demo http://core-staging.patternfly.org/patterns/card-view/html-demos/card-view/
* Draggable data-list should change back to compact

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7297

There are 2 outstanding issues. Looks like both are the result of this PR (cc @srambach) - https://github.com/patternfly/patternfly/pull/7224

1. The "logo" area when using horizontal navigation (no sidebar/vertical navigation) is the same width as a page with vertical navigation. @andrew-ronaldson @lboehling can you confirm if that should be the case? Or when using horizontal navigation, should the logo just take up however much space is needed for the logo? Here's what it looks like now with horizontal nav in the front and vertical nav in the back

Screenshot 2025-03-10 at 7 48 25 PM

FWIW here is what it looks like if the logo area is only as big as the logo

Screenshot 2025-03-10 at 7 52 09 PM

2. We lost the right margin on the page sidebar/vertical navigation. You can see here that there is much less space to the right of the sidebar compared to the left, and when there is a scrollbar, it touches the main content container. @andrew-ronaldson @lboehling can you confirm which is preferred, or if there are any other changes you'd like to see?

Screenshot 2025-03-10 at 7 53 33 PM
Screenshot 2025-03-10 at 7 53 25 PM

 
If we revert the change from https://github.com/patternfly/patternfly/pull/7224, you can see the space on the left side of the nav matches the right, and there is space between the scrollbar and main content container.

Screenshot 2025-03-10 at 7 53 55 PM
Screenshot 2025-03-10 at 7 54 00 PM

### DIFF
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -158,8 +158,8 @@
   }
 
   .#{$button}.pf-m-inline {
-    --#{$button}--m-link--Color: var(--#{$banner}--link--Color);
-    --#{$button}--m-link--hover--Color: var(--#{$banner}--link--hover--Color);
+    --#{$button}--m-link--m-inline--Color: var(--#{$banner}--link--Color);
+    --#{$button}--m-link--m-inline--hover--Color: var(--#{$banner}--link--hover--Color);
     --#{$button}--disabled--Color: var(--#{$banner}--link--disabled--Color);
 
     text-decoration-line: var(--#{$banner}--link--TextDecoration);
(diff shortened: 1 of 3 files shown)


srambach (approved):
These changes all look good to me.

PR #7400, opened 2025-03-11
feat(clipboard-copy): added support for truncation

### LINKED ISSUE #7399: Clipboard copy - support for truncation
re: https://github.com/patternfly/patternfly-react/pull/11610

We'll rely on the `<Truncate>` component to do the truncation, but we need to keep the layout (actions) from wrapping.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7399

@sg00dwin may be able to come up with a way to keep the existing non-flex/`inline` layout, but here's a PR using `inline-flex` for now.


thatblindgeye (approved):
Copy pasted the code from my React PR https://github.com/patternfly/patternfly-react/pull/11610 and added the new class, looks good 🥳

sg00dwin (approved):
/lgtm

PR #7404, opened 2025-03-17
fix(alert-group): only apply incoming animation to the newest alert

srambach:
Tweak to alert animation to apply the incoming animation (sliding down from the top) only to the most recent - i.e. the first - alert. This supports React PR https://github.com/patternfly/patternfly-react/pull/11495 where the animations are connected to adding and removing alerts from the alert group.


mcoker (approved):
LGTM

PR #7416, opened 2025-03-25
fix(menu-toggle): correct avatar alignment within menu-toggle 

### LINKED ISSUE #7410: Bug - [Menu-toggle] - Avatar within a menu-toggle is not in alignment with text and icon
**Describe the problem**
This is an issue with 
Patternfly 6  

https://www.patternfly.org/components/menus/menu-toggle#with-avatar-and-text

Image

Avatar should be centered with text and icon

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/7410

Aligns the menu-toggle icon `img` correctly. Also removed `pf-m-avatar` modifier.

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -567,16 +567,28 @@
   flex-wrap: nowrap;
 }
 
+%pf-v6-menu-toggle__icon--MarginBlock {
+  margin-block-start: calc(var(--#{$menu-toggle}--PaddingBlockStart) * -1);
+  margin-block-end: calc(var(--#{$menu-toggle}--PaddingBlockEnd) * -1);
+}
+
 .#{$menu-toggle}__icon {
   flex-shrink: 0;
-
+  
+  @at-root .#{$menu-toggle}:not(.pf-m-plain) & {
+    @extend %pf-v6-menu-toggle__icon--MarginBlock;
+  }
+  
   &.pf-m-avatar {
     .#{$avatar},
     img,
     svg {
-      margin-block-start: calc(var(--#{$menu-toggle}--PaddingBlockStart) * -1);
-      margin-block-end: calc(var(--#{$menu-toggle}--PaddingBlockEnd) * -1);
+      @extend %pf-v6-menu-toggle__icon--MarginBlock;
     }
+  }
+
+  :where(picture, img) {
+    vertical-align: middle;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -570,13 +570,8 @@
 .#{$menu-toggle}__icon {
   flex-shrink: 0;
 
-  &.pf-m-avatar {
-    .#{$avatar},
-    img,
-    svg {
-      margin-block-start: calc(var(--#{$menu-toggle}--PaddingBlockStart) * -1);
-      margin-block-end: calc(var(--#{$menu-toggle}--PaddingBlockEnd) * -1);
-    }
```
mcoker: We'll want to keep this to avoid a breaking change, but I think taking it out of the docs like you did is great, and we'll also want to take `menu-toggle-icon--HasAvatar` off of these examples

https://github.com/patternfly/patternfly/blob/cbe70fe435e751db884c38bf1b9bfac53ef4bab7/src/patternfly/components/MenuToggle/examples/MenuToggle.md?plain=1#L283

https://github.com/patternfly/patternfly/blob/cbe70fe435e751db884c38bf1b9bfac53ef4bab7/src/patternfly/components/MenuToggle/examples/MenuToggle.md?plain=1#L297

https://github.com/patternfly/patternfly/blob/cbe70fe435e751db884c38bf1b9bfac53ef4bab7/src/patternfly/components/MenuToggle/examples/MenuToggle.md?plain=1#L311

mcoker (changes requested):
Looking good! Next we'll need to make sure this avatar (or any large icon) doesn't change the menu toggle height. I tried this and it seemed OK but I didn't text extensively. Also it breaks plain/icon-only toggles:


```
.pf-v6-c-menu-toggle__icon {
  margin-block-start: calc(-1 * var(--pf-v6-c-menu-toggle--PaddingBlockStart));
  margin-block-end: calc(-1 * var(--pf-v6-c-menu-toggle--PaddingBlockEnd));
}
```

srambach:
I think I'm seeing shift down of a couple pixels on the avatar menu toggle compared to the others, but I'm not sure why.
image

mcoker (approved):
LGTM!

srambach:
> I think I'm seeing shift down of a couple pixels on the avatar menu toggle compared to the others, but I'm not sure why. image

I'm fine with merging and if it seems to really be a problem for consumers, we can put in a bug for the alignment.

srambach (approved):

PR #7419, opened 2025-03-27
fix(font): update RedHat[Text/Display/Mono] to latest

### LINKED ISSUE #7330: Bug - Red Hat Mono font - needs to be updated once https://github.com/RedHatOfficial/RedHatFont/issues/64 is fixed upstream
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

See https://github.com/RedHatOfficial/RedHatFont/issues/64.  Once that issue is addressed upstream, PatternFly needs to adopt the patched font.

cc: @nicolethoen

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7330

@lboehling @andrew-ronaldson the bold font-weights now render lighter in both RedHatText and RedHatDisplay, along with a couple of other small changes. Do we want to make any updates to font-weight with this change?

Here's a backstop report of the title and progress component that shows the bold weight differences, a slash-zero change, and looks like the number "2" renders a little differently now? [backstop-font-3-26-25.pdf](https://github.com/user-attachments/files/19477616/backstop-font-3-26-25.pdf)


andrew-ronaldson:
On one hand I want to align with brand but the other I find this change problematic. The hierarchy had clearer distinction when it was a combination of size and weight but this change shifts to be size only and that is difficult on the lower headings. @lboehling are the current headings aligned with the DPO or was there any adjustments for products needs?

dlabaj (approved):

srambach:
I still see slashed zeros in the preview site - is that an artifact of the way we do the preview and it's not getting the updated font?

mcoker:
@srambach hmm... have you tried an incognito window and/or another browser? It looks updated to me

Screenshot 2025-03-27 at 8 06 47 AM

srambach:
Hmm, I see that on progress, but I was looking at the code block. So I guess Red Hat Mono still has the slashed zero? (ironically it looks like if you set `font-variant: slashed-zero` it removes it...)

mcoker:
@lboehling do you know if there is an update/ETA for when a new font file may be available?

andrew-ronaldson (approved):

mcoker:
Closing in favor of https://github.com/patternfly/patternfly/pull/8368

PR #7421, opened 2025-03-27
fix(rtl): fixes RTL styling

### LINKED ISSUE #7292: Bug - Drawer - doesn't work in RTL
Found by @nicolethoen in https://github.com/patternfly/patternfly-react/pull/11404#pullrequestreview-2541063867

You can see the drawer disappear in when toggling RTL on any drawer or notification drawer demo:
* https://core-staging.patternfly.org/components/drawer/html-demos/expanded-drawer-with-jump-links/
* https://core-staging.patternfly.org/components/notification-drawer/html-demos/expanded-read/

### LINKED ISSUE #7414: Bug - Switch - `rtl` not adjusted correctly
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Within the Switch component when rtl is enabled the switch is not adjusted correctly, instead it seems shifted over, but the animation and off-circle is the same location.

See
https://codesandbox.io/p/sandbox/romantic-noyce-d2glyq

**Expected behavior**
Works as it should

Seems like `--pf-v6-global--inverse--multiplier` isn't changed for rtl, can workaround it by adding

```css
:root[dir="rtl"] {
    --pf-v6-global--inverse--multiplier: -1;
}
```

Can workaround it ourselves for now

**Screenshots**
Image

**What is your environment?**
 - OS: Linux
 - Browser Chromium
 - Version 134

**Any other information?**

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7414
fixes https://github.com/patternfly/patternfly/issues/7292

This fixes RTL styles that reference the `--pf-v6-global--inverse--multiplier` var when using `dir="rtl"`. To test, load any comnponents that use `@include pf-v6-bidirectional-style` (which is used mostly with styles that `translate` something directionally) or `@include pf-v6-mirror-inline-on-rtl`, which flips/mirrors icons mostly (like left/right directional carets)

A few that were broken and should now be fixed:
`pf-v6-bidirectional-style` - drawer, truncate, slider, switch, dual-list-selector
`pf-v6-mirror-inline-on-rtl` - tabs, card, tree view, table, data-list (probably the "expand" button in these)

### DIFF
--- a/src/patternfly/base/patternfly-variables.scss
+++ b/src/patternfly/base/patternfly-variables.scss
@@ -15,6 +15,6 @@
   @include dark.pf-v6-tokens;
 }
 
-@include pf-v6-rtl {
+@include pf-v6-rtl(false) {
   @include pf-v6-set-inverse;
 }
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -370,15 +370,19 @@
 // [dir="rtl"] {
 //   background: red;
 // }
-@mixin pf-v6-rtl {
-  @at-root :where(.#{$pf-prefix}m-dir-rtl, [dir="rtl"]) #{&} {
+@mixin pf-v6-rtl($hasWhere: true) {
+  $sel: if($hasWhere,':where',':is');
+
+  @at-root #{$sel}(.#{$pf-prefix}m-dir-rtl, [dir="rtl"]) #{&} {
     @content;
   }
 }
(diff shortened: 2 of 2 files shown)


srambach (approved):
Looks good to me! Thanks for the real time work through and fix. 🪩 🕺🏻

sg00dwin (approved):

dlabaj (approved):

PR #7434, opened 2025-04-02
fix(alert-group): make animation respect RTL

### LINKED ISSUE #7365: Bug - Alert - animation updates for RTL
* Make sure the alerts go offstage in RTL
* Update any directional classes (eg, `pf-m-offstage-right`) to something other than the cardinal direction. Easy update could be `pf-m-offstage-inline-end` or `pf-m-offstage-[inline/horizontal]` or it could just be something more meaningful that doesn't indicate the direction and is more about what it does in the alert? Possibly just use `pf-m-offstage`?

srambach:
This does two things:

- Adds the RTL calculation to reverse the translate when appropriate
- Renames `.pf-m-offstage-top` and `.pf-m-offstage-right` to `.pf-m-incoming` and `.pf-m-outgoing` to be more descriptive for the state rather than the specific action. (Old classes remain until a breaking change.) I've created variables with the updated names and have used fallbacks to ensure that the old variables will continue to work until a breaking change can remove them.

React will continue to work because of the fallbacks/old classnames still existing, but a followup should be created in order to use the new classes.

Fixes #7365

### DIFF
--- a/src/patternfly/components/Alert/alert-group.scss
+++ b/src/patternfly/components/Alert/alert-group.scss
@@ -38,34 +38,58 @@
     var(--#{$alert-group}--m-toast__item--c-alert--TransitionTimingFunction)
     0s;
 
-  // Alert group item removal reduced motion
-  --#{$alert-group}--m-toast__item--m-offstage-right--TransitionDuration--opacity--default: var(--pf-t--global--motion--duration--fade--default);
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):

mcoker (approved):
🚀

PR #7435, opened 2025-04-03
fix(form-control): remove browser prefixed properties

### LINKED ISSUE #7425: [FormControl] form control's CSS uses non-standard css properties for appearance

Debugging porting to PF 6 I stumbled upon the use of [non-standard css](https://github.com/patternfly/patternfly/blob/65724354f98fa0e44a25f1d82dee54d4883546a3/src/patternfly/components/FormControl/form-control.scss#L199) `-webkit-appearance` and `-moz-appareance` while [appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) is part of the css baseline.

**Any other information?**

https://developer.mozilla.org/en-US/docs/Web/CSS/appearance

mcoker:
fixes #7425


srambach (approved):
:shipit:

PR #7442, opened 2025-04-07
fix(drawer): introduce opacity for reduced motion

### LINKED ISSUE #7360: Animation: Drawer - Add reduced motion
Drawer slides out, but we haven't implemented reduced motion. Since this is a large area that slides, we should create a reduced motion version.

~~Likely we want to fade the opacity out before the space is taken away, and fade it in after creating the space.~~

We can replace the slide-in-out-action with opacity-fade-in-out-action with default transition tokens to start out with and can edit in review.

srambach:
Fixes #7360

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -27,15 +27,13 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --#{$drawer}__panel--RowGap: var(--pf-t--global--spacer--sm);
   --#{$drawer}__panel--PaddingBlockStart: var(--pf-t--global--spacer--sm);
   --#{$drawer}__panel--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
-  --#{$drawer}__panel--TransitionDelay--expand--visibility: 0s; // on expand, transition visible immediately
-  --#{$drawer}__panel--TransitionDelay--collapse--visibility: var(--#{$drawer}__panel--TransitionDuration--opacity); // on collapse, delay until opacity transitions
-  --#{$drawer}__panel--TransitionDelay--visibility: var(--#{$drawer}__panel--TransitionDelay--collapse--visibility); // TODO should be collapse but this BREAKS first time for normal and all the time for reduced motion
-  --#{$drawer}__panel--TransitionDelay: 0s, var(--#{$drawer}__panel--TransitionDelay--visibility);
+  --#{$drawer}__panel--TransitionDelay: 0s, var(--#{$drawer}__panel--TransitionDelay--focus), var(--#{$drawer}__panel--TransitionDelay--focus);
+  --#{$drawer}__panel--TransitionDelay--focus: var(--#{$drawer}__panel--TransitionDuration--fade);
+  --#{$drawer}__panel--TransitionDelay--expand--focus: 0s; // on expand, transition visible immediately
   --#{$drawer}__panel--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--decelerate);
-  --#{$drawer}__panel--TransitionDuration--visibility: var(--pf-t--global--motion--duration--fade--default);
-  --#{$drawer}__panel--TransitionDuration--opacity: var(--pf-t--global--motion--duration--fade--default);
-  --#{$drawer}__panel--TransitionDuration: var(--#{$drawer}__panel--TransitionDuration--opacity), 0s;
-  --#{$drawer}__panel--TransitionProperty: opacity, visibility; // by default, fade the drawer
+  --#{$drawer}__panel--TransitionDuration--fade: var(--pf-t--global--motion--duration--fade--default);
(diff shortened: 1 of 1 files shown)


mcoker:
@srambach pushed some updates. From what we were looking at
* there was a scoping issue like you suspected with this var, keeping the expanded visibility delay from triggering https://github.com/patternfly/patternfly/blob/1b1916a6986d7f3b485e35ad4cdd394640999c2c/src/patternfly/components/Drawer/drawer.scss#L233
* and `transform` (for the `translateX()`) wasn't being transitioned with the default/opacity transition, so it was going off screen immediately.

I pushed an update with the changes below. One thing I didn't test was in nested drawers, so we will want to test that.
* fixes the things above
* removes the `margin`, `flex-basis`, and `box-shadow` transitions. `margin` and `flex-basis` were being used in inline drawers to make the content grow/shrink as the panel was expanded/collapsed. Just took it off and IMO it looks fine - the transition happens fast enough, it looks OK to me. Curious what @lboehling @andrew-ronaldson think. I also removed the `box-shadow` transition - I _think_ it was used so the panel didn't slide off screen (with the box-shadow still visible), then the shadow would just disappear. I'm not seeing an issue with that here with the box-shadow transition disabled.
  * That should address any performance issues, we're only transitioning `transform` now.
* another thing we could do is transition `opacity` in the no-pref animation - fade it in as it expands, and fade out as it collapses. I didn't include that in this PR, but I tested it and the diff below is what I changed to enable it FWIW.

```diff
diff --git a/src/patternfly/components/Drawer/drawer.scss b/src/patternfly/components/Drawer/drawer.scss
index c6e3fe6dc..65aecf54a 100644
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -55,11 +55,10 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   @media screen and (prefers-reduced-motion: no-preference) {
     --#{$drawer}__panel--TransitionDuration--slide: var(--pf-t--global--motion--duration--slide-in--short);
     --#{$drawer}__panel--TransitionDelay--focus: var(--#{$drawer}__panel--TransitionDuration--slide);
-    --#{$drawer}__panel--TransitionDelay: 0s, 0s, 0s, 0s, var(--#{$drawer}__panel--TransitionDelay--focus);
+    --#{$drawer}__panel--TransitionDelay: 0s, 0s, 0s, 0s, var(--#{$drawer}__panel--TransitionDelay--focus), 0s;
     --#{$drawer}__panel--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--decelerate);
-    --#{$drawer}__panel--TransitionDuration: 0s, var(--#{$drawer}__panel--TransitionDuration--slide), 0s, 0s, 0s;
-    --#{$drawer}__panel--TransitionProperty: margin, transform, box-shadow, flex-basis, visibility; // with no preference, slide the panel
-    --#{$drawer}__panel--Opacity: 1;
+    --#{$drawer}__panel--TransitionDuration: 0s, var(--#{$drawer}__panel--TransitionDuration--slide), 0s, 0s, 0s, var(--#{$drawer}__panel--TransitionDuration--slide);
+    --#{$drawer}__panel--TransitionProperty: margin, transform, box-shadow, flex-basis, visibility, opacity; // with no preference, slide the panel
   }
 
   // Drawer panel head

```

andrew-ronaldson (approved):
Beauty never fades... unless you turn on reduced motion.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -46,6 +52,16 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --#{$drawer}--m-panel-bottom__panel--m-resizable--md--FlexBasis--min: #{pf-size-prem(24px)};
   --#{$drawer}--m-panel-bottom__panel--m-resizable--MinHeight: #{pf-size-prem(24px)};
 
+  @media screen and (prefers-reduced-motion: no-preference) {
+    --#{$drawer}__panel--TransitionDuration--slide: var(--pf-t--global--motion--duration--slide-in--short);
+    --#{$drawer}__panel--TransitionDelay--focus: var(--#{$drawer}__panel--TransitionDuration--slide);
+    --#{$drawer}__panel--TransitionDelay: 0s, 0s, 0s, 0s, var(--#{$drawer}__panel--TransitionDelay--focus);
+    --#{$drawer}__panel--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--decelerate);
```
mcoker, line 59: Looks like this is a dupe of [this line](https://github.com/patternfly/patternfly/blob/ef1322550b7ca56940a746c4fbe63cd25e188bce/src/patternfly/components/Drawer/drawer.scss#L33) and could be removed.

mcoker (approved):
Just one little nit, otherwise lgtm 🚀

sg00dwin (commented):
LGTM!

Issue #7377, opened 2025-03-05
Page - add top-level modifiers to indicate nav expanded/collapsed, horizontal nav, etc


mcoker:
Re: https://github.com/patternfly/patternfly-doc-core/issues/26

Right now the style that control nav expanded/collapsed, and whether the main content takes up the entire viewport or not are dependent on markup structure (eg `.pf-v6-c-page__sidebar.pf-m-collapsed + .pf-v6-c-page__main-container`).

This is fragile and is currently not working in our docs framework (from the issue linked above) because the structure is roughly this below and `__sidebar` is not a sibling of `__main-container`.

```
 div.pf-c-page
  div.pf-c-masthead
  astro-island // which has `display: contents;`
    div.pf-c-page__sidebar
  div.pf-c-page__main-container
```

The page component should be able to dictate the grid layout for its children. It would be great if we had classes that could be used to indicate the states we have today:
* sidebar is open
  * on mobile, sidebar overlays content and content spans full width of page grid
  * on desktop, sidebar takes up left part of page grid, main content takes up right part
* sidebar is collapsed 
  * sidebar is off screen, main content spans full width of page grid
* page has horizontal nav
  * content spans full width of page grid

PR #7455, opened 2025-04-11
fix(page): fix main layout with no sidebar

### LINKED ISSUE #7357: Page - Allow main grid without sidebar and masthead
`<Page>`

**Describe the feature**
Currently there is no way to configure Page component to not have the grid whitespace of a missing sidebar when you are missing both the sidebar and header grid area but only have a main layout. One use-case for this is when you are putting a page main content inside an iframe. As you still want the correct layout. If we were to just use PageSection or similar we would not get the correct CSS applied and would have to introduce our own CSS

With the iframe system we have it looks like this
Image
Image

Using the CSS override 
```css
@media (min-width: 75rem) {
  .pf-v6-c-page.no-masthead-sidebar { /* custom class to scope this style to a specific page component instance */
    --pf-v6-c-page__main-container--GridArea: var(--pf-v6-c-page--masthead--main-container--GridArea);
  }
}
```
we get
Image

**Any other information?**
Current CSS introducing the problem
https://github.com/patternfly/patternfly/blob/917bda02e6c5d2eb21a44893ed63f12b2286aefe/src/patternfly/components/Page/page.scss#L340-L344

### LINKED ISSUE #7452: Bug - Page - Whitespace on the left side of the screen from sidebar styling
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  (6.2.0)
- [ ] other

When we show an error as an alert at the top of the screen, it's got a buffer of whitespace between the left side of the screen and the alert due to [this change.](https://github.com/patternfly/patternfly/commit/8fe1357d8971b40d388b0ba4bbae76b34956d704#diff-cc315fba6c5159bc6259ec0b3a8b416ff6a7aeb55ad77b73900a9c5fcb004269L138) 

In https://github.com/opendatahub-io/odh-dashboard

1. Run npm install && npm run dev:start:ext in the frontend folder (after connecting to a cluster)
2. Load up localhost & wait for the home page to show up
3. CTRL+C the previous npm terminal
4. Wait for timeout — see related images

**Expected behavior**

To fix the issue in https://github.com/opendatahub-io/odh-dashboard, we added this change cc @mattnolting - will need to apply the same fix here so we can remove this override in ODH.

```
.pf-v6-c-page {
  --pf-v6-c-page__sidebar--Width: 0;

  &:has(.pf-v6-c-page__sidebar) {
    --pf-v6-c-page__sidebar--Width: 18.125rem;
  }
}
```

**Screenshots**

Current:
Image

Expected:
Image

**What is your environment?**
 - Browser [e.g. chrome, safari] Chrome

RHOAI 2.21

**Any other information?**
Following our discussion @mattnolting - we need to pass a check to verify that the sidebar exists to fix the alignment of the page content.

mcoker:
Partially addresses https://github.com/patternfly/patternfly/issues/7377
Fixes https://github.com/patternfly/patternfly/issues/7452
Fixes https://github.com/patternfly/patternfly/issues/7357

Adds `.pf-m-no-sidebar` to `.pf-v6-c-page` for layouts without a sidebar as an opt-in so it doesn't impact existing page layouts.

### DIFF
--- a/src/patternfly/components/Page/examples/Page.md
+++ b/src/patternfly/components/Page/examples/Page.md
@@ -268,6 +268,7 @@ This component provides the basic chrome for a page, including sidebar and main 
 | `.pf-v6-c-page__main-body` | `<div>` | Creates the body section for a page section. **Required** |
 | `.pf-v6-c-page__main-group` | `<div>` | Creates the group of `.pf-v6-c-page__main-*` sections. Can be used in combination with `.pf-m-sticky-[top/bottom]` to make multiple sections sticky. |
 | `.pf-v6-c-page__drawer` | `<div>` | Creates a container for the drawer component when placing the main page element in the drawer body. |
+| `.pf-m-no-sidebar` | `.pf-v6-c-page` | Modifies the page grid for layouts without a sidebar. |
 | `.pf-m-expanded` | `.pf-v6-c-page__sidebar` | Modifies the sidebar for the expanded state. |
 | `.pf-m-collapsed` | `.pf-v6-c-page__sidebar` | Modifies the sidebar for the collapsed state. |
 | `.pf-m-page-insets` | `.pf-v6-c-page__sidebar-body` | Modifies a sidebar body padding/inset to visually match padding of page elements. |
--- a/src/patternfly/components/Page/page.hbs
+++ b/src/patternfly/components/Page/page.hbs
@@ -1,4 +1,4 @@
-<div class="{{pfv}}page{{#if page--modifier}} {{page--modifier}}{{/if}}"
+<div class="{{pfv}}page{{#if page--HasNoSidebar}} pf-m-no-sidebar{{/if}}{{#if page--modifier}} {{page--modifier}}{{/if}}"
   {{#if page--attribute}}
     {{{page--attribute}}}
   {{/if}}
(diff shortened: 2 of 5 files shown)


srambach:
Without the masthead, do we care that there is no space at the top of the white area? If the use case really is a whole page and no masthead I think it would look better with some. OTOH if it gets used embedded into something else, I'm not sure but also that seems like a bad idea to embed the whole page in something.

srambach (approved):
Code looks good - just wondering about providing spacing at the top if there is no masthead.

mcoker:
@srambach good question! My opinion - this is for no sidebar, so I'd say if that works, we could go ahead and get this through. I updated the PR title from "no masthead/sidebar" to just "no sidebar" since that's what it's about - even though it fixed the use case of no masthead/sidebar.

On the top padding, I'm not sure what all use cases there are for layouts without a masthead. Looking at https://github.com/patternfly/patternfly/issues/7357, if I'm looking at the screenshots right, they are creating their own type of masthead, so they likely wouldn't want any sort of top padding. But looking at https://github.com/patternfly/patternfly/issues/7452, this is just an alert on an empty page, so they likely would want top padding.

There is probably a bigger question of whether those layouts are built the way we would expect, but assuming they are, I think we would want to support both. Possibly something like `.pf-m-no-masthead`, which removes the masthead grid-row and adds top padding to the page layout, and maybe `.pf-m-no-padding-top` to remove it?

We could tag design to review and give their dime, or merge this and create a follow-up to look at that. I'd probably be inclined to merge this as-is so we can go forward with the patch release. WDYT?

srambach:
@mcoker I'm fine to go ahead and merge since it's no worse than it was and it fixes an existing problem.

PR #7457, opened 2025-04-11
fix(A11y): resolved issues with latest pf- a11y package version

### LINKED ISSUE #7193: Bug - A11y issues flagged by updated pf-a11y tool
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

When attempting to update to the latest version of the patternfly-a11y dependency (5.0.0) new a11y issues are getting flagged
See a11y report from the [dependabot's version bump PR](https://github.com/patternfly/patternfly/pull/7191): https://patternfly-pr-7191-a11y.surge.sh/

- [x] card demo has invalid `aria-controls` values (the `aria-controls` on a table's expandable row toggle does not match the id of the expandable content) (fixed by https://github.com/patternfly/patternfly/pull/7194)
- [ ] dividers in a list cannot have a `role="separator"` - I suggest using aria-hidden in those cases?
- [ ] drawers which use a draggable splitter are missing a required `aria-valuenow` attribute. This is because axe assumes that an element with `role="separator"` and `tabindex=0` is being used as an adjustable component and so it needs an `aria-valuenow` prop. In the case of the core docs, since the JS has not enabled actual adjusting of the splitters, adding the attribute `aria-valuenow="50"` will resolve this issue
- [x] multiple table examples have invalid `aria-controls` values (the `aria-controls` on a table's expandable row toggle does not match the id of the expandable content) (fixed by https://github.com/patternfly/patternfly/pull/7194)
- [ ] multiple table examples have a `toggle all` control in their `thead` which has an invalid `aria-controls` value. In such cases, the `aria-controls` value should be a space delimited list of the ids of each row's expandable sections.
- [ ] Tabs (`ul` elements with `role='tablist'`) cannot have children with any role other than `role='tab'`. So all of our examples which include a button beside the tab for additional actions, such as information icons or close buttons, are violating that rule. This may require some redesign?

Once these are all addressed, we could consider turning back on the CI rule that requires the a11y tool to pass in order to merge the PR

thatblindgeye:
Closes #7193 

Also bumping the a11y tool version here, so https://github.com/patternfly/patternfly/pull/7352 can be closed out once this goes in

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@fortawesome/fontawesome": "^1.1.8",
     "@octokit/rest": "^20.1.0",
     "@patternfly/documentation-framework": "6.9.1",
-    "@patternfly/patternfly-a11y": "5.0.0",
+    "@patternfly/patternfly-a11y": "5.1.0",
     "@patternfly/react-code-editor": "6.2.0",
     "@patternfly/react-core": "6.2.0",
     "@patternfly/react-table": "6.2.0",
--- a/patternfly-a11y.config.js
+++ b/patternfly-a11y.config.js
@@ -74,7 +74,8 @@ module.exports = {
     'color-contrast',
     'landmark-no-duplicate-main',
     'landmark-main-is-top-level',
-    'scrollable-region-focusable'
+    'scrollable-region-focusable',
+    'aria-required-children' // Remove once https://github.com/patternfly/patternfly-react/issues/9968 is resolved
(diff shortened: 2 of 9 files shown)


nicolethoen (approved):

mcoker (approved):
🦾🤖
*"bleep bloop boopity bleeboop"* (LGTM)

PR #7458, opened 2025-04-14
fix(CodeEditor): remove command palette underlining

logonoff:
reverts `normalize.scss` changes to the command palette

before:
image

after:
image


mcoker (approved):
🚀

PR #7473, opened 2025-04-18
feat(text-input-group): animate error status

### LINKED ISSUE #7342: Animation: Form control (TextInput, TextArea, FormSelect) add error validation jiggle

### LINKED ISSUE #7343: Animation: Text input groups add error validation jiggle

sg00dwin:
https://github.com/patternfly/patternfly/issues/7343 and https://github.com/patternfly/patternfly/issues/7342

Create `mixin`'s that are applied to animate error status for `text-input-group` and `form-control` with `input` `textarea` `select` 

Working example https://codesandbox.io/p/sandbox/distracted-cray-mzkk8p

### DIFF
--- a/src/patternfly/sass-utilities/mixins.scss
+++ b/src/patternfly/sass-utilities/mixins.scss
@@ -289,8 +289,6 @@
 @mixin pf-v6-animate-danger-jiggle($element) {
   --#{$element}--m-danger--AnimationDuration--Transform: var(--pf-t--global--motion--duration--fade--default);
   --#{$element}--m-danger--AnimationTimingFunction--Transform: var(--pf-t--global--motion--timing-function--default);
-  --#{$element}__status-icon--m-danger--TransitionDuration--Opacity: var(--pf-t--global--motion--duration--icon--default);
-  --#{$element}__status-icon--m-danger--TransitionTimingFunction--Opacity: var(--pf-t--global--motion--timing-function--default);
 
   translate: var(--#{$element}--m-danger--TranslateX, 0);
   animation-name: #{$element}-m-danger-motion;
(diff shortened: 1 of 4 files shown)


src/patternfly/sass-utilities/mixins.scss
```diff
@@ -276,6 +276,58 @@
...
+      
+      66% {
+        --#{$element}--m-danger--TranslateX: 3px;
+      }
+    }
+  }
+}
+
+// Animate icon fade in
+@mixin pf-v6-animate-icon-fade-in($element) {
```
srambach: I was pondering whether we want to rename this `pf-v6-animate-status-icon-fade-in` since the variables are `__status-icon...` but I see that actually only menu-toggle has `__status-icon`. And then just below, the animation is `-status-icon-fade-in` but it's such a generic set of keyframes that I was wondering if this could be super general like just `-fade-in` or something and made available more widely. I think everything is fine, but I was a little confused by the names so maybe we could think through and spell out a naming convention for mixins and animations? I think it's generally consistent so far but I also don't think we have this particular types of naming spelled out anywhere.

sg00dwin: I pushed some changes to make the mixin more general.

sg00dwin:
@mcoker @srambach 
Based on Sarahs comment, I changed the `mixin` to be generic, and introduced an `$fade` argument to enable `fade-in` or `fade-out`. Plus updated its name to `pf-v6-fade-default` since it uses the default global values for its animation properties. Let me know what you think.

mcoker (commented):
This LGTM. A few comments:

For the vars that get generated in `pf-v6-animate-danger-jiggle`
  * These are component specific - eg, `--#{$text-input-group}--m-danger--[...]`. If they're going to be component specific like that, we should add those at the top of the component with the rest of the main component vars. We don't typically assign component vars that have values of `--pf-t--global--[...]` outside of the vars at the top of a component stylesheet.
  * If they don't have to be component specific, these could be something like `--#{$pf-global}--danger--jiggle--Animation[...]` and you could create them once, then every component that uses them could reference those variables.

For the `animation-name`, jiggle is `animation-name: #{$element}-m-danger-motion;` - component specific. And fade is `animation-name: $fade;` where `$fade` is `pf-v6-global-fade-in` - not component specific. Those should probably match in the way they're named. You could create a single animation in our global/common CSS that each component references (so not generated in each component stylesheet), or you could keep it component specific.

Lemme know what you think - depending on what kind of updates we want to make, we could merge this as it is and create a follow-up (and just get it in before we cut the q2 release) or you could go ahead and make the changes now.k

srambach (approved):
L🪇TM

mcoker:
Closing in favor of https://github.com/patternfly/patternfly/pull/7545/ which was branched off of this PR branch.

PR #7485, opened 2025-04-24
fix(truncate): enable inline links to have underline when used with truncate

### LINKED ISSUE #7255: Bug - Truncate - removes underline from link
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

From @DaoDaoNoCode:
If a Truncate component is used within an <a> tag, that underscore goes away because the truncate has style display: inline-grid

Provide steps to reproduce. A [codepen](http://codepen.io) or [codesandbox](https://codesandbox.io/) demonstrating the problem is appreciated.

**Expected behavior**
The underline should still show up on the truncated text - is this correct @lboehling @andrew-ronaldson ?

**Screenshots**
Image

sg00dwin:
https://github.com/patternfly/patternfly/issues/7255

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -770,6 +770,13 @@
   }
 }
 
+// enable inline link to have underline when used with truncate
+.#{$button}__text {
+  // stylelint-disable property-disallowed-list
+  text-decoration: inherit;
+  // stylelint-enable property-disallowed-list
+}
+
 .#{$button}__count {
   display: inline-flex;
   align-items: center;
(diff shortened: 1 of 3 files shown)


srambach (commented):
I found one other spot - Timestamp with help text `pf-v6-c-timestamp pf-m-help-text` - that might need the same treatment?

sg00dwin:
@srambach updated to allow timestamp with tooltip to have a dashed underline

srambach (approved):
LGTM! 🐘

mcoker (approved):

PR #7493, opened 2025-04-30
fix(dual-list-selector): animate tree expansion

### LINKED ISSUE #7325: Animation: Dual list - Expand/collapse interaction
Part of the https://github.com/patternfly/pf-roadmap/issues/217 

### This is the expansion pattern from design

Open expandable content

Caret

Close expandable content

Caret

### LINKED ISSUE #7530: Expandable section - keep animation from firing on page load
Currently expandable section and others transition `display: none`, which requires setting a `@starting-style` that has a byproduct of causing transitions to run on page load if the component is in the expanded state.

Instead of transitioning `display`, we can try transitioning `visibility` along with `height` for collapsed/expanded states, and the `inert` attribute will be added when in the collapsed state to keep children from being visible/focusable.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7325
fixes #7530

### DIFF
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -221,18 +221,13 @@ $pf-v6-c-dual-list-selector__item--MaxNesting: 10;
 
     > .#{$dual-list-selector}__list {
       .#{$dual-list-selector}.pf-m-animate-expand & {
-        // stylelint-disable max-nesting-depth, selector-max-class
-        &:where([hidden]) {
-          display: flex;
-        }
-        // stylelint-enable
-
-        display: none;
+        max-height: 0;
+        visibility: hidden;
         opacity: var(--#{$dual-list-selector}__list--Opacity);
+        transition-delay: 0s, 0s, var(--#{$dual-list-selector}__list--TransitionDuration--fade), var(--#{$dual-list-selector}__list--TransitionDuration--fade);
         transition-timing-function: var(--#{$dual-list-selector}__list--TransitionTimingFunction);
-        transition-duration: var(--#{$dual-list-selector}__list--TransitionDuration--fade), var(--#{$dual-list-selector}__list--TransitionDuration--slide), var(--#{$dual-list-selector}__list--TransitionDuration--fade);
-        transition-property: opacity, translate, display;
-        transition-behavior: allow-discrete;
+        transition-duration: var(--#{$dual-list-selector}__list--TransitionDuration--fade), var(--#{$dual-list-selector}__list--TransitionDuration--slide), 0s, 0s;
+        transition-property: opacity, translate, visibility, max-height;
         translate: 0 var(--#{$dual-list-selector}__list--TranslateY);
       }
     }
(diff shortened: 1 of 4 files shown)


andrew-ronaldson (approved):
Crushing these issues bruh!

mcoker:
@kmcfaul @thatblindgeye looks like dual list selector tree in react conditionally renders the nested items when they're expanded. I _think_ that happens right here - https://github.com/patternfly/patternfly-react/blob/f8b21b594c90960305e3b58b723c1cff7a9decbb/packages/react-core/src/deprecated/components/DualListSelector/DualListSelectorTreeItem.tsx#L159

I think we'll need to either
1) To keep everything the way it is in the DOM now and continue to conditionally render
    * On expand, render the nested tree as we do now, then use the `setTimeout()` method to apply `.pf-m-expanded` on the parent item.
    * On collapse, wait for the `transitionend` event of the nested tree before unmounting
2) Let CSS manage hiding/showing the nested tree. That would probably mean the react component always renders the nested tree. Then the rest can be managed with CSS.

Do you have a preference? Or any other thoughts?

thatblindgeye:
@mcoker gonna try fiddling around with React locally using this PR update, but at the moment I feel like personally I might lean towards making it more opt-in for React for now and updating in the next breaking change to be the default (similar to Alert).

- Making it opt in means it won't be breaking for consumers to have the dual list selector content always rendered rather than dynamically rendered
- Avoids any potential issues relying on the timeout to add that class to get animations working
- Would be consistent with how the other expansion animations on ExpandableSection and Accordion are working right now

thatblindgeye:
@mcoker @kmcfaul I think I have something locally that *may* not require a setTimeout but still work. There is an issue with nested tree view items in the DLS component, though, and seems like it's also in Core.

Toggling the `pf-m-expanded` class on a top most tree view item in the [dual list selector with tree example](https://patternfly-pr-7493.surge.sh/components/dual-list-selector/#tree-view) animates fine (the "Colors" item). But trying to toggle an item nested in that one (e.g. "Green") doesn't have the animations I'd expect.

kmcfaul:
I think I'd lean more towards a core solution with maybe an opt-in prop on react's side to allow the children to always be rendered, but I also think @thatblindgeye's solution avoiding the timeout and toggling the expanded class is good as well. The only issue I see is that the nested tree folders aren't being hidden when collapsed.

mcoker:
@kmcfaul @thatblindgeye fixed the nested item bug and converted it to transition `display` for the hide/show instead of relying on `hidden`

thatblindgeye (commented):
In React this looks great. The only thing is that by default (just the tree example as it curently is in React) the expand animation will still occur, but there's no collapse animation. Adding a `hasAnimations` prop and setting that up gets things working as expected, it's just the current behavior where the children are dynamically rendered rather than always rendered.

Would we be able to add a temporary class to disable animations for the current behavior, then in the next breaking change remove it (and the prop we'd need in React to get animations working)?

mcoker:
@thatblindgeye forgot about that! Added, `.pf-m-animate-expand` that goes on the outer `.pf-v6-c-dual-list-selector` wrapper.

thatblindgeye (approved):
*insert magneto-perfection.gif*

Looks like this should work in React!

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -199,10 +218,44 @@ $pf-v6-c-dual-list-selector__item--MaxNesting: 10;
 
   &.pf-m-expandable {
     --#{$dual-list-selector}__item--PaddingInlineStart: var(--#{$dual-list-selector}__item--m-expandable--PaddingInlineStart);
+
+    > .#{$dual-list-selector}__list {
+      .#{$dual-list-selector}.pf-m-animate-expand & {
+        // stylelint-disable max-nesting-depth, selector-max-class
+        &:where([hidden]) {
+          display: flex;
```
sg00dwin: I'm not following why this rule is needed. `.#{$dual-list-selector}__list` looks to be removed from the dom when in the unexpanded state.

sg00dwin:
I'm not seeing it animate on collapse. Should it?
dual-list-selector-animate

src/patternfly/components/DualListSelector/dual-list-selector.scss
mcoker: @sg00dwin good catch, that was a copy/paste from another component.

mcoker:
@sg00dwin if you're testing in react, it won't collapse because react removes the nested menu from the DOM on collapse. But that's going to change so that react always renders the element - @thatblindgeye and I have tested it on the react side so it should be GTG now!

src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -199,10 +218,34 @@ $pf-v6-c-dual-list-selector__item--MaxNesting: 10;
 
   &.pf-m-expandable {
     --#{$dual-list-selector}__item--PaddingInlineStart: var(--#{$dual-list-selector}__item--m-expandable--PaddingInlineStart);
+
+    > .#{$dual-list-selector}__list {
+      .#{$dual-list-selector}.pf-m-animate-expand & {
```
srambach: Do we want to leave a TODO anywhere about removing the opt-in in a breaking change?

srambach (approved):
Just the one question about if we want to leave a breaking change comment for later.

PR #7499, opened 2025-05-01
fix(page): add sidebar reduced motion animation

### LINKED ISSUE #6960: Page sidebar - add reduced motion default transition

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6960

Please review the tokens used 😅 I got these from the toast alert animation. BTW if there is a reference for which tokens should be used either here, or a general reference sheet, please let me know. I think it would be worthwhile to review all of our animations and make sure we're being consistent.


andrew-ronaldson (approved):
Cool beans!

srambach (approved):
LGTM

PR #7503, opened 2025-05-01
fix(tabs): add opt-in for current animation, performance

### LINKED ISSUE #7437: Tabs - animation and core styles, performance
Follow up from https://github.com/patternfly/patternfly/pull/7336

The accent now defaults to being drawn w/ JS. This will be a breaking change for core-only users (and keep the accent from showing in our docs). We'll need to address this.
  * If animation is an opt-in, that should take care of it. Technically this will be a "breaking" change as the component most likely relies on `style="--[pf-tabs]--accent-start --[pf-tabs]--accent-size"` rendered in the markup by react to make the accent work. So opt-in may be the best way to do it.
  * Not sure if it's possible, but I wonder if we could rely on one of the new vars (mentioned above) being set to disable the default/non-animation accent styles. So if the vars aren't set, the non-animation styles apply.

We should look at the animated properties to ensure they're performant.
* `width` should change to `scale`
* `inset-inline-start` should change to `translate`

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7437

Moves animation behind `.pf-m-animate-current`. Converts `width`/`inset-inline-start` transition to `scale`/`translate` for supercharged turbo performance TO THE MAX!!!

### DIFF
--- a/src/patternfly/components/Tabs/__tabs-list.hbs
+++ b/src/patternfly/components/Tabs/__tabs-list.hbs
@@ -38,7 +38,7 @@ __tabs-list--IsCloseDisabled: bool
     __tabs-item--id="containers"
     __tabs-item--icon-name="fas fa-box"
     __tabs-item--aria-label="Containers"
-    __tabs-item--text="Containers"
+    __tabs-item--text=(ternary __tabs-list--HasMultiLine "Containers <br> and more" "Containers")
     }}
   {{> __tabs-item
     __tabs-item--id="database"
--- a/src/patternfly/components/Tabs/examples/Tabs.md
+++ b/src/patternfly/components/Tabs/examples/Tabs.md
@@ -327,35 +327,39 @@ cssPrefix: pf-v6-c-tabs
 ```
 
 ## Animate current tab accent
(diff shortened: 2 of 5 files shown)


thatblindgeye (approved):
Looks good pulling these latest changes into React -- animated tabs animate and the HTML/class implementation used from Org in the local build looks how it should without animations

srambach (commented):
Seems fine but FWIW, the animated tabs do get messed up in RTL, so if that's not easy to fix, maybe we should remove the inline JS - if so, I'd add a comment on the example clearly saying that you won't see the effect here and link to the React example showing it.

sg00dwin (approved):
LGTM

mcoker:
@srambach good catch with RTL. Since this uses translate now, I needed to change the translate to a negative number in RTL. You can test by clicking on something, then switching to RTL and looking at the computed value.

Also updated the JS to get the "right" value by calcing the `container width` - `current item left` - `current item width`. @thatblindgeye is that what [the RTL util](https://github.com/patternfly/patternfly-react/blob/c10341f7c13354629d46d02f15bf27d071cbd1dd/packages/react-core/src/helpers/util.ts#L549-L575) you added does?

I'd still like to remove this JS once this PR is ready, or as a follow-up.

thatblindgeye:
The util is basically just taking the offsetLeft and adding it to the offsetWidth of the `pf-m-current` tab, and subtracting that from the offsetWidth of the outermost tabs list container.

srambach (approved):
L🐈TM (with js removed)
I wouldn't hate mentioning which classes and variables apply only to the animated version in the component docs.

srambach (approved):

PR #7508, opened 2025-05-06
fix(expandable-section): exclude truncate from animating

### LINKED ISSUE #7507: Bug - Expandable section - exclude truncate variant from animation
This variant can't be animated with the same method as regular expandable section because it uses css `line-clamp` to manage the expanded/collapsed state. I created a follow-up issue to look at that in our next breaking change to see if we can animate it - https://github.com/patternfly/patternfly/issues/7506

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7507


mcoker:
Closing for now - should be addressed in https://github.com/patternfly/patternfly/pull/7511

PR #7509, opened 2025-05-06
feat(button): add click animation

### LINKED ISSUE #7208: Buttons - Animate a clicked state of buttons (Core)
Follow up to design issue https://github.com/patternfly/patternfly-design/issues/1354

@kaylachumley's codepen - https://codepen.io/Kayla-Chumley/pen/OPJgBBN
Material example - https://codepen.io/sajran/pen/Wwyjwa

srambach:
Fixes #7208 

- Uses the :after pseudoelement that has a background radial gradient with 1% transparent and 1% `currentcolor`. `currentcolor` gives the best contrast without having to define a color for each variant (using a background gray doesn't show up well on the filled buttons). Design has initially ok'd the `customcolor` (you can see the slight tint on the secondary buttons and the link buttons)
- Added `overflow: hidden` to button to contain the ripple effect
- The effect happens when coming out of the :active state - in default and hover, the gradient is scaled so large that the transparent "hole" in the middle covers the button. The :active state changes the scale to 0 so that the whole background is infinitely small, then when the state moves out of :active, the scale up creates the ripple effect.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -25,6 +25,14 @@
   --#{$button}--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
   --#{$button}--TransitionProperty: color, background-color, border-width, border-color, padding;
 
+  // Button click (ripple) effect
+  --#{$button}--after--MaskPosition: center;
+  --#{$button}--after--MaskSize: 15000%;
+  --#{$button}--after--TransitionDuration: var(--pf-t--global--motion--duration--3xl);
+  --#{$button}--after--active--TransitionDuration: 0s;
+  --#{$button}--after--Opacity: 0;
+  --#{$button}--after--hover--Opacity: 0.12;
+
   // Hover
   --#{$button}--hover--BackgroundColor: transparent;
   --#{$button}--hover--BorderColor: transparent;
@@ -334,9 +342,13 @@
   text-decoration-style: var(--#{$button}--TextDecorationStyle);
   text-decoration-color: var(--#{$button}--TextDecorationColor);
   white-space: nowrap;
-  cursor: pointer; // needed for when a button does not use <button> - eg, <span>
+  cursor: pointer;
   user-select: none;
   background-color: var(--#{$button}--BackgroundColor);
+  background-image: var(--#{$button}--BackgroundImage);
+  background-repeat: no-repeat;
+  background-position: var(--#{$button}--BackgroundPosition);
+  background-size: var(--#{$button}--BackgroundSize);
   border: 0;
   border-start-start-radius: var(--#{$button}--BorderStartStartRadius, var(--#{$button}--BorderRadius));
   border-start-end-radius: var(--#{$button}--BorderStartEndRadius, var(--#{$button}--BorderRadius));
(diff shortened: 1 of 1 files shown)


andrew-ronaldson:
Could you please remove the clicked state animation from the inline link button. That one stands out to me but I think the rest are in a good place visually.

src/patternfly/components/Button/button.scss
```diff
@@ -346,7 +362,8 @@
   transition-duration: var(--#{$button}--TransitionDuration);
   transition-property: var(--#{$button}--TransitionProperty);
 
-  &::after {
+  // Border pseudo-element
+  &::before {
```
mcoker: I'd be hesitant to make this change. Technically it isn't breaking, but given how common the button is and the number of overrides that likely exist, I could see this causing some styling issues. This may be the best way to make this change, so I'm not saying change it necessarily, but just want to mention that in case you can think of ways we could avoid swapping before/after styles. I can also noodle on it after reviewing this PR more thoroughly.

sg00dwin:
General observation, when the animation is slowed down, it looks like upon click, the background `currentcolor` is scaled up from 0 to cover the button, then `radial-gradient` ripple is scaled from the center. Maybe that initial scale of the `currentcolor` is intentional, just noting it.

Also here are a few items I noticed when testing in Safari and Firefox.

**Safari**
These issues are only on link and control versions. Upon hover there's some background color changes (maybe that's the `currentcolor` scaling from 0). And then on click the `radial-gradient` isn't the same as what's shown on the `varients`. Also, on the icon only button there's something strange going on.  

https://github.com/user-attachments/assets/0954a81e-87ca-4c87-95f7-db0c974ef52f



**Firefox**
In some cases the hover background edge is visible. You can see it on the "Link danger" button and a inline link button with an arrow icon. 
                        
https://github.com/user-attachments/assets/94b866af-b821-4302-bc81-be260c966824

Here's the code for that inline button with arrow
```
<a class="pf-v6-c-button pf-m-inline pf-m-link" href="#">
  <span class="pf-v6-c-button__text">click</span>
  <span class="pf-v6-c-button__icon pf-m-end" style="">
    <i class="fas fa-arrow-right" aria-hidden="true"></i>
  </span>
</a>
```

srambach:
Moved the ripple effect to the button background.

src/patternfly/components/Button/button.scss
```diff
@@ -23,7 +23,7 @@
   --#{$button}--TextDecorationColor: currentcolor;
   --#{$button}--TransitionDuration: var(--pf-t--global--motion--duration--fade--default);
   --#{$button}--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
```
lboehling: can we adjust this token to use `--pf-t--global--motion--timing-function--accelerate` instead? reviewed with the design team today, and thought this adjustment made the ripple look smoother! thank you!

srambach: @lboehling Thanks - I updated that but just to make sure that's what you want, that variable does also affect the color and border transition (e.g. on hover). LMK if that's not what you intended.

mcoker (approved):
LGTM! 🤩

PR #7511, opened 2025-05-07
fix(expandable-section): transition display instead of hidden attr

### LINKED ISSUE #7507: Bug - Expandable section - exclude truncate variant from animation

### LINKED ISSUE #7510: Expandable section - toggle display instead of hidden
As part of the expandable animation, let's gauge the level of effort and potential risks of animating `display` instead of relying on `[hidden]`.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7510
fixes https://github.com/patternfly/patternfly/issues/7507

Allows `hidden` to still be applied (to avoid a breaking change) but the content is now shown/hidden by toggling `display: none`.

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section.scss
+++ b/src/patternfly/components/ExpandableSection/expandable-section.scss
@@ -21,8 +21,8 @@
   --#{$expandable-section}__content--TransitionDuration--fade: var(--#{$expandable-section}__content--TransitionDuration--collapse--fade);
   --#{$expandable-section}__content--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
   --#{$expandable-section}__content--Opacity: 0;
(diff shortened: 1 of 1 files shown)


srambach (approved):
Looks great!

thatblindgeye (approved):
This also looks good when linked with React locally 😍 If detached expandable sections are fine to have the animation always be "slide in from the top down" then 🚢

mcoker:
@thatblindgeye opened this issue for detached - do you remember if we landed on a solution for that? https://github.com/patternfly/patternfly/issues/7515

PR #7513, opened 2025-05-07
fix(accordion):  transition display instead of hidden attr

### LINKED ISSUE #7512: Accordion - toggle display instead of hidden
As part of the expandable animation, let's gauge the level of effort and potential risks of animating `display` instead of relying on `[hidden]`.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7512


thatblindgeye (approved):
Linked to React locally and things look good. There's still that visual thing where clicking open one toggle when another toggle is already open has a slight "snap", but if that's still fine from design per convo yesterday then React should be good without any further work (worth double checking in a core bump PR, though)

srambach (approved):
Looks good

PR #7525, opened 2025-05-13
fix(tree-view): animate visibility

### LINKED ISSUE #7524: Tree view - animate display
Update component to animate display instead of relying on `hidden` to hide/show

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7524

### DIFF
--- a/src/patternfly/components/TreeView/tree-view.scss
+++ b/src/patternfly/components/TreeView/tree-view.scss
@@ -357,12 +357,13 @@ $pf-v6-c-tree-view--MaxNesting: 10 !default;
 
 .#{$tree-view}__list-item {
   .#{$tree-view}__list {
-    display: none;
+    max-height: 0;
+    visibility: hidden;
     opacity: var(--#{$tree-view}__list--Opacity);
+    transition-delay: 0s, 0s, var(--#{$tree-view}__list--TransitionDuration--fade), var(--#{$tree-view}__list--TransitionDuration--fade);
     transition-timing-function: var(--#{$tree-view}__list--TransitionTimingFunction);
     transition-duration: var(--#{$tree-view}__list--TransitionDuration--fade), var(--#{$tree-view}__list--TransitionDuration--slide);
-    transition-property: opacity, translate, display;
-    transition-behavior: allow-discrete;
+    transition-property: opacity, translate, visibility, max-height;
     translate: 0 var(--#{$tree-view}__list--TranslateY);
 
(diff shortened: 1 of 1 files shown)


thatblindgeye (commented):
Similar comment here as the dual list PR re: could we add a temporary class to disable animations for the current, dynamically rendered children behavior?

mcoker:
Validated this works as expected in react when the nested list is always rendered (instead of conditionally) with this change

```diff
diff --git a/packages/react-core/src/components/TreeView/TreeViewListItem.tsx b/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
index 4db1294333..5fd54c267d 100644
--- a/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
+++ b/packages/react-core/src/components/TreeView/TreeViewListItem.tsx
@@ -249,7 +249,7 @@ const TreeViewListItemBase: React.FunctionComponent<TreeViewListItemProps> = ({
         </GenerateId>
         {action && <div className={css(styles.treeViewAction)}>{action}</div>}
       </div>
-      {internalIsExpanded && children}
+      {children}
     </li>
   );
 };
```

srambach (approved):
Looks great. Under 20x load I see a little delay for the space to close up, but the animation is smooth.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -357,11 +357,18 @@ $pf-v6-c-tree-view--MaxNesting: 10 !default;
 
 .#{$tree-view}__list-item {
   .#{$tree-view}__list {
+    max-height: 0;
+    visibility: hidden;
     opacity: var(--#{$tree-view}__list--Opacity);
+    transition-delay: 0s, 0s, var(--#{$tree-view}__list--TransitionDuration--fade), var(--#{$tree-view}__list--TransitionDuration--fade);
```
sg00dwin, line 363: If on close, the `opacity` and `translate` are completed at `100ms` is a delay needed for `visibility` and `max-height`?  When setting them all to `0s` in Chrome dev tools, it looks to function normally.

thatblindgeye (approved):
Looks to still work fine in React. Would have to investigate adding inert (doesn't seem to work as seemlessly as the DLS PR), but also wondering if we'd need it since everything has its tabindex updated automatically between -1 and 0.

sg00dwin (approved):
lgtm

PR #7527, opened 2025-05-13
fix(table): use correct th font weight in grid view

### LINKED ISSUE #7526: Bug - Table - header has too high font-weight on mobile layout
**Describe the problem**
This is an issue with  
- [ ] Patternfly 6  

The Table component in a mobile layout uses `font-weight: bold` which is a higher value than the usual weight used for bold fonts in the rest of Patternfly 6. The relevant part of the code which causes this should be here:
https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Table/table-grid.scss#L275

**Screenshots**

Image

**Expected behavior**
Expected font-weight of the Version/Status/Time/Branch would be the same as "cockpit-base.1" text.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7526

I poked around and looks like this is the only font-weight set to the "bold" keyword


srambach (approved):
**Fabulous!**

PR #7531, opened 2025-05-14
fix(expandable-section): animate transition instead of display

### LINKED ISSUE #7530: Expandable section - keep animation from firing on page load

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7530


mcoker:
Closing this PR in favor of https://github.com/patternfly/patternfly/pull/7493

Issue #7348, opened 2025-02-14
Bug - Toolbar - Clear all filters button below the filters


nicolethoen:
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

In PF6, the 'clear all filters' button is being displayed on its own line. I know it's been discussed amongst the team, but I would like to argue that it's a regression - possibly a result of cleaning up layouts on mobile viewport widths?

OCP + PF6 migration flagged this as an unexpected update or possible bug during their PF6 migration. It feels like this change is causing the toolbar to take up a lot of extra vertical real estate above tables. Could we make it so that at wider viewport widths, the button appears beside the applied filters?

CC: @lboehling

<img width="862" alt="Image" src="https://github.com/user-attachments/assets/826f6c84-a534-4647-9571-f79c962ff3ca" />

This is what it looked like in PF5

<img width="839" alt="Image" src="https://github.com/user-attachments/assets/de6d1aec-8e51-42a1-b35a-d0b3612b0ecd" />

sg00dwin:
Some investigative findings. In V5 there was a modifier class `pf-m-chip-container` added to `pf-v5-c-toolbar__content` and rendered the chip group of "clear all filters" link with `display:flex`

<img width="967" alt="Image" src="https://github.com/user-attachments/assets/aa2a306e-dcde-43ee-8fdc-a68cd1415472" />

----

<img width="863" alt="Image" src="https://github.com/user-attachments/assets/25f9c6d5-bfc0-4f97-8bf2-e426260ce37b" />

An alternative fix would be to wrap the chips and clear filter text with `pf-v6-c-toolbar__content-section`

PR #6987, opened 2024-08-19
fix(toolbar-group): let label groups wrap within the toolbar

srambach:
Allows label groups to wrap within the header rather than flexing them onto one line and having the labels within wrap down excessively.

Here are some examples of the new wrapping with longer label groups
Before:

After:

fixes #6967

PR #7532, opened 2025-05-15
fix(toolbar): align toolbar actions with filtered items

sg00dwin:
address https://github.com/patternfly/patternfly/issues/7348

Moves toolbar action links to be inline with the filtered items.

Requires follow on changes to add `pf-m-toolbar-filtered-items` here. https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Toolbar/ToolbarLabelGroupContent.tsx#L76

https://patternfly-pr-7532.surge.sh/components/toolbar/html-demos#toolbar-attribute-value-checkbox-menu-toggle-filter-on-desktop

note: in the above html-demo there is wider spacing between the filter label groups and the action links, which is intended, because of [this change](https://github.com/patternfly/patternfly/pull/6987) to maintain long label groups presented inline and wrap when needed. The links will be spaced closer if there is only a single label group.
Screenshot 2025-05-16 at 9 38 00 AM

### DIFF
--- a/src/patternfly/components/Toolbar/toolbar.scss
+++ b/src/patternfly/components/Toolbar/toolbar.scss
@@ -110,6 +110,15 @@ $pf-v6--include-toolbar-breakpoints: true !default;
   row-gap: var(--#{$toolbar}__content--RowGap);
   padding-inline-start: var(--#{$toolbar}__content--PaddingInlineStart);
   padding-inline-end: var(--#{$toolbar}__content--PaddingInlineEnd);
+
+  &.pf-m-toolbar-filtered-items {
+    display: flex;
+    align-items: baseline;
+
+    > :first-child {
+      margin-inline-end: var(--pf-t--global--spacer--md);
+    }
+  }
 }
 
 // Toolbar content section
@@ -239,7 +248,6 @@ $pf-v6--include-toolbar-breakpoints: true !default;
 
   // - Toolbar label group
   &.pf-m-label-group {
-    flex: 1;
     flex-wrap: wrap;
     column-gap: var(--#{$toolbar}__group--m-label-group--ColumnGap);
   }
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -110,6 +110,15 @@ $pf-v6--include-toolbar-breakpoints: true !default;
   row-gap: var(--#{$toolbar}__content--RowGap);
   padding-inline-start: var(--#{$toolbar}__content--PaddingInlineStart);
   padding-inline-end: var(--#{$toolbar}__content--PaddingInlineEnd);
+
+  &.pf-m-toolbar-filtered-items {
+    display: flex;
+    align-items: baseline;
+
+    > :first-child {
```
srambach, line 118: I'm wondering if we'd ever expect more than one toolbar group? Then they wouldn't get spaced correctly.
image

Also, can you add this new class to the documentation section?

mcoker (commented):
I think you called this out in the PR description, but this is what it looks like for me by default

Screenshot 2025-05-29 at 5 08 08 PM

Looks like that, and @srambach's [comment](https://github.com/patternfly/patternfly/pull/7532/files#r2114507797) are because `.pf-m-toolbar-filtered-items` doesn't wrap. The PF5 filter group wrapped - https://v5-archive.patternfly.org/components/toolbar/react/with-filters/

Here's what it looks like if it's set to wrap

https://github.com/user-attachments/assets/fa9a6605-2c9d-44a3-9d2d-5d1e91060d20

The only thing is if there are 2+ rows of filters, the clear all buttons wrap below all of them on their own line, but it was like that in PF5, too

Here's this PR with multiple rows and it set to wrap
foo

And PF5
Screenshot 2025-05-29 at 5 11 54 PM

Any reason we wouldn't add `flex-wrap: wrap` to `.pf-m-toolbar-filtered-items`?

Also we'll need to validate with a react dev that they can add `.pf-m-toolbar-filtered-items` to that `<ToolbarContent>`. At first glance, looks like we generate it internally for the filter groups, so I'm guessing it's possible, but we'll need to verify. cc @kmcfaul  - can you verify if we're able to add a class to this `.pf-v6-c-toolbar__content` element that wraps filter labels? 

Screenshot 2025-05-29 at 5 17 28 PM

PR #7537, opened 2025-05-20
fix(table): updated expandable to animate visibility

### LINKED ISSUE #7536: Table - update expandable to animate visibility
Currently animates `display` - update to animate `visibility`

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7536

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -774,6 +774,136 @@ Note: Table column widths will respond automatically when toggling expanded rows
 {{/table}}
 ```
 
+### Animated expandable
+```hbs
+{{#> table table--IsAnimateExpand=true table--id="table-animated-expandable" table--IsGrid=true table--modifier="pf-m-grid-lg" table--IsExpandable=true table--attribute='aria-label="Animated expandable table example"'}}
+  {{#> table-thead}}
+    {{#> table-tr table-tr--index="thead" table-tr--index="thead"}}
+      {{> table-cell-toggle table-tr--aria-controls="table-animated-expandable-content-1 table-animated-expandable-content-2 table-animated-expandable-content-3 table-animated-expandable-content-4"}}
+      {{> table-cell-check}}
+      {{#> table-th table-th--attribute='scope="col"' table-th--IsSortable=true table-th--modifier="pf-m-width-30" table-th--IsSelected="true" table-th--IsAsc="true"}}
+        Repositories
+      {{/table-th}}
+      {{#> table-th table-th--attribute='scope="col"' table-th--IsSortable=true}}
+        Branches
+      {{/table-th}}
+      {{#> table-th table-th--attribute='scope="col"' table-th--IsSortable=true}}
+        Pull requests
+      {{/table-th}}
+      {{> table-cell-empty table-cell--text='Links'}}
+      {{> table-cell-empty}}
+    {{/table-tr}}
+  {{/table-thead}}
+
+  {{#> table-tbody table-tr--index="1" table-tbody--modifier="pf-m-expanded"}}
+    {{#> table-tr table-tr--IsExpanded=true}}
+      {{> table-cell-toggle}}
+      {{> table-cell-check}}
+      {{> table--node}}
+      {{#> table-td table-td--data-label="Branches"}}
+        10
+      {{/table-td}}
+      {{#> table-td table-td--data-label="Pull requests"}}
+        25
+      {{/table-td}}
+      {{#> table-td table-td--data-label="Action"}}
+        <a href="#">Link 1</a>
+      {{/table-td}}
+      {{> table-cell-action}}
+    {{/table-tr}}
+
+    {{#> table-tr table-tr--IsExpandable=true table-tr--IsExpanded=true}}
+      {{> table-cell-empty}}
+      {{> table-cell-empty}}
+      {{#> table-td table-td--attribute=(concat 'colspan="4" id="' table--id '-content-' table-tr--index '"')}}
+        {{#> table-expandable-row-content}}
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+        {{/table-expandable-row-content}}
+      {{/table-td}}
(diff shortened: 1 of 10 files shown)


thatblindgeye (commented):
Tested this against the React PR that resolves the divider issue https://github.com/patternfly/patternfly-react/pull/11815 and it mostly looks good. The one thing I'm noticing is when you click to collapse a table row, the divider line appears as the expandable row is collapsing (causing the table to look like it has an empty row momentarily). I think it's really only noticeable at 25 or 10% animation speed, though.

My mac isn't currently letting my screen record so pictures instead:

PatternFly table example showing how expandable animations causes the table to momentarily look like it has an empty table row as the expandable row is animating out.

srambach:
> Tested this against the React PR that resolves the divider issue [patternfly/patternfly-react#11815](https://github.com/patternfly/patternfly-react/pull/11815) and it mostly looks good. The one thing I'm noticing is when you click to collapse a table row, the divider line appears as the expandable row is collapsing (causing the table to look like it has an empty row momentarily). I think it's really only noticeable at 25 or 10% animation speed, though.

Does it have to do with the `border-block-end: revert` on line 1051 and/or not everything is getting marked as expanded as it should?
There are 3 elements with `.pf-m-expanded` and if you toggle the `table__tr` that's not the expandable row, it creates an extra border

image

srambach:
There's a problem with the [compound expansion example](https://patternfly-pr-7537.surge.sh/components/table/#compound-expansion-example) - looks like maybe changing them to visibility instead of display:none means they are taking space.

thatblindgeye (approved):
Other than ^ lgtm, animations work nicely when pulled into React (with the fixes for the divider bug that got merged in react)

mcoker:
@srambach can you check again?

I'm assuming we should animate compound expandable, too? At least the initial click on a compound expandable item, but also possibly a fade between items once one is expanded when clicking on a different item. cc @andrew-ronaldson wdyt? Here's a link to compound expandable in react - https://www.patternfly.org/components/table/react/compound-expandable/

sg00dwin:
@mcoker I'm not seeing the `/components/table#tree-table` animate on expand/collapse

sg00dwin (approved):
LGTM

kmcfaul (approved):

src/patternfly/components/Table/examples/Table.md
```diff
@@ -1344,8 +1362,10 @@ Note: To apply padding to `.pf-v6-c-table__expandable-row`, wrap the content in
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-v6-c-table__control-row` | `.pf-v6-c-table__expandable > <tr>` | Modifies a compound expandable table control row. |
-| `.pf-m-expanded` | `<tbody>`, `.pf-v6-c-table__compound-expansion-toggle` > `.pf-v6-c-button` | Modifies a tbody with a row and an expandable row. |
 | `.pf-v6-c-table__compound-expansion-toggle` | `<td>` | Modifies a `<td>` on active/focus. |
+| `.pf-m-expanded` | `<tbody>`, `.pf-v6-c-table__compound-expansion-toggle` > `.pf-v6-c-button` | Modifies a tbody with a row and an expandable row. |
```
srambach: This wording is confusing - modifies a tbody that contains ... and an expanded toggle (?)

src/patternfly/components/Table/examples/Table.md
```diff
@@ -1344,8 +1362,10 @@ Note: To apply padding to `.pf-v6-c-table__expandable-row`, wrap the content in
...
+| `.pf-m-no-background` | `.pf-v6-c-table__expandable-row-content` | Modifies the expandable row content to have a transparent background. For in compound expandable when the parent expandable row has no padding with `.pf-m-no-padding`. |
```
srambach, line 1678:
```suggestion
| `.pf-m-no-background` | `.pf-v6-c-table__expandable-row-content` | Modifies the expandable row content to have a transparent background. For use in compound expandable when the parent expandable row has no padding with `.pf-m-no-padding`. |
```

src/patternfly/components/Table/table.scss
```diff
@@ -1008,24 +1050,16 @@
...
-    translate: 0 var(--#{$table}__tbody--m-expanded__expandable-row--TranslateY);
-
-    @starting-style {
-      opacity: var(--#{$table}__expandable-row--Opacity);
-      translate: 0 var(--#{$table}__expandable-row--TranslateY);
-    }
-  }
-
-  &:not(.pf-m-expanded) {
-    display: none;
```
srambach: Does this need to remain or use another method to hide rows that aren't expanded?

srambach (commented):
It looks like compound expandable shows all the non-expanded rows unless `.pf-m-animate-expand` is on the table. Are we keeping `display: none` in that case so we don't break core users?

Also, I think we need to add `.pf-m-animate-expand` in the docs at the appropriate spot.

src/patternfly/components/Table/table.scss
mcoker: Added back, missed that completely. That hides the non-expanded rows in a non-expandable table.

src/patternfly/components/Table/examples/Table.md
mcoker: That was already there (I didn't add it) but I took a stab. It looks like the selectors were wrong, too.

mcoker:
@srambach thanks! I should have caught the non-animated rows showing!

Updated and added new examples with the animated class for regular and compound expandable.

thatblindgeye (changes requested):
Repeat from Slack convo below, but a couple things I'm noticing when pulling these updates into React with updates made there:

- In Core we need to apply the `pf-m-animate-expand` class to the outer table element. Doing that will show an issue seen in React where when anything but the 1st compound expandable section is expanded, there's a border above the expandable content. I assume it's due to some `tr + tr` selector not taking into account this animate-expand modifier? The core compound expandable looks fine when updating the classes so the second compound expansion button+content is expanded, it's not until you add the missing `pf-m-animate-expand` to the outer table where you can reproduce the issue
- If content within the `pf-v6-c-table__expandable-row-content` element has a margin (as it does in React examples/demos for compound expansion), that margin continues to take up space when collapsed. Updating the React examples/demos to use the padding utility class instead of the margin one should resolve the issue, but just wondering if we want to handle a case of margin being used.

Not from Slack, but also examples will need to be updates so that `hidden` and `inert` are added to the collapsed rows.

srambach:
I think now the top padding is missing on tabs that aren't the first one?  You can see in the inspector that `padding-block-start` is 0 if it's not immediately following a control row

https://github.com/user-attachments/assets/ce3c9e42-39d0-45db-b4be-8251885395fb

thatblindgeye (approved):
Looks good when linked to my React draft!

srambach (approved):
I'm good with a follow-on later to see if we can improve the bounce when switching from one compound expandable to another.

PR #7539, opened 2025-05-20
fix(accordion): animate visibility

### LINKED ISSUE #7538: Accordion - update to animate visibility
Currently animating `display`, update to animate `visibility` instead

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7538

### DIFF
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -213,7 +213,6 @@
 
 .#{$accordion}__expandable-content {
   max-height: 0;
-  margin-block-end: var(--#{$accordion}__expandable-content--MarginBlockEnd);
   margin-inline-start: var(--#{$accordion}__expandable-content--MarginInlineStart);
   margin-inline-end: var(--#{$accordion}__expandable-content--MarginInlineEnd);
   font-size: var(--#{$accordion}__expandable-content--FontSize);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Accordion/accordion.scss
```diff
@@ -212,32 +212,30 @@
...
-    overflow-y: auto;
-  }
-
   .#{$accordion}__item.pf-m-expanded & {
-    display: revert;
-
-    @starting-style {
-      --#{$accordion}__expandable-content--Opacity: 0;
-      --#{$accordion}__expandable-content--TranslateY: -.5rem;
+    max-height: 9999px;
```
thatblindgeye, line 235: Rather than setting an arbitrary max-height, would we be able to just unset the `max-height: 0` declaration?

thatblindgeye (commented):
Noticing in React for a fixed implementation, if the accordion item content has a scrollbar, the overflow is removed as the collapsing animation occurs. Adding overflow-y of auto to `.pf-v6-c-accordion__expandable-content` seems to help.

Basically looks like we're only applying the overflow auto when the content is expanded, causing the overflow content to break out of its container when the collapse animation is occuring. 

Video of the behavior I'm seeing just pulling into React, no react changes made:

https://github.com/user-attachments/assets/66d65171-c210-4f5b-95c2-c308c1c6b33b

thatblindgeye (approved):
lgtm, my other comment above about max-height isnt blocking

src/patternfly/components/Accordion/accordion.scss
thatblindgeye: Per convo with Coker we need a px value to transition correctly

srambach (approved):
I also see a lag at 20x CPU loading here.
Fine to merge with that imho.

PR #7542, opened 2025-05-21
fix(nav): animate slide

### LINKED ISSUE #7518: Animation - Nav - use improved scheme for expansion
A [previous spike](https://github.com/patternfly/patternfly/issues/7433) investigated performance issues with the way we were animating the expansion of navigation items. Since then we've started to use a small translate to hint at a slide in, along with an opacity change. Currently we are just using opacity on the nav menu items.

**Describe the feature**
Expandable nav menus should use the same performant method of a slight vertical translate plus opacity to animate expansion and collapse.

mcoker:
fixes #7518


srambach (approved):
🛝 🧭 
I don't see any delay in closing up at 20x cpu load on this one.

PR #7545, opened 2025-05-22
feat(progress): animate progress status jiggle

### LINKED ISSUE #7342: Animation: Form control (TextInput, TextArea, FormSelect) add error validation jiggle

### LINKED ISSUE #7343: Animation: Text input groups add error validation jiggle

### LINKED ISSUE #7516: Progress: Add error jiggle on failure.
Follow up to design issue https://github.com/patternfly/patternfly-design/issues/1353

When the progress bar fails to complete its task I'd like to trigger the same error jiggle we use on other components for validation.

Image

https://github.com/user-attachments/assets/a37d7cb4-d77e-457e-9451-f32ac5e04963

sg00dwin:
Feature https://github.com/patternfly/patternfly/issues/7516

Builds off of pr https://github.com/patternfly/patternfly/pull/7473 which introduces `mixin`s for `pf-v6-animate-danger-jiggle` and `pf-v6-fade-default`

### DIFF
--- a/src/patternfly/base/patternfly-common.scss
+++ b/src/patternfly/base/patternfly-common.scss
@@ -67,3 +67,48 @@
 
   // stylelink-enable declaration-no-important
 }
+
+:root {
+  --#{$pf-global}--danger-jiggle--AnimationDuration--Transform: var(--pf-t--global--motion--duration--fade--default);
+  --#{$pf-global}--danger-jiggle--AnimationTimingFunction--Transform: var(--pf-t--global--motion--timing-function--default);
+}
+
+// Register the property type for the custom property to be animatable
+@property  --#{$pf-global}--danger-jiggle--TranslateX {
+  syntax: "<length>";
+  inherits: false;
+  initial-value: 0;
+}
+
+// Animate danger jiggle
+@keyframes #{$pf-global}-danger-jiggle-motion {
+  33% {
+    --#{$pf-global}--danger-jiggle--TranslateX: -2px;
+  }
+  
+  66% {
+    --#{$pf-global}--danger-jiggle--TranslateX: 3px;
+  }
+}
+
+// Animate fade-in
+@keyframes #{$pf-global}-fade-in {
+  from {
+    opacity: 0;
+  }
+  
+  to {
+    opacity: 1;
+  }
+}
+
+// Animate fade-out
+@keyframes #{$pf-global}-fade-out {
+  from {
+    opacity: 1;
+  }
+  
+  to {
+    opacity: 0;
+  }
+}
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -294,7 +294,7 @@
       @include pf-v6-animate-danger-jiggle;
     }
 
-    .pf-m-status {
+    .#{$form-control}__icon.pf-m-status {
       @include pf-v6-fade-default(#{$form-control}, pf-v6-global-fade-in);
     }
 
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -290,7 +290,13 @@
     --#{$form-control}__icon--m-status--Color: var(--#{$form-control}--m-error__icon--m-status--Color);
     --#{$form-control}--after--BorderWidth: var(--#{$form-control}--m-error--after--BorderWidth);
 
-    @include pf-v6-animate-danger-jiggle("#{$form-control}");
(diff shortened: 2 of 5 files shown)


src/patternfly/base/patternfly-common.scss
```diff
@@ -66,4 +66,49 @@
...
+}
+
+// Animate danger jiggle
+@keyframes #{$pf-global}-danger-jiggle-motion {
+  33% {
+    --#{$pf-global}--danger-jiggle--TranslateX: -2px;
+  }
+  
+  66% {
+    --#{$pf-global}--danger-jiggle--TranslateX: 3px;
```
srambach, line 90: I am not personally worried about RTL-ing these but will call it out in case anyone else is 😛

srambach (approved):
One question but I'm good with it as is.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -290,6 +290,14 @@
     --#{$form-control}__icon--m-status--Color: var(--#{$form-control}--m-error__icon--m-status--Color);
     --#{$form-control}--after--BorderWidth: var(--#{$form-control}--m-error--after--BorderWidth);
 
+    @media (prefers-reduced-motion: no-preference) {
+      @include pf-v6-animate-danger-jiggle;
+    }
+
+    .pf-m-status {
```
mcoker: This isn't as big of a deal with form control because it doesn't typically take nested elements, but this will target any `.pf-m-status` class. We typically do not style modifier classes on their own, and instead style the BEM block/element class with the modifier class. So like `.pf-v6-c-form-control__icon.pf-m-status`

Where this could cause a problem is if someone passed a [custom icon](https://github.com/patternfly/patternfly-react/blob/9d843f99842f39a8233c16c23c001aa22647dd82/packages/react-core/src/components/TextInput/examples/TextInputCustomIcon.tsx#L15) that had `.pf-m-status`, we'd end up styling that, too.

src/patternfly/components/Progress/progress.scss
```diff
@@ -140,6 +140,16 @@
     --#{$progress}__indicator--BackgroundColor: var(--#{$progress}--m-danger__indicator--BackgroundColor);
     --#{$progress}__status-icon--Color: var(--#{$progress}--m-danger__status-icon--Color);
     --#{$progress}--m-inside__measure--Color: var(--#{$progress}--m-danger--m-inside__measure--Color);
+
+    .#{$progress}__bar {
+      @media (prefers-reduced-motion: no-preference) {
+        @include pf-v6-animate-danger-jiggle;
+      }
+
+      .pf-m-status {
```
mcoker: Same comment here about styling a block/element class along with the modifier so we don't accidentally target elements we don't intend to.

src/patternfly/components/TextInputGroup/text-input-group.scss
```diff
@@ -128,6 +128,14 @@
     --#{$text-input-group}--BorderColor: var(--#{$text-input-group}--m-error--BorderColor);
     --#{$text-input-group}--m-hover--BorderColor: var(--#{$text-input-group}--m-hover--m-error--BorderColor);
     --#{$text-input-group}__icon--m-status--Color: var(--#{$text-input-group}__main--m-error__icon--m-status--Color);  
+    
+    @media (prefers-reduced-motion: no-preference) {
+      @include pf-v6-animate-danger-jiggle;
+    }
+    
+    .pf-m-status {
```
mcoker: Same comment here about styling a block/element class along with the modifier so we don't accidentally target elements we don't intend to.

mcoker (changes requested):
Left some comments about standalone modifier classes being styled when they should be attached to block/element selectors.

Also @srambach is going to confirm that the fade/jiggle is OK to run on page load.

src/patternfly/components/Progress/progress.scss
mcoker: Looks like this is meant to target the `__status-icon` element

```suggestion
      .#{$progress}__status-icon {
```

src/patternfly/components/FormControl/form-control.scss
mcoker:
```suggestion
    .#{$form-control}__icon.pf-m-status {
```

src/patternfly/components/TextInputGroup/text-input-group.scss
mcoker:
```suggestion
    .#{$text-input-group}__icon.pf-m-status {
```

mcoker (approved):
shake 'n bake!!

PR #7550, opened 2025-05-28
fix(ExpandableSection): made animation opt-in for detached variant

### LINKED ISSUE #7515: Bug - Expandable section - detached animation direction
Currently the content expands "down" by default. In a detached expandable section, the content can be on any side of the toggle, so we need a way to have the content expand "up".

Chatted with @andrew-ronaldson and we want to disable default animation for detached toggles, and create an opt-in class for the direction, similar to the toggle's `.pf-m-expand-top` class.

thatblindgeye:
Closes #7515 

Some notes:

- ~In React I believe we're fine to *add* things outside of breaking changes -- wasn't sure if we wanted to apply the `pf-m-detached` modifier to the `pf-v6-c-expandable-section` wrapper that only contains the toggle, but I did that here which should be fine to add into React + sorta seemed odd that only one of the wrapper would have that modifier~ Per discussion in retroplanning 6/4/2025, React won't have the detached class as it won't be needed for animations. Right now it's kept in Core though not marked as required; the plan is in the next breaking change to have those classes applied+required for detached variants
- Updated docs to mention the new method of changing the toggle icon rotation direction
- Removes the fade and slide animation for detached variants if neither the top or bottom modifiers are included

### DIFF
--- a/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
+++ b/src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
@@ -128,10 +128,10 @@ cssPrefix: pf-v6-c-expandable-section
 | `.pf-v6-c-expandable-section__content` | `<div>` | Initiates the expandable section content. **Required** |
 | `.pf-m-expanded` | `.pf-v6-c-expandable-section` | Modifies the component for the expanded state. |
 | `.pf-m-detached` | `.pf-v6-c-expandable-section` | Modifies the component for a detached variant. |
+| `.pf-m-expand-top` | `.pf-v6-c-expandable-section__toggle-icon` | Modifies the toggle icon to point up when expanded. We recommend the new method of applying this class directly to the `.pf-v6-c-expandable-section` wrapper element. |
 | `.pf-m-expand-top` | `.pf-v6-c-expandable-section.pf-m-detached` | Modifies the expandable animation and icon rotation directions for detached expandable sections. **Required** when the content is above the toggle. |
 | `.pf-m-expand-bottom` | `.pf-v6-c-expandable-section.pf-m-detached` | Modifies the expandable animation direction for detached expandable sections. **Required** when the content is below the toggle. |
 | `.pf-m-display-lg` | `.pf-v6-c-expandable-section` | Modifies the styling of the component to have large display styling. |
 | `.pf-m-indented` | `.pf-v6-c-expandable-section` | Indicates that the expandable section content is indented and is aligned with the start of the title text to provide visual hierarchy. |
 | `.pf-m-truncate` | `.pf-v6-c-expandable-section` | Indicates that the expandable section content is truncated by default, and not truncated when expanded. |
-| `.pf-m-expand-top` | `.pf-v6-c-expandable-section__toggle-icon` | Modifies the toggle icon to point up when expanded. We recommend the new method of applying this class directly to the `.pf-v6-c-expandable-section` wrapper element. |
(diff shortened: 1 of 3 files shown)


thatblindgeye:
Can also test how it works in React by using this draft PR https://github.com/patternfly/patternfly-react/pull/11851

src/patternfly/components/ExpandableSection/expandable-section.scss
```diff
@@ -122,7 +146,7 @@
   transition: var(--#{$expandable-section}__toggle-icon--Transition); // TODO remove shorthand in breaking change
   transform: rotate(var(--#{$expandable-section}__toggle-icon--Rotate));
 
-  &.pf-m-expand-top {
+  &.pf-m-expand-top { // TODO: Remove this block in breaking change in favor of using modifier on outter expandable section wrapper
```
srambach:
```suggestion
  &.pf-m-expand-top { // TODO: Remove this block in breaking change in favor of using modifier on outer expandable section wrapper
```
😛 🦦

src/patternfly/components/ExpandableSection/expandable-section.scss
```diff
@@ -81,6 +85,26 @@
     gap: var(--pf-v6-c-expandable-section--Gap);
   }
 
+  &.pf-m-expand-top {
+    --#{$expandable-section}__toggle-icon--Rotate: var(--#{$expandable-section}__toggle-icon--m-expand-top--Rotate);
+  }
+
+  &:has(.#{$expandable-section}__content:only-child) {
```
srambach, line 95: Can you add a comment here about the assumption we're making about the structure with the only-child?

src/patternfly/components/ExpandableSection/examples/ExpandableSection.md
```diff
@@ -127,8 +127,11 @@ cssPrefix: pf-v6-c-expandable-section
...
+| `.pf-m-expand-bottom` | `.pf-v6-c-expandable-section.pf-m-detached` | Modifies the expandable animation direction for detached expandable sections. **Required** when the content is below the toggle. |
 | `.pf-m-display-lg` | `.pf-v6-c-expandable-section` | Modifies the styling of the component to have large display styling. |
 | `.pf-m-indented` | `.pf-v6-c-expandable-section` | Indicates that the expandable section content is indented and is aligned with the start of the title text to provide visual hierarchy. |
 | `.pf-m-truncate` | `.pf-v6-c-expandable-section` | Indicates that the expandable section content is truncated by default, and not truncated when expanded. |
-| `.pf-m-expand-top` | `.pf-v6-c-expandable-section__toggle-icon` | Modifies the toggle icon to point up when expanded. |
+| `.pf-m-expand-top` | `.pf-v6-c-expandable-section__toggle-icon` | Modifies the toggle icon to point up when expanded. We recommend the new method of applying this class directly to the `.pf-v6-c-expandable-section` wrapper element. |
```
srambach: wdyt about moving this adjacent to the other `.pf-m-expand-top` so that a reader would see them together? I just wonder if it could be confusing if one is missed.

srambach (approved):
Just minor nits in docs. Looks good to me otterwise. 🦦

src/patternfly/components/ExpandableSection/expandable-section.scss
thatblindgeye: You and Katie calling me outt 😭 (that misspelling was intentional at least)

It's funny because I think I told Katie the same thing about maybe having otter's on the mind 😆

srambach (approved):
Super, thank you!

andrew-ronaldson (approved):

mcoker (approved):
👏

PR #7552, opened 2025-05-28
fix(menu): respect reduced motion for drilldown menu

### LINKED ISSUE #6739: Menu - make drill down menu respect prefers-reduced-motion

srambach:
Fixes #6739 

Adds support for the prefers-reduced-motion media query to the Menu component drilldown menu.

Changes:
Set default transition duration to 0s for all transitions
Apply standard transition durations only for users who don't prefer reduced motion
Instead of removing transitions, I had to set the durations to 0s; otherwise the menu did not behave properly.
Setting the overall menu transition duration was necessary or else the nested menus would still have a transition duration and show the translate happening.
That said, the 0s duration does not seem to affect the fade in of other menus - that seems like it might still be coming from Popper despite the `!important` declaration.


srambach:
Messed up branch; closing

PR #7553, opened 2025-05-28
fix(menu): respect prefers reduced motion for drilldown

### LINKED ISSUE #6739: Menu - make drill down menu respect prefers-reduced-motion

srambach:
Fixes https://github.com/patternfly/patternfly/issues/6739

Adds support for the prefers-reduced-motion media query to the Menu component drilldown menu.

Changes:
Set default transition duration to 0s for all transitions
Apply standard transition durations only for users who don't prefer reduced motion
Instead of removing transitions, I had to set the durations to 0s; otherwise the menu did not behave properly.
Setting the overall menu transition duration was necessary or else the nested menus would still have a transition duration and show the translate happening.
That said, the 0s duration does not seem to affect the fade in of other menus - that seems like it might still be coming from Popper despite the !important declaration.

@andrew-ronaldson I left the transition on the height - do you think it should be removed as well?

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -15,8 +15,12 @@
   --#{$menu}--ZIndex: var(--pf-t--global--z-index--sm);
   --#{$menu}--button--disabled--Color: var(--pf-t--global--text--color--disabled);
   --#{$menu}--icon--disabled--Color: var(--pf-t--global--icon--color--disabled);
-  --#{$menu}--TransitionDuration: var(--pf-t--global--motion--duration--fade--default);
+  --#{$menu}--TransitionDuration: 0s;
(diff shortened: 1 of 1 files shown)


sg00dwin (approved):
LGTM!

srambach:
@andrew-ronaldson 
With reduced motion:

https://github.com/user-attachments/assets/d565f273-c493-4fb5-b1d8-494d2272780f

srambach:
With full motion:

https://github.com/user-attachments/assets/eba962d9-6f02-451a-b69e-5129b0c550c8

andrew-ronaldson (approved):
perfecto

mcoker (approved):
🚀

PR #7563, opened 2025-06-09
fix(button): fix hamburger shift on hover

### LINKED ISSUE #7562: Bug - Button - hamburger icon shifts on hover
Expand icon shifts on hover, collapse icon shifts on hover if page is zoomed in.

mcoker:
Was using `scale: -1` which was also flipping on the y-axis causing the shift. Separated scaleX and scaleY and am only changing scaleX.

fixes https://github.com/patternfly/patternfly/issues/7562


thatblindgeye (approved):
On the one hand the icon looks less happy without it's little jiggle dance 🕺🏼 , but on the hand it is less jarring so in that regard it LGTM

srambach (approved):
L 🍔 TM

Issue #7543, opened 2025-05-21
Bug - Data list - draggable buttons are too big in compact data lists


mcoker:
Reported in https://issues.redhat.com/browse/AAP-42989

You can see the misalignment in https://www.patternfly.org/components/drag-and-drop/react-demos/drag-and-drop-sortable-data-list/

The compact data list in that demo is using regular sized buttons, and compact data lists should use small buttons/toggles. You can see examples of small buttons/toggles in these examples:
* https://www.patternfly.org/components/data-list/html/compact/
* https://www.patternfly.org/components/data-list/html/expandable-compact/

PR #7564, opened 2025-06-09
fix(data-list): adjust padding to align drag icon with text

sg00dwin:
https://github.com/patternfly/patternfly/issues/7543


srambach (approved):
Looks fine to me. @mcoker 's comment in the issue sounds like it should be a .pf-m-small button, but it looks like draggable buttons are styled specifically. So unless there's something I'm missing, this looks good!

mcoker (approved):
LGTM!

PR #7570, opened 2025-06-11
chore(backstop): use pf-m-no-motion, table screenshots

mcoker:
Removes our custom css that disables animations and uses `.pf-m-no-motion` instead. Also runs backstop against table since the animation PR merged.

https://drive.google.com/file/d/1Dj5ji_3ocGJ7Qf4CVLIzedcHV4SwDU6t/view?usp=sharing


sg00dwin (approved):
lgtm

PR #7572, opened 2025-06-12
fix(form): fix field group padding, toggle alignement, demo code

### LINKED ISSUE #7571: Bug - Forms - spacing is off when field groups are used
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Per comment and screenshot from https://github.com/patternfly/patternfly-react/pull/11843#pullrequestreview-2875204408, the spacing is a bit off when field groups are used (possibly when expandable + nonexpandable are used together?)

Visit the [field groups example](https://patternfly-react-pr-11843.surge.sh/components/forms/form#field-groups) from the above linked React PR, then go to the input that has the label "Nested field group 2 (non-expandable)", and notice the divider after the second input there has no top spacing to separate it from the input.

**Expected behavior**

**Screenshots**

**What is your environment?**

**Any other information?**

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7571

* Updates the core example to match react
* Fixes the gap/row spacing from #7571
* Fixes nested toggle alignment from screenshot below

Screenshot 2025-06-11 at 10 35 37 PM

### DIFF
--- a/src/patternfly/components/Form/form.scss
+++ b/src/patternfly/components/Form/form.scss
@@ -130,7 +130,7 @@ $pf-v6-c-form--m-horizontal--breakpoint-map: build-breakpoint-map("sm", "md", "l
   --#{$form}__field-group--GridTemplateColumns--toggle: calc(var(--pf-t--global--spacer--md) * 2 + var(--#{$form}__field-group-toggle-icon--MinWidth) + var(--pf-t--global--spacer--xs)); // based off of the expected width of the group toggle, for use to define a column when the toggle is not present
   --#{$form}__field-group-toggle--PaddingBlockStart: var(--#{$form}__field-group-header--PaddingBlockStart);
   --#{$form}__field-group-toggle--PaddingInlineEnd: var(--pf-t--global--spacer--xs);
-  --#{$form}__field-group__field-group__field-group-toggle--PaddingBlockStart: var(--pf-t--global--spacer--md);
+  --#{$form}__field-group__field-group__field-group-toggle--PaddingBlockStart: var(--pf-t--global--spacer--lg); // remove in breaking change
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Form/form.scss
```diff
@@ -130,7 +130,7 @@ $pf-v6-c-form--m-horizontal--breakpoint-map: build-breakpoint-map("sm", "md", "l
   --#{$form}__field-group--GridTemplateColumns--toggle: calc(var(--pf-t--global--spacer--md) * 2 + var(--#{$form}__field-group-toggle-icon--MinWidth) + var(--pf-t--global--spacer--xs)); // based off of the expected width of the group toggle, for use to define a column when the toggle is not present
   --#{$form}__field-group-toggle--PaddingBlockStart: var(--#{$form}__field-group-header--PaddingBlockStart);
   --#{$form}__field-group-toggle--PaddingInlineEnd: var(--pf-t--global--spacer--xs);
-  --#{$form}__field-group__field-group__field-group-toggle--PaddingBlockStart: var(--pf-t--global--spacer--md);
+  --#{$form}__field-group__field-group__field-group-toggle--PaddingBlockStart: var(--pf-t--global--spacer--lg); // remove in breaking change
```
thatblindgeye, line 133: Could we note in the comment (or open a followup issue that explains) why this should be removed in the breaking change? I don't think I noticed anything drastically change when turning this style off in dev tools, but was also hard to compare before/after because dev tools decides to jump scroll elsewhere when I toggle it

thatblindgeye (approved):
LGTM, just one comment below

srambach (approved):
Looks good and I like having the more robust example.

PR #7578, opened 2025-06-18
fix(table): fix comp exp jump, background fade on collapse

### LINKED ISSUE #7577: Bug - Table - compound expandable animation jumping
In a compound expandable table, animation should not occur when clicking between items in the control row. The animation should only occur when 1) no expansion panel is expanded and one is expanded, and 2) an expansion panel is open and it is closed.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7577

I also noticed that when you collapse the compound expandable row, the background doesn't fade out - that should be fixed now.


srambach (approved):
🛸 LGTM

thatblindgeye (approved):

PR #7582, opened 2025-06-26
fix(data-list): center align the drag icon within the row of core example

### LINKED ISSUE #7579: Bug - data-list - compact button for draggable element isn't spaced correctly
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  

With the changes merged in this recent [PR updates](https://github.com/patternfly/patternfly/pull/7564) to address draggable data list button alignment in the `React` component  [Bug - Data list - draggable buttons are too big in compact data lists](https://github.com/patternfly/patternfly/issues/7543), the `Core` `data-list` doc example will then not aligned properly. Due to the markup structure used in the `handlebars` example.

**Misalignment to be fixed** 
Image

sg00dwin:
fixes https://github.com/patternfly/patternfly/issues/7579

https://patternfly-pr-7582.surge.sh/components/data-list#draggable

### DIFF
--- a/src/patternfly/components/Button/button.hbs
+++ b/src/patternfly/components/Button/button.hbs
@@ -27,6 +27,7 @@
     button--IsBlock='pf-m-block'
     button--IsSettings='pf-m-settings'
     button--IsHamburger='pf-m-hamburger'
+    button--IsDraggable='pf-v6-c-data-list__item-draggable-button'
     button--modifier=button--modifier
   }}
   {{#if button--IsDisabled}}
--- a/src/patternfly/components/Button/button.hbs
+++ b/src/patternfly/components/Button/button.hbs
@@ -27,7 +27,6 @@
     button--IsBlock='pf-m-block'
     button--IsSettings='pf-m-settings'
     button--IsHamburger='pf-m-hamburger'
-    button--IsDraggable='pf-v6-c-data-list__item-draggable-button'
     button--modifier=button--modifier
   }}
   {{#if button--IsDisabled}}
--- a/src/patternfly/components/DataList/examples/DataList.md
+++ b/src/patternfly/components/DataList/examples/DataList.md
@@ -875,7 +875,7 @@ When a list item includes more than one block of content, it can be difficult fo
   {{#> data-list-item data-list-item--id="item-1"}}
     {{#> data-list-item-row}}
       {{#> data-list-item-control}}
-        {{> data-list-item-draggable-button data-list-item-draggable-button--IsDisabled=true data-list-item-draggable-button--attribute=(concat 'id="' data-list--id '-draggable-button-1" aria-describedby="draggable-help" aria-labelledby="' data-list--id '-draggable-button-1 ' data-list--id '-item-1" disabled')}}
+        {{> data-list-item-draggable-button data-list-item-draggable-button--IsDisabled=true data-list-item-draggable-button--attribute=(concat 'id="' data-list--id '-draggable-button-1" aria-describedby="draggable-help" aria-labelledby="' data-list--id '-draggable-button-1 ' data-list--id '-item-1"')}}
         {{> data-list-check}}
       {{/data-list-item-control}}
       {{#> data-list-item-content}}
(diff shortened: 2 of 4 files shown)


srambach (commented):
These changes look good to me. A couple of notes - 
- @mcoker does this qualify as a breaking change? We've added `pf-v6-c-button` and `pf-m-plain` which I think would be ok? But also removed `pf-v6-c-data-list__item-draggable-icon`
- if removing `pf-v6-c-data-list__item-draggable-icon` is ok, then I think we can remove the styling for it (just icon color) and remove it from the docs. Otherwise, I guess we need to add that back onto the icon span?

src/patternfly/components/Button/button.hbs
```diff
@@ -27,6 +27,7 @@
     button--IsBlock='pf-m-block'
     button--IsSettings='pf-m-settings'
     button--IsHamburger='pf-m-hamburger'
+    button--IsDraggable='pf-v6-c-data-list__item-draggable-button'
```
mcoker: This this is a class specific to data-list, can you take this out and move it to [data-list-item-draggable-button](https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/data-list-item-draggable-button.hbs)?

I would just remove this line, then in `data-list-item-draggable-button.hbs`, just pass `button--modifier=(concat pfv "data-list__item-draggable-button")`. My `concat` syntax may be wrong 😅 Just want to get the hard-coded `pf-v6-c-` part out of the string.

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -875,7 +875,7 @@ When a list item includes more than one block of content, it can be difficult fo
   {{#> data-list-item data-list-item--id="item-1"}}
     {{#> data-list-item-row}}
       {{#> data-list-item-control}}
-        {{> data-list-item-draggable-button data-list-item-draggable-button--modifier="pf-m-disabled" data-list-item-draggable-button--attribute=(concat 'id="' data-list--id '-draggable-button-1" aria-describedby="draggable-help" aria-labelledby="' data-list--id '-draggable-button-1 ' data-list--id '-item-1" disabled')}}
+        {{> data-list-item-draggable-button data-list-item-draggable-button--IsDisabled=true data-list-item-draggable-button--attribute=(concat 'id="' data-list--id '-draggable-button-1" aria-describedby="draggable-help" aria-labelledby="' data-list--id '-draggable-button-1 ' data-list--id '-item-1" disabled')}}
```
mcoker:
```suggestion
        {{> data-list-item-draggable-button data-list-item-draggable-button--IsDisabled=true data-list-item-draggable-button--attribute=(concat 'id="' data-list--id '-draggable-button-1" aria-describedby="draggable-help" aria-labelledby="' data-list--id '-draggable-button-1 ' data-list--id '-item-1"')}}
```

Just a nit, but you can remove the `disabled` attr here now that you're passing `button--IsDisabled` to the button.  That will add the `disabled` attr https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/Button/button.hbs#L73-L77

mcoker (changes requested):
Looks good, just a couple of updates please. I left a comment about removing the `disabled` attribute from an hbs block - we should also replace `disabled` with `data-list-item-draggable-button--IsDisabled=true` on this line, too https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/examples/DataList.md?plain=1#L910

Will answer @srambach's comments in another comment

mcoker:
@srambach good catches!

> @mcoker does this qualify as a breaking change? We've added pf-v6-c-button and pf-m-plain which I think would be ok? But also removed pf-v6-c-data-list__item-draggable-icon

The button classes are in the react component, so IMO that part of the update just fixes a bug in the core docs and shouldn't be breaking. However we do need to add `.pf-v6-c-data-list__item-draggable-icon` back to match the react component. And I think we should remove the styles that set the icon color, too, and just use the button's icon colors. Though we'll want to leave the vars in place, so if a user has set them, they'll continue to work as they did before.

I _think_ that update would be this:

* Change the values of these vars to `initial` so they'll trigger a fallback for where they're used:
  https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/data-list.scss#L78-L81
* In these 4 places, add `inherit` as a fallback for the declarations' value vars
  https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/data-list.scss#L319
  https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/data-list.scss#L299
  https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/data-list.scss#L305
  https://github.com/patternfly/patternfly/blob/cb3fa150afa1daf01f542f6a4d3d729b9cbf4293/src/patternfly/components/DataList/data-list.scss#L313

> if removing pf-v6-c-data-list__item-draggable-icon is ok, then I think we can remove the styling for it (just icon color) and remove it from the docs. Otherwise, I guess we need to add that back onto the icon span?

Yep! Need to add it back.

sg00dwin:
@mcoker Addressed comments

- add data-list-item-draggable-icon back
- remove/replace disabled attributes in .md
- move class from button to draggable-button as a button--modifier
-  remove the styles that set the icon color and just use the button's icon colors

PR #7588, opened 2025-07-01
fix(textInputGroup): fix overlapping status icon and utility button

### LINKED ISSUE #7462: Bug - Text Input Group - status icon can overlay close button if a icon prop is not included
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

If the icon prop is not set on TextInputGroupMain, the validation status icon is incorrectly positioned.
`TextInputGroupMain icon={<SearchIcon />}...`

Image

https://codesandbox.io/p/devbox/fast-firefly-k2wgzy?workspaceId=ws_DaxE3H6rUUo4d5D65eiKhT

srambach:
When the icon prop was not set on a text input group, the status icon (error, warning, success) could overlap with the close button due to missing layout adjustments.

What’s changed?
The SCSS now sets `--#{$text-input-group}__text--Position: relative;` whenever a status modifier is present, ensuring the status icon is always spaced correctly from the close button, even if there’s no leading icon.

A new example has been added in text input group.

Closes #7462

### DIFF
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -97,6 +97,21 @@ For the purposes of this example, the `TextInputGroup` is contained in a wrapper
     {{/button}}
   {{/text-input-group-utilities}}
 {{/text-input-group}}
+
+<br/>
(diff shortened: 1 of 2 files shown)


thatblindgeye (approved):

mcoker (approved):
🚀

PR #7653, opened 2025-07-03
fix(backstop): force light mode and fix no motion class

srambach:
This uses `page.emulateMediaFeatures` to force the light mode unless the `--dark` flag is set.
It also fixes the no motion class to include `-v6-` and adds it directly to the html element without a query to find it first.


mcoker (approved):
🚀

PR #7656, opened 2025-07-10
feat(accordion): add borders for high contrast

### LINKED ISSUE #7611: High contrast borders - Accordion
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10992-59484&t=DkgfQ5XT3Uibqm6l-1

### LINKED ISSUE #7626: High contrast focus changes - Accordion

srambach:
Fixes https://github.com/patternfly/patternfly/issues/7611

Adds a pseudoelement border for the item to the `::after`.
Adds the border on the `expandable-content` element so that it scrolls properly when there are multiple `expandable-content-body`s.

Backstop report showing high contrast difference: https://drive.google.com/file/d/1UH15vqvcKXQJJC_vhwBskjj7iaYap16Q/view?usp=sharing

### DIFF
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -161,7 +161,7 @@
     transition-property: opacity;
   }
 
-  &:is(:hover, :active),
+  &:is(:hover, :focus-within),
   &.pf-m-expanded {
     &::after {
       position: absolute;
--- a/src/patternfly/components/Accordion/accordion.scss
+++ b/src/patternfly/components/Accordion/accordion.scss
@@ -20,6 +20,9 @@
   --#{$accordion}__toggle--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
   --#{$accordion}__toggle--BorderRadius: var(--pf-t--global--border--radius--small);
   --#{$accordion}__toggle--ZIndex: var(--pf-t--global--z-index--xs);
+  --#{$accordion}__toggle--BorderWidth: 0;
+  --#{$accordion}__toggle--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$accordion}__toggle--hover--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
 
   // Accordion toggle toggle-start modifier
   // This decreases the gap between icon and text, alternative is calc it to * 2 the token or create a new token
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Accordion/accordion.scss
```diff
@@ -152,6 +161,18 @@
     transition-property: opacity;
   }
 
+  &:is(:hover, :active),
```
mcoker: I'd probably change this to `:focus-within`? Though it's not that big of a deal, we'd typically add this border to the interactive element as a best practice (https://core-staging.patternfly.org/guidelines#hover-styles). So it would be on the `__toggle` for hover/focus, then on the `__item` for expanded. wdyt?

```suggestion
  &:is(:hover, :focus-within),
```

lboehling (approved):
looks good to me! thank you! 🪗

mcoker (commented):
Hmm... when I tab through the accordion items, it looks like the button's default focus outline is obscuring the Hi-C border. Can reproduce in dev tools by applying just `:focus-visible` or both `:focus` and `:focus-visible`. The border looks fine if you either add `outline: 0` to the button or just apply `:focus` (without `:focus-visible`)

In this screenshot, the first item has `:focus` and the second one has `:focus` and `:focus-visible`

Screenshot 2025-07-29 at 2 02 28 PM

srambach:
> Hmm... when I tab through the accordion items, it looks like the button's default focus outline is obscuring the Hi-C border. Can reproduce in dev tools by applying just `:focus-visible` or both `:focus` and `:focus-visible`. The border looks fine if you either add `outline: 0` to the button or just apply `:focus` (without `:focus-visible`)
> 
> In this screenshot, the first item has `:focus` and the second one has `:focus` and `:focus-visible`
> 
@mcoker Do you only see this in dark mode? I.e., in light mode it looks like the focus ring is visible and so the hi-c border is moot - in dark mode Chrome seems to be using a dark focus ring and it obscures the hi-c border while remaining illegible itself.
In FF, the focus ring doesn't obscure it and/or is visible itself. In Safari, I can't even get keyboard focus to work properly.

mcoker:
@lboehling I pushed a change to accordion that changes things a bit. Now when you hover/focus on the toggle when an item is expanded, the toggle itself (at the top of the expanded item that you click on to expand/collapse) gets its own border. Previously you got a border around the entire item (toggle + expanded content) on hover/focus and when it was expanded, so if you hovered/focused the toggle when the item was expanded, there was no border change. Here are a couple of screenshots showing the difference.

Previously it looked like this. The "hover/focus" border goes around the whole item (same border that's applied when it's expanded), so when hovering the toggle, only its background color changed:

Screenshot 2025-08-01 at 3 56 23 PM


What I updated it to. The "hover/focus" is on the toggle itself, so when the item is expanded, there is a border around the whole item, but when you hover/focus on the toggle in an expanded item, it gets its own border. Otherwise, the only indication you're hovering/focusing the toggle to collapse the item is the background color change, which is really subtle.

Screenshot 2025-08-01 at 3 48 20 PM

WDYT?

lboehling (approved):
good w the change 👍

mcoker (approved):

PR #7666, opened 2025-07-17
fix(table): fix compound expandable top padding

### LINKED ISSUE #7665: Bug - Table - compound expandable padding
Reported in 6.3 testing, compound expandable top padding is missing on all items except the first one in a compound expandable row.

Config should be that:
* Regular expandable removes the top-padding
* Compound expandable keeps the top-padding
* That padding can be removed via `pf-m-no-padding` (`<Td noPadding />`)

Image

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7665

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -647,13 +647,6 @@
         }
       }
 
-      // Add padding back to compound expandable
-      > .#{$table}__control-row ~ .#{$table}__expandable-row.pf-m-expanded {
-        > :is(.#{$table}__th, .#{$table}__td):not(.pf-m-no-padding) {
-          padding-block-start: var(--#{$table}--cell--PaddingBlockStart);
-        }
-      }
-
       > .#{$table}__tr:has(~ .#{$table}__expandable-row) {
         border-block-end: 0;
       }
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Table/table.scss
```diff
@@ -1059,13 +1052,21 @@
   position: relative;
   border-block-end: 0 solid transparent;
 
-  @at-root :not(.#{$table}__control-row) ~ .#{$table}__expandable-row {
+  // remove top padding from expandable row cells by default
+  :not(.#{$table}__control-row) ~ & {
```
mcoker, line 1056: The before/after produce the same compiled CSS, I just made the SASS a little simpler.

src/patternfly/components/Table/table.scss
```diff
@@ -647,13 +647,6 @@
         }
       }
 
-      // Add padding back to compound expandable
-      > .#{$table}__control-row ~ .#{$table}__expandable-row.pf-m-expanded {
-        > :is(.#{$table}__th, .#{$table}__td):not(.pf-m-no-padding) {
-          padding-block-start: var(--#{$table}--cell--PaddingBlockStart);
-        }
-      }
```
mcoker, line 655: This was an opt-in fix - removed it and added it to the base styles.

src/patternfly/components/Table/table.scss
```diff
@@ -1059,13 +1052,21 @@
   position: relative;
   border-block-end: 0 solid transparent;
 
-  @at-root :not(.#{$table}__control-row) ~ .#{$table}__expandable-row {
```
mcoker, line 1062: We remove the top-padding from expandable rows by default (for regular expandable), but compound expandable should retain the top-padding. This selector tried (and failed 😢) to do that by only removing top-padding on `__expandable-row` that doesn't have a preceding sibling with class `__control-row`. But this selector will also target `__expandable-row ~ __expandable-row`, which is why the first `__expandable-row` had padding, but all after it had no padding.

thatblindgeye (approved):

PR #7670, opened 2025-07-17
fix(textinputgroup): update placeholder text color for disabled state

### LINKED ISSUE #7661: Bug - TextInputGroup - low contrast placeholder text with isDisabled
**Describe the problem**
This is an issue with Patternfly 6  

Ansible is using a disabled TextInputGroup (within a MenuToggle), with placeholder text in the TextInputGroup.  The combination of placeholder text, isDisabled prop, and dark theme is causing the text to be a low contrast shade of gray that is not consistent with the same combination of placeholder, disabled, and dark theme in similar components like TextArea or TextInput. 

Visit the "[Disabled text input group](https://www.patternfly.org/components/text-input-group/#disabled-text-input-group)" example and change the `value="Disabled"` prop in the example code to `placeholder="Disabled"`.

This is currently a concern for Ansible's PF6 adoption, though if it is intentional more information is requested to understand this treatment.

**Screenshots**
TextInputGroup:
Image
TextInput:
Image

### LINKED ISSUE #7671: Bug - Text input group - disabled text isn't readable
Noticed in https://github.com/patternfly/patternfly/pull/7670

Input text color is the default text color when text input group is disabled. Isn't really visible in default theme since the default text color is so close to the `--on-disabled` text color, however it's very noticeable in dark theme. 

It looks like this line is probably trying to make the change? https://github.com/patternfly/patternfly/blob/ff6269c0e55af61dce80195d195eb1bc412a9a3a/src/patternfly/components/TextInputGroup/text-input-group.scss#L99

evwilkin:
Closes #7661 
Closes #7671 

This PR:
1. Adds new `--#{$text-input-group}__text-input--Color` variable which defaults to the current value of `var(--pf-t--global--text--color--regular)`.
2. Overrides default values for the text input (both placeholder & value) text color for disabled TextInputGroups
3.  Sets the `color` on the text input group text input to the new `--#{$text-input-group}__text-input--Color` variable.

Confirmed w/PF design placeholder text should use the on-disabled token, matching pattern used in similar components such as [FormControl](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/FormControl/form-control.scss#L279).

Before:
- Input placeholder:
  Screenshot 2025-07-17 at 4 45 46 PM
- Input text:
  Screenshot 2025-07-21 at 10 26 58 AM

After:
- Input placeholder:
  Screenshot 2025-07-17 at 4 45 54 PM
- Input text:
  Screenshot 2025-07-21 at 10 26 50 AM

### DIFF
--- a/src/patternfly/components/TextInputGroup/text-input-group.scss
+++ b/src/patternfly/components/TextInputGroup/text-input-group.scss
@@ -44,6 +44,7 @@
   --#{$text-input-group}__text-input--PaddingInlineStart: var(--pf-t--global--spacer--control--horizontal--default);
   --#{$text-input-group}__text-input--MinWidth: 12ch;
   --#{$text-input-group}__text-input--m-hint--Color: var(--pf-t--global--text--color--placeholder);
+  --#{$text-input-group}__text-input--Color: var(--pf-t--global--text--color--regular);
   --#{$text-input-group}__text-input--placeholder--Color: var(--pf-t--global--text--color--placeholder);
   --#{$text-input-group}__text-input--BackgroundColor: transparent;
   --#{$text-input-group}__text-input--OutlineOffset: -6px;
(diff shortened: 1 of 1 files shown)


mcoker (approved):
LGTM! Though I noticed that the regular disabled text color is also wrong - https://github.com/patternfly/patternfly/issues/7671. Do you want to take a stab at that?

evwilkin:
@mcoker thanks for the callout - I've updated this PR to fix #7671 but would appreciate some guidance on if this is the right approach (adding a new variable, setting it as the default & overriding it for `disabled` - I left existing variables as-is).

mcoker (approved):
LGTM!

mattnolting (approved):
A future consideration, something like this would avoid the issue altogether moving forward, unless the goal is to style disabled placeholder text explicitly. Either way, nice work! Approved

.pf-m-disabled {
  &::placeholder {
  ::placeholder {
    color: inherit;
  }
}

PR #7683, opened 2025-07-23
feat(drawer): add high-contrast border

### LINKED ISSUE #7616: High contrast borders - Drawer
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10994-57323&t=DkgfQ5XT3Uibqm6l-1

srambach:
Fixes #7616 

@lboehling Just want to verify that the border is also correct on the secondary background variant.

Note: this changes a background color because it's currently the way the drawer adds a border and I'm just reusing that.

Update 8/1/25: Moved the border from the pseudoelement background to the true border of the panel and used the new high contrast border width.  Old variables are kept so it's not breaking. [One regression test](https://drive.google.com/file/d/1wuX6ekDwii_YY3if6myMfqFKx0QME5ZS/view?usp=sharing) is failing on resizable inline mobile view because there's still a border since it's inline.

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -550,12 +550,12 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
     --#{$drawer}__panel--BorderInlineEndWidth: var(--#{$drawer}__panel--after--Width);
     --#{$drawer}__panel--BorderInlineStartWidth: 0; // turn off default left border
 
-    &.pf-m-inline, // TODO
+    &.pf-m-inline,
     &.pf-m-static {
       > .#{$drawer}__main > .#{$drawer}__panel:not(.pf-m-no-border, .pf-m-resizable) {
         --#{$drawer}__panel--BorderInlineEndWidth: var(--#{$drawer}--m-inline__panel--after--Width);
 
-        // TODOremove in breaking change
+        // TODO remove in breaking change
         padding-inline-start: 0;
         padding-inline-end: var(--#{$drawer}--m-panel-left--m-inline__panel--PaddingInlineEnd);
 
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -21,7 +21,6 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --#{$drawer}__panel--MinWidth: 50%; // change to __panel--md--MinWidth at breaking change
   --#{$drawer}__panel--MaxHeight: auto;
   --#{$drawer}__panel--ZIndex: var(--pf-t--global--z-index--sm);
-  --#{$drawer}__panel--BorderWidth: 0; 
   --#{$drawer}__panel--BorderColor: var(--#{$drawer}__panel--after--BackgroundColor); // TODO use token directly in breaking change
   --#{$drawer}__panel--BorderBlockStartWidth: 0;
   --#{$drawer}__panel--BorderBlockEndWidth: 0;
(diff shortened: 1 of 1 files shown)


mcoker:
Pretty handy that we have a border there already! Looks like we have a z-index issue with things that go in the drawer overlapping the panel border. From https://patternfly-pr-7683.surge.sh/components/notification-drawer/html-demos/expanded-read/

Screenshot 2025-07-29 at 5 56 44 PM

Since this is for inline/static drawers, I check to see if we had the same problem there and we do not - inline/static add padding that is the same as the border width https://github.com/patternfly/patternfly/blob/4bc3a79ecb4fd86784675469f08a8e6542b448dd/src/patternfly/components/Drawer/drawer.scss#L196

Seems like a regular border would be ideal here but because panels can expand the width of the drawer (width of the entire viewport with a notification drawer on mobile), we wouldn't want the transparent border visible in non-HC theme.

We could use the pseudo element and increase the z-index. We would just need to check the use of z-index elsewhere and make sure we don't use too high of a value. wdyt?

lboehling (approved):
LGTM!

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -321,6 +331,11 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   overflow: auto;
   visibility: hidden; // hidden by default
   background-color: var(--#{$drawer}__panel--BackgroundColor);
+  border: var(--#{$drawer}__panel--BorderWidth) solid var(--#{$drawer}__panel--BorderColor);
```
mcoker: Since this is only set/used here, and we have border-widths for all of the sides that will override this, would this ever get used? 

```suggestion
  border: solid var(--#{$drawer}__panel--BorderColor);
```

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -21,6 +21,12 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   --#{$drawer}__panel--MinWidth: 50%; // change to __panel--md--MinWidth at breaking change
   --#{$drawer}__panel--MaxHeight: auto;
   --#{$drawer}__panel--ZIndex: var(--pf-t--global--z-index--sm);
+  --#{$drawer}__panel--BorderWidth: 0;
```
mcoker: I think you can remove this since all of the border sides are defined, I don't think it would ever take effect.

```suggestion
```

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -693,6 +700,12 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
       }
 
       > .#{$drawer}__main > .#{$drawer}__panel {
+        --#{$drawer}--m-expanded__panel--BoxShadow: none;
+
+        &:not(.pf-m-no-border) {
+          --#{$drawer}__panel--BorderColor: var(--#{$drawer}--m-inline--m-expanded__panel--after--BackgroundColor);
+        }
+
```
mcoker: Curious what these styles are needed for? Looks like they're duped from the block above https://github.com/patternfly/patternfly/blob/01bb2b8705c81557d82f0c25674523e8a449d513/src/patternfly/components/Drawer/drawer.scss#L687C1-L693

mcoker (commented):
Just a few comments. Everything else looks great!

On these changes, there are a lot of changes that are like this where the removed style was on `::after` with regular property declarations, and added styles are setting custom properties. I want to say the old styles were directly on elements with regular declarations because declaring vars caused issues in nested drawers (inheriting the var change instead and not resetting properly with the nested drawer), but I'm not seeing any issues from testing nested drawers in this PR, so I think it's good to go! I'm just mentioning as something to keep an eye out for when products pull this in, in case there are nested style issues.

mcoker (approved):

PR #7688, opened 2025-07-28
feat(wizard): add high-contrast borders

### LINKED ISSUE #7609: High contrast borders - Wizard
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10994-65005&t=yNHLMNf7OcjnzTQn-1

srambach:
Fixes #7609 

Added border styling to the nav-link-main element using a ::after pseudoelement
Border appears on hover and current states with high-contrast color
Border width increases for current/selected state
Border follows the same border radius as the background container

Added border styling to the ::before pseudoelement (step number)
Border appears by default with high-contrast color
Border is removed for current and disabled states
Border follows the existing border radius for the step number

Visual regression test at: https://drive.google.com/file/d/1WRRztj5K40iL59tYWDErCfyga-zraOk-/view?usp=sharing
I believe the visible changes are due to the new 1px border on modals. In the non-modal examples that failed I can't actually see a difference (other than the usual artifacts).

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -60,8 +60,8 @@
   --#{$wizard}__nav-link-main--BorderRadius: var(--pf-t--global--border--radius--small);
   --#{$wizard}__nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
   --#{$wizard}__nav-link-main--BorderWidth: 0;
-  --#{$wizard}__nav-link--hover--nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
-  --#{$wizard}__nav-link--m-current--nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--strong);
+  --#{$wizard}__nav-link--hover__nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
+  --#{$wizard}__nav-link--m-current__nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--strong);
 
   // Nav link toggle icon
   --#{$wizard}__nav-link-toggle--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -58,6 +58,11 @@
   --#{$wizard}__nav-link-main--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$wizard}__nav-link-main--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
   --#{$wizard}__nav-link-main--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$wizard}__nav-link-main--BorderColor: transparent;
+  --#{$wizard}__nav-link-main--BorderWidth: var(--pf-t--global--border--width--regular);
+  --#{$wizard}__nav-link--hover--nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$wizard}__nav-link--m-current--nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$wizard}__nav-link--m-current--nav-link-main--BorderWidth: var(--pf-t--global--border--width--strong);
 
   // Nav link toggle icon
   --#{$wizard}__nav-link-toggle--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
@@ -73,13 +78,17 @@
   --#{$wizard}__nav-link--before--Height: var(--pf-t--global--icon--size--font--xl);
   --#{$wizard}__nav-link--before--InsetBlockStart: var(--pf-t--global--spacer--sm);
   --#{$wizard}__nav-link--before--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
+  --#{$wizard}__nav-link--before--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$wizard}__nav-link--before--BorderWidth: var(--pf-t--global--border--width--regular);
   --#{$wizard}__nav-link--before--BorderRadius: var(--pf-t--global--border--radius--large);
   --#{$wizard}__nav-link--before--Color: var(--pf-t--global--text--color--regular);
   --#{$wizard}__nav-link--before--FontSize: var(--pf-t--global--font--size--body--default);
   --#{$wizard}__nav-link--m-current--before--BackgroundColor: var(--pf-t--global--color--brand--default);
   --#{$wizard}__nav-link--m-current--before--Color: var(--pf-t--global--text--color--on-brand--default);
+  --#{$wizard}__nav-link--m-current--before--BorderColor: transparent;
   --#{$wizard}__nav-link--m-disabled--before--BackgroundColor: transparent;
   --#{$wizard}__nav-link--m-disabled--before--Color: var(--pf-t--global--text--color--disabled);
+  --#{$wizard}__nav-link--m-disabled--before--BorderColor: transparent;
 
   // Nav link status icon
   --#{$wizard}__nav-link-status-icon--Color: var(--pf-t--global--icon--color--regular);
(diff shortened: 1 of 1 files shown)


lboehling (changes requested):
This looks great! It does look like the drawer's border gets cut off on the HTML demo with drawer -- but not sure if that's a wizard or drawer problem. Otherwise this lgtm!
Screenshot 2025-08-05 at 10 25 25 AM

mcoker:
@lboehling good catch - that's a pre-existing bug with the drawer panel border. You can see it here, too - https://www.patternfly.org/components/wizard/html-demos/with-drawer-expanded/

Looks like @srambach fixed it in her [HC drawer PR](https://github.com/patternfly/patternfly/pull/7683) - https://pf-pr-7683.surge.sh/components/wizard/html-demos/with-drawer-expanded/

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -58,6 +58,11 @@
   --#{$wizard}__nav-link-main--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$wizard}__nav-link-main--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
   --#{$wizard}__nav-link-main--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$wizard}__nav-link-main--BorderColor: transparent;
+  --#{$wizard}__nav-link-main--BorderWidth: var(--pf-t--global--border--width--regular);
```
mcoker: When we add base border styles that are not intended to be visible by default, and are enabled with some state/class, we tend to do that one of three ways:

* Add a non-zero border width with transparent border (what's done here). Turn it on by changing the border-color in the case you need the border.
* Add a zero border width and whatever border-color will be used. Turn it on by changing the border width.
* Add a zero border width with transparent border. Turn it on by changing both the border width and border color.

Just [digging](https://sarahmhigley.com/writing/whcm-quick-tips/) [around](https://piccalil.li/blog/use-transparent-borders-and-outlines-to-assist-with-high-contrast-mode/) a little, in high contrast modes like WHCM (and what I've seen of [FF HCM](https://firefox-source-docs.mozilla.org/accessible/HCMMediaQueries.html)), if something has a border (even transparent), that border will become visible in WHCM because WHCM overrides colors set in CSS. So which approach we take above doesn't matter in light (or our high contrast theme), but considering future support for WHCM, it would.

So for a change like this, I think we'd ask - should this element have a default border (even when not hovered/current) if a user turned on WHCM? I'm not sure what the answer is there, but since we aren't creating the default border in our HC theme, I'd probably say no, since that seems like design's direction. Though from a pure a11y standpoint, I'd argue that it should, since it's a button and you can click on it and there is no border/boundary by default to show the click area. We also disable the parent `<button>`'s border, so there is no border there in any HC mode, either.

All that said, I would probably update this to use the HC border color token and set the border-width to "0", which would make the borders in WHCM match our HC theme. That would also be more consistent with how other borders behave - like the approach we're taking with adding regular (non-pseudo) borders where we are setting the border-color by default, but not the border-width until we need it.

mcoker: Since it will also support FF/Windows HCM, can you go ahead and update the width to "0" so the border doesn't show by default there? I think you'll end up with less overall code that way, too.

```suggestion
  --#{$wizard}__nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
  --#{$wizard}__nav-link-main--BorderWidth: 0;
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -58,6 +58,11 @@
...
+  --#{$wizard}__nav-link--hover--nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
```
mcoker: Nit on the var name, and since you'll set the color by default, this can update the border width.

```suggestion
  --#{$wizard}__nav-link--hover__nav-link-main--BorderWidth: var(--pf-t--global--border--width--regular);
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -497,6 +510,7 @@
   &:where(:hover, :focus) {
     --#{$wizard}__nav-link--Color: var(--#{$wizard}__nav-link--hover--Color);
     --#{$wizard}__nav-link-main--BackgroundColor: var(--#{$wizard}__nav-link--hover--BackgroundColor);
+    --#{$wizard}__nav-link-main--BorderColor: var(--#{$wizard}__nav-link--hover--nav-link-main--BorderColor);
```
mcoker:
```suggestion
    --#{$wizard}__nav-link-main--BorderWidth: var(--#{$wizard}__nav-link--hover__nav-link-main--BorderWidth);
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -58,6 +58,11 @@
   --#{$wizard}__nav-link-main--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$wizard}__nav-link-main--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
   --#{$wizard}__nav-link-main--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$wizard}__nav-link-main--BorderColor: transparent;
+  --#{$wizard}__nav-link-main--BorderWidth: var(--pf-t--global--border--width--regular);
+  --#{$wizard}__nav-link--hover--nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$wizard}__nav-link--m-current--nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
```
mcoker: I don't think you need this since you've set it already?

```suggestion
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -465,18 +474,22 @@
...
+    inset-block-start: var(--#{$wizard}__nav-link--before--InsetBlockStart); // 8px; // TODO variable
     content: counter(wizard-nav-count);
-}
+    border: var(--#{$wizard}__nav-link--before--BorderWidth) solid var(--#{$wizard}__nav-link--before--BorderColor);
+  }
 
   &.pf-m-current {
     --#{$wizard}__nav-link--Color: var(--#{$wizard}__nav-link--m-current--Color);
     --#{$wizard}__nav-link-main--BackgroundColor: var(--#{$wizard}__nav-link--m-current--BackgroundColor);
+    --#{$wizard}__nav-link-main--BorderColor: var(--#{$wizard}__nav-link--m-current--nav-link-main--BorderColor);
```
mcoker: Shouldn't be needed because it's set already?

```suggestion
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -58,6 +58,10 @@
   --#{$wizard}__nav-link-main--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$wizard}__nav-link-main--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
   --#{$wizard}__nav-link-main--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$wizard}__nav-link-main--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$wizard}__nav-link-main--BorderWidth: 0;
+  --#{$wizard}__nav-link--hover--nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
+  --#{$wizard}__nav-link--m-current--nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--strong);
```
mcoker:
```suggestion
  --#{$wizard}__nav-link--hover__nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
  --#{$wizard}__nav-link--m-current__nav-link-main--BorderWidth: var(--pf-t--global--high-contrast--border--width--strong);
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -466,18 +474,21 @@
...
+    inset-block-start: var(--#{$wizard}__nav-link--before--InsetBlockStart); // 8px; // TODO variable
     content: counter(wizard-nav-count);
-}
+    border: var(--#{$wizard}__nav-link--before--BorderWidth) solid var(--#{$wizard}__nav-link--before--BorderColor);
+  }
 
   &.pf-m-current {
     --#{$wizard}__nav-link--Color: var(--#{$wizard}__nav-link--m-current--Color);
     --#{$wizard}__nav-link-main--BackgroundColor: var(--#{$wizard}__nav-link--m-current--BackgroundColor);
+    --#{$wizard}__nav-link-main--BorderWidth: var(--#{$wizard}__nav-link--m-current--nav-link-main--BorderWidth);
```
mcoker:
```suggestion
    --#{$wizard}__nav-link-main--BorderWidth: var(--#{$wizard}__nav-link--m-current__nav-link-main--BorderWidth); 
```

src/patternfly/components/Wizard/wizard.scss
```diff
@@ -498,6 +509,7 @@
   &:where(:hover, :focus) {
     --#{$wizard}__nav-link--Color: var(--#{$wizard}__nav-link--hover--Color);
     --#{$wizard}__nav-link-main--BackgroundColor: var(--#{$wizard}__nav-link--hover--BackgroundColor);
+    --#{$wizard}__nav-link-main--BorderWidth: var(--#{$wizard}__nav-link--hover--nav-link-main--BorderWidth);
```
mcoker:
```suggestion
    --#{$wizard}__nav-link-main--BorderWidth: var(--#{$wizard}__nav-link--hover__nav-link-main--BorderWidth);
```

mcoker (commented):
Just one tiny nit

mcoker (approved):
🚀

PR #7691, opened 2025-07-28
fix(label-group): adjust close icon color

### LINKED ISSUE #7676: Bug - Label - Incorrect "x" icon and color used in Label Groups
**Describe the problem**
The Label Groups examples are using the incorrect "x" icon and color. The examples are currently showing "fa-times-circle" in pf-t--global--icon--color--subtle, rather than the "fa-times" in pf-t--global--icon--color--regular.

1. Go to patternfly.org
2. Go to "Components"
3. Go to "Label"
4. Scroll down to "Label group with removable categories" and "Vertical label group"

**Expected behavior**
The close "x" should be the "fa-times" icon in pf-t--global--icon--color--regular.
Image

**Screenshots**
Image

**What is your environment?**
 - OS: Mac
 - Browser: Chrome
 - Version 136.0.7103.114

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7676

~~This fixes the color.~~ After chatting with @lboehling, we decided to use a small button here, which has a black icon by default.

Opened a follow-up issue in react to fix the icon - https://github.com/patternfly/patternfly-react/issues/11941

### DIFF
--- a/src/patternfly/components/Label/label-group-close.hbs
+++ b/src/patternfly/components/Label/label-group-close.hbs
@@ -2,5 +2,5 @@
   {{#if label-group-close--attribute}}
     {{{label-group-close--attribute}}}
   {{/if}}>
-  {{> button button--IsPlain=true button--HasNoPadding=true button--IsIcon=true button--icon="times" button--attribute=(concat 'aria-labelledby="' label-group--id '-button ' label-group--id '-label" aria-label="Close label group" id="' label-group--id '-button"')}}
+  {{> button button--IsPlain=true button--IsSmall=true button--IsIcon=true button--icon="times" button--attribute=(concat 'aria-labelledby="' label-group--id '-button ' label-group--id '-label" aria-label="Close label group" id="' label-group--id '-button"')}}
 </div>
--- a/src/patternfly/components/Label/label-group.scss
+++ b/src/patternfly/components/Label/label-group.scss
@@ -41,6 +41,7 @@
   --#{$label-group}__label--MaxWidth: 18ch;
 
   // Close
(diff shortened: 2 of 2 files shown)


mcoker:
@bekah-stephens @lboehling - looks like the label group remove/close icon uses our [plain/no-padding button variation](https://www.patternfly.org/components/button#plain-with-no-padding), which has a grey icon to indicate that it's interactive (turns black on hover), since it doesn't have a background color change on hover. This PR keeps that variation and just updates the button's icon color.

And here is what the label group close/remove icon looks like if I also remove the no-padding variation. I think this looks a lot better and is probably what we're going for instead?

Screenshot 2025-07-28 at 6 35 58 PM

@srambach pointed out that our outlined label close button icons are also grey (because they use the no-padding button variation) - https://www.patternfly.org/components/label#outlined-labels. The filled label close button icons are not grey because they are all set to use unique colors - the `--on-[color]` icon color token for whatever the label color is. For example, `--pf-t--global--icon--color--nonstatus--on-orangered--default`.

FWIW here is what a small plain button (removed the no-padding variation) looks like in a label if I also reduce the font-size to match the label text, in case we wanted to update the label close icons to use regular buttons so they would have some change when you hover/focus on them, indicating they're interactive. That would also mean the outlined label close buttons could be black, too.

Screenshot 2025-07-28 at 6 30 08 PM

All that said, are there any other updates we want to make here?

mcoker:
After chatting with @lboehling, updated to use a small/plain button here instead.

srambach (approved):
👍🏻 
Should we update screen shots soon to keep current?

bekah-stephens (approved):
lgtm!

PR #7693, opened 2025-07-28
fix(wizard): nav item horizontal alignment fixes

### LINKED ISSUE #7687: Bug - Wizard - nav item step issues
First issue - with long step names, the step number is no longer a circle. Whatever creates the circle likely just needs `flex-grow: 0`
    Image

Second issue - as part of a PR to add [nav item step status](https://github.com/patternfly/patternfly/pull/6447), we introduced a visual change that I'm not sure was on purpose.

Prior to that change ^ (and in Figma) steps used to span the width of the wizard nav - the background spanned the full width and the caret/toggle icon was right aligned in the step. You can see it here - https://patternfly-pr-6598.surge.sh/components/wizard/html/basic/

Image

After the change, the step background only spans the width of the text, and the caret is right beside the text. You can see it here - https://www.patternfly.org/components/wizard/html/basic/

Image

@lboehling @andrew-ronaldson can you confirm how it should look?

srambach:
Fixes #7687 

Fixes step number by setting flex-shrink to 0
Fixes the item width by moving flex-grow to the item main rather than the text.

image

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -452,6 +452,7 @@
   &::before {
     position: relative;
     display: inline-flex;
+    flex-shrink: 0;
     align-items: center;
     justify-content: center;
     width: var(--#{$wizard}__nav-link--before--Width);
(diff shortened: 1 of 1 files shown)


mcoker (approved):
🧙‍♂️👍

PR #7701, opened 2025-07-30
fix(table): fix toggle vertical alignment in thead

### LINKED ISSUE #7699: Bug - Table (expandable) - toggle misalignment in THead
**Describe the problem**
This is an issue with PF6

The toggle/checkbox/text in the THead are not aligned correctly in the THead when the expandable toggle is present (they align correctly in the rest of the table).  The alignment is incorrect in both regular & compact tables, but is different between the two (toggle is below checkbox/text on regular table and above on compact):

[Expandable example](https://www.patternfly.org/components/table/html/#expandable-example):

Image

[Compact expandable example](https://www.patternfly.org/components/table/html/#compact-expandable-example):

Image

Ansible - September release but targeting early August code freeze.

evwilkin:
Closes #7699 

This PR fixes alignment issues between the expandable toggle button, selectable checkbox, and column headers in the table thead that affect both expandable tables and expandable & compact tables.  

It adds 2 new variables to set vertical alignment, both default to the existing value of `bottom` and both are changed to `middle` to resolve alignment bugs:
1. `--#{$table}__thead--VerticalAlign: bottom`
    - Updated to `middle` on expandable tables.
2. `--#{$table}__thead__check--VerticalAlign: bottom`
    - Updated to `middle` on tables that are both expandable & compact.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -22,6 +22,7 @@
 
   // * Table thead toggle
   --#{$table}__thead__toggle--PaddingBlockEnd: var(--pf-t--global--spacer--xs);
+  --#{$table}__thead__toggle--VerticalAlign: middle;
 
   // * Table body cell
   --#{$table}__tbody--cell--PaddingBlockStart: var(--#{$table}--cell--Padding--base);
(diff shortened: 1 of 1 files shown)


evwilkin:
From @mattnolting 		

> The issue with the approach is that when th content is set to pf-m-wrap and th content starts wrapping, the expand button then is misaligned.
> Here's what's happening. Button (pf-v6-button ) and table button (pf-v6-c-table__button) are different heights and the icons within them are different sizes. That's creating the misalignment. They also have different padding settings, complicating the misalignment. If they were the same height (created by line height + font size), they would align properly.
> https://github.com/patternfly/patternfly/pull/7050#issuecomment-2364534373
> I attempted to fix it here, though I do believe that this could be simplified. In this comment, you can see the issue - essentially the row height varies because of its contents and causes layout issues as a result.
> The misalignment is because table header cell text is smaller than standard table data cell text, but the icon and button sizes are mixed.

mcoker:
Closing in favor of https://github.com/patternfly/patternfly/pull/7722

PR #7703, opened 2025-07-30
feat(simpleList): added high contrast border for hover/focus/current

### LINKED ISSUE #7605: High contrast borders - Simple list - Selected states
https://www.figma.com/design/wKcOq7IrzfL1gEK2mocd6r/Semantic-dimension-border-width----font-weight-adjustment-test?node-id=10994-65003

rebeccaalpert:
Commented in bit about `:root:where(.pf-v6-theme-high-contrast) {` and added it to root to check.

https://pf-pr-7703.surge.sh/components/simple-list

Selected:

| Before | After |
|-|-|
|Screenshot 2025-07-30 at 12 43 43 PM|Screenshot 2025-07-30 at 12 45 58 PM

| Before | After |
|-|-|
|Screenshot 2025-07-30 at 12 43 37 PM|Screenshot 2025-07-30 at 12 45 52 PM

### DIFF
--- a/src/patternfly/components/SimpleList/simple-list.scss
+++ b/src/patternfly/components/SimpleList/simple-list.scss
@@ -18,7 +18,11 @@
   // Simple list item link
   --#{$simple-list}__item-link--m-link--Color: var(--pf-t--global--text--color--link--default);
   --#{$simple-list}__item-link--m-link--m-current--Color: var(--pf-t--global--text--color--link--hover);
+  --#{$simple-list}__item-link--m-link--m-current--BorderWidth: var(--pf-t--global--border--width--strong);
+  --#{$simple-list}__item-link--m-link--m-current--BorderColor: var(--pf-t--global--border--color--high-contrast);
   --#{$simple-list}__item-link--m-link--hover--Color: var(--pf-t--global--text--color--link--hover);
+  --#{$simple-list}__item-link--m-link--hover--BorderWidth: var(--pf-t--global--border--width--regular);
+  --#{$simple-list}__item-link--m-link--hover--BorderColor: var(--pf-t--global--border--color--high-contrast);
 
   // Simple list title
   --#{$simple-list}__title--PaddingBlockStart: var(--pf-t--global--spacer--sm);
@@ -58,12 +62,35 @@
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--hover--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--hover--Color);
 
+    position:relative;
     text-decoration-line: none;
+
+    &::after {
+      position: absolute;
+      inset: 0;
+      pointer-events: none;
+      content: "";
+      background-color: inherit;
+      border: var(--#{$simple-list}__item-link--m-link--hover--BorderWidth) solid var(--#{$simple-list}__item-link--m-link--hover--BorderColor);
+      border-radius: inherit;
+    }
   }
 
   &.pf-m-current {
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--m-current--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--m-current--Color);
+
+    position: relative;
+    
+    &::after {
+      position: absolute;
+      inset: 0;
+      pointer-events: none;
+      content: "";
+      background-color: inherit;
+      border: var(--#{$simple-list}__item-link--m-link--m-current--BorderWidth) solid var(--#{$simple-list}__item-link--m-link--m-current--BorderColor);
+      border-radius: inherit;
+    }
   }
 }
 
--- a/src/patternfly/components/SimpleList/simple-list.scss
+++ b/src/patternfly/components/SimpleList/simple-list.scss
@@ -14,7 +14,11 @@
   --#{$simple-list}__item-link--hover--Color: var(--pf-t--global--text--color--subtle);
   --#{$simple-list}__item-link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
   --#{$simple-list}__item-link--BorderRadius: var(--pf-t--global--border--radius--tiny);
-
+  --#{$simple-list}__item-link--BorderWidth: 0;
+  --#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$simple-list}__item-link--hover--BorderWidth: var(--pf-t--global--border--width--regular);
+  --#{$simple-list}__item-link--current--BorderWidth: var(--pf-t--global--border--width--strong);
+  
   // Simple list item link
   --#{$simple-list}__item-link--m-link--Color: var(--pf-t--global--text--color--link--default);
   --#{$simple-list}__item-link--m-link--m-current--Color: var(--pf-t--global--text--color--link--hover);
@@ -67,8 +71,7 @@
   &:is(:hover, :focus) {
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--hover--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--hover--Color);
-    --#{$simple-list}__item-link--BorderWidth: var(--pf-t--global--border--width--regular);
-    --#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
+    --#{$simple-list}__item-link--BorderWidth: var(--#{$simple-list}__item-link--hover--BorderWidth);
 
     text-decoration-line: none;
   }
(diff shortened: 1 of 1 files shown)


src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -58,12 +62,35 @@
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--hover--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--hover--Color);
 
+    position:relative;
```
mcoker: Can you move this up to the base element so that it applies all the time, not just on `:hover/:focus`? That way, if someone wants to turn on this border for whatever reason in their own custom theme for a use case that isn't `:hover/:focus`, they can. You probably have to add it as the first rule here, above `display: block;` https://github.com/patternfly/patternfly/blob/ae70c4e89b6779ca4e02c3d05dbaca13b3252537/src/patternfly/components/SimpleList/simple-list.scss#L40-L41

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -58,12 +62,35 @@
...
+
+    &::after {
+      position: absolute;
+      inset: 0;
+      pointer-events: none;
+      content: "";
+      background-color: inherit;
+      border: var(--#{$simple-list}__item-link--m-link--hover--BorderWidth) solid var(--#{$simple-list}__item-link--m-link--hover--BorderColor);
+      border-radius: inherit;
+    }
```
mcoker: Same for this, can you put it under `.#{$simple-list}__item-link` (not nested under `:hover/focus`) so it can be applied any time? This change and the previous one will let you dry ths rest of the code out a little, too.

It does mean you'll probably want to update this line:

```
border: var(--#{$simple-list}__item-link--m-link--hover--BorderWidth) solid var(--#{$simple-list}__item-link--m-link--hover--BorderColor);
```

to this:

```
border: var(--#{$simple-list}__item-link--BorderWidth) solid var(--#{$simple-list}__item-link--BorderColor);
```

So you'll need to add the new vars as base vars.

And you shouldn't need `background-color: inherit;` - since the background-color of the parent is a 25% opacity color, this is applying another layer of that color overlaying the button.

Also no need to `--m-link` in the var names. You'd only put that in there if these were borderwith/color vars for the `.pf-m-link` variation of `__item-link`.

mcoker: Then once you do that, you can just add this var under the `&:is(:hover, :focus)` block

```
--#{$simple-list}__item-link--BorderColor: var(--#{$simple-list}__item-link--hover--BorderColor);
```

Then you can remove all of the stuff you added under `.pf-m-current` and replace it with

````
--#{$simple-list}__item-link--BorderWidth: var(--#{$simple-list}__item-link--m-current--BorderWidth);
--#{$simple-list}__item-link--BorderColor: var(--#{$simple-list}__item-link--m-current--BorderColor);
```

Something like that anyways 😁

mcoker (changes requested):
Left a few comments! It looks good now, other than `background: inherit;` on the pseudo elements - that's making the background too dark and making the button text a little greyed out because it's adding a 25% background overlay on top of the button.

But left some more suggestions for how to make this code a little more efficient. Lemme know if you have any questions!

rebeccaalpert:
Should be all set @mcoker! Thank you.

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -48,6 +49,15 @@
   border: none;
   border-radius: var(--#{$simple-list}__item-link--BorderRadius);
 
+  &::after {
+    position: absolute;
+    inset: 0;
+    pointer-events: none;
+    content: "";
+    border: var(--#{$simple-list}__item-link--BorderWidth) solid var(--#{$simple-list}__item-link--BorderColor);
```
mcoker, line 61: Can you add vars at the top of the stylesheet so we have default values for these? You can set them as:

```
--#{$simple-list}__item-link--BorderWidth: 0;
--#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
```

This will set the `border-color` as a base style, so you shouldn't need to redefine it below for hover/focus/current - you can just update the `border-width`s below

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -57,13 +67,17 @@
   &:is(:hover, :focus) {
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--hover--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--hover--Color);
+    --#{$simple-list}__item-link--BorderWidth: var(--pf-t--global--border--width--regular);
+    --#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
```
mcoker: As a general rule, we only use component vars (not tokens) in the style declarations for a component. We do all of the token assigning at the top of the stylesheet. So you could add this to the vars at the top of the stylesheet:

```
    --#{$simple-list}__item-link--hover--BorderWidth: var(--pf-t--global--border--width--regular);
```

Then you can update this block to:

```suggestion
    --#{$simple-list}__item-link--BorderWidth: var(--#{$simple-list}__item-link--hover--BorderWidth);
```

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -57,13 +67,17 @@
...
 
     text-decoration-line: none;
   }
 
   &.pf-m-current {
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--m-current--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--m-current--Color);
+    --#{$simple-list}__item-link--BorderWidth: var(--pf-t--global--border--width--strong);
+    --#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
```
mcoker: The comment from the hover vars above applies here, too. Just need to create a component var at the top of the stylesheet for the current item-link border-width, then use it here.

mcoker (commented):
Nice, looking much better! Just need to update some of the vars a bit.

rebeccaalpert:
I think this should work? Just pushed those changes.

srambach (approved):
Nice!

lboehling (approved):
yay!

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -14,7 +14,11 @@
   --#{$simple-list}__item-link--hover--Color: var(--pf-t--global--text--color--subtle);
   --#{$simple-list}__item-link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
   --#{$simple-list}__item-link--BorderRadius: var(--pf-t--global--border--radius--tiny);
-
+  --#{$simple-list}__item-link--BorderWidth: 0;
+  --#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$simple-list}__item-link--hover--BorderWidth: var(--pf-t--global--border--width--regular);
+  --#{$simple-list}__item-link--current--BorderWidth: var(--pf-t--global--border--width--strong);
```
mcoker: If the var is for a modifier class `.pf-m-[something]`, the var scheme is `--m-[something`

```suggestion
  --#{$simple-list}__item-link--m-current--BorderWidth: var(--pf-t--global--high-contrast--border--width--strong);
```

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -57,13 +71,15 @@
...
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--hover--Color);
+    --#{$simple-list}__item-link--BorderWidth: var(--#{$simple-list}__item-link--hover--BorderWidth);
 
     text-decoration-line: none;
   }
 
   &.pf-m-current {
     --#{$simple-list}__item-link--BackgroundColor: var(--#{$simple-list}__item-link--m-current--BackgroundColor);
     --#{$simple-list}__item-link--Color: var(--#{$simple-list}__item-link--m-current--Color);
+    --#{$simple-list}__item-link--BorderWidth: var(--#{$simple-list}__item-link--current--BorderWidth);
```
mcoker:
```suggestion
    --#{$simple-list}__item-link--BorderWidth: var(--#{$simple-list}__item-link--m-current--BorderWidth);
```

src/patternfly/components/SimpleList/simple-list.scss
```diff
@@ -14,7 +14,11 @@
   --#{$simple-list}__item-link--hover--Color: var(--pf-t--global--text--color--subtle);
   --#{$simple-list}__item-link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
   --#{$simple-list}__item-link--BorderRadius: var(--pf-t--global--border--radius--tiny);
-
+  --#{$simple-list}__item-link--BorderWidth: 0;
+  --#{$simple-list}__item-link--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$simple-list}__item-link--hover--BorderWidth: var(--pf-t--global--border--width--regular);
```
mcoker: Since adding the new border width tokens, do you mind updating these borders to use them?

```suggestion
  --#{$simple-list}__item-link--hover--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
```

mcoker (changes requested):
Just a few small nits

rebeccaalpert:
The high-contrast--border--width tokens don't seem to be working for me, but I applied these changes.

mcoker (approved):
Nice!

Issue #7702, opened 2025-07-30
High contrast - add border-width tokens


mcoker:
Right now the only HC token we have is `--pf-t--global--border--color--high-contrast` (transparent in default theme, `--pf-t--global--border--color--default` in HC), so if we add a HC border to an element in default theme, we're adding a border-width, too, to avoid having to manipulate border-width in HC mode manually. This results in an unwanted transparent border around things in default theme, creating a gap that can cause visual bugs.

Chatting with @lboehling, we decided to add HC border-width tokens that are 0px in default theme, and set to our standard border width sizes in HC mode. This mimics how `--pf-t--global--border--color--high-contrast` works, too. This allows us to put HC border-widths and border-colors on components in default theme without affecting the layout in any way, and the borders are only shown in HC mode.

PR #7704, opened 2025-07-30
fix(high-contrast): add temp border-width tokens, update modal

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7702

Updates modal just to illustrate it works. We'll need to create follow-ups to review the borders we've added and consider using this approach.


srambach (approved):
🔲

PR #7705, opened 2025-07-30
chore(codeBlock): make aria-labels unique

### LINKED ISSUE #7272: Codeblock - aria-label of copy button should be unique
For both examples, the copy button's aria-label should also convey what will be copied, e.g. "Copy to clipboard basic example code content" since there are multiple copy buttons on the page.

**Any other information?**

rebeccaalpert:
Give each example unique aria-labels for all existing buttons


srambach (approved):
L 🤾🏻‍♀️  TM!

thatblindgeye (approved):

PR #7706, opened 2025-07-30
chore(descriptionList): add unique aria-labels to examples and demos

### LINKED ISSUE #7277: DescriptionList - lists should have unique aria-labels
In pages where multiple DescriptionLists are rendered, each should have a unique aria-label to follow the recommended approach for AT that show each list on a page i.e. in VoiceOver's rotor menu. When there's more than 1 on a page, you see a list of just "definition list" which makes it difficult to traverse via that menu.

**Any other information?**

rebeccaalpert:
Manually added ability to have aria-label. Manually added to demos and checked with VO. For examples, added 1 manually and checked with VO. Then tried using Cursor to generate additional aria-labels based on example title. Manually reviewed Cursor suggestions.

Assisted-by: Cursor (used to generate aria-labels for all but 1 example based on title)

### DIFF
--- a/src/patternfly/components/DescriptionList/description-list.hbs
+++ b/src/patternfly/components/DescriptionList/description-list.hbs
@@ -1,4 +1,6 @@
-<dl class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
+<dl {{#if description-list--aria-label}}
+    aria-label="{{description-list--aria-label}}"
+  {{/if}} class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
   {{#if description-list--attribute}}
     {{{description-list--attribute}}}
   {{/if}}>
--- a/src/patternfly/components/DescriptionList/description-list.hbs
+++ b/src/patternfly/components/DescriptionList/description-list.hbs
@@ -1,6 +1,7 @@
-<dl {{#if description-list--aria-label}}
+<dl class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
+  {{#if description-list--aria-label}}
     aria-label="{{description-list--aria-label}}"
-  {{/if}} class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
+  {{/if}}
   {{#if description-list--attribute}}
     {{{description-list--attribute}}}
   {{/if}}>
--- a/src/patternfly/components/DescriptionList/examples/DescriptionList.md
+++ b/src/patternfly/components/DescriptionList/examples/DescriptionList.md
@@ -8,52 +8,52 @@ cssPrefix: pf-v6-c-description-list
 
 ### Default
 ```hbs
-{{> description-list__example description-list--attribute='aria-label="Default example"' description-list--title="Default DL"}}
+{{> description-list__example description-list--aria-label="Default example"}}
 ```
 
 ### Term help text
 ```hbs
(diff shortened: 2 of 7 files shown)


srambach (approved):
LGTM. Those are some crazy example names 😆

src/patternfly/components/DescriptionList/description-list.hbs
```diff
@@ -1,4 +1,6 @@
-<dl class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
+<dl {{#if description-list--aria-label}}
+    aria-label="{{description-list--aria-label}}"
+  {{/if}} class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
```
mcoker: Typically the format of an hbs file and its `--modifier` and `--attribute` stuff is in this order
* Classes. And we'll add any checks for additional paramaters (eg, `component--IsThing`) that add classes before `component--modifier` so that `component--modifier` is always the last thing in the class list.
* Additional custom HTML attributes set in the hbs file. This is where we'll put `aria-label` (like you have below or `tabindex` or any other specific attributes.
* The output of `component--atribute`. This will always come after any other attributes set in the hbs file. 

```suggestion
<dl class="{{pfv}}description-list{{#if description-list--modifier}} {{description-list--modifier}}{{/if}}"
  {{#if description-list--aria-label}}
    aria-label="{{description-list--aria-label}}"
  {{/if}}
```

src/patternfly/components/DescriptionList/examples/DescriptionList.md
```diff
@@ -8,118 +8,118 @@ cssPrefix: pf-v6-c-description-list
 
 ### Default
 ```hbs
-{{> description-list__example description-list--title="Default DL"}}
+{{> description-list__example description-list--attribute='aria-label="Default example"' description-list--title="Default DL"}}
```
mcoker: I only briefly glanced at this so I could be missing something, but it looks like
* All of these examples have a `description-list--title` parameter already set with unique text per example - except for a handful are all set to `description-list--title="Card DL", but looks like those could be made unique.
* As far as I can tell, `description-list--title` doesn't actually do anything? I don't see any uses of it outside of `DescriptionList.md` where it's being set. 
* You added `description-list--aria-label` but I don't see it being used. Setting it via `description-list--attribute='aria-label="Default example"'` is fine, but it's usually easier to manage by not using the `--attribute` parameter, so using `description-list--aria-label` would be preferred.

If those are true, I might either use the `description-list--title` parameters that already exist as the value for `aria-label` (so `aria-label="{{description-list--title}}"`) and update the ones that have the same value to have unique values, or (this is what I'd actually do) 1) remove all of the `description-list--title` parameters since they don't do anything and it's confusing and 2) do some search/replace action to change `--attribute` to `--aria-label` (like `description-list--attribute='aria-label="Default example"'` to `description-list--aria-label="Default example"`)

mcoker (commented):
Just a couple o' comments, but what you have totally works, too!

thatblindgeye (changes requested):
There's some other files we'll also need to update where a DescriptionList is being used:

- the Card md demo file
- card-template-details handlebars
- card-template-events handlebars
- the Tabs md demo file

I think most of the aria-labels for the current files updates are fine, a nit might be some might be a bit lengthy, but that's sort of the nature of the example naming. I'd be fine with them as-is unless we could quickly tweak the naming - not a blocker, though.

rebeccaalpert:
Should be all set now @mcoker and @thatblindgeye!

thatblindgeye (approved):

Issue #7603, opened 2025-07-02
High contrast borders - Popover


mcoker:
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10790-58284&t=Kd9qGLRo9K8UPc05-1

mcoker:
closed via https://github.com/patternfly/patternfly/pull/7707

PR #7707, opened 2025-07-30
feat(popover): add high contrast borders

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7603
fixes https://github.com/patternfly/patternfly/issues/7613

This will also address calendar month, which is used in a popover https://www.patternfly.org/components/date-and-time/date-picker


srambach (approved):
L 🍿 TM!

lboehling (approved):
lovely! thanks!

PR #7420, opened 2025-03-27
fix(CardTitle): titles over-run boundary for long names w/o breaks

logonoff:
clone of https://github.com/patternfly/react-catalog-view/pull/83 but for CardTitle

fixes https://github.com/patternfly/react-catalog-view/issues/82

before:
image

after:
image

PR #7708, opened 2025-07-31
feat(Button): add focus outline offset

### LINKED ISSUE #6428: Bug - Button - Focus accessibility issue
**Describe the problem**
Focus on the button is difficult to perceive.

https://www.patternfly.org/components/button

**Expected behaviour**
There should be a clear visual focus indicator for all the interactive elements.

**Screenshots**
Before focus:
Screenshot 2024-03-18 at 7 04 00 PM

On Focus:
Screenshot 2024-03-18 at 7 04 31 PM

**What is your environment?**
 - OS: Mac OS Sonoma 14.3
 - Browser: Chrome
 - Version: 122.0.6261.129 (Official Build) (x86_64)

**Any other information?**

kmcfaul:
Closes #6428 

@srambach Is there a particular token we want to use for the outline offset? I looked at other instances of the offset and didn't notice a standard so went with a small value to start.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -765,7 +765,7 @@
     text-decoration-line: var(--#{$button}--hover--TextDecorationLine);
     text-decoration-style: var(--#{$button}--hover--TextDecorationStyle);
     text-decoration-color: var(--#{$button}--hover--TextDecorationColor);
-    outline-offset: #{pf-size-prem(2px)}; // TODO - update to --pf-t--global--focus-ring--position--offset when tokens are pulled in
+    outline-offset: var(--pf-t--global--focus-ring--position--offset);
   }
 
   &:active,
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -765,6 +765,7 @@
     text-decoration-line: var(--#{$button}--hover--TextDecorationLine);
     text-decoration-style: var(--#{$button}--hover--TextDecorationStyle);
     text-decoration-color: var(--#{$button}--hover--TextDecorationColor);
+    outline-offset: #{pf-size-prem(2px)};
   }
 
   &:active,
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -765,7 +765,7 @@
     text-decoration-line: var(--#{$button}--hover--TextDecorationLine);
     text-decoration-style: var(--#{$button}--hover--TextDecorationStyle);
     text-decoration-color: var(--#{$button}--hover--TextDecorationColor);
-    outline-offset: #{pf-size-prem(2px)};
+    outline-offset: #{pf-size-prem(2px)}; // TODO - update to --pf-t--global--focus-ring--position--offset when tokens are pulled in
   }
 
   &:active,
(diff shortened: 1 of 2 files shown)


mcoker:
@kmcfaul there will be tokens for focus outlines coming next quarter probably, so the manual size you have works for now.

srambach (approved):
This looks good afaict, looking cross-browser, light/dark, etc.

mcoker:
Just poking around for a little bit, the titles in this [card demo](https://pf-pr-7708.surge.sh/components/card#actionable-and-selectable) no longer show well (before and after below) due to an overflow issue. We can try and address these, but there will probably be issues like this that pop up considering how common the button in all kinds of layouts.

Screenshot 2025-08-07 at 7 31 26 PM
Screenshot 2025-08-07 at 7 31 36 PM

@lboehling this moves the focus outline 2px outside of the button component. WDYT? Given the upcoming focus state changes, is the recommendation still to move the outline outside of the button? https://github.com/patternfly/patternfly/issues/6428#issuecomment-2004477957

lboehling:
> Just poking around for a little bit, the titles in this [card demo](https://pf-pr-7708.surge.sh/components/card#actionable-and-selectable) no longer show well (before and after below) due to an overflow issue. We can try and address these, but there will probably be issues like this that pop up considering how common the button in all kinds of layouts.
> 
> Screenshot 2025-08-07 at 7 31 26 PM Screenshot 2025-08-07 at 7 31 36 PM
> @lboehling this moves the focus outline 2px outside of the button component. WDYT? Given the upcoming focus state changes, is the recommendation still to move the outline outside of the button? [#6428 (comment)](https://github.com/patternfly/patternfly/issues/6428#issuecomment-2004477957)

The most recent token update pulled in focus-ring specific design tokens! `--pf-t--global--focus-ring--position--offset` is set to 2px 👍

image
 
 There are also global tokens now  added for focus-ring--width & focus-ring--color too if we wanted to use those. 

 
image
image

mcoker:
@logonoff in https://github.com/patternfly/patternfly/pull/7420, you added `overflow: auto;` to `.pf-v6-c-card__title-text`. That's causing an issue with hiding overflow on the title element. Do you remember why that was needed for that fix? I don't see it in the [catalog view PR](https://github.com/patternfly/react-catalog-view/pull/83) that the fix came from, and AFAIK that isn't needed to break long strings of text using `overflow-wrap`.

If @logonoff can confirm that style isn't needed, @kmcfaul do you mind also removing `overflow: auto` from `__title-text`? https://github.com/patternfly/patternfly/blob/9005060274bc955d4d5b6eabd6568c583eeb0f81/src/patternfly/components/Card/card.scss#L315

That should fix the outline focus outline being cut off in [my comment above](https://github.com/patternfly/patternfly/pull/7708#issuecomment-3166376326).

logonoff:
> If @logonoff can confirm that style isn't needed, @kmcfaul do you mind also removing `overflow: auto` from `__title-text`?

I don't remember why we needed `overflow: auto` for that fix, what I can say though is that without the `overflow: auto` style, the bug still appears to be fixed for regular cards:

image

However during testing I noticed that the issue still appears for actionable/selectable cards (the card become scrollable instead of wrapping):

image

But this bug still occurs even with `overflow: auto`, so I think it's safe to remove :smile_cat:

thatblindgeye (approved):

mcoker (approved):
🚀

lboehling (approved):
🌟

Issue #7711, opened 2025-08-01
Bug - Code block - high contrast border update


mcoker:
https://github.com/patternfly/patternfly/pull/7685

We've since added border-width tokens for HC theme that should be used for static elements - let's update code block's pseudo element borders to a regular border.

mcoker:
closed by https://github.com/patternfly/patternfly/pull/7712

PR #7712, opened 2025-08-01
fix(code-block): move border, use hc border width

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7711


srambach (approved):
L 🎃 TM

PR #7713, opened 2025-08-01
chore(textInputGroup): add unique aria labels to examples

### LINKED ISSUE #7289: TextInputGroup - search input group examples should use unique aria-labeling
Since a11y docs will likely include verbiage about using unique aria labeling if there's more than 1 search input on a page, core examples should be updated to provide that unique labeling on inputs/utility buttons.

**Any other information?**

Mash707:
fixes: #7289

### DIFF
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -450,7 +450,7 @@ The React implementation can be found in the [search input](/components/search-i
             {{#> button button--id="submit-advanced-search-expanded-with-autocomplete" button--IsPrimary=true button--IsSubmit=true button--aria-labelledby="submit-advanced-search-expanded-with-autocomplete search-input-group-advanced-search-expanded-with-autocomplete"}}
               Submit
             {{/button}}
-            {{#> button button--id="reset-advanced-search-expanded-with-autocomplete" button--IsLink=true button--IsReset=true button--IsSubmit=true button--aria-labelledby="submit-advanced-search-expanded-with-autocomplete search-input-group-advanced-search-expanded-with-autocomplete"}}
+            {{#> button button--id="reset-advanced-search-expanded-with-autocomplete" button--IsLink=true button--IsReset=true button--IsSubmit=true button--aria-labelledby="reset-advanced-search-expanded-with-autocomplete search-input-group-advanced-search-expanded-with-autocomplete"}}
               Reset
             {{/button}}
           {{/form-actions}}
--- a/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
+++ b/src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
@@ -10,9 +10,9 @@ import './TextInputGroup.css'
 ### Basic
 ```hbs
 {{#> text-input-group text-input-group--id="text-input-group-basic"}}
-  {{#> text-input-group-main text-input-group-main--attribute = 'aria-label = "Basic"'}}
+  {{#> text-input-group-main}}
     {{#> text-input-group-text}}
-      {{> text-input-group-text-input}}
+      {{> text-input-group-text-input text-input-group-text-input--aria-label="Basic"}}
     {{/text-input-group-text}}
   {{/text-input-group-main}}
 {{/text-input-group}}
@@ -21,9 +21,9 @@ import './TextInputGroup.css'
 ### Disabled
 ```hbs
 {{#> text-input-group text-input-group--id="text-input-group-disabled" text-input-group--IsDisabled="true"}}
-  {{#> text-input-group-main text-input-group-main--attribute = 'aria-label = "Disabled"'}}
+  {{#> text-input-group-main}}
     {{#> text-input-group-text}}
-      {{> text-input-group-text-input}}
+      {{> text-input-group-text-input text-input-group-text-input--aria-label="Disabled"}}
     {{/text-input-group-text}}
   {{/text-input-group-main}}
 {{/text-input-group}}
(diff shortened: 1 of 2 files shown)


Mash707:
@thatblindgeye For now I have added aria-labels to `text-input-group-main` but I think we should rather add these to `text-input-group-text-input` since it adds a aria-label by default (this means that there are two aria-labels which I think we should not have). Let me know what you think.

https://github.com/patternfly/patternfly/blob/2147287dde3159a0a3d49097a068065e32051e5c/src/patternfly/components/TextInputGroup/text-input-group-text-input.hbs#L24-L31

Based on the response I'll update the remaining examples.

src/patternfly/components/TextInputGroup/text-input-group--search-input.hbs
```diff
@@ -2,13 +2,26 @@
...
     {{/text-input-group-text}}
   {{/text-input-group-main}}
   {{#if text-input-group--value}}
     {{#> text-input-group-utilities}}
       {{#if text-input-group--search-input--count}}
-        {{> badge badge--modifier="pf-m-read" badge--text=text-input-group--search-input--count}}
+        {{#> badge badge--modifier="pf-m-read"}}
+          <span aria-hidden="true">{{text-input-group--search-input--count}}</span>
+          {{#> screen-reader}}
+            {{#if text-input-group--search-input--IsNavigable}}
```
thatblindgeye, line 14: Looks like the "Search input group, match with navigable options" doesnt have this passed into it

thatblindgeye (changes requested):
One file comment below, and just the following:

- For the "Search input group, advanced search expanded" and "Search input group, advanced search expanded with autocomplete" examples, can we just tweak the "Submit" and "Reset" buttons in the examples so those have unique accessible names? We can keep it simple and apply unique ID's to the buttons themselves, then give each button an aria-labelledby that points to itself and the example heading (in that order). The end result should be an accessible name of "Submit Search input group, advanced search expanded" and so on.

thatblindgeye (approved):

src/patternfly/components/TextInputGroup/text-input-group--search-input.hbs
```diff
@@ -2,13 +2,26 @@
   {{#> text-input-group-main text-input-group-main--modifier="pf-m-icon"}}
     {{#> text-input-group-text}}
       {{> text-input-group-icon}}
-      {{> text-input-group-text-input text-input-group-text-input--aria-label="Search input"}}
+        {{#if text-input-group--search-input--aria-label}}
+          {{> text-input-group-text-input text-input-group-text-input--aria-label=text-input-group--search-input--aria-label }}
+        {{else}}
+          {{> text-input-group-text-input text-input-group-text-input--aria-label="Search input" }}
+        {{/if}}
```
mcoker: FWIW we do have a `ternary` helper for handlebars that would make this less code https://github.com/patternfly/patternfly/blob/9005060274bc955d4d5b6eabd6568c583eeb0f81/scripts/helpers.mjs#L67-L75

```suggestion
      {{> text-input-group-text-input text-input-group-text-input--aria-label=(ternary text-input-group--search-input--aria-label text-input-group--search-input--aria-label 'Search input' }}
```

src/patternfly/components/TextInputGroup/text-input-group--search-input.hbs
```diff
@@ -26,7 +39,21 @@
...
+          button--aria-label=text-input-group--search-input--clear-button-aria-label
+        }}
+      {{else}}
+        {{> button
+          button--IsPlain=true
+          button--IsIcon=true
+          button--icon="times fa-fw"
+          button--aria-label="Clear input"
+        }}
+      {{/if}}
```
mcoker: I'd use a ternary here, too

```suggestion
      {{> button
        button--IsPlain=true
        button--IsIcon=true
        button--icon="times fa-fw"
        button--aria-label=(ternary text-input-group--search-input--clear-button-aria-label text-input-group--search-input--clear-button-aria-label "Clear input")
      }}
```

src/patternfly/components/TextInputGroup/examples/TextInputGroup.md
```diff
@@ -146,22 +146,22 @@ For the purposes of this example, the `TextInputGroup` is contained in a wrapper
...
 ```
 
 ### Search input group, match with navigable options
 ```hbs
-{{> text-input-group--search-input text-input-group--id="text-input-group-search-input-group-match-with-navigable-options" text-input-group-text-input--placeholder="Find by name" text-input-group--value="John Doe" text-input-group--search-input--count="1 / 3" text-input-group--search-input--IsFirstMatch="true"}}
+{{> text-input-group--search-input text-input-group--id="text-input-group-search-input-group-match-with-navigable-options" text-input-group-text-input--placeholder="Find by name" text-input-group--value="John Doe" text-input-group--search-input--count="1 / 3" text-input-group--search-input--IsFirstMatch="true" text-input-group--search-input--IsNavigable="true" text-input-group--search-input--aria-label="Search input group match with navigable options" text-input-group--search-input--clear-button-aria-label="Clear search input group match with navigable options"}}
```
mcoker: The tiniest nittiest nit, just use `true` instead of `"true"` for setting something to true, unless you need to use the string "true" for some reason. In this case it should just be `true`. Our code is a little inconsistent that way, but we're trying to write all new code omitting the quotes unless you're passing a string to the parameter/attribute.

```suggestion
{{> text-input-group--search-input text-input-group--id="text-input-group-search-input-group-match-with-navigable-options" text-input-group-text-input--placeholder="Find by name" text-input-group--value="John Doe" text-input-group--search-input--count="1 / 3" text-input-group--search-input--IsFirstMatch="true" text-input-group--search-input--IsNavigable=true text-input-group--search-input--aria-label="Search input group match with navigable options" text-input-group--search-input--clear-button-aria-label="Clear search input group match with navigable options"}}
```

mcoker (approved):
Looks great to me! Just a few small tweaks if you don't mind.

mcoker (approved):
🚀

PR #7716, opened 2025-08-01
chore(skeleton): added screen reader text

### LINKED ISSUE #7290: Skeleton - at least one skeleton per example should include screen reader text
This goes along with the a11y docs for the component. Basically a screen reader text element should be rendered inside of `pf-v6-c-skeleton` in one of the skeletons per example (to go along with the [when to apply screen reader text](https://www.patternfly.org/components/skeleton/accessibility#when-to-apply-screen-reader-text) verbiage from a11y docs (each example being a main section of the page).

One suggestion would be to utilize the example name, e.g. "Loading default example content".

**Any other information?**

Mash707:
Fixes: #7290


src/patternfly/demos/Skeleton/examples/Skeleton.md
```diff
@@ -12,6 +12,7 @@ section: components
 {{#* inline "page-template-section"}}
   {{#> page-main-section}}
     {{#> gallery gallery--modifier="pf-m-gutter"}}
+      <span class="pf-v6-screen-reader">Loading content</span>
```
Mash707: @thatblindgeye I am not sure if this is the correct place for the span.

src/patternfly/components/Skeleton/examples/Skeleton.md
```diff
@@ -11,12 +11,14 @@ import './Skeleton.css'
 ### Default
 ```hbs
 {{#> skeleton}}
+    <span class="pf-v6-screen-reader">Loading default content</span>
```
mcoker: There is a hbs partial that will generate this element for you - [`screen-reader.hbs`](https://github.com/patternfly/patternfly/blob/main/src/patternfly/utilities/Accessibility/screen-reader.hbs)

```suggestion
    {{#> screen-reader}}Loading default content{{/screen-reader}}
```

src/patternfly/demos/Skeleton/examples/Skeleton.md
thatblindgeye: We'd want this text to be a direct child of a skeleton. What we could do is allow a way to pass some custom screen reader text inside the `card-skeleton.hbs` file, and then update this demo to pass that into the `card-skeleton` instances. We could even have it fallback to some default "Loading card content" text or similar. For this particular demo, whether each card should actually have their own screen reader text is iffy. In certain scenarions it may make sense to do that... for here, we could keep it simple and just apply a screen reader text to the first card.

@mcoker do we intend on ever using the `table-skeleton` at all? We have a file for it but it's not used anywhere. If we do we'd want to make a similar update there.

mcoker: @thatblindgeye not that I know of. You're welcome to remove that file if you'd like. I see it's referencing `{{#> skeleton-element}}` which doesn't appear to exist. FYI, it would be defined either as a file named `skeleton-element.hbs` OR sometimes we have "inline" partials where we define the partial name and code in some other file, but I searched the repo for "skeleton-element" and didn't see an inline partial anywhere.

If skeletons should always have some text, should we just have some logic in skeleton.hbs that's like this, then we'd just specify `skeleton--HasNoText=true` for ones we want to omit (like all but the first card in that demo)

```hbs
{{#unless skeleton--HasNoText}}
  {{#if skeleton--text}}
    {{skeleton--text}}
  {{else}}
    Loading content
  {{/if}}
{{/unless}}
```

src/patternfly/components/Skeleton/examples/Skeleton.md
Mash707: Will the make the changes.

src/patternfly/demos/Skeleton/examples/Skeleton.md
thatblindgeye: We could do that. Would be handy when there's only one section of a skeleton being used in an example/demo, though we'd just need to make sure to provide more unique text when there's multiple sections that need some sort of content

src/patternfly/demos/Skeleton/card-skeleton.hbs
```diff
@@ -2,6 +2,9 @@
   {{#> card-body}}
       {{#> l-flex l-flex--modifier="pf-m-column pf-m-spacer-md"}}
         {{#> skeleton}}
+          {{#if isFirstCard}}
```
thatblindgeye: The naming of this handlebars attribute is pretty close. In a sense they can be similar in naming structure to our React props, one difference being that we'll prefix the attribute with the component and a double hyphen, and the attribute will be pascal case, so something like

```suggestion
          {{#if card-skeleton--IsFirstCard}}
```

That said, implementing the feedback from Coker where the `skeleton.hbs` file has a default value could be a nice to have, though that could be where you can get into how complex handlebars can be/feel. I wouldn't consider implementing that a priority and would be slightly out of scope so if you want to pass on it that's fine.

If you want to make an attempt, ideally the logic should be:

- `skeleton.hbs` has logic either where unless `skeleton--HasNoText` or if `skeleton-IsFirstSkeleton` (or something), then we either render whatever is passed to `skeleton--text` or a default string of `Loading conternt". We could also just use `@partial-block` instead of `skeleton--text`, though I'd leave that up to Coker if he has a preference as I don't
- From there, within `card-skeletong.hbs`, the logic you have now would have to be tweaked slightly if we went with the `skeleton--text` approach, possibly utilizing the handblebars ternary function we have and passing that into `skeleton--text` to check if `card-skeleton--IsFirstCard` is true, passing in the loading text here if so or an empty string if not. Coker may have a less complicated approach, though.

thatblindgeye (commented):
Would you mind also deleting the table-skeleton file? Based on Coker's comments above it doesn't sound necessary and is just another thing we have to worry about in the future.

thatblindgeye (approved):

mcoker (approved):
🔥🔥🔥

Thanks @Mash707!!

Issue #7376, opened 2025-03-04
Bug - Tabs - missing inset value to correctly align tabs with PageSection gutter


rhamilto:
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

When using Tabs directly inside PageBody with PageSections as siblings, there is no existing inset value that will properly align the first tab's underline with the PageBody's gutter.

**How do you reproduce the problem?**
See https://github.com/openshift/console/pull/14805#discussion_r1973768720.

**Expected behavior**
There should be an inset value that does this.

**Is this issue blocking you?**
See https://github.com/openshift/console/pull/14805#discussion_r1973768720.

mcoker:
@rhamilto sorry for the delay in fixing this issue. We updated the spacers for `<Tabs usePageInsets />` so it should be aligned properly now. This change will go out in our upcoming release in early October.

PR #7718, opened 2025-08-01
feat(Page): add high contrast border

### LINKED ISSUE #7590: High contrast borders - Page
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10994-65000&t=Kd9qGLRo9K8UPc05-1

kmcfaul:
Closes #7590.

@mcoker / @srambach Would this approach of updating the existing border on the page main container work for this use case or would the `before`/`after` approach still be preferred? I had initially run into some z-indexing issues on the border & page content as well as seeing the two borders concurrently, before I changed track and updated the current border instead because it already existed. But I can definitely keep adjusting the `after` block if that's still preferred, I think I just need to disable the current border and then resolve the z-indexes.

With the current approach, there is also a 2px difference in the border widths (4px on the normal border and 2px on the high contrast border) that's causing a small layout shift and I'm unsure the best way to solve it. Maybe adding additional padding when high contrast is on?

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -78,7 +78,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__main-container--BorderRadius: var(--pf-t--global--border--radius--medium);
   --#{$page}__main-container--BorderWidth: var(--pf-t--global--border--width--page--default);
   --#{$page}__main-container--BorderColor: var(--pf-t--global--border--color--page--default); 
-  --#{$page}__main-container--Padding--offset: var(--#{$page}__main-container--BorderWidth);
+  --#{$page}__main-container--Padding--offset: 4px;
   --#{$page}__main-container--xs--AlignSelf: stretch;
   --#{$page}__main-container--xs--BorderRadius: 0;
   --#{$page}__main-container--xs--MarginInlineStart: 0;
@@ -371,6 +371,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   flex-direction: column;
   align-self: var(--#{$page}__main-container--AlignSelf);
   max-height: var(--#{$page}__main-container--MaxHeight);
+  padding:  calc(var(--#{$page}__main-container--Padding--offset) - var(--#{$page}__main-container--BorderWidth));
   margin-inline-start: var(--#{$page}__main-container--MarginInlineStart);
   margin-inline-end: var(--#{$page}__main-container--MarginInlineEnd);
   background: var(--#{$page}__main-container--BackgroundColor);
@@ -383,12 +384,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
     --#{$page}__main-container--MarginInlineEnd: var(--#{$page}__main-container--xs--MarginInlineEnd);
     --#{$page}__main-container--MaxHeight: var(--#{$page}__main-container--xs--MaxHeight);
     --#{$page}__main-container--BorderRadius: var(--#{$page}__main-container--xs--BorderRadius);
-  }
-
-  @media (prefers-contrast: more) {
-    --#{$page}__main-container--Padding--offset: #{pf-size-prem(4px)};
-
-    padding: calc(var(--#{$page}__main-container--Padding--offset) - var(--#{$page}__main-container--BorderWidth));
   }
 }
 
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -78,6 +78,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__main-container--BorderRadius: var(--pf-t--global--border--radius--medium);
   --#{$page}__main-container--BorderWidth: #{pf-size-prem(4px)}; // TODO Change to be a page outline token
   --#{$page}__main-container--BorderColor: var(--#{$page}__main-container--BackgroundColor); // TODO Border should match the background to blend in - change to be a page outline token
+  --#{$page}__main-container--BorderOffset: var(--#{$page}__main-container--BorderWidth);
   --#{$page}__main-container--xs--AlignSelf: stretch;
   --#{$page}__main-container--xs--BorderRadius: 0;
   --#{$page}__main-container--xs--MarginInlineStart: 0;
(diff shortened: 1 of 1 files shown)


kmcfaul:
Updated to prevent the layout shift - 

- Other page elements will account for the difference in border width when calcing their padding.
- There is now a small amount (2px) of page-main-container padding only in high contrast mode to compensate for the border width difference in layout.

lboehling (approved):
🌟

src/patternfly/components/Page/page.scss
```diff
@@ -78,6 +78,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__main-container--BorderRadius: var(--pf-t--global--border--radius--medium);
   --#{$page}__main-container--BorderWidth: #{pf-size-prem(4px)}; // TODO Change to be a page outline token
   --#{$page}__main-container--BorderColor: var(--#{$page}__main-container--BackgroundColor); // TODO Border should match the background to blend in - change to be a page outline token
+  --#{$page}__main-container--BorderOffset: var(--#{$page}__main-container--BorderWidth);
```
srambach: Since there isn't a css attribute called border-offset, wdyt about calling this `--#{$page}__main-container--Padding--offset`? It looks like we've used that pattern once or twice before. @mcoker ?

src/patternfly/components/Page/page.scss
```diff
@@ -383,6 +384,14 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
     --#{$page}__main-container--MaxHeight: var(--#{$page}__main-container--xs--MaxHeight);
     --#{$page}__main-container--BorderRadius: var(--#{$page}__main-container--xs--BorderRadius);
   }
+
+  @media (prefers-contrast: more) {
+    --#{$page}__main-container--BorderWidth: var(--pf-t--global--border--width--strong);
+    --#{$page}__main-container--BorderColor: var(--pf-t--global--border--color--high-contrast);
+    --#{$page}__main-container--Padding--offset: #{pf-size-prem(4px)};
```
srambach: @lboehling @mcoker We have a comment[ in the code](https://github.com/kmcfaul/patternfly/blob/79019932142721006596fc4e9430642be5be1158/src/patternfly/components/Page/page.scss#L79) saying that we should create page border outline width and color tokens. Is this a good time to do that? Then it could adjust in the high-contrast tokens. Trying to see if we could get rid of this media query! Since the offset is 4px either way, maybe that would do it?

mcoker: I like that! @lboehling do you have any issues with adding that to figma?

mcoker: That would also let us use that token value to offset any component padding/insets that might need to take into consideration that border width to align properly with other page chrome elements. I haven't looked too much into it, but I'm wondering if that might help resolve this issue with tabs not aligning properly with the page chrome gutter https://github.com/patternfly/patternfly/issues/7376

mcoker (changes requested):
@kmcfaul design added tokens for the page border color (`--pf-t--global--border--color--page--default`) and border width (`--pf-t--global--border--width--page--default`) in https://github.com/patternfly/patternfly/pull/7761. 

Would you mind rebasing and dropping that token in here where appropriate? I'm hoping it can get rid of the custom `@media (prefers-contrast)` styles like @srambach mentioned in https://github.com/patternfly/patternfly/pull/7718/files#r2289094152

kmcfaul:
@mcoker Rebased and updated to use the new tokens, but the media-query may still be required to account for the border width/padding issue. And speaking of that, I'm seeing a small layout shift again even though the calculated sizes should be the same as the previous implementation. Do you have any ideas for addressing this shift?

mcoker:
> And speaking of that, I'm seeing a small layout shift again even though the calculated sizes should be the same as the previous implementation. Do you have any ideas for addressing this shift?

@kmcfaul are you referring to a shift in either the subnav or breadcrumb container? That's where I'm seeing a shift. Assuming that's the case, I believe it's because of a scoping issue with the change of `--#{$page}__main-container--Padding--offset` in this block https://github.com/patternfly/patternfly/blob/1cdf8180ac2c153a6efe3ed47d8430357f7a8300/src/patternfly/components/Page/page.scss#L388-L392

If you update the value of a CSS var, you'll need to update it in the same scope of where it's declared/used. Meaning you'd want to move the `@media` query up to the same scope as where all of the component variables that use `--#{$page}__main-container--Padding--offset` are defined. What I'm not sure on is whether a change within a `@media` query in a scope will also apply to other vars in that same scope but outside of the `@media` query, so maybe give it a shot and let me know if it doesn't work?

The reason the main-section padding doesn't shift is because `--#{$page}__main-container--Padding--offset` is used within the same scope as the change you made on the main container element. https://github.com/patternfly/patternfly/blob/1cdf8180ac2c153a6efe3ed47d8430357f7a8300/src/patternfly/components/Page/page.scss#L502-L503

kmcfaul:
Update - the layout shift must be something with the examples page / seeing the various inline page examples. In a full page example in a new tab, there is no layout shift. So this PR should be good to go unless anyone finds any other gotchas.

src/patternfly/components/Page/page.scss
```diff
@@ -76,8 +76,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__main-container--MarginInlineStart: var(--#{$page}--inset);
   --#{$page}__main-container--MarginInlineEnd: var(--#{$page}--inset);
   --#{$page}__main-container--BorderRadius: var(--pf-t--global--border--radius--medium);
-  --#{$page}__main-container--BorderWidth: #{pf-size-prem(4px)}; // TODO Change to be a page outline token
-  --#{$page}__main-container--BorderColor: var(--#{$page}__main-container--BackgroundColor); // TODO Border should match the background to blend in - change to be a page outline token
+  --#{$page}__main-container--BorderWidth: var(--pf-t--global--border--width--page--default);
+  --#{$page}__main-container--BorderColor: var(--pf-t--global--border--color--page--default); 
+  --#{$page}__main-container--Padding--offset: 4px;
```
srambach:
```suggestion
  --#{$page}__main-container--BorderWidth: var(--pf-t--global--border--width--page--default);
  --#{$page}__main-container--BorderColor: var(--pf-t--global--border--color--page--default); 
```

I think that we can just make this change and revert all the other changes to the file. I don't think we need the offset because subtracting the border width from the page section(s) padding (which was already done) will give the correct spacing and avoid the 2px gap between the high-contrast border and a section with a secondary background. There will be a slight vertical shift as the top border goes from 4 to 2px, but I think that's acceptable as we currently don't try to vertically align the main area with anything else.

If we want to avoid the vertical shift, we can select the first page group/section and adjust, but it's a bit messy and perhaps not worth it.

@lboehling @mcoker

lboehling: i'm good with this 👍 thank you!

lboehling: follow up Q -- we had discussed updating `--pf-t--global--border--width--page--default` & `--pf-t--global--border--color--page--default` to use "main" instead of "page" do we still want to do that? @mcoker @srambach

srambach: I don't hate the idea of that 😄 But I don't consider it a blocker to this.

mcoker: I like it. We could then use page areas as names - "header", "sidebar", possibly "nav" if needed, and "main". Not a blocker to this PR, which I'd like to get merged ASAP but we'd want to make that change this sprint before code freeze.

lboehling: sounds good -- i'll make that change in figma so it's ready the next time we export the tokens.

lboehling (approved):
looks good!

mcoker (approved):
LGTM!

srambach (approved):
L 🍾 TM!

PR #7723, opened 2025-08-05
feat(menu-toggle): add high-contrast border

### LINKED ISSUE #7620: High contrast borders - Menu toggle

srambach:
Fixes #7620 

Uses existing pseudoelement to add borders for plain variant.

@lboehling One question - this exposed the difference in padding for the Count variant - just want to make sure this is correct?
image

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -13,7 +13,7 @@
   --#{$menu-toggle}--BackgroundColor: var(--pf-t--global--background--color--control--default);
   --#{$menu-toggle}--BorderRadius: var(--pf-t--global--border--radius--small);
   --#{$menu-toggle}--BorderColor: var(--pf-t--global--border--color--default);
-  --#{$menu-toggle}--BorderWidth: var(--pf-t--global--border--width--action--default);
+  --#{$menu-toggle}--BorderWidth: var(--pf-t--global--border--width--control--default);
   --#{$menu-toggle}--border--ZIndex: var(--pf-t--global--z-index--xs); // add z-index for toggle border to render above other borders
   --#{$menu-toggle}--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
   --#{$menu-toggle}--TransitionDuration: var(--pf-t--global--motion--duration--fade--short);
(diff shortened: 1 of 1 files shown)


lboehling:
> @lboehling One question - this exposed the difference in padding for the Count variant - just want to make sure this is correct? image

I think this is fine @srambach

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -138,7 +138,10 @@
   --#{$menu-toggle}--m-plain--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--plain--default);
   --#{$menu-toggle}--m-plain--Color: var(--pf-t--global--icon--color--regular);
   --#{$menu-toggle}--m-plain--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
-  --#{$menu-toggle}--m-plain--BorderColor: transparent;
+  --#{$menu-toggle}--m-plain--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$menu-toggle}--m-plain--BorderWidth: var(--pf-t--global--high-contrast--border--width--action--default);
```
lboehling: we just added in new border width tokens for `--pf-t--global--border--width--action--plain--default/hover/clicked` that can be used here instead. this will remove the border on the default state of the plain menu toggle

lboehling: The new border width tokens for HC theme also adjust the border--width--control--hover value as well, so the hover states on the default menu toggles will also increase by a pixel on hover. No action needed on this i believe as long as those tokens are already applied, think we just need those new tokens pulled in to see that change.

lboehling (changes requested):

src/patternfly/components/MenuToggle/menu-toggle.scss
```diff
@@ -138,7 +138,10 @@
   --#{$menu-toggle}--m-plain--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--plain--default);
   --#{$menu-toggle}--m-plain--Color: var(--pf-t--global--icon--color--regular);
   --#{$menu-toggle}--m-plain--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
-  --#{$menu-toggle}--m-plain--BorderColor: transparent;
+  --#{$menu-toggle}--m-plain--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$menu-toggle}--m-plain--BorderWidth: var(--pf-t--global--border--width--action--default);
```
lboehling: small nit -- but adding action--_plain_--[default/hover/clicked] will give the correct styles here for the plain menu toggles!

lboehling (approved):
lovely!

mcoker (approved):
L🦦TM!

Issue #7367, opened 2025-02-24
Menu - ability to make group titles sticky


mcoker:
Requested on the patternfly slack:

> Hey PF team. I am working on a design that has a grouped option menu (PF component - [Titled groups of items](https://www.patternfly.org/components/menus/menu#titled-groups-of-items)).
>
> There is a pain point in a long list of options: When users scroll to an option, they may have no clue of knowing which group the option belongs to, because the group name has been slipped out of sight.
>
> In order to clearly identify the group where each option is located, is it possible to pin the corresponding group name at the top when sliding to options in a group? This way, no matter which option to slide to, the user can always identify the group by the top group name.
>
> I attached a short video as an example. I knew this is not a PF standard component so far. But I am curious about if we can support the similar thing? Thanks!

https://github.com/user-attachments/assets/b305a687-6b45-49ea-a1c3-8dc74d809f62

This will require https://github.com/patternfly/patternfly/issues/7256 as grouped menus are not currently scrollable.

PR #7725, opened 2025-08-05
feat(menu): add high-contrast border

### LINKED ISSUE #7600: High contrast borders - Menu
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10790-58159&t=Kd9qGLRo9K8UPc05-1

srambach:
Fixes #7600 

Added border to existing pseudoelement on the menu item
Added border to a new ::after on the menu - used the pseudo because otherwise the drilldown calculation of the size was off by the amount of the border.

Note - The nav with drilldown menu items have a slight border radius that touches the menu's border. I'm not sure if this is right or not.
image

### DIFF
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -225,6 +225,7 @@
   // TODO: Explore this fallback method
   // padding-block-start: var(--#{$menu}--PaddingBlockStart, var(--#{$menu}--PaddingBlock));
   // padding-block-end: var(--#{$menu}--PaddingBlockEnd, var(--#{$menu}--PaddingBlock));
+  box-sizing: content-box;
   display: grid;
   row-gap: var(--#{$menu}--RowGap);
   width: var(--#{$menu}--Width);
--- a/src/patternfly/components/Menu/menu.scss
+++ b/src/patternfly/components/Menu/menu.scss
@@ -46,9 +46,9 @@
   // * Menu list item
   --#{$menu}__list-item--Color: var(--pf-t--global--text--color--regular);
   --#{$menu}__list-item--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
-  --#{$menu}__list-item--BorderWidth: var(--pf-t--global--high-contrast--border--width--control--default);
-  --#{$menu}__list-item--BorderColor: transparent;
-  --#{$menu}__list-item--hover--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$menu}__list-item--BorderWidth: var(--pf-t--global--border--width--action--plain--default);
+  --#{$menu}__list-item--hover--BorderWidth: var(--pf-t--global--border--width--action--plain--hover);
+  --#{$menu}__list-item--BorderColor: var(--pf-t--global--border--color--high-contrast);
   --#{$menu}__list-item--TransitionDuration: var(--pf-t--global--motion--duration--fade--short);
   --#{$menu}__list-item--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--default);
   --#{$menu}__list-item--TransitionProperty: background-color;
@@ -225,7 +225,6 @@
   // TODO: Explore this fallback method
   // padding-block-start: var(--#{$menu}--PaddingBlockStart, var(--#{$menu}--PaddingBlock));
   // padding-block-end: var(--#{$menu}--PaddingBlockEnd, var(--#{$menu}--PaddingBlock));
-  box-sizing: content-box;
   display: grid;
   row-gap: var(--#{$menu}--RowGap);
   width: var(--#{$menu}--Width);
@@ -299,6 +298,7 @@
 
     :where(.#{$menu}) {
       padding: 0;
+      border: 0;
     }
 
     &.pf-m-drilled-in {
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Menu/menu.scss
```diff
@@ -10,6 +10,8 @@
   --#{$menu}--BackgroundColor: var(--pf-t--global--background--color--floating--default);
   --#{$menu}--BoxShadow: var(--pf-t--global--box-shadow--md);
   --#{$menu}--Color: var(--pf-t--global--text--color--regular);
+  --#{$menu}--BorderWidth: var(--pf-t--global--high-contrast--border--width--control--default);
```
mcoker: should this use the box token?

```suggestion
  --#{$menu}--BorderWidth: var(--pf-t--global--border--width--box--default);
```

src/patternfly/components/Menu/menu.scss
```diff
@@ -235,6 +242,15 @@
...
   // stylelint-enable declaration-no-important
 
+  &::after {
+    position: absolute;
+    inset: 0;
+    pointer-events: none;
+    content: "";
+    border: var(--#{$menu}--BorderWidth) solid var(--#{$menu}--BorderColor);
+    border-radius: inherit;
+  }
```
mcoker: What's the lift to fix the drilldown stuff if we move this to a regular border? I'd expect menu to have issues with children that overlap the border. I know we have a feature request to make [group titles sticky](https://github.com/patternfly/patternfly/issues/7367) that would mess up this border unless we figured out a z-index thing for this pseudo element.


https://github.com/user-attachments/assets/578e5380-ed82-47af-a849-8b1bb991916a

mcoker (commented):
Left a comment about the pseudo border. Also @lboehling should there be a border for selected state? [Single selected](https://pf-pr-7725.surge.sh/components/menus/menu/#option-single-select) and [multi-selected](https://pf-pr-7725.surge.sh/components/menus/menu/#option-multi-select)

srambach:
Updated to put borders onto the menu element rather than the pseudo. This increases the size of the menu, so the React calculation of the menu height will need to take this into consideration (until then, you'll see the last menu item on a drilled in menu without a bottom border, as it's being cut off)

This also changes the menu to use `box-sizing: content-box;` because otherwise the drilled in menus, which are shifted 100% are not moved quite far enough and you'll see an increasing gap on the right side of the menu items.

Regression test shows 1px height change where I had to increase the manually specified size to accommodate the new borders. There's also a shift in the modified-width drilldown expansion caret, which I think results from the menu border but I didn't increase the manual width (I can if you like and then I think the regression would pass).
https://drive.google.com/file/d/1ow4uFRTGH8BgrVh6aAYI6qQRW2lLysFk/view?usp=sharing

This will require a fix in React to account for the borders when calculating the height. (in MenuContent.tsx)

@mcoker @thatblindgeye

lboehling:
> Left a comment about the pseudo border. Also @lboehling should there be a border for selected state? [Single selected](https://pf-pr-7725.surge.sh/components/menus/menu/#option-single-select) and [multi-selected](https://pf-pr-7725.surge.sh/components/menus/menu/#option-multi-select)

No border needed for selected state since the menu items have other indications for its selected state (check mark) that meet the color contrast requirements.

lboehling (approved):
Design looks good to me!

src/patternfly/components/Menu/menu.scss
```diff
@@ -44,6 +46,9 @@
   // * Menu list item
   --#{$menu}__list-item--Color: var(--pf-t--global--text--color--regular);
   --#{$menu}__list-item--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
+  --#{$menu}__list-item--BorderWidth: var(--pf-t--global--high-contrast--border--width--control--default);
+  --#{$menu}__list-item--BorderColor: transparent;
+  --#{$menu}__list-item--hover--BorderColor: var(--pf-t--global--border--color--high-contrast);
```
mcoker: Should we update this to set the color by default and use the action--plain tokens?

```suggestion
  --#{$menu}__list-item--BorderWidth: var(--pf-t--global--border--width--action--plain--default);
  --#{$menu}__list-item--BorderColor: var(--pf-t--global--border--color--high-contrast);
  --#{$menu}__list-item--hover--BorderWidth: var(--pf-t--global--border--width--action--plain--hover);
```

mcoker:
@srambach 

> This also changes the menu to use box-sizing: content-box; because otherwise the drilled in menus, which are shifted 100% are not moved quite far enough and you'll see an increasing gap on the right side of the menu items.

I could be wrong, but this looks like it may be due to the borders on the nested menu, which don't look to be disabled. When I disable the borders on nested menus, looks like the menus position correctly without modifying `box-sizing`.

I wonder if that might mean the height adjustments aren't required, too?

> There's also a shift in the modified-width drilldown expansion caret, which I think results from the menu border but I didn't increase the manual width (I can if you like and then I think the regression would pass).

Looking at the examples, the PR has `--pf-v6-c-menu--Width: 352px;` and core staging has `--pf-v6-c-menu--Width: 350px;`

mcoker (commented):
Also left a separate comment about the drilldown updates.

mcoker (approved):
Greatness

PR #7728, opened 2025-08-05
feat(tabs): add high-contrast border

### LINKED ISSUE #7623: High contrast borders - Tabs

srambach:
Fixes #7623 

Adds a border to the tab__link where the background highlighting is.

[Figma link](https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10790-58308&t=5lTVMAeS41XYuMYJ-1)

### DIFF
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -66,6 +66,10 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__link--FontSize: var(--pf-t--global--font--size--sm);
   --#{$tabs}__link--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
   --#{$tabs}__link--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$tabs}__link--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
+  --#{$tabs}__link--BorderColor: transparent;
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Tabs/tabs.scss
```diff
@@ -66,6 +66,10 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__link--FontSize: var(--pf-t--global--font--size--sm);
   --#{$tabs}__link--BackgroundColor: var(--pf-t--global--background--color--action--plain--default);
   --#{$tabs}__link--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$tabs}__link--BorderWidth: var(--pf-t--global--high-contrast--border--width--regular);
```
lboehling: could may be use the new `--pf-t--global--border--width--action--plain--hover` token here instead? otherwise, this all looks great!

srambach:
I updated to the token @lboehling suggested.
I did leave this with the new border on the nav link element and keep it on the element and switch the color from transparent to the hi-c border color rather than (as @mcoker suggested) to adjust the width on the pseudo element as we've done in other components. The other pseudoelements are already in use and still need to be used as is. Moving to use the `__item-text` caused all sorts of complication, not the least of which is that would not encompass `__item-icon`.

mcoker (approved):
LGTM! If this causes any issues, we could probably refactor it to use `:has()` and put the existing borders on the parent `<li>` for these tab items.

PR #7733, opened 2025-08-06
fix(global): set color-scheme light by default

### LINKED ISSUE #7720: Bug - Checkbox - dark background without .pf-v6-theme-dark when "color-scheme" meta tag is used
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/color-scheme

Summary: A website which supports light and dark themes is suggested to have in its `<head>`

```html
<meta name="color-scheme" content="light dark" />
```

If the user's system has a preference for dark theme, the presence of the tag above will automatically set the CSS `color-scheme: dark` on every component. However, Patternfly's default theme is light (true when the class `pf-v6-theme-dark` is _not_ applied to `html`).

The `color-scheme: dark` style will cause checkboxes (and possibly other input elements) to have a dark background when unchecked. This looks ugly against Patternfly's default light theme.

Create a new minimal Patternfly project, e.g.

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="UTF-8" />
		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
		<meta name="color-scheme" content="light dark" />
		<title>Just a Checkbox</title>
	</head>
	<body>
		<div id="app"></div>
		<script type="module" src="/src/index.tsx"></script>
	</body>
</html>
```

```jsx
import "@patternfly/react-core/dist/styles/base.css";
import { Page, Checkbox, PageSection } from "@patternfly/react-core";

export function App() {
	return (
		<Page>
			<PageSection>
				<Checkbox id="checkbox" />
			</PageSection>
		</Page>
	);
}
```

**Expected behavior**

Checkbox and page is in light theme

Image

No

**Screenshots**

Actual behavior:

Image

**What is your environment?**

Tested on Chromium, Firefox, and on iOS.

Unimportant, but sure: https://github.com/FNNDSC/orthanc-patient-list/, next week

**Any other information?**

Workaround (and possibly a fix):

```css
html:not(.pf-v6-theme-dark) input {
	color-scheme: light;
}
```

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7720

Since designs intent for default theme is a light theme, and we set `color-scheme` explicitly in the CSS in dark theme, this seems like a safe update to make.

Though ultimately for the type of issue reported in https://github.com/patternfly/patternfly/issues/7720, I'd recommend just not including the `<meta>` tag.

### DIFF
--- a/src/patternfly/base/normalize.scss
+++ b/src/patternfly/base/normalize.scss
@@ -122,6 +122,10 @@
     cursor: pointer;
   }
 
+  :where(:root) {
+    color-scheme: light;
+  }
+
   :where(.pf-v6-theme-dark) {
     color-scheme: dark;
   }


srambach (approved):
Seems safe to me.

Issue #7614, opened 2025-07-03
High contrast borders - Clipboard copy


mcoker:
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10994-57314&t=DkgfQ5XT3Uibqm6l-1

mcoker:
@lboehling the design calls for a hover state for clipboard copy, but the whole clipboard copy element isn't clickable - only the button/icon is clickable. Should it have a hover state? If so, should the border change trigger only when hovering the button/icon?

<img width="538" height="396" alt="Image" src="https://github.com/user-attachments/assets/9c6368b7-5a2a-4154-85c5-8f5d2c926ba3" />

lboehling:
Good catch -- we can treat this like the plain action buttons then, and just have the border show on hover!

lboehling:
@mcoker I've updated the design spec in figma: https://www.figma.com/design/wKcOq7IrzfL1gEK2mocd6r/Semantic-dimension-border-width----font-weight-adjustment-test?node-id=10994-57314&t=DgdGE14FdvoOnsNd-4

<img width="780" height="472" alt="Image" src="https://github.com/user-attachments/assets/75085ed6-c9c8-42ce-a769-d13bb2d9dc68" />

mcoker:
With the updated design, this should be resolved in https://github.com/patternfly/patternfly/pull/7734

mcoker:
closed via https://github.com/patternfly/patternfly/pull/7734

PR #7734, opened 2025-08-06
feat(Button): add high contrast to plain/link hover+clicked

### LINKED ISSUE #7591: High contrast borders - Button
https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10790-60003&t=Kd9qGLRo9K8UPc05-1

### LINKED ISSUE #7742: Button - increase no-padding variations clickable area
Since we're drawing borders around these elements in HC, it would be great if the border was offset just a bit. In chatting with @lboehling, we decided to also increase the clickable area of the button by the same amount, which we can do with a negative `inset` value of the existing pseudo element that creates the borders.

This is what I we reviewed and agreed looked ok (2px offset)

**Edit 8.20.25** - we decided to omit inline links from this change since this kind of change would likely need to apply to regular links, too, and that can be tricky since regular links (and the span variation of the pf inline link button) can wrap to multiple lines.

```diff
diff --git a/src/patternfly/components/Button/button.scss b/src/patternfly/components/Button/button.scss
index f90499f33..447445462 100644
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -27,6 +27,7 @@
   --#{$button}--TransitionProperty: color, background, border-width, border-color;
   --#{$button}--ScaleX: 1;
   --#{$button}--ScaleY: 1;
+  --#{$button}--after--Inset: 0;
 
   // Hover
   --#{$button}--hover--BackgroundColor: transparent;
@@ -141,6 +142,7 @@
   --#{$button}--m-link--m-inline--hover--Color: var(--pf-t--global--text--color--brand--hover);
   --#{$button}--m-link--m-inline--hover__icon--Color: var(--pf-t--global--icon--color--brand--hover);
   --#{$button}--m-link--m-inline--hover--TextDecorationColor: currentcolor;
+  --#{$button}--m-link--m-inline--after--Inset: calc(#{pf-size-prem(2px)} * -1);
 
   // Plain
   --#{$button}--m-plain--BorderRadius: var(--pf-t--global--border--radius--small);
@@ -172,7 +174,7 @@
   --#{$button}--m-plain--m-no-padding--hover__icon--Color: var(--pf-t--global--icon--color--regular);
   --#{$button}--m-plain--m-no-padding--hover--BackgroundColor: transparent;
   --#{$button}--m-plain--m-no-padding--m-clicked--BackgroundColor: transparent;
-  --#{$button}__progress--Color: var(--#{$button}__icon--Color);
+  --#{$button}--m-plain--m-no-padding--after--Inset: calc(#{pf-size-prem(2px)} * -1);
 
   // Control
   --#{$button}--m-control--BorderRadius: var(--pf-t--global--border--radius--small);
@@ -399,7 +401,7 @@
 
   &::after {
     position: absolute;
-    inset: 0;
+    inset: var(--#{$button}--after--Inset);
     pointer-events: none;
     content: "";
     border: var(--#{$button}--BorderWidth) solid;
@@ -518,6 +520,7 @@
       --#{$button}--hover--TextDecorationColor: var(--#{$button}--m-link--m-inline--hover--TextDecorationColor);
       --#{$button}--m-link--hover--Color: var(--#{$button}--m-link--m-inline--hover--Color);
       --#{$button}--m-link--hover__icon--Color: var(--#{$button}--m-link--m-inline--hover__icon--Color);
+      --#{$button}--after--Inset: var(--#{$button}--m-link--m-inline--after--Inset);
 
       text-align: start;
       white-space: normal;
@@ -667,7 +670,8 @@
       --#{$button}--PaddingInlineStart: var(--#{$button}--m-plain--m-no-padding--PaddingInlineStart);
       --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-plain--m-no-padding--m-small--PaddingInlineEnd);
       --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-plain--m-no-padding--m-small--PaddingInlineStart);
-  
+      --#{$button}--after--Inset: var(--#{$button}--m-plain--m-no-padding--after--Inset);
+
       min-width: var(--#{$button}--m-plain--m-no-padding--MinWidth);
       background: var(--#{$button}--BackgroundColor); // override the ripple background
     }
```

kmcfaul:
Closes #7591
Closes https://github.com/patternfly/patternfly/issues/7614
Closes https://github.com/patternfly/patternfly/issues/7742

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -33,6 +33,7 @@
   // Hover
   --#{$button}--hover--BackgroundColor: transparent;
   --#{$button}--hover--BorderColor: transparent;
+  --#{$button}--hover--BorderColor--high-contrast: var(--pf-t--global--border--color--high-contrast);
   --#{$button}--hover--BorderWidth: var(--pf-t--global--border--width--action--hover);
   --#{$button}--hover--TextDecorationLine: none;
   --#{$button}--hover--TextDecorationStyle: none;
@@ -44,7 +45,6 @@
   --#{$button}--m-clicked--BackgroundColor: transparent;
   --#{$button}--m-clicked--BorderColor: transparent;
   --#{$button}--m-clicked--BorderWidth: var(--pf-t--global--border--width--action--clicked);
-  --#{$button}--m-clicked--BorderWidth--high-contrast: var(--pf-t--global--border--width--strong);
   --#{$button}--m-clicked--BorderColor--high-contrast: var(--pf-t--global--border--color--high-contrast);
 
   // Primary
@@ -485,7 +485,9 @@
     --#{$button}--PaddingInlineStart: var(--#{$button}--m-link--PaddingInlineStart);
     --#{$button}--Color: var(--#{$button}--m-link--Color);
     --#{$button}--BorderWidth: var(--#{$button}--m-link--BorderWidth);
+    --#{$button}--hover--BorderColor: var(--#{$button}--hover--BorderColor--high-contrast);
     --#{$button}--hover--BorderWidth: var(--#{$button}--m-link--hover--BorderWidth);
+    --#{$button}--m-clicked--BorderColor: var(--#{$button}--m-clicked--BorderColor--high-contrast);
     --#{$button}--m-clicked--BorderWidth: var(--#{$button}--m-link--m-clicked--BorderWidth);
     --#{$button}--BorderColor: var(--#{$button}--BorderColor--high-contrast);
     --#{$button}--BorderRadius: var(--#{$button}--m-link--BorderRadius);
@@ -659,7 +661,9 @@
   // Icon buttons
   &.pf-m-plain {
     --#{$button}--BorderWidth: var(--#{$button}--m-plain--BorderWidth);
+    --#{$button}--hover--BorderColor: var(--#{$button}--hover--BorderColor--high-contrast);
     --#{$button}--hover--BorderWidth: var(--#{$button}--m-plain--hover--BorderWidth);
+    --#{$button}--m-clicked--BorderColor: var(--#{$button}--m-clicked--BorderColor--high-contrast);
     --#{$button}--m-clicked--BorderWidth: var(--#{$button}--m-plain--m-clicked--BorderWidth);
     --#{$button}--BorderColor: var(--#{$button}--BorderColor--high-contrast);
     --#{$button}--BorderRadius: var(--#{$button}--m-plain--BorderRadius);
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -27,6 +27,8 @@
   --#{$button}--TransitionProperty: color, background, border-width, border-color;
   --#{$button}--ScaleX: 1;
   --#{$button}--ScaleY: 1;
+  --#{$button}--BorderWidth--high-contrast: var(--pf-t--global--border--width--regular);
+  --#{$button}--BorderColor--high-contrast: var(--pf-t--global--border--color--high-contrast);
 
   // Hover
   --#{$button}--hover--BackgroundColor: transparent;
@@ -42,6 +44,8 @@
   --#{$button}--m-clicked--BackgroundColor: transparent;
   --#{$button}--m-clicked--BorderColor: transparent;
   --#{$button}--m-clicked--BorderWidth: var(--pf-t--global--border--width--action--clicked);
+  --#{$button}--m-clicked--BorderWidth--high-contrast: var(--pf-t--global--border--width--strong);
+  --#{$button}--m-clicked--BorderColor--high-contrast: var(--pf-t--global--border--color--high-contrast);
 
   // Primary
   --#{$button}--m-primary--Color: var(--pf-t--global--text--color--on-brand--default);
@@ -485,6 +489,18 @@
     --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-link--m-small--PaddingInlineEnd);
     --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-link--m-small--PaddingInlineStart);
 
+    @media (prefers-contrast: more) {
+      &:hover:not(.pf-m-inline) {
+        --#{$button}--BorderWidth: var(--#{$button}--BorderWidth--high-contrast);
+        --#{$button}--BorderColor: var(--#{$button}--BorderColor--high-contrast);
+      }
+    
+      &.pf-m-clicked:not(.pf-m-inline) {
+        --#{$button}--BorderWidth: var(--#{$button}--m-clicked--BorderWidth--high-contrast);
+        --#{$button}--BorderColor: var(--#{$button}--m-clicked--BorderColor--high-contrast);
+      }
+    }
+  
     &.pf-m-inline {
       @at-root span#{&} {
         --#{$button}--m-link--m-inline--Display: var(--#{$button}--span--m-link--m-inline--Display);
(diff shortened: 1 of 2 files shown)


kmcfaul:
Should the other variants of button have a thicker border for the clicked state? I'm noticing that the border becomes thicker for the clicked state in the figma, but this pattern doesn't apply to the other variants which already had a border (hover = clicked border width).

kmcfaul:
The figma file for button is also a little confusing on the CTA button. I've applied the styling to the link CTA but its design is more squared than the rounded like the Figma shows.

mcoker (approved):
Nice! LGTM, though it looks like this has exposed a performance issue with the border and background animations. Here is an issue to follow up on it - https://github.com/patternfly/patternfly/issues/7743

mcoker:
@lboehling looks like the update to the plain/no-padding button may have caused an issue with the label close button. That button has custom padding already, but it pushes the border outside of the label. If I get rid of the custom padding and just use the new fake padding we added to plain/no-padding, it goes from this:

Screenshot 2025-08-15 at 9 19 46 AM

to this

Screenshot 2025-08-15 at 9 22 38 AM

WDYT?

Also I looked up everywhere we use the plain/no-padding button in our components and those spots look ok, you can verify in this PR. For reference, they're:

* form label help button
* table column header help button
* inline clipboard copy buton

mcoker:
@kmcfaul in chatting with @lboehling, we decided to remove the border from the button's inline links. The main issue there is that those present like normal links - so if we draw a border around them, it follows that we would draw a border around regular links, too. However, regular links (and the button's "span" inline link) can wrap to multiple lines, then the border breaks. There is already enough contrast and distinction with links that they don't need the border, so we can get rid of the border we added to inline links.

lboehling:
> @lboehling looks like the update to the plain/no-padding button may have caused an issue with the label close button. That button has custom padding already, but it pushes the border outside of the label. If I get rid of the custom padding and just use the new fake padding we added to plain/no-padding, it goes from this:
> 
> Screenshot 2025-08-15 at 9 19 46 AM
> to this
> 
> Screenshot 2025-08-15 at 9 22 38 AM
> WDYT?
> 
> Also I looked up everywhere we use the plain/no-padding button in our components and those spots look ok, you can verify in this PR. For reference, they're:
> 
> * form label help button
> * table column header help button
> * inline clipboard copy buton

+1 to getting rid of the custom padding. i think that your update looks good. When i went to toggle off the custom padding, it did start to look like the "x" close icon was butting up too closely with the label text. Is there anyway to keep a little gap between the text and the label, and use the updated no-padding padding? You can see the difference below (First 2 with inline start custom padding turned off, Second 2 with it left on). note - im not suggesting we keep the inline custom padding for the close action since it off centers the border, but if there was a way to add in that spacing between the close action and text again that would be 💯. 

Screenshot 2025-08-15 at 10 36 46 AM
Screenshot 2025-08-15 at 10 36 52 AM
Screenshot 2025-08-15 at 10 35 48 AM
Screenshot 2025-08-15 at 10 35 56 AM

The other examples look good! Thanks @mcoker & @kmcfaul

mcoker:
Chatted about the [label close spacing](https://github.com/patternfly/patternfly/pull/7734#issuecomment-3191940130) and decided to 1) remove the custom padding as design suggested, and 2) add back the space that was lost from the padding being removed on the left and right sides of the icon. Just glancing at it, I think that update would look like:
* Set all of these padding/margin variables to "0", and leave a comment to remove them in a breaking change. Usually that looks like `// TODO - remove in breaking change` https://github.com/patternfly/patternfly/blob/9005060274bc955d4d5b6eabd6568c583eeb0f81/src/patternfly/components/Label/label.scss#L225-L230
* Add a gap between the label text and actions via a `gap` on `.pf-v6-c-label`, and use `--pf-t--global--spacer--gap--text-to-element--compact` as the spacer.
* I think you can add back the space to the right of the close icon by setting this negative right margin to "0" and leave a TODO comment to remove it, too https://github.com/patternfly/patternfly/blob/9005060274bc955d4d5b6eabd6568c583eeb0f81/src/patternfly/components/Label/label.scss#L220

src/patternfly/components/Button/button.scss
```diff
@@ -518,7 +538,11 @@
       --#{$button}--hover--TextDecorationColor: var(--#{$button}--m-link--m-inline--hover--TextDecorationColor);
       --#{$button}--m-link--hover--Color: var(--#{$button}--m-link--m-inline--hover--Color);
       --#{$button}--m-link--hover__icon--Color: var(--#{$button}--m-link--m-inline--hover__icon--Color);
-
+      --#{$button}--after--Inset: var(--#{$button}--m-link--m-inline--after--Inset);
```
mcoker: No need for this now. While it does increase the clickable area, it would be inconsistent with our regular non-button links. 

```suggestion
```

src/patternfly/components/Button/button.scss
```diff
@@ -141,6 +148,7 @@
   --#{$button}--m-link--m-inline--hover--Color: var(--pf-t--global--text--color--brand--hover);
   --#{$button}--m-link--m-inline--hover__icon--Color: var(--pf-t--global--icon--color--brand--hover);
   --#{$button}--m-link--m-inline--hover--TextDecorationColor: currentcolor;
+  --#{$button}--m-link--m-inline--after--Inset: calc(#{pf-size-prem(2px)} * -1);
```
```suggestion
```

mcoker (changes requested):
Just a couple of small changes. And chatted with design about the border overflowing compact labels and we landed on using a custom inset for plain/no-padding actions in compact labels by removing the top and bottom inset.

You could do it a bunch of ways, but I would probably(?) put it in [this spot](https://github.com/patternfly/patternfly/blob/02eb35b6b9f51c460d3fdf3ee77eb0559ceb451b/src/patternfly/components/Label/label.scss#L590-L591), you could do something like this. Update from:

```
  .#{$button} {
    --#{$button}--FontSize: var(--#{$label}__actions--c-button--FontSize);
```

to

```
  .#{$button} {
    @at-root .#{$label}.pf-m-compact & {
      --#{$button}--m-plain--m-no-padding--after--Inset: 0 calc(#{pf-size-prem(2px) * -1)
    }

    --#{$button}--FontSize: var(--#{$label}__actions--c-button--FontSize);
```

That should compile to

```
.pf-v6-c-label.pf-m-compact .pf-v6-c-label__actions .pf-v6-c-button {
  --pf-v6-c-button--m-plain--m-no-padding--after--Inset: 0 -2px; 
}
```

srambach (approved):
Just curious if we need the link and plain clicked and hover border color variables? Are they here for completeness?

src/patternfly/components/Button/button.scss
```diff
@@ -172,7 +185,7 @@
   --#{$button}--m-plain--m-no-padding--hover__icon--Color: var(--pf-t--global--icon--color--regular);
   --#{$button}--m-plain--m-no-padding--hover--BackgroundColor: transparent;
   --#{$button}--m-plain--m-no-padding--m-clicked--BackgroundColor: transparent;
-  --#{$button}__progress--Color: var(--#{$button}__icon--Color);
+  --#{$button}--m-plain--m-no-padding--after--Inset: calc(#{pf-size-prem(2px)} * -1);
```
mcoker: Can we change the name of `after--Inset` to `border--offset` for all of these vars? It's more semantic so it isn't necessarily tied to the `::after` element and `inset` property.

mcoker:
@lboehling can you take a look at this PR? Specifically it changes:
* Adds border to link and plain buttons in HC
  * Excludes inline link
* Adds the fake padding to the "no padding" button variation
* Removes the custom padding/margin we had on the close icons in labels in lieu of the new fake padding
* Adds a small gap between the label text + actions
* Adjusts the fake padding on compact labels to remove the top/bottom padding so the button border fits within the label

mcoker (commented):
Looks awesome to me! Just one last request on the var names for the insets, we typically use the word "offset" when we move a border/outline from some relative position, and we can replace `after` with `border` which is easier to read and lets us update the structure down the road (border could go on the `::before` element) and the var name shouldn't need to be updated.

mcoker (approved):
🚀

lboehling (approved):
lgtm! the only thing i'm noticing is that the animation seems really slow when the plain button border fades in, but maybe we can consider that in the follow up @mcoker opened!

PR #7735, opened 2025-08-06
feat(tree-view): add high-contrast border

### LINKED ISSUE #7608: High contrast borders - Tree view
https://www.figma.com/design/wKcOq7IrzfL1gEK2mocd6r/Semantic-dimension-border-width----font-weight-adjustment-test?node-id=10994-65004

srambach:
Fixes #7608 

Adds border as ::after on the __node.
For compact variation, sets the border width to 0 so that the borders don't show, since the hover does not cause highlighting. 

However, questions for @lboehling -
1. In the compact variation, do you want hover to show borders?
2. In the compact variation with a background, do you want to show borders around each node?
^^ answered in slack - ok to go with no borders

[Figma link](https://www.figma.com/design/iT76DS020Ry1Wswfc0Hmes/branch/wKcOq7IrzfL1gEK2mocd6r/High-Contrast-Exploration?m=auto&node-id=10994-65004&t=DSuEchl5SfZjfWQy-1)

### DIFF
--- a/src/patternfly/components/TreeView/tree-view.scss
+++ b/src/patternfly/components/TreeView/tree-view.scss
@@ -18,10 +18,10 @@ $pf-v6-c-tree-view--MaxNesting: 10 !default;
   --#{$tree-view}__node--Color: var(--pf-t--global--text--color--subtle);
   --#{$tree-view}__node--BackgroundColor: transparent;
   --#{$tree-view}__node--BorderRadius: var(--#{$tree-view}__content--BorderRadius);
-  --#{$tree-view}__node--BorderWidth: 0;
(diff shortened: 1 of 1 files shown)


lboehling (commented):
image

Per slack convo -- I think that we should just go with the top level, red borders in the example above for compact tree view (with and without bkg). When i had initially reviewed, i was just thinking about tree view expansion, not row selects. the red border option definitely makes the clickable interaction area the most clear.

Otherwise this looks good thank you!

lboehling (approved):
thank you! looks good!

mcoker (approved):
👏🤖
`LBLEEPBLOOPTM`

PR #7738, opened 2025-08-12
fix(codeeditor): added header border

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7615

### DIFF
--- a/src/patternfly/components/CodeEditor/code-editor.scss
+++ b/src/patternfly/components/CodeEditor/code-editor.scss
@@ -6,8 +6,12 @@
   --#{$code-editor}__controls--Gap: var(--pf-t--global--spacer--gap--action-to-action--plain);
 
   // header
-  --#{$code-editor}__header--before--BorderBlockEndWidth: var(--pf-t--global--border--width--box--default);
-  --#{$code-editor}__header--before--BorderBlockEndColor: var(--pf-t--global--border--color--default);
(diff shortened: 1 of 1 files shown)


lboehling (approved):
noice!

srambach (approved):
This looks great. It is technically a visual breaking change, unless we are considering it a bug (which could be fair).

mcoker:
@srambach good callout. I'm good with it, @lboehling wdyt? Is there anyone else we might want to run this change by in terms of whether this will be disruptive or cause issues?

As an example, we include the code editor's "header" after each of our code examples that will then expand to a full code editor window if you click on the language button, so now that box has borders and rounded corners on the top. I'm not even sure if that's a recommended use of the code editor, but if a product did something similar, that may be something they'd want to review/update when pulling in this version of PF (which is what makes it "breaking")

Screenshot 2025-08-18 at 3 58 29 PM

mcoker:
I just hacked this together but I wonder if something like this might help the design and be a nudge to users that there is code from that list of buttons

Screenshot 2025-08-18 at 4 11 09 PM

lboehling:
@mcoker i love the suggestion you just mocked up for the code editor example. IMO, that has always felt like a weird behavior/misuse of the code editor header on our doc site. We could ask others during office hours if they know of any other product that uses a collapsed code editor in that way?

PR #7748, opened 2025-08-19
feat(DataList): remove li aria, update ids

### LINKED ISSUE #7274: DataList - invalid/incorrect aria attributes in examples
- Per [naming role guidance from W3C](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_role_guidance), listitems should not have aria labeling even if technically it may work. We should remove any aria-label, aria-labelledby, and/or aria-describedby from any `li` items in this component.
- In expandable examples, the toggle has an invalid Id passed to the aria-labelledby attribute - the ID being reference is `...item[1|2|3|...]`, but there should be a hyphen between the word "item" and the number e.g. "data-list-expandable-compact-item-1".

**Any other information?**

kmcfaul:
Closes #7274.

### DIFF
--- a/src/patternfly/components/DataList/examples/DataList.md
+++ b/src/patternfly/components/DataList/examples/DataList.md
@@ -209,7 +209,7 @@ When a list item includes more than one block of content, it can be difficult fo
   {{#> data-list-item data-list-item--id="item-1" data-list-item--expanded="true"}}
     {{#> data-list-item-row}}
       {{#> data-list-item-control}}
-        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 item-1" id="' data-list--id '-toggle1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
+        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 ' data-list--id '-item-1" id="' data-list--id '-toggle1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
       {{/data-list-item-control}}
       {{#> data-list-item-content}}
         {{#> data-list-cell data-list-cell--modifier="pf-m-icon"}}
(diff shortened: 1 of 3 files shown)


src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -209,7 +209,7 @@ When a list item includes more than one block of content, it can be difficult fo
   {{#> data-list-item data-list-item--id="item-1" data-list-item--expanded="true"}}
     {{#> data-list-item-row}}
       {{#> data-list-item-control}}
-        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 ' data-list--id '-item1" id="' data-list--id '-toggle1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
+        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 item-1" id="' data-list--id '-toggle-1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
```
thatblindgeye: For this toggle, we'll either want to remove the hyphen in `-toggle-1` or add a hyphen to `-toggle1`

Then for all the expandable toggles, we should update the `item-[number]` so that it also gets the data-list--id applied to it. Right now the aria-labelledby's second ID is just `item-[number]`, but the text content we want to pull from prepends that with `data-list-expandable-`

thatblindgeye (changes requested):

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -209,7 +209,7 @@ When a list item includes more than one block of content, it can be difficult fo
   {{#> data-list-item data-list-item--id="item-1" data-list-item--expanded="true"}}
     {{#> data-list-item-row}}
       {{#> data-list-item-control}}
-        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 ' data-list--id '-item1" id="' data-list--id '-toggle1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
+        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 item-1" id="' data-list--id '-toggle1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
```
thatblindgeye: This'll apply to all the aria labelledby's here:

```suggestion
        {{#> data-list-toggle button--attribute=(concat 'aria-labelledby="' data-list--id '-toggle1 ' data-list--id '-item-1" id="' data-list--id '-toggle1" aria-label="Toggle details for" aria-expanded="true" aria-controls="' data-list--id '-content1"')}}{{/data-list-toggle}}
```

Just want to ensure these expandable toggle buttons have an accessible name that includes that "primary|secondary|tertiary content". Right now the buttons only have an accesible name of "Toggle details for"

thatblindgeye (changes requested):

thatblindgeye (approved):

PR #7750, opened 2025-08-20
chore(DataList): provided context for demo icons

### LINKED ISSUE #7275: DataList - demos need additional context for icons with number text
Demos include icons with numbers next to them, as seen in the following screenshot:

Image

When traversed via AT, there's no context of what "10" means or is for. We should provide at least some screen reader text to provide that additional context. The same updated will need to be made in React and should follow whatever cadence is used in Core demos.

**Any other information?**

thatblindgeye:
Closes #7275

- Added SR text
- Updated "x" icon to error/danger icon
- Added status color to the applicable icons

### DIFF
--- a/src/patternfly/demos/DataList/data-list-content-2.hbs
+++ b/src/patternfly/demos/DataList/data-list-content-2.hbs
@@ -11,9 +11,9 @@
     {{> data-list--icon-text data-list--icon-text--icon="code-branch" data-list--icon-text--text="5"}}
     {{> data-list--icon-text data-list--icon-text--icon="code" data-list--icon-text--text="9" data-list--icon-text--screenreader-text="Code blocks"}}
     {{> data-list--icon-text data-list--icon-text--icon="cube" data-list--icon-text--text="2" data-list--icon-text--screenreader-text="Workspaces"}}
-    {{> data-list--icon-text data-list--icon-text--icon="check-circle" data-list--icon-text--modifier="pf-m-success" data-list--icon-text--text="11" data-list--icon-text--screenreader-text="Completed"}}
-    {{> data-list--icon-text data-list--icon-text--icon="exclamation-triangle" data-list--icon-text--modifier="pf-m-warning" data-list--icon-text--text="4" data-list--icon-text--screenreader-text="Warnings"}}
-    {{> data-list--icon-text data-list--icon-text--icon="exclamation-circle" data-list--icon-text--modifier="pf-m-danger" data-list--icon-text--text="1" data-list--icon-text--screenreader-text="Errors"}}
+    {{> data-list--icon-text data-list--icon-text--icon="check-circle" icon-content--modifier="pf-m-success" data-list--icon-text--text="11" data-list--icon-text--screenreader-text="Completed"}}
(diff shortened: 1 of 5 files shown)


src/patternfly/demos/DataList/data-list-content-2.hbs
```diff
@@ -9,11 +9,11 @@
...
   {{#> l-flex l-flex--modifier="pf-m-wrap"}}
     {{> data-list--icon-text data-list--icon-text--icon="code-branch" data-list--icon-text--text="5"}}
-    {{> data-list--icon-text data-list--icon-text--icon="code" data-list--icon-text--text="9"}}
-    {{> data-list--icon-text data-list--icon-text--icon="cube" data-list--icon-text--text="2"}}
-    {{> data-list--icon-text data-list--icon-text--icon="check-circle" data-list--icon-text--text="11"}}
-    {{> data-list--icon-text data-list--icon-text--icon="exclamation-triangle" data-list--icon-text--text="4"}}
-    {{> data-list--icon-text data-list--icon-text--icon="times-circle" data-list--icon-text--text="1"}}
+    {{> data-list--icon-text data-list--icon-text--icon="code" data-list--icon-text--text="9" data-list--icon-text--screenreader-text="Code blocks"}}
+    {{> data-list--icon-text data-list--icon-text--icon="cube" data-list--icon-text--text="2" data-list--icon-text--screenreader-text="Workspaces"}}
+    {{> data-list--icon-text data-list--icon-text--icon="check-circle" data-list--icon-text--modifier="pf-m-success" data-list--icon-text--text="11" data-list--icon-text--screenreader-text="Completed"}}
```
mcoker: What you have here is great and probably preferred, but just FYI in cases like this where the `data-list--icon-text` is so limited in scope (it should only have a single icon in it), it's totally fine to just use `icon-content--modifier` directly on `data-list--icon-text`.

```suggestion
    {{> data-list--icon-text data-list--icon-text--icon="check-circle" icon-content--modifier="pf-m-success" data-list--icon-text--text="11" data-list--icon-text--screenreader-text="Completed"}}
```

Also seeing this makes me think we should really enhance the icon component hbs to have parameters/props for `icon--IsWarning`, `icon--IsSmall`, etc

mcoker (approved):
BEEYOUTIFUL

lboehling (approved):
nice!

src/patternfly/demos/DataList/data-list-content-2.hbs
thatblindgeye: Ah wasnt sure if that would pass through so just went with the new attr, updated to just use icon-content--modifier!

mcoker: Yep! Parameters are inherited to all nested things unless they're explicitly set/defined anywhere in the children. FWIW sometimes we add a little context wrapper like this - `id-wrapper.hbs` doesn't exist, so it renders nothing, but it does set a bunch of parameters/props as context for its children

https://github.com/patternfly/patternfly/blob/afa1921db78776ab3b85f580a78fa6e1f8978228/src/patternfly/components/MenuToggle/menu-toggle--check.hbs#L1-L16

mcoker (approved):

nicolethoen (approved):

PR #7751, opened 2025-08-21
chore(HelperText): updated how status is conveyed

### LINKED ISSUE #7282: HelperText - lack of conveying status accurately
With v6 updates to helper text, additional updates still need to be made in example code:

- When there is a status helper text, there should be visually hidden text to provide context of that status like React does (see screenshot below)
  Image
- There are instances where a status helper text is being used, but no icon is rendered (examples and demos). Additionally, React no longer allows an icon to not be rendered if the helper text has a status, but Core still has a "Static" and "With icons" examples. Core should be updated to remove any example that conveys the icon being optional on anything that isn't a "default" helper text item.

**Any other information?**

thatblindgeye:
Closes #7282 

- Updated some handlebars logic so that you shouldn't opt into an icon being rendered, rather it's the default for status helper text items.
- For the SR text, I went with something different than what React has: rather than appending "[status] Status" after the visible helper text, I prepended it with "[status]:". In React we currently allow a way to omit the SR text per a product ask at one point, and in the future it may make sense to allow more flexibility with how the SR text is rendered in React (instead of always rendering it at the end of a helper text item, allow consumers to have it at the beginning). I can update Core to match React for now, though, if we'd prefer that.

### DIFF
--- a/src/patternfly/components/HelperText/helper-text-item-icon.hbs
+++ b/src/patternfly/components/HelperText/helper-text-item-icon.hbs
@@ -2,7 +2,6 @@
   {{#if helper-text-item-icon--attribute}}
     {{{helper-text-item-icon--attribute}}}
   {{/if}}>
-  {{#if helper-text-item--HasIcon}}
     {{#if helper-text-item--IsIndeterminate}}
       <i class="fas fa-fw fa-minus" aria-hidden="true"></i>
     {{else if helper-text-item--IsWarning}}
@@ -11,12 +10,9 @@
       <i class="fas fa-fw fa-check-circle" aria-hidden="true"></i>
     {{else if helper-text-item--IsError}}
       <i class="fas fa-fw fa-exclamation-circle" aria-hidden="true"></i>
-    {{else}}
-      <i class="fas fa-fw fa-minus" aria-hidden="true"></i>
+    {{else if @partial-block}}
+      {{> @partial-block}}
+    {{else if helper-text-item-icon--type}}
+      <i class="fas fa-fw fa-{{helper-text-item-icon--type}}" aria-hidden="true"></i>
     {{/if}}
-  {{else if @partial-block}}
-    {{> @partial-block}}
-  {{else}}
-    <i class="fas fa-fw fa-{{helper-text-item-icon--type}}" aria-hidden="true"></i>
-  {{/if}}
 </span>
--- a/src/patternfly/components/HelperText/helper-text-item-icon.hbs
+++ b/src/patternfly/components/HelperText/helper-text-item-icon.hbs
@@ -2,17 +2,17 @@
   {{#if helper-text-item-icon--attribute}}
     {{{helper-text-item-icon--attribute}}}
   {{/if}}>
-    {{#if helper-text-item--IsIndeterminate}}
-      <i class="fas fa-fw fa-minus" aria-hidden="true"></i>
-    {{else if helper-text-item--IsWarning}}
-      <i class="fas fa-fw fa-exclamation-triangle" aria-hidden="true"></i>
-    {{else if helper-text-item--IsSuccess}}
-      <i class="fas fa-fw fa-check-circle" aria-hidden="true"></i>
-    {{else if helper-text-item--IsError}}
-      <i class="fas fa-fw fa-exclamation-circle" aria-hidden="true"></i>
-    {{else if @partial-block}}
-      {{> @partial-block}}
-    {{else if helper-text-item-icon--type}}
-      <i class="fas fa-fw fa-{{helper-text-item-icon--type}}" aria-hidden="true"></i>
-    {{/if}}
+  {{#if helper-text-item--IsIndeterminate}}
+    <i class="fas fa-fw fa-minus" aria-hidden="true"></i>
+  {{else if helper-text-item--IsWarning}}
+    <i class="fas fa-fw fa-exclamation-triangle" aria-hidden="true"></i>
+  {{else if helper-text-item--IsSuccess}}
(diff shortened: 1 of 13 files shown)


nicolethoen (approved):

src/patternfly/components/HelperText/helper-text-item-icon.hbs
```diff
@@ -2,7 +2,6 @@
   {{#if helper-text-item-icon--attribute}}
     {{{helper-text-item-icon--attribute}}}
   {{/if}}>
-  {{#if helper-text-item--HasIcon}}
     {{#if helper-text-item--IsIndeterminate}}
```
mcoker: Can you fix the indentation of this block?

mcoker (changes requested):
The list below is all the places I spotted something off. It looks like a lot of the issues have basic helper text (no status, not indeterminate). Instead of requiring `--HasNoIcon` to omit the icon, could you change this block and instead of `unless --HasNoIcon`, do something like `{{#ifAny helper-text-item--IsSuccess, helper-text-item--IsError, helper-text-item--IsWarning, helper-text-item--IsIndeterminate, helper-text-item-icon--type}}`? Then you can also just remove the test for `@partial-block` in this file - that shouldn't be needed. https://github.com/patternfly/patternfly/blob/dd7ee4464dba7228fc3cba7ff665996287d8835e/src/patternfly/components/HelperText/helper-text-item.hbs#L15-L17

* There is some weird nested bug on this toolbar demo - https://pf-pr-7751.surge.sh/components/toolbar/html-demos/toolbar-with-validation-on-desktop/
* File upload example needs `--HasNoIcon` - https://github.com/patternfly/patternfly/blob/dd7ee4464dba7228fc3cba7ff665996287d8835e/src/patternfly/components/FileUpload/examples/FileUpload.md?plain=1#L67
* Progress example needs `--HasNoIcon` - https://github.com/patternfly/patternfly/blob/dd7ee4464dba7228fc3cba7ff665996287d8835e/src/patternfly/components/Progress/examples/Progress.md?plain=1#L255-L264
* First form field in this helper text demo needs `--HasNoIcon` - https://github.com/patternfly/patternfly/blob/dd7ee4464dba7228fc3cba7ff665996287d8835e/src/patternfly/demos/HelperText/examples/HelperText.md?plain=1#L19
* All of the `{{> form-helper-text}}` partials in this form demo need `--HasNoIcon` - https://github.com/patternfly/patternfly/blob/dd7ee4464dba7228fc3cba7ff665996287d8835e/src/patternfly/demos/Form/examples/BasicForms.md?plain=1
* Default helper text under the "dynamic" section has an icon on https://www.patternfly.org/components/helper-text/html#dynamic, but not on https://pf-pr-7751.surge.sh/components/helper-text#dynamic - was that intentional? Seems to me like it should have the icon but I'm not 100%.
* Default helper text in this form help text example has the indeterminate icon https://www.patternfly.org/components/forms/form/html#help-text, but this doesn't https://pf-pr-7751.surge.sh/components/forms/form#help-text. That seems like a good change to me, but just wanted to make sure it was intentional? FWIW it isn't clear in Figma - I see "default" with and without the hyphen/indeterminate icon.
    Screenshot 2025-08-28 at 7 00 09 PM
    Screenshot 2025-08-28 at 6 59 34 PM
*

thatblindgeye:
@mcoker let me know if I missed anything - I updated the unless block to that ifAny and removed instances of the HasNoIcon hbs attribute. For your last 2 bullets about the default icon, at least going by React logic we don't allow an icon for default helper text unless the `icon` prop is passed in. We could add a Core example showing "custom icons" similar to what React has if we'd want some documentation that states "hey you *can* add an icon to default helper text, just make sure it isn't the same as another helper text variant/is consistent across your app"

mcoker (commented):
Looks awesome! Only question is around removing the "-" from the basic helper text in the [dynamic example](https://pf-pr-7751.surge.sh/components/helper-text/#dynamic) - want to make sure that was intentional, and if so that design approves. I still see the icon in the helper text stuff in figma (under forms) but it isn't clear if/when that's needed.

Image

thatblindgeye:
@mcoker fwiw this is the React PR that had made Penta updates and removed an icon for a `default` variant: https://github.com/patternfly/patternfly-react/pull/10029

mcoker:
@lboehling do you know if the "-" before the default helper text in "dynamic" helper text is supposed to be there or not? I always assumed it was there for non-status helper text so that as the helper text updates as you type, it doesn't move around too much since all variations of helper text would have some kind of icon. Is that true or should our dynamic example not include the "-" before the regular helper text? https://www.patternfly.org/components/helper-text/html#dynamic
 
Screenshot 2025-09-11 at 2 43 56 PM

lboehling:
> @lboehling do you know if the "-" before the default helper text in "dynamic" helper text is supposed to be there or not? I always assumed it was there for non-status helper text so that as the helper text updates as you type, it doesn't move around too much since all variations of helper text would have some kind of icon. Is that true or should our dynamic example not include the "-" before the regular helper text? https://www.patternfly.org/components/helper-text/html#dynamic
> 
> Screenshot 2025-09-11 at 2 43 56 PM

I don't know the background for this, but I assumed the same as you did -- that it was there to keep the helper text from jumping around if a status was added to it when placed in a list of requirements/helper texts (like in password strength). I don't think it's needed for all helper text w/o status, just in the dynamic option with a list of requirements?

mcoker (approved):
Verified offline that "regular" help text in a dynamic scenario should not have an icon, but intermediate text should. Here's a screenshot of what that would look like if both were present.

Image

PR #7753, opened 2025-08-21
chore(login): fix header markup in main element

### LINKED ISSUE #6928: Bug - Login - header should not be inside main
**Describe the problem**
When using the Login component it has a `header` that is inside `main`

https://www.patternfly.org/components/login-page/react/basic/

**Expected behavior**
Structure should be:
```html
<body>
 <header>
 <!-- Site-wide header content -->
 </header>
 <main>
 <!-- Main content of the page -->
<!-- in case of the login theme the login box would be shown here -->
<!-- IdP links should be shown here -->
<!-- "Login box footer" should be shown here -->
 </main>
 <footer>
 <!-- Site-wide footer content -->
 <!-- users could add some common footer links information here -->
 </footer>
</body>
```

mcoker:
fixes https://github.com/patternfly/patternfly/issues/6928


thatblindgeye (commented):
This is what React currently looks like on Org:

Markup of the PatternFly React Login component

React also doesn't allow customizing the elements for the header, main, or footer components that use those tags (login__main, __footer, and __header classes). Looks like we updated the stuff inside that main element to be hardcoded divs now (the __main-header, and __main-footer classes; __main-body was already a div).

This PR only updates the header component in Core to be a div, but we'd still have footer tag being used for __main-footer.

thatblindgeye (approved):

PR #7754, opened 2025-08-22
chore(expandable section): add missing aria attributes to examples

### LINKED ISSUE #7280: ExpandableSection - examples missing aria attributes
- The toggle needs the `aria-controls` attribute pointing to an ID of the expandable section content
- The expandable section content that can be collapsed should have a role of "region" and an aria-labelledby pointing to the toggle text, to match React.

**Any other information?**

Mash707:
Fixes #7280

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
+++ b/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
@@ -6,7 +6,7 @@
     button--IsLink=true
     button--IsInline=expandable-section--IsTruncate
     button--AriaExpanded=expandable-section--IsExpanded
-    button--aria-controls=(ternary expandable-section--IsDetached (concat expandable-section--id '-content') null)
+    button--aria-controls=(ternary expandable-section--id (concat expandable-section--id '-content') null)
     button--id=(concat expandable-section--id '-toggle')
     button--icon-template=(ternary expandable-section--IsTruncate null "expandable-section-toggle-icon")}}
     {{#if @partial-block}}
--- a/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
+++ b/src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
@@ -7,6 +7,7 @@
     button--IsInline=expandable-section--IsTruncate
     button--AriaExpanded=expandable-section--IsExpanded
     button--aria-controls=(ternary expandable-section--IsDetached (concat expandable-section--id '-content') null)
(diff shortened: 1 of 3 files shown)


src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
```diff
@@ -7,6 +7,7 @@
     button--IsInline=expandable-section--IsTruncate
     button--IsAriaExpanded=expandable-section--IsExpanded
     button--aria-controls=(ternary expandable-section--IsDetached (concat expandable-section--id '-content') null)
```
thatblindgeye: This should be updated to render the aria-controls properly. Right now it's only rendering for detached variants, but we want it regardless of variant as long as an ID is passed.

```suggestion
    button--aria-controls=(ternary expandable-section--id (concat expandable-section--id '-content') null)
```

src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
```diff
@@ -7,6 +7,7 @@
     button--IsInline=expandable-section--IsTruncate
     button--IsAriaExpanded=expandable-section--IsExpanded
```
thatblindgeye: @mcoker are you opposed to replacing instances of `button--IsAriaExpanded` in Core handlebars files with just `button--IsExpanded`? Only 5 instances of IsAriaExpanded in the codebase, and the logic in Button doesn't take that into account when setting the actual aria-expanded attribute (it checks if either IsExpanded or IsAriaExpanded is true, but then only sets aria-expanded based on IsExpanded). Having separate Disabled and AriaDisabled attr is fine since they're distinct things, but there's only an aria-expanded attr so just using IsExpanded is fine.

If you're okay with that, then for this PR we can just update this line to:

```js
button--IsExpanded=(ternary expandable-section--IsExpanded true false)
```

(since we always want aria-expanded rendered whether true or false for this component), and open a followup for the other files (unless you want to update them here @Mash707, shouldn't take long at all).

thatblindgeye (changes requested):

src/patternfly/components/ExpandableSection/expandable-section-toggle.hbs
thatblindgeye: @Mash707 ignore this comment for this PR. Coker and I met up to discuss this and I'll put a followup PR to update this across components.

thatblindgeye (approved):

mcoker (approved):
LGTM!

PR #7755, opened 2025-08-22
fix(breadcrumb): add missing font weight css variable

### LINKED ISSUE #7385: Bug - Breadcrumb - Missing `font-weight` CSS variable
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Within Breadcrumb component I'm noticing that `--pf-v6-c-breadcrumb__item--FontWeight` is being used but is `undefined`. Checking out PF5 in Cockpit I'm noticing it was also missing back there.

1. Inspect a breadcrumb item
2. See font-weight CSS for `.pf-v6-c-breadcrumb__item` class
3. See it undefined

https://www.patternfly.org/components/breadcrumb/react/basic

**Expected behavior**
Either the font-weight is defined through a variable or removed altogether.

No

**Screenshots**
Image

**What is your environment?**

**Any other information?**

Mash707:
Fixes #7385

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -79,7 +79,7 @@
   padding-inline-start: var(--#{$breadcrumb}__link--PaddingInlineStart); // use a mutable value for alignment control
   padding-inline-end: var(--#{$breadcrumb}__link--PaddingInlineEnd); // use a mutable value for alignment control
   font-size: inherit;
-  font-weight: inherit;
+  font-weight: var(--#{$breadcrumb}__link--FontWeight, inherit);
   line-height: inherit;
   color: var(--#{$breadcrumb}__link--Color);
   text-decoration-line: var(--#{$breadcrumb}__link--TextDecorationLine);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Breadcrumb/breadcrumb.scss
```diff
@@ -78,7 +79,7 @@
   padding-inline-start: var(--#{$breadcrumb}__link--PaddingInlineStart); // use a mutable value for alignment control
   padding-inline-end: var(--#{$breadcrumb}__link--PaddingInlineEnd); // use a mutable value for alignment control
   font-size: inherit;
-  font-weight: var(--#{$breadcrumb}__link--FontWeight);
+  font-weight: inherit;
```
mcoker: Technically this would be breaking - if a user had set `--#{$breadcrumb}__link--FontWeight` previously to add a custom font-weight here, this change will remove their custom weight. You can change it to below, then add `--#{$breadcrumb}__link--FontWeight: initial;` in the var list at the top of this stylesheet and that will make the font-weight `inherit` but also allow a user's customization using that variable to continue to work.

```suggestion
  font-weight: var(--#{$breadcrumb}__link--FontWeight, inherit);
```

mcoker (changes requested):
Just one comment, otherwise lgtm!

src/patternfly/components/Breadcrumb/breadcrumb.scss
Mash707: Yup makes sense.

PR #7756, opened 2025-08-22
fix(card): added missing body color variable

### LINKED ISSUE #7386: Bug - Card - Undefined color variable
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

Within Card component I'm noticing that `--pf-v6-c-card__body--Color` is being used but is `undefined`

1. Inspect a Card's body
2. See CSS for `.pf-v6-c-card__body`
3. Notice `color: var(--pf-v6-c-card__body--Color)` and how the variable is `undefined`

https://www.patternfly.org/components/card/react/basic-cards

**Expected behavior**
Either the font-weight is defined through a variable or removed altogether.

No

**Screenshots**
Image

**What is your environment?**

**Any other information?**

Mash707:
Fixes #7386

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -20,6 +20,7 @@
   --#{$card}__title-text--OverflowWrap: break-word;
   --#{$card}--c-button--disabled--Color: var(--pf-t--global--text--color--on-disabled);
   --#{$card}__body--FontSize: var(--pf-t--global--font--size--body--default);
+  --#{$card}__body--Color: var(--pf-t--global--text--color--regular);
   --#{$card}__footer--FontSize: var(--pf-t--global--font--size--body--default);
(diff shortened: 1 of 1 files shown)


srambach (approved):
👍🏻 thanks!

mcoker (changes requested):
This looks great, but since this is changing the card body's text color from `inherit` to our default color, and cards are used quite a bit, would you mind adding `inherit` as a fallback for the body text color to give users a way to revert this change in case they were relying on `inherit`?

https://github.com/patternfly/patternfly/blob/a9c8f5beb220f5be66e61438971bd69d2d2a90c4/src/patternfly/components/Card/card.scss#L445

If you chnage that to `color: var(--#{$card}__body--Color, inherit);` then someone can revert this change by setting `--#{$card}__body--Color: initial;`.

PR #7757, opened 2025-08-25
chore(backstop): update visual regression tests

srambach:
Updates visual regression test. Report includes a couple of inline edit examples that seem to be a glitch because I re-ran the test and it corrected.

https://drive.google.com/file/d/1Tcdh_F99TxdtchkDXOSUu0jXd80dSaNI/view?usp=sharing

One problem noted 
- Text overflows container at mobile width in /components/expandable-section/html/disclosure-variation-indented/ text overflows

Changes noticed and expected:
- Wizard - nav spacing
- Wizard - warning step
- Text input group - placeholder disabled color
- Text input group - new example
- Text input group with navigable options
- Sortable table - header vertical alignment
- Expandable table - vertical alignment of caret
- Label group close button color
- Vertical label group width because of change of button from no padding to small
- Inline edit state invalid - buttons missing - but reappeared in subsequent run
- Error helper text example added
- Expandable section example text change
- Toolbar with validation demo - error icon
- Ghost line reappeared on mobile views
- Expected spinner diffs


mcoker (approved):
Nice!!! Thank you @srambach! 

> Text overflows container at mobile width in /components/expandable-section/html/disclosure-variation-indented/ text overflows

IMO that's OK, we just added more text to that button. I imagine we have a handful of examples that would do that on a very narrow screen.

PR #7774, opened 2025-08-29
feat(high-contrast): add class to switch on high-contrast

### LINKED ISSUE #7624: Add class to enable HC theme
Similar to other themes, we should have a HC class that's added to `<html>` to enable HC theme. To follow our other themes, we can name it `.pf-v6-theme-high-contrast`

Should ideally be completed with/after https://github.com/patternfly/patternfly-org/issues/4738

srambach:
Fixes #7624 

Adds (uncomments) class selector to load high-contrast token set.

### DIFF
--- a/src/patternfly/base/patternfly-variables.scss
+++ b/src/patternfly/base/patternfly-variables.scss
@@ -17,9 +17,9 @@
   @include dark.pf-v6-tokens;
 }
 
-// :root:where(.pf-v6-theme-high-contrast) {
-//   @include highcontrast.pf-v6-tokens;
-// }
+:root:where(.pf-v6-theme-high-contrast) {
+  @include highcontrast.pf-v6-tokens;
+}
 
 @media (prefers-contrast: more) {
   :root {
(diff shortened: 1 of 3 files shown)


src/patternfly/base/patternfly-variables.scss
```diff
@@ -17,19 +17,19 @@
...
 }
 
-// :root:where(.pf-v6-theme-high-contrast) {
-//   @include highcontrast.pf-v6-tokens;
-// }
+:root:where(.pf-v6-theme-high-contrast) {
+  @include highcontrast.pf-v6-tokens;
+}
 
 @media (prefers-contrast: more) {
```
mcoker: You'll need to take out this block and the one below it that applies the CSS using `prefers-contrast` https://github.com/patternfly/patternfly/blob/f3acfe220a7b2b2756103adef041652608b324d0/src/patternfly/base/patternfly-variables.scss#L34-L38

mcoker (changes requested):
Just need to disable the existing blocks that include the tokens via the media query. Also we still have the hard-coded border tokens I added a while back. Looks like some of them are still in use, so you could just change this media query to `:root:where(.pf-v6-theme-high-contrast)` https://github.com/patternfly/patternfly/blob/f3acfe220a7b2b2756103adef041652608b324d0/src/patternfly/base/tokens/tokens-local.scss#L146

PR #7776, opened 2025-08-29
chore(demo): correct aria attributes on jump links examples

### LINKED ISSUE #7286: JumpLinks - aria attributes need correction
Various aria attributes that need some form of correction in examples/demos:

- `aria-current="location"` needs to be applied to the `li` that is currently selected in a nav
- The `nav` and `ul` elements that makes up the navigation in the component should have unique aria labeling - either `aria-label` if there's no visible text label (i.e. the "Jump to section" text) or `aria-labelledby` to point to that visible text. This would require providing unique label text for the 'main' jump links, or providing unique nav text if there are nested lists (like in the "Vertical with inactive subsections" example, since the nested list would make sense to be labeled by its parent nav item).
- For the expandable implementation, the toggle button shouldn't have an aria-label if there's visible text to use. This will be confusing since the aria-label text isnt visible anywhere and doesn't match the text actually used in the button. In react it is possible to have an expandable JumpLinks without a text label in which an aria-label would make sense, but this depends if we intended that to be supported.

**Any other information?**

wise-king-sullyman:
Closes #7286


thatblindgeye (changes requested):
- Let's remove the "navigation" and "list" verbiage from the aria labeling, that'll already get announced by AT
- Can we make updates to the visible text labels for the jump links as https://github.com/patternfly/patternfly-react/pull/11950?
- [vertical with inactive subsection](https://pf-pr-7776.surge.sh/components/jump-links#vertical-with-inactive-subsections) will just need an aria-label applied to the nested list - similar to what I mentioned to Rebecca, for now we can just pass an aria-label to that list with something like "Second inactive subsection" or similar unique-but-generic
- Similar to above re: aria-label on nested list for [vertical with active subsection](https://pf-pr-7776.surge.sh/components/jump-links#vertical-with-active-subsections)
- We just need to remove the aria-label being applied to the expandable toggle buttons as well since there's already the visible text

thatblindgeye (approved):
LGTM!

@mcoker as a followup to this, do we want to update the "Expandable (responsive) with no label" example so that the toggle is just an icon, or remove the example altogether? Currently React allows a toggle withou a visible text label (relying on aria label instead), but we don't explicitly show that in an example. In this Core example, when the toggle appears it has a visible text label.

mcoker (approved):
🚀

PR #7777, opened 2025-09-02
chore(Progress): updated placement and verbiage for status examples

### LINKED ISSUE #7288: Progress - examples with status should have helper text rendered
In React there is a `helperText` prop for Progress, which can be used to convey the status of a Progress -- currently there's nohing other than a visual icon (which is hidden from AT) and progress bar color that conveys status.

This helper text should also be linked to the progress bar via an `aria-describedby` attribute on the `role="progressbar"` element.

**Any other information?**

thatblindgeye:
Closes #7288

### DIFF
--- a/src/patternfly/components/Progress/examples/Progress.md
+++ b/src/patternfly/components/Progress/examples/Progress.md
@@ -203,7 +203,7 @@ If the status that displays with the bar is not a percentage, then the ARIA tag 
 When conveying status, you should ensure:
 - There is visible helper text that explains the status.
 - The helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text/html#basic).
-- The helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
+- The helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar"]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
 
 ### Success
 ```hbs
--- a/src/patternfly/components/Progress/examples/Progress.md
+++ b/src/patternfly/components/Progress/examples/Progress.md
@@ -201,9 +201,9 @@ If the status that displays with the bar is not a percentage, then the ARIA tag 
 ## Status examples
 
 When conveying status, you should ensure:
-- there is visible helper text that explains the status,
-- the helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text#basic), and
-- the helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
+- There is visible helper text that explains the status.
+- The helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text/html#basic).
+- The helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
 
 ### Success
 ```hbs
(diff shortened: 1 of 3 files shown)


kaylachumley (approved):
looks good! didn't see anything too crazy - not sure we want erin to review the copy? up to you!

src/patternfly/components/Progress/examples/Progress.md
```diff
@@ -280,6 +198,96 @@ If the status that displays with the bar is not a percentage, then the ARIA tag
 {{/progress}}
 ```
 
+## Status examples
+
+When conveying status, you should ensure:
+- there is visible helper text that explains the status,
+- the helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text#basic), and
+- the helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
+
```
edonehoo:
```suggestion
When conveying status, you should ensure the following:
- There is visible helper text that explains the status.
- The helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text#basic).
- The helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).

```

edonehoo (changes requested):
Just some small nits (mostly making each point a sentence)!

src/patternfly/components/Progress/examples/Progress.md
```diff
@@ -280,6 +198,96 @@ If the status that displays with the bar is not a percentage, then the ARIA tag
 {{/progress}}
 ```
 
+## Status examples
+
+When conveying status, you should ensure:
+- There is visible helper text that explains the status.
+- The helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text/html#basic).
```
mcoker, line 205: Impressive! Anticipating the [change from `#static` to `#basic`](https://github.com/patternfly/patternfly/pull/7751/files). But generally if I link to another component in the docs, I link to the default/react tab. That link won't work if you're just looking at core docs (like in this PR), but it works on org. Curious what you and @edonehoo think.

```suggestion
- The helper text includes the status icon, as seen in our [basic helper text example](/components/helper-text#basic).
```

src/patternfly/components/Progress/examples/Progress.md
```diff
@@ -280,6 +198,96 @@ If the status that displays with the bar is not a percentage, then the ARIA tag
...
+- The helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
```
mcoker: Looks like you a quotation mark

```suggestion
- The helper text is linked to the `.pf-v6-c-progress__bar[role="progressbar"]` element via an `aria-describedby` attribute, as seen in the [progress helper text example](#helper-text).
```

mcoker (commented):
Just a couple small comments

src/patternfly/components/Progress/examples/Progress.md
thatblindgeye: In this case this should be fine - I think my one caveat would be do we want to link to a React example if someone is browsing the HTML docs? The link not working correctly on staging/local builds is a fair point, but I guess that depends if we should worry as much about that or the final experience on org?

edonehoo: agree! didn't catch that, but I also link to the react tab unless we're referencing a core-specific example

edonehoo: oops my page didn't refresh with Eric's comment before my reply. I didn't realize that the progress doc is an html page since I'm usually in react land, but hm that's a fair point. If it's safe to assume that consumers generally stick with html/css implementation consistently (rather than mingling html & react implementations for something like this) then maybe it is better to link to the html helper text here?

mcoker (approved):
LGTM!

edonehoo (approved):

PR #7787, opened 2025-09-03
feat(high-contrast): use variable for high-contrast

### LINKED ISSUE #7624: Add class to enable HC theme

srambach:
Fixes #7774 

Turns on the high-contrast switcher in docs framework
Removes media queries for prefers-contrast: more and replaces it with the `.pf-v6-theme-high-contrast` variable.


mcoker (approved):
Perfect!

PR #7788, opened 2025-09-03
fix(Icon): remove extra hidden progress icon from success state

### LINKED ISSUE #7285: Bug - Icon - in progress success icon is being announced as though it were "in progress"
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [x] Patternfly 6  
- [ ] other

In the "In progress" example, the success icon is being announced by VoiceOver as though it were in progress (basically the same way the spinner icon is being announced in this example). This would be confusing to users who are able to see the success icon but are still navigating with VoiceOver, as well as convey the wrong information to users in general.

Go to the above example and navigate with VoiceOver, notice that the success icon has a progress indicator announced.

**Expected behavior**
The success icon would be aria-hidden and there would be no progress indicator rendered for it. Note the markup from React when the success icon is rendered:

Image

Versus when the spinner icon is rendered:

Image

**Screenshots**

**What is your environment?**

**Any other information?**

kmcfaul:
Closes #7285.

### DIFF
--- a/src/patternfly/components/Icon/examples/Icon.md
+++ b/src/patternfly/components/Icon/examples/Icon.md
@@ -253,9 +253,6 @@ Inline with size specified: {{#> icon icon--modifier="pf-m-sm pf-m-inline"}}
   {{#> icon-content}}
     <i class="fas fa-check-circle" aria-hidden="true"></i>
   {{/icon-content}}
-  {{#> icon-progress}}
-    {{#> spinner spinner--modifier="pf-m-md"}}Loading...{{/spinner}}
-  {{/icon-progress}}
 {{/icon}}
 ```
 


thatblindgeye (approved):
@mcoker can you just confirm we don't want the `__progress` container at all for an "in progress" icon that shows a loaded/completed state? Based on docs that's the case but just want to confirm in case we want to keep the container for any future reason.

mcoker (approved):
LGTM!

mcoker:
@thatblindgeye good question! It does look like the CSS for in progress icons are setup in a way that they may have transitioned/faded between the progress spinner and whatever icon replaces it at some point, but it doesn't transition currently. https://github.com/patternfly/patternfly/blob/3d9bacdc8d904c43f49fe7bb9913c37848b80c71/src/patternfly/components/Icon/icon.scss#L204-L207

I'd say matching the react implementation is what we should do for core, then we can follow up with design and see if there is supposed to be a transition with the icons to/from in-progress, and if so address that in a future issue. No need to hold up this PR as long as core matches react.

@lboehling @kaylachumley wdyt? For an "in progress" icon that could have some icon in it, then on some action, the icon is replaced with a spinner, then when the action is complete, the spinner goes away and either the original or a new icon is shown - should the icons just change immediately? Or should the they fade in/out, so the old icon fades out and the spinner fades in, and when it's done, the spinner fades out and the new icon fades in?

Issue #7737, opened 2025-08-12
High contrast borders - audit use of border tokens


mcoker:
Because we went from _no HC border tokens_ -> _HC border tokens_ -> _new base tokens & updates to existing tokens to bake in HC support_, once we get toward the end of our initial development, let's review border-width tokens and make sure they're being used correctly.

**Note:** Be sure to check tabs - looks like their height shifts between HC/default. Need to make sure that's working as expected.

Once this is done, remove the temporary HC tokens added here https://github.com/patternfly/patternfly/blob/5c0af4890e23b6d75f8d92448d34eb6120912552/src/patternfly/base/tokens/tokens-local.scss#L127-L165

mcoker:
closed by https://github.com/patternfly/patternfly/issues/7864 and https://github.com/patternfly/patternfly/pull/7791

PR #7791, opened 2025-09-04
chore(high-contrast): remove temporary local high contrast tokens

srambach:
Removes the local high-contrast tokens that were added temporarily.

First step of #7737


mcoker (approved):
L🎳TM!

PR #7799, opened 2025-09-05
fix(calendar-month): add current/range HC borders

### LINKED ISSUE #7793: Bug - Calendar month - missing border on current date and date range
I missed these calendar month updates in https://github.com/patternfly/patternfly/issues/7613

Image

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7793

### DIFF
--- a/src/patternfly/components/CalendarMonth/calendar-month.scss
+++ b/src/patternfly/components/CalendarMonth/calendar-month.scss
@@ -41,6 +41,9 @@
   --#{$calendar-month}__dates-cell--before--InsetInlineEnd: 0;
   --#{$calendar-month}__dates-cell--before--InsetBlockEnd: var(--#{$calendar-month}__dates-cell--PaddingBlockEnd);
   --#{$calendar-month}__dates-cell--before--InsetInlineStart: 0;
+  --#{$calendar-month}__dates-cell--after--BorderColor: var(--pf-t--global--border--color--high-contrast);
+  --#{$calendar-month}__dates-cell--after--BorderBlockStartWidth: 0;
+  --#{$calendar-month}__dates-cell--after--BorderBlockEndWidth: 0;
   --#{$calendar-month}__dates-cell--m-in-range--before--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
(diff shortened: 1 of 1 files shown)


lboehling (approved):
very nice!

src/patternfly/components/CalendarMonth/calendar-month.scss
```diff
@@ -136,21 +143,34 @@
...
+  }
+
+  &::before {
     background-color: var(--#{$calendar-month}__dates-cell--before--BackgroundColor);
   }
 
+  &::after {
+    border: solid var(--#{$calendar-month}__dates-cell--after--BorderColor);
+    border-width: var(--#{$calendar-month}__dates-cell--after--BorderBlockStartWidth) 0 var(--#{$calendar-month}__dates-cell--after--BorderBlockEndWidth);
+  }
```
srambach: Does this bother anyone else enough to break convention and move it to the `::before`? 

Image

mcoker: I had it on the before... _before_... but I had errantly put it on disabled cells, too, and the disabled circle overlapped the border and made it look weird. It looked like this

Screenshot 2025-09-05 at 12 57 48 PM

This is what it looked like on disabled cells if the border is on top of the date circles

Screenshot 2025-09-05 at 1 01 45 PM

So it can make the border look weird if the circle overlaps the border... but I think your comment is probably more correct. I'd probably expect any border/shape from a cell to overlap the border - not the other way around.

FWIW this is what a "hovered" date looks like if the border is on the before (hovering "17")

Screenshot 2025-09-05 at 1 00 08 PM

I think it's fine to move the border to the before, but FWIW this is what that focused element looks like with the border behind - can't really tell at regular zoom (to me anyways) and even when zoomed in, it's hard to really see.

Screenshot 2025-09-05 at 12 56 32 PM

srambach: Hmm yeah the disabled ones cause a weird visual effect when the border is beneath. Given that, I'd stick with how you have it as the lesser of evils 😈

srambach (approved):
That works for me!

PR #7800, opened 2025-09-05
fix(date-picker): add HC border to calendar element

### LINKED ISSUE #7794: Bug - Date picker - "calendar" element needs border, mark for removal
Image

Just needs a border for now. That element isn't in use with react, so I'm guessing it's a straggler from before we used menu/popover and popper and we never removed it. If that looks to be the case, leave a TODO comment to remove it in a breaking change.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7794


srambach (approved):
L 🗓️ TM

PR #7807, opened 2025-09-08
chore(Button): updated how ariaexpanded and class are set

thatblindgeye:


src/patternfly/components/Button/button.hbs
```diff
@@ -44,8 +43,8 @@
       type="button"
     {{/unless}}
   {{/if}}
-  {{#if (concat button--IsExpanded button--IsAriaExpanded)}}
-    aria-expanded="{{button--IsExpanded}}"
+  {{#if (concat button--AriaExpanded)}}
```
mcoker: Do you still need the concat helper?

```suggestion
  {{#if button--AriaExpanded}}
```

src/patternfly/components/Table/TableCells/table-cell-toggle.hbs
```diff
@@ -4,14 +4,13 @@
     button--IsPlain=true
     button--IsIcon=true
     button--icon-template="table-toggle-icon"
-    button--IsExpanded=table-tr--IsExpanded
-    button--IsAriaExpanded=table-tr--IsExpanded
+    button--AriaExpanded=table-tr--IsExpanded
+    button--modifier=(ternary table-tr--IsExpanded "pf-m-expanded" "")
```
mcoker, line 8: 👍

mcoker (commented):
LGTM - just one question on the use of `concat`.

src/patternfly/components/Button/button.hbs
thatblindgeye: Only for you to catch it 😭

mcoker (approved):
send it!

PR #7808, opened 2025-09-08
fix(menu-toggle): use alt border for split toggle divider

### LINKED ISSUE #7804: Bug - Menu toggle - update split toggle border color
The border between buttons in a split toggle menu toggle doesn't have enough contrast, especially in high contrast mode. We can use the  `--pf-t--global--border--color--alt` token for that border instead (in default and HC theme) https://github.com/patternfly/patternfly/blob/e993388031614465e148c0b7c2eaa5ad6073f97b/src/patternfly/base/tokens/tokens-default.scss#L573C3-L573C37

srambach:
Fixes #7804 

Changes the divider for the split toggle to `--pf-t--global--border--color--alt`


mcoker (approved):
LFabulous™

PR #7809, opened 2025-09-08
fix(tabs): add side borders to selected box tab in high contrast

### LINKED ISSUE #7798: Bug - Tabs - add inline high contrast borders to box tabs
> would be nice if selected box tab had left/right borders

So maybe not a requirement, but if it's straight forward, let's add them

Image

srambach:
Fixes #7798 

NOTE: This exposes that the current tab decorator (blue line) is 1px too long on the left side. It comes from line 110, and I think this could be removed along with a couple of other lines - it will break visual regression tests but is also technically a bug fix. I have NOT included it here until confirmation that it's truly not needed and we want to make the change.

`  --#{$tabs}__link--before--InsetInlineStart: calc(var(--#{$tabs}__link--before--border-width--base) * -1);`

image

### DIFF
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -107,10 +107,10 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__link--before--BorderInlineEndWidth: 0;
   --#{$tabs}__link--before--BorderBlockEndWidth: 0;
   --#{$tabs}__link--before--BorderInlineStartWidth: 0;
-  --#{$tabs}__link--m-current--before--BorderInlineStartColor: var(--pf-t--global--border--color--high-contrast);
-  --#{$tabs}__link--m-current--before--BorderInlineStartWidth: var(--pf-t--global--border--width--high-contrast--regular);
-  --#{$tabs}__link--m-current--before--BorderInlineEndColor: var(--pf-t--global--border--color--high-contrast);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Tabs/tabs.scss
```diff
@@ -311,6 +315,10 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
     .#{$tabs}__item.pf-m-current {
       --#{$tabs}__link--BackgroundColor: var(--#{$tabs}__item--m-current__link--BackgroundColor);
       --#{$tabs}__link--before--BorderBlockEndColor: var(--#{$tabs}__link--BackgroundColor);
+      --#{$tabs}__link--before--BorderInlineStartColor: var(--#{$tabs}__link--m-current--before--BorderInlineStartColor);
+      --#{$tabs}__link--before--BorderInlineStartWidth: var(--#{$tabs}__link--m-current--before--BorderInlineStartWidth);
+      --#{$tabs}__link--before--BorderInlineEndColor: var(--#{$tabs}__link--m-current--before--BorderInlineEndColor);
+      --#{$tabs}__link--before--BorderInlineEndWidth: var(--#{$tabs}__link--m-current--before--BorderInlineEndWidth);
```
mcoker: Can you put `--m-box` in front of these? The vars before them are set to `--m-box` vars but for whatever reason they're declared above and down here...

https://github.com/srambach/patternfly/blob/3d9bacdc8d904c43f49fe7bb9913c37848b80c71/src/patternfly/components/Tabs/tabs.scss#L293

mcoker (commented):
Just one thing por favor

mcoker (approved):
L🗂️TM!

PR #7811, opened 2025-09-08
fix(drawer): add top border in high contrast

### LINKED ISSUE #7805: Bug - Drawer - Drawer panel needs top border in some instances
Standard use of the drawer in the page component (like for notification drawer) is one place - http://pf-core-staging.patternfly.org/components/notification-drawer/html-demos/expanded-read

I wonder if we should just add a top border to the drawer panel always and live with the double border it might create in some cases?

mcoker:
~~@lboehling this adds a top border to all drawer panels. This would mean there would be a double border with the panel open if there were already a top border added in the UI, but I wonder if that's worth the trade-off of this adding a border to places there is not currently a border between the drawer and whatever is above it (like the notification drawer)~~

Adds a top border to drawer panels when the HTML structure is attached to the page layout - most common use of this layout is with a notification drawer. This is also the HTML structure generated when using `<Page notificationDrawer={...} />`

fixes https://github.com/patternfly/patternfly/issues/7805

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -154,6 +154,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__main-wizard--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$page}__main-wizard--BorderBlockStartColor: var(--pf-t--global--border--color--default);
   --#{$page}__main-wizard--BorderBlockStartWidth: var(--pf-t--global--border--width--action--default);
+
+  // Drawer section
+  --#{$page}__drawer--c-drawer--BorderBlockStartWidth: var(--pf-t--global--border--width--high-contrast--regular);
+  --#{$page}__drawer--c-drawer--BorderBlockStartColor: var(--pf-t--global--border--color--high-contrast);
 }
 
 .#{$page} {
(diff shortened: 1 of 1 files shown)


lboehling (approved):
i think this looks good! benefits of the border in the full height drawer outweigh the slight weirdness with the double border when it's in page. Wondering if there is anyway to just target the notification drawer specifically to add the top border? if not i think this is g2g!

mcoker:
@lboehling updated to just style the "notification drawer" drawer. Here are places you can see the "notification drawer" that should get the border
* description list in a drawer demo - https://pf-pr-7811.surge.sh/components/description-list/html-demos/in-drawer/
* all drawer demos - https://pf-pr-7811.surge.sh/components/drawer/html-demos
* all notification drawer demos - https://pf-pr-7811.surge.sh/components/notification-drawer/html-demos

You should not see that border on any other drawer - like any of the drawer examples, the primary-detail drawers, etc

Question - when the drawer width spans the width of the viewport (on mobile), do you still want the border?

Here are some screenshots:

## Panel expanded on desktop
Screenshot 2025-09-09 at 4 56 07 PM

## Panel expanded on mobile (with border)
Screenshot 2025-09-09 at 4 56 12 PM

## Panel expanded on mobile (without border)
Screenshot 2025-09-09 at 4 56 22 PM

## Drawer collapsed on mobile
Screenshot 2025-09-09 at 4 56 40 PM

lboehling:
this all looks awesome thank you @mcoker! Yes, i think we should have the border on mobile too. thanks!

src/patternfly/components/Page/page.scss
```diff
@@ -568,10 +572,15 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   flex-shrink: 0;
 }
 
-.#{$page}__drawer .#{$page}__main-container {
-  align-self: revert;
+.#{$page}__drawer {
+  > .#{$drawer} > .#{$drawer}__main > .#{$drawer}__panel {
```
srambach, line 576: Have you ever seen products insert a wrapper in this structure? Just curious if it's common and whether it would suffice to use child vs direct child selector.

mcoker: Not that I'm aware of.... but it's worth noting that with the `<Page notificationDrawer={}>` prop, you can't even modify any of that as far as I'm aware. Everything that's passed to the `noficationDrawer` prop is put in the `drawer__panel` element - https://github.com/patternfly/patternfly-react/blob/d63615d0e82bd93d77b7883fbfafdeba6c7755eb/packages/react-core/src/components/Page/Page.tsx#L325-L329

srambach (approved):
L🗄️ TM

Issue #7796, opened 2025-09-04
Bug - Menu - sticky footer missing high contrast border


mcoker:
Look to see if there is any other sticky thing (or something with a box-shadow) that may need a border

<img width="841" height="471" alt="Image" src="https://github.com/user-attachments/assets/443f0ab4-1214-47cd-85b3-714837a2df5a" />

srambach:
Some other instances of box-shadow that probably need a border:

- [x] [alert group overflow button](https://github.com/patternfly/patternfly/pull/7813)
- [x] [panel scrollable with footer](https://github.com/patternfly/patternfly/pull/7812)
- [x] [sidebar panel (mobile)](https://github.com/patternfly/patternfly/pull/7824/)
- [x] [toolbar - sticky and mobile expanded content](https://github.com/patternfly/patternfly/pull/7814)
- [x] [wizard nav (mobile)](https://github.com/patternfly/patternfly/pull/7816)

PR #7812, opened 2025-09-08
feat(menu, panel): add border to sticky footer

srambach:
Satisfies part of #7796 - sticker footer on the scrollable menu and panel


mcoker (approved):
L🎳TM!

PR #7813, opened 2025-09-08
fix(alert): fix hover styles, HC borders

mcoker:
Goes toward https://github.com/patternfly/patternfly/issues/7796

But I also noticed we had `:hover/focus/active` styles on the alert-group-item (an `<li>` that isn't interactive/focusable) instead of on the overflow button so I 1) removed those styles and 2) put the appropriate changes on the overflow button. 

Also in react, a plain ol' `<li>` wraps the overflow button, not a `li.pf-v6-c-alert-group__item`, so the `:hover` styles I removed weren't working. That should be fixed with the changes above, but I also wrapped the overflow button in an `<li>` in the core docs to match react.


srambach (approved):
good catches!

PR #7814, opened 2025-09-09
fix(toolbar): adds HC borders to sticky and expand content

mcoker:
goes towards https://github.com/patternfly/patternfly/issues/7796

Adds border to sticky, and looks like expandable content has a shadow on mobile so added one there, too.


srambach (approved):
L 🛠️ TM

PR #7829, opened 2025-09-12
fix(dual-list-selector): tree to support focus-within, hc border on hover/focus

### LINKED ISSUE #7795: Bug - Dual list selector w/ tree view - missing high contrast border on selected element
Image

**edit 9/12/25** after talking with @lboehling we're going to match the styling in [Tree View w/ checkboxes](https://www.patternfly.org/components/tree-view#with-checkboxes)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7795

* Adds HC border on hover/focus for tree variant
* Adds background color change when focusing on something (either the expand toggle or checkbox) in a tree item - this is a net new style to the default styles, not just HC - @lboehling is that ok? This matches tree view styles.

Previously looks like this:

Screenshot 2025-09-12 at 2 40 19 PM

Now looks like this:

Screenshot 2025-09-12 at 2 40 56 PM

For comparison this is tree view, which has the same styling:
Screenshot 2025-09-12 at 2 48 31 PM


lboehling (approved):
lgtm! good catch w the focus states

srambach (approved):
L 🤩 TM

PR #7832, opened 2025-09-15
fix(button): remove border width animation

### LINKED ISSUE #7743: Button - high contrast animation performance
In testing the [button high contrast border PR](https://github.com/patternfly/patternfly/pull/7734/), it looks like there is a performance issue with the borders and background color.

srambach:
Fixes #7743

After experimentation, it seems that the perceived jankiness in the plain button hover animation was because we are animating in a 1px border. Because it's simply going from 0 to 1px, there is very room for intermediate states and so it seems jumpy. I've removed border-width from the transition properties. Now the plain button (and other) button borders will animate in based just on the color, but this seems to have a smoother feel.


mcoker (approved):
L🤖TM!

lboehling (approved):
this feels so much smoother to me across the board -- always felt like the border transition felt a little jumpy!

PR #7833, opened 2025-09-16
fix(tabs, pagination): fixes page inset width

### LINKED ISSUE #7376: Bug - Tabs - missing inset value to correctly align tabs with PageSection gutter

srambach:
Fixes #7376

Adjusts the page inset offset of tabs and pagination to account for the adjustments in the main content area padding and border.


mcoker (approved):
LGTM!

PR #7845, opened 2025-09-17
fix(slider): adds borders to make visible in forced colors

### LINKED ISSUE #7835: Slider - add borders for forced colors

srambach:
Fixes #7835 

Adds a transparent border to the rail track, thumb, and step tick to make the slider visible in forced colors mode (in conjunction with high-contrast)

There is some use of box-shadow (thumb hover/active) and linear gradient (rail track color) that aren't addressed here, but for now this at least makes the slider visible in forced colors.

### DIFF
--- a/src/patternfly/components/Slider/slider.scss
+++ b/src/patternfly/components/Slider/slider.scss
@@ -208,7 +208,7 @@
   height: var(--#{$slider}__thumb--Height);
   cursor: pointer;
   background-color: var(--#{$slider}__thumb--BackgroundColor);
-    border: var(--#{$slider}__thumb--BorderWidth) solid var(--#{$slider}__thumb--BorderColor);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Slider/slider.scss
```diff
@@ -200,6 +208,7 @@
   height: var(--#{$slider}__thumb--Height);
   cursor: pointer;
   background-color: var(--#{$slider}__thumb--BackgroundColor);
+    border: var(--#{$slider}__thumb--BorderWidth) solid var(--#{$slider}__thumb--BorderColor);
```
mcoker:
```suggestion
  border: var(--#{$slider}__thumb--BorderWidth) solid var(--#{$slider}__thumb--BorderColor);
```

mcoker (approved):
LGTM! One small spacing update that I'll add.

PR #7847, opened 2025-09-18
fix(switch): adds borders to make visible in forced colors

### LINKED ISSUE #7834: Switch - add borders for forced colors

srambach:
Fixes #7834 

Also adds a border around disabled switches in high-contrast mode. (previously they only showed up on switches that were turned off.)

Note that disabled switches are currently not really distinguishable in forced-colors: active but I don't know another solution. For now it will have to rely on behavior (cursor etc)

@lboehling please take a look!


lboehling (approved):
looks good!

mcoker (approved):
❤️ it. Disabled switches are `disabled`, so you can't focus them, and they're read as disabled to assistive tech. Also we have the "not allowed" cursor on hover when they're disabled, though it would be great to have an alternate display so it's more visible that they're disabled. We can address that if/when we go for full support in forced-colors mode.

PR #7848, opened 2025-09-18
fix(toggle-group): added forced colors selected borders

### LINKED ISSUE #7837: Toggle group - update for forced colors mode
Design feedback was that we could add a thicker border for selected items

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7837

@lboehling needs some design feedback on this one. Updated selected toggles to have an additional border. It isn't very visible IMO:

Screenshot 2025-09-18 at 5 51 52 PM

We could add a hover border-width change of 1px and increase the selected state to 2px? Here's what that would look like with "option 1" on the third row hovered

Screenshot 2025-09-18 at 6 01 18 PM

The only thing that looks bad with that is the compact variant, feels kinda squished, but maybe it's ok?

Screenshot 2025-09-18 at 6 01 29 PM

----

Also updated the border between 2 selected toggles to use the `border--color--alt` token (like we updated with primary split toggles), and the border between 2 disabled toggles to use the disabled border color. The disabled border has less contrast than the regular border, but seems like the more correct use of the token? That border matches the other borders around the disabled toggles in a toggle group when it didn't before. Curious what you think of those changes

This is before the change in HC:
Screenshot 2025-09-18 at 4 57 57 PM

This is after the change:
Screenshot 2025-09-18 at 4 58 13 PM

### DIFF
--- a/src/patternfly/components/ToggleGroup/toggle-group.scss
+++ b/src/patternfly/components/ToggleGroup/toggle-group.scss
@@ -14,6 +14,7 @@
   --#{$toggle-group}__button--hover--BackgroundColor: var(--pf-t--global--background--color--primary--hover);
   --#{$toggle-group}__button--hover--ZIndex: var(--pf-t--global--z-index--xs);
   --#{$toggle-group}__button--hover--before--BorderColor: var(--pf-t--global--border--color--default);
+  --#{$toggle-group}__button--hover--after--BorderWidth: var(--pf-t--global--border--width--high-contrast--regular);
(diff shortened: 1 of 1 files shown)


lboehling:
I'm good with all of the above changes @mcoker! adjusting the border thickness is clever! The pros outweight the cons in the compact version -- it's not covering the text.

mcoker:
@lboehling roger that, thanks! Updated the PR to thicken the borders in forced colors mode - https://pf-pr-7848.surge.sh/components/toggle-group

srambach (approved):
L 🤲🏻 TM

PR #3684, opened 2020-11-18
feat(pagination): added layout mods

mattnolting:
closes #3561 

Caveat: the new modifiers DO NOT apply to `.pf-c-pagination.pf-m-bottom` as `.pf-m-bottom` has a unique and specific, presentation. 

@mcoker removed the `isFullscreen` tag as the examples are difficult to scan in component context.

PR #7849, opened 2025-09-19
feat(Pagination): updated top mobile view

### LINKED ISSUE #7852: Pagination - update top variant on mobile to new design guidelines
Pagination

**Describe the feature**
Design guidelines have been updated, showing the top variant Pagination on mobile viewports to no longer be just static text. Instead, mobile should have the Select and next/previous nav buttons (basically the compact view).

**Any other information?**

thatblindgeye:
Closes #7852 

Draft for now as there's some css cleanup and example updates still to be made if this looks good.

I'm leveraging the existing css media queries to show/hide stuff as needed to visually achieve the compact view of Pagination. This ends up making the `__total-items` container not needed - unless we want to allow a consumer to opt into this new display. If we want that instead, we could maybe add a modifier class (`pf-m-compact-on-mobile` or something,.

I have things setup so that the applicable pagination controls have display none for mobile viewports, so nothing is being repeated in the DOM, and this should trickle down to React where any additional props a consumer is passing in will remain on the applicable controls.

### DIFF
--- a/src/patternfly/components/Pagination/examples/Pagination.md
+++ b/src/patternfly/components/Pagination/examples/Pagination.md
@@ -8,6 +8,7 @@ cssPrefix: pf-v6-c-pagination
 ### Top
 ```hbs
 {{#> pagination}}
+  {{> pagination-total-items-content}}
   {{> pagination-menu-toggle pagination-menu-toggle--id="pagination-menu-toggle-top-example"}}
   {{> pagination-nav-content pagination-nav--aria-label="Pagination nav - top example"}}
 {{/pagination}}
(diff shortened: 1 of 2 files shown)


mcoker:
@thatblindgeye pushed an update https://github.com/patternfly/patternfly/pull/7849/commits/39b334e34b83c5db03bb18fac67d4c987b7c0da0

It scopes any changes that could have an impact to `.pf-m-compact-on-mobile` and only below the mobile breakpoint. That shouldn't interfere with any existing overrides or cause any style changes that could impact non-mobile viewports. WDYT?

Also marked the display-full and display-summary variations to be removed in a breaking change since you mentioned they aren't in react. We added them 5 years ago! https://github.com/patternfly/patternfly/pull/3684

I searched the patternfly-react repo's git logs for "displaySummary/Full" and "display-summary/full" and didn't get any hits. Do you know if they were ever used in react?

thatblindgeye:
Closing this and the linked issue for now as per disccusion during retro-planning, the resolution will likely live in DataView to some capacity rather than introducing a new variant here.

PR #7853, opened 2025-09-22
fix(divider): adds border to make visible in forced colors

### LINKED ISSUE #7840: Divider - disappears in forced colors mode
Divider is currently created with a background image - moving to a border would address the issue in forced colors mode, but given how commonly used the divider is, we wouldn't want to introduce a change stands a good chance of creating style conflicts for products (even if it isn't technically a "breaking change" on our end).

srambach:
Fixes #7840 

Adds a transparent border to the divider pseudo that is the visible part of the divider, so that it shows up in forced colors.


mcoker (approved):
LGTM! Ran visual regressions against the whole repo, didn't see any divider related stuff that was off.

lboehling:
nice!

PR #7859, opened 2025-09-23
fix(page): fix page section inset on mobile

### LINKED ISSUE #7856: Bug - Page - page sections don't align well on mobile
Since we removed the left/right page container borders on mobile, some of the page sections don't align well now.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7856


srambach (approved):
L 🍁 TM

PR #7861, opened 2025-09-23
fix(dual-list-selector): fix overflow of items

### LINKED ISSUE #7854: Bug - Dual list selector - long text overflows with draggable items
Long text in the draggable item - it overflows, but the width exceeds the parent box size. 

Image

Image

`.pf-v6-c-dual-list-selector__item`
* Can we take off `width: 100%`?
* Add `min-width: 0`

`.pf-v6-c-dual-list-selector__item, .pf-v6-c-dual-list-selector__item-main`
* Replace `flex-basis: 100;` with `flex-grow: 1;`

srambach:
Fixes #7854

### DIFF
--- a/src/patternfly/components/DualListSelector/dual-list-selector.scss
+++ b/src/patternfly/components/DualListSelector/dual-list-selector.scss
@@ -312,7 +312,7 @@ $pf-v6-c-dual-list-selector__item--MaxNesting: 10;
 
 .#{$dual-list-selector}__item,
 .#{$dual-list-selector}__item-main {
-  flex-basis: 100%; // TODO
+  flex-grow: 1;
 }
 
 .#{$dual-list-selector}__draggable {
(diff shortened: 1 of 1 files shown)


src/patternfly/components/DualListSelector/dual-list-selector.scss
```diff
@@ -312,7 +312,7 @@ $pf-v6-c-dual-list-selector__item--MaxNesting: 10;
 
 .#{$dual-list-selector}__item,
 .#{$dual-list-selector}__item-main {
-  flex-basis: 100%;
+  flex-basis: 100%; // TODO
```
mcoker: wdyt about replacing this with `flex-grow: 1`? Seems like that might be more predictable than using a percentage based size just to ensure the element takes up the available space by default.

Also did you mean to leave the TODO comment?

mcoker (commented):
Just one comment. Curious what you think about that. The PR is fine as-is if you'd rather go with what you have now.

mcoker (approved):
LGTM!

PR #7869, opened 2025-09-23
fix(card): use high contrast border for secondary

### LINKED ISSUE #7865: Card - update high contrast borders
change the secondary border to use the high contrast regular border width token

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7865

Visual regressions showed no differences. This just uses the more technically correct token for border-width.


srambach (approved):
L ♠️ TM

PR #7870, opened 2025-09-23
fix(panel): updated secondary/raised to use HC border tokens

### LINKED ISSUE #7866: Panel - update high contrast borders
update the secondary and raised borders to use the high contrast regular border width token

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7866


srambach (approved):
L ✋🏻 TM

PR #7872, opened 2025-09-23
fix(label): use HC border width token, plain border for overflow

### LINKED ISSUE #7868: Label - update high contrast borders
* update the default border-width to use the high contrast regular border width token
* update the overflow label to use the action plain border width tokens

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7868


srambach (approved):
L 🌈 TM

PR #7873, opened 2025-09-23
fix(form-control): fix select padding in FF, add select to demo

### LINKED ISSUE #7383: Bug - form-select - Firefox padding issue (due to Firefox changing padding recently)
**Describe the problem**
This is an issue with 
- [x] Patternfly 5  
- [x] Patternfly 6  

**Which components are affected?**

form-select

In a recent version of Firefox, the padding of form select elements has been made consistent with other browsers, after many years (the entire existence of Firefox, Chrome, WebKit) being different.

As PatternFly (rightfully) had a workaround to accommodate the difference for Firefox specifically, the workaround is actually now causing a breaking change in layouts in newer versions of Firefox.

This is an issue in PatternFly 5 and 6, and we're seeing it on the official websites for "form select" being different from other form elements, including inputs and selects. The left padding of form select in Firefox is 4px less than others.

Cockpit issue has been opened at https://github.com/cockpit-project/cockpit/issues/21671 which has screenshots showing it affecting elements in Cockpit.

This is affecting the PatternFly website demos for the components, for both 5 and 6 as well. This specific comment shows the issue in PatternFly 6: https://github.com/cockpit-project/cockpit/issues/21671#issuecomment-2701414685 on the form page, comparing an input and a "form select", which should have the same padding, but do not in Firefox.

**Expected behavior**

All browsers and inputs should have consistent padding.

We currently have a pending PR which undoes the workaround for Firefox by using CSS to target Firefox and set the left padding of forms to the standard left padding for inputs. This PR is at https://github.com/cockpit-project/cockpit/pull/21673

**Screenshots**

Here's the screenshot of the aformentioned PF6 comparison of two input elements (the text input and "form select"): 

Image

**What is your environment?**
 - OS: Linux
 - Browser: Firefox
 - Version 136 and 135 (a few previous versions may also be affected)

Cockpit

**Any other information?**

The fix is to remove the special-casing Firefox-specific CSS to adapt for the newer versions of Firefox, which is specifically this CSS:

https://github.com/patternfly/patternfly/blob/8373776762ffb011f11a25e9851847781ba594ad/src/patternfly/components/FormControl/form-control.scss#L294-L300

Please drop this code block for both PatternFly 6 and also PatternFly 5, so it all apps and websites using PatternFly render as expected in the newer versions of Firefox. Thanks!!!

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7383

* Removes browser prefixed FF styles for `<select>` and fixes padding to align with other form control components
* Updates our basic form demo to add a select menu and updates the `id` (and associated stuff) for a few demos where the id was like `id="-name"` or `id="-email"`

### DIFF
--- a/src/patternfly/components/FormControl/form-control.scss
+++ b/src/patternfly/components/FormControl/form-control.scss
@@ -394,13 +394,6 @@
     & .#{$form-control}__utilities {
       padding-inline-end: var(--#{$form-control}__utilities--select--PaddingInlineEnd);
     }
-
-    // Firefox's select text has additional padding
-    // stylelint-disable-next-line
-    @-moz-document url-prefix() {
-      --#{$form-control}--PaddingInlineEnd: calc(var(--#{$form-control}__select--PaddingInlineEnd) - 1px);
-      --#{$form-control}--PaddingInlineStart: calc(var(--#{$form-control}__select--PaddingInlineStart) - 4px);
-    }
   }
 
   &.pf-m-placeholder > select {
--- a/src/patternfly/demos/Form/examples/BasicForms.md
+++ b/src/patternfly/demos/Form/examples/BasicForms.md
@@ -8,7 +8,24 @@ subsection: forms
 ### Basic
 ```hbs
 {{#> form form--id="form-demo-basic"}}
-  {{#> form-group form-group--id="-name"}}
+  {{#> form-group form-group--id=(concat form--id "-title")}}
+    {{#> form-group-label}}
+      {{#> form-label form-label--attribute=(concat 'for="' form-group--id '"')}}Title{{/form-label}}
+    {{/form-group-label}}
+    {{#> form-group-control}}
+      {{#> form-control controlType="select" form-control--IsPlaceholder="true" form-control--attribute=(concat 'id="' form-group--id '" name="' form-group--id '"')}}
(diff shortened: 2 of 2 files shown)


srambach (approved):
L 👍🏻 TM

thatblindgeye (commented):
Some brief comments, some would be better as a followup since they're pre-existing things in the demos.

- In "Sections with repeatable fields" demo, would make more sense to not use a label element for the "Valid redirect URIs" field since there can be multiple inputs there (could just use a span element with the same classes, similar to what we do for the Basic example where there's a "group" label for multiple checkboxes that have their own unique label). Instead we could add and hook up that text's ID to each input via aria-labelledby to help get unique label names for the inputs.
- In "Complex form" demo, the "Labels" field isn't properly hooked up to the input part of the label widget (just needs an ID on the input). This one would be great to get in as part of this, but it is pre-existing.
- Not anything to do with your updates, but did want to call out for a followup: it seems like non-expandable FieldGroups in core prevent interaction on labels and input elements inside those field groups. React seems fine, but this is pre-existing and unrelated to the updates here - really just mentioning it because I looked at the wrong tab first 😆

thatblindgeye (approved):
Oops since none of the above is blocking

PR #7876, opened 2025-09-24
chore(build): upgrade gulp to v5

wise-king-sullyman:
Needed to support node 22+


mcoker:
@wise-king-sullyman I validated the dist dir is the same compared to the main branch (though you'll need to rebase this branch to pull in one extra commit). I ran `yarn clean && yarn install && yarn build-patternfly` on main (using node 20.13.3) and this branch (node 22.17.1), then ran `diff -r` against the 2 dist dirs those created and the only diff I got was.

```diff
Only in /Users/cmichael/dist-node-upgrade/docs: pages
diff -r dist/package.json /Users/cmichael/dist-node-upgrade/package.json
8c8
<     "node": ">=20.18.3 <22"
---
>     "node": ">=20.18.3"
75,76c75,76
<     "gulp": "^4.0.2",
<     "gulp-cli": "^2.3.0",
---
>     "gulp": "^5.0.1",
>     "gulp-cli": "^3.1.0",
```

Anything else you'd recommend testing outside of the dev server and scripts like backstop?

mcoker:
Actually didn't notice the build was failing. I was able to reproduce locally - I ran `yarn build:pficons && yarn build:pficonfont` and everything worked, including the icons listed on `/icons` on the dev site, which come from the icon font built using those 2 commands. Why aren't those running with `yarn start`?

FWIW after building the font manually, it does show the generated font file as modified, so it's different, although I used a couple of tools to diff the font files and I see no diffs. Also comparing my `/icons` on localhost compared to staging showed no diffs, either, so the file modification seems safe.

wise-king-sullyman:
Alright that error should be fixed now, with an extra tweak so that the build isn't slowed down when it doesn't need to be.

Non-clean builds should still take the same amount of time, but clean builds will take an extra ~5-10 seconds (including for the dev server) because we now need to wait for the icons to be generated before we try to do the copy operations.

> Anything else you'd recommend testing outside of the dev server and scripts like backstop?

Nothing else I can think of!

mcoker (approved):
LGTM! Tested
* full page screenshots
* backstop
* development server changes/hot-reloads
* compiles same `/dist` as main branch

PR #7879, opened 2025-09-25
fix(code-editor): fix borders for addl use cases

### LINKED ISSUE #7878: Bug - Code editor - update header plain to remove border
* Plain header should remove top/left/right border, but leave bottom border.
* Make sure header bottom border renders without content in `__header`
* Read-only tab should also have read only background.
* Language tab should be aligned to right in react/without header-main

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7878

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,5 +114,5 @@
     "registry": "https://registry.npmjs.org/",
     "access": "public"
   },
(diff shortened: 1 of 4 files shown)


srambach (approved):
Great!

PR #7880, opened 2025-09-25
chore: update full page example screenshots

mcoker:


srambach (approved):
👍🏻 Some good cleanup happened in the past year 🤪

PR #7886, opened 2025-10-09
fix(DragDrop): Updated to use grab cursor for draggable items by default

### LINKED ISSUE #7882: Bug - drag drop - update to use grab cursor
We're using `grabbing` on drag, but we should also be suing `grab` when the cursor is over something draggable

##

---

**Jira Issue:** [PF-2742](https://redhat.atlassian.net/browse/PF-2742)

tlabaj:
Updated to use `grab` cursor for draggable items by default.  Use `grabbing` only when `pf-m-dragging` modifier is applied.

closes: #7882

### DIFF
--- a/src/patternfly/components/DragDrop/drag-drop.scss
+++ b/src/patternfly/components/DragDrop/drag-drop.scss
@@ -1,7 +1,7 @@
 @use '../../sass-utilities' as *;
 
 @include pf-root($draggable) {
-  --#{$draggable}--Cursor: auto;
+  --#{$draggable}--Cursor: grab;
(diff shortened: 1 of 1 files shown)


srambach (commented):
Our examples are outdated, and I don't think this fix will change anything in React, as the class `pf-v6-c-draggable` doesn't seem to be used anymore.

It does fix it for the examples as currently implemented in core, but also causes the grab cursor to show up in the examples that are trying to show what happens during dragging, which is confusing.

Created #7901 to update our examples for parity with React. Curious what your thoughts are @mcoker ?

Issue #7196, opened 2024-11-06
Spike: Use container queries


andrew-ronaldson:

srambach:
Implementation in https://github.com/patternfly/patternfly/issues/7312

Issue #7312, opened 2025-01-24
Add support for Container queries


andrew-ronaldson:
PatternFly has received a number of requests to add support for container queries. These are similar to media queries but this version applies styling based on its container.

> Container queries enable you to apply styles to an element based on the size of the element's container. If, for example, a container has less space available in the surrounding context, you can hide certain elements or use smaller fonts. - [Source](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries)

This addition will help products adjust layouts in a more responsive way.

patternfly-jira-sync:
Closed via Jira sync - Jira issue PF-1141 was closed.

PR #7888, opened 2025-10-10
[WIP] Spike exploring container queries in toolbar

srambach:
This WIP exploration adds a set of container breakpoints that are based on the existing breakpoints minus the width of the navigation sidebar.

A modifier for toolbar has been created that changes the toolbar media query for arranging the expandable content to be a media query instead. As a POC, there are individual modifiers for each breakpoint and the background changes color as an easy indication of which container query is activating.

There are a couple of new examples in Toolbar. The first demonstrates that when placed next to the nav sidebar, the container query triggers at the same place that the old media query did. The other examples just demonstrate different breakpoints.

Part of #7196 
This work was Cursor assisted.

### DIFF
--- a/src/patternfly/components/Toolbar/examples/Toolbar.md
+++ b/src/patternfly/components/Toolbar/examples/Toolbar.md
@@ -557,8 +557,11 @@ Several components in the following examples do not include functional and/or ac
 {{/toolbar}}
 ```
 
+## Toolbars responsive to their container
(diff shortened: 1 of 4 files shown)


mcoker (approved):
Spike LGTM!

srambach:
Closing this PR as complete; container query work continues in epic #7312 and toolbar container query specifically in #7967

PR #7893, opened 2025-10-15
fix(Button): updated AriaExpanded to apply correctly

### LINKED ISSUE #7892: Bug - Button - AriaExpanded not set correctly for false value
`button--AriaExpanded` hbs attribute isn't setting aria-expanded correctly when the value is a boolean false.

thatblindgeye:
Closes #7892

Tried going for the quicker solution for now:

- Updated Button hbs so that the `if` block for button--AriaExpanded uses the concat helper; this should apply aria-expanded correctly when button--AriaExpanded is passed a boolean false value as well as a truthy value
- Added ternary helpers to components using button--AriaExpanded to have false as a fallback

Just need douible checked in reivew:

- aria-expanded is set with a true or false value for expandable ~things~ (true when expanded, false when collapsed)
- aria-expanded isn't incorrectly rendered if the ~thing~ isn't expandable (e.g. a basic primary button shouldn't have aria-expanded)

I also added a comment for the concat helper I added for now, I think we could probably have a followup to clean up how we're passing in button--AriaExpanded based on other component attributes (e.g. jump-links--IsExpanded is explicitly passed a stringy "true", but needs to be passed a boolean `false` otherwise a stringy "false" will set `pf-m-expanded` on the jump links.

### DIFF
--- a/src/patternfly/components/Alert/alert-toggle.hbs
+++ b/src/patternfly/components/Alert/alert-toggle.hbs
@@ -5,7 +5,7 @@
   {{> button
     button--IsPlain=true
     button--IsIcon=true
-    button--AriaExpanded=alert--IsExpanded
+    button--AriaExpanded=(ternary alert--IsExpanded alert--IsExpanded false)
     button--id=(concat alert--id '-toggle')
     button--aria-label="Details"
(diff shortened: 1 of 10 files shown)


mcoker (approved):
L🎃TM!

dlabaj (commented):
LGTM

PR #7904, opened 2025-10-20
fix(drawer): fix primary-detail pattern drawer header

### LINKED ISSUE #7803: Bug - Primary detail drawer headers have too much indentation
Here is one example - https://www.patternfly.org/patterns/primary-detail/html-demos/primary-detail-content-body-padding/. We should check the other drawers in the primary detail demos.

Looks like the drawer `__head` element should come before `__body`, instead of being nested inside of it.

srambach:
Fixes #7803 

• **Fixed CSS typo**: `pf-m-paddinng` → `pf-m-padding` in drawer-body.hbs
• **Removed unnecessary attribute**: `drawer-body--HasPadding=true` from Drawer.md example
• **Fixed drawer nesting**: Moved `drawer-head` out of `drawer-body` in page-template-drawer-panel.hbs
• **Simplified header structure**: Removed unnecessary flex layout in primary-detail-template-panel-header.hbs. It now uses `drawer-head` and `drawer-description` elements properly.

Just the description generated by cursor 😺

### DIFF
--- a/src/patternfly/components/Drawer/drawer-body.hbs
+++ b/src/patternfly/components/Drawer/drawer-body.hbs
@@ -1,6 +1,6 @@
 <div class="{{pfv}}drawer__body
   {{~#if drawer-body--HasPadding}}
-    pf-m-paddinng
+    pf-m-padding
   {{/if}}
   {{~#if drawer-body--modifier}}
     {{drawer-body--modifier}}
--- a/src/patternfly/components/Drawer/examples/Drawer.md
+++ b/src/patternfly/components/Drawer/examples/Drawer.md
@@ -185,7 +185,7 @@ import './Drawer.css'
       {{#> drawer-body}}
         Drawer panel body content
       {{/drawer-body}}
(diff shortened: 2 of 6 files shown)


mcoker (changes requested):
Nice use of the drawer description!

Looks like there are a couple more
* Wizard "With drawer" demo has `drawer-body` around `drawer-head`
    https://github.com/patternfly/patternfly/blob/9b4a99e89c65bc6bb0487daee0a2cd5c248cccfc/src/patternfly/demos/Wizard/examples/Wizard.md?plain=1#L210-L217

* Wizard "with drawer" example has `drawer-body` around `drawer-head` along with some other issues
  * text usually comes before actions
  * `drawer-header` doesn't exist
    https://github.com/patternfly/patternfly/blob/9b4a99e89c65bc6bb0487daee0a2cd5c248cccfc/src/patternfly/components/Wizard/examples/Wizard.md?plain=1#L197-L206

mcoker (approved):
🎳

PR #7921, opened 2025-10-22
feat: Added Card description component.

### LINKED ISSUE #7914: Card - CardTitle description
Add new element for description within a CardTitle. This could be react-only if the `cardTitleText` doesn't need any additional wrapping elements added to house the description, i.e. the following is acceptable:
```
  <div className={css(styles.cardTitle)}>
      <div className={css(styles.cardTitleText, className)}>
        {children}
        <Content component="small">{description}</Content>
      </div>
  </div>
```

dlabaj:
Added card description for card component. Closes issue #7914 .


src/patternfly/components/Card/examples/Card.md
```diff
@@ -899,6 +899,7 @@ A card is a generic rectangular container that can be used to build other compon
 | `.pf-v6-c-card` | `<div>` | Creates a card component.  **Required** |
 | `.pf-v6-c-card__title` | `<div>` | Creates a card title container. |
 | `.pf-v6-c-card__title-text` | `<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`, `<h6>`, `<div>` | Creates a card title text element. |
+| `.pf-v6-c-card__description` | Creates a card description element within the card title. |
```
mcoker:
```suggestion
| `.pf-v6-c-card__description` | Creates the description of a card. |
```

I would just say "Creates the description of a card." I don't think we want to put it in the `<CardTitle>` because 1) we don't know what users will put in there and 2) the card title doesn't span the width of the card. Here's a screenshot of the description in `<CardTitle>`.

Image

We probably don't even want to put it in `<CardHeader>` because that is a flex row and doesn't wrap by default. IMO it would be a sibling of `<CardBody>` and come before any card body elements.

src/patternfly/components/Card/card--description.hbs
```diff
@@ -0,0 +1,9 @@
+<{{#if card--description--element}}{{card--description--element}}{{else}}small{{/if}} class="{{pfv}}card__description{{#if card--description--modifier}} {{card--description--modifier}}{{/if}}"
+  {{#if card--description--id}}
+    id="{{card--description--id}}"
+  {{/if}}
+  {{#if card--description--attribute}}
+    {{{card--description--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</{{#if card--description--element}}{{card--description--element}}{{else}}small{{/if}}>
```
mcoker:
```suggestion
<div class="{{pfv}}card__description{{#if card-description--modifier}} {{card-description--modifier}}{{/if}}"
  {{#if card-description--attribute}}
    {{{card-description--attribute}}}
  {{/if}}>
  {{> @partial-block}}
</div>
```

Here are the changes I made

* I don't think there is any reason that needs to support `<small>` at the moment - that can just be a div.
* To be consistent with other hbs naming, that file would be `card-description.hbs`
* All of the `card--description--[...]` stuff would be `card-description--[...]` to match the filename
* You shouldn't need the `--id` block for the time being, but no harm in leaving it, either, if you want to keep it.

src/patternfly/components/Card/card-title.hbs
```diff
@@ -11,4 +11,7 @@
   {{#if @partial-block}}
     {{> @partial-block}}
   {{/if}}
+  {{#if card-title--description}}
+    {{> card-title-description}}
+  {{/if}}
```
mcoker:
```suggestion
```

You shouldn't need this. You would just want to add a new example with a title like "With description" and add a description to the example card with code like below. Be sure to add `isBeta` after the opening backticks for the example (`` ```hbs isBeta ``)

```
{{#> card-description}}
  This is the card description.
{{/card-description}}

**Review by mcoker — COMMENTED** [review]
Nice! Made a few suggestions for the hbs part, just need a new example (mentioned in one of the comments) and then wire up the CSS.

I think for now the CSS can just set a small font size and a subtle text color and use the same padding that card body elements use. You can reference the radio/checkbox description or modal description for those font size and color styles/tokens.

cc @andrew-ronaldson @lboehling here's what it looks like with small and normal sized text for the description (with subtle text color). Do you have a preference for the font-size? Also is the spacing OK or does that need to tighten up between the title and description? It may be a little tricky to tighten that space, though we can do it if needed. Right now it's just equal space between the header/title + description + body elements, similar to [modal descriptions](https://www.patternfly.org/components/modal/html/with-description/) 

Image
Image

File: src/patternfly/components/Card/examples/Card.md [hunk 4]
```diff
@@ -899,6 +920,7 @@ A card is a generic rectangular container that can be used to build other compon
...
```
srambach:
```suggestion
| `.pf-v6-c-card__description` | Creates a card description element. |
```

srambach (commented):
Looks good! Just one minor nit in the docs wording.

srambach (approved):
L ⭐ TM!

src/patternfly/components/Card/examples/Card.md
```diff
@@ -62,6 +62,27 @@ import './Card.css'
   {{/card-footer}}
 {{/card}}
 ```
+### With title and description in head
```
mcoker: Tiny change requested, kind sir. We can just call out that it has a description, and slap a newline between this example and the one before it.

```suggestion

### With description
```

src/patternfly/components/Card/examples/Card.md
```diff
@@ -62,6 +62,27 @@ import './Card.css'
...
+```hbs isBeta
+{{#> card card--id="card-action-example-2"}}
```
mcoker: duplicate id with the one before it

```suggestion
{{#> card card--id="card-description"}}
```

src/patternfly/components/Card/card--description.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="{{pfv}}card__description{{#if card-description--modifier}} {{card-description--modifier}}{{/if}}"
+  {{#if card-description--attribute}}
+    {{{card-description--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
```
mcoker: we'll usually put a new line at the EOF

```suggestion
</div>

```

src/patternfly/components/Card/card.scss
```diff
@@ -29,6 +29,11 @@
   --#{$card}__header--m-wrap--RowGap: var(--pf-t--global--spacer--sm);
   --#{$card}__header--m-wrap--ColumnGap: var(--pf-t--global--spacer--md);
 
+  // Description
+  --#{$card}__description--FontSize: var(--pf-t--global--font--size--body--sm);
+  --#{$card}__description--Color: var(--pf-t--global--text--color--subtle);
```
mcoker: Hmm... did you mean to add these styles in? Right now they aren't applied so the description element looks the same as a regular card body element.

Assuming you do, you'll also want to add an entry in the block below for the description to set it to a disabled text color on a disabled card. https://github.com/dlabaj/patternfly/blob/31c10253b87d0ac9c3bd11c2b883b10134064c83/src/patternfly/components/Card/card.scss#L401-L408

mcoker (changes requested):
Nice! A few comments, the important one being that the color/font-size styles don't appear to be hooked up?

andrew-ronaldson:
Thanks Don!
I made a couple tweaks but will wait for @lboehling to share her thoughts.
I moved the Description inside the card header container and removed the inline padding so it was flush with the title. I had to change the padding top and bottom for the description as well as those values are on the right in the screenshot.
The description looks good with text--color--subtle and size as body--small. I noticed the catalog view has a similar subtitle but it uses a bold weight. I think I prefer regular but Lucia any thoughts?
Screenshot 2025-10-28 at 9 34 37 AM

src/patternfly/components/Card/card.scss
dlabaj: I updated the code too apply this style to the card description. Thanks for catching this.

srambach:
> Thanks Don! I made a couple tweaks but will wait for @lboehling to share her thoughts. I moved the Description inside the card header container and removed the inline padding so it was flush with the title. I had to change the padding top and bottom for the description as well as those values are on the right in the screenshot. 

FWIW @andrew-ronaldson, moving the description inside the `pf-v6-c-card__header-main` will make it wrap to leave space at the right for the actions, in case that matters to you. Putting it just inside `pf-v6-c-card__header` means we need to change the flex layout that is currently keeping the actions from interfering with the title.

mcoker:
@dlabaj chatted with Andrew and sorted it out. Can you update the className to `pf-v6-c-card__subtitle`, and add these styles (though create/use `card--m-subtitle` variables)

```css
.pf-v6-c-card__subtitle {
    color: var(--pf-t--global--text--color--subtle);
    font-size: var(--pf-t--global--font--size--body--sm);
    padding-top: var(--pf-t--global--spacer--sm);
}
```

Then you can add a new "With subtitle" example but show 2 cards in that example (you can put one or two `<br>`'s between them to create some space)

One card has the image/actions up top and the title/subtitle below. You can copy the [header images and actions example](https://www.patternfly.org/components/card#header-images-and-actions) 

Screenshot 2025-10-31 at 4 48 12 PM

The other is a regular card with actions. You can use the [title with inline images and actions example](https://www.patternfly.org/components/card#title-inline-with-images-and-actions), but no need for the long title, it can just say "Title"

Screenshot 2025-10-31 at 4 49 12 PM

src/patternfly/components/Card/card.scss
dlabaj: I removed these now when making the subtitle updates.

dlabaj:
@mcoker @andrew-ronaldson I made the updates per you're suggestions.

src/patternfly/components/Card/examples/Card.md
```diff
@@ -63,6 +63,51 @@ import './Card.css'
...
+  {{#> card-header}}
+    {{#> card-actions}}
+      {{> card--menu-toggle}}
+      {{> card--check}}
+    {{/card-actions}}
+    {{#> card-header-main}}
+      {{> card-title card-title-text--value="This is a really really really really really really really really really really long title" card-title-text--id=(concat card--id '-check-label')}}
+    {{/card-header-main}}
+  {{/card-header}}
+  {{#> card--subtitle}} This is the card subtitle. {{/card--subtitle}}
```
mcoker: Sorry about this but I forgot to mention that this subtitle should go inside of the card title, after the card title text. So the structure should be:

```
card__title
  card__title-text
  card__subtitle
```

And it only needs 3 styles - padding-block-start, font-size, color.

src/patternfly/components/Card/card-subtitle.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="{{pfv}}card__subtitle{{#if card-subtitle--modifier}} {{card-subtitle--modifier}}{{/if}}"
```
mcoker, line 1: Can you rename the file `card--subtitle.hbs` to `card-subtitle.hbs`?

src/patternfly/components/Card/examples/Card.md
```diff
@@ -63,6 +63,51 @@ import './Card.css'
 {{/card}}
 ```
 
+### With subititle
```
srambach:
```suggestion
### With subtitle
```
teeny nit 😄

src/patternfly/components/Card/card.scss
```diff
@@ -323,6 +323,12 @@
   overflow-wrap: var(--#{$card}__title-text--OverflowWrap);
 }
 
+.#{$card}__subtitle {
+  padding-block-start: var(--pf-t--global--spacer--sm);
+  font-size: var(--pf-t--global--font--size--body--sm);
+  color: var(--pf-t--global--text--color--subtle);
```
mcoker, line 329: Can you create component vars for these at the top of the scss file, then use those vars here?

```css
.#{$card} {
  --#{$card}__subtitle--PaddingBlockStart: var(--pf-t--global--spacer--sm);
  --#{$card}__subtitle--FontSize: var(--pf-t--global--font--size--body--sm);
  --#{$card}__subtitle--Color: var(--pf-t--global--text--color--subtle);
}

.#{$card}__subtitle {
  padding-block-start: var(--#{$card}__subtitle--PaddingBlockStart);
  font-size: var(--#{$card}__subtitle--FontSize);
  color: var(--#{$card}__subtitle--Color);
}
```

src/patternfly/components/Card/card.scss
```diff
@@ -335,6 +341,7 @@
   margin-inline-start: auto;
 
   + .#{$card}__title,
+  + .#{$card}__subtitle,
```
mcoker: You can remove this line. This element should never come after `__actions`

src/patternfly/components/Card/card.scss
```diff
@@ -412,6 +419,7 @@
 
 .#{$card}__header,
 .#{$card}__title,
+.#{$card}__subtitle,
```
mcoker: You can remove this line. It doesn't need any additional padding other than the block-start/top padding

mcoker (commented):
Left some comments. We'll probably also want to update `card-title.hbs` to accept a `card-subtitle--value` parameter, the same way it accepts `card-title-text--value`. I made the updates in this codepen if you would like a reference.

https://codepen.io/mcoker/pen/yyeGRzo

srambach (commented):
The spacing looks a little more than might be necessary but I'm not sure what design said.

srambach:
> The spacing looks a little more than might be necessary but I'm not sure what design said.

Oh nevermind I see other comments that are going to affect this!

mcoker (approved):
LGTM!

PR #7924, opened 2025-10-23
feat(tabs): add nav variant

### LINKED ISSUE #7911: Tabs - nav variant
Add new modifier for full navigation styling in Tabs.

srambach:
Fixes #7911 

Also has the old fashioned non-animated tab marker use the tab decoration color that the newer animated version uses.

### DIFF
--- a/src/patternfly/components/Tabs/tabs.scss
+++ b/src/patternfly/components/Tabs/tabs.scss
@@ -125,7 +125,7 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__link--after--BorderBlockStartWidth: 0;
   --#{$tabs}__link--after--BorderInlineEndWidth: 0;
   --#{$tabs}__link--after--BorderInlineStartWidth: 0;
-  --#{$tabs}__item--m-current__link--after--BorderColor: var(--pf-t--global--border--color--clicked);
+  --#{$tabs}__item--m-current__link--after--BorderColor: var(--#{$tabs}--link-accent--color);
   --#{$tabs}__item--m-current__link--after--BorderWidth: var(--pf-t--global--border--width--extra-strong);
 
   // Link accent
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Tabs/tabs.scss
```diff
@@ -125,7 +125,7 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__link--after--BorderBlockStartWidth: 0;
   --#{$tabs}__link--after--BorderInlineEndWidth: 0;
   --#{$tabs}__link--after--BorderInlineStartWidth: 0;
-  --#{$tabs}__item--m-current__link--after--BorderColor: var(--pf-t--global--border--color--clicked);
+  --#{$tabs}__item--m-current__link--after--BorderColor: var(--#{$tabs}--link-accent--color);
```
mcoker: Looks like this broke the existing accent borders since these vars now look to be defined by one another.

I don't think we have to fix this in this issue if that part is tricky. We can just theme `--pf-v6-c-tabs__item--m-current__link--after--BorderColor` instead, which will apply to both animated and default tabs.

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -205,6 +205,12 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__add--PaddingInlineStart: var(--pf-t--global--spacer--sm);
   --#{$tabs}__add--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
 
+  // Nav variant
+  --#{$tabs}--m-nav--inset: var(--pf-t--global--spacer--xl);
+  --#{$tabs}--m-nav__link-accent--color: var(--pf-t--color--red--50);
```
mcoker: in the designs, buttons and other tabs are also red, which looks like a separate theme or maybe a change to the brand color tokens. I'd say we can take this out of the default component (and https://github.com/srambach/patternfly/blob/e613d7842f53406daee5d83df3fbeb5e3d676e14/src/patternfly/components/Tabs/tabs.scss#L522) and set this as an override/theme in the demos and mention it in compass docs on how to style it.

Image
Image

src/patternfly/components/Tabs/tabs.scss
```diff
@@ -205,6 +205,12 @@ $pf-v6-c-tabs--spacer-map: build-spacer-map("none", "sm", "md", "lg", "xl", "2xl
   --#{$tabs}__add--PaddingInlineStart: var(--pf-t--global--spacer--sm);
   --#{$tabs}__add--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
 
+  // Nav variant
+  --#{$tabs}--m-nav--inset: var(--pf-t--global--spacer--xl);
```
mcoker: We can also take this out (and https://github.com/srambach/patternfly/blob/e613d7842f53406daee5d83df3fbeb5e3d676e14/src/patternfly/components/Tabs/tabs.scss#L527) and use the inset modifier/prop in our demos.

mcoker (changes requested):
Looks like default accent colors broke. Otherwise just a couple of small nits that you did right but the POC code was incorrect.

mcoker (approved):
LGTM!

Your last commit allowed setting `--#{$tabs}--link-accent--color` on regular and animated tabs to theme the accent color, but we lost the ability to set either `--#{$tabs}__item--m-current__link--after--BorderColor` or `--#{$tabs}__item--m-current__link--after--BorderColor` on animated tabs to theme it. Since those worked before this change, and would have likely been how someone had themed that accent color, that's breaking. 

Pushed an update to address that. It's basically just the way the CSS was originally for the accent color, but now the non-animated tabs use `--link-accent--color`, so you can set any of those 3 vars above on both animated and non-animated tabs and it themes the accent color.

PR #7926, opened 2025-10-23
feat(Common): added gradient border and pulse animation

### LINKED ISSUE #7915: Button/utility - ai border & thinking styles
We should ship a more generic ai border and thinking modifiers to use with components other than chatbot's  MessageBar (such as Button). So I think these could exist as Button-specific modifiers or utility.

thatblindgeye:
Closes #7915

### DIFF
--- a/src/patternfly/base/patternfly-common.scss
+++ b/src/patternfly/base/patternfly-common.scss
@@ -68,9 +68,59 @@
   // stylelink-enable declaration-no-important
 }
 
+.#{$pf-prefix}m-gradient-border {
+  position: relative;
+
+  &::before {
+    content: "";
+    position: absolute;
+    inset: 0;
+    pointer-events: none;
+    border: var(--pf-t--global--border--width--extra-strong) solid transparent;
+    border-radius: inherit;
+    background: linear-gradient(to right, #f56e6e 0%, #876fd4 65%, #5e40be 100%)
+      border-box;
+    mask: linear-gradient(#000 0 0) padding-box, linear-gradient(#000 0 0);
+    mask-composite: exclude;
+  }
+}
+
+@property --#{$pf-global}--pulse--BoxShadow--Spread {
+  syntax: "<length>";
+  initial-value: 0px;
+  inherits: false;
+}
+
+.#{$pf-prefix}m-pulse {
+  &,
+  &:focus,
+  &:focus-within {
+    animation: #{$pf-global}-pulse 2s infinite ease-in-out;
+    box-shadow: 0px 0px var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(43.2% 0.232 292.759),
+    0px 0px var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(50% 0.134 242.749),
+    0px 0px var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(45.9% 0.187 3.815),
+    0px 0px var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(45.9% 0.187 3.815);
+  }
+
+  &.pf-m-inset {
+    box-shadow: inset 0px 0px var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(43.2% 0.232 292.759),
+    inset 0px 0px var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(50% 0.134 242.749),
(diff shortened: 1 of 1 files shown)


src/patternfly/base/patternfly-common.scss
```diff
@@ -68,9 +68,59 @@
...
+
+.#{$pf-prefix}m-pulse {
+  &,
+  &:focus,
+  &:focus-within {
+    box-shadow: 0 0 var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(43.2% 0.232 292.759deg),
+    0 0 var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(50% 0.134 242.749deg),
+    0 0 var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(45.9% 0.187 3.815deg),
+    0 0 var(--#{$pf-global}--pulse--BoxShadow--Spread) oklch(45.9% 0.187 3.815deg);
+    animation: #{$pf-global}-pulse 2s infinite ease-in-out;
```
srambach: I would consider making the duration a variable.

srambach (commented):
Besides the duration variable, are we going to turn it off for pf-m-no-motion or change/stop it for prefers-reduced-motion?

src/patternfly/base/patternfly-common.scss
```diff
@@ -68,9 +68,59 @@
...
+@property --#{$pf-global}--pulse--BoxShadow--Spread {
+  syntax: "<length>";
+  initial-value: 0;
+  inherits: false;
+}
+
+.#{$pf-prefix}m-pulse {
+  &,
+  &:focus,
+  &:focus-within {
```
mcoker: I don't think you need these? I believe they were in the POC initially to fire on focus, but this animation turn on/off by toggling the class.

mcoker (commented):
I think it's fine as-is. Updates I think we could make at some point:
* The `:focus/:focus-within` styles I mentioned
* Semantic names for these classes named after their role instead of how they look
* vars for the colors and parts that can be customized
* Sarah's comments about no/reduced motion and if there are existing vars we can use

wdyt @thatblindgeye want to update anything or add these to the follow-up issue to do later? https://github.com/patternfly/pf-roadmap/issues/269

thatblindgeye:
@srambach @mcoker made some updates per above:

- Added ability to stop the pulse animation when applicable classes/OS level settings are applied
- Put pulse box shadow colors in vars (named them based sort of where each shadow appears around the box)
- Removed focus selectors

The rest I think would be good for a followup, the class/var naming and the addition of any vars for the gradient border colors/values (wasn't too sure really what to name them or whether to lump the entire linear-gradients in a single var or each color value/stop % its own var)

src/patternfly/base/patternfly-common.scss
```diff
@@ -64,13 +64,69 @@
   --pf-t--global--duration--200: 1ms !important;
   --pf-t--global--duration--100: 1ms !important;
   --pf-t--global--duration--50: 1ms !important;
+  --#{$pf-global}--pulse-pulsing--Animation--Duration: 0 !important;
 
   // stylelink-enable declaration-no-important
 }
 
+.#{$pf-prefix}m-gradient-border {
```
mcoker: I pinged @lboehling about this and since this is something that initiates an AI chat/interaction, she recommended "ai-trigger" or "ai-indicator".

```suggestion
.#{$pf-prefix}m-ai-trigger {
```

mcoker: I chatted with @lboehling about this and she recommended `.pf-m-ai-trigger` or `.pf-m-ai-indicator` since this is a style that indicates it will trigger some sort of AI action (like pulls up an assistant chat or something).

src/patternfly/base/patternfly-common.scss
```diff
@@ -64,13 +64,69 @@
...
+  }
+}
+
+@property --#{$pf-global}--pulse--BoxShadow--Spread {
+  syntax: "<length>";
+  initial-value: 0;
+  inherits: false;
+}
+
+.#{$pf-prefix}m-pulse {
```
mcoker: Ran this by @lboehling too and she was on board with `.pf-m-thinking` since this indicates the AI is working/processing.

mcoker (commented):
LGTM just a couple of super small nits on the class naming after running it by @lboehling

mcoker (approved):
✨

PR #7927, opened 2025-10-23
feat(compass): added component

### LINKED ISSUE #7908: Glass styling
Implement the first pass at glass styling and note its potential caveats with our current components (how they interact and lay on top of each other and the background image). 

Currently, it's been noted that the glass effect (transparent blurred background):
- Can only be applied to a single element in a hierarchy (children cannot also be blurred)
- Doesn't play well with certain other styles options like `z-index`, `transform`
- May not be accessible depending on the background image with element color
- May not be performant

May be good as a utility class, or as modifier(s) to specific components. Note which components may need plain variants as part of this effort to populate out #7910.

### LINKED ISSUE #7909: Compass - new layout components
Harden and create a new suite of compass layout components:

- Compass wrapper with NSEW + center areas
- Center content components:
  - Hero (could be a section variant)
  - Sections
  - Grouped sections (internally may be a flex or grid)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7909


src/patternfly/components/Compass/examples/Compass.md
```diff
@@ -0,0 +1,32 @@
...
+    {{#> compass-content}}
+      {{#> compass-section}}
+        section
+      {{/compass-section}}
+    {{/compass-content}}
+  {{/compass-main}}
+  {{#> compass-panel compass--panel--IsEnd=true}}panel start{{/compass-panel}}
+  {{#> compass-footer}}footer{{/compass-footer}}
+{{/compass}}
+```
```
kmcfaul, line 536: Should we have compass-nav, compass-panel (start and end), compass-main-header, and compass-footer have child compass-sections?

mcoker: yep added

src/patternfly/components/ActionList/examples/ActionList.md
```diff
@@ -148,7 +148,7 @@ In wizards
 ### Vertical action list
 ```hbs isBeta
 Multiple groups
-{{#> action-list action-list--isVertical="true"}}
+{{#> action-list action-list--IsVertical="true"}}
```
thatblindgeye, line 151: Do we wanna remove the quotes around these booleans while we're here?

src/patternfly/components/Tabs/tabs.hbs
```diff
@@ -7,7 +7,7 @@
   {{~#if tabs--HasNoBorderBottom}} pf-m-no-border-bottom{{/if}}
   {{~#if __tabs-list--IsOverflow}} pf-m-overflow{{/if}}
   {{~#if tabs--IsAnimateCurrent}} pf-m-animate-current{{/if}}
-  {{~#if tabs--isNav}} pf-m-nav{{/if}}
+  {{~#if tabs--IsNav}} pf-m-nav{{/if}}
```
thatblindgeye, line 10: This depends what the intended structure is, but we should ensure a `nav` element wraps the Tab being used. It would either live in this file or in the compass_nav file.

If it lived here it could be a followup since we don't currently support it, but we would want to update the wrapper element to be a nav element when IsNav is true (rather than needing to also pass ing `tabs--type="nav"`). Would also need to not apply the role of region on line 16 when IsNav is true so as to not override the nav semantics.

Might make sense to live here so that it could go into React as well.

src/patternfly/components/Compass/compass-footer.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="{{pfv}}compass__footer{{#if compass-footer--modifier}} {{compass-footer--modifier}}{{/if}}"
```
thatblindgeye, line 1: So I'm wondering whether some of these components should use their semantic elements (footer element here, and header and main elsewhere).

Issue is the compass__panel components living outside header, footer, and main. I'm not sure if putting them in `aside` elements would make the most sense (they sorta do I think).

Probably fine to leave them as plain divs for now, but worth mulling over some more, especially because I assume we need (or at least want) the structure of Compass to be as it is now: header, panel, main, panel, and footer all being adjacent siblings (rather than main containing the 2 panels).

thatblindgeye (commented):
The file comments below this may not necessarily be anything to resolve right now, more thoughts I had looking this over. Let me know what you think.

src/patternfly/components/Tabs/tabs.hbs
mcoker: https://github.com/patternfly/patternfly/issues/7941

src/patternfly/components/Compass/compass-footer.hbs
mcoker: https://github.com/patternfly/patternfly/issues/7940

PR #7928, opened 2025-10-23
fix(card): remove padding/margin on clickable cards action element

### LINKED ISSUE #7899: Bug - clickable card actions take up space when invisible
the styles I had to override in the compass POC were `.pf-v6-c-card.pf-m-clickable .pf-v6-c-card__actions { margin: 0; padding; 0; }`

You can go to a clickable card that has no visible radio/checkbox in the top right (they would be in `.pf-v6-c-card__actions`) and see that the actions div takes up space even though there is nothing in it

Image

Image

wise-king-sullyman:
Closes #7899

### DIFF
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -344,6 +344,11 @@
     --#{$card}__actions--MarginBlockStart: 0;
     --#{$card}__actions--MarginBlockEnd: 0;
   }
+
+  &:has(button, a) {
+    padding: 0;
+    margin: 0;
+  }
 }
 
 // Labels are used to handle states of selectable and clickable cards
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -196,6 +196,14 @@
   }
   // stylelint-enable selector-max-class
 
+  // when a card is clickable only, or has the screen reader class on its input
+  &.pf-m-clickable:not(.pf-m-selectable),
+  &:has(input.#{$pf-prefix}u-screen-reader) {
+    --#{$card}__actions--PaddingInlineStart: 0;
+    --#{$card}__actions--MarginBlockStart: 0;
+    --#{$card}__actions--MarginBlockEnd: 0;
+  }
+
   &.pf-m-compact {
     --#{$card}__title-text--FontSize: var(--#{$card}--m-compact__title-text--FontSize);
     --#{$card}__body--FontSize: var(--#{$card}--m-compact__body--FontSize);
@@ -343,11 +351,6 @@
   &.pf-m-no-offset {
     --#{$card}__actions--MarginBlockStart: 0;
     --#{$card}__actions--MarginBlockEnd: 0;
-  }
-
-  &:has(button, a) {
-    padding: 0;
-    margin: 0;
   }
 }
 
--- a/src/patternfly/components/Card/card.scss
+++ b/src/patternfly/components/Card/card.scss
@@ -196,9 +196,9 @@
   }
   // stylelint-enable selector-max-class
 
-  // when a card is clickable only, or has the screen reader class on its input
-  &.pf-m-clickable:not(.pf-m-selectable),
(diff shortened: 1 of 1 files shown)


thatblindgeye (commented):
LGTM, just thinking that wwe probably want to make an update for when a selectable input is also hidden, like in the React [card as tiles](https://www.patternfly.org/components/card#single-selectable-tiles) examples. @mcoker

src/patternfly/components/Card/card.scss
```diff
@@ -344,6 +344,11 @@
     --#{$card}__actions--MarginBlockStart: 0;
     --#{$card}__actions--MarginBlockEnd: 0;
   }
+
+  &:has(button, a) {
+    padding: 0;
+    margin: 0;
+  }
```
srambach: If I'm reading the issue correctly, this should be scoped just to the pf-m-clickable card. As is, this is removing padding and margin for any card actions that contain a link or button - which is probably all or most of them, right?

mcoker:
> we probably want to make an update for when a selectable input is also hidden, like in the React [card as tiles](https://www.patternfly.org/components/card#single-selectable-tiles) examples.

Yep, any time the actions element is in the DOM but invisible, it shouldn't impact the layout of the card.

src/patternfly/components/Card/card.scss
```diff
@@ -196,6 +196,14 @@
   }
   // stylelint-enable selector-max-class
 
+  // when a card is clickable only, or has the screen reader class on its input
+  &.pf-m-clickable:not(.pf-m-selectable),
+  &:has(input.#{$pf-prefix}u-screen-reader) {
```
mcoker: You'd probably want a more narrow scope on the `input` to be able to handle nested cards. `> .#{$card}__selectable-actions` is optional but I'd include it because who knows if someone has some other thing in their card actions that could trigger this style change.

```suggestion
  &.pf-m-clickable:not(.pf-m-selectable),
  &:has(> .#{$card}__header > .#{$card}__actions > .#{$card}__selectable-actions input.#{$pf-prefix}u-screen-reader) {
```

It's probably more efficient to write it this way though (not tested) since targeting the input only needs to start at the `__actions` container?

```suggestion
  @at-root .#{$card}__actions:has(> .#{$card}__selectable-actions input.#{$pf-prefix}u-screen-reader),
  &.pf-m-clickable:not(.pf-m-selectable) {
```

mcoker (commented):
LGTM! Just one little nit so it would work better with nested cards.

mcoker:
@srambach @wise-king-sullyman I pushed the update from [my comment about updating the scope](https://github.com/patternfly/patternfly/pull/7928#discussion_r2462211944) in this commit https://github.com/patternfly/patternfly/pull/7928/commits/fbe2de115933e778fbc28a285a285ff6c762101f

Also it removes the `u-` from `input.#{$pf-prefix}u-screen-reader` and uses `input.#{$pf-prefix}screen-reader` since the react component uses `pf-v6-screen-reader`, not the utility class.

To verify that it works with tiles, I copied/pasted the HTML from the [react cards as tiles example](https://www.patternfly.org/components/card#cards-as-tiles) into the core preview with these changes.

src/patternfly/components/Card/card.scss
```diff
@@ -196,6 +196,14 @@
   }
   // stylelint-enable selector-max-class
 
+  // when the whole card is clickable and the card actions are invisible, remove the padding/margin so it doesn't take up space
+  @at-root .#{$card}__actions:has(> .#{$card}__selectable-actions input.#{$pf-prefix}screen-reader),
```
srambach, line 200: My only comment would be maybe it's worth a comment here to explain what this selector is aiming to do since it's pretty obscure and really only in react.

srambach (approved):
Visual regressions on card looked good. Just one suggestion to add a comment if you like.

mcoker (approved):

mcoker:
@srambach thanks for running visual regressions!!

PR #7925, opened 2025-10-23
feat(table): add plain variant

srambach:
For #7910, adds the plain variant just for table

PR #7929, opened 2025-10-24
feat(data-list): added plain variant

### LINKED ISSUE #7910: Misc - plain variants
Page, PageSection, Drawer, Table and others may need plain variants to support glass.

mcoker:
towards #7925

### DIFF
--- a/src/patternfly/components/DataList/examples/DataList.md
+++ b/src/patternfly/components/DataList/examples/DataList.md
@@ -1074,6 +1074,42 @@ When a list item includes more than one block of content, it can be difficult fo
 | `.pf-m-break-word` | `.pf-v6-c-data-list`, `.pf-v6-c-data-list__item-row`, `.pf-v6-c-data-list__cell`, `.pf-v6-c-data-list__text` | Modifies the data list element so that text breaks to the next line. |
 | `.pf-m-nowrap` | `.pf-v6-c-data-list`, `.pf-v6-c-data-list__item-row`, `.pf-v6-c-data-list__cell`, `.pf-v6-c-data-list__text` | Modifies the data list element so that text does not wrap to the next line. |
 
+### Plain
+```hbs
+{{#> data-list data-list--IsPlain=true data-list--id="data-list-basic" data-list--attribute='aria-label="Basic data list example"'}}
+  {{#> data-list-item data-list-item--id="item-1"}}
+    {{#> data-list-item-row}}
+      {{#> data-list-item-content}}
+        {{#> data-list-cell}}
+          <span id="{{data-list--id}}-{{data-list-item--id}}">Primary content</span>
+        {{/data-list-cell}}
+        {{#> data-list-cell}}
+          Secondary content
+        {{/data-list-cell}}
+      {{/data-list-item-content}}
+    {{/data-list-item-row}}
+  {{/data-list-item}}
+
+  {{#> data-list-item data-list-item--id="item-2"}}
+    {{#> data-list-item-row}}
+      {{#> data-list-item-content}}
+        {{#> data-list-cell data-list-cell--modifier="pf-m-no-fill"}}
+          <span id="{{data-list--id}}-{{data-list-item--id}}">Secondary content (pf-m-no-fill)</span>
+        {{/data-list-cell}}
+        {{#> data-list-cell data-list-cell--modifier="pf-m-no-fill pf-m-align-right"}}
+          Secondary content (pf-m-align-right pf-m-no-fill)
+        {{/data-list-cell}}
+      {{/data-list-item-content}}
+    {{/data-list-item-row}}
+  {{/data-list-item}}
+{{/data-list}}
+```
+
+### Usage
+
+| Class | Applied to | Outcome |
+| -- | -- | -- |
+| `.pf-m-plain` | `.pf-v6-c-data-list` | Modifies a data list to have a transparent background. |
 
 ## Documentation
 ### Overview
--- a/src/patternfly/components/DataList/examples/DataList.md
+++ b/src/patternfly/components/DataList/examples/DataList.md
@@ -1075,7 +1075,7 @@ When a list item includes more than one block of content, it can be difficult fo
 | `.pf-m-nowrap` | `.pf-v6-c-data-list`, `.pf-v6-c-data-list__item-row`, `.pf-v6-c-data-list__cell`, `.pf-v6-c-data-list__text` | Modifies the data list element so that text does not wrap to the next line. |
 
 ### Plain
-```hbs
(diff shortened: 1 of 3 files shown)


src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -1074,6 +1074,42 @@ When a list item includes more than one block of content, it can be difficult fo
 | `.pf-m-break-word` | `.pf-v6-c-data-list`, `.pf-v6-c-data-list__item-row`, `.pf-v6-c-data-list__cell`, `.pf-v6-c-data-list__text` | Modifies the data list element so that text breaks to the next line. |
 | `.pf-m-nowrap` | `.pf-v6-c-data-list`, `.pf-v6-c-data-list__item-row`, `.pf-v6-c-data-list__cell`, `.pf-v6-c-data-list__text` | Modifies the data list element so that text does not wrap to the next line. |
 
+### Plain
+```hbs
```
thatblindgeye: Can we add a general core specific issue to the [Compass followup epic](https://github.com/patternfly/pf-roadmap/issues/269) to ensure we provide example and classname table documentation for new features/styles? Just something for us to track to double check we're providing docs where we would need to across everything we're adding, so not just this specific addition.

thatblindgeye (approved):
LGTM, just something we should add to the followup doc below

src/patternfly/components/DataList/examples/DataList.md
mcoker: @thatblindgeye sounds good! Just FYI for the plain example, the class docs are right below the example. Data-list is documented similar to table where there are little tables below each example section with docs versus a big table at the bottom of the component with all of the classes.

mcoker:
@thatblindgeye would you mind giving this another look? I added `isBeta` to the plain docs and updated the heading hierarchy for data-list examples so it matches table and is more readable.

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -60,7 +59,8 @@ cssPrefix: pf-v6-c-data-list
 | `.pf-m-no-fill` | `.pf-v6-c-data-list__cell` | Modifies a data list cell to not fill the available horizontal space. |
 | `.pf-m-align-right` | `.pf-v6-c-data-list__cell` | Modifies a data list cell to align-right. |
 
-### With headings
+## Data list with headings
+### Data list with headings example
```
thatblindgeye, line 61: Doesn't have to be done here, but as a nit with the heading levels, wew could also update the headings used in the example code from h2's to h5's to continue following the hierarchy (after updating this example title to an h4)

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -632,7 +629,8 @@ When a list item includes more than one block of content, it can be difficult fo
 | `.pf-m-no-padding` | `.pf-v6-c-data-list__expandable-content-body` | Removes padding for the expandable content body. |
 | `.pf-m-icon` | `.pf-v6-c-data-list__cell` | Modifies a data list cell to not grow and align-left when its the first data-list__cell element. |
 
-### Modifiers
+## Data list modifiers
+### Data list flex modifiers example
```
thatblindgeye, line 631: Similar to the above, we could also update the h2s in this example to h5s

thatblindgeye (changes requested):
New heading structure looks good, we just need to increase each heading by 1. The component name should be the h1 on the page, with the top most "Examples" heading being an h2 and then following from there an h3 for the "example grouping" and an h4 for the actual example/sub-section title.

src/patternfly/components/DataList/examples/DataList.md
thatblindgeye: Just to expand on this and the below comment, in the actual example code we currently have stuff like:

```
{{#> data-list-cell}}
  <h2 id="{{data-list--id}}-{{data-list-item--id}}">Primary content</h2>
{{/data-list-cell}}
```

(Link: https://github.com/patternfly/patternfly/blob/9f4c365d3fcb48c436fdee76cb1dd521a82c4b0a/src/patternfly/components/DataList/examples/DataList.md?plain=1#L69-L71)

Which is fine within the context of the example, but since our example render in the natural flow of the rest of the page, we end up with essentially something like:

```
h1: "Component name"
h2: "Examples"
h3: "Example title"
h2: "Primary content"
...
h3: "Usage"
```
Which ends up making the "Usage" section of the page seem like a sub-heading of a heading only used within the example. If we update the example code in these 2 example I called out here so that we do:

```
{{#> data-list-cell}}
  <h4 id="{{data-list--id}}-{{data-list-item--id}}">Primary content</h2>
{{/data-list-cell}}
```

(keeping in mind the tweak we talked about on Meet briefly, we'd want to use an h4 since the example titles would remain h3's), it would result in a more logical heading hiearchy.

mcoker: Ah! That makes sense. I'd advocate we remove the headings in the examples if at all possible.

src/patternfly/components/DataList/examples/DataList.md
```diff
@@ -1254,8 +1250,11 @@ The DataList component provides a flexible alternative to the Table component, w
...
-### Usage
-
+### Data list as grid usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-m-grid{-[none, sm, md, lg, xl, 2xl]}` | `.pf-v6-c-data-list` | Modifies the data list to switch to a grid layout at a specified [breakpoint](/tokens/all-patternfly-tokens). `.pf-m-grid` will display the grid layout at all breakpoints. `.pf-m-grid-none` will display the desktop layout at all breakpoints. **Note:** Without a grid modifier, the data list will display the grid layout by default and switch to the desktop layout at a medium breakpoint. |
+
+## Documentation
+### Overview
+The DataList component provides a flexible alternative to the Table component, wherein individual data points may or may not exist within each row. DataList relies upon PatternFly layouts to achieve desired presentation within `pf-v6-c-data-list__cell`s. DataLists do not have headers. If headers are required, use the [table component](/components/table).
```
srambach:
```suggestion
The DataList component provides a flexible alternative to the Table component, wherein individual data points may or may not exist within each row. DataList relies upon PatternFly layouts to achieve desired presentation within `pf-v6-c-data-list__cell` elements. DataLists do not have headers. If headers are required, use the [table component](/components/table).
```
super nitpicky!

srambach (approved):
Looks good to me, nitpicky wording suggestion can be taken or ignored!

thatblindgeye (approved):
I'm good with the example code specific comments above being a followup, otherwise things look good

PR #7945, opened 2025-10-30
feat(drawer): add pill variant

### LINKED ISSUE #7917: Drawer - pill variant
Add new styles for a `pill` modifier to Drawer:

- Adds a border-radius to the Drawer content.
- With overlay drawers, it should also apply an inset.
- With inline drawers, it should also apply a gap between drawer content section.
- For resizable drawers, it should also modify the splitter styles.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7917

Backstop report - [backstop-drawer-pill.pdf](https://github.com/user-attachments/files/23224950/backstop-drawer-pill.pdf)

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -23,7 +23,6 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   // Panel
   --#{$drawer}__panel--MinWidth: 50%; // change to __panel--md--MinWidth at breaking change
   --#{$drawer}__panel--MaxHeight: auto;
-  --#{$drawer}__panel--AlignSelf: initial;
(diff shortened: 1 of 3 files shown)


src/patternfly/components/Drawer/drawer.scss
```diff
@@ -236,7 +248,7 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
     > .#{$drawer}__main > .#{$drawer}__panel {
       @include pf-v6-bidirectional-style(
         $prop: transform,
-        $ltr-val: translateX(-100%),
+        $ltr-val: translateX(calc(-100% - var(--#{$drawer}--m-expanded__panel--inset))),
         $rtl-val: translateX(#{pf-v6-calc-inverse(-100%)}),
```
srambach, line 252: You could adjust this just like the ltr value.

srambach (approved):
Good with followups in https://github.com/patternfly/patternfly/issues/7947

Issue #7827, opened 2025-09-11
Text Links: Update text decoration styling


andrew-ronaldson:
Follow up to https://github.com/patternfly/patternfly-design-kit/issues/893

Based on feedback we will be updating our text decoration styling to be aligned with RHDS as seen here 
https://ux.redhat.com/foundations/interactions/links/

mcoker:
@andrew-ronaldson were you wanting this to get in the Q3 release or ok to target for Q4?

srambach:
@andrew-ronaldson @lboehling can you clarify whether this affects text decoration other than `<a>` links? E.g., there are custom text decorations on:

- banner (with inline link button)
- breadcrumb (with buttons)
- button inline links
- description list (term help text)
- label (editable)
- progress stepper (with help text)
- timestamp (help text variant)

Note - at least button inline links should probably match since they are often used in the same places as an `<a>` link. Most of the other customizations are changing the style to dashed, so if that's our default I'm not sure if you want helper text etc to be dotted or something else different than dashed.

Also - there's an animation on the RHDS link text decoration - should we include this and if so do we need reduced motion or is it ok?

srambach:
Closed by #7948 
Per design, we'll create follow ups to make any necessary changes to text decoration on other elements.

PR #7948, opened 2025-10-30
fix(anchor link): update text decoration styles

### LINKED ISSUE #7907: Tokens - anchor link text decoration
Add new token for anchor link text decoration, to allow syncing with unified design system's dashed underline text decoration.

We have an existing issue to update the links in PF with the same styling here - https://github.com/patternfly/patternfly/issues/7827

We should check with design and see if this is a global change we want to make to patternfly (considering people may see it as visually breaking) or if we just extend link style vars a bit so it's easier to theme to match the new link styles.

srambach:
Fixes #7907 and #7827 

*Note:* Needs token updates; see local tokens.

Makes the anchor link text decoration match RHDS. Affects anchor tags and inline buttons.

Open question: should custom styling on text decoration elsewhere be updated in any way?

Questions copied from issue #7827:
@andrew-ronaldson @lboehling can you clarify whether this affects text decoration other than <a> links? E.g., there are custom text decorations on:

banner (with inline link button)
breadcrumb (with buttons)
(done) button inline links
description list (term help text)
label (editable)
progress stepper (with help text)
timestamp (help text variant)
Note - at least button inline links should probably match since they are often used in the same places as an <a> link. Most of the other customizations are changing the style to dashed, so if that's our default I'm not sure if you want helper text etc to be dotted or something else different than dashed.

Also - there's an animation on the RHDS link text decoration - should we include this and if so do we need reduced motion or is it ok?


.claude/settings.local.json
```diff
@@ -0,0 +1,10 @@
+{
```
mcoker: Not sure if you meant to check this file in?

src/patternfly/base/normalize.scss
```diff
@@ -108,12 +108,52 @@
   :where(a) {
     color: var(--pf-t--global--text--color--link--default);
     text-decoration-line: var(--pf-t--global--text-decoration--link--line--default);
+    text-decoration-style: var(--pf-t--global--text-decoration--link--style--default);
+    text-decoration-color: var(--pf-t--global--text-decoration--color--default);
+    text-underline-offset: var(--pf-t--global--text-decoration--offset--default);
+
+    // RHDS uses: text-underline-offset: max(var(--pf-t--global--text-decoration--offset--default), 0.28em);
```
mcoker: Could you write it like this?

```suggestion
    text-underline-offset: var(--pf-t--global--text-decoration--offset--default); // RHDS uses: text-underline-offset: max(var(--pf-t--global--text-decoration--offset--default), 0.28em);
```

src/patternfly/components/Button/button.scss
```diff
@@ -530,12 +524,18 @@
       --#{$button}--PaddingInlineEnd: var(--#{$button}--m-link--m-inline--PaddingInlineEnd);
       --#{$button}--PaddingBlockEnd: var(--#{$button}--m-link--m-inline--PaddingBlockEnd);
       --#{$button}--PaddingInlineStart: var(--#{$button}--m-link--m-inline--PaddingInlineStart);
+      --#{$button}--TextUnderlineOffset: var(--#{$button}--m-link--m-inline--TextUnderlineOffset);
+      --#{$button}--hover--TextUnderlineOffset: var(--#{$button}--m-link--m-inline--hover--TextUnderlineOffset);
```
mcoker: Looks like `--#{$button}--TextUnderlineOffset` and `--#{$button}--hover--TextUnderlineOffset` are not defined.

src/patternfly/components/Button/button.scss
```diff
@@ -530,12 +524,18 @@
...
+      --#{$button}--TransitionDuration: var(--#{$button}--m-link--m-inline--TransitionDuration);
+      --#{$button}--TransitionTimingFunction: var(--#{$button}--m-link--m-inline--TransitionTimingFunction);
```
mcoker, line 529: Looks like `--#{$button}--m-link--m-inline--TransitionTimingFunction` is not defined.

src/patternfly/components/Button/button.scss
```diff
@@ -530,12 +524,18 @@
...
+      --#{$button}--TransitionProperty: var(--#{$button}--m-link--m-inline--TransitionProperty);
```
mcoker, line 530: This removes the `color` transition. And if you add `color` back, you'll also want to check `--#{$button}--TransitionDuration: var(--#{$button}--m-link--m-inline--TransitionDuration);` since that updated to the duration used for the underline offset change.

src/patternfly/components/Button/button.scss
```diff
@@ -530,12 +524,18 @@
...
       --#{$button}--BackgroundColor: transparent;
       --#{$button}__progress--InsetInlineStart: var(--#{$button}--m-link--m-inline__progress--InsetInlineStart);
       --#{$button}--TextDecorationLine: var(--#{$button}--m-link--m-inline--TextDecorationLine);
       --#{$button}--TextDecorationStyle: var(--#{$button}--m-link--m-inline--TextDecorationStyle);
       --#{$button}--hover--TextDecorationLine: var(--#{$button}--m-link--m-inline--hover--TextDecorationLine);
       --#{$button}--hover--TextDecorationStyle: var(--#{$button}--m-link--m-inline--hover--TextDecorationStyle);
+      --#{$button}--hover--TextDecorationColor: var(--#{$button}--m-link--m-inline--hover--TextDecorationColor);
```
mcoker: Looks like this is a dupe, overridden by L547 https://github.com/srambach/patternfly/blob/7e6002fb15b4a9120d913efc72ea4dfc46278d02/src/patternfly/components/Button/button.scss#L547

mcoker (approved):
LGTM, left some comments but they're not a big deal and can be fixed in a follow-up if you'd prefer.

Also the content component has its own link styles, including some text-decoration stuff, but I'm not sure why it has styles beyond `color` (to support `:visited`) and doesn't just inherit link styles. So IMO we don't need to make any more updates there. I'd even propose we remove the text-decoration stuff in the content component at some point and let it inherit the global styles.

andrew-ronaldson:
One small nit is that it uses `border--color--hover` but in the RHDS version it uses `text--color--hover` (darker).

src/patternfly/components/Button/button.scss
srambach: Removed and I use the inline link variable directly since it's not needed for the base variant.

src/patternfly/components/Button/button.scss
srambach: I duplicated the color transition settings - not ideal but using those on the text decoration looked really clunky.

src/patternfly/components/Button/button.scss
```diff
@@ -552,9 +550,14 @@
       --#{$button}--m-clicked--BorderWidth: 0;
 
       text-align: start;
+      text-underline-offset: var(--#{$button}--m-link--m-inline--TextUnderlineOffset);
       white-space: normal;
       background: transparent; // don't show the ripple effect for inline links
       outline-offset: #{pf-size-prem(2px)};
+
+      &:hover {
```
```suggestion
      &:hover,
      &:focus {
```

mcoker (approved):
🚀 Just one tiny nit, I'll apply the suggestion and merge.

PR #7963, opened 2025-11-05
feat(drawer): add support for glass panel, add comapss demo

### LINKED ISSUE #7958: Drawer - add support for glass panel
Drawer panel should be able to have glass effect.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7958

This PR does these things:
* Creates a new compass "with drawer" demo
* Creates a `.pf-m-glass` drawer panel variation
* Creates tokens for the panel background attributes - background color, background color opacity, blur. I left the blur token as `--filter` implying you could change that use other filters, too.
  * Also handles dark theme and a "no glass" theme. No glass will likely change this upcoming sprint to "glass", with "no glass" being the new default styles.
* Implements those tokens in the compass panel and the glass drawer panel.

### DIFF
--- a/src/patternfly/base/tokens/tokens-local.scss
+++ b/src/patternfly/base/tokens/tokens-local.scss
@@ -133,5 +133,27 @@
 
   // stylelint-disable custom-property-pattern
   --pf-t--temp--dev--tbd: #BC11E0;
+  // stylelint-enable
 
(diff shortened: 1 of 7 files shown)


srambach (approved):
L 🥂 TM

PR #7965, opened 2025-11-05
fix(links): fix decoration color

### LINKED ISSUE #7959: Bug - Links - wrong token used
Changes to link text decoration as part of #7948 used the wrong token. 

Should be:
inline-link (default)
font-color: --pf-t--global--text--color--link--default
 text-decoration-color: --pf-t--global--border--color--default

inline-link (hover)
font-color: --pf-t--global--text--color--link--hover
 text-decoration-color: --pf-t--global--text--color--link--hover

srambach:
Fixes #7959

Changes back to border color for the text decoration, and changes hover color to currentcolor - but kept in a comment on the temporary token saying what token it should be matching, just for reference until the new tokens are added for real.

### DIFF
--- a/src/patternfly/base/tokens/tokens-local.scss
+++ b/src/patternfly/base/tokens/tokens-local.scss
@@ -3,8 +3,8 @@
 @mixin pf-v6-tokens {
   // temporary until we have updates from design 
   // new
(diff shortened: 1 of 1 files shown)


mcoker (approved):
🔗

PR #7974, opened 2025-11-06
chore(Tabs): applied nav for IsNav

### LINKED ISSUE #7941: Tabs - IsNav updates
https://github.com/patternfly/patternfly/pull/7927#discussion_r2470684125

> This depends what the intended structure is, but we should ensure a nav element wraps the Tab being used. It would either live in this file or in the compass_nav file.

If it lived here it could be a followup since we don't currently support it, but we would want to update the wrapper element to be a nav element when IsNav is true (rather than needing to also pass ing tabs--type="nav"). Would also need to not apply the role of region on line 16 when IsNav is true so as to not override the nav semantics.

Might make sense to live here so that it could go into React as well.

thatblindgeye:
Closes #7941 

The second bullet in the issue should be tackled as part of the rework of the Compass demos in Core.

### DIFF
--- a/src/patternfly/components/Tabs/examples/Tabs.md
+++ b/src/patternfly/components/Tabs/examples/Tabs.md
@@ -264,12 +264,12 @@ cssPrefix: pf-v6-c-tabs
 
 ### Site navigation variation
 ```hbs isBeta
-{{#> tabs tabs--id="nav-tabs" tabs--IsNav="true"}}
+{{#> tabs tabs--id="nav-tabs" tabs--IsNav=true}}
   {{> __tabs-list }}
 {{/tabs}}
 <br/ >
 As sub navigation:
-{{#> tabs tabs--id="nav-tabs" tabs--IsNav="true" tabs--IsSubtab="true"}}
+{{#> tabs tabs--id="nav-tabs" tabs--IsNav=true tabs--IsSubtab=true}}
   {{> __tabs-list }}
 {{/tabs}}
 ```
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -264,12 +264,12 @@ cssPrefix: pf-v6-c-tabs
 
 ### Site navigation variation
 ```hbs isBeta
-{{#> tabs tabs--id="nav-tabs" tabs--IsNav="true"}}
+{{#> tabs tabs--id="nav-tabs" tabs--IsNav=true}}
```
mcoker:
```suggestion
{{#> tabs tabs--id="nav-tabs" tabs--IsNav=true tabs--attribute='aria-label="Site navigation variation example"'}}
```

src/patternfly/components/Tabs/examples/Tabs.md
```diff
@@ -264,12 +264,12 @@ cssPrefix: pf-v6-c-tabs
...
   {{> __tabs-list }}
 {{/tabs}}
 <br/ >
 As sub navigation:
-{{#> tabs tabs--id="nav-tabs" tabs--IsNav="true" tabs--IsSubtab="true"}}
+{{#> tabs tabs--id="nav-tabs" tabs--IsNav=true tabs--IsSubtab=true}}
```
mcoker:
```suggestion
{{#> tabs tabs--id="nav-tabs" tabs--IsNav=true tabs--IsSubtab=true tabs--attribute='aria-label="Site sub-navigation variation example"'}}
```

mcoker (approved):

PR #7981, opened 2025-11-07
fix(drawer): adjust pill for rtl and widths

### LINKED ISSUE #7947: Drawer - pill variant updates
* Add support for the inset in RTL - currently breaks
* Ensure that the inset doesn't interfere with the panel width modifiers. For example, `pf-m-width-100` currently makes the panel go outside of the container.

Image

srambach:
Fixes #7947 
Uses the pill drawer inset to adjust
- the position of the drawer in RTL (subtract the inset from the translate)
- the width of the drawer when specified 
- for 100% subtract twice the offset so it's offset on both sides

Formula syntax assisted by Claude code

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -252,7 +252,7 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
       @include pf-v6-bidirectional-style(
         $prop: transform,
         $ltr-val: translateX(calc(-100% - var(--#{$drawer}--m-expanded__panel--inset))),
-        $rtl-val: translateX(#{pf-v6-calc-inverse(-100%)}),
+        $rtl-val: translateX(#{pf-v6-calc-inverse(calc(-100% - var(--#{$drawer}--m-expanded__panel--inset)))}),
       );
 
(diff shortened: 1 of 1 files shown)


mcoker (commented):
Looks like the reason the inline pill drawer isn't responding to the `.pf-m-width` modifiers is because this var's value doesn't have a unit and is used in a `calc()` https://github.com/srambach/patternfly/blob/f14de9bde750ad33ab6217ea8683f216d6d224cb/src/patternfly/components/Drawer/drawer.scss#L218

Also looked at the react component and it sets `--pf-v6-c-drawer__panel--md--FlexBasis--min` and `--pf-v6-c-drawer__panel--md--FlexBasis--max` inline via `minSize` and `maxSize` props, so a combo of `minSize: {some ideal size}` and `maxSize: 100%` is a valid use case. IMO that's more work than this issue needs. The percentage based width classes work and that should be good enough.

I'd be good with a follow-up that looks at what happens when you use `--FlexBasis`, `--FlexBasis--min`, and `--FlexBasis--max`, and also tests resizable since I don't think that's been looked at yet, either. wdyt?

srambach:
- I put in the 0px so that the calc doesn't fail on the width calculation
- I removed the conditional to subtract the inset on both sides of the pill drawer
- I had to remove the drawer gap for inline (pill) drawers - otherwise the drawer was misaligned by that much 😫

mcoker (approved):
L🗃️TM

PR #7984, opened 2025-11-10
chore(Compass): replaced flex layout in CompassMainHeader

### LINKED ISSUE #7969: Compass - update CompassMainHeader to support title and toolbar elements
Right now we use a generic flex layout in the CompassMainHeader - let's update that to compass component sub-elements like

```
compass__[main-header-content]?
  compass__main-header-title
  compass__main-header-toolbar (or tools or something, doesn't haven't to be a `<Toolbar>®`)
```

thatblindgeye:
Closes #7969

### DIFF
--- a/src/patternfly/components/Compass/compass.scss
+++ b/src/patternfly/components/Compass/compass.scss
@@ -94,15 +94,13 @@
   min-height: 0;
 }
 
-.#{$compass}__main-header {
-  > .#{$compass}__panel {
-    display: flex;
-    gap: var(--#{$compass}__main-header--RowGap) var(--#{$compass}__main-header--ColumnGap);
+.#{$compass}__main-header-content {
+  display: flex;
+  gap: var(--#{$compass}__main-header--RowGap) var(--#{$compass}__main-header--ColumnGap);
+}
 
-    .#{$compass}__main-header-title {
-      flex-grow: 1;
-    }
-  }
+.#{$compass}__main-header-title {
+  flex-grow: 1;
 }
 
 .#{$compass}__content {
--- a/src/patternfly/components/Compass/compass.scss
+++ b/src/patternfly/components/Compass/compass.scss
@@ -95,19 +95,12 @@
 }
 
 .#{$compass}__main-header {
-  & .#{$compass}__panel {
+  > .#{$compass}__panel {
     display: flex;
-    flex-wrap: wrap;
     gap: var(--#{$compass}__main-header--RowGap) var(--#{$compass}__main-header--ColumnGap);
-    align-items: center;
 
     .#{$compass}__main-header-title {
       flex-grow: 1;
-      max-width: 100%;
-    }
-
-    .#{$compass}__main-header-toolbar {
-      max-width: 100%;
     }
   }
 }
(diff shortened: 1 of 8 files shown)


src/patternfly/components/Compass/compass.scss
```diff
@@ -92,6 +94,24 @@
   min-height: 0;
 }
 
+.#{$compass}__main-header {
+  & .#{$compass}__panel {
```
mcoker: The `&` is redundant when using a nested element like this, though you'd want to target the direct descendent in case for some weird reason there were nested panel elements 

```suggestion
  > .#{$compass}__panel {
```

src/patternfly/components/Compass/compass.scss
```diff
@@ -92,6 +94,24 @@
...
+    display: flex;
+    flex-wrap: wrap;
```
mcoker: I'm not sure if these should wrap. Typically with these types of elements, we don't wrap by default. It's up to the user to manage the toolbar size so it leaves room for the title. For card, we have a `.pf-m-wrap` modifier for the header. I looked at the existing demos and the flex layout wraps, but IMO that's because it's the default for the flex layout, not an intentional choice (not by me anyways).

src/patternfly/components/Compass/compass.scss
```diff
@@ -92,6 +94,24 @@
...
+    gap: var(--#{$compass}__main-header--RowGap) var(--#{$compass}__main-header--ColumnGap);
+    align-items: center;
```
mcoker: If they don't wrap by default, we'd want to take this off and leave things top-aligned or baseline aligned or something like that, so that if the title text wraps, it wraps like the card. And because the actions _should_ be taller than the title, the title would probably need `align-self: center`

Image

src/patternfly/components/Compass/compass.scss
```diff
@@ -92,6 +94,24 @@
...
+
+    .#{$compass}__main-header-title {
+      flex-grow: 1;
+      max-width: 100%;
```
mcoker: Curious why you needed `max-width: 100%`? Sometimes the contents of flex items overflow their container, but it's most often due to `min-width: min-content` as the default for a flex item, and you'd disable that with `min-width: 0;`.

src/patternfly/components/Compass/compass.scss
```diff
@@ -92,6 +94,24 @@
...
+    }
+
+    .#{$compass}__main-header-toolbar {
+      max-width: 100%;
```
mcoker: same comment about `max-width: 100%` as on the block above.

src/patternfly/components/Compass/compass.scss
thatblindgeye: Can remove, this was just one of the properties from the Flex layout components I ported over to this.

src/patternfly/components/Compass/compass.scss
thatblindgeye: Similar to above, was one of the things applied with the Flex layout components being used that I ported to this. Wasn't sure exactly what properties we'd want on these new classes after getting rid of those Flex components

src/patternfly/components/Compass/compass.scss
```diff
@@ -92,6 +94,17 @@
...
 }
 
+.#{$compass}__main-header {
+  > .#{$compass}__panel {
+    display: flex;
+    gap: var(--#{$compass}__main-header--RowGap) var(--#{$compass}__main-header--ColumnGap);
+
+    .#{$compass}__main-header-title {
+      flex-grow: 1;
+    }
```
mcoker: You can take this out of the nesting, it should only be used in this context. The lower the specificity the better in general.

mcoker (approved):
🚀

PR #7985, opened 2025-11-10
fix(compass): restructure navigation to include home and search

### LINKED ISSUE #7930: Tabs/Nav - structure needs to support more than just nav items
The top nav (north on the compass) needs to able to consume components beside just tabs/nav items. It could have icon buttons, for example, like in the following design:

https://www.figma.com/design/lJ5s1z11E4Q2Bp6kD3TE0W/Red-Hat-Genie---MVP?node-id=364-52735&m=dev

This may require some reworking of the initial updated structure of the nav built for compass

srambach:
Fixes #7930 

This changes the nav structure to be 
`__nav
  __panel
    __nav-content
      __nav-home
      __nav-main
      __nav-search`

NOTE: it sets the block padding of the nav element to 0 because tabs push it too tall and then the pill ends don't match the round buttons. Technically I think it's still a pixel or two off...input welcome.

ALSO: I didn't add in a home or search icon but can do that if desired and someone provides the svg code for them.


src/patternfly/components/Compass/compass.scss
```diff
@@ -68,13 +80,20 @@
 }
 
 .#{$compass}__nav {
+  --#{$compass}__panel--Padding: 0 var(--#{$compass}__nav--Padding); // assumes there are tabs pushing the height enough not to need block padding on the home/search buttons
```
mcoker: We should probably separate `__panel--Padding` into the 4 sides. Do you mind doing that in this PR? The only other place you'd need to update the use of `__panel--Padding` is the sidebar padding here - https://github.com/srambach/patternfly/blob/432fb30ea563b5f069002910420b37e57e263d85/src/patternfly/components/Compass/compass.scss#L98.

mcoker (commented):
* Left a comment about separating the panel padding vars
* Also would you mind taking the inset variation off of the tabs and adding a `lg` spacer as left/right padding to `__nav-main`? That's a little less space and it still looks ok.
* And attaching the home and search icons. I grabbed them from figma and edited them without testing to make sure they work. They should just need width/height of 1em, class `.pf-v6-svg` and any fill colors set to `currentcolor`

[icons.zip](https://github.com/user-attachments/files/23468677/icons.zip)

mcoker:
Also looks like you need to rebase. Probably just need to remove `compass__hero` since it's a standalone component now.

mcoker (approved):
Perfecto!

PR #7986, opened 2025-11-10
fix: Updated to move compass into generative ui section.

dlabaj:
Updated to move doc updates for compass into generative ui section.

### DIFF
--- a/patternfly-docs/patternfly-docs.config.js
+++ b/patternfly-docs/patternfly-docs.config.js
@@ -14,7 +14,7 @@ module.exports = {
     { section: 'utility-classes' },
     { section: 'patterns' },
     { section: 'extensions' },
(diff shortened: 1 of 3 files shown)


kmcfaul (approved):

mcoker:
@dlabaj since we don't have a generative UI section, I can't find it in the docs now? Also would we want to move the demos, too? https://github.com/patternfly/patternfly/blob/compass-docs-updates/src/patternfly/demos/Compass/examples/Compass.md

nicolethoen:
This update will likely require an additional section added to the docs.config for PatternFly AI / Generative UI to show up.

And yes, the demos should move, too

mcoker (approved):
LGTM!

PR #7987, opened 2025-11-11
fix(compass): add support for expanded

### LINKED ISSUE #7983: Compass - expanded/collapsed sections
Currently, the compass header+sidebars+footer have a react prop that defaults to true for an "expanded" state. We should support styling for when the expanded modifier is present and absent (collapsed/hidden).

Right now, react isn't conditionally rendering any of these divs, so we can either handle hiding the sections in css when the expanded modifier isn't there, or we can update react to not include the section at all.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7983

* Toggle `.pf-m-expanded` on any of the header/sidebars/footer to have it show/hide
* Fixes sidebar padding (regression from https://github.com/patternfly/patternfly/pull/7985)
* Fixes the border that goes on the compass sections in the component example - broke when we moved the example from "components" to the AI section
* Removes an unnecessary padding that was around the right/end sidebar
* Removes some commented out code
* Updates fullscreen screenshots. Most of them are from the text underline change it looks like?


kmcfaul (approved):

src/patternfly/components/Compass/compass-footer.hbs
```diff
@@ -1,4 +1,10 @@
-<div class="{{pfv}}compass__footer{{#if compass-footer--modifier}} {{compass-footer--modifier}}{{/if}}"
+<div class="{{pfv}}compass__footer
+  {{#unless compass-footer--IsCollapsed}}
+    pf-m-expanded
+  {{/unless}}
```
thatblindgeye, line 4: For these 3 components, do we want to just add `inert` as well when it's collapsed to be doubly sure content can't be accessed?

thatblindgeye (commented):
Question/potential change below. Also do we have it tracked to have a demo showing how one could go about handling the expanding functionality in React, or can that be handled by any generic followup for cleaning up demos/examples?

src/patternfly/components/Compass/compass.scss
```diff
@@ -107,8 +159,20 @@
...
+  }
+
   &.pf-m-start {
     grid-area: sidebar-start;
+    translate: -100%;
+  }
+
+  &.pf-m-end {
+    grid-area: sidebar-end;
+    translate: 100%;
```
srambach: You'll need an RTL value of the inverse so it slides the correct direction

srambach (commented):
Does rearranging to put the message bar into the main content mean that it will need to handle expanding, or if that's desired the user will have to use the footer?

src/patternfly/components/Compass/compass.scss
mcoker: \*\*_shakes fist angrily_\*\* darn you RTL, you have foiled me once again!

mcoker:
@srambach @thatblindgeye 
* fixed RTL slide direction
* updated the grid spacing to use the page-chrome-inset token per the working sesh this morning
* added `inert` when not expanded
* made the `__main-footer` expand/collapse like the other sections

I'll open a separate PR to smooth out the animation per @lboehling's comment this morning

thatblindgeye (approved):

srambach (approved):
L 🪗  TM

PR #7988, opened 2025-11-11
fix(compass): add main-footer

### LINKED ISSUE #7938: Compass footer - should expand over main page content as it grows
Currently he MessageBar input expands upward, pushing the main content and sidebars. This causes a significant layout shift and a poor user experience. Additionally, the left and right sidebars are currently affected by the input's height change.

Recommendation: The input area must not alter the main page layout. It should be a fixed or sticky/floating element. If it expands, it should expand over the content or within a pre-allocated, fixed space. The sidebars must remain fixed and independent of any state changes in the chat input.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7938

@lboehling or @andrew-ronaldson for the chatbot message bar, should the space between it and the section above (the row gap) be tighter like this where it matches the space/row-gap between sections in the main area?

Screenshot 2025-11-10 at 9 49 57 PM

Or should it have more space, which matches the gap that's used for the main layout
* between the header and content below it
* between the left/right sidebars and the content in the middle

Screenshot 2025-11-10 at 9 49 40 PM

### DIFF
--- a/src/patternfly/components/Compass/compass-main-footer.hbs
+++ b/src/patternfly/components/Compass/compass-main-footer.hbs
@@ -1,0 +1,6 @@
+<div class="{{pfv}}compass__main-footer{{#if compass-main-footer--modifier}} {{compass-main-footer--modifier}}{{/if}}"
+  {{#if compass-main-footer--attribute}}
+    {{{compass-main-footer--attribute}}}
+  {{/if}}>
+  {{> @partial-block}}
+</div>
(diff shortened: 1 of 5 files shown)


srambach (approved):
Looks good. Aside - might be nice to use the example CSS to shorten the example height but that's separate.

lboehling:
@mcoker I think it should be larger / match the gap of the rest of the main layout but i'll defer to @andrew-ronaldson when he's back!

andrew-ronaldson:
Looks like the spacing around the left/right actions is gone.
side-actions

mcoker:
@andrew-ronaldson yeah that happened in another PR but should be fixed now 👍

PR #7994, opened 2025-11-13
feat(compass): add smooth animation opt-in

### LINKED ISSUE #7993: Compass - animate height/width when expanding panels
The current approach is performant because the only properties that are transitioning are `translate` and `opacity`. However, design mentioned that the current animations feel clunky so we want to try smoothing it out by animating width/height changes so that everything shrinks/grows instead of popping into place.

This approach may come with performance issues, so this is probably best as an opt-in behind a class. Then we can test it and see if it's safe to make as the default behavior.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7993

What it do:
* Adds an opt-in `.pf-m-animate-smooth` on the compass wrapper to opt-in to the non-performant, but cooler looking animation that animates the content width/height as the panels shift in and out of place. Smooth is probably the wrong name, but `.pf-m-smooth-ish-but-probably-choppy-also` was already taken 😅
* In the previous/existing animation, we collapsed the panels, but the layout's `gap` remained - so there was too much space around the main content when panels were collapsed. Oops! That's fixed now, thank goodness, before @thatblindgeye noticed and yelled at me.
* Since the panels widths/heights now transition at the same time as `translate` is happening, self-relevant values for `translate` (like "100%") don't really work. There are now vars for how far sections should move in/out of the way. I just plugged in some numbers for now - we may want to tweak these with design.
  * `--#{$compass}--section--slide--length--header: 10rem;`
  * `--#{$compass}--section--slide--length--sidebar: var(--pf-t--global--spacer--4xl);`
  * `--#{$compass}--section--slide--length--main-footer: 6rem;`
  * `--#{$compass}--section--slide--length--footer: 6rem;`

### DIFF
--- a/src/patternfly/components/Compass/compass.hbs
+++ b/src/patternfly/components/Compass/compass.hbs
@@ -1,6 +1,6 @@
 <div class="{{pfv}}compass
   {{#unless compass--NoSmoothAnimate}}
-    pf-m-animate-smooth
+    pf-m-animate-smoothly
   {{/unless}}
   {{setModifiers
     compass--IsStart='pf-m-start'
--- a/src/patternfly/components/Compass/compass.hbs
+++ b/src/patternfly/components/Compass/compass.hbs
@@ -1,4 +1,12 @@
-<div class="{{pfv}}compass{{#if compass--modifier}} {{compass--modifier}}{{/if}}"
+<div class="{{pfv}}compass
+  {{#unless compass--NoSmoothAnimate}}
+    pf-m-animate-smooth
+  {{/unless}}
+  {{setModifiers
+    compass--IsStart='pf-m-start'
+    compass--IsEnd='pf-m-end'
+    compass--modifier=compass--modifier
+  }}"
   {{#if compass--attribute}}
     {{{compass--attribute}}}
   {{/if}}>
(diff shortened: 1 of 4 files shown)


kmcfaul:
Do we want a react followup adding a flag to `Compass`, or should we leave this as an optional manual css class to pass directly via class/className?

mcoker:
@andrew-ronaldson or @lboehling wdyt? Should we add a react prop to the compass component that lets you easily toggle whether the animation is smooth or more blocky (like we demo'd yesterday), or do we not want to make it that public yet and have it be a class people can manually add if they want that? If we go the latter route, we could include the class in our upcoming compass handbook, and mention that it may be a little jittery depending on what else the page is doing.

thatblindgeye (approved):
> Oops! That's fixed now, thank goodness, before @thatblindgeye noticed and yelled at me.

I can yell at you for taking this away from me.


This lgtm, though

srambach (approved):
I think my only comment is a suggestion to change the classname from `.pf-m-animate-smooth` to `.pf-m-animate-smoothly` or `.pf-m-smooth-animation` 🤓

mcoker:
Since the animation modifiers are `.pf-m-animate-[expand/current]` I'd like to keep the prefix that way - updated to `.pf-m-animate-smoothly`

srambach (approved):
The grammar police and I thank you!

kmcfaul (approved):

mcoker:
@lboehling @andrew-ronaldson made the panels collapse faster. Attaching a couple of videos. Look ok? Do you want to release both (default is performant, non-performant is an opt-in) or think we should just pick one and ditch the other?

https://github.com/user-attachments/assets/4cd79cec-56b7-4044-b7d7-3d600e79a071


https://github.com/user-attachments/assets/03aeef32-919d-48ae-abc9-1b36a3270983

lboehling (approved):
looks great! thank u!

PR #8008, opened 2025-12-01
fix: fixes broken links

edonehoo:
Core counterpart to https://github.com/patternfly/patternfly-org/pull/4882 to update a few broken links resulting from the nav redesign

### DIFF
--- a/src/patternfly/utilities/Text/examples/Text.md
+++ b/src/patternfly/utilities/Text/examples/Text.md
@@ -214,7 +214,7 @@ These colors are best used with the ["inverse" background colors](/foundations-a
 
 These text utilities can be used to modify text within the PatternFly framework. In most cases, using the components with available modifiers should be sufficient to implement most designs, and should be preferred over customizations using these utilities.
 
-Care should be taken especially when applying text colors, as this can have a negative effect on the readability and accessibility of text. Refer to the information on [contrast ratios](/guidelines/colors/#contrast-ratios) for more information.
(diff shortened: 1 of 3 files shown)


src/patternfly/utilities/Text/examples/Text.md
```diff
@@ -214,7 +214,7 @@ These colors are best used with the ["inverse" background colors](/foundations-a
 
 These text utilities can be used to modify text within the PatternFly framework. In most cases, using the components with available modifiers should be sufficient to implement most designs, and should be preferred over customizations using these utilities.
 
-Care should be taken especially when applying text colors, as this can have a negative effect on the readability and accessibility of text. Refer to the information on [contrast ratios](/guidelines/colors/#contrast-ratios) for more information.
+Care should be taken especially when applying text colors, as this can have a negative effect on the readability and accessibility of text. Refer to the information on [contrast ratios](/foundations-and-styles/colors/#contrast-ratios) for more information.
```
mcoker, line 217: I'm not sure if it matters because the link you provided works, but it's in a slightly different format that I'd expect. FWIW if I go to staging (nothing else in the url), then just navigate to the colors section, both the link in the jump links for "contrast ratios" and the little link icon next to the "contrast ratios" section title give me the format below, without the `/` before the `#`.

```suggestion
Care should be taken especially when applying text colors, as this can have a negative effect on the readability and accessibility of text. Refer to the information on [contrast ratios](/foundations-and-styles/colors#contrast-ratios) for more information.
```

mcoker (approved):
Nice, thanks!

src/patternfly/utilities/Text/examples/Text.md
mcoker: I'll submit another PR to update this and fix a few more broken links I found that are not related to any of the nav updates.

PR #8020, opened 2025-12-04
feat: add docked nav

### LINKED ISSUE #7183: Navigation: Core support to enable docked nav variant
Follow up to https://github.com/patternfly/patternfly-design/issues/1243

The goal is to add support for a small docked version of our navigation with an icon view. This is a variant of the navigation component not a replacement.

Also related to https://github.com/patternfly/patternfly-design/issues/1350

Figma link https://www.figma.com/design/VMEX8Xg2nzhBX8rfBx53jp/branch/raktQs3QsxK8IDzPhKThmW/PatternFly-6%3A-Components?m=auto&node-id=3-35&t=7JGjC1c3kBnDTw1w-1

A flyout menu (drawer) option would be beneficial if there the docked nav item has children.

mcoker:
fixes https://github.com/patternfly/patternfly/issues/7183

Links:
* https://pf-pr-8020.surge.sh/components/navigation/html-demos/docked-nav/ - docked nav
* https://pf-pr-8020.surge.sh/components/navigation/html-demos/default-nav/ - regular nav (shows the accent)
* https://pf-pr-8020.surge.sh/ai/generative-uis/compass/html-demos/docked/ - docked compass. Add `.pf-v6-theme-glass` to the `<html>` element to enable glass

Adds these things:
* `.pf-m-docked` to the compass component for the docked + big content layout
* `.pf-v6-c-compass__dock`, `.pf-v6-c-compass__dock-logo`, `.pf-v6-c-compass__dock-main`, and `.pf-v6-c-compass__dock-tools` for the dock contents.
* `.pf-m-docked` to the masthead component that basically just turns it into a vertical layout, and changes it to a flex layout so we can easily add dividers between the sections. For use with regular pf layouts.
* `.pf-m-docked` to the nav component that hides the nav link text and turns it into a vertical layout
  * Text is shown below a nav item if text is passed to it. Otherwise it's just an icon.
* The accent styling to current nav items.
* `.pf-m-docked-nav` to the page component that modifies the grid for the docked layout
* `.pf-m-vertical` to the toolbar component for a vertical layout

TODO:
* add the nav slide out panel

### DIFF
--- a/src/patternfly/components/Compass/compass.scss
+++ b/src/patternfly/components/Compass/compass.scss
@@ -94,7 +94,7 @@
   background-image: var(--#{$compass}--BackgroundImage);
   background-size: cover;
 
-  &.pf-m-docked {
+  &.pf-m-dock {
     grid-template-areas: "dock main";
     grid-template-rows: auto;
     grid-template-columns: auto 1fr;
--- a/src/patternfly/components/Compass/compass.scss
+++ b/src/patternfly/components/Compass/compass.scss
@@ -86,6 +86,29 @@
   background-image: var(--#{$compass}--BackgroundImage);
   background-size: cover;
 
+  &.pf-m-max-canvas {
+    grid-template-areas: "header main";
+    grid-template-rows: auto;
+    grid-template-columns: auto 1fr;
+    align-items: stretch;
+    padding: 0;
+
+    .#{$compass}__main {
+      padding: var(--#{$compass}--Padding);
+    }
+
+    .#{$compass}__header {
+      display: block;
+      margin-block-end: 0;
+
+      > .#{$masthead} {
+        background-color: var(--#{$compass}__panel--BackgroundColor);
+        backdrop-filter: var(--#{$compass}__panel--BackdropFilter);
+        border-color: var(--#{$compass}__panel--BorderColor);
+      }
+    }
+  }
+
   :root:where(.#{$pf-prefix}theme-dark) & {
     --#{$compass}--BackgroundImage: var(--#{$compass}--BackgroundImage--dark);
   }
--- a/src/patternfly/components/Compass/compass.scss
+++ b/src/patternfly/components/Compass/compass.scss
@@ -200,8 +200,8 @@
 
 .#{$compass}__dock {
   display: grid;
-  grid-template-rows: auto 1fr auto;
   grid-area: dock;
+  grid-template-rows: auto 1fr auto;
   row-gap: var(--#{$compass}__dock--RowGap);
   padding-block-start: var(--#{$compass}__dock--PaddingBlockStart);
   padding-block-end: var(--#{$compass}__dock--PaddingBlockEnd);
--- a/src/patternfly/components/Compass/compass.scss
+++ b/src/patternfly/components/Compass/compass.scss
@@ -6,6 +6,14 @@
   --#{$compass}--BackgroundImage--dark: none;
   --#{$compass}--Padding: var(--pf-t--global--spacer--inset--page-chrome);
   --#{$compass}--Gap: var(--pf-t--global--spacer--inset--page-chrome);
+  --#{$compass}__dock--RowGap: var(--pf-t--global--spacer--gap--group-to-group--vertical--default);
+  --#{$compass}__dock--BackgroundColor: var(--pf-t--global--background--color--glass--default);
+  --#{$compass}__dock--BackdropFilter: var(--pf-t--global--background--color--glass--filter);
+  --#{$compass}__dock--BorderColor: var(--pf-t--global--border--color--alt);
+  --#{$compass}__dock--PaddingBlockStart: var(--pf-t--global--spacer--inset--page-chrome);
+  --#{$compass}__dock--PaddingBlockEnd: var(--pf-t--global--spacer--inset--page-chrome);
+  --#{$compass}__dock--PaddingInlineStart: var(--pf-t--global--spacer--sm);
+  --#{$compass}__dock--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$compass}__nav--RowGap: var(--pf-t--global--spacer--gap--group--vertical);
   --#{$compass}__nav--PaddingBlockStart: 0;
   --#{$compass}__nav--PaddingBlockEnd: 0;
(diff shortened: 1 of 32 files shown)


src/patternfly/components/Compass/compass.scss
```diff
@@ -86,6 +86,29 @@
   background-image: var(--#{$compass}--BackgroundImage);
   background-size: cover;
 
+  &.pf-m-max-canvas {
```
srambach: Is "max" a new concept in our modifiers? Can we reuse something like "fill" or "full" (though those usually refer to itself, not its children)? Also, "canvas" isn't used anywhere else - the "main" grid area is what it's maximizing?

kmcfaul:
Do we want to call the new `pf-m-docked` modifiers for masthead & nav something more generic like `pf-m-vertical`?

srambach:
I'm assuming this isn't hardened yet - there are some potential problems at narrow widths
image
image

srambach:
I don't love having two things that are "header." 
image
I still think layout slots are getting muddled with content items that can go in them. E.g. maybe what's shown here is something like a dock with a modifier of .pf-m-inline-start and in there can be a masthead(`<header>`), nav, toolbar, whatever inside it. 
Or, the existing compass layout areas - sidebars, header, and footer - have a docked variant.
But I'd lean more toward separating concerns - layout just handles the box to put things in and if it slides off screen or whatever, and the things inside manage spacing themselves.

src/patternfly/components/Masthead/masthead.scss
```diff
@@ -98,6 +109,33 @@ $pf-v6-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
 
 // - Masthead
 .#{$masthead} {
+  &.pf-m-docked {
```
srambach: Is this supposed to be a real part of masthead? If so, there should be an example and docs in the .md file.

Second thing here, I really don't like .pf-m-docked as the name changing the orientation

src/patternfly/components/Page/page.scss
```diff
@@ -190,6 +195,18 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
     grid-template-columns: var(--#{$page}__sidebar--Width) 1fr;
   }
 
+  &.pf-m-docked {
```
srambach: Since it's not the page that is docked, it seems like this should be named something like  `.pf-m-has-dock` (it has something docked, not is docked) or `pf-m-side-masthead`/`.pf-m-vertical-masthead`/`.pf-whatever-layout` (because it's a different page layout option).

I think I'd rather see something about the layout rather than has-dock because I don't think dock necessarily implies a left side vertical dock.

src/patternfly/components/Compass/compass.scss
mcoker: That was just something temporary - updated to `.pf-m-docked`

src/patternfly/components/Masthead/masthead.scss
mcoker: Added docs and an example.

On the name, `pf-m-vertical` would be fine, but my thinking is this for the docked nav specifically and this lets us be opinionated for use as a dock that wouldn't necessarily apply to a vertical masthead with regular stuff int (the children being center aligned, for example). But it doesn't matter to me, let me know if you want to change it.

src/patternfly/components/Page/page.scss
mcoker: Updated to `.pf-m-dock` to indicate it has a dock, not that the page is docked.

src/patternfly/components/Nav/examples/Navigation.md
```diff
@@ -559,6 +559,26 @@ import './Navigation.css'
 {{/nav}}
 ```
 
+### Docked nav
```
srambach, line 562: FWIW, this doesn't show the selection decorator in the full page view since it's a negative margin.

src/patternfly/demos/Compass/examples/Compass.md
```diff
@@ -372,4 +372,114 @@ wrapperTag: div
     {{> drawer-example-panel drawer-panel-example--IsGlass=true}}
   {{/drawer-main}}
 {{/drawer}}
-```
\ No newline at end of file
+```
+
+### Docked
```
srambach, line 377: On narrow screens, the "make your screen wider" box isn't centered anymore if you care.
Also, if the screen is too short to fit all the nav items, there's no scroll.

src/patternfly/components/Masthead/masthead.scss
srambach: 🤷🏻‍♀️ Could also be 2 modifiers - vertical and docked. But I'm not going to fight you on this 😄

src/patternfly/components/Nav/examples/Navigation.md
mcoker: Yah... I wasn't sure about that. We could add a margin the example CSS. Probably good idea since our visual regression tests only test the full-page examples. WDYT?

src/patternfly/demos/Compass/examples/Compass.md
mcoker: Ah, I'll fix the responsive box if it isn't too complicated. And on the scroll, this is just being used for POC/vibe stuff so we can handle that part later.

mcoker:
> Could also be 2 modifiers - vertical and docked. But I'm not going to fight you on this

🥊🤖

We could, but IMO it isn't necessary. If we end up switching orientation to horizontal but want to keep all the other alignment and stuff that comes with docked, that could be called horizontal or top/bottom or whatever, and the default is vertical. Like most of our components - if we have a variant that just changes orientation (or some other style), the variant for the change is what gets the class (action-list & action-list.pf-m-vertical, nav & nav.pf-m-horizontal, tabs & tabs.pf-m-vertical, etc).

srambach:
Taaaalllll button! (https://pf-pr-8020.surge.sh/components/masthead/html/docked/)
https://github.com/user-attachments/assets/11ff4f28-8d24-4fc5-bc62-9e92cdcb8be9

srambach:
If there are too many items in the dock to fit, nothing scrolls to get to them. You can tab to them but not get there with the mouse.
https://github.com/user-attachments/assets/ffc07207-edd2-4ab3-a961-82607209783f

kmcfaul (approved):

src/patternfly/components/Compass/examples/Compass.md
```diff
@@ -49,6 +49,39 @@ import './Compass.css';
 {{/compass}}
 ```
 
+### Docked
+```hbs isBeta
+{{#> compass compass--HasDock=true}}
+  {{#> compass-dock}}
+    {{#> compass-dock-content}}
```
kmcfaul: Does this div exist? I don't see it in the surge or css.

mcoker: Ah that's a remnant. What it did was make everything flexible so that compass-dock was the compass layout container for the dock, and compass-dock-content effectively the dock component, so you could wrap the dock in a pill/panel or just put whatever you want in there. I removed it after we decided we didn't want users to deviate from the PF docked design.

But from the convos from the past couple of days, it sounds like we want this to be flexible - we've talked about possibly wanting this to be easily swappable with masthead, and there's some world where we may still entertain the pill panel. IMO this is better, and I'd like to make the same change to the page component.

Current compass
```
compass
    compass__dock
        // compass children
```

Current page (mimics non-docked)
```
page
    masthead
        // masthead children
```

Proposed compass
```
compass
    compass__dock
        compass__dock-content
            // dock children
```

Proposed page - IMO this should exist in the default page layout, too but with `page__header` as the layout container. We've probably talked about this before.
```
page
    page__dock
        masthead
            // masthead children
```

compass__dock (container) -> dock-content (opinionated layout) is how we've set up the other compass components with a similar structure where the container can take whatever but the container-content is our default/opnionated layout. like `compass__nav > compass__nav-content` and `main-header -> main-header-content`

andrew-ronaldson (approved):
Go for it!

srambach:
Just one more check - on https://pf-pr-8020.surge.sh/components/page/html/with-dock/ is this correct or not to be worried about - logo shifts over, and the main section touches the right/bottom side of the viewport but not top/left. 

Otherwise 👍🏻 

https://github.com/user-attachments/assets/bb81cb07-c102-44e1-9ebb-56fe35dd3c45

mcoker:
@srambach created a follow-up issue for that https://github.com/patternfly/patternfly/issues/8026. Right now the POCs built from this will be on wide screens with custom content so it's OK for the time being.

srambach (approved):
L 🛳️ TM

PR #8025, opened 2025-12-12
feat(toolbar): implement container queries

### LINKED ISSUE #7967: Toolbar - container queries
Based on experimentation with a container query on toolbar in #7888 clean up and harden a container query for toolbar

---

**Jira Issue:** [PF-2914](https://redhat.atlassian.net/browse/PF-2914)

srambach:
This adds container breakpoints and creates another set of breakpoint modifiers based on a toolbar container for each of the existing (media based) breakpoint modifiers.

Fixes #7967 

Implemented with support from Claude and Cursor.

## Summary by CodeRabbit

* **New Features**
  * Toolbar now supports opt-in container-query responsiveness, enabling container-aware show/hide behavior and improved toggle-group breakpoint handling.
* **Documentation**
  * Added detailed guidance and examples for container-query Toolbars, including breakpoint behavior, media- vs container-query comparisons, and troubleshooting notes.
* **Style**
  * Introduced supporting container-query utilities and breakpoint variables, plus example styling hooks demonstrating resizeable/spacer container interactions.
* **Chores**
  * Updated local ignore rules to ignore the `.claude` directory.


mcoker (commented):
Looks great! As we discussed offline:
* Increasing the specificity of the existing hidden/visible/toggle-group rules to add `.pf-v6-c-toolbar:not({container-class})` could have some unwanted side effects, so it would be great if we can leave the specificity as it is - it would be fine to increase under `.pf-m-container` if needed since that's an opt-in
* Make sure any examples that show container queries have `isBeta` in the start of the hbs block
* Run visual regressions just to make sure nothing else changed

I think it would probably be a good idea to run them against the react repo, too, since I think toggle group examples have markup changes based on breakpoints that the core examples won't show. If you want to take a stab, here's how I would do it:
* Start the react dev server with its core package pointing to your core directory with the toolbar cq branch checked out
* In [backstop.config.js](https://github.com/patternfly/patternfly/blob/2854906afecb625feea448307e1b91495d62b570/backstop.config.js#L1) comment out the first line for `fullscreenRoutes`.
* Add a new `const fullscreenRoutes: [...]` - the code for that is below
* Change `baseUrl` to the react server port (`localhost:8002`)
* Run that and if everything looks like it rendered OK (not checking for toolbar bugs), accept those screenshots. 
* Update the react repo to not point to your local core (just run a regular react dev server), and run the screenshot job again. That should give you a diff of the react toolbar examples with and without your branch changes.

That should be it? Here's the `fullscreenRoutes` array reference above

```
const fullscreenRoutes = [
  '/components/toolbar/react/toolbar-items',
  '/components/toolbar/react/with-adjusted-inset',
  '/components/toolbar/react/sticky-toolbar',
  '/components/toolbar/react/with-groups-of-items',
  '/components/toolbar/react/background-color-variants',
  '/components/toolbar/react/component-managed-toggle-groups',
  '/components/toolbar/react/consumer-managed-toggle-groups',
  '/components/toolbar/react/with-filters',
  '/components/toolbar/react/with-custom-label-group-content',
  '/components/toolbar/react/stacked-example',
  '/components/toolbar/react/toolbar-content-wrapping',
  '/components/toolbar/react/toolbar-group-spacers',
  '/components/toolbar/react/toolbar-item-spacers',
  '/components/toolbar/react-demos/console-log-viewer-toolbar-demo',
  '/components/data-list/react-demos/expandable-control-in-toolbar',
  '/components/table/react-demos/sticky-columns-and-header-with-toolbar'
];
```

srambach:
Rebased and used `:where()` in the selectors to reduce specificity.

[Regression tests run against a new baseline](https://drive.google.com/file/d/1hdxNPA4n8VyumoK3KUVcFWaW63K0ur89/view?usp=sharing) as created in https://github.com/patternfly/patternfly/pull/8039 look ok. Toolbar just shows new examples; others seem likely to be rendering. Let's recheck after the updated tests are merged. One to check is /components/masthead/html-demos/with-toggle-group-filters-expandable-content-expanded-mobile/

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -557,6 +557,252 @@ Several components in the following examples do not include functional and/or ac
...
+      {{> toolbar-group-action-inline}}
+    {{/toolbar-expandable-content}}
+  {{/toolbar-content}}
+{{/toolbar}}
+```
+
+This toolbar will expand at the `md` container breakpoint (478px) instead of the default `lg` (702px).
+
+### Comparing container breakpoints
+This example shows how different breakpoint modifiers affect when the toolbar expands.
```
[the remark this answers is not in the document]
srambach: Hey @coderabbitai, we generally haven't put that space in. I want to stay consistent with what we have.

srambach:
Ran React visual diffs as recommended - just one artifact difference.
https://drive.google.com/file/d/1Tne-uELvP3j1Dj0BzUmE3T_7o47IlNnc/view?usp=sharing

Note for future reference - in addition to the instructions above, I had to change one additional line in backstop.config.js
change: const relativeUrls = Object.keys(fullscreenRoutes);
to: const relativeUrls = fullscreenRoutes;

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -435,31 +522,38 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
+            .#{$toolbar}__toggle {
+              display: none;
+            }
+          }
         }
       }
     }
   }
 }
+// stylelint-enable max-nesting-depth
```
jcmill:
```suggestion
// stylelint-enable max-nesting-depth

```

src/patternfly/components/Toolbar/examples/Toolbar.css
```diff
@@ -79,3 +79,12 @@
...
 }
+
+.ws-core-resizeable-container {
+  resize: horizontal;
+  overflow: auto;
+  border: var(--pf-t--global--border--width--extra-strong) dashed var(--pf-t--global--border--color--default);
+  width: 800px;
+  min-width: 300px;
+  max-width: 100%;
+}
```
jcmill, line 91:
```suggestion
.ws-core-resizeable-container {
  width: 800px;
  min-width: 300px;
  max-width: 100%;
  padding: var(--pf-t--global--spacer--sm);
  overflow: auto;
  resize: horizontal;
  border: var(--pf-t--global--border--width--extra-strong) dashed var(--pf-t--global--border--color--default);
}
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -367,11 +395,40 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
+    @at-root .#{$toolbar}:where(:not([class*="pf-m-container"])) & {
+      position: static;
+      padding: 0;
+      border-block-end: 0;
+      box-shadow: none;
+    }
+  }
+
+  // Default container behavior (lg breakpoint) when using .pf-m-container
+  @at-root .#{$toolbar}:where(.pf-m-container:not([class*="pf-m-container-"])) & {
```
mcoker: If it works as-is, I think it's fine to leave it, but from looking at it I can't see a case for having both `.pf-m-container` and a `.pf-m-container-[size]` class on a toolbar, so I don't think we need the `:not()`

```suggestion
  @at-root .#{$toolbar}:where(.pf-m-container) & {
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -490,31 +577,38 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
-        }
-
-        .#{$toolbar}__toggle {
-          display: none;
-        }
-      }
+    // - Toolbar toggle group (only for non-container toolbars)
+    @at-root .#{$toolbar}:where(:not([class*="pf-m-container"])) {
+      @include pf-v6-toolbar-toggle-group-breakpoint-behavior($breakpoint-name);
+    }
```
mcoker: I'm not sure why this was moved to a mixin? It looks like it generates the same CSS as it did, just wraps it in `.#{$toolbar}:where(:not([class*="pf-m-container"])) {}`. If it is, could have been a mixin in some previous iteration of the PR and is no longer necessary?

If it works without the mixin, I'd vote to remove the mixin and just include the code like it was but wrapped in `.#{$toolbar}:where(:not([class*="pf-m-container"])) {}`. Just a little more difficult to read for it to be a mixin IMO.

.gitignore
```diff
@@ -30,4 +30,4 @@ backstop_data/html_report*
 backstop_data/reports
 .history
 chrome
-.claude/*.local.*
+.claude
```
mcoker: As far as I'm aware, we only need to ignore `.local.` since that will be our local settings for the repo. We want to allow other stuff in `.claude` so we can have repo-wide settings.

```suggestion
.claude/*.local.*
```

jcmill: Ive got this change done locally.

jcmill (approved):

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -559,6 +559,248 @@ Several components in the following examples do not include functional and/or ac
...
+            {{/toolbar-content-section}}
+          {{/toolbar-content}}
+        {{/toolbar}}
+      </div>
+    </div>
+  </div>
+</div>
+```
+
+### Toolbar with container query support for hidden/visible items and groups
```
mcoker: This example looks to only be setting `toolbar--modifier="pf-m-container-md"`, but the title is worded like it's going to show `.pf-m-[hidden/visible]-on`.

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -792,6 +1034,12 @@ As the toolbar component is a hybrid layout and component, some of its elements
 | `.pf-m-no-padding` | `.pf-v6-c-toolbar` | Modifies toolbar to have no padding. |
 | `.pf-m-no-background` | `.pf-v6-c-toolbar` | Modifies toolbar to have no background color. |
 | `.pf-m-vertical` | `.pf-v6-c-toolbar` | Modifies toolbar for a vertical layout. |
+| `.pf-m-container` | `.pf-v6-c-toolbar` | Creates a component-level container that responds to the toolbar's width at the `lg` breakpoint. |
```
mcoker: Let's add the breakpoint value (`px` since that's how it's referenced everywhere else) to these `.pf-m-container-*` class docs.

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -251,6 +252,33 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
+      .#{$toolbar}__toggle {
+        display: inline-flex;
+      }
+    }
+
+    // Override visibility modifiers to use container queries instead of media queries
+    .#{$toolbar}__content-section,
+    .#{$toolbar}__group,
+    .#{$toolbar}__item {
+      @include pf-v6-hidden-visible(flex, $container-name: 'pf-v6-contain-toolbar', $use-viewport-bp: true);
```
mcoker: Looks like this is what's causing it to use the media query breakpoints instead of container query breakpoints for hidden-visible

```suggestion
      @include pf-v6-hidden-visible(flex, $container-name: 'pf-v6-contain-toolbar');
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -367,12 +395,41 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
+      position: static;
+      padding: 0;
+      border-block-end: 0;
+      box-shadow: none;
+    }
   }
 
+  // Default container behavior (lg breakpoint) when using .pf-m-container
+  @at-root .#{$toolbar}:where(.pf-m-container) & {
+    @include pf-v6-container-query("lg", "pf-v6-contain-toolbar", true) {
```
mcoker: This is making the component use media query bp's instead of container query bp's

```suggestion
    @include pf-v6-container-query("lg", "pf-v6-contain-toolbar") {
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -367,12 +395,41 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
+      box-shadow: none;
+    }
+  }
+
+  // Container query behavior with explicit breakpoint modifiers
+  // stylelint-disable max-nesting-depth
+  @each $breakpoint, $breakpoint-value in $pf-v6-c-toolbar--breakpoint-map {
+    @if $breakpoint != "base" {
+      @at-root .#{$toolbar}:where(.pf-m-container-#{$breakpoint}) & {
+        @include pf-v6-container-query($breakpoint, "pf-v6-contain-toolbar", true) {
```
mcoker: Making this block use media query bp's instead of container query bp's

```suggestion
        @include pf-v6-container-query($breakpoint, "pf-v6-contain-toolbar") {
```

src/patternfly/components/Toolbar/toolbar.scss
```diff
@@ -491,30 +549,61 @@ $pf-v6--include-toolbar-breakpoints: true !default;
...
+// Container query responsive behavior
+// stylelint-disable max-nesting-depth
+.#{$toolbar}__group:where(.pf-m-toggle-group) {
+  @each $breakpoint, $breakpoint-value in $pf-v6-c-toolbar--breakpoint-map {
+    @if $breakpoint != "base" {
+      $breakpoint-name: -on-#{$breakpoint};
+
+      @at-root .#{$toolbar}:where(.pf-m-container) &,
+      .#{$toolbar}:where(.pf-m-container-#{$breakpoint}) & {
+        @include pf-v6-container-query($breakpoint, "pf-v6-contain-toolbar", true) {
```
mcoker: one last spot!

```suggestion
        @include pf-v6-container-query($breakpoint, "pf-v6-contain-toolbar") {
```

mcoker (commented):
I think the documentation needs some work. I'll work on that a little and share it tomorrow.

mcoker (commented):
Some notes from our meeting on Friday:

* We are not including support for `.pf-m-hide[on-{bp}]` for the time being since `.pf-m-hide` isn't supported in the react component. We can add it if requested.
* We should add `@container` support for hidden/visible on the `__content` element.
* We can remove the `.pf-m-container-*` classes and just keep `.pf-m-container`. That implies we'll also:
  * Update all the `[class*="pf-m-container"]` to `.pf-m-container`
  * Make the expandable content element position absolute (overlay) below lg `@container` bp, and static (inline) above the lg bp.
* We are not going to support flex alignment and what-not in `@container` for the time being. It's an extra 60k of styles. We can add support for that if needed.
* We will update the expandable content element

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -336,7 +336,7 @@ Several components in the following examples do not include functional and/or ac
 
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-m-[show/hide][-on-[sm/md/lg/xl/2xl]]` | `.pf-v6-c-toolbar__group`, `.pf-v6-c-toolbar__item`, `.pf-v6-c-toolbar__expandable-content` | Modifies toolbar element visibility. |
+| `.pf-m-[show/hide][-on-[sm/md/lg/xl/2xl]]` | `.pf-v6-c-toolbar__group`, `.pf-v6-c-toolbar__item`, `.pf-v6-c-toolbar__expandable-content` | Modifies toolbar element visibility (hide is media-query only). |
```
mcoker: This blurb makes sense to me, but I don't know if it makes sense on its own without more context, and I'm not sure if it's necessary. Otherwise we might want to call out the flex/align styles, too.

IMO we could just cover what `.pf-m-container` does and doesn't do, and let that section be the docs for the classes.

wdyt?

```suggestion
| `.pf-m-[show/hide][-on-[sm/md/lg/xl/2xl]]` | `.pf-v6-c-toolbar__group`, `.pf-v6-c-toolbar__item`, `.pf-v6-c-toolbar__expandable-content` | Modifies toolbar element visibility. |
```

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -767,6 +767,229 @@ Visibility can be set per breakpoint to show or hide items and groups based on v
...
+        {{> toolbar-toggle-group}}
+      {{/toolbar-expandable-content}}
+    {{/toolbar-content}}
+  {{/toolbar}}
+</div>
+```
+
+### Expanded content display
+
+Expanded content with `.pf-m-container` switches from an overlay to an inline layout at the `lg (702px)` container breakpoint.
```
mcoker: Just to be consistent with the text above 

```suggestion
Expanded content with `.pf-m-container` switches from an overlay to an inline layout at the `lg` (702px) container breakpoint.
```

src/patternfly/components/Toolbar/examples/Toolbar.md
```diff
@@ -824,7 +1048,7 @@ As the toolbar component is a hybrid layout and component, some of its elements
 | Class | Applied to | Outcome |
 | -- | -- | -- |
 | `.pf-m-toggle-group` | `.pf-v6-c-toolbar__group` | Modifies toolbar group to control when, and at which breakpoint, filters will be hidden/shown. By default, all filters are hidden until the specified breakpoint is reached. |
-| `.pf-m-[show/hide]` | `.pf-v6-c-toolbar__group.pf-m-toggle-group`, `.pf-v6-c-toolbar__expandable-content` | Modifies toolbar element to hidden. |
+| `.pf-m-[show/hide]` | `.pf-v6-c-toolbar__group.pf-m-toggle-group`, `.pf-v6-c-toolbar__expandable-content` | Modifies toolbar element to hidden (hide is media-query only). |
```
mcoker: Same comment here to possibly drop the "hide is media-query only" text.


```suggestion
| `.pf-m-[show/hide]` | `.pf-v6-c-toolbar__group.pf-m-toggle-group`, `.pf-v6-c-toolbar__expandable-content` | Modifies toolbar element to hidden. |
```

PR #8030, opened 2025-12-16
feat(treeview): allow a list item to look disabled

### LINKED ISSUE #7996: TreeView - disabled styling for TreeViewListItem elements
styling for https://github.com/patternfly/patternfly-react/issues/11893

Since we're using TreeView styled native button elements, we'll need a modifier to style the full node button / node text content button, and toggle icon buttons as disabled.

srambach:
Fixes #7996 

Adds .pf-m-disabled to the tree view list item to change the appearance.
Note: Actions should be disabled separately. 
Also, the compact version with a background color has further reduced contrast of disabled items, so it may not be desirable to use these options together.

Note: I added some disabled items for ease of testing - we can remove them if needed.
See it here: 
[basic tree view](https://pf-pr-8030.surge.sh/components/tree-view#single-selectable)
[compact](https://pf-pr-8030.surge.sh/components/tree-view#compact)
[compact no background](https://pf-pr-8030.surge.sh/components/tree-view#compact-no-background)
[With icons](https://pf-pr-8030.surge.sh/components/tree-view#with-icons), then add pf-m-compact to see that, then add pf-m-no-background to see that version.

## Summary by CodeRabbit

* **New Features**
  * TreeView items and expansion toggles support a disabled state; visuals and interactions reflect disabled styling and non-interactivity across base, compact, no-background, and nested contexts.
  * Disabled toggles render and behave as non-interactive so disabled nodes cannot be expanded or collapsed.

* **Documentation**
  * Examples and docs updated to show the disabled modifier and revised example labels indicating disabled items.

### DIFF
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -37,9 +37,9 @@ cssPrefix: pf-v6-c-tree-view
             {{/tree-view-list-item}}
           {{/tree-view-list}}
         {{/tree-view-list-item}}
-        {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
+        {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true" tree-view-list-item--modifier="pf-m-disabled"}}
           {{#> tree-view-content}}
-            {{> tree-view-node tree-view-node--text="Application 2"}}
+            {{> tree-view-node tree-view-node--text="Application 2 (disabled item)"}}
           {{/tree-view-content}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
--- a/src/patternfly/components/TreeView/examples/TreeView.md
+++ b/src/patternfly/components/TreeView/examples/TreeView.md
@@ -42,12 +42,13 @@ cssPrefix: pf-v6-c-tree-view
             {{> tree-view-node 
               tree-view-node--text="Application 2 (disabled item and toggle)"
               tree-view-node--IsDisabled="true"
-              tree-view-node-toggle--IsDisabled='true'}}
+              tree-view-node-toggle--IsDisabled="true"
+              tree-view-node--type='div'}}
           {{/tree-view-content}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
               {{#> tree-view-content}}
-                {{> tree-view-node tree-view-node--text="Settings"}}
+                {{> tree-view-node tree-view-node--text="Settings" tree-view-node--type='div'}}
               {{/tree-view-content}}
             {{/tree-view-list-item}}
             {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
@@ -401,7 +402,7 @@ A search input can be used to filter tree view items. It is recommended that a t
         {{/tree-view-list-item}}
         {{#> tree-view-list-item tree-view-list-item--IsExpandable="true" tree-view-list-item--IsExpanded="true"}}
           {{#> tree-view-content}}
-            {{> tree-view-node tree-view-node--HasFolderIcon="true" tree-view-node--IsDisabled="true" tree-view-node--text="Application 2 (disabled item but expansion toggle enabled)"}}
+            {{> tree-view-node tree-view-node--HasFolderIcon="true" tree-view-node--IsDisabled="true" tree-view-node-toggle--IsDisabled="true" tree-view-node--text="Application 2 (disabled item and toggle)"}}
           {{/tree-view-content}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item tree-view-list-item--IsExpandable="true"}}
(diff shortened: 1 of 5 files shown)


src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -429,6 +433,11 @@ $pf-v6-c-tree-view--MaxNesting: 10 !default;
   .#{$tree-view}__node-count {
     margin-inline-start: var(--#{$tree-view}__node-count--MarginInlineStart);
   }
+
+  .pf-m-disabled > .#{$tree-view}__content > & {
+    --#{$tree-view}__node--Color: var(--#{$tree-view}__list-item--m-disabled__node--Color);
+    --#{$tree-view}__node-icon--Color: var(--#{$tree-view}__list-item--m-disabled__node-icon--Color);
+  }
```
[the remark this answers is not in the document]
srambach: the item is still meant to be able to be toggled, so i did not change the toggle icon color

mcoker:
Looks like there is a stylelint error

```
src/patternfly/components/TreeView/tree-view.scss
  102:3  ✖  Unexpected empty line before custom property  custom-property-empty-line-before
```

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -347,6 +360,8 @@ $pf-v6-c-tree-view--MaxNesting: 10 !default;
 
   &.pf-m-no-background {
     --#{$tree-view}--m-compact__node-container--nested--BackgroundColor: var(--#{$tree-view}--m-no-background__node-container--BackgroundColor);
+    --#{$tree-view}--m-compact__node-container--m-disabled--nested--Color: inherit;
+    --#{$tree-view}--m-compact__node-icon--m-disabled--nested--Color: inherit;
```
mcoker: `inherit` on a CSS var inherits the var - doesn't apply the value `inherit` for the property it's applied to. To get the value `inherit`, what we typically do is something like this

```
--pf-c-component--Color: black;
--pf-c-component--m-inherit-color--Color: initial;

color: var(--pf-c-component--Color, inherit);

&.pf-m-inherit-color {
  --pf-c-component--Color: var(--pf-c-component--m-inherit-color--Color);
}
```

This is probably working because it's an invalid var and any `color` it's used on will break and revert to the initial value which is `inherit`.

mcoker (commented):
Left a comment about `inherit` on some vars.

I'm curious why the modifier is on the list-item and not the node? Looks like you'll always have to target direct child node since list-items contain nested list-items and nodes, and this isn't intended to be inherited. Nodes will never have nested anything in them, so you could style the children classes without descendent selectors. Also looks like the node is what supports `.pf-m-selectable` and `.pf-m-current`, so might make sense if that is also the item that is disabled.

Also looks like the original issue mentions toggle icons should support being disabled, but I didn't see that here.

And is this the correct color for default compact (with backgrounds) disabled nodes? 

Image
Image

src/patternfly/components/TreeView/tree-view.scss
kmcfaul: In the react PR, I've put temporary 'pf-m-disabled' classes on the treeViewNode when the tree view node is a button (specifically, when selection and toggling for a node isn't separated), and on the treeViewNodeText and treeViewNodeToggle for when the behaviors are separate.

So we would want styling for those three use cases, though disabling the toggle button individually was something I had added myself, since the initial issue was more geared towards disabling the selection behavior for split behavior tree view nodes.

src/patternfly/components/TreeView/examples/TreeView.md
```diff
@@ -39,12 +39,16 @@ cssPrefix: pf-v6-c-tree-view
...
+              tree-view-node--text="Application 2 (disabled item and toggle)"
+              tree-view-node--IsDisabled="true"
+              tree-view-node-toggle--IsDisabled="true"
+              tree-view-node--type='div'}}
           {{/tree-view-content}}
           {{#> tree-view-list newcontext}}
             {{#> tree-view-list-item}}
               {{#> tree-view-content}}
-                {{> tree-view-node tree-view-node--text="Settings"}}
+                {{> tree-view-node tree-view-node--text="Settings" tree-view-node--type='div'}}
```
mcoker:
```suggestion
                {{> tree-view-node tree-view-node--text="Settings" tree-view-node--type="div"}}
```

src/patternfly/components/TreeView/tree-view-node-toggle.hbs
```diff
@@ -1,8 +1,9 @@
-<{{#if tree-view-node-toggle--IsToggle}}button{{else}}span{{/if}} class="{{pfv}}tree-view__node-toggle{{#if tree-view-node-toggle--modifier}} {{tree-view-node-toggle--modifier}}{{/if}}"
+<{{#if tree-view-node-toggle--IsToggle}}button{{else}}span{{/if}} class="{{pfv}}tree-view__node-toggle {{#if tree-view-node-toggle--IsDisabled}}pf-m-disabled{{/if}} {{#if tree-view-node-toggle--modifier}} {{tree-view-node-toggle--modifier}}{{/if}}"
```
mcoker:
```suggestion
<{{#if tree-view-node-toggle--IsToggle}}button{{else}}span{{/if}} class="{{pfv}}tree-view__node-toggle{{#if tree-view-node-toggle--IsDisabled}} pf-m-disabled{{/if}} {{#if tree-view-node-toggle--modifier}} {{tree-view-node-toggle--modifier}}{{/if}}"
```

mcoker (approved):
Had one small nit but commented and committed the suggestion - https://github.com/patternfly/patternfly/pull/8030/changes/3684f4ee472b8f869e4e8491b5362f6d01984dd2

There are a few minor styling tweaks needed but opened a follow up for those so we can get this in react. Here's the follow up - https://github.com/patternfly/patternfly/issues/8053

Otherwise LGTM!! 🌳👀

PR #8034, opened 2025-12-18
feat(table): add support for tfoot

### LINKED ISSUE #8035: Table - support for table footer
To support a "glass" header/footer for table, we need to support for the `[tfoot]` element (compliments `[thead]`). It can just be a basic implementation that supports sticky and mimic `[thead]`.

##

---

**Jira Issue:** [PF-3275](https://redhat.atlassian.net/browse/PF-3275)

mcoker:
Table footer example - https://pf-pr-8034.surge.sh/components/table#table-footer
Sticky footer example - https://pf-pr-8034.surge.sh/components/table#sticky-footer

Towards https://github.com/patternfly/pf-roadmap/issues/282. This was demo code to show glass styles for table header/footer - cleaned it up to be a basic implementation for support for `<tfoot>`. It does not have any special glass styling, it's just support for the table footer. It does support being sticky, like the table header/`<thead>`

Fixes https://github.com/patternfly/patternfly/issues/8035

## Summary by CodeRabbit

* **New Features**
  * Added table footer support including a sticky footer option
  * Introduced public boolean attributes to enable plain tables, sticky headers, and sticky footers

* **Documentation**
  * Updated component docs and demos with footer examples and guidance for the new flags
  * Replaced modifier-based configuration examples with the new boolean attribute approach


src/patternfly/components/Table/table.scss
```diff
@@ -324,6 +325,29 @@
...
+        position: absolute;
+        inset: 0;
+        z-index: var(--#{$table}--m-sticky-footer--border--ZIndex);
+        pointer-events: none;
+        content: "";
+        border-block-start: var(--#{$table}--border-width--base) solid var(--#{$table}--BorderColor);
+      }
+    }
+
+    > .#{$table}__tbody:last-of-type > .#{$table}__tr:last-of-type {
```
jcmill, line 398: Because this executes whenever the sticky-footer modifier is present, what do you think about using the following incase someone doesn't end up using a tfoot (either accidentally or otherwise)? Otherwise that last border will be lost.

`&:has(tfoot) > .#{$table}__tbody:last-of-type > .#{$table}__tr:last-of-type {`

mcoker: Personally I could go either way. 2 downsides to that would be that 1) it increases the specificity, though shouldn't be a problem, and 2) the `:has()` selector isn't as performant as standard selectors, and tables tend to have a lot of nodes, and lots of nodes in `:has()` can make it slower. Tables often have a lot of rows and sometimes contain nested tables, and table performance is something I try to keep an eye on. That said, using a descendent selector like `&:has(> tfoot) > .#{$table}__tbody:last-of-type > .#{$table}__tr:last-of-type {` would probably mitigate any performance issues.

I would lean toward leaving it as-is, unless you can think of a valid use case where it could create a bug? If someone is using `pf-m-sticky-footer` without a `tfoot` I would probably just tell them to remove that class. WDYT - still think we should update it?

thatblindgeye (approved):
Can we also open a React followup for this if one doesnt exist

mcoker:
> Can we also open a React followup for this if one doesnt exist

@thatblindgeye 👍👍 https://github.com/patternfly/patternfly-react/issues/12354

kmcfaul (approved):
Implementation lgtm, but have a question about the design. 

Looking at the preview, I find the footer a bit hard to differentiate from a normal table row, especially when it's non-sticky / sticky but flush with the end of the table. My first reaction seeing the example was that the footer was missing until I looked more carefully at the bottom row, though I did quickly see it was footer after that. 

I'm wondering if anyone from design has any thoughts about this?

mcoker:
@kmcfaul one thing we talked about on Monday was removing the bottom border, so I just pushed that update. @lboehling @andrew-ronaldson any other updates you'd want to make? https://pf-pr-8034.surge.sh/components/table#table-footer

FWIW here it is with a secondary and striped background color

Screenshot 2026-04-14 at 2 06 16 PM
Screenshot 2026-04-14 at 2 06 47 PM

kmcfaul (approved):
I like it better without the bottom border 👍

src/patternfly/components/Table/table.scss
```diff
@@ -373,6 +377,34 @@
...
+      border-block-end: 0;
+    }
+  }
+
+  // - Table plain
+  &.pf-m-plain {
+    --#{$table}--BackgroundColor: transparent;
+  }
```
mcoker, line 406: Should be able to remove this

src/patternfly/components/Table/table.scss
```diff
@@ -373,6 +377,34 @@
...
+        pointer-events: none;
+        content: "";
+        border-block-start: var(--#{$table}--border-width--base) solid var(--#{$table}--BorderColor);
+      }
+    }
+
+    > .#{$table}__tbody:last-of-type > .#{$table}__tr:last-of-type {
+      border-block-end: 0;
+    }
+  }
```
mcoker, line 401: Refactor this to sticky-base/sticky-stuck

src/patternfly/components/Table/table-tfoot.hbs
```diff
@@ -0,0 +1,7 @@
+<tfoot class="{{pfv}}table__tfoot
+  {{#if table-tfoot--modifier}}{{table-tfoot--modifier}}{{/if}}"
```
mcoker, line 2:
```suggestion
<tfoot class="{{pfv}}table__tfoot{{#if table-tfoot--modifier}} {{table-tfoot--modifier}}{{/if}}"
```

andrew-ronaldson (approved):

PR #8041, opened 2026-01-05
chore(backstop): tweak parameters to reduce false positives

srambach:
This adjusts and adds some options to backstop to reduce the number of false positives. 
Some of these might not be necessary - we could try putting the delay back to 100ms for example.

Details of some of the changes according to Claude:
- `--force-device-scale-factor=1`
  - Forces 1x pixel density (no retina/HiDPI scaling)
  - Prevents subpixel rendering variations on different displays
  - Ensures consistent pixel-to-pixel rendering
- `--font-render-hinting=none`
  - Disables OS-level font hinting
  - Font hinting adjusts fonts to pixel grid differently on different systems
  - Removing it makes font rendering more consistent across runs
- `--disable-font-subpixel-positioning`
  - Prevents fonts from being positioned at fractional pixel coordinates
  - Forces integer pixel positions for text
  - Eliminates those "random pixels" you saw under breadcrumbs
- `--disable-gpu`
  - Uses CPU (software) rendering instead of GPU
  - GPU rendering can vary slightly between runs
  - Software rendering is slower but deterministic

`resembleOutputOptions`:
- `errorType: 'movementDifferenceIntensity'`
  - Algorithm for detecting pixel differences
  - Alternative to default 'flat' - better at detecting actual visual changes vs noise
  - More sophisticated detection that considers pixel intensity changes
- `ignoreAntialiasing: true` Most important
  - Ignores small pixel differences caused by antialiasing (font smoothing, edge rendering)
  - Helps reduce false positives from subpixel rendering variations
  - Critical for your diagonal artifact issue - prevents tiny antialiasing differences from being flagged

## Summary by CodeRabbit

* **New Features**
  * Per-URL configurable delay and global scenario defaults (delay, ready selector, remove selectors, mismatch threshold).

* **Improvements**
  * Test engine switched to Playwright with updated readiness scripts.
  * Navigation now waits for network idle and uses standardized launch flags for consistent rendering.
  * Visual comparison tuned to ignore antialiasing and movement noise.

* **Bug Fixes**
  * Reduced concurrent capture limit for improved reliability.

* **Style**
  * Multiple non-functional CSS declaration reorderings and formatting cleanups.


mcoker (approved):
LGTM! From testing, looks like it takes about twice as long to run now. If we want to try reducing that at some point, we can try bumping `asyncCaptureLimit` and reducing the `delay`

mcoker:
I was still getting artifacts that looked to be from partially loaded pages, so I made a couple of changes:
* Main thing is I updated the engine from puppeteer to playwright. We've wanted to do that for a while since playwright supports chrome/FF/safari.
* Added `await page.waitForLoadState('networkidle');` to the `onReady.js` script, which tells playwright to _waits until there are no more than two network connections for at least 500 milliseconds (0.5 seconds)_

Diffs look much better in playwright! I didn't see any more partially loaded pages/images, and no random dark theme elements on pages. I still see some diagonal/slanted diffs, but where I probably got 20 or so using puppeteer (plus a bunch of other partially loaded/dark-theme diffs), there were 7 in playwright. Plus the overall run takes less time (about 52 minutes compared to 56 minutes or so with a full run), even with the new `aitForLoadState('networkidle')`, which sounds like it may be adding 500ms to each test.

PDF report - https://drive.google.com/file/d/1KACVVOaxevAs4nSHPr3UcdTtc5IgrwdQ/view?usp=sharing
Full report - https://drive.google.com/file/d/1Pj8Npi46Ogozkqt5-nkGxlHr8d8xEfFL/view?usp=drive_link

FWIW this is the list of slanted diffs that I spotted

* filename: pf-core__components_table_html_overflow-menu-usage-desktop_0_document_0_mobile.pn
* filename: pf-core__components_table_html_table-with-long-strings-in-the-content_0_document_0_mobile.png
* filename: pf-core__components_input-group_html_variations_0_document_0_mobile.png
* filename: pf-core__components_input-group_html_variations_0_document_1_tablet.png
* filename: pf-core__components_forms_form-control_html_select_0_document_1_tablet.png
* filename: pf-core__components_forms_form-control_html_select_0_document_1_tablet.png
* filename: pf-core__components_card_html-demos_status-card-expanded-with-popover_0_document_0_mobile.png

mcoker:
Re-ran a full job from main with the latest changes, got the same number of pass/fails - though I didn't accept the previously run screenshots, so I accepted and ran a few one-off's to validate and got 0 failures. Those were with `--filter` set to table, form, and input-group, since those normally produce some errors.

PR #8063, opened 2026-01-21
chore(backstop): add custom description and reset params

srambach:
This sets the timeout back to 100, async captures to 5, and adds a command line parameter `--desc="custom description"` that will put the description at the top of the report. I added it so that we can more easily differentiate different runs.


mcoker (approved):
🔙🛑👍

PR #8067, opened 2026-01-22
Post process backstop report name

srambach:
Backstop doesn't differentiate the id shown at the top of a report from the initial part of the screen capture filenames.
This PR introduces a wrapper script that changes the name at the top of the report after it's created.

Note: in testing, it works well but (probably because the html report is opened in the browser immediately after creation) you need to reload the page to see the changed name.

## Summary by CodeRabbit

* **New Features**
  * New test runner CLI that executes visual tests and applies post-processing.
  * Post-processing step can inject a custom report description and apply a dark-theme variant.

* **Chores**
  * Test scripts updated to use the new runner.
  * Reports now use a consistent, fixed report identifier for generated outputs.

### DIFF
--- a/backstop.js
+++ b/backstop.js
@@ -6,10 +6,6 @@ const viewports = [];
 const isDarkTheme = process.argv.includes('--dark');
 const themeSuffix = isDarkTheme ? '_dark' : '';
 
-// check for --desc flag to add a custom description to the report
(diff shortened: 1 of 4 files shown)


srambach:
Closing since it's more complicated than it's worth

PR #8068, opened 2026-01-22
chore(backstop): revert desc feature

srambach:
The --desc option caused problems with the reference file names since they also contain the id string. This reverts that feature.
It also sets the asynch capture number back to 3 for possibly a happy medium of speed and reliability. Also it nicely groups the captures of each test in the 3 viewport settings.

## Summary by CodeRabbit

* **Refactor**
  * Removed support for custom report descriptions
  * Standardized report identifier to a fixed value
  * Increased concurrent screenshot capture operations for higher throughput


mcoker (approved):
🎉

PR #8083, opened 2026-01-31
feat: expand clickable area for standalone radio and checkbox

### LINKED ISSUE #7673: Bug - [SelectColumn] -  label intended to increase clickable area has no effect in PF6
**Describe the problem**

It looks like the label wrapping the input in `SelectColumn`, which was originally added in PatternFly 4 to increase the clickable area, no longer seems to have any effect in PatternFly 6:

https://github.com/patternfly/patternfly-react/blob/87ddd2d8cf02bacbf69c8950daff0137a9fc5c74/packages/react-table/src/components/Table/SelectColumn.tsx#L35

Visit https://www.patternfly.org/components/table#selectable-radio-input and try clicking just outside but near the radio input. It doesn’t get selected.

Then compare this behavior with the archived PatternFly 4 version at https://v4-archive.patternfly.org/v4/components/table#composable-selectable-radio-input. Here, clicking in the surrounding "column area" does select the radio input, and the cursor style differs as well.

**Expected behavior**

Not entirely sure what the behavior should be, but I personally lean toward increasing the actual input size as somehow mentioned in 
https://github.com/patternfly/patternfly/pull/4546#issuecomment-984767013

In my view, making native elements more prominent feels more appropiated than using a *transparent label* to expand the clickable area, which could be unexpected regarding to native behavior.

Not at all. Just noticed this while reviewing the code and was curious about the intention behind such a  wrapping label.

**Any other information?**

Such a label was added at https://github.com/patternfly/patternfly-react/pull/6754 in patternfly-react package after previously added at https://github.com/patternfly/patternfly/pull/4546 in patternfly package.

---

**Jira Issue:** [PF-2398](https://redhat.atlassian.net/browse/PF-2398)

tarunvashishth:
Implement 44x44px minimum touch target for standalone radio and checkbox components using ::before pseudo-element approach. This improves accessibility and usability on mobile devices while maintaining the original visual layout and shape of the components.

The expanded clickable area is achieved through a CSS pseudo-element with an inset of -12px on all sides, providing an invisible but clickable expansion area. The table component inherits this improvement automatically.

- Add --pf-v6-c-radio--m-standalone--before--Inset component variable
- Add --pf-v6-c-check--m-standalone--before--Inset component variable
- Add position: relative to .pf-m-standalone radio and checkbox
- Add ::before pseudo-element with expanded inset for clickable area

Assisted-by: GitHub Copilot

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded the interactive hit area for standalone checkboxes and radio buttons to make them easier to select.
  * Added configurable spacing tokens and an overlay element so the clickable region is larger while preserving layout and cursor behavior.

### DIFF
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -4,6 +4,7 @@
   --#{$check}--GridGap: var(--pf-t--global--spacer--gap--group--vertical) var(--pf-t--global--spacer--gap--text-to-element--default);
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
   --#{$check}--m-standalone--MinHeight: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
+
   // Expanded clickable area for standalone checkbox; provides 44x44px minimum touch target per WCAG guidelines
   --#{$check}--m-standalone--before--Inset: -12px;
 
@@ -40,10 +41,10 @@
 
     // Expanded clickable area using pseudo-element
     &::before {
-      content: '';
       position: absolute;
       inset: var(--#{$check}--m-standalone--before--Inset);
       pointer-events: auto;
+      content: '';
     }
 
     .#{$check}__input {
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -4,6 +4,8 @@
   --#{$check}--GridGap: var(--pf-t--global--spacer--gap--group--vertical) var(--pf-t--global--spacer--gap--text-to-element--default);
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
   --#{$check}--m-standalone--MinHeight: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
+  // Expanded clickable area for standalone checkbox; provides 44x44px minimum touch target per WCAG guidelines
+  --#{$check}--m-standalone--before--Inset: -12px;
 
   // TODO: update to `--#{$check}--FontSize` `--#{$check}--LineHeight`
   --#{$check}__label--disabled--Color: var(--pf-t--global--text--color--disabled);
@@ -31,9 +33,18 @@
   accent-color: var(--#{$check}--AccentColor);
 
   &.pf-m-standalone {
+    position: relative;
     display: inline-grid;
     grid-template-columns: auto;
     min-height: var(--#{$check}--m-standalone--MinHeight);
+
+    // Expanded clickable area using pseudo-element
+    &::before {
+      content: '';
+      position: absolute;
+      inset: var(--#{$check}--m-standalone--before--Inset);
+      pointer-events: auto;
+    }
 
     .#{$check}__input {
       align-self: center;
--- a/src/patternfly/components/Check/check.scss
+++ b/src/patternfly/components/Check/check.scss
@@ -6,7 +6,7 @@
   --#{$check}--m-standalone--MinHeight: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
 
   // Standalone
-  --#{$check}--m-standalone--before--Inset: calc(var(--pf-t--global--spacer--md) * -1);
+  --#{$check}--m-standalone--before--Inset: calc(var(--pf-t--global--spacer--sm) * -1);
 
   // TODO: update to `--#{$check}--FontSize` `--#{$check}--LineHeight`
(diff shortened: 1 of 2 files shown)


tarunvashishth:
Fixes #7673

src/patternfly/components/Check/check.scss
```diff
@@ -5,6 +5,9 @@
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
   --#{$check}--m-standalone--MinHeight: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
 
+  // Expanded clickable area for standalone checkbox; provides 44x44px minimum touch target per WCAG guidelines
+  --#{$check}--m-standalone--before--Inset: -12px;
```
jcmill: @mcoker correct me if I'm wrong here but could we use `calc(var(--pf-t--global--spacer--md) * -1)` here instead of the raw value? Plus if we are trying to hit that min of 44px the 12px isn't quite making it on the width (outputting 37px at -12px).

src/patternfly/components/Check/check.scss
```diff
@@ -5,6 +5,9 @@
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
   --#{$check}--m-standalone--MinHeight: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
 
+  // Expanded clickable area for standalone checkbox; provides 44x44px minimum touch target per WCAG guidelines
```
jcmill: This is a nit but this could just be `// Standalone`.

src/patternfly/components/Check/check.scss
```diff
@@ -31,10 +34,20 @@
   accent-color: var(--#{$check}--AccentColor);
 
   &.pf-m-standalone {
+    position: relative;
     display: inline-grid;
     grid-template-columns: auto;
     min-height: var(--#{$check}--m-standalone--MinHeight);
 
+    // Expanded clickable area using pseudo-element
```
jcmill: Same nit here. Though we don't really need the comment.

src/patternfly/components/Radio/radio.scss
```diff
@@ -5,6 +5,9 @@
   --#{$radio}--AccentColor: var(--pf-t--global--icon--color--brand--default);
   --#{$radio}--m-standalone--MinHeight: calc(var(--#{$radio}__label--FontSize) * var(--#{$radio}__label--LineHeight));
 
+  // Expanded clickable area for standalone radio; provides 44x44px minimum touch target per WCAG guidelines
+  --#{$radio}--m-standalone--before--Inset: -12px;
```
jcmill: Potentially same calc here.

jcmill (commented):
Once again @tarunvashishth, appreciate the work you put in.

src/patternfly/components/Check/check.scss
```diff
@@ -5,6 +5,9 @@
   --#{$check}--AccentColor: var(--pf-t--global--color--brand--default);
   --#{$check}--m-standalone--MinHeight: calc(var(--#{$check}__label--FontSize) * var(--#{$check}__label--LineHeight));
 
+  // Standalone
+  --#{$check}--m-standalone--before--Inset: calc(var(--pf-t--global--spacer--md) * -1);
```
jcmill:
```suggestion
  --#{$check}--m-standalone--before--Inset: calc(var(--pf-t--global--spacer--sm) * -1);
```

src/patternfly/components/Radio/radio.scss
```diff
@@ -5,6 +5,9 @@
   --#{$radio}--AccentColor: var(--pf-t--global--icon--color--brand--default);
   --#{$radio}--m-standalone--MinHeight: calc(var(--#{$radio}__label--FontSize) * var(--#{$radio}__label--LineHeight));
 
+  // Standalone
+  --#{$radio}--m-standalone--before--Inset: calc(var(--pf-t--global--spacer--md) * -1);
```
jcmill:
```suggestion
  --#{$radio}--m-standalone--before--Inset: calc(var(--pf-t--global--spacer--sm) * -1);
```

jcmill (commented):
My bad, Tarun. We talked a bit through the size used here. In order to avoid overlapping with other components, we need to reduce the size to -8px (pr the --sm version of the token). It will still meet AA for WCAG.

jcmill (approved):
✅ @mcoker, I ran a visual regression test and found no changes for the checkbox and radio standalones.

mcoker (approved):
LGTM 🚀

PR #8134, opened 2026-02-13
fix(notification-drawer): prevent list item layout shift when marking as read

### LINKED ISSUE #8091: Bug - Notification drawer item borders cause layout shift
To reproduce:
3. Go here and click on the notification bell to expand notification drawer
4. 1. Click either of the unread items (the first two)
5. 1. Note the layout shift 😭

Expected behavior:
Clicking on an unread item should not cause a layout shift

Probably just need to move the border on `.pf-v6-c-notification-drawer__list-item` to a pseudo element of that same element.

---

**Jira Issue:** [PF-3506](https://issues.redhat.com/browse/PF-3506)

tarunvashishth:
Fixes #8091.

Prevents Notification Drawer layout shift when unread items are marked read by moving the list-item border from the element itself to a `::before` pseudo-element. Visual styles remain the same; box size no longer changes.

Assisted-by: GitHub Copilot

## Summary by CodeRabbit

* **Style**
  * Notification drawer list-item borders now render via a layered overlay for more consistent visuals and preserved corner radii.
  * Removed direct borders from list items and adjusted read-state positioning to align with the overlay approach.
  * Added an adjustable focus outline offset to improve keyboard focus visuals.
  * Preserved existing state-specific border colors (info, warning, danger, success, custom, read).

### DIFF
--- a/src/patternfly/components/NotificationDrawer/notification-drawer.scss
+++ b/src/patternfly/components/NotificationDrawer/notification-drawer.scss
@@ -37,6 +37,7 @@
   --#{$notification-drawer}__list-item--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$notification-drawer}__list-item--BorderWidth: var(--pf-t--global--border--width--box--status--default);
   --#{$notification-drawer}__list-item--BorderRadius: var(--pf-t--global--border--radius--medium);
+  --#{$notification-drawer}__list-item--focus--OutlineOffset: #{pf-size-prem(-4px)};
 
   // List item modifiers
   --#{$notification-drawer}__list-item--m-info__list-item-header-icon--Color: var(--pf-t--global--icon--color--status--info--default);
@@ -220,6 +221,10 @@
     &:focus {
       background-color: var(--#{$notification-drawer}__list-item--m-hoverable--hover--BackgroundColor);
     }
+
+    &:focus-visible {
+      outline-offset: var(--#{$notification-drawer}__list-item--focus--OutlineOffset);
+    }
   }
 }
 
@@ -322,4 +327,3 @@
     transform: rotate(var(--#{$notification-drawer}__group--m-expanded__group-toggle-icon--Rotate));
   }
 }
-
--- a/src/patternfly/components/NotificationDrawer/notification-drawer.scss
+++ b/src/patternfly/components/NotificationDrawer/notification-drawer.scss
@@ -171,8 +171,16 @@
   padding-inline-start: var(--#{$notification-drawer}__list-item--PaddingInlineStart);
   padding-inline-end: var(--#{$notification-drawer}__list-item--PaddingInlineEnd);
   background-color: var(--#{$notification-drawer}__list-item--BackgroundColor);
-  border: var(--#{$notification-drawer}__list-item--BorderWidth) solid var(--#{$notification-drawer}__list-item--BorderColor);
   border-radius: var(--#{$notification-drawer}__list-item--BorderRadius);
+
+  &::before {
+    position: absolute;
+    inset: 0;
+    pointer-events: none;
+    content: '';
+    border: var(--#{$notification-drawer}__list-item--BorderWidth) solid var(--#{$notification-drawer}__list-item--BorderColor);
+    border-radius: inherit;
+  }
 
   &.pf-m-info {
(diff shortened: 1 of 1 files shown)


mcoker (changes requested):
LGTM except for one thing. The focus outline on notification drawer items is now covered by the border. Can you address that? One fix could be to use an outline offset of -4px.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -216,6 +221,10 @@
     &:focus {
       background-color: var(--#{$notification-drawer}__list-item--m-hoverable--hover--BackgroundColor);
     }
+
+    &:focus-visible {
+      outline-offset: var(--#{$notification-drawer}__list-item--focus--OutlineOffset);
```
mcoker: Can you apply this style to the `.#{$notification-drawer}__list-item` element directly - not just on `.pf-m-hoverable&:focus-visible`? Since items have a border, seems fair that when any receive focus, the outline should be offset. FWIW we also do not currently use `outline` for any purposes other than focus outlines.

mcoker (changes requested):
Just one update please, otherwise LGTM!

src/patternfly/components/NotificationDrawer/notification-drawer.scss
tarunvashishth: I’ve applied the style to `.#{$notification-drawer}__list-item`.

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -173,9 +172,21 @@
...
+    inset: 0;
+    pointer-events: none;
+    content: '';
+    border: var(--#{$notification-drawer}__list-item--BorderWidth) solid var(--#{$notification-drawer}__list-item--BorderColor);
+    border-radius: inherit;
+  }
+
+  &:focus-visible {
+    outline-offset: var(--#{$notification-drawer}__list-item--focus--OutlineOffset);
+  }
```
mcoker: Sorry, what I meant was to move `outline-offset` to the main block of declarations for `.#{$notification-drawer}__list-item` - no need to just scope it to `&:focus-visible`

https://github.com/tarunvashishth/patternfly/blob/3ecfd7440d7e31b040fa2e36d7f3f08518cd3351/src/patternfly/components/NotificationDrawer/notification-drawer.scss#L166-L175

mcoker (changes requested):

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -38,6 +37,7 @@
   --#{$notification-drawer}__list-item--BackgroundColor: var(--pf-t--global--background--color--primary--default);
   --#{$notification-drawer}__list-item--BorderWidth: var(--pf-t--global--border--width--box--status--default);
   --#{$notification-drawer}__list-item--BorderRadius: var(--pf-t--global--border--radius--medium);
+  --#{$notification-drawer}__list-item--focus--OutlineOffset: #{pf-size-prem(-4px)};
```
mcoker:
```suggestion
  --#{$notification-drawer}__list-item--OutlineOffset: #{pf-size-prem(-4px)};
```

src/patternfly/components/NotificationDrawer/notification-drawer.scss
```diff
@@ -173,8 +172,17 @@
   padding-inline-start: var(--#{$notification-drawer}__list-item--PaddingInlineStart);
   padding-inline-end: var(--#{$notification-drawer}__list-item--PaddingInlineEnd);
   background-color: var(--#{$notification-drawer}__list-item--BackgroundColor);
-  border: var(--#{$notification-drawer}__list-item--BorderWidth) solid var(--#{$notification-drawer}__list-item--BorderColor);
   border-radius: var(--#{$notification-drawer}__list-item--BorderRadius);
+  outline-offset: var(--#{$notification-drawer}__list-item--focus--OutlineOffset);
```
mcoker:
```suggestion
  outline-offset: var(--#{$notification-drawer}__list-item--OutlineOffset);
```

mcoker (changes requested):
Coderabbit has a good point - no need for `--focus` in the var name now that the style no longer applies specifically for the focus state.

mcoker (approved):

PR #8136, opened 2026-02-16
fix(button): define missing variables and remove unused ones

### LINKED ISSUE #7952: Bug - Button - Unused/undefined variables
There are a few variables in button.scss that aren't defined or are unused:

Variable not defined - this one is a bug in the large (CTA) button font size:
Line 737: --#{$button}--m-display-lg--FontSize is referenced but never defined.
Line 283 defines --#{$button}--m-link--m-display-lg--FontSize (used for link buttons on line 577)
Line 737 references --#{$button}--m-display-lg--FontSize (for non-link display-lg buttons), but this variable is not defined

Variables defined but never used: (determine if needed and use or remove)
--#{$button}__progress--Opacity (line 337) — Defined but never used. The progress element doesn't use opacity.
--#{$button}--m-display-lg--FontWeight (line 282) — Defined but never used. The pf-m-display-lg modifier sets padding and font size but not font weight. (probably goes with the large button text size above
--#{$button}--disabled--BorderColor (line 295) — Defined but never used. The disabled styles use border-color: transparent; directly in the &::after block (line 843) instead of the variable.

---

**Jira Issue:** [PF-2902](https://redhat.atlassian.net/browse/PF-2902)

tarunvashishth:
Fixes #7952 

**PR Summary**
This PR fixes undefined/unused Button CSS variable issues in `src/patternfly/components/Button/button.scss`.

- Added missing `display-lg` token: `--#{$button}--m-display-lg--FontSize`.
- Wired `--#{$button}--m-display-lg--FontWeight` into `.pf-m-display-lg`.
- Replaced hardcoded disabled border color with `--#{$button}--disabled--BorderColor`.
- Wired hover icon transition delay token into hover/focus state.
- Added base corner-radius fallback tokens (`BorderStartStart/StartEnd/EndStart/EndEndRadius`) to match existing `var(...)` usage.
- Removed unused `--#{$button}__progress--Opacity`.

## Summary by CodeRabbit

* **New Features**
  * Added font-size controls for large and link-large button variants
  * Enhanced typography controls for large button variants

* **Style**
  * Added configurable icon hover transition delay
  * Improved disabled button border visibility
  * Refined button progress element styling (removed default hidden opacity)


src/patternfly/components/Button/button.scss
```diff
@@ -278,6 +282,7 @@
   --#{$button}--m-display-lg--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--spacious);
   --#{$button}--m-display-lg--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--spacious);
   --#{$button}--m-display-lg--PaddingInlineStart: var(--pf-t--global--spacer--action--horizontal--spacious);
+  --#{$button}--m-display-lg--FontSize: var(--pf-t--global--font--size--body--lg);
   --#{$button}--m-display-lg--FontWeight: var(--pf-t--global--font--weight--body--bold);
   --#{$button}--m-link--m-display-lg--FontSize: var(--pf-t--global--font--size--body--lg);
```
jcmill: This is a small nit, but this could be moved to the block of tokens from line 95-115 (Link)

src/patternfly/components/Button/button.scss
```diff
@@ -18,6 +18,10 @@
   --#{$button}--BorderColor: transparent;
   --#{$button}--BorderWidth: var(--pf-t--global--border--width--action--default);
   --#{$button}--BorderRadius: var(--pf-t--global--border--radius--pill);
+  --#{$button}--BorderStartStartRadius: var(--#{$button}--BorderRadius);
+  --#{$button}--BorderStartEndRadius: var(--#{$button}--BorderRadius);
+  --#{$button}--BorderEndStartRadius: var(--#{$button}--BorderRadius);
+  --#{$button}--BorderEndEndRadius: var(--#{$button}--BorderRadius);
```
jcmill: Because there is a fallback defined in the button border-radius, we don't need to add these. It creates some unnecessary overhead in an already large file. But these do get defined in places like a horizontal scrolling nav.

src/patternfly/components/Button/button.scss
```diff
@@ -278,8 +279,8 @@
   --#{$button}--m-display-lg--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--spacious);
   --#{$button}--m-display-lg--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--spacious);
   --#{$button}--m-display-lg--PaddingInlineStart: var(--pf-t--global--spacer--action--horizontal--spacious);
+  --#{$button}--m-display-lg--FontSize: var(--pf-t--global--font--size--body--lg);
```
mcoker: This would change the font-size, so we can't use `--pf-t--global--font--size--body--lg`.

Right now the font-size for the `m-display-lg` button is computing to `inherit` since this var is undefined, and the button has `font-size: var(--pf-v6-c-button--FontSize, inherit);`. That's no good because we don't want the button's font-size to change if a parent of the button has a custom font-size. 

So I would just set this to our default font size token, the same as is used for `--pf-v6-c-button--FontSize`

```suggestion
  --#{$button}--m-display-lg--FontSize: var(--pf-t--global--font--size--body--default);
```

src/patternfly/components/Button/button.scss
```diff
@@ -93,6 +93,7 @@
   --#{$button}--m-link--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--plain--default);
   --#{$button}--m-link--PaddingInlineStart: var(--pf-t--global--spacer--action--horizontal--plain--default);
   --#{$button}--m-link--Color: var(--pf-t--global--text--color--brand--default);
+  --#{$button}--m-link--m-display-lg--FontSize: var(--pf-t--global--font--size--body--lg);
```
mcoker: If you rebase, you'll pull in a change for this var that sets it to `--pf-t--global--font--size--body--default`, so you can revert the change in this PR 

https://github.com/patternfly/patternfly/blob/716f0e546cb9da7dade2907e6ad1f32584da027b/src/patternfly/components/Button/button.scss#L293

src/patternfly/components/Button/button.scss
```diff
@@ -745,6 +746,7 @@
     --#{$button}--PaddingBlockEnd: var(--#{$button}--m-display-lg--PaddingBlockEnd);
     --#{$button}--PaddingInlineStart: var(--#{$button}--m-display-lg--PaddingInlineStart);
     --#{$button}--FontSize: var(--#{$button}--m-display-lg--FontSize);
+    --#{$button}--FontWeight: var(--#{$button}--m-display-lg--FontWeight);
```
mcoker: No need for this style since it isn't currently being used. We can just mark the var for removal in our next breaking change release.

```suggestion
```

src/patternfly/components/Button/button.scss
```diff
@@ -278,8 +279,8 @@
   --#{$button}--m-display-lg--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--spacious);
   --#{$button}--m-display-lg--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--spacious);
   --#{$button}--m-display-lg--PaddingInlineStart: var(--pf-t--global--spacer--action--horizontal--spacious);
+  --#{$button}--m-display-lg--FontSize: var(--pf-t--global--font--size--body--lg);
   --#{$button}--m-display-lg--FontWeight: var(--pf-t--global--font--weight--body--bold);
```
mcoker: Can you add a comment to remove this in a breaking change?


```suggestion
  --#{$button}--m-display-lg--FontWeight: var(--pf-t--global--font--weight--body--bold); // TODO - remove in a breaking change
```

src/patternfly/components/Button/button.scss
```diff
@@ -862,7 +865,7 @@
     background: var(--#{$button}--disabled--BackgroundColor);
 
     &::after {
-      border-color: transparent;
+      border-color: var(--#{$button}--disabled--BorderColor);
```
mcoker, line 945: 👍

src/patternfly/components/Button/button.scss
```diff
@@ -820,6 +822,7 @@
     --#{$button}--ScaleY: var(--#{$button}--hover--ScaleY);
     --#{$button}--m-plain--m-no-padding__icon--Color: var(--#{$button}--m-plain--m-no-padding--hover__icon--Color);
     --#{$button}__icon--Color: var(--#{$button}--hover__icon--Color);
+    --#{$button}__icon--TransitionDelay: var(--#{$button}--hover__icon--TransitionDelay);
```
mcoker: Since this isn't needed, we can save a few lines of code and remove the declaration that tries to set it at https://github.com/patternfly/patternfly/blob/716f0e546cb9da7dade2907e6ad1f32584da027b/src/patternfly/components/Button/button.scss#L878

Then we can remove this and `--#{$button}--hover__icon--TransitionDelay` var that was added above.

mcoker (changes requested):
Thanks @tarunvashishth! Noted in a comment but just a heads up that you'll need to rebase and fix a couple of conflicts.

mcoker (approved):
LGTM, visual regressions passed.

PR #8159, opened 2026-02-20
fix(expandable-section): align nested indented content

### LINKED ISSUE #7881: Bug - ExpandableSection - nested indented ExpandableSections look misaligned
**Describe the problem**
This is an issue with 

- [x] Patternfly 6  

Nested indented ExpandableSections look misaligned.

When we have an ExpandableSection inside another ExpandableSection with the `isIndented` prop, then the inner one will look misaligned.

This is due to the "invisible" left padding on the button, which is only visible on hover. So in the default state the ExpandableSection will look more indented than the text above it. Same thing would happen with any link button.

PF5:

image

PF6:

image

image

**Expected behavior**
I think that we could set a negative margin on any child of an ExpandableSection with such "invisible" padding like the link button. So it moves a bit to the left and aligns with text or other elements above / below it. And only on hover the left padding is visible.

We did this as a workaround on a product, where we have lots of nested ExpandableSections aligned with other text.

**What is your environment?**
 - OS: macOS
 - Browser: firefox
 - Version: 143

---

**Jira Issue:** [PF-2737](https://redhat.atlassian.net/browse/PF-2737)

tarunvashishth:
Fixes #7881 

Implements a component-level fix to resolve misalignment of nested indented expandable sections.

- Added a new token:
  `--pf-v6-c-expandable-section--m-indented--nested--Offset`
  derived from the link-button plain horizontal padding token.

- Added a nested-specific alignment rule so direct child indented
  expandable sections inside `.pf-v6-c-expandable-section__content`
  apply a negative `margin-inline-start` to offset the invisible
  link-button start padding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved indentation handling for nested expandable sections, ensuring proper visual alignment when using the indented style variant.

### DIFF
--- a/src/patternfly/components/ExpandableSection/expandable-section.scss
+++ b/src/patternfly/components/ExpandableSection/expandable-section.scss
@@ -60,6 +60,7 @@
 
   // Indented
   --#{$expandable-section}--m-indented__content--PaddingInlineStart: calc(var(--pf-t--global--spacer--action--horizontal--plain--default) + var(--pf-t--global--spacer--gap--text-to-element--default) + var(--#{$expandable-section}__toggle-icon--MinWidth));
+  --#{$expandable-section}--m-indented--nested--Offset: var(--pf-t--global--spacer--action--horizontal--plain--default);
 
   // Truncate
   --#{$expandable-section}--m-truncate__content--LineClamp: 3;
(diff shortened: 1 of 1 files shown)


jcmill:
@tarunvashishth once again, thanks for jumping in and tackling issues. I wanted to make you aware, I requested Designs input on how this specific issue should be handled. It's a bit of a tricky on and could be handled in multiple ways if it's picked up.

jcmill:
@tarunvashishth, we've come full circle. I spoke with design after the interactions in the interview and we are closing the issue and this PR for now. If interest peaks, we will revisit it. Thanks again for the help.

tarunvashishth:
Great to hear! Thanks for the update.

PR #8166, opened 2026-02-25
fix(button): support centered progress spinner for block buttons and restore progress animation.

### LINKED ISSUE #7732: Bug - Button - progress doesn't work with block buttons
Looks like the progress spinner doesn't work as intended in block buttons.

image

@andrew-ronaldson @lboehling looks like a bug to me - is there any reason we wouldn't support the loading spinner in a block button or should we go ahead and update it so that it works (the spinner is centered beside the button text)?

Also it looks when we added the [button click/ripple animation](https://github.com/patternfly/patternfly/pull/6675), we ended up dropping a small animation that occurred on progress/spinner buttons. You can see it here in the v5 archive site, the button text slides in and out to make room for the spinner  - https://v5-archive.patternfly.org/components/button/react/progress-indicators/

Here's the same example on org - note the text shifts automatically without sliding - https://www.patternfly.org/components/button/react/progress-indicators/

I don't remember talking about removing that - I'm assuming I just missed it when reviewing the ripple animation PR. @lboehling @andrew-ronaldson do either of you remember if that was intentional or just something we overlooked and should add back in?

cc @wise-king-sullyman

---

**Jira Issue:** [PF-2521](https://redhat.atlassian.net/browse/PF-2521)

tarunvashishth:
Fix spinner layout for `.pf-m-block` buttons so the progress indicator
is centered alongside button text instead of using absolute positioning.

Restore the progress padding animation that was unintentionally removed
when ripple/click animation updates changed transition handling. The
button text now slides to make space for the spinner as in v5 behavior.

Fixes #7732

## Summary by CodeRabbit

* **Style**
  * Improved the button progress indicator’s positioning and centering.
  * Updated progress transitions to use configurable timing and properties for smoother updates.
  * Enhanced block-style button behavior so progress indicators align consistently and respond better to resizing and state changes.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -339,6 +339,8 @@
   --#{$button}__progress--InsetBlockStart: 50%;
   --#{$button}__progress--InsetInlineStart: var(--pf-t--global--spacer--control--horizontal--default);
   --#{$button}__progress--Color: var(--#{$button}__icon--Color);
+  --#{$button}--m-progress--TransitionProperty: padding, color, background, border-color;
+  --#{$button}--m-progress--TransitionDuration: 250ms;
   --#{$button}--m-progress--PaddingInlineEnd: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--#{$button}__progress--width) / 2);
   --#{$button}--m-progress--PaddingInlineStart: calc(var(--pf-t--global--spacer--control--horizontal--default) + var(--#{$button}__progress--width) / 2);
   --#{$button}--m-in-progress--PaddingInlineEnd: var(--pf-t--global--spacer--control--horizontal--default);
@@ -880,6 +882,8 @@
   &.pf-m-progress {
     --#{$button}--PaddingInlineEnd: var(--#{$button}--m-progress--PaddingInlineEnd);
     --#{$button}--PaddingInlineStart: var(--#{$button}--m-progress--PaddingInlineStart);
+    --#{$button}--TransitionProperty: var(--#{$button}--m-progress--TransitionProperty);
+    --#{$button}--TransitionDuration: var(--#{$button}--m-progress--TransitionDuration);  
   }
 
   &.pf-m-in-progress {
@@ -952,6 +956,16 @@
 
   .#{$spinner} {
     --#{$spinner}--Color: currentcolor;
+  }
+}
+
+.#{$button}.pf-m-block {
+  --#{$button}--AlignItems: center;
+
+  .#{$button}__progress {
+    position: static;
+    inset: auto;
+    transform: none;
   }
 }
 
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -796,6 +796,10 @@
     width: var(--#{$button}--m-block--Width);
     transition-duration: var(--#{$button}--m-progress--TransitionDuration);
     transition-property: var(--#{$button}--m-progress--TransitionProperty);
+
+    .#{$button}__progress {
+      align-self: center;
+    }
   }
 
   &.pf-m-small {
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -339,6 +339,8 @@
   --#{$button}__progress--InsetBlockStart: 50%;
   --#{$button}__progress--InsetInlineStart: var(--pf-t--global--spacer--control--horizontal--default);
   --#{$button}__progress--Color: var(--#{$button}__icon--Color);
+  --#{$button}--m-progress--TransitionProperty: padding, color, background, border-color;
+  --#{$button}--m-progress--TransitionDuration: 250ms;
```
jcmill: On line 27 (--#{$button}--TransitionProperty), you could add padding and then delete both these line (342-343) for the animation that "slides" the text.

src/patternfly/components/Button/button.scss
```diff
@@ -880,6 +882,8 @@
   &.pf-m-progress {
     --#{$button}--PaddingInlineEnd: var(--#{$button}--m-progress--PaddingInlineEnd);
     --#{$button}--PaddingInlineStart: var(--#{$button}--m-progress--PaddingInlineStart);
+    --#{$button}--TransitionProperty: var(--#{$button}--m-progress--TransitionProperty);
+    --#{$button}--TransitionDuration: var(--#{$button}--m-progress--TransitionDuration);
```
jcmill: After the first change, you shouldn't need these either.

src/patternfly/components/Button/button.scss
```diff
@@ -955,6 +959,16 @@
...
 
+.#{$button}.pf-m-block {
+  --#{$button}--AlignItems: center;
+
+  .#{$button}__progress {
+    position: static;
+    inset: auto;
+    transform: none;
+  }
+}
```
jcmill: For this, @mcoker, you may want to confirm. But I would remove this block.
- create a `--#{$button}__progress--position: absolute;` with the other __progress variables
- use it in the position for the `__progress` selector block. and add a `align-self: center` in that block (this will properly center the spinner when the position changes from absolute to relative or static in the `.pf-m-block` chunk. Just switching the element back to the normal flow, there is still a slight px difference in it's vertical positioning which is where the align-self can help).
- For the block modifier we can redefine the variables needed. Like the following in but placed inside `.pf-m-block`:
    `--#{$button}__progress--position: relative;`
    `--#{$button}__progress--InsetInlineStart: 0;`
    `--#{$button}__progress--TranslateY: 0;`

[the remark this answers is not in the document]
jcmill: @tarunvashishth, this is correct... in part. You could move these variables to the first `pf-m-block` and delete this block. You should not need the `__process` block coderabbit is suggesting though.

jcmill (commented):
This is looking good Tarun. Thanks for the turnaround. Just one maintenance-type comment and anything else @mcoker would like to add.

jcmill:
Tarun, I apologize. I misspoke about adding the padding to the `--#{$button}--TransitionProperty`. I hadn't totally thought that through as it's not performative. What you had was correct, adding that padding transition back into the progress tokens specifically. I'm working on rebasing this and will revert my suggestion shortly.

tarunvashishth:
> I apologize. I misspoke about adding the padding to the `--#{$button}--TransitionProperty`. I hadn't totally thought that through as it's not performative. What you had was correct, adding that padding transition back into the progress tokens specifically. I'm working on rebasing this and will revert my suggestion shortly.

No worries, thanks for clarifying. Looking forward to seeing the updated version!

jcmill (approved):

mcoker (approved):
LGTM 👍 Thank you @tarunvashishth!!!

PR #8168, opened 2026-02-25
chore: add icon map for pfIcon helper

### LINKED ISSUE #8167: Add pfIcon mapping for FA icons
There is an icon map in https://github.com/patternfly/patternfly-react/pull/12245/changes#diff-01b2a12d43759bd4327b83ff7a4365c143970e618ef392644e38ef0fffd60966R10-R32 that provides the RH icon equivalent for existing icons.

Let's see if we can create something like that for the pfIcon helper

---

**Jira Issue:** [PF-3628](https://issues.redhat.com/browse/PF-3628)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8167

Updates icons in the banner component to test. You can reference the compass component demos to validate the existing `pfIcon` usage didn't break.

prompt to convert the object in [pfToRhIcons.mjs](https://github.com/patternfly/patternfly-react/blob/main/packages/react-icons/scripts/icons/pfToRhIcons.mjs) to something we can use in core

> convert each item in this object this way. Dasherize the key name and remove the nested object and replace it with the 'name' value. For example...
> If the item is
> AddCircleOIcon: { name: 'rh-ui-add-circle', icon: getIconData('rh-ui-add-circle') },
> The new item should be
> 'add-circle-o-icon': 'rh-ui-add-circle',

## Summary by CodeRabbit

* **New Features**
  * Introduced an icon aliasing system to standardize icon names and improve icon resolution and error reporting.

* **Refactor**
  * Updated Banner examples to use the new icon helper instead of inline icon markup, ensuring consistent icon rendering across the UI.


mcoker:
Closing this for now. The current mapping is more for consumers, not our internal use, so we would need to go through the list and make sure it's all correct, and would work for us. Not the quick-win I was hoping for, and may just serve to complicate things. We could re-open if we decide to create a proper mapping for internal use.

PR #8176, opened 2026-02-26
fix(icons): updated carets to rh micron

### LINKED ISSUE #8173: Update caret icons to rh-icons
We can probably update all uses of `fa-angle-right`, but search for "angle-right" to catch places we just pass the icon name and not the full class. Most likely this is for an expandable thing that expands down. If so, you'll likely replace `fa-angle-right` with `rh-microns-caret-down` and change the `rotate` from 90deg to 180deg. 

If it doesn't expand down, it may be different. Just reference figma for how it should look.

---

**Jira Issue:** [PF-3637](https://issues.redhat.com/browse/PF-3637)

thatblindgeye:
Closes #8173 

- Did not update DualListSelector or pagination implementations that have a single and double angle icon (I get an error trying to use a double rh-microns or rh-ui caret icon)
- Did not update the up/down icon for the TextInputGroup > "Search input group, match with navigable options" example
- Breadcrumb implementations can use review, especially in Wizard (screenshot below)
    

## Release Notes

* **Style**
  * Updated icons across multiple UI components for improved visual consistency.
  * Modified toggle and expand/collapse animation rotations for better visual feedback in accordions, cards, tables, navigation, and other expandable components.


src/patternfly/components/Wizard/wizard-toggle-separator.hbs
```diff
@@ -1 +1 @@
-<i class="fas fa-angle-right {{pfv}}wizard__toggle-separator" aria-hidden="true"></i>
+{{pfIcon "rh-microns-caret-right"}}
```
mcoker: Looks like we lost the `{{pfv}}wizard__toggle-separator` class. It isn't easy to add a class to the react icons in core - can you just add it to a span that wraps the icon for now?


```suggestion
<span class="{{pfv}}wizard__toggle-separator">
  {{pfIcon "rh-microns-caret-right"}}
</span>
```

mcoker (commented):
Here's the backstop report. LGTM other than one thing. 

PDF - https://drive.google.com/file/d/1BWodPy2BYFwgh9VcULmJ-8dWFMvkZIVL/view?usp=sharing
Full report (open 'html_report/index.html') - https://drive.google.com/file/d/1kKFJ8_g4PESlMjdi4YBwc59WbFnK-ZUO/view?usp=sharing

PR #8184, opened 2026-02-28
fix(breadcrumb): ensure inline icons are underlined (#7554)

### LINKED ISSUE #7554: Bug - breadcrumb - icons in breadcrumb are missing an underline
**Describe the problem**
This is an issue with 
- [ ] Patternfly 5  
- [X] Patternfly 6  
- [ ] other

When adding an icon to a breadcrumb the icon does not have an underline underneath it. When using an icon in the middle of a breadcrumb this makes it look like there are two links rather than one.

Add an icon in the middle of a breadcrumb. A [codesandbox](https://codesandbox.io/p/sandbox/billowing-hooks-x4lvqy?file=%2Findex.tsx) demonstrating the problem.

**Expected behavior**
Any elements contained in the breadcrumb should have the underline.

Sort of. There is a workaround identified by @DaoDaoNoCode:

> i tried to do it in a hacky way by removing the text-decoration and adding a bottom border, but i am not sure the spacing as well as the border width. also, i am not sure if it’s an anti-pattern because it needs such a lot of custom styles…

Another possibility identified by Jackson Lee:

> this hack might work better than border-bottom.. https://stackoverflow.com/a/48693294

See [Slack thread](https://redhat-internal.slack.com/archives/C04JMHKSD9C/p1747921836999049) for more background

**Screenshots**

Simple issue:
Image

In context:
Image

**What is your environment?**
 - OS: MacOS
 - Browser: Reproduced in Chrome, Safari, Firefox
 - Version: N/A

OpenShift AI, sprint end June 6

**Any other information?**
cc: @andrew-ronaldson @thatblindgeye 

---

**Jira Issue:** [PF-1843](https://issues.redhat.com/browse/PF-1843)

tarunvashishth:
Fixes #7554 

Screenshot 2026-03-01 at 12 27 59 AM

## Summary by CodeRabbit

* **Style**
  * Improved inline icon and text rendering within breadcrumb navigation elements to ensure proper alignment and display.

### DIFF
--- a/src/patternfly/components/Breadcrumb/breadcrumb.scss
+++ b/src/patternfly/components/Breadcrumb/breadcrumb.scss
@@ -86,6 +86,11 @@
   word-break: break-word;
   background-color: var(--#{$breadcrumb}__link--BackgroundColor);
 
+  > svg,
+  > i {
+    display: inline;
+  }
+
   &:is(:hover, :focus) {
     --#{$breadcrumb}__link--Color: var(--#{$breadcrumb}__link--hover--Color);
     --#{$breadcrumb}__link--TextDecorationLine: var(--#{$breadcrumb}__link--hover--TextDecorationLine);


mcoker:
@tarunvashishth thanks for the PR but doesn't look like this fixes it. 

Screenshot 2026-08-03 at 8 09 02 PM

I [replied back to the original issue](https://github.com/patternfly/patternfly/issues/7554#issuecomment-5173412707) with some more info, and to see if this is still an issue considering it was filed over a year ago. If it is, I'm going to encourage it be a local override in their app. Otherwise I think we'll need to talk with the design team about it and what a solution might look like since this could impact links and buttons anywhere.

Going to close this PR for now. We can resurrect it if it's back on our plate to fix.

PR #8196, opened 2026-03-03
fix(menu-toggle): Fix MenuToggle avatar misalignment with adjacent form controls outside flex layouts

### LINKED ISSUE #7478: Bug - Menu toggle - avatar alignment issues
Follow up from https://github.com/patternfly/patternfly/pull/7416

@srambach noticed that when you have a menu toggle with an avatar next to one without (or next to any form type of element - textinput, button, etc), the avatar menu toggle sits a couple of pixels below the other elements.

Image
Image

This doesn't appear to be the case when used in a toolbar (or flex layout), which is the main use case, but it's worth seeing if there may be a fix for that. cc @sg00dwin

tarunvashishth:
Fixes #7478 

Screenshot 2026-03-03 at 8 56 27 PM

## Summary by CodeRabbit

* **Style**
  * Improved visual alignment of the menu toggle component for better presentation.

### DIFF
--- a/src/patternfly/components/MenuToggle/menu-toggle.scss
+++ b/src/patternfly/components/MenuToggle/menu-toggle.scss
@@ -210,6 +210,7 @@
   font-size: var(--#{$menu-toggle}--FontSize);
   line-height: var(--#{$menu-toggle}--LineHeight);
   color: var(--#{$menu-toggle}--Color);
+  vertical-align: middle;
   cursor: pointer;
   background-color: var(--#{$menu-toggle}--BackgroundColor);
   border: 0;


mcoker (commented):
Thanks @tarunvashishth. Adding `vertical-align` to the menu toggle could cause toggles to shift in users' UIs, and considering how commonly used the menu toggle is, I think that might introduce more risk than benefit in this case. I'm going to go ahead and close this PR and and original issue since we haven't received any feedback about misalignment. Feel free to re-open or drop a comment somewhere it if you think we should revisit!

tarunvashishth:
Yeah, sure, should I delete the branch @mcoker ?

mcoker:
@tarunvashishth yep

PR #8205, opened 2026-03-05
fix(wizard): adjust z-index for nav shadow bug on mobile viewport

### LINKED ISSUE #7225: Bug -  Wizard - nav shadow bug on mobile viewport
re: https://github.com/patternfly/patternfly/pull/7218#issuecomment-2480008248

Ideally the expanded nav shadow should show up under the footer on mobile.

tarunvashishth:
Fixes #7225

When the navigation bar touches the footer, a shadow will be visible on screens with a width less than 576px.  This shadow will be hidden on screens wider than 576px.

Screenshot 2026-03-05 at 1 36 52 PM

Screenshot 2026-03-05 at 1 37 06 PM

## Summary by CodeRabbit

* **Style**
  * Optimized wizard component rendering on mobile devices with improved responsive layering and element stacking.
  * Enhanced footer positioning for better visibility across all screen sizes.

### DIFF
--- a/src/patternfly/components/Wizard/wizard.scss
+++ b/src/patternfly/components/Wizard/wizard.scss
@@ -136,7 +136,8 @@
   --#{$wizard}__toggle--m-expanded__toggle-icon--Rotate: 180deg;
 
   // Nav
-  --#{$wizard}__nav--ZIndex: var(--pf-t--global--z-index--sm);
+  --#{$wizard}__nav--ZIndex: var(--pf-t--global--z-index--xs);
+  --#{$wizard}__nav--mobile--ZIndex: var(--pf-t--global--z-index--md);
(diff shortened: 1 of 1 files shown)


jcmill (commented):
@tarunvashishth, I see what you're doing here. Good call on tackling the mobile size, but I think we could achieve this more simply by keeping `__nav` at `--pf-t--global--z-index--sm` and bumping up the `__footer` z-index token from `xs` to `sm`. Making them match. 

With both set the same z-index, the stacking order would handle where the shadow is displayed across all sizes. In other words, the `__nav` would naturally fall behind the `__footer` due to the DOM order.

tarunvashishth:
Thanks @jcmill for the review and the fix.

jcmill (approved):

mcoker (approved):
Nice! Visual regressions look good.

PR #8206, opened 2026-03-05
fix: updated with notfication badge with unified theme

### LINKED ISSUE #8131: Notification Badge
### Figma Link

https://www.figma.com/design/qRn3S225WwGHIdgGZX0LTC/Unified-Theme--Design-Tokens--Styles---Specs?node-id=9559-14253&p=f&m=dev

### Jira/GitHub Ticket Link

TBD

### Design Screenshot

image

Some of these updates may have been made as part of the form control token updates. need to verify

---

**Jira Issue:** [PF-3555](https://redhat.atlassian.net/browse/PF-3555)

dlabaj:
Updated with notification badge changes per issue #8131 

## Summary by CodeRabbit

* **Style**
  * Standardized button visuals and state appearance by updating border radius and border color tokens; no functional changes.
  * Minor whitespace and formatting cleanups.

* **Refactor**
  * Simplified notification badge icon rendering while preserving visual output and behavior.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -220,7 +220,7 @@
   --#{$button}--m-control--m-small--PaddingInlineStart: var(--pf-t--global--spacer--control--horizontal--compact);
 
   // Stateful
-  --#{$button}--m-stateful--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$button}--m-stateful--BorderRadius: var(--pf-t--global--border--radius--action--plain--default);
   --#{$button}--m-stateful--PaddingInlineStart: var(--pf-t--global--spacer--control--horizontal--default);
   --#{$button}--m-stateful--PaddingInlineEnd: var(--pf-t--global--spacer--control--horizontal--default);
   --#{$button}--m-stateful--m-small--PaddingInlineEnd: var(--pf-t--global--spacer--control--horizontal--compact);
@@ -228,7 +228,7 @@
 
   // Read
   --#{$button}--m-read--BackgroundColor: var(--pf-t--global--background--color--control--default);
-  --#{$button}--m-read--BorderColor: var(--pf-t--global--border--color--default);
+  --#{$button}--m-read--BorderColor: var(--pf-t--global--border--color--control--default);
   --#{$button}--m-read--hover--BackgroundColor: var(--pf-t--global--background--color--control--default);
   --#{$button}--m-read--hover--BorderColor: var(--pf-t--global--border--color--hover);
   --#{$button}--m-read--m-clicked--BackgroundColor: var(--pf-t--global--background--color--control--default);
@@ -688,7 +688,7 @@
     --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-stateful--m-small--PaddingInlineEnd);
     --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-stateful--m-small--PaddingInlineStart);
   }
-
+  
   // Read
   &.pf-m-read {
     --#{$button}--BackgroundColor: var(--#{$button}--m-read--BackgroundColor);
--- a/src/patternfly/components/Button/notification-badge.hbs
+++ b/src/patternfly/components/Button/notification-badge.hbs
@@ -10,14 +10,11 @@
   button--modifier=notification-badge--modifier
   button--count=notification-badge--count
   }}
-    <i class="
-      {{~#if notification-badge--IsTask}}
-        pf-v6-pficon-task
-      {{~else if notification-badge--IsAttention}}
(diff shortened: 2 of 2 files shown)


andrew-ronaldson (approved):
@nicolethoen I can add another issue for the plain variant if you want to keep these small

This PR looks good to me!

nicolethoen:
@andrew-ronaldson yes thank you!

src/patternfly/components/Button/button.scss
```diff
@@ -687,6 +687,12 @@
     --#{$button}--PaddingInlineEnd: var(--#{$button}--m-stateful--PaddingInlineEnd);
     --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-stateful--m-small--PaddingInlineEnd);
     --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-stateful--m-small--PaddingInlineStart);
+
+    &.pf-m-read,
+    &.pf-m-unread,
+    &.pf-m-attention {
+      --#{$button}--BorderRadius: var(--pf-t--global--border--radius--action--plain--default);
+    }
```
mcoker: Could you just add this to `&.pf-m-stateful`, or is there a reason that won't work? Stateful buttons are only used for the notification badge states as far as I'm aware.

Also however you end up doing it, can you create a button var for this style at the top of the CSS file and reference that button var in this block? Like if you applied it to `.pf-m-stateful`, add the var like this

```scss
// at the top of the file with the other vars
--#{$button}--m-stateful--BorderRadius: var(--pf-t--global--border--radius--action--plain--default);

// with the other stateful vars
&.pf-m-stateful {
  --#{$button}--BorderRadius: var(--#{$button}--m-stateful--BorderRadius);
}
```

src/patternfly/components/Button/button.scss
```diff
@@ -701,7 +707,7 @@
 
   // Unread
   &.pf-m-unread {
-    --#{$button}--Color: var(--#{$button}--m-unread--Color);
+   --#{$button}--Color: var(--#{$button}--m-unread--Color);
```
mcoker: Just a nit but can you add that space back?

mcoker (commented):
Nice! Just a couple of things.

src/patternfly/components/Button/notification-badge.hbs
```diff
@@ -10,14 +10,11 @@
...
-      {{~/if}}"
-      aria-hidden="true">
-    </i>
+    {{#if notification-badge--IsTask}}
+      {{pfIcon "rh-ui-task-fill"}}
+    {{else if notification-badge--IsAttention}}
+      {{pfIcon "rh-ui-attention-bell-fill"}}
+    {{else}}
+      {{pfIcon "rh-ui-notification-fill"}}
+    {{/if}}
```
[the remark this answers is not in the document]
mcoker: False. The raw SVGs contain `aria-hidden="true"` so this is a non-issue

andrew-ronaldson:
I added the plain variant [issue ](https://github.com/patternfly/patternfly/issues/8212)

src/patternfly/components/Button/button.scss
dlabaj: fixed. Move to just use stateful.

src/patternfly/components/Button/button.scss
dlabaj: Updated.

src/patternfly/components/Button/button.scss
```diff
@@ -220,15 +236,15 @@
   --#{$button}--m-control--m-small--PaddingInlineStart: var(--pf-t--global--spacer--control--horizontal--compact);
 
   // Stateful
-  --#{$button}--m-stateful--BorderRadius: var(--pf-t--global--border--radius--small);
+  --#{$button}--m-stateful--BorderRadius: var(--pf-t--global--border--radius--action--plain--default);
```
dlabaj, line 223: @mcoker Updated to use stateful per your suggestion in the review.

nicolethoen:
@dlabaj this needs a rebase

src/patternfly/components/Button/button.scss
mcoker: Are you able to remove `.pf-m-read`, `.pf-m-unread`, and `.pf-m-attention`? A button with `.pf-m-stateful` should always be used with one of those statuses, so just styling `.pf-m-stateful` should be all you need.

mcoker (changes requested):
This looks good but looks like you have a conflict. Good news is I think it's just from something prettier changed.

Can you undo the prettier changes? Sorry about that, I don't think any of the core devs use prettier so we don't get those updates. Related, I put up a PR to disable prettier here - https://github.com/patternfly/patternfly/pull/8257

mcoker (approved):
🚀

Issue #7226, opened 2024-11-19
Bug - Modal - update footer action spacing


mcoker:
https://github.com/patternfly/patternfly/blob/23f5d680d5ec78cb6f5929d03b22245740d2278e/src/patternfly/components/ModalBox/modal-box.scss#L76-L77

https://github.com/patternfly/patternfly/blob/23f5d680d5ec78cb6f5929d03b22245740d2278e/src/patternfly/components/ModalBox/modal-box.scss#L245-L251

Need to update to use the correct semantic spacer, and there is no need for a breakpoint based spacer change.

PR #8209, opened 2026-03-05
fix(modal-box): improve footer button layout for long labels

### LINKED ISSUE #7440: Bug - [Modal] - [Modals should adjust to min-content size]
**Describe the problem**
This is an issue with 

- [x ] Patternfly 5  

- [ x] Patternfly 6  

Modal components have predefined size variants with fixed max-width values. This can lead to layout issues in real-world usage.

For example, a project might select the small variant, which works fine in English. However, in other languages with longer translations (e.g. German or French), the footer buttons may overflow because the modal doesn't resize to accommodate the additional text length.
Problem

The fixed max-width doesn't account for varying content sizes due to localization, especially in the modal footer where actions are placed. This results in broken or awkward layouts.

Just put some long text for the buttons in the small Modal variant.

**Expected behavior**
Modal should stretch.

No, we workaround this by setting:

    .pf-v5-c-modal-box { min-inline-size: min-content; }

**Screenshots**

image

Anaconda installer

---

**Jira Issue:** [PF-1451](https://redhat.atlassian.net/browse/PF-1451)

tarunvashishth:
Fixes #7440 and #7226

Fixes layout issues when modal footer buttons have long labels.
Buttons now stack correctly when space is limited and maintain consistent widths, improving readability and alignment.

Screenshot 2026-03-05 at 11 23 52 PM

## Summary by CodeRabbit

* **Style**
  * Enhanced modal dialog footer layout with improved button spacing and responsive wrapping
  * Buttons now adapt better to smaller screens with consistent alignment


src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -238,18 +234,17 @@
 .#{$modal-box}__footer {
   display: flex;
   flex: 0 0 auto;
+  flex-wrap: wrap;
+  gap: var(--pf-t--global--spacer--sm);
```
jcmill: This one is a little tricky in terms of the tokens we need to use. Had to touch base with Design on this one. But lets define two new tokens in the root with the rest of the footer tokens:
`--#{$modal-box}__footer--RowGap: var(--pf-t--global--spacer--gap--group--vertical);`
`--#{$modal-box}__footer--ColumnGap: var(--pf-t--global--spacer--gap--action-to-action--default);`

And then we can define the row-gap and column-gap like the suggestion. The column gap will map to the `--pf-t--global--spacer--sm` you are using (which is correct). The row-gap just follows the patter we have in our action list.

```suggestion
  row-gap: var(--#{$modal-box}__footer--RowGap);
  column-gap: var(--#{$modal-box}__footer--ColumnGap);
```

src/patternfly/components/ModalBox/modal-box.scss
```diff
@@ -238,18 +234,17 @@
...
-  > .#{$button}:not(:last-child) {
-    margin-inline-end: var(--#{$modal-box}__footer--c-button--MarginInlineEnd);
-
-    @media screen and (min-width: $pf-v6-global--breakpoint--sm) {
-      --#{$modal-box}__footer--c-button--MarginInlineEnd: var(--#{$modal-box}__footer--c-button--sm--MarginInlineEnd);
-    }
+  // Allow buttons to stack vertically when there is not enough horizontal space
+  > .#{$button} {
+    flex-grow: 1;
+    white-space: normal;
```
jcmill: Lets remove these.

jcmill (commented):
This looks great. Just a couple of things.

jcmill:
@tarunvashishth, I feel like I'm on repeat, but we really do appreciate your work on these PRs (Really!). We got to talking and found that this solution, along with some others, ended up being a breaking change. So I went ahead and closed out the issue with some suggestions. Feel free to take a look at those if you're curious.

tarunvashishth:
@jcmill Thank you for your feedback. I’ll definitely take a look at the suggestions.

PR #8211, opened 2026-03-06
fix(form-control): add transition styles

### LINKED ISSUE #7429: Bug - NumberInput - isDisabled toggle animation out of sync
**Describe the problem**
When toggling the `isDisabled` state of `NumberInput` the buttons and input change background color at different times.

Might be related to `InputGroup` in general, but only checked `NumberInput`.

Toggle the `isDisabled` here and see how the 
https://codesandbox.io/p/sandbox/great-worker-dfw5fv

**Expected behavior**
The disable animation are sync between buttons and input

N/A

tarunvashishth:
Add transition to form-control so the input background and border changes animate in sync with the increment/decrement buttons when toggling isDisabled.

This resolves the visual delay where the buttons and input updated at different times.

https://github.com/user-attachments/assets/716b9cc0-1fe1-4750-9e49-35d3c41edab0

Fixes #7429

## Summary by CodeRabbit

* **Enhancements**
  * Improved form control animations by introducing configurable transition timing, duration, delay, and transition properties.
  * Updated related pseudo-element animations so they follow the same transition settings for smoother visual behavior.
* **Bug Fixes**
  * Refined the disabled form control state styling so it no longer overrides border style during transitions, ensuring consistent appearance while remaining non-interactive.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -332,6 +332,7 @@
   // Disabled
   --#{$button}--disabled--Color: var(--pf-t--global--text--color--on-disabled);
   --#{$button}--disabled--BackgroundColor: var(--pf-t--global--background--color--disabled--default);
+  --#{$button}--disabled--BackgroundSurface: var(--#{$button}--BackgroundSurface);
   --#{$button}--disabled--TextDecorationColor: currentcolor;
   --#{$button}--disabled--BorderColor: transparent;
   --#{$button}--disabled__icon--Color: var(--pf-t--global--icon--color--on-disabled);
@@ -587,6 +588,7 @@
       --#{$button}--hover--BackgroundColor: transparent;
       --#{$button}--m-clicked--BackgroundColor: transparent;
       --#{$button}--disabled--BackgroundColor: transparent;
+      --#{$button}--disabled--BackgroundSurface: none;
       --#{$button}--disabled--Color: var(--#{$button}--m-link--m-inline--disabled--Color);
       --#{$button}--disabled__icon--Color: var(--#{$button}--m-link--m-inline--disabled__icon--Color);
       --#{$button}--m-link--Color: var(--#{$button}--m-link--m-inline--Color);
@@ -748,6 +750,7 @@
     --#{$button}--disabled--Color: var(--#{$button}--m-plain--disabled--Color);
     --#{$button}--disabled__icon--Color: var(--#{$button}--m-plain--disabled__icon--Color);
     --#{$button}--disabled--BackgroundColor: var(--#{$button}--m-plain--disabled--BackgroundColor);
+    --#{$button}--disabled--BackgroundSurface: none;
     --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-plain--m-small--PaddingInlineEnd);
     --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-plain--m-small--PaddingInlineStart);
 
@@ -941,7 +944,7 @@
   &.pf-m-aria-disabled {
     color: var(--#{$button}--disabled--Color);
     text-decoration-color: var(--#{$button}--disabled--TextDecorationColor);
-    background: var(--#{$button}--disabled--BackgroundColor) var(--#{$button}--BackgroundSurface);
+    background: var(--#{$button}--disabled--BackgroundColor) var(--#{$button}--disabled--BackgroundSurface, none);
 
     &::after {
       border-color: transparent;
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -15,6 +15,7 @@
   --#{$button}--FontWeight: var(--pf-t--global--font--weight--body--default);
   --#{$button}--FontSize: var(--pf-t--global--font--size--body--default);
   --#{$button}--BackgroundColor: transparent;
+  --#{$button}--BackgroundSurface: radial-gradient(circle, transparent 1%, color-mix(in srgb, currentcolor 15%, transparent) 2%) center/15000% 15000%;
   --#{$button}--BorderColor: transparent;
   --#{$button}--BorderWidth: var(--pf-t--global--border--width--action--default);
   --#{$button}--BorderRadius: var(--pf-t--global--border--radius--pill);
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Button/button.scss
jcmill, line 1: We can attack this issue solely from the `form-control.scss` file by adding the same transition properties and variable that the buttons use. So this `button.scss` file can be reverted.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -18,6 +18,14 @@
...
   --#{$form-control}--BackgroundColor: var(--pf-t--global--background--color--control--default);
   --#{$form-control}--Width: 100%;
+  --#{$form-control}--TransitionDelay: 0s;
+  --#{$form-control}--TransitionTimingFunction: var(--pf-t--global--motion--timing-function--accelerate);
+  --#{$form-control}--TransitionDuration: var(--pf-t--global--motion--duration--fade--default);
+  --#{$form-control}--TransitionProperty: color, background-color, border-color;
+  --#{$form-control}__control--TransitionDelay: 0s;
+  --#{$form-control}__control--TransitionTimingFunction: auto;
+  --#{$form-control}__control--TransitionDuration: 0s;
+  --#{$form-control}__control--TransitionProperty: none;
```
jcmill: The `__control` transitions tokens and where they are applied below should be removed. The wrapper's transitions should be sufficient to achieve the desired transition effects.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -71,6 +79,7 @@
   // disabled
   --#{$form-control}--m-disabled--Color: var(--pf-t--global--text--color--on-disabled);
   --#{$form-control}--m-disabled--BackgroundColor: var(--pf-t--global--background--color--disabled--default);
+  --#{$form-control}--m-disabled--before--BorderColor: transparent;
```
jcmill: Let's remove this as it should be handled in the form-control transitions.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -130,6 +139,10 @@
   // Form control icon
   --#{$form-control}__icon--Color: var(--pf-t--global--icon--color--regular);
   --#{$form-control}__icon--FontSize: var(--pf-t--global--icon--size--font--body--default);
+  --#{$form-control}__icon--TransitionDelay: 0s;
+  --#{$form-control}__icon--TransitionTimingFunction: auto;
+  --#{$form-control}__icon--TransitionDuration: 0s;
+  --#{$form-control}__icon--TransitionProperty: none;
```
jcmill: Lets remove these and where they are applied. We might need to address the icons later as this is a little outside the scope of the issue and from the designs, it looks like we might need further direction.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -146,6 +159,10 @@
   --#{$form-control}__toggle-icon--PaddingInlineEnd: 0; // remove in breaking change
   --#{$form-control}__toggle-icon--Color: var(--pf-t--global--icon--color--regular);
   --#{$form-control}__toggle-icon--FontSize: var(--pf-t--global--icon--size--font--body--default); 
+  --#{$form-control}__toggle-icon--TransitionDelay: 0s;
+  --#{$form-control}__toggle-icon--TransitionTimingFunction: auto;
+  --#{$form-control}__toggle-icon--TransitionDuration: 0s;
+  --#{$form-control}__toggle-icon--TransitionProperty: none;
```
jcmill: Same as icon comment above. Lets remove these and where they are applied.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -161,13 +178,18 @@
...
+  transition-duration: var(--#{$form-control}--TransitionDuration);
+  transition-property: var(--#{$form-control}--TransitionProperty);
 
   &::before,
   &::after {
     position: absolute;
     inset: 0;
     pointer-events: none;
     content: "";
+    transition: inherit;
```
jcmill: We wont need this.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -286,8 +312,8 @@
     --#{$form-control}--Color: var(--#{$form-control}--m-disabled--Color);
     --#{$form-control}--m-placeholder--Color: var(--#{$form-control}--m-disabled--Color);
     --#{$form-control}__toggle-icon--Color: var(--#{$form-control}--m-disabled__toggle-icon--Color);
-    --#{$form-control}--before--BorderStyle: none;
-    --#{$form-control}--after--BorderStyle: none;
+    --#{$form-control}--before--BorderColor: var(--#{$form-control}--m-disabled--before--BorderColor);
+    --#{$form-control}--after--BorderColor: var(--#{$form-control}--m-disabled--after--BorderColor);
```
jcmill: Lets change these back to the `--BorderStyle: none;` and remove the associated tokens. We can probably keep this explicit for now.

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -467,4 +501,5 @@
   padding-block-start: var(--#{$form-control}__utilities--PaddingBlockStart);
   padding-inline-end: var(--#{$form-control}__utilities--PaddingInlineEnd);
   pointer-events: none;
+  transition: inherit;
```
jcmill: We could remove this as well.

mcoker (changes requested):
Thanks for making those updates! This looks great except for one thing. The borders (on `::before`/`::after`) are not syncing with the button.

You can probably just do something like this
* Add `transition: inherit;` to https://github.com/tarunvashishth/patternfly/blob/fa2a6f8558891baadbac3c479297becb8fc7874e/src/patternfly/components/FormControl/form-control.scss#L165-L171
* Remove the vars that set the border-styles to none when disabled https://github.com/tarunvashishth/patternfly/blob/fa2a6f8558891baadbac3c479297becb8fc7874e/src/patternfly/components/FormControl/form-control.scss#L289-L290
* In that same `.pf-m-disabled` block from ^, add `&:is(::before, ::after) { border-color: transparent; }`

src/patternfly/components/FormControl/form-control.scss
```diff
@@ -286,10 +295,13 @@
...
     --#{$form-control}__toggle-icon--Color: var(--#{$form-control}--m-disabled__toggle-icon--Color);
-    --#{$form-control}--before--BorderStyle: none;
-    --#{$form-control}--after--BorderStyle: none;
 
     cursor: not-allowed;
+
+    &::before,
+    &::after {
+      border-color: transparent;
+    }
```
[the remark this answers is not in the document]
mcoker: This variable is not in use anywhere in any component CSS, so enabling it at this point could cause unexpected UI changes, which I'd like to avoid. We'll remove that variable in our next major version release.

mcoker (approved):
Perfect, thank you! Visual regressions passed and number input looks good.

https://github.com/user-attachments/assets/10477e2b-a3d5-4c43-b552-7a10d30d797e

PR #8213, opened 2026-03-08
fix(empty-state): align alert title with centered empty state content

### LINKED ISSUE #7310: Bug - Empty State - alerts in an empty state are not centered correctly
Discovered during OCP + PF6 migration

You can see the misalignment in OCP:

Image

And here it is reproduced on the PF website:
 
Image

tarunvashishth:
Fixes #7310 

Screenshot 2026-03-08 at 10 56 19 PM

## Summary by CodeRabbit

* **Style**
  * Refined spacing and alignment of alert titles within empty states to improve visual consistency and readability.


jcmill:
@tarunvashishth, appreciate this. I had suggested one thing but in a conversation with Coker, we need to address some questions with design as there might need to be more significant changes to this component. I'll get back to you.

jcmill:
@tarunvashishth, we got feedback from design and ended up closing the issue. As it stands, we are not intending `empty-state` to support nested components right now. The intent is for other components be displayed outside of `empty-state`. Thank you for taking this on.

PR #8214, opened 2026-03-09
fix(table): align checkbox and action with data label in grid layout

### LINKED ISSUE #6966: table action/kebab position is too high/low

tarunvashishth:
Fixes #6966 

image

## Summary by CodeRabbit

* **Style**
  * Refined compact table spacing for responsive checkboxes, toggles, and action controls.
  * Improved alignment and visual consistency when cells follow selection controls.
  * Corrected action spacing to provide more balanced table layouts across responsive views.


mcoker:
Looks like the original fix addressed some of the spacing, but there are lots of table combos and some were still misaligned. Pushed some updates. Here are visual regressions of tests that match "components/table"

[backstop-8.5.26.pdf](https://github.com/user-attachments/files/30771906/backstop-8.5.26.pdf)
https://drive.google.com/file/d/1Mjq4W3cSWI3-9XHPcHrCsXVOhatCYUyx/view?usp=sharing

A breakdown of the changes:
* Updates the mobile "first-child" top padding var. This is the first visible data cell in the table - like the first row of text. Technically it isn't always the `:first-child` because the toggle or checkbox/radio is the actual first-child, and that was causing a bug in the current table implementation. If you look on org, the first row of text of a regular table gets the padding from the `cell--first-child` var, but the first row of text on a table with a check/radio or toggle doesn't. Now that first row should always get the correct padding. That also makes the space above the first row and below the last row the same. Also added a var for compact tables because compact tables look like they need a different spacer.
* Updates the check/radio top margin. This was a fixed value. Now it is the first row of text's top padding + an offset to center the checkbox/radio with the text and optional action to the right.
* Updates the action cell top margin. This was a fixed value. Now it's the first row of text's top padding minus the action's top padding, which visually centers the icon in the button with the row of text and optional checkbox/radio.

src/patternfly/components/Table/table-grid.scss
```diff
@@ -97,12 +97,13 @@
...
 
   // * Table grid compact table
+  --#{$table}--m-compact--cell--first-child--responsive--PaddingBlockStart: var(--pf-t--global--spacer--xs);
   --#{$table}--m-compact__tr--responsive--PaddingBlockStart: var(--pf-t--global--spacer--sm);
   --#{$table}--m-compact__tr--responsive--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
   --#{$table}--m-compact__tr__td--responsive--PaddingBlockStart: var(--pf-t--global--spacer--xs);
   --#{$table}--m-compact__tr__td--responsive--PaddingBlockEnd: var(--pf-t--global--spacer--xs);
-  --#{$table}--m-compact__check--responsive--MarginBlockStart: #{pf-size-prem(7px)};
-  --#{$table}--m-compact__action--responsive--MarginBlockStart: calc(var(--pf-t--global--spacer--xs) * -1);
+  --#{$table}--m-compact__check--responsive--MarginBlockStart: calc(var(--#{$table}--cell--first-child--responsive--PaddingBlockStart) +  #{pf-size-prem(4px)});
```
mcoker:
```suggestion
  --#{$table}--m-compact__check--responsive--MarginBlockStart: calc(var(--#{$table}--cell--first-child--responsive--PaddingBlockStart) +  #{pf-size-prem(4px)}); // cell padding plus an offset for the check/radio to align with text
```

thatblindgeye (approved):

jcmill (approved):
🏓

PR #8228, opened 2026-03-18
fix(clipboard-copy): update icons and verify tokens

### LINKED ISSUE #8186: Clipboard copy: Update icons and verify token changes
### Figma Link

https://www.figma.com/design/qRn3S225WwGHIdgGZX0LTC/Unified-Theme--Design-Tokens--Styles---Specs?node-id=9552-14259&p=f&m=dev

### Jira/GitHub Ticket Link

TBD

### Design Screenshot

image

*No response*

---

**Jira Issue:** [PF-3642](https://redhat.atlassian.net/browse/PF-3642)

jcmill:
## Summary
Update the Clipboard Copy component to use the correct copy icon and confirm token usage.

## What changed
- **Clipboard Copy icon:** Replaced the copy-button icon from `copy` to `rh-ui-copy-fill` across all Clipboard Copy examples so the component uses the intended design-system icon.
- **File updated:** `src/patternfly/components/ClipboardCopy/examples/ClipboardCopy.md` — all `button--icon="copy"` references are now `button--icon="rh-ui-copy-fill"` (basic, expandable, inline compact, and all variants).

Fixes #8186

## Verification
- Token usage for the Clipboard Copy component has been verified and is correct.

## Summary by CodeRabbit

* **New Features**
  * Added read-only state styling support for the ClipboardCopy component with updated visual styling for read-only expandable content.

* **Documentation**
  * Updated ClipboardCopy component examples with new icon names and documented the read-only state modifier and styling options.

### DIFF
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
@@ -1,6 +1,4 @@
-<div class="{{pfv}}clipboard-copy__expandable-content
-{{#if clipboard-copy-expandable-content--IsReadonlyExpandable}} pf-m-expandable-readonly{{/if}}
-{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div class="{{pfv}}clipboard-copy__expandable-content {{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
   {{#unless clipboard-copy--IsExpanded}}hidden{{/unless}}
   {{#if clipboard-copy-expandable-content--attribute}}
     {{{clipboard-copy-expandable-content--attribute}}}
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
@@ -1,4 +1,6 @@
-<div class="{{pfv}}clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div class="{{pfv}}clipboard-copy__expandable-content
+{{#if clipboard-copy-expandable-content--IsReadonlyExpandable}} pf-m-expandable-readonly{{/if}}
+{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
   {{#unless clipboard-copy--IsExpanded}}hidden{{/unless}}
   {{#if clipboard-copy-expandable-content--attribute}}
     {{{clipboard-copy-expandable-content--attribute}}}
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
@@ -1,4 +1,4 @@
-<div class="{{pfv}}clipboard-copy__expandable-content {{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div class="{{pfv}}clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
   {{#unless clipboard-copy--IsExpanded}}hidden{{/unless}}
   {{#if clipboard-copy-expandable-content--attribute}}
     {{{clipboard-copy-expandable-content--attribute}}}
--- a/src/patternfly/components/ClipboardCopy/clipboard-copy.scss
+++ b/src/patternfly/components/ClipboardCopy/clipboard-copy.scss
@@ -22,6 +22,10 @@
   --#{$clipboard-copy}__expandable-content--BorderRadius: var(--pf-t--global--border--radius--control--form-element);
   --#{$clipboard-copy}__expandable-content--OutlineOffset: var(--pf-t--global--spacer--xs);
 
+  // Readonly
+  --#{$clipboard-copy}--m-readonly__expandable-content--BackgroundColor: var(--pf-t--global--background--color--control--read-only);
+  --#{$clipboard-copy}--m-readonly__expandable-content--BorderColor: var(--pf-t--global--border--color--control--read-only);
+
   // Group
   --#{$clipboard-copy}__group--Gap: var(--pf-t--global--spacer--gap--control-to-control--default);
 
@@ -62,6 +66,11 @@
       display: inline-flex;
     }
   }
+
+  &.pf-m-readonly {
+    --#{$clipboard-copy}__expandable-content--BackgroundColor: var(--#{$clipboard-copy}--m-readonly__expandable-content--BackgroundColor);
+    --#{$clipboard-copy}__expandable-content--BorderColor: var(--#{$clipboard-copy}--m-readonly__expandable-content--BorderColor);
+  }
 }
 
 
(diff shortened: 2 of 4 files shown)


jcmill:
@bekah-stephens, quick question. I noticed the token used for the read-only background color was correct but the actually value was different from the figma file. Does that value need to be updated? It was the difference between something like a -200 and a -300.

bekah-stephens (changes requested):
For the _**default**_ state, the expanded area should use the default control tokens for the background and stroke colors.

Fill: _global/background/color/control/default_
Stroke: _global/border/color/control/default_

For the _**read-only**_ state, the expanded area should use the read-only control tokens for the background and stroke colors.

Fill: _global/background/color/control/read-only_
Stroke: _global/border/color/control/read-only_

Image

src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
```diff
@@ -1,4 +1,6 @@
-<div class="{{pfv}}clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div class="{{pfv}}clipboard-copy__expandable-content
+{{#if clipboard-copy-expandable-content--IsReadonlyExpandable}} pf-m-expandable-readonly{{/if}}
+{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
```
mcoker: FWIW we use a `setModifiers` handlebars thing for components that have multiple classes like this - the [main button handlebars](https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Button/button.hbs) file has a good example of it.

src/patternfly/components/ClipboardCopy/clipboard-copy.scss
```diff
@@ -96,6 +96,11 @@
   pre {
     white-space: pre-wrap;
   }
+
+  &.pf-m-expandable-readonly {
+    background-color: var(--pf-t--global--background--color--control--read-only);
+    border-color: var(--pf-t--global--border--color--control--read-only);
+  }
```
mcoker: Like we discussed, let's move this to `.pf-m-readonly` on `.pf-v6-c-clipboard-copy`, and set the `__expandable-content` styles by redeclaring the `--#{$clipboard-copy}__expandable-content--BackgroundColor` and `--#{$clipboard-copy}__expandable-content--BorderColor` vars

src/patternfly/components/ClipboardCopy/clipboard-copy-expandable-content.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="{{pfv}}clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
+<div class="{{pfv}}clipboard-copy__expandable-content {{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
```
mcoker: Doesn't really matter but you shouldn't need this extra space. Three is a space before `{{clipboard-copy-expandable-content--modifier}}` that will add a space if there are additional classes passed to `clipboard-copy-expandable-content--modifier`

```suggestion
<div class="{{pfv}}clipboard-copy__expandable-content{{#if clipboard-copy-expandable-content--modifier}} {{clipboard-copy-expandable-content--modifier}}{{/if}}"
```

src/patternfly/components/ClipboardCopy/examples/ClipboardCopy.md
mcoker: Can you also add the readonly class to this readonly example, too?

```suggestion
{{#> clipboard-copy clipboard-copy--id="expandable-not-expanded-readonly" clipboard-copy--IsReadonly="true"}}
```

mcoker (changes requested):
Feature changes look good! Can you make a couple of small updates to the docs?

src/patternfly/components/ClipboardCopy/examples/ClipboardCopy.md
```diff
@@ -267,4 +267,5 @@ Lorem ipsum dolor sit amet, consectetur adipiscing elit.&nbsp;
 | `.pf-m-truncate` | `.pf-v6-c-clipboard-copy.pf-m-truncate` | Modifies the inline copy clipboard component for use with text used in trucate component. |
 | `.pf-m-expanded` | `.pf-v6-c-clipboard-copy` | Modifies the clipboard copy component for the expanded state. |
 | `.pf-m-expanded` | `.pf-v6-c-button.pf-m-control` | Modifies the control toggle button for the expanded state. |
+| `.pf-m-readonly` | `.pf-v6-c-clipboard-copy` | Designates the clipboard copy component as read-only and modifies expandable content to match read-only background and border styles. |
```
mcoker: I might also move this line up to come after the line for `.pf-m-expanded` so that all of the modifiers for `.pf-v6-c-clipboard-copy` are next to one another in the table. 

I would probably also put something a little more generic like "Modifies the clipboard copy component for read-only styles" since I don't think it's required to call out the style differences explicitly. Also the more specific styles we list in the docs, the more there is to update if a style changes, and that's something that's easy to miss when making style updates.

src/patternfly/components/ClipboardCopy/examples/ClipboardCopy.md
```diff
@@ -266,5 +266,6 @@ Lorem ipsum dolor sit amet, consectetur adipiscing elit.&nbsp;
 | `.pf-m-block` | `.pf-v6-c-clipboard-copy.pf-m-inline` | Modifies the inline copy clipboard component to have block formatting. |
 | `.pf-m-truncate` | `.pf-v6-c-clipboard-copy.pf-m-truncate` | Modifies the inline copy clipboard component for use with text used in trucate component. |
 | `.pf-m-expanded` | `.pf-v6-c-clipboard-copy` | Modifies the clipboard copy component for the expanded state. |
+| `.pf-m-readonly` | `.pf-v6-c-clipboard-copy` | Modifies the clipboard copy component for read-only styles |
```
mcoker:
```suggestion
| `.pf-m-readonly` | `.pf-v6-c-clipboard-copy` | Modifies the clipboard copy component for read-only styles. |
```

mcoker (commented):
Just a small syntax thing, but I would add `clipboard-copy--IsReadonly="true"` to the non-expandable example, too, even though it doesn't do anything at the moment. In react, we have a [`isReadOnly` prop you apply to `<ClipboardCopy />`](https://www.patternfly.org/components/clipboard-copy#read-only) so I'm thinking this class goes on any read-only clipboard copy component.

bekah-stephens (changes requested):
It looks like there's something off with the collapsed editable field in dark mode (it doesn't match the editable expanded panel). It looks fine in light mode, though? (I marked the visually correct fields with green arrows)

Image

Image

bekah-stephens:
I'm not sure if it's easy for you to update this, but something small I noticed is that the content for the read-only example says "This is an editable version..." which is confusing on first glance. I think it would be good to make those examples say "This is a read-only version..."

image

jcmill:
@bekah-stephens, I did a rebase because the tokens looked wrong and that pretty much fixed everything. Also changed the copy for the read-only items.

bekah-stephens (approved):
lgtm!

src/patternfly/components/ClipboardCopy/examples/ClipboardCopy.md
mcoker: Can you add the readonly class to this, too? Not sure if you saw it but I mentioned updating this in my last review comment here - https://github.com/patternfly/patternfly/pull/8228#pullrequestreview-4022154024

```suggestion
{{#> clipboard-copy clipboard-copy--id="basic-readonly" clipboard-copy--IsReadonly="true"}}
```

mcoker (approved):
🚀

PR #8237, opened 2026-03-23
chore: update tokens from figma

mcoker:
## Summary by CodeRabbit

* **Chores**
  * Refreshed generation timestamps across all theme token files.
  * Reduced Red Hat theme token output to only variant-specific overrides (much smaller override sets).

* **Bug Fixes**
  * Adjusted dark-theme read-only control border color mapping to a darker token.
  * Tweaked high-contrast dark read-only control background to a lighter dark variant.
  * Removed one hover text-decoration offset token from the default theme.


lboehling (approved):
sweeeet

PR #8257, opened 2026-03-26
chore: disabled prettier formatting

mcoker:
Disables prettier. If you have prettier installed, it's causing updates like these, which we don't want - https://github.com/patternfly/patternfly/pull/8206/changes/cf79ab3e795e5b9b7d26ef55325c59a60d44161e

## Summary by CodeRabbit

* **Chores**
  * Switched project to rely on Prettier's default formatting by removing custom formatter settings.
  * Broadened formatter exclusion patterns so a catch‑all rule prevents matching files from being processed.


srambach (approved):
L 🥨 TM

.prettierignore
```diff
@@ -1,2 +1 @@
-package.json
-scripts/template/
+**
```
jcmill: Do we want to match every file in the current directory and explicitly match every file with an extension just to be safe (like .js, .json, .md and so on)? Might be redundant.

ex:
```
*
*.*
```

srambach (approved):
👍🏻

PR #8290, opened 2026-04-08
feat: added dynamic sticky header styles

### LINKED ISSUE #8223: Sticky table header on Glass
Implement sticky positioning for table headers as part of the unified Glass theme.

---

Extension of existing Table component

**Describe the feature**
Enable sticky table headers using Glass background styles. Ensure the header stays visible when scrolling content. Must support current responsive breakpoints and accessibility guidelines.

Design mockups available upon request or reference Glass theme documentation.

**Any other information?**
Sub-issue of Epic: Unified theme: Sticky elements on Glass.

---

**Jira Issue:** [PF-3718](https://redhat.atlassian.net/browse/PF-3718)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8223

[Default backstop report](https://drive.google.com/file/d/1yPVSSv3lgQXNvoEo94OErZYygGbiSc-C/view?usp=sharing)
[Glass backstop report](https://drive.google.com/file/d/1xeE2jm2l9s04Z68UBAlIRzc6giKRUlLi/view?usp=sharing)

https://pf-pr-8290.surge.sh/components/table#sticky-header
https://pf-pr-8290.surge.sh/components/table#sticky-header-with-base-and-stuck

Handlebars/examples updates
* Adds extra rows to some sticky header examples so they scroll and the header can actually be sticky
* Adds extra columns un-truncates some `<th>`s in sticky column examples scroll horizontally.
* Adds a new "Sticky header with base and stuck" example
* Updates icons in a table demo include that's used for sticky table demo

CSS
* Keeps `.pf-m-sticky-header` (the default/existing sticky behavior)
* Creates `.pf-m-sticky-header-base` and `.pf-m-sticky-header-stuck` for the new feature
  * `-base` applies the position/z-index stuff, basically everything but the stylized parts of a sticky header
  * `-stuck` applies all the stylized stuff like background, border, box shadow, etc. This is what should be shown once the user scrolls a little and the sticky header is floating above the table content
* Separates sticky visual styles out into `::before` and `::after`
  * `::before` - the background, border-radius, and box-shadow. Need to put at least the border-radius on a pseudo element since it won't work on a `<thead>` element. All of that sits behind the thead with a negative z-index.
  * `::after` - the border, which needs a positive z-index to sit on top of the thead.
* Updates all sticky headers to use new sticky background
* Adds glass styles to sticky header to make it rounded/shadow
* Adds style transition for "stuck" styles so those fade in (uses fade--default duration)
  *  Fading in the entire pseudo element using `opacity` instead of transitioning background-color, border, box-shadow for performance reasons 
* `.pf-m-nested-column-header` used to share vars/styles with sticky header - separated those from sticky header styles

## Summary by CodeRabbit

* **New Features**
  * Added sticky-header states (base & stuck), a no-wrap header option, and new table modifier mappings.

* **New Templates / Layout**
  * Added a default grid-style table template, a long-table variant, and expanded scrollable tables from 9 to 13 columns.

* **Style**
  * Improved sticky header visuals, transitions and CSS variables; glass-theme overrides; moved sticky styling to modifier-driven overlays.

* **Documentation**
  * Updated demos/examples, aria-label handling, usage notes, and example rows.

* **UX**
  * Switched icons to the platform icon helper and adjusted sticky column target index.

### DIFF
--- a/src/patternfly/components/Table/examples/Table.md
+++ b/src/patternfly/components/Table/examples/Table.md
@@ -3855,6 +3855,132 @@ There are a few ways this can be handled:
           2 days ago
         {{/table-td}}
       {{/table-tr}}
+
+      {{#> table-tr}}
+        {{#> table-td table-td--data-label="Repository name"}}
+          Repository 5
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Branches"}}
+          10
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Pull requests"}}
+          25
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Workspaces"}}
+          5
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Last commit"}}
+          2 days ago
+        {{/table-td}}
+      {{/table-tr}}
+
+      {{#> table-tr}}
+        {{#> table-td table-td--data-label="Repository name"}}
+          Repository 6
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Branches"}}
+          10
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Pull requests"}}
+          25
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Workspaces"}}
+          5
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Last commit"}}
+          2 days ago
+        {{/table-td}}
+      {{/table-tr}}
+
+      {{#> table-tr}}
+        {{#> table-td table-td--data-label="Repository name"}}
+          Repository 7
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Branches"}}
+          10
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Pull requests"}}
+          25
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Workspaces"}}
+          5
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Last commit"}}
+          2 days ago
+        {{/table-td}}
+      {{/table-tr}}
+
+      {{#> table-tr}}
+        {{#> table-td table-td--data-label="Repository name"}}
+          Repository 8
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Branches"}}
+          10
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Pull requests"}}
+          25
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Workspaces"}}
+          5
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Last commit"}}
+          2 days ago
+        {{/table-td}}
+      {{/table-tr}}
+
+      {{#> table-tr}}
+        {{#> table-td table-td--data-label="Repository name"}}
+          Repository 9
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Branches"}}
+          10
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Pull requests"}}
+          25
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Workspaces"}}
+          5
+        {{/table-td}}
+        {{#> table-td table-td--data-label="Last commit"}}
+          2 days ago
+        {{/table-td}}
+      {{/table-tr}}
+
+      {{#> table-tr}}
+        {{#> table-td table-td--data-label="Repository name"}}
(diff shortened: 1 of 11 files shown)


src/patternfly/demos/Table/table-simple-table.hbs
```diff
@@ -152,5 +152,137 @@
       {{/table-td}}
       {{> table-cell-action}}
     {{/table-tr}}
+
+    {{#if table-simple-table--IsLong}}
+      {{#> table-tr table-tr--index='7'}}
+        {{> table-cell-check}}
+        {{#> table-th table-th--data-label="Repository name"}}
```
thatblindgeye: CodeRabbit called it out, but we're using `th` elements for some of these rows, whereas some rows are using `td`. Can we update to use `td` for these new rows?

src/patternfly/components/Table/examples/Table.md
```diff
@@ -3762,101 +3762,24 @@ There are a few ways this can be handled:
...
-          25
-        {{/table-td}}
-        {{#> table-td table-td--data-label="Workspaces"}}
-          5
-        {{/table-td}}
-        {{#> table-td table-td--data-label="Last commit"}}
-          2 days ago
-        {{/table-td}}
-      {{/table-tr}}
+This example shows the use of `.pf-m-sticky-header-base` and `.pf-m-sticky-header-stuck`. `.pf-m-sticky-header-stuck` can be applied dynamically as a table has scrolled to only show sticky styles when the header is "stuck" and floating above the table content.
```
thatblindgeye, line 3774: Not sure if I'm understanding the use-case of using the plain "sticky header" implementation in the example above this one, vs using the -base and -stuck implementation here

thatblindgeye (commented):
Also not sure if we'd need to address the `@at-root` thing coderabbit comment on above?

src/patternfly/components/Table/examples/Table.md
jcmill: I "think" -stuck is meant to be handled with JS. -base is the establishing the sticky behavior but keeps the layers "invisible" until -stuck is applied... again, if I am reading this correctly.

src/patternfly/demos/Table/table-simple-table.hbs
jcmill: Jumping in on this one too. The th tags are applied to column headers only from what I can tell. All others are td within the row itself. Still new to the PF way, so I could be wrong, but to me this seems to be a semantic choice and helps at certain views.

src/patternfly/components/Table/examples/Table.md
mcoker: "sticky header" is the existing sticky styles. The new feature where the sticky styles only apply once an element is scrolled a little is achieved with the combination of `-base` and `-stuck`, where `-base` is used all the time and `-stuck` is applied (by JS) once you've scrolled a bit like @jcmill mentioned

Visually the core examples for "sticky header" and this new feature look the same since core doesn't have a mechanism to dynamically apply the `-stuck` class. I went back and forth on creating this example since it isn't going to illustrate the dynamic nature of the new feature, but created it anyways since 1) we have examples for most features and 2) it gives us something for visual regressions to flag is we break it. Do you think we should remove it? Or update the blurb wording?

mcoker:
@coderabbitai can you re-review the changes in this PR?

src/patternfly/components/Table/templates/table--scrollable--tbody--tr.hbs
```diff
@@ -4,6 +4,7 @@
         table--scrollable--th--modifier=table--scrollable--th--modifier--cell-1-modifier
         table--scrollable--th--content=cell-1--content
         table-th--IsStickyLeft=true
+        table-th--IsNoWrap=true
```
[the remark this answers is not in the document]
mcoker: Parameters in handlebars are passed on by default unless it's set/overwritten somewhere down the line - they doesn't have to be forwarded by default. Are you saying there is somewhere the `table-th--IsNoWrap` is being redefined/overwritten where it will not reach table-th.hbs?

src/patternfly/components/Table/examples/Table.md
thatblindgeye: Ah gotcha. Yeah some stuff is tough to tell in Core due to the static nature. It's probably fine keeping it in, especially to show where the classes need to be applied (outside the doc tables)

thatblindgeye (approved):

jcmill (approved):

PR #8293, opened 2026-04-08
chore(gulp): remove dev server rimraf

wise-king-sullyman:
Removes a rimraf that was added previously to address a docs framework bug.

## Summary by CodeRabbit

* **Chores**
  * Removed automatic cleanup step that deleted the public directory when starting the development server.


mcoker (approved):
🍗

PR #8303, opened 2026-04-13
feat: add compass panel features to panel component

### LINKED ISSUE #8055: Panel - update to support compass panel features
📝 Problem Statement
The Compass Panel is used as a flexible glass container in the Compass layout. To support glass in PatternFly outside of the Compass layout, we will need a globally available flexible glass container. To maintain a single source of truth, we're removing the Compass Panel in favor of using the Panel component. However, the standard Panel currently lacks several specific features required by existing implementations.

💡 Proposed Solution
Port the unique styling features from the Compass Panel into the standard Panel component.

🛠 Scope & Tasks
Ensure the following Compass Panel features exist in the Panel component. Note that some of these may already be available in the Panel. We just need to make sure these features work in the Panel for compass/glass use cases. 

- `.pf-m-no-border`: Logic to toggle border visibility.

- `.pf-m-no-padding`: Option to remove internal spacing.

- `.pf-m-full-height`: Utility to force height: 100%.

- `.pf-m-no-glass`: Removal of frosted/glass background effects.

- `.pf-m-pill`: Implementation of high border-radius for pill-shaped styling.

- `.pf-m-scrollable`: CSS to handle internal overflow scrolling.

- `.pf-m-glass`: opt-in for a glass container.

In glass theme, when used in a compass layout, panel should be glass by default with an opt-out.

✅ Definition of Done

- The features above are supported in the Panel component.

- Visual regression test: A Panel with these modifiers looks identical to the old Compass Panel. This can be achieved by testing the existing Compass demos with Panel instead of Compass Panel.

- Panel component examples and docs are updated with any new/changed features.

- No breaking changes to existing Panel implementations that don't use these flags.

---

**Jira Issue:** [PF-3416](https://redhat.atlassian.net/browse/PF-3416)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8055

Helpful links
* https://pf-pr-8303.surge.sh/ai/generative-uis/compass
* https://pf-pr-8303.surge.sh/components/panel

Visual regressions - [BackstopJS Report.pdf](https://github.com/user-attachments/files/26768270/BackstopJS.Report.pdf)

Things this PR does:
* Updates all compass-panels to the panel component across all of the compass demos. Also updates the documentation to reference glass panels instead of compass panels.
* Adds `.pf-m-no-border` to panel. Removes the border from a panel that has a border. Currently the only use case is a `.pf-m-glass` panel that shouldn't have a border, like the one that wraps the chatbot text box.
* Adds `.pf-m-scrollable-auto-height` to the panel. Unddoes the `max-height` that `.pf-m-scrollable` sets by default. That's needed when a panel is used as a scroll container. In a breaking change, we can remove this class and move the `max-height` to the `.pf-m-raised` variant.
* Adds `.pf-m-full-height` to the panel. 
* Adds `.pf-m-glass` to the panel. This only applies glass styles in glass theme. Otherwise you get normal panel styles. 
* Updates the default panel border-radius from 'sm' to 'md'. This matches card and other content boxes (matches the old compass-panel). This is seen as a bug fix and not a breaking change, should align with other components in PF better. 
  * Retains the 'sm' border-radius on the "raised" variant that should be used for menus/dropdowns/etc, since that's the border radius menus use.
* Updates the block padding for the "main-body" element from "md" to "lg". Similar to the border-radius change, this matches card and other content boxes, and matches the old compass panel. Shouldn't impact panels as menus, since you don't use "main-body" if you have a menu in a panel because you don't want any padding on the panel.
* Removes a couple of old global styles from patternfly.scss that were just bandaids before docs framework could give all of the glass styles.
* Updates the compass examples to remove panels and the hero component to match the [react compass examples[(https://pf-react-staging.patternfly.org/ai/generative-uis/compass)
* Removes the compass responsive bit that hides everything below 1200px. Workspace styles conflict with it, and responsive is being addressed next.

Known issue
* Hero background is broken, but that's not part of the panel/compass-panel updates. https://github.com/patternfly/patternfly/issues/8316

## Summary by CodeRabbit

* **New Features**
  * Added Panel modifiers: no-border, full-height, pill, scrollable-auto-height, plus expanded scrollable/raised/glass behavior.

* **Style**
  * Compass UI switched to the shared Panel structure (headers, sidebars, content); removed the Compass-specific panel and the screen-size warning overlay; updated panel layout, padding, and glass scoping; adjusted default panel radii and flex layout.

* **Documentation**
  * Updated Panel examples, usage table, demos, and example CSS to reflect new modifiers and markup.


src/patternfly/components/Panel/examples/Panel.md
```diff
@@ -168,8 +168,10 @@ cssPrefix: pf-v6-c-panel
...
-| `.pf-m-bordered` | `.pf-v6-c-panel` | Modifies the panel for bordered styles. |
-| `.pf-m-raised` | `.pf-v6-c-panel` | Modifies the panel for raised styles. |
-| `.pf-m-scrollable` | `.pf-v6-c-panel` | Modifies the panel for scrollable styles. |
+| `.pf-m-bordered` | `.pf-v6-c-panel` | Adds a border to the panel. |
+| `.pf-m-no-border` | `.pf-v6-c-panel` | Removes the border from a variant that has a default border. |
+| `.pf-m-raised` | `.pf-v6-c-panel` | Applies elevated styles for use when the panel appears on top of other elements. |
+| `.pf-m-scrollable` | `.pf-v6-c-panel` | Modifies the panel so the main content is scrollable. |
 | `.pf-m-secondary` | `.pf-v6-c-panel` | Modifies the panel for secondary styles. |
 | `.pf-m-pill` | `.pf-v6-c-panel` | Modifies the panel for pill border radius. |
+| `.pf-m-full-height` | `.pf-v6-c-panel` | Makes the panel fill the avaiable height. |
```
mcoker: Add `.pf-m-glass`

src/patternfly/components/Panel/panel.scss
```diff
@@ -121,6 +128,14 @@
 
     backdrop-filter: var(--#{$panel}--m-glass--BackdropFilter);
   }
+
+  &.pf-m-pill {
+    --#{$panel}--BorderRadius: var(--#{$panel}--m-pill--BorderRadius);
```
mcoker, line 146: Add `--#{$panel}--m-pill--BorderRadius` var

src/patternfly/components/Panel/panel.scss
```diff
@@ -112,14 +118,28 @@
...
+  }
+
+  &.pf-m-pill {
+    --#{$panel}--BorderRadius: var(--#{$panel}--m-pill--BorderRadius);
+  }
 
-    backdrop-filter: var(--#{$panel}--m-glass--BackdropFilter);
+  &.pf-m-full-height {
+    height: 100%;
   }
```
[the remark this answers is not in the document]
mcoker: You get "no padding" by omitting `.pf-v6-c-panel__main-body`, so that variant isn't necessary. And there is not a use case for `.pf-m-no-glass`. The other compass panel that had `.pf-m-no-glass` had "glass" style by default, so it needed a class to disable the glass styles. For this change, the panel component is only glass if you set `.pf-m-glass`. If you don't want glass, don't set `.pf-m-glass.`

mcoker:
@coderabbitai please review this PR again

mcoker:
@andrew-ronaldson @kmcfaul @lboehling should be ready for review. Tried to call out everything that changed in the issue description.

thatblindgeye (approved):

kmcfaul (changes requested):
We may want to make this example a full page example / give it a wrapping div with height - https://pf-pr-8303.surge.sh/components/panel/#scrollable-with-auto-height. It's currently showing scrollable items over other content beyond the example area. 

I also noticed a slight performance hit when scrolling in glass theme versus default theme, even without the dev tools open, but I'm not sure if that's just my computer or surge. It would likely be more of a general issue than something with panel either way. Does anyone else notice that? 

Everything else in the PR seems good to me.

kmcfaul (approved):

lboehling (approved):
lgtm!! ✨

PR #8305, opened 2026-04-13
feat(button): add plain variant

### LINKED ISSUE #8212: Notification badge: Add plain variant
Follow up to #8131 work done by @dlabaj 

This issue adds a plain variant of the stateful button. 

image

Design contact: @andrew-ronaldson

---

**Jira Issue:** [PF-3669](https://redhat.atlassian.net/browse/PF-3669)

jcmill:
Closes #8212

Adds plain variant to stateful buttons.

Changes
**button.scss:** For stateful buttons, PaddingInlineStart / PaddingInlineEnd now use --pf-t--global--spacer--action--horizontal--plain--default instead of the control horizontal default, so plain stateful controls match plain action padding in figma.

**Button.md:** Adds a Read Plain demo (read + clicked) next to the existing stateful examples, and extends the .pf-m-plain table description to include stateful use cases.

## Summary by CodeRabbit

* **Style**
  * Adjusted plain icon/button modifier order to change override precedence and refine spacing and hover/click visuals for plain and no‑padding stateful/read variants.

* **Documentation**
  * Added stateful "Plain" examples (Read, Unread, Attention — clicked/non-clicked) including variants with visible counts and screen‑reader labels; expanded plain-button usage guidance.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -673,51 +673,6 @@
     --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-control--m-small--PaddingInlineStart);
   }
 
-  // Stateful
-  &.pf-m-stateful {
-    --#{$button}--BorderRadius: var(--#{$button}--m-stateful--BorderRadius);
-    --#{$button}--PaddingInlineStart: var(--#{$button}--m-stateful--PaddingInlineStart);
-    --#{$button}--PaddingInlineEnd: var(--#{$button}--m-stateful--PaddingInlineEnd);
-    --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-stateful--m-small--PaddingInlineEnd);
-    --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-stateful--m-small--PaddingInlineStart);
-  }
-  
-  // Read
-  &.pf-m-read {
-    --#{$button}--BackgroundColor: var(--#{$button}--m-read--BackgroundColor);
-    --#{$button}--BorderColor: var(--#{$button}--m-read--BorderColor);
-    --#{$button}--hover--BackgroundColor: var(--#{$button}--m-read--hover--BackgroundColor);
-    --#{$button}--hover--BorderColor: var(--#{$button}--m-read--hover--BorderColor);
-    --#{$button}--m-clicked--BackgroundColor: var(--#{$button}--m-read--m-clicked--BackgroundColor);
-    --#{$button}--m-clicked--BorderColor: var(--#{$button}--m-read--m-clicked--BorderColor);
-  }
-
-  // Unread
-  &.pf-m-unread {
-    --#{$button}--Color: var(--#{$button}--m-unread--Color);
-    --#{$button}--BackgroundColor: var(--#{$button}--m-unread--BackgroundColor);
-    --#{$button}__icon--Color: var(--#{$button}--m-unread__icon--Color);
-    --#{$button}--hover--Color: var(--#{$button}--m-unread--hover--Color);
-    --#{$button}--hover--BackgroundColor: var(--#{$button}--m-unread--hover--BackgroundColor);
-    --#{$button}--hover__icon--Color: var(--#{$button}--m-unread--hover__icon--Color);
-    --#{$button}--m-clicked--Color: var(--#{$button}--m-unread--m-clicked--Color);
-    --#{$button}--m-clicked--BackgroundColor: var(--#{$button}--m-unread--m-clicked--BackgroundColor);
-    --#{$button}--m-clicked__icon--Color: var(--#{$button}--m-unread--m-clicked__icon--Color);
-  }
-
-  // Attention
-  &.pf-m-attention {
-    --#{$button}--Color: var(--#{$button}--m-attention--Color);
-    --#{$button}--BackgroundColor: var(--#{$button}--m-attention--BackgroundColor);
-    --#{$button}__icon--Color: var(--#{$button}--m-attention__icon--Color);
-    --#{$button}--hover--Color: var(--#{$button}--m-attention--hover--Color);
-    --#{$button}--hover--BackgroundColor: var(--#{$button}--m-attention--hover--BackgroundColor);
-    --#{$button}--hover__icon--Color: var(--#{$button}--m-attention--hover__icon--Color);
-    --#{$button}--m-clicked--Color: var(--#{$button}--m-attention--m-clicked--Color);
-    --#{$button}--m-clicked--BackgroundColor: var(--#{$button}--m-attention--m-clicked--BackgroundColor);
-    --#{$button}--m-clicked__icon--Color: var(--#{$button}--m-attention--m-clicked__icon--Color);
-  }
-
   // Icon buttons
   &.pf-m-plain {
     --#{$button}--BorderWidth: var(--#{$button}--m-plain--BorderWidth);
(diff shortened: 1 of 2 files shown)


mcoker:
@andrew-ronaldson can you confirm that the left/right padding for the notification badge (all variants) should use the plain (sm) spacer token? Looks like everything is using a small spacer in figma. IIRC I thought the plain variant would look like this compared to the default notfication badge

Screenshot 2026-04-13 at 6 56 16 PM

andrew-ronaldson:
All the stateful buttons have a sm spacer in Figma. I see the react and core examples have md spacers on org but that feels too wide.

mcoker:
> All the stateful buttons have a sm spacer in Figma. I see the react and core examples have md spacers on org but that feels too wide.

Looks like it was a change we made in v6, I'm guessing since they look like control buttons by default, they use the control button spacer. If we update the non-plain notification badge to small left/right spacers, wdyt about using the "plain" spacer token versus the "sm" spacer? Another way of asking is, if users customize the plain button spacer (or we do with some new theme or whatever), would we expect the non-plain notification badge spacers to update, too?

andrew-ronaldson:
I talked to Lucia about this one.
We can use the normal control--horizontal--default spacer for the default stateful button (the same as other control buttons). For the plain style it can be the control--horizontal--plain spacer.

mcoker:
@jcmill fwiw here are the updates I made for the screenshot referenced above. 

1. Move the [stateful/read/unread/attention selector blocks](https://github.com/patternfly/patternfly/blob/629b06d5ca2e856fad82757e0e7cf8347309db65/src/patternfly/components/Button/button.scss#L676-L719) so they come after `.pf-m-plain` instead of before
2. Add the following to the [plain modifier vars](https://github.com/patternfly/patternfly/blob/629b06d5ca2e856fad82757e0e7cf8347309db65/src/patternfly/components/Button/button.scss#L722-L744)

```scss
--#{$button}--m-stateful--PaddingInlineEnd: var(--#{$button}--m-plain--PaddingInlineEnd);
--#{$button}--m-stateful--PaddingInlineStart: var(--#{$button}--m-plain--PaddingInlineStart);
--#{$button}--m-read--hover--BackgroundColor: var(--#{$button}--m-plain--hover--BackgroundColor);
--#{$button}--m-read--m-clicked--BackgroundColor: var(--#{$button}--m-plain--m-clicked--BackgroundColor);
--#{$button}--m-read--BackgroundColor: var(--#{$button}--m-plain--BackgroundColor);
```

src/patternfly/components/Button/examples/Button.md
```diff
@@ -228,6 +228,18 @@ For when a plain/icon button is placed inline with text {{> button button--IsPla
 
 <br><br>
 
+<strong>Read Plain</strong>
+<br>
+{{#> button button--IsStateful=true button--IsRead=true button--IsPlain=true button--icon="rh-ui-notification-fill"}}
+  10 {{#> screen-reader}}items{{/screen-reader}}
```
mcoker: Instead of showing "read plain", let's show a "plain" and "plain with count" block that shows read/unread/attention for both of them. It may make sense as 2 new examples?

src/patternfly/components/Button/button.scss
```diff
@@ -761,6 +721,48 @@
...
+    --#{$button}--BorderRadius: var(--#{$button}--m-stateful--BorderRadius);
+    --#{$button}--PaddingInlineStart: var(--#{$button}--m-stateful--PaddingInlineStart);
+    --#{$button}--PaddingInlineEnd: var(--#{$button}--m-stateful--PaddingInlineEnd);
+    --#{$button}--m-small--PaddingInlineEnd: var(--#{$button}--m-stateful--m-small--PaddingInlineEnd);
+    --#{$button}--m-small--PaddingInlineStart: var(--#{$button}--m-stateful--m-small--PaddingInlineStart);
+  }
+  
+  // Read
+  &.pf-m-read {
+    --#{$button}--BorderColor: var(--#{$button}--m-read--BorderColor);
```
mcoker, line 736: There is a coderabbitai comment about it, but did you mean to remove the background color tokens? We'll need those so the control background color is still applied, and so it isn't a breaking change since customizations of those vars would no longer work.

mcoker (changes requested):

src/patternfly/components/Button/button.scss
jcmill: My mistake, I'll bring those back in.

src/patternfly/components/Button/examples/Button.md
```diff
@@ -249,6 +249,64 @@ For when a plain/icon button is placed inline with text {{> button button--IsPla
...
+{{/button}}
+
+{{#> button button--IsStateful=true button--IsRead=true button--IsPlain=true button--IsClicked=true button--icon="rh-ui-notification-fill"}}
+  10 {{#> screen-reader}}items{{/screen-reader}}
+{{/button}}
+
+<br>
+
+{{#> button button--IsStateful=true button--IsUnread=true button--IsPlain=true button--icon="rh-ui-notification-fill"}}
+  10 {{#> screen-reader}}items{{/screen-reader}}
```
mcoker:
```suggestion
  10 {{#> screen-reader}}unread items{{/screen-reader}}
```

src/patternfly/components/Button/examples/Button.md
```diff
@@ -249,6 +249,64 @@ For when a plain/icon button is placed inline with text {{> button button--IsPla
...
+{{/button}}
+
+{{#> button button--IsStateful=true button--IsUnread=true button--IsPlain=true button--IsClicked=true button--icon="rh-ui-notification-fill"}}
+  10 {{#> screen-reader}}items{{/screen-reader}}
```
mcoker:
```suggestion
  10 {{#> screen-reader}}unread items{{/screen-reader}}
```

src/patternfly/components/Button/examples/Button.md
```diff
@@ -249,6 +249,64 @@ For when a plain/icon button is placed inline with text {{> button button--IsPla
...
+{{/button}}
+
+<br>
+
+{{#> button button--IsStateful=true button--IsAttention=true button--IsPlain=true button--icon="rh-ui-attention-bell-fill"}}
+  10 {{#> screen-reader}}items{{/screen-reader}}
```
mcoker:
```suggestion
  10 {{#> screen-reader}}unread items, needs attention{{/screen-reader}}
```

src/patternfly/components/Button/examples/Button.md
```diff
@@ -249,6 +249,64 @@ For when a plain/icon button is placed inline with text {{> button button--IsPla
...
+{{/button}}
+
+{{#> button button--IsStateful=true button--IsAttention=true button--IsPlain=true button--IsClicked=true button--icon="rh-ui-attention-bell-fill"}}
+  10 {{#> screen-reader}}items{{/screen-reader}}
```
mcoker:
```suggestion
  10 {{#> screen-reader}}unread items, needs attention{{/screen-reader}}
```

mcoker (commented):
Just a few small docs updates to match the existing stateful example. Otherwise LGTM!

mcoker (commented):
Also noticed the a11y tests are failing because the [plain (without count)](https://pf-pr-8305.surge.sh/components/button#stateful) need a label or text - https://pf-pr-8305-a11y.surge.sh/

You could add a `button--aria-label` of something like "All items read" for read, "Unread items" for unread and "Unread items, needs attention" for needs attention.

src/patternfly/components/Button/examples/Button.md
jcmill: All coming in the next push. Thank you sir.

mcoker (approved):
🚀

Issue #8287, opened 2026-04-07
default glass theme background image update


patternfly-jira-sync:
Let's come up with an official strategy/recommendation for how to deliver the right background image for various themes.

The way it works now:

`patternfly.scss` has some styles (that probably shouldn't live there) that [set the backgrounds for the body element and the login component](https://github.com/patternfly/patternfly/blob/86abb1a0e8d5637e15fd27af4c70947b30347702/src/patternfly/patternfly.scss#L8-L24) for defualt/unified light/dark themes.
The compass component has its own [background image styles](https://github.com/patternfly/patternfly/blob/86abb1a0e8d5637e15fd27af4c70947b30347702/src/patternfly/components/Compass/compass.scss#L75), set by a var that the react component uses to customize the image.

Some thoughts around this:

- We found that with the main page background for glass, setting the background on the `.pf-v6-c-page` class didn't work well because users put drawers and banners and things outside of the page component that may not show the background image properly if it's only applied to the page component. Also there are cases where pages may need the background but aren't using the page component - the login page component is a good example of that. So we set the styles on the `<body>` element.

- The way it should probably work is the styles for the page/body, login, and compass go either on the body element or via something like the background image component. Probably can't attach a background to the login or compass components specifically because someone might have a drawer or banner or something that is technically outside of the login/compass component but visible on the page. We ran into this when we assigned the background images to the page component - if a drawer wraps the whole page component, an inline/static drawer panel wouldn't have a background behind it.

I'm thinking we could do one of a few things:

- Just include the CSS similar to how we are doing today (but move it out of patternfly.scss) and use vars so you can set a var to override the background image.

  * Downside is people may want to set the backgrounds their own way, and may not want to set them on `<body>` as we have and want to use their own element.

- The same as above, but don't include the backgrounds by default - have them as an opt-in (via setting a class) either for all of them or individually by setting something like `.pf-v6-theme-glass-background`, `.pf-v6-theme-glass-felt-background`, etc on whatever item they want the background to apply to.

- Provide snippets of CSS for people to assign to body or whatever wrapping element they want to add the background do. It would probably be similar to what we have now - https://github.com/patternfly/patternfly/blob/629b06d5ca2e856fad82757e0e7cf8347309db65/src/patternfly/patternfly.scss#L8-L24

[...]

PR #8319, opened 2026-04-16
chore: update glass/felt workspace background images

mcoker:
towards https://github.com/patternfly/patternfly/issues/8287.

This just updates the existing styles. For #8287 I think we need to come up with a strategy for delivering the background image. 
* Can we use the background image component?
* Should we provide snippets of CSS for setting the background image for various themes?
* Should compass support props for the background image, or should those be removed in favor of one of the first two bullets?

## Summary by CodeRabbit

* **Style**
  * Refreshed background imagery for glass and Red Hat themes with updated SVG assets
  * Improved theme scoping so background images apply consistently to the page and components
  * Updated compass component to rely on theme-driven background images (removed redundant inline wallpaper injection)


lboehling (approved):
nice!

kmcfaul (approved):

PR #8337, opened 2026-04-27
chore: update backstop screenshots

mcoker:
Failure report. This is glass, but it shows all of the failures from non-glass + glass dock styles.

All failures look expected to me
* new examples
* toggle group block padding updates
* full width toggle group
* table sticky column background updates
* fixed compass (local) example css so full page compass examples occupy height of page
* stateful plain buttons

[backstop-4.27.26.pdf](https://github.com/user-attachments/files/27137211/backstop-4.27.26.pdf)

## Summary by CodeRabbit

* **Style**
  * Scoped compass styling to the main page area and simplified visual borders by replacing overlay decorations with a dashed outline on compass variants.
* **Documentation**
  * Updated the Docked example to include an inline masthead section before dock content for clearer layout demonstration.


jcmill (approved):
Looks great.

PR #8356, opened 2026-04-30
feat(glass): backgrounds for glass and compass

### LINKED ISSUE #8287: default glass theme background image update

jcmill:
Closes: #8287

Implements a global background image strategy for glass themes.

## Background image handling:
  - Moved background images from component-level to the root (HTML element)
  - Created theme-specific background image tokens located in token-local theme specific files
  - Each glass theme token file overrides --pf-t--global--background--image--default with fallbacks for its theme-specific variant

## Compass component updates:
  - Removed component-specific background image handling
  - Added --pf-v6-c-compass--BackgroundColor with default secondary background color
  - In glass themes, Compass background becomes transparent so the HTML element's background image shows through
  - Restructured to use compass-container wrapper for improved layout control

## Summary by CodeRabbit

* **New Features**
  * Added a compass-container component as the required layout wrapper and simplified Compass wrapper markup.
  * Added local glass/felt theme token sets and a new glass-theme box-shadow CSS variable.

* **Style**
  * Compass now uses a background-color approach with adjusted height/structure for consistent layout.
  * Normalized global background-image behavior and removed legacy theme image overrides.

* **Documentation**
  * Updated Compass examples and usage guidance to reflect the new container pattern.


src/patternfly/components/Compass/compass.scss
mcoker:
```suggestion
    .#{$compass}__container > .#{$masthead} {
```

mcoker (approved):
🤖👍

PR #8362, opened 2026-05-01
chore: update backstop screenshots

mcoker:
Updates main and glass screenshots. Everything looks expected except it looks like there is a bug in pagination on mobile in glass. The sticky background isn't applying. You can see it in a bunch of examples, like `/components/table/html-demos/compact ` and `/components/table/html-demos/compact `

Screenshot 2026-05-01 at 1 47 51 PM

[glass.pdf](https://github.com/user-attachments/files/27284089/glass.pdf)
[BackstopJS Report.pdf](https://github.com/user-attachments/files/27284090/BackstopJS.Report.pdf)


jcmill (approved):
👀👍

PR #8364, opened 2026-05-01
chore(backstop): add uniq report ids, add all theme images

### LINKED ISSUE #8363: Add baseline screenshots for all of the themes
Needed to do this eventually. With the upcoming PR that changes all of the references of "redhat" and "unified" to "felt" (https://github.com/patternfly/patternfly/pull/8354), that PR will include moving the variable collection that was in a separate file in figma being moved to the main v6 component figma file. So as good a time as any to establish baseline screenshots - then we can run a report against each theme to make sure nothing broke.

---

**Jira Issue:** [PF-4102](https://redhat.atlassian.net/browse/PF-4102)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8363

"files changed" tab is unusable because of the number of image/file changes.

The first two commits are code changes, the rest are file changes I've reviewed with @jcmill

Commit 1 - https://github.com/patternfly/patternfly/pull/8364/changes/4728132967f5c77a36a8eae44aec260fde5bb2c6

* Adds a dynamic "id" for the report, which labels the HTML report it generates with a title at the top, so that if you run multiple reports, it labels each as "default dark", "felt dark glass", etc.
* Adds `backstop:test:all` and `backstop:approve:all` scripts in package.json to kick off and approve all tests with a single command.
  * For reference, took me about 42 minutes for all to run. If running it, you may want to pass `caffeinate` before the command if you intend on stepping away at any point so that your computer doesn't go to sleep, which will cause the active test to fail
  *  Also worth noting that github currently has a 2gb max limit per push, and updating screenshots for all themes creates more than a 2gb change, so you may want to break it up into multiple commits and push each commit so you don't hit the 2gb limit on a single push.

Commit 2 - https://github.com/patternfly/patternfly/pull/8364/changes/24046eda7379dfbb2787f26be9cb0419deaa2a61
* Updates the test/approve:all scripts to always run each test instead of only running the next test if the previous one exited cleanly (replaced `&&` with `;` between the tests) 

## New Features
* Backstop test outputs and reports are now organized by theme combination (felt vs default, with optional dark, glass, and high-contrast variants)
* Added new npm scripts (`backstop:test:all` and `backstop:approve:all`) to run all theme test variants sequentially


jcmill (approved):
🕵️‍♂️

PR #8384, opened 2026-05-08
fix(dock): prevent tab interaction on mobile

### LINKED ISSUE #8339: Bug - Docked page and compass - can tab to items in collapsed dock
On mobile when the docked nav is collapsed, you can still tab to items in the dock.

We can likely just add `visiblity` to the expand/collapse animation. We'll need `visiblity: visible/revert` immediately when going from collapsed -> expanded. When going from expanded -> collapsed, we need `visibility: hidden` that is delayed by the duration of the animation so it hides after it finishes collapsing.

---

**Jira Issue:** [PF-4049](https://redhat.atlassian.net/browse/PF-4049)

jcmill:
Fixes #8339 
Prevents keyboard navigation to items in collapsed dock on mobile.

Added `visibility: hidden` to collapsed dock with transition timing that:
- Shows items immediately when expanding
- Hides items after slide animation completes when collapsing

Changes applied to both Page and Compass components for consistency.

## Summary by CodeRabbit

* **Style**
  * Enhanced dock panel animations with improved visibility transitions when expanding and collapsing.
  * Refined transition timing across different screen sizes for consistent dock behavior throughout the interface.


mcoker (approved):
perfect!

PR #8412, opened 2026-05-18
fix(banner): link colors

### LINKED ISSUE #8403: Bug - Banner - text decoration on Inline links broken.
**Describe the problem**
This is an issue with 

- [ ] Patternfly 5  

- [x] Patternfly 6  

- [ ] other

When we updated the text-decoration styling on links this change applied to all components. Because the banner has specific background colors we need to override the text-decoration styles to inherit the text color tokens.

Ex: on-danger, on-success.

---

**Jira Issue:** [PF-4202](https://redhat.atlassian.net/browse/PF-4202)

jcmill:
Fixes #8403

Fixes link text decoration colors. Links and their underlines now inherit the banner's text color tokens (on-danger, on-success, etc.) instead of using default link colors.

## Summary by CodeRabbit

* **Style**
  * Improved banner link styling so `text-decoration-color` is now consistently applied, including correct behavior for hover states.
  * Added dedicated theming variables for banner-level and link-level text-decoration color, ensuring inline link-styled buttons match banner link appearance.

### DIFF
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -23,7 +23,7 @@
   --#{$banner}--link--TextDecoration: underline;
   --#{$banner}--link--hover--Color: var(--#{$banner}--Color);
   --#{$banner}--link--disabled--Color: var(--pf-t--global--text--color--disabled);
-  --#{$banner}--a--TextDecorationColor: currentColor;
+  --#{$banner}--a--TextDecorationColor: currentcolor;
   --#{$banner}--m-link--TextDecorationColor: var(--#{$banner}--a--TextDecorationColor);
   --#{$banner}--m-link--hover--TextDecorationColor: var(--#{$banner}--a--TextDecorationColor);
 
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -154,9 +154,6 @@
     border-radius: var(--#{$banner}--m-pill--BorderRadius);
   }
     
-    a {
-      color: var(--#{$banner}--link--Color);
-      text-decoration-line: var(--#{$banner}--link--TextDecoration);
   a {
     color: var(--#{$banner}--link--Color);
     text-decoration-line: var(--#{$banner}--link--TextDecoration);
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -157,7 +157,10 @@
     a {
       color: var(--#{$banner}--link--Color);
       text-decoration-line: var(--#{$banner}--link--TextDecoration);
-      text-decoration-color: var(--#{$banner}--a--TextDecorationColor);
+  a {
+    color: var(--#{$banner}--link--Color);
+    text-decoration-line: var(--#{$banner}--link--TextDecoration);
+    text-decoration-color: var(--#{$banner}--a--TextDecorationColor);
 
     &:hover:not(.pf-m-disabled) {
       --#{$banner}--link--Color: var(--#{$banner}--link--hover--Color);
--- a/src/patternfly/components/Banner/banner.scss
+++ b/src/patternfly/components/Banner/banner.scss
@@ -23,7 +23,7 @@
   --#{$banner}--link--TextDecoration: underline;
   --#{$banner}--link--hover--Color: var(--#{$banner}--Color);
   --#{$banner}--link--disabled--Color: var(--pf-t--global--text--color--disabled);
-  --#{$banner}--a--TextDecorationColor: currentcolor;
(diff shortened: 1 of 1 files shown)


thatblindgeye (approved):

andrew-ronaldson (approved):

src/patternfly/components/Banner/banner.scss
```diff
@@ -23,6 +23,9 @@
   --#{$banner}--link--TextDecoration: underline;
   --#{$banner}--link--hover--Color: var(--#{$banner}--Color);
   --#{$banner}--link--disabled--Color: var(--pf-t--global--text--color--disabled);
+  --#{$banner}--a--TextDecorationColor: currentcolor;
```
mcoker: Since we already have `--link`, is there a reason not to use that?

```suggestion
  --#{$banner}--link--TextDecorationColor: currentcolor;
```

src/patternfly/components/Banner/banner.scss
```diff
@@ -150,10 +153,11 @@
...
   }
-
-  a {
-    color: var(--#{$banner}--link--Color);
-    text-decoration-line: var(--#{$banner}--link--TextDecoration);
+    
+    a {
+      color: var(--#{$banner}--link--Color);
+      text-decoration-line: var(--#{$banner}--link--TextDecoration);
+      text-decoration-color: var(--#{$banner}--a--TextDecorationColor);
```
mcoker: This doesn't need to be indented

```suggestion
  a {
    color: var(--#{$banner}--link--Color);
    text-decoration-line: var(--#{$banner}--link--TextDecoration);
    text-decoration-color: var(--#{$banner}--a--TextDecorationColor);
```

src/patternfly/components/Banner/banner.scss
```diff
@@ -174,6 +178,11 @@
     --#{$button}--disabled--Color: var(--#{$banner}--link--disabled--Color);
 
     text-decoration-line: var(--#{$banner}--link--TextDecoration);
+    
+    &.pf-m-link {
+      --#{$button}--TextDecorationColor: var(--#{$banner}--m-link--TextDecorationColor);
+      --#{$button}--hover--TextDecorationColor: var(--#{$banner}--m-link--hover--TextDecorationColor);
+    }
```
mcoker: A few things about this block. 
* Looks like we have button tokens for the inline link `text-decoration-color` styles, so no need to write out the `.pf-m-link` selector block.
  * FWIW `.pf-m-inline` only applies to `.pf-m-link` in button anyways, so you could have omitted the `.pf-m-link` selector anyways.
* I wouldn't have named these vars `--#{$banner}--m-link`. That implies it will style `.pf-v6-c-banner.pf-m-link`. If you had kept those vars, I'd name them something annoying like `--#{$banner}--button-inline-link` or something. Naming component vars for another component can be tricky/wordy. But since the existing var overrides under  `.#{$button}.pf-m-inline` use the `--#{$banner}--link` vars to style the button, I would say just reuse the `--#{$banner}--link--TextDecorationColor`
* Looks like we're missing a style for inline link buttons with `.pf-m-clicked` - I'll open a separate issue for that. It's creating a bug with `:active`

```suggestion
    --#{$button}--m-link--m-inline--TextDecorationColor: var(--#{$banner}--link--TextDecorationColor);
    --#{$button}--m-link--m-inline--hover--TextDecorationColor: var(--#{$banner}--link--TextDecorationColor);
    
    text-decoration-line: var(--#{$banner}--link--TextDecoration);
 ```

**Review by mcoker — CHANGES_REQUESTED** [review]

File: src/patternfly/components/Banner/banner.scss [hunk 4]
```diff
@@ -150,10 +153,11 @@
   &.pf-m-pill {
     border-radius: var(--#{$banner}--m-pill--BorderRadius);
   }
-
+
```
mcoker:
```suggestion

```

src/patternfly/components/Banner/banner.scss
```diff
@@ -23,6 +23,9 @@
   --#{$banner}--link--TextDecoration: underline;
   --#{$banner}--link--hover--Color: var(--#{$banner}--Color);
   --#{$banner}--link--disabled--Color: var(--pf-t--global--text--color--disabled);
+  --#{$banner}--link--TextDecorationColor: currentcolor;
+  --#{$banner}--m-link--TextDecorationColor: var(--#{$banner}--a--TextDecorationColor);
+  --#{$banner}--m-link--hover--TextDecorationColor: var(--#{$banner}--a--TextDecorationColor);
```
mcoker: You can remove these now that they're no longer used.

mcoker (approved):
🦸‍♂️

PR #8415, opened 2026-05-18
chore(ci): group non @patternfly/patternfly-a11y dependency bumps

wise-king-sullyman:
## Summary by CodeRabbit

* **Chores**
  * Reorganized dependency grouping in update configuration to prepare for consolidated updates.
  * Added a dedicated rule to handle the PatternFly accessibility package separately, while keeping the automated updates disabled for now.

[Review Change Stack](https://app.coderabbit.ai/change-stack/patternfly/patternfly/pull/8415?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -20,7 +20,7 @@
       "enabled": false
     },
     {
-      "groupName": "non-pf-a11y-deps",
+      "groupName": "grouped-pf-deps",
       "datasources": ["npm"],
       "packageNames": [
         "@patternfly/patternfly-a11y",
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -20,6 +20,7 @@
       "enabled": false
     },
     {
+      "groupName": "non-pf-a11y-deps",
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -20,6 +20,7 @@
       "enabled": false
     },
     {
+      "groupName": "non-pf-a11y-deps",
```
mcoker: wdyt about a name like "grouped-pf-deps" or something less specific to the excluded package, in case we either yank another package out of the single renovate PR or add a new PF dep but don't want it in the PR?

mcoker (approved):
Just one question about the group name that doesn't really matter. Is there a benefit or standard to the way it's named currently?

.github/renovate.json
wise-king-sullyman: No benefit/standard to the current name, I just threw some words together, can change that rq.

mcoker (approved):

PR #8421, opened 2026-05-19
feat(avatar): color variant

### LINKED ISSUE #8116: Avatar - Add Initials and colorful variants
Follow up to [https://github.com/patternfly/patternfly-design/issues/1518](https://github.com/patternfly/patternfly-design/issues/1518)

These are two new variants for Avatar based on product requests. These variants add more visual interest and helps differentiate user profiles when an image is not specified.

1. User initials - Testing of the smallest avatars showed that the users First initials would be visible. This users the body/small/semibold token.
2. Generic icon with backgrounds. Using our non-status background colors to add more visual interest to our default avatars. 

image

Figma link with devolpment annotations found [here](https://www.figma.com/design/qRn3S225WwGHIdgGZX0LTC/Unified-Theme--Design-Tokens--Styles---Specs?m=auto&node-id=9552-14252&t=GbSxS8vdEiwk05wc-1).

##

---

**Jira Issue:** [PF-3536](https://redhat.atlassian.net/browse/PF-3536)

jcmill:
Closes #8116

Add new color modifiers and initials to a colorful version of the Avatar component. Keeps existing Avatar usage as is.

New features:
  - User initials variant - Display initial with size-appropriate typography tokens
  - Colorful backgrounds - Nine color modifiers (red, orangered, orange, yellow, green, teal, blue, purple, gray) for both icon and initials variants
  - Accessibility with role="img", aria-label, and aria-hidden attributes
  - Full responsive size support (sm, md, lg, xl)

## Summary by CodeRabbit

* **New Features**
  * Added a "colorful" avatar variant with multiple color modifiers, initials support, SVG fallback, and size-specific typography for sm/lg/xl.

* **Documentation**
  * Updated examples to demonstrate colorful avatars (images and initials) and expanded accessibility and usage guidance.

* **Bug Fixes**
  * Improved alt attribute fallback and image source handling to enhance accessibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/patternfly/patternfly/pull/8421?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/patternfly/components/Avatar/avatar--colorful.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="{{pfv}}avatar {{pfv}}avatar--colorful
+    {{~#if avatar--modifier}} {{avatar--modifier}}{{/if}}"
+    role="img"
+    aria-label="{{#if avatar--AriaLabel}}{{avatar--AriaLabel}}{{else}}Avatar{{/if}}">
```
thatblindgeye: Could we instead have an if else block where if avatar--AriaLabel is passed render with aria-label, otherwise render with aria-hidden="true"?

If an aria-label/alt text isn't being supplied the avatar is likely just decorative or there's other context near it that provides relevant info (like an avatar next to a username) so we should hide it.

We should really update the base avatar hbs so that if avatar--AltText isn't passed in we just render an empty string for a similar result, but that doesnt have to be done in here.

thatblindgeye (changes requested):

src/patternfly/components/Avatar/avatar--colorful.hbs
jcmill: Both incoming!

src/patternfly/components/Avatar/avatar.hbs
```diff
@@ -1,6 +1,6 @@
 <img class="{{pfv}}avatar
   {{~#if avatar--modifier}} {{avatar--modifier}}{{/if}}"
-  alt="{{#if avatar--AltText}}{{avatar--AltText}}{{else}}Avatar image{{/if}}"
+  alt="{{#if avatar--AltText}}{{avatar--AltText}}{{else}}{{/if}}"
```
mcoker: Just a nit but typically for partial parameters that take text or a value, we use lower case, so it would be like `avatar--alt-text` (or you could even use `avatar--alt`). Usually sentence case is for boolean things like `IsExpanded` or `HasPadding`.

src/patternfly/components/Avatar/avatar.hbs
```diff
@@ -1,6 +1,6 @@
...
   src="
     {{~#if avatar--Src}}{{avatar--Src}}
```
mcoker: Same comment here about case, this would normally be `avatar--src`

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +96,102 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+    --#{$avatar}__initials--FontWeight: var(--#{$avatar}--m-lg__initials--FontWeight);
+    --#{$avatar}__initials--LineHeight: var(--#{$avatar}--m-lg__initials--LineHeight);
+  }
+
+  &.pf-m-xl {
+    --#{$avatar}__initials--FontSize: var(--#{$avatar}--m-xl__initials--FontSize);
+    --#{$avatar}__initials--LineHeight: var(--#{$avatar}--m-xl__initials--LineHeight);
+  }
+
+  &.#{$avatar}--colorful {
```
mcoker: If we used traditional BEM, this would be correct, but this should be `.pf-m-colorful`

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +96,102 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+    --#{$avatar}--BorderWidth: var(--#{$avatar}--colorful--BorderWidth);
```
mcoker: After updating to `.pf-m-colorful`, just a note to update all of the vars

```suggestion
    --#{$avatar}--BorderWidth: var(--#{$avatar}--m-colorful--BorderWidth);
```

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +96,102 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+
+    overflow: clip;
+    font-size: var(--#{$avatar}__initials--FontSize);
+    font-weight: var(--#{$avatar}__initials--FontWeight);
+    line-height: var(--#{$avatar}__initials--LineHeight);
```
mcoker: Looks like this is set on both color and on the `__initials` element, and it probably only needs to be on one since both point to the same vars. However I wonder if this should be setup so that the colorful variant sets the font-size to whatever the height/width is, which will make icons in it as big as the avatar as long as their height/width is set to `1em` (all of our icons are). That would also work for webfonts. Then the `__initials` element can set the font-size and font-weight appropriate for the letters.

I'm not sure that line-height is needed since this is a centered flex element? Didn't seem to do anything when I removed it.

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +96,102 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+    vertical-align: middle;
+    background-color: var(--#{$avatar}--BackgroundColor);
+    border-color: var(--#{$avatar}--BorderColor);
+    
+    .#{$avatar}__initials {
```
mcoker: You shouldn't need to nest this under the colorful modifier

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +96,102 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+      font-size: var(--#{$avatar}__initials--FontSize);
+      font-weight: var(--#{$avatar}__initials--FontWeight);
+      line-height: var(--#{$avatar}__initials--LineHeight);
+      color: var(--#{$avatar}--Color, var(--#{$avatar}--Color));
```
mcoker: Is this meant to fall back to itself?

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +96,102 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+    }
+
+    &.pf-m-red {
+      --#{$avatar}--BackgroundColor: var(--#{$avatar}--m-red--BackgroundColor);
+      --#{$avatar}--Color: var(--#{$avatar}--m-red--Color);
+      --#{$avatar}--FillColor: var(--#{$avatar}--m-red--FillColor);
```
mcoker: It looks like `--FillColor` sets the color for an icon and `--Color` sets the color for text? If so, I'm wondering if we can always set the `color` property and let icons use `fill="currentcolor"` to get the correct color? That would work better with our icons since our icons use `fill="currentcolor"` and it would also work if someone is using a webfont like font awesome. You could set the `color` of the `.pf-v6-c-avatar` element to the `--FillColor` var values (assuming an icon is a direct child of the avatar class), and set `color` on `.pf-v6-c-avatar__initials` to set the text color.

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -21,9 +21,65 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+  --#{$avatar}--m-lg__initials--FontWeight: var(--pf-t--global--font--weight--heading--bold);
+  --#{$avatar}--m-lg__initials--LineHeight: var(--pf-t--global--font--line-height--heading);
+  --#{$avatar}--m-xl__initials--FontSize: var(--pf-t--global--font--size--4xl);
+  --#{$avatar}--m-xl__initials--LineHeight: var(--pf-t--global--font--line-height--heading);
 }
 
 .#{$avatar} {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
```
mcoker: I wonder if we should limit this to `.pf-m-colorful` so it doesn't inadvertently change someone's existing use of avatar?

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -21,9 +21,65 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+  --#{$avatar}__initials--FontSize: var(--pf-t--global--font--size--body--lg);
+  --#{$avatar}__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$avatar}__initials--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$avatar}--m-sm__initials--FontSize: var(--pf-t--global--font--size--body--sm);
+  --#{$avatar}--m-sm__initials--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$avatar}--m-lg__initials--FontSize: var(--pf-t--global--font--size--2xl);
+  --#{$avatar}--m-lg__initials--FontWeight: var(--pf-t--global--font--weight--heading--bold);
+  --#{$avatar}--m-lg__initials--LineHeight: var(--pf-t--global--font--line-height--heading);
+  --#{$avatar}--m-xl__initials--FontSize: var(--pf-t--global--font--size--4xl);
+  --#{$avatar}--m-xl__initials--LineHeight: var(--pf-t--global--font--line-height--heading);
```
mcoker: Looks like the xl variant needs a font-weight of `var(--pf-t--global--font--weight--heading--default)`

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -21,9 +21,65 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+  --#{$avatar}--m-purple--BorderColor: var(--pf-t--global--border--color--nonstatus--purple--default);
+  --#{$avatar}--m-purple--FillColor: var(--pf-t--global--icon--color--nonstatus--on-purple--default);
+
+  // Initials
+  --#{$avatar}__initials--FontSize: var(--pf-t--global--font--size--body--lg);
+  --#{$avatar}__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$avatar}__initials--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$avatar}--m-sm__initials--FontSize: var(--pf-t--global--font--size--body--sm);
+  --#{$avatar}--m-sm__initials--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$avatar}--m-lg__initials--FontSize: var(--pf-t--global--font--size--2xl);
```
mcoker, line 65: Looks like the lg and xl variants need `font-family` set to `--pf-t--global--font--family--heading`. 

A couple of relevant notes about the fonts/weights
* The `--heading` font (aka "Red Hat Display" or RHD) is intended to be used for larger text. The `--body` font (aka "Red Hat Text" or RHT) RHT doesn't display well above a certain size. I'm not sure at what point we switch to that font, but that's generally why it's used. 
* The `--heading` font-weights should always be paired with the `--heading` font-family. I'm pretty sure the weights don't match between the fonts. And ranges for the `@font-face` rules are different. The RHT font is setup to use a range of [400-500 for the font-weights](https://github.com/patternfly/patternfly/blob/15eec9458136bc57eccc20079f3a7e7ef1173c79/src/patternfly/base/patternfly-fonts.scss#L6). Since `--pf-t--global--font--weight--heading--bold` is "700", it doesn't actually use 700 since it's out of range.

src/patternfly/components/Avatar/avatar.scss
jcmill: Honestly, I can't remember why. It was probably leftovers from an attempt at something.

src/patternfly/components/Avatar/avatar.scss
jcmill: Old habit here.

andrew-ronaldson:
What's the no-color variant?
Screenshot 2026-05-28 at 1 31 36 PM

jcmill:
@andrew-ronaldson That is the default styling of the colorful variants without any color modifiers applied. I can remove those if we don't want them being displayed. On a side note, I just removed the sizing modifiers.

thatblindgeye (approved):

andrew-ronaldson:
> @andrew-ronaldson That is the default styling of the colorful variants without any color modifiers applied. I can remove those if we don't want them being displayed. On a side note, I just removed the sizing modifiers.

I think we can remove them, please.

src/patternfly/components/Avatar/examples/Avatar.md
```diff
@@ -42,14 +76,28 @@ The avatar component provides a default SVG icon. If an image is used it should
 ### Accessibility
 | Attribute | Applied to | Outcome |
 | -- | -- | -- |
-| `alt` | `.pf-v6-c-avatar` | The alt attribute describes the appearance and function of the avatar image. **Required** |
+| `alt` | `.pf-v6-c-avatar` | The alt attribute describes the appearance and function of the avatar image. **Required for image avatars** |
+| `role="img"` | `<div class="pf-v6-c-avatar">` | Indicates the element represents an image. **Required for colorful avatars** |
+| `aria-label` | `<div class="pf-v6-c-avatar">` | Provides an accessible name for the avatar. **Required for colorful avatars** |
```
mcoker: You can just list the class

```suggestion
| `role="img"` | `.pf-v6-c-avatar` | Indicates the element represents an image. **Required for colorful avatars** |
| `aria-label` | `.pf-v6-c-avatar` | Provides an accessible name for the avatar. **Required for colorful avatars** |
```

If you need to call out the element (which you shouldn't here), you can do it like `div.pf-v6-c-avatar`.  You can see some examples in the button docs - https://github.com/patternfly/patternfly/blob/3707ce18e801b2760808368f8f871bd35eb714ea/src/patternfly/components/Button/examples/Button.md?plain=1#L348-L351

src/patternfly/components/Avatar/examples/Avatar.md
```diff
@@ -42,14 +76,28 @@ The avatar component provides a default SVG icon. If an image is used it should
...
+| `aria-hidden="true"` | `.pf-v6-c-avatar__initials`, `<svg>` | Hides decorative content from screen readers when the container has `aria-label`. |
 
 ### Usage
 | Class | Applied to | Outcome |
 | -- | -- | -- |
-| `.pf-v6-c-avatar` | `<img>` |  Initiates an avatar image. **Required** |
+| `.pf-v6-c-avatar` | `<img>`, `<div>` |  Initiates an avatar. **Required** |
+| `.pf-v6-c-avatar--colorful` | `<div class="pf-v6-c-avatar">` | Modifies an avatar to use colorful styling with a border. |
```
mcoker:
```suggestion
| `.pf-v6-c-avatar--colorful` | `.pf-v6-c-avatar` | Modifies an avatar to use colorful styling with a border. |
```

src/patternfly/components/Avatar/avatar--colorful.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="{{pfv}}avatar pf-m-colorful
+    {{~#if avatar--modifier}} {{avatar--modifier}}{{/if}}"
+    role="img"
+    {{#if avatar--AriaLabel}}aria-label="{{avatar--AriaLabel}}"{{else}}aria-hidden="true"{{/if}}>
```
mcoker: It's no big deal but most of the time, the only sentence case text we use in handlebars parameters is the `IsSomething` or `HasSomething` booleans.

```suggestion
    {{#if avatar--aria-label}}aria-label="{{avatar--aria-label}}"{{else}}aria-hidden="true"{{/if}}>
```

Here's an example of button, though button also has `button--AriaExpanded` which I would consider wrong https://github.com/patternfly/patternfly/blob/3707ce18e801b2760808368f8f871bd35eb714ea/src/patternfly/components/Button/button.hbs#L57-L69

src/patternfly/components/Avatar/avatar--colorful.hbs
```diff
@@ -0,0 +1,14 @@
...
+    {{#if avatar--Initials}}
```
mcoker: Same deal here, should be lower case

```suggestion
    {{#if avatar--initials}}
```

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +88,95 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+
+  &.pf-m-xl {
+    --#{$avatar}__initials--FontSize: var(--#{$avatar}--m-xl__initials--FontSize);
+    --#{$avatar}__initials--FontWeight: var(--#{$avatar}--m-xl__initials--FontWeight);
+    --#{$avatar}__initials--FontFamily: var(--#{$avatar}--m-xl__initials--FontFamily);
+  }
+
+  &.pf-m-colorful {
+    --#{$avatar}--BorderWidth: var(--#{$avatar}--m-colorful--BorderWidth);
+    --#{$avatar}--m-colorful--Color: var(--pf-t--global--color--nonstatus--gray--hover);
```
mcoker: This should be up in the top level scoped vars, though it's worth noting this var is already declared here - https://github.com/jcmill/patternfly/blob/f88a89d96094c376accbb560f9733456548ebf42/src/patternfly/components/Avatar/avatar.scss#L29

I'm assuming you'll want to replace that var with this one though since this one is what's being used.

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +88,95 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    overflow: clip;
```
mcoker, line 127: I like this! I would typically use `overflow: hidden` out of habit but `clip` is more appropriate here.

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -21,6 +21,54 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+  --#{$avatar}__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$avatar}--m-sm__initials--FontSize: var(--pf-t--global--font--size--body--sm);
+  --#{$avatar}--m-sm__initials--LineHeight: var(--pf-t--global--font--line-height--body);
+  --#{$avatar}--m-sm__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$avatar}--m-lg__initials--FontSize: var(--pf-t--global--font--size--2xl);
+  --#{$avatar}--m-lg__initials--FontWeight: var(--pf-t--global--font--weight--heading--bold);
+  --#{$avatar}--m-lg__initials--LineHeight: var(--pf-t--global--font--line-height--heading);
+  --#{$avatar}--m-lg__initials--FontFamily: var(--pf-t--global--font--family--heading);
+  --#{$avatar}--m-xl__initials--FontSize: var(--pf-t--global--font--size--4xl);
+  --#{$avatar}--m-xl__initials--LineHeight: var(--pf-t--global--font--line-height--heading);
```
mcoker: Doesn't look like the line-height vars are still used

```suggestion
  --#{$avatar}--m-sm__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
  --#{$avatar}--m-lg__initials--FontSize: var(--pf-t--global--font--size--2xl);
  --#{$avatar}--m-lg__initials--FontWeight: var(--pf-t--global--font--weight--heading--bold);
  --#{$avatar}--m-lg__initials--FontFamily: var(--pf-t--global--font--family--heading);
  --#{$avatar}--m-xl__initials--FontSize: var(--pf-t--global--font--size--4xl);
```

src/patternfly/components/Avatar/examples/Avatar.md
```diff
@@ -8,31 +8,65 @@ cssPrefix: pf-v6-c-avatar
...
+{{> avatar--colorful avatar--modifier="pf-m-yellow" avatar--AriaLabel="Yellow avatar with icon"}}
+{{> avatar--colorful avatar--modifier="pf-m-green" avatar--AriaLabel="Green avatar with icon"}}
+{{> avatar--colorful avatar--modifier="pf-m-teal" avatar--AriaLabel="Teal avatar with icon"}}
+{{> avatar--colorful avatar--modifier="pf-m-blue" avatar--AriaLabel="Blue avatar with icon"}}
+{{> avatar--colorful avatar--modifier="pf-m-purple" avatar--AriaLabel="Purple avatar with icon"}}
+{{> avatar--colorful avatar--modifier="pf-m-gray" avatar--AriaLabel="Gray avatar with icon"}}
+```
+
+### Colorful initials
+Use a color modifier with a user's initial to add visual interest when no image is available.
```
mcoker, line 59: Just a nit but the title says "initials" and the sentence says "initial". Maybe that's intentional?

mcoker (changes requested):
Just a few comments

src/patternfly/components/Avatar/examples/Avatar.md
jcmill: Awesome! Thank's for that.

src/patternfly/components/Avatar/avatar--colorful.hbs
jcmill: All these case issues was me trying to follow the original format of the avatar.hbs file. Should be locked in my brain now.

src/patternfly/components/Avatar/examples/Avatar.md
jcmill: Intentional. In a couple of places (jira and figma) it says "first initials" specifically, when referring to users.

mcoker (commented):
Is `.pf-m-colorful` meant to be used on its own or should it always be paired with a color variant? It seems like the latter, but looks like there is still code in place to make `.pf-m-colorful` on its own look like our default avatar SVG.

Assuming that isn't intentional/supported, I'd probably remove the [background color, color, and border color vars](https://github.com/jcmill/patternfly/blob/466e8b81febec0e194e1991c37683e11372d5ee8/src/patternfly/components/Avatar/avatar.scss#L27-L29) and give them default fallbacks (`transparent`, `inherit`, and `transparent`, or maybe even `revert` for all 3?) The main thing throwing me off for what's there now is that we're using the `nonstatus--gray--hover` token for color, which I'm assuming isn't the actual token we want to use there.

If it is intentional and we want to support `.pf-m-colorful` without a color modifier, I wonder if we should add an example.

Just one comment - right now the color variants are always setting `color` to the text color tokens, when ideally I'm assuming we would want icons to use the icon color tokens and the text (initals element) to text color tokens. I think you could do it by updating the `--Color` vars you have now to use the `--icon` equivalent tokens, then add a new var for the `__initials` element for each color variation that sets `__initials` to the text color. I tried to say that in a [previous comment](https://github.com/patternfly/patternfly/pull/8421#discussion_r3306841706) but I don't think it was clear 😅

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -21,6 +21,57 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+  --#{$avatar}__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$avatar}--m-sm__initials--FontSize: var(--pf-t--global--font--size--body--sm);
+  --#{$avatar}--m-sm__initials--FontWeight: var(--pf-t--global--font--weight--body--bold);
+  --#{$avatar}--m-lg__initials--FontSize: var(--pf-t--global--font--size--2xl);
+  --#{$avatar}--m-lg__initials--FontWeight: var(--pf-t--global--font--weight--heading--bold);
+  --#{$avatar}--m-lg__initials--FontFamily: var(--pf-t--global--font--family--heading);
+  --#{$avatar}--m-xl__initials--FontSize: var(--pf-t--global--font--size--4xl);
+  --#{$avatar}--m-xl__initials--FontWeight: var(--pf-t--global--font--weight--heading--default);
+  --#{$avatar}--m-xl__initials--FontFamily: var(--pf-t--global--font--family--heading);
+  --#{$avatar}--m-gray--Color: var(--pf-t--global--text--color--nonstatus--on-gray--hover);
```
mcoker: I think you meant to define it this way, since `--#{$avatar}--m-gray--Color` is already defined on L28

```suggestion
  --#{$avatar}__initials--m-gray--Color: var(--pf-t--global--text--color--nonstatus--on-gray--hover);
```

However, these initial vars should be named with the modifier first, like `--#{$avatar}--m-gray__initials--Color`

mcoker (commented):
Can we also add a `font-size` property to the wrapping avatar class and have each of the size modifiers set a font-size that's the same as the height/width value the size modifier sets.

src/patternfly/components/Avatar/avatar.scss
```diff
@@ -40,4 +97,107 @@ $pf-v6-c-avatar--sizes: "sm", "md", "lg", "xl";
...
+
+  &.pf-m-xl {
+    --#{$avatar}--FontSize: var(--#{$avatar}--m-xl--FontSize);
+    --#{$avatar}__initials--FontSize: var(--#{$avatar}--m-xl__initials--FontSize);
+    --#{$avatar}__initials--FontWeight: var(--#{$avatar}--m-xl__initials--FontWeight);
+    --#{$avatar}__initials--FontFamily: var(--#{$avatar}--m-xl__initials--FontFamily);
+  }
+
+  &.pf-m-colorful {
+  --#{$avatar}--BorderWidth: var(--#{$avatar}--m-colorful--BorderWidth);
```
mcoker:
```suggestion
    --#{$avatar}--BorderWidth: var(--#{$avatar}--m-colorful--BorderWidth);
```

mcoker (approved):
The smallest nit of nits.

PR #8427, opened 2026-05-20
feat(Page): update glass spacers

### LINKED ISSUE #8393: Bug - Page - Glass spacing
For the glass page sidebar/vertical navigation use 16px spacer instead of 24px spacer around nav. This should likely just change from the “lg” spacer token to “md”

Also we want to reduce the spacer between the masthead/sidebar/content from 24px to 16px. I think we’re using the page chrome token now - this should probably use the gutter token. Generally speaking, page chrome is for outside spacing, and gutter is for inner gaps.

---

**Jira Issue:** [PF-4153](https://redhat.atlassian.net/browse/PF-4153)

kmcfaul:
Closes: #8393

Updates sidebar spacer from lg to md token.
Updates the gap between masthead, sidebar, and content to use gutter token instead of page chrome. Outer spacers around the page should still be using page chrome token.

## Summary by CodeRabbit

* **Style**
  * Refined layout and spacing across themes: adjusted docked-nav, sidebar and main-container paddings/margins, recalculated main-container max-height, and updated XL breakpoint gutter behavior for more consistent display across breakpoints; root masthead horizontal padding now aligns with the inset page chrome spacing.

* **Chores**
  * Regenerated token files, refreshed generation timestamps, and introduced a new global inset page-chrome spacer token used by glass/felt themes.

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -249,9 +249,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__sidebar-main--PaddingBlockEnd--glass: var(--#{$page}__sidebar--PaddingBlockEnd);
   --#{$page}__sidebar-main--PaddingInlineStart--glass: var(--#{$page}__sidebar--PaddingInlineStart);
   --#{$page}__sidebar-main--PaddingInlineEnd--glass: var(--#{$page}__sidebar--PaddingInlineEnd);
-  --#{$page}__sidebar-main--MarginBlockStart--glass: calc(var(--pf-t--global--spacer--sm) * -1);
-  --#{$page}__sidebar-main--MarginBlockEnd--glass: var(--pf-t--global--spacer--md);
-  --#{$page}__sidebar-main--MarginInlineStart--glass: var(--pf-t--global--spacer--md);
+  --#{$page}__sidebar-main--MarginBlockStart--glass: calc(var(--pf-t--global--spacer--xs) * -1);
+  --#{$page}__sidebar-main--MarginBlockEnd--glass: calc(var(--pf-t--global--spacer--inset--page-chrome) - var(--pf-t--global--spacer--xs));
+  --#{$page}__sidebar-main--MarginInlineStart--glass: calc(var(--pf-t--global--spacer--inset--page-chrome) - var(--pf-t--global--spacer--xs));
   --#{$page}__sidebar-main--MarginInlineEnd--glass: var(--pf-t--global--spacer--md);
   --#{$page}__sidebar-main--BackgroundColor--glass: var(--#{$page}__sidebar--BackgroundColor);
   --#{$page}__sidebar-main--BackdropFilter--glass: var(--#{$page}__sidebar--BackdropFilter);
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -243,9 +243,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   // Glass theme
   --#{$page}--BackgroundColor--glass: transparent;
   --#{$page}__sidebar--Width--base--glass: calc(#{pf-size-prem(290px)} + var(--pf-t--global--spacer--inset--page-chrome) * 2);
-  --#{$page}__sidebar--MarginBlockStart--glass: var(--pf-t--global--spacer--inset--page-chrome);
-  --#{$page}__sidebar-body--PaddingInlineStart--glass: var(--pf-t--global--spacer--lg);
-  --#{$page}__sidebar-body--PaddingInlineEnd--glass: var(--pf-t--global--spacer--lg);
+  --#{$page}__sidebar--MarginBlockStart--glass: var(--pf-t--global--spacer--gutter--default);
+  --#{$page}__sidebar-body--PaddingInlineStart--glass: var(--pf-t--global--spacer--md);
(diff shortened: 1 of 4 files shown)


mcoker (commented):
Need to confirm with design:
* whether page-chrome token stays the same or should change to 16px.
* the mobile layout of the sidebar looks correct with the inset/offset

mcoker (changes requested):
[Full backstop report](https://drive.google.com/file/d/1xFku66Wr0mzHls5IMcIapUUyD3o9UXRe/view?usp=sharing)

Spotted these two issues:
* the masthead uses the "lg" spacer for it's left/right padding and that should be updated to use the page chrome spacer.
* the page content in the docked nav layout uses a "lg" top margin when it should use the gutter spacer on mobile (when there is a masthead visible) and the page chrome spacer on desktop (when the content is the first element on the page)

src/patternfly/components/Page/page.scss
```diff
@@ -10,9 +10,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}--inset: var(--pf-t--global--spacer--inset--page-chrome);
 
   // Docked nav
-  --#{$page}--m-dock__main-container--MaxHeight: calc(100% - var(--pf-t--global--spacer--lg) * 2);
+  --#{$page}--m-dock__main-container--MaxHeight: calc(100% - var(--pf-t--global--spacer--inset--page-chrome) * 2);
```
mcoker: Looks like this is duped on https://github.com/kmcfaul/patternfly/blob/1a3c6b7c95c89656658c4f9b93ad6949383e587e/src/patternfly/components/Page/page.scss#L239

```suggestion
```

src/patternfly/components/Page/page.scss
```diff
@@ -10,9 +10,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
   --#{$page}--m-dock__main-container--MarginBlockStart: 0;
```
mcoker: This is duped on https://github.com/kmcfaul/patternfly/blob/1a3c6b7c95c89656658c4f9b93ad6949383e587e/src/patternfly/components/Page/page.scss#L240

```suggestion
```

src/patternfly/components/Page/page.scss
```diff
@@ -10,9 +10,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
-  --#{$page}--m-dock__main-container--desktop--MarginBlockStart: var(--pf-t--global--spacer--lg);
+  --#{$page}--m-dock__main-container--desktop--MarginBlockStart: var(--pf-t--global--spacer--inset--page-chrome);
   --#{$page}--m-dock--ColumnGap: var(--pf-t--global--spacer--inset--page-chrome);
```
mcoker: This is duped on https://github.com/kmcfaul/patternfly/blob/1a3c6b7c95c89656658c4f9b93ad6949383e587e/src/patternfly/components/Page/page.scss#L241


```suggestion
```

src/patternfly/components/Page/page.scss
```diff
@@ -236,22 +236,22 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__drawer--c-drawer--BorderBlockStartColor: var(--pf-t--global--border--color--high-contrast);
 
   // Docked nav
-  --#{$page}--m-dock__main-container--MaxHeight: calc(100% - var(--pf-t--global--spacer--lg) * 2);
-  --#{$page}--m-dock__main-container--MarginBlockStart: var(--pf-t--global--spacer--lg);
+  --#{$page}--m-dock__main-container--MaxHeight: calc(100% - var(--pf-t--global--spacer--inset--page-chrome) - var(--pf-t--global--spacer--gutter--default));
```
mcoker: Can you add a desktop `--MaxHeight` var so we can adjust this value to `calc(100% - var(--page-chrome) * 2)`? You should be able to put it in the same spot that the desktop MarginBlockStart goes - https://github.com/kmcfaul/patternfly/blob/1a3c6b7c95c89656658c4f9b93ad6949383e587e/src/patternfly/components/Page/page.scss#L15

mcoker (changes requested):
Just a few comments on the vars. I still need to run visual regressions, but need to get them updated on the main branch first. Will re-review when those are updated.

src/patternfly/components/Page/page.scss
```diff
@@ -236,22 +236,22 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
+  --#{$page}--m-dock__main-container--MarginBlockStart: var(--pf-t--global--spacer--gutter--default);
```
mcoker: This is changing the non-glass margin, too. We can revert this but set it to `--pf-t--global--spacer--inset--page-chrome` instead of `--pf-t--global--spacer--lg`, then add a `--#{$page}--m-dock__main-container--MarginBlockStart--glass` var that we set to `--pf-t--global--spacer--gutter--default`

mcoker (changes requested):

mcoker (approved):
LGTM!

mcoker (changes requested):
Oops missed one spot. Looks like there are still duped vars here - https://github.com/kmcfaul/patternfly/blob/5cd7e7ef7b1fc9a170d0c5bcba6ce70eac4deab0/src/patternfly/components/Page/page.scss#L238-L241

You can delete the previously defined ones since those are not applying currently anyways. Then can you group these separate blocks so all of the vars are in a single docked nav section?

mcoker (approved):
🎃

lboehling (approved):
lgtm!

PR #8431, opened 2026-05-26
fix(docked): add min width on icon elements

### LINKED ISSUE #8333: Bug - button/menu-toggle - update icon width for docked nav
For docked nav, to mimic the nav component icons, let's set `min-width: 1lh` for the button/menu-toggle icon elements so that there is a little bit of extra space around the icon. This should only apply to the docked variant, and technically only needs to be applied when text is expanded `.pf-m-text-expanded`, but could be used with `.pf-m-docked`, too, if it makes more sense to do that.

---

**Jira Issue:** [PF-4044](https://redhat.atlassian.net/browse/PF-4044)

jcmill:
Closes: #8333

Sets min-width: 1lh on icon elements. This mimics the nav component's icon spacing pattern for docked navigation.

## Summary by CodeRabbit

* **Style**
  * Enforced a minimum icon width in docked/text-expanded button layouts to keep icons aligned and prevent layout shifts when text visibility changes.
  * Applied the same minimum-width constraint to menu toggle icons in docked/text-expanded layouts for more consistent spacing and visual balance across controls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/patternfly/patternfly/pull/8431?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -989,6 +989,10 @@
     justify-content: flex-start;
     width: 100%;
 
+    .#{$button}__icon {
+      min-width: 1lh;
+    }
+
     @media (min-width: $pf-v6-global--breakpoint--dock--desktop) {
       justify-content: center;
 
@@ -1005,12 +1009,6 @@
 
     .#{$button}__text {
       display: revert;
-    }
-  }
-
-  &:where(.pf-m-docked, .pf-m-text-expanded) {
-    .#{$button}__icon {
-      min-width: 1lh;
     }
   }
 }
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -1007,6 +1007,12 @@
       display: revert;
     }
   }
+
+  &:where(.pf-m-docked, .pf-m-text-expanded) {
+    .#{$button}__icon {
+      min-width: 1lh;
+    }
+  }
 }
 
 .#{$button}__icon {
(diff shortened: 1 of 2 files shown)


src/patternfly/components/Button/button.scss
mcoker: Couple of things on this
* You should always attach a `.pf-m-` modifier class to a BEM element, so here all you need to do is add `&`'s to the modifier classes
* You shouldn't need to target `.pf-m-text-expanded`. Just putting it on `.pf-m-docked` should be fine. `.pf-m-docked` is always present for a docked button/menu-toggle, and `.pf-m-text-expanded` doesn't change the icon in a way where you would need to set the `min-width` for it, too.

I would nest it under this block https://github.com/jcmill/patternfly/blob/a7e2f7aa26ed6b2849caa790e232a24e564fff13/src/patternfly/components/Button/button.scss#L988-L990

src/patternfly/components/MenuToggle/menu-toggle.scss
mcoker: Same comment here as for button but you can put it here https://github.com/jcmill/patternfly/blob/a7e2f7aa26ed6b2849caa790e232a24e564fff13/src/patternfly/components/MenuToggle/menu-toggle.scss#L437-L439

mcoker (changes requested):

mcoker (approved):
🚀

PR #8438, opened 2026-06-02
 fix: update stylelint deps

mcoker:
* Pulls in changes from a PR in the design repo to use `rgb()` instead of `rgba()` - https://github.com/patternfly/design-tokens/pull/153
* Updates the stylelint deps to latest, including breaking changes to stylelint 17
* Adds "stylelint-order" as a dev dependency since it no longer comes with "stylelint-config-recess-order"
* Removes "stylelint-use-logical-spec" - that's supported natively now with the "property-layout-mappings" rule
* Updates the following stylelint errors
  * Property order
  * Deprecated rules (`clip` -> `clip-path`, `word-break` -> `overflow-wrap`, `grid-gap` -> `gap`)
  * Disables stylelint for max-selector violations in `:is()/where()` - pre-existing, but there was a bug with stylelint that prevented them from being detected

[BACKSTOP-REPORT.ZIP](https://drive.google.com/file/d/1rj-txBZV2XGnPJn5zj7b2Mry6DQ-0-KZ/view?usp=sharing)

## Summary by CodeRabbit

* **Style**
  * Modernized CSS grid properties across components, replacing legacy syntax with current standards.
  * Updated text-wrapping to use overflow-wrap for improved long-word handling.
  * Refined CSS property ordering for more consistent styling output.
  * Updated Stylelint configuration and plugin set to align linting rules.

* **Accessibility**
  * Improved screen-reader utility using a modern clip-path approach.

* **Content**
  * Updated an example modal icon in documentation.


jcmill (approved):
LGTM

PR #8443, opened 2026-06-04
fix(panel, docked nav): borders

### LINKED ISSUE #8399: Bug - Panel and docked nav borders
Panel

- Add high contrast border

Docked nav

- Add subtle border on desktop for default and text-expanded

---

**Jira Issue:** [PF-4181](https://redhat.atlassian.net/browse/PF-4181)

jcmill:
fixes: #8399

Adds hc border to panel and a border to the __dock-main for the docked nav.

## Summary by CodeRabbit

* **Style**
  * Adjusted desktop dock border sourcing to use regular border width and subtle-regular border color, with explicit desktop overrides for consistency at larger breakpoints.
  * Changed panel “before” default border width to a high-contrast value; the secondary panel modifier no longer overrides border width and now only applies the secondary background.

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -27,7 +27,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__dock--m-expanded__dock-main--BoxShadow: var(--pf-t--global--box-shadow--sm--right);
   --#{$page}__dock-main--desktop--BoxShadow: var(--pf-t--global--box-shadow--glass--default, none);
   --#{$page}__dock-main--BackdropFilter: var(--pf-t--global--background--filter--glass--blur--primary, revert);
-  --#{$page}__dock-main--BorderInlineEndWidth: var(--pf-t--global--border--width--glass--default, 0);
+  --#{$page}__dock-main--BorderInlineEndWidth: 0;
   --#{$page}__dock-main--BorderInlineEndColor: transparent;
   --#{$page}__dock-main--desktop--BorderInlineEndWidth: var(--pf-t--global--border--width--regular);
   --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--subtle);
@@ -286,7 +286,8 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
     --#{$page}__main-container--MarginBlockStart: var(--#{$page}__main-container--MarginBlockStart--glass);
     --#{$page}__main-container--MaxHeight: var(--#{$page}__main-container--MaxHeight--glass);
     --#{$page}--m-dock__main-container--MaxHeight: var(--#{$page}--m-dock__main-container--MaxHeight--glass);
-    --#{$page}--m-dock__main-container--MarginBlockStart: var(--#{$page}--m-dock__main-container--MarginBlockStart--glass);  
+    --#{$page}--m-dock__main-container--MarginBlockStart: var(--#{$page}--m-dock__main-container--MarginBlockStart--glass);
+    --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--glass--default);
   }
 
   @media (min-width: $pf-v6-global--breakpoint--xl) {
@@ -448,9 +449,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   }
 
   @media (min-width: $pf-v6-global--breakpoint--dock--desktop) {
-    --#{$page}__dock-main--BackgroundColor: var(--#{$page}__dock-main--desktop--BackgroundColor);
-    --#{$page}__dock-main--BorderInlineEndColor: var(--#{$page}__dock-main--desktop--BorderInlineEndColor);
-
     position: revert;
     inset: revert;
     visibility: revert;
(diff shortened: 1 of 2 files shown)


mcoker (commented):
Want to confirm one thing with design (@andrew-ronaldson @lboehling). Right now there is a subtle right border when expanded on mobile in non-glass theme. In glass theme, we're using the glass border, which isn't really visible. Since the docked nav on mobile uses the floating background color in glass and non-glass, should we use a subtle border on mobile in glass, too?

That would mean that docked nav has a subtle right border all of the time, except in glass on desktop - then it uses the glass border.

Attached a video of the border in glass on mobile. It's toggling between the subtle border and glass border that it uses now. 

https://github.com/user-attachments/assets/d7d024c2-92f7-48b5-97ab-8e9495234624

src/patternfly/components/Panel/panel.scss
```diff
@@ -9,12 +9,11 @@
   --#{$panel}--BorderRadius: var(--pf-t--global--border--radius--medium);
 
   // border
-  --#{$panel}--before--BorderWidth: 0;
+  --#{$panel}--before--BorderWidth: var(--pf-t--global--border--width--high-contrast--regular);
   --#{$panel}--before--BorderColor: var(--pf-t--global--border--color--high-contrast);
 
   // secondary modifier
   --#{$panel}--m-secondary--BackgroundColor: var(--pf-t--global--background--color--secondary--default);
-  --#{$panel}--m-secondary--before--BorderWidth: var(--pf-t--global--border--width--high-contrast--regular);
```
mcoker, line 17: Need to add this back to avoid a breaking change. I'd also add this to the end of the line so we can flag it for removal in our next breaking change.  `// TODO - remove in a breaking change`

src/patternfly/components/Panel/panel.scss
```diff
@@ -100,7 +99,6 @@
 
   &.pf-m-secondary {
     --#{$panel}--BackgroundColor: var(--#{$panel}--m-secondary--BackgroundColor);
-    --#{$panel}--before--BorderWidth: var(--#{$panel}--m-secondary--before--BorderWidth);
```
mcoker, line 103: Need to add this back.

src/patternfly/components/Page/page.scss
```diff
@@ -27,9 +27,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
   --#{$page}__dock-main--desktop--BoxShadow: var(--pf-t--global--box-shadow--glass--default, none);
   --#{$page}__dock-main--BackdropFilter: var(--pf-t--global--background--filter--glass--blur--primary, revert);
-  --#{$page}__dock-main--BorderInlineEndWidth: var(--pf-t--global--border--width--glass--default, 0);
+  --#{$page}__dock-main--BorderInlineEndWidth: 0;
   --#{$page}__dock-main--BorderInlineEndColor: transparent;
-  --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--glass--default, transparent);
+  --#{$page}__dock-main--desktop--BorderInlineEndWidth: var(--pf-t--global--border--width--regular);
+  --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--subtle);
   --#{$page}__dock--m-expanded__dock-main--BorderInlineEndWidth: var(--pf-t--global--border--width--glass--default, var(--pf-t--global--border--width--regular));
   --#{$page}__dock--m-expanded__dock-main--BorderInlineEndColor: var(--pf-t--global--border--color--glass--default, var(--pf-t--global--border--color--subtle));
```
mcoker: Like we discussed in the meeting, here's how the borders should show:

Mobile
* No border on docked nav (default and glass) since there is a shadow

Desktop
* Default theme - uses subtle border color
* Glass theme - uses glass border color

I tested locally and I think you can just set this var to transparent, and the desktop borders are already configured properly. 

```suggestion
  --#{$page}__dock--m-expanded__dock-main--BorderInlineEndColor: transparent;
```

src/patternfly/components/Page/page.scss
```diff
@@ -27,9 +27,10 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__dock--m-expanded__dock-main--BoxShadow: var(--pf-t--global--box-shadow--sm--right);
   --#{$page}__dock-main--desktop--BoxShadow: var(--pf-t--global--box-shadow--glass--default, none);
   --#{$page}__dock-main--BackdropFilter: var(--pf-t--global--background--filter--glass--blur--primary, revert);
-  --#{$page}__dock-main--BorderInlineEndWidth: var(--pf-t--global--border--width--glass--default, 0);
+  --#{$page}__dock-main--BorderInlineEndWidth: 0;
   --#{$page}__dock-main--BorderInlineEndColor: transparent;
-  --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--glass--default, transparent);
+  --#{$page}__dock-main--desktop--BorderInlineEndWidth: var(--pf-t--global--border--width--regular);
+  --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--subtle);
```
mcoker: Not necessary, but you could clean this up a little and set this var to the glass token, and have the subtle var as the fallback. Then you could remove the glass specific token you added here - https://github.com/jcmill/patternfly/blob/0285f68e95277c4c514fb815e980702439b1bf6c/src/patternfly/components/Page/page.scss#L287

```suggestion
  --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--glass--default, var(--pf-t--global--border--color--subtle));
```

src/patternfly/components/Page/page.scss
```diff
@@ -283,6 +284,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
     --#{$page}__sidebar-main--BorderRadius: var(--#{$page}__sidebar-main--BorderRadius--glass);
     --#{$page}__sidebar-main--BoxShadow: var(--#{$page}__sidebar-main--BoxShadow--glass);
     --#{$page}__main-container--MarginBlockStart: var(--#{$page}__main-container--MarginBlockStart--glass);
+    --#{$page}__dock-main--desktop--BorderInlineEndColor: var(--pf-t--global--border--color--glass--default);
```
mcoker: If you make the suggestion above to use this with the fallback, you can remove this line.

```suggestion
```

src/patternfly/components/Page/page.scss
```diff
@@ -975,4 +976,4 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
       flex-grow: 1;
     }
   }
-}
+}
```
mcoker, line 979: Looks like we need a newline here

mcoker (approved):
🎃

PR #8472, opened 2026-06-23
fix(table): table thead background issue in safari

### LINKED ISSUE #8453: Bug - Table - Fixed header will hide the table rows in Safari browser
**Describe the problem**
This is an issue with 

- [ ] Patternfly 5  

- [x] Patternfly 6  

- [ ] other

Can be viewed on the "Sticky header" example on the PF website, while using Safari.

**Expected behavior**
Sticky header works in Safari, can see all table rows, header also has a drop shadow.

Can hide the pseudo elements (target just safari with @supports), but then we lose the shadow below the header.

**What is your environment?**
 - OS: MacOS
 - Browser: Safari

Nexus

image

---

**Jira Issue:** [PF-4296](https://redhat.atlassian.net/browse/PF-4296)

jcmill:
Closes #8453 
Move background and border ::before and ::after elements from thead to th for Safari support.

## Summary by CodeRabbit

* **Style**
  * Dock component: Updated desktop end-border width and color to use the standard border tokens for more consistent rendering.
  * Panel component: Adjusted default “before” border width to a high-contrast value and ensured the secondary modifier no longer overrides that border width.
  * Table component: Improved sticky header visuals by moving the background and border decoration from the header container pseudo-element to individual header cells.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -301,7 +301,7 @@
     table-layout: fixed;
   }
 
-  &.pf-m-sticky-header-base > .#{$table}__thead, &.pf-m-sticky-header-base .#{$table}__th {
+  &.pf-m-sticky-header-base > .#{$table}__thead, &.pf-m-sticky-header-base > .#{$table}__thead .#{$table}__th {
     &::before,
     &::after {
       opacity: 0;
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -311,7 +311,7 @@
     }
   }
 
-  &.pf-m-sticky-header-stuck > .#{$table}__thead, &.pf-m-sticky-header-stuck .#{$table}__th {
+  &.pf-m-sticky-header-stuck > .#{$table}__thead, &.pf-m-sticky-header-stuck > .#{$table}__thead .#{$table}__th {
     &::before,
     &::after {
       opacity: 1;
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -301,7 +301,7 @@
     table-layout: fixed;
   }
 
-  &.pf-m-sticky-header-base > .#{$table}__thead {
+  &.pf-m-sticky-header-base > .#{$table}__thead, &.pf-m-sticky-header-base .#{$table}__th {
     &::before,
     &::after {
       opacity: 0;
@@ -311,12 +311,7 @@
     }
   }
 
-  &.pf-m-sticky-header,
-  &.pf-m-sticky-header-stuck {
-    --#{$table}__thead--m-nested-column-header--BorderBlockEndWidth: 0;
-  }
-
-  &.pf-m-sticky-header-stuck > .#{$table}__thead {
+  &.pf-m-sticky-header-stuck > .#{$table}__thead, &.pf-m-sticky-header-stuck .#{$table}__th {
     &::before,
     &::after {
       opacity: 1;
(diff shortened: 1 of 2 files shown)


mcoker:
@jcmill looks like you may have created this branch from the panel/docked nav branch? Looks like the page/panel stylesheets have a conflict.

mcoker (commented):
* Looks like there is a bug with nested column headers. I wonder if something like this would work?

```
tr:first-child  {
  .#{$table}__th:first-child { top/left border }
  .#{$table}__th:last-child { top/right border }
}
tr:last-child  {
  .#{$table}__th:first-child { bottom/left border }
  .#{$table}__th:last-child { bottom/right border }
}
```
 
Image

* We still have some styles on thead:before/after and some need to be moved to the thead cells. Not an exhaustive list but here are some I found
  * https://github.com/jcmill/patternfly/blob/484b66c6941a3eddb880866264673a4f28903445/src/patternfly/components/Table/table.scss#L304-L312
  * https://github.com/jcmill/patternfly/blob/484b66c6941a3eddb880866264673a4f28903445/src/patternfly/components/Table/table.scss#L330-L343
  * https://github.com/jcmill/patternfly/blob/484b66c6941a3eddb880866264673a4f28903445/src/patternfly/components/Table/table.scss#L1309-L1316

src/patternfly/components/Table/table.scss
```diff
@@ -301,7 +301,7 @@
     table-layout: fixed;
   }
 
-  &.pf-m-sticky-header-base > .#{$table}__thead {
+  &.pf-m-sticky-header-base > .#{$table}__thead, &.pf-m-sticky-header-base .#{$table}__th {
```
mcoker: I would add a direct child selector for the `thead` do avoid targeting nested tables where the parent table has sticky columns and the nested table does not.

```suggestion
  &.pf-m-sticky-header-base > .#{$table}__thead, &.pf-m-sticky-header-base > .#{$table}__thead .#{$table}__th {
```

src/patternfly/components/Table/table.scss
```diff
@@ -311,12 +311,7 @@
     }
   }
 
-  &.pf-m-sticky-header,
-  &.pf-m-sticky-header-stuck {
-    --#{$table}__thead--m-nested-column-header--BorderBlockEndWidth: 0;
-  }
-
-  &.pf-m-sticky-header-stuck > .#{$table}__thead {
+  &.pf-m-sticky-header-stuck > .#{$table}__thead, &.pf-m-sticky-header-stuck .#{$table}__th {
```
mcoker: Same here

```suggestion
  &.pf-m-sticky-header-stuck > .#{$table}__thead, &.pf-m-sticky-header-stuck > .#{$table}__thead .#{$table}__th {
```


You could also probably write these this way if that's more readable, doesn't matter to me

```scss
  &.pf-m-sticky-header-stuck > .#{$table}__thead {
    &,
    .#{$table}__th {
```

mcoker:
I worked on this a little and added a PR in your repo to show the code diff, although it doesn't have the last commit you just made - https://github.com/jcmill/patternfly/pull/1

And I created a preview at http://8453-table-thead-hidden-in-safari.surge.sh/

The CSS should fix all of the borders except in Safari with sticky, nested column headers in non-glass theme. That border appears initially, then scrolls off screen. To fix that, I've gone back to how we did that before by adding a `tr.pf-m-border-row` as the last thing in the thead for that particular table. For some reason, if you use `height` and `background-color` to mimic a 1px border, that will stay in the thead and act as a bottom border in Safari. That's how it is in the [pf5 sticky nested column header table](https://v5-archive.patternfly.org/components/table/html/nested-column-headers-sticky-header/), too. Having `tr.pf-m-border-row` there doesn't (or shouldn't anyways) conflict with the borders that work for chrome/FF.

So we would say if you have a table with 1) sticky header, and 2) nested column headers, you need a `tr.pf-m-border-row` - which is how it was before our v6 release just a few weeks ago. Technically you only need it for Safari, since it works without that element in chrome/firefox.

This doesn't fix the border-radius and shadow issue in Safari, but those can come later.

mcoker (approved):
NICE WORK!

PR #8487, opened 2026-07-08
fix: address word-break/overflow-wrap bug

### LINKED ISSUE #8484: Bug - Modal - `overflow-wrap` migration from `word-break` was applied wrong
**Describe the problem**
This is an issue with 

- [ ] Patternfly 5  

- [x] Patternfly 6  

- [ ] other

Within Cockpit I saw a regression with PF 6.6.0 bump where our modal suddenly doesn't wrap text correctly. As far as I can tell there aren't any CSS rules on our end that would warrant this.

Root cause determined to be the migration of `word-break` to `overflow-wrap` where the migration was not done correctly.
[According to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/word-break#break-word), `word-break: break-word` has the same behaviour as `overflow-wrap: anywhere`. Whereas the refactored change changed it to `overflow-wrap: break-word`

I'm reading and likely seems related to this behaviour from [`overflow-wrap: break-word`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-wrap#break-word):

> ... soft wrap opportunities introduced by the word break are NOT considered when calculating min-content intrinsic sizes.

Introduced in

- https://github.com/patternfly/patternfly/pull/8438

**Expected behavior**
Not overflowing when long text inputs are provided

Will introduce a CSS override, not blocking. Our override:

```css
.pf-v6-c-clipboard-copy__expandable-content,
.pf-v6-c-clipboard-copy__text,
.pf-v6-c-menu__item-description,
.pf-v6-c-modal-box__body,
.pf-v6-c-notification-drawer__list-item-header-title,
.pf-v6-c-notification-drawer__list-item-description,
.pf-v6-c-notification-drawer__group-toggle-title,
.pf-v6-c-popover__body,
.pf-v6-c-progress__description,
.pf-v6-c-progress__status,
.pf-v6-c-title,
.pf-v6-c-tooltip__content,
.pf-v6-c-wizard__title,
.pf-v6-c-wizard__toggle-list-item,
.pf-v6-c-wizard__nav-link,
.pf-v6-c-wizard__main {
  overflow-wrap: anywhere;
}
```

**Screenshots**
Left browser is reference, right browser is the PF 6.6.0 bump

image

Cockpit 365, July 15

**Any other information?**

---

**Jira Issue:** [PF-4396](https://redhat.atlassian.net/browse/PF-4396)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8484

Follow up to a bug from https://github.com/patternfly/patternfly/pull/8438. I incorrectly replaced `word-break: break-word` with `overflow-wrap: break-word`, when it should have been replaced with `overflow-wrap: anywhere`. This PR goes through all of the changes in #8438 and changes any rules that were `word-break: break-word` to `overflow-wrap: anywhere` instead.

**Note** - there are some instances of `word-wrap: break-word` (`word-wrap` vs `word-break`) replaced with `overflow-wrap: break-word` originally, which is correct, so I left those alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text wrapping across multiple UI elements, including modals, alerts, breadcrumbs, menus, tooltips, notifications, progress displays, and wizards.
  - Long or unbroken text now wraps more gracefully in tighter spaces, reducing overflow and layout issues.
  - Updated title and clipboard copy text behavior for better readability when content is constrained.

### DIFF
--- a/src/patternfly/components/AboutModalBox/about-modal-box.scss
+++ b/src/patternfly/components/AboutModalBox/about-modal-box.scss
@@ -176,7 +176,7 @@
   overflow-x: hidden;
   overflow-y: auto;
   overscroll-behavior: contain;
-  overflow-wrap: break-word;
+  overflow-wrap: anywhere;
   -webkit-overflow-scrolling: touch;
 
   @media screen and (min-width: $pf-v6-global--breakpoint--sm) {
--- a/src/patternfly/components/Alert/alert.scss
+++ b/src/patternfly/components/Alert/alert.scss
@@ -205,7 +205,7 @@
   grid-area: title;
   font-weight: var(--#{$alert}__title--FontWeight);
   color: var(--#{$alert}__title--Color);
-  overflow-wrap: break-word;
+  overflow-wrap: anywhere;
 
   &.pf-m-truncate {
     @include pf-v6-line-clamp("var(--#{$alert}__title--max-lines)");
@@ -215,7 +215,7 @@
 .#{$alert}__description {
   grid-area: description;
   padding-block-start: var(--#{$alert}__description--PaddingBlockStart);
-  overflow-wrap: break-word;
+  overflow-wrap: anywhere;
 
   + .#{$alert}__action-group {
     --#{$alert}__action-group--PaddingBlockStart: var(--#{$alert}__description--action-group--PaddingBlockStart);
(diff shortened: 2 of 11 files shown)


jcmill (approved):
✈️

PR #8490, opened 2026-07-09
fix(tree-view): revert caret rotations

### LINKED ISSUE #8474: Treeview - Revert caret back to right/down.
In 6.5 we made the global change to change the expand/collapse pattern globally to be down/up. There has been product feedback that industry standards for a tree view use the right/down pattern. This is present in design systems like [Carbon](https://carbondesignsystem.com/components/tree-view/usage/) where the navigation and accordions are the down/up carets and the exception is their tree view.

---

**Jira Issue:** [PF-4348](https://redhat.atlassian.net/browse/PF-4348)

jcmill:
Closes: #8474 

Reverted the expand/collapse pattern from down/up to right/down.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the TreeView node toggle icon rotation to better reflect expanded vs. collapsed state.
  * Updated the TreeView toggle icon behavior so the appropriate caret direction is shown when the toggle attribute is present.

### DIFF
--- a/src/patternfly/components/TreeView/tree-view-node-toggle-icon.hbs
+++ b/src/patternfly/components/TreeView/tree-view-node-toggle-icon.hbs
@@ -2,5 +2,5 @@
   {{#if tree-view-node-toggle-icon--attribute}}
     {{{tree-view-node-toggle-icon--attribute}}}
   {{/if}}>
-  {{pfIcon "rh-microns-caret-down"}}
+  {{pfIcon "rh-microns-caret-right"}}
 </span>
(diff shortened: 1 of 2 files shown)


andrew-ronaldson (approved):
🥇

mcoker (commented):
LGTM though curious if you tried updating the icon itself, too, so that the icon without rotation is pointing in the collapsed direction (right), then we rotate it 90deg for the expanded position (down). That should be how all of our other caret/expand icons are setup.

Also be sure to add the issue # to the description or link the issue to the PR so the issue closes when the PR is merged and so that it's easy to find the issue if someone lands on this PR.

jcmill:
Good call. Just updated the pr.

mcoker (approved):
🚀

PR #8491, opened 2026-07-09
fix(table): nested table sorting

### LINKED ISSUE #8485: Bug - Table - sorting is not available for nested tables with v6.6.0
**Describe the problem**
After updating to 6.6.0, QE discovered that sorting no longer works for nested table columns.

Update to v6.6.0

You can only click on arrows in the last couple columns for this example -- the first 5 are dead links.
See https://www.patternfly.org/components/table#nested-column-headers

**Expected behavior**
Sorting should be enabled

**Is this issue blocking you?
This is blocking the ROS namespace UI release

**Screenshots**

image

**What is your environment?**
 - OS: [e.g. iOS]: Mac
 - Browser [e.g. chrome, safari]: Chrome
 - Version [e.g. 22]: Version 149.0.7827.201 (Official Build) (arm64)

ROS Optimizations

**Any other information?**
From Cursor:

This lines up with a PatternFly 6.6.0 CSS regression, not something wrong in your column config.

What changed in 6.6.0
In 6.5.2, nested header borders used a ::after pseudo-element on the <thead>, and only that overlay had pointer-events: none:

// PF 6.5.2
&.pf-m-nested-column-header {
  &::after {
    pointer-events: none;  // only the decorative overlay
    ...
  }
}
In 6.6.0, that was moved to the subheader row itself:

// PF 6.6.0
&.pf-m-nested-column-header {
  .#{$table}__tr:last-child:not(.pf-m-border-row) {
    pointer-events: none;  // the entire subheader row
    border-block-end: ...
  }
}
That matches your symptoms exactly:

"Last reported" works — it's in the first header row with rowSpan: 2, not the last <tr>
All other sortable columns fail — they're in the second row, which now has pointer-events: none applied directly
This looks tied to the sticky header work in [patternfly/patternfly#8290](https://github.com/patternfly/patternfly/pull/8290), where the border styling moved from a ::after on <thead> to the last <tr>.

---

**Jira Issue:** [PF-4397](https://redhat.atlassian.net/browse/PF-4397)

jcmill:
Fixes: #8485 

Change :last tr pointer-events to auto to allow for click through with new background treatment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interaction for nested table column headers by enabling pointer events on the last nested header row, making it easier to interact with nested header layouts.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -1304,7 +1304,7 @@
 
     .#{$table}__tr:last-child:not(.pf-m-border-row) {
       z-index: var(--#{$table}__thead--m-nested-column-header--after--ZIndex);
-      pointer-events: none;
+      pointer-events: auto;
       border-block-end: var(--#{$table}__thead--m-nested-column-header--BorderBlockEndWidth) solid var(--#{$table}__thead--m-nested-column-header--BorderBlockEndColor);
     }
 
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -1304,7 +1304,6 @@
 
     .#{$table}__tr:last-child:not(.pf-m-border-row) {
       z-index: var(--#{$table}__thead--m-nested-column-header--after--ZIndex);
-      pointer-events: auto;
       border-block-end: var(--#{$table}__thead--m-nested-column-header--BorderBlockEndWidth) solid var(--#{$table}__thead--m-nested-column-header--BorderBlockEndColor);
     }
 


src/patternfly/components/Table/table.scss
```diff
@@ -1304,7 +1304,7 @@
 
     .#{$table}__tr:last-child:not(.pf-m-border-row) {
       z-index: var(--#{$table}__thead--m-nested-column-header--after--ZIndex);
-      pointer-events: none;
+      pointer-events: auto;
```
mcoker: WDYT about just removing the declaration? Doesn't look like it's needed.


```suggestion
```

jcmill: I like removing it. I was concerned about removing it for two reasons. First was potential future proofing and second, was a concern due to existing instances and not wanting to deviate too much.

mcoker (approved):
🎃

PR #8495, opened 2026-07-14
fix(plain actions): update tokens to vertical-plain

### LINKED ISSUE #8251: Bug - Update plain actions to use control-vertical-plain token
Audit components for plain actions and ensure top/bottom padding (and anything else like a margin offset for a plain actions padding) is using the `--pf-t--global--spacer--control--vertical--plain` token.

Probably best to look for all uses of `pf-t--global--spacer--control--vertical--default`and update to the plain token if it's a plain styled element.

---

**Jira Issue:** [PF-3823](https://redhat.atlassian.net/browse/PF-3823)

jcmill:
Closes #8251 

Update 12 components to use `--pf-t--global--spacer--control--vertical--plain` for plain-styled actions. Updates were applied to specific `--pf-t--global--spacer--control--vertical--default` or the raw `--pf-t--global--spacer--sm` tokens in block paddings and margin offsets.

## Summary by CodeRabbit

* **Style**
  * Refined vertical spacing across accordion, alert, breadcrumb, card, data list, hint, modal, navigation, popover, table, and tree view components.
  * Standardized control and action spacing for more consistent alignment and interaction areas.
  * Adjusted component padding and margins to improve visual consistency across states and layouts.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -6,9 +6,9 @@
   --#{$button}--JustifyContent: center;
   --#{$button}--Gap: var(--pf-t--global--spacer--gap--text-to-element--default);
   --#{$button}--MinWidth: calc(1lh + var(--#{$button}--PaddingBlockStart) + var(--#{$button}--PaddingBlockEnd));
-  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
+  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--default);
   --#{$button}--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--default);
-  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
+  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--default);
   --#{$button}--PaddingInlineStart: var(--pf-t--global--spacer--action--horizontal--default);
   --#{$button}--Color: var(--pf-t--global--text--color--regular);
   --#{$button}--LineHeight: var(--pf-t--global--font--line-height--body);
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -6,9 +6,9 @@
   --#{$button}--JustifyContent: center;
   --#{$button}--Gap: var(--pf-t--global--spacer--gap--text-to-element--default);
   --#{$button}--MinWidth: calc(1lh + var(--#{$button}--PaddingBlockStart) + var(--#{$button}--PaddingBlockEnd));
-  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--default);
+  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
   --#{$button}--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--default);
-  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--default);
+  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
   --#{$button}--PaddingInlineStart: var(--pf-t--global--spacer--action--horizontal--default);
   --#{$button}--Color: var(--pf-t--global--text--color--regular);
   --#{$button}--LineHeight: var(--pf-t--global--font--line-height--body);
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -89,8 +89,8 @@
   --#{$table}__toggle--c-button--m-expanded__toggle-icon--Rotate: -180deg;
 
   // * Table button
-  --#{$table}__button--PaddingBlockStart: var(--pf-t--global--spacer--sm);
-  --#{$table}__button--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
+  --#{$table}__button--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
+  --#{$table}__button--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
   --#{$table}__button--PaddingInlineStart: var(--pf-t--global--spacer--md);
   --#{$table}__button--PaddingInlineEnd: var(--pf-t--global--spacer--md);
   --#{$table}__button--Color: var(--pf-t--global--text--color--regular);
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -79,8 +79,8 @@
   --#{$table}--cell--hidden-visible--Display: table-cell;
 
   // * Table toggle
-  --#{$table}__toggle--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
-  --#{$table}__toggle--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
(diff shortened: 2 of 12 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -6,9 +6,9 @@
   --#{$button}--JustifyContent: center;
   --#{$button}--Gap: var(--pf-t--global--spacer--gap--text-to-element--default);
   --#{$button}--MinWidth: calc(1lh + var(--#{$button}--PaddingBlockStart) + var(--#{$button}--PaddingBlockEnd));
-  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--default);
+  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
   --#{$button}--PaddingInlineEnd: var(--pf-t--global--spacer--action--horizontal--default);
-  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--default);
+  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
```
mcoker: This is for the default spacer for buttons and the default buttons are not plain

```suggestion
  --#{$button}--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--default);
```

src/patternfly/components/Button/button.scss
```diff
@@ -6,9 +6,9 @@
   --#{$button}--JustifyContent: center;
   --#{$button}--Gap: var(--pf-t--global--spacer--gap--text-to-element--default);
   --#{$button}--MinWidth: calc(1lh + var(--#{$button}--PaddingBlockStart) + var(--#{$button}--PaddingBlockEnd));
-  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--default);
+  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
```
```suggestion
  --#{$button}--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--default);
```

src/patternfly/components/Table/table.scss
```diff
@@ -79,8 +79,8 @@
   --#{$table}--cell--hidden-visible--Display: table-cell;
 
   // * Table toggle
-  --#{$table}__toggle--PaddingBlockStart: var(--pf-t--global--spacer--sm);
-  --#{$table}__toggle--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
+  --#{$table}__toggle--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
+  --#{$table}__toggle--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
```
mcoker: This is a `<td>` and isn't really meant to represent a button's padding. As in, if `--pf-t--global--spacer--control--vertical--plain`'s value changed at some point, we wouldn't want this `<td>`'s padding to change to that new value, too.

Here's why we have a small spacer as the padding:
* Table cells use a common top padding spacer, `--pf-v6-c-table--cell--PaddingBlockStart`, which is 16px https://github.com/patternfly/patternfly/blob/716f0e546cb9da7dade2907e6ad1f32584da027b/src/patternfly/components/Table/table.scss#L39-L43
* This `<td>` has top padding of 8px, and has a plain button (the toggle) as a child, and the button has 8px top padding. We want the icon in the button to align vertically with the text in adjacent cells, so the 8px top padding on the `<td>` plus the 8px padding on the button aligns it with the adjacent cells with 16px top padding.

So what we _could_ do is either
* Update this `<td>`'s top padding to `calc(var(--pf-v6-c-table--cell--PaddingBlockStart) - var(--pf-t--global--spacer--control--vertical--plain))`. Also handle the bottom padding, too. This would probably what I would do?
* Use negative margins, like use `--pf-v6-c-table--cell--PaddingBlockStart` as the top padding for this `<td>` to match all of the other cells, then maybe target the button inside of it and give it a negative margin of `--pf-t--global--spacer--control--vertical--plain`.

But IMO that would be a separate issue. Feel free to create one if you want to fix it. There may be other places we aren't handling these types of adjustments correctly.

src/patternfly/components/Table/table.scss
```diff
@@ -89,8 +89,8 @@
   --#{$table}__toggle--c-button--m-expanded__toggle-icon--Rotate: -180deg;
 
   // * Table button
-  --#{$table}__button--PaddingBlockStart: var(--pf-t--global--spacer--sm);
-  --#{$table}__button--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
+  --#{$table}__button--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
+  --#{$table}__button--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
```
mcoker, line 93: This is good since this is an actual button that's meant to mimic our plain button 👍

src/patternfly/components/Table/table.scss
```diff
@@ -117,8 +117,8 @@
   --#{$table}__tr--m-ghost-row--BackgroundColor: var(--pf-t--global--background--color--primary--default);
 
   // * Table action
-  --#{$table}__action--PaddingBlockStart: var(--pf-t--global--spacer--sm);
-  --#{$table}__action--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
+  --#{$table}__action--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
+  --#{$table}__action--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
```
mcoker: Same comment here as for the `__toggle` above. This would probably be better as a `calc()` of the regular table cell padding value minus the control--vertical--plain token.

src/patternfly/components/TreeView/tree-view.scss
```diff
@@ -66,9 +66,9 @@ $pf-v6-c-tree-view--MaxNesting: 10 !default;
 
   // Toggle icon
   --#{$tree-view}__node-toggle-icon--MinWidth: var(--pf-t--global--icon--size--font--body--default);
-  --#{$tree-view}__node-toggle--PaddingBlockStart: var(--pf-t--global--spacer--sm);
+  --#{$tree-view}__node-toggle--PaddingBlockStart: var(--pf-t--global--spacer--control--vertical--plain);
   --#{$tree-view}__node-toggle--PaddingInlineEnd: var(--pf-t--global--spacer--md);
-  --#{$tree-view}__node-toggle--PaddingBlockEnd: var(--pf-t--global--spacer--sm);
+  --#{$tree-view}__node-toggle--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--plain);
```
mcoker, line 71: It looks like we probably want to update the node padding, too? https://github.com/patternfly/patternfly/blob/716f0e546cb9da7dade2907e6ad1f32584da027b/src/patternfly/components/TreeView/tree-view.scss#L14-L17

mcoker (changes requested):
Some others I found that we should update or ask design about

* Accordion toggle - inline padding still uses `spacer--md`
  * If we're updating accordion and nav, should other things like them be updated?
    * Menu items
    * Dual list selector items
    * Tree view items/nodes
    * Wizard nav/step links
* Toggle group buttons need semantic spacers (regular and compact)
* Random but I think this could use the text-to-element token https://github.com/patternfly/patternfly/blob/716f0e546cb9da7dade2907e6ad1f32584da027b/src/patternfly/components/ToggleGroup/toggle-group.scss#L30

src/patternfly/components/Table/table.scss
jcmill: Cool. I'll revert and create an issue. I think you're "probably what I would do" is a good approach.

src/patternfly/components/Table/table.scss
jcmill: I'll revert here too and bring it up with a new issue.

mcoker (approved):
L🔥TM

PR #8513, opened 2026-07-21
chore: update visual regression screenshots

mcoker:
https://drive.google.com/file/d/1wwseMb5-oWsCNMGjkYNmVVpX5BHXeXf5/view?usp=sharing

Differences noted:
* Tree view icons
* There are a couple instances of titles wrapping in weird ways, but the wrapping is correct and fixes an issue we introduced in 6.5. In 6.5 we [updated the title component's](https://github.com/patternfly/patternfly/pull/8438/changes) (and others) `word-break: break-word` to `overflow-wrap: break-word`, which is wrong - it should have been updated to `overflow-wrap: anywhere`, which is the equivalent of `word-break: break-word`. Here's the PR that fixed it - https://github.com/patternfly/patternfly/pull/8487. If you test 6.4 (or just change `overflow-wrap: anywhere` back to `word-break: break-word`) and look at a card with a couple of actions and a long title, you'll see it wraps the same way it does in the screenshots. This is for URLs and long words in other languages that are common in patternfly. Here are screenshots of before and after, and you can see in the before screenshot where the word didn't wrap, the card layout was making the page content bleed into the chrome spacing since it wasn't wrapping.

Screenshot 2026-07-21 at 9 12 40 AM

Screenshot 2026-07-21 at 9 12 45 AM


jcmill (approved):
Looks great.

PR #8536, opened 2026-07-30
fix: disabled background flicker

### LINKED ISSUE #8535: Bug - Button - background flickers when removing disabled
You can manually trigger it by togglinng the `disabled` attribute to a plain button, but it's easy to see in the [pagination examples](https://www.patternfly.org/components/pagination) by pressing the "next" button - the previous buttons will enable and show the issue.

---

**Jira Issue:** [PF-4515](https://redhat.atlassian.net/browse/PF-4515)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8535

The but comes from the click transition being removed on a disabled button. When the button goes from disabled to enabled, the transition is added back and runs.

This PR keeps the transition on disabled buttons, but the transition is disabled because the `background-position` change on `:active` (that gives the ripple effect) is overridden by the `background` declaration on `:disabed/.pf-m-disabled/.pf-m-aria-disabled`.

Screenshot 2026-08-03 at 3 39 17 PM

## Bug Fixes

- Improved button click ripple rendering across standard and disabled button states.
- Preserved button background colors while displaying the click interaction effect consistently.
- Ensured the click effect layers correctly with button themes and disabled styling.

### DIFF
--- a/src/patternfly/components/Button/button.scss
+++ b/src/patternfly/components/Button/button.scss
@@ -15,6 +15,7 @@
   --#{$button}--FontWeight: var(--pf-t--global--font--weight--body--default);
   --#{$button}--FontSize: var(--pf-t--global--font--size--body--default);
   --#{$button}--BackgroundColor: transparent;
+  --#{$button}--background--click: radial-gradient(circle, transparent 1%, color-mix(in srgb, currentcolor 15%, transparent) 2%) center/15000% 15000%;
   --#{$button}--BorderColor: transparent;
   --#{$button}--BorderWidth: var(--pf-t--global--border--width--action--default);
   --#{$button}--BorderRadius: var(--pf-t--global--border--radius--pill);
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Button/button.scss
```diff
@@ -15,6 +15,7 @@
   --#{$button}--FontWeight: var(--pf-t--global--font--weight--body--default);
   --#{$button}--FontSize: var(--pf-t--global--font--size--body--default);
   --#{$button}--BackgroundColor: transparent;
+  --#{$button}--background--click: radial-gradient(circle, transparent 1%, color-mix(in srgb, currentcolor 15%, transparent) 2%) center/15000% 15000%;
```
thatblindgeye: Should background be capitalized as `--Background`?

mcoker: Yeah probably. I lower cased it since it's less of a 1:1 to a background property, but it will probably always be a background style and it looks weird not capitalized.

thatblindgeye (approved):

jcmill (approved):
U Dah 💣

Issue #8511, opened 2026-07-20
Bug - Tabs aren't flush with edge of container


mcoker:
When scroll buttons are in the DOM and hidden, there is a small gap on either side of the tabs list that keeps the tabs from being flush with the edge of their container.

![image](https://github.com/user-attachments/assets/b15bbcda-1cda-4a66-a7c6-524c7609a85c)

Looks like the problem is because this var is supposed to be the width of the scroll button and it isn't. It's used as a negative margin that's animated to  expand the tabs list out when the scroll buttons are hidden. https://github.com/patternfly/patternfly/blob/5ec242ce60ad35365325cee7171e9b5077f6bca3/src/patternfly/components/Tabs/tabs.scss#L164

We can probably do that a better way. A couple of ideas:

- Simplest fix is probably just updating that var with a `calc()` of all of the things in the button, to get an accurate width.

- Ideally, I think we would optimize the animation performance and only animate the scroll button's position using `translate`, and let the tabs list resize immediately. We would need to run that by design though because the animation would change a little.

---

**Jira Issue:** [PF-4461](https://redhat.atlassian.net/browse/PF-4461)

PR #8540, opened 2026-07-30
chore: update visual regression screenshots

mcoker:
Updates I spotted:
* Tabs height changed for regular and subtabs
* Small gap on the left/right edge of tabs removed for tabs without scroll buttons. I opened a separate issue last week to fix that - https://github.com/patternfly/patternfly/issues/8511
* Primary detail updates
* "Reset" buttons at the end of some forms changed to "Cancel"
* In the dark/high-contrast diff, some sticky bottom paginations on mobile (like https://pf-pr-8540.surge.sh/components/data-list/html-demos/basic) show a background color diff but what's there looks correct (floating bg token). I'm assuming those didn't get picked up last time we updated visual regressions

**Full report - https://drive.google.com/file/d/1wwseMb5-oWsCNMGjkYNmVVpX5BHXeXf5/view?usp=sharing**


jcmill (approved):
LGTM! Looks like glass was affected by the background color on the pagination too.

PR #8542, opened 2026-08-04
feat(docked nav): add expandable nav

### LINKED ISSUE #8478: Docked nav: Add expandable nav items to Docked nav
### Figma Link

https://www.figma.com/design/VMEX8Xg2nzhBX8rfBx53jp/PatternFly-6--Styles---Components?m=auto&node-id=46511-38350&t=0ZNYJuk0dbFcwBhy-1

### Jira/GitHub Ticket Link

https://redhat.atlassian.net/browse/PF-4161

### Design Screenshot

Image

Image

Image

Option C in this [prototype](https://pf-docked-subnav-prototype.vercel.app/) gives an overview of the desired outcome. Clicking on the Expandable nav item opens to full menu allowing users to navigate across multiple sections instead of one flyout at a time. When the user chooses a nav item the docked nav will collapse. 

---

**Jira Issue:** [PF-4394](https://redhat.atlassian.net/browse/PF-4394)

jcmill:
Closes: #8478

Add expandable nav items to the docked nav (page and compass), including a new `.pf-m-expandable-expanded` class as an overlay (with a backdrop) that stays open across breakpoint.

Three demos show the closed, open with `.pf-m-expandable-expanded`, and open with `.pf-m-text-expanded`.

## Summary by CodeRabbit

- **New Features**
  - Added expandable docked navigation with nested subnavigation and expandable-link indicators.
  - Added expanded, expandable-expanded, and text-expanded states across page, compass, masthead, menu toggle, and navigation components.
  - Added expandable docked navigation support in masthead templates.
  - Added examples covering expandable, expanded, and text-expanded navigation variations.

- **Documentation**
  - Expanded Compass, Navigation, Page, and demo documentation with dock modifiers, usage guidance, and examples.


src/patternfly/components/Masthead/masthead.scss
```diff
@@ -210,7 +210,7 @@ $pf-v6-c-masthead--inset-map: build-spacer-map("none", "sm", "md", "lg", "xl", "
       }
     }
 
-    @at-root :is(.#{$page}__dock, .#{$compass}__dock).pf-m-text-expanded & {
+    @at-root :is(.#{$page}__dock, .#{$compass}__dock).pf-m-text-expanded &, :is(.#{$page}__dock, .#{$compass}__dock).pf-m-expandable-expanded & {
```
mcoker: Looks like you've been grouping these using `:is()`, can you do that here, too?

```suggestion
    @at-root :is(.#{$page}__dock, .#{$compass}__dock):is(.pf-m-text-expanded, .pf-m-expandable-expanded) & {
```

src/patternfly/components/Nav/nav-link-expandable-icon.hbs
```diff
@@ -0,0 +1,6 @@
+<span class="{{pfv}}nav__link-expandable-icon{{#if nav-link-icon--modifier}} {{nav-link-icon--modifier}}{{/if}}"
+  {{#if nav-link-icon--attribute}}
+    {{{nav-link-icon--attribute}}}
```
mcoker:
```suggestion
<span class="{{pfv}}nav__link-expandable-icon{{#if nav-link-expandable-icon--modifier}} {{nav-link-expandable-icon--modifier}}{{/if}}"
  {{#if nav-link-expandable-icon--attribute}}
    {{{nav-link-expandable-icon--attribute}}}
```

src/patternfly/components/Nav/nav-link.hbs
```diff
@@ -5,6 +5,7 @@
     nav-link--current='pf-m-current'
     nav-link--IsCurrent='pf-m-current'
     nav-link--IsHover='pf-m-hover'
+    nav-link--IsExpandable='pf-m-hover'
```
mcoker: Doesn't look like you need this? (also it applies the wrong class)

src/patternfly/components/Nav/nav.scss
```diff
@@ -54,7 +54,7 @@
   // * Nav item toggle icon
   --#{$nav}__item__toggle-icon--Rotate: 0;
   --#{$nav}__item--m-expanded__toggle-icon--Rotate: -180deg;
-  --#{$nav}__item__toggle-icon--TransitionDuration--transform: var(--pf-t--global--motion--duration--icon--default);
+  --#{$nav}__item__icon--TransitionDuration--transform: var(--pf-t--global--motion--duration--icon--default);
```
mcoker: looks like this may have been a mistake? 

```suggestion
  --#{$nav}__item__toggle-icon--TransitionDuration--transform: var(--pf-t--global--motion--duration--icon--default);
```

src/patternfly/components/Nav/nav.scss
```diff
@@ -199,6 +204,29 @@
...
+
+    .#{$nav}__toggle, .#{$nav}__subnav {
+      display: none;
+    }
+
+    .#{$nav}__link-expandable-icon {
+      position: absolute;
+      inset-block-end: calc(var(--#{$nav}--m-docked__link-expandable-icon--PaddingBlockEnd) * -1);
+      inset-inline-start: var(--#{$nav}--m-docked__link-expandable-icon--InsetInlineStart);
+      display: initial;
```
mcoker: is this needed?

src/patternfly/components/Nav/nav.scss
```diff
@@ -150,6 +150,11 @@
   --#{$nav}--m-docked__link--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$nav}--m-docked__link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
   --#{$nav}--m-docked__link--m-current--BackgroundColor: var(--pf-t--global--background--color--action--plain--clicked);
+  --#{$nav}--m-docked__link-expandable-icon--Color: var(--pf-t--global--icon--color--subtle);
+  --#{$nav}--m-docked__link-expandable-icon--PaddingBlockEnd: var(--pf-t--global--spacer--control--vertical--compact);
```
mcoker: Since this button doesn't use a compact spacer, we probably shouldn't use that token here. If someone were to theme `--pf-t--global--spacer--control--vertical--compact`, this icon position shouldn't change.

I tried to think about how to predictably position those ellipsis, if the icon size and/or button/padding changed, and I'm not sure there is a single way. It depends on how the button scales as it gets bigger I guess. For this I don't know how much it matters for it to be perfect. You could use a pixel value if you wanted, or half of the `--pf-v6-c-nav__link--PaddingBlockEnd` var (since that is the same as the compact var you were using) as long as the icon's still positioned if someone updates the padding and icon/size.

src/patternfly/components/Nav/nav.scss
```diff
@@ -199,6 +204,29 @@
...
+      position: absolute;
+      inset-block-end: calc(var(--#{$nav}--m-docked__link-expandable-icon--PaddingBlockEnd) * -1);
+      inset-inline-start: var(--#{$nav}--m-docked__link-expandable-icon--InsetInlineStart);
+      display: initial;
+      color: var(--#{$nav}--m-docked__link-expandable-icon--Color);
+      transform: translateX(var(--#{$nav}--m-docked__link-expandable-icon--TransformX));
+
+      svg {
+        scale: var(--#{$nav}--m-docked__link-expandable-icon--Scale);
+      }
```
mcoker: Can you move this to the `__link-expandable-icon`? Typically we don't style by element since we don't know what could always go there. Font awesome icons are still valid (`<i>` elements by default)

src/patternfly/components/Nav/nav.scss
```diff
@@ -199,6 +204,29 @@
...
+      display: none;
+    }
+
+    .#{$nav}__link-expandable-icon {
+      position: absolute;
+      inset-block-end: calc(var(--#{$nav}--m-docked__link-expandable-icon--PaddingBlockEnd) * -1);
+      inset-inline-start: var(--#{$nav}--m-docked__link-expandable-icon--InsetInlineStart);
+      display: initial;
+      color: var(--#{$nav}--m-docked__link-expandable-icon--Color);
+      transform: translateX(var(--#{$nav}--m-docked__link-expandable-icon--TransformX));
```
mcoker: Can you use `translate` here? Unless you need to control the order of the transform functions or something like that, we'll typically write these out using their own properties

src/patternfly/components/Nav/nav.scss
```diff
@@ -199,6 +204,29 @@
...
+
+      svg {
+        scale: var(--#{$nav}--m-docked__link-expandable-icon--Scale);
+      }
+    }
+
+    .#{$nav}__item:has(> .#{$nav}__subnav .#{$nav}__link.pf-m-current) {
+      --#{$nav}__item--accent--ScaleY: var(--#{$nav}__item--current--accent--ScaleY);
+      --#{$nav}__item--accent--TransitionDuration: var(--#{$nav}__item--current--accent--TransitionDuration);
+      --#{$nav}__link--BackgroundColor: var(--pf-v6-c-nav--m-docked__link--hover--BackgroundColor);
```
mcoker: I think this is the token you want here?

```suggestion
      --#{$nav}__link--BackgroundColor: var(--#{$nav}--m-docked__link--m-current--BackgroundColor);
```

src/patternfly/components/Nav/nav.scss
```diff
@@ -400,7 +454,7 @@
 // - Nav link icon
 .#{$nav}__link-icon {
   min-width: 1lh;
-  color: var(--#{$nav}__link-icon--Color);
+  color: inherit;
```
mcoker: Not sure if you meant to remove this? Icons use different color tokens, and this makes it use the text color token. Also it's breaking for the "current" state where this color is updated - https://github.com/patternfly/patternfly/blob/0464bb81e1b668f2e15697c90915dcb228d25a3e/src/patternfly/components/Nav/nav.scss#L446

Though it does seem like we have a bug - on hover/focus, the icon color doesn't change, but the text does and the icon color changes in figma, too. So I would probably just update it so the icon color changes on hover/focus, too, unless there was a different reason you removed the var.

jcmill: That's exactly why I made this change. But I see your point and it should the set to the icon version and not just inherit the text version. Thanks for catching that.

src/patternfly/components/Page/page.scss
```diff
@@ -19,6 +19,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__dock--TransitionDuration--slide: 0s;
   --#{$page}__dock--m-expanded--TransitionDuration--slide: 0s;
   --#{$page}__dock--TransitionTimingFunction--slide: var(--pf-t--global--motion--timing-function--decelerate);
+  --#{$page}__dock--m-expandable-expanded--ZIndex: var(--pf-t--global--z-index--lg);
```
mcoker, line 22: Doesn't look like this is being used

src/patternfly/components/Nav/nav.scss
```diff
@@ -150,6 +151,11 @@
   --#{$nav}--m-docked__link--PaddingInlineEnd: var(--pf-t--global--spacer--sm);
   --#{$nav}--m-docked__link--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
   --#{$nav}--m-docked__link--m-current--BackgroundColor: var(--pf-t--global--background--color--action--plain--clicked);
+  --#{$nav}--m-docked__link-expandable-icon--Color: var(--pf-t--global--icon--color--subtle);
+  --#{$nav}--m-docked__link-expandable-icon--InsetBlockStart: 50%;
+  --#{$nav}--m-docked__link-expandable-icon--InsetInlineStart: 50%;
+  --#{$nav}--m-docked__link-expandable-icon--TransformX: -50%;
+  --#{$nav}--m-docked__link-expandable-icon--Scale: 80%;
```
mcoker, line 158: Curious why there are vars for some of these but not others? IMO the only vars we probably need are the `scale` and probably the `.75em` as `--#{$nav}--m-docked__link-expandable-icon--offset` or something, so users could adjust the scale and the vertical position relative to the main icon. Curious what you think about that?

Also I don't suppose it matters, but we use numbers for `scale`, so `.8` instead of `80%`. They're both valid so it doesn't really matter, but it would be more consistent to use `.8`

src/patternfly/components/Nav/nav.scss
```diff
@@ -208,14 +233,40 @@
...
+        --#{$nav}__link--BackgroundColor: initial;
+        --#{$nav}__item--accent--ScaleY: 0;
+        --#{$nav}__link--m-current--BackgroundColor: initial;
+        
+        .#{$nav}__item:has(> .#{$nav}__link.pf-m-current) {
+          --#{$nav}__link--BackgroundColor: initial;
+          --#{$nav}__item--accent--ScaleY: var(--#{$nav}__item--current--accent--ScaleY);
+          --#{$nav}__link--m-current--BackgroundColor: var(--pf-v6-c-nav--m-docked__link--hover--BackgroundColor);
+        }
+      }
```
mcoker, line 268: It seems like here and [here](https://github.com/patternfly/patternfly/blob/021c5ac23c8aa2d47e2a927ab0e08b4927de0c27/src/patternfly/components/Nav/nav.scss#L222-L226), we're duplicating logic that already exists. I could be missing something but at first glance, it seems like we are. I tried this code (diff below) out and it seems to work. It is probably less readable (long selectors) but it seemed to work. The docked expandable examples looked good and I ran visual regressions and it didn't break anything else.

Not saying we should use it but curious what you think?

```diff
diff --git a/src/patternfly/components/Nav/nav.scss b/src/patternfly/components/Nav/nav.scss
index ce8f37103..ae1f3dbbd 100644
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -219,12 +219,6 @@
       translate: -50% calc(-50% + .75em);
     }
 
-    .#{$nav}__item:has(> .#{$nav}__subnav .#{$nav}__link.pf-m-current) {
-      --#{$nav}__item--accent--ScaleY: var(--#{$nav}__item--current--accent--ScaleY);
-      --#{$nav}__item--accent--TransitionDuration: var(--#{$nav}__item--current--accent--TransitionDuration);
-      --#{$nav}__link--BackgroundColor: var(--#{$nav}--m-docked__link--m-current--BackgroundColor);
-    }
-  
     @media (min-width: #{$pf-v6-global--breakpoint--dock--desktop}) {
       width: fit-content;
 
@@ -252,21 +246,6 @@
         display: revert;
       }
     }
-
-    @at-root :is(.#{$page}__dock, .#{$compass}__dock):is(.pf-m-text-expanded, .pf-m-expandable-expanded) &, 
-    &:is(.pf-m-text-expanded, .pf-m-expandable-expanded) {
-      .#{$nav}__item:has(> .#{$nav}__subnav .#{$nav}__link.pf-m-current) {
-        --#{$nav}__link--BackgroundColor: initial;
-        --#{$nav}__item--accent--ScaleY: 0;
-        --#{$nav}__link--m-current--BackgroundColor: initial;
-        
-        .#{$nav}__item:has(> .#{$nav}__link.pf-m-current) {
-          --#{$nav}__link--BackgroundColor: initial;
-          --#{$nav}__item--accent--ScaleY: var(--#{$nav}__item--current--accent--ScaleY);
-          --#{$nav}__link--m-current--BackgroundColor: var(--pf-v6-c-nav--m-docked__link--hover--BackgroundColor);
-        }
-      }
-    }
   }
 
   .#{$menu} {
@@ -374,6 +353,7 @@
     }
   }
 
+  @at-root :is(.#{$page}__dock, .#{$compass}__dock):not(.pf-m-text-expanded, .pf-m-expandable-expanded) &:has(.#{$nav}__subnav .#{$nav}__link.pf-m-current),
   &:has(> .#{$nav}__link.pf-m-current) {
     --#{$nav}__item--accent--ScaleY: var(--#{$nav}__item--current--accent--ScaleY);
     --#{$nav}__item--accent--TransitionDuration: var(--#{$nav}__item--current--accent--TransitionDuration);
@@ -439,6 +419,7 @@
     background-color: var(--#{$nav}__link--hover--BackgroundColor);
   }
 
+  @at-root :is(.#{$page}__dock, .#{$compass}__dock):not(.pf-m-text-expanded,.pf-m-expandable-expanded) .#{$nav}__item:has(.#{$nav}__subnav .#{$nav}__link.pf-m-current) > .#{$nav}__link,
   &.pf-m-current,
   &.pf-m-current:hover {
     --#{$nav}__link-icon--Color: var(--#{$nav}__link--m-current__link-icon--Color);
```

src/patternfly/demos/Masthead/masthead-template-content-toolbar-content.hbs
```diff
@@ -39,41 +39,108 @@
...
-        {{> menu-toggle menu-toggle--IsPlain=true menu-toggle--icon="rh-ui-thumbnail-view-large" menu-toggle--aria-label="Applications" menu-toggle--text="Applications" menu-toggle--IsDocked=true menu-toggle--IsText=true}}
-      {{/toolbar-item}}
-      {{#> toolbar-item}}
-        {{> button button--IsDocked=true button--IsPlain=true button--IsSettings=true button--aria-label="Settings" button--text="Settings"}}
-      {{/toolbar-item}}
-      {{#> toolbar-item}}
-        {{> menu-toggle menu-toggle--IsPlain="true" menu-toggle--aria-label="Help" menu-toggle--icon="rh-ui-question-mark-circle" menu-toggle--text="Help" menu-toggle--IsDocked=true menu-toggle--IsText=true}}
-      {{/toolbar-item}}
-    {{/toolbar-group}}
+    {{#if masthead-template--HasExpandableDockedNav}}
```
mcoker, line 42: Just a small nit, but if these navs are identical other than the expandable one replaces the middle 2 items with expandable items, we could just conditionally render those 2 items instead of the whole thing so that there is less repeated code. This could also go into its own template if it's a lot of code, and the horizontal nav could, too. Like `masthead-template-horizontal-nav.hbs` and `masthead-template-docked-nav.hbs` and the docked nav hbs can just take an argument for whether it's expandable or not. Up to you if you think either of these are worth it, just wanted to mention.

src/patternfly/components/Nav/nav.scss
```diff
@@ -199,6 +205,25 @@
     --#{$nav}__link--m-current--BackgroundColor: var(--#{$nav}--m-docked__link--m-current--BackgroundColor);
 
     width: 100%;
+
+    .#{$nav}__toggle, .#{$nav}__subnav {
+      display: none;
+    }
```
mcoker, line 211: I wonder if you could add these selectors here since we're already hiding things like this for the docked version? https://github.com/patternfly/patternfly/blob/021c5ac23c8aa2d47e2a927ab0e08b4927de0c27/src/patternfly/components/Nav/nav.scss#L231-L233

src/patternfly/components/Nav/nav.scss
```diff
@@ -208,14 +233,40 @@
...
+    &:is(.pf-m-text-expanded, .pf-m-expandable-expanded) {
       width: 100%;
 
       .#{$nav}__link-text {
         display: revert;
       }
     }
+
+    @at-root :is(.#{$page}__dock, .#{$compass}__dock):is(.pf-m-text-expanded, .pf-m-expandable-expanded) &, 
+    &:is(.pf-m-text-expanded, .pf-m-expandable-expanded) {
```
mcoker, line 246: if I'm not mistaken, it looks like this selector is written 3 times in this file. [Just above this block](https://github.com/patternfly/patternfly/blob/021c5ac23c8aa2d47e2a927ab0e08b4927de0c27/src/patternfly/components/Nav/nav.scss#L236-L243), which was before this PR, this one, then [again below this block](https://github.com/patternfly/patternfly/blob/021c5ac23c8aa2d47e2a927ab0e08b4927de0c27/src/patternfly/components/Nav/nav.scss#L256-L268) which was added in this PR. 

Could those be combined with the pre-existing selector?

src/patternfly/components/Nav/nav.scss
```diff
@@ -208,14 +233,40 @@
...
+      .#{$nav}__link-expandable-icon {
+        display: none;
+      }
+
+      .#{$nav}__toggle, .#{$nav}__subnav {
+        display: revert;
+      }
```
mcoker, line 253: If you combine the 3 selectors mentioned above into 1, could you also add these to the link-text rule that sets `display: revert`? https://github.com/patternfly/patternfly/blob/021c5ac23c8aa2d47e2a927ab0e08b4927de0c27/src/patternfly/components/Nav/nav.scss#L240-L242

src/patternfly/components/Nav/nav.scss
jcmill: I agree. Those are probably the only tokens users would need to be changed at this point.

src/patternfly/components/Nav/nav.scss
jcmill: I'm cool with that change. The only difference I see is how the icon color actually changes on hover when the nav is docked. Before, the icon at the top level (when a subnav item was set to current) was the lighter grey color and you could still hover it to change the color. This change locks that top level icon to the active color as if it were set to current. *Edit - I should add that I think this is more inline with the expected behavior and a great change.

lboehling (changes requested):
Thanks Jake! I'm seeing an unexpected second gray opacity layer with sharp edges on all of the selected docked nav items (not just expandable). Can we remove that? Otherwise the changes look good to me!

Image
Image

mcoker (approved):
🚀 NICE! Visual regressions all passed.

lboehling (approved):
wooh! 🎉

PR #8543, opened 2026-08-04
fix: make entire label clickable

### LINKED ISSUE #8430: Bug - Label - entire label isn't clickable
**Describe the problem**

When a `<Label>` component has an `onClick` handler, the clickable area (the inner `<button>`) does not extend to the full visual bounds of the `<Label>` outline. The outer `<span class="pf-v6-c-label">` renders padding/border that is visually part of the label, but clicking on that outer region does not trigger the `onClick`.

```jsx
<Label onClick={() => console.log('clicked')}>Label 1</Label>
```

Rendered HTML:

```html
<span class="pf-v6-c-label pf-m-filled pf-m-clickable">
  <button class="pf-v6-c-label__content pf-m-clickable" type="button">
    <span class="pf-v6-c-label__text">Label 1</span>
  </button>
</span>
```

The `<button>` (`.pf-v6-c-label__content`) is smaller than the outer `<span>` (`.pf-v6-c-label`), so clicks in the gap between the button edge and the label outline are not handled.

**Expected behavior**

The entire visible area of the label (up to and including the outline/border) should be clickable.

No. Just annoyance.

**Screenshots**

DevTools showing the button's hit area (36.44 x 18) sits inside the label outline:

image

**What is your environment?**

- Mac OS

- Chrome

**Any other information?**

The root cause appears to be that the outer `<span>` carries padding/border that is not covered by the inner `<button>`. Either the button should be styled to fill the span completely, or the click handler should be placed on the outer element.

---

**Jira Issue:** [PF-4238](https://redhat.atlassian.net/browse/PF-4238)

mcoker:
fixes https://github.com/patternfly/patternfly/issues/8430

`pointer-events: none` was preventing the pseudo element from making the entire label clickable. Looks like it was added a while ago as part of the editable label enhancement https://github.com/patternfly/patternfly/pull/4097

After this change, I tested onclicks of all the action/link/removable variations, as well as editable and editable removable labels, and all of the clicks registered on the elements they're intended to. One side effect of this change is that the area above and to the right of the close button is now clickable, but the click event registers on the label, so I updated the close button padding/margin to extend to the top/bottom/right edges of the label.

For testing the close button clickable area, you can add something like this

```scss
.pf-v6-c-label .pf-v6-c-label__actions .pf-v6-c-button {
  background: red;
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved removable label close-button accessibility attributes and disabled-state handling.
  * Corrected action-button spacing, padding, and corner styling within labels.
  * Improved pointer interaction with label elements and compact label layouts.

### DIFF
--- a/src/patternfly/components/Label/label.hbs
+++ b/src/patternfly/components/Label/label.hbs
@@ -31,8 +31,14 @@
   {{/label-content}}
   {{#if label--IsRemovable}}
     {{#> label-actions}}
-      {{> button button--IsDisabled=label--IsDisabled button--IsIcon=true button--icon="rh-microns-close" button--IsPlain=true button--HasNoPadding=true button--attribute=(concat 'id="' label--id '-button"
-    aria-label="Remove" aria-labelledby="' label--id '-button ' label--id '-text"')}}
+      {{> button
+          button--IsDisabled=label--IsDisabled
+          button--IsIcon=true button--icon="rh-microns-close"
+          button--IsPlain=true button--HasNoPadding=true
+          button--id=(concat label--id '-button')
+          button--aria-label="Remove"
+          button--aria-labelledby=(concat label--id '-button ' label--id '-text')
+      }}
     {{/label-actions}}
   {{/if}}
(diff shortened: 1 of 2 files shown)


jcmill (approved):
🟩

PR #8557, opened 2026-08-14
fix(nav): update html examples to display indicator correctly.

### LINKED ISSUE #8489: Docked nav: HTML example has no container
---

**Jira Issue:** [PF-4354](https://redhat.atlassian.net/browse/PF-4354)

jcmill:
Fixes: #8489

Adds a margin to docked nav and other nav HTML examples so the current-page indicator isn’t clipped. Applies margin-inline-start on the preview container, and skips expanded, horizontal, and drilldown variants that don’t need it.

## Summary by CodeRabbit

* **Style**
  * Added consistent medium inline spacing to navigation previews.
  * Updated horizontal navigation previews with the secondary default background.
  * Improved alignment and visual consistency across navigation examples and previews.

### DIFF
--- a/src/patternfly/components/Nav/examples/Navigation.css
+++ b/src/patternfly/components/Nav/examples/Navigation.css
@@ -19,6 +19,8 @@
 }
 
 .ws-core-c-navigation .ws-preview-html {
+  padding-inline-start: var(--pf-t--global--spacer--md);
+  padding-inline-end: var(--pf-t--global--spacer--md);
   background: var(--pf-v6-c-page__sidebar--BackgroundColor); /* use sidebar background for vertical nav examples */
 }
 
@@ -28,8 +30,4 @@
 
 [id^="ws-core-c-navigation-horizontal-subnav"] .ws-preview-html {
   background: transparent; /* reset background for horizontal subnav examples */
-}
-
-.ws-core-c-navigation .ws-preview-html:not(:has(.pf-m-expanded, .pf-m-horizontal, .pf-m-drilldown)) {
-  margin-inline-start: var(--pf-t--global--spacer--md);
 }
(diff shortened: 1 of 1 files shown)


src/patternfly/components/Nav/examples/Navigation.css
```diff
@@ -29,3 +29,7 @@
 [id^="ws-core-c-navigation-horizontal-subnav"] .ws-preview-html {
   background: transparent; /* reset background for horizontal subnav examples */
 }
+
+.ws-core-c-navigation .ws-preview-html:not(:has(.pf-m-expanded, .pf-m-horizontal, .pf-m-drilldown)) {
+  margin-inline-start: var(--pf-t--global--spacer--md);
```
mcoker: Is there a reason this wasn't a padding? IMO the margin looks a little off. We could even add left/right padding to all of these examples with the secondary background color (would include `.pf-m-expanded`). That seems the most consistent.

Current style

Image

Using a padding instead of margin

Image

Updating to left/right padding and including the expanded example

Image

mcoker (commented):
Just a comment, tho I'll defer to whatever design thinks is best.

mcoker (commented):
This LGTM. What do you think about adding the background to horizontal (but not horizontal subnav)?

You can't see the hover/current background changes currently. Here's what it looks like now

Image

Here's with the background added

Image

src/patternfly/components/Nav/examples/Navigation.css
bekah-stephens: I agree that the grouped nav looks off (in the current style screenshot)

bekah-stephens: when I pull up the preview link, it actually looks fine?

mcoker: @bekah-stephens jake updated it so that it looks like the last example

mcoker (approved):

bekah-stephens (approved):

PR #8560, opened 2026-08-17
docs(misc): remove pf-d- prefix from demo docs

kmcfaul:
Removes the css prefix from a couple demo md files. These were appearing in their own CSS variables table with no content.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Back to top, Button, Dashboard, Description List, and Banner examples.
  * Removed obsolete styling metadata from these demos to keep their displayed configurations accurate and consistent.

### DIFF
--- a/src/patternfly/demos/BackToTop/examples/BackToTop.md
+++ b/src/patternfly/demos/BackToTop/examples/BackToTop.md
@@ -1,7 +1,6 @@
 ---
 id: 'Back to top'
 section: components
-cssPrefix: pf-d-back-to-top
 ---
 
 ## Examples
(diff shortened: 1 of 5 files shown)


mcoker:
Can we also remove `cssPrefix` from the banner demo file? Probably isn't hurting anything, but then none of the demo markdown will have a `cssPrefix` defined https://github.com/kmcfaul/patternfly/blob/cd0398eb6348adf24b6e138f57312ed59c6b242a/src/patternfly/demos/Banner/examples/Banner.md?plain=1#L4

mcoker (approved):

PR #8566, opened 2026-08-25
feat(drawer): full height drawer with no content

### LINKED ISSUE #8504: Spike - Investigate best option for drawer

This spike is to test the best option for a drawer that will work wrapping this page and be content height-agnostic. We just want the drawer to be the viewport size and allow the page to scroll behind it.

Options:

- Get rid of drawer content - Michael is leaning towards this being more likely

- Telling the content not to scroll and probably also letting the panel be height independent from content

---

**Jira Issue:** [PF-4446](https://redhat.atlassian.net/browse/PF-4446)

jcmill:
Closes: #8504

Based on the approaches discussed in #8504, this PR implements Option 1: removing `.pf-v6-c-drawer__content` from the page-level drawer so the panel can act as a standalone overlay rather than a layout participant sized by sibling content. At the md breakpoint and up, .pf-v6-c-page__drawer becomes position: fixed and the panel fills the full browser height, sitting above the page while pointer events pass through to the content scrolling behind it. Below md, the drawer remains in normal document flow.

## Summary by CodeRabbit

- **New Features**
  - Added support for drawers without dedicated content areas, rendering as a full-viewport overlay.
  - Improved responsive drawer positioning and panel slide behavior across page layouts, including expanded/panel-left variants.
- **Documentation**
  - Updated Drawer demo to use the Drawer component directly and marked it as beta.
  - Clarified Drawer content-container guidance (including the sibling placement pattern for overlay drawers).
  - Updated page templates to conditionally render main and drawer content when no drawer content is available.

### DIFF
--- a/src/patternfly/components/Drawer/drawer.scss
+++ b/src/patternfly/components/Drawer/drawer.scss
@@ -373,6 +373,13 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
       }
     }
   }
+
+  &:not(:has(.#{$drawer}__content)) {
+    > .#{$drawer}__main > .#{$drawer}__panel {
+      height: var(--#{$drawer}__panel--Height, auto);
+      margin-inline-start: var(--#{$drawer}__panel--MarginInlineStart, revert);
+    }
+  }
 }
 
 // Header area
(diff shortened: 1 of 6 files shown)


rebeccaalpert:
I'm going to go ahead and close this since we're not going in this direction, but thank you so much for all your work!

PR #8568, opened 2026-08-25
feat(drawer): full height drawer with modifier class

### LINKED ISSUE #8504: Spike - Investigate best option for drawer

jcmill:
Closes: #8504

Based on the approaches discussed in #8504, this PR's implementation is more inline with Option 2: keeping `.pf-v6-c-drawer__content` in place and instead giving the drawer panel its own height independent of the content container. This is implemented with the modifier .pf-m-full-height. The styles live in drawer.scss alongside  the Drawer's other panel modifiers, with Page managing the z-index elevation needed to sit above other page chrome. At the md breakpoint and up, the panel becomes position: fixed and spans the full browser height (100dvh), sliding in/out via transform independent of the drawer content's own height. Below md, the panel stays position: relative and behaves like a normal in-flow drawer panel.

## Summary by CodeRabbit

* **New Features**
  * Added viewport overlay support for drawers with fixed, full-height positioning.
  * Added directional drawer behavior and improved overlay interaction at medium and larger screen sizes.
  * Preserved support for expanded, inline, static, and panel-position variants.

* **Documentation**
  * Updated drawer guidance and examples to use the new viewport overlay option.
  * Simplified demonstrations by replacing multiple full-height drawer examples with a single viewport drawer example.


src/patternfly/components/Drawer/drawer.scss
```diff
@@ -6,6 +6,14 @@ $pf-v6-c-drawer--breakpoint-map--width: build-breakpoint-map("base", "lg", "xl",
 $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 
 @include pf-root($drawer) {
+  // Drawer
+  --#{$drawer}--Height: 100%;
+  --#{$drawer}--OverflowX: hidden;
+
+  // Main
+  --#{$drawer}__main--Display: flex;
```
mcoker: Is this needed? Doesn't look like it's ever changed.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -285,6 +299,55 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
...
+      --#{$drawer}--m-full-height__panel--InsetInlineStart: 0;
+      --#{$drawer}--m-full-height__panel--InsetInlineEnd: auto;
+    }
+
+    @media (min-width: $pf-v6-global--breakpoint--md) {
+      --#{$drawer}__panel--Position: fixed;
+      --#{$drawer}__panel--Height: 100%;
+      --#{$drawer}__panel--ZIndex: var(--pf-t--global--z-index--md);
+
+      .#{$drawer}__panel {
```
mcoker: when targeting any children, use the descendent selectors since drawers are often nested

```suggestion
      > .#{$drawer}__main > .#{$drawer}__panel {
```

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -285,6 +299,55 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
...
+
+      &.pf-m-panel-left .#{$drawer}__panel {
+        @include pf-v6-bidirectional-style(
+          $prop: transform,
+          $ltr-val: translateX(-100%),
+          $rtl-val: translateX(#{pf-v6-calc-inverse(-100%)}),
+        );
+      }
+
+      &.pf-m-expanded > .#{$drawer}__main .#{$drawer}__panel {
```
mcoker:
```suggestion
      &.pf-m-expanded > .#{$drawer}__main > .#{$drawer}__panel {
```

mcoker (commented):
From our review earlier. Also just a note that the pill variant isn't spaced properly currently.

rebeccaalpert:
Will this work for resizable drawers? I believe that's what we need. I'm also wondering about the backdrop - this method means we don't have one. I'm not 100% sure how necessary that backdrop is, but I can ask.

rebeccaalpert:
Screenshot 2026-09-01 at 5 15 35 PM

Plugged into the demo app with resizable drawer HTML and it seems like it works, even if the styles are a little bit different - my only question then is whether we need a backdrop. Hummingbird has one for what it's worth, but I don't see them in the Core examples or React examples.

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -249,8 +261,8 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
 
   display: flex;
   flex-direction: column;
-  height: 100%;
-  overflow-x: hidden;
+  height: var(--#{$drawer}--Height);
+  overflow-x: var(--#{$drawer}--OverflowX);
```
mcoker: You'll want a fallback, so that `initial` as the value for the CSS var applies the fallback. Otherwise the value will be invalid and the property will default to `unset` effectively, causing the value to be `initial` for non-inheritable properties, which can cause issues with certain properties.

```suggestion
  height: var(--#{$drawer}--Height, revert);
  overflow-x: var(--#{$drawer}--OverflowX, revert);
```

src/patternfly/components/Drawer/drawer.scss
```diff
@@ -396,7 +489,7 @@ $pf-v6-c-drawer__panel--list--width: (25, 33, 50, 66, 75, 100);
   display: flex;
   flex: 1;
   gap: var(--#{$drawer}__main--Gap, 0);
-  overflow: hidden;
+  overflow: var(--#{$drawer}__main--Overflow);
```
mcoker:
```suggestion
  overflow: var(--#{$drawer}__main--Overflow, revert);
```

src/patternfly/demos/Drawer/examples/Drawer.md
```diff
@@ -56,5 +56,20 @@ wrapperTag: div
 {{> drawer-jump-links drawer-jump-links--id="drawer-expanded-jump-links" drawer-panel--IsOpen="true"}}
 ```
 
+### Expanded resiable viewport drawer
+```hbs isFullscreen isBeta
+{{> page-template page-template--id="drawer-expanded-resiable-viewport-example"}}
+
+{{#> drawer drawer--id="drawer-expanded-resiable-viewport-example-drawer" drawer--IsViewport="true" drawer-panel--IsOpen="true" drawer-panel--IsResizable=true}}
```
rebeccaalpert:
```suggestion
{{#> drawer drawer--id="drawer-expanded-resizable-viewport-example-drawer" drawer--IsViewport="true" drawer-panel--IsOpen="true" drawer-panel--IsResizable=true}}
```

src/patternfly/demos/Drawer/examples/Drawer.md
```diff
@@ -56,5 +56,20 @@ wrapperTag: div
 {{> drawer-jump-links drawer-jump-links--id="drawer-expanded-jump-links" drawer-panel--IsOpen="true"}}
 ```
 
+### Expanded resiable viewport drawer
```
rebeccaalpert:
```suggestion
### Expanded resizable viewport drawer
```

lboehling (approved):
lgtm! this is very minor but if it is covering the masthead (and i think it's ok if it does in this use case), we might want the demo/example to show a close button at the top right to make it clear the drawer should be dismissible. 

It's hard to say re: the backdrop. If users can interact with the page content still, i'd say it's not needed. can it be optional based on use case?

rebeccaalpert:
Sounds good; we'll do an optional backdrop on the React side!

rebeccaalpert (approved):
Thank you!

src/patternfly/components/Drawer/examples/Drawer.md
```diff
@@ -336,7 +336,7 @@ import './Drawer.css'
 | `.pf-v6-c-drawer` | `<div>` | Initiates the drawer container. **Required** |
 | `.pf-v6-c-drawer__section` | `<div>` | Initiates a drawer section area. This element can be used above or below `.pf-v6-c-drawer__main` for titles, toolbars, footers, etc. |
 | `.pf-v6-c-drawer__main` | `<div>` | Initiates the drawer main area. **Required** |
-| `.pf-v6-c-drawer__content` | `<div>` | Initiates the drawer content container. **Required** |
+| `.pf-v6-c-drawer__content` | `<div>` | Initiates the drawer content container. **Required**. May be empty when `.pf-m-viewport` is used as a viewport overlay. |
```
mcoker: WDYT about either removing this (I don't know if it's necessary) or updating to say it has to be empty?

```suggestion
| `.pf-v6-c-drawer__content` | `<div>` | Initiates the drawer content container. **Required**. It is empty when used with `.pf-m-viewport`. |
```

mcoker: WDYT about either removing this (I don't know if it's necessary) or updating to say it has to be empty?

```suggestion
| `.pf-v6-c-drawer__content` | `<div>` | Initiates the drawer content container. **Required**. It is empty when used with `.pf-m-viewport`. |
```

src/patternfly/demos/Page/page-template-drawer.hbs
```diff
@@ -1,5 +1,5 @@
 {{#> page-drawer}}
-  {{#> drawer drawer--id=(concat page-template--id '-drawer') drawer-panel--IsOpen=page-template--drawer-panel--IsOpen}}
+  {{#> drawer drawer--id=(concat page-template--id '-drawer') drawer-panel--IsOpen=page-template--drawer-panel--IsOpen drawer--IsPanelLeft=drawer--IsPanelLeft}}
```
mcoker: Do you need this? Doesn't seem like it would do anything.

```suggestion
  {{#> drawer drawer--id=(concat page-template--id '-drawer') drawer-panel--IsOpen=page-template--drawer-panel--IsOpen}}
```

mcoker (approved):
🧑‍🚀🚀

PR #8570, opened 2026-08-26
fix(docked nav expandable): remove ellipsis on mobile

### LINKED ISSUE #8569: Bug - Docked Nav Expandable - Visible ellipsis on mobile
The ellipsis for expandable items in the docked nav are visible in the mobile view. Because the mobile view is already fully expanded by default, one option could be to set display none based on `.pf-m-expanded`. Similar to when `.pf-m-text-expanded` is applied. But should really look at setting the display to none by default for a mobile first approach (if feasible and makes sense).

---

**Jira Issue:** [PF-4575](https://redhat.atlassian.net/browse/PF-4575)

jcmill:
Fixes #8569

This fixes the expandable-item indicator icon (ellipsis) incorrectly showing in the docked nav on mobile view. Because the nav is fully expanded by default on mobile, this sets the display to none and reverts it in the desktop breakpoint (following a "mobile first" approach).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the docked navigation layout by hiding the expandable icon on smaller screens.
  - Restored the icon on desktop-width displays for a more appropriate responsive experience.

### DIFF
--- a/src/patternfly/components/Nav/nav.scss
+++ b/src/patternfly/components/Nav/nav.scss
@@ -208,6 +208,7 @@
       position: absolute;
       inset-block-start: 50%;
       inset-inline-start: 50%;
+      display: none;
       color: var(--#{$nav}--m-docked__link-expandable-icon--Color);
       scale: var(--#{$nav}--m-docked__link-expandable-icon--Scale);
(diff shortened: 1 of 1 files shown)


mcoker (approved):
👍

PR #8571, opened 2026-09-01
feat(page): adds plain variant

### LINKED ISSUE #8506: Spike - Investigate best option for page

The goal of this spike is to test the feasibility of moving the scrollbar in the PatternFly Page component.

Easy path may be:

- Update the page component to hold a non-PatternFly header and footer and use a grid gap layout

- Sample structure would look like:

- Page

- > PageHeader

- > PageContent

Hard path may be:

- New grid layout or just page main with a layout

---

**Jira Issue:** [PF-4449](https://redhat.atlassian.net/browse/PF-4449)

mcoker:
Fixes https://github.com/patternfly/patternfly/issues/8506 (JIRA https://redhat.atlassian.net/browse/PF-4449)

**PR preview - https://pf-pr-8571.surge.sh/components/page/html-demos/plain/**

The primary goal is to enable a layout similar to https://images.redhat.com/ that has
* Custom header
* Stacked sections with a gap between them that can optionally be wrapped in glass panels
* Custom footer

This demo only looks right in glass. Isn't clear what this should do in non-glass. Non-glass can be a follow up after this merges. 

New features
* Adds `.pf-m-plain` on `.pf-v6-c-page` for this new layout
* Adds `.pf-v6-c-page__footer` that defaults to `<footer>`. It's full-bleed and doesn't have padding. A RH global footer can be dropped inside. Compliments the new `.pf-v6-c-page__header`
  * Updated the page grid to support it behind `:has()` so we don't break anyone's existing grid.

Does these things
* Removes the main container styling (background color, border, padding, etc)
* Removes the height constraints on the page wrapper so it can grow beyond the viewport, and added a min-height of the viewport height.
* Removes overflow/scroll management from the content section so the window is what scrolls now
* Adds a gap on the main container to space the page sections. Gap token is `--pf-t--global--spacer--gutter--default`
* Removes padding from the breadcrumb and subnav sections. TBD on that, may add it back if the goal is to do something like group those in a panel with the page title.

Questions
* Should the footer always be at the bottom of the viewport or could it just end wherever the page content ends, even if that means the footer is floating in the middle of the viewport. 
* What should it look like in non-glass?
* Should the footer have any kind of default or optional styles? Background color, padding, divider, etc?
* Should we have a few demos with a few common layouts to show things like a toolbar + table, breadcrumbs, horizontal subnav, etc?
* Should we change chrome/layout spacers? Specifically the chrome around the main content area, gap between the header + content + footer, gap used between page sections.

## Summary by CodeRabbit

- **New Features**
  - Added footer support with configurable markup and styling.
  - Added a plain Page variant with simplified styling and spacing.
  - Added configurable Page header element types.
  - Added sidebar exclusion and custom footer options.
  - Added no-padding and width-limiting options for main sections.
  - Added demos for plain pages, custom headers, footers, and glass content.

- **Bug Fixes**
  - Improved expandable status card rendering to preserve intended glass styling.

- **Documentation**
  - Updated Page examples and usage tables to reflect beta features and new layout options.

### DIFF
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -429,6 +429,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
     --#{$page}__main-container--MarginBlockEnd: var(--#{$page}--m-plain__main-container--MarginBlockEnd);
 
     height: revert;
+    min-height: 100dvh;
     max-height: revert;
 
     :is(.#{$page}__main, .#{$page}__main-container) {
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -334,7 +334,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
         "header header"
         "sidebar main"
         "footer footer";
-      grid-template-rows: max-content 1fr auto;
     }
   }
 
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -267,6 +267,9 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}--m-plain__main-container--MarginBlockEnd: var(--#{$page}--inset);
   --#{$page}--m-plain__main--RowGap: var(--pf-t--global--spacer--gutter--default);
 
+  // Footer
+  --#{$page}__footer--ZIndex: var(--#{$page}__main-container--ZIndex);
+
   :where(.pf-v6-theme-glass) & {
     --#{$page}--BackgroundColor: var(--#{$page}--BackgroundColor--glass);
     --#{$page}__sidebar--Width--base: var(--#{$page}__sidebar--Width--base--glass);
@@ -476,7 +479,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 
 // Footer
 .#{$page}__footer {
-  z-index: var(--#{$page}__footer-ZIndex);
+  z-index: var(--#{$page}__footer--ZIndex);
   grid-area: footer;
 }
 
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -476,7 +476,7 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
 
 // Footer
 .#{$page}__footer {
-  z-index: var(--#{$page}__header--ZIndex);
+  z-index: var(--#{$page}__footer-ZIndex);
   grid-area: footer;
 }
 
--- a/src/patternfly/components/Page/page.scss
+++ b/src/patternfly/components/Page/page.scss
@@ -172,7 +172,6 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$page}__main-section--m-secondary--BorderBlockEndWidth: var(--pf-t--global--border--width--glass--default, var(--pf-t--global--border--width--high-contrast--regular));
   --#{$page}__main-section--m-secondary--BorderBlockEndColor: var(--pf-t--global--border--color--subtle);
 
-
   // Limit width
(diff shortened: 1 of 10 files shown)


src/patternfly/components/Page/page.scss
```diff
@@ -311,13 +313,29 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   max-height: 100%;
   background-color: var(--#{$page}--BackgroundColor);
 
+  &:has(> .#{$page}__footer) {
+    grid-template-areas:
+      "header"
+      "main"
+      "footer";
+    grid-template-rows: max-content 1fr auto;
+  }
```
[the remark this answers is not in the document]
mcoker: I'm not concerned about docked nav for now. We can support that if someone requests it.

src/patternfly/components/Page/page.scss
```diff
@@ -404,6 +422,35 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
       @include pf-v6-hamburger($collapse: true);
     }
   }
+
+  &.pf-m-plain {
+    --#{$page}__main-container--MarginBlockEnd: var(--#{$page}--m-plain__main-container--MarginBlockEnd);
+
+    height: revert;
+    max-height: revert;
```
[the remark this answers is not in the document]
mcoker: That is OK. The short height is fine for the time being. In the PR description I have this under "questions" (for design)

> Should the footer always be at the bottom of the viewport or could it just end wherever the page content ends, even if that means the footer is floating in the middle of the viewport.

If design says yes, then I will add `min-height: 100dvh;` to `.pf-v6-c-page.pf-m-plain`

rebeccaalpert (commented):
I would add that min-height personally; I plugged most of the changes into the demo app and I can't imagine them accepting the footer where it is:

Image

Everything else looks good to me though.

mcoker:
@rebeccaalpert I agree personally - I had it in initially, but looked at redhat.com and images.redhat.com and they both have floating footers when there is not enough content to push the footer to the bottom of the page so I removed it.

That doesn't mean it's an intentional decision on RH/images though, since all of their pages may have enough content that it never triggers this issue, but that's why I left it off and flagged it as a question. I think @lboehling is going to take a look soon so she can guide us in the right direction.

Screenshot 2026-09-01 at 4 08 49 PM

Screenshot 2026-09-01 at 4 09 26 PM

src/patternfly/components/Page/page.scss
```diff
@@ -311,13 +313,29 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
...
       "header header"
       "sidebar main";
     grid-template-columns: var(--#{$page}__sidebar--Width) 1fr;
+
+    &:has(> .#{$page}__footer) {
+      grid-template-areas:
+        "header header"
+        "sidebar main"
+        "footer footer";
+      grid-template-rows: max-content 1fr auto;
```
jcmill: Unless you feel this needs to be explicitly stated, it could be considered duplicate code from line 321.

src/patternfly/components/Page/page.scss
```diff
@@ -428,6 +475,12 @@ $pf-page-v6--height-breakpoint-map: build-breakpoint-map("base", "sm", "md", "lg
   --#{$masthead}--m-display-inline--GridTemplateColumns: var(--#{$page}--m-dock--c-masthead--m-display-inline--GridTemplateColumns);
 }
 
+// Footer
+.#{$page}__footer {
+  z-index: var(--#{$page}__header--ZIndex);
```
jcmill: Out of my own curiosity, would we not set `--#{$page}__footer--ZIndex: var(--#{$page}__header--ZIndex);`?

```suggestion
  z-index: var(--#{$page}__footer-ZIndex);
```

mcoker: Oops I left that on accident. Committed your suggestion and updated it with a value of the `__main-container` z-index since I don't think the main/footer areas need a particular stacking order and can overlap one another.

jcmill (approved):
👣

lboehling (changes requested):
I agree with @rebeccaalpert about the minheight. I can't imagine the floating footer is intentional. 

also this is a small nit and probably demo specific styling (so feel free to treat it in a follow up), but it would be great if the page main sections behind the header and collapsible card were using the plain modifier, and if the panel behind the header had pf-m-bordered applied like so: 

Image 

otherwise looks good!

rebeccaalpert (approved):
💃

lboehling (approved):
🎉

PR #8577, opened 2026-09-04
fix(table): hover and selected background color updated to action-plan tokens

### LINKED ISSUE #8562: Bug - Modal - Modal background color is the same as Table hover and selected background colors in dark mode
This is an issue with Patternfly 6.6.1.

We have a Table in a Modal that has hoverable and selectable rows.  This looks like this in light mode:

Image

The "Filesystem" in the sidebar and the "boot" row are selected (with the `isRowSelected` property in React), and the "etc" row is being hovered over.

In dark mode, the exact same situation looks like this:

Image

Both the hover color and the "selected" color are the same as the background color of the Modal itself, `--pf-t--global--dark--background--color--300`.  

---

**Jira Issue:** [PF-4567](https://redhat.atlassian.net/browse/PF-4567)

jcmill:
Fixes: #8562

Consulted with Design to change the hover/clicked states to action-plain tokens respectively in order to cover different backgrounds on the table component.

## Summary by CodeRabbit

* **Style**
  * Updated table control-row, expandable, hover, clickable, expanded, and selected-state backgrounds for more consistent styling.
  * Aligned clickable row hover and selected-row colors with global action color tokens.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -226,10 +226,10 @@
   // * Table tr clickable
   --#{$table}__tr--m-clickable--BackgroundColor: transparent;
   --#{$table}__tr--m-clickable--OutlineOffset: calc(-1 * var(--pf-t--global--spacer--xs));
-  --#{$table}__tr--m-clickable--hover--BackgroundColor: var(--pf-t--global--background--color--primary--hover);
+  --#{$table}__tr--m-clickable--hover--BackgroundColor: var(--pf-t--global--background--color--action--plain--hover);
 
   // * Table tr selected
-  --#{$table}__tr--m-selected--BackgroundColor: var(--pf-t--global--background--color--primary--clicked);
+  --#{$table}__tr--m-selected--BackgroundColor: var(--pf-t--global--background--color--action--plain--clicked);
   --#{$table}__tr--m-selected--OutlineOffset: calc(-1 * var(--pf-t--global--spacer--xs));
 
   // * Table tbody clickable


mcoker (commented):
I think we'll also want to update the `tbody` element, which is for expandable rows

https://github.com/jcmill/patternfly/blob/3d607faa84fe008e692810dc7a2929ee8aad7478/src/patternfly/components/Table/table.scss#L238-L242

mcoker:
This is outside of the scope of selected tables, but just poked around and compound expansion is also using the `primary--hover/clicked` backgrounds and looks wrong in a plain table on a modal.

Screenshot 2026-09-10 at 5 46 33 PM

Fixing it is straight forward, except that the [control row background color](https://github.com/jcmill/patternfly/blob/a5a99a0c120b0492d49f6b0d6bbe74f5a517e514/src/patternfly/components/Table/table.scss#L153) needs to change. It's currently using the primary background color - I'm not even sure if that's needed, but assuming it is, it probably just needs to match the rest of the table, so I'd update it to use `--#{$table}--BackgroundColor` 

Screenshot 2026-09-10 at 5 52 03 PM

mcoker (approved):
LGTM! Left a comment on the compound expansion also needing this update. Up to you if you want to do it here or in a separate issue.

lboehling (approved):
looks good! i'd love to get the update in compound expandable too like @mcoker suggested. We can do that in a follow up if you want to get this update merged though. thanks jake!

mcoker (approved):
Rad!

PR #8581, opened 2026-09-08
feat(table): add dynamic sticky footer styling

### LINKED ISSUE #8582: Table - dynamic sticky support for table footer
Table footer should also support dynamic sticky classes similar to header.

---

**Jira Issue:** [PF-4623](https://redhat.atlassian.net/browse/PF-4623)

gabipodolnikova:
## What

- Add `pf-m-sticky-footer-base` and `pf-m-sticky-footer-stuck` table modifiers.
- Add `<tfoot>` template modifier support and documentation examples.
- Match sticky footer glass styling to the sticky header, including cell surfaces, z-index, shadow, and rounded corners.
- Keep the footer visually normal when the stuck modifier is removed at the table boundary.

## Related

- Resolves #8582 
- PF-3996
- PatternFly PR #8034

## Validation

- `./node_modules/.bin/sass src/patternfly/patternfly.scss /tmp/patternfly.css`
- `git diff --check`

The React layer should toggle `pf-m-sticky-footer-stuck` only while the footer is floating.

## Summary by CodeRabbit

* **New Features**
  * Added support for sticky table footer base and stuck states.
  * Improved sticky footer styling across borders, shadows, rounded corners, backgrounds, and glass-theme presentation.
  * Updated sticky header and footer previews with consistent sizing.
* **Documentation**
  * Expanded sticky footer examples with realistic repository data, headers, and summary rows.
  * Clarified descriptions for base and stuck footer states.
  * Updated examples to demonstrate scrollable table content.

### DIFF
--- a/src/patternfly/components/Table/table.scss
+++ b/src/patternfly/components/Table/table.scss
@@ -287,6 +287,7 @@
   --#{$table}--m-sticky-footer--BoxShadow--glass: var(--pf-t--global--box-shadow--sm);
   --#{$table}--m-sticky-footer--BorderColor: var(--#{$table}--BorderColor);
   --#{$table}--m-sticky-footer-stuck--BackgroundColor: var(--#{$table}--m-sticky-footer--BackgroundColor);
+  --#{$table}--m-sticky-footer-stuck--BorderBlockStartWidth: var(--#{$table}--m-sticky-footer--BorderBlockStartWidth);
   --#{$table}--m-sticky-footer-stuck--BoxShadow: var(--#{$table}--m-sticky-footer--BoxShadow);
   --#{$table}--m-sticky-footer-stuck--BorderRadius: var(--#{$table}--m-sticky-footer--BorderRadius);
   --#{$table}--m-sticky-footer--ZIndex: calc(var(--pf-t--global--z-index--xs) + 1);
@@ -296,9 +297,9 @@
     --#{$table}--m-sticky-header--BorderBlockEndWidth: 0;
     --#{$table}--m-sticky-header--BoxShadow: var(--#{$table}--m-sticky-header--BoxShadow--glass);
     --#{$table}--m-sticky-header--BorderRadius: var(--#{$table}--m-sticky-header--BorderRadius--glass);
-    --#{$table}--m-sticky-footer--BorderBlockStartWidth: 0;
-    --#{$table}--m-sticky-footer--BoxShadow: var(--#{$table}--m-sticky-footer--BoxShadow--glass);
-    --#{$table}--m-sticky-footer--BorderRadius: var(--#{$table}--m-sticky-footer--BorderRadius--glass);
+    --#{$table}--m-sticky-footer-stuck--BorderBlockStartWidth: 0;
+    --#{$table}--m-sticky-footer-stuck--BoxShadow: var(--#{$table}--m-sticky-footer--BoxShadow--glass);
+    --#{$table}--m-sticky-footer-stuck--BorderRadius: var(--#{$table}--m-sticky-footer--BorderRadius--glass);
   }
 }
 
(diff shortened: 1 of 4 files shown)


mcoker (changes requested):
This looks really good! Looks like there were some extra styles, so I worked through it locally and put up a PR to your branch here if you want to check it out - https://github.com/gabipodolnikova/patternfly/pull/1/changes

The main updates are
* Add some example content so the table scrolls
* Use `::after` instead of `::before` for the footer styles. That seemed to fix some of the styles that were there but weren't working, then we could remove some of the extra styles that looks like they were added to fix the styles that weren't working.
* Somewhere in there we're disabling a border-block-end right before the `tfoot` - updated that to change the color to transparent (instead of the width to 0) to prevent shifting when going between stuck/not-stuck.

src/patternfly/components/Table/table.scss
```diff
@@ -425,6 +426,13 @@
     }
   }
 
+  &.pf-m-sticky-footer-stuck,
+  &.pf-m-sticky-footer:not(.pf-m-sticky-footer-base) {
```
mcoker: You shouldn't need `:not(.pf-m-sticky-footer-base)`. You should only ever have these 3 combinations of classes

* `.pf-m-sticky-footer` - regular non-dynamic sticky footer
* `.pf-m-sticky-footer-base` - dynamic sticky footer when not stuck
* `.pf-m-sticky-footer-base.pf-m-sticky-footer-stuck` - - dynamic sticky footer when stuck

src/patternfly/components/Table/table.scss
```diff
@@ -275,13 +275,31 @@
...
   --#{$table}--m-sticky-footer--ZIndex: calc(var(--pf-t--global--z-index--xs) + 1);
   --#{$table}--m-sticky-footer--border--ZIndex: calc(var(--pf-t--global--z-index--xs) + 2);
 
   :where(.pf-v6-theme-glass) & {
     --#{$table}--m-sticky-header--BorderBlockEndWidth: 0;
     --#{$table}--m-sticky-header--BoxShadow: var(--#{$table}--m-sticky-header--BoxShadow--glass);
     --#{$table}--m-sticky-header--BorderRadius: var(--#{$table}--m-sticky-header--BorderRadius--glass);
+    --#{$table}--m-sticky-footer-stuck--BorderBlockStartWidth: 0;
+    --#{$table}--m-sticky-footer-stuck--BoxShadow: var(--#{$table}--m-sticky-footer--BoxShadow--glass);
+    --#{$table}--m-sticky-footer-stuck--BorderRadius: var(--#{$table}--m-sticky-footer--BorderRadius--glass);
```
mcoker: You should be able to add these styles to the base state since the pseudo elements with the background styles aren't visible until they're stuck. This change should let you get rid of some extra styles below


```suggestion
    --#{$table}--m-sticky-footer--BorderBlockStartWidth: 0;
    --#{$table}--m-sticky-footer--BoxShadow: var(--#{$table}--m-sticky-footer--BoxShadow--glass);
    --#{$table}--m-sticky-footer--BorderRadius: var(--#{$table}--m-sticky-footer--BorderRadius--glass);
```

src/patternfly/components/Table/table.scss
```diff
@@ -392,25 +409,84 @@
...
+      opacity: 1;
+    }
+  }
+
+  &.pf-m-sticky-footer-stuck,
+  &.pf-m-sticky-footer:not(.pf-m-sticky-footer-base) {
+    --#{$table}--m-sticky-footer--BorderBlockStartWidth: var(--#{$table}--m-sticky-footer-stuck--BorderBlockStartWidth);
+    --#{$table}--m-sticky-footer--BoxShadow: var(--#{$table}--m-sticky-footer-stuck--BoxShadow);
+    --#{$table}--m-sticky-footer--BorderRadius: var(--#{$table}--m-sticky-footer-stuck--BorderRadius);
+  }
```
mcoker: You shouldn't need these now since the styles are already added to the base state.

mcoker (changes requested):
Sorry this took forever! Everything looks good but I think you can simplify it a little and remove some unnecessary code.

src/patternfly/components/Table/table.scss
gabipodolnikova: Thanks! I also found some unused styling so removed those too. I hope now it's okay 🤞 😆

mcoker (approved):
Nice!! Great job! 🚀

